Issue #16536, opened 2014-12-13
Response to ask through actor selection times out


kosii:
The code below always times out:

```scala
implicit val t = Timeout(1 seconds)

val system = ActorSystem("system")
import system.dispatcher

system.actorOf(
  Props(
    new Actor {
      override def receive: Actor.Receive = {
        case r => context.actorSelection(sender.path) ! r
      }
    }
  )
) ? "hello" onComplete println
```

As far as I understood it's because the `VirtualPathContainer` does not reimplement the `receive` method. Is it a bug or it supposed to behave like this?
I did a quick fix and some unit tests. It apparently did not break existing tests, but I can't really evaluate the possible consequences and the correctness (for example I don't really know what `wildcardFanOut` does in an `ActorSelectionMessage`).

I implemented `receive` in `VirtualPathContainer` like this:
```scala
override def !(message: Any)(implicit sender: ActorRef = Actor.noSender): Unit = message match {
  case ActorSelectionMessage(msg, elements, _) ⇒ {
    elements.foreach {
      case SelectChildName(name)       ⇒ getChild(name).tell(msg, sender)
      case SelectChildPattern(pattern) ⇒ // no idea
      case SelectParent                ⇒ getParent.tell(msg, sender)
    }
  }
}
```

rkuhn:
Thanks for finding and raising this! Fixing this ticket could go either way:

* if it shall not work, then we need to reflect that in the docs (ScalaDoc and reference docs)
* if it shall work then we need to fix the code

Could you open a PR with your “quick fix” so we can have a look at what’s involved?

kosii:
Sure, I'll do it this evening

bantonsson:
Fixed by #16690

kosii:
thanks for all the help, it was fun!

PR #16690, opened 2015-01-20
=act #16536 actor selection now completes ask

kosii:


kosii:
#16536

drewhk:
OK TO TEST

akka-actor/src/main/scala/akka/actor/ActorRef.scala
```diff
@@ -599,6 +599,16 @@ private[akka] class VirtualPathContainer(
 
   private val children = new ConcurrentHashMap[String, InternalActorRef]
 
+  override def !(message: Any)(implicit sender: ActorRef = Actor.noSender): Unit = message match {
+    case ActorSelectionMessage(msg, elements, _) ⇒ {
+      elements.foreach {
+        case SelectChildName(name)       ⇒ getChild(name).tell(msg, sender)
+        case SelectChildPattern(pattern) ⇒ // no idea
```
drewhk: This sould deliver the message to all children whose name matches the pattern. This is best effort, so no problem if the ConcurrentHashMap actually changes in-between.

drewhk:
Thanks for the contribution

kosii:
Thanks, I'll complete the `SelectChildPattern(pattern)` case

kosii:
cannot reproduce the test failures in local

kosii:
PLS BUILD

akka-actor/src/main/scala/akka/actor/ActorRef.scala
```diff
@@ -599,6 +599,23 @@ private[akka] class VirtualPathContainer(
...
+    case ActorSelectionMessage(msg, elements, _) ⇒ {
+      elements.foreach {
+        case SelectChildName(name) ⇒ getChild(name).tell(msg, sender)
+        case pattern @ SelectChildPattern(_) ⇒ {
+          import scala.collection.JavaConverters.mapAsScalaConcurrentMapConverter
+          children.asScala filter {
+            case (name, child) ⇒ pattern.pattern.matcher(name).matches
+          } foreach {
+            case (name, child) ⇒ child.tell(msg, sender)
+          }
```
2beaucoup:
```scala
children.asScala foreach {
  case (name, child) ⇒ if (pattern.pattern.matcher(name).matches) child.tell(msg, sender)
}
```

akka-actor/src/main/scala/akka/actor/ActorRef.scala
```diff
@@ -599,6 +599,23 @@ private[akka] class VirtualPathContainer(
...
+        }
+        case SelectParent ⇒ getParent.tell(msg, sender)
+      }
```
patriknw: what happens if it is not an ActorSelectionMessage?

akka-actor/src/main/scala/akka/actor/ActorRef.scala
```diff
@@ -599,6 +599,23 @@ private[akka] class VirtualPathContainer(
 
   private val children = new ConcurrentHashMap[String, InternalActorRef]
 
+  override def !(message: Any)(implicit sender: ActorRef = Actor.noSender): Unit = message match {
+    case ActorSelectionMessage(msg, elements, _) ⇒ {
+      elements.foreach {
```
patriknw: is it not possible to use `ActorSelection.deliverSelection` from here?

akka-actor/src/main/scala/akka/actor/ActorRef.scala
kosii: thanks!

akka-actor/src/main/scala/akka/actor/ActorRef.scala
patriknw: I guess not

akka-actor/src/main/scala/akka/actor/ActorRef.scala
```diff
@@ -599,6 +599,23 @@ private[akka] class VirtualPathContainer(
...
+        case SelectChildName(name) ⇒ getChild(name).tell(msg, sender)
+        case pattern @ SelectChildPattern(_) ⇒ {
```
patriknw: I think it is questionable if we want to support wildcards here?

akka-actor/src/main/scala/akka/actor/ActorRef.scala
```diff
@@ -599,6 +599,23 @@ private[akka] class VirtualPathContainer(
...
+          import scala.collection.JavaConverters.mapAsScalaConcurrentMapConverter
+          children.asScala filter {
+            case (name, child) ⇒ pattern.pattern.matcher(name).matches
+          } foreach {
+            case (name, child) ⇒ child.tell(msg, sender)
+          }
+        }
+        case SelectParent ⇒ getParent.tell(msg, sender)
```
patriknw: this not correct, e.g. `"/temp/../temp/$a"

it should only send `msg` to parent if sel.elements.drop(1).isEmpty otherwise it should send `ActorSelectionMessage(sel.elements.drop(1), msg, sender)`

patriknw: questionable if we should support `..` here?

akka-actor/src/main/scala/akka/actor/ActorRef.scala
kosii: I don't know, but if I recall correctly then the execution ends up here already from an `ActorSelection.deliverSelection`

akka-actor/src/main/scala/akka/actor/ActorRef.scala
kosii: I don't know, does receiving a user defined message or some other subclass of `AutoReceivedMessage` in a `VirtualPathContainer` make sense?

akka-actor/src/main/scala/akka/actor/ActorRef.scala
patriknw: yeah, ActorSelection.deliverSelection is for the actor cell things, and this is not

akka-actor/src/main/scala/akka/actor/ActorRef.scala
patriknw: maybe not now, but I think you should have a default case that delegates to `super.!` to avoid potential match error

akka-actor/src/main/scala/akka/actor/ActorRef.scala
kosii: can I just safely ignore it right along with `..` for now?

akka-actor/src/main/scala/akka/actor/ActorRef.scala
rkuhn: I somehow cannot figure out why `foreach` on the elements makes sense here—each element descends one level into the tree …

akka-actor-tests/src/test/scala/akka/pattern/AskSpec.scala
```diff
@@ -111,6 +111,45 @@ class AskSpec extends AkkaSpec {
...
+    }
+
+    "work when reply uses actor selection wildcard pattern" in {
+      implicit val timeout = Timeout(5 seconds)
+      import system.dispatcher
+      val echo = system.actorOf(Props(new Actor {
+        def receive = {
+          case x ⇒
+            val path = sender().path
+            context.actorSelection(path / ".." / "*") ! x
```
rkuhn: This will complete all outstanding “ask” operations in the system (that had their paths registered), so I don’t think this is a valid test.

drewhk: Except VirtualPathContainers that didn't forward to children?

drewhk: It is easy to test if the test fails without the patch.

akka-actor/src/main/scala/akka/actor/ActorRef.scala
drewhk: ah yes, it should be the head. Maybe also rename "elements" to path-elements or something more eye-catching.

akka-actor-tests/src/test/scala/akka/pattern/AskSpec.scala
rkuhn: Or rather: set up two “asks”, look at their path to have it registered, then verify that both are completed by a single reply. But I still find this ability disturbing.

akka-actor-tests/src/test/scala/akka/pattern/AskSpec.scala
```diff
@@ -111,6 +111,45 @@ class AskSpec extends AkkaSpec {
...
+    }
+
+    "work when reply uses actor selection parent pattern" in {
+      implicit val timeout = Timeout(5 seconds)
+      import system.dispatcher
+      val echo = system.actorOf(Props(new Actor {
+        def receive = {
+          case x ⇒
+            val path = sender().path
+            context.actorSelection(path / ".." / path.name) ! x
```
patriknw: This test only works because of the wrong `elements.foreach`.
`..` and wildcard should not be supported for `/temp` as far as I can see.

akka-actor/src/main/scala/akka/actor/ActorRef.scala
kosii: Thanks for pointing this out, it seemed to work, so I just took it for granted without looking further into (my initial thoughts were that the iterator contains the names of the  children already resolved by `ActorSelection.deliverSelection` for example in case of wildcard elements in the path). I'll fix it soon

akka-actor-tests/src/test/scala/akka/pattern/AskSpec.scala
kosii: I'll fix the implementation and change the tests accordingly

akka-actor/src/main/scala/akka/actor/ActorRef.scala
```diff
@@ -599,6 +599,22 @@ private[akka] class VirtualPathContainer(
...
+    case sel @ ActorSelectionMessage(msg, elements, _) ⇒ {
+      elements.head match {
+        case SelectChildName(name) ⇒
+          val child = getChild(name)
+          if (elements.tail.isEmpty) {
+            child ! msg
+          } else {
+            child ! sel.copy(elements = elements.tail)
+          }
+        case _ ⇒ // .. and * are unsupported
```
viktorklang: Isn't this going to be a bit of a surprise?

rkuhn: please send to deadLetters via an EmptyLocalActorRef as shown in `ActorSelection.deliverSelection` (i.e. conditional on the `wildcardFanout` flag that you’ll also need to extract from the match above)

rkuhn:
Touching things in the internals can feel a little scary but I think this is going to work. If the tests pass with the `require` clause I suggested then we can also pull this into the `release-2.3` branch and include it in Akka 2.3.10. So, to summarize the next steps:

* please add the sending to EmptyLocalActorRef and the `require` clause
* when tests pass, squash and rebase
* once merged, cherry-pick onto release-2.3 and open a new PR

Thanks!

akka-actor/src/main/scala/akka/actor/ActorRef.scala
rkuhn: @viktorklang (concurrent review ;-) ) This took 3 years to come up at all, so I think it makes sense to allow the “obvious” case that is expressed in the tests but still make it impossible to complete all currently outstanding `ask` operations in one fell swoop. Sending to deadLetters in this case will document what is happening.

viktorklang: @rkuhn :) but supporting the wildcard ought to be quite straightforward?

rkuhn: Do you think it should be allowed to send to `/temp/*`? I think that would have non-intuitive consequences and count as “surprising”, more so than explicitly disallowing it. (and yes, it was even implemented earlier in this PR, so it definitely is possible)

viktorklang: @rkuhn I'm all for not supporting it, but I think it should be made visible that it isn't allowed, perhaps that is what you meant by the `require`?

rkuhn: Since `tell` cannot throw the only thing we can do is to send to deadLetters (which I proposed in this comment thread already), unless I overlooked something.

viktorklang: Logging?

-- 
Cheers,
√

rkuhn: Well, that is the purpose of sending to deadLetters, is it not? So far we have used this approach consistently for all message sends that cannot be completed (wherever possible). Do you propose that we change this to print more targeted warnings in all cases? The other consideration is that messages ending up in deadLetters is unavoidable in some cases and therefore we allow people to switch off all those cases with a single configuration setting.

viktorklang: No, if we send to DeadLetters there's no way for the user to know why it ended up there?

viktorklang: Or did you propose we add a wrapper that explains it?

rkuhn: In no case do we have an explanation of how the message ended up in deadLetters, if we add one then we need to do that everywhere. I’m not saying that that is a bad idea, but it sure is not something to be done in this PR. We should open a new ticket about that part and use EmptyLocalActorRef here as I suggested. Then, if we enhance ELAR to include a reason we can add an appropriate one in this instance, too.

viktorklang: No, my point was that we can log AND send to dead letters.

rkuhn: No, double logging is definitely a bad idea, and for that reason we don't do that anywhere else.

viktorklang: Fair enough. So DL plus adequate docs?

-- 
Cheers,
√

rkuhn: Yup :-)

kosii: To get a reference on `EventStream` can I use `provider.systemGuardian.underlying.system.eventStream` or is there a better way?
The other way I found was matching on getParent, but I think it's ugly
```scala
getParent match {
  case refWithCell: ActorRefWithCell ⇒ {
    val emptyRef = new EmptyLocalActorRef(provider, path / sel.elements.map(_.toString), refWithCell.underlying.system.eventStream)
    emptyRef.tell(msg, sender)
  }
  case _ ⇒
    this.provider.deadLetters.tell(msg, sender)
}
```

viktorklang: Hi @kosii,

given that the `VirtualPathContainer` is marked as Akka-private I think it's more sensible to adjust its constructor to also take an `EventStream`. @rkuhn?

rkuhn: If only it were that easy: start-up order is a _pain_. Matching on parent is probably a tad more brittle than taking the `systemGuardian.underlying` route.

viktorklang: In this case I think `eventStream` is available at all time `VirtualPathContainer` is instantiated.

akka-actor/src/main/scala/akka/actor/ActorRef.scala
```diff
@@ -599,6 +599,31 @@ private[akka] class VirtualPathContainer(
...
+   * In [[ActorSelectionMessage]]s only [[SelectChildName]] elements
+   * are supported, otherwise messages are sent to [[EmptyLocalActorRef]].
+   */
+  override def !(message: Any)(implicit sender: ActorRef = Actor.noSender): Unit = message match {
+    case sel @ ActorSelectionMessage(msg, elements, wildcardFanOut) ⇒ {
+      require(elements.nonEmpty)
+
+      elements.head match {
+        case SelectChildName(name) ⇒
+          val child = getChild(name)
```
patriknw: `getChild` will return `null` if the named child doesn't exist, i.e. NPE will be thrown below

you need to check for null, and send to EmptyLocalActorRef if that happens

akka-actor/src/main/scala/akka/actor/ActorRef.scala
```diff
@@ -599,6 +599,31 @@ private[akka] class VirtualPathContainer(
...
+          if (elements.tail.isEmpty) {
+            child ! msg
+          } else {
+            child ! sel.copy(elements = elements.tail)
```
patriknw: we can't support this case either, because this will complete the Future with the remaining ActorSelection message, i.e.  EmptyLocalActorRef here also

test case: `context.actorSelection("/temp/" + name + "/missing") ! x`

patriknw:
@kosii I found two bugs, but I think we are close now :)

kosii:
@patriknw thanks for pointing these out! I fixed it this morning and I'll finalize my commit this evening

kosii:
the `wildcardFanOut`  flag is only used to prevent the user from flooding the system with `DeadLetter`s by doing stuff like `context.actorSelection("/*/*/*") ! "Hi!"` ?

patriknw:
@kosii yes, that is true, you should only send to the `emptyRef` when `!sel.wildcardFanOut`
not sure if that can happen, but doesn't hurt to use that flag here in the same way as in other places

kosii:
does it look alright?

patriknw:
LGTM, thanks

bantonsson:
LGTM, thank you @kosii

rkuhn:
LGTM

kosii:
cool, I rebased onto the current master

Issue #13866, opened 2014-02-07
ConcurrentModificationException in java.util.Map MDC


patriknw:
imported from https://www.assembla.com/spaces/akka/tickets/3866

User report: https://groups.google.com/d/msg/akka-user/3JrDRdCQiIA/R_5c8cVlF-kJ

We should create a copy when converting to the scala immutable Map in DiagnosticLoggingAdapter.setMDC

hepin1989:
one for release-2.3 and one for master

rkuhn:
Upon closer inspection this ticket is not valid: we are indeed already making a copy during `setMDC`, which is also clearly visible in the original mailing list report. The CME was caused by the user through concurrent updates to the map, that is not something we can fix.

hepin1989:
I don't think so.the scala api accept a immutable map.but the java one just accept a Map,which maybe a mutable one.

rkuhn:
What I meant was that `jMdc.asScala.toMap` already does a copy, so there is nothing further for us to fix.

hepin1989:
@rkuhn yes.I don't realize that.thanks,then should close them :P

patriknw:
my bad when creating the ticket, thanks

hepin1989:
my bad too for don't notice the ```toMap``` call ...

PR #16691, opened 2015-01-21
=act #13866 fix ConcurrentModificationException in java.util.Map MDC

He-Pin:
fix this https://github.com/akka/akka/issues/13866 for release 2.3

### DIFF
--- a/akka-actor/src/main/scala/akka/event/Logging.scala
+++ b/akka-actor/src/main/scala/akka/event/Logging.scala
@@ -3,20 +3,20 @@
  */
 package akka.event
 
-import language.existentials
-import akka.actor._
-import akka.{ ConfigurationException, AkkaException }
-import akka.actor.ActorSystem.Settings
-import akka.util.ReentrantGuard
-import java.util.concurrent.atomic.AtomicInteger
+import java.util.Locale
 import java.util.concurrent.TimeoutException
+import java.util.concurrent.atomic.AtomicInteger
+
+import akka.actor.ActorSystem.Settings
+import akka.actor._
+import akka.util.ReentrantGuard
+import akka.{ AkkaException, ConfigurationException }
+
 import scala.annotation.implicitNotFound
 import scala.collection.immutable
-import scala.concurrent.duration._
 import scala.concurrent.Await
-import scala.util.control.NoStackTrace
-import scala.util.control.NonFatal
-import java.util.Locale
+import scala.language.existentials
+import scala.util.control.{ NoStackTrace, NonFatal }
 
 /**
  * This trait brings log level handling to the EventStream: it reads the log
@@ -234,7 +234,9 @@ trait LoggingBus extends ActorEventBus {
  */
 @implicitNotFound("Cannot find LogSource for ${T} please see ScalaDoc for LogSource for how to obtain or construct one.") trait LogSource[-T] {
   def genString(t: T): String
+
   def genString(t: T, system: ActorSystem): String = genString(t)
+
   def getClazz(t: T): Class[_] = t.getClass
 }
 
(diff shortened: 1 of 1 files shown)


drewhk:
OK TO TEST

hepin1989:
failed by SurviveNetworkInstabilitySpec

akka-actor/src/main/scala/akka/event/Logging.scala
```diff
@@ -1116,7 +1116,7 @@ trait DiagnosticLoggingAdapter extends LoggingAdapter {
    * These values can be used in PatternLayout when [[akka.event.slf4j.Slf4jLogger]] is configured.
    * Visit <a href="http://logback.qos.ch/manual/mdc.html">Logback Docs: MDC</a> for more information.
    */
-  def setMDC(jMdc: java.util.Map[String, Any]): Unit = mdc(if (jMdc != null) jMdc.asScala.toMap else emptyMDC)
+  def setMDC(jMdc: java.util.Map[String, Any]): Unit = mdc(if (jMdc != null) jMdc.asScala.foldLeft(Map.empty[String, Any])((map, kv) ⇒ map + kv) else emptyMDC)
```
patriknw: alternative: `.toSeq.toMap` ?

hepin1989: which one is better ,from the perspective of efficient?

patriknw: more efficient would probably be to skip the `toScala` and iterate `jMdc` entries and convert them to tuples (or use MapBuilder)

drewhk: maybe it is worth here to do the extra effort.

drewhk:
Thanks for the contribution, LGTM!

drewhk:
PLS BUILD

akka-actor/src/main/scala/akka/event/Logging.scala
hepin1989: like this?
```scala
  def setMDC(jMdc: java.util.Map[String, Any]): Unit = {
    val mdcValue = if (jMdc != null) {
      val iterator = jMdc.entrySet().iterator()
      var builder = Map.newBuilder[String, Any]
      while (iterator.hasNext) {
        val entry = iterator.next()
        builder += (entry.getKey -> entry.getValue)
      }
      builder.result()
    } else emptyMDC
    mdc(mdcValue)
  }
```

drewhk: :+1:

patriknw: :+1:

hepin1989: OK,I will update then,thanks

akka-actor/src/main/scala/akka/event/Logging.scala
```diff
@@ -1116,7 +1241,18 @@ trait DiagnosticLoggingAdapter extends LoggingAdapter {
...
-  def setMDC(jMdc: java.util.Map[String, Any]): Unit = mdc(if (jMdc != null) jMdc.asScala.toMap else emptyMDC)
+  def setMDC(jMdc: java.util.Map[String, Any]): Unit = {
+    val mdcValue = if (jMdc != null) {
+      val iterator = jMdc.entrySet().iterator()
+      val builder = Map.newBuilder[String, Any]
+      while (iterator.hasNext) {
+        val entry = iterator.next()
+        builder += (entry.getKey -> entry.getValue)
+      }
+      builder.result()
```
rkuhn: This is exactly what `jMdc.asScala.toMap` ends up doing, so I prefer the shorter version for readability.

rkuhn:
Thanks for your work and energy, but I’m sorry to say that the ticket itself was faulty (as commented there).

akka-actor/src/main/scala/akka/event/Logging.scala
hepin1989: I miss the toMap call and think there is just asScala.
will close this.thanks

hepin1989:
sorry for the misunderstanding .I am blind the toMap call

Issue #16693, opened 2015-01-21
add sbt-updates to the plugins.sbt


hepin1989:
add this plugin will easy akka dependencies management

hepin1989:
close it now

PR #16694, opened 2015-01-21
=pro #16693 add sbt-updates to plugins.sbt

He-Pin:
1 change the typesafeResolver to typesafeReleases
2 add sbt-updates as a new plugin

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,4 +1,4 @@
-resolvers += Classpaths.typesafeResolver
+resolvers += Classpaths.typesafeReleases
 
 resolvers += "jgit-repo" at "http://download.eclipse.org/jgit/maven"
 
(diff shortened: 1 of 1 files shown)


project/plugins.sbt
```diff
@@ -25,5 +25,7 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-git" % "0.6.2")
 
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.1.6")
 
+addSbtPlugin("com.timushev.sbt" % "sbt-updates" % "0.1.7")
```
ktoso: Isn't that one of these plugins better kept globally in your `~/.sbt/0.13/...` files?

hepin1989: Hard to say,but i think this one is out of box with minimal effort.

2m:
I also think this one is better in global configuration, because you can use this plugin potentially on any sbt project.

2m:
I am closing this. Thanks for the gathering the dependencies that need to be updated #16695.

2m:
Refs #16693

Issue #16609, opened 2014-12-29
akka.stream: missing case in FlowGraphBuilder.addEdge(junctionOut, flow, sink) ?


suiryc:
Hi,

Running a graph linking a `Junction` to a `GraphFlow` then a `Sink` fails at runtime with `IllegalArgumentException`.
If using a standard `Flow` instead of a `GraphFlow`, or if using a `Source` instead of a `Junction`, or if using an `UndefinedSink` instead of a `Sink`, graph runs without issue.

Example:

    val gflow = Flow[Int, Int]() { implicit builder =>
      import FlowGraphImplicits._

      val in = UndefinedSource[Int]
      val out = UndefinedSink[Int]

      in ~> out

      (in, out)
    }
    val sink = Sink.fold[Int, Int](0)(_ + _)
    val graph = FlowGraph { implicit builder =>
      import FlowGraphImplicits._

      val merge = Merge[Int]

      Source(List(1, 2, 3)) ~> merge
      Source.empty[Int] ~> merge
      merge ~> gflow ~> sink
    }

    graph.run()

Fails with
```
Exception in thread "main" java.lang.IllegalArgumentException: Unsupported value [(GraphFlow(Pipe(List(),List(),OperationAttributes(List())),akka.stream.scaladsl.UndefinedSource@442675e1,akka.stream.scaladsl.PartialFlowGraph@6166e06f,akka.stream.scaladsl.UndefinedSink@49e202ad,Pipe(List(),List(),OperationAttributes(List()))),FoldSink(0))] of type [scala.Tuple2]. Only Pipes and Graphs are supported!
	at akka.stream.scaladsl.FlowGraphInternal$.throwUnsupportedValue(FlowGraph.scala:436)
	at akka.stream.scaladsl.FlowGraphBuilder.addEdge(FlowGraph.scala:705)
	at akka.stream.scaladsl.FlowGraphImplicits$JunctionNextStep.$tilde$greater(FlowGraph.scala:1453)
...
```
Note: here the `Flow` does nothing, actual implementation could do work balancing as detailed in the [cookbook](http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0-M2/scala/stream-cookbook.html#Balancing_jobs_to_a_fixed_pool_of_workers).

So indeed `FlowGraphBuilder.addEdge(junctionOut, flow, sink)` (other than the `UndefinedSink` version) does not handle this particular case.
[...]

PR #16702, opened 2015-01-22
=str #16609 add missing handling of FlowGraphs when connecting

ktoso:
Solves #16609 already, but I think we should also skim over the code to make sure we're not still missing some of the possible combinations.

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowGraphCompileSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowGraphCompileSpec.scala
@@ -440,6 +440,32 @@ class FlowGraphCompileSpec extends AkkaSpec {
         }
       }
 
+      "Junction is connected through GraphFlow" in {
+        val gflow = Flow[Int, String]() { implicit builder ⇒
+          import FlowGraphImplicits._
+
+          val in = UndefinedSource[Int]
+          val out = UndefinedSink[String]
+
+          in ~> Flow[Int].map(_.toString) ~> out
+
+          (in, out)
+        }
+
+        val sink = Sink.fold[Int, Int](0)(_ + _)
+        val graph = FlowGraph { implicit builder ⇒
+          import FlowGraphImplicits._
+
+          val merge = Merge[Int]
+
+          Source(List(1, 2, 3)) ~> merge
+          Source.empty[Int] ~> merge
+          merge ~> gflow.map(_.toInt) ~> sink
+        }
+
+        graph.run()
+      }
+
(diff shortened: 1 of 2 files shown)


patriknw:
LGTM

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowGraphCompileSpec.scala
```diff
@@ -440,6 +440,32 @@ class FlowGraphCompileSpec extends AkkaSpec {
...
+      "Junction is connected through GraphFlow" in {
+        val gflow = Flow[Int, String]() { implicit builder ⇒
+          import FlowGraphImplicits._
+
+          val in = UndefinedSource[Int]
+          val out = UndefinedSink[String]
+
+          in ~> Flow[Int].map(_.toString) ~> out
+
+          (in, out)
```
drewhk: looking at this pair feels confusing every time. Maybe we should have a case class instead of a raw pair? At least when using a case class the parameters can be named.

drewhk:
LGTM

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowGraphCompileSpec.scala
ktoso: Maybe `Flow(in: UndefinedSource[_], out: UndefinedSink)` would be an intuitive thing to return here?

// cc @patriknw @rkuhn

patriknw: um, it would also avoid `in -> out`, which almost looks like `in ~> out`

first we must come up with a name for the case class, then we can consider the `Flow.apply` for creating it (I'm not convinced about the latter)

patriknw: `FlowEndpoints` ?

ktoso: Sounds allright I guess. Another idea `ExposedFlow(in, out)`?

drewhk: Ports(fromUpstream, toDownstream)? It will look like this:

    Ports(fromUpstream = in, toDownstream = out)

ktoso: Made it a separate ticket https://github.com/akka/akka/issues/16715

Issue #16577, opened 2014-12-18
FlexiMerge completionHandling doesn't catch thrown exceptions


sirthias:
Throwing an exception in the CompletionHandling::onComplete handler results in

```
java.lang.IllegalStateException: Processor actor terminated abruptly
	at akka.stream.impl.FanIn.postStop(FanIn.scala:224)
	at akka.actor.Actor$class.aroundPostStop(Actor.scala:475)
...
```

l15k4:
Hey,

I'm getting this exception when `process` Flow throws FlowException that should be handled by decider and the element should be skipped : 
```scala
  val decider: Supervision.Decider = {
    case FlowException(_) =>
      Supervision.Resume
    case ex =>
      log.error(ex, "Unexpected exception in flow execution occurred")
      Supervision.Stop
  }

  implicit val materializer = ActorFlowMaterializer(
    ActorFlowMaterializerSettings(context.system).withSupervisionStrategy(decider)
  )

source.via(process).via(store).runWith(sink)
```

But instead this happens : 

```
[info]     Cause: java.lang.IllegalStateException: Processor actor terminated abruptly
[info]     at akka.stream.impl.fusing.ActorInterpreter.postStop(ActorInterpreter.scala:288)
[info]     at akka.actor.Actor$class.aroundPostStop(Actor.scala:475)
[info]     at akka.stream.impl.fusing.ActorInterpreter.aroundPostStop(ActorInterpreter.scala:271)
[info]     at akka.actor.dungeon.FaultHandling$class.akka$actor$dungeon$FaultHandling$$finishTerminate(FaultHandling.scala:210)
[info]     at akka.actor.dungeon.FaultHandling$class.terminate(FaultHandling.scala:172)
[info]     at akka.actor.ActorCell.terminate(ActorCell.scala:369)
[info]     at akka.actor.ActorCell.invokeAll$1(ActorCell.scala:462)
[info]     at akka.actor.ActorCell.systemInvoke(ActorCell.scala:478)
[info]     at akka.dispatch.Mailbox.processAllSystemMessages(Mailbox.scala:279)
[info]     at akka.dispatch.Mailbox.run(Mailbox.scala:220)
[info]     at akka.dispatch.Mailbox.exec(Mailbox.scala:231)
[info]     at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
[...]

PR #16708, opened 2015-01-23
=str #16577 emit user exception from Flexi* completion handler

patriknw:


ktoso:
LGTM

akka-stream/src/main/scala/akka/stream/impl/FlexiMergeImpl.scala
```diff
@@ -44,7 +44,10 @@ private[akka] class FlexiMergeImpl(_settings: MaterializerSettings,
 
   override protected val inputBunch = new FanIn.InputBunch(inputPorts, settings.maxInputBufferSize, this) {
     override def onError(input: Int, e: Throwable): Unit = {
-      changeBehavior(completion.onError(ctx, inputMapping(input), e))
+      changeBehavior(try completion.onError(ctx, inputMapping(input), e)
+      catch {
+        case NonFatal(e) ⇒ fail(e); mergeLogic.SameState
```
drewhk: can this lead to double onError signals?

drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/impl/FlexiMergeImpl.scala
patriknw: no, inputBunch and primaryOutputs have protection flag for that, so it is idempotent

patriknw:
Refs #16577

PR #16654, opened 2015-01-15
NOT FOR MERGE! #16653 Reproducer for form upload hang

bantonsson:
Refs #16653

Issue #16653, opened 2015-01-15
HTTP Form upload hangs if the server has been port scanned with nc (GNU netcat 0.7.1)


bantonsson:
So we now have an issue where if we check for server availability from the outside by using nc, the following curl upload will fail.

This fails on Linux with GNU netcat 0.7.1 and on OS X if you install GNU netcat by doing `brew install netcat`. The default OS X nc command doesn't provoke this behavior.

Reproducer coming shortly.

/cc @sirthias @jrudolph

patriknw:
merged to master and release-2.3

PR #16709, opened 2015-01-23
=act #16653 swallow exception from setTcpNoDelay

patriknw:
- some versions of netcat (`nc -z`) doesn't allow setTcpNoDelay

@bantonsson provided an excellent reproducer in https://github.com/akka/akka/pull/16654

### DIFF
--- a/akka-actor/src/main/scala/akka/io/TcpConnection.scala
+++ b/akka-actor/src/main/scala/akka/io/TcpConnection.scala
@@ -181,7 +181,12 @@ private[io] abstract class TcpConnection(val tcp: TcpExt, val channel: SocketCha
   def completeConnect(registration: ChannelRegistration, commander: ActorRef,
                       options: immutable.Traversable[SocketOption]): Unit = {
     // Turn off Nagle's algorithm by default
-    channel.socket.setTcpNoDelay(true)
+    try channel.socket.setTcpNoDelay(true) catch {
+      case e: SocketException ⇒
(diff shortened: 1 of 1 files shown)


patriknw:
Refs #16653

akka-actor/src/main/scala/akka/io/TcpConnection.scala
```diff
@@ -181,7 +181,12 @@ private[io] abstract class TcpConnection(val tcp: TcpExt, val channel: SocketCha
   def completeConnect(registration: ChannelRegistration, commander: ActorRef,
                       options: immutable.Traversable[SocketOption]): Unit = {
     // Turn off Nagle's algorithm by default
-    channel.socket.setTcpNoDelay(true)
+    try channel.socket.setTcpNoDelay(true) catch {
+      case e: SocketException ⇒
+        // as reported in #16653 some versions of netcat (`nc -z`) doesn't allow setTcpNoDelay
```
drewhk: Hm. Isn't the problem maybe that nc immediately closes the connection after a handshake? Can't this option set pre-connect?

drewhk:
I guess LGTM as a workaround, but would be nice to know the real cause.

drewhk:
This might be relevant: http://bugs.java.com/view_bug.do?bug_id=7014244

patriknw:
ok, another reason to continue anyway if that setting fails

drewhk:
agreed

rkuhn:
LGTM (there’s nothing else we can do)

Issue #15349, opened 2014-06-04
Sanitize error logging in processor


drewhk:
Currently it logs both on failed behavior (the processor introduced the failure) and on upstream onError (the processor has been notified of failure). These two scenarios should not be mixed up.

PR #16710, opened 2015-01-23
=str #15349 Sanitize error logging

patriknw:


akka-stream-testkit/src/test/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -3,6 +3,7 @@
  */
 package akka.stream.testkit
 
+import scala.language.existentials
```
drewhk: nice!

akka-stream/src/main/scala/akka/stream/impl/ActorProcessor.scala
```diff
@@ -250,7 +250,8 @@ private[akka] abstract class ActorProcessorImpl(val settings: MaterializerSettin
   protected def onError(e: Throwable): Unit = fail(e)
 
   protected def fail(e: Throwable): Unit = {
-    log.error(e, "failure during processing") // FIXME: escalate to supervisor instead
+    // FIXME: escalate to supervisor
```
drewhk: I am not sure about the escalate part. Also we have a ticket about supervision, this FIXME is therefore superceded by that.

akka-stream/src/main/scala/akka/stream/impl/ActorProcessor.scala
```diff
@@ -250,7 +250,8 @@ private[akka] abstract class ActorProcessorImpl(val settings: MaterializerSettin
...
+    log.debug("fail {} due to: {}", self, e.getMessage)
```
drewhk: hmm, maybe we want to protect this debug output with a config option: "stream.debug-log-all-failures" or something like that.

drewhk:
If we want these debug logs, then this PR needs to cover ActorInterpreter, too. I am not sure about the inheritance hierarchy but a quick check of FanIn, FanOut and the StreamOfStreamX actors might be needed, too. I am not sure this debug output is that useful though, but have no strong opinion about it.

akka-stream/src/main/scala/akka/stream/impl/ActorProcessor.scala
patriknw: well, I only kept the FIXME, and it will be revisited by the ticket you mention

akka-stream/src/main/scala/akka/stream/impl/ActorProcessor.scala
patriknw: @drewhk yes, I'm not sure. Shall I remove them and we revisit the whole debuggability later?

akka-stream/src/main/scala/akka/stream/impl/ActorProcessor.scala
drewhk: ok

akka-stream/src/main/scala/akka/stream/impl/ActorProcessor.scala
drewhk: I think it is fine to put this feature behind a config flag.

drewhk: But maybe best to add a comment on the debug ticket and remove the logs from here

patriknw:
@drewhk I have added the setting for extra debug logging, see a2f931a. It doesn't mean that the debugging is a solved problem, but at least the failures are visible if needed.

drewhk:
Fine by me. LGTM

patriknw:
@ktoso are you also fine with this one, so I can get it out the way?

akka-stream/src/main/resources/reference.conf
```diff
@@ -38,6 +38,9 @@ akka {
       # Fully qualified config path which holds the dispatcher configuration
       # to be used by FlowMaterialiser when creating Actors for IO operations.
       file-io-dispatcher = ${akka.io.tcp.file-io-dispatcher}
+      
+      # Enable additional troubleshooting logging at DEBUG log level
+      debug-logging = off
```
ktoso: yeah, this makes sense +1

ktoso:
LTGM, the additional flag makes sense

patriknw:
thanks for quick response

patriknw:
Refs #15349

PR #16633, opened 2015-01-10
+ akka-persistence: add internal message protocol traits, resolve #16612

Andrei-Pozolotin:

Issue #16612, opened 2015-01-04
akka-persistence: JournalProtocol/SnapshotProtocol should extend 3 marker traits


Andrei-Pozolotin:
Currently [JournalProtocol](https://github.com/akka/akka/blob/master/akka-persistence/src/main/scala/akka/persistence/JournalProtocol.scala) and [SnapshotProtocol] (https://github.com/akka/akka/blob/master/akka-persistence/src/main/scala/akka/persistence/Snapshot.scala#L115) do not extend any  marker traits.

This is problematic for persistence plugin development, when custom Journal and Snapshot actors
need process other in-band commands, such as cluster replication.

[For example, here](https://github.com/carrot-garden/akka-persistence-chronicle/blob/master/src/main/scala/akka/persistence/chronicle/Journal.scala#L273) wildcard case will potentially forward wrong commands down to the persistence supertype.

Solution is to provide 3 marker traits and have relevant messages extend them:
```
sealed trait PersistenceCommand
sealed trait PersistenceJournalCommand
sealed trait PersistenceSnapshotCommand
```

patriknw:
sounds like a good addition, thanks

Andrei-Pozolotin:
see https://github.com/akka/akka/pull/16633

PR #16711, opened 2015-01-23
+ akka-persistence: add internal message protocol traits

Andrei-Pozolotin:
2 more snapshot messages used for internal plugin acknowledgement in
https://github.com/akka/akka/blob/master/akka-persistence/src/main/scala/akka/persistence/snapshot/SnapshotStore.scala
- see #16612 
- see #16633

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/SnapshotProtocol.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/SnapshotProtocol.scala
@@ -24,6 +24,7 @@ final case class SnapshotMetadata(persistenceId: String, sequenceNr: Long, times
  */
 @SerialVersionUID(1L)
 final case class SaveSnapshotSuccess(metadata: SnapshotMetadata)
+  extends SnapshotProtocol.Response
 
 /**
  * Sent to a [[PersistentActor]] after failed saving of a snapshot.
@@ -33,6 +34,7 @@ final case class SaveSnapshotSuccess(metadata: SnapshotMetadata)
  */
 @SerialVersionUID(1L)
 final case class SaveSnapshotFailure(metadata: SnapshotMetadata, cause: Throwable)
+  extends SnapshotProtocol.Response
 
 /**
  * Offers a [[PersistentActor]] a previously saved `snapshot` during recovery. This offer is received


Andrei-Pozolotin:
@patriknw please review

ktoso:
OK TO TEST

akka-persistence/src/main/scala/akka/persistence/SnapshotProtocol.scala
```diff
@@ -24,6 +24,7 @@ final case class SnapshotMetadata(persistenceId: String, sequenceNr: Long, times
  */
 @SerialVersionUID(1L)
 final case class SaveSnapshotSuccess(metadata: SnapshotMetadata)
+  extends SnapshotProtocol.Response // Internal snapshot acknowledgement.
```
ktoso: * These messages are user API, so not internal - remove the comment please (internal snapshot messages are listed in `private[persistence] object SnapshotProtocol {`)
* Was thinking if `Response` is the best type here, but I guess it is, even if the user does not *directly* send a request (he does indirectly, by calling a `saveSnapshot` method).
* What about `SnapshotOffer`, shouldn't we also mark it for completeness? It would be a `Message` then IMO.

Andrei-Pozolotin: well, the original intent was
https://github.com/akka/akka/blob/master/akka-persistence/src/main/scala/akka/persistence/Protocol.scala
```
  /**
   * Helps persistence plugin developers to differentiate
   * internal persistence extension messages from their custom plugin messages.
   */
```

1) re "These messages are user API, so not internal" - they are both public api and internal, look:
https://github.com/akka/akka/blob/master/akka-persistence/src/main/scala/akka/persistence/snapshot/SnapshotStore.scala#L32

2) re:  "Response is the best type here" - it probably better to separate internal from public api completely,
despite seeming message duplication

3) re: "What about SnapshotOffer," - it is used as external only, no internal api function,

please confirm what you want to do with this

ktoso: 1) I'd still remove the comment. If something is internally and user used, users may look for it, we shouldn't confuse them with the internal comment (used in akka when we mean "do not touch!"). The trait is OK ofc.
2) sounds good.
3) right, good then.

Summing up I'd only remove the comments

Andrei-Pozolotin: applied

patriknw:
OK TO TEST

patriknw:
LGTM

patriknw:
thanks

Andrei-Pozolotin:
thank you

Issue #16231, opened 2014-11-06
OneBoundedInterpreter jumpBackTable is broken


drewhk:
Currently it is not recalculated when ops finish, therefore it might happen that finished ops are "jumped over". Currently no op exposes this problem, but needs to be fixed.

drewhk:
Ok, no bug there. The problem would be if you could create a chain like this:

    map -> map -> keepGoing -> map -> map

where keepGoing is an op that calls absorbTermination. Since after a completion round the chain would look like this:

    completed -> completed -> keepGoing -> map -> map

If the jumpback table entries for the remaining maps would point into the completed region it would be a bug. But that can only happen if keepGoing is a PushStage, i.e. it cannot override onPull. But that is incorrect code anyway because calling absorbTermination means that onPull needs to be overridden (currently).

I still "fix" this though, we might change some underlying assumptions, I just cannot reproduce the error with the current code state.

PR #16718, opened 2015-01-26
=str #16231: Protect jumpback table from corruption

drewhk:
Which currently cannot even happen (see explanation https://github.com/akka/akka/issues/16231) but it is good to protect against anyway since the current assumptions might change in the future.

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
@@ -3,6 +3,8 @@
  */
 package akka.stream.impl.fusing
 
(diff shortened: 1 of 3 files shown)


ktoso:
LGTM

drewhk:
Waiting for some more feedback

patriknw:
LGTM

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
```diff
@@ -3,6 +3,8 @@
  */
 package akka.stream.impl.fusing
 
+import akka.stream.impl.fusing.Map
```
rkuhn: unused import?

rkuhn:
LGTM

Issue #16425, opened 2014-11-28
Bind failure doesn't propagate to the stream properly


bantonsson:
`akka-http-core-experimental/testOnly akka.http.ClientServerSpec`

drewhk:
Interesting. There is an explicit call to onError() in that code path, and the path seems to be executed.

drewhk:
Ok, there seems to be a race in the test that is caused by the delayed materialization of the actual listen actor, so in my runs it is the *first* stream that fails and not the second, and the second is the one that succeeds. Seems to be solved by awaiting on the local address of the first stream first.

drewhk:
but unbind has some issues which I don't understand. Also, we failed the unbind future when the original bind failed, I would say this is unexpected. We should only fail the localAddress future.

drewhk:
btw, the test also expected an immediate onError, but first a subscription is handed out and then the onError. This is valid behavior even if it is not optimal.

PR #16720, opened 2015-01-26
!str #16425: Fix bind failure tests, unbind should not fail on bind errors

drewhk:
**DO NOT MERGE YET**
- Changes the test so that it does not involve concurrency between the two binds and it is guaranteed that the _second_ attempt will get the bind failure
- Corrected assumption in the test that onError is immediately signaled, now we check expectErrorOrSubscriptionFollowedByError
- Changed semantics of the unbind Future: It now does _not_ fail if the bind itself has failed. I guess most people don't care whether there was a real unbind, they care about that after the (successful) future completion there are no dangling bound ports.

For some reson on my machine the Http version of the test fails (I cannot bind after unbind) but the Tcp version passes. Let's see if this is Win specific.

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/ClientServerSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/ClientServerSpec.scala
@@ -12,12 +12,11 @@ import scala.concurrent.Await
 import scala.concurrent.duration._
 import org.scalatest.{ BeforeAndAfterAll, Matchers, WordSpec }
 import akka.actor.ActorSystem
-import akka.stream.scaladsl.StreamTcp
+import akka.stream.scaladsl._
 import akka.stream.BindFailedException
 import akka.stream.ActorFlowMaterializer
 import akka.stream.testkit.StreamTestKit
 import akka.stream.testkit.StreamTestKit.{ PublisherProbe, SubscriberProbe }
-import akka.stream.scaladsl._
 import akka.http.engine.client.ClientConnectionSettings
 import akka.http.engine.server.ServerSettings
 import akka.http.model._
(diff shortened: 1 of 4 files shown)


drewhk:
"fixes" #16425

akka-http-core/src/test/scala/akka/http/ClientServerSpec.scala
```diff
@@ -42,27 +41,50 @@ class ClientServerSpec extends WordSpec with Matchers with BeforeAndAfterAll {
...
+      val mm2 = binding.connections.to(Sink(probe2)).run()
+      probe2.expectErrorOrSubscriptionFollowedByError()
 
-      Await.result(binding.unbind(mm1), 1.second)
       val probe3 = StreamTestKit.SubscriberProbe[Http.IncomingConnection]()
       val mm3 = binding.connections.to(Sink(probe3)).run()
-      probe3.expectSubscription() // we bound a second time, which means the previous unbind was successful
+      probe3.expectErrorOrSubscriptionFollowedByError()
+
+      // The unbind should NOT fail even though the bind failed.
```
drewhk: misplaced comment, should be on L70

akka-http-core/src/test/scala/akka/http/ClientServerSpec.scala
```diff
@@ -42,27 +41,50 @@ class ClientServerSpec extends WordSpec with Matchers with BeforeAndAfterAll {
...
+      // Now unbind the first
+      Await.result(binding.unbind(mm1), 1.second)
+      probe1.expectComplete()
+
+      val probe4 = StreamTestKit.SubscriberProbe[Http.IncomingConnection]()
+      val mm4 = binding.connections.to(Sink(probe4)).run()
+      probe4.expectSubscription()
+
+      // Bind succeeded, we have a local address
+      Await.result(binding.localAddress(mm4), 1.second)
```
drewhk: This fails on my machine, but the corresponding line in StreamTcpSpec does not. Not sure it is OS specific timing issue, or there is some underlying bug lurking here.

drewhk:
So it passes on Jenkins but not on my machine. Weeeird.

akka-http-core/src/test/scala/akka/http/ClientServerSpec.scala
patriknw: tried it on mine, works fine

patriknw:
LGTM, I don't know why the re-bind doesn't work on your machine

drewhk:
Well, that would be not surprising. The surprising part is that the Tcp version works, and the Http does not.

drewhk:
Should I just add an ignoreIfWindows() to that test case?

drewhk:
Everyone agrees that we should not fail the unbind Future even if there was no successful bind?

bantonsson:
Yes, I think that unbind of an unsuccessful bind should be ok. If something goes horribly wrong during unbind (if that is even possible), that is a different story.

akka-stream/src/main/scala/akka/stream/impl/io/TcpListenStreamActor.scala
```diff
@@ -88,7 +88,7 @@ private[akka] class TcpListenStreamActor(localAddressPromise: Promise[InetSocket
       case f: CommandFailed ⇒
         val ex = BindFailedException
         localAddressPromise.failure(ex)
-        unbindPromise.failure(ex)
+        unbindPromise.success(() ⇒ Future.successful(()))
```
patriknw: would it be worth to put `Future.successful(())` in a constant?

patriknw:
LGTM, again (I'm fine with `ignoreIfWindows()` if we can't understand why)

akka-stream/src/main/scala/akka/stream/impl/io/TcpListenStreamActor.scala
```diff
@@ -99,7 +99,6 @@ private[akka] class TcpListenStreamActor(localAddressPromise: Promise[InetSocket
         pump()
       case f: CommandFailed ⇒
         val ex = new ConnectionException(s"Command [${f.cmd}] failed")
-        unbindPromise.tryFailure(ex)
```
drewhk: I need to check the reason for the CommandFailed. If it was a result of an Unbind then I actually need to fail here.

akka-stream/src/main/scala/akka/stream/impl/io/TcpListenStreamActor.scala
drewhk: Maybe, but this is not a hot path.

patriknw: not hot, then don't worry

rkuhn:
LGTM so far, your latest comment is not immediately obvious though

drewhk:
That means that I need to double check if we actually fail the unbind promise if we get a CommandFailed with the cause Unbind

rkuhn:
Okay, understood now (somehow I was a bit thick last week ;-) ).

drewhk:
needs to rebase anyway

Issue #16715, opened 2015-01-23
Use case class when exposing undefined-ports when building Flow/Sink from Graph


ktoso:
As discussed in https://github.com/akka/akka/pull/16702#discussion_r23388580

It would be nice to expose the "undefined ports" more nicely when building a Flow or Source or Sink from a graph, example to make the case (current api):

```scala
Flow[Int, String]() { implicit builder ⇒
          import FlowGraphImplicits._

          val in = UndefinedSource[Int]
          val out = UndefinedSink[String]

          in ~> Flow[Int].map(_.toString) ~> out

          (in, out) // this tuple may be confusing
```

Proposals made so far:
* `FlowEndpoints(in, out)`
* `ExposedFlow(in, out)`
* `Ports(fromUpstream = in, toDownstream = out)`As discussed in https://github.com/akka/akka/pull/16702#discussion_r23388580

It would be nice to expose the "undefined ports" more nicely when building a Flow or Source or Sink from a graph, example to make the case (current api):

```scala
Flow[Int, String]() { implicit builder ⇒
          import FlowGraphImplicits._

          val in = UndefinedSource[Int]
          val out = UndefinedSink[String]

          in ~> Flow[Int].map(_.toString) ~> out

          (in, out) // this tuple may be confusing
```

Proposals made so far:
* `FlowEndpoints(in, out)`
* `ExposedFlow(in, out)`
* `Ports(fromUpstream = in, toDownstream = out)`

13h3r:
This proposals only covers situation with one input and one output. What about improving API to expose complex flows with several inputs/outputs?
Right now it looks so verbose both on declaring and using.

ktoso:
This is because this issue is aimed at improving special construction methods to "construct a Source from a graph" and "construct a Flow from a graph" where we can guarantee typesafety between `Source[T]` and the exposed `UndefinedSink[T]`, specifically these 3 APIs:

```scala
val so: Source[Int] = 
  Source() { b ⇒ /* ... */ UndefinedSink[Int] }
      
val si: Sink[Int] = 
  Sink() { b ⇒ /* ... */ UndefinedSource[Int] }
      
val fl: Flow[Int, Int] = 
  Flow() { b ⇒ /* ... */ (UndefinedSource[Int], UndefinedSink[Int]) }
```

[...]

PR #16722, opened 2015-01-26
WIP proposal for a little bit of support for N ports in graphs

ktoso:
WIP, A draft of an idea addressing a perhaps nicer way to deal with "[LEGO style](https://github.com/akka/akka/blob/release-2.3-dev/akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphOpsIntegrationSpec.scala#L37)" of exposing multiple ports.
Triggered by discussion in https://github.com/akka/akka/issues/16715

Does this look helpful enough for exposing multiple inputs / outputs you think?

// cc @13h3r @drewhk

### DIFF
--- a/akka-docs-dev/rst/scala/code/docs/stream/StreamPartialFlowGraphDocSpec.scala
+++ b/akka-docs-dev/rst/scala/code/docs/stream/StreamPartialFlowGraphDocSpec.scala
@@ -90,20 +90,17 @@ class StreamPartialFlowGraphDocSpec extends AkkaSpec {
   "build source from partial flow graph" in {
     //#source-from-partial-flow-graph
     val pairs: Source[(Int, Int)] = Source() { implicit b =>
-      import FlowGraphImplicits._
+      out =>
+        import FlowGraphImplicits._
 
-      // prepare graph elements
-      val undefinedSink = UndefinedSink[(Int, Int)]
-      val zip = Zip[Int, Int]
-      def ints = Source(() => Iterator.from(1))
+        // prepare graph elements
+        val zip = Zip[Int, Int]
+        def ints = Source(() => Iterator.from(1))
 
-      // connect the graph
-      ints ~> Flow[Int].filter(_ % 2 != 0) ~> zip.left
-      ints ~> Flow[Int].filter(_ % 2 == 0) ~> zip.right
-      zip.out ~> undefinedSink
-
-      // expose undefined sink
-      undefinedSink
+        // connect the graph
+        ints ~> Flow[Int].filter(_ % 2 != 0) ~> zip.left
+        ints ~> Flow[Int].filter(_ % 2 == 0) ~> zip.right
+        zip.out ~> out
     }
 
     val firstPair: Future[(Int, Int)] = pairs.runWith(Sink.head)
@@ -137,7 +134,7 @@ class StreamPartialFlowGraphDocSpec extends AkkaSpec {
 
     // format: OFF
     val (_, matSink: Future[(Int, String)]) =
-      //#flow-from-partial-flow-graph
+    //#flow-from-partial-flow-graph
     pairUpWithToString.runWith(Source(List(1)), Sink.head)
     //#flow-from-partial-flow-graph
     // format: ON
--- a/akka-docs-dev/rst/scala/code/docs/stream/StreamPartialFlowGraphDocSpec.scala
+++ b/akka-docs-dev/rst/scala/code/docs/stream/StreamPartialFlowGraphDocSpec.scala
@@ -15,6 +15,8 @@ import akka.stream.scaladsl.UndefinedSink
 import akka.stream.scaladsl.UndefinedSource
 import akka.stream.scaladsl.Zip
 import akka.stream.scaladsl.ZipWith
+import akka.stream.scaladsl.Merge
+import akka.stream.scaladsl.Ports
 import akka.stream.testkit.AkkaSpec
 
 import scala.concurrent.Await
@@ -141,5 +143,63 @@ class StreamPartialFlowGraphDocSpec extends AkkaSpec {
     // format: ON
 
     Await.result(matSink, 300.millis) should equal(1 -> "1")
+  }
+
+  "help users in providing complex ports" in {
+    object AudioPorts extends Ports {
+      val nums = Input[Int] // TODO: use I / O words or keep sinks/sources?
+      val words = Input[String]
+
+      val out = Output[String]
+    }
+
+    // building
+    val g = PartialFlowGraph(AudioPorts) { implicit b ⇒
+      import FlowGraphImplicits._
+      val merge = Merge[String]
+      // format: OFF
+      AudioPorts.nums ~> Flow[Int].map(_.toString) ~> merge
+                                  AudioPorts.words ~> merge ~> AudioPorts.out
+      // format: ON
+    }
+
+    // connecting, current style:
+    FlowGraph(g) { b ⇒
+      b.attachSource(AudioPorts.nums, Source(1 to 10))
+      b.attachSource(AudioPorts.words, Source((1 to 10).map(_.toString)))
+      b.attachSink(AudioPorts.out, Sink.ignore)
+    }.run()
+
+    // connecting, proposed style for Ports:
+    FlowGraph(g) { implicit b ⇒
+      import AudioPorts._
+
+      AudioPorts.nums(Source(1 to 10))( /* implicit */ b)
+      AudioPorts.words(Source((1 to 10).map(_.toString)))
+      AudioPorts.out(Sink.ignore)
+    }.run()
+  }
+
+  "throw when not all Ports ports are connected" in {
+    case object AudioPorts extends Ports {
+      val nums = Input[Int]
+      val words = Input[String]
+
+      val out = Output[String]
+    }
+
+    val ex = intercept[IllegalArgumentException] {
+      PartialFlowGraph(AudioPorts) { implicit b ⇒
+        import FlowGraphImplicits._
+        val merge = Merge[String]
+        // format: OFF
+        AudioPorts.nums ~> Flow[Int].map(_.toString) ~> merge
+                                    AudioPorts.words ~> merge ~> Sink.ignore
+        // format: ON
+      }
+    }
+
+    ex.getMessage should include("must contain all ports")
+
   }
 }
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/FlowGraph.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/FlowGraph.scala
@@ -10,6 +10,7 @@ import akka.stream.impl.Ast
 import akka.stream.impl.Ast.FanInAstNode
 import akka.stream.impl.{ DirectedGraphBuilder, Edge }
 import akka.stream.impl.Ast.Defaults._
+import akka.stream.scaladsl.FlowGraphInternal.InternalVertex
 import akka.stream.scaladsl.OperationAttributes._
 import org.reactivestreams._
 
@@ -1300,6 +1301,16 @@ object PartialFlowGraph {
   def apply(partialFlowGraph: PartialFlowGraph)(block: FlowGraphBuilder ⇒ Unit): PartialFlowGraph =
     apply(new FlowGraphBuilder(partialFlowGraph))(block)
 
+  def apply(p: Ports, requireAllPorts: Boolean = true)(block: FlowGraphBuilder ⇒ Unit): PartialFlowGraph = {
+    val partial = apply(block)
+    
+    // TODO: validation could point out which ones are not connected here
+    // TODO: We could also have "OptionalInput", have not thought that one through yet though
+    if (requireAllPorts)
+      require(p.allPorts.forall(partial.graph.contains(_)), s"PartialFlowGraph using $p must contain all ports contained within this Ports object")
+    partial
+  }
+
   private def apply(builder: FlowGraphBuilder)(block: FlowGraphBuilder ⇒ Unit): PartialFlowGraph = {
     // FlowGraphBuilder does a full import on the passed graph, so no defensive copy needed
     block(builder)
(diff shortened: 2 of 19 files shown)


patriknw:
I'm lacking context of why this is needed. Feels a bit magical, while it's not really doing much (that the user can't do herself)

ktoso:
My initial feeling was that it's up to the user to define all ports and pass them around - we'd just support Source/Flow/Sink (1-2 undefined ports) as some case class. In #16715 @drewhk  and @13h3r voiced some interest in "some help" for n-ports, so this PR is a request-for-comments if this would be helpful or too magical.

akka-docs-dev/rst/scala/code/docs/stream/StreamPartialFlowGraphDocSpec.scala
```diff
@@ -142,4 +144,62 @@ class StreamPartialFlowGraphDocSpec extends AkkaSpec {
...
+  "help users in providing complex ports" in {
+    object AudioPorts extends Ports {
+      val nums = Input[Int] // TODO: use I / O words or keep sinks/sources?
+      val words = Input[String]
+
+      val out = Output[String]
+    }
+
+    // building
+    val g = PartialFlowGraph(AudioPorts) { implicit b ⇒
```
drewhk: This is beautiful! Will see how the underlying implementation turns out, but this style will make generalized graph building blocks much more easy to create.

akka-docs-dev/rst/scala/code/docs/stream/StreamPartialFlowGraphDocSpec.scala
```diff
@@ -142,4 +144,62 @@ class StreamPartialFlowGraphDocSpec extends AkkaSpec {
 
     Await.result(matSink, 300.millis) should equal(1 -> "1")
   }
+
+  "help users in providing complex ports" in {
+    object AudioPorts extends Ports {
+      val nums = Input[Int] // TODO: use I / O words or keep sinks/sources?
```
ktoso: (probably should be same as in fleximerge if we want this style of thing - so `createInputPort`?)

akka-docs-dev/rst/scala/code/docs/stream/StreamPartialFlowGraphDocSpec.scala
drewhk: Would be nice to get back a similar class as the object, but one that represents the built graph. Just thinking aloud.

akka-docs-dev/rst/scala/code/docs/stream/StreamPartialFlowGraphDocSpec.scala
```diff
@@ -142,4 +144,62 @@ class StreamPartialFlowGraphDocSpec extends AkkaSpec {
...
+      b.attachSource(AudioPorts.nums, Source(1 to 10))
+      b.attachSource(AudioPorts.words, Source((1 to 10).map(_.toString)))
+      b.attachSink(AudioPorts.out, Sink.ignore)
+    }.run()
+
+    // connecting, proposed style for Ports:
+    FlowGraph(g) { implicit b ⇒
+      import AudioPorts._
+
+      AudioPorts.nums(Source(1 to 10))( /* implicit */ b)
```
drewhk: I would remove this version and keep only the one that takes the Ports class as a parameter. That way the builder can take it internally as an argument, too and register them itself. Also, we can now probably unify PartialGraph and Graph, since Graph is something that takes a NoPorts as arguments and gives a RunnableGraph, anything else that takes a non-empty Ports can return a PartialGraph that gives access to those ports (API needs to be designed for that though).

akka-stream/src/main/scala/akka/stream/scaladsl/FlowGraph.scala
```diff
@@ -1487,3 +1498,40 @@ object FlowGraphImplicits {
...
   }
 }
+
+trait Ports {
+  private var inputs: List[UndefinedSource[_]] = Nil
+  private var outputs: List[UndefinedSink[_]] = Nil
+
+  // TODO draft - could simply be also defined on Undefined elements
+  implicit class DefinableSink[T](undefined: UndefinedSink[T]) {
+    def apply(s: Sink[T])(implicit builder: FlowGraphBuilder) = builder.attachSink(undefined, s) // TODO `apply` or `attach`?
```
drewhk: If you take the Ports only as an argument to (Partial)FlowGraph then you don't need this, you can just populate the builder from the factory method. See my previous comment.

13h3r:
1. I was trying to build end-user API for this feature for a several times and all I get is some kind of case class (because we need names for every in and out). So at this point I have no idea how to represent multiple inputs/outputs better then case class/object.


2. What if we introduce dsl to bind inputs and outpus like 

```scala
val source: Source[String] = ???
val in: UnboundedSource[String] = ???
FlowGraph(g) { b ⇒
  source ~> in
  out ~> sink
}
```

Existing verbose API with `attachSource` a bit confusing after all nice examples with `~>`.

3. Last thought. It is not clear from the documentation, that it is possible to define `UndefinedSink` and `UndefinedSource` and use them multiple times during `Flow` creation from `PartialFlow`. I think we need just an advice how to structure `PartialFlow`s, how materialize them and so on. Like keeping actor's messages in companion object.

akka-docs-dev/rst/scala/code/docs/stream/StreamPartialFlowGraphDocSpec.scala
```diff
@@ -142,4 +144,62 @@ class StreamPartialFlowGraphDocSpec extends AkkaSpec {
...
+      import FlowGraphImplicits._
+      val merge = Merge[String]
+      // format: OFF
+      AudioPorts.nums ~> Flow[Int].map(_.toString) ~> merge
+                                  AudioPorts.words ~> merge ~> AudioPorts.out
+      // format: ON
+    }
+
+    // connecting, current style:
+    FlowGraph(g) { b ⇒
```
13h3r: It is possible to have `AudioPorts` captured in `g` and have some instance class with all actual `Source`s and `Sink`s captured and bind not every in and out, but `g` and actual instance of `AudioPorts`. 

Right now I could not provide API, but just an idea how to improve type safety and force to have all in/outs get bound.

akka-docs-dev/rst/scala/code/docs/stream/StreamPartialFlowGraphDocSpec.scala
ktoso: > I would remove this version and keep only the one that takes the Ports class as a parameter.
I didn't quite get that, could you give an example?

drewhk: Ahh I see this is the connecting part.

ktoso: Yeap. I like the idea @13h3r mentioned below, to allow `Source ~> UndefinedSource`, I'll try that instead of the `nums.apply()` - that doesnt look like any of our apis

ktoso: Or is it too confusing:

```
FlowGraph(g) { b ⇒
      import FlowGraphImplicits._
      // nums: UndefinedSource[Int]
      Source(1 to 10) ~> AudioPorts.nums // "attach this source to this UndefinedSource"

     // may be confusing because
     Source(1 to 10) ~> UndefinedSink[Int] // normal connection, not an "attach"
```

?

akka-docs-dev/rst/scala/code/docs/stream/StreamPartialFlowGraphDocSpec.scala
ktoso: Interesting idea / spin on the above (a bit akka-http routing inspired):

```scala
    val g = PartialFlowGraph(AudioPorts) { implicit b ⇒ ports ⇒
        import FlowGraphImplicits._
        val merge = Merge[String]
      // format: OFF
      ports.nums ~> Flow[Int].map(_.toString) ~> merge
                                  ports.words ~> merge ~> ports.out
      // format: ON
    }
```

yay or nay? Might be "too fancy"

akka-docs-dev/rst/scala/code/docs/stream/StreamPartialFlowGraphDocSpec.scala
rkuhn: Here’s the thing: a port is always both a Source and a Sink. Maybe we should just unify them in this fashion and then use only the `~>` syntax—the types will make sure that we know what is meant.

ktoso: Cool, I'll gladly apply this then :-)

rkuhn: We’ll have to make it fail at runtime, then, as soon as such a port has more than one connected input or output. Could a Merge or a Broadcast also be a Port, I wonder?

ktoso: It currently already fails at runtime if you'd try to attach to the same undefined element twice, so that's covered (`attach*` does this - error message could be perhaps improved though currently it would fail as "[info]   java.lang.IllegalArgumentException: No matching UndefinedSource [akka.stream.scaladsl.UndefinedSource@36fd5d59]" as it is already a "defined" one).

ktoso:
I have pushed some updates. Most notably:

1) Source and Sink are now *given* the in/out ports:

```scala
import FlowGraphImplicits._

Source[Int]() { implicit b ⇒ out ⇒      // out: UndefinedSink[Int]
  Source.single(1) ~> out
}

Sink[Int]() { implicit b => in =>    // in: UndefinedSource[Int]
  in ~> Sink.ignore 
}
```

For Java I made it `Procedure2[FlowGraphBuilder, Undefined*[T]]`.

2) We can now "attachSource" and "attachSink" by using `~>`, so when preparing graph:

```
case object AudioPorts extends Ports {
  val nums = InputPort[Int]
}

// building
AudioPorts.nums ~> Sink.ignore

// defining the UndefinedSource:
Source.single(1) ~> AudioPorts.nums
```

3) Notice how ports are passed in to a flow graph:

```scala
PartialFlowGraph(AudioPorts) { implicit b ⇒ ports ⇒
```

so it looks similar to constructing sink / source using the proposal.

13h3r:
1) I like how `Source` and `Sink` looks now, but how `Flow` will looks like? I this we should keep same idea and provide something like

```scala
Flow[Int]() { implicit b ⇒ in ⇒ out ⇒ 
  in ~> Flow[Int].map(_ + 1) ~> out
}
```

2) and 3) Looks good

ktoso:
For Flow I was thinking already bring in `Ports`, but a "predefined FlowPorts" (upside being that again "looks similar" to the N case, and similar to the 1 case"), so it would be:

```scala
Flow[Int, Int]() { implicit b => ports =>
  ports.in ~> Flow[Int].map(_ + 1) ~> ports.out
}
```

Another upside is that in many graph elements we already provide `in` or `out` already (`FlexiMerge` has a default `out` output port), so perhaps we could get to making all these the same concept.

Thanks for your comments! Awaiting more feedback before I continue working on this one :)

13h3r:
Any chances to have this in 1.0?

ktoso:
Hi there,
status update here is that we're about to finish an interals (graph representation) rewrite and then this PR will be rebased on top of it. The new APIs (this PR) will also include better support for exposing graph backed `Source` / `Sink` / `Flow` as well as the "ports" proposal fo Graphs as discussed in this PR :-)

I hope to get it in this week - so, to answer your question - yes, more-typedness is going to come to 1.0 (even to M4 I hope).

ktoso:
This work is now included and superseeded by the internal graph representation rewrite, WIP here https://github.com/akka/akka/pull/16879

13h3r:
@ktoso I found time to take a look at M4. Looks like API become even more verbose: 

```scala
  val flow = Flow() { implicit b =>
    import FlowGraph.Implicits._
    val bcast = b.add(Broadcast[Int](2))
    val merge = b.add(Merge[String](2))

    bcast ~> Flow[Int].filter(_ % 2 == 0) ~> merge
    bcast ~> Flow[Int].filter(_ % 3 == 0) ~> merge

    bcast.in -> merge.out
  }
```

Things I don't like:
- explicit `b.add` 
- explicit numbers of inputs/outputs
- absence of implicit input and output

Looks like we discussed what we want to achieve and implement something completely different. I understand that there are some other internal refactorings in M4 (with Shapes and so on), but looks like we make API worse.

Probably I miss something in new API. Please point to correct way in this case. 

Thanks!

rkuhn:
Part of the explanation is in the announcement email: we found fundamental problems in the previous approach that required a redesign. The new API is consistently based upon Graphs and Shapes, which explains why the explicit import is needed: a Broadcast is just a `Graph[UniformFanOutShape[T, T], Unit]`, and it can be used in any place that needs a graph of that shape. A shape is a concrete form, hence the number of its ports need to be known—otherwise we couldn’t allow you to return one of the `bcast` outputs as the overall flow output. Concerning “absence of implicit input and output” I don’t know what you mean, but the new Flow DSL is consistent with the FlowGraph DSL and all others in that the Shape (or the ports, where simpler) need to be returned. This allows the types to be inferred quite nicely.

One thing we improved is that you can now write
~~~scala
bcast.filter(_ % 2 == 0) ~> merge
~~~

13h3r:
Roland, thanks for reply. To be clear - I understand changes in M4 and it looks consistent and reasonable. In this ticket I try to find way to improve user API. I agree with explicit number of inputs/outputs. What about getting rid of `b.add` somehow?

rkuhn:
Importing a subgraph into a larger graph needs to be explicit because not only is it a type change (from `Graph[Shape, _]` to `Shape`) but also it is a side-effect (it modifies the builder). I think it would be too confusing to implicitly do the addition, and that would also make it impossible to construct a Zip outside of a graph (for reuse in different locations). One idea that just crossed my mind is that we could add factory methods to the `FlowGraph.Builder`, but I’m not sure how much that would help:
~~~scala
val bcast = b.createBroadcast[Int](2)
~~~

Issue #16553, opened 2014-12-16
TcpStream doesn't complete incoming stream


jrudolph:
This is related to #16552: if `ConfirmedClosed` is never generated the input stream of the Tcp connection will never be completed.

However, even with a fix for #16552 it still doesn't work because the TcpStreamActor shuts itself down before the completion is signaled to both, the input and the output stream, i.e. it doesn't properly coordinate the closing of incoming and outgoing streams.

Alien2150:
Hi, i am not sure if that is related too. But when I use the SSLCipherActor I don't get any completion signals but what I do see is an log-entry: 

[INFO] [01/07/2015 14:12:53.549] [app-akka.actor.default-dispatcher-15] [akka://
app/system/IO-TCP/selectors/$a/1] Message [akka.io.Tcp$Abort$] from Actor[akka:/
/app/system/IO-TCP-STREAM/server-1-%2F0.0.0.0%3A16500/$a#-536894325] to Actor[akka://app/system/IO-TCP/selectors/$a/1#-1566861189] was not delivered. [1] dead l
etters encountered. This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' and 'akka.log-dead-letters-during-shutdown'.

This is the code i am using: https://www.google.com/url?q=https%3A%2F%2Fgist.github.com%2FAlien2150%2F9468c871135fd94869a2&sa=D&sntz=1&usg=AFQjCNEanvTVgDgxtLdTfdy88exEUkbfJQ

Regards,
Thomas

drewhk:
Ok, I am testing with a special probe for now that will ping the selector with a ResumeReading before fishing for a connection event. That will help to fix the stream issues, and later by removing the "ping" code we can test the underlying Akka IO fix.

drewhk:
I will need the fix for https://github.com/akka/akka/issues/16723

Issue #16723, opened 2015-01-26
TcpConnection does not allow Close only Abort after sending a ConfirmedClose


drewhk:
Once we stopped writing with ConfirmedClose, we only parse the Abort command in the corresponding state, which means that once we got uninterested maintaining the read side we must terminate with an RST instead of a FIN.

drewhk:
Well, after seding a FIN packed there is only RST to send, so this is very likely invalid.

drewhk:
Marking this as invalid.

PR #16724, opened 2015-01-26
=str #16553: Fix TCP stream shutdown scenarios

drewhk:
**DO NOT MERGE YET**

Reworked the tests for TCP Streams to better test closing scenarios. Also refactored the connection actor completion handling and fixed some of the issues. Related ticket that is a fallout: https://github.com/akka/akka/issues/16723

Until that ticket is fixed the test scenario marked as pending cannot be made to pass.

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/io/StreamTcpSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/io/StreamTcpSpec.scala
@@ -3,6 +3,8 @@
  */
 package akka.stream.io
 
+import akka.io.Tcp._
+
 import scala.concurrent.Await
 import scala.concurrent.duration._
 import akka.util.ByteString
@@ -68,7 +70,7 @@ class StreamTcpSpec extends AkkaSpec with TcpHelper {
 
     }
 
-    "half close the connection when output stream is closed" in {
+    "work when client closes write, then remote closes write" in {
       val testData = ByteString(1, 2, 3, 4, 5)
       val server = new Server()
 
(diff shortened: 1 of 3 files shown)


drewhk:
Almost fixes https://github.com/akka/akka/issues/16553

akka-stream-tests/src/test/scala/akka/stream/io/StreamTcpSpec.scala
```diff
@@ -77,36 +79,142 @@ class StreamTcpSpec extends AkkaSpec with TcpHelper {
...
+
+      // Client can still write
+      tcpWriteProbe.write(testData)
       serverConnection.read(5)
+      serverConnection.waitRead() should be(testData)
+
+      // Close client side write
+      tcpWriteProbe.close()
+      // FIXME: Why not Closed?
+      serverConnection.expectClosed(PeerClosed)
```
drewhk: I think this is a bug in IO, after a PeerClosed a full close should result in Closed, not another PeerClosed

akka-stream/src/main/scala/akka/stream/impl/io/TcpConnectionStream.scala
```diff
@@ -64,21 +63,8 @@ private[akka] abstract class TcpStreamActor(val settings: MaterializerSettings)
     def handleRead: Receive = {
       case Received(data) ⇒
         pendingElement = data
```
drewhk: no need to store the data if closed, this should be moved into the else statement below

drewhk:
Note to self: Add check that the test connection actor is terminated.

akka-stream-tests/src/test/scala/akka/stream/io/StreamTcpSpec.scala
drewhk: Well, rather ConfirmedClosed

drewhk: No, it seems to be a bug of my own :)

drewhk:
Another round is coming to this, closing it until then.

PR #16725, opened 2015-01-26
Improve error logging in akka.io.SelectionHandler

reid-spencer:
When a java.lang.reflect.InvocationTargetException occurs, it produces a logging
message that is "null" because it's cause returns no message. This notably occurs
when a java.net.BindException occurs during construction of a TCP selector. This
patch causes a log message like this:
   java.net.BindException: Can't assign requested address
instead
   null
which is quite a bit more helpful when diagnosing connection issues.


typesafehub-validator:
Hi @reid-spencer,

We see that you have signed the Typesafe Contributors License Agreement before, however, the CLA has changed since you last signed it.
Please review the new CLA and sign it before we proceed with reviewing this pull request:

reid-spencer:
I have accepted the (new) CLA.

ktoso:
OK TO TEST

akka-actor/src/main/scala/akka/io/SelectionHandler.scala
```diff
@@ -257,7 +257,12 @@ private[io] class SelectionHandler(settings: SelectionHandlerSettings) extends A
                               decision: SupervisorStrategy.Directive): Unit =
         try {
           val logMessage = cause match {
-            case e: ActorInitializationException if e.getCause ne null ⇒ e.getCause.getMessage
+            case e: ActorInitializationException if (e.getCause ne null) && (e.getCause.getMessage ne null) ⇒ e.getCause.getMessage
+            case e: ActorInitializationException if (e.getCause ne null) ⇒
+              e.getCause match {
```
drewhk: Shouldn't we get and log the exception name anyway? Some exceptions will have meaningless messages without the name of the exception itself (sadly).

reid-spencer: Perhaps.  In my projects I tend to have a generic facility for turning exceptions into meaningful messages because, unfortunately, the "standard" for this is shoddy.  It would be awesome if Akka had such a facility that covered the JVM and Akka exceptions because useful log messages are one of the strongest aids for debugging complex asynch programs that do not lend themselves well to using debuggers. I'd be willing to help make this improvement to Akka but it is a bit of a wide-sweeping change. This patch just aimed at taking care of one specific instance where it is important but if the Akka developers want this generalized, I'd be willing to help.

patriknw:
LGTM, thanks

akka-actor/src/main/scala/akka/io/SelectionHandler.scala
```diff
@@ -257,7 +257,12 @@ private[io] class SelectionHandler(settings: SelectionHandlerSettings) extends A
...
+                case ie: java.lang.reflect.InvocationTargetException ⇒ ie.getTargetException.toString
+                case t: Throwable                                    ⇒ t.getClass.getSimpleName
```
rkuhn: I’m not sure whether we have problematic exception types, but in general `getSimpleName` is not reliable when faced with Scala classes. Please use either `.getName` or `Logging.simpleName(t)`.

akka-actor/src/main/scala/akka/io/SelectionHandler.scala
rkuhn: You raise a good point and you seem to have experience with this topic, so help is definitely welcome. It would be best to first open tickets for discussing shortcomings and goals before diving into wide-sweeping change mode. It would be good to also include test cases that verify the improved logging so we don’t accidentally break it again later—would you mind adding one to this PR as well? Thanks!

reid-spencer: @rkuhn - Yes, I can do that.

akka-actor/src/main/scala/akka/io/SelectionHandler.scala
reid-spencer: I wasn't aware of Logging.simpleName .. I'll use that.

akka-actor/src/main/scala/akka/io/SelectionHandler.scala
rkuhn: Thanks!

reid-spencer:
I'm going to have to renege on this due to #16764.

rkuhn:
Sorry for the long pause: this looks good, could you please squash and rebase so we can merge? Thanks.

reid-spencer:
@rkuhn - I will, but under a separate pull request. I botched the first attempt.

rkuhn:
While you’re at it please also create a ticket and format the commit message accordingly:
~~~
=act #<ticket> <description>
~~~
Thanks!

reid-spencer:
Please pull #16789, not this one.

Issue #16552, opened 2014-12-16
TcpConnection: ConfirmedClose doesn't work with `pullMode = true`


jrudolph:
That's because, no read interest is registered and therefore the closing of the connection is missed.

/cc @sirthias

drewhk:
While fixing a related ticket https://github.com/akka/akka/issues/16553 I added a special method in tests that will ping the test connection actor with a ResumeReading before expecting a close message. This comment here is just a remainder that once this ticket is fixed the above mentioned workaround in akka.stream.io.TcpHelper should be removed.

drewhk:
Superceded by ticket: https://github.com/akka/akka/issues/17122

PR #16728, opened 2015-01-27
=str #16553: Fix TCP stream shutdown scenarios

### LINKED ISSUE #16653: HTTP Form upload hangs if the server has been port scanned with nc (GNU netcat 0.7.1)

drewhk:
Properly handle various stream and TCP shutdown scenarios
- Upgraded tests to cover various closing scenarios
- Even upon downstream cancel we pull the events from TCP by sending ResumeReading and draining ignored bytes if needed
- Now also tests if the test connection actor terminates
- workaround for #16552, before we wait on close events in the tests, we send a ResumeReading first
- tests even cancellation after PeerClosed by trying to write (otherwise no ErrorClosed is generated)

Fixes #16653

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/io/StreamTcpSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/io/StreamTcpSpec.scala
@@ -3,6 +3,8 @@
  */
 package akka.stream.io
 
+import akka.io.Tcp._
+
 import akka.stream.BindFailedException
 
 import scala.concurrent.Await
@@ -70,7 +72,7 @@ class StreamTcpSpec extends AkkaSpec with TcpHelper {
 
     }
 
-    "half close the connection when output stream is closed" in {
+    "work when client closes write, then remote closes write" in {
       val testData = ByteString(1, 2, 3, 4, 5)
       val server = new Server()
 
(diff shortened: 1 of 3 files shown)


akka-stream/src/main/scala/akka/stream/impl/io/TcpConnectionStream.scala
```diff
@@ -86,8 +80,21 @@ private[akka] abstract class TcpStreamActor(val settings: MaterializerSettings)
...
     override def cancel(): Unit = {
-      closed = true
-      pendingElement = null
+      if (!closed) {
+        closed = true
+        pendingElement = null
+        if (tcpOutputs.isClosed && (connection ne null))
+          connection ! Abort
+        else
+          connection ! ResumeReading
```
patriknw: can `connection` be null here?

drewhk: I guess so. The null guard ended up in the wrong place.

patriknw:
LGTM

akka-stream-tests/src/test/scala/akka/stream/io/StreamTcpSpec.scala
```diff
@@ -77,36 +79,177 @@ class StreamTcpSpec extends AkkaSpec with TcpHelper {
...
+      serverConnection.read(5)
+      serverConnection.waitRead() should be(testData)
+
+      // Close client side write
+      tcpWriteProbe.close()
+
+      // Need a write on the server side to detect the close event
       serverConnection.write(testData)
-      tcpReadProbe.subscriberProbe.expectNoMsg(1.second)
+      serverConnection.write(testData)
```
rkuhn: We should document this: if the client cancels reading before completing the writing then a passive server will just hang.

akka-stream-tests/src/test/scala/akka/stream/io/StreamTcpSpec.scala
```diff
@@ -77,36 +79,177 @@ class StreamTcpSpec extends AkkaSpec with TcpHelper {
...
+      val serverConnection = server.waitAccept()
+
+      // Client can still write
+      tcpWriteProbe.write(testData)
       serverConnection.read(5)
+      serverConnection.waitRead() should be(testData)
+
+      // Close client side write
+      tcpWriteProbe.close()
+      serverConnection.expectClosed(PeerClosed)
```
rkuhn: this case is better because the server at least gets some notification

akka-stream-tests/src/test/scala/akka/stream/io/StreamTcpSpec.scala
drewhk: Yep, unfortunately I found no way around it. An idle timeout stage will help of course.

rkuhn:
LGTM, great fixes!

Issue #16327, opened 2014-11-18
IllegalStateException in akka.dispatch.BatchingExecutor$Batch.run


maxcom:
Our application provides REST interface to database storage. User uploads data via HTTP interface, we store it to JDBC and then do some additional logic that processes uploaded data. HTTP interface is implemented via Spray 1.3.2; Akka version 2.3.6 (but I have seen this problem in previous 2.3.x). We do not use akka-streams.

We create and then stop an some number of actors per upload; also we use transformation operations on Future in context dispatcher.

Sometimes we see this exception in our logs:

```
[incident-daemon-system-akka.actor.default-dispatcher-233] ERROR akka.dispatch.Dispatcher - exception in sameThreadExecutionContext (a.d.Dispatcher)
java.lang.IllegalStateException: exception in sameThreadExecutionContext
        at akka.dispatch.ExecutionContexts$sameThreadExecutionContext$.reportFailure(Future.scala:76)
        at scala.concurrent.impl.CallbackRunnable.executeWithValue(Promise.scala:40)
        at scala.concurrent.impl.Promise$DefaultPromise.tryComplete(Promise.scala:248)
        at scala.concurrent.Promise$class.complete(Promise.scala:55)
        at scala.concurrent.impl.Promise$DefaultPromise.complete(Promise.scala:153)
        at scala.concurrent.Future$$anonfun$map$1.apply(Future.scala:235)
        at scala.concurrent.Future$$anonfun$map$1.apply(Future.scala:235)
        at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:32)
        at akka.dispatch.BatchingExecutor$Batch$$anonfun$run$1.processBatch$1(BatchingExecutor.scala:67)
        at akka.dispatch.BatchingExecutor$Batch$$anonfun$run$1.apply$mcV$sp(BatchingExecutor.scala:82)
        at akka.dispatch.BatchingExecutor$Batch$$anonfun$run$1.apply(BatchingExecutor.scala:59)
        at akka.dispatch.BatchingExecutor$Batch$$anonfun$run$1.apply(BatchingExecutor.scala:59)
        at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:72)
        at akka.dispatch.BatchingExecutor$Batch.run(BatchingExecutor.scala:58)
        at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:41)
[...]

PR #16744, opened 2015-01-28
Added test to reproduce the stacktrace

skisel:
I was able to reproduce the stacktrace mentioned in the issue https://github.com/akka/akka/issues/16327


rkuhn:
Thanks for the test case, this is quite possibly fixed on master already, can you verify by targeting that branch instead of release-2.3-dev?

PR #16750, opened 2015-01-29
+str #16547 "lazyEmpty" source, allowing external completion

### LINKED ISSUE #16547: Never completed, no element Source
As per this forum topic: https://groups.google.com/forum/#!topic/akka-user/E46rSCLpZYI, sometimes a source which never produces an elements (but also doesn't complete) is useful e.g. when using TCP streams, where only one side transmits, and the reverse stream is only used to detect broken connections.

ktoso:
Implemented this guy during a meetup yesterday, short and nice :-)

The TCK verification for this one is pending, because RC1 TCK does not allow empty publishers. But that's solved on reactive-streams master. I'll create a ticket to bump the dependency and remember to add such (we also have empty) publishers to TCK verification (its only one or two tests for those).

resolves #16547

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
@@ -178,6 +178,17 @@ object Source {
   private[this] val _empty: Source[Nothing] = apply(EmptyPublisher)
 
   /**
+   * Create a `Source` with no elements, which does not complete its downstream,
+   * until externally triggered to do so.
+   *
+   * It materializes a [[scala.concurrent.Promise]] which will be completed
+   * when the downstream stage of this source cancels. This promise can also
+   * be used to externally trigger completion, which the source then signalls
+   * to its downstream.
+   */
+  def lazyEmpty[T]() = LazyEmptySource[T]()
+
+  /**
    * Create a `Source` that immediately ends the stream with the `cause` error to every connected `Sink`.
    */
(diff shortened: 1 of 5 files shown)


akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -178,6 +178,16 @@ object Source {
...
 
   /**
+   * A `Source` with no elements and does not automatically `onComplete`.
+   *
+   * It materializes a [[scala.concurrent.Promise]] which will be completed
+   * when the downstream stage of this source cancels. This promise can also
+   * be used to externally trigger completion, which the source then signalls
+   * to its downstream.
+   */
+  def never[T]() = NeverSource[T]()
```
drewhk: I am not sure about the name. KeepAliveEmtpySource is long, maybe someone has a better idea?

ktoso: never works for me but I'm open for suggestions. (datapoint: In rx it's also never https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/never.md)

rkuhn:
LGTM, although “never” is not what it does. lazyEmpty?

ktoso:
Do we want a "really never" which would then be `never` in addition to this one or just rename this to `lazyEmpty`?

rkuhn:
No, I think this one is “never” enough. Just don’t complete the Promise …

ktoso:
Yeap.

ktoso:
Pushed the rename.

PR #16752, opened 2015-01-29
Test that reproduces the issue #16327

skisel:
Added test to master branch according to request by Roland Kuhn:
https://github.com/akka/akka/pull/16744#issuecomment-72010706

It still fails (as well as in 2.3 branch), so the root cause was because sameThreadExecutionContext cannot avoid thread starvation when Batching is used and Await/blocking is called.

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/dispatch/ExecutionContextSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/dispatch/ExecutionContextSpec.scala
@@ -58,6 +58,29 @@ class ExecutionContextSpec extends AkkaSpec with DefaultTimeout {
       Await.result(p.future, timeout.duration) should be(())
     }
 
+    "be able to avoid starvation in sameThreadExecutionContext when Batching is used and Await/blocking is called" in {
+      implicit val ec = ExecutionContexts.sameThreadExecutionContext
+
+      def batchable[T](f: ⇒ T)(implicit ec: ExecutionContext): Unit = ec.execute(new Batchable {
+        override def isBatchable = true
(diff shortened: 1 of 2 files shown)


rkuhn:
OK TO TEST

rkuhn:
Thanks for this test case, now we only need a matching patch :-) (but having a reproducer is always a very good start)

refs #16327

skisel:
I removed BatchingExecutor mixing from sameThreadExecutionContext because it may cause starvation problems, moreover it seems it does not add any value to the executor. What do you think?

rkuhn:
Further investigation reveals the following:

* the test failure on master is different from the other tickets, it simply times out (as is expected) due to the deferred evaluation of batchable subtasks—this is perfectly fine and normal (*)
* the fix for the actual issue was done in #1970, but at the time we did not see that this might impact Spray for Akka 2.3.x

The way forward is therefore to port Viktor’s PR to the release-2.3 branch and include that with Akka 2.3.10.

(*) sameThreadExecutor has two choices: either run depth first or breadth first. No matter which we choose we can write a test that fails due to deadlock. Giving the user a choice (via `Batchable.isBatchable`) allows them to configure their tasks appropriately, so we’ll leave things as they are on this front.

rkuhn:
@skisel Do you want to try your hand at this, or shall I?

skisel:
Ok, its clear. Yes, I will try my hand at this. Thanks

rkuhn:
This PR can be closed now that the underlying issue has been fixed otherwise. Thanks for pushing this!

Issue #16448, opened 2014-12-03
ensure consistent naming of Failures in Streams API


rkuhn:
The Reactive Streams’ `onError` method is used for signaling errors on the stream infrastructure level, which means that these are [failures](http://www.reactivemanifesto.org/glossary#Failure) on the Akka Streams level of abstraction. All our facilities for handling or emitting these signals should consistently be named “failure” in the API that we expose to Akka Streams users.

PR #16756, opened 2015-01-30
!str #16448 Change error to failure

patriknw:
- one API change OverflowStrategy.error -> OverflowStrategy.fail
- error is still kept in the internals where we are at the reactive
  streams level


patriknw:
Refs #16448

drewhk:
LGTM

PR #16763, opened 2015-01-30
fix all non-deprecation warnings

rkuhn:
… well, except for one place:
- trait ActorRefProvider cannot be fixed without breaking binary compatibility, which is probably not worth it

I suppose we could turn warnings into errors everywhere but in `akka-actor`.


akka-actor-tests/src/test/scala/akka/serialization/SerializeSpec.scala
```diff
@@ -20,7 +20,7 @@ import org.apache.commons.codec.binary.Hex.encodeHex
 
 object SerializationTests {
 
-  val serializeConf = """
+  val serializeConf = s"""
```
patriknw: so there is a warning when using `$` in `"""`, or what is the reason?

akka-actor/src/main/scala/akka/actor/ActorRef.scala
```diff
@@ -326,7 +326,7 @@ private[akka] class LocalActorRef private[akka] (
    * If this method returns true, it will never return false again, but if it
    * returns false, you cannot be sure if it's alive still (race condition)
    */
-  @deprecated("Use context.watch(actor) and receive Terminated(actor)", "2.2") override def isTerminated: Boolean = actorCell.isTerminated
+  override def isTerminated: Boolean = actorCell.isTerminated
```
patriknw: what, why?
if we can't be without it internally, we can change it to `private[akka]` instead of removing it

akka-contrib/src/main/scala/akka/contrib/pattern/DistributedPubSubMediator.scala
```diff
@@ -191,7 +191,7 @@ object DistributedPubSubMediator {
 
       def business: Receive
 
-      def receive = business orElse defaultReceive
+      def receive = business.orElse[Any, Unit](defaultReceive)
```
patriknw: ah, that is how to do it (we have some of those in streams also

patriknw:
LGTM, when we have discussed `isTerminated`

akka-actor/src/main/scala/akka/actor/ActorRef.scala
rkuhn: it is `private[akka]` already by virtue of being on a class marked as such

patriknw: ok

akka-actor-tests/src/test/scala/akka/serialization/SerializeSpec.scala
rkuhn: yes, exactly; this one is quite stupid in that it cannot be switched off, but it has caught a few mistakes for me in the past so I’d like to keep it on

rkuhn:
The test failure was in `akka.cluster.Stress`, most certainly not related.

As a side note: even if we switch off the remaining warnings in ActorRefProvider (by using `-Xlint:-inaccessible`) we still cannot turn warnings into errors because of the deprecations—I think this is a bug in scalac in that one warning will be emitted even if the deprecation warnings themselves are off (the warning says that there have been deprecation warnings).

Issue #13875, opened 2014-02-14
Exclude unreachable observations from downed node in convergence check


patriknw:
imported from https://www.assembla.com/spaces/akka/tickets/3875

Investigate if unreachable observations from downed node can be excluded in the convergence check.
Discussed in context of the failure in SurviveNetworkInstabilitySpec [3871](https://www.assembla.com/spaces/akka/tickets/3871).

patriknw:
SurviveNetworkInstabilitySpec sometimes fails because of this issue.
It is expected that a quarantined node is to be removed from the cluster. The test sometimes fails because the quarantined node is not removed by the leader.

    Set(akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@first, akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@seventh, akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@sixth, akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@third, akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@second, akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@fourth, akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@fifth) contained element akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@second (SurviveNetworkInstabilitySpec.scala:292)

The reason for that is that it is not convergence, and the reason for that is that the quarantined node (`second`) has marked another node (`sixth`) as unreachable and been able to spread that information.

See the second row in the below reachability table.

    unreachable: Set(Member(address = akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@sixth, status = Up), Member(address = akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@second, status = Down))


    reachability: 
    akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@seventh -> akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@second: Unreachable [Terminated] (7)
    akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@second -> akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@sixth: Unreachable [Unreachable] (13)
[...]

Issue #16224, opened 2014-11-05
Cluster node can't join after restart


bantonsson:
As described here https://groups.google.com/forum/#!topic/akka-user/AdRSv2yuwo4 a quickly restarted node starts responding to heartbeats and becomes reachable according to the cluster even though it is a new node and is not allowed to join.

marekzebrowski:
I had very similar situation in my cluster:
I have 3 nodes:
app1, app2 - those are seed nodes, client-  supposed to connect to cluster
after restart of app1, app2, client, app2 was not receiving messages from cluster.

Then, after restarting app2, cluster dissolved - client and app1 were quarantained.
Only after restarting app1, app2 and client cluster formed again.

It was on EC2 so probably flaky networking has something to do with it

patriknw:
Please try [2.3.9](http://akka.io/news/2015/01/19/akka-2.3.9-released.html)

hepin1989:
@patriknw yes,we have upgrade to 2.3.9,and I will update with more information.

previous version.a quick start will spend a long time to let the restart node to rejoin the cluster or could never join it.

hepin1989:
Same issue after upgrade to 2.3.9

I have an idea that:
If in we have ```Node A ```and``` Node B``` in the cluster,and with ```Node A``` as seeds node,and ```Node B ``` join it.

when the ```Node B``` restart and then rejoin with a different UID.how about let the ```Node A```down the previous one first cause it was in the ```unreachable``` state,and then let the restart one join in?

maybe we could add a new configuration to support this as ```akka.cluster.quick-restart = on ```

patriknw:
Then the issue is that you have not downed the node and it is therefore not removed from the cluster. The joining process to seed nodes will retry until it is successful (or you abort it).

Your suggestion of a new feature that will automatically down a node when a new join attempt is made for same host:port with different uid sounds interesting. Please create a new issue for that.

hepin1989:
@patriknw The real issue is not that ```Node B ``` will try and try until it finally rejoin in ,but that ,even I have the setting with ```akka.cluster. auto-down-unreachable-after = 10s ``` in ```Node A```,and after I restart the ```Node B```,I think how about after 1 minute I could rejoin right?Noop,some times we have to wait more time or will have to restart ```Node A``` to regroup the cluster cause there have no progress any further.There would be something be-handing this issue

patriknw:
that should not happen, then I would like to see proof, i.e. logs or a small reproducer

[...]

Issue #15439, opened 2014-06-25
FAILED: SurviveNetworkInstabilitySpec


bantonsson:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/1133

```
[JVM-1] - must down and remove quarantined node *** FAILED *** (59 seconds, 730 milliseconds)
[JVM-1]   Set(akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@a5.moxie:37927, akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@a6.moxie:50250, akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@a5.moxie:33423, akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@a5.moxie:38122, akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@a6.moxie:60022, akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@a3.moxie:46616, akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@a6.moxie:50969) contained element akka.trttl.gremlin.tcp://SurviveNetworkInstabilitySpec@a5.moxie:38122 (SurviveNetworkInstabilitySpec.scala:314)
[JVM-1]   org.scalatest.exceptions.TestFailedException:
[JVM-1]   at org.scalatest.MatchersHelper$.newTestFailedException(MatchersHelper.scala:160)
[JVM-1]   at org.scalatest.words.ResultOfNotWordForAny.contain(ResultOfNotWordForAny.scala:527)
[JVM-1]   at akka.cluster.SurviveNetworkInstabilitySpec$$anonfun$1$$anonfun$apply$mcV$sp$31$$anonfun$apply$mcV$sp$32$$anonfun$apply$mcV$sp$39$$anonfun$apply$mcV$sp$40.apply$mcV$sp(SurviveNetworkInstabilitySpec.scala:314)
[JVM-1]   at akka.cluster.SurviveNetworkInstabilitySpec$$anonfun$1$$anonfun$apply$mcV$sp$31$$anonfun$apply$mcV$sp$32$$anonfun$apply$mcV$sp$39$$anonfun$apply$mcV$sp$40.apply(SurviveNetworkInstabilitySpec.scala:314)
[...]

PR #16766, opened 2015-02-02
=clu #13875 Exclude unreachability observations from downed

patriknw:
- Skip observations from downed node (quarantined is marked down immediately)
  in convergence check
- Skip observations from downed node when picking "reachable" targets for gossip.
- This also means that we must accept gossip with own node marked as unreachable,
  but that should not be spread to the external membership events.

The SurviveNetworkInstabilitySpec failed a lot before this patch. I have verified that it does not fail any more by letting jenkins run this over the weekend.
744 successful runs of SurviveNetworkInstabilitySpec
90  successful runs of all tests in akka-cluster.

This PR also contains a cherry-pick of forgotten forward port of adjustments to SurviveNetworkInstabilitySpec.

### DIFF
--- a/akka-cluster/src/main/scala/akka/cluster/Gossip.scala
+++ b/akka-cluster/src/main/scala/akka/cluster/Gossip.scala
@@ -20,7 +20,7 @@ private[cluster] object Gossip {
     if (members.isEmpty) empty else empty.copy(members = members)
 
   private val leaderMemberStatus = Set[MemberStatus](Up, Leaving)
-  private val convergenceMemberStatus = Set[MemberStatus](Up, Leaving)
+  val convergenceMemberStatus = Set[MemberStatus](Up, Leaving) // FIXME private
   val convergenceSkipUnreachableWithMemberStatus = Set[MemberStatus](Down, Exiting)
   val removeUnreachableWithMemberStatus = Set[MemberStatus](Down, Exiting)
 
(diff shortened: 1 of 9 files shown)


patriknw:
Refs #13875
Refs #15439

hepin1989:
will this related to https://github.com/akka/akka/issues/16224 ?

patriknw:
> does this related to 16224
I don't think so. I'm stuck on 16224, in the sense that I don't know what the problem is apart from the thing I have already mentioned. Please continue discussion there.

akka-cluster/src/test/scala/akka/cluster/ClusterDomainEventPublisherSpec.scala
```diff
@@ -19,8 +19,15 @@ import akka.testkit.ImplicitSender
...
+object ClusterDomainEventPublisherSpec {
+  val config = """
+    akka.actor.provider = "akka.cluster.ClusterActorRefProvider"
+    akka.remote.netty.tcp.port = 0
+    """
+}
+
 @org.junit.runner.RunWith(classOf[org.scalatest.junit.JUnitRunner])
-class ClusterDomainEventPublisherSpec extends AkkaSpec
+class ClusterDomainEventPublisherSpec extends AkkaSpec(ClusterDomainEventPublisherSpec.config)
```
rkuhn: I’m slightly confused: did this test suite not work before?

rkuhn:
LGTM, nice work!

akka-cluster/src/test/scala/akka/cluster/ClusterDomainEventPublisherSpec.scala
patriknw: yes, the change is that the ClusterDomainEventPublisher actor now make use of `selfUniqueAddress` from the `Cluster` extension, and therefore this test must use the CARP

rkuhn: Ah, right, I didn’t connect the dots.

patriknw:
Can I merge this so that we get rid of the jenkins failures, or will someone more review it?

bantonsson:
Sorry, didn't finish it Tuesday.  I'll do that now.

akka-cluster/src/main/scala/akka/cluster/Gossip.scala
```diff
@@ -159,19 +159,26 @@ private[cluster] final case class Gossip(
...
+    val unreachable = reachabilityExcludingDownedObservers.allUnreachableOrTerminated.collect {
+      case node if (node != selfUniqueAddress) ⇒ member(node)
+    }
     unreachable.forall(m ⇒ Gossip.convergenceSkipUnreachableWithMemberStatus(m.status)) &&
       !members.exists(m ⇒ Gossip.convergenceMemberStatus(m.status) && !seenByNode(m.uniqueAddress))
   }
 
+  lazy val reachabilityExcludingDownedObservers: Reachability = {
+    val downed = members.collect { case m if m.status == Down ⇒ m }
+    if (downed.isEmpty) overview.reachability else overview.reachability.removeObservers(downed.map(_.uniqueAddress))
```
bantonsson: Should this optimization be in `removeObservers`? I mean do nothing on an empty `Set`.

bantonsson:
LGTM

akka-cluster/src/main/scala/akka/cluster/Gossip.scala
patriknw: good point,
`downed.map(_.uniqueAddress)` allocates something also when downed is empty, but I guess that is neglectable

bantonsson: Yes, I think that object allocation is negligible.

rkuhn:
Merge away!

akka-cluster/src/main/scala/akka/cluster/Gossip.scala
patriknw: ah, that is how it is spelled :)
I'll fix that and merge.

rkuhn: I’d even say “spelt” :-)

patriknw: always learning from these reviews :-)

Issue #15859, opened 2014-09-11
Solve the circular dependency between internal ActorPublisher and its backing actor in a better way.


drewhk:
The most likely approach that could work is to reverse the responsibilities, instead of ActorPublisher being pushed into the actor via ExposedPublisher it should be the actor that pushes its ref into its corresponding ActorPublisher. Rationale:

 - ActorPublisher already needs to handle concurrent state
 - An actor might back multiple Publishers while a Publisher is usually backed by one actor

rkuhn:
This is an internal cleanup and has nothing to do with the ActorRefPublisher.

patriknw:
also related to the stash workaround

patriknw:
Removed usage of stash mailbox in #16769, but leaving this ticket since it covers more than that.
For example I think we can take advantage of that most of the internal publishers only supports one subscriber.

rkuhn:
Discussion: this is not that big of an issue to warrant a fix, especially since no grandiose improvements come readily to mind.

PR #16769, opened 2015-02-02
=str #15859 Eliminate usage of Stash mailbox

patriknw:
- As reported in #16522 Stash is slow
- Nothing fancy here, but should be more efficient than Stash

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/ExposedPublisherReceive.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/ExposedPublisherReceive.scala
@@ -1,0 +1,30 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.impl
+
+import akka.actor.Actor
+
+/**
+ * INTERNAL API
+ */
+private[akka] abstract class ExposedPublisherReceive(activeReceive: Actor.Receive, unhandled: Any ⇒ Unit) extends Actor.Receive {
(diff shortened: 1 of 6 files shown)


drewhk:
Thanks, this looks like a very good first step!

LGTM

patriknw:
yes, that was my thought, low hanging improvement with a lot of value.
I will not close the ticket, since this doesn't really solve what is described there.

drewhk:
Yep, great work!

drewhk:
Double checked again, and it still looks great :)

akka-stream/src/main/scala/akka/stream/impl/ExposedPublisherReceive.scala
```diff
@@ -0,0 +1,28 @@
...
+private[akka] abstract class ExposedPublisherReceive(activeReceive: Actor.Receive, unhandled: Any ⇒ Unit) extends Actor.Receive {
+  private var stash = List.empty[Any]
+
+  def isDefinedAt(o: Any): Boolean = true
+
+  def apply(o: Any): Unit = o match {
+    case ep: ExposedPublisher ⇒
+      receiveExposedPublisher(ep)
+      // we don't use sender() so this is allright
+      stash.reverse.foreach { msg ⇒
```
patriknw: I will add a `if (stash.nonEmpty)` here, just to make sure that there is no closure instance created for the normal case

patriknw:
See #15859

Issue #15750, opened 2014-08-25
supervision: Ensure that restart directives work as expected when applied to Stream actors (part II Restart)


rkuhn:
relevant Reactive Streams state (subscribers, demand, etc.) must be stashed away in the stream supervisor to allow a Restart to work

Issue #15753, opened 2014-08-25
supervision: add supervisorStrategy setting to ActorBasedFlowMaterializer


rkuhn:
The user-level supervisorStrategy shall only apply to the user-level exceptions unwrapped from the StreamUserException. All others are infrastructure failures that lead to pipeline termination.

Issue #15751, opened 2014-08-25
supervision: Ensure that restart directives work as expected when applied to Stream actors (part III Stop)


rkuhn:
Stop must properly shut down the stream, both upstream and downstream; this must be properly distinguished from normal termination

Issue #15749, opened 2014-08-25
supervision: Ensure that restart directives work as expected when applied to Stream actors (part I Resume)


rkuhn:
local state changes must be done properly before or after invoking user code, ensuring that the actor is not in an invalid state (to make Resume work)

PR #16772, opened 2015-02-04
+str #15750 Add stream supervision

patriknw:
only a prototype to collect early feedback


akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -398,8 +399,24 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]], val forkLimit:
...
-          state.fail(e)
+          currentOp.decide(e) match {
+            case Supervision.Stop ⇒
+              println(s"# fail [$activeOpIndex]: $e") // FIXME
+              state.fail(e)
+            case Supervision.Resume ⇒
+              println(s"# resume [$activeOpIndex]: $e") // FIXME
+              // FIXME do we have to keep track of if user called ctx.push and exception was thrown afterwards
+              // FIXME `pull` can also throw
+              state.pull()
```
drewhk: can't you match on currentState?

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -398,8 +399,24 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]], val forkLimit:
         state.progress()
       } catch {
         case NonFatal(e) if lastOpFailing != activeOpIndex ⇒
+          if (Debug) printDebug()
```
drewhk: why moved here?

drewhk: ah, no it is just duplicated

drewhk:
looks promising so far!

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
patriknw: yes, I should remove that again, residue from debugging

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
patriknw: yes, or I was thinking on defining `resume` and `restart` on the state, and it can do the right thing

drewhk: yes, even better.

ktoso:
Great first step :)

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
```diff
@@ -447,42 +484,330 @@ class InterpreterSpec extends InterpreterSpecKit {
 
     }
 
-    "handle failure inside op" in new TestSetup(Seq(Map((x: Int) ⇒ if (x == 0) throw TE else x))) {
+    "emit failure when op throws" in new TestSetup(Seq(Map((x: Int) ⇒ if (x == 0) throw TE else x, stoppingDecider))) {
```
drewhk: maybe these supervision tests should be moved to another Spec, this file starts to grow very large

patriknw: good point, will do

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
```diff
@@ -447,42 +484,330 @@ class InterpreterSpec extends InterpreterSpecKit {
...
+      upstream.onNext(-1) // boom
       lastEvents() should be(Set(Cancel, OnError(TE)))
+    }
 
+    "resume when op throws in middle of the chain" in new TestSetup(Seq(
+      Map((x: Int) ⇒ x + 1, resumingDecider),
+      Map((x: Int) ⇒ if (x == 0) throw TE else x + 10, resumingDecider),
+      Map((x: Int) ⇒ x + 100, resumingDecider))) {
+
+      lastEvents() should be(Set.empty)
```
drewhk: I don't think you need to repeat this line in all test cases. It is relevant for the basic interpreter spec, but I think these supervision tests can rely on that normal operation "works as expected". Your call though.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
```diff
@@ -447,42 +484,330 @@ class InterpreterSpec extends InterpreterSpecKit {
...
+        lastEvents() should be(Set(RequestOne))
+        upstream.onNext(2)
+        lastEvents() should be(Set(OnNext(103)))
+
+        downstream.requestOne()
+        lastEvents() should be(Set(RequestOne))
+        upstream.onNext(-5) // this will trigger failure of next requestOne (pull)
+        lastEvents() should be(Set(OnNext(99)))
+
+        downstream.requestOne() // this failed, but resume will pull
```
drewhk: Comment is wrong, this is "restart" case. Will see how this will work generally for example if that onPull happened after absorbTermination.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
```diff
@@ -0,0 +1,656 @@
...
+      downstream.requestOne()
+      lastEvents() should be(Set.empty)
+
+      upstream.onNext(4)
+      lastEvents() should be(Set(OnNext(4), RequestOne))
+
+      downstream.cancel()
+      lastEvents() should be(Set(Cancel))
+    }
+
```
patriknw: @drewhk I have added a lot of tests in this InterpreterSupervisionSpec.scala file. Above this line are tests that I think work as expected. I have some troubles below this line, and would appreciate input. See FIXMEs below...

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
```diff
@@ -0,0 +1,656 @@
...
+      downstream.requestOne()
+      lastEvents() should be(Set(RequestOne, OnNext(1)))
+
+      downstream.requestOne()
+      lastEvents() should be(Set(OnNext(-1)))
+
+      upstream.onNext(2) // boom
+      lastEvents() should be(Set.empty)
+
+      downstream.requestOne() // FIXME this throws IllegalStateException: Cannot pull while holding, only pushAndPull
```
drewhk: yeah, if it is in holding state it means that the one upstream and one downstream token is "owned" by the stage. In this case it *must* call a pushAndPull. There is no way around this. (Please note that this is not interpreter specific, this will be the same issue for a hand-crafted Processor. This is true for the other FIXMEs as well)

The same problem will repeat many times: in order to keep boundedness and not lose request tokens, a pull *must* be answered by a push, which unfortunately means to provide an element upon failure.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
```diff
@@ -0,0 +1,656 @@
...
+
+    override def decide(t: Throwable): Supervision.Directive = Supervision.Restart
+
+    override def restart(): Stage[Int, Int] = {
+      sum = 0
+      this
+    }
+  }
+
+  case class BufferingTestStage(decider: Supervision.Decider, absorbTermination: Boolean = false) extends PushPullStage[Int, Int] {
```
drewhk: the name buffering is misleading, a "normal" buffer is always a detached element since it allows upstream and downstream to progress independently when the buffer is neither empty or full.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
```diff
@@ -0,0 +1,656 @@
...
+    var buf: List[Int] = Nil
+    def onPush(elem: Int, ctx: Context[Int]): Directive = {
+      buf = List(elem + 1, elem + 2, elem + 3)
+      ctx.push(elem)
+    }
+
+    override def onPull(ctx: Context[Int]): Directive = {
+      if (buf.isEmpty)
+        ctx.pull()
```
drewhk: this will not work if you have called absorbTermination. You are not allowed to call pull while ctx.isFinishing

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
```diff
@@ -0,0 +1,656 @@
...
+      upstream.onNext(1)
+      lastEvents() should be(Set(OnNext(1)))
+
+      downstream.requestOne()
+      lastEvents() should be(Set(OnNext(2)))
+
+      downstream.requestOne()
+      // 3 => boom
+      // FIXME feels wrong? it should deliver OnNext(4)?
+      lastEvents() should be(Set(RequestOne))
```
drewhk: no, if the resume strategy pulls. This is what I said regarding one-to-many stages. The stage expects a pull from downstream to serve the next element from the list, but instead the resume strategy pulls, so it gets a push from upstream! This can lead to serious violation of expectations in the stage.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
patriknw: agreed

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
patriknw: yes, I was obviously not done with this one

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
patriknw: Thanks, this is interesting. Do you agree that for a restart we could support it, because the state is supposed to be cleared, and a new pull from upstream makes sense?

We can probably not support resume when onPull throws, and must resort to fail (stop).
Note that it is not many of the built in stages that have this problem. As far as I can see it is only `Expand`.

Just for my understanding. Let's say that this custom stage would emit an element downstream here as response to the exception instead of pulling. If a special Filter stage was installed afterwards to discard that kind of error elements everything would work? I'm somewhat confused of why we could not emulate something like that?

drewhk: > Thanks, this is interesting. Do you agree that for a restart we could support it, because the state is supposed to be cleared, and a new pull from upstream makes sense?

Unfortunately, no, because a fresh Stage expects a pull as the first event (except detached stages).

> Just for my understanding. Let's say that this custom stage would emit an element downstream here as response to the exception instead of pulling. If a special Filter stage was installed afterwards to discard that kind of error elements everything would work? I'm somewhat confused of why we could not emulate something like that?

Probably it will not work, because the stage will fail the pull exactly the same way as before (all pulls look the same). But there is something we can try: if a pull fails, we can (as you suggested) try to pull the resumed stage again. That way there is no lost token, but the new pull may fail just as the previous. We have some options here:

 - hope that repeating the pull will solve the issue, so try it until we hit a retry limit, then fail
 - add a user some kind of hook were he can "heal" state. Or alternatively we turn the resume into a restart.

patriknw: I see

As discussed in the meeting I will keep it simple for now, and we can support more fancy things later, if needed.

patriknw:
@drewhk I have pushed something that I think makes sense for the stages.

* expand will always fail the stream
* for custom stages exception when in pulling or holding state will fail the stream

I will continue with `mapAsync` after lunch break.

patriknw:
I have added supervision for mapAsync and mapAsyncUnordered, see  312d40a

Remaining:
* Java api for the OperationAttributes
* `groupBy`
* `splitWhen`
* sources and sinks, but I don't dare to touch that until after the rewrite of the internals since those have changed

patriknw:
I had to rebase

patriknw:
the failure was in mvn, disabled now

patriknw:
PLS BUILD

patriknw:
@rkuhn @drewhk @bantonsson I have added java api and supervision for groupBy and splitWhen.
I think this makes a good scope for M4, so it would be a good time for review.

I will add something to the reference docs also.

patriknw:
I have added reference docs for scala. Will translate to java tomorrow morning.

patriknw:
added reference docs for java also, should be complete now

akka-docs-dev/rst/java/stream-error.rst
```diff
@@ -0,0 +1,69 @@
...
+materializing the stream. The error handling strategies are inspired by actor supervision
+strategies, but the semantics has been adapted to the domain of stream processing.
+
+Supervision Strategies
+======================
+
+There are three ways to handle exceptions from application code:
+
+* ``Stop`` - The stream is completed with failure.
+* ``Resume`` - The element is dropped and the stream continues.
```
drewhk: place a link to the graphs and deadlocks section and mention that dropping can result in deadlocks in certain scenarios, for example broadcast -> flow -> zip

drewhk: actually in this case it will not deadlock just be out of sync.

akka-stream/src/main/scala/akka/stream/impl/GroupByProcessorImpl.scala
```diff
@@ -30,22 +36,34 @@ private[akka] class GroupByProcessorImpl(settings: ActorFlowMaterializerSettings
...
   // some substreams are open now. If downstream cancels, we still continue until the substreams are closed
   val waitNext = TransferPhase(primaryInputs.NeedsInput) { () ⇒
     val elem = primaryInputs.dequeueInputElement()
-    val key = keyFor(elem)
-
-    keyToSubstreamOutput.get(key) match {
-      case Some(substream) if substream.isOpen ⇒ nextPhase(dispatchToSubstream(elem, keyToSubstreamOutput(key)))
-      case None if primaryOutputs.isOpen       ⇒ nextPhase(openSubstream(elem, key))
-      case _                                   ⇒ // stay
+    tryKeyFor(elem) match {
```
drewhk: heh, expected something more complex here, glad it turned out to be simple!

akka-stream/src/main/scala/akka/stream/impl/MapAsyncProcessorImpl.scala
```diff
@@ -126,16 +137,27 @@ private[akka] class MapAsyncProcessorImpl(_settings: ActorFlowMaterializerSettin
...
-          // f threw, propagate failure immediately
-          fail(err)
+          // f threw, handle failure immediately
+          decider(err) match {
+            case Supervision.Stop ⇒
+              fail(err)
+            case Supervision.Resume | Supervision.Restart ⇒
+              // submittedSeqNo was not increased, just continue
+              if (settings.debugLogging)
+                log.debug("Dropped element [{}] due to exception from mapAsync function: {}", elem, err.getMessage)
```
drewhk: maybe say "factory function" to make it clear

akka-stream/src/main/scala/akka/stream/impl/MapAsyncUnorderedProcessorImpl.scala
```diff
@@ -46,8 +48,17 @@ private[akka] class MapAsyncUnorderedProcessorImpl(_settings: ActorFlowMateriali
...
-    case FutureFailure(cause) ⇒
-      fail(cause)
+    case FutureFailure(in, err) ⇒
+      decider(err) match {
+        case Supervision.Stop ⇒
+          fail(err)
+        case Supervision.Resume | Supervision.Restart ⇒
+          inProgressCount -= 1
+          if (settings.debugLogging)
+            log.debug("Dropped element [{}] due to mapAsyncUnordered future was completed with exception: {}",
```
drewhk: maybe say "factory function" to make it clear

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -396,7 +403,7 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]], val forkLimit:
           ("----" * (activeOpIndex - jumpBacks(activeOpIndex) - 1))
       case Completing ⇒ padding + "---|"
       case Cancelling ⇒ padding + "|---"
-      case Failing(e) ⇒ padding + s"---X ${e.getMessage}"
+      case Failing(e) ⇒ padding + s"---X ${e.getMessage} => ${decide(e)}"
```
drewhk: thanks, great addition!

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -152,6 +152,14 @@ class Flow[-In, +Out](delegate: scaladsl.Flow[In, Out]) {
    * downstream may run in parallel and may complete in any order, but the elements that
```
drewhk: so the decision was to not support supervision on user defined transform() elements now

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
```diff
@@ -85,6 +87,22 @@ private[stream] abstract class AbstractStage[-In, Out, PushD <: Directive, PullD
    */
   def onUpstreamFailure(cause: Throwable, ctx: Ctx): TerminationDirective = ctx.fail(cause)
 
+  /**
+   * If an exception is thrown from [[#onPush]] this method is invoked to decide how
+   * to handle the exception. By default this method returns [[Supervision.Stop]].
+   *
+   * If an exception is thrown from [[#onPull]] the stream will always be completed with
+   * failure, because it is not always possible to recover from that state.
```
drewhk: maybe add some doc that the users should use their own try-catch-recover in onPull if they know how their logic is supposed to recover from that state.

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
drewhk: ah no, it is defined on Stage

drewhk:
Looks like a very good start to me! We can tweak things later as real use-case feedback come in from users.

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
patriknw: The Stage itself can define it by overriding the `decide` and `restart` methods. I was thinking about this, and to give custom stages the same capabilities as our internal stages the factory function of `transform` must be something like `GeneralFlowMaterializerSettings => Stage` or `OperationAttributes => Stage`. We can have that in mind, but for now it is out of scope.

patriknw:
squashed, will merge after validation

patriknw:
Refs #15749
Refs #15750
Refs #15751
Refs #15752
Refs #15753

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
```diff
@@ -0,0 +1,523 @@
...
+      upstream.onNext(2)
+      lastEvents() should be(Set(OnNext(2)))
+
+      downstream.requestOne()
+      lastEvents() should be(Set(RequestOne))
+      upstream.onNext(0) // boom
+      lastEvents() should be(Set(RequestOne))
+
+      downstream.requestOne()
+      lastEvents() should be(Set(RequestOne))
```
rkuhn: @drewhk Should two not be impossible? After the “boom” the ball landed upstream so we should just feed more data, right?

l15k4:
Hey, the error handling seems to be broken when a Flow throws exception. It enters the decider that returns `Resume`, but then it immediately fails. Please see https://github.com/akka/akka/issues/16577#issuecomment-76727915

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
rkuhn: @drewhk ping (I have the feeling that this wants to be sorted out)

patriknw: alright, I'm fixing this in https://github.com/akka/akka/pull/17006

PR #16773, opened 2015-02-04
=str Fix compilation error in akka-docs-java-lambda

patriknw:


ktoso:
Would be cool to include it in validation as well btw, or any reasons not to? (I know it's maven)

PR #16775, opened 2015-02-04
=pro PR validating java8 samples of akka streams (using mvn)

ktoso:
Includes `akka-docs-java-lambda` in (for streams documentation) in PR validation.
Also makes it a bit more centralised how we call out to maven.
Not very fancy yet, but good enough for the task at hand - overriding javaHome can be done via env or sbt setting (via env is there for Jenkins).

### DIFF
--- a/akka-samples/akka-docs-java-lambda/pom.xml
+++ b/akka-samples/akka-docs-java-lambda/pom.xml
@@ -6,7 +6,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <publishedAkkaVersion>2.3.7</publishedAkkaVersion>
+    <publishedAkkaVersion>2.3.9</publishedAkkaVersion>
   </properties>
 
   <groupId>sample</groupId>
@@ -27,7 +27,7 @@
     </dependency>
     <dependency>
       <groupId>com.typesafe.akka</groupId>
-      <artifactId>akka-stream_2.10</artifactId>
+      <artifactId>akka-stream-experimental_2.10</artifactId>
       <version>1.0-SNAPSHOT</version>
     </dependency>
     <dependency>
--- a/project/Maven.scala
+++ b/project/Maven.scala
@@ -1,0 +1,54 @@
+/*
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka
+
+import java.io.File
+
+import sbt._
+
+object Maven {
+  val mvn = config("maven")
+
+  val projects = settingKey[Seq[File]]("Projects which should be built by calling maven")
+  val defaultCommands = settingKey[Seq[String]]("Commands to be called on each maven project")
+  val javaHomeOverride = settingKey[Option[File]]("JAVA_HOME to be used while building projects with maven")
+  val mavenLocalRepoOverride = settingKey[Option[File]]("Overrides local maven repository location if set")
+  
+  val mvnExec = taskKey[Unit]("Test all maven projects (as listed in projectsToTest)")
+
+  def akkaBuildJava8HomeEnv() = sys.env.get("AKKA_BUILD_JAVA8_HOME").map(file)
+
+  lazy val settings: Seq[Setting[_]] = Seq(
+    projects in mvn := Nil,
+    defaultCommands in mvn := "clean" :: "test" :: Nil,
+    javaHomeOverride in mvn := akkaBuildJava8HomeEnv(),
+    mvnExec := {
+      val javaHome = (javaHomeOverride in mvn).value
+      val prjs = (projects in mvn).value
+      val cmnds = (defaultCommands in mvn).value
+      prjs foreach { p ⇒ mvnRun(javaHome, p.getAbsolutePath, cmnds: _*) }
+    }
+  )
+
+  /** Runs maven commands (blocking) on given project directory (by cd-ing into it) */
+  def mvnRun(proj: String, commands: String*): Unit =
+    mvnRun(None, proj, commands:_ *)
+
+  /** Runs maven commands (blocking) on given project directory (by cd-ing into it) */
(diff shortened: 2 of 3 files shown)


akka-samples/akka-docs-java-lambda/pom.xml
```diff
@@ -27,7 +27,7 @@
     </dependency>
     <dependency>
       <groupId>com.typesafe.akka</groupId>
-      <artifactId>akka-stream_2.10</artifactId>
+      <artifactId>akka-stream-experimental_2.10</artifactId>
```
ktoso: is there something I missed why this was not `-experimental`?
When we `publishM2` it publishes as experimental (I moved the publish step in front the pr validation task on jenkins - AFAICS this should work)

ktoso:
Failure was before I managed to swap publishM2 with validatePullRequest, re running.

akka-samples/akka-docs-java-lambda/pom.xml
patriknw: I guess this has never been used from this branch. I use sbt.

patriknw:
LGTM, I guess

akka-samples/akka-docs-java-lambda/pom.xml
```diff
@@ -6,7 +6,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <publishedAkkaVersion>2.3.7</publishedAkkaVersion>
+    <publishedAkkaVersion>2.3.9</publishedAkkaVersion>
```
rkuhn: Is this how it is done in the other samples, or do they use `2.3-SNAPSHOT` in order to pick up the automatic rename?

project/Maven.scala
```diff
@@ -0,0 +1,55 @@
...
+import sbt._
+
+object Maven {
+  val mvn = config("maven")
+
+  val projects = settingKey[Seq[File]]("Projects which should be built by calling maven")
+  val defaultCommands = settingKey[Seq[String]]("Commands to be called on each maven project")
+  val javaHomeOverride = settingKey[Option[File]]("JAVA_HOME to be used while building projects with maven")
+  
+  val testAll = taskKey[Unit]("Test all maven projects (as listed in projectsToTest)")
```
rkuhn: projectsToTest?

drewhk:
LGTM

rkuhn:
LGTM apart from the nitpicks

akka-samples/akka-docs-java-lambda/pom.xml
patriknw: this is the stream dependency, and that must be fixed (2.3-SNAPSHOT would be from the release-2.3-dev branch)
messy, but things will be cleaned up when we leave this -dev branch

rkuhn: thanks for clarifying—these things will be cleaned up soon in other ways as well (removing the samples from the main build)

patriknw: yes, and using java8 as primary target

ktoso: can't wait for targeting j8 ;)

project/Maven.scala
ktoso: this is the task so projectsToTest sounds a bit off, it actually runs all tests (for all `projects`). 
Maybe `mvn test` => `mvnTest` would make the most sense here then.

ktoso:
Made mvn aware of custom local maven repo location on jenkins. Watiing for validation

bantonsson:
How does this interact with the `mavenLocalResolver` and `mavenLocalResolverSettings` in the `AkkaBuild.scala`. I'm not completely sure what we are trying to do here. The path that mvn should use to find the local repo should be (was at least) configured for the user running jenkins, and the version of sbt we had at the time didn't pick up on that, so we had to override it in the AkkaBuild by hand, hence the `mavenLocalResolver` and `mavenLocalResolverSettings`.

ktoso:
Very true actually, thanks for the reminder @bantonsson.
I'll revert that change and look deeper into the dep-res maven failure hm.

rkuhn:
failed due to undefined sbt settings

ktoso:
All good now.
I had missed that `publishM2` was disabled on this branch in the main project so the jars wouldn't end up being published anyway.

2m:
It seems that more projects than `akka-docs-java-lambda` were disabled by https://github.com/akka/akka/commit/20c74b6116d165fad291d49fc9a12f936fb16b40 We should add those back here as well.

Merging build files back to master is going to be fun. :)

ktoso:
This being the streams branch I think we only need that one (`akka-docs-java-lambda`) as the other projects shouldn't really be touched on `release-2.3-dev`?

ktoso:
Merging back to master may be easier to just not merge but "rip out the streams directory and update the master build to include those" I guess by the way.

rkuhn:
Yes, that is the plan: there will be no “merge” into master. And the build definition in master will look rather different at that point as well, I assume, so it will be a completely new re-integration.

ktoso:
Shall I merge then? :-) (on my ☎)

2m:
It would be nice to not forget this integration of Maven to sbt build, so we could use this in master as-well instead of extra build steps sh file.

patriknw:
@ktoso what did we conclude? move away from mvn for this project or merge?

ktoso:
Merge as moving away would take too much time (java version weirdness with sbt, can go into details - in essence it should have been easy to say "this project with java8, here's javahome", but it wasn't).

The build will change anyway once we merge into the 2.3 branch (for 1.0 AFAIR), so not worth the time to fix this here since it will need a rewrite then anyway. Pls merge, and my java cookbook afterwards unless problems found in there :)

patriknw:
@ktoso right decision, in my opinon

PR #16776, opened 2015-02-04
=str java documentation for integrating streams

ktoso:
ready to be reviewed now

### DIFF
--- a/akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ActorPublisherDocTest.java
+++ b/akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ActorPublisherDocTest.java
@@ -1,0 +1,162 @@
+/*
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package docs.stream;
+
+import akka.actor.ActorRef;
+import akka.actor.ActorSystem;
+import akka.actor.Props;
+import akka.japi.pf.ReceiveBuilder;
+import akka.stream.ActorFlowMaterializer;
+import akka.stream.FlowMaterializer;
+import akka.stream.actor.AbstractActorPublisher;
+import akka.stream.actor.ActorPublisherMessage;
+import akka.stream.javadsl.KeyedSource;
+import akka.stream.javadsl.MaterializedMap;
+import akka.stream.javadsl.Sink;
+import akka.stream.javadsl.Source;
+import akka.testkit.JavaTestKit;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ActorPublisherDocTest {
+
+  static ActorSystem system;
+  
+
+  @BeforeClass
+  public static void setup() {
+    system = ActorSystem.create("ActorPublisherDocTest");
+  }
+
+  @AfterClass
+  public static void tearDown() {
+    JavaTestKit.shutdownActorSystem(system);
+    system = null;
+  }
+  
+  final FlowMaterializer mat = ActorFlowMaterializer.create(system);
+  
+    //#job-manager
+    public static class JobManagerProtocol {
+      final public static class Job {
+        public final String payload;
+
+        public Job(String payload) {
+          this.payload = payload;
+        }
+
+      }
+
+      public static class JobAcceptedMessage {
+        @Override
+        public String toString() {
+          return "JobAccepted";
+        }
+      }
+      public static final JobAcceptedMessage JobAccepted = new JobAcceptedMessage();
+
+      public static class JobDeniedMessage {
+        @Override
+        public String toString() {
+          return "JobDenied";
+        }
+      }
+      public static final JobDeniedMessage JobDenied = new JobDeniedMessage();
+    }
+    public static class JobManager extends AbstractActorPublisher<JobManagerProtocol.Job> {
+      
+      public static Props props() { return Props.create(JobManager.class); }
+
+      private final int MAX_BUFFER_SIZE = 100;
+      private final List<JobManagerProtocol.Job> buf = new ArrayList<>();
+
+      public JobManager() {
+        receive(ReceiveBuilder.
+          match(JobManagerProtocol.Job.class, job -> buf.size() == MAX_BUFFER_SIZE, job -> {
+            sender().tell(JobManagerProtocol.JobDenied, self());
+          }).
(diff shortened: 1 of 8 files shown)


patriknw:
@ktoso is this ready for review?

ktoso:
Pushing last tests right now actually, I'll ping

ktoso:
ping @patriknw pushed the last part now (reactive streams).
Fun fact, seeing more red code in Java (8) since in Scala since a few years - very weird at times..

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ActorPublisherDocTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+      }
+
+      public static class JobAcceptedMessage {}
+      public static final JobAcceptedMessage JobAccepted = new JobAcceptedMessage () { 
+        @Override
+        public String toString() { return "JobAccepted"; }
+      };
+
+      static class JobDeniedMessage {}
```
patriknw: public

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ActorPublisherDocTest.java
```diff
@@ -0,0 +1,153 @@
...
+        public Job(String payload) {
+          this.payload = payload;
+        }
+
+      }
+
+      public static class JobAcceptedMessage {}
+      public static final JobAcceptedMessage JobAccepted = new JobAcceptedMessage () { 
+        @Override
+        public String toString() { return "JobAccepted"; }
```
patriknw: doesn't it make more sense to have the toString in the `JobAcceptedMessage` class?

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ActorPublisherDocTest.java
```diff
@@ -0,0 +1,153 @@
...
+        public String toString() { return "JobDenied"; }
+      };
+
+    }
+  
+    public static class JobManager extends AbstractActorPublisher<JobManagerProtocol.Job> {
+      
+      public static Props props() { return Props.create(JobManager.class); }
+
+      private final int MaxBufferSize = 100;
```
patriknw: shall we use java constant style?

patriknw: otherwise `maxBufferSize`

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ActorPublisherDocTest.java
```diff
@@ -0,0 +1,153 @@
...
+            final List<JobManagerProtocol.Job> took =
+                buf.subList(0, Math.min(buf.size(), (int) totalDemand()));
+            took.forEach(this::onNext);
+            buf.removeAll(took);
+          } else {
+            final List<JobManagerProtocol.Job> took =
+                buf.subList(0, Math.min(buf.size(), Integer.MAX_VALUE));
+            took.forEach(this::onNext);
+            buf.removeAll(took);
+            deliverBuf(); // recurse
```
patriknw: is it a problem that this is not tailrec?

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/IntegrationDocTest.java
```diff
@@ -0,0 +1,567 @@
...
+        assertTrue(set.contains("after: I"));
+        assertTrue(set.contains("after: J"));
+      }
+    };
+  }
+  
+  
+  
+  @SuppressWarnings("UnusedParameters")
+  static <T> void println(T s) { /* ignore */ }
```
patriknw: yes, I was also thinking about how we should do that. The problem is that the user can't copy the code and just run it, since this doesn't exist. I don't know.

patriknw:
LGTM

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ActorPublisherDocTest.java
ktoso: thanks

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ActorPublisherDocTest.java
ktoso: yup, expected ids here to be used, but ended up being id-less messages

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ActorPublisherDocTest.java
ktoso: java-constant it is :)

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ActorPublisherDocTest.java
ktoso: shouldn't really be a problem here imo (would have to be a really huge buf), but a while loop is maybe safer to get people used to for those. Made it `while(true) + break`

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/IntegrationDocTest.java
ktoso: Check this out!

```java
class MockSystem {
        class Println {
          public <T> void println(T s) {
            if (s.toString().startsWith("after:"))
              probe.ref().tell(s, ActorRef.noSender());
          }
        }

        public final Println out = new Println();
      }
      private final MockSystem System = new MockSystem();
```

:-)

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ActorPublisherDocTest.java
patriknw: thanks

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/IntegrationDocTest.java
patriknw: yes, that is nice, we should use it at all places in the java8 stream sampes

drewhk:
LGTM

PR #16780, opened 2015-02-06
=clu #13875 Exclude unreachability observations from downed (for validation)

patriknw:
- Skip observations from downed node (quarantined is marked down immediately)
  in convergence check
- Skip observations from downed node when picking "reachable" targets for gossip.
- This also means that we must accept gossip with own node marked as unreachable,
  but that should not be spread to the external membership events.

(cherry picked from commit 71ccb4c21b1acd11650315d536997ea5247c547b)

Conflicts:
    akka-cluster/src/main/scala/akka/cluster/ClusterEvent.scala
- Also added mima filters for the changes of the internal classes


rkuhn:
failure was in LeaderElectionWithAccrualFailureDetector

patriknw:
PLS BUILD

patriknw:
Refs #13875

Issue #16683, opened 2015-01-20
Bug in Stream Documentation: StatefulStage


kodemaniak:
I think there is a bug in the StatefulStage documentation (http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0-M2/scala/stream-customize.html#Using_StatefulStage). The example overrides the initial state as a val, but that leads to a null pointer exception. Using a def resolves that issue.

kodemaniak:
BTW, where are the sources for the stream docs. I was unable to find them.

patriknw:
akka-docs-dev/rst/ in release-2.3-dev branch

kodemaniak:
Thanks! I totally missed that directory.

ktoso:
Thanks for reporting @kodemaniak! Resolved now and shipping with M3 :)

PR #16786, opened 2015-02-06
=str #16683 fixes NPE in stateful stage example

ktoso:
resolves https://github.com/akka/akka/issues/16683


drewhk:
Ah, good one, thanks. LGTM

patriknw:
LGTM

rkuhn:
LGTM; this should also be mentioned clearly in the docs because it may seem surprising to users (I mean the scaladoc of the `initial` method)

ktoso:
Good idea, adding docs

ktoso:
```
**Warning:** This method must not be implemented as `val`.
```
Good enough or needs full explanation?

drewhk:
Fine by me, any furhter explanation would need to refer to internals in some way.

rkuhn:
Good enough, I think.

Issue #16745, opened 2015-01-28
Better Error Logging


reid-spencer:
When exceptions occur, getting a useful error message to log is something of an art form rather than simplicity. Getting useful messages logged is very important in highly asynchronous systems such as Akka because often debuggers will simply avoid the (timing related) bug -- the infamous heisenbug. This issue suggests that Akka ought to have a general mechanism for extracting the appropriate error message from an exception. 

Here are some examples of the issues we have with generating useful log messages from exceptions:
* The getMessage method in some exceptions yields a null value (e.g. NullPointerException)
* The getMessage method in some exceptions yields empty string (e.g. SocketTimeOutException)
* The cause object is not used but some other object is instead (e.g. InvocationTargetException)
* The message is unintelligible without the exception class name (e.g. NullPointerException)
* The message is delegated to the cause but the cause is null so the message is null (IllegalArgumentException)
* etc.

To help Akka provide helpful log messages, I am soliciting feedback from Akka developers about this issue, specifically:
* Annoyances you've had tracking a bug that was made harder because of poor log messages from exceptions. 
* Ideas you have about how to, and how not to, implement a general mechanism to address this. 
* Ways to do this with least impact to the code (e.g. implement it in akka logging)
* Specific exceptions you know are problematic.

The motivation for this came from this PR: https://github.com/akka/akka/pull/16725

patriknw:
Thanks, makes a lot of sense to me.
Some kind of utility to extract a good message from an exception sounds like a great start.
+1 for collecting more feedback from users

drewhk:
+1, this is always a problem

rkuhn:
+1 for a general mechanism for this

reid-spencer:
Another thing I've run into with Akka logging is its interface (LoggingAdapter) which is, by my estimation, deficient in two regards: 

1. It doesn't implement the SLF4J Logger interface, which means that:
    * you must use "warning" instead of "warn"
    * "trace" is not supported
    * the arguments for error method are different
    * you can't provide exceptions to warnings, info or debug
    * parameters are substituted with place holders instead of string interpolation
    * the Typesafe ScalaLogging interface cannot be used because it requires SLF4J Logger

2. It doesn't support the Typesafe ScalaLogging interface, which means that:
    * markers are not easily supported
    * computation of messages are not avoided if the debugging level isn't active
    * computation of messages are not diverted to the logging actor

[...]

PR #16789, opened 2015-02-06
=act #16745 Improve error logging in akka.io.SelectionHandler

reid-spencer:
When a java.net.BindException occurs, it produces an exception with a cause and
message that is "null" because it's cause returns no message. This notably occurs
when a java.net.BindException occurs during construction of a TCP selector. This
patch causes a log message like this:
    java.net.BindException: Can't assign requested address
instead of like this:
    null
which is quite a bit more helpful when diagnosing connection issues.


reid-spencer:
This is a replacement for #16725

rkuhn:
OK TO TEST

rkuhn:
LGTM, thanks!

The test failure was unrelated.

rkuhn:
backported to release-2.3

Issue #16726, opened 2015-01-27
add quick-restart support for akka cluster


hepin1989:
how about automatically down a node that is unreachable now when a new join attempt is made for same ```{ host:port}``` with different uid?
by this way, a node could quick restart and then rejoin the cluster again.
ref to https://github.com/akka/akka/issues/16224

otherwise it takes long time to maintain/hot fix.

behrad:
+1

drewhk:
Probably a good idea since the same host port is then proved to be unaccessible with the old UID.

patriknw:
I agree. This is a nice feature. The way I think we should implement it is that a join attempt with different UID is received we down the old member and let the normal downing/removal procedure continue. The new new node will have to try joining again, and eventually when the old has been removed it will be accepted.

hepin1989:
@patriknw Yes,the issue here https://github.com/akka/akka/issues/16224 seems after a quick restart.some nodes see the rejoining one ```UP```and someone see it ```UNREACHABLE```，then there seems have a blocker.

patriknw:
backport to 2.3

Issue #16736, opened 2015-01-28
Cluster.leave andThen system.shutdown


patriknw:
It would be nice with an option to shutdown the actor system after cluster leave, to avoid code like this:

    cluster.leave(cluster.selfAddress)
    Thread.sleep(10000)
    system.shutdown()

It should listen to cluster membership events and perform `system.shutdown` when it sees `MemberRemoved(selfAddress)`

patriknw:
@marekzebrowski I created this ticket based on your feedback in #16224
// cc @hepin1989

hepin1989:
I think we need trapping CTRL-C too

hepin1989:
we are using it like this
```
  engineLifeCycle.addPostStopHook {
    () =>
      import scala.concurrent.duration._
      if (!inCluster) {
        val failure = new IllegalStateException("not join the cluster yet")
        joinClusterPromise.tryFailure(failure)
        leaveClusterPromise.tryFailure(failure)
        downClusterPromise.tryFailure(failure)
      }
      cluster.leave(cluster.selfAddress)
      Await.ready(leaveClusterPromise.future, 10.seconds)
  }
```
&
```
 case MemberUp(member) =>
          log.info("member :{} up,is self :{}", member, member.address == cluster.selfAddress)
          if (member.address == cluster.selfAddress) {
            joinClusterPromise.trySuccess((true, member.address))
            inCluster = true
          }
        case MemberExited(member) =>
          log.info("member :{} exited,is self :{}", member, member.address == cluster.selfAddress)
          if (member.address == cluster.selfAddress) {
            leaveClusterPromise.trySuccess((true, member.address))
          }
        case MemberRemoved(member, previousStatus) =>
          log.info("member :{} removed,previousStatus:{},is self :{}", member, previousStatus, member.address == cluster.selfAddress)
          if (member.address == cluster.selfAddress) {
            downClusterPromise.trySuccess((true, member.address))
          }
```

michaelom:
> It would be nice with an option to shutdown the actor system after cluster leave...

+1

hepin1989:
how about add a listener to the cluster ,like ```registerOnMemberRemoved```?

PR #16792, opened 2015-02-08
=clu #16726 Down member automatically when restarted

patriknw:
- When new uid is seen in join attempt we can down existing
  member and thereby new restarted node will be able to join
  in later retried join attempt without relying on auto-down.

### DIFF
--- a/akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
+++ b/akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
@@ -448,31 +448,36 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
 
       // check by address without uid to make sure that node with same host:port is not allowed
       // to join until previous node with that host:port has been removed from the cluster
-      val alreadyMember = localMembers.exists(_.address == node.address)
-      val isUnreachable = !latestGossip.overview.reachability.isReachable(node)
+      localMembers.find(_.address == node.address) match {
+        case Some(m) if m.uniqueAddress == node ⇒
+          // node retried join attempt, probably due to lost Welcome message
+          logInfo("Existing member [{}] is joining again.", m)
+          if (node != selfUniqueAddress)
+            sender() ! Welcome(selfUniqueAddress, latestGossip)
+        case Some(m) ⇒
+          // node restarted, same host:port as existing member, but with different uid
+          // safe to down and later remove existing member
+          // new node will retry join
+          logInfo("New incarnation of existing member [{}] is trying to join. " +
+            "Existing will be removed from the cluster and then new member will be allowed to join.", m)
+          if (m.status != Down && m.status != Leaving && m.status != Exiting)
+            downing(m.address)
+        case None ⇒
+          // remove the node from the failure detector
+          failureDetector.remove(node.address)
 
-      if (alreadyMember)
-        logInfo("Existing member [{}] is trying to join, ignoring", node)
-      else if (isUnreachable)
-        logInfo("Unreachable member [{}] is trying to join, ignoring", node)
-      else {
+          // add joining node as Joining
+          // add self in case someone else joins before self has joined (Set discards duplicates)
+          val newMembers = localMembers + Member(node, roles) + Member(selfUniqueAddress, cluster.selfRoles)
+          val newGossip = latestGossip copy (members = newMembers)
 
-        // remove the node from the failure detector
-        failureDetector.remove(node.address)
+          updateLatestGossip(newGossip)
 
-        // add joining node as Joining
-        // add self in case someone else joins before self has joined (Set discards duplicates)
-        val newMembers = localMembers + Member(node, roles) + Member(selfUniqueAddress, cluster.selfRoles)
-        val newGossip = latestGossip copy (members = newMembers)
+          logInfo("Node [{}] is JOINING, roles [{}]", node.address, roles.mkString(", "))
+          if (node != selfUniqueAddress)
+            sender() ! Welcome(selfUniqueAddress, latestGossip)
 
-        updateLatestGossip(newGossip)
-
-        logInfo("Node [{}] is JOINING, roles [{}]", node.address, roles.mkString(", "))
-        if (node != selfUniqueAddress) {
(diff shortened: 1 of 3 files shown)


patriknw:
cc: @hepin1989

patriknw:
Refs #16726

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -448,31 +448,36 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
...
+          if (node != selfUniqueAddress)
+            sender() ! Welcome(selfUniqueAddress, latestGossip)
+        case Some(m) ⇒
+          // node restarted, same host:port as existing member, but with different uid
+          // safe to down and later remove existing member
+          // new node will retry join
+          logInfo("New incarnation of existing member [{}] is trying to join. " +
+            "Existing will be removed from the cluster and then new member will be allowed to join.", node)
+          if (m.status != Down && m.status != Leaving && m.status != Exiting)
+            downing(m.address)
```
ktoso: nice :)

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -792,11 +797,6 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
     val localOverview = localGossip.overview
     val localSeen = localOverview.seen
 
-    val hasPartionHandoffCompletedSuccessfully: Boolean = {
```
ktoso: Is this something worth noting as trello / ticket and revisiting?

ktoso:
LGTM!

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -448,31 +448,36 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
 
       // check by address without uid to make sure that node with same host:port is not allowed
       // to join until previous node with that host:port has been removed from the cluster
-      val alreadyMember = localMembers.exists(_.address == node.address)
-      val isUnreachable = !latestGossip.overview.reachability.isReachable(node)
+      localMembers.find(_.address == node.address) match {
+        case Some(m) if m.uniqueAddress == node ⇒
+          // node retried join attempt, probably due to lost Welcome message
+          logInfo("Existing member [{}] is joining again.", node)
```
drewhk: good one! this is important info

drewhk:
LGTM, nice addition!

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
hepin1989: should log the existing member's current state too.

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
hepin1989: should we just down the node which is currently unreachable?and log the node's current state too?

hepin1989:
https://github.com/akka/akka/issues/16736
this one will need done too,try to leaving the cluster when the akka system is going down.
both with this,will help a lot.

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
patriknw: yes, we can log `m` instead of `node`

I don't understand what you mean by "should we just down the node which is currently unreachable".
Removal should go through the normal process.

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
patriknw: These ideas have been abandoned, boy scout in duty.
I noticed that these ideas are still in the Cluster Specification as future ideas. I'll create a ticket to remove that.

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
hepin1989: how about the existing member is in the state of ```UP```?
how about a attack when someone signal a ```JOIN``` command?
cause akka2.4 will add support for binding on different address.I think we may think about this too.
different host : addr but same binding-host : binding-addr,this will cause the UP node down if that was an attack.

I think the downing the existing one should only progress when the existing one is ```UNREACHABLE```.

patriknw: no, the identifier of a member is hostname+port+uid, that will not change with the support of binding to a different external address, unless I have missed something completely (and then we have major problems with that host binding feature).

If we get a join request with new uid and same hostname+port as existing member it is know that the old uid is gone, and can be downed (independent of status or reachability of existing member). It will become unreachable if it's not already unreachable and then it can be removed by the normal removal procedure.

hepin1989: https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/transport/netty/NettyTransport.scala#L127

two nodes could have the same ```bind-hostname```,but binding on different ```hostname```

patriknw:
PLS BUILD

behrad:
will we have this also in 2.3 branch? or just in 2.4 !?

hepin1989:
@behrad seems just in 2.4

behrad:
this is the main and most important feature for node connectivity reliability in different scenarios (JVM crashes, Ctrl+C, Network problems,..)  @hepin1989 , and depending on `auto-down` helps on node restarts but breaks on temporal network departures. 

I assume Akka's goal should be to provide semantics that a robust/HA cluster of nodes can be built upon and this feature helps :)

agreed?

patriknw:
@behrad we have decided to backport it to 2.3 even thought its almost like a new feature, https://github.com/akka/akka/pull/17148

behrad:
great @patriknw Happy to hear this :+1: 

this way even `2.3` users would benefit from this.

hepin1989:
@patriknw wow!

PR #16793, opened 2015-02-08
add akka.util.TypedMultiMap

rkuhn:
This greatly simplifies the Receptionist by providing the ability to
express a polymorphic map that calculates the type of each value from
the type of its key. Another possible use of this map is to express the
Extensions map in a type-safe fashion (no casts needed).

### DIFF
--- a/akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
+++ b/akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
@@ -1,0 +1,106 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.util
+
+import language.higherKinds
+
+/**
+ * An immutable multi-map that expresses the value type as a type function of the key
+ * type. Create it with a type constructor that expresses the relationship:
+ *
+ * {{{
+ * trait Key { type Type = T }
+ * case class MyValue[T](...)
+ *
+ * // type function from Key to MyValue
+ * type KV[K <: Key] = MyValue[K#Type]
+ *
+ * val map = TypedMultiMap.empty[Key, KV]
+ *
+ * // a plain Int => String map would use this function:
+ * type plain[K <: Int] = String
+ *
+ * // a map T => T would use this function:
+ * type identity[T <: AnyRef] = T
+ * }}}
+ *
+ * Caveat: using keys which take type parameters does not work due to conflicts
+ * with the existential interpretation of `Key[_]`. A workaround is to define
+ * a key type like above and provide a subtype that provides its type parameter
+ * as type member `Type`.
+ */
+class TypedMultiMap[T <: AnyRef, K[_ <: T]] private (private val map: Map[T, Set[Any]]) {
+
+  /**
+   * Return the set of keys which are mapped to non-empty value sets.
+   */
+  def keySet: Set[T] = map.keySet
+
+  /**
+   * Return a map that has the given value added to the mappings for the given key.
+   */
+  def inserted(key: T)(value: K[key.type]): TypedMultiMap[T, K] = {
+    val set = map.get(key) match {
+      case Some(s) ⇒ s
+      case None    ⇒ Set.empty[Any]
+    }
+    new TypedMultiMap[T, K](map.updated(key, set + value))
+  }
+
+  /**
+   * Obtain all mappings for the given key.
+   */
+  def get(key: T): Set[K[key.type]] =
+    map.get(key) match {
+      case Some(s) ⇒ s.asInstanceOf[Set[K[key.type]]]
+      case None    ⇒ Set.empty
+    }
+
+  /**
+   * Return a map that has the given value removed from all keys.
+   */
+  def valueRemoved(value: Any): TypedMultiMap[T, K] = {
+    val s = Set(value)
+    val m = map.collect {
+      case (k, set) if set != s ⇒ (k, set - value)
+    }
+    new TypedMultiMap[T, K](m)
+  }
+
+  /**
+   * Return a map that has all mappings for the given key removed.
+   */
+  def keyRemoved(key: T): TypedMultiMap[T, K] = new TypedMultiMap[T, K](map - key)
+
+  /**
+   * Return a map that has the given mapping from the given key removed.
+   */
+  def removed(key: T)(value: K[key.type]): TypedMultiMap[T, K] = {
+    map.get(key) match {
+      case None ⇒ this
+      case Some(set) ⇒
+        if (set(value)) {
+          val newset = set - value
+          val newmap = if (newset.isEmpty) map - key else map.updated(key, newset)
+          new TypedMultiMap[T, K](newmap)
+        } else this
+    }
+  }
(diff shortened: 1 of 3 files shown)


akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
```diff
@@ -0,0 +1,68 @@
...
+ * with the existential interpretation of `Key[_]`. A workaround is to define
+ * a key type like above and provide a subtype that provides its type parameter
+ * as type member `Type`.
+ */
+class TypedMultiMap[T <: AnyRef, K[_ <: T]] private (map: Map[T, Set[Any]]) {
+
+  def keySet: Set[T] = map.keySet
+
+  def inserted(key: T)(value: K[key.type]): TypedMultiMap[T, K] = {
+    val set = map.get(key).getOrElse(Set.empty)
```
patriknw: match instead of getOrElse to save an allocation?

akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
```diff
@@ -0,0 +1,68 @@
...
+    new TypedMultiMap[T, K](map.updated(key, set + value))
+  }
+
+  def get(key: T): Set[K[key.type]] =
```
patriknw: should this return `Set` or `Option[Set]`?

making this Option and also provide a getOrElse is perhaps more consistent with scala collections?

akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
```diff
@@ -0,0 +1,68 @@
...
+      case None ⇒ this
+      case Some(set) ⇒
+        if (set(value)) {
+          val newset = set - value
+          val newmap = if (newset.isEmpty) map - key else map.updated(key, newset)
+          new TypedMultiMap[T, K](newmap)
+        } else this
+    }
+  }
+}
```
patriknw: equals, hashCode, toString

akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
```diff
@@ -0,0 +1,68 @@
...
+  def valueRemoved(value: Any): TypedMultiMap[T, K] = {
+    val s = Set(value)
+    val m = map.collect {
+      case (k, set) if set != s ⇒ (k, set - value)
+    }
+    new TypedMultiMap[T, K](m)
+  }
+
+  def removed(key: T)(value: K[key.type]): TypedMultiMap[T, K] = {
+    map get key match {
```
patriknw: map.get(key) match, please

akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
```diff
@@ -0,0 +1,68 @@
...
+          val newset = set - value
+          val newmap = if (newset.isEmpty) map - key else map.updated(key, newset)
+          new TypedMultiMap[T, K](newmap)
+        } else this
+    }
+  }
+}
+
+object TypedMultiMap {
+  def empty[T <: AnyRef, K[_ <: T]]: TypedMultiMap[T, K] = new TypedMultiMap[T, K](Map.empty)
```
patriknw: is it possible use a `private val _empty` instance and cast?

akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
```diff
@@ -0,0 +1,68 @@
...
+    map get key match {
+      case None ⇒ this
+      case Some(set) ⇒
+        if (set(value)) {
+          val newset = set - value
+          val newmap = if (newset.isEmpty) map - key else map.updated(key, newset)
+          new TypedMultiMap[T, K](newmap)
+        } else this
+    }
+  }
```
patriknw: perhaps also `def removed(key: T): TypedMultiMap[T, K]`

akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
```diff
@@ -0,0 +1,68 @@
...
+ * Caveat: using keys which take type parameters does not work due to conflicts
+ * with the existential interpretation of `Key[_]`. A workaround is to define
+ * a key type like above and provide a subtype that provides its type parameter
+ * as type member `Type`.
+ */
+class TypedMultiMap[T <: AnyRef, K[_ <: T]] private (map: Map[T, Set[Any]]) {
+
+  def keySet: Set[T] = map.keySet
+
+  def inserted(key: T)(value: K[key.type]): TypedMultiMap[T, K] = {
```
patriknw: double parameter lists because of `key.type`?

rkuhn:
Thanks for the review, Patrik, will go over all the comments soon. All in all, do you think this should live in `akka.util`? I guess we’ll need this in a few places, wherever something like a “service registry” is built, but still I don’t want to simply throw this at the team ;-)

akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
rkuhn: yes, doesn’t work otherwise

akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
rkuhn: So far I only added what I needed, but yes, this looks very useful as well.

akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
rkuhn: yes, certainly

akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
rkuhn: awww, I like those dot-less expressions ;-)

akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
rkuhn: good point

akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
rkuhn: For a multi-map I find it more consistent to use if it always returns a Set, given that emptiness can perfectly well be represented that way already.

akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
rkuhn: I’m dreaming of a world where getOrElse would not need an allocation :-(

patriknw:
I think it is useful, and it is easy enough to understand. 
I have used scala mutable MultiMap and I'm not very happy with that.

akka-actor/src/main/scala/akka/util/TypedMultiMap.scala
patriknw: ok

rkuhn:
PLS BUILD

Issue #16635, opened 2015-01-11
akka-persistence: support multiple configurations


Andrei-Pozolotin:
after a bit of prototyping with akka-persistence we discovered that different types of persistent actors need either different settings for the same plugin, or different plugins altogether.

there is a "promise" of this feature in [Persistence.scala](https://github.com/akka/akka/blob/master/akka-persistence/src/main/scala/akka/persistence/Persistence.scala#L89)

can we have it in 2.4?

ktoso:
Here's the ticket for it: https://github.com/akka/akka/issues/15587
It's very likely to be included as persistence gets stable, as the need for it has been voiced already and we'll also need it for the journal migration tool: https://github.com/akka/akka/issues/16332

Sounds like we can close this ticket as duplicate of #15587?

hepin:
need something like https://github.com/akka/akka/issues/16541 done before.
and then the persistent actor need to provide the plugin id or clazz to select the provider

Andrei-Pozolotin:
lets just do it now for 2.4

please suggest how do you see a) actor -> plugin mapper and b) multi config name space?
using ```persistenceId``` to resolve to plugin/config profile seems convoluted. 
how if we add to ```PersistentActor``` explicit configuration name:
```
/**
* Persistent actor configuration name. 
* Must be located in the following 'reference.conf' section:
* akka {
*    persistence {
*       configuration {
*          name { // User-provided custom configuration name.
*              journal { ... }
*              snapshot-store { ... }
*          }
*       }
*    }
* }
*/
def configuration: String = "" // Empty default for legacy config name space.
```

patriknw:
Please discuss this in the existing ticket, #15587

Issue #15587, opened 2014-07-24
akka persistence: multiple plugins


TimPigden:
It would be useful to allow akka persistence to employ different journal  plugins at the same time. One use case is porting event-source data from one database to another as underlying architecture requirements change.
Another is simply different life-cycles of the event-sourced models, for example when dealing with non-server applications with a defined process cycle for internal data vs some data that is persisted long-term - for example to a backend like EventStore - where the events are separately accessed from other applications.

Andrei-Pozolotin:
after a bit of prototyping with akka-persistence we discovered that different types of persistent actors need either different settings for the same plugin, or different plugins altogether.

there is a "promise" of this feature in Persistence.scala

can we have it in 2.4?

Andrei-Pozolotin:
lets just do it now for 2.4

please suggest how do you see a) actor -> plugin mapper and b) multi config name space?
using persistenceId to resolve to plugin/config profile seems convoluted. 
how if we add to PersistentActor explicit configuration name:

/**
* Persistent actor configuration name. 
* Must be located in the following 'reference.conf' section:
* akka {
*    persistence {
*       configuration {
*          name { // User-provided custom configuration name.
*              journal { ... }
*              snapshot-store { ... }
*          }
*       }
*    }
* }
*/
def configuration: String = "" // Empty default for legacy config name space.

patriknw:
@Andrei-Pozolotin I agree, this would be a great feature.

We can change the signature of the `journalFor(persistenceId: String): ActorRef` method.

I'm thinking that a journal should have an identifying String key, that can be defined by overriding a method in the `PersistentActor`. That key match a configuration section, e.g. `akka.persistence.journal.leveldb`.

We probably need a default value for this key, so that it works as it is today, i.e. using what is defined in config `akka.persistence.journal.plugin`.

Similar for snapshot-store.

Would you be interested in taking a stab at it?

Andrei-Pozolotin:
ok. please move this out of the way #16633

bantonsson:
Fixed by #16794

PR #16794, opened 2015-02-09
+per #15587 Make it possible to use multiple persistence plugins

Andrei-Pozolotin:
- see #15587 #16635

### DIFF
--- a/akka-docs/rst/java/code/docs/persistence/PersistenceMultiDocTest.java
+++ b/akka-docs/rst/java/code/docs/persistence/PersistenceMultiDocTest.java
@@ -1,0 +1,31 @@
+/**
+ * Copyright (C) 2009-2014 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package docs.persistence;
+
+import akka.persistence.UntypedPersistentActor;
+
+public class PersistenceMultiDocTest {
+
+    //#default-plugins
+    abstract class ActorWithDefaultPlugins extends UntypedPersistentActor {
+        @Override
+        public String persistenceId() { return "123"; }
+    }
+    //#default-plugins
+
+    //#override-plugins
+    abstract class ActorWithOverridePlugins extends UntypedPersistentActor {
+        @Override
+        public String persistenceId() { return "123"; }
+        // Absolute path to the journal plugin configuration entry in the `reference.conf`
+        @Override
+        public String journalPluginId() { return "akka.persistence.chronicle.journal"; }
+        // Absolute path to the snapshot store plugin configuration entry in the `reference.conf`
+        @Override
+        public String snapshotPluginId() { return "akka.persistence.chronicle.snapshot-store"; }
+    }
+    //#override-plugins
+
+}
--- a/akka-docs/rst/project/migration-guide-persistence-experimental-2.3.x-2.4.x.rst
+++ b/akka-docs/rst/project/migration-guide-persistence-experimental-2.3.x-2.4.x.rst
@@ -178,8 +178,3 @@ persistent actor on the sending side.
 
 Read more about at-least-once delivery in the :ref:`documentation for Scala <at-least-once-delivery>` and 
 :ref:`documentation for Java <at-least-once-delivery-java>`.  
-
-
-
-
-   
--- a/akka-docs/rst/scala/code/docs/persistence/PersistenceMultiDocSpec.scala
+++ b/akka-docs/rst/scala/code/docs/persistence/PersistenceMultiDocSpec.scala
@@ -1,0 +1,57 @@
+/**
+ * Copyright (C) 2009-2014 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+import akka.persistence.PersistentActor
+
+object PersistenceMultiDocSpec {
+
+  val DefaultConfig = """
+  //#default-config
+  # Absolute path to the default journal plugin configuration entry.
+  akka.persistence.journal.plugin = "akka.persistence.journal.inmem"
+  # Absolute path to the default snapshot store plugin configuration entry.
+  akka.persistence.snapshot-store.plugin = "akka.persistence.snapshot-store.local"
+  //#default-config
+  """
+
+  //#default-plugins
+  trait ActorWithDefaultPlugins extends PersistentActor {
+    override def persistenceId = "123"
+  }
+  //#default-plugins
+
+  val OverrideConfig = """
+  //#override-config
+  # Configuration entry for the custom journal plugin, see `journalPluginId`.
+  akka.persistence.chronicle.journal {
+    # Standard persistence extension property: provider FQCN.
+    class = "akka.persistence.chronicle.ChronicleSyncJournal"
+    # Custom setting specific for the journal `ChronicleSyncJournal`.
+    folder = ${user.dir}/store/journal
+    # Standard persistence extension property: plugin actor uses config injection.
+    inject-config = true
+  } 
+  # Configuration entry for the custom snapshot store plugin, see `snapshotPluginId`.
+  akka.persistence.chronicle.snapshot-store {
+    # Standard persistence extension property: provider FQCN.
+    class = "akka.persistence.chronicle.ChronicleSnapshotStore"
+    # Custom setting specific for the snapshot store `ChronicleSnapshotStore`.
+    folder = ${user.dir}/store/snapshot
+    # Standard persistence extension property: plugin actor uses config injection.
+    inject-config = true
+  }
+  //#override-config
+  """
+
+  //#override-plugins
+  trait ActorWithOverridePlugins extends PersistentActor {
+    override def persistenceId = "123"
+    // Absolute path to the journal plugin configuration entry in the `reference.conf`.
+    override def journalPluginId = "akka.persistence.chronicle.journal"
+    // Absolute path to the snapshot store plugin configuration entry in the `reference.conf`.
+    override def snapshotPluginId = "akka.persistence.chronicle.snapshot-store"
+  }
+  //#override-plugins
+
+}
--- a/akka-persistence/src/main/scala/akka/persistence/Persistence.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/Persistence.scala
@@ -5,13 +5,15 @@
 package akka.persistence
 
 import scala.concurrent.duration._
-
 import com.typesafe.config.Config
-
 import akka.actor._
 import akka.dispatch.Dispatchers
 import akka.persistence.journal.AsyncWriteJournal
 import akka.util.Helpers.ConfigOps
+import akka.event.LoggingAdapter
+import akka.event.Logging
+import java.util.concurrent.atomic.AtomicReference
+import scala.annotation.tailrec
 
 /**
  * Persistence configuration.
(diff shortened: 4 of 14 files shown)


Andrei-Pozolotin:
@patriknw please review

akka-docs/rst/project/migration-guide-persistence-experimental-2.3.x-2.4.x.rst
```diff
@@ -180,6 +180,27 @@ Read more about at-least-once delivery in the :ref:`documentation for Scala <at-
...
+      
+      // Absolute path to the snapshot store configuration entry in the `reference.conf`.
+      override def snapshotPluginId = "akka.persistence.chronicle.snapshot-store"
+      
+    }
 
-
-   
\ No newline at end of file
+Then the actor will be serviced by these specific persistence plugins.
```
patriknw: slightly strange to continue sentence after the sample.

"When persistent actor overrides ``journalPluginId`` and ``snapshotPluginId`` methods the actor will be serviced by these specific persistence plugins instead of the defaults."

akka-docs/rst/project/migration-guide-persistence-experimental-2.3.x-2.4.x.rst
```diff
@@ -180,6 +180,27 @@ Read more about at-least-once delivery in the :ref:`documentation for Scala <at-
...
+    
+      override def persistenceId = "EUR/USD"
+      
+      // Absolute path to the journal configuration entry in the `reference.conf`.
+      override def journalPluginId = "akka.persistence.chronicle.journal"
+      
+      // Absolute path to the snapshot store configuration entry in the `reference.conf`.
+      override def snapshotPluginId = "akka.persistence.chronicle.snapshot-store"
+      
+    }
```
patriknw: code sample should be in compiled source file, and included

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
```diff
@@ -73,65 +73,113 @@ final class PersistenceSettings(config: Config) {
...
   /**
-   * Creates a canonical persistent actor id from a persistent actor ref.
+   * Returns a journal actor identified by `journalPluginId`.
+   * When empty, looks in `akka.persistence.journal.plugin` to find configuration entry path.
+   * When configured, uses `journalPluginId` as absolute path to the journal configuration entry.
+   * Configuration entry must contain few required fields, such as `class`. See `src/main/resources/reference.conf`.
    */
-  def persistenceId(persistentActor: ActorRef): String = id(persistentActor)
+  def journalBy(journalPluginId: String): ActorRef = {
+    val configPath = if (isDefault(journalPluginId)) config.getString("journal.plugin") else journalPluginId
```
patriknw: `config.getString("journal.plugin")` should be cached, i.e. a private val in constructor

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
```diff
@@ -73,65 +73,113 @@ final class PersistenceSettings(config: Config) {
...
+    // TODO remove synchronized after https://issues.scala-lang.org/browse/SI-7943
+    snapshotPlugins.synchronized(snapshotPlugins.getOrElseUpdate(configPath, createPlugin(configPath)(snapshotDispatchSelector)))
   }
 
   /**
    * Returns a journal for a persistent actor identified by `persistenceId`.
+   * When empty, looks in `akka.persistence.journal.plugin` to find journal configuration entry path.
+   * When configured, uses `journalPluginId` declared by the persistent actor, which must be live at the time of invocation.
    */
   def journalFor(persistenceId: String): ActorRef = {
```
patriknw: I think we should add `journalPluginId: String` to this method instead of the `register`, and thereby we don't need the `actorPlugins` map.

akka-persistence/src/main/scala/akka/persistence/PersistentView.scala
```diff
@@ -88,7 +88,10 @@ trait PersistentView extends Actor with Snapshotter with Stash with StashFactory
 
   private val extension = Persistence(context.system)
   private val viewSettings = extension.settings.view
-  private lazy val journal = extension.journalFor(persistenceId)
+
+  // Use same plugins as view origin actor.
+  private[persistence] lazy val journal = extension.journalFor(persistenceId)
```
patriknw: It must be possible to define `journalPluginId` and `snapshotPluginId` for a PersistentView also, in the same way as for a PersistentActor.

This will not work anyway, e.g. when PersistentActor is on one machine and the PersistentView on another machine.

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
```diff
@@ -73,65 +73,113 @@ final class PersistenceSettings(config: Config) {
...
+   * Returns a journal actor identified by `journalPluginId`.
+   * When empty, looks in `akka.persistence.journal.plugin` to find configuration entry path.
+   * When configured, uses `journalPluginId` as absolute path to the journal configuration entry.
+   * Configuration entry must contain few required fields, such as `class`. See `src/main/resources/reference.conf`.
    */
-  def persistenceId(persistentActor: ActorRef): String = id(persistentActor)
+  def journalBy(journalPluginId: String): ActorRef = {
+    val configPath = if (isDefault(journalPluginId)) config.getString("journal.plugin") else journalPluginId
+    // TODO remove synchronized after https://issues.scala-lang.org/browse/SI-7943
+    journalPlugins.synchronized(journalPlugins.getOrElseUpdate(configPath, createPlugin(configPath)(journalDispatchSelector)))
```
patriknw: That is a difficult concurrency problem. The tricky part is to only invoke the creator function once.

We have solved that in the registration of Akka extensions: https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/actor/ActorSystem.scala#L745
That is probably not something we would like to duplicate.

It would be possible to register each journal as an extension, but that would require a `ConcurrentHashMap[String, ExtensionId] here for the mapping between the journalPluginId and ExtensionId instance.
I'm not sure if that is the best solution.

patriknw:
@Andrei-Pozolotin overall this is good, thanks.

patriknw:
OK TO TEST

akka-docs/rst/project/migration-guide-persistence-experimental-2.3.x-2.4.x.rst
Andrei-Pozolotin: applied

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
Andrei-Pozolotin: applied

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
Andrei-Pozolotin: applied

akka-persistence/src/main/scala/akka/persistence/PersistentView.scala
Andrei-Pozolotin: applied

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
Andrei-Pozolotin: I do not follow:
1) do you mean you want non-locking implementation now?
2) or your say that ```synchronized``` is not sufficient?

what I meant to say is: ```synchronized``` is sufficient for now, till SI-7943 is resolved in 2.11.6

Andrei-Pozolotin:
@patriknw please review

Andrei-Pozolotin:
@patriknw "OK TO TEST" seems broken

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
patriknw: `synchronized` is correct, but probably not an acceptable solution. We don't know if and when SI-7943 will be solved. Let's see what other have to say about this.

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
```diff
@@ -87,51 +113,72 @@ object Persistence extends ExtensionId[Persistence] with ExtensionIdProvider {
...
+   * Configuration entry must contain few required fields, such as `class`. See `src/main/resources/reference.conf`.
    */
-  def snapshotStoreFor(persistenceId: String): ActorRef = {
-    // Currently returns a snapshot store singleton but this methods allows for later
-    // optimizations where each persistent actor can have its own snapshot store actor.
-    snapshotStore
+  def journalFor(journalPluginId: String): ActorRef = {
+    val configPath = if (isDefault(journalPluginId)) defaultJournalPluginId else journalPluginId
+    // TODO remove synchronized after https://issues.scala-lang.org/browse/SI-7943
+    journalPlugins.synchronized(journalPlugins.getOrElseUpdate(configPath, createPlugin(configPath)(journalDispatchSelector)))
```
patriknw: repeating my previous comment about this, so that other reviewers can think about it:

`synchronized` might not be an acceptable solution, and I'm not sure we can rely on SI-7943.

That is a difficult concurrency problem. The tricky part is to only invoke the creator function once.

We have solved that in the registration of Akka extensions: https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/actor/ActorSystem.scala#L745
That is probably not something we would like to duplicate.

It would be possible to register each journal as an extension, but that would require a `ConcurrentHashMap[String, ExtensionId] here for the mapping between the journalPluginId and ExtensionId instance.
I'm not sure if that is the best solution.

patriknw:
OK TO TEST

Andrei-Pozolotin:
@patriknw I can not make sense of the build log, can you spot the failure?
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/1910/consoleText

ktoso:
Seems like wrong AKKA_BUILD_JAVA_HOME caused `scripts/build/extra-build-steps.sh` to fail.
Fixed now (back to `/usr/lib/jvm/java-6-openjdk-amd64`, prev setting seems to have been a typo - was m2home).

ktoso:
PLS BUILD

Andrei-Pozolotin:
@ktoso I knew it was you. Next time I am wysyłając depesza natychmiast :-)

Andrei-Pozolotin:
@patriknw do we need more time for review?

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
patriknw: This is how my suggestion of using the extension mechanism to achieve the strict get or create semantics looks like in code:

    case class JournalPlugin(journal: ActorRef) extends Extension

    private var journalPluginExtensionId = new AtomicReference[Map[String, ExtensionId[JournalPlugin]]](Map.empty)

    @tailrec final def journalFor(journalPluginId: String): ActorRef = {
      val configPath = if (isDefault(journalPluginId)) defaultJournalPluginId else journalPluginId
      val extensionIdMap = journalPluginExtensionId.get
      extensionIdMap.get(configPath) match {
        case Some(extensionId) ⇒
          extensionId(system).journal
        case None ⇒
          val extensionId = new ExtensionId[JournalPlugin] {
            override def createExtension(system: ExtendedActorSystem): JournalPlugin =
              JournalPlugin(createPlugin(configPath)(journalDispatchSelector))
          }
          journalPluginExtensionId.compareAndSet(extensionIdMap, extensionIdMap.updated(configPath, extensionId))
          journalFor(journalPluginId) //recursive
      }
    }

patriknw:
@Andrei-Pozolotin it should also be mentioned somewhere in the reference documentation.

We need a few more eyes on this. @ktoso @bantonsson ?

Andrei-Pozolotin:
-------- Original Message --------
Subject: Re: [akka] + akka-persistence: multiple plugins (Patrik
Nordwall6794)
>
> @Andrei-Pozolotin <https://github.com/Andrei-Pozolotin> it should also
> be mentioned somewhere in the reference documentation.
>
I am missing the point: what you mean by "it"?

patriknw:
I mean the feature of being able to use several plugins. How to configure. What to override in the actor to use a different plugin than default.

Andrei-Pozolotin:
-------- Original Message --------
Subject: Re: [akka] + akka-persistence: multiple plugins (Patrik
Nordwall6794)
Date: 2/16/2015, 10:04:56 AM
>
> I mean the feature of being able to use several plugins.
>
applied

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
Andrei-Pozolotin: do you have a failing test case somewhere to prove that ```synchronized``` is not enough?

patriknw: For performance reasons I would like to avoid `synchronized` that is used every time a persistent actor is started. Probably not a big deal, but normally we try to keep the akka internals free from such things.
We can ask for second opinion from @rkuhn or @viktorklang.

viktorklang: I'd advise to use the same mechanism that is used for Extensions, its battle tested.

Andrei-Pozolotin: -------- Original Message --------
Subject: Re: [akka] + akka-persistence: multiple plugins (Viktor Klang
(√)6794)
> I'd advise to use the same mechanism that is used for Extensions, its
> battle tested
applied

Andrei-Pozolotin:
@ktoso please restart
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/1948/console
```
+ git fetch origin +pull/16794/head:pullreq/16794
fatal: unable to connect to github.com:
github.com[0: 192.30.252.131]: errno=Connection timed out
```

bantonsson:
PLS BUILD

patriknw:
LGTM

Andrei-Pozolotin:
-------- Original Message --------
Subject: Re: [akka] + akka-persistence: multiple plugins (Patrik
Nordwall6794)
>
> LGTM
>
ready for squash or still more review?

patriknw:
We need some more LGTM before we can merge. A squash would not hurt at this moment.

When you squash it would be great if you can amend the commit comment to follow the [conventions](https://github.com/akka/akka/blob/master/CONTRIBUTING.md).

In this case it would be something like:

    +per #15587 Make it possible to use multiple persistence plugins

Andrei-Pozolotin:
-------- Original Message --------
Subject: Re: [akka] + akka-persistence: multiple plugins (Patrik
Nordwall6794)
>
> squash would not hurt at this moment
>
applied

Andrei-Pozolotin:
@ktoso @bantonsson please review

rkuhn:
Sorry for the radio silence, we are crunching to get streams 1.0-M4 out, will review soon!

Andrei-Pozolotin:
good to hear. we started to worry uppsala is running low on vitamin c :-)

akka-docs/rst/java/code/docs/persistence/PersistenceMultiDocTest.java
```diff
@@ -0,0 +1,31 @@
+/**
+ * Copyright (C) 2009-2014 Typesafe Inc. <http://www.typesafe.com>
```
bantonsson: It's 2015 :wink:

akka-docs/rst/scala/code/docs/persistence/PersistenceMultiDocSpec.scala
```diff
@@ -0,0 +1,57 @@
+/**
+ * Copyright (C) 2009-2014 Typesafe Inc. <http://www.typesafe.com>
```
bantonsson: 2015

bantonsson:
LGTM!

Thanks a lot @Andrei-Pozolotin

akka-docs/rst/java/code/docs/persistence/PersistenceMultiDocTest.java
Andrei-Pozolotin: Yeah, that is because it takes forever to get a review :-)

akka-docs/rst/scala/code/docs/persistence/PersistenceMultiDocSpec.scala
Andrei-Pozolotin: Is it just for new files of for all of the files I touched? 
I would imagine it has to be global akka find/replace, not peace-meal per-project approach?

rkuhn: Indeed, we should do a sweep, you don’t need to worry about this one.

bantonsson: @rkuhn Maybe even automate it? https://github.com/sbt/sbt-header

rkuhn: Yes, looks good; could you look into that? Thanks!

Andrei-Pozolotin:
@patriknw is 2 X LGTM good enough?

ktoso:
Looks great, LGTM, thanks for your work on this!

Andrei-Pozolotin:
thank you guys

Issue #16787, opened 2015-02-06
Convert Cookbook to Java


drewhk:

ktoso:
I'll start with a few here - picking up recipies from the tip.

patriknw:
@ktoso isn't this done?

ktoso:
No, Sadlly the APIs changed so much, that I was unable to update my PR last week for M4.
The most recent is https://github.com/drewhk/akka/pull/22 but doesnt compile yet for M4.

PR #16850, opened 2015-02-11
=str #16787 java cookbook

ktoso:
Missing a few recipes, I'm adding them now.


ktoso:
Pushed fixup of mock `System.out.println()` which caused test failures. Ready to be reviewed now.

ktoso:
refs #16787

ktoso:
failure was pdf generation - file include was wrong in one spot `s/seq/list/`

akka-actor/src/main/scala/akka/util/ByteString.scala
```diff
@@ -53,6 +54,14 @@ object ByteString {
   def fromArray(array: Array[Byte]): ByteString = apply(array)
 
   /**
+   * JAVA API
+   * Creates a new ByteString by copying an int array by converting from integral numbers to bytes.
+   */
+  @varargs
+  def fromInts(array: Int*): ByteString =
```
patriknw: this should not be changed here, but in release-2.3 branch

ktoso: ah, right - thanks!

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeGlobalRateLimit.java
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  final FlowMaterializer mat = ActorFlowMaterializer.create(system);
+
+  static
+  //#global-limiter-actor
+  public class Limiter extends AbstractActor {
+
+    public static class WantToPass { }
+    public static final WantToPass WANT_TO_PASS = new WantToPass();
```
patriknw: Is WantToPass intended to be a singleton instance? Is it possible to use something more idiomatic, such as:

     public static class WantToPass {
         private WantToPass() {}
         public static final WantToPass INSTANCE = new WantToPass();
     }

ktoso: Yes, I think this style caught my eye in some other spec. I'll move to INSTANCE

akka-stream/src/main/scala/akka/stream/javadsl/OperationAttributes.scala
```diff
@@ -50,7 +50,9 @@ object OperationAttributes {
     private[akka] def asScala = scaladsl.OperationAttributes.dispatcher(dispatcher)
   }
 
-  private[akka] val none: OperationAttributes = new OperationAttributes {
+  def none() = _none
+
+  val _none: OperationAttributes = new OperationAttributes {
```
patriknw: why was this needed?

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeGlobalRateLimit.java
patriknw: ok, thanks 
(several other places)

akka-stream/src/main/scala/akka/stream/javadsl/OperationAttributes.scala
ktoso: Had to check up on this (didn't remember).
It's not really *needed*, but thought the reusing the none instance may be useful - shall I revert?

ktoso: Had to check up on this - not really needed, I'll revert!

ktoso: Oh, and about making it public - I think it's useful for end users to have access to none sometimes, in code like `if (x) none else name(...)` etc.

ktoso:
Moved over to https://github.com/drewhk/akka/pull/22, applied Patriks comments.

PR #16867, opened 2015-02-12
=rem #xxxx use a Map watchee -> watchers in RemoteWatcher

trobert:
This commit replace the set of  (Watchee, Watchers) by a Map of Watchee -> Set of Watchers (a multimap actually)
This aims to remove whole iteration on all watches in most cases.

Some more commit should follow with other optimizations, especially for 1 watchee, N watchers case.

### DIFF
--- a/akka-cluster/src/main/scala/akka/cluster/ClusterRemoteWatcher.scala
+++ b/akka-cluster/src/main/scala/akka/cluster/ClusterRemoteWatcher.scala
@@ -102,10 +102,8 @@ private[cluster] class ClusterRemoteWatcher(
    * by super RemoteWatcher.
    */
   def takeOverResponsibility(address: Address): Unit = {
-    watching foreach {
-      case (watchee, watcher) ⇒ if (watchee.path.address == address)
-        unwatchRemote(watchee, watcher)
-    }
+    watching.keys.withFilter(_.path.address == address).foreach(clearAllWatches)
+    unwatchNode(address)
   }
 
 }
--- a/akka-cluster/src/main/scala/akka/cluster/ClusterRemoteWatcher.scala
+++ b/akka-cluster/src/main/scala/akka/cluster/ClusterRemoteWatcher.scala
@@ -36,7 +36,7 @@ private[cluster] object ClusterRemoteWatcher {
  *
  * Specialization of [[akka.remote.RemoteWatcher]] that keeps
  * track of cluster member nodes and is responsible for watchees on cluster nodes.
- * [[akka.actor.AddressTerminate]] is published when node is removed from cluster.
+ * [[akka.actor.AddressTerminated]] is published when node is removed from cluster.
  *
  * `RemoteWatcher` handles non-cluster nodes. `ClusterRemoteWatcher` will take
  * over responsibility from `RemoteWatcher` if a watch is added before a node is member
@@ -52,8 +52,6 @@ private[cluster] class ClusterRemoteWatcher(
     heartbeatInterval,
     unreachableReaperInterval,
     heartbeatExpectedResponseAfter) {
-
-  import RemoteWatcher._
 
   val cluster = Cluster(context.system)
   import cluster.selfAddress
@@ -73,8 +71,6 @@ private[cluster] class ClusterRemoteWatcher(
   override def receive = receiveClusterEvent orElse super.receive
 
   def receiveClusterEvent: Actor.Receive = {
-    case WatchRemote(watchee, watcher) if clusterNodes(watchee.path.address) ⇒
-      () // cluster managed node, don't propagate to super
     case state: CurrentClusterState ⇒
       clusterNodes = state.members.collect { case m if m.address != selfAddress ⇒ m.address }
       clusterNodes foreach takeOverResponsibility
@@ -96,14 +92,18 @@ private[cluster] class ClusterRemoteWatcher(
     case _: MemberEvent ⇒ // not interesting
   }
 
+  override def watchNode(watcheeAddress: Address) =
+    if (!clusterNodes(watcheeAddress)) super.watchNode(watcheeAddress)
+
   /**
    * When a cluster node is added this class takes over the
    * responsibility for watchees on that node already handled
    * by super RemoteWatcher.
    */
-  def takeOverResponsibility(address: Address): Unit = {
-    watching.keys.withFilter(_.path.address == address).foreach(clearAllWatches)
-    unwatchNode(address)
-  }
+  def takeOverResponsibility(address: Address): Unit =
+    if (watchingNodes.contains(address)) {
+      log.debug("Cluster is taking over responsibility of node: [{}]", address)
+      unwatchNode(address)
+    }
 
 }
--- a/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala
+++ b/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala
@@ -8,14 +8,12 @@ import akka.actor._
 import akka.dispatch.sysmsg._
 import akka.event.{ Logging, LoggingAdapter, EventStream }
 import akka.event.Logging.Error
-import akka.serialization.{ JavaSerializer, Serialization, SerializationExtension }
+import akka.serialization.{ Serialization, SerializationExtension }
 import akka.pattern.pipe
 import scala.util.control.NonFatal
 import akka.actor.SystemGuardian.{ TerminationHookDone, TerminationHook, RegisterTerminationHook }
 import scala.util.control.Exception.Catcher
-import scala.concurrent.{ ExecutionContext, Future }
-import scala.concurrent.forkjoin.ThreadLocalRandom
-import com.typesafe.config.Config
+import scala.concurrent.Future
 import akka.ConfigurationException
 import akka.dispatch.{ RequiresMessageQueue, UnboundedMessageQueueSemantics }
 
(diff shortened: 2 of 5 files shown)


patriknw:
@trobert thanks, I will take a look, but I can say right away that I don't think we will do this change in 2.3.x, so you should open another pull request targeted to master

akka-cluster/src/main/scala/akka/cluster/ClusterRemoteWatcher.scala
```diff
@@ -102,10 +102,8 @@ private[cluster] class ClusterRemoteWatcher(
    * by super RemoteWatcher.
    */
   def takeOverResponsibility(address: Address): Unit = {
-    watching foreach {
-      case (watchee, watcher) ⇒ if (watchee.path.address == address)
-        unwatchRemote(watchee, watcher)
-    }
+    watching.keys.filter(_.path.address == address).foreach(clearAllWatches)
```
patriknw: this loops twice, the filter can be moved inside the foreach as an if statement

patriknw: it's mostly creation of the intermediate collection that is the problem

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
```diff
@@ -193,11 +193,11 @@ private[akka] class RemoteWatcher(
 
       // also watch from self, to be able to cleanup on termination of the watchee
       context watch watchee
-      watching += ((watchee, self))
+      watching += watchee → watching.get(watchee).fold(Set(self)) { _ + self }
```
patriknw: slightly more efficient to use `get(watchee) match {`

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
```diff
@@ -287,15 +296,15 @@ private[akka] class RemoteWatcher(
...
-        if (wee.path.address == address) {
-          // this re-watch will result in a RewatchRemote message to this actor
-          // must be a special message to be able to detect if an UnwatchRemote comes in
-          // before the extra RewatchRemote, then the re-watch should be ignored
-          log.debug("Re-watch [{} -> {}]", wer, wee)
-          wee.sendSystemMessage(new Rewatch(wee, wer)) // ➡➡➡ NEVER SEND THE SAME SYSTEM MESSAGE OBJECT TO TWO ACTORS ⬅⬅⬅
-        }
+    for {
+      (wee: InternalActorRef, wers) ← watching if wee.path.address == address
+      wer ← wers if wer.isInstanceOf[InternalActorRef]
```
patriknw: I would make that instanceOf into a `match` inside the next block, then isInstanceOf-asInstanceOf is not needed

akka-cluster/src/main/scala/akka/cluster/ClusterRemoteWatcher.scala
patriknw: also, I'm not sure, but I think it's more efficient to use `foreach { case (key, value)` than `keys`, because `keys` must create some kind of intermediate collection

patriknw:
@trobert I like this, this part is not as scary as I first thought
anyway, open pull request to master, and we can decide later what/if backporting should take place

trobert:
Thanks for the review!

I pushed another commit (the 1->N case), if you can have a look.

This a bit more intrusive, but the unit tests (and also my own testing) seems to be ok

I'll rebase against master and take your remark into account asap (probably next week). Is there any indicative schedule for akka 2.4 ?

akka-cluster/src/main/scala/akka/cluster/ClusterRemoteWatcher.scala
trobert: keys is lightweight, it creates a wrapper around the map. You're right for filter however.
I propose a `for (watchee <- watching.keys`) with a guard condition (efficient and I find this a bit  more readable at the same time)

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
trobert: Not sure about speed argument. `Option.fold` is marked inline and only consist of one `if`
But I change to a `match`, it is more common usage

akka-cluster/src/main/scala/akka/cluster/ClusterRemoteWatcher.scala
ktoso: Or a plain `watching.keys.filterWith(_.path.address == address).foreach(clearAllWatches)` would jsut as well (exactly same as forloop).

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
trobert: I propose

```
for {
      (wee, wers) ← watching if wee.path.address == address
      wer ← wers
      (watchee: InternalActorRef, watcher: InternalActorRef) = (wee, wer)
    }
```

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
```diff
@@ -287,15 +287,12 @@ private[akka] class RemoteWatcher(
...
-        if (wee.path.address == address) {
-          // this re-watch will result in a RewatchRemote message to this actor
-          // must be a special message to be able to detect if an UnwatchRemote comes in
-          // before the extra RewatchRemote, then the re-watch should be ignored
-          log.debug("Re-watch [{} -> {}]", wer, wee)
-          wee.sendSystemMessage(new Rewatch(wee, wer)) // ➡➡➡ NEVER SEND THE SAME SYSTEM MESSAGE OBJECT TO TWO ACTORS ⬅⬅⬅
-        }
+    for {
+      wee ← watching.keys if wee.path.address == address
+      (watchee: InternalActorRef, watcher: InternalActorRef) = (wee, self)
```
ktoso: I'd avoid this pattern in performance sensitive code (`(a, b) = (x, y)`) - this allocates a Tuple2 each time (this may actually be the case where 2 Tuple2 allocations happen for each such line actually if I remember correctly).
A `match` inside would be allocation free and thus faster. :-)

patriknw:
The second commit will require careful consideration, I will not be able to swap in that right now.
Thanks.

akka-cluster/src/main/scala/akka/cluster/ClusterRemoteWatcher.scala
```diff
@@ -36,7 +36,7 @@ private[cluster] object ClusterRemoteWatcher {
  *
  * Specialization of [[akka.remote.RemoteWatcher]] that keeps
  * track of cluster member nodes and is responsible for watchees on cluster nodes.
- * [[akka.actor.AddressTerminate]] is published when node is removed from cluster.
+ * [[akka.actor.AddressTerminated]] is published when node is removed from cluster.
```
drewhk: nice!

akka-cluster/src/main/scala/akka/cluster/ClusterRemoteWatcher.scala
```diff
@@ -73,8 +71,6 @@ private[cluster] class ClusterRemoteWatcher(
   override def receive = receiveClusterEvent orElse super.receive
 
   def receiveClusterEvent: Actor.Receive = {
-    case WatchRemote(watchee, watcher) if clusterNodes(watchee.path.address) ⇒
```
drewhk: hm, why is this removed?

akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala
```diff
@@ -477,8 +461,13 @@ private[akka] class RemoteActorRef private[akka] (
 
   def sendSystemMessage(message: SystemMessage): Unit =
     try {
-      remote.send(message, None, this)
-      provider.afterSendSystemMessage(message)
+      message match {
+        case Watch(watchee, watcher) if watcher != provider.remoteWatcher ⇒
+          provider.remoteWatcher ! RemoteWatcher.WatchRemote(watchee, watcher)
```
drewhk: wait, this change will now no longer send the Watch to the remote side

drewhk:
OK TO TEST

akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala
trobert: yes, this is the idea of the change: only watches from the remoteWatcher are sent to the remote side

akka-cluster/src/main/scala/akka/cluster/ClusterRemoteWatcher.scala
trobert: `WatchRemote` message must be handled by the remote watcher for updating its watching map. This is necessary because remote watcher has now the responsibility to propagate the `DeathWatchNotifcation` to all watchers.

The heartbeat process will be bypassed for cluster members thanks to the override of the `watchNode` method

akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala
drewhk: so you say that remoteWatcher will send the Watch system message instead?

trobert: Exactly. this is to improve the case where different local actors are watching the same remote actor.
See discussion here: https://groups.google.com/d/msg/akka-dev/ao6MLKvxJ0o/INMBcgJ7PXMJ

trobert:
The failure is linked to binary incompatibility. I'm opening a new PR against master as @patriknw suggested.

patriknw:
yes, please

patriknw:
Superseed by ##16911

Issue #16620, opened 2015-01-07
Report cause of bind and connect failures


richdougherty:
When a bind fails in Play we'd like to tell users the reason for the failure. At the moment Akka HTTP gives an `akka.stream.BindFailedException` with a fixed error message. This makes it difficult to debug bind failures without enabling logging. It would be useful to include the error reason or cause in the exception so that it is not necessary to enable logging. I realise this may require changes to the Akka IO protocol so that Akka IO can return error information to Akka HTTP.

At the moment Play only cares about bind errors, but in the future we will use Akka HTTP for outgoing messages, so it would be good to make sure that Akka HTTP returns error information for outgoing connection failures too.

patriknw:
cc @sirthias @jrudolph

jrudolph:
See #13861.

jrudolph:
And there's also #16425 and #16507 filed for streams and http.

richdougherty:
So I'm not the only one. :)

drewhk:
The cause reporting needs to be first fixed at the IO level. Retargeting this ticket to 2.3.x. If this cannot be done in a binary compatible fashion this needs to be retargetad at 2.4-M1

richdougherty:
Makes sense. Play's Akka HTTP support will be part of Play 2.4 and Play 2.4 will target Akka 2.4.

Apart from giving users better errors, we want better error information because we occasionally see a class of bind errors that we haven't been able to diagnose, even with logging turned on. We have been able to fix "normal" bind errors, like when the address is already in use, but there are other bind errors that we don't understand yet. The errors could be caused by a bug in Akka, or they could be a bug in Play's Akka backend or its test code. I'm not sure yet. Once we get better error reporting hopefully we can diagnose it properly.

rkuhn:
While reviewing @skisel’s PR I realized that we will likely not be able to perform the proposed changes in the 2.x timeframe (since we want to retain binary compatibility). Another line of reasoning is that making IO exceptions part of the API sounds like a bad idea because these exceptions are so ill-specified and unreliable in their interpretation (O/S-, version- and locale-dependent).

Hence I am (strongly) leaning in the direction of solving this via logging only—we need to make sure that every failure case does lead to a log event and that gives users the visibility they need without contaminating our API.

bantonsson:
Closed by #16895

richdougherty:
[...]

PR #16878, opened 2015-02-15
#16620 Report cause of bind and connect failures

skisel:
I have added Option[Throwable] into CommandFailed message, and passed exception where it was raised (including binding failure). It seems its breaking-API change, since I was not able to compile and run akka-stream-tests-experimental locally. I would appreciate if you could explain me how to do it right. Also, I'm not sure what is the right branch to base my change onto, please assist. Thank you!

### DIFF
--- a/akka-actor/src/main/scala/akka/io/Tcp.scala
+++ b/akka-actor/src/main/scala/akka/io/Tcp.scala
@@ -97,7 +97,7 @@ object Tcp extends ExtensionId[TcpExt] with ExtensionIdProvider {
    * This is the common trait for all commands understood by TCP actors.
    */
   trait Command extends Message with SelectionHandler.HasFailureMessage {
-    def failureMessage = CommandFailed(this)
+    def failureMessage(cause: Option[Throwable] = None) = CommandFailed(this, cause)
   }
 
   /**
@@ -425,7 +425,7 @@ object Tcp extends ExtensionId[TcpExt] with ExtensionIdProvider {
    * Whenever a command cannot be completed, the queried actor will reply with
    * this message, wrapping the original command which failed.
    */
-  case class CommandFailed(cmd: Command) extends Event
+  case class CommandFailed(cmd: Command, cause: Option[Throwable] = None) extends Event
 
   /**
    * When `useResumeWriting` is in effect as indicated in the [[Register]] message,
--- a/akka-actor/src/main/scala/akka/io/TcpListener.scala
+++ b/akka-actor/src/main/scala/akka/io/TcpListener.scala
@@ -19,7 +19,7 @@ import akka.dispatch.{ UnboundedMessageQueueSemantics, RequiresMessageQueue }
 private[io] object TcpListener {
 
   case class RegisterIncoming(channel: SocketChannel) extends HasFailureMessage with NoSerializationVerificationNeeded {
-    def failureMessage = FailedRegisterIncoming(channel)
+    def failureMessage(cause: Option[Throwable] = None) = FailedRegisterIncoming(channel)
   }
 
   case class FailedRegisterIncoming(channel: SocketChannel) extends NoSerializationVerificationNeeded
@@ -60,7 +60,7 @@ private[io] class TcpListener(selectorRouter: ActorRef,
       ret
     } catch {
       case NonFatal(e) ⇒
-        bindCommander ! bind.failureMessage
+        bindCommander ! bind.failureMessage(Some(e))
         log.debug("Bind failed for TCP channel on endpoint [{}]: {}", bind.localAddress, e)
         context.stop(self)
     }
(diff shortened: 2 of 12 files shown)


skisel:
Forgot to mention that this PR is to fix #16620

akka-actor/src/main/scala/akka/io/Tcp.scala
```diff
@@ -425,7 +425,7 @@ object Tcp extends ExtensionId[TcpExt] with ExtensionIdProvider {
    * Whenever a command cannot be completed, the queried actor will reply with
    * this message, wrapping the original command which failed.
    */
-  case class CommandFailed(cmd: Command) extends Event
+  case class CommandFailed(cmd: Command, cause: Option[Throwable] = None) extends Event
```
rkuhn: Unfortunately it basically impossible to extend a case class in a binary compatible fashion (meaning: I cannot prove its impossibility but I have never seen a working proposal). Therefore we cannot do this in the 2.3.x series (which current Streams work is based upon).

rkuhn: On a related note: Akka 2.4.x is supposed to be as binary compatible to 2.3.x as we can make it, so I’m unsure about how to proceed here.

akka-actor/src/main/scala/akka/io/TcpListener.scala
```diff
@@ -60,7 +60,7 @@ private[io] class TcpListener(selectorRouter: ActorRef,
       ret
     } catch {
       case NonFatal(e) ⇒
-        bindCommander ! bind.failureMessage
+        bindCommander ! bind.failureMessage(Some(e))
```
rkuhn: The downside of exposing these exceptions is that due to a design deficiency in NIO they are platform dependent (i.e. which precise subtype of IOException is used for which purpose and with which message varies according to operating system and version). This means that we offer an API that is not reliable. We could try to lift it up by knowing all the quirks and interpreting the error messages, turning the real exception into something that makes more sense, but that would be a lot of work (in particular for those systems that we do not have readily available for testing).

akka-stream/src/main/scala/akka/stream/impl/io/TcpConnectionStream.scala
```diff
@@ -278,6 +279,11 @@ private[akka] class OutboundTcpStreamActor(processorPromise: Promise[Processor[B
       processorPromise.success(exposedProcessor)
       initSteps.become(Actor.emptyBehavior)
 
+    case CommandFailed(_, Some(ex)) ⇒
+      localAddressPromise.failure(ex)
+      processorPromise.failure(ex)
+      fail(ex)
```
rkuhn: Instead of repeating the implementation, I’d extract the `Option[Throwable]` and use `.getOrElse(new StreamTcpException(...))` to get the exception out.

akka-stream/src/main/scala/akka/stream/impl/io/TcpListenStreamActor.scala
```diff
@@ -86,6 +86,11 @@ private[akka] class TcpListenStreamActor(localAddressPromise: Promise[InetSocket
         unbindPromise.success(() ⇒ { target ! Unbind; unboundPromise.future })
         primaryOutputs.getExposedPublisher.subscribe(flowSubscriber.asInstanceOf[Subscriber[Any]])
         subreceive.become(running)
+      case CommandFailed(_, Some(ex)) ⇒
+        localAddressPromise.failure(ex)
+        unbindPromise.failure(ex)
+        flowSubscriber.onError(ex)
+        fail(ex)
```
rkuhn: Here it is particularly important to use the other implementation since it contains a fix for Reactive Streams compliance (the `tryOnError` part).

akka-stream/src/main/scala/akka/stream/impl/io/TcpListenStreamActor.scala
```diff
@@ -98,6 +103,9 @@ private[akka] class TcpListenStreamActor(localAddressPromise: Promise[InetSocket
       case c: Connected ⇒
         pendingConnection = (c, sender())
         pump()
+      case CommandFailed(_, Some(ex)) ⇒
+        unbindPromise.tryFailure(ex)
```
rkuhn: the `tryFailure` below is conditional, so here reuse would probably be better.

rkuhn:
@skisel Thanks for the contribution! Unfortunately you picked one of the “hairy” tickets, where it is not even obvious whether the fix should be performed at all.

An alternative solution to the original problem—namely that not enough information is available to users in order to debug failures—could be to continue sending the anonymous bind failure message but additionally log the full exception. That way the unreliable (locale dependent) information does not become part of the API, which means that less people will be bitten by relying upon brittle details.

skisel:
Thanks for the review, @rkuhn! Very useful comments. 
I can pick up logging as well, the only question will be - what branch should be the base for this change?

rkuhn:
You should target the `release-2.3` branch in order to let Streams benefit (which I think is your intention). The process is that at some point we will release 2.3.10 (or 2.3.11 if your PR doesn’t make it in time) and then the next Streams release will make use of it.

rkuhn:
new efforts in #16895

Issue #16505, opened 2014-12-10
BUG: Remoting sends AddressTerminated a bit too eagerly


bantonsson:
The remoting code that gates an address when there is a connection failure also eagerly sends an AddressTerminated event which it shouldn't. This can lead to remote death watch firing even though the node has only been Gated and not Quarantined, and the node subsequently coming back.

drewhk:
Hey, this was not forward ported to master

drewhk:
forward ported

PR #16884, opened 2015-02-17
=rem #16505: Do not publish AddressTerminated (missing fwd port)

drewhk:
- also not throw IllegalAssocEx from transport
- added a test for stashing behavior
  (cherry picked from commit b7295a8)

### DIFF
--- a/akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterShardingSpec.scala
+++ b/akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterShardingSpec.scala
@@ -315,6 +315,26 @@ class ClusterShardingSpec extends MultiNodeSpec(ClusterShardingSpec) with STMult
       enterBarrier("after-4")
     }
 
+    "support proxy only mode" in within(10.seconds) {
+      runOn(second) {
+        val proxy = system.actorOf(ShardRegion.proxyProps(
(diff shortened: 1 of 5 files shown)


drewhk:
This commit was not forward ported. I think this was a plain oversight but someone can comment if there was some real reason.

drewhk:
Hmm, a series of Sharding related failures. It seems unlikely that my changes are the issue, so let's retry.

drewhk:
PLS BUILD

drewhk:
now a connection timeout to GH... sigh

drewhk:
PLS BUILD

drewhk:
Ok, so sharding really fails with these patches. Interesting.

patriknw:
Same thing locally, cluster sharding has changed rather much in master.
I will take a look at it.

patriknw:
@drewhk it's a timing issue. Terminated is now triggered later when trying to talk to a shutdown node. In this case node `first` is crashed in an earlier step, but that has not been triggered when coming to the step that now fails.

The fix is to simply move "support proxy only mode" up, and place it after "support passivation and activation of entries". Then it must also `runOn(second)` instead of `sixth`.

Can you fix it here, or do you want me to submit a separate PR?

drewhk:
I can fix it here but probably not today.

PR #16891, opened 2015-02-17
Server-side Websocket API preview (Do not merge)

jrudolph:
A first preview of a possible design of the websocket API on the server side (no implementation yet).

The basic procedure would be as follows:
- HttpServer receives WebSocket upgrade request
- A special header `UpgradeToWebsocket` header is created and added to a barebones request that is sent to the user-handler
- If the user-handler wants to answer websocket requests at the given URI it needs to look for that header and call one of its methods (supplying the websocket handler) to create the upgrade response.
- The special response triggers the update response to be sent and at the same time the connection will be switched over to a websocket one (using a FlexiRoute/FlexiMerge around the usual HttpServer pipeline)

The Websocket API itself consists of two layers:
- high-level message based API: the behavior is defined as a `Flow[Message, Message]`, framing and dealing with control-frames is done automatically
- low-level frame based handler: the user needs to handle control frames and data frames yourself (answer or create ping requests, deal with fragmentation). This could also be implementation private.

Several questions come up:
- Messages can be fragmented into several frames. This allows interleaving control frames and data frames. However, it doesn't allow multiplexing. I.e. it makes sense to model one side of a websocket connection as a `Source[Message]` where each `Message` itself may contain a `Source[ByteString]`. Because of missing multiplexing there's a real danger of head-of-line blocking on the `Message` stream. This is by design and we probably don't need to do anything here because it is somewhat handled by reactive-stream backpressure automatically.
- Each message fragment (= data frame) can have a size of 2^64 bytes. This leads to the question if clients will produce such big frames. If we expect biggish frames it makes sense to dispatch the data of partially-received frames as soon as the data arrives. Unfortunately, this leads to something like the current HTTP pipeline design with the `splitWhen`/`headAndTail` combo which seems like a workaround and is the main reason of the bad performance of akka-http.
- Instead of using the special header, an alternative `HttpRequest/Response` subtype could be created to represent a WebSocket request/response. The header based design, though, seems less intrusive and more true to the spirit of HTTP where most extension functionality is also carried by headers. We could also piggy-back on top of one of the (to-be-written) WebSocket specific headers to expose the functionality of `UpgradeToWebsocket` to users.

/cc @sirthias

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/engine/ws/Frame.scala
+++ b/akka-http-core/src/main/scala/akka/http/engine/ws/Frame.scala
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.engine.ws
+
+import akka.util.ByteString
+
+/**
+ * http://tools.ietf.org/html/rfc6455
+ *
+ * 0                   1                   2                   3
+ * 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+ * +-+-+-+-+-------+-+-------------+-------------------------------+
+ * |F|R|R|R| opcode|M| Payload len |    Extended payload length    |
+ * |I|S|S|S|  (4)  |A|     (7)     |             (16/64)           |
+ * |N|V|V|V|       |S|             |   (if payload len==126/127)   |
+ * | |1|2|3|       |K|             |                               |
+ * +-+-+-+-+-------+-+-------------+ - - - - - - - - - - - - - - - +
+ * |     Extended payload length continued, if payload len == 127  |
+ * + - - - - - - - - - - - - - - - +-------------------------------+
+ * |                               |Masking-key, if MASK set to 1  |
+ * +-------------------------------+-------------------------------+
+ * | Masking-key (continued)       |          Payload Data         |
+ * +-------------------------------- - - - - - - - - - - - - - - - +
+ * :                     Payload Data continued ...                :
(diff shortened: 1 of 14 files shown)


sirthias:
I'd prefer the special-header based design (as described) over the "special HttpRequest/response sub-type". The basic approach sounds good!

akka-http-core/src/main/scala/akka/http/engine/ws/Frame.scala
```diff
@@ -0,0 +1,44 @@
...
+package akka.http.engine.ws
+
+import akka.util.ByteString
+
+/**
+ * http://tools.ietf.org/html/rfc6455
+ *
+ * 0                   1                   2                   3
+ * 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+ * +-+-+-+-+-------+-+-------------+-------------------------------+
```
drewhk: thanks for including this layout

drewhk:
I think it would be useful to have a thin API layer in top that fully strictifies Messages up to a maximum size (and fails if that maximum is violated). Use case: there are many applications that use websockets for small, interactive messaging, and for them the added risk of deadlocks by having stream-of-streams does not worth the added cognitive overhead. I guess this can be easily added on top of the lower level API, but I think it would be useful to provide it out of the box.

rkuhn:
Looking great, I also agree with the reasoning of how to present things. The Flow adaptations in WebSockets will then (soon) be changed into BidiFlows of course ;-)

rkuhn:
I guess this can be closed, and you’ll open a new one once ready?

jrudolph:
Yes, I'll open a new one.

PR #16892, opened 2015-02-17
+hco #16822 add model for `Referer` header

### LINKED ISSUE #16822: Add `Referer` header model
Also for the Java side.

/cc @sirthias

sirthias:
Closes #16822.

Could also serve as a stand-alone example of how to add a new header model class.


rkuhn:
PLS BUILD

akka-http-core/src/main/scala/akka/http/model/parser/HeaderParser.scala
```diff
@@ -99,7 +99,7 @@ object HeaderParser {
     "range",
     "proxy-authenticate",
     "proxy-authorization",
-    "range",
```
rkuhn: this seems misplaced, should this be removed within this PR?

rkuhn:
LGTM apart from the one question above

akka-http-core/src/main/scala/akka/http/model/parser/HeaderParser.scala
2beaucoup: It's a duplication from 3 lines above. Maybe delete the other one to keep the sorting?

sirthias: Yes, right.
The sorting is not actually a requirement here but I wanted to uphold it for clearity, but failed.
Will fix.

sirthias:
Fixed the small ordering problem in `HeaderParser.scala`.

akka-http-core/src/test/scala/akka/http/engine/parsing/HttpHeaderParserSpec.scala
```diff
@@ -153,7 +153,8 @@ class HttpHeaderParserSpec extends WordSpec with Matchers with BeforeAndAfterAll
           |   |   ┌─o-r-i-g-i-n-:- (origin)
           |   |   |                   ┌─e-n-t-i-c-a-t-e-:- (proxy-authenticate)
           |   | ┌─p-r-o-x-y---a-u-t-h-o-r-i-z-a-t-i-o-n-:- (proxy-authorization)
-          |   | | └─r-a-n-g-e-:- (range)
+          |   | | |   ┌─a-n-g-e-:- (range)
+          |   | | └─r-e-f-e-r-e-r-:- (referer)
```
jrudolph: I wonder if we shouldn't either 1) test only invariants of the tree (as far as we can identify important ones) or 2) disable these tests. 

I would for now suggest to disable these tests as AFAIK they never caught an error so far but were a maintenance burden.

jrudolph:
LGTM

akka-http-core/src/test/scala/akka/http/engine/parsing/HttpHeaderParserSpec.scala
rkuhn: I concur.

sirthias: Yup. I'll keep them in but disable them.
They do serve the purpose of providing a nice test bed for the fast head parser / cache implementation, which does exhibit some size and structure restrictions.
However, since the code has proven quite resilient to growth in the header model  (i.e. keeps working as expected)
over the course of the last two years that we now have this code in, disabling these tests should be fine.

sirthias:
Added a commit disabling the header-model-sensitive tests as discussed.

2beaucoup:
Mmh, what's the usage of these disabled tests if they are most likely invalid when reenabled? Why not just add a main method that prints out the report?

sirthias:
Yes, right. Simply disabling was the easier thing to do, but you are right, moving the stuff out of the tests completely is better. Will do. Thanks!

sirthias:
Ok, done.

2beaucoup:
:+1:

rkuhn:
Nice one, should be noted somewhere as an example for a new header.

Issue #13783, opened 2013-12-10
Sidestep Protobuf-java binary incompatibility


viktorklang:
imported from https://www.assembla.com/spaces/akka/tickets/3783

Best idea so far is embedding the protobuf-java-2.4.1 (current version used in Akka) under a different package name into akka-remote.

http://stackoverflow.com/questions/13746737/hand-edit-a-jar-to-change-package-names

akka.com.google... ?

rkuhn:
This is the renaissance ;-) Currently looking into just including the (patched) sources in a different package.

clockfly:
Almost 2 years. Are there plans to fix this in recent release?

I saw there was a PR, and the PR was closed due to disagreement. 

IMHO, it is acceptable for ProtobufSerializer to still depends on google.protobuf  as an exception, as long as user app can choose to exclude protobuf.jar manually when linking akka. Maybe the code is not clean, but it can solve the protobuf version conflict issue. 

It is not good to sleep this issue. Hadoop and other infrastructure services depends on protobuf. Allowing this issue to still exist will make user app extremely difficult to program in akka to connect to these services.

clockfly:
issues related with akka protobuf dependency:
in flink: https://issues.apache.org/jira/browse/FLINK-2075
in spark: https://issues.apache.org/jira/browse/SPARK-2805

PR #16896, opened 2015-02-17
=all #13783 replace protobuf dependency by embedded version

rkuhn:
- created new subproject akka-protobuf (and added COPYING and LICENSE)
- renamed com.google.protobuf -> akka.protobuf everywhere
- also added such renaming step to the results of protoc compilation in
  project/Protobuf.scala
- had to include transcriptions of Netty’s ProtobufEncoder/Decoder to
  make multi-node-testkit compile again


rkuhn:
if this is what we’re doing then this will have to be backported to release-2.3

rkuhn:
fixes #13783.

rkuhn:
Currently on hold due to an unconsidered issue: what about users who are using our ProtobufSerializer for their (real) `com.google.protobuf.Message`?

patriknw:
I think this is a good move. I'm not convinced it can be done in 2.3.x, but 2.4 is a possible target if we allow some breakage.

I think the issue with "users who are using our ProtobufSerializer" is solvable, but the major breakage is that we cannot send these new messages to a node with an old version, since the class name is in the manifest. You are good at patching the bytes, so there might be a way ;-)

rkuhn:
What do you mean with “classname in the bytes”? AFAICS this PR has no influence on the serialized bytes themselves, only on which classes are used to produce them. But I may be missing something (I even hope so, because you are hinting at a way to solve this issue without breakage).

patriknw:
We don't use the ProtobufSerializer ourselves, but if users use it the classname of the protobuf messages will be in the messageManifest. The ProtobufSerializer must be for google.protobuf as before and not for akka.protobuf.

We can change the google protobuf dependency to provided, and let users add that dependency if they use the ProtobufSerializer. That is acceptable for 2.4, in my opinion.

We can actually get rid of the google protobuf dependency completely if we invoke `message.toByteArray` with reflection in the ProtobufSerializer, but that might not be important.

rkuhn:
The message manifest will contain the classname of the message type itself, not of its `com.google.protobuf.Message` supertype, so I don’t see why things should break. The important part would be to choose the right deserialization on the receiving end so that we create either `akka.protobuf.Message` or `com.google.protobuf.Message`, but that should be easily determined by reflection. So now I am nearly convinced that we can make this work fully compatible by supporting both protobuf variants in the very same serializer. Binary compatible, even.

rkuhn:
Is there another reason why you think this should not be in 2.3.x?

patriknw:
yes, that sounds correct

For 2.3.x we can't remove the google dependency, but it would be great if it can be excluded by user. Apart from that It can actually work for 2.3.x also. I have to look at the details of the changes.

bantonsson:
@rkuhn, I think that you are right that we can with some careful initialization make sure that we only "load" the `com.google.protobuf.Message` when it is used, and otherwise we just use the internal one.

akka-remote/src/main/resources/reference.conf
```diff
@@ -20,11 +20,11 @@ akka {
 
 
     serialization-bindings {
-      # Since com.google.protobuf.Message does not extend Serializable but
+      # Since akka.protobuf.Message does not extend Serializable but
       # GeneratedMessage does, need to use the more specific one here in order
       # to avoid ambiguity
       "akka.actor.ActorSelectionMessage" = akka-containers
-      "com.google.protobuf.GeneratedMessage" = proto
+      "akka.protobuf.GeneratedMessage" = proto
```
patriknw: `"com.google.protobuf.GeneratedMessage" = proto` may also be included in 2.4, and for 2.3 it must be included

project/Dependencies.scala
```diff
@@ -24,7 +24,6 @@ object Dependencies {
     // when updating config version, update links ActorSystem ScalaDoc to link to the updated version
     val config        = "com.typesafe"                % "config"                       % "1.2.1"       // ApacheV2
     val netty         = "io.netty"                    % "netty"                        % "3.8.0.Final" // ApacheV2
-    val protobuf      = "com.google.protobuf"         % "protobuf-java"                % "2.5.0"       // New BSD
```
patriknw: for 2.3 this must remain

patriknw:
I looked at the actual diff, and that convinced me that if we solve the `ProtobufSerializer` issue this can done in 2.3

rkuhn:
Thanks!

project/Dependencies.scala
rkuhn: why?

patriknw: We don't only want it to be a binary compatible drop-in, it should also be a source and dependency compatible drop-in. 
Here we remove the transitive dependency that users may depend on having. Or have I missed something?

rkuhn: The question was whether source and binary compatibility is perhaps enough, because build compatibility means that everyone who wants to profit from this change will need to manually exclude a dependency, something that not everybody does every day. But you might right that “drop-in replacement” does bind our hands here.

patriknw: yes, but those who need this will be happy to workaround it by adding the manual exclude
for 2.4 I don't think we have to be that strict

rkuhn: deal!

drewhk:
What is the status here?

drewhk:
@rkuhn ping?

rkuhn:
This needs more design work on how to make it work with shaded and non-shaded Message classes.

PR #16901, opened 2015-02-18
=pro Fix java8Home on mac osx

patriknw:
- for some reason there is a newline at the end

some temporary debug logging, because we have problems with pr validation


patriknw:
@ktoso alright, java8home is resolved correctly from the AKKA_BUILD_JAVA8_HOME environment variable

It still prints warning "Error reading API from class file : java.lang.UnsupportedClassVersionError: AllJavaTests : Unsupported major.minor version 52.0"

but that doesn't fail the builds as we noticed in #16772

but, this time it fails on akka-docs-java-lambda using maven!

I'm clueless.

PR #16911, opened 2015-02-19
Improve remote watching mechanism

trobert:
follow-up: https://github.com/akka/akka/pull/16867

This should include all comments of the previous discussion (especially on loops and casts)

I added a new commit with remove the casts by using InternalActorRef in more places. I am fine to remove it if you don't like it.

### DIFF
--- a/akka-actor/src/main/scala/akka/dispatch/sysmsg/SystemMessage.scala
+++ b/akka-actor/src/main/scala/akka/dispatch/sysmsg/SystemMessage.scala
@@ -239,11 +239,11 @@ private[akka] final case class Supervise(child: ActorRef, async: Boolean) extend
  * INTERNAL API
  */
 @SerialVersionUID(1L)
-private[akka] case class Watch(watchee: InternalActorRef, watcher: InternalActorRef) extends SystemMessage // sent to establish a DeathWatch
+private[akka] final case class Watch(watchee: InternalActorRef, watcher: InternalActorRef) extends SystemMessage // sent to establish a DeathWatch
 /**
  * INTERNAL API
  */
-@SerialVersionUID(1L)
+@SerialVersionUID(1L) // Watch and Unwatch have different signatures, but this can't be changed without breaking serialization compatibility
 private[akka] final case class Unwatch(watchee: ActorRef, watcher: ActorRef) extends SystemMessage // sent to tear down a DeathWatch
 /**
  * INTERNAL API
--- a/akka-cluster/src/multi-jvm/scala/akka/cluster/ClusterDeathWatchSpec.scala
+++ b/akka-cluster/src/multi-jvm/scala/akka/cluster/ClusterDeathWatchSpec.scala
@@ -6,28 +6,15 @@ package akka.cluster
 import language.postfixOps
 import scala.concurrent.Await
 import scala.concurrent.duration._
-import com.typesafe.config.ConfigFactory
-import org.scalatest.BeforeAndAfter
 import akka.remote.testkit.MultiNodeConfig
 import akka.remote.testkit.MultiNodeSpec
 import akka.testkit._
 import akka.testkit.TestEvent._
-import akka.actor.Props
-import akka.actor.Actor
-import akka.actor.Address
-import akka.actor.RootActorPath
-import akka.actor.Terminated
-import akka.actor.Address
+import akka.actor._
 import akka.remote.RemoteActorRef
 import java.util.concurrent.TimeoutException
-import akka.actor.ActorSystemImpl
-import akka.actor.ActorIdentity
-import akka.actor.Identify
-import akka.actor.ActorRef
 import akka.remote.RemoteWatcher
-import akka.actor.ActorSystem
 import akka.cluster.MultiNodeClusterSpec.EndActor
-import akka.actor.Deploy
 
 object ClusterDeathWatchMultiJvmSpec extends MultiNodeConfig {
   val first = role("first")
@@ -172,7 +159,9 @@ abstract class ClusterDeathWatchSpec
         // fifth is not cluster member, so the watch is handled by the RemoteWatcher
         awaitAssert {
           remoteWatcher ! RemoteWatcher.Stats
-          expectMsgType[RemoteWatcher.Stats].watchingRefs should contain((subject5, testActor))
+          val stats = expectMsgType[RemoteWatcher.Stats]
+          stats.watchingRefs should contain(subject5 → testActor)
+          stats.watchingAddresses should contain(address(fifth))
         }
       }
       enterBarrier("remote-watch")
@@ -181,13 +170,13 @@ abstract class ClusterDeathWatchSpec
       awaitClusterUp(first, fourth, fifth)
 
       runOn(first) {
-        // fifth is member, so the watch is handled by the ClusterRemoteWatcher,
-        // and cleaned up from RemoteWatcher
+        // fifth is member, so the node is handled by the ClusterRemoteWatcher,
+        // but the watch is still in RemoteWatcher
         awaitAssert {
           remoteWatcher ! RemoteWatcher.Stats
-          expectMsgType[RemoteWatcher.Stats].watchingRefs.map {
-            case (watchee, watcher) ⇒ watchee.path.name
-          } should not contain ("subject5")
+          val stats = expectMsgType[RemoteWatcher.Stats]
+          stats.watchingRefs.map(_._1.path.name) should contain("subject5")
+          stats.watchingAddresses should not contain address(fifth)
         }
       }
 
--- a/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala
+++ b/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala
@@ -8,14 +8,12 @@ import akka.actor._
 import akka.dispatch.sysmsg._
 import akka.event.{ Logging, LoggingAdapter, EventStream }
 import akka.event.Logging.Error
-import akka.serialization.{ JavaSerializer, Serialization, SerializationExtension }
+import akka.serialization.{ Serialization, SerializationExtension }
 import akka.pattern.pipe
 import scala.util.control.NonFatal
 import akka.actor.SystemGuardian.{ TerminationHookDone, TerminationHook, RegisterTerminationHook }
 import scala.util.control.Exception.Catcher
-import scala.concurrent.{ ExecutionContext, Future }
-import scala.concurrent.forkjoin.ThreadLocalRandom
-import com.typesafe.config.Config
+import scala.concurrent.Future
 import akka.ConfigurationException
 import akka.dispatch.{ RequiresMessageQueue, UnboundedMessageQueueSemantics }
 
@@ -150,7 +148,9 @@ private[akka] class RemoteActorRefProvider(
   // This actor ensures the ordering of shutdown between remoteDaemon and the transport
   @volatile private var remotingTerminator: ActorRef = _
 
-  @volatile private var remoteWatcher: ActorRef = _
+  @volatile private var _remoteWatcher: ActorRef = _
+  private[akka] def remoteWatcher = _remoteWatcher
+
   @volatile private var remoteDeploymentWatcher: ActorRef = _
 
   def init(system: ActorSystemImpl): Unit = {
@@ -183,7 +183,7 @@ private[akka] class RemoteActorRefProvider(
     // this enables reception of remote requests
     transport.start()
 
-    remoteWatcher = createRemoteWatcher(system)
+    _remoteWatcher = createRemoteWatcher(system)
     remoteDeploymentWatcher = createRemoteDeploymentWatcher(system)
   }
 
@@ -422,20 +422,6 @@ private[akka] class RemoteActorRefProvider(
    */
   def quarantine(address: Address, uid: Option[Int]): Unit = transport.quarantine(address, uid)
 
-  /**
-   * INTERNAL API
-   */
-  private[akka] def afterSendSystemMessage(message: SystemMessage): Unit =
-    message match {
-      // Sending to local remoteWatcher relies strong delivery guarantees of local send, i.e.
-      // default dispatcher must not be changed to an implementation that defeats that
-      case rew: RemoteWatcher.Rewatch ⇒
-        remoteWatcher ! RemoteWatcher.RewatchRemote(rew.watchee, rew.watcher)
-      case Watch(watchee, watcher)   ⇒ remoteWatcher ! RemoteWatcher.WatchRemote(watchee, watcher)
-      case Unwatch(watchee, watcher) ⇒ remoteWatcher ! RemoteWatcher.UnwatchRemote(watchee, watcher)
-      case _                         ⇒
-    }
-
 }
 
 private[akka] trait RemoteRef extends ActorRefScope {
@@ -475,10 +461,30 @@ private[akka] class RemoteActorRef private[akka] (
       remote.system.eventStream.publish(Error(e, path.toString, getClass, "swallowing exception during message send"))
   }
 
+  /**
+   * Determine if a watch/unwatch message must be handled by the remoteWatcher actor, or sent to this remote ref
+   */
+  def isWatchIntercepted(watchee: ActorRef, watcher: ActorRef) =
+    if (watchee.path.uid == akka.actor.ActorCell.undefinedUid) {
+      provider.log.debug("actorFor is deprecated, and watching a remote ActorRef acquired with actorFor is not reliable: [{}]", watchee.path)
+      false // Not managed by the remote watcher, so not reliable to communication failure or remote system crash
+    } else {
+      // If watchee != this then watcher should == this. This is a reverse watch, and it is not intercepted
+      // If watchee == this, only the watches from remoteWatcher are sent on the wire, on behalf of other watchers
+      watcher != provider.remoteWatcher && watchee == this
+    }
+
   def sendSystemMessage(message: SystemMessage): Unit =
     try {
-      remote.send(message, None, this)
-      provider.afterSendSystemMessage(message)
+      //send to remote, unless watch message is intercepted by the remoteWatcher
+      message match {
+        case Watch(watchee, watcher) if isWatchIntercepted(watchee, watcher) ⇒
+          provider.remoteWatcher ! RemoteWatcher.WatchRemote(watchee, watcher)
+        //Unwatch has a different signature, need to pattern match arguments against InternalActorRef
+        case Unwatch(watchee: InternalActorRef, watcher: InternalActorRef) if isWatchIntercepted(watchee, watcher) ⇒
+          provider.remoteWatcher ! RemoteWatcher.UnwatchRemote(watchee, watcher)
+        case _ ⇒ remote.send(message, None, this)
+      }
     } catch handleException
 
   override def !(message: Any)(implicit sender: ActorRef = Actor.noSender): Unit = {
--- a/akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
+++ b/akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
@@ -3,23 +3,14 @@
  */
 package akka.remote
 
-import scala.concurrent.duration._
-import akka.actor.Actor
-import akka.actor.ActorLogging
-import akka.actor.ActorRef
-import akka.actor.Address
-import akka.actor.AddressTerminated
-import akka.actor.Props
-import akka.actor.RootActorPath
-import akka.actor.Terminated
-import akka.actor.ExtendedActorSystem
 import akka.ConfigurationException
-import akka.dispatch.{ UnboundedMessageQueueSemantics, RequiresMessageQueue }
-import akka.actor.InternalActorRef
-import akka.dispatch.sysmsg.DeathWatchNotification
-import akka.dispatch.sysmsg.Watch
-import akka.actor.Deploy
+import akka.actor._
+import akka.dispatch.sysmsg.{ DeathWatchNotification, Watch }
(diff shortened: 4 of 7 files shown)


ktoso:
OK TO TEST

ktoso:
Thanks for re-targeting!

patriknw:
@trobert we have not forgot this, we have just not had any spare cycles to ponder it

trobert:
Ok

akka-actor/src/main/scala/akka/dispatch/sysmsg/SystemMessage.scala
```diff
@@ -237,12 +237,12 @@ private[akka] final case class Supervise(child: ActorRef, async: Boolean) extend
...
  */
 @SerialVersionUID(1L)
-private[akka] case class Watch(watchee: InternalActorRef, watcher: InternalActorRef) extends SystemMessage // sent to establish a DeathWatch
+private[akka] final case class Watch(watchee: InternalActorRef, watcher: InternalActorRef) extends SystemMessage // sent to establish a DeathWatch
 /**
  * INTERNAL API
  */
 @SerialVersionUID(1L)
-private[akka] final case class Unwatch(watchee: ActorRef, watcher: ActorRef) extends SystemMessage // sent to tear down a DeathWatch
+private[akka] final case class Unwatch(watchee: InternalActorRef, watcher: InternalActorRef) extends SystemMessage // sent to tear down a DeathWatch
```
rkuhn: Does deserialization of the old with the new still work (and vice versa)? If not then we need to roll this back.

akka-cluster/src/multi-jvm/scala/akka/cluster/ClusterDeathWatchSpec.scala
```diff
@@ -6,28 +6,15 @@ package akka.cluster
...
 import akka.remote.RemoteActorRef
 import java.util.concurrent.TimeoutException
-import akka.actor.ActorSystemImpl
-import akka.actor.ActorIdentity
-import akka.actor.Identify
-import akka.actor.ActorRef
 import akka.remote.RemoteWatcher
-import akka.actor.ActorSystem
 import akka.cluster.MultiNodeClusterSpec.EndActor
-import akka.actor.Deploy
```
rkuhn: good!

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
```diff
@@ -207,57 +194,70 @@ private[akka] class RemoteWatcher(
...
+
+    watching -= watchee
     checkLastUnwatchOfNode(watchee.path.address)
   }
 
   def checkLastUnwatchOfNode(watcheeAddress: Address): Unit = {
-    if (watchingNodes(watcheeAddress) && watching.forall {
-      case (wee, wer) ⇒ wee.path.address != watcheeAddress
-    }) {
+    if (watchingNodes(watcheeAddress) && watching.keys.forall(_.path.address != watcheeAddress)) {
```
rkuhn: Doing this for every termination can be quite costly; can we keep track of this property otherwise?

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
```diff
@@ -287,15 +287,9 @@ private[akka] class RemoteWatcher(
    * does not exist.
    */
   def reWatch(address: Address): Unit =
```
rkuhn: Is this still necessary? My intuition is that we should now have all the information in one place to just fire DeathWatchNotification upon system UID change, but I might be missing something. @patriknw ?

rkuhn:
@trobert Thanks a lot, this looks like a very good idea!

akka-actor/src/main/scala/akka/dispatch/sysmsg/SystemMessage.scala
patriknw: too bad that we have not protobuf:ed these messages

akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala
```diff
@@ -150,7 +148,7 @@ private[akka] class RemoteActorRefProvider(
   // This actor ensures the ordering of shutdown between remoteDaemon and the transport
   @volatile private var remotingTerminator: ActorRef = _
 
-  @volatile private var remoteWatcher: ActorRef = _
+  @volatile private[akka] var remoteWatcher: ActorRef = _
```
patriknw: I would prefer to have this `private _remoteWatcher` and add `private[akka] def remoteWatcher` so that it is clear that it is only set from here.

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
```diff
@@ -107,8 +105,8 @@ private[akka] class RemoteWatcher(
 
   val selfHeartbeatRspMsg = HeartbeatRsp(AddressUidExtension(context.system).addressUid)
 
-  // actors that this node is watching, tuple with (watcher, watchee)
-  var watching: Set[(ActorRef, ActorRef)] = Set.empty
+  // actors that this node is watching, map of watchee -> Set(watchers)
+  var watching: Map[InternalActorRef, Set[InternalActorRef]] = Map.empty
```
patriknw: perhaps use the new akka.util.TypedMultiMap?

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
```diff
@@ -177,28 +174,18 @@ private[akka] class RemoteWatcher(
...
-      log.debug("Ignoring re-watch after being unwatched in the meantime: [{} -> {}]", watcher.path, watchee.path)
-
-  def watchRemote(watchee: ActorRef, watcher: ActorRef): Unit =
+  def watchRemote(watchee: InternalActorRef, watcher: InternalActorRef): Unit = {
     if (watchee.path.uid == akka.actor.ActorCell.undefinedUid)
       logActorForDeprecationWarning(watchee)
-    else if (watcher != self) {
-      log.debug("Watching: [{} -> {}]", watcher.path, watchee.path)
-      addWatching(watchee, watcher)
+    log.debug("Watching: [{} -> {}]", watcher.path, watchee.path)
```
patriknw: perhaps add `assert(watcher != self)`?

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
```diff
@@ -207,57 +194,70 @@ private[akka] class RemoteWatcher(
...
-    }
-
+    // When watchee is stopped it sends DeathWatchNotification to this RemoteWatcher,
+    // which will propagate it to all watchers of this watchee.
+    // addressTerminated case is already handled by the watcher itself in DeathWatch trait
+    if (!addressTerminated)
+      for {
+        watchers ← watching.get(watchee)
+        watcher ← watchers
+        if watcher != self
```
patriknw: can `watcher` be `self`?

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
patriknw: The rewatch mechanism was introduced in 637598a28bc5aafaca27b801f442cf9b20009807
When actor system was restarted quickly the new system replied to heartbeats and Terminated was never triggered for actors in old system.

Is that really solved by this?

The test fails when I remove the reWatch: akka.remote.RemoteNodeShutdownAndComesBack

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
```diff
@@ -287,15 +287,9 @@ private[akka] class RemoteWatcher(
...
-          // this re-watch will result in a RewatchRemote message to this actor
-          // must be a special message to be able to detect if an UnwatchRemote comes in
-          // before the extra RewatchRemote, then the re-watch should be ignored
-          log.debug("Re-watch [{} -> {}]", wer, wee)
-          wee.sendSystemMessage(new Rewatch(wee, wer)) // ➡➡➡ NEVER SEND THE SAME SYSTEM MESSAGE OBJECT TO TWO ACTORS ⬅⬅⬅
-        }
+    watching.keys.withFilter(_.path.address == address) foreach { watchee ⇒
+      val watcher = self.asInstanceOf[InternalActorRef]
+      log.debug("Re-watch [{} -> {}]", watcher.path, watchee.path)
+      watchee.sendSystemMessage(Watch(watchee, watcher)) // ➡➡➡ NEVER SEND THE SAME SYSTEM MESSAGE OBJECT TO TWO ACTORS ⬅⬅⬅
```
patriknw: Previous implementation with the `Rewatch`/  `RewatchRemote` message handled the case when there was a unwatch inbetween. Is that handled now?

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
rkuhn: My question was: can we not just unconditionally fire the DeathWatchNotifications (without reWatch) in case a new UID is seen?

patriknw:
@trobert I have reviewed this, finally. I think it looks great. We need to nail the comments around re-watch.

It might change the order of messages compared to previous implementation, but as far as I can see that should not influence correctness.

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
patriknw: What shall we do the first time, when uid is not know?
The uid is of importance to trigger quarantine.

Triggering DeathWatchNotification upon uid change when the previous uid was know might work, but what makes that work now and not previously? Just curious to understand if I missed something.

rkuhn: same here: I might well be missing something

akka-actor/src/main/scala/akka/dispatch/sysmsg/SystemMessage.scala
trobert: I would have to test, but I'm quite sure it will not interoperate with older format.
Is it too late to change ? 2.4 must be wire-compatible with 2.3 ? I'd like to see this in protobuf too !

Otherwise, the 2nd commit may be dropped or changed to something else (it's merely "cosmetic", not functional)

trobert:
Thanks again for the reviews. I'll try to get my head again into this in the next days, especially the re-watch thing, which seems to be sensible.

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
trobert: I agree, we could introduce another Map to avoid this

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
trobert: TypedMultiMap is used for mapping types to values.
What is needed is a multimap. There is scala.collection.mutable.MultiMap, but only for mutable maps. I think this is a viable option, though

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
trobert: No, using `assert` is a good idea

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
trobert: I think there is no "inbetween" possible with this implementation.
When a heartbeat message with a new uid is received, it sends Watch in the processing of this message.
Previous implementation used Rewatch messages (indirectly) sent to self, but I don't see the reason. Checking the unwatch in between was making sense only because of this.

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
trobert: To my understandig, I think it is risky to trigger DeathWatchNotification when seeing new uid. There can be some new watches established with actors on system with the new uid. There is no way to distinguish them unless we store the uid along with all the watches (the tuples of watcher, watchee)
But I am more likely to miss something than you ;)

trobert:
@patriknw Sorry I did not answer before. I pushed some commits followings the remarks.
To me it is in a good shape to be merged. Especially, I removed the serialization incompatibility.

If this is ok for you, I can rework the commits to make them cleaner.

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
```diff
@@ -177,87 +168,88 @@ private[akka] class RemoteWatcher(
...
-      log.debug("Watching: [{} -> {}]", watcher.path, watchee.path)
-      addWatching(watchee, watcher)
 
-      // also watch from self, to be able to cleanup on termination of the watchee
-      context watch watchee
-      watching += ((watchee, self))
-    }
+    log.debug("Watching: [{} -> {}]", watcher.path, watchee.path)
+    watching.addBinding(watchee, watcher)
+    watchNode(watchee)
```
patriknw: now we do this also for refs with `undefinedUid`, could that change have any bad side effects?

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
```diff
@@ -177,87 +168,88 @@ private[akka] class RemoteWatcher(
...
-        watching -= ((watchee, self))
-      }
-      checkLastUnwatchOfNode(watchee.path.address)
+    log.debug("Unwatching: [{} -> {}]", watcher.path, watchee.path)
+    watching.get(watchee) match {
+      case Some(watchers) if watchers == Set(watcher) ⇒
+        clearAllWatches(watchee)
+        checkLastUnwatchOfNode(watchee)
+      case Some(watchers) ⇒
+        watching += watchee → (watchers - watcher)
```
patriknw: removeBinding?

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
patriknw: I think the unwatch inbetween is handled by that this is iterating the `watching`, so ok.

patriknw:
LGTM, some minor nitpicks
I will create a jenkins repeat job for this branch to make sure that tests a happy.

patriknw:
The repeat job has been running for a few days and did not find any test failures related to this.
In other words: final review by some more team members and then we can merge.
Great job @trobert

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
trobert: Using a match saves a contain call to know if this was the last watcher.
I see now that I could use the same approach in checkLastUnwatchOfNode to save 2 contains call.

What's your opinion (match or removeBinding) ?

patriknw: whatever is most efficient, and a comment about the reason why it was done like that

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
trobert: Previous behavior was to ignore the watch. Was this deliberate ? The log message is not very clear. I'll do some testing

patriknw: yes, it was, but previously the watch was also sent from RemoteActorRefProvider.afterSendSystemMessage, which is now not done.

I think it was cautious with those actorFor ActorRefs here because they don't include uid and therefore the equals would not match when there is a mix of those and ordinary ActorRef.
I think we should try to retain the old behaviour.

trobert: You mean if there is no uid, send the watch over the wire and don't let the RemoteWatcher handle it ? That makes sense to me

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
```diff
@@ -177,87 +168,88 @@ private[akka] class RemoteWatcher(
...
-      // clean up self watch when no more watchers of this watchee
-      if (watching.forall { case (wee, wer) ⇒ wee != watchee || wer == self }) {
-        log.debug("Cleanup self watch of [{}]", watchee.path)
-        context unwatch watchee
-        watching -= ((watchee, self))
-      }
-      checkLastUnwatchOfNode(watchee.path.address)
+    log.debug("Unwatching: [{} -> {}]", watcher.path, watchee.path)
+    watching.get(watchee) match {
+      case Some(watchers) if watchers == Set(watcher) ⇒
```
2beaucoup: I wonder if `case Some(watchers) if watchers.forall(_ == watcher)` would be more efficient here.

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
patriknw: exactly, that was the old behaviour, thanks
keep the log message also

akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala
trobert: This will be slow if the set is very large

I changed to `watchers -= watcher ; if (watchers.isEmpty)` 
This is the way it is done in Multimap, I think it is the fastest.

2beaucoup: > This will be slow if the set is very large

But `forall` fails fast, right?

patriknw: if you change to `forall`, be careful with empty collection, but I guess that it will not be empty here

2beaucoup: yeah, that's the tricky part with `forall` ;)

trobert: I did some bench: https://gist.github.com/trobert/d0983a2b7236e488a94d

`remove` (`-=`) + `isEmpty` is the fastest way to test `set == Set(a)` when the test is true (tie with `size==1 && set(a)`)

When it's false, I must call remove anyway, so other methods won't give any benefits.

2beaucoup: thanks for the verification @trobert! doesn't `sizeContains` score highest?

trobert: It is 1 ops / us higher for PresentSmall benchmark, that is to say when the test is true. So I consider it equivalent to removeEquals.
And in the other case, I need to remove the element from the set, so I will call remove in the end too. 
Hope this is clear enough

2beaucoup: > Hope this is clear enough

well, yes sir. ;)

trobert:
These last commits correct the last remarks. Especially, the `undefinedUid` case is now handled specifically.
I also take care of "reverse watches" (a watch message sent to the watcher, not the watchee). Such messages should be sent on the wire.
And I took benefit of the new map (address->watchee) in the rewatch method. There is no more full iteration on watches now.

rkuhn:
@patriknw More repeat jobs after the latest fixes, then merge?

patriknw:
the last 4 commits looks good
@trobert please squash to one commit
I will run some extra tests runs, and then merge.

trobert:
Here is the squashed commit 
I just changed the signature of the `Stats.watchingRefs` object back to the original `Set[ActorRef, ActorRef]` to avoid messing too much the tests

patriknw:
@trobert It was binary compatibility check that failed. It is only internal changes so you can add the following.

    ProblemFilters.exclude[FinalClassProblem]("akka.dispatch.sysmsg.Watch"),

at https://github.com/akka/akka/blob/master/project/MiMa.scala#L410

and 

    ,
    FilterAnyProblem("akka.remote.RemoteWatcher")

at https://github.com/akka/akka/blob/master/project/MiMa.scala#L519

You check that it works with `sbt mimaReportBinaryIssues`

trobert:
it seems I also need to add filters for:
`akka.remote.RemoteActorRefProvider` ,
 `akka.remote.RemoteWatcher#WatchRemote`,
`akka.remote.RemoteWatcher#UnwatchRemote`, 
`akka.remote.RemoteWatcher$RewatchRemote`, 
`akka.remote.RemoteWatcher#Stats`

patriknw:
ok, please do so, those are also internal
The change in RemoteActorRefProvider was the removal of the internal `afterSendSystemMessage` method

    ProblemFilters.exclude[MissingMethodProblem]("akka.remote.RemoteActorRefProvider.afterSendSystemMessage")

trobert:
I have some issues with the case class, it still reports things like:
`[error]  * method watchee()akka.actor.ActorRef in class akka.remote.RemoteWatcher#WatchRemote has now a different result type; was: akka.actor.ActorRef, is now: akka.actor.InternalActorRef
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.remote.RemoteWatcher#WatchRemote.watchee")
`

even with this filter:
`FilterAnyProblem("akka.remote.RemoteWatcher#WatchRemote"),`

any idea ?

trobert:
Ok I think I found it, I need: `FilterAnyProblem("akka.remote.RemoteWatcher$WatchRemote")` instead

rkuhn:
Yes, that should be correct.

patriknw:
@trobert I'm about to merge this now. Thanks for an outstanding contribution. Are you `thibautrobert` at twitter? I would like to give credit to the right person.

trobert:
Well, that's me, thanks !

trobert:
I was happy to contribute by the way, thanks for the merge, and your open mind. I Hope to have time to do that again :)

patriknw:
Refs #17480

PR #16918, opened 2015-02-23
!htp #16833 Add minimum number of iterations to `PathMatcher::repeat` modifier

### LINKED ISSUE #16833: Add minimum number of iterations to `PathMatcher::repeat` modifier
In comparison to the spray variant, in akka-http `repeat` already allows to specify a maximum number of iterations. As suggested in spray/spray#636 amd spray/spray#660, also allowing to specify the minimum number may be useful as well.

/cc @sirthias

sirthias:
Closes #16833.

Unfortunately this change breaks existing code using `Segments(n)` or `xxx.repeat(n)` because the semantics of these calls change with this PR.
I've played around with alternative solutions that do not have this caveat but didn't like them enough in the end.
IMHO, at this point in the akka-http project life cycle we should rather break the few existing applications already relying on `Segments(n)` or `xxx.repeat(n)` than compromise our public API for the future.

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/server/directives/PathDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/server/directives/PathDirectivesSpec.scala
@@ -199,10 +199,34 @@ class PathDirectivesSpec extends RoutingSpec with Inside {
     "accept [/foo/] and clear the unmatchedPath" in test("")
   }
 
-  """pathPrefix(IntNumber.repeat(separator = "|"))""" should {
-    val test = testFor(pathPrefix(IntNumber.repeat(maxIterations = 5, separator = "|")) { echoCaptureAndUnmatchedPath })
-    "accept [/1|2|3rest]" in test("List(1, 2, 3):rest")
-    "accept [/rest]" in test("List():rest")
+  """pathPrefix(IntNumber.repeat(separator = "."))""" should {
+    {
+      val test = testFor(pathPrefix(IntNumber.repeat(min = 2, max = 5, separator = ".")) { echoCaptureAndUnmatchedPath })
+      "reject [/foo]" in test()
+      "reject [/1foo]" in test()
+      "reject [/1.foo]" in test()
+      "accept [/1.2foo]" in test("List(1, 2):foo")
+      "accept [/1.2.foo]" in test("List(1, 2):.foo")
+      "accept [/1.2.3foo]" in test("List(1, 2, 3):foo")
+      "accept [/1.2.3.foo]" in test("List(1, 2, 3):.foo")
+      "accept [/1.2.3.4foo]" in test("List(1, 2, 3, 4):foo")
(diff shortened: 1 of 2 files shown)


akka-http/src/main/scala/akka/http/server/PathMatcher.scala
```diff
@@ -48,8 +48,25 @@ abstract class PathMatcher[L](implicit val ev: Tuple[L]) extends (Path ⇒ PathM
   def tflatMap[R: Tuple](f: L ⇒ Option[R]): PathMatcher[R] = transform(_.flatMap(f))
 
   /**
-   * Turns this ``PathMatcher`` into one that always matches, consumes zero or more times (with the given separator)
+   * Same as ``repeat(min = iterations, max = iterations)``.
```
jrudolph: s/iterations/count/ ?

jrudolph:
LGTM

akka-http/src/main/scala/akka/http/server/PathMatcher.scala
sirthias: Yes.

sirthias:
Fixed stale copy-pasta detected by @jrudolph.

akka-http-tests/src/test/scala/akka/http/server/directives/PathDirectivesSpec.scala
```diff
@@ -199,10 +199,31 @@ class PathDirectivesSpec extends RoutingSpec with Inside {
...
+  """pathPrefix(IntNumber.repeat(separator = "."))""" should {
+    {
+      val test = testFor(pathPrefix(IntNumber.repeat(min = 2, max = 5, separator = ".")) { echoCaptureAndUnmatchedPath })
+      "reject [/foo]" in test()
+      "reject [/1foo]" in test()
+      "reject [/1.foo]" in test()
+      "accept [/1.2foo]" in test("List(1, 2):foo")
+      "accept [/1.2.foo]" in test("List(1, 2):.foo")
+      "accept [/1.2.3foo]" in test("List(1, 2, 3):foo")
+      "accept [/1.2.3.foo]" in test("List(1, 2, 3):.foo")
```
rkuhn: perhaps add a case for `[/1.2.3.]` just to be safe for end-of-uri slashes?

rkuhn:
Test failure was unrelated, I have relaxed the timeouts in SourceSpec (which were quite a bit too tight).

akka-http-tests/src/test/scala/akka/http/server/directives/PathDirectivesSpec.scala
sirthias: Yup, good idea. Will add.

sirthias:
Added a few more test cases as suggested by @rkuhn.

rkuhn:
Thanks!

PR #16941, opened 2015-02-26
Updating documentation basic sections to M4 API

drewhk:


akka-docs-dev/rst/scala/code/docs/stream/FlowDocSpec.scala
```diff
@@ -138,4 +138,69 @@ class FlowDocSpec extends AkkaSpec {
...
     //#flow-connecting
   }
+
+  "various ways of transforming materialized values" in {
+    import scala.concurrent.duration._
+
+    val throttler = Flow(Source(1.second, 1.second, "test")) { implicit b =>
+      tickSource =>
+        import FlowGraph.Implicits._
+        val zip = b.add(Zip[String, Int](OperationAttributes.inputBuffer(1, 1)))
```
rkuhn: use ZipWith

akka-docs-dev/rst/scala/code/docs/stream/FlowDocSpec.scala
```diff
@@ -138,4 +138,69 @@ class FlowDocSpec extends AkkaSpec {
...
+    val source: Source[Int, Promise[Unit]] = Source.lazyEmpty[Int]()
+
+    // A flow that internally throttles elements to 1/second, and returns a Cancellable
+    // which can be used to shut down the stream
+    val flow: Flow[Int, Int, Cancellable] = throttler
+
+    // A sink that returns the first element of a stream in the returned Future
+    val sink: Sink[Int, Future[Int]] = Sink.head[Int]()
+
+    // Simple selection of materialized values by using Keep.right
```
rkuhn: perhaps say that the default of `via` and `to` is to keep the left value—even if it is in the text, this might bear repeating

akka-docs-dev/rst/scala/code/docs/stream/FlowDocSpec.scala
```diff
@@ -138,4 +138,69 @@ class FlowDocSpec extends AkkaSpec {
...
+    val r1: RunnableFlow[Promise[Unit]] = source.via(flow).to(sink)
+    val r2: RunnableFlow[Cancellable] = source.viaMat(flow)(Keep.right).to(sink)
+    val r3: RunnableFlow[Future[Int]] = source.via(flow).toMat(sink)(Keep.right)
+
+    // Using runWith will always give the materialized values of the stages added
+    // by runWith() itself
+    val r4: Future[Int] = source.via(flow).runWith(sink)
+    val r5: Promise[Unit] = flow.to(sink).runWith(source)
+    val r6: (Promise[Unit], Future[Int]) = flow.runWith(source, sink)
```
rkuhn: I think this demonstrates that these are the right defaults, the behavior can be described concisely :-)

akka-docs-dev/rst/scala/code/docs/stream/FlowDocSpec.scala
drewhk: it doesn't matter, I just wanted a Flow with a materialized value :) Not linked from docs, and not executed. I can fix though.

akka-docs-dev/rst/scala/code/docs/stream/FlowDocSpec.scala
```diff
@@ -138,4 +138,69 @@ class FlowDocSpec extends AkkaSpec {
...
+    val r10: RunnableFlow[(Cancellable, Future[Int])] =
+      source.viaMat(flow)(Keep.right).toMat(sink)(Keep.both)
+
+    // It is also possible to map over the materialized values. In r9 we had a
+    // doubly nested pair, but we want to flatten it out
+    val r11: RunnableFlow[(Promise[Unit], Cancellable, Future[Int])] =
+      r9.mapMaterialized {
+        case ((promise, cancellable), future) =>
+          (promise, cancellable, future)
+      }
```
rkuhn: perhaps show the alternative:
~~~scala
val r = FlowGraph.closed(source, flow, sink)((_, _, _)) { implicit b =>
  source ~> flow ~> sink
}
~~~

drewhk: This is the basics doc, we don't refer to graphs here

drewhk: and you forgot to properly import as function arguments :)

rkuhn: perhaps the following even compiles?
~~~scala
val r = FlowGraph.closed(source, flow, sink)((_, _, _)) { implicit b =>
  _ ~> _ ~> _
}

akka-docs-dev/rst/scala/stream-flows-and-basics.rst
```diff
@@ -85,7 +88,7 @@ instead of modifying the existing instance, so while constructing long flows, re
 In the above example we used the ``runWith`` method, which both materializes the stream and returns the materialized value
 of the given sink or source.
 
-Since a stream can be materialized multiple times, the ``MaterializedMap`` returned is different for each materialization.
+Since a stream can be materialized multiple times, the materialized value returned is different for each materialization.
```
rkuhn: The value is calculated fresh every time, but the result _might_ in certain cases not be different …

akka-docs-dev/rst/scala/stream-quickstart.rst
```diff
@@ -69,8 +70,8 @@ combinator:
 
 .. note::
   The name ``flatMap`` was consciously avoided due to its proximity with for-comprehensions and monadic composition.
-  It is problematic for two reasons: firstly, flattening by concatenation is often undesirable in bounded stream processing
-  due to the risk of deadlock (with merge being the preferred strategy), and secondly, the monad laws would not hold for
+  It is problematic for two reasons: first, flattening by concatenation is often undesirable in bounded stream processing
+  due to the risk of deadlock (with merge being the preferred strategy), and second, the monad laws would not hold for
```
rkuhn: well, this is a matter of taste ;-) (and perhaps British vs. American)

akka-docs-dev/rst/scala/code/docs/stream/FlowDocSpec.scala
rkuhn: Yes, I mention it because these tests are crucial to verifying the API that we provide, and if ZipWith does not work “easily” here then we know what to fix ;-)

akka-docs-dev/rst/scala/code/docs/stream/FlowDocSpec.scala
rkuhn: well, yes, I meant to include the “oneliner” only as a teaser for the later Graph parts, because they are SOOO COOOL :-)

akka-docs-dev/rst/scala/code/docs/stream/FlowDocSpec.scala
drewhk: yes, for example ZipWith does not take attributes?

akka-docs-dev/rst/scala/code/docs/stream/FlowDocSpec.scala
drewhk: ok, adding

akka-docs-dev/rst/scala/stream-quickstart.rst
drewhk: we removed them in other cases, tried to stay consistent

akka-docs-dev/rst/scala/stream-flows-and-basics.rst
drewhk: I don't think it is worth to mention here. It just makes the introductory documentation harder to digest. Those elements which might materialize to the same value can explicitly say that in their API docs.

akka-docs-dev/rst/scala/stream-quickstart.rst
rkuhn: okay

akka-docs-dev/rst/scala/stream-flows-and-basics.rst
rkuhn: Yes, and no (you got me slightly wrong), my proposal is:

> Since a stream can be materialized multiple times, the materialized value will also be calculated anew for each such materialization, usually leading to different values being returned each time.

drewhk: ok, added that sentence

akka-docs-dev/rst/scala/code/docs/stream/FlowDocSpec.scala
```diff
@@ -138,4 +138,78 @@ class FlowDocSpec extends AkkaSpec {
...
+    val (promise, cancellable, future) = r11.run()
+
+    // Type inference works as expected
+    promise.success(0)
+    cancellable.cancel()
+    future.map(_ + 3)
+
+    // The result of r11 can be also achieved by using the Graph API
+    val r12: RunnableFlow[(Promise[Unit], Cancellable, Future[Int])] =
+      FlowGraph.closed(source, flow, sink)((_, _, _)) { implicit b =>
```
rkuhn: upon closer reflection it would probably be helpful to change this name to `builder` here

rkuhn:
LGTM apart from the last nitpick

rkuhn:
LGTM: please merge after fixes rather than stacking things onto this PR since I’d like to incorporate the improvements while working on the rest.

PR #16943, opened 2015-02-26
fix a good swath of FIXMEs

rkuhn:


akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
```diff
@@ -152,13 +151,20 @@ abstract class FlowMaterializer {
...
   def withNamePrefix(name: String): FlowMaterializer
 
-  // FIXME this is scaladsl specific
   /**
    * This method interprets the given Flow description and creates the running
    * stream. The result can be highly implementation specific, ranging from
    * local actor chains to remote-deployed processing networks.
    */
-  def materialize[Mat](runnable: RunnableFlow[Mat]): Mat
+  def materialize[Mat](runnable: Graph[ClosedShape, Mat]): Mat
```
drewhk: so I need to update docs again :) Don't we want to still have a special type? I understand what the above type means, but I am not sure how confusing it is for users.

akka-stream/src/main/scala/akka/stream/Shape.scala
```diff
@@ -55,7 +61,7 @@ abstract class Shape {
   /**
    * Compare this to another shape and determine whether the arrangement of ports is the same (including their ordering).
    */
-  def hasSameShapeAs(s: Shape): Boolean =
+  def hasSamePortsAndShapeAs(s: Shape): Boolean =
```
drewhk: thanks

akka-stream/src/main/scala/akka/stream/impl/ActorProcessor.scala
```diff
@@ -106,10 +106,13 @@ private[akka] abstract class BatchingInputBuffer(val size: Int, val pump: Pump)
 
   protected def onSubscribe(subscription: Subscription): Unit = {
     assert(subscription != null)
-    upstream = subscription
-    // Prefetch
-    upstream.request(inputBuffer.length)
-    subreceive.become(upstreamRunning)
+    if (upstreamCompleted) subscription.cancel()
```
drewhk: good one!

akka-stream/src/main/scala/akka/stream/impl/ActorProcessor.scala
```diff
@@ -132,7 +135,7 @@ private[akka] abstract class BatchingInputBuffer(val size: Int, val pump: Pump)
   }
 
   protected def completed: Actor.Receive = {
-    case OnSubscribe(subscription) ⇒ throw new IllegalStateException("Cannot subscribe shutdown subscriber") // FIXME "shutdown subscriber"?!
+    case OnSubscribe(subscription) ⇒ throw new IllegalStateException("onSubscribe called after onError or onComplete")
```
drewhk: nice

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
rkuhn: Users will use the DSL to materialize things, not this bare-bones facility. And this is the only sensible way to express a generic “closed graph” concept without huge API baggage—the problem with RunnableFlow is that it has a language-specific `mapMaterialized` method on it.

akka-stream/src/main/scala/akka/stream/impl/SubscriberManagement.scala
```diff
@@ -126,8 +126,16 @@ private[akka] trait SubscriberManagement[T] extends ResizableMultiReaderRingBuff
...
+                  val goOn = try {
+                    subscription.dispatch(buffer.read(subscription))
+                    true
+                  } catch {
+                    case _: SpecViolation ⇒
+                      unregisterSubscriptionInternal(subscription)
+                      false
+                  }
+                  if (goOn) dispatchFromBufferAndReturnRemainingRequested(requested - 1, eos)
+                  else Long.MinValue
```
drewhk: what is this value used for?

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -382,7 +382,7 @@ class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Graph
  *
  * Flow with attached input and output, can be executed.
  */
-trait RunnableFlow[+Mat] {
+trait RunnableFlow[+Mat] extends Graph[ClosedShape, Mat] {
```
drewhk: ah, I see

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
drewhk: I've seen that RunnableFlow is still there, just extends this type, so everything is OK!

drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/impl/SubscriberManagement.scala
rkuhn: this is the contract of the method; tests pass, but we will remove this whole file (and a few others) soon I hope

PR #16944, opened 2015-02-26
fix two wrong/flaky tests

rkuhn:
- ordering is not preserved by shufflers in GraphOpsIntegrationSpec
- larger tolerance is needed in GraphBalanceSpec since balancing does
  not keep track of previous imbalances
- also add Source.constant(elem)

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphOpsIntegrationSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphOpsIntegrationSpec.scala
@@ -3,13 +3,13 @@ package akka.stream.scaladsl
 import scala.collection.immutable
 import scala.concurrent.{ Future, Await }
 import scala.concurrent.duration._
-
 import akka.stream.ActorFlowMaterializer
 import akka.stream.ActorFlowMaterializerSettings
 import akka.stream.testkit.AkkaSpec
 import akka.stream.testkit.StreamTestKit.{ OnNext, SubscriberProbe }
 import akka.util.ByteString
 import akka.stream.{ Inlet, Outlet, Shape, Graph }
+import org.scalautils.ConversionCheckedTripleEquals
 
 object GraphOpsIntegrationSpec {
   import FlowGraph.Implicits._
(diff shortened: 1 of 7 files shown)


akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphOpsIntegrationSpec.scala
```diff
@@ -185,19 +185,12 @@ class GraphOpsIntegrationSpec extends AkkaSpec {
...
 
-      result.sorted should be(List(4, 5, 6, 13, 14, 15))
-
-      result.indexOf(4) < result.indexOf(5) should be(true)
-      result.indexOf(5) < result.indexOf(6) should be(true)
-
-      result.indexOf(13) < result.indexOf(14) should be(true)
-      result.indexOf(14) < result.indexOf(15) should be(true)
-
+      result.toSet should ===(Set(4, 5, 6, 13, 14, 15))
```
drewhk: exactly :)

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -323,6 +323,11 @@ object Source extends SourceApply {
    * Every connected `Sink` of this stream will see an individual stream consisting of one element.
    */
   def single[T](element: T): Source[T, Unit] = apply(SynchronousIterablePublisher(List(element), "single")) // FIXME optimize
+  
+  /**
+   * Create a `Source` that will continually emit the given element.
+   */
+  def constant[T](element: T): Source[T, Unit] = apply(() => Iterator.continually(element)) // FIXME optimize
```
drewhk: why not continually? constant does not express that it is repeated.

drewhk:
LGTM apart from the name "constant"

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
rkuhn: `continually` has at least in Scala collections the implication of repeated evaluation, while this is truly a constant value that is emitted. What about `repeat`?

drewhk: ok, repeat is fine

rkuhn:
Oh, the joys of inconsistent Java type inference … will fix.

ktoso:
LGTM

PR #16945, opened 2015-02-26
overhaul stream-design-rst after recent changes

rkuhn:


akka-docs-dev/rst/stream-design.rst
```diff
@@ -101,10 +100,10 @@ The finer points of stream materialization
...
+It might be tempting to allow different pieces of a flow topology to access the materialization results of other pieces in order to customize their behavior, but that would violate composability and reusability as argued above. Therefore the arguments and results of materialization need to be segregated:
 
-  * **Create:** first all execution units (Actors) are created, having access to the set of input parameters for the current materialization and producing key–value pairs that are placed in the MaterializedMap,
-  * **Resolve:** each flow element may declare derived keys that are calculated from other keys and added to the MaterializedMap; a derived key cannot depend on another derived key,
-  * **Initialize:** each flow element is finally initialized with the full MaterializedMap from the previous two phases; this will usually not do anything, but it allows certain elements to calculate their real behavior at this late stage.
+  * The FlowMaterializer is configured with a (type-safe) mapping from keys to values, which is exposed to the processing stages during their materialization.
```
drewhk: What do you mean by this? I don't understand here, it reads like the old MaterializedMap

rkuhn: yes, same principle, but used as (immutable) materialization input

rkuhn: the reasoning is that adding yet another type parameter to things will just blow the complexity budget

drewhk: Still, I cannot pars ethis sentence. Dissecting:

> The FlowMaterializer is configured with a (type-safe) mapping from keys to values

Configured where? What mapping? I don't understand :(

> which is exposed to the processing stages during their materialization.

What is exposed, where? I don't understand this either.

I feel this sentence is too abstract and high-level and it is hard to see what it intends to convey, although I know even the internals. Can't you reformulate this?

rkuhn: None of this exists already:
~~~scala
graph.run(mat.withMappings(...))
~~~
combined with access to the current FlowMaterializer during materialization (the API for which I have not yet figured out; could be as horrible as ThreadLocals, who knows).

drewhk: aaah. Isn't it confusing to speculate about these, even though I know this is a design document?

PR #16949, opened 2015-02-26
Refactor akka.http.server.RejectionHandler

### LINKED ISSUE #16835: Improve writing of rejection handlers
From spray/spray#368:

If we use simple pattern matching the order of the rejections in the rejections list matters more than it should.
Currently the logic is: Apply the rejection conversion logic to the whole list at once, which in most cases will simply produce the result for the *first* rejection in the list.

Ideally, we would transpose this behavior, i.e. define rejection handling logic as a `List[PartialFunction[Rejection, Route]]` whose elements get applied to the rejection list element-by-element. The first "hit" wins. This would mean that the list of conversions would define the priority of the rejections, not the order of the rejections in the list.

See also this thread:
https://groups.google.com/d/msg/spray-user/gdXGSC3wwZI/FLhcys4nEh4J

/cc @sirthias

sirthias:
Closes #16835.

### DIFF
--- a/akka-http/src/main/scala/akka/http/server/RejectionHandler.scala
+++ b/akka-http/src/main/scala/akka/http/server/RejectionHandler.scala
@@ -4,6 +4,8 @@
 
 package akka.http.server
 
+import scala.annotation.tailrec
+import scala.reflect.ClassTag
 import scala.collection.immutable
 import scala.concurrent.ExecutionContext
 import akka.http.model._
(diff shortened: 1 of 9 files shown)


akka-http/src/main/scala/akka/http/server/RejectionHandler.scala
```diff
@@ -12,113 +14,202 @@ import headers._
...
+  /**
+   * Creates a new [[RejectionHandler]] which uses the given one as fallback for this one.
+   */
+  def withFallback(that: RejectionHandler): RejectionHandler =
+    (this, that) match {
+      case (a: BuiltRejectionHandler, b: BuiltRejectionHandler) ⇒
+        new BuiltRejectionHandler(a.cases ++ b.cases, a.notFound orElse b.notFound, a.isSealed || b.isSealed)
+      case _ ⇒ new RejectionHandler {
+        def apply(rejections: immutable.Seq[Rejection]): Option[Route] =
+          this(rejections) orElse that(rejections)
```
rkuhn: wrong `this`, I think?

rkuhn:
Structure LGTM, didn’t check every single rejection case.

akka-http/src/main/scala/akka/http/server/RejectionHandler.scala
sirthias: Indeed. Nice catch!

sirthias:
Fixed bad `this` as detected by @rkuhn's radar.

jrudolph:
I like the new DSL. The rejection handlers are now more explicit than before (and the DSL could be extended easily if necessary).

I cannot oversee the complete impact of this change but the general change to decide handling based on the order of handlers instead of the order of rejections (= position in the route tree) seems to be a good one. It's still possible to customize the behavior (e.g. to restore the old behavior while migrating from spray). So, :+1: from me.

akka-http/src/main/scala/akka/http/server/RejectionHandler.scala
```diff
@@ -130,4 +221,6 @@ object RejectionHandler {
       case (remaining, transformation) ⇒ transformation.transform(remaining)
     }
   }
+
+  private object PrivateRejection extends Rejection
```
2beaucoup: What's the purpose of this indirection?

sirthias:
@jrudolph Yes. My goal here was to leave the main outer API of a `RejectionHandler` pretty much unchanged. It was `PartialFunction[immutable.Seq[Rejection], Route]` before and now it's `immutable.Seq[Rejection] ⇒ Option[Route]`, so no big change.
Also, the interface is open for custom extension, so you can definitely do the same thing as before pretty easily. 

Still, this change probably means that some large applications with complicated route structures now produce different error messages in the case of multiple rejections. Now the precedence between rejections is clearly defined, whereas before it was hard to predict.
So, while this change might break some tests here and there, it's overall a good thing.

akka-http/src/main/scala/akka/http/server/RejectionHandler.scala
sirthias: I use this private rejection for a naive test of whether a given handler has a catchall case.
This test is not perfect, because it doesn't correctly classify things like this:

```scala
handle {
  case x if !x.isInstanceOf[MissingCookieRejection] ⇒ complete(...)
}
```

but it should be good enough for the very large majority of cases.
If I have a really "strange" custom `RejectionHandler` I can still say `myHandler withFallBack default` to seal it manually, so this simple implementation is good enough IMO.
```

2beaucoup: Ahh, Ok. `isDefinedAt` needs an instance. ;)

rkuhn:
I agree with the reasoning and intention, please add this to the list of migration notes from Spray to Akka HTTP (which I’m sure you have begun already, right? ;-) ).

PR #16954, opened 2015-02-26
=htc #16953 Fix handling of empty hosts in Uris

### LINKED ISSUE #16953: Uri Authority instances with empty host but non-empty port or user-info not handled correctly
We have a bug in our handling of empty hosts.
These should all pass but currently some of them fail:

```scala
Uri("http://:8000/foo") shouldEqual Uri("http", Authority(Host.Empty, 8000), Path / "foo")
Uri("http://:80/foo") shouldEqual Uri("http", Authority.Empty, Path / "foo")
normalize("https://:443") shouldEqual "https:"
normalize("https://:444") shouldEqual "https://:444"
normalize("http://:80/foo") shouldEqual "http:/foo"
```

The fix needs backporting to spray.

/cc @jrudolph

sirthias:
Closes #16953.

/cc @jrudolph

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/model/UriSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/model/UriSpec.scala
@@ -16,6 +16,10 @@ class UriSpec extends WordSpec with Matchers {
 
   "Uri.Host instances" should {
 
+    "correctly parse empty hosts" in {
+      Host("") shouldEqual Host.Empty
+    }
+
     "parse correctly from IPv4 literals" in {
       Host("192.0.2.16") shouldEqual IPv4Host("192.0.2.16")
       Host("255.0.0.0") shouldEqual IPv4Host("255.0.0.0")
@@ -25,6 +29,7 @@ class UriSpec extends WordSpec with Matchers {
       Host("3.0.0.0") shouldEqual IPv4Host("3.0.0.0")
       Host("30.0.0.0") shouldEqual IPv4Host("30.0.0.0")
     }
+
     "support inetAddresses round-trip for Inet4Addresses" in {
       def roundTrip(ip: String): Unit = {
         val inetAddr = InetAddress.getByName(ip)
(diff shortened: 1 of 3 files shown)


akka-http-core/src/test/scala/akka/http/model/UriSpec.scala
```diff
@@ -406,7 +418,7 @@ class UriSpec extends WordSpec with Matchers {
       normalize("http://sourceforge.net/projects/uriparser/") shouldEqual "http://sourceforge.net/projects/uriparser/"
       normalize("http://sourceforge.net/project/platformdownload.php?group_id=182840") shouldEqual "http://sourceforge.net/project/platformdownload.php?group_id=182840"
       normalize("mailto:test@example.com") shouldEqual "mailto:test@example.com"
-      normalize("file:///bin/bash") shouldEqual "file:///bin/bash"
+      normalize("file:///bin/bash") shouldEqual "file:/bin/bash"
```
jrudolph: Is this change correct for any scheme? Browsers seem to do the conversion in the other direction for the `file` scheme. The [normalization paragraph of RFC 3986](http://tools.ietf.org/html/rfc3986#section-6.2.3) doesn't seem to specify a general normalization rule of this kind but rather:

> Normalization should not remove delimiters when their associated component is empty unless licensed to do so by the scheme specification.

sirthias: URIs with a scheme but no authority are pretty rare, so looking for examples isn't as easy.
I'm not sure what the "scheme specification" for HTTP has to say about URIs with an empty authority. It's appears to be another unspeced edge case.
Semantically there doesn't appear to be a difference between `http:///foo` and `http:/foo`, so I'd day normalizing the first to the latter is ok.
Since in order to be able to tell these two variants apart we'd have to introduce another special case to our model I'm inclined to simply go with this simpler version and wait for someone to come up with a real-world case where this poses a problem. Then we can go back and decide whether the complication is indeed justified.

rkuhn:
It LGTM, have you guys reached a final verdict?

akka-http-core/src/test/scala/akka/http/model/UriSpec.scala
sirthias: One more thought to maybe settle this issue:
We could change the rendering logic to simply format URIs with a *scheme* but an empty authority to always show the `//` delimiter. So `http:///foo` and `http:/foo` would still have the identical representation but always render to the former rather than the latter.
Maybe that's what you meant in the first place.

jrudolph: That would probably settle it for now. For http it isn't really a question, it's more the question how to handle other schemes and for that I agree that revisiting the decision once use-cases come up would be good enough.

rkuhn: sounds reasonable to me

sirthias:
Ok, changed rendering logic to include the `//` delimiter between a non-empty scheme and an empty authority as discussed.

Issue #16934, opened 2015-02-26
Change ActorFlowMaterializer arguments to FlowMaterializer where possible


drewhk:
Many API methods now take an ActorFlowMaterializer but most of them are actually happy with any FlowMaterializer.

bantonsson:
Fixed by #16976

PR #16976, opened 2015-03-02
HTTP: s/ActorFlowMaterializer/FlowMaterializer where possible

sirthias:
refs #16934.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/Http.scala
@@ -13,11 +13,11 @@ import scala.concurrent.Future
 import akka.event.LoggingAdapter
 import akka.util.ByteString
 import akka.io.Inet
-import akka.stream.ActorFlowMaterializer
+import akka.stream.FlowMaterializer
 import akka.stream.scaladsl._
 import akka.http.engine.client.{ HttpClient, ClientConnectionSettings }
 import akka.http.engine.server.{ HttpServer, ServerSettings }
-import akka.http.model.{ ErrorInfo, HttpResponse, HttpRequest }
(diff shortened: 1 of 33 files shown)


drewhk:
LGTM

sirthias:
Looks like this stream test needs to be relaxed a bit more:

```
- must fairly balance between three outputs *** FAILED *** (525 milliseconds)
[info]   7953 was not 10000 plus or minus 2000 (GraphBalanceSpec.scala:139)
```

bantonsson:
I think that behavior is a bug (unrelated to your changes of course) but the test says `waitForAllDownstreams = true`, and it clearly doesn't. Opened a ticket for that #16978

drewhk:
well, strictly speaking the violation of that spec does not mean that it does not wait for all downstreams (it is nondeterministic), but that is one likely explanation.

bantonsson:
@drewhk, `enqueueAndYield` does a `+1` to pick the next port, which I think is a highly deterministic RoundRobin algorithm if all downstreams have connected when it starts. :wink:

drewhk:
No, it is non-deterministic because it picks from the currently available set. Yes, it starts from +1, but it can wrap around and end up in the original position. The test case punches through 10000 elements, while the buffer size is by default 16 AFAIK, so there is plenty of room for non-deterministic races.

bantonsson:
Yes, I know, but since the buffer is the same for all the downstreams I don't think that the spread should be that big, but who knows. Maybe one of the consumers can't keep up with adding the elements.

drewhk:
Maybe we are affected by non-fairness in dispatcher and actor scheduling, who knows. We need to look into it.

bantonsson:
LGTM

Issue #15766, opened 2014-08-28
Akka TCP on server side might not signal Closed events if client goes away silently


drewhk:
Reported by user on ML: https://groups.google.com/forum/#!topic/akka-user/WdXWjcnVWiQ

It seems like that ServerChannel might not signal close in certain cases if nothing is written to the socket. We should try to reproduce the issue and see what we can do about it if it is reproducible.

bantonsson:
This needs to stay on the 2.3.7 milestone since it was released there.

The better workaround has its own ticket #17000

drewhk:
Superceded by ticket: https://github.com/akka/akka/issues/17122

PR #16984, opened 2015-03-03
NOT FOR MERGE ActorPublisher used for completing requests does not get Cancel message when downstream cancels

2m:
_Currently not for merging. Contains failing test-case._

We have an akka-http server use-case where request is completed with `Source` which is backed up by `ActorPublisher` which `context.stop` itself upon receive of `ActorPublisherMessage.Cancel`.

Problem we are seeing is: when client terminates the connection, `Cancel` is not immediately received by the `ActorPublisher`. `Cancel` message is only received when there is a new message sent from the upstream.

This PR contains one failing and one passing test case for such http server. Also I have added a test case for `ActorPublisher` which validates that `ActorPublisher` when used standalone receives `Cancel` message immediately when downstream cancels.

We have seen this at least since `1.0-M3`.

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/actor/ActorPublisherSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/actor/ActorPublisherSpec.scala
@@ -187,6 +187,19 @@ class ActorPublisherSpec extends AkkaSpec with ImplicitSender {
       s.expectNoMsg(300.millis)
     }
 
+    "cancel without receiving new elements" in {
+      val ref = watch(system.actorOf(senderProps))
+      val p = ActorPublisher[Int](ref)
+      val s = StreamTestKit.SubscriberProbe[Int]()
+      p.subscribe(s)
+      val sub = s.expectSubscription
+      sub.request(1)
+      ref ! 1
+      s.expectNext(1)
+      sub.cancel()
+      expectTerminated(ref)
+    }
+
     "remember requested after restart" in {
       val probe = TestProbe()
       val ref = system.actorOf(testPublisherProps(probe.ref))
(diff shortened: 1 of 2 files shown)


huntc:
@rkuhn FYI this one is a bit of a blocker for us sans determining a work-around.

drewhk:
Is this still a problem?

huntc:
> Is this still a problem?

I believe so. We've worked around the issue though by having the server send heartbeats. If it sends a heartbeat it will then close on the server side correctly if the client has closed.

drewhk:
This is a known issue. NIO is not very friendly to us in this case. This is related to : https://github.com/akka/akka/issues/16552 and https://github.com/akka/akka/issues/15766. I will close this PR if no one objects because it does not add anything new as far as I see. If I am wrong, just tell me here :)

huntc:
I'll defer to @2m on that.

akka-stream-tests/src/test/scala/akka/stream/actor/ActorPublisherSpec.scala
```diff
@@ -187,6 +187,19 @@ class ActorPublisherSpec extends AkkaSpec with ImplicitSender {
...
+      val ref = watch(system.actorOf(senderProps))
+      val p = ActorPublisher[Int](ref)
+      val s = StreamTestKit.SubscriberProbe[Int]()
+      p.subscribe(s)
+      val sub = s.expectSubscription
+      sub.request(1)
+      ref ! 1
+      s.expectNext(1)
+      sub.cancel()
+      expectTerminated(ref)
```
drewhk: this one fails too?

2m: No. This one passes. I added this while searching if it is network related or not. I probably should not have committed it.

2m:
This can be used as an inspiration for tests when fixing tickets mentioned above. I'll close this then.

Issue #16983, opened 2015-03-03
ActorInterpreter should yield to allow cancel even though there is demand


patriknw:
e.g. Source(infinite iterator) and request of many elements followed by cancel should stop the stream.

In RS-TCK RC3 IterablePublisherTest required_spec317_mustNotSignalOnErrorWhenPendingAboveLongMaxValue the actor system cannot be shutdown because of this. The OneBoundedInterpreter is spinning when requested elements is Int.MaxValue

drewhk:
The simple solution is just to implement the yielding at the boundaryStages of ActorInterpreter (probably at the bottom stage is enough for a simple fix, the upper stage is bounded by the buffer size). Since we do not have fusing right now, it is enough. Once we restore fusing we must do the proper yielding, which is able to suspend the interpreter returning a handle that has a resume() method (ctx.enter at the boundaries MUST NOT happen to a suspended interpreter). The hardest part would be to make this work with the forking, and also take care of overflow-to-heap fork markers.

drewhk:
Also, don't forget to swap the two checks at the output boundary -- cancellation check should come first, not second.

Issue #16986, opened 2015-03-03
stream cancellation leaks actors


rkuhn:
Currently ActorInterpreter will not shut down after receiving cancellation from downstream (it will pass on the cancel signal though). Fixing this will mean that OnSubscribe events can go to dead letters, which will mean that upstream flows will just hang around unless we introduce a timeout. We need to fix all angles to this—belt and suspenders.

This is possibly also related to #16900.

drewhk:
Why is this so hard to fix? Isn't this then also a problem with ordinary ActorProcessor? Or if not, why is it harder to fix it here?

rkuhn:
Yes, this would affect all Actors that we create. Your proposal of just keeping them alive until the `OnSubscribe` comes in would take care of this issue, I’ll try it out and close this issue if successful.

rkuhn:
Fixed for ActorInterpreter, still need to look at other ActorProcessors.

patriknw:
yes, the new TCK detected this for PrefixAndTail

drewhk:
Hasn't this been fixed already?

patriknw:
I don't think so. My workaround (// FIXME  unreference real subscriber, should not be needed after #16986) in StreamOfStreamProcessors.scala is still there and the tck-tests (required_spec313_cancelMustMakeThePublisherEventuallyDropAllReferencesToTheSubscriber) fail without it.

drewhk:
What is the issue there? I don't remember the details.

patriknw:
I don't remember the details. :(

drewhk:
also https://github.com/akka/akka/issues/15191

rkuhn:
many custom processors will be removed in the wake of #17090

patriknw:
Solved the issue with stream of streams in #17212, but that was not an actor leak. So more to do here.

PR #16987, opened 2015-03-03
!str #16983 add outputBurstLimit

rkuhn:
This limits the number of elements that the ActorInterpreter will allow
the OneBoundedInterpreter to emit during one message processing.


akka-stream-tests/src/test/scala/akka/stream/impl/ActorInterpreterSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  "An ActorInterpreter" must {
+
+    "pass along early cancellation" in new Setup {
+      processor.subscribe(down)
+      val sub = down.expectSubscription()
+      sub.cancel()
+      up.subscribe(processor)
+      val upsub = up.expectSubscription()
+      upsub.expectCancellation()
```
rkuhn: This currently fails, because I fixed ActorInterpreter to actually shut down upon cancellation, and then there is nobody to generate the upstream’s `cancel()` call. This is the reason for #16986.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorInterpreter.scala
```diff
@@ -82,7 +82,7 @@ private[akka] class BatchingActorInputBoundary(val size: Int)
 
   override def onDownstreamFinish(ctx: BoundaryContext): TerminationDirective = {
     cancel()
-    ctx.exit()
+    ctx.finish()
```
rkuhn: This is the actual actor termination fix.

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -178,6 +178,23 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]], val forkLimit:
 
   private var lastOpFailing: Int = -1
 
+  private def pipeName(op: UntypedOp): String = {
```
drewhk: nice!

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorInterpreter.scala
```diff
@@ -191,12 +226,9 @@ private[akka] class ActorOutputBoundary(val actor: ActorRef, debugLogging: Boole
 
   override def onPush(elem: Any, ctx: BoundaryContext): Directive = {
     onNext(elem)
-    if (downstreamDemand > 0) ctx.pull()
+    if (downstreamDemand > 0) tryBounceBall(ctx)
     else if (downstreamCompleted) ctx.finish()
```
drewhk: swap these, check for completion first. This is cancellation and it is out-of-band.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorInterpreter.scala
drewhk: yep, good catch!

drewhk:
LGTM, only the downsteamCompleted check needs to be swapped with the demand check.

akka-stream-tests/src/test/scala/akka/stream/impl/ActorInterpreterSpec.scala
rkuhn: has been fixed by keeping the Actor around until the upstream subscription has been established

akka-stream/src/main/scala/akka/stream/impl/ActorPublisher.scala
```diff
@@ -64,8 +64,7 @@ private[akka] class ActorPublisher[T](val impl: ActorRef) extends Publisher[T] {
 
   def takePendingSubscribers(): immutable.Seq[Subscriber[_ >: T]] = {
     val pending = pendingSubscribers.getAndSet(Nil)
-    assert(pending ne null, "takePendingSubscribers must not be called after shutdown")
-    pending.reverse
+    if (pending eq null) Nil else pending.reverse
```
rkuhn: Please ponder this one: I cannot currently explain why this patch changed the behavior, and I also cannot explain why this previously worked with the assertion.

rkuhn:
closes #16983

akka-stream-tests/src/test/scala/akka/stream/impl/ActorInterpreterSpec.scala
```diff
@@ -0,0 +1,165 @@
...
+      new Setup(infinite :: Nil) {
+        processor.subscribe(down)
+        val sub = down.expectSubscription()
+        up.subscribe(processor)
+        val upsub = up.expectSubscription()
+        sub.request(100000000)
+        sub.cancel()
+        watch(actor)
+        latch.countDown()
+        for (i ← 1 to N) withClue(s"iteration $i: ") {
```
patriknw: I had no clue about that one, nice

akka-stream-tests/src/test/scala/akka/stream/impl/ActorInterpreterSpec.scala
```diff
@@ -0,0 +1,165 @@
...
+          try down.expectNext(42) catch { case e: Throwable ⇒ fail(e) }
+        }
```
patriknw: here it would be great with a comment saying something like:
// now the cancellation signal is processed and no more elements are pushed

akka-stream-tests/src/test/scala/akka/stream/impl/ActorInterpreterSpec.scala
```diff
@@ -0,0 +1,165 @@
...
+        down.expectComplete()
+        upsub.expectCancellation()
+        expectTerminated(actor)
+      }
+    }
+
+  }
+
+}
```
patriknw: you should turn on "Add new line at end of file" in Eclipse Scala Save Actions

akka-stream-tests/src/test/scala/akka/stream/impl/ActorInterpreterSpec.scala
rkuhn: yes, the only ugly part is that it only applies to test failures, not assertion errors, which is why I have to catch them

akka-stream-tests/src/test/scala/akka/stream/impl/ActorInterpreterSpec.scala
rkuhn: yup

akka-stream-tests/src/test/scala/akka/stream/impl/ActorInterpreterSpec.scala
rkuhn: you mean: “you should turn it on in _that_ workspace as well” ;-)

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorInterpreter.scala
```diff
@@ -142,18 +144,25 @@ private[akka] class BatchingActorInputBoundary(val size: Int)
...
-  private def completed: Actor.Receive = {
-    case OnSubscribe(subscription) ⇒ throw new IllegalStateException("Cannot subscribe shutdown subscriber")
-  }
+}
 
+private[akka] object ActorOutputBoundary {
+  /**
+   * INTERNAL API.
+   */
+  private[akka] case object ContinuePulling extends DeadLetterSuppression
```
patriknw: isn't this only used in ActorOutputBoundary, i.e. it can be `private`?

rkuhn: right; I guess I was sloppy because it doesn’t really matter anyway—I’m still disappointed that for the JVM all of these choices mean `public` :-(

akka-stream/src/main/scala/akka/stream/impl/ActorPublisher.scala
patriknw: I think this is safe. The reason is when `exposedPublisher.shutdown` and `context.stop` is performed. I guess the patch delayed `context.stop`. Anyway, this should be good.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorInterpreter.scala
patriknw: you save 3 lines of comments ;-)

patriknw:
LGTM

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorInterpreter.scala
patriknw: ah, you mean that it is actually java public, and the comment is needed, that is too bad

bantonsson:
LGTM

drewhk:
LGTM, but open a ticket that this needs to be solved in a fundamental fashion (make the interpreter suspendable) because we will need that once we start to fuse things again.

rkuhn:
done, see #17025

Issue #16994, opened 2015-03-04
make materialized value of Flow.join consistent with all other combinators


rkuhn:
Instead of Keep.both it should default to Keep.left.

drewhk:
Has this been done?

Issue #16416, opened 2014-11-27
add BidirectionalFlow


rkuhn:
This is the piece that has two inputs and two outputs, allowing the modeling of an arbitrary protocol layer on top of a Source–Sink pair (a.k.a. Flow):
~~~scala
trait BidirectionalFlow[In1, Out1, In2, Out2] {
  def join[T, U](other: BidirectionalFlow[Out1, T, U, In2]): BidirectionalFlow[In1, T, U, Out2]
  def join(flow: Flow[Out1, In2]): Flow[In1, Out2]
  // plus matching method on FlowOps for the left side
}
~~~
The “flows” 1 and 2 are going in opposite directions, 1 being on the top (left to right) and 2 being on the bottom (right to left).

reid-spencer:
:+1: This would be helpful for me.

drewhk:
This can be done immediately after https://github.com/akka/akka/issues/16902

drewhk:
This can be closed now right?

PR #16995, opened 2015-03-04
!str #16416,#16994 BidiFlow DSL and Flow.join Mat

rkuhn:
- add BidiFlow, with atop and join combinators
- add Flow.join(BidiFlow)
- correct Flow.join’s default materialized value selection to Keep.left
- [x] add factory methods
- [x] add javadsl
- [x] add tests
- [x] add docs

References #16416 #16994

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/BidiFlow.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/BidiFlow.scala
@@ -1,0 +1,142 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.scaladsl
+
+import akka.stream.Graph
+import akka.stream.BidiShape
+import akka.stream.impl.StreamLayout.Module
+import akka.stream.FlowShape
+
+final class BidiFlow[-I1, +O1, -I2, +O2, +Mat](private[stream] override val module: Module) extends Graph[BidiShape[I1, O1, I2, O2], Mat] {
+  override val shape = module.shape.asInstanceOf[BidiShape[I1, O1, I2, O2]]
+
+  /**
+   * Add the given BidiFlow as the next step in a bidirectional transformation
+   * pipeline. By convention protocol stacks are growing to the left: the right most is the bottom
+   * layer, the closest to the metal.
+   * {{{
+   *     +----------------------------+
+   *     | Resulting BidiFlow         |
+   *     |                            |
+   *     |  +------+        +------+  |
+   * I1 ~~> |      |  ~O1~> |      | ~~> OO1
+   *     |  | this |        | bidi |  |
+   * O2 <~~ |      | <~I2~  |      | <~~ II2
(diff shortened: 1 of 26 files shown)


akka-stream/src/main/scala/akka/stream/scaladsl/BidiFlow.scala
```diff
@@ -0,0 +1,140 @@
...
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.scaladsl
+
+import akka.stream.Graph
+import akka.stream.BidiShape
+import akka.stream.impl.StreamLayout.Module
+import akka.stream.FlowShape
+
+final class BidiFlow[-I1, +O1, -I2, +O2, +Mat](private[stream] override val module: Module) extends Graph[BidiShape[I1, O1, I2, O2], Mat] {
```
ktoso: uuu, it's here finally :-)

patriknw:
nice

bantonsson:
Very nice indeed. ASCII art FTW!

rkuhn:
yeah, I was a little bored while de- and re-registering my car ;-)

akka-stream/src/main/scala/akka/stream/scaladsl/BidiFlow.scala
drewhk: We probably need a factory from a Graph[BidiShape, _]. The same should be added to Flow, Source and Sink (maybe RunnableFlow, too, but that is less useful).

drewhk:
LGTM, after the factory is added.

akka-stream/src/main/scala/akka/stream/scaladsl/BidiFlow.scala
rkuhn: Good point, ISTR adding such a `wrap` method to `Flow`, but it should be on all of them.

rkuhn:
test & docs added, also for Java, please review

akka-stream/src/main/scala/akka/stream/scaladsl/JavaConverters.scala
```diff
@@ -17,6 +17,9 @@ private[akka] object JavaConverters {
   implicit final class AddAsJavaFlow[In, Out, Mat](val flow: scaladsl.Flow[In, Out, Mat]) extends AnyVal {
     def asJava: javadsl.Flow[In, Out, Mat] = new javadsl.Flow(flow)
   }
+  implicit final class AddAsJavaBidiFlow[I1, O1, I2, O2, Mat](val flow: scaladsl.BidiFlow[I1, O1, I2, O2, Mat]) extends AnyVal {
+    def asJava: javadsl.BidiFlow[I1, O1, I2, O2, Mat] = new javadsl.BidiFlow(flow)
+  }
```
bantonsson: @patriknw has removed the implicit converters here #17019

bantonsson:
LGTM!

The validation failure is just `Sink.head()` and friends loosing their parenthesis.

PR #17001, opened 2015-03-05
!htc #16988 hide model.parser.HeaderParser from public API, add proper `HeaderParser.parse`

### LINKED ISSUE #16988: Parsing RawHeaders with names containing header separator
`RawHeader` parsing should fail when the header name contains a sequence of a colon and a space (`": "`). Currently parsing returns an invalid result with the `": "` characters moving into the `RawHeader` value field. Instead of returning an invalid result, a parse error should be returned.

An example test case:
```scala
HeaderParser.parseHeader(RawHeader("X: ", "a")) shouldEqual Left(ErrorInfo("Illegal header `X: ` contains invalid character: `:`"))
```

This test fails with:
```
Right(X: : a) did not equal Left(ErrorInfo(Illegal header `X: ` contains invalid character: `:`,))
```

sirthias:
Closes #16988.

The previously used `model.parser.HeaderParser.parseHeader(s)` was no proper user-facing API.
This PR fixes this by removing outside access to the `model.parser.HeaderParser` completely and providing a clean API for parsing a String/String pair into a header model instance.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/engine/parsing/HttpHeaderParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/engine/parsing/HttpHeaderParser.scala
@@ -59,7 +59,7 @@ import akka.http.model.parser.CharacterClasses._
  */
 private[engine] final class HttpHeaderParser private (
   val settings: HttpHeaderParser.Settings,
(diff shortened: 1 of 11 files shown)


rkuhn:
LGTM, @jrudolph ?

jrudolph:
LGTM

Issue #16992, opened 2015-03-04
Source.apply(f, name) doesn't use name parameter


hseeberger:
Taken from `Source.scala` of 1.0-M4: `name` parameter is not used

``` scala
  def apply[T](f: () ⇒ Iterator[T], name: String): Source[T, Unit] = {
    apply(new immutable.Iterable[T] {
      override def iterator: Iterator[T] = f()
    })
  }
```

bantonsson:
Thanks @hseeberger

Issue #15707, opened 2014-08-21
Revisit stream actor naming and hierarchy


patriknw:
Naming and hierarchy was changed here: https://github.com/akka/akka/pull/15701
Tests were removed there, might be useful to bring them back when the proper naming and hierarch is in place.

rkuhn:
Discussion of the supervision semantics resulted in the conclusion that the current way is very likely the right one, but we should still revisit the actor naming to make sure that it is good.

drewhk:
Naming should be probably reflecting the nested nature of the streams blueprint, see https://github.com/akka/akka/issues/16902

patriknw:
note that default names of sources and sinks should be named correctly, now they contain "unknown-operator"

drewhk:
throw an exception where this default name is used, and see where it fails. This might work.

PR #17005, opened 2015-03-05
!str #16992 Rework Source and Sink name parameter

patriknw:
I don't like this, but given that we really need the name parameter for the naming of the shape I found no other way.

`onComplete` javadsl signature was wrong

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
@@ -14,7 +14,10 @@ import org.reactivestreams.{ Publisher, Subscriber, Subscription }
 import scala.annotation.unchecked.uncheckedVariance
 import scala.concurrent.{ Future, Promise }
 
-abstract class SinkModule[-In, Mat](val shape: SinkShape[In]) extends Module {
+/**
+ * INTERNAL API
+ */
+private[akka] abstract class SinkModule[-In, Mat](val shape: SinkShape[In]) extends Module {
 
   def create(materializer: ActorFlowMaterializerImpl, flowName: String): (Subscriber[In] @uncheckedVariance, Mat)
 
@@ -31,15 +34,24 @@ abstract class SinkModule[-In, Mat](val shape: SinkShape[In]) extends Module {
   }
 
   override def subModules: Set[Module] = Set.empty
+
+  def amendShape(attr: OperationAttributes): SinkShape[In] = {
+    attr.nameOption match {
+      case None ⇒ shape
+      case s: Some[String] if s == attributes.nameOption ⇒ shape
+      case Some(name) ⇒ shape.copy(inlet = new Inlet(name + ".in"))
+    }
+  }
 }
 
 /**
+ * INTERNAL API
  * Holds the downstream-most [[org.reactivestreams.Publisher]] interface of the materialized flow.
  * The stream will not have any subscribers attached at this point, which means that after prefetching
  * elements to fill the internal buffers it will assert back-pressure until
  * a subscriber connects and creates demand for elements to be emitted.
  */
-class PublisherSink[In](val attributes: OperationAttributes, shape: SinkShape[In]) extends SinkModule[In, Publisher[In]](shape) {
+private[akka] class PublisherSink[In](val attributes: OperationAttributes, shape: SinkShape[In]) extends SinkModule[In, Publisher[In]](shape) {
 
   override def toString: String = "PublisherSink"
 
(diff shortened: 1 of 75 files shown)


rkuhn:
Just skimmed it for now: did you check that there is no method without parens anymore? Unfortunately this seems to make a difference when it comes to inferred type arguments and overloading, so we should do that consistently as well. An example of what I mean is that `.runWith(Sink.head)` does not work anymore without parens.

patriknw:
@rkuhn `runWith(Sink.head)` actually works if I declare it as this:

    def head[T]: Sink[T, Future[T]]
    def head[T](name: String): Sink[T, Future[T]]

but then `runWith(Sink.head())` doesn't work

Otherwise we can declare it as 

    def head[T](name: String = ""): Sink[T, Future[T]]

and then only `runWith(Sink.head())` will work, and the parens will also be required for `val futureSink = Sink.head[Seq[Int]]()`

I have already used the default parameter value approach at places where there are more parameters.

WDYT?

bantonsson:
I think that we should go with `Sink.head()`(and the same for cancelled), or if we don't then all other with `()` at the end (like ignore) should drop theirs as well.

rkuhn:
I’d recommend to stay away from default arguments as they are not extensible in a binary compatible fashion, and source compatibility can also be challenging. So, thanks for your research, I think using overloaded methods with at least one argument list each is the way to go.

patriknw:
The drawback with the overload is the copies of the docs.

but if we use overloads anyway, why not use the variant without argument list to support `Sink.head`?

rkuhn:
If that works in all cases, then yes. I ran into some issue that might well have involved another bug on my part. The problem is if Scalac can’t figure that one out there is nothing you can write to disambiguate—parens are then not allowed.

patriknw:
ok, I'll try it in large scale

patriknw:
@rkuhn one surprise when I removed default arguments.
Changing:

    def foreach[T](f: T ⇒ Unit, name: String = ""): Sink[T, Future[Unit]]

to:

    def foreach[T](f: T ⇒ Unit, name: String): Sink[T, Future[Unit]]
    def foreach[T](f: T ⇒ Unit): Sink[T, Future[Unit]]

Then compilation error of for example:

    val echoHandler = Sink.foreach[StreamTcp.IncomingConnection] { _.flow.join(Flow[ByteString]).run() }


    missing parameter type for expanded function ((x$9) ⇒ x$9.flow.join(Flow[ByteString]).run())

rkuhn:
Hmm, yes, that only works with default arguments … aargh. Overloading is evil, all kinds of it. Maybe a better approach is to add a `.named` method on Sink and thereby completely factor out the naming aspect?

patriknw:
I can interpret `.named` in several ways, one is `Sink.named("blah").head`

What is the name used for in the shape? Do we really need to make that configurable? Is it possible to grab it afterwards from the `OperationAttribtues`?

rkuhn:
The point is to _set_ the OperationAttributes of this (Sink) stage. I was thinking more like
~~~scala
Sink.head[Int].named("TheHead")
~~~

rkuhn:
To add more color: the inlet of this sink should then have `"TheHead"` in its name, so that when you forget to wire it up you’ll get a nice and explanative error message.

patriknw:
that sounds more in line with the existing `Sink.head[Int].withAttributes`, I will give it a try

patriknw:
@rkuhn I have reworked this completely according to our discussion. Much better.

* Remove name parameter (no overloads), naming is performed using `.withAttributes` or the new
  convenience `.named`. Those adds the OperationAttribute.Name and also change the name of the
  shape Inlet and Outlet.

* Remove Source/Sink parameter list for 0 parameter methods,
  this allows usage of `Sink.head` instead of `Sink.head()`

bantonsson:
LGTM

One question though. What name will an already named source (like `IterableSource`) have if you call `named(...)` on it? are the operational attributes overriden?

patriknw:
I assume last wins. We have an additional ticket for revisiting the actual naming: #15707

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -31,15 +34,24 @@ abstract class SinkModule[-In, Mat](val shape: SinkShape[In]) extends Module {
   }
 
   override def subModules: Set[Module] = Set.empty
+
+  def amendShape(attr: OperationAttributes): SinkShape[In] = {
```
drewhk: why is this needed to be public?

drewhk: stupid question, this is the Module.

patriknw: well, it could be restricted to `protected`

drewhk:
It is not exactly last wins, it is hierarchic. So if you name an iterable it will have three names, the user given, inside that the name "iterable" and inside that the names "emptySource" and "stageFactory". How we calculate a display name from that is up to us. When I implemented toString on Source/Sink/Flow they always used the outermost name. A deepToString would actually show the innermost names and the hierarchic nesting (maybe showing the name hierarchy, too). When we name actors we take the whole sequence AFAIK ("myName-iterable-stageFactory").

patriknw:
Refs #16992

Issue #16866, opened 2015-02-12
Fold, Foreach, etc. should fail the exposed Future if the passed function throws


RichardBradley:
The following test fails with a timeout, but I think it should pass.
I am using akka-http-experimental 1.0-M3.

```
import akka.actor.ActorSystem
import akka.stream.ActorFlowMaterializer
import akka.stream.scaladsl.{Sink, Source}
import org.specs2.mutable.Specification
import scala.concurrent.Future

class ExceptionDuringFoldSpec
  extends Specification {

  private implicit val system = ActorSystem()
  private implicit val fm = ActorFlowMaterializer()

  "an exception during a Flow fold" should {
    "result in a failed Future containing the exception" in {

      val output: Future[List[Int]] = Source(List(1,2,3,4))
        .runWith(Sink.fold(List[Int]()) { (xs, x) =>

        if (x == 3) throw new Exception("test exception")

        xs :+ x
      })

      output must throwAn[Exception].like {
        case e => e.getMessage mustEqual "test exception"
      }.await()
    }
  }
}
```

drewhk:
I confirm. It fails properly when an upstream element fails, but not if the folding function fails.

drewhk:
Btw, this is not a problem for Map, Filter, etc. because they have a downstream to propagate to the error, but here fold is a Sink and the only thing to propagate to is the Promise. The following code:

        override def onPush(elem: T, ctx: Context[U]): Directive = {
          aggregator = f(aggregator, elem)
          ctx.pull()
        }

should be something like 

        override def onPush(elem: T, ctx: Context[U]): Directive = {
          try { 
             aggregator = f(aggregator, elem)
             ctx.pull()
          } catch { case e => promise.tryFailure(e); ctx.fail(e) } 
        }

drewhk:
Same is probably ture for Foreach

viktorklang:
@drewhk More like:

~~~scala
catch { case NonFatal(e) => promise.tryFailure(e); ctx.fail(e) }
~~~

drewhk:
the "should be something like" meant to imply that :)

Issue #16916, opened 2015-02-20
add stream supervision for Sinks and Sources


patriknw:
e.g. iterator source and fold sink

patriknw:
think about ActorPublisher and ActorSubscriber also

patriknw:
remember that the supervision should also influence if the promise of `fold` and `foreach` is to be completed with the failure

cppexpert:
Just found this bug. Looks like it covers what I reported on SO: http://stackoverflow.com/q/32290285/226895

hepin1989:
I encounter this problem when using reactive-kafka 0.7.0,so If this is added,then should be nice for further use case.

patriknw:
There is a fat warning in the docs: "Warning
ZipWith, FlexiMerge, FlexiRoute junction, ActorPublisher source and ActorSubscriber sink components do not honour the supervision strategy attribute yet."

cppexpert:
Perhaps it's not fat enough. I didn't see it in actor integration page of documentation and wasted significant amount of time debugging this.

patriknw:
ok, should be mentioned in that page also (@2m could you fix that ?)

patriknw:
ActorPublisher and ActorSubscriber actually supports ordinary ordinary parental actor supervision without loosing the internal stream state.

cppexpert:
@patriknw Could you please demonstrate how I can "resume" the stream that is generated by `ActorPublisher` ? I'm getting very long list of XML items. Some items I can't parse but I'd like to continue instead of stopping paginated REST API replies.

patriknw:
I'm not sure I understand. You get the XML as an external message to the ActorPublisher and when you handle that message an exception is thrown. That should by default restart the actor and, but not stop the stream.

hepin1989:
ok,recheck the title,seems the fat warning should always be fat there :).

hepin1989:
@patriknw I encounter the problem is that once the protobuf message is corrupted,the subscriber will failed.but that bytes just was dirty data.A better way I think should do a map before it.but the reactive-kafka don;t have that.

ktoso:
Heard more demand for it recently, from @kciesielski

kciesielski:
[...]

PR #17009, opened 2015-03-05
=str #16866 fail fold and foreach future when function throws

patriknw:
Refs #16866

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
@@ -14,6 +14,8 @@ import scala.concurrent.{ Promise, Future }
 import scala.util.{ Success, Failure, Try }
 import akka.stream.FlowMaterializer
 import akka.stream.impl.StreamLayout.Module
+import scala.util.control.NonFatal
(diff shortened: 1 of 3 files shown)


akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -154,6 +156,12 @@ object Sink extends SinkApply {
           promise.success(())
           ctx.finish()
         }
+
+        override def decide(cause: Throwable): Supervision.Directive = {
+          // supervision will be implemented by #16916
+          promise.tryFailure(cause)
+          super.decide(cause)
```
bantonsson: I guess that means that #16916 will allow the supervision to decide if the promise is completed with the failure?

bantonsson:
LGTM

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
patriknw: exactly, I'll add a note to the ticket

drewhk:
LGTM

patriknw:
Refs #16866

Issue #16309, opened 2014-11-14
Fix the enormous code duplication in the streams specs


viktorklang:
FlowIteratorSpec looks virtually identical to
FlowIterableSpec looks virtually identical to
SynchronousPublisherFromIterableSpec looks virtually identical to
…

("Fixing bugs in one just to find that it was copy-pasted all over is not fun." — Unknown)

PR #17016, opened 2015-03-06
=str #16309 DRY FlowIteratorSpec, FlowIterableSpec, SynchronousIterableS...

patriknw:
...pec


akka-stream-tests/src/test/scala/akka/stream/impl/SynchronousIterablePublisherSpec.scala
```diff
@@ -1,172 +0,0 @@
...
-import scala.collection.immutable
-import scala.concurrent.duration._
-import akka.stream.testkit.AkkaSpec
-import akka.stream.testkit.StreamTestKit
-import akka.testkit.TestProbe
-import org.reactivestreams.{ Subscriber, Subscription }
-
-class SynchronousIterablePublisherSpec extends AkkaSpec {
-
-  "A SynchronousPublisherFromIterable" must {
```
drewhk: Wait. We still have SynchronousIterablePublisher right? I only removed the Iterable/IteratorPublisher stuff, but this one still exists. For example Source.single uses SynchronousIterablePublisher and it might be the fastest actually, much faster than the Stage based Iterable Source.

patriknw: I moved everything from this file into FlowIteratorSpec.scala, or did I miss something?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowIteratorSpec.scala
```diff
@@ -13,17 +12,139 @@ import akka.stream.testkit.StreamTestKit
...
+    }
+    val p = Source(iterable).runWith(Sink.publisher())
+    val c = StreamTestKit.SubscriberProbe[Int]()
+    p.subscribe(c)
+    c.expectSubscriptionAndError().getMessage should be("no next")
+    c.expectNoMsg(100.millis)
+  }
+}
+
+class SynchronousIterableSpec extends AbstractFlowIteratorSpec {
```
patriknw: @drewhk here is the test for SynchronousIterablePublisher, I moved it here to be able to reuse things from AbstractFlowIteratorSpec

drewhk: Aaah. OK, then it is fine. I missed that.

drewhk:
LGTM

patriknw:

patriknw:
Refs #16309

Issue #16993, opened 2015-03-04
Is FlattenStrategy placed in right package?


patriknw:
`abstract class FlattenStrategy` is used in both javadsl and scaladsl, but the concrete `concat` for the javadsl is in `javadsl.FlattenStrategy` and the concrete `concat` for the scaladsl is in `stream.FlattenStrategy`.

I think it would be most consistent to not have anything of this in the `stream` package, i.e. two separate classes in `javadsl` and `scaladsl`.

patriknw:
@rkuhn thoughts?

rkuhn:
Yes, I think the DSLs have evolved a little differently than we foresaw back then.

drewhk:
has this been done?

patriknw:
Not completely, see comments in PR #17019

PR #17019, opened 2015-03-06
!str #16993 Separate scaladsl/javadsl for FlattenStategy

patriknw:
`abstract class FlattenStrategy` was used in both javadsl and scaladsl, but the concrete concat
for the javadsl was in javadsl.FlattenStrategy and the concrete concat for the scaladsl is in
stream.FlattenStrategy.

Now there are separate FlattenStategy in scaladsl and javadsl packages and conversion
as we have in other places.


ktoso:
Missing conversion in the opposite way: `asJava` provided in `JavaConverters` as implicit. 
Would be consistent with other parts of the API where we do such conversions.

ktoso:
LGTM otherwise :)

patriknw:
@ktoso good point, I have added the `toJava` method. Could not add it to the implicits because it must be implemented in the concrete subclasses.

rkuhn:
Thinking about it more: we should remove the implicits completely in order to allow Java users to do conversions both ways—simply add the `asJava` method to the Scala classes.

patriknw:
@rkuhn great point, I have replaced it with explicit mehtods
and rebased

bantonsson:
LGTM

akka-stream/src/main/scala/akka/stream/scaladsl/FlattenStrategy.scala
```diff
@@ -15,7 +20,10 @@ object FlattenStrategy {
    * emitting its elements directly to the output until it completes and then taking the next stream. This has the
    * consequence that if one of the input stream is infinite, no other streams after that will be consumed from.
    */
-  def concat[T]: FlattenStrategy[scaladsl.Source[T, _], T] = Concat[T]()
+  def concat[T]: FlattenStrategy[Source[T, _], T] = Concat[T]()
 
-  private[akka] final case class Concat[T]() extends FlattenStrategy[scaladsl.Source[T, _], T]
+  private[akka] final case class Concat[T]() extends FlattenStrategy[Source[T, _], T] {
+    override def asJava[S]: javadsl.FlattenStrategy[S, T] =
```
rkuhn: Why not derive S from the type info we have? (i.e. `javadsl.FlattenStrategy[javadsl.Source[T, M], T]`)

rkuhn:
Apart from the last bit LGTM.

akka-stream/src/main/scala/akka/stream/scaladsl/FlattenStrategy.scala
patriknw: @rkuhn what is confusing me is the `-T` part of `abstract class FlattenStrategy[-T, U]`. Therefore I thought that `T` was not always a `Source[U, _]`.

rkuhn: Not sure what you mean: the variance annotation is only interesting outside of the declaration, on the inside the compiler will make sure that you don’t do anything wrong—meaning if my proposal compiles then it is fine ;-)

Another way to look at this: variance means that a client of this type can see a different type argument, in this case it could be a supertype, but that would also mean that said client will compute a different return type for this method. Whether or not that computation would violate type constraints is automatically verified by the compiler.

patriknw: when trying to change corresponding thing in javadsl.FlattenStrategy to:

    def asScala: scaladsl.FlattenStrategy[scaladsl.Source[U, _], U]

compilation error in javadsl.Flow:

    def flatten[U](strategy: FlattenStrategy[Out, U]): javadsl.Flow[In, U, Mat] =
      new Flow(delegate.flatten(strategy.asScala))


    type mismatch; found : akka.stream.scaladsl.FlattenStrategy[akka.stream.scaladsl.Source[U, _],U] required: akka.stream.scaladsl.FlattenStrategy[Out,?]

drewhk:
LGTM

rkuhn:
Thinking about this in some more depth and trying it out: it does not make sense to have `asJava` and `asScala` methods on a purely symbolic strategy placeholder without any implementation. We can either introduce a simple wrapper `case class JavaStrategy(s: javadsl.FlattenStrategy) extends FlattenStrategy` or we introduce a common supertype, but the user should never see this or care about it.

The reason for this change of mind is that the `javadsl.Concat` of course only applies to `javadsl.Source[T, M]`, but converting a `scaladsl.Concat` would need to turn into something that deals with `scaladsl.Source[T, M]`—a paradox that therefore eats its own tail.

patriknw:
@rkuhn alright, thanks for that insight. I have removed asJava and asScala and got it working without any casts: afde5ed

akka-stream/src/main/scala/akka/stream/javadsl/Graph.scala
```diff
@@ -279,7 +279,7 @@ object FlowGraph {
    */
   def builder[M](): Builder[M] = new Builder()(new scaladsl.FlowGraph.Builder[M])
 
-  final class Builder[Mat]()(private implicit val delegate: scaladsl.FlowGraph.Builder[Mat]) { self ⇒
+  final class Builder[+Mat]()(private implicit val delegate: scaladsl.FlowGraph.Builder[Mat]) { self ⇒
```
patriknw: added this `+` because of "covariant type M occurs in invariant position in type ⇒ akka.stream.javadsl.FlowGraph.Builder[M] of method asJava"

rkuhn: yup, that’s right

rkuhn:
LGTM!

patriknw:
thanks, I'll squash and merge this

patriknw:
Refs #16993

Issue #15884, opened 2014-09-14
LevelDB dependency


ahjohannessen:
LevelDB ought to be a separate dependency, e.g. "akka-persistence-leveldb", as most people likely will use something else in production.

hepin1989:
+1 for moving out to a separate project

Andrei-Pozolotin:
@patriknw please make repo or choose module name for the leveldb plugins

patriknw:
it should be module `akka-persistence-leveldb` (akka-persistence-leveldb-experimental)

PR #17024, opened 2015-03-06
!per #15884 Make LevelDB an optional dependency

Andrei-Pozolotin:
since it is not immediately obvious how to split leveldb plugin into separate module
w/o introducing circular dependency or code duplication, lets make it "optional" for now

### DIFF
--- a/akka-docs/rst/java/persistence.rst
+++ b/akka-docs/rst/java/persistence.rst
@@ -4,7 +4,6 @@
 Persistence
 ###########
 
-
 Akka persistence enables stateful actors to persist their internal state so that it can be recovered when an actor
 is started, restarted after a JVM crash or by a supervisor, or migrated in a cluster. The key concept behind Akka
 persistence is that only changes to an actor's internal state are persisted but never its current state directly
@@ -53,6 +52,22 @@ Akka persistence is a separate jar file. Make sure that you have the following d
     <version>@version@</version>
   </dependency>
 
+Akka persistence extension comes with few built-in persistence plugins, including 
+in-memory heap based journal, local file-system based snapshot-store and LevelDB based journal.
+
+LevelDB based plugins will require the following additional dependency declaration::
+
+  <dependency>
+    <groupId>org.iq80.leveldb</groupId>
+    <artifactId>leveldb</artifactId>
+    <version>0.7</version>
+  </dependency>
+  <dependency>
+    <groupId>org.fusesource.leveldbjni</groupId>
+    <artifactId>leveldbjni-all</artifactId>
+    <version>1.7</version>
+  </dependency>
+
 Architecture
 ============
 
@@ -69,13 +84,14 @@ Architecture
   case of sender and receiver JVM crashes.
 
 * *Journal*: A journal stores the sequence of messages sent to a persistent actor. An application can control which messages
-  are journaled and which are received by the persistent actor without being journaled. The storage backend of a journal is
-  pluggable. The default journal storage plugin writes to the local filesystem, replicated journals are available as
-  `Community plugins`_.
+  are journaled and which are received by the persistent actor without being journaled. The storage backend of a journal is pluggable. 
+  Persistence extension comes with a "leveldb" journal plugin, which writes to the local filesystem, 
+  and replicated journals are available as `Community plugins`_.
 
 * *Snapshot store*: A snapshot store persists snapshots of a persistent actor's or a view's internal state. Snapshots are
-  used for optimizing recovery times. The storage backend of a snapshot store is pluggable. The default snapshot
-  storage plugin writes to the local filesystem.
+  used for optimizing recovery times. The storage backend of a snapshot store is pluggable. 
+  Persistence extension comes with a "local" snapshot storage plugin, which writes to the local filesystem,
+  and replicated snapshot stores are available as `Community plugins`_.
 
 .. _Community plugins: http://akka.io/community/
 
@@ -486,11 +502,25 @@ configuration key. The method can be overridden by implementation classes to ret
 Storage plugins
 ===============
 
-Storage backends for journals and snapshot stores are pluggable in Akka persistence. The default journal plugin
-writes messages to LevelDB (see :ref:`local-leveldb-journal-java`). The default snapshot store plugin writes snapshots
-as individual files to the local filesystem (see :ref:`local-snapshot-store-java`). Applications can provide their own
-plugins by implementing a plugin API and activate them by configuration. Plugin development requires the following
-imports:
+Storage backends for journals and snapshot stores are pluggable in the Akka persistence extension.
+
+Directory of persistence journal and snapshot store plugins is available at the Akka Community Projects page, see `Community plugins`_
+
+Plugins can be selected either by "default", for all persistent actors and views,
+or "individually", when persistent actor or view defines it's own set of plugins.
+
+When persistent actor or view does NOT override ``journalPluginId`` and ``snapshotPluginId`` methods,
+persistense extension will use "default" journal and snapshot-store plugins configured in the ``reference.conf``::
+
+    akka.persistence.journal.plugin = ""
+    akka.persistence.snapshot-store.plugin = ""
+
+However, these entires are provided as empty "", and require explicit user configuration via override in the user ``application.conf``.
+For an example of journal plugin which writes messages to LevelDB see :ref:`local-leveldb-journal-java`. 
+For an example of snapshot store plugin which writes snapshots as individual files to the local filesystem see :ref:`local-snapshot-store-java`. 
+
+Applications can provide their own plugins by implementing a plugin API and activate them by configuration. 
+Plugin development requires the following imports:
 
 .. includecode:: code/docs/persistence/PersistencePluginDocTest.java#plugin-imports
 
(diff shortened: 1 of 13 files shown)


Andrei-Pozolotin:
@patriknw please review

hepin1989:
make the default in-memory and move the leveldb implement to separate module?

patriknw:
Move tests to a 3rd module.
Not sure we can make inmem default. It is probably better to fail if journal is not configured.

Andrei-Pozolotin:
inmem is already default:
https://github.com/akka/akka/blob/master/akka-persistence/src/main/resources/reference.conf#L23

patriknw:
Ok, good then

Andrei-Pozolotin:
-------- Original Message --------
Subject: Re: [akka] !per Patrik Nordwall5884 Make LevelDB an optional
dependency (#17024)
>
> Ok, good then
>
not sure what you mean?

if you still insist you need some kind of persistent journal by default,
I think it is easier to provide some form of simplistic "journal.local"
(similar to "snapshot-store.local", direct file system based journal)
rather then maintain 2 more modules, plus all the refactoring -
- paying with all of this just to kill the level-db dependency

patriknw:
I thought leveldb was default and as response to hepin1989's comment I thought that we could not silently change that to inmem, since that would be a surprising breaking change.

All that was irrelevant for this issue, since leveldb is not default.

So we have two alternatives:
- change the dependency to provided
- move code to leveldb and tests modules

I have no strong opinion, but recently @rkuhn has preferred more small modules over provided dependencies and I tend to agree with that.

Andrei-Pozolotin:
I see. we can treat them not as alternatives, but 2 consecutive steps:
first, "change the dependency to provided"
later, "move code to leveldb and tests modules"

hepin1989:
if we change dependency to provided and user don't provided leveldb dependency at runtime,and don't choose other implement ,then it fails too.
so just as the https://github.com/akka/akka/issues/15884 said,people will keep using other implement ,so keep it in a separate module is OK.and in fact,if you are using leveldb implement as default,you are fail to run the app in  sbt.

even change to separate module is a break change,but I think that worth.If people want to using leveldb implement just ctrl+c , ctrl +v from the migrate guide.It works.

Andrei-Pozolotin:
@rkuhn @patriknw if you do not want "change the dependency to provided" - feel free to close this.

rkuhn:
PLS BUILD

rkuhn:
The change looks good to me, I’m not certain that splitting off the code is a big win.

akka-persistence/src/main/resources/reference.conf
```diff
@@ -161,3 +161,21 @@ akka.persistence.journal.leveldb-shared {
...
+#        <dependency>
+#            <groupId>org.iq80.leveldb</groupId>
+#            <artifactId>leveldb</artifactId>
+#            <version>0.7</version>
+#        </dependency>
+#        <dependency>
+#            <groupId>org.fusesource.leveldbjni</groupId>
+#            <artifactId>leveldbjni-all</artifactId>
+#            <version>1.7</version>
+#        </dependency>
```
patriknw: This should be included in the reference documentation, also.
It's a breaking change so it should also be mentioned in the migration guide.

patriknw:
ok

akka-persistence/src/main/resources/reference.conf
Andrei-Pozolotin: applied

Andrei-Pozolotin:
@patriknw please review & restart build

bantonsson:
OK TO TEST

patriknw:
missing dependency in contrib, according to the build failure?

akka-docs/rst/project/migration-guide-persistence-experimental-2.3.x-2.4.x.rst
```diff
@@ -178,3 +178,19 @@ persistent actor on the sending side.
...
+Now persistence extension will instead use in-memory journal plugin ``akka.persistence.journal.inmem`` as a default.
+This change was needed to decouple persistence extension from the the LevelDB dependency.
+Please see persistence extension ``reference.conf`` for details. 
+
+Converted LevelDB to an optional dependency
+===========================================
+Persistence extension uses LevelDB based plugins for own development and keeps related code in the published jar.
+However previously LevelDB was a ``compile`` scope dependency, and now it is an ``optional;provided`` dependency.
+To continue using LevelDB based persistence plugins for development it is now required for related user projects
+to include an additional explicit dependency declaration for the LevelDB artifacts in the scope ``test``.
```
patriknw: is scope `test` correct here?

akka-docs/rst/java/persistence.rst
```diff
@@ -53,6 +53,22 @@ Akka persistence is a separate jar file. Make sure that you have the following d
...
+  <dependency>
+    <groupId>org.iq80.leveldb</groupId>
+    <artifactId>leveldb</artifactId>
+    <version>0.7</version>
+  </dependency>
+  <dependency>
+    <groupId>org.fusesource.leveldbjni</groupId>
+    <artifactId>leveldbjni-all</artifactId>
+    <version>1.7</version>
+  </dependency>
```
patriknw: this should be added to lambda-persistence.rst also

akka-docs/rst/project/migration-guide-persistence-experimental-2.3.x-2.4.x.rst
Andrei-Pozolotin: yes, in the context of "development":
"To continue using LevelDB based persistence plugins for development"

akka-docs/rst/java/persistence.rst
Andrei-Pozolotin: applied

akka-docs/rst/project/migration-guide-persistence-experimental-2.3.x-2.4.x.rst
patriknw: I find that confusing. Some use it for production. I suggest that you remove "for development" and "in the scope test"

patriknw:
@Andrei-Pozolotin can you take a look at the build failure?

    akka-contrib/multi-jvm:compile) Compilation failed

Andrei-Pozolotin:
ok

akka-docs/rst/project/migration-guide-persistence-experimental-2.3.x-2.4.x.rst
Andrei-Pozolotin: applied

Andrei-Pozolotin:
@patriknw please review

akka-docs/rst/project/migration-guide-persistence-experimental-2.3.x-2.4.x.rst
```diff
@@ -178,3 +178,19 @@ persistent actor on the sending side.
 
 Read more about at-least-once delivery in the :ref:`documentation for Scala <at-least-once-delivery>` and 
 :ref:`documentation for Java <at-least-once-delivery-java>`.  
+
+Default persistence plugins
+===========================
+Previously default ``akka.persistence.journal.plugin`` was set to the LevelDB journal ``akka.persistence.journal.leveldb``.
+Now persistence extension will instead use in-memory journal plugin ``akka.persistence.journal.inmem`` as a default.
+This change was needed to decouple persistence extension from the the LevelDB dependency.
```
patriknw: You changed this in an earlier pull request, and I agree that this is the reason why we must change it.
However, I think it would be better to not have a default at all. I don't want to be responsible in case someone previously used leveldb as the default and then drops in 2.4 into production and it silently stores things in memory.

Users don't read docs or migration guides until there is a problem.

rkuhn: I agree with @patriknw, we cannot change the default back-end, only remove the default itself.

Andrei-Pozolotin: applied, please take a look

patriknw:
@Andrei-Pozolotin I also found this in persistence.rst: 
* "The default journal plugin writes messages to LevelDB"
* "The default snapshot store plugin writes ..."
* "The default journal plugin is ``akka.persistence.journal.leveldb``"
* "The default snapshot store plugin is ``akka.persistence.snapshot-store.local``"

Perhaps also add a sentence in the Configuration section of persistence.rst mentioning that akka.persistence.journal.plugin and akka.persistence.snapshot-store.plugin must be defined.

Andrei-Pozolotin:
-------- Original Message --------
Subject: Re: [akka] !per Patrik Nordwall5884 Make LevelDB an optional
dependency (#17024)
>
> @Andrei-Pozolotin <https://github.com/Andrei-Pozolotin> I also found
> this in persistence.rst:
>
applied

patriknw:
LGTM, thanks!
please squash

Andrei-Pozolotin:
-------- Original Message --------
Subject: Re: [akka] !per Patrik Nordwall5884 Make LevelDB an optional
dependency (#17024)
>
> please squash
>
squashed

hepin1989:
+1

patriknw:
Refs #15884

Andrei-Pozolotin:
thank you

PR #17026, opened 2015-03-08
Changes all occurances of Typesafe copyright to extend to 2015

jtescher:


drewhk:
OK TO TEST

drewhk:
Wow, thanks! LGTM

ktoso:
Thanks!
Related info: @bantonsson was researching that we switch to https://github.com/sbt/sbt-header for copyright but I'm not sure on the status of that. Some edge cases popped up AFAIR.

rkuhn:
Thanks, that’s awesome! Now I really hate to say this, but this PR has bugs: those lines that previously only had `2014` in them should not be changed to `2015` but to `2014–2015`. Your efforts are much appreciated!

jtescher:
Ah! Good point. I'll amend the PR.

jtescher:
@rkuhn fixed those bugs. Everything else looking good?

rkuhn:
Looking good, thanks again!

Issue #17028, opened 2015-03-09
ByteString performs too many copies on reasonably immutable data


colinrgodsey:
PR ~~#17069~~ offers some improvements to ByteString, mostly related to internal array copying. 
PR ~~#17144~~ has changes to allow serialization across all ByteStrings.
(previous PRs: ~~#17029~~, ~~#17070~~)

Relating to conversation here: 
https://groups.google.com/forum/#!topic/akka-dev/wgEe0ERrqVo

Assumptions
- A readonly ByteBuffer wrapper of internal Array in ByteString is safe as the underlying array is not accessible. 
- ~~It is faster to use a wrapped ByteBuffer with NIO directly (asByteBuffer) than copying it to a temporary buffer~~ https://gist.github.com/colinrgodsey/9bc606d09d035ba2334c
- ~~A ByteString(1) produced for a String should be able to hit NIO as a wrapped ByteBuffer with no array copies done besides the original generation of the Array from the String.~~
- It is faster to produce a ByteString1 from an IndexedSeq by copying the contents to a new Array instead of resizing the temp buffer in ByteStringBuilder (#17069)

Future Considerations
- Allow creation of sparse ByteString(1) from a _safe_ IndexedSeq (Vectors, etc if possible)
- Allow reading from an NIO channel directly with ByteString (0-copy read for known sizes, safe ByteString creation)
- Package-private base classes for possible extension under the _akka_ namespace

Other Observations
- Optimizations of TraversableOnce#copyToArray does expose the internal Array to potentially unsafe APIs extending TraversableOnce (existing)
- ~~No serialization for ByteString1, leaving ByteString (as an abstract class) not serializeable (#15170, #17070).~~

colinrgodsey:
@rkuhn I was wondering if I could get some input on the 'other observations' towards the end there. 

I can see why ByteString1 isnt serializable (default ser would copy the entire underlying referenced array), but it could be possible with an (albeit messy) custom ser. As far as I can tell its only safe to send a compact ByteString over the wire, and this seems odd to me. Was wondering if there's some other reason for this im not seeing.

@ktoso hey again

rkuhn:
We’re currently all traveling in preparation for ScalaDays, so our response latency might not be as good as we want it to be.

colinrgodsey:
Woops, this was addressed a few years ago.

PR #17029, opened 2015-03-09
ByteString 0-copy optimizations [DO NOT MERGE]

colinrgodsey:
## deferred to #17069

various optimizations to ByteString described in #17028
- more paths for 0-copy usage of ByteStrings
- usage of wrapped ByteBuffers where possible
- for more see #17028

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/util/ByteStringSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/util/ByteStringSpec.scala
@@ -306,7 +306,7 @@ class ByteStringSpec extends WordSpec with Matchers with Checkers {
           val denseUnsafeBs = ByteString.ByteString1(arr)
           val additionalBs = ByteString.ByteString1(Array[Byte](1, 2, 3))
           val bs = if (isDense) denseUnsafeBs
-          else ByteString.ByteStrings(denseUnsafeBs, denseUnsafeBs)
+          else ByteString.ByteStrings(denseUnsafeBs, additionalBs)
 
           //should directly use the underlying buffer
           if (len < 1) true
--- a/akka-actor-tests/src/test/scala/akka/util/ByteStringSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/util/ByteStringSpec.scala
@@ -4,6 +4,7 @@
 
 package akka.util
 
+import akka.io.DirectByteBufferPool
 import org.scalatest.WordSpec
 import org.scalatest.Matchers
 import org.scalatest.prop.Checkers
@@ -298,7 +299,37 @@ class ByteStringSpec extends WordSpec with Matchers with Checkers {
         }
       }
 
-      "asByteBuffers" in {
+      "asByteBuffer respects canWrapAsByteBuffer" in {
+        check { (slice: ArraySlice[Byte], isDense: Boolean) ⇒
+          val (arr, from, until) = slice
+          val len = until - from
+          val denseUnsafeBs = ByteString.ByteString1(arr)
+          val additionalBs = ByteString.ByteString1(Array[Byte](1, 2, 3))
+          val bs = if (isDense) denseUnsafeBs
+          else ByteString.ByteStrings(denseUnsafeBs, denseUnsafeBs)
+
+          //should directly use the underlying buffer
+          if (len < 1) true
+          else {
+            //make buffer before mutation
+            val testByteBuffer = bs.asByteBuffer
+
+            val origByte = arr(0)
+            val testByte = (origByte - 1).toByte
+            arr(0) = testByte
+
+            require(bs.canWrapAsByteBuffer == isDense, "all dense arrays should wrap!")
+
+            //test whether mutations to the underling array are visible
+            if (bs.canWrapAsByteBuffer)
+              bs(0) == testByte && testByteBuffer.get(0) == testByte
+            else
+              bs(0) == testByte && testByteBuffer.get(0) == origByte
+          }
+        }
+      }
+
+      "calling asByteBuffers" in {
         check { (a: ByteString) ⇒ if (a.isCompact) a.asByteBuffers.size == 1 && a.asByteBuffers.head == a.asByteBuffer else a.asByteBuffers.size > 0 }
         check { (a: ByteString) ⇒ a.asByteBuffers.foldLeft(ByteString.empty) { (bs, bb) ⇒ bs ++ ByteString(bb) } == a }
         check { (a: ByteString) ⇒ a.asByteBuffers.forall(_.isReadOnly) }
@@ -551,6 +582,40 @@ class ByteStringSpec extends WordSpec with Matchers with Checkers {
       "encoding Float in little-endian" in { check { slice: ArraySlice[Float] ⇒ testFloatEncoding(slice, LITTLE_ENDIAN) } }
       "encoding Double in big-endian" in { check { slice: ArraySlice[Double] ⇒ testDoubleEncoding(slice, BIG_ENDIAN) } }
       "encoding Double in little-endian" in { check { slice: ArraySlice[Double] ⇒ testDoubleEncoding(slice, LITTLE_ENDIAN) } }
+    }
+  }
+
+  "A DirectByteBufferPool" must {
+    "ignore read-only ByteBuffers" when {
+      "wrapped from ByteString" in {
+        val pool = new DirectByteBufferPool(10, 10)
+        val arr = Array[Byte](1, 2, 3, 4)
+        val compactBs = ByteString.ByteString1(arr)
+        val writeableByteBuffer = ByteBuffer.wrap(arr.clone())
+
+        require(compactBs.asByteBuffer.isReadOnly, "ByteString didnt produce a read-only buffer")
+
+        require(pool.size == 0, "started nonEmpty")
+        pool.release(compactBs.asByteBuffer)
+        require(pool.size == 0, "tried to repool a read-only buffer!")
+        pool.release(writeableByteBuffer)
+        require(pool.size == 1)
+      }
+    }
+  }
+
+  "A ByteBuffer" must {
+    "have equality in 'compactness' logic to ByteString" in {
+      check { slice: ArraySlice[Byte] ⇒
+        val (data, from, to) = slice
+        val len = to - from
+
+        val bs = ByteString.ByteString1(data, from, len)
+        val bb = bs.asByteBuffer
+        val bbCompact = !(bb.remaining < data.length)
+
+        bbCompact == bs.isCompact || len == 0
+      }
     }
   }
 }
--- a/akka-actor/src/main/scala/akka/io/TcpConnection.scala
+++ b/akka-actor/src/main/scala/akka/io/TcpConnection.scala
@@ -361,15 +361,22 @@ private[io] abstract class TcpConnection(val tcp: TcpExt, val channel: SocketCha
   }
 
   def PendingBufferWrite(commander: ActorRef, data: ByteString, ack: Event, tail: WriteCommand): PendingBufferWrite = {
-    val buffer = bufferPool.acquire()
-    try {
-      val copied = data.copyToBuffer(buffer)
-      buffer.flip()
-      new PendingBufferWrite(commander, data.drop(copied), ack, buffer, tail)
-    } catch {
-      case NonFatal(e) ⇒
-        bufferPool.release(buffer)
-        throw e
+    if (data.headByteString.canWrapAsByteBuffer) {
+      val head = data.headByteString
+      val remaining = data.drop(head.length)
+
+      new PendingBufferWrite(commander, remaining, ack, head.asByteBuffer, tail)
+    } else {
+      val buffer = bufferPool.acquire()
+      try {
+        val copied = data.copyToBuffer(buffer)
+        buffer.flip()
+        new PendingBufferWrite(commander, data.drop(copied), ack, buffer, tail)
+      } catch {
+        case NonFatal(e) ⇒
+          bufferPool.release(buffer)
+          throw e
+      }
     }
   }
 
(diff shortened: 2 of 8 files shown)


typesafehub-validator:
Hi @colinrgodsey,

akka-actor/src/main/scala/akka/util/ByteString.scala
```diff
@@ -155,14 +176,13 @@ object ByteString {
...
     def isCompact: Boolean = (length == bytes.length)
 
+    override def canWrapAsByteBuffer: Boolean = true
+
     def compact: CompactByteString =
       if (isCompact) ByteString1C(bytes) else ByteString1C(toArray)
 
-    def asByteBuffer: ByteBuffer = {
-      val buffer = ByteBuffer.wrap(bytes, startIndex, length).asReadOnlyBuffer
-      if (buffer.remaining < bytes.length) buffer.slice
```
colinrgodsey: logic proved in test to be equivalent to isCompact. If expecting a compact ByteBuffer, one can simply ByteString#compact.asByteBuffer

akka-actor/src/main/scala/akka/util/ByteString.scala
```diff
@@ -132,6 +146,11 @@ object ByteString {
     def apply(bytes: Array[Byte]): ByteString1 = ByteString1(bytes, 0, bytes.length)
     def apply(bytes: Array[Byte], startIndex: Int, length: Int): ByteString1 =
       if (length == 0) empty else new ByteString1(bytes, startIndex, length)
+    def apply(seq: collection.IndexedSeq[Byte]): ByteString1 = {
+      val tmpArr = new Array[Byte](seq.length)
+      seq.copyToArray(tmpArr)
```
colinrgodsey: allows optimizations via copyToArray

akka-actor/src/main/scala/akka/io/TcpConnection.scala
```diff
@@ -361,15 +361,22 @@ private[io] abstract class TcpConnection(val tcp: TcpExt, val channel: SocketCha
...
-    val buffer = bufferPool.acquire()
-    try {
-      val copied = data.copyToBuffer(buffer)
-      buffer.flip()
-      new PendingBufferWrite(commander, data.drop(copied), ack, buffer, tail)
-    } catch {
-      case NonFatal(e) ⇒
-        bufferPool.release(buffer)
-        throw e
+    if (data.headByteString.canWrapAsByteBuffer) {
```
colinrgodsey: this may have the downside of committing tiny writes (as sparse ByteString parts might be small) to the channel. if that is not ideal, this check can easily account for size and make dense buffers when needed (e.g. things below the standard MTU)

colinrgodsey:
@typesafehub-validator agreed to the CLA

bantonsson:
OK TO TEST

akka-actor/src/main/scala/akka/util/ByteString.scala
```diff
@@ -381,6 +408,11 @@ sealed abstract class ByteString extends IndexedSeq[Byte] with IndexedSeqOptimiz
   def asByteBuffers: immutable.Iterable[ByteBuffer]
 
   /**
+   * Indicates whether this ByteString can produce a wrapped ByteBuffer via [[asByteBuffer]].
```
colinrgodsey: scaladoc error? seeings lots of these in the build tho 

[warn] /mnt/akka/jenkinsakka/localhome-a0/workspace/pr-validator-per-commit-jenkins/akka-actor/target/java/akka/util/ByteString.java:207: warning - Tag @link: reference not found: asByteBuffer

colinrgodsey:
https://github.com/colinrgodsey/akka/blob/colinrgodsey-17028-1/akka-actor/src/main/scala/akka/io/TcpConnection.scala#L400

This portion where it rewrites the remainder to the buffer- this needs to check if the buffer is readonly and get another 0-copy bytebuffer instead of trying to write the rest.

akka-actor-tests/src/test/scala/akka/util/ByteStringSpec.scala
```diff
@@ -553,4 +584,38 @@ class ByteStringSpec extends WordSpec with Matchers with Checkers {
...
+    "have equality in 'compactness' logic to ByteString" in {
+      check { slice: ArraySlice[Byte] ⇒
+        val (data, from, to) = slice
+        val len = to - from
+
+        val bs = ByteString.ByteString1(data, from, len)
+        val bb = bs.asByteBuffer
+        val bbCompact = !(bb.remaining < data.length)
+
+        bbCompact == bs.isCompact || len == 0
```
colinrgodsey: need require

akka-actor/src/main/scala/akka/io/TcpConnection.scala
```diff
@@ -360,16 +360,25 @@ private[io] abstract class TcpConnection(val tcp: TcpExt, val channel: SocketCha
...
-      val copied = data.copyToBuffer(buffer)
-      buffer.flip()
-      new PendingBufferWrite(commander, data.drop(copied), ack, buffer, tail)
-    } catch {
-      case NonFatal(e) ⇒
-        bufferPool.release(buffer)
-        throw e
+  //TODO: move canWrapAsByteBuffer into PendingBufferWrite#doWrite
+  //TODO: check chunk size
+  def PendingBufferWrite(commander: ActorRef, data: ByteString, ack: Any, tail: WriteCommand): PendingBufferWrite = {
```
rkuhn: why change the `ack` type?

akka-actor/src/main/scala/akka/io/TcpConnection.scala
```diff
@@ -389,6 +398,12 @@ private[io] abstract class TcpConnection(val tcp: TcpExt, val channel: SocketCha
           if (data eq remainingData) this
           else new PendingBufferWrite(commander, data, ack, buffer, tail) // copy with updated remainingData
 
+        } else if (data.nonEmpty && buffer.isReadOnly) {
+          val head = data.headByteString
+          val remaining = data.drop(head.length)
```
rkuhn: This is fragile: can there ever be a read-only ByteBuffer that has a different length in the future?

akka-actor/src/main/scala/akka/io/TcpConnection.scala
```diff
@@ -389,6 +398,12 @@ private[io] abstract class TcpConnection(val tcp: TcpExt, val channel: SocketCha
...
+
+          //this will always start with the direct buffer, or will change over to atmost one
+          PendingBufferWrite(commander, remaining, ack, tail).doWrite(info)
```
rkuhn: This is technically open recursion, it might be possible to construct a pathological write request that blows the stack.

akka-actor/src/main/scala/akka/util/ByteString.scala
rkuhn: These can be tricky, but I think the right syntax is `[[ByteString#asByteBuffer]]`.

akka-actor/src/main/scala/akka/util/ByteString.scala
```diff
@@ -576,8 +613,14 @@ final class ByteStringBuilder extends Builder[Byte, ByteString] {
         _length += bs.length
       case xs: WrappedArray.ofByte ⇒
         putByteArrayUnsafe(xs.array.clone)
+      case seq: collection.IndexedSeq[Byte] if shouldResizeTempFor(seq.length) ⇒
+        //we will have to resize and copy anyways, might as well flush it
+        clearTemp()
+        _builder += ByteString.ByteString1(seq)
+        _length += seq.length
```
rkuhn: perhaps due to food coma, but I currently don’t see why this case should be added

akka-actor/src/main/scala/akka/io/TcpConnection.scala
colinrgodsey: this was mostly because the class takes an ack: Any, but the factory method took an ack: Event (and couldnt then call the factory method from the class, which has the 'bytebuffer picking' logic)

I can probably align to either ack: Any or ack: Event. I've noticed the api for like akka.io.udp asks for an ack of type 'Event', but I've picked through the logic and it didnt really seem required anywhere. its not sealed or anything, but it might also be overreach for this pr.

akka-actor/src/main/scala/akka/io/TcpConnection.scala
colinrgodsey: there could be because of the sparse headByteString logic.

normally the chain of 'PendingBufferWrite's would just acquire a temp rw-bytebuffer when its created, and pass that buffer to each subsequent PendingBufferWrite that could be made (or just hammers on doWrite inline until its done), the calls release on the buffer after everything.

This change breaks that assumption and on each instance of PendingBufferWrite you're not sure if you have a rw or ro bytebuffer, and thats why the factory method is called again as it checks if the next 'bytestring' chunk (any ByteString1/1C thats at the head of a ByteStrings) can be wrapped as a ro-buffer. 

The way this logic ends up panning out is that it will use read-only buffers as long as possible, and may switch to a rw buffer (but never back to a ro-bytebuffer). The thinking around this was that i didnt want to modify with the logic around releasing buffers from PendingBufferWrite. The connection is set up to only release the last buffer in the 'chain' of PendingBufferWrite, so I wanted to make sure that if one was pulled from the pool, it stayed there for the whole chain.

I'm going to fix this. It is fragile, and the fact I had to write this much about it probably means its super murky. Plus this logic as it is can enable really tiny writes which isn't ideal when making kernel calls.

akka-actor/src/main/scala/akka/io/TcpConnection.scala
colinrgodsey: hmm, so normally there was recursion to 'writeToChannel' on data.nonEmpty, which should only happen if the first chunk of data was written in full to the channel (any incomplete writes or end-of-data stopped the recursion).

The *data.nonEmpty && buffer.isReadOnly, new PBW.doWrite(...)* part is effectively doing the same thing (but with a new PBW. This does feel like abuse of the existing code though, and is murky. ill fix it up.

akka-actor/src/main/scala/akka/util/ByteString.scala
colinrgodsey: ah, so the main reason this is here is to avoid resizing the temp buffer if the next IndexedSeq will flood it out. (causing another copy and allocation just for the temp buffer). The temp buffer itself is also copied again, so if we can produce a compact array from it right off the bat, we avoid 1 definite copy, and 1 potential copy for the temp buffer resize. This also cleans up the builder nicely where the temp buffer isnt even used or allocated unless you're feeding it individual Bytes or some TraversableOnce.

This should probably be for any Seq (not just indexedseq) as it should happen with anything we know the length of (TraversableOnce etc. should always go to temp as we dont have a size hint). 

The only downside I see of doing this is that we will end up producing more sparse ByteStrings instead of dense ones when constructing from IndexedSeqs. So the tradeoff ends up being: less copies, but more sparseness (and sparseness may come in handy with TcpConnection and ByteString#headByteString.canWrapAsByteBuffer, as we can avoid another copy there and potentially dereference more buffers over time).

jrudolph:
IMO the main thing missing here is a proof that using a wrapped (i.e. a non-direct) ByteBuffer has any performance benefits when used with the low-level IO methods. AFAIK any wrapped ByteBuffer will be copied into a direct one when it is passed to IO methods. This is necessary since the user-level native IO methods need to work on native (= off-JVM-heap) buffers at some point so this is a buffer copy you cannot get around. 

Letting the IO methods do the copy to an ad-hoc created direct buffer means that this direct buffer is created out of our control which in previous benchmarks this shown to be adverse to performance compared to our manual pooling approach.

colinrgodsey:
@jrudolph ooh... hmmm. i figured it did a copy no matter what (as even the direct buffers are mutable)

I'm making a small performance test for this... but my eyes are starting to glaze over thinking of the system/implementation specifics for the JVM.....

colinrgodsey:
Well, this will at-least make the PR cleanup easier. Took me a while to get some meaningful data, but even with the copy, direct buffers win out. 

```
[info] - A direct ByteBuffer (with copy) should be faster than a wrapped buffer (13 seconds, 247 milliseconds)
[info]   + failed samples: 0, nSamples: 40000, testBufferSize: 4096, batchNum: 50 
[info]   + 99.0% direct: 575 (5) wrapped: 585 (9) 
[info]   + 98.0% direct: 535 (17) wrapped: 545 (18) 
[info]   + 80.0% direct: 340 (70) wrapped: 345 (81) 
[info]   + 70.0% direct: 300 (136) wrapped: 306 (134) 
[info]   + 50.0% direct: 267 (315) wrapped: 275 (352) 
[info]   + 30.0% direct: 251 (811) wrapped: 259 (804) 
```

Not sure what to do with the test. If any of you need blog-fodder or something, I know I couldnt find any reasonable metrics online. Got it posted here: https://gist.github.com/colinrgodsey/9bc606d09d035ba2334c

colinrgodsey:
So at this point, I'm willing to reduce this PR down to just the builder change if anybody thinks that alone is worth it (the resizing of the temp buffer). 

I also have a chance set that fixes serialization for ByteString, I can just move the builder change to there and trash this particular PR.

rkuhn:
Please don’t draw hasty conclusions from me not responding, I’m still preparing my ScalaDays presentation—I should have some time on Thursday or Friday to take a deeper look again.

colinrgodsey:
@rkuhn no rush! just trying to flush out as much as i can here until you guys can pounce on it

wish I could be there for ScalaDays, best of luck!

rkuhn:
So, finally I’m back (also mentally) :-)

My conclusion at this point is that we should add the builder change, plus a code comment explaining why we didn’t add the wrapping change (for future reference; this research has value in itself).

colinrgodsey:
sounds good!

colinrgodsey:
deferring to #17069

Issue #17008, opened 2015-03-05
On connection failure, a TcpStreamActor doesn't shut down


catalin-ursachi:
(This was encountered in a project using streams-1.0-M3; however, M4 doesn't obviously appear to have fixed it.)

If a TCP connection fails, and the TcpStreamActor managing it receives an error (e.g. ErrorClosed), the .fail(e) method is called. This method cancels the inputs and outputs, but the actor is never shut down.

The tryShutdown() method is only called if the stream completes gracefully - i.e., if .pumpFinished() is called on either the readPump or the writePump.

(If .pumpFailed(e) is called on either pump, that also simply calls fail(e) without shutting down the actor.)

Adding the line: "tryShutdown()" at the end of the .fail() method appears to solve the problem.

(Possibly related to https://github.com/akka/akka/issues/16986, https://github.com/akka/akka/issues/16900 ?)

drewhk:
Thanks for reporting!

PR #17035, opened 2015-03-12
=str #17008 Close the TcpConnectionStream on failure.

RichardBradley:
See #17008

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/io/TcpConnectionStream.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/io/TcpConnectionStream.scala
@@ -224,6 +224,7 @@ private[akka] abstract class TcpStreamActor(val settings: ActorFlowMaterializerS
     tcpOutputs.error(e)
     primaryInputs.cancel()
     primaryOutputs.error(e)
+    tryShutdown()
   }
 
   def tryShutdown(): Unit =


rkuhn:
OK TO TEST

rkuhn:
Thanks for the patch! Would you mind adding a test case that keeps us safe from breaking this again in the future? (That’s how we always do things.)

RichardBradley:
> Would you mind adding a test case that keeps us safe from breaking this again in the future? (That’s how we always do things.)

I'd like to, but I may not be able to do it very soon, and I'd like some advice on how to write such a test.

The issue is a memory leak: after the connection is finished, a `TcpStreamActor` lives on (and it can never be GCed as it's an Actor).

Are there any existing Akka tests which address Actor leaks of this kind, which I can use as a template? 

We have found that tests which address global issues like leaks don't mesh well with the tests being run in parallel by SBT. Are there any patterns in Akka tests which address that issue which I can use?

I don't see any Akka test which directly reference the `TcpStreamActor` class. I presume this class is tested indirectly. Do you know which tests cover this class indirectly?

rkuhn:
Verifying a global property is indeed problematic, but luckily we are after something very local: we only need to verify that one particular Actor shuts down. In this case you should look at StreamTcpManager to see how to create this kind of Actor, the `processorPromise` allows you to extract the Reactive Streams interface to which you can then subscribe SubscriberProbe / PublisherProbe to inject the failure case you want to see, and Actor termination can then be verified by using normal DeathWatch from a TestProbe.

The existing tests work on the stream level, you can find them in `akka-stream-tests/src/test/scala/akka/stream/io/StreamTcpSpec.scala`.

drewhk:
OK TO TEST

patriknw:
shall we merge this so that it is included in M5 and add a separate ticket for adding a test?

drewhk:
There is a ticket here: https://github.com/akka/akka/issues/15191

patriknw:
LGTM

patriknw:
@drewhk good, so then I merge this? is it LGTY?

drewhk:
Yes, this is clearly LGTM

Issue #15681, opened 2014-08-18
Provide host-level client-side API


jrudolph:
Currently, only the connection-level client API is implemented. 

The host-level API allows to specify settings per host and provides a connection pool for requests sent to this host.

The request-level API is used to execute one-off HTTP requests without having to care for any details.

See also spray's documentation of this topic: http://spray.io/documentation/1.2.1/spray-can/http-client/

Related topics to keep in mind for this feature:
  * automatic redirection handling
  * proxy support
  * proper timeouts (see outstanding spray/spray#769)

jrudolph:
If support for proxy servers and automatic redirection is added when implementing this feature, the interplay between both need to be addressed. See spray/spray#953.

jrudolph:
This is part of the bigger initiative to support a high-level HTTP client interface as tracked as #16856.

hepin1989:
connection pool in netty https://github.com/netty/netty/issues/3218 too

Issue #17004, opened 2015-03-05
Add `FlexiRoute.DemandFromNone` DemandCondition


sirthias:
I have a `FlexiRoute` use case where I need to distribute elements from upstream to an Outlet selected by a property of the *incoming element*.
This means that I need to first request an element from upstream *and then* decide which output to await demand from.
Currently I can't do that. What I can do is await `DemandFromAny` to get hold of an upstream element, but then I'd have to await `DemandFrom(selected)` to be sure that there actually is demand from the selected outlet. However, this await will only be fulfilled when the *second* upstream element arrives.

What we need to introduce is another `DemandCondition` called `DemandFromNone`, which allows me to request an upstream element without any demand from downstream. Of course, in the handler for this I would not be allowed to call `ctx.emit`.

rkuhn:
What you actually want is a FlexiKraken, but as a workaround you can use a doubler stage in front of the FlexiRoute in order to implement the scheme you sketched: demandFromAny to get the element (because if there is no demand whatsoever then you don’t need the element anyway), then demandFromX to do the real routing.

sirthias:
Yeah, I just realized that the `DemandFromNone` doesn't suffice. It's really easy to implement (just did it), but then I'd need another `DemandFrom` variant that *doesn't* await another upstream element (because I already have one). Maybe the current design of `FlexiRoute` isn't good enough for routing based on any property of the input elements. Which is somewhat of a pity. Since the "flexi" then isn't that flexi at all.
I'll go with the doubler stage approach you proposed. Thank you for that idea!

Issue #16985, opened 2015-03-03
Provide public infrastructure for using a single actor as `Flow` implementation


sirthias:
Currently akka-stream provides me with `ActorPublisher` and `ActorSubscriber` traits that I can use to easily implement a `Source` or `Sink` (resp.) manually by using these constructors:

```scala
Source.apply[T](Props): Source[T, ActorRef]
Sink.apply[T](Props): Sink[T, ActorRef]
```

I suggest we add

```scala
trait ActorPublisher[T] extends ActorSubscriber with ActorPublisher[T]
```

as well as

```scala
Flow.apply[A, B](Props): Flow[A, B, ActorRef]
```

to let me do the same for flows that are backed by a single actor.

Is there work-around that I can use with the current infrastructure to achieve the same effect?
(Other than splitting my processor logic into two actors, one for the `Sink` and one for the `Source` side.)

sirthias:
Could anyone comment on the question of whether there is a work-around currently?

My use case:
I need to implement a `Flow[A, B, _]` that wraps/manages an inner (nested) `Flow[A, B, _]`. This inner flow needs to be materialized on demand, i.e. only when the first `A` has actually arrived. Also, the inner stream is allowed to terminate (even with an error) without it affecting the outside. In this case the next `A` should simply rematerialize the inner stream.
I guess this is easiest done by completely decoupling the inner stream from the outside with an actor-based stage, but I might be missing something.

rkuhn:
Implementing the stream handling in an Actor is a complex endeavor, I’d rather get rid of ActorPublisher/ActorSubscriber than put more weight on them (yes, I do appreciate that sometimes we need them, but AFAICS only for Sinks and Sources). I think more complex transformation steps should be modeled using `mapAsync`: asking an Actor keeps that Actor oblivious to the backpressure concerns and nicely separates concerns.

sirthias:
That's not going to be good enough.
My use case:
I have a stream of this setup:

```
A ==> [B ==> C] ==> D
```

To the outside this behaves exactly like a `Flow[A, B]`.
Internally a part of the stream (the inner `Flow[B, C]`) needs to be attached with very flexible "joints" that

1. do not materialize the inner `Flow[B, C]` unless the first `B` actually arrives at the joint
2. do not allow completion or errors from the inner flow to carry over to the outside
3. rematerialize the inner flow if whenever necessary

This is very easily solved with a single actor "wrapping" the inner flow.

[...]

Issue #16819, opened 2015-02-10
Implement request timeouts on the server side


jrudolph:
In spray-can there was a two-stage request timeout that first triggered a `Timeout` message to the user handler and if still no response was received a short notice was automatically dispatched to the client.

We thought about a new approach maybe more suitable to a streaming API. 
 * there's a configured fall-back timeout on the engine level which uses the bypass to dispatch a short message after some time (30s in spray)
 * provide a routing element that allows to shorten the timeout of the outer levels and provide a custom timeout handling or message if that timeout triggers

See spray/spray#892 for more details on the discussion.

/cc @sirthias

jrudolph:
This is currently blocked on waiting for timeout support on the stream side.

ktoso:
Refs https://github.com/akka/akka/issues/17346

ktoso:
A request for this comes up once in a while frome people migrating from Spray. Perhaps we should pick this one next? Along with SSL things which are started already.

sirthias:
Ok, I'll take this one on then.

ktoso:
Great, thanks!

ktoso:
Thanks a lot @sirthias !

drewhk:
too much effort to backport, this does not make this round

PR #17036, opened 2015-03-12
htc: Preview of client-side connection pooling

sirthias:
**[Not for merge yet]**

This PR brings client-side connection pooling to akka-http-core (#15681).

The implementation as such is complete.
The only thing not yet completed (and therefore not contained yet) is the tests.

I'm putting it here to enable feedback even before all the remaining implementation kinks have been ironed out.

/cc @jrudolph

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -107,6 +107,40 @@ akka.http {
     parsing = ${akka.http.parsing}
   }
 
+  host-connection-pool {
+    # The maximum number of parallel connections that a connection pool to a
+    # single host endpoint is allowed to establish. Must be greater than zero.
+    max-connections = 4
+
+    # The maximum number of times failed requests are attempted again,
+    # (if the request can be safely retried) before giving up and returning an error.
+    max-retries = 5
+
+    # The maximum number of requests that are dispatched to the target host in
+    # batch-mode across a single connection (HTTP pipelining).
+    # A setting of 1 disables HTTP pipelining, since only one request per
+    # connection can be "in flight" at any time.
+    # Set to higher values to enable HTTP pipelining.
+    # This value must be > 0.
+    # (Note that, independently of this setting, pipelining will never be done
+    # on a connection that still has a non-idempotent request in flight.
+    # See http://tools.ietf.org/html/rfc7230#section-6.3.2 for more info.)
+    pipelining-limit = 1
+
+    # The time after which an idle connection pool (without open connections)
+    # will automatically terminate itself. Set to `infinite` to disable idle timeouts.
+    idle-timeout = 30 s
+
+    # Grace period between beginning a connection pool shutdown and the (internal)
+    # gateway actor not accepting any more requests into the pool. Protects racy
+    # requests that were dispatched to the pool after the Shutdown message was
+    # queued. Should rarely require adjustment.
+    shutdown-grace-period = 2 s
+
+    # Modify to tweak client settings for host connection pools only.
+    client = ${akka.http.client}
+  }
+
   # The (default) configuration of the HTTP message parser for the server and the client.
   # IMPORTANT: These settings (i.e. children of `akka.http.parsing`) can't be directly
   # overridden in `application.conf` to change the parser settings for client and server
--- a/akka-http-core/src/main/scala/akka/http/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/Http.scala
@@ -5,9 +5,14 @@
 package akka.http
 
 import java.net.InetSocketAddress
+import java.util.concurrent.ConcurrentHashMap
+import akka.http.util.FastFuture
 import com.typesafe.config.Config
+import scala.annotation.tailrec
+import scala.util.control.NonFatal
+import scala.util.Try
 import scala.collection.immutable
-import scala.concurrent.Future
+import scala.concurrent.{ Promise, Future }
 import akka.event.LoggingAdapter
 import akka.util.ByteString
 import akka.io.Inet
@@ -15,7 +20,7 @@ import akka.stream.FlowMaterializer
 import akka.stream.scaladsl._
 import akka.http.engine.client._
 import akka.http.engine.server._
-import akka.http.model.{ HttpResponse, HttpRequest }
+import akka.http.model._
 import akka.actor._
 
 class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.Extension {
@@ -173,6 +178,208 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
                   settings: ClientConnectionSettings,
                   log: LoggingAdapter = system.log): ClientLayer =
     BidiFlow.wrap(OutgoingConnectionBlueprint(remoteAddress, settings, log))
+
+  /**
+   * Starts a new connection pool to the given host and configuration and returns a [[Flow]] which dispatches
+   * the requests from all its materializations across this pool.
+   * While the started host connection pool internally shuts itself down automatically after the configured idle
+   * timeout it will spin itself up again if more requests arrive from an existing or a new client flow
+   * materialization. The returned flow therefore remains usable for the full lifetime of the application.
+   *
+   * Since the underlying transport usually comprises more than a single connection the produced flow might generate
+   * responses in an order that doesn't directly match the consumed requests.
+   * For example, if two requests A and B enter the flow in that order the response for B might be produced before the
+   * response for A.
+   * In order to allow for easy response-to-request association the flow takes in a custom, opaque context
+   * object of type ``T`` from the application which is emitted together with the corresponding response.
+   */
+  def newHostConnectionPool[T](host: String, port: Int = 80,
+                               options: immutable.Traversable[Inet.SocketOption] = Nil,
+                               settings: Option[HostConnectionPoolSettings] = None,
+                               defaultHeaders: List[HttpHeader] = Nil,
+                               log: LoggingAdapter = system.log)(implicit fm: FlowMaterializer): Flow[(HttpRequest, T), (Try[HttpResponse], T), HostConnectionPool] = {
+    val effectiveSettings = HostConnectionPoolSettings(settings)
+    val cps = ConnectionPoolSetup(encrypted = false, options, effectiveSettings, defaultHeaders, log)
+    val setup = HostConnectionPoolSetup(host, port, cps)
+    newHostConnectionPool(setup)
+  }
+
+  /**
+   * Starts a new connection pool to the given host and configuration and returns a [[Flow]] which dispatches
+   * the requests from all its materializations across this pool.
+   * While the started host connection pool internally shuts itself down automatically after the configured idle
+   * timeout it will spin itself up again if more requests arrive from an existing or a new client flow
+   * materialization. The returned flow therefore remains usable for the full lifetime of the application.
+   *
+   * Since the underlying transport usually comprises more than a single connection the produced flow might generate
+   * responses in an order that doesn't directly match the consumed requests.
+   * For example, if two requests A and B enter the flow in that order the response for B might be produced before the
+   * response for A.
+   * In order to allow for easy response-to-request association the flow takes in a custom, opaque context
+   * object of type ``T`` from the application which is emitted together with the corresponding response.
+   */
+  def newHostConnectionPool[T](setup: HostConnectionPoolSetup)(
+    implicit fm: FlowMaterializer): Flow[(HttpRequest, T), (Try[HttpResponse], T), HostConnectionPool] = {
+    val gateway = ConnectionPool.startHostConnectionPool(setup)
+    clientFlow[T]((_, _) ⇒ gateway).mapMaterialized(_ ⇒ HostConnectionPool(setup)(gateway))
+  }
+
+  /**
+   * Returns a [[Flow]] which dispatches incoming HTTP requests to the per-ActorSystem pool of outgoing
+   * HTTP connections to the given target host endpoint. For every ActorSystem, target host and pool
+   * configuration a separate connection pool is maintained.
+   * The HTTP layer transparently manages idle shutdown and restarting of connections pools as configured.
+   * The returned [[Flow]] instances therefore remain valid throughout the lifetime of the application.
+   *
+   * Note that the provided caching logic follows a "best-effort" design with regard to pool uniqueness.
+   * I cannot fully guarantee that the application will always run at most one pool for a given host and
+   * configuration.
+   * The cache removes entries for pools that are shut down (e.g. because of an idle timeout) and due to the
+   * inherent raciness of this operation one client might end up re-starting an old pool while another
+   * client starts a new pool if both access the cache around the time of the shutdown.
+   *
+   * Since the underlying transport usually comprises more than a single connection the produced flow might generate
+   * responses in an order that doesn't directly match the consumed requests.
+   * For example, if two requests A and B enter the flow in that order the response for B might be produced before the
+   * response for A.
+   * In order to allow for easy response-to-request association the flow takes in a custom, opaque context
+   * object of type ``T`` from the application which is emitted together with the corresponding response.
+   */
+  def cachedHostConnectionPool[T](host: String, port: Int = 80,
+                                  options: immutable.Traversable[Inet.SocketOption] = Nil,
+                                  settings: Option[HostConnectionPoolSettings] = None,
+                                  defaultHeaders: List[HttpHeader] = Nil,
+                                  log: LoggingAdapter = system.log)(implicit fm: FlowMaterializer): Flow[(HttpRequest, T), (Try[HttpResponse], T), HostConnectionPool] = {
+    val effectiveSettings = HostConnectionPoolSettings(settings)
+    val cps = ConnectionPoolSetup(encrypted = false, options, effectiveSettings, defaultHeaders, log)
+    val setup = HostConnectionPoolSetup(host, port, cps)
+    cachedHostConnectionPool(setup)
+  }
+
+  // every ActorSystem maintains its own connection pools
+  private[this] val hostPoolCache = new ConcurrentHashMap[HostConnectionPoolSetup, ConnectionPool.Gateway]
+
+  /**
+   * Returns a [[Flow]] which dispatches incoming HTTP requests to the per-ActorSystem pool of outgoing
+   * HTTP connections to the given target host endpoint. For every ActorSystem, target host and pool
+   * configuration a separate connection pool is maintained.
+   * The HTTP layer transparently manages idle shutdown and restarting of connections pools as configured.
+   * The returned [[Flow]] instances therefore remain valid throughout the lifetime of the application.
+   *
+   * Note that the provided caching logic follows a "best-effort" design with regard to pool uniqueness.
+   * I cannot fully guarantee that the application will always run at most one pool for a given host and
+   * configuration.
+   * The cache removes entries for pools that are shut down (e.g. because of an idle timeout) and due to the
+   * inherent raciness of this operation one client might end up re-starting an old pool while another
+   * client starts a new pool if both access the cache around the time of the shutdown.
+   *
+   * Since the underlying transport usually comprises more than a single connection the produced flow might generate
+   * responses in an order that doesn't directly match the consumed requests.
+   * For example, if two requests A and B enter the flow in that order the response for B might be produced before the
+   * response for A.
+   * In order to allow for easy response-to-request association the flow takes in a custom, opaque context
+   * object of type ``T`` from the application which is emitted together with the corresponding response.
+   */
+  def cachedHostConnectionPool[T](setup: HostConnectionPoolSetup)(implicit fm: FlowMaterializer): Flow[(HttpRequest, T), (Try[HttpResponse], T), HostConnectionPool] = {
+    val gateway = cachedGateway(setup)
+    clientFlow[T]((_, _) ⇒ gateway).mapMaterialized(_ ⇒ HostConnectionPool(setup)(gateway))
+  }
+
+  /**
+   * Creates a new "super connection pool flow", which routes incoming requests to a (cached) host connection pool
+   * depending on their respective effective URI. Note that incoming requests must have either an absolute URI or
+   * a valid `Host` header.
+   *
+   * Since the underlying transport usually comprises more than a single connection the produced flow might generate
+   * responses in an order that doesn't directly match the consumed requests.
+   * For example, if two requests A and B enter the flow in that order the response for B might be produced before the
+   * response for A.
+   * In order to allow for easy response-to-request association the flow takes in a custom, opaque context
+   * object of type ``T`` from the application which is emitted together with the corresponding response.
+   */
+  def superPool[T](options: immutable.Traversable[Inet.SocketOption] = Nil,
+                   settings: Option[HostConnectionPoolSettings] = None,
+                   defaultHeaders: List[HttpHeader] = Nil,
+                   log: LoggingAdapter = system.log)(implicit fm: FlowMaterializer): Flow[(HttpRequest, T), (Try[HttpResponse], T), Any] = {
+    val effectiveSettings = HostConnectionPoolSettings(settings)
+    val setup = ConnectionPoolSetup(encrypted = false, options, effectiveSettings, defaultHeaders, log)
+    clientFlow[T] { (request, userContext) ⇒
+      val effectiveRequest = request.withEffectiveUri(securedConnection = false)
+      val Uri.Authority(host, port, _) = effectiveRequest.uri.authority
+      cachedGateway(HostConnectionPoolSetup(host.toString(), port, setup))
+    }
+  }
+
+  /**
+   * Fires a single [[HttpRequest]] across the (potentially cached) host connection pool for the request's
+   * effective URI to produce a response future.
+   *
+   * Note that the request must have either an absolute URI or a valid `Host` header, otherwise
+   * the future will be completed with an error.
+   */
+  def singleRequest(request: HttpRequest,
+                    options: immutable.Traversable[Inet.SocketOption] = Nil,
+                    settings: Option[HostConnectionPoolSettings] = None,
+                    defaultHeaders: List[HttpHeader] = Nil,
+                    log: LoggingAdapter = system.log)(implicit fm: FlowMaterializer): Future[HttpResponse] =
+    try {
+      val effectiveSettings = HostConnectionPoolSettings(settings)
+      val setup = ConnectionPoolSetup(encrypted = false, options, effectiveSettings, defaultHeaders, log)
+      val effectiveRequest = request.withEffectiveUri(securedConnection = false)
+      val uri = effectiveRequest.uri
+      val gateway = cachedGateway(HostConnectionPoolSetup(uri.authority.host.toString(), uri.effectivePort, setup))
+      val responsePromise = Promise[HttpResponse]()
+      gateway.actorRef() ! ConnectionPool.PoolRequest(request, responsePromise)
+      responsePromise.future
+    } catch {
+      case e: IllegalUriException ⇒ FastFuture.failed(e)
+    }
+
+  /**
+   * Triggers an orderly shutdown of all host connections pools currently maintained by the [[ActorSystem]].
+   * The returned future is completed when all pools that were live at the time of this method call
+   * have completed their shutdown process.
+   *
+   * If existing pool client flows are re-used or new ones materialized concurrently with or after this
+   * method call the respective connection pools will be restarted and not contribute to the returned future.
+   */
+  def shutdownAllConnectionPools(): Future[Unit] = {
+    import scala.collection.JavaConverters._
+    import system.dispatcher
+    val gateways = hostPoolCache.values().asScala
+    system.log.info("Initiating orderly shutdown of all active host connections pools (w/ configured grace period)...")
+    Future.sequence(gateways.map(_.shutdown())).map(_ ⇒ ())
+  }
+
+  @tailrec
+  private def cachedGateway(setup: HostConnectionPoolSetup)(implicit fm: FlowMaterializer): ConnectionPool.Gateway =
+    hostPoolCache.putIfAbsent(setup, ConnectionPool.CowboyGateway) match {
+      case null ⇒ // we are the first to attempt starting a pool for this setup
+        try {
+          val gateway = ConnectionPool.startHostConnectionPool(setup)
+          hostPoolCache.put(setup, gateway)
+          gateway.whenShuttingDown.onComplete(_ ⇒ hostPoolCache.remove(setup))(system.dispatcher)
+          gateway
+        } catch {
+          case NonFatal(e) ⇒
+            hostPoolCache.remove(setup) // remove Cowboy entry if we weren't able to properly start the pool
+            throw e
+        }
+
+      case ConnectionPool.CowboyGateway ⇒ cachedGateway(setup) // spin while other thread is starting a pool for this setup
+      case pool                         ⇒ pool // return cached instance
+    }
+
+  private def clientFlow[T](poolGateway: (HttpRequest, Any) ⇒ ConnectionPool.Gateway)(
+    implicit system: ActorSystem, fm: FlowMaterializer): Flow[(HttpRequest, T), (Try[HttpResponse], T), Any] =
+    Flow[(HttpRequest, T)].mapAsync(parallelism = 4, {
+      case (request, userContext) ⇒
+        val responsePromise = Promise[HttpResponse]()
+        poolGateway(request, userContext).actorRef() ! ConnectionPool.PoolRequest(request, responsePromise)
+        val result = Promise[(Try[HttpResponse], T)]()
+        responsePromise.future.onComplete(responseTry ⇒ result.success(responseTry -> userContext))(system.dispatcher)
+        result.future
+    })
 }
 
 object Http extends ExtensionId[HttpExt] with ExtensionIdProvider {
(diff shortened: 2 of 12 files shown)


akka-http-core/src/main/resources/reference.conf
```diff
@@ -107,6 +107,34 @@ akka.http {
...
+    # on a connection that still has a non-idempotent request in flight.
+    # See http://tools.ietf.org/html/rfc7230#section-6.3.2 for more info.)
+    pipelining-limit = 1
+
+    # The time after which an idle connection pool (without open connections)
+    # will automatically terminate itself. Set to `infinite` to disable idle timeouts.
+    idle-timeout = 30 s
+
+    # Modify to tweak client settings for host connection pools only.
+    client = ${akka.http.client}
```
hepin1989: connection timeout?
total-max-connection?
read timeout?

akka-http-core/src/main/resources/reference.conf
```diff
@@ -107,6 +107,34 @@ akka.http {
     parsing = ${akka.http.parsing}
   }
 
+  host-connection-pool {
+    # The maximum number of parallel connections that a connection pool to a
+    # single host endpoint is allowed to establish. Must be greater than zero.
+    max-connections = 4
```
hepin1989: max-connections-per-host
&max-connections-total?

sirthias: This value is what you call "max-connections-per-host". There is no global limit.

akka-http-core/src/main/resources/reference.conf
sirthias: The `connecting-timeout` and `idle-timeout` for the connections in the pool can be configured here.

akka-http-core/src/main/resources/reference.conf
hepin1989: looking forward to it.I come up with this from the AHC

jrudolph:
This already looks quite advanced! Do we need request timeouts on this level? I'd say it would make sense to abort and reconnect a slot that hangs (by whatever condition it is determined).

My other question would be what the overall demand behavior is/should be. I'd say there should be a simple one-to-one mapping from upstream demand to downstream demand. I think this is important so that a higher-level queuing logic can decide (wrt request timeouts, request ordering, etc.) promptly what to do with a request. If request queue up in hidden buffers somewhere we may end up in a similar state like in spray where fine-grained queue-handling was hard to achieve. (Maybe this is already implemented like this and just needs tests.)

sirthias:
> Do we need request timeouts on this level? I'd say it would make sense to abort and reconnect a slot that hangs (by whatever condition it is determined).

I think we need support for request-timeouts of the client-side connection flow level anyway (pretty much the client-side dual of #16819, which we are still missing as a ticket).
We already have `connecting-timeout` and `idle-timeout` on the client side, the latter of which is still dysfunctional I think.
Once we have all these on the connection-level they will be automatically be effective in the host connection pool level as well.

> My other question would be what the overall demand behavior is/should be. I'd say there should be a simple one-to-one mapping from upstream demand to downstream demand. I think this is important so that a higher-level queuing logic can decide (wrt request timeouts, request ordering, etc.) promptly what to do with a request. If request queue up in hidden buffers somewhere we may end up in a similar state like in spray where fine-grained queue-handling was hard to achieve. (Maybe this is already implemented like this and just needs tests.)

The pool implements a running stream from `(HttpRequest, T)` to `(HttpResponse, T)]` that "pool clients" can attach to. Demand from all clients is merged into the pool stream and the pool demands from the attached clients in a fair fashion, i.e. a client that is ready to send requests will eventually get a chance to do it.
Unfortunately, because the pool can't predict which response will be produced next response demand from the client can only be forwarded into the pool when *all* clients are ready for the next response.
This means, that a single slow client that is attached to a pool, slows down all other clients as well.
(All this logic is implemented in the `PoolGateway` actor.)

We can improve the currently implemented logic a bit though:
If a client currently has no request "in flight" through the pool then we know that the next response coming out of the pool cannot be for that client. Therefore we can ignore "non-demand" from that client and still forward the response demand from the other clients to the pool.
This however requires that we buffer one request per pool client, which is ok IMHO. I didn't want to implement that a bit more complex logic right now though, and first stabilize everything else.

sirthias:
Updated once more, this time with the first (yet untested) implementation of the top-level (request-level) API. Also, surfaced uncached host connection pool construction to the user.

Now suspending work on this until https://github.com/akka/akka/issues/16985 is cleared, as it is crucial for the host connection pool implementation.

akka-http-core/src/main/scala/akka/http/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -23,9 +23,9 @@ import akka.http.util._
 /**
  * INTERNAL API
  */
-private[http] object HttpClient {
+private[http] object OutgoingConnectionBlueprint {
 
-  case class HttpClientPorts(
+  case class Ports(
```
rkuhn: This should then go away and be replaced with a BidiShape/BidiFlow, right?

sirthias: That would be nice... :)

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -128,44 +138,182 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
+            // dead; in this case the returned ClientFlow will always be completed immediately after materialization
+            hostPool.isShutdown.onComplete(_ ⇒ hostPoolCache.remove(setup))(system.dispatcher)
+            hostPool
+          } catch {
+            case NonFatal(e) ⇒
+              hostPoolCache.remove(setup) // remove Cowboy entry if we weren't able to properly start the pool
+              throw e
+          }
+
+        case CowboyEntry ⇒ { Thread.`yield`(); hostPool() } // spin while other thread is creating a HostPool for this setup
```
rkuhn: `Thread.yield()` is not the best here. In other places we just spin, but might as well `Thread.sleep(1)` (or `LockSupport.parkNanos(...)` if that is too long for your taste).

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -128,44 +138,182 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
+    @tailrec def hostPool(): HostPool =
+      hostPoolCache.putIfAbsent(setup, CowboyEntry) match {
+        case null ⇒ // we are the first to attempt construction of the HostPool for this setup
+          try {
+            val hostPool = ConnectionPool.hostPool(setup)
+            hostPoolCache.put(setup, hostPool)
+            // we remove the cache entry when the pool has shut down; however, due to the inherent raciness, a client
+            // might still end up receiving a client flow for a pool which is in the process of shutting down or already
+            // dead; in this case the returned ClientFlow will always be completed immediately after materialization
+            hostPool.isShutdown.onComplete(_ ⇒ hostPoolCache.remove(setup))(system.dispatcher)
```
rkuhn: the Future should probably be named `whenShutdown`, `isShutdown` sounds a little boolean to me :-)

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -128,44 +138,182 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
+  }
+
+  /**
+   * Triggers an orderly shutdown of all host connections pools currently maintained by the [[ActorSystem]].
+   */
+  def shutdownAllConnectionPools(): Future[Unit] = {
+    import scala.collection.JavaConverters._
+    import system.dispatcher
+    val hostPools = hostPoolCache.values().asScala
+    Future.sequence(hostPools.map(_.doShutdown())).map(_ ⇒ ())
```
rkuhn: if this is supposed to be reliable, then it needs to guard against concurrent or subsequent pool creation as well

akka-http-core/src/main/scala/akka/http/engine/client/OutgoingConnectionBlueprint.scala
rkuhn: https://github.com/akka/akka/blob/release-2.3-dev/akka-stream/src/main/scala/akka/stream/scaladsl/BidiFlow.scala

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -116,7 +116,7 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
     else new Flow(module.growConnect(op, shape.outlet, op.inPort, Keep.right).replaceShape(FlowShape(shape.inlet, op.outPort)))
   }
 
-  private[stream] def andThenMat[U, Mat2, O >: Out](processorFactory: () ⇒ (Processor[O, U], Mat2)): Repr[U, Mat2] = {
+  private[akka] def andThenMat[U, Mat2, O >: Out](processorFactory: () ⇒ (Processor[O, U], Mat2)): Repr[U, Mat2] = {
```
rkuhn: fair enough: I was going to suggest this

akka-stream/src/main/boilerplate/akka/stream/FanInShape.scala.template
```diff
@@ -57,6 +57,21 @@ class UniformFanInShape[T, O](val n: Int, _init: FanInShape.Init[O]) extends Fan
...
+  override def deepCopy(): FanInShape1N[T0, T1, O] = super.deepCopy().asInstanceOf[FanInShape1N[T0, T1, O]]
+
+  val in0 = newInlet[T0]("in0")
+  val in1Vector: Vector[Inlet[T1]] = Vector.tabulate(n)(i => newInlet[T1](s"in${i+1}"))
+  def in(n: Int): Inlet[T1] = {
+    require(n > 0, "n must be > 0")
+    in1Vector(n - 1)
+  }
+}
+
```
rkuhn: This does not seem to be used.

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
```diff
@@ -0,0 +1,213 @@
...
+private[http] object ConnectionPool {
+  import HostConnectionPoolGateway._
+
+  type ClientFlow = Flow[(HttpRequest, Any), (Try[HttpResponse], Any), Http.HostConnectionPool]
+
+  final class HostPool(flow: ClientFlow, val isShutdown: Future[Unit], val doShutdown: () ⇒ Future[Unit]) {
+    def clientFlow[T] = flow.asInstanceOf[Flow[(HttpRequest, T), (Try[HttpResponse], T), Http.HostConnectionPool]]
+  }
+
+  val CowboyEntry = new HostPool(null, null, null) // for reigning in potentially approaching thundering herds
```
rkuhn: And the cowboy shouts: yield! yield! :-)

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
```diff
@@ -0,0 +1,213 @@
...
+  private val poolClientActorName = new SeqActorName("PoolClient")
+  private val poolGatwayActorName = new SeqActorName("PoolGateway")
+
+  /**
+   * Creates a new "super pool flow", which routes incoming requests to "their" host connection pool,
+   * depending on the respective effective URI.
+   */
+  def superPool[T](setup: ConnectionPoolSetup)(implicit system: ActorSystem, fm: FlowMaterializer): Flow[(HttpRequest, T), (Try[HttpResponse], T), Unit] = {
+    Flow.wrap {
+      FlowGraph.partial() { implicit b ⇒
```
rkuhn: This is written as
~~~scala
Flow() { implicit b =>
  ...
}
~~~

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
```diff
@@ -0,0 +1,213 @@
...
+        import FlowGraph.Implicits._
+        val map = b.add(Flow[(HttpRequest, T)].map {
+          case (r, ctx) ⇒ Try(r.withEffectiveUri(securedConnection = false)) -> ctx
+        })
+        val failureSplit = b.add(new FailureSplit[T])
+        val mainFlow = Flow[(HttpRequest, T)]
+          .groupBy { case (req, _) ⇒ req.uri.authority }
+          .map {
+            case (Uri.Authority(host, port, _), hostRequests) ⇒
+              val poolClientFlow = Http().cachedHostConnectionPool[T](HostConnectionPoolSetup(host.toString(), port, setup))
```
rkuhn: This will not work after a cached pool timed out: the `groupBy` group will not be closed and this will not run again.

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
```diff
@@ -0,0 +1,213 @@
...
+              hostRequests.via(poolClientFlow)
+          }
+          .flatten(FlattenStrategy.Concat()) // WRONG! FIXME: Needs to be `FlattenStrategy.Merge` instead! (#15089)
```
rkuhn: This might in theory work, but it is a rather costly way of formulating this: as we discussed this would be better off using `mapAsync` and using a single-request–response protocol. Trying to amortize the stream materialization costs will lead into big troubles (idle timeouts shutting things down, which is nasty to get to work consistently; maximum number of open groups in this expression; both connected to double-bookkeeping here and in cachedHostConnectionPool).

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
```diff
@@ -0,0 +1,213 @@
...
+   * from the gateway. The connection pool stream is kept running, as it may be serving other clients now or in the
+   * future.
+   *
+   * A running host connection pool is stopped after the configured idle timeout
+   */
+  def hostPool(setup: HostConnectionPoolSetup)(implicit system: ActorSystem, fm: FlowMaterializer): HostPool = {
+    val shutdownPromise = Promise[Unit]()
+    val poolGateway = startHostConnectionPool(setup, shutdownPromise)
+    val doShutdown: () ⇒ Future[Unit] = { () ⇒
+      if (!shutdownPromise.isCompleted) // optimization: avoid known dead letter (not guaranteed to work)
```
rkuhn: proper fix: `case object Shutdown extends DeadLetterSuppression`

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
```diff
@@ -0,0 +1,213 @@
...
+        poolGateway ! Shutdown
+      shutdownPromise.future
+    }
+    val flow = Flow[(HttpRequest, Any)] andThenMat { () ⇒
+      val defaultHeaders = prepareDefaultHeaders(setup)
+      val actor = system.actorOf(Props(new PoolClient(poolGateway, defaultHeaders, setup.setup.settings.maxRetries)),
```
rkuhn: This should be `fm.actorOf`.

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
```diff
@@ -0,0 +1,213 @@
...
+        poolClientActorName.next())
+      val mat = Http.HostConnectionPool(setup)(doShutdown)
+      ActorProcessor[(HttpRequest, Any), (Try[HttpResponse], Any)](actor) -> mat
+    }
```
rkuhn: Okay, here’s a better plan: after finishing [AsyncStage](https://github.com/akka/akka/pull/17105) we’ll add support for transformation stages that materialize to a value, and then you can express all of this without a PoolClient actor.

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
rkuhn: To avoid misinterpretations: what I’m saying is that the CHM should contain ActorRefs, not Flows.

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
```diff
@@ -0,0 +1,213 @@
...
+    import hcps._
+    import setup._
+    val connectionFlow = Http().outgoingConnection(host, port, None, options, Some(settings.connectionSettings), log)
+    val poolFlow = newPoolFlow(connectionFlow, new InetSocketAddress(host, port), settings, log)
+    val gatewayFlow =
+      Flow[ResponseContext] andThenMat { () ⇒
+        val actor = system.actorOf(Props(new PoolGateway(settings.idleTimeout, shutdownPromise)), poolGatwayActorName.next())
+        ActorProcessor[ResponseContext, RequestContext](actor) -> actor
+      }
+    gatewayFlow.joinMat(poolFlow)(Keep.left).run()
```
rkuhn: `Keep.left` is always the default (if it isn’t then that’s a bug).

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
```diff
@@ -0,0 +1,187 @@
...
+  def apply(slotIx: Int, connectionFlow: Flow[HttpRequest, HttpResponse, Any],
+            remoteAddress: InetSocketAddress, // TODO: remove after #16168 is cleared
+            settings: HostConnectionPoolSettings)(implicit system: ActorSystem, fm: FlowMaterializer): Graph[Ports, Unit] =
+    FlowGraph.partial() { implicit b ⇒
+      import FlowGraph.Implicits._
+
+      val slotProcessor = b.add {
+        Flow[RequestContext] andThenMat { () ⇒
+          val actor = system.actorOf(Props(new SlotProcessor(slotIx, connectionFlow, settings)), slotProcessorActorName.next())
+          ActorProcessor[RequestContext, List[ProcessorOut]](actor) -> ()
```
rkuhn: Okay, this is the one that needs a bit more thought, I feel that I’m not far from coming up with a good idea, hold on to your hats!

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: Ok, no prob.
Do you happen to have a pointer to some good background material on `Thread.yield()` and the difference to the alternatives?

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: Ok, agreed.

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: It's always going to be a racy thing, so, unless we want `shutdownAllConnectionPools` to be "final" in the sense that you cannot start a new pool after having called it, I'm not sure that there is value in investing more "best effort" than we do here.

akka-http-core/src/main/scala/akka/http/engine/client/OutgoingConnectionBlueprint.scala
sirthias: So the Bidi-stuff is already ready and useable? Cool!

akka-stream/src/main/boilerplate/akka/stream/FanInShape.scala.template
sirthias: Yes. It used to be used in an intermediate version, which got killed later.
However, I kept this around because it might be of general utility.
If you don't like it I'll simply remove it.

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
sirthias: And now you need to explain why "Sleep! Sleep!" or "Park! Park!" is better than "Yield! Yield!"... :)

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
sirthias: ok

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
sirthias: True. That's a bummer.
I was thinking that, for once, we have a nice application of `groupBy`... :)

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
sirthias: Yes. I'll give the `mapAsync` approach a spin.
I have the feeling that it might even be somewhat more concise than the current design.

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
sirthias: Ok

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
sirthias: I wasn't sure whether the `fm` is supposed to be "hijackable" for actor creation.
After all, we might have FlowMaterializers later which are actor-less and have no idea of how to create an actor.

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
sirthias: My feeling is that, once we move to `mapAsync` the need for the `PoolClient` actor would go away anyway.
So I don't yet see where I'd need a "transformation stage that materialized into a value". But I guess I'll find out once I get there.

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
sirthias: Ok.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
sirthias: Hats held on to!

rkuhn:
how’s this coming along?

sirthias:
Things a working for me now on my machine. Nothing left to do from your side.
I'm in the process of getting the initial test suite completed.

rkuhn:
Cool!

drewhk:
wonderful!

sirthias:
Ok, I just pushed the current state, which is basically done.
We still need a lot more tests, which are in the works.
But everything else is complete, so it's not too early to review (if you need something to spend your weekend with... :)

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -173,6 +178,208 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
+   * responses in an order that doesn't directly match the consumed requests.
+   * For example, if two requests A and B enter the flow in that order the response for B might be produced before the
+   * response for A.
+   * In order to allow for easy response-to-request association the flow takes in a custom, opaque context
+   * object of type ``T`` from the application which is emitted together with the corresponding response.
+   */
+  def newHostConnectionPool[T](host: String, port: Int = 80,
+                               options: immutable.Traversable[Inet.SocketOption] = Nil,
+                               settings: Option[HostConnectionPoolSettings] = None,
+                               defaultHeaders: List[HttpHeader] = Nil,
```
patriknw: `List` vs `immutable.Seq`?

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
```diff
@@ -0,0 +1,158 @@
...
+      val fut = whenShuttingDown
+      if (!fut.isCompleted) // try to avoid a restart just for shutting down
+        actorRef() ! ConnectionPool.BeginShutdown
+      fut.flatMap(identityFunc)(system.dispatcher)
+    }
+
+    // retrieves the ActorRef for the running Gateway actor
+    // or starts a new one it if the gateway is already in (or finished with) the shutdown process
+    @tailrec def actorRef(): ActorRef = {
+      var ref = cell.get
```
patriknw: does this have to be a `var`?

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    override def onError(cause: Throwable): Unit = {
+      exposedPublisher.shutdown(Some(cause))
+      super.onError(cause)
+      shutdown()
+    }
+
+    def shutdown(): Unit = {
+      context.become { case _ ⇒ }
```
patriknw: why?

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -173,6 +178,208 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
                   settings: ClientConnectionSettings,
                   log: LoggingAdapter = system.log): ClientLayer =
     BidiFlow.wrap(OutgoingConnectionBlueprint(remoteAddress, settings, log))
+
+  /**
+   * Starts a new connection pool to the given host and configuration and returns a [[Flow]] which dispatches
+   * the requests from all its materializations across this pool.
```
drewhk: Different materializations' responses are delivered only to the relevant materialization?

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -173,6 +178,208 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
+
+  /**
+   * Returns a [[Flow]] which dispatches incoming HTTP requests to the per-ActorSystem pool of outgoing
+   * HTTP connections to the given target host endpoint. For every ActorSystem, target host and pool
+   * configuration a separate connection pool is maintained.
+   * The HTTP layer transparently manages idle shutdown and restarting of connections pools as configured.
+   * The returned [[Flow]] instances therefore remain valid throughout the lifetime of the application.
+   *
+   * Note that the provided caching logic follows a "best-effort" design with regard to pool uniqueness.
+   * I cannot fully guarantee that the application will always run at most one pool for a given host and
```
drewhk: "It"

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -173,6 +178,208 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
```
drewhk: "It"

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolConductor.scala
```diff
@@ -0,0 +1,230 @@
...
+      merge1.out ~> merge2.in0
+      merge2.out ~> doubler ~> route.in
+      retrySplit.out0 ~> merge2.in1
+      retrySplit.out1 ~> merge1.in1
+
+      Ports(merge1.in0, retrySplit.in, route.outlets.asInstanceOf[immutable.Seq[Outlet[RequestContext]]])
+    }
+
+  // almost the same as a MergePreferred,
+  // but as MergePreferred doesn't appear to propagate completion on the secondary input we can use it here
```
drewhk: but this does not prefer any inputs? You are only interested in the eager close right?

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolConductor.scala
```diff
@@ -0,0 +1,230 @@
...
+  private case class SwitchCommand(rc: RequestContext, slotIx: Int)
+
+  private sealed trait SlotState
+  private case object Unconnected extends SlotState
+  private case object Idle extends SlotState
+  private final case class Loaded(openIdempotentRequests: Int) extends SlotState { require(openIdempotentRequests > 0) }
+  private case class Busy(openRequests: Int) extends SlotState { require(openRequests > 0) }
+  private object Busy extends Busy(1)
+
+  private class Merge2(slotCount: Int, maxRetries: Int, pipeliningLimit: Int, log: LoggingAdapter)
```
drewhk: This is a nice class, can we give it a more meaningful name? SlotAllocator or somtething like that?

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolConductor.scala
```diff
@@ -0,0 +1,230 @@
...
+                                  +---------+
+
+  */
+  def apply(slotCount: Int, maxRetries: Int, pipeliningLimit: Int, log: LoggingAdapter): Graph[Ports, Any] =
+    FlowGraph.partial() { implicit b ⇒
+      import FlowGraph.Implicits._
+
+      val merge1 = b.add(new Merge1)
+      val merge2 = b.add(new Merge2(slotCount, maxRetries, pipeliningLimit, log))
+      val doubler = Flow[SwitchCommand].mapConcat(x ⇒ x :: x :: Nil)
```
drewhk: what is the purpose of this doubler?

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: `immutable.Seq` it shall be.

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
sirthias: Nope. Appears to be a left-over from something earlier.
Good catch.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
sirthias: If we are in the `unconnected` state and have a `Cancel` (sent by the downstream) followed by an `OnNext` (sent by the upstream) in the mailbox, we'll end up calling `shutdown()` and `context.stop(self)`. The scaladoc to the latter says that `context.stop` is asynchronous, which means that we might get to handling the `OnNext` before being stopped.
However, the `OnNext` handler in `unconnected` materializes a new `runnableFlow` (which opens a new HTTP connection), that is not automatically stopped/collected when the `SlotProcessor` actor is stopped.

I guess a comment explaining this is warranted here... :)

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: Yes. Each request creates a new response promise for itself which it `mapAsync`s against. So all responses fulfill the response future of their respective request. The materializations only provide the context for the `mapAsync` stages (and their `parallelism` scope).

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: Yeah, I guess the impersonal "it" fits better... :)
(Even though I think that a somewhat more personal touch really does spice up code docs/comments!)

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolConductor.scala
sirthias: Actually I want a "prefer retryInlet merge" here but wasn't sure whether a `FlexiMerge` can actually do something like a `MergePreferred`. I'll improve the comment to make that clearer.

(And the comment has a typo, it should say "... we *can't* use it here".)

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolConductor.scala
sirthias: Absolutely, good idea!
I'll also add more internal docs for the class and the slot state model.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolConductor.scala
sirthias: It's a work-around for https://github.com/akka/akka/issues/17004.
I'll add a respective comment.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
rkuhn: Nope, “async” does not imply that other activity may happen between now and then—removing this line is perfectly safe.

patriknw: yupp, no more messages will be received after stop, i.e. the behavior is irrelevant

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolConductor.scala
drewhk: ok thanks

drewhk:
looks good so far, I couldn't get to decode all the details.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
sirthias: Ok, great. This makes sense and was also my initial assumption (as the difference to a `PoisonPill` is clear).
However, I was tripped up by the scaladoc on `Context::stop`, which is:

> Stop the actor pointed to by the given [[akka.actor.ActorRef]]; this is
> an asynchronous operation, i.e. involves a message send.

Should this scaladoc be improved?

rkuhn: yes, it should: #17206

sirthias:
Superseded by #17222.

PR #17041, opened 2015-03-14
+htc #16972 makes bindAndStartHandlingWith actually run()

### LINKED ISSUE #16972: Http().bindAndStartHandlingWith does not work
there’s a `.run()` missing

ktoso:
Resolves #16972

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/Http.scala
@@ -60,13 +60,13 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
    * connections are being accepted at maximum rate, which, depending on the applications, might
    * present a DoS risk!
    */
-  def bindAndstartHandlingWith(handler: Flow[HttpRequest, HttpResponse, _],
+  def bindAndStartHandlingWith(handler: Flow[HttpRequest, HttpResponse, _],
                                interface: String, port: Int = 80, backlog: Int = 100,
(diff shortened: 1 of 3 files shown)


rkuhn:
LGTM

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -60,13 +60,13 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
    * connections are being accepted at maximum rate, which, depending on the applications, might
    * present a DoS risk!
    */
-  def bindAndstartHandlingWith(handler: Flow[HttpRequest, HttpResponse, _],
+  def bindAndStartHandlingWith(handler: Flow[HttpRequest, HttpResponse, _],
```
ktoso: there was also the question in #16972 on if this method should be called `bindAndHandle` to be more similar to TCP.
// cc @rkuhn @jrudolph @hseeberger @sirthias

hseeberger: Please!

hseeberger: `theseTerriblyLongNamesRemindMeOfMyJeeSpringDays`

rkuhn: ah, right; yes, this makes sense.

ktoso: Yeap, PR here https://github.com/akka/akka/pull/17042

PR #17050, opened 2015-03-18
+act #17012 improve err msg on invalid actor name/path

### LINKED ISSUE #17012: Make actor name errors show the illegal character
Since we no longer use a regexp to verify actor names we can actually show the character that is illegal as well, which is useful. Some illegal characters are hard to notice, e.g. from the message ```Illegal actor name ["2w10"]. Actor paths MUST: not start with `$`, include only ASCII letters and can only contain these special characters: -_.*$+:@&=,!~';.``` it might not be obvious that the name includes the character `"` that is illegal.

ktoso:
Resolves #17012

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/ActorPath.scala
+++ b/akka-actor/src/main/scala/akka/actor/ActorPath.scala
@@ -2,7 +2,7 @@
  * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
  */
 package akka.actor
-import scala.annotation.tailrec
+import scala.annotation.{ switch, tailrec }
 import scala.collection.immutable
 import akka.japi.Util.immutableSeq
 import java.net.MalformedURLException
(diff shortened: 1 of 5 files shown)


ktoso:
Performance wise checking the "is valid or not", was:

```
[info] a.a.ActorPathValidationBenchmark.handLoopActor_1    thrpt       20  36.619        1.261  ops/us
```

and now it's 

```
[info] a.a.ActorPathValidationBenchmark.handLoopActor_1    thrpt       40  41.417        0.604  ops/us
```

So seems actually better, though the only thing here we care about is not getting worse IMO.

bantonsson:
LGTM

akka-actor/src/main/scala/akka/actor/ActorPath.scala
```diff
@@ -20,6 +20,27 @@ object ActorPath {
...
+   * @param element actor path element to be validated
+   * @param fullPath optional fullPath element that may be included for better error messages; null if not given
+   */
+  final def validatePathElement(element: String, fullPath: String = null): Unit =
+    findInvalidPathElementCharPosition(element) match {
+      case ValidPath ⇒ // valid
+      case invalidAt ⇒
+        val fullPathMsg = if (fullPath ne null) s""" (in path [$fullPath])""" else ""
+        throw new InvalidActorNameException(
+          s"""Invalid actor path element [$element]$fullPathMsg, illegal character [${element(invalidAt)}] at position: $invalidAt. """ +
```
patriknw: what happens for `element = ""`?
as far as I can see `findInvalidPathElementCharPosition` will return 0 and `element(invalidAt)` will fail.

ktoso: good catch, `""` in current code would never reach here, we catch it earlier but the method should probably be ready for it anyway. Thanks!

akka-actor/src/main/scala/akka/actor/ActorPath.scala
```diff
@@ -20,6 +20,27 @@ object ActorPath {
...
+
+  /**
+   * Validates the given actor path element and throws an [[InvalidActorNameException]] if invalid.
+   * See [[isValidPathElement()]] for a non-throwing version.
+   *
+   * @param element actor path element to be validated
+   * @param fullPath optional fullPath element that may be included for better error messages; null if not given
+   */
+  final def validatePathElement(element: String, fullPath: String = null): Unit =
+    findInvalidPathElementCharPosition(element) match {
```
drewhk: a simple if-then maybe?

drewhk:
LGTM

akka-actor/src/main/scala/akka/actor/ActorPath.scala
ktoso: made sure it's a `@switch`

akka-actor/src/main/scala/akka/actor/ActorPath.scala
ktoso: Fixed now.

ktoso:
Addressed review comments and merging.

Issue #16938, opened 2015-02-26
Fix GraphConcatSpec


drewhk:
There are two test cases currently marked as pending because they relied on materialization order of the old materializer. These tests should be revived and adapted to the new changes.

bantonsson:
I don't see the issue with the tests, since they test the opposite wiring of each other and should cover both materialization cases, unless our materialization order is random.

drewhk:
Our materialization order is currently non-deterministic (not necessarily random)

bantonsson:
Let me rephrase that. If I have a graph with exactly the same shape, do you mean that two different materializations can have different order of materialization?

bantonsson:
Even more narrow. If I materialize the same graph twice, can the order be different?

drewhk:
Probably not on the same JVM run, but it depends on HashSet ordering.

bantonsson:
Check, then my comment is broken. I'll have to read up on the materialization code.

drewhk:
Enough to look at MaterializerSession it does the traversal.

bantonsson:
I just took a look at this, and it is deterministic since the Set only contains 3 submodules, and becomes an optimized Set3, that traverses in insertion order :wink:

I added a comment to the tests.

PR #17061, opened 2015-03-23
=str #16938 Enable more tests in GraphConcatSpec

bantonsson:
Refs #16938

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphConcatSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphConcatSpec.scala
@@ -103,8 +103,11 @@ class GraphConcatSpec extends TwoStreamsSetup {
       subscriber2.expectSubscriptionAndError(TestException)
     }
 
-    "work with one delayed failed and first nonempty publisher" in {
-      pending // FIXME: This relies on materialization order!!
+    "work with one nonempty and one delayed failed publisher" in {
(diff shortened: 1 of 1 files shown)


bantonsson:
I made this comment in the issue:

> I don't see the issue with the tests, since they test the opposite wiring of each other and should cover both materialization cases, unless our materialization order is random.

bantonsson:
So @drewhk explained how the materializer works and invalidated my previous comment. I quickly double checked this case and it works since the graphs created only contains 3 submodules that are put into a Set that becomes an optimized Set3, that traverses in insertion order.

I added a comment to the tests.

patriknw:
LGTM

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphConcatSpec.scala
```diff
@@ -103,8 +103,11 @@ class GraphConcatSpec extends TwoStreamsSetup {
       subscriber2.expectSubscriptionAndError(TestException)
     }
 
-    "work with one delayed failed and first nonempty publisher" in {
-      pending // FIXME: This relies on materialization order!!
+    "work with one nonempty and one delayed failed publisher" in {
+      // This test and the next one are materialization order dependent and rely on the fact
```
drewhk: we might break this ordering though, but I guess the comment is enough so the first one encountering the issue can modify it accordingly.

drewhk:
LGTM

Issue #15677, opened 2014-08-17
DOC: Document packaging akka apps


ktoso:
Perhaps simply by linking to http://www.scala-sbt.org/sbt-native-packager/index.html

Users asked for this as in `2.2.x` we had some tips (and the sbt plugin) documented at http://doc.akka.io/docs/akka/2.2.1/scala/microkernel.html

PR #17065, opened 2015-03-23
=doc #15677 Document how to use native-packager

patriknw:
- replace docs of AkkaAppPackaging with JavaServerAppPackaging,
  since the former doesn't have the same flexibility and quality
- warn about microkernel deprecation

### DIFF
--- a/akka-docs/rst/scala/microkernel.rst
+++ b/akka-docs/rst/scala/microkernel.rst
@@ -2,15 +2,15 @@
 .. _microkernel-scala:
 
 Microkernel
-===================
+===========
 
 The purpose of the Akka Microkernel is to offer a bundling mechanism so that you can distribute
 an Akka application as a single payload, without the need to run in a Java Application Server or manually
 having to create a launcher script.
 
-The Akka Microkernel is included in the Akka download found at `downloads`_.
-
-.. _downloads: http://akka.io/downloads
+.. warning:: Akka Microkernel will be deprecated and removed. It will be replaced by using an ordinary
+   user defined main class and packaging with `sbt-native-packager <https://github.com/sbt/sbt-native-packager>`_
+   or `Typesafe ConductR <http://typesafe.com/products/conductr>`_.
 
 To run an application with the microkernel you need to create a Bootable class 
 that handles the startup and shutdown the application.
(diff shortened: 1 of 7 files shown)


bantonsson:
@patriknw I think you should `s/latter/former/` in the PR comment since it's the `AkkaAppPackaging` that is not so good.

LGTM

patriknw:
@bantonsson good catch :)

ktoso:
This would be a nice candidate to plugin ConductR's "all in one" plugin: https://github.com/sbt/sbt-typesafe-conductr but later, once it's fully completed. New issue?

akka-docs/rst/scala/microkernel.rst
```diff
@@ -2,15 +2,15 @@
...
 The purpose of the Akka Microkernel is to offer a bundling mechanism so that you can distribute
 an Akka application as a single payload, without the need to run in a Java Application Server or manually
 having to create a launcher script.
 
-The Akka Microkernel is included in the Akka download found at `downloads`_.
-
-.. _downloads: http://akka.io/downloads
+.. warning:: Akka Microkernel will be deprecated and removed. It will be replaced by using an ordinary
+   user defined main class and packaging with `sbt-native-packager <https://github.com/sbt/sbt-native-packager>`_
+   or `Typesafe ConductR <http://typesafe.com/products/conductr>`_.
```
ktoso: :+1:

drewhk:
LGTM. We might want to extend the docs on packaging in the future and provide a fully fledged native-packager template, too.

akka-samples/akka-sample-hello-kernel/build.sbt
```diff
@@ -1,12 +1,28 @@
-import NativePackagerKeys._
```
muuki88: Nice!

akka-samples/akka-sample-hello-kernel/build.sbt
```diff
@@ -1,12 +1,28 @@
...
+  "ch.qos.logback" % "logback-classic" % "1.0.7"
 )
+
+mainClass in Compile := Some("akka.kernel.Main")
+
+enablePlugins(JavaServerAppPackaging)
+
+mappings in Universal ++= {
+  // copy additional scripts from bin directory
+  directory("bin") ++
```
muuki88: That shouldn't be necessary.

akka-samples/akka-sample-hello-kernel/build.sbt
```diff
@@ -1,12 +1,28 @@
...
+  // copy configuration files
+  contentOf("src/main/resources").toMap.mapValues("config/" + _)
+}
+
+// add 'config' directory first in the classpath of the start script
```
muuki88: This is interesting. I would rather recommended setting the config file locations via CLI parameters. This won't play nice with the `LauncherPackaging` or `ClasspathJar`

muuki88: That's not true. It should work, but may look awkward

patriknw: Note that this is instructions/sample of how to build something that is close to what the Akka dist plugin did for Akka Microkernel. I agree that this packaging is not the best thing we can do, but that is for another pull request targeted to master when Akka Microkernel is deprecated for removal.

akka-samples/akka-sample-hello-kernel/build.sbt
patriknw: not necessary, but an example of how you can copy things if needed.
Not uncommon to have additional scripts, I think.

muuki88: Yep, but the `src/universal/bin` directory is always included. So this example maybe a bit misleading. We had this problem in our docs as well, when describing the app archetype as people just copy and pasted everything and created a mess :(

Maybe change it to `data-formats` or something else. Sorry for stressing this line so much, but in my experience users tend to use the hard way if they spot it :p

akka-samples/akka-sample-hello-kernel/build.sbt
muuki88: Good point. Transition should be as painless as possible.

akka-samples/akka-sample-hello-kernel/build.sbt
patriknw: ok, I'll change, thanks

akka-samples/akka-sample-hello-kernel/build.sbt
patriknw: thanks for reviewing

muuki88: you are welcome :) Good docs mean less issues on sbt-native-packager ;)

hepin1989:
it would be better to add something about mapping in the packaging and how to add external folder to the classpath.

akka-samples/akka-sample-hello-kernel/build.sbt
patriknw: I would like to hear your opinion about https://github.com/sbt/sbt-native-packager/issues/521. Let's discuss there.

Issue #15170, opened 2014-05-12
Make ByteString Serializable


bantonsson:
The ByteString class itself isn't serializable, only a subcategory called CompactByteString (things referencing a contiguous chunk of memory) is serializable. Why this difference was introduced isn't clear from the comments or the pull request #402.

There is also the question of what it exactly means to serialize a ByteString. Do you expect it to become one big contiguous array automatically?

colinrgodsey:
submitted PR #17070

colinrgodsey:
@bantonsson to answer your question about compactness: I implemented the serialization to produce a compact bytestring, but of the originating type still. Given that we have to serialize to an array anyways, it doesnt make sense to hold on to the 'sparseness'

bantonsson:
@colinrgodsey I completely agree. That is the way that makes most sense.

PR #17070, opened 2015-03-24
ByteString serialization

colinrgodsey:
## Deferred to #17144

Addresses ByteString serialization (#15170) while keeping binary
compatibility with the previously serializable ByteString

Referenced in previous ByteString comments in #17028

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/util/ByteStringSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/util/ByteStringSpec.scala
@@ -4,12 +4,16 @@
 
 package akka.util
 
+import java.io.{ ByteArrayInputStream, ObjectInputStream, ObjectOutputStream, ByteArrayOutputStream }
+
 import org.scalatest.WordSpec
 import org.scalatest.Matchers
 import org.scalatest.prop.Checkers
 import org.scalacheck.Arbitrary
 import org.scalacheck.Arbitrary.arbitrary
 import org.scalacheck.Gen
+
+import org.apache.commons.codec.binary.Hex.{ encodeHex, decodeHex }
 
 import scala.collection.mutable.Builder
 
@@ -54,6 +58,23 @@ class ByteStringSpec extends WordSpec with Matchers with Checkers {
       from ← Gen.choose(0, xs.length)
       until ← Gen.choose(from, xs.length)
     } yield (xs, from, until)
+  }
+
+  def testSer(obj: AnyRef) = {
+    val os = new ByteArrayOutputStream
+    val bos = new ObjectOutputStream(os)
+    bos.writeObject(obj)
+    val arr = os.toByteArray
+    val is = new ObjectInputStream(new ByteArrayInputStream(arr))
+
+    is.readObject == obj
+  }
+
+  def hexFromSer(obj: AnyRef) = {
+    val os = new ByteArrayOutputStream
+    val bos = new ObjectOutputStream(os)
+    bos.writeObject(obj)
+    String valueOf encodeHex(os.toByteArray)
   }
 
   val arbitraryByteArray: Arbitrary[Array[Byte]] = Arbitrary { Gen.sized { n ⇒ Gen.containerOfN[Array, Byte](n, arbitrary[Byte]) } }
@@ -367,6 +388,22 @@ class ByteStringSpec extends WordSpec with Matchers with Checkers {
               array.toSeq
             })
           }
+        }
+      }
+    }
+
+    "serialize correctly" when {
+      "parsing regular ByteString1C as compat" in {
+        val oldSerd = "aced000573720021616b6b612e7574696c2e42797465537472696e672442797465537472696e67314336e9eed0afcfe4a40200015b000562797465737400025b427872001b616b6b612e7574696c2e436f6d7061637442797465537472696e67fa2925150f93468f0200007870757200025b42acf317f8060854e002000078700000000a74657374737472696e67"
+        val bs = ByteString("teststring", "UTF8")
+        val str = hexFromSer(bs)
+
+        require(oldSerd == str)
+      }
+
+      "given all types of ByteString" in {
+        check { bs: ByteString ⇒
+          testSer(bs)
         }
       }
     }
--- a/akka-actor/src/main/scala/akka/util/ByteString.scala
+++ b/akka-actor/src/main/scala/akka/util/ByteString.scala
@@ -4,7 +4,7 @@
 
 package akka.util
 
-import java.io.OutputStream
+import java.io.{ ObjectInputStream, ObjectOutputStream }
 import java.nio.{ ByteBuffer, ByteOrder }
 import java.lang.{ Iterable ⇒ JIterable }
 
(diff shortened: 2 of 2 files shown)


patriknw:
OK TO TEST

akka-actor/src/main/scala/akka/util/ByteString.scala
```diff
@@ -274,8 +286,44 @@ object ByteString {
...
     def decodeString(charset: String): String = compact.decodeString(charset)
+
+    private[akka] def writeToOutputStream(os: OutputStream): Unit =
+      bytestrings.foreach(_.writeToOutputStream(os))
+
+    protected def writeReplace(): AnyRef = new SerializationProxy(this)
   }
 
+  @SerialVersionUID(1L)
+  private class SerializationProxy(@transient private var orig: ByteString) extends Serializable {
```
bantonsson: I assume that the `SerializationProxy` is a necessity by not being able to assign the `val` fields in any nice way.

bantonsson: I mean otherwise you could have done all the work by having `writeObject` and `readObject` in their respective classes.

akka-actor-tests/src/test/scala/akka/util/ByteStringSpec.scala
```diff
@@ -11,6 +13,8 @@ import org.scalacheck.Arbitrary
 import org.scalacheck.Arbitrary.arbitrary
 import org.scalacheck.Gen
 
+import org.apache.commons.codec.binary.Hex.{ encodeHex, decodeHex }
```
drewhk: just noticed. We have apache commons? what needs it?

akka-actor/src/main/scala/akka/util/ByteString.scala
```diff
@@ -274,8 +286,44 @@ object ByteString {
     def asByteBuffers: scala.collection.immutable.Iterable[ByteBuffer] = bytestrings map { _.asByteBuffer }
 
     def decodeString(charset: String): String = compact.decodeString(charset)
+
+    private[akka] def writeToOutputStream(os: OutputStream): Unit =
+      bytestrings.foreach(_.writeToOutputStream(os))
```
bantonsson: Thinking a bit more about this, there could be a problem with the aggregated `ByteString` becoming very large. I mean 256 bytestrings of 1M each is fine but a 256M one might not be. Maybe there should be some kind of cap on how much to coalesce?

akka-actor-tests/src/test/scala/akka/util/ByteStringSpec.scala
bantonsson: The `SerializeSpec`

bantonsson:
@colinrgodsey thanks a lot for the contribution

akka-actor/src/main/scala/akka/util/ByteString.scala
```diff
@@ -274,8 +286,44 @@ object ByteString {
...
+      val s = in.readInt()
+
+      val arr = new Array[Byte](s)
+      val bs = ByteString1C(arr).toByteString1
+      in.read(arr, 0, s)
+
+      orig = serType match {
+        case ByteString1.SerializationIdentity  ⇒ ByteString1(arr)
+        case ByteString1C.SerializationIdentity ⇒ ByteString1C(arr)
+        case ByteStrings.SerializationIdentity  ⇒ ByteStrings(Vector(ByteString1(arr)))
```
patriknw: perhaps `ByteStrings(Vector(ByteString1(arr)), arr.length)` ?

akka-actor/src/main/scala/akka/util/ByteString.scala
```diff
@@ -132,12 +136,14 @@ object ByteString {
...
       if (length == 0) empty else new ByteString1(bytes, startIndex, length)
+
+    private[akka] val SerializationIdentity = 0.toByte
   }
 
   /**
    * An unfragmented ByteString.
    */
-  final class ByteString1 private (private val bytes: Array[Byte], private val startIndex: Int, val length: Int) extends ByteString {
+  final class ByteString1 private (private val bytes: Array[Byte], private val startIndex: Int, val length: Int) extends ByteString with Serializable {
```
patriknw: SerialVersionUID?

patriknw:
LGTM

akka-actor/src/main/scala/akka/util/ByteString.scala
bantonsson: They should never show up on the wire. Only the `SerializationProxy` gets serialized.

akka-actor/src/main/scala/akka/util/ByteString.scala
drewhk: Can't we avoid this proxy somehow? I am yet to see why this is needed.

akka-actor/src/main/scala/akka/util/ByteString.scala
colinrgodsey: I see that there are possibly some performance differences related to allocation with a ByteString being serialized vs not (slicing over time, etc).... thats a good question. That also relates back into other 'sparseness' concerns in the existing ByteString. For example large IndexedSeqs are copied as flat Arrays via the tmp buffer, when it may be advantageous to spare out all incoming data via the ByteStringBuilder for the same reasons (I believe the Protobuf ByteString does this also).

akka-actor/src/main/scala/akka/util/ByteString.scala
colinrgodsey: yea, AFAIK Serializable is just needed for 'writeReplace' support

akka-actor/src/main/scala/akka/util/ByteString.scala
colinrgodsey: I guess mostly to avoid compatibility issues by forcing all serialization through a single proxy, and this way we can also avoid using vars in the underlying ByteString classes (*if* we wanted to serialize in a different form than as-constructed)

akka-actor/src/main/scala/akka/util/ByteString.scala
colinrgodsey: maybe we can leave ByteStrings sparse, but compact all ByteString1 (and components of ByteStrings) to ByteString1C

akka-actor/src/main/scala/akka/util/ByteString.scala
patriknw: good, thanks

akka-actor/src/main/scala/akka/util/ByteString.scala
bantonsson: @colinrgodsey Yes, I think it is a good compromise to keep the `Bytestrings` sparse but compact the others.

colinrgodsey:
all- ive pushed some improvements hopefully tackling the suggestions in this PR. 

I moved the serialization to a Companion trait on the ByteString types- this simplifies the proxy class a bit and should hopefully prove to be more extensible

akka-actor/src/main/scala/akka/util/ByteString.scala
```diff
@@ -274,8 +322,43 @@ object ByteString {
...
+      orig = Companion(from = serializationId).readFromInputStream(in)
+    }
+
+    private def readResolve(): AnyRef = orig
+  }
+
+  private[akka] object Companion {
+    def apply(from: Byte): Companion = companionMap(from)
+
+    val companionMap = Seq(ByteString1, ByteString1C, ByteStrings).
```
ktoso: `private val`

ktoso:
LGTM from my side, could you please squash these commits into one, with a commit message as explained in [contributing.md](https://github.com/akka/akka/blob/master/CONTRIBUTING.md)? Thanks!

colinrgodsey:
changed that val to private and squashed to a new PR

Deferred to #17144
-------------------------

PR #17073, opened 2015-03-25
+tes #17023 Adding ForwardActor among test-kit actors

### LINKED ISSUE #17023: Create a "standard" forward actor available for tests
`TestActors.EchoActor` is great, but not only one which is very usefull in test. 

Forward actor is very usefull also. It would be nice if testkit would provide implementation of forward actor.

mehmeteking:
Resolves #17023 by adding ForwardActor, a dummy which forwards each message to target specified at its constructor, to TestActors in test-kit.

### DIFF
--- a/akka-testkit/src/main/scala/akka/testkit/TestActors.scala
+++ b/akka-testkit/src/main/scala/akka/testkit/TestActors.scala
@@ -3,7 +3,7 @@
  */
 package akka.testkit
 
-import akka.actor.{ Props, Actor }
+import akka.actor.{ Props, Actor, ActorRef }
 
 /**
  * A collection of common actor patterns used in tests.
@@ -19,6 +19,18 @@ object TestActors {
     }
   }
 
+  /**
+   * ForwardActor forwards all messages as-is to specified ActorRef.
+   *
+   * @param ref target ActorRef to forward messages to
+   */
+  class ForwardActor(ref: ActorRef) extends Actor {
+    override def receive = {
+      case message ⇒ ref forward message
+    }
+  }
+
   val echoActorProps = Props[EchoActor]()
+  def forwardActorProps(ref: ActorRef) = Props(classOf[ForwardActor], ref)
 
 }
--- a/akka-testkit/src/test/scala/akka/testkit/TestActorsSpec.scala
+++ b/akka-testkit/src/test/scala/akka/testkit/TestActorsSpec.scala
@@ -5,7 +5,7 @@ package akka.testkit
 
 class TestActorsSpec extends AkkaSpec with ImplicitSender {
 
-  import TestActors.echoActorProps
+  import TestActors.{ echoActorProps, forwardActorProps }
 
   "A EchoActor" must {
     "send back messages unchanged" in {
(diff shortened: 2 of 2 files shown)


akka-testkit/src/main/scala/akka/testkit/TestActors.scala
```diff
@@ -19,6 +19,18 @@ object TestActors {
...
+   * @param ref target ActorRef to forward messages to
+   */
+  class ForwardActor(ref: ActorRef) extends Actor {
+    override def receive = {
+      case message ⇒ ref forward message
+    }
+  }
+
   val echoActorProps = Props[EchoActor]()
+  val forwardActorProps: (ActorRef ⇒ Props) = Props(classOf[ForwardActor], _)
```
drewhk: why not a simple def?

mehmeteking: I like that val is only evaluated once, also the EchoActor props is val and I have OCD :) Should we prefer def over val when both are applicable?

ktoso: You don't really save evaluations by using the val - the function it holds must be evaluated each time anyway :-) Memory wise a val needs to stay around where as you just call into a method.

I'm with Endre on this one, make it a `def` please.

ktoso:
LGTM except the comment above.

ktoso:
OK TO TEST

akka-testkit/src/main/scala/akka/testkit/TestActors.scala
drewhk: Also, the val version will allocate a Function1, while the def version will not, unless it is used in a closure. Not that it matters, but we were talking about OCD :)

mehmeteking: I made the change.

By "evaluated once", I meant the evaluation that generates Function1, not the Props; i.e. there is a single Function1 with val, but a new Function1 every time an equivalent def is called. Somehow the former was appealing to me, to a healthy person it shouldn't make any difference :)

ktoso:
LGTM, thanks for the contrib!

akka-testkit/src/main/scala/akka/testkit/TestActors.scala
drewhk: > but a new Function1 every time an equivalent def is called.

What do you mean? There will be no Function1 allocated at all. (There will be one allocated when the def is used in a closure, but only at that site, and that is not a common thing to do with props factories).

drewhk:
LGTM

bantonsson:
LGTM

akka-testkit/src/main/scala/akka/testkit/TestActors.scala
```diff
@@ -19,6 +19,18 @@ object TestActors {
...
+   * @param ref target ActorRef to forward messages to
+   */
+  class ForwardActor(ref: ActorRef) extends Actor {
+    override def receive = {
+      case message ⇒ ref forward message
+    }
+  }
+
   val echoActorProps = Props[EchoActor]()
+  def forwardActorProps(ref: ActorRef) = Props(classOf[ForwardActor], ref)
```
patriknw: here it should be safe to use `Props(new ForwardActor(ref))`

patriknw:
LGTM

Issue #16910, opened 2015-02-19
doc: Code inclusion in Fault Tolerance section is wrong


patriknw:
shows things like `with FlatSpecLike with BeforeAndAfterAll`

http://doc.akka.io/docs/akka/2.3.9/scala/fault-tolerance.html

http://doc.akka.io/docs/akka/snapshot/scala/fault-tolerance.html

PR #17076, opened 2015-03-25
=doc #16910 Fix include sections in fault tolerance docs

patriknw:
- and don't show AkkaSpec in doc code


patriknw:
Refs #16910

Issue #16295, opened 2014-11-12
Performance degraded when upgrading From Akka 2.3.6 to Akka 2.3.7


adamhonen:
I’ve upgraded to 2.3.7 and tried running a very basic performance test I do for the sake of sanity from time to time.
It’s something that I use to find big glaring performance bugs – I have my application run using a local database (Couchbase) on my own PC, as well as a small client simulator (also written using Akka) and just check how many operations a second it’s doing. So while this setup cannot be used to do any serious tuning, it is useful for spotting stuff like forgetting the logs in DEBUG.

I’ve tried running with 2.3.6 and 2.3.7 several times (going back and forth) and as basic as this test is, the results are consistent.
I get ~11K ops/sec with 2.3.6 and ~7.3K with 2.3.7.
The version of Akka is of course the only difference between each of the runs.

Looking at the thread view of visualvm I see one immediate difference between the two versions.
With Akka 2.3.6, the dispatcher threads seem to be spending less time in park state.
Of course, I’m not familiar at all with the internal Akka implementation, so maybe that doesn’t mean anything at all – it just looks to me as a useful hint.

Let me know what additional information I can supply in order to investigate this.

![dispatcher-threads](https://cloud.githubusercontent.com/assets/6965115/5018015/526f7cce-6abb-11e4-9f52-e013f5effee1.JPG)

t3hnar:
I can see the same issue, reported in https://github.com/EventStore/EventStore.JVM/issues/30

drewhk:
Superceded by new ticket: https://github.com/akka/akka/issues/17122

PR #17079, opened 2015-03-26
=act #16295 Temporary disable windows-connection-abort-workaround

patriknw:
Until we have a working solution.

Are there any tests that will fail on windows because of this change? Then those tests could be configured with `auto`.

### DIFF
--- a/akka-actor-tests/src/test/resources/reference.conf
+++ b/akka-actor-tests/src/test/resources/reference.conf
@@ -3,3 +3,7 @@ akka {
     serialize-messages = on    
   }
 }
+
+# FIXME Some test depend on this setting when running on windows.
+#       It should be removed when #17122 is solved. 
+akka.io.tcp.windows-connection-abort-workaround-enabled = auto
--- a/akka-actor/src/main/resources/reference.conf
+++ b/akka-actor/src/main/resources/reference.conf
@@ -600,7 +600,7 @@ akka {
       # and undefined behavior.
       # Possible values of this key are on, off and auto where auto will enable the
       # workaround if Windows is detected automatically.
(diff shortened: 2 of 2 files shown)


drewhk:
I think there are a few tests that will fail.

patriknw:
how do I find those tests?

drewhk:
I think the only reliable way is to run the tests on a Windows machine

patriknw:
could you help me with that, please

patriknw:
@drewhk it would be great if you could find the failing tests, so that we can include this in upcoming 2.3.10 release.

drewhk:
Ok, doing it now

drewhk:
Lots of tests break for me, and one of them does not even terminate. Interesting.

drewhk:
I don't remember having so much failing tests...

patriknw:
ok, but they work when you set it to `auto`? then that is what we need to do

drewhk:
Yes, they work with auto.

rkuhn:
I’d much rather see a “probing back-off” strategy implemented (i.e. instead of setting OP_CONNECT right away only do it once every 30 seconds or so) than this temporary regression. But if we cannot get the proper solution done then we’ll at least need to mark those tests that need it with `auto`.

drewhk:
I think periodic ResumeReading would work too, but the proper solution is probably the read-ahead solution discussed in the master ticket: https://github.com/akka/akka/issues/17122

patriknw:
I have added the setting to /akka-actor-tests/src/test/resources/reference.conf

patriknw:
Refs #17255

rkuhn:
LGTM

Issue #16924, opened 2015-02-24
Investigate commented out test case in FlowSpec


drewhk:

bantonsson:
What are we talking about here? The FIXME comment that is as old as the file itself (12:th Sep 2014)? I can't see anything else that is commented out.

drewhk:
that is the part, yes, and it is old, yes :)

drewhk:
"We get a processor actor terminated abruptly" failure to the 3rd subscriber. It should be "processor actor shut-down" at that point.

PR #17085, opened 2015-03-27
=str #16924: Fix FanoutProcessor to not overwrite last termination cause

drewhk:
Also fix FlowSpec to expect onError instead of onComplete for late subscribers

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSpec.scala
@@ -457,7 +457,7 @@ class FlowSpec extends AkkaSpec(ConfigFactory.parseString("akka.actor.debug.rece
       }
     }
 
-    "call future subscribers' onComplete instead of onSubscribed after initial upstream was completed" in {
+    "call future subscribers' onError after onSubscribe if initial upstream was completed" in {
       new ChainSetup(identity, settings.copy(initialInputBufferSize = 1),
         toFanoutPublisher(initialBufferSize = 1, maximumBufferSize = 1)) {
         val downstream2 = StreamTestKit.SubscriberProbe[Any]()
@@ -489,11 +489,10 @@ class FlowSpec extends AkkaSpec(ConfigFactory.parseString("akka.actor.debug.rece
         downstream2.expectNext("a3")
         downstream2.expectComplete()
 
-        // FIXME when adding a sleep before the following link this will fail with IllegalStateExc shut-down
-        // what is the expected shutdown behavior? Is the title of this test wrong?
-        //        val downstream3 = StreamTestKit.SubscriberProbe[Any]()
-        //        publisher.subscribe(downstream3)
-        //        downstream3.expectComplete()
+        val downstream3 = StreamTestKit.SubscriberProbe[Any]()
+        publisher.subscribe(downstream3)
+        downstream3.expectSubscription()
+        downstream3.expectError() should ===(ActorPublisher.NormalShutdownReason)
       }
     }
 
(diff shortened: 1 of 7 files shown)


akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSpec.scala
```diff
@@ -489,11 +489,10 @@ class FlowSpec extends AkkaSpec(ConfigFactory.parseString("akka.actor.debug.rece
...
 
-        // FIXME when adding a sleep before the following link this will fail with IllegalStateExc shut-down
-        // what is the expected shutdown behavior? Is the title of this test wrong?
-        //        val downstream3 = StreamTestKit.SubscriberProbe[Any]()
-        //        publisher.subscribe(downstream3)
-        //        downstream3.expectComplete()
+        val downstream3 = StreamTestKit.SubscriberProbe[Any]()
+        publisher.subscribe(downstream3)
+        downstream3.expectSubscription()
+        downstream3.expectError().getMessage should ===("Cannot subscribe to shut-down Publisher")
```
patriknw: This message is defined in ActorPublisher.NormalShutdownException. Please extract it to a val and compare with that here. It is also written in StreamOfStreamProcessors.scala and SubscriberManagement.scala. Looks like they could use the ActorPublisher.NormalShutdownException instead.

patriknw:
LGTM, apart from the exception message cleanup

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSpec.scala
drewhk: Ok, thanks!

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSpec.scala
```diff
@@ -457,7 +457,7 @@ class FlowSpec extends AkkaSpec(ConfigFactory.parseString("akka.actor.debug.rece
       }
     }
 
-    "call future subscribers' onComplete instead of onSubscribed after initial upstream was completed" in {
+    "call future subscribers' onError after of onSubscribe if initial upstream was completed" in {
```
2m: Delete of.

akka-stream/src/main/scala/akka/stream/impl/ActorPublisher.scala
```diff
@@ -16,8 +16,9 @@ import org.reactivestreams.Subscription
  * INTERNAL API
  */
 private[akka] object ActorPublisher {
-  class NormalShutdownException extends IllegalStateException("Cannot subscribe to shut-down Publisher") with NoStackTrace
-  val NormalShutdownReason: Option[Throwable] = Some(new NormalShutdownException)
+  val NormalShutdownReasonMessage = "Cannot subscribe to shut-down Publisher"
+  class NormalShutdownException extends IllegalStateException(NormalShutdownReasonMessage) with NoStackTrace
+  val NormalShutdownReason: Throwable = new NormalShutdownException
```
rkuhn: The previous value was optimized for not allocating a `Some` during each shutdown, perhaps we can retain this as `val SomeNormalShutdownReason`?

rkuhn:
Other than the deoptimization it LGTM.

2m:
LGTM apart of a minor type in the test case name.

akka-stream/src/main/scala/akka/stream/impl/ActorPublisher.scala
drewhk: Does it matter that much? I expect this not being a hot path. Anyway, the fix is simple.

ktoso:
resolves #16924

PR #17091, opened 2015-03-30
=htp #17082 fix response message generated for `MethodRejection`

### LINKED ISSUE #17082: 405 response message incorrect
Using akka http routing, when you make a request with an unsupported message to an endpoint the error message reads:

HTTP method not allowed, supported methods: HttpMethod(GET)

Whereas it should just say:
HTTP method not allowed, supported methods: GET

Looks like an easy one - in akka-http/src/main/scala/akka/http/server/RejectionHandler.scala, it does:

"HTTP method not allowed, supported methods: " + methods.mkString(", ")

so I guess it needs a ".map(_.name)"

sirthias:
Fixes #17082.

/cc @jrudolph


jrudolph:
Nice. LGTM

ktoso:
LGTM

Issue #16641, opened 2015-01-12
Config setting `remote-address-header` is parsed but never used


luksow:
`remote-address-header` in `akka.http.server` namespace is parsed but then never actually used. This renders `extractClientIP` directive useless unless you put a proxy in front of your server to append proper headers.

luksow:
My initial guess is that it should append header somewhere in `HttpRequestParser` or `HttpServer` but I'm not that familiar with akka-http code (yet!). With little guidance I can PR a fix.

rkuhn:
Thanks for reporting and for offering to try your hand on this. While I cannot immediately suggest where to start on this one (@sirthias?) we’ll definitely welcome and review a PR.

sirthias:
This is an interesting one, demonstrating a fundamental problem that we might not yet have solved properly, namely https://github.com/akka/akka/issues/16168.
What we want is have the 

```scala
def serverFlowToTransport(serverFlow: Flow[HttpRequest, HttpResponse],
  settings: ServerSettings, log: LoggingAdapter): Flow[ByteString, ByteString]
```

method (optionally) add a special request header holding the clients IP address.
Since the flow produced by `serverFlowToTransport` is supposed to be reuseable we need to somehow access the `MaterializedMap` from within the flow (after all the clients IP differs per materialization).

/cc @jrudolph

drewhk:
@jrudolph @sirthias this particular problem can be now solved with the materializedValue accessor in graphs, right? Is some other feature missing?

jrudolph:
For now we will implement it on top of the current one-time flows which should be very simple.

Making the server flow reusable is still a major task that involves a lot of restructuring that is currently not in scope as we have a working solution by reconstructing the flow per connection. Also, conceptually it would mean that the `BidiFlow` would need to be able to take per-materialization parameters (we talked about this somewhere else) which could be done by the `serverFlow` providing a hole as materialized value and combining that hole at (after?) materialization. In any case this seems rather like a workaround like a real solution.

jrudolph:
Fixed in #17658.

Issue #16168, opened 2014-10-29
Allow access to the MaterializedMap inside the Flow


jrudolph:
This is needed e.g. for an IO server API like the one sketched in #16162 and #16166 where it is hard to access per-materialization data inside the flow. One use case is that in akka-http we need to provide the remote address to the user.

To show this in a simpler example, an echo server could be implemented like this:

```scala
val echoFlow = Flow[ByteString].map(ByteString("Hello ") ++ _)
Tcp.generalBind(addr)(conn => conn.handleWith(echoFlow))
```

Now, you may want to access the remote address inside the Flow which handles a request:

```scala
val echoFlow = Flow[ByteString].map(ByteString("Hello ") ++ _ ++ ByteString(" at address ???"))
Tcp.generalBind(addr) { conn => 
  // there's conn.remoteAddress(MaterializedMap) but we don't have a MaterializedMap
  conn.handleWith(echoFlow)
}
```

One idea would be to provide an overload for the `transform` combinator which passes the `MaterializedMap` when the transformer gets instantiated:

```scala
def transform[T](mkTransformer: MaterializedMap ⇒ Transformer[Out, T]): Repr[T]
```

Then, the above (or something similar) could be implemented as:

```scala
val echoFlow = Flow[ByteString].map(ByteString("Hello ") ++ _ ++ ByteString("at address ???"))
Tcp.generalBind(addr) { conn => 
  conn.handleWith {
    Flow[ByteString].transform { matMap =>
      new Transformer[ByteString, ByteString] {
         def onNext(b: ByteString): Seq[ByteString] = Seq(b, ByteString(conn.remoteAddress(map).toString))
      }
    }
  }
}
```

/cc @sirthias

rkuhn:
A proposal for how this could work can be found [here](http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0-M2/stream-design.html#The_finer_points_of_stream_materialization). We’ll need to figure this out before 1.0.

rkuhn:
This will likely be solved by offering the materialized value of a graph as a Source within that graph’s FlowGraph.Builder.

drewhk:
@sirthias @jrudolph there is a PR for this: https://github.com/akka/akka/pull/17097

You can look at the test cases to see how it works: https://github.com/akka/akka/pull/17097/files#diff-f1e18fd8ee3a5d3436daae570bffd03d

Issue #16503, opened 2014-12-09
Create a lightweight driver for FutureSource


viktorklang:
No need to create an actor for it.

Or create an AsyncIterablePublisher that takes an immutableIterable[Future[T]] (where FutureSource would be `new immutable.Iterable { def iterator = Iterator.single(future) }`

rkuhn:
related to #15135

PR #17097, opened 2015-03-30
!str #16168: Expose materialized value in the graph as a "source"

drewhk:
Only docs are missing now, and probably fixing samples.

### DIFF
--- a/akka-docs-dev/rst/java/stream-graphs.rst
+++ b/akka-docs-dev/rst/java/stream-graphs.rst
@@ -194,6 +194,24 @@ together and also turned around with the ``.reversed()`` method. The test
 simulates both parties of a network communication protocol without actually
 having to open a network connection—the flows can just be connected directly.
 
+.. _graph-matvalue-java:
+
+Accessing the materialized value inside the Graph
+-------------------------------------------------
+
+In certain cases it might be necessary to feed back the materialized value of a Graph (partial, closed or backing a
+Source, Sink, Flow or BidiFlow). This is possible by using ``builder.matValue`` which gives an ``Outlet`` that
+can be used in the graph as an ordinary source or outlet, and which will eventually emit the materialized value.
+If the materialized value is needed at more than one place, it is possible to call ``matValue`` any number of times
+to acquire the necessary number of outlets.
+
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/FlowGraphDocTest.java#flow-graph-matvalue
+
+Be careful not to introduce a cycle where the materialized value actually contributes to the materialized value.
+The following example demonstrates a case where the materialized ``Future`` of a fold is fed back to the fold itself.
+
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/FlowGraphDocTest.java#flow-graph-matvalue-cycle
+
 .. _graph-cycles-java:
 
 Graph cycles, liveness and deadlocks
--- a/akka-stream-tests/src/test/java/akka/stream/javadsl/FlexiRouteTest.java
+++ b/akka-stream-tests/src/test/java/akka/stream/javadsl/FlexiRouteTest.java
@@ -50,9 +50,9 @@ public class FlexiRouteTest {
             out1,
             out2,
             Keep.<Future<List<String>>, Future<List<String>>> both(),
-            new Procedure3<Builder, SinkShape<List<String>>, SinkShape<List<String>>>() {
+            new Procedure3<Builder<Pair<Future<List<String>>, Future<List<String>>>>, SinkShape<List<String>>, SinkShape<List<String>>>() {
               @Override
-              public void apply(Builder b, SinkShape<List<String>> o1,
+              public void apply(Builder<Pair<Future<List<String>>, Future<List<String>>>> b, SinkShape<List<String>> o1,
                   SinkShape<List<String>> o2) throws Exception {
                 final UniformFanOutShape<String, String> fair = b.graph(new Fair<String>());
                 b.edge(b.source(in), fair.in());
@@ -80,9 +80,9 @@ public class FlexiRouteTest {
             out1,
             out2,
             Keep.<Future<List<String>>, Future<List<String>>> both(),
-            new Procedure3<Builder, SinkShape<List<String>>, SinkShape<List<String>>>() {
+            new Procedure3<Builder<Pair<Future<List<String>>, Future<List<String>>>>, SinkShape<List<String>>, SinkShape<List<String>>>() {
               @Override
-              public void apply(Builder b, SinkShape<List<String>> o1,
+              public void apply(Builder<Pair<Future<List<String>>, Future<List<String>>>> b, SinkShape<List<String>> o1,
                   SinkShape<List<String>> o2) throws Exception {
                 final UniformFanOutShape<String, String> robin = b.graph(new StrictRoundRobin<String>());
                 b.edge(b.source(in), robin.in());
@@ -112,9 +112,9 @@ public class FlexiRouteTest {
             Sink.<List<Integer>> head(),
             out2,
             Keep.<Future<List<Integer>>, Future<List<String>>> both(),
-            new Procedure3<Builder, SinkShape<List<Integer>>, SinkShape<List<String>>>() {
+            new Procedure3<Builder<Pair<Future<List<Integer>>, Future<List<String>>> >, SinkShape<List<Integer>>, SinkShape<List<String>>>() {
               @Override
-              public void apply(Builder b, SinkShape<List<Integer>> o1,
+              public void apply(Builder<Pair<Future<List<Integer>>, Future<List<String>>> > b, SinkShape<List<Integer>> o1,
                   SinkShape<List<String>> o2) throws Exception {
                 final FanOutShape2<Pair<Integer, String>, Integer, String> unzip = b.graph(new Unzip<Integer, String>());
                 final Outlet<Pair<Integer, String>> src = b.source(Source.from(pairs));
--- a/akka-stream/src/main/boilerplate/akka/stream/javadsl/BidiFlowCreate.scala.template
+++ b/akka-stream/src/main/boilerplate/akka/stream/javadsl/BidiFlowCreate.scala.template
@@ -13,14 +13,14 @@ trait BidiFlowCreate {
   import language.implicitConversions
   private implicit def p[A, B](pair: Pair[A, B]): (A, B) = pair.first -> pair.second
 
-  def create[I1, O1, I2, O2](block: japi.Function[FlowGraph.Builder, BidiShape[I1, O1, I2, O2]]): BidiFlow[I1, O1, I2, O2, Unit] =
+  def create[I1, O1, I2, O2](block: japi.Function[FlowGraph.Builder[Unit], BidiShape[I1, O1, I2, O2]]): BidiFlow[I1, O1, I2, O2, Unit] =
     new BidiFlow(scaladsl.BidiFlow() { b ⇒ block.apply(b.asJava) })
 
-  def create[I1, O1, I2, O2, S <: Shape, M](g1: Graph[S, M], block: japi.Function2[FlowGraph.Builder, S, BidiShape[I1, O1, I2, O2]]): BidiFlow[I1, O1, I2, O2, M] =
+  def create[I1, O1, I2, O2, S <: Shape, M](g1: Graph[S, M], block: japi.Function2[FlowGraph.Builder[M], S, BidiShape[I1, O1, I2, O2]]): BidiFlow[I1, O1, I2, O2, M] =
     new BidiFlow(scaladsl.BidiFlow(g1) { b ⇒ s => block.apply(b.asJava, s) })
 
   [2..21#def create[I##1, O##1, I##2, O##2, [#S1 <: Shape#], [#M1#], M]([#g1: Graph[S1, M1]#], combineMat: japi.Function1[[#M1#], M],
-      block: japi.Function2[FlowGraph.Builder, [#S1#], BidiShape[I##1, O##1, I##2, O##2]]): BidiFlow[I##1, O##1, I##2, O##2, M] =
+      block: japi.Function2[FlowGraph.Builder[M], [#S1#], BidiShape[I##1, O##1, I##2, O##2]]): BidiFlow[I##1, O##1, I##2, O##2, M] =
     new BidiFlow(scaladsl.BidiFlow([#g1#])(combineMat.apply _) { b => ([#s1#]) => block.apply(b.asJava, [#s1#]) })#
   
   ]
--- a/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
@@ -3,11 +3,15 @@
  */
 package akka.stream.impl
 
+import java.util.concurrent.atomic.{ AtomicInteger, AtomicBoolean, AtomicReference }
+
+import akka.stream.impl.StreamLayout.Module
 import akka.stream.scaladsl.{ Keep, OperationAttributes }
 import akka.stream._
 import org.reactivestreams.{ Subscription, Publisher, Subscriber }
 import akka.event.Logging.simpleName
 import scala.collection.mutable
+import scala.util.control.NonFatal
 
 /**
  * INTERNAL API
@@ -313,6 +317,148 @@ private[stream] class VirtualPublisher[T]() extends Publisher[T] {
 /**
  * INTERNAL API
  */
+private[stream] case class MaterializedValueSource[M](
+  shape: SourceShape[M] = SourceShape[M](new Outlet[M]("Materialized.out")),
+  attributes: OperationAttributes = OperationAttributes.name("Materialized")) extends StreamLayout.Module {
+
+  override def subModules: Set[Module] = Set.empty
+  override def withAttributes(attr: OperationAttributes): Module = this.copy(shape = amendShape(attr), attributes = attr)
+  override def carbonCopy: Module = this.copy(shape = SourceShape(new Outlet[M]("Materialized.out")))
+
+  override def replaceShape(s: Shape): Module =
+    if (s == shape) this
+    else throw new UnsupportedOperationException("cannot replace the shape of MaterializedValueSource")
+
+  def amendShape(attr: OperationAttributes): SourceShape[M] = {
+    attr.nameOption match {
+      case None ⇒ shape
+      case s: Some[String] if s == attributes.nameOption ⇒ shape
+      case Some(name) ⇒ shape.copy(outlet = new Outlet(name + ".out"))
+    }
+  }
+
+}
+
+/**
+ * INTERNAL API
+ */
+private[stream] object MaterializedValuePublisher {
+  final val NotRequested = 0
+  final val Requested = 1
+  final val Completed = 2
+
+  final val NoValue = new AnyRef
+}
+
+/**
+ * INTERNAL API
+ */
+private[stream] class MaterializedValuePublisher extends Publisher[Any] {
+  import MaterializedValuePublisher._
+
+  private val value = new AtomicReference[AnyRef](NoValue)
+  private val registeredSubscriber = new AtomicReference[Subscriber[_ >: Any]](null)
+  private val requestState = new AtomicInteger(NotRequested)
+
+  private def close(): Unit = {
+    requestState.set(Completed)
+    value.set(NoValue)
+    registeredSubscriber.set(null)
+  }
+
+  private def tryOrClose(block: ⇒ Unit): Unit = {
+    try block catch {
+      case v: ReactiveStreamsCompliance.SpecViolation ⇒
+        close()
+      // What else can we do here?
+      case NonFatal(e) ⇒
+        val sub = registeredSubscriber.get()
+        if ((sub ne null) &&
+          requestState.compareAndSet(NotRequested, Completed) || requestState.compareAndSet(Requested, Completed)) {
+          sub.onError(e)
+        }
+        close()
+        throw e
+    }
+  }
+
+  def setValue(m: Any): Unit =
+    tryOrClose {
+      if (value.compareAndSet(NoValue, m.asInstanceOf[AnyRef]) && requestState.get() == Requested)
+        pushAndClose(m)
+    }
+
+  /*
+   * Both call-sites do a CAS on their "own" side and a GET on the other side. The possible overlaps
+   * are (removing symmetric cases where you can relabel A->B, B->A):
+   *
+   * A-CAS
+   * A-GET
+   * B-CAS
+   * B-GET - pushAndClose fires here
+   *
+   * A-CAS
+   * B-CAS
+   * A-GET - pushAndClose fires here
+   * B-GET - pushAndClose fires here
+   *
+   * A-CAS
+   * B-CAS
+   * B-GET - pushAndClose fires here
+   * A-GET - pushAndClose fires here
+   *
+   * The proof that there are no cases:
+   *
+   * - all permutations of 4 operations are 4! = 24
+   * - the operations of A and B are cannot be reordered, so there are 24 / (2 * 2) = 6 actual orderings
+   * - if we don't count cases which are a simple relabeling A->B, B->A, we get 6 / 2 = 3 reorderings
+   *   which are all enumerated above.
+   *
+   * pushAndClose protects against double onNext by doing a CAS itself.
+   */
+  private def pushAndClose(m: Any): Unit = {
+    if (requestState.compareAndSet(Requested, Completed)) {
+      val sub = registeredSubscriber.get()
+      ReactiveStreamsCompliance.tryOnNext(sub, m)
+      ReactiveStreamsCompliance.tryOnComplete(sub)
+      close()
+    }
+  }
+
+  override def subscribe(subscriber: Subscriber[_ >: Any]): Unit = {
+    tryOrClose {
+      ReactiveStreamsCompliance.requireNonNullSubscriber(subscriber)
+      if (registeredSubscriber.compareAndSet(null, subscriber)) {
+        ReactiveStreamsCompliance.tryOnSubscribe(subscriber, new Subscription {
+          override def cancel(): Unit = close()
+
+          override def request(n: Long): Unit = {
+            if (n <= 0) {
+              ReactiveStreamsCompliance.tryOnError(
+                subscriber,
+                ReactiveStreamsCompliance.numberOfElementsInRequestMustBePositiveException)
+            } else {
+              if (requestState.compareAndSet(NotRequested, Requested)) {
+                val m = value.get()
+                if (m ne NoValue) pushAndClose(m)
+              }
+            }
+          }
+        })
+      } else {
+        if (subscriber == registeredSubscriber.get())
+          ReactiveStreamsCompliance.rejectDuplicateSubscriber(subscriber)
+        else
+          ReactiveStreamsCompliance.rejectAdditionalSubscriber(subscriber, "MaterializedValuePublisher")
+      }
+    }
+  }
+
+}
+
+/**
+ * INTERNAL API
+ */
 private[stream] abstract class MaterializerSession(val topLevel: StreamLayout.Module) {
   import StreamLayout._
 
@@ -332,12 +478,25 @@ private[stream] abstract class MaterializerSession(val topLevel: StreamLayout.Mo
 
   protected def materializeModule(module: Module, effectiveAttributes: OperationAttributes): Any = {
     val materializedValues = collection.mutable.HashMap.empty[Module, Any]
+    var materializedValuePublishers: List[MaterializedValuePublisher] = Nil
+
     for (submodule ← module.subModules) {
       val subEffectiveAttributes = mergeAttributes(effectiveAttributes, submodule.attributes)
-      if (submodule.isAtomic) materializedValues.put(submodule, materializeAtomic(submodule, subEffectiveAttributes))
-      else materializedValues.put(submodule, materializeComposite(submodule, subEffectiveAttributes))
+      submodule match {
+        case mv: MaterializedValueSource[_] ⇒
+          val pub = new MaterializedValuePublisher
+          materializedValuePublishers ::= pub
+          assignPort(mv.shape.outlet, pub)
+        case atomic if atomic.isAtomic ⇒
+          materializedValues.put(atomic, materializeAtomic(atomic, subEffectiveAttributes))
+        case composite ⇒
+          materializedValues.put(composite, materializeComposite(composite, subEffectiveAttributes))
+      }
     }
-    resolveMaterialized(module.materializedValueComputation, materializedValues)
+
+    val mat = resolveMaterialized(module.materializedValueComputation, materializedValues)
+    materializedValuePublishers foreach { pub ⇒ pub.setValue(mat) }
+    mat
   }
 
   protected def materializeComposite(composite: Module, effectiveAttributes: OperationAttributes): Any = {
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Graph.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Graph.scala
@@ -4,7 +4,6 @@
 package akka.stream.javadsl
 
 import akka.stream._
-import akka.stream.scaladsl
 import akka.japi.Pair
 
 /**
@@ -278,10 +277,10 @@ object FlowGraph {
    * The [[Builder]] is mutable and not thread-safe,
    * thus you should construct your Graph and then share the constructed immutable [[FlowGraph]].
    */
-  def builder(): Builder = new Builder()(new scaladsl.FlowGraph.Builder)
+  def builder[M](): Builder[M] = new Builder()(new scaladsl.FlowGraph.Builder[M])
 
-  class Builder()(private implicit val delegate: scaladsl.FlowGraph.Builder) { self ⇒
-    import scaladsl.FlowGraph.Implicits._
+  final class Builder[Mat]()(private implicit val delegate: scaladsl.FlowGraph.Builder[Mat]) { self ⇒
+    import akka.stream.scaladsl.FlowGraph.Implicits._
 
     def flow[A, B, M](from: Outlet[A], via: Flow[A, B, M], to: Inlet[B]): Unit = delegate.addEdge(from, via.asScala, to)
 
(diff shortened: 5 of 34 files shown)


akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -313,6 +316,61 @@ private[stream] class VirtualPublisher[T]() extends Publisher[T] {
...
+      case Some(name) ⇒ shape.copy(outlet = new Outlet(name + ".out"))
+    }
+  }
+
+}
+
+/**
+ * INTERNAL API
+ */
+private[stream] class MaterializedValuePublisher extends Publisher[Any] {
```
rkuhn: This should be extended slightly to cover all that is needed for a PromisePublisher and then also used for FutureSource.

rkuhn:
Looking very promising: seems that this is easier than I thought!

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -313,6 +316,61 @@ private[stream] class VirtualPublisher[T]() extends Publisher[T] {
...
+
+  def setValue(m: Any): Unit = {
+    if (value.compareAndSet(null, m.asInstanceOf[AnyRef]) && hasRequest.get()) {
+      subscriber.get().onNext(m)
+    }
+  }
+
+  override def subscribe(s: Subscriber[_ >: Any]): Unit = {
+    if (subscriber.compareAndSet(null, s)) {
+      s.onSubscribe(new Subscription {
```
patriknw: use `ReactiveStreamsCompliance`

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -313,6 +316,61 @@ private[stream] class VirtualPublisher[T]() extends Publisher[T] {
...
+        override def cancel(): Unit = hasRequest.set(false)
+        override def request(n: Long): Unit = {
+          // TODO: Signal invalid request count
+          hasRequest.set(true)
+          if (value.get() ne null) s.onNext(value.get())
```
patriknw: two value.get()?

patriknw: is it supposed to emit the value multiple times for multiple `request`?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: The reason I have it here in this standalone form is to avoid depending on higher layer features in this core feature.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: yes, there are TODOs around for doing all that jazz.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: What I meant is that this is _nearly_ good enough to replace the needlessly Actor-based FutureSource, so just keep it here and we’ll use it for other things as well. The only needed extension is that it would need to support failures (i.e. `Try`).

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: you are right, I forgot the onComplete call. Meh.

drewhk:
Unfortunately adding the additional type parameter to Builder is not completely trivial with Java

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -313,6 +316,88 @@ private[stream] class VirtualPublisher[T]() extends Publisher[T] {
...
+
+  def setValue(m: Any): Unit =
+    if (value.compareAndSet(null, m.asInstanceOf[AnyRef]) && requestState.get() == Requested)
+      pushAndClose(m)
+
+  private def pushAndClose(m: Any): Unit = {
+    val sub = registeredSubscriber.get()
+    ReactiveStreamsCompliance.tryOnNext(sub, m)
+    ReactiveStreamsCompliance.tryOnComplete(sub)
+    requestState.set(Completed)
```
drewhk: this needs to be a compareAndSet and should guard the onNext/onComplete calls to avoid duplicate signaling. Not that it can happen right now, since materialization is synchronous (the whole atomic stuff is strictly unnecessary), but hey.

akka-stream/src/main/boilerplate/akka/stream/javadsl/BidiFlowCreate.scala.template
```diff
@@ -13,14 +13,14 @@ trait BidiFlowCreate {
   import language.implicitConversions
   private implicit def p[A, B](pair: Pair[A, B]): (A, B) = pair.first -> pair.second
 
-  def create[I1, O1, I2, O2](block: japi.Function[FlowGraph.Builder, BidiShape[I1, O1, I2, O2]]): BidiFlow[I1, O1, I2, O2, Unit] =
+  def create[I1, O1, I2, O2](block: japi.Function[FlowGraph.Builder[Unit], BidiShape[I1, O1, I2, O2]]): BidiFlow[I1, O1, I2, O2, Unit] =
```
ktoso: Would `_` be less painful for Java users as it desugars to `Object`?

drewhk: But it is not Object but Unit, and it must match the materialized type.

rkuhn: This is an input that we give to the user, so we need to give the proper type (which is Unit). We should be flexible in places where we accept input from the user, where `_` may be an option for Java (needs verification). We must not ever return such a wildcard type to users.

ktoso: Oh okey, I think I get it now.

drewhk:
Ready for final review

akka-stream-tests/src/test/java/akka/stream/javadsl/FlexiRouteTest.java
```diff
@@ -50,9 +50,9 @@ public void mustBuildSimpleFairRoute() throws Exception {
             out1,
             out2,
             Keep.<Future<List<String>>, Future<List<String>>> both(),
-            new Procedure3<Builder, SinkShape<List<String>>, SinkShape<List<String>>>() {
+            new Procedure3<Builder<Pair<Future<List<String>>, Future<List<String>>>>, SinkShape<List<String>>, SinkShape<List<String>>>() {
```
patriknw: :-) >>>>>>>>>>>>>>>>>>>>>>>>

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -313,6 +316,89 @@ private[stream] class VirtualPublisher[T]() extends Publisher[T] {
...
+  private val requestState = new AtomicInteger(NotRequested)
+
+  def setValue(m: Any): Unit =
+    if (value.compareAndSet(null, m.asInstanceOf[AnyRef]) && requestState.get() == Requested)
+      pushAndClose(m)
+
+  private def pushAndClose(m: Any): Unit = {
+    if (requestState.compareAndSet(Requested, Completed)) {
+      val sub = registeredSubscriber.get()
+      ReactiveStreamsCompliance.tryOnNext(sub, m)
```
patriknw: how/where is SpecViolation handled?

drewhk: I thought these calls do what is necessary. What can I do on SpecViolation exceptions?

akka-stream-tests/src/test/java/akka/stream/javadsl/FlexiRouteTest.java
drewhk: <+ <* a 2> <- b 5> <+ <f c d> e>>

drewhk: ah no, even Lisp looks more horrible that way

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -313,6 +316,89 @@ private[stream] class VirtualPublisher[T]() extends Publisher[T] {
...
+
+}
+
+/**
+ * INTERNAL API
+ */
+private[stream] object MaterializedValuePublisher {
+  val NotRequested = 0
+  val Requested = 1
+  val Completed = 2
```
patriknw: sealed trait and case objects?
I guess these will be boxed when placed in the AtomicReference.

drewhk: that is an AtomicInteger. I hope that does not wrap.

ktoso: Is `final val` required to be "compile time constant" (like we do in mailbox debug)?

drewhk: yes, probably

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -313,6 +316,89 @@ private[stream] class VirtualPublisher[T]() extends Publisher[T] {
...
+        }
+      })
+    } else {
+      if (subscriber == registeredSubscriber.get())
+        ReactiveStreamsCompliance.rejectDuplicateSubscriber(subscriber)
+      else
+        ReactiveStreamsCompliance.rejectAdditionalSubscriber(subscriber, "MaterializedValuePublisher")
+    }
+  }
+}
```
ktoso: Should be be covered by TCK `PublisherVerification`, as all our ad-hoc publisher impls.

drewhk: Hmm, will try.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
patriknw: we are not allowed to call the subscriber again, and we should not buble it up to wrong places. I think we mostly do:

    } catch {
      case _: SpecViolation ⇒ // nothing we can do
    }

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
patriknw: ah, looked at wrong field, but is there a reason for not using sealed trait and case objects (as we normally do for enums)

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
ktoso: I can help you out with this (prep the class), should be faster, if you can wait a bit (maybe 15min, so I can finish up something else)?

You'd need to set the TCK up for "publisher with at most 1 element".

ktoso: @drewhk just add this class in `akka-stream-tck-experimental` https://gist.github.com/ktoso/e91f71f7865e397c2d64
Can't test much but better than not being covered :)

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: maybe also set the requestState to Completed. Ok, thanks!

ktoso:
LGTM after `final val` and the TCK test added.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: I just factored this from a boolean field to include a completed state. It just felt like overkill to create three objects and have a reference instead of a simple integer. What would we gain from the case objects?

drewhk:
Added TCK test, handling of SpecViolations and other errors, added Java and Scala documentation with samples

akka-docs-dev/rst/java/stream-graphs.rst
```diff
@@ -194,6 +194,24 @@ together and also turned around with the ``.reversed()`` method. The test
 simulates both parties of a network communication protocol without actually
 having to open a network connection—the flows can just be connected directly.
 
+.. _graph-matvalue-java:
+
+Accessing the materialized value inside the Graph
+-------------------------------------------------
+
+In certain cases it might be necessary to feed back the materialized value of a Graph (partial, closed or backing a
+Source, Sink, Flow or BidiFlow). This is possible by using ``builder.matValue`` which gives an ``Outlet`` that
```
viktorklang: Not sure about abbreviations like that in method names.

What about options such as "product", "endResult", "produces" etc?

drewhk: that would make it worse, since we refer to the thing as the "materialized value" and the above proposal would introduce yet another name that means the same thing. We use this in viaMat, toMat, etc. so I just followed that convention. Maybe we should use m10dValue :)

viktorklang: Fair enough! What's the problem spelling it out?

drewhk: It is just long. While here it might not matter that much, it starts to matter when you use the Flow DSL:

    SourceX.viaMapMaterialized(f1)(Keep.both).viaMapMaterialized(f2)(Keep.left).toMapMaterialized(s)(Keep.right)

compared to

    SourceX.viaMat(f1)(Keep.both).viaMat(f2)(Keep.left).toMat(s)(Keep.right)

viktorklang: Looks more like there's a lack of lifting?

`SourceX.mapMaterialized.via(f1)(Keep.both).via(f2)(Keep.left).to(s)(Keep.right)`

drewhk: er, no? Why? That would need a "stopMapMaterialied" conterpart to be usable, plus it would mean duplicating Flow and creating a MaterializedMappingFlow type, which would in turn lead to explosion at the Graph DSL as well.

viktorklang: I'm sure it's already too late to "fix" this, I'm just knee-jerk-reacting to shortenings in method names. Let's not waste more time here :)

drewhk: I am not sure if it is too late. In fact this might be the final window. I would recommend leaving as it is for this PR and opening a discussion as part of a ticket maybe.

viktorklang: Sounds good. Do what you believe is right, I'll trust your judgement.

drewhk: I am also not particularly happy with "mat" things but so far has found no better solution. I created a ticket: https://github.com/akka/akka/issues/17123

akka-stream/src/main/scala/akka/stream/javadsl/Graph.scala
```diff
@@ -278,10 +277,10 @@ object FlowGraph {
    * The [[Builder]] is mutable and not thread-safe,
    * thus you should construct your Graph and then share the constructed immutable [[FlowGraph]].
    */
-  def builder(): Builder = new Builder()(new scaladsl.FlowGraph.Builder)
+  def builder[Mat](): Builder[Mat] = new Builder()(new scaladsl.FlowGraph.Builder[Mat])
```
viktorklang: Unorthodox type parameter name?

akka-stream/src/main/scala/akka/stream/javadsl/Graph.scala
```diff
@@ -278,10 +277,10 @@ object FlowGraph {
...
 
-  class Builder()(private implicit val delegate: scaladsl.FlowGraph.Builder) { self ⇒
-    import scaladsl.FlowGraph.Implicits._
+  class Builder[Mat]()(private implicit val delegate: scaladsl.FlowGraph.Builder[Mat]) { self ⇒
```
viktorklang: not final?

akka-stream/src/main/scala/akka/stream/javadsl/Graph.scala
```diff
@@ -315,25 +330,26 @@ object FlowGraph {
     def to[I, O](j: UniformFanOutShape[I, O]): ReverseOps[I] = new ReverseOps(j.in)
 
     class ForwardOps[T](out: Outlet[T]) {
```
viktorklang: not final? should be public?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
viktorklang: Consult https://github.com/reactive-streams/reactive-streams-jvm/blob/master/examples/src/main/java/org/reactivestreams/example/unicast/AsyncIterablePublisher.java

akka-stream/src/main/scala/akka/stream/javadsl/Graph.scala
drewhk: Er, why? I really dislike one-letter type parameters except if they have a trivial meaning. We usually use In, Out, for input/output types and M or Mat for materialized types. We have many type parameters, the extra clarity helps.

akka-stream/src/main/scala/akka/stream/javadsl/Graph.scala
drewhk: it can be final, yes

akka-stream/src/main/scala/akka/stream/javadsl/Graph.scala
drewhk: can be final, but must be public. Just look at what it does, how can it be private?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: what is missing here?

akka-stream/src/main/scala/akka/stream/javadsl/Graph.scala
viktorklang: TBH it doesn't tell me anything about what it is—it's even worse because its a real word:

`a piece of coarse material placed on a floor for people to wipe their feet on.`

rkuhn: So, what do you propose? AFAICS we are completely free to change this identifier to anything we want, but `MaterializedType` is out due to obesity.

viktorklang: MaterializedType would be superfluous, since it is a type, so `M` would be perfectly fine, and then add ScalaDoc explaining that M is the Materialized type?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -313,6 +317,118 @@ private[stream] class VirtualPublisher[T]() extends Publisher[T] {
...
+          sub.onError(e)
+        }
+        close()
+        throw e
+    }
+  }
+
+  def setValue(m: Any): Unit =
+    tryOrClose {
+      if (value.compareAndSet(null, m.asInstanceOf[AnyRef]) && requestState.get() == Requested)
```
jrudolph: Is this safe in combination with the below

```scala
if (requestState.compareAndSet(NotRequested, Requested)) {
  val m = value.get()
  f (m ne null) pushAndClose(m)
}
```

I.e. when both run concurrently is it guaranteed that at least one of the `pushAndClose` terms fire? If its already safe, it may warrant an explaining comment. It seems missing both pushAndClose is prevented by allowing it to be called twice, instead, if both `compareAndSet` run concurrently and finish both before the respective other atomic accesses happen.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
viktorklang: No, I meant for reference

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: > I.e. when both run concurrently is it guaranteed that at least one of the pushAndClose terms fire?

I think so. First they do a CAS on their "own" side and a get on the other side. The possible overlaps are (removing symmetric cases where you can relabel A->B, B->A):

- A-CAS
- A-GET
- B-CAS
- B-GET - fires here

- A-CAS
- B-CAS
- A-GET - fires here
- B-GET - fires here (pushAndClose CAS fails)

- A-CAS
- B-CAS
- B-GET - fires here
- A-GET - fires here (pushAndClose CAS fails)

The proof that there are no cases:
 - all permutations of 4 operations are 4! = 24
 - the operations of A and B are cannot be reordered, so there are 24 / (2 * 2) = 6 actual orderings
 - if we don't count cases which are a simple relabeling A->B, B->A, we get 6 / 2 = 3 reorderings which are all enumerated above. 

pushAndClose protects against double onNext by doing a CAS itself. Actually the CAS is probably not necessary on the setValue side, but I like this symmetry and it wouldn't buy much to change it to a SET.

I will add a comment about this.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: Ok thanks! Hopefully we will reduce the number of these ad-hoc Publishers.

akka-stream/src/main/scala/akka/stream/javadsl/Graph.scala
drewhk: I will change it to M then.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: Actually which pushAndClose CAS will fail is nondeterministic, but otherwise correct.

There is a bug though, the above code will not handle null materialized values. I will add a marker object instead.

drewhk:
PLS BUILD

drewhk:
refs: #16168

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
ktoso: Linking to the #16503 "lightweight future publisher" issue.

jrudolph:
On Thursday I thought I had understood how you could use this in HTTP but looking at it again it doesn't seem so obvious any more. The use case in HTTP for needing access to the materialized value is that the materialization of some components depends on the materialized value of others. 

Concretely, we need the materialized ActorRef of a PropsSource during the creation of a stage. So, what would seem like a solution is an out-of-band, pre-run-phase channel that allows to defer the creation of the stage until that other component has been materialized. 

How would you achieve this with this new feature? It seems the feature here forces us to transport the materialized value in-band in the content stream, probably needing a FlexiMerge to get the single, materialized value to the already (pre-)instantiated stage which has to wait until this special value is delivered upon which it can move into its normal running state. This introduces new states and runtime elements for something that could be solved before the pipeline itself has been started at all. Isn't this a layering violation? Is there a simpler way, I'm missing here?

drewhk:
> On Thursday I thought I had understood how you could use this in HTTP but looking at it again it doesn't seem so obvious any more. The use case in HTTP for needing access to the materialized value is that the materialization of some components depends on the materialized value of others.

Buh, this is definitely not a solution for that. I thought that you need access to the materialized value, for example the bound port of the connection. How would the feature you describe even work? Materialization is synchronous, and some materialized values are Futures.

> Concretely, we need the materialized ActorRef of a PropsSource during the creation of a stage. So, what would seem like a solution is an out-of-band, pre-run-phase channel that allows to defer the creation of the stage until that other component has been materialized.

With the upcoming asyncStage support you will be able to send the ActorRef to the stage after the materialized value becomes available. Until the value becomes available, the stage simply backpressures.

> This introduces new states and runtime elements for something that could be solved before the pipeline itself has been started at all. 

WDYM? How can it be solved before the pipeline has been started? I don't see how it is possible.

> Isn't this a layering violation?

WDYM? Turning the materialized value to be an in-band element? The driving motiviation was to make the InetSocketAddress from a binding to be available to the stream. I didn't know that what you actually want is to intertwine stream materialization with the actual running of that stream. I really don't see how that is possible cleanly. We can think about it but it will need a lot of design discussions.

Can you formalize a specification, preferably with a mock API and example so I can exactly see what you need?

jrudolph:
Ah, yes, we had several use-cases for this feature. For the access to the port, I agree this solution is more obvious than for the problem of how to access the ActorRef of the PropsSource.

> Materialization is synchronous, and some materialized values are Futures.

Yes, I guess that's the main problem.

drewhk:
Can you sketch a small example of the feature you want?

jrudolph:
I find it hard to come up with a decent example for a potential API. Conceptually, there's a related, simpler problem which is still a challenge.

Consider that you want to build a pipeline where multiple components need to access the same shared, per-materialization data ("a session", etc.). This is easy to achieve with one-time flows (first, create the shared object, then start the flow and just access the shared object), but how do you do that if you want the flows to be reusable?

As long there's no solution for this case (or the solution is non-obvious or requires reordering the pipeline to somehow inject the value at flow runtime) people (like us) will create one-time flows because those give us a simple, understandable solution.

viktorklang:
Afaik we've been talking about "FlowLocal" in the past, the problem is that
it doesn't work in a distributed setting afaik as it is out-of-band
communication.

On Wed, Apr 8, 2015 at 9:34 AM, Johannes Rudolph <notifications@github.com>
wrote:

> I find it hard to come up with a decent example for a potential API.
> Conceptually, there's a related, simpler problem which is still a challenge.
>
> Consider that you want to build a pipeline where multiple components need
> to access the same shared, per-materialization data ("a session", etc.).
> This is easy to achieve with one-time flows (first, create the shared
> object, then start the flow and just access the shared object), but how do
> you do that if you want the flows to be reusable?
>
> As long there's no solution for this case (or the solution is non-obvious
> or requires reordering the pipeline to somehow inject the value at flow
> runtime) people (like us) will create one-time flows because those give us
> a simple, understandable solution.
>
> —
> <https://github.com/akka/akka/pull/17097#issuecomment-90831064>.
>



-- 
Cheers,
√

drewhk:
I think the current solution, together with the upcoming async stage will enable most of the use cases that were discussed, even if in a more boilerplatey fashion. I will start to think about dynamic layouts though, and see where we can take this to be more flexible and friendly.

jrudolph:
@viktorklang doesn't this depend on the nature of the shared data? If the data isn't distributable in the first place, it also won't be if it is transported in-band. And if it is distributable, distributing it before the flow runs (i.e. out-of-band or actually "before-band") should be possible in the same way as it is afterwards in-band.

jrudolph:
@drewhk yes, but why should you replace working and readable code that produces one-time flows with code, that needs to jump through hoops to get things done? Also, it can only be slower because work that could be done in a simple fashion (simple creation of objects, synchronous flow of logic, clear separation of phases) before the flow runs now needs to be deferred to the runtime of the flow (extra wrappers to transport the extra data in-band, asynchronous execution, dynamically changing execution flow, etc.).

viktorklang:
@jrudolph Since it would be available to both the upstream(s) and downstreams(s) it would require updates to travel in both directions over the wire out-of-band.

drewhk:
@jrudolph The current approach in this PR is working and is not invasive -- it fitst the underlying model. I don't see how it affects Flow reusability. Even if you cannot use this abstraction, it should remain as it is very useful.

>  could be done in a simple fashion (simple creation of objects, synchronous flow of logic, clear separation of phases)

What do you mean? How? If you have a solution then please share because I do not see how this is possible (not saying it isn't but I don't see how it can be done in code without horrible workarounds). You imply it is simple, but I don't think it is. It would help tremendously if you would completely explain what you need.

In summary, I do not yet see how I can do what you want, nor do I really see what is what you want. Can you please completely elaborate what do you think is a hoop-free abstraction and what features do you need? In full detail? 

Don't take my comments as arguing against your requests, but take it as total confusion, I want to see a much more clear specification so I can actually start arguing for or agains, or starting thinking on the design.

rkuhn:
Stepping back a bit, we first need a precise description of the use-cases that we aim to support. Then we shall implement those cases using the features that we already have (including this one), and then we can judge the amount of boilerplate involved and whether/how this can be reduced without compromising on semantic integrity of the abstractions and without making the internal implementation a complete mess.

> (The latter might not seem like an inevitable a priori requirement, but past experience has shown that such messy internals always ended up being connected to core design issues even if we didn’t foretell them at the time of feature invention.)

So, with this plan: @jrudolph and @sirthias please provide exact specifications for the use-cases to be implemented so that we can add them (prototypically) to the test suite—for regression testing as well as further design inspiration.

drewhk:
As a starting point, lets summarize the current state.

 - The construction phase involves building up a description of the components. This description is expressed in StreamLayout.scala and it is an immutable, hierarchic representation of processing modules with arbitrary input and output ports and a materialized value. The representation is constructed by the user using either the Flow or Graph DSL, but internally they look the same
 - The materialization phase takes this internal representation, and does a recursive traversal. The traversal part is expressed in the MaterializerSession class (in StreamLayout.scala), this encodes everything that is not actor specific. The actor specific parts are implemented in ActorFlowMaterializerImpl which subclasses MaterializerSession and is able to materialize atomic modules (those which have no submodules)
   - for every component all of its subcomponents are visited and materialized (recursively, so it starts again from this step for the subcomponent)
   - atomic components are materialized by the implementors of MaterializerSession
   - as a submodule is materialized, we start to connect their Publishers and Subscribers to their corresponding Subscribers, Publishers if they are available (at some point all pieces fall into place and everything becomes connected). This basically starts regions of the graph.
   - for every materialized submodule we get its materialized value, and we put it into a map
   - once we enumerated and materialized all submodules, we execute an AST that describes how to calculate the materialized value of this module (taking the map we constructed above as argument).
   - once this materialized value is computed, we return to the previous step of recursion

All of this logic except the DSL is contained in StreamLayout.scala

The pending questions:
 - What is the use cases we need to support?
 - How should the API/DSL look like?
 - How can the constructs from the API/DSL encoded in the representation retaining all of its niceness?
 - How can the materialization process be changed so that it supports the new constructs, while retaining all of its niceness?

jrudolph:
@rkuhn yes, I guess it makes sense to look at concrete examples.

The original case from above:

> Concretely, we need the materialized ActorRef of a PropsSource during the creation of a stage.

So, in much simplified terms:

```scala
val propsSource = Source[Token](Props(...))
val parser = Flow[ByteString].transform(() => new Parser(needActorRefHere))

FlowGraph.partial() { implicit b =>
  // use propsSource and parser here to build graph
}
```

Then there's the case of getting the peer port to the parser in the HTTP server (to fix #16641). This isn't implemented right now, but could theoretically be like this (using the existing one-time flow):

```scala
// from Http.scala
def bind(...): Source[IncomingConnection, Future[ServerBinding]] = {  
  // ...
  connections.map { conn ⇒
      val flow = Flow(conn.flow, serverBlueprint(conn.localAddress, conn.remoteAddress))(Keep.right) { implicit b ⇒
        (tcp, http) ⇒
          import FlowGraph.Implicits._
          tcp.outlet ~> http.bytesIn
          http.bytesOut ~> tcp.inlet
          (http.httpResponses, http.httpRequests)
      }

      IncomingConnection(conn.localAddress, conn.remoteAddress, flow)
    }
  // ...
}
```

Here, the ports aren't a materialized value but somehow need to be passed to the `serverBlueprint` in some way. It seems we would rather need something like a "MaterializedParameter" than a "MaterializedValue" here. We could also create a Source with the values which will only ever be used one time and connect it to a reusable flow (but does that help in any way?).

With websockets, we have something similar to the ActorRef case where we need to create a "hole" in the flow where the user can plug in the actual websocket handler during runtime. Currently, we have a special `oneTimePublisherSink` and `oneTimeSubscriberSource` which put the publisher and subscriber they materialize into a place where the implementation can find it later. You can see https://github.com/akka/akka/pull/17131/files#diff-d651359e50b8308f2ae5c4075ee2f403R313 for the details.

For me it looks like that for the server-side it isn't clear why the server side graph should be reusable at all because it will have one-time elements (the tcp side source and sink and the ports) in any case. So, maybe that isn't the interesting use-case in the first place? Maybe we should rather look at the client side which should return a reusable flow on the API level?

drewhk:
> Concretely, we need the materialized ActorRef of a PropsSource during the creation of a stage.

The example you provided can be done with this PR and an async Stage (soon to come). One thing is that the Flow must know about the Source[..., ActorRef], unless the type of the materialized value is not known. You can construct the resulting component (will it be a Source?) as a graph, using the materialized value port this PR provides and injecting it into the Stage using the upcoming async stage support. Then this component will be fully reusable, since every materialization will feed the stage with the correct ActorRef.

This is boilerplatey, yes, but we can build sugar on top of it at the DSL layer because it does not need internals to be changed -- all the necessary features are available.

> Here, the ports aren't a materialized value but somehow need to be passed to the serverBlueprint in some way. It seems we would rather need something like a "MaterializedParameter" than a "MaterializedValue" here. We could also create a Source with the values which will only ever be used one time and connect it to a reusable flow (but does that help in any way?).

The way to model is that you keep a "hole" in your reusable component that will be filled in with the changing part. In this case it would be a Source[Port]. This is currently bolierplatey, yes, and would make the reusable component not a BidiStage anymore. Since this does not affect the internal model again, I think it is possible to put sugar on top of it (i.e. solve it at the DSL level).

So far, these examples do seem like as a stream *parametrization* problem, i.e. they don't change the layout of the stream layout dynamically. The distinct features that I decoded from these samples:

 - parametrization by external values, i.e. having modules that are basically a (param) => Graph (conceptually). There are ways to implement it now, but we can probably make it nicer. The challenge here is a proper API (type parameter blowup). This method can in a sense change the layout of the graph, as the parameters are known at the start of materialization, so the materializer can just calculate the proper module from the parameter when it reaches it during materialization. This could be useful for example for parametrizing a worker pool. Every module that does not depend on the parameter can be preoptimized (for example fused).

 - parametrization by materialized values, i.e. modules that are basically a (mat) => AtomicModule (currently stages and in the future Octopus things). This method does not change the layout as the previous one so the materializer is allowed to construct the module and feed to it the materialized value later, once it becomes available (leaving the stage backpressured). This also means that we can allow the user to map over the materialized value (waiting on a future -- flattening it) before feeding it back. Until the materialized value becomes available the module is backpressuring. The reason why this parametrization is not allowed to change the layout is because at that point in materialization where the materialized value becomes available the current module has been already materialized (every submodule is running, otherwise the mat value might not be calculatable). These parametrized modules can be fully preoptimized (for example fused) since they are treated as atomics.

Both of these seem to be doable. The challenge is probably the proper API.

drewhk:
In a way, both features are kind of a "meta-factory" pattern as Graphs are already factories, and these features will parametrize these factories, therefore being "meta-factories". I might see now that this eerily resembles a phased approach, but I do not see yet how that generalization maps to implementation.

Issue #16521, opened 2014-12-11
implement ActorRefSink


rkuhn:
Similar in spirit to #16520 we should have a Sink that wraps an ActorRef. In order to allow the target Actor to be restarted without issues the protocol between the stream endpoint and that Actor needs to allow for a way to reset outstanding demand to a known value. We should also include an identifying token so that a single target can ingest multiple streams.

Issue #16520, opened 2014-12-11
implement ActorRefSource


rkuhn:
Since the supervision of stream elements is going to place restrictions on the actors, I think that providing a mixin “stream nature” might not be nicely manageable. Instead I propose a Source that materializes to an ActorRef, and everything sent to that ref is injected into the stream (and dropped if there is no demand).

We might be tempted to add configurable buffering, but that exists already in the form of the `.buffer(...)` combinator which we should just fuse into the Actor for this Source (when present).

PR #17098, opened 2015-03-30
!str #16520 #16521 Add ActorRefSource and ActorRefSink

patriknw:
will add something to reference docs if you think this is the right approach

### DIFF
--- a/akka-docs-dev/rst/java/stream-integrations.rst
+++ b/akka-docs-dev/rst/java/stream-integrations.rst
@@ -8,7 +8,8 @@ Integrating with Actors
 =======================
 
 For piping the elements of a stream as messages to an ordinary actor you can use the
-``Sink.actorRef``.  
+``Sink.actorRef``. Messages can be sent to a stream via the :class:`ActorRef` that is 
+materialized by ``Source.actorRef``.
 
 For more advanced use cases the :class:`ActorPublisher` and :class:`ActorSubscriber` traits are
 provided to support implementing Reactive Streams :class:`Publisher` and :class:`Subscriber` with
@@ -27,6 +28,25 @@ Akka Streams :class:`Source` or :class:`Sink`.
   These Actors are designed to be implemented using Java 8 lambda expressions. In case you need to stay on a JVM
   prior to 8, Akka provides :class:`UntypedActorPublisher` and :class:`UntypedActorSubscriber` which can be used
   easily from any language level.
+
+Source.actorRef
+^^^^^^^^^^^^^^^
+
+Messages sent to the actor that is materialized by ``Source.actorRef`` will be emitted to the 
+stream if there is demand from downstream, otherwise they will be buffered until request for 
+demand is received.
+
+Depending on the defined :class:`OverflowStrategy` it might drop elements if there is no space
+available in the buffer.
+
+The stream can be completed successfully by sending ``akka.actor.PoisonPill`` or
+``akka.actor.Status.Success`` to the actor reference.
+
+The stream can be completed with failure by sending ``akka.actor.Status.Failure`` to the 
+actor reference.
+
+The actor will be stopped when the stream is completed, failed or cancelled from downstream,
+i.e. you can watch it to get notified when that happens.
 
 Sink.actorRef
 ^^^^^^^^^^^^^
--- a/akka-docs-dev/rst/java/stream-integrations.rst
+++ b/akka-docs-dev/rst/java/stream-integrations.rst
@@ -7,8 +7,12 @@ Integration
 Integrating with Actors
 =======================
 
-:class:`AbstractActorPublisher` and :class:`AbstractActorSubscriber` are two traits that provides support for
-implementing Reactive Streams :class:`Publisher` and :class:`Subscriber` with an :class:`Actor`.
+For piping the elements of a stream as messages to an ordinary actor you can use the
+``Sink.actorRef``.  
+
+For more advanced use cases the :class:`ActorPublisher` and :class:`ActorSubscriber` traits are
+provided to support implementing Reactive Streams :class:`Publisher` and :class:`Subscriber` with
+an :class:`Actor`.
 
 These can be consumed by other Reactive Stream libraries or used as a
 Akka Streams :class:`Source` or :class:`Sink`.
@@ -24,6 +28,20 @@ Akka Streams :class:`Source` or :class:`Sink`.
   prior to 8, Akka provides :class:`UntypedActorPublisher` and :class:`UntypedActorSubscriber` which can be used
   easily from any language level.
 
+Sink.actorRef
+^^^^^^^^^^^^^
+
+The sink sends the elements of the stream to the given `ActorRef`. If the target actor terminates
+the stream will be cancelled. When the stream is completed successfully the given ``onCompleteMessage``
+will be sent to the destination actor. When the stream is completed with failure a ``akka.actor.Status.Failure``
+message will be sent to the destination actor.
+
+.. warning::
+
+   There is no back-pressure signal from the destination actor, i.e. if the actor is not consuming
+   the messages fast enough the mailbox of the actor will grow. For potentially slow consumer actors
+   it is recommended to use a bounded mailbox with zero `mailbox-push-timeout-time` or use a rate
+   limiting stage in front of this stage.
 
 ActorPublisher
 ^^^^^^^^^^^^^^
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefSinkSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefSinkSpec.scala
@@ -1,0 +1,50 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.scaladsl
+
+import akka.stream.ActorFlowMaterializer
+import akka.stream.testkit.AkkaSpec
+import akka.stream.testkit.StreamTestKit
+import akka.actor.Actor
+import akka.actor.ActorRef
+import akka.actor.Props
+
+object ActorRefSinkSpec {
+  case class Fw(ref: ActorRef) extends Actor {
+    def receive = {
+      case msg ⇒ ref forward msg
+    }
+  }
+}
+
+class ActorRefSinkSpec extends AkkaSpec {
+  import ActorRefSinkSpec._
+  implicit val mat = ActorFlowMaterializer()
+
+  "A ActorRefSink" must {
+
+    "send the elements to the ActorRef" in {
+      Source(List(1, 2, 3)).runWith(Sink.actorRef(testActor, onCompleteMessage = "done"))
+      expectMsg(1)
+      expectMsg(2)
+      expectMsg(3)
+      expectMsg("done")
+    }
+
+    "cancel stream when actor terminates" in {
+      val publisher = StreamTestKit.PublisherProbe[Int]()
+      val fw = system.actorOf(Props(classOf[Fw], testActor).withDispatcher("akka.test.stream-dispatcher"))
+      Source(publisher).runWith(Sink.actorRef(fw, onCompleteMessage = "done"))
+      val autoPublisher = new StreamTestKit.AutoPublisher(publisher)
+      autoPublisher.sendNext(1)
+      autoPublisher.sendNext(2)
+      expectMsg(1)
+      expectMsg(2)
+      system.stop(fw)
+      autoPublisher.subscription.expectCancellation()
+    }
+
+  }
+
+}
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefSourceSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefSourceSpec.scala
@@ -1,0 +1,81 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.scaladsl
+
+import scala.concurrent.duration._
+import akka.stream.ActorFlowMaterializer
+import akka.stream.OverflowStrategy
+import akka.stream.testkit.AkkaSpec
+import akka.stream.testkit.StreamTestKit
+import akka.actor.PoisonPill
+import akka.actor.Status
+
+class ActorRefSourceSpec extends AkkaSpec {
+  implicit val mat = ActorFlowMaterializer()
+
+  "A ActorRefSource" must {
+
+    "emit received messages to the stream" in {
+      val s = StreamTestKit.SubscriberProbe[Int]()
+      val ref = Source.actorRef(10, OverflowStrategy.fail).to(Sink(s)).run()
+      val sub = s.expectSubscription
+      sub.request(2)
+      ref ! 1
+      s.expectNext(1)
+      ref ! 2
+      s.expectNext(2)
+      ref ! 3
+      s.expectNoMsg(500.millis)
+    }
+
+    "buffer when needed" in {
+      val s = StreamTestKit.SubscriberProbe[Int]()
+      val ref = Source.actorRef(100, OverflowStrategy.dropHead).to(Sink(s)).run()
+      val sub = s.expectSubscription
+      for (n ← 1 to 20) ref ! n
+      sub.request(10)
+      for (n ← 1 to 10) s.expectNext(n)
+      sub.request(10)
+      for (n ← 11 to 20) s.expectNext(n)
+
+      for (n ← 200 to 399) ref ! n
+      sub.request(100)
+      for (n ← 300 to 399) s.expectNext(n)
+    }
+
+    "terminate when the stream is cancelled" in {
+      val s = StreamTestKit.SubscriberProbe[Int]()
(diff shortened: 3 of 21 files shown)


patriknw:
Refs #16521

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -200,3 +200,24 @@ private[akka] final class PropsSink[In](props: Props, val attributes: OperationA
...
+/**
+ * INTERNAL API
+ */
+private[akka] final class TellSink[In](ref: ActorRef, onCompleteMessage: Any,
+                                       val attributes: OperationAttributes,
+                                       shape: SinkShape[In]) extends SinkModule[In, Unit](shape) {
+
+  override def create(materializer: ActorFlowMaterializerImpl, flowName: String) = {
+    val subscriberRef = materializer.actorOf(
+      TellSinkActor.props(ref, materializer.settings.maxInputBufferSize, onCompleteMessage),
```
drewhk: hmm, do we really need an actor for this?

rkuhn:
This is certainly the most simple approach, but I fear that on the Sink side this is too simplistic: we discussed dropping things on the Source side, and I think the Sink should also drop things instead of just sending. We could recommend people to configure a bounded mailbox for the recipient, another possibility would be to have an extreme simple work pulling protocol:
~~~scala
object Sink {
  def actorRef[T]: Sink[T, ActorRef]
}

case class Request(n: Long, target: ActorRef)
~~~
Then the Subscriber that this is Materialized to will keep a list of requests and serve them in order, which would even allow multiple actors to pull data out of this stream. This could in principle even stop requesting from upstream when there is no demand …

Is this still simple enough to be done right now?

rkuhn:
Ah, another thought that makes this more resilient: allow client actors to set their demand to a specific value instead of using normal additive semantics. This allows them to just unconditionally issue that command after restarts.

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
patriknw: I used an actor because of cancellation when the destination actor terminates, if we don't want that feature we don't need an actor here

patriknw:
@rkuhn renamed to ActorRefSink and added documentation. Please take a second look.
Also renamed the factory for ActorSubscriber props Sink, from apply to actorSubscriber.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefSinkSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+}
+
+class ActorRefSinkSpec extends AkkaSpec {
+  import ActorRefSinkSpec._
+  implicit val mat = ActorFlowMaterializer()
+
+  "A ActorRefSink" must {
+
+    "send the elements to the ActorRef" in {
+      Source(List(1, 2, 3)).runWith(Sink.actorRef(testActor, onCompleteMessage = "done"))
```
drewhk: in many cases this will be more useful than the subscriberProbe

akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefSinkSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+      val publisher = StreamTestKit.PublisherProbe[Int]()
+      val fw = system.actorOf(Props(classOf[Fw], testActor).withDispatcher("akka.test.stream-dispatcher"))
+      Source(publisher).runWith(Sink.actorRef(fw, onCompleteMessage = "done"))
+      val autoPublisher = new StreamTestKit.AutoPublisher(publisher)
+      autoPublisher.sendNext(1)
+      autoPublisher.sendNext(2)
+      expectMsg(1)
+      expectMsg(2)
+      system.stop(fw)
+      autoPublisher.subscription.expectCancellation()
```
drewhk: so nice!

akka-stream/src/main/scala/akka/stream/impl/ActorRefSinkActor.scala
```diff
@@ -0,0 +1,36 @@
...
+ */
+private[akka] object ActorRefSinkActor {
+  def props(ref: ActorRef, highWatermark: Int, onCompleteMessage: Any): Props =
+    Props(new ActorRefSinkActor(ref, highWatermark, onCompleteMessage))
+}
+
+/**
+ * INTERNAL API
+ */
+private[akka] class ActorRefSinkActor(ref: ActorRef, highWatermark: Int, onCompleteMessage: Any) extends ActorSubscriber {
```
drewhk: it would be interesting to experiment later if this can be implemented via a lightweight synthetic actorRef just like PromiseActorRef.

akka-stream/src/main/scala/akka/stream/javadsl/Sink.scala
```diff
@@ -99,6 +91,33 @@ object Sink {
...
+   * When the stream is completed successfully the given `onCompleteMessage`
+   * will be sent to the destination actor.
+   * When the stream is completed with failure a [[akka.actor.Status.Failure]]
+   * message will be sent to the destination actor.
+   *
+   * It will request at most `maxInputBufferSize` number of elements from
+   * upstream, but there is no back-pressure signal from the destination actor,
+   * i.e. if the actor is not consuming the messages fast enough the mailbox
+   * of the actor will grow. For potentially slow consumer actors it is recommended
+   * to use a bounded mailbox with zero `mailbox-push-timeout-time` or use a rate
```
drewhk: unfortunately that would lose the completion message, too :(

akka-stream/src/main/scala/akka/stream/impl/ActorRefSinkActor.scala
```diff
@@ -0,0 +1,36 @@
...
+  import ActorSubscriberMessage._
+
+  override val requestStrategy = WatermarkRequestStrategy(highWatermark)
+
+  context.watch(ref)
+
+  def receive = {
+    case OnNext(elem)      ⇒ ref ! elem
+    case OnError(cause)    ⇒ ref ! Status.Failure(cause)
```
drewhk: shouldn't we stop the actor here? Or it will be done by the ActorSubscriber superclass?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefSinkSpec.scala
patriknw: exactly what I was thinking about, then you have the full facilities of TestKit
@2m might be something to keep in mind when shaping up the stream-testkit

akka-stream/src/main/scala/akka/stream/impl/ActorRefSinkActor.scala
patriknw: good catch, thanks

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -200,3 +200,24 @@ private[akka] final class PropsSink[In](props: Props, val attributes: OperationA
...
+ * INTERNAL API
+ */
+private[akka] final class ActorRefSink[In](ref: ActorRef, onCompleteMessage: Any,
+                                           val attributes: OperationAttributes,
+                                           shape: SinkShape[In]) extends SinkModule[In, Unit](shape) {
+
+  override def create(materializer: ActorFlowMaterializerImpl, flowName: String) = {
+    val subscriberRef = materializer.actorOf(
+      ActorRefSinkActor.props(ref, materializer.settings.maxInputBufferSize, onCompleteMessage),
+      name = s"$flowName-tell")
```
ktoso: `$flowName-actorRef`

akka-stream/src/main/scala/akka/stream/impl/ActorRefSinkActor.scala
```diff
@@ -0,0 +1,36 @@
+package akka.stream.impl
```
ktoso: missing copyright

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
patriknw: thx

patriknw:
I have added one commit with the review feedback.
Another commit with the ActorRefSource, will add reference docs for ActorRefSource when we have sorted out the buffering questions.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefSourceSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+      ref ! 2
+      s.expectNext(2)
+      ref ! 3
+      s.expectNoMsg(500.millis)
+    }
+
+    "be useful together with buffer" in {
+      val s = StreamTestKit.SubscriberProbe[Int]()
+      // FIXME isn't there a race when starting up, if I send to the ref before the buffer has requested?
+      //       Do we ignore that and fix that with fusing, or shall I add an internal buffer to the ActorRefSourceActor?
```
patriknw: @drewhk I would like to hear what you think about this

drewhk: This is exactly what I said in the meeting. Just slapping a buffer stage after an element will not fully solve the buffering issue. Fusing is not necessarily possible, in this case for example the source is backed by a custom actor.

Originally I imagined a simple fire-and-forget interface that wraps a bounded buffer and an AtomicLong which tracks pending demand, using only the Publisher interfaces to the downstream, and a simple fire(elem) call. I guess you can put the buffer to the actor, but of course then there is the implicit additional buffer, its mailbox.

patriknw: That matches my conclusions and sounds like adding an internal buffer would not be harmful. Then the buffer size is also very explicit. Opinions @rkuhn

drewhk: I guess the MaterializedValuePublisher I implemented in the other PR can be taken as a template, augmented with a buffer and proper request tracking to have a simple Fire-and-Forget Publisher if we want to pursue that path.

patriknw:
I have added the internal buffer. Used same overflow strategies (except Backpressure) as in the buffer stage.

patriknw:
Refs #16520

akka-stream/src/main/scala/akka/stream/impl/ActorRefSourceActor.scala
```diff
@@ -0,0 +1,65 @@
...
+  import akka.stream.actor.ActorPublisherMessage._
+  import akka.stream.OverflowStrategy._
+
+  require(overflowStrategy != OverflowStrategy.Backpressure, "Backpressure overflowStrategy not supported")
+
+  private val buffer = FixedSizeBuffer(bufferSize)
+
+  def receive = {
+    case _: Request ⇒
+      // count is tracked by super
```
2m: s/count/totalDemand

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -322,4 +314,33 @@ object Source extends SourceApply {
...
+   * Creates a `Source` that is materialized as an [[akka.actor.ActorRef]].
+   * Messages sent to this actor will be emitted to the stream if there is demand from downstream,
+   * otherwise they will be buffered until request for demand is received.
+   *
+   * Depending on the defined [[akka.stream.OverflowStrategy]] it might drop elements if
+   * there is no space available in the buffer.
+   *
+   * The buffer can be disabled by using `bufferSize` of 0 and then received messages are dropped
+   * if there is no demand from downstream. When `bufferSize` is 0 the `overflowStrategy` is does
+   * not matter.
```
ktoso: not needed `is` in the above sentence?

```
When `bufferSize` is 0 the `overflowStrategy` does not matter.
```
?

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -160,6 +152,32 @@ object Source {
...
+  /**
+   * Creates a `Source` that is materialized as an [[akka.actor.ActorRef]].
+   * Messages sent to this actor will be emitted to the stream if there is demand from downstream,
+   * otherwise they will be buffered until request for demand is received.
+   *
+   * Depending on the defined [[akka.stream.OverflowStrategy]] it might drop elements if
+   * there is no space available in the buffer.
+   *
+   * The buffer can be disabled by using `bufferSize` of 0 and then received messages are dropped
+   * if there is no demand from downstream. When `bufferSize` is 0 the `overflowStrategy` is does
```
ktoso: additional `is` not needed here

ktoso:
LGTM, will be very useful!

akka-docs-dev/rst/java/stream-integrations.rst
```diff
@@ -24,6 +29,30 @@ Akka Streams :class:`Source` or :class:`Sink`.
...
+the stream will be cancelled. When the stream is completed successfully the given ``onCompleteMessage``
+will be sent to the destination actor. When the stream is completed with failure a ``akka.actor.Status.Failure``
+message will be sent to the destination actor.
+
+.. warning::
+
+   There is no back-pressure signal from the destination actor, i.e. if the actor is not consuming
+   the messages fast enough the mailbox of the actor will grow. For potentially slow consumer actors
+   it is recommended to use a bounded mailbox with zero `mailbox-push-timeout-time` or use a rate
+   limiting stage in front of this stage.
```
rkuhn: “… in front of this _Sink_.”

akka-stream/src/main/scala/akka/stream/OverflowStrategy.scala
```diff
@@ -54,6 +54,8 @@ object OverflowStrategy {
    */
   def dropBuffer: OverflowStrategy = DropBuffer
 
+  // FIXME shouldn't there be a strategy that drops the new element?
```
rkuhn: Yes, definitely! I thought that was dropTail, but I was mistaken, please add a ticket.

akka-stream/src/main/scala/akka/stream/impl/ActorRefSourceActor.scala
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * INTERNAL API
+ */
+private[akka] class ActorRefSourceActor(bufferSize: Int, overflowStrategy: OverflowStrategy)
+  extends akka.stream.actor.ActorPublisher[Any] with ActorLogging {
+  import akka.stream.actor.ActorPublisherMessage._
+  import akka.stream.OverflowStrategy._
+
+  require(overflowStrategy != OverflowStrategy.Backpressure, "Backpressure overflowStrategy not supported")
```
rkuhn: better move this `require` into the Props factory method

akka-stream/src/main/scala/akka/stream/javadsl/Sink.scala
rkuhn: yes, but the target Actor can still watch this one and thereby know when the stream ended

rkuhn: ah, the Terminated would be lost as well …

rkuhn:
LGTM: I agree that this is nice and simple and will be very useful—with one addition, namely how can the ActorRefSource be complete/fail the stream?

patriknw:
@rkuhn good point regarding ActorRefSource complete/fail. I suggest that `akka.actor.Status.Failure` can be sent to the ref to fail the stream and `PoisonPill` (or `system.stop`) will complete the stream successfully. WDYT?

rkuhn:
Why not make proper message types in the ActorRefSource companion?

patriknw:
Because then I will be responsible for making proper serialization for these messages, unless we continue to use java serialization.
I have already used `Status.Failure` for the ActorRefSink so using it on this side also makes it symmetrical.

drewhk:
It makes sense for me. And the rest is LGTM

patriknw:
ok, I'll complete this now, so @2m can use it

akka-stream/src/main/scala/akka/stream/impl/ActorRefSourceActor.scala
patriknw: ok

akka-stream/src/main/scala/akka/stream/OverflowStrategy.scala
patriknw: ticket added: #17167

patriknw:
I have implemented and documented the completion/fail: c0185b2

Will squash and merge after pr validation

patriknw:
failure was unrelated, created ticket: https://github.com/akka/akka/issues/17172

patriknw:
rebased and squashed

drewhk:
LGTM, merge it!

patriknw:
yes, waiting for another validation, had some merge conflicts when rebasing

patriknw:
now it was a real test failure, because FixedSizeBuffer had changed (to the better)
one more time!

PR #17099, opened 2015-03-30
=htc #16807 introduce `akka.http.parsing.error-logging-verbosity` setting

### LINKED ISSUE #16807: Change default header parsing error message to be logged as a single line
and allow configuration of more verbose messages for debugging purposes.

See spray/spray#977.

/cc @sirthias

sirthias:
Closes #16807.

/cc @jrudolph

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/engine/parsing/ParserSettings.scala
+++ b/akka-http-core/src/main/scala/akka/http/engine/parsing/ParserSettings.scala
@@ -4,6 +4,7 @@
 
 package akka.http.engine.parsing
 
+import java.util.Locale
 import com.typesafe.config.Config
 import scala.collection.JavaConverters._
 import akka.http.model.{ StatusCode, HttpMethod, Uri }
@@ -21,6 +22,7 @@ final case class ParserSettings(
   maxChunkSize: Int,
   uriParsingMode: Uri.ParsingMode,
   illegalHeaderWarnings: Boolean,
+  errorLoggingVerbosity: ParserSettings.ErrorLoggingVerbosity,
   headerValueCacheLimits: Map[String, Int],
   customMethods: String ⇒ Option[HttpMethod],
   customStatusCodes: Int ⇒ Option[StatusCode]) extends HttpHeaderParser.Settings {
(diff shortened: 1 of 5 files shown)


jrudolph:
LGTM

drewhk:
LGTM

akka-http-core/src/main/scala/akka/http/engine/parsing/ParserSettings.scala
```diff
@@ -66,9 +67,25 @@ object ParserSettings extends SettingsCompanion[ParserSettings]("akka.http.parsi
...
   }
+
+  sealed trait ErrorLoggingVerbosity
+  object ErrorLoggingVerbosity {
+    case object Off extends ErrorLoggingVerbosity
+    case object Simple extends ErrorLoggingVerbosity
+    case object Full extends ErrorLoggingVerbosity
+
+    def apply(string: String): ErrorLoggingVerbosity =
+      string match {
```
patriknw: to be consistent with other such settings:

    string.toLowerCase(Locale.ROOT) match

akka-http-core/src/main/scala/akka/http/engine/parsing/ParserSettings.scala
```diff
@@ -66,9 +67,25 @@ object ParserSettings extends SettingsCompanion[ParserSettings]("akka.http.parsi
...
+        case "off"    ⇒ Off
+        case "simple" ⇒ Simple
+        case "full"   ⇒ Full
+        case _        ⇒ throw new IllegalArgumentException(string + " is not a legal `error-logging-verbosity` setting")
```
patriknw: We try to enclose values in `[ ]` in exception and log messages (we are not always consistent with this).
Here it would make sense, since it would make it more easy to spot empty string value.

patriknw:
LGTM after toLowerCase

akka-http-core/src/main/scala/akka/http/engine/parsing/ParserSettings.scala
sirthias: ok, will change.

akka-http-core/src/main/scala/akka/http/engine/parsing/ParserSettings.scala
sirthias: Ok, will do.

ktoso:
LGTM, merging

PR #17101, opened 2015-03-30
=htc #16804 make `akka.http.model.RangeUnit.Bytes` a `case object`

### LINKED ISSUE #16804: Make sure HTTP domain model objects are case objects
For serializability (to avoid surprises) and to support a proper toString method.

This was originally requested as spray/spray#954.

/cc @sirthias

sirthias:
Closes #16804.

I also went through the complete model package structure and verified that all other `object` instances are indeed either case objects or extend a `case class`.

/cc @jrudolph


ktoso:
LGTM, not expecting any objections I guess :-)

drewhk:
LGTM

PR #17102, opened 2015-03-30
=act #17087 PromiseActorRef (ask pattern) must send right Terminated()

### LINKED ISSUE #17087: Memory leak with context.watch and pattern.ask involved
Hi guys,
Here is a simple example of use case when PromiseActorRef is leaking when get watched.

```scala
import akka.actor.{Actor, ActorSystem, Props, Terminated}
import akka.pattern.ask
import akka.util.Timeout

object MemoryLeak extends App {
  val system = ActorSystem()
  val actor = system.actorOf(Props[MyActor])
  implicit val timeout = Timeout(3.seconds)
  while (true) {
    Thread.sleep(100)
    actor ? "Request"
  }
}

class MyActor extends Actor {
  def receive = {
    case Terminated(actor) => context unwatch actor
    case "Request"         =>
      val actor = sender()
      context.watch(actor)
      actor ! "Response"
  }
}
```

I can see two possible solutions: 
* `context.watch` should ignore `PromiseActorRef`
* Watchers should receive `Terminated`

ktoso:
Resolves #17087

The `PromiseActorRef` had all logic in place, but was putting a wrong actor ref into the `DeathWatchNotification` (telling the watcher that the watcher died ;-)).

### DIFF
--- a/akka-actor/src/main/scala/akka/pattern/AskSupport.scala
+++ b/akka-actor/src/main/scala/akka/pattern/AskSupport.scala
@@ -3,16 +3,16 @@
  */
 package akka.pattern
 
-import language.implicitConversions
-
 import java.util.concurrent.TimeoutException
+
 import akka.actor._
 import akka.dispatch.sysmsg._
-import scala.annotation.tailrec
-import scala.util.control.NonFatal
-import scala.concurrent.{ Future, Promise, ExecutionContext }
 import akka.util.{ Timeout, Unsafe }
-import scala.util.{ Success, Failure }
+
+import scala.annotation.tailrec
+import scala.concurrent.{ ExecutionContext, Future, Promise }
+import scala.language.implicitConversions
+import scala.util.{ Failure, Success }
 
 /**
  * This is what is used to complete a Future that is returned from an ask/? call,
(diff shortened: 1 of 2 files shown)


ktoso:
This will need backport to 2.3

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
```diff
@@ -308,7 +307,7 @@ private[akka] final class PromiseActorRef private (val provider: ActorRefProvide
         watchers foreach { watcher ⇒
           // ➡➡➡ NEVER SEND THE SAME SYSTEM MESSAGE OBJECT TO TWO ACTORS ⬅⬅⬅
           watcher.asInstanceOf[InternalActorRef]
-            .sendSystemMessage(DeathWatchNotification(watcher, existenceConfirmed = true, addressTerminated = false))
+            .sendSystemMessage(DeathWatchNotification(this, existenceConfirmed = true, addressTerminated = false))
```
drewhk: LOL

drewhk:
LGTM

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
t3hnar: :D

viktorklang: OMG

viktorklang: How about (to get rid of the explicit cast):
~~~scala
watchers foreach { case watcher: InternalActorRef => … }

rkuhn: well, so much for reviews … never trust the “simple stuff”!

rkuhn:
LGTM

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
viktorklang: We need moar dependent types ;)

ktoso: @viktorklang that would be really cool :]

PR #17103, opened 2015-03-30
!htc #16803 introduce proper model for type of media-type encoding specs

### LINKED ISSUE #16803: Undefined inclusion of charset parameter in Content-Type header of certain Media Types
Originally raised by @mackler as spray/spray#924

_Originally raised as [this question](https://groups.google.com/d/msg/spray-user/nsU46VnK3rU/43e2RfQvTSMJ)._

Given a custom `MediaType`:

    val `application/mybinary` = spray.http.MediaType.custom(
      mainType = "application",
      subType = "mybinary",
      binary = true                                                                          
    )

and its use when marshaling a binary type:

    implicit val marshaller: Marshaller[SomeBinary] =
        Marshaller.delegate[SomeBinary,Array[Byte]](`application/mybinary`){
          _.toByteArray
        }

and a route:

    complete(sb: SomeBinary)

then a Spray server replies with a `Content-Type` header that includes a `charset` parameter:

    HTTP/1.1 200 OK
    Date: Mon, 04 Aug 2014 10:01:06 GMT
    Content-Type: application/mybinary; charset=UTF-8
    Content-Length: 128

I propose that passing value `true` for the `binary` parameter to `MediaType.custom()` indicates that character encoding is meaningless for the defined media type and that Spray ought to behave accordingly by generating a `Content-Type` header that lacks a `charset` parameter.

### References

The format and use of Media Types in `Content-Type` headers is specified in [section 3.7 of the HTTP 1.1 specification](http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7).  That specification refers to the `charset=UTF-8` in question as the Media Type's `parameter`.

[Section 3.7.1](http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7.1) specifically refers to use of the `charset` parameter in the context of `text` types, and indicates in passing that, "[t]he `charset` parameter is used with _some_ media types to define the character set of the data," (emphasis added) by which use of the word "some" it may be inferred that the `charset` it is _not_ used with some (other) media types.  Thus including a `charset` parameter always regardless of Media Type, and in particular when that type is not a subtype of `text`, is inconsistent with that specification.

Moreover, [section 3.7](http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7) reads, "[t]he presence or absence of a parameter might be significant to the processing of a media-type, depending on its definition within the media type registry."  In reference to the process by which parameter-use definition is included in that registry, the HTTP specification explicitly identifies as authoritative the process outlined in [RFC 1590, "Media Type Registration Procedure"](http://www.ietf.org/rfc/rfc1590.txt).  Appendix A of that document reads:

> In general, parameters in the content-type header field are used to convey supplemental information for various content types, and their use is defined when the content-type and subtype are defined.

If parameter use must be defined for a given Media Type, then it is inappropriate for Spray to include a parameter for an unrecognized Media Type as it does in the example given above.  If a Media Type is unrecognized then the existence and content of use-definitions for its parameters, including `charset`, are not known, thus including a `charset`--or any other--parameter may at worst be contradictory to defined use of the given Media Type and might cause unintended, unpredictable and incorrect behavior in a client application.

sirthias:
Closes #16803.

So far our media-type model has been a bit blurry with regard to a media-types requirements on character encoding. This PR rectifies that and fixes #16803 in the process.

/cc @jrudolph

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/model/MediaType.scala
+++ b/akka-http-core/src/main/scala/akka/http/model/MediaType.scala
@@ -78,7 +78,7 @@ object MediaRange {
     override def isMultipart = mainType == "multipart"
     override def isText = mainType == "text"
     override def isVideo = mainType == "video"
-    def specimen = MediaType.custom(mainType, "custom")
+    def specimen = MediaType.custom(mainType, "custom", MediaType.Encoding.Binary)
   }
 
   def custom(mainType: String, params: Map[String, String] = Map.empty, qValue: Float = 1.0f): MediaRange = {
@@ -165,7 +165,7 @@ object MediaRanges extends ObjectRegistry[String, MediaRange] {
 sealed abstract case class MediaType private[http] (value: String)(val mainType: String,
                                                                    val subType: String,
                                                                    val compressible: Boolean,
-                                                                   val binary: Boolean,
+                                                                   val encoding: MediaType.Encoding,
                                                                    val fileExtensions: immutable.Seq[String],
                                                                    val params: Map[String, String])
   extends japi.MediaType with LazyValueBytesRenderable with WithQValue[MediaRange] {
@@ -191,7 +191,7 @@ sealed abstract case class MediaType private[http] (value: String)(val mainType:
 }
 
 class MultipartMediaType private[http] (_value: String, _subType: String, _params: Map[String, String])
-  extends MediaType(_value)("multipart", _subType, compressible = true, binary = true, Nil, _params) {
+  extends MediaType(_value)("multipart", _subType, compressible = true, encoding = MediaType.Encoding.Open, Nil, _params) {
   override def isMultipart = true
   def withBoundary(boundary: String): MultipartMediaType = withParams {
     if (boundary.isEmpty) params - "boundary" else params.updated("boundary", boundary)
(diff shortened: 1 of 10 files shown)


akka-http-core/src/main/scala/akka/http/model/MediaType.scala
```diff
@@ -200,28 +200,47 @@ class MultipartMediaType private[http] (_value: String, _subType: String, _param
...
+  }
+
   /**
    * Create a custom media type.
    */
-  def custom(mainType: String, subType: String, compressible: Boolean = false, binary: Boolean = false,
-             fileExtensions: immutable.Seq[String] = Nil, params: Map[String, String] = Map.empty,
-             allowArbitrarySubtypes: Boolean = false): MediaType = {
+  def custom(mainType: String, subType: String, compressible: Boolean = false,
+             encoding: MediaType.Encoding = Encoding.Binary, fileExtensions: immutable.Seq[String] = Nil,
```
jrudolph: This changes the default from the former `binary = false`. Was this intended?

akka-http-core/src/main/scala/akka/http/model/MediaType.scala
```diff
@@ -253,33 +274,34 @@ object MediaTypes extends ObjectRegistry[(String, String), MediaType] {
...
-  private def img(subType: String, compressible: Boolean, binary: Boolean, fileExtensions: String*) = register {
-    new NonMultipartMediaType("image", subType, compressible, binary, immutable.Seq(fileExtensions: _*)) {
+  private def img(subType: String, compressible: Boolean, encoding: Encoding, fileExtensions: String*) = register {
+    new NonMultipartMediaType("image", subType, compressible, encoding, immutable.Seq(fileExtensions: _*)) {
       override def isImage = true
     }
   }
   private def msg(subType: String, fileExtensions: String*) = register {
-    new NonMultipartMediaType("message", subType, compressible = true, binary = false, immutable.Seq(fileExtensions: _*)) {
+    new NonMultipartMediaType("message", subType, compressible = true, encoding = Encoding.Fixed(HttpCharsets.`UTF-8`),
```
jrudolph: Is assuming a fixed charset of UTF-8 spec'd somewhere for `msg/` media types? Otherwise, this may be overly strict.

jrudolph:
LGTM otherwise. Nice.

akka-http-core/src/main/scala/akka/http/model/MediaType.scala
sirthias: Yes, that's intended. We cannot use `Fixed` as a default (too restrictive) and `Open` is a bit too relaxed as a default for my taste. So `Binary` appeared to be the best choice.
Another option would be to force the user to supply the encoding by not having a default at all.
Would that be the better choice?

akka-http-core/src/main/scala/akka/http/model/MediaType.scala
sirthias: AFAICS the old RFC 2046 is the relevant here and defines content to be generally US-ASCII encoded.
Since UTF8 is a super-set of US-ASCII I went for UTF8 here, since it appears to be the best default choice nowadays.

akka-http-core/src/main/scala/akka/http/model/MediaType.scala
jrudolph: @sirthias thanks for the explanation. I guess it's only a default value so people can still change it easily for their own custom media-types. My slight preference would probably be to remove the default value but it's a tough call.

akka-http-core/src/main/scala/akka/http/model/MediaType.scala
jrudolph: Hmm, I'm not so sure about this. From RFC 2046:

>    No encoding other than "7bit", "8bit", or "binary" is permitted for
   the body of a "message/rfc822" entity.  The message header fields are
   always US-ASCII in any case, and data within the body can still be
   encoded, in which case the Content-Transfer-Encoding header field in
   the encapsulated message will reflect this.

In any case, it seems there's no charset parameter defined for any of those media types, so maybe binary would be the most general to use here.

akka-http-core/src/main/scala/akka/http/model/MediaType.scala
sirthias: Yeah, I agree. People should make a conscious decision here.
I'll remove the default.

akka-http-core/src/main/scala/akka/http/model/MediaType.scala
sirthias: Ok, I'll switch to `Binary`.

ktoso:
LGTM

sirthias:
Applied changes discussed with @jrudolph.

jrudolph:
LGTM

Issue #17090, opened 2015-03-30
add AsyncStage


rkuhn:
should allow receiving input “from the side” in order to express TimerTransform, mapAsync etc.

PR #17105, opened 2015-03-30
add async stage

rkuhn:
not yet rebased and squashed (and commit-commented properly), will do after review


akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMapAsyncSpec.scala
```diff
@@ -17,8 +17,51 @@ import akka.testkit.TestProbe
...
+      val future = f(elem)
+      val cb = ctx.getAsyncCallback()
+      future.onComplete(cb.invoke)
+      ctx.holdUpstream()
+    }
+
+    override def onPull(ctx: AsyncContext[Out, Try[Out]]) =
+      if (elemInFlight != null) {
+        val e = elemInFlight
+        elemInFlight = null.asInstanceOf[Out]
```
drewhk: ah, I didn't know this.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMapAsyncSpec.scala
```diff
@@ -17,8 +17,51 @@ import akka.testkit.TestProbe
...
+
+    override def onUpstreamFinish(ctx: AsyncContext[Out, Try[Out]]) =
+      if (ctx.isHoldingUpstream) ctx.absorbTermination()
+      else ctx.finish()
+
+    private def pushIt(elem: Out, ctx: AsyncContext[Out, Try[Out]]) =
+      if (ctx.isFinishing) ctx.pushAndFinish(elem)
+      else ctx.pushAndPull(elem)
+  }
+}
```
drewhk: nice one!

akka-stream/src/main/scala/akka/stream/FlowMaterializer.scala
```diff
@@ -32,3 +32,12 @@ abstract class FlowMaterializer {
   def executionContext: ExecutionContextExecutor
 
 }
+
+object NoFlowMaterializer extends FlowMaterializer {
```
drewhk: internal api? or is this useful for users?

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -9,6 +9,7 @@ import scala.util.control.NonFatal
 import akka.stream.stage._
 import akka.stream.Supervision
 import akka.stream.impl.ReactiveStreamsCompliance
+import akka.stream.FlowMaterializer
```
drewhk: I don't like this dependency. Can't this be a type parameter instead?

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -238,22 +245,64 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]], val forkLimit:
...
+      currentOp.precedingWasPull = true
+      currentOp.holdingUpstream = false
       state = Pulling
       null
     }
 
+    override def getAsyncCallback(): AsyncCallback[Any] = {
+      val current = currentOp.asInstanceOf[AsyncStage[Any, Any, Any]]
+      val context = current.context // avoid concurrent access (to avoid @volatile)
+      new AsyncCallback[Any] {
```
drewhk: is this cacheable?

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -282,18 +339,62 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]], val forkLimit:
...
-    override def hold(): FreeDirective = {
-      if (currentOp.holding) throw new IllegalStateException("Cannot hold while already holding")
-      currentOp.holding = true
+    override def holdUpstream(): UpstreamDirective = {
+      currentOp.precedingWasPull = false
+      currentOp.holdingUpstream = true
+      exit()
+    }
+
+    override def holdUpstreamAndPush(elem: Any): UpstreamDirective = {
```
drewhk: what is the use case for these holdXstreamAndY methods?

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -282,18 +339,62 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]], val forkLimit:
...
+    override def isHoldingUpstream: Boolean = currentOp.holdingUpstream
+
+    override def holdDownstream(): DownstreamDirective = {
+      currentOp.holdingDownstream = true
       exit()
     }
 
-    override def isHolding: Boolean = currentOp.holding
+    override def holdDownstreamAndPull(): DownstreamDirective = {
+      if (comingFromUpstream)
```
drewhk: I have an idea here. What if we encode the holding states as a bitfield instead of separate booleans, say holdingBalls, the current direction also as a bitfield (with exactly one bit set), say directionBall, and the reqirements of the directive as directiveNeedsBalls. Then we can use this:

    if ((holdingBalls | directionBall) & directiveNeedsBalls == directiveNeedsBalls)
      // do the action
    else
      calculateNiceException(holdingBalls, directionBall, directiveNeedsBalls)

That way the calculation looks simple and needs less branches, while the error message generation can be centralized and we do not care if it is fast or not.

drewhk: It might be even possible to pull it up to the parent class.

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -513,67 +642,94 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]], val forkLimit:
...
+    override def comingFromSide = true
 
-          override def hold(): FreeDirective = {
-            activeOpIndex = entryPoint
-            super.hold()
-            execute()
-            null
-          }
+    override def push(elem: Any): DownstreamDirective = {
+      ReactiveStreamsCompliance.requireNonNullElement(elem)
```
drewhk: already checked by super.push

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -513,67 +642,94 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]], val forkLimit:
...
 
-          override def pushAndPull(elem: Any): FreeDirective = {
-            ReactiveStreamsCompliance.requireNonNullElement(elem)
-            activeOpIndex = entryPoint
-            super.pushAndPull(elem)
-            execute()
-            null
-          }
+    override def pull(): UpstreamDirective = {
+      activeOpIndex = entryPoint
```
drewhk: hmm, can't this pattern be expressed in a more compact way? Or it will allocate closures unnecessarily?

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
```diff
@@ -455,19 +514,64 @@ sealed trait Context[Out] {
...
   def pushAndPull(elem: Out): FreeDirective
 
 }
 
 /**
+ * An asynchronous callback holder that is attached to an [[AsyncContext]].
+ * Invoking [[AsyncCallback#invoke]] will eventually lead to [[AsyncStage#onAsyncInput]]
+ * being called.
+ */
+trait AsyncCallback[T] {
```
drewhk: is there a reason not being a Function1?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -334,3 +339,85 @@ private[akka] final case class Expand[In, Out, Seed](seed: In ⇒ Seed, extrapol
...
+  }
+
+  override def onPush(elem: In, ctx: AsyncContext[Out, Notification]) = {
+    val future =
+      try f(elem)
+      catch {
+        case NonFatal(ex) if decider(ex) != Supervision.Stop ⇒ null // resume processing
+      }
+    if (future ne null) {
+      val idx = elemsInFlight.enqueue(null)
```
drewhk: this reads weird, but I see what it does. Maybe it would help if you would preallocate a dummy Try[Out] (obviously a dummy failure will do), naming it WaitingSlot or something and using it instead of null. That would make the intent more clear.

rkuhn: good point

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -334,3 +339,85 @@ private[akka] final case class Expand[In, Out, Seed](seed: In ⇒ Seed, extrapol
...
+    }
+    if (elemsInFlight.isFull) ctx.holdUpstream()
+    else ctx.pull()
+  }
+
+  override def onPull(ctx: AsyncContext[Out, (Int, Try[Out])]) = {
+    @tailrec def rec(hasConsumed: Boolean): DownstreamDirective =
+      if (elemsInFlight.isEmpty && ctx.isFinishing) ctx.finish()
+      else if (elemsInFlight.isEmpty || elemsInFlight.peek == null) {
+        if (hasConsumed && ctx.isHoldingUpstream) ctx.holdDownstreamAndPull()
```
drewhk: what is the purpose of hasConsumed?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -334,3 +339,85 @@ private[akka] final case class Expand[In, Out, Seed](seed: In ⇒ Seed, extrapol
...
+        else ctx.holdDownstream()
+      } else elemsInFlight.dequeue() match {
+        case Failure(ex) ⇒ rec(true)
```
drewhk: Shouldn't this fail the stage?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -334,3 +339,85 @@ private[akka] final case class Expand[In, Out, Seed](seed: In ⇒ Seed, extrapol
   final override def restart(): Expand[In, Out, Seed] =
     throw new UnsupportedOperationException("Expand doesn't support restart")
 }
+
+/**
+ * INTERNAL API
+ */
+private[akka] final case class MapAsync[In, Out](parallelism: Int, f: In ⇒ Future[Out], decider: Supervision.Decider)
```
drewhk: I think this class can be refactored to be more readable by concentrating on domain events and actions instead of the raw events and actions. For me the fundamental events are:

 - the queue becomes full or non-full (does not matter if the slots are pending or completed)
 - the front slot in the queue becomes completed
 - upstream finish or downstream cancel
 - failure

and the corresponding actions are:

 - enqueue and launch a Future
 - complete a slot
 - dequeue a completed slot

I think a refactoring around these would improve the readability.

akka-stream/src/main/scala/akka/stream/FlowMaterializer.scala
rkuhn: good point

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
rkuhn: We have too many type parameters already: what else but a FlowMaterializer should it be?

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
rkuhn: Yes, I also do that in the MapAsync implementation. Should be documented.

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
rkuhn: They properly model things that you can do: ball coming from downstream, we’re already holding the upstream ball, as a result we hold downstream and pull from upstream. How else would you express this?

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
rkuhn: good point!

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
drewhk: Interpreter does not have type parameters. I like the fact that the interpreter can be ripped out and used with minimal dependencies. I just try to maintain this "simple core" design.

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
drewhk: but is there a case where you need this?

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
drewhk: feel free to change the names, they are just arbitrary for demonstrating the idea.

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
rkuhn: yes, otherwise I wouldn’t have added them, see the MapAsync implementation

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
rkuhn: That is the _raison d’être_ for `NoFlowMaterializer`.

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
drewhk: ok, I believe you. Hope we can clear these up by using bitmasks.

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
drewhk: That is not the same, because now interpreter relies on the FlowMaterializer interface. I like a more layered approach, even if this is not a separate subproject.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMapAsyncSpec.scala
```diff
@@ -17,8 +17,51 @@ import akka.testkit.TestProbe
 import akka.stream.scaladsl.OperationAttributes.supervisionStrategy
 import akka.stream.Supervision.resumingDecider
 import akka.stream.impl.ReactiveStreamsCompliance
+import scala.util.Try
+import scala.concurrent.ExecutionContext
+import scala.util.Failure
+import scala.util.Success
+import org.scalactic.ConversionCheckedTripleEquals
```
ktoso: :+1: :-)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMapAsyncSpec.scala
ktoso: I like it as well (callback.invoke => onAsyncInput makes complete sense)

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
rkuhn: I don’t know what you mean by “closures”, but yes, this could be implemented differently by restructuring how the boundaries enter their stage/context. I’ll try it out.

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
rkuhn: Yes, Function1 has a ton of methods that this one does not need. Plus: Java API :-)

drewhk: Ok, fine by me. As for the Java API, this is not created by user code anyway, right? Only the call to invoke that matters.

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
drewhk: I assumed some surround(block) solution (block being a function), but if you have better means, do it :)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
rkuhn: whether or not we have previously freed up a slot in the buffer: in that case we need to push the upstream ball to get more inputs

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
rkuhn: apparently not, this was fixed by the existing test suite: the meaning of Supervision.Resume or Supervision.Restart is that the failed result is just ignored. Real failures will not be enqueued into the buffer.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: but that you can always check by buffer.nonFull

rkuhn: hmm, will look into refactoring that one

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
rkuhn: Exactly: Java users would have no clue what it means to “apply” a callback.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: Ah ok, so if it ends up being here it means that supervision has already been invoked

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: I summarized some ideas in the comment at the class header.

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
drewhk: ok, convinced :)

drewhk:
@jrudolph @sirthias I guess this new stage can be very useful for you, too.

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
```diff
@@ -241,30 +241,30 @@ private[akka] object ActorProcessorFactory {
...
-      case Take(n, _)                 ⇒ (ActorInterpreter.props(settings, List(fusing.Take(n))), ())
-      case Collect(pf, _)             ⇒ (ActorInterpreter.props(settings, List(fusing.Collect(settings.supervisionDecider)(pf))), ())
-      case Scan(z, f, _)              ⇒ (ActorInterpreter.props(settings, List(fusing.Scan(z, f, settings.supervisionDecider))), ())
-      case Expand(s, f, _)            ⇒ (ActorInterpreter.props(settings, List(fusing.Expand(s, f))), ())
-      case Conflate(s, f, _)          ⇒ (ActorInterpreter.props(settings, List(fusing.Conflate(s, f, settings.supervisionDecider))), ())
-      case Buffer(n, s, _)            ⇒ (ActorInterpreter.props(settings, List(fusing.Buffer(n, s))), ())
-      case MapConcat(f, _)            ⇒ (ActorInterpreter.props(settings, List(fusing.MapConcat(f, settings.supervisionDecider))), ())
-      case MapAsync(f, _)             ⇒ (MapAsyncProcessorImpl.props(settings, f), ())
+      case Identity(_)                ⇒ (ActorInterpreter.props(settings, List(fusing.Map({ x: Any ⇒ x }, settings.supervisionDecider)), materializer), ())
```
viktorklang: `{ x: Any ⇒ x }` should be hoisted, no reason to allocate a new one :)

drewhk: True. Fortunately we remove identity stages anyway :)

akka-stream/src/main/scala/akka/stream/impl/FixedSizeBuffer.scala
```diff
@@ -17,67 +20,64 @@ private[akka] object FixedSizeBuffer {
    *
    * Returns a specialized instance for power-of-two sized buffers.
    */
-  def apply(size: Int): FixedSizeBuffer =
+  def apply[T](size: Int): FixedSizeBuffer[T] =
```
viktorklang: Doesn't seem to validate 0/negative sizes?

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
```diff
@@ -28,9 +29,30 @@ import akka.stream.Supervision
...
+   */
+  private[stream] var holdingDownstream = false
+  /**
+   * INTERNAL API
+   */
+  private[stream] var precedingWasPull = false
+  /**
+   * INTERNAL API
+   */
   private[stream] var terminationPending = false
```
viktorklang: Add a TODO to use bitflags to conserve memory and cache traffic?

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
viktorklang: You mean as a part of some other PR? :)

akka-stream/src/main/scala/akka/stream/impl/FixedSizeBuffer.scala
drewhk: not a user level API, but I guess it does not hurt to add.

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
drewhk: see https://github.com/akka/akka/pull/17105/files#r27479472

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
drewhk: No, I meant that we do not expose this AST node anywhere, and where we actually use them (empty flows), they are automatically removed after you add something meaningful before or after them. Still, preallocating the identity function is a good idea.

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
viktorklang: Is that the correct link?

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
viktorklang: When would we ever materialize an Identity-node?

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
drewhk: no, it is unbounded recursion now :) There is a comment above somewhere where we discuss using bitfields, not because of memory but because it allows cheaper tests where we actually use these.

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
drewhk: I think we never actually materialize them, they are just placeholders. We can add an exception here and see what breaks :) (We will still need the AST node though)

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
viktorklang: So a win-win thing (memory size and access cost)

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
viktorklang: I think that sounds excellent :)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -334,3 +339,85 @@ private[akka] final case class Expand[In, Out, Seed](seed: In ⇒ Seed, extrapol
...
+      if (elemsInFlight.isEmpty && ctx.isFinishing) ctx.finish()
+      else if (elemsInFlight.isEmpty || elemsInFlight.peek == null) ctx.ignore()
+      else elemsInFlight.dequeue() match {
+        case Failure(ex) ⇒ rec(true)
+        case Success(elem) ⇒
+          if (ctx.isHoldingUpstream) ctx.pushAndPull(elem)
+          else ctx.push(elem)
+      }
+
+    input._2 match {
```
patriknw: perhaps use `ínput match` instead of `_2` and `_1`?

patriknw:
looking good, very useful

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMapAsyncSpec.scala
hepin1989: ne

rkuhn: @hepin1989 If you mean two lines up, then `ne` and `!=` are absolutely identical when involving `null` (well, almost: sometimes the compiler seems to emit `ifnull` and sometimes `ifnonnull` bytecodes, but that should not matter).

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
rkuhn: The extra parameter would have to be on all Stages, which is a no-go: this is an implementation detail that has no business in the user API. The alternative is to provide the “current” FlowMaterializer in a ThreadLocal, but that is too high a price in terms of ugliness.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
rkuhn: hmm, to me that felt repetitive, but I’ll try it on for size

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
rkuhn: I thought long and hard about this, but was unable to come up with an improvement over the status quo. You are welcome to try your hand on it after it has been merged.

A few background thoughts:

  * queue becoming full is trivially modeled by `holdUpstream`
  * queue becoming non-full requires different actions based on whether at the same time an element was dequeued or not
  * unification of the two complex tailrec expressions is impossible due to type constraints

In general layering another abstraction over the “raw” one always included a large amount of extraneous complexity and boilerplate code.

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
hepin1989: @rkuhn yes ,for the ```if (elemInFlight != null)  ```,could not click the line on phone.I know the result is identical,but you are compare the reference,maybe I have been followed the book too much:P.

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
drewhk: This is still a layer violation that I don't like. This would prevent ripping out this class from Akka and reusing elsewhere, for example if a very lightweight stream support is needed somewhere without full akka dependencies. IteratorInterpreter is a good example since NoMaterializer seems more like a hack than a nice solution. I would rather take an Any, no matter how ugly it is, and cast it down at the use site, than to have this extra dependency.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: Ok

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -282,18 +319,48 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]], val forkLimit:
...
-      currentOp.holding = true
+    override def holdUpstream(): UpstreamDirective = {
+      removeBits(PrecedingWasPull)
+      addBits(UpstreamBall)
       exit()
     }
 
-    override def isHolding: Boolean = currentOp.holding
+    override def holdUpstreamAndPush(elem: Any): UpstreamDirective = {
+      ReactiveStreamsCompliance.requireNonNullElement(elem)
```
drewhk: I am also not convinced of this one. This is not an RS stream, this is an unnecessary assumption. We can throw on null element but does it have anything to do with RS?

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -282,18 +319,48 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]], val forkLimit:
...
+      if (incomingBall != UpstreamBall)
+        throw new IllegalStateException("can only holdUpstreamAndPush from onPush")
+      mustHave(DownstreamBall)
```
drewhk: BothBalls

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -282,18 +319,48 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]], val forkLimit:
...
+
+    override def holdDownstream(): DownstreamDirective = {
+      addBits(DownstreamBall)
+      exit()
+    }
+
+    override def holdDownstreamAndPull(): DownstreamDirective = {
+      if (incomingBall != DownstreamBall)
+        throw new IllegalStateException("can only holdDownstreamAndPull from onPull")
+      mustHave(UpstreamBall)
```
drewhk: BothBalls

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -282,18 +319,48 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]], val forkLimit:
...
+      null
+    }
+
+    override def isHoldingDownstream: Boolean = (currentOp.bits & DownstreamBall) != 0
 
     override def pushAndPull(elem: Any): FreeDirective = {
       ReactiveStreamsCompliance.requireNonNullElement(elem)
-      if (!currentOp.holding) throw new IllegalStateException("Cannot pushAndPull without holding first")
-      currentOp.holding = false
+      mustHave(DownstreamBall | UpstreamBall)
```
drewhk: BothBalls?

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -301,21 +368,24 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]], val forkLimit:
 
     override def absorbTermination(): TerminationDirective = {
       updateJumpBacks(activeOpIndex)
-      currentOp.holding = false
+      removeBits(DownstreamBall | UpstreamBall)
```
drewhk: BothBalls?

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
```diff
@@ -27,10 +28,68 @@ import akka.stream.Supervision
...
+  private[stream] def enterAndPush(elem: Out): Unit = {
+    context.enter()
+    context.push(elem)
+    context.execute()
+  }
+
+  /**
+   * INTERNAL API
+   */
+  private[stream] def enterAndPull(): Unit = {
```
drewhk: can these be pushed down to subclasses that actually use them? Only BoundaryStage and AsyncStage uses these.

rkuhn: then I’d need to repeat them

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
rkuhn: yes, here it is right

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
rkuhn: nope: see two lines above

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
rkuhn: Since fusing of operations is an optimization, we should not change semantics based on whether or not it has occurred: pushing an element downstream should be considered async / RS in all cases.

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
rkuhn: okay, I see your point: with the new constant there would be no overhead to doing so

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
drewhk: ok

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
drewhk: > pushing an element downstream should be considered async / RS in all cases.

This is the source of disagreement then, I like this engine to be reusable in other ways and not depending on Akka or RS features.

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
drewhk: exactly. I should have elaborated that the above line is correct, but I like how this mustHave documents the requirements exactly (even if one of them is trivially met at the current point)

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
rkuhn: The real conflict then arises not from the engine itself but from the usage of `Stage` and `Context` for the user-facing API (in the context of a Reactive Streams compliant implementation) as well as the internal implementation. At this point I see no reason to complicate matters for our own use, it should be rather simple to disentangle (but quite a bit of extra code) once we need these to be independent.

drewhk: Ok. That might be not that far ahead :)

rkuhn:
Why TF did this compile in Eclipse _AND_ sbt?

rkuhn:
I consider this now good enough for internal usage, but before properly documenting it for external users we should gain more experience, replace all TimerTransform usage and quite possibly also add inputBunch / outputBunch like abstractions so that we can formulate all stream-of-streams stages on this basis.

rkuhn:
fixes #17090

drewhk:
YAAAY!

PR #17111, opened 2015-03-31
=htp #15792 make `entityAs` directive turn `IllegalArgumentException` into `ValidationRejection`

### LINKED ISSUE #15792: Provide more fine-grained rejections for distinguishing between malformed and invalid data in unmarshalling
@2beaucoup at #15718:

> Would be nice if we could distinguish between malformed and invalid (string instead of int, failed  require) rejections. Seems useful for unmarshalling to case classes would provide more accurate error messages.

sirthias:
Closes #15792.

/cc @jrudolph


jrudolph:
LGTM

drewhk:
LGTM

2beaucoup:
:+1:

PR #17112, opened 2015-03-31
=htc #17110 disallow zero body parts in MIME content

### LINKED ISSUE #17110: Akka MIME unmarshallers neither accept nor properly reject MIME with LF line endings
The Akka MIME unmarshallers do not accept (nor properly reject) MIME with LF line endings.

I know that unix line endings are not valid MIME according to the RFC, but Spray 1.3 used to accept this kind of MIME (via the `org.jvnet.mimepull` library), and Postel's law would seem to apply.

Currently, if you pass Akka a MIME message with LF line endings, it parses successfully as a MIME bundle with zero BodyParts. I think that it should either be an error as the RFC states "[The body must then contain **one or more** body parts](http://www.w3.org/Protocols/rfc1341/7_2_Multipart.html)", or (much better) it should parse the slightly broken MIME.

Here’s a failing test demonstrating the issue. The equivalent test in Spray 1.3.1 passes:

```
import akka.actor.ActorSystem
import akka.http.marshalling.Marshal
import akka.http.model.ContentTypes._
import akka.http.model.{HttpEntity, Multipart}
import akka.http.model.Multipart.FormData
import akka.http.unmarshalling.Unmarshal
import akka.util.ByteString
import akka.http.marshalling.MultipartMarshallers._
import scala.concurrent.{Future, Await}

class MimeLineEndingsSpec
  extends Specification
  with NoTimeConversions {

  implicit val fm = ActorFlowMaterializer()
  val timeout = 1.second
  private def await[T](future: Future[T]): T =
    Await.result(future, timeout)

  "MIME unmarshallers" should {
    "accept LF endings as well as CRLF" in {
      val unmarshalledMime = Multipart.FormData(Source(List(
        FormData.BodyPart("first",
          HttpEntity.Strict(`text/plain`, ByteString("first part"))),
        FormData.BodyPart("second",
          HttpEntity.Strict(`text/plain`, ByteString("second part"))))))

      val marshalledMime = await(
        Marshal(unmarshalledMime).to[HttpEntity].flatMap(_.toStrict(timeout)))

      println(marshalledMime.data.utf8String)

      val mimeWithUnixLineEndings = marshalledMime.copy(
        data = ByteString(marshalledMime.data.utf8String.replace("\r\n", "\n")))

      val unmarshalledMimeWithUnixLineEndings = await(
        Unmarshal(mimeWithUnixLineEndings).to[Multipart.FormData])

      // unmarshalledMimeWithUnixLineEndings here is 'FormData.Strict(Vector())'
      // i.e. the whole body is treated as a MIME preamble and ignored.
      //
      // Shouldn't we get an error like "no MIME boundary found"?
      // Or better (and like Spray 1.3), shouldn't we be able to parse LF MIME?
      unmarshalledMimeWithUnixLineEndings mustEqual unmarshalledMime
    }
  }
}
```

Thanks,

Rich

sirthias:
Closes #17110.

/cc @jrudolph @RichardBradley

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/engine/parsing/BodyPartParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/engine/parsing/BodyPartParser.scala
@@ -56,7 +56,6 @@ private[http] final class BodyPartParser(defaultContentType: ContentType,
 
   private[this] var output = collection.immutable.Queue.empty[Output] // FIXME this probably is too wasteful
   private[this] var state: ByteString ⇒ StateResult = tryParseInitialBoundary
-  private[this] var receivedInitialBoundary = false
   private[this] var terminated = false
 
   def warnOnIllegalHeader(errorInfo: ErrorInfo): Unit =
@@ -79,10 +78,8 @@ private[http] final class BodyPartParser(defaultContentType: ContentType,
     if (output.nonEmpty)
       ctx.push(dequeue())
     else if (ctx.isFinishing) {
-      if (terminated || !receivedInitialBoundary)
-        ctx.finish()
-      else
-        ctx.pushAndFinish(ParseError(ErrorInfo("Unexpected end of multipart entity")))
+      if (terminated) ctx.finish()
+      else ctx.pushAndFinish(ParseError(ErrorInfo("Unexpected end of multipart entity")))
     } else
       ctx.pull()
   }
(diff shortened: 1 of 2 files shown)


jrudolph:
LGTM

drewhk:
LGTM

PR #17121, opened 2015-04-01
=htp #17051 fix BodyPartParser not handling streaming entities correctly

### LINKED ISSUE #17051: Multipart MIME parser may swallow MIME errors depending on stream chunk size
I have found that the `akka.http.unmarshalling.MultipartUnmarshallers` will behave quite differently with invalid MIME streams depending on how the stream has been divided into chunks.

I have uploaded a failing test to https://gist.github.com/RichardBradley/0e73adaa86c4d377bf65/95ff8af7298ad64808700fee59e736a4c1e10031

(Run it with `sbt test`)

I have picked out one example: when a MIME body part is missing its headers, the MIME parser will throw a fairly helpful error if the request is `Strict`:
```
akka.http.model.ParsingException: Illegal character ' ' in header name
```

However, if the same MIME stream is `Default` or `Chunked`, then the error will not be thrown. Instead the BodyPart stream will be truncated (without error) at varying lengths depending on the exact chunk size used.

sirthias:
Closes #17051.

/cc @jrudolph

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/engine/parsing/BodyPartParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/engine/parsing/BodyPartParser.scala
@@ -61,18 +61,19 @@ private[http] final class BodyPartParser(defaultContentType: ContentType,
   def warnOnIllegalHeader(errorInfo: ErrorInfo): Unit =
     if (illegalHeaderWarnings) log.warning(errorInfo.withSummaryPrepended("Illegal multipart header").formatPretty)
 
-  override def onPush(input: ByteString, ctx: Context[Output]): SyncDirective = {
-    try state(input)
-    catch {
-      case e: ParsingException ⇒ fail(e.info)
-      case NotEnoughDataException ⇒
-        // we are missing a try/catch{continue} wrapper somewhere
-        throw new IllegalStateException("unexpected NotEnoughDataException", NotEnoughDataException)
-    }
-    if (output.nonEmpty) ctx.push(dequeue())
-    else if (!terminated) ctx.pull()
-    else ctx.finish()
-  }
+  override def onPush(input: ByteString, ctx: Context[Output]): SyncDirective =
+    if (!terminated) {
+      try state(input)
+      catch {
+        case e: ParsingException ⇒ fail(e.info)
+        case NotEnoughDataException ⇒
+          // we are missing a try/catch{continue} wrapper somewhere
+          throw new IllegalStateException("unexpected NotEnoughDataException", NotEnoughDataException)
+      }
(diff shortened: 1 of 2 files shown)


jrudolph:
Only looked at the tests but those looked reasonable. LGTM.

akka-http-core/src/main/scala/akka/http/engine/parsing/BodyPartParser.scala
```diff
@@ -115,9 +116,8 @@ private[http] final class BodyPartParser(defaultContentType: ContentType,
       }
       rec(offset)
     } catch {
-      case NotEnoughDataException ⇒ continue(input.takeRight(needle.length), 0)(parsePreamble)
+      case NotEnoughDataException ⇒ continue(input.takeRight(needle.length + 2), 0)(parsePreamble)
```
rkuhn: all of these eta expansions perform one allocation per call, it should be way more efficient to store the function in a `val` instead

rkuhn:
I don’t fully understand what was wrong before, but the tests look reasonable.

akka-http-core/src/main/scala/akka/http/engine/parsing/BodyPartParser.scala
sirthias: These lines are only executed on "parsing boundaries" where an intermediate chunk of data coming in from the TCP layer ends. So, things are actually somewhat self-adjusting. If the system becomes more loaded the chunks become bigger (as reading from the socket buffer become less frequent).

Also, it is doubtful that *always* allocating an extra function instance even if the preamble fits in a single chunk (or isn't even present, as in the majority of cases) vs. allocating one on every chunk boundary does indeed provide any benefit.

rkuhn: Okay, thanks for the extra detail, sounds reasonable.

sirthias:
Rebased onto latest release-2.3-dev.
Test failure appears to be unrelated.

rkuhn:
PLS BUILD

rkuhn:
(the mapAsyncUnordered bug has been fixed in #17195)

sirthias:
Apparently the build failed due to `unresolved dependency: org.scala-lang#scala-library;2.10.5: not found`.

rkuhn:
yes, very weird; I waited until the next one passed that stage, let’s try again

rkuhn:
PLS BUILD

rkuhn:
hmm, something fishy on Jenkins, looking into it

rkuhn:
PLS BUILD

rkuhn:
Yeehah! :-)

Issue #17080, opened 2015-03-26
Move socket TestUtils to akka-testkit


patriknw:
We have created several copies of it because of issues with dependencies.

PR #17124, opened 2015-04-01
=tes #17080 Move TestUtils to testkit.SocketUtil

patriknw:
Refs #17080

Good to have in next 2.3.x release, so we can make use of it in streams


akka-actor-tests/src/test/scala/akka/actor/ActorPerfSpec.scala
```diff
@@ -104,7 +104,8 @@ class ActorPerfSpec extends AkkaSpec("akka.actor.serialize-messages = off") with
     System.gc()
     val memAfter = memMx.getHeapMemoryUsage
     driver ! PoisonPill
-    TestUtils.verifyActorTermination(driver, 15 seconds)
```
drewhk: why removing this helper method?

patriknw: we already have `expectTerminated`

ktoso:
LGTM

PR #17125, opened 2015-04-01
=act #15914 Clarify and correct max-received-message-size

patriknw:
Refs #15914

### DIFF
--- a/akka-actor/src/main/resources/reference.conf
+++ b/akka-actor/src/main/resources/reference.conf
@@ -557,6 +557,9 @@ akka {
       # The maximum number of bytes delivered by a `Received` message. Before
       # more data is read from the network the connection actor will try to
       # do other work.
+      # The purpose of this setting is to impose a smaller limit than the 
+      # configured receive buffer size. When using value 'unlimited' it will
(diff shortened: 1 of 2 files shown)


akka-actor/src/main/resources/reference.conf
```diff
@@ -557,6 +557,9 @@ akka {
       # The maximum number of bytes delivered by a `Received` message. Before
       # more data is read from the network the connection actor will try to
       # do other work.
+      # The purpose of this setting is to impose a smaller limit than the 
+      # configured receive buffer size. When using value 'unlimited' it will
+      # try to read all from the receive buffer.
       max-received-message-size = unlimited
```
drewhk: should this remain unlimited by default?

2m:
LGTM

ktoso:
LGTM

akka-actor/src/main/resources/reference.conf
rkuhn: yes, I’d think so: while this is `unlimited` the true limit is controlled in exactly one place—preferable to having two places to modify

rkuhn:
LGTM

PR #17126, opened 2015-04-01
Upgrade HTTP layer APIs to new BidiFlow design

sirthias:
plus a few smaller tweaks.

/cc @jrudolph

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/Http.scala
@@ -5,8 +5,6 @@
 package akka.http
 
 import java.net.InetSocketAddress
-import akka.http.engine.server.HttpServer.HttpServerPorts
-import akka.stream.Graph
 import com.typesafe.config.Config
 import scala.collection.immutable
 import scala.concurrent.Future
@@ -15,16 +13,23 @@ import akka.util.ByteString
 import akka.io.Inet
 import akka.stream.FlowMaterializer
 import akka.stream.scaladsl._
-import akka.http.engine.client.{ HttpClient, ClientConnectionSettings }
-import akka.http.engine.server.{ HttpServer, ServerSettings }
-import akka.http.model.{ HttpResponse, HttpRequest }
+import akka.http.engine.client._
+import akka.http.engine.server._
+import akka.http.model.{ HttpHeader, HttpResponse, HttpRequest }
 import akka.actor._
 
 class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.Extension {
   import Http._
 
   /**
-   * Creates a [[ServerBinding]] instance which represents a prospective HTTP server binding on the given `endpoint`.
+   * Creates a [[Source]] of [[IncomingConnection]] instances which represents a prospective HTTP server binding
+   * on the given `endpoint`.
+   * If the given port is 0 the resulting source can be materialized several times. Each materialization will
+   * then be assigned a new local port by the operating system, which can then be retrieved by the materialized
+   * [[ServerBinding]].
+   * If the given port is non-zero subsequent materialization attempts of the produced source will immediately
+   * fail, unless the first materialization has already been unbound. Unbinding can be triggered via the materialized
+   * [[ServerBinding]].
    */
   def bind(interface: String, port: Int = 80, backlog: Int = 100,
            options: immutable.Traversable[Inet.SocketOption] = Nil,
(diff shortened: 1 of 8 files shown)


akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -99,27 +99,34 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
     bindAndHandle(Flow[HttpRequest].mapAsync(handler), interface, port, backlog, options, settings, log)
 
   /**
-   * Transforms a given HTTP-level server [[Flow]] into a lower-level TCP transport flow.
+   * The type of the server-side HTTP layer as a stand-alone BidiStage
+   * that can be put atop the TCP layer to form an HTTP server.
+   *
+   * {{{
+   *                +------+
+   * HttpResponse ~>|      |~> ByteString
```
drewhk: What should be the convention? Lower layer to the left or right? I used lower layer to the left before, but anything is OK, but it would be nice to establish a convention now.

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -194,7 +200,7 @@ object Http extends ExtensionId[HttpExt] with ExtensionIdProvider {
   case class IncomingConnection(
     localAddress: InetSocketAddress,
     remoteAddress: InetSocketAddress,
-    flow: Flow[HttpResponse, HttpRequest, Unit]) {
+    flow: Flow[HttpResponse, HttpRequest, Any]) {
```
drewhk: thanks!

akka-http-core/src/test/scala/akka/http/engine/client/OutgoingConnectionSpec.scala
```diff
@@ -359,13 +359,13 @@ class HttpClientSpec extends AkkaSpec("akka.loggers = []\n akka.loglevel = OFF")
...
 
-      FlowGraph.closed(HttpClient.clientBlueprint(remoteAddress, settings, NoLogging)) { implicit b ⇒
+      FlowGraph.closed(OutgoingConnectionBlueprint(remoteAddress, settings, NoLogging)) { implicit b ⇒
         client ⇒
           import FlowGraph.Implicits._
-          Source(netIn) ~> client.bytesIn
-          client.bytesOut ~> Sink(netOut)
-          Source(requests) ~> client.httpRequests
-          client.httpResponses ~> Sink(responses)
+          Source(netIn) ~> client.in2
```
drewhk: hm, maybe we can use a nicer API here. Maybe something like (Sink, Source) <~> client <~> (Sink, Source), or alternatively Flow(sink, source) <~> client <~> Flow(sink, source)

akka-stream/src/main/boilerplate/akka/stream/FanInShape.scala.template
```diff
@@ -57,6 +57,21 @@ class UniformFanInShape[T, O](val n: Int, _init: FanInShape.Init[O]) extends Fan
   def in(n: Int): Inlet[T] = inArray(n)
 }
 
+class FanInShape1N[T0, T1, O](val n: Int, _init: FanInShape.Init[O]) extends FanInShape[O](_init) {
```
drewhk: So this is an almost uniform fan-in shape with one extra typed port?

drewhk:
Nice!

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: This convention already appears to be established.
The scaladoc to `BidiFlow::atop`, for example, says,:

> By convention protocol stacks are growing to the left: the right most is the bottom layer, the closest to the metal.

akka-http-core/src/test/scala/akka/http/engine/client/OutgoingConnectionSpec.scala
sirthias: +1

akka-stream/src/main/boilerplate/akka/stream/FanInShape.scala.template
sirthias: Right. I had a need for this in some intermediate version which has been superseded since then.
But since I already wrote this shape I left it in in case you guys deem it of general utility.

akka-http-core/src/main/scala/akka/http/Http.scala
drewhk: ok, I used the reverse convention, but I started before BidiFlow was available :)

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -33,28 +38,22 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
     val endpoint = new InetSocketAddress(interface, port)
     val effectiveSettings = ServerSettings(settings)
 
-    val connections: Source[StreamTcp.IncomingConnection, Future[StreamTcp.ServerBinding]] = StreamTcp().bind(endpoint, backlog, options, effectiveSettings.timeouts.idleTimeout)
-    val serverBlueprint: Graph[HttpServerPorts, Unit] = HttpServer.serverBlueprint(effectiveSettings, log)
+    val connections: Source[StreamTcp.IncomingConnection, Future[StreamTcp.ServerBinding]] =
+      StreamTcp().bind(endpoint, backlog, options, effectiveSettings.timeouts.idleTimeout)
+    val layer = serverLayer(effectiveSettings, log)
```
jrudolph: By moving this line into the `connections.map` we could also fix #17129 for now.

sirthias: Yes, but since #16168 has been closed yesterday I'd rather go for the proper solution now.

rkuhn:
this is subsumed in #17133, right?

drewhk:
yes, it looks like. Can we close this?

sirthias:
Yes, this can be closed if its ok to merge the two somewhat unrelated things via the joint PR #17133.

sirthias:
Subsumed in #17133.

Issue #16737, opened 2015-01-28
NullPointerException in LocalSnapshotStore  


raboof:
While running our tests, I sometimes (but rarely) see the following exception:

```
2015-01-28 13:11:02,081 ERROR [OneForOneStrategy > akka://rfs-rebb-ohm-api-UnparsableContentRoutesSpec/system/snapshot-store]: null
java.lang.NullPointerException: null
        at scala.collection.mutable.ArrayOps$ofRef$.length$extension(ArrayOps.scala:192) ~[scala-library-2.11.5.jar:0.13.5]
        at scala.collection.mutable.ArrayOps$ofRef.length(ArrayOps.scala:192) ~[scala-library-2.11.5.jar:0.13.5]
        at scala.collection.SeqLike$class.size(SeqLike.scala:106) ~[scala-library-2.11.5.jar:0.13.5]
        at scala.collection.mutable.ArrayOps$ofRef.size(ArrayOps.scala:186) ~[scala-library-2.11.5.jar:0.13.5]
        at scala.collection.mutable.Builder$class.sizeHint(Builder.scala:69) ~[scala-library-2.11.5.jar:0.13.5]
        at scala.collection.mutable.ArrayBuilder.sizeHint(ArrayBuilder.scala:22) ~[scala-library-2.11.5.jar:0.13.5]
        at scala.collection.TraversableLike$class.builder$1(TraversableLike.scala:241) ~[scala-library-2.11.5.jar:0.13.5]
        at scala.collection.TraversableLike$class.map(TraversableLike.scala:244) ~[scala-library-2.11.5.jar:0.13.5]
        at scala.collection.mutable.ArrayOps$ofRef.map(ArrayOps.scala:186) ~[scala-library-2.11.5.jar:0.13.5]
        at akka.persistence.snapshot.local.LocalSnapshotStore.snapshotMetadata(LocalSnapshotStore.scala:108) ~[akka-persistence-experimental_2.11-2.3.9.jar:na]
        at akka.persistence.snapshot.local.LocalSnapshotStore.loadAsync(LocalSnapshotStore.scala:47) ~[akka-persistence-experimental_2.11-2.3.9.jar:na]
        at akka.persistence.snapshot.SnapshotStore$$anonfun$receive$1.applyOrElse(SnapshotStore.scala:27) ~[akka-persistence-experimental_2.11-2.3.9.jar:na]
[...]

PR #17127, opened 2015-04-01
=per #16737 avoid NPE when snapshot dir is removed

patriknw:
- also best effort to recreate the directory if it is removed

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala
@@ -28,7 +28,7 @@ private[persistence] class LocalSnapshotStore extends SnapshotStore with ActorLo
 
   private val config = context.system.settings.config.getConfig("akka.persistence.snapshot-store.local")
   private val streamDispatcher = context.system.dispatchers.lookup(config.getString("stream-dispatcher"))
-  private val snapshotDir = new File(config.getString("dir"))
+  private val dir = new File(config.getString("dir"))
 
   private val serializationExtension = SerializationExtension(context.system)
   private var saving = immutable.Set.empty[SnapshotMetadata] // saving in progress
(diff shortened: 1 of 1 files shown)


patriknw:
Refs #16737

drewhk:
LGTM

akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala
```diff
@@ -104,19 +104,27 @@ private[persistence] class LocalSnapshotStore extends SnapshotStore with ActorLo
   private def snapshotFile(metadata: SnapshotMetadata, extension: String = ""): File =
     new File(snapshotDir, s"snapshot-${URLEncoder.encode(metadata.persistenceId, "UTF-8")}-${metadata.sequenceNr}-${metadata.timestamp}${extension}")
 
-  private def snapshotMetadata(persistenceId: String, criteria: SnapshotSelectionCriteria): immutable.Seq[SnapshotMetadata] =
-    snapshotDir.listFiles(new SnapshotFilenameFilter(persistenceId)).map(_.getName).collect {
+  private def snapshotMetadata(persistenceId: String, criteria: SnapshotSelectionCriteria): immutable.Seq[SnapshotMetadata] = {
+    val files = snapshotDir.listFiles(new SnapshotFilenameFilter(persistenceId))
+    if (files eq null) Nil // if the dir was removed
```
2m: This only defers NPE to [here](https://github.com/akka/akka/blob/wip-16737-NPE-LocalSnapshotStore-patriknw/akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala#L47) or [here](https://github.com/akka/akka/blob/wip-16737-NPE-LocalSnapshotStore-patriknw/akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala#L66). Am I missing something?

patriknw: wdym? it's `Nil`, not `null`

akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala
```diff
@@ -104,19 +104,27 @@ private[persistence] class LocalSnapshotStore extends SnapshotStore with ActorLo
...
+  }
 
   override def preStart() {
-    if (!snapshotDir.isDirectory) {
+    snapshotDir()
+    super.preStart()
+  }
+
+  private def snapshotDir(): File = {
+    if (!dir.isDirectory) {
```
hepin1989: if it's not a dir,why not delete it and then create it as dir？

akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala
2m: Wow. Sorry. I was missing something. :D

akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala
hepin1989: a better way i think would be
if(dir.exists){
if it's file
   delete it and then create as dir
}else create as dir

patriknw: It might be an important file, misconfigured, we can't just delete it.
Note that this will fail in preStart and the actor will not even start.

hepin1989: yeah,missed that.

ktoso:
LGTM

PR #17130, opened 2015-04-02
=pro,ben Added benchmark for actor tell and updated sbt-jmh to 0.1.14

bantonsson:


ktoso:
Interesting... what results are you getting for this benchmark @bantonsson?
When I run it I got things in ms range which seems rather weird (or I'm missing something?)

bantonsson:
@ktoso Yes, millis is the right range. Since I don't want to print 1 million invocations (iterations = 1 million, batchSize = 1), and I don't want to try my luck with loops, then the test has a batchSize of 1 million, and the time is per op, which seems to correspond to iteration (i.e. batchSize calls of the method) in SingleShot.

I can't say that I understand all the nuances of JMH, but that's how I've interpreted things. Here are my results.

```
Result: 12827.834 ±(99.9%) 501.952 us/op [Average]
  Statistics: (min, avg, max) = (11970.200, 12827.834, 16915.457), stdev = 1013.969
  Confidence interval (99.9%): [12325.881, 13329.786]
  Samples, N = 50
        mean =  12827.834 ±(99.9%) 501.952 us/op
         min =  11970.200 us/op
  p( 0.0000) =  11970.200 us/op
  p(50.0000) =  12494.284 us/op
  p(90.0000) =  14134.745 us/op
  p(95.0000) =  15699.270 us/op
  p(99.0000) =  16915.457 us/op
  p(99.9000) =  16915.457 us/op
  p(99.9900) =  16915.457 us/op
  p(99.9990) =  16915.457 us/op
  p(99.9999) =  16915.457 us/op
         max =  16915.457 us/op


# Run complete. Total time: 00:00:52

Benchmark               Mode  Cnt      Score     Error  Units
TellOnlyBenchmark.tell    ss   50  12827.834 ± 501.952  us/op
```

ktoso:
Ah I see, thanks for explaining! I have not used `batchSize` and didn't realise it'll get reported like that.
Makes sense for what we want to look at here, esp. that singleshot gives a histogram. :-)

Just in case you didn't find it yet, there's a useful annotation for cases when you know how many ops per benchmark method execution there are and you want it's avg:

```scala
@Benchmark
  @OutputTimeUnit(TimeUnit.MICROSECONDS)
  @Warmup(iterations = 10, batchSize = TellOnlyBenchmark.numMessages)
  @Measurement(iterations = 10, batchSize = TellOnlyBenchmark.numMessages)
  /* I want avg time of one op, I know there's numMessages ops for each tell() */
  @BenchmarkMode(Array(Mode.AverageTime))
  @OperationsPerInvocation(TellOnlyBenchmark.numMessages)
  def tell() = probe.send(actor, message)
```

project/plugins.sbt
```diff
@@ -17,7 +17,7 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-pgp" % "0.8.3")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-s3" % "0.5")
 
-addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.1.6")
+addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.1.12")
```
ktoso: @bantonsson just released sbt-jmh `0.1.13` with updated JMH version to `1.7.1` (the 1.12 shipped with 1.6.x).

(after holidays of course :-))

ktoso: Actually, just bumped to `0.1.14` as today's `1.8` JMH release has an interesting new profiler (not for this benchmark, but in general interesting): http://mail.openjdk.java.net/pipermail/jmh-dev/2015-April/001785.html (gc churn)

bantonsson: Thanks, updating.

bantonsson:
@ktoso I know about that annotation, but all benchmark modes except SingleShotTime  are time based, and I desperately want to avoid the GC pauses.

ktoso:
:+1:

project/AkkaBuild.scala
```diff
@@ -192,7 +192,7 @@ object AkkaBuild extends Build {
     dependencies = Seq(actor, stream, persistence, testkit).map(_ % "compile;compile->test"),
     settings = defaultSettings ++ Seq(
       libraryDependencies ++= Dependencies.testkit
-    ) ++ settings ++ jmhSettings
+    ) ++ settings ++ formatSettings ++ jmhSettings
```
drewhk: nice!

akka-bench-jmh/src/main/scala/akka/actor/TellOnlyBenchmark.scala
```diff
@@ -0,0 +1,151 @@
...
+  class Echo extends Actor {
+    def receive = {
+      case s @ `stop` ⇒
+        context stop self
+      case m ⇒ sender ! m
+    }
+  }
+
+  class DroppingMessageQueue extends UnboundedMailbox.MessageQueue {
+    @volatile var dropping = false
```
drewhk: why do we need dropping?

bantonsson: I want to measure the cost of the `tell` code without the mailbox implementation, and I also don't want the other actor to wake up and skew the measurement.

drewhk: Ok.

ktoso:
Excellent, thanks for bumping the jmh dependency as well. :-)

Issue #16887, opened 2015-02-17
Websockets Server side API and implementation discussion


jrudolph:
**To signal your approval or support for websockets please go to #16848, this ticket is reserved for technical discussion.**

Specs: http://tools.ietf.org/html/rfc6455
Chapter from "High Performance Browser Networking": http://chimera.labs.oreilly.com/books/1230000000545/ch17.html

/cc @sirthias

jrudolph:
First server-side API design preview at #16891.

sirthias:
I'd say the first goal should be to come up with a complete basic implementation (without implemented extensions) that passes the Autobahn testsuite: http://autobahn.ws/testsuite/

sirthias:
Also interesting: Netty's implementation here: https://github.com/netty/netty/tree/master/codec-http/src/main/java/io/netty/handler/codec/http/websocketx

sirthias:
Maybe the single most important extensions that might be in scope once we have the basic implementation done:
Compression: https://tools.ietf.org/html/draft-ietf-hybi-permessage-compression-19

jrudolph:
Here's a diagram of how the server pipeline would look like with Websocket included:

![server-pipeline-with-websocket dot](https://cloud.githubusercontent.com/assets/9868/6318150/617226a2-ba81-11e4-9b33-b0f98c7483ec.png)

Color code:
 - blue: HTTP
 - red: Websocket
 - green: user-supplied

dnrusakov:
Also interesting: https://github.com/wandoulabs/spray-websocket
Pass all Autobahn test cases (both ws and wss)

drewhk:
Looking at the graph picture you posted, we need to make a little tool that extracts a StreamLayout in a format that graphviz can use it. It would be very useful to have automated architecture docs, also possibly be able to link from the graph elements to the scaladoc or the code itself (just like scaladoc inheritance diagrams).

drewhk:
@jrudolph with github.com/akka/akka/pull/17300 merged I can close this?

jrudolph:
Yep.

jrudolph:
Merged in #17300 and released with RC1.

PR #17131, opened 2015-04-02
Preview 2 of server-side websocket support

jrudolph:
This is the next preview of the server-side websocket implementation #16887. The high-level API actually works now.

Since the last preview:
- the low-level frame model has been revamped (though, it will need another overhaul to be able to create Strict messages)
- the high-level message API actually works
- lots of tests for the intermediate layers
- the integration with the HttpServer has been upgraded to the latest streams API

Missing:
- more tests, especially integration tests
- proper and tested fail and close behaviors
- support for per-message-compression

At this point I'm mostly interested in comments about the API (what you see in TestServer, Message, UpgradeToWebsocket) and the integration with the existing http infrastructure (HttpServer). The internals are still subject to change.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/engine/rendering/HttpResponseRendererFactory.scala
+++ b/akka-http-core/src/main/scala/akka/http/engine/rendering/HttpResponseRendererFactory.scala
@@ -4,6 +4,8 @@
 
 package akka.http.engine.rendering
 
+import akka.http.engine.ws.{ WebsocketSwitch, UpgradeToWebsocketResponseHeader, Handshake }
+
 import scala.annotation.tailrec
 import akka.event.LoggingAdapter
 import akka.util.ByteString
@@ -21,7 +23,8 @@ import headers._
  */
 private[http] class HttpResponseRendererFactory(serverHeader: Option[headers.Server],
                                                 responseHeaderSizeHint: Int,
-                                                log: LoggingAdapter) {
+                                                log: LoggingAdapter,
+                                                websocketSwitch: Option[WebsocketSwitch] = None) {
 
   private val renderDefaultServerHeader: Rendering ⇒ Unit =
     serverHeader match {
@@ -149,6 +152,11 @@ private[http] class HttpResponseRendererFactory(serverHeader: Option[headers.Ser
 
             if (renderConnectionHeader)
               r ~~ Connection ~~ (if (close) CloseBytes else KeepAliveBytes) ~~ CrLf
+            else if (connHeader != null && connHeader.hasUpgrade && websocketSwitch.isDefined) {
+              r ~~ connHeader ~~ CrLf
+              val websocketHeader = headers.collectFirst { case u: UpgradeToWebsocketResponseHeader ⇒ u }
+              websocketHeader.foreach(header ⇒ websocketSwitch.get.switchToWebsocket(header.handlerFlow)(header.mat))
+            }
             if (mustRenderTransferEncodingChunkedHeader && !transferEncodingSeen)
               r ~~ `Transfer-Encoding` ~~ ChunkedBytes ~~ CrLf
         }
--- a/akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
@@ -4,6 +4,13 @@
 
 package akka.http.engine.server
 
+import akka.stream.scaladsl.FlexiMerge.{ ReadAny, MergeLogic }
+import akka.stream.scaladsl._
+
+import akka.http.engine.ws._
+import akka.stream.scaladsl.FlexiRoute.{ DemandFrom, RouteLogic }
+import org.reactivestreams.{ Subscriber, Publisher }
+
 import scala.util.control.NonFatal
 import akka.util.ByteString
 import akka.event.LoggingAdapter
@@ -18,6 +25,7 @@ import akka.http.engine.TokenSourceActor
 import akka.http.model._
 import akka.http.util._
 import ParserOutput._
+import akka.http.engine.ws.Websocket.{ SwitchToWebsocketToken }
 
 /**
  * INTERNAL API
@@ -37,7 +45,8 @@ private[http] object HttpServerBluePrint {
           logParsingError(info withSummaryPrepended "Illegal request header", log, parserSettings.errorLoggingVerbosity)
       })
 
-    val responseRendererFactory = new HttpResponseRendererFactory(serverHeader, responseHeaderSizeHint, log)
+    val ws = websocketPipeline
+    val responseRendererFactory = new HttpResponseRendererFactory(serverHeader, responseHeaderSizeHint, log, Some(ws))
 
     @volatile var oneHundredContinueRef: Option[ActorRef] = None // FIXME: unnecessary after fixing #16168
     val oneHundredContinueSource = Source.actorPublisher[OneHundredContinue.type] {
@@ -94,17 +103,43 @@ private[http] object HttpServerBluePrint {
         val bypassOneHundredContinueInput = bypassMerge.in1
         val bypassApplicationInput = bypassMerge.in2
 
+        // HTTP pipeline
         requestParsing.outlet ~> bypassFanout.in
         bypassMerge.out ~> renderer.inlet
         val requestsIn = (bypassFanout.out(0) ~> requestPreparation).outlet
 
         bypassFanout.out(1) ~> bypass ~> bypassInput
         oneHundredContinueSource ~> bypassOneHundredContinueInput
+        val http = FlowShape(requestParsing.inlet, renderer.outlet)
+
+        // Websocket pipeline
+        val websocket = b.add(ws.flow)
+
+        // protocol routing
+        val protocolRouter = b.add(new WebsocketSwitchRouter())
+        val protocolMerge = b.add(new WebsocketMerge)
+
+        protocolRouter.out0 ~> http ~> protocolMerge.in0
+        protocolRouter.out1 ~> websocket ~> protocolMerge.in1
+
+        // protocol switching
+        val wsSwitchTokenMerge = b.add(new StreamUtils.EagerCloseMerge2[AnyRef]("protocolSwitchWsTokenMerge"))
+        val switchTokenBroadcast = b.add(Broadcast[SwitchToWebsocketToken.type](2))
+        ws.switchSource ~> switchTokenBroadcast.in
+        switchTokenBroadcast.out(0) ~> wsSwitchTokenMerge.in1
+        wsSwitchTokenMerge.out /*~> printEvent[AnyRef]("netIn")*/ ~> protocolRouter.in
+        switchTokenBroadcast.out(1) ~> protocolMerge.in2
+        val netIn = wsSwitchTokenMerge.in0
+
+        val netOutPrint = b.add( /*printEvent[ByteString]("netOut")*/ Flow[ByteString])
+        protocolMerge.out ~> netOutPrint.inlet
+        val netOut = netOutPrint.outlet
 
         BidiShape[HttpResponse, ByteString, ByteString, HttpRequest](
           bypassApplicationInput,
-          renderer.outlet,
-          requestParsing.inlet,
+          netOut,
+          netIn,
+
           requestsIn)
     }
   }
@@ -238,6 +273,109 @@ private[http] object HttpServerBluePrint {
             closeRequested = true)
           ctx.absorbTermination()
         case _ ⇒ ctx.fail(error)
+      }
+  }
+
+  case class WebsocketSetup(
+    flow: Flow[ByteString, ByteString, Any],
+    publisherKey: StreamUtils.ReadableCell[Publisher[FrameEvent]],
+    subscriberKey: StreamUtils.ReadableCell[Subscriber[FrameEvent]],
+    switchSource: Source[SwitchToWebsocketToken.type, Any]) extends WebsocketSwitch {
+    @volatile var switchToWebsocketRef: Option[ActorRef] = None
+
+    def switchToWebsocket(handlerFlow: Flow[FrameEvent, FrameEvent, Any])(implicit mat: FlowMaterializer): Unit = {
+      // 1. fill processing hole in the websocket pipeline with user-provided handler
+      Source(publisherKey.value)
+        .via(handlerFlow)
+        .to(Sink(subscriberKey.value))
+        .run()
+
+      // FIXME: is there a race condition between 1 and 2?
+
+      // 2. flip the switch
+      switchToWebsocketRef.get ! TokenSourceActor.Trigger
+    }
+  }
+  def websocketPipeline: WebsocketSetup = {
+    val sinkCell = new StreamUtils.OneTimeWriteCell[Publisher[FrameEvent]]
+    val sourceCell = new StreamUtils.OneTimeWriteCell[Subscriber[FrameEvent]]
+
+    val sink = StreamUtils.oneTimePublisherSink[FrameEvent](sinkCell, "frameHandler.in")
+    val source = StreamUtils.oneTimeSubscriberSource[FrameEvent](sourceCell, "frameHandler.out")
+
+    val flow =
+      Flow[ByteString]
+        .transform[FrameEvent](() ⇒ new FrameEventParser)
+        .via(Flow.wrap(sink, source)((_, _) ⇒ ()))
+        .transform(() ⇒ new FrameEventRenderer)
+
+    lazy val setup = WebsocketSetup(flow, sinkCell, sourceCell, switchToWebsocketSource)
+    lazy val switchToWebsocketSource: Source[SwitchToWebsocketToken.type, ActorRef] =
+      Source.actorPublisher[SwitchToWebsocketToken.type] {
+        Props {
+          val actor = new TokenSourceActor(SwitchToWebsocketToken)
+          setup.switchToWebsocketRef = Some(actor.context.self)
+          actor
+        }
+      }
+    setup
+  }
+  class WebsocketSwitchRouter
+    extends FlexiRoute[AnyRef, FanOutShape2[AnyRef, ByteString, ByteString]](new FanOutShape2("websocketSplit"), OperationAttributes.name("websocketSplit")) {
+
+    override def createRouteLogic(shape: FanOutShape2[AnyRef, ByteString, ByteString]): RouteLogic[AnyRef] =
+      new RouteLogic[AnyRef] {
+        def initialState: State[_] = http
+
+        def http: State[_] = State[Any](DemandFrom(shape.out0)) { (ctx, _, element) ⇒
+          element match {
+            case b: ByteString ⇒
+              // route to HTTP processing
+              ctx.emit(shape.out0)(b)
+              SameState
+
+            case SwitchToWebsocketToken ⇒
+              // switch to websocket protocol
+              websockets
+          }
+        }
+        def websockets: State[_] = State[Any](DemandFrom(shape.out1)) { (ctx, _, element) ⇒
+          // route to Websocket processing
+          ctx.emit(shape.out1)(element.asInstanceOf[ByteString])
+          SameState
+        }
+      }
+  }
+  class WebsocketMerge extends FlexiMerge[ByteString, FanInShape3[ByteString, ByteString, SwitchToWebsocketToken.type, ByteString]](new FanInShape3("websocketMerge"), OperationAttributes.name("websocketMerge")) {
+    def createMergeLogic(s: FanInShape3[ByteString, ByteString, SwitchToWebsocketToken.type, ByteString]): MergeLogic[ByteString] =
+      new MergeLogic[ByteString] {
+        def httpIn = s.in0
+        def wsIn = s.in1
+        def tokenIn = s.in2
+
+        def initialState: State[_] = http
+
+        def http: State[_] = State[AnyRef](ReadAny(httpIn.asInstanceOf[Inlet[AnyRef]], tokenIn.asInstanceOf[Inlet[AnyRef]])) { (ctx, in, element) ⇒
+          element match {
+            case b: ByteString ⇒
+              ctx.emit(b); SameState
+            case SwitchToWebsocketToken ⇒
+              ctx.changeCompletionHandling(closeWhenInCloses(wsIn))
+              websockets
+          }
+        }
+        def websockets: State[_] = State[ByteString](ReadAny(httpIn /* otherwise we won't read the websocket upgrade response */ , wsIn)) { (ctx, _, element) ⇒
+          ctx.emit(element)
+          SameState
+        }
+
+        def closeWhenInCloses(in: Inlet[_]): CompletionHandling =
+          defaultCompletionHandling.copy(onUpstreamFinish = { (ctx, closingIn) ⇒
+            if (closingIn == in) ctx.finish()
+            SameState
+          })
+
+        override def initialCompletionHandling: CompletionHandling = closeWhenInCloses(httpIn)
       }
   }
 }
--- a/akka-http-core/src/main/scala/akka/http/engine/ws/FrameEventParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/engine/ws/FrameEventParser.scala
@@ -1,0 +1,161 @@
+/*
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.engine.ws
+
+import akka.http.util.{ ByteReader, ByteStringParserStage }
+import akka.stream.stage.{ StageState, SyncDirective, Context }
+import akka.util.ByteString
+
+import scala.annotation.tailrec
+
+/**
+ * Streaming parser for the Websocket framing protocol as defined in RFC6455
+ *
+ * http://tools.ietf.org/html/rfc6455
+ *
+ * 0                   1                   2                   3
+ * 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+ * +-+-+-+-+-------+-+-------------+-------------------------------+
+ * |F|R|R|R| opcode|M| Payload len |    Extended payload length    |
+ * |I|S|S|S|  (4)  |A|     (7)     |             (16/64)           |
+ * |N|V|V|V|       |S|             |   (if payload len==126/127)   |
+ * | |1|2|3|       |K|             |                               |
+ * +-+-+-+-+-------+-+-------------+ - - - - - - - - - - - - - - - +
+ * |     Extended payload length continued, if payload len == 127  |
+ * + - - - - - - - - - - - - - - - +-------------------------------+
+ * |                               |Masking-key, if MASK set to 1  |
+ * +-------------------------------+-------------------------------+
+ * | Masking-key (continued)       |          Payload Data         |
+ * +-------------------------------- - - - - - - - - - - - - - - - +
+ * :                     Payload Data continued ...                :
+ * + - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - +
+ * |                     Payload Data continued ...                |
+ * +---------------------------------------------------------------+
+ *
+ * INTERNAL API
+ */
+private[http] class FrameEventParser extends ByteStringParserStage[FrameEvent] {
+  protected def onTruncation(ctx: Context[FrameEvent]): SyncDirective =
+    ctx.fail(new ProtocolException("Data truncated"))
+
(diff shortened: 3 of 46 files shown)


akka-http-core/src/main/scala/akka/http/util/StreamUtils.scala
```diff
@@ -194,6 +197,125 @@ private[http] object StreamUtils {
...
+                  case Left(t1)  ⇒ ctx.emit(s.out0)(t1)
+                  case Right(t2) ⇒ ctx.emit(s.out1)(t2)
+                }
+                SameState
+              }
+          }
+      }
+
+    Flow.wrap {
+      FlowGraph.partial() { implicit b ⇒
```
jrudolph: Can be simplified with Flow.apply

akka-http-core/src/main/scala/akka/http/engine/rendering/HttpResponseRendererFactory.scala
```diff
@@ -149,6 +152,11 @@ private[http] class HttpResponseRendererFactory(serverHeader: Option[headers.Ser
 
             if (renderConnectionHeader)
               r ~~ Connection ~~ (if (close) CloseBytes else KeepAliveBytes) ~~ CrLf
+            else if (connHeader != null && connHeader.hasUpgrade && websocketSwitch.isDefined) {
+              r ~~ connHeader ~~ CrLf
+              val websocketHeader = headers.collect { case u: UpgradeToWebsocketResponseHeader ⇒ u }.headOption
```
2beaucoup: `collectFirst` or plain old `find`?

akka-http-core/src/main/scala/akka/http/engine/ws/Handshake.scala
```diff
@@ -0,0 +1,111 @@
...
+       9.   Optionally, a |Sec-WebSocket-Extensions| header field, with a
+            list of values indicating which extensions the client would like
+            to speak.  The interpretation of this header field is discussed
+            in Section 9.1.
+   */
+  def isWebsocketUpgrade(headers: List[HttpHeader], hostHeaderPresent: Boolean): Option[UpgradeToWebsocket] = {
+    def find[T <: HttpHeader: ClassTag]: Option[T] =
+      headers.collect {
+        case t: T ⇒ t
+      }.headOption
```
2beaucoup: here too (`collectFirst`)

akka-http-core/src/main/scala/akka/http/engine/rendering/HttpResponseRendererFactory.scala
jrudolph: Thanks, @2beaucoup. I'm forgetting about that one all the time :)

akka-http-core/src/main/scala/akka/http/engine/ws/FrameHandler.scala
```diff
@@ -0,0 +1,185 @@
...
+import java.nio.CharBuffer
+import java.nio.charset.{ Charset, CharsetDecoder }
+
+import akka.stream.scaladsl.Flow
+import akka.stream.stage.{ Directive, Context, StatefulStage }
+import akka.util.ByteString
+import FrameEvent.Opcode
+
+object FrameHandler {
+  private val UTF8 = Charset.forName("UTF-8")
```
2beaucoup: reuse `util.UTF8`?

akka-http-core/src/main/scala/akka/http/engine/rendering/HttpResponseRendererFactory.scala
2beaucoup: hehe ;)

akka-http-core/src/main/scala/akka/http/util/StreamUtils.scala
```diff
@@ -193,6 +196,125 @@ private[http] object StreamUtils {
...
+  /** A one time settable cell */
+  class OneTimeWriteCell[T <: AnyRef] extends ReadableCell[T] {
+    @volatile var _value: T = null.asInstanceOf[T]
+    def value: T = {
+      require(_value != null, "Value wasn't set yet")
+      _value
+    }
+
+    def set(value: T): Unit = {
+      require(_value == null, "Value can be only set once.") // one time usage only
```
drewhk: isn't a compareAndSet cleaner? Also, the error message should indicate that null values are not accepted.

akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
```diff
@@ -240,4 +266,75 @@ private[http] object HttpServerBluePrint {
...
+        .transform(() ⇒ new FrameEventRenderer)
+
+    lazy val setup = WebsocketSetup(flow, sinkCell, sourceCell, switchToWebsocketSource)
+    lazy val switchToWebsocketSource: Source[SwitchToWebsocketToken.type, ActorRef] =
+      Source.actorPublisher[SwitchToWebsocketToken.type] {
+        Props {
+          val actor = new TokenSourceActor(SwitchToWebsocketToken)
+          setup.switchToWebsocketRef = Some(actor.context.self)
+          actor
+        }
```
patriknw: Does this work with `akka.actor.serialize-creators=on`, otherwise move to `props` factory method in `TokenSourceActor` companion.

That usage of `actor.context.self` is "interesting". @rkuhn Is it safe before the actor is started?

akka-http-core/src/main/scala/akka/http/engine/ws/FrameEventParser.scala
```diff
@@ -0,0 +1,152 @@
...
+ * +-------------------------------- - - - - - - - - - - - - - - - +
+ * :                     Payload Data continued ...                :
+ * + - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - +
+ * |                     Payload Data continued ...                |
+ * +---------------------------------------------------------------+
+ *
+ */
+class FrameEventParser extends ByteStringParserStage[FramePart] {
+  protected def onTruncation(ctx: Context[FramePart]): SyncDirective =
+    ctx.fail(new RuntimeException("Data truncated"))
```
patriknw: should this be `RuntimeException`?

akka-http-core/src/main/scala/akka/http/engine/ws/FrameEventRenderer.scala
```diff
@@ -0,0 +1,116 @@
...
+
+import akka.util.ByteString
+import akka.stream.stage.{ TerminationDirective, StatefulStage, SyncDirective, Context }
+
+import scala.annotation.tailrec
+
+/**
+ * Todo: should we check all invariants here?
+ */
+class FrameEventRenderer extends StatefulStage[FramePart, ByteString] {
```
patriknw: does this (and other things) have the right visibility?

akka-http-core/src/main/scala/akka/http/engine/ws/FrameHandler.scala
```diff
@@ -0,0 +1,212 @@
...
+      def createMessagePart(data: ByteString, last: Boolean): MessageDataPart = BinaryMessagePart(data, last)
+    }
+    class CollectingTextMessage extends CollectingMessageFrame(Opcode.Text) {
+      var undecodedBytes = ByteString.empty
+      utf8decoder.reset()
+
+      def createMessagePart(data: ByteString, last: Boolean): MessageDataPart = TextMessagePart(decode(data), last)
+
+      def decode(bytes: ByteString, endOfInput: Boolean = false): String = {
+        println(s"Decoding undecoded: [${undecodedBytes.map(_ formatted "%02x").mkString(" ")}] new bytes: [${bytes.map(_ formatted "%02x").mkString(" ")}]")
```
patriknw: println?

akka-http-core/src/main/scala/akka/http/engine/ws/UpgradeToWebsocketsResponseHeader.scala
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.engine.ws
+
+import akka.http.model.headers.CustomHeader
+import akka.stream.FlowMaterializer
+import akka.stream.scaladsl.Flow
+
+private[akka] case class UpgradeToWebsocketResponseHeader(handlerFlow: Flow[FramePart, FramePart, Any])(implicit val mat: FlowMaterializer)
```
patriknw: INTERNAL API doc marker

akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
```diff
@@ -94,17 +101,36 @@ private[http] object HttpServerBluePrint {
...
         val bypassApplicationInput = bypassMerge.in2
 
+        // HTTP pipeline
         requestParsing.outlet ~> bypassFanout.in
         bypassMerge.out ~> renderer.inlet
         val requestsIn = (bypassFanout.out(0) ~> requestPreparation).outlet
 
         bypassFanout.out(1) ~> bypass ~> bypassInput
         oneHundredContinueSource ~> bypassOneHundredContinueInput
+        val http = FlowShape(requestParsing.inlet, renderer.outlet)
```
drewhk: nice one!

akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
```diff
@@ -94,17 +101,36 @@ private[http] object HttpServerBluePrint {
...
+
+        // Websocket pipeline
+        val websocket = b.add(ws.flow)
+
+        // protocol routing
+        val protocolRouter = b.add(new WebsocketSwitchRouter())
+        val protocolMerge = b.add(new StreamUtils.EagerCloseMerge2[ByteString]("protocolMerge"))
+
+        protocolRouter.out0 ~> http ~> protocolMerge.in0
+        protocolRouter.out1 ~> websocket ~> protocolMerge.in1
```
drewhk: you can use ws.flow here directly

akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
```diff
@@ -240,4 +266,75 @@ private[http] object HttpServerBluePrint {
...
+  case class WebsocketSetup(
+    flow: Flow[ByteString, ByteString, Any],
+    publisherKey: StreamUtils.ReadableCell[Publisher[FramePart]],
+    subscriberKey: StreamUtils.ReadableCell[Subscriber[FramePart]],
+    switchSource: Source[SwitchToWebsocketToken.type, Any]) extends WebsocketSwitch {
+    @volatile var switchToWebsocketRef: Option[ActorRef] = None
+
+    def switchToWebsocket(handlerFlow: Flow[FramePart, FramePart, Any])(implicit mat: FlowMaterializer): Unit = {
+      // 1. fill processing hole in the websocket pipeline with user-provided handler
+      Source(publisherKey.value)
```
drewhk: can't it be a flow instead, so you just do "ws join handlerFlow"?

akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
jrudolph: Yep, thanks.

akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
jrudolph: How would I create that flow and what would be the benefit?

akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
jrudolph: I just copied that from the existing, above TokenSourceActor usage. Like the other usage this is a hack to access the ActorRef.

akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
```diff
@@ -240,4 +266,75 @@ private[http] object HttpServerBluePrint {
...
+    val sinkCell = new StreamUtils.OneTimeWriteCell[Publisher[FramePart]]
+    val sourceCell = new StreamUtils.OneTimeWriteCell[Subscriber[FramePart]]
+
+    val sink = StreamUtils.oneTimePublisherSink[FramePart](sinkCell, "frameHandler.in")
+    val source = StreamUtils.oneTimeSubscriberSource[FramePart](sourceCell, "frameHandler.out")
+
+    val flow =
+      Flow[ByteString]
+        .transform[FramePart](() ⇒ new FrameEventParser)
+        .via(StreamUtils.sinkSourceFlow(sink, source))
```
drewhk: isn't this better as a BidiFlow? then you don't need the sinkCell and sourceCell, since you expose the "hole" as ports.

akka-http-core/src/main/scala/akka/http/engine/ws/FrameEventParser.scala
jrudolph: Yes, thanks, I will change that to ProtocolException which will also properly handled downstream.

akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
```diff
@@ -240,4 +266,75 @@ private[http] object HttpServerBluePrint {
         case _ ⇒ ctx.fail(error)
       }
   }
+
+  case class WebsocketSetup(
+    flow: Flow[ByteString, ByteString, Any],
```
drewhk: where is this Flow used?

akka-http-core/src/main/scala/akka/http/engine/ws/FrameEventRenderer.scala
jrudolph: Not yet, everything in the engine.ws package should be `private[http]`.

akka-http-core/src/main/scala/akka/http/util/StreamUtils.scala
jrudolph: Yes, it would also be more correct to actually deal with race conditions.

akka-http-core/src/main/scala/akka/http/engine/ws/FrameHandler.scala
jrudolph: Yes, I've still got lots of println and other debugging statements I didn't yet bother to remove. Will do.

akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
jrudolph: Above, where you already commented on `ws.flow`.

akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
jrudolph: This is related to the materialized value access in the graph: if you can manage to thread the materialized value through the graph in a way that I can get the publisher/subscriber pair (the hole) to the parser to be filled at runtime of the flow then this can change (and all the OneTime... stuff can go). I didn't manage to achieve that yet without having to restructure all kind of things of the whole pipeline, so I would like to postpone any such internal optimization to a later point.

akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
drewhk: ok thanks

akka-http-core/src/main/scala/akka/http/engine/rendering/HttpResponseRendererFactory.scala
```diff
@@ -149,6 +152,11 @@ private[http] class HttpResponseRendererFactory(serverHeader: Option[headers.Ser
...
+              websocketHeader.foreach(header ⇒ websocketSwitch.get.switchToWebsocket(header.handlerFlow)(header.mat))
```
drewhk: Is there a reason why this handlerFlow is attached dynamically from the header instead of taking it statically from the user?

akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
drewhk: I see, I will think about it. Is there a reason for filling in the hole dynamically though?

akka-http-core/src/main/scala/akka/http/engine/rendering/HttpResponseRendererFactory.scala
jrudolph: The reason is that it is embedded in the normal routing: you can decide whether you want to support websockets based on all the feature of an Http request, so the natural thing is leaving the decision to the user until the concrete request arrives.

jrudolph: Also, this is no different than the streaming TCP API: the engine gives you a flow representing the IncomingConnection and you can decide how to handle the data stream dynamically.

akka-http-core/src/main/scala/akka/http/engine/ws/Handshake.scala
```diff
@@ -0,0 +1,112 @@
...
+      connection.exists(_.hasUpgrade) &&
+      version.exists(_.hasVersion(CurrentWebsocketVersion)) &&
+      key.exists(k ⇒ isValidKey(k.key))) {
+
+      val header = new UpgradeToWebsocket {
+        def handleFrames(handlerFlow: Flow[FramePart, FramePart, Any])(implicit mat: FlowMaterializer): HttpResponse =
+          buildResponse(key.get, handlerFlow)
+      }
+      Some(header)
+    } else None
```
drewhk: maybe log if something looks like a WS request but fails to fully comply?

akka-http-core/src/main/scala/akka/http/engine/ws/Masking.scala
```diff
@@ -0,0 +1,65 @@
...
+object Masking {
+  def maskIf(condition: Boolean): Flow[FramePart, FramePart, Unit] =
+    if (condition) Flow[FramePart].transform(() ⇒ new Masking())
+    else Flow[FramePart]
+  def unmaskIf(condition: Boolean): Flow[FramePart, FramePart, Unit] =
+    if (condition) Flow[FramePart].transform(() ⇒ new Unmasking())
+    else Flow[FramePart]
+
+  class Masking extends Masker {
+    val random = new Random()
```
drewhk: ThreadLocalRandom.current()?

jrudolph: Yes, this is not yet finished. This is security-relevant (needs sufficient entropy by spec) and needs to be configurable and have a sane default.

akka-http-core/src/main/scala/akka/http/engine/ws/ProtocolException.scala
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.engine.ws
+
+case class ProtocolException(cause: String) extends RuntimeException(cause)
```
drewhk: do this and the class below need a separate file?

akka-http-core/src/main/scala/akka/http/engine/ws/Websocket.scala
```diff
@@ -0,0 +1,58 @@
...
+        .map(_.collect {
+          case m: MessageDataPart ⇒ m
+        })
+        .via(collectMessage)
+        .via(messageHandler)
+        .via(MessageToFrameRenderer.create(serverSide))
+
+    Flow[FramePart]
+      .via(printEvent("parserOut"))
+      .via(Masking.unmaskIf(serverSide))
```
drewhk: thankfully identity flows are now automatically removed, so this will just collapse if unmaskIf returns the empty flow

akka-http-core/src/test/scala/akka/http/TestServer.scala
```diff
@@ -18,18 +20,35 @@ object TestServer extends App {
...
 
-  val binding = Http().bindAndHandleSync({
-    case HttpRequest(GET, Uri.Path("/"), _, _, _)      ⇒ index
-    case HttpRequest(GET, Uri.Path("/ping"), _, _, _)  ⇒ HttpResponse(entity = "PONG!")
-    case HttpRequest(GET, Uri.Path("/crash"), _, _, _) ⇒ sys.error("BOOM!")
-    case _: HttpRequest                                ⇒ HttpResponse(404, entity = "Unknown resource!")
-  }, interface = "localhost", port = 8080)
+  try {
+    val binding = Http().bindAndHandleSync({
+      case req @ HttpRequest(GET, Uri.Path("/"), _, _, _) if req.header[UpgradeToWebsocket].isDefined ⇒
```
drewhk: Can't UpgradeToWebsocket be part of the unapply? Or that would just pollute the API for the non-WS case?

akka-http-core/src/main/scala/akka/http/util/StreamUtils.scala
```diff
@@ -193,6 +196,125 @@ private[http] object StreamUtils {
...
+      _value = value
+    }
+  }
+  /** Creates a flow out of sink and source */
+  def sinkSourceFlow[T, U](sink: Sink[T, Any], source: Source[U, Any]): Flow[T, U, Unit] =
```
rkuhn: please move to `akka-stream` as `Flow.apply` and close #17145 when done

akka-http-core/src/test/scala/akka/http/TestServer.scala
```diff
@@ -45,4 +64,16 @@ object TestServer extends App {
...
+
+  def echoWebsocketService: Flow[Message, Message, Unit] =
+    Flow[Message] // just let message flow directly to the output
+      .via(akka.http.util.printEvent("echoWebsocketService"))
+
+  def greeterWebsocketService: Flow[Message, Message, Unit] =
+    Flow[Message]
+      .collect {
+        case TextMessage.Strict(name)         ⇒ TextMessage.Strict(s"Hello '$name'")
+        case TextMessage.Streamed(nameStream) ⇒ TextMessage.Streamed(Source.single("Hello ") ++ nameStream mapMaterialized (_ ⇒ ()))
```
drewhk: the concat Mat type has been fixed now

akka-http-core/src/main/scala/akka/http/util/StreamUtils.scala
```diff
@@ -193,6 +196,125 @@ private[http] object StreamUtils {
...
+      FlowGraph.partial() { implicit b ⇒
+        val in = b.add(sink)
+        val out = b.add(source)
+
+        FlowShape[T, U](in, out)
+      }
+    }
+
+  /** A merge for two streams that just forwards all elements and closes the connection eagerly. */
+  class EagerCloseMerge2[T](name: String) extends FlexiMerge[T, FanInShape2[T, T, T]](new FanInShape2(name), OperationAttributes.name(name)) {
```
rkuhn: It is not yet clear to me why this is needed (and if it is then that would be a reason to just include this functionality as a config flag in the normal `Merge`).

From my understanding the WS stream can just be closed properly when the HTTP stream closes (e.g. by attaching an empty Publisher) and the HTTP handler stream can be closed when doing the WS upgrade. If these are true, then this merge variant would be unnecessary. The reason I’m arguing this point is because it somehow feels a little unclean to me, but I cannot quite put my finger onto the stain.

akka-http-core/src/test/scala/akka/http/engine/ws/FramingSpec.scala
```diff
@@ -0,0 +1,289 @@
...
+  import BitBuilder._
+
+  "The Websocket parser/renderer round-trip should work for" - {
+    "the frame header" - {
+      "interpret flags correctly" - {
+        "FIN" in {
+          b"""1000     # flags
+                  0000 # opcode
+              0        # mask?
+               0000000 # length
```
rkuhn: This looks very useful; @2m is that something for the testkit as well?

akka-http-core/src/test/scala/akka/http/TestServer.scala
jrudolph: In which way?

akka-http-core/src/test/scala/akka/http/TestServer.scala
jrudolph: Yes, I wouldn't pollute it on this low level. We can hide this in the (still missing) routing API.

akka-http-core/src/test/scala/akka/http/TestServer.scala
drewhk: I thought you are using mapMaterialized(_ ⇒ ()) because of the concat operator expecting Unit

jrudolph: No, it's because it's returning a tuple.

drewhk: OK

akka-http-core/src/main/scala/akka/http/util/StreamUtils.scala
jrudolph: Yes, I agree in both cases we should be able to close the respective side automatically. I'll try this.

akka-http-core/src/main/scala/akka/http/engine/server/HttpServerBluePrint.scala
rkuhn: This should work, but why not use `actor.self`? OTOH this whole dance should be avoidable by using an AsyncStage instead of the sinkSourceFlow where the materialized value allows sending in the real Flow to be materialized when required. The problem right now is that the AsyncStage would have to manually implement the Publisher and Subscriber interfaces because we have not yet ported the InputBunch / OutputBunch from ActorProcessor to the new scheme—the plan was to do that after 1.0 because it is an internal improvement.

drewhk:
I think this is good for a first round, and we should merge it after minor things has been fixed so that we can work more incrementally. The ugliest part right now is the dynamic registration of the user WS handler. What we currently lack in that regard is that we can only model static holes in graphs that has to be filled in before materialization. This API needs holes that can be filled after the stream is actually running. Currently the only way to do that is to revert to the use of Publisher/Subscriber interfaces.

jrudolph:
Thanks @drewhk. I'm currently fixing a few remaining bugs and then try to wrap up on a state which can be squashed into a first version to be merged.

akka-http-core/src/main/scala/akka/http/engine/ws/Websocket.scala
```diff
@@ -0,0 +1,58 @@
...
+      .via(FrameHandler.create(server = serverSide))
+      .via(printEvent("handlerOut"))
+      .viaMat(StreamUtils.routeEither(Flow[FramePart], userFlow))(Keep.right)
+      .via(Masking.maskIf(!serverSide))
+      .via(printEvent("maskOut"))
```
ktoso: `printEvent` will be possible to replace with the build-in `log(name)` after https://github.com/akka/akka/pull/17162 is merged

// ticket for this task: https://github.com/akka/akka/issues/17184

akka-http-core/src/test/scala/akka/http/engine/ws/BitBuilder.scala
```diff
@@ -0,0 +1,103 @@
...
+package akka.http.engine.ws
+
+import akka.parboiled2
+import parboiled2._
+import akka.util.ByteString
+
+import scala.annotation.tailrec
+import scala.util.{ Try, Failure, Success }
+
+object BitBuilder {
```
ktoso: awesome idea, love it.

akka-http-core/src/test/scala/akka/http/engine/ws/FramingSpec.scala
2m: Yes. I think it fits in the testkit. Useful when writing/testing stages/bidiflows which work with binary.

jrudolph:
Short status update: it turned out that the Java Utf8 decoder isn't as strict as the websocket protocol demands. For that reason other implementations (netty, spray-websockets) needed to rely on a special Utf8Validator. We now have something similar: a complete UTF8 to/from UTF-16 (= Java String) encoder and decoder that is strict enough to make most autobahn tests pass. What's still missing from the internals is getting all the closing scenarios completely right.

rkuhn:
Thanks for the update!

jrudolph:
This is now feature complete and can be reviewed. There are a view more aesthetic issues I'll will fix tomorrow and I'll read through it myself to see if everything makes sense.

jrudolph:
Superseded by #17248

Issue #16932, opened 2015-02-26
Clean up Http server and Client graph wirings in code and use BidiStage


drewhk:
Currently it does not use every piece of the new DSL niceness, because it was introduced late in the recent rewrite. Once BidiStages are available, together with the new DSL features it can be made much nicer looking.

Issue #16933, opened 2015-02-26
Think about materialization value handling in Http


drewhk:
Currently it is not very consistent, although in most of the cases of API methods:

 - take a materialized value type of _
 - produce a materialized value type of Unit

There might be better ways to handle this.

rkuhn:
One main point is that `entity.data` is declared as `Source[ByteString, Unit]` when in fact it should be `Source[ByteString, Any]` (to allow arbitrary materialized values—they are ignored anyway). This would work without any issue if this method were only ever mentioned in Scala source, but HttpEntityCloseDelimited.java kills it; the problem is that Unit is an AnyVal while Java cannot express the concept of Any.

rkuhn:
same as #17039

PR #17133, opened 2015-04-02
str, htc, htp: Change default materialization placeholder type from `Unit` to `Any`

### LINKED ISSUE #17039: Source.concat is difficult to use now that streams have the materialized type annotation
In HTTP 1.0-M4, all the HTTP body streams are of type `Source[ByteString, Unit]` 
(Why not `Source[ByteString, _]`?).

This means that the "concat" helper for Sources is difficult to use:

```
  def preamble: Source[ByteString, Unit] = ...
  def body: Source[ByteString, Unit] = ...
  def postamble: Source[ByteString, Unit] = ...

  val data = preamble ++ body ++ postamble

  val myResponse = HttpResponse(
    entity = HttpEntity.Chunked.fromData(contentType, data)
```

This used to work in 1.0-M3, but now I get
```
Error:(228, 43) type mismatch;
 found   : akka.stream.scaladsl.Source[akka.util.ByteString,((Unit, Unit), Unit)]
 required: akka.stream.scaladsl.Source[akka.util.ByteString,Unit]
```

What's the correct fix here?
Even `Keep.left` doesn't work, due to tuple wrapping (or something?)
```
    preamble 
      .concat(body).mapMaterialized(Keep.left[Unit,Unit])
      .concat(postamble).mapMaterialized(Keep.left[Unit,Unit])
```
gives:
```
Error:(229, 64) type mismatch;
 found   : (Unit, Unit) => Unit
 required: ((Unit, Unit)) => ?
      .concat(body).mapMaterialized(Keep.left[Unit,Unit])
                                                               ^
```

The only thing I can get to work is:
```
    preamble 
      .concat(body)
      .concat(postamble)
      .mapMaterialized(_ => ())
```
which feels clumsy.

 1. Is there a better way of using "++" or "concat" which avoids this issue, but which I missed?
 2. If not, does the DSL need rejigging so that "++" is nicer? Or is "++" expected to be rare?

Thanks,

Rich

sirthias:
Closes #17039 and #16933.

I needed this a basis for the current work in HTTP so I simply got it done altogether.
(For making things a bit easier for me) this PR builds on #17126, so please review the first 3 commits there. Also #17126 should be merged first.

### DIFF
--- a/akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeWorkerPool.scala
+++ b/akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeWorkerPool.scala
@@ -17,7 +17,7 @@ class RecipeWorkerPool extends RecipeSpec {
       val worker = Flow[String].map(_ + " done")
 
       //#worker-pool
-      def balancer[In, Out](worker: Flow[In, Out, Unit], workerCount: Int): Flow[In, Out, Unit] = {
+      def balancer[In, Out](worker: Flow[In, Out, Any], workerCount: Int): Flow[In, Out, Unit] = {
         import FlowGraph.Implicits._
 
         Flow() { implicit b =>
--- a/akka-http-core/src/main/scala/akka/http/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/Http.scala
@@ -15,7 +15,7 @@ import akka.stream.FlowMaterializer
 import akka.stream.scaladsl._
 import akka.http.engine.client._
 import akka.http.engine.server._
-import akka.http.model.{ HttpHeader, HttpResponse, HttpRequest }
+import akka.http.model.{ HttpResponse, HttpRequest }
 import akka.actor._
 
 class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.Extension {
@@ -40,17 +40,11 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
 
     val connections: Source[StreamTcp.IncomingConnection, Future[StreamTcp.ServerBinding]] =
       StreamTcp().bind(endpoint, backlog, options, effectiveSettings.timeouts.idleTimeout)
-    val bluePrint = HttpServerBluePrint(effectiveSettings, log)
+    val layer = serverLayer(effectiveSettings, log)
 
-    connections.map { conn ⇒
-      val flow = Flow(conn.flow, bluePrint)(Keep.right) { implicit b ⇒
-        (tcp, http) ⇒
-          import FlowGraph.Implicits._
-          tcp.outlet ~> http.bytesIn
-          http.bytesOut ~> tcp.inlet
-          (http.httpResponses, http.httpRequests)
-      }
-      IncomingConnection(conn.localAddress, conn.remoteAddress, flow)
+    connections.map {
+      case StreamTcp.IncomingConnection(localAddress, remoteAddress, flow) ⇒
+        IncomingConnection(localAddress, remoteAddress, layer join flow)
     }.mapMaterialized { tcpBindingFuture ⇒
       import system.dispatcher
       tcpBindingFuture.map { tcpBinding ⇒ ServerBinding(tcpBinding.localAddress)(() ⇒ tcpBinding.unbind()) }
@@ -69,11 +63,10 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
                     interface: String, port: Int = 80, backlog: Int = 100,
                     options: immutable.Traversable[Inet.SocketOption] = Nil,
                     settings: Option[ServerSettings] = None,
-                    log: LoggingAdapter = system.log)(implicit fm: FlowMaterializer): Future[ServerBinding] = {
-    bind(interface, port, backlog, options, settings, log).toMat(Sink.foreach { conn ⇒
-      conn.flow.join(handler).run()
-    })(Keep.left).run()
-  }
+                    log: LoggingAdapter = system.log)(implicit fm: FlowMaterializer): Future[ServerBinding] =
+    bind(interface, port, backlog, options, settings, log).to {
+      Sink.foreach { _.flow.join(handler).run() }
+    }.run()
 
   /**
    * Convenience method which starts a new HTTP server at the given endpoint and uses the given ``handler``
@@ -106,22 +99,30 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
     bindAndHandle(Flow[HttpRequest].mapAsync(1, handler), interface, port, backlog, options, settings, log)
 
   /**
-   * Transforms a given HTTP-level server [[Flow]] into a lower-level TCP transport flow.
+   * The type of the server-side HTTP layer as a stand-alone BidiStage
+   * that can be put atop the TCP layer to form an HTTP server.
+   *
+   * {{{
+   *                +------+
+   * HttpResponse ~>|      |~> ByteString
+   *                | bidi |
+   * HttpRequest  <~|      |<~ ByteString
+   *                +------+
+   * }}}
    */
-  def serverFlowToTransport[Mat](serverFlow: Flow[HttpRequest, HttpResponse, Mat],
-                                 settings: Option[ServerSettings] = None,
-                                 log: LoggingAdapter = system.log)(implicit mat: FlowMaterializer): Flow[ByteString, ByteString, Mat] = {
-    val effectiveSettings = ServerSettings(settings)
-    val bluePrint = HttpServerBluePrint(effectiveSettings, log)
+  type ServerLayer = BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, Any]
 
-    Flow(bluePrint, serverFlow)(Keep.right) { implicit b ⇒
-      (server, user) ⇒
-        import FlowGraph.Implicits._
-        server.httpRequests ~> user.inlet
-        user.outlet ~> server.httpResponses
-        (server.bytesIn, server.bytesOut)
-    }
-  }
+  /**
+   * Constructs a [[ServerLayer]] stage using the configured default [[ServerSettings]].
+   */
+  def serverLayer()(implicit mat: FlowMaterializer): ServerLayer = serverLayer(ServerSettings(system))
+
+  /**
+   * Constructs a [[ServerLayer]] stage using the given [[ServerSettings]].
+   */
+  def serverLayer(settings: ServerSettings,
+                  log: LoggingAdapter = system.log)(implicit mat: FlowMaterializer): ServerLayer =
+    BidiFlow.wrap(HttpServerBluePrint(settings, log))
 
   /**
    * Creates a [[Flow]] representing a prospective HTTP client connection to the given endpoint.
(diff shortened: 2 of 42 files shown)


akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeDigest.scala
```diff
@@ -41,7 +41,7 @@ class RecipeDigest extends RecipeSpec {
         }
       }
 
-      val digest: Source[ByteString, Unit] = data.transform(() => digestCalculator("SHA-256"))
```
drewhk: Nope. This is proper Unit.

drewhk:
This is changing too much. Things should *return* Unit as the materialized type if they really do not materialize anything meaningful. Only those parts should be changed where we *accept* something with an unknown materialized type. So it should be

 - Unit if *we* return it and we know it does not have a materialized value
 - Any if we *take it* as a parameter and we do not care about the materialized value

ktoso:
Meh, @drewhk was faster this time :-) 
It's as Endre writes, in `IncommingConnection(..., flow: Flow[ByteString, ByteString, Any]) {` we really want to accept Any, but when returning things we should expose the most specific type - Unit signalling nothing is going to be materialized.

sirthias:
Ok, agreed. I'll get this fixed up ASAP.

jrudolph:
+1 on doing this soon. It would be good if I could build on this rather sooner than later.

sirthias:
I'm on it and almost done.
Let me say this: getting this new type parameter right every (especially with regard to Java interop) is tricky (but doable).

sirthias:
Ok, applied the discussed reduction in scope.
In the end, not that much needs to be changed.

Apart from some smaller cleanups, the changes in the akka-stream module are essentially:

- Make `Flow::section` and `Source::section` accept function widened from Mat `Unit` to Mat `Any`.
  Also change it from `Keep.right` semantics to `Keep.left`, as the latter is the much more frequently required case.

- Make `FlattenStrategy` covariant in its second type param. Unfortunately, on the Java side, `FlattenStrategy.concat` needed to receive a second type param.

- Added `concatMat` to `Flow` and `Source` in order to solve the original problem reported by @RichardBradley in #17039

- While I was at it I also changed
  ```scala
  Source.apply[T](publisher: Publisher[T]): Source[T, Unit]
  ```
  to
  ```scala
  Source.apply[T](publisher: Publisher[_ >: T]): Source[T, Unit]
  ```
  (as well as, accordingly, the respective constructor on `Sink`). As this appears to be an (initially?) obvious thing to do there might be a reason why this hasn't been done so far. Please let me know if this is the case.

drewhk:
Cool! Actually section will probably go away now, since you can do the following now:

    val sectionFlow = Flow[X].map(f).withAttributes(attr)
    val parentFlow = Flow[Y].mapAsync(ff).via(sectionFlow).map(f2).filter(p) ...

all thanks to the now hierarchic internal layout representation.

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeWorkerPool.scala
```diff
@@ -17,7 +17,7 @@ class RecipeWorkerPool extends RecipeSpec {
       val worker = Flow[String].map(_ + " done")
 
       //#worker-pool
-      def balancer[In, Out](worker: Flow[In, Out, Unit], workerCount: Int): Flow[In, Out, Unit] = {
+      def balancer[In, Out, Mat](worker: Flow[In, Out, Any], workerCount: Int): Flow[In, Out, Unit] = {
```
rkuhn: this type parameter is not used

sirthias: Indeed. Left-over from some previous iteration.
Will remove.

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -194,7 +200,7 @@ object Http extends ExtensionId[HttpExt] with ExtensionIdProvider {
   case class IncomingConnection(
     localAddress: InetSocketAddress,
     remoteAddress: InetSocketAddress,
-    flow: Flow[HttpResponse, HttpRequest, Unit]) {
+    flow: Flow[HttpResponse, HttpRequest, Any]) {
```
rkuhn: here we provide a flow to users, so `Unit` was the right type

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -128,44 +135,44 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
-        (client.httpRequests, client.httpResponses)
-    }
-  }
+  /**
+   * Constructs a [[ClientLayer]] stage using the given [[ClientConnectionSettings]].
+   */
+  def clientLayer(remoteAddress: InetSocketAddress, // TODO: remove after #16168 is cleared
+                  settings: ClientConnectionSettings,
+                  log: LoggingAdapter = system.log): ClientLayer =
+    BidiFlow.wrap(OutgoingConnectionBlueprint(remoteAddress, settings, log))
```
rkuhn: why does the factory not directly return a BidiFlow?

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -128,44 +135,44 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
+   *                +------+
+   * }}}
    */
-  def transportToConnectionClientFlow[Mat](transport: Flow[ByteString, ByteString, Mat],
-                                           remoteAddress: InetSocketAddress, // TODO: removed after #16168 is cleared
-                                           settings: Option[ClientConnectionSettings] = None,
-                                           log: LoggingAdapter = system.log): Flow[HttpRequest, HttpResponse, Mat] = {
-    val effectiveSettings = ClientConnectionSettings(settings)
-    val clientBlueprint = HttpClient.clientBlueprint(remoteAddress, effectiveSettings, log)
+  type ClientLayer = BidiFlow[HttpRequest, ByteString, ByteString, HttpResponse, Any]
```
rkuhn: this is a return type that we provide to users, so it should be the concrete type that the blueprint materializes to

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -99,27 +99,34 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
-                                 log: LoggingAdapter = system.log)(implicit mat: FlowMaterializer): Flow[ByteString, ByteString, Mat] = {
-    val effectiveSettings = ServerSettings(settings)
-    val serverBlueprint: Graph[HttpServerPorts, Unit] = HttpServer.serverBlueprint(effectiveSettings, log)
-
-    Flow(serverBlueprint, serverFlow)(Keep.right) { implicit b ⇒
-      (server, user) ⇒
-        import FlowGraph.Implicits._
-        server.httpRequests ~> user.inlet
-        user.outlet ~> server.httpResponses
+  type ServerLayer = BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, Any]
```
rkuhn: same as for ClientLayer

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -99,27 +99,34 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
+   */
+  def serverLayer()(implicit mat: FlowMaterializer): ServerLayer = serverLayer(ServerSettings(system))
 
-  }
+  /**
+   * Constructs a [[ServerLayer]] stage using the given [[ServerSettings]].
+   */
+  def serverLayer(settings: ServerSettings,
+                  log: LoggingAdapter = system.log)(implicit mat: FlowMaterializer): ServerLayer =
+    BidiFlow.wrap(HttpServerBluePrint(settings, log))
```
rkuhn: same as below

akka-stream/src/main/boilerplate/akka/stream/FanInShape.scala.template
```diff
@@ -57,6 +57,21 @@ class UniformFanInShape[T, O](val n: Int, _init: FanInShape.Init[O]) extends Fan
...
+  override def deepCopy(): FanInShape1N[T0, T1, O] = super.deepCopy().asInstanceOf[FanInShape1N[T0, T1, O]]
+
+  val in0 = newInlet[T0]("in0")
+  val in1Vector: Vector[Inlet[T1]] = Vector.tabulate(n)(i => newInlet[T1](s"in${i+1}"))
+  def in(n: Int): Inlet[T1] = {
+    require(n > 0, "n must be > 0")
+    in1Vector(n - 1)
+  }
+}
+
```
rkuhn: you keep trying to sneak this in :-)

I think I know where it will be used, so leave it in this time.

akka-stream/src/main/scala/akka/stream/javadsl/BidiFlow.scala
```diff
@@ -12,7 +12,7 @@ object BidiFlow {
   val factory: BidiFlowCreate = new BidiFlowCreate {}
 
   /**
-   * A graph with the shape of a flow logically is a flow, this method makes
+   * A graph with the shape of a BidiFlow logically is a BidiFlow, this method makes
```
rkuhn: good catch!

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -58,7 +58,7 @@ object Source {
    * that mediate the flow of elements downstream and the propagation of
    * back-pressure upstream.
    */
-  def from[O](publisher: Publisher[O]): javadsl.Source[O, Unit] =
+  def from[O](publisher: Publisher[_ >: O]): javadsl.Source[O, Unit] =
```
rkuhn: This is wrong: a `Publisher[O]` is most adequately represented as a `Source[O]`. Then if a `Source[O2 >: O]` is required the Java way is to fix that at the use-site of the Source and not here. In any case, if we were to do something here then it would need to be `Publisher[_ <: O]` because otherwise our Source would sometimes emit things outside of its allowed type bound.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -57,7 +57,7 @@ object Sink extends SinkApply {
   /**
    * Helper to create [[Sink]] from `Subscriber`.
    */
-  def apply[T](subscriber: Subscriber[T]): Sink[T, Unit] =
+  def apply[T](subscriber: Subscriber[_ <: T]): Sink[T, Unit] =
```
rkuhn: same as above, with inverted polarity

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -171,7 +178,7 @@ object Source extends SourceApply {
    * that mediate the flow of elements downstream and the propagation of
    * back-pressure upstream.
    */
-  def apply[T](publisher: Publisher[T]): Source[T, Unit] =
+  def apply[T](publisher: Publisher[_ >: T]): Source[T, Unit] =
```
rkuhn: Here it is even more wrong, since Scala does have declaration-site variance, so this will never be useful (unless I am missing something).

akka-http-core/src/main/scala/akka/http/model/HttpEntity.scala
```diff
@@ -160,11 +160,11 @@ object HttpEntity {
     override def toStrict(timeout: FiniteDuration)(implicit fm: FlowMaterializer) =
```
rkuhn: line 130 is the most important one to fix

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: No reason. We could easily move the `BidiFlow.wrap` into `OutgoingConnectionBlueprint` if you'd prefer that.

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: Yup.

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: Yes, `Any` should be `Unit` here.
The client-side HTTP layer itself doesn't materialize to anything useful.

akka-stream/src/main/boilerplate/akka/stream/FanInShape.scala.template
sirthias: I simply proposed it.
This is the first time you (or anyone else) made a statement as to whether to leave it in or not... :)

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
sirthias: Of course. No idea what I was thinking here. I'll throw it out.

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
sirthias: Yeah, this is complete BS.
Sorry about that.

rkuhn: no worries, sometimes we get carried away :-)

akka-http-core/src/main/scala/akka/http/model/HttpEntity.scala
sirthias: Yeah, but, AFAICS, the only way to do it is to add a `data.mapMaterialized(_ ⇒ ()))`, which incurs overhead *always*, even if I already pass in a `Source[ByteString, Unit]`.

rkuhn: The solution is simple: use `Any` for the field’s type. Fixing this particular signature was the purpose of the whole exercise AFAIK, so I don’t think it is optional ;-)

sirthias: Well, the member is at same time something we accept from the user as well as something that we supply. So we have conflicting goals of accepting `Any` but supplying `Unit`.
I'll make the change to `Any` here and in all other similar cases.

akka-stream/src/main/boilerplate/akka/stream/FanInShape.scala.template
```diff
@@ -57,6 +57,21 @@ class UniformFanInShape[T, O](val n: Int, _init: FanInShape.Init[O]) extends Fan
...
 
+class FanInShape1N[T0, T1, O](val n: Int, _init: FanInShape.Init[O]) extends FanInShape[O](_init) {
+  def this(n: Int) = this(n, FanInShape.Name("FanInShape1N"))
+  def this(n: Int, name: String) = this(n, FanInShape.Name(name))
+  def this(outlet: Outlet[O], in0: Inlet[T0], inlets1: Array[Inlet[T1]]) = this(inlets1.length, FanInShape.Ports(outlet, in0 :: inlets1.toList))
+  override protected def construct(init: FanInShape.Init[O]): FanInShape[O] = new FanInShape1N(n, init)
+  override def deepCopy(): FanInShape1N[T0, T1, O] = super.deepCopy().asInstanceOf[FanInShape1N[T0, T1, O]]
+
+  val in0 = newInlet[T0]("in0")
+  val in1Vector: Vector[Inlet[T1]] = Vector.tabulate(n)(i => newInlet[T1](s"in${i+1}"))
```
patriknw: We normally use `immutable.IndexedSeq` as type in public api. Is there a reason to use `Vector`?

sirthias: I'll happily change it to `immutable.IndexedSeq`.

drewhk:
If I understand this also closes https://github.com/akka/akka/issues/16932

sirthias:
@drewhk Yes, it does. I'll make the change to `immutable.IndexedSeq` and rebase right now, so we can get this merged.

drewhk:
Yaay!

sirthias:
Done.

rkuhn:
well, I’ve got good news and bad news … the good is that I was about to click the button, the bad is that the button is not green ;-)

sirthias:
Damn. This thing really conflicts with everything.
I'll give it another try.

rkuhn:
quickly merged :-)

drewhk:
Yaay :)

jrudolph:
The only way to get this done!

sirthias:
Thank god!

rkuhn:
nah, was only me, nobody special …

PR #17134, opened 2015-04-02
Expand the documentation of Scheduler

romusz:


typesafehub-validator:
Hi @romusz,

akka-actor/src/main/scala/akka/actor/Scheduler.scala
```diff
@@ -68,7 +68,11 @@ trait Scheduler {
    * Schedules a function to be run repeatedly with an initial delay and a
    * frequency. E.g. if you would like the function to be run after 2 seconds
    * and thereafter every 100ms you would set delay = Duration(2, TimeUnit.SECONDS)
-   * and interval = Duration(100, TimeUnit.MILLISECONDS)
+   * and interval = Duration(100, TimeUnit.MILLISECONDS). If the execution of
+   * the function takes longer than the interval, the subsequent execution will
+   * start immediately after the prior one completes (there will be no overlap
+   * of the function executions). In such cases, the actual execution interval
```
drewhk: not only in those cases. Scheduler ticks can come later than specified, the only contract is that they will not come *earlier*.

patriknw:
LGTM

drewhk:
OK TO TEST

akka-actor/src/main/scala/akka/actor/Scheduler.scala
romusz: thanks for your reply. yes, i understand the scheduler in akka is 'lightweight' therefore ticks can come late. still, i think the additional comments are helpful to clarify the 'semantics' of the `schedule` method. In particular, it's useful to know that the method behaves more like `ScheduledExecutorService.scheduleAtFixedRate` than `ScheduledExecutorService.scheduleWithFixedDelay`, which is not obvious w/o reading the source or running a simple test.

drewhk: Yes, your additions are very helpful. I was just pondering if we should clarify it even more.

rkuhn:
LGTM

rkuhn:
Thanks for the contribution!

akka-actor/src/main/scala/akka/actor/Scheduler.scala
romusz: Yes, info about the ticks would be useful too.

PR #17144, opened 2015-04-07
ByteString serialization

colinrgodsey:
Addresses ByteString serialization (#15170) while keeping binary
compatibility with the previously serializable ByteString1C

Squashed from previous PR ~~#17070~~

Referenced in previous ByteString comments in #17028

### DIFF
--- a/akka-actor/src/main/scala/akka/util/ByteString.scala
+++ b/akka-actor/src/main/scala/akka/util/ByteString.scala
@@ -4,6 +4,7 @@
 
 package akka.util
 
+import java.io.{ ObjectInputStream, ObjectOutputStream }
 import java.nio.{ ByteBuffer, ByteOrder }
 import java.lang.{ Iterable ⇒ JIterable }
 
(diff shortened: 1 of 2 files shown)


drewhk:
OK TO TEST

ktoso:
LGTM, was reviewed in the previous PR already.

@colinrgodsey please adjust the commit message according to our guidelines [CONTRIBUTING.md](https://github.com/akka/akka/blob/master/CONTRIBUTING.md), in this case it would be: `=act #15170 bytestring serialization` Thanks! (You can simply `git commit --amen -m '=act #15170 bytestring serialization' && git push -f` to update this PR)

colinrgodsey:
@ktoso wow, didnt even know about force....

fixed.

patriknw:
LGTM

ktoso:
@colinrgodsey, yeah you can also `push -f` after squashing commits etc, so your work could have remained in the initial PR :-) Just a hint for the future, not a problem at all.

Thanks for the contrib again!

akka-actor/src/main/scala/akka/util/ByteString.scala
```diff
@@ -274,8 +322,43 @@ object ByteString {
...
+      val serializationId = in.readByte()
+
+      orig = Companion(from = serializationId).readFromInputStream(in)
+    }
+
+    private def readResolve(): AnyRef = orig
+  }
+
+  private[akka] object Companion {
+    private val companionMap = Seq(ByteString1, ByteString1C, ByteStrings).
```
viktorklang: Wouldn't this be much more performant with a `match` + `@tableswitch`?

colinrgodsey: hmm, probably shouldve =\

ktoso: How about submitting a PR with that improvement @colinrgodsey? Would be very welcome :-)

PR #17148, opened 2015-04-08
=clu #16726 Down member automatically when restarted

patriknw:
- When new uid is seen in join attempt we can down existing
  member and thereby new restarted node will be able to join
  in later retried join attempt without relying on auto-down.

(cherry picked from commit 37f6a6581cbe4d1ebfa4a478dd9e25c6d86f8ac6)


patriknw:
Refs #16726

PR #17152, opened 2015-04-08
=act #17140 FSM: forMax(Inf) can now override stateTimeout

### LINKED ISSUE #17140: Akka-FSM stateTimeout ignores forMax Duration.Inf
Code to reproduce:

```scala
sealed trait State
import akka.actor._

sealed trait Data
case object Initial extends State
case object Waiting extends State
case object Empty extends Data

class FSMActor extends LoggingFSM[State, Data] {
      startWith(Initial, Empty)
     
      when(Initial) {
        case Event("Wait", _) => goto(Waiting)
      }

      when(Waiting, stateTimeout = 5.seconds) {
        case Event("Message", _) => 
          println(self.path.name + " got message")
          stay() forMax (Duration.Inf)

        case Event(StateTimeout, _) => 
          println(self.path.name + " got StateTimeout :(")
          stay()
      }
}

val system = ActorSystem("system")
val fsm = system.actorOf(Props(new FSMActor), "fsm")
fsm ! "Wait"
fsm ! "Message"
```

Actor shouldn't receive `StateTimeout` message since `stay()` was called with `forMax(Duration.Inf)`.

Version is `2.4-SNAPSHOT`, bug appeared after update on Apr 02.

Related thread on akka-user: https://groups.google.com/forum/#!topic/akka-user/NsAoqPZ2Svo

ktoso:
Resolves #17140 while remaining binary and source compatible

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/FSM.scala
+++ b/akka-actor/src/main/scala/akka/actor/FSM.scala
@@ -8,6 +8,7 @@ import scala.concurrent.duration.Duration
 import scala.collection.mutable
 import akka.routing.{ Deafen, Listen, Listeners }
 import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration._
 
 object FSM {
 
@@ -118,6 +119,9 @@ object FSM {
    */
   final case class LogEntry[S, D](stateName: S, stateData: D, event: Any)
 
+  /** Used by `forMax` to signal "cancel stateTimeout" */
+  private final val SomeMaxFiniteDuration = Some(Long.MaxValue.nanos)
+
   /**
    * This captures all of the managed state of the [[akka.actor.FSM]]: the state
    * name, the state data, possibly custom timeout, stop reason and replies
@@ -141,8 +145,10 @@ object FSM {
      */
     def forMax(timeout: Duration): State[S, D] = timeout match {
       case f: FiniteDuration ⇒ copy(timeout = Some(f))
-      case _                 ⇒ copy(timeout = None)
-    }
+      case Duration.Inf      ⇒ copy(timeout = SomeMaxFiniteDuration) // we map the Infinite duration to a special marker,
+      case _                 ⇒ copy(timeout = None)                  // that means "cancel stateTimeout". This marker is needed
+    }                                                                // so we do not have to break source/binary compat.
+                                                                     // TODO: Can be removed once we can break State#timeout signature to `Option[Duration]`
 
     /**
      * Send reply to sender of the current message, if available.
@@ -648,13 +654,18 @@ trait FSM[S, D] extends Actor with Listeners with ActorLogging {
         this.nextState = null
       }
       currentState = nextState
-      val timeout = if (currentState.timeout.isDefined) currentState.timeout else stateTimeouts(currentState.stateName)
-      if (timeout.isDefined) {
-        val t = timeout.get
-        if (t.isFinite && t.length >= 0) {
-          import context.dispatcher
-          timeoutFuture = Some(context.system.scheduler.scheduleOnce(t, self, TimeoutMarker(generation)))
-        }
+
+      def scheduleTimeout(d: FiniteDuration): Some[Cancellable] = {
+        import context.dispatcher
+        Some(context.system.scheduler.scheduleOnce(d, self, TimeoutMarker(generation)))
+      }
+
+      currentState.timeout match {
+        case SomeMaxFiniteDuration                    ⇒ // effectively disable stateTimeout
+        case Some(d: FiniteDuration) if d.length >= 0 ⇒ timeoutFuture = scheduleTimeout(d)
+        case _ ⇒
+          val timeout = stateTimeouts(currentState.stateName)
+          if (timeout.isDefined) timeoutFuture = scheduleTimeout(timeout.get)
       }
     }
   }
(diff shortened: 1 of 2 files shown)


drewhk:
LGTM

Shoud this be backported to 2.3?

ktoso:
I think so, it can be considered a bug - in the docs we say this should work.

akka-actor/src/main/scala/akka/actor/FSM.scala
```diff
@@ -648,13 +653,18 @@ trait FSM[S, D] extends Actor with Listeners with ActorLogging {
...
-          timeoutFuture = Some(context.system.scheduler.scheduleOnce(t, self, TimeoutMarker(generation)))
-        }
+
+      def scheduleTimeout(d: FiniteDuration): Some[Cancellable] = {
+        import context.dispatcher
+        Some(context.system.scheduler.scheduleOnce(d, self, TimeoutMarker(generation)))
+      }
+
+      currentState.timeout match {
+        case SomeMaxFiniteDuration                    ⇒ // effectively disable stateTimeout
```
viktorklang: Is this a meaningful distinction given a FiniteDuration of Long.MaxValue - 1?

ktoso: effectively what I need to express "such big timeout that the akka scheduler would blow up anyway". 
Thought using MaxValue would be cleanest as marker value. I could dig into the scheduler and find it's max and cut-off anything above it to => "effectively never", but that sounds a bit more magic than picking MaxValue as my marker.

Note: Users are not expected to use it, they'd do `forMax(Inf)` which I encode as this MaxValue marker. (Can't change types, if I could we'd change things to Duration from FiniteDuration and we'd be good).

drewhk: this is just a workaround that the original code did not handle Duration.Inf although the forMax method (L146) accepts a Duration. The proper fix would be to change the signature of State to have a timeout of Option[Duration] instead of Option[FiniteDuration], but that would break binary compatiblility. So Konrad introduced this artifical marker that is introduced at L148 when a Duration.Inf is encountered, and which can be stored in the State case class without changing its signature.

ktoso: Reason for this marker-hack is the FiniteDuration here: `   final case class State[S, D](stateName: S, stateData: D, timeout: Option[FiniteDuration], ...`)

viktorklang: Makes total sense. Apologies for barging in here.

ktoso: Nah, no worries – more often than not such barging in is pretty useful :-)

akka-actor/src/main/scala/akka/actor/FSM.scala
```diff
@@ -141,6 +145,7 @@ object FSM {
      */
     def forMax(timeout: Duration): State[S, D] = timeout match {
       case f: FiniteDuration ⇒ copy(timeout = Some(f))
+      case Duration.Inf      ⇒ copy(timeout = SomeMaxFiniteDuration)
       case _                 ⇒ copy(timeout = None)
```
kevinmeredith: With the addition of `case Duration.Inf => ...`, isn't this line now dead code?

[Duration](http://www.scala-lang.org/api/2.11.5/index.html#scala.concurrent.duration.Duration)'s docs show that it has 2 children: `FiniteDuration` and `Infinite`.

akka-actor/src/main/scala/akka/actor/FSM.scala
drewhk: but @viktorklang 's comment is valid as in you should put a comment there to explain the "why".

akka-actor/src/main/scala/akka/actor/FSM.scala
rkuhn: There’s negative infinity and also `Undefined`.

akka-actor/src/main/scala/akka/actor/FSM.scala
ktoso: Yes agreed, I'll add one

akka-actor/src/main/scala/akka/actor/FSM.scala
```diff
@@ -1126,7 +1136,7 @@ abstract class AbstractFSM[S, D] extends FSM[S, D] {
    * @return current state descriptor
    */
   final def setTimer(name: String, msg: Any, timeout: FiniteDuration): Unit =
-    setTimer(name, msg, timeout, false);
```
rkuhn: aargh, my eyes!

akka-actor/src/main/scala/akka/actor/FSM.scala
```diff
@@ -1126,7 +1136,7 @@ abstract class AbstractFSM[S, D] extends FSM[S, D] {
...
+    setTimer(name, msg, timeout, false)
```
rkuhn: phew, thanks!

ktoso: got them now bright red in my new color scheme :wink:

rkuhn:
LGTM

akka-actor/src/main/scala/akka/actor/FSM.scala
ktoso: Added a comment explaining the reasoning (and todo - we can fix it once we break compat).

akka-actor/src/main/scala/akka/actor/FSM.scala
ktoso: RE: https://twitter.com/Gentmen/status/586192587654041600
The other mentioned values are not classes but values in `Duration`, as seen here:
https://github.com/scala/scala/blob/2.11.x/src/library/scala/concurrent/duration/Duration.scala#L174

Issue #13869, opened 2014-02-11
do not try to hand-over to None in ClusterSingletonManager


rkuhn:
imported from https://www.assembla.com/spaces/akka/tickets/3869

https://groups.google.com/d/msg/akka-user/ROmRV3ug3DE/oyiuRaRAwy0J

PR #17154, opened 2015-04-08
=con #13869 Speedup startup of Cluster singleton

patriknw:
- avoid the hand-over/take-over attempts when starting the
  ClusterSingletonManager for the normal clase when the cluster is
  in a good shape, i.e. no exiting member that might run previous
  singleton instance

### DIFF
--- a/akka-contrib/src/main/scala/akka/contrib/pattern/ClusterSingletonManager.scala
+++ b/akka-contrib/src/main/scala/akka/contrib/pattern/ClusterSingletonManager.scala
@@ -145,7 +145,7 @@ object ClusterSingletonManager {
       /**
        * The first event, corresponding to CurrentClusterState.
        */
-      final case class InitialOldestState(oldest: Option[Address], memberCount: Int)
+      final case class InitialOldestState(oldest: Option[Address], safeToBeOldest: Boolean)
 
       final case class OldestChanged(oldest: Option[Address])
     }
@@ -191,8 +191,9 @@ object ClusterSingletonManager {
 
       def handleInitial(state: CurrentClusterState): Unit = {
         membersByAge = immutable.SortedSet.empty(ageOrdering) ++ state.members.filter(m ⇒
-          m.status == MemberStatus.Up && matchingRole(m))
-        val initial = InitialOldestState(membersByAge.headOption.map(_.address), membersByAge.size)
+          (m.status == MemberStatus.Up || m.status == MemberStatus.Leaving) && matchingRole(m))
+        val safeToBeOldest = !state.members.exists { m ⇒ (m.status == MemberStatus.Down || m.status == MemberStatus.Exiting) }
+        val initial = InitialOldestState(membersByAge.headOption.map(_.address), safeToBeOldest)
         changes :+= initial
       }
 
@@ -421,10 +422,10 @@ class ClusterSingletonManager(
       getNextOldestChanged()
       stay
 
-    case Event(InitialOldestState(oldestOption, memberCount), _) ⇒
+    case Event(InitialOldestState(oldestOption, safeToBeOldest), _) ⇒
       oldestChangedReceived = true
-      if (oldestOption == selfAddressOption && memberCount == 1)
-        // alone, oldest immediately
+      if (oldestOption == selfAddressOption && safeToBeOldest)
+        // oldest immediately
         gotoOldest()
       else if (oldestOption == selfAddressOption)
         goto(BecomingOldest) using BecomingOldestData(None)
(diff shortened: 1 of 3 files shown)


patriknw:
Refs #13869

akka-contrib/src/main/scala/akka/contrib/pattern/ClusterSingletonManager.scala
```diff
@@ -191,8 +191,9 @@ object ClusterSingletonManager {
 
       def handleInitial(state: CurrentClusterState): Unit = {
         membersByAge = immutable.SortedSet.empty(ageOrdering) ++ state.members.filter(m ⇒
-          m.status == MemberStatus.Up && matchingRole(m))
-        val initial = InitialOldestState(membersByAge.headOption.map(_.address), membersByAge.size)
+          (m.status == MemberStatus.Up || m.status == MemberStatus.Leaving) && matchingRole(m))
+        val safeToBeOldest = !state.members.exists { m ⇒ (m.status == MemberStatus.Down || m.status == MemberStatus.Exiting) }
```
rkuhn: Is this to guard against the previously oldest being in the process of leaving? In that case, would it make sense to compute the previously oldest?

akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterSingletonManagerLeaveSpec.scala
```diff
@@ -0,0 +1,133 @@
...
+        expectMsg(10.seconds, "stopped")
+      }
+      enterBarrier("first-stopped")
+
+      runOn(second, third) {
+        val p = TestProbe()
+        p.within(10.seconds) {
+          p.awaitAssert {
+            echoProxy.tell("hello2", p.ref)
+            p.expectMsgType[ActorRef](1.seconds)
```
rkuhn: perhaps verify that the received ref does not reside on `first`?

akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterSingletonManagerStartupSpec.scala
```diff
@@ -0,0 +1,102 @@
...
+  "Startup of Cluster Singleton" must {
+
+    "be quick" in {
+      join(first, first)
+      join(second, first)
+      join(third, first)
+
+      runOn(first) {
+        echoProxy ! "hello"
+        expectMsgType[ActorRef](10.seconds)
```
rkuhn: why can it take 10 seconds? what exactly would “being slow” mean in terms of timeout?

rkuhn:
LGTM apart from some small questions

akka-contrib/src/main/scala/akka/contrib/pattern/ClusterSingletonManager.scala
patriknw: That is done on next line, `membersByAge.headOption`.
I guess you are thinking about that is `safeToBeOldest` also when a non-oldest is leaving?
I rather be careful than trying to optimize too much here. All that are leaving are most likely older than current node and therefore potential owners of the singleton instance.

akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterSingletonManagerStartupSpec.scala
patriknw: This is the essence of what this improvement is about. Previously the whole test took 16 seconds, now it takes 4 seconds.

10 seconds is a "huge" safety margin, but anyway this test would fail with the old solution.
Since it is also including the joining phase I rather not have it too tight.

akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterSingletonManagerLeaveSpec.scala
patriknw: That is implicitly already done, by the barrier, but it doesn't hurt. Added here: https://github.com/akka/akka/commit/1c311f82feb42b282f0c296e5774b6fd4d9049e5

akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterSingletonManagerStartupSpec.scala
rkuhn: Okay, my conclusion would be that `10.seconds` does not adequately represent the sweat and tears that went into this, perhaps commemorate with a comment?

patriknw: right, perhaps this is better: 381c1da

rkuhn: Yes, thanks!

rkuhn:
merge?

patriknw:
It's ready from my point of view. I have squashed.

Issue #16378, opened 2014-11-25
Where to do debug logging/tracing inside flows?


bantonsson:
When moving HTTP onto the new IO API, the question came up about how to do some deb/trace logging inside a flow, or at materialization time.

rkuhn:
This should be a prefabricated Stage that can be plugged into a Flow.

PR #17162, opened 2015-04-09
Logging with log() stage for streams

ktoso:
This WIP is a bit in the middle between "very outside" (how the `.debug()` is attached) and adding ast nodes etc so "inside". I got the feeling that since I need "more things" from the materializer (the flow's name - `materializationName` how I call it), it needs to be a proper built-in thing with it's proper AST node etc. - looking for opinions about that.

Looking for feedback around things that have TODOs. Esp. If this should simply be put on Source / Flow instead of trying to make it opt-in via import like I started out.

refs https://github.com/akka/akka/issues/16378

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLogSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLogSpec.scala
@@ -1,0 +1,173 @@
+/**
+ * Copyright (C) 2014-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.scaladsl
+
+import akka.event.{ DummyClassForStringSources, Logging }
+import akka.stream._
+import akka.stream.OperationAttributes
+import akka.stream.OperationAttributes.LogLevels
+import akka.stream.testkit.{ AkkaSpec, ScriptedTest }
+import akka.testkit.TestProbe
+
+import scala.util.control.NoStackTrace
+
+class FlowLogSpec extends AkkaSpec("akka.loglevel = DEBUG") with ScriptedTest {
+
+  implicit val mat: FlowMaterializer = ActorFlowMaterializer()
+
+  val logProbe = {
+    val p = TestProbe()
+    system.eventStream.subscribe(p.ref, classOf[Logging.LogEvent])
+    p
+  }
+
+  "A Log" must {
+
+    val LogSrc = s"akka.stream.Log(akka://${Logging.simpleName(classOf[FlowLogSpec])})"
+    val LogClazz = classOf[DummyClassForStringSources]
+
+    "on Flow" must {
+
+      "debug each element" in {
+        val debugging = Flow[Int].log("my-debug")
+        Source(1 to 2).via(debugging).runWith(Sink.ignore)
+
+        logProbe.expectMsg(Logging.Debug(LogSrc, LogClazz, "[my-debug] Element: 1"))
+        logProbe.expectMsg(Logging.Debug(LogSrc, LogClazz, "[my-debug] Element: 2"))
+        logProbe.expectMsg(Logging.Debug(LogSrc, LogClazz, "[my-debug] Upstream finished."))
+      }
+
+      "allow disabling element logging" in {
+        val disableElementLogging = OperationAttributes.logLevels(
+          onElement = LogLevels.Off,
+          onFinish = Logging.DebugLevel,
+          onFailure = Logging.DebugLevel)
+
+        val debugging = Flow[Int].log("my-debug")
+        Source(1 to 2)
+          .via(debugging)
+          .withAttributes(disableElementLogging)
+          .runWith(Sink.ignore)
+
+        logProbe.expectMsg(Logging.Debug(LogSrc, LogClazz, "[my-debug] Upstream finished."))
+      }
+
+    }
+
+    "on javadsl.Flow" must {
+      "debug each element" in {
+        val log = Logging(system, "com.example.ImportantLogger")
+
+        val debugging: javadsl.Flow[Integer, Integer, Unit] = javadsl.Flow.of(classOf[Integer])
+          .log("log-1")
+          .log("log-2", new akka.japi.function.Function[Integer, Integer] { def apply(i: Integer) = i })
+          .log("log-3", new akka.japi.function.Function[Integer, Integer] { def apply(i: Integer) = i }, log)
+          .log("log-4", log)
+
+        javadsl.Source.single[Integer](1).via(debugging).runWith(javadsl.Sink.ignore(), mat)
+
+        var counter = 0
+        var finishCounter = 0
+        import scala.concurrent.duration._
+        logProbe.fishForMessage(3.seconds) {
+          case Logging.Debug(_, _, msg: String) if msg contains "Element: 1" ⇒
+            counter += 1
+            counter == 4 && finishCounter == 4
+
+          case Logging.Debug(_, _, msg: String) if msg contains "Upstream finished" ⇒
+            finishCounter += 1
+            counter == 4 && finishCounter == 4
+        }
+      }
+    }
+
+    "on Source" must {
+      "debug each element" in {
+        Source(1 to 2).log("flow-s2").runWith(Sink.ignore)
+
+        logProbe.expectMsg(Logging.Debug(LogSrc, LogClazz, "[flow-s2] Element: 1"))
+        logProbe.expectMsg(Logging.Debug(LogSrc, LogClazz, "[flow-s2] Element: 2"))
+        logProbe.expectMsg(Logging.Debug(LogSrc, LogClazz, "[flow-s2] Upstream finished."))
+      }
+
+      "allow extracting value to be logged" in {
+        case class Complex(a: Int, b: String)
+        Source.single(Complex(1, "42")).log("flow-s3", _.b).runWith(Sink.ignore)
+
+        logProbe.expectMsg(Logging.Debug(LogSrc, LogClazz, "[flow-s3] Element: 42"))
+        logProbe.expectMsg(Logging.Debug(LogSrc, LogClazz, "[flow-s3] Upstream finished."))
+      }
+
+      "log upstream failure" in {
+        val cause = new TestException
+        Source.failed(cause).log("flow-4").runWith(Sink.ignore)
+        logProbe.expectMsg(Logging.Error(cause, LogSrc, LogClazz, "[flow-4] Upstream failed."))
+      }
+
+      "allow passing in custom LoggingAdapter" in {
+        val log = Logging(system, "com.example.ImportantLogger")
+
+        Source.single(42).log("flow-5")(log).runWith(Sink.ignore)
+
+        val src = "com.example.ImportantLogger(akka://FlowLogSpec)"
+        val clazz = classOf[DummyClassForStringSources]
+        logProbe.expectMsg(Logging.Debug(src, clazz, "[flow-5] Element: 42"))
+        logProbe.expectMsg(Logging.Debug(src, clazz, "[flow-5] Upstream finished."))
+      }
+
+      "allow configuring log levels via OperationAttributes" in {
+        val logAttrs = OperationAttributes.logLevels(
+          onElement = Logging.WarningLevel,
+          onFinish = Logging.InfoLevel,
+          onFailure = Logging.DebugLevel)
+
+        Source.single(42)
+          .log("flow-6")
+          .withAttributes(OperationAttributes.logLevels(
+            onElement = Logging.WarningLevel,
+            onFinish = Logging.InfoLevel,
+            onFailure = Logging.DebugLevel))
+          .runWith(Sink.ignore)
+
+        logProbe.expectMsg(Logging.Warning(LogSrc, LogClazz, "[flow-6] Element: 42"))
+        logProbe.expectMsg(Logging.Info(LogSrc, LogClazz, "[flow-6] Upstream finished."))
+
+        val cause = new TestException
+        Source.failed(cause)
+          .log("flow-6e")
+          .withAttributes(logAttrs)
+          .runWith(Sink.ignore)
+        logProbe.expectMsg(Logging.Debug(LogSrc, LogClazz, "[flow-6e] Upstream failed, cause: FlowLogSpec$TestException: Boom!"))
+      }
+    }
+
+    "on javadsl.Source" must {
+      "debug each element" in {
+        val log = Logging(system, "com.example.ImportantLogger")
+
+        javadsl.Source.single[Integer](1)
+          .log("log-1")
+          .log("log-2", new akka.japi.function.Function[Integer, Integer] { def apply(i: Integer) = i })
+          .log("log-3", new akka.japi.function.Function[Integer, Integer] { def apply(i: Integer) = i }, log)
+          .log("log-4", log)
+          .runWith(javadsl.Sink.ignore(), mat)
+
+        var counter = 1
+        import scala.concurrent.duration._
+        logProbe.fishForMessage(3.seconds) {
+          case Logging.Debug(_, _, msg: String) if msg contains "Element: 1" ⇒
+            counter += 1
+            counter == 4
+
+          case Logging.Debug(_, _, msg: String) if msg contains "Upstream finished" ⇒
+            false
+        }
+      }
+    }
+
+  }
+
+  final class TestException extends RuntimeException("Boom!") with NoStackTrace
+
+}
--- a/akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
+++ b/akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
@@ -11,7 +11,7 @@ import akka.stream.impl._
 import com.typesafe.config.Config
 
 import scala.concurrent.duration._
-import akka.japi.{ function ⇒ japi }
+import akka.japi.function
 
 object ActorFlowMaterializer {
 
@@ -145,13 +145,15 @@ abstract class ActorFlowMaterializer extends FlowMaterializer {
 
   def effectiveSettings(opAttr: OperationAttributes): ActorFlowMaterializerSettings
 
-  /** INTERNAL API */
-  private[akka] def system: ActorSystem
-
   /**
    * INTERNAL API: this might become public later
    */
   private[akka] def actorOf(context: MaterializationContext, props: Props): ActorRef
+
+  /**
+   * INTERNAL API
+   */
+  private[akka] def system: ActorSystem
 
 }
 
@@ -247,7 +249,7 @@ final case class ActorFlowMaterializerSettings(
    * overridden for specific flows of the stream operations with
    * [[akka.stream.OperationAttributes#supervisionStrategy]].
    */
-  def withSupervisionStrategy(decider: japi.Function[Throwable, Supervision.Directive]): ActorFlowMaterializerSettings = {
+  def withSupervisionStrategy(decider: function.Function[Throwable, Supervision.Directive]): ActorFlowMaterializerSettings = {
     import Supervision._
     copy(supervisionDecider = decider match {
       case `resumingDecider`   ⇒ resumingDecider
--- a/akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
@@ -53,6 +53,7 @@ private[akka] case class ActorFlowMaterializerImpl(
         case InputBuffer(initial, max)    ⇒ s.withInputBuffer(initial, max)
         case Dispatcher(dispatcher)       ⇒ s.withDispatcher(dispatcher)
         case SupervisionStrategy(decider) ⇒ s.withSupervisionStrategy(decider)
+        case l: LogLevels                 ⇒ s
         case Name(_)                      ⇒ s
       }
     }
@@ -286,30 +287,31 @@ private[akka] object ActorProcessorFactory {
     // Also, otherwise the attributes will not affect the settings properly!
     val settings = materializer.effectiveSettings(att)
     op match {
-      case Identity(_)                ⇒ (ActorInterpreter.props(settings, List(fusing.Map(_identity, settings.supervisionDecider)), materializer), ())
-      case Fused(ops, _)              ⇒ (ActorInterpreter.props(settings, ops, materializer), ())
-      case Map(f, _)                  ⇒ (ActorInterpreter.props(settings, List(fusing.Map(f, settings.supervisionDecider)), materializer), ())
-      case Filter(p, _)               ⇒ (ActorInterpreter.props(settings, List(fusing.Filter(p, settings.supervisionDecider)), materializer), ())
-      case Drop(n, _)                 ⇒ (ActorInterpreter.props(settings, List(fusing.Drop(n)), materializer), ())
-      case Take(n, _)                 ⇒ (ActorInterpreter.props(settings, List(fusing.Take(n)), materializer), ())
-      case Collect(pf, _)             ⇒ (ActorInterpreter.props(settings, List(fusing.Collect(settings.supervisionDecider)(pf)), materializer), ())
-      case Scan(z, f, _)              ⇒ (ActorInterpreter.props(settings, List(fusing.Scan(z, f, settings.supervisionDecider)), materializer), ())
-      case Expand(s, f, _)            ⇒ (ActorInterpreter.props(settings, List(fusing.Expand(s, f)), materializer), ())
-      case Conflate(s, f, _)          ⇒ (ActorInterpreter.props(settings, List(fusing.Conflate(s, f, settings.supervisionDecider)), materializer), ())
-      case Buffer(n, s, _)            ⇒ (ActorInterpreter.props(settings, List(fusing.Buffer(n, s)), materializer), ())
-      case MapConcat(f, _)            ⇒ (ActorInterpreter.props(settings, List(fusing.MapConcat(f, settings.supervisionDecider)), materializer), ())
-      case MapAsync(p, f, _)          ⇒ (ActorInterpreter.props(settings, List(fusing.MapAsync(p, f, settings.supervisionDecider)), materializer), ())
-      case MapAsyncUnordered(p, f, _) ⇒ (ActorInterpreter.props(settings, List(fusing.MapAsyncUnordered(p, f, settings.supervisionDecider)), materializer), ())
-      case Grouped(n, _)              ⇒ (ActorInterpreter.props(settings, List(fusing.Grouped(n)), materializer), ())
+      case Identity(_)                ⇒ (ActorInterpreter.props(settings, List(fusing.Map(_identity, settings.supervisionDecider)), materializer, att), ())
+      case Fused(ops, _)              ⇒ (ActorInterpreter.props(settings, ops, materializer, att), ())
+      case Map(f, _)                  ⇒ (ActorInterpreter.props(settings, List(fusing.Map(f, settings.supervisionDecider)), materializer, att), ())
+      case Filter(p, _)               ⇒ (ActorInterpreter.props(settings, List(fusing.Filter(p, settings.supervisionDecider)), materializer, att), ())
+      case Drop(n, _)                 ⇒ (ActorInterpreter.props(settings, List(fusing.Drop(n)), materializer, att), ())
+      case Take(n, _)                 ⇒ (ActorInterpreter.props(settings, List(fusing.Take(n)), materializer, att), ())
+      case Collect(pf, _)             ⇒ (ActorInterpreter.props(settings, List(fusing.Collect(settings.supervisionDecider)(pf)), materializer, att), ())
+      case Scan(z, f, _)              ⇒ (ActorInterpreter.props(settings, List(fusing.Scan(z, f, settings.supervisionDecider)), materializer, att), ())
+      case Expand(s, f, _)            ⇒ (ActorInterpreter.props(settings, List(fusing.Expand(s, f)), materializer, att), ())
+      case Conflate(s, f, _)          ⇒ (ActorInterpreter.props(settings, List(fusing.Conflate(s, f, settings.supervisionDecider)), materializer, att), ())
+      case Buffer(n, s, _)            ⇒ (ActorInterpreter.props(settings, List(fusing.Buffer(n, s)), materializer, att), ())
+      case MapConcat(f, _)            ⇒ (ActorInterpreter.props(settings, List(fusing.MapConcat(f, settings.supervisionDecider)), materializer, att), ())
+      case MapAsync(p, f, _)          ⇒ (ActorInterpreter.props(settings, List(fusing.MapAsync(p, f, settings.supervisionDecider)), materializer, att), ())
+      case MapAsyncUnordered(p, f, _) ⇒ (ActorInterpreter.props(settings, List(fusing.MapAsyncUnordered(p, f, settings.supervisionDecider)), materializer, att), ())
+      case Grouped(n, _)              ⇒ (ActorInterpreter.props(settings, List(fusing.Grouped(n)), materializer, att), ())
+      case Log(n, e, l, _)            ⇒ (ActorInterpreter.props(settings, List(fusing.Log(n, e, l)), materializer, att), ())
       case GroupBy(f, _)              ⇒ (GroupByProcessorImpl.props(settings, f), ())
       case PrefixAndTail(n, _)        ⇒ (PrefixAndTailImpl.props(settings, n), ())
       case SplitWhen(p, _)            ⇒ (SplitWhenProcessorImpl.props(settings, p), ())
       case ConcatAll(_)               ⇒ (ConcatAllImpl.props(materializer), ())
-      case StageFactory(mkStage, _)   ⇒ (ActorInterpreter.props(settings, List(mkStage()), materializer), ())
+      case StageFactory(mkStage, _)   ⇒ (ActorInterpreter.props(settings, List(mkStage()), materializer, att), ())
       case TimerTransform(mkStage, _) ⇒ (TimerTransformerProcessorsImpl.props(settings, mkStage()), ())
       case MaterializingStageFactory(mkStageAndMat, _) ⇒
         val sm = mkStageAndMat()
-        (ActorInterpreter.props(settings, List(sm._1), materializer), sm._2)
+        (ActorInterpreter.props(settings, List(sm._1), materializer, att), sm._2)
       case DirectProcessor(p, m) ⇒ throw new AssertionError("DirectProcessor cannot end up in ActorProcessorFactory")
     }
   }
--- a/akka-stream/src/main/scala/akka/stream/impl/Stages.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/Stages.scala
@@ -3,7 +3,7 @@
  */
 package akka.stream.impl
 
-import akka.event.Logging
+import akka.event.{ LoggingAdapter, Logging }
 import akka.stream.{ OverflowStrategy, TimerTransformer }
 import akka.stream.OperationAttributes
 import akka.stream.OperationAttributes._
@@ -128,6 +128,11 @@ private[stream] object Stages {
   }
 
   final case class Map(f: Any ⇒ Any, attributes: OperationAttributes = map) extends StageModule {
+    def withAttributes(attributes: OperationAttributes) = copy(attributes = attributes)
+    override protected def newInstance: StageModule = this.copy()
+  }
+
+  final case class Log(name: String, extract: Any ⇒ Any, loggingAdapter: Option[LoggingAdapter], attributes: OperationAttributes = map) extends StageModule {
     def withAttributes(attributes: OperationAttributes) = copy(attributes = attributes)
     override protected def newInstance: StageModule = this.copy()
   }
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
@@ -3,13 +3,14 @@
  */
 package akka.stream.impl.fusing
 
-import scala.annotation.{ tailrec, switch }
+import akka.stream.{ FlowMaterializer, Supervision }
+import akka.stream.impl.ReactiveStreamsCompliance
+import akka.stream.OperationAttributes
+import akka.stream.stage._
+
+import scala.annotation.{ switch, tailrec }
 import scala.collection.breakOut
 import scala.util.control.NonFatal
-import akka.stream.stage._
-import akka.stream.Supervision
-import akka.stream.impl.ReactiveStreamsCompliance
-import akka.stream.FlowMaterializer
 
 // TODO:
 // fix jumpback table with keep-going-on-complete ops (we might jump between otherwise isolated execution regions)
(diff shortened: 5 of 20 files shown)


akka-stream-tests/src/test/scala/akka/stream/scaladsl/pattern/FlowDebugSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+  "A Debug" must {
+
+    import akka.stream.scaladsl.pattern.debug._
+
+    val LogSrc = "akka.stream.debug.Debug(akka://FlowDebugSpec)"
+    val LogClazz = classOf[DummyClassForStringSources]
+
+    "on Flow" must {
+
+      "debug each element" in {
```
drewhk: add a nested example

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
```diff
@@ -74,7 +74,7 @@ private[akka] case class ActorFlowMaterializerImpl(override val settings: ActorF
             mat
 
           case stage: StageModule ⇒
-            val (processor, mat) = processorFor(stage, effectiveAttributes, calcSettings(effectiveAttributes)(settings))
+            val (processor, mat) = processorFor(stage, effectiveAttributes.and(OperationAttributes.materializationName(flowName)), calcSettings(effectiveAttributes)(settings))
```
drewhk: I am not sure about this. Aren't effectiveattributes already containing every relevant name?

akka-stream/src/main/scala/akka/stream/scaladsl/pattern/Debug.scala
```diff
@@ -0,0 +1,58 @@
...
+    stage.andThen(Stages.DebugLogging(extract.asInstanceOf[Any ⇒ Any], adapter)).asInstanceOf[Repr]
+  }
+}
+
+class DebugStage[T](extract: T ⇒ Any, materializationName: String, log: LoggingAdapter) extends PushStage[T, T] {
+
+  override def onPush(elem: T, ctx: Context[T]): Directive = {
+    log.debug("[{}] Element flowing through: {}", materializationName, extract(elem)) // TODO do we want to make the log levels configurable?
+    ctx.push(elem)
+  }
```
ktoso: Ideas: Could use `DiagnosticLoggingAdapter` to put more infos in MDC?

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
ktoso: It only contains names of stages, not the entire `flowName`. And that's the one I need in order to show to be able to correlate log statements for a given materialization. If one materialization fails and another one doesn't, you want to see their logs separately - which is why we need to log "flow-1544" / "flow-2323" etc.

flowName is: `val name = s"$flowName-$nextId-${attr.name}"`, so `flow-1`

ktoso: (I'd move the `and` somewhere cleaner than in-line here).

akka-stream-tests/src/test/scala/akka/stream/scaladsl/pattern/FlowDebugSpec.scala
```diff
@@ -0,0 +1,89 @@
...
+    .withInputBuffer(initialSize = 2, maxSize = 16)
+
+  implicit val mat: FlowMaterializer = ActorFlowMaterializer()
+
+  def logProbe(): TestProbe = {
+    val p = TestProbe()
+    system.eventStream.subscribe(p.ref, classOf[Logging.Debug])
+    system.eventStream.subscribe(p.ref, classOf[Logging.Info])
+    system.eventStream.subscribe(p.ref, classOf[Logging.Warning])
+    system.eventStream.subscribe(p.ref, classOf[Logging.Error])
```
patriknw: replace with subscribe of `LogEvent`?

patriknw: what is the reason for not using the log testing facilities in testkit?

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
```diff
@@ -91,7 +91,7 @@ private[stream] object Stages {
 
   object Fused {
     def apply(ops: immutable.Seq[Stage[_, _]]): Fused =
-      Fused(ops, name(ops.iterator.map(x ⇒ Logging.simpleName(x).toLowerCase).mkString("+")))
+      Fused(ops, name(ops.iterator.map(x ⇒ akka.event.Logging.simpleName(x).toLowerCase).mkString("+")))
```
patriknw: ?

akka-stream/src/main/scala/akka/stream/scaladsl/pattern/Debug.scala
```diff
@@ -0,0 +1,58 @@
...
+final case class DebuggableFlow[In, Out, Mat](stage: Flow[In, Out, Mat]) extends AnyVal with DebuggableLike[Out, Flow[In, Out, Mat]]
+
+final case class DebuggableSource[Out, Mat](stage: Source[Out, Mat]) extends AnyVal with DebuggableLike[Out, Source[Out, Mat]]
+
+trait DebuggableLike[Out, Repr <: FlowOps[Out, _]] extends Any {
+  protected def stage: Repr
+
+  @implicitNotFound("Cannot find ActorSystem to create LoggingAdapter from") // TODO can we require less? I could use the Materializer yet make it's actor sys akka private?
+  def debug(extract: Out ⇒ Any = identity)(implicit sys: ActorSystem): Repr = { // TODO allow implicit DebugConfig for configuring loglevels?
+    val adapter = akka.event.Logging(sys, /* TODO what here? */ "akka.stream.debug.Debug") // TODO decide on the name + could include materialization name?
```
patriknw: It might be inefficient to use the full materialization name in the LoggingAdapter when using slf4j?

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
drewhk: hmm. I am not happy that this creeps in as an attribute.

akka-stream/src/main/scala/akka/stream/scaladsl/pattern/Debug.scala
```diff
@@ -0,0 +1,58 @@
...
+import akka.stream.scaladsl.{ FlowOps, Source, FlowGraph, Flow }
+import akka.stream.stage._
+import akka.stream.scaladsl.FlowGraph.Implicits._
+import scala.annotation.implicitNotFound
+import scala.language.implicitConversions
+
+// TODO do we want this to be a "tag on pattern", or put debug() right on Source / Flow?
+object debug {
+
+  implicit def addDebugFlow[In, Out, Mat](f: Flow[In, Out, Mat])(implicit sys: ActorSystem): DebuggableFlow[In, Out, Mat] =
```
patriknw: if add this as a real FlowOp the implicit ActorSystem would not be needed, right?

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
ktoso: at some point my class was named Logging, causing this. I'll revert those once we settle for a name for it (Logging / Debug etc).

akka-stream/src/main/scala/akka/stream/scaladsl/OperationAttributes.scala
```diff
@@ -57,6 +63,7 @@ object OperationAttributes {
   final case class InputBuffer(initial: Int, max: Int) extends Attribute
   final case class Dispatcher(dispatcher: String) extends Attribute
   final case class SupervisionStrategy(decider: Supervision.Decider) extends Attribute
+  final case class MaterializationName(n: String) extends Attribute
```
patriknw: should this be public api?

akka-stream/src/main/scala/akka/stream/scaladsl/pattern/Debug.scala
patriknw: we could also let the user pass in the `LoggingAdapter`?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/pattern/FlowDebugSpec.scala
ktoso: 1) Oh, right - will do.
2) Ah you mean `EventFilter`, yes that would be cleaner - I'll use it.

akka-stream/src/main/scala/akka/stream/scaladsl/pattern/Debug.scala
ktoso: That would work I think, yes: Would have to downcast to `ActorFlowMaterializer` and get the `LogSource` based on the `supervisor: ActorRef` contained in the materializer.

akka-stream/src/main/scala/akka/stream/scaladsl/OperationAttributes.scala
ktoso: If we go with it probably no. I'll think of another way to get this data to my Stage though - @drewhk points out it feels quite hacky.

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
ktoso: I'll try to figure out another way to get this info to the Stage, but not sure how much I can without changing too much in the materializer hm.

akka-stream/src/main/scala/akka/stream/scaladsl/OperationAttributes.scala
drewhk: Not that it is hacky. My main problem is that attributes are user api while this materialized name is specific to the materializer.

akka-stream/src/main/scala/akka/stream/scaladsl/pattern/Debug.scala
ktoso: Yes, definitely - I'll add that.

ktoso: I didn't quite understand the concern about full materialization name - by "full" you mean the `s"$namePrefix-${nextFlowNameCount()}"` or something else?

patriknw: yes, I would be careful with using names with the counters for the LoggingAdapter, because I think slf4j implementations cache all those names in its logger hierarchy (and here it could potentially be many many such names)

akka-stream/src/main/scala/akka/stream/scaladsl/OperationAttributes.scala
patriknw: I agree, it would be better if we could pass this in some other way. Didn't Roland add the materializer to some context for the async stage? Perhaps that approach can be extended?

ktoso:
Now with:
* exposed actor system in ActorFlowMaterializer - akka private api
* logging provider created inside stage
* allows customizing which message should be logged at which log-level, via op attributes. Should be pretty useful
* logging adapter can be provided by user

TODO possibly got names wrong / and which attrs I pass down.

rkuhn:
please rename PR description so it invites reviews :-)

ktoso:
Yup done, and fixing the validation failure in a sec :)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/pattern/FlowDebugSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+      //      "debug each element" in {
+      //        val p = logProbe()
+      //
+      //        val debugging = Flow[Int].log("my-debug")
+      //        Source(1 to 2).via(debugging).runWith(Sink.ignore)
+      //
+      //        p.expectMsg(Logging.Debug(LogSrc, LogClazz, "[my-debug] Element: 1"))
+      //        p.expectMsg(Logging.Debug(LogSrc, LogClazz, "[my-debug] Element: 2"))
+      //        p.expectMsg(Logging.Debug(LogSrc, LogClazz, "[my-debug] Upstream finished."))
+      //      }
```
patriknw: ?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/pattern/FlowDebugSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+      }
+
+      "allow configuring log levels via OperationAttributes" in {
+        val p = logProbe()
+
+        val logAttrs = OperationAttributes.logLevels(
+          onElement = Logging.WarningLevel,
+          onFinish = Logging.InfoLevel,
+          onFailure = Logging.DebugLevel)
+        println("logLevels = " + logAttrs.logLevels.get.onElement)
```
patriknw: println?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/pattern/FlowDebugSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+        p.expectMsg(Logging.Debug(src, clazz, "[flow-5] Upstream finished."))
+      }
+
+      "allow configuring log levels via OperationAttributes" in {
+        val p = logProbe()
+
+        val logAttrs = OperationAttributes.logLevels(
+          onElement = Logging.WarningLevel,
+          onFinish = Logging.InfoLevel,
+          onFailure = Logging.DebugLevel)
```
patriknw: nice

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
```diff
@@ -104,6 +104,12 @@ private[stream] object Stages {
     override protected def newInstance: StageModule = this.copy()
   }
 
+  // TODO decide on name, is it Logging or Debugging
+  final case class DebugLogging(name: String, extract: Any ⇒ Any, log: Option[LoggingAdapter], attributes: OperationAttributes = map) extends StageModule {
```
patriknw: I guess it is `Logging` (or `Log`) now

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -491,3 +495,51 @@ private[akka] final case class MapAsyncUnordered[In, Out](parallelism: Int, f: I
     if (todo > 0) ctx.absorbTermination()
     else ctx.finish()
 }
+
+/**
+ * INTERNAL API
+ */
+private[akka] final case class DebugLogging[T](name: String, extract: T ⇒ Any, userLogAdapter: Option[LoggingAdapter]) extends PushStage[T, T] {
```
patriknw: s/userLogAdapter/logAdapter/ ?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -491,3 +495,51 @@ private[akka] final case class MapAsyncUnordered[In, Out](parallelism: Int, f: I
...
+
+  private final val DefaultLogLevels = LogLevels(onElement = Logging.DebugLevel, onFinish = Logging.DebugLevel, onFailure = Logging.ErrorLevel)
```
patriknw: put in companion

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
ktoso: Maybe `Log` would be nice, wouldn't conflict with `akka.event.Logging`

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -491,3 +495,51 @@ private[akka] final case class MapAsyncUnordered[In, Out](parallelism: Int, f: I
...
+  }
+
+  /** Idempotent initialization of logger and log levels */
+  def initLogger(ctx: Context[T]): Unit =
+    if (log == null) {
+      logLevels = ctx.attributes.logLevels.getOrElse(DefaultLogLevels)
+
+      log = userLogAdapter getOrElse {
+        val sys = ctx.materializer.asInstanceOf[ActorFlowMaterializerImpl].system
+        Logging(sys, this.getClass)
```
patriknw: will that be a logger name? akka.stream.impl.fusing..?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: right, thanks :)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: yup +1

akka-stream-tests/src/test/scala/akka/stream/scaladsl/pattern/FlowDebugSpec.scala
ktoso: whoops, removed

akka-stream-tests/src/test/scala/akka/stream/scaladsl/pattern/FlowDebugSpec.scala
ktoso: Enabled again, no changes there

patriknw:
don't forget to add javadsl

patriknw:
LGTM, aside from nitpicks

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: Yeah, exactly. Here's full debug event example:
`Debug(Log(akka://FlowDebugSpec),class akka.stream.impl.fusing.Log,[pre-map] Element: 1)`

patriknw: Is it possible to amend that to something more public/stable, e.g. "akka.stream.Log"?
It is used in slf4j config.

ktoso: @patriknw (sorry for the delay). Yes, we can use a String like "akka.stream.Log" to hardcode this.

The event then contains a dummy class as source class which got me worried for a second: `Warning(akka.stream.Log(akka://FlowLogSpec),class akka.event.DummyClassForStringSources`, but I checked how that's used and it seems all good :-)

I'll document that it's logged under this class.

patriknw: sounds good

ktoso:
Added Java DSL, and cleaned up things.

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
```diff
@@ -52,6 +52,7 @@ object ActorFlowMaterializer {
     val system = actorSystemOf(context)
 
     new ActorFlowMaterializerImpl(
+      system,
```
ktoso: passed through explicitly now (previously internally did casting/hacking)

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
ktoso: fixed

ktoso:
Since we now have built in logging, see https://github.com/akka/akka/pull/17162

It would now also be possible to replace the custom errorLogger HTTP uses, like this: https://github.com/ktoso/akka/commit/4fbb610b02c3b10c6afe76de541f53a8f7143ad0

Using `ZeroLevel` to avoid slowing down the normal pass-elements-through case would be advised ofc.

ktoso:
Requesting another LGTM :)

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
patriknw: right, good

patriknw: it should also be an abstract method in `abstract class ActorFlowMaterializer`

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -491,3 +495,61 @@ private[akka] final case class MapAsyncUnordered[In, Out](parallelism: Int, f: I
...
+    super.onDownstreamFinish(ctx)
+  }
+
+  /** Idempotent initialization of logger and log levels */
+  def initLogger(ctx: Context[T]): Unit =
+    if (log == null) {
+      logLevels = ctx.attributes.logLevels.getOrElse(DefaultLogLevels)
+
+      log = logAdapter getOrElse {
+        val sys = ctx.materializer.asInstanceOf[ActorFlowMaterializerImpl].system
```
patriknw: here it should work with `asInstanceOf[ActorFlowMaterializer]`

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -491,3 +495,61 @@ private[akka] final case class MapAsyncUnordered[In, Out](parallelism: Int, f: I
...
+  }
+
+  override def onUpstreamFinish(ctx: Context[T]): TerminationDirective = {
+    initLogger(ctx) // TODO if we have preStart this could be called in it, instead of in all callbacks
+
+    log.log(logLevels.onFinish, "[{}] Upstream finished.", name)
+    super.onUpstreamFinish(ctx)
+  }
+
+  override def onDownstreamFinish(ctx: Context[T]): TerminationDirective = {
```
ktoso: I'm also adding `OffLevel` support here - seen it used in http, where they only want to log failures for example.

patriknw:
javadsl looks good

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: done, thanks

ktoso:
Pushed with latest updates, OperationAttributes was not usable from Java nicely. I'd like to merge this PR only after rebasing on top of the OpAttrs unification.

ktoso:
// Note to self, rebase once https://github.com/akka/akka/pull/17193 merged.

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -618,6 +626,7 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]],
 
     override def incomingBall = 0
 
+    override def attributes: OperationAttributes = OneBoundedInterpreter.this.attributes
```
rkuhn: why not put this into `State` just once?

akka-stream/src/main/scala/akka/stream/javadsl/japi/WithVariance.scala
```diff
@@ -6,6 +6,15 @@ package akka.stream.javadsl.japi
 // TODO Same SAM-classes as in akka.japi, but with variance annotations
 // TODO Remove these in favour of using akka.japi with added variance
 
+object Function {
+  final val Identity = new Function[Any, Any] {
```
rkuhn: use `[Any, Nothing]`, then no cast is needed

rkuhn: (and of course make this value private)

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
```diff
@@ -571,6 +573,9 @@ sealed trait Context[Out] {
    * It can be used to materialize sub-flows.
    */
   def materializer: FlowMaterializer
+
+  /** Returns operation attributes associated with the this Stage */
+  def attributes: OperationAttributes
```
rkuhn: good! we’ll need to invest some effort to make things work when actually fusing stages, but this API will stay the same

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -491,3 +500,78 @@ private[akka] final case class MapAsyncUnordered[In, Out](parallelism: Int, f: I
...
+  private var log: LoggingAdapter = _
+  private var logLevels: OperationAttributes.LogLevels = _
+
+  @inline private def isEnabled(ctx: Context[T], l: LogLevels ⇒ LogLevel): Boolean = {
+    if (logLevels eq null) logLevels = ctx.attributes.logLevels.getOrElse(DefaultLogLevels)
+    l(logLevels).asInt != OffInt
+  }
+
+  override def onPush(elem: T, ctx: Context[T]): SyncDirective = {
+    if (isEnabled(ctx, _.onElement)) {
```
rkuhn: this is too expensive, unfortunately we cannot use function literals like that

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -491,3 +500,78 @@ private[akka] final case class MapAsyncUnordered[In, Out](parallelism: Int, f: I
...
+      initLogger(ctx) // TODO if we have preStart this could be called in it, instead of in all callbacks
```
rkuhn: okay, convinced: you have my leave to add `preStart` and `postStop` (and replace initAsyncInput)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -491,3 +500,78 @@ private[akka] final case class MapAsyncUnordered[In, Out](parallelism: Int, f: I
...
+ */
+private[akka] final case class Log[T](name: String, extract: T ⇒ Any, logAdapter: Option[LoggingAdapter]) extends PushStage[T, T] {
+
+  import Log._
+
+  private var log: LoggingAdapter = _
+  private var logLevels: OperationAttributes.LogLevels = _
+
+  @inline private def isEnabled(ctx: Context[T], l: LogLevels ⇒ LogLevel): Boolean = {
+    if (logLevels eq null) logLevels = ctx.attributes.logLevels.getOrElse(DefaultLogLevels)
```
rkuhn: with this then also in `preStart` this whole method basically goes away, right?

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
```diff
@@ -139,6 +138,11 @@ abstract class ActorFlowMaterializer extends FlowMaterializer {
    */
   private[akka] def actorOf(props: Props, name: String): ActorRef
 
+  /**
+   * INTERNAL API
+   */
+  private[akka] def system: ActorSystem
```
rkuhn: would it also be interesting to have this materializer’s “guardian” actorRef available so that the logged names are unique?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: Great, that will simplify a lot!

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLogSpec.scala
```diff
@@ -0,0 +1,212 @@
...
+
+class FlowLogSpec extends AkkaSpec("akka.loglevel = DEBUG") with ScriptedTest {
+
+  implicit val mat: FlowMaterializer = ActorFlowMaterializer()
+
+  def logProbe(): TestProbe = {
+    val p = TestProbe()
+    system.eventStream.subscribe(p.ref, classOf[Logging.LogEvent])
+    p
+  }
```
rkuhn: why not have just one `val logProbe`?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: Yes exactly - I can just properly init the state of the stage then.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLogSpec.scala
ktoso: no good reason, you're right val may be better here (could expose if more msgs sent than I expected)

rkuhn:
good stuff! apply some more polish and it really shines :-)

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
ktoso: done, thanks :)

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
ktoso: indeed, thanks!

akka-stream/src/main/scala/akka/stream/javadsl/japi/WithVariance.scala
ktoso: Hm, while in theory it should be fine I can't seem to get it to compile (`param: Any): Nothing = param` => nope)

rkuhn: ah, stupid me; but should still be private :-)

ktoso: yes :)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: Added `preStart` and `postStop` :)

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
ktoso: So just expose the guardian here? (Didn't yet)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLogSpec.scala
ktoso: done

ktoso:
+ Applied more polish polish flag 
+ Added `preStart` and `postStop` callbacks as well
? Make cookbook show how to use custom LoggingAdapter - not sure if we want to show this / remove this / or keep the PushStage in the cookbook?

ktoso:
failure was due to merge conflict on jenkins, rebased now.

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
```diff
@@ -22,7 +22,8 @@ import scala.concurrent.{ Await, ExecutionContextExecutor }
 /**
  * INTERNAL API
  */
-private[akka] case class ActorFlowMaterializerImpl(override val settings: ActorFlowMaterializerSettings,
+private[akka] case class ActorFlowMaterializerImpl(override val system: ActorSystem,
+                                                   override val settings: ActorFlowMaterializerSettings,
                                                    dispatchers: Dispatchers,
```
patriknw: dispatchers param not needed, it is `system.dispatchers`

ktoso: good catch, thanks!

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -124,11 +125,12 @@ private[akka] object OneBoundedInterpreter {
 private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]],
                                           onAsyncInput: (AsyncStage[Any, Any, Any], AsyncContext[Any, Any], Any) ⇒ Unit,
                                           materializer: FlowMaterializer,
+                                          attributes: OperationAttributes = OperationAttributes.none,
```
drewhk: I am not convinced about this. Why would the interpreter need to know about attributes? The stages need to know, but I don't like coupling the interpreter now to yet another dependency.

drewhk: Why not just take the LifecylceContext itself?

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -630,16 +639,27 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]],
       (pipeline(op): Any) match {
         case b: BoundaryStage ⇒
           b.context = new EntryState("boundary", op)
+          b.preStart(LifecycleContext)
```
drewhk: try catch? This should generate onFailure for the stage, not the interpreter, I think.

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -630,16 +639,27 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]],
...
+        case s: UntypedOp ⇒
+          s.context = new EntryState("initializing", op)
+          s.preStart(LifecycleContext)
       }
       op += 1
     }
   }
 
+  private def finishCurrentOp(): Unit = {
+    currentOp.postStop(LifecycleContext)
```
drewhk: try catch. Add test cases that throw on preStart or postStop in InterpreterSpec

ktoso: ah yes, good call - thanks

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
ktoso: agree completely, sadly I wasn't able to get it working yet. Would be great to get some help from you when you're back (monday?).

My modifications were mostly focused on setting the context to Failed. But then I wasn't sure where to make it run to propagate the error. I guess I want to do this after all stages have been initialised, and then propagate the onError from my failed stage right away. I wasn't able to grasp how to express this in the interpreter sadly - so your insights would be very helpful here :-)

ktoso:
Reverted to before `preStart` version, squashed and pushed. Please have a last look before we merge.

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
rkuhn: please open a ticket for further refinement

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
rkuhn: We will fix this when implementing the fusing (which is the point at which this solution will stop working anyway).

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -630,16 +633,23 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]],
...
+
         case a: AsyncStage[Any, Any, Any] @unchecked ⇒
           a.context = new EntryState("async", op)
           activeOpIndex = op
-          a.initAsyncInput(a.context)
-        case _ ⇒
+          a.initAsyncInput(a.context) // TODO remove asyncInput? it's like on Init
+
+        case s: UntypedOp ⇒
+          s.context = new EntryState("initializing", op)
```
rkuhn: this is a left-over of preStart

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
```diff
@@ -630,16 +633,23 @@ private[akka] class OneBoundedInterpreter(ops: Seq[Stage[_, _]],
...
       }
       op += 1
     }
   }
 
+  private def finishCurrentOp(): Unit = {
+    pipeline(activeOpIndex) = Finished.asInstanceOf[UntypedOp]
+  }
```
rkuhn: but this is good

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -849,6 +852,28 @@ trait FlowOps[+Out, +Mat] {
...
+   *    .withAttributes(OperationAttributes.logLevels(
+   *      onElement = Logging.InfoLevel,
+   *      onFinish = Logging.InfoLevel,
+   *      onFailure = Logging.InfoLevel))
+   * }}}
+   *
+   * Uses implicit [[LoggingAdapter]] if available, otherwise uses an internally created one,
+   * which uses `akka.stream.Log` as it's source (use this class to configure slf4j loggers).
+   */
+  def log(name: String, extract: Out ⇒ Any = identity)(implicit log: LoggingAdapter = null): Repr[Out, Mat] =
```
rkuhn: does this use the right `identity`?

rkuhn:
LGTM (skipped reviewing the actual implementation, concentrated on the API)

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
ktoso: For including guardian: https://github.com/akka/akka/issues/17298
For using preStart: https://github.com/akka/akka/issues/17299

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -593,6 +594,57 @@ class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Graph
...
+   * Logs elements flowing through the stream as well as completion and erroring.
+   *
+   * By default element and completion signals are logged on debug level, and errors are logged on Error level.
+   * This can be adjusted according to your needs by providing a custom [[OperationAttributes.LogLevels]] atrribute on the given Flow:
+   *
+   * The `extract` function will be applied to each element before logging, so it is possible to log only those fields
+   * of a complex object flowing through this element.
+   *
+   * Uses the given [[LoggingAdapter]] for logging.
+   */
```
drewhk: Add the same information as in the other scaladocs about Emits when, Backpressures when, Completes when. Add the log stage to the summary table.

drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
ktoso: Oh, right - here I can use a cached instance

patriknw:
LGTM

akka-stream/src/main/scala/akka/stream/impl/fusing/Interpreter.scala
ktoso: good catch, thanks!

ktoso:
Please don't merge yet - must address this comment on docs: https://github.com/akka/akka/pull/17162#discussion_r29042803

ktoso:
Addressed all comments, merging.

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
```diff
@@ -132,6 +132,11 @@ private[stream] object Stages {
     override protected def newInstance: StageModule = this.copy()
   }
 
+  final case class Log(name: String, extract: Any ⇒ Any, loggingAdapter: Option[LoggingAdapter], attributes: OperationAttributes = map) extends StageModule {
```
hepin1989: With ```attributes = map```?Shouldn't it be something like ```log```?

ktoso: Ouch, great catch; would you submit a PR?
Thanks a lot for catching these!

Issue #16957, opened 2015-02-27
Rethink actor creation API exposed by ActorFlowMaterializer


drewhk:
Currently we expose the actorOf() method that takes a Props and a name. Unfortunately the materializer cannot set the correct dispatcher without having the current effectiveAttributes availalbe. Apart from that, I am not sure if we should take a name parameter at all, it should be encoded by the attributes itself. Probably the solution is to expose an actorOf that expects only a Props and attributes, and calculates correct name and dispatcher automatically. This also means that Source/Sink create methods should take the effectiveAttributes as parameters, otherwise they cannot pass it to actorOf().

drewhk:
So, thinking it further, it seems suboptimal to pass down the effectiveAttributes to the create() call only because it needs to be passed back to the actorOf method. What we probably need is to expose a different interface instead of ActorFlowMaterializer, which wraps the underlying session, and only accepts a Props in its only actorOf() method. The rest should be calculated automatically by the materializer session.

Issue #16952, opened 2015-02-26
PropsSource overrides Props dispatcher with the one in Materializer Settings


TrustNoOne:
Hi,
as pointed out in the code by @2m on the akka user list, PropsSource calls materializer.actorOf when creating the ActorProducer, overriding any custom dispatcher set in the props object. 

The only way to use a different dispatcher is ping-ponging the stream objects to an actor outside the actor flow.

this  is where PropsSource creates the actor:
https://github.com/akka/akka/blob/38d8f265dad3909f3d597f4c3c9da65a7bf576d2/akka-stream/src/main/scala/akka/stream/impl/Sources.scala#L166

This is where the materializer overrides the dispatcher:
https://github.com/akka/akka/blob/38d8f265dad3909f3d597f4c3c9da65a7bf576d2/akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala#L157

Thanks

drewhk:
Since this is closed: https://github.com/akka/akka/issues/16902 it is possible to properly set attributes on sources. The real question is whether it makes sense to override the dispatcher on a Props from stream settings. The solution would be that if the Props has an explicit dispatcher then it should be taken, otherwise it should fall back to the attributes inherited from the stream it is embedded in.

TrustNoOne:
I can tell you my use case: an ActorProducer that acts as a source but does blocking operations (e.g. reading from a database) thus needing a different dispatcher.

drewhk:
Yes, that is a perfectly valid use-case.

drewhk:
This needs to be tackled first: https://github.com/akka/akka/issues/16957

PR #17175, opened 2015-04-10
=str #16957 Refactor actor creation api for sinks and sources

patriknw:
- also solves #16952, dispatcher for ActorPublisher

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
+++ b/akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
@@ -14,6 +14,7 @@ import akka.actor.Props
 import akka.actor.ActorRef
 import akka.stream.javadsl.japi
 import scala.concurrent.ExecutionContextExecutor
+import akka.stream.scaladsl.OperationAttributes
 
 object ActorFlowMaterializer {
 
@@ -121,6 +122,16 @@ object ActorFlowMaterializer {
     system
   }
 
+  /**
+   * INTERNAL API
+   */
+  private[akka] def downcast(materializer: FlowMaterializer): ActorFlowMaterializer =
+    materializer match {
+      case m: ActorFlowMaterializer ⇒ m
+      case _ ⇒ throw new IllegalArgumentException(s"required [${classOf[ActorFlowMaterializer].getName}] " +
+        s"but got [${materializer.getClass.getName}]")
+    }
+
 }
 
 /**
@@ -134,10 +145,12 @@ abstract class ActorFlowMaterializer extends FlowMaterializer {
 
   def settings: ActorFlowMaterializerSettings
 
+  def effectiveSettings(opAttr: OperationAttributes): ActorFlowMaterializerSettings
+
   /**
-   * INTERNAL API
+   * INTERNAL API: this might become public later
    */
-  private[akka] def actorOf(props: Props, name: String): ActorRef
+  private[akka] def actorOf(context: MaterializationContext, props: Props): ActorRef
 
 }
 
--- a/akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
@@ -19,17 +19,6 @@ import org.reactivestreams._
 
 import scala.concurrent.{ Await, ExecutionContextExecutor }
 
-object ActorFlowMaterializerImpl {
-  import OperationAttributes._
-  private[akka] def calcSettings(opAttr: OperationAttributes)(settings: ActorFlowMaterializerSettings): ActorFlowMaterializerSettings =
-    opAttr.attributes.collect {
-      case InputBuffer(initial, max) ⇒ (s: ActorFlowMaterializerSettings) ⇒ s.withInputBuffer(initial, max)
-      case Dispatcher(dispatcher) ⇒ (s: ActorFlowMaterializerSettings) ⇒ s.withDispatcher(dispatcher)
-      case SupervisionStrategy(decider) ⇒ (s: ActorFlowMaterializerSettings) ⇒
-        s.withSupervisionStrategy(decider)
-    }.reduceOption(_ andThen _).getOrElse((x: ActorFlowMaterializerSettings) ⇒ x)(settings) // FIXME is this the optimal way of encoding this?
-}
-
 /**
  * INTERNAL API
  */
@@ -43,11 +32,23 @@ private[akka] case class ActorFlowMaterializerImpl(override val settings: ActorF
   import ActorFlowMaterializerImpl._
   import akka.stream.impl.Stages._
 
-  def withNamePrefix(name: String): FlowMaterializer = this.copy(namePrefix = name)
+  override def withNamePrefix(name: String): FlowMaterializer = this.copy(namePrefix = name)
 
   private[this] def nextFlowNameCount(): Long = flowNameCounter.incrementAndGet()
 
   private[this] def createFlowName(): String = s"$namePrefix-${nextFlowNameCount()}"
+
+  override def effectiveSettings(opAttr: OperationAttributes): ActorFlowMaterializerSettings = {
+    import OperationAttributes._
+    opAttr.attributes.foldLeft(settings) { (s, attr) ⇒
+      attr match {
+        case InputBuffer(initial, max)    ⇒ s.withInputBuffer(initial, max)
+        case Dispatcher(dispatcher)       ⇒ s.withDispatcher(dispatcher)
+        case SupervisionStrategy(decider) ⇒ s.withSupervisionStrategy(decider)
+        case Name(_)                      ⇒ s
+      }
+    }
+  }
 
   override def materialize[Mat](runnableFlow: Graph[ClosedShape, Mat]): Mat = {
     runnableFlow.module.validate()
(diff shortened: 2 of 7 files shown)


patriknw:
Refs #16957

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
```diff
@@ -19,15 +19,40 @@ import org.reactivestreams._
...
+    val effectiveAttributes: OperationAttributes,
+    val effectiveSettings: ActorFlowMaterializerSettings,
+    val stageName: String,
+    materializerImpl: ActorFlowMaterializerImpl) {
+
+    def materializer: ActorFlowMaterializer = materializerImpl
+
+    def actorOf(props: Props): ActorRef = {
+      val dispatcher =
+        if (props.dispatcher == Dispatchers.DefaultDispatcherId) effectiveSettings.dispatcher
```
drewhk: you should test for props.dispatcher == NoDispatcherGiven instead. (Look at Props.dispatcher)

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -4,22 +4,21 @@
...
 import scala.concurrent.{ Future, Promise }
+import akka.stream.impl.ActorFlowMaterializerImpl.CreateContext
 
 /**
  * INTERNAL API
  */
 private[akka] abstract class SinkModule[-In, Mat](val shape: SinkShape[In]) extends Module {
 
-  def create(materializer: ActorFlowMaterializerImpl, flowName: String): (Subscriber[In] @uncheckedVariance, Mat)
+  def create(context: CreateContext): (Subscriber[In] @uncheckedVariance, Mat)
```
drewhk: this is still tied to ActorFlowMatImpl, but at least it paves the way for generalization

drewhk:
Yes, this is what I imagined. We can make later CreateContext a trait and have an ActorCreateContext for the actorOf method.

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
patriknw: I thought so too, but actually not:

    def dispatcher: String = deploy.dispatcher match {
      case NoDispatcherGiven ⇒ Dispatchers.DefaultDispatcherId
      case x                 ⇒ x
    }

drewhk: well, you are right (I did this before), check deploy.dispatcher

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
patriknw: yes, I was thinking how far I should go, but stayed here. As far as I can see we don't really support custom sources and sinks anyway, or did I miss anything?

I can make CreateContext independent of actor if you like, but then I think the actorOf method should still be on ActorFlowMaterializer, taking the CreateContext and the Props as parameters.
The source/sink have to downcast to ActorFlowMaterializer when needed.

I think that would be better than CreateContext and ActorCreateContext.

I'll give it a quick try.

drewhk: no, we should leave it as it is for now. This is enough to allow us later to pull it more apart (that is what I meant to say).

patriknw:
@drewhk I implemented the alternative, anyway, compare and pick. 5fd0cb7

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
```diff
@@ -166,13 +176,17 @@ private[akka] case class ActorFlowMaterializerImpl(override val settings: ActorF
...
+  override lazy val executionContext: ExecutionContextExecutor = dispatchers.lookup(settings.dispatcher match {
     case Deploy.NoDispatcherGiven ⇒ Dispatchers.DefaultDispatcherId
     case other                    ⇒ other
   })
 
-  private[akka] def actorOf(props: Props, name: String): ActorRef =
-    actorOf(props, name, settings.dispatcher)
+  override def actorOf(context: MaterializationContext, props: Props): ActorRef = {
+    val dispatcher =
+      if (props.dispatcher == Dispatchers.DefaultDispatcherId) context.effectiveSettings.dispatcher
```
drewhk: props.deploy.dispatcher == NoDispatcherGiven

If a Props is using the default dispatcher explicitly then we should not override it

drewhk:
I like the second approach better, actually, but I guess others might have a different opinion.

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
patriknw: ok, good point, will change to that

patriknw:
I also like the second better, if that was not clear :)

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
```diff
@@ -121,6 +122,16 @@ object ActorFlowMaterializer {
     system
   }
 
+  /**
+   * INTERNAL API
+   */
+  private[akka] def downcast(materializer: FlowMaterializer): ActorFlowMaterializer =
```
ktoso: good idea :)

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
```diff
@@ -121,6 +122,16 @@ object ActorFlowMaterializer {
...
+    materializer match {
+      case m: ActorFlowMaterializer ⇒ m
+      case _ ⇒ throw new IllegalArgumentException(s"required [$classOf[ActorFlowMaterializer.getName]" +
```
drewhk: why the classOf[]?

patriknw: that should be `[${classOf[ActorFlowMaterializer].getName}]`

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
```diff
@@ -134,10 +145,12 @@ abstract class ActorFlowMaterializer extends FlowMaterializer {
...
   def settings: ActorFlowMaterializerSettings
 
+  def effectiveSettings(opAttr: OperationAttributes): ActorFlowMaterializerSettings
+
   /**
-   * INTERNAL API
+   * INTERNAL API: this might become public later
    */
-  private[akka] def actorOf(props: Props, name: String): ActorRef
+  private[akka] def actorOf(context: MaterializationContext, props: Props): ActorRef
```
ktoso: No option to provide name? I found `s"$flowName-whichExactlySinkIAm"` useful when looking at "which actor didnt handle a message" when developing things here etc

patriknw: the name should come from the OperationAttributes, which is in the `context`

ktoso: Okey, makes sense :)

ktoso:
LGTM

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
```diff
@@ -43,12 +32,22 @@ private[akka] case class ActorFlowMaterializerImpl(override val settings: ActorF
...
   private[this] def createFlowName(): String = s"$namePrefix-${nextFlowNameCount()}"
 
+  override def effectiveSettings(opAttr: OperationAttributes): ActorFlowMaterializerSettings = {
+    import OperationAttributes._
+    opAttr.attributes.collect {
+      case InputBuffer(initial, max) ⇒ (s: ActorFlowMaterializerSettings) ⇒ s.withInputBuffer(initial, max)
+      case Dispatcher(dispatcher) ⇒ (s: ActorFlowMaterializerSettings) ⇒ s.withDispatcher(dispatcher)
+      case SupervisionStrategy(decider) ⇒ (s: ActorFlowMaterializerSettings) ⇒
+        s.withSupervisionStrategy(decider)
+    }.reduceOption(_ andThen _).getOrElse((x: ActorFlowMaterializerSettings) ⇒ x)(settings) // FIXME is this the optimal way of encoding this?
```
2m: Fold should get rid of this fixme:
``` scala
    opAttr.attributes.foldLeft(settings) { (s, attr) => attr match {
        case InputBuffer(initial, max) ⇒ s.withInputBuffer(initial, max)
        case Dispatcher(dispatcher) ⇒ s.withDispatcher(dispatcher)
        case SupervisionStrategy(decider) ⇒ s.withSupervisionStrategy(decider)
        case Name(_) ⇒ s
      }
    }
```

2m:
LGTM

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
patriknw: ok, I'll fix that

patriknw:
squashed, will merge after pr validation

PR #17177, opened 2015-04-11
Add spec for extractUri directive

vikraman:
I was trying to understand the behavior of `request.uri` for an akka-http server behind a reverse proxy. This is a test case to clarify the same.

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/server/directives/BasicDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/server/directives/BasicDirectivesSpec.scala
@@ -5,6 +5,9 @@
 package akka.http.server
 package directives
 
+import akka.http.model._
+import headers._
+
 class BasicDirectivesSpec extends RoutingSpec {
 
   "The `mapUnmatchedPath` directive" should {
@@ -24,6 +27,24 @@ class BasicDirectivesSpec extends RoutingSpec {
           echoComplete
         }
       } ~> check { responseAs[String] shouldEqual "GET" }
+    }
+  }
+
+  "The `extractUri` directive" should {
+    "extract from the request url" in {
+      Get("https://example.com/foo") ~> {
+        extractUri { echoComplete }
+      } ~> check { responseAs[String] shouldEqual "https://example.com/foo" }
+    }
+    "extract from a X-Request-URI header" in {
+      Get() ~> addHeader(RawHeader("x-request-uri", "http://example.com/foo")) ~> {
+        extractUri { echoComplete }
+      } ~> check { responseAs[String] shouldEqual "http://example.com/" }
+    }
+    "extract from request path and a Host header" in {
+      Get("/foo/bar") ~> addHeader(`Host`("example.com")) ~> {
+        extractUri { echoComplete }
+      } ~> check { responseAs[String] shouldEqual "http://example.com/foo/bar" }
     }
   }
 }


typesafehub-validator:
Hi @vikraman,

ktoso:
OK TO TEST

akka-http-tests/src/test/scala/akka/http/server/directives/BasicDirectivesSpec.scala
```diff
@@ -26,4 +29,22 @@ class BasicDirectivesSpec extends RoutingSpec {
...
+  "The `extractUri` directive" should {
+    "extract from the request url" in {
+      Get("https://example.com/foo") ~> {
+        extractUri { echoComplete }
+      } ~> check { responseAs[String] shouldEqual "https://example.com/foo" }
+    }
+    "extract from a X-Request-URI header" in {
+      Get() ~> addHeader(RawHeader("x-request-uri", "http://example.com/foo")) ~> {
+        extractUri { echoComplete }
+      } ~> check { responseAs[String] shouldEqual "http://example.com/" }
```
rkuhn: where did the `foo` go? (please excuse my ignorance but quick googling did not turn up proper specs for this feature)

rkuhn:
@sirthias @jrudolph ?

akka-http-tests/src/test/scala/akka/http/server/directives/BasicDirectivesSpec.scala
vikraman: I am also confused about this, and I don't know what should be the correct behavior.

jrudolph: akka-http doesn't know anything about the `X-Request-URI` header and just ignores it. So, there's not really much to see here. The directive itself isn't tested because it just return the `uri` field of the `HttpRequest`. How this uri field is composed is tested in lower levels (but maybe not sufficiently documented).

jrudolph:
> I was trying to understand the behavior of request.uri for an akka-http server behind a reverse proxy.

As said above, the `X-Request-URI` header isn't supported out of the box because the proxy behavior to add this specific header isn't standardized AFAIK (if you have better information please let us know so we can reconsider).

Rewriting the request transparently using the header's value would lose information, so that's probably not the way to go. I would suggest building a simple custom directive, instead, for which you exactly implement the behavior you need.

akka-http-tests/src/test/scala/akka/http/server/directives/BasicDirectivesSpec.scala
vikraman: Can you point me to how exactly the header value ends up in uri? I couldn't figure it out.

vikraman:
Yes, I also couldn't find a spec for `X-Request-URI`, I am currently handling it with a custom directive. Thanks for the help.

Closing this pull request since it isn't really useful.

rkuhn:
Au contraire, @vikraman, I found this useful even though it did not result in a code change!

akka-http-tests/src/test/scala/akka/http/server/directives/BasicDirectivesSpec.scala
sirthias: The header value *doesn't* actually make it into the response.
The test will still be green when you completely remove the `addHeader(...)`.

Issue #16209, opened 2014-11-03
Introduce unbuffered Zip version


drewhk:
If Zip is used for timing purposes an unbuffered version is needed. Basically the same Impl just buffer size set to 1 independently of materializer settings.

rkuhn:
We should just have factory methods for Zip for both purposes—both are zipping things.

viktorklang:
@rkuhn isn't it better to force to put a buffer in front of it if you want buffered behavior?

drewhk:
Not exactly the same. The size of internal buffers set the maximum request() that is possible to send. For larger buffers request() traffic can be amortized more, less message sends. The Flow lewel buffer() element does not know about the RS API so even fusing it to the zip would not help (well, unless we detect that it is a buffer and as a special case we unfold it to an RS aware buffer).

patriknw:
We decided to not add any new factory methods for this (at least not now).

PR #17183, opened 2015-04-13
+str #16209 Add convenience for unbuffered zip

patriknw:
- also add missing OperationAttributes param for ZipWith

DOES NOT COMPILE

Would like input before fighting this further. Might be going in wrong direction.
I noticed that OperationAttributes parameter was missing for `ZipWith` and that is not easy to add.
I have tried various types of overloads but it currently fails with:

```
val zip = b.add(ZipWith[Int, Int, Int](_ + _))

missing parameter type for expanded function ((x$3: <error>, x$4) ⇒ x$3.$plus(x$4))
missing parameter type for expanded function ((x$3, x$4) ⇒ x$3.$plus(x$4))
```

### DIFF
--- a/akka-stream/src/main/boilerplate/akka/stream/javadsl/ZipWith.scala.template
+++ b/akka-stream/src/main/boilerplate/akka/stream/javadsl/ZipWith.scala.template
@@ -12,18 +12,66 @@ object ZipWith {
    * Create a new `ZipWith` vertex with the specified input types and zipping-function `f`.
    *
    * @param f zipping-function from the input values to the output value
-   * @param attributes optional attributes for this vertex
    */
   def create[A, B, Out](f: japi.Function2[A, B, Out]): Graph[FanInShape2[A, B, Out], Unit] =
-    scaladsl.ZipWith(f.apply _)
+    create(OperationAttributes.none, f)
+    
+  /**
+   * Create a new `ZipWith` vertex with the specified input types and zipping-function `f`.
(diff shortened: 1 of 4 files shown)


patriknw:
I was also thinking of removing OperationAttribute altogether from all junction factories and instead add `def withAttributes` to `Graph`, but I did not try it.

rkuhn:
The overload won’t work, function literal input types are only inferred if the method is not overloaded(*). I’d recommend going with `.named`/`.withAttributes` for consistency with Flows etc.

(*) Strictly speaking the argument list the contains the function argument must be unique, i.e. you could get around it by having an empty list before the function argument.

drewhk:
Unbuffered is a bitmisleading since there is always a buffer of size 1

patriknw:
@rkuhn thanks, I'll try that path instead

@drewhk um, I'm not fully convinced we need this convenience at all. Isn't this good enough:

    import OperationAttributes._
    Zip(...).withAttributes(inputBuffer(1))

drewhk:
This addition was discussed because of multiple ML questions regarding this. All I wanted is a simple wrapper on the normal build method of zip. I am not sure we should go overboard, I think having a simple SyncWith.apply would help a lot for only 2 parameters, documenting that it is just a convenience over ZipWith.withAttributes()

drewhk:
I am fine dropping this as well. I know how it works anyway :)

PR #17186, opened 2015-04-13
[STASHED] +str #17185 add Sink.simpleUdp()

### LINKED ISSUE #17185: Provide SimpleSender UDP Sink
I needed this in another project, and seems it should be universally useful as well, so PRing here.
Typical use case would be just dumping statistics or logs to some udp address for collection.

It's just the unconnected "simple sender" version. We could/should also provide the Connected one if we think it's valuable (maybe then as `udp(target, connected = true)`?).

ktoso:
Proposal of simple UDP Sink.

Resolves #17185

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/io/SimpleUdpSinkActor.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/io/SimpleUdpSinkActor.scala
@@ -1,0 +1,45 @@
+/**
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.impl.io
+
+import java.net.InetSocketAddress
+
+import akka.actor.{ ActorLogging, ActorRef, Props }
+import akka.io.{ IO, Udp }
+import akka.stream.actor.ActorSubscriberMessage.{ OnComplete, OnError, OnNext }
+import akka.stream.actor.{ ActorSubscriber, RequestStrategy, WatermarkRequestStrategy, ZeroRequestStrategy }
+import akka.util.ByteString
+
+/** INTERNAL API */
+private[akka] object SimpleUdpSinkActor {
+  def props(target: InetSocketAddress): Props =
+    Props(new SimpleUdpSinkActor(target))
+}
+
+/** INTERNAL API */
+private[akka] class SimpleUdpSinkActor(target: InetSocketAddress) extends ActorSubscriber with ActorLogging {
+
+  private var currentRequestStrategy: RequestStrategy = ZeroRequestStrategy
+  override def requestStrategy = currentRequestStrategy
+
+  import context.system
+
+  IO(Udp) ! Udp.SimpleSender
+
+  override def receive: Receive = {
+    case Udp.SimpleSenderReady ⇒
+      currentRequestStrategy = WatermarkRequestStrategy(10) // TODO arbitrary value, how can we do better here (needs dropping to low level)
+      context.become(ready(sender()))
+  }
+
+  def ready(send: ActorRef): Receive = {
+    case OnNext(bytes: ByteString) ⇒ send ! Udp.Send(bytes, target)
+    case OnComplete                ⇒ context.stop(self)
+
+    case OnError(cause) ⇒
+      log.error(cause, "Shutting down (targetted at: {})", target)
+      context.stop(self)
+
+  }
+}
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Sink.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Sink.scala
@@ -8,6 +8,7 @@ import akka.actor.Props
 import akka.stream.javadsl
 import akka.stream.scaladsl
 import akka.stream._
+import akka.util.ByteString
 import org.reactivestreams.Publisher
 import org.reactivestreams.Subscriber
 import scala.concurrent.Future
@@ -114,6 +115,12 @@ object Sink {
    */
   def actorSubscriber[T](props: Props): Sink[T, ActorRef] =
     new Sink(scaladsl.Sink.actorSubscriber(props))
+
+  /**
+   * Creates a `Sink` which uses Akka-IO's unconnected UDP mode to emit each ByteString it receives to the given target address.
+   */
+  def simpleUdp(host: String, port: Int): Sink[ByteString, Unit] =
+    new Sink(scaladsl.Sink.simpleUdp(host, port))
 
   /**
    * A graph with the shape of a sink logically is a sink, this method makes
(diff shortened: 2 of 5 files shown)


ktoso:
Failure was unrelated.

ktoso:
PLS BUILD

akka-stream/src/main/scala/akka/stream/impl/io/SimpleUdpSinkActor.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  IO(Udp) ! Udp.SimpleSender
+
+  override def receive: Receive = {
+    case Udp.SimpleSenderReady ⇒
+      currentRequestStrategy = WatermarkRequestStrategy(10) // TODO arbitrary value, how can we do better here (needs dropping to low level)
+      context.become(ready(sender()))
+  }
+
+  def ready(send: ActorRef): Receive = {
```
drewhk: rename send to something more meaningful

akka-stream/src/main/scala/akka/stream/impl/io/SimpleUdpSinkActor.scala
```diff
@@ -0,0 +1,45 @@
...
+  override def requestStrategy = currentRequestStrategy
+
+  import context.system
+
+  IO(Udp) ! Udp.SimpleSender
+
+  override def receive: Receive = {
+    case Udp.SimpleSenderReady ⇒
+      currentRequestStrategy = WatermarkRequestStrategy(10) // TODO arbitrary value, how can we do better here (needs dropping to low level)
+      context.become(ready(sender()))
```
drewhk: watch the actor and cancel if it goes away.

akka-stream/src/main/scala/akka/stream/javadsl/Sink.scala
```diff
@@ -116,6 +117,12 @@ object Sink {
     new Sink(scaladsl.Sink.actorSubscriber(props))
 
   /**
+   * Creates a `Sink` which uses Akka-IO's unconnected UDP mode to emit each ByteString it receives to the given target address.
+   */
+  def simpleUdp(host: String, port: Int): Sink[ByteString, Unit] =
```
drewhk: hmm, I would go with the connected one in this case. The unconnected can be a Sink[(host, port, ByteString), Unit].

drewhk:
Can you add a Source, too? I think it shouldn't be very hard.

akka-stream/src/main/scala/akka/stream/javadsl/Sink.scala
ktoso: Nice idea for signature of the unconnected one.
I wanted this one to become the connected one anyway :-)

akka-stream/src/main/scala/akka/stream/impl/io/SimpleUdpSinkActor.scala
```diff
@@ -0,0 +1,45 @@
...
+  private var currentRequestStrategy: RequestStrategy = ZeroRequestStrategy
+  override def requestStrategy = currentRequestStrategy
+
+  import context.system
+
+  IO(Udp) ! Udp.SimpleSender
+
+  override def receive: Receive = {
+    case Udp.SimpleSenderReady ⇒
+      currentRequestStrategy = WatermarkRequestStrategy(10) // TODO arbitrary value, how can we do better here (needs dropping to low level)
```
patriknw: it could use the max inputBuffer size from the `effectiveSettings` (OperationAttributes). That will be easier to grab when we have merged https://github.com/akka/akka/pull/17175

akka-stream/src/main/scala/akka/stream/impl/io/SimpleUdpSinkActor.scala
patriknw: `target`?

patriknw: `targetRef`?

patriknw:
regarding adding `Source` also; I'd prefer a separate ticket for that. We must focus on the essentials to reach RC1.

patriknw:
LGTM, cool

drewhk:
I disagree, if we add the send part then people will ask (rightly) for the receive part, too. It should not be hard to create a simple Source.

drewhk:
If we want to leave out something then it can be the unconnected API.

akka-stream/src/main/scala/akka/stream/impl/io/SimpleUdpSinkActor.scala
```diff
@@ -0,0 +1,45 @@
...
+      context.become(ready(sender()))
+  }
```
rkuhn: need to handle `OnComplete` and `OnError` here as well

akka-stream/src/main/scala/akka/stream/impl/io/SimpleUdpSinkActor.scala
```diff
@@ -0,0 +1,45 @@
...
+private[akka] object SimpleUdpSinkActor {
+  def props(target: InetSocketAddress): Props =
+    Props(new SimpleUdpSinkActor(target))
+}
+
+/** INTERNAL API */
+private[akka] class SimpleUdpSinkActor(target: InetSocketAddress) extends ActorSubscriber with ActorLogging {
+
+  private var currentRequestStrategy: RequestStrategy = ZeroRequestStrategy
+  override def requestStrategy = currentRequestStrategy
```
rkuhn: why not `override var requestStrategy: RequestStrategy = ZeroRequestStrategy` ?

akka-stream/src/main/scala/akka/stream/impl/io/SimpleUdpSinkActor.scala
```diff
@@ -0,0 +1,45 @@
...
+  IO(Udp) ! Udp.SimpleSender
+
+  override def receive: Receive = {
+    case Udp.SimpleSenderReady ⇒
+      currentRequestStrategy = WatermarkRequestStrategy(10) // TODO arbitrary value, how can we do better here (needs dropping to low level)
+      context.become(ready(sender()))
+  }
+
+  def ready(send: ActorRef): Receive = {
+    case OnNext(bytes: ByteString) ⇒ send ! Udp.Send(bytes, target)
```
rkuhn: seems a little incongruous to use the simple sender API when in “connected” mode

rkuhn:
On the priorities: I think it is more important to have File IO than UDP in RC1, and doing UDP should expose all four (in/out, connected/unconnected).

drewhk:
Ok, I agree that File IO is important.

hepin1989:
So cause the akka stream will release against to akka 2.3 release too ,so the file stream could not make use of the new Files API?

ktoso:
Ok, so let's stash this guy and ill revive my FileSink. Sounds good, this
one needs more thinking and is not critical for akka - I can use it for my
needs for the time being just lasted there.

lespea:
Are there any plans to resurrect this?

ktoso:
Not much is missing really, would you be interested in taking the PR and adding the last touchups if you're interested in this feature? Happy to guide through the process if you want.

Currently focused on implementing http/2 which takes precedence over this, for better or worse :)

PR #17164, opened 2015-04-09
+doc, str #16716: Table summarizing built-in processing stages

drewhk:
Missing:
- [x] Timer driven stages
- [x] Make scaladocs consistent with table after the table has been reviewed
- [ ] Make javadocs consistent with scaladocs after it has been reviewed
- [x] Link from relevant documentation pages

Please look through this carefully (I will do the same) since it is easy to make typos and mistakes in such a large table. Sorry for the overly long lines, it is hard to make the table relatively maintainable and at the same time keep reStructuredText happy.

PR #17196, opened 2015-04-14
+str #16885 add splitAfter

### LINKED ISSUE #16885: splitAfter combinator
For Websockets (#16887), a `splitAfter` combinator would be useful. The difference to `splitWhen`:

 * `splitWhen` **begins** a new substream when the predicate is true (each substream looks like this wrt the predicate: `true`, `false`, `false`, `false`, ...)
 * `splitAfter` **ends** the current substream when the predicate is true (each substream looks like this wrt the predicate: `false`, `true`, `true`, ...)

Is it possible to emulate the wanted behavior with the existing combinators? I thought we once discussed a generalized split combinator but I cannot find any reference any more.

#### Background for Websockets

In websockets, on the low level the incoming data is a stream of data frames that look like this:

```scala
case class DataFrame(data: ByteString, fin: Boolean)
```

where a stream of `DataFrame`s needs to be transformed into a stream of `Message`s that look like this:

```scala
case class Message(data: Source[ByteString])
```

with the combinator the implementation would look like this:

```scala
def assemble(frames: Source[Frame]): Source[Message] =
  frames
    .splitAfter(_.fin)
    .map(frames: Source[Frame] => Message(frames.map(_.data)))
```

ktoso:
Resolves #16885 

It's very similar to splitWhen, but different enough to warant it's own class I think. Sharing code would end up being more confusing than worth it I think.

@jrudolph please sanity check it has the semantics you need.


akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -487,11 +487,46 @@ class Source[+Out, +Mat](delegate: scaladsl.Source[Out, Mat]) extends Graph[Sour
...
+   *
+   * If the split predicate `p` throws an exception and the supervision decision
+   * is [[akka.stream.Supervision.Stop]] the stream and substreams will be completed
+   * with failure.
+   *
+   * If the split predicate `p` throws an exception and the supervision decision
+   * is [[akka.stream.Supervision.Resume]] or [[akka.stream.Supervision.Restart]]
+   * the element is dropped and the stream and substreams continue.
+   *
+   * See also [[Source.splitAfter]].
```
ktoso: javadsl.Source had a bit of missing scaladoc, fixed. The rest is all consistent.

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -487,11 +487,46 @@ class Source[+Out, +Mat](delegate: scaladsl.Source[Out, Mat]) extends Graph[Sour
...
+   *
+   * If the split predicate `p` throws an exception and the supervision decision
+   * is [[akka.stream.Supervision.Stop]] the stream and substreams will be completed
+   * with failure.
+   *
+   * If the split predicate `p` throws an exception and the supervision decision
+   * is [[akka.stream.Supervision.Resume]] or [[akka.stream.Supervision.Restart]]
+   * the element is dropped and the stream and substreams continue.
+   *
+   * See also [[Source.splitWhen]].
```
drewhk: Please now also add the summary section that is used in all scaladoc entries, see: https://github.com/akka/akka/pull/17164

akka-stream/src/main/scala/akka/stream/impl/SplitAfterProcessorImpl.scala
```diff
@@ -0,0 +1,94 @@
...
+import akka.stream.scaladsl.Source
+
+/**
+ * INTERNAL API
+ */
+private[akka] object SplitAfterProcessorImpl {
+  def props(settings: ActorFlowMaterializerSettings, splitPredicate: Any ⇒ Boolean): Props =
+    Props(new SplitAfterProcessorImpl(settings, splitPredicate))
+
+  private trait SplitDecision
```
drewhk: Hmm, what if we add a SplitBefore and SplitAfter, change the function from Any => Boolean to Any => SplitWhere, then we have one unified implementation for splitWhen and splitAfter (which just emit Continue or SplitBefore/SplitWhen depending on the predicate), and also different other splits.

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
ktoso: oh, sure!

akka-stream/src/main/scala/akka/stream/impl/SplitAfterProcessorImpl.scala
ktoso: Sounds neat, I'll give it a shot.

drewhk: time-box it, it just seemed straightforward.

ktoso: So it's here on [wip-one-impl-to-rule-them-all-ktoso](https://github.com/ktoso/akka/commit/9fabf3cc35471a91642473a1aac50f9e47953850).
Though I think it's gotten more complicated (2 separate modes, mixted together) and more allocations. 

Could be cleaned up a bit, but first wanted feedback if we prefer one impl or two simple ones. Please chime in guys @drewhk @rkuhn @patriknw (2 simple impls is this PR, 1 smart impl is the above commit).

akka-stream/src/main/scala/akka/stream/impl/SplitAfterProcessorImpl.scala
```diff
@@ -0,0 +1,94 @@
...
+    decideSplit(elem) match {
+      case Continue | Drop ⇒ // ignore elem
+      case Split           ⇒ nextPhase(openSubstream)
+    }
+  }
+
+  private def decideSplit(elem: Any): SplitDecision =
+    try {
+      val predicate = splitPredicate(elem)
+      if (predicate) Split else Continue
```
patriknw: the `val predicate` is only used once, so `if (splitPredicate(elem)) Split else Continue`?

ktoso: thanks :)

akka-stream/src/main/scala/akka/stream/impl/SplitAfterProcessorImpl.scala
patriknw: I'm not sure I see the added complexity. It doesn't look much compared to the two different classes.

drewhk: I agree, one less class to maintain.

ktoso: Cool, last question then - with the new class we're able to expose: `def split(Out => SplitDecision)` should we then expose `splitWhen`, `splitAfter` and `split` (split can be used fro dropping elements at will and SplitDrop as well).

2 new methods API wise, but seems pretty powerful, sounds good?

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -390,11 +392,38 @@ class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Graph
...
     new Flow(delegate.splitWhen(p.test).map(_.asJava))
 
   /**
+   * This operation applies the given predicate to all incoming elements and
+   * emits them to a stream of output streams. It *ends* the current substream when the
+   * predicate is true. This means that for the following series of predicate values,
+   * three substreams will be produced with lengths 1, 2, and 3:
+   *
+   * {{{
+   * true,               // element goes into first substream
```
jrudolph: Maybe, also add one or more leading `false` values here, to make it sure that they are included in the first stream.

akka-stream/src/main/scala/akka/stream/impl/SplitAfterProcessorImpl.scala
patriknw: if we go with two methods `splitWhen` and `splitAfter` I don't think we should also expose `split` with the enums. Otherwise only `split`, but the enums are not so nice (esp for java).

ktoso: Ok, so I'll do this: 

* Use the 1 class impl; It's theoretically more powerful than we're exposing - just splitWhen and splitAfter. This solves the immediate ticket. 
* I'll open a ticket to decide if we want to expose `split(A => Decision)` or something else like it (or nothing).

jrudolph:
You should probably also search for and look at the previous suggestion(s) that @sirthias made regarding generalizing the functionality and how the APIs could look like. I couldn't find them quickly but they are probably still there somewhere in the issues.

akka-stream/src/main/scala/akka/stream/impl/SplitAfterProcessorImpl.scala
ktoso: Pushed `SplitWhere` based impls to this PR.

akka-stream/src/main/scala/akka/stream/impl/SplitWhereProcessorImpl.scala
```diff
@@ -0,0 +1,118 @@
...
+package akka.stream.impl
+
+import akka.actor.Props
+import akka.stream.impl.SplitDecision.SplitDecision
+import akka.stream.scaladsl.Source
+import akka.stream.{ ActorFlowMaterializerSettings, Supervision }
+
+import scala.util.control.NonFatal
+
+object SplitDecision {
```
patriknw: internal

ktoso: Ah yes, thanks.

akka-stream/src/main/scala/akka/stream/impl/SplitWhereProcessorImpl.scala
```diff
@@ -0,0 +1,118 @@
...
+  sealed abstract class SplitDecision
+
+  /** Splits before the current element. The current element will be the first element in the new substream. */
+  case object SplitBefore extends SplitDecision
+
+  /** Splits after the current element. The current element will be the last element in the current substream. */
+  case object SplitAfter extends SplitDecision
+
+  /** Drops the current element, completes the current substream. The next element will be emitted into a new substream. */
+  case object SplitDrop extends SplitDecision
```
ktoso: Not implemented, I'll remove

patriknw:
LGTM

ktoso:
The failures are real, in `ResponseParserSpec` which uses a splitWhen... I thought it's "when first element matched" semantics changes, but I think those are OK (needed to add an "first element flag" otherwise I'd emit an empty substream if first element matched on splitWhen). Current semantics:

```
splitAfter: List(false, true, true, false, false, true, false)
1-false
1-true
———————————
2-true
———————————
3-false
3-false
3-true
———————————
4-false


splitWhen: List(false, true, true, false, false, true, false)
5-false
———————————
6-true
———————————
7-true
7-false
7-false
———————————
8-true
8-false


splitAfter: List(true, false, true, true, false, false, true, false)
9-true
10-false
———————————
10-true
———————————
11-true
12-false
12-false
———————————
12-true
13-false


splitWhen: List(true, false, true, true, false, false, true, false)
14-true    // technically not obvious splitWhen == SplitBefore => Source() + Source(14-true)
———————————
15-false
———————————
16-true
———————————
17-true
17-false
17-false
———————————
18-true
18-false
```

Need to look more into it (failures introduced after collapsing the two impls into one). I still feel having them separate was much simpler to deal with.

jrudolph:
One thing these kind of diagrams are not showing is when the substreams are closed. In the case of `splitWhen` it's in the moment when the next substream is opened (or the input stream is completed). In the case of `splitAfter` it's after the last element, and until the next element arrives there's no substream open.

Not sure if it's related to the failure, I haven't yet looked at the failure or the implementation in detail.

jrudolph:
The error in `FlowWithHeadAndTail` is a sign that it could be related to completion: our `headAndTail` cannot deal with empty substreams so it seems empty substreams are now introduced somewhere where they didn't occur before.

jrudolph:
Btw. as I didn't mentioned it in the original ticket: the motivation for `splitAfter` from our side is exactly the changed completion behavior: with `splitWhen` we needed to introduce an otherwise superfluous message `MessageEnd` to close the previous substream eagerly. With `splitAfter` we could clean up the event protocol to let the last actual data event close the substream automatically.

ktoso:
Thanks for the feedback @jrudolph (and sorry I did not reply here sooner). Yes I think it's empty streams, I'll look at completion behaviour so that it matches your need (add tests for it).

patriknw:
what is the status here?

ktoso:
Status here is I need to look over the failures in http (akka-stream-tests-experimental is green), possibly small change in splitWhen semantics, and when fixed should be good to go.

ktoso:
This will slip from RC1, there seems to be some semantical change other than the one I just changed between the old and new impl of splitWhen. Don't want to break HTTP with pulling this in for RC1.

ktoso:
rebased, fixed and pushed :-)
The bug was related to the first element in a `splitWhen` causing an empty substream, which was *not* the behaviour of splitWhen previously (caused by forgetting to set a boolean flag in one spot).

ktoso:
Failure was SSL spec :cry:

ktoso:
PLS BUILD

ktoso:
@drewhk @2m @jrudolph this needs a last review and should be merged I think.

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -509,11 +509,72 @@ class Source[+Out, +Mat](delegate: scaladsl.Source[Out, Mat]) extends Graph[Sour
...
+   *
+   * If the split predicate `p` throws an exception and the supervision decision
+   * is [[akka.stream.Supervision.Stop]] the stream and substreams will be completed
+   * with failure.
+   *
+   * If the split predicate `p` throws an exception and the supervision decision
+   * is [[akka.stream.Supervision.Resume]] or [[akka.stream.Supervision.Restart]]
+   * the element is dropped and the stream and substreams continue.
+   *
+   * '''Emits when''' an element for which the provided predicate is true, opening and emitting a new substream for subsequent element
```
drewhk: add to the table in the docs

ktoso: done, thanks for reminding me!

ktoso:
Final LGTM?

drewhk:
merge

ktoso:
Merged `splitAfter` // cc @jrudolph might be interesting for you for websockets :)

jrudolph:
Thanks, @ktoso. Will try when when implementing the client-side of websocket support.

ktoso:
Sounds great :)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -803,8 +812,47 @@ trait FlowOps[+Out, +Mat] {
    * '''Cancels when''' downstream cancels and substreams cancel
    *
    */
-  def splitWhen[U >: Out](p: Out ⇒ Boolean): Repr[Source[U, Unit], Mat] =
-    andThen(SplitWhen(p.asInstanceOf[Any ⇒ Boolean]))
+  def splitWhen[U >: Out](p: Out ⇒ Boolean): Repr[Out, Mat]#Repr[Source[U, Unit], Mat] = {
+    val f = p.asInstanceOf[Any ⇒ Boolean]
+    withAttributes(name("splitWhen")).andThen(Split(el ⇒ if (f(el)) SplitBefore else Continue))
+  }
```
hepin1989: Why not
```
  def splitWhen[U >: Out](p: Out ⇒ Boolean): Repr[Source[U, Unit], Mat] = {
    val f = p.asInstanceOf[Any ⇒ Boolean]
    andThen(Split(el ⇒ if (f(el)) SplitBefore else Continue, name("splitWhen")))
  }
```
?

ktoso: nice one, would you like to submit that change as PR? :)

ktoso: And on https://github.com/akka/akka/pull/17196/files#diff-d25227230bef7fd44d678521a8f07f1cR854 as well

hepin1989: then how about adding this:
```
  object Split{
    def when(f:Any ⇒ Boolean)= {
      Split(el ⇒ if (f(el)) SplitAfter else Continue, name("splitAfter"))
    }
    
    def after(f:Any ⇒ Boolean)= {
      Split(el ⇒ if (f(el)) SplitAfter else Continue, name("splitAfter"))
    }
  }
```
and then could have:
```
  def groupBy[K, U >: Out](f: Out ⇒ K): Repr[(K, Source[U, Unit]), Mat] =
    andThen(GroupBy(f.asInstanceOf[Any ⇒ Any]))

  def splitWhen[U >: Out](p: Out ⇒ Boolean): Repr[Source[U, Unit], Mat] = {
    andThen(Split.when(p.asInstanceOf[Any ⇒ Boolean]))
  }
```
?

ktoso: That's a nice refactoring, let's do it :) please submit a PR

Issue #15898, opened 2014-09-16
Java API for ActorPath.fromString is missing


patriknw:
because of trait + object
or do I look at the wrong place?

hepin1989:
change the sealed trait to ```sealed abstract class``` or add a new object like ```ActorPathFactory``` ?

hepin1989:
btw,abstract class is a little faster than interface

PR #17201, opened 2015-04-15
#15898 change ActorPath to abstract class to make the methods in the com...

He-Pin:
1,make the ValidSymbols private
2,change the ActorPath from sealed trait to sealed abstract class
3,and an overload method for validatePathElement

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/ActorPath.scala
+++ b/akka-actor/src/main/scala/akka/actor/ActorPath.scala
@@ -17,8 +17,7 @@ object ActorPath {
     case _                               ⇒ throw new MalformedURLException("cannot parse as ActorPath: " + s)
   }
 
-  /** INTERNAL API */
-  private[akka] final val ValidSymbols = """-_.*$+:@&=,!~';"""
+  private final val ValidSymbols = """-_.*$+:@&=,!~';"""
 
   private final val ValidPathCode = -1
   private final val EmptyPathCode = -2
@@ -28,9 +27,17 @@ object ActorPath {
    * See [[isValidPathElement()]] for a non-throwing version.
    *
    * @param element actor path element to be validated
+   */
+  final def validatePathElement(element: String): Unit = validatePathElement(element, fullPath = null)
+
+  /**
+   * Validates the given actor path element and throws an [[InvalidActorNameException]] if invalid.
+   * See [[isValidPathElement()]] for a non-throwing version.
+   *
+   * @param element actor path element to be validated
    * @param fullPath optional fullPath element that may be included for better error messages; null if not given
    */
-  final def validatePathElement(element: String, fullPath: String = null): Unit = {
+  final def validatePathElement(element: String, fullPath: String): Unit = {
     def fullPathMsg = if (fullPath ne null) s""" (in path [$fullPath])""" else ""
 
     (findInvalidPathElementCharPosition(element): @switch) match {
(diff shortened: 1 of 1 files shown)


akka-actor/src/main/scala/akka/actor/ActorPath.scala
```diff
@@ -28,9 +28,17 @@ object ActorPath {
...
+
+  /**
+   * Validates the given actor path element and throws an [[InvalidActorNameException]] if invalid.
+   * See [[isValidPathElement()]] for a non-throwing version.
+   *
+   * @param element actor path element to be validated
    * @param fullPath optional fullPath element that may be included for better error messages; null if not given
    */
-  final def validatePathElement(element: String, fullPath: String = null): Unit = {
+  final def validatePathElement(element: String, fullPath: String): Unit = {
```
rkuhn: This change is not binary compatible (unfortunate, I know).

akka-actor/src/main/scala/akka/actor/ActorPath.scala
```diff
@@ -98,7 +106,7 @@ object ActorPath {
  * when comparing actor paths.
  */
 @SerialVersionUID(1L)
-sealed trait ActorPath extends Comparable[ActorPath] with Serializable {
+sealed abstract class ActorPath extends Comparable[ActorPath] with Serializable {
```
rkuhn: we’ll see what MiMa says to this one, might work

rkuhn:
OK TO TEST

rkuhn:
refs #15898 

Thanks for the PR! Concerning the BC compatibility comment: it should be fine to add that overload for Java and make it `protected` (unless that removes the static forwarder, please try it out).

akka-actor/src/main/scala/akka/actor/ActorPath.scala
```diff
@@ -18,7 +18,7 @@ object ActorPath {
   }
 
   /** INTERNAL API */
-  private[akka] final val ValidSymbols = """-_.*$+:@&=,!~';"""
+  private final val ValidSymbols = """-_.*$+:@&=,!~';"""
```
ktoso: if `private` the `/** INTERNAL API */` marker can be removed

akka-actor/src/main/scala/akka/actor/ActorPath.scala
ktoso: IMO good news - this is a new method on master (I have added it in https://github.com/akka/akka/commit/c31469dd2930669324512a7750ecc1cb8e414d8f , on 2.3 we only have `isValidPathElement`) and it's on an object - not a trait, so we're safe from the "method added to trait" problem.

I think we might get away with this change in this case.

akka-actor/src/main/scala/akka/actor/ActorPath.scala
ktoso: We don't have mima enabled on `master`. 
I think we're currently not passing MiMa between 2.3 and 2.4 - @patriknw did experiment towards this but we didn't resolve all incompatibilities AFAIR

rkuhn: Yes, I know; I’m just keeping a watchful eye on these as we will have to correct all the mistakes in a few weeks time anyway.

akka-actor/src/main/scala/akka/actor/ActorPath.scala
rkuhn: ah, then all is good, thanks (and sorry for not checking)

rkuhn:
LGTM, thanks!

PR #17202, opened 2015-04-15
=str #17200 Stop shard region when MemberRemoved

patriknw:


patriknw:
Refs #17200

patriknw:
Updated with new solution. Two issues:

1) ShardRegion actor must stop itself when the node is shutting down,
   ie. when receiving MemberRemoved(selfAddress)

2) ShardCoordinator must not persist anything when the node is shutting
   down. MemberRemoved of other shard regions will trigger Terminated,
   which must not be persisted, because then the next coordinator will
   replay those events and end up in wrong state. This is a problem
   announced itself when using leaving as illustrated in the new test.

To solve the second issue I have added a new ClusterShuttingDown event
that is published before the MemberRemoved events. Note that Terminated
is triggered by MemberRemoved.

akka-cluster/src/main/scala/akka/cluster/ClusterEvent.scala
```diff
@@ -170,6 +170,17 @@ object ClusterEvent {
   }
 
   /**
+   * This event is published when the cluster node is shutting down,
+   * before the final [[MemberRemoved]] events are published.
+   */
+  case object ClusterShuttingDown extends ClusterDomainEvent
```
rkuhn: we regard these as local-only messages, right?

akka-contrib/src/main/scala/akka/contrib/pattern/ClusterSharding.scala
```diff
@@ -1331,6 +1337,18 @@ class ShardCoordinator(handOffTimeout: FiniteDuration, rebalanceInterval: Finite
       persistentState.regionProxies.foreach(context.watch)
       persistentState.regions.foreach { case (a, _) ⇒ context.watch(a) }
 
+    case ClusterShuttingDown ⇒
+      log.debug("Shutting down ShardCoordinator")
+      // can't stop because supervisor will start it again,
+      // it will soon be stopped when singleton is stopped
+      context.become(shuttingDown)
```
rkuhn: perhaps we should introduce `Actor.ignoringBehavior`?

akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterShardingLeavingSpec.scala
```diff
@@ -0,0 +1,208 @@
...
+      case GetLocations ⇒ sender() ! locations
+      case l: Locations ⇒ locations = l
+    }
+  }
+
+  val idExtractor: ShardRegion.IdExtractor = {
+    case m @ Ping(id) ⇒ (id, m)
+  }
+
+  val shardResolver: ShardRegion.ShardResolver = msg ⇒ msg match {
```
rkuhn: wouldn’t it work without the `msg => msg match ` part?

akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterShardingLeavingSpec.scala
```diff
@@ -0,0 +1,208 @@
...
+          region ! Ping(id)
+          id -> expectMsgType[ActorRef]
+        }).toMap
+        shardLocations ! Locations(locations)
+      }
+      enterBarrier("after-3")
+    }
+
+    "recover after leaving coordinator node" in within(30.seconds) {
+      runOn(second, third, fourth) {
```
rkuhn: this looks extraneous

rkuhn:
LGTM

akka-cluster/src/main/scala/akka/cluster/ClusterEvent.scala
patriknw: yes, that is the intended usage. we don't have custom serialization or serialversionuid for any of these ClusterDomainEvent

akka-contrib/src/main/scala/akka/contrib/pattern/ClusterSharding.scala
patriknw: nice to have category, added ticket #17257

akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterShardingLeavingSpec.scala
patriknw: indeed, copy pasta, replaced 3 occurrences

akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterShardingLeavingSpec.scala
patriknw: yeah :)

patriknw:
final touch incorporates @rkuhn's comments

drewhk:
LGTM

rkuhn:
fixes #17200

2m:
LGTM

akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterShardingLeavingSpec.scala
```diff
@@ -0,0 +1,206 @@
...
+
+    "setup shared journal" in {
+      // start the Persistence extension
+      Persistence(system)
+      runOn(first) {
+        system.actorOf(Props[SharedLeveldbStore], "store")
+      }
+      enterBarrier("peristence-started")
+
+      runOn(first, second) {
```
patriknw: I had big trouble porting this PR to master. This test failed, and it turned out that this is wrong. It should initialize the SharedLeveldbJournal.setStore on all nodes. The reason it fails in master is the new rememberEntries feature, which requires persistence on all nodes, even if the feature is not used. I'll create a ticket about that.

PR #16324, opened 2014-11-18
Bposerow get subscribers of mediator 16141

BPoserow:
Pushing change to support GetTopics message that returns subscribers by topic

Issue #16141, opened 2014-10-24
should add method to get DistributedPubSubMediator's current topics


hepin1989:
I think it would be handy if we could send a message to the Mediator to get the current all registered topics's snapshot,this could help something like debug or any other use case.
maybe like
```
case object GetTopics
``` 
and replay with 
```
case class CurrentTopics(topics:Map[String, Set[ActorRef]])
```

PR #17204, opened 2015-04-15
+con #16141 Add GetTopics message to DistributedPubSubMediator providing list of topics

patriknw:
Squashed #16324


patriknw:
Refs #16141

Issue #16346, opened 2014-11-20
Akka allocates new socket for each unsuccessful connection attempt to a quarantined system


svozniuk:
I'm using Akka 2.3.4 running on Java 1.7.0_72 in Ubuntu 14.04 (also observed on CentOS and Mac)

I'm reproducibly falling into the situation when a remote system is quarantined and for each unsucessful attempt to connect to that remote system a new socket is allocated but never disposed. Eventually this leads to reaching the operating system limits on the number of open sockets (and files --> Too many open files in the system).

You can find a minimal project and a description how to reproduce the situation in this thread
https://groups.google.com/forum/#!topic/akka-user/IKWCjXMpV4A

patriknw:
@ktoso Did you reproduce this? Do you have an environment where you can run the reproducer? We might have fixed this in later versions?

ktoso:
I did not reproduce the issue (did not attempt to yet), just made sure to link to the reproducer @svozniuk had prepared, see here: https://github.com/akka/akka/issues/16346#issuecomment-63799726

svozniuk:
@ktoso @patriknw I have just followed the instructions I provided using the reproducer project. The problem is still there. I also tried with version 2.3.9 - same behavior as before. E.g. these are the sockets opened during 5 minutes after reaching quarantined state. And it's growing

```
tcp6       0      0 127.0.0.1:2552          :::*                    LISTEN      9852/java       
tcp6       0      0 127.0.0.1:53182         127.0.0.1:2552          ESTABLISHED 9852/java       
tcp6       0      0 127.0.0.1:53113         127.0.0.1:2552          ESTABLISHED 9852/java       
tcp6       0      0 127.0.0.1:53148         127.0.0.1:2552          ESTABLISHED 9852/java       
tcp6       0      0 127.0.0.1:53195         127.0.0.1:2552          ESTABLISHED 9852/java       
tcp6       0      0 127.0.0.1:2552          127.0.0.1:53182         ESTABLISHED 9852/java       
tcp6       0      0 127.0.0.1:53201         127.0.0.1:2552          ESTABLISHED 9852/java       
tcp6       0      0 127.0.0.1:2552          127.0.0.1:53195         ESTABLISHED 9852/java       
tcp6       0      0 127.0.0.1:2552          127.0.0.1:53111         ESTABLISHED 9852/java       
tcp6       0      0 127.0.0.1:2552          127.0.0.1:53171         ESTABLISHED 9852/java       
[...]

PR #17205, opened 2015-04-15
=rem #16346 Close connection after refused connection attempt

patriknw:
This fix the issue illustrated by this test scenario: https://github.com/akka/akka/commit/e35350f4cbb2239b43b15a8df5883fdc53746d78

I'm not sure if there are more places, e.g. other cases in that onTermination action.

### DIFF
--- a/akka-remote/src/main/scala/akka/remote/transport/AkkaProtocolTransport.scala
+++ b/akka-remote/src/main/scala/akka/remote/transport/AkkaProtocolTransport.scala
@@ -505,6 +505,7 @@ private[transport] class ProtocolStateActor(initialData: InitialProtocolStateDat
         case _                                   ⇒ Disassociated(Unknown)
       }
       handlerFuture foreach { _ notify disassociateNotification }
+      wrappedHandle.disassociate()
 
     case StopEvent(reason, _, ListenerReady(handler, wrappedHandle)) ⇒
       val disassociateNotification = reason match {


rkuhn:
So I guess we cannot really make the test fail when things don’t work?

akka-remote/src/main/scala/akka/remote/transport/AkkaProtocolTransport.scala
```diff
@@ -505,6 +505,7 @@ private[transport] class ProtocolStateActor(initialData: InitialProtocolStateDat
         case _                                   ⇒ Disassociated(Unknown)
       }
       handlerFuture foreach { _ notify disassociateNotification }
+      wrappedHandle.disassociate()
```
drewhk: oh my. Thanks!

ktoso:
LGTM, other cases seem to disassociate properly, but @drewhk should perhaps sanity check we're not missing it in a related place here?

drewhk:
I don't think so. This can be tested however with the TestTransport.

rkuhn:
LGTM, merge?

patriknw:
@drewhk ok to merge? I know you would like to see a test, but I'm not sure how to do that and it is probably not worth it for this one.

drewhk:
Ok to merge,

patriknw:
Refs #16346

Issue #17206, opened 2015-04-15
properly document the semantics of `context.stop(self)`


rkuhn:
the ScalaDoc is currently a bit vague and ambiguous

PR #17207, opened 2015-04-15
=act #17206 properly document semantics of context.stop(self)

rkuhn:


rkuhn:
fixes #17206

ktoso:
LGTM, nice

akka-actor/src/main/scala/akka/actor/ActorRefProvider.scala
```diff
@@ -338,6 +338,9 @@ trait ActorRefFactory {
   /**
    * Stop the actor pointed to by the given [[akka.actor.ActorRef]]; this is
    * an asynchronous operation, i.e. involves a message send.
+   * If this method is applied to the `self` reference from inside an Actor
+   * then that Actor is guaranteed to not process any further messages before
```
drewhk: before terminating?

rkuhn: yes, WDYM?

drewhk: well, will it process further messages after termination?

rkuhn: ah, I see :-) I’d still leave this in to make it absolutely clear that message dequeuing does not only stop after termination but _right now_.

drewhk: then why don't you just say that it is guarenteed to not process any further message *after this call*

rkuhn: yup, better, done that

Issue #16608, opened 2014-12-29
akka.stream :  null element in output list of a Flow.mapConcat fails assertion 


pbuszka:
If user function in mapConcat returns a list which contains a null as an element akka-stream fails with assertionFailure in akka.stream.impl.fusing.BatchingActorInputBoundary.dequeue line 41

I think if nulls are not allowed as list elements and eventually crash the system then it should be documented in the description of the function or generate a catchable runtime exception.

scala version 2.11.4
akka-stream-experimental_2.11-1.0-M2.jar

code which reproduces the error below, if null is removed from the list everything goes normally
```
package testakka

import akka.actor.ActorSystem
import akka.stream.FlowMaterializer
import akka.stream.scaladsl.Source
import akka.stream.scaladsl.BlackholeSink
import scala.concurrent.Await
import scala.concurrent.duration._
import akka.stream.scaladsl.Sink

object TestMapConcat extends App {
  implicit val system = ActorSystem()
  implicit val materializer = FlowMaterializer()
  
  case class Element(val test: Int)
  
  val source = Source[Int]( () => List(1,2).iterator)
  val sink = Sink.foreach[Element] { x => println(x) }

  val materializedMap = source.mapConcat { x: Int => List[Element](Element(x), null, Element(x+1)) }.to(sink).run()

  val sinkPromise = materializedMap.get(sink)
  Await.result(sinkPromise, 100 seconds)
}
```

rkuhn:
see reactive-streams/reactive-streams#204

patriknw:
this must have been on wrong milestone

rkuhn:
hmm, it seems that we have simply overlooked it when moving things from streams-1.x, thanks for seeing it!

patriknw:
yeah, I noticed it in the `bug` list

PR #17210, opened 2015-04-15
Document mapConcat's Seq should not contain nulls

ktoso:
Resolves https://github.com/akka/akka/issues/16608

Not sure if we want the section in the docs about it or not (1st commit)


patriknw:
LGTM
the stream is failed with NPE if the list contains null, I guess

ktoso:
It fails with `final val ElementMustNotBeNullMsg = "Element must not be null, rule 2.13"`

rkuhn:
LGTM

Issue #15588, opened 2014-07-24
streams: Provide basic streams for files and from InputStream etc.


jrudolph:

rkuhn:
For 1.0 we can just do the JDK6 version, we’ll need to add proper NIO2 support for Akka 2.4.0.

ktoso:
Got those implemented (for 6 and 7+), will clean-up and pull request later this week

rkuhn:
Awesome, thanks!

Issue #16610, opened 2015-01-01
DOC: Add recipe for batcher element


drewhk:
Grouped is not good for batching on "interactive" streams (where long time gaps might happen between elements). Provide a recipe for a custom DetachedOp that works like a conflate that groups element, but with the difference that it caps the group size (i.e. this is a hybrid between buffer and conflate).

rkuhn:
This has been implemented as a combinator instead.

PR #17211, opened 2015-04-16
+str #15588 Java 6 Synchronous File Sink / Source

ktoso:
Note for reviewers: Yes, these are blocking and pretty simple. Got asynchronous impls stashed and can dig them up, however we can't include them directly because they'd require JDK7.

Resolves https://github.com/akka/akka/issues/15588


akka-stream-tests/src/test/scala/akka/stream/scaladsl/SynchronousFileSinkSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+import akka.stream.testkit.AkkaSpec
+import akka.stream.{ ActorFlowMaterializer, ActorFlowMaterializerSettings }
+import akka.util.ByteString
+
+import scala.collection.mutable.ListBuffer
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+class SynchronousFileSinkSpec extends AkkaSpec(
+  """ akka.actor.default-mailbox.mailbox-type = "akka.dispatch.UnboundedMailbox" """) {
```
ktoso: This or I could set the file-io dispatcher to the StreamTest one.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SynchronousFileSourceSpec.scala
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    List(
+      Settings(chunkSize = 8, readAhead = 8),
+      Settings(chunkSize = 128, readAhead = 1),
+      Settings(chunkSize = 256, readAhead = 1),
+      Settings(chunkSize = 256, readAhead = 2),
+      Settings(chunkSize = 512, readAhead = 1),
+      Settings(chunkSize = 512, readAhead = 2),
+      Settings(chunkSize = 4096, readAhead = 1)) foreach { settings ⇒
```
ktoso: this test case was more interesting for the async impl, I'll lessen the test cases we do for the synchronous one here.

akka-stream-testkit/src/test/resources/reference.conf
```diff
@@ -2,7 +2,7 @@
 #
 # All stream tests should use the dedicated `akka.test.stream-dispatcher` or disable this validation by defining:
 # akka.actor.default-mailbox.mailbox-type = "akka.dispatch.UnboundedMailbox"
-akka.actor.default-mailbox.mailbox-type = "akka.stream.testkit.StreamTestDefaultMailbox"
+# akka.actor.default-mailbox.mailbox-type = "akka.stream.testkit.StreamTestDefaultMailbox"
```
patriknw: ?

akka-stream/src/main/resources/reference.conf
```diff
@@ -42,5 +42,10 @@ akka {
       output-burst-limit = 1000
     }
 
+    # Fully qualified config path which holds the dispatcher configuration
+    # to be used by FlowMaterialiser when creating Actors for IO operations,
+    # such as FileSource, FileSink and others.
+    file-io-dispatcher = ${akka.io.tcp.file-io-dispatcher} # TODO better provide a thread pool backed one here by default?
+                                                           # note: tcps dispatcher is the is the akka.actor.default-dispatcher
```
ktoso: TODO: I'm working on the pre-configured dispatcher for this guy here.
I ran into trouble when wanting to provide a default special dispatcher for these, and still allow users to use `.withAttributes(dispatcher...)`, I'll update the PR soon today.

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala.orig
```diff
@@ -0,0 +1,320 @@
+/**
```
ktoso: ignore this please

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SynchronousFileSinkSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+      val written2 = Await.result(completion2, 3.seconds)
+
+      f.length() should ===(written1 + written2)
+      checkFileContents(f, TestLines.mkString("") + lastWrite.mkString("") + "\n")
+    }
+
+  }
+
+  private def targetFile(block: File ⇒ Unit) {
+    val targetFile = new File("synchronous-file-sink.tmp")
```
patriknw: writing tmp file to current directory, perhaps "target/synchronous-file-sink.tmp"

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala.orig
```diff
@@ -0,0 +1,320 @@
...
+ */
+package akka.stream
+
+import java.util.Locale
+import java.util.concurrent.TimeUnit
+
+import akka.actor.{ ActorContext, ActorRef, ActorRefFactory, ActorSystem, ExtendedActorSystem, Props }
+import akka.stream.impl._
+import akka.stream.javadsl.japi
+<<<<<<< HEAD
```
patriknw: ?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SynchronousFileSinkSpec.scala
ktoso: good idea, yes

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
```diff
@@ -0,0 +1,117 @@
...
+  private val fs = new RandomAccessFile(f, "r") // best way to express this in JDK6, OpenOption are available since JDK7
+  private var chan: FileChannel = _
+
+  override def preStart() = {
+    try {
+      chan = fs.getChannel
+    } catch {
+      case ex: Exception ⇒
+        log.error(ex, "Unable to open channel for {}", f.getAbsolutePath)
+        context.stop(self)
```
patriknw: I think this is too early. Should emit error downstreams when `Request` is received?

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
```diff
@@ -0,0 +1,117 @@
...
+        eofReachedAtOffset = chan.position
+        log.debug("No more bytes available to read (got `-1` from `read`), marking final bytes of file @ " + eofReachedAtOffset)
+
+      case _ ⇒
+        buf.flip()
+        readBytesTotal += readBytes
+        availableChunks :+= ByteString(buf).take(readBytes)
+        buffs.release(buf)
+
+      // valid read, continue
```
patriknw: what if any of this throws?

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  private def signalOnNexts(): Unit =
+    if (availableChunks.nonEmpty && totalDemand > 0) {
+      val ready = availableChunks.head
+      availableChunks = availableChunks.tail
+
+      onNext(ready)
+
+      if (totalDemand > 0) signalOnNexts()
+    } else if (eofEncountered) onComplete()
```
patriknw: + context.stop(self)

akka-stream/src/main/scala/akka/stream/impl/SynchronousFileSubscriber.scala
```diff
@@ -0,0 +1,76 @@
...
+
+}
+
+/** INTERNAL API */
+private[akka] class SynchronousFileSubscriber(f: File, completionPromise: Promise[Long], append: Boolean) extends akka.stream.actor.ActorSubscriber
+  with ActorLogging {
+
+  // TODO since ActorPublisher is typed, shouldn't ActorSubscriber be as well?
+
+  override protected val requestStrategy = OneByOneRequestStrategy // TODO better one here ?
```
patriknw: WatermarkRequestStrategy with highWatermark == maxBufferSize from the OperationAttributes?

akka-stream/src/main/scala/akka/stream/impl/SynchronousFileSubscriber.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  override def preStart(): Unit = {
+    try {
+      chan = fs.getChannel
+
+      // manually supporting appending to files - in Java 7 we could use OpenModes: FileChannel.open(f, openOptions.asJava)
+      if (append) chan.position(chan.size())
+    } catch {
+      case ex: Exception ⇒
+        completionPromise.failure(ex)
```
patriknw: Too early, I think you can call `cancel()` here and when `OnSubscribe` is handled in aroundReceive it will be cancelled.
However, we must also stop the actor. Perhaps we should be more opinionated about the actor lifecycle of the ActorPublisher and ActorSubscriber, i.e. stop it when the stream is completed/cancelled/failed?

akka-stream/src/main/scala/akka/stream/impl/SynchronousFileSubscriber.scala
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  def receive = {
+    case ActorSubscriberMessage.OnNext(bytes: ByteString) ⇒
+      try {
+        bytesWritten += chan.write(bytes.asByteBuffer)
+      } catch {
+        case ex: Exception ⇒
+          completionPromise.failure(ex)
+          context.stop(self)
```
patriknw: `cancel()`

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -225,4 +224,17 @@ object Sink extends SinkApply {
...
+  /**
+   * Creates a synchronous (Java 6 compatible) Sink that writes incoming [[ByteString]] elements to the given file.
+   *
+   *
+   * Materializes a [[Future]] that will be completed with the size of the file (in bytes) at the streams completion.
+   *
+   * This source is backed by an Actor which will use the dedicated `akka.stream.file-io-dispatcher`,
+   * unless configured otherwise by using [[ActorOperationAttributes]].
+   */
+  def synchronousFile(f: File, append: Boolean = false): Sink[ByteString, Future[Long]] =
```
patriknw: should this be named `syncronousFile` or just `file` and we pick the best implementation that is available on the running platform?

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala.orig
```diff
@@ -0,0 +1,246 @@
...
+import akka.stream.impl._
+<<<<<<< HEAD
+import akka.stream.{ SinkShape, Inlet, Outlet, Graph, OperationAttributes }
+import akka.stream.OperationAttributes._
+import akka.stream.stage.{ TerminationDirective, Directive, Context, PushStage }
+=======
+import akka.stream.stage.{ Context, PushStage, SyncDirective, TerminationDirective }
+import akka.stream.{ FlowMaterializer, Graph, Inlet, SinkShape, Supervision, javadsl }
+import akka.util.ByteString
+>>>>>>> +str #15588 Java 6 Synchronous File Sink / Source
```
patriknw: ?

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
ktoso: Yes, good catch. I'll also add a test trying to read from not existing file.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SynchronousFileSinkSpec.scala
patriknw: or use File.createTempFile

akka-stream-testkit/src/test/resources/reference.conf
ktoso: Oh, this shouldn't be pushed here - sorry.
// Since it wants the dispatcher to be the test dispatcher it disrupted my setting of io-dispatcher. I'll change that to be only overriden only in my test.

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
ktoso: // note: solving this by onError() triggering stopping of the actor once it has signalled the error

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
ktoso: Will be solved by `ActorPublisher.onComplete()` now handling stopping of actors.

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
ktoso: thanks, yes here it's best to onError, not rely on supervision I think?

akka-stream/src/main/scala/akka/stream/impl/SynchronousFileSubscriber.scala
ktoso: good idea, doing

akka-stream/src/main/scala/akka/stream/impl/SynchronousFileSubscriber.scala
ktoso: Right, also updating the Subscriber to handle stopping.

akka-stream/src/main/scala/akka/stream/impl/SynchronousFileSubscriber.scala
ktoso: thanks

akka-stream/src/main/scala/akka/stream/impl/SynchronousFileSubscriber.scala
```diff
@@ -0,0 +1,76 @@
...
+
+/** INTERNAL API */
+private[akka] class SynchronousFileSubscriber(f: File, completionPromise: Promise[Long], append: Boolean) extends akka.stream.actor.ActorSubscriber
+  with ActorLogging {
+
+  // TODO since ActorPublisher is typed, shouldn't ActorSubscriber be as well?
+
+  override protected val requestStrategy = OneByOneRequestStrategy // TODO better one here ?
+
+  private val fs = new RandomAccessFile(f, "rw") // best way to express this in JDK6, OpenOption are available since JDK7
```
ktoso: (this guy can throw, moved it into preStart)

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
ktoso: My reasoning here was that JDK7+ API will look differently because then we can use http://docs.oracle.com/javase/7/docs/api/java/nio/file/StandardOpenOption.html instead of boolean flag for append (and other interesting modes). Once we require Java 8 we could deprecate this method separately nicely if it's separate

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala.orig
ktoso: sorry about the `*.orig`s :bow:

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
patriknw: stopping supervision is the only that makes sense and that would require another actor, so I would say catch and stop

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
patriknw: ok, makes sense

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
ktoso: :+1: done

ktoso:
Includes cleaned up commit to resolve https://github.com/akka/akka/issues/17229

akka-http/src/main/resources/reference.conf
```diff
@@ -39,5 +39,5 @@ akka.http.routing {
 
   # Fully qualified config path which holds the dispatcher configuration
   # to be used by FlowMaterialiser when creating Actors for IO operations.
-  file-io-dispatcher = ${akka.io.tcp.file-io-dispatcher}
+  file-io-dispatcher = ${akka.io.tcp.file-io-dispatcher} // TODO point to akka.stream.file-io-dispatcher instead?
```
ktoso: @sirthias @jrudolph @patriknw any feelings on this suggestion? I'm asking because `akka.io.tcp.file-io-dispatcher` is by default the default dispatcher of the actor system https://github.com/akka/akka/blob/release-2.3-dev/akka-actor/src/main/resources/reference.conf#L574 and users are expected to configure one by themselfes for IO. For the file-source / file-sink we now provide a preconfigured simple threadpool dispatcher: `akka.stream.file-io-dispatcher`

Should we keep this setting as is here?

sirthias: +1 from me on your `// TODO` proposal of having a default file-io-dispatcher which is *not* the ActorSystem's main one.

patriknw: +1 and shouldn't we create an issue for changing the akka.io.tcp.file-io-dispatcher to point to a dedicated io dispatcher?

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
```diff
@@ -0,0 +1,124 @@
...
+
+      if (totalDemand > 0) signalOnNexts()
+    } else if (eofEncountered) {
+      onComplete()
+    }
+
+  /** BLOCKING I/O READ */
+  def loadChunk() = try {
+    val buf = buffs.acquire()
+    buf.flip().clear()
```
rkuhn: the `flip` does nothing here

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
```diff
@@ -0,0 +1,124 @@
...
+    readBytes match {
+      case -1 | 0 ⇒
+        // had nothing to read into this chunk, will complete now
+        eofReachedAtOffset = chan.position
+        log.debug("No more bytes available to read (got `-1` from `read`), marking final bytes of file @ " + eofReachedAtOffset)
+
+      case _ ⇒
+        buf.flip()
+        readBytesTotal += readBytes
+        availableChunks :+= ByteString(buf).take(readBytes)
```
rkuhn: the `take` looks extraneous

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
```diff
@@ -0,0 +1,124 @@
...
+      signalOnNexts()
+
+      // read chunks until readAhead is fulfilled
+      while (availableChunks.length < readAhead && !eofEncountered && isActive)
+        loadChunk()
+
+      self ! Continue
+    }
+
+  private def signalOnNexts(): Unit =
```
rkuhn: please annotate with `@tailrec` to codify this expectation

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
```diff
@@ -0,0 +1,124 @@
...
+    if (availableChunks.nonEmpty && totalDemand > 0) {
+      val ready = availableChunks.head
+      availableChunks = availableChunks.tail
+
+      onNext(ready)
+
+      if (totalDemand > 0) signalOnNexts()
+    } else if (eofEncountered) {
```
rkuhn: so if the demand drop to zero while still chunks are outstanding but the last one has already been read, then we’ll emit onComplete a little early ;-)

rkuhn:
Careful when merging with Patrik’s termination checker PR ;-)

Two things:

  * also add docs and `javadsl.Sink.synchronousFile`
  * maybe add offset/length for the sources?

akka-http/src/main/resources/reference.conf
ktoso: okey, linked to `akka.stream.default-file-io-dispatcher` then, it's default conf is:

```
    default-file-io-dispatcher {
      type = "Dispatcher"
      executor = "thread-pool-executor"
      throughput = 1

      thread-pool-executor {
        core-pool-size-min = 2
        core-pool-size-factor = 2.0
        core-pool-size-max = 16
      }
    }
```

The `throughput = 1` we thought makes sense for a "file" worker threadpool. Let me know if this can be improved as a nice default.

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
ktoso: Thanks, also:

/me just realised `buffs.acquire()` already calls `clear()`. This must have been a remainder from one of my other implementations.

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
ktoso: reworked this part of the code to be simpler and valid, thanks for keeping an eye out.
Added a test as well.

ktoso:
Yes, I stole the checker ;) Will be careful when rebasing (Patrik's first)

ktoso:
Not sure about `maybe add offset/length for the sources?` "read starting from this offset?" I guess, not sure if a core feature

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
```diff
@@ -0,0 +1,124 @@
...
+}
+
+/** INTERNAL API */
+private[akka] class SynchronousFilePublisher(f: File, bytesReadPromise: Promise[Long], chunkSize: Int, initialBuffer: Int, maxBuffer: Int)
+  extends akka.stream.actor.ActorPublisher[ByteString]
+  with ActorLogging {
+
+  import SynchronousFilePublisher._
+
+  val buffs = new DirectByteBufferPool(chunkSize, maxBuffer)
```
jrudolph: A maybe heretic question: what's the reason for using DirectByteBuffers and FileChannels instead of a simple InputStream based solution (which would then work for any InputStream)?

I would expect this solution to be even slower than the InputStream solution because what would be one native call (`InputStream.read`) now becomes at least two: allocate a direct buffer, FileChannel.read and ByteString(buf). You also don't prevent any buffer copies because `ByteString(buf)` will have to copy the bytes from the native heap into the heap in the same way that the implementation of `InputStream.read` does internally.

Direct buffers usually make sense for asynchronous IO operations but as asynchronous file IO isn't supported on the OS-level (at least under Linux, but also under other OS the interaction with the page cache will always be tricky) using `AsynchronousFileChannel` may not be best solution compared to a simpler one that blocks on a thread controlled by the user instead of blocking in the internal thread pool.

akka-http/src/main/resources/reference.conf
patriknw: I'm not sure if trying to adjust the size based on number of cores are relevant for such a pool or if it is more appropriate to define a fixed size?

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
rkuhn: You are right, I was thinking along the same lines but thought that my Java is so rusty that I don’t know the modern ways.

ktoso: Not heretic at all, thanks pointing it out!
I recall having benchmarked multiple implementations in the past in a train at some point. Can't remember results now, will dig them up and update to IS.

akka-http/src/main/resources/reference.conf
drewhk: Good question. I would say fixed size is likely better since the contended resource is not the CPU, therefore core count is less significant (I wouldn't go as far as to say insignificant though)

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -185,6 +187,32 @@ object Source {
...
+   * Creates a synchronous (Java 6 compatible) Source from a Files contents.
+   * Emitted elements are [[ByteString]] of default 2048 bytes size.
+   *
+   * This source is backed by an Actor which will use the dedicated thread-pool base dispatcher.
+   * You can configure the default dispatcher for this Source by changing the `akka.stream.file-io-dispatcher` or
+   * set it for a given Source by using [[ActorOperationAttributes]].
+   *
+   * It materializes a [[Future]] containing the number of bytes read from the source file upon completion.
+   */
+  def synchronousFile(f: File): Source[ByteString, Future[Long]] =
```
drewhk: I don't think these API methods should be on Source. I would rather create a new object in the io subpackage. We shouldn't cram everything here.

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -185,6 +187,32 @@ object Source {
...
+   * Creates a synchronous (Java 6 compatible) Source from a Files contents.
+   * Emitted elements are `chunkSize` sized [[ByteString]] elements.
+   *
+   * This source is backed by an Actor which will use the dedicated thread-pool base dispatcher.
+   * You can configure the default dispatcher for this Source by changing the `akka.stream.file-io-dispatcher` or
+   * set it for a given Source by using [[ActorOperationAttributes]].
+   *
+   * It materializes a [[Future]] containing the number of bytes read from the source file upon completion.
+   */
+  def synchronousFile(f: File, chunkSize: Int): Source[ByteString, Future[Long]] =
```
drewhk: same as above

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
drewhk: Document that this is not only synchronous but blocking. I know it is clear for us, but better be safe.

drewhk:
I would move the Sink/Source constructor APIs to the io subpackage, having a dedicated object just like Tcp.

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
ktoso: Here it is, true. It remained from different impls. When moved to reading to byteArrays from InputStream it's needed.

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
ktoso: I'm torn. I understand your arguments @jrudolph yet am unable to prove it with a benchmark, currently I'm at (for reading a ~500M file):

```
[info] Benchmark                         (bufSize)  Mode  Cnt    Score     Error  Units
[info] FileSourcesBenchmark.fileChannel       2048  avgt   80  591.803 ±  68.584  ms/op
[info] FileSourcesBenchmark.inputStream       2048  avgt   80  714.225 ± 106.160  ms/op
```
See this commit for code and bench: https://github.com/ktoso/akka/commit/fe822de517845819219b99a1356da7caaae898fe

It's also a question on API - I guess we do want to expose an "file" API, independently how it's implemented. The Input / Output Streams would be additional `Sink.outputStream` I guess? (some of those exist already https://github.com/typesafehub/akka-contrib-extra/tree/master/src/main/scala/akka/contrib/stream )

ktoso:
@jrudolph @rkuhn I have pushed an InputStream backed impl, however am unable to prove it's better.

@drewhk's idea on splitting these out to "IO" is also worth discussing... Do we supply "file", "udp", "inputstream", "outputstream" all on the main objects or do we want to separate them out into some IO object? I personally like `import akka.stream.io_` to add such methods to Source/Sink, but that's perhaps discoverability friendly (as in, it's bad for that :wink:). No strong opinion on creating `StreamIO` or something to group those though.

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
rkuhn: This seems to indicate a performance benefit, and I also follow the point of the `File` API, we should indeed have both, then users can use what suits them. The `InputStream` / `OutputStream` versions should be implementable as simple Iterator and Foreach elements, no?

rkuhn:
+1 for moving to `akka.stream.io` package and turning it around: `FileSource("buh")`, `OutputStreamSink(os)`, …

akka-stream/src/main/scala/akka/stream/impl/SynchronousFilePublisher.scala
jrudolph: @ktoso thanks for the benchmark. I commented more on the additional complexity against a very simple InputStream implementation. If this implementation pays this cost with a performance benefit then I won't object.

ktoso: Sounds like a plan then! And I'll move these sources to the io package as discussed in another place in this PR.
Thanks for keeping an eye on this guys :)

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
ktoso: moving them out

akka-stream/src/main/scala/akka/stream/impl/io/SynchronousFilePublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/** INTERNAL API */
+private[akka] class SynchronousFilePublisher(f: File, bytesReadPromise: Promise[Long], chunkSize: Int, initialBuffer: Int, maxBuffer: Int)
+  extends akka.stream.actor.ActorPublisher[ByteString]
+  with ActorLogging {
+
+  import SynchronousFilePublisher._
+
+  val buffs = new DirectByteBufferPool(chunkSize, maxBuffer)
```
jrudolph: Why is the per-instance pool needed? A pool only makes sense if it is shared between something. Also in light of being reactive IMO the number of buffers just kept around for no good reason should be kept small. You could either use one permanent buffer, or use a shared pool (so the buffers amortize over many usages), or as there's no reason to keep data in the direct pool between invocations even reallocate a buffer for every usage to be extra lean.

akka-stream/src/main/scala/akka/stream/impl/io/SynchronousFilePublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+  var eofReachedAtOffset = Long.MinValue
+
+  var readBytesTotal = 0L
+  var availableChunks: Vector[ByteString] = Vector.empty
```
jrudolph: Why don't you just keep one ByteString and deliver it in one go? IMO the idea to guarantee a maximum chunk size for ByteStrings between streaming stages is a bad precedent that leads people to rely on maximum chunk sizes on the subscriber sides when what they should do is batch process as much data as possible.

A chunk size helps on the input side where you want to guarantee that backpressure actually works by not allowing too much data to enter the system. Holding data back on the output side just prevents batch processing and is an streaming anti-pattern IMO.

akka-stream/src/main/scala/akka/stream/impl/io/SynchronousFilePublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  private var raf: RandomAccessFile = _
+  private var chan: FileChannel = _
+
+  override def preStart() = {
+    try {
+      raf = new RandomAccessFile(f, "r") // best way to express this in JDK6, OpenOption are available since JDK7
+      chan = raf.getChannel
+      readAndSignal(initialBuffer)
```
jrudolph: The pattern to prebuffer one element (to be able to deliver onComplete instantly) may warrant an extra abstraction.

akka-stream/src/main/scala/akka/stream/impl/io/SynchronousFilePublisher.scala
ktoso: You're right, we can ditch it and just allocate a buffer when reading - I'll go with that simple solution for now.

akka-stream/src/main/scala/akka/stream/impl/io/SynchronousFilePublisher.scala
ktoso: If just "start reading to see if can complete" is the need than one can always use a `.buffer(1, backpressure)` after such publisher. The general idea of the file publisher is to always have a bit of data loaded, before `Request(n)` comes in, so two birds with one stone. It's less interesting in this impl because of the synchronous reads... I was implementing an AsynchronousFileChannel based impl where this gets a bit more interesting (though not necessarily faster).

akka-stream/src/main/scala/akka/stream/io/impl/InputStreamPublisher.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  /** BLOCKING I/O READ */
+  def loadChunk() = try {
+    val arr = Array.ofDim[Byte](chunkSize)
+
+    // blocking read
+    val readBytes = is.read(arr)
+
+    readBytes match {
+      case -1 | 0 ⇒
```
huntc: 0 is not EOF - I'm unsure that 0 should be treated as -1

jrudolph: The [Javadoc of InputStream.read](https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#read(byte[],%20int,%20int)) say this:

> This method blocks until input data is available, end of file is detected, or an exception is thrown.

> If len is zero, then no bytes are read and 0 is returned; otherwise, there is an attempt to read at least one byte. If no byte is available because the stream is at end of file, the value -1 is returned; otherwise, at least one byte is read and stored into b.

So, I'd agree with @huntc that reading zero should either be treated as an InvalidStateException or just be ignored to be lenient towards almost working implementations of `InputStream`.

akka-stream/src/main/scala/akka/stream/io/impl/InputStreamPublisher.scala
```diff
@@ -0,0 +1,121 @@
...
+    require(maxBuffer >= initialBuffer, s"maxBuffer must be >= initialBuffer (was $maxBuffer)")
+
+    Props(classOf[InputStreamPublisher], is, completionPromise, chunkSize, initialBuffer, maxBuffer)
+  }
+
+  private final case object Continue extends DeadLetterSuppression
+}
+
+/** INTERNAL API */
+private[akka] class InputStreamPublisher(is: InputStream, bytesReadPromise: Promise[Long], chunkSize: Int, initialBuffer: Int, maxBuffer: Int)
```
huntc: Just curious as to whether you guys sought any inspiration from https://github.com/typesafehub/akka-contrib-extra/blob/master/src/main/scala/akka/contrib/stream/InputStreamPublisher.scala#L36 ?

akka-stream/src/main/scala/akka/stream/io/impl/InputStreamPublisher.scala
ktoso: Good catch, thanks guys

akka-stream/src/main/scala/akka/stream/io/impl/InputStreamPublisher.scala
```diff
@@ -0,0 +1,121 @@
...
+        if (totalDemand > 0) signalOnNexts()
+      }
+    } else if (eofEncountered) onComplete()
+
+  /** BLOCKING I/O READ */
+  def loadChunk() = try {
+    val arr = Array.ofDim[Byte](chunkSize)
+
+    // blocking read
+    val readBytes = is.read(arr)
```
huntc: Should you not use the `blocking` function around blocking calls?

akka-stream/src/main/scala/akka/stream/io/impl/InputStreamPublisher.scala
ktoso: Yes, I know that impl, this is slightly different though - has "bytes read" future etc. It's more a side effect of discussions in this PR about `SynchronousFilePublisher`. If you want a kudos or something we can do that ;-)

huntc: I'm not looking for kudos! Just wanted to ensure that you knew of our implementation more than anything else. It works pretty well. Sounds like you do know of it though, so mission accomplished.

huntc: Incidentally - something to be aware of with these blocked actors: https://github.com/typesafehub/akka-contrib-extra/pull/34

akka-stream/src/main/scala/akka/stream/impl/io/SynchronousFilePublisher.scala
ktoso: I'm a bit torn on this - on one hand, I agree that the "combine many chunks into one bytestring and signal it" is a great idea; on the other hand in order to implement the semantics I think makes sense here it's becoming "weird" in the implementation (much int overflow protection code etc)

To rephrase the idea as I understand it: we can deliver `1` ByteString that is `4*chunkSize` once we get `4` demand, and if more was preloaded we keep that rest around, i.e. the bytestring *is* the buffer. I do not want to treat "bytestring of any size" as "the `element`" however. And I agree that that's a nice idea. 

I'm not sure how much effort to pour into this impl, we could also always document "don't depend on the sizes of elements" and improve/change this later. Note that this implementation will *probably* be thrown out once we can use JDK7+ things.

Opinion on how to proceed @rkuhn?

akka-stream/src/main/scala/akka/stream/io/impl/InputStreamPublisher.scala
huntc: (closing an input/output stream from another thread is not enough - works for OS X, but did not work on Ubuntu.

ktoso: Thanks for the pointer! It seems what you're talking about in that issue is specifically about how System.in / System.out behave? This Source is "any input stream".

huntc: It's really about any input stream really. In essence you've got to kill the thing that feeds it. That's all you can do as closing from another thread won't work on all platforms.

akka-stream/src/main/scala/akka/stream/impl/io/SynchronousFilePublisher.scala
ktoso: Chatted with @drewhk a bit on this.
We came to the conclusion that if you want batching from a Source like this it can be easily built by conflating elements, we even have a ticket for it to ship pre-built: https://github.com/akka/akka/issues/16610

I'd rather keep this implementation as is then. I can change the `Vector[ByteString]` to just a `ByteString` and `take()` from it - looks a bit nicer and same semantics.

akka-stream/src/main/scala/akka/stream/io/impl/InputStreamPublisher.scala
ktoso: it has a dedicated dispatcher.

akka-stream/src/main/scala/akka/stream/impl/io/SynchronousFilePublisher.scala
drewhk: Well, now looking at this, this is an actor. So it won't be fusable with such a batching stage, so you will still suffer the async handoff overhead. Why is this an actor in the first place? Since synchronous API is used, this could be a stage, and then it would be fusable.

jrudolph: @ktoso I see it the other way round :) If you need guaranteed chunk sizes you can add an extra stage providing those. You cannot, however, revert the effect of holding data back by adding another stage. 

Actually, I find it strange (a code smell) that is even possible to hold data back. Reading data ahead to determine the end of the file makes only sense if you cannot know beforehand how much data is in the file. In such a case like when reading from a pipe (or if the file-size changes during reading), reading ahead won't help because it will close spontaneously. And, if you know the file size in advance you don't have to read ahead.

drewhk: Yes, but this stage has a natural chunk size that is the read buffer. Why would it batch up more, what would it gain? Also, how would you interpret a demand of 4 for example?

jrudolph: I don't know, I only see an `availableChunks` buffer in the code which wouldn't be needed if it wouldn't batch. `request(4)` is the same as four times `request(1)`.

drewhk: well, yes but that does not answer the question. What does then request of one translates to? There must be an upper bound on the emitted ByteStrings, it is definitely a no-go to slurp the whole file into one ByteString. The only question is where to put that chunk threshold. Currently it is the size of the read buffer, we can make it a different threshold though. Of course we can for example read ahead while we have a demand of zero, up to some threshold to prefetch data before it is demanded.

ktoso: I feel we mixed two questions into one somehow. 
1) if the signalled chunk is of known size, 
2) if we read ahead 
(whoa even 3) if this should be a stage or an actor)

The current impl has:
1) known chunk size – the read buffer, someone somewhere will set this value, it's not magic so people could end up depending on it even if not exposed as parameter – (does not concat multiple reads into
2) reads ahead, trying to lessen the delay from request to getting data. Maybe not needed - your points on when it would kick in and may be surprising are good ones.
3) is an Actor. When I started this impl it was the JDK7 thingies (async) and stayed an actor. I can easily make it a Stage, however: in a stage I do not get the `Request(N)` explicitly (unless I missed something in StatefulStage API), and don't know how many chunks to read+collapse.

I don't see why/how I'd want to avoid known chunk size (1), *unless* we do the collapsing we talked about. If we do not read ahead we could "read Demand(N) chunks" collapse them into 1 signal and send it (depleting demand). Now we need to figure out how much we want to read at a given time - so the "how many chunks" needs to be configurable (it's a `maxChunkSize` again) - otherwise I'm stuck reading "forever".

(I need to pick up my car from the workshop, be right back - around 30minutes :-))

jrudolph: @ktoso Yes, I agree. My last comment was about 2). And, yes, of course, you need a chunk size for 1). Maybe then it's just about code organization. If you take away read-ahead the code should collapse into a few lines. If you want read-ahead, as someone already commented, you can easily add it with a buffer stage.

drewhk: > 1) known chunk size – the read buffer, someone somewhere will set this value, it's not magic so people could end up depending on it even if not exposed as parameter – (does not concat multiple reads into

My point here that there must be a limit, either because you don't read ahead, and in this case you cannot return more than the read buffer contains, or because you read ahead, but you don't want to slurp the whole universe.

> 2) reads ahead, trying to lessen the delay from request to getting data. Maybe not needed - your points on when it would kick in and may be surprising are good ones

I believe in the read-ahead because it allows pipelining IO and CPU bound tasks. Even if we don't read ahead when having no demand, a request larger than 1 unit will imply a read-ahead anyway.

> 3) is an Actor. When I started this impl it was the JDK7 thingies (async) and stayed an actor. I can easily make it a Stage, however: in a stage I do not get the Request(N) explicitly (unless I missed something in StatefulStage API), and don't know how many chunks to read+collapse.

In the stage case you would not be able to batch, but the read-ahead will naturally arise from being a BoundaryStage at the end and having demands more than 1 from downstream. If it would be fused with other stages, then those stages would not get read-ahead, because they share the same synchronous island. I actually find this desirable, since when we will enable fusing support you can choose if you want pipelined async execution, or not pipelined sync execution.

jrudolph: @drewhk is a fusable stage allowed to make blocking calls?

drewhk: It is allowed, but usually not wise :) It will block the encapsulating actor. If you fuse multiple stages into one actor, then one blocking call would block the whole island together, which is usually fine, but can be very surprising with detached and async stages. I believe though that there are valid use cases to fuse certain stages together with a blocking one.

The intended API will require explcit denotations of fusing boundaries, so it will not be automatically done, only when you mark a composite module (graph) eligible for fusing. There might be also annotations to prevent fusing for certain stages (which would override the inherited PleaseFuse attribute from the encapsulating composite). But this is not yet really designed. We need to build the internal layout introspection/rewriting support, I would like to add a more flexible n-to-n port fusable abstraction (fusable graph stages) and then we can do it.

So for example in the case of this Source, we can set the default attribute to prohibit fusing, so people cannot accidentally put it in a larger graph and fuse it, but users will be able to override it by changing the attribute of the Source itself. This is just speculation currently, I don't have a proper design.

ktoso: Back from the car-repair-shop and investigated the options we're discussing with adding them to the "drain the file" benchmark:

```
[info] FileSourcesBenchmark.fileChannel                   2048  avgt  100  1217.977 ±  61.751  ms/op
[info] FileSourcesBenchmark.fileChannel_noReadAhead       2048  avgt  100  1921.258 ± 116.078  ms/op
[info] FileSourcesBenchmark.fileChannel_stage       2048  avgt   40  2864.094 ± 450.808  ms/op
```

So the naive/ugly readahead does help in those scenarios, the Stage is not very stellar at this point in time (no fusing in action here). 

Let me try to wrap this up (and assuming realistic time bounds):
- we do not have fine control over fusing / not fusing yet. The win by using a stage would be fusing. Translating this guy into a Stage is not very hard if we want to do it later on I believe.
- making the "read ahead as buffer in front of it" works when we fuse, otherwise it's just another async boundary (does not help with perf, I measured that).
- If we want to do the "read ahead and collapse into 1 signal", I need to extend the `ActorPublisher` with `onNext(x, demandDrained = 10)`. I'm not sure we want to expose this as such... It is indeed something more fancy that the AP can do but the Stage not so much. I'm not convienced it's worth it for the 1st impl of this anyway though... 

Summing up:
- if we want a Stage at some point, it's a 15 minute task to change to it from the Actor.
- current best perf for "drain dat file" is the naive readahead (and I think it could be improved) 
- At least the Actor isolates the IO from the rest of the world nicely, and theres the default file-io dispatcher now.

So: how about we go with the simplest impl for now and revisit the topic once fusing is more real? If we don't like the readahead's current form I'm OK with removing it, I'd keep it as an Actor though - sounds good?

----

On a related note, this is what people are currently doing:

```
Source(() ⇒ scala.io.Source.fromFile(file).getLines())
// FileSourcesBenchmark.naive_ioSourceLinesIterator  avgt   20  7067.944 ± 1341.847  ms/op
```

So whichever impl already is very noticable improvement :wink:

ktoso:
Discussion in-flight here: https://github.com/akka/akka/pull/17211#discussion_r28886575 (adding link because github folded that thread)

jrudolph:
@ktoso thanks for sharing the choices and the benchmark.

drewhk:
Keep it as it is. We need to investigate what causes the stage version to be that slow later.

ktoso:
I'll squash and create tickets to follow up on.

ktoso:
Thanks a lot for all the comments guys! I squashed with what seems to be the current consensus and opened a ticket for further discussion and improvements.

ktoso:
failure was weird: `sbt.ResolveException: unresolved dependency: org.scala-lang#scala-library;2.10.5: not found`

ktoso:
PLS BUILD

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/io/StreamFileDocTest.java
```diff
@@ -0,0 +1,83 @@
...
+  @Test
+  public void demonstrateSimpleServerConnection() {
+    final File file = File.createTempFile(getClass().getName(), ".tmp");
+
+    //#file-source
+    SynchronousFileSource.create(file)
+      .runForeach(chunk -> System.out.println(chunk.utf8String()), mat);
+    //#file-source
+
+    file.delete();
```
patriknw: would be nice to have this in finally, or in tearDown

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSinkSpec.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  "OutputStreamSink" must {
+    "write bytes to void OutputStream" in checkThatAllStagesAreStopped {
+      val p = TestProbe()
+      val datas = List(ByteString("a"), ByteString("c"), ByteString("c"))
+
+      val completion = Source(datas)
+        .runWith(OutputStreamSink(() ⇒ new OutputStream {
+          override def write(i: Int): Unit = ()
+          override def write(bytes: Array[Byte]): Unit = p.ref ! ByteString(bytes).utf8String
```
patriknw: nice

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSinkSpec.scala
```diff
@@ -0,0 +1,66 @@
...
+        }))
+
+      p.expectMsg(datas(0).utf8String)
+      p.expectMsg(datas(1).utf8String)
+      p.expectMsg(datas(2).utf8String)
+      Await.ready(completion, 300.millis)
```
patriknw: why so tight timeout?

akka-stream/src/main/scala/akka/stream/actor/ActorSubscriber.scala
```diff
@@ -190,9 +190,10 @@ trait ActorSubscriber extends Actor {
     case OnSubscribe(sub) ⇒
       if (subscription.isEmpty) {
         subscription = Some(sub)
-        if (canceled)
+        if (canceled) {
           sub.cancel()
-        else if (requested != 0)
+          context.stop(self)
```
patriknw: move up one line, in case sub.cancel() throws

akka-stream/src/main/scala/akka/stream/io/InputStreamSource.scala
```diff
@@ -0,0 +1,52 @@
...
+import akka.stream.scaladsl.Source
+import akka.stream.scaladsl.Source._
+import akka.stream.{ OperationAttributes, javadsl }
+import akka.util.ByteString
+
+import scala.concurrent.Future
+
+object InputStreamSource {
+
+  final val DefaultChunkSize = 2048
```
patriknw: default buffer size of BufferedInputStream is 8192, would we benefit from using the same?

akka-stream/src/main/scala/akka/stream/io/InputStreamSource.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Java API
+   *
+   * Creates a Source that will pull data out of the given input stream.
+   * Emitted elements are `chunkSize` sized [[ByteString]] elements.
+   *
+   * It materializes a [[Future]] containing the number of bytes read from the source file upon completion.
+   */
+  def create(createInputStream: javadsl.japi.Creator[InputStream], chunkSize: Int = 2048): javadsl.Source[ByteString, Future[Long]] =
```
patriknw: s/2048/DefaultChunkSize/

patriknw: btw, the default value should be removed here, it's java api

ktoso: Ouch, yes - thanks!

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/io/StreamFileDocTest.java
ktoso: :+1:

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSinkSpec.scala
ktoso: No good reason, can give default 3seconds

akka-stream/src/main/scala/akka/stream/actor/ActorSubscriber.scala
ktoso: +1, thanks

akka-stream/src/main/scala/akka/stream/io/impl/InputStreamPublisher.scala
```diff
@@ -0,0 +1,121 @@
...
+    case ex: Exception ⇒
+      onError(ex)
+  }
+
+  private final def eofEncountered: Boolean = eofReachedAtOffset != Long.MinValue
+
+  override def postStop(): Unit = {
+    bytesReadPromise.trySuccess(readBytesTotal)
+
+    if (is ne null) is.close()
```
patriknw: `close` may throw, so it might be safer to do that last, in case super does something important

akka-stream/src/main/scala/akka/stream/io/InputStreamSource.scala
ktoso: I took 2048 based on `scala.io.Source`. Larger as you suggest is probably even better... running a quick bench.

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSinkSpec.scala
patriknw: thanks, it's just labor consuming with sporadic jenkins failures

akka-stream/src/main/scala/akka/stream/io/InputStreamSource.scala
patriknw: I was just thinking that most input streams will be BufferedInputStream, and there might be some advantage of using same size (but I'm just guessing)

ktoso: Let's go with 8192 as default, for the file case it reduced the jitter a bit

```
8192
[info] Benchmark                                     (bufSize)  Mode  Cnt     Score     Error  Units
[info] FileSourcesBenchmark.fileChannel                   2048  avgt   40  1276.648 ± 107.406  ms/op
[info] FileSourcesBenchmark.fileChannel_noReadAhead       2048  avgt   40  1650.780 ± 115.605  ms/op

2048

[info] Benchmark                                     (bufSize)  Mode  Cnt     Score     Error  Units
[info] FileSourcesBenchmark.fileChannel                   2048  avgt   40  1172.119 ±  66.854  ms/op
[info] FileSourcesBenchmark.fileChannel_noReadAhead       2048  avgt   40  1629.711 ± 208.312  ms/op
```

akka-stream/src/main/scala/akka/stream/io/impl/InputStreamPublisher.scala
ktoso: thanks

ktoso:
Updated according to comments and rebased. Looking forward to merge after validation.

akka-stream/src/main/scala/akka/stream/io/InputStreamSource.scala
jrudolph: 8192 is the maximum size where the implementation of java.io will still try to allocate the intermediate buffer on the stack. For bigger sizes it will use malloc:

http://hg.openjdk.java.net/jdk7/jdk7/jdk/file/9b8c96f96a0f/src/share/native/java/io/io_util.c#l96

I guess that's why BufferInputStream also uses this size. FileChannel, however, seems to use another call path.

ktoso:
PLS BUILD

ktoso:
PLS BUILD

ktoso:
Merge failuse on jenkins, rebased and pushed - hopefully last time.

patriknw:
LGTM (in case I didn't say that before)

ktoso:
Thanks. :-)
I'll go ahead and merge as it seems you need to squash/rebase https://github.com/akka/akka/pull/17217 anyway?

ktoso:
Failure was unrelated `[error] 	akka.http.engine.client.ConnectionPoolSpec`, merging and will open tickef that that test failed to investigate.

PR #17212, opened 2015-04-16
=str #16986 Fix memory leak in PrefixAndTail when using Sink.publisher

patriknw:
The problem was reproduced with the TCK PrefixAndTailTest
required_spec313_cancelMustMakeThePublisherEventuallyDropAllReferencesToTheSubscriber
The tck subscriber was still referenced. Profiling revealed that the
root cause was the VirtualPublisher that holds a reference to the
realPublisher, which was MultiStreamOutputProcessor$SubstreamOutput,
which had the reference to the subscriber. The VirtualPublisher
is created by the Sink.publisher in the test, and the test holds
on to that VirtualPublisher reference.

The solution is to null out realPublisher field in the VirtualPublisher.

The old workaround with the NullSubscriber was removed.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
@@ -126,6 +126,7 @@ private[akka] case class ActorFlowMaterializerImpl(override val settings: ActorF
             }
             val impl = actorOf(props, stageName(effectiveAttributes), effectiveSettings.dispatcher)
             val publisher = new ActorPublisher[Any](impl)
+            // Resolve cyclic dependency with actor. This MUST be the first message no matter what.
             impl ! ExposedPublisher(publisher)
             for ((in, id) ← inputs.zipWithIndex) {
               assignPort(in, FanIn.SubInput[Any](impl, id))
@@ -281,6 +282,7 @@ private[akka] object ActorProcessorFactory {
 
   def apply[I, O](impl: ActorRef): ActorProcessor[I, O] = {
     val p = new ActorProcessor[I, O](impl)
+    // Resolve cyclic dependency with actor. This MUST be the first message no matter what.
     impl ! ExposedPublisher(p.asInstanceOf[ActorPublisher[Any]])
     p
   }
--- a/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
@@ -311,7 +311,13 @@ private[stream] class VirtualSubscriber[T](val owner: VirtualPublisher[T]) exten
  */
 private[stream] class VirtualPublisher[T]() extends Publisher[T] {
   @volatile var realPublisher: Publisher[T] = null
-  override def subscribe(s: Subscriber[_ >: T]): Unit = realPublisher.subscribe(s)
+  override def subscribe(s: Subscriber[_ >: T]): Unit = {
+    val sub = realPublisher.subscribe(s)
+    // unreference the realPublisher to facilitate GC and
+    // Sink.publisher is supposed to reject additional subscribers anyway
+    realPublisher = RejectAdditionalSubscibers[T]
+    sub
+  }
 }
 
 /**
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowFromFutureSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowFromFutureSpec.scala
@@ -67,7 +67,7 @@ class FlowFromFutureSpec extends AkkaSpec {
 
     "produce elements with multiple subscribers" in {
       val promise = Promise[Int]()
-      val p = Source(promise.future).runWith(Sink.publisher)
+      val p = Source(promise.future).runWith(Sink.fanoutPublisher(1, 1))
       val c1 = StreamTestKit.SubscriberProbe[Int]()
       val c2 = StreamTestKit.SubscriberProbe[Int]()
       p.subscribe(c1)
@@ -83,30 +83,9 @@ class FlowFromFutureSpec extends AkkaSpec {
       c2.expectComplete()
     }
 
-    "produce elements to later subscriber" in {
-      val promise = Promise[Int]()
-      val p = Source(promise.future).runWith(Sink.publisher)
-      val keepAlive = StreamTestKit.SubscriberProbe[Int]()
-      val c1 = StreamTestKit.SubscriberProbe[Int]()
-      val c2 = StreamTestKit.SubscriberProbe[Int]()
-      p.subscribe(keepAlive)
-      p.subscribe(c1)
-
-      val sub1 = c1.expectSubscription()
-      sub1.request(1)
-      promise.success(1)
-      c1.expectNext(1)
-      c1.expectComplete()
-      p.subscribe(c2)
-      val sub2 = c2.expectSubscription()
-      sub2.request(1)
-      c2.expectNext(1)
-      c2.expectComplete()
-    }
-
     "allow cancel before receiving element" in {
       val promise = Promise[Int]()
-      val p = Source(promise.future).runWith(Sink.publisher)
+      val p = Source(promise.future).runWith(Sink.fanoutPublisher(1, 1))
       val keepAlive = StreamTestKit.SubscriberProbe[Int]()
       val c = StreamTestKit.SubscriberProbe[Int]()
       p.subscribe(keepAlive)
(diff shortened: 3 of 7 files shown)


patriknw:
Refs #16986

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -311,7 +311,16 @@ private[stream] class VirtualSubscriber[T](val owner: VirtualPublisher[T]) exten
  */
 private[stream] class VirtualPublisher[T]() extends Publisher[T] {
   @volatile var realPublisher: Publisher[T] = null
-  override def subscribe(s: Subscriber[_ >: T]): Unit = realPublisher.subscribe(s)
+  override def subscribe(s: Subscriber[_ >: T]): Unit = {
+    val sub = realPublisher.subscribe(s)
+    realPublisher match {
+      case _: SupportingMultipleSubscribers   ⇒
```
drewhk: Isn't this completely Akka specific now? I mean, if you take a Source(thirdPartyPublisher).runWith(Sink.publisher) then this will remove the multi-subscribe support from thirdPartyPublisher.

drewhk:
I am not sure about this. Why is someone holding to the VirtualPublisher? Why is that considered a leak? Maybe it is, but it needs a little bit more explanation.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
patriknw: That is the case as soon as you add any stage to the flow. One could argue that Sink.publisher must always return a single-subscription publisher. We have even documented it as "can handle one [[org.reactivestreams.Subscriber]]".

I can turn the marker interface the other way around if Endre's point is considered a problem. I just thought it would be more places to mark.

patriknw:
The problem is that we don't release the reference to the subscriber when it is cancelled (until the publisher is released), and that is not allowed by reactive streams spec and tck.

I agree that it is a bit hard of the tck to keep the reference to the publisher, but that is the kind of mistakes that users do also, so I don't think it hurts that we help them here.

drewhk:
My problem here is that it does not seem to fix the root issue. Currently the references look like this:

    X -> V -> P -> S

Where V is the VirtualPublisher, P is the wrapped Publisher and S is the Subscriber. This PR adds a band-aid by removing the V -> P link, but it does not feel like a real solution. The existence of V is simply due to the fact that I was not able to simplify the API and at the same time return the original publisher P. In other words, in an ideal case V would not exist and the chain would be:

    X -> P -> S

The real leak in my eyes is the P -> S link, not the V -> P since it just adds a simple indirection, it does not retain anything the user wouldn't if V would not exist.

patriknw:
ok, but then we are back to my previous workaround, that breaks the `P -> S` link with the `NullSubscriber` when cancellation arrives.

The advantage of breaking the link in `V` is that we release a lot of more objects, i.e. the whole `PrefixAndTailImpl` and everything it holds on to.

Anyway, aren't you planning on rewriting `PrefixAndTailImpl` and friends anyway since it is based on the pump infrastructure, and then this (or the previous `NullSubscriber`) could be good enough for RC1?

drewhk:
It is ok by me to have this in RC1. Just make sure then to file a ticket to revisit this issue later.

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
```diff
@@ -284,4 +284,10 @@ private[akka] object ActorProcessorFactory {
     impl ! ExposedPublisher(p.asInstanceOf[ActorPublisher[Any]])
     p
   }
+
+  def supportingMultipleSubscribers[I, O](impl: ActorRef): ActorProcessor[I, O] = {
+    val p = new ActorProcessor[I, O](impl) with SupportingMultipleSubscribers
+    impl ! ExposedPublisher(p.asInstanceOf[ActorPublisher[Any]])
```
ktoso: wouldn't hurt to have the `// Resolve cyclic dependency with actor. This MUST be the first message no matter what.` here as well I guess

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: It is late and I might be missing something, but why don’t we implement what @patriknw says and remove the SupportingMultipleSubscribers marker entirely? Why should some VirtualPublishers support multi-subscribe?

patriknw: What I have implemented here with a volatile var is not enough for "must always return a single-subscription publisher". My implementation here is aiming at clearing the reference as soon as possible but it can still be there for concurrent subscribe calls. Just saying.

rkuhn: Well, what I was getting at is that “roughly one subscriber” is good enough for our purposes, so we might want to keep it simple.

patriknw: ok, I'll take another look at what happens when I simply remove the `SupportingMultipleSubscribers` marker, but I think I added it for the reason that we have tests that will fail otherwise. E.g. FuturePublisher, EmptyPublisher and SynchronousIterablePublisher have native support for more than one subscriber.

drewhk:
So, I guess it is OK for now, but I think this will need eventually a better solution.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
patriknw: @rkuhn I would like that we decide what to do here, so I can merge.

The following tests fail if I remove the SupportingMultipleSubscribers marker:

* Failed Source must emit error immediately
* Empty Source must complete immediately
* Singleton Source must produce elements to later subscriber
* A Flow based on a Future must produce elements with multiple subscribers
* A Flow based on a Future must produce elements to later subscriber
* A Flow based on a Future must allow cancel before receiving element

If we remove the marker I have to remove or rewrite those tests.

rkuhn: Hmm. Merge as is and add a ticket, please.

drewhk: This is why I disliked this change in the first place - it removes a feature from the wrapped publisher (the multi-subscriber feature). The marker interface fixes the akka provided publishers, but only those. Another option could be to mark the single-subscriber things instead, in this case we don't change the semantics of 3rd party publishers.

rkuhn: Just to prevent misunderstandings: what I want eventually is that the behavior is consistently “single-subscriber”, i.e. the removal of that feature is completely fine. I just don’t want to spend the effort right now of figuring out why those unrelated tests fail as a result (with the exception of one case that tests for a feature that shouldn’t be there).

drewhk: Even if you go for "consistently single-subscriber" you should not remove the feature from fanoutPublisher since that is the whole point of that Sink. I guess the tests fail because they rely on the fact that the exposed publisher supports the multi-subscriber scenario the original publisher provides.

patriknw: exactly, I think it can be rather surprising for users that `Source(p).runWith(Sink.publisher)` may work with multiple subscribers, but `Source(p).map(identity).runWith(Sink.publisher)` doesn't.

PS. Why would one use  `Source(p).runWith(Sink.publisher)` in the first place?

patriknw: ticket created: #17245
I will merge this as is.

drewhk: I am not sure if it is surprising or not. Anyway, if we really want to go the one-subscription enforced way, then we only need to special case the fanoutPublisher. All other test failures test the wrong thing simply.

rkuhn: I still don’t understand this point: the `fanoutPublisher` is a Sink that materializes to a Publisher that supports multiple subscribers. What we are discussing here has nothing to do with this, it is about how we pass along Publishers internally. The output is again determined by the Sink the user attaches, and if that is a `Sink.publisher`, why should it ever support multiple subscriptions?

drewhk: ah yes, you are right, fanoutPublisher is a proper sink, it will not use the virtualpublisher.

drewhk: btw, it is not even working to do Source(p).runWith(Sink.publisher), and the solution will put an Identity stage there anyway, so I guess the question is moot. We can simply make VirtualPublisher a single-subscriber one and fix/remove the failing tests.

patriknw: ok, but since we still have some publishers that supports multiple subscribers and here mark them as such, we should do that with the fanoutPublisher also to be consistent. I missed that. That change is: 19aa5eb
Ok?

I think making Sink.publisher single subscriber only is the right way to go, but that is for #17245

patriknw: alright, I'll change the tests!

patriknw:
@rkuhn @drewhk I have removed the marker interface and rewritten the failing tests.

rkuhn:
Thanks, Patrik, LGTM.

Issue #17168, opened 2015-04-10
avoid copy in Props.withDispatcher


patriknw:
if the given dispatcher is the same as current value

PR #17215, opened 2015-04-17
=act #17168 avoid copy in Props for the same mailbox or dispatcher

He-Pin:
refs https://github.com/akka/akka/issues/17168

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/Props.scala
+++ b/akka-actor/src/main/scala/akka/actor/Props.scala
@@ -4,20 +4,18 @@
 
 package akka.actor
 
+import java.lang.reflect.{ Constructor, Modifier, ParameterizedType, TypeVariable }
+
+import akka.actor.Deploy.{ NoDispatcherGiven, NoMailboxGiven }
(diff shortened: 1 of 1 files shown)


patriknw:
OK TO TEST

akka-actor/src/main/scala/akka/actor/Props.scala
```diff
@@ -215,12 +213,12 @@ final case class Props(deploy: Deploy, clazz: Class[_], args: immutable.Seq[Any]
   /**
    * Returns a new Props with the specified dispatcher set.
    */
-  def withDispatcher(d: String): Props = copy(deploy = deploy.copy(dispatcher = d))
+  def withDispatcher(d: String): Props = if (d == dispatcher) this else copy(deploy = deploy.copy(dispatcher = d))
```
patriknw: I think the test must be on `deploy.dispatcher` to account for `NoDispatcherGiven` vs `DefaultDispatcherId`

rkuhn:
LGTM, thanks!

PR #17218, opened 2015-04-17
=htc #16795 add missing test

### LINKED ISSUE #16795: HTTP client not robust
client deadlocks after a few requests, see https://groups.google.com/d/msg/akka-user/Gp0TezTei94/x5py6rcY1RIJ

sirthias:
Closes #16795.

/cc @jrudolph

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/engine/client/HighLevelOutgoingConnectionSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/engine/client/HighLevelOutgoingConnectionSpec.scala
@@ -1,0 +1,71 @@
+/*
+ * Copyright (C) 2009-2014 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.engine.client
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import akka.http.{ Http, TestUtils }
+import akka.http.model.HttpEntity._
+import akka.http.model._
+import akka.stream.ActorFlowMaterializer
+import akka.stream.scaladsl._
+import akka.stream.testkit.AkkaSpec
+
+class HighLevelOutgoingConnectionSpec extends AkkaSpec("akka.loggers = []\n akka.loglevel = OFF") {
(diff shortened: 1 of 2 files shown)


sirthias:
Failed seems unrelated (in `akka.stream.scaladsl.GraphPreferredMergeSpec`).

rkuhn:
LGTM, but does it really need 1000 requests? Takes 7 sec on Jenkins …

sirthias:
Ok, I'll reduce it to 100 requests. Should be almost as effective.

> Takes 7 sec on Jenkins

On my machine (i7-4960HQ CPU @ 2.60GHz) it takes 3.8 sec.
And here is a pointer to a nice "actual iron" 4-core machine with about the same [CPU PassMark](https://www.cpubenchmark.net/high_end_cpus.html) score as mine, 32 GB RAM, 2 x 240 GB SSD and 30 TB of traffic for 59 EUR/month:
https://www.hetzner.de/hosting/produkte_rootserver/ex40ssd

I bet this is a fraction of what you are currently paying to amazon, for at least twice the oomph.

It might make sense to go for even more iron, like a 6-core for 109 EUR/month (https://www.hetzner.de/de/hosting/produkte_rootserver/px90ssd) or an 8-core for 189 EUR/month (https://www.hetzner.de/de/hosting/produkte_rootserver/dx151), and make Jenkins deliver those PR validations a lot more presto.

sirthias:
Reduced to 100 test requests to not be so hard on Jenkins, the old man.

rkuhn:
«Thanks!», said the grumpy old gaffer.

jrudolph:
> I bet this is a fraction of what you are currently paying to amazon, for at least twice the oomph.

It seems the 4770 would be roughly comparable to a c3.2xlarge - c4.2xlarge instance in terms of CPU power which currently is about 200 $/month for a reserved instance.

patriknw:
thanks for suggestions, fyi we are currently running this on our own old servers, so they are pretty cheap (they are not fast, but they have many cores)

PR #17219, opened 2015-04-17
+clu #16736 add registerOnMemberRemoved to get notified when current mem...

He-Pin:
refs https://github.com/akka/akka/issues/16736
add registerOnMemberRemoved
and I think maybe add support for the 
`cluster.join`,`cluster.down` and `cluster.leave` to return a future should be nice.
like

```
join(address:Address):Future[Address]
down(address:Address):Future[Address]
leave(address:Address):Future[Address]
```

### DIFF
--- a/akka-cluster/src/main/scala/akka/cluster/Cluster.scala
+++ b/akka-cluster/src/main/scala/akka/cluster/Cluster.scala
@@ -4,32 +4,24 @@
 
 package akka.cluster
 
-import language.implicitConversions
-import akka.actor._
-import akka.actor.Status._
-import akka.ConfigurationException
-import akka.dispatch.MonitorableThreadFactory
-import akka.event.Logging
-import akka.pattern._
-import akka.remote._
-import akka.routing._
-import akka.util._
-import scala.concurrent.duration._
-import scala.concurrent.forkjoin.ThreadLocalRandom
-import scala.annotation.tailrec
-import scala.collection.immutable
 import java.io.Closeable
-import java.util.concurrent.atomic.AtomicBoolean
-import java.util.concurrent.atomic.AtomicReference
-import scala.concurrent.{ ExecutionContext, Await }
-import com.typesafe.config.ConfigFactory
-import akka.remote.DefaultFailureDetectorRegistry
-import akka.remote.FailureDetector
-import com.typesafe.config.Config
-import akka.event.LoggingAdapter
 import java.util.concurrent.ThreadFactory
-import scala.util.control.NonFatal
+import java.util.concurrent.atomic.AtomicBoolean
+
+import akka.ConfigurationException
+import akka.actor._
+import akka.dispatch.MonitorableThreadFactory
+import akka.event.{ Logging, LoggingAdapter }
+import akka.pattern._
+import akka.remote.{ DefaultFailureDetectorRegistry, FailureDetector, _ }
+import com.typesafe.config.{ Config, ConfigFactory }
+
 import scala.annotation.varargs
+import scala.collection.immutable
+import scala.concurrent.duration._
+import scala.concurrent.{ Await, ExecutionContext }
+import scala.language.implicitConversions
+import scala.util.control.NonFatal
 
 /**
  * Cluster Extension Id and factory for creating Cluster extension.
@@ -65,8 +57,8 @@ class Cluster(val system: ExtendedActorSystem) extends Extension {
   import ClusterEvent._
 
   val settings = new ClusterSettings(system.settings.config, system.name)
-  import settings._
   import InfoLogger._
+  import settings._
 
   /**
    * The address including a `uid` of this cluster member.
@@ -119,7 +111,6 @@ class Cluster(val system: ExtendedActorSystem) extends Extension {
    */
   private[cluster] val scheduler: Scheduler = {
     if (system.scheduler.maxFrequency < 1.second / SchedulerTickDuration) {
-      import scala.collection.JavaConverters._
       logInfo("Using a dedicated scheduler for cluster. Default scheduler can be used if configured " +
         "with 'akka.scheduler.tick-duration' [{} ms] <=  'akka.cluster.scheduler.tick-duration' [{} ms].",
         (1000 / system.scheduler.maxFrequency).toInt, SchedulerTickDuration.toMillis)
@@ -314,7 +305,7 @@ class Cluster(val system: ExtendedActorSystem) extends Extension {
    * a certain size.
    */
   def registerOnMemberUp[T](code: ⇒ T): Unit =
-    registerOnMemberUp(new Runnable { def run = code })
+    registerOnMemberUp(new Runnable { def run() = code })
 
   /**
    * Java API: The supplied callback will be run, once, when current cluster member is `Up`.
@@ -322,8 +313,27 @@ class Cluster(val system: ExtendedActorSystem) extends Extension {
    * to defer some action, such as starting actors, until the cluster has reached
    * a certain size.
    */
-  def registerOnMemberUp(callback: Runnable): Unit = clusterDaemons ! InternalClusterAction.AddOnMemberUpListener(callback)
+  def registerOnMemberUp(callback: Runnable): Unit =
+    clusterDaemons ! InternalClusterAction.AddOnMemberUpListener(callback)
+  /**
+   * The supplied thunk will be run, once, when current cluster member is `Removed`.
+   * and if the cluster have been shutdown,that thunk will run on the caller thread immediately.
+   * Typically used together `cluster.leave(cluster.selfAddress)` and then `system.shutdown()`.
+   */
+  def registerOnMemberRemoved[T](code: ⇒ T): Unit =
+    registerOnMemberRemoved(new Runnable { override def run(): Unit = code })
 
+  /**
+   * Java API: The supplied thunk will be run, once, when current cluster member is `Removed`.
+   * and if the cluster have been shutdown,that thunk will run on the caller thread immediately.
+   * Typically used together `cluster.leave(cluster.selfAddress)` and then `system.shutdown()`.
+   */
+  def registerOnMemberRemoved(callback: Runnable): Unit = {
+    if (_isTerminated.get())
+      callback.run()
+    else
+      clusterDaemons ! InternalClusterAction.AddOnMemberRemovedListener(callback)
+  }
   // ========================================================
   // ===================== INTERNAL API =====================
   // ========================================================
--- a/akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
+++ b/akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
@@ -129,9 +129,11 @@ private[cluster] object InternalClusterAction {
 
   /**
    * Comand to [[akka.cluster.ClusterDaemon]] to create a
-   * [[akka.cluster.OnMemberUpListener]].
+   * [[akka.cluster.OnMemberStatusChangedListener]].
    */
   final case class AddOnMemberUpListener(callback: Runnable) extends NoSerializationVerificationNeeded
+
+  final case class AddOnMemberRemovedListener(callback: Runnable) extends NoSerializationVerificationNeeded
 
   sealed trait SubscriptionMessage
   final case class Subscribe(subscriber: ActorRef, initialStateMode: SubscriptionInitialStateMode, to: Set[Class[_]]) extends SubscriptionMessage
@@ -165,7 +167,9 @@ private[cluster] final class ClusterDaemon(settings: ClusterSettings) extends Ac
   def receive = {
     case msg: GetClusterCoreRef.type ⇒ coreSupervisor forward msg
     case AddOnMemberUpListener(code) ⇒
-      context.actorOf(Props(classOf[OnMemberUpListener], code).withDeploy(Deploy.local))
+      context.actorOf(Props(classOf[OnMemberStatusChangedListener], code, Up).withDeploy(Deploy.local))
+    case AddOnMemberRemovedListener(code) ⇒
+      context.actorOf(Props(classOf[OnMemberStatusChangedListener], code, Removed).withDeploy(Deploy.local))
     case PublisherCreated(publisher) ⇒
       if (settings.MetricsEnabled) {
         // metrics must be started after core/publisher to be able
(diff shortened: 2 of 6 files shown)


akka-cluster/src/main/scala/akka/cluster/Cluster.scala
```diff
@@ -314,7 +314,7 @@ class Cluster(val system: ExtendedActorSystem) extends Extension {
    * a certain size.
    */
   def registerOnMemberUp[T](code: ⇒ T): Unit =
```
hepin1989: I think the signature should be ```def registerOnMemberUp(code: ⇒ Unit): Unit ```,I don't know why it need a ```T``` here,any pointer?

patriknw:
OK TO TEST

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -1114,21 +1118,30 @@ private[cluster] final class JoinSeedNodeProcess(seedNodes: immutable.IndexedSeq
...
-private[cluster] class OnMemberUpListener(callback: Runnable) extends Actor with ActorLogging {
+private[cluster] class OnMemberStatusChangedListener(callback: Runnable, prefer: MemberStatus) extends Actor with ActorLogging {
   import ClusterEvent._
-  val cluster = Cluster(context.system)
-  // subscribe to MemberUp, re-subscribe when restart
+  private val cluster = Cluster(context.system)
+  private val to = prefer match {
+    case Up ⇒
+      classOf[MemberUp]
+    case Removed | Down ⇒
```
rkuhn: There seem to be remnants of another feature in here as well: `Down` is not enabled for tracking by this PR.

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -1140,8 +1153,8 @@ private[cluster] class OnMemberUpListener(callback: Runnable) extends Actor with
     }
   }
 
-  def isSelfUp(m: Member): Boolean =
-    m.uniqueAddress == cluster.selfUniqueAddress && m.status == MemberStatus.Up
+  private def checkChanged(m: Member, previousStatus: MemberStatus = null): Boolean =
+    m.uniqueAddress == cluster.selfUniqueAddress && (if (prefer == Down) previousStatus == prefer else m.status == prefer)
```
rkuhn: I’d suggest to remove the `Down` handling, which means getting rid of `previousState` as well, and then changing the name of this method to `isTriggered` because that is what it does.

rkuhn:
Thanks a lot, @hepin1989!

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
hepin1989: In fact I think it would be nice to make the 
```down(address:Address)```
```leave(address:Address)```
```join(address:Address)```
to return a ```Future```
or add method like ```leave(address:Address,timeout:FinitTimeOut)``` which return a Future,so I add that here.
wdyt?

Comment (rkuhn) [inline akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala]: Hmm, programmatic actions like the ones you cite are not the only ways to join or leave a cluster, therefore I’d rather have the generic hooks than future-returning actions. It is pretty trivial to write an extension that then uses these hooks to provide one Future for each purpose.

Comment (patriknw) [inline akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala]: yes, Down is not published as a cluster event, since it is a transition for "user" command

File: akka-cluster/src/main/scala/akka/cluster/Cluster.scala [hunk 4]
```diff
@@ -324,6 +324,18 @@ class Cluster(val system: ExtendedActorSystem) extends Extension {
...
+   * Typically used together `cluster.leave(cluster.selfAddress)` and then `system.shutdown()`.
+   */
+  def registerOnMemberRemoved[T](code: ⇒ T): Unit =
+    registerOnMemberRemoved(new Runnable { override def run(): Unit = code })
+
+  /**
+   * Java API: The supplied thunk will be run, once, when current cluster member is `Removed`.
+   * Typically used together `cluster.leave(cluster.selfAddress)` and then `system.shutdown()`.
+   */
+  def registerOnMemberRemoved(callback: Runnable): Unit = clusterDaemons ! InternalClusterAction.AddOnMemberRemovedListener(callback)
```
patriknw: I think there is something missing for when the cluster is shutdown

akka-cluster/src/main/scala/akka/cluster/Cluster.scala
```diff
@@ -313,17 +317,45 @@ class Cluster(val system: ExtendedActorSystem) extends Extension {
...
+
+  /**
+   * Java API: The supplied thunk will be run, once, when current cluster member is `Removed`.
+   * Typically used together `cluster.leave(cluster.selfAddress)` and then `system.shutdown()`.
+   * @throws TerminatedClusterException(String:message) Throws when the cluster have been shutdown.
+   */
+  @throws(classOf[TerminatedClusterException])
+  def registerOnMemberRemoved(callback: Runnable): Unit = {
+    if (_isTerminated.get())
+      throw new TerminatedClusterException("Cluster have been shutdown.")
```
rkuhn: Shouldn’t this just run the callback? The system must have been previously created, so running this one makes sense; running the `registerOnMemberUp` OTOH might not be correct so there the exception is the right thing to do.

rkuhn:
Thanks, looking very good, just one last thing to resolve.

akka-cluster/src/main/scala/akka/cluster/Cluster.scala
hepin1989: then need an execution context?

patriknw: I was thinking that it should just run it, in current thread.

ActorSystem.registerOnTermination will not run it, just ignored.

Cluster.subscribe will not send MemberRemoved if it is shutdown.

Perhaps we should not complicate this and simply not run it, i.e. as your previous code.

No exception please, then we need to add that in all other places to make it consistent.

rkuhn: Yes, running this one in the current thread and the other one not at all sounds like a sensible thing to do; please also document that clearly so people are not surprised. The rationale is that once the cluster has started up, user code might wish to install some cleanup handling, but the cluster might already be shutting down, and depending on the race is not healthy.

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -1114,34 +1118,43 @@ private[cluster] final class JoinSeedNodeProcess(seedNodes: immutable.IndexedSeq
  *
  * The supplied callback will be run, once, when current cluster member is `Up`.
  */
-private[cluster] class OnMemberUpListener(callback: Runnable) extends Actor with ActorLogging {
+private[cluster] class OnMemberStatusChangedListener(callback: Runnable, prefer: MemberStatus) extends Actor with ActorLogging {
```
drewhk: The parameter name "prefer" is misleading. It does not simply prefer, it actually listens to that event only.

drewhk:
LGTM apart from nitpick

akka-cluster/src/main/scala/akka/cluster/Cluster.scala
```diff
@@ -263,9 +266,14 @@ class Cluster(val system: ExtendedActorSystem) extends Extension {
    * An actor system can only join a cluster once. Additional attempts will be ignored.
    * When it has successfully joined it must be restarted to be able to join another
    * cluster or to join the same cluster again.
+   * @throws TerminatedClusterException(message:String) Throws when the cluster have been shutdown.
    */
-  def join(address: Address): Unit =
-    clusterCore ! ClusterUserAction.JoinTo(address)
+  def join(address: Address): Unit = {
+    if (_isTerminated.get())
+      throw new TerminatedClusterException("Cluster have been shutdown,you must restart the actor system to rejoin the cluster.")
```
patriknw: I'm completely convinced that we should add this. Those commands are not guaranteed to succeed in any case and they are intended to be idempotent for safe retry.

akka-cluster/src/main/scala/akka/cluster/Cluster.scala
```diff
@@ -313,17 +326,44 @@ class Cluster(val system: ExtendedActorSystem) extends Extension {
...
+    registerOnMemberRemoved(new Runnable { override def run(): Unit = code })
+
+  /**
+   * Java API: The supplied thunk will be run, once, when current cluster member is `Removed`.
+   * and if the cluster have been shutdown,that thunk will run on the caller thread immediately.
+   * Typically used together `cluster.leave(cluster.selfAddress)` and then `system.shutdown()`.
+   */
+  def registerOnMemberRemoved(callback: Runnable): Unit = {
+    if (_isTerminated.get())
+      callback.run()
```
patriknw: I think this is convenient

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -165,7 +167,9 @@ private[cluster] final class ClusterDaemon(settings: ClusterSettings) extends Ac
   def receive = {
     case msg: GetClusterCoreRef.type ⇒ coreSupervisor forward msg
     case AddOnMemberUpListener(code) ⇒
-      context.actorOf(Props(classOf[OnMemberUpListener], code).withDeploy(Deploy.local))
+      context.actorOf(Props(classOf[OnMemberStatusChangedListener], code, Up).withDeploy(Deploy.local))
+    case AddOnMemberRemovedListener(code) ⇒
+      context.actorOf(Props(classOf[OnMemberStatusChangedListener], code, Removed).withDeploy(Deploy.local))
```
patriknw: Perhaps we should be more careful with what thread we are using for running the `code`. E.g. if the code is blocking, e.g. waiting for system termination => deadlock.

patriknw: this comment should have been located below...

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -1112,36 +1116,45 @@ private[cluster] final class JoinSeedNodeProcess(seedNodes: immutable.IndexedSeq
...
-    case MemberUp(m) ⇒
-      if (isSelfUp(m))
+    case MemberRemoved(member, _) ⇒
+      if (isTriggered(member))
         done()
   }
 
-  def done(): Unit = {
+  private def done(): Unit = {
     try callback.run() catch {
```
patriknw: ...here. Perhaps we should be more careful with what thread we are using for running the code. E.g. if the code is blocking, e.g. waiting for system termination => deadlock.

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -1112,36 +1116,45 @@ private[cluster] final class JoinSeedNodeProcess(seedNodes: immutable.IndexedSeq
...
+      classOf[MemberUp]
+    case Removed ⇒
+      classOf[MemberRemoved]
+  }
+
   override def preStart(): Unit =
-    cluster.subscribe(self, classOf[MemberUp])
+    cluster.subscribe(self, to)
   override def postStop(): Unit =
     cluster.unsubscribe(self)
```
patriknw: To make sure that the `callback` is always run when shutting down we should invoke it in `postStop` if that is not already done earlier.

patriknw:
This is trickier that it seems like, but we'll get there.
Thanks, @hepin1989

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
hepin1989: I think the code should not be run in the actor or even the actor system's dispatcher if we want to get around of this ,or using a deadlock detector when the system is shut downing.or we could add a note to the the doc about the```DO NOT BLOCK```.

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
hepin1989: will address after the comment above addressed.

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
patriknw: I was thinking that a common thing to do would be `system.shutdown(); system.awaitTermination()`, but thinking some more the `awaitTermination` in the callback does not make any sense. Also, we don't spawn execution of ActorSystem `registerTermination` callbacks. Please go with 3, i.e. leave it as is here, and add a doc note.

akka-cluster/src/main/scala/akka/cluster/Cluster.scala
drewhk: I agree (I parsed your comment as "I'm **not** completely convinced that we should add this")

hepin1989: could we rejoin the cluster after the cluster have been terminated?

patriknw: wow, those small words are important. I mean I'm *not* completely....

patriknw: > could we rejoin the cluster after the cluster have been terminated?

that is not supported, you have to restart the actor system

hepin1989: so that's why I throw.Anyway,I will remove the exception If you think that's better😋

patriknw: thanks

akka-cluster/src/main/scala/akka/cluster/Cluster.scala
patriknw: I don't know, I guess this comes from ActorSystem.registerOnTermination.
Let's leave it as is.

akka-docs/rst/java/cluster-usage.rst
```diff
@@ -108,8 +108,9 @@ that the last join request is retried. Retries can be disabled by setting the pr
 An actor system can only join a cluster once. Additional attempts will be ignored.
 When it has successfully joined it must be restarted to be able to join another
 cluster or to join the same cluster again. It can use the same host name and port
-after the restart, but it must have been removed from the cluster before the join
-request is accepted.
+after the restart, when it come up as new incarnation of existing member in the cluster,
+trying to join in ,then the existing one will be removed from the cluster and then it will be allowed to join.
```
patriknw: thanks

akka-cluster/src/main/scala/akka/cluster/Cluster.scala
rkuhn: There were some cases where type inference did not work with `Unit`, while this signature says “give me whatever”.

patriknw:
LGTM

patriknw:
this has already been reviewed in another PR for some time, and I think it looks good now, so merging...

patriknw:
thanks a lot for the contribution @hepin1989

akka-cluster/src/main/scala/akka/cluster/Cluster.scala
hepin1989: thanks.

hseeberger: `code: => Unit` is obviously wrong if you want to except anything. There's a type for "anything" in Scala which is called `Any`. That would definitely work and you could get rid of the type parameter.

hepin1989: thanks very much for the detail of the type system.

Issue #17064, opened 2015-03-23
HTTP Client doesn't support "//" at the beginning or a URL path


sirthias:
Akka HTTP equivalent of this ticket: https://github.com/spray/spray/issues/1019
All that's required is a forward-port of this PR: https://github.com/spray/spray/pull/1026

/cc @jrudolph

wjur:
If this is only a forward-port and none of you is working on it then I can do it.

sirthias:
Excellent! Thanks @wjur!

Issue #16852, opened 2015-02-11
Add automatic retry of idempotent requests


jrudolph:
Spray contains retry logic that automatically retries idempotent requests (like GET requests) a configurable number of times before failing.

This could also include a feature suggested as spray/spray#747 where multiple IP addresses are tried if the hostname resolves to more than one.

This is part of the bigger initiative to support a high-level HTTP client interface as tracked as #16856.

/cc @sirthias

Issue #15906, opened 2014-09-17
Add request-level client-side API


rkuhn:
depends upon #15681

sirthias:
This is part of the bigger initiative to support a high-level HTTP client interface as tracked as #16856.

PR #17222, opened 2015-04-18
Support for client-side HTTP connection pooling

sirthias:
Supersedes https://github.com/akka/akka/pull/17036 and closes these 4 tickets:
- #15681
- #15906
- #16852
- #17064

/cc @jrudolph

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -107,6 +107,42 @@ akka.http {
     parsing = ${akka.http.parsing}
   }
 
+  host-connection-pool {
+    # The maximum number of parallel connections that a connection pool to a
+    # single host endpoint is allowed to establish. Must be greater than zero.
+    max-connections = 4
+
+    # The maximum number of times failed requests are attempted again,
+    # (if the request can be safely retried) before giving up and returning an error.
+    max-retries = 5
+
+    # The maximum number of open requests accepted into the pool across all
+    # materializations of any of its client flows.
+    # Protects against (accidentally) overloading a single pool with too many client flow materializations.
+    # Note that with N concurrent materializations the max number of open request in the pool
+    # will never exceed N * max-connections * pipelining-limit.
+    # Must be a power of 2 and > 0!
+    max-open-requests = 32
+
+    # The maximum number of requests that are dispatched to the target host in
+    # batch-mode across a single connection (HTTP pipelining).
+    # A setting of 1 disables HTTP pipelining, since only one request per
+    # connection can be "in flight" at any time.
+    # Set to higher values to enable HTTP pipelining.
+    # This value must be > 0.
+    # (Note that, independently of this setting, pipelining will never be done
+    # on a connection that still has a non-idempotent request in flight.
+    # See http://tools.ietf.org/html/rfc7230#section-6.3.2 for more info.)
+    pipelining-limit = 1
+
+    # The time after which an idle connection pool (without pending requests)
+    # will automatically terminate itself. Set to `infinite` to completely disable idle timeouts.
+    idle-timeout = 30 s
+
+    # Modify to tweak client settings for host connection pools only.
+    client = ${akka.http.client}
+  }
+
   # The (default) configuration of the HTTP message parser for the server and the client.
   # IMPORTANT: These settings (i.e. children of `akka.http.parsing`) can't be directly
   # overridden in `application.conf` to change the parser settings for client and server
--- a/akka-http-core/src/main/scala/akka/http/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/Http.scala
@@ -5,9 +5,13 @@
 package akka.http
 
 import java.net.InetSocketAddress
+import java.util.concurrent.ConcurrentHashMap
+import akka.http.util.FastFuture
 import com.typesafe.config.Config
+import scala.util.control.NonFatal
+import scala.util.Try
 import scala.collection.immutable
-import scala.concurrent.Future
+import scala.concurrent.{ ExecutionContext, Promise, Future }
 import akka.event.LoggingAdapter
 import akka.util.ByteString
 import akka.io.Inet
@@ -15,7 +19,7 @@ import akka.stream.FlowMaterializer
 import akka.stream.scaladsl._
 import akka.http.engine.client._
 import akka.http.engine.server._
-import akka.http.model.{ HttpResponse, HttpRequest }
+import akka.http.model._
 import akka.actor._
 
 class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.Extension {
@@ -33,21 +37,17 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
    */
   def bind(interface: String, port: Int = 80, backlog: Int = 100,
            options: immutable.Traversable[Inet.SocketOption] = Nil,
-           settings: Option[ServerSettings] = None,
+           settings: ServerSettings = ServerSettings(system),
            log: LoggingAdapter = system.log)(implicit fm: FlowMaterializer): Source[IncomingConnection, Future[ServerBinding]] = {
     val endpoint = new InetSocketAddress(interface, port)
-    val effectiveSettings = ServerSettings(settings)
-
     val connections: Source[StreamTcp.IncomingConnection, Future[StreamTcp.ServerBinding]] =
-      StreamTcp().bind(endpoint, backlog, options, effectiveSettings.timeouts.idleTimeout)
-
+      StreamTcp().bind(endpoint, backlog, options, settings.timeouts.idleTimeout)
     connections.map {
       case StreamTcp.IncomingConnection(localAddress, remoteAddress, flow) ⇒
-        val layer = serverLayer(effectiveSettings, log)
+        val layer = serverLayer(settings, log)
         IncomingConnection(localAddress, remoteAddress, layer join flow)
-    }.mapMaterialized { tcpBindingFuture ⇒
-      import system.dispatcher
-      tcpBindingFuture.map { tcpBinding ⇒ ServerBinding(tcpBinding.localAddress)(() ⇒ tcpBinding.unbind()) }
+    }.mapMaterialized {
+      _.map(tcpBinding ⇒ ServerBinding(tcpBinding.localAddress)(() ⇒ tcpBinding.unbind()))(fm.executionContext)
     }
   }
 
@@ -62,7 +62,7 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
   def bindAndHandle(handler: Flow[HttpRequest, HttpResponse, _],
                     interface: String, port: Int = 80, backlog: Int = 100,
                     options: immutable.Traversable[Inet.SocketOption] = Nil,
-                    settings: Option[ServerSettings] = None,
+                    settings: ServerSettings = ServerSettings(system),
                     log: LoggingAdapter = system.log)(implicit fm: FlowMaterializer): Future[ServerBinding] =
     bind(interface, port, backlog, options, settings, log).to {
       Sink.foreach { _.flow.join(handler).run() }
@@ -79,7 +79,7 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
   def bindAndHandleSync(handler: HttpRequest ⇒ HttpResponse,
                         interface: String, port: Int = 80, backlog: Int = 100,
                         options: immutable.Traversable[Inet.SocketOption] = Nil,
-                        settings: Option[ServerSettings] = None,
+                        settings: ServerSettings = ServerSettings(system),
                         log: LoggingAdapter = system.log)(implicit fm: FlowMaterializer): Future[ServerBinding] =
     bindAndHandle(Flow[HttpRequest].map(handler), interface, port, backlog, options, settings, log)
 
@@ -94,7 +94,7 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
   def bindAndHandleAsync(handler: HttpRequest ⇒ Future[HttpResponse],
                          interface: String, port: Int = 80, backlog: Int = 100,
                          options: immutable.Traversable[Inet.SocketOption] = Nil,
-                         settings: Option[ServerSettings] = None,
+                         settings: ServerSettings = ServerSettings(system),
                          log: LoggingAdapter = system.log)(implicit fm: FlowMaterializer): Future[ServerBinding] =
     bindAndHandle(Flow[HttpRequest].mapAsync(1, handler), interface, port, backlog, options, settings, log)
 
@@ -131,14 +131,13 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
   def outgoingConnection(host: String, port: Int = 80,
                          localAddress: Option[InetSocketAddress] = None,
                          options: immutable.Traversable[Inet.SocketOption] = Nil,
-                         settings: Option[ClientConnectionSettings] = None,
+                         settings: ClientConnectionSettings = ClientConnectionSettings(system),
                          log: LoggingAdapter = system.log): Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]] = {
-    val effectiveSettings = ClientConnectionSettings(settings)
     val remoteAddr = new InetSocketAddress(host, port)
-    val layer = clientLayer(remoteAddr, effectiveSettings, log)
+    val layer = clientLayer(remoteAddr, settings, log)
 
     val transportFlow = StreamTcp().outgoingConnection(remoteAddr, localAddress,
-      options, effectiveSettings.connectingTimeout, effectiveSettings.idleTimeout)
+      options, settings.connectingTimeout, settings.idleTimeout)
 
     layer.joinMat(transportFlow) { (_, tcpConnFuture) ⇒
       import system.dispatcher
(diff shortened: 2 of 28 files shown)


akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -173,6 +174,217 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
+   * Fires a single [[HttpRequest]] across the (potentially cached) host connection pool for the request's
+   * effective URI to produce a response future.
+   *
+   * Note that the request must have either an absolute URI or a valid `Host` header, otherwise
+   * the future will be completed with an error.
+   */
+  def singleRequest(request: HttpRequest,
+                    options: immutable.Traversable[Inet.SocketOption] = Nil,
+                    settings: HostConnectionPoolSettings = HostConnectionPoolSettings(system),
+                    defaultHeaders: immutable.Seq[HttpHeader] = Nil,
```
2beaucoup: I still think this can go. ;)

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -173,6 +174,217 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
+      case null ⇒ // we are the first to attempt starting a pool for this setup
+        try {
+          val gateway = ConnectionPool.startHostConnectionPool(setup)
+          promise.success(gateway)
+          gateway.whenShuttingDown.onComplete(_ ⇒ hostPoolCache.remove(setup))(system.dispatcher)
+          promise.future
+        } catch {
+          case NonFatal(e) ⇒
+            promise.failure(e)
+            hostPoolCache.remove(setup) // remove Cowboy entry if we weren't able to properly start the pool
```
2beaucoup: Is this comment still valid?

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: Yes, absolutely. No need for this here.
I'll take it out. Thanks for the reminder!

sirthias: Yes, absolutely. No need for this here.
I'll take it out. Thanks for the reminder!

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: Nope. No more cowboys here. This is a civilized place now.
I'll remove the signpost to the former wild-west territory.

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: Actually, looking at it a bit more closely, there is a reason for having this here.
Not such much to supply additional headers to the request (which the user could easily do on her side before the call), but rather because the `defaultHeaders` are part of the `ConnectionPoolSetup` and as such the cache key.
If we don't give the user the ability to supply default headers here they cannot target a host pool that was configured with `defaultHeaders` before.

Plus, the additional parameter doesn't really hurt anyone here, so I'd be inclined to keep it after all.

2beaucoup: So a user would have to take care to provide the same `defaultHeaders` that was used in previous setups to get cached connections?

sirthias: Yes. The cache maintains separate pools per ActorSystem, target endpoint and pool configuration. The default headers are part of the configuration.

So, if you want your `singleRequest` call to go out over a pool that was previously configured with a certain `defaultHeaders` settings then you'll have to supply the same setting here. Otherwise a new pool will be spun up.

2beaucoup: I think it would be more intuitive to remove the `defaultHeaders` from the settings/cache key and wrap the flow with default header apllication instead. That way setups with different `deafultHeaders` would be able to share connection instances. WDYT?

akka-http-core/src/main/resources/reference.conf
```diff
@@ -107,6 +107,50 @@ akka.http {
...
+    # The time after which an idle connection pool (without open connections)
+    # will automatically terminate itself. Set to `infinite` to disable idle timeouts.
+    idle-timeout = 30 s
+
+    # Grace period between beginning a connection pool shutdown and the (internal)
+    # gateway actor not accepting any more requests into the pool. Protects racy
+    # requests that were dispatched to the pool after the Shutdown message was
+    # queued.
+    # Set to zero to completely skip the grace-period (and accept the risk that
+    # requests sent around the time of an idle shutdown might get lost).
```
rkuhn: Not entirely clear: will it Just Work if I first send requests and then shut down the pool? Or will shutting down the pool kill requests that were previously enqueued? I have not yet read the internals, but from a user’s perspective it should Just Work and this grace period should be unnecessary.

akka-http-core/src/main/scala/akka/http/Http.scala
rkuhn: I’m leaning towards @2beaucoup here: the current behavior is more surprising to users.

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
```diff
@@ -0,0 +1,141 @@
...
+                                               +-------------------+                             
+                                               |                   |                             
+                                        +----> | Connection Slot 1 +---->                        
+                                        |      |                   |    |                        
+                                        |      +---+---------------+    |                        
+                                        |          |                    |                        
+                       +-----------+    |      +-------------------+    |      +---------------+
+      RequestContext   |           +----+      |                   |    +----> |               |  ResponseContext
+    +----------------> | Conductor |---------> | Connection Slot 2 +---------> | responseMerge +------------------>
+                       |           +----+      |                   |    +----> |               |
```
rkuhn: there seems to be one space missing on this line before the first `|` ;-)

rkuhn: (or the github formatter screws this up—the view definitely does, but I copied it into a terminal window)

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: Yes, we could of course move out the defaultHeader application.
I like it.
Will do.

akka-http-core/src/test/scala/akka/http/engine/client/ConnectionPoolSpec.scala
```diff
@@ -0,0 +1,358 @@
...
+      val (Success(response), 42) = responseOut.expectNext()
+      response.headers should contain(RawHeader("Req-Host", s"localhost:$serverPort"))
+    }
+
+    "add default headers to requests if they don't contain them" in new TestSetup {
+      val (requestIn, responseOut, responseOutSub, hcp) =
+        cachedHostConnectionPool[Int](defaultHeaders = List(RawHeader("X-Custom-Header", "Default")))
+
+      val requestInSub = requestIn.expectSubscription()
+      requestInSub.expectRequest(4)
```
rkuhn: this will break when some internals are changed, perhaps just `expectRequest()`?

akka-http-core/src/test/scala/akka/http/engine/client/ConnectionPoolSpec.scala
```diff
@@ -0,0 +1,358 @@
...
+    "open a second connection if the first one is loaded" in new TestSetup {
+      val (requestIn, responseOut, responseOutSub, hcp) = cachedHostConnectionPool[Int]()
+
+      val requestInSub = requestIn.expectSubscription()
+      requestInSub.expectRequest(4)
+      requestInSub.sendNext(HttpRequest(uri = "/a") -> 42)
+      requestInSub.sendNext(HttpRequest(uri = "/b") -> 43)
+
+      responseOutSub.request(2)
+      acceptIncomingConnection()
```
rkuhn: so it also opens a second connection while the first is being established? otherwise it would be more reliable to cork the server’s processing until the second connection is made

akka-http-core/src/test/scala/akka/http/engine/client/ConnectionPoolSpec.scala
```diff
@@ -0,0 +1,358 @@
...
+          id
+        case x ⇒ fail(x.toString)
+      }.runFold(0)(_ + _)
+
+      acceptIncomingConnection()
+      acceptIncomingConnection()
+      acceptIncomingConnection()
+      acceptIncomingConnection()
+
+      Await.result(idSum, 10.seconds) shouldEqual N * (N + 1) / 2
```
rkuhn: nice!

akka-http-core/src/test/scala/akka/http/engine/client/ConnectionPoolSpec.scala
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    "transparently restart after idle shutdown" in new TestSetup() {
+      val (requestIn, responseOut, responseOutSub, hcp) =
+        cachedHostConnectionPool[Int](idleTimeout = 1.second, shutdownGracePeriod = 1.second)
+
+      val gateway = Await.result(hcp.gateway, 500.millis)
+      gateway.whenShuttingDown.isCompleted shouldEqual false
+      val whenShutdownCompleted = Await.result(gateway.whenShuttingDown, 1500.millis) // verify shutdown start (after idle)
+      Await.result(whenShutdownCompleted, 1500.millis) // verify shutdown completed (after grace period)
```
rkuhn: What happens if you leave out this step? Will it just not shut down if re-triggered during the grace period?

akka-http-core/src/test/scala/akka/http/engine/client/ConnectionPoolSpec.scala
```diff
@@ -0,0 +1,358 @@
...
+    def value = nr.toString
+  }
+
+  implicit class MustContain[T](specimen: Seq[T]) {
+    def mustContainLike(pf: PartialFunction[T, Unit]): Unit =
+      specimen.collectFirst(pf) getOrElse fail("did not contain")
+  }
+
+  object NoErrorComplete extends SingletonException
+}
```
rkuhn: Tests are looking very good!

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
```diff
@@ -0,0 +1,239 @@
...
+    final case class RetryRequest(rc: RequestContext) extends SlotEvent
+  }
+
+  case class Ports(
+    requestContextIn: Inlet[RequestContext],
+    responseOut: Outlet[ResponseContext],
+    slotEventOut: Outlet[SlotEvent]) extends Shape {
+
+    override val inlets = requestContextIn :: Nil
+    override def outlets = responseOut :: slotEventOut :: Nil
```
rkuhn: `val` vs `def`?

akka-http-core/src/main/resources/reference.conf
sirthias: > will it Just Work if I first send requests and then shut down the pool?

Yes, this is not a problem. The grace period is needed because we automatically shut down the pool after an idle-timeout. If I sent a request shortly after the idle-shutdown command has been queued this request would be lost without a grace period.

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
sirthias: Hmmm... looks good for me here on github as well as in my IDEA editor window.

akka-http-core/src/test/scala/akka/http/engine/client/ConnectionPoolSpec.scala
sirthias: If I only plan on sending one element then, yes, `expectRequest()` is better.
But what do I do if I plan on sending more element (e.g. 4)?

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
```diff
@@ -0,0 +1,239 @@
...
+      new Inlet(requestContextIn.toString),
+      new Outlet(responseOut.toString),
+      new Outlet(slotEventOut.toString))
+
+    override def copyFromPorts(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]): Shape =
+      Ports(
+        inlets.head.asInstanceOf[Inlet[RequestContext]],
+        outlets.head.asInstanceOf[Outlet[ResponseContext]],
+        outlets.last.asInstanceOf[Outlet[SlotEvent]])
+  }
```
rkuhn: How about using the `FanOutShape` infrastructure? As an example consider
~~~scala
class FanOutShape2[I, O0, O1](_init: FanOutShape.Init[I]) extends FanOutShape[I](_init) {
  def this(name: String) = this(FanOutShape.Name[I](name))
  override protected def construct(init: FanOutShape.Init[I]): FanOutShape[I] = new FanOutShape2(init)
  
  val out0 = newOutlet[O0]("out0")
  val out1 = newOutlet[O1]("out1")
}
~~~

akka-http-core/src/test/scala/akka/http/engine/client/ConnectionPoolSpec.scala
sirthias: > so it also opens a second connection while the first is being established?

Yes. After one request has been scheduled across an "unconnected" connection this connection is then considered either "busy" or "loaded" (depending on whether the request method is idempotent or not, respectively).

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
```diff
@@ -0,0 +1,239 @@
...
+
+  private class SlotProcessor(slotIx: Int, connectionFlow: Flow[HttpRequest, HttpResponse, Any],
+                              settings: HostConnectionPoolSettings)(implicit fm: FlowMaterializer)
+    extends ActorSubscriber with ActorPublisher[List[ProcessorOut]] with ActorLogging {
+
+    var exposedPublisher: akka.stream.impl.ActorPublisher[Any] = _
+    var inflightRequests = immutable.Queue.empty[RequestContext]
+    val runnableFlow = Source.actorPublisher[HttpRequest](Props(new FlowInportActor(self)))
+      .via(connectionFlow)
+      .toMat(Sink.actorSubscriber[HttpResponse](Props(new FlowOutportActor(self))))(_ -> _)
```
rkuhn: there is a combiner named `Keep.both`

akka-http-core/src/test/scala/akka/http/engine/client/ConnectionPoolSpec.scala
sirthias: This step only confirms that the shutdown is indeed completed eventually. We could also leave this out or move it to the very end. Once a shutdown has been started it will always be completed after the grace period.
Any request scheduled onto the same endpoint & pool config will end up starting a new pool instance.
The grace period only protects against not loosing racy requests.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
sirthias: 'val' it shall be.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
sirthias: Yes, good point. I'll remove the `Ports` here altogether and switch to `FanOutShape2`.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
sirthias: Yes, right. Thanks!

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
```diff
@@ -0,0 +1,239 @@
...
+
+    def waitingForSubscribePending: Receive = {
+      case SubscribePending ⇒
+        exposedPublisher.takePendingSubscribers() foreach (s ⇒ self ! ActorPublisher.Internal.Subscribe(s))
+        context.become(unconnected)
+    }
+
+    val unconnected: Receive = {
+      case OnNext(rc: RequestContext) ⇒
+        assert(totalDemand > 0)
```
rkuhn: This is not a good idea (and not proper RS behavior); I have created #17244 to solve this at the ActorSubscriber level.

2beaucoup:
I'm a little confused by `ConnectionPool.Gateway`, `HostConnectionPoolGateway` and `HostConnectionPoolGateway.PoolGateway`. Which does what? Some comments (or alternative names?) would be nice. :)

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
sirthias: Why is it not a good idea? It simply signals a potential bug in the `SlotProcessor` code here.
The `OnNext` comes in via the `ActorSubscriber` upstream interface while the `totalDemand` shows downstream demand (`ActorPublisher` side). What I assert against here is that we have somehow requested from upstream without having seen demand from downstream first.

sirthias:
@2beaucoup Well, the naming is crystal clear, isn't it? ;-)

No, you are right. This can be improved. I'll rename to something better.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
```diff
@@ -0,0 +1,239 @@
...
+    def handleDisconnect(error: Option[Throwable]): Unit = {
+      log.debug("Slot {} disconnected after {}", slotIx, error getOrElse "regular connection close")
+      val results: List[ProcessorOut] = inflightRequests.map { rc ⇒
+        if (rc.retriesLeft == 0) {
+          val reason = error.fold[Throwable](new RuntimeException("Unexpected disconnect"))(identityFunc)
+          ResponseDelivery(ResponseContext(rc, Failure(reason)))
+        } else SlotEvent.RetryRequest(rc.copy(retriesLeft = rc.retriesLeft - 1))
+      }(collection.breakOut)
+      inflightRequests = immutable.Queue.empty
+      onNext(SlotEvent.Disconnected(slotIx, results.size) :: results)
```
rkuhn: There might be a case where this is against the rules: imagine there not being any requests in flight with at the same time no more downstream demand and then the TCP connection closes for some weird reason. I think the simple fix is to not send empty collections.

2beaucoup:
@sirthias Klar wie Kloßbrühe! ;)

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
rkuhn: Nope, this can also be the result of a RS spec violation of the upstream Publisher, in which case shutting down the ActorSystem without even printing a log seems a little harsh ;-) Seriously, using `assert` in Actors is (currently) not ideal.

akka-http-core/src/test/scala/akka/http/engine/client/ConnectionPoolSpec.scala
rkuhn: remember what was requested and expect more if needed—there is also AutoPublisher

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSettings.scala
```diff
@@ -0,0 +1,53 @@
...
+import com.typesafe.config.Config
+import scala.collection.immutable
+import scala.concurrent.duration.{ FiniteDuration, Duration }
+import akka.actor.ActorRefFactory
+import akka.http.util._
+
+final case class HostConnectionPoolSetup(host: String, port: Int, setup: ConnectionPoolSetup)
+
+final case class ConnectionPoolSetup(
+  encrypted: Boolean,
```
rkuhn: This will then probably be replaced with `Option[NegotiateNewSession]`, right?

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
sirthias: Ok, I'll simply remove the assert then.

akka-http-core/src/test/scala/akka/http/engine/client/ConnectionPoolSpec.scala
sirthias: Ok, thanks.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSettings.scala
sirthias: I'm not quite sure yet of how exactly things will look. But something along these lines, yes.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolGateway.scala
```diff
@@ -0,0 +1,139 @@
...
+    // start the pool flow with this actor acting as source as well as sink
+    Source(ActorPublisher(self)).via(poolFlow).runWith(Sink(ActorSubscriber[ResponseContext](self)))
+
+    activateIdleTimeoutIfNecessary()
+
+    def requestStrategy = ZeroRequestStrategy
+
+    def receive: Receive = {
+
+      /////////////// FROM POOL DOWNSTREAM //////////////
```
rkuhn: this is a bit confusing: isn’t this “from the pool’s upstream end”?

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolGateway.scala
```diff
@@ -0,0 +1,139 @@
...
+
+      case Request(_) ⇒ dispatchRequests() // the pool is ready to take on more requests
+
+      case Cancel ⇒
+        // somehow the pool shut down, however, we don't do anything here because we'll also see an
+        // OnComplete or OnError which we use as the sole trigger for cleaning up
+        ()
```
rkuhn: this line should be removed

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolGateway.scala
```diff
@@ -0,0 +1,139 @@
...
+private object HostConnectionPoolGateway {
+  import ConnectionPool.{ PoolRequest, BeginShutdown }
+
+  case class RequestContext(request: HttpRequest, responsePromise: Promise[HttpResponse], retriesLeft: Int) {
+    require(retriesLeft >= 0)
+  }
+  case class ResponseContext(rc: RequestContext, response: Try[HttpResponse])
+
+  private case object ShutdownNow
+
```
rkuhn: Please add a comment that this Actor wraps a Flow, so its Publisher and Subscriber interfaces are “internal”.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolGateway.scala
sirthias: Well, these things are "coming up" from the pool. Here we are "atop" the pool and the pool is on our (internal) downstream side. I'll improve the comment.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolGateway.scala
sirthias: It's here to keep scalariform happy. Otherwise the comment will be strangely aligned.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolGateway.scala
sirthias: Yes, will do.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolGateway.scala
```diff
@@ -0,0 +1,139 @@
...
+import scala.annotation.tailrec
+import scala.concurrent.{ Future, Promise }
+import scala.concurrent.duration.{ Duration, FiniteDuration }
+import scala.util.{ Failure, Try }
+import akka.actor.{ ActorLogging, Cancellable }
+import akka.stream.FlowMaterializer
+import akka.stream.actor.{ ActorPublisher, ActorSubscriber, ZeroRequestStrategy }
+import akka.stream.actor.ActorPublisherMessage._
+import akka.stream.actor.ActorSubscriberMessage._
+import akka.stream.impl.ResizableMultiReaderRingBuffer
```
rkuhn: This use-case does neither resize nor have multiple readers, please use `akka.stream.impl.FixedSizeBuffer`.

sirthias: Nice one. I was looking for something like that but didn't find it. Probably because I was looking for a "ring" buffer impl.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolConductor.scala
```diff
@@ -0,0 +1,221 @@
...
+            case Loaded(n) if n == failed ⇒ Unconnected
+            case Busy(n) if n > failed    ⇒ Busy(n - failed)
+            case Busy(n) if n == failed   ⇒ Unconnected
+            case _                        ⇒ throw new IllegalStateException(s"Disconnect(_, $failed) on $slotState connection?")
+          }
+
+        /**
+         * Implements the following Connection Slot selection strategy
+         *  - Select the first idle connection in the pool, if there is one.
+         *  - If none is idle select the first unconnected connection, if there is one.
```
rkuhn: it actually selects the last unconnected connection

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolConductor.scala
```diff
@@ -0,0 +1,221 @@
...
+              case _ ⇒ arg.openIdempotentRequests < pipeliningLimit
+            }
+            slotStates(ix) match {
+              case Idle                             ⇒ ix
+              case Unconnected                      ⇒ bestSlot(ix + 1, ix, Unconnected)
+              case x: Loaded if isBetterThanBest(x) ⇒ bestSlot(ix + 1, ix, x)
+              case _                                ⇒ bestSlot(ix + 1, bestIx, bestState)
+            }
+          } else bestIx
+      }
```
rkuhn: Nice one, I like it!

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
rkuhn: weird

akka-http-core/src/main/scala/akka/http/engine/client/ConnectionPool.scala
```diff
@@ -0,0 +1,141 @@
...
+                          remoteAddress: InetSocketAddress, settings: HostConnectionPoolSettings, log: LoggingAdapter)(
+                            implicit system: ActorSystem, fm: FlowMaterializer): Flow[RequestContext, ResponseContext, Any] =
+    Flow() { implicit b ⇒
+      import settings._
+      import FlowGraph.Implicits._
+
+      val conductor = b.add(HostConnectionPoolConductor(maxConnections, maxRetries, pipeliningLimit, log))
+      val slots = Vector
+        .tabulate(maxConnections)(HostConnectionPoolSlot(_, connectionFlow, remoteAddress, settings))
+        .map(b.add(_))
```
rkuhn: did I ever say that I totally love Scala?

sirthias: I believe not.
But it's certainly good to take a moment in quiet content,
reflect on the beauty of what was given to us,
and praise the creator.

And then jump right back into the flame war surrounding `CanBuildFrom`.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolSlot.scala
sirthias: Right, although I'm not what you mean by not sending empty collections.
I'd say the proper fix is to turn the `SlotEventSplit` into a route that always maintains demand for `SlotEvent`s. The conductor will always demand `SlotEvent`s, so if we repeat the "prepend with doubler-stage" trick here we should be able to always pull out `SlotEvent`s (if no undemanded response is blocking the way, which shouldn't happen).

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -173,6 +174,217 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
+
+  private def clientFlow[T](settings: HostConnectionPoolSettings)(
+    poolGateway: (HttpRequest, Any) ⇒ Future[ConnectionPool.Gateway])(
+      implicit system: ActorSystem, fm: FlowMaterializer): Flow[(HttpRequest, T), (Try[HttpResponse], T), Any] = {
+    // a connection pool can never have more than pipeliningLimit * maxConnections requests in flight at any point
+    val parallelism = settings.pipeliningLimit * settings.maxConnections
+    Flow[(HttpRequest, T)].mapAsyncUnordered(parallelism, {
+      case (request, userContext) ⇒
+        val responsePromise = Promise[HttpResponse]()
+        poolGateway(request, userContext).onComplete(dispatch(request, responsePromise))(system.dispatcher)
```
rkuhn: should this use `fm.executionContext` instead?

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -173,6 +174,217 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
+   * configuration a separate connection pool is maintained.
+   * The HTTP layer transparently manages idle shutdown and restarting of connections pools as configured.
+   * The returned [[Flow]] instances therefore remain valid throughout the lifetime of the application.
+   *
+   * Note that the provided caching logic follows a "best-effort" design with regard to pool uniqueness.
+   * It cannot fully guarantee that the application will always run at most one pool for a given host and
+   * configuration.
+   * The cache removes entries for pools that are shut down (e.g. because of an idle timeout) and due to the
+   * inherent raciness of this operation one client might end up re-starting an old pool while another
+   * client starts a new pool if both access the cache around the time of the shutdown.
```
rkuhn: It might be preferable (semantically) to never remove entries from the `hostPoolCache` in order to prevent double creation—abiding by API limits is a serious business. What is the downside?

akka-http-core/src/main/scala/akka/http/Http.scala
sirthias: I'd say yes. Maybe we should switch to that dispatcher everywhere, since `fm.executionContext` is easier to tweak (if required) than `system.dispatcher`.

akka-http-core/src/main/scala/akka/http/Http.scala
```diff
@@ -173,6 +174,217 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
...
+      case e: IllegalUriException ⇒ FastFuture.failed(e)
+    }
+
+  /**
+   * Triggers an orderly shutdown of all host connections pools currently maintained by the [[ActorSystem]].
+   * The returned future is completed when all pools that were live at the time of this method call
+   * have completed their shutdown process.
+   *
+   * If existing pool client flows are re-used or new ones materialized concurrently with or after this
+   * method call the respective connection pools will be restarted and not contribute to the returned future.
```
rkuhn: This can remove entries from `hostPoolCache` if that feature is removed above.

akka-http-core/src/main/scala/akka/http/Http.scala
rkuhn: yes

rkuhn:
Looking very good!

akka-http-core/src/main/scala/akka/http/Http.scala
jrudolph: Isn't the problem that the cache is unbounded in size? E.g. if you write a web-crawler it would start to hoard pools for lots of hosts even if it never crawls back.

sirthias: I feel somewhat uneasy with structures that only grow (w/o bounds) and never shrink. If we never remove entries the `hostPoolCache` would be such a case.

For scenarios like the one @jrudolph described an automatic idle-timeout, which doesn't require any manual cleaning, can be quite helpful.

To be honest, I think the "two concurrent pool" problem will rarely be an issue. The default idle-timeout is 30 seconds, so the chance that a client re-accesses an existing pool at the "wrong moment", after the idle-timeout has kicked in but before the cache entry has been removed, is quite low. And even then you'll have to construct a second pool client to same endpoint & config to end up with two concurrent pools.

I'd even be open to increasing to default timeout to something like 2 min, to reduce the risk even more.

akka-http-core/src/test/scala/akka/http/engine/client/ConnectionPoolSpec.scala
drewhk: +1 for AutoPublisher.

akka-http-core/src/main/scala/akka/http/Http.scala
rkuhn: Reducing the risk is not what I’m talking about: if there is a race then there is a race, which means that instead of the simple “there will be only one” we need to document that under rare circumstances the number per target is potentially unbounded—not nice. If we cannot make the idle timeout work reliably then we will have to place the burden on the user, offering an API that shuts down a pool so that it cannot come back. Then we can add a check to the read side of the CHM that verifies that the entry it got is still valid and retries if not.

jrudolph: Is it really unbounded? IIUC in the worst case the old pool will be around for another idle time but no new requests (besides those in the grace period) will be scheduled on it. So, after those requests have been processed it will be idle and die.

akka-http-core/src/main/scala/akka/http/engine/client/HostConnectionPoolConductor.scala
sirthias: Indeed. While it doesn't really matter I'll fix this (easier to follow during debugging).

akka-http-core/src/main/scala/akka/http/Http.scala
rkuhn: The worst case is:

  * pool is idle for 30sec
  * request comes in so that it reads the old pool’s reference and restarts it while the gateway is removed from the map
  * request comes in later that creates a new pool
  * repeat _ad nauseam_

The unbounded worst case assumes that the pool reactivated in step 2 is then used forever after by that client.

sirthias: Actually I think we can change the logic a bit to make the "two concurrent pools" problem go away properly:
After a pool gateway received the `BeginShutdown` trigger we simply shut down the old pool immediately and use the grace period to forward potential "late" requests to the new pool. This way the reception of `BeginShutdown` marks the clear boundary between the old pool shutting down and the new pool being started.

We can even delay the start of the new pool until the old pool has fully completed its tear down, thereby preventing new connections to the target hosts being attempted while old ones are still live.

That should solve this thing properly and still allow us to remove cache entries automatically.

jrudolph: But this assumes an unbounded number of clients that keep references and use the ever newly created pools. I don't think this situation is too bad. If you want to be sure, all clients can use the same reference from the start. You could even say that it is a feature of a cache in general that it doesn't need to be used (or at least we could define it that way here).

jrudolph: My comment was @rkuhn

rkuhn: @sirthias That sounds like a very good plan!

jrudolph: @sirthias what happens to gateways that have been delivered previously (pruned from the cache but still referenced)? They will still need to be able to forward requests to any later created actual pool instance. But that would mean that you need to decouple the things you hand out from the things you put into the cache.

akka-http-core/src/main/scala/akka/http/Http.scala
2beaucoup: Alternatively we could just add the correct host header internally and let the user do `.map(_.withDefaultHeaders(...))` *before* submitting requests to the pool. That way we can can rid of `defaultHeaders` completely.

sirthias:
Ok, applied all review feedback and, most importantly, reworked the complete pool caching/termination/restart design to achieve these goals:

- guarantee a max of one live pool per ActorSystem, target endpoint and config at any time
- keep terminated pools and their support infrastructure fully GC-able (i.e. no permanent cache entries)
- remove the need for configurable grace-periods

The latter point is achieved by switching all potentially newly incoming requests to a subsequent incarnation immediately after a pool shutdown is started. While the shutdown is in process all potentially still arriving ("late") requests also get forwarded to the subsequent incarnation.
While this approach still leaves open a theoretical chance of a request to a pool ending up as a dead letter due to extremely unfair scheduling (as any explicit grace period would as well) I'd consider it by far good enough for actual use.

akka-http-core/src/main/scala/akka/http/Http.scala
2beaucoup: @sirthias ping

sirthias:
Fixed validation error from merge problem.

@2beaucoup The `defaultHeaders` on `singleRequest` have been removed with the rework yesterday.

sirthias:
PLS BUILD

2beaucoup:
@sirthias Actually I meant to remove `defaultHeaders` from the other methods as well, only add the host header and let users use `withDefaultHeaders` themselves. See https://github.com/2beaucoup/akka/commit/b37d4b5558432f742ba7cc4ed4717051a568c05e.

sirthias:
@2beaucoup Ah, I see. I like it. Also because the API simplification would make things easier on the Java side, where we don't have default parameters. I'll simply apply your patch. Thanks!

patriknw:
PLS BUILD

patriknw:
@sirthias merge conflict

sirthias:
Updated and rebased once more (removed `defaultHeaders` param completely).

rkuhn:
PLS BUILD

2beaucoup:
@sirthias :+1:

PR #17224, opened 2015-04-19
test using a custom class as parameter directive

yanns:


ktoso:
OK TO TEST

ktoso:
Thanks for the additional test! 
I'm not entirely sure if we need to pull it in, CCing @sirthias @jrudolph to decide on this :-)

yanns:
@ktoso not sure as well about testing it.
Another possibility would be to document it.
But I don't know exactly where it can fit it. Maybe a new chapter in http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0-M5/scala/http/directives/basic-directives/?

sirthias:
Yes, Yann, I think something like this should be part of the documentation.
The best place is probably going to be the [Marshalling chapter](http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0-M5/scala/http/marshalling.html), which is, unfortunately, not completed yet.
Maybe we could postpone this addition by two weeks and you could check, whether the docs we'll hopefully have written by then do cover your questions?

yanns:
@sirthias OK with your proposition.

rkuhn:
Thanks, gents, I’ll close this then for now, it can be reopened as needed (or just used as a basis for a new PR).

yanns:
The marshalling is still in TODO.
If you want to, I can try to help here.

Issue #16699, opened 2015-01-21
Fix all the FIXMEs in streams and http


rkuhn:
(the same holds for all TODOs)

Bernd-Vogt:
I'm really looking forward to the new Http-Manager Java API.

And really really hope that the code shown in [JavaTestServer](https://github.com/akka/akka/blob/releasing-akka-stream-and-http-experimental-1.0-M3/akka-http-core/src/test/java/akka/http/model/japi/JavaTestServer.java) can be used soon.

PR #17228, opened 2015-04-20
=str #16699 fix some FIXMEs

patriknw:


akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -96,8 +97,16 @@ object Source {
    * stream will see an individual flow of elements (always starting from the
    * beginning) regardless of when they subscribed.
    */
-  def from[O](iterable: java.lang.Iterable[O]): javadsl.Source[O, Unit] =
-    new Source(scaladsl.Source(akka.stream.javadsl.japi.Util.immutableIterable(iterable)))
+  def from[O](iterable: java.lang.Iterable[O]): javadsl.Source[O, Unit] = {
+    // this adapter is not immutable if the the underlying java.lang.Iterable is modified
+    // but there is not anything we can do to prevent that from happening.
+    // ConcurrentModificationException will be thrown in some cases.
```
ktoso: Worth promoting this comment to a javadoc level comment?

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -363,7 +372,7 @@ class Source[+Out, +Mat](delegate: scaladsl.Source[Out, Mat]) extends Graph[Sour
    * @param n must be positive, and `d` must be greater than 0 seconds, otherwise [[IllegalArgumentException]] is thrown.
    */
   def groupedWithin(n: Int, d: FiniteDuration): javadsl.Source[java.util.List[Out @uncheckedVariance], Mat] =
-    new Source(delegate.groupedWithin(n, d).map(_.asJava)) // FIXME optimize to one step
+    new Source(delegate.groupedWithin(n, d).map(_.asJava)) // TODO optimize to one step
```
ktoso: New task: `DO all the TODOs` :wink:

ktoso: (this is fine, doesn't have to done this sprint I think)

ktoso:
LGTM, the changes made look OK

rkuhn:
LGTM

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
patriknw: yeah, TODO is not as urgent as FIXME (a final release should not contain any FIXMEs)

ktoso: ok, makes sense :)

akka-stream/src/main/scala/akka/stream/OperationAttributes.scala
```diff
@@ -62,9 +62,22 @@ final case class OperationAttributes private (attributes: immutable.Seq[Operatio
    * INTERNAL API
    */
   private[akka] def nameLifted: Option[String] =
-    attributes.collect {
-      case Name(name) ⇒ name
-    }.reduceOption(_ + "-" + _) // FIXME don't do a double-traversal, use a fold instead
+    if (attributes.isEmpty)
+      None
+    else {
```
drewhk: Thanks!

akka-stream/src/main/scala/akka/stream/impl/FanoutProcessor.scala
```diff
@@ -79,11 +79,9 @@ private[akka] abstract class FanoutOutputs(val maxBufferSize: Int, val initialBu
     case SubscribePending ⇒
       subscribePending()
     case RequestMore(subscription, elements) ⇒
-      // FIXME can we avoid this cast?
       moreRequested(subscription.asInstanceOf[ActorSubscriptionWithCursor[Any]], elements)
```
drewhk: can we move the cast to the pattern match?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -173,7 +173,7 @@ private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immut
   override def onPull(ctx: Context[immutable.Seq[T]]): SyncDirective =
     if (ctx.isFinishing) {
       val elem = buf.result()
-      buf.clear() //FIXME null out the reference to the `buf`?
+      buf.clear()
```
drewhk: we can null it out though.

drewhk:
Thanks, LGTM!

akka-stream/src/main/scala/akka/stream/impl/FanoutProcessor.scala
patriknw: strange, I thought I tried that, but it works, changing to that instead:

    case RequestMore(subscription: ActorSubscriptionWithCursor[Any], elements) ⇒
       moreRequested(subscription, elements)

patriknw: ah, now I remember, since it's in partial function we will not see if its another type, but I can add additional cases for that and throw something

drewhk: ok, I see. I am not sure if it is worth those extra cases

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
patriknw: why? isn't that premature optimization? if it's finishing here the whole stage is about to be garbage collected soon anyway

akka-stream/src/main/scala/akka/stream/impl/FanoutProcessor.scala
patriknw: yeah, I guess that was my thought as well. You decide, this is how the alternative looks like:

    case RequestMore(subscription: ActorSubscriptionWithCursor[Any], elements) ⇒
      moreRequested(subscription, elements)
      pump.pump()
    case Cancel(subscription: ActorSubscriptionWithCursor[Any]) ⇒
      unregisterSubscription(subscription)
      pump.pump()
    case RequestMore(subscription, _) =>
      throw new IllegalStateException(s"Expected ${classOf[ActorSubscriptionWithCursor[Any]]}, but was " +
        (if (subscription == null) null else subscription.getClass))
    case Cancel(subscription) =>
      throw new IllegalStateException(s"Expected ${classOf[ActorSubscriptionWithCursor[Any]]}, but was " +
        (if (subscription == null) null else subscription.getClass))

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: Ah true, finishing a stage sets the backing array slot to a FinishedOp. GC was not the primary intention for that feature though, but it is nice anyway :)

akka-stream/src/main/scala/akka/stream/impl/FanoutProcessor.scala
drewhk: It does not worth it, keep it as it is.

patriknw: ok

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
patriknw: good idea

Issue #15833, opened 2014-09-09
Add session renegotiation to SSL/TLS


bantonsson:
The SSL/TLS support in #15402 doesn't support session renegotiation.

rkuhn:
Preliminary sketch of the types (based on web-based research, so comments welcome):
~~~scala
sealed trait SessionRead
case class SessionNegotiated(session: SSLSession, parameters: SSLParameters) extends SessionRead
case class SessionReadData(data: ByteString) extends SessionRead

sealed trait SessionWrite
case class SessionRenegotiate(parameters: SSLParameters, invalidate: Boolean) extends SessionWrite
case class SessionWriteData(data: ByteString) extends SessionWrite

def addSSL(f: Flow[ByteString, ByteString], parameters: SSLParameters): Flow[SessionWrite, SessionRead]
def addNoSSL(f: Flow[ByteString, ByteString]) = Flow[SessionWrite].collect{case SessionWriteData(d) => d}.via(f).map(SessionReadData)
~~~
The last two will eventually be represented as standalone pieces with four ports, so the method signatures only serve to document the intent.

On the HTTP side the HttpRequestParser will have to keep track of which bytes were received with which session and reject requests that are “split” between them. The HttpRequest will include a field that holds the session that was in effect when the request was received, and the entity’s stream will need to be terminated upon renegotiation. A renegotiation will then be effected by including this request in the HttpResponse; the response itself will only be sent once the new session has been established, so that follow-up requests are delivered with the new parameters in effect.

Issue #17062, opened 2015-03-23
SslTlsCipher (probably) stops processing in certain renegotiation scenarios


sirthias:
We recently got a bug report against spray's client-side which describes a connection not making any more progress in certain scenarios when SSL/TLS encryption is on and the server wants to verify a client certificate: https://github.com/spray/spray/issues/1010

We traced it down to our SSL/TLS implementation not handling `BUFFER_UNDERFLOW` conditions in our `decrypt` phase correctly. While we didn't have time to fully reproduce the problem with a dedicated test we attempted a fix with https://github.com/spray/spray/commit/877c175d7e7dbe3820db5e0b718aa6db81d91a84 which was then reported as effective by the customer.

From a superficial glance at `SslTlsCipher.scala` the same problem appears to be present with the current Akka Stream SSL/TLS implementation.

/cc @jrudolph

PR #17217, opened 2015-04-17
=str #16986 Handle early cancelation issue

patriknw:
and =str #15191 Verify that stage actors are stopped

Some more things to verify and add, but you can take a first look and see if it is on right track. Note 2 commits if you want to review the two tickets in isolation.

Issue #16167, opened 2014-10-29
Improve server- and client-side stream-TLS APIs


sirthias:
More thinking on the basic client- and server-side APIs for TCP, HTTP and SSL made us realize that we need to improve things once more.
This is the ticket for SSL, there is also [one for HTTP](https://github.com/akka/akka/issues/16162) and [one for TCP](https://github.com/akka/akka/issues/16166).

This is our API proposal:

```scala
object SSL {
  def upgradeClientSide(connectionFlow: ConnectionFlow, ...): Tcp.ConnectionFlow
  def upgradeServerSide(connectionFlow: ConnectionFlow, ...): Tcp.ConnectionFlow
}
```

To see all proposed APIs for TCP, HTTP and SSL together see this gist:
https://gist.github.com/sirthias/88391662fb030da45083

/cc @jrudolph

PR #17231, opened 2015-04-20
+str #15833 TLS with session renegotiation

rkuhn:


rkuhn:
fixes #15833, #16167, #16106 (well, removal is pending)

rkuhn:
cc @sirthias @jrudolph

sirthias:
Probably also closes #17062, right?

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+
+    val trustManagerFactory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm)
+    trustManagerFactory.init(trustStore)
+
+    val context = SSLContext.getInstance("TLS")
+    context.init(keyManagerFactory.getKeyManagers, trustManagerFactory.getTrustManagers, new SecureRandom)
+    context
+  }
+
+  class Timeout()(implicit system: ActorSystem) extends AsyncStage[ByteString, ByteString, Unit] {
```
drewhk: This is an overall timeout, not an idle-timeout right? Maybe it deserves a comment.

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+      if (ctx.isHoldingDownstream) ctx.pushAndPull(elem)
+      else ctx.holdUpstream()
+    }
+
+    override def onPull(ctx: AsyncContext[ByteString, Unit]) =
+      if (ctx.isFinishing) ctx.pushAndFinish(last)
+      else if (ctx.isHoldingUpstream) ctx.pushAndPull(last)
+      else ctx.holdDownstream()
+
+    override def onUpstreamFinish(ctx: AsyncContext[ByteString, Unit]) =
```
drewhk: should we clean up the timer?

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+      }
+      x
+    }
+
+    val cipherSuites = NegotiateNewSession.withCipherSuites("TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA", "TLS_RSA_WITH_AES_128_CBC_SHA")
+    def clientTls(closing: Closing) = StreamTls(sslContext, cipherSuites, Client, closing)
+    def serverTls(closing: Closing) = StreamTls(sslContext, cipherSuites, Server, closing)
+
+    trait Named {
+      def name: String = getClass.getName.reverse.dropWhile(c ⇒ "$0123456789".indexOf(c) != -1).takeWhile(_ != '$').reverse
```
drewhk: long line

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    trait CommunicationSetup extends Named {
```
drewhk: add a comment to explaint that these traits are meant to be composed together, and how it works in a nutshell

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+      def flow(leftClosing: Closing, rightClosing: Closing, flow: Flow[SslTlsInbound, SslTlsOutbound, Any]): Flow[SslTlsOutbound, SslTlsInbound, Any]
+    }
+
+    object ClientInitiates extends CommunicationSetup {
+      def flow(leftClosing: Closing, rightClosing: Closing, flow: Flow[SslTlsInbound, SslTlsOutbound, Any]) =
+        clientTls(leftClosing) atop serverTls(rightClosing).reversed join flow
```
drewhk: nice ones

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
drewhk: ah I see, it does not use trait composition, sorry for the noise

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+
+    object ServerInitiatesViaTcp extends CommunicationSetup {
+      def flow(leftClosing: Closing, rightClosing: Closing, flow: Flow[SslTlsInbound, SslTlsOutbound, Any]) = {
+        val local = server(clientTls(rightClosing).reversed join flow)
+        serverTls(leftClosing) join StreamTcp().outgoingConnection(local)
+      }
+    }
+
+    val communicationPatterns =
+      Seq(
```
drewhk: Thanks, more flexible than before!

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+    val cipherSuites = NegotiateNewSession.withCipherSuites("TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA", "TLS_RSA_WITH_AES_128_CBC_SHA")
+    def clientTls(closing: Closing) = StreamTls(sslContext, cipherSuites, Client, closing)
+    def serverTls(closing: Closing) = StreamTls(sslContext, cipherSuites, Server, closing)
+
+    trait Named {
+      def name: String = getClass.getName.reverse.dropWhile(c ⇒ "$0123456789".indexOf(c) != -1).takeWhile(_ != '$').reverse
+    }
+
+    trait CommunicationSetup extends Named {
+      def flow(leftClosing: Closing, rightClosing: Closing, flow: Flow[SslTlsInbound, SslTlsOutbound, Any]): Flow[SslTlsOutbound, SslTlsInbound, Any]
```
drewhk: maybe rename "flow" (both the param and the method name) to something more descriptive.

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+
+    def server(flow: Flow[ByteString, ByteString, Any]): InetSocketAddress = {
+      val server = StreamTcp()
+        .bind(new InetSocketAddress("localhost", 0))
+        .to(Sink.foreach(c ⇒ c.flow.join(flow).run()))
+        .run()
+      Await.result(server, 2.seconds).localAddress
+    }
+
+    object ClientInitiatesViaTcp extends CommunicationSetup {
```
drewhk: Nice ones!

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+
+    object ImpatientRHS extends PayloadScenario {
+      override def flow =
+        Flow[SslTlsInbound]
+          .mapConcat {
+            case SessionTruncated       ⇒ SessionTruncated :: Nil
+            case SessionBytes(s, bytes) ⇒ bytes.map(b ⇒ SessionBytes(s, ByteString(b)))
+          }
+          .take(5)
+          .mapAsync(10, x ⇒ later(500.millis, system.scheduler)(Future.successful(x)))
```
drewhk: why 10 if you take only 5 anyway?

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+      def inputs = List(ByteString("hello"), ByteString.empty, ByteString(" world")).map(SendBytes)
+      def output = ByteString("hello world")
+    }
+
+    object EmptyBytesLast extends PayloadScenario {
+      def inputs = List(ByteString("hello"), ByteString.empty).map(SendBytes)
+      def output = ByteString("hello")
+    }
+
+    object ImpatientRHS extends PayloadScenario {
```
drewhk: This needs some comments, "impatient" is not descriptive enough. So this takes any bytes it receives, chunks them up in single-byte fragments, takes the first 5 of them, and then passes them on delayed.

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+      def output = ByteString("TLS_RSA_WITH_AES_128_CBC_SHAhello")
+    }
+
+    object SessionRenegotiationFirstTwo extends PayloadScenario {
+      override def flow = logCipherSuite
+      def inputs = NegotiateNewSession.withCipherSuites("TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA") :: send("hello") :: Nil
+      def output = ByteString("TLS_ECDHE_RSA_WITH_AES_128_CBC_SHAhello")
+    }
+
+    val scenarios =
```
drewhk: Really nice! I had a much more limited solution, nice to see it generalized to perfection!

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+            .dropWhile(_.size < scenario.output.size)
+            .runWith(Sink.head)
+
+        Await.result(f, 3.seconds).utf8String should be(scenario.output.utf8String)
+      }
+    }
+
+  }
+
+  // TODO: Add parallel
```
drewhk: this scenario meant no loopback. I.e. the output of the receiver side is not fed back in any form to its input. There are two "parallel" data flows not affecting each other (except closing and failures of course)

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+  // TODO: Add chained parallel, loopback
```
drewhk: this meant the combination of the parallel setting client-server-client but looping back at the outermost. I am not sure if this is worth to implement at all.

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+  // TODO: Add inverted chains (client-server-server-client and client-server-client-server and server-client-client-server)
+  // TODO: Add no bytes, short bytes, long stream, large bytes
```
drewhk: this has been done

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+
+        Await.result(f, 3.seconds).utf8String should be(scenario.output.utf8String)
+      }
+    }
+
+  }
+
+  // TODO: Add parallel
+  // TODO: Add chained parallel, loopback
+  // TODO: Add inverted chains (client-server-server-client and client-server-client-server and server-client-client-server)
```
drewhk: probably not that hard to add, I am not sure if it buys anything.

akka-stream/src/main/scala/akka/stream/impl/Transfer.scala
```diff
@@ -136,7 +136,14 @@ private[akka] object Always extends TransferState {
 /**
  * INTERNAL API
  */
-private[akka] final case class TransferPhase(precondition: TransferState)(val action: () ⇒ Unit)
+private[akka] final case class TransferPhase(precondition: TransferState)(val action: () ⇒ Unit) {
+
+  def |(other: TransferPhase) = TransferPhase(precondition || other.precondition)(() ⇒ {
```
drewhk: hm, still here. I can't wait to see how you used it!

rkuhn: oh, I completely overlooked this one, it is not used

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,419 @@
...
+  protected val outputBunch = new OutputBunch(outputCount = 2, self, this)
+  outputBunch.markAllOutputs()
+
+  protected val inputBunch = new InputBunch(inputCount = 2, settings.maxInputBufferSize, this) {
+    override def onError(input: Int, e: Throwable): Unit = fail(e)
+  }
+
+  class ChoppingBlock(idx: Int, name: String) extends TransferState {
+    override def isReady: Boolean = buffer.nonEmpty
+    override def isCompleted: Boolean = false
```
drewhk: these always make me suspicious. Maybe it works here, but I have found multiple of bugs in the past that were caused by this.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -464,6 +464,18 @@ trait FlowOps[+Out, +Mat] {
...
+    transform(() ⇒ new PushStage[Out, Out] {
+      private var open = false
+      override def onPush(elem: Out, ctx: Context[Out]) =
+        if (open) ctx.push(elem)
+        else if (p(elem)) ctx.pull()
+        else {
+          open = true
+          ctx.push(elem)
+        }
+    })
```
drewhk: .named()

akka-stream/src/main/scala/akka/stream/ssl/SslTlsCipher.scala
```diff
@@ -1,376 +1,59 @@
...
-    failOutputs(e)
-    context.stop(self)
-  }
+case class SendBytes(bytes: ByteString) extends SslTlsOutbound
 
-  override protected def pumpFailed(e: Throwable): Unit = fail(e)
+sealed abstract class ClientAuth
+object ClientAuth {
+  case object None extends ClientAuth
+  case object Want extends ClientAuth
```
drewhk: Document these. What is the difference between Want and Need?

akka-stream/src/main/scala/akka/stream/ssl/SslTlsCipher.scala
```diff
@@ -1,376 +1,59 @@
 /**
  * Copyright (C) 2009-2014 Typesafe Inc. <http://www.typesafe.com>
  */
-
 package akka.stream.ssl
```
drewhk: shouldn't this be under IO? Maybe stream.io.ssl?

akka-stream/src/main/scala/akka/stream/ssl/StreamTls.scala
```diff
@@ -0,0 +1,85 @@
+package akka.stream.ssl
+
+import akka.stream.impl.StreamLayout.Module
+import akka.stream._
+import akka.stream.ssl._
+import akka.util.ByteString
+import javax.net.ssl.SSLContext
+
+object StreamTls {
```
drewhk: Hm, looking more and more at these I found it less and less nice to have StreamXXX.

akka-stream/src/main/scala/akka/stream/ssl/StreamTls.scala
```diff
@@ -0,0 +1,85 @@
...
+  }
+  case object IgnoreCancel extends Closing {
+    override def ignoreCancel = true
+    override def ignoreComplete = false
+  }
+  case object IgnoreComplete extends Closing {
+    override def ignoreCancel = false
+    override def ignoreComplete = true
+  }
+  case object IgnoreBoth extends Closing {
```
drewhk: Isn't this dangerous?

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,419 @@
...
+  import SslTlsCipherActor._
+
+  protected val outputBunch = new OutputBunch(outputCount = 2, self, this)
+  outputBunch.markAllOutputs()
+
+  protected val inputBunch = new InputBunch(inputCount = 2, settings.maxInputBufferSize, this) {
+    override def onError(input: Int, e: Throwable): Unit = fail(e)
+  }
+
+  class ChoppingBlock(idx: Int, name: String) extends TransferState {
```
drewhk: Add comments what a choppingblock means

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,419 @@
...
+     * Pour as many bytes as are available either on the chopping block or in
+     * the inputBunch’s next ByteString into the supplied ByteBuffer, which is
+     * expected to be in “read left-overs” mode, i.e. everything between its
+     * position and limit is retained. In order to allocate a fresh ByteBuffer
+     * with these characteristics, use `prepare()`.
+     */
+    def chopInto(b: ByteBuffer): Unit = {
+      b.compact()
+      if (buffer.isEmpty) {
+        buffer = inputBunch.dequeue(idx) match {
```
drewhk: this implies that this TransferState can only be used in conjunction with a NeedsInput. Can't this be codified somehow?

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,419 @@
...
+          case bs: ByteString ⇒ bs
```
drewhk: ah, this handles both user and transport inputs right? Add a comment, otherwise it is confusing how you can get a ByteString or a SendBytes on the same thing (well, you don't, but that is what needs to be commented :) )

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,419 @@
...
+        if (bs.nonEmpty) buffer = bs ++ buffer
+        prepare(b)
+      }
+
+    /**
+     * Prepare a fresh ByteBuffer for receiving a chop of data.
+     */
+    def prepare(b: ByteBuffer): Unit = {
+      b.clear()
+      b.limit(0)
```
drewhk: why do you need to set the limit to zero?

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,419 @@
...
+   * these user-data states, the engine may at any point in time also be
+   * handshaking. This is mostly transparent, but it has an influence on the
+   * outbound direction:
+   *
+   *  - if the local user triggered a re-negotiation, cork all user data until
+   *    that is finished
+   *  - if the outbound direction has been closed, trigger outbound readiness
+   *    based upon HandshakeStatus.NEED_WRAP
+   *
+   * These conditions lead to the introduction of a synthetic TransferState
```
drewhk: I was considering the same before, but I got lost in the details at some point :)

akka-stream/src/main/scala/akka/stream/ssl/StreamTls.scala
rkuhn: Hah! Living on the edge :-)

No, actually your suggestion of closing down when both have been seen has been implemented, but a test for that behavior remains to be written. Good point.

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
rkuhn: because then I can just always `.compact()` and it works—compacting a cleared buffer yields a completely full buffer (don’t you love that API?)

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,419 @@
...
+    def isReady = lastHandshakeStatus == NEED_WRAP
+    def isCompleted = false
+  }
+
+  val engineInboundOpen = new TransferState {
+    def isReady = !engine.isInboundDone()
+    def isCompleted = false
+  }
+
+  var corkUser = true
```
drewhk: what happens to a corked user? He gets screwed? :D

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
drewhk: You know my opinion on the API ;)

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,419 @@
...
+  }
+
+  val transportHasData = inputBunch.inputsOrCompleteAvailableFor(TransportIn) || transportInChoppingBlock
+
+  // bidirectional case
+  val outbound = (userHasData || engineNeedsWrap) && outputBunch.demandAvailableFor(TransportOut)
+  val inbound = transportHasData && outputBunch.demandOrCancelAvailableFor(UserOut)
+
+  // half-closed
+  val outboundHalfClosed = engineNeedsWrap && outputBunch.demandAvailableFor(TransportOut)
```
drewhk: since these are not real half-closes anyway, we might wanna rename these?

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,419 @@
...
+      catch {
+        case ex: SSLException ⇒
+          if (tracing) log.debug(s"SSLException during doWrap: $ex")
+          completeOrFlush()
+      }
+    }
+
+  val bidirectional = TransferPhase(outbound || inbound) { () ⇒
+    if (tracing) log.debug("bidirectional")
+    val continue = doInbound(isOutboundClosed = false, inbound)
```
drewhk: do you think fairness can be an issue here? Probably not, but I was not fully able to convince myself.

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,419 @@
...
+
+  def completeOrFlush(): Unit =
+    if (engine.isOutboundDone()) nextPhase(completedPhase)
+    else nextPhase(flushingOutbound)
+
+  private def doInbound(isOutboundClosed: Boolean, inboundState: TransferState): Boolean =
+    if (inputBunch.isDepleted(TransportIn) && transportInChoppingBlock.isEmpty) {
+      if (tracing) log.debug("closing inbound")
+      try engine.closeInbound()
+      catch { case ex: SSLException ⇒ outputBunch.enqueue(UserOut, SessionTruncated) }
```
drewhk: add a comment why we do this (I recall that I had a comment with a link to a Netty ticket, but if not I can dig it up)

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,419 @@
...
+            transportInChoppingBlock.putBack(transportInBuffer)
+          case _ ⇒
+            if (transportInBuffer.hasRemaining()) doUnwrap()
+            else flushToUser()
+        }
+      case CLOSED ⇒
+        flushToUser()
+        nextPhase(flushingOutbound)
+      case BUFFER_UNDERFLOW ⇒
+        flushToUser()
```
drewhk: underflow means that we just need new transport bytes right?

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,419 @@
...
+      case BUFFER_OVERFLOW ⇒
+        flushToUser()
+        transportInChoppingBlock.putBack(transportInBuffer)
```
drewhk: ah, because we failed the unwrap, we need to get our bytes back into place to retry, right?

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,419 @@
...
+  private def doWrap(): Unit = {
+    val result = engine.wrap(userInBuffer, transportOutBuffer)
+    lastHandshakeStatus = result.getHandshakeStatus
+    if (tracing) log.debug(s"wrap: status=${result.getStatus} handshake=$lastHandshakeStatus remaining=${userInBuffer.remaining} out=${transportOutBuffer.position}")
+    if (lastHandshakeStatus == FINISHED) handshakeFinished()
+    runDelegatedTasks()
+    result.getStatus match {
+      case OK ⇒
+        flushToTransport()
+        userInChoppingBlock.putBack(userInBuffer)
```
drewhk: btw, this is why I was yearning for a FIFO circular buffer.

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
rkuhn: no: while TLS itself does not really do half-closed, all the handling of its state works best if we assume that it does—it is a really weird setup

drewhk: ok

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
rkuhn: Hmm, this leads to doing both directions when both are ready, sounds fair to me. It would be an issue if the bytes from the peer keep coming in at all times (i.e. always chopping 16kiB) while the local user only sends small messages, in which case the outbound direction will be throttled, but I don’t see a good way around this.

rkuhn: A solution (in user space) would be a bunching stage in front of UserIn that just groups things into >16kiB chunks.

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
rkuhn: yes, exactly: “current packet incomplete”

akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
```diff
@@ -155,6 +155,7 @@ private[akka] object FanIn {
       if (input.inputsDepleted) {
         if (marked(id)) markedDepleted += 1
         depleted(id) = true
+        onDepleted(id)
```
rkuhn: This is killing FlexiMergeDocTest (of all things!), but validly so: eagerClose is now way too eager, leading to `onComplete` before the last element can be emitted. What was the rationale behind this change @drewhk ?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: well, and more …

akka-stream/src/main/scala/akka/stream/ssl/SslTlsCipher.scala
rkuhn: TLS is not really IO, it is only encryption … crypto …

akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
drewhk: The rationale was that it was depleted but the callback was never called. I considered this a bug.

akka-stream/src/main/scala/akka/stream/ssl/SslTlsCipher.scala
drewhk: It is a protocol, and I consider protocols IO.

akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
rkuhn: Yes, I came to the same conclusion. The only place that uses this is FlexiMerge, which built upon the misbehavior, but I’ll fix it.

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
rkuhn: we would, if this wasn’t in an internal test suite

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
rkuhn: yes, it is quite long :-)

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
rkuhn: right

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
rkuhn: yes, added a comment and changed the name

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
rkuhn: I don’t think this adds anything fundamentally new.

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
rkuhn: ditto

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
rkuhn: what should this gain?

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
rkuhn: yup; I’m removing all TODOs now

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
rkuhn: I thought about this and the reason is that it is very rarely the right thing to just stop this pump (which is what completion would do); in almost all cases we need to tell the engine and keep exchanging data.

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
rkuhn: done

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
rkuhn: not that I can see: this is handled through several layers of indirection

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
rkuhn: This is explained in the JavaDoc of `closeInbound`, so I think this is sufficiently clear.

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
rkuhn: yes, exactly

akka-stream/src/main/scala/akka/stream/ssl/SslTlsCipher.scala
rkuhn: Hmm, I don’t think we should start documenting the JSSE APIs—these are explained in the JDK JavaDocs. But it is probably alright to put in a pointer.

akka-stream/src/main/scala/akka/stream/ssl/StreamTls.scala
rkuhn: BTW: that test has been written and I verified (by hand for now—Patrik’s PR is not yet merged) that the TLS stages do terminate.

rkuhn:
Now code complete, only docs missing.

akka-stream/src/main/scala/akka/stream/ssl/SslTlsCipher.scala
drewhk: I didn't know this is JSSE parlance. It deserves then a pointer to those docs.

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
drewhk: No, the reason why we don't treat it as an error. Netty simply swallows the exception.

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
drewhk: that's what I said, too. I just tried to explain the TODOs because they are terse and was meant to me as a reminder.

rkuhn: Okay, thanks!

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,398 @@
...
+        ClientInitiates,
+        ServerInitiates,
+        ClientInitiatesViaTcp,
+        ServerInitiatesViaTcp)
+
+    trait PayloadScenario extends Named {
+      def flow: Flow[SslTlsInbound, SslTlsOutbound, Any] =
+        Flow[SslTlsInbound]
+          .map {
+            var session: SSLSession = null
```
patriknw: nasty ;-)

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,398 @@
...
+
+    object CancellingRHSIgnoresBoth extends PayloadScenario {
+      override def flow =
+        Flow[SslTlsInbound]
+          .mapConcat {
+            case SessionTruncated       ⇒ SessionTruncated :: Nil
+            case SessionBytes(s, bytes) ⇒ bytes.map(b ⇒ SessionBytes(s, ByteString(b)))
+          }
+          .take(5)
+          .mapAsync(5, x ⇒ later(500.millis, system.scheduler)(Future.successful(x)))
```
patriknw: ah, I was just about to ask if this wasn't the `after` pattern

akka-stream-tests/src/test/scala/akka/stream/io/TlsSpec.scala
```diff
@@ -0,0 +1,398 @@
...
+            .runWith(Sink.head)
+
+        Await.result(f, 8.seconds).utf8String should be(scenario.output.utf8String)
+
+        commPattern.cleanup()
+
+        // flush log so as to not mix up logs of different test cases
+        if (log.isDebugEnabled)
+          EventFilter.debug("stopgap", occurrences = 1) intercept {
+            log.debug("stopgap")
```
patriknw: clever

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,443 @@
...
+    Props(new SslTlsCipherActor(settings, sslContext, firstSession, tracing, role, closing))
+
+  final val TransportIn = 0
+  final val TransportOut = 0
+
+  final val UserOut = 1
+  final val UserIn = 1
+}
+
+class SslTlsCipherActor(settings: ActorFlowMaterializerSettings, sslContext: SSLContext,
```
patriknw: public ?

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,443 @@
...
+
+  nextPhase(bidirectional)
+
+  protected def fail(e: Throwable): Unit = {
+    // FIXME: escalate to supervisor
+    if (tracing) log.debug("fail {} due to: {}", self, e.getMessage)
+    inputBunch.cancel()
+    outputBunch.error(TransportOut, e)
+    outputBunch.error(UserOut, e)
+    context.stop(self)
```
patriknw: I wonder if this is correct? I fixed something similar here: https://github.com/akka/akka/pull/17217/files#diff-248e41e9341c718693127e44f10fc77cR244
but I think that it relies on something else in that PR to work.

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -0,0 +1,443 @@
...
+  }
+
+  override protected def pumpFailed(e: Throwable): Unit = fail(e)
+
+  override protected def pumpFinished(): Unit = {
+    inputBunch.cancel()
+    outputBunch.complete()
+    log.debug(s"STOP Outbound Closed: ${engine.isOutboundDone} Inbound closed: ${engine.isInboundDone}")
```
patriknw: use variable placeholders or if (tracing)

patriknw:
LGTM

akka-stream/src/main/scala/akka/stream/ssl/StreamTls.scala
rkuhn: Yes, it is named `akka.stream.io.SslTls` now.

rkuhn:
PLS BUILD

drewhk:
LGTM

rkuhn:
also fixes #17232

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
patriknw: @rkuhn did you verify that this is alright?

rkuhn: No, indeed I did not, thanks for the reminder! My thinking was that this stage can only fail once everything is up and running, but now I see that this is not true: if we get onError before everything is up and running then we terminate early—good catch! Since you fixed those: if this happens TLS is out of the picture and it boils down to letting the pump handle it, right? In this case, could you take care of this one as well? Otherwise I’ll pick it up later when I’m finished with the japi stuff.

PR #17234, opened 2015-04-20
=str 15707 Define default names for sources and sinks

patriknw:
- Not intended to close the ticket, but improve the situation somewhat
  by defining default names of all sources and sinks.
- The stage names (actor names) are still rather weird

For example the following

```
Source.single(1).named("aa")
  .map(identity).named("bb")
  .map(identity).named("cc")
  .runWith(Sink.publisher)
```

is materilaized with names:
flow-1-0-cc-bb-aa-singleSource : akka.stream.impl.PublisherSource@1787f2a0
flow-1-1-cc-bb-map : Map(<function1>,OperationAttributes(List(Name(map))))
flow-1-2-cc-map : Map(<function1>,OperationAttributes(List(Name(map))))
flow-1-3-publisherSink : PublisherSink

but that is out of scope for this PR

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/SubstreamSubscriptionTimeoutSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/SubstreamSubscriptionTimeoutSpec.scala
@@ -145,7 +145,7 @@ class SubstreamSubscriptionTimeoutSpec(conf: String) extends AkkaSpec(conf) {
   private def watchGroupByActor(flowNr: Int): ActorRef = {
     implicit val t = Timeout(300.millis)
     import akka.pattern.ask
-    val path = s"/user/$$a/flow-${flowNr}-1-groupBy"
+    val path = s"/user/$$a/flow-${flowNr}-1-publisherSource-groupBy"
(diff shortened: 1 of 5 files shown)


ktoso:
Weird but definitely an improvement over `unknown-operation` :-)

ktoso:
LGTM, I'll do the same right away in my pending PRs.

rkuhn:
LGTM

The “weird” naming of your initial example is completely expected and as it should be—for naming sub-flows you’ll need to declare them separately and combine using `via`. This also makes sense for organizing the source code into values with those names.

drewhk:
Yes, every call to withAttributes will wrap the preceding assembly. So what you got in your example is:

    [[[[source] -> map] -> map] -> publisherSink]

Since soure is wrapped in aa, bb and cc, the actor names represent this nesting.
if you put the two maps into a separate flow and give them a common withAttributes, it will look like

    [source -> [map -> map] -> publisherSink]

If you put the two maps into a separate flow for each, and give them separate attributies it will look like

    [source -> [map] -> [map] -> publisherSink]

(btw, this is why I liked my LinearLayout+toString additions, you could exactly see the nesting structure)

drewhk:
LGTM

Issue #17089, opened 2015-03-30
StreamTestKit: shape up and include for release


rkuhn:
doing the minimal work needed to get something out for 1.0

Issue #16070, opened 2014-10-10
Add Tap/Drain for probes in StreamTestKit


drewhk:

rkuhn:
As part of the streams TestKit we should have a ProbeSink that materializes to a probe (same for Source).

rkuhn:
These should automatically track demand.

2m:
Streams TestKit now contains `TestSink.probe` and `TestSource.probe`.

PR #17238, opened 2015-04-21
=str #17089 stream testkit

2m:
This change pulls out Subscriber/Publisher probes to main sources of `akka-stream-testkit` project and adds factory objects for various subscribers/publishers.


akka-stream-testkit/src/test/scala/akka/stream/testkit/ChainSetup.scala
```diff
@@ -22,8 +22,8 @@ class ChainSetup[In, Out](
            toPublisher: (Source[Out, _], ActorFlowMaterializer) ⇒ Publisher[Out])(implicit system: ActorSystem) =
```
patriknw: should ChainSetup be public?
I think it is too advanced for the casual user, `TestPublisher.probe` is in general very easy to use in the wrong way.

akka-stream-testkit/src/test/scala/akka/stream/testkit/ScriptedTest.scala
```diff
@@ -121,14 +123,14 @@ trait ScriptedTest extends Matchers {
     def mayRequestMore: Boolean = remainingDemand > 0
```
patriknw: should ScriptedTest be public?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefSinkSpec.scala
```diff
@@ -33,16 +34,14 @@ class ActorRefSinkSpec extends AkkaSpec {
...
-      autoPublisher.sendNext(1)
-      autoPublisher.sendNext(2)
+      val sourceRef = watch(Source.actorRef(10, OverflowStrategy.fail).to(Sink.actorRef(fw, onCompleteMessage = "done")).run())
+      sourceRef ! 1
+      sourceRef ! 2
       expectMsg(1)
       expectMsg(2)
       system.stop(fw)
-      autoPublisher.subscription.expectCancellation()
+      expectTerminated(sourceRef)
```
patriknw: this does not test the same thing any more, it does not verify "cancel stream when actor terminates"

akka-stream-testkit/src/test/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -1,177 +1,9 @@
...
-      probe.expectMsgType[Subscribe].subscription.asInstanceOf[PublisherProbeSubscription[I]]
-
-    def expectRequest(subscription: Subscription, n: Int): Unit = probe.expectMsg(RequestMore(subscription, n))
-
-    def expectNoMsg(): Unit = probe.expectNoMsg()
-    def expectNoMsg(max: FiniteDuration): Unit = probe.expectNoMsg(max)
-
-    def getPublisher: Publisher[I] = this
-  }
+object Utils {
```
patriknw: should not be public

patriknw: and file name does not match content

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowBufferSpec.scala
```diff
@@ -49,16 +49,14 @@ class FlowBufferSpec extends AkkaSpec {
...
 
     "accept elements that fit in the buffer while downstream is silent" in {
-      val publisher = StreamTestKit.PublisherProbe[Int]()
-      val subscriber = StreamTestKit.SubscriberProbe[Int]()
+      val publisher = TestPublisher.auto[Int]()
+      val subscriber = TestSubscriber.probe[Int]()
 
-      Source(publisher).buffer(100, overflowStrategy = OverflowStrategy.backpressure).runWith(Sink(subscriber))
-
-      val autoPublisher = new StreamTestKit.AutoPublisher(publisher)
```
patriknw: nice!

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -0,0 +1,293 @@
...
+
+  /**
+   * Probe that implements [[org.reactivestreams.Publisher]] interface.
+   */
+  def probe[T]()(implicit system: ActorSystem): Probe[T] = new Probe()
+
+  /**
+   * Probe that implements [[org.reactivestreams.Publisher]] interface and tracks demand.
+   */
+  def auto[T](initialPendingRequests: Long = 0)(implicit system: ActorSystem): Auto[T] = new Auto(initialPendingRequests)
```
patriknw: I'm thinking if we should name the `auto` to `probe` instead, to make it the preferred way, and name current `probe` to `manualProbe`

patriknw:
Looking good!
I think we should only provide `TestPublisher` and `TestSubscriber` as public and make other things `private[akka]`. Easier to add things than changing them.

patriknw:
ah, they are in `src/test`, sorry for the noise

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
drewhk: +1

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -0,0 +1,293 @@
...
+  /**
+   * Implementation of [[org.reactivestreams.Publisher]] that allows various assertions.
+   * This probe does not track demand. Therefore you need to expect demand before sending
+   * elements downstream.
+   */
+  class Probe[I] private[TestPublisher] ()(implicit system: ActorSystem) extends Publisher[I] {
+    private val probe: TestProbe = TestProbe()
+
+    /**
+     * Subscriber a given [[org.reactivestreams.Subscriber]] to this probe publisher.
```
drewhk: Subscribes

akka-stream-testkit/src/test/scala/akka/stream/testkit/ChainSetup.scala
drewhk: Also ChainSetup feels a bit old-school, before the times where Source and Flow has been separated. I guess this has been ported quite literally, but this could be expressed much easier with the new DSL. We should not expose this anyway to users.

akka-stream-testkit/src/test/scala/akka/stream/testkit/ScriptedTest.scala
drewhk: No, in this form no.

drewhk: Also, this does much more than a user needs. This does fuzzing of the internal protocol by trying out different timings, but users are expected to use the higher level DSL and therefore rarely need this fuzzing.

akka-stream-tests/src/test/scala/akka/stream/actor/ActorPublisherSpec.scala
```diff
@@ -139,7 +139,7 @@ class ActorPublisherSpec extends AkkaSpec(ActorPublisherSpec.config) with Implic
       val probe = TestProbe()
       val ref = system.actorOf(testPublisherProps(probe.ref))
       val p = ActorPublisher[String](ref)
-      val s = StreamTestKit.SubscriberProbe[String]()
+      val s = TestSubscriber.probe[String]()
       p.subscribe(s)
       val sub = s.expectSubscription
```
drewhk: I always disliked this style. It is the subscriber that requests, so it reads weird to call the method on the subscription. Maybe we should add an Auto style subscriber probe that handles the subscription part automatically.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowExpandSpec.scala
```diff
@@ -66,21 +62,19 @@ class FlowExpandSpec extends AkkaSpec {
...
       sub.request(1)
       subscriber.expectNext(1)
 
-      autoPublisher.sendNext(2)
-      autoPublisher.sendComplete()
+      publisher.sendNext(2)
+      publisher.sendComplete()
 
       sub.request(1)
       subscriber.expectNext(2)
```
drewhk: the above pattern should be sugared. Maybe requestNext()?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowExpandSpec.scala
```diff
@@ -97,43 +91,41 @@ class FlowExpandSpec extends AkkaSpec {
...
       subscriber.expectNext(1)
       sub.request(1)
       subscriber.expectNext(1)
 
-      var pending = autoPublisher.pendingRequests
+      var pending = publisher.pending
       // Deplete pending requests coming from input buffer
       while (pending > 0) {
-        autoPublisher.subscription.sendNext(2)
+        publisher.sendNextDisregardDemand(2)
```
drewhk: hm. Maybe have a method, autoPublisher.manualProbe() instead?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowGraphCompileSpec.scala
```diff
@@ -227,10 +226,10 @@ class FlowGraphCompileSpec extends AkkaSpec {
 
     "build with implicits and variance" in {
       FlowGraph.closed() { implicit b ⇒
-        def appleSource = b.add(Source(PublisherProbe[Apple]))
-        def fruitSource = b.add(Source(PublisherProbe[Fruit]))
-        val outA = b add Sink(SubscriberProbe[Fruit]())
-        val outB = b add Sink(SubscriberProbe[Fruit]())
+        def appleSource = b.add(Source(TestPublisher.probe[Apple]))
```
drewhk: hmm, thinking a bit, what if the probe just extends Source[T, Unit] instead? Then it can be used in the DSL without the extra hassle, it does not need a materialization trip to get to the probe. I think it is not hard to implement (and the same for the Subscriber side)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMapAsyncUnorderedSpec.scala
```diff
@@ -121,8 +118,8 @@ class FlowMapAsyncUnorderedSpec extends AkkaSpec {
         .to(Sink(c)).run()
       val sub = c.expectSubscription()
       sub.request(10)
-      val expected = (OnComplete :: List(1, 2, 4, 5).map(OnNext.apply)).toSet
-      c.probe.receiveWhile(2.seconds, messages = 5) { case x ⇒ x }.toSet should be(expected)
+      val expected = (TestSubscriber.OnComplete :: List(1, 2, 4, 5).map(TestSubscriber.OnNext.apply)).toSet
```
drewhk: Would be nice to make this sugared, too.

    probe.expectEvents (ev1, ev2, ev3)

and/or the alternative

    probe.expectScript(action1 -> List(ev1, ev2), action2 -> Nil)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphFlexiMergeSpec.scala
```diff
@@ -186,12 +185,11 @@ class GraphFlexiMergeSpec extends AkkaSpec {
...
       }.run()
 
-      val s = SubscriberProbe[String]
+      val s = TestSubscriber.probe[String]
       p.subscribe(s)
       val sub = s.expectSubscription()
       sub.request(10)
-      (s.probe.receiveN(6).map { case OnNext(elem) ⇒ elem }).toSet should be(
-        Set("a", "b", "c", "d", "e", "f"))
+      (for (_ ← 1 to 6) yield s.expectNext()).toSet should be(Set("a", "b", "c", "d", "e", "f"))
```
drewhk: make this a pattern, probe.expectNextN for example. Maybe it should also handle completion, truncating the Seq if completed early (this would make the "Await(x.grouped().runWith(Sink.head)) should be (y)" pattern go await, since it would be x.runWith(probe).expectNextN() should be(y) given that probe also extends Sink)

drewhk:
LGTM, but I would be super happy to see some sugar to make this more smooth experience without spending too much effort.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefSinkSpec.scala
2m: Yes. I experimented with a few test cases when converting them from `TestPublisher.Probe`  to `Source.actorRef` and left this one converted. But you are right that it does not test the same thing anymore. I will revert.

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
2m: Agreed.

akka-stream-tests/src/test/scala/akka/stream/actor/ActorPublisherSpec.scala
2m: Will add auto style subscriber with one subscription tracking and sugar methods.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowExpandSpec.scala
2m: Sure. Will add to auto style subscriber probe.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowExpandSpec.scala
2m: This would not work. It requires access to the tracked subscription and not the underlying publisher.

drewhk: I meant that it wraps everything that the autoPublisher has except the tracked demand. The autoPublisher has the subscription AFAIK, and then it can give it to the manual probe.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMapAsyncUnorderedSpec.scala
2m: I doubt that this needs additional sugar. I think this assertion can be rewriten to:
``` scala
c.expectNext(1, 2, 4, 5)
c.expectComplete()
```

drewhk: ok

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowGraphCompileSpec.scala
2m: I experimented with extending a Sink and while not extending Sink itself (since it is final) I created a new Module for TestSink.probe which allows doing assertions in fluent fashion like so:
``` scala
FlowGraph.closed(in1, in2, TestSink.probe[String])((i1, i2, o) ⇒ o) { implicit b ⇒
  (in1, in2, o) ⇒
    val m1 = b.add(fairString)
    val m2 = b.add(fairString)

    // format: OFF
    in1.outlet ~> m1.in(0)
    in2.outlet ~> m1.in(1)

    Source(List("A", "B", "C", "D", "E", "F")) ~> m2.in(0)
                                        m1.out ~> m2.in(1)
                                                  m2.out ~> o.inlet
  // format: ON
}.run()
 .request(20)
 .expectNextUnordered("a", "b", "c", "d", "e", "f", "A", "B", "C", "D", "E", "F")
 .expectComplete()
```

drewhk: Hmm. It is sad that Sink is final. We can still make probe.sink work, because if you have the Module, you can create a Sink from it, since there is an akka private constructor: https://github.com/akka/akka/blob/release-2.3-dev/akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala#L76

It is not as nice though:

    x.to(Sink.publisher(probe)) // current
    x.to(probe) // If probe could be a Sink
    x.to(probe.sink) // If probe cannot be a sink, but it can still create one

What makes me wonder though is if we want to expose methods like .via(Graph[FlowShape]) and .to(Graph[SinkShape]) in the DSL. Then x.to(probe) would work and Sink could remain final.

@rkuhn WDYT?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphFlexiMergeSpec.scala
2m: Added `expectNextUnordered`.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowExpandSpec.scala
2m: Your suggestion is to expose the tracked subscription. It was exposed before this change. I made it private so we have more leeway in the future. `probe.subscription.sendNext` is now encapsulated in `probe.sendNextDisregardDemand`. Do you think it is a bad idea to not expose subscription in a probe?

drewhk: I didn't mean to expose it, it can stay a private field, but I thought that you can pass it to the unsafe probe in a private constructor. But I realized that the unsafe probe does not track the subscription field anyway. In other words, it is fine as it is now.

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -0,0 +1,293 @@
...
+     */
+    def pending: Long = pendingRequests
+
+    def sendNext(elem: T): Unit = {
+      if (pendingRequests == 0) pendingRequests = subscription.expectRequest()
+      pendingRequests -= 1
+      subscription.sendNext(elem)
+    }
+
+    def sendNextDisregardDemand(elem: T): Unit = subscription.sendNext(elem)
```
ktoso: `sendNextDisregardingDemand`?

drewhk: forceNext, unsafeSendNext

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -0,0 +1,293 @@
...
+
+    def expectRequest(): Long = subscription.expectRequest()
+
+    def expectCancellation(): Unit = subscription.expectCancellation()
+  }
+
+  sealed trait PublisherEvent extends DeadLetterSuppression
+  final case class Subscribe(subscription: Subscription) extends PublisherEvent
+  final case class CancelSubscription(subscription: Subscription) extends PublisherEvent
+  final case class RequestMore(subscription: Subscription, elements: Long) extends PublisherEvent
```
ktoso: If users see this perhaps `Request` would be better? `RequestMore` is old naming we had in some of the initial reactive streams proposals

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -0,0 +1,293 @@
...
+      expectError()
+    }
+
+    def expectSubscriptionAndComplete(): Unit = {
+      val sub = expectSubscription()
+      sub.request(1)
+      expectComplete()
+    }
+
+    def expectNextOrError(element: I, cause: Throwable): Either[Throwable, I] = {
```
ktoso: missing `expectNextOrComplete`?

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -0,0 +1,293 @@
...
+  }
+
+  sealed trait SubscriberEvent extends DeadLetterSuppression
+  final case class OnSubscribe(subscription: Subscription) extends SubscriberEvent
+  final case class OnNext[I](element: I) extends SubscriberEvent
+  final case object OnComplete extends SubscriberEvent
+  final case class OnError(cause: Throwable) extends SubscriberEvent
+}
+
+private[testkit] object StreamTestKit {
```
ktoso: Maybe separate file as `TestSubscription`?

akka-stream-testkit/src/test/scala/akka/stream/testkit/StreamTestKit.scala
ktoso: It ocnly contains the `TE`, so moving it to `StreamTestKit` or something like that would work well I think :)

drewhk: I am not sure we should expose it to users though.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowExpandSpec.scala
ktoso: +1

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphFlexiMergeSpec.scala
```diff
@@ -272,11 +270,11 @@ class GraphFlexiMergeSpec extends AkkaSpec {
       }
 
       val p = g.run()
-      val s = SubscriberProbe[String]
+      val s = TestSubscriber.probe[String]
       p.subscribe(s)
       val sub = s.expectSubscription()
       sub.request(100)
-      (s.probe.receiveN(2).map { case OnNext(elem) ⇒ elem }).toSet should be(Set("((+,b),f)", "((+,a),e)"))
+      (for (_ ← 1 to 2) yield s.expectNext()).toSet should be(Set("((+,b),f)", "((+,a),e)"))
```
ktoso: This looks like we want `expectNextAll(Seq[T])` in addition to the `expectNextUnordered(Seq[T])`?

akka-stream-testkit/src/test/scala/akka/stream/testkit/StreamTestKit.scala
ktoso: Very true actually

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
2m: I refactored it a bit and now this is all akka private api.

ktoso: excellent :-)

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
2m: I like `unsafeSendNext` most.

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
2m: Added.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphFlexiMergeSpec.scala
2m: There is already a `def expectNext(e1: I, e2: I, es: I*)` which expects multiple elements in order.

ktoso: Almost, doesn't make the case "I have a seq, and know those elements will come in sequence", I'd have to `expectNext(s.head, s.tail.head, s.tail.tail)`, I'd like to say `expectNextN(s: Seq[T])`

2m: I see. Sure.

rkuhn:
needs a rebase; otherwise merge? (I did not review yet, but I’ll trust you)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMapAsyncUnorderedSpec.scala
```diff
@@ -110,19 +107,17 @@ class FlowMapAsyncUnorderedSpec extends AkkaSpec {
...
           if (n == 3) throw new RuntimeException("err3") with NoStackTrace
           else n
         })
         .withAttributes(supervisionStrategy(resumingDecider))
-        .to(Sink(c)).run()
-      val sub = c.expectSubscription()
-      sub.request(10)
-      val expected = (OnComplete :: List(1, 2, 4, 5).map(OnNext.apply)).toSet
-      c.probe.receiveWhile(2.seconds, messages = 5) { case x ⇒ x }.toSet should be(expected)
+        .runWith(TestSink.probe[Int])
```
drewhk: Nice!

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphFlexiMergeSpec.scala
```diff
@@ -185,18 +184,13 @@ class GraphFlexiMergeSpec extends AkkaSpec {
...
-        Set("a", "b", "c", "d", "e", "f"))
-      s.expectComplete()
+        .request(10)
+        .expectNextUnordered("a", "b", "c", "d", "e", "f")
+        .expectComplete()
     }
 
     "be able to have two fleximerges in a graph" in {
-      val p = FlowGraph.closed(in1, in2, out)((i1, i2, o) ⇒ o) { implicit b ⇒
+      FlowGraph.closed(in1, in2, TestSink.probe[String])((i1, i2, o) ⇒ o) { implicit b ⇒
```
drewhk: nice!

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -0,0 +1,443 @@
...
+    override type Self = Probe[T]
+
+    private lazy val subscription = expectSubscription()
+
+    def request(n: Long): Self = {
+      subscription.request(n)
+      this
+    }
+
+    def requestNext(element: T): Self = {
```
drewhk: this one is very useful

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowExpandSpec.scala
```diff
@@ -22,68 +22,54 @@ class FlowExpandSpec extends AkkaSpec {
...
-      val sub = subscriber.expectSubscription()
+      Source(publisher).expand(seed = i ⇒ i)(extrapolate = i ⇒ (i, i)).to(Sink(subscriber)).run()
 
       for (i ← 1 to 100) {
         // Order is important here: If the request comes first it will be extrapolated!
-        autoPublisher.sendNext(i)
-        sub.request(1)
-        subscriber.expectNext(i)
+        publisher.sendNext(i)
+        subscriber.requestNext(i)
```
drewhk: nice!

drewhk:
LGTM

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -0,0 +1,443 @@
...
+    def expectRequest(): Long = publisherProbe.expectMsgPF() {
+      case RequestMore(sub, n) if sub eq this ⇒ n
+    }
+
+    def expectCancellation(): Unit = publisherProbe.fishForMessage() {
+      case CancelSubscription(sub) if sub eq this ⇒ true
+      case RequestMore(sub, _) if sub eq this     ⇒ false
+    }
+
+    def sendNext(element: I): Unit = subscriber.onNext(element)
```
patriknw: Not for this PR, but create a ticket if you like the idea:
We could track demand and throw exception if this is invoked when there is no demand.
We could have a power user flag that doesn't do that, to be able to simulate an invalid publisher.

drewhk: Isn't it already there with PublisherProbe? It also has an unsafeSendNext

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -0,0 +1,443 @@
...
+    def expectNext(e1: I, e2: I, es: I*): Self = expectNextN(e1 +: e2 +: es)
+
+    def expectNextUnordered(e1: I, e2: I, es: I*): Self = expectNextUnorderedN(e1 +: e2 +: es)
+
+    /**
+     * Expect and return a data element.
+     */
+    def expectNext(): I = probe.expectMsgType[OnNext[I]].element
+
+    def expectNextN(all: Seq[I]): Self = {
```
patriknw: `immutable.Seq`

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -0,0 +1,443 @@
...
+      all.foreach(e ⇒ probe.expectMsg(OnNext(e)))
+      self
+    }
+
+    def expectNextUnorderedN(all: Seq[I]): Self = {
```
patriknw: `Set` instead of `Seq`?

ktoso: I can see how expecting a certain element multiple times makes sense, let's keep `immutable.Seq`

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -0,0 +1,443 @@
...
+      @annotation.tailrec def expectOneOf(all: Seq[I]): Unit = all match {
+        case Nil ⇒
+        case list ⇒
+          val next = expectNext()
+          assert(all.exists(_ == next), s"expected one of $all, but received $next")
+          expectOneOf(all.filterNot(_ == next))
```
patriknw: if `all` is a `Seq` shouldn't this only remove one element?

ktoso: Yes, I'd vote for adjusting impl to work like you suggest (didn't review in full before just noticed your comment above)

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
patriknw: yes, that is a good point, and then below should be changed

patriknw:
have you tried with java, or is that out of scope?

patriknw:
LGTM, expecting some time for rebasing this one so don't make it perfect (the ticket only expects some shape up)

2m:
Today I tried it with Java and it works after a little bit of shuffling classes around. Will address feedback comments and rebase by tomorrow standup.

patriknw:
sounds great, thanks!

2m:
Rebased and added documentation stubs with example code in Scala and Java.

akka-docs-dev/rst/java/stream-testkit.rst
```diff
@@ -0,0 +1,25 @@
...
+Probe Sink
+==========
+
+Using probe as a `Sink` allows manual control over demand and assertions over elements coming downstream. Streams testkit provides a sink that materializes to a :class:`TestSubscriber.Probe`.
+
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/StreamTestKitDocTest.java#test-sink-probe
+
+Probe Source
+============
+
```
patriknw: it should mention that most tests can use a strict collection `Source(iterable)` as source

akka-docs-dev/rst/java/stream-testkit.rst
```diff
@@ -0,0 +1,25 @@
...
+###############
+Testing streams
+###############
+
+Akka Streams comes with an :mod:`akka-stream-testkit` module that provides tools which can be used for controlling and asserting various parts of the stream pipeline.
+
+Probe Sink
+==========
+
```
patriknw: `Sink.actorRef` works in many cases and then using ordinary actor testkit should be familiar to many users

akka-samples/akka-docs-java-lambda/pom.xml
```diff
@@ -31,6 +31,11 @@
       <version>1.0-SNAPSHOT</version>
     </dependency>
     <dependency>
+      <groupId>com.typesafe.akka</groupId>
+      <artifactId>akka-stream-testkit-experimental_2.10</artifactId>
```
patriknw: interesting, what will happen if this is run with 2.11?
doesn't have to be solved now

patriknw: it will not be run with 2.11, no problem

patriknw:
LGTM
mention of Source(iterable), Source.actorRef and Sink.actorRef would give bonus points

akka-docs-dev/rst/java/stream-testkit.rst
```diff
@@ -0,0 +1,25 @@
...
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/StreamTestKitDocTest.java#test-sink-probe
+
+Probe Source
+============
+
+A source that materializes to :class:`TestPublisher.Probe` can be used for asserting demand or controlling when stream is completed or ended with an error.
+
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/StreamTestKitDocTest.java#test-source-probe
+
+*TODO*
```
drewhk: File a ticke to extend these sections

2m: Added #17291 which also should take care of Patrik's suggestions.

rkuhn:
Test failures are unrelated, merge?

2m:
One failure was related - fix for ConnectionPoolSpec string matching did not work on a jenkins machine. If the current validation passes or fails with SynchFileSourceSpec - we can merge.

rkuhn:
fixes #17089

rkuhn:
also fixes #16070?

Issue #17176, opened 2015-04-10
failed: GraphBalanceSpec


patriknw:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/2273/consoleFull

PR #17241, opened 2015-04-21
=str #17176 do not check for arbitrary conditions that depends on thread scheduling

2m:
Checking for at least one landed element across balanced sinks should be enough.

References #17176

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphBalanceSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphBalanceSpec.scala
@@ -2,11 +2,13 @@ package akka.stream.scaladsl
 
 import scala.concurrent.Await
 import scala.concurrent.duration._
+import scala.concurrent.Future
 
 import akka.stream.ActorFlowMaterializer
 
 import akka.stream.ActorFlowMaterializerSettings
 import akka.stream.testkit._
+import akka.stream.testkit.scaladsl._
 import akka.stream.testkit.Utils._
 
 class GraphBalanceSpec extends AkkaSpec {
@@ -51,7 +53,7 @@ class GraphBalanceSpec extends AkkaSpec {
           val balance = b.add(Balance[Int](2, waitForAllDownstreams = true))
           Source(List(1, 2, 3)) ~> balance.in
           balance.out(0) ~> Sink(s1)
-          balance.out(1) ~> p2Sink.inlet
+          balance.out(1) ~> p2Sink
       }.run()
 
       val sub1 = s1.expectSubscription()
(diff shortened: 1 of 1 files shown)


drewhk:
LGTM

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphBalanceSpec.scala
```diff
@@ -127,18 +129,21 @@ class GraphBalanceSpec extends AkkaSpec {
...
+          balance.out(1) ~> o2
+          balance.out(2) ~> o3
       }.run()
 
-      Await.result(r1, 3.seconds) should be(numElementsForSink +- 2000)
-      Await.result(r2, 3.seconds) should be(numElementsForSink +- 2000)
-      Await.result(r3, 3.seconds) should be(numElementsForSink +- 2000)
+      import system.dispatcher
+      val sum = Future.sequence(results).map { res ⇒
+        res should not contain 0
```
patriknw: Isn't this running in some other thread, not the test main thread, and therefore assertion error here will not fail the test.

patriknw: do we think > 0 elements is good enough check to verify "fairly balance between three outputs" ?

2m: It is. But there is `Await.result` on that future below which will wait for result and therefore execute this assertion.

2m: > do we think > 0 elements is good enough check to verify "fairly balance between three outputs" ?

I am not sure if any other larger arbitrary number is better, as it increases the chance of this testcase failing.

drewhk: Yes, we should probably add a separate test case that just manually requests elements one-by-one at different outputs and so verify that balancing works. I.e. sink1 requests 1, awaits element, then sink2 requests 2, awaits element, etc.

patriknw: > But there is Await.result on that future below which will wait for result and therefore execute this assertion.

clever, I missed that

ktoso:
LGTM

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphBalanceSpec.scala
```diff
@@ -109,37 +111,65 @@ class GraphBalanceSpec extends AkkaSpec {
...
+      p3.requestNext(3)
+      p2.requestNext(4)
+      p1.requestNext(5)
+      p3.requestNext(6)
+      p1.requestNext(7)
+
+      p1.expectComplete()
+      p2.expectComplete()
+      p3.expectComplete()
     }
```
patriknw: great

patriknw:
LGTM

drewhk:
merge?

PR #17243, opened 2015-04-21
=str Use some more DeadLetterSuppression

patriknw:
noticed `SubstreamOnComplete` dead letter when running normal http client/server


drewhk:
There are also many in remoting.

ktoso:
LGTM

Issue #17145, opened 2015-04-07
Resurrect constructor to create Flow from Sink and Source


jrudolph:
(or maybe I just couldn't find it but @rkuhn encouraged me to create a ticket).

drewhk:
There isn't any right now, but it is easy to add.

drewhk:
(added to RC1 due to its simplicity)

drewhk:
This is invalid, there is already a wrap method.

drewhk:
but the Java counterpart is missing

jrudolph:
Maybe because it has already been merged?

jrudolph:
And yes, the Java counterpart is missing.

drewhk:
I fix this ticket in the scope of this, since exactly those method's signature will change: https://github.com/akka/akka/issues/17254

drewhk:
Fixed

Issue #17129, opened 2015-04-02
http server blueprint is reused but cannot be


jrudolph:
The server blueprint currently isn't reusable but it actually is reused in `HttpExt.bind`. Because reusage isn't prevented currently this currently introduces hard to detect errors.

The short-term solution should be not to reuse the HTTP server blueprint.

In the long-term it should be attempted to make the server blueprint reusable. Currently this isn't possible because it needs to access the ActorRef of the token source for Expect 100 support but there's no way to access it currently (can hopefully be fixed when #16168 is implemented). So, the token source writes the ActorRef into a mutable volatile var that is shared over all connections (!). Currently, the cell is overwritten which each new materialization of the blueprint, i.e. Expect 100 support only works on the latest connection and can be disrupted by older connections also trying to use Expect 100 support.

Also, for similar reasons initial websocket support won't be reusable for now.

sirthias:
Yes. Hopefully https://github.com/akka/akka/pull/17097 will be merged soon and we can fix this properly.

PR #17248, opened 2015-04-21
Server-side Websocket support

### LINKED ISSUE #16887: Websockets Server side API and implementation discussion

jrudolph:
I rebased, cut into proper commits and cleaned up things. I also added a first simple directive for websocket in akka-http.

Fixes #16887, #17145, and a preliminary fix for #17129.

/cc @sirthias

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/engine/ws/FrameOutHandler.scala
+++ b/akka-http-core/src/main/scala/akka/http/engine/ws/FrameOutHandler.scala
@@ -1,0 +1,132 @@
+/*
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.engine.ws
+
+import scala.concurrent.duration.FiniteDuration
+
+import akka.stream.stage._
+import akka.http.util.Timestamp
+import FrameHandler.{ UserHandlerCompleted, ActivelyCloseWithCode, PeerClosed, DirectAnswer }
+import Websocket.Tick
+
+/**
+ * Implements the transport connection close handling at the end of the pipeline.
+ *
+ * INTERNAL API
+ */
+private[http] class FrameOutHandler(serverSide: Boolean, _closeTimeout: FiniteDuration) extends StatefulStage[AnyRef, FrameStart] {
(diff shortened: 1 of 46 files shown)


akka-http-core/src/main/scala/akka/http/engine/ws/FrameOutHandler.scala
```diff
@@ -0,0 +1,132 @@
...
+private[http] class FrameOutHandler(serverSide: Boolean, _closeTimeout: FiniteDuration) extends StatefulStage[AnyRef, FrameStart] {
+  def initial: StageState[AnyRef, FrameStart] = Idle
+  def closeTimeout: Timestamp = Timestamp.now + _closeTimeout
+
+  object Idle extends CompletionHandlingState {
+    def onPush(elem: AnyRef, ctx: Context[FrameStart]): SyncDirective = elem match {
+      case start: FrameStart   ⇒ ctx.push(start)
+      case DirectAnswer(frame) ⇒ ctx.push(frame)
+      case PeerClosed(code, reason) if !code.exists(Protocol.CloseCodes.isError) ⇒
+        // let user complete it, FIXME: maybe make configurable? immediately, or timeout
```
jrudolph: The [spec says](http://tools.ietf.org/html/rfc6455#section-5.5.1):

> If an endpoint receives a Close frame and did not previously send a
   Close frame, the endpoint MUST send a Close frame in response.  (When
   sending a Close frame in response, the endpoint typically echos the
   status code it received.)  It SHOULD do so as soon as practical.  An
   endpoint MAY delay sending a Close frame until its current message is
   sent (for instance, if the majority of a fragmented message is
   already sent, an endpoint MAY send the remaining fragments before
   sending a Close frame).

We currently, fully support half-closed connections, so we wait for the user-handler to complete and only then send our close frame. "as soon as practical" is currently implemented as "as soon as the user-handler deems right". Is this enough or do we need automatic closing (after a timeout?) WDYT?

Issue #17239, opened 2015-04-21
Move akka.stream.japi classes to akka.japi.function


patriknw:
with new names

patriknw:
not merged to master yet, but there is a PR for that

PR #17250, opened 2015-04-21
=act #17239 Add akka.japi.function with variance

patriknw:
- and deprecate corresponding classes in akka.japi

I'm not sure how adding overloaded method for deprecation works with java8 lambdas, e.g. `AbstractPersistentActorWithAtLeastOnceDelivery`.

I have not deprecated the `Creator` and not changed `Props`. I'm not sure if we need to, or if we should have both `akka.japi.Creator` and `akka.japi.function.Creator`?


patriknw:
this is also missing migration guide, so far

akka-persistence/src/main/scala/akka/persistence/AtLeastOnceDelivery.scala
```diff
@@ -380,6 +384,8 @@ abstract class AbstractPersistentActorWithAtLeastOnceDelivery extends AbstractPe
    * This method will throw [[AtLeastOnceDelivery.MaxUnconfirmedMessagesExceededException]]
    * if [[numberOfUnconfirmed]] is greater than or equal to [[#maxUnconfirmedMessages]].
    */
-  def deliver(destination: ActorPath, deliveryIdToMessage: akka.japi.Function[java.lang.Long, Object]): Unit =
+  def deliver(destination: ActorPath, deliveryIdToMessage: akka.japi.function.Function[java.lang.Long, Object]): Unit =
     super.deliver(destination, id ⇒ deliveryIdToMessage.apply(id))
+
+  // FIXME @deprecated deliver with akka.japi.Function. How would java8 know which one to pick when param is lambda?
```
ktoso: Bad news on this one - javac won't be able to figure out which target type you want this lambda to be expanded to when it sees `deliver(..., in -> { ... })` because both match (arity + types + return type).

The error would be along the lines of:

```
Example.java:25: error: reference to doIt is ambiguous
        whoops.doIt(() -> {
              ^
  both method doIt(Fun) in Whoops and method doIt(FunBetter) in Whoops match
```

(It's solvable as `deliver(..., (akka.japi.function.Function<Long, Object>)(l) -> l)` but that's not really helping...)

patriknw: Thanks for checking, I suspected that. Good news is that this is still experimental.

patriknw:
@rkuhn @ktoso @2m please review this, important for 2.4 binary compatibility

patriknw:
We should also add the generated functions.

patriknw:
added the generated functions

akka-persistence/src/main/scala/akka/persistence/AtLeastOnceDelivery.scala
rkuhn: But does this objection not apply to all changes in this PR (with the debatable exception of UntypedActor)?

rkuhn:
I’m not sure about this one: we cannot keep the deprecated variants without destroying Java 8 lambda usage, right? But we also cannot just break source compatibility on this scale without deprecation—or can we? We _could_ potentially keep binary compatibility by keeping the old signatures as SYNTHETIC (i.e. bridge) methods.

akka-persistence/src/main/scala/akka/persistence/AtLeastOnceDelivery.scala
patriknw: You mean that we can break akka-persistence in any case? We have always tried to not break it if we don't have to.

It is only AbstractPersistentActor and AtLeastOnceDelivery that are intended to be used from java 8.

rkuhn: What about the Future methods?

patriknw: good point, um I don't know what to do.

What would happen if we introduced the variance in the old functions?
We had some thought about that it would be binary compatible but not source compatible if used from Scala. Would that be acceptable for 2.4? Then we would have to break binary compatibility of streams when moving it to 2.4, but that might be alright.

patriknw:
I have included the japi functions, cherry-picked as is from release-2.3 without any deprecations, in the [BC PR](https://github.com/akka/akka/pull/17354). Whatever we decide here that should be the first step anyway, and it might be the only thing we can do.

akka-persistence/src/main/scala/akka/persistence/AtLeastOnceDelivery.scala
rkuhn: I think it makes sense to have a complete set of these functions in `akka.japi.function`, even if they duplicate things (with slight variation) that exist in `akka.japi`. We may break compatibility in Akka 3.0, but for now we could just punt and do nothing, i.e. back out the changes that use the new functions where problematic (since that usage does not gain us anything besides internal consistency anyway). Then we can talk with Josh _et al_ about how we could tackle such a change.

patriknw: yes, that was my conclusion also, after thinking some more about it (I dislike that we can't keep the code base dry and clean, but that is the price we have to pay).
I think that means that I can close this PR without action, but I will check if I did any other changes that we actually want.

rkuhn: good, thanks!

patriknw:
nothing good here

Issue #17194, opened 2015-04-14
Attempt cleaning of DirectByteBuffer iso relying on GC of native mem


viktorklang:
http://stackoverflow.com/questions/1854398/how-to-garbage-collect-a-direct-buffer-java

apply to https://github.com/akka/akka/blob/release-2.3-dev/akka-actor/src/main/scala/akka/io/DirectByteBufferPool.scala

jrudolph:
WDYM? 

As there's no specified way of securely releasing a direct buffer the only way is relying on GC.

The pool will only over-allocate buffers( which later need to be GC'd) if the system is under serious load (i.e. for Akka IO more concurrent IO is going on than the size of the pool). If your application needs to live under those conditions you need to tweak settings anyway.

viktorklang:
@jrudolph The point is that under high load it will generate garbage that lives in a more limited heap and takes longer to free—exactly conditions that are suboptimal. Having a means of -trying- to do it more reliably can't be bad?

jrudolph:
@viktorklang Fair enough. What remains would be a trade-off where instantly deallocating the buffer would introduce even more work under high load but which admittedly is the safer and more predictable solution (because it is less likely to OOM fatally), against a more fundamental solution which would entail sizing the pool appropriately to the expected load.

viktorklang:
@jrudolph Under ideal circumstances it would adaptively resize itself to a local optimum, but statically sizing the pool seems to require a "total view", which can be hard for developers and possibly impossible for Ops. I am of course open to alternative solutions, the one outlined in the Issue seems to be quite unintrusive / easy to reason about.

jrudolph:
Yes, I think I can agree on that. 

To clarify: the outlined solution in SO is trying to use `sun.misc.Cleaner` reflectively to clean a direct buffer instead of just dropping the reference.

viktorklang:
Affirmative

hepin1989:
https://github.com/netty/netty/blob/master/common/src/main/java/io/netty/util/internal/Cleaner0.java for direct and https://github.com/netty/netty/blob/master/common/src/main/java/io/netty/util/internal/PlatformDependent0.java#L415 for unsafe

PR #17258, opened 2015-04-22
=act #17216 fix DirectByteBufferPoll locking

### LINKED ISSUE #17216: fix open-coded spinlock in DirectByteBufferPool
from Akara on the mailing list:
---
Scala 2.11.6, Akka 2.3.9, Spray 1.3.2.

When testing a particular workload, we ran into CPU spikes. The avg CPU load is less than 5% but about 6-7 minutes into the run, we start to see CPU spiking to near 100% lasting for several seconds. This repeats itself every 6-7 minutes. We can't correlate this with a particular GC activity just yet. We tested against Java 7_45 but tried Java 8 without much change.

Spray is configured on the default-dispatcher (Spray default).

Trying to capture thread dumps at the time spikes happening reveal 9 threads spinning on the followings:
~~~
"default-akka.actor.default-dispatcher-11" prio=10 tid=0x0000000001b2b800 nid=0x3752 runnable [0x00007f1191d40000]
   java.lang.Thread.State: RUNNABLE
	at akka.io.DirectByteBufferPool.takeBufferFromPool(DirectByteBufferPool.scala:41)
	at akka.io.DirectByteBufferPool.acquire(DirectByteBufferPool.scala:31)
	at akka.io.TcpConnection.doRead(TcpConnection.scala:231)
	at akka.io.TcpConnection$$anonfun$connected$1.applyOrElse(TcpConnection.scala:87)
	at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:171)
	at akka.actor.Actor$class.aroundReceive(Actor.scala:465)
	at akka.io.TcpConnection.aroundReceive(TcpConnection.scala:28)
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:516)
	at akka.actor.ActorCell.invoke(ActorCell.scala:487)
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:254)
	at akka.dispatch.Mailbox.run(Mailbox.scala:221)
	at akka.dispatch.Mailbox.exec(Mailbox.scala:231)
~~~
2 threads show this spinning:
~~~
"default-akka.actor.default-dispatcher-8" prio=10 tid=0x0000000001d6f000 nid=0x374f runnable [0x00007f11928d2000]
   java.lang.Thread.State: RUNNABLE
	at akka.io.DirectByteBufferPool.offerBufferToPool(DirectByteBufferPool.scala:66)
	at akka.io.DirectByteBufferPool.release(DirectByteBufferPool.scala:34)
	at akka.io.TcpConnection.doRead(TcpConnection.scala:245)
	at akka.io.TcpConnection$$anonfun$connected$1.applyOrElse(TcpConnection.scala:87)
	at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:171)
	at akka.actor.Actor$class.aroundReceive(Actor.scala:465)
	at akka.io.TcpConnection.aroundReceive(TcpConnection.scala:28)
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:516)
	at akka.actor.ActorCell.invoke(ActorCell.scala:487)
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:254)
	at akka.dispatch.Mailbox.run(Mailbox.scala:221)
	at akka.dispatch.Mailbox.exec(Mailbox.scala:231)
~~~
1 thread is on I/O:

~~~
"default-akka.actor.default-dispatcher-4" prio=10 tid=0x0000000001be2800 nid=0x36ff runnable [0x00007f11924b0000]
   java.lang.Thread.State: RUNNABLE
	at sun.nio.ch.SocketChannelImpl.read(SocketChannelImpl.java:295)
	- locked <0x00000007d3b7e940> (a java.lang.Object)
	at akka.io.TcpConnection.innerRead$1(TcpConnection.scala:216)
	at akka.io.TcpConnection.doRead(TcpConnection.scala:232)
	at akka.io.TcpConnection$$anonfun$connected$1.applyOrElse(TcpConnection.scala:87)
	at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:171)
	at akka.actor.Actor$class.aroundReceive(Actor.scala:465)
	at akka.io.TcpConnection.aroundReceive(TcpConnection.scala:28)
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:516)
	at akka.actor.ActorCell.invoke(ActorCell.scala:487)
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:254)
	at akka.dispatch.Mailbox.run(Mailbox.scala:221)
	at akka.dispatch.Mailbox.exec(Mailbox.scala:231)
~~~

All in all, during this time there is no app logic being executed on the default dispatcher at all (12 threads on the default-dispatcher).

rkuhn:
instead of spinning unboundedly use ReentrantLock.tryLock—it is a best
effort pool anyway and contention shall not lock the whole application

fixes #17216

### DIFF
--- a/akka-actor/src/main/scala/akka/io/DirectByteBufferPool.scala
+++ b/akka-actor/src/main/scala/akka/io/DirectByteBufferPool.scala
@@ -4,9 +4,9 @@
 
 package akka.io
 
-import java.util.concurrent.atomic.AtomicBoolean
 import java.nio.ByteBuffer
-import annotation.tailrec
+import java.util.concurrent.locks.ReentrantLock
+import java.util.concurrent.TimeUnit
 
 trait BufferPool {
   def acquire(): ByteBuffer
@@ -23,9 +23,9 @@ trait BufferPool {
  * being freed by normal garbage collection.
  */
 private[akka] class DirectByteBufferPool(defaultBufferSize: Int, maxPoolEntries: Int) extends BufferPool {
-  private[this] val locked = new AtomicBoolean(false)
-  private[this] val pool: Array[ByteBuffer] = new Array[ByteBuffer](maxPoolEntries)
-  private[this] var buffersInPool: Int = 0
+  private[this] val lock = new ReentrantLock
+  private[this] val pool = new Array[ByteBuffer](maxPoolEntries)
+  private[this] var buffersInPool = 0
 
   def acquire(): ByteBuffer =
     takeBufferFromPool()
@@ -36,32 +36,32 @@ private[akka] class DirectByteBufferPool(defaultBufferSize: Int, maxPoolEntries:
   private def allocate(size: Int): ByteBuffer =
     ByteBuffer.allocateDirect(size)
 
-  @tailrec
-  private final def takeBufferFromPool(): ByteBuffer =
-    if (locked.compareAndSet(false, true)) {
-      val buffer =
-        try if (buffersInPool > 0) {
+  private final def takeBufferFromPool(): ByteBuffer = {
+    var buffer: ByteBuffer = null
+
+    if (lock.tryLock(1, TimeUnit.MILLISECONDS))
+      try
+        if (buffersInPool > 0) {
           buffersInPool -= 1
-          pool(buffersInPool)
-        } else null
-        finally locked.set(false)
+          buffer = pool(buffersInPool)
+        }
+      finally lock.unlock()
 
-      // allocate new and clear outside the lock
-      if (buffer == null)
-        allocate(defaultBufferSize)
-      else {
-        buffer.clear()
-        buffer
-      }
-    } else takeBufferFromPool() // spin while locked
+    // allocate new and clear outside the lock
+    if (buffer == null)
+      allocate(defaultBufferSize)
+    else {
+      buffer.clear()
+      buffer
+    }
+  }
 
-  @tailrec
   private final def offerBufferToPool(buf: ByteBuffer): Unit =
-    if (locked.compareAndSet(false, true))
-      try if (buffersInPool < maxPoolEntries) {
-        pool(buffersInPool) = buf
-        buffersInPool += 1
-      } // else let the buffer be gc'd
-      finally locked.set(false)
-    else offerBufferToPool(buf) // spin while locked
+    if (lock.tryLock(1, TimeUnit.MILLISECONDS))
+      try
+        if (buffersInPool < maxPoolEntries) {
+          pool(buffersInPool) = buf
+          buffersInPool += 1
+        } // else let the buffer be gc'd
+      finally lock.unlock()
 }


akka-actor/src/main/scala/akka/io/DirectByteBufferPool.scala
```diff
@@ -36,32 +36,32 @@ private[akka] class DirectByteBufferPool(defaultBufferSize: Int, maxPoolEntries:
...
-        buffersInPool += 1
-      } // else let the buffer be gc'd
-      finally locked.set(false)
-    else offerBufferToPool(buf) // spin while locked
+    if (lock.tryLock(1, TimeUnit.MILLISECONDS))
+      try
+        if (buffersInPool < maxPoolEntries) {
+          pool(buffersInPool) = buf
+          buffersInPool += 1
+        } // else let the buffer be gc'd
```
patriknw: here we can use `sun.misc.Cleaner` as pointed out in #17194, but that is for another PR

patriknw:
LGTM

akka-actor/src/main/scala/akka/io/DirectByteBufferPool.scala
rkuhn: Yes, that approach would have to be tried out properly and is not a 5min fix :-)

rkuhn:
I did a medium-deep audit of our current usage of `compareAndSet` (and Unsafe friends) and did not find any other offenders.

patriknw:
thanks, I have cherry-picked this for master

drewhk:
I have not reviewed this yet

akka-actor/src/main/scala/akka/io/DirectByteBufferPool.scala
```diff
@@ -36,32 +36,32 @@ private[akka] class DirectByteBufferPool(defaultBufferSize: Int, maxPoolEntries:
...
 
-  @tailrec
-  private final def takeBufferFromPool(): ByteBuffer =
-    if (locked.compareAndSet(false, true)) {
-      val buffer =
-        try if (buffersInPool > 0) {
+  private final def takeBufferFromPool(): ByteBuffer = {
+    var buffer: ByteBuffer = null
+
+    if (lock.tryLock(1, TimeUnit.MILLISECONDS))
```
drewhk: why no else on this block? The previous code kept retrying, so eventually it entered a critical section. Now we try to enter the critical section, and if we fail, we do something outside the critical section. Why is it safe?

jrudolph:
I just wanted to comment that the best-effort approach was suggested before but wasn't accepted. Now, if you change to a best-effort approach, why not just bound the number of spinning iterations?

akka-actor/src/main/scala/akka/io/DirectByteBufferPool.scala
patriknw: It's best effort reuse of the buffers, if we can't grab one within 1ms we create a new one. It's indeed different, but not wrong as far as I can see.

rkuhn: There’s no point in blocking this thread longer than 1ms for an optimization that we can do well without. If we can’t access the pool right now we’ll have to allocate a new buffer.

drewhk: Ok, I see.

drewhk:
I think this is a good question.

rkuhn:
@jrudolph Ah, I did not remember that. It basically comes down to how costly a DirectByteBuffer allocation is: is it worth potentially doing two context switches within a millisecond in order to recuperate one of them? My impression was that these particular objects are particularly nasty for the GC to cope with, which led me to this trade-off. Spinning for 100ns (or so) would express that DirectByteBuffers are actually quite cheap. Spinning for 1ms does not sound like a good idea. So, does anybody have some numbers? We don’t need to settle this now, this should be good enough for 2.3.10, but we can certainly open another ticket to properly revisit this.

patriknw:
I guess a thread that is spinning will quickly consume all its iterations. tryLock is probably doing something smarter?

rkuhn:
It will spin a few times (adaptively) and then properly park the thread—including the information on what it is blocked so that the kernel can make intelligent decisions about it.

patriknw:
However, I feel somewhat uncomfortable with this change without the `sun.misc.Cleaner`, since it can potentially result in a lot more buffers being allocated. A simple synchronized might be the safe way to go until we can do this together with the Cleaner.

drewhk:
It is also worth to mention that these explicit locks do not get all the JVM lock optimization stuff unlike synchronized blocks.

jrudolph:
I agree, it's hard to decide. In the end, any number, iterations or milliseconds, would be arbitrary since the behavior would depend on the expected contention (and as @rkuhn says the expected benefit). It probably cannot be intuitively answered which one is better without actually benchmarking it. My guess would have been that 1 millisecond is already way in the diminishing returns territory.

rkuhn:
I just tried it: I placed a `println` in the `else` branch, but I was unable to provoke it, with arbitrary other CPU hogging activities going on. It is extremely unlikely that this change will lead to additional DirectByteBuffers being allocated in sizable numbers.

Wrt. `synchronized`: in the pathological scheduling case that triggered this ticket this will completely inhibit all IO in the system while that one thread is holding the lock and not getting CPU cycles. Since this mechanism is an optimization I think it is not warranted to build it such that it can block the whole system.

patriknw:
@rkuhn thanks for checking, let's go with this one

drewhk:
OK

hepin1989:
http://blog.takipi.com/java-8-stampedlocks-vs-readwritelocks-and-synchronized/

akka-actor/src/main/scala/akka/io/DirectByteBufferPool.scala
viktorklang: Seems like there should be a ticket/todo to not use locks?

rkuhn: Why? As discussed extensively elsewhere on this PR we do _not_ want to just drop things to the floor if we cannot “immediately” get access to the pool, and a lock is the perfect and well-optimized way of expressing this concern. Removing locks is not self-justifying.

viktorklang: Since we aren't dealing with readers vs writers with this lock, there's nothing that says that the next thread to obtain the lock isn't another of the same type anyway. Correct me if I'm wrong, but to me this seems like a stop-gap solution rather than a design—which to me would imply a todo (which in itself is very cheap and adds to the documentation of the implementation).

rkuhn: Your reply leaves me entirely confused—may I suggest that you read the whole thing carefully and explain again what you mean? (We can also leave this until next Wednesday, that would be much preferred actually.)

viktorklang: Urgency was not intended nor implied. Let's talk at the end of next week.

jrudolph:
It seems this change led to a performance regression in 2.3.10: I had a quick check and the spray server-benchmark drops from 192k RPS to 72k RPS going from 2.3.9 to 2.3.10. The benchmark stresses akka io as much as possible with about 50% of CPU time spent in the TcpConnection actor and consequently the pool is much contented so this may not be representative of realistic loads.

drewhk:
Thanks @jrudolph! Can you check with a bounded-spin version of the original spinning code? (it it is not a lot of work)

jrudolph:
I tried and not getting the bounds right will kill the JVM with OOM: Direct buffer memory. I didn't implement the cleaner which would fix that but could introduce another performance regression. I think this needs some tinkering and benchmarking to get the details right.

Using synchronized instead of this lock makes the performance recover to ~ 170k RPS but if @rkuhn is correct it wouldn't fix the original issue.

drewhk:
> I tried and not getting the bounds right will kill the JVM with OOM: Direct buffer memory. I didn't implement the cleaner which would fix that but could introduce another performance regression. I think this needs some tinkering and benchmarking to get the details right.

Ok

> Using synchronized instead of this lock makes the performance recover to ~ 170k RPS but if @rkuhn is correct it wouldn't fix the original issue.

That was just a curiosity from my side because synchronized is heavily optimized in the JVM compared to explicit locks. I just wanted to see how much is the difference, not really related to this PR.

rkuhn:
Thanks a lot for the quick testing, @jrudolph! Viktor and I talked about this an hour ago and we have an idea of how to implement a completely wait-free pool. The proper solution should also include the Cleaner so it will take a bit more time than we have today—let’s fix it next week.

rkuhn:
#17281 created

PR #17262, opened 2015-04-22
=act #17216 fix DirectByteBufferPoll locking (for validation)

patriknw:
instead of spinning unboundedly use ReentrantLock.tryLock—it is a best
effort pool anyway and contention shall not lock the whole application

(cherry picked from commit 518fedb33c22c69deae019090d4236c9c5175fb5)


patriknw:
Refs #17216

PR #17266, opened 2015-04-22
+htc resurrect low-level server-side Java API and JavaTestServer

jrudolph:
Modelled after javadsl.StreamTcp.

Client-side and websocket support will be built once the respectives PRs are merged.

/cc @sirthias


akka-http-core/src/main/scala/akka/http/model/japi/Http.scala
```diff
@@ -0,0 +1,167 @@
+/*
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.model.japi
```
rkuhn: wrong package: shouldn’t this be in `akka.http.server.javadsl`?

jrudolph: We don't have the packages migrated, have you seen our proposal how to name things in #17279 ? We wanted to wait until things are merged before we introduce an all-conflicting renaming.

jrudolph: So, this just reinstantiates the previous status-quo. But otherwise you are of course right, that the `Http` class should not live in the model.

rkuhn: Ah, thanks for the background.

PR #17267, opened 2015-04-22
Fix for broken poll() function

mplatvoet:
The pollNode() function has issues that leads to missing and duplicate values. I've tried to explain the issues,with a proposed fix, in comments in the code.

Note: Proposed the fixes from the raw github editor _without actually building_ it.

### DIFF
--- a/akka-actor/src/main/java/akka/dispatch/AbstractBoundedNodeQueue.java
+++ b/akka-actor/src/main/java/akka/dispatch/AbstractBoundedNodeQueue.java
@@ -167,9 +167,17 @@ public abstract class AbstractBoundedNodeQueue<T> {
             final Node<T> deq = getDeq();
             final Node<T> next = deq.next();
             if (next != null) {
+                //Can't rely on next.value after casDeq because 'next' can be popped by another thread already
+                //before this one finishes. 
+                //Thus reading and storing the value before casDeq is needed. 
+                T nextVal = next.value;
                 if (casDeq(deq, next)) {
-                    deq.value = next.value;
-                    next.value = null;
+                    deq.value = nextVal;
+                    
+                    //Only set this to null if next still contains the expected
+                    //value. If not, "deq.value = nextVal" has already happened on a different thread.
+                    //and we don't want to override that since that's another value popping.
+                    next.casNullValue(nextVal);
                     return deq;
                 } // else we retry (concurrent consumers)
             } else if (getEnq() == deq) return null; // If we got a null and head meets tail, we are empty
@@ -188,7 +196,9 @@ public abstract class AbstractBoundedNodeQueue<T> {
     }
 
     public static class Node<T> {
-        protected T value;
+        //must be volatile, writes from other threads otherwise can be missed since values
+        //can allready be cached by the current thread. Effectively returning a stale/different value.
+        protected volatile T value;
         @SuppressWarnings("unused")
         private volatile Node<T> _nextDoNotCallMeDirectly;
         protected int count;
@@ -202,11 +212,17 @@ public abstract class AbstractBoundedNodeQueue<T> {
           Unsafe.instance.putOrderedObject(this, nextOffset, newNext);
         }
         
+        protected void casNullValue(T expected) {
+           Unsafe.instance.compareAndSwapObject(this, valueOffset, expected, null); 
+        }
+        
         private final static long nextOffset;
+        private final static long valueOffset;
         
         static {
             try {
                 nextOffset = Unsafe.instance.objectFieldOffset(Node.class.getDeclaredField("_nextDoNotCallMeDirectly"));
+                valueOffset = Unsafe.instance.objectFieldOffset(Node.class.getDeclaredField("value"));
             } catch(Throwable t){
                 throw new ExceptionInInitializerError(t);
             } 


typesafehub-validator:
Hi @mplatvoet,

drewhk:
OK TO TEST

patriknw:
Is there [an issue](https://github.com/akka/akka/issues) for this?

mplatvoet:
Probably not, just came across this when skimming through the code

akka-actor/src/main/java/akka/dispatch/AbstractBoundedNodeQueue.java
```diff
@@ -167,9 +167,17 @@ public final T poll() {
             final Node<T> deq = getDeq();
             final Node<T> next = deq.next();
             if (next != null) {
+                //Can't rely on next.value after casDeq because 'next' can be popped by another thread already
+                //before this one finishes.
```
rkuhn: This is not true: only one thread can win the CAS operation, and that one then ownes `next`. The loser goes back to the outer loop and reads the new `next` pointer from the new `deq` that it obtains (because we just shifted it).

rkuhn: Ah, I needed to write this down plus an external interruption to understand you properly, forget my objection, continuing to think.

akka-actor/src/main/java/akka/dispatch/AbstractBoundedNodeQueue.java
```diff
@@ -167,9 +167,17 @@ public final T poll() {
...
+                T nextVal = next.value;
                 if (casDeq(deq, next)) {
-                    deq.value = next.value;
-                    next.value = null;
+                    deq.value = nextVal;
+                    
+                    //Only set this to null if next still contains the expected
+                    //value. If not, "deq.value = nextVal" has already happened on a different thread.
+                    //and we don't want to override that since that's another value popping.
+                    next.casNullValue(nextVal);
```
rkuhn: This does not work since the queue permits the same object to be enqueued multiple times, which means that the CAS would succeed when it shouldn’t.

rkuhn:
Thanks for finding this, @mplatvoet, thinking about this some more I come to two conclusions:

* we cannot drop the reference to the dequeued element with the current scheme of doing things; a possible solution would be to CAS a new Node into the dequeue position with a `null` value (cost: 1 allocation)

* this queue is not actually lock-free, it contains a hidden-coded spinlock in the enqueue case

It should be noted, though, that while this is not a suitable MPMC queue, it is a perfectly fine MPSC queue (apart from the spinlock), so it does work for Actors unless used with a BalancingDispatcher.

akka-actor/src/main/java/akka/dispatch/AbstractBoundedNodeQueue.java
mplatvoet: You're right, didn't think of that. Though you're probably better off with the CAS compared to bluntly setting `next.value` to `null` since that possibly always incorrectly overwrites values.

This simply means that `next.value` can't be set to `null` in multiple consumer scenarios (without introducing more concurrency controls). The obvious other option is to accept a slight (temporary?) memory leak on the last node. Which actually doesn't sound that bad for a queue that is expected to move work around during it's lifetime.

viktorklang: Leaking at the last node is a huge no-go, imagine many actors with infrequent messages and large messages on top of that.

viktorklang:
I think it was originally documented to be an MPSC queue, I think the appropriate fix is to reinstate that documentation-wise.

drewhk:
This should be closed then?

viktorklang:
IMO, yes. But would be interested in hearing what others think.

mplatvoet:
Documenting this is a single consumer queue sounds like the most reasonable solution since I was let to believe it was a MCMP queue just by looking at docs. And turning this into a MCMP queue requires way more guards and thus sacrifices performance.

viktorklang:
@mplatvoet Well said!

mplatvoet:
One thing that did sprang to mind though was that if this is truly used and intended as a SCMP queue there is no need for CASsing on the deque altogether (and neither is looping). Since there is only one thread updating the field. The thread can't race itself. 

Don't do the next part, but: 
If it can be guaranteed that during the lifetime of this AbstractBoundedNodeQueue the same thread (identity equals) is doing the querying (so polling, peeking, size, etc) the deque field doesn't even need to be volatile. But if other threads also query (like peeking and size) don't do this because values can be missed and that is the argument to keep it volatile.

viktorklang:
Hi Mark,

yes, other threads will peek and size() it but not take elements from it.

On Mon, May 25, 2015 at 4:13 PM, Mark Platvoet <notifications@github.com>
wrote:

> One thing that did sprang to mind though was that if this is truly used
> and intended as a SCMP queue there is no need for CASsing on the deque
> altogether (and neither is looping). Since there is only one thread
> updating the field. The thread can't race itself.
>
> Don't do the next part, but:
> If it can be guaranteed that during the lifetime of this
> AbstractBoundedNodeQueue the same thread (identity equals) is doing the
> querying (so polling, peeking, size, etc) the deque field doesn't even need
> to be volatile. But if other threads also query (like peeking and size)
> don't do this because values can be missed and that is the argument to keep
> it volatile.
>
> —
> <https://github.com/akka/akka/pull/17267#issuecomment-105240527>.
>



-- 
Cheers,
√

drewhk:
Please reopen a ticket or something if you think this needs more discussion, but I am closing this PR for now.

Issue #16940, opened 2015-02-26
Avoid unnecessary copying in StreamLayout


drewhk:
Currently the only reason we need to copy is because we return the full, flattened downstreams and upstreams Maps. Since internal connections of an imported composite module are, well, internal, they are not necessary to copy -- the materializer can disambiguate while traversing. Exposed ports of the imported graphs still needs to be copied, and a mapping should be maintained.

PR #17280, opened 2015-04-23
=str #16940: Implement shallow copy for stream layouts

drewhk:


drewhk:
Preview for the shallow copy in layout. One test still failing and validation needs to be updated.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -230,63 +237,46 @@ private[akka] object StreamLayout {
...
-    shape: Shape,
-    override val connections: List[(OutPort, InPort)],
-    override val materializedValueComputation: MaterializedValueNode,
-    attributes: OperationAttributes) extends Module {
+  final case class CopiedModule(shape: Shape, attributes: OperationAttributes, copyOf: Module) extends Module {
+    override val subModules: Set[Module] = Set(copyOf)
+
+    override def withAttributes(attr: OperationAttributes): Module = this.copy(attributes = attr)
+
+    override def carbonCopy: Module = this.copy(shape = shape.deepCopy())
```
drewhk: If I lift this up we can get rid of copy methods on atomic modules. Since I was not sure of all the consequences it might have I left those as they are now. We can revisit it later.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -65,7 +65,7 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
    * flow into the materialized value of the resulting Flow.
    */
   def viaMat[T, Mat2, Mat3](flow: Flow[Out, T, Mat2])(combine: (Mat, Mat2) ⇒ Mat3): Flow[In, T, Mat3] = {
-    if (this.isIdentity) flow.asInstanceOf[Flow[In, T, Mat3]]
+    if (this.isIdentity) flow.asInstanceOf[Flow[In, T, Mat2]].mapMaterialized(combine(().asInstanceOf[Mat], _))
```
drewhk: This was the bug that caused only one test failure. Now added proper tests.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphMatValueSpec.scala
```diff
@@ -60,7 +60,7 @@ class GraphMatValueSpec extends AkkaSpec {
     }
 
     // Exposes the materialized value as a stream value
-    val foldFeedbackSource: Source[Future[Int], Future[Int]] = Source(foldSink) { implicit b ⇒
+    lazy val foldFeedbackSource: Source[Future[Int], Future[Int]] = Source(foldSink) { implicit b ⇒
```
drewhk: ah sorry, this is just the remainings of some debugging.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphConcatSpec.scala
```diff
@@ -154,6 +155,28 @@ class GraphConcatSpec extends TwoStreamsSetup {
...
       subscriber.expectError(TestException)
     }
+
+    "work with Source DSL" in {
+      val testSource = Source(1 to 5).concat(Source(6 to 10)).grouped(1000)
+      Await.result(testSource.runWith(Sink.head), 3.seconds) should ===(1 to 10)
+
+      val runnable = testSource.toMat(Sink.ignore)(Keep.left)
+      val (m1, m2) = runnable.run()
+
```
drewhk: I checked that m1 m2 are indeed Unit, which somehow I removed.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: I have no clue why it worked before though.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -468,8 +465,47 @@ private[stream] class MaterializedValuePublisher extends Publisher[Any] {
...
+  private def enterScope(enclosing: CopiedModule): Unit = {
+    subscribersStack ::= mutable.Map.empty.withDefaultValue(null)
+    publishersStack ::= mutable.Map.empty.withDefaultValue(null)
+    moduleStack ::= enclosing.copyOf
+  }
+
+  // Exits the scope of the copied module and propagates Publishers/Subscribers to the enclosing scope assigning
+  // them to the copied ports instead of the original ones (since there might be multiple copies of the same module
+  // leading to port identity collisions)
+  private def exitScope(enclosing: CopiedModule): Unit = {
```
rkuhn: please remove this argument and use `val enclosing = moduleStack.head` instead

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -524,15 +564,21 @@ private[stream] abstract class MaterializerSession(val topLevel: StreamLayout.Mo
   }
 
   final protected def assignPort(in: InPort, subscriber: Subscriber[Any]): Unit = {
-    subscribers.put(in, subscriber)
-    val publisher = publishers(topLevel.upstreams(in))
-    if (publisher ne null) attach(publisher, subscriber)
+    subscribers.update(in, subscriber)
```
rkuhn: might also use `subscribers(in) = subscriber` (same thing)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: The case looks nasty but is correct … ugh

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: Nope, it does not work that way

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: yep, true

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: see line 483, we don't save the wrapper on the stack but we save the wrapped module. Btw the only reason for that to be stateful so that I don't need to thread every current state through the subclasses (like ActorMaterializerImpl)

rkuhn: Why? The three stacks are always “in sync”, so when popping the subscribers and publishers you also pop the corresponding layout. Or is there another thing going on?

drewhk: Yes, the CopiedModule does not have flattened donwstreams+upstreams maps to isolate the internals. Normal composites do have those flattened maps. Therefore I save on the stack the wrapped module (which is needed at many places). I don't save the enclosing module because it does not need to be stored as state. Plus, this way I can set the stack to contain an initial value of topModule instead of creating a dummy CopiedModule (if the type of the stack would be CopiedModule).

rkuhn: ah, I see, that subtlety escaped me (so the purpose of asking this question has been fulfilled ;-) )

drewhk: Maybe there is a better way, but I was happy to get this working :)

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: Well, it’s a trade-off—we could just store the CopiedModule on the stack and adapt `currentLayout` to call `copyOf`—but I think this is okay, it saves some pointer chasing. Please add a comment to that stack that it stores the copies’ originals.

rkuhn:
Nicely done! After the last round of lipstick we can merge this one.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: (Wrong comment I guess) So I tried that road as well, but since the stack needs an initial entry I needed to make an unnecessary copy of topLevel to make it the right type.

drewhk: Interesting, Github showed your comment as a reply to a different place. Now it is at the right place.

drewhk:
Fixed tests and added comment explaining the scope stack

patriknw:
LGTM

PR #17283, opened 2015-04-23
log materialization errors in `bindAndHandle`

sirthias:
Closes #16507.

/cc @jrudolph

PR #17287, opened 2015-04-24
Improve error message if serverLayer is reused.

### LINKED ISSUE #17129: http server blueprint is reused but cannot be

jrudolph:
Fixes #17129 for now.

Based on @sirthias' #17283.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/Http.scala
@@ -120,12 +120,14 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
   type ServerLayer = BidiFlow[HttpResponse, ByteString, ByteString, HttpRequest, Unit]
 
   /**
-   * Constructs a [[ServerLayer]] stage using the configured default [[ServerSettings]].
+   * Constructs a [[ServerLayer]] stage using the configured default [[ServerSettings]]. The returned [[BidiFlow]]
+   * can only be materialized once.
    */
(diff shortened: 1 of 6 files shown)


jrudolph:
Added a commit which fixes a small issue in the Websocket implementation.

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
```diff
@@ -53,7 +53,7 @@ private[akka] case object CancelledSubscription extends Subscription {
 /**
  * INTERNAL API
  */
-private[akka] case object RejectAdditionalSubscibers extends Publisher[Nothing] {
+private[akka] case object RejectAdditionalSubscribers extends Publisher[Nothing] {
```
patriknw: thanks

patriknw:
LGTM

jrudolph:
Failed in `akka.stream.io.TlsSpec`.

jrudolph:
Rebased.

PR #17289, opened 2015-04-24
refactor SslTlsActor and stop it reliably

rkuhn:
- this also discovered two omissions in the Transfer infrastructure:
  - inputsAvailableFor should be completed when the input is cancelled
  - demandAvailableFor should be completed when the output is errored

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
@@ -186,7 +186,7 @@ private[akka] object FanIn {
     }
 
     def inputsAvailableFor(id: Int) = new TransferState {
-      override def isCompleted: Boolean = depleted(id)
+      override def isCompleted: Boolean = depleted(id) || cancelled(id)
       override def isReady: Boolean = pending(id)
     }
 
--- a/akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
@@ -69,8 +69,8 @@ private[akka] class SslTlsCipherActor(settings: ActorFlowMaterializerSettings, s
    * block.
    */
   class ChoppingBlock(idx: Int, name: String) extends TransferState {
-    override def isReady: Boolean = buffer.nonEmpty
-    override def isCompleted: Boolean = false
+    override def isReady: Boolean = buffer.nonEmpty || inputBunch.isPending(idx) || inputBunch.isDepleted(idx)
+    override def isCompleted: Boolean = inputBunch.isCancelled(idx)
 
     private var buffer = ByteString.empty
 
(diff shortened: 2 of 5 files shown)


akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
```diff
@@ -186,7 +186,7 @@ private[akka] object FanIn {
     }
 
     def inputsAvailableFor(id: Int) = new TransferState {
-      override def isCompleted: Boolean = depleted(id)
+      override def isCompleted: Boolean = depleted(id) || cancelled(id)
```
drewhk: hm, but it is the downstream that is cancelled. Isn't this the wrong place to handle this?

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -69,8 +69,8 @@ private[akka] class SslTlsCipherActor(settings: ActorFlowMaterializerSettings, s
    * block.
    */
   class ChoppingBlock(idx: Int, name: String) extends TransferState {
-    override def isReady: Boolean = buffer.nonEmpty
-    override def isCompleted: Boolean = false
+    override def isReady: Boolean = buffer.nonEmpty || inputBunch.isPending(idx) || inputBunch.isDepleted(idx)
+    override def isCompleted: Boolean = inputBunch.isCancelled(idx)
```
drewhk: so my spidey senses were right :)

rkuhn: Yes, and I’m saying this on the record: you were exactly right :-) The part that this handles gracefully is when termination happens “unexpectedly”, as is the case for `fail(e)`.

akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
rkuhn: No, when _we_ cancelled this input, this input cannot possibly contribute to our success anymore, hence `isCompleted`.

drewhk: Ah, ok. Isn't it enough to flip isDepleted on cancellations?

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -203,38 +203,76 @@ private[akka] class SslTlsCipherActor(settings: ActorFlowMaterializerSettings, s
...
+  }
+
+  val inboundClosed = TransferPhase(outbound || inboundHalfClosed) { () ⇒
+    if (tracing) log.debug("inboundClosed")
+    val continue = doInbound(isOutboundClosed = false, inboundHalfClosed)
+    if (continue) {
+      if (tracing) log.debug("inboundClosed continue")
+      doOutbound(isInboundClosed = true)
+    }
+  }
```
drewhk: moving this large chunk messes with diff view. Were there any changes here?

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -203,38 +203,76 @@ private[akka] class SslTlsCipherActor(settings: ActorFlowMaterializerSettings, s
...
     def isReady = lastHandshakeStatus == NEED_WRAP
-    def isCompleted = false
+    def isCompleted = engine.isOutboundDone
   }
 
   val engineInboundOpen = new TransferState {
-    def isReady = !engine.isInboundDone()
-    def isCompleted = false
+    def isReady = true
+    def isCompleted = engine.isInboundDone
```
drewhk: yes, this is the correct condition since inboundDone implies it cannot be ready again. (and no need to flip isReady anymore since isCompleted overrides it)

drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
rkuhn: no, I just moved it up so that I have all the run conditions on one screen

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -435,7 +433,12 @@ private[akka] class SslTlsCipherActor(settings: ActorFlowMaterializerSettings, s
     inputBunch.cancel()
     outputBunch.error(TransportOut, e)
     outputBunch.error(UserOut, e)
-    context.stop(self)
+    log.debug("{}\n{}", inputBunch, outputBunch)
```
rkuhn: will remove

akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
rkuhn: No, because there is value in distinguishing whether we cancelled it or the outside completed (think inputsOrCompleteAvailableFor)

drewhk: Ok, that's true

akka-stream/src/main/scala/akka/stream/impl/io/SslTls.scala
```diff
@@ -435,7 +433,12 @@ private[akka] class SslTlsCipherActor(settings: ActorFlowMaterializerSettings, s
...
+    pump()
+  }
+
+  override def postStop(): Unit = {
+    if (tracing) log.debug("postStop")
```
patriknw: nothing in super.postStop?

patriknw:
LGTM

Issue #17290, opened 2015-04-24
preStart / postStop for AbstractStage / ActorInterpreter


ktoso:
Stages need `preStart` and `postStop` similarly as Actors do.
We can also remove initAsyncInput probably once this is in.

I have this WIP but after latest rebases many test failures came up.

PR #17295, opened 2015-04-24
WIP: preStart / postStop for stages

ktoso:
This is the ripped out state of preStart / postStop from the log PR.

Aims to resolve: https://github.com/akka/akka/issues/17290


ktoso:
I'm resuming work on this guy to unblock other tickets that need it.

PR #17300, opened 2015-04-24
+htc #16887 the Java side of the Websocket server-side API

jrudolph:
I already used the new package naming convention but this will still conflict with the big renaming. It can be reviewed now and then I'll rebase once the package-renaming PR has been merged.

/cc @sirthias

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/model/ws/Message.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/model/ws/Message.scala
@@ -1,0 +1,145 @@
+/*
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.javadsl.model.ws
+
+import akka.http.model.ws
+import akka.stream.javadsl.Source
+import akka.util.ByteString
+
+/**
+ * Represents a Websocket message. A message can either be a binary message or a text message.
+ */
+sealed abstract class Message {
+  /**
+   * Is this message a text message? If true, [[asTextMessage]] will return this
+   * text message, if false, [[asBinaryMessage]] will return this binary message.
+   */
+  def isText: Boolean
+
+  /**
+   * Returns this TextMessage if it is a text message, throws otherwise.
+   */
+  def asTextMessage: TextMessage
+
+  /**
+   * Returns this BinaryMessage if it is a binary message, throws otherwise.
+   */
+  def asBinaryMessage: BinaryMessage
+
+  def asScala: ws.Message
+}
+
+object Message {
+  def adapt(msg: ws.Message): Message = msg match {
+    case t: ws.TextMessage   ⇒ TextMessage.adapt(t)
+    case b: ws.BinaryMessage ⇒ BinaryMessage.adapt(b)
+  }
+}
+
+/**
+ * Represents a Websocket text message. A text message can either be strict in which case
+ * the complete data is already available or it can be streamed in which case [[getStreamedText]]
+ * will return a Source streaming the data as it comes in.
+ */
+abstract class TextMessage extends Message {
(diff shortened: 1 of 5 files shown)


akka-http-core/src/main/scala/akka/http/javadsl/model/ws/Message.scala
```diff
@@ -0,0 +1,145 @@
...
+   */
+  def getStrictData: ByteString
+
+  /**
+   * Returns a source of the binary message data.
+   */
+  def getStreamedData: Source[ByteString, _]
+
+  def isText: Boolean = false
+  def asTextMessage: TextMessage = throw new ClassCastException("This message is not a binary message.")
```
patriknw: is this message correct?

jrudolph: No, thanks. The corresponding one is wrong as well.

patriknw:
LGTM

jrudolph:
Fixed the message and the compile error.

akka-http-core/src/main/scala/akka/http/javadsl/model/ws/UpgradeToWebsocket.scala
```diff
@@ -0,0 +1,38 @@
...
+   *
+   * See http://tools.ietf.org/html/rfc6455#section-1.9
+   */
+  def getRequestedProtocols(): JIterable[String]
+
+  /**
+   * Returns a response that can be used to answer a Websocket handshake request. The connection will afterwards
+   * use the given handlerFlow to handle Websocket messages from the client.
+   */
+  def handleMessagesWith(handlerFlow: Flow[Message, Message, _], materializer: FlowMaterializer): HttpResponse
```
drewhk: not for this PR, but in the future we can accept GraphShape[Flow[Message, Message], _] since the DSLs now accept it in place of a Flow, and flow is actually a subclass of that graphshape. I will open a ticket to review these.

akka-http-core/src/main/scala/akka/http/model/ws/UpgradeToWebsocket.scala
```diff
@@ -32,4 +37,15 @@ trait UpgradeToWebsocket extends HttpHeader {
...
+  import scala.collection.JavaConverters._
+  def getRequestedProtocols(): Iterable[String] = requestedProtocols.asJava
+  def handleMessagesWith(handlerFlow: stream.javadsl.Flow[javadsl.model.ws.Message, javadsl.model.ws.Message, _], materializer: FlowMaterializer): HttpResponse =
+    handleMessages(adaptJavaFlow(handlerFlow))(materializer)
+
+  def handleMessagesWith(handlerFlow: stream.javadsl.Flow[javadsl.model.ws.Message, javadsl.model.ws.Message, _], subprotocol: String, materializer: FlowMaterializer): HttpResponse =
+    handleMessages(adaptJavaFlow(handlerFlow), subprotocol = Some(subprotocol))(materializer)
+
+  private[this] def adaptJavaFlow(handlerFlow: stream.javadsl.Flow[javadsl.model.ws.Message, javadsl.model.ws.Message, _]): Flow[Message, Message, Any] =
+    Flow[Message].map(javadsl.model.ws.Message.adapt).via(handlerFlow.asScala).map(_.asScala)
```
drewhk: the asScala is no longer needed because via accepts any Graph[FlowShape] so you can pass a scaladsl.Flow into the Java APIs and vice versa. You don't need to fix it now, just a heads up.

drewhk:
LGTM

jrudolph:
@drewhk good to know, thanks.

drewhk:
Created ticket: https://github.com/akka/akka/issues/17303

PR #17304, opened 2015-04-24
=str minor doc syntax correction

patriknw:


drewhk:
LGTM

akka-docs-dev/rst/java/stream-quickstart.rst
```diff
@@ -26,8 +26,7 @@ which will be responsible for materializing and running the streams we are about
 
 The :class:`ActorFlowMaterializer` can optionally take :class:`ActorFlowMaterializerSettings` which can be used to define
 materialization properties, such as default buffer sizes (see also :ref:`stream-buffers-java`), the dispatcher to
-be used by the pipeline etc. These can be overridden on a flow, source and sink, but this
-will be discussed in depth in :ref:`stream-section-configuration`.
+be used by the pipeline etc. These can be overridden on a flow, source and sink with :class:`OperationAttribute`.
```
ktoso: the dispatcher is defined on `ActorOperationAttributes`

patriknw: You are right. What about: These can be overridden `withAttributes` on a `Flow`, `Source` and `Sink`.

ktoso: Sounds OK, users will find what to put in there via scaladoc / types then.

patriknw: technically it is on `Graph` but the reader probably doesn't know about these things at this point in the documentation.

ktoso: right, mentioning all types one can apply this to here sounds good :)

Issue #17274, opened 2015-04-23
Make ForkJoinPool asyncMode configurable


richdougherty:
The [`asyncMode` argument in the `ForkJoinPool` constructor](https://github.com/scala/scala/blob/77437ffa521a6d1b073283624a722848a8c0b33c/src/forkjoin/scala/concurrent/forkjoin/ForkJoinPool.java#L2852-L2857) switches the pool between LIFO and FIFO scheduling. The JRE/Scala default is FIFO (`asyncMode = false`) and [Akka's default is LIFO](https://github.com/akka/akka/blob/00f6a58e7c1e4795ad8920745c01916cc26947ca/akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala#L379) (`asyncMode = true`).

I've [switched](https://github.com/playframework/playframework/pull/4278) to use FIFO scheduling in Play, but the value isn't configurable through Akka's configuration, so I needed to [implement](https://github.com/playframework/playframework/pull/4278/files) a custom configurator. It would be great if Akka itself made the `asyncMode` option configurable via a normal setting. E.g.

```
fork-join-executor {
  ...
  async-mode = true
  ...
}
```

drewhk:
Hmm, I didn't know this setting exists. It would be interesting to see how
streams behave performance wise when LIFO or FIFO is used or if they are
sensitive to it at all.

-Endre

hepin1989:
https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala#L379

richdougherty:
@drewhk, yes this setting may have an effect on stream behaviour. It would be good to make it configurable so we can experiment with different settings for different workloads.

patriknw:
sounds interesting

hepin1989:
in fact ``` asyncMode ? FIFO_QUEUE : LIFO_QUEUE``` so I think @richdougherty 's comment is wrong.
and akka's default is true so will be FIFO_QUEUE and JDK is false will be LIFO_QUEUE

drewhk:
yes, thanks for clarifying. would be interesting to see what happens when we change the value.

richdougherty:
[...]

PR #17315, opened 2015-04-27
+act #17274 make ForkJoinPool asyncMode configurable

He-Pin:
refs https://github.com/akka/akka/issues/17274

### DIFF
--- a/akka-actor/src/main/resources/reference.conf
+++ b/akka-actor/src/main/resources/reference.conf
@@ -276,6 +276,10 @@ akka {
 
         # Max number of threads to cap factor-based parallelism number to
         parallelism-max = 64
+
+        # Setting to "FIFO" to use queue like peeking mode which "poll" or "LIFO" to use stack
+        # like peeking mode which "pop".
+        task-peeking-mode = "FIFO"
       }
 
       # This will be used if you have set "executor = "thread-pool-executor""
--- a/akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
+++ b/akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
@@ -5,23 +5,22 @@
 package akka.dispatch
 
 import java.util.concurrent._
-import akka.event.Logging.{ Debug, Error, LogEventException }
-import akka.actor._
-import akka.dispatch.sysmsg._
-import akka.event.{ BusLogging, EventStream }
-import com.typesafe.config.{ ConfigFactory, Config }
-import akka.util.{ Unsafe, Index }
-import scala.annotation.tailrec
-import scala.concurrent.forkjoin.{ ForkJoinTask, ForkJoinPool, ForkJoinWorkerThread }
-import scala.concurrent.duration.Duration
-import scala.concurrent.ExecutionContext
-import scala.concurrent.ExecutionContextExecutor
-import scala.concurrent.duration.FiniteDuration
-import scala.util.control.NonFatal
-import scala.util.Try
 import java.{ util ⇒ ju }
 
-final case class Envelope private (val message: Any, val sender: ActorRef)
+import akka.actor._
+import akka.dispatch.sysmsg._
+import akka.event.EventStream
+import akka.event.Logging.{ Debug, Error, LogEventException }
+import akka.util.{ Index, Unsafe }
+import com.typesafe.config.Config
+
+import scala.annotation.tailrec
+import scala.concurrent.{ ExecutionContext, ExecutionContextExecutor }
+import scala.concurrent.duration.{ Duration, FiniteDuration }
+import scala.concurrent.forkjoin.{ ForkJoinPool, ForkJoinTask, ForkJoinWorkerThread }
+import scala.util.control.NonFatal
+
+final case class Envelope private (message: Any, sender: ActorRef)
 
 object Envelope {
   def apply(message: Any, sender: ActorRef, system: ActorSystem): Envelope = {
@@ -46,7 +45,8 @@ final case class TaskInvocation(eventStream: EventStream, runnable: Runnable, cl
 /**
  * INTERNAL API
  */
-private[akka] trait LoadMetrics { self: Executor ⇒
+private[akka] trait LoadMetrics {
+  self: Executor ⇒
   def atFullThrottle(): Boolean
 }
 
@@ -62,11 +62,15 @@ private[akka] object MessageDispatcher {
   // since this is a compile-time constant, scalac will elide code behind if (MessageDispatcher.debug) (RK checked with 2.9.1)
   final val debug = false // Deliberately without type ascription to make it a compile-time constant
   lazy val actors = new Index[MessageDispatcher, ActorRef](16, _ compareTo _)
+
   def printActors(): Unit =
     if (debug) {
       for {
         d ← actors.keys
-        a ← { println(d + " inhabitants: " + d.inhabitants); actors.valueIterator(d) }
+        a ← {
+          println(d + " inhabitants: " + d.inhabitants)
+          actors.valueIterator(d)
+        }
       } {
         val status = if (a.isTerminated) " (terminated)" else " (alive)"
         val messages = a match {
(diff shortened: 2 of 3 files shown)


ktoso:
OK TO TEST

akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
```diff
@@ -54,19 +54,25 @@ private[akka] trait LoadMetrics { self: Executor ⇒
...
-  final val debug = false // Deliberately without type ascription to make it a compile-time constant
+  final val debug = false
+  // Deliberately without type ascription to make it a compile-time constant
   lazy val actors = new Index[MessageDispatcher, ActorRef](16, _ compareTo _)
+
   def printActors(): Unit =
     if (debug) {
       for {
         d ← actors.keys
-        a ← { println(d + " inhabitants: " + d.inhabitants); actors.valueIterator(d) }
```
ktoso: And one less `;` in the akka codebase, good ;-)

akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
```diff
@@ -84,14 +90,15 @@ private[akka] object MessageDispatcher {
...
-  import MessageDispatcher._
   import AbstractMessageDispatcher.{ inhabitantsOffset, shutdownScheduleOffset }
+  import MessageDispatcher._
   import configurator.prerequisites
 
   val mailboxes = prerequisites.mailboxes
   val eventStream = prerequisites.eventStream
 
-  @volatile private[this] var _inhabitantsDoNotCallMeDirectly: Long = _ // DO NOT TOUCH!
+  @volatile private[this] var _inhabitantsDoNotCallMeDirectly: Long = _
```
ktoso: Would be nice to keep those comments in-line (multiple cases like this one)

hepin1989:
[error] (akka-actor/*:mimaReportBinaryIssues) akka-actor: Binary compatibility check failed!

akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
2beaucoup: I can still see it. ;)

ktoso: lol yeah - wishful thinking from my side

akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
```diff
@@ -422,9 +441,11 @@ class ForkJoinExecutorConfigurator(config: Config, prerequisites: DispatcherPrer
   }
 
   class ForkJoinExecutorServiceFactory(val threadFactory: ForkJoinPool.ForkJoinWorkerThreadFactory,
-                                       val parallelism: Int) extends ExecutorServiceFactory {
-    def createExecutorService: ExecutorService = new AkkaForkJoinPool(parallelism, threadFactory, MonitorableThreadFactory.doNothing)
+                                       val parallelism: Int,
+                                       val asyncMode: Boolean) extends ExecutorServiceFactory {
```
ktoso: That's the binary compat failure. Technically it's not akka-private so people may have used it.
We could add an additional constructor here to be binary compatible - with the default value as before-this-patch.

// cc @rkuhn

akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
2beaucoup: +1

hepin1989: @2beaucoup you can't see it now:)

2beaucoup: Yeah. But all the whitespace changes make reviewing hard. :(

hepin1989: sorry for that changes :(``

rkuhn: You know that you can add a query parameter '?w=1' to the diff view to suppress whitespace changes, right? The only downside is that for comments you'll have to use a pristine tab.

akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
```diff
@@ -60,13 +60,18 @@ private[akka] object MessageDispatcher {
 
   // dispatcher debugging helper using println (see below)
   // since this is a compile-time constant, scalac will elide code behind if (MessageDispatcher.debug) (RK checked with 2.9.1)
-  final val debug = false // Deliberately without type ascription to make it a compile-time constant
+  final val debug = false
+  // Deliberately without type ascription to make it a compile-time constant
```
patriknw: move the comment above the `val` if not on same line

akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
2beaucoup: Yeah, I'm aware of this workaround. :)

akka-actor/src/main/resources/reference.conf
```diff
@@ -276,6 +276,9 @@ akka {
 
         # Max number of threads to cap factor-based parallelism number to
         parallelism-max = 64
+
+        # Enable establishing local first-in-first-out scheduling mode for forked tasks that are never joined.
+        async-mode = on
```
patriknw: default is `false`, so we might want to keep this default `off` for 2.3 to not change behavior?

hepin1989: default is true

drewhk: The comment is a bit confusing. Akka users might have not used FJP therefore "forked tasks that are never joined" does not help. Also the tasks that Akka schedules are never joined anyway.

drewhk:
LGTM after binary compat resolved, and a better comment in the reference.conf

akka-actor/src/main/resources/reference.conf
hepin1989: so just FIFO or LIFO?I don't how to describe that in a proper way.could you please just comment here so I could copy and paste?

patriknw: I see, akka currently use `true`. Good

akka-actor/src/main/resources/reference.conf
```diff
@@ -276,6 +276,10 @@ akka {
 
         # Max number of threads to cap factor-based parallelism number to
         parallelism-max = 64
+
+        # Set to first-in-first-out scheduling mode when you set "async-mode = on" or
+        # last-in-first-out scheduling mode when you set "async-mode = off".
+        async-mode = on
```
rkuhn: I know that this comes from FJP, but it still reads awkwardly: “asynchrony” cannot really be switched on and off with this toggle. I’d prefer to be explicit and parse the strings `LIFO` and `FIFO` (explaining that one is a stack and the other a queue should be enough to give people an idea of what this does).

rkuhn:
Thanks @hepin1989!

patriknw:
LGTM

akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
```diff
@@ -432,12 +455,18 @@ class ForkJoinExecutorConfigurator(config: Config, prerequisites: DispatcherPrer
         m.withName(m.name + "-" + id)
       case other ⇒ other
     }
+    val asyncMode = config.getString("task-peeking-mode") match {
+      case "FIFO"      ⇒ true
+      case "LIFO"      ⇒ false
+      case unsupported ⇒ throw new IllegalStateException(s""""task-peeking-mode" could only set to "FIFO" or "LILO",but you provided:[$unsupported].""")
```
2beaucoup: this msg can be improved and lacks some whitespaces. ;)

patriknw: and it should be `IllegalArgumentException`

drewhk:
Do we want to have this in 2.3.11? This will also need a forward port to master.

patriknw:
It's ready and I vote fore including it in 2.3.11.

rkuhn:
Agreed.

hepin1989:
will rebase

patriknw:
why closed? you don't have to open a new PR for rebasing

hepin1989:
sorry,I force updated on another computer,but the code is not right,and when I merge the conflict,seems like I missed something.so I came to here.

patriknw:
ok, superseded by https://github.com/akka/akka/pull/17423

PR #17319, opened 2015-04-27
=act #17316 fix wrong setting in pinned-disp reference.conf

### LINKED ISSUE #17316: Seems like obsolete config property name `allow-core-pool-timeout`
In `akka-actor/src/main/resources/reference.conf`
<pre>
...
io {
  pinned-dispatcher {
    thread-pool-executor.allow-core-pool-timeout = off
  }
}
...
</pre>
<pre>
...
  thread-pool-executor {
    allow-core-timeout = on
  }
...
</pre>
I believe `allow-core-pool-timeout` should be renamed to `allow-core-timeout`.

ktoso:
Backport of https://github.com/akka/akka/pull/17318
Resolves #17316

### DIFF
--- a/akka-actor/src/main/resources/reference.conf
+++ b/akka-actor/src/main/resources/reference.conf
@@ -515,7 +515,7 @@ akka {
     pinned-dispatcher {
       type = "PinnedDispatcher"
       executor = "thread-pool-executor"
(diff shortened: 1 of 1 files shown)


patriknw:
LGTM, will change default behavior, but I think that is alright since it is a bug

drewhk:
LGTM

Issue #13711, opened 2013-11-08
move cluster tools in akka-contrib to separate modules


rkuhn:
imported from https://www.assembla.com/spaces/akka/tickets/3711

We need to determine how we want people to interact with the cluster. My intuition is that things like PubSub and Singleton are good candidates to consider for canonicalization.

jboner:
I like that structuring. I think that all high-level modules that build on top of akka-cluster should be in their own modules. There will be many more of these in the future.

PR #17321, opened 2015-04-27
!con #13711 move cluster tools and sharding to separate modules

patriknw:
This is "only" moving files, and renaming packages and config properties.


patriknw:
this is the next step to be able to progress with the cluster tools

project/AkkaBuild.scala
```diff
@@ -125,6 +125,19 @@ object AkkaBuild extends Build {
...
+    id = "akka-cluster-tools",
+    base = file("akka-cluster-tools"),
+    dependencies = Seq(cluster % "compile->compile;test->test;multi-jvm->multi-jvm")
+  ) configs (MultiJvm)
+  
+  lazy val clusterSharding = Project(
+    id = "akka-cluster-sharding",
+    base = file("akka-cluster-sharding"),
+    dependencies = Seq(cluster % "compile->compile;test->test;multi-jvm->multi-jvm", 
+        persistence % "compile;test->provided", clusterTools)
```
ktoso: why `test->provided`?

ktoso:
LGTM, *this is a small step for...* ;-)

project/AkkaBuild.scala
patriknw: That is a good question. It comes from the same definition in akka-contrib. I think it is that leveldb and commons-io are needed for the tests. I'll try it out to see if I understand.

patriknw: When I try other things it fails running tests with "missing or invalid dependency detected while loading class file 'LeveldbIdMapping.class'."

I checked the pom and it looks correct, nothing about level db or provided.

ktoso: okey then, thanks for verifying!

patriknw:
I added a section to the docs about the dependencies. I will squash before merge.

project/AkkaBuild.scala
patriknw: I would be happy to lear what this actually means in sbt terms. For me it reads "take the test dependencies in akka-persistence and make them provided here", but that is not what it is doing.

ktoso: I believe those are Ivy's "configurations", see here: http://ant.apache.org/ivy/history/latest-milestone/tutorial/conf.html 
and the `->`: http://ant.apache.org/ivy/history/latest-milestone/ivyfile/configurations.html

I do understand them the same way you do by the way.

rkuhn: My reading of such things is «the "test" config of this one depends on the "provided" configuration of that one», which seems consistent with the observations.

patriknw: thanks for the translation :)

akka-cluster-sharding/build.sbt
```diff
@@ -0,0 +1,18 @@
...
+
+AkkaBuild.defaultSettings
+
+Formatting.formatSettings
+
+Unidoc.scaladocSettingsNoVerificationOfDiagrams
+
+Unidoc.javadocSettings
+
+OSGi.clusterSharding
```
rkuhn: as a general remark: should we not inline these settings here instead of keeping them centralized?

akka-cluster-sharding/build.sbt
```diff
@@ -0,0 +1,18 @@
...
+
+MultiNode.multiJvmSettings
+
+libraryDependencies ++= Dependencies.clusterSharding
```
rkuhn: for these we might switch to declaring some libraries centrally and then using symbolic names in the sub-projects—this would keep the version information in one place while also keeping the dependency management local

akka-cluster-sharding/src/main/resources/reference.conf
```diff
@@ -0,0 +1,54 @@
...
+# This is the reference config file that contains all the default settings.
+# Make your edits/overrides in your application.conf.
+
+
+# //#sharding-ext-config
+# Settings for the ClusterShardingExtension
+akka.cluster.sharding {
+  # The extension creates a top level actor with this name in top level user scope,
+  # e.g. '/user/sharding'
+  guardian-name = sharding
```
rkuhn: empty lines after each setting make this more readable where no syntax highlighting is available

akka-cluster-sharding/build.sbt
patriknw: sounds good to me, same thing with dependencies

please create a separate issue

akka-cluster-sharding/src/main/resources/reference.conf
patriknw: ok

akka-cluster-sharding/build.sbt
patriknw: the libraries can be defined centrally, but the specific usage of them can be defined here

akka-cluster-sharding/build.sbt
rkuhn: done: https://github.com/akka/akka/issues/17451

rkuhn:
LGTM

patriknw:
Refs #13711

PR #17324, opened 2015-04-28
=str #17323 Source.actorRef now emits buffered els before completion

### LINKED ISSUE #17323: Possible bug: sending Status.Success message to an Source.actorRef can discard previous messages
I was experimenting with akka-streams 1.0-RC1 and was confused by the behavior of a sample program I wrote:

```
import akka.actor.ActorSystem
import akka.actor.Status
import akka.actor.actorRef2Scala
import akka.stream._
import akka.stream.OverflowStrategy
import akka.stream.scaladsl.Sink

object PossibleBug extends App {
  implicit val mat = ActorFlowMaterializer()
  
  val a = Source.actorRef(10, OverflowStrategy.fail).to(Sink.foreach(println)).run()
  a ! "abc"
  a ! "def"
  Thread.sleep(100)           // (1)
  a ! Status.Success("ok")    // (2)
  
  Console.readLine()
  
  system.shutdown()
}
```

The program as it is prints two messages: "abc" and "def"
With Thread.sleep (line (1)) commented out it doesn't print anything
With Status.Success (line (2)) commented out it prints two messages: "abc" and "def"

it looks like there is a race between consuming regular messages and Status.Success message.

ktoso:
Landed and pushed :-)
Resolves #17323 and also avoids a possible NPE which would happen if `bufferSize = 0` and demand is signalled - the condition in the while loop was not completely correct (did not guard from `buffer = null`).

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/ActorRefSourceActor.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/ActorRefSourceActor.scala
@@ -32,18 +32,19 @@ private[akka] class ActorRefSourceActor(bufferSize: Int, overflowStrategy: Overf
   def receive = {
     case _: Request ⇒
       // totalDemand is tracked by super
-      while (totalDemand > 0L && !buffer.isEmpty)
-        onNext(buffer.dequeue())
+      if (bufferSize != 0)
+        while (totalDemand > 0L && !buffer.isEmpty)
+          onNext(buffer.dequeue())
 
     case Cancel ⇒
       context.stop(self)
 
     case _: Status.Success ⇒
-      context.stop(self) // will complete the stream successfully
+      if (bufferSize == 0 || buffer.isEmpty) context.stop(self) // will complete the stream successfully
+      else context.become(drainBufferThenComplete)
 
     case Status.Failure(cause) if isActive ⇒
-      onError(cause)
-      context.stop(self)
+      onErrorThenStop(cause)
 
     case elem if isActive ⇒
       if (totalDemand > 0L)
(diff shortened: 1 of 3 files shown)


akka-stream/src/main/scala/akka/stream/impl/ActorRefSourceActor.scala
```diff
@@ -32,18 +32,19 @@ private[akka] class ActorRefSourceActor(bufferSize: Int, overflowStrategy: Overf
   def receive = {
     case _: Request ⇒
       // totalDemand is tracked by super
-      while (totalDemand > 0L && !buffer.isEmpty)
-        onNext(buffer.dequeue())
+      if (bufferSize != 0)
```
ktoso: otherwise NPE happens, added a test which shows this

akka-stream/src/main/scala/akka/stream/impl/ActorRefSourceActor.scala
```diff
@@ -63,12 +64,31 @@ private[akka] class ActorRefSourceActor(bufferSize: Int, overflowStrategy: Overf
...
+      context.stop(self)
+
+    case Status.Failure(cause) if isActive ⇒
+      // errors must be signalled as soon as possible,
+      // even if previously valid completion was requested via Status.Success
+      onErrorThenStop(cause)
+
+    case _: Request ⇒
+      // totalDemand is tracked by super
+      while (totalDemand > 0L && !buffer.isEmpty)
```
ktoso: // here we know buffer is not null

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -355,11 +353,14 @@ object Source extends SourceApply {
...
-   * [[akka.actor.Status.Success]] to the actor reference.
+   * The stream can be completed successfully by sending the actor referende an [[akka.actor.Status.Success]]
+   * messagein which case already buffered elements will be signalled before signalling completion,
+   * or by sending a [[akka.actor.PoisonPill]] in which case completion will be signalled immediatly.
    *
    * The stream can be completed with failure by sending [[akka.actor.Status.Failure]] to the
-   * actor reference.
+   * actor reference. In case the Actor is still draining its internal buffer (after having received
+   * an [[akka.actor.Status.Success]]) before signalling completion and it receives a [[akka.actor.Status.Failure]],
+   * the failure will be signalled downstream immediatly (instead of the completion signal).
```
ktoso: I guess we should stick to UK spelling or US spelling... We have a mix currently ("ll" is UK)

akka-stream/src/main/scala/akka/stream/impl/ActorRefSourceActor.scala
patriknw: good catch

patriknw:
LGTM

drewhk:
LGTM

ktoso:
Failure was SSL

PR #17332, opened 2015-04-28
StreamLayout debug mode

RichardBradley:
Is this "debug" val on intentionally?

I know you haven't officially addressed HTTP performance yet, but we have been doing some perf profiling of our Akka HTTP based app, and the "validate()" calls behind this "debug" flag are very expensive at runtime.

I can't find any warnings from the validate() calls in our logs, so it seems like this isn't needed at runtime.

Unless I'm missing something, this should be safe to turn off now.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
@@ -18,8 +18,10 @@ import akka.event.Logging.simpleName
  */
 private[akka] object StreamLayout {
 
-  // compile-time constant
-  final val Debug = true
+  // Turn this on to add run-time checks about Stream consistency during
+  // stream operations.
+  // (compile-time constant)
+  final val Debug = false
 
   // TODO: Materialization order
   // TODO: Special case linear composites


drewhk:
OK TO TEST

drewhk:
Hi Richard, we kept this option on for RC1 so we see detailed error logs if we messed up something before we release the final. I think we can shut it down in the next release. The purpose of this flag was to find inconsistencies in our internals and get detailed information about them. As you noticed it is quite costly to do this check, so this will be turned on. 

I guess we can merge this so it is already turned off in the next release.

RichardBradley:
Thanks, that makes sense.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -18,8 +18,10 @@ import akka.event.Logging.simpleName
  */
 private[akka] object StreamLayout {
 
-  // compile-time constant
-  final val Debug = true
+  // Turn this on to add run-time checks about Stream consistency during
```
drewhk: it is about consistency of the underlying data structure describing stream layouts (Stream consistency would mean something more general)

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -18,8 +18,10 @@ import akka.event.Logging.simpleName
...
+  // stream operations.
```
drewhk: it only validates during materialization, not during opertions (unless a stream operation does materialization of streams of course)

RichardBradley:
Thanks -- it sounds like my comment is quite a bit off the mark.
It's probably easiest to reject this PR and for the Akka team to turn it off when you're ready. 
I can use a forked version of Akka which includes this patch in the meantime.

Issue #17184, opened 2015-04-13
Replace errorLogger in HTTP by using built in log() stage


ktoso:
Since we now have built in logging, see https://github.com/akka/akka/pull/17162

It would now also be possible to replace the custom errorLogger HTTP uses, like this: https://github.com/ktoso/akka/commit/4fbb610b02c3b10c6afe76de541f53a8f7143ad0

Using `ZeroLevel` to avoid slowing down the normal pass-elements-through case would be advised ofc.

miciek:
@ktoso Is this still relevant? Where is the `errorLogger` used?

ktoso:
Good catch, AFAICS this can be closed now. 
We have an errorLogger stage but it's using the passed in logger correctly.

If you want to help out a bit there's a nice low hanging fruit in that area though:
The usage of transform+pushstage in: https://github.com/akka/akka/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala#L251
and https://github.com/akka/akka/blob/master/akka-http-core/src/main/scala/akka/http/impl/util/package.scala#L92 
should be rewritten to be a `GraphStage`, let me know if you'd like to fix that :-)

ktoso:
I made it a separate ticket https://github.com/akka/akka/issues/19899

PR #17335, opened 2015-04-29
#17322 Fix race condition during websocket connection establishment

### LINKED ISSUE #17322: Race between sending out Http Upgrade response and sending first websocket bytes
I.e if the user-supplied handler is immediately ready to send a message, this message is on the wire before the HTTP response. Hilarity ensues.

/cc @sirthias

jrudolph:
The problem was that the switch consists of two parts, triggering the protocol switch and sending out the HTTP response. Both were done asynchronously which lead to the race condition. The solution was to make the switching and sending out of the HTTP response an atomic operation signalled by the ResponseRenderer.

This needed some restructuring of the server pipeline which overall simplified the setup. The previous switch token source could be replaced by a loopback in the graph which directly feeds the switching signal from the response renderer to the protocol switch.

Fixes #17322.

/cc @sirthias


rkuhn:
LGTM (although I have not deeply understood the mechanism employed)—needs one more review to be merged.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -273,29 +278,11 @@ private[http] object HttpServerBluePrint {
       }
   }
 
-  case class WebsocketSetup(
-    flow: Flow[ByteString, ByteString, Any],
-    publisherKey: StreamUtils.ReadableCell[Publisher[FrameEvent]],
-    subscriberKey: StreamUtils.ReadableCell[Subscriber[FrameEvent]],
-    switchSource: Source[SwitchToWebsocketToken.type, Any]) extends WebsocketSwitch {
-    @volatile var switchToWebsocketRef: Option[ActorRef] = None
```
ktoso: nice to see this one gone

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -345,36 +331,44 @@ private[http] object HttpServerBluePrint {
...
-              ctx.emit(b); SameState
-            case SwitchToWebsocketToken ⇒
-              ctx.changeCompletionHandling(closeWhenInCloses(wsIn))
-              websockets
+            case ResponseRenderingOutput.HttpData(bytes) ⇒
+              ctx.emit(bytes); SameState
+            case ResponseRenderingOutput.SwitchToWebsocket(responseBytes, handlerFlow) ⇒
+              ctx.emit(responseBytes)
+              installHandler(handlerFlow)
+              websocket
```

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerTestSetupBase.scala
```diff
@@ -0,0 +1,70 @@
...
+    }.mkString("\n")
+
+  val netInSub = netIn.expectSubscription()
+  val netOutSub = netOut.expectSubscription()
+  val requestsSub = requests.expectSubscription()
+  val responsesSub = responses.expectSubscription()
+
+  def expectRequest: HttpRequest = {
+    requestsSub.request(1)
+    requests.expectNext()
```
drewhk: you can use requestNext to simplify these two lines.

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerTestSetupBase.scala
```diff
@@ -0,0 +1,70 @@
...
+  implicit def system: ActorSystem
+  implicit def materializer: FlowMaterializer
+
+  val requests = TestSubscriber.manualProbe[HttpRequest]
+  val responses = TestPublisher.manualProbe[HttpResponse]
+
+  def settings = ServerSettings(system).copy(serverHeader = Some(Server(List(ProductVersion("akka-http", "test")))))
+
+  val (netIn, netOut) = {
+    val netIn = TestPublisher.manualProbe[ByteString]
```
drewhk: you are recommended to use the normal (auto) probe instead of the manual, is usually powerful enough and simplifies life.

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
```diff
@@ -109,10 +98,10 @@ package object util {
...
   }
 
-  private[http] def errorLogger(log: LoggingAdapter, msg: String): PushStage[ByteString, ByteString] =
-    new PushStage[ByteString, ByteString] {
-      override def onPush(element: ByteString, ctx: Context[ByteString]): SyncDirective = ctx.push(element)
-      override def onUpstreamFailure(cause: Throwable, ctx: Context[ByteString]): TerminationDirective = {
+  private[http] def errorLogger[T](log: LoggingAdapter, msg: String): PushStage[T, T] =
+    new PushStage[T, T] {
+      override def onPush(element: T, ctx: Context[T]): SyncDirective = ctx.push(element)
+      override def onUpstreamFailure(cause: Throwable, ctx: Context[T]): TerminationDirective = {
```
ktoso: @jrudolph This should be replaced with `.log()` - but there's a separate ticket for it so no worries about it in this PR: https://github.com/akka/akka/issues/17184

ktoso:
LGTM, merging and releasing RC2 because we want to be on RS 1.0

ktoso:
The minor code cleanups can be done later on

PR #17370, opened 2015-05-04
Fix caching of HTTP Date header

gmethvin:
I noticed that the code to cache the HTTP date header in the `HttpResponseRenderer` is comparing milliseconds to seconds in `now - 1000 > cachedSeconds`. That would mean it updates the cached value every time the method is called. Also the method of overriding the date in tests only overrides the generated date string, not the date used for caching.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
@@ -37,18 +37,19 @@ private[http] class HttpResponseRendererFactory(serverHeader: Option[headers.Ser
 
   private def dateHeader: Array[Byte] = {
     var (cachedSeconds, cachedBytes) = cachedDateHeader
-    val now = System.currentTimeMillis
-    if (now - 1000 > cachedSeconds) {
+    val now = currentTimeMillis()
+    if (now / 1000 > cachedSeconds) {
       cachedSeconds = now / 1000
       val r = new ByteArrayRendering(48)
-      dateTime(now).renderRfc1123DateTimeString(r ~~ headers.Date) ~~ CrLf
+      DateTime(now).renderRfc1123DateTimeString(r ~~ headers.Date) ~~ CrLf
       cachedBytes = r.get
       cachedDateHeader = cachedSeconds -> cachedBytes
     }
     cachedBytes
   }
 
-  protected def dateTime(now: Long) = DateTime(now) // split out so we can stabilize by overriding in tests
+  // split out so we can stabilize by overriding in tests
+  protected def currentTimeMillis(): Long = System.currentTimeMillis()
 
   def newRenderer: HttpResponseRenderer = new HttpResponseRenderer
 
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/rendering/ResponseRendererSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/rendering/ResponseRendererSpec.scala
@@ -74,6 +74,23 @@ class ResponseRendererSpec extends FreeSpec with Matchers with BeforeAndAfterAll
             |"""
         }
       }
+      "a custom status code and no headers and different dates" in new TestSetup() {
+        val initial = DateTime(2011, 8, 25, 9, 10, 0).clicks
+        var extraMillis = 0L
+        (0 until 10000 by 500) foreach { millis ⇒
+          extraMillis = millis
+          HttpResponse(200) should renderTo {
+            s"""HTTP/1.1 200 OK
+              |Server: akka-http/1.0.0
+              |Date: Thu, 25 Aug 2011 09:10:0${extraMillis / 1000 % 60} GMT
+              |Content-Length: 0
+              |
+              |"""
+          }
+        }
+
+        override def currentTimeMillis() = initial + extraMillis
+      }
 
       "to a transparent HEAD request (Strict response entity)" in new TestSetup() {
         ResponseRenderingContext(
@@ -559,7 +576,7 @@ class ResponseRendererSpec extends FreeSpec with Matchers with BeforeAndAfterAll
         Await.result(future, 250.millis) -> renderer.isComplete
       }
 
-    override def dateTime(now: Long) = DateTime(2011, 8, 25, 9, 10, 29) // provide a stable date for testing
+    override def currentTimeMillis() = DateTime(2011, 8, 25, 9, 10, 29).clicks // provide a stable date for testing
   }
 
   def source[T](elems: T*) = Source(elems.toList)


akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
```diff
@@ -33,22 +33,23 @@ private[http] class HttpResponseRendererFactory(serverHeader: Option[headers.Ser
     }
 
   // as an optimization we cache the Date header of the last second here
-  @volatile private[this] var cachedDateHeader: (Long, Array[Byte]) = (0L, null)
-
+  private[this] var cachedDateHeader: (Long, Array[Byte]) = (0L, null)
   private def dateHeader: Array[Byte] = {
     var (cachedSeconds, cachedBytes) = cachedDateHeader
```
patriknw: does this allocate a tuple?

patriknw:
LGTM

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
gmethvin: I believe it does. It would probably be more efficient to do
```scala
var cachedSeconds = cachedDateHeader._1
var cachedBytes = cachedDateHeader._2
```
Unsure how much of a performance difference it'd make in this case though.

gmethvin:
@patriknw I updated the PR with a slightly more efficient implementation.

patriknw:
LGTM, thanks

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
```diff
@@ -33,22 +33,22 @@ private[http] class HttpResponseRendererFactory(serverHeader: Option[headers.Ser
     }
 
   // as an optimization we cache the Date header of the last second here
-  @volatile private[this] var cachedDateHeader: (Long, Array[Byte]) = (0L, null)
-
+  private[this] var cachedDateHeader: (Long, Array[Byte]) = (0L, null)
```
ktoso: are you sure removing this `@volatile` is safe?

ktoso: // cc @sirthias @jrudolph

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
```diff
@@ -33,22 +33,22 @@ private[http] class HttpResponseRendererFactory(serverHeader: Option[headers.Ser
...
   private def dateHeader: Array[Byte] = {
-    var (cachedSeconds, cachedBytes) = cachedDateHeader
-    val now = System.currentTimeMillis
```
sirthias: Wouldn't it suffice to simply append a  `/ 1000` here to fix the core bug?
All the other changes appear to be unnecessary from my current point of view.

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
```diff
@@ -33,22 +33,22 @@ private[http] class HttpResponseRendererFactory(serverHeader: Option[headers.Ser
...
-    if (now - 1000 > cachedSeconds) {
```
sirthias: Scrap my comment from above.
What is needed is replacing the `-` here with `/`.

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
```diff
@@ -33,22 +33,22 @@ private[http] class HttpResponseRendererFactory(serverHeader: Option[headers.Ser
...
+      cachedSeconds = nowSeconds
       val r = new ByteArrayRendering(48)
-      dateTime(now).renderRfc1123DateTimeString(r ~~ headers.Date) ~~ CrLf
-      cachedBytes = r.get
-      cachedDateHeader = cachedSeconds -> cachedBytes
+      DateTime(now).renderRfc1123DateTimeString(r ~~ headers.Date) ~~ CrLf
+      cachedDateHeader = cachedSeconds -> r.get
     }
-    cachedBytes
+    cachedDateHeader._2
```
sirthias: If you switch to reading the `cachedSeconds` and `cachedBytes` independently then we could also switch to writing them independently and get rid of the tuple altogether, no?

jrudolph: Actually, wouldn't splitting make the thing unsafe even with volatile? With the tuple and with or without volatile it seems to be safe because the tuple can always be treated as an atomic entity.

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
sirthias: I guess we could indeed get rid of the `volatile` here and simply allow for some fuzzyness.
Updates lost to the missing publishing would simply cause a bit more work on another core, which would hopefully be outweighed by removing the `volatile` overhead.
We might even split the `Long` and the bytes and remove the tuple.

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
sirthias: Yes, theoretically the `Long` and the bytes don't fully match up anymore.
But this shouldn't really matter.
It's just the date header, even if we render one that is 1 or 2 seconds off, no problem.

The question is whether it's a *requirement* that a single server only issues monotonically increasing Date stamps.
I'd say it isn't.

jrudolph: "Unsafe" because seconds and the rendered header wouldn't necessarily correspond to another any more. But this may not be a problem since every update to the header is locally still an improvement over the previously cached header version. So the worst thing that could happen, is a) that one thread still sees and uses the version of the last second because it hasn't seen the update of the timestamp yet or b) that it sees the stale timestamp and redundantly updates the header value.

jrudolph: Yes, right monotonicity may be a requirement. Is it safe to assume that all parts of the rendered `Array[Byte]` are visible at the same time?

sirthias: Yes. So this makes it possible that a server produces two requests A and B going out over the same connection to the same client in that sequence, where A has a `Date` header with a later timestamp than B.
I'm not sure whether this has the potential of breaking anything.

Having thought about it a bit more I'd probably vote for being on the safe side and **keeping the volatile tuple** rather than opening up a hole for a potentially extremely hard to debug problem in complicated enterprise deployments (which proxies, caches, and what-not).

sirthias: Which brings me back to my thought that the simple fix in (the old) line 41 is actually all we really need.

ktoso: Thanks for chiming in guys. Yes I'd much rather keep the monotonicity guarantee, god knows who might depend on such properties in api clients or whatnot.

jrudolph: +1 keeping the `@volatile` should also make sure that all parts of the `Array[Byte]` have been published

gmethvin:
Just read the above discussion. It seems like the only thing that needs to be fixed is changing the `-` to a `/`, so I updated the PR to do that, plus my other changes to make it testable.

sirthias:
Great, thanks a lot, Greg!
LGTM!

Issue #17351, opened 2015-04-30
Stage: onUpstreamFinish is (or may be?) called after previous ctx.pushAndFinish


jrudolph:
It's then even possible to inject more elements in onUpstreamFinish. I'd say this is a bug as explicitly finishing should cancel upstream which is racy in reactive streams semantics wrt receiving extra elements (or onComplete) but I don't find it natural in Stage semantics that such a race isn't prevented.

drewhk:
Once you completed the stage with pushAndFinish it should never result in a new onUpstreamFinish -- it is synchronous. In fact, calling pushAndFinish rewrites the internal stage pipeline to contain a special Finished op at the place of the original op. If you are right, then this is a bug, but currently it is hard to see how this can happen. Do you have a reproducer?

drewhk:
This might be related to forking in pushAndFinish.

jrudolph:
Ok, then we agree about the semantics. I'll try to come up with a reproduction.

drewhk:
I might know where the bug is. There is a FIXME even :) When an pushFinish is called and the next stage also calls pushFinish, then that op is not replaced with a Finished op, which is a bug.

drewhk:
but I would still like to see your reproducer because I am not sure that bug can cause what you see, unless you fuse stages, which as far as I know is not done.

drewhk:
I think this is the bug. Upstream boundary stage can call pushAndFinish, and if your code also calls it, then it will fail in the way you observed.

jrudolph:
Yep, this doesn't even seem to be a race condition.

I wrote this failing test in FlowStageSpec:

```scala
    "not trigger onUpstreamFinished after pushAndFinish" in assertAllStagesStopped {
      val in = TestPublisher.manualProbe[Int]()
      val flow =
        Source(in)
          .transform(() ⇒ new StatefulStage[Int, Int] {

            def initial: StageState[Int, Int] = new State {
              override def onPush(element: Int, ctx: Context[Int]) =
                ctx.pushAndFinish(element)
            }
            override def onUpstreamFinish(ctx: Context[Int]): TerminationDirective =
              terminationEmit(Iterator(42), ctx)
          })
          .runWith(Sink.publisher)

      val inSub = in.expectSubscription()

      val out = TestSubscriber.manualProbe[Int]()
      flow.subscribe(out)
      val outSub = out.expectSubscription()

      inSub.sendNext(23)
      inSub.sendComplete()

[...]

PR #17373, opened 2015-05-04
=str #17351: Fix pushAndFinish improper stage termination

drewhk:
The call to pushAndFinish did not properly finish the op that it was called upon, therefore not preventing an onPull from dowstream stages that came from the forked execution of pushAndFinish: first a push, then a finish is executed, but onPull could arrive first.

This fix adds a new PushFinished op type, that has the only difference from Finished that it propagates onUpstreamFinish. This is to ensure that the forked onUpstreamFinish from the original stage still take effect.

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
@@ -3,6 +3,8 @@
  */
 package akka.stream.impl.fusing
 
+import akka.stream.stage.{ TerminationDirective, SyncDirective, Context, PushStage }
+
 import scala.util.control.NoStackTrace
 import akka.stream.Supervision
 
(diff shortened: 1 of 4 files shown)


drewhk:
Fixes: https://github.com/akka/akka/issues/17351

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
```diff
@@ -460,6 +462,52 @@ class InterpreterSpec extends InterpreterSpecKit {
 
     }
 
+    // This test is related to issue #17351
```
drewhk: comment at wrong place...

drewhk:
Failure is unrelated (SslTls)

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
```diff
@@ -470,7 +518,31 @@ class InterpreterSpec extends InterpreterSpecKit {
...
+
+      downstream.requestOne()
+      lastEvents() should be(Set(RequestOne))
+
+      upstream.onNext("foo")
+      lastEvents() should be(Set(OnNext("foo"), OnComplete, Cancel))
+
+    }
+
+    "implement take-take wiht pushAndFinish from upstream" in new TestSetup(Seq(
```
rkuhn: `s/wiht/with/`

rkuhn:
LGTM; I have thought it through as far as I could and couldn’t find a flaw, but I have the feeling that I have not fully explored all corner cases.

patriknw:
LGTM

PR #17374, opened 2015-05-05
=pro check binary compatibility against all bc-compatible versions

2m:
Also verified that BC checks pass with 2.11.

Released a temporary version of mima sbt plugin to a separate bintray repo until official updated version is released.

### DIFF
--- a/project/AkkaBuild.scala
+++ b/project/AkkaBuild.scala
@@ -12,7 +12,7 @@ import com.typesafe.sbt.SbtScalariform
 import com.typesafe.sbt.SbtScalariform.ScalariformKeys
 import com.typesafe.sbt.osgi.SbtOsgi.{ OsgiKeys, defaultOsgiSettings }
 import com.typesafe.tools.mima.plugin.MimaPlugin.mimaDefaultSettings
-import com.typesafe.tools.mima.plugin.MimaKeys.previousArtifact
+import com.typesafe.tools.mima.plugin.MimaKeys.previousArtifacts
 import com.typesafe.tools.mima.plugin.MimaKeys.reportBinaryIssues
 import com.typesafe.tools.mima.plugin.MimaKeys.binaryIssueFilters
 import com.typesafe.sbt.SbtSite.site
@@ -50,7 +50,7 @@ object AkkaBuild extends Build {
     base = file("."),
     settings = parentSettings ++ Release.settings ++ Unidoc.settings ++ Publish.versionSettings ++
       SphinxSupport.settings ++ Dist.settings ++ s3Settings ++ mimaSettings ++ unidocScaladocSettings ++
-      StatsDMetrics.settings ++ 
+      StatsDMetrics.settings ++
       Protobuf.settings ++ inConfig(JavaDoc)(Defaults.configSettings) ++ Seq(
       testMailbox in GlobalScope := System.getProperty("akka.testMailbox", "false").toBoolean,
       parallelExecution in GlobalScope := System.getProperty("akka.parallelExecution", "false").toBoolean,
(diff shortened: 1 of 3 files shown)


patriknw:
LGTM, excellent

2m:
Failure was due to sbt version update. Same problem as fixed by the #16201. Instead of using the same fix, here we only upgrade to sbt 0.13.5 because we are still not using autoplugins.

project/AkkaBuild.scala
```diff
@@ -1105,33 +1105,49 @@ object AkkaBuild extends Build {
...
       val fullId = crossVersion.value match {
          case _ : CrossVersion.Binary => id + "_" + scalaBinaryVersion.value
          case _ : CrossVersion.Full => id + "_" + scalaVersion.value
-         case CrossVersion.Disabled => id 
+         case CrossVersion.Disabled => id
        }
-      Some(organization.value % fullId % version) // the artifact to compare binary compatibility with
+      // check against all binary compatible artifacts
+      val artifacts = for (version <- first to last)
+        yield organization.value % fullId % s"2.3.$version"
```
rkuhn: wouldn’t `%%` work here as well?

project/AkkaBuild.scala
```diff
@@ -1105,33 +1105,49 @@ object AkkaBuild extends Build {
...
   )
 
-  def akkaPreviousArtifact(id: String): Def.Initialize[Option[sbt.ModuleID]] = Def.setting {
+  def akkaPreviousArtifacts(id: String): Def.Initialize[Set[sbt.ModuleID]] = Def.setting {
     if (enableMiMa) {
       // Note: This is a little gross because we don't have a 2.3.0 release on Scala 2.11.x
       // This should be expanded if there are more deviations.
-      val version: String = 
+      val (first, last) = {
+        val lastAkka = 10
```
rkuhn: could this just be extracted from the current version by subtracting one?

rkuhn: Ah, that would require us to use `2.3.12-SNAPSHOT` instead of `2.3-SNAPSHOT`.

rkuhn:
LGTM! We should wrap this up as a first step and refine the filters to apply only to specific version ranges later.

project/AkkaBuild.scala
patriknw: it's 11 now!

project/plugins.sbt
```diff
@@ -11,7 +11,9 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "0.7.1")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-osgi" % "0.6.0")
 
-addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.6")
+addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1-b71a3e88f0a21d6ce5efe7c0580dba9784295358")
+
```
patriknw: it's published now?

ktoso: good catch, yes it is.
I made a separate bump PR, https://github.com/akka/akka/pull/17605 so if we include the bump here the other one can be closed

patriknw:
LGTM

rkuhn:
So, is this good to merge?

project/plugins.sbt
2m: There is a new version of mima plugins, but it does not have multiple artifacts functionality.

2m:
Not really. It needs a new release of mima plugin. I will keep this for clean up day.

rkuhn:
Okay, sounds good.

2m:
I will retarget this to master.

Issue #17355, opened 2015-04-30
Avoid case class for settings


patriknw:
case classes are extremely difficult to evolve in a binary compatible way so we should avoid it for things like flow materializer settings

drewhk:
Good point

drewhk:
What should we do with attributes?

PR #17378, opened 2015-05-05
!str #17355: Don't use case classes for settings

drewhk:


drewhk:
refs https://github.com/akka/akka/issues/17355

drewhk:
Shoud we cover attributes? Also, should I check Http as well? I think yes, but feedback is welcome. /cc @jrudolph @sirthias

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSpec.scala
```diff
@@ -529,7 +529,7 @@ class FlowSpec extends AkkaSpec(ConfigFactory.parseString("akka.actor.debug.rece
 
   "A broken Flow" must {
     "cancel upstream and call onError on current and future downstream subscribers if an internal error occurs" in {
-      new ChainSetup(faultyFlow, settings.copy(initialInputBufferSize = 1), toFanoutPublisher(initialBufferSize = 1, maximumBufferSize = 16)) {
+      new ChainSetup(faultyFlow, settings.withInputBuffer(initialSize = 1, maxSize = 1), toFanoutPublisher(initialBufferSize = 1, maximumBufferSize = 16)) {
```
patriknw: perhaps add a `withInputBuffer(size: Int)`?

patriknw:
LGTM, I don't know about the attributes. Everything case class will be a pain to evolve (at least after experimental phase). The attributes typically carry one or a few values and we can always add a new one (naming is the difficult part).

PR #17384, opened 2015-05-05
Bearer Token Authentication

pjan:
I took a first shot at adding directives for Bearer Token Authentication, collecting input here on whether this is valuable, and if, what should be changed.

Some highlights and initial questions:
- I renamed `UserCredentials` to the more general `Credentials`, and its `username` param to the more appropriate `identifier` in case it covers both `HttpBasicAuthentication` and `BearerTokenAuthentication` (?)
- `HttpBasicAuthenticator` and `BearerTokenAuthenticator` have almost exactly the same implementation, short of the naming of `provideUserName` and `provideToken` methods. I feel this might benefit from having a single `Authenticator` implementation with a `provideIdentifier` method instead (?)

Happy to receive comments and suggestions,

pjan

### DIFF
--- a/akka-http-java/src/main/scala/akka/http/impl/server/RouteImplementation.scala
+++ b/akka-http-java/src/main/scala/akka/http/impl/server/RouteImplementation.scala
@@ -8,7 +8,7 @@ import scala.language.implicitConversions
 import scala.annotation.tailrec
 import scala.collection.immutable
 import akka.http.javadsl.model.ContentType
-import akka.http.scaladsl.server.directives.AuthenticationDirectives.UserCredentials
+import akka.http.scaladsl.server.directives.AuthenticationDirectives.Credentials
 import akka.http.scaladsl.server.directives.ContentTypeResolver
 import akka.http.scaladsl.server.directives.FileAndResourceDirectives.DirectoryRenderer
 import akka.http.scaladsl.model.HttpHeader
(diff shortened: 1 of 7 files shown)


typesafehub-validator:
Hi @pjan,

ktoso:
OK TO TEST

pjan:
I suspect the build failure is due to the switch to JDK8 for PR validation yesterday?

patriknw:
yes, I have switched back to JDK6 until we fix genjavadoc

pjan:
Any input on this? The PR potentially changes some of the existing interfaces, and I believe it would be better if this is done before 1.0 reaches final?

pjan:
@sirthias or @jrudolph; I believe this is your terrain. Any input on this?

jrudolph:
It looks promising AFAICS but we don't have enough capacity to validate it properly at least this week and maybe not even before the release. Besides, it will conflict with changes in #17502 so it will have to be updated once that one has been merged.

TBH, we are reluctant at this point to accept changes that increase the API surface, as all those changes also need to have corresponding APIs on the Java side (as you have seen). I think your best bet would be to propose a change (after #17502 has been merged) that improves on the existing API elements as you suggested here and maybe introduces infrastructure to prevent code duplication so that it would be easier to add support for this kind of authentication (and other) either directly in this code base or as a third party library.

rkuhn:
So, what do we do with this PR? #17502 has been merged, should this be reworked in this PR or a new one?

pjan:
I should be able to find sometime this weekend to review and amend it based on the changes made in #17502. Either rebased in this PR, or in a new one. - this one would keep the history of the dialog though.

rkuhn:
Okay, in order to fit the PR queue on one screen my preference is to close and open a new one, linking to this discussion. Thanks a lot!

PR #17386, opened 2015-05-05
!pro #17380 Build with Java 8

patriknw:


ktoso:
@patriknw is the latest fix enough to make it pass if we'd switch the JDK on jenkins to 8? 
Latest failure was because 1.8 target was given to 1.6 JDK.

patriknw:
yes, this is supposed to work. The reason I have not changed to jdk 8 in the pr-validator is that we got some new test failures when I ran this branch in a repeat job during the weekend. The blocker is #17440.

#17314 is also failing sometimes, but it did so also before java 8.

patriknw:
PLS BUILD ALL

patriknw:
I have switched the pr-validator job to jdk8 and added some hardening of the failing tests here.

The problematic RemoteReDeploymentSpec did not fail one single time of tonights 90 repeat runs. The fix is only to increase the timeouts, so we should anyway investigate why that is needed but it is not as urgent.

project/AkkaBuild.scala
```diff
@@ -292,13 +298,13 @@ object AkkaBuild extends Build {
...
-    scalacOptions in Compile ++= Seq("-encoding", "UTF-8", "-target:jvm-1.6", "-feature", "-unchecked", "-Xlog-reflective-calls", "-Xlint"),
+    scalacOptions in Compile ++= Seq("-encoding", "UTF-8", "-target:jvm-1.8", "-feature", "-unchecked", "-Xlog-reflective-calls", "-Xlint"),
     scalacOptions in Compile ++= (if (allWarnings) Seq("-deprecation") else Nil),
     scalacOptions in Test := (scalacOptions in Test).value.filterNot(_ == "-Xlog-reflective-calls"),
     // -XDignore.symbol.file suppresses sun.misc.Unsafe warnings
-    javacOptions in compile ++= Seq("-encoding", "UTF-8", "-source", "1.6", "-target", "1.6", "-Xlint:unchecked", "-XDignore.symbol.file"),
+    javacOptions in compile ++= Seq("-encoding", "UTF-8", "-source", "1.8", "-target", "1.8", "-Xlint:unchecked", "-XDignore.symbol.file"),
     javacOptions in compile ++= (if (allWarnings) Seq("-Xlint:deprecation") else Nil),
-    javacOptions in doc ++= Seq("-encoding", "UTF-8", "-source", "1.6"),
+    javacOptions in doc ++= Seq("-encoding", "UTF-8", "-source", "1.8", "-Xdoclint:none"),
```
patriknw: It looks like this `-Xdoclint:none` setting is not used when pr-validation runs javaunidoc:doc.

When I test locally it works fine when I first `clean` and then `javaunidoc:doc`. A second `javaunidoc:doc` without cleaning generates these 100 errors.

I don't know how to solve this. HELP NEEDED!

patriknw: the failure log: https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/2636/consoleFull

ktoso: Oh perhaps it's because `javacOptions in ...`, which is not the `Test` scope, which would mean those are not seen by the validator task... I'll have a look!

ktoso: I'm finally back to looking at this, hope to solve it in a bit

akka-actor/src/main/scala/akka/io/Tcp.scala
```diff
@@ -336,9 +336,9 @@ object Tcp extends ExtensionId[TcpExt] with ExtensionIdProvider {
 
   /**
    * Write `count` bytes starting at `position` from file at `filePath` to the connection.
-   * The count must be > 0. The connection actor will reply with a [[CommandFailed]]
-   * message if the write could not be enqueued. If [[WriteCommand#wantsAck]]
-   * returns true, the connection actor will reply with the supplied [[WriteCommand#ack]]
+   * The count must be &gt; 0. The connection actor will reply with a [[CommandFailed]]
```
ktoso: Think we could build in this replacement right into genjavadoc? Seems like this must be applied everywhere, right?

akka-camel/src/main/scala/akka/camel/internal/ActivationMessage.scala
```diff
@@ -1,8 +1,7 @@
-package akka.camel.internal
-
 /**
  * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
  */
+package akka.camel.internal
```
ktoso: nice catch

akka-cluster-metrics/src/main/scala/akka/cluster/metrics/Provision.scala
```diff
@@ -15,14 +15,14 @@ import scala.util.Failure
...
+ * Provide sigar instance as `SigarProxy`.
  *
  * User can provision sigar classes and native library in one of the following ways:
  *
- * 1) Use [[https://github.com/kamon-io/sigar-loader Kamon sigar-loader]] as a project dependency for the user project.
+ * 1) Use <a href="https://github.com/kamon-io/sigar-loader">Kamon sigar-loader</a> as a project dependency for the user project.
  * Metrics extension will extract and load sigar library on demand with help of Kamon sigar provisioner.
  *
- * 2) Use [[https://github.com/kamon-io/sigar-loader Kamon sigar-loader]] as java agent: `java -javaagent:/path/to/sigar-loader.jar`
+ * 2) Use <a href="https://github.com/kamon-io/sigar-loader">Kamon sigar-loader</a> as java agent: `java -javaagent:/path/to/sigar-loader.jar`
```
ktoso: ok, good to know

ktoso:
LGTM, I scanned all the changes in this PR and looks good. 
Think we can merge this to continue with dependent things?

akka-actor/src/main/scala/akka/io/Tcp.scala
patriknw: I thought it was built in

patriknw:
one more time, I don't trust this at all

patriknw:
PLS BUILD ALL

patriknw:
@rkuhn genjavadoc is passing on jdk8 now so I'm merging this, BUT I'm extremely confused. Open the log https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/2682/consoleFull and search for "error: ". 833 matches.

For example:

    [warn] /mnt/akka/jenkinsakka/localhome-a3/workspace/pr-validator-per-commit-jenkins/akka-actor/target/java/akka/actor/Status.java:38: error: class Status is already defined in package akka.actor
    [warn]   static public  interface Status extends scala.Serializable {
    [warn]    

I thought these were the real errors and yesterday I wasted several hours in believing that things were not working (genjavadoc failed with errors and then seeing these errors), and then I think the only thing I fixed was a few more invalid links in our source. How can one see what is a real problem and what is just nonsense?

patriknw:
refs #17380

rkuhn:
@patriknw That is sbt logging being confusing: as soon as there is a single “real” error, all lines are formatted with red `[error]` marker. There are a ton of messages that javadoc cannot find things before it actually generates the HTML output, and those are all harmless, I think that just leads to arguments and return types not being linked. Finding the real ones is the challenge …

patriknw:
...yes, my problem is that I don't know how to find the real problem. I have nothing to search for that is unique for the real errors and it is not the first error.

rkuhn:
well, it is the first error after the lines
~~~
[info] Standard Doclet version 1.8.0_31
[info] Building tree for all the packages and classes...
~~~

Issue #15917, opened 2014-09-17
Create marshallers for interacting with JSON data


jrudolph:

jrudolph:
Added #16812 to provide json4s support.

ktoso:
PlayJson support is a community project now: https://github.com/guillaumebreton/play-json-support

ktoso:
Related notes: since play support will live outside, we should use this style for all support traits - including spray-json. // or alternatively bring "spray-json" in as "akka-json" - this sounds less likely though?

This should probably be addressed before 1.0

// discussed this with @hseeberger

ktoso:
We have finally reached a final consensus about where marshallers shall live.

We can not include play-json in akka itself. We do want to support Spray and the existing marshallers.
We will encourage the community to build and support external marshallers as separate repositories, such as @guillaumebreton's play support module.

This means that:
- current marshallers stay in akka/akka. We support them.
- akka-http-play is maintained by the community person, in his own repository // https://github.com/akka/akka-http-marshallers/issues/7
- If a customer will need support with play-json we'll see how to solve it then.
- akka-http-marshallers goes away.
- we close the this meta ticket, as we will not include more marshallers in akka itself. They should be community repos.

Issue #17106, opened 2015-03-30
akka-http support for PlayJson


mrkaspa:
In akka-http is there something similar to akka.http.marshallers.sprayjson.SprayJsonSupport._ but for the PlayJson Library?

ktoso:
Hi @mrkaspa,
not currently, but it would be very simple to roll your own, take a look at how the spray traits look like: https://github.com/akka/akka/blob/release-2.3-dev/akka-http-marshallers/akka-http-spray-json/src/main/scala/akka/http/marshallers/sprayjson/SprayJsonSupport.scala#L39

My immediate feeling is that these support traits need not neccessarily live inside the akka/akka repository, what do you think @akka/akka-team?

mrkaspa:
@ktoso I've done it, should i make a branch and a PR or do I create my own library?

jrudolph:
We once decided to support these kinds of integrations each with a small module like the one for spray-json. There's a meta-ticket for adding more integration for json libraries at https://github.com/akka/akka/issues/15917

hepin1989:
maybe like json4s?which have different backend

mrkaspa:
I’ve forked and downloaded the akka repository on my mac, I’ve opened the project in Intellij. I’ve created a new scala module inside the akka-http-marshallers for the play json support but the imports fot the akka libraries don’t work, how can I fix this, and I have to add the the dependency to the play-json library where should it go?, thanks

jrudolph:
@mrkaspa you need to add the project in the sbt build file first and then update IntelliJ with the changes (however you do that). See this commit https://github.com/akka/akka/commit/db3fc16f4cfedef1d815f1cb8ac27b1f16f61fab which introduced the spray-json support as a template.

ktoso:
We have discussed this module in depth in the [https://github.com/akka/akka/pull/17387#issuecomment-99606569](discussion here [#17387]).

**Summary:** because of lifecycles of projects this module should best live outside of akka and play projects - in order to be able to keep up with either version update.

Closing and expecting @guillaumebreton to roll his library out soon! :-)

PR #17387, opened 2015-05-05
+htt #15917 Adding play json support

guillaumebreton:
- Add play json dependency
- Add play json support to marshal/unmarshal json
- Add test using JsonSupportSpec

Note : at line 40 in PlayJsonSupport.scala, I leave a "// FIXME: identify charset by instance, not by name!
" comment as I don't have any idea on how to fix this.

### DIFF
--- a/akka-http-marshallers-scala/akka-http-play-json/src/main/scala/akka/http/scaladsl/marshallers/playjson/PlayJsonSupport.scala
+++ b/akka-http-marshallers-scala/akka-http-play-json/src/main/scala/akka/http/scaladsl/marshallers/playjson/PlayJsonSupport.scala
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2009-2014 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.scaladsl.marshallers.playjson
+
+import scala.language.implicitConversions
+import scala.concurrent.ExecutionContext
+import akka.stream.FlowMaterializer
+import akka.http.scaladsl.marshalling.{ ToEntityMarshaller, Marshaller }
+import akka.http.scaladsl.unmarshalling.{ FromEntityUnmarshaller, Unmarshaller }
+import akka.http.scaladsl.model.{ ContentTypes, HttpCharsets }
+import akka.http.scaladsl.model.MediaTypes.`application/json`
+import play.api.libs.json._
+import play.api.libs.functional.syntax._
+import akka.http.scaladsl.model.StatusCodes._
+import scala.util.control.NoStackTrace
+
+/**
+ * A trait providing automatic to and from JSON marshalling/unmarshalling using an in-scope *play-json* protocol.
+ */
+trait PlayJsonSupport {
+
+  type Printer = (JsValue ⇒ String)
+
+  def read[T](jsValue: JsValue)(implicit reads: Reads[T]): T = {
+    reads.reads(jsValue) match {
+      case s: JsSuccess[T] ⇒ s.get
+      case e: JsError      ⇒ throw JsResultException(e.errors)
+    }
+  }
+
+  implicit def playJsonUnmarshallerConverter[T](reads: Reads[T])(implicit ec: ExecutionContext, mat: FlowMaterializer): FromEntityUnmarshaller[T] =
+    playJsonUnmarshaller(reads, ec, mat)
+  implicit def playJsonUnmarshaller[T](implicit reads: Reads[T], ec: ExecutionContext, mat: FlowMaterializer): FromEntityUnmarshaller[T] =
+    playJsValueUnmarshaller.map(read[T])
+  implicit def playJsValueUnmarshaller(implicit ec: ExecutionContext, mat: FlowMaterializer): FromEntityUnmarshaller[JsValue] =
+    Unmarshaller.byteStringUnmarshaller.forContentTypes(`application/json`).mapWithCharset { (data, charset) ⇒
+      if (charset == HttpCharsets.`UTF-8`) Json.parse(data.toArray)
+      else Json.parse(data.decodeString(charset.nioCharset.name)) // FIXME: identify charset by instance, not by name!
+    }
+
+  implicit def playJsonMarshallerConverter[T](writes: Writes[T])(implicit printer: Printer = Json.prettyPrint, ec: ExecutionContext): ToEntityMarshaller[T] =
+    playJsonMarshaller[T](writes, printer, ec)
+  implicit def playJsonMarshaller[T](implicit writes: Writes[T], printer: Printer = Json.prettyPrint, ec: ExecutionContext): ToEntityMarshaller[T] =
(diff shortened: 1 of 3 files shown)


ktoso:
OK TO TES

ktoso:
OK TO TEST

akka-http-marshallers-scala/akka-http-play-json/src/main/scala/akka/http/scaladsl/marshallers/playjson/PlayJsonSupport.scala
```diff
@@ -0,0 +1,50 @@
+/*
```
ktoso: Thanks, that's the right place to put it.

guillaumebreton:
I checked the jenkins result, and I am not sure if I can do something about it.

project/AkkaBuild.scala
```diff
@@ -425,9 +425,10 @@ object AkkaBuild extends Build {
     id = "akka-http-marshallers-scala-experimental",
     base = file("akka-http-marshallers-scala"),
     settings = parentSettings ++ Seq(
-      version := streamAndHttpVersion
+      version := streamAndHttpVersion,
+      resolvers += "Typesafe Repo" at "http://repo.typesafe.com/typesafe/releases/"
```
ktoso: Are you sure we need to add the resolver here? I'd have thought we include it already hm..

project/AkkaBuild.scala
```diff
@@ -1547,6 +1552,9 @@ object Dependencies {
     // For akka-http spray-json support
     val sprayJson   = "io.spray"                     %% "spray-json"                   % "1.3.1"       // ApacheV2
 
+    //For akka-http play-json support
+    val playJson    = "com.typesafe.play"            %% "play-json"                    % "2.3.8"       // ApacheV2
```
ktoso: @jroper do you have an opinion about what `play-json` version should be supported here (it's for a simple "support" trait for akka-http). Play `2.4` is in RC1 mode AFAICS, and akka-http is experimental, also in RC mode - would it make sense to depdend on `2.4-RC1` here instead then?

// cc @rkuhn

ktoso:
The build failure most likely was due to us switching to JDK8 for PR validation tonight - not related to your PR - don't worry, we'll sort it out :-)

akka-http-marshallers-scala/akka-http-play-json/src/main/scala/akka/http/scaladsl/marshallers/playjson/PlayJsonSupport.scala
```diff
@@ -0,0 +1,50 @@
...
+
+import scala.language.implicitConversions
+import scala.concurrent.ExecutionContext
+import akka.stream.FlowMaterializer
+import akka.http.scaladsl.marshalling.{ ToEntityMarshaller, Marshaller }
+import akka.http.scaladsl.unmarshalling.{ FromEntityUnmarshaller, Unmarshaller }
+import akka.http.scaladsl.model.{ ContentTypes, HttpCharsets }
+import akka.http.scaladsl.model.MediaTypes.`application/json`
+import play.api.libs.json._
+import play.api.libs.functional.syntax._
```
pjan: As far as I can tell, I believe this import is not required here?

jroper:
I'm not sure about this... it introduces a circular dependency between our builds.

While at a module level, there is no circular dependency, it means we can never have a build where Play depends on the latest released version of akka-http, and akka-http depends on the latest released version of play-json.

It may also introduce circular dependencies whenever dbuild is used against Akka and Play, and this will most likely break the Reactive Platform build.

It will also introduce chicken egg scenarios - when Scala 2.12 is released, you won't be able to upgrade to it in akka-http because you'll need a play-json 2.12 build, and Play won't be able to upgrade to it because we'll need an akka-http 2.12 build.

Either we need to pull play-json out of the Play git repository, or pull this out of the Akka git repository, so that they can have independent lifecycles, and break the circular dependency.

So, I am very against this change.

rkuhn:
Since Play depends on Akka already in a high-level sense, I agree that this module should live in the Play ecosystem.

project/AkkaBuild.scala
guillaumebreton: I tried to add play-json dependency without it and it failed. I may have missed something, but it only appears in the atmos project.

hepin1989:
yes,as a play plugin

guillaumebreton:
@jroper @rkuhn  Thanks for the review. This is a perfectly valid point, sorry I missed it. Should I close the PR myself ?

@hepin1989 I'm not sure about the play plugin implementation. My point here was to support the play json lib directly in akka http, without the complete play stack. If I write a play plugin, I will be force to use it inside a play stack, am I right ?

ktoso:
Good point @jroper, thanks for chiming in. Closing this PR for now and hoping it'll find a place to live :-)

rkuhn:
I don’t think it should be a Play plugin, it should just live together with play-json itself so that it can without cycles depend upon akka-http. Play users should be able to use play-json already without such a plugin.

hepin1989:
@guillaumebreton so I think you want to using play-json with akka-http right?then may you need a akka extension.

I don't know that play-json dependents on play.should that be ?play-ws seems bound to play.

hepin1989:
a better way I think just like @jroper said.split out play-json then you could create an akka extension,which just dependent on play-json.
and play-json it's self shoudl not dependent on play but just jackson.

ktoso:
An akka extension is not really needed here - the implicits in the trait must be brought into scope to be used, nothing an extension can help with. 

+1 for taking this to live next to play-json

jroper:
play-json doesn't depend on Play, but is a part of the Play build out of convenience - as soon as you start splitting things out of builds, it adds friction, releases become more effort, sweeping changes become difficult to coordinate, versions and compatibility becomes harder to reconcile.  So you don't split things out unless you have a good reason to, and so far there's been no good reason to split play-json out of Play.

We could pull this into Play, to live alongside play-json, though it is a bit odd, because it has nothing to do with Play - Play's relation to akka-http is that it depends on lower level akka-http APIs, not these higher level ones.  So it would feel a little out of place.  If play-json was its own build, then it would be a completely different story.

My preference for now would be to have this live entirely as its own module, outside of akka-http and outside of Play.  If or when we pull play-json out of Play, then it would make sense to combine this into it.

ktoso:
I was not aware of it being tied build-wise with Play. Thanks for explaining how things fit together :-)
Seems we have all info we need to make an informed decision about this project being best hosted as separate repo/project.

I hope you'll be able to pull it off @guillaumebreton! Let us know if you'd need any help or hints

guillaumebreton:
Thanks a lot @jroper and @ktoso :) it makes things very clear ! I will extract it in a separate project :+1:

guillaumebreton:
@ktoso  Found the issue #17106 which might be linked to this discussion.

ktoso:
Linking to the repo: https://github.com/guillaumebreton/play-json-support

PR #17390, opened 2015-05-06
=act #16327 fix blockOn requirement failure

### LINKED ISSUE #16327: IllegalStateException in akka.dispatch.BatchingExecutor$Batch.run

rkuhn:
The problem was that resubmitUnbatched can destroy the contents of
parentBlockContext, leading to the observed requirement failure in
blockOn.

fixes #16327 

needs forward-port to master

### DIFF
--- a/akka-actor/src/main/scala/akka/dispatch/BatchingExecutor.scala
+++ b/akka-actor/src/main/scala/akka/dispatch/BatchingExecutor.scala
@@ -78,22 +78,23 @@ private[akka] trait BatchingExecutor extends Executor {
     }
   }
 
+  private[this] val _blockContext = new ThreadLocal[BlockContext]()
+
   private[this] final class BlockableBatch extends AbstractBatch with BlockContext {
-    private var parentBlockContext: BlockContext = _
     // this method runs in the delegate ExecutionContext's thread
     override final def run(): Unit = {
       require(_tasksLocal.get eq null)
       _tasksLocal set this // Install ourselves as the current batch
-      val prevBlockContext = BlockContext.current
+      val prevBlockContext = _blockContext.get
+      _blockContext.set(BlockContext.current)
       BlockContext.withBlockContext(this) {
-        parentBlockContext = prevBlockContext
         try processBatch(this) catch {
           case t: Throwable ⇒
             resubmitUnbatched()
             throw t
         } finally {
           _tasksLocal.remove()
-          parentBlockContext = null
+          _blockContext.set(prevBlockContext)
         }
       }
     }
@@ -102,8 +103,7 @@ private[akka] trait BatchingExecutor extends Executor {
       // if we know there will be blocking, we don't want to keep tasks queued up because it could deadlock.
       resubmitUnbatched()
       // now delegate the blocking to the previous BC
-      require(parentBlockContext ne null)
-      parentBlockContext.blockOn(thunk)
+      _blockContext.get.blockOn(thunk)
     }
   }
 
--- a/akka-actor-tests/src/test/scala/akka/dispatch/ExecutionContextSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/dispatch/ExecutionContextSpec.scala
@@ -3,8 +3,10 @@ package akka.dispatch
 import java.util.concurrent.{ ExecutorService, Executor, Executors }
 import java.util.concurrent.atomic.AtomicInteger
 import scala.concurrent._
+import scala.concurrent.duration._
 import akka.testkit.{ TestLatch, AkkaSpec, DefaultTimeout }
 import akka.util.SerializedSuspendableExecutionContext
+import akka.dispatch.ExecutionContexts.sameThreadExecutionContext
 
 @org.junit.runner.RunWith(classOf[org.scalatest.junit.JUnitRunner])
 class ExecutionContextSpec extends AkkaSpec with DefaultTimeout {
@@ -80,6 +82,29 @@ class ExecutionContextSpec extends AkkaSpec with DefaultTimeout {
         latch.countDown()
       }
       Await.ready(latch, timeout.duration)
+    }
+
+    "work with tasks that use blocking{} multiple times" in {
+      system.dispatcher.isInstanceOf[BatchingExecutor] should be(true)
+      import system.dispatcher
+
+      val f = Future(()).flatMap { _ ⇒
+        // this needs to be within an OnCompleteRunnable so that things are added to the batch
+        val p = Future.successful(42)
+        // we need the callback list to be non-empty when the blocking{} call is executing
+        p.onComplete { _ ⇒ () }
+        val r = p.map { _ ⇒
+          // trigger the resubmitUnbatched() call
+          blocking { () }
+          // make sure that the other task runs to completion before continuing
+          Thread.sleep(500)
+          // now try again to blockOn()
+          blocking { () }
+        }
+        p.onComplete { _ ⇒ () }
+        r
+      }
+      Await.result(f, 3.seconds) should be(())
     }
   }
 
--- a/project/AkkaBuild.scala
+++ b/project/AkkaBuild.scala
@@ -1094,6 +1094,8 @@ object AkkaBuild extends Build {
       ProblemFilters.exclude[FinalMethodProblem]("akka.dispatch.BatchingExecutor#Batch.run"),
       ProblemFilters.exclude[MissingMethodProblem]("akka.dispatch.BatchingExecutor#Batch.akka$dispatch$BatchingExecutor$Batch$$parentBlockContext_="),
       ProblemFilters.exclude[MissingMethodProblem]("akka.dispatch.BatchingExecutor#Batch.this"),
+      ProblemFilters.exclude[MissingMethodProblem]("akka.dispatch.BatchingExecutor.akka$dispatch$BatchingExecutor$_setter_$akka$dispatch$BatchingExecutor$$_blockContext_="),
+      ProblemFilters.exclude[MissingMethodProblem]("akka.dispatch.BatchingExecutor.akka$dispatch$BatchingExecutor$$_blockContext"),
       
       // Exclude observations from downed, #13875
       ProblemFilters.exclude[MissingMethodProblem]("akka.cluster.ClusterEvent.diffReachable"),


patriknw:
I don't know why pr validation is not picking up this.

patriknw:
PLS BUILD

akka-actor/src/main/scala/akka/dispatch/BatchingExecutor.scala
```diff
@@ -89,21 +89,22 @@ private[akka] trait BatchingExecutor extends Executor {
...
-          parentBlockContext = null
-        }
+        } finally _tasksLocal.remove()
+        parentBlockContext = null
       }
     }
 
     override def blockOn[T](thunk: ⇒ T)(implicit permission: CanAwait): T = {
+      // need to rescue the parent BC before re-running this Batch
+      val parent = parentBlockContext
```
viktorklang: `require(parent ne null)`?

akka-actor/src/main/scala/akka/dispatch/BatchingExecutor.scala
```diff
@@ -89,21 +89,22 @@ private[akka] trait BatchingExecutor extends Executor {
...
+      parentBlockContext = null
```
viktorklang: Haven't swapped this functionality into my head yet (espresso still not in blood stream), why is it safe/desirable to null it out here?

drewhk:
You said that this is deterministic under certain scenarios. Can you add a test-case?

rkuhn:
@drewhk Right, I only realized this after pushing the PR, I’ll see what I can do (and if my hunch was right).

akka-actor/src/main/scala/akka/dispatch/BatchingExecutor.scala
rkuhn: Nope, that line would not help any: either we have a bug that needs fixing, and then an NPE would tell the same story, or we don’t have a bug.

akka-actor/src/main/scala/akka/dispatch/BatchingExecutor.scala
rkuhn: To make a clean slate: resubmitUnbatched may or may not schedule this Batch to run, so don’t retain a reference in either case.

akka-actor/src/main/scala/akka/dispatch/BatchingExecutor.scala
viktorklang: Yes, and no. an assertion documents intent. An NPe does not.

akka-actor/src/main/scala/akka/dispatch/BatchingExecutor.scala
```diff
@@ -89,21 +89,22 @@ private[akka] trait BatchingExecutor extends Executor {
...
           case t: Throwable ⇒
+            parentBlockContext = null
             resubmitUnbatched()
             throw t
-        } finally {
-          _tasksLocal.remove()
-          parentBlockContext = null
-        }
+        } finally _tasksLocal.remove()
+        parentBlockContext = null
```
viktorklang: Is it intentional to put this outside of the `finally` and if so, that might be good to document on this line.

akka-actor/src/main/scala/akka/dispatch/BatchingExecutor.scala
viktorklang: Ok, that makes sense.

akka-actor/src/main/scala/akka/dispatch/BatchingExecutor.scala
rkuhn: Here I disagree: dereferencing `parent` clearly documents the intent that it is not assumed to be `null`, the code cannot possibly work without this, and the intended program semantics are that this is always non-`null`. We have countless uses of variables in our codebase that do not `require(x ne null)` before dereferencing `x`, why is this one different?

akka-actor/src/main/scala/akka/dispatch/BatchingExecutor.scala
rkuhn: Yes, very much intentional: we _must not_ modify `parentBlockContext` after `resubmitUnbatched()`, I’ll add a comment.

akka-actor/src/main/scala/akka/dispatch/BatchingExecutor.scala
viktorklang: That is a very fair objection, given the impact of bugs in this code I think extra precaution/clarity/design intent makes sense. But you're the boss :)

ktoso:
PLS BUILD

rkuhn:
Please re-review, the fix is completely different (the other was not sufficient) and there is a test now.

akka-actor/src/main/scala/akka/dispatch/BatchingExecutor.scala
```diff
@@ -78,22 +78,23 @@ private[akka] trait BatchingExecutor extends Executor {
...
   private[this] final class BlockableBatch extends AbstractBatch with BlockContext {
-    private var parentBlockContext: BlockContext = _
     // this method runs in the delegate ExecutionContext's thread
     override final def run(): Unit = {
       require(_tasksLocal.get eq null)
       _tasksLocal set this // Install ourselves as the current batch
-      val prevBlockContext = BlockContext.current
+      val prevBlockContext = _blockContext.get
+      _blockContext.set(BlockContext.current)
       BlockContext.withBlockContext(this) {
```
viktorklang: Doesn't `withBlockContext` already handle `prev` itself?

Also, when you fix this bug you'll want to forward-port it to Scala: https://github.com/scala/scala/blob/2.11.x/src/library/scala/concurrent/BatchingExecutor.scala

rkuhn: Well, we cannot access the stack frame of `withBlockContext`, so we cannot use that knowledge here AFAICS. And yes, Scala needs to be fixed as well. BTW, why doesn’t the Scala version have the non-Blockable version as well?

viktorklang: TBH I don't know right now, there were fixes made against the stdlib version. I think looking at the last changes to the Akka one and comparing that to the last cahnges to the Scala one could give a couple of hints as to what's most "current".

rkuhn: Issue created: https://issues.scala-lang.org/browse/SI-9304

hepin1989: DynamicVariable?

rkuhn: All of these closure allocations make me sad. Perhaps, some day, we can optimize them away in the distant future, but it does not look like that today. It is a shame that the optimizer cannot be used due to binary compatibility constraints.

drewhk: end users of Akka can use the optimizer though.

rkuhn: … and by doing so opt out of binary compatibility, plus that won’t fix the internal usage of `getOrElse` or `DynamicVariable` in Akka.

drewhk: > … and by doing so opt out of binary compatibility

Yes, that is their choice.

hepin1989: thanks.

patriknw:
LGTM

drewhk:
LGTM

rkuhn:
Failure due to me being surprised by `private[this] val` generating a public setter (!) and getter, added filters to MiMa for this.

patriknw:
wat, isn't it better to use `private val` then. I know `private[this]` is more idiomatic scala, but...

Issue #17392, opened 2015-05-06
Include generated japi.function from akka-stream


patriknw:

rkuhn:
fixed in #17396

patriknw:
I'll include this in master via https://github.com/akka/akka/pull/17250

PR #17396, opened 2015-05-06
+act #17392 Include generated japi.function from akka-stream

patriknw:
- add boilerplate plugin
- make them Serializable to be able to grab line number for Java 8 lambdas

### DIFF
--- a/akka-actor/src/main/boilerplate/akka/japi/function/Functions.scala.template
+++ b/akka-actor/src/main/boilerplate/akka/japi/function/Functions.scala.template
@@ -1,0 +1,29 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.japi.function
+
+[3..22#/**
+ * A Function interface. Used to create 1-arg first-class-functions is Java.
+ * `Serializable` is needed to be able to grab line number for Java 8 lambdas.
+ */
+@SerialVersionUID(##1L)
+trait Function1[[#-T1#], +R] extends java.io.Serializable {
+  @throws(classOf[Exception])
+  def apply([#arg1: T1#]): R
+}#
+
+]
+
+[2..#/**
+ * A Consumer interface. Used to create 1-arg consumers in Java.
+ * A Procedure is like a Function, but it doesn't produce a return value.
+ * `Serializable` is needed to be able to grab line number for Java 8 lambdas.
+ */
+@SerialVersionUID(##1L)
+trait Procedure1[[#-T1#]] extends java.io.Serializable {
+  @throws(classOf[Exception])
+  def apply([#arg1: T1#]): Unit
+}#
+
+]
--- a/akka-actor/src/main/scala/akka/japi/function/Function.scala
+++ b/akka-actor/src/main/scala/akka/japi/function/Function.scala
@@ -5,40 +5,50 @@ package akka.japi.function
 
 /**
  * A Function interface. Used to create first-class-functions is Java.
+ * `Serializable` is needed to be able to grab line number for Java 8 lambdas.
  */
-trait Function[-T, +R] {
+@SerialVersionUID(1L)
+trait Function[-T, +R] extends java.io.Serializable {
   @throws(classOf[Exception])
   def apply(param: T): R
 }
 
 /**
  * A Function interface. Used to create 2-arg first-class-functions is Java.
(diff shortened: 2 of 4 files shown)


ktoso:
PLS BUILD

patriknw:
This is for 2.3.11, and I would like @rkuhn to review it

rkuhn:
Good, then, that I’m back on the grid :-)

akka-actor/src/main/scala/akka/japi/function/Function.scala
```diff
@@ -5,22 +5,27 @@ package akka.japi.function
...
-trait Function2[-T1, -T2, +R] {
+@SerialVersionUID(1L)
+trait Function2[-T1, -T2, +R] extends java.io.Serializable {
   @throws(classOf[Exception])
   def apply(arg1: T1, arg2: T2): R
 }
 
 /**
  * A Procedure is like a Function, but it doesn't produce a return value.
+ * `Serializable` is needed to be able to grab line number for Java 8 lambdas.
```
rkuhn: but the `Serializable` is missing?

akka-actor/src/main/boilerplate/akka/japi/function/Functions.scala.template
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.japi.function
+
+[3..22#/**
+ * A Function interface. Used to create 1-arg first-class-functions is Java.
+ * `Serializable` is needed to be able to grab line number for Java 8 lambdas.
+ */
+@SerialVersionUID(##1L)
+trait Function1[[#T1#], R] extends java.io.Serializable {
```
rkuhn: shouldn’t they be contra-/co-variant?

rkuhn: (I think that was my mistake back then, but nevertheless it should be fixed)

akka-actor/src/main/scala/akka/japi/function/Function.scala
patriknw: ah, good catch

akka-actor/src/main/boilerplate/akka/japi/function/Functions.scala.template
patriknw: right, now they look like:

    /**
     * A Function interface. Used to create 3-arg first-class-functions is Java.
     * `Serializable` is needed to be able to grab line number for Java 8 lambdas.
     */
    @SerialVersionUID(1L)
    trait Function3[-T1, -T2, -T3, +R] extends java.io.Serializable {
      @throws(classOf[Exception])
      def apply(arg1: T1, arg2: T2, arg3: T3): R
    }
    
    /**
     * A Function interface. Used to create 4-arg first-class-functions is Java.
     * `Serializable` is needed to be able to grab line number for Java 8 lambdas.
     */
    @SerialVersionUID(1L)
    trait Function4[-T1, -T2, -T3, -T4, +R] extends java.io.Serializable {
      @throws(classOf[Exception])
      def apply(arg1: T1, arg2: T2, arg3: T3, arg4: T4): R
    }

    /**
     * A Consumer interface. Used to create 2-arg consumers in Java.
     * A Procedure is like a Function, but it doesn't produce a return value.
     * `Serializable` is needed to be able to grab line number for Java 8 lambdas.
     */
    @SerialVersionUID(1L)
    trait Procedure2[-T1, -T2] extends java.io.Serializable {
      @throws(classOf[Exception])
      def apply(arg1: T1, arg2: T2): Unit
    }
    
    /**
     * A Consumer interface. Used to create 3-arg consumers in Java.
     * A Procedure is like a Function, but it doesn't produce a return value.
     * `Serializable` is needed to be able to grab line number for Java 8 lambdas.
     */
    @SerialVersionUID(1L)
    trait Procedure3[-T1, -T2, -T3] extends java.io.Serializable {
      @throws(classOf[Exception])
      def apply(arg1: T1, arg2: T2, arg3: T3): Unit
    }

patriknw:
updated

akka-actor/src/main/boilerplate/akka/japi/function/Functions.scala.template
ktoso: Good catch. Yes variance annotations looks right on the new ones.

rkuhn:
LGTM

PR #17397, opened 2015-05-06
=pro improve PR validation speed tremendously

ktoso:
Good news everyone: Introducing PR validation with module dependency analysis!
It skips testing of modules which do not depend on the changed modules, so that changes in persistence (this commit does a mock change in persistence) only test things which depend on persistence, see logs:

```
[info] Comparing [master] with [HEAD] to determine changed modules in PR...
[info] Detected changes in directories: [akka-persistence]
[info] Project [akka-persistence-experimental] must be verified, because depends on akka-persistence-experimental_2.11
[info] Skipping validation of [akka-sample-remote-java], as PR does NOT affect this project...
[info] Changes in PR are affecting project [akka-persistence-experimental] - proceeding with test:test
... 
[info] Changes in PR are affecting project [akka-sample-persistence-scala] - proceeding with test:test
[info] Changes in PR are affecting project [akka-sample-persistence-java] - proceeding with test:test
[info] Changes in PR are affecting project [akka-contrib] - proceeding with test:test
[info] Changes in PR are affecting project [akka-docs] - proceeding with test:test
...
```

The dependency analysis uses @jrudolph's dependency graph plugin directly to analyse the deps (based on ivy reports) - This tremendously speeds up PR validation. 
We should possibly increase the times a day we run the "nightly" one so we have some during the day coverage of "everything"?

Also: 
- enable parallel execution // TODO things did not blow up for me, but _might_
- exclude perf tests (TODO mark more as such)
- implement dependency tracking for testing of only these project which could have been affected by a given PR

### DIFF
--- a/akka-persistence/build.sbt
+++ b/akka-persistence/build.sbt
@@ -21,3 +21,5 @@ MimaKeys.previousArtifact := akkaPreviousArtifact("akka-persistence-experimental
 fork in Test := true
 
 javaOptions in Test := MultiNode.defaultMultiJvmOptions
+
+
--- a/project/GitHub.scala
+++ b/project/GitHub.scala
@@ -1,6 +1,11 @@
 package akka
 
 object GitHub {
+
+  def envTokenOrThrow: String =
+    sys.env.getOrElse("PR_VALIDATOR_GH_TOKEN",
+      throw new Exception("No PR_VALIDATOR_GH_TOKEN env var provided, unable to reach github!"))
+
   def url(v: String): String = {
     val branch = if (v.endsWith("SNAPSHOT")) "master" else "v" + v
     "http://github.com/akka/akka/tree/" + branch
--- a/project/ValidatePullRequest.scala
+++ b/project/ValidatePullRequest.scala
@@ -50,9 +50,9 @@ object ValidatePullRequest extends AutoPlugin {
   override def trigger = allRequirements
 
   override lazy val projectSettings = inConfig(ValidatePR)(Defaults.testTasks) ++ Seq(
-    testOptions in ValidatePR += Tests.Argument("-l akka.testkit.PerformanceTest"),
-    testOptions in ValidatePR += Tests.Argument("-l akka.testkit.LongRunningTest"),
-    testOptions in ValidatePR += Tests.Argument("-l akka.testkit.TimingTest"),
+    testOptions in ValidatePR += Tests.Argument(TestFrameworks.ScalaTest, "-l", "performance"),
+    testOptions in ValidatePR += Tests.Argument(TestFrameworks.ScalaTest, "-l", "long-running"),
+    testOptions in ValidatePR += Tests.Argument(TestFrameworks.ScalaTest, "-l", "timing"),
 
     targetBranch in ValidatePR := {
         sys.env.get(TargetBranchEnvVarName) orElse
@@ -161,7 +161,7 @@ object ValidatePullRequest extends AutoPlugin {
       val theVoid = Def.task { () } // when you stare into the void, the void stares back at you
 
       if ((projectIsAffectedByChanges in ValidatePR).value) {
-        log.info(s"Changes in PR are affecting project [${name.value}] - proceeding with test:test")
+        log.info(s"Changes in PR are affecting project [${name.value}] - proceeding with pr-validation:test")
         theVoid.dependsOn(test in ValidatePR)
       } else {
         log.info(s"Skipping validation of [${name.value}], as PR does NOT affect this project...")
--- a/project/ValidatePullRequest.scala
+++ b/project/ValidatePullRequest.scala
@@ -3,19 +3,171 @@
  */
 package akka
 
-import sbt._
-import sbt.Keys._
 import com.typesafe.tools.mima.plugin.MimaKeys.reportBinaryIssues
+import net.virtualvoid.sbt.graph.IvyGraphMLDependencies
+import net.virtualvoid.sbt.graph.IvyGraphMLDependencies.ModuleId
+import org.kohsuke.github._
+import sbt.Keys._
+import sbt._
+
+import scala.collection.immutable
+import scala.util.matching.Regex
 
 object ValidatePullRequest extends AutoPlugin {
 
+  val ValidatePR = config("pr-validation") extend Test
+
+  override lazy val projectConfigurations = Seq(ValidatePR)
+
+  /*
+    Assumptions:
+      Env variables set "by Jenkins" are assumed to come from this plugin:
+      https://wiki.jenkins-ci.org/display/JENKINS/GitHub+pull+request+builder+plugin
+   */
+
+  // settings
+  val PullIdEnvVarName = "ghprbPullId" // Set by "GitHub pull request builder plugin"
+
+  val TargetBranchEnvVarName = "PR_TARGET_BRANCH"
+  val TargetBranchJenkinsEnvVarName = "ghprbTargetBranch"
+  val targetBranch = settingKey[String]("Branch with which the PR changes should be diffed against")
+
+  val SourceBranchEnvVarName = "PR_SOURCE_BRANCH"
+  val SourcePullIdJenkinsEnvVarName = "ghprbPullId" // used to obtain branch name in form of "pullreq/17397"
+  val sourceBranch = settingKey[String]("Branch containing the changes of this PR")
+
+  // asking github comments if this PR should be PLS BUILD ALL
+  val githubEnforcedBuildAll = taskKey[Boolean]("Checks via GitHub API if comments included the PLS BUILD ALL keyword")
+  val buildAllKeyword = taskKey[Regex]("Magic phrase to be used to trigger building of the entire project instead of analysing dependencies")
+
+  // determining touched dirs and projects
+  val changedDirectories = taskKey[immutable.Set[String]]("List of touched modules in this PR branch")
+  val projectIsAffectedByChanges = taskKey[Boolean]("True if this project is affected by the PR and should be rebuilt")
+
+  // running validation
   val validatePullRequest = taskKey[Unit]("Additional tasks for pull request validation")
 
   override def trigger = allRequirements
 
-  override lazy val projectSettings = Seq(
-    validatePullRequest := (),
-    validatePullRequest <<= validatePullRequest.dependsOn(test in Test),
+  override lazy val projectSettings = inConfig(ValidatePR)(Defaults.testTasks) ++ Seq(
+    testOptions in ValidatePR += Tests.Argument("-l akka.testkit.PerformanceTest"),
+    testOptions in ValidatePR += Tests.Argument("-l akka.testkit.LongRunningTest"),
+    testOptions in ValidatePR += Tests.Argument("-l akka.testkit.TimingTest"),
+
+    targetBranch in ValidatePR := {
+        sys.env.get(TargetBranchEnvVarName) orElse
+        sys.env.get(TargetBranchJenkinsEnvVarName) getOrElse // Set by "GitHub pull request builder plugin"
+        "master"
+    },
+
+    sourceBranch in ValidatePR := {
+      sys.env.get(SourceBranchEnvVarName) orElse
+      sys.env.get(SourcePullIdJenkinsEnvVarName).map("pullreq/" + _) getOrElse // Set by "GitHub pull request builder plugin"
+      "HEAD"
+    },
+
+    changedDirectories in ValidatePR := {
+      val log = streams.value.log
+
+      val targetId = (targetBranch in ValidatePR).value
+      val prId = (sourceBranch in ValidatePR).value
+
+      // TODO could use jgit
+      log.info(s"Comparing [$targetId] with [$prId] to determine changed modules in PR...")
+      val gitOutput = "git diff %s..%s --name-only".format(targetId, prId).!!.split("\n")
+
+      val moduleNames =
+        gitOutput
+          .map(l ⇒ l.trim.takeWhile(_ != '/'))
+          .filter(_ startsWith "akka-")
+          .toSet
+
+      log.info("Detected changes in directories: " + moduleNames.mkString("[", ", ", "]"))
+      moduleNames
+    },
+
+    buildAllKeyword in ValidatePR := """PLS BUILD ALL""".r,
+
+    githubEnforcedBuildAll in ValidatePR := {
+      sys.env.get(PullIdEnvVarName).map(_.toInt) match {
+        case None => false // can't ask github if no issue ID given
+        case Some(prId) =>
+          val log = streams.value.log
+          val buildAllMagicPhrase = (buildAllKeyword in ValidatePR).value
+          log.info("Checking GitHub comments for PR validation options...")
+
+          try {
+            import scala.collection.JavaConverters._
+            val gh = GitHubBuilder.fromEnvironment().withOAuthToken(GitHub.envTokenOrThrow).build()
+            val comments = gh.getRepository("akka/akka").getIssue(prId).getComments.asScala
+
+            comments exists { c =>
+              val triggersBuildAll = buildAllMagicPhrase.findFirstIn(c.getBody).isDefined
+              if (triggersBuildAll)
+                log.info(s"GitHub PR comment [ ${c.getUrl} ] contains [$buildAllMagicPhrase], forcing BUILD ALL mode!")
+              triggersBuildAll
+            }
+          } catch {
+            case ex: Exception =>
+              log.warn("Unable to reach GitHub! Exception was: " + ex.getMessage)
+              false
+          }
+      }
+    },
+
+    projectIsAffectedByChanges in ValidatePR := (githubEnforcedBuildAll in ValidatePR).value || {
+      val log = streams.value.log
+      log.debug(s"Analysing project (for inclusion in PR validation): [${name.value}]")
+
+      // if in any scope, any of the changed modules is within this projects dependencies, we must test it:
+      val shouldBeBuilt = (changedDirectories in ValidatePR).value.exists { modifiedProject ⇒
+        Set(Compile, Test, Runtime, Provided, Optional) exists { ivyScope: sbt.Configuration ⇒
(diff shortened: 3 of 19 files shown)


ktoso:
Very weird failure `Exception in thread "main" java.lang.NoClassDefFoundError: `

ktoso:
PLS BUILD

ktoso:
PLS BUILD

hepin1989:
+10000
I was thought about that but never came up with filing a issue.

project/ValidatePullRequest.scala
```diff
@@ -3,19 +3,137 @@
...
+  // running validation
   val validatePullRequest = taskKey[Unit]("Additional tasks for pull request validation")
 
   override def trigger = allRequirements
 
   override lazy val projectSettings = Seq(
-    validatePullRequest := (),
-    validatePullRequest <<= validatePullRequest.dependsOn(test in Test),
+    testOptions in ValidatePR += Tests.Argument("-l akka.testkit.PerformanceTest"),
+    testOptions in ValidatePR += Tests.Argument("-l akka.testkit.LongRunningTest"),
```
patriknw: I think we have marked a lot with `LongRunningTest`, e.g. cluster tests. Perhaps we should remove that marker for some important tests?

project/ValidatePullRequest.scala
```diff
@@ -3,19 +3,137 @@
...
+        System.getProperty(TargetBranchEnvVarName),
+        System.getProperty(TargetBranchJenkinsEnvVarName), // Set by "GitHub pull request builder plugin"
+        "master").find(_ != null).get
+    },
+
+    sourceBranch in ValidatePR := {
+      List(
+        System.getProperty(SourceBranchEnvVarName),
+        System.getProperty(SourceBranchJenkinsEnvVarName), // Set by "GitHub pull request builder plugin"
+        "HEAD").find(_ != null).get
```
patriknw: what happens if these variables are not set?
can we fall back to running everything, so that the validatePr task can be run locally?

project/ValidatePullRequest.scala
```diff
@@ -3,19 +3,137 @@
...
+    projectIsAffectedByChanges in ValidatePR := {
+      val log = streams.value.log
+      log.debug("Analysing project (for inclusion in PR validation): " + name.value)
+
+      // if in any scope, any of the changed modules is within this projects dependencies, we must test it:
+      val shouldBeBuilt = (changedDirectories in ValidatePR).value.exists { modifiedProject ⇒
+        Set(Compile, Test, Runtime, Provided, Optional) exists { ivyScope: sbt.Configuration ⇒
+          log.debug(s"Analysing $ivyScope scoped dependencies...")
+          val modifiedModule = modifiedProject match {
+            case p if p contains "persistence" => ModuleId("com.typesafe.akka", s"$p-experimental_2.11", "2.4-SNAPSHOT")
```
patriknw: is this because of `experimental`? don't we have more experimental modules? typed?

patriknw:
GREAT!

project/ValidatePullRequest.scala
ktoso: Yes exactly, I think we may get away with "look for both" in dependencies actually, so this match goes away.

project/ValidatePullRequest.scala
ktoso: Yes, lets go over the usages of these.

ktoso: (I'll do that and ping the issue)

project/ValidatePullRequest.scala
ktoso: If the env vars are not set it defaults to running what differs between HEAD and master - could be useful as such when sending PRs towards master... Or would you prefer we run-all when these are not set?

project/ValidatePullRequest.scala
ktoso: Fixed now, no need for the match! (if any "experimental" or "non experimental" appended names happen to be in a dependency we want to test it)

ktoso:
Idea, how about we don't exclude the slow tests - since thanks to the dependency tracking build times are much lower anyway now?

project/ValidatePullRequest.scala
patriknw: I see, that is good

patriknw:
Unless you change something in akka-actor, I vote for not running everything in pr-validation, many of the tagged tests often require several runs anyway to be trustworthy.
We must however take failures from the nightly jobs much more seriously and follow up on them immediately, when things are fresh.

patriknw:
... and for non-trivial PR we should point a repeat job at it before merging

patriknw:
... it would be nice with a command like PLS BUILDALL to run everything

ktoso:
BUILDALL That's actually not that hard to do! I think it's worth doing it (we know when we want to test all)

akka-persistence/build.sbt
```diff
@@ -21,3 +21,5 @@ MimaKeys.previousArtifact := akkaPreviousArtifact("akka-persistence-experimental
 fork in Test := true
 
 javaOptions in Test := MultiNode.defaultMultiJvmOptions
+
+
```
jrudolph: Accidental whitespace addition?

ktoso: Kind of, a "mock change" in persistence to sanity check it all works as expected (triggers everything that uses persistence) on jenkins.

ktoso: (Will remove before merging)

jrudolph:
Nice one! I guess there's also a (maybe even simpler?) alternative implementation possible which doesn't depend on sbt-dependency-graph but uses sbt structures directly. But I guess that's probably an exercise  that can be left for the next version.

project/ValidatePullRequest.scala
```diff
@@ -3,19 +3,174 @@
...
+          .map(l ⇒ l.trim.takeWhile(_ != '/'))
+          .filterNot(_ == "project")
+          .filter(_ startsWith "akka-")
+          .toSet
+
+      log.info("Detected changes in directories: " + moduleNames.mkString("[", ", ", "]"))
+      moduleNames
+    },
+
+    buildAllKeyword in ValidatePR := """PLS BUILD ALL""".r,
```
ktoso: Check this out! :shipit:

ktoso:
Pushed with new ALL feature. I'll be back in a bit to investigate the failures on jenkins.

project/ValidatePullRequest.scala
drewhk: wow!

project/GitHub.scala
```diff
@@ -1,6 +1,12 @@
 package akka
 
 object GitHub {
+
+  def envTokenOrThrow: String =
+    Option(System.getenv("PR_VALIDATOR_GH_TOKEN")) getOrElse {
```
fsvehla: `sys.env`?

project/ValidatePullRequest.scala
```diff
@@ -3,19 +3,174 @@
...
+
+    githubEnforcedBuildAll in ValidatePR := {
+      Option(System.getProperty(PullIdEnvVarName)).map(_.toInt) match {
```
fsvehla: `sys.props`?

ktoso: sure, thanks

project/GitHub.scala
ktoso: good one, thanks

ktoso:
PLS BUILD

ktoso:
PLS BUILD

patriknw:
LGTM, apart from excluding all LongRunning. Would you like me to select some tests of the LongRunning that we should always run, i.e. remove the LongRunning tag?

ktoso:
Yes please, your feedback on which tests we always want to run would be most welcome - thanks

patriknw:
Here is a selection of tests that we can remove the `LongRunningTest` tag from to get some kind of smoke test coverage:
* ClusterRoundRobinSpec
* NodeUpSpec
* ClusterDeathWatchSpec
* SurviveNetworkInstabilitySpec
* ClusterMetricsEnabledSpec (new akka-cluster-metrics module)
* RemoteRoundRobinSpec
* NewRemoteActorSpec
* RemoteNodeDeathWatchSpec
* RemoteDeliverySpec

patriknw:
note that tests in contrib don't have the LongRunning tag, that might be something that we want to revisit when we move those to their own modules

drewhk:
LGTM, but some of the stuff is just black magic for me

project/ValidatePullRequest.scala
```diff
@@ -3,19 +3,171 @@
...
+  // settings
+  val PullIdEnvVarName = "ghprbPullId" // Set by "GitHub pull request builder plugin"
+
+  val TargetBranchEnvVarName = "PR_TARGET_BRANCH"
+  val TargetBranchJenkinsEnvVarName = "ghprbTargetBranch"
+  val sourceBranch = settingKey[String]("Branch containing the changes of this PR")
+
+  val SourceBranchEnvVarName = "PR_SOURCE_BRANCH"
+  val SourceBranchJenkinsEnvVarName = "ghprbSourceBranch"
+  val targetBranch = settingKey[String]("Branch with which the PR changes should be diffed against")
```
2m: Should this line be swapped with line 31 above?

project/ValidatePullRequest.scala
```diff
@@ -3,19 +3,171 @@
...
+            val f2 = resolutionFile(includeScalaVersion = false)
+            if (f1.exists()) f1 else f2
+          }
+
+          val deps = IvyGraphMLDependencies.graph(ivyReportFile.getAbsolutePath)
+          deps.nodes.foreach { m ⇒ log.debug(" -> " + m.id) }
+
+          // if this project depends on a modified module, we must test it
+          deps.nodes.exists { m =>
+            val depends = modifiedModuleIds exists { _.name == m.id.name } // match just by name, we'd rather include too much than too little
```
2m: Since you match only by name this line could be simplified to `val depends = m.id.name startsWith modifiedProject`, right? And then `modifiedModuleIds` can be removed.

project/ValidatePullRequest.scala
ktoso: Hah, good catch! Thanks

project/ValidatePullRequest.scala
```diff
@@ -3,19 +3,171 @@
...
+          }
+        }
+      }
+
+      shouldBeBuilt
+    },
+
+    validatePullRequest := Def.taskDyn {
+      val log = streams.value.log
+      val theVoid = Def.task { () } // when you stare into the void, the void stares back at you
```
2m: This looks like Eye of ~~Nietzsche~~ Sauron

project/ValidatePullRequest.scala
```diff
@@ -3,19 +3,171 @@
...
+
+    changedDirectories in ValidatePR := {
+      val log = streams.value.log
+
+      val targetId = (targetBranch in ValidatePR).value
+      val prId = (sourceBranch in ValidatePR).value
+
+      // TODO could use jgit
+      log.info(s"Comparing [$targetId] with [$prId] to determine changed modules in PR...")
+      val gitOutput = "git diff %s..%s --stat".format(targetId, prId).!!.split("\n")
```
2m: This produces the following output on my machine when converted to List:
```
List( .../src/test/scala/akka/camel/internal/ActivationTrackerTest.scala     | 3 ++-,  1 file changed, 2 insertions(+), 1 deletion(-))
```

which then of course makes the code below match nothing. I propose to change the command to `git diff %s..%s --name-only` which outputs
```
List(akka-camel/src/test/scala/akka/camel/internal/ActivationTrackerTest.scala)
```

allowing to simplify the code below.

ktoso: Oh nice, thanks!

project/ValidatePullRequest.scala
hepin1989: good at imagnation

project/ValidatePullRequest.scala
```diff
@@ -3,19 +3,171 @@
...
+      shouldBeBuilt
+    },
+
+    validatePullRequest := Def.taskDyn {
+      val log = streams.value.log
+      val theVoid = Def.task { () } // when you stare into the void, the void stares back at you
+
+      if ((projectIsAffectedByChanges in ValidatePR).value) {
+        log.info(s"Changes in PR are affecting project [${name.value}] - proceeding with test:test")
+        theVoid.dependsOn(test in Test)
```
2m: This should be `test in ValidatePR` (and the log string above `pr-validation:test`) as otherwise tag filters which are set on `ValidatePR` scope will not kick in.

ktoso: whoa, good catch thanks

project/ValidatePullRequest.scala
```diff
@@ -3,19 +3,171 @@
...
+import org.kohsuke.github._
+import sbt.Keys._
+import sbt._
+
+import scala.collection.immutable
+import scala.util.matching.Regex
 
 object ValidatePullRequest extends AutoPlugin {
 
+  val ValidatePR = config("pr-validation") extend Test
```
2m: For `test in ValidatePR` to work more is needed than this. `... extend Test` only specifies where to look for settings if a setting is not found under `ValidatePR` scope.

Add all test related tasks under the new configuration scope by:
```
override lazy val projectSettings = inConfig(ValidatePR)(Defaults.testTasks) ++ ...
```

Then add all the settings of `ValidatePR` to every project that has this autoplugin enabled:
```
override lazy val projectConfigurations = Seq(ValidatePR)
```

And finaly, since we have this neat sample build transformation, which unfortunately does not consider AutoPlugins add `.configs(ValidatePullRequest.ValidatePR)` [here](https://github.com/akka/akka/blob/e52ce3afaf2ecee2e68ec1677ec1e29568d320a3/project/Sample.scala#L68).

project/ValidatePullRequest.scala
ktoso: (-:

project/ValidatePullRequest.scala
```diff
@@ -3,19 +3,171 @@
...
   val validatePullRequest = taskKey[Unit]("Additional tasks for pull request validation")
 
   override def trigger = allRequirements
 
   override lazy val projectSettings = Seq(
-    validatePullRequest := (),
-    validatePullRequest <<= validatePullRequest.dependsOn(test in Test),
+    testOptions in ValidatePR += Tests.Argument("-l akka.testkit.PerformanceTest"),
+    testOptions in ValidatePR += Tests.Argument("-l akka.testkit.LongRunningTest"),
+    testOptions in ValidatePR += Tests.Argument("-l akka.testkit.TimingTest"),
```
2m: When I tried it, it errored out with `Runpath must be either zero or two args`. It seems that it is related [to this issue](https://github.com/scalatest/scalatest/pull/231#issuecomment-40059704) which is fixed in `2.2.0-RC1`. We can update to `2.2.4` in this PR as well.

I also noticed that tests where excluded only when I changed it to `Tests.Argument("-n", "akka.testkit.PerformanceTest")` which is very strange as `-n` should include tags as [ScalaTest documentation](http://www.scalatest.org/user_guide/using_the_runner) suggests. Maybe a bug in ScalaTest sbt framework?

project/ValidatePullRequest.scala
ktoso: Oh wow, that's weird. Thanks a lot for helping out here. Done.

project/ValidatePullRequest.scala
```diff
@@ -3,19 +3,171 @@
...
 
   override def trigger = allRequirements
 
-  override lazy val projectSettings = Seq(
-    validatePullRequest := (),
-    validatePullRequest <<= validatePullRequest.dependsOn(test in Test),
+  override lazy val projectSettings = inConfig(ValidatePR)(Defaults.testTasks) ++ Seq(
+    testOptions in ValidatePR += Tests.Argument("-l akka.testkit.PerformanceTest"),
+    testOptions in ValidatePR += Tests.Argument("-l akka.testkit.LongRunningTest"),
+    testOptions in ValidatePR += Tests.Argument("-l akka.testkit.TimingTest"),
```
2m: When I tried it, it errored out with `Runpath must be either zero or two args`. It seems that it is related [to this issue](https://github.com/scalatest/scalatest/pull/231#issuecomment-40059704) which is fixed in `2.2.0-RC1`. We can update to `2.2.4` in this PR as well.

I also noticed that tests where excluded only when I changed it to `Tests.Argument("-n", "akka.testkit.PerformanceTest")` which is very strange as `-n` should include tags as [ScalaTest documentation](http://www.scalatest.org/user_guide/using_the_runner) suggests. Maybe a bug in ScalaTest sbt framework?

project/ValidatePullRequest.scala
```diff
@@ -3,19 +3,171 @@
...
+
+      shouldBeBuilt
+    },
+
+    validatePullRequest := Def.taskDyn {
+      val log = streams.value.log
+      val theVoid = Def.task { () } // when you stare into the void, the void stares back at you
+
+      if ((projectIsAffectedByChanges in ValidatePR).value) {
+        log.info(s"Changes in PR are affecting project [${name.value}] - proceeding with test:test")
```
2m: Change this to `pr-validation:test` as well.

project/ValidatePullRequest.scala
2m: Finally I figured it out! Tag name, not the full class name, needs to be given to scalatest. The following lines need to be changed to:
```
testOptions in ValidatePR += Tests.Argument("-l", "timing"),
testOptions in ValidatePR += Tests.Argument("-l", "long-running"),
testOptions in ValidatePR += Tests.Argument("-l", "performance"),
```

ktoso:
`akka.typed.PerformanceSpec` failed. 
All modules which were expected to execute did, the ignores worked as well from what I see.

I'd like to merge this now, in case we need more tweaks in it I'll do them as we notice what needs tweaking.

2m:
LGTM

patriknw:
+1

ktoso:
Merged and will keep an eye on all PRs if they work as expected.

Issue #13935, opened 2014-03-18
Throw same exception when extension init fails


kodemaniak:
imported from https://www.assembla.com/spaces/akka/tickets/3935

When a configured journal plugin is missing from the classpath, the following exception is thrown:

11:23:37.614 [kiarchive-akka.actor.default-dispatcher-19] ERROR akka.actor.OneForOneStrategy - actor name [snapshot-store] is not unique!
akka.actor.ActorInitializationException: exception during creation
        at akka.actor.ActorInitializationException$.apply(Actor.scala:164) ~[com.typesafe.akka.akka-actor_2.10-2.3.0.jar:na]
        at akka.actor.ActorCell.create(ActorCell.scala:596) ~[com.typesafe.akka.akka-actor_2.10-2.3.0.jar:na]
        at akka.actor.ActorCell.invokeAll$1(ActorCell.scala:456) ~[com.typesafe.akka.akka-actor_2.10-2.3.0.jar:na]
        at akka.actor.ActorCell.systemInvoke(ActorCell.scala:478) ~[com.typesafe.akka.akka-actor_2.10-2.3.0.jar:na]
        at akka.dispatch.Mailbox.processAllSystemMessages(Mailbox.scala:263) ~[com.typesafe.akka.akka-actor_2.10-2.3.0.jar:na]
        at akka.dispatch.Mailbox.run(Mailbox.scala:219) ~[com.typesafe.akka.akka-actor_2.10-2.3.0.jar:na]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1110) ~[na:1.7.0_09-icedtea]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:603) ~[na:1.7.0_09-icedtea]
        at java.lang.Thread.run(Thread.java:722) ~[na:1.7.0_09-icedtea]
Caused by: akka.actor.InvalidActorNameException: actor name [snapshot-store] is not unique!
        at akka.actor.dungeon.ChildrenContainer$NormalChildrenContainer.reserve(ChildrenContainer.scala:130) ~[com.typesafe.akka.akka-actor_2.10-2.3.0.jar:na]
        at akka.actor.dungeon.Children$class.reserveChild(Children.scala:77) ~[com.typesafe.akka.akka-actor_2.10-2.3.0.jar:na]
[...]

PR #17400, opened 2015-05-06
=act #13935 Throw same exception when extension init fails

patriknw:
- The problem is that when an extension partly fails the next
  attempt will typically generate another failure, such as
  "actor name [snapshot-store] is not unique"
- We have seen this problem for both persistence and cluster
  extensions
- Extensions are now only given one chance to initialize and
  thereafter fail fast with same exception as the the first failure


patriknw:
Refs #13935

drewhk:
LGTM

patriknw:
failure was unrelated (known issue)

ktoso:
LGTM

Issue #17341, opened 2015-04-30
Actor starvation when self-messages and different dispatchers are involved


clockfly:
## Problem Statement:

We have a benchmark test which process 10 million message per second(project https://github.com/intel-hadoop/gearpump), during the test, I found sometimes some message cannot be delivered to target actor.

## Test case description:
The test is like this:
I have 600 upstream actor, and 600 downstream actors distribued in 60 JVM. the upstream actors will crazily send data to downstream actors in a shuffle way. The upstream actors rely on self-sending message to trigger the repeation of sending. 

upstream tasks and downstream tasks will share a custom dispatcher(not default dispatcher, or remote-default-dispatcher).

Upstream Actor logic:
1. upstream actor A send itself a tick message
2. in receive of A, sending one messge to downstream actors. After sending, send itself another tick message
3. repeat step 2

## Trouble-shooting and findings
I compiled akka source to add log trace, and confirmed that the lost message is already inserted into target Mailbox queue(SingleConsumerOnlyUnboundedMailbox), and the Malibox(Runnable) is already inserted into internal task queue of fork-join pool. **The strange part** is that the mbox was **NOT** get scheduled to a real thead and run.

### Finding1: about akka Version
I find version akka 2.3.6 don't have this problem, while version 2.3.9 has.

### Finding2: about self-sending message
For my test, the upstream actors need to send themselves repeatedly tick message so that they can trigger themselves to send more message downstream. If I change the self-sending message, to be an indirect self sending, then the problem is gone. 
For example, instead of a sending directly to itself A->A, I create a intermediate actor C, A->C->A,

### Finding 3: About changes in latest akka version
Since 2.3.6 doesn't have this problem, I browser through the akka history, and find a suspicious change. https://github.com/akka/akka/pull/16152/files#diff-2c9d0638ce3a1f007623ac9d1ead1725R379

The intent of this PR is to make MailBox a subclass of ForkjoinTask, so that we can optimize the performance by fully using fork-join ideas. I doubt the changes in this PR will also create fairness problem. 

Here is some change in AkkaForkJoinPool
```scala
     extends ForkJoinPool(parallelism, threadFactory, unhandledExceptionHandler, true) with LoadMetrics {
-    override def execute(r: Runnable): Unit =		+    override def execute(r: Runnable): Unit = {
-      if (r eq null) throw new NullPointerException else super.execute(new AkkaForkJoinTask(r))		+      if (r eq null) throw new NullPointerException("The Runnable must not be null")
[...]

PR #17405, opened 2015-05-07
=act #17341: Revert starvation prone optimization 4e0bd6d

drewhk:
- also adding test to reproduce the problem

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/dispatch/ForkJoinPoolStarvationSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/dispatch/ForkJoinPoolStarvationSpec.scala
@@ -1,0 +1,66 @@
+package akka.dispatch
+
+import akka.actor.{ ActorRef, Actor, Props }
+import akka.testkit.{ ImplicitSender, AkkaSpec }
+import com.typesafe.config.ConfigFactory
+
+object ForkJoinPoolStarvationSpec {
+  val config = ConfigFactory.parseString(
+    """
+      |actorhang {
+      |
+      |  task-dispatcher {
+      |    mailbox-type = "akka.dispatch.SingleConsumerOnlyUnboundedMailbox"
+      |    throughput = 100
+      |    fork-join-executor {
+      |      parallelism-factor = 2
+      |      parallelism-max = 2
+      |      parallelism-min = 2
+      |    }
+      |  }
+      |}
+    """.stripMargin)
+
+  class SelfBusyActor extends Actor {
+    self ! "tick"
+
+    override def receive = {
+      case "tick" ⇒
+        self ! "tick"
+    }
+  }
+
+  class InnocentActor extends Actor {
+
+    override def receive = {
+      case "ping" ⇒
+        sender ! "All fine"
+    }
+  }
+
+}
+
+class ForkJoinPoolStarvationSpec extends AkkaSpec(ForkJoinPoolStarvationSpec.config) with ImplicitSender {
+  import ForkJoinPoolStarvationSpec._
+
+  val Iterations = 1000
+
+  "AkkaForkJoinPool" must {
+
+    "not starve tasks arriving from external dispatchers under high internal traffic" in {
+      // Two busy actors that will occupy the threads of the dispatcher
+      // Since they submit to the local task queue via fork, they can starve external submissions
+      system.actorOf(Props(new SelfBusyActor).withDispatcher("actorhang.task-dispatcher"))
+      system.actorOf(Props(new SelfBusyActor).withDispatcher("actorhang.task-dispatcher"))
+
+      val innocentActor = system.actorOf(Props(new InnocentActor).withDispatcher("actorhang.task-dispatcher"))
+
+      for (_ ← 1 to Iterations) {
+        // External task submission via the default dispatcher
+        innocentActor ! "ping"
+        expectMsg("All fine")
+      }
+    }
+
+  }
+}
--- a/akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
+++ b/akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
@@ -12,7 +12,7 @@ import akka.event.{ BusLogging, EventStream }
 import com.typesafe.config.{ ConfigFactory, Config }
 import akka.util.{ Unsafe, Index }
 import scala.annotation.tailrec
-import scala.concurrent.forkjoin.{ ForkJoinTask, ForkJoinPool, ForkJoinWorkerThread }
+import scala.concurrent.forkjoin.{ ForkJoinTask, ForkJoinPool }
 import scala.concurrent.duration.Duration
 import scala.concurrent.ExecutionContext
 import scala.concurrent.ExecutionContextExecutor
@@ -377,16 +377,8 @@ object ForkJoinExecutorConfigurator {
                                threadFactory: ForkJoinPool.ForkJoinWorkerThreadFactory,
                                unhandledExceptionHandler: Thread.UncaughtExceptionHandler)
     extends ForkJoinPool(parallelism, threadFactory, unhandledExceptionHandler, true) with LoadMetrics {
-    override def execute(r: Runnable): Unit = {
-      if (r eq null) throw new NullPointerException("The Runnable must not be null")
-      val task =
-        if (r.isInstanceOf[ForkJoinTask[_]]) r.asInstanceOf[ForkJoinTask[Any]]
-        else new AkkaForkJoinTask(r)
-      Thread.currentThread match {
-        case worker: ForkJoinWorkerThread if worker.getPool eq this ⇒ task.fork()
-        case _ ⇒ super.execute(task)
-      }
-    }
+    override def execute(r: Runnable): Unit =
+      if (r eq null) throw new NullPointerException else super.execute(new AkkaForkJoinTask(r))
 
     def atFullThrottle(): Boolean = this.getActiveThreadCount() >= this.getParallelism()
   }
--- a/akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
+++ b/akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
@@ -14,7 +14,6 @@ import akka.event.Logging.Error
 import scala.concurrent.duration.Duration
 import scala.concurrent.duration.FiniteDuration
 import scala.annotation.tailrec
-import scala.concurrent.forkjoin.ForkJoinTask
 import scala.util.control.NonFatal
 import com.typesafe.config.Config
 /**
@@ -51,7 +50,7 @@ private[akka] object Mailbox {
  * INTERNAL API
  */
 private[akka] abstract class Mailbox(val messageQueue: MessageQueue)
-  extends ForkJoinTask[Unit] with SystemMessageQueue with Runnable {
+  extends SystemMessageQueue with Runnable {
 
   import Mailbox._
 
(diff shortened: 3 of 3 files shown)


akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
```diff
@@ -399,9 +391,6 @@ object ForkJoinExecutorConfigurator {
     override def getRawResult(): Unit = ()
     override def setRawResult(unit: Unit): Unit = ()
     final override def exec(): Boolean = try { runnable.run(); true } catch {
-      case ie: InterruptedException ⇒
```
drewhk: this should not be reverted right?

drewhk:
refs https://github.com/akka/akka/issues/17341

akka-actor-tests/src/test/scala/akka/dispatch/ForkJoinPoolStarvationSpec.scala
```diff
@@ -0,0 +1,66 @@
...
+    "not starve tasks arriving from external dispatchers under high internal traffic" in {
+      // Two busy actors that will occupy the threads of the dispatcher
+      // Since they submit to the local task queue via fork, they can starve external submissions
+      system.actorOf(Props(new SelfBusyActor).withDispatcher("actorhang.task-dispatcher"))
+      system.actorOf(Props(new SelfBusyActor).withDispatcher("actorhang.task-dispatcher"))
+
+      val innocentActor = system.actorOf(Props(new InnocentActor).withDispatcher("actorhang.task-dispatcher"))
+
+      for (_ ← 1 to Iterations) {
+        // External task submission via the default dispatcher
```
rkuhn: no default dispatchers are harmed in this test :-) AFAICS this only uses the ScalaTest thread and the special dispatcher you created

akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
rkuhn: indeed, and please also leave the currentStatus changes out of the revert

akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
```diff
@@ -208,13 +207,13 @@ private[akka] abstract class Mailbox(val messageQueue: MessageQueue)
...
 
-  final def canBeScheduledForExecution(hasMessageHint: Boolean, hasSystemMessageHint: Boolean): Boolean = currentStatus match {
+  final def canBeScheduledForExecution(hasMessageHint: Boolean, hasSystemMessageHint: Boolean): Boolean = status match {
     case Open | Scheduled ⇒ hasMessageHint || hasSystemMessageHint || hasSystemMessages || hasMessages
     case Closed           ⇒ false
     case _                ⇒ hasSystemMessageHint || hasSystemMessages
   }
 
-  override final def run(): Unit = {
+  final def run = {
```
rkuhn: well, this is a matter of taste, but technically our style guide demands the red version

akka-actor-tests/src/test/scala/akka/dispatch/ForkJoinPoolStarvationSpec.scala
```diff
@@ -0,0 +1,66 @@
...
+        innocentActor ! "ping"
+        expectMsg("All fine")
+      }
+    }
```
rkuhn: Good test, thanks!

rkuhn:
LGTM after reverting the unrelated reversions ;-)

akka-actor-tests/src/test/scala/akka/dispatch/ForkJoinPoolStarvationSpec.scala
drewhk: Ok, I correct the comment

akka-actor-tests/src/test/scala/akka/dispatch/ForkJoinPoolStarvationSpec.scala
drewhk: Thank our reporters :)

akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
```diff
@@ -226,21 +225,6 @@ private[akka] abstract class Mailbox(val messageQueue: MessageQueue)
     }
   }
 
-  override final def getRawResult(): Unit = ()
```
drewhk: is this needed? It was in the reverse patch and it does not seem to affect anything, but some confirmation would be helpful

patriknw:
LGTM

akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
rkuhn: Yes, once you don’t extend ForkJoinTask anymore these need to go.

rkuhn:
LGTM

drewhk:
Failure is very likely unrelated: 

[info] - must redeliver on reset *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (10 seconds) during expectMsg:
[info]   at scala.Predef$.assert(Predef.scala:179)
[info]   at akka.testkit.TestKitBase$class.expectMsgPF(TestKit.scala:356)
[info]   at akka.testkit.TestKit.expectMsgPF(TestKit.scala:718)
[info]   at akka.persistence.PersistentChannelSpec$$anonfun$1$$anonfun$apply$mcV$sp$17$$anonfun$apply$mcV$sp$19.apply$mcVI$sp(PersistentChannelSpec.scala:138)
[info]   at scala.collection.immutable.Range.foreach$mVc$sp(Range.scala:141)
[info]   at akka.persistence.PersistentChannelSpec$$anonfun$1$$anonfun$apply$mcV$sp$17.apply$mcV$sp(PersistentChannelSpec.scala:138)
[info]   at akka.persistence.PersistentChannelSpec$$anonfun$1$$anonfun$apply$mcV$sp$17.apply(PersistentChannelSpec.scala:127)

viktorklang:
Is there a ticket to fix the problem?

akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
```diff
@@ -377,16 +377,8 @@ object ForkJoinExecutorConfigurator {
                                threadFactory: ForkJoinPool.ForkJoinWorkerThreadFactory,
                                unhandledExceptionHandler: Thread.UncaughtExceptionHandler)
     extends ForkJoinPool(parallelism, threadFactory, unhandledExceptionHandler, true) with LoadMetrics {
-    override def execute(r: Runnable): Unit = {
-      if (r eq null) throw new NullPointerException("The Runnable must not be null")
-      val task =
-        if (r.isInstanceOf[ForkJoinTask[_]]) r.asInstanceOf[ForkJoinTask[Any]]
```
viktorklang: Was it verified whether it was the reuse of the Task instance or the fork() that caused the problem?

akka-actor-tests/src/test/scala/akka/dispatch/ForkJoinPoolStarvationSpec.scala
viktorklang: I posit that the test (ForkJoinPoolStarvationSpec) is broken since it passes every time when I added back what this commit reverts.

drewhk: No it is not. It failed reliably *every single time* I had run it. Also, see the original ticket for plenty of reproducers (pick any of you like): https://github.com/akka/akka/issues/17341

akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
drewhk: Not the reuse, but the call to fork(). See: https://github.com/akka/akka/issues/17341#issuecomment-99065785 and also my report: https://github.com/akka/akka/issues/17341#issuecomment-99504739

akka-actor-tests/src/test/scala/akka/dispatch/ForkJoinPoolStarvationSpec.scala
drewhk: Reproducer 1: https://github.com/akka/akka/issues/17341#issuecomment-97792949
Reproducer 2: https://github.com/akka/akka/issues/17341#issuecomment-99427533
Reproducer for supposed root cause: https://github.com/akka/akka/issues/17341#issuecomment-99065785
(and the reproducer included here)
Checking with different JDK versions: https://github.com/akka/akka/issues/17341#issuecomment-99681029

akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
viktorklang: If it is only fork() which is the problem, why revert the reuse of the Task (avoiding allocations)?

akka-actor-tests/src/test/scala/akka/dispatch/ForkJoinPoolStarvationSpec.scala
viktorklang: ActorHang never terminates, but ForkJoinPoolStarvationSpec passes every single time without reverting the code. So ForkJoinPoolStarvationSpec is broken?

drewhk: I just double checked again. with your original code this test fails 10 out of 10 times. With your changes reverted (i.e. after this PR) it passes 10 times out of 10 times. It is not broken at all. It might be system specific to some extent, but that does not change the fact that it is 100% reproducable on my machine and also other machines (the two independent users who reproduced the problem).

akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
drewhk: Because your whole commit has been reverted. We don't want to take risks on 2.3.x anymore, this bug is critical already. The original code worked for many-many versions in a row. If we want to experiment further, it will go into 2.4, but 2.3 remains conservative in this regard.

akka-actor-tests/src/test/scala/akka/dispatch/ForkJoinPoolStarvationSpec.scala
viktorklang: Thanks for double-checking, I wiped everything clean and checked out v2.3.9 and tested the spec again, and now it fails reliably for me as well.
Cache related?

akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala
viktorklang: I think that's a fair stance, @drewhk.

The appropriate fix seems to just not use `.fork()`.

Let me know if you want help on fixing it for 2.4

rkuhn: Really tackling this means talking with Doug and possibly rolling our own variant of FJP, the current revert is only a stop gap measure to reduce risk. "talking with Doug" is not something that we track on github, though ;-)

viktorklang: @rkuhn As it happens I have planned to talk to him next week. In any case, if everything works if we omit the `fork()` call then we still ought to reuse the FJT to conserve allocations IMO. Long term plan to not use ExecutorServices at all seems like a sensible thing at this point :)

PR #17408, opened 2015-05-07
=act #17071 Update typesafe-config to 1.3.0

### LINKED ISSUE #17071: update Typesafe config to 1.3.0-M3
from Havoc:

Here's what's new:
https://github.com/typesafehub/config/blob/master/NEWS.md#130-m1-march-6-2015

It probably makes sense to move current dev versions of akka and play
to this release and see if anything explodes. I would do a 1.3.0 final
before your projects go final, of course.

At this time all the open issues are feature requests or I couldn't
reproduce them. I don't know of anything that's broken. So if you find
anything let me know.

I think akka was implementing the `Config` interface, which has new
methods added, so it probably will not compile. `Config` is "sealed by
documentation" by the way since Java doesn't have sealed. :-)

However, the optimizations to path parsing in this release may mean
the caching you were doing isn't needed anymore. Details of those
optimizations are at https://github.com/typesafehub/config/issues/205
in case you feel like deleting code. I think even more optimization
could be possible but it's already fast so you will have to tell me if
it matters to you.

ktoso:
Resolves https://github.com/akka/akka/issues/17071

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/ActorSystem.scala
+++ b/akka-actor/src/main/scala/akka/actor/ActorSystem.scala
@@ -63,14 +63,14 @@ object ActorSystem {
    * then tries to walk the stack to find the callers class loader, then falls back to the ClassLoader
    * associated with the ActorSystem class.
    *
-   * @see <a href="http://typesafehub.github.io/config/v1.2.1/" target="_blank">The Typesafe Config Library API Documentation</a>
+   * @see <a href="http://typesafehub.github.io/config/v1.3.0/" target="_blank">The Typesafe Config Library API Documentation</a>
    */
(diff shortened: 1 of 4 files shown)


ktoso:
I just realised that this will fail validation because it needs Java 8, yet we don't run with it yet on Jenkins.
Effectively this ticket depends on updating `gen-javadoc` then (and using JDK8 on jenkins)

patriknw:
LGTM

drewhk:
LGTM

patriknw:
PLS BUILD

ktoso:
PLS BUILD

patriknw:
needs a rebase, probably the dependencies file

ktoso:
PLS BUILD

PR #17412, opened 2015-05-07
=act #16969 Props.create must not check STATIC flag

### LINKED ISSUE #16969: Props.create must not check for STATIC flag
The presence of this flag for anonymous classes inside static methods was a [bug](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8034044) that will be fixed.

Proposal from Steve Ramage on the mailing list: «a better check would be is to look at the getDeclaredConstructors and see if there is an implicit constructor that matches the arguments provided, if it was static then it should exist, otherwise it will have an implicit argument for the outer class.»

ktoso:
Resolves https://github.com/akka/akka/issues/16969

### DIFF
--- a/akka-actor-tests/src/test/java/akka/actor/ActorCreationTest.java
+++ b/akka-actor-tests/src/test/java/akka/actor/ActorCreationTest.java
@@ -5,49 +5,38 @@
 package akka.actor;
 
 import static org.junit.Assert.*;
+
+import akka.testkit.TestActors;
 import org.junit.Test;
 
 import akka.japi.Creator;
 
 public class ActorCreationTest {
 
-  @Test
-  public void testWrongCreator() {
-    try {
-      Props.create(new Creator<Actor>() {
-        @Override
-        public Actor create() throws Exception {
-          return null;
-        }
-      });
-      assert false;
-    } catch(IllegalArgumentException e) {
-      assertEquals("cannot use non-static local Creator to create actors; make it static (e.g. local to a static method) or top-level", e.getMessage());
-    }
-  }
-  
   static class C implements Creator<UntypedActor> {
     @Override
     public UntypedActor create() throws Exception {
       return null;
     }
   }
-  
+
   static class D<T> implements Creator<T> {
     @Override
     public T create() {
       return null;
     }
   }
-  
+
   static class E<T extends UntypedActor> implements Creator<T> {
     @Override
     public T create() {
       return null;
     }
   }
-  
-  static interface I<T> extends Creator<UntypedActor> {}
+
+  static interface I<T> extends Creator<UntypedActor> {
+  }
+
   static class F implements I<Object> {
     @Override
     public UntypedActor create() {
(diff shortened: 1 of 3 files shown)


akka-actor-tests/src/test/java/akka/actor/ActorCreationTest.java
```diff
@@ -117,5 +141,25 @@ public void testRejectAbstractActorClass() {
       assertEquals(String.format("Actor class [%s] must not be abstract", H.class.getName()), e.getMessage());
     }
   }
-  
+
+  private static Creator<UntypedActor> createAnonymousCreatorInStaticMethod() {
+    return new Creator<UntypedActor>() {
+      @Override
```
ktoso: This is the weird case - this anonymous class would be marked as `static` because it's inside a static method, which violates: `An anonymous class is always an inner class (8.1.3); it is never static (8.1.1, 8.5.1).`

http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8034044

ktoso:
When you'll have time, can you sanity check this please @rkuhn ?

akka-actor-tests/src/test/java/akka/actor/ActorCreationTest.java
```diff
@@ -117,5 +141,25 @@ public void testRejectAbstractActorClass() {
...
+      public UntypedActor create() throws Exception {
+        return null;
+      }
+    };
+  }
+
+  @Test
+  public void testWrongAnonymousClassCreatedInStaticMethodCreator() {
+    final Creator<UntypedActor> thisIsNotStatic = createAnonymousCreatorInStaticMethod();
```
rkuhn: Why would we want to forbid this usage? The point of the check is to tell people that it is not okay to implicitly capture the outer class reference, but there is no such reference in play here. The intention of the ticket was to allow this case after the quoted bug was fixed (which removed the erroneous static flag that made it work previously).

ktoso: I don't know why but I understood the inversed case from the ticket somehow. Will fix.

ktoso:
Reimplemented by looking at matching constructors.

ktoso:
w8, found a mistake..

ktoso:
Ok, current impl does what we need it to I believe. Checked perf as well, the checks take less than 0.2 μs in benchmarks (various classes), so should be ok, even though we fail the check via exceptions.

akka-actor/src/main/scala/akka/actor/Props.scala
```diff
@@ -130,6 +131,21 @@ object Props {
...
+  private final def isNotClosingOverEnclosingClass(clazz: Class[_]): Boolean = {
+    def findMatchingConstructor(argTypes: Array[Class[_]]): Boolean =
+      try { val c = clazz.getConstructor(argTypes: _*); true }
+      catch { case ex: NoSuchMethodException ⇒ false }
+    def findMatchingDeclaredConstructor(argTypes: Array[Class[_]]): Boolean =
+      try { val c = clazz.getDeclaredConstructor(argTypes: _*); true }
+      catch { case ex: NoSuchMethodException ⇒ false }
+
+    //noinspection EmptyCheck
+    def defaultEmptyConstructor = clazz.getConstructors.length == 0 && findMatchingDeclaredConstructor(Array.empty)
```
rkuhn: according to the docs `length == 0` only if the class does not have public constructors, which means that this check will never be true

rkuhn: ah, no, I overlooked that important detail …

akka-actor/src/main/scala/akka/actor/Props.scala
```diff
@@ -130,6 +131,21 @@ object Props {
...
+    def matchingConstructor = clazz.getDeclaredConstructors.exists { c ⇒ findMatchingConstructor(c.getParameterTypes) }
```
rkuhn: but here I’m pretty certain that the check does nothing: for every constructor there obviously will be a matching constructor with the same argument list

ktoso: There is a slight twist to it here: For a `declared constructor`, we look for a `constructor`.
See here what happens for an anonymous class (dcs = declared constructors, cs = constructors):

```
Props.create(new C() {});
dcs = List(List(akka.actor.ActorCreationTest))
cs  = List()

      final Creator<UntypedActor> anonymousCreatorFromStaticMethod = new P("hello") {
        // captures enclosing class
      };

dcs = List(List(akka.actor.ActorCreationTest, java.lang.String))
cs  = List()
```

rkuhn: My point was that this check can be replace with `clazz.getConstructors.length > 0` AFAICS.

rkuhn:
@akka/akka-team more reviews?

rkuhn:
LGTM apart from that one remaining question.

akka-actor/src/main/scala/akka/actor/Props.scala
patriknw: wow, what is the performance impact of doing all these checks?

ktoso: Hm... I thought I posted a benchmark together with this PR, seems not or it was lost in a commit-line comment maybe. I remember it not being significant (entire method taking 0.2 micros), but I'll check again - esp. with validating Rolands latest hint

drewhk:
@ktoso, ping?

ktoso:
I discovered that junit tests are not running on the HEAD here... Need to solve this first.

patriknw:
@ktoso what is the status here? now or later?

ktoso:
Later; I'm not sure if I got the semantics right. Can fit it in for next week I think

patriknw:
good

drewhk:
Can we close this to not pollute the open PRs space?

ktoso:
Yeah, let's close; meant to come back to it but it's not happening recently. Closing

PR #17416, opened 2015-05-07
=act #16327 fix BatchingExecutor.blockOn (for verification)

rkuhn:
It contained a difficult to hit race condition that was exploited with
the help of a custom same-thread execution context by Play (its Iteratee
trampoline). In short: don’t resubmit the current Batch if it contains
an unsynchronized variable.


patriknw:
this has already been merged, by cherry-pick https://github.com/akka/akka/commit/25ddd1b

Issue #13919, opened 2014-03-12
Remove deprecated actorFor


patriknw:
imported from https://www.assembla.com/spaces/akka/tickets/3919

hepin1989:
@patriknw this one should be marked to 24-M1

PR #17424, opened 2015-05-08
!act #13919 Make actorFor and isTerminated internal 

patriknw:
- until we have replaced all internal usages of it
- for `isTerminated` we could later decide that it is good to keep as an
  internal facility and then we can remove the deprecation annotations

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/ActorRef.scala
+++ b/akka-actor/src/main/scala/akka/actor/ActorRef.scala
@@ -89,9 +89,6 @@ object ActorRef {
  * Two actor references are compared equal when they have the same path and point to
  * the same actor incarnation. A reference pointing to a terminated actor doesn't compare
  * equal to a reference pointing to another (re-created) actor with the same path.
- * Actor references acquired with `actorFor` do not always include the full information
- * about the underlying actor identity and therefore such references do not always compare
- * equal to references acquired with `actorOf`, `sender`, or `context.self`.
  *
  * If you need to keep track of actor references in a collection and do not care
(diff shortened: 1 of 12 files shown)


patriknw:
refs #13919

patriknw:
This should probably be merged after the BC changes, because it will need some additional mima filters.

akka-actor/src/main/scala/akka/actor/ActorRef.scala
```diff
@@ -134,7 +131,8 @@ abstract class ActorRef extends java.lang.Comparable[ActorRef] with Serializable
    * The contract is that if this method returns true, then it will never be false again.
    * But you cannot rely on that it is alive if it returns false, since this by nature is a racy method.
    */
-  @deprecated("Use context.watch(actor) and receive Terminated(actor)", "2.2") def isTerminated: Boolean
+  @deprecated("Use context.watch(actor) and receive Terminated(actor)", "2.2")
```
drewhk: INTERNAL API?

drewhk:
LGTM

ktoso:
+1 on the one missing INTERNAL marker.
LGTM

patriknw:
@rkuhn ok to merge, or do you want to look at this one?

akka-actor/src/main/scala/akka/actor/ActorRef.scala
rkuhn: yes, good point

patriknw: yes, thanks

rkuhn:
LGTM

Issue #17418, opened 2015-05-07
`ctx.pull` during `isFinishing` leads to endless loop in OneBoundedInterpreter


jrudolph:
See #17417 for an example.

/cc @sirthias

drewhk:
Solution is to throw an exception to the user. Hopefully this is does not need an extra branch but we can leverage the new bit-fields we have already.

PR #17430, opened 2015-05-08
+str #17418: Improved error handling for termination cases in Interpreter

drewhk:


drewhk:
refs https://github.com/akka/akka/issues/17418

rkuhn:
LGTM

drewhk:
TLS again

jrudolph:
A higher-level (API-level) test-case would have been nice which also asserts on the error message to expect. As this was an API misuse case in the first place, the important thing would be to assure that the user is able to fix the Stage implementation. What the user gets now is this: 

```
java.lang.IllegalStateException: operation requires [no balls and not isFinishing] while holding [no balls and isFinishing] and receiving [downstream ball and isFinishing]
	at akka.stream.impl.fusing.OneBoundedInterpreter$State$class.mustHave(Interpreter.scala:283)
	at akka.stream.impl.fusing.OneBoundedInterpreter$$anon$3.mustHave(Interpreter.scala:467)
	at akka.stream.impl.fusing.OneBoundedInterpreter$State$class.pull(Interpreter.scala:306)
	at akka.stream.impl.fusing.OneBoundedInterpreter$$anon$3.pull(Interpreter.scala:467)
	at akka.stream.stage.StageState.onPull(Stage.scala:311)
	at akka.stream.stage.StatefulStage.onPull(Stage.scala:374)
	at akka.stream.stage.StatefulStage.onPull(Stage.scala:335)
	at akka.stream.impl.fusing.OneBoundedInterpreter$$anon$4.absorbTermination(Interpreter.scala:503)
	at akka.http.scaladsl.coding.GzipDecompressor.onUpstreamFinish(Gzip.scala:117)
```

Also the stack trace doesn't point to the erroneous line but to the related but misleading `absorbTermination` call.

jrudolph:
Ah, it doesn't point to the right line because the pull is done in `StageState.onPull` which is inherited by the broken user-State class. So, while this has now been improved on the implementation level it should also (instead?) be fixed on the API-level where it's still the framework code which makes the forbidden call.

drewhk:
I am not sure I follow. What do you mean "fixed on the API-level"?

rkuhn:
Probably that the default implementation of onPull should not pull when finishing.

drewhk:
That is only relevant for PushStage, and the illegal call is not actually the pull, but the previous absorbTermination. PushStage should not call that ever. I only used a PushStage in the test because I was lazy to write the pull code.

jrudolph:
@drewhk the API should prevent that the user calls `pull` while `isFinishing` not the implementation. The difference is maybe subtle but it means that the user gets an actionable error message that points to the problem instead of the above stack trace.

@rkuhn yes, it should probably mirror the behavior of `IntermediateState` (https://github.com/akka/akka/blob/release-2.3-dev/akka-http-core/src/main/scala/akka/http/impl/util/ByteStringParserStage.scala#L28) which reports a truncation error. In many cases the problem is indeed some kind of truncation problem where the user still expects elements after the upstream has finished.

Still, also for the case where the user explicitly calls `pull` a proper error is needed which says "pull isn't allowed after absorbTermination has been called".

drewhk:
> the API should prevent that the user calls pull while isFinishing not the implementation. The difference is maybe subtle but it means that the user gets an actionable error message that points to the problem instead of the above stack trace.

I still don't follow.

drewhk:
> Still, also for the case where the user explicitly calls pull a proper error is needed which says "pull isn't allowed after absorbTermination has been called".

For this we need to introspect the stack trace. absorbTermination might result in onPull being called, so it will be the root of the stacktrace. The current error message implementation was a choice of economy. We encode flags and requirements as a bitfield so we don't have tons of branches everywhere, then when an error happens, we produce a human readable error message from the state of the bits. Any additional information adds costs, except the stacktrace introspection, because we only pay the cost of it during errors.

jrudolph:
@drewhk the above stack trace is what you get after reverting the fix for #17417. There are two problems with the stack trace: 1.) the message of the exception is worded in terms that do not relate to the API 2.) there's still a bug in `StageState.onPull` which still triggers this behavior (and maybe in PushStage as well).

drewhk:
> the message of the exception is worded in terms that do not relate to the API 2.

The message can be improved, yes. For example it could contain the name of the operation that was called. Otherwise, the "balls" are not part of the API, but they are part of the concept. I am not sure how to name it, but with more complex ops, like pushAndPull and friends (with AsyncStage) you will need this information.

> there's still a bug in StageState.onPull which still triggers this behavior (and maybe in PushStage as well).

What do you mean by bug? It reports the error properly. Or you mean it gets into an infinite loop again?

jrudolph:
> For this we need to introspect the stack trace. absorbTermination might result in onPull being called, so it will be the root of the stacktrace. The current error message implementation was a choice of economy. We encode flags and requirements as a bitfield so we don't have tons of branches everywhere, then when an error happens, we produce a human readable error message from the state of the bits.

I don't think that's necessary. As `isFinishing` and `pull` are both implemented on `Context` it should be possible to guard `ctx.pull` if `isFinishing` is true in a proper way. (Avoiding branches is premature optimization IMO as there are megamorphic call-sites all over the place which will easily dwarf the cost of having some (well-predictable) branches that will support the user in fixing bugs.)

drewhk:
> I don't think that's necessary. As isFinishing and pull are both implemented on Context it should be possible to guard ctx.pull if isFinishing is true in a proper way. 

The problem is that this is probably easy for pull, but there are more complex operations with more complex requirements. I still believe reconstructing the error message from the bitfields is the right way to go instead of sprinkling the code all over with various error messages. We can improve on the default error message calculation code.

> Avoiding branches is premature optimization IMO as there are megamorphic call-sites all over the place which will easily dwarf the cost of having some (well-predictable) branches that will support the user in fixing bugs.

Then I rather remove the megamorphic call sites. As I see you have two problems:

 - the error message: we can easily improve on this one
 - the stack trace: the stack trace you pasted *is* the correct one since it properly includes the information that the call to absorbTermination is the root of the problem. And, the current exception is actually thrown *at the call to pull*, just like you recommend: https://github.com/akka/akka/pull/17430/files#diff-189f1f3ce82d58e032f973bc1e85a0a2R306

jrudolph:
> What do you mean by bug? It reports the error properly. Or you mean it gets into an infinite loop again?

No, the infinite loop is gone. I think you got that one eradicated properly. I'm already pretty happy about that one :)

I'm just saying we need proper guards in the API (Stage, StageState, Context, etc.) against stupid users and not just safe-guards in the implementation. These are little things which help users tremendously.

drewhk:
The only difference is that I don't call isFinishing but I test multiple conditions in one go from the bitfield.

drewhk:
I created a ticket for improving the error message itself: https://github.com/akka/akka/issues/17470

With PushStage I can add an exception if onPull is called with isFinishing. Since these stages are optimized anyway (execution jumps over them while pulling, not calling onPull at all), the only time it will be called is when someone does an absorbTermination

As for the stack-trace, the issue here is how absorbTermination works. If it is called while the "ball" was upstream, but we never got it back, but an onUpstreamFinish instead, then we need to "recover" the ball, and it is done by calling onPull on the op.

jrudolph:
> We can improve on the default error message calculation code.

Yay!

> And, the current exception is actually thrown at the call to pull, just like you recommend

Yes, exactly and that's good. The problem is that the call to pull originates from the API itself, from `StageState.onPull` which as @rkuhn recommended should itself include a guard to avoid calling `pull` in these cases.

> Then I rather remove the megamorphic call sites.

I know you'd like to :) But this is rather hard a) generically, because it's likely, we suffer from [The Inlining Problem](http://www.azulsystems.com/blog/cliff/2011-04-04-fixing-the-inlining-problem) in streams as in every other generic code and b) in case of the (fusing streams) interpreter specifically because trampolining prevents inlining because it necessarily depends on it (as every other interpreter loop). That's where most of the speedup between compiled and interpreted code comes from.

drewhk:
> Yes, exactly and that's good. The problem is that the call to pull originates from the API itself, from StageState.onPull which as @rkuhn recommended should itself include a guard to avoid calling pull in these cases.

Ok, why not, StatefulStage is already adding some costs, and I can always fall back to PushPullStage if I want to avoid those costs :)

As for the megamorphic call sites, of course it is impossible to do for Stages -- they are meant to be extensible in the first place. I can only do it for the States -- instead of using subclasses and overriding some of the methods, I can merge everything into a giga-context class with no inheritance. It will be a hell to maintain, but at least the option is there if we ever need it.

rkuhn:
@jrudolph What I meant was that wherever we provide implementations we should not make this particular mistake, and this applies to both PushStage and StageState.

drewhk:
Fair enough. For PushStage adding the extra check is no cost for the normal case, and in the case of StageState it is acceptable because it is a convenience API over the low-level counterparts.

PR #17437, opened 2015-05-10
Fix error in documentation

### LINKED ISSUE #17438: Documentation contains an error
````
import akka.pattern.ask

class ExampleActor extends Actor {
  val other = context.actorOf(Props[OtherActor], "childName") // will be destroyed and re-created upon restart by default

  def receive {
    case Request1(msg) => other ! refine(msg)     // uses this actor as sender reference, reply goes to us
    case Request2(msg) => other.tell(msg, sender()) // forward sender reference, enabling direct reply
    case Request3(msg) =>
      implicit val timeout = Timeout(5.seconds)
      sender() ! (other ? msg)  // will reply with a Future for holding other's reply
  }
}
````

should be 

```
 * import akka.pattern.ask
 * import scala.concurrent.Await
 *
 * class ExampleActor extends Actor {
 *   val other = context.actorOf(Props[OtherActor], "childName") // will be destroyed and re-created upon restart by default
 *
 *   def receive {
 *     case Request1(msg) => other ! refine(msg)     // uses this actor as sender reference, reply goes to us
 *     case Request2(msg) => other.tell(msg, sender()) // forward sender reference, enabling direct reply
 *     case Request3(msg) =>
 *       implicit val timeout = Timeout(5.seconds)
 *       sender() ! Await.result(other ? msg, 2 * timeout)  // will reply with a Future for holding other's reply
 *   }
 * }
```

### PR DESCRIPTION
**Pull request by gkb**
It doesn't make sense to send a message composed of a future to another
actor. Futures aren't serializable.

Fixes https://github.com/akka/akka/issues/17438

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/ActorRef.scala
+++ b/akka-actor/src/main/scala/akka/actor/ActorRef.scala
@@ -39,6 +39,7 @@ object ActorRef {
  * Scala:
  * {{{
  * import akka.pattern.ask
+ * import scala.concurrent.Await
  *
  * class ExampleActor extends Actor {
  *   val other = context.actorOf(Props[OtherActor], "childName") // will be destroyed and re-created upon restart by default
@@ -48,7 +49,9 @@ object ActorRef {
  *     case Request2(msg) => other.tell(msg, sender()) // forward sender reference, enabling direct reply
  *     case Request3(msg) =>
  *       implicit val timeout = Timeout(5.seconds)
- *       sender() ! (other ? msg)  // will reply with a Future for holding other's reply
+ *       (other ? msg) pipeTo sender()
+ *       // the ask call will get a future from other's reply
+ *       // when the future is complete, send its value to the original sender
  *   }
  * }
  * }}}
@@ -56,6 +59,7 @@ object ActorRef {
  * Java:
  * {{{
  * import static akka.pattern.Patterns.ask;
+ * import static akka.pattern.Patterns.pipe;
  *
  * public class ExampleActor extends UntypedActor {
  *   // this child will be destroyed and re-created upon restart by default
@@ -73,7 +77,9 @@ object ActorRef {
  *
  *     } else if (o instanceof Request3) {
  *       Msg msg = ((Request3) o).getMsg();
- *       getSender().tell(ask(other, msg, 5000)); // reply with Future for holding the other's reply (timeout 5 seconds)
+ *       pipe(ask(other, msg, 5000)).to(getSender());
+ *       // the ask call will get a future from other's reply
+ *       // when the future is complete, send its value to the original sender
  *
  *     } else {
  *       unhandled(o);

### CODE REVIEW

File: akka-actor/src/main/scala/akka/actor/ActorRef.scala [hunk 1]
```diff
@@ -48,7 +49,7 @@ object ActorRef {
  *     case Request2(msg) => other.tell(msg, sender()) // forward sender reference, enabling direct reply
  *     case Request3(msg) =>
  *       implicit val timeout = Timeout(5.seconds)
- *       sender() ! (other ? msg)  // will reply with a Future for holding other's reply
+ *       (other ? msg) pipeTo sender() // will reply with a Future for holding other's reply
```
hepin1989: comment should be updated too.

gkb: Done

akka-actor/src/main/scala/akka/actor/ActorRef.scala
```diff
@@ -48,7 +49,7 @@ object ActorRef {
  *     case Request2(msg) => other.tell(msg, sender()) // forward sender reference, enabling direct reply
  *     case Request3(msg) =>
  *       implicit val timeout = Timeout(5.seconds)
- *       sender() ! (other ? msg)  // will reply with a Future for holding other's reply
+ *       (other ? msg) pipeTo sender() // will pipe the Future from other back to sender
```
hepin1989: pipe the value of the future not the future it's self

hepin1989:
and here:
https://github.com/akka/akka/blob/master/CONTRIBUTING.md

after you finished you should rebase and squash the commit into one.
like ```!doc #{issue number} {commit message}```

gkb:
Done.

patriknw:
OK TO TEST

patriknw:
LGTM, thanks

akka-actor/src/main/scala/akka/actor/ActorRef.scala
```diff
@@ -48,7 +49,7 @@ object ActorRef {
  *     case Request2(msg) => other.tell(msg, sender()) // forward sender reference, enabling direct reply
  *     case Request3(msg) =>
  *       implicit val timeout = Timeout(5.seconds)
- *       sender() ! (other ? msg)  // will reply with a Future for holding other's reply
+ *       (other ? msg) pipeTo sender() // will get a Future from other and pipe its value to sender
```
hepin1989: ```from other``` I don;t think so.
the ```ask``` return a ```future``` of the other's reply.and the future is comes from the local ask support actor not the other that is asked.
how about 
```the ask call will get a Future of other's first reply,and then we pipe its value to the sender when it complete.```

gkb: Fixed it.

drewhk:
Does this doc example have a Java counterpart? If yes, then it should be updated, too, and then we can merge this.

gkb:
I updated the Java counterpart as well.

gkb:
Is there anything else this pull request needs before it gets merged?

ktoso:
PLS BUILD

ktoso:
@gkb Thanks for the contrib! It's a nice one. We just need to run PR validation once more, since now our build is using JDK8 and the javadoc generation is a bit stricter. Once it passes validation we'll merge.

Issue #17310, opened 2015-04-24
Provide "Line by Line" either Source or parseLines as built-in op


ktoso:
When updating activator templates we realised `FileSource` is not exactly enough for the common use case, we still need the Cookbook `parseLines` recipe for it to be useful for CSV processing.

We should ship `parseLines` or something similar as built-in - because line-by-line text processing is a super common use case.

patriknw:
we might also want to investigate if a `BufferedReader.readLine()` source is more efficient, but I don't know if there is a corresponding thing for async io

drewhk:
This is somewhat related to the other ticket of framing support https://github.com/akka/akka/issues/15101

PR #17446, opened 2015-05-11
+str #17310: Basic framing support

drewhk:
(Still needs scaladocs and java tests)

### DIFF
--- a/akka-docs-dev/rst/scala/stream-cookbook.rst
+++ b/akka-docs-dev/rst/scala/stream-cookbook.rst
@@ -96,22 +96,8 @@ Parsing lines from a stream of ByteStrings
 characters (or, alternatively, containing binary frames delimited by a special delimiter byte sequence) which
 needs to be parsed.
 
-We express our solution as a :class:`StatefulStage` because it has support for emitting multiple elements easily
-through its ``emit(iterator, ctx)`` helper method. Since an incoming ByteString chunk might contain multiple lines (frames)
-this feature comes in handy.
-
-To create the parser we only need to hook into the ``onPush`` handler. We maintain a buffer of bytes (expressed as
-a :class:`ByteString`) by simply concatenating incoming chunks with it. Since we don't want to allow unbounded size
-lines (records) we always check if the buffer size is larger than the allowed ``maximumLineBytes`` value, and terminate
-the stream if this invariant is violated.
-
-After we updated the buffer, we try to find the terminator sequence as a subsequence of the current buffer. To be
-efficient, we also maintain a pointer ``nextPossibleMatch`` into the buffer so that we only search that part of the
-buffer where new matches are possible.
-
-The search for a match is done in two steps: first we try to search for the first character of the terminator sequence
-in the buffer. If we find a match, we do a full subsequence check to see if we had a false positive or not. The parsing
-logic is recursive to be able to parse multiple lines (records) contained in the decoding buffer.
+The :class:`Framing` helper object contains a convenience method to parse messages from a stream of ``ByteStrings``
+and in particular it has basic support for parsing text lines:
 
 .. includecode:: code/docs/stream/cookbook/RecipeParseLines.scala#parse-lines
 
--- a/akka-stream-tests/src/test/scala/akka/stream/io/FramingSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/io/FramingSpec.scala
@@ -1,0 +1,229 @@
+/**
+ * Copyright (C) 2014-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.io
+
+import java.nio.ByteOrder
+
+import akka.stream.io.Framing.FramingException
+import akka.stream.{ ActorFlowMaterializer, ActorFlowMaterializerSettings }
+import akka.stream.scaladsl._
+import akka.stream.stage.{ TerminationDirective, SyncDirective, Context, PushPullStage }
+import akka.stream.testkit.AkkaSpec
+import akka.util.{ ByteString, ByteStringBuilder }
+
+import scala.collection.immutable
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import scala.concurrent.forkjoin.ThreadLocalRandom
+import scala.util.Random
+
+class FramingSpec extends AkkaSpec {
+
+  val settings = ActorFlowMaterializerSettings(system)
+  implicit val materializer = ActorFlowMaterializer(settings)
+
+  class Rechunker extends PushPullStage[ByteString, ByteString] {
+    private var rechunkBuffer = ByteString.empty
+
+    override def onPush(chunk: ByteString, ctx: Context[ByteString]): SyncDirective = {
+      rechunkBuffer ++= chunk
+      rechunk(ctx)
+    }
+
+    override def onPull(ctx: Context[ByteString]): SyncDirective = {
+      rechunk(ctx)
+    }
+
+    override def onUpstreamFinish(ctx: Context[ByteString]): TerminationDirective = {
+      if (rechunkBuffer.isEmpty) ctx.finish()
+      else ctx.absorbTermination()
+    }
+
+    private def rechunk(ctx: Context[ByteString]): SyncDirective = {
+      if (!ctx.isFinishing && ThreadLocalRandom.current().nextBoolean()) ctx.pull()
+      else {
+        val nextChunkSize =
+          if (rechunkBuffer.isEmpty) 0
+          else ThreadLocalRandom.current().nextInt(0, rechunkBuffer.size + 1)
+        val newChunk = rechunkBuffer.take(nextChunkSize)
+        rechunkBuffer = rechunkBuffer.drop(nextChunkSize)
+        if (ctx.isFinishing && rechunkBuffer.isEmpty) ctx.pushAndFinish(newChunk)
+        else ctx.push(newChunk)
+      }
+    }
+  }
+
+  val rechunk = Flow[ByteString].transform(() ⇒ new Rechunker).named("rechunker")
+
+  "Delimiter bytes based framing" must {
+
+    val delimiterBytes = List("\n", "\r\n", "FOO").map(ByteString(_))
+    val baseTestSequences = List("", "foo", "hello world").map(ByteString(_))
+
+    def completeTestSequences(delimiter: ByteString): immutable.Iterable[ByteString] =
+      for (prefix ← 0 until delimiter.size; s ← baseTestSequences)
+        yield delimiter.take(prefix) ++ s
+
+    "work with various delimiters and test sequences" in {
+      for (delimiter ← delimiterBytes; _ ← 1 to 100) {
+        val f = Source(completeTestSequences(delimiter))
+          .map(_ ++ delimiter)
+          .via(rechunk)
+          .via(Framing.delimiter(delimiter, 256))
+          .grouped(1000)
+          .runWith(Sink.head)
+
+        Await.result(f, 3.seconds) should be(completeTestSequences(delimiter))
+      }
+    }
+
+    "Respect maximum line settings" in {
+      // The buffer will contain more than 1 bytes, but the individual frames are less
+      Await.result(
+        Source.single(ByteString("a\nb\nc\nd\n")).via(Framing.lines("\n", 1)).grouped(100).runWith(Sink.head),
+        3.seconds) should ===(List("a", "b", "c", "d"))
+
+      an[FramingException] should be thrownBy {
+        Await.result(
+          Source.single(ByteString("ab\n")).via(Framing.lines("\n", 1)).grouped(100).runWith(Sink.head),
+          3.seconds)
+      }
+    }
+
+    "work with empty streams" in {
+      Await.result(
+        Source.empty.via(Framing.lines("\n", 256)).runFold(Vector.empty[String])(_ :+ _),
+        3.seconds) should ===(Vector.empty)
+    }
+
+    "report truncated frames" in {
+      an[FramingException] should be thrownBy {
+        Await.result(
+          Source.single(ByteString("I have no end"))
+            .via(Framing.lines("\n", 256, allowTruncation = false))
+            .grouped(1000)
+            .runWith(Sink.head),
+          3.seconds)
+      }
+    }
+
+    "allow truncated frames if configured so" in {
+      Await.result(
+        Source.single(ByteString("I have no end"))
+          .via(Framing.lines("\n", 256, allowTruncation = true))
+          .grouped(1000)
+          .runWith(Sink.head),
+        3.seconds) should ===(List("I have no end"))
+    }
+
+  }
+
+  "Length field based framing" must {
+
+    val referenceChunk = ByteString(scala.util.Random.nextString(0x100001))
+
+    val byteOrders = List(ByteOrder.BIG_ENDIAN, ByteOrder.LITTLE_ENDIAN)
+    val frameLengths = List(0, 1, 2, 3, 0xFF, 0x100, 0x101, 0xFFF, 0x1000, 0x1001, 0xFFFF, 0x100001)
+    val fieldLengths = List(1, 2, 3, 4)
+    val fieldOffsets = List(0, 1, 2, 3, 15, 16, 31, 32, 44, 107)
+
+    def encode(payload: ByteString, fieldOffset: Int, fieldLength: Int, byteOrder: ByteOrder): ByteString = {
+      val header = {
+        val h = (new ByteStringBuilder).putInt(payload.size)(byteOrder).result()
+        byteOrder match {
+          case ByteOrder.LITTLE_ENDIAN ⇒ h.take(fieldLength)
+          case ByteOrder.BIG_ENDIAN    ⇒ h.drop(4 - fieldLength)
+        }
+      }
+
+      ByteString(Array.ofDim[Byte](fieldOffset)) ++ header ++ payload
+    }
+
+    "work with various byte orders, frame lengths and offsets" in {
+      for {
+        _ ← 1 to 10
+        byteOrder ← byteOrders
+        fieldOffset ← fieldOffsets
+        fieldLength ← fieldLengths
+      } {
+
+        val encodedFrames = frameLengths.filter(_ < (1 << (fieldLength * 8))).map { length ⇒
+          val payload = referenceChunk.take(length)
+          encode(payload, fieldOffset, fieldLength, byteOrder)
+        }
+
+        Await.result(
+          Source(encodedFrames)
+            .via(rechunk)
+            .via(Framing.lengthField(fieldLength, fieldOffset, Int.MaxValue, byteOrder))
+            .grouped(10000)
+            .runWith(Sink.head),
+          3.seconds) should ===(encodedFrames)
+      }
+
+    }
+
+    "work with empty streams" in {
+      Await.result(
+        Source.empty.via(Framing.lengthField(4, 0, Int.MaxValue, ByteOrder.BIG_ENDIAN)).runFold(Vector.empty[ByteString])(_ :+ _),
+        3.seconds) should ===(Vector.empty)
+    }
+
+    "report oversized frames" in {
+      an[FramingException] should be thrownBy {
+        Await.result(
+          Source.single(encode(referenceChunk.take(100), 0, 1, ByteOrder.BIG_ENDIAN))
+            .via(Framing.lengthField(1, 0, 99, ByteOrder.BIG_ENDIAN)).runFold(Vector.empty[ByteString])(_ :+ _),
+          3.seconds)
+      }
+
+      an[FramingException] should be thrownBy {
+        Await.result(
+          Source.single(encode(referenceChunk.take(100), 49, 1, ByteOrder.BIG_ENDIAN))
+            .via(Framing.lengthField(1, 0, 100, ByteOrder.BIG_ENDIAN)).runFold(Vector.empty[ByteString])(_ :+ _),
+          3.seconds)
+      }
+    }
+
+    "report truncated frames" in {
+      for {
+        //_ ← 1 to 10
+        byteOrder ← byteOrders
+        fieldOffset ← fieldOffsets
+        fieldLength ← fieldLengths
+        frameLength ← frameLengths if frameLength < (1 << (fieldLength * 8)) && (frameLength != 0)
+      } {
+
+        val fullFrame = encode(referenceChunk.take(frameLength), fieldOffset, fieldLength, byteOrder)
+        val partialFrame = fullFrame.dropRight(1)
+
+        an[FramingException] should be thrownBy {
+          Await.result(
+            Source(List(fullFrame, partialFrame))
+              .via(rechunk)
+              .via(Framing.lengthField(fieldLength, fieldOffset, Int.MaxValue, byteOrder))
+              .grouped(10000)
+              .runWith(Sink.head),
+            3.seconds)
+        }
+      }
+    }
+
+    "support simple framing adapter" in {
+      val rechunkBidi = BidiFlow.wrap(rechunk, rechunk)(Keep.left)
+      val codecFlow =
+        Framing.simpleFramingProtocol(1024)
+          .atop(rechunkBidi)
+          .atop(Framing.simpleFramingProtocol(1024).reversed)
+          .join(Flow[ByteString]) // Loopback
+
+      val testMessages = List.fill(100)(referenceChunk.take(Random.nextInt(1024)))
+      Await.result(
+        Source(testMessages).via(codecFlow).grouped(1000).runWith(Sink.head),
+        3.seconds) should ===(testMessages)
+    }
+
+  }
+
+}
--- a/akka-stream/src/main/scala/akka/stream/io/Framing.scala
+++ b/akka-stream/src/main/scala/akka/stream/io/Framing.scala
@@ -1,0 +1,285 @@
+/**
+ * Copyright (C) 2014-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.io
+
+import java.nio.ByteOrder
+
+import akka.stream.scaladsl.{ Keep, BidiFlow, Flow }
+import akka.stream.stage._
+import akka.util.{ ByteIterator, ByteStringBuilder, ByteString }
+
+import scala.annotation.tailrec
+
+object Framing {
+
+  /**
+   * Creates a Flow that handles decoding a stream of unstructured byte chunks into a stream of frames where the
+   * incoming chunk stream uses a specific byte-sequence to mark frame boundaries.
+   *
+   * The decoded frames will include the separator sequence. If this is not desired, this Flow can be augmented with a
+   * simple ''map'' operation that removes this separator.
+   *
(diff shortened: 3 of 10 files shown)


drewhk:
Refs #17310

akka-stream-tests/src/test/scala/akka/stream/io/FramingSpec.scala
```diff
@@ -0,0 +1,182 @@
+package akka.stream.io
```
ktoso: license header

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -0,0 +1,173 @@
...
+
+import scala.annotation.tailrec
+
+object Framing {
+
+  def delimiter(delimiterBytes: ByteString, maximumFrameLength: Int): Flow[ByteString, ByteString, Unit] =
+    Flow[ByteString].transform(() ⇒ new DelimiterFramingStage(delimiterBytes, maximumFrameLength))
+      .named("delimiterFraming")
+
+  def lines(delimiter: String, maximumLineBytes: Int): Flow[ByteString, String, Unit] =
```
ktoso: maybe `def lines(maximumLineBytes: Int, delimiter: String = "\n")` as a reasonable default?

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -0,0 +1,173 @@
...
+    Framing.delimiter(ByteString(delimiter), maximumLineBytes).map(_.utf8String)
+      .named("lineParsing")
```
ktoso: Parsing or Framing?

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -0,0 +1,173 @@
...
+        }
+      }
+    }
+
+  }
+
+  class LengthFieldFramingStage(
+    val lengthFieldLength: Int,
+    val lengthFieldOffset: Int,
+    val byteOrder: ByteOrder) extends PushPullStage[ByteString, ByteString] {
```
hepin1989: shoud add ```lengthIncludesLengthFieldLength:Boolean```?

akka-stream/src/main/scala/akka/stream/io/Framing.scala
drewhk: for lines it can be called parsing I think.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
drewhk: I am not sure. I rather make people aware that it can decode only one kind of line terminators, and they should specify that terminator explicitly.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
drewhk: hm, that might be a good idea.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -0,0 +1,173 @@
...
+    private var buffer = ByteString.empty
+    private val minimumChunkSize = lengthFieldOffset + lengthFieldLength
+    private val intDecoder: IntDecoder = byteOrder match {
+      case ByteOrder.BIG_ENDIAN    ⇒ new BigEndianCodec(lengthFieldLength)
+      case ByteOrder.LITTLE_ENDIAN ⇒ new LittleEndianCodec(lengthFieldLength)
+    }
+    private var frameSize = Int.MaxValue
```
hepin1989: should limit the max length of the frame.

hepin1989: I think there should be a configurable strategy for the max frame size exceed .and so should have a ```failFast:Boolean```
when false,fail when the frame is exceed.when true ,fail when the header is exceed.

drewhk: A maximum frame setting makes sense.

> when false,fail when the frame is exceed.when true ,fail when the header is exceed.

I don't understand what you mean here. The header encodes the length of the frame, the frame itself cannot exceed any limit. You meant dropping large frames instead of failing the stream?

I don't want to complicate this needlessly. If someone needs a richer functionality, he can build the stage himself.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
drewhk: I would rather take a function Int => Int so the user can tune the length field decoding as much as he wants.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
hepin1989: no what I means is if we set the ```failFast``` to true,this stage will ```fail``` ASAP it notice that the ```length``` of the frame will exceed.but when we set to ```false ``` will means we may do something like ```dropping``` or ```fail``` when the the maxLength of the frame is read. 

And you are right.we could build more complicated one.and you just show the path.:P

hepin1989: for the maximum frame size,cause the maybe someone trying to attach you.

drewhk: > will means we may do something like dropping or fail when the the maxLength of the frame is read.

But that is impossible. We dont "read", but we receive arbitrary many bytes from the upstream. That chunk might contain many valid frames not just one. If someone wants to limit the chunkSize, he can always put such stage before this one.

hepin1989: thanks,I am not that ```stream``` yet.:P

akka-stream/src/main/scala/akka/stream/io/Framing.scala
ktoso: Hm I see what you mean.
My worry comes from the fast that `lines` and `delimiter` now are pretty much the same, with the exception of output type - ByteString/String. I can just as well use delimiter for "lines", so I thought that `lines` would be the "I'm lazy, give me lines split by \n" method.

Not a big issue, but got me thinking.

drewhk: Line endings are always OS specific, what I can do though is to use the OS default as the default value.

ktoso: Agreed, that's a good default.

drewhk: btw, they are of course pretty much the same :) This is just a convenience method to avoid all the boilerplate of creating ByteStrings, or converting them to Strings.

akka-stream-tests/src/test/scala/akka/stream/io/FramingSpec.scala
```diff
@@ -0,0 +1,182 @@
...
+      }
+    }
+
+    "Respect maximum line settings" in {
+      // The buffer will contain more than 1 bytes, but the individual frames are less
+      Await.result(
+        Source.single(ByteString("a\nb\nc\nd\n")).via(Framing.lines("\n", 1)).grouped(100).runWith(Sink.head),
+        3.seconds) should ===(List("a", "b", "c", "d"))
+
+      an[IllegalStateException] should be thrownBy {
```
patriknw: is it `IllegalStateException` or `IllegalArgumentException`?
I think it looks like invalid input or invalid settings, i.e. `IllegalArgumentException`.

akka-stream-tests/src/test/scala/akka/stream/io/FramingSpec.scala
```diff
@@ -0,0 +1,182 @@
...
+        byteOrder ← byteOrders
+        fieldOffset ← fieldOffsets
+        fieldLength ← fieldLengths
+        frameLength ← frameLengths if frameLength < (1 << (fieldLength * 8)) && (frameLength != 0)
+      } {
+
+        val fullFrame = encode(referenceChunk.take(frameLength), fieldOffset, fieldLength, byteOrder)
+        val partialFrame = fullFrame.dropRight(1)
+
+        an[IllegalStateException] should be thrownBy {
```
patriknw: IllegalStateException?

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -0,0 +1,173 @@
+package akka.stream.io
```
patriknw: copyright

akka-stream-tests/src/test/scala/akka/stream/io/FramingSpec.scala
drewhk: Why? The stream terminated early, isn't that an illegal state? There was no invalid argument provided.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -0,0 +1,173 @@
...
+
+import akka.stream.scaladsl.Flow
+import akka.stream.stage.{ TerminationDirective, PushPullStage, SyncDirective, Context }
+import akka.util.{ ByteIterator, ByteStringBuilder, ByteString }
+
+import scala.annotation.tailrec
+
+object Framing {
+
+  def delimiter(delimiterBytes: ByteString, maximumFrameLength: Int): Flow[ByteString, ByteString, Unit] =
```
patriknw: `delimiterBytes: ByteString` is not good parameter name. `def delimiter(delimiter: ByteString, ` might look funny. `separator`?

akka-stream/src/main/scala/akka/stream/io/Framing.scala
patriknw: I'm not sure that it is good to provide a default that will make things break across two machines. Then it is better to make it explicit.

patriknw: I guess it is difficult to support maximumLineLength in number of chars?

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -0,0 +1,173 @@
...
+          tryPull(ctx)
+        } else if (possibleMatchPos + separatorBytes.size > buffer.size) {
+          // We have found a possible match (we found the first character of the terminator
+          // sequence) but we don't have yet enough bytes. We remember the position to
+          // retry from next time.
+          nextPossibleMatch = possibleMatchPos
+          tryPull(ctx)
+        } else {
+          if (buffer.slice(possibleMatchPos, possibleMatchPos + separatorBytes.size)
+            == separatorBytes) {
```
patriknw: keep with previous line?

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -0,0 +1,173 @@
...
+            nextPossibleMatch += 1
+            doParse(ctx)
+          }
+        }
+      }
+    }
+
+  }
+
+  class LengthFieldFramingStage(
```
patriknw: public?

patriknw:
looking good

akka-stream-tests/src/test/scala/akka/stream/io/FramingSpec.scala
patriknw: I was thinking that the stream failed because the input to the stream was invalid.
I'm not sure what we normally use. Perhaps look at what is used in http for unmarshalling (e.g. invalid json).

drewhk: Yes, the input is invalid, but that input is not an argument. You can also argue that closing a stream is a valid operation, but not in that *state* :)

akka-stream/src/main/scala/akka/stream/io/Framing.scala
drewhk: Hmm, I thought Framing.delmiter(delimiterBytes = ...) is the cleanest. I want to use delmiter framing, using the bytes passed as the delimiter. Why call it separator, when the approach is called delimiter based? The alternative is to rename this method to delimiterBased.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
drewhk: I don't think that gains much. counting characters can be more expensive with utf8, and would mean that we cannot reuse the delmiter framing anymore as an implementation. If someone needs more control, he can write his own stage.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
drewhk: ok

akka-stream/src/main/scala/akka/stream/io/Framing.scala
drewhk: should be private

akka-stream/src/main/scala/akka/stream/io/Framing.scala
patriknw: It's just that encoding the type in the name is not generally recommended, e.g. you would not write `maximumFrameLengthInt`, and in another place you have used `maximumLineBytes` which refers to the size in bytes and not the ByteString as type. Naming is hard.

akka-stream-tests/src/test/scala/akka/stream/io/FramingSpec.scala
```diff
@@ -0,0 +1,182 @@
...
+        val nextChunkSize =
+          if (rechunkBuffer.isEmpty) 0
+          else ThreadLocalRandom.current().nextInt(0, rechunkBuffer.size + 1)
+        val newChunk = rechunkBuffer.take(nextChunkSize)
+        rechunkBuffer = rechunkBuffer.drop(nextChunkSize)
+        if (ctx.isFinishing && rechunkBuffer.isEmpty) ctx.pushAndFinish(newChunk)
+        else ctx.push(newChunk)
+      }
+    }
+  }
```
rkuhn: This might make a good addition to the StreamTestKit.

akka-stream-tests/src/test/scala/akka/stream/io/FramingSpec.scala
rkuhn: The deciding question for me would be: “will this input always fail?” If yes, then it is an illegal argument, otherwise it depends on an illegal state.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
drewhk: Ok, I rename the argument to delimiter.

akka-stream-tests/src/test/scala/akka/stream/io/FramingSpec.scala
drewhk: A broken TCP connection for example cannot be considered an illegal argument since it is not known whether it is broken or not upfront.

rkuhn: My point was just that I needed to write down the difference between these two exception types for my own sanity’s sake—I think any kind of parsing error should be its own exception type (there might already be some data format errors defined in the JDK).

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -0,0 +1,173 @@
...
+  def lines(delimiter: String, maximumLineBytes: Int): Flow[ByteString, String, Unit] =
+    Framing.delimiter(ByteString(delimiter), maximumLineBytes).map(_.utf8String)
+      .named("lineParsing")
+
+  def lenghtField(fieldLength: Int, fieldOffset: Int = 0, byteOrder: ByteOrder = ByteOrder.LITTLE_ENDIAN): Flow[ByteString, ByteString, Unit] =
+    Flow[ByteString].transform(() ⇒ new LengthFieldFramingStage(fieldLength, fieldOffset, byteOrder))
+      .named("lengthFieldFraming")
+
+  private trait IntDecoder {
+    def decode(bs: ByteIterator): Int
```
drewhk: I was thinking of making this a Long, but I think it is fair just to throw on overflow in the parser. I don't think this stage is very suitable for gigabyte large messages anyway :)

drewhk: (of course I need to add the overflow check)

akka-stream-tests/src/test/scala/akka/stream/io/FramingSpec.scala
drewhk: I will define a FramingException and use it everywhere. I guess that is the cleanest.

akka-stream-tests/src/test/scala/akka/stream/io/FramingSpec.scala
```diff
@@ -0,0 +1,182 @@
...
+    "work with empty streams" in {
+      Await.result(
+        Source.empty.via(Framing.lines("\n", 256)).runFold(Vector.empty[String])(_ :+ _),
+        3.seconds) should ===(Vector.empty)
+    }
+
+    "report truncated frames" in {
+      an[IllegalStateException] should be thrownBy {
+        Await.result(
+          Source.single(ByteString("I have no end")).via(Framing.lines("\n", 256)).grouped(1000).runWith(Sink.head),
```
rkuhn: :-)

Sidenote: this needs to be prominently documented, including how to ignore this condition (either by adding a configuration flag or by documenting that a delimiter should be always appended to ensure proper termination when desired).

akka-stream/src/main/scala/akka/stream/io/Framing.scala
rkuhn: The JDK does not have support for counting characters correctly, so bytes are the only supported measure here. I also suggest to not pick any defaults, and especially not a default that is platform dependent.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
rkuhn: agreed

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -0,0 +1,173 @@
...
+        decoded <<= 8
+        decoded += bs.next().toInt & 0xFF
+        count -= 1
+      }
+      decoded
+    }
+  }
+
+  private class LittleEndianCodec(val length: Int) extends IntDecoder {
+    private val highestNibble = (length - 1) * 8
```
rkuhn: hmm, a nibble is a half-octet

akka-stream-tests/src/test/scala/akka/stream/io/FramingSpec.scala
patriknw: +1

akka-stream/src/main/scala/akka/stream/io/Framing.scala
drewhk: ah, you noticed :) I hoped I can fix it in the next commit without anyone noticing :D

akka-stream/src/main/scala/akka/stream/io/Framing.scala
drewhk: I don't think I want to create a Parsing object just for this one Flow.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -0,0 +1,173 @@
...
+    def decode(bs: ByteIterator): Int
+  }
+
+  private class BigEndianCodec(val length: Int) extends IntDecoder {
+    override def decode(bs: ByteIterator): Int = {
+      var count = length
+      var decoded = 0
+      while (count > 0) {
+        decoded <<= 8
+        decoded += bs.next().toInt & 0xFF
```
rkuhn: I doesn’t really make a difference in practice, but for these bit operations I prefer to use `|=` for consistency.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -0,0 +1,173 @@
...
+          // retry from next time.
+          nextPossibleMatch = possibleMatchPos
+          tryPull(ctx)
+        } else {
+          if (buffer.slice(possibleMatchPos, possibleMatchPos + separatorBytes.size)
+            == separatorBytes) {
+            // Found a match
+            val parsedFrame = buffer.slice(0, possibleMatchPos).compact
+            buffer = buffer.drop(possibleMatchPos + separatorBytes.size)
+            nextPossibleMatch -= possibleMatchPos + separatorBytes.size
```
rkuhn: we just dropped the beginning of the buffer which was a full frame, so this should probably restart at zero

akka-stream-tests/src/test/scala/akka/stream/io/FramingSpec.scala
drewhk: True, I add a flag wether the last fragment should be emitted on truncation or the fail the stream.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
drewhk: yes, right:)

rkuhn:
LGTM so far; these will then also need sugar BidiFlow elements that combine encoding and decoding in a nice and easy package.

drewhk:
I am not sure about the encoding part. For the delmiter case the only encoding needed is to add the delimiter itself, which is trivial, and in some cases there is no need to add the delmiter at all because it is already present. For the length-field case, we can provide a simple encoder, but many times these frames are more complex packets, the decoding stage is only to simplify the further parsing stages. In those cases the encoding part usually involves constructing the whole message.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
rkuhn: … in which case the name should be “lineFraming” ;-) (I agreed that technically we could call this parsing, but we don’t have to)

rkuhn:
Yes, of course. But when writing your own protocol and you just want frames over TCP then there should be a nice out-of-the-box BidiFlow for that, just like we had for the old Pipelines.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
drewhk: meh, I change it then

drewhk:
True, I can add a sugar for that case. No fancy configuration settings, only a maximum length bound, and just doing a 4-byte header and be happy with it.

rkuhn:
Yup, exactly.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -0,0 +1,173 @@
...
+
+  def delimiter(delimiterBytes: ByteString, maximumFrameLength: Int): Flow[ByteString, ByteString, Unit] =
+    Flow[ByteString].transform(() ⇒ new DelimiterFramingStage(delimiterBytes, maximumFrameLength))
+      .named("delimiterFraming")
+
+  def lines(delimiter: String, maximumLineBytes: Int): Flow[ByteString, String, Unit] =
+    Framing.delimiter(ByteString(delimiter), maximumLineBytes).map(_.utf8String)
+      .named("lineParsing")
+
+  def lenghtField(fieldLength: Int, fieldOffset: Int = 0, byteOrder: ByteOrder = ByteOrder.LITTLE_ENDIAN): Flow[ByteString, ByteString, Unit] =
```
jrudolph: s/lenghtField/lengthField/

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -0,0 +1,173 @@
...
+import scala.annotation.tailrec
+
+object Framing {
+
+  def delimiter(delimiterBytes: ByteString, maximumFrameLength: Int): Flow[ByteString, ByteString, Unit] =
+    Flow[ByteString].transform(() ⇒ new DelimiterFramingStage(delimiterBytes, maximumFrameLength))
+      .named("delimiterFraming")
+
+  def lines(delimiter: String, maximumLineBytes: Int): Flow[ByteString, String, Unit] =
+    Framing.delimiter(ByteString(delimiter), maximumLineBytes).map(_.utf8String)
```
jrudolph: Utf8 is a variable length encoding, so `utf8String` doesn't work on arbitrary chunks of ByteStrings. I had to implement proper stream UTF8-encoding/decoding so I guess it would make sense to move `akka.http.impl.engine.ws.Utf8Decoder` to streams and also provide a streaming interface for it which should be trivial. It currently doesn't have a streaming interface because for websocket I had to embed it inside of another stage.

The same is true for the other direction. An `Array[Char]` is a sequence of UTF-16 code units (called "character" in java for legacy reasons). UTF-16 is a variable length encoding itself to encode 32-bit Unicode code points. So, an arbitrarily chunked stream of `Array[Char]` cannot be naively converted to Unicode code points because code points could be split in half.

akka-stream/src/main/scala/akka/stream/io/Framing.scala
jrudolph: +1 and also needs a check above for restricting the `fieldLength` parameter to 2^31 above.

jrudolph: I meant to 31 bits, i.e. 1 <= fieldLength <= 4.

drewhk: Thank, true!

akka-stream/src/main/scala/akka/stream/io/Framing.scala
drewhk: I rely on the delimiter not being part of a valid codepoint, that is true. I don't want to overcomplicate this for now, so I will document the limitation and we can make it smarter later.

jrudolph: Yes, I agree, assuming UTF8 encoding and not splitting on half codepoints may be good enough for now.

rkuhn:
This looks ripe for just pushing over the line, perhaps @2m can help?

drewhk:
I am happy to finish, but I want to look a bit more into the quarantine bug. So if @2m has some time, feel free to grab it and give it a final push.

2m:
I can take this now.

drewhk:
No need anymore

2m:
Yup. I just saw the validator.

drewhk:
you can make a final review if you want though.

akka-docs-dev/rst/scala/stream-cookbook.rst
```diff
@@ -96,22 +96,8 @@ Parsing lines from a stream of ByteStrings
...
-the stream if this invariant is violated.
-
-After we updated the buffer, we try to find the terminator sequence as a subsequence of the current buffer. To be
-efficient, we also maintain a pointer ``nextPossibleMatch`` into the buffer so that we only search that part of the
-buffer where new matches are possible.
-
-The search for a match is done in two steps: first we try to search for the first character of the terminator sequence
-in the buffer. If we find a match, we do a full subsequence check to see if we had a false positive or not. The parsing
-logic is recursive to be able to parse multiple lines (records) contained in the decoding buffer.
+The :class:`Framing` helper object contains a convenience method to parse messages from a stream of ``ByteStrings``
```
2m: s/method/flow

2m:
LGTM. Very nice!

Issue #17454, opened 2015-05-12
Revisit settings for cluster-tools


patriknw:
Default values should be in config. The number of parameters for props factories have grown beyond manageable.

Issue #13909, opened 2014-03-05
Remote death watch is wrong tool for the job in ClusterClient


patriknw:
imported from https://www.assembla.com/spaces/akka/tickets/3909

For cluster clients it should be possible to re-establish a connection after a long network partition, without restarting the system.

PR #17455, opened 2015-05-12
!cto #17454 Add/change settings for cluster-tools

patriknw:
Default values should be in config. The number of parameters for props factories have grown beyond manageable.

Using same structure as the settings in Akka Streams, i.e. can be created from system/config, and amended by `withX` methods.

### DIFF
--- a/akka-cluster-tools/src/main/resources/reference.conf
+++ b/akka-cluster-tools/src/main/resources/reference.conf
@@ -6,7 +6,7 @@
 # Make your edits/overrides in your application.conf.
 
 # //#pub-sub-ext-config
-# Settings for the DistributedPubSubExtension
+# Settings for the DistributedPubSub extension
 akka.cluster.pub-sub {
   # Actor name of the mediator actor, /user/distributedPubSubMediator
   name = distributedPubSubMediator
--- a/akka-cluster-tools/src/main/resources/reference.conf
+++ b/akka-cluster-tools/src/main/resources/reference.conf
@@ -47,7 +47,7 @@ akka.actor {
 
 
 # //#receptionist-ext-config
-# Settings for the ClusterReceptionistExtension
+# Settings for the ClusterClientReceptionist extension
 akka.cluster.client.receptionist {
   # Actor name of the ClusterReceptionist actor, /user/receptionist
   name = receptionist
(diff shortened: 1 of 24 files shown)


patriknw:
When this has been merged I can unlock the [watch issue in ClusterClient](https://github.com/akka/akka/issues/13909).

patriknw:
that was because of jdk8, I have changed it back

patriknw:
PLS BUILD

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -264,48 +314,93 @@ class ClusterReceptionistExtension(system: ExtendedActorSystem) extends Extensio
...
+   * Java API: Create settings from a configuration with the same layout as
+   * the default configuration `akka.cluster.client.receptionist`.
+   */
+  def create(config: Config): ClusterReceptionistSettings = apply(config)
+
+  /**
+   * INTERNAL API
+   */
+  private[akka] def roleOption(role: String): Option[String] = role match {
+    case null | "" ⇒ None
```
drewhk: Option(null) gives None,  so the below line can be Some(role), and you only need to check for "".

akka-cluster-tools/src/main/resources/reference.conf
```diff
@@ -75,3 +93,44 @@ akka.cluster.client {
 # //#cluster-client-mailbox-config
 
 
+akka.cluster.singleton.default {
```
drewhk: why the .default suffix?

drewhk:
Is this binary compatible?

rkuhn:
none of this is BC, package names and configuration paths have changed

drewhk:
and factory methods/constructors

akka-cluster-tools/src/main/resources/reference.conf
patriknw: You can have many singletons, and this only defines the settings that are used if you create those settings via `ClusterSingletonManagerSettings(system)`, but I think you might be right that it will make sense anyway without the `.default`. I'll take a look at that.

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
patriknw: thanks

patriknw:
We are breaking source compatibility of this when moving it out from contrib and take the opportunity to clean up before making it non-experimental. Those breaking things are supposed to be mentioned in the migration guide.

drewhk:
Ok, makes sense

drewhk:
LGTM

patriknw:
rebased, any more reviewers here?

patriknw:
@rkuhn Here is a real javadoc failure: https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/2687/consoleFull

How do I find what is the problem? What do I search for in the log?

patriknw:
The problem was wrong link to akka.cluster.pubsub.DistributedPubSubExtension. How could I find that from the log?

I found it by running scaladoc

patriknw:
Refs #17454

rkuhn:
Finding these errors in the log is unfortunately a bit difficult within the browser: you’d need to grep for `\.java:[0-9]*: error:` (since `error:` alone is way too prolific otherwise).

Issue #15331, opened 2014-06-03
allow ShardAllocationStrategy to obtain results asynchronously


rkuhn:
It would make sense to query an Actor while trying to decide on the policy, and forcing that to be synchronous is not a good idea. There is no a priori reason why the allocation decision cannot be handed off.

ktoso:
+1, we've bumped into the need of this recently "in the wild".

Issue #16050, opened 2014-10-07
allow graceful ShardRegion shutdown


rkuhn:
For maintenance purposes it may be necessary to passivate all entries on a given node; this is internally implemented for HandOff, but currently not exposed to the user.

dragisak:
+1

hannesstockner:
+1

prettynatty:
:+1:

dcaoyuan:
+1

anderius:
+1

patriknw:
Hi all supporters,

I invite you to take a look at the [pull request](https://github.com/akka/akka/issues/17527) for this feature and verify that the functionality match your expectations.

anderius:
Sorry if it is obvious somewhere, but is this now due for the next milestone release?

I am eager to test it out! :-)

patriknw:
It will be in 2.4-M2, but that is a few weeks away, due to ScalaDays.
You can try it out with tomorrow's timestamped snapshot, you will find it here: http://repo.akka.io/snapshots/com/typesafe/akka/akka-cluster-sharding_2.11/

PR #17456, opened 2015-05-12
!con #15331 Support async allocateShard and rebalance

patriknw:


patriknw:
Refs #15331

patriknw:
When this has been merged I can unlock the highly requested [graceful shard region shutdown](https://github.com/akka/akka/issues/16050).

patriknw:
jdk8 failure

patriknw:
PLS BUILD

akka-cluster-sharding/src/test/scala/akka/cluster/sharding/LeastShardAllocationStrategySpec.scala
```diff
@@ -18,7 +18,7 @@ class LeastShardAllocationStrategySpec extends AkkaSpec {
   "LeastShardAllocationStrategy" must {
     "allocate to region with least number of shards" in {
       val allocations = Map(regionA -> Vector("shard1"), regionB -> Vector("shard2"), regionC -> Vector.empty)
-      allocationStrategy.allocateShard(regionA, "shard3", allocations) should ===(regionC)
+      allocationStrategy.allocateShard(regionA, "shard3", allocations).value.get.get should ===(regionC)
```
drewhk: why not Await? Isn't the current approach depending on already completed Futures?

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ClusterSharding.scala
```diff
@@ -1723,14 +1750,22 @@ class ShardCoordinator(handOffTimeout: FiniteDuration, shardStartTimeout: Finite
...
+            // continue when future is completed
+            shardsFuture.map { shards ⇒ RebalanceResult(shards)
+            }.recover {
+              case _ ⇒ RebalanceResult(Set.empty)
+            }.pipeTo(self)
         }
+      }
+
+    case RebalanceResult(shards) ⇒
+      continueRebalance(shards)
```
drewhk: what happens if there are multiple rebalance Futures in flight? The ordering will not be guaranteed between them, can that be a problem? (I looked at the tests but I am unsure if this scenario is covered or not)

drewhk:
LGTM if the "racing rebalances" case is covered

akka-cluster-sharding/src/test/scala/akka/cluster/sharding/LeastShardAllocationStrategySpec.scala
patriknw: yes, and `LeastShardAllocationStrategy` is supposed to be completed immediately. I can change to Await, though.

drewhk: Yes, the current ones will work with the above test. It will be just surprising if in the future someone changes the impl to be properly async and then suddenly these test cases fail. Maybe not a big issue, though.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ClusterSharding.scala
```diff
@@ -1778,4 +1813,39 @@ class ShardCoordinator(handOffTimeout: FiniteDuration, shardStartTimeout: Finite
 
   def allocateShardHomes(): Unit = persistentState.unallocatedShards.foreach { self ! GetShardHome(_) }
 
+  def continueGetShardHome(shard: ShardId, region: ActorRef, getShardHomeSender: ActorRef): Unit =
+    if (!rebalanceInProgress.contains(shard)) {
+      persistentState.shards.get(shard) match {
+        case Some(ref) ⇒ getShardHomeSender ! ShardHome(shard, ref)
```
patriknw: if concurrent `allocateShard` are performed for the same `shard` the first one will be handled in the `None` case below and subsequent results will end in this `Some` case, i.e. use the first decision

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ClusterSharding.scala
```diff
@@ -1778,4 +1813,39 @@ class ShardCoordinator(handOffTimeout: FiniteDuration, shardStartTimeout: Finite
...
+            }
+          } else
+            log.debug("Allocated region {} for shard [{}] is not (any longer) one of the registered regions: {}",
+              region, shard, persistentState)
+      }
+    }
+
+  def continueRebalance(shards: Set[ShardId]): Unit =
+    shards.foreach { shard ⇒
+      if (!rebalanceInProgress(shard)) {
```
patriknw: concurrent rebalance will be handled by this check, i.e. only first one will trigger a rebalance

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ClusterSharding.scala
patriknw: I have added two comments below that I think answers that question

akka-cluster-sharding/src/test/scala/akka/cluster/sharding/LeastShardAllocationStrategySpec.scala
patriknw: I agree.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ClusterSharding.scala
drewhk: Thanks for the explanation!

drewhk:
LGTM now for real ;)

patriknw:
any more reviewers here?

patriknw:
rebased

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ClusterSharding.scala
```diff
@@ -1693,20 +1712,29 @@ class ShardCoordinator(handOffTimeout: FiniteDuration, shardStartTimeout: Finite
...
-              persist(ShardHomeAllocated(shard, region)) { evt ⇒
-                persistentState = persistentState.updated(evt)
-                log.debug("Shard [{}] allocated at [{}]", evt.shard, evt.region)
-
-                sendHostShardMsg(evt.shard, evt.region)
-                sender() ! ShardHome(evt.shard, evt.region)
+              val getShardHomeSender = sender()
+              val regionFuture = allocationStrategy.allocateShard(getShardHomeSender, shard, persistentState.regions)
+              regionFuture.value match {
+                case Some(Success(region)) ⇒
```
ktoso: nice trick :)

ktoso:
LGTM

patriknw:
thanks for reviewing

Issue #17465, opened 2015-05-12
In ActorInterpreter / ActorProcessor exceptions are created in postStop that may not be used


jrudolph:
Not sure what the purpose of those is. It seems they are catch-all fallbacks to make sure that downstreams are notified of an unexpected death of the actor backing a Publisher. As `postStop` is also called for the expected shutdown (and usually in much bigger volume) creating an exception including stacktrace is costly if the exception is thrown away right afterwards.

I propose to make the exception extend from `NoStackTrace` (the stack trace from `postStop` is probably not useful anyway) and/or restrict the creation of those exceptions to cases where it is still possible to relay it to downstream (i.e. downstream isn't completed yet).

(This turned up during benchmarking last week with percentages <10% but still significant, so I thought I post it here so it doesn't get lost for later optimization attempts.)

drewhk:
Thanks, makes sense.

PR #17466, opened 2015-05-12
!str #17465: Don't create expensive exceptions for abrupt termination

drewhk:


drewhk:
refs: https://github.com/akka/akka/issues/17465

ktoso:
LGTM and thanks @jrudolph for pointing this out!

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
```diff
@@ -163,6 +165,13 @@ abstract class ActorFlowMaterializer extends FlowMaterializer {
  */
 class MaterializationException(msg: String, cause: Throwable = null) extends RuntimeException(msg, cause)
 
+/**
+ * This exception signals that an actor implementing a Reactive Streams Subscriber, Publisher or Processor
+ * has been terminated without being notified by an onError, onComplete or cancel signal. This usually happens
+ * when an ActorSystem is shut down while stream processing actors are still running.
+ */
+object AbruptTerminationException extends Exception("Processor actor terminated abruptly") with NoStackTrace
```
patriknw: RuntimeException?

patriknw:
LGTM after making it a RuntimeException

akka-stream/src/main/scala/akka/stream/ActorFlowMaterializer.scala
drewhk: true. I think it should also be a class instead of object and contain the path of the actor that throws it

patriknw: yes, that would be useful

Issue #17464, opened 2015-05-12
mapConcat is unnecessary strict with requiring a Seq. An Iterable would be fine, too.


drewhk:

patriknw:
Iterator does not have a bounded size. No problem?

drewhk:
I think it isn't. Since we don't store the thing just pull out values as needed it can even stream an infinite amount of elements if someone desires so. Also, currently Sets don't work although they are strict.

drewhk:
Funny enough, this would make implementation of IterableSource a

    Source.empty.mapConcat(iterable)

:)

PR #17467, opened 2015-05-12
+str #1747: mapConcat accepts immutable.Iterable

drewhk:


drewhk:
I am not sure what type the Java API should accept though.

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -215,36 +215,7 @@ object Source extends SourceApply {
...
-          } else {
-            val elem = iterator.next()
-            if (iterator.hasNext) ctx.push(elem)
-            else ctx.pushAndFinish(elem)
-          }
-        }
-      }
-
-    }).withAttributes(DefaultAttributes.iterableSource)
+    Source.single(()).mapConcat((_: Unit) ⇒ iterable).withAttributes(DefaultAttributes.iterableSource)
```
drewhk: this is somewhat less efficient than before since passes along a dummy token (that amortizes though if the iterable is long enough), but this is more DRY. I am not sure which one to pick.

drewhk:
refs https://github.com/akka/akka/issues/17464

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
patriknw: I guess a benchmark must decide that

patriknw:
java api: `java.lang.Iterable`?
It's not immutable, but might be good enough with documentation similar to Source(iterable): https://github.com/akka/akka/blob/release-2.3-dev/akka-stream/src/main/scala/akka/stream/javadsl/Source.scala#L106-L119

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -378,7 +378,7 @@ trait FlowOps[+Out, +Mat] {
    * '''Cancels when''' downstream cancels
    *
    */
-  def mapConcat[T](f: Out ⇒ immutable.Seq[T]): Repr[T, Mat] = andThen(MapConcat(f.asInstanceOf[Any ⇒ immutable.Seq[Any]]))
+  def mapConcat[T](f: Out ⇒ immutable.Iterable[T]): Repr[T, Mat] = andThen(MapConcat(f.asInstanceOf[Any ⇒ immutable.Iterable[Any]]))
```
drewhk: I will update the docs then

drewhk:
Yes, I think that is OK

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
viktorklang: Perhaps a dumb question but why would you need `.withAttributes(DefaultAttributes.iterableSource)´? (imagine if a user would write: `Source.single(()).mapConcat((_: Unit) ⇒ iterable)`—I submit that they would not know to pass such an attribute)

drewhk: that attribute is just a name attribute, and we collected the default names in DefaultAttributes. I.e. the user will not see the singleton source and the mapConcat as different stages, but it will be packaged up in a composite stage which will display as "iterableSource". The withAttributes is not needed for it to work.

drewhk: basically it is equivalent to:

    ( Source.single(()).mapConcat(_ : Unit) => iterable) ).withAttributes(OperationAttributes.name("iterableSource"))

viktorklang: Ah, it would be super if it was named `iterableSourceName`, that would have saved me from taking your time :)

drewhk: I think the reason is that it might not just be name attributes that go into some stdlib construct, but for example buffer sizes. A Sink.head for example can have apart from a name a buffersize = 1 attribute, too.

viktorklang: Sure, makes sense.
I still worry whether users will understand when to use attributes

-- 
Cheers,
√

hepin1989:
see the document here:
https://github.com/akka/akka/blob/master/akka-cluster/src/main/scala/akka/cluster/Cluster.scala#L272
```
   * JAVA API: Use akka.japi.Util.immutableSeq to convert a java.lang.Iterable
   * to the type needed for the seedNodes parameter.
```
I think how about provide an ```immutableIterable``` ?

patriknw:
What is the relation between cluster seed nodes and stream mapConcat? Stream has a separate javadsl which will have mapConcat returning java.lang.Iterable. Did I misunderstand the question?

hepin1989:
You think that ```java.lang.Iterable``` is mutable,but the scaladsl provide an immutable one,and I notice that in the cluster module ,you provide an wrapper for the Seq,so I think how about provide an ```
akka.japi.Util.immutableIterable``` just as the ```akka.japi.Util.immutableSeq``` you provided for the method ```Cluster.joinSeedNodes```.

patriknw:
@hepin1989 How would such an conversion be implemented? I don't think it is possible to safely convert a (mutable) java.lang.Iterable to and immutable.Iterable. I think we can do it for dedicated API such as this, with documentation about the caveat.

hepin1989:
@patriknw haha,yeah,that's could still be unsafe,just like array.sorry for that.

drewhk:
wat

drewhk:
ok, just mapConcat didn't to pushAndFinish and one test was relying on it. I.e. behavior of iterable source changed (in a completely valid way though) but now I added the pushAndFinish behavior since it is more efficient.

Issue #17445, opened 2015-05-11
FAILED: akka.typed.PerformanceSpec


2m:
https://jenkins.akka.io:8498/job/akka-multi-node/1039/consoleFull

This is after https://github.com/akka/akka/issues/17137 has been fixed.

patriknw:
Looks like it is failing every time in this repeat job: https://jenkins.akka.io:8498/job/akka-local-repeat/

(and that is not with java8)

PR #17476, opened 2015-05-13
=typ #17445 Reduce number of iterations in typed.PerformanceSpec

patriknw:
it was generating a lot of GC, which I think made it fail


patriknw:
Refs #17445

rkuhn:
LGTM

patriknw:
the failure was bin compat in master, but that has been fixed and merged by other pr

Issue #17150, opened 2015-04-08
FAILED: AtLeastOnceDelivery


drewhk:
[info] - must restore state from snapshot *** FAILED ***
    [info]   java.lang.AssertionError: assertion failed: expected ReqAck, found UnconfirmedWarning(Vector(UnconfirmedDelivery(4,akka://AtLeastOnceDeliverySpec/user/$g,Action(4,a-4))))
    [info]   at scala.Predef$.assert(Predef.scala:165)

https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2269/consoleFull

drewhk:
Another one, not sure if it is related

https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2386/consoleFull

[info] - must re-deliver lost messages after restart *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: expected Action(4,a-4), found Action(3,a-3)
[info]   at scala.Predef$.assert(Predef.scala:165)
[info]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:354)
[info]   at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:330)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:738)
[info]   at akka.persistence.AtLeastOnceDeliverySpec$$anonfun$1$$anonfun$apply$mcV$sp$3.apply$mcV$sp(AtLeastOnceDeliverySpec.scala:208)
[info]   at akka.persistence.AtLeastOnceDeliverySpec$$anonfun$1$$anonfun$apply$mcV$sp$3.apply(AtLeastOnceDeliverySpec.scala:190)
[info]   at akka.persistence.AtLeastOnceDeliverySpec$$anonfun$1$$anonfun$apply$mcV$sp$3.apply(AtLeastOnceDeliverySpec.scala:190)
[info]   at org.scalatest.Transformer$$anonfun$apply$1.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer$$anonfun$apply$1.apply(Transformer.scala:22)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   ...

PR #17478, opened 2015-05-13
=per #17150 Harden AtLeastOnceDeliverySpec

patriknw:
- UnconfirmedWarning message spilled over from previous test step,
  using separate probes instead of testActor
- the resend interval was too short so that unexpected resend occured
  as seen by the `failed: expected Action(4,a-4), found Action(3,a-3)`


ktoso:
LGTM

patriknw:
Refs #17150

Issue #16808, opened 2015-02-10
Prevent unnecessary reverse DNS lookups when creating client SSLEngine


jrudolph:
See spray/spray#993.

On the client side the SSLEngine needs to created with the target hostname so that the TLS SNI extension works (starting from Java 7). If `InetSocketAddress.getHostName` is called for this purpose with an IP address, a unnecessary reverse DNS lookup will be initiated. Using `InetSocketAddress.getHostString` would be the right solution but is only available from Java 7, so using reflection may be the best solution.

RichardBradley:
See also #17059 -- please check that HTTP connections also don't do a reverse DNS lookup to generate the `Host` header when connecting to a host by IP address.

agolubev:
Hey there,
I checked DNS lookup problem with HTTP connection. The problem exists for HTTP as well.
Following snippet is used
```scala
val conn=Http().outgoingConnection("74.125.226.70")
(Source.single(HttpRequest()) via conn runWith Sink.head).onComplete(println(_))
```
Issue is in [http-core HttpRequestRendererFactory line 96](https://github.com/akka/akka/blob/releasing-akka-stream-and-http-experimental-1.0-RC1/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala#L96) that's creating Host with InetSocketAddress 
Solution that I can see will be similar to what was done for spray/spray#993. 

Still reflection maybe not the best solution here. Is it an option to change RequestRenderingContext?

jrudolph:
@agolubev I think that's what @RichardBradley also commented on above.

I don't think it would suffice to change `RequestRenderingContext`. Instead, it would be necessary to replace all usages of `InetSocketAddress` from the API-layer up to this place. 

I'd say we can go with the reflective solution for akka 2.3 and when support for Java 1.6 is dropped (which AFAIK will be done with Akka 2.4 which is due soon enough) this hack can be removed again. WDYT?

agolubev:
Ok, let me create reflective fix and I'll do pull request then.
Also will try to reproduce the same scenario with HTTPS. There is no documentation on this point so would appreciate any hint

agolubev:
My bad. I switched SBT to 1.6 in a wrong way. That's why the issue was reopened.
Just made two version of fix (and did two pull requests). They both work with 1.8 and 1.6

jrudolph:
[...]

PR #17481, opened 2015-05-13
Prevent unnecessary reverse DNS lookups #16808

agolubev:


akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
```diff
@@ -44,6 +44,15 @@ private[http] class HttpRequestRendererFactory(userAgentHeader: Option[headers.`
 
       def render(h: HttpHeader) = r ~~ h ~~ CrLf
 
+      val hostString: InetSocketAddress ⇒ String = {
+        // Retrieve the original host string that was given (IP or DNS name) if running Java 7 or later
+        // using the getHostString() method. This avoids a reverse DNS query from calling getHostName()
+        // if the original host string is an IP address.
+        val clazz = classOf[InetSocketAddress]
+        address ⇒ clazz.getMethods.find(_.getName == "getHostString").map(_ invoke address).
```
ryanlecompte: find + map = collectFirst

agolubev:
Ok, just changed it to collectFirst

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
```diff
@@ -44,6 +44,16 @@ private[http] class HttpRequestRendererFactory(userAgentHeader: Option[headers.`
...
 
+      val hostString: InetSocketAddress ⇒ String = {
+        // Retrieve the original host string that was given (IP or DNS name) if running Java 7 or later
+        // using the getHostString() method. This avoids a reverse DNS query from calling getHostName()
+        // if the original host string is an IP address.
+        val clazz = classOf[InetSocketAddress]
+        address ⇒ clazz.getMethods.collectFirst {
+          case m if m.getName == "getHostString" => m.invoke(address)
+        }.getOrElse(address.getHostName).asInstanceOf[String]
+      }
```
viktorklang: With a slight modification it is possible to avoid doingthe method lookup for each call:

~~~scala
val hostString: InetSocketAddress => String =
  try {
    val m = classOf[InetSocketAddress].getDeclaredMethod("getHostString")
    require(m.getReturnType == classOf[String])
    require(m.getParameterCount == 0)
    address => m.invoke(address).asInstanceOf[String]
  } catch {
    case NonFatal(_) =>
    address => address.getHostName()
  }
~~~

agolubev: Yeah, my bad. Also would it be better to move hostString val to one level up - from onPush to HttpRequestRenderer or even to HttpRequestRendererFactory? Just to execute this as less times as possible. 

Also it seems to me that ```require``` statements in try/catch will throw exception the same as ```m.invoke``` (if any) and this exception will be ignored as we'll switch to 1.6 ```getHostName```. So here ```require``` is more for visibility and documentation so I believe a comment with explanation why we are using reflection here should be put instead as it is not common for this project. Does it sound reasonable?

viktorklang: >Yeah, my bad. Also would it be better to move hostString val to one level up - from onPush to HttpRequestRenderer or even to HttpRequestRendererFactory? Just to execute this as less times as possible.

That sounds like a fantastic idea.

> So here require is more for visibility and documentation 

It's primarily there to make sure we get the right `getHostString` method, if `getDeclaredMethod` or any of the `require` throws, then we'll fall back to using `getHostName`.

I considered falling back to `getHostName` whenever the reflective call fails, but I think that's the wrong way to go about it as it could fail for many reasons and it doesn't mean that the `getHostName` would succeed either. Makes sense?

Documenting why reflection is used is a very appropriate thing to do IMO.

hepin1989: should we backport the ```Dns``` to release 2.3?

agolubev: > It's primarily there to make sure we get the right getHostString method


Ok, just validated that ```m.invoke(address)``` is checking for number of parameters in method so require(m.getParameterCount == 0) is redundant from this standpoint. However, return type is not validating by ```invoke``` so we do need fist require call to fall back to ```getHostName```

ktoso:
OK TO TEST

ktoso:
Refs #16808

ktoso:
@viktorklang @agolubev are we sure we want the flow control here based on require and exceptions instead of checking explicitly with ifs?

Other than this question LGTM

viktorklang:
@ktoso What do you propose more technically?

agolubev:
I would say that with reflection we always have some level of uncertainty and to reduce risks we should use try/catch to make it more stable

ktoso:
I was thinking to avoid the try catch as we know what signatures are expected. But you're right we're being unsafe here in any case since it's called with reflection anyway. Never mind then, LGTM :-)

viktorklang:
@ktoso yeah, the reason why I thought `require` was a good fit is because it's about adding verification to the reflective lookup, which in itself may throw exceptions, so having a single fallback codepath seemed like a good thing :)

ktoso:
Right, agree with that sentiment after having rethought it for a moment :)
// Failure was due to our battles with JDK8 PR validation - unrelated to this PR.

ktoso:
Ok, so while code tise it's all good. However I need to ask you to open a new PR and target it to the `release-2.3-dev` branch, which is where streams work is ongoing (current PR is targeting a releasing branch, which we do not merge things into (`releasing-akka-stream-and-http-experimental-1.0-RC1`)

Otherwise LGTM, so let's re-target and get it in :)

agolubev:
Ok, will do PR for release-2.3-dev

Issue #17010, opened 2015-03-05
Akka does not log the error when ActorSystem shuts down due to failed assert


RichardBradley:
We have found that Errors (assertion failures, out of memory errors etc.) can cause Akka 2.3.x to exit without logging any failure info.

Original repro: certain types of TCP connections seem to crash our Akka HTTP 1.0-M3 app -- that is being tracked at #17015 .
This issue tracks the more worrying problem that Akka exits without logging any failure info in that case, making this extremely difficult to track down.

After much struggle, I believe that the issue is that the entire message queue of the `Slf4jLogger` actor is discarded on `ActorSystem` shutdown.
In our case, this includes the enqueued "`ERROR`" log describing the System Guardian actor failure.

I have uploaded a repro project at https://gist.github.com/RichardBradley/f7b1b7c3358779e3559d 
If you run this with "`sbt run`", you will see that "`myApp.log`" has lots of log messages, but that it does not include all log messages that were enqueued, and importantly it does not include the details of the Actor assertion failure which was escalated and caused ActorSystem shutdown.

What can I do to ensure that errors which are fatal to the ActorSystem get logged, if I am using SLF4J / Logback?

With respect, why does Akka have its own logging system based on Actors in the first place? Perhaps Akka should be calling SLF4J directly, using an "`AsyncAppender`" to give non-blocking behaviour? That would solve this issue (the appender's queue is drained on JVM exit by Logback), and perhaps many others.

Thanks,

Rich

RichardBradley:
I’ve pushed a second demo that shows the start of a possible fix: https://gist.github.com/RichardBradley/fc0cc3d312d985e65794 .

If we strip out the Actor subscribers to `LoggingBus` and replace them with code that directly hands off all logging messages to SLF4J as soon as they arrive, then the problem goes away. To avoid blocking I/O on actor threads, we need to add an AsyncAppender to the Logback config XML.

You can run the same demo from above, and see that Logback will drain the AsyncAppender queue on application shutdown. It's also quite a bit faster than the Akka Actor based logging version.

What do you think? Akka could then get rid of a whole load of complexity (like the “startDefaultLoggers” methods in Logging.scala, the custom log-level definitions which duplicate those in SLF4J, etc.). I suppose it would be a massively breaking change.

bantonsson:
Thanks for reporting this issue.

The main reason that akka has its own message bus based logging system is simply to not block when logging. Logging is commonly a very large contention point in a highly concurrent system.

[...]

PR #17490, opened 2015-05-15
=act #17010 Drain log messages on system shutdown

patriknw:
- by using a special mailbox that emits the remaining log messages
  to the StandardOutLogger

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/event/LoggerSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/event/LoggerSpec.scala
@@ -27,6 +27,14 @@ object LoggerSpec {
       }
     """).withFallback(AkkaSpec.testConf)
 
+  val slowConfig = ConfigFactory.parseString("""
+      akka {
+        stdout-loglevel = "ERROR"
+        loglevel = "ERROR"
+        loggers = ["akka.event.LoggerSpec$SlowLogger"]
+      }
+    """).withFallback(AkkaSpec.testConf)
+
   val noLoggingConfig = ConfigFactory.parseString("""
       akka {
         stdout-loglevel = "OFF"
@@ -91,6 +99,19 @@ object LoggerSpec {
     }
   }
 
+  class SlowLogger extends Logging.DefaultLogger {
+    override def aroundReceive(r: Receive, msg: Any): Unit = {
+      msg match {
+        case event: LogEvent ⇒
+          if (event.message.toString.startsWith("msg1"))
+            Thread.sleep(500) // slow
+          super.aroundReceive(r, msg)
+        case _ ⇒ super.aroundReceive(r, msg)
+      }
+
+    }
+  }
+
   class ActorWithMDC extends Actor with DiagnosticActorLogging {
     var reqId = 0
 
@@ -148,6 +169,25 @@ class LoggerSpec extends WordSpec with Matchers {
       val out = createSystemAndLogToBuffer("defaultLogger", defaultConfig, true)
       out.size should be > (0)
     }
+
+    "drain logger queue on system shutdown" in {
+      val out = new java.io.ByteArrayOutputStream()
+      Console.withOut(out) {
+        val sys = ActorSystem("defaultLogger", slowConfig)
+        sys.log.error("msg1")
+        sys.log.error("msg2")
+        sys.log.error("msg3")
+        TestKit.shutdownActorSystem(sys, verifySystemShutdown = true)
+        out.flush()
+        out.close()
+      }
+
+      val logMessages = new String(out.toByteArray).split("\n")
+      logMessages.head should include("msg1")
+      logMessages.last should include("msg3")
+      logMessages.size should ===(3)
+    }
+
   }
 
   "An actor system configured with the logging turned off" must {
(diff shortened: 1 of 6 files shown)


patriknw:
Refs #17010

akka-actor-tests/src/test/scala/akka/event/LoggerSpec.scala
```diff
@@ -148,6 +169,25 @@ class LoggerSpec extends WordSpec with Matchers {
       val out = createSystemAndLogToBuffer("defaultLogger", defaultConfig, true)
       out.size should be > (0)
     }
+
+    "drain logger queue on system shutdown" in {
+      val out = new java.io.ByteArrayOutputStream()
+      Console.withOut(out) {
```
drewhk: wow, good to know this

akka-actor/src/main/scala/akka/event/LoggerMailbox.scala
```diff
@@ -0,0 +1,44 @@
...
+ * INTERNAL API
+ */
+private[akka] class LoggerMailbox(owner: ActorRef, system: ActorSystem)
+  extends UnboundedMailbox.MessageQueue with LoggerMessageQueueSemantics {
+
+  override def cleanUp(owner: ActorRef, deadLetters: MessageQueue): Unit = {
+    if (hasMessages) {
+      var envelope = dequeue
+      while (envelope ne null) {
+        Logging.StandardOutLogger.tell(envelope.message, envelope.sender)
```
drewhk: how can we be sure that this actor still lives? Also, do we want to cap the messages we try to emit?

drewhk: it might be also useful to add some marker that these are drained after shutdown

ktoso: A limit on how many we drain would be nice to have here. (`logger-max-drained-during-shutdown`?).
Would it not be a problem with the current while loop when there's someone still calling `system.log.warn(...)` and some other thread issued shutdown? Could the shutdown not happen because this actor would not let go because he'd get more and more log statements?

ktoso:
LGTM, but I'd feel a bit safer with a limit on the while loop, unless I missed something and `system.log.info()` calls after shutdown was called will not reach the logger, then it's not an issue I think.

akka-actor/src/main/scala/akka/event/LoggerMailbox.scala
patriknw: it is not a real actor, but I will look into your questions/suggestions on wednesday

patriknw: StandardOutLogger is a MinimalActorRef so tell should work, but I can change it to invoke `print` instead, if that is easier to understand

patriknw: We should have a max count, for several reasons. Another config property :(

drewhk: just comment that it is a MinimalActorRef, that makes it clear enough

ktoso: Comment works for me :-) Did not notice reviewing that it's not a real one

akka-actor-tests/src/test/scala/akka/event/LoggerSpec.scala
rkuhn: like a BAOS!

akka-actor-tests/src/test/scala/akka/event/LoggerSpec.scala
```diff
@@ -148,6 +169,25 @@ class LoggerSpec extends WordSpec with Matchers {
...
+        sys.log.error("msg3")
+        TestKit.shutdownActorSystem(sys, verifySystemShutdown = true)
+        out.flush()
+        out.close()
+      }
+
+      val logMessages = new String(out.toByteArray).split("\n")
+      logMessages.head should include("msg1")
+      logMessages.last should include("msg3")
+      logMessages.size should be === (3)
```
rkuhn: this is the non-type-safe version, right?

akka-actor/src/main/resources/reference.conf
```diff
@@ -452,6 +454,10 @@ akka {
         # com.typesafe.config.Config) parameters.
         mailbox-type = "akka.dispatch.BoundedControlAwareMailbox"
       }
+      
+      logger-queue {
+        mailbox-type = "akka.event.LoggerMailboxType"
```
rkuhn: perhaps add a comment like “don’t touch” or something scary ;-)

akka-actor/src/main/scala/akka/event/LoggerMailbox.scala
rkuhn: `cleanUp` is called after switching out the mailbox, which is why this kind of loop works (it is in the normal `cleanUp` as well), see https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/dispatch/AbstractDispatcher.scala#L201. Therefore I don’t think we need a limit, we just always drain what is in there.

rkuhn:
LGTM apart from three missing comments :-)

drewhk:
LGTM, please fix the rest and merge

akka-actor-tests/src/test/scala/akka/event/LoggerSpec.scala
ktoso: true, `should === 3` would be the type safe one.
Minor issue though.

ktoso:
LGTM from my side, adding a comment around the logger-queue would be a nice to have.

patriknw:
fixed the comments, will merge after validation

drewhk:
nice to see this in!

Issue #16799, opened 2015-02-10
include akka-data-replication


rkuhn:
think about how and where to pull in Patrik’s library

ktoso:
We had some thoughts on this in this ticket https://github.com/akka/akka/issues/13711#issuecomment-65658295 either as a) separate module b) in the cluster tools grab bag would work

patriknw:
this should be a separate module, especially since it will be experimental and the others will not

bantonsson:
@patriknw :+1: my thoughts exactly

Issue #17252, opened 2015-04-22
java.io.Serializer should not supercede a customer serializer.


rcoh:
I recently ran into a pretty surprising behavior in akka serialization -- if your parent class does not inherit from java.io.Serializable, akka will choose arbitrarily between the two! This "feature" is [documented](http://www.google.com/url?q=http%3A%2F%2Fdoc.akka.io%2Fdocs%2Fakka%2Fsnapshot%2Fscala%2Fserialization.html&sa=D&sntz=1&usg=AFQjCNFgyjhi3OzqGto9cCwHMknWkhxFtA) but it's pretty odd, and definitely not what I would expect as a user. `Product` in scala implements java.io.Serializable leading to the conflict. The current behavior is a warning message about the conflict.

A much more reasonable behavior seems like it would be always picking a user defined serializer if appropriate and falling back on Serializable only if no user defined serializers match. It's very counterintuitive that to _disable_ the java serializers, you must inherit from them. The random choice behavior seems like the right thing to do if two user defined serializers both apply.

patriknw:
thanks

patriknw:
We need to discuss the exact solution more, but the problem is interesting.

rcavalcanti:
We have a similar issue in Lagom (https://github.com/lagom/lagom/issues/786). 

Lagom offers the possibility to define serialisers build from play-json serializers (you guys probably not how it works). 

It turns out that if we have a `Format` for a base trait, akka will give precedence to java serializer because a trait doesn't not implement `java.io. Serializable`.

```
sealed trait UserEvent
case class UserUpdatedEvt(...) extends UserEvent
object UserEvent {
  implicit val format: Format[UserEvent] = new Format[UserEvent] {
    override def reads(json: JsValue): JsResult[UserEvent] = // omitted for conciseness
    override def writes(o: UserEvent): JsValue = // omitted for conciseness
  }
}

object UserSerializerRegistry extends JsonSerializerRegistry {
  override def serializers: Seq[JsonSerializer[_]] = Seq(
    JsonSerializer[UserEvent]
  )
}

```
When attempting to persist an Event, we get:
```
[...]

PR #17499, opened 2015-05-17
+dis #16799 Add Distributed Data module

patriknw:
- previously known as [akka-data-replication](https://github.com/patriknw/akka-data-replication)

The documentation is not completely finished, but that should not stop you from starting reviewing this.


akka-docs/rst/scala/distributed-data.rst
```diff
@@ -0,0 +1,387 @@
...
+
+There are some limitations that you should be aware of.
+
+CRDTs cannot be used for all types of problems, and eventual consistency does not fit
+all domains. Sometimes you need strong consistency.
+
+It is not intended for *Big Data*. The number of top level entries should not exceed 100000.
+When a new node is added to the cluster all these entries are transferred (gossiped) to the
+new node. The entries are split up in chunks and all existing nodes collaborate in the gossip,
+but it will take a while (tens of seconds) to transfer all entries and this means that you
```
hepin1989: could we make a pull after the first start up to speed up this progress?

patriknw: I'm sure there are several ways that we can improve the sync of new nodes, but it is pretty good as is, and we can do that as a separate improvement later, when needed.

patriknw:
I have added code examples to the docs

patriknw:
failure was genjavadoc, probably wrong link or something

akka-distributed-data/src/main/resources/reference.conf
```diff
@@ -0,0 +1,45 @@
...
+##############################################
+
+# This is the reference config file that contains all the default settings.
+# Make your edits/overrides in your application.conf.
+
+
+#//#distributed-data
+# Settings for the DistributedData extension
+akka.cluster.distributed-data {
+  # Actor name of the Replicator actor, /user/replicator
```
rkuhn: shouldn’t it move into the `/system/` namespace now?

akka-distributed-data/src/main/scala/akka/cluster/ddata/GCounter.scala
```diff
@@ -0,0 +1,116 @@
...
+final class GCounter private[akka] (
+  private[akka] val state: Map[UniqueAddress, Long] = Map.empty)
+  extends ReplicatedData with ReplicatedDataSerialization with RemovedNodePruning {
+
+  type T = GCounter
+
+  /**
+   * Current total value of the counter.
+   */
+  def value: Long = state.values.sum
```
rkuhn: Would it make sense to parameterize `ReplicatedData` with the type of data that you can “get” like this? And have a generic method
~~~scala
class XYZ extends ReplicatedData[T] ... {
  ...
  def value: T
}
~~~

akka-distributed-data/src/main/scala/akka/cluster/ddata/LWWRegister.scala
```diff
@@ -0,0 +1,171 @@
...
+ */
+package akka.cluster.ddata
+
+import akka.cluster.Cluster
+import akka.cluster.UniqueAddress
+import akka.util.HashCode
+
+object LWWRegister {
+
+  abstract class Clock[A] {
```
rkuhn: why not a (functional) interface?

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORMap.scala
```diff
@@ -0,0 +1,224 @@
...
+  def updated(node: Cluster, key: String, initial: A)(modify: A ⇒ A): ORMap[A] =
+    updated(node.selfUniqueAddress, key, initial)(modify)
+
+  /**
+   * Java API: Replace a value by applying the `modify` function on the existing value.
+   *
+   * If there is no current value for the `key` the `initial` value will be
+   * passed to the `modify` function.
+   */
+  def updated(node: Cluster, key: String, initial: A, modify: akka.japi.Function[A, A]): ORMap[A] =
```
rkuhn: should these move to using `akka.japi.function.Function`?

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORMap.scala
```diff
@@ -0,0 +1,224 @@
...
+   */
+  private[akka] def remove(node: UniqueAddress, key: String): ORMap[A] = {
+    new ORMap(keys.remove(node, key), values - key)
+  }
+
+  override def merge(that: ORMap[A]): ORMap[A] = {
+    val mergedKeys = keys.merge(that.keys)
+    var mergedValues = Map.empty[String, A]
+    mergedKeys.elementsMap.keysIterator.foreach {
+      case key: String ⇒
```
rkuhn: why partial function syntax?

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORMap.scala
```diff
@@ -0,0 +1,224 @@
...
+        (this.values.get(key), that.values.get(key)) match {
+          case (Some(thisValue), Some(thatValue)) ⇒
+            if (thisValue.getClass != thatValue.getClass) {
+              val errMsg = s"Wrong type for merging [$key] in [${getClass.getName}], existing type " +
+                s"[${thisValue.getClass.getName}], got [${thatValue.getClass.getName}]"
+              throw new IllegalArgumentException(errMsg)
+            }
+            val mergedValue = thisValue.merge(thatValue.asInstanceOf[thisValue.T]).asInstanceOf[A]
```
rkuhn: why is the type equality not statically known? it would seem that having the same `A` implies that IAE will not ever be thrown and these casts should be unnecessary …

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORMap.scala
```diff
@@ -0,0 +1,224 @@
...
+          case (None, None) ⇒ throw new IllegalStateException(s"missing value for $key")
+        }
+    }
+
+    new ORMap(mergedKeys, mergedValues)
+  }
+
+  override def needPruningFrom(removedNode: UniqueAddress): Boolean = {
+    keys.needPruningFrom(removedNode) || values.exists {
+      case (_, data: RemovedNodePruning) ⇒ data.needPruningFrom(removedNode)
```
rkuhn: so this is why this happens only once in a blue moon ;-)

akka-distributed-data/src/main/scala/akka/cluster/ddata/GCounter.scala
```diff
@@ -0,0 +1,116 @@
...
+ * which only increment and merge are possible. Incrementing the counter
+ * adds 1 to the count for the current node. Divergent histories are
+ * resolved by taking the maximum count for each node (like a vector
+ * clock merge). The value of the counter is the sum of all node counts.
+ *
+ * This class is immutable, i.e. "modifying" methods return a new instance.
+ */
+@SerialVersionUID(1L)
+final class GCounter private[akka] (
+  private[akka] val state: Map[UniqueAddress, Long] = Map.empty)
```
rkuhn: One thought that occurred to me when reading all the other nice data structures: the focus here lies on replicating some algebraic data types, and using `Long` sticks out like a sore thumb. Should we maybe switch to BigInt to make it nice and “round”?

akka-docs/rst/general/configuration.rst
```diff
@@ -484,3 +484,35 @@ akka-testkit
...
+   
+.. _config-cluster-sharding:
+
+akka-cluster-sharding
+~~~~~~~~~~~~---------
+
+.. literalinclude:: ../../../akka-cluster-sharding/src/main/resources/reference.conf
+   :language: none
+   
+.. _config-distributed-data:
```
rkuhn: :-)

akka-distributed-data/src/main/scala/akka/cluster/ddata/GCounter.scala
patriknw: Java API?

patriknw: how would we serialize it efficiently?

akka-docs/rst/scala/distributed-data.rst
```diff
@@ -0,0 +1,499 @@
...
+====================
+
+The ``akka.cluster.ddata.Replicator`` actor provides the API for interacting with the data.
+The ``Replicator`` actor must be started on each node in the cluster, or group of nodes tagged 
+with a specific role. It communicates with other ``Replicator`` instances with the same path 
+(without address) that are running on other nodes . For convenience it can be used with the
+``akka.cluster.ddata.DistributedData`` extension.
+
+Below is an example of an actor that schedules tick messages to itself and for each tick 
+adds or remove elements from a ``ORSet`` (observed-remove set). It also subscribes to
```
rkuhn: adds or removes

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
```diff
@@ -0,0 +1,1577 @@
...
+      key: String, initial: A, readConsistency: ReadConsistency, writeConsistency: WriteConsistency,
+      request: Option[Any], modify: akka.japi.Function[A, A]) =
+      this(key, readConsistency, writeConsistency, request)(Update.modifyWithInitial(initial, data ⇒ modify.apply(data)))
+
+  }
+
+  sealed trait UpdateResponse extends NoSerializationVerificationNeeded {
+    def key: String
+  }
+  final case class UpdateSuccess(key: String, request: Option[Any]) extends UpdateResponse
```
rkuhn: This should contain the updated value as well.

rkuhn:
Reviewed the data types and the docs, but not the Replicator or the tests. Looks very good so far!

patriknw:
thanks for review, so far

akka-distributed-data/src/main/scala/akka/cluster/ddata/GCounter.scala
rkuhn: BigInteger is a plain Java class, BigInt just a Scala wrapper. Serialization would of course cost proportional to the actual size, which most of the time will be small. BigInteger provides serialization to `byte[]` itself.

patriknw: ok, that sounds possible
What is the advantage over using a Long? Is BigInteger used out there?

rkuhn: The advantage is that it is a much nicer approximation of the concept of “integer number”, I think it’s more fitting for the nice but not terribly efficient data structures that we are distributing—I just don’t like the overflow exception :-)

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
```diff
@@ -0,0 +1,1577 @@
...
+  sealed trait ReadConsistency {
+    def timeout: FiniteDuration
+  }
+  case object ReadLocal extends ReadConsistency {
+    override def timeout: FiniteDuration = Duration.Zero
+  }
+  final case class ReadFrom(n: Int, timeout: FiniteDuration) extends ReadConsistency {
+    require(n >= 2, "ReadFrom n must be >= 2, use ReadLocal for n=1")
+  }
+  final case class ReadQuorum(timeout: FiniteDuration) extends ReadConsistency
```
rkuhn: Thinking about the discussion we had the other day, this should probably be called `ReadMajority`, since quorum has a more statically defined connotation.

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
```diff
@@ -0,0 +1,1577 @@
...
+  final case class WriteTo(n: Int, timeout: FiniteDuration) extends WriteConsistency {
+    require(n >= 2, "WriteTo n must be >= 2, use WriteLocal for n=1")
+  }
+  final case class WriteQuorum(timeout: FiniteDuration) extends WriteConsistency
+  final case class WriteAll(timeout: FiniteDuration) extends WriteConsistency
+
+  /**
+   * Java API: The [[ReadLocal]] instance
+   */
+  def readLocalInstance = ReadLocal
```
rkuhn: perhaps drop the `Instance` suffix for more consistency with the other names?

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
```diff
@@ -0,0 +1,1577 @@
...
+  /**
+   * Java API: The [[WriteLocal]] instance
+   */
+  def writeLocalInstance = WriteLocal
+
+  case object GetKeys
+  /**
+   * Java API: The [[GetKeys]] instance
+   */
+  def GetKeysInstance = GetKeys
```
rkuhn: … or consistently uppercase the first letter :-)

rkuhn:
One weak point of the design is denominating the shared data structure with a String name, making all interactions with the data somewhat dynamically typed. Should we consider using an approach like for the typed Receptionist with its parametric ServiceKey? The type of the key fixes the type of the data structure and in most cases a key will be an `object`, but AFAICS it would also work with case classes:
~~~scala
trait ReplicationKey[T <: ReplicatedData] {
  def name: String // manifest; should default to classname?
  def args: Array[String] // light-weight serialization interface
}

case class UserMap(realm: String) extends ReplicationKey[ORMap[User]] {
  def name = "UserMap"
  def args = Array(realm)
}
~~~
Then we can still decide whether to use name&args for everything internally or whether to look up the class and use its `equals` method.

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
```diff
@@ -0,0 +1,1577 @@
...
+  /**
+   * Java API: The [[FlushChanges]] instance
+   */
+  def flushChangesInstance = FlushChanges
+
+  /**
+   * Marker trait for remote messages serialized by
+   * [[akka.cluster.ddata.protobuf.ReplicatorMessageSerializer]].
+   */
+  trait ReplicatorMessage extends Serializable
```
rkuhn: stupid question: should all these have a `@SerialVersionUID` nevertheless, just in case, for some reason?

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
patriknw: I agree, even though others (Riak, Cassandra) call it quorum.

patriknw:
The typed keys sounds interesting. Can you clarify how the serialization would work? What is the args Array?

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
patriknw: well, the `Serializable` is the workaround for making the serializer pick unique, issue #17252

Adding the SerialVersionUID everywhere kind of implies that we would maintain java serializability compatibility for all data types. That is not something I sign up for.

akka-distributed-data/src/main/scala/akka/cluster/ddata/GCounter.scala
patriknw: I have to do some benchmarks before I can accept this.

akka-distributed-data/src/main/scala/akka/cluster/ddata/GCounter.scala
patriknw: I think I tried that in some early version, but I never found a use case for that abstraction.

There is also a problem; sometimes we would need different types for Java and Scala APIs.

akka-distributed-data/src/main/resources/reference.conf
patriknw: yes

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORMap.scala
patriknw: no reason, thanks

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
patriknw: Sounds good, but I have to try it out.

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
patriknw: yeah, thanks

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
rkuhn: Of course. My thinking went more along the lines of whether alternative serializers like avro or kryo somehow use this UID as well, and if someone wanted to provide their own serialization of these types then that could help them. In the end it does not matter whether we add this annotation or not, people who want to misunderstand the presence of `Serializable` will do so either way. But I don’t have a strong opinion on this particular point.

akka-distributed-data/src/main/scala/akka/cluster/ddata/GCounter.scala
rkuhn: Okay, good points.

akka-distributed-data/src/main/scala/akka/cluster/ddata/GCounter.scala
rkuhn: Sounds good to me!

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
```diff
@@ -0,0 +1,1577 @@
...
+          var mergedRemovedNodePruning = other.pruning
+          for ((key, thisValue) ← pruning) {
+            mergedRemovedNodePruning.get(key) match {
+              case None ⇒ mergedRemovedNodePruning = mergedRemovedNodePruning.updated(key, thisValue)
+              case Some(thatValue) ⇒
+                mergedRemovedNodePruning = mergedRemovedNodePruning.updated(key, thisValue merge thatValue)
+            }
+          }
+
+          copy(data = cleaned(data, mergedRemovedNodePruning), pruning = mergedRemovedNodePruning).merge(other.data)
```
rkuhn: perhaps add a comment that confirms that the intent is to clean up both sides before merging, it took me a while to completely understand this—or inline the merging here

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
```diff
@@ -0,0 +1,1577 @@
...
+      private def cleaned(c: ReplicatedData, p: Map[UniqueAddress, PruningState]): ReplicatedData = p.foldLeft(c) {
+        case (c: RemovedNodePruning, (removed, PruningState(_, PruningPerformed))) ⇒
+          if (c.needPruningFrom(removed)) c.pruningCleanup(removed) else c
+        case (c, _) ⇒ c
+      }
+
+      def addSeen(node: Address): DataEnvelope = {
+        var changed = false
+        val newRemovedNodePruning = pruning.map {
+          case (removed, pruningNode) ⇒
```
rkuhn: `s/pruningNode/pruningState/`?

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
```diff
@@ -0,0 +1,1577 @@
...
+    clockTask.cancel()
+  }
+
+  def matchingRole(m: Member): Boolean = role.forall(m.hasRole)
+
+  def receive = normalReceive
+
+  val normalReceive: Receive = {
+    case Get(key, consistency, req)                 ⇒ receiveGet(key, consistency, req, sender())
+    case Update(key, _, _, req) if !isLocalSender() ⇒ receiveInvalidUpdate(key, req)
```
rkuhn: the message made it over the wire, why reject it?

akka-docs/rst/scala/distributed-data.rst
```diff
@@ -0,0 +1,499 @@
...
+If consistency is important, you can ensure that a read always reflects the most recent
+write by using the following formula::
+
+    (nodes_written + nodes_read) > N 
+
+where N is the total number of nodes in the cluster, or the number of nodes with the role that is
+used for the ``Replicator``.
+
+For example, in a 7 node cluster you get this strong read consistency by writing to 4 nodes and
+reading from 4 nodes, or writing to 5 nodes and reading from 3 nodes.
```
rkuhn: We should make it clear that this does not mean “strong” as in “strong consistency”, especially since both words appear close to each other. Concurrent updates will still be processed concurrently and then merged, meaning that global invariants cannot be preserved regardless of consistency settings. Guaranteeing that reads see all previously committed writes does not imply serializability or linearizability.

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
```diff
@@ -0,0 +1,1577 @@
...
+ * As reply of the `Get` a [[Replicator.GetSuccess]] is sent to the sender of the
+ * `Get` if the value was successfully retrieved according to the supplied consistency
+ * level within the supplied timeout. Otherwise a [[Replicator.GetFailure]] is sent.
+ * If the key does not exist the reply will be [[Replicator.NotFound]].
+ *
+ * You will always read your own writes. For example if you send a `Update` message
+ * followed by a `Get` of the same `key` the `Get` will retrieve the change that was
+ * performed by the preceding `Update` message. However, the order of the reply messages are
+ * not defined, i.e. in the previous example you may receive the `GetSuccess` before
+ * the `UpdateSuccess`.
```
rkuhn: This is the main point I’d like to discuss: providing this message-ordering-like guarantee is a slightly unusual formulation of “read your own writes” and it has a rather high implementation cost. Normally the quoted statement would mean that if you send an update and receive your successful response then a subsequent query will see that update (plus any concurrently completed ones). The current implementation does not fundamentally improve upon this AFAICS, other writes could still intervene, so why should we not simplify the Replicator and remove the buffers? (We should probably collapse incoming requests pertaining to the same key and consistency level into a single ReadAggregator action, but that is an optional optimization.)

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
```diff
@@ -0,0 +1,1577 @@
...
+      changed = changed.updated(key, oldDigest)
+    }
+
+    val dig = if (envelope.data == DeletedData) DeletedDigest else LazyDigest
+    dataEntries = dataEntries.updated(key, (envelope, dig))
+  }
+
+  def getDigest(key: String): Digest = {
+    dataEntries.get(key) match {
+      case Some((envelope @ DataEnvelope(_, _), LazyDigest)) ⇒
```
rkuhn: this match can be simplified to
~~~scala
case Some((envelope, LazyDigest)) =>
~~~

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
```diff
@@ -0,0 +1,1577 @@
...
+        case Some(envelope) ⇒
+          val msg = if (envelope.data == DeletedData) DataDeleted(key) else Changed(key, envelope.data)
+          subs.foreach { _ ! msg }
+        case None ⇒
+      }
+    }
+
+    if (subscribers.nonEmpty) {
+      for ((key, oldDigest) ← changed; subs ← subscribers.get(key)) {
+        if (oldDigest != getDigest(key))
```
rkuhn: when would this be false?

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
patriknw: because many data types make use of version vectors, which must use the local node

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
patriknw: It has only used when there is an Update with readConsistency != ReadLocal, but I think that is important for making some updates sane.

If you for example emit this sequence of updates:
1. `Update(WriteLocal, ReadQuorum)(_.increment(1))`
2. `Update(WriteLocal, ReadQuorum)(_.increment(1))`
3. `Update(WriteLocal, ReadLocal)(current => if (current.value >= 2) current.increment(2) else current.increment(1))`

I would expect the result to be 4, i.e. subsequent updates see previous updates (read your own writes).

Read your own writes is a very desired UX requirement in my opinion.

rkuhn: Of course it is, I was not arguing against reading your own writes, my argument was that this usually entails waiting for confirmation of your writes before doing the reads, because otherwise the write could just fail anyway. So, what I’m saying is that in normal databases writes that are done concurrently by the same client are, well, concurrent. The same client then means using different connections from the pool, of course. The actor interface of just doing those three message sends in a row would then correspond to using three connections to talk to the same database, whereas the current implementation corresponds to funneling all accesses to one key through the same connection—for all clients.

patriknw:
@rkuhn I tried the typed keys. I was able to change all things in the Replicator and its messages, but then I realized that the serialized key (full class) must be embedded in the `Get` message and all its response messages, because those are allowed remotely. I'm not sure it is worth it.

Note that there is already a type for the `Update` via the `initial` parameter:

    replicator ! Update("users", initial = ORMap.empty[User], WriteLocal)(_ + (user.id -> user))

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
patriknw: Some more examples:

1. `Update(WriteLocal)(_ + "a")`
2. `Get`

Here you will get the "a" as expected, but with the following you would not (unless it is your lucky day):

1. `Update(WriteLocal, ReadQuorum)(_ + "a")`
2. `Get`

One more...

1. `Update(WriteLocal, ReadQuorum)(_ + "a")`
2. `Update(WriteLocal, ReadQuorum)(_ - "a")`

The 2nd could then overtake the 1st.

Yes, user could wait for the reply (by stashing incoming requests) themselves, but it will not be as efficient as current solution (note that it is only the buffering during the read part of the update).

Since we have these semantics naturally for updates with `readConsistency == ReadLocal` and that is what will be used most I think it would cause a lot of confusing to not have the same semantics for updates with `readConsistency != ReadLocal`.

@huntc Do you have an opinion about this, as a user?

akka-docs/rst/scala/distributed-data.rst
patriknw: Good point, I was trying to describe "ensure that a read always reflects the most recent write" (strong read consistency). I agree that we should reword to avoid confusion.

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORMap.scala
patriknw: or java8 `java.util.function.Function`?

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORMap.scala
patriknw: without the last `asInstanceOf[A]` I get `type mismatch; found : mergedValue.type (with underlying type thisValue.T) required: A` on the next line.

Without the `thatValue.asInstanceOf[thisValue.T]` I get `type mismatch; found : thatValue.type (with underlying type A) required: thisValue.T` on this line

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
huntc: All that I can say at this point is that I'm yet to stumble on a use-case for `ReadLocal`. That doesn't mean it is undesirable to have; it could also mean that I should be using it in places.

rkuhn: What are you using instead?

huntc: `ReadQuorum` and `WriteQuorum`.

rkuhn: @huntc When placing a key in a map or incrementing a counter or removing an element from a set, all that `ReadQuorum` does is to make it slower AFAICS. The only difference it could make is if the operation is contingent on the previously observed state, but as argued elsewhere this is fishy: whatever consistency settings you choose there is no way to maintain global invariants with this scheme, so the condition being fulfilled is an illusion for many use-cases.

@patriknw The problems I see are twofold: the current implementation is a lot more complex than it could be without this feature, hence the feature needs good justification (I have not yet been convinced but that may change of course), and secondly “read your writes” usually only applies to a single client while here all usage of a certain key is penalized with huge extra latency while one particular client asks for external confirmation.

In summary, we should move this forward by determining whether there is actual value in ReadQuorum and friends, because “read your writes” can certainly be implemented without it in a much simpler fashion.

huntc: We use ReadQuorum sparingly and only when we need to make an important decision such as scaling up or down. The objective is to make the decision given a reasonable view on the state. There are no guarantees that the state is correct, but there's a reasonable probability. 

When we update state, we use WriteQuorum. I'm not quite sure why as we generally don't care and don't handle failure. May we should handle failure though... May be we did that to help any ReadQuorum along subsequently... I don't recall.

patriknw: @rkuhn After some more thinking (long walks are good for that purpose) I'm fine with removing the readConsistency feature from the `Update` (it was the difference in semantics that was my concern). It is enough if we describe that `Get(ReadQuorum)` continued by `Update(WriteQuorum)` when the `GetSuccess` arrives for those (rare) use cases. That is rather easy thanks to request parameter. It doesn't have the same ordering properties in relation to other commands as current solution, but that would not be surprising since a separate `Get` is emitted.

Another use case for it: remove element in ORSet requires that you have seen the element, and therefore you might want to try to fetch latest first.

@huntc When you use update with ReadQuorum/WriteQuorum you should care about the `ReadFailure`, because if you get that your update will not be applied.

The `UpdateTimeout`, on the other hand, does not mean that the update completely failed or was rolled back. It means that the direct replication of the `Update` could not be fulfill according to
the given consistency level and timeout. It is always done locally. It may still have been replicated to some nodes, and will eventually be replicated to all nodes with the gossip protocol.

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
patriknw: I'm not sure, but I don't think it's needed. I might have been thinking that for changes back and forth resulting in same content should not emit an event, but that is not a valid crdt as far as I can see. I'll remove it.

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
rkuhn: Thanks for the consideration and I agree with your reasoning, in essence this makes the primitives orthogonal and thereby simplifies the implementation without compromising on API consistency. Good call. If someone wants the convenience of the old behavior they can simply create a proxy actor that does the Get/Update combo and the buffering—and if this turns out to be a common concern then we’ll see a contribution for it.

huntc: > When you use update with ReadQuorum/WriteQuorum you should care about the ReadFailure, because if you get that your update will not be applied.

I don't use `Update` with `ReadQuorum`/`WriteQuorum`. I use the flavour that defaults to `ReadLocal` along with an explicit `WriteQuorum`.

However I'll go through the code today and ensure that `UpdateResponse` messages are handled given a log event if nothing else. I have noticed a few unhandled `UpdateSuccess` messages for example.

akka-distributed-data/src/main/scala/akka/cluster/ddata/LWWRegister.scala
patriknw: changed

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
patriknw: I don't see the use case where that should be needed, especially not when we remove the read part of `Update`. The current value will always be what the user changed it to in the `modify` function. It can be different on other nodes, but that is not something we should care about.

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
patriknw: but on the other hand, the function will not accidentally make it through serialization for such cases. I removed this, in case someone cares to write a fancy serializer.

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
patriknw: @huntc great, then this change will not affect you

patriknw:
@rkuhn see latest 6 commits for changes of what we have discussed, and some more. Next step (not today) is to remove the read part of the `Update`.

patriknw:
@rkuhn still some javadoc errors in here. The first error according to the perl script is:

    [error] /Users/patrik/dev/akka/akka-distributed-data/target/java/akka/cluster/ddata/LWWRegister$.java:10: error: reference not found
    [error]    * The default {@link Clock} is using max value of <code>System.currentTimeMillis()</code>

Why is such a link in LWWRegister invalid? Is there a workaround?

rkuhn:
@patriknw The link is valid in `LWWRegister` but not in `LWWRegister$`, you’ll need to use `LWWRegister.Clock`.

patriknw:
@rkuhn I have removed the readConsistency in the `Update` as we agreed on, see cb62621. Note the adjustments of the ShoppingCart sample, and the description in the docs.

patriknw:
@rkuhn Fixed some javadoc issues, but a few more to go:

    [error] /Users/patrik/dev/akka/akka-distributed-data/target/java/akka/cluster/ddata/Replicator.java:110: error: reference not found
    [error]  * You can retrieve all keys of a local replica by sending {@link Replicator.GetKeys} message to the
    [error]                                                                   ^
    [error] /Users/patrik/dev/akka/akka-distributed-data/target/java/akka/cluster/ddata/Replicator.java:119: error: reference not found
    [error]  * send an explicit {@link Replicator.FlushChanges} message to the <code>Replicator</code> to notify the subscribers
    [error]                            ^
    [error] /Users/patrik/dev/akka/akka-distributed-data/target/java/akka/cluster/ddata/Replicator.java:603: error: reference not found
    [error]    * @see Replicator.GetReplicaCount
    [error]           ^
    [error] /Users/patrik/dev/akka/akka-distributed-data/target/java/akka/cluster/ddata/Replicator.java:365: error: reference not found
    [error]    * and it is also possible to send an explicit {@link FlushChanges} message to
    [error]                                                         ^
    [error] /Users/patrik/dev/akka/akka-distributed-data/target/java/akka/cluster/ddata/VersionVector.java:61: error: reference not found
    [error]    * Java API: The {@link VersionVector.After} instance
    [error]                           ^
    [error] /Users/patrik/dev/akka/akka-distributed-data/target/java/akka/cluster/ddata/VersionVector.java:66: error: reference not found
    [error]    * Java API: The {@link VersionVector.Before} instance
    [error]                           ^
    [error] /Users/patrik/dev/akka/akka-distributed-data/target/java/akka/cluster/ddata/VersionVector.java:71: error: reference not found
    [error]    * Java API: The {@link VersionVector.Same} instance
    [error]                           ^
    [error] /Users/patrik/dev/akka/akka-distributed-data/target/java/akka/cluster/ddata/VersionVector.java:76: error: reference not found
    [error]    * Java API: The {@link VersionVector.Concurrent} instance
    [error]                           ^
    [error] /Users/patrik/dev/akka/akka-distributed-data/target/java/akka/cluster/ddata/VersionVector$.java:19: error: reference not found
    [error]    * Java API: The {@link VersionVector.After} instance
    [error]                           ^
    [error] /Users/patrik/dev/akka/akka-distributed-data/target/java/akka/cluster/ddata/VersionVector$.java:24: error: reference not found
    [error]    * Java API: The {@link VersionVector.Before} instance
    [error]                           ^
    [error] /Users/patrik/dev/akka/akka-distributed-data/target/java/akka/cluster/ddata/VersionVector$.java:29: error: reference not found
    [error]    * Java API: The {@link VersionVector.Same} instance
    [error]                           ^
    [error] /Users/patrik/dev/akka/akka-distributed-data/target/java/akka/cluster/ddata/VersionVector$.java:34: error: reference not found
    [error]    * Java API: The {@link VersionVector.Concurrent} instance
    [error]                           ^
    [info] Building index for all the packages and classes...
    [info] Building index for all classes...
    [info] 12 errors

patriknw:
@rkuhn I have reworked everything with the typed keys as we discussed: a34c306

I also removed the stash from the ShoppingCart sample as we agreed on: 20c9110

rkuhn:
An alternative approach to the `String => Key` mapping within the Replicator would be to only work with the typed keys in-JVM and do the conversion to/from Strings in the serializer; do you think this would make it simpler?

rkuhn:
LVGTM!

patriknw:
The types are not known until user sends a command (Update, Get, Subscribe) and I don't want the overhead of replicating the type information (serialized keys). I don't think the current approach of using Strings internally is complicated. It only needs to keep one map with the types for the subscriptions, to be able to emit the right key instances in the `Changed` events.

rkuhn:
Okay, makes sense.

patriknw:
added the java documentation

patriknw:
@ktoso @2m @drewhk are you alright with squashing and merging this tomorrow, so that it can be released in 2.4-M2? Roland has reviewed it (thoroughly) and said LVGTM.

There is one remaining task before final release. Converting the samples to Java and moving the samples to activator templates, but that can be done after this PR.

akka-distributed-data/src/main/resources/reference.conf
```diff
@@ -0,0 +1,61 @@
...
+
+# This is the reference config file that contains all the default settings.
+# Make your edits/overrides in your application.conf.
+
+
+#//#distributed-data
+# Settings for the DistributedData extension
+akka.cluster.distributed-data {
+  # Actor name of the Replicator actor, /system/replicator
+  name = replicator
```
ktoso: minor: kind of generic name, would data-replicator be better?

patriknw: good point, I think we tend to write actor names in camel: dataReplicator (it's weird that I'm thinking about the extra bytes that it adds to the remote messages)

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORSet.scala
```diff
@@ -0,0 +1,299 @@
...
+ * In this implementation we compare the `dot` of the present element to the version vector
+ * in the Set it is absent from. If the element dot is not "seen" by the Set version vector,
+ * that means the other set has yet to see this add, and the item is in the merged
+ * Set. If the Set version vector dominates the dot, that means the other Set has removed this
+ * element already, and the item is not in the merged Set.
+ *
+ * This class is immutable, i.e. "modifying" methods return a new instance.
+ */
+@SerialVersionUID(1L)
+final class ORSet[A] private[akka] (
```
ktoso: excellent, and I finally read up a bit on the dots :)

akka-distributed-data/src/main/scala/akka/cluster/ddata/ReplicatedData.scala
```diff
@@ -0,0 +1,82 @@
...
+ * Java API: Interface for implementing a [[ReplicatedData]] in
+ * Java.
+ */
+abstract class AbstractReplicatedData extends ReplicatedData {
+  // it is not possible to use a more strict type, because it is erased somehow, and
+  // the implementation is anyway required to implement
+  // merge(that: ReplicatedData): ReplicatedData
+  type T = AbstractReplicatedData
+
+}
```
ktoso: There is a way to make this typesafe in Java, via "self recursive type parameters" (sometimes called F-bounds):

Here's some docs I did on the pattern: http://ktoso.github.io/scala-types-of-types/#self-recursive-type

```scala
class Data[T] extends Data[Data[T]] {
  def merge(d: T): T = ???
}
```

usage in java:
```java
  static class X extends Data<X> { }
  static class Y extends Data<Y> { }
  {
    X x = new X();
    Y y = new Y();

    x.merge(x);
    y.merge(y);
    x.merge(y); // good, compilation fails: merge(X) cannot be applied to Y
    y.merge(x); // good, compilation fails: merge(Y) cannot be applied to X
  }
```

So this could be improved in next versions for Java I think :-)

ktoso:
Really awesome work Patrik!
Read through most of the code again now but not very deeply - LGTM for pulling it in to M2 and let's move on from there :-)

If you have a todo-list somewhere about what could use some more work would be great to make it tickets on akka/akka now.

akka-distributed-data/src/main/scala/akka/cluster/ddata/ReplicatedData.scala
patriknw: interesting, then it would be something like:

    abstract class AbstractReplicatedData[D <: AbstractReplicatedData[D]] extends ReplicatedData

but I could not immediately get that to work (but I think that is because of T in ReplicatedData, perhaps we could use this pattern for both java and scala.

Please create a ticket and we can explore it.

ktoso: I believe it works in both Scala and Java, so yeah :)
Ticket created https://github.com/akka/akka/issues/17770 (and distributed-data now has it's own label, yay!)

akka-distributed-data/src/main/resources/reference.conf
ktoso: > (it's weird that I'm thinking about the extra bytes that it adds to the remote messages)

not weird at all, had the same thing... ;-) Well, another reason to get the compaction of those at some point.

patriknw:
@ktoso thanks for the review, I will create tickets for the remaining things (samples conversion)

akka-distributed-data/src/main/scala/akka/cluster/ddata/ReplicatedData.scala
patriknw: thanks

patriknw:
@huntc I have extracted your contributions to separate commits here. The ORMultiMap commit will be an excellent pointer for anyone else that wants to know what is needed to contribute a new data type.

patriknw:
Credits to @jboner included in the [commit message](https://github.com/akka/akka/commit/cbe5dd2cf5d30556126b89f0c37b837edd11f8d1)

patriknw:
I will merge this after pr validation. Yay!

ktoso:
:tada: :tada: congrats on bringing in a new module @patriknw! :tada: :tada:

rkuhn:
Yeehah! Great work, Patrik!

huntc:
Really awesome! Well done Patrik!

smlin:
Great! Are you planning use it in akka-persistence? I think it can be helpful for multiple distributed PersistentActors with one persistentId. Like in the Eventuate.

hepin1989:
can't wait for the akka2.4-m2.

Issue #17495, opened 2015-05-16
FileAndResourceDirectives.getFromResourceDirectory with sub directories may list files 


matlockx:
When passing a directory which contains subdirectories to FileAndResourceDirectives.getFromResource, request which map to a subdirectory will create a response with the file listing of that subdirectory. 
In detail: When the path of the request ends with a '/' the request will be rejected, if not the request comes through and the file listing will be returned.

jrudolph:
Can you post relevant parts of your route? Do you really mean `getFromResource` or `getFromResourceDirectory`?

jrudolph:
I can remember that we discussed this issue in the context of spray a long time ago. The problem is that it's possible to use `ClassLoader.getResource(dirName).openStream` to list contents of classpath resource directories (if the classpath entry is a file system directory). It doesn't seem to work when the classpath entry is a jar.

I guess the fix is then to check if the URL returned from `ClassLoader.getResource` is a `file` URL, and if yes, check that it doesn't point to a directory.

/cc @sirthias

matlockx:
Actually you are right, it's getFromResourceDirectory. Sorry for that.

The path is something like:

```scala
    pathPrefix("game") {
	    	get {
                  getFromResourceDirectory("public")
                }
   }
```

jrudolph:
Just to confirm: are you experiencing this when everything is packaged into a jar or does this happen during development (e.g. with sbt) when the referenced resources are in a classpath directory?

matlockx:
I get it during development with sbt. 
I just checked it in production and there it's a NPE.... 

ava.lang.NullPointerException: null
        at java.io.FilterInputStream.close(FilterInputStream.java:181) ~[na:1.8.0_40]
        at sun.net.www.protocol.jar.JarURLConnection$JarURLInputStream.close(JarURLConnection.java:108) ~[na:1.8.0_40]
        at spray.routing.directives.FileAndResourceDirectives$class.getFromResource(FileAndResourceDirectives.scala:128) ~[io.spray.spray-routing_2.11-1.3.2.jar:na]
        at iwg.api.routes.StaticGameContentRoute.getFromResource(StaticGameContentRoutes.scala:15) ~[io.lotteries.iwg-ticketpool_2.11-0.4.17.jar:0.4.17]
        at spray.routing.directives.FileAndResourceDirectives$class.getFromResource(FileAndResourceDirectives.scala:104) ~[io.spray.spray-routing_2.11-1.3.2.jar:na]

[...]

PR #17503, opened 2015-05-18
=htp #17495 don't serve the content of resource directories

aruediger:


jrudolph:
@2beaucoup Interesting. Do you understand why it would work? It seems strange that content length and content wouldn't agree about the amount of data. Wouldn't we rely on a strange behavior of `FileUrlConnection`?

Can you add a test that tries to get a resource directory from a dependency jar?

And thanks for helping out :)

/cc @sirthias

2beaucoup:
Dunno @jrudolph. I just tried to detect directories within resources and checking against `conn.getContentLength > 0` looked like the only relyable hint. I wasn't able to find a reference in the docs though.

I'll try to come up with a test against a real jar to confirm that the fix is valid. If I recall correctly you were doing something similar (using a dependency jar) in the past. Could you give me a pointer? Thx!

2beaucoup:
FTR: This succeeds on my machine:
```scala
val jarFile = s"${sys.props("user.home")}/.ivy2/cache/com.google.guava/guava/jars/guava-12.0.jar"
"dir" in {
  val jarCon = new URL(s"jar:file:$jarFile!/META-INF").openConnection
  jarCon.getContentLength shouldEqual 0
}
"dir nonexistent" in {
  val jarCon = new URL(s"jar:file:$jarFile!/META-INFX").openConnection
  jarCon.getContentLength shouldEqual -1
}
"file" in {
  val jarCon = new URL(s"jar:file:$jarFile!/META-INF/MANIFEST.MF").openConnection
  jarCon.getContentLength shouldBe > (0)
}
```

jrudolph:
Ok, this would then almost look like a solution. It seems we then cannot serve empty files from resources any more but that may be the lesser evil.

To test I guess we could take the typesafe config jar as granted while running the tests, then we take a package directory from that jar, make sure to use the right classloader, and see what happens...

jrudolph:
The alternative solution would be to inspect the scheme of the url returned by "theClassLoader.getResource(resourceName)" and fish out `file` and `jar` schemes and handle them differently.

(I checked the OpenJDK source code of `JarURLConnection.getContentLength` and `FileURLConnection.getContentLength`. FileURLConnection naively reports the value of `File.length` even if ["The return value is unspecified if this pathname denotes a directory."]( https://docs.oracle.com/javase/7/docs/api/java/io/File.html#length()). JarURLConnection reports the value of JarEntry.getSize which doesn't even talk about directories.)

2beaucoup:
They also behave differently [here](https://github.com/akka/akka/blob/release-2.3-dev/akka-http-scala/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala#L102), where a `NullPointerException` is thrown for jar dirs.

What would you handle differently re. jar vs. file? We could also match on `JarURLConnection` / `FileURLConnection` instead of looking at the schemes. WDYT?

2beaucoup:
I added more tests. Maybe we should borrow from [play's impl}(https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/utils/Resources.scala#L24-L46) for reliable directory detection.

drewhk:
needs a rebase it seems

2beaucoup:
rebased, added more tests and removed the usage of `URLConnection`s

/cc @jrudolph

akka-http-scala/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -208,6 +202,35 @@ object FileAndResourceDirectives extends FileAndResourceDirectives {
...
+        val startIndex = if (jarFile.startsWith("file:")) 5 else 0
+        val bangIndex = jarFile.indexOf("!")
+        val jarFilePath = jarFile.substring(startIndex, bangIndex)
+        val resourcePath = jarFile.substring(bangIndex + 2)
+        val jar = new java.util.zip.ZipFile(jarFilePath)
+        try {
+          val entry = jar.getEntry(resourcePath)
+          if (entry.isDirectory) None
+          else {
+            // isDirectory returns false for URLs pointing to directories without trailing slash
```
jrudolph: Yes, no surprise because it's implemented as `name.endsWith("/");`. I'm not sure if there have to be entries included for directories in the zip file format at all or if they are synthetic.

2beaucoup: LOL. I'll ditch the `isDirectory` check then and just check if `getInputStream(entry)` doesn't return `null`. URLs with trailing slashes doesn't get this far anyway.

drewhk:
LGTM?

jrudolph:
Thanks @2beaucoup, I like it and with the new tests I somewhat confident that this is better than the previous solution. LGTM

PR #17506, opened 2015-05-18
!htc #16826 introduce HttpCookiePair for usage in `Cookie`-header

### LINKED ISSUE #16826: Improve Cookie header model
From spray/spray#753:

> The Cookie header should only return cookie pairs without also specifying all of the possible settings that may be specified in the HttpCookie class.

Also see the discussion in spray/spray#739.

/cc @sirthias @2beaucoup

jrudolph:
This cleans up a small inconsistency in the cookie model  (discovered by @2beaucoup in spray) where the Cookie header model contained unused data leading to potential confusion. 

I'm a bit unsure of the naming of `HttpCookiePair` and `HttpCookie`, which could also be named `HttpCookie` and `HttpCookieWithAttributes` or similar. If you have better suggestions please get in touch ;)

Fixes #16826.

/cc @sirthias

### DIFF
--- a/akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/directives/CookieDirectivesExamplesSpec.scala
+++ b/akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/directives/CookieDirectivesExamplesSpec.scala
@@ -13,10 +13,10 @@ class CookieDirectivesExamplesSpec extends RoutingSpec {
   "cookie" in {
     val route =
       cookie("userName") { nameCookie =>
-        complete(s"The logged in user is '${nameCookie.content}'")
+        complete(s"The logged in user is '${nameCookie.value}'")
       }
 
-    Get("/") ~> Cookie(HttpCookie("userName", "paul")) ~> route ~> check {
+    Get("/") ~> Cookie("userName" -> "paul") ~> route ~> check {
       responseAs[String] shouldEqual "The logged in user is 'paul'"
     }
     // missing cookie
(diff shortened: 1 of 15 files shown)


jrudolph:
Fixed broken examples in docs.

jrudolph:
Fails with lots of genjavadoc errors, is PR validation currently broken?

ktoso:
@jrudolph we're batelling with the move to JDK8 for PR validation these days. Today it should be solved, see: https://github.com/akka/akka/pull/17511 https://github.com/akka/akka/pull/17510

drewhk:
PLS BUILD

drewhk:
Hmm, HttpCookiePair is a bit non-descriptive for me. HttpCookieKeyValue?

jrudolph:
Actually, "cookie-pair" is the name from the spec. I'm actually more concerned about the name `HttpCookie` because it contains not just key and value but also all the attributes which define to which domains/path etc. it applies. It's hard to decide because most implementation don't even make the distinction and mix the things that can be used in the `Set-Cookie` and the `Cookie` headers.

drewhk:
Well, if the functionality is well-documented then I guess the current names are fine. There are no good established naming it seems, and I don't see any obvious better choices.

drewhk:
However, given that there is Cookie, HttpCookie and HttpCookiePair, the distinction between Cookie and HttpCookie is not clear. Maybe HttpCookieEntry instead of HttpCookie?

drewhk:
LGTM, we only need to settle on the names.

jrudolph:
@2beaucoup any ideas about the naming?

2beaucoup:
@jrudolph I like it like it is:

`Cookie` header contains `HttpCookiePair`s and `Set-Cookie` headers contain a `HttpCookie`

Maybe we can come up with a naming scheme that illustrates that `HttpCookie` is used for **setting** new cookies (`Cookie = `server -> client) and `HttpCookiePair`s are used for **resubmitting** already set cookies (`Set-Cookie` = client -> server).

sirthias:
My 2 cents on the naming: I think the names are good enough as is.
As @2beaucoup said:

> `Cookie` header contains `HttpCookiePair`s and `Set-Cookie` headers contain a `HttpCookie`.

Makes sense to me.

drewhk:
Agreed. Let's rebase-validate-merge.

jrudolph:
Ok, I'll rebase.

jrudolph:
Rebased.

jrudolph:
Failed in TlsSpec

jrudolph:
This can be merged AFAICS.

Issue #17415, opened 2015-05-07
CircuitBreaker throws IllegalStateException


rkuhn:
from #16327:
~~~
2015-04-27 14:06:06.498502500 [incident-daemon-system-akka.actor.default-dispatcher-35] ERROR akka.dispatch.Dispatcher - exception in sameThreadExecution
Context (a.d.Dispatcher)
2015-04-27 14:06:06.498504500 java.lang.IllegalStateException: Illegal transition attempted from: Closed with failure count = 10 to Open
2015-04-27 14:06:06.498505500   at akka.pattern.CircuitBreaker.transition(CircuitBreaker.scala:229)
2015-04-27 14:06:06.498506500   at akka.pattern.CircuitBreaker.akka$pattern$CircuitBreaker$$tripBreaker(CircuitBreaker.scala:236) 2015-04-27 14:06:06.498524500     at akka.pattern.CircuitBreaker$Closed$.callFails(CircuitBreaker.scala:369) 
2015-04-27 14:06:06.498525500   at akka.pattern.CircuitBreaker$State$$anonfun$callThrough$1.apply(CircuitBreaker.scala:299) 
2015-04-27 14:06:06.498526500   at akka.pattern.CircuitBreaker$State$$anonfun$callThrough$1.apply(CircuitBreaker.scala:297) 
2015-04-27 14:06:06.498529500   ... 25 common frames omitted
2015-04-27 14:06:06.498529500 Wrapped by: java.lang.IllegalStateException: exception in sameThreadExecutionContext
2015-04-27 14:06:06.498530500   at akka.dispatch.ExecutionContexts$sameThreadExecutionContext$.reportFailure(Future.scala:77) 
2015-04-27 14:06:06.498535500   at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:32) 
2015-04-27 14:06:06.498535500   at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:55) 
2015-04-27 14:06:06.498536500   at akka.dispatch.BatchingExecutor$Batch.run(BatchingExecutor.scala:73) 
2015-04-27 14:06:06.498539500   at akka.dispatch.ExecutionContexts$sameThreadExecutionContext$.unbatchedExecute(Future.scala:74) 
2015-04-27 14:06:06.498539500   at akka.dispatch.BatchingExecutor$class.execute(BatchingExecutor.scala:120) 
[...]

PR #17513, opened 2015-05-19
=act #17415 fix concurrency bug in CircuitBreaker (for validation)

patriknw:
- transition did not account for concurrent transistions

(cherry picked from commit 2a73e77f66202b1bd5a5918329b9dd34448ce7b6)

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/pattern/CircuitBreakerStressSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/pattern/CircuitBreakerStressSpec.scala
@@ -1,0 +1,83 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.pattern
+
+import scala.concurrent.Promise
+import scala.concurrent.duration._
+import scala.concurrent.forkjoin.ThreadLocalRandom
+import akka.actor.Actor
+import akka.actor.ActorLogging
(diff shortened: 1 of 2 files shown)


patriknw:
Refs #17415

patriknw:
That looks like a relevant failure that I should investigate before we merge this.

    [info] - A CircuitBreaker *** FAILED *** (3 seconds, 143 milliseconds)
    [info]   982 was not equal to 0 (CircuitBreakerStressSpec.scala:74)
    [info]   at org.scalatest.Matchers$ShouldMethodHelper$.shouldMatcher(Matchers.scala:6231)
    [info]   at org.scalatest.Matchers$AnyShouldWrapper.should(Matchers.scala:6265)
    [info]   at akka.pattern.CircuitBreakerStressSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply(CircuitBreakerStressSpec.scala:74)
    [info]   at akka.pattern.CircuitBreakerStressSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply(CircuitBreakerStressSpec.scala:71)

Note that this has already been merged to release-2.3 branch, so any changes should be ported there also.

patriknw:
The problem was in the new test. The reason it failed in master but not in release-2.3 is that we have improved the call timeout handling in master. Anyway, the test was not doing the right thing.

Issue #17494, opened 2015-05-15
Each incoming connection leaks one ActorInterpreter


catalin-ursachi:
As of RC1, an ActorInterpreter leak appears to have been introduced to Akka HTTP - precisely one actor is leaked per each incoming connection.

The following test sets up a simple Akka server; counts the initial number of ActorInterpreters; sends 100 requests and waits for the responses; allows for a 1-second delay; and counts the number of ActorInterpreters again. Under M4, the initial and final counts were both equal to 2; the test passes. Under RC1, the final count is precisely 100 greater than the initial count. The test doesn't pass.

Allowing for a 20-second delay doesn't help.

```scala
package akka.http

import akka.actor._
import akka.http.scaladsl.Http
import akka.http.scaladsl.client.RequestBuilding
import akka.http.scaladsl.model._
import akka.stream.ActorFlowMaterializer
import akka.stream.impl.fusing.ActorInterpreter
import akka.stream.scaladsl.{Flow, Sink, Source}
import org.specs2.matcher.MatchResult
import org.specs2.mutable.Specification
import org.specs2.time.NoTimeConversions
import scala.collection.mutable
import scala.concurrent.duration._
import scala.concurrent.{Await, Future}

class IncomingConnectionActorInterpreterLeakSpec
  extends Specification
  with RequestBuilding
  with NoTimeConversions {

  val defaultTimeout = 20.seconds

  /**
   * A simple server which replies to all requests with 200 OK and mirrors the request's entity (to confirm that it's
   * been streamed properly both ways). We are interested in the server's [[ActorSystem]]
   */
  class TestServer() {

    implicit val actorSystem = ActorSystem("ServerActorSystem")
    implicit val materializer = ActorFlowMaterializer(namePrefix = Some("ServerMaterializer"))

    val testServerFuture = Http().bind("127.0.0.1", 0).to {
      Sink.foreach { incomingConnection =>
        incomingConnection.handleWith(Flow[HttpRequest].map { request =>
          HttpResponse(entity = request.entity)
        })
      }
    }.run()

    val server = Await.result(testServerFuture, defaultTimeout)

    val host: String = server.localAddress.getHostName
    val port: Int = server.localAddress.getPort
  }

  /**
   * A simple test client which sends requets to a particular host and port.
   * We are not interested in its [[ActorSystem]]; there do not appear to be any actor leaks from outgoing connections.
   */
  class TestClient(host: String, port: Int) {

[...]

Issue #17453, opened 2015-05-12
When prefixAndTail() is called on a stream with a lazy empty tail, and prefixAndTail() is called again on the tail, the stream hangs.


catalin-ursachi:
(This problem has appeared as of RC1; we didn't encounter it in previous versions of Akka Streams)

Say we've got a source, as given by:

`val source = Source.single(Elem) concatMat Source.lazyEmpty[Elem.type] (...)`

We apply .prefixAndTail(1) on the source, and complete the promise materialized from the "lazy empty source". We then have a prefix and a tail; the tail, if ran, would turn out to contain no elements.

We then call .prefixAndTail(1) on the tail. We run the resulting source with a Sink.head, getting a Future[Seq[...], Source[...]]. We wait for the future to complete.

If we repeat the above 10,000 times, the future completes in some 9,970. It hangs in the other 30.

Funnily enough, if before calling .prefixAndTail(1) the second time, we precede it by .map(identity) (which, in principle, should of course do nothing), then the stream *does* complete 100% of the time.

The following set of tests illustrates the issue. All three tests should pass; but only the first and third do. The second fails with e.g. `'9980' is not equal to '10000'` (i.e. 0.2% of attempts failed)

```
package test

import akka.actor.ActorSystem
import akka.stream.ActorFlowMaterializer
import akka.stream.scaladsl.{Sink, Source}
import org.specs2.mutable.Specification
import org.specs2.time.NoTimeConversions
import scala.collection.immutable.Seq
import scala.concurrent.Await
import scala.concurrent.duration._
import scala.util.Try

class PrefixAndTailHangingErrorSpec
  extends Specification
  with NoTimeConversions {

  implicit val system = ActorSystem()
  implicit val materializer = ActorFlowMaterializer()
  
  case object Elem

  def foldSourceToList[T](source: Source[T, _]): List[T] =
    Await.result(source.runFold(List.empty[T])(_ :+ _), 5.seconds)

  // A source containing a single element, but which doesn't complete until the materialized promise is completed.
  def sourceWithLazyTail() =
    (Source.single(Elem) concatMat Source.lazyEmpty[Elem.type])((_, promise) => promise)

  // We apply prefixAndTail to the source, and complete the materialized promise.
  def applyOnce() = {
    val (promise, future) = sourceWithLazyTail()
      .prefixAndTail(1)
      .toMat(Sink.head)((_, _))
      .run()

    promise.success(())

    future
  }

[...]

PR #17514, opened 2015-05-19
=str #17453: Fix substream RS compliance, awaitAllStagesStopped and FanIn leak

drewhk:
- Substreams emitted were not RS compliant
- awaitAllStagesStopped was racy and therefore did not check the right thing and sometimes caused innocent tests to fail
- FanIn did not stop the actor on early failures and did not wait properly for all Subscriptions

### DIFF
--- a/akka-stream-testkit/src/test/scala/akka/stream/testkit/Utils.scala
+++ b/akka-stream-testkit/src/test/scala/akka/stream/testkit/Utils.scala
@@ -7,6 +7,7 @@ import akka.stream.impl._
 import akka.testkit.TestProbe
 import com.typesafe.config.ConfigFactory
 import scala.concurrent.duration._
+import scala.util.Try
 import scala.util.control.NoStackTrace
 
 object Utils {
@@ -19,9 +20,10 @@ object Utils {
   def assertAllStagesStopped[T](block: ⇒ T)(implicit materializer: FlowMaterializer): T =
     materializer match {
       case impl: ActorFlowMaterializerImpl ⇒
-        impl.supervisor ! StreamSupervisor.StopChildren
-        val result = block
         val probe = TestProbe()(impl.system)
+        probe.send(impl.supervisor, StreamSupervisor.StopChildren)
+        probe.expectMsg(StreamSupervisor.StoppedChildren)
+        val result = block
         probe.within(5.seconds) {
           probe.awaitAssert {
             impl.supervisor.tell(StreamSupervisor.GetChildren, probe.ref)
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowPrefixAndTailSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowPrefixAndTailSpec.scala
@@ -4,11 +4,11 @@
 package akka.stream.scaladsl
 
 import scala.collection.immutable
-import scala.concurrent.Await
+import scala.concurrent.{ Future, Await }
 import scala.concurrent.duration._
+import scala.util.Try
 import scala.util.control.NoStackTrace
-import akka.stream.ActorFlowMaterializer
-import akka.stream.ActorFlowMaterializerSettings
+import akka.stream.{ OperationAttributes, ActorFlowMaterializer, ActorFlowMaterializerSettings }
 import org.reactivestreams.Subscriber
 import akka.stream.testkit._
 import akka.stream.testkit.Utils._
@@ -26,7 +26,7 @@ class FlowPrefixAndTailSpec extends AkkaSpec {
 
     def newHeadSink = Sink.head[(immutable.Seq[Int], Source[Int, _])]
 
-    "work on empty input" in {
+    "work on empty input" in assertAllStagesStopped {
       val futureSink = newHeadSink
       val fut = Source.empty.prefixAndTail(10).runWith(futureSink)
       val (prefix, tailFlow) = Await.result(fut, 3.seconds)
@@ -36,7 +36,7 @@ class FlowPrefixAndTailSpec extends AkkaSpec {
       tailSubscriber.expectSubscriptionAndComplete()
     }
 
-    "work on short input" in {
+    "work on short input" in assertAllStagesStopped {
       val futureSink = newHeadSink
       val fut = Source(List(1, 2, 3)).prefixAndTail(10).runWith(futureSink)
       val (prefix, tailFlow) = Await.result(fut, 3.seconds)
@@ -68,7 +68,7 @@ class FlowPrefixAndTailSpec extends AkkaSpec {
       Await.result(fut2, 3.seconds) should be(1 to 10)
     }
 
-    "handle negative take count" in {
+    "handle negative take count" in assertAllStagesStopped {
       val futureSink = newHeadSink
       val fut = Source(1 to 10).prefixAndTail(-1).runWith(futureSink)
       val (takes, tail) = Await.result(fut, 3.seconds)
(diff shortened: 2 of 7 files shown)


akka-stream-testkit/src/test/scala/akka/stream/testkit/Utils.scala
```diff
@@ -19,9 +20,10 @@ object Utils {
   def assertAllStagesStopped[T](block: ⇒ T)(implicit materializer: FlowMaterializer): T =
     materializer match {
       case impl: ActorFlowMaterializerImpl ⇒
-        impl.supervisor ! StreamSupervisor.StopChildren
-        val result = block
         val probe = TestProbe()(impl.system)
+        probe.send(impl.supervisor, StreamSupervisor.StopChildren)
+        probe.expectMsg(StreamSupervisor.StoppedChildren)
```
drewhk: previously StopChildren was in race with the execution of "block"

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowPrefixAndTailSpec.scala
```diff
@@ -192,6 +192,23 @@ class FlowPrefixAndTailSpec extends AkkaSpec {
...
+      val sub = TestSubscriber.manualProbe[Int]()
+
+      val f = Source(pub).prefixAndTail(1).runWith(Sink.head)
+      val s = pub.expectSubscription()
+      s.sendNext(0)
+
+      val (_, tail) = Await.result(f, 3.seconds)
+
+      val tailPub = tail.runWith(Sink.publisher)
+      s.sendComplete()
```
drewhk: At this point, since prefixAndTail was not seen the input stream completed, it already emitted a substream for the tail instead of Source.empty. We finish the input stream causing the substream to complete.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowPrefixAndTailSpec.scala
```diff
@@ -192,6 +192,23 @@ class FlowPrefixAndTailSpec extends AkkaSpec {
...
+
+      tailPub.subscribe(sub)
+      sub.expectSubscriptionAndComplete()
```
drewhk: originally substream immediately sent onComplete which is no longer RS compliant

akka-stream/src/main/scala/akka/stream/impl/ActorProcessor.scala
```diff
@@ -201,7 +201,7 @@ private[akka] class SimpleOutputs(val actor: ActorRef, val pump: Pump) extends D
     }
   }
 
-  override def isClosed: Boolean = downstreamCompleted
+  override def isClosed: Boolean = downstreamCompleted && (subscriber ne null)
```
drewhk: this should not matter, but better safe than sorry. This will not allow the outputs to be considered closed while there is no subscriber.

akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
```diff
@@ -241,7 +241,7 @@ private[akka] abstract class FanIn(val settings: ActorFlowMaterializerSettings,
   protected def fail(e: Throwable): Unit = {
     if (settings.debugLogging)
       log.debug("fail due to: {}", e.getMessage)
-    inputBunch.cancel()
+    nextPhase(completedPhase)
```
drewhk: this ensures that if we are in the WaitingForUpstreamSubscription state then we will wait for all subscriptions before we finish. (nextPhase does not immediately take effect if we are in WaitingForUpstreamSubscription)

akka-stream/src/main/scala/akka/stream/impl/StreamOfStreamProcessors.scala
```diff
@@ -82,9 +82,11 @@ private[akka] object MultiStreamOutputProcessor {
     private def closePublisher(withState: CompletedState): Unit = {
       subscriptionTimeout.cancel()
       state.getAndSet(withState) match {
-        case Attached(sub)     ⇒ closeSubscriber(sub, withState)
         case _: CompletedState ⇒ throw new IllegalStateException("Attempted to double shutdown publisher")
-        case Open              ⇒ // No action needed
+        case Attached(sub) ⇒
+          if (subscriber eq null) tryOnSubscribe(sub, CancelledSubscription)
```
drewhk: This will execute when the actor has an enqueued subscriber (Attached) which it didn't yet handle (subscriber eq null), so it needs to onSubscribe first. 

What seems to be missing though if after closePublisher the actor does not stop, it *will* handle the subscribe message again, so I should put there a guard (checking for completion).

akka-stream-testkit/src/test/scala/akka/stream/testkit/Utils.scala
ktoso: Good catch

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
```diff
@@ -256,6 +256,8 @@ private[akka] object StreamSupervisor {
   final case class Children(children: Set[ActorRef])
   /** Testing purpose */
   final case object StopChildren
+  /** Testing purpose */
+  final case object StoppedChildren
```
ktoso: nitpick: `ChildrenStopped`?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowPrefixAndTailSpec.scala
ktoso: In theory we could apply the TCK to the substream, how about it?

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
drewhk: well, it has only sent stop, we don't know if the children are actually stopped

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowPrefixAndTailSpec.scala
drewhk: there is a TCK test for prefixAndTail, but it does not exercise this code path.

ktoso: Ok, I think I see how it wouldn't hit this path... There is a test for completely empty stream, but that's slightly different.

akka-stream/src/main/scala/akka/stream/impl/ActorFlowMaterializerImpl.scala
ktoso: Ok, got it :)

rkuhn:
LGTM—but this is subtle stuff so _appearances can be deceiving …_

drewhk:
At least all the previous tests pass, and one more that didn't previously ;)

2m:
LGTM. Just checked and this does not fix https://github.com/akka/akka/issues/17494

drewhk:
Kind of expected. Have you found out yet which stages are actually leaking?

akka-stream/src/main/scala/akka/stream/impl/StreamOfStreamProcessors.scala
ktoso: The fix here makes sense. 

I'm not quite sure in what scenario the actor would not stop in `closeSubscriber()`, AFAICS since state here is `Attached` we won't go into the `Open`state on the closeSubscriber, so the only path that does nothing to the actor is `case Failed(e: SpecViolation) ⇒ // nothing to do`.. Unless you mean that after receiving complete/error the actor is wrongly implemented and does not stop - but then it's his fault and not this code?

ktoso:
Investigated and LGTM :-)

Would like to understand deeper the care you are worried about in https://github.com/akka/akka/pull/17514#discussion_r30593201 @drewhk but that's an extra.

akka-stream/src/main/scala/akka/stream/impl/StreamOfStreamProcessors.scala
drewhk: This is exactly something I wanted to completely convince myself of:)

ktoso: Ok I see :-) I'm not sure about the `SpecViolation` case as explained above - if the downstream would be stopped properly hm.

drewhk: I will investigate this separately, let's merge this to reduce the leaks that we know exist.

drewhk:
Fixes #17514

akka-stream/src/main/scala/akka/stream/impl/StreamOfStreamProcessors.scala
ktoso: +1 agreed

PR #17525, opened 2015-05-19
=pro Disable doclint

patriknw:
- release script fails, probably due to
  `[error] (akka-actor/genjavadoc:doc) javadoc returned nonzero exit code`
- I don't know why, but this is is a workaround to be able to release


rkuhn:
The issues is that we javadoc the test classes and that chokes on Mockito (why are we using that anyway?). So the real fix is to not emit javadoc for the test classes.

patriknw:
Ok, I have no idea why tests are javadoc'd and how to fix it.

rkuhn:
proper fix is in https://github.com/akka/akka/pull/17528

patriknw:
great, then we close this

PR #17526, opened 2015-05-20
=con #13909 Don't use remote watch in ClusterClient

patriknw:
- because it will result in quarantine if failure
  detection triggers and that kind of coupling is
  exactly what is not desired for a ClusterClient
- replace by simple heartbeat failure detection

### DIFF
--- a/akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
+++ b/akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
@@ -4,10 +4,8 @@
 package akka.cluster.client
 
 import java.net.URLEncoder
-
 import scala.collection.immutable
 import scala.concurrent.duration._
-
 import akka.actor.Actor
 import akka.actor.ActorIdentity
 import akka.actor.ActorLogging
@@ -37,6 +35,8 @@ import akka.japi.Util.immutableSeq
 import akka.routing.ConsistentHash
 import akka.routing.MurmurHash
 import com.typesafe.config.Config
+import akka.actor.DeadLetterSuppression
+import akka.remote.DeadlineFailureDetector
 
 object ClusterClientSettings {
   /**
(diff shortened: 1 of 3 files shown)


patriknw:
Refs #13909

patriknw:
note to self: add some DeadLetterSuppression here

patriknw:
any reviewers from @akka/akka-team here (8 days old)?

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -174,12 +189,17 @@ class ClusterClient(settings: ClusterClientSettings) extends Actor with Stash wi
 
   require(initialContacts.nonEmpty, "initialContacts must be defined")
 
+  val acceptableHeartbeatPauseNanos: Long = acceptableHeartbeatPause.toNanos
```
drewhk: why not reuse the DeadlineFailureDetector?

patriknw: I was thinking that it was so simple that I don't needed any fancy stuff, but I guess you are right, I'll give that a try

drewhk:
LGTM

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -174,12 +191,17 @@ class ClusterClient(settings: ClusterClientSettings) extends Actor with Stash wi
 
   require(initialContacts.nonEmpty, "initialContacts must be defined")
 
+  // FIXME use new constructor of DeadlineFailureDetector with heartbeatInterval
+  val failureDetector = new DeadlineFailureDetector(acceptableHeartbeatPause)
```
patriknw: note to self: remember to change this before merging, i.e. when #17604 has been merged

patriknw:
@drewhk I have updated to use the `DeadlineFailureDetector`, see 1907b16
. That is better. Thanks.

drewhk:
Thanks, nice!

patriknw:
rebased, and fixed the FIXME, will merge after validation

PR #17528, opened 2015-05-20
do not generate javadoc for tests

rkuhn:


project/AkkaBuild.scala
```diff
@@ -313,7 +313,8 @@ object AkkaBuild extends Build {
     // compile options
     scalacOptions in Compile ++= Seq("-encoding", "UTF-8", "-target:jvm-1.8", "-feature", "-unchecked", "-Xlog-reflective-calls", "-Xlint"),
     scalacOptions in Compile ++= (if (allWarnings) Seq("-deprecation") else Nil),
-    scalacOptions in Test := (scalacOptions in Test).value.filterNot(_ == "-Xlog-reflective-calls"),
+    scalacOptions in Test := (scalacOptions in Test).value.filterNot(opt =>
+      opt == "-Xlog-reflective-calls" || opt.contains("genjavadoc")),
```
patriknw: alright, magic

patriknw:
LGTM, thanks

project/AkkaBuild.scala
rkuhn: No, not at all: we previously added the configuration that emits the synthetic Java sources to the Test configuration, now we don’t.

patriknw: where do we add this opt to scalacOptions in the first place?
I can't find the string "genjavadoc" anywhere else than `scalacOptions in Compile += "-P:genjavadoc:fabricateParams=true"`, but that can't be the thing.
Is it added by unidoc?

rkuhn: Yes, it is in genjavadocExtraSettings, pulled in from `project/Unidoc.scala`.

Issue #17524, opened 2015-05-19
Use assertAllStagesStopped in Http tests


drewhk:
@patriknw created this helper method (in stream.testkit.Utils) which helped to uncover many of the leaks we had in streams-core. We whould use this in Http tests, too (where applicable).

drewhk:
Be sure to base it on most recent head (after the fix in https://github.com/akka/akka/pull/17514)

sirthias:
Is this closed by #17529?

drewhk:
I think it covers only a subset. You should validate where it makes sense to use it in other places. If you find that all relevant tests are covered then this ticket can be closed.

ktoso:
Let's add the checks soon, in 1.1.

ktoso:
That's a nice one to pick up if you had the time @johanandren :) // tomorrow of course

johanandren:
Adding closeNetworkInput/expectNetworkClose/assertAllStagesStopped to HttpServerSpec
    test makes it fail with ```assertion failed: expected OnComplete, found OnSubscribe(...)`. Not sure if it is because I have no idea what I am doing or there actually is something weird going on.

Tried using ```netOut.probe.fishForPF``` for ```OnSubscribe => false``` + ```OnComplete => true```, but then it just fails because of never getting the ```OnComplete```

johanandren:
Would be happy for some feedback from @jrudolph or @sirthias about this.

ktoso:
The PR under discussion is https://github.com/akka/akka/pull/18749/files

jrudolph:
@johanandren  funny, that you find that just now. I think this is directly related to a change I just proposed: https://github.com/akka/akka/pull/18748/files#diff-f3a816c2a9e353d4e7c075320c6f53ceR599

Automatic probes are not smart enough to expect subscription first and then completion.

jrudolph:
Related to this issue, there's still #18005 open.

johanandren:
Tried applying https://github.com/spray/akka/commit/fa6f13783b716875823cd11de09c66636e96a6a8 first, then it fails because it never gets any ```OnComplete``` (the same as my try with ```fishForPF```)

johanandren:
#18005 would be later, when ```expectNetworkClose()``` has not failed and the actual ```assertAllStagesStopped``` logic as far as I understand

jrudolph:
@johanandren wrt #18005 I didn't mean your current issue but the original issue tracked here.

[...]

PR #17529, opened 2015-05-20
=htc #17494 fix interpreter leak due to unsubscribed websocket handler hole

### LINKED ISSUE #17494: Each incoming connection leaks one ActorInterpreter

jrudolph:
The solution is to install a dummy handler when the connection is closed while still being in http mode.

Fixes #17494 and also introduces a few `assertAllStagesStopped` as suggested in #17524.

/cc @sirthias

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -350,9 +350,20 @@ private[http] object HttpServerBluePrint {
             case ResponseRenderingOutput.SwitchToWebsocket(responseBytes, handlerFlow) ⇒
               ctx.emit(responseBytes)
               installHandler(handlerFlow)
+              ctx.changeCompletionHandling(defaultCompletionHandling)
               websocket
           }
         }
+
+        // Completion handling that installs a dummy handler, to make sure no processors leak because they have
+        // never been subscribed to, see #17494.
+        override def initialCompletionHandling: CompletionHandling =
+          CompletionHandling(
+            onUpstreamFinish = { (ctx, in) ⇒ installDummyHandler(); defaultCompletionHandling.onUpstreamFinish(ctx, in) },
+            onUpstreamFailure = { (ctx, in, t) ⇒ installDummyHandler(); defaultCompletionHandling.onUpstreamFailure(ctx, in, t) })
+
+        def installDummyHandler(): Unit = installHandler(Flow[FrameEvent])
+
         def websocket: State[_] = State[ByteString](Read(wsIn)) { (ctx, in, bytes) ⇒
           ctx.emit(bytes)
           SameState
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebsocketServerSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebsocketServerSpec.scala
@@ -7,6 +7,7 @@ package akka.http.impl.engine.ws
 import akka.http.impl.engine.ws.Protocol.Opcode
 import akka.http.scaladsl.model.ws._
 import akka.stream.scaladsl.{ Sink, Flow, Source }
+import akka.stream.testkit.Utils
 import akka.util.ByteString
 import org.scalatest.{ Matchers, FreeSpec }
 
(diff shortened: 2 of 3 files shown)


drewhk:
LGTM

patriknw:
LGTM, I guess

drewhk:
Is there a test that fails without the fix?

jrudolph:
Yes, the tests in ClientServerSpec.scala fail since I added `assertAllStagesStopped`.

drewhk:
OK

PR #17531, opened 2015-05-20
=pro update sbt-unidoc to 0.3.3

ktoso:
This update solves our blocker issue about compile not being triggered properly before genjavadoc kicks in.


ktoso:
// cc @patriknw

patriknw:
Got it, thanks

PR #17532, opened 2015-05-20
=pro run multi-jvm tests when validating pr

2m:
The essence of this is one more autoplugin `MultiNodeWithPrValidation` that gets triggered on a project which has both `MultiNode` and `ValidatePullRequest` autoplugins enabled.


project/package.scala
```diff
@@ -0,0 +1,7 @@
+package object akka {
+
+  implicit class RichBoolean(val b: Boolean) extends AnyVal {
+    final def apply[A](a: => A): Option[A] = if (b) Some(a) else None
+  }
```
ktoso: I would prefer to avoid such very broad conversions (on primitive type, providing an apply()), or if really needed define very near it's use-site.

project/ValidatePullRequest.scala
```diff
@@ -209,4 +208,13 @@ object ValidatePullRequest extends AutoPlugin {
...
   )
+
+  def whenValidatingPr(task: TaskKey[_]) =
+    validatePullRequest := {
+      (projectBuildMode in ValidatePR).value match {
+        case BuildSkip => // do not run multi-jvm tests if project is skipped during pr validation
+        case _ => task.value
+      }
+      validatePullRequest.value
+    }
```
ktoso: Nice one! Slightly adjust the comment maybe to explain "when BuildSkip we do not do anything here" and "task.value" will force the tasks execution

project/ValidatePullRequest.scala
```diff
@@ -30,22 +30,21 @@ object ValidatePullRequest extends AutoPlugin {
...
   }
   case object BuildProjectChangedQuick extends BuildMode {
-    override def task = Zero.dependsOn(test in ValidatePR)
+    override def task = Zero.dependsOn(test in ValidatePR).dependsOn(unidoc in Compile)
     def log(projectName: String, l: Logger) =
       l.info(s"Building [$projectName] as the root `project/` directory was affected by this PR.")
   }
   final case class BuildCommentForcedAll(phrase: String, c: GHIssueComment) extends BuildMode {
-    override def task = Zero.dependsOn(test in Test)
+    override def task = Zero.dependsOn(test in Test).dependsOn(unidoc in Compile)
```
ktoso: Very good catch, thanks

project/MultiNode.scala
```diff
@@ -27,49 +44,65 @@ object MultiNode {
...
+/**
+ * This autoplugin adds Multi Jvm tests to validatePullRequest task.
+ * It is needed, because ValidatePullRequest autoplugin does not depend on MultiNode and
+ * therefore test:executeTests is not yet modified to include multi-jvm tests when ValidatePullRequest
+ * build strategy is being determined.
+ *
+ * Making ValidatePullRequest depend on MultiNode is impossible, as then ValidatePullRequest
+ * autoplugin would trigger only on projects which have both of these plugins enabled.
+ */
+object MultiNodeWithPrValidation extends AutoPlugin {
```
ktoso: Maybe move this one to `ValidatePullRequest`?

akka-cluster-tools/build.sbt
```diff
@@ -11,8 +11,8 @@ Unidoc.javadocSettings
 
 OSGi.clusterTools
 
-MultiNode.multiJvmSettings
-
 libraryDependencies ++= Dependencies.clusterTools
 
 //MimaKeys.previousArtifact := akkaPreviousArtifact("akka-cluster-tools").value
+
+enablePlugins(MultiNode)
```
ktoso: yay :)

project/ValidatePullRequest.scala
ktoso: For this to work add `scalaJavaUnidocSettings` to `projectSettings` on line 122, otherwise it's a reference to undefined setting during the run. 
After fixing this I got the "duplicated class" problems from genjavadoc (I assume, looks like it) – we're almost there it seems.

project/Unidoc.scala
```diff
@@ -7,47 +7,44 @@ import sbt.Keys._
...
+  override lazy val projectSettings = scaladocSettings ++ javadocSettings
 
-  def scaladocSettings: Seq[sbt.Setting[_]] = {
-    scaladocSettingsNoVerificationOfDiagrams ++
-      (if (scaladocDiagramsEnabled) Seq(doc in Compile ~= scaladocVerifier) else Seq.empty)
-  }
+  val validateDiagrams = settingKey[Boolean]("Validate generated scaladoc diagrams")
+
+  val javadocSettings = if (CliOption.genjavadocEnabled) Seq(
+    scalacOptions in Compile += "-P:genjavadoc:fabricateParams=true",
```
ktoso: This seems to be the failure culprit:

```
[error] bad option: -P:genjavadoc:fabricateParams=true
```

ktoso: the same setting was used previously... so what's different here hm

project/package.scala
patriknw: +1

project/Unidoc.scala
ktoso: @2m no need for `genjavadocExtraSettings`?

project/package.scala
2m: It reduces boilerplate when used together with boolean flags that we have quite a few. Therefore it is not possible to add it to a narrower scope.

For example previously we had:
```
if (CliOption.sbtLogNoFormat) List("-Dakka.test.nocolor=true") else Nil
```
And now we have instead:
```
CliOption.sbtLogNoFormat("-Dakka.test.nocolor=true").toList
```

WDYT?

project/Unidoc.scala
2m: `genjavadocExtraSettings` was used [before](https://github.com/akka/akka/blob/53c5d974eac9d77df9315120374e73d5c7306dc1/project/Unidoc.scala#L26) as well.

ktoso: Sorry, the comment was supposed to be "need for `genjavadocExtraSettings`" it was there but now it's gone

ktoso: (does not solve the bad option problem we're having though, so maybe a non issue)

ktoso:
LGTM, one nitpick about moving javadoc things out of the scaladoc plugin.

2m:
Replaced boolean extension methods with `CliOption` case class and friends. After validation this is good to be merged after a squash.

project/CliOptions.scala
```diff
@@ -0,0 +1,25 @@
...
+  trait CliOptionParser[T] {
+    def parse(path: String, default: T): CliOption[T]
+  }
+
+  object CliOptionParser {
+    implicit object BooleanCliOptionParser extends CliOptionParser[Boolean] {
+      def parse(path: String, default: Boolean) =
+        CliOption(sys.props.getOrElse(path, default.toString).toBoolean)
+    }
+  }
```
ktoso: :+1:

ktoso:
LGTM, I hope validation thinks so as well :-)

2m:
Me too.

ktoso:
Needs a rebase

project/CliOptions.scala
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka
+
+case class CliOption[T](value: T)
+
+object CliOption {
+  def apply[T](path: String, default: T)(implicit ev: CliOptionParser[T]): CliOption[T] = ev.parse(path, default)
+
+  implicit class BooleanCliOption(cliOption: CliOption[Boolean]) {
```
rkuhn: `extends AnyVal`?

project/RootSettings.scala
```diff
@@ -5,18 +5,10 @@ package akka
...
 object RootSettings extends AutoPlugin {
 
-  lazy val docs = ProjectRef(file("."), "akka-docs")
-
   // settings for root project goes here
   override val projectSettings = Seq(
-    validatePullRequest <<= validatePullRequest.dependsOn(unidoc in Compile),
-    validatePullRequest <<= validatePullRequest.dependsOn(SphinxSupport.generate in Sphinx in docs)
   )
 }
```
rkuhn: Does this file still serve a purpose now?

project/ValidatePullRequest.scala
```diff
@@ -209,4 +208,42 @@ object ValidatePullRequest extends AutoPlugin {
     // add reportBinaryIssues to validatePullRequest on minor version maintenance branch
     validatePullRequest <<= validatePullRequest.dependsOn(reportBinaryIssues)
   )
+
+  def whenValidatingPr(task: TaskKey[_]) =
+    validatePullRequest := {
+      (projectBuildMode in ValidatePR).value match {
+        case BuildSkip => // do not run the additional task if project is skipped during pr validation
+        case _ => task.value
```
rkuhn: are failures of this task propagated even when ignoring its value?

project/RootSettings.scala
2m: Yes it does. There are two plugins that depend on `RootSettings`: `TimeStampede` and `UnidocRoot`.

project/Doc.scala
```diff
@@ -7,47 +10,36 @@ import sbt.Keys._
...
-  }
-
-  // for projects with few (one) classes there might not be any diagrams
-  def scaladocSettingsNoVerificationOfDiagrams: Seq[sbt.Setting[_]] = {
+  override lazy val projectSettings = {
     inTask(doc)(Seq(
       scalacOptions in Compile <++= (version, baseDirectory in ThisBuild) map scaladocOptions,
-      autoAPIMappings := scaladocAutoAPI
-    ))
+      autoAPIMappings := CliOptions.scaladocAutoAPI.value
```
rkuhn: This looks suspiciously like a macro evaluation, please rename the `value` method—what about `get`?

project/Doc.scala
```diff
@@ -82,3 +74,63 @@ object Unidoc {
...
+/**
+ * For projects with few (one) classes there might not be any diagrams.
+ */
+object ScaladocNoVerificationOfDiagrams extends AutoPlugin {
+
+  override def trigger = noTrigger
+  override def requires = Scaladoc
+
+  override lazy val projectSettings = Seq(
+    Scaladoc.validateDiagrams in Compile := false
```
rkuhn: Wouldn’t it be simpler to just add this one line to those projects that need it?

project/Doc.scala
```diff
@@ -82,3 +74,63 @@ object Unidoc {
...
+      unidocProjectFilter in JavaUnidoc := docProjectFilter,
+      apiMappings in ScalaUnidoc := (apiMappings in (Compile, doc)).value
+    ))
+  }
+
+  override lazy val projectSettings =
+    CliOptions.genjavadocEnabled.ifTrue(scalaJavaUnidocSettings).getOrElse(scalaUnidocSettings) ++
+    settings(Seq(Project(base=file("akka-samples"), id="akka-samples")), Seq(
+      Project(base=file("akka-remote-tests"), id="akka-remote-tests"),
+      Project(base=file("akka-jmh"), id="akka-jmh")
```
rkuhn: This looks a bit fragile, why not leave the selection of what to exclude in AkkaBuild where it is statically verified?

project/RootSettings.scala
rkuhn: Yes, but would removing the dependency actually change anything?

2m: Yes. Instead of `enablePlugins(akka.RootSettings)` in the root `build.sbt` we would need to enable every autoplugin separately.

project/ValidatePullRequest.scala
ktoso: Hmm, I think you're right Roland but we'll need to verify to be sure I guess. In the validator itself it's fine as `Def.taskDyn` is used https://github.com/akka/akka/blob/master/project/ValidatePullRequest.scala#L201
In taskDyn when you inspect a task you can't see what it depends on, because the dependencies only appear once you run it, that's what we need here as well.

project/CliOptions.scala
2m: This would not make `BooleanCliOption` a value class, because `cliOption` is not primitive. If this was dotty we could make both of these a value class to avoid allocating objects. Is there some other reason to make `BooleanCliOption` extend `AnyVal`?

project/Doc.scala
2m: In that case we would be repeating this specific option in every project that does not need diagram verification. Now it is nicely abstracted away.

2m:
Made dynamic task execution work with a bit of magic from http://stackoverflow.com/questions/24615783/combining-taskdyn-and-tags-in-sbt thanks to @jsuereth

project/CliOptions.scala
rkuhn: WDYM? Extending `AnyVal` (and making `val cliOption`) would mean that this erases to `CliOption` and is stored as that without a wrapper. This is not performance sensitive code, so it is not needed, it is just a habit/pattern to model extension methods using implicit value classes. The only reason for not doing this would be if this were declared within a trait (because then it does not work).

ktoso:
Finally all green!

project/CliOptions.scala
rkuhn: BTW: an alternative formulation would be using phantom types on a method that lives directly on CliOption :-)

ktoso: that'd make our build the most type-fancy kid on the block ;-)

project/RootSettings.scala
rkuhn: Hmm, I can’t seem to make sense of this, please help: my naïve understanding is that removing this plugin means that the root project (`/build.sbt`) will need to explicitly enable the UnidocRoot and TimeStampede plugins. If that is true then I’d prefer that, it would make things more explicit without losing functionality—but I think I’m still missing something.

rkuhn: and BTW: I think we might need an automated test suite for our build as well ;-) (something that verifies that the right set of tests are actually run and all the required sources are compiled)

rkuhn:
LGTM

project/RootSettings.scala
patriknw: :-)

patriknw:
LGTM

project/RootSettings.scala
2m: After a bit of though I agree with removing `RootSettings` autoplugin and adding other plugins explicitly, mainly because this is how we do it in sub projects. Root project should not be treated differently.

2m:
Added one last commit which removed `RootSettings` autoplugin and also kept values from tasks to get rid of the 
```
[warn] /home/martynas/projects/akka/project/ValidatePullRequest.scala:217: a pure expression does nothing in statement position; you may be omitting necessary parentheses
[warn]       validationTasks.map(taskKey => Def.task { taskKey.value; () } ).foldLeft(zero) { (acc, current) =>
[warn]
```
when compiling akka build.

Validated again that no additional tasks are run when project validation is skipped.

Will squash and merge after successful validation.

project/RootSettings.scala
2m: > and BTW: I think we might need an automated test suite for our build as well ;-)

This can not be verified without running the actual tasks because we started to use dynamic settings and tasks. But we could do a report after a run and validate if the required tasks have been run after the fact.

rkuhn: Sounds good, and now we even know the reason why we picked one over the other, thanks! :-)

PR #17534, opened 2015-05-20
=htp #16808 fix unnecessary reverse DNS lookups for >Java1.7

agolubev:
PR from #17481 but re-targeted to release-2.3-dev branch

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
@@ -8,13 +8,13 @@ import java.net.InetSocketAddress
 import scala.annotation.tailrec
 import akka.event.LoggingAdapter
 import akka.util.ByteString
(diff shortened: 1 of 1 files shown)


ktoso:
Thanks for re-targeting! As reviewed previously, LGTM

ktoso:
OK TO TEST

drewhk:
LGTM

drewhk:
Fixes #16808

jrudolph:
@agolubev This is broken for me in Java 6 OpenJDK. Did you check that it works? Is Java 6 compatibility somehow checked in PR validation?

ktoso:
Good catch... We now are running validation with JDK8 (because `master`), and I think *all* validation is now on JDK8 so probably a bad idea and this could have slipped through hm..

jrudolph:
Yes, also the problem is that to properly test it, you would have to compile the files in the same way the release is compiled and then just run (compile?) the tests with Java 6.

agolubev:
I see, let me check for configuration you mentioned. Compile as in released and then run under 1.6

PR #17536, opened 2015-05-21
=pro Use provided dependency from persistence in bench-jmh

patriknw:
- compilation error when running release script

### DIFF
--- a/project/AkkaBuild.scala
+++ b/project/AkkaBuild.scala
@@ -94,7 +94,7 @@ object AkkaBuild extends Build {
   lazy val benchJmh = Project(
     id = "akka-bench-jmh",
     base = file("akka-bench-jmh"),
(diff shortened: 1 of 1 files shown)


patriknw:
no need for pr-validation

patriknw:
@ktoso Do you know if it is possible to compile, but not run, `bench-jmh` in ordinary builds?

ktoso:
@patriknw could you refine what you mean by ordinary builds? Validation / during dev time?
It should be compiled, and never run on jenkins as-is already, but I guess you mean something different?

patriknw:
It did not fail in any of our jobs, we have this dependeny change for a long time. Release script triggered compilation error.

Note that it is not included in the aggregate project

ktoso:
I see, thanks for noticing, now is a good time to clean up all these things. Adding it to the aggregate makes sense, compilation is slow for that project but we want to include it anyway.

PR #17543, opened 2015-05-21
=pro update sbt-jmh and include for compilation in root

ktoso:
Bump JMH to the latest version and also make use of it like an auto plugin.
Also makes sure we compile benchmarks in validation (yup, smart pr validation just works with it).


ktoso:
This should go in after @2m's https://github.com/akka/akka/pull/17532 is merged.

rkuhn:
LGTM

Issue #17361, opened 2015-05-03
collapse java/scala artifacts into one and document content assist filters


rkuhn:
In particular `akka-http-{java,scala}` and `akka-http-testkit-{java,scala}` shall be merged into one project each. The reason is that the split is not complete anyway (with the Java parts depending on their Scala counterparts) and the javadsl/scaladsl packages need to be filtered by the user or in the IDE in any case since third-party dependencies may pull in the wrong flavor as well.

The thought behind this is consistency: the users shall always get the same look and feel. Doing a best-effort split will just be confusing for those parts where a split is not feasible or where it is broken accidentally.

/cc @sirthias @jrudolph

Just to be clear: no code movements are implied, this is only about the published artifacts (and therefore the project layout).

We will then also have to document how to filter out the undesired half of the DSL in IntelliJ or Eclipse, respectively.

drewhk:
I think we should try to do this in RC3. @sirthias @jrudolph any of you want to pick this up, or should I attempt it myself? This affects mergeability of your branches so your call.

jrudolph:
To understand the implications: why would it affect mergeability if no code is moved around? What is the suggested approach?

drewhk:
Let's ask @rkuhn what "therefore the project layout" exactly means in this context given "no code movements are implied" ;)

jrudolph:
:)

rkuhn:
What I meant is that no classes move to different packages, but of course lots of files will move around in the repository :-)

rkuhn:
To maximize git-friendliness this should be done as a completely separate action that does not change any code (to make rename-detection more efficient).

jrudolph:
Ok, good to know. I understood that only class-files would be merged together during the publishing process.

Then, I'd say please do the renaming and we care for the merge problems which are hopefully only few thanks to git's overall greatness.

PR #17545, opened 2015-05-21
+str #17361: Unified http java/scala projects except marshallers

drewhk:


drewhk:
refs #17361

jrudolph:
Does this change have a development productivity overhead? I.e. is compilation as fast as before and does anything that compiled in parallel before still do? I'm asking because mixed Java/Scala compilation can be quite slow. For akka-http-core it's already hardly acceptable so hopefully this doesn't spread to other parts with this change.

drewhk:
The tests were fine on my machine. As for compilation speed, I cannot say anything because I compiled from clean state. Otherwise, I delegate this question to @rkuhn

rkuhn:
@jrudolph Your questions intrigued me, so I did some experiments with `akka-http-tests-experimental/test:compile` (inserting a `-java` for the `release-2.3-dev` branch comparison). A clean compile actually got faster with this PR (128s -> 112s), adding a method to the Scala Http extension went slightly up (from 6s for 10 sources to 8s for 14 sources), changing that method was 2s (only that file) in both cases. So, I’d say that this change does not negatively impact productivity AFAICS.

LGTM

drewhk:
It builds for me locally. Hm

drewhk:
PLS BUILD

jrudolph:
Cool, @rkuhn thanks for checking.

PR #17566, opened 2015-05-26
clu #17565 Invoke OnMemberRemoved callback when  cluster.shutdown

patriknw:
- must also be done when the listener actor stops before the
  MemberRemoved event has been received
- add test for this
- clarify docs with example that shuts down actor system and
  exit jvm

### DIFF
--- a/akka-samples/akka-sample-cluster-java/src/main/java/sample/cluster/factorial/FactorialFrontendMain.java
+++ b/akka-samples/akka-sample-cluster-java/src/main/java/sample/cluster/factorial/FactorialFrontendMain.java
@@ -1,7 +1,11 @@
 package sample.cluster.factorial;
 
+import java.util.concurrent.TimeUnit;
+
+import scala.concurrent.duration.FiniteDuration;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
+
 import akka.actor.ActorSystem;
 import akka.actor.Props;
 import akka.cluster.Cluster;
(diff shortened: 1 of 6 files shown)


patriknw:
Refs #17565

patriknw:
PLS BUILD

akka-samples/akka-sample-cluster-java/src/main/java/sample/cluster/factorial/FactorialFrontendMain.java
```diff
@@ -30,10 +34,22 @@ public void run() {
...
+      public void run() {
+        // exit JVM when ActorSystem has been terminated
+        final Runnable exit = new Runnable() {
+          @Override
+          public void run() {
+            System.exit(-1);
+          }
+        };
+        system.registerOnTermination(exit);
+        // if the ActorSystem does not terminate
```
ktoso: As mentioned elsewhere, I'd reword to  "in case ActorSystem shutdown takes longer than 10 seconds, exit the JVM forcefully anyway"?

ktoso:
LGTM otherwise :)

drewhk:
LGTM

akka-samples/akka-sample-cluster-java/src/main/java/sample/cluster/factorial/FactorialFrontendMain.java
patriknw: +1

PR #17577, opened 2015-05-26
+str #17298 include stream supervisor in log() source

### LINKED ISSUE #17298: Log() improvement: include guardian in logged info?
We could expose the given streams guardian and include it's name in the log messages - such that messages from two Materializers are distinguishable.

ktoso:
In order to make it use a _nice_ LogSource some implicit resolution trickery was needed, reason: if LogSource is a String it appends `...(systemName)` so we'd get `name(supervisorAddr)(systemName)`, which is overdoing it since `supervisorAddr` contains the system name already.

Resolves #17298

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -4,22 +4,17 @@
 package akka.stream.impl.fusing
 
 import akka.event.Logging.LogLevel
-import akka.event.Logging
-import akka.event.LoggingAdapter
+import akka.event.{ LogSource, Logging, LoggingAdapter }
 import akka.stream.OperationAttributes.LogLevels
-import akka.stream.impl.FixedSizeBuffer
-import akka.stream.impl.ReactiveStreamsCompliance
+import akka.stream.impl.{ FixedSizeBuffer, ReactiveStreamsCompliance }
 import akka.stream.stage._
-import akka.stream.Supervision
-import akka.stream._
+import akka.stream.{ Supervision, _ }
 
 import scala.annotation.tailrec
 import scala.collection.immutable
 import scala.concurrent.Future
 import scala.util.control.NonFatal
-import scala.util.Failure
-import scala.util.Success
-import scala.util.Try
+import scala.util.{ Failure, Success, Try }
 
 /**
  * INTERNAL API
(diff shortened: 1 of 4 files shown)


akka-stream/src/main/scala/akka/stream/stage/Stage.scala
```diff
@@ -531,6 +532,29 @@ sealed trait TerminationDirective extends SyncDirective
...
 
+object LifecycleContext {
+  private final val DefaultLoggerName = "akka.stream.Log"
+
+  /**
+   * INTERNAL API
+   * Must be located here to be visible for implicit resolution, when LifecycleContext is passed to [[Logging]]
+   * More specific LogSource than `fromString`, which would add the ActorSystem name in addition to the supervision to the log source.
+   */
+  implicit private[akka] final val fromLifecycleContext = new LogSource[LifecycleContext] {
```
patriknw: for an internal utility I would prefer to be explicit

patriknw:
LGTM (with small concern about the implicit)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -540,7 +535,9 @@ private[akka] final case class Log[T](name: String, extract: T ⇒ Any, logAdapt
     logLevels = ctx.attributes.logLevels.getOrElse(DefaultLogLevels)
     log = logAdapter match {
       case Some(l) ⇒ l
-      case _       ⇒ Logging(ctx.materializer.asInstanceOf[ActorFlowMaterializer].system, DefaultLoggerName)
+      case _ ⇒
+        val mat = ActorFlowMaterializer.downcast(ctx.materializer) // TODO if not in Actor materializer we could fallback to println?
```
drewhk: can't we throw a more context specific exception if we cannot downcast? Like logger is only usable with the actor based materializer?

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
drewhk: +1

drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
ktoso: Ah, that can be passed in explicitly actually - good idea (`Logging(sys, ctx)(fromLivecycleContext)`) 
I somehow missed that it can be explicitly passed in

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: Good idea, I'll add that

ktoso:
Removed the implicit, it's now used explicitly - and nearer the `Log` class

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: Done!

Issue #17554, opened 2015-05-22
Improve flow control of system message delivery


patriknw:
I have noticed that sending system messages for 5000 watch requests sometimes cause high sustained cpu load resulting in remote failure detection in the end.

I used `system-message-buffer-size = 20000`.

patriknw:
I should also mention that it does not always happen. Sometimes it works fine without long cpu usage.

patriknw:
The cpu load also continues a long time (1 minute) after the system has been quarantined, but eventually it stops.

patriknw:
It was noticed when using release-2.3 branch corresponding to version 2.3.11

patriknw:
I have debugged the reproducer that I created last week. I have found that it can be solved by limiting the initial send to some max outstanding messages, i.e. only buffer them if `resendBuffer.nonAcked.size > 500`.

I also noticed that when it goes into the bad spin there is also one nacked message that is resent over and over. That is also something I would like to understand the reason for. I will debug that more.

cc @drewhk

PR #17582, opened 2015-05-27
=rem #17554 Improve flow control of system message delivery

patriknw:
When watching many (5000) actors at the same time the
following problems were found:
- first send of a sys msg is sent without any flow control
  => limit the number of outstanding sys msg by using
     the buffer to send them later (ordinary resend)
- when msg cannot be written sys msg is dropped (relying on resend),
  but that cause message re-ordering and negative acknowledgment,
  which is very costly
  => buffer the sys msg on write failure
  => minor optimization of AckedReceiveBuffer

I also made the resend-limit configurable.

### DIFF
--- a/akka-remote/src/main/resources/reference.conf
+++ b/akka-remote/src/main/resources/reference.conf
@@ -258,6 +258,14 @@ akka {
     # Messages that were negatively acknowledged are always immediately
     # resent.
     resend-interval = 2 s
+    
+    # Maximum number of unacknowledged system messages that will be resent
+    # each 'resend-interval'. If you watch many (> 1000) remote actors you can
+    # increase this value to for example 600, but a too large limit (e.g. 10000)
+    # may flood the connection and might cause false failure detection to trigger.
+    # Test such a configuration by watching all actors at the same time and stop
+    # all watched actors at the same time.
+    resend-limit = 200
 
     # WARNING: this setting should not be not changed unless all of its consequences
     # are properly understood which assumes experience with remoting internals
--- a/akka-remote/src/main/scala/akka/remote/AckedDelivery.scala
+++ b/akka-remote/src/main/scala/akka/remote/AckedDelivery.scala
@@ -101,7 +101,9 @@ case class AckedSendBuffer[T <: HasSequenceNumber](
   def acknowledge(ack: Ack): AckedSendBuffer[T] = {
     if (ack.cumulativeAck > maxSeq)
       throw new IllegalArgumentException(s"Highest SEQ so far was $maxSeq but cumulative ACK is ${ack.cumulativeAck}")
-    val newNacked = (nacked ++ nonAcked) filter { m ⇒ ack.nacks(m.seq) }
+    val newNacked =
+      if (ack.nacks.isEmpty) Vector.empty
+      else (nacked ++ nonAcked) filter { m ⇒ ack.nacks(m.seq) }
     if (newNacked.size < ack.nacks.size) throw new ResendUnfulfillableException
     else this.copy(
       nonAcked = nonAcked.filter { m ⇒ m.seq > ack.cumulativeAck },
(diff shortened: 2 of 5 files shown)


akka-remote/src/main/resources/reference.conf
```diff
@@ -258,6 +258,14 @@ akka {
...
     # resent.
     resend-interval = 2 s
+    
+    # Maximum number of unacknowledged system messages that will be resent
+    # each 'resend-interval'. If you watch many (> 1000) remote actors you can
+    # increase this value to for example 600, but a too large limit (e.g. 10000)
+    # may flood the connection and might cause false failure detection to trigger.
+    # Test such a configuration by watching all actors at the same time and stop
+    # all watched actors at the same time.
+    resend-limit = 200
```
drewhk: last time we calculated this kind of limit from the resend timer and a constant to avoid exposing yet another option.

akka-remote/src/main/scala/akka/remote/AckedDelivery.scala
```diff
@@ -101,7 +101,9 @@ case class AckedSendBuffer[T <: HasSequenceNumber](
   def acknowledge(ack: Ack): AckedSendBuffer[T] = {
     if (ack.cumulativeAck > maxSeq)
       throw new IllegalArgumentException(s"Highest SEQ so far was $maxSeq but cumulative ACK is ${ack.cumulativeAck}")
-    val newNacked = (nacked ++ nonAcked) filter { m ⇒ ack.nacks(m.seq) }
+    val newNacked =
+      if (ack.nacks.isEmpty) Vector.empty
```
drewhk: good one.

akka-remote/src/main/scala/akka/remote/AckedDelivery.scala
```diff
@@ -181,7 +183,8 @@ case class AckedReceiveBuffer[T <: HasSequenceNumber](
       prev = bufferedMsg.seq
     }
 
-    (this.copy(buf = buf filterNot deliver.contains, lastDelivered = updatedLastDelivered), deliver, ack)
+    val newBuf = if (deliver.isEmpty) buf else buf.filterNot(deliver.contains)
```
drewhk: +1

akka-remote/src/main/scala/akka/remote/Endpoint.scala
```diff
@@ -372,8 +365,9 @@ private[remote] class ReliableDeliverySupervisor(
       val sequencedSend = send.copy(seqOpt = Some(nextSeq()))
       tryBuffer(sequencedSend)
       // If we have not confirmed the remote UID we cannot transfer the system message at this point just buffer it.
-      // GotUid will kick resendAll() causing the messages to be properly written
-      if (uidConfirmed)
+      // GotUid will kick resendAll() causing the messages to be properly written.
+      // Flow control by not sending more when we already have many outstanding.
+      if (uidConfirmed && resendBuffer.nonAcked.size <= settings.SysResendLimit)
```
drewhk: totally makes sense.

drewhk:
LGTM, but we should consider calculating that setting instead of exposing another config option.

akka-remote/src/main/resources/reference.conf
patriknw: I know, now I think it actually needs tuning. E.g. watching 5000 actors takes 200 seconds, which I think someone want to tune if they have that use case.

drewhk: Maybe we should inform remoting about the number of watched actors from RemoteWatcher? I think an atomic variable is enough, and we need to refresh it from time to time.

patriknw: I think the factor that we cannot know is the production environment cpu and network capacity.

The default of 200 msg per 2 s is probably too low, but I guess we want to err on that side.

ktoso:
LGTM

drewhk:
Please rebase and merge

ktoso:
LGTM, failure was `ClusterShardingLeavingSpec`, but I don't think it's caused by these changes - feel free to merge?

rkuhn:
Yes: merge!

ktoso:
refs #17554

PR #17590, opened 2015-05-27
=doc: Update withRangeSupport.rst reference to RFC7233

danielwegener:
Linked http-bis-draft is now RFC7233. I checked the [delta](https://www.ietf.org/rfcdiff?url1=draft-ietf-httpbis-p5-range-24&url2=rfc7233) and there are no functional changes compared to the original implementation against `draft-ietf-httpbis-p5-range-24` (https://github.com/spray/spray/pull/612)


ktoso:
LGTM, though sanity check by @jrudolph or @sirthias probably wouldn't hurt

jrudolph:
LGTM.

@danielwegener is the authoritative source on this topic btw. He contributed the original Range support to spray ;)

Thanks for checking for changes in the final RFC.

ktoso:
Oh, I did not know that :-) Thanks @danielwegener!

PR #17592, opened 2015-05-28
=per #16802 no need to verify serialization on journal protocol

### LINKED ISSUE #16802: No serialization-bindings for akka.persistence.journal.AsyncWriteTarget$WriteMessages
akka persistence 2.3.9 wrapped the persistent event with `WriteMessages`.
When the journal actor is a remote actor, the system try to use `akka.remote.MessageSerializer` to serialize the `WriteMessages`, and it finally choose the default `JavaSerializer`.

And then, if the user's persistent event is not Serializable, the serialization will fail.

ktoso:
Resolves #16802

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/Protocol.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/Protocol.scala
@@ -4,6 +4,8 @@
 
 package akka.persistence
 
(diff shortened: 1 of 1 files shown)


patriknw:
LGTM

PR #17596, opened 2015-05-28
=act #17415 fix wrong test for CircuitBreaker (for validation)

patriknw:
(cherry picked from commit a41d1d67a99f6d9964f7e99e6f80afe80ac6d390)

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/pattern/CircuitBreakerStressSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/pattern/CircuitBreakerStressSpec.scala
@@ -6,23 +6,24 @@ package akka.pattern
 import scala.concurrent.Promise
 import scala.concurrent.duration._
 import scala.concurrent.forkjoin.ThreadLocalRandom
-
 import akka.actor.Actor
 import akka.actor.ActorLogging
 import akka.actor.Props
 import akka.actor.Status.Failure
 import akka.testkit.AkkaSpec
 import akka.testkit.ImplicitSender
+import java.util.concurrent.TimeoutException
 
 object CircuitBreakerStressSpec {
   case object JobDone
   case object GetResult
-  case class Result(doneCount: Int, failCount: Int, circCount: Int)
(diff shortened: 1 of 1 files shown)


patriknw:
Refs #17415

akka-actor-tests/src/test/scala/akka/pattern/CircuitBreakerStressSpec.scala
```diff
@@ -43,11 +44,14 @@ object CircuitBreakerStressSpec {
...
         breaker.withCircuitBreaker(job).pipeTo(self)
+      case Failure(_: TimeoutException) ⇒
+        timeoutCount += 1
+        breaker.withCircuitBreaker(job).pipeTo(self)
       case Failure(_) ⇒
         failCount += 1
         breaker.withCircuitBreaker(job).pipeTo(self)
       case GetResult ⇒
-        sender() ! Result(doneCount, failCount, circCount)
+        sender() ! Result(doneCount, timeoutCount, failCount, circCount)
```
drewhk: where is this calculated result used? I see it being calculated and reported, but I don't see code that does anything with it.

patriknw: it is only checking the failCount here: https://github.com/akka/akka/pull/17596/files#diff-a6a84cc4febbb297157a42a8d412a9d0R78

but for debugging (println) it was good to see the distribution different things

drewhk: so then the fix part is basically to not include the TimeoutExceptions in the other failure counters?

patriknw: yes, it failed on master, because we had improved the timeout handling

patriknw: this is for release-2.3 just to keep them in sync

Issue #17576, opened 2015-05-26
Akka Serialization: manifest should be String, not Class based


ktoso:

patriknw:
Related #17434

Issue #17434, opened 2015-05-09
Make Serializer more friendly with persistence


ahjohannessen:
I intend to use protobuf to serialize all my domain events in our akka applications and would like to have a clear separation between types used in the domain and the types used for serialization. So it goes something like this (where  DT is domain type and PBT is protobuf type):

 persist(DT) -> ... -> Akka MessageSerializer -> Select serializer based on DT -> .. MySerializer that knows how to DT<->PBT -> Akka MessageSerializer -> .. -> Storage

As it works today, if I say that I wish to use manifest then Akka MessageSerializer will use the manifest from DT and not what I wish, PBT.

If we could have a `def provideManifest(payload: AnyRef)` method on `Serializer` that had a default implementation, doing what akka otherwise does today, then Akka MessageSerializer would delegate to the selected serializer to provide a manifest. In MySerializer I would get DT in and provide manifest for the corresponding PBT.

ahjohannessen:
One solution is to wrap akka `MessageSerializer` in order to get the manifest of the protobuf type, but I see that more like a workaround, because then I need to make sure that my wrapper is in sync with Akka `MessageSerializer` when Akka is upgraded etc.

patriknw:
sounds interesting

patriknw:
@ahjohannessen please take a look at the proposed solution of a new `SerializerWithStringManifest` in https://github.com/akka/akka/pull/17601
Is that what you need?

ahjohannessen:
Thanks @patriknw :+1:

PR #17601, opened 2015-05-28
+act #17576 Support serializer with string manifest

patriknw:
- useful when evolution is needed, e.g. Akka Persistence

### DIFF
--- a/akka-actor/src/main/scala/akka/serialization/Serialization.scala
+++ b/akka-actor/src/main/scala/akka/serialization/Serialization.scala
@@ -13,6 +13,7 @@ import java.io.NotSerializableException
 import scala.util.{ Try, DynamicVariable, Failure }
 import scala.collection.immutable
 import scala.util.control.NonFatal
+import scala.util.Success
 
 object Serialization {
 
@@ -91,7 +92,7 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
 
   /**
    * Deserializes the given array of bytes using the specified serializer id,
-   * using the optional type hint to the Serializer and the optional ClassLoader ot load it into.
+   * using the optional type hint to the Serializer.
    * Returns either the resulting object or an Exception if one was thrown.
    */
   def deserialize[T](bytes: Array[Byte], serializerId: Int, clazz: Option[Class[_ <: T]]): Try[T] =
@@ -105,8 +106,36 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
     }
 
   /**
+   * Deserializes the given array of bytes using the specified serializer id,
+   * using the optional type hint to the Serializer.
+   * Returns either the resulting object or an Exception if one was thrown.
+   */
+  def deserialize(bytes: Array[Byte], serializerId: Int, manifest: String): Try[AnyRef] =
+    Try {
+      val serializer = try serializerByIdentity(serializerId) catch {
+        case _: NoSuchElementException ⇒ throw new NotSerializableException(
+          s"Cannot find serializer with id [$serializerId]. The most probable reason is that the configuration entry " +
+            "akka.actor.serializers is not in synch between the two systems.")
+      }
+      serializer match {
+        case s2: SerializerWithStringManifest ⇒ s2.fromBinary(bytes, manifest)
+        case s1 ⇒
+          if (manifest == "")
+            s1.fromBinary(bytes, None)
+          else {
+            system.dynamicAccess.getClassFor[AnyRef](manifest) match {
+              case Success(classManifest) ⇒
+                s1.fromBinary(bytes, Some(classManifest))
+              case Failure(e) ⇒
+                throw new NotSerializableException(
+                  s"Cannot find manifest class [$manifest] for serializer with id [$serializerId].")
+            }
+          }
+      }
+    }
+
+  /**
    * Deserializes the given array of bytes using the specified type to look up what Serializer should be used.
-   * You can specify an optional ClassLoader to load the object into.
    * Returns either the resulting object or an Exception if one was thrown.
    */
   def deserialize[T](bytes: Array[Byte], clazz: Class[T]): Try[T] =
@@ -118,10 +147,8 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
    * Throws akka.ConfigurationException if no `serialization-bindings` is configured for the
    *   class of the object.
    */
-  def findSerializerFor(o: AnyRef): Serializer = o match {
-    case null  ⇒ NullSerializer
-    case other ⇒ serializerFor(other.getClass)
-  }
+  def findSerializerFor(o: AnyRef): Serializer =
+    if (o eq null) NullSerializer else serializerFor(o.getClass)
 
   /**
    * Returns the configured Serializer for the given Class. The configured Serializer
@@ -205,5 +232,6 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
    */
   val serializerByIdentity: Map[Int, Serializer] =
     Map(NullSerializer.identifier -> NullSerializer) ++ serializers map { case (_, v) ⇒ (v.identifier, v) }
+
 }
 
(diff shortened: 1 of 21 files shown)


patriknw:
@ktoso I need to add some more docs, but would like your opinion on the solution first

patriknw:
@agemooij you might be interested in this new `SerializerWithStringManifest`. Feedback is welcome.

ahjohannessen:
:+1:

akka-persistence/src/test/scala/akka/persistence/serialization/SerializerSpec.scala
```diff
@@ -241,6 +316,31 @@ class MyPayloadSerializer extends Serializer {
...
+
+  def fromBinary(bytes: Array[Byte], manifest: String): AnyRef = manifest match {
+    case ManifestV2 ⇒
+      val parts = new String(bytes, "UTF-8").split(":")
+      MyPayload2(data = parts(0) + ".", n = parts(1).toInt)
+    case ManifestV1 ⇒
+      MyPayload2(data = s"${new String(bytes, "UTF-8")}.", n = 0)
+    case other ⇒
+      throw new Exception(s"unexpected manifest [$other]")
+  }
```
ktoso: :+1:

ahjohannessen:
Any chance this gets into 2.3.x?

patriknw:
note to self: DistributedPubSubMessageSerializer also needs same treatment

hepin1989:
how about make the manifest as a long/indexed so it could be a long value?

patriknw:
@hepin1989 that doesn't work, because we want to support migration from `Serializer` to `SerializerWithStringManifest` in a compatible way. One character encoding should be compact enough.

hepin1989:
make sense.thanks

ahjohannessen:
Also, having the type name / string is useful for akka persistence journals like eventstore where you can index on type automatically. A long would be hard to reason about.

akka-actor/src/main/scala/akka/serialization/Serialization.scala
```diff
@@ -105,8 +106,34 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
     }
 
   /**
+   * Deserializes the given array of bytes using the specified serializer id,
+   * using the optional type hint to the Serializer and the optional ClassLoader ot load it into.
+   * Returns either the resulting object or an Exception if one was thrown.
```
ktoso: I don't see an "optional ClassLoader" param

akka-actor/src/main/scala/akka/serialization/Serialization.scala
```diff
@@ -105,8 +106,34 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
...
+    Try {
+      val serializer = try serializerByIdentity(serializerId) catch {
+        case _: NoSuchElementException ⇒ throw new NotSerializableException(
+          s"Cannot find serializer with id [$serializerId]. The most probable reason is that the configuration entry " +
+            "akka.actor.serializers is not in synch between the two systems.")
+      }
+      serializer match {
+        case s2: SerializerWithStringManifest ⇒ s2.fromBinary(bytes, manifest).asInstanceOf[T]
+        case s1 if manifest == "" || (manifest eq null) ⇒
+          s1.fromBinary(bytes, None).asInstanceOf[T]
```
ktoso: AFAICS the `manifest eq null` shouldn't really happen in normal operation, as we `if !hasManifest ""` in other places. It's OK to keep the null case here, just making sure that it's not expected due something I missed

akka-actor/src/main/scala/akka/serialization/Serialization.scala
patriknw: ah, copy pasta

akka-persistence/src/main/scala/akka/persistence/serialization/MessageSerializer.scala
```diff
@@ -30,6 +30,8 @@ class MessageSerializer(val system: ExtendedActorSystem) extends BaseSerializer
   val PersistentImplClass = classOf[PersistentImpl]
   val AtLeastOnceDeliverySnapshotClass = classOf[AtLeastOnceDeliverySnap]
 
+  private lazy val serialization = SerializationExtension(system)
```
ktoso: why lazy?

akka-actor/src/main/scala/akka/serialization/Serialization.scala
patriknw: right, it is better to add a null check were `serializer.manifest` is used

akka-persistence/src/main/scala/akka/persistence/serialization/MessageSerializer.scala
patriknw: initializaton order of extension and serializer, I think the SerializationExtension creates this instance

akka-persistence/src/test/scala/akka/persistence/serialization/SerializerSpec.scala
```diff
@@ -241,6 +316,31 @@ class MyPayloadSerializer extends Serializer {
...
+
+  val ManifestV1 = classOf[MyPayload].getName
+  val ManifestV2 = "MyPayload-V2"
+
+  def identifier: Int = 77125
+
+  def manifest(o: AnyRef): String = ManifestV2
+
+  def toBinary(o: AnyRef): Array[Byte] = o match {
+    case MyPayload2(data, n) ⇒ s".$data:$n".getBytes("UTF-8")
```
ktoso: Can we find some place to put the `"UTF-8"`?

akka-persistence/src/test/scala/akka/persistence/serialization/SerializerSpec.scala
```diff
@@ -148,6 +167,57 @@ class MessageSerializerPersistenceSpec extends AkkaSpec(customSerializers) {
...
+
+        // now the system is updated, OldPayload is replaced by MyPayload, and the
+        // OldPayloadSerializer is adjusted to migrate OldPayload
+        val bytes = decodeHex(oldData.toCharArray)
+
+        val deserialized = serializer.fromBinary(bytes, None).asInstanceOf[PersistentRepr]
+
+        deserialized.payload should be(MyPayload("OldPayload(A)"))
+      }
+    }
```
ktoso: great test, +1

ktoso:
Looks great, on this layer I think that's all we need.
The more complex upcasts we talked about will be performed in the adapter layer.

When writing docs include a mini guide how to use this in persistence's case (like the tests show already).

akka-persistence/src/test/scala/akka/persistence/serialization/SerializerSpec.scala
patriknw: since we are > jdk6 now, we can use java.nio.charset.StandardCharsets

ktoso: :heart: JDK8

patriknw:
I think I'll update the ordinary serialization docs, but create a separate ticket for "Illustrate real serializer for Persistence, including evolution of event types"

ktoso:
Sounds like a good plan +1

ktoso:
refs https://github.com/akka/akka/issues/17576

drewhk:
LGTM so far

patriknw:
@ktoso I have updated:
* fixed your feedback
* added docs (the comprehensive persistence docs will be handled by #17613)
* converted serializers in cluster-metrics and cluster-tools, the reason I can't convert more (e.g. ClusterMessageSerializer) is that they are public, and can therefore not be changed in a binary compatible way (we might change when we do a breaking release)

patriknw:
it's a separate commit, if you only want to look at the changes

ktoso:
LGTM

agemooij:
Hey guys, sorry for the late reply. In our projects so far we have used a custom serializer that prefixes the serialized bytes with a key-version combo, serialized using simple binary encoding. Our deserializers pick the right code based on the key and the version. This maps very well onto this new manifest String so I'm all for it since it saves us some complexity!

I would have preferred a binary encoding but I totally see the need for making this a String. String parsing will probably not add all that much extra deserialization time.

As soon as this stuff becomes available in 2.3.x I'll work it into our upcoming "akka serialization for persistence with support for explicit and compile-time checked migration code" library, code named "stamina" because that first name was a bit long :smile:

Looking forward to show you guys what we have built at Scala Days Amsterdam.

patriknw:
Thanks for taking a look at it. Stamina looks like a great library.

We don't plan to include this in 2.3.x, but in 2.4-M2, and 2.4.0 is not that far away.

See you in Amsterdam.

patriknw:
Refs #17576

patriknw:
also refs #17434

Issue #17606, opened 2015-05-29
ActorPath is not binary compatible


patriknw:
Found by @hepin1989, more details in https://github.com/patriknw/akka-data-replication/issues/83

ktoso:
Note to self: notify akka-user that this is known-issue and solved in next Milestone.

ktoso:
l

PR #17608, opened 2015-05-29
=act #17606 Revert abstract class change of ActorPath

patriknw:
It was not binary compatible. Placed static forwarders for Java API
in ActorPaths instead.

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/ActorPath.scala
+++ b/akka-actor/src/main/scala/akka/actor/ActorPath.scala
@@ -8,6 +8,47 @@ import akka.japi.Util.immutableSeq
 import java.net.MalformedURLException
 import java.lang.{ StringBuilder ⇒ JStringBuilder }
 
+/**
+ * Java API
+ */
+object ActorPaths {
+  // static forwarders to `object ActorPath`, since `trait ActorPath`
+  // could not be changed to `abstract ActorPath` in a binary compatible way
+
+  /**
+   * Parse string as actor path; throws java.net.MalformedURLException if unable to do so.
+   */
+  def fromString(s: String): ActorPath = ActorPath.fromString(s)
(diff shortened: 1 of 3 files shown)


patriknw:
Refs #17606

drewhk:
LGTM

akka-actor/src/main/scala/akka/actor/ActorPath.scala
```diff
@@ -8,6 +8,47 @@ import akka.japi.Util.immutableSeq
 import java.net.MalformedURLException
 import java.lang.{ StringBuilder ⇒ JStringBuilder }
 
+/**
+ * Java API
+ */
+object ActorPaths {
```
ktoso: AFAIR it needs to be a "companion", that is have a `class` of the same name for the static forwarders to be generated?

hepin1989: it has no field ,so it's ok.

patriknw: yes, I think it is ok, but I should add usage of it in some java test class

ktoso: good to know it's not a problem if no fields involved :)
test - yes please :)

ktoso:
LGTM after adding usage from a Java spec, better be safe than sorry

akka-actor/src/main/scala/akka/actor/ActorPath.scala
patriknw: added test c81b7273a1e716575ce8a85d018bb866f3435145

ktoso:
Thanks, LGTM

2m:
LGTM

Issue #17393, opened 2015-05-06
make stream and http tests pass with `serialize-messages=on`


rkuhn:
mark stream-internal local messages with NoSerializationVerificationNeeded

rkuhn:
In particular the ActorProcessor and ActorPublisher message types are needed.

drewhk:
in progress: https://github.com/akka/akka/pull/17615, http is next

PR #17615, opened 2015-05-29
!str #17393: Make stream-tests pass with serialize-messages=on

drewhk:
Please note that ActorPublisher and ActorSubscriber messages are no longer having a SerialVersionUID and I think they should not be allowed to be serialized. Most of the Subscriptions and Subscribers are not serializable therefore the handshake won't work anyway in quite a lot of the cases.
- [x] Http

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
@@ -4,7 +4,7 @@ import java.util.concurrent.atomic.AtomicReference
 import scala.annotation.tailrec
 import scala.concurrent.{ Future, Promise }
 import akka.http.HostConnectionPoolSetup
-import akka.actor.{ Props, ActorSystem, ActorRef }
+import akka.actor.{ Deploy, Props, ActorSystem, ActorRef }
 import akka.http.scaladsl.Http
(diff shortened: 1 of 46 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -40,7 +40,7 @@ private[http] class PoolGateway(hcps: HostConnectionPoolSetup,
 
   private val state = {
     val shutdownCompletedPromise = Promise[Unit]()
-    val props = Props(new PoolInterfaceActor(hcps, shutdownCompletedPromise, this))
+    val props = Props(new PoolInterfaceActor(hcps, shutdownCompletedPromise, this)).withDeploy(Deploy.local)
```
2m: Why are these needed? Props are serializable with `NoScopeGiven` as well.

2m:
LGTM

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
drewhk: This explicitly states that these are deployed locally.

drewhk:
Fixes #17393

Issue #17461, opened 2015-05-12
Add a parallel version of foreach


drewhk:
Basically just sugar around mapAsyncUnordered. Looks like:

    source.foreachParallel(4) { doStuff() }

Can be useful for accepting connections from a connection source (TCP or HTTP)

agolubev:
should it be something like 
```scala
def foreachParallel[T](parallelism: Int)(f: Out => T): Repr[T, Mat] =
    andThen(MapAsyncUnordered(
      parallelism,
      {out: Out => Future(f(out))}.asInstanceOf[Any ⇒ Future[Any]]))
```
for scala dsl?

drewhk:
yes, exactly :) (+Java counterpart)

agolubev:
Ok, let me make few statements here to be sure everything will be working as expected:
* ```foreachParallel``` will take function like  ```f:Out => Unit``` as an argument
* return type for ```foreachParallel``` will be ```Future[Unit]```
* when some ```f``` fails with exception or several ```f``` failed, ```foreachParallel``` will fail ```Future```  with first exception when stream is completed (if ```supervisionStrategy(resumingDecider)```) - this is what I'm not sure about
* when some ```f```` fails with exception or many exceptions, ```foreachParallel``` will fail Future at once with  exception  (if ```supervisionStrategy(stop)```)
* ```foreachParallel``` will be placed in class ```scaladsl.Source```
* as in ```Source``` there is method ```runForeach```, our final method name will be ```runForeachParallel``` (instead of foreachParallel)

Basically we can have four exception handling policies here:
* stop on each exception
* ignore all exception
* failed future on complete with first exception
* failed future on complete with last exception

So for three cases out of four most exceptions are lost. Still, maybe it would be good to drop exceptions with ```ForeachParallel``` in "optimistic mode". However, if it matters we can compose stream with ```recoverWith```

PR #17621, opened 2015-06-01
+str #17461 add a parallel version of foreach

agolubev:
Added foreachParallel method to scala, java DSLs, updated docs, added tests (used FlowMapAsyncUnorderedSpec tests)


akka-docs-dev/rst/java/stream-integrations.rst
```diff
@@ -142,7 +142,7 @@ Integrating with External Services
 ==================================
 
 Stream transformations and side effects involving external non-stream based services can be
-performed with ``mapAsync`` or ``mapAsyncUnordered``.
+performed with ``mapAsync``, ``mapAsyncUnordered`` or ``foreachParallel``.
```
ktoso: Not sure if we want to mention foreach here.

akka-docs-dev/rst/stages-overview.rst
```diff
@@ -52,6 +52,7 @@ Stage                  Emits when
...
 mapAsync               the Future returned by the provided function finishes for the next element in sequence                                      the number of futures reaches the configured parallelism and the downstream backpressures                                       upstream completes and all futures has been completed  and all elements has been emitted [1]_
 mapAsyncUnordered      any of the Futures returned by the provided function complete                                                               the number of futures reaches the configured parallelism and the downstream backpressures                                       upstream completes and all futures has been completed  and all elements has been emitted [1]_
+foreachParallel        any of the Futures returned by the provided function complete                                                               the number of futures reaches the configured parallelism and the downstream backpressures                                       upstream completes and all futures has been completed  and all elements has been emitted [1]_
```
ktoso: Same question here. If we do in fact want to mention this method here then we must also mention `runForeach` (terminal operation) and othes like `runFold`

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -440,6 +440,36 @@ trait FlowOps[+Out, +Mat] {
...
+   *
+   * '''Cancels when''' downstream cancels
+   *
+   * @see [[#mapAsyncUnordered]]
+   */
+  def foreachParallel[T](parallelism: Int)(f: Out ⇒ T)(implicit ec: ExecutionContext): Repr[T, Mat] =
+    andThen(MapAsyncUnordered(
+      parallelism,
+      { out: Out ⇒ Future(f(out))(ec) }.asInstanceOf[Any ⇒ Future[Any]]))
+
```
ktoso: Oh so it's *not* like the usual foreach...!
I expected this to be
- a terminal operation, so it would be `runForeachParallel`
- `Out => Unit` since foreach is usually the `=> Unit` counterpart of `map` hm...

Summing up, I'm not sure this method as implemented should be called foreach, it's more like a "mapAsync but please do the Future stuff for me".

agolubev: I see. Just found in Sink:  
```scala
def foreach[T](f: T ⇒ Unit): Sink[T, Future[Unit]] = {
```
So I guess runForeachParallel will be in Sink and will have signature like this:
```scala
def foreachParallel[T](parallelism: Int)(f: Out ⇒ Unit)(implicit ec: ExecutionContext): Sink[Out, Future[Unit]] 
```
correct?

akka-docs-dev/rst/java/stream-integrations.rst
drewhk: I think we should, because it is very much related to the mapAsync stages.

akka-docs-dev/rst/stages-overview.rst
drewhk: We can add new tables for built-in Sinks and Sources, but then this will grow too large. Let's keep it as it is now, and I think foreachParallel makes sense here.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: Yes, it should be a sink like that. Actually you can easily convert your code by adding a Sink.ignore and keeping the materialized value to be the foreach completion future.

drewhk:
Please convert it to a Sink and rebase

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
ktoso: Yup, thanks!

akka-docs-dev/rst/java/stream-integrations.rst
ktoso: :+1:

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
agolubev: Issue with foreach is that it fails Future as soon as any of it's elements fails. So perhaps need to create another foreach in Sink that fail only when stream is completed or not fail at all.

drewhk: No, the mapAsync understands supervision, if you set the proper attributes then it can for example interpret the Resume directive and just simply ignore failed futures. So if you create a Sink.foreachParallel, then the code would look roughly like this:

  Sink.foreachParallel(2){doStuff}.withAttributes(OperationAttributes.supervision(...))

Since attributes are inherited in wrapped stages unless they override them, this will not need any extra code apart from documentation.

agolubev: Thanks for clarification. I actually send several questions here https://github.com/akka/akka/issues/17461#issuecomment-108707474 and it was one of them

drewhk:
This PR is now superceded.

Issue #16290, opened 2014-11-12
Add load tests for streams looking for performance regressions and memory leaks


jrudolph:

benjamingeer:
We recently switched back from Akka HTTP to Spray because we found that Akka HTTP leaks memory. I haven't had time to make a minimal test case for you, but it involved hammering our REST API using [Apache Bench](http://httpd.apache.org/docs/2.4/programs/ab.html) for 20 minutes with 50 concurrent connections, and watching the application's total heap allocation in [VisualVM](http://visualvm.java.net/). With Akka HTTP, the average allocation increased steadily, up to the JVM's maximum heap size:

![screen_shot_2015-02-09_at_12 18 02](https://cloud.githubusercontent.com/assets/558389/6249280/73e534f8-b786-11e4-87f7-a4f5bec5a50b.png)

With Spray, the average allocation is roughly constant:

![screen_shot_2015-02-17_at_16 31 05](https://cloud.githubusercontent.com/assets/558389/6249314/9f695dde-b786-11e4-895c-08e4c04f9d96.png)

jrudolph:
@benjamingeer thanks for reporting. I think this warrants an extra ticket. I recreated your report in https://github.com/akka/akka/issues/16900.

drewhk:
Getting closer, we need to implement fusable stream-of-stream ops and the fusing method itself. Moving out from M1 though.

PR #17623, opened 2015-06-01
=str #16290 add micro benchmarks for materialization and graph builder

2m:
Results on my machine:

```
[info] Benchmark                                             Mode  Cnt        Score        Error   Units
[info] GraphBuilderBenchmark.graph_with_100_junctions       thrpt    3    83686.308 ±  28597.452  ops/ms
[info] GraphBuilderBenchmark.graph_with_100_nested_imports  thrpt    3  1413110.778 ± 243132.891  ops/ms
```

and 

```
[info] Benchmark                                                Mode  Cnt         Score          Error   Units
[info] MaterializationBenchmark.flow_with_100_operations       thrpt    3    185240.747 ±    61163.429  ops/ms
[info] MaterializationBenchmark.graph_with_100_junctions       thrpt    3    122679.627 ±    30274.789  ops/ms
[info] MaterializationBenchmark.graph_with_100_nested_imports  thrpt    3   2659986.545 ±   954281.544  ops/ms
[info] MaterializationBenchmark.simple_flow                    thrpt    3  16856152.724 ± 28923405.310  ops/ms
[info] MaterializationBenchmark.simple_graph                   thrpt    3   5683167.370 ± 12897652.713  ops/ms
```

### DIFF
--- a/akka-bench-jmh/src/main/scala/akka/stream/MaterializationBenchmark.scala
+++ b/akka-bench-jmh/src/main/scala/akka/stream/MaterializationBenchmark.scala
@@ -1,0 +1,122 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.stream
+
+import java.util.concurrent.TimeUnit
+import akka.actor.ActorSystem
+import akka.stream.scaladsl._
+import org.openjdk.jmh.annotations._
+
+object MaterializationBenchmark {
+
+  val flowWithMapBuilder = (numOfCombinators: Int) => {
+    var source = Source.single(())
+    for (_ <- 1 to numOfCombinators) {
+      source = source.map(identity)
+    }
+    source.to(Sink.ignore)
+  }
+
+  val graphWithJunctionsBuilder = (numOfJunctions: Int) =>
+    FlowGraph.closed() { implicit b =>
+      import FlowGraph.Implicits._
+
+      val broadcast = b.add(Broadcast[Unit](numOfJunctions))
+      var outlet = broadcast.out(0)
+      for (i <- 1 until numOfJunctions) {
+        val merge = b.add(Merge[Unit](2))
+        outlet ~> merge
+        broadcast.out(i) ~> merge
+        outlet = merge.out
+      }
+
+      Source.single(()) ~> broadcast
+      outlet ~> Sink.ignore
+    }
+
+  val graphWithNestedImportsBuilder = (numOfNestedGraphs: Int) => {
+    var flow: Graph[FlowShape[Unit, Unit], Unit] = Flow[Unit].map(identity)
+    for (_ <- 1 to numOfNestedGraphs) {
+      flow = FlowGraph.partial(flow) { b ⇒
+        flow ⇒
+          FlowShape(flow.inlet, flow.outlet)
+      }
+    }
+
+    FlowGraph.closed(flow) { implicit b ⇒
+      flow ⇒
+        import FlowGraph.Implicits._
+        Source.single(()) ~> flow ~> Sink.ignore
+    }
+  }
+
+  val graphWithImportedFlowBuilder = (numOfFlows: Int) => {
+    val flow = Flow[Unit].map(identity)
+    FlowGraph.closed() { b ⇒
+      val source = b.add(Source.single(()))
+      var outlet = source
+      for (i <- 0 until numOfFlows) {
+        val flowShape = b.add(flow)
+        b.addEdge(outlet, flowShape.inlet)
+        outlet = flowShape.outlet
+      }
+
+      val sink = b.add(Sink.ignore)
+      b.addEdge(outlet, sink)
+    }
+  }
+}
+
+@State(Scope.Benchmark)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Array(Mode.SampleTime))
+class MaterializationBenchmark {
+  import MaterializationBenchmark._
+
+  implicit val system = ActorSystem("MaterializationBenchmark")
+  implicit val mat = ActorFlowMaterializer()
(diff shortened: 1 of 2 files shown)


akka-bench-jmh/src/main/scala/akka/stream/MaterializationBenchmark.scala
```diff
@@ -0,0 +1,117 @@
...
+import java.util.concurrent.TimeUnit
+import akka.actor.ActorSystem
+import akka.stream.scaladsl._
+import org.openjdk.jmh.annotations._
+
+object MaterializationBenchmark {
+
+  val complexGraphBuilder = () =>
+    FlowGraph.closed() { implicit b =>
+      val numOfJunctions = 100
```
ktoso: Would be cool if `numOfJunctions` was a `@Param("10", "100", "1000")` to see how it grows

ktoso: // but then it can't be val and would have to be in the class.

drewhk:
Also, it would be nice to count the actual number of stages that has been materialized. Otherwise it is hard to exactly compare the results.

akka-bench-jmh/src/main/scala/akka/stream/MaterializationBenchmark.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  implicit val system = ActorSystem("MaterializationBenchmark")
+  implicit val mat = ActorFlowMaterializer()
+
+  val simpleFlow = Source.single(()).map(identity).to(Sink.ignore)
+
+  val complexFlow = {
+    var source = Source.single(())
+    for (_ <- 1 to 100) {
+      source = source.map(identity)
```
drewhk: we should also measure the actual graph wiring time, too (i.e. everything before calling run())

drewhk: and we probably don't need for that a for-loop since JMH takes care of repetitions.

akka-bench-jmh/src/main/scala/akka/stream/MaterializationBenchmark.scala
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  @Benchmark
+  @OperationsPerInvocation(100000)
+  def flow_with_100_operations() {
+    complexFlow.run()
+  }
+
+  @Benchmark
+  @OperationsPerInvocation(100000)
```
drewhk: is it? I thought that this invokes the materializer once

akka-bench-jmh/src/main/scala/akka/stream/MaterializationBenchmark.scala
```diff
@@ -0,0 +1,117 @@
...
+      }
+
+      Source.single(()) ~> broadcast
+      outlet ~> Sink.ignore
+    }
+
+  val graphWithNestedImportsBuilder = () => {
+    var flow: Graph[FlowShape[Unit, Unit], Unit] = Flow[Unit].map(identity)
+    for (_ <- 1 to 100) {
+      flow = FlowGraph.partial(flow) { implicit b ⇒
```
drewhk: what does this intend to do? Btw, I would prefer calling b.add() instead. 
What I would like to see is a graph that is imported 1000 times to another graph. Then we can compare that with creating 1000 graphs directly instead of importing.

drewhk:
Let's discuss tomorrow the details on the benchmarks, I think we should tweak a few things.

akka-bench-jmh/src/main/scala/akka/stream/MaterializationBenchmark.scala
2m: The loop wires 100 maps in sequence. And this code is not part of the benchmark.

akka-bench-jmh/src/main/scala/akka/stream/MaterializationBenchmark.scala
2m: Yup. You are right. I though this tells JMH how much operations to do per invocation, but it actually hints JMH of how many operations will be performed.

akka-bench-jmh/src/main/scala/akka/stream/MaterializationBenchmark.scala
2m: This is nesting a graph in a graph 100 times.

akka-bench-jmh/src/main/scala/akka/stream/MaterializationBenchmark.scala
2m: Great! Added parameter. Also removed simple benchmarks since these are now covered with a parameter value `1`.

akka-bench-jmh/src/main/scala/akka/stream/MaterializationBenchmark.scala
2m: > we should also measure the actual graph wiring time

This is measured in `GraphBuilderBenchmark`

akka-bench-jmh/src/main/scala/akka/stream/MaterializationBenchmark.scala
2m: > What I would like to see is a graph that is imported 1000 times to another graph

This is now benched with `graphWithImportedFlowBuilder` [here](https://github.com/2m/akka/commit/c2af6ea75b1732c965c10dc03f5e2fa62ba33447#diff-f6a9f63376a34a457e1943c5f461285cR55). However I do not understand what do you want to contrast this one with.

drewhk:
LGTM. Do you want to add more tests or can I merge away?

2m:
Please merge. These include flow imports and graph construction benchmarks as well.

Issue #13226, opened 2013-04-11
We should prune the vector clocks when a node leaves the cluster.


bantonsson:
imported from https://www.assembla.com/spaces/akka/tickets/3226

Issue #16890, opened 2015-02-17
FAILED: LeaderDowningNodeThatIsUnreachableSpec


ktoso:
`LeaderDowningNodeThatIsUnreachableSpec` seems to have been fleaky. 
in 2060 it failed, in 2061 it was OK again.

https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2060/consoleFull

```
[JVM-3] - must be able to DOWN a 'last' node that is UNREACHABLE *** FAILED *** (6 seconds, 531 milliseconds)
[JVM-3]   java.util.concurrent.TimeoutException: Futures timed out after [5000 milliseconds]
[JVM-3]   at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:219)
[JVM-3]   at scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:223)
[JVM-3]   at scala.concurrent.Await$$anonfun$result$1.apply(package.scala:116)
[JVM-3]   at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:53)
[JVM-3]   at scala.concurrent.Await$.result(package.scala:116)
[JVM-3]   at akka.remote.testkit.MultiNodeSpec$AwaitHelper.await(MultiNodeSpec.scala:263)
[JVM-3]   at akka.remote.testkit.MultiNodeSpec.node(MultiNodeSpec.scala:364)
[JVM-3]   at akka.cluster.MultiNodeClusterSpec$class.address(MultiNodeClusterSpec.scala:156)
[JVM-3]   at akka.cluster.LeaderDowningNodeThatIsUnreachableSpec.address(LeaderDowningNodeThatIsUnreachableSpec.scala:37)
[JVM-3]   at akka.cluster.LeaderDowningNodeThatIsUnreachableSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply$mcV$sp(LeaderDowningNodeThatIsUnreachableSpec.scala:52)
[JVM-3]   at akka.cluster.LeaderDowningNodeThatIsUnreachableSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply(LeaderDowningNodeThatIsUnreachableSpec.scala:49)
[JVM-3]   at akka.cluster.LeaderDowningNodeThatIsUnreachableSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply(LeaderDowningNodeThatIsUnreachableSpec.scala:49)
[...]

PR #17624, opened 2015-06-01
=clu #13226 Prune vector clocks from removed member

patriknw:
Refs #13226

### DIFF
--- a/akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
+++ b/akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
@@ -510,6 +510,7 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
     else {
       logInfo("Welcome from [{}]", from.address)
       latestGossip = gossip seen selfUniqueAddress
+      assertLatestGossip()
       publish(latestGossip)
       if (from != selfUniqueAddress)
         gossipTo(from, sender())
@@ -657,10 +658,30 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
           (remoteGossip, !remoteGossip.seenByNode(selfUniqueAddress), Newer)
         case _ ⇒
           // conflicting versions, merge
-          (remoteGossip merge localGossip, true, Merge)
+          // We can see that a removal was done when it is not in one of the gossips has status
+          // Down or Exiting in the other gossip.
+          // Perform the same pruning (clear of VectorClock) as the leader did when removing a member.
+          // Removal of member itself is handled in merge (pickHighestPriority)
+          val prunedLocalGossip = localGossip.members.foldLeft(localGossip) { (g, m) ⇒
+            if (Gossip.removeUnreachableWithMemberStatus(m.status) && !remoteGossip.members.contains(m)) {
+              log.debug("Cluster Node [{}] - Pruned conflicting local gossip: {}", selfAddress, m)
+              g.prune(VectorClock.Node(vclockName(m.uniqueAddress)))
+            } else
+              g
+          }
+          val prunedRemoteGossip = remoteGossip.members.foldLeft(remoteGossip) { (g, m) ⇒
+            if (Gossip.removeUnreachableWithMemberStatus(m.status) && !localGossip.members.contains(m)) {
+              log.debug("Cluster Node [{}] - Pruned conflicting remote gossip: {}", selfAddress, m)
+              g.prune(VectorClock.Node(vclockName(m.uniqueAddress)))
+            } else
+              g
+          }
+
+          (prunedRemoteGossip merge prunedLocalGossip, true, Merge)
       }
 
       latestGossip = winningGossip seen selfUniqueAddress
+      assertLatestGossip()
 
       // for all new joining nodes we remove them from the failure detector
       latestGossip.members foreach {
@@ -886,7 +907,14 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
       // removing REMOVED nodes from the `reachability` table
       val newReachability = localOverview.reachability.remove(removed)
       val newOverview = localOverview copy (seen = newSeen, reachability = newReachability)
-      val newGossip = localGossip copy (members = newMembers, overview = newOverview)
+      // Clear the VectorClock when member is removed. The change made by the leader is stamped
+      // and will propagate as is if there are no other changes on other nodes.
+      // If other concurrent changes on other nodes (e.g. join) the pruning is also
+      // taken care of when receiving gossips.
+      val newVersion = removed.foldLeft(localGossip.version) { (v, node) ⇒
+        v.prune(VectorClock.Node(vclockName(node)))
+      }
+      val newGossip = localGossip copy (members = newMembers, overview = newOverview, version = newVersion)
 
       updateLatestGossip(newGossip)
 
@@ -1016,7 +1044,12 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
     val seenVersionedGossip = versionedGossip onlySeen (selfUniqueAddress)
     // Update the state with the new gossip
     latestGossip = seenVersionedGossip
+    assertLatestGossip()
   }
+
+  def assertLatestGossip(): Unit =
+    if (Cluster.isAssertInvariantsEnabled && latestGossip.version.versions.size > latestGossip.members.size)
+      throw new IllegalStateException(s"Too many vector clock entries in gossip state ${latestGossip}")
 
   def publish(newGossip: Gossip): Unit = {
     publisher ! PublishChanges(newGossip)
--- a/akka-cluster/src/main/scala/akka/cluster/VectorClock.scala
+++ b/akka-cluster/src/main/scala/akka/cluster/VectorClock.scala
@@ -187,5 +187,11 @@ final case class VectorClock(
     VectorClock(mergedVersions)
   }
 
+  def prune(removedNode: Node): VectorClock =
+    if (versions.contains(removedNode))
+      copy(versions = versions - removedNode)
+    else
+      this
+
   override def toString = versions.map { case ((n, t)) ⇒ n + " -> " + t }.mkString("VectorClock(", ", ", ")")
 }
--- a/akka-cluster/src/multi-jvm/scala/akka/cluster/NodeChurnSpec.scala
+++ b/akka-cluster/src/multi-jvm/scala/akka/cluster/NodeChurnSpec.scala
@@ -1,0 +1,122 @@
+/**
+ * Copyright (C) 2009-2014 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.cluster
+
+import scala.collection.immutable
+import scala.language.postfixOps
+import scala.concurrent.duration._
+import akka.actor.Address
+import akka.cluster.MemberStatus._
+import akka.remote.testkit.MultiNodeConfig
+import akka.remote.testkit.MultiNodeSpec
+import akka.testkit._
+import com.typesafe.config.ConfigFactory
+import org.scalatest.BeforeAndAfter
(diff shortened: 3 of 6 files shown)


akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -371,7 +371,14 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
     case other             ⇒ super.unhandled(other)
   }
 
-  def initJoin(): Unit = sender() ! InitJoinAck(selfAddress)
+  def initJoin(): Unit = {
```
drewhk: now I know what this change is for :)

patriknw: yeah, this might have been the original reason for the idea, but it makes sense to me in all cases

patriknw:
The pr-validation failure was LeaderDowningNodeThatIsUnreachableSpec which is currently failing a lot in master also, see #16890

drewhk:
I will take a second look at this next Monday

akka-cluster/src/multi-jvm/scala/akka/cluster/NodeChurnSpec.scala
```diff
@@ -0,0 +1,122 @@
...
+          println(s"# leaving ${Cluster(node).selfAddress} -> ${vnode}") // FIXME
+          Cluster(node).leave(Cluster(node).selfAddress)
+          //          Cluster(node).down(Cluster(node).selfAddress)
+        }
+        awaitRemoved(systems)
+        enterBarrier("members-removed-" + n)
+        systems.foreach(_.terminate().await)
+        log.info("end of round-" + n)
+        // log listener will send to testActor if payload size exceed configured log-frame-size-exceeding
+        expectNoMsg(10.seconds)
```
rkuhn: perhaps just do the blocking version after the last iteration?

rkuhn:
LGTM

patriknw:
@drewhk I would be very interested in your thoughts around this one, next week.

patriknw:
rebased, let's revisit in August

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -685,6 +702,13 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
         }
       }
 
+      // FIXME remove
```
viktorklang: ?

patriknw:
@drewhk I would like to merge this PR this week and would like to see your thoughts. Would it be possible for you to review it soonish?

drewhk:
yep, sorry for missing the ping

akka-cluster/src/main/scala/akka/cluster/VectorClock.scala
```diff
@@ -187,5 +187,11 @@ final case class VectorClock(
     VectorClock(mergedVersions)
   }
 
+  def prune(removedNode: Node): VectorClock =
```
drewhk: This removes a "column" for a given node? (i.e. the version vector gets shorter)

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -657,10 +658,26 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
           (remoteGossip, !remoteGossip.seenByNode(selfUniqueAddress), Newer)
         case _ ⇒
           // conflicting versions, merge
-          (remoteGossip merge localGossip, true, Merge)
+          val prunedLocalGossip = localGossip.members.foldLeft(localGossip) { (g, m) ⇒
```
drewhk: My current understanding is that these prunings (on local and remote) will remove any entry m where
 - the entry is removeUnreachableWithMemberStatus (i.e. is of a status that is removable?)
 - and it is not contained by the other gossip

Now, trying to figure out what happens in various scenarios. Suppose all entries I talk about have removeUnreachableWithMemberStatus true (maybe this is wrong to assume, correct me if wrong)

*Scenario 1*: M is contained in both gossips. In this case it will stay in both pruned versions, and as a result in the merged version

*Scenario 2*: M is contained in only one gossip (lets name it G1, and G2 is the other gossip).
 - G1_pruned will not contain M since G2 does not contain it
 - G2_pruned will contain M since G1 does contain it
 - G1_G2_merged will contain M? How does this work here?

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -886,7 +910,13 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
       // removing REMOVED nodes from the `reachability` table
       val newReachability = localOverview.reachability.remove(removed)
       val newOverview = localOverview copy (seen = newSeen, reachability = newReachability)
-      val newGossip = localGossip copy (members = newMembers, overview = newOverview)
+      val newVersion = removed.foldLeft(localGossip.version) { (v, node) ⇒
+        v.prune(VectorClock.Node(vclockName(node)))
+      }
+      val newGossip = localGossip copy (members = newMembers, overview = newOverview, version = newVersion)
```
drewhk: I see. So it is the leader that removes the entry, and this removal is then propagated by the others. This is a safe solution since the leader is always unique (but might not exist if there is a nonconvergent period).

So the only non-monotonic decision is the first removal, which is concentrated to the leader, the rest of the removal is handled by the gossip merge in a monotonic, eventually consistent fashion.

drewhk:
The only point I do not yet get is how the merge handles when one pruned gossip containing M is merged with another pruned gossip that does not contain M. Is it handled properly by their merge function? How does merge work when a node is added, not removed? (i.e how can it distinguish that the difference between the entry-set of the merged two gossips are due to removal or addition?)

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -685,6 +702,13 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
         }
       }
 
+      // FIXME remove
+      if (latestGossip.version.versions.size > latestGossip.members.size) {
```
drewhk: isn't this equal to assertLatestGossip() with additional information printed?

drewhk: ah no, this does not throw

patriknw: I only used this for sanity check when running tests, it should be removed as the FIXME says

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
patriknw: The idea (if I remember correctly) is that the change made by the leader is stamped by the leader node (incrementing its slot in the vclock). That will propagate as is if there are no other changes on other nodes.

If there are changes on other nodes, the merge logic kicks in. (I'll have to revive my memories of that part)

If there are changes from the removed node itself we might have a minor problem. That change might be lost. On the other hand we don't allow any changes from a leaving node (e.g. joining is avoided).

akka-cluster/src/main/scala/akka/cluster/VectorClock.scala
patriknw: depends on you mental model of the VectorClock. I normally see it as a table:

    node    version
    node-1: 3
    node-2: 1
    node-3: 6

`prune(node-3)` would result in:

    node    version
    node-1: 3
    node-2: 1

i.e. removing a row in the table

drewhk: yes, the same, I just imagined it horizontally :)

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
drewhk: Ok, that sounds correct to me.

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
drewhk: Ah, my fault, G2_pruned will not contain M anyway, since G2 did not contain it in the first place. This looks correct then.

drewhk:
After resolving my problem myself, LGTM :)

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
patriknw: > Scenario 1: M is contained in both gossips. In this case it will stay in both pruned versions, and as a result in the merged version

Right, then there has not been any removal (by the leader) and no pruning is involved.

> Scenario 2: M is contained in only one gossip (lets name it G1, and G2 is the other gossip)

G1_pruned will still contain the member M. Pruning only clears entry in vclock. Removal of M itself is handled by `merge` (`pickHighestPriority`).

The idea is (if I remember correctly) that this merge logic for the conflicting case will do the same pruning as the leader did. We see that it was removed because it is not in one of the gossips, and we know that it existed before and is on the way out since the other has status Down or Exiting.

patriknw:
Thanks for review, I might actually put my own LGTM on this, because I had to review it myself to remember what I did on Feb 11 :-)

I'll do the minor cleanups of the FIXMEs and turn some of these review comments into code comments.

patriknw:
Minor update of comments and such. Will merge after validation.

akka-cluster/src/main/scala/akka/cluster/VectorClock.scala
ktoso: very nice :)

akka-cluster/src/multi-jvm/scala/akka/cluster/NodeChurnSpec.scala
```diff
@@ -0,0 +1,122 @@
...
+      cluster.joinSeedNodes(seedNodes)
+      awaitMembersUp(roles.size)
+    }
+
+    "join and remove transient nodes without growing gossip payload" taggedAs LongRunningTest in {
+      // This test is configured with log-frame-size-exceeding and the LogListener
+      // will send to the testActor if unexpected increase in message payload size.
+      // It will fail after a while if vector clock entries of removed nodes are not pruned.
+      for (n ← 1 to rounds) {
+        log.info("round-" + n)
```
viktorklang: Log?

patriknw: makes it easier to understand logs if this test fails

viktorklang: Are the logs withheld in case it succeeds?

patriknw: we have never been able to implement that unfortunately

viktorklang: ouch, that would be a nifty thing to have

patriknw: it would, I did an [attempt](https://github.com/akka/akka/pull/1735) a long time ago, but we disagreed on the level of ambition and never got time to work on it again

viktorklang: 2013 was a good year!
Would we be able to "simply" switch the LoggerActor?

patriknw: It's not that simple, as you can see in https://github.com/akka/akka/pull/1735
I don't remember all complications.
E.g. we also have tests that intercept logs and verify them.

PR #17627, opened 2015-06-01
=htc #17403 fix connection pool reusing existing connections too early

### LINKED ISSUE #17403: HTTP client: If a client application does not read a response body, then future requests will stall
(Closely related to #16898 -- this is basically the client-side mirror of that issue.)

I have some HTTP client code which makes repeated requests to the same host.
After upgrading this code to use the pooled-connection HTTP client in 1.0-RC1, I have found that some requests never complete, even though the remote host has replied.

I believe that the issue is that my client code does not read the full response body in all cases. For example, where it can determine that the response is an error from the status code or the Content-Type it does not read the full response body.

I assume that the HTTP client pool does not notice that the TCP connection "down" stream is stalled in this way, and puts it back in the pool for reuse. The next request to the same host is sent on the "up" stream of that connection, but the next response is never read, as the "down" stream is stalled.

The following test demonstrates the issue:

```
import akka.actor.ActorSystem
import akka.http.scaladsl.model.{HttpResponse, StatusCodes, HttpRequest}
import scala.concurrent.{Future, Await}

class StalledConnectionSpec
  extends Specification
  with NoTimeConversions {

  implicit val fm = ActorFlowMaterializer()

  "HTTP clients who do not read the response body" should {

    "not stall future requests to the same host" in {

      val request = HttpRequest(uri = "http://en.wikipedia.org/wiki/Main_Page")
      val httpClient: HttpRequest => Future[HttpResponse] = Http().singleRequest(_)

      // Make a request, don't read the body:
      val response1 = Await.result(httpClient(request), 10.seconds)
      response1.status mustEqual StatusCodes.OK

      // Make a second request to the same host:
      // The same TCP connection will be re-used by the pool, but it
      // is now blocked waiting for the body of `response1` to be read.
      // The request will time out waiting for that...
      val response2 = Await.result(httpClient(request), 10.seconds)
      response2.status mustEqual StatusCodes.OK
    }
  }
}
```

I think that the TCP connection should not be released to the pool for reuse until the full response has been read by the client code.

/cc @sirthias

sirthias:
Fixes #17403

/cc @jrudolph

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
@@ -10,16 +10,17 @@ import scala.collection.immutable
 import akka.event.LoggingAdapter
 import akka.stream.scaladsl._
 import akka.stream._
+import akka.http.scaladsl.util.FastFuture
 import akka.http.scaladsl.model.HttpMethod
 import akka.http.impl.util._
 
 private object PoolConductor {
   import PoolFlow.RequestContext
-  import PoolSlot.{ SlotEvent, SimpleSlotEvent }
+  import PoolSlot.{ RawSlotEvent, SlotEvent }
 
   case class Ports(
     requestIn: Inlet[RequestContext],
-    slotEventIn: Inlet[SlotEvent],
+    slotEventIn: Inlet[RawSlotEvent],
     slotOuts: immutable.Seq[Outlet[RequestContext]]) extends Shape {
 
     override val inlets = requestIn :: slotEventIn :: Nil
@@ -34,7 +35,7 @@ private object PoolConductor {
     override def copyFromPorts(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]): Shape =
       Ports(
         inlets.head.asInstanceOf[Inlet[RequestContext]],
-        inlets.last.asInstanceOf[Inlet[SlotEvent]],
+        inlets.last.asInstanceOf[Inlet[RawSlotEvent]],
         outlets.asInstanceOf[immutable.Seq[Outlet[RequestContext]]])
   }
 
@@ -47,11 +48,15 @@ private object PoolConductor {
     +--------->|   Merge   +---->| Selector  +-------------->| (MapConcat) +---->|  (Flexi   +-------------->
                |           |     |           |               |             |     |   Route)  +-------------->
                +----+------+     +-----+-----+               +-------------+     +-----------+       to slots     
-                    ^                  ^                                                               
-                    |                  | SimpleSlotEvent                                              
+                    ^                  ^ 
+                    |                  | SlotEvent
+                    |             +----+----+
+                    |             | flatten | mapAsync
+                    |             +----+----+
+                    |                  | RawSlotEvent
                     | Request-         |                                                               
                     | Context     +---------+
-                    +-------------+  retry  |<-------- Slot Event (from slotEventMerge)
+                    +-------------+  retry  |<-------- RawSlotEvent (from slotEventMerge)
                                   |  Split  |
                                   +---------+
 
@@ -67,10 +72,15 @@ private object PoolConductor {
       val doubler = Flow[SwitchCommand].mapConcat(x ⇒ x :: x :: Nil) // work-around for https://github.com/akka/akka/issues/17004
       val route = b.add(new Route(slotCount))
       val retrySplit = b.add(new RetrySplit())
+      val flatten = Flow[RawSlotEvent].mapAsyncUnordered(slotCount) {
+        case x: SlotEvent.Disconnected                ⇒ FastFuture.successful(x)
+        case SlotEvent.RequestCompletedFuture(future) ⇒ future
+        case x                                        ⇒ throw new IllegalStateException("Unexpected " + x)
+      }
 
       retryMerge.out ~> slotSelector.in0
       slotSelector.out ~> doubler ~> route.in
-      retrySplit.out0 ~> slotSelector.in1
+      retrySplit.out0 ~> flatten ~> slotSelector.in1
       retrySplit.out1 ~> retryMerge.in1
 
       Ports(retryMerge.in0, retrySplit.in, route.outlets.asInstanceOf[immutable.Seq[Outlet[RequestContext]]])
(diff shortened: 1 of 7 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
```diff
@@ -67,10 +72,15 @@ private object PoolConductor {
       val doubler = Flow[SwitchCommand].mapConcat(x ⇒ x :: x :: Nil) // work-around for https://github.com/akka/akka/issues/17004
       val route = b.add(new Route(slotCount))
       val retrySplit = b.add(new RetrySplit())
+      val flatten = Flow[RawSlotEvent].mapAsyncUnordered(slotCount) {
```
drewhk: Is the use of unordered correct here?

drewhk: or there can be at most one Event in circulation per slot?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
```diff
@@ -22,15 +24,16 @@ private object PoolSlot {
...
   final case class ResponseDelivery(response: ResponseContext) extends ProcessorOut
-  sealed trait SlotEvent extends ProcessorOut
-  sealed trait SimpleSlotEvent extends SlotEvent
+  sealed trait RawSlotEvent extends ProcessorOut
+  sealed trait SlotEvent extends RawSlotEvent
   object SlotEvent {
-    final case class RequestCompleted(slotIx: Int) extends SimpleSlotEvent
-    final case class Disconnected(slotIx: Int, failedRequests: Int) extends SimpleSlotEvent
-    final case class RetryRequest(rc: RequestContext) extends SlotEvent
+    final case class WhenRequestCompleted(future: Future[RequestCompleted]) extends RawSlotEvent
```
drewhk: the name is a bit misleading. You can just call it RequestCompletedFuture

drewhk:
LGTM, if mapAsyncUnordered is correct here.

drewhk:
@sirthias can I merge this? Is mapAsyncUnordered OK?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
```diff
@@ -149,8 +152,21 @@ private object PoolSlot {
...
+          case x: HttpEntity.Strict ⇒ x -> FastFuture.successful(())
+          case x: HttpEntity.Default ⇒
+            val (newData, whenCompleted) = StreamUtils.captureTermination(x.data)
+            x.copy(data = newData) -> whenCompleted
+          case x: HttpEntity.CloseDelimited ⇒
+            val (newData, whenCompleted) = StreamUtils.captureTermination(x.data)
+            x.copy(data = newData) -> whenCompleted
+          case x: HttpEntity.Chunked ⇒
+            val (newChunks, whenCompleted) = StreamUtils.captureTermination(x.chunks)
+            x.copy(chunks = newChunks) -> whenCompleted
```
rkuhn: This does not handle all entity types: what about IndefiniteLength?

jrudolph: `IndefiniteLength` is not a `ResponseEntity`.

sirthias: `HttpEntity.IndefiniteLength` is not a `ResponseEntity`. Therefore we won't see it here.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
sirthias: Unordered is fine. We don't want a slow response on one connection block us from seeing an update from another connection.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
sirthias: Ok, will do.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
rkuhn: I am massively confused by the type hierarchy here, it seems, but I trust you.

rkuhn:
LGTM (modulo the point Endre raised about the mapAsyncUnordered—I have a hunch that it is fine but there may be dirty details there)

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
sirthias: It's all really simple... :)
4 super types (`RequestEntity`, `ResponseEntity`, `BodyPartEntity` and `UniversalEntity`) "select" sub sets of the 4 actual implementations: `Strict`, `Default`, `CloseDelimited` and `IndefiniteLength`.
Maybe we should provide a handy matrix in the docs that shows with super types comprises which implementation.

sirthias: Actually its five implementations, `Chunked` was missing from my list above.
We definitely need this matrix.... :)

sirthias:
Ok, I renamed `WhenRequestCompleted` as suggested by @drewhk and rebased onto the current release-2.3-dev.
AFAICS there really is no issue with the `mapAsyncUnordered`. The responses from a single connection are naturally ordered and there is no connection between connections (as funny as that may sound). And even if a future for one connection should somehow overtake a previous future for the same connection then this wouldn't be a problem.

So, relax and smell the roses. :)

drewhk:
Wonderful :) Will merge after validation.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
rkuhn: My problem when reading the code on github was that it is not at all obvious how Chunked extends any of RequestEntity or ResponseEntity, but I might just have missed something.

rkuhn:
Good point: the roses do indeed smell wonderful out in the garden :-)

Issue #17617, opened 2015-05-29
Provide metadata extractor / tagger infrastructure for Persistence


ktoso:
- Allows to extract metadata from a payload.
- must be generic so that we can `CassandraTagger extends Tagger[CassandraSpecificMetaData]`
- bindings must be per-journal scoped as different journals need different taggers

ktoso:
After multiple back and forths the design seems to steer more in the direction of adapters: https://github.com/akka/akka/issues/17579

Note for readers: 
* difference being the Adapter is between the persistent actor and the journal and wraps/changes messages
* the tagger would be a separate lifeform which can be asked to tag things.
* we decided the adapter is powerful and useful enough to expose fully.

ktoso:
Further work on this is in: #17579

ktoso:
This is solved by providing an EventAdapter, see https://github.com/akka/akka/issues/17632

Issue #17579, opened 2015-05-27
Persistence "model adapter" for separating models and upcasting


ktoso:
Add a "man in the middle" between PersistentActor and the Journal which can be used to.

No API is yet decided, first we'll need to draft a few example impls.

Use cases this adapter should cover (possibly all, or we split it up, not sure yet):

```
    "domain model to data model" case
                                       +---------+
+---------+     domain => BSON         | journal |
|persist()+---------------------------->         |
+---------+                            |         |
                                       |         |
           <---------------------------+         |
              BSON => domain           |         |
                                       |         |
                                       +---------+

```

- give users more control over "domain model to db **model mapping**", see also https://github.com/akka/akka/issues/17434
- this allows for `Domain => BSON` https://github.com/ironfish/akka-persistence-mongo/issues/88
- allows for "pass through" serialization as in "if my Journal understands BSON, I can give it BSON!"

```
    "upcasting events" case

+                                  +---------+
<----------------------------------+ journal |
|     model-v1 => [0..n] model-v2  |         |
|                                  |         |
|                                  +---------+
+                                      

```
- provide a place where **event upcasting on recovery** could be performed
  - figure out if/how/which we support: `bytes => logic => upcasted event` and/or `old-model => upcasted-model`
  - some inspiration to be gained from [http://www.axonframework.org/docs/2.4/single.html#event-upcasting](Axon upcasters)
  - upcasting may require 1 "v1 event" to emit more than 1 "v2 event", example is when migrating from `ClientDetailsChanged` to `AddressChanged` + `PersonalDataChanged`

- if not needed, it should be an **identity pass through by default**
  - if needed by the Journal, the user must configure it in code - as it's a function and could be combined based on some logic with more upcasters (e.g. when I have "domain-module-a" and "domain-module-b" and both provide adapters for their classes)
  - if needed, and not configured the journal should throw exceptions eagerly

```
                domain-event                          
                    =>                 +---------+
+---------+     tagged(domain-event)   | journal |
|persist()+---------------------------->         |
+---------+                            |         |
                                       |         |
                                       |         |
                                       |         |
                                       |         |
                                       +---------+

```

- additional: since **tagging** is also journal specific, some journals need it and some not, this seems to be a good place to encourage to use for tagging as well, as it's bound to the journal, and gets the event so it can extract tags

olger:
Could explain a bit what in 'identity pass through by default' the identity is ?

olger:
[...]

Issue #17616, opened 2015-05-29
Allow 1:n event upcasting in serializers for persistence


ktoso:
Create a special type in Persistence that indicates that the result of deserialization is in fact multiple events. (`EventsSeq` maybe).

Thanks to this we can support upcasting that reads one old event, and emitts multiple smaller new events.

ktoso:
`EventAdapter` allows this see https://github.com/akka/akka/issues/17632

PR #17632, opened 2015-06-01
+str #17579 #17617 EventAdapter

ktoso:
Initial PR for <strike>ModelAdapter</strike> EventAdapter.

Noteworthy:
- optionally takes ActorSystem as the adapter may need the config or serializers etc
- one adapter per journal
- instead of identity adapter currently avoiding to call it if not set, did not benchmark if it matters or not

refs #17579 #17617

### DIFF
--- a/akka-docs/rst/java/persistence.rst
+++ b/akka-docs/rst/java/persistence.rst
@@ -256,7 +256,7 @@ The ordering between events is still guaranteed ("evt-b-1" will be sent after "e
 .. _defer-java:
 
 Deferring actions until preceding persist handlers have executed
------------------------------------------------------------------
+----------------------------------------------------------------
 
 Sometimes when working with ``persistAsync`` you may find that it would be nice to define some actions in terms of
 ''happens-after the previous ``persistAsync`` handlers have been invoked''. ``PersistentActor`` provides an utility method
@@ -502,6 +502,54 @@ is defined by the ``maxUnconfirmedMessages`` method. If this limit is exceed the
 not accept more messages and it will throw ``AtLeastOnceDelivery.MaxUnconfirmedMessagesExceededException``.
 The default value can be configured with the ``akka.persistence.at-least-once-delivery.max-unconfirmed-messages``
 configuration key. The method can be overridden by implementation classes to return non-default values.
+
+.. _event-adapters-java:
+
+Event Adapters
+==============
+
+.. note::
+
+  Complete documentation featuring use-cases and implementation examples for this feature will follow shortly.
+
+In long running projects using event sourcing sometimes the need arrises to detach the data model from the domain model
+completely.
+
+Event Adapters help in situations where:
+
+- **Version Migration** – existing events stored in *Version 1* should be "upcasted" to a new *Version 2* representation,
+  and the process of doing so involves actual code, not just changes on the serialization layer. For these scenarios
+  the ``toJournal`` function is usually an identity function, however the ``fromJournal`` is implemented as
+  ``v1.Event=>v2.Event``, performing the neccessary mapping inside the fromJournal method.
+  This technique is sometimes refered to as "upcasting" in other CQRS libraries.
+- **Separating Domain and Data models** – thanks to EventAdapters it is possible to completely separate the domain model
+  from the model used to persist data in the Journals. For example one may want to use case classes in the
+  domain model, however persist their protocol-buffer (or any other binary serialization format) counter-parts to the Journal.
+  A simple ``toJournal:MyModel=>MyDataModel`` and ``fromJournal:MyDataModel=>MyModel`` adapter can be used to implement this feature.
+- **Journal Specialized Data Types** – exposing data types understood by the underlying Journal, for example for data stores which
+  understand JSON it is possible to write an EventAdapter ``toJournal:Any=>JSON`` such that the Journal can *directly* store the
+  json instead of serializing the object to its binary representation.
+
+Implementing an EventAdapter is rather stright forward:
+
+.. includecode:: code/docs/persistence/PersistenceEventAdapterDocTest.java#identity-event-adapter
+
+Then in order for it to be used on events coming to and from the journal you must bind it using the below configuration syntax:
+
+.. includecode:: ../scala/code/docs/persistence/PersistenceEventAdapterDocSpec.scala#event-adapters-config
+
+It is possible to bind multiple adapters to one class *for recovery*, in which case the ``fromJournal`` methods of all
+bound adapters will be applied to a given matching event (in order of definition in the configuration). Since each adapter may
+return from ``0`` to ``n`` adapted events (called as ``EventSeq``), each adapter can investigate the event and if it should
+indeed adapt it return the adapted event(s) for it, other adapters which do not have anything to contribute during this
+adaptation simply return ``EventSeq.empty``. The adapted events are then delivered in-order to the ``PersistentActor`` during replay.
+
+.. note::
+  More advanced techniques utilising advanced binary serialization formats such as protocol buffers or kryo / thrift / avro
+  will be documented very soon. These schema evolutions often may need to reach into the serialization layer, however
+  are much more powerful in terms of flexibly removing unused/deprecated classes from your classpath etc.
+
+
 
 Storage plugins
 ===============
--- a/akka-docs/rst/scala/code/docs/persistence/PersistenceEventAdapterDocSpec.scala
+++ b/akka-docs/rst/scala/code/docs/persistence/PersistenceEventAdapterDocSpec.scala
@@ -1,0 +1,233 @@
+/*
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package docs.persistence
+
+import akka.actor.{ ExtendedActorSystem, Props }
+import akka.persistence.journal.{ EventAdapter, EventSeq }
+import akka.persistence.{ PersistentActor, RecoveryCompleted }
+import akka.testkit.{ AkkaSpec, TestProbe }
+import com.google.gson.{ Gson, JsonElement }
+
+import scala.collection.immutable
+
+class PersistenceEventAdapterDocSpec(config: String) extends AkkaSpec(config) {
+
+  def this() {
+    this("""
+      akka.persistence.snapshot-store.plugin = "akka.persistence.snapshot-store.local"
+
+      //#event-adapters-config
+      akka.persistence.journal {
+        inmem {
+          event-adapters {
+            tagging        = "docs.persistence.MyTaggingEventAdapter"
+            user-upcasting = "docs.persistence.UserUpcastingEventAdapter"
+            item-upcasting = "docs.persistence.ItemUpcastingEventAdapter"
+          }
+
+          event-adapter-bindings {
+            "docs.persistence.Item"        = tagging
+            "docs.persistence.TaggedEvent" = tagging
+            "docs.persistence.v1.Event"    = [user-upcasting, item-upcasting]
+          }
+        }
+      }
+      //#event-adapters-config
+
+
+      akka.persistence.journal {
+        auto-json-store {
+          class = "akka.persistence.journal.inmem.InmemJournal" # reuse inmem, as an example
+
+          event-adapters {
+            auto-json = "docs.persistence.MyAutoJsonEventAdapter"
+          }
+
+          event-adapter-bindings {
+            "docs.persistence.DomainEvent" = auto-json  # to journal
+            "com.google.gson.JsonElement"  = auto-json  # from journal
+          }
+        }
+
+        manual-json-store {
+          class = "akka.persistence.journal.inmem.InmemJournal" # reuse inmem, as an example
+
+          event-adapters {
+            manual-json    = "docs.persistence.MyManualJsonEventAdapter"
+          }
+
+          event-adapter-bindings {
+            "docs.persistence.DomainEvent" = manual-json  # to journal
+            "com.google.gson.JsonElement"  = manual-json  # from journal
+          }
+        }
+      }
+    """)
+  }
+
+  "MyAutomaticJsonEventAdapter" must {
+    "demonstrate how to implement a JSON adapter" in {
+      val p = TestProbe()
+
+      val props = Props(new PersistentActor {
+        override def persistenceId: String = "json-actor"
+        override def journalPluginId: String = "akka.persistence.journal.auto-json-store"
+
+        override def receiveRecover: Receive = {
+          case RecoveryCompleted => // ignore...
+          case e                 => p.ref ! e
+        }
+
+        override def receiveCommand: Receive = {
+          case c => persist(c) { e => p.ref ! e }
+        }
+      })
+
+      val p1 = system.actorOf(props)
+      val m1 = Person("Caplin", 42)
+      val m2 = Box(13)
+      p1 ! m1
+      p1 ! m2
+      p.expectMsg(m1)
+      p.expectMsg(m2)
+
+      val p2 = system.actorOf(props)
+      p.expectMsg(m1)
+      p.expectMsg(m2)
+    }
+  }
+
+  "MyManualJsonEventAdapter" must {
+    "demonstrate how to implement a JSON adapter" in {
+      val p = TestProbe()
+
+      val props = Props(new PersistentActor {
+        override def persistenceId: String = "json-actor"
+        override def journalPluginId: String = "akka.persistence.journal.manual-json-store"
+
+        override def receiveRecover: Receive = {
+          case RecoveryCompleted => // ignore...
+          case e                 => p.ref ! e
+        }
+
+        override def receiveCommand: Receive = {
+          case c => persist(c) { e => p.ref ! e }
+        }
+      })
+
+      val p1 = system.actorOf(props)
+      val m1 = Person("Caplin", 42)
+      val m2 = Box(13)
+      p1 ! m1
+      p1 ! m2
+      p.expectMsg(m1)
+      p.expectMsg(m2)
+
+      val p2 = system.actorOf(props)
+      p.expectMsg(m1)
+      p.expectMsg(m2)
+    }
+  }
+}
+
+trait DomainEvent
+case class Person(name: String, age: Int) extends DomainEvent
+case class Box(length: Int) extends DomainEvent
+
+case class MyTaggingJournalModel(payload: Any, tags: immutable.Set[String])
+
+//#identity-event-adapter
+class MyEventAdapter(system: ExtendedActorSystem) extends EventAdapter {
+  override def manifest(event: Any): String =
+    "" // when no manifest needed, return ""
+
+  override def toJournal(event: Any): Any =
+    event // identity
+
+  override def fromJournal(event: Any, manifest: String): EventSeq =
+    EventSeq.single(event) // identity
+}
+//#identity-event-adapter
+
+/**
+ * This is an example adapter which completely takes care of domain<->json translation.
+ * It allows the journal to take care of the manifest handling, which is the FQN of the serialized class.
+ */
+class MyAutoJsonEventAdapter(system: ExtendedActorSystem) extends EventAdapter {
+  private val gson = new Gson
+
+  override def manifest(event: Any): String =
+    event.getClass.getCanonicalName
+
+  override def toJournal(event: Any): Any = gson.toJsonTree(event)
+
+  override def fromJournal(event: Any, manifest: String): EventSeq = EventSeq.single {
+    event match {
+      case json: JsonElement =>
+        val clazz = system.dynamicAccess.getClassFor[Any](manifest).get
+        gson.fromJson(json, clazz)
+    }
+  }
+}
+
+class MyUpcastingEventAdapter(system: ExtendedActorSystem) extends EventAdapter {
+  override def manifest(event: Any): String = ""
+
+  override def toJournal(event: Any): Any = ???
+
+  override def fromJournal(event: Any, manifest: String): EventSeq = ???
+}
+
+/**
+ * This is an example adapter which completely takes care of domain<->json translation.
+ * It manually takes care of smuggling through the manifest even if the journal does not do anything in this regard,
+ * which is the case in this example since we're persisting to the inmem journal, which does nothing in terms of manifest.
+ */
+class MyManualJsonEventAdapter(system: ExtendedActorSystem) extends EventAdapter {
+
+  private val gson = new Gson
+
+  override def manifest(event: Any): String = event.getClass.getCanonicalName
+
+  override def toJournal(event: Any): Any = {
+    val out = gson.toJsonTree(event).getAsJsonObject
+
+    // optionally can include manifest in the adapted event.
+    // some journals will store the manifest transparently
+    out.addProperty("_manifest", manifest(event))
+
+    out
+  }
+
+  override def fromJournal(event: Any, m: String): EventSeq = event match {
+    case json: JsonElement =>
+      val manifest = json.getAsJsonObject.get("_manifest").getAsString
+
+      val clazz = system.dynamicAccess.getClassFor[Any](manifest).get
+      EventSeq.single(gson.fromJson(json, clazz))
+  }
+}
+
+class MyTaggingEventAdapter(system: ExtendedActorSystem) extends EventAdapter {
+  override def manifest(event: Any): String = ""
+
+  override def fromJournal(event: Any, manifest: String): EventSeq = event match {
+    case j: MyTaggingJournalModel => EventSeq.single(j)
+  }
+
+  override def toJournal(event: Any): Any = {
+    event match {
+      case Person(_, age) if age >= 18 => MyTaggingJournalModel(event, tags = Set("adult"))
+      case Person(_, age)              => MyTaggingJournalModel(event, tags = Set("minor"))
+      case _                           => MyTaggingJournalModel(event, tags = Set.empty)
+    }
+  }
+}
+
+object v1 {
+  trait Event
+  trait UserEvent extends v1.Event
+  trait ItemEvent extends v1.Event
+}
--- a/akka-persistence/src/main/scala/akka/persistence/JournalProtocol.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/JournalProtocol.scala
@@ -4,9 +4,9 @@
 
 package akka.persistence
 
-import scala.collection.immutable
-
 import akka.actor._
+
+import scala.collection.immutable
 
 /**
  * INTERNAL API.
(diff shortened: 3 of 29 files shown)


akka-persistence/src/main/scala/akka/persistence/JournalProtocol.scala
```diff
@@ -46,6 +48,13 @@ private[persistence] object JournalProtocol {
     extends Request
 
   /**
+   * Register an model adapter for the Journal
+   * @param adapter to be applied on all events going to this journal
+   */
+  final case class RegisterModelAdapter(adapter: ModelAdapter[Any])
```
patriknw: naming: EventAdapter?

akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
```diff
@@ -47,7 +47,9 @@ trait AsyncWriteJournal extends Actor with WriteJournalBase with AsyncRecovery {
       // Send replayed messages and replay result to persistentActor directly. No need
       // to resequence replayed messages relative to written and looped messages.
       asyncReplayMessages(persistenceId, fromSequenceNr, toSequenceNr, max) { p ⇒
-        if (!p.deleted || replayDeleted) persistentActor.tell(ReplayedMessage(p), p.sender)
+        if (!p.deleted || replayDeleted) {
+          persistentActor.tell(ReplayedMessage(adaptFromJournal(p)), p.sender)
```
patriknw: I think we also need to provide a type hint to `adaptFromJournal`, similar the string `manifest` for serializers

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
```diff
@@ -215,6 +253,34 @@ class Persistence(val system: ExtendedActorSystem) extends Extension {
     system.systemActorOf(pluginActorProps, pluginActorName)
   }
 
+  private def createAdapter(configPath: String): Option[ModelAdapter[Any]] = {
+    val pluginConfig = system.settings.config.getConfig(configPath)
+    if (!pluginConfig.hasPath("adapter")) None
+    else {
+      val adapterClassName = pluginConfig.getString("adapter")
```
patriknw: I might have completely misunderstood your "tagging prototype", but I thought you had something similar to the serialization extension that would support one tagger/adapter per configured event type? I thought that would be nice, but this is perhaps enough?

akka-persistence/src/main/scala/akka/persistence/JournalProtocol.scala
ktoso: +1, sounds good

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
ktoso: Oh - no I think you got me right, see here for tagger impl: https://github.com/ktoso/akka/commit/410a4a8738296187076d1b74074a6ee15798c8a4#diff-aa1e901f1dd62bc57d4b969106f64c27R70 It's as serializers "pick most specific one" and allowed for multiple ones to be combined into one (that part we do not need for adapters).

I somehow thought "one adapter per journal", but now that you mention it "serializer like registration" for those makes much sense, so I'll bring back that idea. :)

patriknw: yes, and that's why I wanted to combine the adapter with the tagger, so that user doesn't have to configure (mostly) same thing twice

drewhk:
the comment tag "str" is wrong :)

akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
ktoso: Right, needed for "I do my mapping manually"

akka-persistence/src/test/scala/akka/persistence/EventAdapterSpec.scala
```diff
@@ -0,0 +1,204 @@
...
+         |    unwrap              = "${classOf[EventAdapterSpec].getCanonicalName}$$PayloadUnwrapper"
+         |  }
+         |
+         |  # define bindings for all journals
+         |  event-adapter-bindings {
+         |    "${classOf[EventAdapterSpec].getCanonicalName}$$UserDataChanged"   = age
+         |    "${classOf[EventAdapterSpec].getCanonicalName}$$TaggedDataChanged" = age
+         |    "${classOf[EventAdapterSpec].getCanonicalName}$$MyTagJournal$$JournalModel"      = unwrap
+         |  }
+         |
```
ktoso: @patriknw so that's the style of configuration we can do:

- adapters defined in one place
- bindings defined at top level, or:
- for each plugin separately

ktoso: (this test fails now, I messed up something with when which adapter is used)

patriknw: I think it is easiest to understand with one config section for each plugin, i.e. 

    akka.persistence.journal.leveldb {
      event-adapters { ... }
      event-adapter-bindings { ... }
    }

akka-docs/rst/scala/code/docs/persistence/PersistenceEventAdapterDocSpec.scala
```diff
@@ -0,0 +1,195 @@
...
+          # define adapters available for all plugins for binding here
+        }
+
+        inmem {
+          event-adapters {
+            tagging = "docs.persistence.MyTaggingEventAdapter"
+          }
+
+          event-adapter-bindings {
+            "docs.persistence.MySnapshot" = my-snapshot
```
patriknw: where is `my-snapshot` defined?

akka-docs/rst/scala/code/docs/persistence/PersistenceEventAdapterDocSpec.scala
```diff
@@ -0,0 +1,195 @@
...
+ * This is an example adapter which completely takes care of domain<->json translation.
+ * It manually takes care of smuggling through the manifest even if the journal does not do anything in this regard,
+ * which is the case in this example since we're persisting to the inmem journal, which does nothing in terms of manifest.
+ */
+class MyManualJsonEventAdapter(system: ExtendedActorSystem) extends EventAdapter[JsonElement] {
+
+  val gson = new Gson // TODO: serisouly, simplest to plug in, all type-class based ones fail in this scenario
+
+  override val includeManifest = false // we take care of it manually
+  override def manifest(event: Any): String = event.getClass.getCanonicalName
```
patriknw: why not always include the manifest, i.e. remove the `def includeManifest`? if not needed it can be `""`

akka-persistence/src/main/scala/akka/persistence/Persistent.scala
```diff
@@ -63,6 +64,11 @@ trait PersistentRepr extends PersistentEnvelope with Message {
   def withPayload(payload: Any): PersistentRepr
 
   /**
+   * Creates a new persistent message with the specified `manifest`.
+   */
+  def withManifest(payload: String): PersistentRepr
```
patriknw: s/payload/manifest/

akka-persistence/src/main/scala/akka/persistence/journal/EventAdapter.scala
```diff
@@ -0,0 +1,57 @@
...
+  def fromJournal(event: JournalModel, manifest: String): Any
+
+  /** Returns whether this adapter needs a manifest in the `fromJournal` method. */
+  def includeManifest = false
+
+  /**
+   * Return the manifest (type hint) that will be provided in the `fromJournal` method.
+   * Use `""` if manifest is not needed.
+   */
+  def manifest(event: Any): String = ""
```
patriknw: I think this should be an abstract method, i.e. no default impl

akka-persistence/src/main/scala/akka/persistence/journal/WriteJournalBase.scala
```diff
@@ -4,21 +4,36 @@
...
-    rb.filter(persistentPrepareWrite).asInstanceOf[immutable.Seq[PersistentRepr]] // filter instead of flatMap to avoid Some allocations
+    rb.collect {
+      case p: PersistentRepr ⇒
+        p.prepareWrite()
+        adaptToJournal(p)
+    } // collect instead of flatMap to avoid Some allocations
+
+  /** INTERNAL API */
+  private[akka] final def adaptFromJournal(repr: PersistentRepr): PersistentRepr = {
+    val adapter = eventAdapters.get(repr.payload.getClass)
```
patriknw: Something is missing here. The stored format is typically the same for all events, e.g. String, Json, but there is no way to map that class to different adapters. I think something similar to the serializerId is needed, i.e. the id is stored and on the way back the adapter is looked up by id instead of by class.

akka-persistence/src/test/scala/akka/persistence/EventAdapterSpec.scala
```diff
@@ -0,0 +1,217 @@
...
+  def this(journalName: String, journalConfig: Config) {
+    this(journalName, journalConfig, ConfigFactory.parseString(
+      s"""
+         |akka.persistence.journal {
+         |
+         |  event-adapters {
+         |    age                 = "${classOf[EventAdapterSpec].getCanonicalName}$$UserAgeTaggingAdapter"
+         |    replay-pass-through = "${classOf[EventAdapterSpec].getCanonicalName}$$ReplayPassThroughAdapter"
+         |    unwrap              = "${classOf[EventAdapterSpec].getCanonicalName}$$PayloadUnwrapper"
+         |  }
```
patriknw: I'm not sure we should complicate it with these two possible locations of `event-adapters`.
It is possible to extract common definitions with typesafe config anyway using syntax like:

    inmem.event-adapters = ${common-event-adapters}

project/Dependencies.scala
```diff
@@ -36,6 +36,11 @@ object Dependencies {
     // TODO remove with metrics from akka-cluster
     val sigar         = "org.fusesource"              % "sigar"                        % "1.6.4"       // ApacheV2
 
+    object Docs {
+      val sprayJson   = "io.spray"                   %%  "spray-json"                  % "1.3.2"
+      val gson        = "com.google.code.gson"        % "gson"                         % "2.3.1"
```
patriknw: "test" scope?

patriknw:
This is in the right direction. I don't think we can get away from adding an `adapterId` also.

akka-docs/rst/scala/code/docs/persistence/PersistenceEventAdapterDocSpec.scala
ktoso: Sounds good

akka-persistence/src/main/scala/akka/persistence/Persistent.scala
ktoso: thanks!

akka-persistence/src/main/scala/akka/persistence/journal/EventAdapter.scala
ktoso: ok :+1: and removed the `includeManifest method.

project/Dependencies.scala
ktoso: thanks!

akka-persistence/src/test/scala/akka/persistence/EventAdapterSpec.scala
ktoso: ok and concating lists works as well... simplifying now

akka-docs/rst/scala/code/docs/persistence/PersistenceEventAdapterDocSpec.scala
ktoso: bug, thanks!

akka-persistence/src/main/scala/akka/persistence/journal/WriteJournalBase.scala
ktoso: Right, I see what you mean - working on it.

ktoso: See https://github.com/ktoso/akka/commit/29d82b1875ab9a6639bb96bc7a25015f3d8c2714 for the "adapterId" implementation. Today we figured out during the meeting that even this is not enough, as it does not cover:

- initially no adapter
- now we need to add an adapter for some type
- that type could be "json", so it's hard to restrict to one

patriknw: What is the exact use case for that? Isn't that overkill? How did you create json if you didn't use an adapter in the first place?

ktoso: Enjoy the midsommar! :wink: (over here it's raining) (no releases today, no worries)

One of the possible scenarios is (it also starts to play into the ideas needed for https://github.com/akka/akka/issues/17616):

"old app"
- `persist(OldAppEvent)`
- catch all "toJson" adapter
  - stored id is something, lets say 1001

We start the new version of the app:
- "we need to move away from OldAppEvents!"
  - they are all stored as json, that's what the adapter gets
  - the model has become complex and we want multiple adapters for different types that are "inside the json"
  - the adapter gets json to the fromJournal() - current impl it would for all of these have to be the 1001 adapter
- **new from here-on:** we add EventAdapters with `fromJournal: Json => NewAppEvent`
  - adapter knows based on something inside the json if it should emit events or not (we provide `EventSeq.empty`)
  - can emit "nothing"
  - allow 1:n which we did not yet address

How it could be done with current:
- was stored without adapter so adapterId is 0
- users are kind of stuck with that decision if we only allow picking the adapter by the adapter id, we could allow binding to the 0 adapterId (which in theory should be our "internal reserved id" for identity)...

akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteProxy.scala
```diff
@@ -29,10 +29,13 @@ private[persistence] trait AsyncWriteProxy extends AsyncWriteJournal with Stash
 
   override def receive = {
     case SetStore(ref) ⇒
+      println("SetStore(ref) = " + SetStore(ref) + ", " + context.system)
```
patriknw: change to log.debug?

akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteProxy.scala
```diff
@@ -29,10 +29,13 @@ private[persistence] trait AsyncWriteProxy extends AsyncWriteJournal with Stash
...
       store = ref
       unstashAll()
       context.become(initialized)
-    case _ ⇒ stash()
+    case x ⇒
+      println("x = " + x)
```
patriknw: remove or debug?

ktoso: Oh, actually will remove here - somewhere on this branch I broke `SharedLeveldbJournalSpec`, this is from investigating it

akka-persistence/src/main/scala/akka/persistence/journal/SyncWriteJournal.scala
```diff
@@ -41,11 +41,16 @@ trait SyncWriteJournal extends Actor with WriteJournalBase with AsyncRecovery {
...
+        if (!p.deleted || replayDeleted)
+          adaptFromJournal(p).foreach { adaptedPersistentRepr ⇒
+            persistentActor.tell(ReplayedMessage(adaptedPersistentRepr), adaptedPersistentRepr.sender)
+          }
       } map {
         case _ ⇒ ReplayMessagesSuccess
       } recover {
-        case e ⇒ ReplayMessagesFailure(e)
+        case e ⇒
+          e.printStackTrace()
```
patriknw: remove

patriknw:
@ktoso noticed that you updated, is it ready for review?

ktoso:
Yes I think you can give it a read, I know it breaks `SharedLeveldbJournalSpec` however have not figured out why it does yet (so looking into it now).

akka-docs/rst/scala/code/docs/persistence/PersistenceEventAdapterDocSpec.scala
```diff
@@ -0,0 +1,233 @@
...
+
+  override def toJournal(event: Any): Any = {
+    event match {
+      case Person(_, age) if age >= 18 => MyTaggingJournalModel(event, tags = Set("adult"))
+      case Person(_, age)              => MyTaggingJournalModel(event, tags = Set("minor"))
+      case _                           => MyTaggingJournalModel(event, tags = Set.empty)
+    }
+  }
+}
+
```
ktoso: Docs here are incomplete (impls not included in docs, but that will be part of the documenting schema evolution ticket)

akka-docs/rst/java/persistence.rst
```diff
@@ -503,6 +503,53 @@ not accept more messages and it will throw ``AtLeastOnceDelivery.MaxUnconfirmedM
...
+.. note::
+
+  Complete documentation featuring use-cases and implementation examples for this feature will follow shortly.
+
+In long running projects using event sourcing sometimes the need arrises to detach the data model from the domain model
+completely.
+
+Event Adapters help in situations where:
+
+- **Upcasting** – existing events stored in *Version 1* should be "upcasted" to a new *Version 2* representation,
```
patriknw: I really dislike the upcasting term. I know where it comes from, but it doesn't feel right (and requires explanation). I prefer migration.

ktoso: Ok, how about we call it "model migrations" here? I'd like to add the word upcasting in something like "This technique is sometimes refered to as 'upcasting' in some other libraries"? In case someone familiar with Axon or something like it tries to find "how do I do upcasting in akka?"

akka-docs/rst/java/persistence.rst
```diff
@@ -503,6 +503,53 @@ not accept more messages and it will throw ``AtLeastOnceDelivery.MaxUnconfirmedM
...
+  and the process of doing so involves actual code, not just changes on the serialization layer. For these scenarios
+  the ``toJournal`` function is usually an identity function, however the ``fromJournal`` is implemented as
+  ``v1.Event=>v2.Event``, performing the neccessary upcasting inside the fromJournal method.
+- **Separating Domain and Data models** – thanks to EventAdapters it is possible to completely separate the domain model
+  from the model used to persist data in the Journals. For example one may want to use nice to use case classes in the
```
patriknw: had to read this several times "to use nice to use".
I think this is enough: For example one may want to use case classes...

akka-docs/rst/java/persistence.rst
```diff
@@ -503,6 +503,53 @@ not accept more messages and it will throw ``AtLeastOnceDelivery.MaxUnconfirmedM
...
+  domain model, however persist their protocol-buffer (or any other binary serialization format) counter-parts to the Journal.
+  A simple ``toJournal:MyModel=>MyDataModel`` and ``fromJournal:MyDataModel=>MyModel`` adapter can be used to implement this feature.
+- **Journal Specialized Data Types** – exposing data types understood by the underlying Journal, for example for data stores which understand JSON its possible
```
patriknw: it's

akka-docs/rst/java/persistence.rst
```diff
@@ -503,6 +503,53 @@ not accept more messages and it will throw ``AtLeastOnceDelivery.MaxUnconfirmedM
...
+
+Implementing an EventAdapter is rather stright forward:
+
+.. includecode:: code/docs/persistence/PersistenceEventAdapterDocTest.java#identity-event-adapter
+
+Then in order for it to be used on events coming to and from the journal you must bind it using the below configuration syntax:
+
+.. includecode:: ../scala/code/docs/persistence/PersistenceEventAdapterDocSpec.scala#event-adapters-config
+
+It is possible to bind multiple adapters to one class, in which case the event will be given (in order of definition in the configuration)
```
patriknw: perhaps clarify that this is only for `fromJournal`

akka-persistence/src/main/scala/akka/persistence/Persistent.scala
```diff
@@ -95,9 +101,10 @@ object PersistentRepr {
     payload: Any,
     sequenceNr: Long = 0L,
     persistenceId: String = PersistentRepr.Undefined,
+    manifest: String = PersistentRepr.Undefined,
     deleted: Boolean = false,
     sender: ActorRef = null): PersistentRepr =
-    PersistentImpl(payload, sequenceNr, persistenceId, deleted, sender)
+    PersistentImpl(payload, sequenceNr, persistenceId, manifest, /*adapterId,*/ deleted, sender)
```
patriknw: remove

ktoso: thanks!

akka-persistence/src/test/scala/akka/persistence/EndToEndEventAdapterSpec.scala
```diff
@@ -0,0 +1,241 @@
...
+    system.actorOf(Props(classOf[EndToEndAdapterActor], name, "akka.persistence.journal." + journalName, probe), name)
+
+  def withActorSystem[T](name: String, config: Config)(block: ActorSystem ⇒ T): T = {
+    val system = ActorSystem(name, journalConfig withFallback config)
+    try block(system) finally Await.ready(system.terminate(), 3.seconds)
+  }
+
+  "EventAdapters in end-to-end scenarios" must {
+
+    "use the same adapter when reading as was used when writing to the journal, by its adapterId" in
```
ktoso: fixed test title not to include adapterId now, push soon

akka-persistence/src/main/scala/akka/persistence/journal/EventAdapter.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * Convert domain event to journal event type.
+   *
+   * Some journal may require a specific type to be returned to them,
+   * for example if a primary key has to be associated with each event then a journal
+   * may require adapters to return `com.example.myjournal.EventWithPrimaryKey(event, key)`.
+   *
+   * The `toJournal` adaptation must be an 1-to-1 transformation.
+   * It is now allowed to drop incoming events during the `toJournal` adaptation.
```
patriknw: s/now/not/ ?

akka-persistence/src/main/scala/akka/persistence/journal/EventAdapter.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  /** Java API */
+  final def single(event: Any): EventSeq = new SingleEventSeq(event)
+
+  /** Java API */
+  @varargs final def create(events: Any*): EventSeq = EventsSeq(events.toList)
+  final def apply(events: Any*): EventSeq = EventsSeq(events.toList)
+}
+final case class SingleEventSeq(event: Any) extends EventSeq { // TODO try to make it a value class, would save allocations
+  override def events: immutable.Seq[Any] = List(event)
```
patriknw: val

akka-docs/rst/java/persistence.rst
ktoso: agreed, thanks :)

akka-docs/rst/java/persistence.rst
ktoso: :+1:

akka-persistence/src/main/scala/akka/persistence/journal/EventAdapter.scala
ktoso: :+1: thanks

akka-persistence/src/main/scala/akka/persistence/journal/WriteJournalBase.scala
```diff
@@ -4,16 +4,39 @@
...
 
+  lazy val persistence = Persistence(context.system)
+  private def eventAdapters = persistence.adaptersFor(self)
+
   protected def preparePersistentBatch(rb: immutable.Seq[PersistentEnvelope]): immutable.Seq[PersistentRepr] =
-    rb.collect {
-      case p: PersistentRepr ⇒ p.update(sender = Actor.noSender) // don't store sender
+    rb.collect { // collect instead of flatMap to avoid Some allocations
+      case p: PersistentRepr ⇒
+        p.update(sender = Actor.noSender) // don't store sender
```
patriknw: it is immutable, so you must pass in this result in next stmt

    adaptToJournal(p.update(sender = Actor.noSender))

akka-persistence/src/main/scala/akka/persistence/journal/EventAdapter.scala
ktoso: Thanks. 
I was trying to make this one a value class, however things get complicated then sadly...
(`EventSeq` must be an interface and must `extend Any` (universal trait), which destroys the Java API forwarders etc). I hope to try this again, we could save on some allocations here.

akka-persistence/src/main/scala/akka/persistence/journal/WriteJournalBase.scala
ktoso: whoops, silly rebase mistake; thanks!

patriknw:
LGTM, aside from some minor things

akka-docs/rst/java/persistence.rst
patriknw: "version migrations"?
(mentioning upcasting is fine)

ktoso:
Addressed comments and fixed bug in LevelDBStore which failed validation previously.
I'll merge once Validation passes.

More work around docs for this coming soon anyway, but in next Milestone.

akka-persistence/src/main/scala/akka/persistence/journal/leveldb/LeveldbStore.scala
```diff
@@ -112,11 +111,11 @@ class SharedLeveldbStore extends { val configPath = "akka.persistence.journal.le
   import AsyncWriteTarget._
 
   def receive = {
-    case WriteMessages(msgs)                        ⇒ sender() ! writeMessages(msgs)
+    case WriteMessages(msgs)                        ⇒ sender() ! writeMessages(preparePersistentBatch(msgs))
```
patriknw: great catch

ktoso:
Fails here on JavaDoc which I wasn't able to figure out precisely yet.

ktoso:
Pushed the javadoc fixup, should be OK to merge once validation confirms its happy.

patriknw:
thanks @ktoso, I hope you did that from bed ;-)

ktoso:
Correct, after a round of mid-day sleeping ;-) Happy to catch up a bit (no pressure) here and there.

Issue #16524, opened 2014-12-11
Consider removing some fault handling options in PersistentActor


patriknw:
Right now: PersistenceFailure is emitted to user receive handler, if it is unhandled it throws ActorKilledException, which by default supervision strategy stops the actor but can be intercepted differently by user

I'm not sure what the best solution is, but right now it feels like there are too many alternatives. The default of stopping the actor is good though.

ktoso:
If we invent an replacement I feel it should be able to "continue replaying in presence of corrupted events". In the current schema this is doable by handling `PersistenceFailure` and for example logging `"whoops, 2323 seems to be in a bad state?"`, the actor then continues replay.

patriknw:
Regarding "continue replaying in presence of corrupted events": that is also some kind of migration issue, and I think it can be solved by the `EventSeq` as proposed in #17616, i.e. the exception is handled in the serializer instead of in PersistentActor.

PR #17634, opened 2015-06-01
!per #16524 Make persistent failures fatal

patriknw:
I'm still exploring, but wanted to share current state so that we can discuss (tomorrow).

This is what I have done so far:
- removed RecoveryFailure and PersistentFailure messages, `unhandled` mess removed
- any failure during recovery or persist will unconditionally stop the actor (logging included)
- I dislike the use of ActorKilledException, using stop instead

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/BackoffSupervisor.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/BackoffSupervisor.scala
@@ -1,0 +1,148 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.persistence
+
+import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.forkjoin.ThreadLocalRandom
+import akka.actor.Actor
+import akka.actor.ActorLogging
+import akka.actor.ActorRef
+import akka.actor.DeadLetterSuppression
+import akka.actor.Props
+import akka.actor.Terminated
+import java.util.Optional
+import scala.concurrent.duration.Duration
+
+object BackoffSupervisor {
+
+  /**
+   * @param childProps the [[akka.actor.Props]] of the child actor that
+   *   will be started and supervised
+   * @param childName name of the child actor
+   * @param minBackoff minimum (initial) duration until the child actor will
+   *   started again, if it is terminated
+   * @param maxBackoff the exponential back-off is capped to this duration
+   * @param randomFactor after calculation of the exponential back-off an additional
+   *   random delay based on this factor is added, e.g. 0.2 adds up to 20%
+   *   delay
+   */
+  def props(
+    childProps: Props,
+    childName: String,
+    minBackoff: FiniteDuration,
+    maxBackoff: FiniteDuration,
+    randomFactor: Double): Props = {
+    require(minBackoff > Duration.Zero, "minBackoff must be > 0")
+    require(maxBackoff >= minBackoff, "maxBackoff must be >= minBackoff")
+    require(0.0 <= randomFactor && randomFactor <= 1.0, "randomFactor must be between 0.0 and 1.0")
+    Props(new BackoffSupervisor(childProps, childName, minBackoff, maxBackoff, randomFactor))
+  }
+
+  /**
+   * Send this message to the [[BackoffSupervisor]] and it will reply with
+   * [[BackoffSupervisor.CurrentChild]] containing the `ActorRef` of the current child, if any.
+   */
+  final case object GetCurrentChild
+
+  /**
+   * Java API: Send this message to the [[BackoffSupervisor]] and it will reply with
+   * [[BackoffSupervisor.CurrentChild]] containing the `ActorRef` of the current child, if any.
+   */
+  def getCurrentChild = GetCurrentChild
+
+  final case class CurrentChild(ref: Option[ActorRef]) {
+    /**
+     * Java API: The `ActorRef` of the current child, if any
+     */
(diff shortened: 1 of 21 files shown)


ktoso:
refs #16524

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -92,15 +91,44 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
...
+   * failure.
+   *
    * @param cause failure cause.
+   * @param event the event that was to be persisted
    */
-  private[persistence] def onReplayFailure(cause: Throwable): Unit = ()
+  protected def onPersistFailure(cause: Throwable, event: Any, seqNr: Long): Unit = {
+    log.error(cause, "Failed to persist event type [{}] with sequence number [{}] for persistenceId [{}].",
+      event.getClass.getName, seqNr, persistenceId)
+  }
```
ktoso: Ok, so we'll move to methods for all these things (snapshot failures as well)

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -92,15 +91,44 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
...
+    event match {
+      case Some(evt) ⇒
+        log.error(cause, "Exception in receiveRecover when replaying event type [{}] with sequence number [{}] for " +
+          "persistenceId [{}].", evt.getClass.getName, lastSequenceNr, persistenceId)
+      case None ⇒
+        log.error(cause, "Persistence failure when replaying events for persistenceId [{}]. " +
+          "Last known sequence number [{}]", persistenceId, lastSequenceNr)
+    }
+    context.stop(self)
+  }
```
ktoso: should this not also be user overridable like onPersistFailure? (and always stop after calling it)

ktoso: On the other hand, cleanup can always be done in `postStop`...

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
patriknw: not snapshots, in my opinion, those are not fatal

ktoso: I meant only the moving from messages to callbacks for consistency between those two?

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
patriknw: I was thinking about that, but could not find a use case for it, apart from amending the logging. I don't mind opening it up though, but then the stop will be outside.

ktoso: Could be reason enough IMO, and it's consistent then with onPersistFailure

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
patriknw: I think messages are fine for snapshots, it illustrates the asynchronous (do it in the background) nature of snapshots.
We talked about changing the method `saveSnapshot` to a message also, but my opinion is that a method is good, even though the reply is a message.

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
patriknw: alright, it would also be consistent with how I think PersistentView will work. There it would invoke such a callback, but not stop, since it will periodically retry the replay.

ktoso: +1

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
ktoso: Now I remember, yeah. Summary of handlers is then:

- saveSnapshot - SnapshotSuccess / SnapshotFailure
- TODO: recover - messages / onReplayFailure / RecoveryComplete 
  - // TODO recover or replay?
- persist, persistAsync - onPersistFailure

patriknw:
@ktoso I have added several things to this PR, see the commits. I think it should be complete now, so please review again.

The first commit is the changes from PR #17698 to stay somewhat merge sane. I will merge that PR first, and then rebase here. Please try to review the sharding changes related to this PR by looking at commit fea9d72

akka-persistence/src/main/scala/akka/persistence/BackoffSupervisor.scala
```diff
@@ -0,0 +1,139 @@
...
+
+  def startChild(): Unit = {
+    child = Some(context.watch(context.actorOf(childProps, childName)))
+  }
+
+  def receive = {
+    case Terminated(ref) if child.contains(ref) ⇒
+      child = None
+      val rnd = 1.0 + ThreadLocalRandom.current().nextDouble() * randomFactor
+      val restartDelay = maxBackoff.min(minBackoff * math.pow(2, restartCount)) * rnd
```
patriknw: need some kind of overflow protection here

patriknw:
@ktoso I have fixed the bug: 9684260

patriknw:
rebased

ktoso:
Reviewed the commits made after last review, LGTM - merge at will

patriknw:
thanks

Issue #17630, opened 2015-06-01
FAILED: PersistenceFSMActorSpec


drewhk:
[info] Persistent FSM Actor
[info] - must function as a regular FSM  (817 milliseconds)
[info] - must function as a regular FSM on state timeout *** FAILED *** (1 second, 164 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (1099411045 nanoseconds) during expectMsg while waiting for Transition(Actor[akka://PersistenceSpec/user/$b#2115551935],Shopping,Inactive)
[info]   at scala.Predef$.assert(Predef.scala:165)
[info]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:353)
[info]   at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:330)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:738)
[info]   at

https://jenkins.akka.io:8498/job/akka-multi-node/1078/consoleFull

patriknw:
@leonidb in case you have some spare cycles, otherwise we'll look into it. Note that our CI environment is not very predictive when it come to timing sensitive tests (e.g. rather long 1 second GC pauses are not uncommon).

drewhk:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2488/consoleFull
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2487/consoleFull

leonidb:
@drewhk , @patriknw 
I'm looking at it

leonidb:
@patriknw 
Looks like a timeout related issue.
There are also failures in tests not related to timeouts, but I think that they are affected by previous test failures.

Those tests do pass locally, but are probably too strict for CI environment.

I can change the acceptable timeout behavior from:
```scala
      within(0.9 seconds, 1.1 seconds) {
        expectMsg(Transition(recoveredFsmRef, Shopping, Inactive))
      }
```
to 
```scala
      within(0.5 seconds, 1.5 seconds) {
        expectTerminated(recoveredFsmRef)
      }
```

If that's not enough, we can use longer timeouts in the test actor - 2 seconds and then make the test check if the timeout is ```within(1 seconds, 3 seconds)``` - but it will make the tests run slightly longer.

Tell me what you think.

patriknw:
yes, `within(0.5 seconds, 1.5 seconds)` looks good
1 second stop the world GC at any point is possible.
[...]

PR #17639, opened 2015-06-02
=per #17630 FAILED: PersistenceFSMActorSpec

leonidb:
- Changed the timeouts to make tests less sensitive to environment issues

### DIFF
--- a/akka-persistence/src/test/java/akka/persistence/fsm/AbstractPersistentFsmActorTest.java
+++ b/akka-persistence/src/test/java/akka/persistence/fsm/AbstractPersistentFsmActorTest.java
@@ -124,7 +124,7 @@ public class AbstractPersistentFsmActorTest {
             FSM.Transition stateTransition = expectMsgClass(FSM.Transition.class);
             assertTransition(stateTransition, fsmRef, UserState.LOOKING_AROUND, UserState.SHOPPING);
 
-            new Within(duration("0.9 seconds"), duration("1.1 seconds")) {
+            new Within(duration("0.9 seconds"), duration("1.9 seconds")) {
                 @Override
                 protected void run() {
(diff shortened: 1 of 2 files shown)


ktoso:
OK TO TEST

ktoso:
Thanks for following up on this!
Refs #17630

leonidb:
@ktoso 
Sure, hope this fixes the issue.

leonidb:
@ktoso @patriknw 
On a slightly unrelated note, what's the convenient way to discuss possible improvements of ```PersistentFSM``` - like removing code duplication between the original ```FSM``` class and the clone, timers improvements, users feedback (hopefully :)) etc.?

ktoso:
> On a slightly unrelated note, what's the convenient way to discuss possible improvements of PersistentFSM - like removing code duplication between the original FSM class and the clone, timers improvements, users feedback (hopefully :)) etc.?

There's two channels that work best for such discussions:
- akka-dev - the mailing list dedicated to developing akka - https://groups.google.com/forum/#!forum/akka-dev
- opening issues, if it's something actionable that might end with a PR

There's also the gitter channel if you have something short but we're not constantly monitoring it (like we do with akka-user and akka-dev).

ktoso:
LGTM after marking the timing sensitive tests (those with `within` I'd say) as TimingTest.

Refs https://github.com/akka/akka/issues/17630

ktoso:
Thanks for the update. I guess tagging doesn't work with junit? No idea, will try to check this before we merge. LGTM otherwise.

leonidb:
@ktoso 
I haven't found any usages of the tag in JUnit based tests. I don't think it's possible.
A possible workaround would be to extend the timeout validation limit in the JUnit test.

ktoso:
I'm happy to merge this in its current state, it's an improvement to stability in any case :-)

PR #17502, opened 2015-05-18
Assorted http changes

jrudolph:
This PR relies on the changes of #17500 and needs to be rebased on top of that before tests pass.

(I'm splitting up @sirthias PR #17452 to allow the individual parts to be merged ASAP. I didn't change the commit from #17452, so you may have already reviewed the changes in the context of that one.)

PR #17643, opened 2015-06-03
!htp #17384 - Generalised Security Directives to accept Bearer Token Authentication

pjan:
As per promise to @rkuhn, an improved shot to allow Bearer Token Authentication, taking into accounts the comments made by @jrudolph in #17384, post the #17502 merge. Since the `OAuth2BearerToken` is part of the akka codebase, I opted to add support for it within the http module rather than provide additional facilities to provide it as a third party library.
- Renamed `UserCredentials` to the more general `Credentials`, and its `userName` param to the more appropriate `identifier` in case it covers both `HttpBasicAuthentication` and `BearerTokenAuthentication`.
- Dropped `Basic` from the directives, as they now similarly accept `Authenticator`s that handle `BasicHttpCredentials` and `OAuth2BearerToken`
- Added and amended the tests for both scala and java side

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/SecurityDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/SecurityDirectivesSpec.scala
@@ -11,26 +11,37 @@ import akka.http.scaladsl.model.headers._
 import akka.http.scaladsl.server.AuthenticationFailedRejection.{ CredentialsRejected, CredentialsMissing }
 
 class SecurityDirectivesSpec extends RoutingSpec {
-  val dontAuth = authenticateBasicAsync[String]("MyRealm", _ ⇒ Future.successful(None))
-  val doAuth = authenticateBasicPF("MyRealm", { case UserCredentials.Provided(name) ⇒ name })
-  val authWithAnonymous = doAuth.withAnonymousUser("We are Legion")
+  val dontBasicAuth = authenticateBasicAsync[String]("MyRealm", _ ⇒ Future.successful(None))
+  val dontOAuth2Auth = authenticateOAuth2Async[String]("MyRealm", _ ⇒ Future.successful(None))
+  val doBasicAuth = authenticateBasicPF("MyRealm", { case Credentials.Provided(identifier) ⇒ identifier })
+  val doOAuth2Auth = authenticateOAuth2PF("MyRealm", { case Credentials.Provided(identifier) ⇒ identifier })
+  val authWithAnonymous = doBasicAuth.withAnonymousUser("We are Legion")
 
   val challenge = HttpChallenge("Basic", "MyRealm")
 
   "basic authentication" should {
     "reject requests without Authorization header with an AuthenticationFailedRejection" in {
       Get() ~> {
-        dontAuth { echoComplete }
+        dontBasicAuth { echoComplete }
       } ~> check { rejection shouldEqual AuthenticationFailedRejection(CredentialsMissing, challenge) }
     }
     "reject unauthenticated requests with Authorization header with an AuthenticationFailedRejection" in {
       Get() ~> Authorization(BasicHttpCredentials("Bob", "")) ~> {
-        dontAuth { echoComplete }
+        dontBasicAuth { echoComplete }
       } ~> check { rejection shouldEqual AuthenticationFailedRejection(CredentialsRejected, challenge) }
+    }
+    "reject requests with an OAuth2 Bearer Token Authorization header with 401" in {
+      Get() ~> Authorization(OAuth2BearerToken("myToken")) ~> Route.seal {
+        dontOAuth2Auth { echoComplete }
+      } ~> check {
+        status shouldEqual StatusCodes.Unauthorized
+        responseAs[String] shouldEqual "The supplied authentication is invalid"
+        header[`WWW-Authenticate`] shouldEqual Some(`WWW-Authenticate`(challenge))
+      }
     }
     "reject requests with illegal Authorization header with 401" in {
       Get() ~> RawHeader("Authorization", "bob alice") ~> Route.seal {
-        dontAuth { echoComplete }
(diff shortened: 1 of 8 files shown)


ktoso:
OK TO TEST

drewhk:
Ah, why is this PR tagged as "per", I thought this is persistence related :)

pjan:
@drewhk sorry for the confusion. Updated.

jrudolph:
The API you suggest hides the fact which kind of authentication credentials the user provided, Basic or Bearer authentication. You could even say, that it treats Bearer authentication as a kind of Basic authentication where id and password have the same value.

While, as you said in your original PR, "HttpBasicAuthenticator and BearerTokenAuthenticator have almost exactly the same implementation", I wonder if we should logically treat them that way. Our policy is to be explicit about the HTTP primitives we use in the routing structure.

We are currently a bit short on capacity for driving this further, so, I'm a unsure about where to go with this for now.

drewhk:
Can we close this PR then and open a discussion ticket instead (if there is not yet one)?

sirthias:
I like the intent of providing full support for the `OAuth2BearerToken` as, as @pjan says, we already have a proper model for it but don't yet go all the way for supporting OAuth2 in the routing layer.

Regarding the current impl proposal I am, like @jrudolph, a bit hesitant to drop the `Basic` suffix from all the names. The `authenticateBasic` directive for example is quite clear on what it does. A fully general `authenticate` directive would have to be able to deal with *any* kind of authentication. Not just basic and OAuth. For example, there also is [Digest Authentication](https://en.wikipedia.org/wiki/Digest_access_authentication), which is quite different and, AFAICS, wouldn't fit the `authenticate` API proposed here.

How about we simply add `authenticateOAuth2` and support without touching the existing `authenticateBasic` structures? While that might appear to duplicate a few things I think it would actually improve clarity.

What do you think?

rkuhn:
I agree with keeping the API clean and simple and obvious, possible implementation code duplication can be sorted out independent from this primary concern.

pjan:
Thanks for the constructive feedback and concrete pointers @sirthias and @rkuhn. I'm happy to refactor it to keep the `authenticateBasic` untouched, and mirror it for `authenticateOAuth2` if that's what everyone agrees on.

I understand you're all very limited in bandwidth, and happily try to squeeze out some time to make this happen in return for all the great work you've been doing with Akka. Much much appreciated!

sirthias:
Great! Thanks a lot, @pjan!

drewhk:
We will cut RC4 sometime early next week, after that, we will not add new features until after 1.0

pjan:
Will hence flesh this out tomorrow. Thanks for the notice @drewhk

pjan:
Updated based on the comments. Tests fail on seemingly unrelated parts.

I'm blocking some time to make changes according to comments so this can be finished before RC4 as per notice of @drewhk

rkuhn:
Indeed, the failure is unrelated. Thanks for your commitment!

jypma:
If this is to be changed, I'd like to chime in with another use case. We use both OAuth2 and Basic auth. In our setup, we actually need to send the HTTP basic username and password to a backend system for verification. It'd currently a bit cumbersome that `UserCredential` doesn't allow one to get the password out directly; this basically means we have to manually extract the `HttpCredentials` instead, and pass them into the authenticator separately.

I understand the strong recommendation to do a secure `==` on the passwords, but could the password at least be made available as well? Maybe by a case (sub)class or extractor one can match on?

drewhk:
@pjan this will be merged after we released 1.0, i.e. 1-2 weeks from now. Keep this PR open and we take care of the rest for you.

pjan:
@drewhk now that we're on 1.0, do I need to make any changes?

2m:
@pjan could you rebase this PR so it merges cleanly? Also please squash the last commit which only changes one line to the previous one.

Also please create an issue and then modify commit messages of this PR to follow the [contribution guidelines](https://github.com/akka/akka/blob/master/CONTRIBUTING.md#creating-commits-and-writing-commit-messages). This way we can follow what issues are fixed in what milestone. :)

drewhk:
@jrudolph is this LGTM for you?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/SecurityDirectivesSpec.scala
```diff
@@ -11,26 +11,37 @@ import akka.http.scaladsl.model.headers._
...
     }
     "reject unauthenticated requests with Authorization header with an AuthenticationFailedRejection" in {
       Get() ~> Authorization(BasicHttpCredentials("Bob", "")) ~> {
-        dontAuth { echoComplete }
+        dontBasicAuth { echoComplete }
       } ~> check { rejection shouldEqual AuthenticationFailedRejection(CredentialsRejected, challenge) }
     }
+    "reject requests with an OAuth2 Bearer Token Authorization header with 401" in {
+      Get() ~> Authorization(OAuth2BearerToken("myToken")) ~> Route.seal {
+        dontBasicAuth { echoComplete }
```
rkuhn: shouldn’t this use OAuth?

rkuhn:
@jrudolph Could you check this? Looks good to me, but I’d value experienced input on security directives …

@pjan Would you mind updating so that tests pass and the merge button is green? (and please also make one commit as foreseen in CONTRIBUTING.md) If you can’t make time I’ll pick this up next week and finish it off.

pjan:
@rkuhn @jrudolph took advantage of my one-day old son keeping me awake. Rebased, updated and squashed. Let me know if any other issues need to be addressed.

sirthias:
LGTM and: Congratulations, @pjan, amazing that you find the time for Akka PRs with such a fresh addition to the family! :)

ktoso:
LGTM, thanks a lot for working on this @pjan :-)
I think it should also get some docs, otherwise it's not as easy to discover. I'll open a ticket for documenting this.

PR #17646, opened 2015-06-03
=clu Improve cluster downing

patriknw:
- avoid using Down and Exiting member from being used for joining
- delay shut down of Down member until the information is spread
  to all reachable members, e.g. downing several nodes via one node
- akka.cluster.down-removal-margin setting
  Margin until shards or singletons that belonged to a
  downed/removed partition are created in surviving partition.
  Used by singleton and sharding.
- remove the retry count parameters/settings for singleton in
  favor of deriving those from the removal-margin


rkuhn:
LGTM

patriknw:
rebased

patriknw:

Issue #17586, opened 2015-05-27
Shouldn't PersistentActor be notified if deleting a snapshot fails?


ktoso:
Currently it would be silently ignored.

This relates to https://github.com/akka/akka/issues/15188 in which we should change the SnapshotStore such that `def delete(metadata: SnapshotMetadata)` returns a `Future[Unit]` and not `Unit` like now.

Yes it's a rare case I agree, but feels a bit weird that we don't provide any info if it fails.

patriknw:
makes sense to have same messages as for saveSnapshot

Issue #15188, opened 2014-05-14
Fully asynchronous snapshot store SPI


krasserm:
Simplified example [here](https://github.com/krasserm/akka-persistence-cassandra/blob/v-0.3/src/main/scala/akka/persistence/cassandra/snapshot/CassandraSnapshotStore.scala#L24).

ktoso:
An interesting point we may want to discuss soon. 
// cc @akka/akka-team

ktoso:
Read through the details of what was meant here, yes absolutely we'll have to change `delete` to return a Future.

PR #17647, opened 2015-06-03
!+per snapshot store improvements, (async, errors and configurable attempts)

ktoso:
Resolves https://github.com/akka/akka/issues/17586 and https://github.com/akka/akka/issues/15188 and now https://github.com/akka/akka/issues/17751 as well as the removal of `saved`: https://github.com/akka/akka/issues/17755

Should be complete, includes migration guide and covers both delete scenarios.

TODO:
- [x] sanity check Java side of things is OK


akka-persistence/src/main/scala/akka/persistence/snapshot/SnapshotStore.scala
```diff
@@ -44,10 +44,18 @@ trait SnapshotStore extends Actor {
       delete(metadata)
```
patriknw: is this still ok? should the `evt` be sent to the sender when the future is completed?

akka-persistence/src/main/scala/akka/persistence/snapshot/SnapshotStore.scala
```diff
@@ -81,7 +89,7 @@ trait SnapshotStore extends Actor {
    * @param metadata snapshot metadata.
```
patriknw: while you are here, can you add explicit return type to `saved` (and any other methods in the api)

akka-persistence/src/main/scala/akka/persistence/snapshot/japi/SnapshotStore.scala
```diff
@@ -25,10 +25,10 @@ abstract class SnapshotStore extends SSnapshotStore with SnapshotStorePlugin {
   final def saved(metadata: SnapshotMetadata) =
```
patriknw: `: Unit =`

akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala
```diff
@@ -58,13 +60,18 @@ private[persistence] class LocalSnapshotStore extends SnapshotStore with ActorLo
...
   }
 
-  def delete(metadata: SnapshotMetadata): Unit = {
+  def delete(metadata: SnapshotMetadata): Future[Unit] = {
     saving -= metadata
-    snapshotFile(metadata).delete()
+    Future {
+      snapshotFile(metadata).delete()
+      ()
+    }(ExecutionContexts.sameThreadExecutionContext)
```
patriknw: perhaps

    snapshotFile(metadata).delete()
    Future.successful(())

akka-persistence/src/test/scala/akka/persistence/SnapshotFailureRobustnessSpec.scala
```diff
@@ -30,6 +36,26 @@ object SnapshotFailureRobustnessSpec {
...
+      "akka.persistence.snapshot-store.local-delete-fail"
+
+    override def receiveRecover: Receive = {
+      case SnapshotOffer(md, s) ⇒ probe ! ((md, s))
+      case other                ⇒ probe ! other
+    }
+
+    override def receiveCommand = {
+      case Cmd(payload)            ⇒ persist(payload)(_ ⇒ saveSnapshot(payload))
+      case DeleteSnapshot(seqNr)   ⇒ deleteSnapshot(seqNr, (new Date).getTime) // TODO shouldn't timestamp be optional?
```
patriknw: `System.currentTimeMillis()`

patriknw: if the seqNr is mandatory I don't see the need for the timestamp at all

akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala
ktoso: :+1: thanks

akka-persistence/src/main/scala/akka/persistence/snapshot/SnapshotStore.scala
ktoso: :+1:

akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala
ktoso: Actually no, the `Future {}` is nicer because if an exception is thrown it's captured in the exception nicely

patriknw: ok, probably a matter of taste, but I would prefer

    Future.fromTry(Try(snapshotFile(metadata).delete()))

(sameThreadExecutionContext is internal)

ktoso: nice one, never used that before :-) I'll do that

akka-persistence/src/main/scala/akka/persistence/snapshot/SnapshotStore.scala
ktoso: As I read if it is valid, flow is:

* save future fails 
* gets converted to SaveSnapshotFailure and piped `to (self, p)` 
* which means sender for `SaveSnapshotFailure` is received, we issue a delete (that line)
* and send the event to the original persistent actor.

akka-persistence/src/test/scala/akka/persistence/SnapshotFailureRobustnessSpec.scala
ktoso: In theory both can be used and then are used in:

```
  private[persistence] def matches(metadata: SnapshotMetadata): Boolean =
    metadata.sequenceNr <= maxSequenceNr && metadata.timestamp <= maxTimestamp
```

Question is if we want to keep as-is, change it to Option or model explicitly as another deletion mode?

patriknw: That is confusing. I would expect that `deleteSnapshot` deletes one single snapshot that is uniquely identified by the sequence number. For fuzzy matching there is `deleteSnapshots`.

ktoso:
This PR also addresses https://github.com/akka/akka/issues/15188 "fully async snapshot store"

akka-persistence/src/test/scala/akka/persistence/SnapshotFailureRobustnessSpec.scala
ktoso: Agreed, I'll remove the timestamp from this message

ktoso: Ah now I see... it's reusing the metadata type that is stored with snapshots. Let me re think this

ktoso:
Squashed a bit into "steps", and now also resolving https://github.com/akka/akka/issues/17755 (removal of `saved` in favour of `receivePluginInternal` (or should we name it `receiveOther`?)).

ktoso:
Please review when you have a moment @patriknw

akka-docs/rst/project/migration-guide-persistence-experimental-2.3.x-2.4.x.rst
```diff
@@ -204,3 +203,67 @@ To continue using LevelDB based persistence plugins it is now required for relat
 to include an additional explicit dependency declaration for the LevelDB artifacts. 
 This change allows production akka deployments to avoid need for the LevelDB provisioning. 
 Please see persistence extension ``reference.conf`` for details. 
+
```
patriknw: um, this migration guide was intended for the changes that we made between 2.3.3 to 2.3.4.
I would guess that most users have already migrated to >= 2.3.4 and this guide is not interesting for them.
Should we put new (2.3.11 -> 2.4.0) changes in the ordinary migration guide.

ktoso: Yeah, good idea :+1:

ktoso: Maybe just a section in this file? Technically it is the "2.3 -> 2.4" file.

akka-docs/rst/project/migration-guide-persistence-experimental-2.3.x-2.4.x.rst
```diff
@@ -204,3 +203,67 @@ To continue using LevelDB based persistence plugins it is now required for relat
...
+    // previously
+    class MySnapshots extends SnapshotStore {
+      // old API:
+      // def saved(meta: SnapshotMetadata): Unit = doThings()
+
+      // new API:
+      def saveAsync(metadata: SnapshotMetadata, snapshot: Any): Future[Unit] = {
+         val f: Future[Unit] = ??? // completion or failure triggers messages in receivePluginInternal
+         self ! MyCustomMessage(metadata.persistenceId)
```
patriknw: no pipeTo?

akka-persistence/src/main/scala/akka/persistence/snapshot/SnapshotStore.scala
```diff
@@ -14,14 +14,16 @@ import akka.persistence._
...
-trait SnapshotStore extends Actor {
+trait SnapshotStore extends Actor with ActorLogging {
   import SnapshotProtocol._
   import context.dispatcher
 
   private val extension = Persistence(context.system)
   private val publish = extension.settings.internal.publishPluginCommands
 
-  final def receive = {
+  final def receive = receiveSnapshotStore orElse receivePluginInternal
```
patriknw: there is a warning here, use:

    receiveSnapshotStore.orElse[Any, Unit](receivePluginInternal)

akka-persistence/src/main/scala/akka/persistence/snapshot/SnapshotStore.scala
```diff
@@ -37,21 +40,49 @@ trait SnapshotStore extends Actor {
...
+      } finally sender() ! evt // sender is persistentActor
+
     case d @ DeleteSnapshot(metadata) ⇒
-      delete(metadata)
+      val p = sender()
+      deleteAsync(metadata) map {
+        case _ ⇒ DeleteSnapshotSuccess(metadata)
+      } recover {
+        case e ⇒ DeleteSnapshotFailure(metadata, e)
+      } to (self, p)
```
patriknw: no need to freeze the sender when using pipeTo

akka-persistence/src/main/scala/akka/persistence/snapshot/SnapshotStore.scala
```diff
@@ -37,21 +40,49 @@ trait SnapshotStore extends Actor {
...
+      try tryReceivePluginInternal(evt) finally sender() ! evt // sender is persistentActor
+
     case d @ DeleteSnapshots(persistenceId, criteria) ⇒
-      delete(persistenceId, criteria)
+      val p = sender()
+      deleteAsync(persistenceId, criteria) map {
+        case _ ⇒ DeleteSnapshotsSuccess(criteria)
+      } recover {
+        case e ⇒ DeleteSnapshotsFailure(criteria, e)
+      } to (self, p)
```
patriknw: same here

akka-persistence/src/main/scala/akka/persistence/snapshot/SnapshotStore.scala
ktoso: it's more for documentation reasons (followed the pattern that was applied in this file), that "sender is the p - persistent actor".

Shall I remove the frozen p's everywhere?

akka-persistence/src/main/scala/akka/persistence/snapshot/japi/SnapshotStore.scala
```diff
@@ -16,19 +16,16 @@ import akka.persistence.snapshot.{ SnapshotStore ⇒ SSnapshotStore }
 abstract class SnapshotStore extends SSnapshotStore with SnapshotStorePlugin {
   import context.dispatcher
 
-  final def loadAsync(persistenceId: String, criteria: SnapshotSelectionCriteria) =
-    doLoadAsync(persistenceId, criteria).map(_.asScala)
+  override final def loadAsync(persistenceId: String, criteria: SnapshotSelectionCriteria) =
```
patriknw: explicit return type

patriknw:
LGTM, aside from some minor comments

akka-persistence/src/main/scala/akka/persistence/snapshot/SnapshotStore.scala
ktoso: Thanks!

akka-docs/rst/project/migration-guide-persistence-experimental-2.3.x-2.4.x.rst
patriknw: rename the file, and remove the (and 2.4.x) in the header

ktoso: Ah `migration-guide-2.3.x-2.4.x.rst` - yeah, makes sense. Will do

akka-docs/rst/project/migration-guide-persistence-experimental-2.3.x-2.4.x.rst
ktoso: yup, adding

akka-persistence/src/main/scala/akka/persistence/snapshot/SnapshotStore.scala
patriknw: doesn't matter, up to you (it's technically not needed)

btw I normally use `pipeTo` instead of `to`, is there a difference?

ktoso: I'll go with the comments then.

to is a more explicit pipeTo when you want to show the sender:
`to(recipient: ActorRef, sender: ActorRef) == pipeTo(recipient: ActorRef)(implicit sender: ActorRef)`

`pipeTo` doesn't work nicely in this fluent style - i.e. this doesn't compile: `} pipeTo (self)(sender()) // sender is persistentActor`

akka-persistence/src/main/scala/akka/persistence/snapshot/SnapshotStore.scala
ktoso: Went with:

```scala
    case evt: DeleteSnapshotsSuccess ⇒
      try tryReceivePluginInternal(evt) finally senderPersistentActor() ! evt
  }

  /** Documents intent that the sender() is expected to be the PersistentActor */
  @inline private final def senderPersistentActor(): ActorRef = sender()
```

akka-persistence/src/main/scala/akka/persistence/snapshot/japi/SnapshotStore.scala
ktoso: done thanks

ktoso:
Addressed comments and squashed a bit - it's 3 separate ones, not completely separated per-feature (they are pretty connected) but mostly.

ktoso:
Added a small test for the optional converter, just to be sure.

ktoso:
Will merge after validation.

ktoso:
PLS BUILD

akka-docs/rst/project/migration-guide-2.3.x-2.4.x.rst
```diff
@@ -315,3 +292,97 @@ actor external actor of how to allocate shards or rebalance shards.
 
 For the synchronous case you can return the result via ``scala.concurrent.Future.successful`` in Scala or 
 ``akka.dispatch.Futures.successful`` in Java.
+
+################
+Akka Persistence
+################
```
patriknw: isn't this wrong, this is the same level as in the top.

I think we can have:

    Akka Persistence
    =============

and

    Persistence Plugin API
    ==================

patriknw:
LGTM, aside from one remark regarding the migration guide

patriknw:
failed on `akka-sample-persistence-java-lambda/compile:compile` (easy to forget that one)

ktoso:
Ah lambda, I always forget about it, thanks for checking - will address comments and fix that proj now

akka-docs/rst/project/migration-guide-2.3.x-2.4.x.rst
ktoso: Very true, fixed

ktoso:
Rebased / fixed comments, will merge after validation

PR #17654, opened 2015-06-04
BodyPartParser fixes

jrudolph:
This only fixes the newly found part of #17051 and actually breaks another test that succeeded for the wrong reasons.

/cc @sirthias @RichardBradley


drewhk:
There is one real failure in validation: 

[info]   - a missing header-separating CRLF (in Default entity) *** FAILED *** (24 milliseconds)
[info]     java.util.NoSuchElementException: Future.failed not completed with a throwable.
[info]     at scala.concurrent.Future$$anonfun$failed$1.apply(Future.scala:192)
[info]     at scala.concurrent.Future$$anonfun$failed$1.apply(Future.scala:190)
[info]     at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:32)
[info]     at scala.concurrent.Future$InternalCallbackExecutor$Batch$$anonfun$run$1.processBatch$1(Future.scala:643)
[info]     at scala.concurrent.Future$InternalCallbackExecutor$Batch$$anonfun$run$1.apply$mcV$sp(Future.scala:658)
[info]     at scala.concurrent.Future$InternalCallbackExecutor$Batch$$anonfun$run$1.apply(Future.scala:635)
[info]     at scala.concurren

jrudolph:
Yes, this isn't complete yet.

sirthias:
Ok, added one more commit which should now fix #17051 for good, and rebased.

jrudolph:
Cool, this seemed easy enough in hindsight :) LGTM

sirthias:
Yeah. Figuring out what's going on was the hardest part.
The whole thing unveils a more fundamental problem that we have in our parsers.

When generating errors, in principle, we'd have to have the parsers produce errors in two categories:

1. Errors within the *start part* of a request, response or entity part
2. Errors within the *data stream* of a request, response or entity part

The errors from the first category need to be surfaced on the *outer* stream level (i.e. the request, response or entity part stream). The errors from the second category need to be surfaced on the *inner* stream level (i.e. the respective data stream).

Currently we don't make that distinction at all.
In the body part parser there currently don't seem to be errors of the 2nd kind.
But in the main request/response parser there are.

jrudolph:
@sirthias yes, I agree. This needs to be considered for every stream of streams scenario anew. It also arose in the Websocket implementation.

> In the body part parser there currently don't seem to be errors of the 2nd kind.

I think that's correct. As the parser usually just scans for the boundary there seems to be little that can go wrong.

drewhk:
LGTM

Is there anything missing, or can I merge?

sirthias:
Nothing missing from my side. Good to merge, I think.

drewhk:
What are the corresponding tickets to close? Do you close them?

jrudolph:
It was only #17051 which I just closed.

PR #17659, opened 2015-06-05
+htp #16810 use flushing deflate call on Java > 6

### LINKED ISSUE #16810: For Deflate & Gzip compression use proper flushing APIs instead of hack
Currently we use this hack to ensure that the `java.util.zip.Deflater` flushes its buffers after each chunk:

```scala
  protected def flushWithBuffer(buffer: Array[Byte]): ByteString = {
    // trick the deflater into flushing: switch compression level
    // FIXME: use proper APIs and SYNC_FLUSH when Java 6 support is dropped
    deflater.deflate(EmptyByteArray, 0, 0)
    deflater.setLevel(Deflater.NO_COMPRESSION)
    val res1 = drain(buffer)
    deflater.setLevel(Deflater.BEST_COMPRESSION)
    val res2 = drain(buffer)
    res1 ++ res2
  }
```

Why this works is completely undocumented and e.g. doesn't work on Android.

Instead, the [deflate overload introduced in Java 7 that allows specifying a flush mode](https://docs.oracle.com/javase/7/docs/api/java/util/zip/Deflater.html#deflate(byte[],%20int,%20int,%20int)) should be used.

See spray/spray#1005.

jrudolph:
Fixes #16810.

/cc @sirthias

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
@@ -28,15 +28,22 @@ private[http] object StreamUtils {
   /**
    * Creates a transformer that will call `f` for each incoming ByteString and output its result. After the complete
    * input has been read it will call `finish` once to determine the final ByteString to post to the output.
+   * Empty ByteStrings are discarded.
    */
   def byteStringTransformer(f: ByteString ⇒ ByteString, finish: () ⇒ ByteString): Stage[ByteString, ByteString] = {
     new PushPullStage[ByteString, ByteString] {
-      override def onPush(element: ByteString, ctx: Context[ByteString]): SyncDirective =
-        ctx.push(f(element))
+      override def onPush(element: ByteString, ctx: Context[ByteString]): SyncDirective = {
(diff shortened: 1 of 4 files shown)


jrudolph:
Failures in TlsSpec.

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -27,16 +27,23 @@ private[http] object StreamUtils {
 
   /**
    * Creates a transformer that will call `f` for each incoming ByteString and output its result. After the complete
-   * input has been read it will call `finish` once to determine the final ByteString to post to the output.
+   * input has been read it will call `finish` once to determine the final ByteString to post to the output. Empty elements
+   * are filtered out.
```
viktorklang: May I suggest "Empty ByteStrings are discarded."

jrudolph: Yes, thanks. Updated.

drewhk:
LGTM

drewhk:
Fixes #16810

Issue #17609, opened 2015-05-29
Create overload for UpgradeToWebsocket.handleMessages that takes separate source and sink


jrudolph:
Just for cosmetic reasons to supply those separately if both ends of the bidirectional websocket channel aren't logically linked.

See http://stackoverflow.com/questions/30479777/integrating-an-ack-based-actor-with-akka-stream?noredirect=1#comment49124618_30479777 for a bit of background.

/cc @sirthias

fommil:
I'd be happy with it taking a `Flow` if there was a simple one liner to convert a Sink/Source to a Flow for this purpose.

drewhk:
Flow.wrap is what you are after

fommil:
`Flow.wrap` takes a `Graph`

fommil:
can this then be closed?

jrudolph:
Let's keep it open to either add the overload or at least a documentation comment how to achieve the same with `Flow.wrap`.

fommil:
hold on

```
wrap[I, O, M1, M2, M](sink: Graph[SinkShape[I], M1], source: Graph[SourceShape[O], M2])(f: (M1, M2) ⇒ M)
```

That means providing  function. This isn't as convenient as I thought it was.

jrudolph:
You can use one of the `Keep.XXX` for the function.

drewhk:
And Source/Sink are subclasses of Graph[SourceShape]/Graph[SinkShape]

fommil:
neither the `f` nor the `Keep` is documented, so I can only guess as to what it means. Assuming `M1` and `M2` are the `Outgoing` and `Incoming` messages, respectively, then I don't see when one would ever have to choose between such messages.

danielwegener:
@jrudolph Just as I stumbled upon this and its kind of the same place: Is there yet any user api to negotiate the ws-protocol between server and the client?  `handleWebsocketMessages` does seem to deal only with an unspecified protocol (and rejects any other). The following does the trick somehow but it might pretty much be the wrong place.

```scala
def handleWebsocketMessagesWithProtocol(protocol: String, handler: Flow[Message, Message, Any]): Route =
      extractFlowMaterializer { implicit mat ⇒
        optionalHeaderValueByType[UpgradeToWebsocket]() {
          case Some(upgrade) if !upgrade.requestedProtocols.contains(protocol) ⇒ complete(upgrade.handleMessages(handler, None))
[...]

PR #17552, opened 2015-05-22
=htp add explicit result types to formField/parameter implicits

jrudolph:
Otherwise, the applicability of the implicit conversions can depend on the order of compiler-internal loading/typing which is hardly predictable.

Thanks to @odersky for [giving the right hint](https://groups.google.com/d/topic/scala-language/KZhcHb6e8DI/discussion).

/cc @sirthias

Issue #17067, opened 2015-03-24
Port XXE fix from spray


jrudolph:
In the process it might make sense to pull in an explicit dependency on xerces like play does. As we have to do this only for akka-http-xml this dependency seems well-contained enough and will guarantee that the code will work reliably across JDKs.

spray fix: spray/spray#1024
play fix: https://github.com/playframework/playframework/pull/3480

In the future we may rely on a fix directly in Scala XML: https://github.com/scala/scala-xml/issues/17

/cc @sirthias

jrudolph:
Fixed in #17660.

PR #17660, opened 2015-06-05
Fixes for #16839, #17067, #17609

### LINKED ISSUE #16839: Factor out `EOI` in header parsing so that header values won't be accidentally truncated if not parseable
This may be fixed with the new parboiled2-based grammar but we should assure that it really is.

See spray/spray#280

/cc @sirthias

jrudolph:
Seven headers still allowed trailing garbage (#16839).

I added even more implicit def result types (#17552) because the compilation problems came back another time...

/cc @sirthias

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
@@ -451,7 +451,7 @@ private[http] object HttpHeaderParser {
       // TODO: optimize by running the header value parser directly on the input ByteString (rather than an extracted String)
       val (headerValue, endIx) = scanHeaderValue(input, valueStart, valueStart + maxHeaderValueLength)()
       val trimmedHeaderValue = headerValue.trim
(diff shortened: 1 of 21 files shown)


jrudolph:
Added a fix for #17609. @fommil would that suffice for your purposes?

fommil:
@jrudolph the `FlowMaterializer` needs some explanation. what's its purpose here? I thought you were going to add a convenience that only required a source and sink (on the `Flow` object)

jrudolph:
The `FlowMaterializer` is completely unused. Thanks for spotting that!

Wrt adding simple `Flow.wrap` overloads ignoring the materialized values WDYT @drewhk, @rkuhn?

jrudolph:
Removed the materializer everywhere.

fommil:
awesome, thanks @jrudolph !

rkuhn:
Concerning the `Flow.wrap` materialized value transformation, we should probably add a `Keep.none` that you can use instead of the `(_, _) => ()` literal. An overload that does not take that argument list will probably not be legal.

drewhk:
+1 for Keep.none

drewhk:
Added ticket for Keep.none: https://github.com/akka/akka/issues/17702

Issue #17662, opened 2015-06-05
Make Sink.ignore return Future[Unit] rather than Unit


viktorklang:
Makes it much easier to deal with "when done do this"

2m:
Fixed in #17666

PR #17666, opened 2015-06-05
+str - #17662 - Changes Sink.ignore to return a Future[Unit]

viktorklang:
Not only does a Future signal that the processing happens elsewhere, but it is also highly useful when wanting to do something at the end of processing a stream (or substream for that matter) or join a substream with its enclosing stream (mapAsync for instance)


drewhk:
LGTM

viktorklang:
Can I merge this?

PR #17674, opened 2015-06-06
=str - Introduces a TCK-verified SingleElementPublisher to optimize S…

viktorklang:
…ource.single and things depending on it

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
@@ -243,13 +243,16 @@ object Source extends SourceApply {
    * Every connected `Sink` of this stream will see an individual stream consisting of one element.
    */
   def single[T](element: T): Source[T, Unit] =
-    apply(SingleElementPublisher(element, "SingleSource")).withAttributes(DefaultAttributes.singleSource) // FIXME optimize
+    apply(SingleElementPublisher(element, "SingleSource")).withAttributes(DefaultAttributes.singleSource)
 
   /**
    * Create a `Source` that will continually emit the given element.
    */
   def repeat[T](element: T): Source[T, Unit] =
-    apply(() ⇒ Iterator.continually(element)).withAttributes(DefaultAttributes.repeat) // FIXME optimize
+    apply(new immutable.Iterable[T] {
+      override val iterator: Iterator[T] = Iterator.continually(element)
+      override def toString: String = "Iterable.continually(" + element + ")"
+    }).withAttributes(DefaultAttributes.repeat)
 
   /**
    * A `Source` with no elements, i.e. an empty stream that is completed immediately for every connected `Sink`.


ktoso:
LGTM :-)

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -242,7 +238,7 @@ object Source extends SourceApply {
    * Every connected `Sink` of this stream will see an individual stream consisting of one element.
    */
   def single[T](element: T): Source[T, Unit] =
-    apply(SynchronousIterablePublisher(List(element), "SingleSource")).withAttributes(DefaultAttributes.singleSource) // FIXME optimize
+    apply(SingleElementPublisher(element, "SingleSource")).withAttributes(DefaultAttributes.singleSource) // FIXME optimize
```
drewhk: remove the TODO comment

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
```diff
@@ -40,6 +39,36 @@ private[akka] final case class ErrorPublisher(t: Throwable, name: String) extend
   override def toString: String = name
 }
 
+private[akka] final case class SingleElementPublisher[T](value: T, name: String) extends Publisher[T] {
+  import ReactiveStreamsCompliance._
+
+  private[this] class SingleElementSubscription(subscriber: Subscriber[_ >: T]) extends Subscription {
+    private[this] var done: Boolean = false
```
drewhk: doesn't need to be volatile?

viktorklang: Since it is only ever written to by the calling thread (the subscriber) the subscriber has to arrange for its own visibility between its own calls so I figured no such volatile was required.

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
viktorklang: With pleasure

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
drewhk: yeah, only the initial value should be seen, since constructor and subscribe is called from a different thread than the one who calls request()

viktorklang: Sure, but onSubscribe must `happen-before` the Subscriber can call `request` which must `happen-before` any subsequent such calls or the calls to `onError` or `onComplete` (which would be done in the calling thread (of the subscriber) anyway. So `onSubscribe` must arrange for the safe publication of said `Subscription` to the `Subscriber`.

With code like this it is always possible to be wrong, but I think it holds. :)

viktorklang:
Can I merge this?

drewhk:
Needs rebase

viktorklang:
rebased

PR #17677, opened 2015-06-06
=htp - Update cookie parsing rules based on twitter.com interop

briantopping:
Update cookie parsing rules based on twitter.com return results:
- RFC 6265 states "Note that a leading %x2E ("."), if present is ignored even though that character is not permitted", but that was not reflected in these rules
- The date that was coming back seemed reasonable of "05-Jun-2017", but the parser rule for a date with dashes only accepted a two-digit year

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
@@ -78,7 +78,7 @@ private[parser] trait CommonRules { this: Parser with StringBuilding ⇒
   def `day-name` = rule(
     "Sun" ~ push(0) | "Mon" ~ push(1) | "Tue" ~ push(2) | "Wed" ~ push(3) | "Thu" ~ push(4) | "Fri" ~ push(5) | "Sat" ~ push(6))
 
-  def date1 = rule { day ~ ' ' ~ month ~ ' ' ~ year }
+  def date1 = rule { day ~ `date-sep` ~ month ~ `date-sep` ~ year }
 
   def day = rule { digit2 }
 
@@ -97,7 +97,8 @@ private[parser] trait CommonRules { this: Parser with StringBuilding ⇒
 
   // def `rfc850-date` = rule { `day-name-l` ~ ", " ~ date2 ~ ' ' ~ `time-of-day` ~ " GMT" }
 
-  def date2 = rule { day ~ '-' ~ month ~ '-' ~ digit2 }
+  // per #17714, parse two digit year to https://tools.ietf.org/html/rfc6265#section-5.1.1
+  def date2 = rule { day ~ '-' ~ month ~ '-' ~ digit2 ~> (y ⇒ if (y <= 69) y + 2000 else y + 1900) }
 
   def `day-name-l` = rule(
     "Sunday" ~ push(0) | "Monday" ~ push(1) | "Tuesday" ~ push(2) | "Wednesday" ~ push(3) | "Thursday" ~ push(4) |
@@ -246,7 +247,7 @@ private[parser] trait CommonRules { this: Parser with StringBuilding ⇒
   // https://tools.ietf.org/html/rfc1034#section-3.5 relaxed by https://tools.ietf.org/html/rfc1123#section-2
   // to also allow digits at the start of a label
   def `domain-value` = rule {
-    capture(oneOrMore(oneOrMore(oneOrMore(ALPHANUM)).separatedBy('-')).separatedBy('.')) ~ OWS
+    optional('.') ~ capture(oneOrMore(oneOrMore(oneOrMore(ALPHANUM)).separatedBy('-')).separatedBy('.')) ~ OWS
   }
 
   def `path-av` = rule {
--- a/akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
@@ -228,6 +228,12 @@ class HttpHeaderSpec extends FreeSpec with Matchers {
 
     "Date" in {
       "Date: Wed, 13 Jul 2011 08:12:31 GMT" =!= Date(DateTime(2011, 7, 13, 8, 12, 31))
+      "Date: Wed, 13-Jul-2011 08:12:31 GMT" =!= Date(DateTime(2011, 7, 13, 8, 12, 31)).renderedTo(
+        "Wed, 13 Jul 2011 08:12:31 GMT")
+      "Date: Wed, 13-Jul-11 08:12:31 GMT" =!= Date(DateTime(2011, 7, 13, 8, 12, 31)).renderedTo(
+        "Wed, 13 Jul 2011 08:12:31 GMT")
+      "Date: Mon, 13-Jul-70 08:12:31 GMT" =!= Date(DateTime(1970, 7, 13, 8, 12, 31)).renderedTo(
+        "Mon, 13 Jul 1970 08:12:31 GMT")
       "Date: Fri, 23 Mar 1804 12:11:10 UTC" =!= Date(DateTime(1804, 3, 23, 12, 11, 10)).renderedTo(
         "Fri, 23 Mar 1804 12:11:10 GMT")
     }
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/CharacterClasses.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/CharacterClasses.scala
@@ -75,6 +75,7 @@ private[http] object CharacterClasses {
   val DIGIT05 = CharPredicate('0' to '5')
   def DIGIT19 = CharPredicate.Digit19
   val colonSlashEOI = CharPredicate(':', '/', akka.parboiled2.EOI)
+  val `date-sep` = CharPredicate("""- """)
 
   require(`qdtext-base`.isMaskBased) // make sure we didn't introduce any non-7bit-chars by accident which
   require(`ctext-base`.isMaskBased) // would make the CharPredicate fall back to the much slower


briantopping:
I'll get the subject right one way or another lol

drewhk:
What is your opinion @jrudolph @sirthias ?

drewhk:
OK TO TEST

jrudolph:
RFC 6265 specifies the year as:

```
year            = 2*4DIGIT ( non-digit *OCTET )
```

which means we must be able to parse any years of length 2 to 4. Changing `date2` to also accept 4 character-years seems a bit strange :) 

I think we need to review which kind of dates to accept for `Set-Cookie` comparing the cookie RFC 6265 with RFC 7231 linked in the sources about general date parsing with the existing parser which already seems to define some extra rules and then fix that 2-digit years aren't accepted at all (see #17714).

sirthias:
My proposal here would be to not relax `date2` to also accept 4 digit dates but to relax `date1` to also accept dashes as delimiters.
Plus, as @jrudolph already pointed out, it'd be great if you could also fix #17714 while you are at it by adding a small "post-processing" action to `date2` that implements the year expansion logic quoted by @jrudolph in #17714. This way we only need to touch this part of the code once and minimize merge conflict risk.

jrudolph:
I think one theoretic problem is that we have only one definition of `HTTP-Date` but conflicting rules of how to do the expansions (a quick glance revealed "MUST interpret a timestamp that appears to be more    than 50 years in the future as representing the most recent year in the past that had the same last two digits." from [RFC 7231](http://tools.ietf.org/html/rfc7231#section-7.1.1.1), and "If the year-value is greater than or equal to 70 and less than or equal to 99, increment the year-value by 1900." from [RFC 6265](https://tools.ietf.org/html/rfc6265#section-5.1.1).

I think we should just choose one definition and document our choice. Also not all the dates in HTTP may use the same format but I think it is ok to use a superset that covers all known date formats as long as there are no known ambiguities.

briantopping:
@sirthias That makes sense about `date2`, I had misinterpreted the name to mean "a second form of date", not "a date with a two digit year".

sirthias:
@briantopping I don't think your take is actually correct and not a misinterpretation at all. We simply followed the names of the rules in the actual specs, and `date2` really appears to be nothing but a 2nd variant. However, if we adapt `date1` instead of `date2` then we can easily attach the year-expansion in `date2` without having to discriminate between 4-digit and 2-digit dates.

Apart from that I agree with @jrudolph in that we should pick one spec for the expansion. For me the newer [RFC 7231](http://tools.ietf.org/html/rfc7231#section-7.1.1.1) seems like the way to go.
And being as lenient as we can as long as nothing becomes ambiguous is also a good thing.

briantopping:
Ugh, that's not right, I think I did my merge improperly...

sirthias:
Damn, just realized that I wrote "I don't think your take is actually correct" instead of "I don't think your take is actually *incorrect*" as I meant to. Sorry about that, Brian!

briantopping:
@sirthias I'm barely awake, didn't even notice, wouldn't worry about it if I did ;)

sirthias:
@briantopping Ok, great. And "Good Morning" to the west-coast! :)

briantopping:
@sirthias I've been in Thailand for most of the last six months, but giving two weeks of training next week in San Diego, so I started adjusting my schedule a week ago. Gonna bicycle to the 24 hour starbucks after rinsing my head out... :)

akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
```diff
@@ -78,7 +78,9 @@ private[parser] trait CommonRules { this: Parser with StringBuilding ⇒
   def `day-name` = rule(
     "Sun" ~ push(0) | "Mon" ~ push(1) | "Tue" ~ push(2) | "Wed" ~ push(3) | "Thu" ~ push(4) | "Fri" ~ push(5) | "Sat" ~ push(6))
 
-  def date1 = rule { day ~ ' ' ~ month ~ ' ' ~ year }
+  def `date-sep` = rule { " " | "-" }
```
sirthias: For rules that simply test membership in a certain set of chars it's most efficient if you simply add a new `CharPredicate` instance to `akka.http.impl.model.parser.CharacterClasses` instead of solving it via a parser rule.

akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
```diff
@@ -228,6 +228,8 @@ class HttpHeaderSpec extends FreeSpec with Matchers {
 
     "Date" in {
       "Date: Wed, 13 Jul 2011 08:12:31 GMT" =!= Date(DateTime(2011, 7, 13, 8, 12, 31))
+// TODO this test fails because we only intend to read this format, not have bidirection generation of it
+//      "Date: Wed, 13-Jul-2011 08:12:31 GMT" =!= Date(DateTime(2011, 7, 13, 8, 12, 31))
```
sirthias: you can simply add a `.renderedTo(...)` clause at the very end to get around the problem you describe in the `TODO`.
The subsequent example shows how.

sirthias:
Nice! Do you also want to fix #17714 while you are at it?
The year expansion logic discussed above ("interpret a timestamp that appears to be more than 50 years in the future as representing the most recent year in the past that had the same last two digits") would be the way to go I think.

briantopping:
Sure, I'll give #17714 a go since I'm here. Will flatten these commits once I'm done.

briantopping:
Okay, I fixed this to https://tools.ietf.org/html/rfc6265#section-5.1.1 in #17714 instead of the "50 years in the future" heuristic because it seems the internet can't agree on a pivot number and the RFC is required for our use cases. WDYT?

briantopping:
I just totally exploded this branch, hold on...

akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
```diff
@@ -4,6 +4,9 @@
 
 package akka.http.impl.model.parser
 
+import java.time.LocalDate
+import java.util.Calendar
+
```
sirthias: These imports make me nervous! :)
They appear to be unused, right?

akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
```diff
@@ -403,7 +406,13 @@ private[parser] trait CommonRules { this: Parser with StringBuilding ⇒
   private def digitInt(c: Char): Int = c - '0'
   private def digitLong(c: Char): Long = (c - '0').toLong
 
-  private def createDateTime(year: Int, month: Int, day: Int, hour: Int, min: Int, sec: Int, wkday: Int) = {
+  private def createDateTime(rawYear: Int, month: Int, day: Int, hour: Int, min: Int, sec: Int, wkday: Int) = {
```
sirthias: If you add the year expansion logic here the additional comparisons will be run *everytime*, even for the large majority of cases where we matched a 4-digit year and know that we will always hit the default branch.
If you add the year expansion logic to the end of the `date2` rule it will only run when actually required.
Additional benefit: you'll only need a single comparison (`_ <= 69`) instead of 4.

sirthias:
If you prefer the year expansion of RFC6265 over the one from RFC7231 then let's go with that one.
The implementation is shorter and a bit more efficient anyways.

drewhk:
I guess after those minor tweaks are done, this is mergeable?

sirthias:
@drewhk Yes, I'd say so.

akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
briantopping: Oops, they were there from the older "50 year" comparison. Will remove.

briantopping:
Ok, I believe that should do it. Thanks for all the feedback, it's been great to learn from you guys!

akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
```diff
@@ -97,7 +97,7 @@ private[parser] trait CommonRules { this: Parser with StringBuilding ⇒
 
   // def `rfc850-date` = rule { `day-name-l` ~ ", " ~ date2 ~ ' ' ~ `time-of-day` ~ " GMT" }
 
-  def date2 = rule { day ~ '-' ~ month ~ '-' ~ digit2 }
+  def date2 = rule { day ~ '-' ~ month ~ '-' ~ digit2 ~> { (year) => push(rfc6265Year(year))} }
```
sirthias: How about we simplify this to:

```scala
 ... ~> (y => if (y <= 69) y + 2000 else y + 1900) // https://tools.ietf.org/html/rfc6265#section-5.1.1
```

Saves us a few lines of code and is arguably even clearer.

briantopping: Agreed and pushed. I'm new to your parser, very nice!

sirthias:
Excellent, thanks, Brian!
Ready for merge from my side.

briantopping:
Jenkins got an OOM error.

drewhk:
Jenkins+slaves are/were being restarted

drewhk:
PLS BUILD

briantopping:
Weird... 

```
[ERROR] [06/18/2015 13:56:51.393] [docs-http-scaladsl-server-directives-FutureDirectivesExamplesSpec-akka.actor.default-dispatcher-3] [ActorSystem(docs-http-scaladsl-server-directives-FutureDirectivesExamplesSpec)] StreamTestDefaultMailbox assertion failed: assertion failed: Don't use anonymous actor classes, actor class for Actor[akka://docs-http-scaladsl-server-directives-FutureDirectivesExamplesSpec/user/$b#-1551774463] was [akka.actor.Actor]
java.lang.AssertionError: assertion failed: Don't use anonymous actor classes, actor class for Actor[akka://docs-http-scaladsl-server-directives-FutureDirectivesExamplesSpec/user/$b#-1551774463] was [akka.actor.Actor]
	at scala.Predef$.assert(Predef.scala:179)
	at akka.stream.testkit.StreamTestDefaultMailbox.create(StreamTestDefaultMailbox.scala:29)
	at akka.dispatch.Dispatcher$$anon$1.<init>(Dispatcher.scala:93)
	at akka.dispatch.Dispatcher.createMailbox(Dispatcher.scala:93)
	at akka.actor.dungeon.Dispatch$class.init(Dispatch.scala:47)
	at akka.actor.ActorCell.init(ActorCell.scala:369)
	at akka.actor.RepointableActorRef.newCell(RepointableActorRef.scala:117)
	at akka.actor.RepointableActorRef.point(RepointableActorRef.scala:103)
	at akka.actor.ActorCell.handleSupervise(ActorCell.scala:614)
	at akka.actor.ActorCell.supervise(ActorCell.scala:606)
	at akka.actor.ActorCell.invokeAll$1(ActorCell.scala:463)
	at akka.actor.ActorCell.systemInvoke(ActorCell.scala:478)
	at akka.dispatch.Mailbox.processAllSystemMessages(Mailbox.scala:263)
	at akka.dispatch.Mailbox.run(Mailbox.scala:219)
	at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(AbstractDispatcher.scala:397)
```

drewhk:
waaa?

2beaucoup:
Please double check the ticket # in the commit msg.

rkuhn:
Yes, the PR number (with swapped digits) should not be in the commit message, otherwise I’d click the button now.

drewhk:
will merge after changed commit message, or we can take over and do it ourselves.

briantopping:
Hi sorry for the delay, I've been traveling for the last 24 hours.  Can fix in about four hours from now.

briantopping:
BTW, I can totally see where that's a problem, no problem on the request and apologies for the fat fingers..

rkuhn:
Thanks for the quick fix!

briantopping:
Thanks for the merge :)

PR #17678, opened 2015-06-06
=con #17670 Fix potential ReceivePipeline MatchError

### LINKED ISSUE #17670: ReceivePipeline can cause MatchError in decorated actor
If an interceptor using the ReceivePipeline pattern directly calls its inner Receive's apply method with a (possible transformed) message, but the actor that finally receives the message does not handle it, a MatchError can be thrown.

For example, the following:
```scala
import akka.actor._
import akka.contrib.pattern._

implicit val system = ActorSystem("actorSystem")

class SomeActor extends Actor with ReceivePipeline { def receive: Actor.Receive = { case m: String => }  }

trait Interceptor { this: ReceivePipeline =>
    pipelineOuter(inner => { case n: Int => inner(n) })
}

class Decorated extends SomeActor with Interceptor

val decorated = system.actorOf(Props(new Decorated))

decorated ! 1
```

results in:
```
[ERROR] [06/06/2015 11:24:01.525] [actorSystem-akka.actor.default-dispatcher-7] [akka://actorSystem/user/$a] 1 (of class java.lang.Integer)
scala.MatchError: 1 (of class java.lang.Integer)
	at scala.PartialFunction$$anon$1.apply(PartialFunction.scala:253)
	at scala.PartialFunction$$anon$1.apply(PartialFunction.scala:251)
	at $line14.$read$$iw$$iw$$iw$$iw$$iw$$iw$$iw$$iw$$iw$$iw$SomeActor$$anonfun$receive$1.applyOrElse(<console>:21)
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
	at $line15.$read$$iw$$iw$$iw$$iw$$iw$$iw$$iw$$iw$$iw$$iw$Interceptor$$anonfun$1$$anonfun$apply$1.applyOrElse(<console>:22)
	at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:170)
	at akka.actor.Actor$class.aroundReceive(Actor.scala:475)
	at $line14.$read$$iw$$iw$$iw$$iw$$iw$$iw$$iw$$iw$$iw$$iw$SomeActor.akka$contrib$pattern$ReceivePipeline$$super$aroundReceive(<console>:21)
	at akka.contrib.pattern.ReceivePipeline$class.aroundReceive(ReceivePipeline.scala:45)
	at $line14.$read$$iw$$iw$$iw$$iw$$iw$$iw$$iw$$iw$$iw$$iw$SomeActor.aroundReceive(<console>:21)
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:516)
	at akka.actor.ActorCell.invoke(ActorCell.scala:487)
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:256)
	at akka.dispatch.Mailbox.run(Mailbox.scala:223)
	at akka.dispatch.Mailbox.exec(Mailbox.scala:233)
```

I have additional test cases with a fix for this and can send pull request if you like.
Thanks,
Jeremy Stone

jeremystone:
Patch that should hopefully fix #17670

### DIFF
--- a/akka-contrib/docs/receive-pipeline.rst
+++ b/akka-contrib/docs/receive-pipeline.rst
@@ -2,8 +2,8 @@
 
 Receive Pipeline Pattern
 ========================
-Receive Pipeline Pattern lets you define general interceptors for your messages
-and plug any arbitrary amount of them to your Actors.
+The Receive Pipeline Pattern lets you define general interceptors for your messages
+and plug an arbitrary amount of them into your Actors.
 It's useful for defining cross aspects to be applied to all or many of your Actors.
 
 Some Possible Use Cases
@@ -16,33 +16,44 @@ Some Possible Use Cases
 
 Interceptors
 ------------
-So how does an interceptor look? Well, Interceptors are defined by decorator functions
-of type :class:`Receive => Receive`, where it gets the inner :class:`Receive` by parameter and
-returns a new :class:`Receive` with the decoration applied.
-Most of the times your decorators will be defined as a regular :class:`Receive` with cases
-for the messages of your interest and at some point delegate on the inner :class:`Receive`
-you get by parameter. We will talk about ignored and unhandled messages later.
+Multiple interceptors can be added to actors that mixin the :class:`ReceivePipeline` trait. 
+These interceptors internally define layers of decorators around the actor's behavior. The first interceptor
+defines an outer decorator which delegates to a decorator corresponding to the second interceptor and so on, 
+until the last interceptor which defines a decorator for the actor's :class:`Receive`. 
+
+The first or outermost interceptor receives messages sent to the actor. 
+
+For each message received by an interceptor, the interceptor will typically perform some 
+processing based on the message and decide whether 
+or not to pass the received message (or some other message) to the next interceptor.
+
+An :class:`Interceptor` is a type alias for
+:class:`PartialFunction[Any, Delegation]`. The :class:`Any` input is the message
+it receives from the previous interceptor (or, in the case of the first interceptor,
+the message that was sent to the actor).    
+The :class:`Delegation` return type is used to control what (if any) 
+message is passed on to the next interceptor.
 
 A simple example
 ----------------
-We have the simplest :class:`Receive` possible, it just prints any type of message, and we decorate
-it to create an interceptor that increments :class:`Int` messages and delegates the result to the
-inner (printer) :class:`Receive`.
+To pass a transformed message to the actor 
+(or next inner interceptor) an interceptor can return :class:`Inner(newMsg)` where :class:`newMsg` is the transformed message.
 
-.. includecode:: @contribSrc@/src/test/scala/akka/contrib/pattern/ReceivePipelineSpec.scala#interceptor
+The following interceptor shows this. It intercepts :class:`Int` messages, 
+adds one to them and passes on the incremented value to the next interceptor. 
+
+.. includecode:: @contribSrc@/src/test/scala/akka/contrib/pattern/ReceivePipelineSpec.scala#interceptor-sample1
 
 Building the Pipeline
 ---------------------
-So we're done defining decorators which will create the interceptors. Now we will see
-how to plug them into the Actors to construct the pipeline.
-
-To give your Actor the ability to pipeline the receive, you'll need to mixin with the
+To give your Actor the ability to pipeline received messages in this way, you'll need to mixin with the
 :class:`ReceivePipeline` trait. It has two methods for controlling the pipeline, :class:`pipelineOuter`
-and :class:`pipelineInner`, both receiving a decorator function. The first one adds the interceptor at the
+and :class:`pipelineInner`, both receiving an :class:`Interceptor`. 
+The first one adds the interceptor at the
 beginning of the pipeline and the second one adds it at the end, just before the current
 Actor's behavior.
 
-In this example we mixin our Actor with :class:`ReceivePipeline` trait and
+In this example we mixin our Actor with the :class:`ReceivePipeline` trait and
 we add :class:`Increment` and :class:`Double` interceptors with :class:`pipelineInner`.
 They will be applied in this very order.
 
@@ -62,21 +73,23 @@ Let's see it in an example. We have the following model:
 
 .. includecode:: @contribSrc@/src/test/scala/akka/contrib/pattern/ReceivePipelineSpec.scala#mixin-model
 
-And this two interceptors defined each one in its own trait. The first one intercepts any messages having
+and these two interceptors defined, each one in its own trait: 
+
+.. includecode:: @contribSrc@/src/test/scala/akka/contrib/pattern/ReceivePipelineSpec.scala#mixin-interceptors
+
+The first one intercepts any messages having
 an internationalized text and replaces it with the resolved text before resuming with the chain. The second one
 intercepts any message with an author defined and prints it before resuming the chain with the message unchanged.
 But since :class:`I18n` adds the interceptor with :class:`pipelineInner` and :class:`Audit` adds it with
 :class:`pipelineOuter`, the audit will happen before the internationalization.
 
-.. includecode:: @contribSrc@/src/test/scala/akka/contrib/pattern/ReceivePipelineSpec.scala#mixin-interceptors
-
-So if we mixin both interceptors in our Actor, we will see the following output for these example messages.
+So if we mixin both interceptors in our Actor, we will see the following output for these example messages:
 
 .. includecode:: @contribSrc@/src/test/scala/akka/contrib/pattern/ReceivePipelineSpec.scala#mixin-actor
 
 Unhandled Messages
 ------------------
-With all that behaviors chaining occurring on, what happens to unhandled messages? Let me explain it with
+With all that behaviors chaining occurring, what happens to unhandled messages? Let me explain it with
 a simple rule.
 
 .. note::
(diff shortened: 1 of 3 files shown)


akka-contrib/src/main/scala/akka/contrib/pattern/ReceivePipeline.scala
```diff
@@ -38,10 +38,10 @@ trait ReceivePipeline extends Actor {
     val around = aroundCache match {
       case Some((`receive`, cached)) ⇒ cached
       case _ ⇒
-        val zipped = pipeline.foldRight[Receive](receive)((outer, inner) ⇒ outer(inner).orElse[Any, Unit](inner))
+        val zipped = pipeline.foldRight[Receive](receive.orElse(PartialFunction(unhandled)))((outer, inner) ⇒ outer(inner).orElse[Any, Unit](inner))
         aroundCache = Some((receive, zipped))
         zipped
     }
-    super.aroundReceive(around, msg)
+    around(msg)
```
ktoso: pretty sure we must call `super.aroundReceive` somewhere, is this change safe?

jeremystone: Indeed. Have re-jigged to use super.aroundReceive as the zero for the fold to pick up the unhandled msg case

drewhk:
OK TO TEST

akka-contrib/src/main/scala/akka/contrib/pattern/ReceivePipeline.scala
```diff
@@ -38,10 +38,13 @@ trait ReceivePipeline extends Actor {
     val around = aroundCache match {
       case Some((`receive`, cached)) ⇒ cached
       case _ ⇒
-        val zipped = pipeline.foldRight[Receive](receive)((outer, inner) ⇒ outer(inner).orElse[Any, Unit](inner))
+        val zipped = pipeline.foldRight[Receive] {
+          case msg => super.aroundReceive(receive, msg)
```
ktoso: I'm not sure if this is the exact semantics we need here, is it ok to call `super.aroundReceive` multiple times? Seems a bit weird.

ktoso:
Would be nice @ktonga if you could have a look in this PR as well, since you authored the initial impl - thanks a lot in advance!

akka-contrib/src/main/scala/akka/contrib/pattern/ReceivePipeline.scala
ktonga: You're right @ktoso, in a mixin context ```super.aroundReceive``` might not be just the default impl, and if any of the interceptors intentionally discard messages ```super.aroundReceive``` won't be called.

We should still use the original ```receive``` as zero but with fallback in ```unhandled```. ```receive.orElse(PartialFunction(unhandled))```

I saw that this was your first approach for fixing it @jeremystone. What was wrong with it that you changed it?

jeremystone: When @ktoso suggested that `super.aroundReceive` should be called somewhere, I agreed: what right has `ReceivePipeline` to enforce the specific use of `unhandled` as the fallback when the base actor might deliberately have overridden `aroundReceive` with some other behaviour. 

Placing the `super.aroundReceive` call in the middle of the fold makes the fallback less invasive. But now as you point out, @ktonga, it might not get called at all. Clearly, calling `super.aroundReceive` at the end of the method _as well_ will not do as then it will then often be called twice.

My gut feeling is that it is almost as if `ReceivePipeline` is ceding too much power over the flow control to the interceptors (allowing them to call each other directly). An alternative would be to have each interceptor as a `PartialFunction[Any, InterceptorResult]` with something like:
```
sealed trait InterceptorResult
case class Inner(transformedMsg: Any) extends InterceptorResult
case object HandledCompletely extends InterceptorResult
```
The ReceivePipeline would construct a `Receive` from these interceptors (so that _it_ would be responsible for the message flow between interceptors) and just pass this to `super.aroundReceive` along with the original message at the end (as per the original implementation). 

Of course this would be an API changer.

ktonga: I like the idea @jeremystone, seems to be a definitive solution for the flow control problem. I don't think it is a problem to change the API since it is in the contrib module.

jeremystone: OK. I'd be happy to knock something up for this - and see how it looks.

ktoso:
So if I understand the discussion and code above correctly this only needs a rebase and is ready to be merged, right?

jeremystone:
Just got round to pushing change addressing flow control as discussed, so
this needs review. Is a minor API change so need docs fixing up if it's OK.

akka-contrib/src/main/scala/akka/contrib/pattern/ReceivePipeline.scala
```diff
@@ -2,46 +2,113 @@ package akka.contrib.pattern
...
+      }
+    }
+
+    toReceive(zipped)
+  }
+
+  private def toReceive(handler: Handler) = new Receive {
+    // Cache the result locally to avoid evaluating potentially 
+    // side-effecting code twice in isDefinedAt/apply...
+    var resultCache = Option.empty[(Any, HandlerResult)]
```
ktonga: ```resultCache``` is never cleaned, it could unnecessarily prevent the message to be GCed. Maybe ```apply``` should empty it.

jeremystone: Good point. Will make that change.

jeremystone:
Before we go any further with this, one thing that is missing (that could be achieved before) is the ability to do something within an interceptor _after_ the message has been processed by the actor's receive (e.g. for the documented use case of timing message processing). 

Looks like this could be provided by altering the `Inner` `InterceptorResult` to something like:
```
  case class Inner(transformedMsg: Any, after: Unit => Unit) extends InterceptorResult
```
along with some appropriate convenience factory methods providing defaults.

Then inside the `combinedAdvisor` ...
```scala
      outerInterceptor.andThen {
        case Inner(msg, after) =>
          val result = innerHandler(msg)
          after()
          result
        case HandledCompletely => Done
      }
```
Any thoughts on this?

ktoso:
needs a `try innerHandler() finally after()` but looks good.

ktoso:
AFAICS this is ready right? If so, could you please squash it into one commit titling it `=con #17670 Fix potential ReceivePipeline MatchError`? Thanks @jeremystone !

jeremystone:
Yes...so long as everyone happy with alterations to API. Will not be in a
position to squash the commit myself till the weekend.

There's also some documentation to do. Do you want this as part of the same
commit or as separate issue and pr?

ktoso:
Docs in the same commit would be excellent :)

akka-contrib/docs/receive-pipeline.rst
```diff
@@ -16,33 +16,46 @@ Some Possible Use Cases
...
 Interceptors
 ------------
-So how does an interceptor look? Well, Interceptors are defined by decorator functions
-of type :class:`Receive => Receive`, where it gets the inner :class:`Receive` by parameter and
-returns a new :class:`Receive` with the decoration applied.
-Most of the times your decorators will be defined as a regular :class:`Receive` with cases
-for the messages of your interest and at some point delegate on the inner :class:`Receive`
-you get by parameter. We will talk about ignored and unhandled messages later.
+Multiple interceptors can be added to actors that mixin the :class:`ReceivePipeline` trait. 
+These interceptors can be thought of as layers: each outer interceptor advising the
```
ktoso: I'm not sure we should use the `advice` word, the API does not expose any "advice" in methods/types.
I know it's AOP wording and am familiar with it, but would like to avoid it leaking into the docs, since the API does not use such words.

akka-contrib/docs/receive-pipeline.rst
```diff
@@ -16,33 +16,46 @@ Some Possible Use Cases
...
+next interceptor in, and so on, until the last or innermost interceptor which advises the 
+actor's :class:`Receive`. 
+
+The first or outermost interceptor receives messages sent to the actor. 
+
+For each message received by an interceptor, the interceptor will typically perform some 
+processing based on the message and decide whether 
+or not to pass the received message (or some other message) to the next interceptor.
+
+So what does an interceptor look like?
```
ktoso: Maybe remove this sentence, does not fit the flow of the doc now with extended docs.

akka-contrib/docs/receive-pipeline.rst
```diff
@@ -84,8 +99,24 @@ a simple rule.
...
 
-But some times it is desired for interceptors to break the chain. You can do it by explicitly ignoring
-the messages with empty cases or just not calling the inner interceptor received by parameter.
+But sometimes it is desired for interceptors to break the chain. You can do it by explicitly indicating 
+that the message has been completely handled by the interceptor by returning the 
+:class:`HandledCompletely` result.
 
 .. includecode:: @contribSrc@/src/test/scala/akka/contrib/pattern/ReceivePipelineSpec.scala#unhandled
 
+Around advice
```
ktoso: Here too; We're not using the word Advice anywhere in the APIs, so let's not add it here to avoid confusing users

akka-contrib/src/main/scala/akka/contrib/pattern/ReceivePipeline.scala
```diff
@@ -2,46 +2,115 @@ package akka.contrib.pattern
 
 import akka.actor.Actor
 
+object ReceivePipeline {
+  /**
+   * Result returned by an interceptor PF.
+   */
+  sealed trait InterceptorResult
```
ktoso: `Interception`?

akka-contrib/docs/receive-pipeline.rst
```diff
@@ -16,33 +16,46 @@ Some Possible Use Cases
...
+or not to pass the received message (or some other message) to the next interceptor.
+
+So what does an interceptor look like? 
+
+An :class:`Interceptor` is a type alias for
+:class:`PartialFunction[Any, InterceptorResult]`. The :class:`Any` input is the message
+it receives from the previous interceptor (or, in the case of the first interceptor,
+the message that was sent to the actor).    
+The :class:`InterceptorResult` return type is used to control what (if any) 
+message is passed on to the next interceptor.
```
ktoso: very good docs improvements in general!

akka-contrib/docs/receive-pipeline.rst
jeremystone: OK. Could use decorate/decoration/delegate terminology as before, though the interceptor at the API level is now a more functional "description of the advice" than it was - so not sure this is right either unless this fact is made clear in the docs.
I will try to come up with an alternative.

akka-contrib/docs/receive-pipeline.rst
jeremystone: OK

akka-contrib/docs/receive-pipeline.rst
jeremystone: Before I commit anything, how about the following which uses the decorator/delegate terminology and renames InterceptorResult to Delegation.

(Internally the code uses advice/advisor terminology - albeit privately - so will rename to line up and lessen any confusion here, too.)

Interceptors
------------
Multiple interceptors can be added to actors that mixin the `ReceivePipeline` trait. 
These interceptors internally define layers of decorators around the actor's behavior. The first interceptor
defines an outer decorator which delegates to a decorator corresponding to the second interceptor and so on, 
until the last interceptor which defines a decorator for the actor's `Receive`. 

[etc etc as before...]

ktoso: Hi, and please excuse the delayed response!
Yes this sounds good to me, thanks!

ktoso:
This is LGTM AFAICS, though needs to be squashed into one commit with title `	=con #17670 Fix potential ReceivePipeline MatchError`

patriknw:
LGTM
@ktoso can we merge this?

ktoso:
Yes I think so, been LGTMed a lot ;-)

jeremystone:
Cool!

PR #17684, opened 2015-06-08
=htc #17683 drop already consumed input in `continue` (+ ByteString compaction)

### LINKED ISSUE #17683: HttpMessageParser aggregates input in `continue`
This leads to a temporary memory leak and increasing parsing times for long messages where the `input` ByteString starts to consist of more and more elements leading to bad performance because of linear scanning in `ByteString.slice`.

/cc @sirthias

jrudolph:
Fixes #17683.

/cc @sirthias

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
@@ -252,7 +252,9 @@ private[http] abstract class HttpMessageParser[Output >: MessageOutput <: Parser
   def continue(input: ByteString, offset: Int)(next: (ByteString, Int) ⇒ StateResult): StateResult = {
     state =
       math.signum(offset - input.length) match {
-        case -1 ⇒ more ⇒ next(input ++ more, offset)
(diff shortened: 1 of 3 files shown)


jrudolph:
TlsSpec

viktorklang:
What's the performance impact on the compaction?

drewhk:
The cost of compacting is copying data. The cost of not compacting is accidentally retaining lots of ByteString snippets if a ByteString buffer is consumed in a "sliding" fashion. Theoretically you can retain unbounded number of ByteStrings that you don't even use. Back then when I was toying with streaming byte-oriented parsers I put a threshold of consumed bytes (how many bytes the buffer has been "shifted/sliding" through) after I compacted to amortize the cost.

jrudolph:
@drewhk yes, exactly. Retaining bytes was a real issue in spray at some point. For this reason we then adopted the policy in spray to compact ByteStrings that are likely to retain parts of byte buffers at the API boundary where the ByteString are released out of the control of the implementation layer.

There's a cost involved, no question, but we are already so far from a zero-copy infrastructure that being predictable seems like the more valuable goal in this case.

drewhk:
LGTM

Issue #15410, opened 2014-06-17
Remove name from props factories for cluster singleton manager and proxy


hseeberger:
Currently it's necessary to provide a name for the singleton actor when creating the singleton manager actor instance. Accordingly it's necessary to provide the full path including this name when creating the singleton proxy actor instance. 

This is error-prone, because one has to provide both the name of the singleton manager and the singleton actor. Hopefully it's possible to get rid of the explicit actor name on both ends.

PR #17696, opened 2015-06-11
!clt #15410 Change path in ClusterSingletonProxy.props

patriknw:
- to avoid giving the name of both singleton manager and singleton instance

### DIFF
--- a/akka-cluster-tools/src/main/resources/reference.conf
+++ b/akka-cluster-tools/src/main/resources/reference.conf
@@ -136,6 +136,9 @@ akka.cluster.singleton {
 }
 
 akka.cluster.singleton-proxy {
+  # The actor name of the singleton actor that is started by the ClusterSingletonManager
+  singleton-name = ${akka.cluster.singleton.singleton-name}
+  
   # The role of the cluster nodes where the singleton can be deployed. 
   # If the role is not specified then any node will do.
   role = ""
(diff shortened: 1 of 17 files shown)


patriknw:
Refs #15410

akka-cluster-tools/src/main/resources/reference.conf
```diff
@@ -138,6 +138,9 @@ akka.cluster.singleton {
 }
 
 akka.cluster.singleton-proxy {
+  # The actor name of the singleton actor that is started by the ClusterSingletonManager
+  singleton-name = ${akka.cluster.singleton.singleton-name}
```
rkuhn: This feels somehow a little off (but might be okay, depends on the explanation): The reason was to remove duplication so that the configuration is more “in one place” than before, therefore I kind-of expected this configuration to be taken directly from the `akka.cluster.singleton` section, since now the user need to override two config keys with identical contents (`reference.conf` is resolved fully before being used as fallback, so this alias does not expand to users’ overridden values).

rkuhn:
LGTM apart from that one comment.

akka-cluster-tools/src/main/resources/reference.conf
patriknw: There is almost never a need to change this name, because it is qualified by the name of singleton manager, which is created by the user with `context.actorOf(..., name = )`, but since it was configurable before and someone might rely on using specific path names I kept it configurable.

Another thing, when you create `ClusterSingletonProxySettings` from a config it is a `Config` with the same layout as defined in reference.conf `akka.cluster.singleton-proxy`. Therefore I wanted to define all settings of the proxy in that section.

What would be possible is to grab the singleton-name from `akka.cluster.singleton.singleton-name` when creating `ClusterSingletonProxySettings` with `def apply(system: ActorSystem): ClusterSingletonProxySettings`

rkuhn: Thanks for the details, makes sense to keep it as is then.

patriknw:

patriknw:
ran tests again locally, can't wait for jenkins (which is out of order right now)

Issue #15110, opened 2014-05-01
ClusterSingletonProxy should optionally drop messages instead of stashing


hseeberger:
Currently I don't see any way to disable the stashing of messages when the cluster singleton isn't available. There should be an option to disable stashing, i.e. drop messages.

patriknw:
It should probably not use stash at all, to make the buffering more explicit with a buffer size parameter. Drop oldest messages also makes more sense.

patriknw:
Similar for ClusterClient, I think

PR #17697, opened 2015-06-11
!clt #15110 Use buffer instead of stash in singleton proxy and cluster client

patriknw:


patriknw:
Refs #15110

.gitignore
```diff
@@ -27,6 +27,7 @@
 .scalastyle
 .settings
 .cache*
+.tmpBin
```
rkuhn: Is this a temporal bin in which I can collect all those short time snippets that are not properly used to recycle them at the end of the day? That would be nice …

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -228,14 +238,19 @@ class ClusterClient(settings: ClusterClientSettings) extends Actor with Stash wi
     case ActorIdentity(_, Some(receptionist)) ⇒
       log.info("Connected to [{}]", receptionist.path)
       scheduleRefreshContactsTick(refreshContactsInterval)
-      unstashAll()
+      sendBuffered(receptionist)
```
rkuhn: This is another DoS attack waiting to happen …

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -270,6 +285,22 @@ class ClusterClient(settings: ClusterClientSettings) extends Actor with Stash wi
     else if (contacts.size == 1) (initialContactsSel ++ contacts) foreach { _ ! GetContacts }
     else contacts foreach { _ ! GetContacts }
   }
+
+  def buffer(msg: Any): Unit =
+    if (buffer.size == settings.bufferSize) {
+      log.debug("Receptionist not available, buffer is full, dropping message type [{}]", msg.getClass.getName)
```
rkuhn: here it might make more sense to drop head instead of tail

rkuhn:
LGTM apart from commented tweaks.

.gitignore
patriknw: I'm not sure what comes from the IDE and what is from sbt, but I think this one is from the IDE. I talked to one of the IDE developers at ScalaDays and said that it would be nice if all those "temporary" files could be located in **one** directory.

rkuhn: Oh, so this is not a time-reuse device … what a bummer :-( (but thanks for the explanation ;-) )

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
patriknw: You mean sending too many at the same time?

rkuhn: Yes, we invariably end up building in flow control later.

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
patriknw: ok, then I guess I need to use `java.util.LinkedList` instead of Vector.

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
patriknw: Yes, full flow control would be nice, but that is not in scope now, in my opinion. The buffer is intended to be small (we could set a max cap?).

rkuhn: agreed upon the scope, please add a ticket

.gitignore
patriknw: :-) (now I got the joke)

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
rkuhn: Yes, that would be better.

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
patriknw: added #17720

patriknw:
PLS BUILD

patriknw:
@rkuhn I have changed so that it drops old instead of new, see 02a23e1 and a33ef48

patriknw:
Failed to create a temp file on /localhome/jenkinsakka/workspace/pr-validator-per-commit-jenkins

rkuhn:
LGTM

patriknw:
PLS BUILD

akka-cluster-tools/src/main/scala/akka/cluster/singleton/ClusterSingletonProxy.scala
```diff
@@ -219,16 +234,41 @@ class ClusterSingletonProxy(singletonPathString: String, settings: ClusterSingle
...
+          buffer(msg)
       }
   }
+
+  def buffer(msg: Any): Unit =
+    if (settings.bufferSize == 0)
+      log.debug("Singleton not available and buffering is disabled, dropping message [{}]", msg.getClass.getName)
+    else if (buffer.size == settings.bufferSize) {
+      val (m, _) = buffer.removeFirst()
+      log.debug("Singleton not available, buffer is full, dropping first message [{}]", m.getClass.getName)
```
anderius: This log and the log above is very useful. Would it be possible to have it on info? At least the first time it happens (and log again if it happens again after the buffer was not full)?

If I put the logger on debug, I will probably get a lot of messages all the time?

akka-cluster-tools/src/main/scala/akka/cluster/singleton/ClusterSingletonProxy.scala
```diff
@@ -219,16 +234,41 @@ class ClusterSingletonProxy(singletonPathString: String, settings: ClusterSingle
...
+      buffer.addLast((msg, sender()))
+    } else {
+      log.debug("Singleton not available, buffering message type [{}]", msg.getClass.getName)
+      buffer.addLast((msg, sender()))
+    }
+
+  def sendBuffered(): Unit = {
+    log.debug("Sending buffered messages to current singleton instance")
```
anderius: This one, for instance, is not always interesting..

anderius:
This looks very good! I tried to implement something similar, but failed because stashing re-ordered the messages in the queue.

My scenario (that this will solve) is that I have GPS messages coming in from a number of devices, and if the system for some reason slows down, I will rather toss away old messages than new ones.

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -270,6 +297,26 @@ class ClusterClient(settings: ClusterClientSettings) extends Actor with Stash wi
     else if (contacts.size == 1) (initialContactsSel ++ contacts) foreach { _ ! GetContacts }
     else contacts foreach { _ ! GetContacts }
   }
+
+  def buffer(msg: Any): Unit =
```
anderius: I almost don't read Scala, but will it be possible (in Java) to override the method 'buffer' and 'sendBuffer'?

What i am trying to achieve is the following: I have GPS updates from many devices, and instead of throwing away the oldest, I could throw away older messages from the same device. That is, replacing 'buffer' with a map mapping from device ID to a list of messages (or the last one, if I only care about that one).

patriknw: Opening up such classes for extension makes it very difficult for us to change things (also in a binary compatible way). I actually made this class `final` for this reason.

I could do an exception for the ClusterClient, but then I will make everything private/final except `def buffer` and `def sendBuffered`.

and yes, then you can override them in Java

anderius: I agree with your points... Maybe it is safer to have the class final (and easier to change). It is not a 'must-have' for my team (dropping the oldest message as you do now will help us a lot).

patriknw: good, I appreciate that we keep it simple (where possible)

regarding buffering in sharding; it actually drops new messages when the buffer is full, perhaps better to change to dropping old?

anderius: For my use-cases that would typically be better. But some way of plugging into that would be nice.

On the other hand, if you have retry implemented on the sender, the order would be less effected if you continued to drop the new messages. (there is not a strict ordering guarantee anyway, but dropping the oldest would cause them to be resent, and the ordering would be even more off)

PR #17704, opened 2015-06-12
=rem #17554 Improve flow control of system message delivery (for validation)

patriknw:
When watching many (5000) actors at the same time the
following problems were found:
- first send of a sys msg is sent without any flow control
  => limit the number of outstanding sys msg by using
     the buffer to send them later (ordinary resend)
- when msg cannot be written sys msg is dropped (relying on resend),
  but that cause message re-ordering and negative acknowledgment,
  which is very costly
  => buffer the sys msg on write failure
  => minor optimization of AckedReceiveBuffer

I also made the resend-limit configurable.

(cherry picked from commit ecfc271e9a9d7efcf76945632d89c78740291cc6)

### DIFF
--- a/akka-remote/src/main/scala/akka/remote/Endpoint.scala
+++ b/akka-remote/src/main/scala/akka/remote/Endpoint.scala
@@ -200,13 +200,6 @@ private[remote] class ReliableDeliverySupervisor(
   val autoResendTimer = context.system.scheduler.schedule(
     settings.SysResendTimeout, settings.SysResendTimeout, self, AttemptSysMsgRedelivery)
 
-  // Aim for a maximum of 100 resend/s = 0.1 resend/ms
-  val maxResendRate = 100.0
-  val resendLimit =
-    math.min(
-      1000,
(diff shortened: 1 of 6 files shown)


akka-remote/src/main/scala/akka/remote/Endpoint.scala
```diff
@@ -716,7 +710,7 @@ private[remote] class EndpointWriter(
   val writing: Receive = {
     case s: Send ⇒
       if (!writeSend(s)) {
-        if (s.seqOpt.isEmpty) enqueueInBuffer(s)
+        enqueueInBuffer(s)
```
rkuhn: does this mean that non-acked messages are now also taking up buffer space?

rkuhn: ah, no, it is the other way around, sorry for the noise

rkuhn:
LGTM

patriknw:
The build failure was due to mima, changed internals. I have added filter.

ktoso:
LGTM

patriknw:
Refs #17554

Issue #16259, opened 2014-11-10
Automate API test coverage check for Java & Scala DSL


patriknw:

rkuhn:
We also need to re-enable the DSL consistency specs.

ktoso:
Note:
- we have consistency specs enabled again
- keeping around since we want full coverage for javadsl, which I don't think we have at the current point (at least touching all APIs).

drewhk:
This is too flaky to do manually, we probably need a coverage tool configured to only look for the usage of API methods.

drewhk:
Is this ticket really needed anymore? If someone has a concrete proposal, please open a separate ticket. I mean we can have a ticket called "Improve Akka Streams" and keep it open forever :)

PR #17710, opened 2015-06-13
+str #17226 add dropWhile and takeWhile

agolubev:
this is proof of concept. Please review if approach is good enough to be finished and merged in.
I tried another approach with using filter stage and closure for filter function, did not find same usage anywhere else though.


akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -36,6 +36,40 @@ private[akka] final case class Filter[T](p: T ⇒ Boolean, decider: Supervision.
...
+private[akka] final case class TakeWhile[T](p: T ⇒ Boolean, decider: Supervision.Decider) extends PushStage[T, T] {
+  var taking = true
+
+  override def onPush(elem: T, ctx: Context[T]): SyncDirective =
+    if (taking && p(elem)) {
+      ctx.push(elem)
+    } else {
+      taking = false
+      ctx.pull
+    }
```
ktoso: Should have similar semantics to `take(n)`, so it should `ctx.pushAndFinish(elem)` instead of keeping to pull. once condition becomes false.

(Remove taking var as well - it is not needed AFAICS)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -36,6 +36,40 @@ private[akka] final case class Filter[T](p: T ⇒ Boolean, decider: Supervision.
...
+private[akka] final case class DropWhile[T](p: T ⇒ Boolean, decider: Supervision.Decider) extends PushStage[T, T] {
+  var dropping = true
+
+  override def onPush(elem: T, ctx: Context[T]): SyncDirective =
+    if (dropping && p(elem)) {
+      ctx.pull
+    } else {
+      dropping = false
+      ctx.push(elem)
+    }
```
ktoso: Looks ok, question is if we want to inverse the if or not ("dropping for a long time is the more often case or just dropping for a little bit is?")

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -457,6 +457,10 @@ trait FlowOps[+Out, +Mat] {
    */
   def filter(p: Out ⇒ Boolean): Repr[Out, Mat] = andThen(Filter(p.asInstanceOf[Any ⇒ Boolean]))
 
+  def takeWhile(p: Out ⇒ Boolean): Repr[Out, Mat] = andThen(TakeWhile(p.asInstanceOf[Any ⇒ Boolean]))
+
+  def dropWhile(p: Out ⇒ Boolean): Repr[Out, Mat] = andThen(DropWhile(p.asInstanceOf[Any ⇒ Boolean]))
+
```
ktoso: OK, let's add docs and Java API (remember that we also have tables of ops in the docs).

ktoso:
OK TO TEST

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
agolubev: I inversed it. Assuming streams are rather taking prevalent

drewhk:
Very nice! You are getting the hang of Stages :)

Please just add a Java smoke test, too. One test case is enough, since logic is shared, but it helps us to see when we break Java API.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowDropWhileSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+      Source.empty[Int].dropWhile(_ < 2).runWith(Sink.publisher).subscribe(subscriber)
+      val s = subscriber.expectSubscription()
+      s.request(1)
+      subscriber.expectComplete()
+    }
+
+    "continue if error" in assertAllStagesStopped {
+      val testException = new Exception("test") with NoStackTrace
+      val subscriber = TestSubscriber.manualProbe[Int]()
+      Source(1 to 4).dropWhile(a ⇒ if (a == 3) throw testException else true).withAttributes(supervisionStrategy(resumingDecider))
```
rkuhn: perhaps change `true` into `a < 3` to make it slightly more interesting ;-)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowDropWhileSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  val settings = ActorFlowMaterializerSettings(system)
+
+  implicit val materializer = ActorFlowMaterializer(settings)
+
+  "A DropWhile" must {
+
+    "drop while predicate is true" in assertAllStagesStopped {
+      val subscriber = TestSubscriber.manualProbe[Int]()
+      Source(1 to 4).dropWhile(_ < 3).runWith(Sink.publisher).subscribe(subscriber)
```
rkuhn: You could use `akka.stream.testkit.scaladsl.TestSink.probe` directly in the `.runWith` which gives a nice and fluent DSL for the following assertions.

rkuhn:
Thanks for the contribution, looking good!

drewhk:
Can you please squash the commits into one? Then I will merge this.

ktoso:
:warning: Missing Java Api Detected :warning: 
Please duplicate the signatures defined on `javadsl.Flow` onto `javadsl.Source`

ktoso:
Good reminder that we're missing the consistency spec currently: https://github.com/akka/akka/issues/16259

drewhk:
Good catch @ktoso!

Please add Java API and a couple of smoke-tests for them.

agolubev:
added methods to java.Source, few test and squashed all commits

drewhk:
LGTM

ktoso:
refs https://github.com/akka/akka/issues/17226

Issue #17638, opened 2015-06-02
Add preStart, postStop to FlexiMerge/Route


drewhk:
This is needed to solve the actor leak in the WS path if connection is aborted.

drewhk:
@jrudolph, the feature is now available, so you can fix the actor leak using preStart/postStop on the FlexiRoute (or FlexiMerge)

jrudolph:
Thanks, good to know. I will check it later.

PR #17724, opened 2015-06-15
+str #17638: Add preStart/postStop to FlexiRoute and FlexiMerge

drewhk:


drewhk:
PLS BUILD

ktoso:
LGTM

akka-stream/src/main/scala/akka/stream/impl/FlexiRouteImpl.scala
```diff
@@ -31,6 +31,16 @@ private[akka] class FlexiRouteImpl[T, S <: Shape](_settings: ActorFlowMaterializ
   // needed to ensure that at most one element is emitted from onInput
   private val emitted = Array.ofDim[Boolean](outputCount)
 
+  override def preStart(): Unit = {
+    super.preStart()
```
ktoso: was tempted to wrap in try/finally yet currently we know it's the Actor#preStart so not needed I guess

drewhk:
Fixes #17638

akka-stream/src/main/scala/akka/stream/impl/FlexiRouteImpl.scala
drewhk: if the super throws then we should not attempt the stage preStart IMHO. (i.e. we did not start the stage)

ktoso: that reasoning also makes sense, good :)

Issue #17661, opened 2015-06-05
OperationAttributes name is confusing


ktoso:
*Back in my days...* (:wink:) we used to call stages "ops", from that time we have `OperationAttributes`.
We do not really use the word operation anywhere in akka streams - it's all "stages/junctions/flows" now, 
so OperationAttributes sounds a bit off.

We should rename before 1.0.

// cc @drewhk

drewhk:
I think it was renamed at one point from Attributes. I think that was the cleanest so far.

PR #17727, opened 2015-06-15
!str #17661: New naming: Materializer, Attributes

drewhk:


akka-bench-jmh/src/main/scala/akka/stream/FlowMapBenchmark.scala
```diff
@@ -62,7 +62,7 @@ class FlowMapBenchmark {
     val settings = ActorFlowMaterializerSettings(system)
```
ktoso: `ActorFlowMaterializerSettings` should be `ActorMaterializerSettings`, since now it is `ActorMaterializer`

ktoso:
For reference, what I think we've settled with recently:
`Materializer`, `ActorMaterializer`, `MaterializerSettings`, `ActorMaterializerSettings`, `akka.stream.Attributes` AFAIR

rkuhn:
Yes, indeed, and `ActorAttributes` as well.

hepin1989:
naming is always the hardest thing.

drewhk:
Another PR is coming, closing this one.

PR #17731, opened 2015-06-15
=str code comment the Inlet/InPort dichotomy

rkuhn:


akka-stream/src/main/scala/akka/stream/Shape.scala
```diff
@@ -6,10 +6,28 @@ package akka.stream
 import scala.collection.immutable
 import scala.collection.JavaConverters._
 
+/**
+ * An input port of a StreamLayout.Module. This type logically belongs
+ * into the impl package but must live here due to how `sealed` works.
```
patriknw: can it be `private [akka]`?

rkuhn: No, since we use it in public signatures, in particular in Flexi*.

patriknw: then it does not logically belong to the impl package, and I think this is more confusing than helpful for the user. Perhaps the reason should be in a non-scaladoc comment instead?

ktoso: The comment is a bit confusing TBH - maybe explain why exactly, or make it `//` comments?

drewhk: INTERNAL API is what you are looking for :)

drewhk:
LGTM after clearly marking the port classes as INTERNAL API

akka-stream/src/main/scala/akka/stream/Shape.scala
patriknw: but he just said that they are not internal

drewhk: Ah, ok. But then the comment is wrong about belonging to impl.

patriknw: yes, that was my point with the question :-)

rkuhn: Yes, TBH I also thought they were internal until I checked … thanks for the nudge, @patriknw! Will rework.

drewhk: Alternative is to really move it into impl, to the StreamLayout file, and mark it as INTERNAL API. "sealed" only works in scala anyway.

rkuhn: Well, we use it in public API, users will use this type, so it is not `INTERNAL API` in any sense.

drewhk: maybe rename to Untyped then?

drewhk:
ping

drewhk:
ping ping

rkuhn:
pong pong

drewhk:
ready for merge?

rkuhn:
No :-( but I don’t want to rename them, so I’ll need to make the comment actually fit reality—unless you really thing that UntypedInlet is better for everyone.

drewhk:
I don't think there has been any issue on the ML so far about this, so the names can remain.

Issue #15644, opened 2014-08-08
Make PersistentActor.persist iterate over a Traversable type instead of immutable.Seq 


Fristi:
Would it be possible to let `PersistentActor.persist` accept a traversable type instead of a immutable.Seq? I stumbled upon writing this today;

`persist(u.items.map(CharacterScanList.Added))(evs => state = evs.foldLeft(state)(_ apply _))`

Where `evs` is a `Seq[CharacterScanList.Added]`. The `u.items` property appears to be a list and seems not be converted as `collection.immutable.Seq[_]` so it is using the overload:

`def persist[A](event: A)(handler: A ⇒ Unit): Unit`

So folding seemed like the right solution. But actually it isn't the right way? It will persist a Seq[CharacterScanList.Added] now as event while it should persist a batch of CharacterScanList.Added events.

rkuhn:
Thanks for pointing this out: we need to remove the overloaded method (i.e. introduce `persistAll`) and then I think we can permit `Seq[T]` even though that is mutable (because we only traverse it once within the method).

sibex:
Yesterday I've spend almost an hour to figure out why the scala compiler demands an event handler of type `Seq[Event] => Unit`. It's because the default `scala.Seq` is `scala.collection.Seq` and not `scala.collection.immutable.Seq`. Actually a no-brainer. But please remove/rename the overloaded method.

There is another issue. The API documentation says:
 > Asynchronously persists `events` in specified order. This is equivalent to calling
    `persist[A](event: A)(handler: A => Unit)` multiple times with the same `handler`,
    except that `events` are persisted atomically with this method.

This is misleading. The `events` are **not** persisted atomically as a block in an ACID way. The current implementation is:
```scala
  final def persist[A](events: immutable.Seq[A])(handler: A ⇒ Unit): Unit =
    events.foreach(persist(_)(handler))
```
Other persistent actors can write to the log between the events. Or for some of the events `persist` may fail. If the new `persistAll` can somehow achieve the promised atomicity then it should take an _all or nothing_ atomic event handler of type `immutable.Seq[A] => Unit`.

patriknw:
@sibex Thanks for feedback.

ktoso:
Actions here:
- remove `persist(Seq)`
- add `persistBatch(Seq)`
- atomicity guarantees are to be defined as follows:
  - persist - each write is atomic
  - perstiBatch the batch is atomic

Issue #15377, opened 2014-06-09
Rethink PersistentBatch - do we need to force atomic batch writes onto plugins?


ktoso:
Based on a discussion from: https://github.com/akka/akka/pull/15339/files#r13540786

This task is about rethinking, and possibly loosening up requirements given to journals, about writing a batch as atomic "blob" of data. More details below.

michaelom:
After reading this thread, I still wonder if you should relax the requirement of atomic batch writes, or at least clarify the what it means in the context of event sourcing. The word “atomic” is very prone to misinterpretation, depending if you approach the problem on database-level or on API-level.

Let’s take a look at Greg’s “1-command, N-events” scenario. I would argue you can archive all this without “atomic” writes. The requirements from an API-perspective are:

* a) PersistentActor: during recovery, replay the persistent batch (or up to a sequence number) and include events of an "event-batch" if and only if all events of the persistent batch have been written successfully to the underlying storage implementation. 

* b) PersistentView: same as PersistentActor, as PersistentView is effectively always in recovery mode.

* c) Deletion: it must be possible to mark individual events of an event-batch as deleted, without breaking a) and b)

For most data stores an atomic write (a transaction) is the most natural way to achieve a), b), and c).

I imagine a document based-storage (such as MongoDB) could achieve all of this with “marker-documents” at the start and the end of the batch. Not sure if this is a feasible approach, but at least I would not have to squash a batch of events into a single document in order to satisfy atomicity. From an API level it is conceived as atomic (all or nothing) but it does not require atomic transactions from a database point of view.

gregoryyoung:
A batch has no meaning in reads only in terms of an atomic write. There
should be no knowledge that something was in a batch when it was written.

On Friday, January 23, 2015, michaelom <notifications@github.com> wrote:

> After reading this thread, I still wonder if you should relax the
> requirement of atomic batch writes, or at least clarify the what it means
> in the context of event sourcing. The word “atomic” is very prone to
> misinterpretation, depending if you approach the problem on database-level
> or on API-level.
>
> Let’s take a look at Greg’s “1-command, N-events” scenario. I would argue
> you can archive all this without “atomic” writes. The requirements from an
> API-perspective are:
>
>    -
>
>    a) PersistentActor: during recovery, replay the persistent batch (or
>    up to a sequence number) and include events of an "event-batch" if and only
>    if all events of the persistent batch have been written successfully to the
>    underlying storage implementation.
>    -
>
[...]

PR #17742, opened 2015-06-16
!per #15644 Rename persist(seq) to persistAll

patriknw:


akka-persistence/src/main/scala/akka/persistence/PersistentActor.scala
```diff
@@ -200,8 +204,8 @@ abstract class UntypedPersistentActor extends UntypedActor with Eventsourced wit
    * @param events events to be persisted
    * @param handler handler for each persisted `events`
    */
-  final def persistAsync[A](events: JIterable[A])(handler: A ⇒ Unit): Unit =
```
patriknw: note that this signature was broken, and therefore I don't care about deprecating it

ktoso: Oh, good catch!

akka-docs/rst/project/migration-guide-2.3.x-2.4.x.rst
```diff
@@ -176,6 +176,13 @@ In case you want to preserve the old behavior of providing the actor's path as t
 implement it yourself either as a helper trait or simply by overriding ``persistenceId`` as follows::
 
     override def persistenceId = self.path.toStringWithoutAddress
+    
+persist sequence of events
```
ktoso: upper case P ;-)

ktoso:
Refs #15644

ktoso:
LGTM, and after we do https://github.com/akka/akka/issues/15377 those might become the "atomic batch", so we'd add comments to them.

patriknw:
the failure was in ClusterShardingLeavingSpec (known issue)

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -254,8 +254,12 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
    * @param events events to be persisted
    * @param handler handler for each persisted `events`
    */
+  final def persistAll[A](events: immutable.Seq[A])(handler: A ⇒ Unit): Unit =
+    events.foreach(persist(_)(handler)) // TODO the atomic part should be handled by issue #15377
+
+  @deprecated("use persistAll instead", "2.4")
   final def persist[A](events: immutable.Seq[A])(handler: A ⇒ Unit): Unit =
-    events.foreach(persist(_)(handler))
+    persistAll(events)(handler)
```
ktoso: I'm tempted to propose immediate deletion instead of deprecation, as it's existence in deprecated status still causes confusion: `persist(List(2))`. If possible I'd vote for removing right away, so the maximum number of breakages come in one version update - M2

ktoso:
LGTM, would like to remove the deprecated method right away though, since there's already API changes let's bundle them up in M2

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
patriknw: I was thinking along the same lines first, but finding the places where it is used is very hard. If we remove it and the user is not changing callers it will break the application silently in completely wrong way, i.e. the Seq will be stored instead of the events.

ktoso: That's very true - I hope most users IDEs can properly figure out the deprecated one is used and highlight it in the editor. So the plan would be to keep it in during the Milestone phase and then drop in RC? Let's create a ticket for it so we don't forget (I for one have terrible memory for these kinds of things :wink:)

ktoso: Summing up: this PR LGTM

patriknw: We must keep the deprecated for 2.4. We can't rely on that users migrate via milestones. It would be the worst kind of breaking change that we can do.

PR #17743, opened 2015-06-16
Round robin logic sequencing fix

### LINKED ISSUE #17738: RoundRobinRoutingLogic negative index after Long wrap around
Guido Medina proposed improvements here: https://groups.google.com/d/msg/akka-user/bMfdK9ox4-I/gvRSAZtzEqEJ

guidomedina:
Round robin logic sequencing will fail when it reaches Long.MAX_VALUE + 1 due to negative modulus result - fixes #17738

Will sign the agreement in a minute.

### DIFF
--- a/akka-actor/src/main/scala/akka/routing/RoundRobin.scala
+++ b/akka-actor/src/main/scala/akka/routing/RoundRobin.scala
@@ -5,8 +5,6 @@ package akka.routing
 
 import java.util.concurrent.atomic.AtomicLong
 import scala.collection.immutable
-import akka.actor.ActorContext
-import akka.actor.Props
 import akka.dispatch.Dispatchers
 import com.typesafe.config.Config
 import akka.actor.SupervisorStrategy
@@ -23,11 +21,14 @@ object RoundRobinRoutingLogic {
  */
 @SerialVersionUID(1L)
 final class RoundRobinRoutingLogic extends RoutingLogic {
-  val next = new AtomicLong(0)
+  val next = new AtomicLong
 
   override def select(message: Any, routees: immutable.IndexedSeq[Routee]): Routee =
-    if (routees.isEmpty) NoRoutee
-    else routees((next.getAndIncrement % routees.size).asInstanceOf[Int])
+    if (routees.nonEmpty) {
+      val size = routees.size
+      val nextSeq = next.getAndIncrement
+      routees((if (nextSeq < 0) size + (nextSeq % size) - 1 else nextSeq % size).asInstanceOf[Int])
+    } else NoRoutee
 
 }
 


typesafehub-validator:
Hi @guidomedina,

guidomedina:
To test the continuity theory, use int n = Integer.MAX_VALUE and then in a for each while increasing n try modulus of any positive number, once there is proof for the theory try the formula on the pull request, something like will prove it:

```
int n = Integer.MAX_VALUE
// Consider 3 the number of routees or pick any positive number.
for (int i = 0; i++; i<100) System.out.println(print ((n++) % 3));
```

drewhk:
OK TO TEST

akka-actor/src/main/scala/akka/routing/RoundRobin.scala
```diff
@@ -23,11 +21,16 @@ object RoundRobinRoutingLogic {
...
+  val next = new AtomicInteger
 
   override def select(message: Any, routees: immutable.IndexedSeq[Routee]): Routee =
     if (routees.isEmpty) NoRoutee
-    else routees((next.getAndIncrement % routees.size).asInstanceOf[Int])
+    else {
+      val nextSeq = next.getAndIncrement
+      // If nextSeq is Integer.MAX_VALUE + 1 then nextSeq is now Integer.MIN_VALUE resulting in a negative modulus operation.
+      // Math.abs would at inverted the order breaking the continuity of the sequence: N, N - 1, ..., 0, N, ....
+      routees(if (nextSeq < 0) routees.size + (nextSeq % routees.size) else nextSeq % routees.size)
```
drewhk: Hmm, if the modulus is not a power of two, then there cannot be a continuity anway unless you keep another compensation variable that you use after overflow.

guidomedina: Yes, you are absolutely correct, the Math.abs would at make it go backwards, the other version I have is the CAS one where the atomic integer is always reset to zero, but that one is way slower than the mod version due to new atomic integer adder faster than than CAS.

drewhk: I think the occasional jump in round-robinness is acceptable, especially if you keep it a Long. Reversing the sequence is of course problematic.

guidomedina: Well, there are 3 options here each with is draw back:
* AtomicInteger: Occasional jump but no need to cast from int to long.
* AtomicLong: Occasional jump unlikely to happen, always has to cast from long to int.
* AtomicInteger with CAS: The most correct solution but way slower.

guidomedina: Need to add a fix, will just force commit, once reach max value the formula should be:
`x < 0 ? size + (x % size) - 1 : x % size`

instead of

`x < 0 ? size + (x % size) : x % size`

guidomedina:
I just forced commit, re-check.

briantopping:
I don't think you have to force commit, just commit on the branch that the PR is on. Everything works properly.

guidomedina:
Yes, I know, I do it because some other projects prefer to have squashed commits to make merger/reviewer life easier.

guidomedina:
I have the CLA and I would like to sign as personal contributor, once I filled it out, where do I send it? Do I need to physically sign it, scan it and send it? or would suffice with filling the form and send it without signing it or sign with full name or something?

drewhk:
You only need to sign it on-line, you can then print it out for yourself if you want.

guidomedina:
Couldn't figure out how to sign it online, pointers or link?

patriknw:
@guidomedina You sign the CLA online here: http://www.typesafe.com/contribute/cla

As for the force commit, we prefer to just push new commits during the review phase, and in the end we will ask you to squash into one commit before merging.

We have another little convention regarding the the commit comments. This one should be something like:

    =act #17738 Improve performance of RoundRobinRoutingLogic

The details of that is described in [CONTRIBUTING.md](https://github.com/akka/akka/blob/master/CONTRIBUTING.md)

Thanks

guidomedina:
I'm a bit embarrass with the following question:

I have digitally signed the CLA but I can't find anywhere where to send it, Acrobat Reader allowed me to fill the document and digitally sign it but now it is asking for an e-mail or phone to send it, maybe it is something that for some people is so obvious, but not to me :(

ktoso:
Hi Guido, concerning the CLA it works like this: you sign it online - with the github account you're using to submit the PR and then we see the PR as "ready to go". No need to send us a paper copy of it, though you can keep it for your reference. So far it seems you did not sign it yet with the @guidomedina account though.

guidomedina:
I download it -a local copy-, filled and digitally sign it but I don't know what to do next, I was expecting Acrobat Reader to upload it or do something automatic.

ktoso:
That's the optional part actually, the part we need is:

```
Step 2) Once you or your employer has agreed to the CLA then you will need to [confirm with your GitHub account].
```

By clicking that link you'll get to a form about signing the CLA, and can do it online there.

guidomedina:
I was being an idiot...

ktoso:
No worries, and no idiots involved :-) I see it signed now, thanks! Will review code shortly

guidomedina:
Things have changed, the objective of this issue was to try to optimize RoundRobinLogic, then I realized than after Java 7 update N -unknown to me- increasing an atomic number is now done via assembler, not even CAS is involved anymore.

So CAS is slower than atomic adder and modulus operation together but then I realized that once you have reached Long.MAX_VALUE -unlikely unless you are doing DNA sequencing and need thousands of routees to compute something- it would go to Long.MIN_VALUE and modulus would return a negative number.

The first solution was to always do Math.abs but that would invert the sequence order, backward instead of forward due to modulus nature, hence the last version which will do occasional jumps if the number of routees is not a power of two.

With 4 threads I benchmark the CAS version which is the most correct but it tends to be N times slower than any of the explained versions, though we are taking slower on a very micro scale, say, 7m ops vs 30m ops

So in conclusion: The current version with AtomicInteger will have occasional jump when you reach Integer.MAX_VALUE but it will never cast a Long to Int and theoretically it should a bit faster than AtomicLong and fix the negative index bug.

Sorry for the very long paragraph.

patriknw:
@guidomedina thanks for benchmarking it. Have you also compared the proposed solution with the old AtomicLong?

akka-actor/src/main/scala/akka/routing/RoundRobin.scala
```diff
@@ -23,11 +21,15 @@ object RoundRobinRoutingLogic {
...
+  val next = new AtomicInteger
 
   override def select(message: Any, routees: immutable.IndexedSeq[Routee]): Routee =
-    if (routees.isEmpty) NoRoutee
-    else routees((next.getAndIncrement % routees.size).asInstanceOf[Int])
+    if (routees.nonEmpty) {
+      val nextSeq = next.getAndIncrement
+      // If nextSeq is Integer.MAX_VALUE + 1 then nextSeq is now Integer.MIN_VALUE resulting in a negative modulus operation.
+      // Math.abs would at inverted the order breaking the direction of the sequence -backward vs forward-
+      routees(if (nextSeq < 0) routees.size + (nextSeq % routees.size) - 1 else nextSeq % routees.size)
```
patriknw: perhaps extract `routees.size` into a `val` here?

akka-actor/src/main/scala/akka/routing/RoundRobin.scala
```diff
@@ -45,7 +47,7 @@ final class RoundRobinRoutingLogic extends RoutingLogic {
  * The router is therefore also the children's supervisor.
  *
  * The supervision strategy of the router actor can be configured with
- * [[#withSupervisorStrategy]]. If no strategy is provided, routers default to
+ * [[withSupervisorStrategy]]. If no strategy is provided, routers default to
```
patriknw: revert this change

guidomedina:
Yes, just did, 4 threads here is the result:

```
Benchmark            Mode  Cnt         Score          Error  Units
CASvsMOD.intCAS_1   thrpt    5  25521821.888 ±  1420008.668  ops/s
CASvsMOD.intCAS_4   thrpt    5   8068441.827 ±   545763.489  ops/s
CASvsMOD.intCAS_8   thrpt    5   8393989.579 ±  2825077.371  ops/s
CASvsMOD.intMOD_1   thrpt    5  31612244.740 ±   383351.317  ops/s
CASvsMOD.intMOD_4   thrpt    5  33413174.307 ±   172333.435  ops/s
CASvsMOD.intMOD_8   thrpt    5  32856744.780 ±  1068276.595  ops/s
CASvsMOD.longMOD_1  thrpt    5  38246766.516 ± 30255056.574  ops/s
CASvsMOD.longMOD_4  thrpt    5  32698458.184 ±   315833.471  ops/s
CASvsMOD.longMOD_8  thrpt    5  32710202.630 ±   492204.706  ops/s
```

guidomedina:
JMH benchmark source of important bit, including Long cast to Int:

```
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.runner.Runner;
import org.openjdk.jmh.runner.RunnerException;
import org.openjdk.jmh.runner.options.Options;
import org.openjdk.jmh.runner.options.OptionsBuilder;

import java.io.IOException;
import java.util.concurrent.atomic.AtomicInteger;
import java.util.concurrent.atomic.AtomicLong;

/**
 * @author Guido Medina, created by gmedina on 31/05/15.
 */
public class CASvsMOD {

  private static final AtomicInteger nextInt = new AtomicInteger();
  private static final AtomicLong nextLong = new AtomicLong();

  private static int nextInt(int routees) {
    return nextInt.accumulateAndGet(1, (index, increment) -> ++index < routees ? index : 0);
  }

  @Benchmark
  public void intCAS_1() {
    nextInt(1);
  }

  @Benchmark
  public void intCAS_4() {
    nextInt(4);
  }

  @Benchmark
  public void intCAS_8() {
    nextInt(8);
  }

  private int nextIntWithMod(int routees) {
    final int nextSeq = nextInt.getAndIncrement();
    return nextSeq < 0 ? nextSeq + (nextSeq % routees) - 1 : nextSeq % routees;
  }

  @Benchmark
  public void intMOD_1() {
    nextIntWithMod(1);
  }

  @Benchmark
  public void intMOD_4() {
    nextIntWithMod(4);
  }

  @Benchmark
  public void intMOD_8() {
    nextIntWithMod(8);
  }

  private int nextLong(int routees) {
    final long nextSeq = nextLong.getAndIncrement();
    return (int) (nextSeq < 0 ? nextSeq + (nextSeq % routees) - 1 : nextSeq % routees);
  }

  @Benchmark
  public void longMOD_1() {
    nextLong(1);
  }

  @Benchmark
  public void longMOD_4() {
    nextLong(4);
  }

  @Benchmark
  public void longMOD_8() {
    nextLong(8);
  }

  public static void main(String[] args) throws RunnerException, IOException {
    final Options opt = new OptionsBuilder().
      include(CASvsMOD.class.getSimpleName()).
      warmupIterations(5).
      measurementIterations(5).
      mode(Mode.Throughput).
      jvmArgs("-server").
      threads(4).
      forks(1).
      build();
    new Runner(opt).run();
  }
}
```

patriknw:
@guidomedina thanks, then it is not slower than the long.

Random fact: it would take 29247 years to overflow the long at the speed of 10000000 msg/second.

guidomedina:
It boils down to having a broken theory in the code *-correctness-* if not doing any calculation after MAX_VALUE, there isn't a performance difference to be honest, I can set the type to Long and leave the check that "will never happen" but keeping the correctness?

patriknw:
I'm fine with your proposed solution, thanks, and I sleep better when I know that it is not a bug in practise.

guidomedina:
What I mean was: Proposed solution with Long instead of Integer, it would take me no time to re-commit.

patriknw:
I mean that I'm fine with the Integer solution. Let's see if others have an opinion.

guidomedina:
I think I found the perfect solution without occasional jumps, testing: http://stackoverflow.com/questions/5844084/java-circular-shift-using-bitwise-operations

guidomedina:
Scratch that, there is a perfect solution just best effort.

guidomedina:
Corret me if I'm wrong, I have tried again the JMH of CAS vs Modulo, JMH result confuses me, if there is the case that CAS is actually doing better than Modulo then that would be the perfect solution, when you have some spared time take a look at the following Gist where I posted both result and source code, the CAS source code would be compatible with Java 6+: https://gist.github.com/guidomedina/5f4c5f116997baa05170

guidomedina:
But it would be unlikely due to new XADD low level operation for Atomic Integer/Long beating CAS by miles.

drewhk:
I think the current solution is fine, there are additional costs of enqueueing to the routee actors anyway, the incrementAndGet should be fast enough.

ktoso:
Secconded, let's stick to `incrementAndGet` "it's what it does". Comparing with CASing can be a rabbit hole and we'd need to compare all different workloads and versions of vms etc to "make sure what is faster", whereas the job we need to get done here is increment-and-get anyway.

LGTM for current impl from my side.

guidomedina:
If CAS would at been as fast as `getAndIncrement` then I wouldn't hesitate to use it as complete and correct solution for one main reason:

It never jumps the sequence, the CAS method on the JMH Gist is basically a circular integer `0, 1, ..., N, 0, ...`

drewhk:
there is always a jump if the count of routees is not a power of two, since otherwise the range of the Int/Long is undivisable by the modulus, therefore there will be a jump, unless a correction factor is introduced. I don't mind the occasional jump, we can document it.

guidomedina:
Not for this version, but it is slower than `getAndIncrement` by a factor of 3 -on a micro-level-, capping to 8m ops/sec on my laptop:

```
  private static int nextInt(int routees) {
    int prev, next;
    do {
      prev = nextInt.get();
      next = prev + 1;
    } while (!nextInt.compareAndSet(prev, next < routees ? next : 0));
    return prev;
  }
```

guidomedina:
And this would be the Scala version, it took me few minutes because I'm still a Scala newbie:

```
final class RoundRobinRoutingLogic extends RoutingLogic {
  val next = new AtomicInteger

  override def select(message: Any, routees: immutable.IndexedSeq[Routee]): Routee =
    if (routees.nonEmpty) {
      var currVal: Int = 0
      var nextVal: Int = 0
      do {
        currVal = next.getAndIncrement
        nextVal = currVal + 1
      } while (!next.compareAndSet(currVal, if (nextVal < routees.size) nextVal else 0))
      routees(currVal)
    } else NoRoutee
}
```

guidomedina:
**Note:** Corrected version.

There is no jump for this version but it is slower than `getAndIncrement` by a factor of 3 *-on a micro-level-* capping to 8m ops on my laptop with 4 threads and 9m ops on a decent server with 8 CPUs and threads:

```
final class RoundRobinRoutingLogic extends RoutingLogic {
  val next = new AtomicInteger

  override def select(message: Any, routees: immutable.IndexedSeq[Routee]): Routee =
    if (routees.nonEmpty) {
      var currVal = 0
      var nextVal = 0
      val size = routees.size
      do {
        currVal = next.get
        nextVal = currVal + 1
      } while (!next.compareAndSet(currVal, if (nextVal < size) nextVal else 0))
      routees(currVal)
    } else NoRoutee

}
```

ktoso:
Wrote a benchmark to collect all the datapoints in one place, was too confusing to skim comments :-)
https://gist.github.com/ktoso/68d47c117c547343c58b

Throughput wise I don't think we have a problem as it most likely would be dominated by other factors.
Perhaps adding a StrictRoundRobin would be the solution? Not decided yet, need to ponder this a bit more..

guidomedina:
Correct selectCas, my initial version had an error, is `nextInt.get` instead of `next.getAndIncrement`

guidomedina:
Correction, I see selectCas and selectCas2, remove selectCas and just keep selectCas2

guidomedina:
**atomic_int** is the fastest which is the current proposed version with avg of 1m ops/ms
**cas_int2** is the most correct version performing OK with avg of 850K ops/ms

guidomedina:
CAS versions are not good, a further analysis shows that there is a possibility of error, if a routee is removed from the list *(new size = old size - 1)* and `next.get == old size - 1`, an index out of bounds exception is thrown, making the CAS useless and fixable by doing a modulo operation which leads us back to the original solutions, that said, the only two usable and correct solutions *-with occasional jump one more often than the other-* from your benchmark are:
* **atomic_int** *-selectInt(...) method-*
* **atomic_long** *-selectLong(...) method-*

guidomedina:
Well, select CAS is fixable with the following, it would just take an extra cycle in the while if that happen:

```
override def select(message: Any, routees: immutable.IndexedSeq[Routee]): Routee =
    if (routees.nonEmpty) {
      var currVal = 0
      var nextVal = 0
      val size = routees.size
      do {
        currVal = next.get
        nextVal = currVal + 1
        // if currVal >= size it will do an extra iteration.
      } while (!next.compareAndSet(currVal, if (nextVal < size) nextVal else 0) && currVal >= size)
      routees(currVal)
    } else NoRoutee
```

guidomedina:
Edited previous comment.

guidomedina:
Are you sure your benchmark is correct, I'm measuring the important bit which is the picking of the index and I can't get close to your numbers, I mean, 1m ops/ms with 10 threads?, Are you sure your bench is running with concurrency enabled?

That sounds too high to be true, I can only get 33K ops/ms with 8 threads, I modified my benchmark to match yours and ran it, CAS is always slower with a factor of 3~4 times, results & source code: https://gist.github.com/guidomedina/5f4c5f116997baa05170

ktoso:
I'm not sure if is - just wanted to collect what we're talking about in one file instead of collecting things from comments each time :-)
33k ops sounds really little for something that's basically just a CAS or XADD... whats your bench box?
The numbers above I got from my laptop (like I mentioned, not super serious benchmarking, just wanted to collect these) `Intel(R) Core(TM) i7-4850HQ CPU @ 2.30GHz`.

guidomedina:
Intel® Core™ i5-4200U CPU @2.6GHz *-concurrently used they really go up to a max of 2.2GHz-* remember you are using as measurement unit ms, so 33K ops/ms would be equivalent to 33m ops/s

guidomedina:
@ktoso this is a more serious benchmark, done on a Debian Server with 8 CPUs which clearly shows that Long (64K ops/ms) is better than Integer (53K ops/ms), so I should probably keep the Long version with the adjustment for negative values after Long passes Long.MAX_VALUE.

```
Benchmark         (routees)   Mode  Cnt      Score      Error   Units
CASvsMOD.intCAS           8  thrpt   10   9355.365 ±   21.177  ops/ms
CASvsMOD.intCAS         512  thrpt   10   8785.338 ±   36.723  ops/ms
CASvsMOD.intMOD           8  thrpt   10  53464.546 ±  209.795  ops/ms
CASvsMOD.intMOD         512  thrpt   10  39683.685 ±  168.896  ops/ms
CASvsMOD.longMOD          8  thrpt   10  64774.359 ± 1443.127  ops/ms
CASvsMOD.longMOD        512  thrpt   10  64490.088 ±  272.969  ops/ms
```

guidomedina:
Closing pull request, wrong branch and too messy. Recreating new one under the right branch.

Issue #16542, opened 2014-12-15
Don't store sender ActorRef


patriknw:
We should investigate if there is a need any more to store the sender ActorRef in each persistent event.

PR #17744, opened 2015-06-16
=per #16542 Don't store sender in PersistentRepr

patriknw:
- I think it originated from channels, or some idea that
  the sender should be revived (as good as possible) during replay,
  but that is pretty useless
- It must still be in PersistentRepr for remote serialization
- I didn't want to change to the built in sender when looping to the
  journal because keeping it together with the message makes it easier
  to do batching (queueing)

### DIFF
--- a/akka-persistence/src/test/scala/akka/persistence/serialization/SerializerSpec.scala
+++ b/akka-persistence/src/test/scala/akka/persistence/serialization/SerializerSpec.scala
@@ -148,7 +148,7 @@ class MessageSerializerPersistenceSpec extends AkkaSpec(customSerializers) {
   "A message serializer" when {
     "not given a manifest" must {
       "handle custom Persistent message serialization" in {
-        val persistent = PersistentRepr(MyPayload("a"), 13, "p1", true, testActor)
+        val persistent = PersistentRepr(MyPayload("a"), 13, "p1", true)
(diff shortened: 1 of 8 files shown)


patriknw:
Refs #16542

patriknw:
ah, I didn't run the tck tests locally

akka-persistence/src/test/scala/akka/persistence/serialization/SerializerSpec.scala
```diff
@@ -221,6 +221,31 @@ class MessageSerializerPersistenceSpec extends AkkaSpec(customSerializers) {
       }
     }
 
+    "given PersistentRepr serialized with Akka 2.3.11 Scala 2.10" must {
```
ktoso: To be more realistic this probably should be Scala 2.11, since 2.4 is not released for 2.10. Worth the trouble you think?

patriknw: I actually did that on purpose, to make sure that we can replay these things when they are serialized with 2.10 - 2.3.11

ktoso: Right that's important, good then :)

ktoso:
LGTM

PR #17745, opened 2015-06-16
=str add VirtualProcessor

rkuhn:
This is only for reviewing the idea, the name is of course wrong and it also needs tests (and be used of course).

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
@@ -4,7 +4,6 @@
 package akka.stream.impl
 
 import java.util.concurrent.atomic.{ AtomicInteger, AtomicBoolean, AtomicReference }
-
 import akka.stream.impl.StreamLayout.Module
 import akka.stream.scaladsl.Keep
 import akka.stream._
@@ -12,6 +11,8 @@ import org.reactivestreams.{ Processor, Subscription, Publisher, Subscriber }
 import scala.collection.mutable
 import scala.util.control.NonFatal
 import akka.event.Logging.simpleName
+import scala.annotation.tailrec
+import java.util.concurrent.atomic.AtomicLong
 
 /**
  * INTERNAL API
@@ -106,9 +107,8 @@ private[akka] object StreamLayout {
         AmorphousShape(shape.inlets ++ that.shape.inlets, shape.outlets ++ that.shape.outlets),
         downstreams ++ that.downstreams,
         upstreams ++ that.upstreams,
-        if (f eq Keep.left) matComputation1
-        else if (f eq Keep.right) matComputation2
-        else Combine(f.asInstanceOf[(Any, Any) ⇒ Any], matComputation1, matComputation2),
+        // would like to optimize away this allocation for Keep.{left,right} but that breaks side-effecting transformations
+        Combine(f.asInstanceOf[(Any, Any) ⇒ Any], matComputation1, matComputation2),
         attributes)
     }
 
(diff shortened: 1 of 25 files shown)


akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -293,6 +293,51 @@ private[akka] object StreamLayout {
...
+  import V._
+  import ReactiveStreamsCompliance._
+
+  private val subscriptionStatus = new AtomicReference[AnyRef]
+  private val terminationStatus = new AtomicReference[Termination]
+
+  override def subscribe(s: Subscriber[_ >: T]): Unit =
+    if (subscriptionStatus.compareAndSet(null, s)) () // wait for onSubscribe
+    else {
+      val sub = subscriptionStatus.getAndSet(s).asInstanceOf[Subscription]
```
drewhk: this will fail on the second subscribe call with a ClassCastException

rkuhn: Right, this should be turned into a SpecViolation instead.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -293,6 +293,51 @@ private[akka] object StreamLayout {
...
+      tryOnSubscribe(s, sub)
+      terminationStatus.getAndSet(Allowed) match {
```
drewhk: this will be null if onSubscribe was first, since it does not set terminationStatus after the successful CAS on subscriptionStatus, and this will enter this branch since the CAS will fail here on subscriptionStatus. Maybe you wanted this to have the initial value of Allowed?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: It is not a spec violation, just that this VirtualProcessor does not support fanout. This is btw one of the drawbacks of this approach, if you use Sink.publisher to access the features a possibly more powerful Publisher to the left, now it is reduced to single-subscription. I think it is fine, just for the record.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: No, you just need to ignore the null, since setting it to Allowed in the getAndSet will force the onComplete to fail the CAS and call the onComplete itself

drewhk:
We can test it against the TCK by appending this processor to a real, actor backed identity processor:
 - attach to the left end
 - attach to the right end
 - attach to both ends

drewhk:
I have verified it, and it looks correct (I can present an informal "proof")

TODO:
 - test the mapMaterialized case where the Publisher/Subscriber is accessed and used in the mapper
 - add TCK tests with this appended to left, right, both ends of an IdentityProcessor
 - remove all the Virtual stuff and replace it with this one
 - remove and inline the attach() method in MaterializerSession
 - remove the createIdentityProcessor method in MaterializerSession

rkuhn:
okay, going forward with this now

rkuhn:
Implemented all the things and found a bonus bug as well (more correctly: an inconsistency in how `Keep.left` would completely drop the right computation instead of only discarding its value).

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -550,6 +620,7 @@ private[stream] abstract class MaterializerSession(val topLevel: StreamLayout.Mo
         case mv: MaterializedValueSource[_] ⇒
           val pub = new MaterializedValuePublisher
           materializedValuePublishers ::= pub
+          materializedValues.put(mv, ())
```
drewhk: You should run the materialization benchmarks (or add a new one) to see how much impact does this have (given that the default is Keep.left almost always). If it is significantly slower ( > 5%) then I am against this.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -293,38 +293,108 @@ private[akka] object StreamLayout {
...
+    tryOnNext(subscriptionStatus.get.asInstanceOf[Subscriber[T]], t)
+  }
+
+  private final class Sub(s: Subscription) extends AtomicLong with Subscription {
+    override def cancel(): Unit = {
+      subscriptionStatus.set(InertSubscriber)
+      s.cancel()
+    }
+    @tailrec
+    override def request(n: Long): Unit = {
```
drewhk: ewww. Why is this needed? There is something frightening about double accounting of requests.

drewhk:
Please also change the materialization of the Identity module to use this instead of Transform(() => Map(identity))

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: This is not optional: without it the tests for accessing the materialized value fail. I don’t expect much impact on real-world benchmarks since this only affects usage of `FlowGraph.Builder.materializedValue`.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: Without this the TCK fails: it rightfully signals reentrant signaling towards the Subscriber since requesting from onSubscribe may now immediately trigger the asynchronous invocation of onNext. Therefore requests during onSubscribe are delayed by this method.

rkuhn:
Concerning changing the materialization of Identity nodes: is the intention that these will be removed before fusing anyway? In that case I agree.

rkuhn:
@drewhk I have replaced Identity by this VirtualProcessor and then by pure chance I saw the same failure as in https://jenkins.akka.io/job/pr-validator-per-commit-jenkins/2953/ locally: ConcatAllImpl suffers from stopping the ActorProcessor too eagerly (when onSubscribe has not yet been received from the substream). Please review the second commit carefully.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: Ok.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: Well, let's accept this as a feature then.

drewhk:
> Concerning changing the materialization of Identity nodes: is the intention that these will be removed before fusing anyway? In that case I agree.

These cannot be removed completely trivially in all cases, see

   incomingConneciton.flow.join(Flow[ByteString]) // echo server

I guess it is possible to detect this in the materialization phase, but fusing cannot help here.

rkuhn:
Failure was remaining child actors in TlsSpec—is that one of the typical failure modes or a new one?

@drewhk Please review especially the last three commits since they fix problems that are unrelated to the VirtualProcessor itself, they were racy before and might have come up now due to timing changes.

drewhk:
yes, that is the typical failure.

drewhk:
mergeable?

rkuhn:
Yes, I’d say it is :-)

rkuhn:
Have you seen the improved fix in 5b420c5?

drewhk:
yes, please squash

rkuhn:
It is already squashed: the commits fix different things.

Issue #17752, opened 2015-06-17
Latest version of Sphinx breaks PDF generation task


kukido:
After following documentation instructions from http://doc.akka.io/docs/akka/snapshot/dev/documentation.html I was not able to generate PDF and was getting the following error:

```
[debug] ! Undefined control sequence.
[debug] <argument> \node@class@name
[debug]
[debug] l.18759 ...ection{Partitioning \DUspan{}{{[}*{]}}}
[debug]
[debug] ?
[debug] ! Emergency stop.
[debug] <argument> \node@class@name
[debug]
[debug] l.18759 ...ection{Partitioning \DUspan{}{{[}*{]}}}
[debug]
[debug] !  ==> Fatal error occurred, no output PDF file produced!
[debug] Transcript written on AkkaJava.log.
[debug] make: *** [AkkaJava.pdf] Error 1
```

As it was my workstation, I tried to repeat the procedure in "clean room" environment of VM. Followed the instructions and got exactly the same error.

Obviously, something was wrong. With help from @rkuhn and @ktoso was able to eliminate MacTeX, and the only variable left was Sphinx. Since HTML generation was fine, it was never a main suspect :-)

@ktoso mentioned his version of Sphinx - `1.2.3`, while the latest automatically installed by `pip` is `1.3.1`.

```
You are using pip version 6.1.1, however version 7.0.3 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Sphinx (1.2.3)
```

I was able to restore the build by forcing `1.2.3` version of Sphinx with `pip`. Will submit the documentation patch shortly.

patriknw:
Thanks for sharing the problem and solution. It would be great if we could make it work with latest, so that we are not stuck on an old version.

kukido:
The following two lines in cluster.rst are causing the generation to fail:

```
Partitioning :ref:`[*] <niy>`
=============================
...
Stateful Actor Replication :ref:`[*] <niy>`
===========================================
```

Looking into Sphinx documentation to see what was changed.

kukido:
The first version of Sphinx not working with current docs is `1.3`

 Looking at version `1.3` [changelog](http://sphinx-doc.org/changes.html#release-1-3-released-mar-10-2015) I see only one change which could have an effect: [Error building sphinx_rtd_theme demo docs with pdf target](https://github.com/sphinx-doc/sphinx/issues/1745)

Now, back to akka documentation. It seems that taking "footnotes" from sections titles solves the issue.

[...]

PR #17753, opened 2015-06-17
=doc #17752 Compatibility with latest version of Sphinx

kukido:


2m:
What is more interesting it does not fail with 1.3.1 on `release-2.3-dev` (streams) branch when generating streams documentation `akka-docs-dev/sphinx:generatePdf` but fails when generating general docs `akka-docs/sphinx:generatePdf`. It would be great to find the underlying cause of the failure.

kukido:
I've done some research and experimentations with rst/Sphinx, removing "footnotes" from two sections solves the compatibility issue and looks like simple/elegant solution. I was able to build the docs with `1.3.1` after making the changes. #17752 has more details.

2m:
I have rebased your changes here: #17947

PR #17761, opened 2015-06-17
=pro verify BC against all binary compatible versions

2m:
This is https://github.com/akka/akka/pull/17374 ported to master.

### DIFF
--- a/project/AkkaBuild.scala
+++ b/project/AkkaBuild.scala
@@ -53,7 +53,7 @@ object AkkaBuild extends Build {
         samplePersistenceJava, samplePersistenceScala, samplePersistenceJavaLambda,
         sampleRemoteJava, sampleRemoteScala, sampleSupervisionJavaLambda,
         sampleDistributedDataScala, sampleDistributedDataJava),
-      
+
       S3.host in S3.upload := "downloads.typesafe.com.s3.amazonaws.com",
       S3.progress in S3.upload := true,
       mappings in S3.upload <<= (Release.releaseDirectory, version) map { (d, v) =>
@@ -404,17 +404,28 @@ object AkkaBuild extends Build {
     mavenLocalResolverSettings ++
     JUnitFileReporting.settings ++ StatsDMetrics.settings
 
-  def akkaPreviousArtifact(id: String): Def.Initialize[Option[sbt.ModuleID]] = Def.setting {
+  def akkaPreviousArtifacts(id: String): Def.Initialize[Set[sbt.ModuleID]] = Def.setting {
     if (enableMiMa) {
-      val version: String = "2.4.0" // FIXME verify all 2.3.x versions
-      val fullId = crossVersion.value match {
-        case _ : CrossVersion.Binary => id + "_" + scalaBinaryVersion.value
-        case _ : CrossVersion.Full => id + "_" + scalaVersion.value
-        case CrossVersion.Disabled => id
+      val versions = {
+        val akka23Versions = Seq("2.3.11", "2.3.12", "2.3.13", "2.3.14")
+        val akka24Versions = Seq("2.4.0")
+        val akka24NewArtifacts = Seq(
+          "akka-cluster-sharding",
+          "akka-cluster-tools",
+          "akka-persistence",
+          "akka-distributed-data-experimental",
+          "akka-persistence-query-experimental"
+        )
+        scalaBinaryVersion.value match {
+          case "2.11" if !akka24NewArtifacts.contains(id) => akka23Versions ++ akka24Versions
+          case _ => akka24Versions // Only Akka 2.4.x for scala > than 2.11
+        }
       }
-      Some(organization.value % fullId % version) // the artifact to compare binary compatibility with
+
+      // check against all binary compatible artifacts
+      versions.map(organization.value %% id % _).toSet
     }
-    else None
+    else Set.empty
   }
 
   def loadSystemProperties(fileName: String): Unit = {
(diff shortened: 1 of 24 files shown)


project/plugins.sbt
```diff
@@ -17,7 +17,8 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "0.7.1")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-osgi" % "0.6.0")
 
-addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.7")
+addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.7-1-g841f792")
+resolvers += Resolver.url("2m-sbt-plugins", url("https://dl.bintray.com/2m/sbt-plugins/"))(Resolver.ivyStylePatterns)
```
ktoso: When can we release it "proper"? Who can?

project/AkkaBuild.scala
```diff
@@ -361,17 +361,27 @@ object AkkaBuild extends Build {
...
+
+      // check against all binary compatible artifacts
+      val artifacts =
+        for (version <- first23.to(last23).map("2.3." + _))
+          yield organization.value %% id % version
+
+      artifacts.toSet
     }
-    else None
+    else Set.empty
```
ktoso: nice

ktoso:
LGTM, though would like to release the plugin "for real" :)

project/plugins.sbt
2m: I know @dotta can. :) But we need to merge https://github.com/typesafehub/migration-manager/pull/64 first.

2m:
PLS BUILD

ktoso:
@2m should we merge this now?

ktoso:
Or we can't release the plugin still - only got repo admin rights?

ktoso:
@2m what's the status here?

patriknw:
@2m are we waiting for anything here or can we merge this?

project/plugins.sbt
patriknw: looks like it has been merged, could we release the proper version?

2m:
I was working on this yesterday. The only thing that we need here is to release `sbt-mima-plugin` which I am on currently.

patriknw:
thanks

patriknw:
any news here? Note that currently we are checking against 2.4.0. If we are to check against at least latest 2.3.x we need a way to define different mima-filters for different versions.

2m:
Play 2.4.3 uses Akka 2.3.13, and Play 2.3.10 is using Akka 2.3.4. Spark 1.5.1 is using Akka 2.3.11. It looks like we need to go back as much back as 2.3.4 when checking 2.4.x BC.

patriknw:
To be honest I think those that are using Akka 2.3.4 have serious problems anyway.
If we can use the same mima filter for 2.3.4 I don't mind checking, but trying to reproduce the exact filter that for 2.3.4 is a major undertaking (I think).

ktoso:
+1 for checking `2.3.11`, `2.3.14`, `2.4.0`

2m:
Addded version conditional BC issue filters. Note that on a subsequent (if any) 2.3.x release `bcIssuesBetween23and24` filters will have to be moved to the new version.

project/AkkaBuild.scala
```diff
@@ -404,17 +404,26 @@ object AkkaBuild extends Build {
...
-        case _ : CrossVersion.Full => id + "_" + scalaVersion.value
-        case CrossVersion.Disabled => id
+      val versions = {
+        val akka23Versions = Seq("2.3.11", "2.3.12", "2.3.13", "2.3.14")
+        val akka24Versions = Seq("2.4.0")
+        scalaBinaryVersion.value match {
+          case "2.11" if !Seq(
+            "akka-cluster-sharding",
+            "akka-cluster-tools",
+            "akka-persistence",
```
patriknw: what does this mean? is it excluding those modules from compat check?

2m: It means no `2.3.x` versions for these modules.

patriknw: but akka-persistence-2.4-SNAPSHOT would still be checked against akka-persistence-2.4.0?

2m: Yes, it will be matched by the next case statement.

ktoso: can we make this a bit more named methods? I also was a bit confused what actually will be tested

patriknw: ah, now I see. please extract to something like
```scala
val newArtifactsInAkka24 = Set(
            "akka-cluster-sharding",
            "akka-cluster-tools",
            "akka-persistence",
            "akka-distributed-data-experimental",
            "akka-persistence-query-experimental")
```

project/plugins.sbt
```diff
@@ -17,7 +17,8 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "0.7.1")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-osgi" % "0.6.0")
 
-addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.7")
+addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.1.7-32-g3bebac4")
+resolvers += Resolver.url("2m-sbt-plugins", url("https://dl.bintray.com/2m/sbt-plugins/"))(Resolver.ivyStylePatterns)
```
ktoso: If we merge in this form let's have a ticket for updating once MiMa gets a release.

2m: Okay. https://github.com/akka/akka/issues/18765

patriknw:
LGTM

patriknw:
@2m looks like the akka-docs jenkins job failed with unresolved dependency: com.typesafe#sbt-mima-plugin;0.1.7-32-g3bebac4: not found

https://jenkins.akka.io:8498/job/akka-docs/680/console

Issue #17759, opened 2015-06-17
Exceptions thrown in Subscription are not gracefully handled


jroper:
The following source:

```scala
  def errorSource[T] = Source(new Publisher[T] {
    def subscribe(s: Subscriber[_ >: T]) = {
      s.onSubscribe(new Subscription {
        def cancel() = ()
        def request(n: Long) = sys.error("error")
      })
    }
  })
```

As you can see request simply throws an error, rather than passing an error to the subscriber.  If I then:

```scala
Await.result(errorSource[Int].runFold(0)(_ + _), 5.seconds)
```

I expected that code to throw some sort of exception, but it doesn't, it times out.

drewhk:
I think RS Spec rule 2.13 is relevant here:

> onSubscribe, onNext, onError or onComplete MUST return normally except when any provided parameter is null in which case it MUST throw a java.lang.NullPointerException to the caller, for all other situations the only legal way for a Subscriber to signal failure is by cancelling its Subscription. In the case that this rule is violated, any associated Subscription to the Subscriber MUST be considered as cancelled, and the caller MUST raise this error condition in a fashion that is adequate for the runtime environment.

drewhk:
Ah, sorry it is 3.16: 

> Calling Subscription.request MUST return normally. The only legal way to signal failure to a Subscriber is via the onError method.

ktoso:
You may want to use the TCK ( https://github.com/reactive-streams/reactive-streams-jvm/tree/master/tck ) when implementing Publisher or Subscriber manually, there's a lot of weird corner cases one would notmally not think of - those tests should help you implement a valid Publisher.

jroper:
@ktoso did you mean @jroper? :smile:

I know it's not implemented correctly, but there's implementing the spec, and then there's gracefully handling spec non conformance. The first is the happy case, any old implementation can do that.  The second is about robustness, and is the difference between libraries that don't have a good development experience and libraries that provide an excellent development experience, it's about helping developers understand what they've done wrong when they've done something wrong, by providing a meaningful error message at a pertinent location.

ktoso:
argh sorry guys, hit `@j` and `[enter]` :wink:

jrudolph:
No worries, I just recently tried to build a custom reactive stream processor at which point the TCK came very handy. (For saving time, that is. In the end I threw away my custom implementation and went with a combination of akka-stream components because I couldn't get the TCK to pass with reasonable effort)

drewhk:
[...]

PR #17762, opened 2015-06-17
=str #17759: Handle when request() throws in ActorInterpreter

drewhk:


drewhk:
Fixes #17762

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorInterpreter.scala
```diff
@@ -145,7 +148,7 @@ private[akka] class BatchingActorInputBoundary(val size: Int, val name: String)
 
     case OnComplete                ⇒ onComplete()
     case OnError(cause)            ⇒ onError(cause)
```
ktoso: should then these be `tryOnError`, according to the same reasoning then?

drewhk: that's not the onError you are looking for ;)

drewhk: (line 127/130)

ktoso: lol, I see :-)

ktoso:
LGTM

PR #17763, opened 2015-06-17
=act #17690 add additional validation for name param in RootActorPath

### LINKED ISSUE #17690: No consistency in actorSelection for ActorPath & String
Does work
```scala
val path = RootActorPath(address, "/user/sharding/*Coordinator/singleton/*")
system.actorSelection(   path.toString   ) ! Identify("singleton")
```

Does NOT work
```scala
val path = RootActorPath(address, "/user/sharding/*Coordinator/singleton/*")
system.actorSelection(   path   ) ! Identify("singleton")
```

ktoso:
The `RootActorPath` is a very specific thing - the ref to the root Actor.
The use case described in #17690 worked (after toString-ing) by accident, and we should not allow constructing such RootActorPath - the parameter is a name which may not include path separators (same as path element).

Resolves #17690

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/ActorPath.scala
+++ b/akka-actor/src/main/scala/akka/actor/ActorPath.scala
@@ -254,6 +254,10 @@ sealed trait ActorPath extends Comparable[ActorPath] with Serializable {
  */
 @SerialVersionUID(1L)
 final case class RootActorPath(address: Address, name: String = "/") extends ActorPath {
+  require(name.length == 1 || name.indexOf('/', 1) == -1,
+    "/ may only exist at the beginning of the root actors name, " +
+      "it is a path separator and is not legal in ActorPath names: [%s]" format name)
+  require(name.indexOf('#') == -1, "# is a fragment separator and is not legal in ActorPath names: [%s]" format name)
 
   override def parent: ActorPath = this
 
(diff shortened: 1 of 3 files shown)


akka-actor/src/main/scala/akka/actor/ActorPath.scala
```diff
@@ -254,6 +254,10 @@ sealed trait ActorPath extends Comparable[ActorPath] with Serializable {
  */
 @SerialVersionUID(1L)
 final case class RootActorPath(address: Address, name: String = "/") extends ActorPath {
+  require(name.length == 1 || name.indexOf('/', 1) == -1,
```
t3hnar: oh...that will break a lot of stuff... I think you should put a note into migration guide

t3hnar: You know that I can still have `/user/blabla/*` there with a `copy` method ;)

ktoso: - note in migration guide might not hurt, sure;
- nope, you can't - `copy` calls the constructor - thus the `require` will catch such attempts

ktoso:
Added migration guide

ktoso:
Any more reviewers?

patriknw:
LGTM

PR #17773, opened 2015-06-18
!htc #17656 do not render host header default port values

### LINKED ISSUE #17656: When generating a Host header don't include default port
/cc @sirthias

sirthias:
Fixes #17656.

/cc @jrudolph

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/client/LowLevelOutgoingConnectionSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/client/LowLevelOutgoingConnectionSpec.scala
@@ -30,7 +30,7 @@ class LowLevelOutgoingConnectionSpec extends AkkaSpec("akka.loggers = []\n akka.
         requestsSub.sendNext(HttpRequest())
         expectWireData(
           """GET / HTTP/1.1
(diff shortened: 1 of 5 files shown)


akka-http-core/src/test/scala/akka/http/impl/engine/client/LowLevelOutgoingConnectionSpec.scala
```diff
@@ -350,7 +350,7 @@ class LowLevelOutgoingConnectionSpec extends AkkaSpec("akka.loggers = []\n akka.
   class TestSetup {
     val requests = TestPublisher.manualProbe[HttpRequest]
     val responses = TestSubscriber.manualProbe[HttpResponse]
-    val remoteAddress = new InetSocketAddress("example.com", 80)
+    val remoteAddress = new InetSocketAddress("example.com", 0)
```
jrudolph: This change seems a bit strange. If the `remoteAddress` is only used to construct the `Host` instance below, replacing the `remoteAddress` with the `Host` object directly seems clearer.

jrudolph:
lgtm

drewhk:
thanks, LGTM

PR #17777, opened 2015-06-18
Make jackson marshallers adapt Jackson's best practices

matsluni:
Hi,

I tried to fix this one. It's my first attempt.
I think the current fix is not thread safe, but I was unsure how to do it. What would be the best way the ensure thread safety for the objectMapper?

WDYT?

### DIFF
--- a/akka-http-marshallers-java/akka-http-jackson/src/main/scala/akka/http/javadsl/marshallers/jackson/Jackson.scala
+++ b/akka-http-marshallers-java/akka-http-jackson/src/main/scala/akka/http/javadsl/marshallers/jackson/Jackson.scala
@@ -13,8 +13,11 @@ import akka.http.javadsl.server.{ Unmarshaller, Marshaller }
 import akka.http.impl.server.{ UnmarshallerImpl, MarshallerImpl }
 
 object Jackson {
-  def json[T <: AnyRef]: Marshaller[T] = _jsonMarshaller.asInstanceOf[Marshaller[T]]
-  def jsonAs[T](clazz: Class[T]): Unmarshaller[T] =
+  private val objectMapper: ObjectMapper = new ObjectMapper().enable(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY)
+  def json[T <: AnyRef]: Marshaller[T] = jsonMarshaller(objectMapper).asInstanceOf[Marshaller[T]]
+  def json[T <: AnyRef](objectMapper: ObjectMapper): Marshaller[T] = jsonMarshaller(objectMapper).asInstanceOf[Marshaller[T]]
+  def jsonAs[T](clazz: Class[T]): Unmarshaller[T] = jsonAs(objectMapper, clazz)
+  def jsonAs[T](objectMapper: ObjectMapper, clazz: Class[T]): Unmarshaller[T] =
     UnmarshallerImpl[T] { (_ec, _flowMaterializer) ⇒
       implicit val ec = _ec
       implicit val mat = _flowMaterializer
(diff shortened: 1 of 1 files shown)


ktoso:
// cc @sirthias @jrudolph

ktoso:
Hi and thanks for the PR!
I thought just a val would be ok, but not sure about thread safety of the object mapper itself btw

matsluni:
I thought it would be nice for an enduser to supply an own customized version of an objectMapper, that is why I used a `var` and have a setter. I am not sure about the thread safety of the objectMapper either.

jrudolph:
Yay, @matsluni, welcome back to the interwebs! :umbrella: :closed_umbrella: :umbrella: :heart_eyes_cat: 

Have you seen the guide here: http://wiki.fasterxml.com/JacksonBestPracticesPerformance? It says something about the thread-safety of the mapper. I'd say we change the signature of the methods to make it possible to pass an `ObjectMapper` in and then we create another set of overloads that look like the current ones which use an immutable global ObjectMapper in cases where you don't want to configure one yourself. This one can put into a `private val` of the object. WDYT?

matsluni:
Yeah! Have some right now and thought I try one issue :-)

> Have you seen the guide here: http://wiki.fasterxml.com/JacksonBestPracticesPerformance

You are right! Haven't seen it.

So, if I understand correct you suggest a `private val objectMapper` for the case the user doesn't supply its own version. And additionally provide overloads like `def json(objectMapper: ObjectMapper)`... for the case the user wants to use its own customized version of an objectMapper?

Correct? 

I will update my fix and the PR.

jrudolph:
Exactly :)

matsluni:
I updated the fix. Please have a look.

I also changed `private val _jsonMarshaller`... to `private def jsonMarshaller(objectMapper: ObjectMapper)`... to avoid the code duplication. But this will create a new objectWriter every time.

WDYT?

matsluni:
I am thinking about the use case in which the user uses its own customized version of the objectMapper. Every time de-/serialization takes place the user must supply the correct instance of the objectMapper. Wouldn't it be easier and less error prone if one could set up the Jackson instance upfront with the right configuration and use this later just with: `Jackson.<Pet>json()???

jrudolph:
Did you push anything new? It seems it's still the same code?

matsluni:
Hah, you're right, missed the push. Now it should be there.

akka-http-marshallers-java/akka-http-jackson/src/main/scala/akka/http/javadsl/marshallers/jackson/Jackson.scala
```diff
@@ -23,16 +26,16 @@ object Jackson {
...
           }
       }
     }(ClassTag(clazz))
 
-  private val _jsonMarshaller: Marshaller[AnyRef] =
+  private def jsonMarshaller(objectMapper: ObjectMapper): Marshaller[AnyRef] =
     MarshallerImpl[AnyRef] { implicit ec ⇒
       marshalling.Marshaller.StringMarshaller.wrap(`application/json`) { (value: AnyRef) ⇒
-        val writer = new ObjectMapper().enable(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY).writer()
+        val writer = objectMapper.enable(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY).writer()
```
jrudolph: We cannot change possibly shared setting here. Let's do it for the predefined ObjectMapper during initialization.

jrudolph:
> Wouldn't it be easier and less error prone if one could set up the Jackson instance upfront with the right configuration and use this later just with: Jackson.<Pet>json()???

Do we prevent this in any way? I think that's how you should do it. Or do you mean that there should be a factory that allows to create marshallers using one ObjectMapper for different target types?

akka-http-marshallers-java/akka-http-jackson/src/main/scala/akka/http/javadsl/marshallers/jackson/Jackson.scala
matsluni: You're right.

matsluni:
With the new approach and the possibility to use a customized ObjectMapper, this ObjectMapper has to be provided every time in the user code where json de-/serialization happens. 
I thought I might be better to configure the Jackson Object with user given ObjectMapper at instantiation time (or right after) of the Jackson Object and then just use it like before e.g. `Jackson.json()`. So, the user can not forget to provide his own configured ObjectMapper. But I don't now if this is possible.

2beaucoup:
Have you tried passing the `ObjectMapper` implicitly?

jrudolph:
@2beaucoup this is for the dark side, so implicits are out of the question.

guidomedina:
I have worked for few years with Jackson so my 2 cents:
* `ObjectMapper` initialization is heavy and once initialized it is thread safe.
* The problem with having an internal `ObjectMapper` instance is that there are too many modules out there to account for, like Jackson's Joda time module, etc, etc, so you will not have a choice there than to require the user to give you one.

ktoso:
So seems that with the current style it's easy to use your own mapper, is there anything missing from this PR still or shall we merge?

ktoso:
OK TO TESt

ktoso:
OK TO TEST

matsluni:
Hi @ktoso, it is ok for me to leave like that for now. It is an improvement for the user to use a customized version of the ObjectMapper. @jrudolph, WDYT?

drewhk:
What is the conclusion here then? Should we merge?

ktoso:
Yes, let's merge - it's an improvement over the existing API

Issue #17786, opened 2015-06-19
Cancelling the bindStream should not cancel the already accepted connections


drewhk:
It should wait until children are stopped instead, relying on subscription timeout to clear never used connections.

drewhk:
And this should be configurable via Attributes

Issue #17789, opened 2015-06-19
Handle subscription timeout in all Pulblishers


drewhk:

rkuhn:
Indeed!

PR #17790, opened 2015-06-19
=str #17786: Cancelling bind stream should not cancel children

drewhk:
Also added subscription timeout for connection actors

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/io/TcpSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/io/TcpSpec.scala
@@ -3,34 +3,21 @@
  */
 package akka.stream.io
 
-import akka.actor.{ ActorSystem, Kill }
-import akka.stream.scaladsl.Tcp.OutgoingConnection
-
-import scala.collection.immutable
-import scala.concurrent.{ Future, Await }
+import akka.actor.{ActorSystem, Kill}
 import akka.io.Tcp._
-import akka.stream.{ BindFailedException, ActorFlowMaterializer, ActorFlowMaterializerSettings, StreamTcpException }
 import akka.stream.scaladsl.Tcp.IncomingConnection
-import akka.stream.scaladsl.{ Flow, _ }
+import akka.stream.scaladsl.{Flow, _}
 import akka.stream.testkit.TestUtils.temporaryServerAddress
 import akka.stream.testkit.Utils._
 import akka.stream.testkit._
-import akka.util.{ Helpers, ByteString }
+import akka.stream.{ActorFlowMaterializer, BindFailedException, StreamTcpException}
+import akka.util.{ByteString, Helpers}
 
 import scala.collection.immutable
-import akka.stream.{ ActorFlowMaterializer, StreamTcpException, BindFailedException }
-
 import scala.concurrent.Await
 import scala.concurrent.duration._
-import akka.util.ByteString
-import akka.stream.scaladsl.Flow
-import akka.stream.testkit._
-import akka.stream.testkit.Utils._
-import akka.stream.scaladsl._
-import akka.stream.testkit.TestUtils.temporaryServerAddress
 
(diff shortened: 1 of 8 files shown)


akka-stream/src/main/scala/akka/stream/impl/io/TcpConnectionStream.scala
```diff
@@ -214,9 +224,20 @@ private[akka] abstract class TcpStreamActor(val settings: ActorFlowMaterializerS
     case Aborted            ⇒ fail(new StreamTcpException("The connection has been aborted"))
   }
 
+  def handleSubscriptionTimeout: Receive = {
+    case SubscriptionTimeout ⇒
+      val millis = settings.subscriptionTimeoutSettings.timeout.toMillis
+      if (!primaryOutputs.demandAvailable) {
```
drewhk: yeah, I know, this is not the same as not being subscribed. I will revisit this when I take the "real" ticket: https://github.com/akka/akka/issues/17789

drewhk:
Fixes #17786

akka-stream/src/main/scala/akka/stream/impl/io/TcpConnectionStream.scala
```diff
@@ -216,7 +217,14 @@ private[akka] abstract class TcpStreamActor(val settings: ActorFlowMaterializerS
 
   final override def receive = new ExposedPublisherReceive(activeReceive, unhandled) {
     override def receiveExposedPublisher(ep: ExposedPublisher): Unit = {
+      import context.dispatcher
       primaryOutputs.subreceive(ep)
+      subscriptionTimer = Some(
+        context.system.scheduler.scheduleOnce(
+          settings.subscriptionTimeoutSettings.timeout,
```
ktoso: So we have a setting like this... yet https://github.com/akka/akka/blob/release-2.3-dev/akka-stream/src/main/scala/akka/stream/actor/ActorPublisher.scala#L131 does not use it, it defaults to `Inf`, should it default to this value instead?

ktoso: (Could be a separate ticket)

drewhk: but that is ActorPublisher, it has nothing to do with pump style actors.

drewhk: Ah you mean, that does not use it, either. Anyway we pushed the whole issue to 1.x, so I will not touch it.

ktoso: Ah, okey :+1:

drewhk:
Interesting why that test fails.

akka-stream-tests/src/test/scala/akka/stream/io/TcpSpec.scala
```diff
@@ -470,6 +474,38 @@ class TcpSpec extends AkkaSpec("akka.io.tcp.windows-connection-abort-workaround-
...
+      val total = folder.run()
+      val rejected = folder.run()
+
+      Await.result(total, 3.seconds) should ===(1000)
+
+      a[StreamTcpException] should be thrownBy {
+        Await.result(rejected, 5.seconds) should ===(1000)
+      }
+    }
+
```
2m: These two tests cases are missing an `unbind` call.

2m:
LGTM after `unbind` is called on tcp bindings.

akka-stream-tests/src/test/scala/akka/stream/io/TcpSpec.scala
drewhk: No need to, because of the take(1) on the bind stream, it is cancelled anyway. If there is a leak because of the failed test, then the actor system shutdown will take care of it at the end of the Spec.

2m: I see, okay.

Issue #16394, opened 2014-11-26
Add `recover` operation


sirthias:
Recovering from stream failures with

    Flow[A, B]::recover[C >: B](PartialFunction[Throwable, C]): Flow[A, C]

appears to be a basic operation that is still missing.

rkuhn:
We also need its big brother (corrected signature from @smlin above):
~~~scala
def recoverWith[C :> B](PartialFunction[Throwable, Source[C]): Flow[A, C]
~~~

ktoso:
RecoverWith ticket: https://github.com/akka/akka/issues/18045

PR #17795, opened 2015-06-21
+str #16394 recover operation

agolubev:
Please verify if approach is Ok. I will need to add Java API,  docs and more tests for Scala and Java


ktoso:
OK TO TEST

agolubev:
And again cannot find that the cause is because of my commit. Is there any ticket about removing stack traces from test output?

ktoso:
Failure was `TlsSpec`, so it's unrelated - we're working on fixing TlsSpec but it's a tricky beast.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -56,6 +57,37 @@ private[akka] final case class Collect[In, Out](decider: Supervision.Decider)(pf
...
+
+  override def onPush(elem: T, ctx: Context[T]): SyncDirective = {
+    ctx.push(elem)
+  }
+
+  override def onPull(ctx: Context[T]): SyncDirective =
+    if (recovered.isDefined) {
+      val value = recovered.get
+      recovered = None
+      if (ctx.isFinishing) ctx.pushAndFinish(value) else ctx.push(value)
```
drewhk: it is always in finishing, since recovered is only set to False in onUpstreamFailure

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -56,6 +57,37 @@ private[akka] final case class Collect[In, Out](decider: Supervision.Decider)(pf
...
+private[akka] final case class Recover[T](pf: PartialFunction[Throwable, T]) extends PushPullStage[T, T] {
+  import Collect.NotApplied
+  var recovered: Option[T] = None
+
+  override def onPush(elem: T, ctx: Context[T]): SyncDirective = {
+    ctx.push(elem)
+  }
+
+  override def onPull(ctx: Context[T]): SyncDirective =
+    if (recovered.isDefined) {
```
drewhk: I would prefer a match over the Option

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -56,6 +57,37 @@ private[akka] final case class Collect[In, Out](decider: Supervision.Decider)(pf
...
+    pf.applyOrElse(t, NotApplied) match {
+      case NotApplied ⇒ ctx.fail(t)
+      case result: T @unchecked ⇒
+        recovered = Some(result)
+        ctx.absorbTermination()
+    }
+  }
+
+  override def onUpstreamFinish(ctx: Context[T]): TerminationDirective =
+    if (recovered.isEmpty) ctx.finish()
```
drewhk: it can only be empty, since it either finishes with a complete, or a failure.

drewhk:
LGTM apart from small nitpicks and some missing tests.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -350,6 +350,8 @@ trait FlowOps[+Out, +Mat] {
 
   private final val _identity = (x: Any) ⇒ x
 
+  def recover[T](pf: PartialFunction[Throwable, T]): Repr[T, Mat] = andThen(Recover(pf.asInstanceOf[PartialFunction[Any, Any]]))
```
patriknw: one important (non-obvious) part that must be documented is that the recovery element may overtake other normally emitted elements, since onError is not driven by demand

drewhk: Ah, yes, +1! Using this stage does not prevent losing messages that arrived *before* onError.

agolubev:
Ok. Added tests, Java tests, Java API to Flow and Source. Rebased and squashed

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -415,6 +416,23 @@ class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Graph
   def dropWhile(p: function.Predicate[Out]): javadsl.Flow[In, Out, Mat] = new Flow(delegate.dropWhile(p.test))
 
   /**
+   * Recover allows to send last element on failure and gracefully complete the stream
+   * Using recover does not prevent from losing messages that arrived before onError
+   * as they pass through this processing step.
```
drewhk: What do you mean by "as they pass through"? I think this should be reformulated to something like this:

"Since the underlying failure signal onError arrives out-of-band, it might jump over existing elements. This stage can recover the failure signal, but not the skipped elements, which will be dropped."

drewhk:
after final doc nitpick, it is ready to merge! Great job @agolubev! 

Btw, I don't know if you know gitter (it is a chat app, and you can sign-in using your github account), but there is an akka/dev channel if you want to discuss development related topics with the team.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -351,6 +351,22 @@ trait FlowOps[+Out, +Mat] {
...
+   * '''Emits when''' element is available from the upstream or upstream is failed and pf returns an element
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes or upstream failed with exception pf can handle
+   *
+   * '''Cancels when''' downstream cancels
+   *
+   */
+  def recover[T >: Out](pf: PartialFunction[Throwable, T]): Repr[T, Mat] = andThen(Recover(pf.asInstanceOf[PartialFunction[Any, Any]]))
```
ktoso: Wouldn't it be cool to be able to `recover` with a `Source[T]` instead of just T?

ktoso: i.e. http://reactivex.io/documentation/operators/catch.html

drewhk: yes, but that is not as simple here.

agolubev:
Thanks for pointing to gitter. Will use it.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
hepin1989: ```recoverWith```?

agolubev: RecoverWith source  can be constructed with FlexiMerge junction. It will continue from alternative source after error.
Would rather have this as a separate ticket.

ktoso: Adding a `recoverWith` sounds good to me. 

The thought I had is that it would be very simple to implement `recover` in terms of `recoverWith` (i.e. `recoverWith(Source.single(a))`) - much fundamental.

drewhk: Yep, except that you need to carefully clear demand, then prefetch again from the new source, while hopefully not receiving stray messages from the previous Publisher. And of course proper handling of re-subscription, etc. So it is not trivial.

drewhk:
needs a rebase. This probably will make into 1.1, we will cut RC4 today.

agolubev:
It was rebased actually yesterday (June 23). 
Could you shed some light on releasing process.. What is the branch for 1.1 (is it still release-2.3-dev but after build is made?) and where I can check when it's released (is it this page https://github.com/akka/akka/releases ?)

patriknw:
LGTM, thanks for contributing

patriknw:
Refs #16394

ktoso:
LGTM, I'd like the `recoverWith` as well but that can be done in the future. Thanks for contributing!
Sadly it seems it needs a rebase again.

drewhk:
Keep the PR open, and we will merge it after 1.0 comes out. Currently we have to hold this back, but it will make it into 1.1. The branch is the same, and I am happy to rebase it for you when the time is right.

ktoso:
I'll rebase it

ktoso:
Rebased it https://github.com/akka/akka/issues/17991

Issue #17261, opened 2015-04-22
Shard actor should only be persistent when rememberEntries=true


patriknw:
The Shard actor (child of ShardRegion) is a PersistentActor for the rememberEntries feature. I think it should not be a persistent actor when the feature is disabled. E.g. depending on connectivity to the journal for recovery.

PR #17797, opened 2015-06-22
=cls #17261 Use persistent shard only when rememberEntities=true

patriknw:
The first commit is a trivial split of the sharding classes into separate files.

The second commit extracts the persistent parts of the `Shard` actor into a `PersistentShard` subclass, which is only used when rememberEntities is enabled.


ktoso:
LGTM

ktoso:
Shall we merge this one as well before M2 tomorrow? Looks good from my end.

patriknw:
Refs #17261

PR #17802, opened 2015-06-22
=str #16973: Stream TCP actor should watch IO actor and handle termiation

### LINKED ISSUE #16973: Watch the connection actor from the Stream actor

drewhk:
Fixes #16973

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/io/TcpSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/io/TcpSpec.scala
@@ -3,17 +3,18 @@
  */
 package akka.stream.io
 
+import akka.actor.{ ActorSystem, Kill }
 import akka.stream.scaladsl.Tcp.OutgoingConnection
 
 import scala.collection.immutable
 import scala.concurrent.{ Future, Await }
 import akka.io.Tcp._
 
-import akka.stream.BindFailedException
+import akka.stream.{ ActorFlowMaterializer, StreamTcpException, BindFailedException }
 
 import scala.concurrent.Await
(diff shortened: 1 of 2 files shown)


rkuhn:
LGTM

akka-stream-tests/src/test/scala/akka/stream/io/TcpSpec.scala
```diff
@@ -380,6 +381,23 @@ class TcpSpec extends AkkaSpec("akka.io.tcp.windows-connection-abort-workaround-
...
+    "handle when connection actor terminates unexpectedly" in {
+      import system.dispatcher
+
+      val serverAddress = temporaryServerAddress()
+      val binding = Tcp().bindAndHandle(Flow[ByteString], serverAddress.getHostName, serverAddress.getPort)
+
+      val result = Source.lazyEmpty[ByteString].via(Tcp().outgoingConnection(serverAddress)).runFold(0)(_ + _.size)
+
+      // Getting rid of existing connection actors by using a blunt instrument
+      system.actorSelection(akka.io.Tcp(system).getManager.path / "selectors" / "$a" / "*") ! Kill
```
rkuhn: Ah, forgot to say: this should be done in a separate ActorSystem if we ever run test cases in parallel.

drewhk: well, can't you annotate tests to not run in parallel?

drewhk: ah, I just add a system in scope, that is what you meant.

PR #17808, opened 2015-06-23
fix actorsystem shutdown in tests

drewhk:


drewhk:
(addendum of missed commit from previous merged PR)

PR #17819, opened 2015-06-24
=str Add missing abstract modifier to ActorPublisherMessage.Cancel

RichardBradley:
This marker base class is missing an "abstract" modifier.

This is causing "warning: match may not be exhaustive" for me in certain cases

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/actor/ActorPublisher.scala
+++ b/akka-stream/src/main/scala/akka/stream/actor/ActorPublisher.scala
@@ -52,7 +52,7 @@ object ActorPublisherMessage {
    * subscription.
    */
   final case object Cancel extends Cancel with NoSerializationVerificationNeeded
-  sealed class Cancel extends ActorPublisherMessage
+  sealed abstract class Cancel extends ActorPublisherMessage
 
   /**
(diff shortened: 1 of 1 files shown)


drewhk:
OK TO TEST

akka-stream/src/main/scala/akka/stream/actor/ActorPublisher.scala
```diff
@@ -52,7 +52,7 @@ object ActorPublisherMessage {
    * subscription.
    */
   final case object Cancel extends Cancel with NoSerializationVerificationNeeded
-  sealed class Cancel extends ActorPublisherMessage
+  sealed abstract class Cancel extends ActorPublisherMessage
```
patriknw: it looks like we are missing the Java API for `SubscriptionTimeoutExceeded`. Would you mind adding that also?
It should be similar to the below `cancelInstance`.
Please also move those close together with classes/objects.

RichardBradley: OK, have done.

patriknw:
LGTM, thanks

drewhk:
Thanks, merging.

PR #17823, opened 2015-06-24
=clu #16890 Add missing barrier in LeaderDowningNodeThatIsUnreachableSpec

patriknw:


patriknw:
Refs #16890

patriknw:
I don't think there will be any objections here

PR #17824, opened 2015-06-24
Fix Http client cancellation

drewhk:
- fixed splitWhen cancellation
- added eagerCancel mode to Broadcast
- made Http use eagerCancel in client
- removed wrong GraphJunctionAtrributesSpec that fails always


akka-http-core/src/test/scala/akka/http/impl/engine/client/ClientCancellationSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+    "support cancellation in pooled outgoing connection" in {
+      testCase(
+        Flow[HttpRequest]
+          .map((_, ()))
+          .via(Http().cachedHostConnectionPool(address.getHostName, address.getPort)(noncheckedMaterializer))
+          .map(_._1.get))
+    }
+
+    "support cancellation in simple outgoing connection with TLS" in {
+      pending
```
drewhk: I cannot figure out how to make the HTTPS server to work, hence pending

sirthias:
Nice. LGTM.
I like the `eagerClose` flag.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSplitWhenSpec.scala
```diff
@@ -114,6 +114,72 @@ class FlowSplitWhenSpec extends AkkaSpec {
...
+      inputs3.expectNoMsg(1.second)
+
+      substream4.cancel()
+      inputs3.expectNoMsg(1.second)
+      masterStream3.expectNoMsg(1.second)
+
+      masterStream3.cancel()
+      inputs3.expectCancellation()
+
+    }
```
rkuhn: shouldn’t these be separate test cases?

drewhk: They all test the same thing

drewhk: I just tried many variations to be sure and wanted to keep all.

rkuhn:
LGTM

drewhk:
wonderful :)

Issue #17821, opened 2015-06-24
binary incompatibility in release-2.3 branch


patriknw:
Perhaps this is already fixed?

```[info] akka-actor: found 1 potential binary incompatibilities
[error]  * method toString()java.lang.String in object akka.japi.Pair does not have a correspondent in new version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("akka.japi.Pair.toString")```

https://jenkins.akka.io:8498/job/akka-release-2.3/200/consoleFull

ktoso:
Weird that validation did not show it hmm... I added the ignore to master,
please add on release-2.3

PR #17825, opened 2015-06-24
=clu #16890 Add missing barrier in LeaderDowningNodeThatIsUnreachableSpec

patriknw:
for 2.3

also fixing #17821 Skip toString bin compat complaint


patriknw:
Refs #16890

Issue #15640, opened 2014-08-07
Nested "persists" fail silently


DomBlack:
I accidentally ended up having nested persists within my Processor. My tests where failing due to timeouts, and it took me a while to figure out. It appears if you nest a persist within another persist's handler, it will silently fail.

Either this behaviour should be supported or an `IllegalStateException` should be thrown. The documentation isn't clear either way on what the intended behaviour should be.

Test Actor (added within PersistentActorSpec)
-----------------------------------------------------------
```scala
class NestedPersistProcessor(name: String) extends ExamplePersistentActor(name) {
    val receiveCommand: Receive = commonBehavior orElse {
      case Cmd(data) ⇒
        persist(Evt(s"${data}-a")) { event =>
          updateState(event)
          persist(Evt(s"${data}-b")) { _ =>
            updateState(event)
          }
        }
    }
  }
```

Test Method
-----------------
```scala
"nested persists works" in {
      val p = system.actorOf(Props(classOf[NestedPersistProcessor], "NestedTest"))
      p ! Cmd("1")
      p ! GetState
      expectMsg(List("1-a", "1-b"))
    }
```

Result of Test
-------------------
```
assertion failed: timeout (10 seconds) during expectMsg while waiting for List(1-a, 1-b)
```

ktoso:
@DomBlack this has been fixed - nesting persists now works (`2.4-M2`), and we're releasing right now.

PR #17828, opened 2015-06-25
~per #15640 support nested persist calls 

ktoso:
It is safe and "makes sense" to have nested persist calls, so that's supported now.
It is a bit confusing with nesting persistAsync, and while it can be made work I think it would require some more work, so I'm not doing that now - instead explicitly throwing if this is detected. If that's reasonable I'll add docs about it.

Refs #15640

### DIFF
--- a/akka-docs/rst/java/persistence.rst
+++ b/akka-docs/rst/java/persistence.rst
@@ -273,6 +273,41 @@ of the command for which this ``defer`` handler was called.
   The callback will not be invoked if the actor is restarted (or stopped) in between the call to
   ``defer`` and the journal has processed and confirmed all preceding writes.
 
+.. _nested-persist-calls-java:
+
+Nested persist calls
+--------------------
+It is possible to call ``persist`` and ``persistAsync`` inside their respective callback blocks and they will properly
+retain both the thread safety (including the right value of ``sender()``) as well as stashing guarantees.
+
+In general it is encouraged to create command handlers which do not need to resort to nested event persisting,
+however there are situations where it may be useful. It is important to understand the ordering of callback execution in
+those situations, as well as their implication on the stashing behaviour (that ``persist()`` enforces). In the following
+example two persist calls are issued, and each of them issues another persist inside its callback:
+
+.. includecode:: code/docs/persistence/PersistenceDocTest.java#nested-persist-persist
+
+When sending two commands to this ``PersistentActor``, the persist handlers will be executed in the following order:
+
+.. includecode:: code/docs/persistence/PersistenceDocTest.java#nested-persist-persist-caller
+
+First the "outer layer" of persist calls is issued and their callbacks applied, after these have successfully completed
+the inner callbacks will be invoked (once the events they are persisting have been confirmed to be persisted by the journal).
+And only after all these handlers have been successfully invoked, the next command will delivered to the persistent Actor.
+In other words, the stashing of incoming commands that is guaranteed by initially calling ``persist()`` on the outer layer
+is extended until all nested ``persist`` callbacks have been handled.
+
+It is also possible to nest ``persistAsync`` calls, using the same pattern:
+
+.. includecode:: code/docs/persistence/PersistenceDocTest.java#nested-persistAsync-persistAsync
+
+In this case no stashing is happening, yet the events are still persisted and callbacks executed in the expected order:
+
+.. includecode:: code/docs/persistence/PersistenceDocTest.java#nested-persistAsync-persistAsync-caller
+
+While it is possible to nest mixed ``persist`` and ``persistAsync`` with keeping their respective semantics
+it is not a recommended practice as it may lead to overly complex nesting.
+
 Failures
 --------
 
--- a/akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
@@ -522,6 +522,44 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
     }
   }
 
+  private def flushBatch() {
+    def addToBatch(p: PersistentEnvelope): Unit = p match {
+      case a: AtomicWrite ⇒
+        journalBatch :+= a.copy(payload =
+          a.payload.map(_.update(persistenceId = persistenceId, sequenceNr = nextSequenceNr(), writerUuid = writerUuid)))
+      case r: PersistentEnvelope ⇒
+        journalBatch :+= r
+    }
+
+    def maxBatchSizeReached: Boolean =
+      journalBatch.size >= maxMessageBatchSize
+
+    // When using only `persistAsync` and `defer` max throughput is increased by using
+    // batching, but when using `persist` we want to use one atomic WriteMessages
+    // for the emitted events.
+    // Flush previously collected events, if any, separately from the `persist` batch
+    if (pendingStashingPersistInvocations > 0 && journalBatch.nonEmpty)
+      flushJournalBatch()
+
+    eventBatch.reverse.foreach { p ⇒
+      addToBatch(p)
+      if (!writeInProgress || maxBatchSizeReached) flushJournalBatch()
+    }
+
+    eventBatch = Nil
+  }
+
+  private def peekApplyHandler(payload: Any): Unit = {
+    val batchSizeBeforeApply = eventBatch.size
+    try pendingInvocations.peek().handler(payload)
+    finally {
+      val batchSizeAfterApply = eventBatch.size
+
+      if (batchSizeAfterApply > batchSizeBeforeApply)
+        flushBatch()
+    }
+  }
+
   /**
    * Common receive handler for processingCommands and persistingEvents
    */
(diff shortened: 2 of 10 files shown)


akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -511,6 +523,32 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
...
+  private def addToBatch(p: PersistentEnvelope): Unit = p match {
+    case p: PersistentRepr ⇒
+      journalBatch :+= p.update(persistenceId = persistenceId, sequenceNr = nextSequenceNr(), sender = sender())
+    case r: PersistentEnvelope ⇒
+      journalBatch :+= r
+  }
+
+  private def maxBatchSizeReached: Boolean =
+    journalBatch.size >= maxMessageBatchSize
+
```
patriknw: is this block identical to the removed block (just a move)?

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -538,7 +578,9 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
         // while message is in flight, in that case we ignore the call to the handler
         if (id == instanceId) {
           try {
+            insidePersistCallback = true
```
ktoso: TODO: this must be try/finally

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -538,7 +578,9 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
...
             pendingInvocations.peek().handler(l)
+            insidePersistCallback = false
```
patriknw: what happens if the handler throws, and resume supervision is used?
should this be done in a finally?

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -522,7 +560,9 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
         if (id == instanceId) {
           updateLastSequenceNr(p)
           try {
+            insidePersistCallback = true
             pendingInvocations.peek().handler(p.payload)
+            insidePersistCallback = false
```
patriknw: in finally?

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
ktoso: yeah, added a todo comment on github already :) must be in try/finally

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -306,6 +312,12 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
    * @param handler handler for each persisted `event`
    */
   final def persistAsync[A](event: A)(handler: A ⇒ Unit): Unit = {
+    if (insidePersistCallback) {
+      val ex = new UnsupportedOperationException("calling persistAsync from within the context of a persist/persistAsync callback is not supported!")
+      try onPersistFailure(ex, event, sequenceNr) finally context.stop(self)
```
patriknw: I don't think this is a persist failure, I think you can just throw the exception (similar to as if the handler throwed something)

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -5,6 +5,9 @@
 package akka.persistence
 
 import java.util.concurrent.atomic.AtomicInteger
+import akka.util.Reflect
```
patriknw: not used

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -5,6 +5,9 @@
...
+
+import scala.UnsupportedOperationException
```
patriknw: import not needed, I'd rather use the `java.lang.UnsupportedOperationException`, but I guess this is just an alias of that

akka-persistence/src/test/scala/akka/persistence/PersistentActorSpec.scala
```diff
@@ -885,6 +951,60 @@ abstract class PersistentActorSpec(config: Config) extends PersistenceSpec(confi
...
+      persistentActor ! "a"
+      persistentActor ! "b"
+
+      expectMsg("a")
+      expectMsg("a-outer")
+      expectMsg("a-inner")
+      // and only then process "b"
+      expectMsg("b")
+      expectMsg("b-outer")
+      expectMsg("b-inner")
```
patriknw: we must document this behavior, I guess it is a good behavior, but it was not obvious for me that it is like that

akka-persistence/src/test/scala/akka/persistence/PersistentActorSpec.scala
```diff
@@ -885,6 +951,60 @@ abstract class PersistentActorSpec(config: Config) extends PersistenceSpec(confi
...
+
+      expectMsg("a")
+      receiveN(6) should ===((1 to nestedPersists).map("a-" + _))
+      // and only then process "b"
+      expectMsg("b")
+      receiveN(6) should ===((1 to nestedPersists).map("b-" + _))
+    }
+    "not allow nesting persistAsync calls" in {
+      // currently disalow (loudly) calling persistAsync inside callbacks
+      // TODO rethink it again if it wouldn't be doable... may make sense if persisted data is CRDTs?
```
patriknw: what is it that breaks?

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
ktoso: I need to check what actually happens when the handler throws... Supervision should kick in then?
Needs to be explicitly documented as well I think "what happens when handler throws", can add this section during this PR

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
ktoso: thanks

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
ktoso: right, thanks

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
patriknw: yes, it should behave exactly as when exc from ordinary receive, and that is why we have those finally and catch and things to make sure the state is correct (especially important for resume)

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
ktoso: yes, it's just moved.
It was inside a `State`, but now it's possible to add to a batch from different places (because `persist()(persist)`)

akka-persistence/src/test/scala/akka/persistence/PersistentActorSpec.scala
ktoso: the callbacks didn't get called, did not dig deep enough - will investigate more now

akka-persistence/src/test/scala/akka/persistence/PersistentActorSpec.scala
ktoso: yes, glad you also find it logical - I'll write a section about nested persist calls, with ordering explained

akka-persistence/src/test/scala/akka/persistence/PersistentActorSpec.scala
```diff
@@ -891,6 +1017,88 @@ abstract class PersistentActorSpec(config: Config) extends PersistenceSpec(confi
...
+      val persistentActor = system.actorOf(Props(classOf[NestedPersistInAsyncEnforcesStashing], name, testActor))
+      persistentActor ! "a"
+
+      expectMsg("a")
+      expectMsg("a-outer-async")
+      expectMsg("a-inner")
+      persistentActor ! "b"
+      expectMsg("done")
+      // which means that b only got applied after the inner persist() handler finished
+      // so it keeps the persist() semantics, even though we should not recommend this style it can come in handy I guess
```
patriknw: cool

akka-docs/rst/java/persistence.rst
```diff
@@ -273,6 +273,41 @@ of the command for which this ``defer`` handler was called.
   The callback will not be invoked if the actor is restarted (or stopped) in between the call to
   ``defer`` and the journal has processed and confirmed all preceding writes.
 
+.. _nested-persist-calls-java:
+
```
patriknw: there is a lambda-persistence.rst also

patriknw:
LGTM after lambda-persistence.rst
Very good docs and tests! Great work.

ktoso:
Failure was unrelated https://github.com/akka/akka/issues/17848

PR #17830, opened 2015-06-25
=str: Materializer should report being shut down on sys termination

drewhk:
Minor tweak to report being shut down even if it was not via the shutdown() method.


ktoso:
LGTM

patriknw:
LGTM

PR #17831, opened 2015-06-25
Introduces `fold` as a Flow transformation and generalizes Sink.fold …

viktorklang:
I think there's quite a few use cases for running a fold and then building on top of the result of that, perhaps splitting or "decompressing" the folded result value into new flow elements downstream.

This PR also includes a few other improvements.


akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -74,7 +74,7 @@ private[akka] final object Collect {
   final val NotApplied: Any ⇒ Any = _ ⇒ Collect.NotApplied
 }
 
-private[akka] final case class Collect[In, Out](decider: Supervision.Decider)(pf: PartialFunction[In, Out]) extends PushStage[In, Out] {
+private[akka] final case class Collect[In, Out](pf: PartialFunction[In, Out], decider: Supervision.Decider) extends PushStage[In, Out] {
```
viktorklang: This change is to harmonize with the definition and parameter ordering of the other Stage definitions.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -107,37 +103,8 @@ object Sink extends SinkApply {
...
-        }
-      }
-
-      (stage, promise.future)
-    }
-
-    Flow[T].transformMaterializing(newForeachStage).to(Sink.ignore).named("foreachSink")
-  }
+  def foreach[T](f: T ⇒ Unit): Sink[T, Future[Unit]] =
+    Flow[T].map(f).toMat(Sink.ignore)(Keep.right).named("foreachSink")
```
viktorklang: I made this change for the simple reason that I really, really, do not like custom stages since the materializer has no clue what the operation does. With map + ignore at least we can have the Materializer see that and optimize accordingly.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -151,11 +118,9 @@ object Sink extends SinkApply {
    * element is dropped and the stream continues.
    *
    * @see [[#mapAsyncUnordered]]
-   */
+   */ // FIXME Why is this method even useful?
```
viktorklang: Sorry for adding this comment, but it was the only way for me to remember to write this comment:

What is the purpose of having this operation as a prime API construct when `mapAsyncUnordered` + `Sink.ignore` is expressive enough for the (unknown) use-case?

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
```diff
@@ -299,34 +300,34 @@ private[akka] object ActorProcessorFactory {
     // USE THIS TO AVOID CLOSING OVER THE MATERIALIZER BELOW
     // Also, otherwise the attributes will not affect the settings properly!
     val settings = materializer.effectiveSettings(att)
+    def interp(s: Stage[_, _]): (Props, Unit) = (ActorInterpreter.props(settings, List(s), materializer, att), ())
```
viktorklang: A bit of Boy Scouting, I wanted to remove the visual clutter.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -164,41 +129,8 @@ object Sink extends SinkApply {
...
-        }
-      }
-
-      (stage, promise.future)
-    }
-
-    Flow[T].transformMaterializing(newFoldStage).to(Sink.ignore).named("foldSink")
-  }
+  def fold[U, T](zero: U)(f: (U, T) ⇒ U): Sink[T, Future[U]] =
+    Flow[T].fold(zero)(f).toMat(Sink.head)(Keep.right).named("foldSink")
```
viktorklang: Another custom stage removed in favor of a named one.

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
drewhk: thanks

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
drewhk: good call, ignore returning a future enables this nicely.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
drewhk: because users asked for it. There will be always a pressure between keeping things minimal, and having convenience methods. The usual trend is that users like the convenience methods, so we add a couple of them if there are enough people asking.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -151,11 +118,9 @@ object Sink extends SinkApply {
    * element is dropped and the stream continues.
    *
    * @see [[#mapAsyncUnordered]]
-   */
+   */ // FIXME Why is this method even useful?
   def foreachParallel[T](parallelism: Int)(f: T ⇒ Unit)(implicit ec: ExecutionContext): Sink[T, Future[Unit]] =
-    Flow[T].andThen(
-      MapAsyncUnordered(parallelism,
-        { out: T ⇒ Future(f(out)) }.asInstanceOf[Any ⇒ Future[Unit]])).toMat(Sink.ignore)(Keep.right)
+    Flow[T].mapAsyncUnordered(parallelism)(t ⇒ Future(f(t))).toMat(Sink.ignore)(Keep.right)
```
drewhk: nice

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -542,6 +542,25 @@ trait FlowOps[+Out, +Mat] {
...
   /**
+   * Similar to `scan` but only emits its result when the upstream completes,
+   * after which it also completes. Applies the given function towards its current and next value,
+   * yielding the next current value.
+   *
+   * If the function `f` throws an exception and the supervision decision is
+   * [[akka.stream.Supervision.Restart]] current value starts at `zero` again
+   * the stream will continue.
+   *
+   * '''Emits when''' the function scanning the element returns a new element
```
drewhk: emits when upstream completes (copy pasta from scan)

drewhk: please also add this to the reference table in the documentation

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -542,6 +542,25 @@ trait FlowOps[+Out, +Mat] {
   def scan[T](zero: T)(f: (T, Out) ⇒ T): Repr[T, Mat] = andThen(Scan(zero, f.asInstanceOf[(Any, Any) ⇒ Any]))
 
   /**
+   * Similar to `scan` but only emits its result when the upstream completes,
+   * after which it also completes. Applies the given function towards its current and next value,
+   * yielding the next current value.
+   *
+   * If the function `f` throws an exception and the supervision decision is
+   * [[akka.stream.Supervision.Restart]] current value starts at `zero` again
+   * the stream will continue.
```
drewhk: maybe have a resume that skips the failed element?

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
drewhk: (accepting connections is one of the use cases for this)

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
drewhk: good!

drewhk:
LGTM after docs are corrected, JavaDSL is harmonized, and foreachParallel left unharmed ;)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: Great catch, thanks, will fix. Which table are you referring to?

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
viktorklang: Thanks, yes I think so too :)

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
viktorklang: Ok. I won't argue given the very simple implementation :)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: This is the one: https://github.com/akka/akka/blob/release-2.3-dev/akka-docs-dev/rst/stages-overview.rst

viktorklang: perfect, thanks!

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
drewhk: Yeah, so the most useful application is with TCP (or HTTP)

    Tcp().bind(...).foreachParallel(4) { connection => connection.join(handler).run() }

But there might be other uses. Since users might not immediately see that mapAsyncUnordered is perfect for this, (given its signature that says Future, and they might not see how Futures are involved in this case), I think this is a little help that we can give them to use the right tool.

viktorklang:
Comments addessed!

drewhk:
Maybe try out Resume with Fold? I don't remember but it might work out of the box, but at least a test case should be there.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: Ideally that should be configured on a case-by-case basis. Given the correspondence between scan and fold I think whatever we pick needs to be consistent between the two.

drewhk: Yes, this is configurable, you can set the supervision strategy stage by stage in the user API. My question is whether these stages (scan/fold) support the Resume strategy right now (maybe they do, but I don't remember)

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
viktorklang: I guess the downside is that using this means that it's hard to do an orderly shutdown because the termination of foreachParallel does not imply that the things being `run()` have terminated?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: Alright, I'll take it for a spin in the REPL and see what happens if Resume is used with Fold/Scan

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
drewhk: this is similar to foreach, whatever task foreach launched asynchronously in the background is out of scope from our shutdown.

In the case of a connection stream, if you terminate the bind-actor, then it terminates its accepted connections. If you just cancel the bind-stream, it will keep existing children around, and shut down the actor when all the children are gone. Unsubscribed children stop themselves after 5 seconds timeout, or when the parent actor stops.

viktorklang: Yeah, which is why I prefer to run a Unit-fold iso foreach :)

viktorklang:
Fixed the DSL inconsistencies and added the missing materialization of Take and Drop While.

viktorklang:
@drewhk PR validator fails on TcpSpec:

"[info] - must properly full-close if requested *** FAILED *** (80 milliseconds)
[info]   akka.stream.StreamTcpException: Connection failed."

Is this a known currently failing build thing?

drewhk:
No, I have only seen a bind failed before, but that looked like a fluke. Open a ticket. But what I am more worried about is one of the previous failures on this PR:

[info] ActorMaterializerSpec:
Coroner Thread Count starts at 16 in akka.stream.ActorMaterializerSpec
[info] ActorMaterializer
[info] - must report shutdown status properly (2 milliseconds)
[info] - must properly shut down actors associated with it *** FAILED *** (3 seconds, 31 milliseconds)
[info]   Expected exception akka.stream.AbruptTerminationException to be thrown, but java.util.concurrent.TimeoutException was thrown. (ActorMaterializerSpec.scala:24)

I think that might be actually triggered by your change to fold.

viktorklang:
@drewhk I'll dig a bit deeper into this

viktorklang:
TcpSpec looks a bit weird as it is not using any timeout dilation also.

drewhk:
Have you found the issue? I have unfortunately no spare time to look into this right now, but this is seemingly related to when the actor below the interpreter is pulled out unexpectedly. There are already improvements in place to handle the materialized value in these cases, but you delegated that task now to Sink.head, which probably causes another edge-case that was not forseen. 

This is the place where these kind of errors are handled in the ActorInterpreter:
https://github.com/akka/akka/blob/release-2.3-dev/akka-stream/src/main/scala/akka/stream/impl/fusing/ActorInterpreter.scala#L374

The idea is basically the following:
 - if the actor is stopped, we first try to invoke the interpreter with an AbruptTermination failure. If the interpreter stages have been closed already, this has no effect, since there is a flag against doubly invoking onError on the stages
 - the above step is to ensure that if the plug has been pulled on the actor unexpectedly (spec violation throws, actorsystem stopped, etc.) then it gives a chance to the stages to deal with the materialized values (failing the future of a Fold for example)
 - If the above step fails, or the error propagation stucks, the finally block tries to signal failure to the downstream Subscriber anyway. Again, this is guarded by a flag, so if the error (or normal completion) has already been reported, this does nothing. It does ensure though, that it closes the downstream publisher no matter what, even if the interpreter failed to signal the completion/failure. Since the actor is going down, the interpreter will not get a second chance, so we must do this to cover all bases.

I tried to make this work in various error scenarios, but you might have tripped something unexpected.

viktorklang:
@drewhk Haven't had time to check, I'm in meetings for 5h straight. It does seem like it is some sort of weird edge case because there's nothing magic happening in the impl of Fold.

drewhk:
Yeah, the difference is that now it is no longer Fold that handles the materialized value, but it passes this responsibility to Sink.head.

viktorklang:
@drewhk Seems like a bug to me :)

drewhk:
Yes, exactly :) And you have the honor of fixing it :)

viktorklang:
Alright, I've done quite some tests and folding itself looks like it works. So it does have to be some sort of cancellation/onError thingydevice going on as these pass on my box:

~~~scala
    val input = 1 to 100
    val expected = input.fold(0)(_ + _)
    val inputSource = Source(input).filter(_ ⇒ true).map(identity)
    val foldSource = inputSource.fold[Int](0)(_ + _).filter(_ ⇒ true).map(identity)
    val foldFlow = Flow[Int].filter(_ ⇒ true).map(identity).fold(0)(_ + _).filter(_ ⇒ true).map(identity)
    val foldSink = Sink.fold[Int, Int](0)(_ + _)

    "work when using Source.runFold" in assertAllStagesStopped {
      Await.result(inputSource.runFold(0)(_ + _), 3.seconds) should be(expected)
    }

    "work when using Source.fold" in assertAllStagesStopped {
      Await.result(foldSource runWith Sink.head, 3.seconds) should be(expected)
    }

    "work when using Sink.fold" in assertAllStagesStopped {
      Await.result(inputSource runWith foldSink, 3.seconds) should be(expected)
    }

    "work when using Flow.fold" in assertAllStagesStopped {
      Await.result(inputSource via foldFlow runWith Sink.head, 3.seconds) should be(expected)
    }

    "work when using Source.fold + Flow.fold + Sink.fold" in assertAllStagesStopped {
      Await.result(foldSource via foldFlow runWith foldSink, 3.seconds) should be(expected)
    }
~~~

drewhk:
I am pretty sure fold works, but I am not sure why providing the materialized future from a non-actor based stage fails the test.

viktorklang:
@drewhk Yes, that is a very good question. I am trying to follow that logic but to be honest—it ain't super-readable.

viktorklang:
The plot thickens, this is the exception that the Future gets:

"[info]   java.lang.AssertionError: assertion failed: expected no StreamSupervisor children, but got [Actor[akka://TcpSpec/user/$a/flow-15-2-fold#-1328863943]]"

drewhk:
Thats from the asserAllStagesStopped in the test. How does it end up in the
future? Are you sure it comes from the future and not the assertion?
2015. jún. 25. 23:05 ezt írta ("Viktor Klang (√)" <notifications@github.com

viktorklang:
@drewhk Doh! I was too generous with the trying and catching.

What actually happens is that the Future gets completed with a "class akka.stream.StreamTcpException" which says "Connection failed."

drewhk:
Er, which test are we talking about? I was talking about ActorMaterializerSpec, but you seem to talk about TcpSpec now. The ActorMaterializerSpec does not touch any TCP code.

viktorklang:
Yeah, If you loook at https://github.com/akka/akka/pull/17831#issuecomment-115237966 that is what I have been looking at :)

drewhk:
Ok, and if you look at my comment, I was thinking that you are doing this:

> No, I have only seen a bind failed before, but that looked like a fluke. Open a ticket. But what I am more worried about is one of the previous failures on this PR:
>
> [info] ActorMaterializerSpec:

viktorklang:
@drewhk I get TcpSpec to consistently fail with that error on my box.

I'm having a stab at ActorMaterializerSpec :)

drewhk:
Ok, then fix TcpSpec first. I thought it is a spurious faiolure?

viktorklang:
@drewhk No, it's pretty consistent :) But I am not sure it's a problem with Fold as it is a connection refused problem...

drewhk:
that is very likely not a fold problem

drewhk:
So, what you should test, is if the fold actually getting all the data (i.e. put a logging stage before) just the connection finishes abruptly. Because the "fullClose" mode in TCP waits for the ack of the last write, but then asks the connection to close fully, and it might be non-reliable in practice (i.e. it might do a connection reset instead)

viktorklang:
@drewhk I think the ActorMaterializerSpec may be completely unrelated to Fold:

This still fails:


    "properly shut down actors associated with it" in {
      val m = ActorMaterializer.create(system)

      val f = Source.single(1).map({ x ⇒ m.shutdown(); x }).runWith(Sink.ignore)(m)

      an[AbruptTerminationException] should be thrownBy Await.result(f, 3.seconds)
    }

[info] - must properly shut down actors associated with it *** FAILED *** (239 milliseconds)
[info]   Expected exception akka.stream.AbruptTerminationException to be thrown, but no exception was thrown. (ActorMaterializerSpec.scala:24)
[info]   at org.scalatest.Assertions$class.newAssertionFailedException(Assertions.scala:495)
[info]   at org.scalatest.Assertions$.checkExpectedException(Assertions.scala:1582)
[info]   at org.scalatest.words.ResultOfBeWordForAnType.thrownBy(ResultOfBeWordForAnType.scala:39)
[info]   at akka.stream.ActorMaterializerSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply$mcV$sp(ActorMaterializerSpec.scala:29)
[info]   at akka.stream.ActorMaterializerSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply(ActorMaterializerSpec.scala:24)
[info]   at akka.stream.ActorMaterializerSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply(ActorMaterializerSpec.scala:24)
[info]   at org.scalatest.Transformer$$anonfun$apply$1.apply$mcV$sp(Transformer.scala:22)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.WordSpecLike$$anon$1.apply(WordSpecLike.scala:953)
[info]   at org.scalatest.Suite$class.withFixture(Suite.scala:1122)
[info]   at akka.stream.testkit.AkkaSpec.withFixture(AkkaSpec.scala:55)
[info]   at org.scalatest.WordSpecLike$class.invokeWithFixture$1(WordSpecLike.scala:950)
[info]   at org.scalatest.WordSpecLike$$anonfun$runTest$1.apply(WordSpecLike.scala:962)
[info]   at org.scalatest.WordSpecLike$$anonfun$runTest$1.apply(WordSpecLike.scala:962)
[info]   at org.scalatest.SuperEngine.runTestImpl(Engine.scala:306)
[info]   at org.scalatest.WordSpecLike$class.runTest(WordSpecLike.scala:962)
[info]   at akka.stream.testkit.AkkaSpec.runTest(AkkaSpec.scala:55)
[info]   at org.scalatest.WordSpecLike$$anonfun$runTests$1.apply(WordSpecLike.scala:1021)
[info]   at org.scalatest.WordSpecLike$$anonfun$runTests$1.apply(WordSpecLike.scala:1021)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:413)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:401)
[info]   at scala.collection.immutable.List.foreach(List.scala:318)
[info]   at org.scalatest.SuperEngine.traverseSubNodes$1(Engine.scala:401)
[info]   at org.scalatest.SuperEngine.org$scalatest$SuperEngine$$runTestsInBranch(Engine.scala:390)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:427)
[info]   at org.scalatest.SuperEngine$$anonfun$traverseSubNodes$1$1.apply(Engine.scala:401)
[info]   at scala.collection.immutable.List.foreach(List.scala:318)
[info]   at org.scalatest.SuperEngine.traverseSubNodes$1(Engine.scala:401)
[info]   at org.scalatest.SuperEngine.org$scalatest$SuperEngine$$runTestsInBranch(Engine.scala:396)
[info]   at org.scalatest.SuperEngine.runTestsImpl(Engine.scala:483)
[info]   at org.scalatest.WordSpecLike$class.runTests(WordSpecLike.scala:1021)
[info]   at akka.stream.testkit.AkkaSpec.runTests(AkkaSpec.scala:55)
[info]   at org.scalatest.Suite$class.run(Suite.scala:1424)
[info]   at akka.stream.testkit.AkkaSpec.org$scalatest$WordSpecLike$$super$run(AkkaSpec.scala:55)
[info]   at org.scalatest.WordSpecLike$$anonfun$run$1.apply(WordSpecLike.scala:1067)
[info]   at org.scalatest.WordSpecLike$$anonfun$run$1.apply(WordSpecLike.scala:1067)
[info]   at org.scalatest.SuperEngine.runImpl(Engine.scala:545)
[info]   at org.scalatest.WordSpecLike$class.run(WordSpecLike.scala:1067)
[info]   at akka.stream.testkit.AkkaSpec.org$scalatest$BeforeAndAfterAll$$super$run(AkkaSpec.scala:55)
[info]   at org.scalatest.BeforeAndAfterAll$class.liftedTree1$1(BeforeAndAfterAll.scala:257)
[info]   at org.scalatest.BeforeAndAfterAll$class.run(BeforeAndAfterAll.scala:256)
[info]   at akka.stream.testkit.AkkaSpec.run(AkkaSpec.scala:55)
[info]   at org.scalatest.tools.Framework.org$scalatest$tools$Framework$$runSuite(Framework.scala:466)
[info]   at org.scalatest.tools.Framework$ScalaTestTask.execute(Framework.scala:677)
[info]   at sbt.TestRunner.runTest$1(TestFramework.scala:84)
[info]   at sbt.TestRunner.run(TestFramework.scala:94)
[info]   at sbt.TestFramework$$anon$2$$anonfun$$init$$1$$anonfun$apply$8.apply(TestFramework.scala:219)
[info]   at sbt.TestFramework$$anon$2$$anonfun$$init$$1$$anonfun$apply$8.apply(TestFramework.scala:219)
[info]   at sbt.TestFramework$.sbt$TestFramework$$withContextLoader(TestFramework.scala:207)
[info]   at sbt.TestFramework$$anon$2$$anonfun$$init$$1.apply(TestFramework.scala:219)
[info]   at sbt.TestFramework$$anon$2$$anonfun$$init$$1.apply(TestFramework.scala:219)
[info]   at sbt.TestFunction.apply(TestFramework.scala:224)
[info]   at sbt.Tests$.sbt$Tests$$processRunnable$1(Tests.scala:211)
[info]   at sbt.Tests$$anonfun$makeSerial$1.apply(Tests.scala:217)
[info]   at sbt.Tests$$anonfun$makeSerial$1.apply(Tests.scala:217)
[info]   at sbt.std.Transform$$anon$3$$anonfun$apply$2.apply(System.scala:45)
[info]   at sbt.std.Transform$$anon$3$$anonfun$apply$2.apply(System.scala:45)
[info]   at sbt.std.Transform$$anon$4.work(System.scala:64)
[info]   at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:237)
[info]   at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:237)
[info]   at sbt.ErrorHandling$.wideConvert(ErrorHandling.scala:18)
[info]   at sbt.Execute.work(Execute.scala:244)
[info]   at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:237)
[info]   at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:237)
[info]   at sbt.ConcurrentRestrictions$$anon$4$$anonfun$1.apply(ConcurrentRestrictions.scala:160)
[info]   at sbt.CompletionService$$anon$2.call(CompletionService.scala:30)
[info]   at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[info]   at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[info]   at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[info]   at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[info]   at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[info]   at java.lang.Thread.run(Thread.java:744)

drewhk:
If you revert back to the original fold, then I am pretty sure it passes. As I said, I suppose the problem comes from Sink.ignore not properly finishing the future somehow. In the original code Fold handled the materialized Future directly, and now it does not. I think Sink.ignore should get an AbruptTerminationException, but it does not for some reason.

viktorklang:
@drewhk As you can see there is no fold involved and it still gets the failure.

drewhk:
yes, but there is Sink.ignore involved.

viktorklang:
@drewhk Absolutely, but this PR does nothing to Sink.ignore :)

viktorklang:
@drewhk So, if this PR somehow breaks the test case I showed (without using fold) and it doesn't mess with the implementation of Sink.ignore, then I think we have bigger problems on our hands (or I'm missing something important)

viktorklang:
@drewhk Which then, I think, points to https://github.com/akka/akka/commit/0fc30e63f5c3dabd70ac1f3032881d91e0477ea4

drewhk:
So you won't fix the bug then? I told you in the beginning that I expect this failure because of the changed behavior, where Sink.ignore handles the Future now, instead of the Fold stage. It still looks like that is exactly what happens. 

You should check:
 - If Sink.ignore gets the onError signal at all (just add a println) or any other completion signal
 - if it gets it, then why does the test fail.

drewhk:
I can take over this PR if you want.

viktorklang:
@drewhk I'm having a look at the bug as we speak. As I said, the curious issue is that the test fails without -any- usage of Fold, and Sink.ignore is not changed in this PR. I get an onComplete Signal and not the exception. But my assumption might be invalid, this is the modified test case:

~~~scala
"properly shut down actors associated with it" in {
  val m = ActorMaterializer.create(system)

  val f = Source.single(1).map({ x ⇒ m.shutdown(); x }).runWith(Sink.ignore)(m)

  an[AbruptTerminationException] should be thrownBy Await.result(f, 3.seconds)
}
~~~

drewhk:
> As I said, the curious issue is that the test fails without -any- usage of Fold

Which is not curious at all for me. The test originally used Fold, and this way, it implicitly relied on the ActorInterpreter postStop() logic. Since you changed the implementation, it uncovered a weird bug there.

viktorklang:
Haha, ok. :)

This passes:

~~~scala
    "properly shut down actors associated with it" in {
      val m = ActorMaterializer.create(system)

      val f = Source.lazyEmpty[Int].toMat(Sink.fold(0) { _ + _ })(Keep.right).run()(m)

      m.shutdown()

      an[AbruptTerminationException] should be thrownBy Await.result(f, 3.seconds)
    }
~~~

drewhk:
Wait a minute, of course you have an oncomplete signal in your version

drewhk:
the difference is betwween Source.lazyEmpty and Source.single. As soon as map emits the first element, it also has a shutdown in progress, but that does not affect Sink.ignore at all, which just gets the onComplete, not the exception!

drewhk:
I used Source.lazyEmpty exactly to not have a completion that is in race with the shutdown

viktorklang:
This also passes:

~~~scala
    "properly shut down actors associated with it" in {
      val m = ActorMaterializer.create(system)

      val f = Source.lazyEmpty[Int].fold(0)(_ + _).runWith(Sink.head)(m)

      m.shutdown()

      an[AbruptTerminationException] should be thrownBy Await.result(f, 3.seconds)
    }
~~~

drewhk:
This fails:

      val f = Source.lazyEmpty[Int].map(identity).runWith(Sink.ignore)(m)
      m.shutdown()

      an[AbruptTerminationException] should be thrownBy
        Await.result(f, 3.seconds)

drewhk:
So, the issue is that the guy is killed before the exposedPublisher variable is set, and therefore it cannot invoke shutdown on the Publisher.

drewhk:
basically postStop is called on the actor before it is aware of its own exposed Publisher instance, therefore its attempt to finish the downstream fails. And since the actor is stopped, there is also no way for the actor to correct this later.

drewhk:
This was not seen before because Fold was able to report the failure to its exposed Future even though the exposedPublisher was not set (and therefore failed to fail the downstream). Since now the Future is provided as a downstream of Fold, the onError signal is simply lost.

viktorklang:
@drewhk TBH it sounds like you may be the more appropriate person to tackle that problem. Let me know if you need my assistance with it. :/

drewhk:
Yes, this is way more internal than I hoped. I recommend to make the currently failed test pending, and open a ticket that I will assign to myself immediately. This PR looks correct, so I don't want to keep open unnecessarily.

viktorklang:
Only failure now is from ClientServerSpec which, from what I can see, has nothing to do with this PR (another Akka IO bug?)

drewhk:
That rather looks like a flaky test. I guess sometimes we get another error messge that we don't expected in the test.

viktorklang:
@drewhk I'm double-checking how exceptions are dealt with during materialization (which is what will happen in that test)

viktorklang:
@drewhk To me, the strategy for dealing with those exception seems fishy

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -635,9 +611,18 @@ private[stream] abstract class MaterializerSession(val topLevel: StreamLayout.Mo
       s"The top level module cannot be materialized because it has unconnected ports: ${(topLevel.inPorts ++ topLevel.outPorts).mkString(", ")}")
     try materializeModule(topLevel, topLevel.attributes)
     catch {
-      case NonFatal(e) ⇒
-        panic(e)
-        throw e
+      case NonFatal(cause) ⇒
+        // Cancels all intermediate Publishers and fails all intermediate Subscribers.
+        // (This is an attempt to clean up after an exception during materialization)
+        val panicError = new MaterializationPanic(cause)
```
viktorklang: @drewhk I embedded "panic" here since it wasn't used anywhere else. But to me, it's hard to reason about if and why it is correct.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -635,9 +611,18 @@ private[stream] abstract class MaterializerSession(val topLevel: StreamLayout.Mo
...
+        val errorPublisher = new ErrorPublisher(cause, "")
```
drewhk: You should give it the MaterializationPanic as cause, not the "cause"

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: WDYM?

drewhk:
WDYM fishy? Care to explain?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
viktorklang: Yeah, you discovered that too. :)

To my defense, it's not extremely intuitive :)

viktorklang:
@drewhk There is no, to my knowledge, documentation of the exception handling strategy for materialization so it is hard to "know" that things are handled at the correct place.

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
```diff
@@ -99,7 +100,7 @@ private[akka] case class ActorMaterializerImpl(
             assignPort(stage.outPort, processor)
             mat
 
-          case tls: TlsModule ⇒
+          case tls: TlsModule ⇒ // TODO isn't this a bit of a hack?
```
viktorklang: @drewhk It seems weird, to me, to special handle Tls in this way

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: what is not intuitive? I cannot really say anything to these comments, because I don't see what you have problems with. 

I try to explain what happens here though. If there is an exception during materialization (note, these are not "in-stream" exceptions, but exceptions when wiring up the graph), then we need to do our best to clean up. Unfortunately at this point we have a partially materialized graph, with potentially dangling Publishers and Subscribers, some of which might be 3rd party library entities actually.

This means that we just take all the Publishers/Subscriber that we know of, and try to shut them down, even if this means a Spec violation (there is basically nothing else we can do here, this is a best effort approach).

As for the stacks, these implement the scoped materialization, that is necessary to avoid copying graphs when you import them. These implement something similar to variable scopes in a language like Scala.

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
drewhk: There is no abstraction right now that allows to build n-to-n stages at the user level, unlike Stages for 1-to-1 or FlexiMerge/Route to 1-to-n or n-to-1 (and FlexiMerge/Route are limited). I need to actually implement the proper generalization of PushPullStage to arbitrary graph junctions. That is my next plan, since it was always postponed for various reasons.

viktorklang: Ok, in that case, I love TODOs/FIXMEs so that the casual reader thinks this is permanent :)

viktorklang:
Sigh, very flaky tests.

Now this from Maven: [maven-out] Tests in error: 
[maven-out]   groupedPartOfInfiniteStream(docs.stream.StreamTestKitDocTest): Futures timed out after [100 milliseconds]

Ignore?

drewhk:
No, please increase the timeouts in that test. anything sub-second will be flaky on Jenkins. Especially since they only add test overhead when they fail. Or open a ticket and I will search for "millis" and eliminate it.

viktorklang:
@drewhk Feedback addressed!

akka-stream/src/main/scala/akka/stream/impl/BlackholeSubscriber.scala
```diff
@@ -41,8 +44,7 @@ private[akka] class BlackholeSubscriber[T](highWatermark: Int, onComplete: Promi
   protected def requestMore(): Unit =
     if (requested < lowWatermark) {
       val amount = highWatermark - requested
-      subscription.get().request(amount)
       requested += amount
+      subscription.request(amount)
```
rkuhn: good catch!

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -390,6 +390,15 @@ class Source[+Out, +Mat](delegate: scaladsl.Source[Out, Mat]) extends Graph[Sour
     new Source(delegate.scan(zero)(f.apply))
 
   /**
+   * Similar to `scan` but only emits the current value once, when completing.
+   * Its current value which starts at `zero` and then
+   * applies the current and next value to the given function `f`,
+   * yielding the next current value.
```
rkuhn: Ah, it seems that we forgot to synchronize the doc comments between Flow and Source for Java: we should add the supervision and semantics documentation sections for all operations here as well.

rkuhn:
LGTM, great stuff, Viktor!

drewhk:
Please rebase, then I will merge this.

viktorklang:
@drewhk Rebased. Merge away!

akka-stream/src/main/scala/akka/stream/impl/BlackholeSubscriber.scala
viktorklang: Dey see me fishin', dey hatin'
:fishing_pole_and_fish:

drewhk:
Hooray!

Issue #15916, opened 2014-09-17
PersistentActor/ProcessorImpl should use HighSequenceNr to replay events efficiently


nefilim:
At the moment it looks like events are replayed from 0 (or snapshot.sequenceNr) to Long.maxValue and only after that is completed a ReadHighestSequenceNr is performed (in ProcessorImpl.onReplaySuccess). 

I propose that ReadHighestSequenceNr is done first and that value is used to:

1. replay events from 0 (or snapshot.sequenceNr) to HighSequenceNr 
2. OR do not attempt to replay events if the HighSequenceNr is 0, probably a new actor 

I believe without these changes it is not viable to use akka-persistence with DynamoDB and it's provisioned throughput model for a "large scale" deployment, especially when the system needs to support a very large number of persistent actors, see my comments & calculations with regard to akka-persistence(-dynamodb) here: https://github.com/sclasen/akka-persistence-dynamodb/issues/3

CptnKirk:
Has this been resolved in the 2.4 milestones?  I just spent a few days tracking down an issue where messages weren't replayed.  The root cause is the journal impl not properly working around this issue for sparse journals.

This feels like a somewhat major bug, and not a nice-to-have optimization.  What's the point of having 

```
def asyncReadHighestSequenceNr(processorId: String, fromSequenceNr: Long): Future[Long] 
```
as part of your journal interface if the framework isn't going to use the value to properly limit the

```
def asyncReplayMessages(processorId: String, fromSequenceNr: Long, toSequenceNr: Long, max: Long)(replayCallback: (PersistentRepr) => Unit): Future[Unit]
```
call?

Now that I know the problem I can fork the journal impl and fix it.  But still, seems like an easy to fix issue that shouldn't take a year to get into a production Akka release.  Sorry for the grump.  Been up for a while tracking this down.

PS:  There should probably be a *ReadLowestSequenceNr call as well.  Without a snapshot, the fromSeq should start at the last non-deleted sequence (unless overridden).  Not 0.  This is also causing havoc in my system.

Imagine a system without snapshots with 1 million records, each with 1 million mutations but coded such that the journal only keeps the last 5 mutations per processorId.  Properly bounding deleteMessages() and replayMessages() is hugely important.  And it's the framework that supplies these bounds.  It's up to the journal impl to faithfully process over them.  And in this case, the framework is being way more expensive than it has to be.  

Sure the journal impl could try and work around issues like these.  But it shouldn't have to, especially not for this long.

[...]

PR #17839, opened 2015-06-26
=per #15916 Read highestSeqNr first in replay

patriknw:
**Skip the first commit, it comes from #17838, which will be merged first**

If we are uncertain if this is 100% correct we could wait with this PR until after 2.4-M2
- we need to read the higestSeqNr anyway and it is better
  to do it first and limit the asyncReadHighestSequenceNr
  to that (instead of Long.MaxValue)
- return the highestSeqNr in the ReplayMessagesSuccess
- this also removes one become state in PersistentActor recovery
  logic

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/journal/AsyncRecovery.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/journal/AsyncRecovery.scala
@@ -24,6 +24,9 @@ trait AsyncRecovery {
    * as deleted. In this case a replayed message's `deleted` method must return
    * `true`.
    *
+   * The `toSequenceNr` is the lowest of what was returned by [[#asyncReadHighestSequenceNr]]
+   * and what the user specified as recovery [[akka.persistence.Recovery]] parameter.
+   *
    * @param persistenceId persistent actor id.
    * @param fromSequenceNr sequence number where replay should start (inclusive).
    * @param toSequenceNr sequence number where replay should end (inclusive).
@@ -38,7 +41,10 @@ trait AsyncRecovery {
 
   /**
    * Plugin API: asynchronously reads the highest stored sequence number for the
-   * given `persistenceId`.
+   * given `persistenceId`. The persistent actor will use the highest sequence
+   * number after recovery as the starting point when persisting new events.
+   * This sequence number is also used as `toSequenceNr` in subsequent call
+   * to [[#asyncReplayMessages]] unless the user has specified a lower `toSequenceNr`.
    *
    * @param persistenceId persistent actor id.
    * @param fromSequenceNr hint where to start searching for the highest sequence
--- a/akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
@@ -14,6 +14,7 @@ import scala.util.control.NonFatal
 import scala.util.Try
 import scala.util.Success
 import scala.util.Failure
+import akka.AkkaException
 
 /**
  * Abstract journal, optimized for asynchronous, non-blocking writes.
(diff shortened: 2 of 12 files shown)


patriknw:
Refs #15916

patriknw:
failure was unrelated (known issue)

patriknw:
rebased

akka-persistence/src/main/scala/akka/persistence/journal/AsyncRecovery.scala
```diff
@@ -38,7 +41,10 @@ trait AsyncRecovery {
 
   /**
    * Plugin API: asynchronously reads the highest stored sequence number for the
-   * given `persistenceId`.
+   * given `persistenceId`. The persistent actor will use the highest sequence
+   * number after recovery as the starting point when persisting new events.
+   * This sequence number is also used as `toSequenceNr` in subsequent call
+   * to [[#asyncReplayMessages]] unless the user has specified a lower `toSequenceNr`.
```
patriknw: @ktoso I added these docs

ktoso: very good :+1:

akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
```diff
@@ -14,6 +14,7 @@ import scala.util.control.NonFatal
 import scala.util.Try
 import scala.util.Success
 import scala.util.Failure
+import akka.AkkaException
```
ktoso: we have that? Oh!

akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
```diff
@@ -93,30 +94,29 @@ trait AsyncWriteJournal extends Actor with WriteJournalBase with AsyncRecovery {
...
-          adaptFromJournal(p).foreach { adaptedPersistentRepr ⇒
-            persistentActor.tell(ReplayedMessage(adaptedPersistentRepr), Actor.noSender)
-          }
-      } map {
-        case _ ⇒ ReplayMessagesSuccess
-      } recover {
+
+      asyncReadHighestSequenceNr(persistenceId, fromSequenceNr).flatMap { highSeqNr ⇒
+        val toSeqNr = math.min(toSequenceNr, highSeqNr)
+        if (highSeqNr == 0L || fromSequenceNr > toSeqNr)
```
ktoso: hm... so you think it's a success?
For me it sounded more like a journal in a weird state and would like some warning at least?

ktoso: Would a user randomly guess a number to start recovering from such that it would be wrong like (`>` triggers) this and still OK? My line of thinking was if we're recovery from a known sequenceId, and it's not there probably something fishy is going on.

patriknw: My conclusion was that we can't fail, at least some of our tests fail when I tried that.
Isn't `toSequenceNr` given by the user, e.g. 13, and fromSequenceNr comes from the snapshot, e.g. 17? That example should mean "no replay".

akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
patriknw: ah, residue, should be removed

akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
ktoso: I need to track where which Nr comes from, will come back with well-informed-opinion in a bit :)

ktoso: I went through the entire flow now. You're right that `fromSequenceNr` comes from the snapshot store. 

The case you mention will not happen however:
- recover(to = 13), results in:
  - load snapshot(to 13), so the snapshot store would not yield the 17 snapshot, it may yield a lower one or None
- a) we get `snapshotOffer = None` and `to = 13` => `snapshotOffer = None` is translated to a `from = 0`, start replay from 0 => OK
- - a) we get `snapshotOffer = Some(5)` and `to = 13` => `snapshotOffer` bumps the fromSequenceNr, `from = 5 + 1`, start replay from 6, since snapshot is applied => OK

So the case when the fromSequenceNr is lower than toSeqNr is when:
- user: recover to 18
- snapshotstore: snapshot offer available for 14
- recovery continues `fromSequenceNr = 15`
- journal: here a possibility for a mismatch, it could yield: `asyncReadHighestSequenceNr => 12`
  - why could this happen: because the journal is somehow out of sync with the snapshot store. For example someone retrored a old backup of the journal, but NOT of the snapshot store, so it has latter sequence ids
- weird state: if we say it's Future.successful then the actor is effectively in seqNr 15 state, from the snapshot offer...

This may actually be OK now that I have written it out, since the snapshot will be offered anyway always, so we'll get this state back again each time, and the sequenceNr does not overrwrite anything, there's a gap in the journal, but we don't hit it since we use the snapshot...

I think it would break if we would recover in such situation without using the snapshot. But perhaps that's expected.

ktoso: Ok, let's merge now and I'll explain the problem in another issue that we can revisit

Issue #17851, opened 2015-06-26
Mark Distributed Data as experimental


patriknw:
Somehow I forgot to mark it as experimental

PR #17852, opened 2015-06-26
=cdd #17851 Mark distributed data as experimental

patriknw:


akka-distributed-data/build.sbt
```diff
@@ -3,6 +3,8 @@ import com.typesafe.tools.mima.plugin.MimaKeys
 
 AkkaBuild.defaultSettings
 
+AkkaBuild.experimentalSettings
```
patriknw: is this correct? or does this overwrite the above defaultSettings?

ktoso:
Thanks Patrik!

akka-distributed-data/build.sbt
ktoso: Yes, it's correct - just adds some text to the description

patriknw:
Refs #17851

PR #17856, opened 2015-06-26
This is not the commit message you're looking for

viktorklang:
Various improvements, not for merging but for discussion!

### DIFF
--- a/akka-stream/src/main/boilerplate/akka/stream/FanInShape.scala.template
+++ b/akka-stream/src/main/boilerplate/akka/stream/FanInShape.scala.template
@@ -7,18 +7,24 @@ import scala.collection.immutable
 import scala.annotation.varargs
 
 object FanInShape {
-  sealed trait Init[+O]
-  case class Name(name: String) extends Init[Nothing]
-  case class Ports[O](out: Outlet[O], ins: immutable.Seq[Inlet[_]]) extends Init[O]
+  sealed trait Init[+O] {
+    def outlet: Outlet[O]
+    def inlets: immutable.Seq[Inlet[_]]
+    def name: String
+  }
+  final case class Name(override val name: String) extends Init[Nothing] {
+    override def outlet: Outlet[Nothing] = Outlet(s"$name.out")
+    override def inlets: immutable.Seq[Inlet[_]] = Nil
+  }
+  final case class Ports[O](override val outlet: Outlet[O], override val inlets: immutable.Seq[Inlet[_]]) extends Init[O] {
+    override def name: String = "FanIn"
+  }
 }
 
-abstract class FanInShape[O](init: FanInShape.Init[O]) extends Shape {
+abstract class FanInShape[O] private (_out: Outlet[O], _registered: Iterator[Inlet[_]], _name: String) extends Shape {
   import FanInShape._
 
-  final private[this] val (_out, _registered, _name) = init match {
-    case Name(name)   => (new Outlet[O](s"$name.out"), Nil.iterator, name)
-    case Ports(o, it) => (o, it.iterator, "FanIn")
-  }
+  def this(init: FanInShape.Init[O]) = this(init.outlet, init.inlets.iterator, init.name)
 
   final def out: Outlet[O] = _out
   final override def outlets: immutable.Seq[Outlet[_]] = _out :: Nil
@@ -26,14 +32,14 @@ abstract class FanInShape[O](init: FanInShape.Init[O]) extends Shape {
 
   private var _inlets: Vector[Inlet[_]] = Vector.empty
   protected def newInlet[T](name: String): Inlet[T] = {
-    val p = if (_registered.hasNext) _registered.next().asInstanceOf[Inlet[T]] else new Inlet[T](s"${_name}.$name")
+    val p = if (_registered.hasNext) _registered.next().asInstanceOf[Inlet[T]] else Inlet[T](s"${_name}.$name")
     _inlets :+= p
     p
   }
 
   protected def construct(init: Init[O]): FanInShape[O]
 
-  def deepCopy(): FanInShape[O] = construct(Ports[O](new Outlet(_out.toString), inlets.map(i => new Inlet(i.toString))))
+  def deepCopy(): FanInShape[O] = construct(Ports[O](_out.carbonCopy(), inlets.map(_.carbonCopy())))
   final def copyFromPorts(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]): FanInShape[O] = {
     require(outlets.size == 1, s"proposed outlets [${outlets.mkString(", ")}] do not fit FanInShape")
     require(inlets.size == _inlets.size, s"proposed inlets [${inlets.mkString(", ")}] do not fit FanInShape")
--- a/akka-stream/src/main/scala/akka/stream/Shape.scala
+++ b/akka-stream/src/main/scala/akka/stream/Shape.scala
@@ -26,13 +26,26 @@ sealed abstract class OutPort
  * is the InPort (which does not bear an element type because Modules only
  * express the internal structural hierarchy of stream topologies).
  */
-final class Inlet[-T](override val toString: String) extends InPort
+object Inlet {
+  def apply[T](toString: String): Inlet[T] = new Inlet[T](toString)
+}
+
+final class Inlet[-T] private (override val toString: String) extends InPort {
+  def carbonCopy(): Inlet[T] = Inlet(toString)
+}
+
 /**
  * An Outlet is a typed output to a Shape. Its partner in the Module view
  * is the OutPort (which does not bear an element type because Modules only
  * express the internal structural hierarchy of stream topologies).
  */
-final class Outlet[+T](override val toString: String) extends OutPort
+object Outlet {
+  def apply[T](toString: String): Outlet[T] = new Outlet[T](toString)
+}
+
+final class Outlet[+T] private (override val toString: String) extends OutPort {
+  def carbonCopy(): Outlet[T] = Outlet(toString)
+}
 
 /**
  * A Shape describes the inlets and outlets of a [[Graph]]. In keeping with the
@@ -148,9 +161,7 @@ object ClosedShape extends ClosedShape {
  * meaningful type of Shape when the building is finished.
  */
 case class AmorphousShape(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]) extends Shape {
-  override def deepCopy() = AmorphousShape(
-    inlets.map(i ⇒ new Inlet[Any](i.toString)),
-    outlets.map(o ⇒ new Outlet[Any](o.toString)))
+  override def deepCopy() = AmorphousShape(inlets.map(_.carbonCopy()), outlets.map(_.carbonCopy()))
   override def copyFromPorts(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]): Shape = AmorphousShape(inlets, outlets)
 }
 
@@ -162,7 +173,7 @@ final case class SourceShape[+T](outlet: Outlet[T]) extends Shape {
   override val inlets: immutable.Seq[Inlet[_]] = Nil
   override val outlets: immutable.Seq[Outlet[_]] = List(outlet)
 
-  override def deepCopy(): SourceShape[T] = SourceShape(new Outlet(outlet.toString))
+  override def deepCopy(): SourceShape[T] = SourceShape(outlet.carbonCopy())
   override def copyFromPorts(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]): Shape = {
     require(inlets.isEmpty, s"proposed inlets [${inlets.mkString(", ")}] do not fit SourceShape")
     require(outlets.size == 1, s"proposed outlets [${outlets.mkString(", ")}] do not fit SourceShape")
@@ -179,7 +190,7 @@ final case class FlowShape[-I, +O](inlet: Inlet[I], outlet: Outlet[O]) extends S
   override val inlets: immutable.Seq[Inlet[_]] = List(inlet)
   override val outlets: immutable.Seq[Outlet[_]] = List(outlet)
 
-  override def deepCopy(): FlowShape[I, O] = FlowShape(new Inlet(inlet.toString), new Outlet(outlet.toString))
+  override def deepCopy(): FlowShape[I, O] = FlowShape(inlet.carbonCopy(), outlet.carbonCopy())
   override def copyFromPorts(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]): Shape = {
     require(inlets.size == 1, s"proposed inlets [${inlets.mkString(", ")}] do not fit FlowShape")
     require(outlets.size == 1, s"proposed outlets [${outlets.mkString(", ")}] do not fit FlowShape")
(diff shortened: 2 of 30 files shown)


akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
```diff
@@ -46,22 +54,28 @@ private[akka] object FanIn {
       }
     }
 
-    private val marked = Array.ofDim[Boolean](inputCount)
+    private val states = Array.ofDim[State](inputCount)
```
viktorklang: This scheme removes 4 x Array[Boolean] (allocation and indirection)

akka-stream/src/main/scala/akka/stream/javadsl/FlattenStrategy.scala
```diff
@@ -17,11 +17,11 @@ object FlattenStrategy {
    * emitting its elements directly to the output until it completes and then taking the next stream. This has the
    * consequence that if one of the input stream is infinite, no other streams after that will be consumed from.
    */
-  def concat[T, U]: FlattenStrategy[Source[T, U], T] = Concat[T, U]()
+  def concat[T, U]: FlattenStrategy[Source[T, U], T] = Concat.asInstanceOf[FlattenStrategy[Source[T, U], T]]
```
viktorklang: no need for allocating new Concat instances

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -247,7 +247,8 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
   /** INTERNAL API */
   override private[stream] def andThen[U](op: StageModule): Repr[U, Mat] = {
     //No need to copy here, op is a fresh instance
-    if (this.isIdentity) new Flow(op).asInstanceOf[Repr[U, Mat]]
+    if (op.isInstanceOf[Stages.Identity]) this.asInstanceOf[Repr[U, Mat]]
```
viktorklang: No need to do anything if we are `andThen`-ing with Identity

akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
drewhk: Yes, Like!

akka-stream/src/main/boilerplate/akka/stream/FanInShape.scala.template
```diff
@@ -7,18 +7,24 @@ import scala.collection.immutable
 import scala.annotation.varargs
 
 object FanInShape {
-  sealed trait Init[+O]
-  case class Name(name: String) extends Init[Nothing]
-  case class Ports[O](out: Outlet[O], ins: immutable.Seq[Inlet[_]]) extends Init[O]
+  sealed trait Init[+O] {
```
viktorklang: This encoding avoids having to create and retain a Tuple3 instance

drewhk:
I skimmed it and they look great so far! (I will do a more detailed review later) Thanks!

In general, please split this into parts that:
 - change public API
 - don't change public API

The latter cannot go in until 1.1 unfortunately.

drewhk:
Ah, and don't forget Java APIs! You should run the akka-samples/akka-docs-java lambda project (it is a separate sbt project not included in the main build because of Java8 dep), to see if the Java doc examples are all compiling.

akka-stream/src/main/scala/akka/stream/Shape.scala
```diff
@@ -26,13 +26,26 @@ sealed abstract class OutPort
  * is the InPort (which does not bear an element type because Modules only
  * express the internal structural hierarchy of stream topologies).
  */
-final class Inlet[-T](override val toString: String) extends InPort
+object Inlet {
+  def apply[T](toString: String): Inlet[T] = new Inlet[T](toString)
+}
+
+final class Inlet[-T] private (override val toString: String) extends InPort {
+  def carbonCopy[R](): Inlet[R] = Inlet(toString)
```
viktorklang: I think this encoding (instead of manual copying using `new` and passing in the correct name) and it meshes very well with other classes' carbonCopy methods.

akka-stream/src/main/scala/akka/stream/javadsl/FlattenStrategy.scala
viktorklang: @drewhk I think this change might be the only public api breaking change in this PR, but it is trivially fixable by keeping a private Concat instance and using that for `def concat`. Wydt?

viktorklang:
TcpSpec spurious failure

drewhk:
Yeah, I think no public API were harmed, I was wondering about the changes in Init, but since all code compiles and the tests run, source compatibility was unharmed.

drewhk:
LGTM, really nice changes!

viktorklang:
Thanks @drewhk, feels good to contribute a bit back :)

akka-stream/src/main/scala/akka/stream/javadsl/FlattenStrategy.scala
```diff
@@ -17,11 +17,12 @@ object FlattenStrategy {
...
    */
-  def concat[T, U]: FlattenStrategy[Source[T, U], T] = Concat[T, U]()
+  def concat[T, U]: FlattenStrategy[Source[T, U], T] = _concat.asInstanceOf[FlattenStrategy[Source[T, U], T]]
 
+  private[this] final val _concat = Concat[Any]()
   /**
    * INTERNAL API
    */
-  private[akka] final case class Concat[T, U]() extends FlattenStrategy[Source[T, U], T]
+  private[akka] final case class Concat[T]() extends FlattenStrategy[Source[T, Any], T]
```
viktorklang: @drewhk WDYT about making this @deprecated and recommending "def concat" instead?

viktorklang: And saying that it will go away in 1.1?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -905,7 +907,7 @@ trait FlowOps[+Out, +Mat] {
    *
    */
   def flatten[U](strategy: FlattenStrategy[Out, U]): Repr[U, Mat] = strategy match {
-    case _: FlattenStrategy.Concat[Out] | _: javadsl.FlattenStrategy.Concat[Out, _] ⇒ andThen(ConcatAll())
+    case _: scaladsl.FlattenStrategy.Concat[_] | _: javadsl.FlattenStrategy.Concat[_] ⇒ andThen(ConcatAll())
```
viktorklang: Matching on `Out` here doesn't really make sense as it is erased.

viktorklang:
@drewhk Let me know if you need anything more from me here!

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -300,6 +300,7 @@ private[stream] object VirtualProcessor {
   case object Completed extends Termination
   case class Failed(ex: Throwable) extends Termination
 
+  // FIXME this is not a legal Subscriber in the RS sense as a Subscriber may never be attached to more than 1 Publisher
```
patriknw: is this a FIXME or "by design"?

akka-stream/src/main/scala/akka/stream/javadsl/FlattenStrategy.scala
patriknw: WDYM? this is internal api

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
```diff
@@ -341,8 +344,8 @@ abstract class StatefulStage[In, Out] extends PushPullStage[In, Out] {
    */
   abstract class State extends StageState[In, Out]
 
-  private var emitting = false
-  private var _current: StageState[In, Out] = _
+  private[this] var emitting = false
+  private[this] var _current: StageState[In, Out] = _
```
patriknw: I'm not sure about this. I know it's idomatic scala, but we have seen some surprises around it when it comes to binary compatibility evolution, and we "normally" don't use it anyway

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
viktorklang: Good question!

I think I added it there as a reminder that this needs to be discussed.
I think the solution is to drop it and use a CancellingSubscriber instead.

akka-stream/src/main/scala/akka/stream/javadsl/FlattenStrategy.scala
viktorklang: @patriknw I wasn't sure it wasn't used by something outside of this compilation unit, but that's unlikely, right?
If so then the proper fix is to make it a case object.

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
viktorklang: I think we'll see interesting things when we start to optimize.

akka-stream/src/main/scala/akka/stream/javadsl/FlattenStrategy.scala
patriknw: INTERNAL API means that we can change it

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
patriknw: ok, the extra private method can't matter for binary compatibility, it must have been another scenario I was thinking of.
Feel free to keep the `[this]`.
(I bet that hotspot will do it's job though)

patriknw:
LGTM

patriknw:
a proper commit message would be nice

viktorklang:
@patriknw For sure. Will add as I fix Concat.

akka-stream/src/main/boilerplate/akka/stream/FanInShape.scala.template
rkuhn: The number of objects created and retained is the same as before, but I don't mind either way.

viktorklang: Not the Tuple3?

-- 
Cheers,
√

akka-stream/src/main/scala/akka/stream/Shape.scala
rkuhn: Shouldn't this return an inlet of type T?

akka-stream/src/main/boilerplate/akka/stream/FanInShape.scala.template
rkuhn: Ah, right: that was created but not retained.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
viktorklang: Should I wait for #17831 to merge?
https://github.com/akka/akka/blob/wip-flow-fold-%E2%88%9A/akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala#L82

akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
```diff
@@ -183,13 +204,19 @@ private[akka] object FanIn {
     }
 
     def inputsAvailableFor(id: Int) = new TransferState {
-      override def isCompleted: Boolean = depleted(id) || cancelled(id)
-      override def isReady: Boolean = pending(id)
+      override def isCompleted: Boolean = {
+        val state = states(id)
+        has(state, Depleted) || has(state, Cancelled)
+      }
```
rkuhn: hmm, what about just keeping the old API such that this change goes away? (and most of the others as well)

viktorklang: I did that first, but keeping the old API means reading from `states` for every check. This change is an optimization (both in memory size and access)

akka-stream/src/main/scala/akka/stream/Shape.scala
viktorklang: True, the rationale was merely to be able to switch the type of an Inlet (given that they have no behavior).
I'll switch to remove the type parameter.

akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
rkuhn: Yes, but it also creates more opportunity for error, e.g. in `idToDequeue` above. What about adding `hasAnyOf` and `hasAllOf` methods?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: This is a private object precisely for the reason that it is correct only as used here, not in general. The VirtualProcessor uses this object just to atomically switch behavior in the right way, this InertSubscriber is never subscribed to any Publisher. Therefore this FIXME should be removed. If you want to use the CancellingSubscriber then please instantiate that here as a singleton in order to retain the same efficiency.

rkuhn:
LVGTM!

akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
viktorklang: Sadly that only covers a some of the use-cases. Things like "has(Marked) && !has(Pending)"?

rkuhn: Are you certain that you are not over-optimizing prematurely? I’d rather see JMH proof that this helps before making the code more brittle. The method I’m talking about would fetch a byte from an array and do a bit op on it, and it would be highly likely that the CPU would even elide the load for the second invocation, since it knows that the value does not change (no `@volatile` in sight). If this really is a hot path then the right things should happen.

viktorklang: I can definitely revert so that every flag check reads off the array, but I'm not sure how less brittle that is.

rkuhn: In the currently commented case it is rather obvious where `state` is coming from, but especially the while loop in `idToDequeue` would be trivial to break by not remembering that there is a `id <=> state` correspondence that needs to be maintained manually. In `dequeue` the distance between `id` and `state` and its usage is also quite big. In contrast `hasFlag(id, Depleted)` has all the pieces in one place. If it is measurably slower then there would be a reason for this reduced readability …

viktorklang: Alright, pushed the suggested fix.

akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
```diff
@@ -46,22 +53,39 @@ private[akka] object FanIn {
...
     private var markedPending = 0
-    private val depleted = Array.ofDim[Boolean](inputCount)
-    private val completed = Array.ofDim[Boolean](inputCount)
     private var markedDepleted = 0
-    private val cancelled = Array.ofDim[Boolean](inputCount)
+
+    private[this] final def hasState(index: Int, flag: Int): Boolean =
+      (states(index) & flag) != 0
+    private[this] final def setState(index: Int, flag: Int, on: Boolean): Unit =
+      states(index) = if (on) (states(index) | flag).toByte else (states(index) & ~flag).toByte
```
viktorklang: @rkuhn the downside of your proposal is that every write needs to be preceded by a read.

rkuhn: for the CPU that is anyway the case since we’re writing bytes here

viktorklang:
Rebased on release-2.3-dev

drewhk:
Thanks

viktorklang:
@drewhk let me know if you need any action from me here to merge.

drewhk:
You have the honor to merge :)

viktorklang:
Aye!

PR #17859, opened 2015-06-28
Removes the need for TerminationCallbacks to have a ReentrantGuard,

viktorklang:
replacing both the Lock and a CountDownLatch with an AtomicReference
and a Promise.

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/ActorSystem.scala
+++ b/akka-actor/src/main/scala/akka/actor/ActorSystem.scala
@@ -5,6 +5,7 @@
 package akka.actor
 
 import java.io.Closeable
+import java.util.concurrent.atomic.AtomicReference
 import java.util.concurrent.{ ConcurrentHashMap, ThreadFactory, CountDownLatch, TimeoutException, RejectedExecutionException }
 import java.util.concurrent.TimeUnit.MILLISECONDS
 import com.typesafe.config.{ Config, ConfigFactory }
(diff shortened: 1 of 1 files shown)


akka-actor/src/main/scala/akka/actor/ActorSystem.scala
```diff
@@ -783,44 +784,36 @@ private[akka] class ActorSystemImpl(val name: String, applicationConfig: Config,
...
-    final def add(callback: Runnable): Unit = {
-      latch.getCount match {
-        case 0 ⇒ throw new RejectedExecutionException("Must be called prior to system shutdown.")
-        case _ ⇒ lock withGuard {
-          if (latch.getCount == 0) throw new RejectedExecutionException("Must be called prior to system shutdown.")
-          else callbacks ::= callback
-        }
+    final def add(callback: Runnable): Unit =
+      get match {
+        case null ⇒ throw new RejectedExecutionException("Must be called prior to system shutdown.")
```
viktorklang: There is a slight difference in semantics for this exception, where it before threw the exception *after all* callbacks had executed, it now throws the exception if shutdown has started. This can easily be fixed but there does not seem to be a test for that semantics so it was uncertain if it was by design or not.

viktorklang: It is of course possible to change this line to:

~~~scala
case null ⇒ Await.ready(done.future, Duration.Inf); throw new RejectedExecutionException("Must be called prior to system shutdown.")
~~~

to get the same semantics as before, but it would be a sad day for sure.

viktorklang:
Lol, realized I had already fixed it for master a while back. Nothing to see here. :)

Issue #17796, opened 2015-06-21
Avoid duplicate change events for same data


patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node/1090/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/akka-nightly/2365/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3041/consoleFull

PR #17861, opened 2015-06-29
=cdd #17796 Avoid duplicate change events

patriknw:
The original test failure was in ReplicatedServiceRegistrySpec, and that should be fixed by this.


patriknw:
Refs #17796

PR #17863, opened 2015-06-29
fix Java8 DiningHakkers sample

rkuhn:


ktoso:
Silly mistake :-) LGTM!

Issue #17399, opened 2015-05-06
add boilerplate remover for fan-in junctions


rkuhn:
Currently users need to build a graph when the only thing they need is just one junction. We can fix this by adding something like
~~~scala
object Source {
  ...
  def combine[T, U](inputs: Source[T, Any]*)(strategy: Int => Graph[UniformFanInShape[T, U], Any]): Source[U, Unit]
~~~
that would allow elegant code like
~~~scala
Source.combine(s1, s2, s3)(Merge)
~~~
This could replace `Source.concat` since it subsumes it.

rkuhn:
Should also provide the mirror operation of combining Sinks.

agolubev:
@drewhk, I believe this ticket should be closed?

rkuhn:
This was done in #17865.

PR #17865, opened 2015-06-30
+str #17399 add boilerplate remover for fan-in junctions

agolubev:
Please verify if implementation approach is good enough. I'm still not sure about return type for sink mirrored version.
If it's Ok I'll need to add following: 
- docs
- java API
- sink tests
- more source tests
- check formatting for boilerplate code

### DIFF
--- a/akka-docs-dev/rst/scala/stream-graphs.rst
+++ b/akka-docs-dev/rst/scala/stream-graphs.rst
@@ -152,6 +152,19 @@ For defining a ``Flow[T]`` we need to expose both an inlet and an outlet:
 
 .. includecode:: code/docs/stream/StreamPartialFlowGraphDocSpec.scala#flow-from-partial-flow-graph
 
+Combining Sources and Sinks with simplified API
+-----------------------------------------------
+
+There is simplified API you can use to combine sources and sinks with junctions like: ``Broadcast[T]``, ``Balance[T]``,
+``Merge[In]`` and ``Concat[A]`` without the need for using the Graph DSL. The combine method takes care of constructing
+the necessary graph underneath. In following example we combine two sources into one (fan-in):
+
+.. includecode:: code/docs/stream/StreamPartialFlowGraphDocSpec.scala#source-combine
+
+The same can be done for a ``Sink[T]`` but in this case it will be fan-out:
+
+.. includecode:: code/docs/stream/StreamPartialFlowGraphDocSpec.scala#sink-combine
+
 Building reusable Graph components
 ----------------------------------
 
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/SinkSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/SinkSpec.scala
@@ -3,10 +3,12 @@
  */
 package akka.stream.scaladsl
 
-import akka.stream.testkit._
 import akka.stream.ActorMaterializer
+import akka.stream.testkit.TestPublisher.ManualProbe
+import akka.stream.testkit._
 
 class SinkSpec extends AkkaSpec {
+
   import FlowGraph.Implicits._
 
   implicit val mat = ActorMaterializer()
@@ -86,6 +88,42 @@ class SinkSpec extends AkkaSpec {
         s.request(3)
         p.expectNext(i)
         p.expectComplete()
+      }
+    }
+
+    "combine to many outputs with simplified API" in {
+      val probes = Seq.fill(3)(TestSubscriber.manualProbe[Int]())
+      val sink = Sink.combine(Sink(probes(0)), Sink(probes(1)), Sink(probes(2)))(Broadcast(_))
+
+      Source(List(0, 1, 2)).runWith(sink)
+
+      val subscriptions = probes.map(_.expectSubscription())
+
+      subscriptions.foreach { s ⇒ s.request(1) }
+      probes.foreach { p ⇒ p.expectNext(0) }
+
+      subscriptions.foreach { s ⇒ s.request(2) }
+      probes.foreach { p ⇒
+        p.expectNextN(List(1, 2))
+        p.expectComplete
+      }
+    }
+
+    "combine to two sinks with simplified API" in {
+      val probes = Seq.fill(2)(TestSubscriber.manualProbe[Int]())
+      val sink = Sink.combine(Sink(probes(0)), Sink(probes(1)))(Broadcast(_))
+
+      Source(List(0, 1, 2)).runWith(sink)
+
+      val subscriptions = probes.map(_.expectSubscription())
+
+      subscriptions.foreach { s ⇒ s.request(1) }
+      probes.foreach { p ⇒ p.expectNext(0) }
+
+      subscriptions.foreach { s ⇒ s.request(2) }
+      probes.foreach { p ⇒
+        p.expectNextN(List(1, 2))
+        p.expectComplete
       }
     }
 
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
@@ -153,6 +153,51 @@ class SourceSpec extends AkkaSpec {
       gotten.toSet should ===(Set(0, 1, 2, 3, 4))
       out.expectComplete()
     }
+
+    "combine from many inputs with simplified API" in {
+      val probes = Seq.fill(3)(TestPublisher.manualProbe[Int]())
+      val source = for (i ← 0 to 2) yield Source(probes(i))
+      val out = TestSubscriber.manualProbe[Int]
+
+      Source.combine(source(0), source(1), source(2))(Merge(_)).to(Sink(out)).run()
+
+      val sub = out.expectSubscription()
+      sub.request(3)
+
+      for (i ← 0 to 2) {
+        val s = probes(i).expectSubscription()
+        s.expectRequest()
+        s.sendNext(i)
+        s.sendComplete()
+      }
+
+      val gotten = for (_ ← 0 to 2) yield out.expectNext()
+      gotten.toSet should ===(Set(0, 1, 2))
+      out.expectComplete()
+    }
+
+    "combine from two inputs with simplified API" in {
+      val probes = Seq.fill(2)(TestPublisher.manualProbe[Int]())
+      val source = Source(probes(0)) :: Source(probes(1)) :: Nil
+      val out = TestSubscriber.manualProbe[Int]
+
+      Source.combine(source(0), source(1))(Merge(_)).to(Sink(out)).run()
+
+      val sub = out.expectSubscription()
+      sub.request(3)
+
+      for (i ← 0 to 1) {
+        val s = probes(i).expectSubscription()
+        s.expectRequest()
+        s.sendNext(i)
+        s.sendComplete()
+      }
+
+      val gotten = for (_ ← 0 to 1) yield out.expectNext()
+      gotten.toSet should ===(Set(0, 1))
+      out.expectComplete()
+    }
+
   }
 
   "Repeat Source" must {
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
@@ -226,6 +226,15 @@ object Source {
       case s: Source[T, M] ⇒ s
       case other           ⇒ new Source(scaladsl.Source.wrap(other))
     }
+
+  /**
+   * Combines several sources with fan-in strategy like `Merge` or `Concat` and returns `Source`.
+   */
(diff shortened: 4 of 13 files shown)


ktoso:
OK TO TEST

ktoso:
Refs #17399

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
```diff
@@ -153,6 +153,29 @@ class SourceSpec extends AkkaSpec {
       gotten.toSet should ===(Set(0, 1, 2, 3, 4))
       out.expectComplete()
     }
+
```
viktorklang: No tests for divide?

agolubev: Had intention to add them (as well as docs and java api) after positive feedback on implementation approach.
Still if you need this to see API usage I add them right away.

viktorklang: I completely understand that rationale, but if you look at it from the readers' point of view, there's really nothing that verifies that it does work and if so what it does.

drewhk:
I think this approach looks good. You can also change now the implementation of Source.concat to use this feature instead of creating the graph manually. So please proceed with tests :)

agolubev:
Yeah, thanks, will proceed.

akka-stream/src/main/boilerplate/akka/stream/scaladsl/SinkApply.scala.template
```diff
@@ -47,4 +47,21 @@ trait SinkApply {
...
+   * Combine several sinks with fun-out strategy like `Broadcast` or `Balance` and returns `Sink`.
+   */
+  def combine[T, U]([#output1: Sink[U, Any]#])(strategy: Int => Graph[UniformFanOutShape[T, U], Unit]): Sink[T, Unit] =
+    Sink([#output1#])(([#M1#]) => ()) { implicit b ⇒
+      ([#s1#]) ⇒
+        import FlowGraph.Implicits._
+        val d = b.add(strategy(1))
+
+        [#d.out(0) ~> s1.inlet#
+        ]
```
viktorklang: why newline here?

akka-stream/src/main/boilerplate/akka/stream/scaladsl/SourceApply.scala.template
```diff
@@ -49,4 +49,22 @@ trait SourceApply {
   
   ]
 
+  [2..#/**
+   * Combines several sources with fun-in strategy like `Merge` or `Concat` and returns`Source`.
```
viktorklang: missing space in "returns`Source`"

akka-stream/src/main/boilerplate/akka/stream/scaladsl/SinkApply.scala.template
agolubev: Idea was to visually divide boilerplate section with out ~> in.

akka-stream/src/main/boilerplate/akka/stream/scaladsl/SourceApply.scala.template
```diff
@@ -49,4 +49,22 @@ trait SourceApply {
...
+      ([#s1#]) ⇒
+        import FlowGraph.Implicits._
+        val c = b.add(strategy(1))
+
+        [#s1 ~> c.in(0)#
+        ]
+
+        SourceShape(c.out)
+    })
+  #
```
viktorklang: I wonder if using templating for this is the optimal solution (given the code duplication and scaling),
wdyt about something like the following?

~~~scala

 def combine[T, U](first: Source[T, _], second: Source[T, _], rest: Source[T, _]*)(strategy: Int ⇒ Graph[UniformFanInShape[T, U], Unit]): Source[U, Unit] =
    Source.wrap(FlowGraph.partial() { implicit b ⇒
      import FlowGraph.Implicits._

      val c = b.add(strategy(rest.size + 2))

      @tailrec def addRest(idx: Int, i: Iterator[Source[T, _]]): SourceShape[U] =
        if (i.hasNext) {
          i.next() ~> c.in(idx)
          addRest(idx + 1, i)
        } else SourceShape(c.out)

      first ~> c.in(0)
      second ~> c.in(1)
      addRest(2, rest.iterator)
    })

~~~

agolubev: Make sense. Will connect all in-outs with help of internal function

agolubev:
Ok, I moved Scala API to be non boilerplate as was suggested.

For Java API it would be something like this:
```scala
@varargs def combine[T, U](strategy: function.Function[java.lang.Integer, Graph[UniformFanOutShape[T, U], Unit]],
                           outputs: Sink[U, _]*): Sink[T, Unit] =
  new Sink[T, Unit](scaladsl.Sink.combine(outputs.map(_.asScala): _*)(num ⇒ strategy.apply(num)))
```
However we have two obstacles here:
* varargs must be as the last argument. It makes us rearrange function attributes and create difference between Java and Scala API (still it's not critical).
* also there is an issue in Scala compiler https://issues.scala-lang.org/browse/SI-8743 which is fixed only in 2.11.3 - it hardly can be workarounded with the given API.

Consequently I left Java API in boilerplate as it was.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -252,4 +253,22 @@ object Sink extends SinkApply {
   def actorSubscriber[T](props: Props): Sink[T, ActorRef] =
     new Sink(new ActorSubscriberSink(props, DefaultAttributes.actorSubscriberSink, shape("ActorSubscriberSink")))
 
+  /**
+   * Combine several sinks with fun-out strategy like `Broadcast` or `Balance` and returns `Sink`.
+   */
+  def combine[T, U](sinks: Sink[U, _]*)(strategy: Int ⇒ Graph[UniformFanOutShape[T, U], Unit]): Sink[T, Unit] =
```
viktorklang: To enforce that it is not used with less than 2, this should (IMO) be:

~~~scala
def combine[T, U](sink1: Sink[U, _], sink2:Sink[U, _], sinkN: Sink[U, _]*)(strategy: Int ⇒ Graph[UniformFanOutShape[T, U], Unit]): Sink[T, Unit] = …
~~~

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SinkSpec.scala
```diff
@@ -89,6 +91,25 @@ class SinkSpec extends AkkaSpec {
       }
     }
 
+    "combine from many inputs with simplified API" in {
+      val probes = Seq.fill(3)(TestSubscriber.manualProbe[Int]())
+      val sink = Sink.combine(Sink(probes(0)), Sink(probes(1)), Sink(probes(2)))(Broadcast(_))
```
viktorklang: There's no test for when an empty varargs is used. or a single Sink.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
agolubev: Yeah, make sense.

agolubev:
Ok, it seems like build verifies that Java DSL is equal to Scala DSL. Should I align them or add exception to test?

viktorklang:
Aligning the Java API sounds like the best option IMO.

drewhk:
Due to a Scala issue, we couldn't make the varargs version proper for Java, and therefore we removed it (you can still use a Collection).

drewhk:
But Java API equivalents for combine is necessary, but you can add an exception on the varargs version, creating a Collection based alternative in Java.

agolubev:
Yeah, just committed with List in Java API, as it seems it's the closest to scala Seq. (http://www.scala-lang.org/docu/files/collections-api/collections_46.html)

ktoso:
It's moving in the right direction, thanks @agolubev! Though we won't include it in the 1.0 today, we'll get back to it and get it into 1.1

drewhk:
Hi @agolubev can you rebase this PR? i would like to merge it as it seems ready to me.

agolubev:
Ok, will do

agolubev:
Seems like HttpEntitySpec failed. 
Let me know if something relates to my commit

agolubev:
@drewhk it's 5 days passed - perhaps it will require rebase once again.

drewhk:
PLS BUILD

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SinkSpec.scala
drewhk: +1, that test case is important

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -107,6 +108,26 @@ object Sink extends SinkApply {
     Flow[T].map(f).toMat(Sink.ignore)(Keep.right).named("foreachSink")
 
   /**
+   * Combine several sinks with fun-out strategy like `Broadcast` or `Balance` and returns `Sink`.
```
drewhk: "fan-out"

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -226,6 +226,15 @@ object Source {
       case s: Source[T, M] ⇒ s
       case other           ⇒ new Source(scaladsl.Source.wrap(other))
     }
+
+  /**
+   * Combines several sources with fun-in strategy like `Merge` or `Concat` and returns `Source`.
```
drewhk: "fan-in"

drewhk:
Looking good, please add a small section to the doc pages, then I merge this.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SinkSpec.scala
agolubev: now combine definition is like
```def combine[T, U](first: Sink[U, _], second: Sink[U, _], rest: Sink[U, _]*)``` 
so I'll add test for two sinks

**Comment by agolubev** [comment]
Ok, added small doc and rebased

**Comment by agolubev** [comment]
TcpSpec "shutdown if Flows have not been subscribed" case failed

**Comment by agolubev** [comment]
@drewhk, I finished with this

File: akka-docs-dev/rst/scala/stream-graphs.rst [hunk 9]
```diff
@@ -152,6 +152,20 @@ For defining a ``Flow[T]`` we need to expose both an inlet and an outlet:
 
 .. includecode:: code/docs/stream/StreamPartialFlowGraphDocSpec.scala#flow-from-partial-flow-graph
 
+Combining Sources and Sinks with simplified API
+-----------------------------------------------
+
+There is simplified API you can use to combine sources and sinks with junctions like: ``Broadcast[T]``, ``Balance[T]``,
+``Merge[In]`` and ``Concat[A]``. With method ``combine`` you don't need to connect junction
```
drewhk: ".. without the need for using the Graph DSL. The combine method takes care of constructing the necessary graph underneath."

akka-docs-dev/rst/scala/stream-graphs.rst
```diff
@@ -152,6 +152,20 @@ For defining a ``Flow[T]`` we need to expose both an inlet and an outlet:
...
+with sources or sinks. It will construct simple graph for you. In following example we combine two sources into
+one (fun-in):
```
drewhk: fan-in

drewhk:
Java documentation section is missing yet.

agolubev:
build failed with following:
```
maven-out] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:testCompile (default-testCompile) on project akka-docs-java-lambda: Compilation failure
[maven-out] [ERROR] required: Graph<UniformFanInShape<Integer,Integer>,BoxedUnit>
[maven-out] [ERROR] found:    Graph
[maven-out] [ERROR] /mnt/akka/jenkinsakka/localhome-a0/workspace/pr-validator-per-commit-jenkins/akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/StreamPartialFlowGraphDocTest.java:[168,25] error: cannot find symbol
```
this is strange because when I put problematic test case into SinkTest and set target java to 1.8 it works ok.

agolubev:
@drewhk any ideas?

agolubev:
Sorry for so many build failures. It took me sometime to figure out how to work properly with lambda samples and docs for java

agolubev:
@drewhk, it's ready

drewhk:
Almost ;) Please squash the two remaining commits, and then I will merge after validation. Awesome work @agolubev !

agolubev:
build failed with TcpSpec issue. Fix is in PR #18369

drewhk:
PLS BUILD

Issue #17846, opened 2015-06-26
Use CRDTs instead of PersistentActor to remember the state of the ShardCoordinator


smlin:
Currently the cluster-sharding module requires an akka-persistence journal for it to work (and database of course). With CRDTs it would be possible to remove this requirement.

patriknw:
That is a very interesting thought, that I also have had.

The difficult part is if we can get enough consistency.
Let me illustrate with an example.

* 7 nodes: n1, n2, n3, n4, n5, n6, n7
* coordinator lives on n1
* shard s7 for entity e700 is allocated to n6 and the coordinator successfully writes this to a majority of nodes, which happens to be n1, n2, n3, n4
* e700 is started on n6
* network partition, (n1, n2, n3, n4) / (n5, n6, n7)
* downing decision to keep (n5, n6, n7), and shutdown (n1, n2, n3, n4)
* n5 becomes new coordinator in the cluster that now consists of (n5, n6, n7), it reads from majority, e.g. n5 and n6, which don't contain any information about s7 (because it was not replicated there before the split)
* s7 for entity e700 is allocated again, now to n7, and now e700 is running on both n6 and n7

This could also happen if (n1, n2, n3, n4) crashed, instead of network partition.

smlin:
Such "unhappy" partitioning is possible only when allocation of s7 to n6 happens before the data about this event was replicated.
I also thought about inconsistency and came to a conclusion that it can be solved by instantiating a coordinator on each node, from which only one will be allowed to write to CRDT collection (change coordinator state). I think it should be the one on the oldest node. So e700 may be started on n6 only after data about this had been received by coordinator6 from coordinator1.

Earlier I was planning to make a small PR to demonstrate this concept, but I have time only after Monday.

patriknw:
it is indeed interesting, and we should try to find a solution, even if it means a "relaxed" mode, which perhaps can't be used with persistent entities

smlin:
I make PR to demonstrate this concept. https://github.com/akka/akka/pull/17871
1) After some thoughts I was decided to save the Coordinator as singleton, but to allow the changing of the distributed data only from a region inside which shard would be allocated. It is allowing do not worry about inconsistency between the distributed data, Coordinator and ShardRegions.
2) Currently I do not know how make the rebalancing and proxy. But I think it is solvable problem.

PR #17871, opened 2015-06-30
=cls #17846 Use CRDTs instead of PersistentActor to remember the state of the ShardCoordinator

oseval:
Currently this is not real PR. Only to show my attempt to change the claster-sharding dependency from the akka-persistence to the akka-ddata. I would incrementally update it when found time for this .

Idea:
1) Distributed state of the ShardRegions are all we need in the cluster-sharding.
2) Distributed state can be handled by akka-ddata module.
3) Inconsistency is possible if only we allowing two ShardRegions allocate equals Shards simultaneously.
4) It is possible avoid inconsistency if we have singleton Coordinator who can allow to the ShardRegion allocate the Shard. R1 ask C to allow him allocate the S1. C1 allow. R1 allocate S1 and then change DData + inner state. The Coordinator will be receiving these changes and then update inner state.
5) If the Coordinator fails, then it can be recovered from DData which be actual for all nodes inside a cluster. 

Welcome for comments!

It is very possible if I missing something in this concept. Correct me please.

And I am sorry for my English.

### DIFF
--- a/akka-cluster-sharding/src/main/resources/reference.conf
+++ b/akka-cluster-sharding/src/main/resources/reference.conf
@@ -13,77 +13,89 @@ akka.cluster.sharding {
   # The extension creates a top level actor with this name in top level system scope,
   # e.g. '/system/sharding'
   guardian-name = sharding
-  
+
   # Specifies that entities runs on cluster nodes with a specific role.
   # If the role is not specified (or empty) all nodes in the cluster are used.
   role = ""
-  
+
   # When this is set to 'on' the active entity actors will automatically be restarted
   # upon Shard restart. i.e. if the Shard is started on a different ShardRegion
   # due to rebalance or crash.
   remember-entities = off
-  
+
   # If the coordinator can't store state changes it will be stopped
   # and started again after this duration, with an exponential back-off
   # of up to 5 times this duration.
   coordinator-failure-backoff = 5 s
-  
+
   # The ShardRegion retries registration and shard location requests to the
   # ShardCoordinator with this interval if it does not reply.
   retry-interval = 2 s
-  
+
   # Maximum number of messages that are buffered by a ShardRegion actor.
   buffer-size = 100000
-  
+
   # Timeout of the shard rebalancing process.
   handoff-timeout = 60 s
-  
+
   # Time given to a region to acknowledge it's hosting a shard.
   shard-start-timeout = 10 s
-  
+
   # If the shard is remembering entities and can't store state changes
   # will be stopped and then started again after this duration. Any messages
   # sent to an affected entity may be lost in this process.
   shard-failure-backoff = 10 s
-  
+
   # If the shard is remembering entities and an entity stops itself without
   # using passivate. The entity will be restarted after this duration or when
   # the next message for it is received, which ever occurs first.
   entity-restart-backoff = 10 s
-  
+
   # Rebalance check is performed periodically with this interval.
   rebalance-interval = 10 s
-  
+
   # Absolute path to the journal plugin configuration entity that is to be
   # used for the internal persistence of ClusterSharding. If not defined
-  # the default journal plugin is used. Note that this is not related to 
+  # the default journal plugin is used. Note that this is not related to
   # persistence used by the entity actors.
   journal-plugin-id = ""
-  
+
   # Absolute path to the snapshot plugin configuration entity that is to be
   # used for the internal persistence of ClusterSharding. If not defined
-  # the default snapshot plugin is used. Note that this is not related to 
+  # the default snapshot plugin is used. Note that this is not related to
   # persistence used by the entity actors.
   snapshot-plugin-id = ""
-  
-  # The coordinator saves persistent snapshots after this number of persistent
-  # events. Snapshots are used to reduce recovery times. 
+
+  # Parameter which determines how the coordinator will be store a state
+  # valid values either "persistence" or "ddata"
+  state-store-mode = "persistence"
+
+  # The shard saves persistent snapshots after this number of persistent
+  # events. Snapshots are used to reduce recovery times.
   snapshot-after = 1000
-  
+
   # Setting for the default shard allocation strategy
   least-shard-allocation-strategy {
     # Threshold of how large the difference between most and least number of
     # allocated shards must be to begin the rebalancing.
     rebalance-threshold = 10
-    
+
     # The number of ongoing rebalancing processes is limited to this number.
     max-simultaneous-rebalance = 3
   }
-  
-  # Settings for the coordinator singleton. Same layout as akka.cluster.singleton.  
+
+  # Timeout of waiting the initial distributed state (an initial state will be queried again if the timeout happened)
+  # works only for state-store-mode = "ddata"
+  waiting-for-state-timeout = 5 s
+
+  # Timeout of waiting for update the distributed state (update will be retried if the timeout happened)
+  # works only for state-store-mode = "ddata"
+  updating-state-timeout = 5 s
+
+  # Settings for the coordinator singleton. Same layout as akka.cluster.singleton.
   coordinator-singleton = ${akka.cluster.singleton}
-  
-  # The id of the dispatcher to use for ClusterSharding actors. 
+
+  # The id of the dispatcher to use for ClusterSharding actors.
   # If not specified default dispatcher is used.
   # If specified you need to define the settings of the actual dispatcher.
   # This dispatcher for the entity actors is defined by the user provided
--- a/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
+++ b/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
@@ -98,7 +98,7 @@ private[akka] class Shard(
   extractShardId: ShardRegion.ExtractShardId,
   handOffStopMessage: Any) extends Actor with ActorLogging {
 
-  import ShardRegion.{ handOffStopperProps, EntityId, Msg, Passivate }
+  import ShardRegion.{ handOffStopperProps, EntityId, Msg, Passivate, ShardInitialized }
   import ShardCoordinator.Internal.{ HandOff, ShardStopped }
   import Shard.{ State, RestartEntity, EntityStopped, EntityStarted }
   import akka.cluster.sharding.ShardCoordinator.Internal.CoordinatorMessage
@@ -112,6 +112,10 @@ private[akka] class Shard(
   var messageBuffers = Map.empty[EntityId, Vector[(Msg, ActorRef)]]
 
   var handOffStopper: Option[ActorRef] = None
+
+  initialized()
+
+  def initialized(): Unit = context.parent ! ShardInitialized(shardId)
 
   def totalBufferSize = messageBuffers.foldLeft(0) { (sum, entity) ⇒ sum + entity._2.size }
 
@@ -297,6 +301,9 @@ private[akka] class PersistentShard(
 
   var persistCount = 0
 
+  // would be initialized after recovery completed
+  override def initialized(): Unit = {}
+
   override def receive = receiveCommand
 
   override def processChange[A](event: A)(handler: A ⇒ Unit): Unit = {
@@ -316,7 +323,10 @@ private[akka] class PersistentShard(
     case EntityStarted(id)                 ⇒ state = state.copy(state.entities + id)
     case EntityStopped(id)                 ⇒ state = state.copy(state.entities - id)
     case SnapshotOffer(_, snapshot: State) ⇒ state = snapshot
-    case RecoveryCompleted                 ⇒ state.entities foreach getEntity
+    case RecoveryCompleted ⇒
+      state.entities foreach getEntity
+      super.initialized()
+      log.debug("Shard recovery completed {}", shardId)
   }
 
   override def entityTerminated(ref: ActorRef): Unit = {
--- a/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
+++ b/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
@@ -3,6 +3,8 @@
  */
 package akka.cluster.sharding
 
+import akka.persistence._
+
 import scala.collection.immutable
 import scala.concurrent.Future
 import scala.concurrent.duration._
@@ -15,16 +17,15 @@ import akka.actor.Deploy
 import akka.actor.NoSerializationVerificationNeeded
 import akka.actor.Props
 import akka.actor.ReceiveTimeout
+import akka.actor.Stash
 import akka.actor.Terminated
 import akka.cluster.Cluster
 import akka.cluster.ClusterEvent.ClusterShuttingDown
 import akka.cluster.ClusterEvent.CurrentClusterState
+import akka.cluster.ddata.LWWRegisterKey
+import akka.cluster.ddata.LWWRegister
+import akka.cluster.ddata.Replicator._
 import akka.dispatch.ExecutionContexts
-import akka.persistence.PersistentActor
-import akka.persistence.RecoveryCompleted
-import akka.persistence.SaveSnapshotFailure
-import akka.persistence.SaveSnapshotSuccess
-import akka.persistence.SnapshotOffer
 import akka.pattern.pipe
 
 /**
@@ -40,7 +41,16 @@ object ShardCoordinator {
    */
   private[akka] def props(typeName: String, settings: ClusterShardingSettings,
                           allocationStrategy: ShardAllocationStrategy): Props =
-    Props(new ShardCoordinator(typeName: String, settings, allocationStrategy)).withDeploy(Deploy.local)
+    Props(new PersistentShardCoordinator(typeName: String, settings, allocationStrategy)).withDeploy(Deploy.local)
+
+  /**
+   * INTERNAL API
+   * Factory method for the [[akka.actor.Props]] of the [[ShardCoordinator]] actor with state based on ddata.
+   */
+  private[akka] def props(typeName: String, settings: ClusterShardingSettings,
+                          allocationStrategy: ShardAllocationStrategy,
+                          replicator: ActorRef): Props =
+    Props(new DDataShardCoordinator(typeName: String, settings, allocationStrategy, replicator)).withDeploy(Deploy.local)
 
   /**
    * Interface of the pluggable shard allocation and rebalancing logic used by the [[ShardCoordinator]].
@@ -51,12 +61,12 @@ object ShardCoordinator {
     /**
      * Invoked when the location of a new shard is to be decided.
      * @param requester actor reference to the [[ShardRegion]] that requested the location of the
-     *   shard, can be returned if preference should be given to the node where the shard was first accessed
+     *                  shard, can be returned if preference should be given to the node where the shard was first accessed
      * @param shardId the id of the shard to allocate
      * @param currentShardAllocations all actor refs to `ShardRegion` and their current allocated shards,
-     *   in the order they were allocated
+     *                                in the order they were allocated
      * @return a `Future` of the actor ref of the [[ShardRegion]] that is to be responsible for the shard, must be one of
-     *   the references included in the `currentShardAllocations` parameter
+     *         the references included in the `currentShardAllocations` parameter
      */
     def allocateShard(requester: ActorRef, shardId: ShardId,
                       currentShardAllocations: Map[ActorRef, immutable.IndexedSeq[ShardId]]): Future[ActorRef]
@@ -64,9 +74,9 @@ object ShardCoordinator {
     /**
      * Invoked periodically to decide which shards to rebalance to another location.
      * @param currentShardAllocations all actor refs to `ShardRegion` and their current allocated shards,
-     *   in the order they were allocated
+     *                                in the order they were allocated
      * @param rebalanceInProgress set of shards that are currently being rebalanced, i.e.
-     *   you should not include these in the returned set
+     *                            you should not include these in the returned set
      * @return a `Future` of the shards to be migrated, may be empty to skip rebalance in this round
      */
     def rebalance(currentShardAllocations: Map[ActorRef, immutable.IndexedSeq[ShardId]],
@@ -95,12 +105,12 @@ object ShardCoordinator {
     /**
      * Invoked when the location of a new shard is to be decided.
      * @param requester actor reference to the [[ShardRegion]] that requested the location of the
-     *   shard, can be returned if preference should be given to the node where the shard was first accessed
+     *                  shard, can be returned if preference should be given to the node where the shard was first accessed
      * @param shardId the id of the shard to allocate
      * @param currentShardAllocations all actor refs to `ShardRegion` and their current allocated shards,
-     *   in the order they were allocated
+     *                                in the order they were allocated
      * @return a `Future` of the actor ref of the [[ShardRegion]] that is to be responsible for the shard, must be one of
-     *   the references included in the `currentShardAllocations` parameter
+     *         the references included in the `currentShardAllocations` parameter
      */
     def allocateShard(requester: ActorRef, shardId: String,
                       currentShardAllocations: java.util.Map[ActorRef, immutable.IndexedSeq[String]]): Future[ActorRef]
@@ -108,9 +118,9 @@ object ShardCoordinator {
     /**
      * Invoked periodically to decide which shards to rebalance to another location.
      * @param currentShardAllocations all actor refs to `ShardRegion` and their current allocated shards,
-     *   in the order they were allocated
+     *                                in the order they were allocated
      * @param rebalanceInProgress set of shards that are currently being rebalanced, i.e.
-     *   you should not include these in the returned set
+     *                            you should not include these in the returned set
      * @return a `Future` of the shards to be migrated, may be empty to skip rebalance in this round
      */
     def rebalance(currentShardAllocations: java.util.Map[ActorRef, immutable.IndexedSeq[String]],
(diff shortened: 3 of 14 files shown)


smlin:
@patriknw I have some questions about cluster sharding. I need to know the "underlying idea". Currently only two questions.
1) Is it critical that the Coordinator reallocates the shards of the terminated regions immediately? Is it possible if the shard from a terminated region will be allocated only after sending a message to this shard?
2) Is it applicable if the terminated region would be remove itself from the distributed state by sending message to the replicator inside the `postStop`?

patriknw:
Hi @smlin
I have not had time to look at your code yet. Will try to get here tomorrow.

There are two things that are persisted.
1. Coordinator state, the allocations of all shards
2. The started entities within each shard, see the Remember Entities section in the documentation.

Lets try to figure out the coordinator state first.
For that the allocation is always done for the first message to a shard, also for rebalance and crash.

It is possible to send message to replicator in postStop, but crash must also be handled and if the node is removed from the cluster the replicator might not be alive any more.

smlin:
@patriknw PR is not ready for review yet. So you could do not waste your time. I will notify to you when it would be ready.

patriknw:
@smlin My thinking around the design.

I think the coordinator should remain a singleton, because it is important with a central decision maker of where the shards are to be allocated. We could start with just saving the coordinator state in a LWWRegister, using WriteMajority and ReadMajority.

The LWWRegister is pretty safe here since the singleton failover is not instant, but we can later create a more customized data type that makes use of akka.cluster.Member.isOlderThan, we want to use the data from the newest singleton. That can be rather useful for singletons in general.

That has the small consistency problems that I sketched in the issue, but perhaps I'm looking too much for the bad scenarios, and perhaps we can make it more safe somehow.

The state of the `PersistenShard` could be stored in one `ORSet` per shard id.

I will be on vacation the next 4 weeks, so if I'm unresponsive here that is not because I'm not interested. Thanks for working on this.

smlin:
@patriknw Thank you too.
I am was created first version of code. Currently it is mostly not completed and not works yet, but I wrote a lot of comments everywhere. And I was not working on PersistentShard because this is a much more work in a row - hope it could be doing step by step. I will be wait for response. 
Enjoy your holiday!
I will be working on this on my vacation at the next week:)

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -387,112 +385,100 @@ class ShardCoordinator(typeName: String, settings: ClusterShardingSettings,
...
-          persistentState = persistentState.updated(evt)
-        case _: ShardHomeDeallocated ⇒
-          persistentState = persistentState.updated(evt)
+  override def receive: Receive = waitForState
+
+  def waitForState: Receive = {
+    case g @ GetSuccess(RegionsKey, _) =>
+      val regions: Map[ActorRef, Set[(ShardId, ActorRef)]] = g.get(RegionsKey).entries
+      regions.foreach { case (region, shards) =>
+        // TODO: Would the Coordinator notified by the `Terminated` message if the region was terminated before `watch`
```
patriknw: yes, it will

smlin:
I was done the first step as @patriknw  offered. I will be move the PersistentShard onto the CRDT in next step, since this is possible. 

Now PR is ready for review. 
@rkuhn Could somebody review this PR as long as Patrik on vacation?

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -15,16 +15,24 @@ import akka.actor.Deploy
...
+import akka.cluster.ddata.Replicator.Get
+import akka.cluster.ddata.Replicator.GetSuccess
+import akka.cluster.ddata.Replicator.GetFailure
+import akka.cluster.ddata.Replicator.ModifyFailure
+import akka.cluster.ddata.Replicator.NotFound
+import akka.cluster.ddata.Replicator.ReadMajority
+import akka.cluster.ddata.Replicator.Update
+import akka.cluster.ddata.Replicator.UpdateSuccess
+import akka.cluster.ddata.Replicator.UpdateTimeout
+import akka.cluster.ddata.Replicator.WriteMajority
```
rkuhn: please use a wildcard import for these

project/AkkaBuild.scala
```diff
@@ -136,9 +136,9 @@ object AkkaBuild extends Build {
     id = "akka-cluster-sharding",
     base = file("akka-cluster-sharding"),
     dependencies = Seq(cluster % "compile->compile;test->test;multi-jvm->multi-jvm",
-        persistence % "compile;test->provided", clusterTools)
+        persistence % "compile;test->provided", distributedData % "compile;test->provided", clusterTools)
```
rkuhn: Is there still a dependency towards persistence?

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -359,127 +368,116 @@ object ShardCoordinator {
...
+      state = g.get(CoordinatorStateKey).value
+      state.regionProxies.foreach(context.watch)
+      state.regions.foreach { case (a, _) ⇒ context.watch(a) }
+      state.shards.foreach { case (a, r) ⇒ sendHostShardMsg(a, r) }
       allocateShardHomes()
+      activate()
+
+    case GetFailure(CoordinatorStateKey, _) ⇒
+      log.error(
+        "The ShardCoordinator was unable to get an initial state within 'waiting-for-state-timeout' (was retried): {} millis",
```
rkuhn: “was retried” may be misunderstood, better would be “retrying”

rkuhn:
Thanks a lot for this idea and PR, @smlin! There are some aspects that I cannot currently think through in full, revolving around the semantics of shard allocation. Currently you send back the result immediately before knowing that the CRDT update has been disseminated. The old implementation makes sure that the allocation has been persisted before it can be handed out and acted upon, which ensures that two different regions cannot create the same shard when the coordinator’s machine fails and a new coordinator makes a different decision.

The other point is related: the old implementation pulled shard location information lazily into the regions when needed, but CRDT replication would allow us to just listen to the same data at all regions, saving roundtrips to the coordinator.

One thing I really like about this approach is that the CRDT’s lifetime is coupled to the lifetime of the cluster itself. Persistence means that allocations can outlive the whole cluster, but when starting from scratch we may as well just reallocate things instead. The only missing feature then is that instantiated entities will not automatically be resurrected, but my thinking goes in the direction of making that an optional add-on since not everybody needs that and then we would have completely separated the sharding from the persistence concerns.

project/AkkaBuild.scala
smlin: Currently it is the "first step" and I removed the persistence just out of the coordinator. But the persistence stay alive inside the PersistenceShard. This is needed when `remember-entities = on`.

smlin:
@rkuhn I have two ways. 
First - I have to fix the "double shards allocation" issue and write some test for checking this. And then I would be start the "next step" (and next PR) at bounds of which the distributed data could be wrote only inside their owners. The data about the allocated shards inside the shards, data about the started regions inside the regions. And of course I will to remove all dependencies onto the persistence.
Second - I can do all of this inside the current PR, but it is a lot of changes in a row. Of course I will do as you will say. But I think the first case is better.

rkuhn:
Yes, I agree that the first proposal is the way to go, optimizing the use of the CRDTs can and should a second step; please create a ticket so that we have it properly on the radar.

akka-cluster-sharding/src/main/resources/reference.conf
```diff
@@ -13,77 +13,83 @@ akka.cluster.sharding {
...
-    
+
     # The number of ongoing rebalancing processes is limited to this number.
     max-simultaneous-rebalance = 3
   }
-  
-  # Settings for the coordinator singleton. Same layout as akka.cluster.singleton.  
+
+  # Timeout of waiting the initial distributed state (an initial state will be queried again if the timeout happened)
+  waiting-for-state-timeout = 30 s
```
patriknw: a more reasonable default is `5 s`

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -359,127 +368,116 @@ object ShardCoordinator {
...
+        updatingStateTimeout.toMillis)
+      sendUpdate(evt)
+
+    case ModifyFailure(CoordinatorStateKey, error, cause, Some(evt: DomainEvent)) ⇒
+      log.error(
+        cause,
+        "The ShardCoordinator was unable to update a distributed state with error {}. Was retried with event {}.",
+        error,
+        evt)
+      sendUpdate(evt)
```
patriknw: ModifyFailure happens only if **you** throw an exception from the the modify function, i.e. it will happen again if you try again. Better to stop or re-throw the cause

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -359,127 +368,116 @@ object ShardCoordinator {
...
         gracefulShutdownInProgress -= region
-        saveSnapshotWhenNeeded()
-        persist(ShardRegionRegistered(region)) { evt ⇒
-          val firstRegion = persistentState.regions.isEmpty
+        val firstRegion = state.regions.isEmpty
 
-          persistentState = persistentState.updated(evt)
-          context.watch(region)
-          sender() ! RegisterAck(self)
+        update(ShardRegionRegistered(region))
```
patriknw: Perhaps continue with watch and RegisterAck when you receive the `UpdateSuccess`? You can use the request (context) parameter in the `Update` message to pass along e.g. `sender`.

If you receive `UpdateTimeout` we should perhaps not retry the update here, since the region also has a retry mechanism.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -359,127 +368,116 @@ object ShardCoordinator {
...
   var aliveRegions = Set.empty[ActorRef]
-  var persistCount = 0
 
   import context.dispatcher
   val rebalanceTask = context.system.scheduler.schedule(rebalanceInterval, rebalanceInterval, self, RebalanceTick)
 
   Cluster(context.system).subscribe(self, ClusterShuttingDown.getClass)
 
+  implicit val node = Cluster(context.system)
+  val CoordinatorStateKey = LWWRegisterKey[State](s"${typeName}CoordinatorState")
```
patriknw: Later we could try to implement the member age aware register, using the `akka.cluster.Member.isOlderThan`, we want to use the data from the newest singleton. That can be rather useful for singletons in general.

If you want to take a stab at that new data type you find useful info here: http://doc.akka.io/docs/akka/snapshot/scala/distributed-data.html#Custom_Data_Type
and the following commit is a good example of what pieces that are required: 7041c76ba910e4474b1004060c5cbc08a121b5e1

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -598,69 +587,74 @@ class ShardCoordinator(typeName: String, settings: ClusterShardingSettings,
...
+    state = state.updated(evt)
+    sendUpdate(evt)
+  }
+
+  def getState(): Unit =
+    replicator ! Get(CoordinatorStateKey, ReadMajority(waitingForStateTimeout))
+
+  def sendUpdate(evt: DomainEvent) =
+    replicator ! Update(CoordinatorStateKey, LWWRegister(State.empty), WriteMajority(updatingStateTimeout), Some(evt)) { reg ⇒
+      reg.withValue(reg.value.updated(evt))
```
patriknw: shouldn't this be `reg.withValue(state)`?

I think the approach of applying the event again is problematic when you retry the updates. ordering, duplicate operations?

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -359,127 +368,116 @@ object ShardCoordinator {
...
+      activate()
+
+    case GetFailure(CoordinatorStateKey, _) ⇒
+      log.error(
+        "The ShardCoordinator was unable to get an initial state within 'waiting-for-state-timeout' (was retried): {} millis",
+        waitingForStateTimeout.toMillis)
+      getState()
+
+    case NotFound(CoordinatorStateKey, _) ⇒ activate()
+    case m                                ⇒ stash()
```
patriknw: Stashing here is a bit dangerous since it might take a long time until we receive a successful Get(majority), e.g. when nodes are unreachable (but not yet removed from cluster). Can't we just drop all other messages? They are resent anyway, right?

patriknw:
@smlin I took a first look. I think it is in the right direction. I appreciate the step-wise approach you take on this. Great work!

To set the expectations right, we will probably have to support both this and the persistent sharding and we need to discuss how to package those. Anyway, don't worry about that now. Please continue in this PR, and separate the different steps with separate commits.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
patriknw: I mean:

    val s = state
    replicator ! Update(...) { reg ⇒
      reg.withValue(s)
    }

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
smlin: Not exactly so. It depends on state inside - if after the failure there was another successful update then at next time the failure could be not occurs.
Anyway I think errors in logs slightly better then kill cluster-sharding at all.

smlin:
@rkuhn @patriknw I am stopped on consistency requirements to case of duplicated shards. To avoid such duplication a coordinator must ensure changes of state to a whole cluster (WriteAll in case of CRDTs) or synchronize state with all regions before moving to the active state. But in second step I will change the source of distributed state changes from a coordinator to the regions and shards. So after second step duplication will be avoided automatically "by design". 
Now question: is WriteMajority enough for merge this PR or it must be implemented with the guaranteed consistency?
Currently I think that this step is finished and need for review before start the second step.

smlin:
I have to notice one case when avoiding the shards duplication is impossible. If cluster was fragmented (due to broken link between two datacenters or some other cases), then such situation would be end up with the two identical coordinators in the both clusters. 
In case when coordinator extends the PersistentActor this will lead to crash both coordinators (due concurrent writes by two persistent actors) and could be repaired only by cleaning persistent data from database. 
In case of CRDT this eventually would lead to stand up two identical shards/entities and could be crash these entities if they are persistent actors.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -359,127 +359,129 @@ object ShardCoordinator {
...
+          sendUpdate(evt)
+      }
+
+    case ModifyFailure(CoordinatorStateKey, error, cause, Some(evt: DomainEvent)) ⇒
+      log.error(
+        cause,
+        "The ShardCoordinator was unable to update a distributed state with error {}. Was retrying with event {}.",
+        error,
+        evt)
+      sendUpdate(evt)
```
patriknw: `ModifyFailure` only happens if you modify function in the `Update` message throws an exception, and that should never happen unless you have done a programming error. No point in trying to do it again. Will probably fail again. Instead you should just re-throw the cause.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -359,127 +359,129 @@ object ShardCoordinator {
...
-        case ShardRegionProxyTerminated(proxy) ⇒
-          if (persistentState.regionProxies.contains(proxy))
-            persistentState = persistentState.updated(evt)
-        case ShardHomeAllocated(shard, region) ⇒
-          persistentState = persistentState.updated(evt)
-        case _: ShardHomeDeallocated ⇒
-          persistentState = persistentState.updated(evt)
+          val firstRegion = state.regions.isEmpty
+
+          state = state.updated(evt)
```
patriknw: The state was already updated in the [update](https://github.com/akka/akka/pull/17871/files#diff-e3b09e26a65bebe3b08cd0b802fde5e8R601) method.

The order of the events applied to the state is important. The original order can be changed if we apply the event here, because of the [retry](https://github.com/akka/akka/pull/17871/files#diff-e3b09e26a65bebe3b08cd0b802fde5e8R445) and that UpdateSuccess messages are not guaranteed to be received in the same order as corresponding Update messages were emitted.

I see two options.
1) Apply the even and set the local state before sending Update to the replicator. Some of the code here must also be moved to that point. No retry. We rely on the gossip replication in case WriteMajority timeout.
2) We apply the event and set the local state here instead of doing it before sending Update to the replicator. We retry until UpdateSuccess. We stash all incoming messages until we receive UpdateSuccess, i.e. we perform only one update at a time.

The second alternative is more safe in case of fail over to another coordinator instance.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -359,127 +359,129 @@ object ShardCoordinator {
...
+      activate()
+
+    case GetFailure(CoordinatorStateKey, _) ⇒
+      log.error(
+        "The ShardCoordinator was unable to get an initial state within 'waiting-for-state-timeout' (was retrying): {} millis",
+        waitingForStateTimeout.toMillis)
+      getState()
+
+    case NotFound(CoordinatorStateKey, _) ⇒ activate()
+  }
```
patriknw: perhaps a comment that it is alright to drop all other messages, they will be retried

patriknw:
> then such situation would be end up with the two identical coordinators in the both clusters

We must assume that cluster singleton keeps it promise to be a singleton. There is not much we can do if  the cluster is split into two separate clusters each hosting its own singleton instance. That will not happen by the network partition itself, but by the used downing strategy. That is why we recommend against using auto-downing when using Akka Persistence. We are attacking that issue from other angles.

We never merge half complete solutions, so please continue to update this PR with more commits for the additional steps.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
smlin: Could it happens due to message/updates reordering? And would fixed after some time.

smlin:
> We are attacking that issue from other angles.
Am I right understand this mean that issue must not be solved in this PR?

patriknw:
> Am I right understand this mean that issue must not be solved in this PR?

yes

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
patriknw: I don't think so, typical error would be things like `NullPointerException`, i.e. it should never happen unless we have a bug somewhere and then we don't want to continue anyway

smlin:
@patriknw It turns out that ddata realization (without straightforward message exchanges between coordinator, regions and shards) is tremendously slow since gossips have a timeout. So I see two variants:

1) The `Replicator` must sending gossips immediately to the nodes where subscribers exists.
2) Currently this PR seems to be working and I think it is anyway better then saving a state to disk. 
Using a ddata not only for the coordinator state gives slightly more safety and no more.

Probability of the shards duplication even at the moments of cluster failures seems to be very small. I don't know exactly, but it possible that such failure could happen only if `auto-down` enabled. But if it is enabled then saving the coordinator state to disk also have the problems.

Currently I think it is complete solution. Any better solution required absolutely different idea and realization. Sorry, I have no a time to do such realization.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -155,6 +155,12 @@ object ShardRegion {
   @SerialVersionUID(1L) final case object GracefulShutdown extends ShardRegionCommand
 
   /**
+   * We must be sure that a shard is initialized before to start send messages to it.
```
patriknw: what is the reason for the changes in this file?

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
```diff
@@ -96,7 +96,7 @@ private[akka] class Shard(
   extractShardId: ShardRegion.ExtractShardId,
   handOffStopMessage: Any) extends Actor with ActorLogging {
 
-  import ShardRegion.{ handOffStopperProps, EntityId, Msg, Passivate }
+  import ShardRegion.{ handOffStopperProps, EntityId, Msg, Passivate, ShardInitialized }
```
patriknw: what is the reason for the changes in this file?

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -359,127 +359,117 @@ object ShardCoordinator {
  * @see [[ClusterSharding$ ClusterSharding extension]]
  */
 class ShardCoordinator(typeName: String, settings: ClusterShardingSettings,
-                       allocationStrategy: ShardCoordinator.ShardAllocationStrategy)
-  extends PersistentActor with ActorLogging {
+                       allocationStrategy: ShardCoordinator.ShardAllocationStrategy,
+                       replicator: ActorRef)
+  extends Actor with ActorLogging with Stash {
```
patriknw: In 2.4 we need to support both the old persistent implementation and the new ddata implementation. The reason is that akka-cluster-sharding is not experimental and therefore it cannot by default use the experimental distributed-data module.

Would it be possible to extract the specific persistent/ddata parts to separate subclasses? AbstractShardCoordinator, PersistentShardCoordinator, DDataShardCoordinator.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -598,69 +579,76 @@ class ShardCoordinator(typeName: String, settings: ClusterShardingSettings,
     case _: CurrentClusterState ⇒
   }
 
+  def activate() = {
+    context.become(active)
+    unstashAll()
```
patriknw: is stash used at all?

patriknw:
@smlin I agree that the solution in this PR is good and it is a great advantage that it is so similar to current persistent implementation, since we need to support both approaches in 2.4 (as described in my comment above). Would you be able to wrap up that part or would you like that I make those adjustments?

smlin:
I will try do it by myslef. Thank you for review.

patriknw:
Thanks!

patriknw:
To clarify my idea. We should make it possible to switch between the two implementations with a config setting. The dependency to distributed-data should be in scope provided, so the user must define the dependency and switch the config setting to use this implementation. Does that make sense?

smlin:
Yes, it make sense. Perhaps now it could be done also for akka-persistence dependency.

smlin:
@patriknw I wrote two implementations of the coordinator which depends on config parameter. I duplicated tests for ddata coordinator - currently I don't know how make it more elegantly. Possible you could help me.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
patriknw: I would still like to understand why this was changed

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -393,62 +396,19 @@ class ShardCoordinator(typeName: String, settings: ClusterShardingSettings,
...
-        persist(ShardRegionRegistered(region)) { evt ⇒
-          val firstRegion = persistentState.regions.isEmpty
-
-          persistentState = persistentState.updated(evt)
+        update(ShardRegionRegistered(region)) { evt ⇒
+          val firstRegion = state.regions.isEmpty
+          state = state.updated(evt)
           context.watch(region)
-          sender() ! RegisterAck(self)
+          region ! RegisterAck(self)
```
patriknw: sender() and region is the same, right?
but now we use region here, but sender() a few lines above

patriknw:
@smlin excellent separation! I have a few comments, but it looks like this can be merged soon. We are releasing on 2.4-RC1 on Friday so it would be great if we can merge it on Thursday at the latests.

Could you also add a section in the docs of the alternative ddata mode:
* akka-docs/rst/scala/cluster-sharding.rst
* akka-docs/rst/java/cluster-sharding.rst

Regarding the tests I think you could split them up into an abstract class and two implementations. The only difference should be the config (and some initial creation of actors). See example of similar split:

* https://github.com/akka/akka/blob/master/akka-persistence/src/test/scala/akka/persistence/PersistentActorSpec.scala#L1139
* https://github.com/akka/akka/blob/master/akka-cluster/src/multi-jvm/scala/akka/cluster/ClientDowningNodeThatIsUnreachableSpec.scala#L22-L34

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
smlin: Sorry, I forgot answer to you here.
Sometimes persistent shard could be failure during a replay process due to temporary journal unavailability. The user messages which sent to such shard just after creation, but before a failure would be lost. This code removes that problem.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -302,11 +308,13 @@ class ShardRegion(
 
   def receive = {
     case Terminated(ref)                         ⇒ receiveTerminated(ref)
+    case ShardInitialized(shardId)               ⇒ initializeShard(shardId, sender())
     case evt: ClusterDomainEvent                 ⇒ receiveClusterEvent(evt)
     case state: CurrentClusterState              ⇒ receiveClusterState(state)
     case msg: CoordinatorMessage                 ⇒ receiveCoordinatorMessage(msg)
     case cmd: ShardRegionCommand                 ⇒ receiveCommand(cmd)
     case msg if extractEntityId.isDefinedAt(msg) ⇒ deliverMessage(msg, sender())
+    case msg: RestartShard                       ⇒ deliverMessage(msg, sender())
```
smlin: The `RestartShard` message handling has not existed before - it is a bug fix.

patriknw: great catch, it's scheduled by `receiveTerminated`

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
patriknw: good catch
note for other reviewers: this is handled in ShardRegion.scala

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
patriknw: as mentioned elsewhere: Sometimes persistent shard could be failure during a replay process due to temporary journal unavailability. The user messages which sent to such shard just after creation, but before a failure would be lost. This code removes that problem.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -560,9 +590,10 @@ class ShardRegion(
...
           name))
-        shards = shards.updated(id, shard)
         shardsByRef = shardsByRef.updated(shard, id)
-        shard
+        None
+      case Some(props) ⇒
+        None
       case None ⇒
         throw new IllegalStateException("Shard must not be allocated to a proxy only ShardRegion")
     })
```
patriknw: The changes in this file looks correct

patriknw:
LGTM
@smlin please squash into one commit and I will merge it. Excellent contribution!

I will add a few notes about experimental and such in the docs, but I can take care of that.

smlin:
@patriknw Squashed.

patriknw:
thanks!

patriknw:
Refs #17846

Issue #17729, opened 2015-06-15
PermGen leak


patriknw:
When shutting down actor system.
Reported here: https://groups.google.com/d/msg/akka-user/XDVyWUT2u7E/o4jWaq9Le5cJ

patriknw:
I have not confirmed the leak

krchniam:
Sorry for the delay, I've created sample project: https://github.com/krchniam/akka-remote-bug-17729.

When deployed on GlassFish 3.1.2.2. After subsequent disabling application, there is one leaked thread from ExecutionContext.Implicits.global (two threads are used during shutdown, ForkJoinPool-3-worker-7 remains in state WAITING).

patriknw:
thanks, is it possible to reproduce the problem without any external dependencies?

krchniam:
I tried to cut dependencies to minimum (akka with all it`s dependencies and slf4j), so project runs on every JEE application server. My point is that, redeploying application with akka remoting on JEE server is causing PermGen leak (thread is holding WebAppClassLoader or EarLibClassLoader).

patriknw:
ok, I was looking for something that could reproduce it without JEE app server.
Thanks anyway, we will look into it. You are welcome to submit a pull request if you know what is wrong.

krchniam:
I'm not sure why is ```Remoting#shutdown``` using ```import scala.concurrent.ExecutionContext.Implicits.global``` instead of ```import system.dispatcher``` (if it is still running during remoting shutdown). That imports creates global ```ExecutionContext```. When akka is shut down, ```system.dispatcher``` thread pool will be shut down. But ```scala.concurrent.ExecutionContext.Implicits.global``` remains active with one thread holding ClassLoader.  My workaround shutting down ```scala.concurrent.ExecutionContext.Implicits.global``` via ```ActorSystem``` termination callback.

drewhk:
I think it was about system.dispatcher not running at that point, but we need to revisit this. Using Implicits.global is definitely not the way to go. Thanks for reporting!

patriknw:
@krchniam I have tried your sample with latest Glassfish 4.1, and I can confirm that number of live WebappClassLoader instances increase with one for each re-deployment and that used metaspace is increasing.

It is not solved by replacing the use of `Implicits.global`. I have replaced it with `system.dispatcher` in the `ShutdownAndFlush` usage in remoting. I verifed that the callbacks are actually running with the system.dispatcher. We should do that change anyway.

[...]

PR #17876, opened 2015-07-01
=rem #17729 Don't use Implicits.global in remoting shutdown

patriknw:
I have verified (with printlns) that the callback is running with the system.dispatcher in all cases when running the tests in akka-remote.

### DIFF
--- a/akka-remote/src/main/scala/akka/remote/Remoting.scala
+++ b/akka-remote/src/main/scala/akka/remote/Remoting.scala
@@ -136,7 +136,6 @@ private[remote] class Remoting(_system: ExtendedActorSystem, _provider: RemoteAc
     eventPublisher.notifyListeners(RemotingErrorEvent(new RemoteTransportException(msg, cause)))
 
   override def shutdown(): Future[Unit] = {
-    import scala.concurrent.ExecutionContext.Implicits.global
     endpointManager match {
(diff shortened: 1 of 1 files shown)


akka-remote/src/main/scala/akka/remote/Remoting.scala
```diff
@@ -146,6 +145,7 @@ private[remote] class Remoting(_system: ExtendedActorSystem, _provider: RemoteAc
           endpointManager = None
         }
 
+        import system.dispatcher
```
drewhk: is it still alive and kicking here?

patriknw: As far as I can see, it is. endpointManager is an actor.
I have verified (with printlns) that the callback is running with the system.dispatcher in all cases when running the tests in akka-remote.

patriknw:
Refs #17729

akka-remote/src/main/scala/akka/remote/Remoting.scala
drewhk: Yes, I think the caller of this shutdown() method actually blocks on an Await, so this is probably safe.

drewhk:
LGTM

ktoso:
LGTM, nice detective work!

Issue #17712, opened 2015-06-13
NPE if cluster creation fails


patriknw:
If the cluster core creation fails due to await timeout it tries to shutdown, and then the readView is null.

    Exception in thread "main" java.lang.NullPointerException
    #011at akka.cluster.Cluster.shutdown(Cluster.scala:358)
    #011at akka.cluster.Cluster.liftedTree1$1(Cluster.scala:176)
    #011at akka.cluster.Cluster.<init>(Cluster.scala:171)

PR #17878, opened 2015-07-01
=clu #17712 Guard against readView not created

patriknw:
- readView might be null if Cluster init fails before it is created,
  i.e. shutdown is called from constructor


patriknw:
Refs #17712

ktoso:
LGTM

PR #17880, opened 2015-07-01
Normalize whitespace for resource paths

### LINKED ISSUE #17873: http: getFromResourceDirectory fails if the absolute path to the server contains a space
Example:
```scala
pathPrefix("assets") {
  getFromResourceDirectory("public")
}
```
Note the space in `/path/player 2/lib/...`:
```
[ERROR] [06/30/2015 23:28:01.375] [player-akka.actor.default-dispatcher-13] [ActorSystem(player)] Error during processing of request HttpRequest(HttpMethod(GET),http://localhost:8080/assets/stylesheets/common.css,...)
java.io.FileNotFoundException: /path/player%202/lib/player_2.11-0.1-SNAPSHOT-web-assets.jar (No such file or directory)
	at java.util.zip.ZipFile.open(Native Method)
	at java.util.zip.ZipFile.<init>(ZipFile.java:220)
	at java.util.zip.ZipFile.<init>(ZipFile.java:150)
	at java.util.zip.ZipFile.<init>(ZipFile.java:121)
	at akka.http.scaladsl.server.directives.FileAndResourceDirectives$ResourceFile$.apply(FileAndResourceDirectives.scala:210)
```

This is happening in this method: https://github.com/akka/akka/blob/releasing-akka-stream-and-http-experimental-1.0-RC4/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala#L199.

Looks like it’s a [known problem with `j.u.z.ZipFile`](https://code.google.com/p/reflections/issues/detail?id=99). I was able to reproduce it on Windows 7 and on Ubuntu. *akka-http* version `RC4`.

aruediger:
Hard to unit test since I had to temporarily add a jar file with spaces to my path.

fixes #17873


stanch:
Have you tried with other escaped characters?

2beaucoup:
No. If you have a test case I can check locally.

stanch:
I don’t have anything distilled at the moment, but I can craft it in the evening (GMT). Anyway, it’s not necessary to modify the jar filenames, because the path is absolute. It’s enough to put the app itself into a directory with spaces or non-latin characters. Some Cyrillic stuff should expose the original issue: `/path/to/тест/app/...`.

2beaucoup:
Cyrillic doesn't work on my machine but umlauts do. I can only test on Win7 with German locale ATM.

I'm working via unit tests so there is no real "app" here. I can simulate the behaviour with the test classpath though.

stanch:
> Cyrillic doesn't work on my machine but umlauts do.

Do you mean that the original issue manifests itself with Cyrillic? Or some other issue?

> I'm working via unit tests so there is no real "app" here. I can simulate the behaviour with the test classpath though.

In that case the app will be the test runner :) You can change the name of the directory where your akka code is contained. But I do understand that it’s not something you can control from within the tests to make them reproducible. I wonder if it’s possible to insert some wacky characters into the package names to the same effect (not sure if that’s permitted by Java/Scala).

2beaucoup:
You just have to create a jar with a funky name, put it in an even funkier directory, add it to the classpath and load things from it. ;)

The issue with Cyrillic is, that the class loader doesn't even recognise classes and resources contained in such a jar/folder and the ClassLoader.getResource always returns null in that case.

stanch:
> The issue with Cyrillic is, that the class loader doesn't even recognise classes and resources contained in such a jar/folder and the ClassLoader.getResource always returns null in that case.

What if the folders inside the jar have latin names, but the outside folder has a Cyrillic name, like in my example? I know it will probably not help establishing a test case for this, but bear with me for a moment :)

stanch:
@2beaucoup Could you try running this sample with your version of `akka-http-experimental`? https://github.com/stanch/akka-http-spaces

2beaucoup:
@stanch let's try the other way around: why don't you just clone https://github.com/2beaucoup/akka and `publishLocal` the `space-in-path` branch?

stanch:
That was the first thing I tried, but I had problems building akka. I will try again tonight, I just thought that running my sample would take much less time :)

2beaucoup:
```
> Some(ResourceFile(jar:file:/[...]/akka-http-spaces/?????????%20?????/lib/test.jar!/test.txt,0,1435790924000))
```

stanch:
That’s great! Thanks for running it :)

2beaucoup:
added a test for this

2beaucoup:
rebased. could someone plz trigger the tests?

ktoso:
OK TO TEST

stanch:
> I had problems building akka

For the record, I was just missing `graphviz`, but it was too late in the night for me to realize it. Other than that, I have to say that for such a big project akka is surprisingly easy to build!

2beaucoup:
> For the record, I was just missing graphviz, but it was too late in the night for me to realize it.

https://github.com/akka/akka/blob/master/akka-docs/rst/dev/building-akka.rst#scaladoc-dependencies ;)

stanch:
@2beaucoup Thanks! I think this should be at the root of the repo ;)

stanch:
@2beaucoup I tried to do `sbt +publishLocal` as well and got this error. Any ideas?

```
[error] /home/stanch/projects/akka/akka-http-core/src/main/scala/akka/http/impl/util/JavaMapping.scala:100: method with dependent type (scalaObject: SIn)inMapping.J cannot be converted to function value
[error]         scaladsl.Flow[SIn].map(inMapping.toJava).viaMat(javaObject)(scaladsl.Keep.right).map(outMapping.toScala)
[error]                                          ^
[error] /home/stanch/projects/akka/akka-http-core/src/main/scala/akka/http/impl/util/JavaMapping.scala:103: method with dependent type (javaObject: JIn)inMapping.S cannot be converted to function value
[error]           scaladsl.Flow[JIn].map(inMapping.toScala).viaMat(scalaObject)(scaladsl.Keep.right).map(outMapping.toJava)
[error]                                            ^
[info] No documentation generated with unsucessful compiler run
[error] two errors found
[info] Compiling 123 Scala sources and 114 Java sources to /home/stanch/projects/akka/akka-http-core/target/classes...
[error] /home/stanch/projects/akka/akka-http-core/src/main/scala/akka/http/impl/util/JavaMapping.scala:100: method with dependent type (scalaObject: SIn)inMapping.J cannot be converted to function value
[error]         scaladsl.Flow[SIn].map(inMapping.toJava).viaMat(javaObject)(scaladsl.Keep.right).map(outMapping.toScala)
[error]                                          ^
[error] /home/stanch/projects/akka/akka-http-core/src/main/scala/akka/http/impl/util/JavaMapping.scala:103: method with dependent type (javaObject: JIn)inMapping.S cannot be converted to function value
[error]           scaladsl.Flow[JIn].map(inMapping.toScala).viaMat(scalaObject)(scaladsl.Keep.right).map(outMapping.toJava)
[error]                                            ^
[error] two errors found
```

Sorry if I’m derailing this discussion :)

2beaucoup:
@jrudolph any objections from your side?

2beaucoup:
@stanch what exactly are you trying to publish?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -207,12 +207,11 @@ object FileAndResourceDirectives extends FileAndResourceDirectives {
         if (file.isDirectory) None
         else Some(ResourceFile(url, file.length(), file.lastModified()))
       case "jar" ⇒
-        val jarFile = url.getFile
-        val startIndex = if (jarFile.startsWith("file:")) 5 else 0
-        val bangIndex = jarFile.indexOf("!")
-        val jarFilePath = jarFile.substring(startIndex, bangIndex)
-        val resourcePath = jarFile.substring(bangIndex + 2)
-        val jar = new java.util.zip.ZipFile(jarFilePath)
+        val path = new URI(url.getPath).getPath // remove "file:" prefix and normailze whitespace
```
jrudolph: s/normailze/normalize/

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
```diff
@@ -104,12 +104,18 @@ class FileAndResourceDirectivesSpec extends RoutingSpec with Inspectors with Ins
...
-    "reject requests to directory resources from an Archive " in {
+    "reject requests to directory resources from an archive " in {
       Get() ~> getFromResource("com/typesafe/config") ~> check { handled shouldEqual (false) }
     }
-    "reject requests to directory resources from an Archive with trailing slash" in {
+    "reject requests to directory resources from an archive with trailing slash" in {
       Get() ~> getFromResource("com/typesafe/config/") ~> check { handled shouldEqual (false) }
     }
+    "return the resource from an archive with spaces and umlauts" in {
+      Get() ~> getFromResource("test-resource.txt") ~> check {
```
jrudolph: Maybe add a comment that it will load this file from the jar file above. Easy enough to spot in the diff but not so easy if you just look at the test.

jrudolph:
LGTM aside from the small things.

stanch:
> @stanch what exactly are you trying to publish?

This:

> @stanch let's try the other way around: why don't you just clone https://github.com/2beaucoup/akka and publishLocal the space-in-path branch?

```
git clone git@github.com:2beaucoup/akka.git && \
  cd akka && \
  git checkout space-in-path && \
  sbt +publish-local
```

2beaucoup:
typo fixed, comment added and updated. thx @jrudolph!

2beaucoup:
@stanch cross-publishing fails for me with the same error. 2.10-only works though.

jrudolph:
Maybe it really is broken, the failing lines are still somewhat new.

jrudolph:
I created a ticket to fix that: https://github.com/akka/akka/issues/17943

stanch:
@jrudolph Thanks! Shouldn’t Jenkins report that? Or it’s only testing with `2.10`?

jrudolph:
Can you try the simple fix here: https://github.com/akka/akka/pull/17944? I couldn't reproduce the problem, how did you cross-build to get the error?

2beaucoup:
@jrudolph just add the `+` prefix like @stanch did [here](https://github.com/akka/akka/pull/17880#issuecomment-119575037).

stanch:
Actually I think `crossScalaVersions` are not defined, so `+publich-local` is compiling with a default (old) version of `2.10`, hence the errors.

jrudolph:
@stanch Yep, that's what I just figured as well. So, I just wrongly inferred that you must mean it fails under 2.11 when it was failing under 2.10.3.

stanch:
What do you think about adding a cross-build config to avoid these surprises?

2beaucoup:
Confirmed. (Just build successfully against 2.11.7)

drewhk:
LGTM. Is this mergeable? @jrudolph ?

jrudolph:
Yep, LGTM.

jrudolph:
Yay! Thanks once again @2beaucoup.

2beaucoup:
always a pleasure @jrudolph

Issue #15404, opened 2014-06-14
Cluster.subscribe should not accept zero events


hseeberger:
Currently one can use subscribe to cluster domain events without specifying any event type:

``` scala
Cluster(context.system).subscribe(listener)
```

Some might expect to subscribe to all possible events, i.e. to `classOf[ClusterDomainEvent]`. But what happens is "nothing".

Therefore the signature should be changed to take a single `Class` parameter first and then repeated ones.

PR #17882, opened 2015-07-01
=clu #15404 Require at least one subscribe class

patriknw:


patriknw:
Refs #15404

akka-cluster/src/main/scala/akka/cluster/Cluster.scala
```diff
@@ -223,8 +223,10 @@ class Cluster(val system: ExtendedActorSystem) extends Extension {
    *
    * Note that for large clusters it is more efficient to use `InitialStateAsSnapshot`.
    */
-  @varargs def subscribe(subscriber: ActorRef, initialStateMode: SubscriptionInitialStateMode, to: Class[_]*): Unit =
+  @varargs def subscribe(subscriber: ActorRef, initialStateMode: SubscriptionInitialStateMode, to: Class[_]*): Unit = {
+    require(to.length > 0, "at least one `ClusterDomainEvent` class is required")
```
hepin1989: check the class like 
to.foreall(classof[ClusterDomainEvent].isAssignableFrom) too ?

patriknw: great suggestion, that actually revealed a "bug" in ClusterMetricsDisabledSpec
see 7fbaaa110445df21de2671e58f20d6fdc6a2f5ab

Issue #17805, opened 2015-06-22
Don't clear self when restarting


msfrank:
documentation page /akka-docs/rst/general/jmm.rst has the following incorrect code block:

```
 18   // Completely safe, "self" is OK to close over
 19   // and it's an ActorRef, which is thread-safe
 20   Future { expensiveCalculation() } onComplete { f => self ! f.value.get }
```

this should be removed or explicitly marked as an incorrect usage, not correct usage.

ktoso:
Closing over self is legal and valid.
What you probably mean is closing over sender, which indeed is not valid, but that's something else.

drewhk:
@ktoso, see the discussion on the ML

drewhk:
i.e. after the original actor died, calling the closed over self() will result in DeadLetters being returned. This might or might not be fine (after all, even if you share the result of self() the message will end up in DeadLetters anyway). While not completely incorrect, the documentation is misleading.

ktoso:
Ok, cought up with the mailing list, I see what is meant here now.

Yeah we can improve it on it.
Thanks for bringing it up!

patriknw:
The problem is restarting. I agree that it is unintuitive that self changes because of a restart, given that it is defined as a `val`. I wonder if the nulling isn't too aggressive. We should investigate if that nulling is really needed for restarts.

drewhk:
but wait, isn't self defined on the ActorCell? That is not nulled when restarted, it is actually a val in Cell.

drewhk:
And in the trait Actor:

    implicit final val self = context.self //MUST BE A VAL, TRUST ME

patriknw:
and it is not nulled by reflection?

then I must have misunderstood the reproducer provided in the mailing list

drewhk:
Ah, it does via reflection. Sneaky. But not to null, but deadLetters. I don't know why we do that for restarts though, I don't see what it buys.

patriknw:
eactly

fredfp:
I'm the original poster of the ML thread.

As @patriknw pointed out, my main problem was indeed restarting.

If the restarting behavior is not changed, I'd like to see the doc updated to illustrate how to send a message to self in a Future's callback. It is something I do very often in my code and I always referred to the section of the doc quoted above stating that it is safe to close over self.

[...]

PR #17888, opened 2015-07-01
=act #17805 Don't change self on restart"

patriknw:
- we hold on to the same ActorRef instance anyway in the new
  Actor instance

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/ActorCell.scala
+++ b/akka-actor/src/main/scala/akka/actor/ActorCell.scala
@@ -567,7 +567,7 @@ private[akka] class ActorCell(
   protected def create(failure: Option[ActorInitializationException]): Unit = {
     def clearOutActorIfNonNull(): Unit = {
       if (actor != null) {
-        clearActorFields(actor)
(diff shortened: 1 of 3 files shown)


patriknw:
Refs #17805

drewhk:
LGTM

patriknw:
do you think this is important enough to backport to 2.3.12?

please double check that it is ok to change the signature of this internal method and still be binary compatible

drewhk:
I don't think there is a danger of binary compatibility here, but you can just add an overloaded method directly if you want to please MiMa.

patriknw:
my reasoning why it is safe is that the signature change is in ActorCell and the only place we invoke it is from the FaultHandling mixin, and those two should always be compiled together
`trait FaultHandling { this: ActorCell`

ktoso:
Looks valid to me, LGTM

ktoso:
I'd vote for backporting as well, it's low effort and keeps the behaviour between the branches in sync which may make things simpler in support.

patriknw:
ok, I'll fix the backport

Issue #17770, opened 2015-06-18
More typesafety for ReplicatedData#merge


ktoso:
As discussed in https://github.com/akka/akka/pull/17499/files#r32720845
It is possible to make the merge function on a distributed data type "only merge apples with apples, and oranges with oranges" by using *self recursive type parameters* (F-bounds).

Here's some docs I did on the pattern: http://ktoso.github.io/scala-types-of-types/#self-recursive-type

```scala
class Data[T] extends Data[Data[T]] {
  def merge(d: T): T = ???
}
```

usage in java:
```java
  static class X extends Data<X> { }
  static class Y extends Data<Y> { }
  {
    X x = new X();
    Y y = new Y();

    x.merge(x);
    y.merge(y);
    x.merge(y); // good, compilation fails: merge(X) cannot be applied to Y
    y.merge(x); // good, compilation fails: merge(Y) cannot be applied to X
  }
```

So this could be improved in next versions for Java I think :-)
Maybe a ticket?

This would possibly make it possible to get rid of `AbstractReplicatedData` and only have `ReplicatedData`, while improving type-safety.

patriknw:
I tried with `abstract class AbstractReplicatedData[D <: AbstractReplicatedData[D]] {`

I'm stuck on one annoying thing when merging the values in `ORMap`. 
https://github.com/akka/akka/commit/482aca115081f95af1e4fabdcebcb4ce7c927094#diff-e15b95357b14e8543fb6bab6f7408abdR165
`type mismatch; found : thatValue.type (with underlying type A) required: _$6`

I can't find a way to work around it with casts, either.

PR #17893, opened 2015-07-02
!cdd #17770 Use self recursive type for ReplicatedData

patriknw:


patriknw:
This is only a prototype. Doesn't work yet.

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORMap.scala
```diff
@@ -164,8 +162,9 @@ final class ORMap[A <: ReplicatedData] private[akka] (
               s"[${thisValue.getClass.getName}], got [${thatValue.getClass.getName}]"
             throw new IllegalArgumentException(errMsg)
           }
+          val mergedValue = thisValue.merge(thatValue)
           // TODO can we get rid of these (safe) casts?
-          val mergedValue = thisValue.merge(thatValue.asInstanceOf[thisValue.T]).asInstanceOf[A]
+          //          val mergedValue = thisValue.merge(thatValue.asInstanceOf[thisValue.T]).asInstanceOf[A]
```
patriknw: This is the problem that I'm stuck on. Merging the values in ORMap. 

    type mismatch; found : thatValue.type (with underlying type A) required: _$6

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORMap.scala
```diff
@@ -33,12 +33,10 @@ object ORMap {
  * This class is immutable, i.e. "modifying" methods return a new instance.
  */
 @SerialVersionUID(1L)
-final class ORMap[A <: ReplicatedData] private[akka] (
+final class ORMap[A <: AbstractReplicatedData[_]] private[akka] (
```
viktorklang: Should that be `[A <: AbstractReplicatedData[A]` ?

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORMap.scala
```diff
@@ -33,12 +33,10 @@ object ORMap {
...
   private[akka] val keys: ORSet[String],
   private[akka] val values: Map[String, A])
-  extends ReplicatedData with ReplicatedDataSerialization with RemovedNodePruning {
-
-  type T = ORMap[A]
+  extends AbstractReplicatedData with ReplicatedDataSerialization { // FIXME with RemovedNodePruning {
```
viktorklang: No type param for AbstractReplicatedData?

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORMap.scala
ktoso: yeah, in other places as well, I'm helping out on getting this PR up to type-soundness right now :-)

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORMap.scala
patriknw: this might what caused the confusion in merge, thanks

patriknw:
@ktoso It turns out that the internals in the Replicator just blows up with ugliness, existential types, and other things that I don't want there. Here is a good compromise. Using self recursive type only for the Java API, `AbstractReplicatedData`.

patriknw:
This PR is not ready. I will also update the docs, and complete the TwoPhaseSet documentation example (serializer and stuff)

ktoso:
I see, yeah you would have to carry around the ugly `A <: ...` everywhere I guess.
Makes sense to use it only for japi then.

patriknw:
@ktoso I have completed the TwoPhaseSet documentation sample, which illustrates how to write a custom data type in java, including serialization. See bde5a06

ktoso:
LGTM, awesome to see it worked out with the recursive type :)

patriknw:
yeah, thanks for the great idea

patriknw:
Refs #17770

PR #17899, opened 2015-07-02
=str #17853 chain terminationEmit elements to the currently emitting ones

### LINKED ISSUE #17853: StatefulStage terminationEmit sometimes replaces previously emitted elements
The following code fails on Akka Streams 1.0-RC4 (with Akka 2.3.11 and Scala 2.11.6).

I would expect the elements given to `terminationEmit` to be emitted after the elements given in the last `emit`. This works as expected when there is enough demand to emit all elements immediately. However in the case below there is only enough demand for the first element and the upstream finishes before the second element is emitted. The result is that the second element is never emitted, replaced by the `terminationEmit`.

```scala
def testStage() = new StatefulStage[String, String] {
  override def initial = new State {
    override def onPush(elem: String, ctx: Context[String]) =
      emit(Iterator(elem, elem), ctx)
  }
  override def onUpstreamFinish(ctx: Context[String]) =
    terminationEmit(Iterator("byebye"), ctx)
}

Source.single("hi")
  .transform(testStage)
  .runWith(TestSink.probe[String]())
  .request(1) // works when requesting more than 1 here
  .expectNext("hi")
  .request(2)
  .expectNext("hi")
  .expectNext("byebye")
  .expectComplete()
```

2m:
Fixes #17853

This little code change took a while to come up with. At first I tried chaining `andThen`s, but stateful stage does not do any more state transitions when `ctx.isFinishing`. So I resorted to joining iterators.

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowStageSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowStageSpec.scala
@@ -433,6 +433,25 @@ class FlowStageSpec extends AkkaSpec(ConfigFactory.parseString("akka.actor.debug
       out.expectComplete()
     }
 
+    "chain elements to currently emitting on upstream finish" in assertAllStagesStopped {
+      Source.single("hi")
+        .transform(() ⇒ new StatefulStage[String, String] {
+          override def initial = new State {
+            override def onPush(elem: String, ctx: Context[String]) =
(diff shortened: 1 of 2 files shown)


drewhk:
Yep, simple enough :)
LGTM

ktoso:
Nice, LGTM

2m:
Changed `case class EmittingState` to `private`. Will merge after validation.

Issue #17735, opened 2015-06-16
failed: ClusterClientSpec


patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly2/24/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/akka-nightly/2349/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node/1090/consoleFull

johanandren:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3637/consoleFull
(pretty sure this is the same)

patriknw:
ok, I'd rather have a new on, since this is closed on 2.4-M3: https://github.com/akka/akka/issues/18437

PR #17903, opened 2015-07-03
=clt #17735 Harden ClusterClientSpec

patriknw:
- number-of-contacts is by default 3, and in this test
  with 4 server nodes we shutdown all but one in the end
  and sometimes the client has all other except the remaining
  node in its list of contacts, so it will never get contact
  with the remaining node

### DIFF
--- a/akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
+++ b/akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
@@ -308,9 +308,13 @@ final class ClusterClient(settings: ClusterClientSettings) extends Actor with Ac
   }
 
   def sendGetContacts(): Unit = {
-    if (contacts.isEmpty) initialContactsSel foreach { _ ! GetContacts }
-    else if (contacts.size == 1) (initialContactsSel ++ contacts) foreach { _ ! GetContacts }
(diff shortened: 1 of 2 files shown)


akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -308,9 +308,13 @@ final class ClusterClient(settings: ClusterClientSettings) extends Actor with Ac
...
-    if (contacts.isEmpty) initialContactsSel foreach { _ ! GetContacts }
-    else if (contacts.size == 1) (initialContactsSel ++ contacts) foreach { _ ! GetContacts }
-    else contacts foreach { _ ! GetContacts }
+    val sendTo =
+      if (contacts.isEmpty) initialContactsSel
+      else if (contacts.size == 1) (initialContactsSel ++ contacts)
+      else contacts
+    if (log.isDebugEnabled)
+      log.debug(s"""Sending GetContacts to [${sendTo.mkString(",")}]""")
+    sendTo.foreach { _ ! GetContacts }
```
patriknw: no intentional changes in functionality here, except the added debug logging

ktoso:
Explanation makes sense to me, LGTM

patriknw:
Refs #17735

Issue #17781, opened 2015-06-19
A little question: Why routerConfig withFallback itself?


MQ-EL:
```scala
def actorOf(system: ActorSystemImpl, props: Props, supervisor: InternalActorRef, path: ActorPath,
              systemService: Boolean, deploy: Option[Deploy], lookupDeploy: Boolean, async: Boolean): InternalActorRef = {
    props.deploy.routerConfig match {

     ...

    case router ⇒
        val lookup = if (lookupDeploy) deployer.lookup(path) else None
        val fromProps = Iterator(props.deploy.copy(routerConfig = props.deploy.routerConfig   withFallback router))
        val d = fromProps ++ deploy.iterator ++ lookup.iterator reduce ((a, b) ⇒ b withFallback a)
        val p = props.withRouter(d.routerConfig)
```

The Line729 located In ActorRefProvider.scala:

```scala
val fromProps = Iterator(props.deploy.copy(routerConfig = props.deploy.routerConfig withFallback router))
```

The router is the same object as props.deploy.routerConfig IMO, did it has some deep implicit meaning?

The code in Line 730 doesn't needed with fallback of deploy, just fallback of deploy.routerConfig is enough? 
```scala
val d = fromProps ++ deploy.iterator ++ lookup.iterator reduce ((a, b) ⇒ b withFallback a)
```

Maybe refactor to this(?):
```scala
val c = props.deploy :: deploy.toList ::: lookup.toList reduce ((a,b) ⇒ b.routerConfig withFallback a.routerConfig)
```
Do i something wrong or miss some key knowledge?

MQ-EL:
Hi! Is this question too simple to answer?(may be i have a little Clumsy), could some body help me?

patriknw:
Thanks for asking, reporting. I have swapped out how the deploy fallbacks works at the moment.

Have you found anything that is not working as expected or is a just a matter of code cleanup?

MQ-EL:
1. For avoid unnesscery fallback of deploy object, do routerConfig fallback is enough IMO.
2. router is as the same object as props.deploy.routerConfig, so doesn't need fallback itself 

 It may be refactor as follows, 

```scala
val lookup = if(lookupDeploy) deployer.lookup(path) else None
val r = props.deploy :: deploy.toList ::: lookup.toList reduce ((a,b) ⇒ b.routerConfig withFallback a.routerConfig)
val p = props.withRouter(r)

[...]

PR #17905, opened 2015-07-03
=act #17781 Optimize the creation of router actor

MQ-EL:
1. Avoid the unnesscery fallback of deploy object
2. remove the fallback of router object with itself

refs: #17781

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/ActorRefProvider.scala
+++ b/akka-actor/src/main/scala/akka/actor/ActorRefProvider.scala
@@ -769,13 +769,12 @@ private[akka] class LocalActorRefProvider private[akka] (
 
       case router ⇒
         val lookup = if (lookupDeploy) deployer.lookup(path) else None
-        val fromProps = Iterator(props.deploy.copy(routerConfig = props.deploy.routerConfig withFallback router))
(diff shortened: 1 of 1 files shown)


ktoso:
OK TO TEST

ktoso:
OK TO TEST

MQ-EL:
I cannot find that the cause is because of my commit. Can Somebody verify it?

MQ-EL:
I think the test mode of akka has some problem, because the first commit failed, the third passed, but they are the same code.

ktoso:
LGTM, could you please squash into one commit and title it like so: `=act #17781 Optimize the creation of router actor`?

MQ-EL:
Ok, I create a new pr. Is it right?

ktoso:
Either works, thanks.
You can easily squash commits by doing:

```
git rebase --interactive HEAD^^^
// make the "to squash" commits "s"quash, and keep one "p"ick
push -f // to update this PR
```

ktoso:
Thanks, moving over to https://github.com/akka/akka/pull/18087 then!

MQ-EL:
Ha Ha, Thank you for sharing

PR #17906, opened 2015-07-03
=act #17805 Don't change self on restart (for validation)

patriknw:
- we hold on to the same ActorRef instance anyway in the new
  Actor instance


patriknw:
Refs #17805

Issue #17854, opened 2015-06-26
akka-http RC4 fails under load with 'Connection reset by peer' (regression)


mo-gr:
The following program works like a charm with akka-http RC3

    object Main extends App {
      implicit val actorSystem: ActorSystem = ActorSystem("main-actor-system")
      implicit val actorFlowMaterializer = ActorFlowMaterializer()
      val bindingFuture = Http(actorSystem).bindAndHandle((r: RequestContext) => r.complete("ok"), "0.0.0.0", 8080)
    }

When I hit it with apache bench like like so 

    ab -c 1 -n 1000 http://localhost:8080/

However, when I upgrade to RC4 and change the ActorFlowMaterializer to ActorMaterializer, I get `apr_socket_recv: Connection reset by peer (54)` from apache bench after something between 2 and 100 requests.

I have no idea, what could be the issue...

Me environment:
- java version "1.8.0_45"
- MacOS 10.10.3
- scala 2.11.6

The output from akka while failing is:

```
[info] Loading project definition from /Users/mgrauel/Coding/scompositor/project
[info] Set current project to scompositor (in build file:/Users/mgrauel/Coding/scompositor/)
[info] Running foo.Main 
[INFO] [06/26/2015 17:09:26.340] [main-actor-system-akka.actor.default-dispatcher-16] [akka://main-actor-system/system/IO-TCP/selectors/$a/3] Message [akka.io.SelectionHandler$ChannelReadable$] from Actor[akka://main-actor-system/deadLetters] to Actor[akka://main-actor-system/system/IO-TCP/selectors/$a/3#-658865784] was not delivered. [1] dead letters encountered. This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' and 'akka.log-dead-letters-during-shutdown'.
[INFO] [06/26/2015 17:09:26.483] [main-actor-system-akka.actor.default-dispatcher-14] [akka://main-actor-system/system/IO-TCP/selectors/$a/9] Message [akka.io.SelectionHandler$ChannelReadable$] from Actor[akka://main-actor-system/deadLetters] to Actor[akka://main-actor-system/system/IO-TCP/selectors/$a/9#-237500142] was not delivered. [2] dead letters encountered. This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' and 'akka.log-dead-letters-during-shutdown'.
[...]

PR #17916, opened 2015-07-06
=act: TCP actor should unwatch handler/commander after Close

drewhk:
/cc @sirthias @jrudolph 

This supposedly fixes the issue when Terminated for watched handler actor comes immediately after Close, while the TCP actor still has a pending write.

### DIFF
--- a/akka-actor/src/main/scala/akka/io/TcpConnection.scala
+++ b/akka-actor/src/main/scala/akka/io/TcpConnection.scala
@@ -38,6 +38,16 @@ private[io] abstract class TcpConnection(val tcp: TcpExt, val channel: SocketCha
   private[this] var readingSuspended = pullMode
   private[this] var interestedInResume: Option[ActorRef] = None
   var closedMessage: CloseInformation = _ // for ConnectionClosed message in postStop
+  private var watchedActor: ActorRef = context.system.deadLetters
+
+  def signDeathPact(actor: ActorRef): Unit = {
+    unsignDeathPact()
+    watchedActor = actor
+    context.watch(watchedActor)
+  }
+
+  def unsignDeathPact(): Unit =
+    if (watchedActor ne context.system.deadLetters) context.unwatch(watchedActor)
 
   def writePending = pendingWrite ne EmptyPendingWrite
 
@@ -268,6 +278,8 @@ private[io] abstract class TcpConnection(val tcp: TcpExt, val channel: SocketCha
       peerClosed = true
       context.become(peerSentEOF(info))
     case _ if writePending ⇒ // finish writing first
+      // Our registered actor is now free to terminate cleanly
+      unsignDeathPact()
       if (TraceLogging) log.debug("Got Close command but write is still pending.")
       context.become(closingWithPendingWrite(info, closeCommander, closedEvent))
     case ConfirmedClosed ⇒ // shutdown output and wait for confirmation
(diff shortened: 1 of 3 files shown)


akka-actor/src/main/scala/akka/io/TcpConnection.scala
```diff
@@ -257,32 +267,37 @@ private[io] abstract class TcpConnection(val tcp: TcpExt, val channel: SocketCha
...
-      if (TraceLogging) log.debug("Got ConfirmedClose command, sending FIN.")
-
-      // If peer closed first, the socket is now fully closed.
-      // Also, if shutdownOutput threw an exception we expect this to be an indication
-      // that the peer closed first or concurrently with this code running.
-      // also see http://bugs.sun.com/view_bug.do?bug_id=4516760
-      if (peerClosed || !safeShutdownOutput())
+                  closedEvent: ConnectionClosed): Unit = {
+    // Our registered actor is now free to terminate cleanly
+    if (!(closedEvent.isConfirmed || closedEvent.isPeerClosed)) unsignDeathPact()
```
jrudolph: Written negatively like this, I find it hard to understand when it applies. What happens if `handleClose` is called several times? It's a bit hard to trace.

Couldn't this more succinctly be added further down in the `case _ if writePending =>` body? That's the only case that should matter here, right? The other possibility would be just gracefully handling (= ignoring) the `Terminated` in the `closingWithPendingWrite` state.

jrudolph:
FWIW this PR alone doesn't fix the problem from #17854 (and @drewhk didn't say it does) because in the case described there the `closingWithPendingWrite` state cannot be reached (`Close` will only be used if `lastWriteAcked` => no write pending). The issue there is the wrong use of `ConfirmedClose` in the other case.

jrudolph:
(However, for the fix for #17854 to rely on `Close` instead of `ConfirmedClose` this fix is a prerequisite.)

akka-actor/src/main/scala/akka/io/TcpConnection.scala
drewhk: you are right, it is much cleaner to move it to the pattern match with writePending. Handling the Terminated is another option, I found it a little bit more explicit though to unwatch, but have no strong opinion here.

drewhk:
Yep, these two builds on each other.

akka-actor/src/main/scala/akka/io/TcpConnection.scala
```diff
@@ -257,32 +267,36 @@ private[io] abstract class TcpConnection(val tcp: TcpExt, val channel: SocketCha
...
-    case ConfirmedClosed ⇒ // shutdown output and wait for confirmation
-      if (TraceLogging) log.debug("Got ConfirmedClose command, sending FIN.")
-
-      // If peer closed first, the socket is now fully closed.
-      // Also, if shutdownOutput threw an exception we expect this to be an indication
-      // that the peer closed first or concurrently with this code running.
-      // also see http://bugs.sun.com/view_bug.do?bug_id=4516760
-      if (peerClosed || !safeShutdownOutput())
+                  closedEvent: ConnectionClosed): Unit = {
+    closedEvent match {
```
jrudolph: Spurious whitespace change.

drewhk: no, that is just the next line for L269 (it aligns with info:ConnectionInfo parameter)

jrudolph: Ah, I meant that the change wasn't necessary AFAICS. :)

drewhk: Ah, it is just the match on a new line. Because I removed the added "if" and now it is unnecessary. True, fixing

drewhk:
Refs: https://github.com/akka/akka/issues/17932

PR #17917, opened 2015-07-06
=str #17854: fullClose should not Abort

drewhk:


jrudolph:
How about splitting it up into two commits, one that fixes the `Abort` problem and one that fixes the `ConfirmedClose` / `Close` sequence problem in `fullClose` mode.

akka-stream/src/main/scala/akka/stream/impl/io/TcpConnectionStream.scala
```diff
@@ -154,17 +142,25 @@ private[akka] abstract class TcpStreamActor(val settings: ActorMaterializerSetti
...
-          connection ! Close
-          tryShutdown()
-        } else
-          connection ! ConfirmedClose
+
+        if (halfClose) {
+          if (tcpInputs.isClosed) {
+            // Reading has stopped, either because of cancel, or PeerClosed, just Close now
+            connection ! Close
+            tryShutdown()
```
jrudolph: These two lines now show up several times. Maybe it should be given a name now... :)

drewhk: well, three times :) I add a name.

drewhk:
I am not sure it worth an extra commit, but why not.

ktoso:
LGTM, validated on my machine a few hundred times.

drewhk:
extra Jenkins run won't hurt

PR #17921, opened 2015-07-06
-str - Improvements and renames in internal streams classes 

viktorklang:
Do not merge, intended for discussion at the moment.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
+++ b/akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
@@ -129,12 +129,11 @@ object ActorMaterializer {
    * INTERNAL API
    */
   private[akka] def downcast(materializer: Materializer): ActorMaterializer =
-    materializer match {
+    materializer match { //FIXME this method is going to cause trouble for other Materializer implementations
       case m: ActorMaterializer ⇒ m
       case _ ⇒ throw new IllegalArgumentException(s"required [${classOf[ActorMaterializer].getName}] " +
         s"but got [${materializer.getClass.getName}]")
     }
-
 }
 
 /**
--- a/akka-stream/src/main/scala/akka/stream/Attributes.scala
+++ b/akka-stream/src/main/scala/akka/stream/Attributes.scala
@@ -5,6 +5,7 @@ package akka.stream
 
 import akka.event.Logging
 
+import scala.annotation.tailrec
 import scala.collection.immutable
 import akka.stream.impl.Stages.StageModule
 import akka.japi.function
@@ -63,37 +64,36 @@ final case class Attributes private (attributeList: immutable.Seq[Attributes.Att
   /**
    * INTERNAL API
    */
-  private[akka] def nameLifted: Option[String] =
-    if (attributeList.isEmpty)
-      None
-    else {
-      val sb = new java.lang.StringBuilder
-      val iter = attributeList.iterator
-      while (iter.hasNext) {
-        iter.next() match {
-          case Name(name) ⇒
-            if (sb.length == 0) sb.append(name)
-            else sb.append("-").append(name)
-          case _ ⇒
+  private[akka] def nameLifted: Option[String] = Option(nameOrDefault(null))
+
+  /**
+   * INTERNAL API
+   */
+  private[akka] def nameOrDefault(default: String = "unknown-operation"): String = {
+    @tailrec def concatNames(i: Iterator[Attribute], first: String, buf: StringBuilder): String =
+      if (i.hasNext)
+        i.next() match {
+          case Name(n) ⇒
+            if (buf ne null) concatNames(i, null, buf.append('-').append(n))
+            else if (first ne null) {
+              val b = new StringBuilder(
+                (first.length() + n.length()) match {
+                  case x if x < 0                ⇒ throw new IllegalStateException("Names too long to concatenate")
+                  case y if y > Int.MaxValue / 2 ⇒ Int.MaxValue
+                  case z                         ⇒ Math.max(Integer.highestOneBit(z) * 2, 32)
+                })
+              concatNames(i, null, b.append(first).append('-').append(n))
+            } else concatNames(i, n, null)
+          case _ ⇒ concatNames(i, first, buf)
         }
-      }
-      if (sb.length == 0) None
-      else Some(sb.toString)
+      else if (buf eq null) first
+      else buf.toString
+
+    concatNames(attributeList.iterator, null, null) match {
+      case null ⇒ default
+      case some ⇒ some
     }
-
-  /**
-   * INTERNAL API
-   */
-  private[akka] def nameOrDefault(default: String = "unknown-operation"): String = nameLifted match {
-    case Some(name) ⇒ name
-    case _          ⇒ default
   }
-
-  /**
-   * INTERNAL API
-   */
-  private[akka] def nameOption: Option[String] =
-    attributeList.collectFirst { case Name(name) ⇒ name }
 
   /**
    * INTERNAL API
--- a/akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
@@ -68,7 +68,7 @@ private[akka] case class ActorMaterializerImpl(
   override def materialize[Mat](runnableGraph: Graph[ClosedShape, Mat]): Mat = {
     if (haveShutDown.get())
       throw new IllegalStateException("Attempted to call materialize() after the ActorMaterializer has been shut down.")
-    if (StreamLayout.Debug) runnableGraph.module.validate()
+    if (StreamLayout.Debug) StreamLayout.validate(runnableGraph.module)
 
     val session = new MaterializerSession(runnableGraph.module) {
       private val flowName = createFlowName()
@@ -299,8 +299,6 @@ private[akka] object ActorProcessorFactory {
     val settings = materializer.effectiveSettings(att)
     def interp(s: Stage[_, _]): (Props, Unit) = (ActorInterpreter.props(settings, List(s), materializer, att), ())
     op match {
-      case Identity(_)                ⇒ throw new AssertionError("Identity cannot end up in ActorProcessorFactory")
-      case Fused(ops, _)              ⇒ (ActorInterpreter.props(settings, ops, materializer, att), ())
       case Map(f, _)                  ⇒ interp(fusing.Map(f, settings.supervisionDecider))
       case Filter(p, _)               ⇒ interp(fusing.Filter(p, settings.supervisionDecider))
       case Drop(n, _)                 ⇒ interp(fusing.Drop(n))
@@ -328,6 +326,7 @@ private[akka] object ActorProcessorFactory {
         val s_m = mkStageAndMat()
         (ActorInterpreter.props(settings, List(s_m._1), materializer, att), s_m._2)
       case DirectProcessor(p, m) ⇒ throw new AssertionError("DirectProcessor cannot end up in ActorProcessorFactory")
+      case Identity(_)           ⇒ throw new AssertionError("Identity cannot end up in ActorProcessorFactory")
     }
   }
 
--- a/akka-stream/src/main/scala/akka/stream/impl/Modules.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/Modules.scala
@@ -3,17 +3,14 @@
  */
 package akka.stream.impl
 
-import java.io.{ InputStream, File }
 import java.util.concurrent.atomic.AtomicBoolean
 import akka.actor.{ ActorRef, Cancellable, PoisonPill, Props }
-import akka.stream.ActorAttributes.Dispatcher
 import akka.stream.impl.StreamLayout.Module
 import akka.stream._
-import akka.util.ByteString
 import org.reactivestreams._
 import scala.annotation.unchecked.uncheckedVariance
 import scala.concurrent.duration.FiniteDuration
-import scala.concurrent.{ Future, Promise }
+import scala.concurrent.{ Promise }
 import scala.util.{ Failure, Success }
 
 /**
@@ -34,13 +31,13 @@ private[akka] abstract class SourceModule[+Out, +Mat](val shape: SourceShape[Out
 
   override def subModules: Set[Module] = Set.empty
 
-  def amendShape(attr: Attributes): SourceShape[Out] =
-    attr.nameOption match {
-      case None ⇒ shape
-      case s: Some[String] if s == attributes.nameOption ⇒ shape
-      case Some(name) ⇒ shape.copy(outlet = Outlet(name + ".out"))
-    }
+  protected def amendShape(attr: Attributes): SourceShape[Out] = {
+    val thisN = attributes.nameOrDefault(null)
+    val thatN = attr.nameOrDefault(null)
 
+    if ((thatN eq null) || thisN == thatN) shape
+    else shape.copy(outlet = Outlet(thatN + ".out"))
+  }
 }
 
 /**
--- a/akka-stream/src/main/scala/akka/stream/impl/Stages.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/Stages.scala
@@ -5,13 +5,12 @@ package akka.stream.impl
 
 import akka.event.{ LoggingAdapter, Logging }
 import akka.stream.impl.SplitDecision.SplitDecision
-import akka.stream.{ OverflowStrategy, TimerTransformer }
-import akka.stream.Attributes
+import akka.stream.impl.StreamLayout._
+import akka.stream.{ OverflowStrategy, TimerTransformer, Attributes }
 import akka.stream.Attributes._
 import akka.stream.stage.Stage
 import org.reactivestreams.Processor
-import StreamLayout._
-
+import akka.event.Logging.simpleName
 import scala.collection.immutable
 import scala.concurrent.Future
 
@@ -117,16 +116,6 @@ private[stream] object Stages {
   }
 
   final case class Identity(attributes: Attributes = Attributes.name("identity")) extends StageModule {
-    def withAttributes(attributes: Attributes) = copy(attributes = attributes)
-    override protected def newInstance: StageModule = this.copy()
-  }
-
-  object Fused {
-    def apply(ops: immutable.Seq[Stage[_, _]]): Fused =
-      Fused(ops, name(ops.iterator.map(x ⇒ Logging.simpleName(x).toLowerCase).mkString("+")))
-  }
-
-  final case class Fused(ops: immutable.Seq[Stage[_, _]], attributes: Attributes = fused) extends StageModule {
     def withAttributes(attributes: Attributes) = copy(attributes = attributes)
     override protected def newInstance: StageModule = this.copy()
   }
--- a/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
@@ -23,6 +23,76 @@ private[akka] object StreamLayout {
   // compile-time constant
   final val Debug = false
 
+  final def validate(m: Module, level: Int = 0, doPrint: Boolean = false, idMap: mutable.Map[AnyRef, Int] = mutable.Map.empty): Unit = {
+    val ids = Iterator from 1
+    def id(obj: AnyRef) = idMap get obj match {
+      case Some(x) ⇒ x
+      case None ⇒
+        val x = ids.next()
+        idMap(obj) = x
+        x
+    }
+    def in(i: InPort) = s"${i.toString}@${id(i)}"
+    def out(o: OutPort) = s"${o.toString}@${id(o)}"
+    def ins(i: Iterable[InPort]) = i.map(in).mkString("In[", ",", "]")
+    def outs(o: Iterable[OutPort]) = o.map(out).mkString("Out[", ",", "]")
+    def pair(p: (OutPort, InPort)) = s"${in(p._2)}->${out(p._1)}"
+    def pairs(p: Iterable[(OutPort, InPort)]) = p.map(pair).mkString("[", ",", "]")
+
+    import m._
+
+    val inset: Set[InPort] = shape.inlets.toSet
+    val outset: Set[OutPort] = shape.outlets.toSet
+    var problems: List[String] = Nil
+
+    if (inset.size != shape.inlets.size) problems ::= "shape has duplicate inlets: " + ins(shape.inlets)
+    if (inset != inPorts) problems ::= s"shape has extra ${ins(inset -- inPorts)}, module has extra ${ins(inPorts -- inset)}"
+    if (inset.intersect(upstreams.keySet).nonEmpty) problems ::= s"found connected inlets ${inset.intersect(upstreams.keySet)}"
+    if (outset.size != shape.outlets.size) problems ::= "shape has duplicate outlets: " + outs(shape.outlets)
+    if (outset != outPorts) problems ::= s"shape has extra ${outs(outset -- outPorts)}, module has extra ${outs(outPorts -- outset)}"
+    if (outset.intersect(downstreams.keySet).nonEmpty) problems ::= s"found connected outlets ${outset.intersect(downstreams.keySet)}"
+    val ups = upstreams.toSet
+    val ups2 = ups.map(_.swap)
+    val downs = downstreams.toSet
+    val inter = ups2.intersect(downs)
+    if (downs != ups2) problems ::= s"inconsistent maps: ups ${pairs(ups2 -- inter)} downs ${pairs(downs -- inter)}"
+    val (allIn, dupIn, allOut, dupOut) =
+      subModules.foldLeft((Set.empty[InPort], Set.empty[InPort], Set.empty[OutPort], Set.empty[OutPort])) {
+        case ((ai, di, ao, doo), m) ⇒ (ai ++ m.inPorts, di ++ ai.intersect(m.inPorts), ao ++ m.outPorts, doo ++ ao.intersect(m.outPorts))
+      }
+    if (dupIn.nonEmpty) problems ::= s"duplicate ports in submodules ${ins(dupIn)}"
+    if (dupOut.nonEmpty) problems ::= s"duplicate ports in submodules ${outs(dupOut)}"
+    if (!isSealed && (inset -- allIn).nonEmpty) problems ::= s"foreign inlets ${ins(inset -- allIn)}"
+    if (!isSealed && (outset -- allOut).nonEmpty) problems ::= s"foreign outlets ${outs(outset -- allOut)}"
+    val unIn = allIn -- inset -- upstreams.keySet
+    if (unIn.nonEmpty && !isCopied) problems ::= s"unconnected inlets ${ins(unIn)}"
+    val unOut = allOut -- outset -- downstreams.keySet
+    if (unOut.nonEmpty && !isCopied) problems ::= s"unconnected outlets ${outs(unOut)}"
+
+    def atomics(n: MaterializedValueNode): Set[Module] =
+      n match {
+        case Ignore                  ⇒ Set.empty
+        case Transform(f, dep)       ⇒ atomics(dep)
+        case Atomic(m)               ⇒ Set(m)
+        case Combine(f, left, right) ⇒ atomics(left) ++ atomics(right)
+      }
+    val atomic = atomics(materializedValueComputation)
+    if ((atomic -- subModules - m).nonEmpty) problems ::= s"computation refers to non-existent modules [${atomic -- subModules - m mkString ","}]"
+
+    val print = doPrint || problems.nonEmpty
+
+    if (print) {
+      val indent = " " * (level * 2)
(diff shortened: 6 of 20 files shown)


akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
```diff
@@ -129,12 +129,11 @@ object ActorMaterializer {
    * INTERNAL API
    */
   private[akka] def downcast(materializer: Materializer): ActorMaterializer =
-    materializer match {
+    materializer match { //FIXME this method is going to cause trouble for other Materializer implementations
```
ktoso: in what sense? We use it on places where we need actors or an actor system, these ops won't work without one (can't remember all of them but `log()` *without* providing a `LoggingAdapter` is one usage, with providing one explicitly its ok)

viktorklang: The problem is that you can only provide a single Materializer, and even if that were carrying an ActorMaterializer this piece of logic would fail. Needs reengineering IMO.

akka-stream/src/main/scala/akka/stream/Attributes.scala
```diff
@@ -63,37 +64,36 @@ final case class Attributes private (attributeList: immutable.Seq[Attributes.Att
...
-  }
+  private[akka] def nameOrDefault(default: String = "unknown-operation"): String = {
+    @tailrec def concatNames(i: Iterator[Attribute], first: String, buf: StringBuilder): String =
+      if (i.hasNext)
+        i.next() match {
+          case Name(n) ⇒
+            if (buf ne null) concatNames(i, null, buf.append('-').append(n))
+            else if (first ne null) {
+              val b = new StringBuilder(
+                (first.length() + n.length()) match {
```
viktorklang: This is arguably a bit over-engineered.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -283,14 +291,21 @@ private[akka] object StreamLayout {
...
-        | ${downstreams.map { case (in, out) ⇒ s"   $in -> $out" }.mkString("\n")}
+        | ${downstreams.iterator.map { case (in, out) ⇒ s"   $in -> $out" }.mkString("\n")}
         | Upstreams:
-        | ${upstreams.map { case (out, in) ⇒ s"   $out -> $in" }.mkString("\n")}
+        | ${upstreams.iterator.map { case (out, in) ⇒ s"   $out -> $in" }.mkString("\n")}
       """.stripMargin
 
+    override def describe: String =
+      attributes.nameOrDefault(null) match {
+        case null ⇒ "(" + subModules.iterator.map(_.describe).mkString(" ~> ") + ")"
```
viktorklang: @drewhk Sadly, since subModules is a Set, there's no way to get stable ordering of operations. Any ideas how to achieve that?

~~~scala
scala> res3 via res5
res6: akka.stream.scaladsl.Source[String,Unit] = fold ~> futureSource ~> mapAsyncUnordered ~> filter ~> map

scala> res3 via res5
res7: akka.stream.scaladsl.Source[String,Unit] = filter ~> fold ~> futureSource ~> mapAsyncUnordered ~> map

scala> res3 via res5
res8: akka.stream.scaladsl.Source[String,Unit] = filter ~> fold ~> futureSource ~> mapAsyncUnordered ~> map

scala> res3 via res5
res9: akka.stream.scaladsl.Source[String,Unit] = fold ~> futureSource ~> filter ~> mapAsyncUnordered ~> map
~~~

rkuhn: This is definitely wrong, printing the structure will require proper interpretation of all the wiring inside of it.

viktorklang: Oh, it's absolutely wrong. (hence the comment). Sadly the `proper interpretation of all the wiring inside of it` seems trickier than I had hoped for.

akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
drewhk: I am not sure I follow. This is just an internal convenience method where we already needed to cast to ActorMaterializer, How would this fail with an ActorMaterializer?

akka-stream/src/main/scala/akka/stream/Attributes.scala
drewhk: well, if it makes it faster, then I think it is fine.

akka-stream/src/main/scala/akka/stream/impl/Modules.scala
```diff
@@ -34,13 +31,13 @@ private[akka] abstract class SourceModule[+Out, +Mat](val shape: SourceShape[Out
 
   override def subModules: Set[Module] = Set.empty
 
-  def amendShape(attr: Attributes): SourceShape[Out] =
-    attr.nameOption match {
-      case None ⇒ shape
-      case s: Some[String] if s == attributes.nameOption ⇒ shape
-      case Some(name) ⇒ shape.copy(outlet = Outlet(name + ".out"))
-    }
+  protected def amendShape(attr: Attributes): SourceShape[Out] = {
```
drewhk: looks good

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
```diff
@@ -126,6 +125,7 @@ private[stream] object Stages {
       Fused(ops, name(ops.iterator.map(x ⇒ Logging.simpleName(x).toLowerCase).mkString("+")))
   }
 
+  //FIMXE this is not used.
```
drewhk: we have ticket for fusing, no need for FIXME

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: So, how I did this before: there was a LinearComposite which was a CompositeModule variant that was always linear, only changing to CompositeModule if linearity was violated (i.e. general graph). The toString on LinearComposite then was easy. I sadly don't find the branch, I hope I didn't delete it. I will need a LinearComposite probably anyway to help fusing.

This won't help with nonlinear modules of course, that belongs to the nasty world of graph layout algorithms. I had some ideas how to make something useful for debug that looks reasonable for most of the cases without going too far with graph layout algos. Instead we will provide a DOT printer that you can use with Graphviz or some other layout tool.

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
viktorklang: This seems to be a common situation for me, and I fear that it is prohibitive to require searching through the issue database (given its current size) whenever one wants to understand why a piece of code is as it is. Here's my empirical stance on the matter: either the code is there and has a FIXME, or the code is pulled out until the person working on the Issue adds it again. Code that doesn't do anything is a major liability.

akka-stream/src/main/scala/akka/stream/Attributes.scala
viktorklang: The main improvement is not having to allocate any new builder nor string for when there's only 1 name.
And for the case where there are only 2 names, there will only be 1 allocation of the builders' internal buffer.
The length calculation, to be correct, needs to deal with size overflow etc—hence the match-case. :/

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
viktorklang: The LinearComposite sounds exactly like what we ought to have. Any hints as to how/where to add it? (where are the dragons)

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
drewhk: You can just remove this class instead then. It will be very obvious when taking the appropriate ticket to add the required stage. It might look completely different anyway.

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -62,7 +44,7 @@ final class Source[+Out, +Mat](private[stream] override val module: Module)
       new Source(
         module
           .growConnect(flowCopy, shape.outlet, flowCopy.shape.inlets.head, combine)
-          .replaceShape(SourceShape(flowCopy.shape.outlets.head)))
+          .replaceShape(SourceShape(flowCopy.shape.outlets.head))) // FIXME why is not .wrap() needed here?
```
viktorklang: @drewhk @ktoso Could you explain this to me?

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -243,22 +238,36 @@ object Source extends SourceApply {
...
    * A `Source` with no elements, i.e. an empty stream that is completed immediately for every connected `Sink`.
    */
   def empty[T]: Source[T, Unit] = _empty
-  private[this] val _empty: Source[Nothing, Unit] = apply(EmptyPublisher).withAttributes(DefaultAttributes.emptySource)
+  private[this] val _empty: Source[Nothing, Unit] =
+    new Source(
+      new PublisherSource[Nothing](
+        EmptyPublisher,
+        DefaultAttributes.emptySource,
+        shape("EmptySource")))
```
viktorklang: There's quite a bit of repetition here when creating these sources. Is there any reason why the PublisherSource can't create its own SourceShape given that it is handed the name anyway?

viktorklang: @drewhk thoughts?

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
viktorklang: Deal!

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: One of the dragons have been eliminated (which forced me to not merge it that time) and that was copying of imported graphs. Now we have a marker CopiedModule instead, and this helped to simplify the wiring code. 

The idea roughly:
 - create a LinearComposite case class that extends Module and which only have at most one external (unwired) output and at most one external input port.
 - override growConnect to handle the case when the Module is wired linearly
 - once non-linear wiring is detected, convert the thing to the existing CompositeModule (since no longer possible to represent the thing as a chain, it is now a graph)
 - wrap() should return a LinearComposite, so we always assume that a graph is linear, until we detect its not.

At least this is how I did it. External users (the user DSLs) should not be aware of the existence of this internal optimization. The MaterializerSession can optionally make use of this extra knowledge to make materialization faster, but this is a nice to have.

akka-stream/src/main/scala/akka/stream/Attributes.scala
drewhk: I thought so. Since materialization uses this, and materialization is currently the perf bottleneck, I think this makes sense here.

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
drewhk: because you just grow the module. Wrapping would introduce a new hierarchy level. 

A little bit of history here:
First, I had a different type for "ModuleInProgress" which signified that it was not a final module, but a thing being built. While this was self-documenting, it turned out to be a huge duplication most of the time, so we dropped it. We just treat the current module as the "builder" implicitly now. Since Source is both a module, or a builder, this means that its submodule is being grown, or "finalized" if imported in a graph.

An example:
 - empty module: []
 - grow with map: [map]
 - grow with viaMat(someFlow):  [map -> someFlow]
 - wrap(): [[map -> someFlow]]
 - grow with filter: [[map -> someFlow] -> filter]

Another example:
 - one source: [single -> map ->]
 - one sink: [ -> filter -> foreach]
 - wiring them together: [single -> map -> filter -> foreach]

I.e. it tries to work on the same hierarchy level, growing the top-level module. If you use .named, it will wrap:
 - one source [single -> map ->]
 - adding a name: [named[single -> map ->] ->]
 - growing: [named[single -> map ->] -> filter ->]
 - one sink: [-> mapAsync -> foreach]
 - adding sink: [named[single -> map ->] -> filter -> mapAsync -> foreach]

akka-stream/src/main/scala/akka/stream/io/SynchronousFileSource.scala
```diff
@@ -25,8 +25,7 @@ object SynchronousFileSource {
    * It materializes a [[Future]] containing the number of bytes read from the source file upon completion.
    */
   def apply(f: File, chunkSize: Int = DefaultChunkSize): Source[ByteString, Future[Long]] =
-    new Source(new SynchronousFileSource(f, chunkSize, DefaultAttributes, Source.shape("SynchronousFileSource")))
-      .named(DefaultAttributes.nameOption.get)
+    new Source(new SynchronousFileSource(f, chunkSize, DefaultAttributes, Source.shape("SynchronousFileSource")).wrap()) // TO DISCUSS: I had to add wrap() here to make the name available
```
viktorklang: @drewhk But why do I need to call wrap here if I removed the `.named(DefaultAttributes.nameOption.get)` section?

I fear that remembering where and when `wrap()` needs to be called will lead to very subtle bugs.

drewhk: What was the exact problem? Anyway, named() called withAttributes() on the Source which called wrap().

> I fear that remembering where and when wrap() needs to be called will lead to very subtle bugs.

Yes. Calling wrap() does not change the semantics of the running stream though, it changes the module boundaries, i.e. modularity. The Layout cannot have any magical means to figure out this for you. You, as the designer of the user level DSL have the responsibility to choose the semantics that you want, and implement it properly. There are many possible semantics and the Layout cannot choose any of them. The Layout is just an immutable ADT for building hierarchical module networks, and it has only a few basic operations: grow, connect and wrap.

viktorklang: Without the wrap, the test in SynchronousFileSourceSpec that checks the name of the source fails.

>Yes. Calling wrap() does not change the semantics of the running stream though, it changes the module boundaries, i.e. modularity. The Layout cannot have any magical means to figure out this for you. You, as the designer of the user level DSL have the responsibility to choose the semantics that you want, and implement it properly. There are many possible semantics and the Layout cannot choose any of them. The Layout is just an immutable ADT for building hierarchical module networks, and it has only a few basic operations: grow, connect and wrap.

Perhaps the problem I have is partly with the name(s). What is the rationale behind the current names? (they lack documentation AFAICT)

drewhk: > What is the rationale behind the current names

Nothing, I chose them, and noone have complained until this point :) You can freely change them if you like.

viktorklang: Lol! Ok :)

Would you mind spending a could of minutes explaining to an idiot like myself what the desired semantics of those methods are and then I can see if I can improve on the names :)

drewhk: Yes. So a module is either atomic or composite. It is the atomic modules that are turned into Publishers/Subscribers, the composite ones are just an organizational abstraction. Composite modules contain submodules, possibly recursively. All modules treat all unwired input and output ports as external ports.

 - grow(other: Module): takes the current module and adds the submodules of the other module to it (or, looking another way, it makes a new module which contains the submodules of both) without changing existing wiring between the submodules (i.e. if submodules were wired in one of the modules, they are still wired in the new module). This does not add a hierarchy level. Grow also can define how to calculate the materialized value of the resulting module from the materialized values of the original modules.
 - connect(Outport, Inport): takes a module, and "destroys" one external input and one external output port, wiring them together. So the number of external ports will be less by 2. This does not add a hierarchy level. Materialized values are not changed.
 - wrap(): Creates a new module that contains the original module as a submodule. This adds a hierarchy level. This also means that grow will now affect this new toplevel module, the structure of the original module is unchanged (since it is now a submodule). Materialized values are not changed.

viktorklang: so:
`grow` == `compose`
`connect` == `wire` | `fuse` | `join` ?
`wrap` == `nest`

The reason I don't like `grow` is that it is too generic, it doesn't say if it grows veritically or horizontally etc. 
`connect` could work, and there are some options that are more `physical` than connect.
`wrap` has been hard, for me, to, no pun intended, wrap my head around. The wrapping is an effect, but it doesn't say anything about the purpose. `Nest` is perhaps clearer about that?

viktorklang: Thanks for explaining, Endre, it makes much more sense now.

drewhk: well, grow an be also "union".
connect = wire and wrap = nest looks good to me.

viktorklang: "unionize" ;-)

viktorklang: Alright,

I renamed `grow` to `compose`, `wrap` to `nest` and `connect` to `wire` and renamed `growConnect` to `fuse`.
Will proceed by writing some scaladoc for them.

ktoso: Those names sound good :+1:

akka-stream/src/main/scala/akka/stream/impl/io/IOSources.scala
```diff
@@ -20,7 +20,7 @@ import scala.concurrent.{ Future, Promise }
 private[akka] final class SynchronousFileSource(f: File, chunkSize: Int, val attributes: Attributes, shape: SourceShape[ByteString])
   extends SourceModule[ByteString, Future[Long]](shape) {
   override def create(context: MaterializationContext) = {
-    val mat = ActorMaterializer.downcast(context.materializer)
+    val mat = ActorMaterializer.downcast(context.materializer) //FIXME why does this need to rely on a concrete Materializer impl?
```
ktoso: for that line below:

```
     val ref = mat.actorOf(context, props.withDispatcher(dispatcher))
```

It would be possible to make it actor-less for the synchronous file source, I don't think we need to do that for 1.0 in any case, in general we could I guess if it's worth it, yet I'd rather not implement actor-less for asynchronous-io once we get to it though.

viktorklang: I don't think we need to do it for 1.0, but I'll leave the FIXME so that we can implement it with an AsyncStage instead later.

ktoso: Right, `AsyncStage` should be enough. I think this impl happened before we had async stage... not sure. Please leave a readme (or maybe better, ticket)

drewhk:
This is currently not working with the names right? Can you just keep the rest of the changes so I can merge them?

akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
viktorklang: If someone would try to materialize a SynchronousFileSource for instance, with a non ActorMaterializer, it would simply fail. And there'd be no way for them to write their Materializer so that it wouldn't fail.

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
```diff
@@ -299,8 +299,6 @@ private[akka] object ActorProcessorFactory {
     val settings = materializer.effectiveSettings(att)
     def interp(s: Stage[_, _]): (Props, Unit) = (ActorInterpreter.props(settings, List(s), materializer, att), ())
     op match {
-      case Identity(_)                ⇒ throw new AssertionError("Identity cannot end up in ActorProcessorFactory")
```
viktorklang: Checking first, for every materialization, a stage that shouldn't be there at all seems a bit wasteful :)

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
```diff
@@ -299,8 +299,6 @@ private[akka] object ActorProcessorFactory {
...
-      case Fused(ops, _)              ⇒ (ActorInterpreter.props(settings, ops, materializer, att), ())
```
viktorklang: Removed Fused

akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
drewhk: Yes exactly. But how is this method causing that? SynchronousFileSource is Actor based, so only Akka actor based materializers have a chance anyway, because it needs Akka features (hence the cast, which this method only makes more convenient and gives a uniform failure message). Custom materializers that want to support Akka actor based stages need to extend ActorMaterializer (which is not the same as ActorMaterializerImpl) which is public and abstract.

akka-stream/src/main/scala/akka/stream/Attributes.scala
ktoso: Seems I forgot to comment here on my previous review round: I'm OK with this snippet, esp. since in materialization code path.

akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
viktorklang: From ActorMaterializer:

~~~scala
  /**
   * INTERNAL API: this might become public later
   */
  private[akka] def actorOf(context: MaterializationContext, props: Props): ActorRef

  /**
   * INTERNAL API
   */
  private[akka] def system: ActorSystem

  /** INTERNAL API */
  private[akka] def supervisor: ActorRef
~~~

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
ktoso: A man can dream: Wouldn't it be great if match would reorder cases depending on runtime information? 
// that's not an optimisation JITs do AFAIR, right?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -89,10 +186,10 @@ private[akka] object StreamLayout {
         attributes)
     }
 
-    def grow(that: Module): Module = grow(that, Keep.left)
+    def compose(that: Module): Module = compose(that, Keep.left)
 
-    def grow[A, B, C](that: Module, f: (A, B) ⇒ C): Module = {
-      if (Debug) validate()
+    def compose[A, B, C](that: Module, f: (A, B) ⇒ C): Module = {
```
ktoso: minor (i.e. not a blocker for merging this): other methods got scaladoc, would be nice if those got it too.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -52,14 +123,40 @@ private[akka] object StreamLayout {
...
+     * transforming the materialized values of `this` and `that` using the
+     * provided function `f`
+     * @param that a Module to fuse with
+     * @param from the data source to wire
+     * @param to the data sink to wire
+     * @param f the function to apply to the materialized values
+     * @return a Module representing the fusion of `this` and `that`
+     */
+    final def fuse[A, B, C](that: Module, from: OutPort, to: InPort, f: (A, B) ⇒ C): Module =
+      this.compose(that, f).wire(from, to)
```
ktoso: I like the new names here, make much sense

akka-stream/src/main/scala/akka/stream/impl/io/IOSources.scala
```diff
@@ -20,6 +20,7 @@ import scala.concurrent.{ Future, Promise }
 private[akka] final class SynchronousFileSource(f: File, chunkSize: Int, val attributes: Attributes, shape: SourceShape[ByteString])
   extends SourceModule[ByteString, Future[Long]](shape) {
   override def create(context: MaterializationContext) = {
+    // FIXME rewrite to be based on AsyncStage rather than dangerous downcasts
```
ktoso: +1

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -153,14 +135,18 @@ final class Source[+Out, +Mat](private[stream] override val module: Module)
    */
   def ++[Out2 >: Out, M](second: Graph[SourceShape[Out2], M]): Source[Out2, (Mat, M)] = concat(second)
 
+  /**
+   * Wraps the current Source and returns a Source with the given Attributes
```
ktoso: Minor: "Nests" would be more consistent wording with current API.

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -243,22 +236,36 @@ object Source extends SourceApply {
...
-    apply(new immutable.Iterable[T] {
-      override val iterator: Iterator[T] = Iterator.continually(element)
-      override def toString: String = "Iterable.continually(" + element + ")"
-    }).withAttributes(DefaultAttributes.repeat)
+    new Source(
+      new PublisherSource(
+        SingleElementPublisher(
+          new immutable.Iterable[T] {
+            override val iterator: Iterator[T] = Iterator.continually(element)
+            override def toString: String = "repeat(" + element + ")"
```
ktoso: good catch on the toString here.

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
ktoso: I may have missed a discussion, we don't need `Fused` now? Is it because LinearModules coming?

ktoso:
Found some minor nitpicks, other than that LGTM
(I'd like to understand why Fuse is gone however)

akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
drewhk: Yes, currently is saying "this might become public later". There is no full support for custom materializers yet, and this cast is just only one of the reasons.

But you say that this cast should be removed, which is basically excluding all Akka integration patterns, since our materializer will no longer to be able to access Akka provided features easily. The way to fix this is not to remove this method, but to eventually make ActorMaterializer properly supported and extensible. I.e. if you want to put a FIXME, then put it there instead.

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
drewhk: currently it is unused. It is just a wrapper over a Seq of stages. It probably needs some changes anyway in the future (attributes for stages for example).

ktoso: I see. Now I remember that Viktor's `optimize` directly looked at Map nodes etc. Ok cool.

akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
viktorklang: This has the signature of a responsibility problem, if something on the outside needs to know whether something is an ActorMaterializer or not and -fail- if it isn't, then I'd argue that it's not being honest about its dependencies.
Materialization should be contained in the Materializer, perhaps we need to take a step back and see if a restructuing would eliminate the need for downcasting.

drewhk: I don't follow what is the problem here, but whatever. Keep this FIXME here then.

viktorklang: Agreed. We'll fix it later for sure.

viktorklang:
@ktoso Fused is removed because it isn't used. There's a ticket to add fusion, and that ticket will have to be responsible for the implementation, so I'm removing the dead code.

This is ready to merge afaict, :)

viktorklang:
Well, besides the commit message, but let me know if everything in here's fine first

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
viktorklang: Thanks! :)

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
viktorklang: Will fix

ktoso: Thanks @viktorklang :) Once fixed give it some OK commit message and we'll merge.
The plan is to release 1.0 tomorrow.

viktorklang: ok. will hopefully have time during this evening

-- 
Cheers,
√

viktorklang:
@ktoso Docs added and proper commit message created.

2m:
LGTM, nice allocation optimizations.

ktoso:
Muchas gracias Señor!

Issue #17842, opened 2015-06-26
Move BackoffSupervisor from akka persistence package


Horusiath:
Right now a [BackoffSupervisor](https://github.com/akka/akka/blob/master/akka-persistence/src/main/scala/akka/persistence/BackoffSupervisor.scala) is implemented inside akka.persistence package. Since it doesn't depend on anything persistence-specific, the idea is to move it into akka core under akka.pattern module.

patriknw:
thanks for the ticket

patriknw:
If we do this we should also remember to add a description of it in the actor Fault Tolerance documentation section.

ktoso:
I think it's a nice idea, back-off is a very nice universally useful tool to have :)

patriknw:
ok, let's do it

michaelom:
Exponential backoff during restart is a very useful feature (particularly in a CQRS setup).
+1

ktoso:
// cc @Horusiath, we moved this class to `akka.pattern`

PR #17925, opened 2015-07-07
!per +act #17842 move BackoffSupervisor to akka.pattern

ktoso:
Resolves https://github.com/akka/akka/issues/17842

### DIFF
--- a/akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
+++ b/akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
@@ -1,12 +1,11 @@
 /**
  * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
  */
-package akka.persistence
+package akka.pattern
 
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.forkjoin.ThreadLocalRandom
 import akka.actor.Actor
-import akka.actor.ActorLogging
 import akka.actor.ActorRef
 import akka.actor.DeadLetterSuppression
 import akka.actor.Props
@@ -17,6 +16,8 @@ import scala.concurrent.duration.Duration
 object BackoffSupervisor {
 
   /**
+   * Props for creating an [[BackoffSupervisor]] actor.
+   *
    * @param childProps the [[akka.actor.Props]] of the child actor that
    *   will be started and supervised
    * @param childName name of the child actor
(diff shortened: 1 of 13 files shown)


akka-docs/rst/general/supervision.rst
```diff
@@ -199,6 +199,25 @@ external resource, which may also be one of its own children. If a third party
...
+This pattern is useful when the started actor fails because some external resource is not available,
+and we need to give it some time to start-up again. One of the prime examples when this is useful is
+when a :ref:`PersistentActor <persistence-scala>` fails with an persistence failure - which indicates that
+the database may be down or overloaded, in such situations it makes most sense to give it a little bit of time
+to recover before the peristent actor is restarted.
+
+The following snippet shows how to create a backoff supervisor which will start the given echo actor
+in increasing intervals of 3, 6, 12, 24 and finally 30 seconds:
+
+.. includecode:: ../scala/code/docs/pattern/BackoffSupervisorDocSpec.scala#backoff
```
ktoso: So this doc is a "general" doc, so lang agnostic, however showing how to use this seemed useful. Do we need to split up into java/supervision and scala/supervision or create a new page for that pattern? 

cc @2m @drewhk @viktorklang

drewhk: I think you can link two code examples, i.e. one for Java and Scala.

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
```diff
@@ -24,8 +25,8 @@ object BackoffSupervisor {
    *   started again, if it is terminated
    * @param maxBackoff the exponential back-off is capped to this duration
    * @param randomFactor after calculation of the exponential back-off an additional
-   *   random delay based on this factor is added, e.g. 0.2 adds up to 20%
-   *   delay
+   *   random delay based on this factor is added, e.g. `0.2` adds up to `20%` delay.
```
drewhk: it is highly encouraged to add randomization. This should be mentioned in the docs.

drewhk:
LGTM otherwise

akka-docs/rst/general/supervision.rst
ktoso: Yeah, sounds good

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
ktoso: done, added encouragement in docs

Issue #15569, opened 2014-07-20
Document testing persistent actors (incompatibility with TestActorRef)


ktoso:
As seen here: https://github.com/akka/akka/issues/15293
And here: https://groups.google.com/forum/#!topic/akka-user/FNOGnAFej9Y

People are confused when they try to use `TestActorRef` with akka persistence and it does not work. We should document that they are not compatible.

Future thinking: we should do a section in the docs on testing persistent actors in general?

PR #17926, opened 2015-07-07
=doc #15569 document TestActorRef being bad for persistence

ktoso:
Initial resolution of https://github.com/akka/akka/issues/15569 such that people don't try to combine `TestActorRef` with Persistence, it just won't work - persistence needs async things to happen for it to work.

In the long run, this could be reworded once we have a test-kit for persistence and point to it instead.


drewhk:
LGTM, we should probably backport this to 2.3

ktoso:
Yeah, agreed.

Issue #17923, opened 2015-07-07
Streams: javadsl.Source.elements(elems: T*) has scala Seq as attribute


agolubev:

ktoso:
Good point, this should be fixed.

Fun fact, it's not a trivial "add `@varargs` and it works", just doing so (which sould work, and does in the REPL actually) results in:

```
[error]  found   : Seq[(some other)T(in method elements)]
[error]  required: Seq[T(in method elements)]
[error]   @varargs def elements[T](elems: T*): Source[T, Unit] =
[error]                ^
```

Thought I'd fix this while waiting on something, but seems it turned into a bigger task suddenly hm

agolubev:
Yeah, it's the same issue I have here #17865 it's  https://issues.scala-lang.org/browse/SI-8743. Still if solution will be found it works for both cases

ktoso:
Summary: I did not find a way to make this work under Scala 2.10; even implementing a base class in plain java does not help, as scalac needs to generate bridges anyway using scala for the companion object - see here for hack attempt: https://github.com/ktoso/akka/commit/2605001dc9cf132670ca940e32ec644e364dd97a

Proposed solutions (all bad tradeoffs):
- remove the method from both scaladsl/javadsl - because consistency
- remove the method from javadsl, until we can drop scala 2.10 - which is the point at which we merge with master
- drop compile time safety in javadsl, then it just works: `  @varargs def elements[T](elems: Any*): Source[T, Unit] = {`, we could add the T once we drop 2.10 then...
- this would be runtime safe, not compile time safe: 
```scala
  @varargs def elements[T](clazz: Class[T], elems: Any*): Source[T, Unit] = {
    require(elems.forall(_.getClass.isAssignableFrom(clazz)), s"elems must all be of type ${clazz.getCanonicalName}")
    new Source(scaladsl.Source(() ⇒ elems.iterator.asInstanceOf[Iterator[T]]))
  }
```
- *joke solution*: it works when `@varargs def elements[T](elems: T*): Unit` :wink:

So... we need to make a decision here: drop the method completely, or drop typesafety for javadsl. Opinions @drewhk @viktorklang @2m ?

viktorklang:
@ktoso In general varargs in scala is not to be considered an immutable collection so perhaps it is wisest just to keep those from immutable.Iterable?

ktoso:
This one is for JavaAPI, but in general yes and we also have that already:

```
[...]

PR #17950, opened 2015-07-09
-str #17923 remove javadsl's elements method, due to compiler bug

ktoso:
Dropping the method because of non-working `@varargs` in this case for javadsl (there was no scaladsl version of this).

More details in the ticket https://github.com/akka/akka/issues/17923
Resolves https://github.com/akka/akka/issues/17923


drewhk:
technically the comment messge should start with "!" indicating a breaking change. (+ means addition, which does not break things, "!" means removal or any other breaking change)

drewhk:
LGTM, it seems like this is not covered by DslConsistencySpec

ktoso:
Ah, `!` true. Yeah and it had no tests.

Issue #17891, opened 2015-07-02
Integration with reactive streams Processors


jroper:
Akka streams should provide mechanisms for working with Reactive Streams processors.

Here's a real world use case that I've just had to do.  Let's say I have a `Processor`, and I want to add an Akka flow step to it:

```java
Processor<HttpRequest, HttpResponse> processor = ...
Flow<List<String>, HttpRequest> myFlow = ...

Pair<Subscriber<List<String>>, Publisher<HttpRequest>> pair = 
    Source.<List<String>>subscriber()
    .via(myFlow)
    .toMat(Sink.<HttpRequest>publisher(), 
        Keep.<Subscriber<List<String>>, Publisher<HttpRequest>>both())
    .run(materializer);
pair.second().subscribe(processor);
Processor<List<String>, HttpResponse> newProcessor = 
    new DelegateProcessor<>(pair.first(), processor);
```

And that's assuming that you've already written this class:

```java
public class DelegateProcessor<In, Out> implements Processor<In, Out> {

    private final Subscriber<In> subscriber;
    private final Publisher<Out> publisher;

    public DelegateProcessor(Subscriber<In> subscriber, Publisher<Out> publisher) {
        this.subscriber = subscriber;
        this.publisher = publisher;
    }

    @Override
    public void subscribe(Subscriber<? super Out> s) {
        publisher.subscribe(s);
    }

    @Override
    public void onSubscribe(Subscription s) {
        subscriber.onSubscribe(s);
    }

    @Override
    public void onNext(In in) {
        subscriber.onNext(in);
    }

    @Override
    public void onError(Throwable t) {
        subscriber.onError(t);
    }

    @Override
    public void onComplete() {
        subscriber.onComplete();
    }
}
```

What it should look like is this:

```java
Processor<HttpRequest, HttpResponse> processor = ...
Flow<List<String>, HttpRequest> myFlow = ...

Processor<List<String>, HttpResponse> newProcessor = Flow.from(() -> processor)
    .via(myFlow)
    .toProcessor(materializer);
```

Note that `Flow.from()` takes a zero arg function, since flows can be reused, but processors can't - though in this case it's fine to return the same processor in each time, because the flow is converted to a processor, and so will only ever be used once.

ktoso:
2 features:

- via existing Processor
- given a Flow attach subscriber/publisher sink/source and expose a Processor

drewhk:
[...]

PR #17953, opened 2015-07-09
+str #17891: Add direct support for RS Processors

drewhk:
refs #17891

### DIFF
--- a/akka-docs-dev/rst/java/stream-integrations.rst
+++ b/akka-docs-dev/rst/java/stream-integrations.rst
@@ -399,7 +399,9 @@ Using an Akka Streams :class:`Flow` we can transform the stream and connect thos
 The :class:`Publisher` is used as an input :class:`Source` to the flow and the
 :class:`Subscriber` is used as an output :class:`Sink`.
 
-A :class:`Flow` can also be materialized to a :class:`Subscriber`, :class:`Publisher` pair:
+A :class:`Flow` can also be also converted to a :class:`RunnableGraph[Processor[In, Out]]` which
+materializes to a :class:`Processor` when ``run()`` is called. ``run()`` itself can be called multiple
+times, resulting in a new :class:`Processor` instance each time.
 
 .. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ReactiveStreamsDocTest.java#flow-publisher-subscriber
 
@@ -429,4 +431,9 @@ by using the Subscriber-:class:`Source`:
 
 .. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ReactiveStreamsDocTest.java#sink-subscriber
 
+It is also possible to use re-wrap :class:`Processor` instances as a :class:`Flow` by
+passing a factory function that will create the :class:`Processor` instances:
 
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ReactiveStreamsDocTest.java#use-processor
+
+Please note that a factory is necessary to achieve reusability of the resulting :class:`Flow`.
--- a/akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
+++ b/akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
@@ -4,10 +4,9 @@
 package docs.stream
 
 import akka.stream.ActorMaterializer
-import akka.stream.scaladsl.Flow
+import akka.stream.scaladsl.{ RunnableGraph, Flow, Sink, Source }
 import akka.stream.testkit._
-import akka.stream.scaladsl.Sink
-import akka.stream.scaladsl.Source
+import org.reactivestreams.Processor
 
 class ReactiveStreamsDocSpec extends AkkaSpec {
   import TwitterStreamQuickstartDocSpec._
@@ -78,11 +77,10 @@ class ReactiveStreamsDocSpec extends AkkaSpec {
     val storage = impl.storage
 
     //#flow-publisher-subscriber
-    val (in: Subscriber[Tweet], out: Publisher[Author]) =
-      authors.runWith(Source.subscriber[Tweet], Sink.publisher[Author])
+    val processor: Processor[Tweet, Author] = authors.toProcessor.run()
 
-    tweets.subscribe(in)
-    out.subscribe(storage)
+    tweets.subscribe(processor)
+    processor.subscribe(storage)
     //#flow-publisher-subscriber
 
     assertResult(storage)
@@ -133,6 +131,17 @@ class ReactiveStreamsDocSpec extends AkkaSpec {
     //#sink-subscriber
 
     assertResult(storage)
+  }
+
+  "use a processor" in {
+
+    //#use-processor
+    // An example Processor factory
+    def createProcessor: Processor[Int, Int] = Flow[Int].toProcessor.run()
+
+    val flow: Flow[Int, Int, Unit] = Flow(() => createProcessor)
+    //#use-processor
+
   }
 
 }
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
@@ -8,6 +8,8 @@ import akka.stream._
 import akka.japi.{ Util, Pair }
 import akka.japi.function
 import akka.stream.scaladsl
+import akka.stream.scaladsl.{ Keep, Sink, Source }
+import org.reactivestreams.{ Subscription, Publisher, Subscriber, Processor }
 import scala.annotation.unchecked.uncheckedVariance
 import scala.collection.immutable
 import scala.concurrent.Future
@@ -30,6 +32,9 @@ object Flow {
   /** Create a `Flow` which can process elements of type `T`. */
   def create[T](): javadsl.Flow[T, T, Unit] =
     adapt(scaladsl.Flow[T])
+
+  def create[I, O](processorFactory: function.Creator[Processor[I, O]]): javadsl.Flow[I, O, Unit] =
+    adapt(scaladsl.Flow(() ⇒ processorFactory.create()))
 
   /** Create a `Flow` which can process elements of type `T`. */
   def of[T](clazz: Class[T]): javadsl.Flow[T, T, Unit] =
@@ -72,12 +77,41 @@ class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Graph
 
   /**
    * Transform this [[Flow]] by appending the given processing steps.
+   *
+   * {{{
+   *     +----------------------------+
+   *     | Resulting Flow             |
+   *     |                            |
+   *     |  +------+        +------+  |
+   *     |  |      |        |      |  |
+   * In ~~> | this | ~Out~> | flow | ~~> T
+   *     |  |      |        |      |  |
+   *     |  +------+        +------+  |
+   *     +----------------------------+
+   * }}}
+   *
+   * The materialized value of the combined [[Flow]] will be the materialized
+   * value of the current flow (ignoring the other Flow’s value), use
+   * [[Flow#viaMat viaMat]] if a different strategy is needed.
    */
   def via[T, M](flow: Graph[FlowShape[Out, T], M]): javadsl.Flow[In, T, Mat] =
     new Flow(delegate.via(flow))
 
   /**
    * Transform this [[Flow]] by appending the given processing steps.
+   * {{{
+   *     +----------------------------+
+   *     | Resulting Flow             |
+   *     |                            |
+   *     |  +------+        +------+  |
+   *     |  |      |        |      |  |
+   * In ~~> | this | ~Out~> | flow | ~~> T
+   *     |  |      |        |      |  |
+   *     |  +------+        +------+  |
+   *     +----------------------------+
+   * }}}
+   * The `combine` function is used to compose the materialized values of this flow and that
+   * flow into the materialized value of the resulting Flow.
    */
   def viaMat[T, M, M2](flow: Graph[FlowShape[Out, T], M], combine: function.Function2[Mat, M, M2]): javadsl.Flow[In, T, M2] =
     new Flow(delegate.viaMat(flow)(combinerToScala(combine)))
@@ -802,6 +836,16 @@ class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Graph
   def log(name: String): javadsl.Flow[In, Out, Mat] =
     this.log(name, javaIdentityFunction[Out], null)
 
+  /**
+   * Converts this Flow to a [[RunnableGraph]] that materializes to a Reactive Streams [[org.reactivestreams.Processor]]
+   * which implements the operations encapsulated by this Flow. Every materialization results in a new Processor
+   * instance, i.e. the returned [[RunnableGraph]] is reusable.
+   *
+   * @return A [[RunnableGraph]] that materializes to a Processor when run() is called on it.
+   */
+  def toProcessor: RunnableGraph[Processor[In @uncheckedVariance, Out @uncheckedVariance]] = {
+    new RunnableGraphAdapter(delegate.toProcessor)
+  }
 }
 
 /**
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
@@ -14,8 +14,7 @@ import akka.stream.Attributes._
 import akka.stream._
 import akka.stream.impl.{ ActorPublisherSource, StreamLayout }
 import akka.util.ByteString
-import org.reactivestreams.Publisher
-import org.reactivestreams.Subscriber
+import org.reactivestreams.{ Processor, Publisher, Subscriber }
 import scala.annotation.unchecked.uncheckedVariance
 import scala.collection.JavaConverters._
 import scala.concurrent.{ Promise, Future }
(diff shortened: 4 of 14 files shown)


2m:
LGTM

akka-docs-dev/rst/java/stream-integrations.rst
```diff
@@ -429,4 +431,7 @@ by using the Subscriber-:class:`Source`:
 
 .. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/ReactiveStreamsDocTest.java#sink-subscriber
 
+It is also possible to use :class:`Processor` interfaces directly with both :class:`Source` and :class:`Flow` by
```
ktoso: Maybe a word or two that that's not really the recommended route?

```
This feature exists primarily for integrating with existing Processor implementations, and it is not recommended to implement this interface directly in end-user projects, unless there is a good reason to do so.
```

?

akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
```diff
@@ -135,4 +133,16 @@ class ReactiveStreamsDocSpec extends AkkaSpec {
     assertResult(storage)
   }
 
+  "use a processor" in {
+
+    //#use-processor
+    // An example Processor factory
+    def createProcessor: Processor[Int, Int] = Flow[Int].toProcessor.run()
```
ktoso: I would prefer `asProcessor`, because consistent with JavaConversions and our other conversions of "equivalent things". `to` sounds like our `to(sink)`.

ktoso: Looks good on Flow here though, let's see how it gets adopted

akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
```diff
@@ -135,4 +133,16 @@ class ReactiveStreamsDocSpec extends AkkaSpec {
...
+
+    val src: Source[Int, Unit] = Source.single(42).viaProcessor(() => createProcessor)
+    val flow: Flow[Int, Int, Unit] = Flow[Int].viaProcessor(() => createProcessor)
```
ktoso: OK

akka-docs-dev/rst/java/stream-integrations.rst
drewhk: This should go to the top of the section then, this is not unique to Processors.

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -72,12 +74,64 @@ class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Graph
...
+   *     |  |      |        |      |  |
+   * In ~~> | this | ~Out~> | flow | ~~> T
+   *     |  |      |        |      |  |
+   *     |  +------+        +------+  |
+   *     +----------------------------+
+   * }}}
+   *
+   * The materialized value of the combined [[Flow]] will be the materialized
+   * value of the current flow (ignoring the other Flow’s value), use
+   * [[Flow#viaMat viaMat]] if a different strategy is needed.
```
ktoso: great catch, those help a lot!

akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
drewhk: will see

akka-docs-dev/rst/java/stream-integrations.rst
ktoso: Yes, maybe as a note?

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -260,6 +259,18 @@ class Source[+Out, +Mat](delegate: scaladsl.Source[Out, Mat]) extends Graph[Sour
...
+   * Transform this [[Source]] by appending a Reactive Streams compliant [[org.reactivestreams.Processor]]. This
+   * method takes a factory that instantiates the [[org.reactivestreams.Processor]] for every materialization
+   * of the resulting [[Source]].
+   *
+   * The materialized value of the combined [[Source]] will be the materialized
+   * value of the current source.
+   */
+  def viaProcessor[T](processorFactory: function.Creator[Processor[Out @uncheckedVariance, T]]): Source[T, Mat] = {
+    new Source(delegate.viaProcessor(() ⇒ processorFactory.create()))
+  }
```
ktoso: Please copy the same graph from scaladsl to this one, it helps a lot when learning those things, see: https://github.com/akka/akka/pull/17953/files#diff-a4d5987f74d69093ccf1264366e4bbc1R111

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -282,6 +306,26 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
...
+    Source.subscriber[In].via(this).toMat(Sink.publisher[Out])(Keep.both[Subscriber[In], Publisher[Out]])
+      .mapMaterializedValue {
+        case (sub, pub) ⇒ new Processor[In, Out] {
+          override def onError(t: Throwable): Unit = sub.onError(t)
+          override def onSubscribe(s: Subscription): Unit = sub.onSubscribe(s)
+          override def onComplete(): Unit = sub.onComplete()
+          override def onNext(t: In): Unit = sub.onNext(t)
+          override def subscribe(s: Subscriber[_ >: Out]): Unit = pub.subscribe(s)
+        }
+      }
```
ktoso: Seems valid to me; do we want to TCK test it?

drewhk: Indriectly is already tested, see modifications to TCK :)

ktoso: Ah, didn't notice that's in TCK - excellent :-)

akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
viktorklang: I'd prefer not to have `viaProcessor` on Source, instead having a canonical way: `Source.single(42) via Flow[Int].processor(() => createProcessor())`

viktorklang: Perhaps even: `Flow.process(() => createProcessor())`

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: We don't have a choice.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -53,6 +53,30 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
...
+   *     |  |      |        |           |  |
+   * In ~~> | this | ~Out~> | processor | ~~> T
+   *     |  |      |        |           |  |
+   *     |  +------+        +-----------+  |
+   *     +---------------------------------+
+   * }}}
+   * The materialized value of the combined [[Flow]] will be the materialized
+   * value of the current flow.
+   */
+  def viaProcessor[T](processorFactory: () ⇒ Processor[Out @uncheckedVariance, T]): Flow[In, T, Mat] = {
```
viktorklang: Not sold on the name here. `transform` is not called `viaTransform`. I'd prefer: `process`

akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
drewhk: I can remove it from Source, yes. As for "process", I think that is too generic. viaProcessor states exactly what it is (overloading via directly is a no-go, it has unfortunate side-effects).

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: All TCK identityX test now use this method to make a Processor from a Flow, so this is tested against several backing Flows indirectly (originally it created a Publisher/Subscriber Pair, but now this is built in with this method)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: Process is too generic. Via overload is not really an option, and this is very clean what it does. "viaTransform" is not the same, the analogue would be "viaStage".

akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
viktorklang: Then every method should be "viaMap" "viaFilter" etc :)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: There's no `stage` method, the equivalent *is* `transform` which takes a () => Stage, compared to () => Processor.
Or am I wrong?

akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
drewhk: Yes, but this is an analogue to via() which takes a Flow, which is a prepackaged processing entity, just like Processor can be something heavyweight inside. I wanted an overload on via, but had lots of issues with that.

drewhk: Alternative is to have a .wrap() on Flow which converts a Processor to a Flow. Then you say Flow.via(Flow.wrap(processor))

ktoso: Similar feelings about that one as Endre: let's stick with `via...`, 
devs will get used to `Flow = Processor`, "stuff flows through it", "which API was that for flowing through stuff... oh right `via`", so they'll via and autocomplete/find the `viaProcessor`

ktoso: wrap proposal also nice, +1

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: No because "transform" refers to the act of transforming. "via" expresses wiring through a flow-like thing. Therefore, your analogue is wrong. Since "transform" is basically "via a Stage" then it should be called "viaStage" by your logic.

akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
drewhk: via and wrap are equivalent, one can emulate the other in this case, but via is more useful in practice (it does not break the fluidity of the DSL, wrap is a bit of boilerplate)

viktorklang: I disagree: *everything* on Flow is via. map, filter, the whole shebang.

@drewhk You mean `Source.via(Flow.wrap(processor))`? That would be OK if Source.wrap(publisher) and Sink.wrap(subscriber) would be the way to hoist those (symmetry and consistency ftw).

Flow.process(() => mkProcessor()) would be symmetric to Flow.transform(() => mkStage())

Or, perhaps what we have realized is that `transform` is the wrong name.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: No, that's -your- logic. My logic is to drop -via- for everything but piecing together sources with flows or flows with flows.

According to my logic, `transform` ought to be called `stage`.

akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
drewhk: Well, I still think "transform" and "viaProcessor" are the best options so far. I don't like calling "transform" viaStage, because it hides the intent of it being a generic stream transformation operation. You can argue that "viaProcessor" is also a generic stream transformation (in fact it is more generic), but that is not the preferred one. Calling it "process" invites trouble, because of the inevitable question of "what is the difference between process and transform? Which one should I use?"

viktorklang: It's very simple: use `process` when all you get is a Processor, use `transform` when all you get is a Stage.

viktorklang: Perhaps I am unclear: I am -extremely against- `viaProcessor` as -nearly every single method on Flow- -semantically- is *via* but doesn't carry it in its name.

I'm not married to the idea of calling it `process` but the `via` part of that suggestion is only inconsistent.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: > No, that's -your- logic.

No, you said:

> transform is not called viaTransform

Which I corrected, because it would be viaStage by that logic where something looks like viaX(X). 

> My logic is to drop -via- for everything but piecing together sources with flows or flows with flows.

Well, we cannot drop "via" for Flows. We can introduce a factory from Processor to Flow, that works fine. I don't like it because it is extra boilerplate though to write. 

> According to my logic, transform ought to be called stage.

Why rename after the final RC? If "viaProcessor" offends your eyes that much, then I introduce a Flow.apply() that takes a processor.

akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
drewhk: >  -nearly every single method on Flow- -semantically- is via but doesn't carry it in its name.

This is not true. All the non-via methods add exactly one stage, while .via() expresses the intent of adding a prepackaged entity. This is where we disagree, because for me .via() is for adding larger modules (a Flow or Processor), therefore for me it is *not the same*

But let's have a Flow.apply then, that settles this issue.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: >Well, we cannot drop "via" for Flows. We can introduce a factory from Processor to Flow, that works fine. I don't like it because it is extra boilerplate though to write.

That's not what I said. I said nothing but the `via` method should have via in its name.

>Why rename after the final RC?

Not strictly needed albeit the current name is suboptimal and a remnant from when things were called different things.

>If "viaProcessor" offends your eyes that much, then I introduce a Flow.apply() that takes a processor.

akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
viktorklang: A Processor is also a single stage. Just as Stage is a single stage.

drewhk: > A Processor is also a single stage. Just as Stage is a single stage.

That's your subjective interpretation. I have mine, too, which is different.

But let's go with the Flow factory method then.

viktorklang: Deal :)

drewhk:
So, the factory method version is actually quite nice, in fact, there are less problems with the types. I was a bit worried about overload hell, but FlowApply does not collide thankfully.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -292,6 +312,14 @@ object Flow extends FlowApply {
   private def shape[I, O](name: String): FlowShape[I, O] = FlowShape(Inlet(name + ".in"), Outlet(name + ".out"))
 
   /**
+   * Creates a Flow from a Reactive Streams [[org.reactivestreams.Processor]]
+   */
+  def apply[I, O](processorFactory: () ⇒ Processor[I, O]): Flow[I, O, Unit] = {
```
viktorklang: The downside of overloading here is that now we have wasted using Function1 never to be used again in Flow.apply.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: How about making this a Sink rather than a method on Flow?

viktorklang:
Great job @drewhk!

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: That is not possible, but there is an analogue, i.e. you can make a Flow that if joined with another Flow materializes to a Processor. To be honest, that would be the 100% consistent version:
 - Add a Sink to a Source to make it a Publisher
 - Add a Source to a Sink to make it a Subscriber
 - Join a Flow with a Flow to make it a Processor

... but I am afraid that in this case it is also the most confusing solution.

viktorklang: Yep, that'd definitely be confusing! But given a single Flow you could pair it with the identity flow :)

Issue #17956, opened 2015-07-09
Missing javadsl in akka-http-testkit-experimental_2.11 1.0-RC4.jar


2m:
Reported in the [mailing list](https://groups.google.com/forum/#!searchin/akka-user/Missing$20javadsl$20from/akka-user/PkdKIPMmjlo/amFLu6yNQd0J). `javadsl` classes are surely missing even when I `akka-http-testkit-experimental/publishLocal` on the latest `release-2.3-dev`

```
Archive:  /home/martynas/.ivy2/local/com.typesafe.akka/akka-http-testkit-experimental_2.10/1.0-SNAPSHOT/jars/akka-http-testkit-experimental_2.10.jar
  inflating: /tmp/testp/META-INF/MANIFEST.MF  
   creating: /tmp/testp/akka/
   creating: /tmp/testp/akka/http/
   creating: /tmp/testp/akka/http/scaladsl/
   creating: /tmp/testp/akka/http/scaladsl/testkit/
  inflating: /tmp/testp/akka/http/scaladsl/testkit/MarshallingTestUtils$$anonfun$marshal$1.class  
  inflating: /tmp/testp/akka/http/scaladsl/testkit/MarshallingTestUtils$class.class  
  inflating: /tmp/testp/akka/http/scaladsl/testkit/MarshallingTestUtils.class  
  inflating: /tmp/testp/akka/http/scaladsl/testkit/RouteTest$$anonfun$actorSystemNameFrom$1.class  
  inflating: /tmp/testp/akka/http/scaladsl/testkit/RouteTest$$anonfun$check$1.class  
  inflating: /tmp/testp/akka/http/scaladsl/testkit/RouteTest$$anonfun$entityAs$1.class  
  inflating: /tmp/testp/akka/http/scaladsl/testkit/RouteTest$$anonfun$header$1.class  
  inflating: /tmp/testp/akka/http/scaladsl/testkit/RouteTest$$anonfun$responseAs$1.class  
  inflating: /tmp/testp/akka/http/scaladsl/testkit/RouteTest$DefaultHostInfo$.class  
  inflating: /tmp/testp/akka/http/scaladsl/testkit/RouteTest$DefaultHostInfo.class  
  inflating: /tmp/testp/akka/http/scaladsl/testkit/RouteTest$TildeArrow$$anon$1$$anonfun$1.class  
[...]

PR #17958, opened 2015-07-09
=pro #17956 include javadsl.testkit in http-testkit artifact by disabling osgi config

jrudolph:
This fixes it for now. As http and http-core don't have any OSGI configuration, not providing any for http-testkit shouldn't break things any further.


2m:
LGTM, thanks!

PR #17965, opened 2015-07-10
Java side documentation progress

jrudolph:
Here's the somewhat self-contained set of finished documentation. I'm going to extend it with more examples and detail over the next days until the cut.

/cc @sirthias


jrudolph:
Added a commit detailing the `Route` => `Flow` conversion on the Scala side. Would that have helped you @fommil?

jrudolph:
(I slightly updated the footnote.)

fommil:
yes, that would have saved me several hours of frustration yesterday :smile:

akka-docs-dev/rst/scala/http/common/http-model.rst
```diff
@@ -93,9 +93,9 @@ HttpEntity.Strict
 
 HttpEntity.Default
   The general, unchunked HTTP/1.1 message entity.
-  It has a known length and presents its data as a ``Source[ByteString]`` which can only materialized once.
+  It has a known length and presents its data as a ``Source[ByteString]`` which can be only materialized once.
   It is an error if the provided source doesn't produce exactly as many bytes as specified.
-  On the wire, a ``Strict`` entity and a ``Default`` entity cannot be distinguished.
+  The distinction of ``Strict`` and ``Default`` is an API-only one. One the wire, both kinds of entities look the same.
```
ktoso: nice rewording

ktoso:
LGTM, looks good :-)

jrudolph:
Added a chapter about "handlers" on the java-side.

akka-docs-dev/rst/java/code/docs/http/javadsl/HandlerExampleSpec.java
```diff
@@ -0,0 +1,157 @@
...
+import akka.http.javadsl.testkit.TestRoute;
+import org.junit.Test;
+
+public class HandlerExampleSpec extends JUnitRouteTest {
+    @Test
+    public void testSimpleHandler() {
+        //#simple-handler-example-full
+        class TestHandler extends akka.http.javadsl.server.AllDirectives {
+            //#simple-handler
+            Handler handler = new Handler() {
```
ktoso: btw, since we now have Functions generated for `japi` in 2.3.x (I think since 2.3.11?), and bumped the streams dependency, wouldn't it make sense to drop Handler and use those `akka.japi.function.FunctionN` interfaces?

jrudolph: Wouldn't that be an argument for all SAM classes? I think it somewhat makes sense when the parameters are fixed to provide at least an alias. We could change the classes to extend from the matching FunctionX interfaces, though, and change the `handleWith` signatures to take matching `FunctionN`, instead. WDYT?

ktoso: Yeah, for many SAM interfaces now we could move to `akka.japi.function` since we have them (we did not previously have them for so many params).

I thought of this direction https://github.com/ktoso/akka/commit/8784c18f24e46e34c4138dbd8982a96ccdf12c2a#diff-58791b179be1c95f3090b06bb72bf473R1 but then we need to name then `handleWithN` since javac is not-very-good-with-overloads: https://gist.github.com/ktoso/76248c629bd614c81624

If we change then to extend the matching function interfaces and have function in method params as well, why keep the `Handler2`? If users want to they can make their Handler, and if we don't require a type in APIs, let's not have it.

Maybe I'm overthinking, but I feel that the API proposed in the above commit I linked "looks more like akka streams", so maybe javadsl users would be more familiar with it... The `handleWith##` uglieness is a cost though :~

Do you have an opinion on this @rkuhn @viktorklang (using Roland while it's still monday ;-))

ktoso: Ok, it seems we'll want to provide the "like akka-streams javadsl" style API here. I'll handle that, and Johannes focuses on remaining docs :)

ktoso:
Merging and will follow up with the changes discussed in https://github.com/ktoso/akka/commit/8784c18f24e46e34c4138dbd8982a96ccdf12c2a#commitcomment-12135171

Issue #17959, opened 2015-07-09
PersistentFsmActor persists StateChangeEvent on stay - intended behavior?


bardurdam:
I've noticed that `PersistentFsmActor` persists `StateChangeEvent` even when the state does not change. Is this intended behavior?

I am aware that `FSM` changes state for every message is processed, handled or not. And that `stay` actually means `goto(currentState)`.

I guess that redundant persistence of `StateChangeEvent`s can be avoided by making `def applyState(nextState: State)` in `PersistentFsmActor` only append the `StateChangeEvent` when the state actually changed. This might be accomplished by by comparing `nextState` with `currentState`, if `currentState`was made accessible or possibly by comparing `stateName` with that of `nextState.stateName`. 
I'm not sure how to take timeouts into account without `currentState` being accessible from `PersistentFsmActor`.

The example below produces the following events:
```scala
IncreaseCount
StateChangeEvent("Open", None)
StateChangeEvent("Open", None)
StateChangeEvent("Open", None)
StateChangeEvent("Open", None)
```

Example:

```scala
object Demo extends App {
  val system = ActorSystem("demo")
  val door = system.actorOf(Props(new PersistentDoorFsm()))

  door ! OpenDoor
  door ! OpenDoor
  door ! "foo"
  door ! "bar"
}

object PersistentDoorFsm {

  sealed trait State extends FSMState

  case object Open extends State {
    def identifier: String = "Open"
  }
  case object Closed extends State {
    def identifier: String = "Closed"
  }

  sealed trait Data
  case class Count(int: Int) extends Data

  sealed trait Evt
  case object IncreaseCount extends Evt

  case object OpenDoor
  case object CloseDoor
}

class PersistentDoorFsm(implicit val domainEventClassTag: ClassTag[Evt])
  extends PersistentFsmActor[State,Data,Evt] {

  def persistenceId: String = "door-fsm-stream"

  startWith(Closed, Count(0))

  when(Closed) {
    case Event(OpenDoor, _)  ⇒ 
      goto(Open) applying IncreaseCount
  }

  when(Open) {
    case Event(CloseDoor, _) ⇒ 
      goto(Closed) applying IncreaseCount
  }

  // unhandled will call stay

  def applyEvent(domainEvent: Evt, currentData: Data): Data = {
    (domainEvent, currentData) match {
      case (IncreaseCount, Count(cnt)) ⇒ Count(cnt+1)
      case _                           ⇒ currentData
    }
  }
}
```

[...]

PR #17974, opened 2015-07-12
=per #17959 PersistentFsmActor persists StateChangeEvent on stay

leonidb:
Prevent PersistentFSM from saving state change event on stay

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/fsm/PersistentFSM.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/fsm/PersistentFSM.scala
@@ -18,10 +18,12 @@ import scala.reflect.ClassTag
  * A FSM implementation with persistent state.
  *
  * Supports the usual [[akka.actor.FSM]] functionality with additional persistence features.
- * State and State Data are persisted on every state change.
  * `PersistentFSM` is identified by 'persistenceId' value.
+ * State changes are persisted atomically together with domain events, which means that either both succeed or both fail,
+ * i.e. a state transition event will not be stored if persistence of an event related to that change fails.
  * Persistence execution order is: persist -&gt; wait for ack -&gt; apply state.
  * Incoming messages are deferred until the state is applied.
+ * State Data is constructed based on domain events, according to user's implementation of applyEvent function.
  */
 trait PersistentFSM[S <: FSMState, D, E] extends PersistentActor with PersistentFSMBase[S, D, E] with ActorLogging {
   import akka.persistence.fsm.PersistentFSM._
@@ -80,14 +82,36 @@ trait PersistentFSM[S <: FSMState, D, E] extends PersistentActor with Persistent
    * Persist FSM State and FSM State Data
    */
   override private[akka] def applyState(nextState: State): Unit = {
-    val eventsToPersist: immutable.Seq[Any] = nextState.domainEvents.toList :+ StateChangeEvent(nextState.stateName.identifier, nextState.timeout)
-    var nextData: D = stateData
-    persistAll[Any](eventsToPersist) {
-      case domainEventTag(event) ⇒
-        nextData = applyEvent(event, nextData)
-      case StateChangeEvent(stateIdentifier, timeout) ⇒
-        super.applyState(nextState using nextData)
-        nextState.afterTransitionDo(stateData)
+    var eventsToPersist: immutable.Seq[Any] = nextState.domainEvents.toList
+
+    //Prevent StateChangeEvent persistence when staying in the same state, except when state defines a timeout
+    if (nextState.notifies || nextState.timeout.nonEmpty) {
+      eventsToPersist = eventsToPersist :+ StateChangeEvent(nextState.stateName.identifier, nextState.timeout)
+    }
+
+    if (eventsToPersist.isEmpty) {
+      //If there are no events to persist, just apply the state
+      super.applyState(nextState)
+    } else {
+      //Persist the events and apply the new state after all event handlers were executed
+      var nextData: D = stateData
+      var handlersExecutedCounter = 0
+
+      def applyStateOnLastHandler() = {
+        handlersExecutedCounter += 1
+        if (handlersExecutedCounter == eventsToPersist.size) {
+          super.applyState(nextState using nextData)
+          nextState.afterTransitionDo(stateData)
+        }
+      }
+
+      persistAll[Any](eventsToPersist) {
+        case domainEventTag(event) ⇒
+          nextData = applyEvent(event, nextData)
+          applyStateOnLastHandler()
+        case StateChangeEvent(stateIdentifier, timeout) ⇒
+          applyStateOnLastHandler()
+      }
     }
   }
 }
--- a/akka-persistence/src/test/scala/akka/persistence/fsm/PersistentFSMSpec.scala
+++ b/akka-persistence/src/test/scala/akka/persistence/fsm/PersistentFSMSpec.scala
@@ -5,7 +5,7 @@
 package akka.persistence.fsm
 
 import akka.actor._
-import akka.persistence.PersistenceSpec
+import akka.persistence.{PersistentActor, RecoveryCompleted, PersistenceSpec}
 import akka.persistence.fsm.PersistentFSM._
 import akka.testkit._
 import com.typesafe.config.Config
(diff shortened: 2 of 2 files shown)


ktoso:
OK TO TEST

ktoso:
PLS BUILD

akka-persistence/src/main/scala/akka/persistence/fsm/PersistentFSMActor.scala
```diff
@@ -17,9 +17,12 @@ import scala.reflect.ClassTag
  * FSM actor implementation with persistent state
  *
  * Supports the usual [[akka.actor.FSM]] functionality with additional persistence features.
- * State and State Data are persisted on every state change.
- * FSM is identified by 'persistenceId' value.
+ * State changes are persisted. FSM is identified by 'persistenceId' value.
  * Persistence execution order is: persist -&gt; wait for ack -&gt; apply state. Incoming messages are deferred until the state is applied.
+ *
+ * Persistent FSM DSL provides a possibility to persist domain events, which are saved atomically with the State change event,
+ * meaning they either both peresited successfully, or both fail.
```
ahjohannessen: *peresited* -> persisted :)

akka-persistence/src/main/scala/akka/persistence/fsm/PersistentFSMActor.scala
```diff
@@ -17,9 +17,12 @@ import scala.reflect.ClassTag
...
+ * State Data is constructed based on domain events, according to user's implementatio of applyEvent function.
```
ahjohannessen: *implementatio* -> implementation

leonidb:
Thanks, @ahjohannessen, 
it was a middle-of-the-night commit :)

leonidb:
@ktoso 
Any comments on this one, before I start rebasing and squashing?

akka-persistence/src/main/scala/akka/persistence/fsm/PersistentFSMActor.scala
```diff
@@ -17,9 +17,12 @@ import scala.reflect.ClassTag
  * FSM actor implementation with persistent state
  *
  * Supports the usual [[akka.actor.FSM]] functionality with additional persistence features.
- * State and State Data are persisted on every state change.
- * FSM is identified by 'persistenceId' value.
+ * State changes are persisted. FSM is identified by 'persistenceId' value.
```
ktoso: Maybe: `State changes are persisted on every state change (except when `stay` is used)`

akka-persistence/src/main/scala/akka/persistence/fsm/PersistentFSMActor.scala
```diff
@@ -17,9 +17,12 @@ import scala.reflect.ClassTag
...
  * Persistence execution order is: persist -&gt; wait for ack -&gt; apply state. Incoming messages are deferred until the state is applied.
+ *
+ * Persistent FSM DSL provides a possibility to persist domain events, which are saved atomically with the State change event,
```
ktoso: remove `DSL` please

akka-persistence/src/main/scala/akka/persistence/fsm/PersistentFSMActor.scala
ktoso: Ok, let's do:

- Remove the section https://github.com/akka/akka/pull/17974/files#diff-ef6f00459ea008459f3774f7875a3e5eR23
- Make the above: "State changes are persisted atomically together with domain events, which means that either both succeed or both fail, i.e. a state transition event will not be stored if persistence of an event related to that change fails".

Is that right?

leonidb: Sounds good.

akka-persistence/src/test/scala/akka/persistence/fsm/PersistentFSMActorSpec.scala
```diff
@@ -223,6 +223,56 @@ abstract class PersistentFSMActorSpec(config: Config) extends PersistenceSpec(co
...
+      expectMsg(NonEmptyShoppingCart(List(shirt, shoes, coat)))
+
+      expectMsg(NonEmptyShoppingCart(List(shirt, shoes, coat)))
+
+      expectTerminated(fsmRef)
+
+      val persistentEventsStreamer = system.actorOf(PersistentEventsStreamer.props(persistenceId, testActor))
+
+      expectMsg(ItemAdded(Item("1", "Shirt", 59.99F)))
+      expectMsgClass(classOf[StateChangeEvent])   //because a timeout is defined, State Change is persisted
```
ktoso: minor, `expectMsgType[StateChangeEvent]` is less typing :)

akka-persistence/src/test/scala/akka/persistence/fsm/PersistentFSMActorSpec.scala
```diff
@@ -349,6 +399,20 @@ object PersistentFSMActorSpec {
     def props(persistenceId: String, reportActor: ActorRef) =
       Props(new WebStoreCustomerFSMActor(persistenceId, reportActor))
   }
+
+  class PersistentEventsStreamer(id: String, client: ActorRef) extends PersistentView {
```
ktoso: Let's make it a `PersistentActor`, as we'll try to remove `PersistentView` soon.

akka-persistence/src/test/scala/akka/persistence/fsm/PersistentFSMActorSpec.scala
leonidb: I'll fix it

leonidb:
@ktoso 
What about the implementation itself? Is it OK? 
```def applyState(nextState: State): Unit``` got pretty complicated...

I've tried to use ```deferAsync``` as a handler that is called after all event handlers were called, but it doesn't defer the incoming events. Would be handy to have a "sync" version of ```deferAsync``` or some other special handler called when all handlers were executed.

As I haven't found anything like the above, I've used a counter.

patriknw:
LGTM, I'm fine with the counter.
Please rebase. We are releasing 2.3-M3 on Friday and it would be great to include this.
cc @ktoso

patriknw:
please also squash into one commit

leonidb:
@patriknw 
Great! I'll try to finish it today.

leonidb:
@ktoso 
The merge with your changes led to a test failure. I think that this failure was due to a small bug that existed in PersistentFSM, and the current behavior is actually the correct one. See the comments to my last commit.
Please take a look, I'll squash the last 2 commits if everything is OK.

leonidb:
@ktoso @patriknw 
After giving it some thought, I think I was wrong. In the regular FSM every command handled by the actor resets the timeout, and then the new state either sets the timeout again or leaves it as is.

So, rewriting the explanation of the test failure from my last commit:
1. ```AddItem``` is handled -> ```StateChangeEvent``` is persisted, setting a timeout to 1 sec
2. ```GetCurrentCart``` is handled -> ```StateChangeEvent``` is not persisted since the event doesn't  cause a state to change, doesn't emit any domain events and doesn't set the timeout to anything.
 In the usual ```FSM``` flow not setting timeout resets the previously set timeout.
 In ```PersistentFSM``` prior to this bug fix a None timeout was persisted, leading to recovery without state timeout.
After the bug fix ```PersistentFSM``` doesn't store the ```StateChangeEvent```, leading to recovery with the timeout of 1 sec.

Can be fixed by storing the last applied timeout  in a ```var``` in ```PersistentFSM``` and using it to determine if the new state changes the previously set timeout.

leonidb:
@ktoso @patriknw 
Alternatively, this can be considered a valid behavior, since in ```PersistentFSM``` events that don't affect domain and don't change state shouldn't normally reset receive timeout?

This does feel like a corner case which should rarely happen in real ```PersistentFSM``` implementation.

Maybe this can be addressed later, as  a part of the effort to remove the duplication between ```FSM``` and ```PersistentFSM```?

Not sure if this is worth fixing right now.
What do you think?

patriknw:
I'm not sure what semantics that are desired, but I'm fine with leaving it as a corner case and adjust in a separate PR, if needed. Please add a separate issue, or write a comment in the DRY PersistentFSM/FSM issue.

patriknw:
PLS BUILD

patriknw:
@ktoso this is ready right, I'd like to merge before we introduce any new merge conflict

ktoso:
LGTM, let's open a ticket to investigate exact semantics of timeouts in FSM - would you open the ticket @leonidb? Thanks!

patriknw:
great, let's merge after final pr validation

leonidb:
> investigate exact semantics of timeouts in FSM

@ktoso Do you mean ```PersistentFSM``` timeout following recovery issue that has led to the test failure, or a more general investigation of ```FSM``` state timeouts?

ktoso:
I mean what you described in comments https://github.com/akka/akka/pull/17974#issuecomment-130554047 https://github.com/akka/akka/pull/17974#issuecomment-130559219 I'm not sure yet myself which behaviour is valid, will need to focus on it, so a issue would help :)

patriknw:
Refs #17959

Issue #15870, opened 2014-09-12
Fix OSGI for all akka-http projects


jrudolph:
Also see #15628

drewhk:
/cc @sirthias @jrudolph 
Can we fix this until 1.0?

sirthias:
Well, I can't. I don't know any OSGi foo. I guess we'll need the help of some OSGi expert.
Maybe @hseeberger can help?

hseeberger:
<rant>Any library that uses class loaders explicitly is not OSGi friendly. That includes Akka. I have said numerous times that I would remove OSGi "support".</rant>

Having said that, I don't consider myself an OSGi expert any more. That was long ago in some dark ages ...

rkrzewski:
I've managed to get Spray running in OSGi environment as a hobby project: https://github.com/rkrzewski/spray-osgi. I found myself jumping through a lot of flaming hoops to get it to run (see  [1] and the rest of the package). I concur that Akka is not the most OSGi friendly library around...
I guess I could help with getting the streams/http OSGi headers straight, but are you sure that there are people out there that want to use those as bundles? ;)

[1] https://github.com/rkrzewski/spray-osgi/blob/master/com.typesafe.akka.osgi.ds/src/akka/osgi/ds/impl/ActorFacade.scala

rkrzewski:
I've produced a set of bundles for Scala 2.10 over the weekend: streams, http(-core) and marshallers. 
I'll post a PR for review tonight.

ktoso:
Sounds excellent! We are preparing to release `1.0` today actually.
If you'd make the PR before "very late" I think it could make it into this cut. Let me know if this sounds realistic :-)

rkrzewski:
Oh. I'm at the office right now, but in this case I'll do the PR ASAP.

ktoso:
Thanks a lot and don't worry, we can wait a bit for this to come in, I'll happily delay the release until evening if we can fix osgi in 1.0 thanks to that :-)

rkrzewski:
It's in. Handling `scala.xml` in 2.11 still needs fixing though.

PR #17979, opened 2015-07-13
+pro #15870 OSGi headers for experimental Streams and HTTP modules

rkrzewski:


project/AkkaBuild.scala
```diff
@@ -1382,14 +1385,37 @@ object AkkaBuild extends Build {
...
     val httpTestKit = exports(Seq("akka.http.scaladsl.testkit.*"))
 
+    val httpSprayJson = exports(Seq("akka.http.scaladsl.marshallers.sprayjson"),
+        imports = Seq(streamAndHttpImport("akka.stream.*"),
+            streamAndHttpImport("akka.http.*")))
+
+    val httpXml = exports(Seq("akka.http.scaladsl.marshallers.xml"),
+        imports = Seq(streamAndHttpImport("akka.stream.*"),
+            streamAndHttpImport("akka.http.*")))
+    // TODO Scala 2.11+ needs also versionedImport("scala.xml.*", "1.0", "1.1")
```
rkrzewski: It would be great if someone having a clue about SBT could give me a hand here

project/AkkaBuild.scala
```diff
@@ -1382,14 +1385,37 @@ object AkkaBuild extends Build {
...
-    val http = Nil // FIXME #15689
+    val http = exports(Seq("akka.http.impl.server",
+        "akka.http.scaladsl.server.*", "akka.http.javadsl.server.*",
+        "akka.http.scaladsl.client", "akka.http.scaladsl.coding", "akka.http.scaladsl.common",
+        "akka.http.scaladsl.marshalling", "akka.http.scaladsl.unmarshalling"),
+        imports = Seq(streamAndHttpImport("akka.stream.*"),
+            streamAndHttpImport("akka.http.*"),
+            streamAndHttpImport("akka.parboiled2.*")))
 
     val httpTestKit = exports(Seq("akka.http.scaladsl.testkit.*"))
```
rkrzewski: Is this module expected to be used in OSGi runtime?

ktoso: Yes, it's the same style of testkit that we provide for actors - a set of utilities to make testing simpler.

ktoso:
refs #15870

ktoso:
OK TO TEST

project/AkkaBuild.scala
2beaucoup: related: https://github.com/akka/akka/issues/17956 (javadsl is missing in artifacts)

rkrzewski: If you really mean to support running the tests _inside_ an OSGi framework with PAX Exam or similar, testkit modules need OSGi headers too.

ktoso: I'm not sure, we probably should - I mean what if someone uses stuff like arquillian etc. Does that count as osgi container? It may I gues..

rkrzewski:
I have amended the commit with OSGi settings for akka-stream-testkit and akka-http-testkit
akka-http-testkit is problematic because it depends on scalatest. scalatest is available as a bundle, but a proorly constructed one: it has mandatory dependencies on a ton of stuff (jmock, easymock, mockito, selenium, ant and more) and some of the dependencies is _not_ available as bundles, most notably ... junit and hamcrest.
Does it make sense to use akka-http-testkit without scalatest in the runtime classpath? Making the dependency optional would give the bundle a chance to be loaded.

rkrzewski:
merge conflict, will rebase.

rkrzewski:
There's one remaining problem: `scala.xml` needs to be imported with different version when building against Scala 2.11+. I'd like do do something like this:
```scala
def scalaXmlImport: Seq[String] = 
  if(scalaVersion == "2.10") Nil
  else Seq(versionedImport("scala.xml.*", "1.0", "1.1"))

val httpXml = exports(Seq("akka.http.scaladsl.marshallers.xml"),
        imports = Seq(streamAndHttpImport("akka.stream.*"),
            streamAndHttpImport("akka.http.*")) ++ scalaXmlImport)
```
but I don't know the proper SBT idiom to do the ```scalaVersion == "2.10"``` part.

rkrzewski:
Done. I verified locally that correct imports are generated both in 2.10 and 2.11 builds.
If you prefer those 2 commits squashed, let me know.

project/AkkaBuild.scala
```diff
@@ -1382,16 +1385,43 @@ object AkkaBuild extends Build {
...
+    val http = exports(Seq("akka.http.impl.server",
+        "akka.http.scaladsl.server.*", "akka.http.javadsl.server.*",
+        "akka.http.scaladsl.client", "akka.http.scaladsl.coding", "akka.http.scaladsl.common",
+        "akka.http.scaladsl.marshalling", "akka.http.scaladsl.unmarshalling"),
+        imports = Seq(streamAndHttpImport("akka.stream.*"),
+            streamAndHttpImport("akka.http.*"),
+            streamAndHttpImport("akka.parboiled2.*")))
 
-    val httpCore = Nil // FIXME #15689
+    val httpTestKit = exports(Seq("akka.http.scaladsl.testkit.*"),
```
2beaucoup: javadsl?

rkrzewski: fixed

project/AkkaBuild.scala
```diff
@@ -1428,11 +1458,17 @@ object AkkaBuild extends Build {
       "com.google.protobuf")
 
     def exports(packages: Seq[String] = Seq(), imports: Seq[String] = Nil) = osgiSettings ++ Seq(
-      OsgiKeys.importPackage := imports ++ defaultImports,
+      OsgiKeys.importPackage := imports,
+      OsgiKeys.importPackage <++= scalaVersion { v =>
```
ktoso: `"2.10"` is `scalaBinaryVersion` :)

ktoso:
Looks great, thank you @rkrzewski!
After merging all other PRs I'll get to this one and verify we're not missing anything, then - release 1.0!

project/AkkaBuild.scala
jrudolph: Are you sure this change is at the right place? This will add `scala.xml` to any package that uses `exports`, right?

rkrzewski: @jrudolph no worries, BND is smart and will add it to `Import-Package` header only when actually needed. I do this in this place, because it needs to be in front of `scala.*` to be effective.

@ktoso should I change the condition to `scalaBinaryVersion { v => if(v == "2.10") ...` or should it stay as is? I actually nicked that from cpsPlugin settings declaration at line 170

2m:
I wrote a [quick script](https://gist.github.com/2m/51fc0490ffea7bdc4cb9) to verify that jars contain all of the compiled class files. So this LGTM.

ktoso:
thanks! That's super helpful, so confidently merging this.

project/AkkaBuild.scala
```diff
@@ -1382,16 +1385,43 @@ object AkkaBuild extends Build {
...
+    val http = exports(Seq("akka.http.impl.server",
+        "akka.http.scaladsl.server.*", "akka.http.javadsl.server.*",
+        "akka.http.scaladsl.client", "akka.http.scaladsl.coding", "akka.http.scaladsl.common",
+        "akka.http.scaladsl.marshalling", "akka.http.scaladsl.unmarshalling"),
+        imports = Seq(streamAndHttpImport("akka.stream.*"),
+            streamAndHttpImport("akka.http.*"),
+            streamAndHttpImport("akka.parboiled2.*")))
 
-    val httpCore = Nil // FIXME #15689
+    val httpTestKit = exports(Seq("akka.http.scaladsl.testkit.*", "akka.http.javaadsl.testkit.*"),
```
ktoso: excellent, thanks @rkrzewski

rkrzewski:
:sunglasses:

project/AkkaBuild.scala
jrudolph: Bam! `javaadsl` spoiled the show. Hard to spot, though...

ktoso: aaaaargh :<

rkrzewski: sorry about that :(

jrudolph: No worries about that, this is so easy to get wrong. The problem is that there are no safeguards that would catch a problem like this.

ktoso: Yeah, no worries - I mean, I reviewed and looked at this exact line and still missed the typo.
Will be fixed soon in 1.1 tho

rkrzewski: Hmm, shouldn't the script written by @2m have detected this problem?

Also, the verification script should check that the generated jar does not contain _more_ classes than `target/classes` folder, because it's easy to accidentally re-export packages from an upstream dependency if a too broad Export declaration is used (which is completely valid, and occasionally useful in OSGi environment)

PR #17983, opened 2015-07-14
+htc #17982 introduce ModeledCompanion.parseFromValueString, MediaType.parse, and ContentType.parse

### LINKED ISSUE #17982: Provide the ability to create a ContentType from a String
The `ContentType` companion object should provide a `parse` object, in a similar fashion to `MediaType`, to make it simple to create content types from Strings.

Consider a situation where you're integrating with a third party library, let's say some sort of file repository, and for each file, it includes a content type property as a String, and you want to serve these files using Akka HTTP.  Akka HTTP, as far as I can see, provides no means to create a `ContentType` from a String, you can create a `ContentType` from a `MediaType` and a `HttpCharset`, and both of these can be parsed from strings, but using that would first require parsing the value to extract both the media type and the charset out, if it exists.  The only way I can see to do this is via a round about method of mocking a header map, parsing it, then trying to get the content type header from that parsed map.

jrudolph:
Fixes #17982.

/cc @sirthias, @jroper

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
@@ -79,6 +79,13 @@ object ContentType {
       }
     new ContentType(mediaType, definedCharset) {}
   }
+
+  /**
+   * Tries to parse a ``ContentType`` value from the given String. Returns ``Right(contentType)`` if successful and
+   * ``Left(errors)`` otherwise.
+   */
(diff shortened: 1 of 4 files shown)


jrudolph:
It would be nice if this could be released as well, this failed for Jenkins reasons I think.

ktoso:
PLS BUILD

2m:
LGTM

ktoso:
Very nice, LGTM

Issue #17984, opened 2015-07-14
ContentNegotiationSpec almost fails completely if tests are properly enabled


jrudolph:
The syntax `(exampleName: String) in (f: String => U)` doesn't actually run `f`. Once that is fixed 13 out of 16 tests fail.

From a glance I cannot say if the implementation is more wrong or the tests. WDYT, @sirthias?

I'll push a PR shortly which enables the tests but fails them.

jrudolph:
#17987 enables the tests.

2beaucoup:
In the first test:

https://github.com/spray/akka/blob/w/17984-enable-content-negotiation-tests/akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala#L123

calls

https://github.com/spray/akka/blob/w/17984-enable-content-negotiation-tests/akka-http/src/main/scala/akka/http/scaladsl/marshalling/PredefinedToEntityMarshallers.scala#L24

`withFixedCharset()` -> `withOpenCharset()`

Could that be the cause?

Stack:
```
  at akka.http.scaladsl.marshalling.Marshaller$.withOpenCharset(Marshaller.scala:112)
	  at akka.http.scaladsl.marshalling.PredefinedToEntityMarshallers$class.byteArrayMarshaller(PredefinedToEntityMarshallers.scala:24)
	  at akka.http.scaladsl.marshalling.Marshaller$.byteArrayMarshaller(Marshaller.scala:66)
	  at akka.http.scaladsl.marshalling.PredefinedToEntityMarshallers$class.$init$(PredefinedToEntityMarshallers.scala:16)
	  at akka.http.scaladsl.marshalling.Marshaller$.<init>(Marshaller.scala:67)
	  at akka.http.scaladsl.marshalling.Marshaller$.<clinit>(Marshaller.scala:-1)
	  at akka.http.scaladsl.marshalling.ContentNegotiationSpec$$anonfun$testHeaders$1$$anonfun$2.apply(ContentNegotiationSpec.scala:123)
	  at akka.http.scaladsl.marshalling.ContentNegotiationSpec$$anonfun$testHeaders$1$$anonfun$2.apply(ContentNegotiationSpec.scala:122)
	  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
	  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
[...]

PR #17987, opened 2015-07-14
ContentNegotiation fixes (and breakage)

### LINKED ISSUE #17985: Marshal assumes List for charset negotiation
However, `HttpRequest.acceptedCharsetRanges` may return any other immutable Seq depending on the actual value of ranges in `Accept-Charset`. This is usually not a problem because the header parser only creates Lists at this position. Still, this is a interoperability problem, e.g. with tests where a request may be manually created.

/cc @sirthias

jrudolph:
Fixes #17985. Only fixes the "not run tests part" from #17984.

/cc @sirthias, @2beaucoup

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
@@ -13,83 +13,82 @@ import MediaTypes._
 import HttpCharsets._
 
 class ContentNegotiationSpec extends FreeSpec with Matchers {
-
   "Content Negotiation should work properly for requests with header(s)" - {
 
-    "(without headers)" in test { accept ⇒
+    "(without headers)" test { accept ⇒
       accept(`text/plain`) should select(`text/plain`, `UTF-8`)
       accept(`text/plain` withCharset `UTF-16`) should select(`text/plain`, `UTF-16`)
     }
 
-    "Accept: */*" in test { accept ⇒
+    "Accept: */*" test { accept ⇒
       accept(`text/plain`) should select(`text/plain`, `UTF-8`)
       accept(`text/plain` withCharset `UTF-16`) should select(`text/plain`, `UTF-16`)
     }
 
-    "Accept: */*;q=.8" in test { accept ⇒
+    "Accept: */*;q=.8" test { accept ⇒
       accept(`text/plain`) should select(`text/plain`, `UTF-8`)
       accept(`text/plain` withCharset `UTF-16`) should select(`text/plain`, `UTF-16`)
     }
 
-    "Accept: text/*" in test { accept ⇒
+    "Accept: text/*" test { accept ⇒
       accept(`text/plain`) should select(`text/plain`, `UTF-8`)
       accept(`text/xml` withCharset `UTF-16`) should select(`text/xml`, `UTF-16`)
       accept(`audio/ogg`) should reject
     }
 
-    "Accept: text/*;q=.8" in test { accept ⇒
+    "Accept: text/*;q=.8" test { accept ⇒
       accept(`text/plain`) should select(`text/plain`, `UTF-8`)
       accept(`text/xml` withCharset `UTF-16`) should select(`text/xml`, `UTF-16`)
       accept(`audio/ogg`) should reject
     }
 
-    "Accept: text/*;q=0" in test { accept ⇒
+    "Accept: text/*;q=0" test { accept ⇒
       accept(`text/plain`) should reject
       accept(`text/xml` withCharset `UTF-16`) should reject
       accept(`audio/ogg`) should reject
     }
 
-    "Accept-Charset: UTF-16" in test { accept ⇒
+    "Accept-Charset: UTF-16" test { accept ⇒
       accept(`text/plain`) should select(`text/plain`, `UTF-16`)
       accept(`text/plain` withCharset `UTF-8`) should reject
     }
 
-    "Accept-Charset: UTF-16, UTF-8" in test { accept ⇒
+    "Accept-Charset: UTF-16, UTF-8" test { accept ⇒
       accept(`text/plain`) should select(`text/plain`, `UTF-8`)
       accept(`text/plain` withCharset `UTF-16`) should select(`text/plain`, `UTF-16`)
     }
 
-    "Accept-Charset: UTF-8;q=.2, UTF-16" in test { accept ⇒
+    "Accept-Charset: UTF-8;q=.2, UTF-16" test { accept ⇒
       accept(`text/plain`) should select(`text/plain`, `UTF-16`)
       accept(`text/plain` withCharset `UTF-8`) should select(`text/plain`, `UTF-8`)
     }
 
-    "Accept-Charset: UTF-8;q=.2" in test { accept ⇒
+    "Accept-Charset: UTF-8;q=.2" test { accept ⇒
       accept(`text/plain`) should select(`text/plain`, `ISO-8859-1`)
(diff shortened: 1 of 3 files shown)


jrudolph:
This should fail for 13 previously disabled tests. It could either be that
 * the tests are broken (unlikely)
 * the implementation is broken
 * the testing infrastructure is broken (it was broken before the first commit of this PR, so no guarantees...)

jrudolph:
I think this may be an oversight from removing the previous plural `Marshallers` classes, so that the `implicit val marshallers` in the test isn't effective in any way which would be good news because it would mean that maybe only the test infrastructure is broken.

jrudolph:
Changing the test code to

```scala
  trait X // otherwise unknown type, preventing that a `ToEntityMarshaller[X]` is already in scope
  object X extends X
  def testHeaders[U](headers: HttpHeader*)(body: ((ContentType*) ⇒ Option[ContentType]) ⇒ U): U = {
    val request = HttpRequest(headers = headers.toVector)
    body { contentTypes ⇒
      import scala.concurrent.ExecutionContext.Implicits.global
      implicit val marshallers: ToEntityMarshaller[X] =
        Marshaller.oneOf(contentTypes map {
          case ct @ ContentType(mt, Some(cs)) ⇒
            println(s"With cs $cs")
            Marshaller.withFixedCharset(mt, cs)((s: X) ⇒ HttpEntity(ct, "The X"))
          case ContentType(mt, None) ⇒ Marshaller.withOpenCharset(mt)((s: X, cs) ⇒ HttpEntity(ContentType(mt, cs), "The X"))
        }: _*)

      Await.result(Marshal(X).toResponseFor(request)
        .fast.map(response ⇒ Some(response.entity.contentType))
        .fast.recover { case _: Marshal.UnacceptableResponseContentTypeException ⇒ None }, 1.second)
    }
  }
```

makes most tests pass but two.

2beaucoup:
I'm also suspecting implict failures. (See how the wrong marshaller gets picked up [here](https://github.com/akka/akka/issues/17984#issuecomment-121214765).)

jrudolph:
I think I have a fix.

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
```diff
@@ -13,83 +16,88 @@ import MediaTypes._
...
+    "Accept-Charset: UTF-8;q=.2, UTF-16" test { accept ⇒
       accept(`text/plain`) should select(`text/plain`, `UTF-16`)
       accept(`text/plain` withCharset `UTF-8`) should select(`text/plain`, `UTF-8`)
     }
 
-    "Accept-Charset: UTF-8;q=.2" in test { accept ⇒
-      accept(`text/plain`) should select(`text/plain`, `ISO-8859-1`)
+    "Accept-Charset: UTF-8;q=.2" test { accept ⇒
+      // FIXME: which rule would enforce that behavior?
+      // accept(`text/plain`) should select(`text/plain`, `ISO-8859-1`)
```
jrudolph: @sirthias, @2beaucoup any idea, which rule would govern behavior like that?

2beaucoup: Maybe a leftover from 8c8400303fa21112a6764b992fc83bb519e844f2?

Switching to `UTF-8` here (and `ISO-8859-1` below) should be fine.

jrudolph: Ah yes, good find. I guess then we need to turn the test around.

jrudolph: Hmm, but it still makes no sense as choosing an explicit charset should always take that one, right?

jrudolph: I would change it to

```scala
    "Accept-Charset: ISO-8859-1;q=.2" test { accept ⇒
      accept(`text/plain`) should select(`text/plain`, `ISO-8859-1`)
      accept(`text/plain` withCharset `UTF-8`) should reject
    }
```

This would definitely test something else but maybe the test was just bogus?

jrudolph: It seems this test case was transferred from spray's https://github.com/spray/spray/blob/master/spray-http/src/test/scala/spray/http/ContentNegotiationSpec.scala#L79 (which I don't understand either :))

jrudolph: I changed it now to the above suggestion (though, I'm not sure if that test adds something interesting).

2beaucoup: LGTM! I was playing around with that one as well and arrived at the same result. ;)

Looks like the logic changed in that we now reject instead of falling back to the default charset (`ISO-8859-1` back then) in that case.

jrudolph: Cool, so we agree. Thanks again for helping out @2beaucoup :)

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
```diff
@@ -13,83 +16,87 @@ import MediaTypes._
 import HttpCharsets._
 
 class ContentNegotiationSpec extends FreeSpec with Matchers {
-
   "Content Negotiation should work properly for requests with header(s)" - {
 
-    "(without headers)" in test { accept ⇒
-      accept(`text/plain`) should select(`text/plain`, `UTF-8`)
+    "(without headers)" test { accept ⇒
```
ktoso: tricky! thanks for noticing!

ktoso:
LGTM, thanks

akka-http-tests/src/test/scala/akka/http/scaladsl/marshalling/ContentNegotiationSpec.scala
sirthias: Sorry for being way late to the party here, but I'd like to shine some light on this test and hopefully clear up the confusion.
This test was originally written for [section 14.2 of RFC 2616](http://tools.ietf.org/html/rfc2616#section-14.2), which says:

> If no "*" is present in an Accept-Charset field, then all character sets not explicitly mentioned get a quality value of 0, except for ISO-8859-1, which gets a quality value of 1 if not explicitly mentioned.

However, in the latest specs `ISO-8859-1` doesn't have this special status anymore and the test should therefore fail (as it did). So the fix here was indeed necessary and not incorrect. :)

jrudolph: @sirthias cool, thanks for the explanation and digging that out.

2beaucoup: Ahh, all is well then. Thanks for clarifying @sirthias!

sirthias: To me this proves once more that you can never add too many pointers to the relevant specs right next to the code. If I had put a pointer to RFC2616 section 14.2 next to this test you guys would have figured things out in no time.

Issue #17338, opened 2015-04-29
add OutputStreamSource and InputStreamSink


rkuhn:
these materialize to the `java.io` types in order to interface with legacy libraries that are based on them; propagating back-pressure will block those libraries

agolubev:
@drewhk there are three approaches I can see here (further speculations are about ```OutputStreamSource``` as ```InputStreamSink``` would be similar):
1. Simplest one is to use ```InputStreamPublisher``` and ```PipedOutputStream```. So the ```OutputStreamSource``` and ```InputStreamSink``` will be syntax sugar (need to keep additional buffer though).
2. Second option is to create ```OutputStreamSourcePublisher``` the similar way as ```InputStreamPublisher(is: InputStream, completionPromise: Promise[Long], chunkSize: Int, initialBuffer: Int, maxBuffer: Int)```. Meaning synchronize only with chunks of fixed size. Method ```write``` will block only when buffer is full. Flush will do nothing
3. As third option there can be implemented something manageable like: 
a) chunk size can vary depends on what was sent via write methods
b) send downstream when buffer is full or when flush is called (when got back-pressure event)
c) block when buffer is full, when flush is called, or on every write method (unblock when data is sent  to down stream)

drewhk:
I think it is important to have flush() actually block until all the buffered things have been passed via onNext. It is probably also nice to not block on write when there is enough space in the buffer. I would not use PipedOutputStream.

agolubev:
Spent some time on this and came to the point when need to organize concurrent access to buffer.
From one side publisher (similar to InputStreamSource) will hold this buffer and respond to back pressure messages. From another - XxxxxStreamAdapter needs to update it and block calls if needed. 

The simplest and most consistent way is to make both publisher and adapter as actors. 
But here I'm not sure from governance standpoint. I mean - is it safe to create custom actor along with publisher actor within stream hierarchy. Is there any example of such custom actor in existing code?

If this strategy is not right it requires a lot of synchronizations around buffer.

ktoso:
Sounds to me like it would be better implemented inside one actor, otherwise it'll get a bit weird, trying to coordinate who can touch what when.

agolubev:
Got it. Thanks

agolubev:
@ktoso One more question... If I do publisherActorRef.ask(message, timeout) from non actor - does it mean I'm creating second actor in system?
[...]

PR #17995, opened 2015-07-15
+str #17338 add OutputStreamSource and InputStreamSink 

agolubev:
added OutputStreamSource only so far. Tests show main cases and flush blocking. 
Let me know if implementation approach is Ok to proceed with InputStreamSink

### DIFF
--- a/akka-stream-tests/src/test/java/akka/stream/io/InputStreamSinkTest.java
+++ b/akka-stream-tests/src/test/java/akka/stream/io/InputStreamSinkTest.java
@@ -1,0 +1,48 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.io;
+
+import akka.japi.Pair;
+import akka.stream.StreamTest;
+import akka.stream.javadsl.AkkaJUnitActorSystemResource;
+import akka.stream.javadsl.Sink;
+import akka.stream.javadsl.Source;
+import akka.stream.testkit.AkkaSpec;
+import akka.stream.testkit.Utils;
+import akka.util.ByteString;
+import org.junit.ClassRule;
+import org.junit.Test;
+import scala.concurrent.Future;
+import scala.concurrent.duration.FiniteDuration;
+
+import java.io.InputStream;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.assertTrue;
+
+public class InputStreamSinkTest  extends StreamTest {
+    public InputStreamSinkTest() {
+        super(actorSystemResource);
+    }
+
+    @ClassRule
+    public static AkkaJUnitActorSystemResource actorSystemResource = new AkkaJUnitActorSystemResource("InputStreamSink",
+            Utils.UnboundedMailboxConfig());
+    @Test
+    public void mustReadEventViaInputStream() throws Exception {
+        final FiniteDuration timeout = FiniteDuration.create(300, TimeUnit.MILLISECONDS);
+
+        final Sink<ByteString, Pair<InputStream,Future<Long>>> sink = InputStreamSink.create(timeout);
+        final List<ByteString> list = Collections.singletonList(ByteString.fromString("a"));
+        final Pair<InputStream, Future<Long>> p = Source.from(list).runWith(sink, materializer);
+
+        byte[] a = new byte[1];
+        p.first().read(a);
+        assertTrue(Arrays.equals("a".getBytes(), a));
+    }
+
+}
--- a/akka-stream-tests/src/test/java/akka/stream/io/OutputStreamSourceTest.java
+++ b/akka-stream-tests/src/test/java/akka/stream/io/OutputStreamSourceTest.java
@@ -1,0 +1,56 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.io;
+
+import akka.actor.ActorRef;
+import akka.japi.Pair;
+import akka.japi.function.Procedure;
+import akka.stream.StreamTest;
+import akka.stream.javadsl.AkkaJUnitActorSystemResource;
+import akka.stream.javadsl.Keep;
+import akka.stream.javadsl.Sink;
+import akka.stream.javadsl.Source;
+import akka.stream.testkit.AkkaSpec;
+import akka.stream.testkit.Utils;
+import akka.testkit.JavaTestKit;
+import akka.util.ByteString;
+import com.typesafe.config.ConfigFactory;
+import org.junit.ClassRule;
+import org.junit.Test;
+import scala.concurrent.Future;
+import scala.concurrent.duration.FiniteDuration;
+import scala.runtime.BoxedUnit;
+
+import java.io.OutputStream;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.assertEquals;
+
+public class OutputStreamSourceTest extends StreamTest {
+    public OutputStreamSourceTest() {
+        super(actorSystemResource);
+    }
+
+    @ClassRule
+    public static AkkaJUnitActorSystemResource actorSystemResource = new AkkaJUnitActorSystemResource("OutputStreamSource",
+            Utils.UnboundedMailboxConfig());
+    @Test
+    public void mustSendEventsViaOutputStream() throws Exception {
+        final FiniteDuration timeout = FiniteDuration.create(300, TimeUnit.MILLISECONDS);
+        final JavaTestKit probe = new JavaTestKit(system);
+
+        final Source<ByteString, Pair<OutputStream, Future<Long>>> source = OutputStreamSource.create(timeout);
+        final Pair<OutputStream, Future<Long>> p = source.to(Sink.foreach(new Procedure<ByteString>() {
+            public void apply(ByteString elem) {
+                probe.getRef().tell(elem, ActorRef.noSender());
+            }
+        })).run(materializer);
+
+        p.first().write("a".getBytes());
+        assertEquals(ByteString.fromString("a"), probe.receiveOne(timeout));
+        p.first().close();
+
+    }
+
+}
--- a/akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
@@ -1,0 +1,204 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.io
+
+import java.io.{ IOException, InputStream }
+import java.util.concurrent.TimeoutException
+
+import akka.actor.{ Deploy, ActorSystem, Props }
+import akka.stream._
+import akka.stream.actor.ActorSubscriber.OnSubscribe
+import akka.stream.actor.ActorSubscriberMessage
+import akka.stream.actor.ActorSubscriberMessage.{ OnComplete, OnNext, OnError }
+import akka.stream.impl.Stages.DefaultAttributes
+import akka.stream.impl.StreamLayout.Module
+import akka.stream.impl.StreamSupervisor.Children
+import akka.stream.impl.io.{ InputStreamSink, InputStreamSubscriber }
+import akka.stream.impl.{ SinkModule, ActorMaterializerImpl, StreamSupervisor }
+import akka.stream.scaladsl.{ Keep, Sink }
+import akka.stream.testkit.AkkaSpec
+import akka.stream.testkit.Utils._
+import akka.stream.testkit.scaladsl.TestSource
+import akka.testkit.TestProbe
+import akka.util.ByteString
+
+import scala.concurrent.duration._
+import scala.concurrent.{ Await, Future, Promise }
+import scala.util.control.NoStackTrace
+
+class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
+  import system.dispatcher
+
+  val settings = ActorMaterializerSettings(system).withDispatcher("akka.actor.default-dispatcher")
+  implicit val materializer = ActorMaterializer(settings)
+
+  val timeout = 300.milliseconds
+  val byteString = ByteString("abc")
+  val byteArray = "abc".toArray
+  def newArray() = new Array[Byte](3)
+
+  def expectBlocked[T](f: Future[T]) =
+    the[Exception] thrownBy Await.result(f, timeout) shouldBe a[TimeoutException]
+
+  def expectSuccess[T](f: Future[T], value: T) =
+    Await.result(f, timeout) should be(value)
+
+  def testSink(probe: TestProbe): Sink[ByteString, (InputStream, Future[Long])] = {
+    class TestInputStreamSink(override val timeout: FiniteDuration,
+                              override val attributes: Attributes,
+                              shape: SinkShape[ByteString])
+      extends InputStreamSink(timeout, attributes, shape) {
+      override protected def getSubscriber(bytesWrittenPromise: Promise[Long], settings: ActorMaterializerSettings) =
+        Props(new TestInputStreamSubscriber(bytesWrittenPromise, settings.maxInputBufferSize)).withDeploy(Deploy.local)
+
+      override protected def newInstance(shape: SinkShape[ByteString]): SinkModule[ByteString, (InputStream, Future[Long])] =
+        new TestInputStreamSink(timeout, attributes, shape)
+    }
+
+    class TestInputStreamSubscriber(p: Promise[Long], bufSize: Int)
+      extends InputStreamSubscriber(p, bufSize) {
+      protected[akka] override def aroundReceive(receive: Receive, msg: Any): Unit = {
+        super.aroundReceive(receive, msg)
+        probe.ref ! msg
+      }
+    }
+    new Sink(new TestInputStreamSink(timeout, DefaultAttributes.inputStreamSink, Sink.shape("TestInputStreamSink")))
+  }
+
+  "InputStreamSink" must {
+    "read bytes from InputStream" in assertAllStagesStopped {
+      val (probe, (inputStream, f)) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+
+      probe.sendNext(byteString)
+      val arr = newArray()
+      inputStream.read(arr)
+      assert(arr === byteArray)
+
+      probe.sendComplete()
+      inputStream.close()
+      expectSuccess(f, 3)
+    }
+
+    "returns less than was expected when the data source has provided some but not enough data" in assertAllStagesStopped {
+      val (probe, (inputStream, f)) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+
+      probe.sendNext(ByteString("ab"))
+      val arr = newArray()
+      inputStream.read(arr)
+      assert(arr === "ab\u0000".toArray)
+
+      probe.sendComplete()
+      inputStream.close()
+      expectSuccess(f, 2)
+    }
+
+    "block read until get requested number of bytes from upstream" in assertAllStagesStopped {
+      val (probe, (inputStream, _)) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+
+      val arr = newArray()
+      val f = Future(inputStream.read(arr))
+
+      expectBlocked(f)
+      probe.sendNext(byteString)
+      expectSuccess(f, 3)
+
+      probe.sendComplete()
+      inputStream.close()
+    }
+
+    "fill up buffer by default" in assertAllStagesStopped {
+      import system.dispatcher
+      val (probe, (inputStream, _)) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+
+      probe.sendNext(byteString)
+      probe.sendNext(ByteString("def"))
+
+      val arr1 = newArray()
+      val arr2 = newArray()
+      val f1 = Future(inputStream.read(arr1))
+      val f2 = Future(inputStream.read(arr2))
+      Await.result(f1, timeout) should be(3)
+      Await.result(f2, timeout) should be(3)
+
+      assert(arr1 === byteArray)
+      assert(arr2 === "def".toArray)
+
+      probe.sendComplete()
+      inputStream.close()
+    }
+
+    "throw error when reactive stream is closed" in assertAllStagesStopped {
+      val (probe, (inputStream, _)) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+
+      probe.sendNext(byteString)
+
+      inputStream.close()
+      probe.expectCancellation()
+      the[Exception] thrownBy inputStream.read(newArray()) shouldBe a[IOException]
+    }
+
+    "return all data when upstream is completed" in assertAllStagesStopped {
+      val sinkProbe = TestProbe()
+      val (probe, (inputStream, _)) = TestSource.probe[ByteString].toMat(testSink(sinkProbe))(Keep.both).run()
+      sinkProbe.expectMsgClass(classOf[OnSubscribe])
+
+      probe.sendNext(ByteString("a"))
+      sinkProbe.expectMsgClass(classOf[OnNext])
+
+      probe.sendComplete()
+      sinkProbe.expectMsg(OnComplete)
+
+      val arr = newArray()
+      val f = Future(inputStream.read(arr))
+      expectSuccess(f, 1)
+      assert(arr === "a\u0000\u0000".toArray)
+    }
+
+    "return -1 when read after stream is completed" in assertAllStagesStopped {
+      val (probe, (inputStream, _)) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+
+      probe.sendNext(byteString)
+      val arr = newArray()
+      inputStream.read(arr)
+      assert(arr === byteArray)
+      probe.sendComplete()
+      Await.result(Future(inputStream.read(arr)), timeout) should ===(-1)
+
+      inputStream.close()
+    }
+
+    "return IOException when stream is failed" in assertAllStagesStopped {
+      val sinkProbe = TestProbe()
+      val (probe, (inputStream, _)) = TestSource.probe[ByteString].toMat(testSink(sinkProbe))(Keep.both).run()
+      val ex = new RuntimeException("Stream failed.") with NoStackTrace
+      sinkProbe.expectMsgClass(classOf[OnSubscribe])
+
+      probe.sendNext(byteString)
+      sinkProbe.expectMsgClass(classOf[OnNext])
+
+      val arr = newArray()
+      inputStream.read(arr)
+      sinkProbe.expectMsgClass(classOf[InputStreamSubscriber.Read])
+
+      probe.sendError(ex)
+      sinkProbe.expectMsgClass(classOf[OnError])
+      val p = Future(inputStream.read(arr))
+      p.onFailure { case e ⇒ assert(e.isInstanceOf[IOException] && e.getCause.equals(ex)); Unit }
+      p.onSuccess { case _ ⇒ fail() }
+    }
+
+    "use dedicated blocking-io-dispatcher by default" in assertAllStagesStopped {
+      val sys = ActorSystem("dispatcher-testing", UnboundedMailboxConfig)
+      val mat = ActorMaterializer()(sys)
+
+      try {
+        TestSource.probe[ByteString].runWith(InputStreamSink())(mat)
+        mat.asInstanceOf[ActorMaterializerImpl].supervisor.tell(StreamSupervisor.GetChildren, testActor)
+        val ref = expectMsgType[Children].children.find(_.path.toString contains "inputStreamSink").get
+        assertDispatcher(ref, "akka.stream.default-blocking-io-dispatcher")
+      } finally shutdown(sys)
+    }
+  }
+
+}
--- a/akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
@@ -4,18 +4,16 @@
 package akka.stream.io
 
 import java.io.{ IOException, InputStream }
-import java.util.concurrent.TimeoutException
+import java.util.concurrent.{ BlockingQueue, TimeoutException }
 
-import akka.actor.{ Deploy, ActorSystem, Props }
+import akka.actor.{ ActorSystem, Deploy, Props }
 import akka.stream._
 import akka.stream.actor.ActorSubscriber.OnSubscribe
-import akka.stream.actor.ActorSubscriberMessage
-import akka.stream.actor.ActorSubscriberMessage.{ OnComplete, OnNext, OnError }
+import akka.stream.actor.ActorSubscriberMessage.{ OnComplete, OnError, OnNext }
 import akka.stream.impl.Stages.DefaultAttributes
-import akka.stream.impl.StreamLayout.Module
 import akka.stream.impl.StreamSupervisor.Children
 import akka.stream.impl.io.{ InputStreamSink, InputStreamSubscriber }
-import akka.stream.impl.{ SinkModule, ActorMaterializerImpl, StreamSupervisor }
+import akka.stream.impl.{ ActorMaterializerImpl, SinkModule, StreamSupervisor }
 import akka.stream.scaladsl.{ Keep, Sink }
 import akka.stream.testkit.AkkaSpec
 import akka.stream.testkit.Utils._
@@ -49,15 +47,15 @@ class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
                               override val attributes: Attributes,
                               shape: SinkShape[ByteString])
       extends InputStreamSink(timeout, attributes, shape) {
-      override protected def getSubscriber(bytesWrittenPromise: Promise[Long], settings: ActorMaterializerSettings) =
-        Props(new TestInputStreamSubscriber(bytesWrittenPromise, settings.maxInputBufferSize)).withDeploy(Deploy.local)
+      override protected def getSubscriber(sharedBuffer: BlockingQueue[ByteString]) =
+        Props(new TestInputStreamSubscriber(sharedBuffer)).withDeploy(Deploy.local)
 
       override protected def newInstance(shape: SinkShape[ByteString]): SinkModule[ByteString, (InputStream, Future[Long])] =
         new TestInputStreamSink(timeout, attributes, shape)
     }
 
-    class TestInputStreamSubscriber(p: Promise[Long], bufSize: Int)
-      extends InputStreamSubscriber(p, bufSize) {
+    class TestInputStreamSubscriber(sharedBuffer: BlockingQueue[ByteString])
+      extends InputStreamSubscriber(sharedBuffer) {
       protected[akka] override def aroundReceive(receive: Receive, msg: Any): Unit = {
         super.aroundReceive(receive, msg)
         probe.ref ! msg
@@ -73,11 +71,28 @@ class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
       probe.sendNext(byteString)
       val arr = newArray()
       inputStream.read(arr)
-      assert(arr === byteArray)
+      arr should ===(byteArray)
 
       probe.sendComplete()
       inputStream.close()
       expectSuccess(f, 3)
+    }
+
+    "read bytes correctly if requested by InputStream not in chunk size" in assertAllStagesStopped {
+      val (probe, (inputStream, f)) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+
+      probe.sendNext(byteString)
+      probe.sendNext(ByteString("def"))
+      val arr = new Array[Byte](2)
+      inputStream.read(arr)
+      arr should ===("ab".toArray)
+      inputStream.read(arr)
+      arr should ===("cd".toArray)
+      inputStream.read(arr)
+      arr should ===("ef".toArray)
+
+      inputStream.close()
+      expectSuccess(f, 6)
     }
 
     "returns less than was expected when the data source has provided some but not enough data" in assertAllStagesStopped {
@@ -86,7 +101,7 @@ class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
       probe.sendNext(ByteString("ab"))
       val arr = newArray()
       inputStream.read(arr)
-      assert(arr === "ab\u0000".toArray)
+      arr should ===("ab\u0000".toArray)
 
       probe.sendComplete()
       inputStream.close()
@@ -94,7 +109,7 @@ class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
     }
 
     "block read until get requested number of bytes from upstream" in assertAllStagesStopped {
-      val (probe, (inputStream, _)) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+      val (probe, (inputStream, future)) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
 
       val arr = newArray()
       val f = Future(inputStream.read(arr))
@@ -104,6 +119,9 @@ class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
       expectSuccess(f, 3)
 
       probe.sendComplete()
+      inputStream.read(newArray())
+      //complete stream future before input stream is closed
+      expectSuccess(future, 3)
       inputStream.close()
     }
 
@@ -121,8 +139,8 @@ class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
       Await.result(f1, timeout) should be(3)
       Await.result(f2, timeout) should be(3)
 
-      assert(arr1 === byteArray)
-      assert(arr2 === "def".toArray)
+      arr1 should ===(byteArray)
+      arr2 should ===("def".toArray)
 
       probe.sendComplete()
       inputStream.close()
@@ -152,7 +170,7 @@ class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
       val arr = newArray()
       val f = Future(inputStream.read(arr))
       expectSuccess(f, 1)
-      assert(arr === "a\u0000\u0000".toArray)
+      arr should ===("a\u0000\u0000".toArray)
     }
 
     "return -1 when read after stream is completed" in assertAllStagesStopped {
@@ -161,8 +179,9 @@ class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
       probe.sendNext(byteString)
       val arr = newArray()
       inputStream.read(arr)
-      assert(arr === byteArray)
+      arr should ===(byteArray)
       probe.sendComplete()
+
       Await.result(Future(inputStream.read(arr)), timeout) should ===(-1)
 
       inputStream.close()
@@ -179,12 +198,15 @@ class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
 
       val arr = newArray()
       inputStream.read(arr)
-      sinkProbe.expectMsgClass(classOf[InputStreamSubscriber.Read])
+      sinkProbe.expectMsg(InputStreamSubscriber.ReadNotification)
 
       probe.sendError(ex)
       sinkProbe.expectMsgClass(classOf[OnError])
       val p = Future(inputStream.read(arr))
-      p.onFailure { case e ⇒ assert(e.isInstanceOf[IOException] && e.getCause.equals(ex)); Unit }
+      p.onFailure {
+        case e ⇒
+          (e.isInstanceOf[IOException] && e.getCause.equals(ex)) should ===(true); Unit
+      }
       p.onSuccess { case _ ⇒ fail() }
     }
 
@@ -198,6 +220,27 @@ class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
         val ref = expectMsgType[Children].children.find(_.path.toString contains "inputStreamSink").get
         assertDispatcher(ref, "akka.stream.default-blocking-io-dispatcher")
       } finally shutdown(sys)
+    }
+  }
+  "InputStreamAdapter" must {
+    "send data request then read from shared buffer, then send read notification for back pressure" in assertAllStagesStopped {
+      val sinkProbe = TestProbe()
+      val (probe, (inputStream, _)) = TestSource.probe[ByteString].toMat(testSink(sinkProbe))(Keep.both).run()
+
+      sinkProbe.expectMsgClass(classOf[OnSubscribe])
+
+      val arr = newArray()
+      val p = Future(inputStream.read(arr))
+
+      sinkProbe.expectMsg(InputStreamSubscriber.Request)
+      probe.sendNext(byteString)
+      sinkProbe.expectMsgClass(classOf[OnNext])
+      Await.result(p, timeout) should ===(3)
+      arr should ===(byteArray)
+      sinkProbe.expectMsg(InputStreamSubscriber.ReadNotification)
+
+      probe.sendComplete()
+      inputStream.close()
     }
   }
 
--- a/akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
@@ -1,0 +1,173 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.io
+
+import java.io.{ IOException, OutputStream }
+import java.util.concurrent.TimeoutException
+
+import akka.actor.{ Deploy, ActorSystem, Props }
+import akka.stream._
+import akka.stream.actor.ActorPublisher.Internal.Subscribe
+import akka.stream.actor.ActorPublisherMessage.{ Cancel, Request }
+import akka.stream.impl.Stages.DefaultAttributes
+import akka.stream.impl.StreamSupervisor.Children
+import akka.stream.impl.io.{ OutputStreamPublisher, OutputStreamSource }
+import akka.stream.impl.{ ActorMaterializerImpl, StreamSupervisor }
+import akka.stream.scaladsl.{ Keep, Source }
+import akka.stream.testkit.Utils._
+import akka.stream.testkit._
+import akka.stream.testkit.scaladsl.TestSink
+import akka.testkit.TestProbe
+import akka.util.ByteString
+
+import scala.concurrent.duration.Duration.Zero
+import scala.concurrent.duration._
+import scala.concurrent.{ Await, Future, Promise }
+
+class OutputStreamSourceSpec extends AkkaSpec(UnboundedMailboxConfig) {
+  import system.dispatcher
+
+  val settings = ActorMaterializerSettings(system).withDispatcher("akka.actor.default-dispatcher")
+  implicit val materializer = ActorMaterializer(settings)
+
+  val timeout = 300.milliseconds
+  val bytes = "abc".getBytes
+  val byteString = ByteString("abc")
+
+  def expectBlocked[T](f: Future[T]) =
+    the[Exception] thrownBy Await.result(f, timeout) shouldBe a[TimeoutException]
+
+  def expectSuccess[T](f: Future[T], value: T) =
+    Await.result(f, timeout) should be(value)
+
+  def testSource(probe: TestProbe): Source[ByteString, (OutputStream, Future[Long])] = {
+    class TestOutputStreamSource(override val timeout: FiniteDuration,
+                                 override val attributes: Attributes,
+                                 shape: SourceShape[ByteString])
+      extends OutputStreamSource(timeout, attributes, shape) {
+      override def getPublisher(p: Promise[Long], settings: ActorMaterializerSettings) =
+        Props(new TestOutputStreamPublisher(p, settings.maxInputBufferSize)).withDeploy(Deploy.local)
+    }
+
+    class TestOutputStreamPublisher(bytesReadPromise: Promise[Long], bufSize: Int)
+      extends OutputStreamPublisher(bytesReadPromise, bufSize) {
+      protected[akka] override def aroundReceive(receive: Receive, msg: Any): Unit = {
+        super.aroundReceive(receive, msg)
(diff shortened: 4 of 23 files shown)


ktoso:
OK TO TEST

agolubev:
@ktoso any feedback on this? or it's not clear and I need to fill gaps?

ktoso:
Sorry, under a lot of pressure recently. I'll give it a read now

agolubev:
Thanks. Still have doubts about implementation approach (with introducing light actor). Now I'm more for just having Adapter and Published as one class with additional synchronization inside. Let me know your thoughts

agolubev:
Ok, committed both OutputStreamSource and InputStreamSink. Still, Java API is not properly done.

agolubev:
hm, seems like failed for TCP test

agolubev:
@ktoso please take a look when you have sometime.

akka-stream/src/main/scala/akka/stream/impl/io/IOSources.scala
```diff
@@ -73,3 +75,66 @@ private[akka] final class InputStreamSource(createInputStream: () ⇒ InputStrea
...
+
+  private[this] def sendMessage(message: OutputStreamPublisher.OutputStreamPublisherInMessage) =
+    try {
+      Await.result(publisher.ask(message)(Timeout(timeout)), timeout) match {
+        case OutputStreamPublisher.Failure(msg) ⇒ throw new IOException(msg)
+        case OutputStreamPublisher.Ok           ⇒ ()
+      }
+    } catch {
+      case e: IOException ⇒ throw e
+      case e: Exception   ⇒ throw new IOException(e)
```
ktoso: `NonFatal`?

akka-stream/src/main/scala/akka/stream/impl/io/IOSources.scala
```diff
@@ -73,3 +75,66 @@ private[akka] final class InputStreamSource(createInputStream: () ⇒ InputStrea
...
+
+  override def withAttributes(attr: Attributes): Module =
+    new OutputStreamSource(timeout, attr, amendShape(attr))
+}
+
+/**
+ * INTERNAL API
+ * OutputStreamAdapter that interacts with OutputStreamPublisher
+ */
+sealed class OutputStreamAdapter(val publisher: ActorRef, val timeout: FiniteDuration) extends OutputStream {
```
ktoso: also known as `final` :)

akka-stream/src/main/scala/akka/stream/impl/io/IOSources.scala
```diff
@@ -73,3 +75,66 @@ private[akka] final class InputStreamSource(createInputStream: () ⇒ InputStrea
...
+
+  import akka.pattern.ask
+
+  private[this] def sendMessage(message: OutputStreamPublisher.OutputStreamPublisherInMessage) =
+    try {
+      Await.result(publisher.ask(message)(Timeout(timeout)), timeout) match {
+        case OutputStreamPublisher.Failure(msg) ⇒ throw new IOException(msg)
+        case OutputStreamPublisher.Ok           ⇒ ()
+      }
```
ktoso: Yeah, this might work good enough actually; digging into the actors now

akka-stream/src/main/scala/akka/stream/io/OutputStreamSource.scala
```diff
@@ -0,0 +1,43 @@
...
+  def apply(timeout: FiniteDuration = 5.seconds): Source[ByteString, (OutputStream, Future[Long])] = {
+    new Source(new OutputStreamSource(timeout, DefaultAttributes, Source.shape("OutputStreamSource")))
+  }
+  /**
+   * Creates a synchronous (Java 6 compatible) Source.
+   *
+   * It materializes a [[Future]] containing the number of bytes written to OutputStream upon completion and
+   * [[OutputStream]] to interacting with reactive stream.
+   */
+  def create(): javadsl.Source[ByteString, Pair[OutputStream, Future[java.lang.Long]]] =
```
ktoso: java api needs corresponding `create(FiniteDuration)` call

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamPublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+
+/** INTERNAL API */
+private[akka] object OutputStreamPublisher {
+  def props(bytesReadPromise: Promise[Long], bufSize: Int) = {
+    require(bufSize > 0, "buffer size must be > 0")
+    Props(classOf[OutputStreamPublisher], bytesReadPromise, bufSize)
+      .withDeploy(Deploy.local)
+  }
+
+  trait OutputStreamPublisherInMessage
```
ktoso: sealed :)

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamPublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  case class Write(bs: ByteString) extends OutputStreamPublisherInMessage
+
+  case class Flush() extends OutputStreamPublisherInMessage
+
+  case class Close() extends OutputStreamPublisherInMessage
+
+  trait OutputStreamPublisherOutMessage
+
+  case class Ok() extends OutputStreamPublisherOutMessage
```
ktoso: case object

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamPublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  trait OutputStreamPublisherInMessage
+
+  case class Write(bs: ByteString) extends OutputStreamPublisherInMessage
+
+  case class Flush() extends OutputStreamPublisherInMessage
+
+  case class Close() extends OutputStreamPublisherInMessage
+
+  trait OutputStreamPublisherOutMessage
```
ktoso: sealed

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamPublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+  def receive = {
+    //Downstream messages
+    case ActorPublisherMessage.Request(elements) ⇒
+      signalDownstream()
+    case ActorPublisherMessage.Cancel ⇒
+      trySignalOnSend(signalErrorUpstream(_, "Stream is cancelled"))
+      context.stop(self)
+
+    //Upstream messages from OutputStreamAdapter
+    case OutputStreamPublisher.Write(bs) ⇒ onWrite(bs) //TODO what if it's been cancelled? need test
```
ktoso: `    case OutputStreamPublisher.Write(bs) if isActive` should be enough here (for all of these)

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamPublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  override def postStop(): Unit = {
+    super.postStop()
+    bytesReadPromise.trySuccess(bytesGotTotal)
+  }
+
+  def isBufferFull = availableChunks.size == bufSize
+
+  def onWrite(byteString: ByteString) = {
+    if (!isCompleted && !isCanceled && !isErrorEmitted) {
```
ktoso: AFAIR `isActive` is enough, check its impl please

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamPublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+      availableChunks :+= byteString
+
+      if (isBufferFull)
+        notifyWhenSend = Some(sender)
+      else
+        sender ! OutputStreamPublisher.Ok
```
ktoso: seems ok

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamPublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+      notifyWhenSendAll match {
+        case Some(sender) ⇒
+          signalOkUpstream(sender)
+          notifyWhenSendAll = None
+        case None ⇒
+      }
+  }
+  /*
+  Considering OutputStreamAdapter as upstream
+   */
```
ktoso: weird doc style, for impl ones like these `//` isok

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamPublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+
+    //Upstream messages from OutputStreamAdapter
+    case OutputStreamPublisher.Write(bs) ⇒ onWrite(bs) //TODO what if it's been cancelled? need test
+    case OutputStreamPublisher.Flush()   ⇒ flush()
+    case OutputStreamPublisher.Close()   ⇒ flush(); onCompleteThenStop()
+  }
+
+  def signalDownstream(): Unit =
+    if (isActive) {
+      while (totalDemand > 0 && availableChunks.nonEmpty) {
```
ktoso: Here that's OK, since availableChunks won't be huge. In general hint that we had to optimise those to send in "smaller bursts" sometimes. But here it's not a problem I think, keep as is.

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamPublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+        val ready = availableChunks.head
+        availableChunks = availableChunks.tail
+        bytesGotTotal += ready.size
```
ktoso: so it's "bytesSentDownstream" and not "got" :)

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamPublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+      }
+      trySignalOnSendAll()
+    }
+
+  def trySignalOnSend(notify: (ActorRef) ⇒ Unit = signalOkUpstream(_)) = {
+    notifyWhenSend match {
+      case Some(sender) ⇒
+        notify(sender)
+        notifyWhenSend = None
+      case None ⇒
```
ktoso: add `// do nothing...` so we know it's intentional

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamPublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+        val ready = availableChunks.head
+        availableChunks = availableChunks.tail
+        bytesGotTotal += ready.size
+        onNext(ready)
+        trySignalOnSend()
+      }
+      trySignalOnSendAll()
+    }
+
+  def trySignalOnSend(notify: (ActorRef) ⇒ Unit = signalOkUpstream(_)) = {
```
ktoso: I understand what it does, makes sense; maybe the name could be better, it's basically "ack write" or "confirm write to upstream"?

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamPublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+  /*
+  Considering OutputStreamAdapter as upstream
+   */
+  def signalOkUpstream(sender: ActorRef) = sender ! OutputStreamPublisher.Ok
+
+  def signalErrorUpstream(actor: ActorRef, msg: String) = actor ! OutputStreamPublisher.Failure(msg)
+
+  override def postStop(): Unit = {
+    super.postStop()
+    bytesReadPromise.trySuccess(bytesGotTotal)
```
ktoso: good

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamPublisher.scala
```diff
@@ -0,0 +1,125 @@
...
+        sender ! OutputStreamPublisher.Ok
+
+      signalDownstream()
+    } else {
+      signalErrorUpstream(sender, "Stream is terminated, no writes are possible")
+    }
+  }
+
+  def flush() = {
+    notifyWhenSendAll = Some(sender)
```
ktoso: difference between `notifyWhenSendAll` and `notifyWhenSend` is a bit fuzzy, rename them or explain in impl docs?

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala
```diff
@@ -0,0 +1,119 @@
...
+
+  var bytesRequestedDownstream: Option[(ActorRef, Int)] = None
+
+  var bytesWritten: Long = 0
+
+  var availableChunks: Vector[ByteString] = Vector.empty[ByteString]
+
+  override protected val requestStrategy = new RequestStrategy() {
+    override def requestDemand(remainingRequested: Int): Int = bytesRequestedDownstream match {
+      case Some((actor, length)) ⇒ if (remainingRequested > 0) 0 else 1
```
ktoso: random trivia, until recently scala allocates a lot of tuples for these extractors (2.12 improves on that AFAIR).
Not that it hurts much here, since we're IO dominated anyway

ktoso: On the other hand, since actor nor length are used let's just match `_: SOme`

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala
```diff
@@ -0,0 +1,119 @@
...
+      case None                  ⇒ bufSize - availableChunks.size - remainingRequested
+    }
+  }
+
+  def receive = {
+    case ActorSubscriberMessage.OnNext(bytes: ByteString) ⇒
+      availableChunks :+= bytes
```
ktoso: maybe should be `isActive` guarded?

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+import org.scalatest.concurrent.ScalaFutures
+
+import scala.concurrent.duration._
+import scala.concurrent.{ Await, Future }
+
+class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) with ScalaFutures {
+
+  val settings = ActorMaterializerSettings(system).withDispatcher("akka.actor.default-dispatcher")
+  implicit val materializer = ActorMaterializer(settings)
+  val timeout = 300.milliseconds
```
ktoso: good timeout for jekins, :+1:

akka-stream/src/main/scala/akka/stream/io/OutputStreamSource.scala
```diff
@@ -0,0 +1,43 @@
...
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+/**
+ * Source which allows to use [[OutputStream]] to interact with reactive stream.
+ */
+object OutputStreamSource {
+  import scala.language.postfixOps
+
+  final val DefaultChunkSize = 8192
```
ktoso: do other sources/sinks here also use hardcoded chunk size?

akka-stream/src/main/scala/akka/stream/io/OutputStreamSource.scala
```diff
@@ -0,0 +1,43 @@
...
+    new Source(new OutputStreamSource(timeout, DefaultAttributes, Source.shape("OutputStreamSource")))
+  }
+  /**
+   * Creates a synchronous (Java 6 compatible) Source.
+   *
+   * It materializes a [[Future]] containing the number of bytes written to OutputStream upon completion and
+   * [[OutputStream]] to interacting with reactive stream.
+   */
+  def create(): javadsl.Source[ByteString, Pair[OutputStream, Future[java.lang.Long]]] =
+    apply().asJava.asInstanceOf[javadsl.Source[ByteString, Pair[OutputStream, Future[java.lang.Long]]]]
```
ktoso: Can you add a test it really works with `j.l.Long`? There's been weird cases when conversion blew up... forgot the exact cases though, better to cover with tests

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+  val byteString = ByteString("abc")
+
+  def expectBlocked[T](f: Future[T]) =
+    the[Exception] thrownBy Await.result(f, timeout) shouldBe a[TimeoutException]
```
ktoso: nice :)

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+      val f1 = Future(inputStream.read(arr1))
+      val f2 = Future(inputStream.read(arr2))
+      Await.result(f1, timeout) should be(3)
+      Await.result(f2, timeout) should be(3)
+
+      assert(arr1 === "abc".toArray)
+      assert(arr2 === "def".toArray)
+
+      probe.sendComplete()
+    }
```
ktoso: can we expect on the completion here somehow? i.e. further reads do what?

ktoso: (it's implicitly checked by the `assertAllStagesStopped` I guess, but still)

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+    "throw error when reactive stream is closed" in assertAllStagesStopped {
+      val (probe, (inputStream, _)) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+
+      probe.sendNext(byteString)
+      inputStream.close()
+      probe.expectCancellation()
+
+      val arr = new Array[Byte](3)
+      //this is the cause of dead letter
+      the[Exception] thrownBy inputStream.read(arr) shouldBe a[IOException]
```
ktoso: nice

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    "unblock read when upstream is completed" in assertAllStagesStopped {
+      import scala.concurrent.ExecutionContext.Implicits.global
```
ktoso: Seems we could `import system.dispatcher`?

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+
+      val (probe, (inputStream, _)) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+      probe.sendNext(ByteString("a"))
+      val arr = new Array[Byte](3)
+      val f = Future(inputStream.read(arr))
+
+      expectBlocked(f)
+      probe.sendComplete()
+      expectSuccess(f, 1)
+      assert(arr === "a\u0000\u0000".toArray)
```
ktoso: :-)

akka-stream/src/main/scala/akka/stream/io/OutputStreamSource.scala
agolubev: will remove it as I'm not doing chunks anymore. assuming that there is some contract about ByteString size.

akka-stream/src/main/scala/akka/stream/io/OutputStreamSource.scala
agolubev: the problem here is having scala pair. it must be japi.Pair

agolubev: but it requires creating class for Source with pair as Mat

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
agolubev: last read if it's in blocked state will return everything that is in buffer (assuming less then client requested). Then IO exception will be thrown. Will add this test if it's not yet there

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
```diff
@@ -0,0 +1,120 @@
...
+
+      outputStream.write(bytes)
+      s.request(1)
+      probe.expectNext(byteString)
+      outputStream.close()
+      probe.expectComplete()
+      Await.result(f, timeout) should be(3)
+    }
+
+    "flush call blocks until sent to downstream" in assertAllStagesStopped {
```
ktoso: Yes, I guess that's the expected semantics when dealing with outputstream

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
```diff
@@ -66,3 +68,72 @@ private[akka] final class OutputStreamSink(createOutput: () ⇒ OutputStream, va
...
+
+  override def withAttributes(attr: Attributes): Module =
+    new InputStreamSink(timeout, attr, amendShape(attr))
+}
+
+/**
+ * INTERNAL API
+ * InputStreamAdapter that interacts with InputStreamSubscriber
+ */
+sealed private class InputStreamAdapter(subscriber: ActorRef, val timeout: FiniteDuration) extends InputStream {
```
ktoso: final?

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
```diff
@@ -66,3 +68,72 @@ private[akka] final class OutputStreamSink(createOutput: () ⇒ OutputStream, va
...
+
+    val props = InputStreamSubscriber.props(bytesWrittenPromise, settings.maxInputBufferSize)
+
+    val ref = mat.actorOf(context, props)
+    (akka.stream.actor.ActorSubscriber[ByteString](ref),
+      (new InputStreamAdapter(ref, timeout).asInstanceOf[InputStream], bytesWrittenPromise.future))
+  }
+
+  override protected def newInstance(shape: SinkShape[ByteString]): SinkModule[ByteString, (InputStream, Future[Long])] =
+
```
ktoso: remove this additional newline

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
```diff
@@ -66,3 +68,72 @@ private[akka] final class OutputStreamSink(createOutput: () ⇒ OutputStream, va
...
+  private[this] def sendMessage(message: InputStreamSubscriber.InputStreamSubscriberInMessage): ByteString =
+    try {
+      Await.result(subscriber.ask(message)(Timeout(timeout)), timeout) match {
+        case InputStreamSubscriber.Failure(msg) ⇒ throw new IOException(msg)
+        case InputStreamSubscriber.Bytes(b)     ⇒ b
+      }
+    } catch {
+      case e: IOException ⇒ throw e
+      case e: Exception   ⇒ throw new IOException(e)
```
ktoso: `NonFatal`

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
```diff
@@ -66,3 +68,72 @@ private[akka] final class OutputStreamSink(createOutput: () ⇒ OutputStream, va
...
+
+  @scala.throws(classOf[IOException])
+  override def read(a: Array[Byte], begin: Int, length: Int): Int = {
+    val bytes = sendMessage(InputStreamSubscriber.Read(length)).toArray
+    System.arraycopy(bytes, 0, a, begin, bytes.length)
+    bytes.length
+  }
+
+  @scala.throws(classOf[IOException])
+  override def close(): Unit = subscriber.tell(InputStreamSubscriber.Close(), ActorRef.noSender)
```
ktoso: maybe also via `sendMessage`? What if close fails, I guess I'd like to know when using this stream

akka-stream/src/main/scala/akka/stream/impl/io/IOSources.scala
```diff
@@ -73,3 +75,66 @@ private[akka] final class InputStreamSource(createInputStream: () ⇒ InputStrea
...
+
+  override def create(context: MaterializationContext) = {
+    val mat = ActorMaterializer.downcast(context.materializer)
+
+    val bytesReadPromise = Promise[Long]()
+    val settings = mat.effectiveSettings(context.effectiveAttributes)
+
+    val maxInputBufferSize = settings.maxInputBufferSize
+
+    val props = OutputStreamPublisher.props(bytesReadPromise, maxInputBufferSize).withDeploy(Deploy.local)
```
ktoso: I was tempted to propose the IO dispatcher to be used for these actors... Technically it's not "file io" it could be other stuff, but it's blocking... I'd want to put anything blocking far away from the default dispatcher.

Opinions @viktorklang @jrudolph ?

akka-stream/src/main/scala/akka/stream/io/OutputStreamSource.scala
ktoso: btw, this is unused :)

akka-stream/src/main/scala/akka/stream/io/InputStreamSink.scala
```diff
@@ -0,0 +1,43 @@
...
+    new Sink(new InputStreamSink(timeout, DefaultAttributes, Sink.shape("InputStreamSink")))
+  }
+
+  /**
+   * Creates a synchronous (Java 6 compatible) Sink
+   *
+   * It materializes a [[Future]] containing the number of bytes written to InputStream upon completion and
+   * [[InputStream]] to interacting with reactive stream.
+   */
+  def create(): javadsl.Sink[ByteString, Pair[InputStream, Future[java.lang.Long]]] =
```
ktoso: needs corresponding Java API for `create(chunksize, timeout)`

ktoso:
Sorry it took me so long to get to review this @agolubev! 
I think it looks great actually, needs some minor polishing but I'd pull it in soon actually :)

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
agolubev: in Subscriber we are doing only context.stop(self) in cancel method of subscriber. and after this actor can not send anything back I believe.

agolubev:
Thanks for the feedback!

akka-stream/src/main/scala/akka/stream/impl/io/IOSources.scala
viktorklang: Shouldn't it use the dispatcher configured for the stage?

ktoso: That's actually true as well; @agolubev, have you seen how the FileSource deals with dispatchers?

agolubev: Yeah ```val dispatcher = IOSettings.fileIoDispatcher(context)``` with config in "akka.stream.file-io-dispatcher".
So going to introduce "akka.stream.blocking-io-dispatcher" configuration. WDT?

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala
agolubev: I checked, subscriber does not have isActive, only publishers have

ktoso: Ah, right - thanks for checking

jrudolph:
Error lines in https://jenkins.akka.io/job/pr-validator-per-commit-jenkins/3277/:

```
 14105 Failed: Total 234, Failed 1, Errors 0, Passed 233, Pending 6
 14106 Failed tests:
 14107 	docs.stream.cookbook.RecipeDroppyBroadcast
 14126 (akka-docs-dev/test:test) sbt.TestsFailedException: Tests unsuccessful
 14127 Total time: 866 s, completed Jul 23, 2015 6:58:42 PM
```

agolubev:
it's already been fixed in 2.3-dev. Will rebase my branch.

akka-stream/src/main/resources/reference.conf
```diff
@@ -47,6 +47,9 @@ akka {
     # such as FileSource, FileSink and others.
     file-io-dispatcher = "akka.stream.default-file-io-dispatcher"
 
+    # Blocking IO like OutputStreamSource and InputStreamSink uses by default the same settings
+    blocking-io-dispatcher = "akka.stream.default-file-io-dispatcher"
```
viktorklang: I'm not fully on board with why this is desirable.

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
```diff
@@ -66,3 +72,94 @@ private[akka] final class OutputStreamSink(createOutput: () ⇒ OutputStream, va
   override def withAttributes(attr: Attributes): Module =
     new OutputStreamSink(createOutput, attr, amendShape(attr))
 }
+
+/**
+ * INTERNAL API
+ * Creates simple synchronous (Java 6 compatible) Sink which provide all incoming element on read request via
+ * InputStream.
+ */
+private[akka] sealed class InputStreamSink(timeout: FiniteDuration, val attributes: Attributes, shape: SinkShape[ByteString])
```
viktorklang: `final` iso `sealed`?

ktoso: previously commented already that `final` would be nicer here :-)
btw, not really `iso` "because java" :wink:

viktorklang: I don't understand, nothing in this PR extends InputStreamSink?
Wdym "because java"?

akka-stream/src/main/resources/reference.conf
ktoso: +1, let's just make sure it uses the configured (by `withAttributes`) dispatcher, and no default then?

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
ktoso: > Wdym "because java"?

just that in javac ignores sealed-ness of things.
Agree on `final` being better here :)

agolubev: it's not final as we have InputStreamSinkJava (maybe another concern here). Added this to be able to have japi.Pair[InputStream, Future[JLong] for Java API

akka-stream/src/main/resources/reference.conf
viktorklang: Consistency is dog. Where are other things running by default? :)

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
viktorklang: InputStreamSinkJava does not extend InputStreamSink, so I'm not sure I follow why it cannot be `final`. Am I missing something?

akka-stream/src/main/resources/reference.conf
ktoso: Historically only "file io" had a default blocking dispatcher (both in IO and in http (and spray previouwly))

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
agolubev: You are right. My bad

akka-stream/src/main/resources/reference.conf
viktorklang: I'm fine as long as it is dog :)

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
viktorklang: Thanks for confirming, I was fearing for my sanity for a while there.

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
```diff
@@ -66,3 +72,94 @@ private[akka] final class OutputStreamSink(createOutput: () ⇒ OutputStream, va
...
+
+  override def withAttributes(attr: Attributes): Module =
+    new InputStreamSink(timeout, attr, amendShape(attr))
+}
+
+/**
+ * INTERNAL API
+ * For Java API usages (returns Java API Pair)
+ */
+private[akka] final class InputStreamSinkJava(timeout: FiniteDuration, val attributes: Attributes, shape: SinkShape[ByteString])
```
viktorklang: Small change but I believe the tradition is to put "Java" first when it comes to naming classes intended for Java.

agolubev: Will change this. Thanks.

akka-stream/src/main/resources/reference.conf
agolubev: @ktoso What's the discussion summary? Should I left dispatcher as it is now?

agolubev:
Just pushed whatever was reviewed in a second try (except blocking-io-dispatcher)

agolubev:
Ref #18002 - build fail is a separate issue

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
```diff
@@ -66,3 +72,94 @@ private[akka] final class OutputStreamSink(createOutput: () ⇒ OutputStream, va
...
+ * INTERNAL API
+ * InputStreamAdapter that interacts with InputStreamSubscriber
+ */
+private[akka] final class InputStreamAdapter(subscriber: ActorRef, val timeout: FiniteDuration) extends InputStream {
+
+  import akka.pattern.ask
+
+  @scala.throws(classOf[IOException])
+  override def read(): Int = {
+    val a = new Array[Byte](1)
```
viktorklang: No need for allocating this array:

~~~scala
sendMessage(InputStreamSubscriber.Read(1))(0)
~~~

agolubev: Make sense

agolubev:
@ktoso I fixed all feedback points so please review when you have time. I also can rebase to eliminate test failed issue.

akka-stream-tests/src/test/java/akka/stream/io/OutputStreamSourceTest.java
```diff
@@ -0,0 +1,56 @@
...
+    public void mustSendEventsViaOutputStream() throws Exception {
+        final FiniteDuration timeout = FiniteDuration.create(300, TimeUnit.MILLISECONDS);
+        final JavaTestKit probe = new JavaTestKit(system);
+
+        final Source<ByteString, Pair<OutputStream, Future<Long>>> source = OutputStreamSource.create(timeout);
+        final Pair<OutputStream, Future<Long>> p = source.toMat(Sink.foreach(new Procedure<ByteString>() {
+            public void apply(ByteString elem) {
+                probe.getRef().tell(elem, ActorRef.noSender());
+            }
+        }), Keep.<Pair<OutputStream, Future<Long>>,Future<BoxedUnit>> left()).run(materializer);
```
rkuhn: `Keep.left` is the default, this should be simplified to `source.to(...).run(...)`.

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
```diff
@@ -0,0 +1,145 @@
...
+    Await.result(f, timeout) should be(value)
+
+  "InputStreamSink" must {
+    "read bytes from InputStream" in assertAllStagesStopped {
+      val (probe, (inputStream, f)) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+
+      probe.sendNext(byteString)
+      val arr = new Array[Byte](3)
+      inputStream.read(arr)
+      assert(arr === "abc".toArray)
```
rkuhn: perhaps factor out `val byteArray` and `def newArray` just like `val byteString`

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
```diff
@@ -0,0 +1,145 @@
...
+    "throw IOException when read after stream is completed" in assertAllStagesStopped {
+      val (probe, (inputStream, f)) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+
+      probe.sendNext(byteString)
+      val arr = new Array[Byte](3)
+      inputStream.read(arr)
+
+      probe.sendComplete()
+      //this is the cause of dead letter
+      the[Exception] thrownBy Await.result(Future(inputStream.read(arr)), timeout) shouldBe a[IOException]
```
rkuhn: This is not correct: InputStream.read() is supposed to return -1 upon having reached EOF.

rkuhn: there should be another test case that fails the stream, in which case the IOException is the right behavior

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
```diff
@@ -0,0 +1,130 @@
...
+    Await.result(f, timeout) should be(value)
+
+  "OutputStreamSource" must {
+    "read bytes from OutputStream" in assertAllStagesStopped {
+      val ((outputStream, f), probe) = OutputStreamSource().toMat(TestSink.probe[ByteString])(Keep.both).run
+      val s = probe.expectSubscription()
+
+      outputStream.write(bytes)
+      s.request(1)
+      probe.expectNext(byteString)
```
rkuhn: please assert that these are the correct three bytes

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    "block flush call until send all buffer to downstream" in assertAllStagesStopped {
+      val ((outputStream, _), probe) = OutputStreamSource().toMat(TestSink.probe[ByteString])(Keep.both).run
+      val s = probe.expectSubscription()
+
+      outputStream.write(bytes)
+      val f = Future(outputStream.flush())
+
+      probe.expectNoMsg(timeout)
```
rkuhn: please move this down one line and set the duration to zero (because `expectBlocked` already waits long enough)

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
```diff
@@ -0,0 +1,130 @@
...
+    "not block flushes when buffer is empty" in assertAllStagesStopped {
+      val ((outputStream, _), probe) = OutputStreamSource().toMat(TestSink.probe[ByteString])(Keep.both).run
+      val s = probe.expectSubscription()
+
+      outputStream.write(bytes)
+
+      val f = Future(outputStream.flush())
+      s.request(1)
+      expectSuccess(f, ())
+      probe.expectNext(byteString)
```
rkuhn: check bytes

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
rkuhn: (same below)

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
```diff
@@ -0,0 +1,130 @@
...
+
+    "throw error when reactive stream is closed" in assertAllStagesStopped {
+      val ((outputStream, _), probe) = OutputStreamSource().toMat(TestSink.probe[ByteString])(Keep.both).run
+
+      probe.expectSubscription()
+      outputStream.close()
+      probe.expectComplete()
+      //source of dead letters
+      the[Exception] thrownBy outputStream.write(bytes) shouldBe a[IOException]
+    }
```
rkuhn: there’s a test case missing: what about writing to the stream after the subscriber has cancelled the reactive stream?

akka-stream/src/main/resources/reference.conf
rkuhn: My own slightly presumptuous summary is that we should rename the `file-io-dispatcher` to `blocking-io-dispatcher` and use it for this as well (i.e. “what @ktoso said”).

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
```diff
@@ -66,3 +72,90 @@ private[akka] final class OutputStreamSink(createOutput: () ⇒ OutputStream, va
...
+ */
+private[akka] final class JavaInputStreamSink(timeout: FiniteDuration, val attributes: Attributes, shape: SinkShape[ByteString])
+  extends SinkModule[ByteString, japi.Pair[InputStream, Future[JLong]]](shape) {
+  import akka.dispatch.ExecutionContexts.{ sameThreadExecutionContext ⇒ ec }
+
+  val sink = new InputStreamSink(timeout, attributes, shape)
+
+  override def create(context: MaterializationContext) = {
+    val result = sink.create(context)
+    (result._1, new japi.Pair(result._2._1, result._2._2.map(long2Long)(ec)))
```
rkuhn: this is a classical case of where the `akka.dispatch.ExecutionContexts.sameThreadExecutionContext` is appropriate

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala
```diff
@@ -0,0 +1,133 @@
...
+import akka.stream.actor.{ ActorSubscriberMessage, RequestStrategy }
+import akka.util.ByteString
+
+import scala.concurrent.Promise
+
+/** INTERNAL API */
+object InputStreamSubscriber {
+  def props(completionPromise: Promise[Long], bufSize: Int): Props = {
+    require(bufSize > 0, "buffer size must be > 0")
+    Props(classOf[InputStreamSubscriber], completionPromise, bufSize).withDeploy(Deploy.local)
```
rkuhn: please use `Props(new ...)` here, it is faster and safer

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  def receive = {
+    case ActorSubscriberMessage.OnNext(bytes: ByteString) ⇒
+      availableChunks :+= bytes
+      tryToReply(sendRest = false)
+
+    case ActorSubscriberMessage.OnError(cause) ⇒
+      tryToReply(sendRest = true)
+      bytesWrittenPromise.failure(cause)
```
rkuhn: this is not consistent with the FileSink; we need to make a decision on how these shall behaved, see #18565

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala
```diff
@@ -0,0 +1,133 @@
...
+      context.stop(self)
```
rkuhn: the InputStream also needs to be switched into a “failed” mode here, which means that the Actor will need to have access to it

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala
```diff
@@ -0,0 +1,133 @@
...
+    bytesWrittenPromise.trySuccess(bytesSentDownstream)
+    super.postStop()
+  }
+
+  def tryToReply(sendRest: Boolean) =
+    bytesRequestedDownstream match {
+      case Some((actor, length)) ⇒
+        if (!sendRest) {
+          if (availableBytes >= length)
+            sendBytes(actor, length)
```
rkuhn: InputStream.read routinely returns less than was expected when the data source has provided some but not enough data; I think we should do the same here.

rkuhn:
Thinking a bit more about the actual data transfer: it would be more efficient and in particular allow higher throughput to use a LinkedBlockingQueue for the communication between the Input/OutputStream and the Actor: the stream impl would send wake-up messages to the Actor (because the Actor must not use blocking enqueue/dequeue operations) while the stream impl uses blocking put/take operations.

rkuhn:
Otherwise this looks very promising, thanks @agolubev!

agolubev:
@rkuhn, thanks for great feedback. Will fix obvious staff on weekend and perhaps will follow up then with LinkedBlockingQueue related questions

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
agolubev: >This is not correct: InputStream.read() is supposed to return -1 upon having reached EOF.

the initial idea was that InputStream is considered to be closed when onComplete came and in this case read returns IOException. Consequently, there is no case when -1 returns
However we can think of closing stream as some detached state for InputStream and in this case -1 totally make sense, and IOException will happen only when InputStream.close is called

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala
agolubev: Yeah, this is the key point. I had a lot of troubles cause potentially any request from InputStream can be timed out because reactive stream already completed

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
agolubev: Not sure I understand you. Method expectNext checks with assert() if next element is ByteString("abc")

akka-stream/src/main/resources/reference.conf
agolubev: Ok, just in case. I'll rename streams file-io-dispatcher. 
akka-actor will stay with it's
```file-io-dispatcher = "akka.actor.default-dispatcher"``` and akka-http will point to blocking ```file-io-dispatcher = ${akka.stream.blocking-io-dispatcher}```

File: akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala (hunk 47 above)
Comment (agolubev) [inline akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala]: @rkuhn  in akka-streams there is no use of ```def apply[T <: Actor: ClassTag](creator: ⇒ T): Props``` method.
Is your concern about starting up stream and that user can use InputStream.read() before materialization finished?

Comment (hepin1989) [inline akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala]: He means in the companian object of ```InputStreamSubscriber```,we could make use of the ```
def apply[T <: Actor: ClassTag](creator: ⇒ T): Props``` which is more typesafe and avoid the reflection .Once you changed the actor's constructor,then it will have compile error,and you could detect the compile error quickly:).

Comment (agolubev) [inline akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala]: @hepin1989, I got your point. Still I'm looking for reasoning why InputStreamSubscriber is different than all other Subscrbers/Publishers actors. Cause in this case I'll change to "new.." behavior within this ticket. However, if there is only a general rule - then we need a separate ticket to make it consistent everywhere

Comment (hepin1989) [inline akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala]: @agolubev I think @rkuhn just come up with this once he looked at your PR.
so I think you may open a ticket.this seems like an optimization.

File: akka-stream-tests/src/test/java/akka/stream/io/InputStreamSinkTest.java [hunk 51]
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @ClassRule
+    public static AkkaJUnitActorSystemResource actorSystemResource = new AkkaJUnitActorSystemResource("InputStreamSink",
+            Utils.UnboundedMailboxConfig());
+    @Test
+    public void mustReadEventViaInputStream() throws Exception {
+        final FiniteDuration timeout = FiniteDuration.create(300, TimeUnit.MILLISECONDS);
+
+        final Sink<ByteString, Pair<InputStream,Future<Long>>> sink = InputStreamSink.create(timeout);
```
drewhk: do we really need a Future<Long> here?

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
```diff
@@ -0,0 +1,247 @@
...
+  val settings = ActorMaterializerSettings(system).withDispatcher("akka.actor.default-dispatcher")
+  implicit val materializer = ActorMaterializer(settings)
+
+  val timeout = 300.milliseconds
+  val byteString = ByteString("abc")
+  val byteArray = "abc".toArray
+  def newArray() = new Array[Byte](3)
+
+  def expectBlocked[T](f: Future[T]) =
+    the[Exception] thrownBy Await.result(f, timeout) shouldBe a[TimeoutException]
```
drewhk: using a probe and expectNoMsg?

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
```diff
@@ -0,0 +1,247 @@
...
+      expectSuccess(f, 6)
+    }
+
+    "returns less than was expected when the data source has provided some but not enough data" in assertAllStagesStopped {
+      val (probe, (inputStream, f)) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+
+      probe.sendNext(ByteString("ab"))
+      val arr = newArray()
+      inputStream.read(arr)
+      arr should ===("ab\u0000".toArray)
```
drewhk: why the trailing zero?

drewhk: ah because of array size...

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
```diff
@@ -66,3 +74,163 @@ private[akka] final class OutputStreamSink(createOutput: () ⇒ OutputStream, va
...
 }
+
+/**
+ * INTERNAL API
+ * Creates simple synchronous (Java 6 compatible) Sink which provide all incoming element on read request via
+ * InputStream.
+ */
+private[akka] class InputStreamSink(val timeout: FiniteDuration,
+                                    val attributes: Attributes, shape: SinkShape[ByteString])
+  extends SinkModule[ByteString, (InputStream, Future[Long])](shape) {
```
drewhk: do we really need the Future[Long]? If we do, then I recommend creating an explicit SinkInputStream interface that exposes that Future, then there is no need to return a pair.

drewhk: (same for output side)

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala
```diff
@@ -0,0 +1,120 @@
...
+  var sentCompleteNotification = false
+
+  override protected val requestStrategy = WatermarkRequestStrategy(highWatermark = bufSize)
+
+  def receive = {
+    case ActorSubscriberMessage.OnNext(bytes: ByteString) ⇒
+      if (bytes.nonEmpty) {
+        // ignore empty ByteStrings
+        val added = sharedBuffer.offer(bytes)
+        assert(added, "Bytes must always be added to buffer as adapter consuming them")
```
drewhk: please don't use assert, they produce AssertionErrors which is not handled by NonFatal, and therefore bubbles right through the whole actor hierarchy with the final effect of killing the ActorSystem. Please use require() instead.

drewhk: please do a search-replace on asserts if there are more.

drewhk:
In general looks OK, but there might be improvements lurking:
- reuse the blocking queue for events to reduce the need for asks (even if not completely eliminate)
- try how all this looks like as a GraphStage(WithMaterializedValue)

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
agolubev: I added Future[Long] to be consistent with existing IO API (InputStreamSource and OutputStreamSink)

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSubscriber.scala
agolubev: Got it, thanks

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
drewhk: I think it is better to add it as a field of the returned InputStream. Just return an interface that makes the field accessible.

rkuhn: good point, and good solution as well!

agolubev:
@drewhk thanks, 
* make sense for using queue for events and data 
* will look into GraphStage(WithMaterializedValue) - still don't have understanding

drewhk:
@agolubev here is a simple example how to build a source with GraphStage:

      abstract class SourceStage[Out, M](name: String) extends GraphStageWithMaterializedValue[SourceShape[Out], M] {
        val out: Outlet[Out] = Outlet[Out](name + ".in")
        override val shape: SourceShape[Out] = SourceShape(out)
      }

      class TestAsyncSource[Out](asyncApi: () ⇒ Future[Out]) extends SourceStage[Out, Unit]("testSource") {

        override def createLogicAndMaterializedValue: (GraphStageLogic, Unit) = {

          val logic = new GraphStageLogic(shape) {
            private val callback: AsyncCallback[Try[Out]] = getAsyncCallback(onAsyncElem)

            private def onAsyncElem(event: Try[Out]): Unit = event match {
              case Success(elem) ⇒ push(out, elem)
              case Failure(ex)   ⇒ failStage(ex)
            }

                asyncApi().onComplete(callback.invoke)(interpreter.materializer.executionContext)
              }
            })
          }

          (logic, ())
        }

      }

drewhk:
It is a simple source that emits asynchronous events from a future based API in a backpressured way. This is of course not what you need, but I wanted to show how to get asynchronous events inside such a stage (there is no ActorRef to send to).

agolubev:
Thanks @drewhk.

drewhk:
You are the first real outsider to test this API, so your experience with it is very valuable.

rkuhn:
This is superseded by #18707.

PR #17996, opened 2015-07-15
=htc #17992 prevent single failing server connections from cancelling the binding

### LINKED ISSUE #17992: bindAndHandleWith unbinds if a connection fails (which can be caused by client)
If you create a simple server using `Http().bindAndHandleWith`, any failure in the connection stream will cause the connection source to be cancelled, which subsequently causes the server to unbind.

The problem appears to be the mapAsyncUnordered at Http.scala#L97. The future here comes from the 'identityFinishReporter', which fails when a connection stream fails.

This comes before the handler is used, so user code cannot produce this behaviour, but the TCP stream can! In particular, when a connection is reset by a client unexpectedly a `akka.stream.StreamTcpException: The connection closed with error An existing connection was forcibly closed by the remote host` is sometimes* produced.

In other words, a badly behaved client can take down the server by forcibly closing a single connection!

*I couldn't produce this particularly reliably, I don't know exactly when a connection is deemed to have completed gracefully as opposed to forcibly.

jrudolph:
Fixes #17992.

/cc @sirthias

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
@@ -278,7 +278,7 @@ private[http] object StreamUtils {
   }
 
   /**
-   * Returns a no-op flow that materialized to a future that will be completed when the flow gets a
+   * Returns a no-op flow that materializes to a future that will be completed when the flow gets a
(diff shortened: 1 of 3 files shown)


2m:
LGTM

ktoso:
Thanks, makes sense; LGTM

PR #17562, opened 2015-05-25
+str #16965 add generalized UnzipWith

agolubev:
This is continuation of pull request #17533 cause I moved vhanged to 2.3-dev
Here is the list of previous feedback:
- Move to 2.3 dev branch - Done
- Need to add header with (c) to GraphUnzipWithSpec.scala - Done
- Have only one commit for review - Done 
- Mirror to javadsl UnzipWith - Done
- Update javadsl.Graph to support UnzipWith - Done
- Fix spaces in GenJunctions.scala.template: - Done
- Refactor UnzipWith.scala.template to remove redundant line - Done
- Place identity function level up in scaladsl/Graph.scala - substituted with “identity” standard scala type
- Add UnzipWith to docs - added to stream-graphs.rs for scala and java and to stages-overview with tables.

PR #17998, opened 2015-07-15
Agolubev #16965 unzip with 2.3 dev based

### LINKED ISSUE #16965: add generalized UnzipWith
This should be a more approachable form of FlexiRoute that works using a single function:
~~~scala
def unzip[A, B] = UnzipWith((p: (A, B)) => (p._1 :: Nil, p._2 :: Nil))
~~~
meaning that upon receiving an input each of the outputs receives zero or more (strictly available) elements. The real Unzip will of course not be implemented like this in order to avoid the extra allocations, but as user feature this is way easier than having to worry about demand and a RouteLogic implementation.

ktoso:
Pushing https://github.com/akka/akka/pull/17562 over the finish-line (uses tuples from akka-actor).

Resolves #16965

Thanks a lot @agolubev!


2m:
LGTM

akka-docs-dev/rst/java/stream-graphs.rst
```diff
@@ -29,6 +29,7 @@ Akka Streams currently provide these junctions (for a detailed list see :ref:`st
 
  - ``Broadcast[T]`` – *(1 input, N outputs)* given an input element emits to each output
  - ``Balance[T]`` – *(1 input, N outputs)* given an input element emits to one of its output ports
+ - ``UnzipWith[In,A,B,...]`` – *(1 input, N outputs)* takes a function of 1 input that given a value for each input emits N output elements (where N <= 20)
```
ktoso: lol, just noticed we have `[]` in java documentation here in general. Fixing

Issue #18007, opened 2015-07-15
Don't use deprecated scala.concurrent.util.Unsafe -- to be removed in 2.12.0


adriaanm:
You could [inline it](https://github.com/scala/scala/pull/4629/files#diff-39840923e78c5b9ecc80ab7b8daf2e57) into Akka's `Unsafe` wrapper, I suppose. We have to make the core Scala library Java 9 friendly in 2.12, since we move slowly. The first step was taken in https://github.com/scala/scala/pull/4629, which removes our fork of forkjoin (leaving a compatibility stub in place for the equivalent classes in `java.util.concurrent._ `).

ktoso:
Inlining seems like a perfectly good solution to allow you guys to remove the "scala Unsafe" - done that https://github.com/akka/akka/issues/18009

The next release of Akka which does not use `scala.concurrent.util.Unsafe` would be `2.4-M3` then, which is roughly a few weeks away (less than more) - timing sounds good enough?

Seems like that's enough, since `2.3.x` will not be released against Scala `2.12` anyway, so no need to change there.

ktoso:
The bigger problem for us will be the actual `sun.misc.Unsafe`. 
There's ongoing discussions on a number of mailing lists, we'll see what the way forward will be soon (Also, useful link to analysis of Usafe removal impact: http://bit.ly/sunmiscunsafe )

adriaanm:
Awesome! Thanks for the lightning fast response :)

PR #18009, opened 2015-07-15
Remove scala.concurrent.util.Unsafe and Java8 class copies

ktoso:
Resolves https://github.com/akka/akka/issues/18007 for `2.4.x` (to be included in `2.4-M3`),
and also drops classes that we don't have a need to keep copies of now for metrics in testkit lib.

The https://github.com/akka/akka/commit/f3c6ff5665dfea5f6d2c2c51b1be2e9c04f3cbd3 could be easily backported to 2.3.x, though I'm not sure if it gains us anything:
- 2.3 will not be released for Scala `2.12`
  - and I don't think we'll support `2.3.x` on Java 9 (or will we? no idea...) - due to jigsaw probably messing up more than that anyway (well, the entire "jigsaw wants to take away unsafe" story will be a big deal for Akka anyway).

// cc @adriaanm @viktorklang @rkuhn


viktorklang:
Nice work Konrad!

adriaanm:
:fireworks:

PR #18031, opened 2015-07-17
=str,doc Harden RateTransformationDocSpec

ktoso:


2m:
Thanks for the fix. I included it in #18043, as it needs the same changes in the java sample code as well.

ktoso:
Sounds like a plan :)

Issue #18032, opened 2015-07-18
Hook on PersistentActor after message is persisted


jgordijn:
I want to extend the behaviour of the `PersistentActor` such that a function is called (with the persisted event) every time a message is persisted. If the `PersistentActor` provided a hook `onPersistSuccess` it would be possible to extend the behaviour.

The use case is that I want to publish every persisted message on a bus. I would like to mixin a trait into the existing actors to provide this behaviour, but that means that the existing Actor needs to call a new function `persistAndPublish` instead of `persist`. I would like to use this hook so that mixing in this trait is enough.

michaelom:
+1

rkuhn:
I’m not convinced that this is a good idea: it is yet another way in which PersistentActor gets more convenient for the case of keeping variable(s) in the Actor which are acted upon by callbacks instead of properly containing state-dependent information in different behaviors. The common pattern is to have an `applyEvent` method that either updates a variable or transforms a given state into a new state and this method would be the place to put the publishing action (or any cross-cutting concern). Adding a second but invisible way to do the same thing is probably not a good idea.

michaelom:
>The common pattern is to have an applyEvent method that either updates a variable or transforms a given state into a new state and this method would be the place to put the publishing action (or any cross-cutting concern). Adding a second but invisible way to do the same thing is probably not a good idea.

After working on an ES-solution with Akka for almost a year, my experience is, that PersistentActors get very, very complex. Some of the complexity also comes from state/behaviors-dependent actions when persisting events. Yet some actions are executed irrespective of the actors state or behaviour. 

e.g. for state A:
 ```persist(event, persistedEvent -> {
                actionA();
                applyEvent(persistedEvent);
                actionC();
                actionX();
            });```

e.g. for state B:
 ```persist(event, persistedEvent -> {
                applyEvent(persistedEvent);
                actionD();
                actionX();
            });```

Currently we have to repeat `actionX()` for each state. There is no way to declare cross-cutting actions.

By the way, the `appyEvent`-method is not the place I put my publishing logic in. :-)
`appyEvent` does exactly what it says it does.

jgordijn:
@rkuhn I understand your hesitation, but there is no other way to define behaviour that should always be performed for every event. 

[...]

PR #18033, opened 2015-07-18
+per #18032 Add onPersistSuccess hook on PersistentActor

jgordijn:


akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -546,6 +546,7 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
           updateLastSequenceNr(p)
           try {
             peekApplyHandler(p.payload)
+            onPersistSuccess(p.payload)
```
ktoso: what if this throws?
(needs try finally)

ktoso:
OK TO TEST

ktoso:
On one hand I understand your motivation for this, on the other hand we had explicitly planned to not provide this callback, as persist() already has the callback "right in it" and it could simply be `persist()(publishEvent andThen somethingElse)`

ktoso:
refs https://github.com/akka/akka/issues/18032

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
jgordijn: Ok, so the way I understand this, is that in the finally block I should check if there are other messages persisted and flush those. I will look into that.

jgordijn:
@ktoso Yes I am currently doing something like `persist()(publishEvent andThen somethingElse).`, but this forces the user of the trait to write this for every `persist` call in their code. I need this hook so I can make sure that the trait user will not forget to call a different function or use the `andThen` construction. 

I understand 'as persist() already has the callback "right in it" ', but there's no generic way to extend this. I want to open PersistentActor for extension.

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
ktoso: I mean, what if the users implementation of `onPersistSuccess` throws.

ktoso:
> @ktoso Yes I am currently doing something like persist()(publishEvent andThen somethingElse)., but this forces the user of the trait to write this for every persist call in their code. I need this hook so I can make sure that the trait user will not forget to call a different function or use the andThen construction.

I understand, and would actually be in favour of this reasoning, esp. as I've last week come to appreciate pub-sub-ing on persisted events to feed an akka-stream by these – it still needs a round of ACKs from the rest of the team (because this is something we spent a lot of time discussing recently :-)), and sadly most are currently on vacation :smile: 

What I can promise, is an open discussion about this soon though :-)

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
jgordijn: IMHO it should be handled in the same way the callback is handled. 

The only omission is I guess the functionality inside `peekApplyHandler` to flush any added persists. Am I missing something else?

jgordijn:
@ktoso thanks for taking the time to look into this.

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
ktoso: Yeah, I think you're right. Would like to include a test for this case then - to make sure the onPersistFailure is properly invoked (I expect it to be, but if it's designed in behaviour, it needs a test :-))

ktoso: Another thought (sorry, just at my notebook in random short bursts): since persists succeeded, yet the handler may have thrown, technically we still should have gotten the onPersistSuccess callback, since persistence succeeded, and then the handler failed, so slightly different no?

jgordijn: No need to be sorry, great thought. I totally overlooked that one. Currently, when the callback throws, all messages get unstashed and the exception is thrown so that the Actor will (if not implemented a different supervisor strategy) restart.

What should really happen? The persist callback failed (unable to update the internal state for instance). What does this mean for the `onPersistSuccess` if it uses the internal state? 
In my case I want to publish on the `EventStream` that a message is successfully persisted, so the hook should be called always when persistence succeeded, which is the case here. 

In my current user code and in your proposed code (`(publishEvent andThen somethingElse)`) the hook is called first. This essentially turns the problem around, but might be more straight forward to reason about. I've got the feeling that this might be the best solution. Always call the `onPersistSuccess` first and when that succeeded call the persist callback. Then document that the `onPersistSuccess` is called before the callback. When the `onPersistSuccess` throws an Exception the callback will not be called. 

I don't think it is possible to make this foolproof, but this seems to be the most safe way and straight forward solution to me. What do you think?

jgordijn:
I've changed the order of application of the callback handler and onPersistSuccess. I've also added a test that the handler is not called when the `onPersistSuccess` threw an Exception.

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -544,8 +544,10 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
         // while message is in flight, in that case we ignore the call to the handler
         if (id == instanceId) {
           updateLastSequenceNr(p)
+          val applyHandler = applyHandlerOn(p.payload) _
```
rkuhn: this is a bit over the top allocation-wise ;-) please call applyHandlerOn directly in the code below

akka-persistence/src/test/scala/akka/persistence/PersistentActorSpec.scala
```diff
@@ -1132,6 +1177,49 @@ abstract class PersistentActorSpec(config: Config) extends PersistenceSpec(confi
...
+      persistentActor ! "a"
+      expectMsg(982417)
+      listener.expectMsg("message: 982417")
+    }
+    "not call the callback hook when onPersistSuccess throws an exception" in {
+      val persistentActor = namedPersistentActor[ExceptionInHookAfterPersistActor]
+      val listener = TestProbe()
+      system.eventStream.subscribe(listener.ref, classOf[String])
+      persistentActor ! "a"
+      expectNoMsg(100.millis)
```
rkuhn: This line can also be dropped: for a negative assertion it is too short, but extraneous messages will be detected in following tests anyway.

ktoso:
> That's why I think the onPersistSuccess should be called first. When the implementer decides to put
> some logic in this hook and maybe change the state, we cannot be sure that the actor is in a consistent
> state when an exception is thrown. That's why I think that the handler from persist should not be called
> in this case.

When an exception is thrown, in any case, we can't assume the actor is in a consistent state I'd say.

> IMHO the read model should not directly be updated via the events that are emitted in this way.  
> There is no guarantee that you will receive all events. My idea is to have a PersistentView listening and 
> issuing Update commands to self to update from the journal. This will give far better guarantees.

I think it depends – I can imagine an event stream which is "catching up from the database (events `1 to m`)" while new events keep being added on the write side. At some point the polling "has caught up", and it could consume the event directly from the event-bus. It can know if it can use this event or should use polling based on "to which seqNr it has reached" – if the incoming event is `currentSeqNr+1` it can directly push it through the stream (if demand is available on it), otherwise it can drop the event and keep polling once demand arrives. I think this is a very tempting optimisation we could easily implement using the new query side.

> In effect it is replacing db polling with a signal when to fetch new data.

Yeah, it can be used like that, though directly feeding other systems (pub-sub-ish) also seems like a valid solution to me.

akka-persistence/src/test/scala/akka/persistence/PersistentActorSpec.scala
```diff
@@ -1132,6 +1177,49 @@ abstract class PersistentActorSpec(config: Config) extends PersistenceSpec(confi
       expectMsg(List("b-1", "b-2"))
     }
 
+    "be extendible by hooking into onPersisted, which is called after persist" in {
```
rkuhn: `s/extendible/extensible/`

rkuhn:
The PR looks good apart from the small things I commented on, but as mentioned in #18032 I am not convinced that this is a good idea overall.

jgordijn:
@ktoso you're right. Using Update is not really necessary. Thanks for the pointer.

I will process the other comments.

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -585,6 +586,12 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
   }
 
   /**
+   * Hook to be able to extend the PersistentActor with behaviour when a message is persisted.
+   * @param msg The message that was persisted to the journal
+   */
+  def onPersistSuccess(msg: Any): Unit = {}
```
patriknw: our convention is to use `()` instead of `{}` here

patriknw:
Thanks for the PR, I need to think some of the feature though. I'm by default against growing the api.

patriknw:
Refs #18032

patriknw:
See conclusion in the issue.

Issue #18034, opened 2015-07-18
Streams: OutputStreamSubscriber calls println


agolubev:

ktoso:
Ouch, good catch thanks.

PR #18035, opened 2015-07-18
=str #18034 remove println call from OutputStreamSubscriber

agolubev:
just removed it as exception is sending to user with promise.failure


ktoso:
refs https://github.com/akka/akka/issues/18034

ktoso:
OK TO TEST

ktoso:
Great catch, thanks a lot for fixing it!

PR #18039, opened 2015-07-19
Update stream-composition.rst

netcomm:
"nestedFlow will inherit from nestedSource",should be "from nestedSink"?

PR #18042, opened 2015-07-20
Update stream-composition.rst

netcomm:
fix a mistake


2m:
Nice catch. And it looks like it was only on scala side of the docs. Thanks!

2m:
This came together with #18039

PR #18024, opened 2015-07-17
=doc #18012 rate transformation docs

2m:
Refs #18012

Issue #18012, opened 2015-07-16
Remove last todo-pages from akka-http docs


ktoso:
Hi all,
we have missed a number of TODO pages and released 1.0 without them - this is no good.

The pages are:

java:
- [x] stream-rate.rst -- 2m
- [x] http/client-side/connection-level.rst `// has scala-part`
- [x] http/client-side/host-level.rst `// has scala-part`
- [x] http/client-side/https-support.rst `// has scala-part`
- [x] http/client-side/request-level.rst `// has scala-part`
- [x] http/routing-dsl/handlers.rst `// mostly complete`
- [x] http/server-side/low-level-server-side-api.rst
- [x] http/server-side/websocket-support.rst

scala:
- [x] stream-rate.rst -- 2m
- [x] http/client-side/connection-level.rst
- [x] http/server-side/low-level-server-side-api.rst
- [ ] http/server-side/migration-from-spray.rst https://github.com/akka/akka/issues/17972
- [x] http/routing-dsl/directives/security-directives/index.rst `// TODO explain http in this picture`

common: 
- [x] http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0/scala/http/common/json-support.html
- [x] http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0/scala/http/common/marshalling.html
- [x] http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0/scala/http/common/unmarshalling.html
- [x] http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0/scala/http/common/de-coding.html
- [x] http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0/scala/http/common/json-support.html
- [x] http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0/scala/http/common/xml-support.html

additional (not needed asap):
- [ ] explain the magnet https://github.com/akka/akka/issues/17749

We should fix these ASAP, as it's kind of a screw up on our side with the 1.0... We'll start with those which have a scala side and need the java side (a few), but would need assistance for the others, can we count on you guys to help out here? // @jrudolph @sirthias

ktoso:
Please ping the ticket here if working on any of these. // cc @2m

2m:
I am working on `stream-rate` scala+java

sirthias:
I'll take care of the http stuff on the scala side.

ktoso:
Thanks Mathias! We're thinking how to ship these now. 
[...]

PR #18043, opened 2015-07-20
=doc #18012 rate transformation doc improvements

2m:
Addresses feedback from #18024. Thanks @ktoso!
- use () when calling scala methods with side effects
- concat immutable lists instead of modifying mutable ones
- use Collections.singletonList instead of sublassing ArrayList
- and typo fixes


ktoso:
LGTM, thanks

ktoso:
btw good that we don't have "stream()" methods, would be uber confusing when mixed with java8 stuff

2m:
Failure was timing related. An effort to fix it was in #18031. I include fix here, as it needs a java counterpart as well.

2m:
Failure unrelated.

Issue #18050, opened 2015-07-21
Integrate EventAdapters to the Query side


ktoso:
Since EventAdapters can be used to detach the data model from the domain model, they must be a first class citizen in the query side as well.

**Idea 1:**
~~Make configurable (same as the write side, transparently) and tuneable via:~~

```scala
ReadJournal {
  def query...
  def withEventAdapters(config: Config): ReadJournal
  def withEventAdapters(adapters: (Class[_], EventAdapter)*): ReadJournal
```

~~ReadJournals do not have to implement anything to make this possible, however the methods would be nicely accessible for users on ReadJournal.~~

**Idea 2:**
Make it easy enough to obtain `EventAdapters` such that it can be applied as a map step. This again needs construction in code as `(Class[_] -> EventAdapter)*`. So similar to the first idea, however detached from the readjournal. 

**Idea 3:**
Something else?

ahjohannessen:
Would it make sense to define `EventAdapter` in terms of `EventReadAdapter` and `EventWriteAdapter`:

```scala
trait EventReadAdapter {
  def fromJournal(event: Any): ...
}
trait EventWriteAdapter {
  def toJournal(event: Any): ...
}
```
where current `EventAdapter` would extend both, but the `ReadJournal` would only require `EventReadAdapter` and thus minimizing the API surface area to only relevant methods.

ktoso:
Yes good point!
It initially was 2 interfaces like above actually, but there we only thought about the write side still;
Since the read side needs it as well, splitting up makes sense I think (can imagine something starting out as `EventAdapter` and then the "very old code" gets moved to only the read side's adapters).

ahjohannessen:
The thing that prompted me was this: https://github.com/akka/akka/pull/18051/files#diff-ff86a6201dd1acf1ec044cffadbaabfcR69 :)

ktoso:
True true :smile:

ktoso:
1) We'll make more or less like so `stream.map(EventAdapters.apply)`, so no additional API on the journal
2) we'll make it two interfaces https://github.com/akka/akka/issues/18137

patriknw:
I don't think event adapters should be used in the stream transformations because `fromJournal` takes a manifest parameter and that is typically not known at that point.

I think event adapters should be applied by the journal query plugin. That is already done by the leveldb query plugin.
Such an plugin can grab the event adapters via `Persistence(system).adapterFor`. Nothing more needed for this, except from making `EventAdapters` public.

Issue #16541, opened 2014-12-15
Improve the Query Side


patriknw:
[Akka Persistence on the Query Side: The Conclusion](https://groups.google.com/forum/#!topic/akka-user/MNDc9cVG1To)

also discussed in [Improving Akka Persistence wrt CQRS/ES/DDD](https://groups.google.com/forum/#!msg/akka-user/SL5vEVW7aTo/KfqAXAmzol0J) and [Event Stores for Akka Persistence for CQRS?](https://groups.google.com/forum/#!msg/akka-user/4kbYcwWS2OI/hpmAkxnB9D4J)

ktoso:
Should be possible to ship a version of it with M3 (experimental).

PR #18051, opened 2015-07-22
+per #16541 initial version of the Persistence Query module

ktoso:
Ladies and Gentlemen,
Thanks to the [Akka Streams 1.0 release last week](http://akka.io/news/2015/07/15/akka-streams-1.0-released.html), we now would like to present this new experimental module for initial feedback, addressing a long history of discussions and awesome feedback we have gathered from the community (see the great summary: [Akka Persistence on the Query Side: The Conclusion](https://groups.google.com/forum/#!topic/akka-user/MNDc9cVG1To)). 

Introducing the Akka Streams based improved Query Side to Akka Persistence.
This is only the initial PR, more examples, polish and example implementations will be pushed soon.

The current form of the API is the culmination of these discussions with the community, a very long "maturing in the back of your mind" process and a few days spent together around ScalaDays for polishing and trying it out in various example/mock implementations with @patriknw and @dotta. 

The API is intentionally very "open" (we provide the marker traits `Query[T, M]`, `Hint`, and a number of basic case classes), because we do not want to restrict some of the great journals out there - instead, we want to encourage various implementations to expose that kind of API which is best suited for the underlying store. For example, SQL based journals may provide `SqlQuery("SELECT * ...")` if that's what they want to expose. Journals will have to document their behaviours very thoroughly.

Other important information:
- this module (`akka-persistence-query-experimental`) will be (and stay) **experimental** in Akka `2.4`
- we'll provide a number of _example_ `ReadJournal` implementations, however we currently do not intend to maintain one for production usage - we ask the community to help out, similarily like it has happened organically with Akka Persistence.
- integration with the new `EventAdapter`s known from the write-side in Akka 2.4 is under discussion/development https://github.com/akka/akka/issues/18050


ktoso:
Refs #16541

jrudolph:
Error lines in https://jenkins.akka.io/job/pr-validator-per-commit-jenkins/3269/:

```
  1659 Unable to find credentials for [Artifactory Realm @ proxy-ch.typesafe.com].
  1660 Unable to find credentials for [Artifactory Realm @ proxy-ch.typesafe.com].
 12321 (akka-docs/sphinx:generatePdf) Sphinx pdf generation failed.  See debug output for details.
 12322 Total time: 713 s, completed Jul 22, 2015 1:19:57 PM
```

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+tag - for example if the journal stored the events as json it may try to find those with the field ``tag`` set to this value etc.).
+
+.. includecode:: code/docs/persistence/query/PersistenceQueryDocSpec.scala#events-by-tag
+
+As you can see, we can use all the usual stream combinators available from `Akka Streams`_ on the resulting query stream,
+including for example taking the first 10 and cancelling the stream. It is worth pointing out that the built-in ``EventsByTag``
+query has an optionally supported offset parameter (of type ``Long``) which the journals can use to implement resumable-streams.
+For example a journal may be able to use a WHERE clause to begin the read starting from a specific row, or in a datastore
+that is able to order events by insertion time it could treat the Long as a timestamp and select only older events.
+Again, specific sapabilities are specific to the journal you are using, so you have to
```
ahjohannessen: *capabilities*

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+.. _Community plugins: http://akka.io/community/#plugins-to-akka-persistence-query
+
+Performance and denormalization
+===============================
+When building systems using :ref:`event-sourcing` and CQRS (`Command & Query Responsibility Segragation`_) techniques
+it is tremendously important to realise that the write-side has completely different needs from the read-side,
+and separating those concerns into datastores that are optimised for either side makes it possible to offer the best
+expirience for the write and read sides independently.
+
+For example in add bidding systems, it is important to "take the write" and respond to the bidder that we have accepted
```
2m: s/add/ad

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+the bid as soon as possible, which means that write-throughput is of highest importance for the write-side – often this
+means that data stores which are able to scale to accomodate these requirements have a less expressive query side.
+
+On the other hand the same application may have some complex statistics view or we may have analists working with the data
+to figure out best bidding strategies and trends – this often requires some kind of expressive query capabilities like
+for example SQL or writing Spark jobs to analyse the data. To offer data more efficiently to those kinds of
```
2m: Unfinished sentence. :)

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+.. _read-journal-plugin-api-scala:
+
+Query plugins
+=============
+
+Query plugins are various (mostly community driven) :class:`ReadJournal` implementations for all kinds
+of available datastores. The complete list of available plugins is maintained on the Akka Persistence Query `Community Plugins`_ page.
+
+This section aims to provide tips and guide plugin developers through implementing a custom query plugin.
+Most users will not need to implement journals themselfs, except if targeting a not yet supported datastore.
```
2m: s/themselfs/themselves

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+  Since different data stores provide different query capabilities journal plugins **must extensively document**
+  their exposed semantics as well as handled query scenarios.
+
+ReadJournal plugin API
+----------------------
+
+Journals *MUST* return a *failed* ``Source`` if they are unable to execute the passed in query.
+For example if the user accidentally passed in an ``SqlQuery()`` to a key-value journal.
+
+Below is a simple journal implementations:
```
2m: s/implementations/implementation

akka-docs/rst/scala/code/docs/persistence/query/PersistenceQueryDocSpec.scala
```diff
@@ -0,0 +1,212 @@
...
+    override def query[T, M](q: Query[T, M], hints: Hint*): Source[T, M] =
+      q match {
+        case EventsByTag(tag, offset) ⇒
+          val props = MyEventsByTagPublisher.props(tag, offset, refreshInterval(hints))
+          Source.actorPublisher[EventEnvelope](props)
+            .mapMaterializedValue(_ ⇒ noMaterializedValue)
+
+        case unsupported ⇒
+          Source.failed[T](
+            new UnsupportedOperationException(s"Query $unsupported not supported by ${getClass.getName}"))
```
2m: This sticks out of the page in the generated HTML doc output.

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
+.. _persistence-query-scala:
+
+#################
+Persistence Query
+#################
+
```
2m: This page is not linked from anywhere. Probably should be added to `akka-docs/rst/scala/index-actors.rst`.

2m:
LGTM. Very clean API!

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
```diff
@@ -0,0 +1,86 @@
...
+  def props(tag: String, offset: Long, refreshInterval: FiniteDuration): Props =
+    Props(new MyEventsByTagPublisher(tag, offset, refreshInterval))
+
+  private case object Continue
+}
+
+//#events-by-tag-publisher
+class MyEventsByTagPublisher(tag: String, offset: Long, refreshInterval: FiniteDuration)
+  extends ActorPublisher[EventEnvelope] {
+  import MyEventsByTagPublisher._
```
rkuhn: I’d declare the `case object Continue` in here to avoid having to pull the companion into the docs as well.

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  private val limit = 1000
+
+  private var currentId = 0L
+  var buf = Vector.empty[EventEnvelope]
+
+  import context.dispatcher
+  val continueTask = context.system.scheduler.schedule(
+    refreshInterval, refreshInterval, self, Continue)
```
rkuhn: perhaps just give `context.dispatcher` as explicit argument—saves one line and removes doubt as to where else this implicit is used

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
```diff
@@ -0,0 +1,86 @@
...
+
+    case Cancel ⇒
+      context.stop(self)
+  }
+
+  def query(): Unit =
+    if (buf.isEmpty) {
+      try {
+        // Could be an SQL query, for example:
+        //      "SELECT id, persistent_repr FROM journal WHERE tag = like ? and " +
```
rkuhn: `= like` is not correct; could we show some actual code instead of `???` without pulling in all the internets as dependencies?

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
```diff
@@ -0,0 +1,86 @@
...
+        //        "id >= ? ORDER BY id limit ?"
+        val result: Vector[(Long, Array[Byte])] = ???
+        currentId = if (result.nonEmpty) result.last._1 else currentId
```
rkuhn: in particular this line only becomes clear once `currentId` is actually used

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+  @tailrec final def deliverBuf(): Unit =
+    if (totalDemand > 0 && buf.nonEmpty) {
+      if (totalDemand <= Int.MaxValue) {
+        val (use, keep) = buf.splitAt(totalDemand.toInt)
+        buf = keep
+        use foreach onNext
+      } else {
+        val (use, keep) = buf.splitAt(Int.MaxValue)
```
rkuhn: isn’t this a no-op?

akka-docs/rst/scala/code/docs/persistence/query/PersistenceQueryDocSpec.scala
```diff
@@ -0,0 +1,212 @@
...
+
+    private def refreshInterval(hints: Seq[Hint]): FiniteDuration =
+      hints.collectFirst { case RefreshInterval(interval) ⇒ interval }
+        .getOrElse(defaulRefreshInterval)
+
+    private def noMaterializedValue[M]: M =
+      null.asInstanceOf[M]
+  }
+
+  //#my-read-journal
```
rkuhn: including an empty line on purpose?

akka-docs/rst/scala/code/docs/persistence/query/PersistenceQueryDocSpec.scala
```diff
@@ -0,0 +1,212 @@
...
+
+    // TODO from config
+    private val defaulRefreshInterval: FiniteDuration = 3.seconds
+
+    override def query[T, M](q: Query[T, M], hints: Hint*): Source[T, M] =
+      q match {
+        case EventsByTag(tag, offset) ⇒
+          val props = MyEventsByTagPublisher.props(tag, offset, refreshInterval(hints))
+          Source.actorPublisher[EventEnvelope](props)
+            .mapMaterializedValue(_ ⇒ noMaterializedValue)
```
rkuhn: This is not nice: how would a type-safe implementation of this method look like? It might be that the type signature is too ambitious …

akka-docs/rst/scala/code/docs/persistence/query/PersistenceQueryDocSpec.scala
```diff
@@ -0,0 +1,212 @@
...
+  //#projection-into-different-store
+  class TheOneWhoWritesToQueryJournal(id: String) extends Actor {
+    val store = new DummyStore()
+
+    var state: ComplexState = ComplexState()
+
+    def receive = {
+      case m =>
+        state = updateState(state, m)
+        if (state.readyToSave) store.save(Record(state))
```
rkuhn: forgot to reply here (needed by mapAsync below)

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+Akka persistence query complements :ref:`persistence-scala` by providing an universal asynchronous stream based
```
rkuhn: `s/an/a/`

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+query interface that various journal plugins can implement in order to expose their query capabilities.
+
+The most typical use case of persistence query is implementing the so-called query side (also known as "read side"),
```
rkuhn: please remove trailing comma

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+in the popular CQRS architecture pattern - in which the writing side of the application (e.g. implemented using akka
+persistence) is completely separated from the query side which can be implemented using akka persistence query.
```
rkuhn: I’d replace “using” by “with the help of” to create more distance—Akka Persistence Query is _not_ an implementation of the read side.

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+
+While queries may be performed directly on the same datastore, it is also a very common pattern to use the queries
+to create projections of the write-side's events and store them into a separate journal which is optimised for more
+complex queries - this migration of data, possibly including some transformation along the way, can be easily implemented
+using akka persistence query as well - by pulling out of one Journal and storing into another one.
```
rkuhn: This might want to be worded more strongly, namely that this is the recommended approach.

akka-docs/rst/scala/persistence-query.rst
ahjohannessen: I guess you mean *For example, in a bidding system it is...* or *For example, in bidding systems it is...*

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+
+  This module is marked as **“experimental”** as of its introduction in Akka 2.4.0. We will continue to
+  improve this API based on our users’ feedback, which implies that while we try to keep incompatible
+  changes to a minimum the binary compatibility guarantee for maintenance releases does not apply to the
+  contents of the ``akka.persistence.query`` package.
+
+Design overview
+===============
+
+Akka persistence query is purposely designed to be a very loose API.
```
rkuhn: I’d say “loosely specified” instead of “loose”.

akka-docs/rst/scala/code/docs/persistence/query/PersistenceQueryDocSpec.scala
ktoso: > It might be that the type signature is too ambitious …

True, on one hand - from the user's point of view it's great; the implementations though need to perform casts. We would really like to make use of types "based on the query": we know that *"query for persistence ids"* == `Source[String, _]`, so it's `Query[String, Unit]`. For the user of the API it makes complete sense; I did not find a way we could preserve typesafety for both ends, something to investigate further I guess?

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+
+Journal implementers are encouraged to put this identified in a variable known to the user, such that one can access it via
+``journalFor(NoopJournal.identifier)``, however this is not enforced.
+
+Read journal implementations are available as `Community plugins`_.
+
+
+Predefined queries
+------------------
+Akka persistence query comes with a number of ``Query`` objects built in and suggests journal implementors to implement
```
rkuhn: please title-case when using the full feature name

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+them according to the semantics described below. It is important to notice that while these query types are very common
+a journal is not obliged to implement all of them - for example because in a given journal such query would be
+significantly inefficient.
+
+.. note::
+  Refer to the documentation of the ``ReadJournal`` plugin you are using for a specific list of supported query types.
+  For example, Journal plugins should document their stream completion strategies.
+
+The predefined queries are:
```
rkuhn: if this is meant to be an enumeration then the individual points should be put into a bullet-point list

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+persistence ids as they come into the system:
+
+.. includecode:: code/docs/persistence/query/PersistenceQueryDocSpec.scala#all-persistence-ids-live
+
+If your usage does not require a live stream, you can disable refreshing by using *hints*, providing the built-in
+``NoRefresh`` hint to the query:
+
+.. includecode:: code/docs/persistence/query/PersistenceQueryDocSpec.scala#all-persistence-ids-snap
+
+``EventsByPersistenceId`` is a query equivalent to replaying an :ref:`PersistentActor <event-sourcing>`,
```
rkuhn: `s/an/a/`

akka-persistence-query/src/main/scala/akka/persistence/query/Hint.scala
```diff
@@ -0,0 +1,34 @@
...
+ * typically specific to the journal implementation.
+ */
+trait Hint
+
+/**
+ * If the underlying datastore only supports queries that are completed when they reach the
+ * end of the "result set", the journal has to submit new queries after a while in order
+ * to support "infinite" event streams that include events stored after the initial query has completed.
+ *
+ * A plugin may optionally support this [[Hint]] for defining such a refresh interval.
```
rkuhn: “A plugin may optionally support this hint” is probably true for all hints; it should be enough to document this on the `trait Hint` itself.

akka-persistence-query/src/main/scala/akka/persistence/query/PersistenceQuery.scala
```diff
@@ -0,0 +1,89 @@
...
+   *
+   * Returns the [[akka.persistence.query.javadsl.ReadJournal]] specified by the given read journal configuration entry.
+   */
+  final def getReadJournalFor(readJournalPluginId: String): javadsl.ReadJournal =
+    new javadsl.ReadJournal(readJournalFor(readJournalPluginId))
+
+  private def createPlugin(configPath: String): scaladsl.ReadJournal = {
+    require(!isEmpty(configPath) && system.settings.config.hasPath(configPath),
+      s"'reference.conf' is missing persistence read journal plugin config path: '${configPath}'")
+    val pluginActorName = configPath
```
rkuhn: This sounds like it should be used as the name of an Actor somewhere, but that does not seem to be the case.

ktoso: Good catch, thanks

akka-persistence-query/src/main/scala/akka/persistence/query/Hint.scala
ktoso: Very true, all hints are optional.

akka-persistence-query/src/main/scala/akka/persistence/query/scaladsl/ReadJournal.scala
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Usage:
+ * {{{
+ * val journal = PersistenceQuery(system).readJournalFor(queryPluginConfigPath)
+ * val events = journal.query(EventsByTag("mytag", 0L))
+ * }}}
+ *
+ * For Java API see [[akka.persistence.query.javadsl.ReadJournal]].
+ */
+abstract class ReadJournal { self ⇒
```
rkuhn: please remove `self`

project/Dependencies.scala
```diff
@@ -18,6 +18,8 @@ object Dependencies {
   )
 
   object Compile {
+    val akkaStream = "com.typesafe.akka" %% "akka-stream-experimental" % "1.0"
```
rkuhn: please add a FIXME to replace by project dependency before 2.4.0

project/Dependencies.scala
```diff
@@ -116,7 +120,7 @@ object Dependencies {
 
   val osgi = l ++= Seq(osgiCore, osgiCompendium, Test.logback, Test.commonsIo, Test.pojosr, Test.tinybundles, Test.scalatest.value, Test.junit)
 
-  val docs = l ++= Seq(Test.scalatest.value, Test.junit, Test.junitIntf, Docs.sprayJson, Docs.gson)
+  val docs = l ++= Seq(Test.scalatest.value, Test.junit, Test.junitIntf, Docs.sprayJson, Docs.gson, akkaStream)
```
rkuhn: why is this necessary?

akka-persistence-query/src/main/scala/akka/persistence/query/scaladsl/ReadJournal.scala
ktoso: Right, thanks.

Was a leftover from trying out integrating event-adapters: https://github.com/akka/akka/issues/18050 One of the ways to integrate them is to implement them on the ReadJournal directly here. I dropped it, since it felt a bit like mixing up concerns. More impl ideas in that ticket.

akka-persistence-query/src/test/scala/akka/persistence/query/PersistenceQuerySpec.scala
```diff
@@ -0,0 +1,70 @@
...
+      }
+    }
+
+    "throw if unable to find query journal by config key" in {
+      withActorSystem() { system ⇒
+        intercept[IllegalArgumentException] {
+          PersistenceQuery.get(system).readJournalFor(MockReadJournal.Identifier + "-unknown")
+        }.getMessage should include("missing persistence read journal")
+      }
+    }
```
rkuhn: we’ll need to include a somewhat real implementation so that we can also write some sensible tests for the whole infrastructure

rkuhn:
Looking pretty good as a starting point, we’ll need to flesh it out (in subsequent PRs) so that it can actually be tried out without depending on externally contributed implementations.

akka-persistence-query/src/test/scala/akka/persistence/query/PersistenceQuerySpec.scala
ktoso: do you think we should provide it for leveldb?

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
```diff
@@ -0,0 +1,86 @@
...
+import scala.concurrent.duration.FiniteDuration
+
+object MyEventsByTagPublisher {
+  def props(tag: String, offset: Long, refreshInterval: FiniteDuration): Props =
+    Props(new MyEventsByTagPublisher(tag, offset, refreshInterval))
+
+  private case object Continue
+}
+
+//#events-by-tag-publisher
```
patriknw: perhaps move the start tag above the companion, to show where `Continue` comes from?

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
patriknw: +1

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
patriknw: writing it down as jdbc code (PreparedStatement + iterate ResultSet) would be possible

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
patriknw: right

    else {
      buf foreach onNext
      buf = Vector.empty
    }

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+
+Akka persistence query complements :ref:`persistence-scala` by providing an universal asynchronous stream based
+query interface that various journal plugins can implement in order to expose their query capabilities.
+
+The most typical use case of persistence query is implementing the so-called query side (also known as "read side"),
+in the popular CQRS architecture pattern - in which the writing side of the application (e.g. implemented using akka
+persistence) is completely separated from the query side which can be implemented using akka persistence query.
+
+While queries may be performed directly on the same datastore, it is also a very common pattern to use the queries
+to create projections of the write-side's events and store them into a separate journal which is optimised for more
```
patriknw: s/journal/data store/ (or even database)

akka-docs/rst/scala/persistence-query.rst
```diff
@@ -0,0 +1,212 @@
...
+
+.. includecode:: code/docs/persistence/query/PersistenceQueryDocSpec.scala#events-by-tag
+
+As you can see, we can use all the usual stream combinators available from `Akka Streams`_ on the resulting query stream,
+including for example taking the first 10 and cancelling the stream. It is worth pointing out that the built-in ``EventsByTag``
+query has an optionally supported offset parameter (of type ``Long``) which the journals can use to implement resumable-streams.
+For example a journal may be able to use a WHERE clause to begin the read starting from a specific row, or in a datastore
+that is able to order events by insertion time it could treat the Long as a timestamp and select only older events.
+Again, specific sapabilities are specific to the journal you are using, so you have to
+
```
patriknw: perhaps mention something about that a journal may use event adapters for tagging of events?

patriknw:
looking good

akka-docs/rst/scala/persistence-query.rst
ktoso: well, I mean "ad bidding", but can generalise for bigging systems

akka-docs/rst/scala/persistence-query.rst
ktoso: thanks :)

akka-docs/rst/scala/persistence-query.rst
ktoso: thanks :) 

I also have to port it to java :~

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
ktoso: ok

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
ktoso: okey

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
ktoso: ok, can try a JDBC thing here

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
ktoso: Giving it a shot then

akka-docs/rst/scala/persistence-query.rst
patriknw: yeah, I didn't dare to say ;-)

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
ktoso: good catch, thanks

akka-docs/rst/scala/code/docs/persistence/query/PersistenceQueryDocSpec.scala
ktoso: no, will trim. (it does not get shown in rendered output btw)

akka-docs/rst/scala/code/docs/persistence/query/PersistenceQueryDocSpec.scala
ktoso: Shipping as-is as we don't see a simple solution here yet.
Will ask for community feedback to sattle on some API

akka-docs/rst/scala/code/docs/persistence/query/PersistenceQueryDocSpec.scala
ktoso: thanks!

akka-docs/rst/scala/persistence-query.rst
ktoso: thanks :)

akka-docs/rst/scala/persistence-query.rst
ktoso: roger

akka-docs/rst/scala/persistence-query.rst
ktoso: yeah, very true; rewording

akka-docs/rst/scala/persistence-query.rst
ktoso: Will do

akka-docs/rst/scala/persistence-query.rst
ktoso: True (must have been thinking of windsurfing https://en.wikipedia.org/wiki/Shaka_sign, hang loose) ;-)

akka-docs/rst/scala/persistence-query.rst
ktoso: not really, too much content inside (incl. code snippets), enumeration would be a bit weird here. I;ll reword the sentence instead maybe

project/Dependencies.scala
ktoso: done

project/Dependencies.scala
ktoso: good catch, it's not

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
ktoso: Included Continue inside here, so not including companion

akka-docs/rst/scala/persistence-query.rst
ktoso: ok

akka-docs/rst/scala/persistence-query.rst
ktoso: I'll add that in a separate PR - have a ticket open for "query side and event adapters"

ktoso:
Pushed amended state, will add java part to it

akka-docs/rst/scala/persistence-query.rst
ktoso: here https://github.com/akka/akka/issues/18050

akka-docs/rst/java/code/docs/persistence/PersistenceQueryDocTest.java
```diff
@@ -0,0 +1,380 @@
...
+
+        final Props props = MyEventsByTagPublisher.props(tag, offset, refreshInterval(hints));
+
+        return (Source<T, M>) Source.<EventEnvelope>actorPublisher(props)
+                                    .mapMaterializedValue(noMaterializedValue());
+      } else {
+        // unsuported
+        return Source.<T>failed(
+          new UnsupportedOperationException(
+            "Query $unsupported not supported by " + getClass().getName()))
```
patriknw: no string interpolation in java

akka-docs/rst/java/code/docs/persistence/PersistenceQueryDocTest.java
```diff
@@ -0,0 +1,380 @@
...
+                     .mapMaterializedValue(noMaterializedValue());
+      }
+    }
+
+    private FiniteDuration refreshInterval(Hint[] hints) {
+      FiniteDuration ret = defaultRefreshInterval;
+      for (Hint hint : hints) {
+        if (hint instanceof RefreshInterval) {
+          ret = ((RefreshInterval) hint).interval();
```
patriknw: I'd return here

akka-docs/rst/java/code/docs/persistence/PersistenceQueryDocTest.java
```diff
@@ -0,0 +1,380 @@
...
+        }
+      }
+      return ret;
```
patriknw: and return defaultRefreshInterval
(no need for the var `ret`)

akka-docs/rst/java/code/docs/persistence/PersistenceQueryDocTest.java
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    private <I, M> akka.japi.function.Function<I, M> noMaterializedValue() {
+      return param -> (M) null;
+    }
+  }
+  //#my-read-journal
+
+  void demonstrateBasicUsage() {
+    final ActorSystem system = ActorSystem.create();
```
patriknw: perhaps put this as an instance variable

akka-docs/rst/java/code/docs/persistence/query/MyEventsByTagJavaPublisher.java
```diff
@@ -0,0 +1,112 @@
...
+  private final String tag;
+
+  private final String CONTINUE = "CONTINUE";
+  private final int LIMIT = 1000;
+  private long currentOffset;
+  private List<EventEnvelope> buf = new ArrayList<>();
+
+  private Cancellable continueTask;
+
+  public MyEventsByTagJavaPublisher(Connection connection, String tag, Long offset, FiniteDuration refreshInterval) {
```
patriknw: too long line

akka-docs/rst/java/code/docs/persistence/query/MyEventsByTagJavaPublisher.java
```diff
@@ -0,0 +1,112 @@
...
+    this.connection = connection;
+    this.tag = tag;
+    this.currentOffset = offset;
+
+    this.continueTask = context().system().scheduler().schedule(refreshInterval, refreshInterval, self(), CONTINUE, context().dispatcher(), self());
```
patriknw: long line

akka-docs/rst/java/code/docs/persistence/query/MyEventsByTagJavaPublisher.java
```diff
@@ -0,0 +1,112 @@
...
+          final Long id = in.first();
+          final byte[] bytes = in.second();
+
+          final PersistentRepr p = serialization.deserialize(bytes, PersistentRepr.class).get();
+
+          return new EventEnvelope(id, p.persistenceId(), p.sequenceNr(), p.payload());
+        }).collect(toList());
+      } catch(Exception e) {
+          onErrorThenStop(e);
+      }
```
patriknw: ResultSet  and PreparedStatement should be closed in finally

akka-docs/rst/java/code/docs/persistence/query/MyEventsByTagJavaPublisher.java
```diff
@@ -0,0 +1,112 @@
...
+    SerializationExtension.get(context().system());
+
+  private final Connection connection;
+
+  private final String tag;
+
+  private final String CONTINUE = "CONTINUE";
+  private final int LIMIT = 1000;
+  private long currentOffset;
+  private List<EventEnvelope> buf = new ArrayList<>();
```
patriknw: use LinkedList, remove(0) in ArrayList is very expensive

akka-docs/rst/java/persistence-query.rst
```diff
@@ -0,0 +1,236 @@
...
+persistence) is completely separated from the "query side". Akka Persistence Query itself is *not* directly the query
+side of an application, however it can help to migrate data from the write side to the query side database. In very
+simple scenarios Persistence Query may be powerful enough to fulful the query needs of your app, however we highly
+recommend (in the spirit of CQRS) of splitting up the write/read sides into separate datastores as the need arrises.
+
+While queries can be performed directly on the same datastore, it is also a very common pattern to use the queries
+to create *projections* of the write-side's events and store them into a separate datastore which is optimised for more
+complex queries. This architectural pattern of projecting the data into a query optimised datastore, with possibly some
+transformation or canculations along the way is the core use-case and recommended style of using Akka Persistence Query
+- pulling out of one Journal and storing into another one.
```
patriknw: Feels like this section is almost telling the same thing as the previous section.
I like the first one, see if something is unique here and if it can be incorporated in the first.

akka-docs/rst/java/persistence-query.rst
```diff
@@ -0,0 +1,236 @@
...
+=============
+
+In order to issue queries one has to first obtain an instance of a ``ReadJournal``.
+Read journals are implemented as `Community plugins`_, each targeting a specific datastore (for example Cassandra or JDBC
+databases). For example, given a library that provides a ``akka.persistence.query.noop-read-journal`` obtaining the related
+journal is as simple as:
+
+.. includecode:: code/docs/persistence/PersistenceQueryDocTest.java#basic-usage
+
+Journal implementers are encouraged to put this identified in a variable known to the user, such that one can access it via
```
patriknw: s/identified/identifier/

akka-docs/rst/java/persistence-query.rst
```diff
@@ -0,0 +1,236 @@
...
+.. includecode:: code/docs/persistence/PersistenceQueryDocTest.java#all-persistence-ids-live
+
+If your usage does not require a live stream, you can disable refreshing by using *hints*, providing the built-in
+``NoRefresh`` hint to the query:
+
+.. includecode:: code/docs/persistence/PersistenceQueryDocTest.java#all-persistence-ids-snap
+
+``EventsByPersistenceId`` is a query equivalent to replaying a :ref:`PersistentActor <event-sourcing>`,
+however, since it is a stream it is possible to keep it alive and watch for additional incoming events persisted by the
+persistent actor identified by the given ``persistenceId``. Most journal will have to revert to polling in order to achieve
```
patriknw: s/journal/journals/

akka-docs/rst/java/persistence-query.rst
```diff
@@ -0,0 +1,236 @@
...
+tag - for example if the journal stored the events as json it may try to find those with the field ``tag`` set to this value etc.).
+
+.. includecode:: code/docs/persistence/PersistenceQueryDocTest.java#events-by-tag
+
+As you can see, we can use all the usual stream combinators available from `Akka Streams`_ on the resulting query stream,
+including for example taking the first 10 and cancelling the stream. It is worth pointing out that the built-in ``EventsByTag``
+query has an optionally supported offset parameter (of type ``Long``) which the journals can use to implement resumable-streams.
+For example a journal may be able to use a WHERE clause to begin the read starting from a specific row, or in a datastore
+that is able to order events by insertion time it could treat the Long as a timestamp and select only older events.
+Again, specific capabilities are specific to the journal you are using, so you have to
```
patriknw: incomplete sentence

akka-docs/rst/java/persistence-query.rst
```diff
@@ -0,0 +1,236 @@
...
+and separating those concerns into datastores that are optimised for either side makes it possible to offer the best
+expirience for the write and read sides independently.
+
+For example, in a bidding system it is important to "take the write" and respond to the bidder that we have accepted
+the bid as soon as possible, which means that write-throughput is of highest importance for the write-side – often this
+means that data stores which are able to scale to accomodate these requirements have a less expressive query side.
+
+On the other hand the same application may have some complex statistics view or we may have analists working with the data
+to figure out best bidding strategies and trends – this often requires some kind of expressive query capabilities like
+for example SQL or writing Spark jobs to analyse the data. Trefore the data stored in the write-side needs to be
```
patriknw: s/Trefore/Therefore/

akka-docs/rst/java/persistence-query.rst
```diff
@@ -0,0 +1,236 @@
...
+projected into the other read-optimised datastore.
+
+.. note::
+  When refering to **Materialized Views** in Akka Persistence think of it as "some persistent storage of the result of a Query".
+  In other words, it means that the view is created once, in order to be afterwards queries multiple times, as in this format
```
patriknw: s/queries/queried/

akka-docs/rst/java/persistence-query.rst
```diff
@@ -0,0 +1,236 @@
...
+  it may be more efficient or interesting to query it (instead of the source events directly).
+
+Materialize view to Reactive Streams compatible datastore
+---------------------------------------------------------
+
+If the read datastore exposes it an `Reactive Streams`_ interface then implementing a simple projection
```
patriknw: remove **it**

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
patriknw: ping

patriknw:
LGTM, after final touch

ahjohannessen:
Gotta love collaboration, feedback is great :+1: :smile:

akka-docs/rst/java/code/docs/persistence/PersistenceQueryDocTest.java
ktoso: ouch, thanks!

akka-docs/rst/java/code/docs/persistence/PersistenceQueryDocTest.java
ktoso: yeap, good idea

akka-docs/rst/java/code/docs/persistence/PersistenceQueryDocTest.java
ktoso: +1

akka-docs/rst/java/code/docs/persistence/PersistenceQueryDocTest.java
ktoso: The reason that it's here is for it to never actually be created (those are "compile tests")

akka-docs/rst/java/code/docs/persistence/query/MyEventsByTagJavaPublisher.java
ktoso: fixed

akka-docs/rst/java/code/docs/persistence/query/MyEventsByTagJavaPublisher.java
ktoso: It's kinda funny how long it is actually...
Agreed, chopping down

akka-docs/rst/java/code/docs/persistence/query/MyEventsByTagJavaPublisher.java
ktoso: whoops, good catch, thanks!

ktoso: actually, `try (ResultSet rs = s.executeQuery()) {` :)

akka-docs/rst/java/code/docs/persistence/query/MyEventsByTagJavaPublisher.java
ktoso: true, thanks

akka-docs/rst/java/persistence-query.rst
ktoso: good catch, the 1st part I amended recently due to feedback; Seems now they cover the same topic. Removing the 2nd part.

akka-docs/rst/java/persistence-query.rst
ktoso: thanks

akka-docs/rst/java/persistence-query.rst
ktoso: thanks; removed it, forgot what I was to re-emphasise again (well, likely that "read your journal docs", but that's been mentioned a lot already)

akka-docs/rst/java/code/docs/persistence/query/MyEventsByTagJavaPublisher.java
patriknw: AutoCloseable yay

ktoso: *oh my god, it's the future!* (feels nice to use post-6 things ;))

akka-docs/rst/scala/code/docs/persistence/query/MyEventsByTagPublisher.scala
ktoso: thanks

akka-docs/rst/scala/persistence-query.rst
ktoso: Java part... done!

ktoso:
Addressed all feedback, merging after validation (failure was javadoc)

akka-docs/rst/scala/code/docs/persistence/query/PersistenceQueryDocSpec.scala
scullxbones: Probably not the ideal spot to discuss, but the context of the discussion between @ktoso and @rkuhn is helpful.

I started to implement a `ReadJournal` today for my mongo journal and the `query` method seemed a bit off.  The casting wasn't great, but also I didn't really like communicating with the user that a query was not supported via exception.  More type-safety would be nice.

Intuitively this seems like a good spot for typeclasses.  I'm thinking that the (read) journal implementers would provide implicit objects as witnesses for the queries they support, e.g.
(Excuse the hastily put together pseudo-pseudocode, there must be a better signature to remove the extra T & M parameters on the trait)

```scala
package akka.persistence.query
abstract class ReadJournal { 
...
trait SupportedQueryWitness[A <: Query[T,M],T,M]
}

package my.read.journal
object SupportedQueries {
  implicit object allPids extends SupportedQueryWitness[AllPersistenceIds,String,Unit]
}
```

Then `ReadJournal.query` could require a witness for that `Query[T,M]` to be in implicit scope.  The downside of this I see is that the user of the read journal needs a compile-time dependency on the journal implementation, rather than the typical run-time approach.  Of course they may need this anyways for more interesting `Query[T,M]`s.

Anyways, just some initial reactions.  Thanks for all the work you do, it's greatly appreciated!

lol
```diff
@@ -0,0 +1,172 @@
+adoc-api/akka/persistence/query/scaladsl/ReadJournal.html...
```
xuwei-k: unnecessary log file?

ktoso: Thanks for noticing @xuwei-k, removed https://github.com/akka/akka/commit/9b6b43335f5de6f1f3289ee8614d5d79c75a1bab 
(used this file to debug javadoc failure using `./scripts/find-javadoc-error.pl`)

akka-docs/rst/scala/code/docs/persistence/query/PersistenceQueryDocSpec.scala
ktoso: Thanks, that's great feedback @scullxbones - we're not completely happy yet with the API, so all kinds of ideas are very welcome.
Initially we thought that the same API should be implementable from Java as well as Scala, but since we have javadsl and scaladsl separate anyway, maybe we can go ahead and make the Scala API something nice (and make Java "as good as possible").

I have created a new ticket to gather feedback, I'll copy your comment there so that it does not get lost => https://github.com/akka/akka/issues/18229

PR #18081, opened 2015-07-27
Sliding stream implementation

tdhd:
Hi everyone,

this is an implementation of the sliding operator for akka streams. It should be 100% compatible with the standard scala collections implementation.

I will add documentation and clean the code when I have more time.

Any suggestions?


ktoso:
Thanks for the PR! We'll review it shortly :-)

ktoso:
OK TO TEST

tdhd:
Oh, I forgot to add the javadsl.

ktoso:
No worries, add it when you have time – we'll review in the meantime :-)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
```diff
@@ -0,0 +1,79 @@
+/**
+ * Copyright (C) 2014 Typesafe Inc. <http://www.typesafe.com>
```
ktoso: 2015 :)

tdhd:
Added the javadsl. Do I need to add anything else for Java?

tdhd:
It looks like the test case I provided did work. Not sure why the integration test failed.

tdhd:
Among others, this error occurs

```
[ERROR] [07/27/2015 20:51:13.282] [ClientServerSpec-akka.actor.default-dispatcher-5] [akka://ClientServerSpec/system/IO-TCP/selectors/$a/2] Bind failed for TCP channel on endpoint [localhost/127.0.0.1:59129]
java.net.BindException: Address already in use
```

The output of the attached test case
```
[info] FlowSlidingSpec:
Coroner Thread Count starts at 16 in akka.stream.scaladsl.FlowSlidingSpec
[info] Sliding
[info] - must work with n = 3, step = 1 (31 milliseconds)
[info] - must work with n = 3, step = 2 (16 milliseconds)
[info] - must work with n = 3, step = 3 (17 milliseconds)
[info] - must work with n = 2, step = 3 (17 milliseconds)
[info] - must work with n = 2, step = 1 (22 milliseconds)
Coroner Thread Count started at 16, ended at 16, peaked at 21 in akka.stream.scaladsl.FlowSlidingSpec
```

tdhd:
Can you retest it?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -268,6 +268,40 @@ private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immut
 /**
  * INTERNAL API
  */
+private[akka] final case class Sliding[T](n: Int, step: Int) extends PushPullStage[T, immutable.Seq[T]] {
+  private var buf = immutable.Stack.empty[T]
```
ktoso: `Vector` should be a better match here unless I'm missing something.
You can also give it a size hint (equal to `n`) once the previous "chunk" has been signalled to avoid it having to resize.

tdhd: I was picking `Stack` because it has O(1) append complexity. Vector has almost the same but depends on certain assumptions (http://docs.scala-lang.org/overviews/collections/performance-characteristics.html). `Vector` should be also ok, sure!

Also I didn't pick a builder since I don't know if you can append to a vector builder for example. If not wouldn't I have to always reverse the vector when I output it?

ktoso: Indeed, Stack is constant cost, however not very fast in general. Vector may need to resize when appending stuff to it, that's why amortized constant cost. However in the situation we have here we can use a builder, with size hint, and get some excellent performance:

```
[info] Benchmark                           Mode  Cnt       Score       Error   Units
[info] TestBenchmark.stack_1000           thrpt   20       7.446 ±     0.338  ops/ms
[info] TestBenchmark.vector_1000          thrpt   20   19401.495 ±   754.851  ops/ms
[info] TestBenchmark.vector_builder_1000  thrpt   20  142296.432 ± 40669.008  ops/ms
```

https://gist.github.com/ktoso/6f1ed3dc192b24d93f97 (trust no-one, bench everything)

tdhd:
I've changed the implementation to `Vector` with builder. What do you think about it?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+  "Sliding" must {
+    "work with n = 3, step = 1" in assertAllStagesStopped {
+      Source(1 to 6).sliding(n = 3, step = 1).runForeach(testActor ! _) onSuccess {
+        case _ ⇒ testActor ! "done"
+      }
+      expectMsg(Vector(1, 2, 3))
+      expectMsg(Vector(2, 3, 4))
+      expectMsg(Vector(3, 4, 5))
+      expectMsg(Vector(4, 5, 6))
+      //expectMsg(Vector.empty[Int])
```
ktoso: remove this comment line please :) sliding never emits empty collections

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+        case _ ⇒ testActor ! "done"
+      }
+      expectMsg(Vector(1, 2))
+      expectMsg(Vector(2, 3))
+      expectMsg(Vector(3, 4))
+      expectMsg(Vector(4, 5))
+      expectMsg(Vector(5, 6))
+      //expectMsg(Vector.empty[Int])
+      expectMsg("done")
+    }
```
ktoso: can you add a test that `Source.empty.sliding(1)` completes right away (without accidentally yielding a `Vector.empty`)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -268,6 +268,48 @@ private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immut
 /**
  * INTERNAL API
  */
+private[akka] final case class Sliding[T](n: Int, step: Int) extends PushPullStage[T, immutable.Seq[T]] {
+  private val buf = {
+    val b = Vector.newBuilder[T]
+    b.sizeHint(n)
+    b
+  }
```
ktoso: exactly, thanks :-)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -268,6 +268,48 @@ private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immut
...
+
+  override def onPush(elem: T, ctx: Context[immutable.Seq[T]]): SyncDirective = {
+    buf += elem
+    val emit = buf.result()
```
ktoso: Let's avoid this, this way we're generating a lot of `immutable.Vector` instances even though they are not needed. 

Instead let's have a counter (`bufferedElements: Int`) that tracks this information (as the Builder does not expose it).

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -268,6 +268,48 @@ private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immut
...
+    if (emit.size < n) {
+      ctx.pull()
+    } else if (emit.size == n) {
+      ctx.push(emit)
+    } else {
+      buf.clear()
+      buf.sizeHint(n)
```
ktoso: good (sizeHint must be called after a clear)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -268,6 +268,48 @@ private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immut
...
+    b
+  }
+
+  override def onPush(elem: T, ctx: Context[immutable.Seq[T]]): SyncDirective = {
+    buf += elem
+    val emit = buf.result()
+    if (emit.size < n) {
+      ctx.pull()
+    } else if (emit.size == n) {
+      ctx.push(emit)
```
ktoso: why not `buf.clear` after we do the push here? Then the `else` branch can go away

tdhd: `buf.clear` would delete elements which are needed in a possible next step of the window. I don't think its possible to do it as in `Grouped`, right?

ktoso: ah, sorry I forgot about the `step` feature of this operation which requires us to keep around the elements. Looks good!

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -346,6 +346,24 @@ class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Graph
...
+   *
+   * '''Emits when''' enough elements have been collected within the window or upstream completed
+   *
+   * '''Backpressures when''' a window has been assembled and downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def sliding(n: Int, step: Int): javadsl.Flow[In, java.util.List[Out @uncheckedVariance], Mat] =
```
ktoso: needs an overload, like so:

```scala
def sliding(n: Int) = sliding(n, step = 1)
```

because Scala's sliding is: `def sliding[B >: A](size: Int, step: Int = 1): GroupedIterator[B] =`

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -569,6 +569,23 @@ trait FlowOps[+Out, +Mat] {
...
+   *
+   * '''Emits when''' enough elements have been collected within the window or upstream completed
+   *
+   * '''Backpressures when''' a window has been assembled and downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def sliding(n: Int, step: Int): Repr[immutable.Seq[Out], Mat] = andThen(Sliding(n, step))
```
ktoso: Please add default param `step: Int = 1` to be just like Scala's `sliding`

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -346,6 +346,24 @@ class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Graph
...
+    new Flow(delegate.sliding(n, step).map(_.asJava)) // TODO optimize to one step
+
+  /**
```
ktoso: the methods need to be copied for `javadsl/Source.scala` as well

tdhd:
I've noticed the implementation still has a bug in it. I added a test case that demonstrates that and will review the code shortly to fix it.

tdhd:
I fixed the implementation of `sliding` to succeed in another two test cases where `step > n`. To account for that case, the buffer is filled until it reaches `step` elements in it and then drops all elements. In the case where `step >> n` the buffer would get quite large.

I would argue that we can keep the implementation like that because only when `step >> n` the size hint to the vector would be very much off. What do you think?

tdhd:
@ktoso any update on this?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -268,6 +268,62 @@ private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immut
...
+        else ctx.pull()
+      }
+    }
+  }
+
+  override def onPull(ctx: Context[immutable.Seq[T]]): SyncDirective =
+    if (ctx.isFinishing) {
+      val emit = buf.result()
+      if (emit.size == n) {
+        ctx.finish()
```
drewhk: I don't get this, why cannot you emit the result here? Is it because it was already emitted? Please put a comment here

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -268,6 +268,62 @@ private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immut
...
+  var bufferedElements = 0
+
+  override def onPush(elem: T, ctx: Context[immutable.Seq[T]]): SyncDirective = {
+    buf += elem
+    bufferedElements += 1
+    if (bufferedElements < n) {
+      ctx.pull()
+    } else if (bufferedElements == n) {
+      ctx.push(buf.result())
+    } else {
```
drewhk: I don't completely get this branch. When we get here, it is true that !(bufferedElements < n)&&!(bufferedElements == n) which is the same as bufferedElements > n. But I don't see how is that possible, or why should it happen. Maybe a comment?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
```diff
@@ -0,0 +1,100 @@
...
+
+  "Sliding" must {
+    "work with n = 3, step = 1" in assertAllStagesStopped {
+      Source(1 to 6).sliding(n = 3, step = 1).runForeach(testActor ! _) onSuccess {
+        case _ ⇒ testActor ! "done"
+      }
+      expectMsg(Vector(1, 2, 3))
+      expectMsg(Vector(2, 3, 4))
+      expectMsg(Vector(3, 4, 5))
+      expectMsg(Vector(4, 5, 6))
```
drewhk: add an incomplete group (1 to 7 for example)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
```diff
@@ -0,0 +1,100 @@
...
+      expectMsg(Vector(5, 6))
+      expectMsg("done")
+    }
+
+    "work with n = 3, step = 3" in assertAllStagesStopped {
+      Source(1 to 6).sliding(n = 3, step = 3).runForeach(testActor ! _) onSuccess {
+        case _ ⇒ testActor ! "done"
+      }
+      expectMsg(Vector(1, 2, 3))
+      expectMsg(Vector(4, 5, 6))
```
drewhk: Add an incomplete group.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
```diff
@@ -0,0 +1,100 @@
...
+      expectMsg("done")
+    }
+
+    "work with n = 2, step = 3" in assertAllStagesStopped {
+      Source(1 to 6).sliding(n = 2, step = 3).runForeach(testActor ! _) onSuccess {
+        case _ ⇒ testActor ! "done"
+      }
+      expectMsg(Vector(1, 2))
+      expectMsg(Vector(4, 5))
+      expectMsg("done")
```
drewhk: this is nice, shows that 6 is gone.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
```diff
@@ -0,0 +1,100 @@
...
+
+    "work with n = 2, step = 1" in assertAllStagesStopped {
+      Source(1 to 6).sliding(n = 2, step = 1).runForeach(testActor ! _) onSuccess {
+        case _ ⇒ testActor ! "done"
+      }
+      expectMsg(Vector(1, 2))
+      expectMsg(Vector(2, 3))
+      expectMsg(Vector(3, 4))
+      expectMsg(Vector(4, 5))
+      expectMsg(Vector(5, 6))
```
drewhk: add an incomplete group

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    "work with n = 3, step = 4" in assertAllStagesStopped {
+      Source(1 to 12).sliding(n = 3, step = 4).runForeach(testActor ! _) onSuccess {
+        case _ ⇒ testActor ! "done"
+      }
+      expectMsg(Vector(1, 2, 3))
+      expectMsg(Vector(5, 6, 7))
+      expectMsg(Vector(9, 10, 11))
+      expectMsg("done")
```
drewhk: nice one, 12 is dropped properly.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
```diff
@@ -0,0 +1,100 @@
...
+      expectMsg(Vector(1, 2, 3))
+      expectMsg(Vector(7, 8, 9))
+      expectMsg("done")
+    }
+
+    "work with empty sources" in assertAllStagesStopped {
+      Source.empty.sliding(1).runForeach(testActor ! _) onSuccess {
+        case _ ⇒ testActor ! "done"
+      }
+      expectMsg("done")
```
drewhk: maybe add a different step count scenario as well

drewhk:
Looking quite good, will merge after final touchups

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
tdhd: do you mean in `onPull`?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
tdhd: `bufferedElements > n` does happen because in the branch `bufferedElements == n` we don't clear the buffer. This is because the buffer is needed to be kept in case `step > n`.

tdhd:
@drewhk I added more test cases which all succeed. I hope my explanation above is clear.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: aaah, nice one. thanks!

drewhk:
Please squash the commits into one, and use the commit message:

    +str: Sliding window operation

drewhk:
LGTM, will merge after squash

tdhd:
I squashed my commits into the one you suggested. On github it still shows my commits though. It's the first time I squashed commits, so I am a bit unsure if that's the way you wanted it.

ktoso:
Assuming you did `rebase --interactive ...` and now have 1 commit locally, you'll need to `push --force` so that the 3 commits here are overriden by the one commit you have locally. Here is a tutorial on squashing http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

tdhd:
I did `rebase -i <one before my first commit hash>` and squashed everything into one commit. Then I tried to `push` which failed because the branches had diverged. I then did a `pull` and merged the branches. `git log` now shows all of my previous commits and `+str: Sliding window operation` and `Merge branch...`. I guess that when I pulled again and merged I somehow undid the squashing?

ktoso:
> I did `rebase -i <one before my first commit hash>` and squashed everything into one commit. Then I tried to `push` which failed because the branches had diverged. 

At this point you should have `push --force`, since indeed they have diverged "remote has 3 commits", and you want to push the "1 commit" version. :smile: 

> I then did a `pull` and merged the branches. `git log` now shows all of my previous commits and `+str: Sliding window operation` and `Merge branch...`. I guess that when I pulled again and merged I somehow undid the squashing?

ktoso:
You can rebase and squash again and then push :)

tdhd:
Sorry for the confusion but when I try `git rebase -i 98548d10c7ac4d8679d5d2fc0dbcb4eda4df5225` and squash everything into one I get `Could not apply 75550e6dc42a529327b8546094d2a345ae125746... added stream sliding testcase`.

ktoso:
Run a `git rebase --interactive master` :)

tdhd:
It should have worked now. :)

ktoso:
@tdhd indeed, looks good :+1: Awaiting validation then

drewhk:
LGTM

PR #18084, opened 2015-07-28
+doc #17613 document persistence schema evolution

### LINKED ISSUE #17613: doc: Illustrate real serializer for Persistence, including evolution of event types
* some inspiration can be found in `MessageSerializerPersistenceSpec`
* use protobuf, show optional and repeated fields
* https://martin.kleppmann.com/2012/12/05/schema-evolution-in-avro-protocol-buffers-thrift.html
* show evolution of domain events, adding fields, deleting fields, renaming fields, change class names (remove old class)

ktoso:
Will resolve https://github.com/akka/akka/issues/17613

I had a section on "rename event" but it was "either protobuf does it for you, or refer to deleting events" will work a bit more on covering all cases, need a break from this document for a while though.


akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+
+Serializing to human-readable documents
+---------------------------------------
+Some projects may prefer to store events in human-readable formats instead of formats optimised for write/read throughput.
+The style of serialization used in Akka Persistence is completely left up to the concrete Journal plugin implementations -
+this means that some community provided journals may offer human readable serialization, however it is not the default style
+to persist events.
+
+Additionally some datastores are able to store "documents" as JSON (or BSON) natively, wich may have additional benefits
```
ahjohannessen: wich -> *which*

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+(e.g. a JSON type), in the ``EventAdapter`` such that the Journal can directly store it in the datastore without applying
+any additional serialization. This technique is described as sub-section of the :ref:`detach-domain-from-data-model-scala` pattern.
+
+.. note::
+  If in need of human-readable events on the *write-side* of youe application reconsider if preparing materialized views
+  using :ref:`akka-persistence-query-scala` would not be an efficient way to go about this, without compromising the
+  write-side's throughput characteristics.
+
+  If indeed you want to use a human-readable representation on the write-side, pick a Persistence plugin
+  that provides that functionality, or – implenent one yourself.
```
ahjohannessen: implenent -> *implement*

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+The above described technique of detaching the domain model from the data model can be used in interesting ways
+when the journal can be smart about the type of events it receives. For example, let us consider a fictional journal
+plugin implementation that is able to receive ``JSON`` objects, and store them natively in the underlying datastore.
+An example of a datastore which may well implement this pattern is MongoDB, however other databases such as PostgreSQL
+and Cassandra could also do it because of their built-in JSON capabilities.
+
+In this aproach, the :class:`EventAdapter` is used as the marshalling layer: it serializes the events to/from JSON.
+The journal plugin notices that the incoming event type is JSON (for example by performing a ``match`` on the incoming
+event) and
```
ahjohannessen: incomplete sentence

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+.. _rename-field-scala:
+
+Rename fields
+-------------
+
+**Situation:**
+When first designing the system we introduced an ``SeatReverved`` event, it contained many fields, including
+a ``flightCode`` field. After investigating the model, we notice that the flight code does not really relate to the seat
+reserved event, thus we should remove
```
ahjohannessen: This seems incomplete

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
```diff
@@ -0,0 +1,330 @@
...
+    case object Aisle   extends SeatType { override val code = "A" }
+    case object Other   extends SeatType { override val code = "O" }
+    case object Unknown extends SeatType { override val code = ""  }
+
+  }
+  
+  case class SeatReserved(code: String, seatType: SeatType)
+  //#protobuf-read-optional-model
+
+  val proto = """
```
rkuhn: This should not compile: there’s an `object proto` in this class already …

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
```diff
@@ -0,0 +1,330 @@
...
+      case _ => Unknown
+    }
+    case object Window  extends SeatType { override val code = "W" }
+    case object Aisle   extends SeatType { override val code = "A" }
+    case object Other   extends SeatType { override val code = "O" }
+    case object Unknown extends SeatType { override val code = ""  }
+
+  }
+  
+  case class SeatReserved(code: String, seatType: SeatType)
```
rkuhn: what is the relation between this `code` and the `SeatType.code`? such reuse can be confusing

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
```diff
@@ -0,0 +1,330 @@
...
+    case p: Person => s"""${p.name}|${p.surname}""".getBytes(Utf8)
+  }
+
+  // since includeManifest was true, we will get the Person clazz here:
+  def fromBinary(bytes: Array[Byte], clazz: Option[Class[_]]): AnyRef = clazz match {
+    case Some(c) if c == classOf[Person] =>
+      val nameAndSurname = new Predef.String(bytes, Utf8)
+      val Array(name, surname) = nameAndSurname.split("[|]")
+      Person(name, surname)
+  }
```
rkuhn: This means mismatches will produce nondescript MatchError; should probably raise a more descriptive exception.

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
```diff
@@ -0,0 +1,330 @@
...
+object DomainModel {
+  final case class Customer(name: String)
+  final case class Seat(code: String) {
+    def bookFor(customer: Customer): SeatBooked = SeatBooked(code, customer)
+  }
+
+  final case class SeatBooked(code: String, customer: Customer)
+}
+
+/** Data model - imagine highly optimised for schema evolution and persistence */
```
rkuhn: why imagine?

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
+.. _persistence-schema-evolution-scala:
+
+##############################
+Persistence - Schema Evolution
+##############################
+
+When working on long running projects using :ref:`persistence-scala`, or any kind of `Event Sourcing`_ architectures,
+schema evolution becomes one of the more important technical aspects of developing your application since both
+requirements as well as our own understanding of the business domain may (and will) change in time.
```
rkuhn: perhaps split this sentence after “application”; the “both” also disturbs the reading flow a little—perhaps “The requirements will change as well as our understanding …”

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+In fact, if a project matures to the point where you need to evolve its schema to adapt to changing business
+requirements you can view this as first signs of its success – if you wouldn't need to adapt anything over an apps
+lifecycle that could mean that no-one is really using it actively.
+
+In this chapter we will investigate various schema evolution strategies and techniques from which you can pick and
+choose the ones that match your domain and challange at hand.
```
rkuhn: challenge

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+Schema evolution in event-sourced systems
+=========================================
+
+In recent years we have observed a tremendous move towards immutable append-only datastores, with event-sourcing being
+the prime technique successfully being used in these settings. For an excellent overview why and how immutable data makes scalability
+and systems design much simpler you may want to read Pat Helland's excellent `Immutability Changes Everything`_ whitepaper.
+
+Since with `Event Sourcing`_ the **events are immutable** and usually never deleted – the way schema evolution is handled in such
+systems differs from how one would go about it in a mutable database setting (e.g. in typical CRUD database applications).
```
rkuhn: please remove the dash and “in such systems”

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+.. _Immutability Changes Everything: http://www.cidrdb.org/cidr2015/Papers/CIDR15_Paper16.pdf
+.. _Event Sourcing: http://martinfowler.com/eaaDev/EventSourcing.html
+
+Picking the right serialization format
+======================================
+
+Picking the serialization format is one of the most important decisions you will be making in setting up
+your project. It affects which kind of evolutions are simple (or hard) to do, how much work is required to
+add a new datatype and last but not least performance serialization.
```
rkuhn: Maybe mention that while an important decision it can (at a cost) be fixed later, so people shouldn’t be scared too much.

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+Provided default serializers
+----------------------------
+
+Akka Persistence provides `Google Protocol Buffers`_ based serializers (using :ref:`Akka Serialization <serialization-scala>`)
+for it's own message types such as ``PersistentRepr``, ``AtomicWrite`` and snapshots. Journal plugin implementations
+*may* choose to use those provided serializers, or pick a serializer which suits the underlying database more.
```
rkuhn: `s/more/better/` (and perhaps move before “suits”)

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+.. note::
+  Serialization is **NOT** handled automatically by Akka Persistence itself. Instead, it only provides the above described
+  serializers, and in case a ``AsyncWriteJournal`` plugin implementation chooses to use them directly, the above serialization
+  scheme will be used.
+
+  Please refer to your write journal's documentation to learn more how it handles serialization!
+
+  For example, some journals may choose to not use Akka Serialization *at all* and instead store the data in a format
+  that is more "native" for the underlying datastore, e.g. using JSON or some other kind of format that the target
+  datastore understands natively.
```
rkuhn: `s/natively/directly/`

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+The below figure explains how the default serialization scheme works, and how it fits together with serializing the
+user provided message itself, which we will from here on refer to as the ``payload`` (highlighted in yellow):
+
+.. figure:: ../images/persistent-message-envelope.png
+   :align: center
+
+   Akka Persistence provided serializers wrap the user payload in an envelope containing all persistence-relevant information.
+   **The payload will be serialized using the user configured serializer (if none is provided explicitly, Java serialization will be used).**
```
rkuhn: is this really correct? will it not want to store JSON as JSON and not bytes?

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
ktoso: Whoop! :boom:

akka-docs/rst/images/persistence-drop-event-serializer.svg
```diff
@@ -0,0 +1,3 @@
...
...th="63.018734" height="79.897957" stroke="black" stroke-linecap="round" stroke-linejoin="round" stroke-width="1"/><text transform="translate(301.34997 60.91326)" fill="black"><tspan font-family="Helvetica" font-size="12" font-weight="500" x="1.5015545" y="11" textLength="50.015625">Serializer</tspan></text></g><line x1="296.98822" y1="82.87499" x2="254.58112" y2="82.87499" marker-end="url(#FilledArrow_Marker)" stroke="black" stroke-linecap="round" stroke-linejoin="round" stroke-width="1"/><rect x="262.90395" y="70.87499" width="19" height="24" fill="white"/><text transform="translate(267.90395 75.87499)" fill="black"><tspan font-family="Helvetica" font-size="12" font-weight="500" x=".49804688" y="11" textLength="8.0039062">E</tspan></text><line x1="296.98822" y1="60.11223" x2="260.18112" y2="60.11223" marker-end="url(#NonNavigable_Marker)" stroke="black" stroke-linecap="round" stroke-linejoin="round" stroke-width="1"/><rect x="263.40395" y="48.11223" width="18" height="24" fill="white"/><text transform="translate(268.40395 53.11223)" fill="black"><tspan font-family="Helvetica" font-size="12" font-weight="500" x=".33496094" y="11" textLength="7.330078">T</tspan></text></g></g></svg>
```
patriknw: shouldn't it be

    -- O -->
    <-- O --

akka-docs/rst/images/persistence-drop-event.svg
```diff
@@ -0,0 +1,3 @@
...
...="black" stroke-linecap="round" stroke-linejoin="round" stroke-width="1"/><rect x="280.5521" y="70.75" width="19" height="24" fill="white"/><text transform="translate(285.5521 75.75)" fill="black"><tspan font-family="Helvetica" font-size="12" font-weight="500" x=".49804688" y="11" textLength="8.0039062">E</tspan></text><line x1="329.84314" y1="60.612243" x2="260.68112" y2="60.612243" marker-end="url(#NonNavigable_Marker)" stroke="black" stroke-linecap="round" stroke-linejoin="round" stroke-width="1"/><rect x="280.0521" y="48.612243" width="20" height="24" fill="white"/><text transform="translate(285.0521 53.612243)" fill="black"><tspan font-family="Helvetica" font-size="12" font-weight="500" x=".33300781" y="11" textLength="9.3339844">O</tspan></text><line x1="187.09439" y1="84.75" x2="123.08368" y2="84.75" marker-end="url(#FilledArrow_Marker)" stroke="black" stroke-linecap="round" stroke-linejoin="round" stroke-width="1"/><rect x="145.63266" y="72.75" width="19" height="24" fill="white"/><text transform="translate(150.63266 77.75)" fill="black"><tspan font-family="Helvetica" font-size="12" font-weight="500" x=".49804688" y="11" textLength="8.0039062">E</tspan></text></g></g></svg>
```
patriknw: O vs E?

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
```diff
@@ -0,0 +1,330 @@
...
+  //#protobuf-read-optional
+  /**
+   * Example serializer impl which uses protocol buffers generated classes (proto.*)
+   * to perform the to/from binary marshalling.
+   */
+  class AddedFieldsSerializerWithProtobuf extends Serializer {
+    override def identifier = 67876
+    override def includeManifest = true
+
+    override def fromBinary(bytes: Array[Byte], manifest: Option[Class[_]]): AnyRef =
```
patriknw: shouldn't we always promote SerializerWithStringManifest for persistence?

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
```diff
@@ -0,0 +1,330 @@
...
+  def includeManifest: Boolean = true
+
+  def toBinary(obj: AnyRef): Array[Byte] = obj match {
+    case p: Person => s"""${p.name}|${p.surname}""".getBytes(Utf8)
+  }
+
+  // since includeManifest was true, we will get the Person clazz here:
+  def fromBinary(bytes: Array[Byte], clazz: Option[Class[_]]): AnyRef = clazz match {
+    case Some(c) if c == classOf[Person] =>
+      val nameAndSurname = new Predef.String(bytes, Utf8)
```
patriknw: Predef.String ?

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
```diff
@@ -0,0 +1,330 @@
...
+  override def identifier: Int = 8337
+
+  val OldPayloadClassName = "docs.persistence.OldPayload" // class NOT available anymore
+  val MyPayloadClassName = classOf[MyPayload].getName
+
+  override def manifest(o: AnyRef): String = o.getClass.getName
+
+  override def toBinary(o: AnyRef): Array[Byte] = o match {
+    case MyPayload(data) => s"""$data""".getBytes(Utf8)
+    // previously halso handled "old" events here...
```
patriknw: halso?

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+In recent years we have observed a tremendous move towards immutable append-only datastores, with event-sourcing being
+the prime technique successfully being used in these settings. For an excellent overview why and how immutable data makes scalability
+and systems design much simpler you may want to read Pat Helland's excellent `Immutability Changes Everything`_ whitepaper.
+
+Since with `Event Sourcing`_ the **events are immutable** and usually never deleted – the way schema evolution is handled in such
+systems differs from how one would go about it in a mutable database setting (e.g. in typical CRUD database applications).
+The system needs to be able to continue to work in the presence of "old" events which were stored under the "old" schema,
+and we want to limit the complexity to the data layer, exposing a consistent view over all of the events of a given type
+to :class:`PersistentActor`` s and persistence queries, which allows these layers to focus on the business problems instead
+handling the different schemas explicitly in the business logic layers.
```
patriknw: instead of

akka-docs/rst/scala/persistence-schema-evolution.rst
patriknw: clarify that if the journal use the provided protobuf serializers for serializing the wrapper types (e.g. PersistentRepr) the payload will be serialized using the user configured serializer

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+More advanced techniques (e.g. :ref:`remove-event-class-scala`) will dive into using the manifests for increasing the
+flexibility of the persisted vs. exposed types even more, however for now we will focus on the simpler evolution techniques,
+concerning simply configuring the payload serializers.
+
+By default the ``payload`` will be serialized using Java Serialization. This is fine for testing and initial phases
+of your development (while you're still figuring out things and the data will not need to stay persisted forever),
+however once you move to production you really should *pick a different serializer for your payloads*.
+
+.. warning::
+  Do not rely on Java serialization (which will be picked by Akka by default if you don't specify any serializaers)
```
patriknw: serializaers

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+Serializing to human-readable documents
+---------------------------------------
+Some projects may prefer to store events in human-readable formats instead of formats optimised for write/read throughput.
+The style of serialization used in Akka Persistence is completely left up to the concrete Journal plugin implementations -
+this means that some community provided journals may offer human readable serialization, however it is not the default style
+to persist events.
+
+Additionally some datastores are able to store "documents" as JSON (or BSON) natively, wich may have additional benefits
+for later debugging or searchability through the events. In general we highly encourage the use projections using
```
patriknw: strange sentence, perhaps "In general we highly encourage using :ref:`akka-persistence-query-scala` to migrate..."

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+**Situation:**
+You need to add a field to an existing message type. For example, a ``SeatReservation(code: String)`` now
+needs to have an associated code which indicates if it is a window or aisle seat.
+
+**Solution:**
+Adding fields is the most common change you'll need to apply to your messages so make sure the serialization format
+you picked for your payloads can handle it apropriately, i.e. such changes should be *binary compatible*.
+This is easily achieved using the right serializer toolkit – we recommend something like `Google Protocol Buffers`_ or
+`Apache Thrift`_ however other tools may fit your needs just as well – picking a serializer backend is something
```
patriknw: perhaps also include Avro in the recommended?

ktoso: Ok, but then let's add a mini section summarising the trade-offs (like the Will-writeup), since AFAIR renaming a field causes pain in Avro etc.

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: Whoops :) Will complete it soon.

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+on what the user actually intended to change (instead of the composite ``UserDetailsChanged`` that we had in version 1
+of our model).
+
+.. figure:: ../images/persistence-event-adapter-1-n.png
+   :align: center
+
+   The ``EventAdapter`` splits the incoming event into smaller more fine grained events during recovery.
+
+During recovery however, we now need to convert the old ``V1`` model into the ``V2`` representation of the change.
+Depending if the old event contains a name change, we either emit the ``UserNameChanged`` or we don't,
```
patriknw: That is not possible to keep track of without building up state in the event adapter, and that would become very complicated. I think it is reasonable that old events generate both UserNameChanged and AddressChanged no matter what change it actually was.

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+.. includecode:: code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala#string-serializer-skip-deleved-event-by-manifest
+
+The EventAdapter we implemented is aware of ``EventDeserializationSkipped`` events (our "Tombstones"),
+and emits and empty ``EventSeq`` whenever such object is encoutered:
+
+.. includecode:: code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala#string-serializer-skip-deleved-event-by-manifest-adapter
+
+.. _detach-domain-from-data-model-scala:
+
```
patriknw: somewhere in this section it should also be described that it is possible to completely remove the class file, and that the serialization config for that class can be amended to the serializer is only registered for fromBinary (requiring a WithStringManifestSerializer)

patriknw:
great!

ktoso:
Note to self for tomorrow: Maybe explain or link to [stamina](https://github.com/scalapenos/stamina)?

patriknw:
feedback from @agemooij would be great, especially to ensure that stamina could be used together with the things described here (serializers/event adapters)

agemooij:
@patriknw This looks very interesting. I've been swamped with work but I'll try to go through this PR ASAP.

ktoso:
Awesome, thanks in advance!
You may also be interested in the EventAdapters PR itself to get the complete picture: https://github.com/akka/akka/pull/17632

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+lifecycle that could mean that no-one is really using it actively.
+
+In this chapter we will investigate various schema evolution strategies and techniques from which you can pick and
+choose the ones that match your domain and challange at hand.
+
+.. note::
+  This page proposes a number of possible solutions to the schema evolution problem and explains how some of the
+  utilities Akka provides can be used to achieve this, it is by no means a complete (closed) set of solutions.
+
+  Sometimes based on the capabilities of your serialization formats you may be able to evolve your schema in
```
agemooij: Some missing commas: "Sometimes, based on..., you may be able to..."

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+In recent years we have observed a tremendous move towards immutable append-only datastores, with event-sourcing being
+the prime technique successfully being used in these settings. For an excellent overview why and how immutable data makes scalability
+and systems design much simpler you may want to read Pat Helland's excellent `Immutability Changes Everything`_ whitepaper.
+
+Since with `Event Sourcing`_ the **events are immutable** and usually never deleted – the way schema evolution is handled in such
+systems differs from how one would go about it in a mutable database setting (e.g. in typical CRUD database applications).
+The system needs to be able to continue to work in the presence of "old" events which were stored under the "old" schema,
+and we want to limit the complexity to the data layer, exposing a consistent view over all of the events of a given type
+to :class:`PersistentActor`` s and persistence queries, which allows these layers to focus on the business problems instead
```
agemooij: The extra backtick after `PersistentActor`  doesn't seem to belong there.

ktoso: indeed, thanks! :)

akka-docs/rst/scala/persistence-schema-evolution.rst
agemooij: This sentence is very very long indeed and I didn't understand it until I read it three times. Also, is "data layer" the term you guys are using for the individual user-implementations of `PersistentActor`? I would probably call that the "business layer" or the "business logic layer". 

Suggestion:

>The system needs to be able to continue to work in the presence of "old" events which were stored under the "old" schema. We also want to limit complexity in the business logic layer, exposing a consistent view over all of the events of a given type to `PersistentActor`s and persistence queries. This allows the business logic layer to focus on solving business problems instead of having to explicitly deal with different schemas.

akka-docs/rst/scala/persistence-schema-evolution.rst
agemooij: "...a new datatype, and, last but not least, how serialization performs."

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+Akka Persistence provides `Google Protocol Buffers`_ based serializers (using :ref:`Akka Serialization <serialization-scala>`)
+for it's own message types such as ``PersistentRepr``, ``AtomicWrite`` and snapshots. Journal plugin implementations
+*may* choose to use those provided serializers, or pick a serializer which suits the underlying database more.
+
+.. note::
+  Serialization is **NOT** handled automatically by Akka Persistence itself. Instead, it only provides the above described
+  serializers, and in case a ``AsyncWriteJournal`` plugin implementation chooses to use them directly, the above serialization
+  scheme will be used.
+
+  Please refer to your write journal's documentation to learn more how it handles serialization!
```
agemooij: "...learn more **about** how it..."

akka-docs/rst/scala/persistence-schema-evolution.rst
agemooij: I would remove the brackets around "if none is provided ... used" and turn it into its own sentence. This better matches the importance it is given in the paragraphs below.

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+   Akka Persistence provided serializers wrap the user payload in an envelope containing all persistence-relevant information.
+   **The payload will be serialized using the user configured serializer (if none is provided explicitly, Java serialization will be used).**
+
+The blue colored regions of the ``PersistentMessage`` indicate what is serialized using the generated protocol buffers
+serializers, and the yellow payload indicates the user provided event (by calling ``persist(payload)(...)``).
+As you can see, the ``PersistentMessage`` acts as an envelope around the payload, adding various fields related to the
+origin of the event (``persistenceId``, ``sequenceNr`` and more).
+
+More advanced techniques (e.g. :ref:`remove-event-class-scala`) will dive into using the manifests for increasing the
+flexibility of the persisted vs. exposed types even more, however for now we will focus on the simpler evolution techniques,
```
agemooij: "...even more. **However,** for now we will..."

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+concerning simply configuring the payload serializers.
+
+By default the ``payload`` will be serialized using Java Serialization. This is fine for testing and initial phases
+of your development (while you're still figuring out things and the data will not need to stay persisted forever),
+however once you move to production you really should *pick a different serializer for your payloads*.
```
agemooij: "...forever. **However,** once you move to..."

s/really should/should really/

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+.. note::
+  Read the :ref:`Akka Serialization <serialization-scala>` docs to learn more about defining custom serializers,
+  to improve performance and maintainability of your system. Do not depend on Java serialization for production deployments.
+
+The below snippet explains in the minimal amount of lines how a custom serializer can be registered.
+For more in-depth explanations on how serialization picks the serializer to use etc, please refer to its documentation.
+
+First we start by defining our domain model class, here representing a person:
+.. includecode:: code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala#simplest-custom-serializer-model
```
agemooij: This needs a newline I think. The generated html docs have the raw `includecode` statement embedded instead of the actual code.

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+.. includecode:: code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala#simplest-custom-serializer-config
+
+Deserialization will be performed by the same serializer which serialized the message initially
+because of the ``identifier`` being stored together with the message.
+
+For more advanced techniques for using serializers refer to the :ref:`Akka Serialization <serialization-scala>` documentation,
+giving the :ref:`string-manifest-serializer-scala` special attention, as it is tremendously useful for Persistence based applications
+(as we will see in some of the below explained evolutions).
+
```
agemooij: How about

> Please refer to the :ref:`Akka Serialization <serialization-scala>` documentation for more advanced use of serializers, especially the :ref:`string-manifest-serializer-scala` section since it is very useful for Persistence based applications dealing with schema evolutions, as we will see in some of the examples below.

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+Additionally some datastores are able to store "documents" as JSON (or BSON) natively, wich may have additional benefits
+for later debugging or searchability through the events. In general we highly encourage the use projections using
+:ref:`akka-persistence-query-scala` to migrate data stored using a binary format (optimised for write throughput)
+stored on the "write side" of the application, to the "read side" of the application – which more often than not
+is a completely different database, which for example is more suited for storing the events in a human readable format.
+
+One of the possible ways of supporting this is to emit "the type that the journal understands how to store natively"
+(e.g. a JSON type), in the ``EventAdapter`` such that the Journal can directly store it in the datastore without applying
+any additional serialization. This technique is described as sub-section of the :ref:`detach-domain-from-data-model-scala` pattern.
```
agemooij: ...is described **in more detail as part of the** :ref:`detach-domain-from-data-model-scala` pattern.

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+.. note::
+  If in need of human-readable events on the *write-side* of youe application reconsider if preparing materialized views
```
agemooij: ...of **your** application, reconsider **whether** preparing materialized views...

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+.. _remove-event-class-scala:
+
+Remove event class and ignore events
+------------------------------------
+
+**Situation:**
+While investigating app performance you notice that insane amounts of ``CustomerBlinked`` events are being stored
+for every customer each time he/she blinks. Upon investigation you decide that the event does not add any value
+and should be deleted. You still have to be able recover from a journal which contains those old CustomerBlinked events though.
```
agemooij: ...to be able **to** recover...

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+
+**Naive solution - drop events in EventAdapter:**
+
+The problem of removing an event type from the domain model is not as much it's removal, as the implications
```
agemooij: s/it's/its/

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+   :align: center
+
+   The ``EventAdapter`` can drop old events (**O**) by emitting an empty :class:`EventSeq`.
+   Other events can simply be passed through (**E**).
+
+This however does not address the underlying cost of having to deserialize all the events during recovery,
+even those which will be filtered out by the adapter. In the next section we will improve the above explained mechanism
+to avoid deserializing events which would be filtered out by the adapter anyway, thus allowing to save precious time
+during a recovery containing lots of such events (without actually having to delete them).
+
```
agemooij: This section is a bit bare since it  doesn't actually explain what an event adapter is or link to such an explanation. An example would be nice but if the concept of an event adapter is properly explained (or linked to) then the diagram is good enough I think.

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+**Improved solution - deserialize into tombstone:**
+
+In the just described technique we have saved the PersistentActor from receiving un-wanted events by filtering them
+out in the ``EventAdapter``, however the event itself still was deserialized and loaded into memory.
+This has two notable downsides:
+
+  - first, that the deserialization was actually performed, so we spent some of out time budget on the
+    deserialization, event though the event does not contribute anything to the persistent actors state.
```
agemooij: "**even** though"instead of "event though"

akka-docs/rst/scala/persistence-schema-evolution.rst
agemooij: The whole section feels very incomplete. It needs more than the rest of that sentence I think.

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,386 @@
...
+   The ``EventAdapter`` splits the incoming event into smaller more fine grained events during recovery.
+
+During recovery however, we now need to convert the old ``V1`` model into the ``V2`` representation of the change.
+Depending if the old event contains a name change, we either emit the ``UserNameChanged`` or we don't,
+and the address change is handled similarily:
+
+.. includecode:: code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala#split-events-during-recovery
+
+Thanks to returning an ``EventSeq`` from the event adapter the recovered event can be converted to multiple events before
+being delivered to the persistent actor.
```
agemooij: **By** returning an `EventSeq` from the event adapter, the recovered event...

agemooij:
I definitely like the new power introduced with event adapters and the option of writing a String manifest but I'm missing some example that ties it all together for a non-trivial application, including all the custom configuration and boilerplate required.

The biggest problem I see for non-trivial applications is keeping all of this low-level serialization code and configuration organized together and making sure it still works after you make some changes. The boilerplate-heavy and by-name nature of the link between the configuration and the code was one of the main reasons why we decided to built a solution that decoupled the code from the configuration in such a way that the compiler and a testkit can guard against misconfiguration and regressions.

I think you guys are underestimating the number of individual events in a regular Persistence application and how much overhead the provided hardcoded serializers and protobuf examples would bring to the development process. 

My current project has dozens of persistent actors (spread over multiple microservice apps) that each produce between 5 and 25 types of events. Those events will change over time with about 20% of all events being at version 2, 3, or higher after a few months because you want to go into production with an MVP as soon as possible, which means you will usually change significant parts of your system after you ave started producing persistent events.

Keeping this manageable requires a focus on as little boilerplate as possible and as much automated regression testing as possible. Writing the kinds of custom serializers shown in the example sections, with the 1-to-1 mapping between generated protobuf code and Scala case classes and explicit event adapter that keep the old event classes around next to the new ones, is asking quite a lot from your users IMHO. There has to be a simpler way using a lot less boilerplate.

Of course the question is whether Akka should provide that solution or whether it should just provide enough powerful, low-level features for other people to build on top of. That is a discussion we can have another time perhaps but I do think that this documentation should be a bit more aware of the requirements of non-trivial applications and of how the solutions presented scale for such apps.

agemooij:
In the second half of the schema evolution document, it might be good to first describe the different kinds of changes that might happen to your code base over time (i.e. add fields, rename fields, remove fields, split events, remove events, etc.) before describing solution techniques and how/whether they solve which of the core problems.

I think a proper production-ready application needs to have solved all of these so it would be good to first give an overview of the problem as a whole before discussing how to go about solving it using several schema evolution techniques.

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: thanks

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: thanks :)

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: whoops, good catch guys - finishing it up

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
ktoso: changed to the code to "letter and row"

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
ktoso: added `throw` in `case _`

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
ktoso: :)

ktoso: It's a nice song though https://www.youtube.com/watch?v=RwUGSYDKUxU

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: :+1:

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: ko

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: thanks

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: ok

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: Thanks, I'll reword the entire section to explain more explicitly where which serializer is used.

akka-docs/rst/images/persistence-drop-event-serializer.svg
ktoso: Hm, here I mean that "O is old, when it gets out of the journal we reject it; when E gets out we pass it through"

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
ktoso: true, I'll change here

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
ktoso: fixed

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
ktoso: typo, thanks

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: by data layer here I meant "too much complex stuff in the serializer", but I like your wording I'll use it - thanks!

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: right, something weird happened here. fixed

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: Read up on Avro, it's more troublesome for those changes. I've added it and some more hints to the "picking a serialization format" section

ahjohannessen: For a Scala specific project, ScalaPB (https://github.com/trueaccord/ScalaPB) is great for protobuf and we use it with great success for serialization of events :)

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: The code example explains it better I think, it's about if the `UserDetailsChanged(name = null)` case. Should I remove?

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: I can add a link to it in "picking a serializer", opinions @patriknw ?

ahjohannessen: Here is the serializer we use for that, stolen from akka remoting and adjusted for ScalaPB: https://gist.github.com/ahjohannessen/ce43cf45607b9dd9050b

ahjohannessen: ScalaPB reduces *a lot* of boilerplate compared to going with plain protobuf

ahjohannessen: The only difference from the akka serializer is in `toBinary`:
```scala
case message: GeneratedMessage
```
vs. the google type you guys match on.

patriknw: Then you use the generated case classes as the events that you use in the actors?
I think it is beneficial to use different representation for storage and domain, especially when it comes to schema evolution.

ahjohannessen: We use bijections to map from the protobuf generated types to the plain scala case classes. Hence, my happiness for `EventAdapter` :+1:  :)

ktoso: I'm not sure yet about adding too many links in the docs here.
We could add an "add your helpful libs list", like this:

```

Helpful serialization libraries
-------------------------------
This section aims to provide a list of projects which may be helpful to build your serializers with,
by for example providing scala bindings to otherwise very "java style" serialization libraries.

Serialization:
- `ScalaPB`_ - is protobuf compiler plugin that generates Scala sources instead of Java sources like the plain protoc compiler does.

Various:
- `Stamina`_ - is a project aimed to help with schema evolution where the schema for each event is versioned (V1, V2, ...).
```

However I'm slightly worried that it may get outdated soon.

patriknw: I see, makes sense.

patriknw: I agree with you Konrad.

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: done, thanks

akka-docs/rst/scala/persistence-schema-evolution.rst
ahjohannessen: @ktoso @patriknw ScalaPB is definitely worth mentioning, IMO. Does not need to be much more than a gentle push in the right direction.

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: thanks

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: thanks

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: I really like too long sentences, don't I? :wink:

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: ouch, thanks!

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: nice, thanks :)

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: :+1:

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: :+1:

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: :+1:

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: Good catch, I'll link to the eventadapter docs

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: thanks

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: Yes and no. Wouldn't a better place for it be plain serialization docs? 
AFAICS it's a plugin to protoc, i.e. does not affect any evolution patterns, may just be nicer to use from scala. This docs page is all about patterns, trying to not focus too hard on proto.

ahjohannessen: I understand and agree, might not be the proper place for this. Yep, might be a better place in the serialization docs. Just wanted to shed some light on that library as it is really nice when using scala and protobuf :)

ktoso:
Thanks *a lot* for the feedback @agemooij! I added a section introducing the problem space as you suggested before we dive into the solutions. I also de-duplicated some things and addressed all comments in the PR. 

I totally agree that the ultimate goal here is the-least-boilerplate-possible for developers, but for now we'll start out by providing the cutpoints and see how libraries can make use of this. We'll need to ask for the help of the community here I think, for example I'd love to see Stamina develop more and maybe fill in that void for example :-)

akka-docs/rst/scala/index-actors.rst.orig
```diff
@@ -0,0 +1,22 @@
+Actors
```
patriknw: .orig file

akka-docs/rst/scala/persistence.rst
```diff
@@ -634,10 +634,7 @@ indeed adapt it return the adapted event(s) for it, other adapters which do not
 adaptation simply return ``EventSeq.empty``. The adapted events are then delivered in-order to the ``PersistentActor`` during replay.
 
 .. note::
```
patriknw: higher up in this section there is a note: "Complete documentation featuring use-cases and implementation examples for this feature will follow shortly."

Should that be removed now?

akka-docs/rst/scala/persistence.rst
```diff
@@ -634,10 +634,7 @@ indeed adapt it return the adapted event(s) for it, other adapters which do not
...
-  More advanced techniques utilising advanced binary serialization formats such as protocol buffers or kryo / thrift / avro
-  will be documented very soon. These schema evolutions often may need to reach into the serialization layer, however
-  are much more powerful in terms of flexibly removing unused/deprecated classes from your classpath etc.
-
+  For more advanced schema evolution techniques refer to the :ref:`persistence-schema-evolution-scala` documentation.
```
patriknw: perhaps put same note in the section "Custom serialization"?

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,471 @@
...
+
+.. _Immutability Changes Everything: http://www.cidrdb.org/cidr2015/Papers/CIDR15_Paper16.pdf
+.. _Event Sourcing: http://martinfowler.com/eaaDev/EventSourcing.html
+
+Types of schema evolution
+-------------------------
+Before we explain the various techniques that can be used to safely evolve the schema of your persistent events
+over time, we first need to define what the actual problem is, and what the typical styles of changes are.
+
+Since events are never deleted, we need to have a way to be able to recover (read) old events, in such way
```
patriknw: s/recover/replay/

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,471 @@
...
+- :ref:`split event into multiple smaller events <split-large-event-into-smaller-scala>`.
+
+The following sections will explain some patterns which can be used to safely evolve your schema when facing those changes.
+
+Picking the right serialization format
+======================================
+
+Picking the serialization format is a very important decision you will have to make while building your application.
+It affects which kind of evolutions are simple (or hard) to do, how much work is required to add a new datatype, and,
+last but not least, how serialization performance serialization.
```
patriknw: "serialization performance serialization" ?

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,471 @@
...
+
+If you find yourself realising you have picked "the wrong" serialization format, it is always possible to change
+the format used for storing new events, however you would have to keep the old deserialization code in order to
+be able to replay events that were persisted using the old serialization scheme. It is possible to "rebuild"
+an event-log from one serialization format to another one, however it may be a more involved process if you need
+to perform this on a live system.
+
+Binary serialization formats that we have seen work well for long-lived applications include the very flexible IDL based:
+`Google Protobuf`_, `Apache Thrift`_ or `Apache Avro`_ (for which schema evolution is more "entire schema" based, instead of
```
patriknw: instead of putting the Avro remark in () it might be easier to read if was a new sentence starting with "Avro schema evolution is..."

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
```diff
@@ -0,0 +1,419 @@
...
+
+//#simplest-custom-serializer
+/**
+ * Simplest possible serializer, uses a string representation of the Person class.
+ *
+ * Usually a serializer like this would use a library like:
+ * protobuf, kryo, avro, cap'n proto, flatbuffers, SBE or some other dedicated serializer backend
+ * to perform the actual to/from bytes marshalling.
+ */
+class SimplestPossiblePersonSerializer extends Serializer {
```
patriknw: I'd prefer that this sample used SerializerWithStringManifest, because we always recommend that over the other Serializer for Akka Persistence. Users copy paste without reading.
This change can be done in a separate PR.

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,471 @@
...
+This however does not address the underlying cost of having to deserialize all the events during recovery,
+even those which will be filtered out by the adapter. In the next section we will improve the above explained mechanism
+to avoid deserializing events which would be filtered out by the adapter anyway, thus allowing to save precious time
+during a recovery containing lots of such events (without actually having to delete them).
+
+**Improved solution - deserialize into tombstone:**
+
+In the just described technique we have saved the PersistentActor from receiving un-wanted events by filtering them
+out in the ``EventAdapter``, however the event itself still was deserialized and loaded into memory.
+This has two notable _downsides_:
```
patriknw: format _downsides_ ?

akka-docs/rst/scala/persistence-schema-evolution.rst
```diff
@@ -0,0 +1,471 @@
...
+   include additional data for the event (e.g. tags), for ease of later querying.
+
+We will use the following domain and data models to showcase how the separation can be implemented by the adapter:
+
+.. includecode:: code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala#detach-models
+
+The :class:`EventAdapter` takes care of converting from one model to the other one (in both directions),
+alowing the models to be completely detached from each other, such that they can be optimised independently
+as long as the mapping logic is able to convert between them:
+
```
patriknw: Perhaps also mention that same thing can be done in a serializer instead of event adapter if the end result is bytes. E.g. protobuf, the serializer would map between domain model classes and bytes via generated protobuf builders.

patriknw:
LGTM
I read the whole master piece one more time. Awesome! Thanks for all the hard work of pushing it over the finish line. Let's merge this and convert it to Java in a separate PR (that can be done after 2.4-M3).

akka-docs/rst/scala/persistence.rst
ktoso: yes, thanks!

ktoso: Removed this note, and now added this here:

```
.. note::
  More advanced techniques utilising advanced binary serialization as well as human readable formats are documented in
  :ref:`persistence-schema-evolution-scala`. These schema evolutions often may need to reach into the serialization layer,
  however are much more powerful in terms of flexibly removing unused/deprecated classes from your classpath and more.
```

akka-docs/rst/scala/persistence.rst
ktoso: +1

akka-docs/rst/scala/persistence.rst
ktoso: Or rather, 

```
.. note::
  For more advanced schema evolution techniques refer to the :ref:`persistence-schema-evolution-scala` documentation.
```

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: ok

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: whoop

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: ok

akka-docs/rst/scala/code/docs/persistence/PersistenceSchemaEvolutionDocSpec.scala
ktoso: done, good catch

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: Ah, rst... `*downsides*`, thanks

akka-docs/rst/scala/persistence-schema-evolution.rst
ktoso: Indeed, I'll add a sentence on it

ktoso:
Addressed all comments. Not sure about the masterpiece status of it, but I hope it helps users to get thinking and solving these things :wink:

patriknw:
excellent

patriknw:
LOL

ahjohannessen:
Wuhuu :+1: :)

Issue #18076, opened 2015-07-24
Include persistenceId in AtomicWrite


chbatey:
If the AtomicWrite included the ~~partitionId~~ persistenceId it would make it explicit that an AtomicWrite contains entries for a single persistenceId.

Also the min/max seqNr on the AtomicWrite would be useful. Not a biggy but would make my life a little easer when deciding the persistence strategy for Cassandra to best achieve atomicity.

ktoso:
Thanks for the issue Chris! Will look into it soon :-)

scullxbones:
Good call on the ticket.  I'm seeing some of the same complexities re: persistenceId & min/max sequence.  I modeled my 2.4 journal to support N persistenceIds per, forcing a single will simplify my schema quite a bit.  Thank you!

ktoso:
Great to hear some confirmation for the need here, thanks @scullxbones! :-)
Have you seen the PR https://github.com/akka/akka/issues/18085 - should address all the needs right?

scullxbones:
That's where I found this ticket, from the PR.  Looks like exactly what I need - for each AW, the triplet of `persistenceId`, min sequence, max sequence.  So glad I didn't release with the multi-persistenceId-per-AW data model, so complicated. :smile:

Thanks @ktoso!

ktoso:
Great, thanks for chiming in! :-)

PR #18085, opened 2015-07-28
+per #18076 makes AtomicWrite more user friendly

ktoso:
For our intents and purposes, an `AtomicWrite` will always contain messages to one `persistenceId`.
In theory journals could support atomic write to multiple at once, but let's not go there :wink: 

This is a small usability improvement as asked for by @chbatey in https://github.com/akka/akka/issues/18076

One thing I'm not sure about is the `require()` in the constructor... On one hand, I don't expect millions of events there, on the other hand it needs to iterate all of them...

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/Persistent.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/Persistent.scala
@@ -38,6 +38,20 @@ object AtomicWrite {
 }
 
 final case class AtomicWrite(payload: immutable.Seq[PersistentRepr]) extends PersistentEnvelope with Message {
+
+  // only check that all persistenceIds are equal when there's more than one in the Seq
+  if (payload match {
+    case l: List[PersistentRepr]   => l.tail.nonEmpty
+    case v: Vector[PersistentRepr] => v.size > 1
+    case _                         => true // some other collection type, let's just check
+  }) require(payload.forall(_.persistenceId == payload.head.persistenceId),
+      "AtomicWrite must contain messages for the same persistenceId, " +
+        s"yet different persistenceIds found: ${payload.map(_.persistenceId).toSet}")
+
+  def persistenceId = payload.head.persistenceId
+  def lowestSequenceNr = payload.head.sequenceNr // this assumes they're gapless; they should be (it is only our code creating AWs)
+  def highestSequenceNr = payload.last.sequenceNr // TODO: could be optimised, since above require traverses already
+
   override def sender: ActorRef = ActorRef.noSender
   override def size: Int = payload.size
 }
--- a/akka-persistence/src/test/scala/akka/persistence/AtomicWriteSpec.scala
+++ b/akka-persistence/src/test/scala/akka/persistence/AtomicWriteSpec.scala
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.persistence
+
+import org.scalatest.{ Matchers, WordSpec }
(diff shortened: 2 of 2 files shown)


akka-persistence/src/main/scala/akka/persistence/Persistent.scala
```diff
@@ -38,6 +38,15 @@ object AtomicWrite {
 }
 
 final case class AtomicWrite(payload: immutable.Seq[PersistentRepr]) extends PersistentEnvelope with Message {
+
+  // TODO: quite costy... do we need to do this? We only internally create AWs hm...
+  require(!payload.exists(_.persistenceId != payload.head.persistenceId),
+    s"AtomicWrite must contain messages for the same persistenceId, yet different persistenceIds found: ${payload.map(_.persistenceId).toSet}")
```
ktoso: yay or nay?
It's only our internal code creating AtomicWrite's anyway hm...

akka-persistence/src/main/scala/akka/persistence/Persistent.scala
```diff
@@ -38,6 +38,15 @@ object AtomicWrite {
...
+
+  def persistenceId = payload.head.persistenceId
+  def lowestSequenceNr = payload.map(_.sequenceNr).min // this assumes they're gapless; they should be
```
ktoso: again, assumption here that they're gapless - they are because we create them.

Well, not even that, we could `.head.sequenceNr` and `.last.sequenceNr`, if we "trust our impl, and noone else creates those".

ktoso: in theory should also guard against empty seqs... it's a question of how much we trust only us creating these I guess.

ktoso:
// cc @patriknw, as it relates to how we think about the `AtomicWrite`

akka-persistence/src/main/scala/akka/persistence/Persistent.scala
patriknw: I think `head` and `last` is better

akka-persistence/src/main/scala/akka/persistence/Persistent.scala
patriknw: doesn't this read better with `forall( == )`?

patriknw: you can optimize it by adding an if check in front of it

    // do we use List?
    if (payload.tail.nonEmpty) require(...

    // do we use Vector? (would be stupid)
    if (payload.size > 1) require(...

patriknw:
LGTM, after adjusting the `require`

ktoso:
Will amend according to comments and merge :)

ktoso:
Updated PR; quick sanity check please? :)

akka-persistence/src/main/scala/akka/persistence/Persistent.scala
```diff
@@ -38,6 +38,20 @@ object AtomicWrite {
 }
 
 final case class AtomicWrite(payload: immutable.Seq[PersistentRepr]) extends PersistentEnvelope with Message {
+
+  // only check that all persistenceIds are equal when there's more than one in the Seq
+  if (payload match {
+    case l: List[PersistentRepr]   => l.tail.nonEmpty
+    case v: Vector[PersistentRepr] => v.size > 1
```
patriknw: great

ktoso: thanks for the quick review :)

patriknw:
LGTM

PR #18089, opened 2015-07-29
=htp: fix match headers by symbol name

kulikov:
``` scala
val headerName: Symbol = 'Referer
headerName.toString == "'Referer"
```

`'Referer` does not match for all.

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/HeaderDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/HeaderDirectivesSpec.scala
@@ -53,6 +53,68 @@ class HeaderDirectivesSpec extends RoutingSpec with Inside {
     }
   }
 
+  "The headerValueByName directive" should {
+    lazy val route =
+      headerValueByName("Referer") { referer ⇒
+        complete(s"The referer was $referer")
+      }
+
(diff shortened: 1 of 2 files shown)


ktoso:
OK TO TEST

ktoso:
Thanks, LGTM!

ktoso:
// cc @jrudolph @sirthias

jrudolph:
Ha, that bug must have been in there forever. @kulikov, thanks for the fix. 

I wonder: Do you like using symbols with these directives? As this has probably never worked, not even in spray, and no one ever noticed, you may be the only one. So, maybe we should just remove this "feature". On the other hand, we also support symbols for parameters, formFields, and probably also path segments.

ktoso:
I think I'd keep support for symbols here, since it's consistent with other parts of the DSL - people may have their custom headers that they'd like to extract using this convenient style (I like symbols for there things, but maybe I've written too many Scalding jobs ;-)).

sirthias:
This is indeed quite a find, thanks @kulikov!
We should backport this to spray.
Amazing that this never came up before.

ktoso:
Failure was unrelated, merging the fix. 

// failure https://github.com/akka/akka/issues/18099

ktoso:
Refs https://github.com/akka/akka/issues/18100 for tracking what goes into 1.1

PR #18095, opened 2015-07-30
=ben http microbech

2m:
Interesting that using pool is consistently a bit slower.

```
[info] Benchmark                            Mode    Cnt    Score    Error  Units
[info] HttpBenchmark.single_request       sample  61408  653.282 ± 31.117  us/op
[info] HttpBenchmark.single_request_pool  sample  56476  710.526 ± 33.105  us/op
```

### DIFF
--- a/akka-bench-jmh-dev/src/main/scala/akka/http/HttpBenchmark.scala
+++ b/akka-bench-jmh-dev/src/main/scala/akka/http/HttpBenchmark.scala
@@ -1,0 +1,76 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http
+
+import akka.actor.ActorSystem
+import akka.stream.ActorMaterializer
+import akka.stream.scaladsl._
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.Http.ServerBinding
+import akka.http.scaladsl.model._
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.unmarshalling._
+import java.util.concurrent.TimeUnit
+import org.openjdk.jmh.annotations._
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import scala.util.Try
+import com.typesafe.config.ConfigFactory
+
+@State(Scope.Benchmark)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@BenchmarkMode(Array(Mode.SampleTime))
+class HttpBenchmark {
+
+  val config = ConfigFactory.parseString(
+    """
+      akka {
+        loglevel = "ERROR"
+      }""".stripMargin).withFallback(ConfigFactory.load())
+
+  implicit val system = ActorSystem("HttpBenchmark", config)
+  implicit val materializer = ActorMaterializer()
+
+  var binding: ServerBinding = _
+  var request: HttpRequest = _
+  var pool: Flow[(HttpRequest, Int), (Try[HttpResponse], Int), _] = _
+
+  @Setup
+  def setup() = {
+    val route = {
+      path("test") {
+        get {
+          complete("ok")
+        }
+      }
+    }
+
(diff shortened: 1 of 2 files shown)


ktoso:
Good starting point, thanks for creating the project :) LGTM

drewhk:
Can we change it to emit ops/time instead of time/op?

akka-bench-jmh-dev/src/main/scala/akka/http/HttpBenchmark.scala
```diff
@@ -0,0 +1,76 @@
...
+  def shutdown() = {
+    Await.ready(Http().shutdownAllConnectionPools(), 1.second)
+    binding.unbind()
+    system.shutdown()
+    system.awaitTermination()
+  }
+
+  @Benchmark
+  def single_request() = {
+    import scala.concurrent.ExecutionContext.Implicits.global
```
drewhk: why not the system.dispatcher?

akka-bench-jmh-dev/src/main/scala/akka/http/HttpBenchmark.scala
```diff
@@ -0,0 +1,76 @@
...
+    val response = Await.result(Http().singleRequest(request), 1.second)
+    Await.result(Unmarshal(response.entity).to[String], 1.second)
+  }
+
+  @Benchmark
+  def single_request_pool() = {
+    import scala.concurrent.ExecutionContext.Implicits.global
```
drewhk: system.dispatcher?

drewhk:
Good first step for a high-level benchmark.

2m:
`Mode.SampleTime` gives time/op but it also gives percentiles, which are interesting in this case IMO.

```
[info] Result: 704.137 ±(99.9%) 35.044 us/op [Average]
[info]   Statistics: (min, avg, max) = (336.896, 704.137, 55377.920), stdev = 2545.644
[info]   Confidence interval (99.9%): [669.092, 739.181]
[info]   Samples, N = 57140
[info]         mean =    704.137 ±(99.9%) 35.044 us/op
[info]          min =    336.896 us/op
[info]   p( 0.0000) =    336.896 us/op
[info]   p(50.0000) =    410.112 us/op
[info]   p(90.0000) =    447.488 us/op
[info]   p(95.0000) =    472.576 us/op
[info]   p(99.0000) =  13641.155 us/op
[info]   p(99.9000) =  35436.495 us/op
[info]   p(99.9900) =  51614.305 us/op
[info]   p(99.9990) =  55377.920 us/op
[info]   p(99.9999) =  55377.920 us/op
[info]          max =  55377.920 us/op
```

We can add `Mode.Throughput` to benchmark modes to have both time/ops and ops/time.

drewhk:
@sirthias Can you look at this and see if any improvement is possible? Do you think it measures what we intend to measure?

jrudolph:
Nice! Having a stable benchmark setup to run regularly could help catching regression (though, I wonder how well jmh is suited for this not-so-micro benchmark).

I tuned down parallelism to `akka.actor.default-dispatcher.fork-join-executor.parallelism-max = 3`, otherwise running the tests will keep cores busy for no good reason. It doesn't change the results much, though. I get results very similar to the ones reported at the top (running on an i7-3840QM).

In profiles (obtained with jstack, hprof, or perf), I don't see anything unexpected. What comes up at the top are several components of stream materialization. Actual HTTP implementation doesn't seem to use more than ~ 10% of all cycles. So, if we want to tweak something at the HTTP layer it would have to be tweaking or decreasing our use of streams in some way.

The additional use of streams when using `cachedHostConnectionPool` could also explain the additional cost.

A basically single-threaded benchmark like this also contains all the costs of running single-threaded processes with actors (a hand-off through an actor mailbox will likely wake up one thread from deep sleep).

Here's a new flame graph captured during running the `singleRequest` benchmark: http://jrudolph.github.io/single-request-bench-flames.svg (it doesn't show much more than previous ones AFAICS).

drewhk:
@2m we can merge this right?

ktoso:
@2m can we merge here?

PR #18102, opened 2015-07-31
+str Adding Split fan out graph component

DomBlack:
I'm opening this for initial feed back on a possible new Streams cookbook patten.

This commit adds a proof of concept split graph component to the Scala DSL. It is
useful to have a component which can split into separate outlets based on either
a predicate or a function returning an either (think user input validation).

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
@@ -5,11 +5,12 @@ package akka.stream.scaladsl
 
 import akka.stream.impl.Junctions._
 import akka.stream.impl.GenJunctions._
-import akka.stream.impl.Stages.{ MaterializingStageFactory, StageModule }
+import akka.stream.impl.Stages.{ Split, MaterializingStageFactory, StageModule }
 import akka.stream.impl._
 import akka.stream.impl.StreamLayout._
 import akka.stream._
 import Attributes.name
+import akka.stream.scaladsl.FlexiRoute.{ DemandFromAll, RouteLogic }
 import scala.collection.immutable
 import scala.annotation.unchecked.uncheckedVariance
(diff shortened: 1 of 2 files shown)


jroper:
+1, I'm having to implement this pattern several times over right now (using map/broadcast/filter), a built in split would be very useful.

drewhk:
This needs Java API as well.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -316,6 +317,50 @@ class Concat[T] private (override val shape: UniformFanInShape[T, T],
...
 }
 
+object Split {
+  /** Creates an Split using the given function */
+  def apply[In, A, B](f: In ⇒ Either[A, B]): Split[In, A, B] = {
+    val shape = new FanOutShape2[In, A, B]("Split")
+    new Split(shape, new FlexiRouteModule(
+      shape,
+      (s: FanOutShape2[In, A, B]) ⇒ new RouteLogic[In] {
+        override def initialState: State[_] = State[Any](DemandFromAll(s.out0, s.out1)) {
```
drewhk: Hmm, this is a limitation of the current API. This strictly speaking does not necessarily needs demand from all ports, after the predicate has been evaluated for an element.

rkuhn:
We now have much better tooling to support such a stage: GraphStage lets your formulate this in a quite straight-forward fashion without needing to have DemandFromAll semantics.

In other news, FlexiRoute is going away, we are in the process of replacing all of its usage with said GraphStage. Would you be interested in updating this PR or shall we close it?

DomBlack:
I'm sure if I have time right now to look into the newer GraphStage.

Perhaps for now close this PR and I'll reopen it (or a new one) when I get time.

rkuhn:
Okay, sounds good to me. Thanks anyway!

Issue #18115, opened 2015-08-04
Backport permgen fix to 2.3


benmccann:
Can you please backport https://github.com/akka/akka/commit/49e6e7f38cb6e738e839f4763189531a5c0c3351 to the 2.3 branch? Play 2.4 uses Akka 2.3.11 and crashes frequently in development mode with OutOfMemoryError, which we believe we've traced to this issue.

benmccann:
I'm not sure what the difference is between the release-2.3 branch and release-2.3-dev, so I've submitted PRs to backport to both branches

release-2.3: https://github.com/akka/akka/pull/18116
release-2.3-dev: https://github.com/akka/akka/pull/18117

ktoso:
Hi @benmccann,
thanks for the PRs. `release-2.3` is the right one to target this to, as `release-2.3-dev` is only for akka streams and http. Thanks!

benmccann:
Fixed. THanks!

PR #18117, opened 2015-08-04
=rem #17729 Don't use Implicits.global in remoting shutdown

benmccann:


ktoso:
The `release-2.3-dev` branch contains only changes to streams / http, the other PR is OK though - https://github.com/akka/akka/issues/18115 - thanks!

Issue #17694, opened 2015-06-10
Provide automatic aggregation of chunks in parser


jrudolph:
Just an idea I wanted to put out of my head.

By throwing away the additional information of chunk boundaries (if configured) we can aggregate all data the parser currently has available instead of having to emit entity parts in chunk sizes as dictated by the peer.

This was a real issue in my FileUpload example where the SynchronousFileSource produces chunks of only 8kb while the parser on the other side will get the data at chunk sizes determined by the direct buffers of the TCP layer (128k / 8k = 16 times streaming overhead).

/cc @sirthias

jrudolph:
Correction: the conversion would be to `CloseDelimited` or `IndefiniteLength` as the length is not known.

This could also prevent two other unfortunate scenarios:
  - The peer sends only chunks of size 1 leading to additional resource usage due the big number of streaming elements sent around
  - The peer sends big chunks that currently need to be aggregated in the parser. With the automatic conversion incoming data could be dispatched as soon as possible.

One problem with this suggestion is that `CloseDelimited` entities currently aren't allowed for requests.

jrudolph:
To get around the entity type issue the simplest solution would be **not to change** the entity type at all but just aggregate small chunks from the network into bigger chunks coming out of the API. In all cases where the transfer-encoding is just used as that: an implementation detail of transferring data over the connection this aggregation should be fine. If you rely on chunking to carry semantics (e.g. the chosen chunk boundaries or by adding chunk extensions), then you would need to turn aggregation off, but this seems like the less frequent use-case (also cf. the discussion around changing the entity type wrt. compression).

rkuhn:
What does the spec say? Is it not allowed for proxies to re-chunk anyway? If so then I’m all for this proposal.

sirthias:
So far I haven't found anything completely authoritative, but this is what a [section of RFC6202](http://tools.ietf.org/html/rfc6202#section-3.2) says:

> The separation of the response stream into application messages needs
> to be performed at the application level and not at the HTTP
> level.  In particular, it is not possible to use the HTTP chunks
> as application message delimiters, since intermediate proxies
> might "re-chunk" the message stream (for example, by combining
> different chunks into a longer one).

jrudolph:
It would be useful in any case and as @sirthias basically confirms, the default should probably be on aggregating chunks. The question would then be if also the also direction would make sense where half-received chunks are dispatched if the rest is still missing.

rkuhn:
[...]

PR #18144, opened 2015-08-07
=htp: reactivate file-chunking-* settings

aruediger:


ktoso:
OK TO TEST

ktoso:
Thanks for the PR @2beaucoup! We've added you to the jenkins whitelist now, all PRs will automatically be validated without one of us having to OK it :-) Thanks again for the continued stream of great PRs :+1:

drewhk:
This is LGTM to me, @jrudolph?

jrudolph:
For me there are two questions:

 * Does chunking still serve any purpose on streams and if yes, which one? :)
 * If yes, why not piggy back on the chunking configuration of the file source directly.

For me, the previous need for chunking in spray was more a hack around missing proper backpressure support. So, AFAIR we got rid of any chunking directives by purpose.

@2beaucoup, what's your stance on this? Are there use cases I am missing?

2beaucoup:
Thx fpr your comments @jrudolph. That was the input I was looking for when I opened this PR.

I agree, that chunking has no purpose in the context of serving static files.

But what about resources where the content length is either not available (e.g. streaming of live content) or expensive to calculate?

jrudolph:
There are a few things that play together in this topic and which maybe need some consideration:

 1. Do chunks carry any meaning on the HTTP (routing/low-level) API level?
 2. What are the stream-level consequences of chunk sizes?
 3. How often do applications need to tweak chunking? Is it significant enough to warrant a directive?

To give a few answers:

 1. Yes, API chunks are currently visible on the network but only for chunked transfer-encoding. However, we would rather make chunking decisions opaque in the default case to have more flexibility in the implementation as suggested in #17694.
 2. For limiting resources, after-the-fact rechunking cannot have any positive effect. As it stands, adding more stages or cutting smaller chunks will always introduce more work. The only place where you can turn knobs to change resource usage effectively can be at the original source which allocates the original buffers. By configuring these original buffers (like network receive buffers) you choose how much data you let into your system. (Decompressing stages that expand data are "original sources" as well.)
 3. I really don't know if there are "good applications" of chunking for now. Adding a directive now could users give the impression that chunks mattered more than they actually do. So, I'd like to be lazy with introducing something on the HTTP level which may not be so interesting for most users anyway.

2beaucoup:
@jrudolph agreed. Chunking doesn't serve a purpose on the routing level and especially not for files. I'll just remove the leftover settings from the test.

On the other hand maybe we should think about the value of chunking in the context of serving `WithoutKnownLength` request/response entities.

drewhk:
What is the conclusion here? Should I merge?

2beaucoup:
@drewhk no. I'll open another PR that removes the remnants of chunking for now.

jrudolph:
Thanks, @2beaucoup, I think it makes sense to have another discussion before coming to a conclusion.

2beaucoup:
superseded by #18316

PR #18147, opened 2015-08-07
try out an idea for Akka Typed Java API

rkuhn:
(also clean up Inbox.sync since there is not async)

See mostly Counter.scala and CounterTest.scala.


akka-typed/src/test/java/akka/typed/javadsl/Counter.java
```diff
@@ -0,0 +1,37 @@
...
+		}
+	}
+	
+	static final Behavior<State<Counter>> zero = new StateBehavior<>(new C(0));
+
+	public static Behavior<State<Counter>> initial() {
+		return zero;
+	}
+
+}
```
ktoso: I like the idea; All interface/impl in one place and vars safely hidden away... :+1:

rkuhn: There are four drawbacks:

* messages cannot reasonably be printed out since they are lambdas
* messages can be written such that they do complex compositions or other operations within the target Actor’s scope
* messages easily end up being non-serializable
* providing custom serializers or using them for “included” (i.e. captured) values will be tough

Given the amount of code needed to implement this (i.e. all of `State.scala`—four actual lines of code) it is trivial for anyone to put this into their projects, so I’m a little torn between hinting at this for custom implementation or documenting it carefully with the drawbacks. I am particularly worried about the lack of traceability that this will induce.

@pvlugter @bantonsson do you have opinions/insights? @ivantopo @dpsoft ?

akka-typed/src/main/scala/akka/typed/javadsl/State.scala
```diff
@@ -0,0 +1,21 @@
...
+trait State[T] extends Serializable {
+  def apply(in: T): T
+}
+
+class StateBehavior[T](t: T) extends Behavior[State[T]] {
+  def management(ctx: ActorContext[State[T]], msg: Signal): Behavior[State[T]] = {
+    ScalaDSL.Same
+  }
+  def message(ctx: ActorContext[State[T]], msg: State[T]): Behavior[State[T]] = {
+    new StateBehavior(msg(t))
```
rkuhn: This should of course avoid the allocation when possible.

drewhk:
Can we close this and reopen later when it becomes relevant again?

rkuhn:
This PR was to opened to invite comments, which didn’t work, so it has outlived its usefulness.

PR #18150, opened 2015-08-08
Update SnapshotExample.scala

fabiofumarola:
Should the method to terminate the actor system a shutdown and not a terminate?


typesafehub-validator:
Hi @fabiofumarola,

ktoso:
Thanks for the PR, however it's OK as it is, see deprecation note:

```scala
@deprecated("Use the terminate() method instead", "2.4")
def shutdown(): Unit
```

The new `terminate()` is better because it yields a future.

Issue #18160, opened 2015-08-10
FAILED: LightArrayRevolverSchedulerSpec


ktoso:
https://jenkins.akka.io:8498/job/akka-nightly/2449/consoleFull

Split out from https://github.com/akka/akka/issues/18071

ktoso:
The scheduler spec failed tight around the timeout:

```
akka.actor.LightArrayRevolverSchedulerSpec

[info] - must not be affected by long running task *** FAILED *** (5 seconds, 521 milliseconds)
[info]   4.002183008913953 did not equal 4.4 plus or minus 0.3 (SchedulerSpec.scala:254)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.MatchersHelper$.newTestFailedException(MatchersHelper.scala:160)
this was on a0
```

It's a pretty timing sensitive test:

```scala
      val startTime = System.nanoTime
      val n = 22
      val latch = new TestLatch(n)
      system.scheduler.schedule(225.millis, 225.millis) {
        Thread.sleep(80)
        latch.countDown()
      }
      Await.ready(latch, 6.seconds)
      // Rate
      n * 1000.0 / (System.nanoTime - startTime).nanos.toMillis should ===(4.4 +- 0.3)
```

I'd leave as is as it has not reappeared

ktoso:
cc @patriknw

Leave as is or increase tolerance?

2m:
Still failing: https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2976/consoleFull

patriknw:
closing because it has not reoccurred in a long time

jrudolph:
Happened once again:

https://jenkins.akka.io:8498/job/akka-nightly/5203/testReport/junit/akka.actor/LightArrayRevolverSchedulerSpec/A_Scheduler_must_not_be_affected_by_long_running_task/

raboof:
This test is indeed rather timing-sensitive and was already tagged 'timing' - perhaps we want to `-Dakka.test.tags.exclude=timing` on Jenkins?

jrudolph:
But where would they run instead?

patriknw:
obsolete

PR #18164, opened 2015-08-10
=18160 harden SchedulerSpec slightly

ktoso:
Lightly improve spec to be more tolerant against timing problems.
Increased sleep inside such that "if it would affect the tasks" it'd be visible right away
Resolves https://github.com/akka/akka/issues/18160


patriknw:
LGTM

drewhk:
LGTM

Issue #18162, opened 2015-08-10
FAILED: LeveldbPersistentFSMActorSpec


ktoso:
Split out from https://github.com/akka/akka/issues/18071
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2655/consoleFull

// also `a0`

patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2766/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2838/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2844/consoleFull

ktoso:
Related one is: https://github.com/akka/akka/issues/18236#issuecomment-136366239 (closed it as duplicate)

patriknw:
closing because it has not reoccurred in a long time

PR #18166, opened 2015-08-10
=per #18162 harden PersistentFSMActorSpec, less timing sensitive

ktoso:
While the last assertion `within(1.9 seconds, 2.9 seconds)` "makes sense" it's very timing sensitive I'd say... Let's avoid it here;

Resovles https://github.com/akka/akka/issues/18162

### DIFF
--- a/akka-docs/rst/scala/persistence.rst
+++ b/akka-docs/rst/scala/persistence.rst
@@ -655,16 +655,16 @@ adaptation simply return ``EventSeq.empty``. The adapted events are then deliver
 
 Persistent FSM
 ==============
-``PersistentFSMActor`` handles the incoming messages in an FSM like fashion.
+``PersistentFSM`` handles the incoming messages in an FSM like fashion.
 Its internal state is persisted as a sequence of changes, later referred to as domain events.
 Relationship between incoming messages, FSM's states and transitions, persistence of domain events is defined by a DSL.
 
 A Simple Example
 ----------------
-To demonstrate the features of the ``PersistentFSMActor`` trait, consider an actor which represents a Web store customer.
+To demonstrate the features of the ``PersistentFSM`` trait, consider an actor which represents a Web store customer.
 The contract of our "WebStoreCustomerFSMActor" is that it accepts the following commands:
 
-.. includecode:: ../../../akka-persistence/src/test/scala/akka/persistence/fsm/PersistentFSMActorSpec.scala#customer-commands
+.. includecode:: ../../../akka-persistence/src/test/scala/akka/persistence/fsm/PersistentFSMSpec.scala#customer-commands
 
 ``AddItem`` sent when the customer adds an item to a shopping cart
 ``Buy`` - when the customer finishes the purchase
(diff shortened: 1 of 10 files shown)


ktoso:
I'm contemplating how to make the timeout part of the state object... though that's a bit though.
Maybe the test should listen to debug events and notice the timeout value there.

patriknw:
LGTM, I guess

akka-persistence/src/main/scala/akka/persistence/fsm/FSM.scala
```diff
@@ -33,7 +33,12 @@ object FSM {
    * [[akka.actor.FSM.SubscribeTransitionCallBack]] before sending any
    * [[akka.actor.FSM.Transition]] messages.
    */
-  final case class CurrentState[S](fsmRef: ActorRef, state: S)
+  sealed case class CurrentState[S](fsmRef: ActorRef, state: S) {
+    def timeout: Option[FiniteDuration] = None
+    def withTimeout(_timeout: Option[FiniteDuration]) = new CurrentState(fsmRef, state) {
```
drewhk: why not a new field? Binary compat?

ktoso: Yeah, it would break compat since it was a public case class

ktoso: Actually... it's the `persistence.fsm.FSM`, can be breaking then!

patriknw: +1
we have not even released this in anything else than in 2.4 milestones

drewhk:
LGTM

patriknw:
LGTM

akka-docs/rst/scala/persistence.rst
```diff
@@ -655,16 +655,16 @@ adaptation simply return ``EventSeq.empty``. The adapted events are then deliver
 
 Persistent FSM
 ==============
-``PersistentFSMActor`` handles the incoming messages in an FSM like fashion.
+``PersistentFSM`` handles the incoming messages in an FSM like fashion.
```
ktoso: cc @leonidb this makes the classes more consistent with `FSM` (and how we refer to them anyway "persistent fsm" :smile:)

akka-persistence/src/main/java/akka/persistence/fsm/japi/pf/FSMStateFunctionBuilder.java
```diff
@@ -4,7 +4,8 @@
 
 package akka.persistence.fsm.japi.pf;
 
-import akka.persistence.fsm.FSM;
+import akka.persistence.fsm.PersistentFSM;
+import akka.persistence.fsm.PersistentFSMBase;
```
ktoso: users never see the `*Base`

akka-persistence/src/main/scala/akka/persistence/fsm/PersistentFSMBase.scala
```diff
@@ -7,212 +7,10 @@ import akka.actor._
 import akka.japi.pf.{ UnitPFBuilder, UnitMatch, FSMTransitionHandlerBuilder }
 
 import language.implicitConversions
-import scala.annotation.varargs
-import scala.concurrent.duration.Duration
 import scala.collection.mutable
 import akka.routing.{ Deafen, Listen, Listeners }
 import scala.concurrent.duration.FiniteDuration
 
-object FSM {
```
ktoso: This is now all in `object PersistentFSM` to make importing simpler to discover

ktoso:
Added second commit which addresses https://github.com/akka/akka/issues/18168 (as discussed on gitter) 488b1f4d6ddf9e9cc9ab769293979250c5c7d936

patriknw:
rename LGTM

Issue #18053, opened 2015-07-22
FAILED: Diagrams not generated (wat)


ktoso:
Diagram generation failed for scaladoc on nightly, need to check why;

```
java.lang.RuntimeException: ScalaDoc diagrams not generated!
	at scala.sys.package$.error(package.scala:27)
	at akka.Scaladoc$.scaladocVerifier(Doc.scala:72)
	at akka.Scaladoc$$anonfun$projectSettings$5$$anonfun$apply$1.apply(Doc.scala:34)
	at akka.Scaladoc$$anonfun$projectSettings$5$$anonfun$apply$1.apply(Doc.scala:31)
	at scala.Function1$$anonfun$compose$1.apply(Function1.scala:47)
	at sbt.$tilde$greater$$anonfun$$u2219$1.apply(TypeFunctions.scala:40)
	at sbt.std.Transform$$anon$4.work(System.scala:63)
	at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:226)
	at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:226)
	at sbt.ErrorHandling$.wideConvert(ErrorHandling.scala:17)
	at sbt.Execute.work(Execute.scala:235)
	at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:226)
	at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:226)
	at sbt.ConcurrentRestrictions$$anon$4$$anonfun$1.apply(ConcurrentRestrictions.scala:159)
	at sbt.CompletionService$$anon$2.call(CompletionService.scala:28)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
[error] (akka-docs/compile:doc) ScalaDoc diagrams not generated!
```

https://jenkins.akka.io:8498/job/akka-multi-node/1131/console

2m:
This is now failing constantly:
[...]

PR #18167, opened 2015-08-10
=pro 18053 Disable diagram validation in akka-bench

patriknw:


ktoso:
Ah, that's how... Great, LGTM

patriknw:
amended, looks like it is needed in akka-docs also

patriknw:
Refs #18053

Issue #17896, opened 2015-07-02
stashing/unstashing messages in AbstractPersistentView seems broken


michaelom:
I am using akka 2.3.11 with lambda support.

My actor is stashing messages of type X via `stash`, scheduling a tick-message to itself and switching its message handler. Upon receiving the tick-message, my actor calls `unstashAll()` and `context().unbecome()`. I use this technique to wait for a message for a certain amount of time without blocking the actor.

UnstashAll() does exactly what it should do - it prepends all stashed messages to the actor's inbox. Unfortunately all unstashed messages are wrapped in a ReplayedMessage. I was expecting the stashed message type to be preprended to the mailbox (type X). Instead I find a ReplayedMessage.

So in order to get my stashed message I have to call ReplayedMessage.persistent().payload(). I consider this a bug.

Am I using the API in a way it is not designed for? If so, what's the right way to stash und unstash messages in an AbstractPersistentView? Please advise.

ktoso:
Hi @michaelom,
`PersistentView` is going away, we're replacing it with a solution which uses Akka Streams (PR coming any day now). While the View will most likely still be around for a bit (deprecated though), I would not recommend building apps around it.

Having that said, thanks for the report, I think this is most likely a bug.
We have tickets for increasing the coverage of stashing tests in PersistentActor (which shares some of the ideas with PersistentView), so we may well fix it during that ticket: https://github.com/akka/akka/issues/16528

michaelom:
Hi @ktoso 

Thank you for your quick reply.
Looking at your roadmap, PersistentView will be around for some time to come. Working around PersistentView is not really an option right now.

I assume this means no fix for 2.3.x and no fix for 2.4? Can you confirm please.

ktoso:
Persistence changes will not be backported to 2.3, correct. It is an experimental module on 2.3 and unless critical bugs, we won't backport changes there.

patriknw:
We can absolutely fix bugs in PersistentView in 2.4 even though it is phased out (via deprecation). Especially if you can help us track down the issue.

PR #18182, opened 2015-08-11
=per #17896 Support stashing of replayed in PersistentView

patriknw:


patriknw:
Refs #17896

ktoso:
LGTM :)

akka-persistence/src/main/scala/akka/persistence/PersistentView.scala
```diff
@@ -361,6 +361,10 @@ trait PersistentView extends Actor with Snapshotter with Stash with StashFactory
     override def recoveryRunning: Boolean = false
 
     override def stateReceive(receive: Receive, message: Any): Unit = message match {
+      case ReplayedMessage(p) ⇒
+        // we can get ReplayedMessage here if it was stashed by user during replay
+        // unwrap the payload
+        PersistentView.super.aroundReceive(receive, p.payload)
```
drewhk: hmm, is there a valid use case for stashing replayed messages?

patriknw: user asked for it, see #17896, and it is obviously not good to emit the internal `ReplayedMessage` wrapper

drewhk: yes, unwrapping is necessary if we support stashing, otherwise an explicit failure would be needed. It feels like a bit risky to stash replayed messages, but if users need it I trust your judgement to add support for it.

michaelom: >hmm, is there a valid use case for stashing replayed messages?

In our use case we are using a PersistentView to create a (denormalised) read model. Most of your implementations require no stashing - just mapping logic. However, sometimes we join replayed events with some data before writing to the read model. That data might not be available during event consumption (due to eventual consistency, load or a network failure). So we stash all incoming events for a few seconds, then unstash and check for availability of that data again. We repeat that until the data is available.

So basically we can delay event consumption with stash/unstash in a non-blocking fashion.

I am aware this could lead to OOMs but given the right circumstances (low event publishing frequency) it works very well.

patriknw: thanks for clarifying, the bug will be fixed in 2.4-M3

Issue #17894, opened 2015-07-02
Consider supporting Future.succesful(Nil) in asyncWriteMessages


patriknw:
That should make it easier for plugin writers to optimize for the "all good" case.

Remember to change this part of the plugin api docs:
"Note that it is possible to reduce number of allocations by caching some result `Seq` for the happy path, i.e. when no messages are rejected."

PR #18185, opened 2015-08-11
+per #17894 Support Future(Nil) in asyncWriteMessages

patriknw:
Future.successful(Nil) is an alternative way to signal all good
in the happy case, for reduced allocations.

### DIFF
--- a/akka-bench-jmh/build.sbt
+++ b/akka-bench-jmh/build.sbt
@@ -7,3 +7,5 @@ disablePlugins(Unidoc)
 AkkaBuild.defaultSettings
 
 AkkaBuild.dontPublishSettings
+
+Dependencies.benchJmh
--- a/akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
@@ -35,6 +35,7 @@ trait AsyncWriteJournal extends Actor with WriteJournalBase with AsyncRecovery {
       val cctr = resequencerCounter
       resequencerCounter += messages.foldLeft(0)((acc, m) ⇒ acc + m.size) + 1
 
+      val atomicWriteCount = messages.count(_.isInstanceOf[AtomicWrite])
       val prepared = Try(preparePersistentBatch(messages))
       val writeResult = (prepared match {
         case Success(prep) ⇒
@@ -44,7 +45,7 @@ trait AsyncWriteJournal extends Actor with WriteJournalBase with AsyncRecovery {
           // exception from preparePersistentBatch => rejected
           Future.successful(messages.collect { case a: AtomicWrite ⇒ f })
       }).map { results ⇒
-        if (results.size != prepared.get.size)
+        if (results.nonEmpty && results.size != atomicWriteCount)
           throw new IllegalStateException("asyncWriteMessages returned invalid number of results. " +
             s"Expected [${prepared.get.size}], but got [${results.size}]")
         results
@@ -54,7 +55,9 @@ trait AsyncWriteJournal extends Actor with WriteJournalBase with AsyncRecovery {
         case Success(results) ⇒
           resequencer ! Desequenced(WriteMessagesSuccessful, cctr, persistentActor, self)
 
-          val resultsIter = results.iterator
+          val resultsIter =
+            if (results.isEmpty) Iterator.fill(atomicWriteCount)(AsyncWriteJournal.successUnit)
+            else results.iterator
           var n = cctr + 1
           messages.foreach {
             case a: AtomicWrite ⇒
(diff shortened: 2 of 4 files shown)


patriknw:
Refs #17894

akka-persistence/src/main/scala/akka/persistence/journal/leveldb/LeveldbStore.scala
```diff
@@ -42,9 +42,24 @@ private[persistence] trait LeveldbStore extends Actor with WriteJournalBase with
...
+          } catch {
+            case NonFatal(e) ⇒
+              if (result.isEmpty)
+                result = Vector.fill(n)(AsyncWriteJournal.successUnit)
+              result :+= Failure(e)
+          }
+        }
+
+        result
+      }
```
patriknw: I'm not sure we should do this change in the leveldb impl. It's so much uglier.

ktoso: I wonder how much perf impact it has, it's quite a few allocations less.. hm

drewhk: what was the reason for the change?

patriknw: as far as I can see the old impl allocates one Success(Unit) instance per message, while the new impl only allocates one iterator (for the happy path)

patriknw: The purpose with the PR is to support Future(Nil) in asyncWriteMessages so that implementors of journal plugins can use that for optimizations (reduce allocations). It was requested by some (and we also discussed it early).

patriknw: We don't have to change the leveldb impl, and I don't think we should (we don't care about that optimization)

patriknw:
the build failure is real, I'll fix

akka-persistence/src/main/scala/akka/persistence/journal/leveldb/LeveldbStore.scala
patriknw: I have fixed the test failure, and it was actually a [bug in my implementation here](https://github.com/akka/akka/commit/2fc091c33ffc6bd286f3ed9f743b69794a62c81c). Found by tck.

I vote for not changing leveldb impl. Ok @ktoso ?

ktoso: We can leave it out from this PR. We don't recommend LevelDB for anything other than toy projects anyway...

I'm curious how it impacted perf of happy case (raw throughput with persistAsync), since we have it implemented could you run the `PersistentActorThroughputBenchmark` with it before we revert the change?

patriknw: will do

patriknw: you didn't say that it would take 2 x 36 minutes ;-)
I'll do the second run when having dinner or something.

Here is the first run, with the optimized code:

    [info] Benchmark                                                                                    Mode  Cnt       Score      Error  Units
    [info] PersistentActorThroughputBenchmark.actor_normalActor_reply_baseline                         thrpt  200  146964.946 ± 2766.057  ops/s
    [info] PersistentActorThroughputBenchmark.persistentActor_noPersist_reply                          thrpt  200  149247.984 ± 2760.938  ops/s
    [info] PersistentActorThroughputBenchmark.persistentActor_persistAsync_reply                       thrpt  200  107536.480 ± 2585.782  ops/s
    [info] PersistentActorThroughputBenchmark.persistentActor_persistAsync_replyRightOnCommandReceive  thrpt  200  110368.506 ± 2786.160  ops/s
    [info] PersistentActorThroughputBenchmark.persistentActor_persist_reply                            thrpt  200    9601.518 ±   81.373  ops/s

ktoso: lol sorry, that's a lot of runs :d
Looking forward to it :)

// me writing blogpost..

patriknw: here is the second run, with the change here reverted:

    [info] PersistentActorThroughputBenchmark.actor_normalActor_reply_baseline                         thrpt  200  143879.774 ± 2125.070  ops/s
    [info] PersistentActorThroughputBenchmark.persistentActor_noPersist_reply                          thrpt  200  150487.166 ± 2621.352  ops/s
    [info] PersistentActorThroughputBenchmark.persistentActor_persistAsync_reply                       thrpt  200  111451.952 ± 2536.184  ops/s
    [info] PersistentActorThroughputBenchmark.persistentActor_persistAsync_replyRightOnCommandReceive  thrpt  200  114634.707 ± 2604.804  ops/s
    [info] PersistentActorThroughputBenchmark.persistentActor_persist_reply                            thrpt  200    9686.627 ±   36.865  ops/s 

i.e. no difference

ktoso: Interesting... thanks for checking! It's not dominating the work in any case but good to see it's practically no impact.

Ok to remove this optimisation from our Journal then :)

patriknw:
reverted the leveldb change, rebased

ktoso:
LGTM

PR #18186, opened 2015-08-11
=act split Props java api implementation in separate files.

andreaTP:
This require a split into 3 compilation units, but no semantic modifications.

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/AbstractProps.scala
+++ b/akka-actor/src/main/scala/akka/actor/AbstractProps.scala
@@ -1,0 +1,68 @@
+/**
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.actor
+
+import java.lang.reflect.{ Modifier, ParameterizedType, TypeVariable }
+
+import akka.dispatch._
+import akka.japi.Creator
+import akka.routing._
+import akka.util.Reflect
+
+import scala.annotation.varargs
+import scala.language.existentials
+import scala.reflect.ClassTag
+import akka.japi.Util.immutableSeq
+
+/**
+ * Java API: Factory for Props instances.
+ */
+private[akka] trait AbstractProps {
+
+  def validate(clazz: Class[_]) =
+    if (Modifier.isAbstract(clazz.getModifiers))
+      throw new IllegalArgumentException(s"Actor class [${clazz.getName}] must not be abstract")
+
+  /**
+   * Java API: create a Props given a class and its constructor arguments.
+   */
+  @varargs
+  def create(clazz: Class[_], args: AnyRef*): Props = new Props(deploy = Props.defaultDeploy, clazz = clazz, args = args.toList)
+
+  /**
+   * Create new Props from the given [[akka.japi.Creator]].
+   *
+   * You can not use a Java 8 lambda with this method since the generated classes
+   * don't carry enough type information.
+   *
+   * Use the Props.create(actorClass, creator) instead.
+   */
+  def create[T <: Actor](creator: Creator[T]): Props = {
+    val cc = creator.getClass
+    if ((cc.getEnclosingClass ne null) && (cc.getModifiers & Modifier.STATIC) == 0)
+      throw new IllegalArgumentException("cannot use non-static local Creator to create actors; make it static (e.g. local to a static method) or top-level")
+    val ac = classOf[Actor]
+    val coc = classOf[Creator[_]]
+    val actorClass = Reflect.findMarker(cc, coc) match {
+      case t: ParameterizedType ⇒
+        t.getActualTypeArguments.head match {
+          case c: Class[_] ⇒ c // since T <: Actor
+          case v: TypeVariable[_] ⇒
+            v.getBounds collectFirst { case c: Class[_] if ac.isAssignableFrom(c) && c != ac ⇒ c } getOrElse ac
+          case x ⇒ throw new IllegalArgumentException(s"unsupported type found in Creator argument [$x]")
+        }
+      case c: Class[_] if (c == coc) ⇒
+        throw new IllegalArgumentException(s"erased Creator types are unsupported, use Props.create(actorClass, creator) instead")
+    }
+    create(classOf[CreatorConsumer], actorClass, creator)
+  }
+
+  /**
+   * Create new Props from the given [[akka.japi.Creator]] with the type set to the given actorClass.
+   */
+  def create[T <: Actor](actorClass: Class[T], creator: Creator[T]): Props = {
+    create(classOf[CreatorConsumer], actorClass, creator)
+  }
+}
--- a/akka-actor/src/main/scala/akka/actor/Props.scala
+++ b/akka-actor/src/main/scala/akka/actor/Props.scala
@@ -4,13 +4,9 @@
 
 package akka.actor
 
-import java.lang.reflect.{ Constructor, Modifier, ParameterizedType, TypeVariable }
-
 import akka.actor.Deploy.{ NoDispatcherGiven, NoMailboxGiven }
 import akka.dispatch._
-import akka.japi.Creator
 import akka.routing._
-import akka.util.Reflect
 
 import scala.annotation.varargs
 import scala.collection.immutable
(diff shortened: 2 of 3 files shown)


hepin1989:
how about make ```producer``` a private[akka] api?

patriknw:
> how about make producer a private[akka] api?

What do you mean? IndirectActorProducer?

patriknw:
OK TO TEST

hepin1989:
yes

andreaTP:
@hepin1989  I just ported the existing version, I think that the topic goes outside this PR purpose.

ktoso:
Yep, let's keep this PR a simple file move


-- 
Cheers,
Konrad 'ktoso’ Malawski
Akka @ Typesafe

On 11 August 2015 at 16:48:39, Andrea (notifications@github.com) wrote:

@hepin1989 I just ported the existing version, I think that the topic goes outside this PR purpose.

patriknw:
@hepin1989 IndirectActorProducer is important for dependency injection, http://doc.akka.io/docs/akka/2.3.12/java/untyped-actors.html#Dependency_Injection

andreaTP:
ok, able to reproduce, is a problem with Java interop.
Will fix it, have I got to rebase at the end?

andreaTP:
fixed and rebased, ready to re-test.

Sorry.

akka-actor/src/main/scala/akka/actor/Props.scala
```diff
@@ -156,10 +112,9 @@ object Props {
  * }}}
  */
 @SerialVersionUID(2L)
-final case class Props(deploy: Deploy, clazz: Class[_], args: immutable.Seq[Any]) {
+final case class Props(deploy: Deploy = Props.defaultDeploy, clazz: Class[_], args: immutable.Seq[Any]) {
 
-  if (Modifier.isAbstract(clazz.getModifiers))
-    throw new IllegalArgumentException(s"Actor class [${clazz.getName}] must not be abstract")
+  Props.isAbstract(clazz)
```
ktoso: :-1: for making an `isSomething` method throw;

Seems this PR is not only "move" but some new methods?

akka-actor/src/main/scala/akka/actor/AbstractProps.scala
```diff
@@ -0,0 +1,68 @@
...
+import akka.japi.Util.immutableSeq
+
+/**
+ * Java API: Factory for Props instances.
+ */
+trait AbstractProps {
+
+  def isAbstract(clazz: Class[_]) =
+    if (Modifier.isAbstract(clazz.getModifiers))
+      throw new IllegalArgumentException(s"Actor class [${clazz.getName}] must not be abstract")
```
ktoso: "isSomething" feels POJO getter-like, would avoid these method names for validation methods

akka-actor/src/main/scala/akka/actor/Props.scala
andreaTP: ```validate``` is ok to me!

I had to make some really tiny modifications to keep everything working properly, this is not strictly needed, but this is spare code in the body of a class, refactoring it out to a companion object method could be considered good practice, moreover it is in the AbstractProps because is a really JVM dependant check.
Does this sounds good to you?

andreaTP:
Do not know your PRs rules, have I got to squash commits into one or is ok to leave the history of the changes?

ktoso:
The contributing guidelines are written up here: https://github.com/akka/akka/blob/master/CONTRIBUTING.md

Yes, please squash and make the commit name match our style there (`=act split things into more files` (example name))

akka-actor/src/main/scala/akka/actor/Props.scala
```diff
@@ -156,10 +112,9 @@ object Props {
  * }}}
  */
 @SerialVersionUID(2L)
-final case class Props(deploy: Deploy, clazz: Class[_], args: immutable.Seq[Any]) {
+final case class Props(deploy: Deploy = Props.defaultDeploy, clazz: Class[_], args: immutable.Seq[Any]) {
```
ktoso: Why is this needed?
In general adding default params is better in the latter (not initial) parameters

akka-actor/src/main/scala/akka/actor/Props.scala
```diff
@@ -156,10 +112,9 @@ object Props {
...
 
-  if (Modifier.isAbstract(clazz.getModifiers))
-    throw new IllegalArgumentException(s"Actor class [${clazz.getName}] must not be abstract")
+  Props.validate(clazz)
```
ktoso: ok, that's better I guess

akka-actor/src/main/scala/akka/actor/Props.scala
andreaTP: Actually this is no more needed, so I remove this and put the modification into ```AbstractProps```.
will see this into the next commit.

andreaTP:
ok, hope this pr looks better now

akka-actor/src/main/scala/akka/actor/Props.scala
ktoso: great :)

ktoso:
LGTM if validation is happy :)

andreaTP:
PLS BUILD

andreaTP:
Am I supposed to do anything else to get this merged?

ktoso:
I'd like another pair of eyes on it but from my perspective LGTM,
we'll merge soon (currently rushing to release 2.4-M3 and focusing on persistence tickets, sorry for the dalay)

andreaTP:
perfectly ok with this, thanks a lot! 
The question was only because these are my first commits here and I'm not sure on the workflow here :-)

ktoso:
I see, the workflow usually is that we merge after a PR gets "enough" LGTMs, where "enough" may vary on complexity of the PR. This one seems rather simple, yet still there was some small nuances that are now fixed. I'd like a second pair of eyes on it before we merge :)

akka-actor/src/main/scala/akka/actor/AbstractProps.scala
```diff
@@ -0,0 +1,68 @@
...
+import scala.language.existentials
+import scala.reflect.ClassTag
+import akka.japi.Util.immutableSeq
+
+/**
+ * Java API: Factory for Props instances.
+ */
+trait AbstractProps {
+
+  def validate(clazz: Class[_]) =
```
patriknw: why public

akka-actor/src/main/scala/akka/actor/AbstractProps.scala
```diff
@@ -0,0 +1,68 @@
...
+
+import scala.annotation.varargs
+import scala.language.existentials
+import scala.reflect.ClassTag
+import akka.japi.Util.immutableSeq
+
+/**
+ * Java API: Factory for Props instances.
+ */
+trait AbstractProps {
```
patriknw: what is the motivation for splitting it?

akka-actor/src/main/scala/akka/actor/AbstractProps.scala
andreaTP: just because Props is public too, if private works for you I can go for it

ktoso: let's private it :)

akka-actor/src/main/scala/akka/actor/AbstractProps.scala
ktoso: https://github.com/unicredit/akka.js :wink:

akka-actor/src/main/scala/akka/actor/AbstractProps.scala
andreaTP: ok

akka-actor/src/main/scala/akka/actor/AbstractProps.scala
andreaTP: I also believe that code separation jvm dependant / agnostic framework is a good design just in style with this great project.
More projects can benefits from this in the future.

andreaTP:
```private```'d and squashed commits

ktoso:

ktoso:
Superseeded by https://github.com/akka/akka/pull/18238 (thanks!)

PR #18194, opened 2015-08-12
=htc #18044 fail HTTP/1.0 requests with 400 if no default-host-header is set

### LINKED ISSUE #18044: HTTP/1.0 requests without Host header fail noisily with 500
Http 1.0 and 1.1 requests take different code paths if the request is missing a `Host` header. An Http/1.1 request fails in https://github.com/akka/akka/blob/release-2.3-dev/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpRequestParser.scala#L195 while an Http/1.0 request gets through the parser to https://github.com/akka/akka/blob/release-2.3-dev/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala#L71 where it fails uncontrolled and noisily when no `defaultHostHeader` is defined.

It would probably be easiest to move the `defaultHostHeader` processing to the parser as well, so we can fail there in all cases.

See https://groups.google.com/d/topic/akka-user/_fG27EJCZJA/discussion

/cc @2beaucoup @sirthias

jrudolph:
Fixes #18044.

/cc @sirthias @2beaucoup

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -55,10 +55,23 @@ private[http] object HttpServerBluePrint {
       }.withDeploy(Deploy.local)
     }, errorMsg = "Http.serverLayer is currently not reusable. You need to create a new instance for each materialization.")
 
+    def establishAbsoluteUri(requestOutput: RequestOutput): RequestOutput = requestOutput match {
+      case start: RequestStart ⇒
+        try {
+          val effectiveUri = HttpRequest.effectiveUri(start.uri, start.headers, securedConnection = false, defaultHostHeader)
+          start.copy(uri = effectiveUri)
+        } catch {
+          case e: IllegalUriException ⇒
+            MessageStartError(StatusCodes.BadRequest, ErrorInfo("Request is missing required `Host` header", e.getMessage))
+        }
+      case x ⇒ x
+    }
+
     val requestParsingFlow = Flow[ByteString].transform(() ⇒
       // each connection uses a single (private) request parser instance for all its requests
       // which builds a cache of all header instances seen on that connection
       rootParser.createShallowCopy(() ⇒ oneHundredContinueRef).stage).named("rootParser")
+      .map(establishAbsoluteUri)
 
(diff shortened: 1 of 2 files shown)


drewhk:
LGTM

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -55,10 +55,23 @@ private[http] object HttpServerBluePrint {
...
     }, errorMsg = "Http.serverLayer is currently not reusable. You need to create a new instance for each materialization.")
 
+    def establishAbsoluteUri(requestOutput: RequestOutput): RequestOutput = requestOutput match {
+      case start: RequestStart ⇒
+        try {
+          val effectiveUri = HttpRequest.effectiveUri(start.uri, start.headers, securedConnection = false, defaultHostHeader)
+          start.copy(uri = effectiveUri)
+        } catch {
+          case e: IllegalUriException ⇒
+            MessageStartError(StatusCodes.BadRequest, ErrorInfo("Request is missing required `Host` header", e.getMessage))
```
2beaucoup: Shouldn't this be a 5xx in case of HTTP/1.0 requests?

2beaucoup: /cc @jrudolph

jrudolph: It depends on the perspective: if you haven't defined a default host, it could mean that you just don't want to support HTTP/1.0 in which case 400 seems appropriate. Or you see it as a server side misconfiguration leading to a 500 which seems paradox because why would we allow such a misconfiguration in the first place?

2beaucoup: > if you haven't defined a default host, it could mean that you just don't want to support HTTP/1.0

Or it could mean that you simply don't know about this setting. ;) A quick note a la "in order to support HTTP/1.0 you have to specify..." wouldn't hurt I guess.

IMO the msg doesn't help the poor 1.0 client since it doesn't know about host headers. ;)

PR #18199, opened 2015-08-12
=act split ReflectiveDynamicAccess implementation in separate file.

andreaTP:
Split interface and implementation in two files.

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/DynamicAccess.scala
+++ b/akka-actor/src/main/scala/akka/actor/DynamicAccess.scala
@@ -4,7 +4,6 @@
 package akka.actor
 
 import scala.collection.immutable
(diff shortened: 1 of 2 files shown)


ktoso:
OK TO TEST

patriknw:
I appreciate the contributions. We don't strictly use one file per class and here those two are very much related and I think it is better to keep them in same file.

andreaTP:
@patriknw thanks for the feedback, but ReflectiveDynamicAccess is only an implementation of the general interface of DynamicAccess. The fact that is the only one doesn't justify the stickyness to me.
In Akka.js we are providing a completely different implementation and hope this can be detached.

ktoso:

ktoso:
Superseeded by https://github.com/akka/akka/pull/18239 (thanks!)

Issue #16090, opened 2014-10-13
ClusterSharding should fail loudly when no journal is available


ktoso:
When you start `ClusterSharding`, and there's no journal available sharding fill silently fail, as in get stuck after extracting the shard id and looking for the shard home, as in:

```

[DEBUG] [10/13/2014 18:26:29.925] [run-main-0] [EventStream(akka://myapp)] logger log1-Logging$DefaultLogger started
[DEBUG] [10/13/2014 18:26:29.926] [run-main-0] [EventStream(akka://myapp)] Default Loggers started
[INFO] [10/13/2014 18:26:29.996] [run-main-0] [Remoting] Starting remoting
[INFO] [10/13/2014 18:26:30.200] [run-main-0] [Remoting] Remoting started; listening on addresses :[akka.tcp://myapp@127.0.0.1:2551]
[INFO] [10/13/2014 18:26:30.219] [run-main-0] [Cluster(akka://myapp)] Cluster Node [akka.tcp://myapp@127.0.0.1:2551] - Starting up...
[INFO] [10/13/2014 18:26:30.650] [run-main-0] [Cluster(akka://myapp)] Cluster Node [akka.tcp://myapp@127.0.0.1:2551] - Registered cluster JMX MBean [akka:type=Cluster]
[INFO] [10/13/2014 18:26:30.650] [run-main-0] [Cluster(akka://myapp)] Cluster Node [akka.tcp://myapp@127.0.0.1:2551] - Started up successfully
[INFO] [10/13/2014 18:26:30.666] [myapp-akka.actor.default-dispatcher-4] [Cluster(akka://myapp)] Cluster Node [akka.tcp://myapp@127.0.0.1:2551] - Metrics will be retreived from MBeans, and may be incorrect on some platforms. To increase metric accuracy add the 'sigar.jar' to the classpath and the appropriate platform-specific native libary to 'java.library.path'. Reason: java.lang.ClassNotFoundException: org.hyperic.sigar.Sigar
[INFO] [10/13/2014 18:26:30.672] [myapp-akka.actor.default-dispatcher-4] [Cluster(akka://myapp)] Cluster Node [akka.tcp://myapp@127.0.0.1:2551] - Metrics collection has started successfully
[...]

PR #18202, opened 2015-08-13
=per #16090 Fail and warn loudly in SharedLeveldbJournal and ClusterSharding

patriknw:


patriknw:
Refs #16090

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -444,6 +450,9 @@ class ShardRegion(
 
   def register(): Unit = {
     coordinatorSelection.foreach(_ ! registrationMessage)
+    if (shardBuffers.nonEmpty && retryCount >= 5)
+      log.warning("Trying to register to coordinator at [{}], but no acknoledgment. Total [{}] buffered messages.",
+        coordinatorSelection, totalBufferSize)
```
ktoso: :+1:

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -499,12 +517,28 @@ class ShardRegion(
...
             } else {
               val buf = shardBuffers.getOrElse(shardId, Vector.empty)
               shardBuffers = shardBuffers.updated(shardId, buf :+ ((msg, snd)))
+
+              // log some insight to how buffers are filled up every 10% of the buffer capacity
+              val tot = totBufSize + 1
+              if (tot % (bufferSize / 10) == 0) {
+                val logMsg = s"ShardRegion for [$typeName] is using [${100.0 * tot / bufferSize} %] of its buffer capacity."
+                if (tot <= bufferSize / 2)
+                  log.info(logMsg)
```
ktoso: Nice, think that's OK to be on INFO?

akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteProxy.scala
```diff
@@ -27,37 +27,65 @@ private[persistence] trait AsyncWriteProxy extends AsyncWriteJournal with Stash
...
 
   def asyncReadHighestSequenceNr(persistenceId: String, fromSequenceNr: Long): Future[Long] =
-    (store ? ReplayMessages(persistenceId, fromSequenceNr = 0L, toSequenceNr = 0L, max = 0L)).map {
-      case ReplaySuccess(highest) ⇒ highest
+    store match {
+      case Some(s) ⇒
+        (s ? ReplayMessages(persistenceId, fromSequenceNr = 0L, toSequenceNr = 0L, max = 0L)).map {
+          case ReplaySuccess(highest) ⇒ highest
+        }
+      case None ⇒ storeNotInitialized
```
ktoso: yup, great

ktoso:
LGTM

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
patriknw: my thought was to log it at info at 10%, 20%, 30%, 40%, 50%, and then above that as warning
see `if (tot <= bufferSize / 2)`

ktoso: Yeah, I got that. Not sure how noisy it will be (if buf fluctuates a lot around 20% for example).
Let's go with INFO - we can always tune down if someone complains

patriknw: I think all messages will be delivered once it got reply from coordinator, so it goes down to zero.

ktoso: ok, great :)

Issue #15955, opened 2014-09-22
Persistence: SnapshotSelectionCriteria and deleting Snapshots with a lower bound


corradio:
Hi,

This related to Akka persistence.
Why isn't it possible to delete snapshots with a lower bound using SnapshotSelection Criteria (http://doc.akka.io/api/akka/2.3.2/index.html#akka.persistence.SnapshotSelectionCriteria)?
It seems useful to be able to delete snapshots after a certain point, if e.g. a state update function had a bug in a certain version, and we want to replay all messages to recreate snapshots.

Thanks!

ktoso:
Not sure if this should be included in 2.4. Deleting concrete snapshots is always possible with `deleteSnapshot(seqNr)`, needs some discussion

PR #18205, opened 2015-08-13
+per #15955 Add lower bounds to SnapshotSelectionCriteria

patriknw:


patriknw:
Refs #15955

akka-persistence/src/main/scala/akka/persistence/SnapshotProtocol.scala
```diff
@@ -123,6 +135,13 @@ object SnapshotSelectionCriteria {
   /**
    * Java API.
    */
+  def create(maxSequenceNr: Long, maxTimestamp: Long,
+             minSequenceNr: Long, minTimestamp: Long) =
+    SnapshotSelectionCriteria(maxSequenceNr, maxTimestamp, minSequenceNr, minTimestamp)
```
ktoso: Additional Java APIs?

`def create(maxSequenceNr: Long, maxTimestamp: Long) =`?

patriknw: there was already a `def create(maxSequenceNr: Long, maxTimestamp: Long)`
we can't provide all permutations (default values)

ktoso: yeah because all are Long... okey, all good then

ktoso:
LGTM

Issue #15889, opened 2014-09-15
Add TCK test of persistent snapshot with same seqNr


patriknw:
A snapshot journal must be able to handle save of two snapshots with same sequence number, i.e. no persist in-between.

ktoso:
Expected behaviour being overwriting the old snapshot I presume.
Previously our old filebase snapshot store would actually store multiple files with a timestamp,

patriknw:
The behavior could also be ignoring the second attempt, i.e. idempotent. I'm not sure if there are any pros and cons implementation wise for plugins?

PR #18206, opened 2015-08-13
=per #15889 Add tck test for save snapshot with same seqNr

patriknw:


patriknw:
Refs #15889

patriknw:
the test failure was because the snap meta didn't include timestamp in its sorting, fixed

akka-persistence/src/main/scala/akka/persistence/SnapshotProtocol.scala
```diff
@@ -16,6 +16,16 @@ package akka.persistence
...
 //#snapshot-metadata
 
+object SnapshotMetadata {
+  implicit val ordering: Ordering[SnapshotMetadata] = Ordering.fromLessThan[SnapshotMetadata] { (a, b) ⇒
+    if (a eq b) false
+    else if (a.persistenceId != b.persistenceId) a.persistenceId.compareTo(b.persistenceId) < 0
+    else if (a.sequenceNr != b.sequenceNr) a.sequenceNr < b.sequenceNr
+    else if (a.timestamp != b.timestamp) a.timestamp < b.timestamp
+    else false
+  }
```
ktoso: :+1:

akka-persistence-tck/src/main/scala/akka/persistence/snapshot/SnapshotStoreSpec.scala
```diff
@@ -106,5 +106,16 @@ abstract class SnapshotStoreSpec(config: Config) extends PluginSpec(config) {
...
+      val md = metadata(4)
+      snapshotStore.tell(SaveSnapshot(md, s"s-5-modified"), senderProbe.ref)
+      val md2 = senderProbe.expectMsgPF() { case SaveSnapshotSuccess(md2) ⇒ md2 }
+      md2.sequenceNr should be(md.sequenceNr)
+      snapshotStore.tell(LoadSnapshot(pid, SnapshotSelectionCriteria(md.sequenceNr), Long.MaxValue), senderProbe.ref)
+      val result = senderProbe.expectMsgType[LoadSnapshotResult]
+      result.snapshot.get.snapshot should be("s-5-modified")
+      result.snapshot.get.metadata.sequenceNr should be(md.sequenceNr)
+      // metadata timestamp may have been changed
+    }
```
ktoso: I'm missing the `overwrite` part somehow... what am I forgetting?

patriknw: `should be("s-5-modified")` verifies that the snapshot is loaded, previous was `"s-5"`

patriknw: we don't overwrite in the file based store, but semantically a save followed by a load "looks like" an overwrite

ktoso: Ok, now I know what I missed, this:

```scala
def writeSnapshots(): Seq[SnapshotMetadata] = {
    1 to 5 map { i ⇒
      val metadata = SnapshotMetadata(pid, i + 10)
      snapshotStore.tell(SaveSnapshot(metadata, s"s-${i}"), senderProbe.ref)
```

LGTM then :)

ktoso:
LGTM, please merge :)

Issue #18106, opened 2015-08-02
DData Replicator should merge Update result instead of setting it


rkuhn:
Currently the result of the Update’s modify function is

* stored directly into local storage
* possibly sent to others (unless WriteLocal) to be _merged_ into local storage

This is slightly inconsistent in itself, but what is more problematic is that user error can lead to the loss of updates: consider an update function that does not use the previous value and just sets an empty collection or zero counter, then it depends on whether the previous value was already replicated whether it will contribute to the final result or not.

As a general rule, all information added to a data item must always be incorporated by merging—with most of the merges being trivial “fast-forward” ones.

patriknw:
My thinking has been that such a merge is always unnecessary, since the user is supposed to operate on the current value that is passed in as parameter to modify function. I see what you mean with the example of setting it to zero, which will overwrite the local value before it was replicated elsewhere.

I don't mind adding the extra merge.

rkuhn:
Here’s the use-case I had in mind: [MultiMasterCRDT.scala](https://github.com/ReactiveDesignPatterns/CodeSamples/blob/master/chapter12/src/main/scala/com/reactivedesignpatterns/chapter12/MultiMasterCRDT.scala#L74). For a custom CRDT modeling states there is not much point in deriving the new state from the old since they are just singleton objects anyway—sure, it would be possible to add a method that “cancels” any given state, but it is easy to just not use it.

patriknw:
or use `merge` yourself, but I agree that we should add the extra merge

PR #18210, opened 2015-08-14
=cdd #18106 Merge update value with existing

patriknw:


patriknw:
Refs #18106

ktoso:
Makes sense LGTM

Issue #18108, opened 2015-08-02
doc: Change wording about persistent batch given to plugin


ktoso:
As found in recent discussions on the atomicity guarantees etc in the Cassandra plugin: https://github.com/krasserm/akka-persistence-cassandra/pull/69 where @krasserm notes:

>  after thinking more about the specification, I see an issue with the assumption made in the AsyncWriteJournal docs that
>
> > ... the messages of the batch are unrelated
>
> This is actually not the case. From a business logic perspective, later messages in the messages batch may actually depend on earlier messages. For example, later messages in the batch may have been generated during command processing in a PersistentActor whose state is a f(earlier messages).

The wording should be changed somewhat - the events are not completely unrelated, and attempts should be made to store them ordered.

*However*, I don't think this is a hard requirement - the only atomicity guarantee we enforce on all plugins is the AtomicWrite, the batches maybe can not be implemented safely depending on database used. In cassandra it's possible and not too hard to do for example, but in things like mongo it would be a huge pain.

PR #18213, opened 2015-08-14
=per #18108 document that batch contains the same persistenceId events

ktoso:
This documents the status quo of the persistentBatch - the events are all for the same `persistenceId`.
We could entertain the option of loosening this up in the future.

@patriknw do we want to document the status quo or leave it unspecified and reword the docs somehow else? The issue about old wording is the "unrelated" part, which is not entirely true (e2 may have been caused by e1).

Refs #18108

### DIFF
--- a/akka-persistence/src/main/java/akka/persistence/journal/japi/AsyncWritePlugin.java
+++ b/akka-persistence/src/main/java/akka/persistence/journal/japi/AsyncWritePlugin.java
@@ -18,11 +18,12 @@ interface AsyncWritePlugin {
    *
    * The batch is only for performance reasons, i.e. all messages don't have to
    * be written atomically. Higher throughput can typically be achieved by using
-   * batch inserts of many records compared inserting records one-by-one, but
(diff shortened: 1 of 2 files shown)


patriknw:
If we document it like that we can not change it without breaking existing implementations. I don't want to lock us into that. If it is important to document it at all you could say that in current implementation all events come from from same persistenceId but we might change that in a minor version for optimization reasons so a journal implementation must assume that they may come from different persistenceIds

ktoso:
Makes sense – In that case, I think we can simply leave it unspecified. 
I will remove the phrase "unrelated" though as this was the reason for the ticket.

patriknw:
ok

ktoso:
Updated. Please check if wording makes sense

patriknw:
LGTM

patriknw:
wait

patriknw:
no, it's alright

ktoso:
Better safe than sorry :) Thanks!

Issue #18061, opened 2015-07-23
doc: Clarify the interaction of TestActorRef with Persistent actor


giampaolotrapasso:
Please clarify in Testing documentation that using TestActorRef with a PersistentActor (or with a AtLeastOnceDelivery trait mixed-in) leads to an unexpected behavior and must be avoided.

TestActorRef forces the CallingThreadDispatcher, which does not mix well with Actors which need to perform work in the background like the indicated cases. 

This issue comes from this conversation on user list. https://groups.google.com/forum/#!topic/akka-user/r_WW2FUZajo

ktoso:
Hi @giampaolotrapasso,
sounds like a duplicate of https://github.com/akka/akka/issues/15569 which was resolved by https://github.com/akka/akka/pull/17940 already.

Is there something you're missing from that docs update?

giampaolotrapasso:
Sorry, my bad.

ktoso:
No problem, thanks for raising the problem!
In case the added docs don't address your confusion please let us know

giampaolotrapasso:
My humble suggestions: 
- the information is in a light blue box (note), maybe a warning level would make it more visible. (Testing page)
- don't know if it's good to report the warning also in the Persistence page of documentation (Testing section).

ktoso:
Good points; adding it to the testing section of persistence makes a lot of sense - thanks!

PR #18214, opened 2015-08-14
=per,doc #18061 doc that Persistence won't work with TestActorRef

ktoso:
Resolves https://github.com/akka/akka/issues/18061 by:
- making the persistence note in the testing doc a warning
- adding a warning in the persistence docs, section on testing that TestActorRef won't work. 
- and linking to the "right way to do testing" in persistence docs right away


patriknw:
LGTM, excellent!

Issue #18112, opened 2015-08-03
Deleting local snapshots using criteria should take into account the timestamp 


ktoso:
Discovered by Stephen McDonald in the akka-user thread: https://groups.google.com/forum/#!searchin/akka-user/localsnapshotstore$20ignores$20timestamps/akka-user/0CpgVgi0PE8/6gYW5v6zjToJ

> I came across some surprising behaviour with the LocalSnapshotStore - I was trying to implement what I imagine would be a common pattern: remove older snapshots each time a new snapshot is created. I did this with `deleteSnapshots(SnapshotSelectionCriteria(meta.sequenceNr, meta.timestamp - 1))`, only to find it would delete all snapshots for the given persistence ID and sequenceNr, regardless of timestamp given.
>
> After staring at my own code for a very long time, I looked at the source for LocalSnapshotStore and found that it seems to ignore timestamps altogether:
> 
> https://github.com/akka/akka/blob/master/akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala#L161
> 
> I initially thought this to be a bug, but looking at the history it appears to have been intentionally removed:
> 
> https://github.com/akka/akka/commit/63baaf1b2b3d0fc6c1971307fd1e66d74106c968
> 
> I was able to fix this with a minor change in the attached diff. Is there any interest in this fix? Or am I naive to some history around this? 
> 
> I guess if it's the intended behaviour, I can easily enough subclass it and override the relevant bits to take timestamps into account - does that sound reasonable?

Attached diff:

```
diff --git a/akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala b/akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala
index 4dc1ff6..b513816 100644
--- a/akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/snapshot/local/LocalSnapshotStore.scala
@@ -25,6 +25,7 @@ import scala.util._
  * Local filesystem backed snapshot store.
  */
 private[persistence] class LocalSnapshotStore extends SnapshotStore with ActorLogging {
+
   private val FilenamePattern = """^snapshot-(.+)-(\d+)-(\d+)""".r

   import akka.util.Helpers._
[...]

PR #18215, opened 2015-08-14
=per #18112 Use timestamp in deleteSnapshots

patriknw:
It was actually used for finding the right metadata, but the local store
deleted all files with matching seqNr.
Note that we use 0L as undefined value for the timestamp when deleting
single snapshot (and therefore it makes sense to delete all in that case)


patriknw:
Refs #18112

ktoso:
I see, LGTM

PR #18222, opened 2015-08-14
=con #17414 doc mix-in order for PersistentActor and ReceivePipeline

### LINKED ISSUE #17414: doc: ReceivePipeline and PersistentActor don't play nicely
ReceivePipeline must be mixed in before PersistentActor for them to work.

They both override ```aroundReceive``` but whereas ReceivePipeline calls ``super.aroundReceive``, PersistentActor doesn't. 

Consequently this works as expected:
```scala
 class PersistentReplierActor extends PersistentActor with ReceivePipeline {...}
```
but this doesn't:
```scala 
 class PersistentReplierActor extends ReceivePipeline with PersistentActor {...}
```

ktoso:
resolves https://github.com/akka/akka/issues/17414


ktoso:
Added a test as well

ktoso:
cc @patriknw this can make it in todays release as well IMO

patriknw:
LGTM

Issue #18002, opened 2015-07-15
FAILED: FlowTest#mustBeAbleToRecover timing issue


ktoso:
It slightly went over-time in: https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3231/consoleFull // https://github.com/akka/akka/pull/17988

```
rror] Test akka.stream.javadsl.FlowTest.mustBeAbleToRecover failed: java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsg while waiting for 1, took 3.044 sec
[error]     at scala.Predef$.assert(Predef.scala:179)
[error]     at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:338)
[error]     at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:315)
[error]     at akka.testkit.TestKit.expectMsg(TestKit.scala:718)
[error]     at akka.testkit.JavaTestKit.expectMsgEquals(JavaTestKit.java:378)
[error]     at akka.stream.javadsl.FlowTest.mustBeAbleToRecover(FlowTest.java:548)
[error]     ...
```

jrudolph:
This still happens with high probability.

ktoso:
Thanks for reopening // cc @agolubev

agolubev:
Got it. Will check on weekend.

agolubev:
Ok, there are two tests mustBeAbleToRecover. One for Flow and one for Source. So need to fix the second one.

PR #18227, opened 2015-08-16
=str #18002 FlowTest#mustBeAbleToRecover timing issue

agolubev:


agolubev:
Ref #18002

ktoso:
OK TO TEST

drewhk:
Hmm, why did this change fix the issue? What is the underlying timing issue?

agolubev:
Map is throwing exception (on the third element) which is handled as first priority. Consequently, all other messages potentially can be skipped. Fix removes list as source and introduce manual probe to divide good messages from one that throws exception and then recovers.

drewhk:
Hm, but that looks like a bug in recovery. It should not skip elements until the recovery from the previous has been completed.

agolubev:
Flow.recover based on Context.absorbTermination() mechanism which, I believe, stable.
There is a long discussion about treating failure in stream https://github.com/akka/akka/commit/051cf14dd0402848e7cbe7ac60bfc8fd7c46d6ff#commitcomment-8732566
The notion I got is: when failure happens it starts to destroy stream right away. Failure is propagating separately from all other events and can outrun some or all of them. This is exactly the case when recover test failed. The only we can do on failure is to send last event before destroy. 
Per my understanding there is no possibility to send more then one message for recovering, stop destroying stream, recover more then once.

drewhk:
Ah, this is about the recover operation, not about supervision on the stage level. Yes, this is expected behavior then.

drewhk:
LGTM

agolubev:
Thanks

Issue #18234, opened 2015-08-17
Code restructuring to help Akka.js


ktoso:
This ticket outlines our strategy that we'll want to follow to include [akka.js](https://github.com/unicredit/akka.js) needed changes into the akka/akka repository such that we can:
- understand the total impact of these changes before merging into `master`
- keep akka/akka's `master` branch during our `2.4.0-RC#` phase undisturbed by any PRs which are not strictly related to the RC,
- and at the same time do not block the development of akka.js while do so.

As discussed, we do want to help akka.js in the effort, however need to do this without impacting the RC phase of 2.4.0.

The plan is:
- to use the **[akka-js-dev](https://github.com/akka/akka/tree/akka-js-dev) branch on akka/akka** as target for app PRs which are needed for Akka.js to proceed swiftly
- target Akka.js PRs to that branch (can be as one long-running PR, or as multiple ones), in order to get an overview how much impact it has
  - if there is some PR we can not accept, we can safely not merge it there and other ones we can merge ASAP

We commit to:
- review the incoming PRs as they come in (early feedback),
- rebase the `akka-js-dev` branch regularly (or cherry-pick to it) such that it does not get out of sync with `master` too much,
- eventually (most likely after RC phase?) merge in as much of the changes as possible (with the caveat that we may not be able to include all PRs, we'll try to include as much as we can though)

Some references to previous discussions:
- https://github.com/unicredit/akka.js/issues/23 - explains techniques used by akka.js to keep compat
- https://github.com/unicredit/akka.js/issues/14 some good ideas how the impact could be minimised with some sbt magic (in the akka.js projects)

// pinging interested parties: @andreaTP @yawnt @patriknw and FYI @rkuhn (on vacation :tropical_drink: currently :wink:)

andreaTP:
:+1:

yawnt:
looking good! thank you :+1:

ktoso:
I'm happy that you're happy with the plan :smile: 
Let's make it a reality then :-)

ktoso:
**Strategy update:**

- we're going to cherry pick the changes from the akka-js branch onto master (I'll do that)
- akka.js team can issue PRs onto `master` from now on :-)
  - and we'll work on getting them in

Good luck on the presentation @andreaTP, liked the demo, there clearly is a lot of potential there! :-)

andreaTP:
That's great!
:satisfied:

ktoso:
[...]

Issue #17556, opened 2015-05-22
failed: RemoteNodeShutdownAndComesBackSpec


patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2442/consoleFull

    [JVM-1] - must properly reset system message buffer state when new system with same Address comes up *** FAILED *** (14 seconds, 585 milliseconds)
    [JVM-1]   java.lang.AssertionError: assertion failed: timeout (5 seconds) during fishForMessage, hint:
    [JVM-1]   at scala.Predef$.assert(Predef.scala:165)
    [JVM-1]   at akka.testkit.TestKitBase$class.recv$1(TestKit.scala:401)
    [JVM-1]   at akka.testkit.TestKitBase$class.fishForMessage(TestKit.scala:405)
    [JVM-1]   at akka.testkit.TestKit.fishForMessage(TestKit.scala:738)
    [JVM-1]   at akka.remote.RemoteNodeShutdownAndComesBackSpec$$anonfun$1$$anonfun$apply$mcV$sp$1$$anonfun$apply$mcV$sp$2.apply$mcV$sp(RemoteNodeShutdownAndComesBackSpec.scala:121)
    [JVM-1]   at akka.remote.testkit.MultiNodeSpec.runOn(MultiNodeSpec.scala:338)

drewhk:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2495/consoleText

patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly2/10/consoleFull

patriknw:
here is a new one: https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/2901/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/akka-nightly/2366/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3432/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3438/consoleFull

johanandren:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3611/consoleFull

johanandren:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3612/consoleText

patriknw:
[...]

PR #18238, opened 2015-08-17
=act split Props java api implementation in separate files.

andreaTP:


ktoso:

ktoso:
OK TO TEST

akka-actor/src/main/scala/akka/actor/AbstractProps.scala
```diff
@@ -0,0 +1,68 @@
...
+import scala.language.existentials
+import scala.reflect.ClassTag
+import akka.japi.Util.immutableSeq
+
+/**
+ * Java API: Factory for Props instances.
+ */
+private[akka] trait AbstractProps {
+
+  def validate(clazz: Class[_]) =
```
patriknw: `private[akka]` ?

andreaTP: should it be ```private[actor]``` ?

ktoso: `private[akka]` is enough :)

patriknw: note that it should also have the INTERNAL API marker in the Scala doc comment, the reason is that it will be public from Java

patriknw:
LGTM after `private[akka] validate`

ktoso:
Same as Patrik's: LGTM after private[akka] validate

andreaTP:
fixed, tnx

ktoso:
Cool, LGTM

andreaTP:
?

patriknw:
that is currently a flaky test, tracked by #17556

PR #18243, opened 2015-08-17
+act #15819 Create ask() variant that can be used without sender(); Add PromiseRef as an alternative to ask pattern

### LINKED ISSUE #15819: create ask() variant that can be used without sender()
Currently `ask` can only be used with Actors that reply to `sender()` since the PromiseActorRef is not exposed for other usage. We should offer support for both variants:
~~~scala
val p = PromiseActorRef(system, timeout, "name")
target ! Cmd("hello", p)
val f: Future[Any] = p.future
~~~
and
~~~scala
implicit val timeout = ...
val f: Future[Any] = target ? Cmd("hello", _)
~~~

ataraxer:
Closes #15819

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/pattern/AskSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/pattern/AskSpec.scala
@@ -17,6 +17,12 @@ import language.postfixOps
 class AskSpec extends AkkaSpec with ScalaFutures {
 
   "The “ask” pattern" must {
+    "send request to actor and wrap the answer in Future" in {
+      implicit val timeout = Timeout(5.seconds)
+      val echo = system.actorOf(Props(new Actor { def receive = { case x ⇒ sender() ! x } }))
+      val f = echo ? "ping"
+      f.futureValue should ===("ping")
+    }
 
     "return broken promises on DeadLetters" in {
       implicit val timeout = Timeout(5 seconds)
@@ -224,3 +230,4 @@ class AskSpec extends AkkaSpec with ScalaFutures {
   }
 
 }
+
--- a/akka-actor/src/main/scala/akka/pattern/AskSupport.scala
+++ b/akka-actor/src/main/scala/akka/pattern/AskSupport.scala
@@ -127,6 +127,122 @@ trait AskSupport {
     actorSelection.internalAsk(message, timeout, sender)
 }
 
+/**
+ * This object contains implementation details of the “ask” pattern,
+ * which can be combined with "replyTo" pattern.
+ */
+trait ExplicitAskSupport {
+
+  /**
+   * Import this implicit conversion to gain `?` and `ask` methods on
+   * [[akka.actor.ActorRef]], which will defer to the
+   * `ask(actorRef, askSender => message)(timeout)` method defined here.
+   *
+   * {{{
+   * import akka.pattern.ask
+   *
+   * // same as `ask(actor, askSender => Request(askSender))`
+   * val future = actor ? { askSender => Request(askSender) }
+   *
+   * // same as `ask(actor, Request(_))`
+   * val future = actor ? (Request(_))
+   *
+   * // same as `ask(actor, Request(_))(timeout)`
+   * val future = actor ? (Request(_))(timeout)
+   * }}}
+   *
+   * All of the above use a required implicit [[akka.util.Timeout]] and optional implicit
+   * sender [[akka.actor.ActorRef]].
+   */
+  implicit def ask(actorRef: ActorRef): ExplicitlyAskableActorRef = new ExplicitlyAskableActorRef(actorRef)
+
+  /**
+   * Sends a message asynchronously and returns a [[scala.concurrent.Future]]
+   * holding the eventual reply message; this means that the target actor
+   * needs to send the result to the `sender` reference provided. The Future
+   * will be completed with an [[akka.pattern.AskTimeoutException]] after the
+   * given timeout has expired; this is independent from any timeout applied
+   * while awaiting a result for this future (i.e. in
+   * `Await.result(..., timeout)`).
+   *
+   * <b>Warning:</b>
+   * When using future callbacks, inside actors you need to carefully avoid closing over
+   * the containing actor’s object, i.e. do not call methods or access mutable state
+   * on the enclosing actor from within the callback. This would break the actor
+   * encapsulation and may introduce synchronization bugs and race conditions because
+   * the callback will be scheduled concurrently to the enclosing actor. Unfortunately
+   * there is not yet a way to detect these illegal accesses at compile time.
+   *
+   * <b>Recommended usage:</b>
+   *
+   * {{{
+   * val f = ask(worker, replyTo => Request(replyTo))(timeout)
+   * f.map { response =>
+   *   EnrichedMessage(response)
+   * } pipeTo nextActor
+   * }}}
+   */
+  def ask(actorRef: ActorRef, messageFactory: ActorRef ⇒ Any)(implicit timeout: Timeout): Future[Any] =
+    actorRef.internalAsk(messageFactory, timeout, ActorRef.noSender)
+  def ask(actorRef: ActorRef, messageFactory: ActorRef ⇒ Any, sender: ActorRef)(implicit timeout: Timeout): Future[Any] =
+    actorRef.internalAsk(messageFactory, timeout, sender)
+
+  /**
+   * Import this implicit conversion to gain `?` and `ask` methods on
+   * [[akka.actor.ActorSelection]], which will defer to the
+   * `ask(actorSelection, message)(timeout)` method defined here.
+   *
+   * {{{
+   * import akka.pattern.ask
+   *
+   * // same as `ask(selection, askSender => Request(askSender))`
+   * val future = selection ? { askSender => Request(askSender) }
+   *
+   * // same as `ask(selection, Request(_))`
+   * val future = selection ? (Request(_))
+   *
+   * // same as `ask(selection, Request(_))(timeout)`
+   * val future = selection ? (Request(_))(timeout)
+   * }}}
+   *
+   * All of the above use a required implicit [[akka.util.Timeout]] and optional implicit
+   * sender [[akka.actor.ActorRef]].
+   */
+  implicit def ask(actorSelection: ActorSelection): ExplicitlyAskableActorSelection = new ExplicitlyAskableActorSelection(actorSelection)
+
+  /**
+   * Sends a message asynchronously and returns a [[scala.concurrent.Future]]
+   * holding the eventual reply message; this means that the target actor
+   * needs to send the result to the `sender` reference provided. The Future
+   * will be completed with an [[akka.pattern.AskTimeoutException]] after the
+   * given timeout has expired; this is independent from any timeout applied
+   * while awaiting a result for this future (i.e. in
+   * `Await.result(..., timeout)`).
+   *
+   * <b>Warning:</b>
+   * When using future callbacks, inside actors you need to carefully avoid closing over
+   * the containing actor’s object, i.e. do not call methods or access mutable state
+   * on the enclosing actor from within the callback. This would break the actor
+   * encapsulation and may introduce synchronization bugs and race conditions because
+   * the callback will be scheduled concurrently to the enclosing actor. Unfortunately
+   * there is not yet a way to detect these illegal accesses at compile time.
+   *
+   * <b>Recommended usage:</b>
+   *
+   * {{{
+   * val f = ask(worker, replyTo => Request(replyTo))(timeout)
+   * f.map { response =>
+   *   EnrichedMessage(response)
+   * } pipeTo nextActor
+   * }}}
+   *
+   */
+  def ask(actorSelection: ActorSelection, messageFactory: ActorRef ⇒ Any)(implicit timeout: Timeout): Future[Any] =
+    actorSelection.internalAsk(messageFactory, timeout, ActorRef.noSender)
+  def ask(actorSelection: ActorSelection, messageFactory: ActorRef ⇒ Any, sender: ActorRef)(implicit timeout: Timeout): Future[Any] =
+    actorSelection.internalAsk(messageFactory, timeout, sender)
+}
+
 object AskableActorRef {
   /**
    * INTERNAL API: for binary compatibility
@@ -184,6 +300,44 @@ final class AskableActorRef(val actorRef: ActorRef) extends AnyVal {
 
 }
 
+/*
+ * Implementation class of the “ask” with explicit sender pattern enrichment of ActorRef
+ */
+final class ExplicitlyAskableActorRef(val actorRef: ActorRef) extends AnyVal {
+
+  def ask(message: ActorRef ⇒ Any)(implicit timeout: Timeout, sender: ActorRef = Actor.noSender): Future[Any] =
+    internalAsk(message, timeout, sender)
+
+  def ?(message: ActorRef ⇒ Any)(implicit timeout: Timeout, sender: ActorRef = Actor.noSender): Future[Any] =
+    internalAsk(message, timeout, sender)
+
+  /**
+   * INTERNAL API: for binary compatibility
+   */
+  private[pattern] def internalAsk(messageFactory: ActorRef ⇒ Any, timeout: Timeout, sender: ActorRef): Future[Any] = actorRef match {
+    case ref: InternalActorRef if ref.isTerminated ⇒
+      val message = messageFactory(ref.provider.deadLetters)
+      actorRef ! message
+      Future.failed[Any](new AskTimeoutException(s"""Recipient[$actorRef] had already been terminated. Sender[$sender] sent the message of type "${message.getClass.getName}"."""))
+    case ref: InternalActorRef ⇒
+      if (timeout.duration.length <= 0) {
+        val message = messageFactory(ref.provider.deadLetters)
+        Future.failed[Any](new IllegalArgumentException(s"""Timeout length must not be negative, question not sent to [$actorRef]. Sender[$sender] sent the message of type "${message.getClass.getName}"."""))
+      } else {
+        val a = PromiseActorRef(ref.provider, timeout, targetName = actorRef, "unknown", sender)
+        val message = messageFactory(a)
+        a.messageClassName = message.getClass.getName
+        actorRef.tell(message, a)
+        a.result.future
+      }
+    case _ if sender eq null ⇒
+      Future.failed[Any](new IllegalArgumentException(s"""No recipient provided, question not sent to [$actorRef]."""))
+    case _ ⇒
+      val message = messageFactory(sender.asInstanceOf[InternalActorRef].provider.deadLetters)
+      Future.failed[Any](new IllegalArgumentException(s"""Unsupported recipient ActorRef type, question not sent to [$actorRef]. Sender[$sender] sent the message of type "${message.getClass.getName}"."""))
+  }
+}
+
 object AskableActorSelection {
   /**
    * INTERNAL API: for binary compatibility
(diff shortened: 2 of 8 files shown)


ktoso:
OK TO TEST

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
```diff
@@ -207,53 +255,106 @@ final class AskableActorSelection(val actorSel: ActorSelection) extends AnyVal {
...
+
+/**
  * Akka private optimized representation of the temporary actor spawned to
  * receive the reply to an "ask" operation.
  *
  * INTERNAL API
  */
-private[akka] final class PromiseActorRef private (val provider: ActorRefProvider, val result: Promise[Any], _mcn: String)
-  extends MinimalActorRef {
+private[akka] final class PromiseActorRef(val provider: ActorRefProvider)
```
ataraxer: This should probably be renamed to `PromiseRefImpl` for consistency with Akka Typed code and to be less ambiguous with public `PromiseRef`. However I've found mentions of `PromiseActorRef` in documentation and it is also used in public `ActorPromiseActorRef` java API (although I couldn't figure out what's it purpose or how is it supposed to be used) so I've decided to leave it unchanged for now.

ataraxer:
Tests failed with bianry compatibility issues. Reproduced locally:
```
akka > akka-actor/mimaReportBinaryIssues
[info] Resolving com.typesafe#config;1.2.1 ...
[info] akka-actor: found 2 potential binary incompatibilities (filtered 238)
[error]  * method apply(akka.actor.ActorRefProvider,akka.util.Timeout,java.lang.String)akka.pattern.PromiseActorRef in object akka.pattern.PromiseActorRef does not have a correspondent in new version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("akka.pattern.PromiseActorRef.apply")
[error]  * method this(akka.actor.ActorRefProvider,scala.concurrent.Promise)Unit in class akka.pattern.PromiseActorRef does not have a correspondent in new version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("akka.pattern.PromiseActorRef.this")
[trace] Stack trace suppressed: run last akka-actor/*:mimaReportBinaryIssues for the full output.
[error] (akka-actor/*:mimaReportBinaryIssues) akka-actor: Binary compatibility check failed!
[error] Total time: 0 s, completed Aug 18, 2015 10:18:29 AM
```

`PromiseActorRef` is internal API -- added reported issues to exceptions:
```scala
ProblemFilters.exclude[MissingMethodProblem]("akka.pattern.PromiseActorRef.apply"),
ProblemFilters.exclude[MissingMethodProblem]("akka.pattern.PromiseActorRef.this")
```

```
akka > akka-actor/mimaReportBinaryIssues
[info] Resolving com.typesafe#config;1.2.1 ...
[info] akka-actor: found 0 potential binary incompatibilities (filtered 240)
[success] Total time: 1 s, completed Aug 18, 2015 10:24:08 AM
```

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
```diff
@@ -396,29 +497,39 @@ private[akka] final class PromiseActorRef private (val provider: ActorRefProvide
...
   }
 
-  @deprecated("Use apply with messageClassName and sender parameters", "2.4")
-  def apply(provider: ActorRefProvider, timeout: Timeout, targetName: String): PromiseActorRef =
-    apply(provider, timeout, targetName, "unknown", Actor.noSender)
+  private[akka] def apply(provider: ActorRefProvider, timeout: Timeout): PromiseRef = {
+    if (timeout.duration.length > 0) {
+      new PromiseActorRef(provider).setTimeout(timeout)
+    } else {
+      throw new IllegalArgumentException(s"Timeout length must not be negative.")
```
ktoso: Include the timeout value, like: `", was $timeout."`

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
```diff
@@ -73,8 +76,21 @@ trait AskSupport {
...
+
+  /**
+   * A variant of `ask()` which allows to explicitly provide a reference to the actor
+   * which the result needs to be sent to in a message:
+   *
+   * {{{
+   *   ask(worker, respondTo => Message(respondTo))(timeout)
+   * }}}
+   */
+  def ask(actorRef: ActorRef, message: ActorRef ⇒ Any)(implicit timeout: Timeout): Future[Any] =
```
ktoso: This introduces a tricky question about Java API:
- We should provide Java API for it, since on Scala 2.11 it's not nice to use this ask, it would have to be `ask(ActorRef, japi.Function1)`
- Scala 2.12 comes out, and `ActorRef => Any` can now be used from Java 8 as `e -> m`, won't java8 freak out that it's ambiguous which method we wanted - the "java api" (added for Scala 2.11 + JDK7) or the "scala api, which now works with Java 8"?

So... Since 2.4 requires JDK8, Java users will most likely be able to bump from 2.11 to 2.12 since they don't depend on Scala compat (I'd hope), and use the new API via the scala Functions. Summing up - it's best we do not add specific Java API for this method? Unless I missed something.

ataraxer: I think it would be better to provide `ask(ActorRef, japi.Function)` now and address this problem after 2.12 release, so that `japi.Function` replacement with lambdas can be solved in a more uniform way across all of Java APIs. A single case of API which provides lambdas interop would introduce inconsistency, which is worse than potential ambiguity.

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
```diff
@@ -150,36 +180,54 @@ final class AskableActorRef(val actorRef: ActorRef) extends AnyVal {
...
+    internalAsk(_ ⇒ message, timeout, ActorRef.noSender)
 
   def ?(message: Any)(implicit timeout: Timeout, sender: ActorRef = Actor.noSender): Future[Any] =
+    internalAsk(_ ⇒ message, timeout, sender)
+
+  def ?(message: ActorRef ⇒ Any)(implicit timeout: Timeout): Future[Any] =
+    internalAsk(message, timeout, ActorRef.noSender)
+
+  def ?(message: ActorRef ⇒ Any)(implicit timeout: Timeout, sender: ActorRef): Future[Any] =
     internalAsk(message, timeout, sender)
```
ataraxer: Unfortunately it seems that this won't work. We can't provide a single signature `?(ActorRef => Any)(implicit Timeout, ActorRef)`, because multiple overloads with default arguments are not allowed.

On the other hand current implementation fails miserably with "ambiguous reference to overloaded definition", due to erasure, which means that we need to choose to leave either `?(ActorRef => Any)(implicit Timeout, ActorRef)` or `?(ActorRef => Any)(implicit Timeout)`.

First option would require to explicitly specify an implicit sender, adding some cruft to using ask pattern from non actor context (which is one of the primary use cases of the pattern):
```scala
implicit val sender = ActorRef.noSender
target ? (Request(_))
```

Second one will make is impossible to pass a reference to sender.

ataraxer: Other options:
- provide a separate import for this use-case as a `richAsk` or `extendedAsk`;
- choose a different name:
  - `??` -- too close to `???`;
  - `>?<` -- takes longer to write;
  - `=?>` -- welcome to "this is why nobody likes symbolic operators" zone.

This also raises a question of consistency of usage across `akka-actor` and `akka-typed`, considering that the latter already provides a `?(Any => ActorRef)(Timeout)` option.

cc @rkuhn: any thoughts?

ktoso:
FYI, this won't make it into the RC1 release (we're cutting it tomorrow). Let's keep the discussion going though, it's a quite involved feature which we should "get *right*" before pulling in.

hepin1989:
@ktoso This feature would be great if it would be inclouded in akka2.4.x.but I can see that akka team is rushing:)

ktoso:
It can be in `2.4.x`, it just won't make it for the `RC1`.

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
rkuhn: We should not introduce a new symbol—remember `!!` and `!!!`?—and overloading will not work (even in those cases where it does work, type inference will fail for the function literal), so providing it as another import is the only option. Thanks for the exploration!

BTW: the symmetry with `akka-typed` is the reason for doing this, explicit replyTo address style is not only useful in that context but also here—implicit `sender` is sometimes just a tad too hidden.

ataraxer:
OK, I think I'm done for now:
- moved new API to `akka.pattern.extended` package which can aggregate other conflicting pattern APIs in future; this also ensures that the API is backward compatible in case someone uses a wildcard import like `import akka.pattern._`;
- added Java API for both Scala 2.11 and 2.12 options so it can be used immediately after 2.12 release;
- added tests for `ask` and `expanded.ask`;
- added tests for `PromiseRef`;
- cleaned up the docs;

ataraxer:
Ping?

akka-actor-tests/src/test/scala/akka/pattern/AskSpec.scala
```diff
@@ -1,7 +1,8 @@
 /**
  * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
  */
-package akka.pattern
+/* Wrapper package has to be added in order to prevent implicits from clashing. */
+package akka.pattern.wrapper
```
rkuhn: This file does not add implicits, so this comment seems incorrect; otherwise this file would now also be in the wrong source directory.

akka-actor/src/main/scala/akka/pattern/package.scala
```diff
@@ -40,6 +40,51 @@ import akka.actor._
...
+   *
+   * // for Akka compiled with Scala 2.11
+   * ask(actor, new akka.japi.Function<ActorRef, Object> {
+   *   Object apply(ActorRef askSender) {
+   *     return new Request(askSender);
+   *   }
+   * });
+   * }}}
+   */
+  package object extended extends ExplicitAskSupport
```
rkuhn: This should not be a package object and it should also not be within the package object for `akka.pattern`—package objects are strange beasts. I think this should be a normal `object` within the `akka.pattern` package.

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
```diff
@@ -73,9 +192,9 @@ trait AskSupport {
    *
    */
   def ask(actorRef: ActorRef, message: Any)(implicit timeout: Timeout): Future[Any] =
-    actorRef.internalAsk(message, timeout, ActorRef.noSender)
```
rkuhn: these changes should probably not be done: the old ask support was the most efficient solution for that problem, calling internalActorAsk means creating one closure per call that is then not really used (i.e. it only transports the final value anyway)

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
```diff
@@ -150,38 +270,31 @@ final class AskableActorRef(val actorRef: ActorRef) extends AnyVal {
...
-    case ref: InternalActorRef ⇒
-      if (timeout.duration.length <= 0)
-        Future.failed[Any](new IllegalArgumentException(s"""Timeout length must not be negative, question not sent to [$actorRef]. Sender[$sender] sent the message of type "${message.getClass.getName}"."""))
-      else {
-        val a = PromiseActorRef(ref.provider, timeout, targetName = actorRef, message.getClass.getName, sender)
-        actorRef.tell(message, a)
-        a.result.future
-      }
-    case _ ⇒ Future.failed[Any](new IllegalArgumentException(s"""Unsupported recipient ActorRef type, question not sent to [$actorRef]. Sender[$sender] sent the message of type "${message.getClass.getName}"."""))
-  }
```
rkuhn: This method must remain: it is needed for binary compatibility.

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
```diff
@@ -189,13 +302,13 @@ object AskableActorSelection {
    * INTERNAL API: for binary compatibility
    */
   private[pattern] def ask$extension(actorSel: ActorSelection, message: Any, timeout: Timeout): Future[Any] =
-    actorSel.internalAsk(message, timeout, ActorRef.noSender)
```
rkuhn: these changes should also be removed

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
```diff
@@ -207,53 +320,153 @@ final class AskableActorSelection(val actorSel: ActorSelection) extends AnyVal {
...
+}
+
+/*
+ * Implementation class of the “ask” pattern with explicit sender enrichment of ActorSelection
+ */
+final class ExplicitlyAskableActorSelection(val actorSel: ActorSelection) extends AnyVal {
+
+  /**
+   * INTERNAL API: for binary compatibility
+   */
```
rkuhn: for new classes there can be no BC concerns: only add those methods that are really needed and comment them meaningfully

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
```diff
@@ -207,53 +320,153 @@ final class AskableActorSelection(val actorSel: ActorSelection) extends AnyVal {
...
+    internalSelectionAsk(actorSel, messageFactory, timeout, ActorRef.noSender)
+
+  def ?(messageFactory: ActorRef ⇒ Any)(implicit timeout: Timeout, sender: ActorRef = Actor.noSender): Future[Any] =
+    internalSelectionAsk(actorSel, messageFactory, timeout, sender)
+}
+
+/**
+ * INTERNAL API: for binary compatibility
+ */
+private[pattern] object InternalAsk {
```
rkuhn: the methods in here should be moved into the ExplicitlyAskableActorRef/Selection, respectively

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
```diff
@@ -207,53 +320,153 @@ final class AskableActorSelection(val actorSel: ActorSelection) extends AnyVal {
...
+  def internalActorAsk(actorRef: ActorRef, messageFactory: ActorRef ⇒ Any, timeout: Timeout, sender: ActorRef) = actorRef match {
+    case ref: InternalActorRef if ref.isTerminated ⇒
+      val message = messageFactory(ActorRef.noSender)
+      actorRef ! message
+      Future.failed[Any](new AskTimeoutException(s"""Recipient[$actorRef] had already been terminated. Sender[$sender] sent the message of type "${message.getClass.getName}"."""))
     case ref: InternalActorRef ⇒
-      if (timeout.duration.length <= 0)
+      if (timeout.duration.length <= 0) {
+        val message = messageFactory(ActorRef.noSender)
```
rkuhn: This is just a glorified `null`; we should probably use `system.deadletters` instead (to give a real ActorRef).

rkuhn: (same below)

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
```diff
@@ -207,53 +320,153 @@ final class AskableActorSelection(val actorSel: ActorSelection) extends AnyVal {
...
+  private val ActorStopResult = Failure(new ActorKilledException("Stopped"))
+}
+
+/**
  * Akka private optimized representation of the temporary actor spawned to
  * receive the reply to an "ask" operation.
  *
  * INTERNAL API
  */
-private[akka] final class PromiseActorRef private (val provider: ActorRefProvider, val result: Promise[Any], _mcn: String)
```
rkuhn: I think this class will need to remain binary compatible (it is used by prominent “naughty” projects).

akka-actor/src/main/scala/akka/pattern/Patterns.scala
```diff
@@ -48,6 +48,37 @@ object Patterns {
   def ask(actor: ActorRef, message: Any, timeout: Timeout): Future[AnyRef] = scalaAsk(actor, message)(timeout).asInstanceOf[Future[AnyRef]]
 
   /**
+   * A variation of ask which allows to implement "replyTo" pattern by including
+   * sender reference in message.
+   *
+   * For use with Akka compiled with Scala 2.12.
```
rkuhn: Hmm, it would be nice to just do “the right thing” for each release, but that might be impossible. Given that `japi.Function` works fine independent of Scala version I’d remove this variant without replacement—in a different universe this would be the only one we need …

rkuhn:
Sorry for the delay, this fell through the cracks :-( Overall it looks promising but I do not yet understand why the internals were changed, both for untyped and typed. Especially in the untyped case our hands are pretty tied because this was such a useful feature that people were naughty and started using internal APIs; if we can avoid it then we should definitely not break binary compatibility.

ataraxer:
Sorry for the delay on my part too -- had a bit of vacation time myself :)

I had reworked the original implementation from the ground up effectively turning it into an addition only PR. Binary compatibility FTW :)

"Explicit sender" pattern is now explicitly enabled via `akka.pattern.extended.ask` import and is implemented separately from the original ask to avoid changing the code in breaking ways.

`PromiseRef` has been reimplemented as a union of `ActorRef` and `Promise` represented as class members instead of type-level union (which turned out to be quite tricky to implement and fragile to changes) with optional `Implicits` provided to make it's use transparent wherever `ActorRef` or `Promise` is applicable.

`FutureRef` has also been added to mirror the `Promise/Future` relationship and add some flexibility to the further application of `FutureRef` and `PromiseRef`, which turned out to be really useful abstractions on themselves, enabling to express one-time computations performed by an actor which are exposed in a type-safe level as a value of a `Promise` instead of a message dispatch.

akka-actor/src/main/scala/akka/pattern/extended/package.scala
```diff
@@ -0,0 +1,49 @@
...
+ * actor ask (Request(_))   // use it by implicit conversion
+ * }}}
+ *
+ * For Java the patterns are available as static methods of the [[akka.pattern.Patterns]]
+ * class:
+ * {{{
+ * import static akka.pattern.Patterns.ask;
+ *
+ * // for Akka compiled with Scala 2.12
+ * ask(actor, askSender -> new Request(askSender));
```
ataraxer: This comment will be removed: leftover from the original PR.

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
```diff
@@ -184,6 +300,42 @@ final class AskableActorRef(val actorRef: ActorRef) extends AnyVal {
...
+        Future.failed[Any](new IllegalArgumentException(s"""Timeout length must not be negative, question not sent to [$actorRef]. Sender[$sender] sent the message of type "${message.getClass.getName}"."""))
+      } else {
+        val a = PromiseActorRef(ref.provider, timeout, targetName = actorRef, "unknown", sender)
+        val message = messageFactory(a)
+        a.messageClassName = message.getClass.getName
+        actorRef.tell(message, a)
+        a.result.future
+      }
+    case _ ⇒
+      val message = messageFactory(sender.asInstanceOf[InternalActorRef].provider.deadLetters)
```
ataraxer: I'm a bit concerned about this use of `deadLetters`, which is inconsistent from those defined for `InternalActorRef`s above. In the case of local actor communication within the same `ActorSystem` it will be essentially the same mailbox as in `InternalActorRef`s case. However with cross-system communication and/or remoting it becomes a `deadLetters` of sender's system, instead of remote one, which seems at the very least strange semantically. Is it something to be concerned about? Maybe it should be replaced with some kind of dummy `ActorRef` which does nothing at all.

On the other hand, if provided reference is not an `InternalActorRef` there can be no dispatch at all, so in some sense this letter can only be "dead" for the local system... Does this makes sense?

drewhk:
I personally like this. @rkuhn ?

akka-actor/src/main/scala/akka/pattern/PromiseRef.scala
```diff
@@ -0,0 +1,195 @@
...
+     * Provides a conversion to transparently use PromiseRef as ActorRef.
+     */
+
+    implicit def toActorRef(promiseRef: PromiseRef[_]): ActorRef = promiseRef.ref
+
+    /**
+     * Provides a conversion to transparently use PromiseRef to dispatch messages via Scala DSL.
+     */
+    implicit def toScalaActorRef(promiseRef: PromiseRef[_]): ScalaActorRef = promiseRef.ref
+  }
```
rkuhn: I’m not sure about these, we normally don’t do that. At the cost of one more import these conversions make the code one bit less obvious in order to save a few keystrokes—unless of course I’m missing something.

akka-actor/src/main/scala/akka/pattern/PromiseRef.scala
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * Provides a conversion to transparently use FutureRef as ActorRef.
+     */
+    implicit def toScalaActorRef(futureRef: FutureRef[_]): ScalaActorRef = futureRef.ref
+
+    /**
+     * Provides a conversion to transparently use FutureRef to dispatch messages via Scala DSL.
+     */
+    implicit def toActorRef(futureRef: FutureRef[_]): ActorRef = futureRef.ref
+  }
```
rkuhn: same here; and if these should go then the language import should go with them

akka-actor/src/main/scala/akka/pattern/PromiseRef.scala
```diff
@@ -0,0 +1,195 @@
...
+private[akka] final class FutureRefImpl[T](val ref: ActorRef, val future: Future[T])
+  extends FutureRef[T]
+
+private[akka] final class AskPromiseRef private (promiseActorRef: PromiseActorRef)
+  extends PromiseRefImpl[Any](promiseActorRef, promiseActorRef.result)
+
+private[akka] object AskPromiseRef {
+  def apply(provider: ActorRefProvider, timeout: Timeout): AskPromiseRef = {
+    if (timeout.duration.length > 0) {
+      val promiseActorRef = PromiseActorRef(provider, timeout, "unknown", "unknown", ActorRef.noSender)
```
rkuhn: please use `provider.deadLetters` instead of `null` (which is what `noSender` means)

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
rkuhn: Using deadLetters like this is the right thing to do, it is consistent with the rest of Akka. However, this line will throw an NPE if no sender was given, so please check that case and bail out without invoking the function in this case.

rkuhn:
Thanks for the updates @ataraxer , this looks pretty good!

ataraxer:
Updated and rebased to master.

akka-actor/src/main/scala/akka/pattern/AskSupport.scala
```diff
@@ -238,6 +392,41 @@ final class AskableActorSelection(val actorSel: ActorSelection) extends AnyVal {
...
+        val a = PromiseActorRef(ref.provider, timeout, targetName = actorSel, "unknown", sender)
+        val message = messageFactory(a)
+        a.messageClassName = message.getClass.getName
+        actorSel.tell(message, a)
+        a.result.future
+      }
+    case null ⇒
+      Future.failed[Any](new IllegalArgumentException(s"""No recipient provided, question not sent to [$actorSel]."""))
+    case _ ⇒
+      val message = messageFactory(sender.asInstanceOf[InternalActorRef].provider.deadLetters)
```
rkuhn: sorry for being unclear: it is the `sender` that can be `null`, not the `ref`

ataraxer: Oh, of course, my bad. Rushed it a bit.

rkuhn:
Thanks @ataraxer !

ataraxer:
Thanks for guiding me through my first contribution to Akka! It's been a lovely experience (:

rkuhn:
Great, you can always come back for more :-)

ktoso:
Wohoo, awesome this made it in! Thanks @ataraxer :-)

Issue #17898, opened 2015-07-02
ByteString.toString is very dangerous in logging


fommil:
I've encountered this problem a couple of times, e.g. https://github.com/spray/spray/issues/1060

If a `ByteString` ends up in a log message somewhere it basically blows up the log. GB log entries are not uncommon when sending MB-sized messages.

Often the logging happens deep within the akka.io or spray/etc frameworks and it is not possible to selectively tune the logging.

A safe solution would be to implement a `ByteString.toString` that does something like `take(100).toString`

2beaucoup:
see  #13709 (and #16395)

fommil:
yes they are somewhat related, but having a "safe" `toString` for `ByteString` would certainly be a good and simple first step.

jrudolph:
+1

fommil:
Is there any possibility of getting this bumped in priority for the next RC or stable release?

I basically can't trust akka-io in production environments until this is fixed, because any network error anywhere in the system, can potentially produce logging **events** (even at WARNING level) that are GB large.

I can't see any usecase for actually calling the current `toString` as it is effectively unreadable.

viktorklang:
While I can definitely sympathize with the situation, it may not be as simple as it is made out to be.

Given that ByteString does not currently override toString, it should be safer to change it in a dot-release, but I am weary against potentially breaking current code that somehow relies on the current behavior.

Things to consider are `ByteString`s compared by string representation (deduplication or otherwise, which is why I'd be against `take(100).toString` etc).

(keep in mind that all invocations of `toString` on something of `AnyRef` type may end up creating a String of size Integer.MAX_VALUE)

My 2¢

fommil:
@viktorklang I'd be ok with a call to the super implementation for `toString` which would have the object ref and avoid the problem you raised (although anyone who's doing that should probably be sent straight to hell)

patriknw:
We are careful with not invoking `toString` on messages (`AnyRef`), for the same reasons. If there iare any places where we do that on `ByteString` in the Akka code base we should fix that.

That said, I agree with that we should implement a more safe `toString` in 2.4, but I'm also not sure about changing this in 2.3.x

viktorklang:
@fommil That suggestion would break all code that considers the toString-representation equality as equality. ByteString("foo").toString == ByteString("foo").toString would suddenly be `false`.

fommil:
[...]

PR #18247, opened 2015-08-18
=act #17898 Limit length of ByteString.toString

patriknw:


patriknw:
Refs #17898

akka-actor/src/main/scala/akka/util/ByteString.scala
```diff
@@ -409,6 +409,13 @@ sealed abstract class ByteString extends IndexedSeq[Byte] with IndexedSeqOptimiz
   override def indexWhere(p: Byte ⇒ Boolean): Int = iterator.indexWhere(p)
   override def indexOf[B >: Byte](elem: B): Int = iterator.indexOf(elem)
 
+  override def toString(): String = {
+    if (size > 100)
+      take(100).toString + "..."
```
ktoso: Would it be useful to `" [and $size more]..."`?

patriknw: sounds good

patriknw:
@ktoso updated with "and more"

ktoso:
LGTM

Issue #17768, opened 2015-06-18
Suppress sigar load exception


patriknw:
Looks like the latest cluster metrics changes prints huge stack traces when sigar cannot be loaded. Noticed this in the cluster sample. Might be correct, but creating this ticket to remember to verify it.

    org.hyperic.sigar.SigarException: no libsigar-universal64-macosx.dylib in java.library.path
    [JVM-3]     at org.hyperic.sigar.Sigar.loadLibrary(Sigar.java:172)
    [JVM-3]     at org.hyperic.sigar.Sigar.<clinit>(Sigar.java:100)
    [JVM-3]     at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
    [JVM-3]     at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
    [JVM-3]     at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
    [JVM-3]     at java.lang.reflect.Constructor.newInstance(Constructor.java:408)
    [JVM-3]     at akka.actor.ReflectiveDynamicAccess$$anonfun$createInstanceFor$2.apply(DynamicAccess.scala:78)
    [JVM-3]     at scala.util.Try$.apply(Try.scala:191)
    [JVM-3]     at akka.actor.ReflectiveDynamicAccess.createInstanceFor(DynamicAccess.scala:73)
    [JVM-3]     at akka.actor.ReflectiveDynamicAccess$$anonfun$createInstanceFor$3.apply(DynamicAccess.scala:84)
    [JVM-3]     at akka.actor.ReflectiveDynamicAccess$$anonfun$createInstanceFor$3.apply(DynamicAccess.scala:84)
    [JVM-3]     at scala.util.Success.flatMap(Try.scala:230)
    [JVM-3]     at akka.actor.ReflectiveDynamicAccess.createInstanceFor(DynamicAccess.scala:84)
    [JVM-3]     at akka.cluster.SigarMetricsCollector.<init>(ClusterMetricsCollector.scala:712)
    [JVM-3]     at akka.cluster.SigarMetricsCollector.<init>(ClusterMetricsCollector.scala:717)
    [JVM-3]     at akka.cluster.MetricsCollector$$anonfun$7.apply(ClusterMetricsCollector.scala:792)

patriknw:
[...]

PR #18253, opened 2015-08-18
=clm #17768 Silence stacktrace in cluster sample

patriknw:
It works, but the stack trace is very annoying since it will always be printed by Sigar itself
(so stupid of such a library to use stderr and stdout).

```
[error] org.hyperic.sigar.SigarException: no libsigar-universal64-macosx.dylib in java.library.path
[error]     at org.hyperic.sigar.Sigar.loadLibrary(Sigar.java:174)
[error]     at org.hyperic.sigar.Sigar.<clinit>(Sigar.java:102)
```

A workaround is to run with `-Djava.library.path=./target/native`, then it will only print
the stack trace when it is not provisioned and next time it's silent.

Also updated to latest sigar-loader, but that did not make
any difference for this issue


drewhk:
LGTM

patriknw:
PLS BUILD

ktoso:
LGTM

ktoso:
Refs #17768

Issue #13523, opened 2013-07-28
shall cluster-client expose sender in replies?


rkuhn:
imported from https://www.assembla.com/spaces/akka/tickets/3523

roland:

Another thing I wonder is whether our choice of exposing the responding ActorRef is the right one: an ActorRef signifies the capability to access a resource and passing an ActorRef along proliferates that capability. Maybe the receptionist should send on the reply with its own sender address and if an actor wants to introduce itself to external clients it needs to do so by passing the ActorRef as part of the message, opt-in instead of opt-out.

patrik:

One thing to consider is that the messages sent via the client/receptionist inwards must be wrapped, which must be remembered when replying back to such a reference. I think I prefer deadLetters as sender if we don't want to expose the cluster destination ref as sender. Please create a ticket and we can think more about it.

PR #18259, opened 2015-08-18
!clt #13523 Don't expose sender in cluster client replies

patriknw:


patriknw:
Refs #13523

hepin1989:
seems I need update my code:) once I upgrade to RC1

ktoso:
I'm not completely up to date with the reasoning for this change... Why is it bad to expose the ActorRef?

patriknw:
The user should opt-in to talking directly from client to an actor in the cluster. Previosly it was too easy to just use the sender without thinking about the consequences.

ktoso:
Hmm, I see... makes sense I guess. LGTM

Issue #15040, opened 2014-04-23
ScatterGatherFirstCompleted router doesn't respond for 0 routees


hseeberger:
The contract that the ScatterGatherFirstCompleted router always should respond with the first response or with a Status.Failure is broken, because for 0 routees it does't respond at all.

Currently the ScatterGatherFirstCompleted router can only be used with ask, because ask will also result in a failure, yet it's a different one (Try, not Status), which is nasty. Even more nasty, one can't use the ScatterGatherFirstCompleted router with fire-and-forget, because one will wait forever for a response.

The fix seems trivial to me. In
```
final case class ScatterGatherFirstCompletedRoutingLogic(within: FiniteDuration) extends RoutingLogic {
  override def select(message: Any, routees: immutable.IndexedSeq[Routee]): Routee =
    if (routees.isEmpty) NoRoutee
    else ScatterGatherFirstCompletedRoutees(routees, within)
}
```

simply remove the if-case.

hseeberger:
Thx

PR #18260, opened 2015-08-18
=act #15040 ScatterGatherFirstCompleted reply when no routees

patriknw:


patriknw:
Refs #15040

akka-actor-tests/src/test/scala/akka/routing/ScatterGatherFirstCompletedSpec.scala
```diff
@@ -86,4 +89,15 @@ class ScatterGatherFirstCompletedSpec extends AkkaSpec with DefaultTimeout with
 
   }
 
+  "Scatter-gather pool" must {
+
+    "without routees should reply immediately" in {
+      val probe = TestProbe()
+      val router = system.actorOf(ScatterGatherFirstCompletedPool(nrOfInstances = 0, within = 5.seconds).props(Props.empty))
```
ktoso: I thought that we may consider `0` illegal here, but the case Heiko spoke of can well be triggered by all the initially in-the-pool routees to die, leaving the pool having 0 routees... then the optimisation makes sense, right?

patriknw: Yes, i just thought this was an easy way to test it

ktoso: Okey, sounds good. Making the 0 throw right away during construction probably does not buy much here in terms of safety for users, they'll notice anyway.

ktoso:
LGTM

akka-actor-tests/src/test/scala/akka/routing/ScatterGatherFirstCompletedSpec.scala
patriknw: There might be corner cases with not allowing 0, e.g. could work with 0 + resizer.

ktoso: I see :) The PR looks good in any case, feel free to merge when you get the time

PR #18261, opened 2015-08-18
+htc #18250 add HttpEntity#apply(contentType, source) for Chunked

### LINKED ISSUE #18250: Add overload for HttpEntity for unknown length stream
A "streaming" http echo currently looks like this:

```scala
val httpEcho = Flow[HttpRequest].map(r ⇒ 
  HttpResponse(entity = Chunked.fromData(octetStream, r.entity.dataBytes)))
```

As HttpEntity does not have an overload that does not take a stream length (there is `(contentType, length, source)`).

We should add:

```scala
def apply(contentType: ContentType, data: Source[ByteString, Any]): Chunked =
    Chunked.fromData(contentType, data)
```

to allow:

```scala
val httpEcho = Flow[HttpRequest].map(r ⇒ 
  HttpResponse(entity = HttpEntity(octetStream, r.entity.dataBytes)))
```

Should `    def fromData(contentType: ContentType, chunks: Source[ByteString, Any]): Chunked =` also be exposed as `apply()`? It's not marked as Java API (fromData), but looks a bit weird for Scala DSL

ktoso:
Resolves #18250

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntities.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntities.java
@@ -7,9 +7,10 @@ package akka.http.javadsl.model;
 import java.io.File;
 
 import akka.http.impl.util.JavaAccessors;
+import akka.http.scaladsl.model.*;
+import akka.http.scaladsl.model.HttpEntity;
 import akka.util.ByteString;
 import akka.stream.javadsl.Source;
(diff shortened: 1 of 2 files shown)


ktoso:
Ooh, right, forgot to ping: Please review @sirthias @jrudolph

jrudolph:
LGTM. Makes sense to provide this for consistency.

Issue #17750, opened 2015-06-16
Investigate if expected-response-after can be removed


jknight12882:
Ticket in response to this thread https://groups.google.com/d/topic/akka-user/2JKtDw7dcJs/discussion

patriknw:
thanks

PR #18263, opened 2015-08-19
=rem,clu #17750 Decrease default expected-response-after

patriknw:
`expected-response-after` is used by remote watcher and cluster failure detector as an extra margin for when establishing heartbeating for a new connection. I don't want to remove it now. It could be good for compensating for extra delays for new connections such as dns lookups and handshakes. The default value can be decreased though.

### DIFF
--- a/akka-cluster/src/main/resources/reference.conf
+++ b/akka-cluster/src/main/resources/reference.conf
@@ -146,9 +146,9 @@ akka {
       monitored-by-nr-of-members = 5
       
       # After the heartbeat request has been sent the first failure detection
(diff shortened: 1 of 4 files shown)


patriknw:
Refs #17750

patriknw:
wow, jvm crash: "Could not load hsdis-amd64.so"

patriknw:
PLS BUILD

drewhk:
LGTM

ktoso:
LGTM 

> wow, jvm crash: "Could not load hsdis-amd64.so"

Whoa.

Issue #18137, opened 2015-08-06
Make EventAdapter into ReadEventAdapter and WriteEventAdapter


ktoso:
See here https://github.com/akka/akka/issues/18050#issuecomment-123756225

patriknw:
@ktoso looks like this is related to #18050, so let's do those together after 2.4-M3

ktoso:
Yes, let's do that.
The task here seems simple, though it has some implications on the code obtaining adapters, so it's a bit more involved thanjust "split the interface"

PR #18265, opened 2015-08-19
+per #18137 EventAdapter => can be Read / Write or both

ktoso:
Resolves https://github.com/akka/akka/issues/18137

### DIFF
--- a/akka-docs/rst/java/code/docs/persistence/PersistenceEventAdapterDocTest.java
+++ b/akka-docs/rst/java/code/docs/persistence/PersistenceEventAdapterDocTest.java
@@ -12,7 +12,7 @@ public class PersistenceEventAdapterDocTest {
   @SuppressWarnings("unused")
   static
   //#identity-event-adapter
-  class MyEventAdapter extends EventAdapter {
+  class MyEventAdapter implements EventAdapter {
     @Override
     public String manifest(Object event) {
       return ""; // if no manifest needed, return ""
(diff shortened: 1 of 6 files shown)


akka-docs/rst/java/code/docs/persistence/PersistenceEventAdapterDocTest.java
```diff
@@ -12,7 +12,7 @@
   @SuppressWarnings("unused")
   static
   //#identity-event-adapter
-  class MyEventAdapter extends EventAdapter {
+  class MyEventAdapter implements EventAdapter {
```
ktoso: made it an interface, makes more sense since it's just abstract classes anyway

akka-persistence/src/main/scala/akka/persistence/journal/EventAdapters.scala
```diff
@@ -96,6 +96,18 @@ private[akka] object EventAdapters {
     new EventAdapters(backing, bindings, system.log)
   }
 
+  def instanciateAdapter(adapterFQN: String, system: ExtendedActorSystem): Try[EventAdapter] = {
```
patriknw: spelling looks weird, instantiate

ktoso: good catch, thanks

patriknw:
LGTM

PR #18267, opened 2015-08-19
+str: Add Sink.toSeq and Sink.single

RichardBradley:
I thought that the "Draining a stream to a strict collection" recipe could do with moving into the `Sink` helper in Akka, because it seems like such a fundamental operation that it was strange not to have a built in op for it.

I also added a "`single`" sink because this is another very common pattern which I think could do with a canonical implementation.

I have updated the documentation examples and the Akka source to use the new `Sink.toSeq` and `Sink.single` where appropriate. Hopefully this demonstrates the ubiquity of these two use-cases.

In both cases, the recipe is a little fiddly so I think users would benefit from a correct and tested canonical implementation in the base library.
### Implementation notes

The [recipe given in the docs](http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0/scala/stream-cookbook.html#Draining_a_stream_to_a_strict_collection) for `Sink.toSeq` uses "`group`" + "`head`". I've ported that into `Sink` directly here.
Is that more efficient than using "`fold`" to aggregate a collection? 

What happens if the collection is empty? I don't see a "`headOption`" Sink. I've added a "recover", but it seems like it might be better to write a new `Subscriber` for "`toSeq`" like the "`HeadSinkSubscriber`"? How would the performance of that compare to the "group" version?

I had to introduce an implicit `ExecutionContext` for both Sinks. This makes them awkward to use in places. This would go away if they used a custom subscriber like the "`HeadSinkSubscriber`". Or they could require an implicit `FlowMaterializer`? (The latter is more heavyweight, but more likely to be in scope.)
### Java DSL naming

The "`DslFactoriesConsistencySpec`" test forces me to name the Sink "toSeq" in the Java version as well.
I've added a "toList" alias for it, as this is a much more natural name in Java.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
@@ -88,6 +88,22 @@ object Sink extends SinkApply {
       .mapMaterializedValue { _.recover { case _: NoSuchElementException ⇒ Nil } }
 
   /**
+   * A `Sink` that drains a stream and returns the sole element
+   * that the stream produced.
+   * If the stream did not produce exactly one element, a failed Future will be
+   * returned.
+   */
+  def single[T](implicit ec: ExecutionContext): Sink[T, Future[T]] =
+    Flow[T].grouped(2).toMat(Sink.head) { (_, firstTwo) ⇒
+      firstTwo.map {
+        // (Empty streams are alredy rejected by Sink.head.)
+        case Seq(element) ⇒ element
+        case _ ⇒ throw new IllegalStateException(
+          "Expected exactly one element from the stream, but there were at least two.")
+      }
+    }
+
+  /**
    * A `Sink` that materializes into a [[org.reactivestreams.Publisher]].
    * that can handle one [[org.reactivestreams.Subscriber]].
    */
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
@@ -13,6 +13,7 @@ import akka.stream.Attributes._
 import akka.stream.stage.{ TerminationDirective, Directive, Context, PushStage, SyncDirective }
 import org.reactivestreams.{ Publisher, Subscriber }
 
+import scala.collection.immutable
 import scala.concurrent.{ ExecutionContext, Future, Promise }
 import scala.util.{ Failure, Success, Try }
 
@@ -75,6 +76,16 @@ object Sink extends SinkApply {
    * A `Sink` that materializes into a `Future` of the first value received.
    */
   def head[T]: Sink[T, Future[T]] = new Sink(new HeadSink[T](DefaultAttributes.headSink, shape("HeadSink")))
+
+  /**
+   * A `Sink` that materializes into a `Future` of all elements in the stream
+   * as a strict collection.
+   * If the stream is inifite, this will never complete but it will consume
+   * all of your memory.
+   */
+  def toSeq[T](implicit ec: ExecutionContext): Sink[T, Future[immutable.Seq[T]]] =
+    Flow[T].grouped(Int.MaxValue).toMat(Sink.head)(Keep.right)
+      .mapMaterializedValue { _.recover { case _: NoSuchElementException ⇒ Nil } }
 
   /**
    * A `Sink` that materializes into a [[org.reactivestreams.Publisher]].
--- a/akka-docs-dev/rst/java/stream-cookbook.rst
+++ b/akka-docs-dev/rst/java/stream-cookbook.rst
@@ -54,14 +54,9 @@ collection itself, so we can just call ``mapConcat(l -> l)``.
 Draining a stream to a strict collection
 ----------------------------------------
 
-**Situation:** A finite sequence of elements is given as a stream, but a scala collection is needed instead.
+**Situation:** A finite sequence of elements is given as a stream, but a Java collection is needed instead.
 
-In this recipe we will use the ``grouped`` stream operation that groups incoming elements into a stream of limited
-size collections (it can be seen as the almost opposite version of the "Flattening a stream of sequences" recipe
-we showed before). By using a ``grouped(MAX_ALLOWED_SIZE)`` we create a stream of groups
-with maximum size of ``MaxAllowedSeqSize`` and then we take the first element of this stream by attaching a ``Sink.head()``. What we get is a
-:class:`Future` containing a sequence with all the elements of the original up to ``MAX_ALLOWED_SIZE`` size (further
-elements are dropped).
+The ``Sink.toList`` built-in sink can drain a finite stream into a Java collection.
 
 .. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeToStrict.java#draining-to-list
 
--- a/akka-docs-dev/rst/scala/code/docs/http/scaladsl/HttpClientExampleSpec.scala
+++ b/akka-docs-dev/rst/scala/code/docs/http/scaladsl/HttpClientExampleSpec.scala
@@ -20,6 +20,7 @@ class HttpClientExampleSpec extends WordSpec with Matchers {
 
     implicit val system = ActorSystem()
     implicit val materializer = ActorMaterializer()
+    import system.dispatcher
 
     val connectionFlow: Flow[HttpRequest, HttpResponse, Future[Http.OutgoingConnection]] =
       Http().outgoingConnection("akka.io")
@@ -42,6 +43,7 @@ class HttpClientExampleSpec extends WordSpec with Matchers {
 
     implicit val system = ActorSystem()
     implicit val materializer = ActorMaterializer()
+    import system.dispatcher
 
     // construct a pool client flow with context type `Int`
     val poolClientFlow = Http().cachedHostConnectionPool[Int]("akka.io")
--- a/akka-docs-dev/rst/scala/code/docs/http/scaladsl/HttpClientExampleSpec.scala
+++ b/akka-docs-dev/rst/scala/code/docs/http/scaladsl/HttpClientExampleSpec.scala
@@ -26,7 +26,7 @@ class HttpClientExampleSpec extends WordSpec with Matchers {
     val responseFuture: Future[HttpResponse] =
       Source.single(HttpRequest(uri = "/"))
         .via(connectionFlow)
-        .runWith(Sink.head)
+        .runWith(Sink.single)
     //#outgoing-connection-example
   }
 
@@ -48,7 +48,7 @@ class HttpClientExampleSpec extends WordSpec with Matchers {
     val responseFuture: Future[(Try[HttpResponse], Int)] =
       Source.single(HttpRequest(uri = "/") -> 42)
         .via(poolClientFlow)
-        .runWith(Sink.head)
+        .runWith(Sink.single)
     //#host-level-example
   }
 
--- a/akka-docs-dev/rst/scala/code/docs/stream/BidiFlowDocSpec.scala
+++ b/akka-docs-dev/rst/scala/code/docs/stream/BidiFlowDocSpec.scala
@@ -131,6 +131,7 @@ class BidiFlowDocSpec extends AkkaSpec with ConversionCheckedTripleEquals {
   import BidiFlowDocSpec._
 
   implicit val mat = ActorMaterializer()
+  import mat.executionContext
 
   "A BidiFlow" must {
 
@@ -152,7 +153,7 @@ class BidiFlowDocSpec extends AkkaSpec with ConversionCheckedTripleEquals {
       // test it by plugging it into its own inverse and closing the right end
       val pingpong = Flow[Message].collect { case Ping(id) => Pong(id) }
       val flow = stack.atop(stack.reversed).join(pingpong)
-      val result = Source((0 to 9).map(Ping)).via(flow).grouped(20).runWith(Sink.head)
+      val result = Source((0 to 9).map(Ping)).via(flow).runWith(Sink.toSeq)
       Await.result(result, 1.second) should ===((0 to 9).map(Pong))
       //#compose
     }
@@ -160,14 +161,14 @@ class BidiFlowDocSpec extends AkkaSpec with ConversionCheckedTripleEquals {
     "work when chopped up" in {
       val stack = codec.atop(framing)
       val flow = stack.atop(chopUp).atop(stack.reversed).join(Flow[Message].map { case Ping(id) => Pong(id) })
-      val f = Source((0 to 9).map(Ping)).via(flow).grouped(20).runWith(Sink.head)
+      val f = Source((0 to 9).map(Ping)).via(flow).runWith(Sink.toSeq)
       Await.result(f, 1.second) should ===((0 to 9).map(Pong))
     }
 
     "work when accumulated" in {
       val stack = codec.atop(framing)
       val flow = stack.atop(accumulate).atop(stack.reversed).join(Flow[Message].map { case Ping(id) => Pong(id) })
-      val f = Source((0 to 9).map(Ping)).via(flow).grouped(20).runWith(Sink.head)
+      val f = Source((0 to 9).map(Ping)).via(flow).runWith(Sink.toSeq)
       Await.result(f, 1.second) should ===((0 to 9).map(Pong))
     }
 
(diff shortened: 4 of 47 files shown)


drewhk:
OK TO TEST

drewhk:
I have to disagree with the choices here. There is no need for Sink.single, I would rather add a conterpart to "take", "limit" which instead of dropping additional elements it errors out. Then the current Sink.single would be limit(1).to(Sink.head).

As for toSeq, that was a deliberate decision. It is an invitation for trouble to try to drain a potentially infinite stream to a strict collection. If such a built-in functionality is desired, it should take an explicit limit parameter to cap the size of the collection and should error if that limit is exceeded.

RichardBradley:
> I have to disagree with the choices here. 

I disagree (naturally, otherwise I wouldn't have created this PR :-) )

I have been using Akka HTTP with Streams a lot over the past few months (pretty much full time writing webapps with Akka HTTP since the first preview release last year. These helpers are based on pragmatic requirements of Akka HTTP apps, rather than abstract design. I expect that most Akka HTTP apps will have "toSeq" and "single" in their "utils" packages, as it comes up a lot when using the streaming HTTP APIs.

> There is no need for Sink.single, I would rather add a conterpart to "take", "limit" which instead of dropping additional elements it errors out. Then the current Sink.single would be limit(1).to(Sink.head).

That would work.
Consuming a stream that you know to have a single element seems to be a surprisingly common occurrence in Akka HTTP, hence my proposal here.
Currently everyone seems to use "Sink.head" (see e.g. "HttpClientExampleSpec" and some of the HTTP parsing code), but IMO it's ugly to ignore any extra elements and it seems more type safe to assert that there is one element.

It also has a pleasing symmetry with `Source.single`

> As for toSeq, that was a deliberate decision. It is an invitation for trouble to try to drain a potentially infinite stream to a strict collection. If such a built-in functionality is desired, it should take an explicit limit parameter to cap the size of the collection and should error if that limit is exceeded.

I don't think that's a problem.
Take a look at the Scala (or Java) collections API -- all the potentially-infinite collections have a "toList" operator which doesn't have a "max size" param. Without one, the convention is that the library should attempt to create the list and either succeed or run out of memory. I think Akka could do the same without surprising anyone.

The Akka code is littered with `x.grouped(1234).head` statements, as you can see from the diff here. It's distracting to have these magic numbers all over the place with the estimated size of the stream. Also, if your magic number is too small, there will be no error with that version, which seems like a serious bug to me. 

I definitely think Akka should provide a `toSeq` sink because of these pitfalls.

> Some checks haven’t completed yet

(There might be some test failures in the first build -- I wasn't able to run all tests locally. I'll fix up when this finishes.)

RichardBradley:
> Test FAILed.

I'll come back to this tomorrow or later this week.

RichardBradley:
Should I email the akka-dev list to ask for opinions on this? Or raise an issue?
Should I cc any of the main Akka HTTP contribs? e.g. sirthias or jrudolph?
Thanks

ktoso:
Hi @RichardBradley,
Endre (who's working on streams right now) has public holiday these 2 days, we'll get back to the discussion soon :-) The rest of the team was focused on releasing `2.4.0-RC1` today, so sorry about no more replies here.

RichardBradley:
Ok, thanks. No particular rush from my end.
We should resolve #18292 before merging in `Sink.single`, as it causes HTTP deadlocks in some cases.
The `Sink.toSeq` should be less problematic :-)
I can split the two up if necessary, but let's have the design discussion first.

ktoso:
I agree with Endre that we do not want a toSeq operation, since it leads devs down the wrong way (too easy to try to toSeq an infinite stream). The places where you see the `grouped().runWith(Sink.head)` in the Akka code base are mostly tests, which were written before we had a streams test-kit, so it was our "poor-man's test-kit", I don't think there's much of that pattern in actual code, and for tests we currently have a great testkit-for streams where you can `.expectNextN(List(...))` which handles that case.

A `limit` on the other hand would be pretty interesting :) Let's get back to this issue next week (with a proper in-depth review then). Thanks for your patience :)

jrudolph:
Just saw this. @RichardBradley, could you maybe provide a few examples of why you think those combinators are needed when working with akka-http? Because, I somewhat agree with @drewhk that you shouldn't be forced to use operations like this in most code.

On the other hand, as you observed as well, we already went through a weird dance in akka-http tests where we fluctuated from `.grouped(1000).runWith(Sink.head)` (which is all over the place) to a custom abstraction that lived in `akka.http.impl.util` back to `.grouped(1000).runWith(Sink.head)` because of "we don't want to introduce private, temporary infrastructure". This is a strong sign that some kind of API is missing. I'm not sure if it is `toSeq`, though, which seems to low-level and should be replaced by higher-level domain-specific concepts to prevent it from spreading everywhere.

RichardBradley:
> Just saw this. @RichardBradley, could you maybe provide a few examples of why you think those combinators are needed when working with akka-http?

The examples seen here in Akka HTTP itself are representative:
 * test code
  * why dismiss test code as not deserving a nice API?
  * the alternative is lots of ugly `.grouped(1000).runWith(Sink.head)` snippets
 * HTTP clients etc, where you expect exactly one response out of a `Source[HttpResponse]`
 * Interfacing Streams with regular collections
  * (if this is never a concern, why is it addressed in the cookbook?)

> This is a strong sign that some kind of API is missing. I'm not sure if it is toSeq, though, which seems to low-level and should be replaced by higher-level domain-specific concepts to prevent it from spreading everywhere.

I don't think there is a hidden "higher-level domain-specific concept" here.
I think that sometimes you just want to run a stream to completion and capture the output in a `Seq`

RichardBradley:
> we do not want a toSeq operation, since it leads devs down the wrong way (too easy to try to toSeq an infinite stream)

I think this justification about infinite streams is not realistic.
The vast majority of real useful Streams are not infinite (i/o, HTTP request / responses etc.).
Also, it is quite conventional for possibly-unbounded structures to offer a "toStrict" style API which will not complete for infinite cases. See e.g `TraversableOnce.toSeq` or `HttpRequest.toStrict`.

jrudolph:
Thanks for the feedback @RichardBradley.

Matters are a bit different with streams than with just a collection API. Due to the asynchronous nature, an infinite stream cannot be distinguished from a slowly producing one. Also, a running stream "lives on its own", i.e. similarly to an actor it won't get garbage collected once all user-level references are gone. Also, `toSeq` in particular, will aggregate unlimited amounts of data (same as for `toStrict`).

So, I think you could make two different points here:

 * Should there be any terminal operations (i.e. Sinks) provided at all that are either unbounded in space or time?
 * Which sets of operations should be provided out-of-the-box?

For the first point, the sane and consistent thing seems to be to make sure that *all* terminal operations need to provide bounds for time and space consumption. Because, regarding the time limit, `toSeq`, `head`, `single`, `toStrict`, or `foreach` do not really differ.

Regarding the second point, opinions have differed about that right from the beginning. 

There's one stance that (akka-)stream should only provide the basic set of operations and combinators out-of-the-box and everything else needs to be hand-written by the user. This is mostly to keep the API surface as small as possible and partly also because we've recognized that pre-packaged higher-level combinators may easily have almost-but-not-quite the right semantics a user would expect and the user would need to fall back to lower-level ones in many cases. 

The other stance would be that a rich set of combinators should be provided just to avoid that users need to reinvent all the abstractions themselves. An advantage of providing a high-level abstraction directly is that it may sometimes have a simpler semantics than each of the parts it is composed of. In this case, not providing the higher-level API may actually impede the possibility to evolve the low-level API because more (and a-priori potentially unknown) users of the low-level API will start to depend on the de-facto set of low-level semantics they implement their own version of the high-level API on.

jrudolph:
> I don't think there is a hidden "higher-level domain-specific concept" here.

I think there is :) E.g. in testing  `grouped()...` is used to assert that a stream will deliver a certain sequence of elements in a certain time. This is such a higher-level concept that is completely unrelated to whether those elements need to be collected first or how the comparison is implemented in detail. It shows that a testing API is missing (or wasn't used for some reason) which would allow writing the assertion in a more concise form.

ktoso:
>  E.g. in testing grouped()... is used to assert that a stream will deliver a certain sequence of elements in a certain time. This is such a higher-level concept that is completely unrelated to whether those elements need to be collected first or how the comparison is implemented in detail. It shows that a testing API is missing

This part I agree with very much. In fact, the missing APIs for testing already have a PR: https://github.com/akka/akka/pull/18503 (usage: https://github.com/akka/akka/pull/18503/files#diff-133e483519cf529d626c1f559af0ab55R55 ) :smile:

RichardBradley:
> Matters are a bit different with streams than with just a collection API. Due to the asynchronous nature, an infinite stream cannot be distinguished from a slowly producing one. Also, a running stream "lives on its own", i.e. similarly to an actor it won't get garbage collected once all user-level references are gone. Also, toSeq in particular, will aggregate unlimited amounts of data (same as for toStrict).

All these things apply to a number of similar pre-existing concepts, each of which offers a "toStrict" which may not complete or may consume unbounded resources. I don't think Akka streams is especially unusual in this regard. In each of these cases, the programmer must decide up-front, using information that is not available to the compiler, whether it is safe to make these calls (or handle the failure case).

 * `scala.concurrent.Future` has `Await.result`. If the future take unbounded time to complete, then so will Await
 * `scala.collection.immutable.Stream` has `.toSeq`. If the stream is infinite, then this call will aggregate unlimited amounts of data. You can't distinguish up-front between a slow or an infinite stream.
* `java.io.InputStream` has `IOUtils.toString`.

> I think there is [a hidden "higher-level domain-specific concept" here] E.g. in testing grouped()... is used to assert that a stream will deliver a certain sequence of elements in a certain time.

There's no explicit time limit set on `grouped()` in the tests here.

The pattern you can see all over the Akka test suite is just

```stream.grouped(10000).runWith(Sink.head) === List(expected)```

this is entirely equivalent to the `toSeq` case, and avoids the distracting magic number:

```stream.runWith(Sink.toSeq) === List(expected)```

(The "toSeq" one has an additional advantage in that we also verify that the stream terminates correctly.)

Sorry, I don't see the difference between these two? Could you expand on what higher-level domain-specific concept `.grouped(10000).runWith(Sink.head)` encapsulates v.s. `.runWith(Sink.toSeq)`?

**Comment by RichardBradley** [comment]
> the sane and consistent thing seems to be to make sure that all terminal operations need to provide bounds for time and space consumption. Because, regarding the time limit, toSeq, head, single, toStrict, or foreach do not really differ.

That sounds good to me.

It might be tedious to make it explicit though. There's already an implicit time bound, isn't there? Or if not, all of these return `Future`s, so the user can apply their own max wait time.

**Comment by ktoso** [comment]
@RichardBradley, please review my PR before continuing to argue for the timeout-less versions (i.e. grouped as well), it has explicit timeouts: https://github.com/akka/akka/pull/18503 :-)

**Comment by RichardBradley** [comment]
> There's one stance that (akka-)stream should only provide the basic set of operations and combinators out-of-the-box and everything else needs to be hand-written by the user.

That sounds fine. I think that "toSeq" is a slightly special case because a) it's a fundamental op and b) it can be implemented far more efficiently by Akka than in userland and c) the current "recipe" is quite ugly, requiring this "10000" magic number.

**Comment by jrudolph** [comment]
> Could you expand on what higher-level domain-specific concept .grouped(10000).runWith(Sink.head) encapsulates v.s. .runWith(Sink.toSeq)

I'd say there should be a test method to assert which elements to expect from a stream and in which time frame.

**Comment by RichardBradley** [comment]
> I'd say there should be a test method to assert which elements to expect from a stream and in which time frame.

The time frame is not actually the item under test in 99% of the tests affected by this patch, or for most userland tests which will use similar constructs.

`BidiFlowDocSpec.scala` is a reasonable example:

```
"A BidiFlow" must {

    "compose" in {
      //#compose
      /* construct protocol stack
       *         +------------------------------------+
       *         | stack                              |
       *         |                                    |
       *         |  +-------+            +---------+  |
       *    ~>   O~~o       |     ~>     |         o~~O    ~>
       * Message |  | codec | ByteString | framing |  | ByteString
       *    <~   O~~o       |     <~     |         o~~O    <~
       *         |  +-------+            +---------+  |
       *         +------------------------------------+
       */
      val stack = codec.atop(framing)

      // test it by plugging it into its own inverse and closing the right end
      val pingpong = Flow[Message].collect { case Ping(id) => Pong(id) }
      val flow = stack.atop(stack.reversed).join(pingpong)
      val result = Source((0 to 9).map(Ping)).via(flow).runWith(Sink.toSeq)
      Await.result(result, 1.second) should ===((0 to 9).map(Pong))
      //#compose
    }
```

This test is all about what will come out of the stream. It adds a "1 second" wait time, but that's not really important.

If I were writing this in `specs2`, I'd just use the "await" helper, and settle for the default expect timeout:

```
    val result = Source((0 to 9).map(Ping)).via(flow).runWith(Sink.toSeq)
    result must beEqualTo((0 to 9).map(Pong)).await()
```

This is much clearer and a much more accurate representation of what the test is trying to assert than the alternatives proposed here:

```
   Source((0 to 9).map(Ping)).via(flow)
    .runWith(TestSink.probe[Pong])
    .request(10)
    .expectNext((0 to 9).map(Pong))
    .expectTermination()
    .within(1.second)
```

(I may have gotten the #18503 syntax a bit wrong here -- I think it's something like that above)

Sorry to keep banging on about this. If you really hate this idea, then I can continue to use my userland "Sinks.toSeq" implementation, but it's a bit less efficient than a built-in would be, and I do think Akka would benefit from this.

File: akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala [hunk 1]
```diff
@@ -77,6 +78,32 @@ object Sink extends SinkApply {
   def head[T]: Sink[T, Future[T]] = new Sink(new HeadSink[T](DefaultAttributes.headSink, shape("HeadSink")))
 
   /**
+   * A `Sink` that materializes into a `Future` of all elements in the stream
+   * as a strict collection.
+   * If the stream is inifite, this will never complete but it will consume
+   * all of your memory.
+   */
+  def toSeq[T](implicit ec: ExecutionContext): Sink[T, Future[immutable.Seq[T]]] =
```
rkuhn: this method should not take an ExecutionContext: the recovery step is wait-free and therefore can run on the sameThreadExecutionContext

rkuhn:
Sorry, I’m late to the party: it is obvious that collecting a finite stream into a sequence is a popular operation, I think we should support it. I’d recommend keeping the concerns orthogonal:
~~~scala
val single = source.limit(1).runWith(Sink.head)
val seq = source.limit(65536).within(5.seconds).runWith(Sink.toSeq)
~~~
This implies that `Sink.single` is superfluous as shown above (limit will have to assert completion before emitting the final element in order to live up to its expectations). Does that work for everyone?

Side remark: instead of having an unusual alias in the Java DSL we should instead recognize this difference in the consistency spec.

RichardBradley:
Sounds good to me, thanks.
Do you want me to try to fix up my patch, or would the Akka team prefer to do this work? 
(I'm happy for you to use any or all of this commit as the basis of a fix if you'd like.)

drewhk:
There is a method on testProbe that allows testing against a given strict sequence, so if this is just about testing, then that is already covered.

rkuhn:
Sorry for the delay, @RichardBradley you’re welcome to add the `limit` operator if you want; if not I’ll take this over later this week.

drewhk:
Ok, so it is agreed to have Sink.toSeq and limit. We already have `completionTimeout`. So the usual pattern will be:

    src.completionTimeout(5.seconds).limit(1000).runWith(Sink.toSeq)

@RichardBradley feel free to have limit and Sink.toSeq implemented yourself, we are a bit busy :) You can also try to implement the toSeq sink in terms of the new GraphStage. It will look like roughly:

    class MySink[T] extends GraphStageWithMaterializedValue[SinkShape[T], Future[Seq[T]]] {
        val in: Inlet[T] = Inlet("mySink.in")
        val shape: SinkShape[T](in)
    
        val createLogicAndMaterializedValue(inheritedAttributes: Attributes) = {
            val promise = Promise[Seq[T]]
            val logic = new GraphStageLogic(shape) {
              // Your state comes here
          
              // Your logic goes here
                 def onPush() = ...{
                   val elem = grab(in)
                 }
     
                 def onUpstreamFinish() = ...
              }
            })
    
            (logic, promise.future)
        }
    }

RichardBradley:
>  feel free to have limit and Sink.toSeq implemented yourself, we are a bit busy

Thanks. I would like to, but I don't think I'll get around to it any time soon either.
I'll keep this on my list in case I do get to it before you guys :-)
I have emailed a few people at my company who have been working with me on our Akka HTTP app to see if they'd like to have a go.

> You can also try to implement the toSeq sink in terms of the new GraphStage.

That looks better -- this should be done in the most efficient way possible, if it's an official operator. I haven't looked at the recent work on fusing, but we'd want toSeq to take part in that too, where applicable.

drewhk:
Ok. I close this for now but please reopen if there is some progress.

RichardBradley:
Thanks; I agree that a half-completed PR is a poor place to track this :-)
I have raised #18921 to track this enhancement.

Issue #17693, opened 2015-06-10
Provide a "Source.queue" as an alternative to `Source.actorRef` that can reject elements


jrudolph:
`Source.actorRef` basically implements a queue (by setting buffer sizes) that may overflow at any time. In case of overflow, only the consumer of the elements can be notified but not the sender.

If you need a queue that can reject elements if it overflows you need to implement it yourself.

See http://zuchos.com/blog/2015/06/10/more-reactive-publisher-aka-publisher-vol-2/ and https://github.com/Zuchos/akka-http-with-streams/blob/blogpost2/src/main/scala/pl/zuchos/example/actors/DataPublisher.scala

A possible signature could look like this:

```scala
def queue[T](bufferSize: Int): Source[T, T => Future[Boolean]] 
```

It would materialize to a function that can be used to try adding an element to the queue and returns a future that indicates if adding the element was successful.

jrudolph:
Also, it could make sense to abstract over the queue, so that you could use a PriorityQueue etc.

drewhk:
Make sense to me. We can provide a Sink version of this that has a similar materialized type, but returns a Future when the next element is available for consumption. These two simple abstractions would allow integration with existing send-ack actor patterns without much hassle.

fommil:
sweet! I'll definitely have to look at this in the next RC

ymeymann:
This would come handy, as we are trying to use Akka Http to receive data and push it to Kafka with https://github.com/softwaremill/reactive-kafka

cppexpert:
It doesn't work for me in 2.4.4. Could you please take a look ?
http://stackoverflow.com/questions/37113877/how-can-i-use-and-return-source-queue-to-caller-without-materializing-it

drewhk:
What do you mean does not work? Do you have a reproducer and a detailed bug description?

cppexpert:
I mean that I don't receive values from the stream in `.runFold/.runForEach`. In edit of my SO question I added self-contained code snippet in which I reproduced the issue. It doesn't even work with `Sink` for me. Maybe I'm using it incorrectly but in the past I used actorPublisher sink without any problems. So apparently it's something about `Source.queue` I don't understand. Code is this

```
val source = Source.queue[String](50, backpressure)
val sink = Sink.seq[String] // Sink.queue[String]()
val queue = source.toMat(sink)(Keep.left).run()
val queue1: SourceQueueWithComplete[String] = queue
for (i <- 1 until 10) {
[...]

Issue #17967, opened 2015-07-12
Create a variant of `Sink.actorRef` that uses simple acking for backpressure


fommil:
The complement of #17610 to ease migration paths without having to go full-on `ActorSubscriber` / `ActorPublisher`.

It would be good to have a way of creating a `Sink` from an existing `actor: ActorRef` that will only send a message to the underlying `actor` when it receives an `Ack` from that actor.

This will require some initial "Subscribe" message from streams to the actor, with an `Ack` response. All failure cases should just result in a `PoisonPill` (or whatever `Sink.actorRef` currently does).

If the user's actor sends a message without an `Ack`, fail... it's an illegal state.

rkuhn:
ping @drewhk

agolubev:
@drewhk, I can take this.
Please let me know if all in ticket description make sense and making something similar to ActorRefSink is a right direction for implementation.

rkuhn:
fixed in #18773

PR #18271, opened 2015-08-20
+str #17693 add Source.queue and Sink.queue

agolubev:
Approach with () => Future[Boolean] for Source or  () => Future[Out] for Sink is leaky.
When it's completed or failed there is no way to notify client that stream is stopped (at least with given API) - Future can only fail with timeout.
Ref #17693

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
@@ -1,0 +1,117 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.scaladsl
+
+import akka.actor.Status
+import akka.pattern.{ AskTimeoutException, pipe }
+import akka.stream.ActorMaterializer
+import akka.stream.testkit.Utils._
+import akka.stream.testkit.{ AkkaSpec, _ }
+
+import scala.concurrent.duration._
+import scala.concurrent.{ Await, Future }
+import scala.util.control.NoStackTrace
+
+class AcknowledgeSinkSpec extends AkkaSpec {
+  implicit val ec = system.dispatcher
+  implicit val mat = ActorMaterializer()
+
+  val ex = new RuntimeException("ex") with NoStackTrace
+
+  val noMsgTimeout = 300.millis
+
+  def assertSuccess(value: Any, fb: Future[Option[Any]]): Unit =
+    Await.result(fb, 1.second) should be(Some(value))
+
+  "An AcknowledgeSink" must {
+
+    "send the elements as result of future" in assertAllStagesStopped {
+      val queue = Source(List(1, 2, 3)).runWith(Sink.queue(3))
+      assertSuccess(1, queue.pull())
+      assertSuccess(2, queue.pull())
+      assertSuccess(3, queue.pull())
+      queue.pull().pipeTo(testActor)
+      expectMsg(None)
+    }
+
+    "allow to have only one future waiting for result in each point of time" in assertAllStagesStopped {
+      val probe = TestPublisher.manualProbe[Int]()
+      val queue = Source(probe).runWith(Sink.queue(3))
+      val sub = probe.expectSubscription()
+      val future = queue.pull()
+      val future2 = queue.pull()
+      an[IllegalStateException] shouldBe thrownBy { Await.result(future2, 300.millis) }
+
+      sub.sendNext(1)
+      future.pipeTo(testActor)
+      expectMsg(Some(1))
+
+      sub.sendComplete()
+    }
+
+    "wait for next element from upstream" in assertAllStagesStopped {
+      val probe = TestPublisher.manualProbe[Int]()
+      val queue = Source(probe).runWith(Sink.queue(3))
+      val sub = probe.expectSubscription()
+
+      queue.pull().pipeTo(testActor)
+      expectNoMsg(noMsgTimeout)
+
+      sub.sendNext(1)
+      expectMsg(Some(1))
+      sub.sendComplete()
+    }
+
+    "fail future on stream failure" in assertAllStagesStopped {
+      val probe = TestPublisher.manualProbe[Int]()
+      val queue = Source(probe).runWith(Sink.queue(3))
+      val sub = probe.expectSubscription()
+
+      queue.pull().pipeTo(testActor)
+      expectNoMsg(noMsgTimeout)
+
+      sub.sendError(ex)
+      expectMsg(Status.Failure(ex))
+    }
+
+    "fail future when stream failed" in assertAllStagesStopped {
+      val probe = TestPublisher.manualProbe[Int]()
+      val queue = Source(probe).runWith(Sink.queue(3, 100.milli))
+      val sub = probe.expectSubscription()
+      sub.sendError(ex) // potential race condition
+
+      val future = queue.pull()
+      future.onFailure { case e ⇒ e.getClass() should be(classOf[AskTimeoutException]); Unit }
+      future.onSuccess { case _ ⇒ fail() }
+
+      Await.ready(future, 1.second)
+    }
+
+    "timeout future when stream cannot provide data" in assertAllStagesStopped {
+      val probe = TestPublisher.manualProbe[Int]()
+      val queue = Source(probe).runWith(Sink.queue(3))
+      val sub = probe.expectSubscription()
+
+      queue.pull().pipeTo(testActor)
+      expectNoMsg(noMsgTimeout)
+
+      sub.sendNext(1)
+      expectMsg(Some(1))
+      sub.sendComplete()
+    }
+
+    "work when buffer is 0" in assertAllStagesStopped {
+      val probe = TestPublisher.manualProbe[Int]()
+      val queue = Source(probe).runWith(Sink.queue(0))
+      val sub = probe.expectSubscription()
+      sub.sendNext(1)
+
+      queue.pull().pipeTo(testActor)
+      sub.sendNext(2)
+      expectMsg(Some(2))
+      sub.sendComplete()
+    }
+
+  }
+}
--- a/akka-stream/src/main/scala/akka/stream/impl/AcknowledgePublisher.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/AcknowledgePublisher.scala
@@ -1,0 +1,86 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.impl
+
+import akka.actor.{ ActorRef, Props }
+import akka.stream.OverflowStrategy
+import akka.stream.OverflowStrategy._
+import akka.stream.actor.ActorPublisherMessage.Request
+import akka.stream.impl.AcknowledgePublisher.{ Rejected, Ok }
+
+/**
+ * INTERNAL API
+ */
+private[akka] object AcknowledgePublisher {
+  def props(bufferSize: Int, overflowStrategy: OverflowStrategy) =
+    Props(new AcknowledgePublisher(bufferSize, overflowStrategy))
+
+  case class Ok()
+  case class Rejected()
+}
+
+/**
+ * INTERNAL API
+ */
+private[akka] class AcknowledgePublisher(bufferSize: Int, overflowStrategy: OverflowStrategy)
+  extends ActorRefSourceActor(bufferSize, overflowStrategy) {
+
+  var backpressedElem: Option[ActorRef] = None
+
+  override def requestElem: Receive = {
+    case _: Request ⇒
+      // totalDemand is tracked by super
+      if (bufferSize != 0)
+        while (totalDemand > 0L && !buffer.isEmpty) {
+          //if buffer is full - sent ack message to sender in case of Backpressure mode
+          if (buffer.isFull) backpressedElem match {
+            case Some(ref) ⇒
+              ref ! Ok(); backpressedElem = None
+            case None ⇒ //do nothing
+          }
+          onNext(buffer.dequeue())
+        }
+  }
+
+  override def receiveElem: Receive = {
+    case elem if isActive ⇒
+      if (totalDemand > 0L) {
+        onNext(elem)
+        sendAck(true)
+      } else if (bufferSize == 0) {
+        log.debug("Dropping element because there is no downstream demand: [{}]", elem)
+        sendAck(false)
+      } else if (!buffer.isFull)
+        enqueueAndSendAck(elem)
+      else overflowStrategy match {
+        case DropHead ⇒
+          buffer.dropHead()
+          enqueueAndSendAck(elem)
+        case DropTail ⇒
+          buffer.dropTail()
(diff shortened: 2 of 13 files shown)


ktoso:
OK TO TEST

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+      Await.result(future, 200.millis).get should ===(1)
+      sub.sendComplete()
+    }
+
+    "wait for next element from upstream" in assertAllStagesStopped {
+      val probe = TestPublisher.manualProbe[Int]()
+      val receiveFun = Source(probe).runWith(Sink.queue(3))
+      val sub = probe.expectSubscription()
+      val future = receiveFun()
+      an[Exception] shouldBe thrownBy { Await.result(future, 200.millis) }
```
drewhk: how does this work? We expect an exception here, but at line 53 we expect a result.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+      Await.result(future, 200.millis).get should ===(1)
+      sub.sendComplete()
+    }
+
+    "fail future on stream failure" in assertAllStagesStopped {
+      val probe = TestPublisher.manualProbe[Int]()
+      val receiveFun = Source(probe).runWith(Sink.queue(3))
+      val sub = probe.expectSubscription()
+      val future = receiveFun()
+      an[Exception] shouldBe thrownBy { Await.result(future, 200.millis) }
```
drewhk: how come we expect an exception before we send the error?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+      sub.sendError(ex)
+      Await.ready(future, 200.millis)
+    }
+
+    "fail future when stream failed" in assertAllStagesStopped {
+      val probe = TestPublisher.manualProbe[Int]()
+      val receiveFun = Source(probe).runWith(Sink.queue(3, 100.milli))
+      val sub = probe.expectSubscription()
+      sub.sendError(ex) // potential race condition
+      val future = receiveFun()
```
drewhk: you can use awaitAssert to avoid the race condition here

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+      Await.ready(future, 200.millis)
```
drewhk: shouldn't this Future fail?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+  implicit val mat = ActorMaterializer()
+  val ex = new RuntimeException with NoStackTrace
+
+  def assertFuture(value: Any, fb: Future[Option[Any]]): Unit =
+    Await.result(fb, 200.millis) should be(Some(value))
+
+  def assertComplete(fb: Future[Option[Any]]): Unit =
+    Await.result(fb, 200.millis) should be(None)
+
+  "An AcknowledgeSink" must {
```
drewhk: I am a bit confused by the error scenarios in this test suite. Can you explain them a bit?

akka-stream/src/main/scala/akka/stream/impl/AcknowledgePublisher.scala
```diff
@@ -0,0 +1,62 @@
...
+      }
+  }
+
+  def enqueueAndSendAck(elem: Any): Unit = {
+    buffer.enqueue(elem)
+    sendAck(true)
+  }
+
+  def sendAck(ack: Boolean): Unit =
+    context.sender() ! ack
```
drewhk: I know it is an extra allocation to wrap the boolean, but we usually don't like naked types. Please introduce a wrapper message for that boolean.

akka-stream/src/main/scala/akka/stream/impl/AcknowledgeSubscriber.scala
```diff
@@ -0,0 +1,88 @@
...
+
+    case OnNext(elem) ⇒
+      if (maxBuffer != 0) {
+        buffer :+= elem
+        trySendElementDownstream()
+      } else requester match {
+        case Some(ref) ⇒
+          requester = None
+          ref ! Some(elem)
+        case None ⇒ log.debug("Dropping element because there is no downstream demand: [{}]", elem)
```
drewhk: shouldn't it buffer the element and backpressure upstream at this point?

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -184,4 +183,9 @@ object Sink extends SinkApply {
   def actorSubscriber[T](props: Props): Sink[T, ActorRef] =
     new Sink(new ActorSubscriberSink(props, DefaultAttributes.actorSubscriberSink, shape("ActorSubscriberSink")))
 
+  def queue[T](bufferSize: Int, timeout: FiniteDuration = 5.seconds): Sink[T, () ⇒ Future[Option[T]]] = {
```
drewhk: I know that the raw function type is elegant ("() ⇒ Future[Option[T]]") but I would prefer a Trait instead with a named method, "pull()" for example.

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -364,4 +362,9 @@ object Source extends SourceApply {
     new Source(new ActorRefSource(bufferSize, overflowStrategy, DefaultAttributes.actorRefSource, shape("ActorRefSource")))
   }
 
+  def queue[T](bufferSize: Int, overflowStrategy: OverflowStrategy, timeout: FiniteDuration = 5.seconds): Source[T, T ⇒ Future[Boolean]] = {
```
drewhk: I know that the raw function type is elegant ("() ⇒ Future[Boolean]") but I would prefer a Trait instead with a named method, "offer()" for example.

akka-stream/src/main/scala/akka/stream/impl/AcknowledgePublisher.scala
```diff
@@ -0,0 +1,62 @@
...
+          enqueueAndSendAck(elem)
+        case DropBuffer ⇒
+          buffer.clear()
+          enqueueAndSendAck(elem)
+        case DropNew ⇒
+          sendAck(false)
+        case Fail ⇒
+          onErrorThenStop(new Fail.BufferOverflowException(s"Buffer overflow (max capacity was: $bufferSize)!"))
+        case Backpressure ⇒
+        // there is a precondition check in Source.actorRefSource factory method
```
drewhk: This Source would be much more useful if Backpressure mode would be supported. I.e. when you offer an element, and the buffer just got full by enqueueing that element, then the future should only complete after a new buffer space got freed. This way the Future would propagate properly the backpressure information.

drewhk:
This ticket is also relevant here: https://github.com/akka/akka/issues/17967

agolubev:
Thanks for the feedback will look into later today.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
agolubev: This means feature will be succeed when element send from upstream but not earlier

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
agolubev: here again we are waiting any event from upstream. In this case it's failure

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
agolubev: yes, it should. Will add verification

akka-stream/src/main/scala/akka/stream/impl/AcknowledgeSubscriber.scala
agolubev: this line works when buffer size is 0 so we drop everything if there is no request waiting for element

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
agolubev: There are several scenarios when stream is finished or failed:
1) future has been received and upstream sent success - future returns None
2) upstream done with  success and only then re requesting next element. We'll receive timeout as stream is finished
3) we requested next element and then upstream sent failure. Our Future will fail
4) upstream done with failure and only then we requesting new element. We'll receive timeout as stream is finished

agolubev:
@drewhk, I done with fixed, rebased and added docs

agolubev:
@drewhk, can you check if this pull request good enough to merge?

drewhk:
yep, will review

akka-stream/src/main/scala/akka/stream/impl/AcknowledgeSubscriber.scala
drewhk: got it, thanks!

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+import scala.util.Failure
+import scala.util.control.NoStackTrace
+
+class AcknowledgeSinkSpec extends AkkaSpec {
+  import system.dispatcher
+  implicit val mat = ActorMaterializer()
+  val ex = new RuntimeException("ex") with NoStackTrace
+
+  def assertFuture(value: Any, fb: Future[Option[Any]]): Unit =
+    Await.result(fb, 200.millis) should be(Some(value))
```
drewhk: I think you should increase these timeouts to 1.second, they only matter anyway if the tests fail, but short timeouts might trigger on our Jenkins boxes depending on their load and mood :)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+      assertComplete(queue.pull())
+    }
+
+    "allow to have only one future waiting for result in each point of time" in assertAllStagesStopped {
+      val probe = TestPublisher.manualProbe[Int]()
+      val queue = Source(probe).runWith(Sink.queue(3))
+      val sub = probe.expectSubscription()
+      val future = queue.pull()
+      val future2 = queue.pull()
+      an[Exception] shouldBe thrownBy { Await.result(future2, 200.millis) }
```
drewhk: specify the subtype of the exception (it makes the test easier to understand. This is an IllegalStateException for example)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+      Await.result(future, 200.millis).get should ===(1)
+      sub.sendComplete()
+    }
+
+    "wait for next element from upstream" in assertAllStagesStopped {
+      val probe = TestPublisher.manualProbe[Int]()
+      val queue = Source(probe).runWith(Sink.queue(3))
+      val sub = probe.expectSubscription()
+      val future = queue.pull()
+      an[Exception] shouldBe thrownBy { Await.result(future, 200.millis) }
```
drewhk: This is a confusing case since you rely on a TimeoutException. More idiomatic would be:

    val future = queue.pull()
    future.pipeTo(testProbe)
    expectNoMsg(200.millis)
    sub.sendNext(1)
    expectMsg(...)

"expectNoMsg" describes more clearly what is expected here, compared to "an[Exception] shouldBe thrownBy Await".

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+      Await.result(future, 200.millis).get should ===(1)
+      sub.sendComplete()
+    }
+
+    "fail future on stream failure" in assertAllStagesStopped {
+      val probe = TestPublisher.manualProbe[Int]()
+      val queue = Source(probe).runWith(Sink.queue(3))
+      val sub = probe.expectSubscription()
+      val future = queue.pull()
+      an[Exception] shouldBe thrownBy { Await.result(future, 200.millis) }
```
drewhk: same as above

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+      val queue = Source(probe).runWith(Sink.queue(3, 100.milli))
+      val sub = probe.expectSubscription()
+      sub.sendError(ex) // potential race condition
+      val future = queue.pull()
+      future.onFailure { case e ⇒ e.getClass() should be(classOf[AskTimeoutException]); Unit }
+      future.onSuccess { case _ ⇒ fail() }
+      Await.ready(future, 200.millis)
+    }
+
+    "timeout future when stream cannot provide data" in assertAllStagesStopped {
```
drewhk: isn't this the same as "wait for next element from upstream"?

drewhk:
Looks like this is at the finish line, LGTM.

 - little tests cleanup
 - rebase
 - validation
 - merge
 - ???
 - profit

agolubev:
Thanks for the feedback

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
agolubev: You are right. Will remove this one

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
agolubev: that looks great. will fix test code

agolubev:
@drewhk, I finished from my side.

drewhk:
Thanks! Great work again @agolubev!

agolubev:
Cool. Thanks

Issue #18273, opened 2015-08-20
LeveldbStore numericId must not be used from Futures


patriknw:
numericId must only be called from the actor

we made a mistake in one of the refactorings, e.g. asyncReadHighestSequenceNumber

PR #18278, opened 2015-08-20
=per #18273 Make leveldb numericId thread safe

patriknw:


patriknw:
Refs #18273

akka-persistence/src/main/scala/akka/persistence/journal/leveldb/LeveldbIdMapping.scala
```diff
@@ -19,19 +19,30 @@ private[persistence] trait LeveldbIdMapping extends Actor { this: LeveldbStore 
 
   private val idOffset = 10
   private var idMap: Map[String, Int] = Map.empty
```
ktoso: I'd be tempted to wrap in atomic-ref instead... Maybe doesn't matter since db is the slowest thing here anyway hm...

ktoso:
I may be too-early-optimising, LGTM in any case

akka-persistence/src/main/scala/akka/persistence/journal/leveldb/LeveldbIdMapping.scala
patriknw: 2 things must be updated together, the idMap and the write to leveldb, see `numericId ` that is calling `writeIdMapping`.

I'm pretty sure leveldb is full of locks and synchronized anyway.

Additionally `numericId` is normally a very quick operation (lookup in hashmap), i.e. low contention, apart from when new persistenceIds are created (and then the contention will be the db anyway).

patriknw:
The failure was a timeout in `AllPersistenceIdsSpec`. I don't think it is related to this PR, but we might have to increase the default timeouts for the tests in persistence.query, as we have in persistence. I'll do that in the other PR.

patriknw:
PLS BUILD

akka-persistence/src/main/scala/akka/persistence/journal/leveldb/LeveldbIdMapping.scala
ktoso: thanks for the reasoning behind it :) Makes sense

Issue #17372, opened 2015-05-04
deprecate blocking mailboxes


rkuhn:
We should eventually remove the blocking versions and only allow non-blocking bounded/unbounded mailboxes. The reason is that the blocking part is not location transparent.

rkuhn:
… and also document the non-blocking bounded queue properly

PR #18280, opened 2015-08-20
=act #17372 deprecate blocking mailboxes

ktoso:
Resolves https://github.com/akka/akka/issues/17372

Please review @patriknw, I am not 100% sure if that covers everything and if we're happy with such logging of the deprecation warning.

### DIFF
--- a/akka-actor/src/main/resources/reference.conf
+++ b/akka-actor/src/main/resources/reference.conf
@@ -420,6 +420,7 @@ akka {
         mailbox-type = "akka.dispatch.UnboundedMailbox"
       }
 
+      # deprecated (because relies on blocking)
       bounded-queue-based {
         # FQCN of the MailboxType, The Class of the FQCN must have a public
         # constructor with (akka.actor.ActorSystem.Settings,
@@ -427,6 +428,7 @@ akka {
         mailbox-type = "akka.dispatch.BoundedMailbox"
       }
 
+      # deprecated (because relies on blocking)
       unbounded-deque-based {
         # FQCN of the MailboxType, The Class of the FQCN must have a public
         # constructor with (akka.actor.ActorSystem.Settings,
@@ -434,6 +436,7 @@ akka {
         mailbox-type = "akka.dispatch.UnboundedDequeBasedMailbox"
       }
 
+      # deprecated (because relies on blocking)
       bounded-deque-based {
         # FQCN of the MailboxType, The Class of the FQCN must have a public
         # constructor with (akka.actor.ActorSystem.Settings,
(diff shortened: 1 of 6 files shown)


akka-actor/src/main/resources/reference.conf
```diff
@@ -420,6 +420,7 @@ akka {
         mailbox-type = "akka.dispatch.UnboundedMailbox"
       }
 
+      # deprecated (because relies on blocking)
```
patriknw: I don't think the purpose of the ticket was to deprecate all bounded mailboxes. Bounded mailbox can be good if used with zero `mailbox-push-timeout-time`.

akka-actor/src/main/scala/akka/Sample.scala
```diff
@@ -0,0 +1,31 @@
...
+     * with the actor logic.
+     */
+    override def receive: Receive = {
+      case x ⇒ x
+    }
+  }).withMailbox("akka.actor.mailbox.unbounded-deque-based"))
+
+  readLine()
+  system.terminate()
+}
```
patriknw: was this committed accidentally?

ktoso: yes, please ignore

akka-actor/src/main/resources/reference.conf
ktoso: Right, that's something I was wondering about... Some of the mailboxes do have a timeout parameter - those we want to keep then?

patriknw: so there are some that don't have the timeout?

ktoso: My mistake, all bounded queues have a timeout (timeout handling added via a trait that I missed to look into). But if it's not those we want to deprecate, I don't know which ones.

patriknw: I can't see any. `capacity` and `pushTimeOut` always comes together. I think the ticket is about emitting a warning when pushTimeout != 0. Nothing should be deprecated as far as I can see. Well, we could start removing the `pushTimeOut` parameter and (use 0 as default). That would be very complicated to do in a binary compatible way, since the mailboxes are case classes :-(

ktoso: Hm... I thought of that initially but seems I started from the wrong side (from inside the mailboxes), checking the timeout could be done while creating the mailbox... I'll give it a shot

ktoso:
Closing this PR, will solve it another way

akka-actor/src/main/resources/reference.conf
patriknw: good, thanks

Issue #15412, opened 2014-06-18
Cluster aware routers together with allow-local-routees and cluster roles behaves inconsistently


drewhk:
See the user report here https://groups.google.com/forum/#!topic/akka-user/f81POHJVWu8
(Together with code that reproduces the behavior)

1. if allow-local-routees is set to true, then local routees are added even though the role doesn't match what's set in the use-role property.
2. As in above, depending on how and when you start the router it might have the local routee or not. If you start the router before the worker node joins it will have the local routee before it joins but not after. If you start the router after the worker node joins it will always contain the local routee. So in either way this is inconsistent.

PR #18281, opened 2015-08-20
=clu #15412 Add paths(system) method to Group router

patriknw:
to be able to use the role correctly in cluster aware routers

This solution is very similar to what we did for nrOfInstances
in Pool routers, see PR https://github.com/akka/akka/pull/15822

### DIFF
--- a/akka-actor/src/main/scala/akka/routing/RouterConfig.scala
+++ b/akka-actor/src/main/scala/akka/routing/RouterConfig.scala
@@ -128,9 +128,16 @@ private[akka] trait PoolOverrideUnsetConfig[T <: Pool] extends Pool {
  * Java API: Base class for custom router [[Group]]
  */
 abstract class GroupBase extends Group {
-  def getPaths: java.lang.Iterable[String]
+  @deprecated("Implement getPaths with ActorSystem parameter instead", "2.4")
+  def getPaths: java.lang.Iterable[String] = null
 
+  @deprecated("Use paths with ActorSystem parameter instead", "2.4")
   override final def paths: immutable.Iterable[String] = immutableSeq(getPaths)
+
+  def getPaths(system: ActorSystem): java.lang.Iterable[String]
+
+  override final def paths(system: ActorSystem): immutable.Iterable[String] =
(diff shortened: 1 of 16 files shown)


patriknw:
Refs #15412

akka-actor/src/main/scala/akka/routing/RouterConfig.scala
```diff
@@ -140,7 +140,10 @@ abstract class GroupBase extends Group {
  */
 trait Group extends RouterConfig {
 
-  def paths: immutable.Iterable[String]
+  @deprecated("Implement paths with ActorSystem parameter instead", "2.4")
+  def paths: immutable.Iterable[String] = null
+
+  def paths(system: ActorSystem): immutable.Iterable[String]
```
ktoso: oh, great that the same trick as the other ticket in this area works for this ticket as well :)

project/MiMa.scala
```diff
@@ -417,9 +417,14 @@ object MiMa extends AutoPlugin {
       ProblemFilters.exclude[MissingMethodProblem]("akka.remote.testkit.MultiNodeSpec#Replacement.akka$remote$testkit$MultiNodeSpec$Replacement$$$outer"),
 
 
-      // method nrOfInstances(akka.actor.ActorSystem)Int in trait akka.routing.Pool does not have a correspondent in old version
+      // method nrOfInstances(akka.actor.ActorSystem) in trait akka.routing.Pool does not have a correspondent in old version
       // ok to exclude, since we don't call nrOfInstances(sys) for old implementations
       ProblemFilters.exclude[MissingMethodProblem]("akka.routing.Pool.nrOfInstances"),
+      
+      // method nrOfInstances(akka.actor.ActorSystem) in trait akka.routing.Group does not have a correspondent in old version
```
ktoso: method `paths(ActorSystem)`?

ktoso:
LGTM

project/MiMa.scala
patriknw: copy pasta, thx

PR #18289, opened 2015-08-21
=cls #17846 Use provided scope for the distributed-data dependency

patriknw:
The scope in the pom.xml was "compile".

### DIFF
--- a/project/AkkaBuild.scala
+++ b/project/AkkaBuild.scala
@@ -135,8 +135,12 @@ object AkkaBuild extends Build {
   lazy val clusterSharding = Project(
     id = "akka-cluster-sharding",
     base = file("akka-cluster-sharding"),
+    // TODO akka-distributed-data dependency should be provided in pom.xml artifact.
+    //      If I only use "provided" here it works, but then we can't run tests.
+    //      Scope "test" is alright in the pom.xml, but would have been nicer with
+    //      provided.
     dependencies = Seq(cluster % "compile->compile;test->test;multi-jvm->multi-jvm",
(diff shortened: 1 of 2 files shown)


project/AkkaBuild.scala
```diff
@@ -136,7 +136,7 @@ object AkkaBuild extends Build {
     id = "akka-cluster-sharding",
     base = file("akka-cluster-sharding"),
     dependencies = Seq(cluster % "compile->compile;test->test;multi-jvm->multi-jvm",
-        persistence % "compile;test->provided", distributedData % "compile;test->provided", clusterTools)
+        persistence % "compile;test->provided", distributedData % "provided;test", clusterTools)
```
patriknw: Now it actually becomes scope "test" in the pom.xml. It should be "provided" in the pom.xml, but I don't know how to do that. If I only use "provided" here it works, but then we can't run tests. Scope "test" is alright in the pom, so if we don't have any better suggestions we'll use that.

patriknw:
Refs #17846

project/AkkaBuild.scala
ktoso: have you tried `provided;test->provided`?

ktoso: (Can you add a comment here describing what happens in the pom? Maybe some sbt wizard could help here at some point)

ktoso:
LGTM in any case (would like the TODO comment though, to remind us why/how it works)

project/AkkaBuild.scala
patriknw: yes, I think I tried that and several other combinations. I'll add the TODO

Issue #18287, opened 2015-08-21
failed: ClusterShardingSpec ddata mode


patriknw:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3480/consoleFull

    [JVM-6] - must recover coordinator state after coordinator crash *** FAILED ***
    [JVM-6]   akka://ClusterShardingSpec/user/counterRegion/3/3 did not equal akka.tcp://ClusterShardingSpec@a5.moxie:51499/user/counterRegion/3/3 (ClusterShardingSpec.scala:497)

smlin:
@patriknw I reproduced it. When the nodes are joins the coordinator sometimes receive the info about the new regions faster then the replicator receives `MemberUp` message. So that info simple does not replicated. It could be partially fixed by making a rejection of registration for the regions from the nodes which are not `Up`.  A minimal chance of inconsistency anyway exists.

I did a fix. https://github.com/akka/akka/pull/18293

PR #18293, opened 2015-08-21
=cls #18287 failed: ClusterShardingSpec ddata mode

oseval:


ktoso:
OK TO TEST

patriknw:
Refs #18287

patriknw:
@smlin Thanks for looking into the issue. It's not a trivial problem and will do some thinking about it tomorrow.

patriknw:
@smlin I looked at the logs from the first failure. After crashing first, the coordinator is started at the fifth node. It gets the initial `GetSuccess` from the replicator and activates itself:

    Sharding Coordinator was moved to the active state State(Map(),Map(),Set(),Set())

State is empty. I agree that that can be because the second node is already crashed and the third and fourth node are not full Up members yet, i.e. the first did not replicate the shard locations in the previous test step to any other node (it was only on first). Then we crash first and the information is lost, i.e. empty state.

I agree that ignoring Register messages from nodes that are not full members yet makes sense.

I might want to add an additional safety by delaying the initial getState until it has received a the initial `CurrentClusterState` and made sure that replica count via `Replicator.ReplicaCount` is >= members.size.
I will do that in a separate pull request. I have a few other things to improve in ClusterSharding anyway.

Let's merge this as the first step.

smlin:
@patriknw I have some thoughts about the coordinator:
1) The Coordinator is just a judge and not a sharding state holder. 
2) The Coordinator could not performs its duties when it does not have full info about all regions  
3) The Coordinator could recover all state from the regions instead of any storage, but it must await to get info from all nodes/regions before moving to the active state. 

I think possible to realize a synchronization of the coordinator with all nodes/regions before moving to the active state. But if one or more nodes temporarily unavailable then such synchronization will be is a long process and during synchronization all cluster-sharding functionality will be locked. 

I don't know how much is bad doing such synchronization on a recovering of the coordinator.

patriknw:
That is a rather different design, with its own set of problems. I don't have the bandwidth to see all implications of that now.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -403,23 +394,33 @@ abstract class ShardCoordinator(typeName: String, settings: ClusterShardingSetti
...
-      aliveRegions += region
-      if (state.regions.contains(region))
-        region ! RegisterAck(self)
-      else {
-        gracefulShutdownInProgress -= region
-        update(ShardRegionRegistered(region)) { evt ⇒
-          val firstRegion = state.regions.isEmpty
-          state = state.updated(evt)
-          context.watch(region)
+      if (region.path.address.hostPort == self.path.address.hostPort || members.exists(_.hostPort == region.path.address.hostPort)) {
```
patriknw: why use `hostPort`?

    if (region.path.address == self.path.address || members.contains(region.path.address)) {

smlin: I was not sure that would be working properly in all cases. Changed.

patriknw:
LGTM

patriknw:
@smlin thanks for contributing, again

patriknw:
Refs #18287

Issue #18275, opened 2015-08-20
[akka-streams] Java DSL: Flow.of(MyClass.class).via(otherFlow) throws ClassCastException


jypma:
```
import akka.stream.javadsl.Flow;

Flow<MyClass,Object,?> otherFlow = ...

Flow<MyClass,Object,?> myFlow = Flow.of(MyClass.class).via(otherFlow)
```
results in
```
akka.stream.javadsl.Flow cannot be cast to akka.stream.scaladsl.Flow	
	at  akka.stream.scaladsl.Flow.viaMat(Flow.scala:78)
	at akka.stream.scaladsl.Flow.via(Flow.scala:59)
	at akka.stream.javadsl.Flow.via(Flow.scala:99)
```
because `scaladsl.Flow:78` assumes that if `this.isIdentity` is true (which it is on the Java version), that `this` must in fact be an instance of `scalads.Flow` (which it isn't).

ktoso:
Ouch, thanks for reporting!

PR #18300, opened 2015-08-24
+str #18275 Java DSL: Flow.of(MyClass.class).via(otherFlow) throws CCE

agolubev:
Ref #18275

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
@@ -3,29 +3,22 @@
  */
 package akka.stream.scaladsl
 
-import scala.language.higherKinds
-
 import akka.event.LoggingAdapter
-import akka.stream.impl.Stages.{ Recover, MaterializingStageFactory, StageModule }
-import akka.stream.impl.StreamLayout.{ EmptyModule, Module }
-import akka.stream._
 import akka.stream.Attributes._
-import akka.stream.stage._
-import akka.stream.impl.{ Stages, StreamLayout }
+import akka.stream._
 import akka.stream.impl.SplitDecision._
 import akka.stream.impl.Stages.{ DirectProcessor, MaterializingStageFactory, StageModule }
 import akka.stream.impl.StreamLayout.{ EmptyModule, Module }
+import akka.stream.impl.{ Stages, StreamLayout }
+import akka.stream.stage._
 import akka.util.Collections.EmptyImmutableSeq
-import org.reactivestreams.{ Subscription, Publisher, Subscriber, Processor }
-import org.reactivestreams.Processor
-import scala.annotation.implicitNotFound
+import org.reactivestreams.{ Processor, Publisher, Subscriber, Subscription }
+
 import scala.annotation.unchecked.uncheckedVariance
 import scala.collection.immutable
-import scala.concurrent.duration.{ Duration, FiniteDuration }
 import scala.concurrent.Future
+import scala.concurrent.duration.{ Duration, FiniteDuration }
 import scala.language.higherKinds
-import akka.stream.stage._
-import akka.stream.impl.{ Stages, StreamLayout, FlowModule }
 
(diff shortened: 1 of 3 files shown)


ktoso:
OK TO TEST

ktoso:
LGTM, thanks a lot!

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -75,8 +68,13 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
...
    */
   def viaMat[T, Mat2, Mat3](flow: Graph[FlowShape[Out, T], Mat2])(combine: (Mat, Mat2) ⇒ Mat3): Flow[In, T, Mat3] = {
-    if (this.isIdentity) flow.asInstanceOf[Flow[In, T, Mat2]].mapMaterializedValue(combine(().asInstanceOf[Mat], _))
-    else {
+    if (this.isIdentity) {
+      val flowInstance: Flow[In, T, Mat2] = if (flow.isInstanceOf[javadsl.Flow[In, T, Mat2]])
+        flow.asInstanceOf[javadsl.Flow[In, T, Mat2]].asScala
+      else
+        flow.asInstanceOf[Flow[In, T, Mat2]]
+      flowInstance.mapMaterializedValue(combine(().asInstanceOf[Mat], _))
```
drewhk: why is this step needed?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -121,8 +119,12 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
    * Sink into the materialized value of the resulting Sink.
    */
   def toMat[Mat2, Mat3](sink: Graph[SinkShape[Out], Mat2])(combine: (Mat, Mat2) ⇒ Mat3): Sink[In, Mat3] = {
-    if (isIdentity) sink.asInstanceOf[Sink[In, Mat3]]
-    else {
+    if (isIdentity) {
+      if (sink.isInstanceOf[javadsl.Sink[In, Mat3]])
+        sink.asInstanceOf[javadsl.Sink[In, Mat3]].asScala
+      else
+        sink.asInstanceOf[Sink[In, Mat3]]
```
drewhk: and why here it is not needed?

drewhk:
LGTM, very nice catch! The only thing I want to know is the need for mapping the materialized value

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
agolubev: this is adding consistency to API GraphConcatSpec test case "work with Flow DSL". Mat for identity is Unit. So in case we'll call Identity.viaMat(flow)(Keep.both) it will return (Unit, matOfFlow)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
agolubev: We'll need calling combine here actually. I'll try to add some test case tomorrow.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: ah, this line was there earlier, I missed that. Thanks for the explanation.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: thanks!

agolubev:
Ok, committed fix for toMat and Keep.both. Let me know if it makes sense

drewhk:
Everything is OK now, please squash then I will merge after validation

drewhk:
Another great contribution, thanks! :)

Issue #18299, opened 2015-08-23
Persistence - Register event adapters @ runtime


q42mark:
I think it's quite common to tag events per aggregate type. With the current setup, you would need to write a EventAdapter per aggregate type. If you have several aggregate types, this results in boilerplate code. I've made an attempt to cut that down by introducing a `Tag` type which contains a tag name as string and the actual type. It would be possible create a the `WriteEventAdapter` at runtime by using the `Tag`. See the example below: 

```scala
implicit val actorSystem = ActorSystem()
implicit val journal = PersistenceQuery(actorSystem).readJournalFor(LeveldbReadJournal.Identifier)

// Containing the return type + the tag name
trait Tag {
  type Type
  val manifest: Manifest[Type]
  val tag: String
}

object Tag {
  def apply[T : Manifest](t: String) = new Tag {
    type Type = T
    val manifest = implicitly[Manifest[T]]
    val tag = t
  }
}

//A event type persisted by a PersistentActor
sealed trait UserEvent {
  val userId: Int
}

//Create a Tag for userEvent
val userEvent = Tag[UserEvent]("userEvent")

//Utility function to create a WriteEventAdapter based on tag
def tagType(t: Tag) = new WriteEventAdapter {
  override def manifest(event: Any): String = ""
  override def toJournal(event: Any): Any = {
    //not sure if this works, but we cannot use match/case here (type erasure)
    if(t.manifest.canEqual(event)) Tagged(event, Set(t.tag))
    else event
  }
}


//Would be nice to register this WriteEventAdapter to persistence, however it's not a class
val userEventAdapter = tagType(userEvent)

//This is a utility function to get a stream of events by the Tag, just to illustrate the use of Tag
def eventsBy(tag: Tag, offset: Long = 0L)(implicit readJournal: ReadJournal) =
  readJournal.query(EventsByTag(tag.tag, offset)).map(_.event.asInstanceOf[tag.Type])

//Query userEvents
eventsBy(userEvent).map(_.userId)
```

However we cannot register a `WriteEventAdapter` at runtime. At the moment its only possible to register EventAdapters by using the HOCON config: http://doc.akka.io/docs/akka/2.4.0-RC1/scala/persistence.html#event-adapters-scala

Would it be possible to have `Persistence(system).registerWriteAdapter(t: WriteEventAdapter)` or something like that?

ktoso:
Good idea about dynamically registering adapters. This was not implemented yet but definitely was the intent.
[...]

PR #18301, opened 2015-08-24
+per #18299 allow registering user-created instances of EventAdapters

ktoso:
Refs https://github.com/akka/akka/issues/18299

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/Persistence.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/Persistence.scala
@@ -5,16 +5,17 @@
 package akka.persistence
 
 import java.util.concurrent.atomic.AtomicReference
+
 import akka.actor._
-import akka.dispatch.Dispatchers
 import akka.event.{ Logging, LoggingAdapter }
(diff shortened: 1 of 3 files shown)


patriknw:
Why do we need this?
How do we prevent race conditions when starting up (e.g. how do we make sure that user code for registering adapters is run before persistent actors managed by cluster sharding are started)?

Fristi:
@patriknw I see the problems of registering the adapters at runtime. Isn't it possible to have a fluent builder API upon creating an actor system to register them via code, or is it not the right place to do for some reason? I can imagine this is not trivial to do, but configuring serializers and adapters by configuration requires a lot of boilerplate (can't abstract over it) and it is not compiled (no static checks).

patriknw:
I think we have to start out simple here. Configuration of event adapters is in line with how other such things are defined in Akka.

patriknw:
@ktoso do you agree?

ktoso:
I'm OK with dropping this PR. Event adapters should usually be something one thinks about long-term-ish anyway. At run-time created adaptation code can reside inside a given adapter anyway, so it's possible to create adapters inside a "MyDynamicAdapter" already, without exposing so much complexity in API.

patriknw:
good point about that the dynamic/advanced parts can reside inside a more general adapter

Issue #18302, opened 2015-08-24
Backport ClusterClient failure detection improvement


patriknw:
Backport of #13909 to 2.3.x

For cluster clients it should be possible to re-establish a connection after a long network partition, without restarting the system.

PR #18304, opened 2015-08-24
=con #18302 Backport ClusterClient failure detection improvement

patriknw:
Note that this implementation of the ClusterClient cannot be used with old server, i.e. 2.3.12 and 2.3.13 compatibility. I think that is fair enough since this is in contrib and the old implementation was seriously broken.

### DIFF
--- a/akka-contrib/src/main/scala/akka/contrib/pattern/ClusterClient.scala
+++ b/akka-contrib/src/main/scala/akka/contrib/pattern/ClusterClient.scala
@@ -29,6 +29,8 @@ import akka.routing.ConsistentHash
 import akka.routing.MurmurHash
 import akka.actor.Stash
 import akka.actor.Cancellable
+import akka.remote.DeadlineFailureDetector
+import akka.actor.DeadLetterSuppression
 
 object ClusterClient {
 
@@ -79,6 +81,8 @@ object ClusterClient {
    */
   private[pattern] object Internal {
     case object RefreshContactsTick
+    case object HeartbeatTick
+    val Heartbeat = "H"
   }
 }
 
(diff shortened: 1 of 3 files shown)


patriknw:
Refs #18302

akka-contrib/src/main/scala/akka/contrib/pattern/ClusterClient.scala
```diff
@@ -311,10 +332,14 @@ object ClusterReceptionist {
...
   private[pattern] object Internal {
     @SerialVersionUID(1L)
-    case object GetContacts
+    case object GetContacts extends DeadLetterSuppression
     @SerialVersionUID(1L)
     case class Contacts(contactPoints: immutable.IndexedSeq[ActorSelection])
     @SerialVersionUID(1L)
+    case object Heartbeat extends DeadLetterSuppression
+    @SerialVersionUID(1L)
+    case object HeartbeatRsp extends DeadLetterSuppression
```
patriknw: Regarding the 2.3.12 vs 2.3.13 compatibility, we could support 2.3.12 client with 2.3.13 server if we use `Identify` and `ActorIdentity` instead of these new heartbeat messages. WDYT?

ktoso:
LGTM from my side, I don't have strong feelings about the optimisation mentioned.
One one hand it's more compat, on the other through it may be more confusing and adds a bit more traffic to the "heartbeating"... If possible to avoid special casing the 2.3 branch here with this logic I'd opt for that

patriknw:
The idea with using the `Identify` messages was based on sending those instead of the specialised heartbeat messages, so it should not be extra traffic.

patriknw:
changed to Identify/ActorIdentity messages as heartbeats: 490e42d

Issue #18303, opened 2015-08-24
MatchError in AsyncWriteJournal


WellingR:
Using akka-persistence_2.11-2.4.0-RC1
I see some MatchErrors occur in the log:

    scala.MatchError: Success(DeleteMessagesSuccess(1)) (of class scala.util.Success)
        at akka.persistence.journal.AsyncWriteJournal$$anonfun$1$$anonfun$applyOrElse$11.apply(AsyncWriteJournal.scala:156)
        at akka.persistence.journal.AsyncWriteJournal$$anonfun$1$$anonfun$applyOrElse$11.apply(AsyncWriteJournal.scala:156)
        at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:32)
        at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:55)
        at akka.dispatch.BatchingExecutor$BlockableBatch$$anonfun$run$1.apply$mcV$sp(BatchingExecutor.scala:91)
        at akka.dispatch.BatchingExecutor$BlockableBatch$$anonfun$run$1.apply(BatchingExecutor.scala:91)
        at akka.dispatch.BatchingExecutor$BlockableBatch$$anonfun$run$1.apply(BatchingExecutor.scala:91)
        at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:72)
        at akka.dispatch.BatchingExecutor$BlockableBatch.run(BatchingExecutor.scala:90)
        at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:39)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)

It appears to be caused by the following code 
```scala
    case d @ DeleteMessagesTo(persistenceId, toSequenceNr, persistentActor) ⇒
      breaker.withCircuitBreaker(asyncDeleteMessagesTo(persistenceId, toSequenceNr)) map {
        case _ ⇒ DeleteMessagesSuccess(toSequenceNr)
      } recover {
        case e ⇒ DeleteMessagesFailure(e, toSequenceNr)
      } pipeTo persistentActor onComplete {
        case _ if publish ⇒ context.system.eventStream.publish(d)
      }
```
[...]

PR #18305, opened 2015-08-24
=per #18303 avoid causing match error in AsyncWriteJournal

ktoso:
Fixes https://github.com/akka/akka/issues/18303

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
@@ -145,7 +145,7 @@ trait AsyncWriteJournal extends Actor with WriteJournalBase with AsyncRecovery {
         }.recover {
           case e ⇒ ReplayMessagesFailure(e)
         }.pipeTo(replyTo).onSuccess {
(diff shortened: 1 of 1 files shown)


akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
```diff
@@ -144,8 +144,8 @@ trait AsyncWriteJournal extends Actor with WriteJournalBase with AsyncRecovery {
           highSeqNr ⇒ RecoverySuccess(highSeqNr)
         }.recover {
           case e ⇒ ReplayMessagesFailure(e)
-        }.pipeTo(replyTo).onSuccess {
-          case _ if publish ⇒ context.system.eventStream.publish(r)
+        } pipeTo replyTo onSuccess {
```
patriknw: don't mix infix and dot notation

patriknw:
LGTM

akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
ktoso: oh, now I see why that notation was used here. Reverted this line

Issue #18306, opened 2015-08-24
ToC links require two clicks


patriknw:
I have to click twice on the table of contents on the right side of the pages to jump to the section.

ktoso:
Hmm... I can confirm it, surprisingly "on the same sources" it does cause problems when viewed locally, it does when viewed on the akka site.

PR #18307, opened 2015-08-24
=doc #18306 resolve problem with navigating using right-hand TOC

ktoso:
Resolves https://github.com/akka/akka/issues/18306

The right-hand-side TOC navigation now properly works - there was too much stuff going on at the same click()


patriknw:
LGTM (I guess)

ktoso:
Yeap, habe verified it, also the http://doc.akka.io/docs/akka/2.4.0-RC1/scala/persistence.html work properly now.

Issue #13584, opened 2013-09-05
Forced join when network split


patriknw:
imported from https://www.assembla.com/spaces/akka/tickets/3584

The leader cannot move joining nodes to Up when there are unreachable nodes (no convergence). Some users might want to add more nodes also when there is a network split.
Probably those nodes should have a special Up status, but be taken in use by for example routers. They are then moved to real Up by the leader on convergence.

patriknw:
Returning to backlog due to lack of interest/priority. My prototype can be revived from https://github.com/akka/akka/issues/17625

patriknw:
Taking note of that we got a user request for something similar as this issue. Discussed in #18067

hveiga:
Hello @patriknw, I see this got moved to Backlog due to lack of interest/priority. I would like to see this implemented as part of 2.4.0 release (if possible). I am happy to help wrapping up whatever is remaining and create a pull request, only thing is that I would just need a bit of guidance. Is that something you would be interested in?

patriknw:
Thanks for wanting to help out. We need that! Have you looked at my prototype in https://github.com/akka/akka/pull/17625 ? 

The big question is what we want? The immediate answer is of course "we want it all, let members join and leave however they want, also during network partitions". That is a rather big change to how the cluster membership is currently designed and would be major undertaking.

However, if the answer is more specific we might be able to support some scenarios, as I tried in the prototype.

WDYT?

hveiga:
I have been looking into the code and I believe that's what I had in mind. Another Member status where a node could join the cluster even if there is no convergence. Once convergence is reached, leader will move the Member from Weakly Up to Up. I have created this transition schema:

![weaklyup](https://cloud.githubusercontent.com/assets/1147034/9388639/57aea766-472c-11e5-9fbc-1b04e13b2555.png)

I believe that is what we are trying to achieve. Can you double check?

I believe everything is pretty much done except for the config parameter that we will be needed to enable or disable this feature to don't break Gossips coming from 2.3.X that don't know about WeaklyUp. Also, I believe there are some tests failing and maybe we need to merge the base 2.4-M3 (or the latest) because I believe there was some changes from where you implemented this.

Thanks and let me know your thoughts on this :)

patriknw:
Sounds good to me. I'll try to talk Roland and team next week to see if we can come to some conclusion.
Thanks!

[...]

PR #18308, opened 2015-08-24
+clu #13584 Adding option to enable/disable weaklyup. 

hveiga:


akka-cluster/src/main/resources/reference.conf
```diff
@@ -34,6 +34,11 @@ akka {
     # in case of a network partition the persistent actors in the non-surviving partitions
     # must be stopped before corresponding persistent actors are started somewhere else.
     down-removal-margin = 20s
+
+    # Allow new members to join the cluster when there is no convergence.
```
patriknw: strictly speaking it's already possible to join (status = Joining), so we should try to find a better description and property name

patriknw:
@hveiga I have talked to team today and we are interested in going forward with this feature. Could you open a new pull request targeted at master. I don't mind if you squash mine and your commits and that you are the author of it.

akka-cluster/src/main/resources/reference.conf
```diff
@@ -34,6 +34,11 @@ akka {
...
+    # These new members will have the state 'WeaklyUp' until convergence is reached and leader
+    # can move them to 'Up'. This feature must be off if some members are running Akka 2.3.X.
+    allow-join-new-members-while-no-convergence = off
```
patriknw: I agree that we should put this under a feature flag instead of trying to find a compatible protocol.

hveiga: Sure, I will change the name of the flag to be more strict about the status and create a pull request to master. Hopefully I can do it by the end of the day today. Thanks for reviewing!

PR #18313, opened 2015-08-25
WIP Octopus

drewhk:


akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -0,0 +1,167 @@
...
+
+  val NoEvent = -1
+}
+
+class GraphInterpreter(private val assembly: GraphAssembly) {
+  import GraphInterpreter._
+
+  val connectionStates = Array.fill[Any](assembly.connectionCount)(null)
+  private val mask = 255
+  private val eventQueue = Array.ofDim[Int](256)
```
drewhk: this array will be dynamically sized

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -0,0 +1,167 @@
...
+      assembly.in(i).interpreter = this
+      assembly.in(i).id = i
+      assembly.out(i).interpreter = this
+      assembly.out(i).id = i
+      i += 1
+    }
+  }
+
+  private def dequeue(): Int = {
+    if (queueHead == queueTail) -1
```
rkuhn: use NoEvent here

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -0,0 +1,167 @@
...
+    else {
+      val elem = eventQueue(queueHead & mask)
```
rkuhn: should clear eventQueue as well

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -0,0 +1,167 @@
...
+  private var queueHead: Int = 0
+  private var queueTail: Int = 0
+
+  def init(): Unit = {
+    var i = 0
+    while (i < assembly.connectionCount) {
+      assembly.in(i).interpreter = this
+      assembly.in(i).id = i
+      assembly.out(i).interpreter = this
+      assembly.out(i).id = i
```
rkuhn: this will need to invoke an `init` hook within the GraphStage so that demand can be generated immediately where required

rkuhn:
This is looking very promising, in particular the interpreter itself is extremely simple. When it comes to the API for GraphStage the loss of static verification is of course caused by its increased flexibility, but I would like to retain the compiler enforced invariants of the OneBoundedInterpreter in the normal Stages—whether we swap out the execution model for those is a purely internal question that can be tackled at a later point in time.

On your question about how to represent GraphStage as reusable blueprint I can only offer that it must be a Graph with a proper Shape in order to use all the nice wiring DSL support. The logic would then be part of the AST node as well, so that the materializer can then generate and initialize the GraphAssembly. This is probably no news for you, but I wanted to spell it out anyway ;-)

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: there will be preStart and postStop.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: yep, noticed too

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: well, it is an integer so there is no leak here. Can be left -1 though just to be sure.

drewhk:
> whether we swap out the execution model for those is a purely internal question that can be tackled at a later point in time.

That was the idea, there should be not much difficulty to run existing stages on this interpreter.

> This is probably no news for you, but I wanted to spell it out anyway ;-)

Well, the details will be the interesting part :)

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
rkuhn: ah, right intuition but wrong place: the state that needs clearing is in the associated `connectionStates` slot (if it contains an element or an exception, at least)

drewhk: yes, but it can be only cleared once it has been processed, so it should be done along L150

rkuhn: yes, exactly

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/GraphInterpreterSpec.scala
```diff
@@ -0,0 +1,551 @@
...
+      lastEvents() should ===(Set(RequestOne(source)))
+
+      sink.requestOne()
+      lastEvents() should ===(Set(OnNext(sink, 2)))
+
+      source.onNext(3)
+
+      //      // Unstuck it
+      //      sink.requestOne()
+      //      lastEvents() should ===(Set(OnNext(sink, 2)))
```
patriknw: is this a FIXME, TODO, or removal?

drewhk: this is the cycle that I intended to be divergent but it is not. This branch is still WIP so things are in an intermediate state.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -0,0 +1,167 @@
...
+  private val mask = 255
+  private val eventQueue = Array.ofDim[Int](256)
+  private var queueHead: Int = 0
+  private var queueTail: Int = 0
+
+  def init(): Unit = {
+    var i = 0
+    while (i < assembly.connectionCount) {
+      assembly.in(i).interpreter = this
+      assembly.in(i).id = i
```
patriknw: is it worth putting `assembly.in(i)` in a `val`?

patriknw:
looks clean and promising

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: yes, probably

drewhk:
This is ready for a review round. There will be more tests added and the async side-channel support is still missing, but the core functionality is there.

(@rkuhn take a look if you are interested)

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/ActorGraphInterpreterSpec.scala
```diff
@@ -0,0 +1,123 @@
+package akka.stream.impl.fusing
```
ktoso: Don't forget license headers later on when finished

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -0,0 +1,379 @@
...
+    private val IndexMask = size - 1
+
+    private def requestBatchSize = math.max(1, inputBuffer.length / 2)
+    private var batchRemaining = requestBatchSize
+
+    val out: Outlet[Any] = Outlet[Any]("UpstreamBoundary" + id)
+
+    private def dequeue(): Any = {
+      val elem = inputBuffer(nextInputElementCursor)
+      assert(elem ne null)
```
patriknw: we are sure that it can never ever be null?
otherwise require

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -0,0 +1,379 @@
...
+        inputBuffer((nextInputElementCursor + inputBufferElements) & IndexMask) = elem.asInstanceOf[AnyRef]
+        inputBufferElements += 1
+        if (isAvailable(out)) push(out, dequeue())
+      }
+    }
+
+    def onError(e: Throwable): Unit =
+      if (!upstreamCompleted) {
+        upstreamCompleted = true
+        fail(out, e)
```
patriknw: `clear` ?

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
rkuhn: I’d even say that we should flat-out ban `assert` from the code base outside of tests.

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
```diff
@@ -117,6 +117,19 @@ private[akka] case class ActorMaterializerImpl(
             assignPort(tls.plainIn, FanIn.SubInput[Any](impl, SslTlsCipherActor.UserIn))
             assignPort(tls.cipherIn, FanIn.SubInput[Any](impl, SslTlsCipherActor.TransportIn))
 
+          case graph: GraphModule ⇒
+            val props = ActorGraphInterpreter.props(graph.assembly, graph.shape, effectiveSettings(effectiveAttributes))
+            val impl = actorOf(props, stageName(effectiveAttributes), effectiveSettings(effectiveAttributes).dispatcher)
```
rkuhn: perhaps cache `effectiveSettings(effectiveAttributes)`?

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -0,0 +1,379 @@
...
+      interpreter.execute(eventLimit)
+      if (interpreter.isCompleted) context.stop(self)
+      else if (interpreter.isSuspended && !resumeScheduled) {
+        resumeScheduled = true
+        self ! Resume
+      }
+    } catch {
+      case NonFatal(e) ⇒
+        tryAbort(e)
+        context.stop(self)
```
patriknw: should the stop be placed before tryAbort, just to be sure that it is done

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
drewhk: copy pasta from OneBoundedInterpreter, this should be require(). The element cannot be null, it is enforced by the spec.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
drewhk: yes

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
drewhk: good idea

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
drewhk: yes, that makes sense

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/GraphInterpreterSpec.scala
```diff
@@ -0,0 +1,454 @@
+package akka.stream.impl.fusing
```
patriknw: copyright

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -0,0 +1,250 @@
...
+    for ((outlet, i) ← shape.outlets.iterator.zipWithIndex) {
+      assembly.outs(i + shape.inlets.size) = outlet
+      assembly.outOwners(i + shape.inlets.size) = 0
+    }
+
+    GraphModule(assembly, shape, Attributes.none)
+  }
+
+  // FIXME: How?
+  override def withAttributes(attr: Attributes): Graph[S, Unit] = ???
```
patriknw: should this be ???

patriknw:
LGTM

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: no, good that you mention. I have not yet figured this out.

patriknw:
yay

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  /**
+   * Return true if the given output port is ready to be pushed.
+   */
+  final def isAvailable[T](out: Outlet[T]): Boolean = interpreter.outAvailable(conn(out))
+
+  /**
+   * Obtain a callback object that can be used asynchronously to re-enter the
+   * current [[AsyncStage]] with an asynchronous notification. The [[invoke()]] method of the returned
```
rkuhn: `s/AsyncStage/GraphStage/`

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -0,0 +1,255 @@
...
+    GraphModule(assembly, shape, Attributes.none)
+  }
+
+  /**
+   * This method throws an [[UnsupportedOperationException]] by default. The subclass can override this method
+   * and provide a correct implementation that creates an exact copy of the stage with the provided new attributes.
+   */
+  override def withAttributes(attr: Attributes): Graph[S, Unit] =
+    throw new UnsupportedOperationException(
+      "withAttributes not supported by default by FlexiMerge, subclass may override and implement it")
```
rkuhn: `s/FlexiMerge/GraphStage/`

rkuhn: Couldn’t this be done by emitting a Graph that delegates to the same shape and a copied module?
~~~scala
override def withAttributes(attr: Attributes) =
  new Graph[S, Unit] {
    def shape = GraphStage.this.shape
    lazy val module = GraphStage.this.module.copy(attributes = attr)
  }
~~~

rkuhn: (or forcing the `module` here if the GC link is not desired)

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -0,0 +1,434 @@
...
+    def in: Inlet[T]
+  }
+
+  /**
+   * INTERNAL API
+   *
+   * A GraphAssembly represents a small stream processing graph to be executed by the interpreter. Instances of this
+   * class **must not** be mutated after construction.
+   *
+   * The arrays [[ins]] and [[outs]] correspond to the notion of a *connection* in the [[GraphInterpreter]]. Each slot
```
rkuhn: please remove links like these, they don’t work and fail the Java 8 javadoc

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -0,0 +1,434 @@
...
+   * *i* contains the input and output port corresponding to connection *i*. Slots where the graph is not closed (i.e.
+   * ports are exposed to the external world) are marked with *null* values. For example if an input port *p* is
+   * exposed, then outs(p) will contain a *null*.
+   *
+   * The arrays [[inOwners]] and [[outOwners]] are lookup tables from a connection id (the index of the slot)
+   * to a slot in the [[stages]] array, indicating which stage is the owner of the given input or output port.
+   * Slots which would correspond to non-existent stages (where the corresponding port is null since it represents
+   * the currently unknown external context) contain the value [[GraphInterpreter#Boundary]].
```
rkuhn: this link also won’t work

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -0,0 +1,434 @@
...
+   *
+   *            +---------------------------------------+-----------------+
+   * inOwners:  | index to stages array                 | Boundary (-1)   |
+   *            +----------------+----------------------+-----------------+
+   * ins:       | exposed inputs | internal connections | nulls           |
+   *            +----------------+----------------------+-----------------+
+   * outs:      | nulls          | internal connections | exposed outputs |
+   *            +----------------+----------------------+-----------------+
+   * outOwners: | Boundary (-1)  | index to stages array                  |
+   *            +----------------+----------------------------------------+
```
rkuhn: very good, thanks!

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -0,0 +1,434 @@
...
+        outOwners.mkString("[", ",", "]") +
+        ")"
+  }
+}
+
+/**
+ * INERNAL API
+ *
+ * From an external viewpoint, the GraphInterpreter takes an assembly of graph processing stages encoded as a
+ * [[GraphInterpreter#GraphAssembly]] object and provides facilities to execute and interact with this assembly.
```
rkuhn: this link is also broken

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -0,0 +1,434 @@
...
+ * Receiving an event is usually the following sequence:
+ *  - id of connection to be processed is dequeued
+ *  - the type of the event is determined by the object in the corresponding connectionStates slot
+ *  - the availability of the port is set on the receiver side to be true (inAvailable or outAvailable)
+ *  - using the inHandlers/outHandlers table the corresponding callback is called on the stage logic.
+ *
+ * Because of the FIFO construction of the queue the interpreter is fair, i.e. a pending event is always executed
+ * after a bounded number of other events. This property, together with suspendability means that even infinite cycles can
+ * be modeled, or even dissolved (if preempted and a "stealing" external even is injected; for example the non-cycle
+ * edge of a balance is pulled, dissolving the original cycle).
```
rkuhn: Nice!

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * INTERNAL API
+ */
+object GraphStages {
+
+  class Identity[T] extends GraphStage[FlowShape[T, T]] {
+    val in = Inlet[T]("in")
+    val out = Outlet[T]("out")
+
+    val shape = FlowShape(in, out)
```
2m: Here, and on other stages we should add `override` keyword.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -0,0 +1,434 @@
...
+          outAvailable(connection) = false
+          outHandlers(connection).onDownstreamFinish()
+          completeConnection(stageId)
+        }
+      case PushCompleted(elem) ⇒
+        inAvailable(connection) = true
+        connectionStates(connection) = elem
+        processElement(elem)
+        enqueue(connection, Completed)
+      case pushedElem ⇒ processElement(pushedElem)
```
rkuhn: null out element in connectionStates before processing

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -0,0 +1,390 @@
...
+  final case class ExposedPublisher(id: Int, publisher: ActorPublisher[Any]) extends BoundaryEvent
+
+  final case class AsyncInput(logic: GraphStageLogic, evt: Any, handler: (Any) ⇒ Unit) extends BoundaryEvent
+
+  case object Resume extends BoundaryEvent
+
+  final class BoundarySubscription(val parent: ActorRef, val id: Int) extends Subscription {
+    override def request(elements: Long): Unit = parent ! RequestMore(id, elements)
+    override def cancel(): Unit = parent ! Cancel(id)
+    override def toString = "BoundarySubscription" + System.identityHashCode(this)
```
rkuhn: no, please don’t use identity hash

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -0,0 +1,390 @@
...
+    val offset = assembly.connectionCount - outputs.length
+    i = 0
+    while (i < outputs.length) {
+      interpreter.attachDownstreamBoundary(i + offset, outputs(i))
+      i += 1
+    }
+    interpreter.init()
+  }
+
+  override def receive: Receive = {
```
rkuhn: please move OnNext and RequestMore to the beginning of the match, this is implemented by a linear scan

rkuhn:
Nice one, Endre! I have not scrutinized every single line, but this looks good, although it should be documented to be a mostly-internal power tool: end users will probably not have a clear enough mental picture to use push/pull without making mistakes—unless we add some really good documentation in this regard.

Ah, and the ActorGraphInterpreter probably wants some TCK love :-)

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: I was thinking about the delegate thing, but that would change the type of the class which can be confusing. Might be acceptable though. The current throwing approach is basically what FlexiMerge takes, and I just used it for now until we figure it out properly.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: ok

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: ok

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: nope, it will be changed to Empty as soon as grab() is called on that port. Until then, it needs to be kept there.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
drewhk: yeah, parent+id is enough here. This is copy pasta from FanIn I think.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
drewhk: Makes sense, and OnNext should be the first, as RequestMore is batched. I will move and comment that the order is important.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: The problem is that with the current layout subclasses cannot do anything besides what I described, which led me to comment.

drewhk:
This is not yet done, it was merged to
 - avoid growing this PR to infinity
 - enable me to work on much needed stages (e.g. idle timeout) right now

I will revisit the interpreter again, as it lacks tons of tests (completion is not tested at all, no TCK, etc.).

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
rkuhn: ah, right

rkuhn:
Yes, absolutely, I fully agree.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: ok, let's try the delegate approach and see how that goes.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
patriknw: @drewhk this is one of the special cases that doesn't work, nested object I guess

drewhk: But what about:

> please remove links like these, they don’t work and fail the Java 8 javadoc

Which comments on the links to ins and outs?

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
patriknw: @drewhk what are these actually linking to? of course they will not work if they are not linking to something public. These should probably be changed to simple backticks.

drewhk: Public fields of a case class.

patriknw: ``[[#ins]]`` ?

it could be a problem because it is nested in an object, but then it should be backticked

drewhk: I guess backtick will work for this case (these are internals anyway, we will mostly look at the comment, not the actual scaladoc/javadoc)

Issue #18314, opened 2015-08-25
Cluster sharding messages lost when region is terminated


patriknw:
Well, we don't guarantee that messages are not lost, but current implementation is not best effort.

The new `akka.cluster.down-removal-margin` comes into play.

During that period messages are still routed to the old location, even though we have got the Terminated message.

I think we can avoid the message loss by not replying to `GetShardHome` during the period.

PR #18317, opened 2015-08-25
=clt #18314 Reduce message lost when region terminates

patriknw:
The new `akka.cluster.down-removal-margin` comes into play.
During that period messages are still routed to the old location, even though we have got the Terminated message.

We can reduce (best effort) the message loss by not replying to GetShardHome during the period.

### DIFF
--- a/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
+++ b/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
@@ -382,6 +382,7 @@ abstract class ShardCoordinator(typeName: String, settings: ClusterShardingSetti
   var gracefulShutdownInProgress = Set.empty[ActorRef]
   var aliveRegions = Set.empty[ActorRef]
   var members = Set.empty[Address]
+  var regionTerminationInProgress = Set.empty[ActorRef]
 
   import context.dispatcher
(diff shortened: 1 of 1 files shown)


patriknw:
I think that failure was the same as was fixed in other PR. Will rebase.

ktoso:
Great improvement, LGTM

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -564,10 +570,10 @@ abstract class ShardCoordinator(typeName: String, settings: ClusterShardingSetti
       log.debug("ShardRegion terminated: [{}]", ref)
       state.regions(ref).foreach { s ⇒ self ! GetShardHome(s) }
 
-      gracefulShutdownInProgress -= ref
```
drewhk: what is the reason for this moved into the block below?

patriknw: There is no difference in practise, but it feels (reads) better to modify the state of the actor after persist of ShardRegionTerminated has been done. The reason it doesn't matter is that no other message is processed in-between (stashing) and if the persist fails the actor will be stopped anyway.

drewhk: Ok, thanks!

drewhk:
LGTM

patriknw:
Refs #18314

PR #18322, opened 2015-08-25

hveiga:


hveiga:
Please see https://github.com/akka/akka/pull/18323

PR #18323, opened 2015-08-26

hveiga:


akka-cluster/src/main/resources/reference.conf
```diff
@@ -34,7 +34,14 @@ akka {
...
     # must be stopped before corresponding persistent actors are started somewhere else.
     down-removal-margin = 20s
-    
+
+    # By default, the leader will not move 'Joining' members to 'Up' during a network
+    # split. This feature allows the leader to accept 'Joining' members to be 'WeaklyUp'
+    # so they become part of the cluster even during a network split. The leader will
+    # move 'WeaklyUp' members to 'Up' status once convergence has been reached. This
+    # feature must be off if some members are running Akka 2.3.X.
+    allow-weaklyup-members = off
```
patriknw: perhaps `allow-weakly-up-members` ?

patriknw:
Since we put it under a feature flag and disabled by default I'm actually considering including it in 2.4.0.
It doesn't look at all scary to me, but I might be biased.
Please take a look @ktoso @drewhk @2m

patriknw:
Documentation must be updated also.

hveiga:
I can easily change the flag name to `allow-weakly-up-members`. Is https://github.com/akka/akka/blob/master/akka-docs/rst/scala/cluster-usage.rst or https://github.com/akka/akka/blob/master/akka-docs/rst/common/cluster.rst the documentation file I should be updating? I can create a new PR with both changes.

patriknw:
Yes, that are the doc files + https://github.com/akka/akka/blob/master/akka-docs/rst/java/cluster-usage.rst

Please update that in this PR. You can just push more commits here and we will squash before merging.

hveiga:
Sounds good. I'll get it done as soon as I can.

akka-cluster/src/main/resources/reference.conf
ktoso: +1

hveiga:
This should be ready.

hveiga:
It's better to be picky :). I can add a new section with the diagram I put together for the initial ticket, however, the format is different to the current one so I will have to modify it to make it consistent. I will push these changes for you to re-re-review. I hope it's the last change :)

akka-cluster/src/main/resources/reference.conf
```diff
@@ -34,7 +34,14 @@ akka {
     # in case of a network partition the persistent actors in the non-surviving partitions
     # must be stopped before corresponding persistent actors are started somewhere else.
     down-removal-margin = 20s
-    
+
+    # By default, the leader will not move 'Joining' members to 'Up' during a network
+    # split. This feature allows the leader to accept 'Joining' members to be 'WeaklyUp'
+    # so they become part of the cluster even during a network split. The leader will
```
drewhk: This only explains the "upsides" of this feature and the weaklyup state. If I only read this I would be to tempted to turn this on. We should explain here in one simple sentence that this state has various limitations compared to be fully Up.

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -789,6 +789,9 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
         leaderActionCounter = 0
         leaderActionsOnConvergence()
       } else {
+        if (cluster.settings.AllowWeaklyUpMembers)
```
drewhk: this only works on the side where the leader is, right?

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -838,12 +847,8 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
     val localOverview = localGossip.overview
     val localSeen = localOverview.seen
 
-    def enoughMembers: Boolean = {
-      localMembers.size >= MinNrOfMembers && MinNrOfMembersOfRole.forall {
-        case (role, threshold) ⇒ localMembers.count(_.hasRole(role)) >= threshold
-      }
-    }
-    def isJoiningToUp(m: Member): Boolean = m.status == Joining && enoughMembers
+    val enoughMembers: Boolean = isMinNrOfMembersFulfilled
```
drewhk: enoughmembers is only used once in this cope as far as I see. If you inline it then it can save some evaluations since it is the second term of a && expression.

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
```diff
@@ -920,6 +925,33 @@ private[cluster] class ClusterCoreDaemon(publisher: ActorRef) extends Actor with
     }
   }
 
+  def moveJoiningToWeaklyUp(): Unit = {
+    val localGossip = latestGossip
+    val localMembers = localGossip.members
+
+    val enoughMembers: Boolean = isMinNrOfMembersFulfilled
```
drewhk: again, inlining this to the && expression where it used will save unnecessary evaluation (since this val is not used anywhere else)

akka-cluster/src/main/scala/akka/cluster/ClusterRemoteWatcher.scala
```diff
@@ -72,23 +73,28 @@ private[cluster] class ClusterRemoteWatcher(
...
-      if (m.address != selfAddress) {
-        clusterNodes -= m.address
-        if (previousStatus == MemberStatus.Down) {
-          quarantine(m.address, Some(m.uniqueAddress.uid))
-        }
-        publishAddressTerminated(m.address)
-      }
-    case _: MemberEvent ⇒ // not interesting
+    case MemberUp(m)                      ⇒ memberUp(m)
+    case MemberWeaklyUp(m)                ⇒ memberUp(m)
```
drewhk: is it safe here to treat it as a fully up member?

akka-cluster/src/main/scala/akka/cluster/routing/ClusterRouterConfig.scala
```diff
@@ -363,7 +363,7 @@ private[akka] trait ClusterRouterActor { this: RouterActor ⇒
   }
 
   def isAvailable(m: Member): Boolean =
-    m.status == MemberStatus.Up &&
+    (m.status == MemberStatus.Up || m.status == MemberStatus.WeaklyUp) &&
```
drewhk: is it OK that all routers automatically use WeaklyUp members? Should this be configurable per router, and if yes, opt-in or opt-out?

akka-docs/rst/java/cluster-usage.rst
```diff
@@ -167,6 +167,27 @@ leaving member will be shutdown after the leader has changed status of the membe
...
+
+If a node is ``unreachable`` then gossip convergence is not possible and therefore any
+``leader`` actions are also not possible. However, we still might want new nodes to join
+the cluster in this scenario.
+
+With a configuration option you can allow this behavior::
+
+    akka.cluster.allow-weakly-up-members = on
+
+When ``allow-weakly-up-members`` is enabled and there is no gossip convergence,
```
drewhk: This only explains why this is good. This tempts people to turn this on, without thinking about the consequences. They might also wonder why there is an Up state even when you can have this. Please explain the downside very carefully here.

drewhk:
If I understand correctly, this will allow adding this weak member nodes during partitions but *only on the side where the current leader is if there is any*. How useful is this in practice?

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
patriknw: No, it will be done on both sides of a network partition. The leader is the lowest address of the reachable nodes, i.e. there is always a leader, and there may be several. It is convergence that controls that one leader performs certain tasks, but this task is performed by all leaders.

drewhk: Ok, I see now.

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
patriknw: it's used from `def isJoiningToUp` and I think that is called in a loop

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
patriknw: it's called from collect below

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
drewhk: ok, missed that it is a def below...

akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala
drewhk: thanks

akka-cluster/src/main/scala/akka/cluster/ClusterRemoteWatcher.scala
patriknw: I think that is fine, otherwise it will be watched by the remote watcher (and may cause premature quarantine). Here it will not trigger Terminated until the member is removed, and that is still something that we only do on convergence.

akka-cluster/src/main/scala/akka/cluster/routing/ClusterRouterConfig.scala
patriknw: they already act on unreachable/reachable
we could of course make it a user decision

akka-cluster/src/main/resources/reference.conf
patriknw: good point

akka-cluster/src/main/scala/akka/cluster/routing/ClusterRouterConfig.scala
drewhk: Hm, thinking about it, I don't see much dangers with this feature, so we can leave this fixed here.

drewhk:
LGTM, only the docs need to be a bit more precise showing the tradeoff clearly.

patriknw:
@hveiga We have been discussing back and forth if we should include this in 2.4.0 or wait with it until 2.4.1. We think it is safe to add now, but we want to mark it as experimental so that we have possibility to change based on feedback.

I will cut the 2.4.0-RC2 release today, so I will create a new PR of this, with that adjustment (and perhaps add some more to the docs). It will still be your commit/contribution.

Sounds good to everyone?

patriknw:
Continued in https://github.com/akka/akka/pull/18398

patriknw:
Thanks a lot for contributing @hveiga !
There is a follow up ticket for Distributed Data in case you would like to continue, #18399

hveiga:
Thanks to you @patriknw for the support! I will take a look to that ticket and see if I can help.

hveiga:
Just curious, will this be part of 2.4.0-RC2 (and eventually 2.4.0) or 2.4.X? If it is going to be 2.4.0-RC2 you might want to mark the original ticket to that milestone.

patriknw:
Refs #13584

patriknw:
good catch, 2.4.0-RC2 it is

Issue #15646, opened 2014-08-11
Consider delaying initial GetShardHome replies until after watch


patriknw:
Problem illustrated by the attached program and logs in https://groups.google.com/d/msg/akka-user/Wauhjk_9PVg/Lk8X6BQY9wQJ

For this startup scenario we could improve the effort by giving the Terminated messages from the initial watch a chance to be processed before replying to GetShardHome requests.

Issue #17909, opened 2015-07-03
[akka-cluster] Getting `sending HandOverToMe to [None]` on newer versions (2.3.9 and 2.3.11) while it's Ok in older version (2.3.6)


AmirKarimi:
I'm using akka cluser sharding. When akka version is `2.3.9` or `2.3.11` I get `sending HandOverToMe to [None]` messages and the cluster is getting ready too late (35 seconds) while there is no `sending HandOverToMe to [None]` in version `2.3.6` and also cluster gets ready faster (3 seconds). 

I've build a sample to demonstrate it here: https://github.com/AmirKarimi/akka-cluster-shardin-bug-2.3.9

ktoso:
Thanks for reporting, we're a bit short on manpower over this month (holidays season), but will try to have a look at it as soon as we can.

patriknw:
`HandOverToMe to [None]` is not an error. There might have been some other bug in 2.3.6 that caused it to get ready faster. I don't see any issues when running your sample.

AmirKarimi:
@patriknw Please change the akka version to 2.3.11 and run the sample twice.

AmirKarimi:
@patriknw I got your point! You mean it's OK to get ready late?

patriknw:
I see. I think the bug we fixed after 2.3.6 was that we immediately triggered death watch on early connection failures instead of always using the heartbeat based failure detector. That is why it is so quick in 2.3.6 (but it is also not correct in all situations).

In this case the sharding coordinator thinks that the shard is located at old actor system instance with the 0 (random port) and therefore tries to route the messages there. It will do so until death watch is triggered by the failure detector.

You can see that it works better if you change to port 2553 instead of 0.

There is no guarantee in cluster sharding that messages will not be lost, so you have to use some kind of at-least-once mechanism if you can't allow for message loss.

I understand that this can be considered inconvenient as this startup scenario is rather common. I'm not sure we can find a way to improve this, but I will keep this issue open and think more about it.

patriknw:
Related to #15646

AmirKarimi:
@patriknw Thanks for explanation.

Another interesting situation is that, when you remove leveldb folders (with default persistence configuration) cluster gets ready fast again.

There are more information [here](https://groups.google.com/forum/#!searchin/akka-user/cluster$20ready/akka-user/axqoHHnqC1Q/uLaUTsjYlXYJ)

patriknw:
[...]

PR #18324, opened 2015-08-26
=cls #15646 Optimize the initial watch in shard coordinator

patriknw:
Two improvements to the coordinator startup (state recovery) that
should make it operational faster and reduce the amount of lost messages
during startup.
- Let the quick (those not involving failure detection) Terminated messages
  be processed before starting to reply to GetShardHome.
- Consider regions that don't belong to the current cluster
  to be terminated.

### DIFF
--- a/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
+++ b/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
@@ -3,21 +3,22 @@
  */
 package akka.cluster.sharding
 
-import akka.persistence._
-
 import scala.collection.immutable
 import scala.concurrent.Future
 import scala.concurrent.duration._
 import scala.util.Success
+
 import akka.actor._
+import akka.actor.DeadLetterSuppression
 import akka.cluster.Cluster
 import akka.cluster.ClusterEvent._
-import akka.cluster.ddata.LWWRegisterKey
+import akka.cluster.MemberStatus
 import akka.cluster.ddata.LWWRegister
+import akka.cluster.ddata.LWWRegisterKey
 import akka.cluster.ddata.Replicator._
 import akka.dispatch.ExecutionContexts
 import akka.pattern.pipe
-import akka.actor.DeadLetterSuppression
+import akka.persistence._
 
 /**
  * @see [[ClusterSharding$ ClusterSharding extension]]
@@ -239,6 +240,8 @@ object ShardCoordinator {
     @SerialVersionUID(1L) final case class ShardHomeAllocated(shard: ShardId, region: ActorRef) extends DomainEvent
     @SerialVersionUID(1L) final case class ShardHomeDeallocated(shard: ShardId) extends DomainEvent
 
+    case object StateInitialized
+
     object State {
       val empty = State()
     }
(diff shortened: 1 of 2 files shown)


patriknw:
It should improve things that users have complained about, such as #15646 and #17909, but I'm on the fence if this is to risky changes to include in RC2.

patriknw:
@smlin it would be great if you could review these changes, since you have this code swapped in recently

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -554,11 +553,45 @@ abstract class ShardCoordinator(typeName: String, settings: ClusterShardingSetti
...
 
   def update[E <: DomainEvent](evt: E)(f: E ⇒ Unit): Unit
 
+  def watchStateActors(): Unit = {
+
+    // Optimization:
+    // Consider regions that don't belong to the current cluster to be terminated.
+    // This is an optimization that makes it operational faster and reduces the
+    // amount of lost messages during startup.
+    val nodes = Cluster(context.system).state.members.map(_.address)
```
smlin: I think cluster state must always be taken from one place either from `MemberUp` or from `Cluster(context.system).state.members`. Otherwise this could lead to inconsistency.

patriknw: The reason for taking it from cluster.state here is that this is used from `RecoveryCompleted` which happens before any other external message is received, i.e. no MemberUp received yet.

I don't see that these members have to be be fully in sync with the ones you use for `Register`, but we could use the same for `Register` instead of subscribing to MemberUp

smlin:
@patriknw I think this is only an improvement and does not solve common problem. LGTM anyway.

This problem seems to be impossible to completely solve by such method. Since there no any guarantees that a region would not be removed by mistake/inconsistency. I think best method is lock all regions and shards inside the coordinator and unlock them when a region will confirm this data or terminated. So the coordinator could respond on `GetShardHome` for unlocked shards and for new shards when all data will be unlocked.

drewhk:
@patriknw Can you elaborate a bit more on this optimization? Where are those Terminated messages come from, what triggers them?

patriknw:
When the coordinator starts up it recovers its state and watch the registered region actor refs. When using persistence and starting up a new cluster the refs of the old cluster are all stale, but it takes long time (FD triggering) until that is detected. During that period messages are routed to those old region refs (which does not exist).

drewhk:
Ah, I see. And the Terminated will arrive quicker due to them being non-existent anymore

patriknw:
Yes, some will arrive quickly when using same hostname and ports. Others will not arrive until failure detector triggers.

drewhk:
thanks for the explanation

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
smlin: "we could use the same for Register instead of subscribing to MemberUp" - this exactly what I mean.

patriknw: sounds good, I'll change that

patriknw:
rebased and using cluster.state instead of subscribe

patriknw:
@ktoso do you have any opinions about this one, or can I merge it?

ktoso:
In the middle of reading it :)

ktoso:
Looks good, LGTM.

patriknw:
thanks for review!

patriknw:
Refs #15646

Issue #18156, opened 2015-08-10
failed: AdaptiveLoadBalancingRouterSpec


patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly2/78/consoleFull

ktoso:
Different instance of the failure in https://github.com/akka/akka/issues/18071

https://jenkins.akka.io:8498/job/akka-multi-node-nightly2/54/consoleFull

ktoso:
The `AdaptiveLoadBalancingRouter` fails with 

```

[JVM-1] [WARN] [07/15/2015 08:17:21.060] [AdaptiveLoadBalancingRouterSpec-akka.actor.default-dispatcher-17] [akka.tcp://AdaptiveLoadBalancingRouterSpec@a0.moxie:33794/system/cluster/core/daemon] Cluster Node [akka.tcp://AdaptiveLoadBalancingRouterSpec@a0.moxie:33794] - Marking node(s) as UNREACHABLE [Member(address = akka.tcp://AdaptiveLoadBalancingRouterSpec@a6.moxie:37022, status = Up)]
[JVM-3] [WARN] [07/15/2015 08:17:21.142] [AdaptiveLoadBalancingRouterSpec-akka.actor.default-dispatcher-14] [akka.tcp://AdaptiveLoadBalancingRouterSpec@a6.moxie:37022/system/cluster/core/daemon] Cluster Node [akka.tcp://AdaptiveLoadBalancingRouterSpec@a6.moxie:37022] - Marking node(s) as UNREACHABLE [Member(address = akka.tcp://AdaptiveLoadBalancingRouterSpec@a0.moxie:33794, status = Up)]
[JVM-1] [INFO] [07/15/2015 08:17:21.784] [AdaptiveLoadBalancingRouterSpec-akka.actor.default-dispatcher-17] [akka.cluster.Cluster(akka://AdaptiveLoadBalancingRouterSpec)] Cluster Node [akka.tcp://AdaptiveLoadBalancingRouterSpec@a0.moxie:33794] - Ignoring received gossip status from unreachable [UniqueAddress(akka.tcp://AdaptiveLoadBalancingRouterSpec@a6.moxie:37022,765315385)] 
[JVM-1] - must prefer node with more free heap capacity *** FAILED *** (21 seconds, 248 milliseconds)
[JVM-1]   2284 did not equal 3000 (AdaptiveLoadBalancingRouterSpec.scala:188)
[...]

PR #18325, opened 2015-08-26
=clu #18156 Harden AdaptiveLoadBalancingRouterSpec

patriknw:
Some logs shows some heavy GC that might be the reason for the unreachable. Increasing the failure detector timeout, since this tests is supposed to stress the memory.


patriknw:
Refs #18156

Issue #17555, opened 2015-05-22
nodes can talk after quarantine


patriknw:
Reproduced (does not always happen) with the remote-sample.

Start:

    sbt "run-main sample.remote.calculator.LookupApplication Calculator"
    sbt "run-main sample.remote.calculator.LookupApplication Lookup"

The log from the Lookup:

    [info] Running sample.remote.calculator.LookupApplication Lookup
    [INFO] [05/22/2015 09:41:16.648] [run-main-0] [Remoting] Starting remoting
    [INFO] [05/22/2015 09:41:16.845] [run-main-0] [Remoting] Remoting started; listening on addresses :[akka.tcp://LookupSystem@127.0.0.1:2553]
    [INFO] [05/22/2015 09:41:16.848] [run-main-0] [Remoting] Remoting now listens on addresses: [akka.tcp://LookupSystem@127.0.0.1:2553]
    Started LookupSystem
    Sub result: 77 - 89 = -12
    Sub result: 4 - 50 = -46
    Sub result: 15 - 88 = -73
    Sub result: 50 - 63 = -13
    Add result: 36 + 0 = 36
    Add result: 44 + 27 = 71
    Add result: 74 + 36 = 110
    Add result: 90 + 3 = 93
    Sub result: 41 - 7 = 34
    Add result: 28 + 45 = 73

    ### HERE I denied access to port 2552 with 
    (sudo pfctl -sr 2>/dev/null; echo "block drop quick on lo0 proto tcp from any to any port = 2552") | sudo pfctl -f - 2>/dev/null
    ###

    [WARN] [05/22/2015 09:41:38.878] [LookupSystem-akka.remote.default-remote-dispatcher-5] [akka.tcp://LookupSystem@127.0.0.1:2553/system/remote-watcher] Detected unreachable: [akka.tcp://CalculatorSystem@127.0.0.1:2552]
    [WARN] [05/22/2015 09:41:38.882] [LookupSystem-akka.remote.default-remote-dispatcher-6] [Remoting] Association to [akka.tcp://CalculatorSystem@127.0.0.1:2552] having UID [-1495856266] is irrecoverably failed. UID is now quarantined and all messages to this UID will be delivered to dead letters. Remote actorsystem must be restarted to recover from this situation.
    Calculator terminated
[...]

PR #18326, opened 2015-08-26
=rem #17555: Quarantine should clear pending connections

drewhk:
**DO NOT MERGE YET**

The message handler for the Quarantine command does shut down already registered writers and readers, but it did not clear the pending connections waiting for a handoff.

I have troubles reproducing the problem, so this is not 100% fix, but it looks like a bug anyway, even if it is not the one reported in the ticket

### DIFF
--- a/akka-remote/src/main/scala/akka/remote/Remoting.scala
+++ b/akka-remote/src/main/scala/akka/remote/Remoting.scala
@@ -549,6 +549,35 @@ private[remote] class EndpointManager(conf: Config, log: LoggingAdapter) extends
         case (Some((endpoint, currentUid)), Some(quarantineUid)) if currentUid == quarantineUid ⇒ context.stop(endpoint)
         case _ ⇒ // nothing to stop
       }
+
+      def matchesQuarantine(handle: AkkaProtocolHandle): Boolean = {
+        handle.remoteAddress == address &&
+          uidToQuarantineOption.forall(_ == handle.handshakeInfo.uid)
+      }
+
+      // Stop all matching pending read handoffs
+      pendingReadHandoffs = pendingReadHandoffs.filter {
+        case (pendingActor, pendingHandle) ⇒
+          val drop = matchesQuarantine(pendingHandle)
+          // Side-effecting here
+          if (drop) {
+            pendingHandle.disassociate()
+            context.stop(pendingActor)
+          }
+          !drop
+      }
+
+      // Stop all matching stashed connections
+      stashedInbound = stashedInbound.map {
+        case (writer, associations) ⇒
+          writer -> associations.filter { assoc ⇒
+            val handle = assoc.association.asInstanceOf[AkkaProtocolHandle]
+            val drop = matchesQuarantine(handle)
+            if (drop) handle.disassociate()
+            !drop
+          }
+      }
+
       uidToQuarantineOption foreach { uid ⇒
         endpoints.markAsQuarantined(address, uid, Deadline.now + settings.QuarantineDuration)
         eventPublisher.notifyListeners(QuarantinedEvent(address, uid))
--- a/akka-remote/src/test/scala/akka/remote/RemotingSpec.scala
+++ b/akka-remote/src/test/scala/akka/remote/RemotingSpec.scala
@@ -707,6 +707,87 @@ class RemotingSpec extends AkkaSpec(RemotingSpec.cfg) with ImplicitSender with D
 
     }
 
+    "should properly quarantine stashed inbound connections" in {
+      val localAddress = Address("akka.test", "system1", "localhost", 1)
+      val rawLocalAddress = localAddress.copy(protocol = "test")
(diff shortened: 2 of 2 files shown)


drewhk:
Refs #17555

akka-remote/src/main/scala/akka/remote/Remoting.scala
```diff
@@ -549,6 +549,20 @@ private[remote] class EndpointManager(conf: Config, log: LoggingAdapter) extends
...
+
+      // Stop all stashed pending connections
+      pendingReadHandoffs = pendingReadHandoffs.filterNot {
+        case (pendingActor, pendingHandle) ⇒
+          val addressMatches = pendingHandle.remoteAddress == address
+          // This yields true also if the Quarantine command did not contain a UID
+          val uidMatches = uidToQuarantineOption.forall(_ == pendingHandle.handshakeInfo.uid)
+
+          // Side-effecting here
+          context.stop(pendingActor)
```
patriknw: is it to correct to stop all pending actors, independent of the quarantine address and uid?

drewhk: meh, of course not!

drewhk:
Fixed the previous issue, and also added proper Quarantined information to be propagated to the remote connection if applicable.

akka-remote/src/main/scala/akka/remote/Remoting.scala
```diff
@@ -549,6 +549,25 @@ private[remote] class EndpointManager(conf: Config, log: LoggingAdapter) extends
...
+          val keep = addressMatches & uidMatches
+
+          // Side-effecting here
+          if (!keep) {
+            if (uidToQuarantineOption.nonEmpty)
+              pendingHandle.disassociate(AssociationHandle.Quarantined)
+            else pendingHandle.disassociate()
+            context.stop(pendingActor)
+          }
+          keep
```
patriknw: now we keep the ones that match the quarantine and throw away the others, is that intended?

drewhk: No, I messed it up again.

drewhk:
More fixes, and now with a test-case (not very straightforward though)

akka-remote/src/test/scala/akka/remote/RemotingSpec.scala
```diff
@@ -707,6 +707,87 @@ class RemotingSpec extends AkkaSpec(RemotingSpec.cfg) with ImplicitSender with D
...
+
+        // Quarantine the connection
+        RARP(thisSystem).provider.quarantine(remoteAddress, Some(remoteUID))
+
+        // Even though the connection is stashed it will be disassociated
+        inboundHandleProbe.expectMsgType[AssociationHandle.Disassociated]
+
+      } finally shutdown(thisSystem)
+
+    }
```
patriknw: wow, that's indeed not a trivial test

patriknw:
LGTM

drewhk:
Squashed

akka-remote/src/main/scala/akka/remote/Remoting.scala
```diff
@@ -549,6 +549,35 @@ private[remote] class EndpointManager(conf: Config, log: LoggingAdapter) extends
         case (Some((endpoint, currentUid)), Some(quarantineUid)) if currentUid == quarantineUid ⇒ context.stop(endpoint)
         case _ ⇒ // nothing to stop
       }
+
+      def matchesQuarantine(handle: AkkaProtocolHandle): Boolean = {
+        handle.remoteAddress == address &&
+          uidToQuarantineOption.forall(_ == handle.handshakeInfo.uid)
```
viktorklang: How about `contains` iso forall?

akka-remote/src/main/scala/akka/remote/Remoting.scala
```diff
@@ -549,6 +549,35 @@ private[remote] class EndpointManager(conf: Config, log: LoggingAdapter) extends
...
+          keep
+      }
+
+      // Stop all matching stashed connections
+      stashedInbound = stashedInbound.map {
+        case (writer, associations) ⇒
+          writer -> associations.filter { assoc ⇒
+            val handle = assoc.association.asInstanceOf[AkkaProtocolHandle]
+            val keep = !matchesQuarantine(handle)
+            if (!keep) handle.disassociate()
```
viktorklang: Double negation?

drewhk: I can change it to drop, then only a single negation is needed

akka-remote/src/main/scala/akka/remote/Remoting.scala
drewhk: contains does not return true for the empty Option, while forall does.

patriknw:
LGTM, please forward port to master

drewhk:
Thanks!

Issue #18328, opened 2015-08-26
Optimize ORSet.merge


patriknw:
I noticed that the extra merge introduced by #18106 makes the local write very slow compared to not merging. Especially we should see if we can detect and optimize the "fast forward" merge path.

Impact illustrated by the PerformanceSpec.

With merge:

    [JVM-1] ## 2000 ORSet Update WriteLocal took 3534 ms, 565 TPS
    [JVM-1] ## 2000 ORSet Update WriteLocal took 3637 ms, 549 TPS
    [JVM-1] ## 2000 ORSet Update WriteLocal took 3464 ms, 577 TPS

Without the extra merge:

    [JVM-1] ## 2000 ORSet Update WriteLocal took 15 ms, 133333 TPS
    [JVM-1] ## 2000 ORSet Update WriteLocal took 12 ms, 166666 TPS
    [JVM-1] ## 2000 ORSet Update WriteLocal took 16 ms, 125000 TPS

Issue #18235, opened 2015-08-17
FAILED: akka.cluster.ddata.PerformanceSpec


ktoso:
Just a bit fleaky IMO: https://jenkins.akka.io:8498/job/akka-multi-node-nightly2/86/consoleFull

```
[info] * akka.cluster.ddata.PerformanceSpec
[JVM-1] PerformanceSpecMultiJvmNode1:
[JVM-1] Performance
[JVM-2] PerformanceSpecMultiJvmNode2:
[JVM-2] Performance
[JVM-5] PerformanceSpecMultiJvmNode5:
[JVM-3] PerformanceSpecMultiJvmNode3:
[JVM-5] Performance
[JVM-3] Performance
[JVM-4] PerformanceSpecMultiJvmNode4:
[JVM-4] Performance
[JVM-2] - must setup cluster (2 seconds, 634 milliseconds)
[JVM-5] - must setup cluster (2 seconds, 626 milliseconds)
[JVM-1] - must setup cluster (2 seconds, 727 milliseconds)
[JVM-3] - must setup cluster (2 seconds, 624 milliseconds)
[JVM-4] - must setup cluster (2 seconds, 619 milliseconds)
[JVM-1] - must be great for ORSet Update WriteLocal *** FAILED *** (6 seconds, 216 milliseconds)
[JVM-1]   java.util.concurrent.TimeoutException: Timeout of 6 seconds with time factor of 1.0
[JVM-1]   at akka.testkit.TestLatch.ready(TestLatch.scala:41)
[JVM-1]   at akka.testkit.TestLatch.ready(TestLatch.scala:26)
[JVM-1]   at scala.concurrent.Await$$anonfun$ready$1.apply(package.scala:169)
[JVM-1]   at scala.concurrent.Await$$anonfun$ready$1.apply(package.scala:169)
[JVM-1]   at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:53)
[JVM-1]   at scala.concurrent.Await$.ready(package.scala:169)
[JVM-1]   at akka.cluster.ddata.PerformanceSpec$$anonfun$repeat$1$$anonfun$apply$1.apply$mcV$sp(PerformanceSpec.scala:91)
[JVM-1]   at akka.remote.testkit.MultiNodeSpec.runOn(MultiNodeSpec.scala:338)
[...]

PR #18329, opened 2015-08-26
=cdd #18235 Harden ddata PerformanceSpec

patriknw:
I noticed that the extra merge introduced by #18328 makes the local write very slow compared to not merging. That might be the reason why this test started failing. I will increase the timeout in this issue and we can do performance optimizations in #18328


patriknw:
that failure in unreleated test was caused by a long GC

patriknw:
PLS BUILD

drewhk:
LGTM

patriknw:
Refs #18235

PR #18282, opened 2015-08-20
akka-http-testkit websocket support (#17914) + WS subprotocols in WebsocketDirectives (#17726)

jrudolph:
Will fix #17726 and #17914.

/cc @sirthias

PR #18330, opened 2015-08-26
+htp #16438 add Java-side rejection handling support

### LINKED ISSUE #16438: Java routing: provide API for rejections
Currently, there's only `RequestContext.notFound()`. Full-fledged rejection support would be nice.

jrudolph:
The idea is a listener-based approach where each pre-defined Scala rejection corresponds to one callback method in a `RejectionHandler`. An application can decide which rejections to handle by just overriding handler methods that are interesting to the application. This allows us to implement rich custom rejection handling without the need to reproduce the complete rejection class hierarchy to the Java side. However, it is currently not possible to create one of pre-defined rejections. It is possible to create custom rejections on the Java side by extending `CustomRejection`. This is a trade-off that should suffice in most situations.

Fixes #16438.

This is currently based on #18282 and needs to be rebased once that one has been merged.

/cc @sirthias

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/ContentTypeRange.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/ContentTypeRange.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.javadsl.model;
+
+/**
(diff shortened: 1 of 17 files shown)


ktoso:
LGTM, only one thing to rename perhaps. Thanks and sorry this stayed open so long @jrudolph!

jrudolph:
Rebased and applied the suggested name change.

drewhk:
merged

jrudolph:
Thanks!

Issue #17253, opened 2015-04-22
deadlock when initializing Akka cluster: document minimum thread pool size


clockfly:
### Frequency: random 

```
java.util.concurrent.TimeoutException: Futures timed out after [60000 milliseconds]
        at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:219)
        at scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:223)
        at scala.concurrent.Await$$anonfun$result$1.apply(package.scala:116)
        at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:53)
        at scala.concurrent.Await$.result(package.scala:116)
        at akka.cluster.Cluster.liftedTree1$1(Cluster.scala:172)
        at akka.cluster.Cluster.<init>(Cluster.scala:171)
        at akka.cluster.Cluster$.createExtension(Cluster.scala:42)
        at akka.cluster.Cluster$.createExtension(Cluster.scala:37)
        at akka.actor.ActorSystemImpl.registerExtension(ActorSystem.scala:711)
        at akka.actor.ExtensionId$class.apply(Extension.scala:79)
        at akka.cluster.Cluster$.apply(Cluster.scala:37)
        at 
     ......
akka.cluster.ClusterActorRefProvider.createRemoteWatcher(ClusterActorRefProvider.scala:66)
        at akka.remote.RemoteActorRefProvider.init(RemoteActorRefProvider.scala:186)
        at akka.cluster.ClusterActorRefProvider.init(ClusterActorRefProvider.scala:58)
        at akka.actor.ActorSystemImpl.liftedTree2$1(ActorSystem.scala:618)
        at akka.actor.ActorSystemImpl._start$lzycompute(ActorSystem.scala:615)
        at akka.actor.ActorSystemImpl._start(ActorSystem.scala:615)
        at akka.actor.ActorSystemImpl.start(ActorSystem.scala:632)
        at akka.actor.ActorSystem$.apply(ActorSystem.scala:141)
        at akka.actor.ActorSystem$.apply(ActorSystem.scala:118)
```

### Investigation
The timeout happen in class Cluster, when trying to GetClusterCoreRef

```
  private[cluster] val clusterCore: ActorRef = {
    implicit val timeout = system.settings.CreationTimeout
    try {
      Await.result((clusterDaemons ? InternalClusterAction.GetClusterCoreRef).mapTo[ActorRef], timeout.duration)
    } catch {
[...]

PR #18332, opened 2015-08-27
=clu #17253 Improve cluster startup thread usage

patriknw:
When using a dispatcher (default or separate cluster dispatcher)
with less than 5 threads the Cluster extension initialization
could deadlock.

It was reproducable by adding a sleep before the Await of GetClusterCoreRef
in the Cluster extension constructor. The reason was that other cluster actors were
started too early and they also tried to get the Cluster extension and thereby blocking
dispatcher threads.

Note that the Cluster extension is started via ClusterActorRefProvider before
ActorSystem.apply returns.

The improvement is to start the cluster child actors lazily when the
GetClusterCoreRef is received.


patriknw:
Refs #17253

patriknw:
ah, required some mima filters also for the internal changes

drewhk:
LGTM

patriknw:
the failure was unrelated (but new) issue https://github.com/akka/akka/issues/18336

patriknw:
PLS BUILD

Issue #18337, opened 2015-08-27
disable down-removal-margin by default


patriknw:
For manual downing it is not needed. For auto-down it doesn't add any extra safety, since that is not handling network partitions anyway.

The setting is still useful if you implement downing strategies that handle network partitions, e.g. by keeping the larger side of the partition and shutting down the smaller side.

PR #18338, opened 2015-08-27
=clu #18337 Disable down-removal-margin by default

patriknw:
For manual downing it is not needed. For auto-down it doesn't add any extra safety, since that
is not handling network partitions anyway.

The setting is still useful if you implement downing strategies that handle network partitions,
e.g. by keeping the larger side of the partition and shutting down the smaller side.

### DIFF
--- a/akka-cluster/src/main/resources/reference.conf
+++ b/akka-cluster/src/main/resources/reference.conf
@@ -29,11 +29,16 @@ akka {
     # Disable with "off" or specify a duration to enable auto-down.
     auto-down-unreachable-after = off
     
-    # Margin until shards or singletons that belonged to a downed/removed
+    # Time margin after which shards or singletons that belonged to a downed/removed
     # partition are created in surviving partition. The purpose of this margin is that 
     # in case of a network partition the persistent actors in the non-surviving partitions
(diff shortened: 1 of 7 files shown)


patriknw:
Refs #18337

patriknw:
failure was unrelated (but new) issue: https://github.com/akka/akka/issues/18339

patriknw:
PLS BUILD

akka-cluster/src/main/resources/reference.conf
```diff
@@ -33,7 +33,12 @@ akka {
     # partition are created in surviving partition. The purpose of this margin is that
```
drewhk: in the above sentence (L32) it says "until" but is that correct? Isn't it "after"?

drewhk:
I agree, this can be safely disabled to be the default.

drewhk:
LGTM

akka-cluster/src/main/resources/reference.conf
patriknw: you are right, I'll correct that

ktoso:
LGTM

Issue #18296, opened 2015-08-22
Persistence - AsyncRecovery readHighestSequenceNr clarification


chbatey:
At the moment the asyncReadHighestSequenceNr fromSequenceNr parameter is described as a hint where to start searching from, it would be nice to specify exactly what will be passed in in the presence/absence of snapshots.

I could be wrong but for akka 2.4-M2 for a persistenceId with no snapshots 0 was passed in, as of 2.4-RC1 1 is passed in. The cassandra journal previously returned the fromSequenceNr if it found no events either because it is a new persistenceId or all messages have been deleted/snapshotted which led to new persistenceIds starting at sequence number 2 rather than 1. 

WDYT?

patriknw:
This might have changed, yes. Now the fromSequenceNr param is 1 or snapshotSeqNr + 1. That makes sense for the asyncReplayMessages but I agree that readHighestSequenceNr should have one less.

patriknw:
Thanks a lot for reporting, by the way :-)

PR #18355, opened 2015-08-28
=per #18296 Use 0L or snapshot seqNr as asyncReadHighestSequenceNr param

patriknw:
For example, a new persistent actor (no snapshots, no events) should use
0L so that it makes sense that journal returns 0L and the first persisted
event gets 1L.


ktoso:
LGTM

Issue #18356, opened 2015-08-28
Metrics based resizer for routers


kailuowang:
Based on discussion with @ktoso and @richdougherty, implement a Resizer that can auto scale the routee pool based on performance metrics.  

Basically it should perform three resizing operations periodically:
1. downsize after it hasn't seen all routees fully utilized for a period of time.
2. explore to nearby pool sizes to collect metrics 
3. optimize to a nearby pool size that has better performance metrics

When the pool is full, it randomly choose between exploring and optimizing.

patriknw:
Shouldn't this be closed?

ktoso:
Right, thanks @patriknw

PR #18357, opened 2015-08-28
+act #18356 Metrics based resizer for router

kailuowang:
UPDATE: a new commit brings back the binary compatibility, so it's no longer a breaking change. 

The change to the API is an aditional perameter added to the resize method of akka.routing.Resizer
So for users who implemented their own Resizer, they will have to make an easy change to their implemenation.
However it's not obvious from the akka documentation that user should/can implement their own Resizer.

This commit doesn't include change to akka documentation. I decided to hold on that before I get review feedbacks.

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/routing/MetricsBasedResizerSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/routing/MetricsBasedResizerSpec.scala
@@ -1,0 +1,345 @@
+/**
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.routing
+
+import java.time.LocalDateTime
+
+import akka.actor._
+import akka.testkit._
+import akka.testkit.TestEvent._
+
+import OptimalSizeExploringResizer._
+import MetricsBasedResizerSpec._
+import akka.util.Timeout
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import scala.util.{ Try, Random }
+import akka.pattern.ask
+
+object MetricsBasedResizerSpec {
+
+  class TestLatchingActor(implicit timeout: Timeout) extends Actor {
+    import context.system
+
+    def receive = {
+      case latch: TestLatch ⇒
+        Try(Await.ready(latch, timeout.duration))
+    }
+  }
+
+  def routee(implicit system: ActorSystem, timeout: Timeout): ActorRefRoutee =
+    ActorRefRoutee(system.actorOf(Props(new TestLatchingActor)))
+
+  def routees(num: Int = 10)(implicit system: ActorSystem, timeout: Timeout) = (1 to num).map(_ ⇒ routee).toVector
+
+  case class TestRouter(routees: Vector[ActorRefRoutee], resizer: Resizer)(implicit system: ActorSystem, timeout: Timeout) {
+
+    system.registerOnTermination(close())
+
+    var msgs: Set[TestLatch] = Set()
+
+    def mockSend(l: TestLatch = TestLatch(),
+                 routeeIdx: Int = Random.nextInt(routees.length),
+                 wait: Boolean = true)(implicit sender: ActorRef): TestLatch = {
+      val target = routees(routeeIdx)
+      target.send(l, sender)
+      msgs = msgs + l
+      if (wait) waitForMessageToArrive()
+      l
+    }
+
+    def waitForMessageToArrive(): Unit = Thread.sleep(1.milliseconds.dilated.toMillis)
+
+    def close(): Unit = msgs.foreach(_.open())
+
+    def sendToAll()(implicit sender: ActorRef): Seq[TestLatch] = {
+      val sentMessages = (0 until routees.length).map(i ⇒ mockSend(routeeIdx = i, wait = false))
+      waitForMessageToArrive()
+      sentMessages
+    }
+
+  }
+
+}
+
+@org.junit.runner.RunWith(classOf[org.scalatest.junit.JUnitRunner])
+class MetricsBasedResizerSpec extends AkkaSpec(ResizerSpec.config) with DefaultTimeout with ImplicitSender {
+
+  override def atStartup: Unit = {
+    // when shutting down some Resize messages might hang around
+    system.eventStream.publish(Mute(EventFilter.warning(pattern = ".*Resize")))
+  }
+
+  "MetricsBasedResizer isTimeForResize" must {
+
+    "be true with empty history" in {
+      val resizer = DefaultOptimalSizeExploringResizer()
+      resizer.record = ResizeRecord(checkTime = 0)
+      resizer.isTimeForResize(0) should ===(true)
+    }
+
+    "be false if the last resize is too close within actionInterval enough history" in {
+      val resizer = DefaultOptimalSizeExploringResizer(actionInterval = 10.seconds)
+      resizer.record = ResizeRecord(checkTime = System.nanoTime() - 8.seconds.toNanos)
+
+      resizer.isTimeForResize(100) should ===(false)
+    }
+
+    "be true if the last resize is before actionInterval ago" in {
+      val resizer = DefaultOptimalSizeExploringResizer(actionInterval = 10.seconds)
+      resizer.record = ResizeRecord(checkTime = System.nanoTime() - 11.seconds.toNanos)
+
+      resizer.isTimeForResize(100) should ===(true)
+    }
+
+  }
+
+  "MetricsBasedResizer reportMessageCount" must {
+
+    "record last messageCounter correctly" in {
+      val resizer = DefaultOptimalSizeExploringResizer()
+      resizer.reportMessageCount(Vector(routee), 3)
+      resizer.record.messageCount shouldBe 3
+    }
+
+    "record last totalQueueLength correctly" in {
+      val resizer = DefaultOptimalSizeExploringResizer()
+      val router = TestRouter(routees(2), resizer)
+
+      resizer.reportMessageCount(router.routees, router.msgs.size)
+
+      resizer.record.totalQueueLength shouldBe 0
+
+      router.mockSend()
+      router.mockSend()
+
+      resizer.reportMessageCount(router.routees, router.msgs.size)
+      resizer.record.totalQueueLength shouldBe 2
+
+    }
+
+    "start an underutilizationStreak when not fully utilized" in {
+      val resizer = DefaultOptimalSizeExploringResizer()
+      resizer.reportMessageCount(routees(2), 0)
+      resizer.record.underutilizationStreak should not be empty
+      resizer.record.underutilizationStreak.get.start.isBefore(LocalDateTime.now.plusSeconds(1)) shouldBe true
+      resizer.record.underutilizationStreak.get.start.isAfter(LocalDateTime.now.minusSeconds(1)) shouldBe true
+    }
+
+    "stop an underutilizationStreak when fully utilized" in {
+      val resizer = DefaultOptimalSizeExploringResizer()
+      resizer.record = ResizeRecord(
+        underutilizationStreak = Some(UnderUtilizationStreak(start = LocalDateTime.now.minusHours(1), highestUtilization = 1)))
+
+      val router = TestRouter(routees(2), resizer)
+      router.sendToAll()
+
+      resizer.reportMessageCount(router.routees, router.msgs.size)
+      resizer.record.underutilizationStreak shouldBe empty
+    }
+
+    "leave the underutilizationStreak start date unchanged when not fully utilized" in {
+      val start: LocalDateTime = LocalDateTime.now.minusHours(1)
+      val resizer = DefaultOptimalSizeExploringResizer()
+      resizer.record = ResizeRecord(
+        underutilizationStreak = Some(UnderUtilizationStreak(start = start, highestUtilization = 1)))
+
+      resizer.reportMessageCount(routees(2), 0)
+      resizer.record.underutilizationStreak.get.start shouldBe start
+    }
+
+    "leave the underutilizationStreak highestUtilization unchanged if current utilization is lower" in {
+      val resizer = DefaultOptimalSizeExploringResizer()
+      resizer.record = ResizeRecord(
+        underutilizationStreak = Some(UnderUtilizationStreak(start = LocalDateTime.now, highestUtilization = 2)))
+
+      val router = TestRouter(routees(2), resizer)
+      router.mockSend()
+
+      resizer.reportMessageCount(router.routees, router.msgs.size)
+      resizer.record.underutilizationStreak.get.highestUtilization shouldBe 2
+
+    }
+
+    "update the underutilizationStreak highestUtilization if current utilization is higher" in {
+      val resizer = DefaultOptimalSizeExploringResizer()
+      resizer.record = ResizeRecord(
+        underutilizationStreak = Some(UnderUtilizationStreak(start = LocalDateTime.now, highestUtilization = 1)))
+
+      val router = TestRouter(routees(3), resizer)
+      router.mockSend(routeeIdx = 0)
+      router.mockSend(routeeIdx = 1)
+
+      resizer.reportMessageCount(router.routees, router.msgs.size)
+      resizer.record.underutilizationStreak.get.highestUtilization shouldBe 2
+
+    }
+
+    "not record a performance log when it's not fully utilized in two consecutive checks" in {
+      val resizer = DefaultOptimalSizeExploringResizer()
+      val router = TestRouter(routees(2), resizer)
+      resizer.reportMessageCount(router.routees, router.msgs.size)
+
+      router.sendToAll()
+      resizer.reportMessageCount(router.routees, router.msgs.size)
+
+      resizer.performanceLog shouldBe empty
+    }
+
+    "not record the performance log when no message is processed" in {
+      val resizer = DefaultOptimalSizeExploringResizer()
+      resizer.record = ResizeRecord(
+        totalQueueLength = 2,
+        messageCount = 2,
+        checkTime = System.nanoTime())
+
+      val router = TestRouter(routees(2), resizer)
+
+      router.sendToAll()
+
+      resizer.reportMessageCount(router.routees, router.msgs.size)
+
+      resizer.performanceLog shouldBe empty
+    }
+
+    "record the performance log with the correct pool size" in {
+      val resizer = DefaultOptimalSizeExploringResizer()
+      val router = TestRouter(routees(2), resizer)
+      val msgs = router.sendToAll()
+      resizer.reportMessageCount(router.routees, router.msgs.size)
+      msgs.head.open()
+
+      router.sendToAll()
+      resizer.reportMessageCount(router.routees, router.msgs.size)
+      resizer.performanceLog.get(2) should not be empty
+    }
+
+    "record the performance log with the correct process speed" in {
+      val resizer = DefaultOptimalSizeExploringResizer()
+      val router = TestRouter(routees(2), resizer)
+      val msgs = router.sendToAll()
+      router.sendToAll() //make sure the routees are still busy after the first batch of messages get processed.
+
+      val before = LocalDateTime.now
+      resizer.reportMessageCount(router.routees, router.msgs.size) //updates the records
+
+      msgs.foreach(_.open()) //process two messages
+
+      Thread.sleep(1) // wait for routees to update their mail boxes
+
+      resizer.reportMessageCount(router.routees, router.msgs.size)
+
+      val after = LocalDateTime.now
+      resizer.performanceLog(2).toMillis shouldBe (java.time.Duration.between(before, after).toMillis / 2 +- 1)
+    }
+
+    "update the old performance log entry with updated speed " in {
+      val oldSpeed = 50
+      val resizer = DefaultOptimalSizeExploringResizer(
+        weightOfLatestMetric = 0.5)
+
+      resizer.performanceLog = Map(2 → oldSpeed.milliseconds)
+
+      val router = TestRouter(routees(2), resizer)
+      val msgs = router.sendToAll()
+
+      router.sendToAll() //make sure the routees are still busy after the first batch of messages get processed.
+
+      val before = LocalDateTime.now
+      resizer.reportMessageCount(router.routees, router.msgs.size) //updates the records
+
+      msgs.foreach(_.open()) //process two messages
+
+      Thread.sleep(1) // wait for routees to update their mail boxes
+
+      resizer.reportMessageCount(router.routees, router.msgs.size)
+
+      val after = LocalDateTime.now
+      val newSpeed = java.time.Duration.between(before, after).toMillis / 2
+
+      resizer.performanceLog(2).toMillis shouldBe ((newSpeed + oldSpeed) / 2 +- 1)
+    }
+
+  }
+
+  "MetricsBasedResizer resize" must {
+    "downsize to close to the highest retention when a streak of underutilization started downsizeAfterUnderutilizedFor" in {
+      val resizer = DefaultOptimalSizeExploringResizer(
+        downsizeAfterUnderutilizedFor = 72.hours,
+        downsizeRatio = 0.5)
+
+      resizer.record = ResizeRecord(underutilizationStreak = Some(
+        UnderUtilizationStreak(start = LocalDateTime.now.minusHours(73), highestUtilization = 8)))
+      resizer.resize(routees(20)) should be(4 - 20)
+    }
+
+    "does not downsize on empty history" in {
+      val resizer = DefaultOptimalSizeExploringResizer()
+      resizer.resize(routees()) should be(0)
+    }
+
+    "always go to lowerBound if below it" in {
+      val resizer = DefaultOptimalSizeExploringResizer(lowerBound = 50, upperBound = 100)
+      resizer.resize(routees(20)) should be(30)
+    }
+
+    "always go to uppperBound if above it" in {
+      val resizer = DefaultOptimalSizeExploringResizer(upperBound = 50)
+      resizer.resize(routees(80)) should be(-30)
+    }
+
+    "explore when there is performance log but not go beyond exploreStepSize" in {
+      val resizer = DefaultOptimalSizeExploringResizer(
+        exploreStepSize = 0.3,
+        explorationProbability = 1)
+      resizer.performanceLog = Map(11 → 1.milli, 13 → 1.millis, 12 → 3.millis)
+
+      val exploreSamples = (1 to 100).map(_ ⇒ resizer.resize(routees(10)))
+      exploreSamples.forall(change ⇒ Math.abs(change) >= 1 && Math.abs(change) <= (10 * 0.3)) should be(true)
+
+    }
+  }
+
+  "MetricsBasedResizer optimize" must {
+    "optimize towards the fastest pool size" in {
+      val resizer = DefaultOptimalSizeExploringResizer(explorationProbability = 0)
+      resizer.performanceLog = Map(7 → 5.millis, 10 → 3.millis, 11 → 2.millis, 12 → 4.millis)
+      resizer.resize(routees(10)) should be(1)
+      resizer.resize(routees(12)) should be(-1)
+      resizer.resize(routees(7)) should be(2)
+    }
+
+    "ignore further away sample data when optmizing" in {
+      val resizer = DefaultOptimalSizeExploringResizer(explorationProbability = 0, numOfAdjacentSizesToConsiderDuringOptimization = 4)
+      resizer.performanceLog = Map(
+        7 → 5.millis,
+        8 → 2.millis,
+        10 → 3.millis,
+        11 → 4.millis,
+        12 → 3.millis,
+        13 → 1.millis)
+
+      resizer.resize(routees(10)) should be(-1)
+    }
+  }
+
+  "MetricsBasedResizer" must {
+
+    def poolSize(router: ActorRef): Int =
+      Await.result(router ? GetRoutees, timeout.duration).asInstanceOf[Routees].routees.size
+
+    "start with lowerbound pool size" in {
+
+      val resizer = DefaultOptimalSizeExploringResizer(lowerBound = 2)
+      val router = system.actorOf(RoundRobinPool(nrOfInstances = 0, resizer = Some(resizer)).props(Props(new TestLatchingActor)))
+      Thread.sleep(10)
+
+      poolSize(router) shouldBe resizer.lowerBound
+
+    }
+
+  }
+
+}
--- a/akka-actor/src/main/scala/akka/routing/Resizer.scala
+++ b/akka-actor/src/main/scala/akka/routing/Resizer.scala
@@ -7,6 +7,9 @@ import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.atomic.AtomicLong
 
+import akka.AkkaException
+import akka.event.Logging.Error.NoCause
+
 import scala.collection.immutable
 import scala.concurrent.duration._
 import scala.concurrent.duration.FiniteDuration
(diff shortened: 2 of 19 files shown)


ktoso:
OK TO TEST

ktoso:
Thanks a lot for the PR, will review it soon! Have a great weekend in the mean time :smile:

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
```diff
@@ -0,0 +1,226 @@
...
+          optimize(currentSize)
+      }
+
+    if (proposedChange + currentSize > upperBound) {
+      upperBound - currentSize
+    } else if (proposedChange + currentSize < lowerBound) {
+      lowerBound - currentSize
+    } else
+      proposedChange
+  }
```
richdougherty: An equivalent way to write this:
```scala
Math.max(lowerBound, Math.min(proposedChange, upperBound))
```

richdougherty:
Thanks @kailuowang, that looks awesome! Does it work well in practice?

I think all the logic looks good to me—although I won't promise I managed to review it all, I mainly read the comments!—but I have some questions about the downsize logic. If I understand correctly, if the pool fails to reach full capacity for 72 hours then it will downsize to a size based on its max utilisation over that period. The size is configurable, but would be (max size x 1.2) by default. Is that right?

If so, it might be better to resize to something like (max size x 0.8) by default. The reason is that we want to max the pool out occasionally so that the resizer can start exploring/optimising. If we downsize to a value that's above the max pool size then it might be hard to max out the pool, so we may never get a chance to explore/optimise. That's why I suggest downsizing to a value below the optimum.

I think 72h is a good period to discover the max utilisation, but it might take a while to downsize the pool when the server first starts up. I wonder if it would be possible to have a smaller time interval when the server first starts up? Changing time intervals is probably getting a bit complicated though, so I wouldn't worry too much about it, just something to think about. :)

ktoso:
Refs https://github.com/akka/akka/issues/18356

kailuowang:
@richdougherty thanks Rich. 

I haven't put this resizer into practice yet. There is an integration test in which the resizer successfully move close to the optimal pool size within a couple dozens tries. I did have some light use of the [actor based implementation](https://github.com/iheartradio/work-pulling-pattern) from which is resizer is adapted. And it also was able to move the pool size towards the optimal one in decent speed - for a 15 second action interval, it took less than 10 minutes to find the optimal size and stick around there. 

Yeah, downsizing to below the highest utilization makes better sense. So I changed it to downsizeRatio which allows more flexibility and made  the default value 0.8. 
I agree it's not very efficient to wait for 72 hour before the first downsize, I guess we can setup two time period, one used only once for initial downsize, and another for regular downsize afterwards. but as you said it does make things a bit complicated. Let me think about it more.

akka-actor-tests/src/test/scala/akka/routing/MetricsBasedResizerSpec.scala
```diff
@@ -0,0 +1,380 @@
+package akka.routing
```
ktoso: Needs a copyright header

akka-actor/src/main/scala/akka/routing/Resizer.scala
```diff
@@ -71,6 +94,7 @@ case object DefaultResizer {
       backoffRate = resizerConfig.getDouble("backoff-rate"),
       messagesPerResize = resizerConfig.getInt("messages-per-resize"))
 
+  @deprecated("User Resizer.fromConfig instead", since = "2.4.1")
```
ktoso: since `2.4.0`, since that's what we're moving towards currently :-)

ktoso: I need to think why/if this actually needs deprecation - is it not possible to make the "new" Resizer fit here?

kailuowang: I wasn't sure here as well. I made this deprecation mostly due to API design. 
I could make `DefaultResizer.fromConfig` to return either a `DefaultConfig` or a `MetricsBasedResizer` based on configuration, but I didn't like this API design. Moving this `fromConfig` method to `Resizer` provides an API that tells people there are multiple implementations of the `Resizer` trait.
It appeared to me that this method is mostly used internally, and that this deprecation would have low short-term impact. But if that's not the case, I can remove this deprecation and make it only returns a `DefaultResizer`, the `Resizer.fromConfig` can continue return either `MetricsBasedResizer` or `DefaultResizer` based on configuration.

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
```diff
@@ -0,0 +1,221 @@
...
+  explorationProbability: Double = 0.4,
+  weightOfLatestMetric: Double = 0.5,
+  /**
+ * Leave package accessible for testing purpose
+ */
+  private[routing] var performanceLog: PerformanceLog = Map.empty,
+  /**
+ * Leave package accessible for testing purpose
+ */
+  private[routing] var record: ResizeRecord = ResizeRecord()) extends MetricsBasedResizer {
```
ktoso: Can we put the vars as normal fields inside the case class then, instead of in the constructor?
Should they participate in hashcode/equality?

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
```diff
@@ -0,0 +1,221 @@
...
+  def resize(currentRoutees: immutable.IndexedSeq[Routee]): Int = {
+    val currentSize = currentRoutees.length
+    val proposedChange =
+      if (record.underutilizationStreak.fold(false)(_.start.isBefore(LocalDateTime.now.minus(downsizeAfterUnderutilizedFor)))) {
+        val downsizeTo = (record.underutilizationStreak.get.highestUtilization * downsizeRatio).toInt
+        Math.min(downsizeTo - currentSize, 0)
+      } else if (performanceLog.isEmpty || record.underutilizationStreak.isDefined) {
+        0
+      } else {
+        if (Random.nextDouble() < explorationProbability)
```
ktoso: please use a `ThreadLocalRandom` (put it in a field), and bound the upper limit of `nextDouble` explicitly

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
kailuowang: Will do. I put them there for convenience when writing unit tests (so that I can create a resizer with these two fields preset in a single statement). Their participation in hashcode/equality didn't worry me.

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
kailuowang: will do.

akka-actor/src/main/scala/akka/routing/Resizer.scala
```diff
@@ -54,6 +57,26 @@ trait Resizer {
...
+      case (false, false) ⇒ None
+      case (true, true) ⇒
+        throw new ResizerInitializationException(s"cannot enable both resizer and metrics-based-resizer", null)
+    }
+  }
+}
+
+@SerialVersionUID(1L)
+class ResizerInitializationException(message: String, cause: Throwable)
+  extends AkkaException(message, cause) {
```
ktoso: skipping the `{}` would be nice here :)

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   * Leave package accessible for testing purpose
+   */
+  private[routing] var performanceLog: PerformanceLog = Map.empty
+  /**
+   * Leave package accessible for testing purpose
+   */
+  private[routing] var record: ResizeRecord = ResizeRecord()
+
+  private val random = ThreadLocalRandom.current()
```
patriknw: this should probably be a `def`, since it is bound to the caller's thread

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
```diff
@@ -0,0 +1,222 @@
...
+  checkParamAsProbability(chanceOfScalingDownWhenFull, "chanceOfScalingDownWhenFull")
+  checkParamAsPositiveNum(numOfAdjacentSizesToConsiderDuringOptimization, "numOfAdjacentSizesToConsiderDuringOptimization")
+  checkParamAsPositiveNum(exploreStepSize, "exploreStepSize")
+  checkParamAsPositiveNum(downsizeRatio, "downsizeRatio")
+  checkParamAsProbability(explorationProbability, "explorationProbability")
+  checkParamAsProbability(weightOfLatestMetric, "weightOfLatestMetric")
+
+  //accessible only for testing purpose
+
+  val historySampleRate: Duration = actionInterval dividedBy 100
```
patriknw: should this be public?

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
```diff
@@ -0,0 +1,222 @@
...
+
+  private[routing] def updatedStats(currentRoutees: immutable.IndexedSeq[Routee], messageCounter: Long): (PerformanceLog, ResizeRecord) = {
+    val now = LocalDateTime.now
+    val currentSize = currentRoutees.length
+
+    val messagesInRoutees = currentRoutees map {
+      case ActorRefRoutee(a: ActorRefWithCell) ⇒
+        a.underlying match {
+          case cell: ActorCell ⇒
+            cell.mailbox.numberOfMessages + (if (cell.currentMessage != null) 1 else 0)
```
patriknw: we must look at the performance of calling `numberOfMessages` for the normal mailboxes, it might not be good

patriknw:
@kailuowang I would like to understand what we are trying to achieve with this router, on a high level. It only works for local actors. It is using the mailbox size (and current processing state) to measure if if the actors are busy. Anything else that it base the decisions on?

In what way is it be better to increase the router pool under high load compared to having a larger pool size in the first place?

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
kailuowang: Actually this one should probably better be configurable. Will make the change once I get back to my laptop.

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
kailuowang: I thought about the performance when writing this implementation. Didn't worry too much because this is only called every action period which is usually more than 5 second and its doesn't block anything other than the resizing effort. On the other hand we have to get the mailbox size to collect the metric. Let me know if there is another better way to collect mailbox that I am not aware of.

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
kailuowang: Konrad suggested place it in a field and I agreed. 
This random is only used when resizing which shouldn't be called concurrently.

kailuowang:
@patriknw, it uses mailbox size to calculate and collect processing speed metrics. And optimize the router pool size based on the size-performence metrics it collected. So instead of simply increasing the pool size under pressure, it finds and stick with the optimal pool size.

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
patriknw: it should be a `def` so that it is bound to the thread that is doing the resizing and not to the one that runs the constructor once

patriknw:
This sounds very cool and such.

It would be interesting to see some kind of benchmark that shows that this actually creates better performance than a fixed size pool (that is not of "optimal" size). Do you think it would be possible to create such a benchmark? We have a the JMH benchmark project where this could perhaps be added: https://github.com/akka/akka/tree/master/akka-bench-jmh

What pool size will it tune in to if you have pure cpu bound tasks?

kailuowang:
@patriknw I did some test with pure cpu bound tasks. The task itself takes around 100 ms to process.  On my i7 4 core mac pro, the resizer tuned in to the pool size of 4-6 after about 4 minutes (one resize per 5 seconds), and stuck around there (occasionally went up to 9-10, but came back down to 4-6 after) for 40 minutes until I terminated the test. 

For the benchmark test, I may need some advices. Should it be a cpu bound task to test against? I originally started the whole thing for I/O bound tasks. Is it possible to test I/O bound tasks in bench-jmh?

patriknw:
Let's skip the jmh benchmark. I'm actually more interested in a few sanity checks, if you don't mind.

It just passed the first sanity check for the cpu bound test :-)

What if you replace the 100 ms cpu bound processing with simulating waiting for IO? I think easiest is to just put a Thread.sleep(100) there. Make sure that you run that test with a dispatcher with a large pool size (e.g. 100 threads) so that is not the limiting factor.

kailuowang:
@patriknw I did the 100 ms sleep test.  This time the resizer obviously couldn't easily find an optimal size (the processing speed is constant regardless of the pool size). During the 30 minutes test, it kept changing the pool size in random directions. It spent the majority of the time between pool size 10 to 18. The highest pool size it went is 20.

patriknw:
@kailuowang Thanks for testing that. Theoretically it would be best at max pool size, since that would give higher throughput for that scenario (as long as the max size is not too big).

My next step with this PR is to review the details, and understand the algorithm.

kailuowang:
@patriknw you are right. Let me double check if my dispatcher was set correct which might prevent the resizer from getting higher throughput with larger pool size.

kailuowang:
@patriknw I updated the dispatcher settings with a different throughput to 1, it was 100 which might be too long for an actor to hold on to a thread (not quite sure this is the cause here). I also changed the actionInterval from 5s to 1s, now the resizer went up to the upperbound (which is 100) in about 10 minutes (equivalent to 50 minutes with a 5s action interval which is the previous setting). 
It keep experimenting between 85 and 100. 
This is the behavior I expected to see.

patriknw:
Sounds good, thanks!

akka-actor-tests/src/test/scala/akka/routing/MetricsBasedResizerSpec.scala
```diff
@@ -0,0 +1,387 @@
...
+      val router = TestRouter(routees(2), resizer)
+      val msgs = router.sendToAll()
+      router.sendToAll() //make sure the routees are still busy after the first batch of messages get processed.
+
+      val before = LocalDateTime.now
+      resizer.reportMessageCount(router.routees, router.msgs.size) //updates the records
+
+      msgs.foreach(_.open()) //process two messages
+
+      Thread.sleep(20)
```
patriknw: why is sleep needed? does the test rely on that something is completed within 20 milliseconds?

akka-actor-tests/src/test/scala/akka/routing/MetricsBasedResizerSpec.scala
```diff
@@ -0,0 +1,387 @@
...
+      val resizer = DefaultMetricsBasedResizer(
+        lowerBound = 3,
+        upperBound = 100,
+        actionInterval = JDuration.ofMillis(20))
+
+      val router = system.actorOf(RoundRobinPool(nrOfInstances = 0, resizer = Some(resizer)).props(
+        Props(new Actor {
+          def receive = {
+            case n: Int if n <= 0 ⇒ //done
+            case n: Int           ⇒ Try(Await.ready(TestLatch(), (n * 200).micros))
```
patriknw: how sensitive is this to GC pauses and random sluggish execution?

akka-actor-tests/src/test/scala/akka/routing/MetricsBasedResizerSpec.scala
```diff
@@ -0,0 +1,387 @@
...
+      def send(): Unit = {
+        val baselineWait = 1
+        val extraWait = Math.max(poolSize(router) - optimalSize, 0) * 3 //punish when pool size go beyond optimal size
+        router ! (baselineWait + extraWait)
+      }
+
+      val numOfMessage = 30000
+      for (i ← 0 until numOfMessage) {
+        send()
+        if (i > 0 && i % 1000 == 0) Thread.sleep(1) //1000 messages per 1 millis
```
patriknw: same question here, I'm worried that we will see false test failures in our CI environment

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
```diff
@@ -0,0 +1,222 @@
...
+   * it can collect metrics.
+   * Caution: this method is not thread safe.
+   *
+   * @param currentRoutees
+   * @param messageCounter
+   */
+  def reportMessageCount(currentRoutees: immutable.IndexedSeq[Routee], messageCounter: Long): Unit
+}
+
+case object DefaultMetricsBasedResizer {
```
patriknw: remove Default from the name? (and also rename the file)

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
```diff
@@ -0,0 +1,222 @@
...
+  private[routing]type PoolSize = Int
+
+  private[routing] case class UnderUtilizationStreak(start: LocalDateTime, highestUtilization: Int)
+
+  private[routing] case class ResizeRecord(
```
patriknw: we add a scaladoc marker 

    /**
     * INTERNAL API
     */

to things that are private to some package, because they are public from Java

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
```diff
@@ -0,0 +1,222 @@
+/**
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.routing
+
+import java.time.{ Duration, LocalDateTime }
```
patriknw: we use `scala.concurrent.duration.Duration` / `FiniteDuration` for all other durations, so I think we should stick to that to make APIs consistent

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
```diff
@@ -0,0 +1,222 @@
...
+  checkParamLowerBound(numOfAdjacentSizesToConsiderDuringOptimization, 2, "numOfAdjacentSizesToConsiderDuringOptimization")
+  checkParamAsProbability(chanceOfScalingDownWhenFull, "chanceOfScalingDownWhenFull")
+  checkParamAsPositiveNum(numOfAdjacentSizesToConsiderDuringOptimization, "numOfAdjacentSizesToConsiderDuringOptimization")
+  checkParamAsPositiveNum(exploreStepSize, "exploreStepSize")
+  checkParamAsPositiveNum(downsizeRatio, "downsizeRatio")
+  checkParamAsProbability(explorationProbability, "explorationProbability")
+  checkParamAsProbability(weightOfLatestMetric, "weightOfLatestMetric")
+
+  def isTimeForResize(messageCounter: Long): Boolean = {
+    record.checkTime.fold(true)(_.isBefore(LocalDateTime.now.minus(actionInterval)))
```
patriknw: `isTimeForResize` is called for each and every message so this must be low overhead.
I think we have to use a `Long` nanoTime instead of LocalDateTime and maybe pattern match on `checkTime` instead of `fold`.

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
```diff
@@ -0,0 +1,222 @@
...
+      if (fullyUtilized && record.underutilizationStreak.isEmpty && record.checkTime.isDefined) {
+        val totalMessageReceived = messageCounter - record.messageCount
+        val queueSizeChange = record.totalQueueLength - totalQueueLength
+        val totalProcessed = queueSizeChange + totalMessageReceived
+        if (totalProcessed > 0) {
+          val duration = Duration.between(record.checkTime.get, now)
+          val speed = duration dividedBy totalProcessed
+          //exponentially decrease the weight of old speed metrics data
+          val toUpdate = performanceLog.get(currentSize).fold(speed) { oldSpeed ⇒
+            val nanos = (oldSpeed.toNanos * (1d - weightOfLatestMetric)) + (speed.toNanos * weightOfLatestMetric)
```
patriknw: s/1d/1.0/

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
```diff
@@ -0,0 +1,222 @@
...
+
+import scala.collection.immutable
+import scala.concurrent.forkjoin.ThreadLocalRandom
+import com.typesafe.config.Config
+
+import akka.actor._
+
+import DefaultMetricsBasedResizer._
+
+trait MetricsBasedResizer extends Resizer {
```
patriknw: I would like to find a better, more specific, name for this. Metrics is so general.
What does it use? mailbox size? throughput?
Perhaps something with Adaptive or Optimizing in the name?
Sorry that I don't have any good suggestions.

akka-actor/src/main/scala/akka/routing/Resizer.scala
patriknw: I think we should remove the deprecation. It still does the right thing. Perhaps not that useful, but we can't deprecated things in minor releases (2.4.1).

akka-actor/src/main/scala/akka/routing/Resizer.scala
```diff
@@ -273,11 +295,13 @@ private[akka] final class ResizablePoolCell(
       resizer.isTimeForResize(resizeCounter.getAndIncrement()) && resizeInProgress.compareAndSet(false, true)) {
       super.sendMessage(Envelope(ResizablePoolActor.Resize, self, system))
     }
+
     super.sendMessage(envelope)
   }
 
   private[akka] def resize(initial: Boolean): Unit = {
     if (resizeInProgress.get || initial) try {
+      tryReportMessageCount()
```
patriknw: is this the only way? can't it be done inside the ordinary `resizer.resize`?  you need the counter? can it have it's own counter?

akka-actor/src/main/scala/akka/routing/Resizer.scala
```diff
@@ -309,10 +343,12 @@ private[akka] class ResizablePoolActor(supervisorStrategy: SupervisorStrategy)
   val resizerCell = context match {
     case x: ResizablePoolCell ⇒ x
     case _ ⇒
-      throw ActorInitializationException("Router actor can only be used in RoutedActorRef, not in " + context.getClass)
+      throw ActorInitializationException("Resizable actor can only be used when resizer is defined, not in " + context.getClass)
```
patriknw: s/Resizable actor/Resizable router actor/ ?

akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteRestartedQuarantinedSpec.scala
```diff
@@ -87,7 +87,7 @@ abstract class RemoteRestartedQuarantinedSpec
 
         testConductor.shutdown(second).await
 
-        within(30.seconds) {
+        within(30.seconds.dilated) {
```
patriknw: within is already dilated

akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteRestartedQuarantinedSpec.scala
```diff
@@ -132,7 +132,7 @@ abstract class RemoteRestartedQuarantinedSpec
         // TODO sometimes it takes long time until the new connection is established,
         //      It seems like there must first be a transport failure detector timeout, that triggers
         //      "No response from remote. Handshake timed out or transport failure detector triggered".
-        probe.expectMsgType[ActorIdentity](30.second).ref should not be (None)
+        probe.expectMsgType[ActorIdentity](30.second.dilated).ref should not be (None)
```
patriknw: expectMsgType is already dilated

akka-actor-tests/src/test/scala/akka/routing/MetricsBasedResizerSpec.scala
kailuowang: This is for routees to update their mailboxs' sizes on the other threads. I just changed it to 1 millis and added some comment. 
I would love to learn if there is a more appropriate way to achieve this.

patriknw: I'm mostly interested in that we don't get test failures if it takes longer time than you expect. Garbage collection pauses up to a second is normal on our CI servers and there can also be other type unexpected cpu scheduling.

akka-actor-tests/src/test/scala/akka/routing/MetricsBasedResizerSpec.scala
kailuowang: This is an integration test. This code is executed 30000 times, so as long as they statistically behave the way it is designed to (processing time proportional to the `n` passed in), the resizer shall be able to find the optimal size and the test should pass.

akka-actor-tests/src/test/scala/akka/routing/MetricsBasedResizerSpec.scala
kailuowang: see response above.

akka-actor-tests/src/test/scala/akka/routing/MetricsBasedResizerSpec.scala
patriknw: I see

akka-actor/src/main/scala/akka/routing/DefaultMetricsBasedResizer.scala
kailuowang: how about ThroughputAdaptiveResizer?

akka-actor/src/main/scala/akka/routing/Resizer.scala
kailuowang: I tried to avoid this as well. But I need the message counter to calculate the number of messages the router passed through during the period (since last resize). This can't be tracked in `resizer.resize` since it's not called on every message. It is possible to count the message in `resizer.isTimeForResize` as it's called every message but it's obviously a bad choice to add state mutation in a inquiry method.
That's why I couldn't think of a better design than the introduction of  tryReportMessageCount()

kailuowang:
@patriknw in regards to the Duration, is there a scala.concurrent.duration.Duration - java.time.Duration conversion class I should use? I know it's trivial to write one, but would love to keep it DRY.

akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteRestartedQuarantinedSpec.scala
kailuowang: I added these dilated code because jenkins failed on them consecutively. Adding them seems fixed. 
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3558/console
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3562/console
I will try remove these after the other changes pass jenkins.

patriknw: I understand, but those are related to something else. Please remove double dilation.
By the way we run with dilation factor 1.0 so this change has no effect.

patriknw:
`java.time.Duration` is all nice, but you should be able to implement things without it, right?
Since we already use scala Duration we can't start mixing and get inconsistencies in APIs.
We might change to `java.time.Duration` but that is a major change.

kailuowang:
just changed to scala.concurrent.duration.Duration and removed the dilated in remote tests.

kailuowang:
@patriknw sorry I missed a previous feedback regarding deprecation. Just made the change to remove it. Please let me know if there is anything else I missed. Thanks very much!

akka-actor/src/main/scala/akka/routing/MetricsBasedResizer.scala
```diff
@@ -0,0 +1,242 @@
...
+      checkTime = System.nanoTime())
+
+    (newPerformanceLog, newRecord)
+
+  }
+
+  def resize(currentRoutees: immutable.IndexedSeq[Routee]): Int = {
+    val currentSize = currentRoutees.length
+    val proposedChange =
+      if (record.underutilizationStreak.fold(false)(_.start.isBefore(LocalDateTime.now.minus(downsizeAfterUnderutilizedFor.asJava)))) {
```
patriknw: place `LocalDateTime.now` outside the fold ?

akka-actor/src/main/scala/akka/routing/MetricsBasedResizer.scala
```diff
@@ -0,0 +1,242 @@
...
+import scala.concurrent.duration._
+
+import com.typesafe.config.Config
+
+import akka.actor._
+import akka.util.JavaDurationConverters._
+
+import MetricsBasedResizer._
+
+trait MetricsBasedResizer extends Resizer {
```
patriknw: I think we should come up with a better name that is somewhat more specific. @ktoso do you have any name suggestion for this thing?

patriknw:
Looks good to me apart from the name and missing documentation.

Please add documentation to 
https://github.com/akka/akka/blob/master/akka-docs/rst/java/routing.rst
https://github.com/akka/akka/blob/master/akka-docs/rst/scala/routing.rst

This will be included in 2.4.1, so we will not merge it until 2.4.0 has been released.
Thanks for contributing.

akka-actor/src/main/scala/akka/routing/MetricsBasedResizer.scala
ktoso: `SelfTuningResizer`?

ktoso: `DynamicSelfOptimisingResizer`?
`OptimalSizeExploringResizer`?

akka-actor/src/main/scala/akka/util/JavaDurationConverters.scala
```diff
@@ -0,0 +1,15 @@
+package akka.util
```
ktoso: missing license header

akka-actor/src/main/scala/akka/util/JavaDurationConverters.scala
```diff
@@ -0,0 +1,15 @@
...
+ * INTERNAL API
+ */
+private[akka] object JavaDurationConverters {
+  final implicit class JavaDurationOps(val self: JDuration) extends AnyVal {
+    def asScala: Duration = Duration.fromNanos(self.toNanos)
+  }
+
+  final implicit class ScalaDurationOps(val self: Duration) extends AnyVal {
+    def asJava: JDuration = JDuration.ofNanos(self.toNanos)
+  }
```
ktoso: cool, thanks

akka-actor/src/main/scala/akka/routing/MetricsBasedResizer.scala
```diff
@@ -0,0 +1,242 @@
...
+ * For documentation about the the parameters, see the reference.conf -
+ * akka.actor.deployment.default.metrics-based-resizer
+ *
+ * CAUTION: [[resize]] also changes the inner state of the resizer,
+ * do not call it unnecessarily or even concurrently
+ *
+ *
+ */
+@SerialVersionUID(1L)
+case class DefaultMetricsBasedResizer(
```
ktoso: Some of the logic must be explained here, if I'm a user I don't know what this resizer is doing based on this doc.
We also have to explain it in https://github.com/akka/akka/blob/master/akka-docs/rst/scala/routing.rst#dynamically-resizable-pool (as sub-section), a small walk-through and rationale is enough :-)

ktoso:
Read through again, looks great :-)
It needs documentation though, otherwise no-one will find or understand what it does based on current scaladocs. Best would be to add docs to routing.rst, and show how it would work on an example scenario.

Thanks a lot, almost there! :-)

akka-actor/src/main/scala/akka/routing/MetricsBasedResizer.scala
kailuowang: OptimalSizeExploring sounds most specific to me. Will use.

patriknw: sounds great, thanks :)

kailuowang:
Added the documentation, renamed to OptimalSizeExploringResizer.
Thanks for all the feedback!

ktoso:
Great docs, LGTM from my side!

patriknw:
@kailuowang Sorry for being slow with the reviews. Docs looks great and this is ready.  Please squash to one commit and we will merge it (and it will be included in 2.4.1). Thanks for a great contribution.

kailuowang:
@patriknw squashed, thanks for all the feedbacks. I am very glad to be able to contribute to this amazing project.

Issue #18176, opened 2015-08-11
failed: ClusterShardingLeavingSpec


patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2759/consoleFull

Fails in the joining phase in the beginning of the test

drewhk:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly2/108/consoleFull

PR #18363, opened 2015-08-31
=all #13783 replace protobuf dependency by embedded version

patriknw:


patriknw:
The first commit is a rebase of https://github.com/akka/akka/pull/16896

The second commit makes ProtobufSerializer independent of protobuf version
* well, as long as they provide the parseFrom and toByteArray
* It is using reflection to find the `parseFrom` and `toByteArray` methods to avoid
  dependency to `com.google.protobuf`.
* also special case com.google.protobuf when loading serialization binding
* migration guide

patriknw:
Refs #13783

project/AkkaBuild.scala
```diff
@@ -87,7 +87,7 @@ object AkkaBuild extends Build {
   lazy val actorTests = Project(
     id = "akka-actor-tests",
     base = file("akka-actor-tests"),
-    dependencies = Seq(testkit % "compile;test->test")
+    dependencies = Seq(testkit % "compile;test->test", protobuf)
```
patriknw: It doesn't look like this is needed

hepin1989:
for it's a embead version ,why not just based on 2.6.1?

patriknw:
Refs  #13783

ktoso:
> for it's a embead version ,why not just based on 2.6.1?

Would that be binary compatible with Akka 2.3 which used this version of protobuf?

hepin1989:
I am using protobuf 2.6.1 and it evicted protobuf 2.5.0,I have not encounter any probleam,but not that sure for a larger scope.

patriknw:
Moving to 2.6.1 would probably be great (assuming that it is wire compatible) if this was done a month ago. Might be risky to do it now, since we are already on bonus time for 2.4.0. Does it solve any issue or increase performance dramatically?

patriknw:
ah, forgot mima

hepin1989:
seems not ,and here is the (changelog)[https://github.com/google/protobuf/blob/master/CHANGES.txt].
anyway ,embed is a good start.

patriknw:
failure was unrelated know issue #18176

patriknw:
PLS BUILD

patriknw:
I did two manual tests to make sure that I didn't messed up anything.

1. Persistent replay of messages stored with old version.

2. Cluster with old and new nodes sending plain protobuf messages (and thereby using the ProtobufSerializer) between them.

All good!

drewhk:
LGTM

patriknw:
rebased

ktoso:
Changes look safe to me, LGTM

Issue #18099, opened 2015-07-30
FAILED: TcpSpec


ktoso:
`Expected exception akka.stream.StreamTcpException to be thrown, but java.util.concurrent.TimeoutException was thrown`
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3304/consoleFull
```
[info] - must handle when connection actor terminates unexpectedly *** FAILED *** (3 seconds, 36 milliseconds)
[info]   Expected exception akka.stream.StreamTcpException to be thrown, but java.util.concurrent.TimeoutException was thrown. (TcpSpec.scala:384)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.Assertions$class.newAssertionFailedException(Assertions.scala:496)
[info]   at org.scalatest.Assertions$.newAssertionFailedException(Assertions.scala:1538)
[info]   at org.scalatest.Assertions$.checkExpectedException(Assertions.scala:1572)
[info]   at org.scalatest.words.ResultOfBeWordForAType.thrownBy(ResultOfBeWordForAType.scala:39)
[info]   at akka.stream.io.TcpSpec$$anonfun$1$$anonfun$apply$mcV$sp$27.apply$mcV$sp(TcpSpec.scala:397)
[info]   at akka.stream.io.TcpSpec$$anonfun$1$$anonfun$apply$mcV$sp$27.apply(TcpSpec.scala:384)
[info]   at akka.stream.io.TcpSpec$$anonfun$1$$anonfun$apply$mcV$sp$27.apply(TcpSpec.scala:384)
[info]   at org.scalatest.Transformer$$anonfun$apply$1.apply$mcV$sp(Transformer.scala:22)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.WordSpecLike$$anon$1.apply(WordSpecLike.scala:953)
[info]   at org.scalatest.Suite$class.withFixture(Suite.scala:1122)
[info]   at akka.stream.testkit.AkkaSpec.withFixture(AkkaSpec.scala:55)
[...]

PR #18369, opened 2015-08-31
+str #18099 FAILED: TcpSpec

agolubev:
there was a race condition between two connections. From time to time second connection that supposed to be rejected connects before the first one.

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/io/TcpSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/io/TcpSpec.scala
@@ -15,7 +15,7 @@ import akka.stream.{ ActorMaterializer, BindFailedException, StreamTcpException 
 import akka.util.{ ByteString, Helpers }
 
 import scala.collection.immutable
-import scala.concurrent.Await
+import scala.concurrent.{ Promise, Await }
 import scala.concurrent.duration._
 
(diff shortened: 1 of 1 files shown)


agolubev:
Ref #18099

2m:
OK TO TEST

akka-stream-tests/src/test/scala/akka/stream/io/TcpSpec.scala
```diff
@@ -503,7 +503,11 @@ class TcpSpec extends AkkaSpec("akka.io.tcp.windows-connection-abort-workaround-
 
     "shut down properly even if some accepted connection Flows have not been subscribed to" in assertAllStagesStopped {
       val address = temporaryServerAddress()
-      val takeTwoAndDropSecond = Flow[IncomingConnection].grouped(2).take(1).map(_.head)
+      val firstClientConnected = Promise[Unit]()
+      val takeTwoAndDropSecond = Flow[IncomingConnection].map(conn ⇒ {
+        if (!firstClientConnected.isCompleted) firstClientConnected.success()
```
drewhk: no need for the if, just use trySuccess()

drewhk:
LGTM

drewhk:
Test failure is weird...

akka-stream-tests/src/test/scala/akka/stream/io/TcpSpec.scala
agolubev: will change

agolubev:
Test "Recipe for a droppy broadcast.must work" failed. I think it does not "work" :)

drewhk:
LGTM, Thanks again!

Issue #18342, opened 2015-08-27
doc: Akka persistence complains about missing configuration which is present


reinierl:
When I configure akka-persistence to use a LevelDB journal plugin as documented on e.g. http://doc.akka.io/docs/akka/2.4.0-RC1/scala/persistence.html#Local_LevelDB_journal, it fails with:

```
[info] [ERROR] [08/27/2015 16:23:14.630] [nuvveproxysystem-akka.actor.default-dispatcher-6] [akka://nuvveproxysystem/user/cpmanager] requirement failed: 'reference.conf' is missing persistence plugin config path: 'leveldb'
[info] akka.actor.ActorInitializationException: exception during creation
[info] 	at akka.actor.ActorInitializationException$.apply(Actor.scala:172)
[info] 	at akka.actor.ActorCell.create(ActorCell.scala:605)
[info] 	at akka.actor.ActorCell.invokeAll$1(ActorCell.scala:460)
[info] 	at akka.actor.ActorCell.systemInvoke(ActorCell.scala:482)
[info] 	at akka.dispatch.Mailbox.processAllSystemMessages(Mailbox.scala:282)
[info] 	at akka.dispatch.Mailbox.run(Mailbox.scala:223)
[info] 	at akka.dispatch.Mailbox.exec(Mailbox.scala:234)
[info] 	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
[info] 	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
[info] 	at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
[info] 	at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
```

the relevant reference.conf part is:

```
akka {
  persistence {
    journal {
      plugin = "leveldb"

      leveldb {
        dir = "target/persistence/journal"
      }
    }

    snapshot-store.local.dir = "target/persistence/snapshots"
  }
}
```

If I add the full leveldb plugin configuration at the root level in the config, it starts working. That is when the reference.conf contains this:

```
akka {
  persistence {
    journal {
      plugin = "leveldb"

      leveldb {
        dir = "target/persistence/journal"
      }
    }

    snapshot-store.local.dir = "target/persistence/snapshots"
  }
}

leveldb {
  dir = "target/persistence/journal"
[...]

PR #18376, opened 2015-09-01
=per #18342 Improve docs for leveldb plugin config

patriknw:
- clarify how to enable the plugin
- added empty class property in fallback config in reference
  to have a proper place to document that and throw a more
  specific exception if it is not defined
- also some formatting of reference.conf

### DIFF
--- a/akka-docs/rst/java/persistence.rst
+++ b/akka-docs/rst/java/persistence.rst
@@ -525,6 +525,8 @@ saved snapshot matches the specified ``SnapshotSelectionCriteria`` will replay a
   Since it is acceptable for some applications to not use any snapshotting, it is legal to not configure a snapshot store,
   however Akka will log a warning message when this situation is detected and then continue to operate until
   an actor tries to store a snapshot, at which point the the operation will fail (by replying with an ``SaveSnapshotFailure`` for example).
+  
+  Note that :ref:`cluster_sharding_java` is using snapshots, so if you use Cluster Sharding you need to define a snapshot store plugin.
 
 Snapshot deletion
 -----------------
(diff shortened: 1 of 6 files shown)


patriknw:
Refs #18342

patriknw:
Please review @ktoso

akka-docs/rst/java/persistence.rst
```diff
@@ -859,7 +876,11 @@ Local snapshot store
 --------------------
 
 Local snapshot store plugin config entry is ``akka.persistence.snapshot-store.local`` and it writes snapshot files to
-the local filesystem. The default storage location is a directory named ``snapshots`` in the current working
+the local filesystem. Enable this plugin by defining config property:
```
ktoso: Maybe add an explicit note that it's OK to run without a snapshot store configured even if you use persistence, until you actually call saveSnapshot somewhere (and that cluster sharding will use snapshotting?).

akka-persistence/src/main/resources/reference.conf
```diff
@@ -5,32 +5,29 @@
...
-#           The class must have a constructor without parameters or constructor with 
-#           one `com.typesafe.config.Config` parameter.
-# `plugin-dispatcher` : Absolute configuration path to the akka dispatcher configuration entry. This string field is optional.  
-
-# Note that journal and snapshot store plugins included with the extension are suitable for testing purposes only.
-# You should change extension defaults or override `journalPluginId` and `snapshotPluginId` in the persistent actor or view.
-# Directory of persistence journal and snapshot store plugins is available at the Akka Community Projects page 
-# http://akka.io/community/
+# Directory of persistence journal and snapshot store plugins is available at the 
+# Akka Community Projects page http://akka.io/community/
```
ktoso: good idea

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
```diff
@@ -253,7 +253,11 @@ class Persistence(val system: ExtendedActorSystem) extends Extension {
 
   private def createPlugin(configPath: String, pluginConfig: Config): ActorRef = {
     val pluginActorName = configPath
-    val pluginClassName = pluginConfig.getString("class")
+    val pluginClassName = pluginConfig.getString("class") match {
+      case "" ⇒ throw new IllegalArgumentException("Plugin class name must be defined in config property " +
```
ktoso: good catch

ktoso:
LGTM, maybe add the note about snapshot store unless it's mentioned somewhere already

patriknw:
I added some more notes about that snapshot store is optional

akka-docs/rst/scala/persistence.rst
```diff
@@ -516,6 +516,8 @@ saved snapshot matches the specified ``SnapshotSelectionCriteria`` will replay a
   Since it is acceptable for some applications to not use any snapshotting, it is legal to not configure a snapshot store,
   however Akka will log a warning message when this situation is detected and then continue to operate until
   an actor tries to store a snapshot, at which point the the operation will fail (by replying with an ``SaveSnapshotFailure`` for example).
+  
+  Note that :ref:`cluster_sharding_scala` is using snapshots, so if you use Cluster Sharding you need to define a snapshot store plugin.
```
ktoso: Thanks :)

PR #18382, opened 2015-09-01
=ben add benchmark runner and move stream benchmarks to dev project

2m:
Added benchmark runner which prints out results in a tab separated format so it can be included into [Google Docs](https://docs.google.com/spreadsheets/d/1Sdqx2qj_L6-gpiXerCZHQcmAWzgh9p-pN2N0lBi87Wk/edit?usp=sharing) aggregate document easily.

Also moved stream benchmarks to akka-jmh-bench-dev project, so all stream related benchmarks can be run by `akka-bench-jmh-dev/run`.


akka-bench-jmh-dev/src/main/scala/akka/BenchRunner.scala
```diff
@@ -0,0 +1,28 @@
...
+import org.openjdk.jmh.results.RunResult
+import org.openjdk.jmh.runner.Runner
+import org.openjdk.jmh.runner.options.CommandLineOptions
+
+object BenchRunner extends App {
+  import scala.collection.JavaConversions._
+
+  val args2 = args.toList match {
+    case "quick" :: tail => "-i 1 -wi 1 -f1 -t1".split(" ").toList ::: tail
+    case "thorough" :: tail => "-i 10 -wi 10 -f1 -t1".split(" ").toList ::: tail
```
ktoso: I can see myself mis-smelling this a 100 times (is there a simpler word for it?):wink:

ktoso:
LGTM :)

akka-bench-jmh-dev/src/main/scala/akka/BenchRunner.scala
drewhk: I love these shortcuts! Btw, the thorough should have 3 forks, maybe at the cost of less (warmup) iterations.

patriknw: full?

patriknw:
LGTM

akka-bench-jmh-dev/src/main/scala/akka/BenchRunner.scala
ktoso: full would be great :)

2m:
Changed "thorough" to "full", changed to 3 threads and decresed warmup iterations to 4. It takes around 30 minutes to run all the benchmarks now.

Issue #18091, opened 2015-07-29
Akka Streams: .take(0) should always immediately complete.


AnIrishDuck:
This:

```scala
Await.result(Source.empty.take(0).runWith(Sink.ignore), 1.second)
```
Successfully completes. OTOH, this:

```scala
Await.result(Source.lazyEmpty.take(0).runWith(Sink.ignore), 1.second)
```

Does not. This seems counter-intuitive and wrong: taking zero elements should always yield a stream that immediately completes.

This issue was present in several 1.0 release candidates and is also present in 1.0 itself.

drewhk:
Thanks for reporting.

AnIrishDuck:
Just ran into this, which seems related:

```scala
var requested = 0L

val test = new Publisher[Long]() {
  override def subscribe(s: Subscriber[_ >: Long]) = s.onSubscribe(new Subscription() {
    def cancel() = { }
    def request(l: Long) = {
      requested += l
      (0L until l).foreach(s.onNext)
    }
  })
}

Await.result(Source(test).take(4).runWith(Sink.ignore), 2.seconds)

requested ==== 4
```

Gives this:

```
'8' is not equal to '4'
Expected :4
Actual   :8
```

It appears that what `take(n)` requests from downstream is unrelated to how many items it is capped at emitting.

IMHO if `take(n)` knows exactly how many elements it needs, it should only request that amount.

drewhk:
Addresssed by https://github.com/akka/akka/pull/18386
As for the second issue, it is a duplicate of (https://github.com/akka/akka/issues/16967) it has been fixed earlier via this: https://github.com/akka/akka/pull/18094

PR #18386, opened 2015-09-02
=str #18091: Take should eagerly complete if arguments are zero or less

drewhk:


ktoso:
Nice, LGTM

Refs #18091

Issue #18377, opened 2015-09-01
FAILED: RecipeDroppyBroadcast


drewhk:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3569/consoleFull

PR #18387, opened 2015-09-02
=str #18377: Harden DroppyBroadcastSpec to not lose messages

drewhk:
On Jenkins the publisher might be too fast compared to the consumer in some cases and therefore losing the elements. This modification keeps the two rates tightly coupled so it is no longer flaky.


drewhk:
Refs #18387

drewhk:
I was a nasty boy and merged this: since this fails spuriously and the logic is super simple here.

ktoso:
Post merge LGTMing anyway :-)

PR #18393, opened 2015-09-03
=str #17226 remove in-place DropWhile impl, it exists now

ktoso:
Replaces in-place impl of a drop while with the proper one

refs #17226


drewhk:
LGTM

PR #18395, opened 2015-09-03
=per #18394 persist EventAdapter manifest in default proto serializer

### LINKED ISSUE #18394: EventAdapter's manifest not persisted in default proto serializer
Refs https://github.com/krasserm/akka-persistence-cassandra/issues/75

ktoso:
Resolves #18394

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/serialization/MessageSerializer.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/serialization/MessageSerializer.scala
@@ -130,6 +130,7 @@ class MessageSerializer(val system: ExtendedActorSystem) extends BaseSerializer 
 
     if (persistent.persistenceId != Undefined) builder.setPersistenceId(persistent.persistenceId)
     if (persistent.sender != Actor.noSender) builder.setSender(Serialization.serializedActorPath(persistent.sender))
+    if (persistent.manifest != PersistentRepr.Undefined) builder.setManifest(persistent.manifest)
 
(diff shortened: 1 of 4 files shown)


akka-persistence/src/main/scala/akka/persistence/serialization/MessageSerializer.scala
```diff
@@ -130,6 +130,7 @@ class MessageSerializer(val system: ExtendedActorSystem) extends BaseSerializer
 
     if (persistent.persistenceId != Undefined) builder.setPersistenceId(persistent.persistenceId)
     if (persistent.sender != Actor.noSender) builder.setSender(Serialization.serializedActorPath(persistent.sender))
+    if (persistent.manifest != PersistentRepr.Undefined) builder.setManifest(persistent.manifest)
```
patriknw: ouch

patriknw:
LGTM, thanks for the quick fix

akka-persistence/src/main/scala/akka/persistence/serialization/MessageSerializer.scala
ktoso: Yeah :< Wasn't caught by tests since they used the in-mem journal. I'll also want to enable the same tests using leveldb and then merge this.

ktoso:
Enabled all tests on leveldb as well (some were only inmem, since it's not simple to have many instances of the leveldb plugin running which the tests needed)

ktoso:
This is complete now, merge at will :)

patriknw:
thanks, will merge after validation

PR #18398, opened 2015-09-04

patriknw:


patriknw:
Refs #13584

patriknw:
@drewhk please take a look at 7267f75 Did you have anything else in mind for the doc additions?

drewhk:
Can WeaklyUp members participate in sharding, or in distributed-data? Maybe that should be clarified as well. Otherwise LGTM

patriknw:
Good point. I have added that. In summary:
singleton: no
sharding: no
pub/sub: yes
metrics: yes
routers: yes
distributed data: currently not, but should be supported as described in #18399

Issue #18226, opened 2015-08-15
document context.stop(self)


rkuhn:
The Actors page (Stopping Actors) should point this out prominently and have a code snippet for it as well. self-stopping should also be mentioned near the lifecycle diagram.

PR #18413, opened 2015-09-08
=doc #18226 document actor stopping

johanandren:
Documents how/from where an actor can be stopped, includes code samples for both Java and Scala


patriknw:
Refs #18226

patriknw:
LGTM

Issue #18072, opened 2015-07-24
concat should subscribe to next when prev completes


viktorklang:
From @kodemaniak:

~~~scala
Source.single(1).concat(Source.empty)
        .runWith(TestSink.probe[Int])
        .request(1)
        .expectNext(1)
        .expectComplete()
~~~

This ought to work but doesn't. subscriptions are not demand driven and onComplete and onError should be sent without demand being present, they are termination signals.

PR #18428, opened 2015-09-09
+str #18072 fix concat with empty source

agolubev:
it handle two cases: when empty source concat with another empty source and when non empty concats with empty source.
Fix does not do exactly what issue #18072 suggests (concat should subscribe to next when prev completes) it still subscribes for all sources at once. However fix allows to send onComplete outside of pump.
So let me know what do you think.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
@@ -56,6 +56,9 @@ private[akka] object FanIn {
     private var markedPending = 0
     private var markedDepleted = 0
 
+    private var receivedInput = false
+    private var completedCounter = 0
+
     private[this] final def hasState(index: Int, flag: Int): Boolean =
       (states(index) & flag) != 0
     private[this] final def setState(index: Int, flag: Int, on: Boolean): Unit =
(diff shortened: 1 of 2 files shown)


drewhk:
OK TO TEST

akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
```diff
@@ -151,6 +155,8 @@ private[akka] object FanIn {
 
     def isCancelled(input: Int): Boolean = cancelled(input)
 
+    def isAllCompleted(): Boolean = (0 to inputCount - 1).map(completed).forall(_ == true)
```
viktorklang: why map + forall?

agolubev: yeah, forall is enough

drewhk: this is a linear scan. Can't this be implemented as a counter, so evaluating it is constant time?

viktorklang: +1 @drewhk

drewhk: OTOH I will probably remove this file completely in the upcoming days, so I am not sure how much optimization effort should be put into this, but the counter feels simple enough. We need the test cases though anyway.

agolubev: Can add counter today.
> We need the test cases though anyway.

Have one test case with Source.empty.concat(Source.empty). Are you thinking about other scenarios?

drewhk: No I just meant that even though this class will be deprecated, the test cases reproducing the original issue will stay and will be very useful.

agolubev: I see. 
Actually can also add test case where it's shown that subscription happening at the beginning for all sources even before pump starts working. Do you think it will be useful?

drewhk: If it is easy to add, then please do it. At least it will indicate if something has changed.

agolubev:
rebased

agolubev:
@drewhk, counter and test case are added.

drewhk:
LGTM

drewhk:
Thank you again, great job as usual!

agolubev:
Cool! Thanks.

Issue #17317, opened 2015-04-27
FAILED failure while attaching new conductor


ktoso:
```
[info] * akka.remote.RemoteNodeDeathWatchSlow
[JVM-2] [INFO] [04/27/2015 01:13:31.327] [ScalaTest-main] [akka.remote.Remoting] Starting remoting
[JVM-2] [INFO] [04/27/2015 01:13:31.644] [ScalaTest-main] [akka.remote.Remoting] Remoting started; listening on addresses :[akka.tcp://RemoteNodeDeathWatchSpec@a5.local:57475]
[JVM-3] [INFO] [04/27/2015 01:13:34.928] [ScalaTest-main] [akka.remote.Remoting] Starting remoting
[JVM-3] [INFO] [04/27/2015 01:13:36.124] [ScalaTest-main] [akka.remote.Remoting] Remoting started; listening on addresses :[akka.tcp://RemoteNodeDeathWatchSpec@a6.local:36278]
[JVM-1] [INFO] [04/27/2015 01:13:39.020] [ScalaTest-main] [akka.remote.Remoting] Starting remoting
[JVM-1] [INFO] [04/27/2015 01:13:41.963] [ScalaTest-main] [akka.remote.Remoting] Remoting started; listening on addresses :[akka.tcp://RemoteNodeDeathWatchSpec@a0.local:54866]
[JVM-2] *** RUN ABORTED *** (11 seconds, 651 milliseconds)
[JVM-2]   java.lang.RuntimeException: failure while attaching new conductor
[JVM-2]   at akka.remote.testkit.MultiNodeSpec.attachConductor(MultiNodeSpec.scala:387)
[JVM-2]   at akka.remote.testkit.MultiNodeSpec.<init>(MultiNodeSpec.scala:392)
[JVM-2]   at akka.remote.testkit.MultiNodeSpec.<init>(MultiNodeSpec.scala:252)
[JVM-2]   at akka.remote.RemoteNodeDeathWatchSpec.<init>(RemoteNodeDeathWatchSpec.scala:79)
[JVM-2]   at akka.remote.RemoteNodeDeathWatchSlowSpec.<init>(RemoteNodeDeathWatchSpec.scala:73)
[JVM-2]   at akka.remote.RemoteNodeDeathWatchSlowMultiJvmNode2.<init>(RemoteNodeDeathWatchSpec.scala:71)
[JVM-2]   at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
[...]

Issue #18339, opened 2015-08-27
failed: SurviveNetworkInstabilitySpec


patriknw:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3534/consoleFull

drewhk:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2875/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3632/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly2/119/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/3033/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/3036/consoleFull

PR #18430, opened 2015-09-09
=mnt #17317 Harden reconnect when attaching to test conductor

patriknw:
The failing tests showed big variations in startup times, i.e.
JVM1 not started/bound when other JVMs tries to connect to the
test conductor. The reconnect logic was in place but it was
bound to 10 client-reconnects (10 seconds). It's better to let
it by default reconnect until the connect-timeout (20 seconds)
is elapsed.


patriknw:
Refs #17317

drewhk:
LGTM

patriknw:
failure was unrelated: https://github.com/akka/akka/issues/18339

patriknw:
PLS BUILD

PR #18432, opened 2015-09-09
=cls #18176 Harden ClusterShardingLeavingSpec

patriknw:
In logs it is clear that the fourth node is moved to Up,
but it takes more than 5 sec to disseminate that info


patriknw:
Refs #18176

PR #18435, opened 2015-09-09
=rem #17556 Do not fail if identity or terminated arrives

johanandren:
Trying to stabilize RemoteNodeShutdownAndComesBackSpec which fails
intermittently on the test servers by replacing an unecessary check
with receiveWhile which does not fail if it times out.

Cannot repeat failure locally, so this is a "maybe" fixes the problem after discussions on gitter.

### DIFF
--- a/akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteNodeShutdownAndComesBackSpec.scala
+++ b/akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteNodeShutdownAndComesBackSpec.scala
@@ -118,9 +118,9 @@ abstract class RemoteNodeShutdownAndComesBackSpec
         watch(subjectNew)
 
         subjectNew ! "shutdown"
-        fishForMessage(5.seconds) {
-          case _: ActorIdentity       ⇒ false
-          case Terminated(subjectNew) ⇒ true
+        // we are waiting for a Terminated here, but it is ok if it does not arrive
+        receiveWhile(5.seconds) {
+          case _: ActorIdentity ⇒ true
         }
       }
 


akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteNodeShutdownAndComesBackSpec.scala
```diff
@@ -118,8 +118,8 @@ abstract class RemoteNodeShutdownAndComesBackSpec
         watch(subjectNew)
 
         subjectNew ! "shutdown"
-        fishForMessage(5.seconds) {
-          case _: ActorIdentity       ⇒ false
+        receiveWhile(5.seconds) {
+          case _: ActorIdentity       ⇒ true
```
drewhk: shouldn't this be false? I.e. we keep pulling the stray ActorIdentity responses from previous test steps until we find the Terminated?

patriknw:
The test relied on that the DeathWatchNotification is sent out immediately when the actor system is terminated. That is best effort so it could potentially be missed. It could be something else, though.

akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteNodeShutdownAndComesBackSpec.scala
johanandren: Might be my mistake, but I understood receiveWhile such that it is the matching that matters, not the return value of the PF?

akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteNodeShutdownAndComesBackSpec.scala
```diff
@@ -118,8 +118,8 @@ abstract class RemoteNodeShutdownAndComesBackSpec
...
           case Terminated(subjectNew) ⇒ true
```
patriknw: I think this will always take 5 seconds now. We would like to continue earlier if we receive the `Terminated`, because then we know that the other side is done.

receiveWhile:

     * Receive a series of messages until one does not match the given partial
     * function or the idle timeout is met (disabled by default) or the overall
     * maximum duration is elapsed or expected messages count is reached.
     * Returns the sequence of messages.

If I understand that correctly it should work if you remove the `case Terminated` (and write a comment).
but you better try it out, so that I'm not wrong.

johanandren: That was my first idea, but then I thought you corrected me to do it like this, good because I didn't really understand how this would work (except for always waiting the 5 seconds).

patriknw:
You were right, I didn't know exactly how it worked.

akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteNodeShutdownAndComesBackSpec.scala
```diff
@@ -118,9 +118,8 @@ abstract class RemoteNodeShutdownAndComesBackSpec
         watch(subjectNew)
 
         subjectNew ! "shutdown"
-        fishForMessage(5.seconds) {
-          case _: ActorIdentity       ⇒ false
-          case Terminated(subjectNew) ⇒ true
+        receiveWhile(5.seconds) {
+          case _: ActorIdentity ⇒ true
```
patriknw: a comment here would be great, saying that we expect Terminated but alright if it doesn't arrive

patriknw:
LGTM

drewhk:
LGTM, great work, @johanandren!

johanandren:
PLS BUILD

johanandren:
Something going on with the jenkins server right now "Permission denied"

patriknw:
I don't know. Might be another restart by Toni.

patriknw:
Refs #17556

PR #18436, opened 2015-09-09
=doc Clarify sharding data compatibility

patriknw:


smlin:
@patriknw I think you must to clarify that in most cases a sharding data could be simple removed.  Since during migration from 2.3 to 2.4 a cluster anyway will restarted.

patriknw:
@smlin good point, updated with new attempt

patriknw:
@ktoso good to squash and merge?

akka-docs/rst/project/migration-guide-2.3.x-2.4.x.rst
```diff
@@ -288,6 +288,26 @@ actor external actor of how to allocate shards or rebalance shards.
...
+it should not load such old data, but it can be a problem if you have used a 2.4
+milestone release. In that case you should remove the persistent data that the 
+Cluster Sharding coordinator stored. Note that this is not application data.
+
+Consult the Journal Plugin and Snapshot Store Plugin documentation or backend
+data store documentation for information about how to remove such data.
+
+The new ``persistenceId`` is ``s"/sharding/${typeName}Coordinator"``.
+The old ``persistenceId`` is ``s"/user/sharding/${typeName}Coordinator/singleton/coordinator"``.  
+
```
ktoso: I recently had this idea that this could be made easier if we'd make the coordinator env variable aware of sth like `-Dakka.sharding.pleaseDeleteShardingDataOnStart=true`, sounds ticket worthy @patriknw?

ktoso:
Docs LGTM

akka-docs/rst/project/migration-guide-2.3.x-2.4.x.rst
patriknw: That is a great suggestion. I think it is difficult to ensure that it is only run once (on the first coordinator node when starting up fresh cluster). The safe variant, which I prefer, is to provide it as a standalone utility that can be run from CLI before starting up cluster.

ktoso: Yup, good idea to make it manual instead. I opened a ticket https://github.com/akka/akka/issues/18485

patriknw:
merging this now and we can amend the docs when we do the toool

Issue #18245, opened 2015-08-18
Misleading example in "How To Cleanup when Member is Removed"


patriknw:
http://doc.akka.io/docs/akka/snapshot/scala/cluster-usage.html#How_To_Cleanup_when_Member_is_Removed

scheduleOnce is always run when system is terminated.
Here it doesn't matter since we do System.exit(-1) in both cases.
Would be better with example that does exit(-1) vs exit(0).
Perhaps use `awaitTermination(10.seconds)` and catch `TimeoutException`

Issue #16495, opened 2014-12-09
Remove the need for Scheduler to run all callbacks on shutdown


rkuhn:
This should be mentioned in the API docs of the individual methods. We might also think about adding the ability to opt out of this behavior.

jroper:
Just another datapoint for confusion of scheduled tasks being executed before shutdown:

https://github.com/playframework/playframework/issues/4905

I've noticed this problem myself before too. I can understand that for schedule once tasks, it would often, or maybe usually be useful that they be executed at shutdown, but recurring tasks don't make sense to me - in my particular use case, I had a recurring task that was polling Play contributor data out of GitHub every 24 hours. Running this task again on shutdown doesn't make sense, and especially in the context of Play, where every change to the code base results in an Akka shutdown.

rkuhn:
Yes, indeed, that is not usually what is intended. We’ll have to rethink that part—possibly by drawing the line between the ActorSystem’s Scheduler (that is only intended to be used for timeouts) and a to-be-written component that can send messages at particular calendar times. We might even need to rename things to make the difference clear.

patriknw:
this is now a duplicate of https://github.com/akka/akka/issues/20256 keeping this one

patriknw:
Another workaround: Use `system.registerOnTermination` and in that callback cancel your scheduled task. Your callback will be run before the scheduler is shutdown, which is what makes the scheduled task to run.

PR #18441, opened 2015-09-10
=clu #18245 Improve registerOnMemberRemoved example

patriknw:
I will convert the java counterpart after feedback.

### DIFF
--- a/akka-samples/akka-sample-cluster-scala/src/main/scala/sample/cluster/factorial/FactorialFrontend.scala
+++ b/akka-samples/akka-sample-cluster-scala/src/main/scala/sample/cluster/factorial/FactorialFrontend.scala
@@ -9,6 +9,8 @@ import akka.actor.Props
 import akka.cluster.Cluster
 import akka.routing.FromConfig
 import akka.actor.ReceiveTimeout
+import scala.util.Try
+import scala.concurrent.Await
 
 //#frontend
 class FactorialFrontend(upToN: Int, repeat: Boolean) extends Actor with ActorLogging {
@@ -61,12 +63,20 @@ object FactorialFrontend {
     //#registerOnRemoved
     Cluster(system).registerOnMemberRemoved {
       // exit JVM when ActorSystem has been terminated
-      system.registerOnTermination(System.exit(-1))
-      // in case ActorSystem shutdown takes longer than 10 seconds,
-      // exit the JVM forcefully anyway
-      system.scheduler.scheduleOnce(10.seconds)(System.exit(-1))(system.dispatcher)
+      system.registerOnTermination(System.exit(0))
       // shut down ActorSystem
       system.terminate()
+
+      // In case ActorSystem shutdown takes longer than 10 seconds,
+      // exit the JVM forcefully anyway.
+      // We must spawn a separate thread to not block current thread,
+      // since that would have blocked the shutdown of the ActorSystem.
+      new Thread {
+        override def run(): Unit = {
+          if (Try(Await.ready(system.whenTerminated, 10.seconds)).isFailure)
+            System.exit(-1)
+        }
+      }.start()
     }
     //#registerOnRemoved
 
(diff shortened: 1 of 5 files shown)


akka-samples/akka-sample-cluster-scala/src/main/scala/sample/cluster/factorial/FactorialFrontend.scala
```diff
@@ -61,10 +61,18 @@ object FactorialFrontend {
...
+      // exit the JVM forcefully anyway.
+      // Use a Deadline since the scheduled task will also be run when
+      // the ActorSystem is terminated normally.
+      val shutdownDeadline = Deadline.now + 10.seconds
+      system.scheduler.scheduleOnce(shutdownDeadline.timeLeft + 10.millis) {
+        if (shutdownDeadline.isOverdue())
+          System.exit(-1)
+        // otherwise it was a successful shutdown and we exit the JVM via the
+        // system.registerOnTermination hook
+      }(system.dispatcher)
```
patriknw: This is in fact very complicated. I can't find an easier way that is correct so please speak up if you have any idea.

I don't want to use Await to block inside the registerOnMemberRemoved since that would block the actor system shutdown.

Using the system termination Future with `after` doesn't help either, as far as I can see.

akka-samples/akka-sample-cluster-scala/src/main/scala/sample/cluster/factorial/FactorialFrontend.scala
```diff
@@ -61,10 +61,18 @@ object FactorialFrontend {
...
+      system.registerOnTermination(System.exit(0))
       // in case ActorSystem shutdown takes longer than 10 seconds,
-      // exit the JVM forcefully anyway
-      system.scheduler.scheduleOnce(10.seconds)(System.exit(-1))(system.dispatcher)
+      // exit the JVM forcefully anyway.
+      // Use a Deadline since the scheduled task will also be run when
+      // the ActorSystem is terminated normally.
+      val shutdownDeadline = Deadline.now + 10.seconds
+      system.scheduler.scheduleOnce(shutdownDeadline.timeLeft + 10.millis) {
+        if (shutdownDeadline.isOverdue())
```
drewhk: Ah, this is because the scheduler runs pending events on shutdown? Would it be a heresy to use a normal JVM timer here?

patriknw: I would not go near JVM timers. I have had so many problems with them (memory leaks).

We have this issue https://github.com/akka/akka/issues/16495 regarding the scheduler shutdown. At some point we must do something about that. It's so unintuitive.

drewhk: But the JVM timer is not dangerous here since
 - it is only used for one task and only during shutdown
 - memory leak is not relevant since the JVM will be killed anyway

Another alternative is to spawn a deamon thread that is calling the Await for the system shutdown and aborts if necessary.

patriknw: yes, but since the System.exit is just an example and users will copy paste without thinking I would prefer to not involve external timers.

A separate thread is pretty good. Is it better than this hack?

drewhk: I think it is better, because that check for .isOverdue is hard to understand (scheduler fires events on shutdown). Using a Thread, while trivial is easy to explain at least, and I think it is a valid explanation that you should not use the scheduler of the thing that you want to shut down to coordinate the shutdown itself.

This might be a pattern that we should provide though.

patriknw: I agree. I'll change to that. Thanks for the suggestion.

patriknw:
@drewhk I have changed to the thread variant

drewhk:
LGTM

patriknw:
Refs #18245

Issue #18439, opened 2015-09-10
PersistentRepr.manifest cleaned up by IdentityEventAdapters


Horusiath:
`WriteJournalBase.adaptToJournal` method uses event adapters to map PersistentRepr &rArr; PersistentRepr instance. Unfortunately I've seen that manifest is always taken from event adapters, and never retained from an original message. In default configuration - `IdentityEventAdapters` - this means, that the input message manifest is always erased.

patriknw:
Normally there is no manifest in the incoming PersistentRepr, but as @Horusiath explained to me for test purposes one may want to send `WriteMessages` directly to the journal.

ktoso:
Good catch, thanks for reporting @Horusiath!

PR #18442, opened 2015-09-10
=per #18439 Preserve original manifest for IdentityEventAdapter

patriknw:
- when WriteMessages is sent directly to the  journal for testing
  purposes we want to preserve the original manifest instead of
  letting IdentityEventAdapter clearing it out

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/journal/WriteJournalBase.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/journal/WriteJournalBase.scala
@@ -6,7 +6,6 @@ package akka.persistence.journal
 
 import akka.actor.Actor
 import akka.persistence.{ Persistence, PersistentEnvelope, PersistentRepr }
-
 import scala.collection.immutable
 import akka.persistence.AtomicWrite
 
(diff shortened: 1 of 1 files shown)


patriknw:
Refs #18439

patriknw:
unrelated failure, which is about to be fixed by https://github.com/akka/akka/pull/18432

drewhk:
I guess this is LGTM but I might not be in full understanding of the consequences here.

patriknw:
Yes, I might wait with merging this until Konrad is back.

akka-persistence/src/main/scala/akka/persistence/journal/WriteJournalBase.scala
```diff
@@ -33,11 +32,17 @@ private[akka] trait WriteJournalBase {
...
+    val repr2 = repr.withPayload(adapter.toJournal(payload))
+
+    // IdentityEventAdapter returns "" as manifest and normally the incoming PersistentRepr
+    // doesn't have an assigned manifest, but when WriteMessages is sent directly to the
+    // journal for testing purposes we want to preserve the original manifest instead of
+    // letting IdentityEventAdapter clearing it out.
+    if (adapter == IdentityEventAdapter || adapter.isInstanceOf[NoopWriteEventAdapter])
+      repr2
+    else
+      repr2.withManifest(adapter.manifest(payload))
```
patriknw: I think it can be worth avoiding the allocation from `withPayload` when it's `IdentityEventAdapter`, i.e.

    if (adapter == IdentityEventAdapter || adapter.isInstanceOf[NoopWriteEventAdapter])
      repr
    else
      repr.withPayload(adapter.toJournal(payload)).withManifest(adapter.manifest(payload))

patriknw:
@ktoso ok to merge?

ktoso:
LGTM, great catch and fix. 
Thanks for reporting @Horusiath

Issue #18346, opened 2015-08-27
Document scheduler behaviour when given runnable throws


ktoso:
The scheduler docs do not explain what happens if given a runnable, and the code inside throws, we should document that. This is about the "repeating scheduled task", i.e. if 2nd time throws, will there be a 3rd time (AFAIR no, there won't).

via: http://stackoverflow.com/questions/32255919/akka-scheduler-stops-on-exception-is-it-expected

hepin1989:
how about add some meta data to the scheduling and when a exception throw,invoke a callback?
eg :
schedule*(meta:Option(Any) ...) and when there is an exception,invoke a callback like 
scheduler.exceptionHandler(cause:Throwable,meta:Option[Any])?
and user could register a exception handler to the scheduler,I think this should be useful.

ktoso:
I'd rather want to keep it very simple. We do not encourage the "schedule a runnable" style anyway, it's better to schedule a message to an actor.

hepin1989:
I don't know why even I have add quartz to our project,our team mate keep using akka's default scheduler for game long scheduling,maybe it's too convenient.

ktoso:
Akka's scheduler is definitely not built for long-term scheduling, it's optimised for "a lot of events, frequently" (i.e. the usual actors case - when many want to do something periodically). 

If you want "long term" definitely go for quartz or something else (that's schedule would survive shutting down apps).

hepin1989:
I know that,them just use it for all.:(,I think there should be more developer do it like this.

PR #18446, opened 2015-09-10
=act #18346 Document scheduler task throws

patriknw:


patriknw:
Refs #18346

patriknw:
failure was unrelated https://github.com/akka/akka/issues/17314

patriknw:
@drewhk please review

akka-actor/src/main/scala/akka/actor/Scheduler.scala
```diff
@@ -74,6 +74,9 @@ trait Scheduler {
    * of the function executions). In such cases, the actual execution interval
    * will differ from the interval passed to this method.
    *
+   * If the function throws an exception the repeated scheduling is aborted,
```
drewhk: That's an important information :)
It is logged at error level though, right?

drewhk: this raises the question of who timeouts on the timeouts ;)

drewhk:
LGTM

akka-actor/src/main/scala/akka/actor/Scheduler.scala
patriknw: It is logged. In the test I added:

    [ERROR] [09/16/2015 11:40:12.744] [LightArrayRevolverSchedulerSpec-akka.actor.default-dispatcher-3] [TaskInvocation] TEST (akka.actor.SchedulerSpec$$anonfun$1$$anonfun$apply$mcV$sp$4$$anonfun$apply$mcV$sp$5$$anon$1)

I think it is logged via the `ExecutionContext.reportFailure` and for our dispatchers we publish that Error to the eventStream

ktoso:
LGTM

PR #18448, opened 2015-09-10
!per #18050 Make event adapter lookup more efficient

patriknw:
When looking into "Integrate EventAdapters to the Query side" I found a few other things.

It looks like `adaptFromJournal` is rather inefficient since the `eventAdapters` is a `def` and `adaptersFor` is O(n) with number of configured plugins (and rather costly also for 1 plugin).

The bigger problem is that we can't change this without breaking binary compatibility. At least I have not found a way. Breaking the journal plugins that was built for 2.4.0-RC2 is not nice, but now is better than later.

I'm not sure if this is the right thing to do, or if there is some alternative. Also, do we have to make another RC because of this? Let's discuss on Monday.

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/journal/EventAdapters.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/journal/EventAdapters.scala
@@ -16,8 +16,10 @@ import scala.collection.mutable.ArrayBuffer
 import scala.reflect.ClassTag
 import scala.util.Try
 
-/** INTERNAL API */
-private[akka] class EventAdapters(
+/**
+ * `EventAdapters` serves as a per-journal collection of bound event adapters.
+ */
+class EventAdapters(
   map: ConcurrentHashMap[Class[_], EventAdapter],
   bindings: immutable.Seq[(Class[_], EventAdapter)],
   log: LoggingAdapter) {
@@ -62,7 +64,10 @@ private[akka] object EventAdapters {
   def apply(system: ExtendedActorSystem, config: Config): EventAdapters = {
     val adapters = configToMap(config, "event-adapters")
     val adapterBindings = configToListMap(config, "event-adapter-bindings")
-    apply(system, adapters, adapterBindings)
+    if (adapters.isEmpty && adapterBindings.isEmpty)
+      IdentityEventAdapters
+    else
+      apply(system, adapters, adapterBindings)
   }
 
   private def apply(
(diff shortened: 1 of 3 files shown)


akka-persistence/src/main/scala/akka/persistence/journal/EventAdapters.scala
```diff
@@ -16,8 +16,10 @@ import scala.collection.mutable.ArrayBuffer
 import scala.reflect.ClassTag
 import scala.util.Try
 
-/** INTERNAL API */
-private[akka] class EventAdapters(
+/**
+ * `EventAdapters` serves as a per-journal collection of bound event adapters.
+ */
+class EventAdapters(
```
patriknw: This change is needed to make the public `Persistence(system).adaptersFor` meaningful.

patriknw:
@ktoso please review

akka-persistence/src/main/scala/akka/persistence/journal/EventAdapters.scala
```diff
@@ -62,7 +64,10 @@ private[akka] object EventAdapters {
   def apply(system: ExtendedActorSystem, config: Config): EventAdapters = {
     val adapters = configToMap(config, "event-adapters")
     val adapterBindings = configToListMap(config, "event-adapter-bindings")
-    apply(system, adapters, adapterBindings)
+    if (adapters.isEmpty && adapterBindings.isEmpty)
+      IdentityEventAdapters
+    else
+      apply(system, adapters, adapterBindings)
```
ktoso: OK

akka-persistence/src/main/scala/akka/persistence/journal/WriteJournalBase.scala
```diff
@@ -13,8 +13,8 @@ import akka.persistence.AtomicWrite
 private[akka] trait WriteJournalBase {
   this: Actor ⇒
 
-  lazy val persistence = Persistence(context.system)
-  private def eventAdapters = persistence.adaptersFor(self)
+  val persistence = Persistence(context.system)
+  private val eventAdapters = persistence.adaptersFor(self)
```
ktoso: I think this is Ok. If we'd consider going back to making ad-hoc registrations possible the EventAdapters would need to become mutable I guess?

ktoso: Though that may be a non-issue, since we discussed last time that the ad-hoc have limited usability and we'll most likely not support that...

ktoso:
LGTM AFACIS

drewhk:
I don't know the context, so I don't see what is the source of speed improvement here? Can someone explain?

patriknw:
for every replayed event `adaptFromJournal` is called, and it calls the `def eventAdapters` which calls `adaptersFor`, which does a linear scan of all registered plugins. https://github.com/akka/akka/blob/master/akka-persistence/src/main/scala/akka/persistence/Persistence.scala#L185

The solution is to change the `def` to a `val`.

drewhk:
Ok, thanks. LGTM

Issue #17425, opened 2015-05-08
Possible bug in AbstractBoundedNodeQueue


drewhk:
See user submitted PR: https://github.com/akka/akka/pull/17267

drewhk:
Probably no issue apart from ensuring that the corresponding mailbox is marked as SingleConsumerSemantics. Moving out from 2.3.11 milestone.

PR #18449, opened 2015-09-11
=act #17425 Document NonBlockingBoundedMailbox to be single-consumer

patriknw:


patriknw:
Refs #17425

akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
```diff
@@ -386,7 +386,10 @@ class NodeMessageQueue extends AbstractNodeQueue[Envelope] with MessageQueue wit
   }
 }
 
-//Discards overflowing messages into DeadLetters
+/**
+ * Lock-free bounded non-blocking multiple-producer single-consumer queue.
+ * Discards overflowing messages into DeadLetters.
```
hepin1989: should we mention this(discarding) in the doc?

drewhk:
LGTM

akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
patriknw: done, thx

PR #18453, opened 2015-09-11
=per #18451 Recommend configuring stash capacity when persistence is used

patriknw:


patriknw:
Refs #18451

rkuhn:
Good point! Perhaps also include a documentation snippet that shows how to make that value significantly lower for a specific dispatcher (in case of ClusterSharding `10000*<partition size>` could well be too much).

akka-docs/rst/java/lambda-persistence.rst
```diff
@@ -117,7 +117,14 @@ about successful state changes by publishing events.
...
 the ``persist`` call and the execution(s) of the associated event handler. This also holds for multiple ``persist``
-calls in context of a single command.
+calls in context of a single command. Incoming messages are :ref:`stashed <stash-lambda-java>` until the ``persist``
+is completed. You should be careful to not send more messages to a persistent actor than it can keep up with,
+otherwise the number of stashed messages will grow. It can be wise to protect against `OutOfMemoryError`
+by defining a maximum stash capacity in the mailbox configuration.
+
+.. code-block:: ruby
+
+    akka.actor.default-mailbox.stash-capacity=10000
```
ktoso: Could be written as (to get code block look):

```
by defining a maximum stash capacity in the mailbox configuration::

    akka.actor.default-mailbox.stash-capacity=10000
```

Issue #18459, opened 2015-09-14
ClusterClient doesn't reconnect when server restarts quickly


patriknw:
The failure detection that was introduced in 2.3.13 does not detect a quick server restart.

This is only an issue in 2.3.13, not in master and not in 2.3.12 (which has other issues).

Issue #18440, opened 2015-09-10
failed: ClusterClientSpec


johanandren:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3637/consoleFull

Could be the same as #17735

patriknw:
this might be a duplicate of https://github.com/akka/akka/issues/18155

I'm running it in a repeat job with debug logging to be able to understand what is wrong

ktoso:
One more:

```
[Association failed with [akka.trttl.gremlin.tcp://ClusterClientSpec@a5.moxie:33345]] Caused by: [Connection refused: a5.moxie/10.0.1.235:33345]
[JVM-1] [WARN] [09/14/2015 00:20:14.425] [ClusterClientSpec-akka.actor.default-dispatcher-4] [akka://ClusterClientSpec/deadLetters] received dead letter from Actor[akka://ClusterClientSpec/user/client3#1249827410]: GetContacts
[JVM-1] - must re-establish connection to receptionist after partition *** FAILED ***
[JVM-1]   java.lang.AssertionError: assertion failed: timeout (1 second) during expectMsgClass waiting for class akka.cluster.client.ClusterClientSpec$Reply
[JVM-1]   at scala.Predef$.assert(Predef.scala:165)
[JVM-1]   at akka.testkit.TestKitBase$class.expectMsgClass_internal(TestKit.scala:435)
[JVM-1]   at akka.testkit.TestKitBase$class.expectMsgType(TestKit.scala:417)
[JVM-1]   at akka.testkit.TestKit.expectMsgType(TestKit.scala:735)
[JVM-1]   at akka.cluster.client.ClusterClientSpec$$anonfun$1$$anonfun$apply$mcV$sp$22$$anonfun$apply$mcV$sp$23$$anonfun$apply$mcV$sp$24$$anonfun$apply$mcV$sp$25.apply$mcV$sp(ClusterClientSpec.scala:249)
[JVM-1]   at akka.cluster.client.ClusterClientSpec$$anonfun$1$$anonfun$apply$mcV$sp$22$$anonfun$apply$mcV$sp$23$$anonfun$apply$mcV$sp$24$$anonfun$apply$mcV$sp$25.apply(ClusterClientSpec.scala:247)
[...]

Issue #18155, opened 2015-08-10
failed: ClusterClientSpec


patriknw:
https://jenkins.akka.io:8498/job/akka-nightly/2499/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2761/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2771/consoleFull

patriknw:
Duplicate of #18440

PR #18461, opened 2015-09-14
=con #18459 Detect quick server restart in ClusterClient

patriknw:
- the problem was that the new server replied with ActorIdentity("H", None)
  which is the auto-reply from EmptyLocalActorRef and we treated that as
  valid heartbeat replies
- including test case that fails before the fix

### DIFF
--- a/akka-contrib/src/main/scala/akka/contrib/pattern/ClusterClient.scala
+++ b/akka-contrib/src/main/scala/akka/contrib/pattern/ClusterClient.scala
@@ -200,7 +200,11 @@ class ClusterClient(
         // heartbeat messages for compatibility with 2.3.12
         receptionist ! Identify(Heartbeat)
       }
-    case ActorIdentity(Heartbeat, _) ⇒
+    case ActorIdentity(Heartbeat, None) ⇒
+      // Server replied, but without ref, i.e. new incarnation (new uid) of receptionist,
+      // which may happen if server is restarted. Don't call failureDetector.
+      ()
+    case ActorIdentity(Heartbeat, Some(_)) ⇒
       // heartbeat reply, using Identify/ActorIdentify messages instead
       // of specialized heartbeat messages for compatibility with 2.3.12
       failureDetector.heartbeat()
--- a/akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterClientSpec.scala
+++ b/akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterClientSpec.scala
@@ -8,6 +8,7 @@ import scala.concurrent.duration._
 import com.typesafe.config.ConfigFactory
 import akka.actor.Actor
 import akka.actor.ActorRef
+import akka.actor.ActorSystem
 import akka.actor.Props
 import akka.cluster.Cluster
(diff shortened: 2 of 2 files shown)


patriknw:
Refs #18459

patriknw:
@drewhk please review

akka-contrib/src/main/scala/akka/contrib/pattern/ClusterClient.scala
```diff
@@ -200,7 +200,11 @@ class ClusterClient(
         // heartbeat messages for compatibility with 2.3.12
         receptionist ! Identify(Heartbeat)
       }
-    case ActorIdentity(Heartbeat, _) ⇒
+    case ActorIdentity(Heartbeat, None) ⇒
+      // Server replied, but without ref, i.e. new incarnation (new uid) of receptionist,
+      // which may happen if server is restarted. Don't call failureDetector.
```
drewhk: While probably not the most optimal here (we know there is another server there, which might have a contact there but we will wait until FD trigger), I agree that we should not complicate this anymore.

patriknw: yes, I was also thinking that since 2.4 does not use this ActorIdentity thingies it could be perceived as a regression (slower behavior) when updating to 2.4 later so I prefer to stick to timeout based failure detection

akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterClientSpec.scala
```diff
@@ -244,5 +249,48 @@ class ClusterClientSpec extends MultiNodeSpec(ClusterClientSpec) with STMultiNod
...
+        }
+
+        c ! ClusterClient.Send("/user/service2", "shutdown", localAffinity = true)
+        Thread.sleep(2000) // to ensure that it is sent out before shutting down system
+      }
+
+      runOn(remainingServerRoleNames.toSeq: _*) {
+        system.awaitTermination(20.seconds)
+        // start new system on same port
+        val sys2 = ActorSystem(system.name,
```
drewhk: does it need to be done on all remaining server roles?

drewhk:
LGTM

akka-contrib/src/multi-jvm/scala/akka/contrib/pattern/ClusterClientSpec.scala
patriknw: um, it's actually only one server JVM left in the test, others have been exited in earlier test steps. However, I don't know which one is left here. Yeah, rather complicated, but this works. I can add a comment.

drewhk: Just add a comment please (I realized L254 already asserted that it is a single one), because it is not intuitive.

ktoso:
LGTM

patriknw:
The failure was in ClusterClientSpec, but in the test before the new test. We already have know failures of that https://github.com/akka/akka/issues/18155 https://github.com/akka/akka/issues/18440

patriknw:
I'll add the new test to master

PR #18403, opened 2015-09-04
=sam update dependencies from snapshot to RC2

KadekM:
As I mentioned to @ktoso on gitter...

Makes the examples buildable and runnable.

Although now PersistentView is deprecated (raises warning in ViewExample). Should I also replace it with Persistence Queries ?

PR #18466, opened 2015-09-14
=sam update leveldb dependencies and doc links

patriknw:
(cherry picked from commit 8291bc0aca504851d0619815c98c760a7d458d18)

This is the same as in https://github.com/akka/akka/pull/18403
Let's see if this PR also fails validation.

### DIFF
--- a/akka-samples/akka-sample-persistence-scala/build.sbt
+++ b/akka-samples/akka-sample-persistence-scala/build.sbt
@@ -6,5 +6,7 @@ scalaVersion := "2.11.7"
 
 libraryDependencies ++= Seq(
   "com.typesafe.akka" %% "akka-actor" % "2.4-SNAPSHOT",
-  "com.typesafe.akka" %% "akka-persistence" % "2.4-SNAPSHOT"
+  "com.typesafe.akka" %% "akka-persistence" % "2.4-SNAPSHOT",
+  "org.iq80.leveldb"            % "leveldb"          % "0.7",
+  "org.fusesource.leveldbjni"   % "leveldbjni-all"   % "1.8"
 )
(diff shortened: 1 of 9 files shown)


patriknw:
Alright, it failed in the same way. I have no clue.

    [info] Compiling 4 Scala sources to /localhome/jenkinsakka/workspace/pr-validator-per-commit-jenkins/akka-samples/akka-sample-persistence-scala/target/classes...
    [error] missing or invalid dependency detected while loading class file 'ActorSystem.class'.
    [error] Could not access term typesafe in package com,
    [error] because it (or its dependencies) are missing. Check your build definition for
    [error] missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
    [error] A full rebuild may help if 'ActorSystem.class' was compiled against an incompatible version of com.
    [error] one error found

akka-samples/akka-sample-persistence-scala/build.sbt
```diff
@@ -6,5 +6,7 @@ scalaVersion := "2.11.7"
 
 libraryDependencies ++= Seq(
   "com.typesafe.akka" %% "akka-actor" % "2.4-SNAPSHOT",
-  "com.typesafe.akka" %% "akka-persistence" % "2.4-SNAPSHOT"
```
patriknw: akka-actor was removed when we had the build failure, I have added it again
stange that akka-sample-persistence-java works without that dependency

2m: I remember that sample build transformer does not support transitive dependencies when transforming from `libraryDependencies` to `projectDependencies`. That is why we need to explicitly declare all akka dependencies in samples. In that case `akka-sample-persistence-java` should not compile. But it does. :)

patriknw:
great, so the problem was the transitive dependency that must be explicit in the samples (at least some of them)

Issue #16597, opened 2014-12-22
StreamTCP: idleTimeout - apply to Http


ktoso:
The `idleTimeout` was introduced in: https://github.com/akka/akka/issues/15401 but is never really handled, see: https://github.com/akka/akka/blob/release-2.3-dev/akka-stream/src/main/scala/akka/stream/impl/io/StreamTcpManager.scala#L80 (that underscore is idleTimeout), it's not propagated anywhere, thus code like:

```scala
StreamTcp().bind(localhost, idleTimeout = 1.second)
```

will never close.

ktoso:
While starting on this we had a small discussion where this shall live, so taking the question to github for more feedback:

Should this be implemented as:

1) Like it was in Spray, but moved into akka-io's `TcpConnection`. Spray used a receive timeout to trigger "if we get a timeout message, shut down" basically. API wise we need to perform the same timeout differentiation in IO as we have in streams (`idleTimeout` and `connectTimeout`, not just `timeout` in `Connect` / `Bind`).

2) We could model it as stream element which in general cancels a stream if no activity for some amount of time. Have not explored this yet.

3) Something else I missed?

// cc  @rkuhn @sirthias @jrudolph

drewhk:
Since idletimeout is something that can be useful for things other than TCP or HTTP I would opt for making it a reusable stream element.

sirthias:
+1 for a reusable, not-TCP-specific solution

ktoso:
Cool, I'm on it then.
In case someone has opinions on "this is not a canonical operation" let me know, so we can keep it an importable pattern perhaps?

ktoso:
Sadly forced to park this due to needing to first solve https://github.com/akka/akka/issues/16410 (we need `TimerStage`) which isn't an easy thing to tackle right now.

drewhk:
We have AsyncStage now, so we can add idletimeout is a simple form, when it treats input and output idle timeouts independently.

jrudolph:
I wonder if treating input and output timeouts independently would be useful for request/response protocols. For those it is likely that only one side is currently active and the other one is currently idle. With AsyncStage (haven't look at the api yet) shouldn't it be much easier to build asynchronously communicating stages to implement a bi-directional solution?

drewhk:
[...]

Issue #16410, opened 2014-11-27
Replace (now hidden) TimerTransform


ktoso:
As part of https://github.com/akka/akka/issues/16393 we hid `TimerTransform` (it's the "old style transformer"), we need to replace it with a "new-style" TimerTransformer ( see this PR: https://github.com/akka/akka/pull/16287 for what I mean by "new" )

ktoso:
This should be solved by AsyncStage: https://github.com/akka/akka/issues/16410

hepin1989:
shouldn't  this be replaced by something like TimerStage?shouldn't this be a part of the 1.0 release ?

ktoso:
By `AsyncStage` and yes it is planned to be done.

viktorklang:
How about adding a `defer` method on `AsyncContext`which returns a `Future[Unit]`?

viktorklang:
misclick

ktoso:
I was tinkering with exposing a method like `schedule(when, theAsyncInput)` to `AsyncContext`

viktorklang:
when == delay?

perhaps 'def defer[T](delay: FiniteDuration, value: T): Future[T]'

-- 
Cheers,
√

ktoso:
Yes, `delay`.
Why Future though? We want the input to come in to the `def onAsyncInput(event: Ext, ctx: AsyncContext[Out, Ext]): Directive` method.

viktorklang:
Then you need to produce the Ext _before_ deferring it, but perhaps that is
what one needs all the time?

-- 
Cheers,
√

ktoso:
I think most of the uses would need a "tick" from this method (that was my use case as well), but very good point!

PR #18225, opened 2015-08-14
WIP: draft of AsyncStage with schedule(), to replace TimerTransform

ktoso:
General direction draft (in reference to https://github.com/akka/akka/issues/16410).
I'm curious if you think this direction looks good or not, rather simple to implement the rest for it (though the question remains of "where to get the scheduler from" to keep the Interpreter pure).

PR #18467, opened 2015-09-14
!str #16410 #16597: Removed TimerTransform, added Timeout stages

drewhk:
Introducing timer support on GraphStage to replace TimerTransform, also implementing various timeout stages using the feature.

In the meantime
- pulled up via and viaMat to FlowOps
- added exception handling to the GraphInterpreter
- ported the still relevant part of the TimerTransformer tests
- implemented proper default upstream and downstream completion handlers
- added beforePreStart and afterPostStop as internal hooks on GraphStage


akka-http-core/src/main/scala/akka/http/impl/util/package.scala
```diff
@@ -127,6 +127,66 @@ package object util {
 
 package util {
 
+  import akka.http.scaladsl.model.{ ContentType, HttpEntity }
```
patriknw: why do we place so much in package.scala?

drewhk: What do you mean?

akka-stream/src/main/scala/akka/stream/io/Timeouts.scala
```diff
@@ -0,0 +1,222 @@
...
+
+  /**
+   * If the time between two processed elements *in any direction* exceed the provided timeout, the stream is failed.
+   *
+   * There is a difference between this stage and having two idleTimeout Flows assembled into a BidiStage.
+   * If the timout is configured to be 1 seconds, then this stage will not fail even though there are elements flowing
+   * every second in one direction, but no elements are flowing in the other direction. I.e. this stage considers
+   * the *joint* frequencies of the elements in both directions.
+   */
+  // FIXME: Where should system come from? (Actually we need a Scheduler and ExecutionContext)
```
patriknw: I think timer scheduling is a core facility that a `Materializer` have to provide. Can we add two abstract methods to the `Materializer` ?

    def schedule(initialDelay: FiniteDuration, interval: FiniteDuration, runnable: Runnable)

    def scheduleOnce(delay: FiniteDuration, runnable: Runnable)

My reasoning for not requiring `def scheduler: akka.actor.Scheduler` is that then other materializer implemntations can easier use something else than the Akka Scheduler if they like.

ExecutionContext is already defined in `Materializer`.

drewhk: That is not a bad idea. That way all the stages need is a `Materializer` not a specific subclass. Sold :)

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
```diff
@@ -306,6 +310,125 @@ abstract class AsyncStage[In, Out, Ext]
...
+   * timer does not exist, has previously been canceled or if it was a
+   * single-shot timer that was already triggered.
+   */
+  final def isTimerActive(timerKey: Any): Boolean = timers contains timerKey
+
+  /**
+   * Will be called when the scheduled timer is triggered.
+   * @param timerKey key of the scheduled timer
+   */
+  def onTimer(timerKey: Any, ctx: AsyncContext[Out, Scheduled]): Directive
```
patriknw: `Scheduled` is `private[stream]` but it is exposed in the public method, is that alright?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/TimerTransformerSpec.scala
```diff
@@ -1,133 +0,0 @@
...
-      val seq = receiveWhile(2 seconds) {
-        case t: Tick ⇒ t
-      }
-      seq should have length 5
-      expectNoMsg(1 second)
-    }
-
-  }
-
-}
```
patriknw: are the deleted tests (that are still relevant) covered by new tests?

drewhk: Not yet, hence the DO NOT MERGE marker. First the approach should be decided (detached TimerStage or GraphStage) then I will port the existing tests.

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
patriknw: I simply don't like hiding things in package.scala, why not use ordinary class files?

drewhk: Should I refactor http.impl.util into an object instead? I think there are a couple of package objects in http, so those should be changed as well. I think that should be another ticket and PR though (these are used all around the place).

patriknw: yes, not this PR, that's something for @ktoso

akka-stream-tests/src/test/scala/akka/stream/scaladsl/TimerTransformerSpec.scala
patriknw: I see. Could you make it easier for me by pointing out the API of the two alternatives and describe the the differences (pros/cons).

Comparing the new GroupedWithin with the old TimerTransformer based GroupedWithin it looks rather much more complicated to me.

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
ktoso: I don't mind the package objects as they are in Http currently but if you have feelings about them I'll open a ticket and re-visit this :) https://github.com/akka/akka/issues/18477

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
```diff
@@ -127,6 +127,66 @@ package object util {
...
+  import akka.stream.{ Outlet, Inlet, FlowShape }
+  import scala.concurrent.duration.FiniteDuration
+
+  private[http] class ToStrict(timeout: FiniteDuration, system: ActorSystem, contentType: ContentType)
+    extends GraphStage[FlowShape[ByteString, HttpEntity.Strict]] {
```
ktoso: excellent, no more TimerTransform thanks to this guy then :)

akka-stream/src/main/scala/akka/stream/io/Timeouts.scala
ktoso: I'm all for exposing just `schedule`/`scheduleOnce` instead of anything akka specific, in fact that's how I wanted to impl the AsyncStage scheduling: https://github.com/akka/akka/pull/18225/files#diff-9b5ba5308f1cef1574d15d27aa125b9bR663 :)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/TimerTransformerSpec.scala
drewhk: One is the TimerStage that you see as reimplementing takeWithin, dropWithin and groupedWithin. These are using the OneBoundedInterpreter and the AsyncStage abstraction (which TimerStage is a subclass of). The other approach does not give timer specific helpers yet (but it is possible of course) but uses GraphStage directly, you can see these stages in the `Timeouts` object. They are mostly comparable with the takeWithin/dropWithin implementations. 

The GraphStage approach is more boilerplate, as it is something that assumes generic graph stages, so having a simple flow stage is more characters. The logic I think is simpler though. Supposedly the TimerStage gives more safety features by the static return types, but I am not sure that at this complexity level they actually help. I personally like the GraphStage better, but it looks more complex due to it having more boilerplate. It also have the benefit of not adding yet-another DSL (TimerStage is a new one) just reuses the most generic DSL.

I have not yet tried to implement GroupedWithin as a GraphStage, so I don't have a comparison between those, maybe I can try to build one.

The difference to TimerTransformer is that it handled emitting of multiple elements in a simpler way, but it was not pullable (you cannot emit very long sequences effectively) and fusable. It is also a much more limited abstraction.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -726,3 +729,139 @@ private[akka] object Log {
...
+/**
+ * INTERNAL API
+ */
+private[stream] final case class GroupedWithin[T](n: Int, d: FiniteDuration) extends TimerStage[T, immutable.Seq[T]] {
+  import TimerKeys._
+  private var buf: Vector[T] = Vector.empty
+
+  override def preStart(ctx: AsyncContext[immutable.Seq[T], Scheduled]) = {
+    super.preStart(ctx)
+    schedulePeriodically(GroupedWithinTimerKey, d)
```
ktoso: I think I'd prefer extending `AsyncContext` to be able to schedule in a normal `AsyncStage`, one less concept to pick from when implementing stuff, in order to avoid: *"hmm... timer is async stuff, I know async stage already. I guess it can schedule. Oh it can't... oh there's TimerStage."*

akka-stream/src/main/scala/akka/stream/io/Timeouts.scala
drewhk: Well, almost, but you had to cast down in the end. What @patriknw idea is that it is actually defined on Materializer, which means that no casting is necessary - and forces all Materializers to implement it (or at least think about it). Since timers are quite common in any concurrency related construct, I think it is very reasonable.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: This is why I like the GraphStage better, since you are not forced to use only one callback (onAsync). The problem if you add the thing to AsyncStage itself, then if people forget to call `super.onAsync` then the timer event will be lost.

`preStart` and `postStop` are already a problem of course.

akka-stream/src/main/scala/akka/stream/io/Timeouts.scala
ktoso: Yup, agreed let's make them implement scheduling, or they can cop-out of implementing those explicitly if they want to :+1:

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: Just a sanity check for myself:
- since `schedule` will be moved to `Materializer`, I can use it with AsyncStage - if my stage is rather simple I think I'd use that still.
- it's also usable with the GraphStage

win-win and everybody is happy, right?

drewhk: TimerStage is just a convenience class, you can do the exact same thing with a raw AsyncStage.

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
```diff
@@ -306,6 +310,125 @@ abstract class AsyncStage[In, Out, Ext]
...
+  final case class Timer(id: Int, task: Cancellable)
+}
+
+/**
+ * This is a variant of [[DetachedStage]] that can receive timer events.
+ *
+ * When overriding the preStart() and postStop() methods make sure that you call
+ * the inherited methods as the timer support relies on these.
+ */
+abstract class TimerStage[In, Out] extends AsyncStage[In, Out, TimerStage.Scheduled] {
```
patriknw: I think named timers, automatic cancellation, pre start queueing, and all such things that this `TimerStage` provides are very important for the usability. So if we support scheduling in the GraphStage it should provide the same.

drewhk: I fully agree, if we go the GraphStage way then these will be migrated there. The other benefit of GraphStage will be that it can use different callbacks, i.e. it will be easier to mix async signals and timer signals.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/TimerTransformerSpec.scala
patriknw: Thanks for describing, I think we need the timer DSL sugar as mentioned elsewhere.

drewhk: Yep, that will migrate from TimerStage to GraphStage. That way I can add cleanup and init method also in extra "prepreStart" "postpostStop" methods not accessible for users, i.e. they don't need to call super.

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
patriknw: The current TimerStage is perhaps too difficult to use for the casual user due to the AsyncStage so it's not a good end user api for the 90% cases anyway. I get the impression that GraphStage is more flexible and should therefore be our low level building block, and then we can add something on top of that specially targeted for the simple/limited use cases.

akka-stream/src/main/scala/akka/stream/io/Timeouts.scala
hepin1989: I think that should be very convenient to have them on ```Meterializer```

akka-stream/src/main/scala/akka/stream/io/Timeouts.scala
```diff
@@ -0,0 +1,222 @@
...
+
+      final override protected def onTimerTick(): Unit =
+        if (!initialHasPassed)
+          failStage(new TimeoutException(s"The first element has not yet passed through in $timeout."))
+
+      override def preStart(): Unit = {
+        super.preStart()
+        val callback = getAsyncCallback[Unit]((_) ⇒ onTimerTick())
+        import system.dispatcher
+        timerCancellable = Some(system.scheduler.scheduleOnce(timeout)(callback.invoke(())))
```
hepin1989: not context.scheduler?

drewhk: There is no "context.scheduler". This is not an ActorContext.

hepin1989: I means we need get the scheduler from the materializer later right?

drewhk: Yes, exactly. We will add the necessary methods to the Materializer interface itself.

drewhk:
@ktoso @patriknw This is ready for final review.

drewhk:
eh, well, let's pass the compiler first ;)

akka-stream-tests/src/test/scala/akka/stream/io/TimeoutsSpec.scala
```diff
@@ -0,0 +1,238 @@
+package akka.stream.io
```
ktoso: license header missing

akka-stream/src/main/scala/akka/stream/io/Timeouts.scala
```diff
@@ -0,0 +1,166 @@
...
+object Timeouts {
+
+  /**
+   * If the first element has not passed through this stage before the provided timeout, the stream is failed.
+   */
+  def initalTimeout[T](timeout: FiniteDuration): Flow[T, T, Unit] =
+    Flow.wrap(new InitialTimeout[T](timeout))
+
+  /**
+   * If the completion of the stream does not happen until the provided timeout, the stream is failed.
```
ktoso: `... is failed with an [[TimeoutException]].`?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -726,3 +729,121 @@ private[akka] object Log {
...
+      override def onUpstreamFailure(ex: Throwable): Unit = failStage(ex)
+    })
+
+    final override protected def onTimer(key: Any): Unit = allow = true
+
+    scheduleOnce("DropWithinTimer", timeout)
+  }
+
+  override def toString = "CompletionTimeout"
+}
```
ktoso: `SimpleLinearGraphStage` looks quite nice actually, :+1: 

Do I remember correctly you mentioning that he graph interpreter is slightly slower than the OneBoundedInterpreter?
May we end up wanting to use normal Stages instead of this because of perf, or am I misunderstanding when which interpreter is used? (This is more for my understanding than comment to this PR)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -726,3 +729,121 @@ private[akka] object Log {
...
+      override def onUpstreamFinish(): Unit = completeStage()
+      override def onUpstreamFailure(ex: Throwable): Unit = failStage(ex)
+    })
+
+    final override protected def onTimer(key: Any): Unit = allow = true
+
+    scheduleOnce("DropWithinTimer", timeout)
+  }
+
+  override def toString = "CompletionTimeout"
```
ktoso: typo `override def toString = "DropWithin"`?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -726,3 +729,121 @@ private[akka] object Log {
...
+      override def onUpstreamFailure(ex: Throwable): Unit = failStage(ex)
+    })
+
+    final override protected def onTimer(key: Any): Unit =
+      completeStage()
+
+    scheduleOnce("TakeWithinTimer", timeout)
+  }
+
+  override def toString = "CompletionTimeout"
```
ktoso: typo `override def toString = "TakeWithin"`?

ktoso:
Read through a few times, looks good I think, and `Timeouts` look very good.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: yes, "TakeWithinTimeout".

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: yep, copy pasta

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: 50% performance compared to the OneBounded currently. I have not yet tested though if I remove all Maps (adding ID to ports)

akka-stream/src/main/scala/akka/stream/io/Timeouts.scala
drewhk: good point

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
```diff
@@ -127,6 +127,55 @@ package object util {
...
+
+    val in = Inlet[ByteString]("in")
+    val out = Outlet[HttpEntity.Strict]("out")
+    override val shape = FlowShape(in, out)
+
+    override def createLogic: GraphStageLogic = new GraphStageLogic {
+      var bytes = ByteString.newBuilder
+      private var emptyStream = false
+
+      scheduleOnce("ToStrictTimeoutTimer", timeout)
```
patriknw: nice!

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
```diff
@@ -127,6 +127,55 @@ package object util {
 
 package util {
 
+  import akka.http.scaladsl.model.{ ContentType, HttpEntity }
+  import akka.stream.{ Outlet, Inlet, FlowShape }
+  import scala.concurrent.duration.FiniteDuration
+
+  private[http] class ToStrict(timeout: FiniteDuration, system: ActorSystem, contentType: ContentType)
```
patriknw: I can't see that `system` is used here

akka-stream-tests/src/test/scala/akka/stream/io/TimeoutsSpec.scala
patriknw: still missing

akka-stream-tests/src/test/scala/akka/stream/io/TimeoutsSpec.scala
```diff
@@ -0,0 +1,238 @@
...
+      Await.result(
+        Source(1 to 100).via(Timeouts.initalTimeout(2.seconds)).grouped(200).runWith(Sink.head),
+        3.seconds) should ===(1 to 100)
+    }
+
+    "pass through error unmodified" in assertAllStagesStopped {
+      a[TE] shouldBe thrownBy {
+        Await.result(
+          Source(1 to 100).concat(Source.failed(TE("test")))
+            .via(Timeouts.initalTimeout(2.seconds))
```
patriknw: might want to speed up the test by having a shorter timeout?

akka-stream-tests/src/test/scala/akka/stream/io/TimeoutsSpec.scala
```diff
@@ -0,0 +1,238 @@
...
+      Await.result(
+        Source(1 to 100).via(Timeouts.completionTimeout(2.seconds)).grouped(200).runWith(Sink.head),
+        3.seconds) should ===(1 to 100)
+    }
+
+    "pass through error unmodified" in assertAllStagesStopped {
+      a[TE] shouldBe thrownBy {
+        Await.result(
+          Source(1 to 100).concat(Source.failed(TE("test")))
+            .via(Timeouts.completionTimeout(2.seconds))
```
patriknw: shorter?

akka-stream/src/main/scala/akka/stream/Materializer.scala
```diff
@@ -31,6 +35,24 @@ abstract class Materializer {
...
+
+  /**
+   * Interface for stages that need timer services for their functionality. Schedules a
+   * repeated task with the given interval between invocations.
+   *
+   * @return A [[akka.actor.Cancellable]] that allows cancelling the timer. Cancelling is best effort, if the event
+   *         has been already enqueued it will not have an effect.
+   */
+  def schedulePeriodically(interval: FiniteDuration, task: Runnable): Cancellable
+
```
patriknw: good!

patriknw: by the way, don't we need initialDelay and interval?

ktoso: Hah was about to ask about initialDelay.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -267,7 +271,10 @@ private[stream] final class GraphInterpreter(
   def finish(): Unit = {
     var i = 0
     while (i < logics.length) {
-      if (!isStageCompleted(i)) logics(i).postStop()
+      if (!isStageCompleted(i)) {
+        logics(i).postStop()
```
patriknw: can this throw, and in that case what happens?

akka-stream/src/main/scala/akka/stream/io/Timeouts.scala
```diff
@@ -0,0 +1,166 @@
...
+
+      final override protected def onTimer(key: Any): Unit =
+        if (!initialHasPassed)
+          failStage(new TimeoutException(s"The first element has not yet passed through in $timeout."))
+
+      scheduleOnce("InitialTimeout", timeout)
+    }
+
+    override def toString = "InitialTimeoutTimer"
+  }
```
patriknw: nice!

akka-stream/src/main/scala/akka/stream/io/Timeouts.scala
```diff
@@ -0,0 +1,166 @@
...
+        }
+        override def onUpstreamFinish(): Unit = completeStage()
+        override def onUpstreamFailure(ex: Throwable): Unit = failStage(ex)
+      })
+
+      final override protected def onTimer(key: Any): Unit =
+        if (nextDeadline.isOverdue())
+          failStage(new TimeoutException(s"No elements passed in the last $timeout."))
+
+      schedulePeriodically("IdleTimeoutCheckTimer", interval = timeout / 8)
```
patriknw: do we need to do it that often? perhaps something like

    min(max(timeout / 8, 100.millis), timeout / 2)

patriknw:
LGTM, awesome!

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
drewhk: no longer needed, sorry missed on my radar

akka-stream-tests/src/test/scala/akka/stream/io/TimeoutsSpec.scala
drewhk: this is a new one, but yes :)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: no longer true, will update

akka-stream-tests/src/test/scala/akka/stream/io/TimeoutsSpec.scala
drewhk: It was a tradeoff between flakyness and speed. Do you think 1 second is safe?

akka-stream/src/main/scala/akka/stream/Materializer.scala
drewhk: we haven't had that in the TimerTransformer. I am not sure if we really need it, but of course it is easy to add (and of course also easy to emulate with two timers).

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: good point, these need to be covered

akka-stream/src/main/scala/akka/stream/io/Timeouts.scala
drewhk: / 8 is around 12,5% error which I deemed to be sufficient. I agree that it is a good idea to cap it.

akka-stream-tests/src/test/scala/akka/stream/io/TimeoutsSpec.scala
patriknw: I think I was wrong, this test is quick, it will normally complete with the TE failure immediately, right?

akka-stream/src/main/scala/akka/stream/Materializer.scala
patriknw: I think we started without it in the Source timer, but then someone needed it and we added it, so I was thinking we should add it immediately now.

The two timers is a good idea though for the rare cases when it is needed. I like that, since we have those named timers.

patriknw: BTW how is the Source timer implemented now? Should it make use of these materializer methods?

drewhk: I forgot about that source TBH. I don't even know what it uses, but probably not TimerTransformer snice I removed that ;)

If it is a custom actor, then I will try to replace it with a GraphStage

patriknw: one more thought, let's add the initial delay in the materializer methods, but skip it in the `GraphStage. schedulePeriodically`

drewhk: we can add a separate method, `schedulePeriodicallyWithInitialDelay`? that makes the common case still simple, while the custom case accessible.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
patriknw: ok, then there is at a few other places aslo

akka-stream/src/main/scala/akka/stream/Materializer.scala
patriknw: but the materializer methods are not end user api (or at least not normal end users)

drewhk: No, I meant adding it to the GraphStageLogic, i.e. "overload" just not with the same name

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: I will do these in another PR, these are the missing tests/features on the interpreter I mentioned earlier. I am still working with a half-tested, half-done engine :D (living dangerous)

akka-stream/src/main/scala/akka/stream/Materializer.scala
patriknw: +1

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
patriknw: :+1:

drewhk:
refs #16410 #16597

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
```diff
@@ -127,6 +127,53 @@ package object util {
 
 package util {
 
+  import akka.http.scaladsl.model.{ ContentType, HttpEntity }
+  import akka.stream.{ Outlet, Inlet, FlowShape }
+  import scala.concurrent.duration.FiniteDuration
+
+  private[http] class ToStrict(timeout: FiniteDuration, contentType: ContentType)
```
jrudolph: I've seen this just now. I wonder if we aren't missing a kind of half-the-way abstraction that more closely follows the old `TimerTransform` if we need to double the lines of code in this change for such a simple thing?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -55,28 +54,10 @@ sealed trait HttpEntity extends jm.HttpEntity {
...
-
-        def onTimer(timerKey: Any): immutable.Seq[HttpEntity.Strict] =
-          throw new java.util.concurrent.TimeoutException(
-            s"HttpEntity.toStrict timed out after $timeout while still waiting for outstanding data")
-      }
-
-    // TODO timerTransform is meant to be replaced / rewritten, it's currently private[akka]; See https://github.com/akka/akka/issues/16393
-    dataBytes.via(Flow[ByteString].timerTransform(transformer).named("toStrict")).runWith(Sink.head)
-  }
+  def toStrict(timeout: FiniteDuration)(implicit fm: Materializer): Future[HttpEntity.Strict] =
```
jrudolph: I guess what we need here is a simple generic `aggregateBytes(timeout, maxBytes)` Flow so that this can get

```scala
dataBytes.via(aggreateBytes(timeout, maxBytes)).map(HttpEntity.Strict(contentType, _))
```

I guess the generic `aggregateBytes` flow could even be more generic by allow it to work on every collection that supports `++` and `size`, or a builder that support `append`, `size`, and `result`.

drewhk: yep, good idea! File a ticket for that please.

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
drewhk: Well, the DSL zoo is a bit too wide right now, in fact, we will remove FlexiMerge and FlexiRoute. We can certainly add more friendly mid-tier DSLs later, but these are always missing "one more feature" and I am happy that we have at least a fairly powerful bottom-tier DSL now. I.e. in the future yes, right now, probably no. 

But you can easily add layers in HTTP on top of the GraphStage if needed. For this stage it didn't feel worth it, but if you find nice abstractions we can always migrate them to the main library.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
jrudolph: Maybe let's do it while resolving #18592

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
jrudolph: I see. Thanks for explaining.

PR #18473, opened 2015-09-15
=str: GraphStage improvements

drewhk:
Collection of small things

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/ActorGraphInterpreterSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/ActorGraphInterpreterSpec.scala
@@ -3,17 +3,14 @@
  */
 package akka.stream.impl.fusing
 
-import java.util.concurrent.TimeoutException
-
-import akka.actor.{ ActorSystem, Cancellable, Scheduler }
 import akka.stream._
 import akka.stream.scaladsl._
 import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
-import akka.stream.testkit.{ TestPublisher, AkkaSpec, TestSubscriber }
+import akka.stream.testkit.AkkaSpec
+import akka.stream.testkit.Utils._
 
 import scala.concurrent.Await
 import scala.concurrent.duration._
-import akka.stream.testkit.Utils._
 
 class ActorGraphInterpreterSpec extends AkkaSpec {
   implicit val mat = ActorMaterializer()
@@ -27,6 +24,19 @@ class ActorGraphInterpreterSpec extends AkkaSpec {
         Source(1 to 100).via(identity).grouped(200).runWith(Sink.head),
         3.seconds) should ===(1 to 100)
 
+    }
+
+    "be able to reuse a simple identity graph stage" in assertAllStagesStopped {
+      val identity = new GraphStages.Identity[Int]
+
+      Await.result(
+        Source(1 to 100)
+          .via(identity)
+          .via(identity)
+          .via(identity)
+          .grouped(200)
+          .runWith(Sink.head),
+        3.seconds) should ===(1 to 100)
     }
 
     "be able to interpret a simple bidi stage" in assertAllStagesStopped {
@@ -70,6 +80,52 @@ class ActorGraphInterpreterSpec extends AkkaSpec {
 
     }
 
+    "be able to interpret and resuse a simple bidi stage" in assertAllStagesStopped {
+      val identityBidi = new GraphStage[BidiShape[Int, Int, Int, Int]] {
+        val in1 = Inlet[Int]("in1")
+        val in2 = Inlet[Int]("in2")
+        val out1 = Outlet[Int]("out1")
+        val out2 = Outlet[Int]("out2")
+        val shape = BidiShape(in1, out1, in2, out2)
+
+        override def createLogic: GraphStageLogic = new GraphStageLogic {
+          setHandler(in1, new InHandler {
+            override def onPush(): Unit = push(out1, grab(in1))
+
+            override def onUpstreamFinish(): Unit = complete(out1)
+          })
+
+          setHandler(in2, new InHandler {
+            override def onPush(): Unit = push(out2, grab(in2))
+
+            override def onUpstreamFinish(): Unit = complete(out2)
+          })
+
+          setHandler(out1, new OutHandler {
+            override def onPull(): Unit = pull(in1)
+
+            override def onDownstreamFinish(): Unit = cancel(in1)
+          })
+
+          setHandler(out2, new OutHandler {
+            override def onPull(): Unit = pull(in2)
+
+            override def onDownstreamFinish(): Unit = cancel(in2)
+          })
+        }
+
+        override def toString = "IdentityBidi"
+      }
+
+      val identityBidiF = BidiFlow.wrap(identityBidi)
+      val identity = (identityBidiF atop identityBidiF atop identityBidiF).join(Flow[Int].map { x ⇒ x })
+
+      Await.result(
+        Source(1 to 10).via(identity).grouped(100).runWith(Sink.head),
+        3.seconds) should ===(1 to 10)
+
+    }
+
     "be able to interpret a rotated identity bidi stage" in assertAllStagesStopped {
       // This is a "rotated" identity BidiStage, as it loops back upstream elements
       // to its upstream, and loops back downstream elementd to its downstream.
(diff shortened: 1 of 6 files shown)


drewhk:
Please take a look at this one, too, I will need these changes.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/ActorGraphInterpreterSpec.scala
```diff
@@ -70,6 +80,52 @@ class ActorGraphInterpreterSpec extends AkkaSpec {
 
     }
 
+    "be able to interpret and resuse a simple bidi stage" in assertAllStagesStopped {
```
ktoso: reuse

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/ActorGraphInterpreterSpec.scala
```diff
@@ -70,6 +80,52 @@ class ActorGraphInterpreterSpec extends AkkaSpec {
...
+
+        override def toString = "IdentityBidi"
+      }
+
+      val identityBidiF = BidiFlow.wrap(identityBidi)
+      val identity = (identityBidiF atop identityBidiF atop identityBidiF).join(Flow[Int].map { x ⇒ x })
+
+      Await.result(
+        Source(1 to 10).via(identity).grouped(100).runWith(Sink.head),
+        3.seconds) should ===(1 to 10)
```
patriknw: shouldn't we slowly start using the testkit facilities?

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -48,13 +48,13 @@ private[stream] object GraphInterpreter {
    * A GraphAssembly represents a small stream processing graph to be executed by the interpreter. Instances of this
    * class **must not** be mutated after construction.
    *
-   * The arrays [[ins]] and [[outs]] correspond to the notion of a *connection* in the [[GraphInterpreter]]. Each slot
+   * The arrays ins and outs correspond to the notion of a *connection* in the [[GraphInterpreter]]. Each slot
```
patriknw: could use backtick?

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/ActorGraphInterpreterSpec.scala
drewhk: Why add a probe for this case? I usually like to use the high level API unless I explicitly need to control requests and such.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: ok, that can work.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -55,7 +58,7 @@ abstract class GraphStage[S <: Shape] extends Graph[S, Unit] {
  *    of the enclosing [[GraphStage]]
  *  * Possible mutable state, accessible from the [[InHandler]] and [[OutHandler]] callbacks, but not from anywhere
  *    else (as such access would not be thread-safe)
- *  * The lifecycle hooks [[preStart()]] and [[postStop()]]
+ *  * The lifecycle hooks preStart() and postStop()
```
patriknw: it's not a problem with such links in general, those should work as [[#preStart]] and [[#postStop]]
Otherwise backtick

drewhk: Do these work with Java8 Javadoc?

patriknw:
removing scaladoc links doesn't look like an attractive change

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/ActorGraphInterpreterSpec.scala
patriknw: ok, I thought the Await, grouped, Sink.head dance was nicely covered by the testkit

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
patriknw: yes, most of them work just fine, there are special cases that will not work, but those are rare. Testing with java8 javadoc is not a priority right now, so it's better to have working scaladoc links now, and we can fix the broken once later instead of not having any links at all.

drewhk: @rkuhn asked for the changes, that's why I did them.

drewhk:
See comments on the original PR: https://github.com/akka/akka/pull/18313

drewhk:
I am fine either way, just let's agree on something :)

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
patriknw: Broken links in scaladoc should of course be fixed, `preStart` and `postStop` looks like ordinary methods so I don't understand why that would not work.
Feels like you aggressively removed too many links.

patriknw:
scaladoc is the tool you have at this branch, so let's fix links that scaladoc complains about, and worry about genjavadoc+java8 when we move there

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: Well, I just followed the "please remove links like these, they don’t work and fail the Java 8 javadoc" sentence. You can interpret the "links like these" in various ways :)

drewhk:
Ok, fine by me

drewhk:
What about the rest of the changes? Are they OK?

patriknw:
Yes, i only have objections to changing links to plain text.

drewhk:
Thats ok I revert those.
2015. szept. 15. 20:07 ezt írta ("Patrik Nordwall" <notifications@github.com

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/ActorGraphInterpreterSpec.scala
ktoso: Good catch @patriknw, I think that should be added to the StreamTestKit actually, we have it all over the place in all tests.

drewhk: There is an expectNextn that takes a sequence, but creating the probe itself, then creating a Sink with the probe is not less work or boilerplate. If we really want to make this more succinct, then an `expectSink(expectedSeq)` could work, although the exception somehow needs to be passed to the test thread. 

I don't like to use probes unless they are strictly necessary.

drewhk: also, ===() has nicer error reporting when there is a mismatch.

patriknw: draining to a strict collection would be useful in testkit, i.e. doing the same ss you do here, but in less characters

drewhk: I guess an implicit toStrict is ok for our cases

ktoso: As a matter of fact, I'm almost done with the needed TestKit updates :wink: 

```
Source(1 to 10).via(identity).runWIth(TestSink.probe).toStrict(300.millis) should ===(1 to 10)
```

// needed a quick break from Http, this was fun :smile:

drewhk: Why not just:

    Source(1 to 10).via(identity).shouldEmit(1 to 10)

ktoso: Hm that'd need implicits on Source/Flow, we can do that later I guess - for now it's a bit more utility methods on the TestSink, see here: https://github.com/akka/akka/pull/18503

PR #18475, opened 2015-09-15
=htt http server fails to unbind

2m:
I have found this behavior while researching other issue. It seems that server continues to respond to requests after unbind is called and returned future is completed.

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/ServerSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/ServerSpec.scala
@@ -1,0 +1,58 @@
+package akka.http.scaladsl
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
+import akka.http.scaladsl.server.Directives
+import akka.http.scaladsl.unmarshalling.Unmarshal
+import akka.stream.ActorMaterializer
+import akka.stream.scaladsl._
+import org.scalatest.{ Matchers, WordSpec }
+
+import scala.util._
+import scala.concurrent.{ Await, ExecutionContext, Future }
+import scala.concurrent.duration._
+
+class ServerSpec extends WordSpec with Matchers with Directives {
+
+  "server" should {
+    "not accept any more connections" when {
+      "unbound" in {
+        implicit val sys = ActorSystem("ClientSpec")
+        implicit val mat = ActorMaterializer()
+        import sys.dispatcher
+
+        val binding = Await.result(Http().bindAndHandle(
+          path("test") {
+            complete("ok")
+          }, "127.0.0.1", 0), 1.second)
+
+        val hostname = binding.localAddress.getHostString
+        val port = binding.localAddress.getPort
+
+        val pool = Http().cachedHostConnectionPool[Int](hostname, port)
+        val request = HttpRequest(uri = s"http://$hostname:$port/test")
+
+        for (i ← 1 to 16) {
+          Await.result(clientRequest(request, pool), 2.second) shouldBe "ok"
+        }
+
+        Await.ready(binding.unbind(), 1.second)
+
+        for (i ← 1 to 16) {
+          Await.result(clientRequest(request, pool), 2.second) shouldBe "fail"
+        }
+      }
+    }
+  }
+
+  def clientRequest(request: HttpRequest, pool: Flow[(HttpRequest, Int), (Try[HttpResponse], Int), Http.HostConnectionPool])(implicit mat: ActorMaterializer, ec: ExecutionContext) =
+    Source.single(request -> 42)
+      .via(pool)
+      .mapAsync(4) {
+        case (Success(r), _) ⇒ Unmarshal(r).to[String]
+        case (Failure(_), _) ⇒ Future.successful("fail")
+      }
+      .runWith(Sink.head)
+
+}
+


akka-http-tests/src/test/scala/akka/http/scaladsl/ServerSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+    "not accept any more connections" when {
+      "unbound" in {
+        implicit val sys = ActorSystem("ClientSpec")
+        implicit val mat = ActorMaterializer()
+        import sys.dispatcher
+
+        val binding = Await.result(Http().bindAndHandle(
+          path("test") {
+            complete("ok")
+          }, "127.0.0.1", 0), 1.second)
```
2beaucoup: `TestUtils.temporaryServerHostnameAndPort()`?

ktoso:
FYI @sirthias @jrudolph

jrudolph:
Have you checked if the behavior is the same for TCP? `Http` does not add anything over what TCP offers: https://github.com/akka/akka/blob/release-2.3-dev/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala#L69

jrudolph:
Isn't the behavior as expected? Unbinding does not mean that the existing connections are forced to close and if you use the pool, it will have opened persistent connections that stay open for a while so that they may not be affected by unbinding at all.

ktoso:
Thanks for checking in on this @jrudolph!
I now see what you mean... It may be a bit unexpected for the unknowing user though (I myself thought it's wrong and the ticket originates from a customer interaction actually AFAIR).

Seems like the thing to do to get the expected (by the test) semantics is to `Http().shutdownAllConnectionPools()` - I confirmed this to work right now.

So, for my understanding what's going on here:
- the server unbinds, i.e. is not accepting new connections
- there's existing connections kept alive by the client pool
- those are not killed when the unbind future completes
- we can kill the pool's connection by `Http().shutdownAllConnectionPools()` on the client side, before requesting the next batch
  - this forces the client to open new connections, which there is no server available to handle
  - so it's the scenario which the test expected -> profit.

Pros of current unbind behaviour:
- I'd usually want to service all existing users nicely and then go away - which the current unbind behaviour does.

Downside of current unbind behaviour:
- I don't really know when it's "safe" (as in, "existing clients have closed connections, everyone is happy and I can kill the JVM") to `exit()` the JVM anyway, since the unbind returns too early for that...
  - waiting for the above situation may not be something one might want to do in any case (could wait forever, right?)

@2m could you expand on the use-case where this surfaced?

@jrudolph @sirthias do you guys think implementing a more "terminate(timeout)" like operation on the binding is not useful? It would: unbind, wait for existing connections to be closed, complete the future; (or, if timeout elapsed, kill these connections forcibly).

Curious to hear your thoughts on this, and @2m's use-case explanation. Then let's make a call if this is OK behaviour indeed and we document it or we think about the other "terminate-like" semantics as an addition (?) (if we need it, we need to decide when).

Hope this makes sense :)

sirthias:
In spray we do indeed have a `CloseAll` command. which does what you are looking for: a full graceful shutdown of the complete HTTP layer (server- as well as client-side).

Akka-http currently doesn't have such a feature but I'd second @jrudolph in that it's not the job of `unbind` to do it. `unbind` unbinds the listening port, nothing more, nothing less.
I wouldn't really expect the unbinding to automatically kill existing connections.
Also, unbinding is a server-side-only function.

We could add `gracefulShutdown` method, which mimics what spray does with `CloseAll`:

On the server-side:

1. unbind all listening ports (i.e. stop accepting new server-side connections)
2. stop reading new requests from existing connections
3. wait for requests that are currently being processed to send out their responses (with a timeout)
4. close all existing connections

On the client-side

1. stop accepting new requests into connection pools
2. wait for responses to in-flight requests (with a timeout)
3. close all existing connections

ktoso:
Right @sirthias, sounds like we should provide such a facility. 
I'll work on it after I finish documenting directives today - sounds like a good one that'll require some digging in internals so I'll use that to ask various questions when/if I find something I don't get :)

drewhk:
Yes, TCP unbind does not close existing connections. You can do connections.take(1) and it will work, although take will cancel the connection stream after one connection, but that existing connection will be served.

drewhk:
Maybe we can add an unbindAndCloseAll on top of the unbind command?

ktoso:
Yes, sounds good. It's something best done in the IO layer in any case I think.

sirthias:
@drewhk I'd think that `gracefulShutdown` would also do the unbind. Would you like the `unbindAndCloseAll` to only work on the server-side?

sirthias:
One more thought, we might not need additional timeouts in fact, since all requests (on the server- as well as the client-side) should already have a timeout attached to them. If we simply stop accepting new requests and wait for the defined timeouts to expire we should be good.

drewhk:
>  I'd think that gracefulShutdown would also do the unbind.

Yes, I was talking about the TCP level.

sirthias:
@drewhk I see. However, unfortunately I don't think we'd be able to rely on a TCP-level `unbindAndCloseAll`, since we'd need to snychronize the shutdown with the HTTP-level requests/responses.

Another question that we'd have to answer: What happens with requests/responses that are currently being received when the shutdown is triggered? Since they could, in principle, be infinite we'd have to abort them after a timeout, which wouldn't be that graceful anymore... (but I don't see another way.)

drewhk:
>  I see. However, unfortunately I don't think we'd be able to rely on a TCP-level unbindAndCloseAll, since we'd need to snychronize the shutdown with the HTTP-level requests/responses.

Ok, but TCP might still need a closing version independently of this ticket.

> What happens with requests/responses that are currently being received when the shutdown is triggered?

Maybe require a time limit as an argument in gracefulShutdown and force the shutdown after that time if connections are still present. This is where the TCP close feature might enter the picture though.

ktoso:
> Another question that we'd have to answer: What happens with requests/responses that are currently being received when the shutdown is triggered? Since they could, in principle, be infinite we'd have to abort them after a timeout, which wouldn't be that graceful anymore... (but I don't see another way.)

Any kind of "graceful" to me implies that involved parties "play nice". And if they don't, we'd have the "oh well, we tried - now killing all connections the hard way" timeout - that one I'd give explicitly in the shutdown method. Sounds like the needed parts for this are forming from this discussion, thanks guys!

drewhk:
What is the fate of this PR?

drewhk:
@2m @ktoso @sirthias @jrudolph what should be done with this PR?

2m:
This PR is a reproducer for a failure because of the missing feature. I will close it know as proper tests will be created when #18649 is implemented.

PR #18481, opened 2015-09-15
=ben interpreter benchmarks

2m:
Results in the bench doc. It is interesting that bare OneBounded interpreter is faster than GraphInterpreter. However their performance evens out when they get wrapped into actors as is seen in the bench results (`FlowMapBenchmark` has the same results when using `GraphStage.Identity` or `Flow[Int].map(identity)`)

EDIT: link to [bench results doc](https://docs.google.com/spreadsheets/d/1Sdqx2qj_L6-gpiXerCZHQcmAWzgh9p-pN2N0lBi87Wk/edit?usp=sharing).

### DIFF
--- a/akka-bench-jmh-dev/src/main/scala/akka/stream/InterpreterBenchmark.scala
+++ b/akka-bench-jmh-dev/src/main/scala/akka/stream/InterpreterBenchmark.scala
@@ -1,0 +1,158 @@
+package akka.stream
+
+import akka.event._
+import akka.stream.impl.fusing.{ InterpreterSpecKit, GraphInterpreterSpec, GraphStages, Map => MapStage, OneBoundedInterpreter }
+import akka.stream.impl.fusing.GraphStages.Identity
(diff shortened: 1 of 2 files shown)


drewhk:
Do you have the numbers somewhere?

akka-bench-jmh-dev/src/main/scala/akka/stream/InterpreterBenchmark.scala
```diff
@@ -0,0 +1,158 @@
...
+      sink.requestOne()
+      interpreter.execute(Int.MaxValue)
+    }
+    lock.acquire()
+  }
+
+  @Benchmark
+  @OperationsPerInvocation(100000)
+  def onebounded_interpreter_100k_elements() {
+    val lock = new Lock()
```
drewhk: why not a CountdownLatch?

2m: No real reason. Lock was used in the previous benchmarks.

drewhk:
LGTM

Issue #16948, opened 2015-02-26
FAILED: ClusterSingletonManager


2m:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2058/

Multiple barrier failures starting with:

```
[JVM-3] - must startup 6 node cluster *** FAILED ***
[JVM-3]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class akka.actor.ActorIdentity
[JVM-3]   at scala.Predef$.assert(Predef.scala:165)
[JVM-3]   at akka.testkit.TestKitBase$class.expectMsgClass_internal(TestKit.scala:438)
...
```

patriknw:
https://jenkins.akka.io:8498/job/akka-nightly/2272/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3649/consoleFull

PR #18490, opened 2015-09-16
=clt #16948 Use min retries for singleton leaving scenario

patriknw:
- In 2.4 we derive the number of hand-over/take-over retries from
  the removal margin, but we decided to set that to 0 by default, since
  it is intended for network partition scenarios. maxTakeOverRetries
  became 1. So there must be also be a  min number of retries property.
- The test failed for the leaving scenario because the singleton
  instance was stopped hard without sending the terminationMessage when
  the maxTakeOverRetries was exceeded.

### DIFF
--- a/akka-cluster-tools/src/main/resources/reference.conf
+++ b/akka-cluster-tools/src/main/resources/reference.conf
@@ -141,6 +141,11 @@ akka.cluster.singleton {
   # the previous oldest confirms that the hand over has started or the previous 
   # oldest member is removed from the cluster (+ akka.cluster.down-removal-margin).
   hand-over-retry-interval = 1s
+  
+  # The number of retries are derived from hand-over-retry-interval and
+  # akka.cluster.down-removal-margin (or ClusterSingletonManagerSettings.removalMargin),
+  # but it will never be less than this property.
+  min-number-of-hand-over-retries = 10
 }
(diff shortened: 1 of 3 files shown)


patriknw:
Refs #16948

patriknw:
a failing test that actually found a real bug :-)

akka-cluster-tools/src/main/resources/reference.conf
```diff
@@ -141,6 +141,12 @@ akka.cluster.singleton {
   # the previous oldest confirms that the hand over has started or the previous 
   # oldest member is removed from the cluster (+ akka.cluster.down-removal-margin).
   hand-over-retry-interval = 1s
+  
+  # The number of retries are derived from hand-over-retry-interval and
+  # akka.cluster.down-removal-margin (or ClusterSingletonManagerSettings.removalMargin),
+  # but if removal margin is disabled (or short) the minimum number of retries
```
drewhk: strictly speaking the *minimum* number of retries is always defined by this property.

drewhk:
LGTM

akka-cluster-tools/src/main/resources/reference.conf
patriknw: right, I'll improve

Issue #18489, opened 2015-09-16
parse "off" as MaxLong for akka.http.server.parsing.max-content-length


cubefriendly:
If I want to disable the max-content-length limit for my service, I need to give "MaxValue" as a param in the config.

But MaxValue can be seen as a magic number and makes the configuration less clear.

It would be good if "off" would be parsed to MaxLong and therefore make it explicit in the config

ktoso:
Looking at the config option now, it does support the bytes syntax, i.e. `2G` or `200M`.
So instead of a MaxValue one can give it "10G" which or some other very large number which in turn can be dictated by your requirements. I'm pretty sure your service definitely should have some upper limit of file uploads it can take, I mean, even YouTube restricts those (at least it did a while ago :-)).

With file uploads one can also make the limit in code, maybe logic dependent ("authed user" => 10G, others => 1G): http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0/scala/stream-cookbook.html#Limit_the_number_of_bytes_passing_through_a_stream_of_ByteStrings

Before we allow this "off" style setting, I'd like to re-visit this question again, given that the value can be given in a nice format and not just a very big long. I'd rather not allow disabling it, since then we have to, for consistency, do the same for max-chunk-size and other settings, which will encourage people to disable it altogether everywhere - which is a bad thing to do.

// cc @jrudolph @sirthias

ktoso:
@drewhk provided a nice hint, that we should just enable this for these two settings:

```
max-content-length         = infinite
max-chunk-size             = 1m
```

Also, I'll use infinite instead of `off` as it better relays the meaning and is used in such places already.

cubefriendly:
Having infinite instead of off is good for me. 
I understand your concern and I am 100% with you that it does not make sense to disable the limit by default. But it should be possible to do so.

ktoso:
Cool, figured out the way to get this solved, here's the PR: https://github.com/akka/akka/pull/18491

cubefriendly:
looks good!! thank you

PR #18491, opened 2015-09-16
=htc #18489 allow `infinite` value in max-content-length

ktoso:
Resolves https://github.com/akka/akka/issues/18489

### DIFF
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -196,9 +196,16 @@ akka.http {
     max-header-name-length     = 64
     max-header-value-length    = 8k
     max-header-count           = 64
-    max-content-length         = 8m
     max-chunk-ext-length       = 256
     max-chunk-size             = 1m
+
+    # Maximum content length which should not be exceeded by incoming HttpRequests.
+    # For file uploads which use the entityBytes Source of an incoming HttpRequest it is safe to
+    # set this to a very high value (or to `infinite` if feeling very adventurous) as the streaming
+    # upload will be back-pressured properly by Akka Streams.
+    # Please note however that this setting is a global property, and is applied to all incoming requests,
+    # not only file uploads consumed in a streaming fashion, so pick this limit wisely.
+    max-content-length         = 8m
 
     # Sets the strictness mode for parsing request target URIs.
     # The following values are defined:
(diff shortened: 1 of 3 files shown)


akka-http-core/src/main/scala/akka/http/impl/util/EnhancedConfig.scala
```diff
@@ -39,8 +39,8 @@ private[http] class EnhancedConfig(val underlying: Config) extends AnyVal {
     case x          ⇒ getIntBytes(path)
   }
 
-  def getPossiblyInfiniteLongBytes(path: String): Long = underlying.getString(path) match {
+  def getPossiblyInfiniteBytes(path: String): Long = underlying.getString(path) match {
     case "infinite" ⇒ Long.MaxValue
-    case x          ⇒ getIntBytes(path)
```
ktoso: actually had a bug in here, it would try to parse as Int, yet we need Long. It wasn't used in any place though :)

akka-http-core/src/main/resources/reference.conf
```diff
@@ -196,10 +196,17 @@ akka.http {
...
     max-chunk-ext-length       = 256
     max-chunk-size             = 1m
 
+    # Maximum content length which should not be exceeded by incoming HttpRequests.
+    # For file uploads which use the entityBytes Source of an incoming HttpRequest it is safe to
+    # set this to a very high value (or to `infinite` if feeling very adventurous) as the streaming
+    # upload will be back-pressured properly by Akka Streams.
+    # Please note however that this setting is a global property, and is applied to all incoming requests,
+    # not only file uploads consumed in a streaming fashion, so pick this limit wisely.
+    max-content-length         = 8m
```
patriknw: good description

patriknw:
LGTM

drewhk:
Is it possible to provide a content limit directive?

ktoso:
@drewhk good idea. So the Parser gets shallow copied for each request so we could set a size for a given request on the copied one... Gotta look into it deeper, would you mind opening a separate ticket for it?

drewhk:
No, no, that's not needed. All you need is to attach a limiter Flow to the entity bytes flow. The other parser settings should not be controlled from a directive.

drewhk:
I don't know how directives work, but might be possible.

ktoso:
Ok, so you mean on the stream level. Probably good enough. Separate ticket in any case I think :)

jrudolph:
LGTM

ktoso:
Failure was unrelated.

Issue #18406, opened 2015-09-07
Sink.source


jroper:
If you have an API that takes a `Sink` to consume a stream, and another that takes a `Source` to produce a stream, and you want the stream from the first API to be connected to the stream in the second API, the only way to currently do that in Akka streams is using `Sink.publisher`, and then mapping the materialized value to convert the publisher to a `Source`, and then use it.

There's nothing too difficult about doing this, except that is a major optimisation could happen if there was a `Sink.source` method that returned `Sink[T, Source[T, Unit]]`, since when such a source was run, materialization could be made a noop that simply returned the passed in `Source`.  Now that may not sound interesting, but it does get interesting if you map the materialized value, at that point, you'll be able to materialize the source to the API that accepts the source.

A concrete use case of this is if you have `Sink` to consume an HTTP request body (as you do in Play), and you want to proxy that request to another location through an HTTP client, which takes a `Source` for the request body.  Essentially your code could look like this:

```scala
def myAction = EssentialAction { req => 
  Sink.source[ByteString].mapMaterializedValue { source =>
    ws.url("http://example.org").post(source).map { _ =>
      Ok("Successfully proxied request!")
    }
  }
}
```

drewhk:
Hmm, interesting. We see more and more creative uses of "mapMaterializedValue" :) 
The only thing I would modify here is to not put this on Sink/Source but on a different class not to confuse beginners. Or find a better name for this funcitonality (although I would prefer now not adding to Sink/Source but opening new objects for this purpose).

drewhk:
> There's nothing too difficult about doing this, except that is a major optimisation could happen if there was a Sink.source method that returned Sink[T, Source[T, Unit]], since when such a source was run, materialization could be made a noop that simply returned the passed in Source

Um, but what you get is a Sink, which means that it will need a Publisher anyway. There is not much optimization that can happen here compared to the original solution you use.

rkuhn:
This ticket is invalid as per the discussion in https://github.com/akka/akka/pull/18494/files#r41512546.

drewhk:
[...]

PR #18494, opened 2015-09-16
+str #18406 Sink.source

agolubev:
Add Sink.pipedSink that materializes with Source to be used in API calls. Ref #18406


ktoso:
OK TO TEST

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -220,4 +220,11 @@ object Sink extends SinkApply {
     new Sink(new AcknowledgeSink(bufferSize, DefaultAttributes.acknowledgeSink, shape("AcknowledgeSink"), timeout))
   }
 
+  /**
+   * Crates a `Sink` that is materialized into `Source` to connect with another flow.
+   * This sink gets backpressure from second flow and does not have internal buffer.
+   */
+  def pipedSink[T](): Sink[T, Source[T, Unit]] =
+    new Sink(new PipedSink(DefaultAttributes.pipedSink, shape("PipedSink")))
```
drewhk: Why not just

    Sink.publisher.mapMaterialized(Source(_)).named("PipedSink")

It does the exact same, without an extra class.

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -247,3 +250,55 @@ private[akka] final class AcknowledgeSink[In](bufferSize: Int, val attributes: A
     new AcknowledgeSink[In](bufferSize, attr, amendShape(attr), timeout)
   override def toString: String = "AcknowledgeSink"
 }
+
```
drewhk: I am not sure the 3 classes below are needed, there is a one-liner that does the same using existing API as far as I understand.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
agolubev: ahh, that mapMaterialized magic again. Yeah, makes sense. Will update.

agolubev:
@drewhk, now it's super short solution =)

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -220,4 +220,11 @@ object Sink extends SinkApply {
     new Sink(new AcknowledgeSink(bufferSize, DefaultAttributes.acknowledgeSink, shape("AcknowledgeSink"), timeout))
   }
 
+  /**
+   * Crates a `Sink` that is materialized into `Source` to connect with another flow.
+   * This sink gets backpressure from second flow and does not have internal buffer.
+   */
+  def pipedSink[T](): Sink[T, Source[T, Unit]] =
+    Sink.publisher[T].mapMaterializedValue(Source(_)).named("PipedSink")
```
ktoso: In general good, though it may be worth not dropping down to the Publisher level, as it will be hard to optimise this away during materialization, won't it? I.e. When I'd like to fuse this "sink-source" with the other stages.

ktoso: I'll defer to @drewhk for the decision if we want it in this form or not I guess, he knows more about what we can do with fusing..

agolubev: Ok, let's see what @drewhk will say

drewhk: @ktoso you have no other choice, since after the materialization of this Sink the upper part of the stream will be already running. At that point there is no fusing possible.

ktoso: I see, thanks Endre. LGTM then

drewhk:
@rkuhn, any objections against merging this?

rkuhn:
I’m not sure what exactly this solves, but it will need to present a reusable Source (meaning: it must use fanoutPublisher).

drewhk:
I am not sure, that could lead to silent corruption, while the current approach fails if you want to materialize it multiple times.

drewhk:
For what it solves, see the original ticket by @jroper

rkuhn:
Neither of which sounds desirable, so why exactly are we adding this? The ticket that is linked to does not describe any actual benefits, AFAICS you rebutted James’ claim. `Sink.source[T]` should just be replaced with `Flow[T].….to(…)` in order to do forward chaining of operations where a Sink is needed, unless I am missing something.

drewhk:
I didn't rebutt anything as far as I know. What I rebutted is that a performance optimization is possible.

rkuhn:
Well, that optimization was the only point that I could discern, what am I missing?

drewhk:
I though it was about convenience. Let's hear @jroper

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
jroper: Why does the upper part of the stream have to be running after materialization of the sink?  Why can't materialization say "oh, I don't actually have a sink yet, so there's no point in starting the source yet, so I won't materialize it"?  I don't see why it needs to run the upstream.

jroper:
I guess the question is this: Should an end user, when using nothing but Akka streams, ever need to create a publisher or subscriber?  Currently, to implement my use case, they must, there is no other way to do it.  Since reactive streams APIs are designed to be an SPI for integrating different reactive streams implementations, it seems that if to do something in Akka streams you need to go through reactive streams, that that indicates a missing feature in Akka streams.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
rkuhn: This question is so far from how things are working that I have difficulty understanding it: the sole purpose of materialization is to turn a stream description into a running stream engine. Saying “well, I don’t feel like actually doing my job in this particular case” does not sound appealing to me.

rkuhn:
Currently there is exactly one case in which users of Akka Streams have to use a Publisher, and that is when formulating a dynamic topology, i.e. when at materialization time the shape of the stream is not yet fully known. This is an extremely rare case because when you write a program in general you do have an idea of what it is going to do.

Examples of dynamic behavior that are not dynamic stream topologies include the chat room: here the layout is actually fixed, it is just new participants that get added or removed. This can be implemented using a `Source.actorRef` for injecting new input and output streams, no Publishers required.

The question then is: shall this dynamic layout capability have its own representation in the DSL layer? If yes, then it needs to be well-behaved (i.e. use fanoutPublisher) as mentioned above. There are drawbacks of using dynamic layouts, though, which @drewhk meant when saying that such usage may lead to the loss of elements (one leg starts consuming while the other is not yet attached); these drawbacks are inherent to the feature that is modeled.

My personal opinion is that we should not add sugar for an advanced feature that does not live up to the standards maintained by the rest of the DSL, but I am open to be convinced otherwise.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
drewhk: > Why does the upper part of the stream have to be running after materialization of the sink?

Because materialization is about running. It is even called `run()`. If it would not start the upper part of the stream that would mean the introduction of *another* lifted representation layer, and I am already slightly annoyed with having one.

> Why can't materialization say "oh, I don't actually have a sink yet, so there's no point in starting the source yet, so I won't materialize it"? I don't see why it needs to run the upstream.

Because the materializer only deals with Publishers and Subscribers at that point and changing this would lead to unnecessary complications and serious performance implications, given that it would result in double/triple indirections in various places.

The streams philosophy (as the API is currently) favors encoding dependencies in your modules directly, rather than in function arguments. I.e. if your code needs a Source, then your code should not accept a Source as a parameter to a function, but return a Sink instead. While this ideal cannot be met in 100% of the cases, in general the unit of modularity is an Akka Streams component, not a function. Also, this prefers mostly bottom-up organization (although top down is also possible).

jroper: > I.e. if your code needs a Source, then your code should not accept a Source as a parameter to a function, but return a Sink instead.

Either that is wrong, or Akka HTTP is going against the flow of how Akka streams is meant to be used, since Akka Streams accepts Sources as parameters all over the place, for client request body streaming, for server response body streaming, in WebSockets, everywhere.  I just don't think the ideals being expressed here are at all consistent with the reality of providing a usable API.

jroper: In fact, we had that exact philosophy on Play - APIs should return iteratees, not accept enumerators.  It's great in theory, in practice, it breaks down when you need to do simple plumbing - and it turns out that when dealing with IO, most people just want to do simple plumbing.  So for a number of releases, due to our ideals, we were making life difficult for most of our users.  In the end we introduced new APIs that broke these ideals, but made these common plumbing use cases very straight forward, and an interesting thing about that is that since then, there have been much fewer questions about iteratees/enumerators on the mailing list, we made things that should have been easy, easy, and people found it easy.

drewhk: For those cases when its necessary, the pre-materialzation trick works. This is why this PR was created. The difference is that you wanted to detach a graph into two parts which still share the same materialization which is invalid.

drewhk: @jroper so you are proposing of getting rid of the explicit materialization step?

rkuhn: Guys, I think you might be partially talking past each other. Perhaps it helps to add a completely different viewpoint: on the topic of whether to accept a Source or provide a Sink there is no universally preferred answer. Whoever provides the API needs to make a choice about which party shall be in control of the materialization (i.e. the setting in motion of things). For Akka HTTP this decision was pretty simple in that it is much more difficult to do the wrong thing when the API demands a Source—for one it is impossible to forget to materialize a provided Sink (disregarding the other technical difficulties that this alternative would pose).

In a sense this decision is very similar to toolkit vs. framework: a toolkit will provide tools to be assembled by the user, a framework will require the user to provide the customizable pieces and build the application from those. In this sense Akka HTTP is more framework than toolkit—to a degree.

In any case I think this discussion has long since strayed from its context (the PR) and I do not see any reason to change my conclusion on the code itself.

drewhk: > The streams philosophy (as the API is currently) **favors** encoding dependencies in your modules directly

Emphasis mine, I never said that you cannot use it any other way.

> **While this ideal cannot be met in 100% of the cases**, in general the unit of modularity ... 

Emphasis mine, I never said that it is the only way. Nevertheless, this is the more natural way given the API, since you cannot split up a Flow to a Source and Sink if a pair of functions require those. 

You cannot cut up a Flow (like with using the Sink.source here) without actually materializing one half of it. That is what I said, please read it again. there is no "talking past" on my side. We provided the convenience method for prematerializing stuff, but what I cannot do is to throw out the current materializer and provide means for detaching part of the graph in different DSL elements but still materializing together, and still maintain reusability. These are conflicting goals.

rkuhn: @drewhk The talk about “removing materialization” is what I consider “talking past each other”, nobody is suggesting that (and if they were then I’d provide a host of good arguments against such an endeavor). Another thing I don’t understand is the “convenience method for prematerializing stuff”, we should definitely not have such a thing. Separate materialization is only needed under rare circumstances and shall not be encouraged by the API.

drewhk: I guess you sort it out guys, the PR is here, by our fine contributor @agolubev and it satisfies the given signature in the original ticket. What it does not do is to cut up a graph into two independently materializable, still jointly materializing parts.

rkuhn:
Very sorry about this confusion, @agolubev, but the ticket you solved (and very well, I might add!) turned out to be invalid. This should not happen, but unfortunately it did, please accept my apologies!

agolubev:
@rkuhn, no problem. Discussion above touched interesting topics that are good to know.

rkuhn:
Thanks for your understanding!

Issue #18104, opened 2015-07-31
References to code samples are broken in akka-stream-and-http documentation


cppexpert:
References to code samples are broken in akka-stream-and-http documentation.

For example in http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0/scala/http/routing-dsl/directives/parameter-directives/parameters.html#parameters it renders

```
Required parameter
... includecode2:: ../../../../code/docs/http/scaladsl/server/directives/ParameterDirectivesExamplesSpec.scala
snippet:	required-1
```

ktoso:
Thanks for reporting and sorry for the trouble, we'll fix that soon!

// cc @sirthias @jrudolph

cppexpert:
Guys, any progress on this ? It might be easy fix in some config I suspect.

ktoso:
Not yet, sorry – vacation season struck our throughput, catching up with tickets now.

ktoso:
Resolved by #18495 and listed missing docs: https://github.com/akka/akka/issues/18496 will work on them today.

cppexpert:
Guys, it's still broken.

Sample page
http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0/scala/http/routing-dsl/directives/parameter-directives/parameters.html#parameters

ktoso:
Because there has not been a release, docs are generally published immutably according to their version number - fix will come in 1.1 (which is rather soon, no dates yet though)

ktoso:
The fix here we could somewhat backport, however it would not be complete still, because of the missing things I found: https://github.com/akka/akka/issues/18496

// cc @rkuhn in case you think we should fix the some cases (it's a search-replace) and release 1.0 again. The above linked things though (the ticket) will still be missing

rkuhn:
We’ll release 1.1 soon, don’t spend time on back-porting: streams are experimental for a reason.

PR #18495, opened 2015-09-16
=doc #18104 fixes typo which made includecode2 not work in many places

ktoso:
Note that some of the files don't exist, fixing this very soon...

Resolves https://github.com/akka/akka/issues/18104

// cc @jrudolph @sirthias (unless I missed something, the `... includecode2` does not work, and `..`does, as is expected)


ktoso:
AFAICS no reason not to merge this, going ahead.

PR #18502, opened 2015-09-17
=rem Improve the TimeoutReason error message

patriknw:


patriknw:
@drewhk please review, anything else that should be included in the err message?

akka-remote/src/main/scala/akka/remote/transport/AkkaProtocolTransport.scala
```diff
@@ -463,7 +463,8 @@ private[transport] class ProtocolStateActor(initialData: InitialProtocolStateDat
     } else {
       // send disassociate just to be sure
       sendDisassociate(wrappedHandle, Unknown)
-      stop(FSM.Failure(TimeoutReason("No response from remote. Handshake timed out or transport failure detector triggered.")))
+      stop(FSM.Failure(TimeoutReason(s"No response from remote when $stateName. " +
```
drewhk: Does the stateName really help users? Maybe put it at the end of the sentence in a parenthesis:

    (internal state was $stateName)

To indicate it is DEBUG info for us mostly

drewhk:
LGTM apart from the small nitpick

akka-remote/src/main/scala/akka/remote/transport/AkkaProtocolTransport.scala
patriknw: good point

ktoso:
LGTM

PR #18514, opened 2015-09-18
simple experiment to show the effect of array vs map in GraphInterpreter

drewhk:
**DO NOT MERGE**
_This is an experiment_

This demonstrates that the GraphInterpreter is actually competitive with the OneBoundedInterpreter.

### DIFF
--- a/akka-bench-jmh-dev/src/main/scala/akka/stream/InterpreterBenchmark.scala
+++ b/akka-bench-jmh-dev/src/main/scala/akka/stream/InterpreterBenchmark.scala
@@ -54,7 +54,10 @@ class InterpreterBenchmark {
     val lock = new Lock()
     lock.acquire()
     val sink = OneBoundedDataSink(data100k.size, lock)
-    val ops = Vector.fill(numberOfIds)(MapStage(identity[Int], Supervision.stoppingDecider))
+    val ops = Vector.fill(numberOfIds)(new PushPullStage[Int, Int] {
+      override def onPull(ctx: _root_.akka.stream.stage.Context[Int]) = ctx.pull()
+      override def onPush(elem: Int, ctx: _root_.akka.stream.stage.Context[Int]) = ctx.push(elem)
+    })
     val interpreter = new OneBoundedInterpreter(OneBoundedDataSource(data100k) +: ops :+ sink,
       (op, ctx, event) ⇒ (),
       Logging(NoopBus, classOf[InterpreterBenchmark]),
--- a/akka-stream/src/main/scala/akka/stream/Shape.scala
+++ b/akka-stream/src/main/scala/akka/stream/Shape.scala
@@ -17,6 +17,11 @@ import scala.collection.JavaConverters._
 sealed abstract class InPort { self: Inlet[_] ⇒
   final override def hashCode: Int = System.identityHashCode(this)
   final override def equals(that: Any): Boolean = this eq that.asInstanceOf[AnyRef]
+
+  /**
+   * INTERNAL API
+   */
+  private[stream] var id: Int = 0
 }
 /**
  * An output port of a StreamLayout.Module. This type logically belongs
@@ -27,6 +32,11 @@ sealed abstract class InPort { self: Inlet[_] ⇒
 sealed abstract class OutPort { self: Outlet[_] ⇒
   final override def hashCode: Int = System.identityHashCode(this)
   final override def equals(that: Any): Boolean = this eq that.asInstanceOf[AnyRef]
+
+  /**
+   * INTERNAL API
+   */
+  private[stream] var id: Int = 0
 }
 
 /**
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
@@ -89,6 +89,10 @@ private[stream] object GraphInterpreter {
     def materialize(interpreter: GraphInterpreter): (Array[InHandler], Array[OutHandler], Array[GraphStageLogic]) = {
       val logics = Array.ofDim[GraphStageLogic](stages.length)
       for (i ← stages.indices) {
+        // Must come first
+        stages(i).asInstanceOf[GraphStage[Shape]].shape.inlets.zipWithIndex.foreach { case (inlet, idx) ⇒ inlet.id = idx }
+        stages(i).asInstanceOf[GraphStage[Shape]].shape.outlets.zipWithIndex.foreach { case (outlet, idx) ⇒ outlet.id = idx }
+
         logics(i) = stages(i).createLogic
         logics(i).interpreter = interpreter
       }
@@ -98,12 +102,12 @@ private[stream] object GraphInterpreter {
 
       for (i ← 0 until connectionCount) {
         if (ins(i) ne null) {
-          inHandlers(i) = logics(inOwners(i)).inHandlers(ins(i))
-          logics(inOwners(i)).inToConn += ins(i) -> i
+          inHandlers(i) = logics(inOwners(i)).inHandlers(ins(i).id)
+          logics(inOwners(i)).inToConn(ins(i).id) = i
         }
         if (outs(i) ne null) {
-          outHandlers(i) = logics(outOwners(i)).outHandlers(outs(i))
-          logics(outOwners(i)).outToConn += outs(i) -> i
+          outHandlers(i) = logics(outOwners(i)).outHandlers(outs(i).id)
+          logics(outOwners(i)).outToConn(outs(i).id) = i
         }
       }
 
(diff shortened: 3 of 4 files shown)


akka-bench-jmh-dev/src/main/scala/akka/stream/InterpreterBenchmark.scala
```diff
@@ -54,7 +54,10 @@ class InterpreterBenchmark {
     val lock = new Lock()
     lock.acquire()
     val sink = OneBoundedDataSink(data100k.size, lock)
-    val ops = Vector.fill(numberOfIds)(MapStage(identity[Int], Supervision.stoppingDecider))
+    val ops = Vector.fill(numberOfIds)(new PushPullStage[Int, Int] {
```
drewhk: having a PushPullStage instead of a PushStage disables the jumpBack table which speeds up things in this case, but not in others. I wanted to see the difference between the two interpreters when this optimization is not applicable.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -89,6 +89,10 @@ private[stream] object GraphInterpreter {
     def materialize(interpreter: GraphInterpreter): (Array[InHandler], Array[OutHandler], Array[GraphStageLogic]) = {
       val logics = Array.ofDim[GraphStageLogic](stages.length)
       for (i ← stages.indices) {
+        // Must come first
+        stages(i).asInstanceOf[GraphStage[Shape]].shape.inlets.zipWithIndex.foreach { case (inlet, idx) ⇒ inlet.id = idx }
```
drewhk: bit hacky, should be a while loop for performance

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -75,20 +75,20 @@ abstract class GraphStageLogic {
   /**
    * INTERNAL API
    */
-  private[stream] var inHandlers = scala.collection.Map.empty[Inlet[_], InHandler]
+  private[stream] var inHandlers = Array.ofDim[InHandler](16) // FIXME: proper size
```
drewhk: these should be sized accordingly to the Shape of course. This was just a quick hack to get this working.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -75,20 +75,20 @@ abstract class GraphStageLogic {
...
   /**
    * INTERNAL API
    */
-  private[stream] var outHandlers = scala.collection.Map.empty[Outlet[_], OutHandler]
+  private[stream] var outHandlers = Array.ofDim[OutHandler](16)
```
drewhk: these should be sized accordingly to the Shape of course. This was just a quick hack to get this working.

akka-stream/src/main/scala/akka/stream/Shape.scala
```diff
@@ -17,6 +17,11 @@ import scala.collection.JavaConverters._
 sealed abstract class InPort { self: Inlet[_] ⇒
   final override def hashCode: Int = System.identityHashCode(this)
   final override def equals(that: Any): Boolean = this eq that.asInstanceOf[AnyRef]
+
+  /**
+   * INTERNAL API
+   */
+  private[stream] var id: Int = 0
```
drewhk: This is the ugly change that is needed to make this work

akka-stream/src/main/scala/akka/stream/Shape.scala
```diff
@@ -27,6 +32,11 @@ sealed abstract class InPort { self: Inlet[_] ⇒
 sealed abstract class OutPort { self: Outlet[_] ⇒
   final override def hashCode: Int = System.identityHashCode(this)
   final override def equals(that: Any): Boolean = this eq that.asInstanceOf[AnyRef]
+
+  /**
+   * INTERNAL API
+   */
+  private[stream] var id: Int = 0
```
drewhk: This is the ugly change that is needed to make this work

akka-stream/src/main/scala/akka/stream/Shape.scala
```diff
@@ -17,6 +17,11 @@ import scala.collection.JavaConverters._
 sealed abstract class InPort { self: Inlet[_] ⇒
   final override def hashCode: Int = System.identityHashCode(this)
```
patriknw: perhaps delegate to `super.hashCode` instead (i.e. Object.hashCode)?

drewhk:
My results for these:

    Benchmark                                                  (numberOfIds)   Mode  Cnt      Score     Error   Units
    InterpreterBenchmark.graph_interpreter_100k_elements                   1  thrpt   30  13006,880 ?  40,293  ops/ms
    InterpreterBenchmark.graph_interpreter_100k_elements                   5  thrpt   30   4709,305 ?   9,150  ops/ms
    InterpreterBenchmark.graph_interpreter_100k_elements                  10  thrpt   30   2585,515 ?   8,334  ops/ms
    InterpreterBenchmark.onebounded_interpreter_100k_elements              1  thrpt   30   9264,221 ? 680,750  ops/ms
    InterpreterBenchmark.onebounded_interpreter_100k_elements              5  thrpt   30   4204,527 ?  24,592  ops/ms
    InterpreterBenchmark.onebounded_interpreter_100k_elements             10  thrpt   30   2312,714 ?  42,599  ops/ms

I.e. the GraphInterpreter is as fast, or faster than the OneBounded, also has less deviance/error (more predictable throughput).

My machine:
Intel i7-1930K, 3.4GHz, 24GB RAM

drewhk:
Also, turned out that Map was creating `Option`instances like mad, even though the `apply()` method was used for retreival which I expected to not use an `Option` (but being unsafe as a tradeoff).

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -98,12 +102,12 @@ private[stream] object GraphInterpreter {
 
       for (i ← 0 until connectionCount) {
```
patriknw: replace with while loop

akka-stream/src/main/scala/akka/stream/Shape.scala
drewhk: yeah, that is a good idea. The point was to seal-off the method.

ktoso:
Great results, seems like we'll be able to use just this interpreter :)

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: agreed. This should be not noticable though in the throughput test, since this is initialization code. We probably will need a "latency" benchmark as well, which measures the time from init to the first element downstream.

patriknw:
I like how this is shaping up, great work!

drewhk:
Will see, this is just a single experiment. Also, OneBounded can be still faster if jumpBackTable is enabled, and there is a long enough chain of PushStages. That optimization is very hard to do with the GraphInterpreter (although not impossible, in fact it can be done at the GraphAssembly level, i.e. statically)

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
patriknw: yes, not for this test but we need materialization to be fast and ranges are extremely slow (see for example http://www.scalablescala.com/roller/scala/entry/why_is_using_for_foreach)

drewhk: I will phase out all of these (I have experienced the "performance" of ranges before, also in the other interpreter). What was important here is that the design is such that it only needs a few linear passes over arrays, and being able to reuse most of the arrays without the need for copy (like inOwners, outOwners).

rkuhn:
One aspect I have not yet fully penetrated: is it completely impossible that the same port object is used in two different materializations? The initial value for `id` should probably be `-1` and then when setting it to the real value there should be a requirement that it is still `-1`, making things fail fast instead of corrup “silently”.

If this approach works then it looks like a viable solution to the map overhead problem.

drewhk:
A port objects instance identity is used for various purposes, so it is pretty much true that a port can only belong to one module. Otherwise when two modules with the same port instance are used in the same graph it would lead to incredibly subtle bugs.

drewhk:
I.e. it does not matter how many times a thing is materialized, the point is that ports cannot be shared between modules. (if a port belongs to a certain module, then it has a certain position, i.e. id is always fixed.)

rkuhn:
Okay, I’d still sleep better if this assumption was codified.

drewhk:
but how can that be done?

rkuhn:
As I said: initialize to -1 and only allow writing to it if the previous value is -1 or the same that you’re writing.

drewhk:
Ah ok.

drewhk:
This is already in, so the experiment can be closed

PR #18516, opened 2015-09-18
=str: TickSource is now a GraphStage

drewhk:
Another custom actor bites the dust


drewhk:
I had to change the infrastructure slightly to accomodate for materialized values. Currently no materialized value support for fused islands but a temporary solution which works for the single-stage islands we currently have.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
```diff
@@ -240,4 +245,49 @@ object GraphStages {
     override def toString = "Balance"
   }
 
+  private object TickSource {
+    class TickSourceCancellable(cancelled: AtomicBoolean) extends Cancellable {
+      @volatile var callback: AsyncCallback[Unit] = _
```
drewhk: Ah stupid me, a Promise will work just as good.

patriknw:
LGTM

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
```diff
@@ -240,4 +246,53 @@ object GraphStages {
...
+          override def onPull() = () // Do nothing
+        })
+
+        override protected def onTimer(timerKey: Any) =
+          if (isAvailable(out)) push(out, tick)
+      }
+
+      (logic, cancellable)
+    }
+  }
```
ktoso: `override def toString = "TickSource"`?

drewhk: good point, missed that.

Issue #18486, opened 2015-09-16
Source.subscriber's Subscriber should log errors if subscribed more than once


ktoso:
This was posted to akka-user:

```scala
object Main2 extends App{
  implicit val system = ActorSystem("streamakka")
  implicit val materializer = ActorMaterializer()
  val sub1=Source.subscriber[String].to(Sink.foreach(println)).run()

  val pub1=Source.repeat("p1").runWith(Sink.publisher)
  val pub2=Source.repeat("p2").runWith(Sink.publisher)

  pub1.subscribe(sub1)
  pub2.subscribe(sub1)
}
```

Supposedly this does not throw, but only takes from one of the streams. Instead it should error on the 2nd subscription.

drewhk:
fat fingers, meant 1.1 at the first time, not invalid

ktoso:
Update: instead of throwing we should log an error that a subscriber was onSubscribe-ed multiple times,
as explained in detail here: https://github.com/akka/akka/pull/18521/files#r41328256

(otherwise we'd violate the RS spec).

drewhk:
Reopened, since it is not fully solved yet.

rkuhn:
This loses out against the rock (RS spec) and the hard place (being light-weight for VirtualProcessor).

PR #18521, opened 2015-09-21
+str #18486 Source.subscriber's Subscriber throws if subscribed more than once

agolubev:
The idea here is to throw IlligalStateException from VirtualProcessor.subscribe when tryOnSubscribe closed subscription (when Subscriber already has subscription per reactive spec).
Ref #18486

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
@@ -379,11 +379,9 @@ private[stream] final class VirtualProcessor[T] extends Processor[T, T] {
               case Completed  ⇒ tryOnComplete(s)
               case Failed(ex) ⇒ tryOnError(s, ex)
               case Allowed    ⇒ // all good
-
             }
           } catch {
-            case ex @ canNotSubscribeTheSameSubscriberMultipleTimesException ⇒ throw ex
-            case NonFatal(ex) ⇒ sub.cancel()
+            case NonFatal(ex) ⇒ if (isCancelled) throw ex else sub.cancel()
           }
       }
   }
--- a/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
@@ -360,6 +360,8 @@ private[stream] final class VirtualProcessor[T] extends Processor[T, T] {
   private val subscriptionStatus = new AtomicReference[AnyRef]
   private val terminationStatus = new AtomicReference[Termination]
 
+  private[this] def isCancelled() = subscriptionStatus.get == InertSubscriber
+
   override def subscribe(s: Subscriber[_ >: T]): Unit = {
     requireNonNullSubscriber(s)
     if (subscriptionStatus.compareAndSet(null, s)) () // wait for onSubscribe
@@ -367,19 +369,21 @@ private[stream] final class VirtualProcessor[T] extends Processor[T, T] {
       subscriptionStatus.get match {
         case sub: Subscriber[_] ⇒ rejectAdditionalSubscriber(s, "VirtualProcessor")
         case sub: Sub ⇒
-          try {
-            subscriptionStatus.set(s)
-            tryOnSubscribe(s, sub)
+          subscriptionStatus.set(s)
+          try tryOnSubscribe(s, sub) catch { case NonFatal(ex) ⇒ sub.cancel(); return }
+          if (!isCancelled) {
             sub.closeLatch() // allow onNext only now
-            terminationStatus.getAndSet(Allowed) match {
-              case null       ⇒ // nothing happened yet
-              case Completed  ⇒ tryOnComplete(s)
-              case Failed(ex) ⇒ tryOnError(s, ex)
-              case Allowed    ⇒ // all good
+            try {
+              terminationStatus.getAndSet(Allowed) match {
+                case null       ⇒ // nothing happened yet
+                case Completed  ⇒ tryOnComplete(s)
+                case Failed(ex) ⇒ tryOnError(s, ex)
+                case Allowed    ⇒ // all good
+              }
+            } catch {
+              case NonFatal(ex) ⇒ sub.cancel()
             }
-          } catch {
-            case NonFatal(ex) ⇒ sub.cancel()
-          }
+          } else throw new IllegalStateException("Cannot be subscribed more then once")
       }
   }
 
--- a/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
@@ -369,21 +369,22 @@ private[stream] final class VirtualProcessor[T] extends Processor[T, T] {
       subscriptionStatus.get match {
         case sub: Subscriber[_] ⇒ rejectAdditionalSubscriber(s, "VirtualProcessor")
         case sub: Sub ⇒
-          subscriptionStatus.set(s)
-          try tryOnSubscribe(s, sub) catch { case NonFatal(ex) ⇒ sub.cancel(); return }
-          if (!isCancelled) {
+          try {
+            subscriptionStatus.set(s)
+            tryOnSubscribe(s, sub)
+            if (isCancelled) throw canNotSubscribeTheSameSubscriberMultipleTimesException
             sub.closeLatch() // allow onNext only now
-            try {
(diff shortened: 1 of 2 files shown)


agolubev:
In the same time when Subscriber.onSubscribe throws exception - this breaks tck tests.

ktoso:
OK TO TEST

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -367,19 +369,21 @@ private[stream] final class VirtualProcessor[T] extends Processor[T, T] {
       subscriptionStatus.get match {
         case sub: Subscriber[_] ⇒ rejectAdditionalSubscriber(s, "VirtualProcessor")
         case sub: Sub ⇒
-          try {
-            subscriptionStatus.set(s)
-            tryOnSubscribe(s, sub)
+          subscriptionStatus.set(s)
+          try tryOnSubscribe(s, sub) catch { case NonFatal(ex) ⇒ sub.cancel(); return }
```
patriknw: we don't use `return`

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -367,19 +369,21 @@ private[stream] final class VirtualProcessor[T] extends Processor[T, T] {
...
+                case Failed(ex) ⇒ tryOnError(s, ex)
+                case Allowed    ⇒ // all good
+              }
+            } catch {
+              case NonFatal(ex) ⇒ sub.cancel()
             }
-          } catch {
-            case NonFatal(ex) ⇒ sub.cancel()
-          }
+          } else throw new IllegalStateException("Cannot be subscribed more then once")
```
patriknw: isn't this exception defined in the ReactiveStreamsComplience utility class?

agolubev: thanks for the hint.

agolubev:
@patriknw, it's ready for review

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -370,14 +372,17 @@ private[stream] final class VirtualProcessor[T] extends Processor[T, T] {
...
             sub.closeLatch() // allow onNext only now
             terminationStatus.getAndSet(Allowed) match {
               case null       ⇒ // nothing happened yet
               case Completed  ⇒ tryOnComplete(s)
               case Failed(ex) ⇒ tryOnError(s, ex)
               case Allowed    ⇒ // all good
+
             }
           } catch {
+            case ex @ canNotSubscribeTheSameSubscriberMultipleTimesException ⇒ throw ex
```
patriknw: This is not correct, this will catch all exceptions.
canNotSubscribeTheSameSubscriberMultipleTimesException is like any bind variable in the match.
It would have had to be `canNotSubscribeTheSameSubscriberMultipleTimesException` or starting with upper case.

Perhaps:

      } catch {
         case NonFatal(ex) ⇒ if (isCancelled) throw ex else sub.cancel()
      }

agolubev: Yeah, my bad. 
Code snippet make sense as tryComplete and tryError do not cancel VirtualProcessor in the way onSubscription does

drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -370,6 +372,7 @@ private[stream] final class VirtualProcessor[T] extends Processor[T, T] {
           try {
             subscriptionStatus.set(s)
             tryOnSubscribe(s, sub)
+            if (isCancelled) throw canNotSubscribeTheSameSubscriberMultipleTimesException
```
viktorklang: This violates Reactive Streams spec rule 1:9: https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.0/README.md#specification

agolubev: Hm, than we need to close case #18486 as invalid request

viktorklang: Simple fix :)

ktoso: Yes and no.

The wording I used in #18486 was indeed unfortunate (wrong), as it violates Rule 1.9.
However! The ticket was actually aimed to resolve the case as shown in the snippet of that ticket, and originated from https://groups.google.com/forum/#!searchin/akka-user/https$3A$2F$2Fgithub.com$2Fakka$2Fakka$2Fissues$2F18486%7Csort:relevance/akka-user/EMuOMaXsG84/SgzXdOPmAgAJ

In other words, it is about [Rule 2.12](https://github.com/reactive-streams/reactive-streams-jvm/tree/v1.0.0#2.12) and [Rule 2.13](https://github.com/reactive-streams/reactive-streams-jvm/tree/v1.0.0#2.13) - i.e. onSubscribe can only be called once + "... *for all other situations the only legal way for a Subscriber to signal failure is by cancelling its Subscription. In the case that this rule is violated, any associated Subscription to the Subscriber MUST be considered as cancelled, and the caller MUST raise this error condition in a fashion that is adequate for the runtime environment.*".

The ticket description should be amended to "log error", and this PR should do so then, as Rule 2.13 mandates.
Sounds about right? :)

agolubev: @ktoso, if you change the issue subjct - I'll pull the fix for this :)

viktorklang: Note that the tracking of additional subscribers is on a best effort basis as the rule for no concurrent signalling still applies. But yes, @ktoso, I agree with your sentiment! :)

ktoso: I changed the title of the ticket, thanks @agolubev !

drewhk:
Oh, I though it was settled and I merged it. I don't know about the logging though, the IdentityProcessor is actorless and therefore adding a logger there might turn out to be problematic. Not throwing in this case feels weird, spec or not. Somehow a Source needs to signal that it is not reusable at materialization time. This needs a bit more thought.

viktorklang:
revert the merge?

drewhk:
yep, done.

drewhk:
btw, the proper solution would be onError-ing the second subscriber isn't it?

drewhk:
> Publisher.subscribe MUST call onSubscribe on the provided Subscriber prior to any other signals to that Subscriber and MUST return normally, except when the provided Subscriber is null in which case it MUST throw a java.lang.NullPointerException to the caller, for all other situations the only legal way to signal failure (or reject the Subscriber) is by calling onError (after calling onSubscribe).

viktorklang:
@drewhk yep. I assume that we don't properly verify this implementation if it is currently passing the tests.

agolubev:
to the @drewhk point. In general we have two cases:
1) When one publisher is calling subscribe twice. Second call must do OnError for Subscriber. And this was implemented long time ago (ReactiveStreamsCompliance.rejectAdditionalSubscriber is called).
2) This ticket is about case when one Subscriber is called with OnSubscribe twice. And currently it's only canceling subscription for second call as there is no other way to tell Publisher that something is wrong (with existing API). With this behavior Source is canceled and not used further so anything that is providing data to source already receive cancel message. Don't think that sending OnError to functioning Subscriber is a good idea.
And yes, there is no logging currently for VirtualProcessor I need to figure out how to do this for non actors.

viktorklang:
@agolubev
Re: 2) the appropriate action is to `cancel` the additional Subscription and log the condition if possible.

drewhk:
Ah, Ok, I guess we all misunderstood the case. Btw, @viktorklang this code **is** TCK checked. In the light of this, I no longer think that the original ticket is valid.

viktorklang:
@drewhk I trust your call on this! :)

agolubev:
@drewhk, should we add logging error (per @ktoso point)? if it's possible as it's not an actor

drewhk:
I would like to keep VirtualProcessor lightweight I would keep it as it is. @ktoso @viktorklang your opinions?

viktorklang:
@drewhk I agree.

ktoso:
Probably not worth making it heavy just for that, let's keep as is.

Issue #15159, opened 2014-05-10
include failed requirement type in mailbox configuration error message


rkuhn:
Currently it says

~~~
produced message queue type [class akka.dispatch.UnboundedPriorityMailbox$MessageQueue] does not fulfill requirement for actor class [class services.CourseActor]
~~~

and it would be helpful to state what the requirement was.

michaelahlers:
Just ran into this problem, too. The limitation isn't clear from any documentation I've seen.

devisaneer:
I also just ran into this exact same problem - I'm not complaining in any way at all (good vides from me about akka), though a one liner limitation in the Mailbox section of the doco could be cool. As a side note/Q?  Is there any out-of-the-box way to use Stash with a Priority in  2.4 snapshot?

PR #18526, opened 2015-09-21
=act #15159 Better error when maibox does not fulfill requirements

johanandren:


patriknw:
LGTM

patriknw:
since we have the opportunity of not changing the bits between RC3 and 2.4.0 this will also wait until 2.4.1

patriknw:
Refs #15159

Issue #18527, opened 2015-09-22
failed: SteppingInMemPersistentActorStashingSpec


patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/2914/consoleFull

    [info] - should handle async callback not happening until next message has been stashed *** FAILED *** (117 milliseconds)
    [info]   java.lang.AssertionError: assertion failed: expected List(a, c, b), found List(a)
    [info]   at scala.Predef$.assert(Predef.scala:165)
    [info]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:351)
    [info]   at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:327)
    [info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:735)
    [info]   at akka.persistence.SteppingInMemPersistentActorStashingSpec$$anonfun$2$$anonfun$apply$mcV$sp$4.apply$mcV$sp(PersistentActorStashingSpec.scala:167)
    [info]   at akka.persistence.SteppingInMemPersistentActorStashingSpec$$anonfun$2$$anonfun$apply$mcV$sp$4.apply(PersistentActorStashingSpec.scala:147)
    [info]   at akka.persistence.SteppingInMemPersistentActorStashingSpec$$anonfun$2$$anonfun$apply$mcV$sp$4.apply(PersistentActorStashingSpec.scala:147)

patriknw:
cc @johanandren

patriknw:
https://jenkins.akka.io:8498/job/akka-nightly/2623/consoleFull

johanandren:
Hmm, that shouldn't be race:y, it should either nto reach the assert at all or succeed...

patriknw:
Here is another one, looks slightly different I think: https://jenkins.akka.io:8498/job/akka-nightly/2639/consoleFull

johanandren:
Hmm, pretty sure thats unrelated, its in in `akka.typed.ActorContextSpec`

patriknw:
Sorry, I misread, StepWise is not the same as the stepping journal :-)

PR #18538, opened 2015-09-23
=per #18527 fix race condition in async stash test

johanandren:
I think there may have been a race between the next message being received and the async callback block being triggered in the test. Fixed with retrying within 3 seconds.

### DIFF
--- a/akka-persistence/src/test/scala/akka/persistence/PersistentActorStashingSpec.scala
+++ b/akka-persistence/src/test/scala/akka/persistence/PersistentActorStashingSpec.scala
@@ -163,8 +163,12 @@ class SteppingInMemPersistentActorStashingSpec extends PersistenceSpec(
       SteppingInmemJournal.step(journal)
       SteppingInmemJournal.step(journal)
 
-      persistentActor ! GetState
-      expectMsg(List("a", "c", "b"))
+      within(3.seconds) {
+        awaitAssert {
+          persistentActor ! GetState
+          expectMsg(List("a", "c", "b"))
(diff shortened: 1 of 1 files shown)


johanandren:
References #18527

akka-persistence/src/test/scala/akka/persistence/PersistentActorStashingSpec.scala
```diff
@@ -163,8 +163,12 @@ class SteppingInMemPersistentActorStashingSpec extends PersistenceSpec(
...
       SteppingInmemJournal.step(journal)
 
-      persistentActor ! GetState
-      expectMsg(List("a", "c", "b"))
+      within(3.seconds) {
+        awaitAssert {
+          persistentActor ! GetState
+          expectMsg(List("a", "c", "b"))
+        }
+      }
```
ktoso: The problem could also be solved by injecting the an "replyTo" (or rather `stateProbe`) test probe into the persistentActor, such that it informs the probe on applying `updateState` about the current one, then the race between step-induced processing and `GetState` wouldn't happen.

I don't really mind this solution, it's just a test and bound to complete successfully quickly IMO. So it can stay as is.

ktoso:
Proposed alternate solution in https://github.com/akka/akka/pull/18538/files#r40200077 however I don't really mind the current one, since it's bound to complete successfully quickly anyway. 

LGTM once you're happy with it

patriknw:
LGTM, with this or with Konrad's proposal

PR #18541, opened 2015-09-23
+rem #18353: Prune reliable deliver actors

drewhk:
Prune reliable delivery actors more aggressively:
- They are dropped if there is no system message state to be stored
- They are removed after a long period (default 5 days) inactivity when there is no traffic of any kind nor an active association

### DIFF
--- a/akka-remote/src/main/scala/akka/remote/Endpoint.scala
+++ b/akka-remote/src/main/scala/akka/remote/Endpoint.scala
@@ -21,7 +21,7 @@ import akka.serialization.Serialization
 import akka.util.ByteString
 import akka.{ OnlyCauseStackTrace, AkkaException }
 import java.io.NotSerializableException
-import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.{ TimeUnit, TimeoutException, ConcurrentHashMap }
 import scala.annotation.tailrec
 import scala.concurrent.duration.{ Duration, Deadline }
 import scala.util.control.NonFatal
@@ -168,6 +168,7 @@ private[remote] object ReliableDeliverySupervisor {
 
   case object IsIdle
   case object Idle
+  case object TooLongIdle
 
   def props(
     handleOrActive: Option[AkkaProtocolHandle],
@@ -200,6 +201,8 @@ private[remote] class ReliableDeliverySupervisor(
   val autoResendTimer = context.system.scheduler.schedule(
     settings.SysResendTimeout, settings.SysResendTimeout, self, AttemptSysMsgRedelivery)
 
+  private var bufferWasInUse = false
+
   override val supervisorStrategy = OneForOneStrategy(loggingEnabled = false) {
     case e @ (_: AssociationProblem) ⇒ Escalate
     case NonFatal(e) ⇒
@@ -207,12 +210,14 @@ private[remote] class ReliableDeliverySupervisor(
       log.warning("Association with remote system [{}] has failed, address is now gated for [{}] ms. Reason: [{}] {}",
         remoteAddress, settings.RetryGateClosedFor.toMillis, e.getMessage, causedBy)
       uidConfirmed = false // Need confirmation of UID again
-      if ((resendBuffer.nacked.nonEmpty || resendBuffer.nonAcked.nonEmpty) && bailoutAt.isEmpty)
-        bailoutAt = Some(Deadline.now + settings.InitialSysMsgDeliveryTimeout)
-      context.become(gated)
-      currentHandle = None
-      context.parent ! StoppedReading(self)
-      Stop
+      if (bufferWasInUse) {
+        if ((resendBuffer.nacked.nonEmpty || resendBuffer.nonAcked.nonEmpty) && bailoutAt.isEmpty)
+          bailoutAt = Some(Deadline.now + settings.InitialSysMsgDeliveryTimeout)
+        context.become(gated)
+        currentHandle = None
+        context.parent ! StoppedReading(self)
+        Stop
+      } else Escalate
   }
 
   var currentHandle: Option[AkkaProtocolHandle] = handleOrActive
@@ -237,6 +242,7 @@ private[remote] class ReliableDeliverySupervisor(
   var writer: ActorRef = createWriter()
   var uid: Option[Int] = handleOrActive map { _.handshakeInfo.uid }
   var bailoutAt: Option[Deadline] = None
+  var maxSilenceTimer: Option[Cancellable] = None
   // Processing of Acks has to be delayed until the UID after a reconnect is discovered. Depending whether the
   // UID matches the expected one, pending Acks can be processed, or must be dropped. It is guaranteed that for
   // any inbound connections (calling createWriter()) the first message from that connection is GotUid() therefore
@@ -255,6 +261,7 @@ private[remote] class ReliableDeliverySupervisor(
     (resendBuffer.nacked ++ resendBuffer.nonAcked) foreach { s ⇒ context.system.deadLetters ! s.copy(seqOpt = None) }
     receiveBuffers.remove(Link(localAddress, remoteAddress))
     autoResendTimer.cancel()
+    maxSilenceTimer.foreach(_.cancel())
   }
 
   override def postRestart(reason: Throwable): Unit = {
@@ -291,7 +298,7 @@ private[remote] class ReliableDeliverySupervisor(
       context.parent ! StoppedReading(self)
       if (resendBuffer.nonAcked.nonEmpty || resendBuffer.nacked.nonEmpty)
         context.system.scheduler.scheduleOnce(settings.SysResendTimeout, self, AttemptSysMsgRedelivery)
-      context.become(idle)
+      goToIdle()
     case g @ GotUid(receivedUid, _) ⇒
       bailoutAt = None
       context.parent ! g
@@ -321,8 +328,8 @@ private[remote] class ReliableDeliverySupervisor(
             new java.util.concurrent.TimeoutException("Delivery of system messages timed out and they were dropped."))
         writer = createWriter()
         // Resending will be triggered by the incoming GotUid message after the connection finished
-        context.become(receive)
-      } else context.become(idle)
+        goToActive()
+      } else goToIdle()
     case AttemptSysMsgRedelivery               ⇒ // Ignore
     case s @ Send(msg: SystemMessage, _, _, _) ⇒ tryBuffer(s.copy(seqOpt = Some(nextSeq())))
     case s: Send                               ⇒ context.system.deadLetters ! s
@@ -338,16 +345,32 @@ private[remote] class ReliableDeliverySupervisor(
       writer = createWriter()
       // Resending will be triggered by the incoming GotUid message after the connection finished
       handleSend(s)
-      context.become(receive)
+      goToActive()
     case AttemptSysMsgRedelivery ⇒
       if (resendBuffer.nacked.nonEmpty || resendBuffer.nonAcked.nonEmpty) {
         writer = createWriter()
         // Resending will be triggered by the incoming GotUid message after the connection finished
-        context.become(receive)
+        goToActive()
       }
+    case TooLongIdle ⇒
+      throw new HopelessAssociation(localAddress, remoteAddress, uid,
+        new TimeoutException("Remote system has been silent for too long. " +
+          s"(more than ${settings.QuarantineSilentSystemTimeout.toUnit(TimeUnit.HOURS)} hours)"))
     case EndpointWriter.FlushAndStop ⇒ context.stop(self)
     case EndpointWriter.StopReading(w, replyTo) ⇒
       replyTo ! EndpointWriter.StoppedReading(w)
+  }
+
+  private def goToIdle(): Unit = {
+    if (bufferWasInUse && maxSilenceTimer.isEmpty)
+      maxSilenceTimer = Some(context.system.scheduler.scheduleOnce(settings.QuarantineSilentSystemTimeout, self, TooLongIdle))
+    context.become(idle)
+  }
+
+  private def goToActive(): Unit = {
+    maxSilenceTimer.foreach(_.cancel())
+    maxSilenceTimer = None
+    context.become(receive)
   }
 
   def flushWait: Receive = {
@@ -381,6 +404,7 @@ private[remote] class ReliableDeliverySupervisor(
   private def tryBuffer(s: Send): Unit =
     try {
       resendBuffer = resendBuffer buffer s
+      bufferWasInUse = true
     } catch {
       case NonFatal(e) ⇒ throw new HopelessAssociation(localAddress, remoteAddress, uid, e)
     }
--- a/akka-remote/src/main/scala/akka/remote/Remoting.scala
+++ b/akka-remote/src/main/scala/akka/remote/Remoting.scala
@@ -329,6 +329,8 @@ private[remote] object EndpointManager {
         readonlyToAddress -= endpoint
       }
 
+    def addressForWriter(writer: ActorRef): Option[Address] = writableToAddress.get(writer)
+
     def writableEndpointWithPolicyFor(address: Address): Option[EndpointPolicy] = addressToWritable.get(address)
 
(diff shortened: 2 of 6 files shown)


drewhk:
Refs #18353

akka-remote/src/main/scala/akka/remote/AckedDelivery.scala
```diff
@@ -86,12 +86,14 @@ class ResendUnfulfillableException
...
  *               will be not stored but rejected with [[java.lang.IllegalArgumentException]]
+ * @param hasBeenUsed Indicates whether the buffer has been used at all
  */
 case class AckedSendBuffer[T <: HasSequenceNumber](
   capacity: Int,
   nonAcked: IndexedSeq[T] = Vector.empty[T],
   nacked: IndexedSeq[T] = Vector.empty[T],
-  maxSeq: SeqNo = SeqNo(-1)) {
+  maxSeq: SeqNo = SeqNo(-1),
+  hasBeenUsed: Boolean = false) {
```
patriknw: this is unfortunately not internal, so we have to be creative around binary compatibility issue

akka-remote/src/main/scala/akka/remote/Endpoint.scala
```diff
@@ -207,12 +208,16 @@ private[remote] class ReliableDeliverySupervisor(
       log.warning("Association with remote system [{}] has failed, address is now gated for [{}] ms. Reason: [{}] {}",
         remoteAddress, settings.RetryGateClosedFor.toMillis, e.getMessage, causedBy)
       uidConfirmed = false // Need confirmation of UID again
-      if ((resendBuffer.nacked.nonEmpty || resendBuffer.nonAcked.nonEmpty) && bailoutAt.isEmpty)
-        bailoutAt = Some(Deadline.now + settings.InitialSysMsgDeliveryTimeout)
-      context.become(gated)
-      currentHandle = None
-      context.parent ! StoppedReading(self)
-      Stop
+      if (!resendBuffer.hasBeenUsed) {
```
patriknw: I'd rather swap the statements to avoid the negation

akka-remote/src/main/scala/akka/remote/Endpoint.scala
```diff
@@ -338,18 +345,32 @@ private[remote] class ReliableDeliverySupervisor(
...
     case AttemptSysMsgRedelivery ⇒
       if (resendBuffer.nacked.nonEmpty || resendBuffer.nonAcked.nonEmpty) {
         writer = createWriter()
         // Resending will be triggered by the incoming GotUid message after the connection finished
-        context.become(receive)
+        goToActive()
       }
+    case TooLongIdle ⇒
+      throw new HopelessAssociation(localAddress, remoteAddress, uid,
+        new TimeoutException("Remote system has been silent for too long."))
```
patriknw: include the QuarantineSilentSystemTimeout in the error message?

akka-remote/src/main/scala/akka/remote/Endpoint.scala
```diff
@@ -338,18 +345,32 @@ private[remote] class ReliableDeliverySupervisor(
...
   }
 
+  private def goToIdle(): Unit = {
+    if (resendBuffer.hasBeenUsed && maxSilenceTimer.isEmpty)
+      maxSilenceTimer = Some(context.system.scheduler.scheduleOnce(settings.QuarantineSilentSystemTimeout, self, TooLongIdle))
+    context.become(idle)
+  }
+
+  private def goToActive(): Unit = {
+    maxSilenceTimer.foreach(_.cancel())
```
patriknw: maxSilenceTimer = None ?

akka-remote/src/main/scala/akka/remote/Remoting.scala
```diff
@@ -409,11 +411,9 @@ private[remote] class EndpointManager(conf: Config, log: LoggingAdapter) extends
...
   var transportMapping: Map[Address, AkkaProtocolTransport] = Map()
 
-  def retryGateEnabled = settings.RetryGateClosedFor > Duration.Zero
-  val pruneInterval: FiniteDuration = if (retryGateEnabled) settings.RetryGateClosedFor * 2 else Duration.Zero
-  val pruneTimerCancellable: Option[Cancellable] = if (retryGateEnabled)
-    Some(context.system.scheduler.schedule(pruneInterval, pruneInterval, self, Prune))
-  else None
+  val pruneInterval: FiniteDuration = settings.RetryGateClosedFor * 2
+  val pruneTimerCancellable: Cancellable =
+    context.system.scheduler.schedule(pruneInterval, pruneInterval, self, Prune)
```
patriknw: what about `Duration.Zero`, which was handled previously?

akka-remote/src/test/scala/akka/remote/ActorsLeakSpec.scala
```diff
@@ -0,0 +1,176 @@
+package akka.remote
```
patriknw: copyright

akka-remote/src/test/scala/akka/remote/ActorsLeakSpec.scala
```diff
@@ -0,0 +1,176 @@
...
+
+  "Remoting" must {
+
+    "not leak actors" in {
+      val ref = system.actorOf(Props[EchoActor], "echo")
+      val echoPath = RootActorPath(RARP(system).provider.getDefaultAddress) / "user" / "echo"
+
+      val initialActors =
+        (collectLiveActors(system.actorFor("/system/endpointManager")) ++
+          collectLiveActors(system.actorFor("/system/transports"))).toSet
```
patriknw: actorFor? can't you identify them?

akka-remote/src/test/scala/akka/remote/ActorsLeakSpec.scala
```diff
@@ -0,0 +1,176 @@
...
+
+      //Clean shutdown case
+      for (_ ← 1 to 3) {
+
+        val remoteSystem = ActorSystem(
+          "remote",
+          ConfigFactory.parseString("akka.remote.netty.tcp.port = 2553")
```
patriknw: don't use hardcoded ports, since that prevents parallel execution on same machine (when we need that)

akka-remote/src/test/scala/akka/remote/ActorsLeakSpec.scala
drewhk: Ah, that was a quick hack, will replace with Identify.

akka-remote/src/test/scala/akka/remote/ActorsLeakSpec.scala
drewhk: thanks, forgot that

akka-remote/src/main/scala/akka/remote/Remoting.scala
drewhk: that remained from old times when it was possible to disable gating. Nothing used retryGateEnabled.

akka-remote/src/test/scala/akka/remote/ActorsLeakSpec.scala
```diff
@@ -0,0 +1,176 @@
...
+          // This will make sure that no SHUTDOWN message gets through
+          Await.ready(
+            RARP(system).provider.transport.managementCommand(ForceDisassociate(remoteAddress)),
+            3.seconds)
+
+        } finally {
+          remoteSystem.shutdown()
+        }
+
+        within(6.seconds) {
```
patriknw: I don't think within is used for the EventFilter, unfortunately. You probably already found that since you have configured `akka.test.filter-leeway`

akka-remote/src/test/scala/akka/remote/ActorsLeakSpec.scala
```diff
@@ -0,0 +1,176 @@
...
+          EventFilter.warning(pattern = "Association with remote system", occurrences = 1).intercept {
+            remoteSystem.awaitTermination()
```
patriknw: no timeout here?

akka-remote/src/main/scala/akka/remote/Endpoint.scala
drewhk: True! Need to somehow test this.

akka-remote/src/main/scala/akka/remote/Endpoint.scala
drewhk: good point.

akka-remote/src/main/scala/akka/remote/AckedDelivery.scala
drewhk: aaah. Yes you are right, this is a public class. I can encode this flag in the actor itself.

akka-remote/src/test/scala/akka/remote/ActorsLeakSpec.scala
drewhk: yep, need to remove this.

akka-remote/src/test/scala/akka/remote/ActorsLeakSpec.scala
drewhk: ah right, it does not have one by default.

akka-remote/src/test/scala/akka/remote/ActorsLeakSpec.scala
```diff
@@ -0,0 +1,176 @@
...
+
+        remoteSystem.actorSelection(echoPath).tell(Identify(1), probe.ref)
+        probe.expectMsgType[ActorIdentity].ref.nonEmpty should be(true)
+
+        // Watch a remote actor - this results in system message traffic
+        system.actorSelection(RootActorPath(remoteAddress) / "user" / "stoppable") ! Identify(1)
+        val remoteActor = expectMsgType[ActorIdentity].ref.get
+        watch(remoteActor)
+        remoteActor ! "stop"
+        expectMsgType[Terminated]
```
patriknw: `expectTerminated` ?

akka-remote/src/main/scala/akka/remote/Endpoint.scala
patriknw: so this is the fix, apart from the QuarantineSilentSystemTimeout? anything else that I missed?

drewhk: yep, surprisingly :)

akka-remote/src/main/scala/akka/remote/Remoting.scala
patriknw: ok, so RetryGateClosedFor is required to be > zero

drewhk: It turns out its not:

    val RetryGateClosedFor: FiniteDuration = {
      config.getMillisDuration("akka.remote.retry-gate-closed-for")
    } requiring (_ >= Duration.Zero, "retry-gate-closed-for must be >= 0")

Equality must go.

patriknw: but can we change that? Could it have been used as a way to disable gating?

drewhk: I don't think so, It would still drop some messages, although for the shortest time (scheduler granularity+messaging delay)

drewhk: I.e. there is no real zero anyway. Alternatively, I can change zero to a small minimum value automatically. That will remain compatible. A warning can be logged too.

patriknw: but previously nothing was scheduled for zero

drewhk: But that was not a documented feature, it was a bug. The docs are quite clear about that there is always a Gated state: http://doc.akka.io/docs/akka/2.3.14/scala/remoting.html#Lifecycle_and_Failure_Recovery_Model

Do you think we should keep it as it is, although this was not documented and contradicts the docs? Before 2.3 there was an explicit flag, then we removed the feature along with the corresponding tests and this remained somehow.

Not an easy question, maybe best to go conservative, log a Warning for the zero setting, and keep the current behavior.What about 2.4? Is it too late to fix it there?

drewhk: Ah, no, this is not a question, this timer does not control the gating at all, it is the timer that checks for markers that can be removed. I.e. if you set the value to 0 this would mean that the gated markers remain there unremoved.

drewhk: Will double check tomorrow though, but now I remember why I touched this.

patriknw: ok, I don't know, if it is a bug to allow zero then we can of course change, but otherwise it a bit hard to fail the system startup (and claim that it is a drop in replacement).

This have to go into 2.4.1, unless we have to make another 2.4.0 RC for other reasons, so the situation is pretty much the same there.

drewhk: No, the solution is to treat zero as a minimal timeout (maybe) for gating, and set a minimum timer interval for the prune timer (which is the above timer).

I.e. people who didn't use zero are unaffected, people who used zero still had gating but for minimal time (scheduler granularity + delay dependent), but now they don't have a memory leak either :)

patriknw: :+1:

patriknw:
@drewhk any chance that you can complete this and forward port to master?

drewhk:
I am not sure if I can finish this today, but I can do it tomorrow.

viktorklang:
@drewhk Ping :)

drewhk:
I finish this today

drewhk:
ClusterClientSpec failed. Not sure if it is related:

```
[JVM-1] - must re-establish connection to receptionist after server restart *** FAILED ***
[JVM-1]   java.lang.AssertionError: assertion failed: expected bonjour4-ack, found bonjour3-ack
[JVM-1]   at scala.Predef$.assert(Predef.scala:179)
[JVM-1]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:339)
[JVM-1]   at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:324)
[JVM-1]   at akka.testkit.TestKit.expectMsg(TestKit.scala:718)
[JVM-1]   at akka.contrib.pattern.ClusterClientSpec$$anonfun$1$$anonfun$apply$mcV$sp$26$$anonfun$apply$mcV$sp$27$$anonfun$apply$mcV$sp$28.apply$mcV$sp(ClusterClientSpec.scala:261)
[JVM-1]   at akka.remote.testkit.MultiNodeSpec.runOn(MultiNodeSpec.scala:338)
[JVM-1]   at 
```

drewhk:
@patriknw do you have a quick idea before I start to deep dive into that test?

akka-remote/src/main/scala/akka/remote/AckedDelivery.scala
```diff
@@ -86,6 +86,7 @@ class ResendUnfulfillableException
  * @param nacked   Sequence of messages that has been explicitly negative acknowledged.
  * @param maxSeq The maximum sequence number that has been stored in this buffer. Messages having lower sequence number
  *               will be not stored but rejected with [[java.lang.IllegalArgumentException]]
+ * @param hasBeenUsed Indicates whether the buffer has been used at all
```
patriknw: `hasBeenUsed` is not a parameter any more

patriknw:
@drewhk the failure is because the ""bonjour3-ack" spills over from the previous test step. This has been fixed in master: https://github.com/akka/akka/blob/master/akka-cluster-tools/src/multi-jvm/scala/akka/cluster/client/ClusterClientSpec.scala#L254

Either you apply the same fix to 2.3, or we just build one more time and if good we ignore it.

patriknw:
LGTM

akka-remote/src/main/scala/akka/remote/AckedDelivery.scala
drewhk: thanks!

drewhk:
I will check the build again. More testing does not hurt.

PR #18542, opened 2015-09-23
=rem #18339 Use explicit handshake timeout

patriknw:
- instead of using transport failure detector
- add a new config property akka.remote.handshake-timeout, but
  for netty.tcp and netty.ssl the existing netty.tcp.connection-timeout
  setting will be used

### DIFF
--- a/akka-remote/src/main/resources/reference.conf
+++ b/akka-remote/src/main/resources/reference.conf
@@ -98,6 +98,11 @@ akka {
 
     # Acknowledgment timeout of management commands sent to the transport stack.
     command-ack-timeout = 30 s
+    
+    # The timeout for outbound associations to perform the handshake.
+    # If the transport is akka.remote.netty.tcp or akka.remote.netty.ssl
+    # the configured connection-timeout for the transport will be used instead.
+    handshake-timeout = 15 s
 
     # If set to a nonempty string remoting will use the given dispatcher for
     # its internal actors otherwise the default dispatcher is used. Please note
--- a/akka-remote/src/main/scala/akka/remote/transport/AkkaProtocolTransport.scala
+++ b/akka-remote/src/main/scala/akka/remote/transport/AkkaProtocolTransport.scala
@@ -3,6 +3,7 @@
  */
 package akka.remote.transport
 
+import java.util.concurrent.TimeoutException
 import akka.ConfigurationException
 import akka.actor.SupervisorStrategy.Stop
 import akka.actor._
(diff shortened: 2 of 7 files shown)


patriknw:
Refs #18339

patriknw:
This should be merged for 2.4.1, it is probably only a problem when using test transport.

akka-remote/src/main/scala/akka/remote/transport/AkkaProtocolTransport.scala
```diff
@@ -402,6 +420,16 @@ private[transport] class ProtocolStateActor(initialData: InitialProtocolStateDat
...
 
+    case Event(HandshakeTimer, OutboundUnderlyingAssociated(_, wrappedHandle)) ⇒
+      sendDisassociate(wrappedHandle, Unknown)
+      stop(FSM.Failure(TimeoutReason("No response from remote for outbound association. Handshake timed out after " +
+        s"[${settings.HandshakeTimeout.toMillis} ms].")))
+
+    case Event(HandshakeTimer, InboundUnassociated(_, wrappedHandle)) ⇒
+      sendDisassociate(wrappedHandle, Unknown)
+      stop(FSM.Failure(TimeoutReason("No response from remote for inbound association. Handshake timed out after " +
+        s"[${settings.HandshakeTimeout.toMillis} ms].")))
```
patriknw: I'm unsure about the timeout for inbound associations. It was not handled previously, and is perhaps not needed, but it might not hurt either?

akka-remote/src/main/resources/reference.conf
```diff
@@ -98,6 +98,11 @@ akka {
 
     # Acknowledgment timeout of management commands sent to the transport stack.
     command-ack-timeout = 30 s
+    
+    # The timeout for outbound associations to perform the handshake.
+    # If the transport is akka.remote.netty.tcp or akka.remote.netty.ssl
+    # the configured connection-timeout for the transport will be used instead.
```
drewhk: I am not sure about that. The connection-timeout relates to TCP events, which is handled by the kernel, while the Akka Protocol is at the JVM level. I am not sure these timeouts mix well.

akka-remote/src/main/scala/akka/remote/transport/AkkaProtocolTransport.scala
drewhk: I think it should have this timeout.

drewhk:
Can you add the two timeout test cases to AkkaProtocolSpec? That tests this actor in isolation.

akka-remote/src/main/resources/reference.conf
ktoso: I remember a customer issue in which SSL was taking so long to init that connections timed out. So from that perspective, if read this PR correctly it would be easier to tune for their case as well.

The error message would also have been nicer, so I'm for this PR, though maybe I'm missing some more fuzzy interplay with other remoting things

patriknw: I was thinking that otherwise the user would always have to change both those settings. When would you use different timeout here than the connection-timeout?

drewhk: on a busy system the actual TCP connection might finish much quicker than a remoting handshake, but maybe I am just too paranoid. Less knobs to tune is maybe better.

patriknw: but isn't that same issue from the user's perspective? at least all sane users.

There is also a backwards compatibility aspect of this.
Let's say that someone has configured connection-timeout = 30s, then we can't suddenly introduce a new property that fail the connection attempt after 15s (which is the default for handshake-timeout)

drewhk: Both are good points, convinced :)

patriknw:
@drewhk I added a test for the timeouts. I'm a bit lost in those low level test utilities, but at least it tests something. See 8418b14

I decided to start the timer for outbound connections immediately when the actor is started so that also timeouts before WaitingHandshake are covered.

patriknw:
PLS BUILD

patriknw:
@drewhk please take a look at the latest here so this can be merged

drewhk:
This PR received the Seal of approval from the Remoting Department.

patriknw:
thanks, ah bin compatibility issue

Issue #17695, opened 2015-06-11
Provide an API for Cluster Sharding state retrieval


2m:
Would be great to have a user API where one could subscribe for Cluster Sharding state changes.

patriknw:
Do you have any other use case in mind than logging?

2m:
This could enable a monitoring tool to provide an overview of what shard regions are in what nodes. What sharded entries are active and what passivated.

patriknw:
ok, thanks for clarifying, makes sense

2m:
Oh. I thought you started working on it right away. :)

patriknw:
no, wrong label

PR #18543, opened 2015-09-23
+clu #17695 add a way to inspect the current sharding state

johanandren:
First stab at it:
- Update the existing `GetCurrentRegions`/`CurrentRegions` to contain the actor paths
- A new message pair: `GetShardRegionState`/`CurrentShardRegionState` allows for querying a region for it's current shards and the current `EntityIds` of it

The first change allows to query the returned actor paths to get the entire state with the second. Not sure if this is too user-unfriendly. The end of the test does exactly this.

Refs #17695

### DIFF
--- a/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
+++ b/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
@@ -9,7 +9,7 @@ import akka.actor.ActorRef
 import akka.actor.Deploy
 import akka.actor.Props
 import akka.actor.Terminated
-import akka.cluster.sharding.Shard.ShardCommand
+import akka.cluster.sharding.Shard.{ GetCurrentShardState, ShardCommand }
 import akka.persistence.PersistentActor
 import akka.persistence.SnapshotOffer
 import akka.actor.Actor
@@ -41,6 +41,11 @@ private[akka] object Shard {
   }
 
   /**
+   * A query for information about the shard
+   */
+  sealed trait ShardQuery
+
+  /**
    * `State` change for starting an entity in this `Shard`
    */
   @SerialVersionUID(1L) final case class EntityStarted(entityId: EntityId) extends StateChange
@@ -49,6 +54,14 @@ private[akka] object Shard {
    * `State` change for an entity which has terminated.
    */
   @SerialVersionUID(1L) final case class EntityStopped(entityId: EntityId) extends StateChange
+
+  @SerialVersionUID(1L) case object GetCurrentShardState extends ShardQuery
+
+  @SerialVersionUID(1L) final case class CurrentShardState(shardId: ShardRegion.ShardId, entityIds: Set[EntityId])
+
+  @SerialVersionUID(1L) case object GetShardStats extends ShardQuery
+
+  @SerialVersionUID(1L) final case class ShardStats(shardId: ShardRegion.ShardId, entityCount: Int)
 
   object State {
     val Empty = State()
@@ -101,6 +114,7 @@ private[akka] class Shard(
   import ShardRegion.{ handOffStopperProps, EntityId, Msg, Passivate, ShardInitialized }
   import ShardCoordinator.Internal.{ HandOff, ShardStopped }
   import Shard.{ State, RestartEntity, EntityStopped, EntityStarted }
+  import Shard.{ ShardQuery, GetCurrentShardState, CurrentShardState, GetShardStats, ShardStats }
   import akka.cluster.sharding.ShardCoordinator.Internal.CoordinatorMessage
   import akka.cluster.sharding.ShardRegion.ShardRegionCommand
   import settings.tuningParameters._
@@ -129,6 +143,7 @@ private[akka] class Shard(
     case msg: CoordinatorMessage                 ⇒ receiveCoordinatorMessage(msg)
     case msg: ShardCommand                       ⇒ receiveShardCommand(msg)
     case msg: ShardRegionCommand                 ⇒ receiveShardRegionCommand(msg)
+    case msg: ShardQuery                         ⇒ receiveShardQuery(msg)
     case msg if extractEntityId.isDefinedAt(msg) ⇒ deliverMessage(msg, sender())
   }
 
@@ -145,6 +160,11 @@ private[akka] class Shard(
     case HandOff(`shardId`) ⇒ handOff(sender())
     case HandOff(shard)     ⇒ log.warning("Shard [{}] can not hand off for another Shard [{}]", shardId, shard)
     case _                  ⇒ unhandled(msg)
+  }
+
+  def receiveShardQuery(msg: ShardQuery): Unit = msg match {
+    case GetCurrentShardState ⇒ sender() ! CurrentShardState(shardId, refById.keySet)
+    case GetShardStats        ⇒ sender() ! ShardStats(shardId, state.entities.size)
   }
 
   def handOff(replyTo: ActorRef): Unit = handOffStopper match {
--- a/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
+++ b/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
@@ -4,17 +4,11 @@
 package akka.cluster.sharding
 
 import java.net.URLEncoder
-import scala.collection.immutable
-import akka.actor.Actor
-import akka.actor.ActorLogging
-import akka.actor.ActorRef
-import akka.actor.ActorSelection
-import akka.actor.Address
-import akka.actor.Deploy
-import akka.actor.PoisonPill
-import akka.actor.Props
-import akka.actor.RootActorPath
-import akka.actor.Terminated
+import akka.pattern.AskTimeoutException
+import akka.util.Timeout
+
+import akka.pattern.{ ask, pipe }
+import akka.actor._
 import akka.cluster.Cluster
 import akka.cluster.ClusterEvent.ClusterDomainEvent
 import akka.cluster.ClusterEvent.CurrentClusterState
@@ -23,6 +17,11 @@ import akka.cluster.ClusterEvent.MemberRemoved
 import akka.cluster.ClusterEvent.MemberUp
 import akka.cluster.Member
 import akka.cluster.MemberStatus
+
+import scala.collection.immutable
+import scala.concurrent.duration._
+import scala.concurrent.Future
+import scala.reflect.ClassTag
 
 /**
  * @see [[ClusterSharding$ ClusterSharding extension]]
@@ -167,19 +166,26 @@ object ShardRegion {
    */
   def gracefulShutdownInstance = GracefulShutdown
 
-  /*
+  sealed trait ShardRegionQuery
+
+  /**
    * Send this message to the `ShardRegion` actor to request for [[CurrentRegions]],
    * which contains the addresses of all registered regions.
-   * Intended for testing purpose to see when cluster sharding is "ready".
+   * Intended for testing purpose to see when cluster sharding is "ready" or to monitor
+   * the state of the shard regions.
    */
-  @SerialVersionUID(1L) final case object GetCurrentRegions extends ShardRegionCommand
+  @SerialVersionUID(1L) final case object GetCurrentRegions extends ShardRegionQuery
 
+  /**
+   * Java API:
+   */
   def getCurrentRegionsInstance = GetCurrentRegions
 
   /**
    * Reply to `GetCurrentRegions`
    */
   @SerialVersionUID(1L) final case class CurrentRegions(regions: Set[Address]) {
+
     /**
      * Java API
      */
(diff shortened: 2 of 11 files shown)


johanandren:
Wat!? 
wat-1414024905

johanandren:
"It works on my computer"

johanandren:
PLS BUILD

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
```diff
@@ -50,6 +50,10 @@ private[akka] object Shard {
    */
   @SerialVersionUID(1L) final case class EntityStopped(entityId: EntityId) extends StateChange
 
+  @SerialVersionUID(1L) final case object GetCurrentShardState
```
patriknw: this is in `private[akka] object Shard`, I think they should be in `ShardRegions` companion

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
```diff
@@ -147,6 +152,10 @@ private[akka] class Shard(
     case _                  ⇒ unhandled(msg)
   }
 
+  def receiveCurrentEntityQuery(ref: ActorRef): Unit = {
+    ref ! CurrentShardState(shardId, refById.keySet)
```
patriknw: is refById url encoded? The ids we return should be the original ids.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
```diff
@@ -50,6 +50,10 @@ private[akka] object Shard {
    */
   @SerialVersionUID(1L) final case class EntityStopped(entityId: EntityId) extends StateChange
 
+  @SerialVersionUID(1L) final case object GetCurrentShardState
+
+  @SerialVersionUID(1L) final case class CurrentShardState(shardId: ShardRegion.ShardId, entityIds: Set[EntityId])
```
patriknw: also add java api `getEnityIds: java.util.Set[String]`

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -179,7 +175,21 @@ object ShardRegion {
   /**
    * Reply to `GetCurrentRegions`
    */
-  @SerialVersionUID(1L) final case class CurrentRegions(regions: Set[Address]) {
+  @SerialVersionUID(2L) final case class CurrentRegions(regionActorPaths: Set[ActorPath]) {
```
patriknw: this is unfortunately not binary backwards compatible (and I don't think we should squeeze it into 2.4.0)

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -190,6 +200,35 @@ object ShardRegion {
 
   }
 
+  /**
+   * Send this message to the `ShardRegion` actor to request for
+   * [[CurrentShardRegionState]] which describes the current running shards and the
+   * shard entities on running on those in the region.
```
patriknw: that sentence is strange

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -190,6 +200,35 @@ object ShardRegion {
...
+    /**
+     * Java API
+     */
+    def getShards(): java.util.Set[ShardState] = {
+      import scala.collection.JavaConverters._
+      shards.asJava
+    }
+  }
+
+  @SerialVersionUID(1L) final case class ShardState(shardId: ShardId, entityIds: Set[EntityId])
```
patriknw: java api for `entityIds`

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -458,6 +498,19 @@ class ShardRegion(
     }
   }
 
+  def receiveRegionStateQuery(ref: ActorRef): Unit = {
```
patriknw: s/ref/replyTo/

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -458,6 +498,19 @@ class ShardRegion(
...
+    import akka.pattern.{ ask, pipe }
+    import scala.concurrent.duration._
+    implicit val timeout: Timeout = 1.seconds
```
patriknw: I agree that this is not important enough to be defined in config, but I think we could increase it to at least 3.seconds ?

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -458,6 +498,19 @@ class ShardRegion(
...
+    Future.sequence(shards.map {
+      case (shardId, ref) ⇒
+        (ref ? Shard.GetCurrentShardState).mapTo[Shard.CurrentShardState]
+    }).map { shardStates ⇒
+      CurrentShardRegionState(shardStates.map(state ⇒
+        ShardRegion.ShardState(state.shardId, state.entityIds)).toSet)
+    }.pipeTo(ref)
```
patriknw: how should we handle timeout? sending back Status.Failure is too general in my opinion, perhaps we need another reply message for the the error case?

patriknw: or should we just reply with empty ShardState?

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
patriknw: I'm thinking about what API we want to provide. We don't want one single query to get all regions and all their entities, because that would result in huge messages (when millions of entities)). With similar reasoning I'm not sure about handing out the paths (refs) to the regions in this query, because then user would continue that by asking all regions for all entities and we would send a lot of data over the network.

Instead I think we should encourage local tasks. Let's say the user want to get an aggregated summary of number of entities on each node. Then it would be enough to hand out the Address of the regions here, and then they can have their own aggregation actor on each node that queries the local region for the entity information.

Perhaps not convenient but makes user's think before abusing.

wdyt?

Note that coordinator and region actors are for one entity type, so user have to know the entity types anyway.

akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingGetStateSpec.scala
```diff
@@ -0,0 +1,182 @@
...
+        }
+      }
+
+      enterBarrier("sharded actors started")
+
+    }
+
+    "get shard state" in {
+      within(10.seconds) {
+        awaitAssert {
```
patriknw: It's often good to have a separate TestProbe for the things inside the awaitAssert, to avoid spilling over messages from one attempt to the next. Also note that the within defined here will currently be used for the things inside the awaitAssert, such as receiveWhile. That is often not what you intend, but with a separate TestProbe you will also get separate within timeouts for free.

akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingSpec.scala
```diff
@@ -323,7 +323,7 @@ abstract class ClusterShardingSpec(config: ClusterShardingSpecConfig) extends Mu
         expectMsg(2)
 
         region ! GetCurrentRegions
-        expectMsg(CurrentRegions(Set(Cluster(system).selfAddress)))
+        expectMsgType[CurrentRegions].regions shouldEqual Set(Cluster(system).selfAddress)
```
patriknw: we are slowly moving to using `===` instead

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
johanandren: I didn't make it public since you would anyways go through the `ShardRegion` with the messages `GetShardRegionState`/`CurrentShardRegionState`, considered it an internal impl detail. Motivation was partly that there is no public access to the actual `Shard` actor anywhere else.

You think it should be a public API?

patriknw: sorry, I missed where it was used. good as is

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
johanandren: I'm pretty sure it is not url encoded, `refById` is a `Map[EntityId, ActorRef]`

patriknw: good, thanks for checking

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
johanandren: This is pretty much how I have thought about it and what is in the PR right now, it is possible to query the `ShardRegion` about what regions there are (internally forwarded request to the `ShardCoordinator`), which returns `ActorPath`s of all regions in the cluster, so that you (the user) can talk to each `ShardRegion` if you want.

Then there is the separate `GetCurrentShardRegionState`/`CurrentShardRegionState` which you can send to _one_ `ShardRegion` to get info about it's running entities. So, we do not provide the aggregate for the entire cluster ourselves - the user will have to do it - and as you say, hopefully think twice about it in their use case.

patriknw: My point is that `ActorPath` is too easy to use. They will send `GetCurrentShardRegionState` to the actor paths and do the aggregation at one central place (and thereby sending all ids over the net). I think this query should be as before, i.e. handing out addresses.

Then we could add another convenient aggregation query that return the full summary of everything, but only counts.

johanandren: Ah, ok, I didn't understand that. So, `CurrentRegions` back to how it was, and no obvious way for the users to figure out where all regions are through this API.

So, if I understand it correctly, the summary would be something like `CurrentShardStats(stats: Seq[(ShardId, StatsAboutShard-with-enitty-count-etc)]])`?

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
johanandren: I think a specific message is better.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
ktoso: I like the "only counts" style of response here. // Can definitely see how people would abuse it if given full information.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
johanandren: Thinking a bit more about it, should that be any kind of failure getting the states? 

I'm thinking log the failure at warning and send that back. a `GetShardRegionStateFailure` since the sender can't do anything about it anyways. Bad idea?

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
hepin1989: add a query for this,```Query("{regex}")```then get the ``CurrentRegions``` to ```RegionsResponse(regionActorPaths: Set[ActorPath],queryHint:Any)```?sorry I don't familiar with sharding ,but I think a query for a dedicated path/sub-path would be  great.

patriknw: ok, we should not go over board with this. The presented use cases are rather limited so far.
Let's add the CurrentShardStats, other things are possible with user implemented "map reduce" style actors on the local nodes.

patriknw: > no obvious way for the users to figure out where all regions are 

well, it's rather obvious (but not too easy for abuse), there is one region actor on each node (address) that is returned by CurrentRegions, so the user can send a message to that node to some user defined actor and then that actor can access the local regions via the extension as usual

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
patriknw: I don't think `ShardState(shardId, Set.empty[EntityId])` is too bad either, or if we want be more explicit that the timeout is different from empty we could have `ShardState(shardId: ShardId, entityIds: Option[Set[EntityId]])`

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
johanandren: Internal message, so no Java api

patriknw:
@johanandren  This looks good. Need some kind of mention in the rst docs also.

johanandren:
Two things I'm unsure about:

* Timeout for the cluster region stats requests since it will be fetched from across the cluster - should it be configurable? Hardcoded to 10s for now.
* Should the `GetClusterRegionStats`/`ClusterRegionStats` have protobuf serialization since they go across the wire?

patriknw:
1) Should the timeout be part of the request message?
2) yes, protobuf please

johanandren:
I think timeout as part of message is a good idea. Will do that.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala
```diff
@@ -147,6 +162,19 @@ private[akka] class Shard(
...
+    case GetShardStats        ⇒ replyToShardStats(sender())
+  }
+
+  def replyToCurrentEntityQuery(ref: ActorRef): Unit = {
+    ref ! CurrentShardState(shardId, refById.keySet)
+  }
+
+  def replyToShardStats(ref: ActorRef): Unit = {
+    ref ! ShardStats(shardId, state.entities.size)
+  }
```
patriknw: I'd inline those two methods

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -190,6 +193,80 @@ object ShardRegion {
...
+   * Intended for testing purpose to see when cluster sharding is "ready" or to monitor
+   * the state of the shard regions.
+   */
+  @SerialVersionUID(1L) case class GetClusterShardingStats(timeout: FiniteDuration) extends ShardRegionQuery
+
+  /**
+   * Reply to [[GetClusterShardingStats]], contains statistics about all the sharding regions
+   * in the cluster.
+   */
+  @SerialVersionUID(1L) final case class ClusterShardingStats(regions: Map[Address, ShardRegionStats])
```
patriknw: Java API for the `Map`

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -190,6 +193,80 @@ object ShardRegion {
...
+   * Intended for testing purpose to see when cluster sharding is "ready" or to monitor
+   * the state of the shard regions.
+   *
+   * For the statistics for the entire cluster, see [[GetClusterShardingStats$]].
+   */
+  @SerialVersionUID(1L) case object GetShardRegionStats extends ShardRegionQuery
+
+  def getRegionStatsInstance = GetShardRegionStats
+
+  @SerialVersionUID(1L) final case class ShardRegionStats(stats: Map[ShardId, Int])
```
patriknw: Java API for the Map

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -190,6 +193,80 @@ object ShardRegion {
...
+   * which contains statistics about the currently running sharded entities in the
+   * entire region.
+   * Intended for testing purpose to see when cluster sharding is "ready" or to monitor
+   * the state of the shard regions.
+   *
+   * For the statistics for the entire cluster, see [[GetClusterShardingStats$]].
+   */
+  @SerialVersionUID(1L) case object GetShardRegionStats extends ShardRegionQuery
+
+  def getRegionStatsInstance = GetShardRegionStats
```
patriknw: add

    /** Java API */

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -419,13 +497,26 @@ class ShardRegion(
       gracefulShutdownInProgress = true
       sendGracefulShutdownToCoordinator()
 
+    case _ ⇒ unhandled(cmd)
```
patriknw: good

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -458,6 +549,53 @@ class ShardRegion(
...
+  def askAllShards[T: ClassTag](msg: Any): Future[Seq[(ShardId, T)]] = {
+    implicit val timeout: Timeout = 3.seconds
+    Future.sequence(shards.toSeq.map {
+      case (shardId, ref) ⇒ (ref ? msg).mapTo[T].map(t ⇒ (shardId, t))
+    })
+  }
+
+  def replyToClusterShardingStats(waitMax: FiniteDuration, ref: ActorRef): Unit = {
+    implicit val timeout: Timeout = waitMax
+    val regionsAndSelf = regions.keySet + self
```
patriknw: A region does not know about all other regions unless it has asked the coordinator for the `GetShardHome`, which will be done when the shard is first used in this specific region. I think the stats must be aggregated via the coordinator to be correct.

akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingGetStateSpec.scala
johanandren: It doesn't show in the snippet here, but I did update according to this comment!

patriknw:
LGTM, great work Johan! This looks ready to me, so please squash to one commit.

PR #18548, opened 2015-09-24
Fix various interpreter issues, add extensive port transition tests

drewhk:


drewhk:
New fixes, new tests, fixed benchmark compilation.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -203,19 +213,19 @@ private[stream] final class GraphInterpreter(
...
-  // connection event has been processed. The inAvailable and outAvailable arrays usually protect access to this
+  // connection event has been processed. The inStates and outStates arrays usually protect access to this
   // field while it is in transient state.
   val connectionStates = Array.fill[Any](assembly.connectionCount)(Empty)
 
   // Indicates whether the input port is pullable. After pulling it becomes false
   // Be aware that when inAvailable goes to false outAvailable does not become true immediately, only after
   // the corresponding event in the queue has been processed
-  val inAvailable = Array.fill[Boolean](assembly.connectionCount)(true)
+  val inStates = Array.fill[PortState](assembly.connectionCount)(Available)
```
drewhk: The inStates and outStates arrays probably can be packed together in the same array, encoding the two sides as bit-fields of an Int:
 - lowest two bits are encoding the inStates state
 - next two bits are encoding the outStates state
 - additional flags can encode the PushCompleted and CompletedWithValue states instead of wrapper objects (probably one flag for "hasElement" so it can replace isInstanceOf[HasElement], and another flag if a pushcomplete is in flight)

drewhk:
Will need to regain some performance, but let's get this correct first.

rkuhn:
Yes, this does look like the result of some painful debugging! I have not yet understood all single changes, but I trust you.

akka-bench-jmh-dev/src/main/scala/akka/stream/InterpreterBenchmark.scala
```diff
@@ -26,34 +26,33 @@ class InterpreterBenchmark {
...
+        val identities = Vector.fill(numberOfIds)(new Identity[Int])
+        val source = new GraphDataSource("source", data100k)
+        val sink = new GraphDataSink[Int]("sink", data100k.size)
+
+        val b = builder(identities: _*)
+          .connect(source, identities.head.in)
+          .connect(identities.last.out, sink)
+
+        // FIXME: This should not be here, this is pure setup overhead
+        for (i <- (0 until identities.size - 1)) {
```
drewhk: these benchmarks must be fixed, these initialization steps should not be measured here. Even if we want to measure the initialization speed, the above is simply not part of the initialization since it is just an ugly test tool to create the GraphAssembly (i.e. it emulates a fuser)

PR #18560, opened 2015-09-25
=str,htp clean up build warnings

rkuhn:
- explicitly provide Unit values and place parens around tuple creation
- remove structural type usage in TestUtils
- fix Java double-casts
- use unused Java values by asserting their non-nullness
- work around inability to place case class in trait (scripted test)

The remaining warnings about using private types in public methods are
bogus as reported in https://issues.scala-lang.org/browse/SI-9490.

### DIFF
--- a/akka-docs-dev/rst/java/code/docs/http/javadsl/HttpClientExampleDocTest.java
+++ b/akka-docs-dev/rst/java/code/docs/http/javadsl/HttpClientExampleDocTest.java
@@ -16,6 +16,7 @@ import akka.http.javadsl.model.*;
 import akka.http.javadsl.Http;
 import scala.util.Try;
 
+@SuppressWarnings("unused")
 public class HttpClientExampleDocTest {
 
     // compile only test
--- a/akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/RejectionHandlerExamplesSpec.scala
+++ b/akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/RejectionHandlerExamplesSpec.scala
@@ -23,13 +23,13 @@ object MyRejectionHandler {
         complete(HttpResponse(BadRequest, entity = "No cookies, no service!!!"))
       }
       .handle { case AuthorizationFailedRejection ⇒
-        complete(Forbidden, "You're out of your depth!")
+        complete((Forbidden, "You're out of your depth!"))
       }
       .handleAll[MethodRejection] { methodRejections ⇒
         val names = methodRejections.map(_.supported.name)
-        complete(MethodNotAllowed, s"Can't do that! Supported: ${names mkString " or "}!")
+        complete((MethodNotAllowed, s"Can't do that! Supported: ${names mkString " or "}!"))
       }
-      .handleNotFound { complete(NotFound, "Not here!") }
+      .handleNotFound { complete((NotFound, "Not here!")) }
       .result()
 
   object MyApp extends App {
--- a/akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/directives/PathDirectivesExamplesSpec.scala
+++ b/akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/directives/PathDirectivesExamplesSpec.scala
@@ -16,7 +16,7 @@ class PathDirectivesExamplesSpec extends RoutingSpec {
 
   //# path-dsl
   // matches /foo/
-  path("foo" /)
+  path("foo"./)
 
   // matches e.g. /foo/123 and extracts "123" as a String
   path("foo" / """\d+""".r)
--- a/akka-docs-dev/rst/scala/code/docs/stream/io/StreamFileDocSpec.scala
+++ b/akka-docs-dev/rst/scala/code/docs/stream/io/StreamFileDocSpec.scala
@@ -32,6 +32,7 @@ class StreamFileDocSpec extends AkkaSpec(UnboundedMailboxConfig) {
     //#file-source
     import akka.stream.io._
     //#file-source
+    Thread.sleep(0) // needs a statement here for valid syntax and to avoid "unused" warnings
   }
 
   {
--- a/akka-http-core/src/main/java/akka/http/impl/util/Util.java
+++ b/akka-http-core/src/main/java/akka/http/impl/util/Util.java
@@ -23,17 +23,17 @@ public abstract class Util {
     // needed to provide covariant conversions that the Java interfaces don't provide automatically.
     // The alternative would be having to cast around everywhere instead of doing it here in a central place.
     public static <U, T extends U> Option<U> convertOption(scala.Option<T> o) {
-        return (Option<U>)(Option) akka.japi.Option.fromScalaOption(o);
+        return (Option<U>)(Object) akka.japi.Option.fromScalaOption(o);
     }
     @SuppressWarnings("unchecked") // no support for covariance of Publisher in Java
     // needed to provide covariant conversions that the Java interfaces don't provide automatically.
     // The alternative would be having to cast around everywhere instead of doing it here in a central place.
     public static <U, T extends U> Source<U, scala.Unit> convertPublisher(Source<T, scala.Unit> p) {
-        return (Source<U, scala.Unit>)(Source) p;
+        return (Source<U, scala.Unit>)(Object) p;
     }
     @SuppressWarnings("unchecked")
     public static <T, U extends T> Source<U, scala.Unit> upcastSource(Source<T, scala.Unit> p) {
-        return (Source<U, scala.Unit>)(Source) p;
+        return (Source<U, scala.Unit>)(Object) p;
     }
     @SuppressWarnings("unchecked")
     public static scala.collection.immutable.Map<String, String> convertMapToScala(Map<String, String> map) {
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/ContentRange.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/ContentRange.java
@@ -27,7 +27,7 @@ public abstract class ContentRange {
     }
     @SuppressWarnings("unchecked")
     public static ContentRange create(long first, long last, Option<Long> instanceLength) {
-        return ContentRange$.MODULE$.apply(first, last, ((Option<Object>) (Option) instanceLength).asScala());
+        return ContentRange$.MODULE$.apply(first, last, ((Option<Object>) (Object) instanceLength).asScala());
     }
     public static ContentRange createUnsatisfiable(long length) {
         return new akka.http.scaladsl.model.ContentRange.Unsatisfiable(length);
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
@@ -126,6 +126,7 @@ private[http] object OutgoingConnectionBlueprint {
           case (ctx, `terminationBackchannelInput`) ⇒
             ctx.finish()
             SameState
+          case (_, _) ⇒ SameState
         },
         onUpstreamFailure = defaultCompletionHandling.onUpstreamFailure)
     }
@@ -220,6 +221,7 @@ private[http] object OutgoingConnectionBlueprint {
               ctx.fail(new ResponseParsingError(parser.onPull().asInstanceOf[ErrorOutput]))
             }
             SameState
+          case (_, _) ⇒ SameState
         },
         onUpstreamFailure = defaultCompletionHandling.onUpstreamFailure)
     }
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -204,9 +204,13 @@ private[http] object HttpServerBluePrint {
             }
 
           case (ctx, _, OneHundredContinue) ⇒
-            assert(requestStart.expect100ContinueResponsePending)
+            require(requestStart.expect100ContinueResponsePending)
             ctx.emit(ResponseRenderingContext(HttpResponse(StatusCodes.Continue)))
             requestStart = requestStart.copy(expect100ContinueResponsePending = false)
+            SameState
+
+          case (ctx, _, msg) ⇒
+            ctx.fail(new IllegalStateException(s"unexpected message of type [${msg.getClass}], expecting only HttpResponse or OneHundredContinue"))
             SameState
         }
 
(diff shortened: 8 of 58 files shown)


akka-docs-dev/rst/java/code/docs/http/javadsl/HttpClientExampleDocTest.java
```diff
@@ -55,6 +56,7 @@ public void testHostLevelExample() {
         .via(poolClientFlow)
         .runWith(Sink.<Pair<Try<HttpResponse>, Integer>>head(), materializer);
     //#host-level-example
+    assert(responseFuture != null);
```
patriknw: could also have been suppressed with annotation?

akka-docs-dev/rst/scala/code/docs/stream/io/StreamFileDocSpec.scala
```diff
@@ -32,6 +32,7 @@ class StreamFileDocSpec extends AkkaSpec(UnboundedMailboxConfig) {
     //#file-source
     import akka.stream.io._
     //#file-source
+    Thread.sleep(0)
```
patriknw: wat?

akka-http-core/src/main/java/akka/http/javadsl/model/ContentRange.java
```diff
@@ -27,7 +27,7 @@ public static ContentRange create(long first, long last, long instanceLength) {
     }
     @SuppressWarnings("unchecked")
     public static ContentRange create(long first, long last, Option<Long> instanceLength) {
-        return ContentRange$.MODULE$.apply(first, last, ((Option<Object>) (Option) instanceLength).asScala());
+        return ContentRange$.MODULE$.apply(first, last, ((Option<Object>) (Object) instanceLength).asScala());
```
patriknw: it's too bad that we still use $.MODULE$ syntax, we could define an internal accessor class in scaladsl for such things

akka-http-core/src/main/java/akka/http/impl/util/Util.java
```diff
@@ -23,17 +23,17 @@
     // needed to provide covariant conversions that the Java interfaces don't provide automatically.
     // The alternative would be having to cast around everywhere instead of doing it here in a central place.
     public static <U, T extends U> Option<U> convertOption(scala.Option<T> o) {
-        return (Option<U>)(Option) akka.japi.Option.fromScalaOption(o);
+        return (Option<U>)(Object) akka.japi.Option.fromScalaOption(o);
```
drewhk: er, why?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -204,10 +204,14 @@ private[http] object HttpServerBluePrint {
             }
 
           case (ctx, _, OneHundredContinue) ⇒
-            assert(requestStart.expect100ContinueResponsePending)
+            require(requestStart.expect100ContinueResponsePending)
```
drewhk: good

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -126,6 +126,7 @@ private[http] object OutgoingConnectionBlueprint {
           case (ctx, `terminationBackchannelInput`) ⇒
             ctx.finish()
             SameState
+          case (_, _) ⇒ SameState
```
patriknw: and this is alright? we don't silently ignore any programming mistakes? alternative is to throw IllegalStateException?

akka-stream-testkit/src/test/scala/akka/stream/testkit/TestUtils.scala
```diff
@@ -8,14 +8,24 @@ import java.nio.channels.DatagramChannel
...
-  import scala.language.reflectiveCalls
-  // Structural type needed since DatagramSocket and ServerSocket has no common ancestor apart from Object
-  type GeneralSocket = {
+  trait GeneralSocket extends Any {
     def bind(sa: SocketAddress): Unit
     def close(): Unit
-    def getLocalPort(): Int
+    def getLocalPort: Int
+  }
+  implicit class GeneralDatagramSocket(val s: DatagramSocket) extends GeneralSocket {
```
drewhk: nice

drewhk:
LGTM

akka-http-tests-java8/src/test/java/docs/http/javadsl/server/HandlerExampleDocTest.java
```diff
@@ -75,6 +73,7 @@ public void testCalculator() {
             //#handler2
             final Handler2<Integer, Integer> multiply =
                 new Handler2<Integer, Integer>() {
+            		static final long serialVersionUID = 1L;
                     @Override
```
patriknw: indentation?

akka-docs-dev/rst/java/code/docs/http/javadsl/HttpClientExampleDocTest.java
rkuhn: it’s a line either way; do you have a strong preference? or anyone else?

akka-docs-dev/rst/scala/code/docs/stream/io/StreamFileDocSpec.scala
rkuhn: There needs to be a statement here, otherwise scalariform chokes on the (admittedly also invalid) syntax. Will add a comment.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -250,7 +250,7 @@ private[stream] object ActorGraphInterpreter {
           rejectAdditionalSubscriber(subscriber, s"${Logging.simpleName(this)}")
       }
 
-    def exposedPublisher(publisher: ActorPublisher[Any]): Unit = {
+    private[akka] def exposedPublisher(publisher: ActorPublisher[Any]): Unit = {
```
patriknw: why? isn't all this internal?

patriknw:
LGTM, aside from some nitpicks

akka-docs-dev/rst/java/code/docs/http/javadsl/HttpClientExampleDocTest.java
patriknw: the intention of that line is not very clear, without a comment (so someone might refactor it away)
suppress warning is just one keyboard click away in eclipse

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
rkuhn: ah, this was a test balloon, will remove: the reason is https://issues.scala-lang.org/browse/SI-9490

akka-docs-dev/rst/java/code/docs/http/javadsl/HttpClientExampleDocTest.java
ktoso: no strong preference from my side here; can be annotation or assert

akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/directives/PathDirectivesExamplesSpec.scala
```diff
@@ -16,7 +16,7 @@ class PathDirectivesExamplesSpec extends RoutingSpec {
 
   //# path-dsl
   // matches /foo/
-  path("foo" /)
+  path("foo"./)
```
ktoso: ok, probably clearer

akka-docs-dev/rst/scala/code/docs/stream/io/StreamFileDocSpec.scala
ktoso: maybe just `()` with comment then? `sleep(0)` is super weird :)

akka-http-tests-java8/src/main/java/akka/http/javadsl/server/examples/simple/SimpleServerApp8.java
```diff
@@ -40,6 +40,7 @@ public void test() {
     @Override
     public Route createRoute() {
         Handler addHandler = new Handler() {
+        	static final long serialVersionUID = 1L;
```
ktoso: slightly weird indent

akka-docs-dev/rst/java/code/docs/http/javadsl/HttpClientExampleDocTest.java
rkuhn: The disadvantage of the annotation is that it would need extra boilerplate in order to not show up in the scraped doc snippet.

ktoso:
Some minor things where commented, LGTM in general though

akka-http-tests-java8/src/main/java/akka/http/javadsl/server/examples/simple/SimpleServerApp8.java
rkuhn: AARGH, TABS VERSUS SPACES! (sorry :-) )

akka-docs-dev/rst/java/code/docs/http/javadsl/HttpClientExampleDocTest.java
patriknw: in doc tests you would place it in outer scope, here on `testHostLevelExample`

drewhk: I prefer the annotation, too, these assertions are rather mysterious. If nothing else works, a method "supressUnusedWarning" that just does this assert(x != null) is more readable.

akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/directives/PathDirectivesExamplesSpec.scala
drewhk: I guess this was postfix ops warning?

ktoso: yup

patriknw: is that really good in documentation? would you write it like that in a real app?

rkuhn: Yes, I certainly would: this matches our usage of the duration DSL as well, where we have added the dot in (hopefully) all cases. Postfix operators are frowned upon for good reasons and importing the needed implicit just in the right scopes would be too bulky. What’s wrong with the line as it is now?

akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/RejectionHandlerExamplesSpec.scala
```diff
@@ -23,13 +23,13 @@ object MyRejectionHandler {
         complete(HttpResponse(BadRequest, entity = "No cookies, no service!!!"))
       }
       .handle { case AuthorizationFailedRejection ⇒
-        complete(Forbidden, "You're out of your depth!")
+        complete((Forbidden, "You're out of your depth!"))
```
2beaucoup: What's the reasoning behind these?

drewhk:
Truly heroic effort, LGTM.

akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/directives/PathDirectivesExamplesSpec.scala
patriknw: it's no problem with postfix when used within `( )` as here and I thought it was a convention to use it like that in the routing dsl.

I'm fine with `"foo"./`

akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/RejectionHandlerExamplesSpec.scala
rkuhn: I don’t know why anyone would want to use auto-tupling, so I cannot tell you why the line previously was as it was ;-) (seriously, though: the associated warning has caught real bugs before and therefore we abide by its imposed restrictions).

akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/directives/PathDirectivesExamplesSpec.scala
rkuhn: Unfortunately the linter is not smart enough to recognize that, which was the reason for abandoning postfix notation for durations as well.

akka-http-core/src/main/java/akka/http/impl/util/Util.java
rkuhn: Because I also don’t want to have Java warnings: `Option` is a raw type.

akka-http-core/src/main/java/akka/http/javadsl/model/ContentRange.java
rkuhn: maybe I’m missing some context: this should be internal code only, where we can do ugly things if they leave the DSL in a cleaner shape

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
rkuhn: we would previously have thrown MatchError and terminated the ActorSystem, but yes, I can future-proof this one as well

rkuhn: ah, wrong context: this is actually fine, the compiler is just not smart enough

akka-http-core/src/main/java/akka/http/javadsl/model/ContentRange.java
patriknw: is the IDE clever enough to understand this nowadays?

akka-http-tests-java8/src/test/java/docs/http/javadsl/server/HandlerExampleDocTest.java
rkuhn: I honestly have no idea what Eclipse has been doing here …

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
drewhk: add a comment there for the future though

akka-http-core/src/main/java/akka/http/javadsl/model/ContentRange.java
rkuhn: Works for me (tm) (most of the time).

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
rkuhn: viewed in context it is pretty obvious that there are only these two inlets defined

Issue #18533, opened 2015-09-22
Setting akka.http.server.parsing.illegal-header-warnings = off doesn't work


analytically:
When using the following config:

```
akka {
  http {
    server.parsing.illegal-header-warnings = off
    client.parsing.illegal-header-warnings = off
  }
}
```

I still see

```[WARN] a.actor.ActorSystemImpl - Illegal response header: Illegal 'expires' header: Illegal weekday in date 1990-01-01T00:00:00: is 'Fri' but should be 'Mon'```

Using Akka HTTP 1.0.

**Comment by ktoso** [comment]
Thanks for reporting! I'll have a look right away :)

**Comment by ktoso** [comment]
Actually it seems to work properly, I've verified both with current latest version as well as the 1.0 version of Akka Http. Are you sure the settings are actually being picked up? Where do you put them, to application.conf (if so, are you using it? also known as: not supplying a custom conf that ignores this file?)?

(you can check it yourself using this [branch](https://github.com/ktoso/akka/tree/wip-illegal-headers-warning-bad-ktoso) by `sbt shell; project akka-http-tests-experimental; test:run`)

**Comment by analytically** [comment]
Try this:

```
import akka.actor.ActorSystem
import akka.http.scaladsl.Http
import akka.http.scaladsl.client.RequestBuilding._
import akka.http.scaladsl.model.headers.{BasicHttpCredentials, Authorization}
import akka.http.scaladsl.model._
import akka.http.scaladsl.server.directives.UserCredentials
import akka.http.scaladsl.unmarshalling.Unmarshal
import akka.http.scaladsl.util.FastFuture
import akka.stream.ActorMaterializer
import com.typesafe.config.{ Config, ConfigFactory }

object TestClient extends App {
  val testConf: Config = ConfigFactory.parseString("""
                                                     akka {
                                                       http {
                                                         server.parsing.illegal-header-warnings = off
                                                         client.parsing.illegal-header-warnings = off
                                                       }
                                                     }
    akka.log-dead-letters = off""")
  implicit val system = ActorSystem("ServerTest", testConf)
  import system.dispatcher
  implicit val materializer = ActorMaterializer()

  Http().singleRequest(Get(Uri("http://localhost:8000"))).flatMap { response =>
    response.status match {
      case StatusCodes.OK => FastFuture.successful(StatusCodes.OK)
      case _ => Unmarshal(response.entity).to[String].flatMap { entity =>
[...]

PR #18562, opened 2015-09-25
=htc #18533 make configuration of parsing less confusing

ktoso:
resolves https://github.com/akka/akka/issues/18533 
- It's now less DRY, but very clear on how it actually works.
- added to docs info how to reason about these sections, I think it's important to explain explicitly like this (it may be confusing if `singleRequest()` is client or pool (it's pool)).


akka-http-core/src/test/scala/akka/http/impl/engine/client/ClientConfigurationSpec.scala
```diff
@@ -0,0 +1,28 @@
...
+class ClientConfigurationSpec extends AkkaSpec {
+  "Reference configurations" should {
+    for {
+      (first, second) ← "akka.http.client" -> "akka.http.host-connection-pool.client" ::
+        "akka.http.client.parsing" -> "akka.http.host-connection-pool.client.parsing" ::
+        "akka.http.server.parsing" -> "akka.http.client.parsing" ::
+        Nil
+    } s"be consistent for: `$first` and `$second`" in {
+      configShouldBeEqual("akka.http.client.parsing", "akka.http.host-connection-pool.client.parsing")
+    }
```
ktoso: currently complete equality, if we change things only for the server for example this can be changed to check if keys are the same

akka-http-core/src/main/resources/reference.conf
```diff
@@ -88,7 +88,77 @@ akka.http {
     }
 
     # Modify to tweak parsing settings on the server-side only.
-    parsing = ${akka.http.parsing}
+    #
+    # IMPORTANT:
+    # Please note that this section is replicated in `client` and `http-connection-pool`
+    parsing = {
```
jrudolph: Using the equal sign in `parsing =` seems to be non-standard, right?

akka-http-core/src/test/scala/akka/http/impl/engine/client/ClientConfigurationSpec.scala
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (C) 2009-2014 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.impl.engine.client
+
+import akka.stream.testkit.AkkaSpec
+
+class ClientConfigurationSpec extends AkkaSpec {
```
jrudolph: +1 for adding those if the above configuration duplication is really the only way to go.

akka-http-core/src/main/resources/reference.conf
ktoso: good catch thanks :)

jrudolph:
+1 for the additional ScalaDoc pointing to the right configuration section.

akka-http-core/src/main/resources/reference.conf
ktoso: fixed in 2 spots

akka-http-core/src/test/scala/akka/http/impl/engine/client/ClientConfigurationSpec.scala
```diff
@@ -0,0 +1,28 @@
...
+    } s"be consistent for: `$first` and `$second`" in {
+      configShouldBeEqual("akka.http.client.parsing", "akka.http.host-connection-pool.client.parsing")
+    }
+  }
+
+  private def configShouldBeEqual(first: String, second: String): Unit = {
+    val config = system.settings.config
+    try config.getConfig(first) should ===(config.getConfig(second))
+    catch {
+      case t: Throwable ⇒ throw new Exception(s"Config [$first] did not equal [$second]!", t)
```
patriknw: perhaps use `withClue` instead?

patriknw:
LGTM

akka-http-core/src/test/scala/akka/http/impl/engine/client/ClientConfigurationSpec.scala
ktoso: Oh nice! Seems I've gotten behind in my ScalaTest skills, need to catch up on those :)

ktoso:
Previous impl actually broke using parsers detached from the `Http()` object (they can be used independently, so we need the `akka.http.parsing` section in any case).

I went back to the beginning and figured out how to properly implement cascading (as in CSS) overriding of settings and added a number of tests and new docs, gains:

- no duplication, just empty parsing values in the "server / client" sections
- ability to set defaults in `akka.http.parsing` and override it in `akka.http.server.parsing` or akka.http.client`
- the `host-connection-pool` things ARE clients, so the `akka.http.client.parsing` settings apply to them (this is most definitely what people want – when I'm using `Http().singleRequest()` I sure think of it as a client (see the original ticket).
- if indeed different settings for pool and client are needed, the `akka.http.host-connection-pool.client` values do override `akka.http.client`.

Settings rule is simple and consistent: most specific setting wins: 
`akka.http.parsing <  akka.http.client.parsing < akka.http.host-connection-pool.client.parsing`.

Please have a look, I'm rather strongly convinced these are understandable, DRY and powerful semantics that we want

sirthias:
Nice! LGTM.

drewhk:
LGTM, nice one!

PR #18572, opened 2015-09-27
=cdd #18328 Optimize ORSet merge

patriknw:
Some Sunday amusement.

I was able to completely eliminate the overhead of the extra merge that was introduced by #18106. Several other optimizations resulted in speedups of 4-5 times. Overview of the JMH benchmark results in attached [ORSetBenchmark.pdf](https://github.com/akka/akka/files/3105/ORSetBenchmark.pdf) 

Each commit contains one improvement and the commit message the JMH results for the change.

### DIFF
--- a/akka-distributed-data/src/main/scala/akka/cluster/ddata/ORSet.scala
+++ b/akka-distributed-data/src/main/scala/akka/cluster/ddata/ORSet.scala
@@ -10,8 +10,6 @@ import akka.cluster.Cluster
 import akka.cluster.UniqueAddress
 import akka.util.HashCode
 
-// TODO this class can be optimized, but I wanted to start with correct functionality and comparability with riak_dt_orswot
-
 object ORSet {
   private val _empty: ORSet[Any] = new ORSet(Map.empty, VersionVector.empty)
   def empty[A]: ORSet[A] = _empty.asInstanceOf[ORSet[A]]
--- a/akka-distributed-data/src/main/scala/akka/cluster/ddata/ORSet.scala
+++ b/akka-distributed-data/src/main/scala/akka/cluster/ddata/ORSet.scala
@@ -63,8 +63,12 @@ object ORSet {
           }
       }
 
-    val newDots = dropDots(dot.versions.toList, Nil)
-    new VersionVector(versions = TreeMap.empty[UniqueAddress, Long] ++ newDots)
+    if (dot.versions.isEmpty)
+      VersionVector.empty
+    else {
+      val newDots = dropDots(dot.versions.toList, Nil)
+      new VersionVector(versions = VersionVector.emptyVersions ++ newDots)
+    }
   }
 
   /**
@@ -74,20 +78,26 @@ object ORSet {
   private[akka] def mergeCommonKeys[A](commonKeys: Set[A], lhs: ORSet[A], rhs: ORSet[A]): Map[A, ORSet.Dot] = {
     commonKeys.foldLeft(Map.empty[A, ORSet.Dot]) {
       case (acc, k) ⇒
-        val lhsDots = lhs.elementsMap(k).versions
-        val rhsDots = rhs.elementsMap(k).versions
-        val commonDots = lhsDots.filter {
-          case (thisDotNode, v) ⇒ rhsDots.get(thisDotNode).exists(_ == v)
+        val lhsDots = lhs.elementsMap(k)
+        val lhsDotsVersions = lhsDots.versions
+        val rhsDotsVersions = rhs.elementsMap(k).versions
+        if (lhsDotsVersions.size == 1 && rhsDotsVersions.size == 1 && lhsDotsVersions.head == rhsDotsVersions.head) {
+          // one single common dot
+          acc.updated(k, lhsDots)
+        } else {
+          val commonDots = lhsDotsVersions.filter {
+            case (thisDotNode, v) ⇒ rhsDotsVersions.get(thisDotNode).exists(_ == v)
+          }
+          val commonDotsKeys = commonDots.keys
+          val lhsUniqueDots = lhsDotsVersions -- commonDotsKeys
+          val rhsUniqueDots = rhsDotsVersions -- commonDotsKeys
+          val lhsKeep = ORSet.subtractDots(new VersionVector(lhsUniqueDots), rhs.vvector)
+          val rhsKeep = ORSet.subtractDots(new VersionVector(rhsUniqueDots), lhs.vvector)
+          val merged = lhsKeep.merge(rhsKeep).merge(new VersionVector(versions = commonDots))
+          // Perfectly possible that an item in both sets should be dropped
+          if (merged.versions.isEmpty) acc
+          else acc.updated(k, merged)
         }
-        val commonDotsKeys = commonDots.keys
-        val lhsUniqueDots = lhsDots -- commonDotsKeys
-        val rhsUniqueDots = rhsDots -- commonDotsKeys
-        val lhsKeep = ORSet.subtractDots(new VersionVector(lhsUniqueDots), rhs.vvector)
-        val rhsKeep = ORSet.subtractDots(new VersionVector(rhsUniqueDots), lhs.vvector)
-        val merged = lhsKeep.merge(rhsKeep).merge(new VersionVector(versions = commonDots))
-        // Perfectly possible that an item in both sets should be dropped
-        if (merged.versions.isEmpty) acc
-        else acc.updated(k, merged)
     }
   }
 
--- a/akka-distributed-data/src/main/scala/akka/cluster/ddata/ORSet.scala
+++ b/akka-distributed-data/src/main/scala/akka/cluster/ddata/ORSet.scala
@@ -54,20 +54,28 @@ object ORSet {
       remaining match {
         case Nil ⇒ acc
         case (d @ (node, v1)) :: rest ⇒
-          vvector.versions.get(node) match {
-            case Some(v2) if v2 >= v1 ⇒
-              // dot is dominated by version vector, drop it
-              dropDots(rest, acc)
-            case _ ⇒
-              dropDots(rest, d :: acc)
-          }
+          val v2 = vvector.versionAt(node)
+          if (v2 >= v1)
+            // dot is dominated by version vector, drop it
+            dropDots(rest, acc)
+          else
+            dropDots(rest, d :: acc)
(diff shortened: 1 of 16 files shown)


drewhk:
I review this now

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORSet.scala
```diff
@@ -148,6 +207,21 @@ final class ORSet[A] private[akka] (
 
   type T = ORSet[A]
 
+  // Optimization for add/remove followed by merge, which is very common
+  // and merge should just fast forward to the new set
+  private var ancestor: ORSet[A] = null
```
drewhk: Hmm. This is tricky here. So this is a cache field, which is mutable. As far as I see it is safe to share this over threads because it is safe to see a null here (it is just a cache miss). What I was a bit more worried about is the check-and-act code like:

    if (this.ancestor eq null) this else this.ancestor

Although this seems to be safe since if this.ancestor becomes null after the condition have been checked it only results in assigning a null (i.e. just losing this optimization). 

However, this needed quite a bit of thought to see, so I would appreciate this to be documented here. Especially since someone in the future might do an actual "if not-null then dereference" which is no longer thread safe.

drewhk:
Rest of it looks right, I couldn't really understand the "size 1" optimization because I would need to read up on this data structure. The rest of the transformations were understandable without this though.

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORSet.scala
patriknw: It's actually a cache between calls of the same thread, you can think of it as a thread local.
The Replicator actor invokes the user's modify function, which returns a new ORSet instance, with the ancestor field set, see for example the `add` method. Then (in same thread) the Replication calls `merge`, which makes use of the `ancestor` field to perform quick merge.

patriknw:
Thanks for reviewing, @drewhk.

akka-distributed-data/src/main/scala/akka/cluster/ddata/ORSet.scala
drewhk: So it is guaranteed that only one thread will access that field at any time, if I understood this correctly? I would still be more happy if that would be commented :)

patriknw: yes, unless the `ORSet` is used outside of the `Replicator` infrastructure (and I think you are right in that the worst thing that can happen is that performs a full merge). I'll add comment.

akka-distributed-data/src/main/scala/akka/cluster/ddata/VersionVector.scala
```diff
@@ -3,21 +3,37 @@
...
-  val empty: VersionVector = new VersionVector(TreeMap.empty[UniqueAddress, Long])
+  private val emptyVersions: TreeMap[UniqueAddress, Long] = TreeMap.empty
+  val empty: VersionVector = new VersionVector(Left(emptyVersions))
+
   def apply(): VersionVector = empty
+
+  def apply(versions: TreeMap[UniqueAddress, Long]): VersionVector =
+    if (versions.isEmpty) empty
+    else if (versions.size == 1) apply(versions.head._1, versions.head._2)
+    else new VersionVector(Left(versions))
```
viktorklang: I'd probably do this:

~~~scala
versions.size match {
  case 0 => empty
  case 1 => 
   val (k, v) => versions.head
   apply(k, v)
  case _ => new VersionVector(Left(versions))
}
~~~

patriknw: Yeah, that reads nice, but are you sure that it doesn't allocate a tuple?
I think we confirmed that it does in constructor val assignments.

drewhk: Yes, I also think it allocates a tuple. This is the reason I also avoid the unapply style variable assignment in performance sensitive code.

patriknw: Evidence, this is what javap says about `class Foo { val versions: List[(String,String)] = Nil; def doIt(): Unit = { val (k,v) = versions.head}}`. Note `32: new`.

```
public void doIt();
    Code:
       0: aload_0
       1: invokevirtual #17                 // Method versions:()Lscala/collection/immutable/List;
       4: invokevirtual #23                 // Method scala/collection/immutable/List.head:()Ljava/lang/Object;
       7: checkcast     #25                 // class scala/Tuple2
      10: astore_2
      11: aload_2
      12: ifnull        66
      15: aload_2
      16: invokevirtual #28                 // Method scala/Tuple2._1:()Ljava/lang/Object;
      19: checkcast     #30                 // class java/lang/String
      22: astore_3
      23: aload_2
      24: invokevirtual #33                 // Method scala/Tuple2._2:()Ljava/lang/Object;
      27: checkcast     #30                 // class java/lang/String
      30: astore        4
      32: new           #25                 // class scala/Tuple2
      35: dup
      36: aload_3
      37: aload         4
      39: invokespecial #37                 // Method scala/Tuple2."<init>":(Ljava/lang/Object;Ljava/lang/Object;)V
      42: astore        5
      44: aload         5
      46: astore_1
      47: aload_1
      48: invokevirtual #28                 // Method scala/Tuple2._1:()Ljava/lang/Object;
      51: checkcast     #30                 // class java/lang/String
      54: astore        6
      56: aload_1
      57: invokevirtual #33                 // Method scala/Tuple2._2:()Ljava/lang/Object;
      60: checkcast     #30                 // class java/lang/String
      63: astore        7
      65: return
      66: new           #39                 // class scala/MatchError
      69: dup
      70: aload_2
      71: invokespecial #42                 // Method scala/MatchError."<init>":(Ljava/lang/Object;)V
      74: athrow
```

viktorklang: Wow. major wtf. Doesn't head return a tuple?

ktoso: It does, but the `val (..., ....) = actualTuple` allocates another one (sic!). I think this is fixed in Scala 2.12 by the way if I remember correctly (I was excited about a fix recently).

viktorklang: Massive WTF moment.

patriknw: yeah, it's so surprising

akka-distributed-data/src/main/scala/akka/cluster/ddata/VersionVector.scala
```diff
@@ -103,12 +119,60 @@ final case class VersionVector private[akka] (
...
+  /**
+   * INTERNAL API
+   */
+  private[akka] def versionAt(node: UniqueAddress): Long =
+    versions match {
+      case Left(treeMap) ⇒ treeMap.get(node) match {
+        case Some(v) ⇒ v
+        case None    ⇒ Timestamp.Zero
+      }
+      case Right((n, v)) ⇒
```
viktorklang: Doesn't these also allocate Tuples then?

akka-distributed-data/src/main/scala/akka/cluster/ddata/VersionVector.scala
```diff
@@ -184,10 +248,19 @@ final case class VersionVector private[akka] (
...
-    else compare(this.versions.iterator, that.versions.iterator, if (order eq Concurrent) FullOrder else order)
+    else compare(this.versionsIterator, that.versionsIterator, if (order eq Concurrent) FullOrder else order)
   }
 
   /**
+   * INTERNAL API
+   */
+  private[akka] def versionsIterator: Iterator[(UniqueAddress, Long)] =
+    versions match {
+      case Right(tuple)  ⇒ Iterator.single(tuple)
```
viktorklang: Given all these matches, wouldn't it make sense to create a sealed hierarchy with one side operating on tuple and the other on Map?

akka-distributed-data/src/main/scala/akka/cluster/ddata/VersionVector.scala
patriknw: no (I checked something similar with javap). pattern matcher is smarter, there is at least some sanity

akka-distributed-data/src/main/scala/akka/cluster/ddata/VersionVector.scala
patriknw: @viktorklang Do you mean like this: https://github.com/akka/akka/commit/2a65671070a6c1672d53ee8f449b0d50a4873948

viktorklang: Yes, exactly! :)

On Mon, Nov 16, 2015 at 8:54 PM, Patrik Nordwall <notifications@github.com>
wrote:

> In
> akka-distributed-data/src/main/scala/akka/cluster/ddata/VersionVector.scala
> <https://github.com/akka/akka/pull/18572#discussion_r44973533>:
>
> >    }
> >
> >    /**
> > +   * INTERNAL API
> > +   */
> > +  private[akka] def versionsIterator: Iterator[(UniqueAddress, Long)] =
> > +    versions match {
> > +      case Right(tuple)  ⇒ Iterator.single(tuple)
>
> @viktorklang <https://github.com/viktorklang> Do you mean like this:
> 2a65671
> <https://github.com/akka/akka/commit/2a65671070a6c1672d53ee8f449b0d50a4873948>
>
> —
> <https://github.com/akka/akka/pull/18572/files#r44973533>.
>



-- 
Cheers,
√

patriknw: cool, it breaks binary compatibility, but I guess that is something we can do for this experimental module (and the VersionVector is not very much an end-user type)

drewhk:
Anything that keeps this from merging?

patriknw:
yes, I introduced a regression in one of the later refactorings here. I must investigate that first.

patriknw:
I found the regression (small refactoring mistake, hard to find). I will merge after validation.

patriknw:
Refs #18328

Issue #18535, opened 2015-09-22
Problem with unmarshalling json content with  entity(as[T]) 


redlion99:
I'm trying to use the entity router to unmarshall a request with json content,  A compile error said:

```scala
Error:(26, 18) could not find implicit value for parameter um: akka.http.scaladsl.unmarshalling.FromRequestUnmarshaller[Template]
        entity(as[Template]){
                 ^
```

in MarshallingDirectives I found:

```scala
def as[[T]] [(implicit um: FromRequestUnmarshaller[T])]  = um
```

although in SprayJsonSupport only unmarshaller for FromEntityUnmarshaller provided

```scala
implicit def sprayJsonUnmarshaller[T](implicit reader: RootJsonReader[T], mat: Materializer):FromEntityUnmarshaller[T] =
    sprayJsValueUnmarshaller.map(jsonReader[T].read)
```

ktoso:
Hi @redlion99,
did you read about how Spray JSON works? https://github.com/spray/spray-json#providing-jsonformats-for-case-classes

You'll need to provide a format for your case-class, like this: 

```scala
trait MyJsonProtocol extends DefaultJsonProtocol {
  implicit val templateFormat = jsonFormat4(Template)
}
object MyJsonProtocol extends MyJsonProtocol

import MyJsonProtocol._
import spray.json._

// will make use of the implicit templateFormat
as[Template]
```

ktoso:
I'll actually improve the docs file:///Users/ktoso/code/akka/akka-docs-dev/target/sphinx/html/scala/http/common/json-support.html to show a full example how to utilise this

redlion99:
@ktoso  To answer your question, yes, I have import the format for Template, which is in file JsProtocol like this:

             import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
             import akka.http.scaladsl.server.Directives._
             import com.convertlab.channel.email.core.models.JsProtocol._
             import com.convertlab.channel.email.core.models.{EmailTables, Template}

My question is when I use as[Template]  an `implicit um: FromRequestUnmarshaller[T]` is required. But I can not find implementation of a `FromRequestUnmarshaller`, instead I only find  an unmarshaller named `FromEntityUnmarshaller` in `SprayJsonSupport`, I guess that's why I got that error.

redlion99:
by the way, the version I'm using is 1.0  from Maven

<dependency>
            <groupId>com.typesafe.akka</groupId>
[...]

PR #18578, opened 2015-09-28
=htc,doc #18535 improved docs on spray-json usage

ktoso:
Resolves https://github.com/akka/akka/issues/18535

I also increased the max depth of the top-level TOC for streams and http. While this is a crude solution, it already helps a lot with discoverability of docs (I can finally `Ctrl+F "json"`, which previously required digging "oh, because it's _common_"). Proper fixing of the search bar (I do remember about it) will properly address this issue.

### DIFF
--- a/akka-docs-dev/rst/scala/code/docs/http/scaladsl/SprayJsonExampleSpec.scala
+++ b/akka-docs-dev/rst/scala/code/docs/http/scaladsl/SprayJsonExampleSpec.scala
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package docs.http.scaladsl
+
+import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport
+import akka.http.scaladsl.server.Directives
+import org.scalatest.{ Matchers, WordSpec }
+
+class SprayJsonExampleSpec extends WordSpec with Matchers {
+
+  "spray-json example" in {
+    //#example
+    import spray.json._
+
+    // domain model
+    final case class Item(name: String, id: Long)
+    final case class Order(items: List[Item])
+
+    // collect your json format instances into a support trait:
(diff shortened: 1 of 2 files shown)


akka-docs-dev/rst/scala/code/docs/http/scaladsl/SprayJsonExampleSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+    //#example
+    import spray.json._
+
+    // domain model
+    final case class Item(name: String, id: Long)
+    final case class Order(items: List[Item])
+
+    // collect your json format instances into a support trait:
+    trait JsonSupport extends SprayJsonSupport with DefaultJsonProtocol {
+      implicit val itemFormat = jsonFormat2(Item)
```
drewhk: add a comment where jsonFormatX comes from

akka-docs-dev/rst/scala/http/common/json-support.rst
```diff
@@ -21,7 +21,7 @@ that an implicit ``spray.json.RootJsonReader`` and/or ``spray.json.RootJsonWrite
 
 This is how you enable automatic support for (un)marshalling from and to JSON with `spray-json`_:
 
-1. Add a library dependency onto ``"com.typesafe.akka" %% "akka-http-spray-json-experimental" % "1.x"``.
+1. Add a library dependency onto ``"com.typesafe.akka" %% "akka-http-spray-json-experimental" % "@version@"``.
```
drewhk: what is this variable resolving to?

akka-docs-dev/rst/scala/code/docs/http/scaladsl/SprayJsonExampleSpec.scala
ktoso: good one, I'll add that comment in another PR then :)

akka-docs-dev/rst/scala/http/common/json-support.rst
ktoso: the version of streams, so "`1.0`"

PR #18579, opened 2015-09-28
=doc fix typo in sharding docs

ktoso:
//cc @patriknw

### DIFF
--- a/akka-docs/rst/scala/cluster-sharding.rst
+++ b/akka-docs/rst/scala/cluster-sharding.rst
@@ -262,8 +262,7 @@ Note that stopped entities will be started again when a new message is targeted 
 Graceful Shutdown
 -----------------
 
-You can send the message ``ClusterSharding.GracefulShutdown`` message (``ClusterSharding.gracefulShutdownInstance
-in Java) to the ``ShardRegion`` actor to handoff all shards that are hosted by that ``ShardRegion`` and then the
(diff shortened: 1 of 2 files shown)


patriknw:
LGTM, thx

akka-docs/rst/scala/cluster-sharding.rst
```diff
@@ -262,8 +262,7 @@ Note that stopped entities will be started again when a new message is targeted
 Graceful Shutdown
 -----------------
 
-You can send the message ``ClusterSharding.GracefulShutdown`` message (``ClusterSharding.gracefulShutdownInstance
-in Java) to the ``ShardRegion`` actor to handoff all shards that are hosted by that ``ShardRegion`` and then the
```
hepin1989: I think you just need change this(``ClusterSharding.gracefulShutdownInstance
``-in Java)

hepin1989:
I think the ```entity entity identifier ``` may should be ```entity identifier``` too.I just want to send a PR for the typo things too:).

ktoso:
I don't see where the `entity entity identifier` is, @hepin1989? Feel free to submit a PR to fix it in case I'm just blind :)

akka-docs/rst/scala/cluster-sharding.rst
ktoso: I want to remove stuff about Java API from the Scala API docs.
This probably originates from the fact that sharding originally had 1 documentation page for both APIs :)

PR #18503, opened 2015-09-17
+str #18501 improvements and fixes in StreamTestKit (adds toStrict)

ktoso:
Adds `toStrict` as well as a number of other methods for better consistency.
Fixes bug in `expectNextOrError` and `expectNextOrComplete` in which the `expected` element was ignored - any element would have passed the assertion.

PR #18585, opened 2015-09-29
=str fix double-dilation in StreamTestKit

ktoso:
https://github.com/akka/akka/pull/18503/files#r40636956


ktoso:
`expectMsgType ` does dilation already, as noticed by @patriknw

patriknw:
thx

PR #18605, opened 2015-09-30
=htp replace all FlexiMerge/Route with GraphStage

rkuhn:
Early preview of the working principle for review before diving in for real: please check mostly what is in `Websocket.scala`.
- also found a usability improvement (better error message for when ports don’t have handlers at the right time
- will also add nicer errors for when calling `pull()` and friends before `preStart()` (currently just gives a NPE)
- and fixed some more warnings related to https://issues.scala-lang.org/browse/SI-9490


akka-http-core/src/main/scala/akka/http/impl/engine/ws/Websocket.scala
```diff
@@ -120,35 +120,81 @@ private[http] object Websocket {
...
+        private object EagerTerminateOutput extends OutHandler {
+          override def onPull(): Unit = ()
+        }
+
+        private object IgnoreTerminateOutput extends OutHandler {
+          override def onPull(): Unit = ()
+          override def onDownstreamFinish(): Unit = ()
+        }
+
+        private def emit[T](out: Outlet[T], elem: T, andThen: ⇒ Unit): Unit = {
```
drewhk: yeah, nice one. Probably needs an `elems: Iterable[T]` version, as it is quite common. An opposite side (input) version can be `read` which can take either a count of elements, or a condition where to stop reading (for vs. while).

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Websocket.scala
```diff
@@ -120,35 +120,81 @@ private[http] object Websocket {
...
+          if (isAvailable(out)) {
+            push(out, elem)
+            andThen
+          } else {
+            setHandler(out, new Emitting(out, elem, getHandler(out), () ⇒ andThen))
```
drewhk: should we handle if there is another "built-in" handler is already in place there? I.e. require the user to explicitly abort the previous emit for example before installing a new one?

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Websocket.scala
```diff
@@ -120,35 +120,81 @@ private[http] object Websocket {
...
+        private def emit[T](out: Outlet[T], elem: T): Unit = {
+          if (isAvailable(out)) {
+            push(out, elem)
+          } else {
+            setHandler(out, new Emitting(out, elem, getHandler(out), DoNothing))
+          }
         }
+
+        private class Emitting[T](out: Outlet[T], elem: T, next: OutHandler, andThen: () ⇒ Unit) extends OutHandler {
+          if (getHandler(out).isInstanceOf[Emitting[_]])
```
drewhk: ah I see, it is checked. That will probably be an `isInstanceOf[InternalAction]` or something like that, accompanied by an `abort(port)` which allows removing the handler (or something similar).

drewhk: btw, please install a default handler to each stage, because currently the handler Arrays are null. Simply filling them with a constant InvalidInHandler and InvalidOutHandler that just throw on every event "no handler is installed" could be helpful.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Websocket.scala
drewhk: can imagine something similar specialized for bytes, too. (i.e. expresses counts of bytes instead of counts of ByteString instances)

ktoso: make the `andThen: () => Unit`? We're sticking to being explicit in most APIs in streams

rkuhn: Yes, I came to the same conclusion for other reasons as well.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Websocket.scala
rkuhn: I added a check during materialization that leads to a nice error message, is that enough?

drewhk: that can work, yes

rkuhn:
@sirthias @jrudolph please take a look at the changes in `Websocket.scala`: this is the current draft of the replacement API for FlexiMerge/Route and I’ll continue applying and evolving it tomorrow across all of akka-http-core.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameOutHandler.scala
```diff
@@ -44,6 +43,9 @@ private[http] class FrameOutHandler(serverSide: Boolean, _closeTimeout: FiniteDu
       case UserHandlerCompleted ⇒
         become(new WaitingForPeerCloseFrame())
         ctx.push(FrameEvent.closeFrame(Protocol.CloseCodes.Regular))
+      case UserHandlerErredOut(ex) ⇒
+        become(new WaitingForPeerCloseFrame())
+        ctx.push(FrameEvent.closeFrame(Protocol.CloseCodes.UnexpectedCondition, "internal error"))
```
rkuhn: @sirthias @jrudolph please take a look, I think this was missing but I’m not sure about what exactly should be done in this case

sirthias: Ok, Johannes will take a look.... (he's currently sitting next to me.... :)

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -127,7 +176,44 @@ abstract class GraphStageLogic {
   /**
    * INTERNAL API
    */
-  private[stream] var interpreter: GraphInterpreter = _
+  private[this] var _interpreter: GraphInterpreter = _
+
+  /**
+   * INTENRAL API
+   */
+  private[stream] def interpreter_=(gi: GraphInterpreter) = _interpreter = gi
```
drewhk: yeah, good move.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -127,7 +176,44 @@ abstract class GraphStageLogic {
...
+
+  /**
+   * INTERNAL API
+   */
+  private[stream] def interpreter: GraphInterpreter =
+    if (_interpreter == null)
+      throw new IllegalStateException("not yet initialized: only setHandler is allowed in GraphStageLogic constructor")
```
drewhk: "setHandler is only allowed"

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -127,7 +176,44 @@ abstract class GraphStageLogic {
...
+
+  /**
+   * INTENRAL API
+   */
+  private[stream] def interpreter_=(gi: GraphInterpreter) = _interpreter = gi
+
+  /**
+   * INTERNAL API
+   */
+  private[stream] def interpreter: GraphInterpreter =
```
drewhk: are there hot paths that might benefit from accessing _interpreter idirectly nstead?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -263,6 +364,232 @@ abstract class GraphStageLogic {
   final protected def isClosed[T](out: Outlet[T]): Boolean = interpreter.outStates(conn(out)) eq Closed
 
   /**
+   * Read a number of elements from the given inlet and continue with the given function,
```
drewhk: can these features go to a subclass?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -263,6 +364,232 @@ abstract class GraphStageLogic {
...
+      }
+    }
+  }
+
+  /**
+   * Install a handler on the given inlet that emits received elements on the
+   * given outlet before pulling for more data. `doTerminate` controls whether
+   * completion or failure of the given inlet shall lead to stage termination or not.
+   */
+  final protected def passAlong[Out, In <: Out](from: Inlet[In], to: Outlet[Out], doTerminate: Boolean): Unit =
```
drewhk: maybe a version that has an `until: () => Boolean`?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -263,6 +364,232 @@ abstract class GraphStageLogic {
...
+    setHandler(from, new InHandler {
+      override def onPush(): Unit = {
+        val elem = grab(from)
+        emit(to, elem, () ⇒ pull(from))
+      }
+      override def onUpstreamFinish(): Unit = if (doTerminate) super.onUpstreamFinish()
```
drewhk: should on !doTerminate deinstall this handler?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -127,7 +176,44 @@ abstract class GraphStageLogic {
...
+  private[stream] def interpreter: GraphInterpreter =
+    if (_interpreter == null)
+      throw new IllegalStateException("not yet initialized: only setHandler is allowed in GraphStageLogic constructor")
+    else _interpreter
+
+  /**
+   * Input handler that terminates the stage upon receiving completion.
+   * The stage fails upon receiving a failure.
+   */
+  final protected def eagerTerminateInput: InHandler = new EagerTerminateInput
```
drewhk: these below probably need some bikeshedding :)

drewhk:
Don't really know the Http side of it, but otherwise looks like a good direction.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: nah, I love the “totally ignorant” part best :-)

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: to be replaced with what?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: my guess is that more dynamic usage will be better off just implementing the whole handler directly

rkuhn: in other words: let’s flesh this out on an “as needed internally” basis

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: sure, any particular reason?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: the idea is to make the error reporting reliable; Hotspot will aggressively inline the accessor on any hot path

akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameOutHandler.scala
rkuhn: What was the verdict? (I’ll start running with this approach now, so if you want to alter my course now is exactly the right time)

rkuhn: sorry: context was “please comment on the overall approach”, not this line

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: but then there is no protection to the call from pull(from), so if this continues while upstream finished then it will fail right?

drewhk: I see the doTerminate flag to be dangerous in this form.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: to not grow the class to large. Things that only use the bare minimum features might not need all that baggage (this might include timers, too.)

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: fine, let's not rush ahead.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: ah, that is correct, but replacing this handler will not fix things: the problematic code is enqueued on the output port already; I’ll change it to use tryPull (unless you have another brilliant idea)

drewhk: that will silently stop processing. I think a callback is useful in that case. Or maybe this sugar method just does not pull its weight?

rkuhn: It happens to be exactly what is needed for a merge that does not until all (or a special) ports have finished, hence I added it.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameOutHandler.scala
jrudolph: I cannot say for sure right now. I will be able to look into it next week. AFAICS either this case didn't happen (in which case I should have documented it, or otherwise, it is indeed missing (and missing a test as well). So, could you add a ticket to figure this out? I will then have a look next week. Thanks!

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Websocket.scala
```diff
@@ -212,13 +131,85 @@ private[http] object Websocket {
...
+    val shape = new FanOutShape2(in, bypass, user)
+
+    def createLogic = new GraphStageLogic(shape) {
+
+      setHandler(in, new InHandler {
+        override def onPush(): Unit = {
+          grab(in) match {
+            case b: BypassEvent with MessagePart ⇒ emit(bypass, b, () ⇒ emit(user, b, () ⇒ pull(in)))
+            case b: BypassEvent                  ⇒ emit(bypass, b, () ⇒ pull(in))
+            case m: MessagePart                  ⇒ emit(user, m, () ⇒ pull(in))
```
viktorklang: Doesn't it generate significant garbage to allocate functions for every emit?

rkuhn: Obviously it generates garbage, whether significant remains to be measured. But the real question is: can it be avoided? Because equally obviously the continuation may need to be suspended, meaning that the creation of an Object is unavoidable (short of building in a crystal ball).

viktorklang: `() => pull(in)` looks extremely hoistable.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -300,6 +410,231 @@ abstract class GraphStageLogic private[stream] (inCount: Int, outCount: Int) {
...
 
   /**
+   * Read a number of elements from the given inlet and continue with the given function,
+   * suspending execution if necessary. This action replaces the [[InHandler]]
+   * for the given inlet if suspension is needed and reinstalls the current
+   * handler upon receiving the last `onPush()` signal (before invoking the `andThen` function).
+   */
+  final protected def readN[T](in: Inlet[T], n: Int)(andThen: Array[T] ⇒ Unit): Unit =
+    if (n < 0) throw new IllegalArgumentException("cannot read negative number of elements")
+    else if (n == 0) andThen(new Array[Any](0).asInstanceOf[Array[T]])
```
viktorklang: Why allocate new empty arrays?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -88,12 +130,12 @@ private object TimerMessages {
 abstract class GraphStageLogic private[stream] (inCount: Int, outCount: Int) {
   import GraphInterpreter._
   import TimerMessages._
+  import GraphStageLogic._
 
   def this(shape: Shape) = this(shape.inlets.size, shape.outlets.size)
 
   private val keyToTimers = mutable.Map[Any, Timer]()
   private val timerIdGen = Iterator from 1
```
viktorklang: Is this safe to be an Int and why start at 1?

rkuhn: Yes, it is safe, and it can start at any number (it would break if a timer is cancelled and rescheduled four billion times while the message is in flight from the scheduler—which is more than just hypothetical).

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: In fact: yes, this would be necessary if actually going with Arrays because otherwise there could be CCEs (which is why there is currently a ClassTag missing as well). In order to sidestep these issues I changed it to an ArrayBuffer.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Websocket.scala
rkuhn: yes, it does (thanks)

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: I also changed this from an Iterator to a var in my other branch to reduce memory overhead.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameOutHandler.scala
rkuhn: created #18685

akka-stream/src/main/boilerplate/akka/stream/FanOutShape.scala.template
```diff
@@ -20,12 +21,12 @@ object FanOutShape {
   }
 }
 
-abstract class FanOutShape[I] private (_in: Inlet[I], _registered: Iterator[Outlet[_]], _name: String) extends Shape {
+abstract class FanOutShape[-I] private (_in: Inlet[I @uncheckedVariance], _registered: Iterator[Outlet[_]], _name: String) extends Shape {
```
rkuhn: All this `@uncheckedVariance` dance is needed because an Inlet/Outlet really wants use-site variance (because it does not have any methods on it) but expressing that involves additional type parameters or existential wildcards that are just not bearable (and sometimes even not expressible).

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -398,12 +761,6 @@ abstract class GraphStageLogic private[stream] (inCount: Int, outCount: Int) {
 
   // Internal hooks to avoid reliance on user calling super in preStart
   protected[stream] def beforePreStart(): Unit = {
```
rkuhn: these hooks might want to go completely—there is currently no use for them

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
```diff
@@ -79,11 +79,11 @@ private object PoolConductor {
       }
 
       retryMerge.out ~> slotSelector.in0
-      slotSelector.out ~> doubler ~> route.in
-      retrySplit.out0 ~> flatten ~> slotSelector.in1
-      retrySplit.out1 ~> retryMerge.in1
+      slotSelector.out ~> route.in
+      retrySplit.out(0).filter(!_.isInstanceOf[SlotEvent.RetryRequest]) ~> flatten ~> slotSelector.in1
```
ktoso: Feels clunky, let's eventually add `filterNot`. (`filter(!_.bla)` being a code smell on collections, I react the same when I see it on here)

rkuhn: good point (#18687)

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -171,78 +173,75 @@ private[http] object HttpServerBluePrint {
...
-          case other ⇒ throw new IllegalStateException(s"Unexpected MergeLogicContext [${other.getClass.getName}]")
-        }
-        finish(ctx)
+        emit(out, ResponseRenderingContext(HttpResponse(status, entity = msg), closeRequested = true), () ⇒ completeStage())
       }
 
-      def finish(ctx: MergeLogicContextBase): State[Any] = {
-        ctx.finish() // shouldn't this return a `State` rather than `Unit`?
-        SameState // it seems weird to stay in the same state after completion
+      setHandler(oneHundredContinue, ignoreTerminateInput) // RK: not sure if this is always correct
```
ktoso: uh, `100 continue` is a tricky one... have you seen this @sirthias @jrudolph ?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -329,79 +328,90 @@ private[http] object HttpServerBluePrint {
...
-          ctx.emit(shape.out1)(element.asInstanceOf[ByteString])
-          SameState
-        }
-      }
+      })
+
+      setHandler(httpOut, conditionalTerminateOutput(() ⇒ target == httpOut))
+      setHandler(wsOut, conditionalTerminateOutput(() ⇒ target == wsOut))
+
+      val pullIn = () ⇒ pull(in)
```
ktoso: nice trick

ktoso:
AFAICS looks good.

akka-stream/src/main/scala/akka/stream/javadsl/Graph.scala
```diff
@@ -277,49 +304,41 @@ object FlowGraph {
...
-      def via[U](f: FlowShape[U, T]): ReverseOps[U] = to((out <~ f).inlet)
-      def via[U](j: UniformFanInShape[U, T]): ReverseOps[U] = to((out <~ j).inlet)
-      def via[U](j: UniformFanOutShape[U, T]): ReverseOps[U] = to((out <~ j).inlet)
+      def fromOutlet(dst: Outlet[_ <: T]): Builder[Mat] = { out <~ dst; self }
+      def from(dst: SourceShape[_ <: T]): Builder[Mat] = { out <~ dst; self }
+      def fromFanIn[U](j: UniformFanInShape[U, _ <: T]): Builder[Mat] = { out <~ j; self }
+      def fromFanOut[U](j: UniformFanOutShape[U, _ <: T]): Builder[Mat] = { out <~ j; self }
+      def via[U](f: FlowShape[U, _ <: T]): ReverseOps[U] = to((out <~ f).inlet)
+      def viaFanIn[U](j: UniformFanInShape[U, _ <: T]): ReverseOps[U] = to((out <~ j).inlet)
+      def viaFanOut[U](j: UniformFanOutShape[U, _ <: T]): ReverseOps[U] = to((out <~ j).inlet)
```
patriknw: +1

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -299,6 +437,239 @@ abstract class GraphStageLogic private[stream] (inCount: Int, outCount: Int) {
    */
   final protected def isClosed[T](out: Outlet[T]): Boolean = (interpreter.portStates(conn(out)) & OutClosed) != 0
 
+  private val emptyArray = new ArrayBuffer[Any](0)
```
patriknw: does it have to be an `ArrayBuffer` or would `Nil` work?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -299,6 +437,239 @@ abstract class GraphStageLogic private[stream] (inCount: Int, outCount: Int) {
...
+   */
+  final protected def emitMultiple[T](out: Outlet[T], elems: Iterator[T], andThen: () ⇒ Unit): Unit =
+    if (elems.hasNext) {
+      if (isAvailable(out)) {
+        push(out, elems.next())
+        if (elems.hasNext) {
+          setOrAddEmitting(out, new EmittingIterator(out, elems, getNonEmittingHandler(out), andThen))
+        } else {
+          andThen()
+        }
```
patriknw: possible curly brace reduction

patriknw:
looking good

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: `Nil` should work and it is better, thanks!

akka-docs-dev/rst/scala/code/docs/stream/BidiFlowDocSpec.scala
```diff
@@ -50,7 +50,7 @@ object BidiFlowDocSpec {
     // construct and add the bottom flow, going inbound
     val inbound = b.add(Flow[ByteString].map(fromBytes))
     // fuse them together into a BidiShape
-    BidiShape(outbound, inbound)
+    BidiShape.fromFlows(outbound, inbound)
```
drewhk: I like this one better than the `fromGraphs`. @viktorklang ?

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -106,126 +108,109 @@ private[http] object OutgoingConnectionBlueprint {
...
+    private val out = Outlet[HttpRequest]("out")
 
-    def createMergeLogic(p: PortT) = new MergeLogic[HttpRequest] {
+    val shape = new FanInShape2(requests, responses, out)
 
-      val requestInput = p.in0
-      val terminationBackchannelInput = p.in1
+    override def createLogic = new GraphStageLogic(shape) {
+      passAlong(requests, out, doFinish = false, doFail = true)
+      setHandler(out, eagerTerminateOutput)
```
drewhk: doesn't out needs to be pulled continuously and the values discarded? or this channel only carries completion?

akka-docs-dev/rst/scala/code/docs/stream/BidiFlowDocSpec.scala
viktorklang: Sounds good. As long as it is easy to get dot-completion/intellisense. I'll curate a proposed update on my PR:

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
drewhk: interesting, I am one of those weird ones that like explicit negation in the predicate instead of in the method name ;)

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Websocket.scala
```diff
@@ -212,13 +129,86 @@ private[http] object Websocket {
...
       BidiShape(
-        routePreparation.inlet,
+        split.in,
         messagePreparation.outlet,
         messageRendering.inlet,
         merge.out)
     }.named("ws-message-api")
   }
 
+  private object bypassRouter extends GraphStage[FanOutShape2[Output, BypassEvent, MessagePart]] {
```
drewhk: why the small letter?

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/FlowGraphDocTest.java
```diff
@@ -131,7 +134,7 @@ public void demonstrateMatValue() throws Exception {
       (b, fold) -> {
         return new Pair<>(
           fold.inlet(),
-          b.from(b.materializedValue()).via(flatten).out());
+          b.from(b.materializedValue()).via(b.graph(flatten)).out());
```
drewhk: I think I like `b.add()` better than `b.graph()`

viktorklang: Yeah, what does "graph" do?

akka-stream/src/main/boilerplate/akka/stream/FanOutShape.scala.template
drewhk: Does this affect people when they need to create their own shapes?

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
```diff
@@ -151,48 +147,6 @@ private[akka] case class ActorMaterializerImpl(val system: ActorSystem,
           ActorProcessorFactory[Any, Any](
             actorOf(opprops, stageName(effectiveAttributes), effectiveSettings.dispatcher)) -> mat
       }
-
```
drewhk: muhahahahaa!!!

akka-stream/src/main/scala/akka/stream/impl/FanIn.scala
```diff
@@ -292,10 +291,3 @@ private[akka] abstract class FanIn(val settings: ActorMaterializerSettings, val
 
 }
```
drewhk: cannot this whole file go? Is there something reused from here by stream-of-stream ops?

akka-stream/src/main/scala/akka/stream/impl/FanOut.scala
```diff
@@ -316,11 +315,3 @@ private[akka] class Unzip(_settings: ActorMaterializerSettings) extends FanOut(_
     }
   })
 }
```
drewhk: same here, how come this file cannot be removed?

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -302,6 +310,22 @@ private[stream] final class GraphInterpreter(
   }
 
   /**
+   * Dynamic handler changes are communicated from a GraphStageLogic by this method.
+   */
+  def setHandler(connection: Int, handler: InHandler): Unit = {
```
drewhk: oh, I didn't expect these. But of course you are right, the interpreter has everything flattened into a connection id indexed array.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -18,8 +18,9 @@ object Merge {
    * Create a new `Merge` with the specified number of input ports.
    *
    * @param inputPorts number of input ports
+   * @param eagerClose if true, the merge will complete as soon as one of its inputs completes.
    */
-  def apply[T](inputPorts: Int): Merge[T] = new Merge(inputPorts)
+  def apply[T](inputPorts: Int, eagerClose: Boolean = false): Merge[T] = new Merge(inputPorts, eagerClose)
```
drewhk: there is a table of stages and their close semantics, you will need to update that one too.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -129,7 +189,54 @@ abstract class GraphStageLogic private[stream] (inCount: Int, outCount: Int) {
...
+  /**
+   * Input handler that terminates the stage upon receiving completion.
+   * The stage fails upon receiving a failure.
+   */
+  final protected def eagerTerminateInput: InHandler = new EagerTerminateInput
+  /**
+   * Input handler that does not terminate the stage upon receiving completion.
+   * The stage fails upon receiving a failure.
+   */
+  final protected def ignoreTerminateInput: InHandler = new IgnoreTerminateInput
```
drewhk: Good, no extra fields added.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -165,6 +295,14 @@ abstract class GraphStageLogic private[stream] (inCount: Int, outCount: Int) {
   }
 
   /**
+   * Requests an element on the given port unless the port is already closed.
+   * Calling this method twice before an element arrived will fail.
+   * There can only be one outstanding request at any given time. The method [[hasBeenPulled()]] can be used
+   * query whether pull is allowed to be called or not.
+   */
+  final protected def tryPull[T](in: Inlet[T]): Unit = if (!isClosed(in)) pull(in)
```
drewhk: does this need a dedicated method?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: I dislike the extra field. I removed quite a bit of memory overhead in the other PR, so I would not like to see it again :)

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: ah, since you are only allowing timers from preStart(). This solves one of my issues, as this allows the timer related things to be pulled out to a separate class that is not a subclass.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -299,6 +437,239 @@ abstract class GraphStageLogic private[stream] (inCount: Int, outCount: Int) {
...
 
+  private val emptyArray = new ArrayBuffer[Any](0)
+
+  /**
+   * Read a number of elements from the given inlet and continue with the given function,
+   * suspending execution if necessary. This action replaces the [[InHandler]]
+   * for the given inlet if suspension is needed and reinstalls the current
+   * handler upon receiving the last `onPush()` signal (before invoking the `andThen` function).
+   */
+  final protected def readN[T](in: Inlet[T], n: Int)(andThen: Seq[T] ⇒ Unit): Unit =
```
drewhk: do you think these methods (this and below) should be on the default GraphStageLogic? We can push them to a subclass, or a separate non-subclass that can be composed separately.

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
rkuhn: `out` is the output, which just gets the elements from the `requests` input; the `responses` input is indeed continuously pulled a few lines below this one

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Websocket.scala
rkuhn: ah, this was a `val` initially; will fix

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/FlowGraphDocTest.java
rkuhn: I didn’t want to do this particular rename within this PR because it probably conflicts with a lot of other things—it is on my list to do next, but I’ll wait until our big refactorings are merged.

akka-stream/src/main/boilerplate/akka/stream/FanOutShape.scala.template
rkuhn: Yes, if they want to define custom shapes from Scala with variance; this should be rarely necessary in general (i.e. with generics), though.

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
rkuhn: :-D

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: It is gone already, replaced by `Nil`.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: What would be the gain? There is no runtime overhead in having them ready everywhere and having just one class is the most discoverable strategy for users.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: I lifted it from multiple improvised use sites, which tells me that there is utility in having it handy.

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/FlowGraphDocTest.java
drewhk: ok, put a fixme or ticket somewhere

rkuhn: #18692

akka-http-core/src/main/scala/akka/http/impl/engine/ws/FrameHandler.scala
```diff
@@ -176,7 +177,9 @@ private[http] object FrameHandler {
     }
   }
 
-  sealed trait MessagePart {
+  sealed trait Output
```
jrudolph: Good change.

jrudolph:
Looks good as far as I can judge this with my limited knowledge about how the new API works. I ran the server-side autobahn test suite to check if this still works (it exercises closing logic quite a bit and I guess that could have easily changed in the transition). It found no failures. :+1: from me.

rkuhn:
Thanks, @jrudolph !

rkuhn:
@drewhk This means that you’ll have some logic conflict when rebasing your OneBoundedInterpreter removal.

Issue #18447, opened 2015-09-10
Remember to adjust persistence mima filters for 2.4.0


patriknw:
We must start checking binary compatibility with 2.4.0, especially for persistence since we have excluded that package in 2.3.x compatibility check.

PR #18606, opened 2015-09-30
=pro #18447 Check binary compatibility with 2.4.0

patriknw:
This changes the binary compatibility baseline to 2.4.0. We intend to check against several versions, which is in progress in https://github.com/akka/akka/pull/17761.


patriknw:
Refs #18447

patriknw:
PLS BUILD

patriknw:
Let's merge this, so that new pull requests are validated with it. Further improvements can be done in #17761.

PR #15596, opened 2014-07-28
Update AbstractNodeQueue.java to avoid redundant reads of head in peekNode

nitsanw:
Avoid spinning on head when a peek/poll observes a null next but the queue is not empty.

PR #18621, opened 2015-10-01
=act correct two comments in AbstractNodeQueue.java

rkuhn:
cc @nomagic @nitsanw

### DIFF
--- a/akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
+++ b/akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
@@ -14,7 +14,8 @@ import java.util.concurrent.atomic.AtomicReference;
  */
 @SuppressWarnings("serial")
 public abstract class AbstractNodeQueue<T> extends AtomicReference<AbstractNodeQueue.Node<T>> {
-    // Extends AtomicReference for the "head" slot (which is the one that is appended to) since Unsafe does not expose XCHG operation intrinsically
+    // Extends AtomicReference for the "head" slot (which is the one that is appended to) since
+    // Unsafe does not expose XCHG operation intrinsically before JDK 8
     @SuppressWarnings("unused")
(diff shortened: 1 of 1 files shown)


rkuhn:
(see also https://github.com/akka/akka/pull/15596#issuecomment-50446085)

akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
```diff
@@ -14,7 +14,8 @@
  */
 @SuppressWarnings("serial")
 public abstract class AbstractNodeQueue<T> extends AtomicReference<AbstractNodeQueue.Node<T>> {
-    // Extends AtomicReference for the "head" slot (which is the one that is appended to) since Unsafe does not expose XCHG operation intrinsically
+    // Extends AtomicReference for the "head" slot (which is the one that is appended to) since
+    // Unsafe does not expose XCHG operation intrinsically before JDK 8
```
viktorklang: Since 2.4 is JDK8+ this would no longer be an issue, tho?

rkuhn: Well, the comment is still correct :-)

But what would we gain by changing?

viktorklang: I guess what we gain is that extenders no longer can mess around with the full AR interface :)

rkuhn: That is true. But I’m not yet convinced that this is a strong enough reason to change.

viktorklang: That makes two of us :)

rkuhn:
This can be merged by whoever sees the green button (as a token of approval).

PR #18631, opened 2015-10-02
=htc #18397 fix ArrayIndexOutOfBoundsException in HttpHeaderParser

### LINKED ISSUE #18397: Do we really need header caching?
We've been trying to track down a very elusive bug in one of our systems, where we get a `400 Bad Request` error after a number of requests to the server. We finally tracked it down as far as the HTTP Header caching in Spray, for some reason the `nodeCount` variable in HttpHeaderParser is being set to a value > 32k, which, since it later is converted to a Short, results in overflow.

We have fixed the problem by disabling HTTP header caching (`spray.can.server.parsing.header-cache.default = 0`). We have tried to make a reproducible minimal project but have been unable to.

My question is, what effects can we expect from disabling header caching entirely? I don't really understand why it is even necessary, since the headers have to be parsed anyway in every request (since the values might have changed). Is this to save memory by reusing HttpHeader objects?

sirthias:
Closes #18397 (and fixes an unrelated off-by-one error as the cherry on the cake :-)

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpHeaderParser.scala
@@ -138,7 +138,7 @@ private[engine] final class HttpHeaderParser private (
   }
 
   private def parseRawHeader(input: ByteString, lineStart: Int, cursor: Int, nodeIx: Int): Int = {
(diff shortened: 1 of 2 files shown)


sirthias:
@ktoso This one should also get merged into 1.1.

ktoso:
Thanks, LGTM

PR #18640, opened 2015-10-05
cleanup some test configuration

patriknw:


akka-actor/src/main/resources/reference.conf
```diff
@@ -82,10 +82,6 @@ akka {
     # Timeout for ActorSystem.actorOf
     creation-timeout = 20s
 
-    # Frequency with which stopping actors are prodded in case they had to be
-    # removed from their parents
-    reaper-interval = 5s
```
patriknw: this is not used anywhere

akka-remote/src/test/scala/akka/remote/RemotingSpec.scala
```diff
@@ -637,7 +635,7 @@ class RemotingSpec extends AkkaSpec(RemotingSpec.cfg) with ImplicitSender with D
       val config = ConfigFactory.parseString(s"""
         akka.remote.enabled-transports = ["akka.remote.test"]
         akka.remote.retry-gate-closed-for = 5s
-        akka.remote.log-lifecylce-events = on
+        akka.remote.log-remote-lifecycle-events = on
```
drewhk: I am not sure if this is actually needed. Maybe comment it out, so it can be easily enabled?

Good catch btw.

akka-actor/src/main/resources/reference.conf
drewhk: I always wondered about this setting.

drewhk:
LGTM

ktoso:
LGTM

rkuhn:
LGTM

Issue #18641, opened 2015-10-05
FAILED: TlsEndpointVerificationSpec


ktoso:
Seems to have been caused by strictly checking equality of ports now:

```
[WARN] [10/02/2015 15:36:02.054] [TlsEndpointVerificationSpec-akka.test.stream-dispatcher-10] [ActorSystem(TlsEndpointVerificationSpec)] Illegal request, responding with status '400 Bad Request': Request is missing required `Host` header: 'Host' header value of request to `https://akka.example.org/` doesn't match request target authority: Host header: Some(Host: akka.example.org:8080)
request target authority: //akka.example.org
[info] - should accept certificates signed by known CA *** FAILED *** (117 milliseconds)
[info]   400 Bad Request did not equal 200 OK (TlsEndpointVerificationSpec.scala:44)
[info]   org.scalatest.exceptions.TestFailedException:
```

Introduced in https://github.com/akka/akka/pull/18633

PR #18633, opened 2015-10-02
=htp #18632 check ports of URI and Host header in effectiveUri

aruediger:

PR #18642, opened 2015-10-05
=htc #18641 request must be made to the right port

ktoso:
Failure caused by more strictness introduced in: https://github.com/akka/akka/pull/18633 (ports must match)
Resolves https://github.com/akka/akka/issues/18641 

Was: 

```
[WARN] [10/02/2015 15:36:02.054] [TlsEndpointVerificationSpec-akka.test.stream-dispatcher-10] [ActorSystem(TlsEndpointVerificationSpec)] 
Illegal request, responding with status '400 Bad Request': 
Request is missing required `Host` header: 'Host' header value of 
request to `https://akka.example.org/` doesn't match request target authority: 
Host header: Some(Host: akka.example.org:8080)
request target authority: //akka.example.org
```

Please review @rkuhn @drewhk

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/client/TlsEndpointVerificationSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/client/TlsEndpointVerificationSpec.scala
@@ -40,7 +40,7 @@ class TlsEndpointVerificationSpec extends AkkaSpec("""
     "accept certificates signed by known CA" in {
       val pipe = pipeline(ExampleHttpContexts.exampleClientContext, hostname = "akka.example.org") // example context does include custom CA
 
-      whenReady(pipe(HttpRequest(uri = "https://akka.example.org/")), timeout) { response ⇒
+      whenReady(pipe(HttpRequest(uri = "https://akka.example.org:8080/")), timeout) { response ⇒
         response.status shouldEqual StatusCodes.OK
       }
     }


drewhk:
Can you elaborate on this a little bit more?

ktoso:
I'm re-reading the spec now, and the part we're talking about here is: http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23 

So technically the spec does not require rejecting such requests (i.e. `URL "bla:8080" + Host: bla` which is illegal, however the spec does not explicitly say that one must reject it - it only does so for missing Host headers).

The related akka-user discussion is: https://groups.google.com/forum/#!searchin/akka-user/port$20header/akka-user/eAgsGyUzMu4/BkeGTZCwBgAJ

So technically requests with not matching poty in `Host` header are illegal, yet the spec does not require rejecting them... Currently (since that commit) we reject them, which caused this test to fail.

Opinions on this @sirthias @jrudolph @rkuhn?

drewhk:
I though it is not as simple :)

ktoso:
/me leaning towards making things as strict as possible until someone complains. As a general principle that's how Spray / HTTP have been built. Thanks @drewhk for reminding me that.

2beaucoup:
Sorry for breaking that test. Should have tested more thoroughly here. :(

From http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23:

>  The "Host" header field in a request <b>provides the host and port
>  information from the target URI</b>, enabling the origin server to
>  distinguish among resources while servicing requests for multiple
>  host names on a single IP address.
>  
>  ... 
>  
>  A server MUST respond with  a 400 (Bad Request) status code to any
>  HTTP/1.1 request message that lacks a Host header field and to any
>  request message that contains more than one Host header field **or a
>  Host header field with an invalid field-value**.

So the previous behaviour of rejecting non matching hosts looks right to me but should check the port as well.

Anyway: Thanks for cleaning up after me!

ktoso:
No problem, thanks for chiming in @2beaucoup! :)

I can't seem to find this section:

```
A server MUST respond with a 400 (Bad Request) status code to any
HTTP/1.1 request message that lacks a Host header field and to any
request message that contains more than one Host header field or a
Host header field with an invalid field-value.
```

in the linked RFC, where's that paragraph from?

jrudolph:
LGTM. 

The section looks the same in the current spec: http://tools.ietf.org/html/rfc7230#section-5.4

One thing to note here is that this whole scenario seems to be one where the client uses [an absolute target URI in the request line](http://tools.ietf.org/html/rfc7230#section-5.3.2) (which is uncommon), so the check will not be used in most cases.

An absolute target URI, however, is required if the request goes to a proxy and for the proxy itself the now implemented rule is not valid and [RFC 7230 Section 5.4 instead mandates](http://tools.ietf.org/html/rfc7230#section-5.4):

> When a proxy receives a request with an absolute-form of
   request-target, the proxy MUST ignore the received Host header field
   (if any) and instead replace it with the host information of the
   request-target.  A proxy that forwards such a request MUST generate a
   new Host field-value based on the received request-target rather than
   forward the received Host field-value.

It seems that the user in the [linked akka-user post](https://groups.google.com/forum/#!searchin/akka-user/port%2420header/akka-user/eAgsGyUzMu4/BkeGTZCwBgAJ) tried to implement a proxy server with akka-http for which the rules may be too strict. I guess, for now implementing an HTTP proxy with akka-http may simply not be supported.

On the other hand, something seems fishy in the logs from the post which contain a line like this:

```
[ERROR] [09/28/2015 20:46:21.163] [Proxy-akka.actor.default-dispatcher-386] [ActorSystem(Proxy)] Internal server error, sending 500 response
akka.http.scaladsl.model.IllegalUriException: 'Host' header value of request to `openoox.com:8080` doesn't match request target authority: Host header: Some(Host: openoox.com:8080) request target authority:
```

Do we understand how this happens or this there another bug lurking inside?

2beaucoup:
Sorry (again), wrong RFC. It's this one: https://tools.ietf.org/html/rfc7230#section-5.4.

drewhk:
> Do we understand how this happens or this there another bug lurking inside?

+1, first I thought that a simple equality testing in test to a string failed because String::toString vs. Option::toString, and it still looks like of something similar happening.

2beaucoup:
I agree with @jrudolph that the strict approach in akka-http isn't suited for a proxy impl.

Maybe we should think about a "relaxed mode" that only checks for absolute minimum correctness that proxies need.

jrudolph:
I created #18643 to clarify the situation for the other issue.

ktoso:
I'll merge this one and we'll continue work on this in the ticket @jrudolph created, in order to make PR validation useful again (all validations fail on this test currently).

PR #18655, opened 2015-10-06
#18058 Bubble up tls errors

### LINKED ISSUE #18058: SslTls gulps any exceptions and just completes the stream

AnIrishDuck:
Fixes #18058. I'm not sure where exactly to merge this, please advise.


ktoso:
OK TO TEST

ktoso:
Nice! Will review soon! :+1:

akka-stream/src/main/scala/akka/stream/impl/io/SslTlsCipherActor.scala
```diff
@@ -299,6 +299,8 @@ private[akka] class SslTlsCipherActor(settings: ActorMaterializerSettings, sslCo
       } catch {
         case ex: SSLException ⇒
           if (tracing) log.debug(s"SSLException during doUnwrap: $ex")
+          fail(ex, closeTransport = false)
+          engine.closeInbound()
```
rkuhn: Is this not redundant after having seen an exception from the engine?

rkuhn:
LGTM

akka-stream/src/main/scala/akka/stream/impl/io/SslTlsCipherActor.scala
```diff
@@ -422,10 +425,13 @@ private[akka] class SslTlsCipherActor(settings: ActorMaterializerSettings, sslCo
 
   initialPhase(2, bidirectional)
 
-  protected def fail(e: Throwable): Unit = {
+  protected def fail(e: Throwable, closeTransport: Boolean=true): Unit = {
     if (tracing) log.debug("fail {} due to: {}", self, e.getMessage)
     inputBunch.cancel()
-    outputBunch.error(TransportOut, e)
+    if(closeTransport) {
```
drewhk: why is this needed?

AnIrishDuck: Closing the transport here prevents the SSLEngine from entering the `flushingOutbound` state when `completeOrFlush` is called after this function. This prevents the necessary TLS error messages from getting sent.

akka-stream/src/main/scala/akka/stream/impl/io/SslTlsCipherActor.scala
AnIrishDuck: It's not redundant, in that the provided test fails without the `engine.closeInbound()` call. The connection is never closed, and the reaper detects the dangling actors and fails the test.

*Why* this is true is not clear to me. Reading the [documentation](https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/SSLEngine.html) for `SSLEngine`:

> A close can be initiated by one of: an SSLException, an inbound closure handshake message, or one of the close methods. In all cases, closure handshake messages are generated by the engine, and wrap() should be repeatedly called until the resulting SSLEngineResult's status returns "CLOSED", or isOutboundDone() returns true. All data obtained from the wrap() method should be sent to the peer.

So it sounds like if `doUnwrap()` raises `SSLException`, it should have already called `closeInbound()`. But that's clearly *not* the case, because the `closeInbound()` call is not a no-op. I'm not sure if this is an `SSLEngine` bug, bad / outdated `SSLEngine` documentation, or (most likely) me not fully understanding `SSLEngine`.

It's possible that some `SSLException`s are not considered "fatal", and thus will not close the inbound stream. I can't find any reference to this possibility in `SSLEngine` documentation.

I am frankly not convinced that my own closure logic is completely correct. `SSLEngine` is a complicated beast. But it does pass the test now, which means it is at least correct in cases where it was not before.

rkuhn: Thanks for the investigation and the explanation, as perverse as it sounds it makes perfect sense to me—meaning it matches my previous experience with SSLEngine.

ktoso:
LGTM

ktoso:
OK TO TEST

ktoso:
Letting validation do it's job before merge.

drewhk:
this needs squashing to a single commit first. Otherwise excellent job!

AnIrishDuck:
OK ... while I'm rebasing, what branch should I apply this to? release-2.3-dev doesn't seem right, but I'm not familiar with your branching / release practices...

rkuhn:
Yes, Streams & HTTP is still developed on branch release-2.3-dev—that will change in a few weeks when we move these projects to master.

AnIrishDuck:
I squashed all the commits together and rebased to the current release-2.3-dev, should be good to go. Let me know if I need to do anything else.

ktoso:
Looks good from our side, thanks :)
Would be nice if the commit could be named `=str #18058 Emit errors from TLS handshake` as outlined in https://github.com/akka/akka/blob/master/CONTRIBUTING.md#creating-commits-and-writing-commit-messages 

Otherwise this is ready to be merged in a short bit :)

ktoso:
I'll go ahead with merging this, small inconsistency in commit message style doesn't hurt that much ;)

ktoso:
Thanks a lot for the contrib @AnIrishDuck !

Issue #17214, opened 2015-04-16
optionally allow a payload with GET, HEAD, DELETE, CONNECT request methods


rkuhn:
This may be configurable—the specification says that such payload has no predefined semantics, but it is also not forbidden.

Careful: for TRACE it actually is forbidden, so just ignoring isEntityAccepted based on some configuration is not the right thing to do.

rkuhn:
Thinking a bit more: the server should probably always accept it while on the client it needs to be enabled with a configuration setting (cf. Postel’s law).

sirthias:
Things to remind user of who are trying to rely on GET request entities which are "undefined" by the spec:

1. Existing proxies implementation might refuse to proxy the request payload.

2. By default the responses to GET requests are cachable. (["unless otherwise indicated by the Cache-Control header field"](http://tools.ietf.org/html/rfc7231#section-4.3.1))
It is unlikely that existing cache implementations include the request entity into the cache key, which means that a server relying on GET request entities must make sure to always set cache control headers properly itself.

In general we should discourage users from creating APIs which rely on such "undefined" features. If they choose to ignore this advice they will be on their own and see all kinds of strange issues in the wild, that we will be probably be unable to provide help for.

sirthias:
Also interesting is [section 4.3 of RFC 2616](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.3), which was the state-of-the-art spec for almost all HTTP related products that are currently in the field. It says:

> A server SHOULD read and forward a message-body on any request; **if the request method does not include defined semantics for an entity-body, then the message-body SHOULD be ignored when handling the request.**

This [was relaxed with RFC 7231](http://tools.ietf.org/html/rfc7231#appendix-B):

> To be consistent with the method-neutral parsing algorithm of
> [RFC7230], the definition of GET has been relaxed so that requests
> can have a body, even though a body has no meaning for GET.
> (Section 4.3.1)

So apparently the change was done for simplifying parser implementations rather than enabling new API styles.

2beaucoup:
FTR: This came from https://groups.google.com/forum/#!topic/akka-user/9Ue5m9YC3l8

sirthias:
Interesting pointer from @analytically:
Dropbox decided to switch one of their API calls from GET to POST to get around the "GETs don't have an entity" problem: https://blogs.dropbox.com/developers/2015/03/limitations-of-the-get-method-in-http/

[...]

PR #18667, opened 2015-10-07
=htc accept entity for DELETE and GET methods

analytically:


ktoso:
Thanks for the contrib!
I think this will need more love however as it would be best if we could do the client side at the same time - and add tests for both cases. I'm not sure from the top of my head how the client part should be implemented - could you give it some thought or maybe @sirthias / @jrudolph / @rkuhn have some hints ready since they discussed the feature in depth previously?

Refs https://github.com/akka/akka/issues/17214

ktoso:
OK TO TEST

jrudolph:
Error lines in https://jenkins.akka.io/job/pr-validator-per-commit-jenkins/3827/:

```
 12854 Failed: Total 478, Failed 1, Errors 0, Passed 477, Pending 22
 12855 Failed tests:
 12856 	akka.http.impl.engine.parsing.RequestParserSpec
 16335 (akka-http-core-experimental/test:test) sbt.TestsFailedException: Tests unsuccessful
 16336 (akka-docs-dev/sphinx:generatePdf) Sphinx pdf generation failed.  See debug output for details.
 16337 Total time: 772 s, completed Oct 7, 2015 6:16:29 PM
```

sirthias:
Looking at this patch as well as at what would be required in order to hide acceptance of DELETE entities behind a config setting I am inclined to simply go for the change that @analytically proposes here.

The risk that a user unintentionally attaches an entity to a DELETE request (and would thus benefit from explicit configuration) is not really significant compared to the amount of structural change and added complexity required for making `HttpMethod#isEntityAccepted` depend on runtime configuration.

So +1 from me for this PR as it stands.

jrudolph:
I agree with @sirthias.

rkuhn:
Should we then go all-in and allow entities on all verbs that are not required to reject payloads?

analytically:
IMO TRACE, CONNECT and HEAD should never allow an entity. GET is debatable.

sirthias:
AFAICS @analytically proposed this change to DELETE because Dropbox somehow decided to use it in its API. How about we simply make the change for DELETE and wait for other users to show us "proof" of real-world usage before going any further?

rkuhn:
For GET I do know that the HTTP spec does not force a rejection and elasticsearch (and others) are using it in their API, so that one should definitely go in as well. For the others I’d have to consult the spec (which I have misplaced once again—which one was it?).

rkuhn:
Okay, [TRACE](https://tools.ietf.org/html/rfc7231#section-4.3.8) is the only one where a payload is forbidden, for all others under discussion it has “no specified semantics”.

analytically:
I will update the PR.

rkuhn:
Thanks!

2beaucoup:
I'd vote for keeping `isEntityAccepted = false` for `CONNECT`, `HEAD` and `TRACE` for now.

analytically:
https://groups.yahoo.com/neo/groups/rest-discuss/conversations/messages/9962

rkuhn:
Okay, so the summary then is: we have seen arguments for why DELETE and GET may need an entity, and absent such arguments we leave CONNECT and HEAD alone. No matter what Mr. Fielding says ;-)

sirthias:
Yes, +1 from my side.

jrudolph:
Error lines in https://jenkins.akka.io/job/pr-validator-per-commit-jenkins/3839/:

```
 14184 Failed: Total 480, Failed 16, Errors 0, Passed 464, Pending 22
 14185 Failed tests:
 14186 	akka.http.impl.engine.client.LowLevelOutgoingConnectionSpec
 14187 	akka.http.impl.engine.rendering.RequestRendererSpec
 17656 (akka-http-core-experimental/test:test) sbt.TestsFailedException: Tests unsuccessful
 17657 Total time: 783 s, completed Oct 8, 2015 12:56:33 PM
```

jrudolph:
Error lines in https://jenkins.akka.io/job/pr-validator-per-commit-jenkins/3844/:

```
 14155 Failed: Total 480, Failed 16, Errors 0, Passed 464, Pending 22
 14156 Failed tests:
 14157 	akka.http.impl.engine.client.LowLevelOutgoingConnectionSpec
 14158 	akka.http.impl.engine.rendering.RequestRendererSpec
 17632 (akka-http-core-experimental/test:test) sbt.TestsFailedException: Tests unsuccessful
 17633 Total time: 801 s, completed Oct 8, 2015 1:50:39 PM
```

analytically:
Tests fine locally.

analytically:
Will someone merge this? The tests run fine locally.

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
```diff
@@ -99,7 +99,7 @@ private[http] class HttpRequestRendererFactory(userAgentHeader: Option[headers.`
       }
 
     def renderContentLength(contentLength: Long) =
-      if (method.isEntityAccepted) r ~~ `Content-Length` ~~ contentLength ~~ CrLf else r
+      if (method.isEntityAccepted && (contentLength > 0 || !method.isSafe)) r ~~ `Content-Length` ~~ contentLength ~~ CrLf else r
```
2beaucoup: what's the purpose of this?

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/RequestParserSpec.scala
```diff
@@ -448,17 +448,26 @@ class RequestParserSpec extends FreeSpec with Matchers with BeforeAndAfterAll {
...
+            |""" should parseToError(422: StatusCode, ErrorInfo("CONNECT requests must not have an entity"))
+        }
+        "with an illegal entity using HEAD" in new Test {
           """HEAD /resource/yes HTTP/1.1
             |Content-length: 3
             |Host: x
             |
-            |foo""" should parseToError(422: StatusCode, ErrorInfo("HEAD requests must not have an entity"))
-          """DELETE /resource/yes HTTP/1.1
+            |""" should parseToError(422: StatusCode, ErrorInfo("HEAD requests must not have an entity"))
```
2beaucoup: "foo" can stay so that it matches `Content-length: 3`.

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
analytically: Safe methods usually don't have a body, hence no need for setting `Content-Length: 0`.

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/RequestParserSpec.scala
analytically: Doesn't make much of a difference, but pushed the change.

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
2beaucoup: `isSafe` has nothing to do with bodies. that's exaclty what `isEntityAccepted` is for.

sirthias: I agree with @2beaucoup in that we shouldn't change the logic of the previous implementation here.

analytically: So that means that we add a `Content-Length: 0` header to GET requests? Seems overkill to me.

2beaucoup: good point. maybe we should change `isEntityAccepted` from Boolean to "yes/no/unkilely" (GET/DELETE = unilkely) and omit rendering of `Content-Length` in the unlikely case if it's zero.

analytically: DELETE is not a safe method. Only GET/HEAD/OPTIONS/TRACE. But only GET and OPTIONS accept an entity.

2beaucoup: `isSafe` = is safe for retry (a.k.a. idempotent).

analytically: isSafe and isIdempotent are not the same.

analytically: I still think this change is valid and correct. What else should I do here?

2beaucoup: One option would be to introduce an `isEntityExpected` prop which is only true for methods with defined payload-semantics and for which `Content-Length` would always be rendered even if = 0.

We could eventually use this to switch between strict and relaxed request parsing.

@sirthias?

sirthias: Won't this line end up rendering all normal (i.e. empty) DELETE requests with a `Content-Length: 0`?
I think @2beaucoup is right, we now have essentially 3 types of HTTP methods:

1. The one where entities are *regular* (OPTIONS, PATCH, POST, PUT)
2. The one where entities are *disallowed* (CONNECT, HEAD, TRACE)
3. The ones where entities are *tolerated* (DELETE, GET)

We accept entities on request with methods whose `requestEntities` property is *regular* or *tolerated*, but we only render `Content-Length: 0` headers for the former type.

So, I'd propose we do the following:

1. Replace the `isEntityAccepted` property with one named `requestEntities`.

2. Introduce a simple ADT:
    ```scala
    sealed trait RequestEntityAcceptance
    object RequestEntityAcceptance {
      case object Regular extends RequestEntityAcceptance
      case object Tolerated extends RequestEntityAcceptance
      case object Disallowed extends RequestEntityAcceptance
    }
    ```

3. Change the rendering logic to only render `Content-Length: 0` headers for requests with `Regular` methods.

WDYT?

2beaucoup: Sounds good! s/Regular/Expected?

sirthias: Yup, `Expected` is better.

analytically: Ok implemented your proposal.

rkuhn:
Thanks a lot, @analytically !

PR #18670, opened 2015-10-08
+doc fails fast when includecode has no content

ktoso:
Will now eagerly log errors (sadly does not fail the build, that's doable with exceptions, but then the cause is hidden inside `last`, so it's less visible - esp. on jenkins).

> [error] /Users/ktoso/code/akka/akka-docs-dev/rst_preprocessed/scala/http/routing-dsl/directives/security-directives/authorize.rst:11: ERROR: Snippet (/../../akka-http/src/main/scala/akka/http/scaladsl/server/directives/SecurityDirectives.scala#authorize) 
> not found!


patriknw:
LGTM

ktoso:
Thanks @patriknw

PR #18672, opened 2015-10-08
+doc #18670 log errors when included snippet empty

ktoso:
Back port of (so we have it on all active branches) https://github.com/akka/akka/pull/18670


ktoso:
Failure was unrelated

Issue #18659, opened 2015-10-07
Remove sugar APIs for FlowGraph construction and introduce canonical alternative


viktorklang:
Rename:
FlowGraph.partial => create (Java) & apply (Scala)
FlowGraph.closed => runnable (Java) & runnable (Scala)

Remove:
SinkApply
SourceApply
FlowApply
BidiFlowApply
SinkCreate
SourceCreate
FlowCreate
BidiFlowCreate

This is not an exhaustive list, the end goal is to create a very small surface-area for the FlowGraph-builder so that it's easy to know what to use and how to use it.

drewhk:
Optimally the fluid DSL should gain new powers as well, providing the most common use cases where the specialized apply methods were used.

viktorklang:
@drewhk Which fluid DSL?

drewhk:
FlowOps

viktorklang:
@drewhk Right, I wasn't sure if you mean that or FlowGraph.Implicits._!

rkuhn:
This was done in #18700.

PR #18675, opened 2015-10-08
DO NOT MERGE YET: -str - Removes tons of sugar, this is a placeholder commit message

viktorklang:
A proposal for #18659


akka-docs-dev/rst/scala/code/docs/stream/CompositionDocSpec.scala
```diff
@@ -137,17 +137,17 @@ class CompositionDocSpec extends AkkaSpec {
...
+    val source = Source.fromGraph( FlowGraph.create() { implicit builder =>
       val merge = builder.add(Merge[Int](2))
       Source.single(0)      ~> merge
       Source(List(2, 3, 4)) ~> merge
 
       // Exposing exactly one output port
-      merge.out
-    }
+      SourceShape(merge.out)
+    })
```
patriknw: I'm not sure this is a good usability improvement?

akka-http-core/src/main/scala/akka/http/impl/util/JavaMapping.scala
```diff
@@ -113,7 +113,7 @@ private[http] object JavaMapping {
       def toScala(javaObject: javadsl.Flow[JIn, JOut, M]): S =
         scaladsl.Flow[SIn].map(inMapping.toJava(_)).viaMat(javaObject)(scaladsl.Keep.right).map(outMapping.toScala(_))
       def toJava(scalaObject: scaladsl.Flow[SIn, SOut, M]): J =
-        javadsl.Flow.wrap {
+        javadsl.Flow.fromGraph {
```
patriknw: +1

akka-http-core/src/main/scala/akka/http/impl/util/JavaMapping.scala
```diff
@@ -123,7 +123,7 @@ private[http] object JavaMapping {
   def javaToScalaAdapterFlow[J, S](implicit mapping: JavaMapping[J, S]): scaladsl.Flow[J, S, Unit] =
     scaladsl.Flow[J].map(mapping.toScala(_))
   def adapterBidiFlow[JIn, SIn, SOut, JOut](implicit inMapping: JavaMapping[JIn, SIn], outMapping: JavaMapping[JOut, SOut]): scaladsl.BidiFlow[JIn, SIn, SOut, JOut, Unit] =
-    scaladsl.BidiFlow.wrap(javaToScalaAdapterFlow(inMapping), scalaToJavaAdapterFlow(outMapping))(scaladsl.Keep.none)
+    scaladsl.BidiFlow.fromGraphsMat(javaToScalaAdapterFlow(inMapping), scalaToJavaAdapterFlow(outMapping))(scaladsl.Keep.none)
```
patriknw: interesting name

akka-stream-tests/src/test/java/akka/stream/javadsl/BidiFlowTest.java
```diff
@@ -177,7 +176,7 @@ public void mustWorkAsAFlowThatIsOpenOnTheRight() throws Exception {
 
   @Test
   public void mustWorkWhenAtopItsInverse() throws Exception {
-    final Flow<Integer,String,BoxedUnit> f = bidi.atop(inverse).join(Flow.<Integer> empty().map(
+    final Flow<Integer,String,BoxedUnit> f = bidi.atop(inverse).join(Flow.of(Integer.class).map(
```
patriknw: +1

akka-docs-dev/rst/scala/code/docs/stream/CompositionDocSpec.scala
viktorklang: The idea with the change is to make the power-level distinction between the FlowGraph-builder and "normal" Flow/Source/Sink-combinators clear and reduce the surface area of the FlowGraph-related APIs.

The pattern is consistent: <Type>.fromGraph( FlowGraph.create() { implicit builder => <Type>Shape(…) }

I'd love to see alternate solutions to reduce the surface area of the FlowGraph API and make the power-level distinction clearer.

akka-http-core/src/main/scala/akka/http/impl/util/JavaMapping.scala
viktorklang: Yeah, I realized that we ought to be consistent with methods that take a function to transform the materialized value.
"Consistency is dog." :)

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
```diff
@@ -65,7 +69,51 @@ private[akka] final case class SingleElementPublisher[T](value: T, name: String)
...
+    private[this] var done: Boolean = false
+    override def cancel(): Unit = {
+      done = true
+      promise.trySuccess(None)
+    }
+
+    override def request(elements: Long): Unit = {
+      if (elements < 1) rejectDueToNonPositiveDemand(subscriber)
+      if (!done) {
+        done = true
```
patriknw: how is the visibility of `done` handled, so that we don't enter here twice?

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
```diff
@@ -65,7 +69,51 @@ private[akka] final case class SingleElementPublisher[T](value: T, name: String)
...
+        promise.future foreach {
+          case Some(v) ⇒
+            tryOnNext(subscriber, v)
+            tryOnComplete(subscriber)
+          case None ⇒
+            tryOnComplete(subscriber)
+        }
```
patriknw: no need to catch `SpecViolation`?

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
viktorklang: request and cancel is only called by Subscriber, and there is only one. :)

akka-docs-dev/rst/scala/code/docs/stream/FlowDocSpec.scala
```diff
@@ -142,17 +143,17 @@ class FlowDocSpec extends AkkaSpec {
...
-        (zip.in1, zip.out)
-    }
+        FlowShape(zip.in1, zip.out)
+    })
 
     //#flow-mat-combine
-    // An empty source that can be shut down explicitly from the outside
-    val source: Source[Int, Promise[Unit]] = Source.lazyEmpty[Int]
+    // An source that can be signalled explicitly from the outside
+    val source: Source[Int, Promise[Option[Int]]] = Source.maybe[Int]
```
drewhk: Hm, I think we should replace this with a tick source, that materializes to a Cancellable. Might be better for this doc example as it is more intuitive to get what the Cancellable does compared to a Promise[Option[Int]]

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
viktorklang: I thought about that and concluded that it would be better (since the Publisher has no logger) to have the ExecutionContext get the `reportFailure` call.

akka-docs-dev/rst/scala/code/docs/stream/CompositionDocSpec.scala
drewhk: Does the documentation text needs to be updated? These are doc samples.

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
viktorklang: Might deserve a comment to explain?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/BidiFlowSpec.scala
```diff
@@ -20,26 +18,22 @@ class BidiFlowSpec extends AkkaSpec with ConversionCheckedTripleEquals {
 
   implicit val mat = ActorMaterializer()
 
-  val bidi = BidiFlow() { b ⇒
-    val top = b.add(Flow[Int].map(x ⇒ x.toLong + 2).withAttributes(name("top")))
-    val bottom = b.add(Flow[ByteString].map(_.decodeString("UTF-8")).withAttributes(name("bottom")))
-    BidiShape(top.inlet, top.outlet, bottom.inlet, bottom.outlet)
-  }
+  val bidi = BidiFlow.fromGraphs(
```
drewhk: hm, `fromGraphs` does not ring so nice.

akka-docs-dev/rst/scala/code/docs/stream/FlowDocSpec.scala
viktorklang: Yeah, I didn't read the docs while refactoring.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/BidiFlowSpec.scala
viktorklang: @drewhk Yeah, I wasn't 100% happy with it either, but the consistency appealed to me. Any suggested alternatives?

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
drewhk: piggybacks on the synchronization of the downstream AFAIK. I think a test-case should be added for source.maybe.runWith(Sink.head) as Sink.head is not backed by an actor either.

akka-stream/src/main/scala/akka/stream/impl/Consts.scala
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.impl
+
+import akka.japi.function.{ Function ⇒ JFun, Function2 ⇒ JFun2 }
+import akka.japi.{ Pair ⇒ JPair }
+
+private[stream] object Consts {
```
drewhk: more like CachedFunctions?

viktorklang: ConstFuns ;-)

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
drewhk: If there is a synchronous downstream than this will throw on the caller'ss (materializer's) thread. Probably not an issue, just wanted to add.

drewhk:
I like it in general, especially the removal of the boilerplates (and the corresponding overload hell). What I don't like is `fromGraphs` and I am not sure about `maybe`. The problem with `fromGraphs`(also the old `wrap` btw) that it does not say what shapes it takes. With `fromGraph` that is not an issue since that is evident, but `Flow.fromGraphs` is not really conweying that it needs a source and sink shaped graph.

akka-stream/src/main/scala/akka/stream/impl/Consts.scala
drewhk: ConstantFun ;)

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
viktorklang: no? It will throw on the ExecutionContext's thread.

viktorklang:
@drewhk `maybe` is a separate commit, so that can easily be discussed in isolation, I am not committed to either, I just wanted to explore the lazyEmpty space a bit. `fromGraphs` I am not married to at all so any suggestion is welcome, it is an improvement over `apply` in either case ;)

akka-stream/src/main/scala/akka/stream/impl/Consts.scala
viktorklang: haha, I like that

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
drewhk: If you use a Sink.head, which is also actorless, then I think it will be:
 - materializer calls subscribe
 - subscribe calls onSubscribe on "head"
 - head requests and assume it has a bug (requests -1 for example)
 - exception is thrown. We are still on the materializing thread, which throws it back to caller. 

Am I missing something here?

viktorklang: @drewhk You're commenting on the wrong line? See line 113, if it really is synchronous then it will throw on tryOnSubscribe, which will throw a SpecViolation, which will be caught and dropped. Or am I missing something?

akka-docs-dev/rst/scala/code/docs/stream/CompositionDocSpec.scala
patriknw: ok, it might be good, so when working with the FlowGraph I only need to learn one thing (but that thing involves more pieces).
Is the compilation error message good, e.g. when returning the wrong thing instead of `SourceShape` or wrong type of `SourceShape`?

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
patriknw: yeah, it's probably not a problem in practise, but a bit subtile (implicit). Even though there is only one subscriber subsequent calls to `request` can be done from different threads.

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
patriknw: ok, please add the comment

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -20,32 +20,27 @@ import scala.language.{ higherKinds, implicitConversions }
...
+   * followed by completion.
+   * If the materialized promise is completed with a None, no value will be produced downstream and completion will
+   * be signalled immediately.
+   * If the materialized promise is completed with a failure, then the returned source will terminate with that error.
+   * If the downstream of this source cancels before the promise has been completed, then the promise will be completed
+   * with None.
    */
-  def lazyEmpty[T](): Source[T, Promise[Unit]] =
-    new Source[T, Promise[Unit]](scaladsl.Source.lazyEmpty)
+  def maybe[T]: Source[T, Promise[Option[T]]] =
```
patriknw: I promise, maybe. :)

(I like it)

akka-docs-dev/rst/scala/code/docs/stream/CompositionDocSpec.scala
ktoso: As long as compilation errors are helpful I like this change. 
Discourages dropping into graph land, and if you do you only care about graph-land.

Docs will need re-writing to explain this.

akka-http-core/src/main/scala/akka/http/impl/util/JavaMapping.scala
ktoso: Does not look very nice but at least it's *very* visually recognisable - "oh, fromX[Mat], I know that."

akka-stream-tck/src/test/scala/akka/stream/tck/PromiseSourceTest.scala
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright (C) 2014 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.tck
+
+import org.reactivestreams.Publisher
+import org.reactivestreams.Subscriber
+import akka.stream.scaladsl.{ Keep, Source, Sink }
+
+class PromiseSourceTest extends AkkaPublisherVerification[Int] {
```
ktoso: `MaybeSourceTest`?

akka-stream-tests/src/test/java/akka/stream/javadsl/BidiFlowTest.java
```diff
@@ -35,84 +35,83 @@ public BidiFlowTest() {
...
+                                      .of(ByteString.class).map(new Function<ByteString, String>() {
+                                          @Override
+                                          public String apply(ByteString arg) {
+                                              return arg.decodeString("UTF-8");
+                                          }
+                                      }));
+                              return new BidiShape<Integer, Long, ByteString, String>(top
+                                      .inlet(), top.outlet(), bottom.inlet(), bottom.outlet());
+                          }
+                      }));
```
ktoso: note: we'll be going with 2-spaces in java code.

akka-stream/src/main/scala/akka/stream/impl/Consts.scala
```diff
@@ -0,0 +1,25 @@
...
+  private[this] val JavaIdentityFunction = new JFun[Any, Any] {
+    @throws(classOf[Exception]) override def apply(param: Any): Any = param
+  }
+
+  val JavaPairFunction = new JFun2[AnyRef, AnyRef, AnyRef JPair AnyRef] {
+    def apply(p1: AnyRef, p2: AnyRef): AnyRef JPair AnyRef = JPair(p1, p2)
+  }
```
ktoso: maybe add a TODO that we should provide those cached functions in `akka.japi.Util` directly

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
```diff
@@ -73,14 +72,14 @@ class SourceSpec extends AkkaSpec {
     }
   }
 
-  "Lazy Empty Source" must {
-    "complete materialized future when stream cancels" in {
-      val neverSource = Source.lazyEmpty
-      val pubSink = Sink.publisher
+  "Maybe Source" must {
```
ktoso: missing test for completing the Promise with a `Some(a)`?

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
```diff
@@ -65,7 +69,51 @@ private[akka] final case class SingleElementPublisher[T](value: T, name: String)
...
     }
-  def apply[T]: Publisher[T] = this.asInstanceOf[Publisher[T]]
+
+  override def toString: String = name
+}
+
+/**
+ * INTERNAL API
+ */
+private[akka] final case class TriggeredOptionPublisher[T](
```
ktoso: make the name match up with the implemented API method - `MaybePublisher`?

akka-docs-dev/rst/scala/code/docs/stream/CompositionDocSpec.scala
viktorklang: If one specifies the wrong return value (not a Shape):

inferred type arguments [(akka.stream.Inlet[Int], akka.stream.Outlet[Int])] do not conform to method create's type parameter bounds [S <: akka.stream.Shape]

If one specifies the wrong shape:

[error]  found   : akka.stream.Graph[akka.stream.SinkShape[Int],akka.actor.Cancellable]
[error]  required: akka.stream.Graph[akka.stream.FlowShape[?,?],?]

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
viktorklang: Yes, but if that is the case, then the Subscriber is responsible for making sure that it is properly migrated between threads.

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
viktorklang: Even better, I decided to also report the caught SpecViolation to ec reportFailure

akka-docs-dev/rst/scala/code/docs/stream/CompositionDocSpec.scala
patriknw: ok, so the errors don't not give any hint about that a `SourceShape` is expected

I guess that is the disadvantage compared to having something like 

    Source.fromGraph { implicit builder =>

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
patriknw: ok, a comment would be nice

akka-docs-dev/rst/scala/code/docs/stream/CompositionDocSpec.scala
viktorklang: I didn't try the exact example in the code above, but to me it signalled what was expected:

`akka.stream.Graph[akka.stream.FlowShape[?,?],?]`

patriknw: ok, I was looking for SourceShape, since the code here was about that. Then it's good, I guess.

viktorklang: Thanks for asking for clarification and apologies for not being clear :)

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
drewhk: @viktorklang I trust you, but if you would write a test case for that it would be even better (since we would see if we break it)

drewhk:
Are there Java samples in need of changing in akka-samples/akka-docs-java-lambda?

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
viktorklang: @drewhk Have we done the same for the other implementations?

drewhk: No, and we probably should. I expect things to work, but who knows for real :)

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
drewhk: +1

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
viktorklang: Then I suggest we open a separate ticket on that so it gets done properly and for all, deal?

drewhk:
LGTM, really like it. So far it seems like we have a chance 1.1 will be net negative in total (excluding test code) :)

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
drewhk: yes sure, make this merged first.

viktorklang:
@drewhk Thanks Endre!

There are 4 things that haven't been done yet:

1) Make sure that everything compiles. (How do I make sure that everything is compiled, even the examples etc?)
2) I haven't renamed FlowGraph.closed to runnable yet, I'll do that tonight
3) The documentation needs attention, I do not know what needs to be updated here.
4) The commit messages needs a bit of love, I can peel this PR apart and create multiple ones for each commit, building on top of each other.

Let me know how you want to proceed with this.
I have just pushed an update addressing most if not all of the comments.

viktorklang:
Something else that needs to get done, we need to update the migration documentation...

drewhk:
> 1) Make sure that everything compiles. (How do I make sure that everything is compiled, even the examples etc?)

in project akka-stream-and-http do a test:compile, that covers all the main code. Most samples are on akka-docs-dev, but there are some in akka-samples/akka-docs-java-lambda which is a separate sbt project (due to Java8) i.e. you need to run those test from that sbt project.

> 3) The documentation needs attention, I do not know what needs to be updated here.

Unfortunately I don't know. The graph sections definitely, and probably the modularity section, too. But there are likely other places as well (and of course the Java counterpart of everything). 

> Something else that needs to get done, we need to update the migration documentation...

We need to start one first :)

viktorklang:
>in project akka-stream-and-http do a test:compile, that covers all the main code. Most samples are on akka-docs-dev, but there are some in akka-samples/akka-docs-java-lambda which is a separate sbt project (due to Java8) i.e. you need to run those test from that sbt project.

Is there any way we can fix that, it is a productivity drain to not be able to catch all of this from a given project. causes friction, at least for me.

>Unfortunately I don't know. The graph sections definitely, and probably the modularity section, too. But there are likely other places as well (and of course the Java counterpart of everything).

Since we're not talking about the "final" version I think the documentation can be incrementally improved.

drewhk:
> Is there any way we can fix that, it is a productivity drain to not be able to catch all of this from a given project. causes friction, at least for me.

This will be better once part of this gets merged into 2.4, where we will not need separate projects for Java8 stuff. Until then, this is what we have.

patriknw:
`akka-docs-dev` is java8 enabled, so I'm not sure why we have `akka-docs-java-lambda` for streams

drewhk:
That's new to me. This is why I was confused then the last time. Then why aren't the doc samples migrated from that project to akka-docs-dev?

pkinsky:
re:
```scala
val source = Source.fromGraph( FlowGraph.create() { implicit builder =>
  /* ... */
  SourceShape(merge.out)
}
```

It's slightly more verbose, but this change is going to make the FlowGraph DSL much, much easier to explain.

viktorklang:
@pkinsky Thanks Paul, that is music to my ears :)

viktorklang:
Superseded by: https://github.com/akka/akka/pull/18700

Issue #17025, opened 2015-03-06
make OneBoundedInterpreter pausable/resumable


rkuhn:
Currently it can only sensibly be paused when the ball has exited at a BoundaryStage (which is implemented in ActorInterpreter). Once we implement more complex fusing we will want to limit the number of steps performed per actor invocation in order to retain the ability to get input elements or termination signals into it.

hepin1989:
you are going to remove ```OneBoundedInterpreter``` in https://github.com/akka/akka/pull/18677,so this one is invalid right?

drewhk:
Yes, this is correct

drewhk:
will close after merged my PR

drewhk:
done

PR #18677, opened 2015-10-08
=str: Replacing OneBoundedInterpreter with GraphInterpreter

drewhk:
Currently fails on Http tests for unknown reasons.


akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
```diff
@@ -25,20 +25,22 @@ object GraphStages {
     val out = Outlet[T]("out")
     override val shape = FlowShape(in, out)
 
-    protected abstract class SimpleLinearStageLogic extends GraphStageLogic(shape) {
-      setHandler(out, new OutHandler {
-        override def onPull(): Unit = pull(in)
-      })
+    protected val basicOutHandler = new OutHandler {
```
drewhk: ah, this is not needed for anything, will remove.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/LifecycleInterpreterSpec.scala
```diff
@@ -138,4 +140,54 @@ class LifecycleInterpreterSpec extends InterpreterSpecKit {
 
   }
 
+  private[akka] case class PreStartAndPostStopIdentity[T](
```
drewhk: This is just a move from InterpreterSpecKit, but I have not yet removed it from there.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
```diff
@@ -564,4 +567,51 @@ class InterpreterSpec extends InterpreterSpecKit {
 
   }
 
+  private[akka] case class Doubler[T]() extends PushPullStage[T, T] {
```
drewhk: This is just a move from InterpreterSpecKit, but I have not yet removed it from there.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphStageTimersSpec.scala
```diff
@@ -39,13 +39,17 @@ class GraphStageTimersSpec extends AkkaSpec {
...
   class TestStage(probe: ActorRef, sideChannel: SideChannel) extends SimpleLinearGraphStage[Int] {
-    override def createLogic = new SimpleLinearStageLogic {
+    override def createLogic(inheritedAttributes: Attributes) = new TimerGraphStageLogic(shape) {
       val tickCount = Iterator from 1
 
       setHandler(in, new InHandler {
         override def onPush() = push(out, grab(in))
       })
 
+      setHandler(out, new OutHandler {
```
drewhk: I removed `SimpleLinearStageLogic` because I extracted timer related stuff to the subclass `TimerGraphStageLogic`

akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
```diff
@@ -31,11 +31,11 @@ object ActorMaterializer {
    * the processing steps. The default `namePrefix` is `"flow"`. The actor names are built up of
    * `namePrefix-flowNumber-flowStepNumber-stepName`.
    */
-  def apply(materializerSettings: Option[ActorMaterializerSettings] = None, namePrefix: Option[String] = None, optimizations: Optimizations = Optimizations.none)(implicit context: ActorRefFactory): ActorMaterializer = {
+  def apply(materializerSettings: Option[ActorMaterializerSettings] = None, namePrefix: Option[String] = None)(implicit context: ActorRefFactory): ActorMaterializer = {
     val system = actorSystemOf(context)
 
     val settings = materializerSettings getOrElse ActorMaterializerSettings(system)
-    apply(settings, namePrefix.getOrElse("flow"), optimizations)(context)
```
drewhk: yeah, optimizations is gone. These should be external graph->graph functions simply.

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
```diff
@@ -92,108 +93,122 @@ private[stream] object Stages {
...
+    override val attributes: Attributes = identityOp
+
+    def apply[T]: SymbolicStage[T, T] = this.asInstanceOf[SymbolicStage[T, T]]
+
+    override def create(): Stage[Any, Any] = fusing.Map(identity, supervision)
   }
 
-  final case class Map(f: Any ⇒ Any, attributes: Attributes = map) extends StageModule {
-    override def withAttributes(attributes: Attributes) = copy(attributes = attributes)
+  final case class Map[In, Out](f: In ⇒ Out, attributes: Attributes = map) extends SymbolicStage[In, Out] {
```
drewhk: yeah, typesafe :)

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -145,12 +150,15 @@ private[stream] object GraphInterpreter {
...
+          val logic = logics(inOwners(i))
+          inHandlers(i) = logic.handlers(ins(i).id).asInstanceOf[InHandler]
+          logics(inOwners(i)).portToConn(ins(i).id) = i
         }
         if (outs(i) ne null) {
-          outHandlers(i) = logics(outOwners(i)).outHandlers(outs(i).id)
-          logics(outOwners(i)).outToConn(outs(i).id) = i
+          val logic = logics(outOwners(i))
+          val inCount = logic.inCount
+          outHandlers(i) = logic.handlers(outs(i).id + inCount).asInstanceOf[OutHandler]
```
drewhk: there are no longer separate `inHandlers` and `outHandlers`, they are all stored in the same array to reduce object header overhead.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -30,14 +31,17 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
...
-        flow.asInstanceOf[javadsl.Flow[In, T, Mat2]].asScala
-      else
-        flow.asInstanceOf[Flow[In, T, Mat2]]
+      val flowInstance: Flow[In, T, Mat2] =
+        if (flow.isInstanceOf[javadsl.Flow[In, T, Mat2]])
+          flow.asInstanceOf[javadsl.Flow[In, T, Mat2]].asScala
+        else if (flow.isInstanceOf[scaladsl.Flow[In, T, Mat2]])
+          flow.asInstanceOf[Flow[In, T, Mat2]]
+        else
+          Flow.wrap(flow).asInstanceOf[Flow[In, T, Mat2]]
```
drewhk: Of course the graph can be, well, a graph, not a flow even if it has that shape :). This fixes that issue.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -188,19 +192,15 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
   }
 
   /** INTERNAL API */
-  override private[stream] def andThen[U](op: StageModule): Repr[U, Mat] = {
+  // FIXME: Only exists to keep old stuff alive
+  private[stream] override def deprecatedAndThen[U](op: StageModule): Repr[U, Mat] = {
```
drewhk: stupid name, but this method will be removed as all stream-of-stream stages are migrated to the GraphStage infrastructure. `StageModule` and this method exist only to support them until removal.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -370,7 +370,7 @@ trait FlowOps[+Out, +Mat] {
    * '''Cancels when''' downstream cancels
    *
    */
-  def recover[T >: Out](pf: PartialFunction[Throwable, T]): Repr[T, Mat] = andThen(Recover(pf.asInstanceOf[PartialFunction[Any, Any]]))
```
drewhk: typesafe, yeah :)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1096,8 +1090,9 @@ trait FlowOps[+Out, +Mat] {
   def withAttributes(attr: Attributes): Repr[Out, Mat]
 
   /** INTERNAL API */
-  private[scaladsl] def andThen[U](op: StageModule): Repr[U, Mat]
+  private[scaladsl] def andThen[T](op: SymbolicStage[Out, T]): Repr[T, Mat] =
+    via(SymbolicGraphStage(op))
```
drewhk: real beauty here, andThen is just a sugar now on top of via. Subclasses (Source, Flow) only need to provide a viaMat implementation now.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -821,17 +820,6 @@ object FlowGraph extends GraphApply {
       override def withAttributes(attr: Attributes): Repr[Out, Mat] =
         throw new UnsupportedOperationException("Cannot set attributes on chained ops from a junction output port")
 
-      override private[scaladsl] def andThen[U](op: StageModule): Repr[U, Mat] = {
```
drewhk: andThen are no longer needed, since FlowOps implements it.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -12,14 +12,15 @@ import akka.stream.impl.fusing.GraphInterpreter.GraphAssembly
 import scala.collection.mutable
 import scala.concurrent.duration.FiniteDuration
 
-abstract class GraphStageWithMaterializedValue[S <: Shape, M] extends Graph[S, M] {
+abstract class GraphStageWithMaterializedValue[+S <: Shape, +M] extends Graph[S, M] {
```
drewhk: Thanks @rkuhn for this!

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -111,20 +100,14 @@ abstract class GraphStageLogic private[stream] (inCount: Int, outCount: Int) {
   /**
    * INTERNAL API
    */
-  private[stream] var inHandlers = Array.ofDim[InHandler](inCount)
-  /**
-   * INTERNAL API
-   */
-  private[stream] var outHandlers = Array.ofDim[OutHandler](outCount)
+  // Using common array to reduce overhead for small port counts
+  private[stream] var handlers = Array.ofDim[Any](inCount + outCount)
```
drewhk: one array instead of two. More ways to break it, but it is more compact, too.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -111,20 +100,14 @@ abstract class GraphStageLogic private[stream] (inCount: Int, outCount: Int) {
...
   /**
    * INTERNAL API
    */
-  private[stream] var inToConn = Array.ofDim[Int](inHandlers.length)
-  /**
-   * INTERNAL API
-   */
-  private[stream] var outToConn = Array.ofDim[Int](outHandlers.length)
+  // Using common array to reduce overhead for small port counts
+  private[stream] var portToConn = Array.ofDim[Int](handlers.length)
```
drewhk: one array instead of two. More ways to break it, but it is more compact, too.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -315,23 +294,60 @@ abstract class GraphStageLogic private[stream] (inCount: Int, outCount: Int) {
...
+   */
+  def preStart(): Unit = ()
+
+  /**
+   * Invoked after processing of external events stopped because the stage is about to stop or fail.
+   */
+  def postStop(): Unit = ()
+}
+
+abstract class TimerGraphStageLogic(_shape: Shape) extends GraphStageLogic(_shape) {
```
drewhk: I pushed this down as a subclass, since it hooks into the beforePreStart/afterPostStop hooks, otherwise it could be a separate support class that you just compose with a GraphStageLogic. If anyone has a good solution what to do here (trait composition won't work for Java) please share.

drewhk: (Btw, this saved a lot of memory for stages not using timers)

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -315,23 +294,60 @@ abstract class GraphStageLogic private[stream] (inCount: Int, outCount: Int) {
...
+  import TimerMessages._
+
+  private var keyToTimers: mutable.Map[Any, Timer] = null // create on-demand
```
drewhk: Any empty mutable Map is costly, but since I moved this functionality to a subclass this optimization can be probably reverted.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -430,7 +442,7 @@ trait InHandler {
   /**
    * INTERNAL API
    */
-  private[stream] var ownerStageLogic: GraphStageLogic = _
+  private[stream] var inOwnerStageLogic: GraphStageLogic = _
```
drewhk: renamed so it does not clash with the same variable in OutHandler. This means that one can implement both traits with one object. This wastes two references (since they both point to the owner logic) but saves an object header which was a measurable size reduction.

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
```diff
@@ -37,6 +39,143 @@ private[stream] object AbstractStage {
   final val NoTerminationPending = 0x8000
   final val BothBalls = UpstreamBall | DownstreamBall
   final val BothBallsAndNoTerminationPending = UpstreamBall | DownstreamBall | NoTerminationPending
+
+  private class PushPullGraphLogic[In, Out, Ext](
```
drewhk: This is the class which adapts a PushPullStage to a GraphStageLogic. Basically an emulation layer on top of the new GraphStage infrastructure.

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
```diff
@@ -37,6 +39,143 @@ private[stream] object AbstractStage {
...
+    private val shape: FlowShape[In, Out],
+    val attributes: Attributes,
+    val stage: AbstractStage[In, Out, Directive, Directive, Context[Out], LifecycleContext])
+    extends GraphStageLogic(shape) with AsyncContext[Out, Ext] {
+
+    final override def materializer: Materializer = interpreter.materializer
+
+    // FIXME: Can be removed from here and from Context
+    override private[stream] def enter(): Unit = ???
```
drewhk: these are no longer needed, I can remove them as soon as I remove the old interpreter.

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
```diff
@@ -37,6 +39,143 @@ private[stream] object AbstractStage {
...
+    override private[stream] def execute(): Unit = ???
+
+    private def ctx: AsyncContext[Out, Ext] = this
+
+    private var absorbed = false
+
+    {
+      // No need to refer to the handle in a private val
+      val handler = new InHandler with OutHandler {
```
drewhk: using one object to reduce space usage.

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
```diff
@@ -37,6 +39,143 @@ private[stream] object AbstractStage {
...
+    }
+
+    final override def holdDownstreamAndPull(): DownstreamDirective = {
+      pull(shape.inlet)
+      null
+    }
+
+    final override def isHoldingDownstream: Boolean = isAvailable(shape.outlet)
+
+    final override def isHoldingUpstream: Boolean = !(isClosed(shape.inlet) || hasBeenPulled(shape.inlet))
```
drewhk: This turns out to be not completely correct, seems like the old interpreter could still report `isHoldingUpstream == true` when the upstream was actually closing. We can either change this behavior to match what I coded here, or we can attempt to emulate the old one.

drewhk:
One Interpreter

rkuhn:
EPIC!

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
viktorklang: Good call. Given the current timer implementation I'm pretty sure there's more memory to save. For instance, how much memory does a mutable.Map take?

viktorklang:
I have to say, this is the most exciting pull request I've read in ages, I love it. Keep up the *awesome* work!

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: A lot. There is a reason I nulled it out instead of having an empty mutable map :)

viktorklang: How about using a TreeMap, and if the key-type can be pinned down to a primitive type, a specialized TreeMap?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -30,14 +31,17 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
...
     if (this.isIdentity) {
-      val flowInstance: Flow[In, T, Mat2] = if (flow.isInstanceOf[javadsl.Flow[In, T, Mat2]])
-        flow.asInstanceOf[javadsl.Flow[In, T, Mat2]].asScala
-      else
-        flow.asInstanceOf[Flow[In, T, Mat2]]
+      val flowInstance: Flow[In, T, Mat2] =
+        if (flow.isInstanceOf[javadsl.Flow[In, T, Mat2]])
+          flow.asInstanceOf[javadsl.Flow[In, T, Mat2]].asScala
+        else if (flow.isInstanceOf[scaladsl.Flow[In, T, Mat2]])
+          flow.asInstanceOf[Flow[In, T, Mat2]]
```
patriknw: perhaps use pattern match?

viktorklang: I like this solution: https://github.com/akka/akka/pull/18675/files#diff-d25227230bef7fd44d678521a8f07f1cR37

patriknw:
looking good

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: yep, that looks nice.

akka-stream-tck/src/test/scala/akka/stream/tck/FusableProcessorTest.scala
```diff
@@ -18,7 +18,7 @@ class FusableProcessorTest extends AkkaIdentityProcessorVerification[Int] {
     implicit val materializer = ActorMaterializer(settings)(system)
 
     // withAttributes "wraps" the underlying identity and protects it from automatic removal
-    Flow[Int].andThen[Int](Identity()).named("identity").toProcessor.run()
+    Flow[Int].via(Stages.identityGraph.asInstanceOf[Graph[FlowShape[Int, Int], Unit]]).named("identity").toProcessor.run()
```
rkuhn: this deserves a proper “factory” (a.k.a. “cast”) method

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
```diff
@@ -4,17 +4,24 @@
...
 
 import scala.util.control.NoStackTrace
 import akka.stream.Supervision
 
-class InterpreterSpec extends InterpreterSpecKit {
+class InterpreterSpec extends AkkaSpec with GraphInterpreterSpecKit {
   import Supervision.stoppingDecider
 
+  /*
+   * These tests were writtern for the previous veryion of the interpreter, the so called OneBoundedInterpreter.
```
rkuhn: version

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
```diff
@@ -294,46 +294,12 @@ private[akka] object ActorProcessorFactory {
...
-      case Split(d, _)                ⇒ (SplitWhereProcessorImpl.props(settings, d), ())
-      case ConcatAll(_)               ⇒ (ConcatAllImpl.props(materializer), ())
-      case StageFactory(mkStage, _)   ⇒ interp(mkStage())
-      case MaterializingStageFactory(mkStageAndMat, _) ⇒
-        val s_m = mkStageAndMat()
-        (ActorInterpreter.props(settings, List(s_m._1), materializer, att), s_m._2)
+      case GroupBy(f, _)         ⇒ (GroupByProcessorImpl.props(settings, f), ())
+      case PrefixAndTail(n, _)   ⇒ (PrefixAndTailImpl.props(settings, n), ())
+      case Split(d, _)           ⇒ (SplitWhereProcessorImpl.props(settings, d), ())
+      case ConcatAll(_)          ⇒ (ConcatAllImpl.props(materializer), ())
```
rkuhn: so these are the remaining ones to be translated …

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
```diff
@@ -92,108 +93,122 @@ private[stream] object Stages {
...
+    extends PushPullGraphStage[In, Out, Ext](
+      symbolicStage.create,
+      symbolicStage.attributes) {
   }
 
-  final case class MaterializingStageFactory(
-    mkStageAndMaterialized: () ⇒ (Stage[_, _], Any),
-    attributes: Attributes = stageFactory) extends StageModule {
-    override def withAttributes(attributes: Attributes) = copy(attributes = attributes)
+  val identityGraph = SymbolicGraphStage[Any, Any, Any](Identity)
```
rkuhn: this should be private (with underscore) plus a public getter that casts it for type T

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: :-)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: nice!

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: good idea

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
rkuhn: emulating the old behavior wastes space, so unless it has real usage advantages I’d go with the current implementation

rkuhn:
Great work, @drewhk ! Looking forward to merging this and #18605 and then ripping out the final pump-based bits and pieces :-)

akka-stream/src/main/scala/akka/stream/stage/Stage.scala
drewhk: agreed!

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
drewhk: yes, and the tcp+tls

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
drewhk: yes, good idea.

akka-stream-tck/src/test/scala/akka/stream/tck/FusableProcessorTest.scala
drewhk: There is one in @viktorklang's pull request that solves this better (using the `fromGraph` if I recall).

viktorklang: Yes, you remember correctly. I think I managed to get the solution quite slick, if I may say so :)

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
viktorklang: Is that due to be done in the guise of this PR or are there other issues for those?

rkuhn: no issues, the presence of these lines is a disturbance in the force that is strong enough by itself

rkuhn: (but no, should not go into this PR, let’s do things in a somewhat orderly fashion)

drewhk:
Updated, will need a rebase

drewhk:
also fixes: https://github.com/akka/akka/issues/18701#issuecomment-150232664

drewhk:
Fixes also: https://github.com/akka/akka/issues/17025

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
```diff
@@ -57,10 +57,10 @@ private object PoolSlot {
       import FlowGraph.Implicits._
 
       val slotProcessor = b.add {
-        Flow[RequestContext].andThenMat { () ⇒
+        Flow.fromProcessor { () ⇒
           val actor = system.actorOf(Props(new SlotProcessor(slotIx, connectionFlow, settings)).withDeploy(Deploy.local),
             slotProcessorActorName.next())
-          (ActorProcessor[RequestContext, List[ProcessorOut]](actor), ())
+          ActorProcessor[RequestContext, List[ProcessorOut]](actor)
         }.mapConcat(identity)
```
viktorklang: use `ConstantFun.scalaIdentityFunction`

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
```diff
@@ -93,112 +96,129 @@ private[stream] object Stages {
...
   }
 
-  final case class Identity(attributes: Attributes = identityOp) extends StageModule {
-    override def withAttributes(attributes: Attributes) = copy(attributes = attributes)
+  object Identity extends SymbolicStage[Any, Any] {
+    override val attributes: Attributes = identityOp
+
+    def apply[T]: SymbolicStage[T, T] = this.asInstanceOf[SymbolicStage[T, T]]
+
+    override def create(attr: Attributes): Stage[Any, Any] = fusing.Map(identity, supervision(attr))
```
viktorklang: Shouldn't there be a specific fusing.Identity such that it can be properly elided?

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
```diff
@@ -93,112 +96,129 @@ private[stream] object Stages {
...
+      symbolicStage.attributes) {
   }
 
-  final case class MaterializingStageFactory(
-    mkStageAndMaterialized: () ⇒ (Stage[_, _], Any),
-    attributes: Attributes = stageFactory) extends StageModule {
-    override def withAttributes(attributes: Attributes) = copy(attributes = attributes)
+  val identityGraph = SymbolicGraphStage[Any, Any, Any](Identity)
+
+  sealed trait SymbolicStage[-In, +Out] {
```
viktorklang: Why is it called SymbolicStage?

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -145,20 +150,21 @@ private[stream] object GraphInterpreter {
...
-            case null ⇒ throw new IllegalStateException(s"no handler defined in stage $l for port ${outs(i)}")
-            case h    ⇒ outHandlers(i) = h
+          val logic = logics(outOwners(i))
+          val inCount = logic.inCount
+          logic.handlers(outs(i).id + inCount) match {
+            case null          ⇒ throw new IllegalStateException(s"no handler defined in stage $logic for port ${outs(i)}")
+            case h: OutHandler ⇒ outHandlers(i) = h
           }
-          l.outToConn(outs(i).id) = i
+          logic.portToConn(outs(i).id + inCount) = i
```
viktorklang: Since `outs(i).id + inCount` is used in several places—store and reuse?

akka-stream/src/main/scala/akka/stream/impl/fusing/IteratorInterpreter.scala
```diff
@@ -3,64 +3,71 @@
...
-      }
+    private var hasNext = input.hasNext
 
-    }
+    setHandler(out, new OutHandler {
+      override def onPull(): Unit = {
+        if (!hasNext) completeStage()
+        else {
+          val elem = input.next()
+          hasNext = input.hasNext
```
viktorklang: why store hasNext before `push(out, elem)`?

~~~scala
val elem = input.next()
push(out, elem)
hasNext = input.hasNext
if (!hasNext)
  complete(out)
~~~

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/GraphInterpreterSpecKit.scala
```diff
@@ -253,4 +259,133 @@ trait GraphInterpreterSpecKit extends AkkaSpec {
...
+
+    def lastEvents(): Set[TestEvent] = {
+      val events = lastEvent
+      lastEvent = Set.empty
+      events
+    }
+
+    class UpstreamOneBoundedProbe[T] extends UpstreamBoundaryStageLogic[T] {
+      val out = Outlet[T]("out")
+      out.id = 0
```
viktorklang: When would not `out.id` be initialized to 0?

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -786,6 +774,12 @@ object FlowGraph extends GraphApply {
 
       override def viaMat[T, Mat2, Mat3](flow: Graph[FlowShape[Out, T], Mat2])(combine: (Mat, Mat2) ⇒ Mat3) =
         throw new UnsupportedOperationException("Cannot use viaMat on a port")
+
+      override private[scaladsl] def deprecatedAndThen[U](op: StageModule): PortOps[U, Mat] = {
+        b.deprecatedAndThen(outlet, op)
+        new PortOps(op.shape.outlet.asInstanceOf[Outlet[U]], b)
```
viktorklang: `op.shape.outlet.as[U]`

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -643,7 +642,7 @@ object FlowGraph extends GraphApply {
       module.shape.outlet.asInstanceOf[Outlet[M]]
```
viktorklang: `module.shape.outlet.as[M]`

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
drewhk: ok

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
drewhk: it is elided properly, Flow, etc does it automatically.

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
drewhk: because you can still extract the symbolic representation of the stage (is it a Map, with which mapping function, etc.)

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
viktorklang: Everyone deserves ConstantFun

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: sure!

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
viktorklang: Ok, so this is never called.

akka-stream/src/main/scala/akka/stream/impl/fusing/IteratorInterpreter.scala
drewhk: ah, this was ported from a PushPull stage that requires push to be the last statement

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/GraphInterpreterSpecKit.scala
drewhk: if there would be multiple outputs. I think this can be pulled up to the parent class though.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
drewhk: oh, nice!

drewhk:
aaaargh, conflicts again! :(

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
drewhk: except if you join an identity flow with an identity flow, but I guess that does not make any sense anyway.

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
viktorklang: Technically it isn't (IS-A) a Stage though?

drewhk: it depends what do you call a Stage. Only subclasses of `Stage`, or every atomic module with `FlowShape`. I usually use the latter, but it can be confusing here.

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
viktorklang: Flow[Int].via(Flow[Int]) does not connect the first to the second,

~~~scala
def viaMat[T, Mat2, Mat3](flow: Graph[FlowShape[Out, T], Mat2])(combine: (Mat, Mat2) ⇒ Mat3): Flow[In, T, Mat3] = {
    if (this.isIdentity) {
      Flow.fromGraph(flow.asInstanceOf[Graph[FlowShape[In, T], Mat2]])
        .mapMaterializedValue(combine(().asInstanceOf[Mat], _))
    } else {
      val flowCopy = flow.module.carbonCopy
      new Flow(
        module
          .fuse(flowCopy, shape.outlet, flowCopy.shape.inlets.head, combine)
          .replaceShape(FlowShape(shape.inlet, flowCopy.shape.outlets.head)))
    }
  }
~~~

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
viktorklang: It's going to be confusing if we use the same word to mean different things in the same context.

What about FlowOp?

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
drewhk: I meant `join()`.

viktorklang:
@drewhk Yeah, may I suggest that the team focus on getting existing PRs into their branches in the order of least conflicts and then focus on shepherding PRs to be merged speedily? (Time wasted rebasing is not time invested. :))

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
viktorklang: You mean joining two identity flows? :-D

That out to be a Noop :)

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
2m: and `written`

akka-stream/src/main/scala/akka/stream/impl/io/IOSettings.scala
```diff
@@ -1,14 +1,12 @@
...
-  def fileIoDispatcher(context: MaterializationContext): String = {
-    val mat = ActorMaterializer.downcast(context.materializer)
-    context.effectiveAttributes.attributeList.collectFirst { case d: Dispatcher ⇒ d.dispatcher } getOrElse {
-      mat.system.settings.config.getString("akka.stream.file-io-dispatcher")
-    }
-  }
-}
\ No newline at end of file
+
+  final val SyncFileSourceDefaultChunkSize = 8192
```
viktorklang: Shouldn't this be taken from akka config?

rkuhn:
I squashed and pushed this branch as wip-RIP-OneBoundedInterpreter2, but no PR yet since it still needs to be rebased.

PR #18689, opened 2015-10-09
=per Clarify concurrency of asyncWriteMessages

patriknw:


akka-persistence/src/main/java/akka/persistence/journal/japi/AsyncWritePlugin.java
```diff
@@ -66,6 +78,8 @@
    * Java API, Plugin API: synchronously deletes all persistent messages up to
    * `toSequenceNr`.
    *
+   * This call is protected with a circuit-breaker.
```
ktoso: good catch

ktoso:
LGTM :)

// "cleaning up open tabs day" ;)

PR #18691, opened 2015-10-09
+htc add missing javadsl.Http.outgoingConnectionTls overload + remove extra params

jrudolph:


akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
```diff
@@ -330,7 +333,6 @@ class Http(system: ExtendedActorSystem) extends akka.actor.Extension {
    * for encryption on the connection.
    */
   def newHostConnectionPoolTls[T](host: String, port: Int,
-                                  options: JIterable[Inet.SocketOption],
```
ktoso: is there an overload with these options already?

jrudolph: No, at some point I moved the options into the settings but it seems I forgot to remove all the parameters I had added before.

ktoso: I see. Makes sense

ktoso:
Thanks!

Issue #17275, opened 2015-04-23
Websocket Client Side API


jrudolph:
Subticket of #16848 to add client-side websocket support. The existing (#17248) websocket implementation can be completely be used with the client-side of the handshake and the actual API still missing.

fommil:
this one is rather important for testing a websocket server, if nothing else

jrudolph:
I think for testing it would suffice if you could access the registered websocket handler Flow to test it without going through the network. This would be consistent with the rest of testing which also doesn't involve parsing / networking / rendering.

fommil:
@jrudolph I like to have a small sample of tests on the actual socket as well. It does catch things.

jrudolph:
I don't object. There will be client-side websocket support but we will add better support for testing Websocket routes to the testkit before that. I created https://github.com/akka/akka/issues/17914 to track this.

jrudolph:
Slowly getting there:

![screenshot from 2015-09-29 15 22 49](https://cloud.githubusercontent.com/assets/9868/10165155/fa68a316-66bd-11e5-8545-0e4d49ce15df.png)

ktoso:
Awesome, thanks for the status update :+1: Almost there now.. :-)

spangaer:
At the risk of coming across inpatient, any idea when you'd submit something?
I'm currently looking at building an http reverse proxy on top of akka http. For allowing WebSocket forwarding, having a client API seems like the shortest road to success (might not be the cleanest). I was just starting to think about what I would need to break open to get it to work when your update came along.

ktoso:
Hi @jrudolph, what's the status here - much work still needed to get it over the finish line?

FYI we're planning a `1.1` release end of this week, it's mostly in order to ship all the things that we've fixed since `1.0`. The next release will merge with Akka 2.4 and thus require JDK8+.

jrudolph:
Thanks, @ktoso. Good to know, I'll try to get it in at least in a first version.

ktoso:
Would be great, thanks!

spangaer:
FYI, checked out your branch and used it in reverse proxy via:
UpgradeToWebsocket.handleMessages(Http().websocketClientFlow(...)) and it works like a charm.
(technically it's a bit more complex as I replicate the websocketClientFlow via websocketClientLayer and so on so I can separately connection address from URI, so I'm grateful that those calls are public)

[...]

Issue #18563, opened 2015-09-25
Include Autobahn WS tests in the build


ktoso:
They're an external python implemented testsuite, we could include them in the build somehow.

autobahn.ws/testsuite/installation.html#testing

@jrudolph's hints on that:

> I thought about including the autobahn suite. I think the main challenges are:
> 
> - whoever runs the tests needs the suite to be installed (or it must be installed every time from scratch, or e.g. be run from a docker container, not sure if someone already tried to package it up in that way)
> - you need to parse the result and feed it back to the build tool
> - you need to specify the set of tests run and maybe tweak which ones need to be strictly passed and which ones don't, I found a few issues where I think the test suite is too strict in what it accepts (IIRC this was about race conditions that affected the closing behavior)

ktoso:
(Claiming it as a "pet project", if someone else wants to jump in and help please ping here)

ktoso:
Note to self: https://github.com/jrudolph/dockerfiles/tree/master/autobahn-testsuite dockerized autobahn suite

ktoso:
I want to do this this week, hopefully won't be too painful to set up.

ktoso:
This is now solved and running as a nightly task.

Tasks performed here:
- some infra cleanups to make it possible to run @jrudolph's autobahn images 
- the job is: https://jenkins.akka.io:8498/job/akka-http-websockets

How it works: 
- basically as described in: https://github.com/akka/akka/blob/release-2.3-dev/akka-http-core/RunWebsocketAutobahnTestSuite.md
  - for client: we run the test server image, then run test app using sbt, download results, then stop the container
  - for server: we start the server (in a weird way, in background, in subshell), we run the test client against it, thanks to `--net=host`, results are generated via `-v` mounted filesystem
- results are reported and stored using the static html reports jenkins plugin
- it's currently set up to run on a4, but all nodes are able to run it (made sure via cheffing all the same way)

Results are accessible as HTML pages:
- here: https://jenkins.akka.io:8498/job/akka-http-websockets/35/
  - client: https://jenkins.akka.io:8498/job/akka-http-websockets/35/Autobahn_TCK_%E2%80%93_WebSocket_Client/
  - server https://jenkins.akka.io:8498/job/akka-http-websockets/35/Autobahn_TCK_%E2%80%93_WebSocket_Server/

Build fails with:

```
[...]

PR #18693, opened 2015-10-09
#17275 Websocket Client implementation

jrudolph:
/cc @sirthias

### DIFF
--- a/akka-http-core/RunWebsocketAutobahnTestSuite.md
+++ b/akka-http-core/RunWebsocketAutobahnTestSuite.md
@@ -1,0 +1,51 @@
+# Test the client side
+
+Start up the testsuite with docker:
+
+```
+docker run -ti --rm=true -p 8080:8080 -p 9001:9001 jrudolph/autobahn-testsuite
+```
+
+Then in sbt, to run all tests, use
+
+```
+akka-http-core-experimental/test:run-main akka.http.impl.engine.ws.WSClientAutobahnTest
+```
+
+or, to run a single test, use
+
+```
+akka-http-core-experimental/test:run-main akka.http.impl.engine.ws.WSClientAutobahnTest 1.1.1
+```
+
+After a run, you can access the results of the run at http://localhost:8080/cwd/reports/clients/index.html.
+
+You can supply a configuration file for autobahn by mounting a version of `fuzzingserver.json` to `/tmp/fuzzingserver.json`
+of the container, e.g. using this docker option:
+
+```
+-v /fullpath-on-host/my-fuzzingserver-config.json:/tmp/fuzzingserver.json
+```
+
+# Test the server side
+
+Start up the test server in sbt:
+
+```
+akka-http-core-experimental/test:run-main akka.http.impl.engine.ws.WSServerAutobahnTest
+```
+
+Then, run the test suite with docker:
+
+```
+docker run -ti --rm=true -v `pwd`/reports:/tmp/server-report jrudolph/autobahn-testsuite-client
+```
+
+This will put the result report into a `reports` directory in the current working directory on the host.
+
+You can supply a configuration file for autobahn by mounting a version of `fuzzingclient.json` to `/tmp/fuzzingclient.json`
+of the container, e.g. using this docker option:
+
+```
+-v /fullpath-on-host/my-fuzzingclient-config.json:/tmp/fuzzingclient.json
+```
--- a/akka-http-core/src/main/scala/akka/http/ClientConnectionSettings.scala
+++ b/akka-http-core/src/main/scala/akka/http/ClientConnectionSettings.scala
@@ -4,6 +4,9 @@
 
 package akka.http
 
+import java.util.Random
+
+import akka.http.impl.engine.ws.Randoms
 import akka.io.Inet.SocketOption
 
 import scala.concurrent.duration.{ Duration, FiniteDuration }
@@ -21,6 +24,7 @@ final case class ClientConnectionSettings(
   connectingTimeout: FiniteDuration,
   idleTimeout: Duration,
   requestHeaderSizeHint: Int,
+  websocketRandomFactory: () ⇒ Random,
   socketOptions: immutable.Traversable[SocketOption],
   parserSettings: ParserSettings) {
 
@@ -36,6 +40,7 @@ object ClientConnectionSettings extends SettingsCompanion[ClientConnectionSettin
       c getFiniteDuration "connecting-timeout",
       c getPotentiallyInfiniteDuration "idle-timeout",
       c getIntBytes "request-header-size-hint",
+      Randoms.SecureRandomInstances, // can currently only be overridden from code
       SocketOptionSettings.fromSubConfig(root, c.getConfig("socket-options")),
       ParserSettings.fromSubConfig(root, c.getConfig("parsing")))
   }
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -5,8 +5,10 @@
 package akka.http.impl.engine.server
 
 import java.net.InetSocketAddress
+import java.util.Random
 
 import akka.http.ServerSettings
+import akka.http.scaladsl.model.ws.Message
 import akka.stream.io._
 import org.reactivestreams.{ Subscriber, Publisher }
 import scala.util.control.NonFatal
@@ -145,7 +147,7 @@ private[http] object HttpServerBluePrint {
 
         // protocol routing
         val protocolRouter = b.add(new WebsocketSwitchRouter())
-        val protocolMerge = b.add(new WebsocketMerge(ws.installHandler))
+        val protocolMerge = b.add(new WebsocketMerge(ws.installHandler, settings.websocketRandomFactory))
 
         protocolRouter.out0 ~> http ~> protocolMerge.in0
         protocolRouter.out1 ~> websocket ~> protocolMerge.in1
@@ -355,7 +357,7 @@ private[http] object HttpServerBluePrint {
         }
       }
   }
-  class WebsocketMerge(installHandler: Flow[FrameEvent, FrameEvent, Any] ⇒ Unit) extends FlexiMerge[ByteString, FanInShape2[ResponseRenderingOutput, ByteString, ByteString]](new FanInShape2("websocketMerge"), Attributes.name("websocketMerge")) {
+  class WebsocketMerge(installHandler: Flow[FrameEvent, FrameEvent, Any] ⇒ Unit, websocketRandomFactory: () ⇒ Random) extends FlexiMerge[ByteString, FanInShape2[ResponseRenderingOutput, ByteString, ByteString]](new FanInShape2("websocketMerge"), Attributes.name("websocketMerge")) {
     def createMergeLogic(s: FanInShape2[ResponseRenderingOutput, ByteString, ByteString]): MergeLogic[ByteString] =
       new MergeLogic[ByteString] {
         var websocketHandlerWasInstalled: Boolean = false
@@ -370,7 +372,13 @@ private[http] object HttpServerBluePrint {
               ctx.emit(bytes); SameState
             case ResponseRenderingOutput.SwitchToWebsocket(responseBytes, handlerFlow) ⇒
               ctx.emit(responseBytes)
-              installHandler(handlerFlow)
+
+              val frameHandler = handlerFlow match {
+                case Left(frameHandler) ⇒ frameHandler
+                case Right(messageHandler) ⇒
+                  Websocket.stack(serverSide = true, maskingRandomFactory = websocketRandomFactory).join(messageHandler)
+              }
+              installHandler(frameHandler)
               ctx.changeCompletionHandling(defaultCompletionHandling)
               websocketHandlerWasInstalled = true
               websocket
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
@@ -6,13 +6,9 @@ package akka.http.impl.engine.ws
 
 import java.util.Random
 
-import akka.http.impl.engine.parsing.ParserOutput.MessageStartError
-
 import scala.collection.immutable
 import scala.collection.immutable.Seq
 import scala.reflect.ClassTag
-
-import akka.parboiled2.util.Base64
 
 import akka.stream.scaladsl.Flow
 
@@ -91,12 +87,10 @@ private[http] object Handshake {
       // FIXME See #18709
       // val extensions = find[`Sec-WebSocket-Extensions`]
 
-      def isValidKey(key: String): Boolean = Base64.rfc2045().decode(key).length == 16
-
       if (upgrade.exists(_.hasWebsocket) &&
         connection.exists(_.hasUpgrade) &&
         version.exists(_.hasVersion(CurrentWebsocketVersion)) &&
-        key.exists(k ⇒ isValidKey(k.key))) {
+        key.exists(k ⇒ k.isValid)) {
 
         val header = new UpgradeToWebsocketLowLevel {
           def requestedProtocols: Seq[String] = clientSupportedSubprotocols
@@ -156,7 +150,7 @@ private[http] object Handshake {
     def buildRequest(uri: Uri, extraHeaders: immutable.Seq[HttpHeader], subprotocols: Seq[String], random: Random): (HttpRequest, `Sec-WebSocket-Key`) = {
       val keyBytes = new Array[Byte](16)
       random.nextBytes(keyBytes)
-      val key = `Sec-WebSocket-Key`(Base64.rfc2045().encodeToString(keyBytes, false))
+      val key = `Sec-WebSocket-Key`(keyBytes)
       val protocol =
         if (subprotocols.nonEmpty) `Sec-WebSocket-Protocol`(subprotocols) :: Nil
         else Nil
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
@@ -4,15 +4,21 @@
 
 package akka.http.impl.engine.ws
 
-import akka.http.scaladsl.model.headers._
-import akka.http.scaladsl.model.ws.{ Message, UpgradeToWebsocket }
-import akka.http.scaladsl.model.{ StatusCodes, HttpResponse, HttpProtocol, HttpHeader }
-import akka.parboiled2.util.Base64
-import akka.stream.Materializer
-import akka.stream.scaladsl.Flow
+import java.util.Random
 
+import scala.collection.immutable
 import scala.collection.immutable.Seq
 import scala.reflect.ClassTag
+
+import akka.parboiled2.util.Base64
+
+import akka.stream.scaladsl.Flow
+
+import akka.http.impl.util._
+
+import akka.http.scaladsl.model.headers._
+import akka.http.scaladsl.model.ws.{ Message, UpgradeToWebsocket }
+import akka.http.scaladsl.model._
 
 /**
  * Server-side implementation of the Websocket handshake
@@ -22,7 +28,8 @@ import scala.reflect.ClassTag
 private[http] object Handshake {
   val CurrentWebsocketVersion = 13
 
-  /*
+  object Server {
+    /*
       From: http://tools.ietf.org/html/rfc6455#section-4.2.1
 
       1.   An HTTP/1.1 or higher GET request, including a "Request-URI"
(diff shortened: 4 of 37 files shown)


ktoso:
Wohoo, great! Will review once back home :)

ktoso:
Wanted to re-read the RFC before reviewing (just did so on the flight), so now reviewing :)

ktoso:
Refs https://github.com/akka/akka/issues/17275

akka-http-core/RunWebsocketAutobahnTestSuite.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+This will put the result report into a `reports` directory in the current working directory on the host.
+
+You can supply a configuration file for autobahn by mounting a version of `fuzzingclient.json` to `/tmp/fuzzingclient.json`
+of the container, e.g. using this docker option:
+
+```
+-v /fullpath-on-host/my-fuzzingclient-config.json:/tmp/fuzzingclient.json
+```
```
ktoso: Thanks for the instructions, will give them a shot and also will want to put these as part of the build - refs https://github.com/akka/akka/issues/18563

akka-http-core/src/main/scala/akka/http/ClientConnectionSettings.scala
```diff
@@ -36,6 +40,7 @@ object ClientConnectionSettings extends SettingsCompanion[ClientConnectionSettin
       c getFiniteDuration "connecting-timeout",
       c getPotentiallyInfiniteDuration "idle-timeout",
       c getIntBytes "request-header-size-hint",
+      Randoms.SecureRandomInstances, // can currently only be overridden from code
```
ktoso: can stay as-is currently (not sure what one might want to substitute here other than the secure random that this yields), maybe make this a `// TODO` comment though, so it's easier to find.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -355,7 +357,7 @@ private[http] object HttpServerBluePrint {
         }
       }
   }
-  class WebsocketMerge(installHandler: Flow[FrameEvent, FrameEvent, Any] ⇒ Unit) extends FlexiMerge[ByteString, FanInShape2[ResponseRenderingOutput, ByteString, ByteString]](new FanInShape2("websocketMerge"), Attributes.name("websocketMerge")) {
+  class WebsocketMerge(installHandler: Flow[FrameEvent, FrameEvent, Any] ⇒ Unit, websocketRandomFactory: () ⇒ Random) extends FlexiMerge[ByteString, FanInShape2[ResponseRenderingOutput, ByteString, ByteString]](new FanInShape2("websocketMerge"), Attributes.name("websocketMerge")) {
```
ktoso: note that this will conflict with @rkuhn's `FlexiMerge => GraphStage` rewrite in https://github.com/akka/akka/pull/18605.
I'd suggest first merging the WS things then Roland's rewrite.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
```diff
@@ -57,69 +64,202 @@ private[http] object Handshake {
...
-  def isWebsocketUpgrade(headers: List[HttpHeader], hostHeaderPresent: Boolean): Option[UpgradeToWebsocket] = {
-    def find[T <: HttpHeader: ClassTag]: Option[T] =
-      headers.collectFirst {
-        case t: T ⇒ t
+    */
+    def isWebsocketUpgrade(headers: List[HttpHeader], hostHeaderPresent: Boolean): Option[UpgradeToWebsocket] = {
+      def find[T <: HttpHeader: ClassTag]: Option[T] =
+        headers.collectFirst {
+          case t: T ⇒ t
+        }
```
ktoso: nice. I always forget how to force this to actually work (by providing the tag)

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
```diff
@@ -57,69 +64,202 @@ private[http] object Handshake {
...
+
+      // val host = find[Host]
+      val upgrade = find[Upgrade]
+      val connection = find[Connection]
+      val key = find[`Sec-WebSocket-Key`]
+      val version = find[`Sec-WebSocket-Version`]
+      // val origin = find[Origin]
```
ktoso: `// FIXME`? Origin validation sounds like a very important thing to me.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
```diff
@@ -57,69 +64,202 @@ private[http] object Handshake {
...
-      headers.collectFirst {
-        case t: T ⇒ t
+    */
+    def isWebsocketUpgrade(headers: List[HttpHeader], hostHeaderPresent: Boolean): Option[UpgradeToWebsocket] = {
+      def find[T <: HttpHeader: ClassTag]: Option[T] =
+        headers.collectFirst {
+          case t: T ⇒ t
+        }
+
+      // val host = find[Host]
```
ktoso: For my understanding: `Host` we don't need at all here, because the `handleWithWebSocket` directive could have been guarded with a `host` directive if the user wanted to restrict acces for a given host (or use host info in the flow).

jrudolph: Basically, I mention all the things here from the list of the headers from the RFC mentioned above. I should maybe add  more comments why things don't need to be checked at this place because they will be validated at another place.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
```diff
@@ -22,7 +28,8 @@ import scala.reflect.ClassTag
 private[http] object Handshake {
   val CurrentWebsocketVersion = 13
```
ktoso: Since we're passing the `autobahn`... 
I guess I don't need to ask about all the spec rules about "when client wants 7 but you only have 13 you must reply with 13" etc, right?

(just one sanity check, since I expect all these things to be covered in the autobahn suite - won't ping those kinds of questions :-))

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
jrudolph: Yes, it is important, but it is optional for Websocket, so we don't handle it on the lower layer. If you want to restrict origins you can do so from routing.

akka-http-core/src/main/scala/akka/http/ClientConnectionSettings.scala
jrudolph: It's changed in tests to support repeatable runs.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
ktoso: Yes please, a small comment on these is enough ("because can be handled by outer directive" or something like that) I "got" that one rather quickly, but maybe someone else would have been puzzled.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -612,12 +692,23 @@ object Http extends ExtensionId[HttpExt] with ExtensionIdProvider {
...
-  case class OutgoingConnection(localAddress: InetSocketAddress, remoteAddress: InetSocketAddress)
+  final case class OutgoingConnection(localAddress: InetSocketAddress, remoteAddress: InetSocketAddress)
+
+  /**
+   * Represents the response to a websocket upgrade request.
+   */
+  sealed trait WebsocketUpgradeResponse {
+    def response: HttpResponse
+  }
+  final case class InvalidUpgradeResponse(response: HttpResponse, cause: String) extends WebsocketUpgradeResponse
```
jrudolph: This needs more documentation.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
```diff
@@ -57,69 +64,202 @@ private[http] object Handshake {
...
+      // val host = find[Host]
+      val upgrade = find[Upgrade]
+      val connection = find[Connection]
+      val key = find[`Sec-WebSocket-Key`]
+      val version = find[`Sec-WebSocket-Version`]
+      // val origin = find[Origin]
+      val protocol = find[`Sec-WebSocket-Protocol`]
+      val supportedProtocols = protocol.toList.flatMap(_.protocols)
+      // FIXME: support extensions
+      // val extensions = find[`Sec-WebSocket-Extensions`]
```
ktoso: To what extent are we planning to support extensions in the "http:next" scope?

ktoso: I'm only aware of one existing extension currently (compression), it's a draft AFAIR hm...

jrudolph: We don't support it in http:next.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
```diff
@@ -57,69 +64,202 @@ private[http] object Handshake {
...
+        }
+
+      // val host = find[Host]
+      val upgrade = find[Upgrade]
+      val connection = find[Connection]
+      val key = find[`Sec-WebSocket-Key`]
+      val version = find[`Sec-WebSocket-Version`]
+      // val origin = find[Origin]
+      val protocol = find[`Sec-WebSocket-Protocol`]
+      val supportedProtocols = protocol.toList.flatMap(_.protocols)
```
ktoso: nitpick: aren't those better described as "`subprotocols`"? https://tools.ietf.org/html/rfc6455#section-1.9

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
```diff
@@ -57,69 +64,202 @@ private[http] object Handshake {
             list of values indicating which extensions the client would like
             to speak.  The interpretation of this header field is discussed
             in Section 9.1.
-   */
-  def isWebsocketUpgrade(headers: List[HttpHeader], hostHeaderPresent: Boolean): Option[UpgradeToWebsocket] = {
-    def find[T <: HttpHeader: ClassTag]: Option[T] =
-      headers.collectFirst {
-        case t: T ⇒ t
+    */
+    def isWebsocketUpgrade(headers: List[HttpHeader], hostHeaderPresent: Boolean): Option[UpgradeToWebsocket] = {
```
2beaucoup: drop the `is` prefix?

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
jrudolph: If there's not yet a test for it yet we should add one and see what happens. The autobahn test suite doesn't really cover the handshake.

jrudolph: I.e. it doesn't cover failing handshakes IIRC.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
ktoso: Okey, makes sense (small comment here please "do CORS stuffs on the routing layer instead"), thanks :)

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
ktoso: Ok, sounds reasonable to me.
I ticketified it explicitly, in case someone has an actual need for it there's a ticket - https://github.com/akka/akka/issues/18709 (I currently don't see the need to jump in on them right away)

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
ktoso: Glad that I asked then! 
Yes, let's please add a test case for "client wants lower version", it's pretty simple as we'll simply reply with what we have instead (13)

refs https://tools.ietf.org/html/rfc6455#section-4.4

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
```diff
@@ -57,69 +64,202 @@ private[http] object Handshake {
...
+        version.exists(_.hasVersion(CurrentWebsocketVersion)) &&
+        key.exists(k ⇒ isValidKey(k.key))) {
+
+        val header = new UpgradeToWebsocketLowLevel {
+          def requestedProtocols: Seq[String] = supportedProtocols
+
+          def handle(handler: Either[Flow[FrameEvent, FrameEvent, Any], Flow[Message, Message, Any]], subprotocol: Option[String]): HttpResponse = {
+            require(subprotocol.forall(chosen ⇒ supportedProtocols.contains(chosen)),
+              s"Tried to choose invalid subprotocol '$subprotocol' which wasn't offered by the client: [${requestedProtocols.mkString(", ")}]")
+            buildResponse(key.get, handler, subprotocol)
```
ktoso: seems good :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
```diff
@@ -57,69 +64,202 @@ private[http] object Handshake {
...
+      3.  A |Connection| header field with value "Upgrade".
+
+      4.  A |Sec-WebSocket-Accept| header field.  The value of this
+          header field is constructed by concatenating /key/, defined
+          above in step 4 in Section 4.2.2, with the string "258EAFA5-
+          E914-47DA-95CA-C5AB0DC85B11", taking the SHA-1 hash of this
+          concatenated value to obtain a 20-byte value and base64-
+          encoding (see Section 4 of [RFC4648]) this 20-byte hash.
+    */
+    def buildResponse(key: `Sec-WebSocket-Key`, handler: Either[Flow[FrameEvent, FrameEvent, Any], Flow[Message, Message, Any]], subprotocol: Option[String]): HttpResponse =
```
ktoso: I'd try to replace the `Either` with a named type.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
```diff
@@ -57,69 +64,202 @@ private[http] object Handshake {
...
+  object Client {
+    case class NegotiatedWebsocketSettings(subprotocol: Option[String])
+
+    /**
+     * Builds a WebSocket handshake request.
+     */
+    def buildRequest(uri: Uri, extraHeaders: immutable.Seq[HttpHeader], subprotocols: Seq[String], random: Random): (HttpRequest, `Sec-WebSocket-Key`) = {
+      val keyBytes = new Array[Byte](16)
+      random.nextBytes(keyBytes)
+      val key = `Sec-WebSocket-Key`(Base64.rfc2045().encodeToString(keyBytes, false))
```
ktoso: Would it make sense to put the `Base...` logic into its apply method (giving it a `keyBytes`)?

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
jrudolph: I had that before but since this is completely internal and `Either` captures exactly what we want to say, it didn't seem worth the extra lines.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
ktoso: good observation, +1 :)

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
```diff
@@ -57,69 +64,202 @@ private[http] object Handshake {
...
+          headerExists(`Sec-WebSocket-Accept`.forKey(key), showExactOther = false)
+
+      expectations(response) match {
+        case None ⇒
+          val subs = response.header[`Sec-WebSocket-Protocol`].flatMap(_.protocols.headOption)
+
+          if (subprotocols.isEmpty && subs.isEmpty) Right(NegotiatedWebsocketSettings(None)) // no specific one selected
+          else if (subs.nonEmpty && subprotocols.contains(subs.get)) Right(NegotiatedWebsocketSettings(Some(subs.get)))
+          else Left(s"response that indicated that the given subprotocol was not supported. (client supported: ${subprotocols.mkString(", ")}, server supported: $subs)")
+        case Some(problem) ⇒ Left(problem)
```
ktoso: validation looks good :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Masking.scala
```diff
@@ -51,12 +51,9 @@ private[http] object Masking {
       def onPush(part: FrameEvent, ctx: Context[FrameEvent]): SyncDirective =
         part match {
           case start @ FrameStart(header, data) ⇒
-            if (header.length == 0) ctx.push(part)
-            else {
```
ktoso: good catch, did autobahn things catch this or did you notice it?

ktoso: (trying to get a feel how useful the autobahn tests are)

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Randoms.scala
```diff
@@ -0,0 +1,15 @@
...
+package akka.http.impl.engine.ws
+
+import java.security.SecureRandom
+import java.util.Random
+
+object Randoms {
+  /** A factory that creates SecureRandom instances */
+  private[http] case object SecureRandomInstances extends (() ⇒ Random) {
+    override def apply(): Random = new SecureRandom()
+  }
```
ktoso: good idea :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Websocket.scala
```diff
@@ -27,6 +27,16 @@ import akka.http.scaladsl.model.ws._
...
 
+  /**
+   * A stack of all the higher WS layers between raw frames and the user API.
+   */
+  def stack(serverSide: Boolean,
+            maskingRandomFactory: () ⇒ Random,
+            closeTimeout: FiniteDuration = 3.seconds): BidiFlow[FrameEvent, Message, Message, FrameEvent, Unit] =
+    masking(serverSide, maskingRandomFactory) atop
+      frameHandling(serverSide, closeTimeout) atop
+      messageAPI(serverSide, closeTimeout)
```
ktoso: :heart: `atop`

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Masking.scala
jrudolph: Yes, that was found by autobahn. The framing protocol details are the main targets of autobahn.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
jrudolph: +1

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
jrudolph: I created #18711 to review the handshake error scenario separately from this PR.

jrudolph: (Because it only concerns the server side)

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
jrudolph: Yes, makes sense.

jrudolph:
Added Java API.

akka-http-core/src/main/scala/akka/http/ClientConnectionSettings.scala
ktoso: ah yes, that is a good reason.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
ktoso: // cc @rkuhn; suggesting we first merge WS client side, then your rewrite.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
ktoso: thanks!

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Masking.scala
ktoso: Thanks for the info! :)

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
jrudolph: In this case, these are the "client-supported sub protocols".

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Masking.scala
jrudolph: Btw. this was undetected before because the logic is only run for the client side of the autobahn test suite which wasn't run before. My own tests for the same direction made the same wrong assumption as the old code (because masking is somewhat useless if you mask zero bytes, it only adds 4 bytes of overhead to the frame, still it is more consistent to require it all the time) :)

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
rkuhn: Yes, makes sense.

ktoso:
Sorry I didn't manage to finish reviewing today, will continue soon. Thanks for the java api!

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -316,6 +316,23 @@ private[http] object StreamUtils {
     }
     Flow[T].transformMaterializing(newForeachStage)
   }
+
+  /**
+   * Similar to Source.lazyEmpty but doesn't rely on materialization. Can only be used once.
```
ktoso: I see, so it exposes the `OneTimeValce#open()` before materialization so you can complete it eagerly... Ok!

jrudolph:
@ktoso when you are finished with your first round of reviewing let me know, I will have another try at the improvements you suggested.

ktoso:
Yep, I'll ping once I did the full first round here - sorry for it being so bursty this week :)

ktoso:
@jrudolph I'm sorry it took so long to review - eng meeting was very intense this time (well, as always I guess).

This looks good, let's address the comments I had and merge @jrudolph!
WDYT I try setting up autobahn tests as a nightly build on monday?

jrudolph:
Thanks for all the comments, @ktoso. I just pushed an updated version.

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Handshake.scala
ktoso: Ok, let's keep it

ktoso:
Looks good, thanks @jrudolph!
Still some small things to work on around WS but I think we'll be able to take them on later, nothing urgent as I see it.

jrudolph:
Cool, thanks for merging!

Issue #18133, opened 2015-08-06
Sink.actorSubscriber() should verify the props?


hepin1989:
I think the props should be verified for the Actor should be a ActorSubscriber,or it was design to like that.

ktoso:
Good catch, we can't always do it (if `Props(new ...)`) but for the class based one we can :)

hepin1989:
then should make producer method privite[akka]

agolubev:
@ktoso, should it be just verification if actor given in ActorSubscriberSink.props inherits ActorSubscriber?
Should the same be done for ActorPublisherSource?

ktoso:
@agolubev both would benefit from doing such check (for their respective classes).

hepin1989:
this one should be close now.

ktoso:
True, thanks @hepin1989!

hepin1989:
Why I am asking ```producer``` is once I really encounter that the type return ```Actor```,I just cann't rememer,maybe the java side.we will see :)

Issue #18187, opened 2015-08-11
make producer of props a method of private[akka]


hepin1989:
eg:```private[akka] def producer:IndirectActorProducer ```

hepin1989:
https://github.com/akka/akka/issues/18133

patriknw:
If I understand the request correctly it is about making it possible to access `producer.actorClass` for validation purposes. Sounds alright to make it `private[akka]`.

PR #18697, opened 2015-10-10
=str #18133 Sink.actorSubscriber() should verify the props

agolubev:
My concern here is that before this Sink.actorSubscriber() should be ActorSubscriber (relaxed) and now it's a "must".
Perhaps review will trigger some discussion.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
@@ -4,6 +4,7 @@
 package akka.stream.scaladsl
 
 import akka.actor.{ ActorRef, Props }
+import akka.stream.actor.ActorSubscriber
 import akka.stream.impl.Stages.DefaultAttributes
 import akka.stream.impl.StreamLayout.Module
 import akka.stream.impl._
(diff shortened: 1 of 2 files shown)


ktoso:
OK TO TEST

agolubev:
Hm, build failed with "Zip - must work in the happy case"

hepin1989:
shoudn't we use the https://github.com/akka/akka/issues/18187?
sometimes you could not get the right class but just ```classof[Actor]```

agolubev:
Perhaps you are right and I need to use Props.producer as Props.actorClass initializes cachedActorClass. However #18187 was merged into master and release-2.3. If nothing changed - current streams development happening in release-2.3-dev so need to merge PR here as well.
@hepin1989  can you specify case when Props.actorClass will give classOf[Actor]?

hepin1989:
@agolubev I think should upgrade the akka dependencies of akka stream to akka 2.3.14 first and then make use of that,am I right?

hepin1989:
@agolubev take a look at ```def create[T <: Actor](creator: Creator[T])``` I have once tested that on worksheet.I think we could only test part of it,it's a practical test,not full,but still make it more safe I think.

agolubev:
Actually release-2.3-dev branch has Akka source as well (not sure about exact version though). I believe we need to wait for message from core team as they are managing merging and dependencies.

hepin1989:
@agolubev yes,anyway ,more typesafe more happy:)

patriknw:
akka-stream has jar dependency to akka-actor, please update that to 2.3.14

ktoso:
Refs #18133

agolubev:
Ok, I merged in @patriknw fix and now sbt compile task can see ```props.producer``` method.
Still the difference in behavior of Props.producer.actorClass and Props.actorClass is following
```scala
  private[this] def cachedActorClass: Class[_ <: Actor] = {
    if (_cachedActorClass eq null)
      _cachedActorClass = producer.actorClass
    _cachedActorClass
  }
```
```_cachedActorClass``` initialized - and that's it. 
I also made few experiments and verify class of props.actorClass  when props constructed with ``` def create[T <: Actor](creator: Creator[T])``` and it returns correct class (not just Actor)
Can add several test cases with "create" if I haven't convinced you.

patriknw:
Looks like `producer` is not needed for this.

patriknw:
LGTM

rkuhn:
Thanks @agolubev !

validation failure filed as #18756

PR #18698, opened 2015-10-10
=per #18751 Persistence: Return completed current events stream for cleaned journal

eshepelyuk:
Fix for #18751

### DIFF
--- a/akka-persistence-query/src/main/scala/akka/persistence/query/journal/leveldb/EventsByPersistenceIdPublisher.scala
+++ b/akka-persistence-query/src/main/scala/akka/persistence/query/journal/leveldb/EventsByPersistenceIdPublisher.scala
@@ -181,7 +181,7 @@ private[akka] class CurrentEventsByPersistenceIdPublisher(
     deliverBuf()
     if (highestSeqNr < toSequenceNr)
       toSeqNr = highestSeqNr
-    if (highestSeqNr == 0L || (buf.isEmpty && currSeqNo > toSequenceNr) || currSeqNo == fromSequenceNr)
+    if (buf.isEmpty && (currSeqNo > toSequenceNr || currSeqNo == fromSequenceNr))
       onCompleteThenStop()
     else
       self ! Continue // more to fetch
--- a/akka-persistence-query/src/main/scala/akka/persistence/query/journal/leveldb/EventsByPersistenceIdPublisher.scala
+++ b/akka-persistence-query/src/main/scala/akka/persistence/query/journal/leveldb/EventsByPersistenceIdPublisher.scala
@@ -181,7 +181,7 @@ private[akka] class CurrentEventsByPersistenceIdPublisher(
     deliverBuf()
     if (highestSeqNr < toSequenceNr)
       toSeqNr = highestSeqNr
-    if (highestSeqNr == 0L || (buf.isEmpty && currSeqNo > toSequenceNr))
+    if (highestSeqNr == 0L || (buf.isEmpty && currSeqNo > toSequenceNr) || currSeqNo == fromSequenceNr)
       onCompleteThenStop()
     else
       self ! Continue // more to fetch
--- a/akka-cluster/src/multi-jvm/scala/akka/cluster/SurviveNetworkInstabilitySpec.scala
+++ b/akka-cluster/src/multi-jvm/scala/akka/cluster/SurviveNetworkInstabilitySpec.scala
@@ -34,7 +34,10 @@ object SurviveNetworkInstabilityMultiJvmSpec extends MultiNodeConfig {
   val eighth = role("eighth")
 
   commonConfig(debugConfig(on = false).withFallback(
-    ConfigFactory.parseString("akka.remote.system-message-buffer-size=100")).
+    ConfigFactory.parseString("""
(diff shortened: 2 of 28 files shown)


typesafehub-validator:
Hi @eshepelyuk,

patriknw:
OK TO TEST

akka-persistence-query/src/main/scala/akka/persistence/query/journal/leveldb/EventsByPersistenceIdPublisher.scala
```diff
@@ -181,7 +181,7 @@ private[akka] class CurrentEventsByPersistenceIdPublisher(
     deliverBuf()
     if (highestSeqNr < toSequenceNr)
       toSeqNr = highestSeqNr
-    if (highestSeqNr == 0L || (buf.isEmpty && currSeqNo > toSequenceNr))
+    if (highestSeqNr == 0L || (buf.isEmpty && currSeqNo > toSequenceNr) || currSeqNo == fromSequenceNr)
```
patriknw: so that extra condition means that nothing was replayed?
do we need the `highestSeqNr == 0L` ?

patriknw:
do we have the same issue in `CurrentEventsByTagPublisher`?

eshepelyuk:
Let me try to add a test for `CurrentEventsByTagPublisher` and check the behavior. If there's the same error - I will try to fix it.
Sounds ok ?

akka-persistence-query/src/main/scala/akka/persistence/query/journal/leveldb/EventsByPersistenceIdPublisher.scala
patriknw: Note that there is a `maxBufSize` which means that the buf might be half filled from a first replay request and then when this is called for the second replay request we must still drain the buffer before completing the stream. Shouldn't the condition be `buf.isEmpty && (currSeqNo > toSequenceNr || currSeqNo == fromSequenceNr)`?

eshepelyuk: I've tried to remove `highestSeqNr == 0L` and use `buf.isEmpty && (currSeqNo > toSequenceNr || currSeqNo == fromSequenceNr)`.
Tests are working, I've added some more cases.

eshepelyuk:
- modified condition of stream completion for `currentEventsByPersistenceId` query
- test for `currentEventsByTag` after journal cleanup

akka-persistence-query/src/main/scala/akka/persistence/query/journal/leveldb/EventsByPersistenceIdPublisher.scala
```diff
@@ -181,7 +181,7 @@ private[akka] class CurrentEventsByPersistenceIdPublisher(
     deliverBuf()
     if (highestSeqNr < toSequenceNr)
       toSeqNr = highestSeqNr
-    if (highestSeqNr == 0L || (buf.isEmpty && currSeqNo > toSequenceNr))
+    if (buf.isEmpty && (currSeqNo > toSequenceNr) || currSeqNo == fromSequenceNr)
```
patriknw: I'd find it more readable as:

    if (buf.isEmpty && (currSeqNo > toSequenceNr || currSeqNo == fromSequenceNr))

eshepelyuk: Sorry, that was a weird typo. Fixed

akka-persistence-query/src/test/scala/akka/persistence/query/journal/leveldb/EventsByTagSpec.scala
```diff
@@ -121,6 +123,23 @@ class EventsByTagSpec extends AkkaSpec(EventsByTagSpec.config)
...
+      val b = system.actorOf(TestActor.props("b"))
+      b ! TestActor.DeleteCmd(Long.MaxValue)
+      expectMsg(s"${Long.MaxValue}-deleted")
+
+      val c = system.actorOf(TestActor.props("c"))
+      c ! TestActor.DeleteCmd(Long.MaxValue)
+      expectMsg(s"${Long.MaxValue}-deleted")
+
+      val src = queries.currentEventsByTag("green", Long.MaxValue)
+      src.map(_.event).runWith(TestSink.probe[Any]).request(1).expectComplete()
```
patriknw: note that `Long.MaxValue` is the starting offset, i.e. corresponding to `fromSequenceNr` so this test doesn't really do what the title says.

Tags work differently because events are not deleted from the tag-stream when you do `deleteMessages`. There is currently no support for deleting tagged events.

Anyway, I suggest that you change the corresponding condition in the same way in `CurrentEventsByTagPublisher`, given that it does not break any other test.

eshepelyuk:
Pushed suggested improvement, deleted irrelevent test.

patriknw:
LGTM, thanks. Please create an issue so that we can track this fix. Squash to one commit.

eshepelyuk:
I will close this PR and reopen new one.

PR #18705, opened 2015-10-11
=typ #18699 pass Unhandled message to unhandled()

rkuhn:


ktoso:
Nice catch, LGTM

akka-typed/src/test/scala/akka/typed/ActorContextSpec.scala
```diff
@@ -426,6 +437,29 @@ class ActorContextSpec extends TypedSpec(ConfigFactory.parseString(
...
+          child ! Stop
+      }.expectFailureKeep(500.millis) {
+        case (f, (subj, child)) ⇒
+          f.child should ===(subj)
+          Failed.Stop
+      }.expectMessage(500.millis) {
+        case (msg, (subj, child)) ⇒
+          msg should ===(GotSignal(PostStop))
+      }
+    })
```
ktoso: ah, it's been a while since we had a commit in typed, reminds me how nice to read these are :)

johanandren:
LGTM

akka-typed/src/main/scala/akka/typed/Impl.scala
```diff
@@ -20,24 +21,35 @@ private[typed] class ActorAdapter[T](_initialBehavior: () ⇒ Behavior[T]) exten
...
-  private def next(b: Behavior[T]): Unit = {
+  private def next(b: Behavior[T], msg: Any): Unit = {
+    if (isUnhandled(b)) unhandled(msg)
     behavior = canonicalize(ctx, b, behavior)
     if (!isAlive(behavior)) {
       context.stop(self)
     }
   }
 
+  override def unhandled(msg: Any): Unit = msg match {
```
drewhk: is this method overridable by users? If so, isn't it dangerous to let them override the DeathPact behavior?

rkuhn: No, this has always been the case, people know what they are doing :-)

drewhk: Oh, so you can override this in the current actors as well? Wow.

drewhk: Can we change this in typed? As in have an internal hook that only we can override, and have the DeathPact handling there?

rkuhn: Yes, you’re looking at exactly that: Behavior has no way to interfere with the underlying Actor implementation.

patriknw:
LGTM, please merge if this is done

patriknw:
Refs #18699

PR #18707, opened 2015-10-12
Agolubev #17338 add output stream source graph stage based

agolubev:
This is for ticket #17338. 
Current brunch implements OuputStreamSource with GraphStageWithMaterializedValue as @drewhk suggested (InsputStreamSink stays unchanged). First version is in PR #17995 and based on ActorPublisher/ActorSubscriber.
Please let me know if this approach is good enough to fix the ticket.
Currently one test javadsl/scaladsl is failed as in javadsl new apply method is missed (had to introduce scaladsl.SourceApply.apply with builder materializer type)

### DIFF
--- a/akka-stream-tests/src/test/java/akka/stream/io/InputStreamSinkTest.java
+++ b/akka-stream-tests/src/test/java/akka/stream/io/InputStreamSinkTest.java
@@ -1,0 +1,48 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.io;
+
+import akka.japi.Pair;
+import akka.stream.StreamTest;
+import akka.stream.javadsl.AkkaJUnitActorSystemResource;
+import akka.stream.javadsl.Sink;
+import akka.stream.javadsl.Source;
+import akka.stream.testkit.AkkaSpec;
+import akka.stream.testkit.Utils;
+import akka.util.ByteString;
+import org.junit.ClassRule;
+import org.junit.Test;
+import scala.concurrent.Future;
+import scala.concurrent.duration.FiniteDuration;
+
+import java.io.InputStream;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.assertTrue;
+
+public class InputStreamSinkTest  extends StreamTest {
+    public InputStreamSinkTest() {
+        super(actorSystemResource);
+    }
+
+    @ClassRule
+    public static AkkaJUnitActorSystemResource actorSystemResource = new AkkaJUnitActorSystemResource("InputStreamSink",
+            Utils.UnboundedMailboxConfig());
+    @Test
+    public void mustReadEventViaInputStream() throws Exception {
+        final FiniteDuration timeout = FiniteDuration.create(300, TimeUnit.MILLISECONDS);
+
+        final Sink<ByteString, InputStream> sink = InputStreamSink.create(timeout);
+        final List<ByteString> list = Collections.singletonList(ByteString.fromString("a"));
+        final InputStream stream = Source.from(list).runWith(sink, materializer);
+
+        byte[] a = new byte[1];
+        stream.read(a);
+        assertTrue(Arrays.equals("a".getBytes(), a));
+    }
+
+}
--- a/akka-stream-tests/src/test/java/akka/stream/io/OutputStreamSourceTest.java
+++ b/akka-stream-tests/src/test/java/akka/stream/io/OutputStreamSourceTest.java
@@ -1,0 +1,56 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.io;
+
+import akka.actor.ActorRef;
+import akka.japi.Pair;
+import akka.japi.function.Procedure;
+import akka.stream.StreamTest;
+import akka.stream.javadsl.AkkaJUnitActorSystemResource;
+import akka.stream.javadsl.Keep;
+import akka.stream.javadsl.Sink;
+import akka.stream.javadsl.Source;
+import akka.stream.testkit.AkkaSpec;
+import akka.stream.testkit.Utils;
+import akka.testkit.JavaTestKit;
+import akka.util.ByteString;
+import com.typesafe.config.ConfigFactory;
+import org.junit.ClassRule;
+import org.junit.Test;
+import scala.concurrent.Future;
+import scala.concurrent.duration.FiniteDuration;
+import scala.runtime.BoxedUnit;
+
+import java.io.OutputStream;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.assertEquals;
+
+public class OutputStreamSourceTest extends StreamTest {
+    public OutputStreamSourceTest() {
+        super(actorSystemResource);
+    }
+
+    @ClassRule
+    public static AkkaJUnitActorSystemResource actorSystemResource = new AkkaJUnitActorSystemResource("OutputStreamSource",
+            Utils.UnboundedMailboxConfig());
+    @Test
+    public void mustSendEventsViaOutputStream() throws Exception {
+        final FiniteDuration timeout = FiniteDuration.create(300, TimeUnit.MILLISECONDS);
+        final JavaTestKit probe = new JavaTestKit(system);
+
+        final Source<ByteString, OutputStream> source = OutputStreamSource.create(timeout);
+        final OutputStream s = source.to(Sink.foreach(new Procedure<ByteString>() {
+            public void apply(ByteString elem) {
+                probe.getRef().tell(elem, ActorRef.noSender());
+            }
+        })).run(materializer);
+
+        s.write("a".getBytes());
+        assertEquals(ByteString.fromString("a"), probe.receiveOne(timeout));
+        s.close();
+
+    }
+
+}
--- a/akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
@@ -1,0 +1,241 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.io
+
+import java.io.{ IOException, InputStream }
+import java.util.concurrent.TimeoutException
+
+import akka.actor.{ ActorSystem, NoSerializationVerificationNeeded }
+import akka.stream._
+import akka.stream.impl.StreamSupervisor.Children
+import akka.stream.impl.io.InputStreamSinkStage
+import akka.stream.impl.{ ActorMaterializerImpl, StreamSupervisor }
+import akka.stream.scaladsl.{ Keep, Sink }
+import akka.stream.stage.InHandler
+import akka.stream.testkit.AkkaSpec
+import akka.stream.testkit.Utils._
+import akka.stream.testkit.scaladsl.TestSource
+import akka.testkit.TestProbe
+import akka.util.ByteString
+
+import scala.concurrent.duration._
+import scala.concurrent.{ Await, Future }
+import scala.util.Random
+import scala.util.control.NoStackTrace
+
+class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
+  import system.dispatcher
+
+  val settings = ActorMaterializerSettings(system).withDispatcher("akka.actor.default-dispatcher")
+  implicit val materializer = ActorMaterializer(settings)
+
+  val timeout = 300.milliseconds
+  def randomArray(size: Int) = {
+    val a = new Array[Byte](size)
+    Random.nextBytes(a)
+    a
+  }
+
+  val byteArray = randomArray(3)
+  val byteString = ByteString(byteArray)
+
+  def newArray() = new Array[Byte](3)
+
+  def expectSuccess[T](f: Future[T], value: T) =
+    Await.result(f, timeout) should be(value)
+
+  object InputStreamSinkTestMessages {
+    case object Push extends NoSerializationVerificationNeeded
+    case object Finish extends NoSerializationVerificationNeeded
+    case class Failure(ex: Throwable) extends NoSerializationVerificationNeeded
+  }
+
+  def testSink(probe: TestProbe): Sink[ByteString, InputStream] = {
+    class InputStreamSinkTestStage(val timeout: FiniteDuration)
+      extends InputStreamSinkStage(timeout) {
+
+      override def createLogicAndMaterializedValue = {
+        val (logic, inputStream) = super.createLogicAndMaterializedValue
+        val inHandler = logic.inHandlers(in.id)
+        logic.inHandlers(in.id) = new InHandler {
+          override def onPush(): Unit = {
+            probe.ref ! InputStreamSinkTestMessages.Push
+            inHandler.onPush()
+          }
+          override def onUpstreamFinish(): Unit = {
+            probe.ref ! InputStreamSinkTestMessages.Finish
+            inHandler.onUpstreamFinish()
+          }
+          override def onUpstreamFailure(ex: Throwable): Unit = {
+            probe.ref ! InputStreamSinkTestMessages.Failure(ex)
+            inHandler.onUpstreamFailure(ex)
+          }
+        }
+        (logic, inputStream)
+      }
+    }
+    Sink.fromGraph(new InputStreamSinkTestStage(timeout))
+  }
+
+  "InputStreamSink" must {
+    "read bytes from InputStream" in assertAllStagesStopped {
+      val (probe, inputStream) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+
+      probe.sendNext(byteString)
+      val arr = newArray()
+      inputStream.read(arr)
+      arr should ===(byteArray)
+
+      probe.sendComplete()
+      inputStream.close()
+    }
+
+    "read bytes correctly if requested by InputStream not in chunk size" in assertAllStagesStopped {
+      val sinkProbe = TestProbe()
+      val (probe, inputStream) = TestSource.probe[ByteString].toMat(testSink(sinkProbe))(Keep.both).run()
+
+      probe.sendNext(byteString)
+      val byteArray2 = randomArray(3)
+      probe.sendNext(ByteString(byteArray2))
+
+      sinkProbe.expectMsgAllOf(InputStreamSinkTestMessages.Push, InputStreamSinkTestMessages.Push)
+
+      val arr = new Array[Byte](2)
+      inputStream.read(arr)
+      arr should ===(Array(byteArray(0), byteArray(1)))
+      inputStream.read(arr)
+      arr should ===(Array(byteArray(2), byteArray2(0)))
+      inputStream.read(arr)
+      arr should ===(Array(byteArray2(1), byteArray2(2)))
+
+      inputStream.close()
+    }
+
+    "returns less than was expected when the data source has provided some but not enough data" in assertAllStagesStopped {
+      val (probe, inputStream) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+
+      val data = randomArray(2)
+      probe.sendNext(ByteString(data))
+      val arr = newArray()
+      inputStream.read(arr) should ===(2)
+      arr should ===(Array(data(0), data(1), 0))
+
+      probe.sendComplete()
+      inputStream.close()
+    }
+
+    "block read until get requested number of bytes from upstream" in assertAllStagesStopped {
+      val (probe, inputStream) = TestSource.probe[ByteString].toMat(InputStreamSink())(Keep.both).run()
+
+      val arr = newArray()
+      val f = Future(inputStream.read(arr))
+      the[Exception] thrownBy Await.result(f, timeout) shouldBe a[TimeoutException]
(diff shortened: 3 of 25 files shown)


ktoso:
OK TO TEST

akka-stream-tests/src/test/java/akka/stream/io/OutputStreamSourceTest.java
```diff
@@ -0,0 +1,56 @@
...
+        final JavaTestKit probe = new JavaTestKit(system);
+
+        final Source<ByteString, OutputStream> source = OutputStreamSource.create(timeout);
+        final OutputStream s = source.to(Sink.foreach(new Procedure<ByteString>() {
+            public void apply(ByteString elem) {
+                probe.getRef().tell(elem, ActorRef.noSender());
+            }
+        })).run(materializer);
+
+        s.write("a".getBytes());
```
viktorklang: What encoding?

agolubev: Hm, OutputStream works with byte array, OutputStreamSource with ByteString which is wrapper for byte array as well. 
To be fully consistent with converting string to bytes and then string to ByteString will use getBytes instead of fromString below.

agolubev:
Let me summarize my experience for GraphStage(WithMaterializedValue) API (main code is placed in ```OutputStreamSourceStage```), I believe some things are missed and some I just did not find:
1) ```GraphStageWithMaterializedValue``` is a good way of dealing with actors underneath with help of some wrapper. So we are thread safe with actor underneath but we don’t dealing with sending messages just calling methods
2) Source module boilerplate code is not needed anymore and can be removed
3) As soon as data (byte arrays) is not going through actor messages (however it will go through stream as akka messages anyway) and because I can not use ```GraphStage``` with ```akka.stream.blocking-io-dispatcher``` I removed such dependency
4) Source.apply has builder with Mat as [Unit] - so need to introduce another factory method to have it as generic [Mat]
5) Code that’s creating Source from ```GraphStageWithMaterializedValue``` looks strange
```scala
 Source.apply { implicit b: FlowGraph.Builder[OutputStream] ⇒
      val source = b.add(new OutputStreamSourceStage(16, timeout))
      source.outlet
    }
```
6) ```GraphStage``` is useful for some flow transformations and supposed to be used with Builder and Graphs a lot. Now it’s uses only as layer to abstract over ```SubscriberActor```
7) Did not find a way to get actor settings from within ```GraphStage```. It supposed to be read buffer size max from there.
8) Also was trying to implement blocking call (flush for example) with just ```() => Future[Unit]``` from ```OutputStreamAdapter``` to ```GraphStage``` (```GraphStage``` does not have API for calls from outside of reactive stream context). Future[Unit] will be finished when all elements sent downstream on stream is canceled.
```scala
val logic = new GraphStageLogic {
     def wakeUp(msg: AdapterToStageMessage): Future[Unit] = {
        implicit val ex = interpreter.materializer.executionContext
        val p = Promise[Unit]
        Future(message.invoke((msg, p)))
        p.future
      }
      ......
}

(logic, new OutputStreamAdapter(dataQueue, downstreamStatus, logic.wakeUp, timeout))
```
The following code gets me warning: "reflective access of structural type member value message should be enabled"
Did not find the simple way to fix it (believe this will impact performance in a bad way)

Also few points for general approach
1) The main purpose for using ```BlockingQueue``` was to reduce number of big data messages with byte arrays. However downstream will send bytes in messages anyway. So ```OutputStreamSource``` suppose to be just another node that transferring data in messages. My point is there will be performance impact anyway just because of nature of standard stream components. So this is performance of one stream element versus consistency (with other IO Sinks/Sources) choice
2) Removing ask’s with block calls to ```BlockingQueues```... It’s not always a good idea to have ```BlockingQueue``` with one element (like for status). Also we should have timeouts anyway to support fail fast idea. So will have some kind of timeouts anyway.
3) In previous implementation Source was materialized to ```(OutputStream, Future[Long])``` with number of written bytes when it’s completed. This was done alike other ```InputStreamSource``` and ```OutputStreamSink```. Those are behave as pipes and have lack of visibility so that was the reason Future[Long] was introduced. ```OutputStreamSource```, on the other side, requires explicit writing calls so developer will have pretty much visibility on what’s going on there. With this in mind I removed ```Future[Long]``` from ```OutputStreamSource```

To sum up, this is third general refactoring (or reimplementation) of this functionality. I believe we should have pretty good vision on implementation options for now. So let me know what approach I need to stick to to finalize this fix (hope that last one is good enough :) ).

akka-stream/src/main/scala/akka/stream/io/OutputStreamSource.scala
```diff
@@ -0,0 +1,61 @@
...
+  import scala.language.postfixOps
+
+  /**
+   * Creates a synchronous (Java 6 compatible) Source.
+   *
+   * It materializes a [[Future]] containing the number of bytes written to OutputStream upon completion and
+   * [[OutputStream]] to interacting with reactive stream.
+   */
+  def apply(timeout: FiniteDuration = 5.seconds): Source[ByteString, OutputStream] = {
+    Source.apply2 { implicit b: FlowGraph.Builder[OutputStream] ⇒
```
drewhk: Since `OutputStreamSourceStage` has the right shape already, all you need is:

    Source.wrap(new OutputStreamSourceStage(...))

drewhk: Which reads roughly like "add the nice DSL to the thing whch is already a Source-linke thing"

akka-stream-tests/src/test/java/akka/stream/io/OutputStreamSourceTest.java
viktorklang: String.getBytes uses the platform encoding whereas ByteString.fromString uses UTF-8

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamPublisher.scala
```diff
@@ -0,0 +1,183 @@
...
+    } else throw new IOException("OutputStream is closed")
+
+  @scala.throws(classOf[IOException])
+  override def write(b: Int) = {
+    sendData(ByteString(b))
+  }
+
+  @scala.throws(classOf[IOException])
+  override def write(b: Array[Byte], off: Int, len: Int) = {
+    sendData(ByteString(b).drop(off))
```
viktorklang: Most probably cheaper to do: ByteString(b, off, len)

akka-stream-tests/src/test/java/akka/stream/io/InputStreamSinkTest.java
```diff
@@ -0,0 +1,48 @@
...
+    public void mustReadEventViaInputStream() throws Exception {
+        final FiniteDuration timeout = FiniteDuration.create(300, TimeUnit.MILLISECONDS);
+
+        final Sink<ByteString, Pair<InputStream,Future<Long>>> sink = InputStreamSink.create(timeout);
+        final List<ByteString> list = Collections.singletonList(ByteString.fromString("a"));
+        final Pair<InputStream, Future<Long>> p = Source.from(list).runWith(sink, materializer);
+
+        byte[] a = new byte[1];
+        p.first().read(a);
+        assertTrue(Arrays.equals("a".getBytes(), a));
```
viktorklang: Put a specific character encoding for all invocations of String.getBytes(), otherwise it'll use platform default.

viktorklang: https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#getBytes()

akka-stream-tests/src/test/java/akka/stream/io/InputStreamSinkTest.java
```diff
@@ -0,0 +1,48 @@
...
+    @ClassRule
+    public static AkkaJUnitActorSystemResource actorSystemResource = new AkkaJUnitActorSystemResource("InputStreamSink",
+            Utils.UnboundedMailboxConfig());
+    @Test
+    public void mustReadEventViaInputStream() throws Exception {
+        final FiniteDuration timeout = FiniteDuration.create(300, TimeUnit.MILLISECONDS);
+
+        final Sink<ByteString, Pair<InputStream,Future<Long>>> sink = InputStreamSink.create(timeout);
+        final List<ByteString> list = Collections.singletonList(ByteString.fromString("a"));
+        final Pair<InputStream, Future<Long>> p = Source.from(list).runWith(sink, materializer);
```
viktorklang: Iso using List here you can use Source.single(ByteString(…))

akka-stream-tests/src/test/java/akka/stream/io/OutputStreamSourceTest.java
agolubev: yeah, I'll fix this

akka-stream/src/main/scala/akka/stream/io/OutputStreamSource.scala
agolubev: Thanks, that's what I missed

akka-stream-tests/src/test/java/akka/stream/io/InputStreamSinkTest.java
agolubev: it's using file.encoding system property. I got it. Will change it to have UTC-8

viktorklang: Make that UTF-8 :)

akka-stream-tests/src/test/java/akka/stream/io/OutputStreamSourceTest.java
viktorklang: awesome, thanks!

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
```diff
@@ -0,0 +1,243 @@
...
+
+class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
+  import system.dispatcher
+
+  val settings = ActorMaterializerSettings(system).withDispatcher("akka.actor.default-dispatcher")
+  implicit val materializer = ActorMaterializer(settings)
+
+  val timeout = 300.milliseconds
+  val byteString = ByteString("abc")
+  val byteArray = "abc".toArray
```
viktorklang: java.lang.String does not have a toArray method, so I assume it is added by Scala, and I don't know what char encoding it uses. I'd recommend doing: ByteString.fromString("abc", "UTF-8").toArray

viktorklang: One idea would be to not use Strings for the byte arrays, since the values themselves are not interesting.
instead, it would be easier to generate random byte-arrays of a given length and compare to those instead.

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
```diff
@@ -0,0 +1,135 @@
...
+  import system.dispatcher
+
+  val settings = ActorMaterializerSettings(system)
+  implicit val materializer = ActorMaterializer(settings)
+
+  val timeout = 300.milliseconds
+  val bytes = "abc".getBytes
+  val byteString = ByteString("abc")
+
+  def expectBlocked[T](f: Future[T]) =
```
viktorklang: blocked?

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
```diff
@@ -0,0 +1,135 @@
...
+      val (outputStream, probe) = OutputStreamSource().toMat(TestSink.probe[ByteString])(Keep.both).run
+
+      probe.expectSubscription()
+      outputStream.close()
+      probe.expectComplete()
+      the[Exception] thrownBy outputStream.write(bytes) shouldBe a[IOException]
+    }
+
+    "throw IOException when writing to the stream after the subscriber has cancelled the reactive stream" in assertAllStagesStopped {
+      pending
```
viktorklang: Why pending?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphBackedFlowSpec.scala
```diff
@@ -141,7 +141,7 @@ class GraphFlowSpec extends AkkaSpec {
       "work with a Sink when having KeyedSource inside" in {
         val probe = TestSubscriber.manualProbe[Int]()
 
-        val source = Source.apply(Source.subscriber[Int]) { implicit b ⇒
+        val source = Source(Source.subscriber[Int]) { implicit b ⇒
```
viktorklang: :)

akka-stream/src/main/boilerplate/akka/stream/scaladsl/SourceApply.scala.template
```diff
@@ -7,6 +7,12 @@ import akka.stream.{ Outlet, Shape, Graph }
 
 trait SourceApply {
 
+  def apply2[Out,Mat](buildBlock: FlowGraph.Builder[Mat] => Outlet[Out]): Source[Out, Mat] = {
```
viktorklang: Why is this needed?

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
```diff
@@ -66,3 +74,163 @@ private[akka] final class OutputStreamSink(createOutput: () ⇒ OutputStream, va
...
+/**
+ * INTERNAL API
+ * Creates simple synchronous (Java 6 compatible) Sink which provide all incoming element on read request via
+ * InputStream.
+ */
+private[akka] class InputStreamSink(val timeout: FiniteDuration,
+                                    val attributes: Attributes, shape: SinkShape[ByteString])
+  extends SinkModule[ByteString, (InputStream, Future[Long])](shape) {
+
+  protected def getSubscriber(sharedBuffer: BlockingQueue[ByteString]) =
```
viktorklang: We always use explicit return type parameters, would you mind adding them? Thank you! :)

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
agolubev: flush and close suppose to be blocked until all data is pulled from downstream or stream is canceled.
Should I rename method? or you have other concern?

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
agolubev: This test case requires some synchronization. We need to know when stage received "canceled" message from downstream. I created test version of SourceState for previous implementation - will figure something out with GraphStage as well. Until the whole approach is not approved would like to keep it pending

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphBackedFlowSpec.scala
agolubev: :) will change this back as change is outside of ticket scope - (until you want me to have it stayed)

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
viktorklang: expectTimeout? (and should perhaps take a timeout parameter)

akka-stream/src/main/boilerplate/akka/stream/scaladsl/SourceApply.scala.template
agolubev: Initially I looked for apply with [Out,Mat] instead of [Out,Unit] but @drewhk gave me a hint that Source.wrap will work here

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphBackedFlowSpec.scala
viktorklang: Why not `Source.subscriber[Int]`?

akka-stream/src/main/boilerplate/akka/stream/scaladsl/SourceApply.scala.template
viktorklang: Ah, yes, good point. Source.wrap will do it!

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
```diff
@@ -0,0 +1,135 @@
...
+
+      outputStream.close()
+      probe.expectComplete()
+    }
+
+    "block writes when buffer is full" in assertAllStagesStopped {
+      val (outputStream, probe) = OutputStreamSource().toMat(TestSink.probe[ByteString])(Keep.both).run
+      val s = probe.expectSubscription()
+
+      //default max-input-buffer-size = 16
```
viktorklang: Instead of this implicit dependency, how about manually specifying it as a setting in the graph?

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
viktorklang: ok

akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
agolubev: Will fix this. Please don't review InputStreamSink/InputStreamSubscriber part as this is still actor based (previous version). Now I'm trying to implement with using GraphStage as @drewhk  suggested

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
agolubev: Yes, it makes sense.

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
agolubev: Ok, will it be like GraphStage.withAttributes fro set and then GraphStage.module.attributes to read?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphBackedFlowSpec.scala
agolubev: Yeah this is example I found and went the wrong way - will simplify it to Source.subscriber[Int]

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
agolubev: Ok, found example       
```scala
val p = SynchronousFileSource(testFile, chunkSize)
        .withAttributes(bufferAttributes)
        .runWith(Sink.publisher)
```

ktoso: yup exactly

agolubev:
@viktorklang, implemented InputStreamSink as GraphStage as well, did rebase.
Now the whole feature is available for code review

agolubev:
Zip spec happy scenario failed for some reason during test run.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -61,6 +61,24 @@ abstract class GraphStage[S <: Shape] extends GraphStageWithMaterializedValue[S,
   def createLogic: GraphStageLogic
 }
 
+/**
+ * A SourceStage represents a reusable graph stream processing stage. A SourceStage consists of a [[akka.stream.Shape]] which describes
+ * its output port.
+ */
+abstract class SourceStage[Out, M](name: String) extends GraphStageWithMaterializedValue[SourceShape[Out], M] {
```
drewhk: I would add convenience methods, `push(elem) ` == `push(out, elem)` and similarly for `pull()`

drewhk:
PLS BUILD

drewhk:
LGTM. This will need a documentation section in the *Integration* section of the reference docs, but that can be another PR (and can be done by us). Can you open a ticket for that?

agolubev:
Ok, will crate additional ticket

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
agolubev: @drewhk, I tried it in few different ways and the most simple was:
```scala
abstract class SourceStage[Out, M](name: String) extends GraphStageWithMaterializedValue[SourceShape[Out], M] {
  def push(elem:Out) = createLogicAndMaterializedValue._1.push(out,elem)
```
which is calling createLogicAndMaterializedValue and creating GraphStageLogic another time. As far as I understand this call must be made only by materializer. Can we skip this for now (I don't see easy way of implementing this)?

drewhk: ah no, that method should be on a subclass of GraphStageLogic, but nevermind we can do this later.

agolubev: ok, will do rebase with squash than

drewhk: Thanks, great work!

agolubev:
@drewhk,  PR is rebased

drewhk:
In it goes.

PR #18710, opened 2015-10-12
Only try further constructors if constructor can't be found.

### LINKED ISSUE #18708: [persistence-query] Read journal constructor exceptions are swallowed
I'm writing a cassandra read journal:
```
class CassandraReadJournal(
    system: ExtendedActorSystem,
    config: Config
) extends ReadJournal with EventsByTagQuery {
```
Since it's not done, its constructor throws an exception (invalid cassandra query).

However, [PersistenceQuery](https://github.com/akka/akka/blob/master/akka-persistence-query/src/main/scala/akka/persistence/query/PersistenceQuery.scala#L82) catches any constructor exception and then tries to find alternative constructors (which don't exist), finally throwing a generic exception. There's no way to even get a try/catch in between the constructor, since most initialization typically happens in `val` assignments.

I'd recommend `PersistenceQuery` chaining the constructor attempts using a `recover { case x:NoSuchMethodException => ...try next... } `, so exceptions on constructors that were found propagate correctly.

jypma:
This way, any exception in a read journal's constructor can be
propagated to the caller correctly. Fixes #18708.

### DIFF
--- a/akka-persistence-query/src/main/scala/akka/persistence/query/PersistenceQuery.scala
+++ b/akka-persistence-query/src/main/scala/akka/persistence/query/PersistenceQuery.scala
@@ -78,9 +78,12 @@ class PersistenceQuery(system: ExtendedActorSystem) extends Extension {
     log.debug(s"Create plugin: ${configPath} ${pluginClassName}")
     val pluginClass = system.dynamicAccess.getClassFor[AnyRef](pluginClassName).get
 
-    system.dynamicAccess.createInstanceFor[ReadJournalProvider](pluginClass, (classOf[ExtendedActorSystem], system) :: (classOf[Config], pluginConfig) :: Nil)
-      .orElse(system.dynamicAccess.createInstanceFor[ReadJournalProvider](pluginClass, (classOf[ExtendedActorSystem], system) :: Nil))
-      .orElse(system.dynamicAccess.createInstanceFor[ReadJournalProvider](pluginClass, Nil))
+    def instantiate(args: collection.immutable.Seq[(Class[_], AnyRef)]) =
+      system.dynamicAccess.createInstanceFor[ReadJournalProvider](pluginClass, args)
+
+    instantiate((classOf[ExtendedActorSystem], system) :: (classOf[Config], pluginConfig) :: Nil)
+      .recoverWith { case x: NoSuchMethodException ⇒ instantiate((classOf[ExtendedActorSystem], system) :: Nil) }
+      .recoverWith { case x: NoSuchMethodException ⇒ instantiate(Nil) }
       .recoverWith {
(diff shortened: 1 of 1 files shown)


typesafehub-validator:
Hi @jypma,

ktoso:
OK TO TEST

ktoso:
We have a similar style of logic in `Persistence` itself, so sounds like it would be best to keep their behaviour consistent: https://github.com/akka/akka/blob/master/akka-persistence/src/main/scala/akka/persistence/Persistence.scala#L262 (i.e. let's change there as well).

jypma:
@ktoso Is there a reason that a persistence plugin can't define requiring an `ExtendedActorSystem` as argument, but a persistence query plugin can? If we're changing it anyways, I'd rather extract a single method `instantiatePlugin[T](system: ExtendedActorSystem, pluginConfig:Config)` that is used in both cases. Or maybe that's what you meant, but it'd be extending on the existing functionality of `Persistence.scala`.

ktoso:
As long as we don't break existing journals we can add this capability to the Persistence plugins :)

jypma:
brrr now I see, `Persistence` plugins are actors, `PersistenceQuery` plugins are singleton instances... is that on purpose?

ktoso:
Nothing in this world is an accident :wink: Yes, because the QueryJournals are actually exposed to users - in order to provide additional type-safe methods to query things.

patriknw:
LGTM
Have you settled on not changing any other things?

jypma:
I'm working on it, I'm trying to refactor the instantiation logic of both out into a shared subclass of `ExtensionId`.

patriknw:
It is completely different for actors, you will not get any exceptions that are thrown by the actor's constructor, because that is executing in a different thread. I don't think we should complicate this. Current change looks good to me.

jypma:
@patriknw good point. Plus, being an actor, the persistence plugin can of course already get to the actor system it's running in. @ktoso , what do you say? BTW my CLA should be in by now.

ktoso:
LGTM from my side as well. Thanks a lot!

ktoso:
Well, one last thing I'd like to ask you to do @jypma is to change the commit message to match our guidelines: https://github.com/akka/akka/blob/master/CONTRIBUTING.md#creating-commits-and-writing-commit-messages In this case it would be `=per #18708 only try further constructors if constructor can't be found. `.

It makes it easier for us to track changes back to discussions etc.

ktoso:
Merged, thanks a lot! :-)

jypma:
uhm did you merge before or after I pushed the commit change?

ktoso:
After, it's all fine :)

jypma:
ok great, thanks!

patriknw:
Thanks for contributing!

Issue #18479, opened 2015-09-15
Uri parsing error


RomanIakovlev:
I see the following exception when creating the `Uri` object:

    akka.http.scaladsl.model.IllegalUriException: Illegal URI reference: Invalid input '=', expected '+', query-char, '%', '&', '#' or 'EOI' (line 1, column 35): /?EVENT=itemGroup&GROUP=10380?page=2
                                  ^
	at akka.http.scaladsl.model.IllegalUriException$.apply(ErrorInfo.scala:40) ~[com.typesafe.akka.akka-http-core-experimental_2.11-1.0.jar:na]
	at akka.http.scaladsl.model.Uri$.fail(Uri.scala:748) ~[com.typesafe.akka.akka-http-core-experimental_2.11-1.0.jar:na]
	at akka.http.impl.model.parser.UriParser.fail(UriParser.scala:61) ~[com.typesafe.akka.akka-http-core-experimental_2.11-1.0.jar:na]
	at akka.http.impl.model.parser.UriParser.parseUriReference(UriParser.scala:33) ~[com.typesafe.akka.akka-http-core-experimental_2.11-1.0.jar:na]
	at akka.http.scaladsl.model.Uri$.apply(Uri.scala:204) ~[com.typesafe.akka.akka-http-core-experimental_2.11-1.0.jar:na]
	at akka.http.scaladsl.model.Uri$.apply(Uri.scala:176) ~[com.typesafe.akka.akka-http-core-experimental_2.11-1.0.jar:na]

The stack trace is provided from the first entry of Akka library.

I'm not responsible for producing the uri, `/?EVENT=itemGroup&GROUP=10380?page=2`, but it looks valid to me, if strange.

Akka version is 2.4-RC1. I tried to search for similar issues, but haven't found any.

2beaucoup:
Is the second "?" supposed to be a "&"? The parser groups the query params separated by "&" and sees the second param as `GROUP=10380?page=2` which is invalid as it has two "="s.

RomanIakovlev:
I guess it is, but, again it's not my url, I'w writing a web crawler in Akka and that was one of the failed cases I've found. Nevertheless, this URL is valid, because subsequent entries of `?` are supposed to be parsed as data within the query component, as per http://tools.ietf.org/html/rfc3986#section-3.4.

I see 2 options here. First one is to treat `?` as `&`, i.e. add new parameter. Second one is to parse the value of `GROUP` parameter as `10380?page=2`.

[...]

PR #18715, opened 2015-10-13
!htp #18479 defer parsing of query key-value-pairs

aruediger:
Contains the following changes:
- URIs now contain the raw query string with an added `query()` method that parses on demand
- removed `Uri.ParsingMode.RelaxedWithRawQuery`
- add method to add query params to scala to mirror java and switched to `japi.Pair` for the params
- introduce `akka.http.javadsl.model.Query`
- removes  `withQueryParam` (scala) and `addParameter` (java) to prevent re-parsing during query construction

/cc @sirthias @jrudolph

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/FormData.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/FormData.java
@@ -4,34 +4,51 @@
 
 package akka.http.javadsl.model;
 
-import akka.http.impl.model.parser.CharacterClasses;
-import akka.http.impl.util.JavaMapping;
-import akka.http.impl.util.StringRendering;
-import akka.http.scaladsl.model.Uri.Query;
-import akka.http.scaladsl.model.UriRendering;
 import akka.japi.Pair;
+import java.util.Map;
 
 /**
  * Simple model for `application/x-www-form-urlencoded` form data.
  */
-public abstract class FormData {
+public final class FormData {
 
-  public abstract Query fields();
+  private final Query fields;
 
+  public FormData(Query fields) {
+    this.fields = fields;
+  }
+
+  /**
+   * Converts this FormData to a RequestEntity using UTF8 encoding.
+   */
   public RequestEntity toEntity() {
     return toEntity(HttpCharsets.UTF_8);
   }
 
+  /**
+   * Converts this FormData to a RequestEntity using the given encoding.
+   */
   public RequestEntity toEntity(HttpCharset charset) {
-    // TODO this logic is duplicated in scaladsl.model.FormData, spent hours trying to DRY it but compiler freaked out in a number of ways... -- ktoso
-    final akka.http.scaladsl.model.HttpCharset c = (akka.http.scaladsl.model.HttpCharset) charset;
-    final StringRendering render = (StringRendering) UriRendering.renderQuery(new StringRendering(), this.fields(), c.nioCharset(), CharacterClasses.unreserved());
-    return HttpEntities.create(ContentType.create(MediaTypes.APPLICATION_X_WWW_FORM_URLENCODED, charset), render.get());
+    return HttpEntities.create(ContentType.create(MediaTypes.APPLICATION_X_WWW_FORM_URLENCODED, charset), fields.render(charset));
   }
 
+  /**
+   * Returns empty FormData.
+   */
+  public static final FormData EMPTY = new FormData(Query.EMPTY);
+
+  /**
+   * Creates the FormData from the given parameters.
+   */
   @SafeVarargs
-  public static FormData create(Pair<String, String>... fields) {
-    return akka.http.scaladsl.model.FormData.create(fields);
+  public static FormData create(Pair<String, String>... params) {
+    return new FormData(Query.create(params));
   }
 
+  /**
+   * Creates the FormData from the given parameters.
+   */
+  public static FormData create(Map<String, String> params) {
+    return new FormData(Query.create(params));
+  }
 }
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/Query.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/Query.java
@@ -1,0 +1,108 @@
+/**
+ * Copyright (C) 2009-2014 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.javadsl.model;
+
+import akka.http.impl.model.JavaQuery;
+import akka.http.scaladsl.model.*;
+import akka.japi.Option;
+import akka.japi.Pair;
+import akka.parboiled2.CharPredicate;
+import akka.parboiled2.ParserInput$;
+
+import java.nio.charset.Charset;
+import java.util.List;
+import java.util.Map;
+
+public abstract class Query {
+    /**
+     * Returns the value of the first parameter with the given key if it exists.
+     */
+    public abstract Option<String> get(String key);
+
+    /**
+     * Returns the value of the first parameter with the given key or the provided default value.
+     */
+    public abstract String getOrElse(String key, String _default);
+
+    /**
+     * Returns the value of all parameters with the given key.
+     */
+    public abstract List<String> getAll(String key);
+
+    /**
+     * Returns a `List` of all parameters of this Query. Use the `toMap()`
+     * method to filter out entries with duplicated keys.
+     */
+    public abstract List<Pair<String, String>> toList();
+
+    /**
+    * Returns a key/value map of the parameters of this Query. Use
+    * the `toList()` method to return all parameters if keys may occur
+    * multiple times.
+    */
+    public abstract Map<String, String> toMap();
+
+    /**
+     * Returns a `Map` of all parameters of this Query. Use the `toMap()`
+     * method to filter out entries with duplicated keys.
+     */
+    public abstract Map<String, List<String>> toMultiMap();
+
+    /**
+     * Returns a copy of this instance with a query parameter added.
+     */
+    public abstract Query withParam(String key, String value);
+
+    /**
+     * Renders this Query into its string representation using the given charset.
+     */
+    public abstract String render(HttpCharset charset);
+
+    /**
+     * Renders this Query into its string representation using the given charset and char predicate.
+     */
+    public abstract String render(HttpCharset charset, CharPredicate keep);
+
+    /**
+     * Returns an empty Query.
+     */
+    public static final Query EMPTY = new JavaQuery(UriJavaAccessor.emptyQuery());
+
+    /**
+     * Returns a Query created by parsing the given undecoded string representation.
+     */
+    public static Query create(String rawQuery) {
+        return new JavaQuery(akka.http.scaladsl.model.Uri.Query$.MODULE$.apply(rawQuery));
+    }
+
+    /**
+     * Returns a Query created by parsing the given undecoded string representation with the provided parsing mode.
+     */
+    public static Query create(String rawQuery, akka.http.scaladsl.model.Uri.ParsingMode parsingMode) {
+        return new JavaQuery(UriJavaAccessor.queryApply(rawQuery, parsingMode));
+    }
+
+    /**
+     * Returns a Query created by parsing the given undecoded string representation with the provided charset and parsing mode.
+     */
+    public static Query create(String rawQuery, Charset charset, akka.http.scaladsl.model.Uri.ParsingMode parsingMode) {
+        return new JavaQuery(akka.http.scaladsl.model.Uri.Query$.MODULE$.apply(ParserInput$.MODULE$.apply(rawQuery), charset, parsingMode));
+    }
+
+    /**
+     * Returns a Query from the given parameters.
+     */
+    @SafeVarargs
+    public static Query create(Pair<String, String>... params) {
+        return new JavaQuery(UriJavaAccessor.queryApply(params));
+    }
+
+    /**
+     * Returns a Query from the given parameters.
+     */
+    public static Query create(Map<String, String> params) {
+        return new JavaQuery(UriJavaAccessor.queryApply(params));
+    }
+}
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/Uri.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/Uri.java
@@ -4,12 +4,14 @@
 
 package akka.http.javadsl.model;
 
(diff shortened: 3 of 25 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
```diff
@@ -30,6 +30,14 @@ sealed abstract case class Uri(scheme: String, authority: Authority, path: Path,
   def isEmpty: Boolean
 
   /**
+   * Parses the given String into a Query instance.
+   */
+  def query(charset: Charset = UTF8, mode: Uri.ParsingMode = Uri.ParsingMode.Relaxed): Query = queryString match {
```
2beaucoup: I could make `charset` and `mode` members of the URI so this method gets parameter-less.

akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
```diff
@@ -95,22 +103,27 @@ sealed abstract case class Uri(scheme: String, authority: Authority, path: Path,
...
   /**
    * Returns a copy of this Uri with a Query created using the given map.
    */
-  def withQuery(map: Map[String, String]): Uri = copy(query = Query(map))
+  def withQuery(map: Map[String, String]): Uri = withQuery(Query(map))
+
+  /**
+   * Returns a copy of this Uri with a Query containing the provided key-value-pair.
+   */
+  def withQueryParam(key: String, value: String): Uri = withQuery(((key -> value) +: query().reverse).reverse)
```
2beaucoup: Can we get rid of this?

akka-http-core/src/main/java/akka/http/javadsl/model/Uri.java
```diff
@@ -80,7 +81,7 @@
      * Returns an `Iterable` of all query parameters of this Uri. Use the `parameterMap()`
      * method to filter out entries with duplicated keys.
      */
-    public abstract Iterable<Map.Entry<String, String>> parameters();
+    public abstract Iterable<Pair<String, String>> parameters();
```
2beaucoup: Can we get rid of `addParameter()` below?

ktoso:
OK TO TEST

akka-http-core/src/main/scala/akka/http/scaladsl/model/Uri.scala
jrudolph: I would suggest another approach: with deferred query parsing we never loose any information any more between the request parser and later stages. Therefore, couldn't we have a standard `lazy val query: Query` which uses parsing mode strict and UTF8 charset by default?

We could then even get completely rid of the parsing mode and just introduce another method which allows to reparse the query string with any other (like `Relaxed`) rules.

2beaucoup: Ok, we can omit the charset. So we would have
```scala
private def query(mode: Uri.ParsingMode): Query = queryString match {
  case Some(q) ⇒ new UriParser(q, uriParsingMode = mode).parseQuery()
  case None    ⇒ Query.Empty
}
lazy val query = query(Uri.ParsingMode.Strict)
```
vs.
```scala
private def query(mode: Uri.ParsingMode = Uri.ParsingMode.Relaxed): Query = queryString match {
  case Some(q) ⇒ new UriParser(q,  uriParsingMode = mode).parseQuery()
  case None    ⇒ Query.Empty
}
```
Does the lazy val really pull its weight here?

2beaucoup: It could be misleading that the query would always get parsed in strict mode by default even though the containing URI could have been parsed in relaxed mode.

jrudolph: Yes, good point, I forgot that relaxed parsing applies to path components, as well. I wouldn't get rid of the charset parameter, I'd just like to have the common case to be not more complicated than before.

jrudolph: Btw. is there a difference between `queryString = Some("")` and `queryString = None`? Otherwise, you could get rid of the option wrapper. (I guess it's whether there's a `'?'` sign or not. We need to decide if we want to keep this distinction.)

2beaucoup: It's the same for `fragment.` If there's a `#` then the result is `Some("")`.

2beaucoup: I'd repeat my top comment (for the default, lazy val case) + add the possibility to provide custom `charset` and  `mode`.

jrudolph: We cannot make `charset` a field because that would prevent serialization, we would need to keep its name.

sirthias: IMHO neither the charset nor the mode is a property of the URI.
Rather these are parameters to the (de)serialization process.
As such I think they should stay as parameters to the parse/render methods.

2beaucoup: We have
```scala
def apply(input: ParserInput, charset: Charset, mode: Uri.ParsingMode)
def query(charset: Charset = UTF8, mode: Uri.ParsingMode = Uri.ParsingMode.Relaxed)
```

for which we want to provide shortcuts with defaults for `charset` and `mode`. We could

drewhk:
What is the fate of this PR?

2beaucoup:
OK TO TEST

2beaucoup:
I've adapted the tests to the new behaviour (and they succeed locally).

The following questions remain:
- Should we add a lazy val for the query? If yes, what charset/mode should be used?
- Should we remove the param modifiers `withQueryParam` (scala) and `addParameter` (java) as they need to reparse the query string?
- The current version leaves the query string undecoded and decodes it while parsing into the `Query` instance. This stands in contrast to the other URI parts. Should we decode eagerly and parse into key-val-pairs lazily? (I'd say yes. See https://github.com/2beaucoup/akka/commit/63d036c778501e2d3fcbfc6ede1122bbda9d88c2#diff-dcccad7d939d98961ea90818f2c4974aR273) Otherwise `queryString` should be renamed to e.g.`rawQueryString`.

WDYT?

/cc @sirthias @jrudolph

2beaucoup:
PLS BUILD

sirthias:
My opinion here on the open questions:

> Should we add a lazy val for the query? If yes, what charset/mode should be used?

I'd say no, a lazy val probably wouldn't add a lot of value. I don't think many application really do access the query several times. And for the few that do the overhead should be negligible.

> Should we remove the param modifiers withQueryParam (scala) and addParameter (java) as they need to reparse the query string?

Yes, I think these methods are not crucial and, as you say, somewhat don't fit into the new model.

> The current version leaves the query string undecoded and decodes it while parsing into the Query instance. This stands in contrast to the other URI parts. Should we decode eagerly and parse into key-val-pairs lazily? (I'd say yes....) 

Yes, I strongly agree here! We need to be consistent as to when de-/encoding happens.
The property of the Uri model to only ever contain *decoded* stuff is imporant to uphold, IMHO.

Thanks, @2beaucoup, for pushing this forward.
Important stuff!

2beaucoup:
Thx @sirthias!

>> The current version leaves the query string undecoded and decodes it while parsing into the Query instance. This stands in contrast to the other URI parts. Should we decode eagerly and parse into key-val-pairs lazily? (I'd say yes....)

> Yes, I strongly agree here! We need to be consistent as to when de-/encoding happens.
> The property of the Uri model to only ever contain decoded stuff is imporant to uphold, IMHO.

Yeah, I played with the idea. The problem here is that FormData decoding introduces new rules like + -> ' ' which requires decoding in one run to prevent double decoding (&2b -> + -> ' '). Therefore I renamed `queryString` to `rawQueryString` in the last commit.

I also introduced a Java model for queries make step-by-step construction without constant reparsing possible. I'll sprinkle some comments if we agree that that makes sense.

sirthias:
> The problem here is that FormData decoding introduces new rules like + -> ' ' which requires decoding in one run to prevent double decoding (&2b -> + -> ' ').

Hmm... ok, I see.
Where is this distinction between the different query parsing modes currently done?
To me it looks like the UriParser *always* does the `+` to ' ' translation for query strings, no matter whether they are then parsed into FormData or not.

2beaucoup:
It's done [here](https://github.com/akka/akka/pull/18715/files#diff-52e496fada3f67a2626c662c33c0d392R169) when parsing the (undecoded) query string into key value pairs (a.k.a. FormData).

The [`query()` method](https://github.com/akka/akka/pull/18715/files#diff-e0948a9dfaf145ac83bbbefe9080d77dR35) has params for charset and mode.

ktoso:
https://github.com/akka/akka/pull/18715/files#diff-e0948a9dfaf145ac83bbbefe9080d77dR35 looks great! Thanks @2beaucoup

sirthias:
@2beaucoup Yes, sorry, my question was unclear.
What I was trying to do is fully understanding your point that we cannot decode the query eagerly because FormData unmarshalling requires access to the raw query.
But I get it now.

In order to provide the same feature that we had before (enabling access to the raw query, before decoding) we need the queryString be a `rawQueryString`. Otherwise we'd essentially remove the enable-full-custom-parsing-of-weird-query-strings-feature altogether.

Even though I currently can't remember, why we introduced it in the first place I am pretty sure that removing it would just cause someone to step up and request it being added back sooner or later.

So, your move from `queryString` to `rawQueryString` is probably the only solution, even though it is kind of ugly that we now have decoded as well as undecoded stuff in an `Uri` instance at the same.
(But then we already did so before, just better hidden inside of the Query model.)

So, +1 from my side for this patch as it stands now.

2beaucoup:
@sirthias The only motivation for keeping the raw query string was correctness of form data parsing. I guess to an API user the decoded string is of more value than it's undecoded form. Therefore I suggest to add
```scala
def queryString(charset: Charset = UTF8) = decode(rawQueryString, charset, 0)
```
WDYT?

sirthias:
@2beaucoup Yes, convenient access to the decoded queryString is certainly something we should provide!

akka-http-core/src/main/java/akka/http/javadsl/model/FormData.java
```diff
@@ -4,34 +4,51 @@
...
 
-  public abstract Query fields();
+  private final Query fields;
 
+  public FormData(Query fields) {
+    this.fields = fields;
+  }
+
+  /**
+   * Convertes this FormnData to a RequestEntity using UTF8 encoding.
```
ktoso: FormnData => FormData

2beaucoup: Convertes => Converts ;)

akka-http-core/src/main/java/akka/http/javadsl/model/Query.java
```diff
@@ -0,0 +1,79 @@
...
+import akka.japi.Option;
+import akka.japi.Pair;
+import akka.parboiled2.CharPredicate;
+import akka.parboiled2.ParserInput$;
+
+import java.nio.charset.Charset;
+import java.util.List;
+import java.util.Map;
+
+public abstract class Query {
```
ktoso: Let's document this class (and the scala one as well), a little bit of JavaDoc explaining that it's part of the Uri?

2beaucoup:
@sirthias While adding this convinience I wrapped my head around the charset stuff. I haven't found a single place where it's not UTF-8. Same for `Path.apply` (except [here](https://github.com/2beaucoup/akka/blob/deferred-query-parsing/akka-http-core/src/test/scala/akka/http/scaladsl/model/UriSpec.scala#L225)).

Can we get rid of it? Why is the charset even involved in decoding where the input is already a java string?

ktoso:
LGTM from my side, looks good after final touchups!
I'll refer to @sirthias about the encoding though; I don't think it hurts much to have Utf8 as default like we have nowadays.

sirthias:
@2beaucoup We need the charset for being able to interpret percent-encoded byte sequences. I guess you are right in saying that the very large majority of users will be happy with UTF-8. But I am *sure* that someone will come and complain, if we remove the parameter altogether.
Currently the cost of the parameter to the user is close to zero and one might even argue that it's a kind of documentation, showing that some kind of decoding might be happening inside of the method.
So I'd vote to keep the parameter.

ktoso:
So... anything else left to discuss here or ready to merge?
We'll be releasing a milestone tomorrow :)

2beaucoup:
I think it's OK for now.

I've pushed a squashed version which addresses your comments.

ktoso:
Great, thanks @2beaucoup!

ktoso:
refs #18479

ktoso:
Whoops, error was compilation error // cc @2beaucoup 

```
[error] /localhome/jenkinsakka/workspace/pr-validator-per-commit-jenkins/akka-http-core/src/main/scala/akka/http/impl/model/JavaQuery.scala:26: overriding method toMultiMap in class Query of type ()java.util.Map[String,Iterable[String]];
[error]  method toMultiMap has incompatible type
[error]   override def toMultiMap: ju.Map[String, ju.List[String]] = query.toMultiMap.map { case (k, v) ⇒ (k, v.asJava) }.asJava
[error]                ^
```

If fixed on time this could go into `2.0-M1` :)

2beaucoup:
Should be good to go now.

akka-http-core/src/main/java/akka/http/javadsl/model/Query.java
2beaucoup: I just documented here as `JavaQuery` is INTERNAL and the comments would be redundant.

akka-http-core/src/main/resources/reference.conf
```diff
@@ -277,10 +277,6 @@ akka.http {
     #
     # `relaxed`: all visible 7-Bit ASCII chars are allowed
     #
-    # `relaxed-with-raw-query`: like `relaxed` but additionally
-    #     the URI query is not parsed, but delivered as one raw string
-    #     as the `key` value of a single Query structure element.
-    #
```
2beaucoup: Removal of this mode was the main motivation of this PR.

ktoso:
LGTM, thanks a lot @2beaucoup !

PR #18720, opened 2015-10-13
=doc Fix sentence in `Actors/Send messages/Ask`

markus1189:
The sentence

> Gives you a way to avoid blocking.

does not make sense alone, I added it to the previous one.

The relevant part was introduced in ce128740abf8272d795d528ce58e043747e7636d, maybe @patriknw can have a look and see whether that matches his intended meaning?


ktoso:
LGTM, thanks!

PR #18723, opened 2015-10-14
htc: check for client-side response stream truncation

### LINKED ISSUE #17344: Error-close client-side response stream if there are responses pending at connection end
Currently the low-level client-side connection stream simply completes the response stream if the underlying HTTP connection is cleanly closed, even if there are more (potentially pipelined) responses pending.

This ticket proposes to change this behavior to signal the premature closing of a connection, with (potentially pipelined) responses still pending, as a "truncation error" rather than a clean stream completion.

Supersedes #16865.

/cc @jrudolph @richardbradley

sirthias:
Closes #17344.

There is one question open:
How do we best make the `one2one` wrapper method that this patch currently only defines in the `One2OneBidiFlowSpec` available to the public?
It probably shouldn't go into the core `FlowOps`. What's the best alternative?

/cc @jrudolph

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/One2OneBidiFlowSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/One2OneBidiFlowSpec.scala
@@ -1,0 +1,87 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.scaladsl
+
+import java.util.concurrent.atomic.AtomicInteger
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import org.scalactic.ConversionCheckedTripleEquals
+import akka.stream.ActorMaterializer
+import akka.stream.testkit.AkkaSpec
+import akka.stream.testkit._
+
+class One2OneBidiFlowSpec extends AkkaSpec with ConversionCheckedTripleEquals {
+  implicit val mat = ActorMaterializer()
+
+  "A One2OneBidiFlow" must {
+
+    def test(flow: Flow[Int, Int, Unit]) =
+      Source(List(1, 2, 3)).via(flow).grouped(10).runWith(Sink.head)
+
+    "be fully transparent for valid one-to-one streams" in {
+      val f = One2OneBidiFlow[Int, Int](-1) join Flow[Int].map(_ * 2)
+      Await.result(test(f), 1.second) should ===(Seq(2, 4, 6))
+    }
+
+    "be fully transparent to errors" in {
+      val f = One2OneBidiFlow[Int, Int](-1) join Flow[Int].map(x ⇒ 10 / (x - 2))
+      an[ArithmeticException] should be thrownBy Await.result(test(f), 1.second)
+    }
+
+    "trigger an `OutputTruncationException` if the wrapped stream terminates early" in {
+      val f = One2OneBidiFlow[Int, Int](-1) join Flow[Int].filter(_ < 3)
+      a[One2OneBidiFlow.OutputTruncationException.type] should be thrownBy Await.result(test(f), 1.second)
+    }
+
+    "trigger an `UnexpectedOutputException` if the wrapped stream produces out-of-order elements" in new Test() {
+      inIn.sendNext(1)
+      inOut.requestNext() should ===(1)
+
+      outIn.sendNext(2)
+      outOut.requestNext() should ===(2)
+
+      outOut.request(1)
+      outIn.sendNext(3)
+      outOut.expectError(new One2OneBidiFlow.UnexpectedOutputException(3))
+    }
+
+    "drop surplus output elements" in new Test() {
+      inIn.sendNext(1)
+      inOut.requestNext() should ===(1)
+
+      outIn.sendNext(2)
+      outOut.requestNext() should ===(2)
+
+      outOut.cancel()
+      outIn.expectCancellation()
+    }
+
+    "backpressure the input side if the maximum number of pending output elements has been reached" in {
+      val MAX_PENDING = 24
+
+      val out = TestPublisher.probe[Int]()
+      val seen = new AtomicInteger
+
+      Source(1 to 1000)
+        .log("", seen.set)
+        .via(One2OneBidiFlow[Int, Int](MAX_PENDING) join Flow.wrap(Sink.ignore, Source(out))(Keep.left))
+        .runWith(Sink.ignore)
+
+      Thread.sleep(50)
+      val x = seen.get()
+      (1 to 8) foreach out.sendNext
+      Thread.sleep(50)
+      seen.get should ===(x + 8)
+    }
+  }
+
+  class Test(maxPending: Int = -1) {
+    val inIn = TestPublisher.probe[Int]()
+    val inOut = TestSubscriber.probe[Int]()
+    val outIn = TestPublisher.probe[Int]()
+    val outOut = TestSubscriber.probe[Int]()
+
+    Source(inIn).via(One2OneBidiFlow[Int, Int](maxPending) join Flow.wrap(Sink(inOut), Source(outIn))(Keep.left)).runWith(Sink(outOut))
+  }
+}
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/One2OneBidiFlow.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/One2OneBidiFlow.scala
@@ -1,0 +1,87 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.scaladsl
+
+import akka.stream._
+import akka.stream.stage.{ OutHandler, InHandler, GraphStageLogic, GraphStage }
+
+import scala.concurrent.duration.Deadline
+import scala.util.control.NoStackTrace
+
+object One2OneBidiFlow {
+
+  case class UnexpectedOutputException(element: Any) extends RuntimeException with NoStackTrace
+  case object OutputTruncationException extends RuntimeException with NoStackTrace
+
+  /**
+   * Creates a generic ``BidiFlow`` which verifies that another flow produces exactly one output element per
+   * input element, at the right time. Specifically it
+   *
+   * 1. triggers an ``UnexpectedOutputException`` if the inner flow produces an output element before having
+   *    consumed the respective input element.
+   * 2. triggers an `OutputTruncationException` if the inner flow completes before having produced an output element
+   *    for every input element.
+   * 3. Backpressures the input side if the maximum number of pending output elements has been reached,
+   *    which is given via the ``maxPending`` parameter. You can use -1 to disable this feature.
+   * 4. Drops surplus output elements, i.e. ones that the inner flow tries to produce after the input stream
+   *    has signalled completion. Note that no error is triggered in this case!
+   */
+  def apply[I, O](maxPending: Int): BidiFlow[I, I, O, O, Unit] =
+    BidiFlow.wrap(new One2OneBidi[I, O](maxPending))
+
+  class One2OneBidi[I, O](maxPending: Int) extends GraphStage[BidiShape[I, I, O, O]] {
+    val inIn = Inlet[I]("inIn")
+    val inOut = Outlet[I]("inOut")
+    val outIn = Inlet[O]("outIn")
+    val outOut = Outlet[O]("outOut")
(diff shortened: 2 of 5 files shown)


akka-stream-tests/src/test/scala/akka/stream/scaladsl/One2OneBidiFlowSpec.scala
```diff
@@ -0,0 +1,84 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.scaladsl
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import org.scalactic.ConversionCheckedTripleEquals
```
ktoso: oh, we have scalactic in compile scope?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/One2OneBidiFlowSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+   * Wraps a flow with one-to-one guarding, which
+   *
+   * 1. triggers an `OutputTruncationException` if the wrapped stream terminates
+   *    before exactly one output has been produced for each input element
+   * 2. Backpressures surplus output elements (i.e. ones that are being produced without
+   *    having a corresponding input element)
+   * 3. makes sure that output elements are being produced in lockstep with input elements,
+   *    i.e. no output element is ever being produced before the corresponding input element hasn't been received
+   *
+   * // TODO: make this available on Flows; Where do we place it best?
```
ktoso: Hm, that's a weird one... We don't have similar things to it yet, maybe some new place hm.

// cc @drewhk @rkuhn

akka-stream-tests/src/test/scala/akka/stream/scaladsl/One2OneBidiFlowSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+  import FlowGraph.Implicits._
+
+  implicit val mat = ActorMaterializer()
+
+  /**
+   * Wraps a flow with one-to-one guarding, which
+   *
+   * 1. triggers an `OutputTruncationException` if the wrapped stream terminates
+   *    before exactly one output has been produced for each input element
+   * 2. Backpressures surplus output elements (i.e. ones that are being produced without
```
jrudolph: Doesn't that contradict 3.? Or otherwise, if it is allowed to produce several output elements per input element, how do you associate the surplus with incoming elements?

sirthias: The inner (wrapped) flow can try to produce several output elements for one input element.
However, the Bidi stage will only ever let one single output element for each input element reach the outside.

akka-stream/src/main/scala/akka/stream/scaladsl/One2OneBidiFlow.scala
```diff
@@ -0,0 +1,91 @@
...
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.scaladsl
+
+import akka.stream._
+
+import scala.util.control.NoStackTrace
+
+/**
+ * A generic BidiFlow which verifies that another flow produces exactly one output element per
```
jrudolph: Isn't the current implemented logic "at least one output element per input element" instead of "exactly one output element"? What's the reason not to implement the "exactly one output element" rule more closely at least as far as it is dynamically possible. As far as possible because we can only associate input elements to output elements after the input and the wrapped flow are completed. So, the best dynamic approximation would be that we keep an `pendingOutput` counter that never goes below 0.

So, for a supposed identity flow an allowed sequence would be

```
input: 1
output: 1
```

but not

```
output: 1 <- fails here, because we haven't seen any input
input: 1
```

and

```
input: 1
input: 2
output: 1
input: 3
output: 2
output: 3
```

or with bugs (which this flow should prevent but is not able to in this case):

```
input: 1
input: 2
output: 1
output: 1 // oops, bug in the wrapped flow: outputs two elements for the first input, we cannot detect this
```

in which case it may fail later at

```
input: 1
input: 2
output: 1
output: 1 // oops, bug in the wrapped flow: outputs two elements for the first input
output: 2 // <- fail here
```

or

```
input: 1
input: 2
output: 1
output: 1 // oops, bug in the wrapped flow: outputs two elements for the first input
input: 3
output: 2 // oops, we mis-associated another wrong input element with the output element
output: 3 <- but at least we can fail here
```

sirthias: Unfortunately I don't think we can reliably distinguish

```
output: 1
input: 1
```

from

```
input: 1
output: 1
```

at least not with the current bypass-based design.
We could do it with an `AtomicInteger` or something similar, but then we'd have to solve the several-materializations problem.

Do you have a good alternative design in mind?

jrudolph: I guess the implementation problem is that the bypass operates asynchronously, so you don't know in which order you will receive input and output elements. I wonder if there's an ever simpler implementation which puts a simple side-effect stage before the wrapped flow that increments an AtomicInteger, and a similar stage after the wrapped flow that decrements the same counter and then
 * throws if the counter falls below 0
 * asserts that the counter is zero at completion

Shouldn't this work as well? Not sure, how much you need to have to work around materialization to share the counter between the two stages (in the worst case the component becomes non-reusable).

jrudolph: Ok, same idea :)

jrudolph: If we don't find a solution, I agree that this is a solution that at least solves #17344 nicely.

jrudolph:
LGTM for the cause it is supposed to solve.

If this should be made more generally available from akka-stream users it may make sense to create a follow-up ticket to figure out 1) how to access it more easily and 2) how to enforce the additional constraint. WDYT, @ktoso ?

akka-stream/src/main/scala/akka/stream/scaladsl/One2OneBidiFlow.scala
sirthias: Yes, that was pretty much my solution evolution here.
Initially I started out with an AtomicInteger but quickly discovered that materializations become a problem.
Once again we'd need access to a value in one stage from another stage.
Maybe the new octopus stage would allow us to implement the complete bidi stage in a single entity? (Which could then create and use an AtomicInteger internally?)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/One2OneBidiFlowSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+   *    having a corresponding input element)
+   * 3. makes sure that output elements are being produced in lockstep with input elements,
```
drewhk: how can this make sure this? Maybe it enforces it, that is what you mean (as in failing if violated)?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/One2OneBidiFlowSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+   *    i.e. no output element is ever being produced before the corresponding input element hasn't been received
+   *
+   * // TODO: make this available on Flows; Where do we place it best?
+   */
+  def one2one[I, O, Mat](flow: Flow[I, O, Mat]): Flow[I, O, Mat] =
+    Flow(flow) { implicit b ⇒
```
drewhk: this is `flow.join(One2OneBidiFlow)` AFAIK

akka-stream-tests/src/test/scala/akka/stream/scaladsl/One2OneBidiFlowSpec.scala
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    "trigger an `OutputTruncationException` if the wrapped stream terminates early" in {
+      val f = one2one(Flow[Int].filter(_ < 3))
+      a[One2OneBidiFlow.OutputTruncationException.type] should be thrownBy Await.result(test(f), 1.second)
+    }
+
+    "backpressure surplus output elements" in {
+      val f = one2one(Flow[Int].mapConcat(x ⇒ x :: x :: Nil))
+      Await.result(test(f), 1.second) should ===(Seq(1, 1, 2))
```
drewhk: this actually drops, right?

akka-stream/src/main/scala/akka/stream/scaladsl/One2OneBidiFlow.scala
```diff
@@ -0,0 +1,91 @@
...
+  def apply[I, O]: BidiFlow[I, I, O, O, Unit] =
+    BidiFlow() { implicit b ⇒
+      import FlowGraph.Implicits._
+      val bypassFanout = b.add(Broadcast[I](2, eagerCancel = true))
+      val bypassMerge = b.add(new BypassMerge[O])
+      bypassFanout.out(1) ~> Flow[I].map(_ ⇒ ()) ~> bypassMerge.in0
+      BidiShape[I, I, O, O](bypassFanout.in, bypassFanout.out(0), bypassMerge.in1, bypassMerge.out)
+    }
+
+  class BypassMerge[T] extends FlexiMerge[T, FanInShape2[Unit, T, T]](new FanInShape2("BypassMerge"),
```
drewhk: this should be a GraphStage now. I think it will be easier, too.

akka-stream/src/main/scala/akka/stream/scaladsl/One2OneBidiFlow.scala
drewhk: yes, use GraphStage and you won't need the AtomicInteger. It will be very clean I think.

drewhk: Here is a simple IdleTimeout stage as an example (we have no reference docs for GraphStage, but there are some scaladocs already): https://github.com/akka/akka/blob/release-2.3-dev/akka-stream/src/main/scala/akka/stream/io/Timeouts.scala#L123

akka-stream-tests/src/test/scala/akka/stream/scaladsl/One2OneBidiFlowSpec.scala
sirthias: You are right, this comment is somewhat misleading.
What I mean is that for every input element you will only ever see one output element and this element will always be produced *after* the respective input element has been received.
I.e. it's impossible to receive output X *before* input X has been sent.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/One2OneBidiFlowSpec.scala
sirthias: Ah, yes.
Nice!

akka-stream-tests/src/test/scala/akka/stream/scaladsl/One2OneBidiFlowSpec.scala
sirthias: It backpressures and then truncates, which amounts to the surplus elements being dropped, yes.

akka-stream/src/main/scala/akka/stream/scaladsl/One2OneBidiFlow.scala
sirthias: Excellent!
Thanks, @drewhk!
I'll rewrite the whole thing based on the `GraphStage`, which should nicely allow us to detect (and prevent) the "output before input" case Johannes described above.

sirthias:
Converted to GraphStage, which is indeed a lot nicer to work with.
Thanks a lot, @drewhk!
Now the `One2OneBidiFlow` can also check for and throw `UnexpectedOutputException`, which was what @jrudolph had in mind, I think.

sirthias:
Rebased and fixed unused parameter spotted by @2beaucoup.

ktoso:
LGTM, thanks!
As for exposing the stage - let's not (at least for now).

drewhk:
LGTM, and +1 for not exposing it for now

sirthias:
Ok, thanks.
It of course exposed now, just not conveniently tied into the flow DSL.

ktoso:
Right, it being `public` is fine, let's think about "nicely expose" once we see more need for it.

PR #18725, opened 2015-10-14
Improve docs for Actors and Fault Tolerance

markus1189:
Fixes some more issues I found while reading the docs:
- missing `by` in the `Actors` docs
- replace `in runtime` with `at runtime` in the `Fault Tolerance` docs


2m:
Thank you!

Issue #18575, opened 2015-09-28
Publish MemberJoined event


hseeberger:
As discussed with @patriknw, add `MemberJoined` to the `ClusterEvent`s.

patriknw:
I agree that it is no point in hiding these things. Joining members are anyway included in the `CurrentClusterState`.

The use case we discussed was to find a way to know when a join attempt is successful.

hseeberger:
Any chance this is going to happen in 2.4.1? If help needed, I could probably come up with a PR.

ktoso:
If you could PR I think we can surely include it :)

hseeberger:
Fair enough. So there are no compatibility concerns?

hseeberger:
Looks like this won't break binary compatibility, i.e. I'll create a PR.

patriknw:
That would be great.

patriknw:
This was fixed in 2.4.1, right? wrong milestone

hseeberger:
Yep.

PR #18729, opened 2015-10-15
Publish MemberJoined

hseeberger:
See #18575.

### DIFF
--- a/akka-cluster/src/main/scala/akka/cluster/ClusterEvent.scala
+++ b/akka-cluster/src/main/scala/akka/cluster/ClusterEvent.scala
@@ -122,6 +122,13 @@ object ClusterEvent {
   }
 
   /**
+   * Member status changed to Joining.
+   */
+  final case class MemberJoined(member: Member) extends MemberEvent {
+    if (member.status != Joining) throw new IllegalArgumentException("Expected Joining status, got: " + member)
+  }
+
+  /**
    * Member status changed to WeaklyUp.
    * A joining member can be moved to `WeaklyUp` if convergence
    * cannot be reached, i.e. there are unreachable nodes.
@@ -136,6 +143,13 @@ object ClusterEvent {
    */
   final case class MemberUp(member: Member) extends MemberEvent {
     if (member.status != Up) throw new IllegalArgumentException("Expected Up status, got: " + member)
+  }
+
+  /**
+   * Member status changed to Leaving.
+   */
+  final case class MemberLeft(member: Member) extends MemberEvent {
+    if (member.status != Leaving) throw new IllegalArgumentException("Expected Leaving status, got: " + member)
   }
 
   /**
@@ -278,8 +292,10 @@ object ClusterEvent {
         case (_, newMember :: oldMember :: Nil) if newMember.status != oldMember.status ⇒ newMember
       }
       val memberEvents = (newMembers ++ changedMembers) collect {
+        case m if m.status == Joining  ⇒ MemberJoined(m)
         case m if m.status == WeaklyUp ⇒ MemberWeaklyUp(m)
         case m if m.status == Up       ⇒ MemberUp(m)
+        case m if m.status == Leaving  ⇒ MemberLeft(m)
         case m if m.status == Exiting  ⇒ MemberExited(m)
         // no events for other transitions
       }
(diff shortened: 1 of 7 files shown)


hseeberger:
I'm not sure if this PR passes the tests or not, because on my machine I get flaky results, sometimes successful passes and sometimes errors which can't be reproduced with `testQuick`.

I was expecting PR validation through a CI server. But it seems that's not in place. Am I missing something?

patriknw:
OK TO TEST

akka-cluster/src/main/scala/akka/cluster/ClusterEvent.scala
```diff
@@ -278,6 +285,7 @@ object ClusterEvent {
         case (_, newMember :: oldMember :: Nil) if newMember.status != oldMember.status ⇒ newMember
       }
       val memberEvents = (newMembers ++ changedMembers) collect {
+        case m if m.status == Joining  ⇒ MemberJoined(m)
         case m if m.status == WeaklyUp ⇒ MemberWeaklyUp(m)
         case m if m.status == Up       ⇒ MemberUp(m)
         case m if m.status == Exiting  ⇒ MemberExited(m)
```
patriknw: shall we add `Leaving` also, to make it complete?

patriknw:
perhaps make use of it in `ClusterReadView` also? (that is used by JMX)

patriknw:
thanks Heiko, I will review in more detail next week

akka-cluster/src/main/scala/akka/cluster/ClusterEvent.scala
hseeberger: Not sure. Is the difference between `Leaving` and `Exiting` important for users? I always thought of that more as an internal detail.

hseeberger:
Ah, "OK TO TEST" are the magic words ;-)

hseeberger:
OK TO TEST

hseeberger:
@patriknw, `NodeChurnSpec` failure is related to `ClusterReadView` which treats all `MemberEvent`s the same way. So, should we change the way `ClusterReadView` works or fix the tests?

akka-cluster/src/main/scala/akka/cluster/ClusterEvent.scala
patriknw: We thought of joining as an internal detail also, I'd rather make it complete now and expose events for all state transitions. As a bonus we can make use of them in the ReadView so that JMX shows correct status.

patriknw:
@hseeberger please update `ClusterReadView` to make use of the new events.

akka-cluster/src/main/scala/akka/cluster/ClusterEvent.scala
hseeberger: OK, adding `MemberLeft`.

hseeberger:
@patriknw, I think `ClusterReadView` already makes use of the two new events, because it handles all `MemberEvent`s in a general way. What else do you have in mind?

patriknw:
I see, good. Then we need to investigate why the NodeChurnSpec fails.

patriknw:
@hseeberger I have been debugging the NodeChurnSpec and it was almost driving me crazy. I could not for my life understand how this change would make it fail. Finally I found the issue. This line is missing an assert: https://github.com/akka/akka/blob/master/akka-cluster/src/multi-jvm/scala/akka/cluster/NodeChurnSpec.scala#L70

It should be: 
 
    c.state.members.forall(_.status == MemberStatus.Up) should be(true)

That caused the leave command to be issued early and thereby ignored (the node must be in Up state).

hseeberger:
@patriknw, thanks! I will update my PR. Let's keep fingers crossed it passes then ;-)

hseeberger:
Finally passed PR. Thanks, @patriknw!

patriknw:
Thank you!
LGTM

hseeberger:
All right, now that we have achieved this, would you, @patriknw, please shed some light onto me poor ignoramus? What exactly is going on when a member leaves? What's the difference btw. leaving and exiting?

patriknw:
Look at the [state diagram](http://doc.akka.io/docs/akka/2.4.0/common/cluster.html#State_Diagram_for_the_Member_States__akka_cluster_allow-weakly-up-members=off_)

the transition from Up to Leaving is triggered by a user command, while the transition from Leaving to Exiting is performed by the cluster (the leader when there is convergence). Rather similar as the relation between Joining and Up.

hseeberger:
Ah, now I see. Thanks for the good description.

johanandren:
LGTM

hseeberger:
Will this PR be part of 2.4.1?

patriknw:
absolutely, it is done as far as I can see

hseeberger:
Yep. I was asking, because the PR isn't assigned to a milestone.

ktoso:
We don't assign PRs to milestones, we do that for issues.

ktoso:
I'll review and merge, first more customer work though.

drewhk:
LGTM

drewhk:
Thanks @hseeberger

PR #18732, opened 2015-10-16
=sam Updated invalid doc links

ch33hau:
I'm reading some tutorial pages from akka-samples and noticed that the doc links are pointing to **2.4-SNAPSHOT** which is not working, I guess the correct one should be **snapshot** (correct me if I'm wrong)...

For eg, 
    http://doc.akka.io/docs/akka/2.4-SNAPSHOT/scala/distributed-data.html#Limitations
vs
    http://doc.akka.io/docs/akka/snapshot/scala/distributed-data.html#Limitations

I have did a _grep_ and updated them, appreciate if you could review it

### DIFF
--- a/akka-samples/akka-sample-camel-java/tutorial/index.html
+++ b/akka-samples/akka-sample-camel-java/tutorial/index.html
@@ -8,7 +8,7 @@
 <div>
 <p>
 This tutorial contains 3 samples of 
-<a href="http://doc.akka.io/docs/akka/2.4-SNAPSHOT/java/camel.html" target="_blank">Akka Camel</a>.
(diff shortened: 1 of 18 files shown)


ktoso:
Hi @ch33hau,
thanks for investigating however this is not needed.
The `2.4-SNAPSHOT` is replaced with "the version currently being released" during the release script's run, so it always is the "right version" (as in 2.4.1 for example).

ktoso:
The snippet here fixes the version during the build: https://github.com/akka/akka/blob/master/project/scripts/release#L284

ch33hau:
@ktoso thanks for telling me!

Issue #17383, opened 2015-05-05
add `intersperse` or `interleave` combinator


rkuhn:
As a streamified generic version of the collections’ mkString method:
~~~scala
def interleave(separator: Out): Repr[Out]
def interleave(initial: Out, separator: Out, terminal: Out): Repr[Out]
~~~

adamw:
+1

ktoso:
Implemented https://github.com/akka/akka/issues/18734

PR #18734, opened 2015-10-18
+str #17383 implement intersperse

ktoso:
Resolves https://github.com/akka/akka/issues/17383

Will be used by json streaming rendering.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
@@ -26,6 +26,8 @@ private[akka] case object EmptyPublisher extends Publisher[Nothing] {
  * INTERNAL API
  */
 private[akka] final case class ErrorPublisher(t: Throwable, name: String) extends Publisher[Nothing] {
+  ReactiveStreamsCompliance.requireNonNullElement(t)
+
   import ReactiveStreamsCompliance._
   override def subscribe(subscriber: Subscriber[_ >: Nothing]): Unit =
     try {
@@ -41,6 +43,8 @@ private[akka] final case class ErrorPublisher(t: Throwable, name: String) extend
 
 private[akka] final case class SingleElementPublisher[T](value: T, name: String) extends Publisher[T] {
   import ReactiveStreamsCompliance._
+
+  requireNonNullElement(value)
 
   private[this] class SingleElementSubscription(subscriber: Subscriber[_ >: T]) extends Subscription {
     private[this] var done: Boolean = false
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -239,6 +239,49 @@ private[akka] final case class Fold[In, Out](zero: Out, f: (Out, In) ⇒ Out, de
 /**
  * INTERNAL API
  */
+private[akka] final case class Intersperse[T](start: Option[T], inject: T, end: Option[T]) extends StatefulStage[T, T] {
+  private var needsToEmitStart = start.isDefined
+
+  override def initial: StageState[T, T] =
+    start match {
+      case Some(initial) ⇒ firstWithInitial(initial)
+      case _             ⇒ first
+    }
+
+  def firstWithInitial(initial: T) = new StageState[T, T] {
+    override def onPush(elem: T, ctx: Context[T]) = {
+      needsToEmitStart = false
+      emit(Iterator(initial, elem), ctx, running)
+    }
+  }
+
+  def first = new StageState[T, T] {
+    override def onPush(elem: T, ctx: Context[T]) = {
+      become(running)
+      ctx.push(elem)
+    }
+  }
+
+  def running = new StageState[T, T] {
+    override def onPush(elem: T, ctx: Context[T]): SyncDirective =
+      emit(Iterator(inject, elem), ctx)
+  }
+
+  override def onUpstreamFinish(ctx: Context[T]): TerminationDirective = {
+    end match {
+      case Some(e) if needsToEmitStart ⇒
+        terminationEmit(Iterator(start.get, end.get), ctx)
+      case Some(e) ⇒
+        terminationEmit(Iterator(end.get), ctx)
+      case _ ⇒
+        terminationEmit(Iterator(), ctx)
+    }
+  }
+}
+
+/**
+ * INTERNAL API
+ */
 private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immutable.Seq[T]] {
   private val buf = {
     val b = Vector.newBuilder[T]
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
@@ -5,7 +5,8 @@ package akka.stream.javadsl
 
 import akka.event.LoggingAdapter
 import akka.japi.{ Pair, function }
-import akka.stream.impl.StreamLayout
+import akka.stream.impl.Stages.Intersperse
+import akka.stream.impl.{ ReactiveStreamsCompliance, StreamLayout }
 import akka.stream.{ scaladsl, _ }
 import akka.stream.stage.Stage
 import org.reactivestreams.Processor
@@ -405,6 +406,65 @@ class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Graph
    */
   def fold[T](zero: T)(f: function.Function2[T, Out, T]): javadsl.Flow[In, T, Mat] =
     new Flow(delegate.fold(zero)(f.apply))
+
+  /**
+   * Intersperses stream with provided element, similar to how [[scala.collection.immutable.List.mkString]]
+   * injects a separator between a List's elements.
+   *
+   * Additionally can inject start and end marker elements to stream.
+   *
+   * Examples:
+   *
+   * {{{
+   * Source<Integer, ?> nums = Source.from(Arrays.asList(0, 1, 2, 3));
+   * nums.intersperse(",");            //   1 , 2 , 3
+   * nums.intersperse("[", ",", "]");  // [ 1 , 2 , 3 ]
+   * }}}
+   *
+   * In case you want to only prepend or only append an element (yet still use the `intercept` feature
+   * to inject a separator between elements, you may want to use the following pattern instead of the 3-argument
+   * version of intersperse (See [[Source.concat]] for semantics details):
+   *
+   * {{{
+   * Source.single(">> ").concat(flow.intersperse(","))
+   * flow.intersperse(",").concat(Source.single("END"))
+   * }}}
+   *
+   * '''Emits when''' upstream emits (or before with the `start` element if provided)
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def intersperse[T >: Out](start: T, inject: T, end: T): javadsl.Flow[In, T, Mat] =
+    new Flow(delegate.intersperse(start, inject, end))
+
+  /**
+   * Intersperses stream with provided element, similar to how [[scala.collection.immutable.List.mkString]]
+   * injects a separator between a List's elements.
+   *
+   * Additionally can inject start and end marker elements to stream.
+   *
+   * Examples:
+   *
+   * {{{
+   * Source<Integer, ?> nums = Source.from(Arrays.asList(0, 1, 2, 3));
+   * nums.intersperse(",");            //   1 , 2 , 3
+   * nums.intersperse("[", ",", "]");  // [ 1 , 2 , 3 ]
+   * }}}
+   *
+   * '''Emits when''' upstream emits (or before with the `start` element if provided)
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def intersperse[T >: Out](inject: T): javadsl.Flow[In, T, Mat] =
+    new Flow(delegate.intersperse(inject))
 
   /**
    * Chunk up this stream into groups of elements received within a time window,
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
@@ -554,6 +554,64 @@ class Source[+Out, +Mat](delegate: scaladsl.Source[Out, Mat]) extends Graph[Sour
     new Source(delegate.fold(zero)(f.apply))
 
   /**
+   * Intersperses stream with provided element, similar to how [[scala.collection.immutable.List.mkString]]
+   * injects a separator between a List's elements.
+   *
+   * Additionally can inject start and end marker elements to stream.
+   *
+   * Examples:
+   *
+   * {{{
+   * Source<Integer, ?> nums = Source.from(Arrays.asList(0, 1, 2, 3));
+   * nums.intersperse(",");            //   1 , 2 , 3
+   * nums.intersperse("[", ",", "]");  // [ 1 , 2 , 3 ]
+   * }}}
+   *
+   * In case you want to only prepend or only append an element (yet still use the `intercept` feature
+   * to inject a separator between elements, you may want to use the following pattern instead of the 3-argument
+   * version of intersperse (See [[Source.concat]] for semantics details):
+   *
+   * {{{
+   * Source.single(">> ").concat(list.intersperse(","))
+   * list.intersperse(",").concat(Source.single("END"))
+   * }}}
+   * '''Emits when''' upstream emits (or before with the `start` element if provided)
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def intersperse[T >: Out](start: T, inject: T, end: T): javadsl.Source[T, Mat] =
+    new Source(delegate.intersperse(start, inject, end))
+
+  /**
+   * Intersperses stream with provided element, similar to how [[scala.collection.immutable.List.mkString]]
+   * injects a separator between a List's elements.
+   *
+   * Additionally can inject start and end marker elements to stream.
+   *
+   * Examples:
+   *
+   * {{{
+   * Source<Integer, ?> nums = Source.from(Arrays.asList(0, 1, 2, 3));
+   * nums.intersperse(",");            //   1 , 2 , 3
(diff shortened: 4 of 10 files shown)


ktoso:
Failure was unrelated (https://github.com/akka/akka/issues/18735  )

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -239,6 +239,61 @@ private[akka] final case class Fold[In, Out](zero: Out, f: (Out, In) ⇒ Out, de
...
+
+  override def onUpstreamFinish(ctx: Context[T]): TerminationDirective = {
+    end match {
+      case Some(e) ⇒
+        // TODO could be exposed as method on Stage?
+        become(finishing)
+        ctx.absorbTermination()
+      case _ ⇒ ctx.finish()
+    }
+  }
```
ktoso: TODO make use of more vals

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -239,6 +239,61 @@ private[akka] final case class Fold[In, Out](zero: Out, f: (Out, In) ⇒ Out, de
...
+}
```
ktoso: This impl actually fails this test:

```scala

    "inject element between existing elements, when downstream is fold" in {
      val concated = Source(List(1, 2, 3))
        .map(_.toString)
        .intersperse(",")
        .runFold("")(_ + _)

      concated.futureValue should ===("1,2,3")
    }
```

@rkuhn @drewhk what am I doing wrong here hm...

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -239,6 +239,63 @@ private[akka] final case class Fold[In, Out](zero: Out, f: (Out, In) ⇒ Out, de
...
+      ctx.push(elem)
+    }
+  }
+
+  def running = new StageState[T, T] {
+    override def onPush(elem: T, ctx: Context[T]): SyncDirective =
+      emit(Iterator(inject, elem), ctx)
+  }
+
+  def finishing: StageState[T, T] = new StageState[T, T] {
```
rkuhn: this is not used

ktoso: good catch, thanks.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -239,6 +239,63 @@ private[akka] final case class Fold[In, Out](zero: Out, f: (Out, In) ⇒ Out, de
...
+      emit(Iterator(initial, elem), ctx, running)
+    }
+  }
+
+  def first = new StageState[T, T] {
+    override def onPush(elem: T, ctx: Context[T]) = {
+      become(running)
+      ctx.push(elem)
+    }
+  }
```
ktoso: TODO will try to make those as much vals as possible

rkuhn: nope: allocating them only when needed is better than doing it always

ktoso: cool, keeping as is

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -239,6 +239,63 @@ private[akka] final case class Fold[In, Out](zero: Out, f: (Out, In) ⇒ Out, de
...
+
+  override def onUpstreamFinish(ctx: Context[T]): TerminationDirective = {
+    end match {
+      case Some(e) if needsToEmitStart ⇒
+        terminationEmit(Iterator(start.get, end.get), ctx)
+      case Some(e) ⇒
+        terminationEmit(Iterator(end.get), ctx)
+      case _ ⇒
+        terminationEmit(Iterator(), ctx)
```
rkuhn: please use a cached empty Iterator (yes, I know, this is a bug in scala-library …)

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -554,6 +554,56 @@ class Source[+Out, +Mat](delegate: scaladsl.Source[Out, Mat]) extends Graph[Sour
...
+   *
+   * '''Emits when''' upstream emits (or before with the `start` element if provided)
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def intersperse[T >: Out](start: T, inject: T, end: T): javadsl.Source[Out, Mat] =
```
rkuhn: I’m torn API-wise: for mkString it is okay to use String arguments because there is an empty string, but here there may not be an “empty T” so we should allow either start or end to be left out. Making it four overloads is messy, making it Options is also not perfect …

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -605,6 +605,62 @@ trait FlowOps[+Out, +Mat] {
...
+   * '''Emits when''' upstream emits (or before with the `start` element if provided)
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def intersperse[T >: Out](start: T, inject: T, end: T): Repr[Out, Mat] = {
+    ReactiveStreamsCompliance.requireNonNullElement(start)
```
rkuhn: good point! do we have those on all value injection points?

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -407,6 +408,56 @@ class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Graph
     new Flow(delegate.fold(zero)(f.apply))
 
   /**
+   * Itersperses stream with provided element, similar to how [[scala.collection.immutable.List.mkString]]
```
johanandren: Itersperces -> Intersperces

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -554,6 +554,56 @@ class Source[+Out, +Mat](delegate: scaladsl.Source[Out, Mat]) extends Graph[Sour
...
+   * '''Emits when''' upstream emits (or before with the `start` element if provided)
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def intersperse[T >: Out](inject: T): javadsl.Source[Out, Mat] =
+    new Source(delegate.intersperse(Option.empty, inject, Option.empty))
```
rkuhn: what is this `Option.empty` about? shouldn’t this be the 1-arg method instead?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
ktoso: AFAIR no, can add in this PR

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -407,6 +408,56 @@ class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Graph
...
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def intersperse[T >: Out](start: T, inject: T, end: T): javadsl.Flow[In, Out, Mat] =
+    new Flow(delegate.intersperse(start, inject, end))
+
+  /**
+   * Itersperses stream with provided element, similar to how [[scala.collection.immutable.List.mkString]]
```
johanandren: Same as above - Itersperces

rkuhn:
LGTM after a conclusion on the API.

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
ktoso: Whoops, that a reminder from the impl that had the intersperse(Option, T, Option). Seems like missing tests that I didn't catch this

rkuhn: why does it even compile?

rkuhn: Ah, I see: you got the type signature wrong ;-) This should return a `Source[T, Mat]`.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -239,6 +239,63 @@ private[akka] final case class Fold[In, Out](zero: Out, f: (Out, In) ⇒ Out, de
 /**
  * INTERNAL API
  */
+private[akka] final case class Intersperse[T](start: Option[T], inject: T, end: Option[T]) extends StatefulStage[T, T] {
```
drewhk: does this really need to be a StatefulStage?

ktoso: I can make it a PushPullStage I guess - would that be much more efficient?
Is `emit` available on those?

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
drewhk: you still have source/flow concat, so I don't see it an issue:

     Source.single(start) ++ src.intersperce(separator)

ktoso: What's nice with that one is to be able to say "it's just like `mkString`".

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: we have it inside the processors I think.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: No emit, but I found it a bit overkill to have an Iterator/Iterable for maximum two elem output sequences.

You can have a two-elem array, a `read` index and an `end` index. Then you only need the code to emit at most two length sequences:

[start], elem1
sep, elem2
sep, elem3
sep, elem4
[end]

maybe overkill but I think it is simple enough.

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
drewhk: but it is not like mkString, because String define an empty String. I like the current API signature (1-param and 3-param) since the uncommon variants are still easy to achieve with concat (the only tricky part is the actual separator injection).

rkuhn: Hmm, okay. There’s no “perfect” solution here, unfortunately :-(

ktoso: @drewhk agreed. I can make it an example in the docs, how to use concat to achieve the "just prefix, no suffix".

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
ktoso: ouch, thanks!

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
ktoso: Yes, it does fail properly when running.
I did add it to `Source.single()` now though

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: done

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
ktoso: thanks

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: Made it a ticket https://github.com/akka/akka/issues/18736 want to push through json streaming today, can come back optimising this stage after

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
```diff
@@ -42,6 +44,8 @@ private[akka] final case class ErrorPublisher(t: Throwable, name: String) extend
 private[akka] final case class SingleElementPublisher[T](value: T, name: String) extends Publisher[T] {
   import ReactiveStreamsCompliance._
 
+  requireNonNullElement(value)
```
viktorklang: Nice catch :-)

ktoso: spidey senses kicked in ;]

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
viktorklang: Iterator.empty is better :)

ktoso: duh, seems that fix is on a different branch - checked after Roland commented and had Iterator.empty there :)
PR coming soon, will fix that one by the way

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -605,6 +605,71 @@ trait FlowOps[+Out, +Mat] {
...
+   * '''Emits when''' upstream emits (or before with the `start` element if provided)
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def intersperse[T >: Out](start: T, inject: T, end: T): Repr[T, Mat] = {
+    ReactiveStreamsCompliance.requireNonNullElement(start)
```
viktorklang: I don't think we throw RS SpecViolations anywhere else in the DSL. Perhaps use `require`?

viktorklang: This validation should go into the Intersperse constructor (I believe that's how the other ops do it)

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -233,16 +233,19 @@ object Source extends SourceApply {
   /**
    * Create a `Source` that will continually emit the given element.
    */
-  def repeat[T](element: T): Source[T, Unit] =
+  def repeat[T](element: T): Source[T, Unit] = {
+    ReactiveStreamsCompliance.requireNonNullElement(element)
```
viktorklang: I don't think this should be checked here. Is there any reason to believe that a null element in here would be different from pulling out a null from an Iterable?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
ktoso: good idea on moving them in, will do.

ktoso: The other addition here: https://github.com/akka/akka/pull/18734/files#diff-3eeced81b30519a848c0230d8ee399ecR47 
actually also happens "during DSL time", as it's a Publisher, not actor.

viktorklang: Consistency is dog :)

how are other nulls dealt with?

Source(() => Iterator.single(null)) etc :)

ktoso: > Consistency is dog :)

There's different flavours of dog :-)
The current state is consistent in the sense of "as soon as possible" - in the iterator we only can fail when looking into it, here we can fail right away. For dev time as-soon-as-possible is usually the best expirience IMO, you know where the null was passed in. Once it's flowing you get "well, some element in that pipeline was null, let's see hm hm..." instead of seeing it right away in the stacktrace – "oh, single was given a null".

I'd prefer the ASAP failing, but if you have strong feelings towards "always the same way - i.e. once running" I can adjust.

rkuhn: I’m with Konrad here, we should go for “ASAP”, with a best-effort for P.

drewhk: Yes, but let's make it a simple require, RS compliance exceptions should be thrown where RS is involved. Technically the DSL can be materialized to non-RS stuff.

ktoso: That's true as well, how about making those `require()`?

rkuhn: My comment was about when to complain, not how: `require` makes sense to me.

drewhk: ASAP+`require` is the consensus here.

viktorklang: affirmative

-- 
Cheers,
√

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -407,6 +408,65 @@ class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends Graph
...
+   * Intersperses stream with provided element, similar to how [[scala.collection.immutable.List.mkString]]
+   * injects a separator between a List's elements.
+   *
+   * Additionally can inject start and end marker elements to stream.
+   *
+   * Examples:
+   *
+   * {{{
+   * Source<Integer, ?> nums = Source.from(Arrays.asList(0, 1, 2, 3));
+   * nums.intersperse(",");            //   1 , 2 , 3
```
2beaucoup: the `0` is missing in all embedded examples

akka-stream/src/main/scala/akka/stream/impl/CompletedPublishers.scala
```diff
@@ -26,6 +26,8 @@ private[akka] case object EmptyPublisher extends Publisher[Nothing] {
  * INTERNAL API
  */
 private[akka] final case class ErrorPublisher(t: Throwable, name: String) extends Publisher[Nothing] {
+  ReactiveStreamsCompliance.requireNonNullElement(t)
+
   import ReactiveStreamsCompliance._
```
2beaucoup: switch these two lines?

PR #18745, opened 2015-10-20
[WIP] Framed entity streaming incl. Json Streaming (uploaded entity / response rendering)

### LINKED ISSUE #18358: Provide Marshallers to be able to Stream JSON for Source[A, _]
It should be possible to implement AFAICS,
like for Futures http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0/scala/http/common/marshalling.html#http-marshalling-scala

ktoso:
This is the combined and somewhat enriched result of @fsat's and mine work on json streaming.

Key features:
- syntax is just like people know and love from `entity(as[Person]) { person => }`,
  - now you can also: `entity(stream[Person]) { persons => }`
- adds `Framing.json` (needs some more love though)
- Any framing can be used (!) - PoC how a user would implement "CSV streaming" in 10 lines or less is attached as a test
- rendering the response is the usual `complete(Source[])`
  - a `Marshaller` and `FramingWithContentType` is required from the user (or like in SprayJsonSupport, we provide bits for the user)
  - FWCT allows us to reject routes if incoming request does not match content type that we expect
- allows control over parallelism in rendered stream (!) - `complete(source.renderAsync(5))` or `complete(source.renderAsyncUnordered(5))`

TODOs:
- [x] perf improvements for the JSON framing, it does too much currently (a lot of ByteStrings created), instead we can just count the `{` and `}` and concat incoming chunks of ByteStrings
- [x] Remove hardcoded application/json in one spot, should come from the unmarshaller.
- [ ] Figure out JavaDSL - can be a separate ticket

I may need to park this PR in its current state for a bit, but please do review so I can polish it up addressing comments along the way (the entire thing, independent commits are a bit chaotic).

Resolves #18358


typesafehub-validator:
Hi @fsat,

ktoso:
TODO: Investigate why `$ curl -O --verbose -X POST -H"Content-Type: application/json" --data @javaone-tweets.json http://localhost:8080/tweets` to `        entity(stream[Tweet]) { tweets: Source[Tweet, _] ⇒` runs the stream either slow or too late (?). Not sure yet, but it's not as expected.

akka-http-tests/src/test/scala/akka/http/scaladsl/marshallers/sprayjson/JsonStreamingSpec.scala
```diff
@@ -0,0 +1,170 @@
...
+      }
+
+      val json =
+        """{"name":"Smith","boardMember":false,"fname":"Frank","age":42,"id":12345}""" + "\n" +
+          """{"name":"Smith","boardMember":false,"fname":"Frank","age":42,"id":12345}""" + "\n" +
+          """{"name":"Smith","boardMember":false,"fname":"Frank","age":42,"id":12345}""" // TODO what if someone wants eager newline?
+
+      Get("/") ~> route ~> check {
+        responseAs[String] shouldEqual json
+        contentType shouldEqual ContentTypes.`application/json`
```
johanandren: Shouldn't it rather be 'application/x-ldjson' (https://en.wikipedia.org/wiki/Line_Delimited_JSON) or use an additional 'boundary=NL' part in the content type (https://groups.google.com/d/msg/nodejs/0ohwx0vF-SY/nf_wuQ0oUx0J)?

akka-http-tests/src/test/scala/akka/http/scaladsl/marshallers/sprayjson/JsonStreamingSpec.scala
```diff
@@ -0,0 +1,170 @@
...
+    """.stripMargin
+
+  val lineByLine =
+    """|{"fname":"Frank","name":"Smith","age":42,"id":1337,"boardMember":false}
+       |{"fname":"Bob","name":"Smith","age":42,"id":1337,"boardMember":false}
+       |{"fname":"Hank","name":"Smith","age":42,"id":1337,"boardMember":false}
+    """.stripMargin
+
+  val arrayPretty =
+    """|[
```
johanandren: This isn't a valid array in JSON, missing commas, but maybe there is something special about streaming arrays?

ktoso: Oh, good point – nope, it should have comas

akka-http/src/main/scala/akka/http/scaladsl/server/JsonEntityStreaming.scala
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * Specialised rendering mode for streaming elements as JSON.
+ *
+ * See also: <a href="https://en.wikipedia.org/wiki/JSON_Streaming">JSON Streaming on Wikipedia</a>.
+ */
+trait JsonSourceRenderingMode extends SourceRenderingMode
+
+// TODO framing needs a max-entity size.
+object JsonSourceRenderingMode {
```
johanandren: Not sure if this matters, but won't the `override def between = ByteString("c")` below allocate a new identical object for each chunk?

ktoso: true, can be implemented as a `val`

akka-http-tests/src/test/scala/akka/http/scaladsl/marshallers/sprayjson/JsonStreamingSpec.scala
ktoso: Doesn't seem there is consensus about it... https://groups.google.com/forum/#!topic/nodejs/0ohwx0vF-SY
adding the boundary sounds like a good idea though

ktoso: @sirthias @jrudolph WDYT, mime type of returned "new line separated streaming json" to be `application/json;boundary=NL`

ktoso: though boundary seems to imply `--....`? Leaving as is currently, there's not enough consensus about the other formats to stick to them just yet I feel.

sirthias: Hard to say. I haven't look into common approaches to JSON streaming so far.
But I'd say that something with content-type `application/json` should be legal JSON no matter what parameters are potentially attached to the content-type. That pretty much means that the top-level element must either be an array, whose elements are streamed (with commas in-between, as required) or an object, whose members are streamed (with commas in-between, as required). However, I'm not sure whether the latter approach is actually used anywhere in the wild.

If we break the JSON spec by introducing "funny" stream boundaries then we'd loose the ability to simply "re-chunk" the stream (e.g. by buffering). I think there is prettyness in *not* breaking the spec when streaming.

ktoso:
note to self: test using empty sources (in array mode should render `[]`).

akka-http-tests/src/test/scala/akka/http/scaladsl/marshallers/sprayjson/JsonStreamingSpec.scala
jrudolph: I agree with Mathias, I wouldn't use a non-standard extension of a common standard mime type because it's too easy to break something. (E.g. would it return something for client sending an `Accept: application/json` header?) 

Potential consequences would be less severe for a lesser used mime type like `application/x-ldjson`, so that's what I would prefer. Additionally, it would probably make sense to keep the media type easily configurable.

ktoso: Sounds like we have a plan here - keeping application/json and will make sure it's easy to configure if people want the fancy ones.

.java-version
```diff
@@ -0,0 +1 @@
+1.8
```
jrudolph: What uses this?

akka-http-tests/src/test/scala/akka/http/scaladsl/marshallers/sprayjson/JsonStreamingSpec.scala
jrudolph: I wouldn't default to `application/json` before making sure that clients which use `Accept: application/json` don't get illegal JSON.

ktoso: Content negotiation is still in play here, see tests that reject if user does not want json :)

.java-version
ktoso: Oh, missed this comment. I use this for `jenv`, but if someone minds it I'm happy to ignore it from git commits.

ktoso: http://www.jenv.be/

drewhk:
@ktoso temporarily closing this due to inactivity.

ktoso:
Yep, let's do that. 
I'll bring it back after 2.0 :+1:

akka-stream/src/main/scala/akka/stream/io/Framing.scala
```diff
@@ -59,6 +61,57 @@ object Framing {
...
+   * The framing works independently of formatting, i.e. it will still emit valid JSON elements even if two
+   * elements are separated by multiple newlines or other whitespace characters. And of course is insensitive
+   * (and does not impact the emitting frame) to the JSON object's internal formatting.
+   *
+   * Framing raw JSON values (such as integers or strings) is supported as well.
+   *
+   * @param maximumObjectLength The maximum length of allowed frames while decoding. If the maximum length is exceeded
+   *                         this Flow will fail the stream.
+   */
+  def json(maximumObjectLength: Int): Flow[ByteString, ByteString, Unit] =
```
ktoso: TODO compare perf with https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/json/JsonObjectDecoder.java

akka-stream/src/main/scala/akka/stream/impl/JsonBracketCounting.scala
```diff
@@ -0,0 +1,151 @@
...
+ * Copyright (C) 2014-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.impl
+
+import akka.stream.io.Framing.FramingException
+import akka.util.ByteString
+
+import scala.annotation.switch
+
+object JsonBracketCounting {
```
ktoso: TODO compare perf with https://github.com/netty/netty/blob/4.1/codec/src/main/java/io/netty/handler/codec/json/JsonObjectDecoder.java

Issue #18554, opened 2015-09-24
Akka cluster splitbrain on startup


brianwawok:
I was sent here from the user_group by @patriknw 

https://groups.google.com/forum/#!topic/akka-user/5RmljydLk3g

I am using AKKA 2.3.13

```
cluster {
  min-nr-of-members = 2
  seed-nodes = [
    "akka.tcp://System1@ip-197:9010",
    "akka.tcp://System1@ip-85:9010",
    "akka.tcp://System1@ip-86:9010"]
  auto-down-unreachable-after = off
  failure-detector {
    threshold = 16.0
    acceptable-heartbeat-pause = 10 s
    expected-response-after = 12 s
  }
```

I use the cluster singleton to split up work. 1 guy becomes the master, and the other 2 become workers.

Cool, 75% of the time it works.  25% of the time - it fails on startup. For some reason the nodes form 2 small clusters, one with 2 nodes, and one with only 1 node.  Once the cluster is up all is well.

My deployment procedure is:

1) Kill -9 all nodes
2) Wait 5 seconds
3) Start all 3 nodes one after another (few ms lag between them).

As you can see, the 197 IP is listed first.. so I guess that means everyone should converge on him. On my last event of this, I have 1 cluster of 197 and 85, with 85 elected the singleton.. and a 2nd cluster with 86 as the only member.

Crux of the problem is these log lines conflict on who the singleton is

```
Singleton identified: akka.tcp://MySystem@IP.85:9010/user/singleton/overmaster
Singleton identified: akka://MySystem/user/singleton/overmaster
Singleton identified: akka://MySystem/user/singleton/overmaster
```

1 and 2 agree that 2 is the singleton. 3 thinks that he is the singleton.

Log1: https://gist.github.com/brianwawok/d895d7f22891ec4b07c3
Log2: https://gist.github.com/brianwawok/3593efe4be78a5c70cb2
Log3: https://gist.github.com/brianwawok/3b71ff75a98ffa983c41

patriknw:
@brianwawok this is a bug. Thanks for reporting. Until we have fixed it you can work around it by using a delay (such as 5 seconds) between starting the nodes. Another way to mitigate the risk of this happening is to increase the `akka.cluster.leader-actions-interval` to let's say 4 seconds.

brianwawok:
it does look like my 3rd server has 3 or 4 second lead on the time, not sure if small clock skew is helping this along...

patriknw:
[...]

PR #18746, opened 2015-10-21
=clu #18554 Make oldest assignment deterministic when joining

patriknw:
- the reported issue is fixed by the immediate leaderActions
  (moving to Up)  when joining the first node to itself
- the other changes are precautions just in case


patriknw:
Refs #18554

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -325,7 +325,7 @@ class ShardRegion(
   def receiveClusterEvent(evt: ClusterDomainEvent): Unit = evt match {
     case MemberUp(m) ⇒
       if (matchingRole(m))
-        changeMembers(membersByAge + m)
+        changeMembers(membersByAge - m + m) // replace
```
ktoso: thanks for the comment, for a second thought was confused ;)

akka-cluster-tools/src/test/java/akka/cluster/singleton/ClusterSingletonManagerTest.java
```diff
@@ -46,77 +46,6 @@ public void demo() {
     //#create-singleton-proxy
   }
 
-  static//documentation of how to keep track of the oldest member in user land
-  //#singleton-proxy
```
ktoso: oh seems it wasn't ever linked to from docs. OK

ktoso:
LGTM.

Would it make sense to mention the `first joining itself will immediately be moved to Up` in the docs at least with a few words?

patriknw:
I think that is an implementation detail, which no happen to coincide with what users always have expected (first node joining is also the oldest and it will of course be moved to Up asap)

ktoso:
True, good as is then :)

akka-cluster-tools/src/main/scala/akka/cluster/singleton/ClusterSingletonProxy.scala
```diff
@@ -203,8 +201,9 @@ final class ClusterSingletonProxy(singletonManagerPath: String, settings: Cluste
    */
   def add(m: Member): Unit = {
     if (matchingRole(m))
-      trackChange {
-        () ⇒ membersByAge += m
+      trackChange { () ⇒
+        membersByAge -= m // replace
```
viktorklang: For consistency?

~~~scala
membersByAge = membersByAge - m + m // replace
~~~

akka-cluster/src/main/scala/akka/cluster/ClusterEvent.scala
```diff
@@ -275,7 +275,8 @@ object ClusterEvent {
       val newMembers = newGossip.members -- oldGossip.members
       val membersGroupedByAddress = List(newGossip.members, oldGossip.members).flatten.groupBy(_.uniqueAddress)
       val changedMembers = membersGroupedByAddress collect {
-        case (_, newMember :: oldMember :: Nil) if newMember.status != oldMember.status ⇒ newMember
+        case (_, newMember :: oldMember :: Nil) if newMember.status != oldMember.status || newMember.upNumber != oldMember.upNumber ⇒
+          newMember
       }
       val memberEvents = (newMembers ++ changedMembers) collect {
```
viktorklang: Hmm, didn't that performance article recommend using `diff` and `union` iso -- and ++?

patriknw: @viktorklang yes, that was something I learned from that presentation (surprised me that they are not implemented in the same way). I have changed: https://github.com/akka/akka/commit/811e28595f24de89210ceb85684c93fdb75ce0cb

viktorklang: :+1:

PR #18748, opened 2015-10-21
Automatic probe improvements suggestion (not for merging)

jrudolph:
IMO the automatic probes should take care of any demand handling on both sides.

This is only a tiny sketch, I tried more changes, but migrating all of the streams tests would take unreasonable time.

Going forward, to me it seems, the current implementation is currently somewhat undecided about what it wants to be. As far as I understand automatic probes of StreamTestKit should be the go-to tool for most low-level streams testing. However, the `Probe` classes are poorly documented and it's interaction with its superclass is easy to get wrong. E.g. it is easy to confuse the `expectNext` vs. `requestNext` methods. Also, the failure modes of the expectations are really ugly because they piggy-back on `TestProbe` expectations so that you get exceptions with stack traces where the actual test code is buried deeply and the messages contain too little detail (because many of them fail with `expectMsgType`). Instead, a testing tool should be focused on the failure situation so that it is simple to grasp a problem if it arises.

If automatic probes are deemed important (I think they are) they should be treated as such. I would suggest the following changes:
- don't extend from ManualProbe but make the underlying manual probe a field with a name and documentation that tells that you are usually not supposed to access it
- import and implement only the useful subset of the ManualProbe methods
- handle demand automatically
  - the subscriber-side `Probe.request` method should be replaced with an `ensureDemand(atLeast: Long)` which makes sure that the current `pendingRequests` counter is at least as big as the given value or otherwise requests more
  - the publisher-side `Probe.expectRequest` method should be replaced with an `expectDemand(atLeast: Long)` which ensures that the underlying `pendingRequests` counter is at least as big as the given value or, otherwise, expect more demand from the underlying probe until the counter reaches the given value or the expectation times out
- it would furthermore be nice if none of the expectation fail with an unhelpful error message that was generated by `TestProbe.expectMsgType` but that assertions are somehow wrapped to produce precise and detailed messages with compact stack traces

### DIFF
--- a/akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
+++ b/akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
@@ -167,7 +167,11 @@ object TestPublisher {
       this
     }
 
-    def expectRequest(): Long = subscription.expectRequest()
+    def expectRequest(): Long = {
+      val req = subscription.expectRequest()
+      pendingRequests += req
+      req
+    }
 
     def expectCancellation(): Self = {
       subscription.expectCancellation()
@@ -559,6 +563,8 @@ object TestSubscriber {
 
     override type Self = Probe[T]
 
+    var pendingRequests: Long = 0
+
     private lazy val subscription = expectSubscription()
 
     /** Asserts that a subscription has been received or will be received */
@@ -566,12 +572,14 @@ object TestSubscriber {
 
     def request(n: Long): Self = {
       subscription.request(n)
+      pendingRequests += n
       this
     }
 
     def requestNext(element: T): Self = {
-      subscription.request(1)
+      if (pendingRequests == 0) request(1)
       expectNext(element)
+      pendingRequests -= 1
       this
     }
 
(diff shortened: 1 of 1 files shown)


ktoso:
Something I noticed when helping out on the cassandra or kafka akka libs is that people used the manual ones, because they looked like the right thing to do, however they had no idea about demand and why/how, and blamed akka not working (i.e. they didn't signal demand, so it did nothing).

Related to such improvements is perhaps separating the "low level with manual demand" from the automatic ones very explicitly?

drewhk:
There are cases where one needs to start with a manual probe, but later wants the automatic methods, and vice versa.

Maybe separate the two probes into separate classes, but provide `toManual` and `toAutomatic` methods that upgrade or downgrade the functionality (carefully invalidating the original probe)?

drewhk:
Btw, non-detached, linear ops rarely need a probe, so we should also nudge the developers to use higher level tools, providing the missing pieces if necessary.

drewhk:
I am closing this in favor of ticket https://github.com/akka/akka/issues/18910. If someone wants to take this, he need to take the ticket. Until then, I close this PR.

Issue #18613, opened 2015-09-30
Possible performance regression in ConsistentHashingPool


drewhk:
Not fully confirmed, but looks like it happens during initialization, maybe some resize gone n^2?

See user report on ML: https://groups.google.com/forum/#!topic/akka-user/p344JTPpIh8

patriknw:
I don't think it's a problem with ConsistentHashingPool specifically, but with all pools configured with many routees. It looks like the time is spent in 

Here is some trace logging inside RepointableActorRef `sendSystemMessage` the for 10 routees. Each system message must iterate the queue until the next last element. You can imagine how that explodes for 100k.

    # tryEnqueue insertIntoIndex -1 Supervise(Actor[akka://ConsistentHashingRouterSpec/user/router1/$a#2121480742],false)
    # tryEnqueue insertIntoIndex -1 Supervise(Actor[akka://ConsistentHashingRouterSpec/user/router1/$a#2121480742],false)
    # tryEnqueue insertIntoIndex -1 Supervise(Actor[akka://ConsistentHashingRouterSpec/user/router1/$b#-1926314642],false)
    # tryEnqueue insertIntoIndex -1 Supervise(Actor[akka://ConsistentHashingRouterSpec/user/router1/$b#-1926314642],false)
    # tryEnqueue insertIntoIndex 2 Supervise(Actor[akka://ConsistentHashingRouterSpec/user/router1/$b#-1926314642],false)
    # queue.add insertIntoIndex 2 queue.size = 3  Supervise(Actor[akka://ConsistentHashingRouterSpec/user/router1/$b#-1926314642],false)
    # tryEnqueue insertIntoIndex -1 Supervise(Actor[akka://ConsistentHashingRouterSpec/user/router1/$c#310134968],false)
    # tryEnqueue insertIntoIndex -1 Supervise(Actor[akka://ConsistentHashingRouterSpec/user/router1/$c#310134968],false)
    # tryEnqueue insertIntoIndex 2 Supervise(Actor[akka://ConsistentHashingRouterSpec/user/router1/$c#310134968],false)
    # tryEnqueue insertIntoIndex 3 Supervise(Actor[akka://ConsistentHashingRouterSpec/user/router1/$c#310134968],false)
    # queue.add insertIntoIndex 3 queue.size = 4  Supervise(Actor[akka://ConsistentHashingRouterSpec/user/router1/$c#310134968],false)
[...]

PR #18754, opened 2015-10-21
=act #18613 Optimize RepointableActorRef for only sys msg

patriknw:
- Before the cell is started the RepointableActorRef queue the messages
  (both system and non-system messages) into a queue. System messages
  are inserted right before the last non-system message by iterating
  the queue, i.e. becomes slow when there are many system messages already
  enqueued.
- This is a problem for pools with many routees.
- The fix is to keep track of that there are only system messages
  and then only add the message to the queue without any iteration.

JMH benchmark before:

```
[info] Benchmark                                 (size)  Mode  Cnt       Score      Error  Units
[info] RouterPoolCreationBenchmark.testCreation    1000    ss  300   13204.048 ± 3081.576  us/op
[info] RouterPoolCreationBenchmark.testCreation    2000    ss  300   41939.524 ± 6178.087  us/op
[info] RouterPoolCreationBenchmark.testCreation    3000    ss  300   70752.881 ± 4344.992  us/op
[info] RouterPoolCreationBenchmark.testCreation    4000    ss  300  120620.885 ± 3296.342  us/op
```

JMH benchmark after:

```
[info] Benchmark                                 (size)  Mode  Cnt      Score       Error  Units
[info] RouterPoolCreationBenchmark.testCreation    1000    ss  300  10867.705 ±  2817.712  us/op
[info] RouterPoolCreationBenchmark.testCreation    2000    ss  300  21018.967 ±  5363.646  us/op
[info] RouterPoolCreationBenchmark.testCreation    3000    ss  300  33191.025 ± 14053.836  us/op
[info] RouterPoolCreationBenchmark.testCreation    4000    ss  300  41751.298 ± 12960.081  us/op
```

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/RepointableActorRef.scala
+++ b/akka-actor/src/main/scala/akka/actor/RepointableActorRef.scala
@@ -183,17 +183,35 @@ private[akka] class UnstartedCell(val systemImpl: ActorSystemImpl,
   private[this] final val lock = new ReentrantLock
 
   // use Envelope to keep on-send checks in the same place ACCESS MUST BE PROTECTED BY THE LOCK
-  private[this] final val queue = new JLinkedList[Any]()
+  private[this] final val queue = new JLinkedList[Envelope]()
+
+  // ACCESS MUST BE PROTECTED BY THE LOCK
+  private[this] var sysmsgQueue: LatestFirstSystemMessageList = SystemMessageList.LNil
 
   import systemImpl.settings.UnstartedPushTimeout.{ duration ⇒ timeout }
 
   def replaceWith(cell: Cell): Unit = locked {
     try {
-      while (!queue.isEmpty) {
-        queue.poll() match {
-          case s: SystemMessage ⇒ cell.sendSystemMessage(s)
-          case e: Envelope      ⇒ cell.sendMessage(e)
+      def drainSysmsgQueue(): Unit = {
+        // using while in case a sys msg enqueues another sys msg
+        while (sysmsgQueue.nonEmpty) {
+          var sysQ = sysmsgQueue.reverse
+          sysmsgQueue = SystemMessageList.LNil
+          while (sysQ.nonEmpty) {
+            val msg = sysQ.head
+            sysQ = sysQ.tail
+            msg.unlink()
+            cell.sendSystemMessage(msg)
+          }
         }
+      }
+
+      drainSysmsgQueue()
+
+      while (!queue.isEmpty) {
+        cell.sendMessage(queue.poll())
+        // drain sysmsgQueue in case a msg enqueues a sys msg
+        drainSysmsgQueue()
       }
     } finally {
       self.swapCell(cell)
@@ -236,25 +254,11 @@ private[akka] class UnstartedCell(val systemImpl: ActorSystemImpl,
     lock.lock // we cannot lose system messages, ever, and we cannot throw an Error from here as well
     try {
       val cell = self.underlying
-      if (cellIsReady(cell)) {
+      if (cellIsReady(cell))
         cell.sendSystemMessage(msg)
-      } else {
-        // systemMessages that are sent during replace need to jump to just after the last system message in the queue, so it's processed before other messages
-        val wasEnqueued = if ((self.lookup ne this) && (self.underlying eq this) && !queue.isEmpty()) {
-          @tailrec def tryEnqueue(i: JListIterator[Any] = queue.listIterator(), insertIntoIndex: Int = -1): Boolean =
-            if (i.hasNext())
-              tryEnqueue(i,
-                if (i.next().isInstanceOf[SystemMessage]) i.nextIndex() // update last sysmsg seen so far
-                else insertIntoIndex) // or just keep the last seen one
-            else if (insertIntoIndex == -1) queue.offer(msg)
-            else Try(queue.add(insertIntoIndex, msg)).isSuccess
-          tryEnqueue()
-        } else queue.offer(msg)
-
-        if (!wasEnqueued) {
-          system.eventStream.publish(Warning(self.path.toString, getClass, "dropping system message " + msg + " due to enqueue failure"))
-          system.deadLetters ! DeadLetter(msg, self, self)
-        } else if (Mailbox.debug) println(s"$self temp queueing system $msg")
+      else {
+        sysmsgQueue ::= msg
+        if (Mailbox.debug) println(s"$self temp queueing system $msg")
       }
     } finally lock.unlock()
   }
--- a/akka-bench-jmh/src/main/scala/akka/actor/RouterPoolCreationBenchmark.scala
+++ b/akka-bench-jmh/src/main/scala/akka/actor/RouterPoolCreationBenchmark.scala
@@ -1,0 +1,43 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.actor
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import akka.routing.RoundRobinPool
+import akka.testkit.TestActors
+import akka.testkit.TestProbe
+import org.openjdk.jmh.annotations._
+import java.util.concurrent.TimeUnit
+
+@State(Scope.Benchmark)
+@BenchmarkMode(Array(Mode.SingleShotTime))
+@Fork(3)
+@Warmup(iterations = 20)
+@Measurement(iterations = 100)
+class RouterPoolCreationBenchmark {
+  implicit val system: ActorSystem = ActorSystem()
+  val probe = TestProbe()
+
+  Props[TestActors.EchoActor]
+
+  @Param(Array("1000", "2000", "3000", "4000"))
+  var size = 0
+
+  @TearDown(Level.Trial)
+  def shutdown() {
+    system.terminate()
+    Await.ready(system.whenTerminated, 15.seconds)
+  }
+
+  @Benchmark
+  @OutputTimeUnit(TimeUnit.MICROSECONDS)
+  def testCreation: Boolean = {
+    val pool = system.actorOf(RoundRobinPool(size).props(TestActors.echoActorProps))
+    pool.tell("hello", probe.ref)
+    probe.expectMsg(5.seconds, "hello")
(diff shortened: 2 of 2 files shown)


patriknw:
Refs #18613

akka-actor/src/main/scala/akka/actor/RepointableActorRef.scala
```diff
@@ -222,6 +226,8 @@ private[akka] class UnstartedCell(val systemImpl: ActorSystemImpl,
         if (cellIsReady(cell)) {
           cell.sendMessage(msg)
         } else if (!queue.offer(msg)) {
+          println(s"# enqueuedNonSystemMsg $msg") // FIXME
+          enqueuedNonSystemMsg = true
```
rkuhn: This needs to happen in the other branch (i.e. if the offer was successful).

rkuhn:
LGTM after fixup.

akka-actor/src/main/scala/akka/actor/RepointableActorRef.scala
patriknw: ouch, thanks

viktorklang: Have we considered keeping a var to a SystemMessage that we prepend to until we repoint it, where we reverse it and send them (the SystemMessages) to the "real" actor ref?

patriknw:
@viktorklang suggestion with a separate system messages queue is excellent. I have updated to that. Please review carefully.
cc @rkuhn 

Latest benchmark:

```
JMH benchmark before:

    [info] Benchmark                                 (size)  Mode  Cnt       Score      Error  Units
    [info] RouterPoolCreationBenchmark.testCreation    1000    ss  300   13204.048 ± 3081.576  us/op
    [info] RouterPoolCreationBenchmark.testCreation    2000    ss  300   41939.524 ± 6178.087  us/op
    [info] RouterPoolCreationBenchmark.testCreation    3000    ss  300   70752.881 ± 4344.992  us/op
    [info] RouterPoolCreationBenchmark.testCreation    4000    ss  300  120620.885 ± 3296.342  us/op

JMH benchmark after:

    [info] Benchmark                                 (size)  Mode  Cnt      Score      Error  Units
    [info] RouterPoolCreationBenchmark.testCreation    1000    ss  300   7738.721 ± 1806.297  us/op
    [info] RouterPoolCreationBenchmark.testCreation    2000    ss  300  15497.588 ± 4532.852  us/op
    [info] RouterPoolCreationBenchmark.testCreation    3000    ss  300  28704.005 ± 6322.458  us/op
    [info] RouterPoolCreationBenchmark.testCreation    4000    ss  300  37783.516 ± 6778.437  us/op
```

viktorklang:
@patriknw wow, that looks like a great improvement!

akka-actor/src/main/scala/akka/actor/RepointableActorRef.scala
```diff
@@ -183,17 +183,26 @@ private[akka] class UnstartedCell(val systemImpl: ActorSystemImpl,
   private[this] final val lock = new ReentrantLock
 
   // use Envelope to keep on-send checks in the same place ACCESS MUST BE PROTECTED BY THE LOCK
-  private[this] final val queue = new JLinkedList[Any]()
+  private[this] final val queue = new JLinkedList[Envelope]()
```
viktorklang: Ahhh, the fresh smell of types!

akka-actor/src/main/scala/akka/actor/RepointableActorRef.scala
```diff
@@ -236,26 +245,10 @@ private[akka] class UnstartedCell(val systemImpl: ActorSystemImpl,
...
-            else if (insertIntoIndex == -1) queue.offer(msg)
-            else Try(queue.add(insertIntoIndex, msg)).isSuccess
-          tryEnqueue()
-        } else queue.offer(msg)
-
-        if (!wasEnqueued) {
-          system.eventStream.publish(Warning(self.path.toString, getClass, "dropping system message " + msg + " due to enqueue failure"))
-          system.deadLetters ! DeadLetter(msg, self, self)
-        } else if (Mailbox.debug) println(s"$self temp queueing system $msg")
-      }
```
viktorklang: Wow, much cleaner!

akka-actor/src/main/scala/akka/actor/RepointableActorRef.scala
```diff
@@ -183,17 +183,26 @@ private[akka] class UnstartedCell(val systemImpl: ActorSystemImpl,
   private[this] final val lock = new ReentrantLock
 
   // use Envelope to keep on-send checks in the same place ACCESS MUST BE PROTECTED BY THE LOCK
-  private[this] final val queue = new JLinkedList[Any]()
+  private[this] final val queue = new JLinkedList[Envelope]()
+
+  // ACCESS MUST BE PROTECTED BY THE LOCK
+  private[this] var sysmsgQueue: LatestFirstSystemMessageList = SystemMessageList.LNil
```
viktorklang: :+1:

patriknw:
@viktorklang @rkuhn I suspect that the improved performance revealed a bug creeping around in the darkness of the dungeon. The failing test is "not double-deregister" in CallingThreadDispatcherModelSpec

I can produce it locally, and I see that two `Terminate()` are sent, and as far as I can see we have nothing preventing the second to be ignored. This is executed twice: https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/actor/dungeon/FaultHandling.scala#L171

rkuhn:
How can this possibly happen? We could of course getAndSet instead of putVolatile and bail out if the previous already was terminated, but I would find that bothersome on a deeper level. Even if multiple Terminate commands are enqueued, after processing the first nothing else must ever happen. Perhaps your additional queue is not drained to deadLetters properly?

akka-actor/src/main/scala/akka/actor/RepointableActorRef.scala
```diff
@@ -236,26 +245,10 @@ private[akka] class UnstartedCell(val systemImpl: ActorSystemImpl,
     lock.lock // we cannot lose system messages, ever, and we cannot throw an Error from here as well
     try {
       val cell = self.underlying
-      if (cellIsReady(cell)) {
+      if (cellIsReady(cell))
         cell.sendSystemMessage(msg)
-      } else {
-        // systemMessages that are sent during replace need to jump to just after the last system message in the queue, so it's processed before other messages
-        val wasEnqueued = if ((self.lookup ne this) && (self.underlying eq this) && !queue.isEmpty()) {
```
patriknw: @rkuhn is it these checks that you are thinking about?
feels strange to check here, perhaps it should be checked in `replaceCell` for each sys msg?

patriknw: These checks should be added back, but that is not enough.
I think it's the calling thread dispatcher that is playing games with me.
I'll look into it more next week.

patriknw:
Nailed it! This is what was going on:
1. called `replaceWith` in repointable, no sys msg so far
2. one enqueued normal message, delivered
3. the normal message generated 2 system messages, which are enqueued in the new `sysmsgQueue`
4. but the `sysmsgQueue` was already been drained in `replaceWith`, and the new system messages were never delivered

It worked previously because the single `queue` is mutable.

Solved in eb43c57

akka-actor/src/main/scala/akka/actor/RepointableActorRef.scala
patriknw: This is something that I'm still unsure about. What was the purpose of the `(self.lookup ne this) && (self.underlying eq this)` check? I tried to add it back like:

    if ((self.lookup ne this) && (self.underlying eq this)) {
          sysmsgQueue ::= msg

but that cause `akka.routing.RoutingSpec` to fail due to dropped Watch sys msg.

In the old code, `queue.offer` will never return `false` so `wasEnqueued` will always be `true`.

viktorklang: @patriknw I think the reason is just what the comment says: "// systemMessages that are sent during replace need to jump to just after the last system message in the queue, so it's processed before other messages"

patriknw: Right, system messages can/should always be processed before other messages, as far as I know. 
Do you see anything in the new implementation (see also `replaceWith`) that changes this behavior?

viktorklang: @patriknw Hmm, I'm too rusty on that pieceof logic. One thing I'd check if I were you would be possible reentrancy of `replaceWith`, i.e. is it possible for the same thread that is currently calling `replaceWith` to call it again before it exits?

patriknw: To make my question clear, I'm not looking for any more bug. I found the issue, I just wanted second review confirmation that I have not messed up anything. Thanks.

viktorklang: The Klang can only prove the existence of bugs, not the non-existence of bugs. ;)

patriknw: :-) at least I tried

rkuhn:
LGTM

akka-actor/src/main/scala/akka/actor/RepointableActorRef.scala
drewhk: so these lines are safe to remove?

drewhk:
Looks ok, but I am unsure about that big red blob of removed code. This brings back memories though, making the system message list safer via value classes :)

akka-actor/src/main/scala/akka/actor/RepointableActorRef.scala
patriknw: yeah, all they do is to insert the sys msg before the first non-sys msg (and now we have two separate queues, and always drain the sys msg queue first)

drewhk: a-ha, that makes sense.

patriknw:
squashed, will merge after another pr validation

PR #18769, opened 2015-10-23
=cdd #18768 Cache serialization of read and write msg

patriknw:
Something I did on the plane last week.

The WriteAggregator and ReadAggregator typically send
the same message to several replicas and by caching the serialized bytes
we avoid doing the same thing for each message

add a test for WriteAggregator

### DIFF
--- a/akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
+++ b/akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
@@ -1359,13 +1359,15 @@ private[akka] class WriteAggregator(
 
   def receive = {
     case WriteAck ⇒
-      remaining -= sender().path.address
+      remaining -= senderAddress()
       if (remaining.size == doneWhenRemainingSize)
         reply(ok = true)
     case SendToSecondary ⇒
       secondaryNodes.foreach { replica(_) ! writeMsg }
     case ReceiveTimeout ⇒ reply(ok = false)
   }
+
+  def senderAddress(): Address = sender().path.address
 
   def reply(ok: Boolean): Unit = {
     if (ok && envelope.data == DeletedData)
--- a/akka-distributed-data/src/main/scala/akka/cluster/ddata/protobuf/ReplicatorMessageSerializer.scala
+++ b/akka-distributed-data/src/main/scala/akka/cluster/ddata/protobuf/ReplicatorMessageSerializer.scala
@@ -3,6 +3,7 @@
  */
 package akka.cluster.ddata.protobuf
 
+import scala.concurrent.duration._
 import java.util.concurrent.TimeUnit
 import scala.collection.JavaConverters._
 import scala.collection.breakOut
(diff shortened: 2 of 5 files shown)


akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
```diff
@@ -1367,6 +1367,8 @@ private[akka] class WriteAggregator(
     case ReceiveTimeout ⇒ reply(ok = false)
   }
 
+  def senderAddress(): Address = sender().path.address
```
viktorklang: why?

akka-distributed-data/src/main/scala/akka/cluster/ddata/protobuf/ReplicatorMessageSerializer.scala
```diff
@@ -80,9 +171,9 @@ class ReplicatorMessageSerializer(val system: ExtendedActorSystem)
 
   def toBinary(obj: AnyRef): Array[Byte] = obj match {
     case m: DataEnvelope   ⇒ dataEnvelopeToProto(m).toByteArray
-    case m: Write          ⇒ writeToProto(m).toByteArray
-    case WriteAck          ⇒ dm.Empty.getDefaultInstance.toByteArray
-    case m: Read           ⇒ readToProto(m).toByteArray
+    case m: Write          ⇒ writeCache.getOrAdd(m, () ⇒ writeToProto(m).toByteArray)
```
viktorklang: Instead of allocating this function on each Write, why not prove the function when constructing the SmallCache?

akka-distributed-data/src/main/scala/akka/cluster/ddata/protobuf/ReplicatorMessageSerializer.scala
```diff
@@ -21,12 +21,103 @@ import akka.serialization.BaseSerializer
...
+}
 
 /**
  * Protobuf serializer of ReplicatorMessage messages.
  */
 class ReplicatorMessageSerializer(val system: ExtendedActorSystem)
   extends SerializerWithStringManifest with SerializationSupport with BaseSerializer {
+  import ReplicatorMessageSerializer.SmallCache
+
+  private val readCache = new SmallCache[Read, Array[Byte]](4)
```
viktorklang: What is the foreseen implication of having a non-expiring cache for this?

akka-distributed-data/src/main/scala/akka/cluster/ddata/protobuf/ReplicatorMessageSerializer.scala
```diff
@@ -21,12 +21,103 @@ import akka.serialization.BaseSerializer
...
+
+    /**
+     * Add entry to the cache.
+     * Overwrites oldest entry.
+     */
+    def add(a: A, b: B): Unit = add((a, b))
+
+    def add(x: (A, B)): Unit = {
+      val i = n.incrementAndGet()
+      elements(i & mask) = x
```
viktorklang: What if `x` is `null`?

akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala
patriknw: needed for the new test `WriteAggregatorSpec`, which is not really related to the cache (I just increased the tests coverage a bit)

akka-distributed-data/src/main/scala/akka/cluster/ddata/protobuf/ReplicatorMessageSerializer.scala
patriknw: good point, thanks

akka-distributed-data/src/main/scala/akka/cluster/ddata/protobuf/ReplicatorMessageSerializer.scala
patriknw: I didn't consider it to be a big deal to keep around a few if no more activity, but I can add simple TTL eviction

akka-distributed-data/src/main/scala/akka/cluster/ddata/protobuf/ReplicatorMessageSerializer.scala
patriknw: I'm not using null, but I think it can actually work, because the cache elements are initially `null.

patriknw:
@viktorklang I have added the cache eviction and changed so that the `getOrAdd` function is passed in to the constructor instead, see 3302b14. Thanks for feedback!

patriknw:
Refs #18768

Issue #18728, opened 2015-10-14
peristAll throws when nil


ahjohannessen:
If one invokes `persistAll` with an empty sequence it will cause the persistent actor to crash. Repro:

```scala
import scala.io.StdIn
import akka.actor.{Props, ActorSystem}
import akka.persistence.PersistentActor

object OhNoes extends App {

  val system = ActorSystem("Crash")
  system.actorOf(Props[Persister]) ! "Oh Noes"

  StdIn.readLine()
  system.terminate()

  class Persister extends PersistentActor {
    def persistenceId  = "crash-bob"
    def receiveCommand = { case "Oh Noes" ⇒ persistAll(Nil)(_ ⇒ ()) }
    def receiveRecover = { case x ⇒ }
  }
}
```

Exception being thrown:

```
[ERROR] [...] [Crash-akka.actor.default-dispatcher-4] [akka://Crash/user/$a] tail of empty list
java.lang.UnsupportedOperationException: tail of empty list
	at scala.collection.immutable.Nil$.tail(List.scala:422)
	at scala.collection.immutable.Nil$.tail(List.scala:417)
	at akka.persistence.AtomicWrite.<init>(Persistent.scala:44)
	at akka.persistence.Eventsourced$class.persistAll(Eventsourced.scala:310)
...
```

One can imagine that `C => S => Err \/ Seq[E]` is a common use case where issuing a command may result in no event. The fix is easily done in user land like:

```scala
    if(evts.isEmpty) done else {
      persistAll(evts)(applyEvent)
      deferAsync()(_ ⇒ done)
    }
```

But why not let Akka do the right thing and allow for the user to:

```scala
  persistAll(evts)(applyEvent)
  deferAsync()(_ ⇒ done)
```

ping @ktoso @patriknw

patriknw:
Good catch.

ktoso:
Thanks for reporting @ahjohannessen, great catch!

ahjohannessen:
:+1:

ahjohannessen:
:+1:

PR #18772, opened 2015-10-24
=per #18728 allow persistAll(Nil)

ktoso:
Resolves https://github.com/akka/akka/issues/18728

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
@@ -303,11 +303,13 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
    * @param handler handler for each persisted `events`
    */
   def persistAll[A](events: immutable.Seq[A])(handler: A ⇒ Unit): Unit = {
-    events.foreach { event ⇒
-      pendingStashingPersistInvocations += 1
-      pendingInvocations addLast StashingHandlerInvocation(event, handler.asInstanceOf[Any ⇒ Unit])
+    if (events.nonEmpty) {
+      events.foreach { event ⇒
+        pendingStashingPersistInvocations += 1
+        pendingInvocations addLast StashingHandlerInvocation(event, handler.asInstanceOf[Any ⇒ Unit])
+      }
+      eventBatch = AtomicWrite(events.map(PersistentRepr.apply(_, sender = sender()))) :: eventBatch
     }
-    eventBatch = AtomicWrite(events.map(PersistentRepr.apply(_, sender = sender()))) :: eventBatch
   }
 
   @deprecated("use persistAll instead", "2.4")
@@ -350,12 +352,13 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
    * @param events events to be persisted
    * @param handler handler for each persisted `events`
    */
-  def persistAllAsync[A](events: immutable.Seq[A])(handler: A ⇒ Unit): Unit = {
-    events.foreach { event ⇒
-      pendingInvocations addLast AsyncHandlerInvocation(event, handler.asInstanceOf[Any ⇒ Unit])
+  def persistAllAsync[A](events: immutable.Seq[A])(handler: A ⇒ Unit): Unit =
+    if (events.nonEmpty) {
+      events.foreach { event ⇒
+        pendingInvocations addLast AsyncHandlerInvocation(event, handler.asInstanceOf[Any ⇒ Unit])
+      }
+      eventBatch = AtomicWrite(events.map(PersistentRepr.apply(_, sender = sender()))) :: eventBatch
     }
-    eventBatch = AtomicWrite(events.map(PersistentRepr.apply(_, sender = sender()))) :: eventBatch
-  }
 
   @deprecated("use persistAllAsync instead", "2.4")
   def persistAsync[A](events: immutable.Seq[A])(handler: A ⇒ Unit): Unit =
--- a/akka-persistence/src/main/scala/akka/persistence/Persistent.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/Persistent.scala
@@ -4,13 +4,13 @@
 
 package akka.persistence
 
-import scala.collection.immutable
 import java.lang.{ Iterable ⇒ JIterable }
 import java.util.{ List ⇒ JList }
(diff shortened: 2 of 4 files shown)


akka-persistence/src/test/scala/akka/persistence/PersistentActorSpec.scala
```diff
@@ -248,6 +248,14 @@ object PersistentActorSpec {
         persistAsync(event) { evt ⇒ sender() ! s"${evt.data}-b-${sendMsgCounter.incrementAndGet()}" }
     }
   }
+  class PersistAllNilPersistentActor(name: String) extends ExamplePersistentActor(name) {
+
+    val receiveCommand: Receive = commonBehavior orElse {
+      case Cmd(data) ⇒
+        persistAll(Nil)(_ ⇒ ())
```
patriknw: So this loops the journal and then calls the handler? I think that is necessary for ordering.

akka-persistence/src/main/scala/akka/persistence/Persistent.scala
```diff
@@ -41,7 +41,7 @@ final case class AtomicWrite(payload: immutable.Seq[PersistentRepr]) extends Per
 
   // only check that all persistenceIds are equal when there's more than one in the Seq
   if (payload match {
-    case l: List[PersistentRepr]   ⇒ l.tail.nonEmpty
+    case l: List[PersistentRepr]   ⇒ l.nonEmpty && l.tail.nonEmpty // avoids calling .size
```
patriknw: Looks like `AtomicWrite` is not really designed for empty `payload`, there are problems in `lowestSequenceNr` and `highestSequenceNr` also. I don't mind if we we use it internally for looping to the journal, but should avoid calling `asynWriteMessages` with empty `AtomicWrite` ?

ktoso: Yeah, good point - I'll fix this in another way then

akka-persistence/src/test/scala/akka/persistence/PersistentActorSpec.scala
rkuhn: Hmm, persistAll invokes the handler once for every element—which means not at all in this case. I don’t think there is a need to loop anything here.

akka-persistence/src/main/scala/akka/persistence/Persistent.scala
rkuhn: The fix should rather be to not emit any write for an empty list.

akka-persistence/src/test/scala/akka/persistence/PersistentActorSpec.scala
ktoso: Yeah, the callback is never called so doesn't really matter if we do not call it without looping vs. not call it with looping :)

akka-persistence/src/main/scala/akka/persistence/Persistent.scala
ktoso: > The fix should rather be to not emit any write for an empty list. 

Yeah, very true. I'll amend the PR

ktoso:
Updated the PR.

akka-persistence/src/test/scala/akka/persistence/PersistentActorSpec.scala
```diff
@@ -785,6 +795,13 @@ abstract class PersistentActorSpec(config: Config) extends PersistenceSpec(confi
       expectMsg("x-b-2")
       expectNoMsg(100.millis)
     }
+    "support calling persistAll with Nil" in {
+      val persistentActor = namedPersistentActor[PersistAllNilPersistentActor]
+      persistentActor ! Cmd("x")
+      expectMsg("before-nil")
+      expectMsg("after-nil")
+      expectMsg("x")
```
rkuhn: shouldn’t this message come first?

ktoso: If preceeding calls were `persist()` then yes, but they're `deferAsync()`.
The persist test is more interesting, you're right - I'll change the test to exercise that path instead.

rkuhn:
LGTM apart from the test that I would expect to fail—unless I’m missing something or it is too late here :-)

ktoso:
Test is fine, but boring case :)
Have a good evening! Thanks for all the reviews so late.

akka-persistence/src/test/scala/akka/persistence/PersistentActorSpec.scala
rkuhn: Ah, I wasn’t consciously aware of the optimization within `deferAsync` to immediately execute when possible—dare I say that this might be confusing?

rkuhn:
Please retain the boring test case, I see the rationale now.

akka-persistence/src/test/scala/akka/persistence/PersistentActorSpec.scala
ktoso: Hm, I wonder if it is confusing... I may be biased as I implemented the current semantics.
`deferAsync` defers in respect to `persist*` calls, not "in general".
If you think we should address this let's open a ticket.

rkuhn: Well, now I do remember being part of the decision to do this optimization—why go through the journal when that is not needed. I’m not sure it’s bad enough to ticketify, let’s watch out for related user feedback.

ktoso: :+1: waiting for user feedback sounds good. I'll add the test and merge :)

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -303,11 +303,13 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
    * @param handler handler for each persisted `events`
    */
   def persistAll[A](events: immutable.Seq[A])(handler: A ⇒ Unit): Unit = {
-    events.foreach { event ⇒
-      pendingStashingPersistInvocations += 1
-      pendingInvocations addLast StashingHandlerInvocation(event, handler.asInstanceOf[Any ⇒ Unit])
+    if (events.nonEmpty) {
```
patriknw: :-) good one

PR #18773, opened 2015-10-25
Agolubev #17967 sink.actor ref with acking

agolubev:
Ref #17967

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefBackpressureSinkSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefBackpressureSinkSpec.scala
@@ -1,0 +1,115 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.scaladsl
+
+import akka.actor.{ Actor, ActorRef, Props }
+import akka.stream.ActorMaterializer
+import akka.stream.testkit.Utils._
+import akka.stream.testkit._
+import akka.stream.testkit.scaladsl._
+
+object ActorRefBackpressureSinkSpec {
+  val initMessage = "start"
+  val completeMessage = "done"
+  val ackMessage = "ack"
+
+  class Fw(ref: ActorRef) extends Actor {
+    def receive = {
+      case `initMessage` ⇒
+        sender() ! ackMessage
+        ref forward initMessage
+      case `completeMessage` ⇒
+        ref forward completeMessage
+      case msg: Int ⇒
+        sender() ! ackMessage
+        ref forward msg
+    }
+  }
+
+  case object TriggerAckMessage
+
+  class Fw2(ref: ActorRef) extends Actor {
+    var actorRef: ActorRef = Actor.noSender
+
+    def receive = {
+      case TriggerAckMessage ⇒
+        actorRef ! ackMessage
+      case msg ⇒
+        actorRef = sender()
+        ref forward msg
+    }
+  }
+
+}
+
+class ActorRefBackpressureSinkSpec extends AkkaSpec {
+  import ActorRefBackpressureSinkSpec._
+  implicit val mat = ActorMaterializer()
+
+  def createActor[T](c: Class[T]) =
+    system.actorOf(Props(c, testActor).withDispatcher("akka.test.stream-dispatcher"))
+
+  "An ActorRefSink" must {
+
+    "send the elements to the ActorRef" in assertAllStagesStopped {
+      val fw = createActor(classOf[Fw])
+      Source(List(1, 2, 3)).runWith(Sink.actorRefWithAck(fw,
+        initMessage, ackMessage, completeMessage))
+      expectMsg("start")
+      expectMsg(1)
+      expectMsg(2)
+      expectMsg(3)
+      expectMsg(completeMessage)
+    }
+
+    "send the elements to the ActorRef2" in assertAllStagesStopped {
+      val fw = createActor(classOf[Fw])
+      val probe = TestSource.probe[Int].to(Sink.actorRefWithAck(fw,
+        initMessage, ackMessage, completeMessage)).run()
+      probe.sendNext(1)
+      expectMsg("start")
+      expectMsg(1)
+      probe.sendNext(2)
+      expectMsg(2)
+      probe.sendNext(3)
+      expectMsg(3)
+      probe.sendComplete()
+      expectMsg(completeMessage)
+    }
+
+    "cancel stream when actor terminates" in assertAllStagesStopped {
+      val fw = createActor(classOf[Fw])
+      val publisher = TestSource.probe[Int].to(Sink.actorRefWithAck(fw,
+        initMessage, ackMessage, completeMessage)).run().sendNext(1)
+      expectMsg(initMessage)
+      expectMsg(1)
+      system.stop(fw)
+      publisher.expectCancellation()
+    }
+
+    "send message only when backpressure received" in assertAllStagesStopped {
+      val fw = createActor(classOf[Fw2])
+      val publisher = TestSource.probe[Int].to(Sink.actorRefWithAck(fw,
+        initMessage, ackMessage, completeMessage)).run()
+      expectMsg(initMessage)
+
+      publisher.sendNext(1)
+      expectNoMsg()
+      fw ! TriggerAckMessage
+      expectMsg(1)
+
+      publisher.sendNext(2)
+      publisher.sendNext(3)
+      publisher.sendComplete()
+      fw ! TriggerAckMessage
+      expectMsg(2)
+      fw ! TriggerAckMessage
+      expectMsg(3)
+
+      expectMsg(completeMessage)
+    }
+
+  }
+
+}
--- a/akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkStage.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkStage.scala
@@ -1,0 +1,93 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.impl
+
+import java.util
+
+import akka.actor._
+import akka.dispatch.sysmsg.{ DeathWatchNotification, SystemMessage, Watch }
+import akka.stream.stage.GraphStageLogic.StageActorRef
+import akka.stream.{ Inlet, SinkShape, ActorMaterializer, Attributes }
+import akka.stream.Attributes.InputBuffer
+import akka.stream.stage._
+
+/**
+ * INTERNAL API
+ */
+private[akka] class ActorRefBackpressureSinkStage[In](ref: ActorRef, onInitMessage: Any,
+                                                      ackMessage: Any,
+                                                      onCompleteMessage: Any,
+                                                      onFailureMessage: (Throwable) ⇒ Any)
+  extends GraphStage[SinkShape[In]] {
+  val in: Inlet[In] = Inlet[In]("ActorRefBackpressureSink.in")
+  override val shape: SinkShape[In] = SinkShape(in)
+
+  val maxBuffer = module.attributes.getAttribute(classOf[InputBuffer], InputBuffer(16, 16)).max
+  require(maxBuffer > 0, "Buffer size must be greater than 0")
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) {
+      implicit var self: StageActorRef = _
+
+      val buffer: util.Deque[In] = new util.ArrayDeque[In]()
+      var acknowledgementReceived = false
+      var completeReceived = false
+
+      override def keepGoingAfterAllPortsClosed: Boolean = true
+
+      private val callback: AsyncCallback[Unit] = getAsyncCallback((_: Unit) ⇒ {
+        if (!buffer.isEmpty) sendData()
+        else acknowledgementReceived = true
+      })
+
+      private val deathWatchCallback: AsyncCallback[Unit] =
+        getAsyncCallback((Unit) ⇒ completeStage())
+
+      private def receive(evt: (ActorRef, Any)): Unit = {
+        evt._2 match {
+          case `ackMessage`      ⇒ callback.invoke(())
+          case Terminated(`ref`) ⇒ deathWatchCallback.invoke(())
+          case _                 ⇒ //ignore all other messages
+        }
+      }
+
+      override def preStart() = {
+        self = getStageActorRef(receive)
+        self.watch(ref)
+        ref ! onInitMessage
+        pull(in)
+      }
+
+      private def sendData(): Unit = {
+        if (!buffer.isEmpty) {
+          ref ! buffer.poll()
+          acknowledgementReceived = false
+        }
+        if (buffer.isEmpty && completeReceived) finish()
+      }
+
+      private def finish(): Unit = {
+        ref ! onCompleteMessage
+        completeStage()
+      }
+
+      setHandler(in, new InHandler {
+        override def onPush(): Unit = {
+          buffer offer grab(in)
+          if (acknowledgementReceived) sendData()
+          if (buffer.size() < maxBuffer) pull(in)
+        }
+        override def onUpstreamFinish(): Unit = {
+          if (buffer.isEmpty) finish()
+          else completeReceived = true
+        }
+        override def onUpstreamFailure(ex: Throwable): Unit = {
+          ref ! onFailureMessage(ex)
+          failStage(ex)
+        }
+      })
+    }
+
+  override def toString = "ActorRefBackpressureSink"
+}
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Sink.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Sink.scala
@@ -8,6 +8,7 @@ import java.io.{ InputStream, OutputStream, File }
 import akka.actor.{ ActorRef, Props }
 import akka.dispatch.ExecutionContexts
 import akka.japi.function
+import akka.stream.impl.Stages.DefaultAttributes
 import akka.stream.impl.StreamLayout
 import akka.stream.{ javadsl, scaladsl, _ }
 import akka.util.ByteString
@@ -152,6 +153,23 @@ object Sink {
    */
   def actorRef[In](ref: ActorRef, onCompleteMessage: Any): Sink[In, Unit] =
     new Sink(scaladsl.Sink.actorRef[In](ref, onCompleteMessage))
+
+  /**
+   * Sends the elements of the stream to the given `ActorRef` that sends back back-pressure signal.
+   * First element is always `onInitMessage`, then stream is waiting for acknowledgement message
+   * `ackMessage` from the given actor which means that it is ready to process
+   * elements. It also requires `ackMessage` message after each stream element
+   * to make backpressure work.
+   *
+   * If the target actor terminates the stream will be canceled.
+   * When the stream is completed successfully the given `onCompleteMessage`
+   * will be sent to the destination actor.
+   * When the stream is completed with failure - result of `onFailureMessage(throwable)`
+   * message will be sent to the destination actor.
+   */
+  def actorRefWithAck[In](ref: ActorRef, onInitMessage: Any, ackMessage: Any, onCompleteMessage: Any,
+                          onFailureMessage: function.Function[Throwable, Any]): Sink[In, Unit] =
+    new Sink(scaladsl.Sink.actorRefWithAck[In](ref, onInitMessage, ackMessage, onCompleteMessage, onFailureMessage.apply))
 
   /**
    * Creates a `Sink` that is materialized to an [[akka.actor.ActorRef]] which points to an Actor
--- a/akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
+++ b/akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
@@ -3,23 +3,21 @@
  */
 package akka.stream.stage
 
-import java.util
-import java.util.concurrent.atomic.{ AtomicReferenceFieldUpdater, AtomicReference }
+import java.util.concurrent.atomic.AtomicReference
 
 import akka.actor._
-import akka.actor.dungeon.DeathWatch
-import akka.dispatch.sysmsg.{ Unwatch, Watch, DeathWatchNotification, SystemMessage }
-import akka.event.{ LoggingAdapter, Logging }
-import akka.event.Logging.{ Warning, Debug }
+import akka.dispatch.sysmsg.{ DeathWatchNotification, SystemMessage, Unwatch, Watch }
+import akka.event.LoggingAdapter
 import akka.stream._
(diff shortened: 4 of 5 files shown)


ktoso:
OK TO TEST

akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefBackpressureSinkSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+
+object ActorRefBackpressureSinkSpec {
+  val initMessage = "start"
+  val completeMessage = "done"
+
+  class Fw(ref: ActorRef) extends Actor {
+    def receive = {
+      case msg ⇒
+        ref forward msg
+        if (msg != completeMessage) sender ! Success("Ok")
```
patriknw: please use `sender()`

akka-stream/src/main/scala/akka/stream/javadsl/Sink.scala
```diff
@@ -114,6 +115,22 @@ object Sink {
     new Sink(scaladsl.Sink.actorRef[In](ref, onCompleteMessage))
 
   /**
+   * Sends the elements of the stream to the given `ActorRef` that sends back back-pressure signal.
+   * First element is always `onInitMessage`, then stream is waiting for acknowledgement message
+   * [[akka.actor.Status.Success]] from the given actor which means that it is ready to process
```
patriknw: perhaps make this a `Any` parameter also instead of `Status.Success`?

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * INTERNAL API
+ */
+private[akka] class ActorRefBackpressureSinkActor(ref: ActorRef, maxBuffer: Int, onInitMessage: Any,
+                                                  onCompleteMessage: Any) extends ActorSubscriber {
+  import ActorSubscriberMessage._
+  var buffer: Vector[Any] = Vector.empty
+
+  var backpressureReceived = false
```
rkuhn: please rename: what is received is not backpressure but a confirmation or acknowledgement

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
```diff
@@ -0,0 +1,68 @@
...
+    Props(new ActorRefBackpressureSinkActor(ref, maxBuffer, onInitMessage, onCompleteMessage))
+}
+
+/**
+ * INTERNAL API
+ */
+private[akka] class ActorRefBackpressureSinkActor(ref: ActorRef, maxBuffer: Int, onInitMessage: Any,
+                                                  onCompleteMessage: Any) extends ActorSubscriber {
+  import ActorSubscriberMessage._
+  var buffer: Vector[Any] = Vector.empty
```
rkuhn: using a Vector is nice but pretty wasteful: it would be a lot more efficient to use a mutable queue here, e.g. `java.util.LinkedList`

rkuhn: … or rather ArrayDeque

rkuhn:
The implementation would be smaller and also integrate better with op fusion plans if it were based upon GraphStage instead of ActorSubscriber, would you mind investigating that route?

akka-stream/src/main/scala/akka/stream/javadsl/Sink.scala
hepin1989: +1 for ack message to Any.

agolubev:
Yeah, sure. I will try GraphStage based solution. Have few concerns: we need to watch actor and receive ackMessage. But we'll see if it can be handled gracefully.

rkuhn:
Yes, you’ll need to create a custom little ActorRef by extending MinimalActorRef—we wanted to do that for other reasons as well.

drewhk:
I can try to sketch out tomorrow how to do that. We will need that to make TCP fusable as well.

agolubev:
@drewhk, I have two questions so far. What to put for MinimalActorRef.path and how to do watch for actorRef inside GraphStage without context.

agolubev:
waiting for #18677 to uncomment half of test cases

rkuhn:
I’m working on finishing that up, stay tuned.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -683,6 +683,13 @@ abstract class GraphStageLogic private[stream] (inCount: Int, outCount: Int) {
     }
   }
 
+  final def getAsyncCallback(handler: () ⇒ Unit): AsyncCallbackNoArgs = {
+    new AsyncCallbackNoArgs {
+      override def invoke(): Unit =
+        interpreter.onAsyncInput(GraphStageLogic.this, Unit, (Unit) ⇒ handler())
+    }
+  }
+
```
rkuhn: no, please don’t: this harms type inference (overloading is bad) and it is unnecessary; just send `()` as the signal

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  val maxBuffer = module.attributes.getAttribute(classOf[InputBuffer], InputBuffer(16, 16)).max
+  require(maxBuffer > 0, "Buffer size must be greater than 0")
+
+  override def createLogicAndMaterializedValue: (GraphStageLogic, Unit) = {
+
+    val buffer: util.Deque[Out] = new util.ArrayDeque[Out]()
+    var acknowledgementReceived = false
+    var completeReceived = false
+    var pullRequestIsSent = true
```
rkuhn: please put the variables inside the GraphStageLogic, doing it like this creates lots of 1-element arrays

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
```diff
@@ -0,0 +1,108 @@
...
+
+    val logic = new GraphStageLogic(shape) {
+
+      private val callback: AsyncCallbackNoArgs = getAsyncCallback(() ⇒ {
```
rkuhn: this will look like `private val callback = getAsyncCallback((_: Unit) => { ... }` instead

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
```diff
@@ -0,0 +1,108 @@
...
+        if (!buffer.isEmpty) sendData()
+        else acknowledgementReceived = true
+      })
+
+      private val deathWatchCallback: AsyncCallbackNoArgs =
```
rkuhn: I’d rather just have one callback and use the argument that it is given to distinguish between these two cases

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
```diff
@@ -0,0 +1,108 @@
...
+        getAsyncCallback(() ⇒ completeStage())
+
+      implicit val internalRefCallback = new akka.actor.MinimalActorRef {
+        override def provider: ActorRefProvider =
+          throw new UnsupportedOperationException("ActorRefBackpressureSink MinimalActor does not provide")
+        override def path: ActorPath =
+          ActorMaterializer.downcast(interpreter.materializer).supervisor.path / "ActorRefWithAckMinimalActor"
```
rkuhn: Please compute this outside of the MinimalActorRef in order to retain only the ActorPath and not the whole ActorMaterializer.

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
```diff
@@ -0,0 +1,108 @@
...
+        override def !(message: Any)(implicit sender: ActorRef = Actor.noSender): Unit = {
+          message match {
+            case msg if msg == ackMessage ⇒ callback.invoke()
```
rkuhn:
~~~scala
case `ackMessage` => callback.invoke()
~~~

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
```diff
@@ -0,0 +1,108 @@
...
+      private def sendPullIfAllowed(): Unit =
+        if (!pullRequestIsSent) {
+          pullRequestIsSent = true
+          pull(in)
+        }
+
+      setHandler(in, new InHandler {
+        override def onPush(): Unit = {
+          buffer offer grab(in)
+          pullRequestIsSent = false
```
rkuhn: this is tracked already by the GraphStage, you can use `hasBeenPulled(in)`

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
```diff
@@ -0,0 +1,108 @@
...
+      private def finish(): Unit = {
+        ref ! onCompleteMessage
+        completeStage()
+      }
+
+      private def sendPullIfAllowed(): Unit =
+        if (!pullRequestIsSent) {
+          pullRequestIsSent = true
+          pull(in)
+        }
```
rkuhn: in fact, this whole method is not needed

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
```diff
@@ -0,0 +1,108 @@
...
+
+      setHandler(in, new InHandler {
+        override def onPush(): Unit = {
+          buffer offer grab(in)
+          pullRequestIsSent = false
+          if (acknowledgementReceived) sendData()
+          if (buffer.size() < maxBuffer) sendPullIfAllowed()
```
rkuhn: within onPush you can be certain that pull is allowed, no ifs and buts :-)

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
```diff
@@ -0,0 +1,108 @@
...
+          message match {
+            case msg if msg == ackMessage ⇒ callback.invoke()
+            case _                        ⇒ //ignore all other messages
+          }
+        }
+        override def sendSystemMessage(message: SystemMessage): Unit = message match {
+          case DeathWatchNotification(_, _, _) ⇒ deathWatchCallback.invoke()
+          case _                               ⇒ //ignore all other messages
+        }
+      }
```
rkuhn: Would be great to encapsulate this in a reusable fashion inside `GraphStageLogic`, e.g. exposing it as
~~~scala
val ref = createStageActorRef(msg => /* do stuff with whatever is received, also SysMsg */)
...
ref.watch(otherRef)
~~~

akka-stream/src/main/scala/akka/stream/javadsl/Sink.scala
```diff
@@ -114,6 +115,22 @@ object Sink {
     new Sink(scaladsl.Sink.actorRef[In](ref, onCompleteMessage))
 
   /**
+   * Sends the elements of the stream to the given `ActorRef` that sends back back-pressure signal.
+   * First element is always `onInitMessage`, then stream is waiting for acknowledgement message
+   * `ackMessage` from the given actor which means that it is ready to process
+   * elements. It also requires `ackMessage` message after each stream element
+   * to make backpressure works.
```
rkuhn: `s/works/work/`

akka-stream/src/main/scala/akka/stream/javadsl/Sink.scala
```diff
@@ -114,6 +115,22 @@ object Sink {
...
+   *
+   * If the target actor terminates the stream will be canceled.
+   * When the stream is completed successfully the given `onCompleteMessage`
+   * will be sent to the destination actor.
+   * When the stream is completed with failure a [[akka.actor.Status.Failure]]
```
rkuhn: perhaps make this configurable, too? could be `onFailureMessage: (Throwable) => Any = Status.Failure`

akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefBackpressureSinkSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+import akka.stream.testkit.scaladsl._
+
+object ActorRefBackpressureSinkSpec {
+  val initMessage = "start"
+  val completeMessage = "done"
+  val ackMessage = "ack"
+
+  class Fw(ref: ActorRef) extends Actor {
+    def receive = {
+      case msg if msg == initMessage ⇒
```
rkuhn: you can either use capital letters (i.e. name it `InitMessage`) or surround by backticks

rkuhn:
Thanks for trying this out, it’s looking good. Now enable those tests and see if they work :-)

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
```diff
@@ -0,0 +1,108 @@
...
+ */
+private[akka] class ActorRefBackpressureSinkStage[Out](ref: ActorRef, onInitMessage: Any,
+                                                       ackMessage: Any,
+                                                       onCompleteMessage: Any)
+  extends SinkStage[Out, Unit]("ActorRefBackpressureSink") {
+
+  val maxBuffer = module.attributes.getAttribute(classOf[InputBuffer], InputBuffer(16, 16)).max
+  require(maxBuffer > 0, "Buffer size must be greater than 0")
+
+  override def createLogicAndMaterializedValue: (GraphStageLogic, Unit) = {
```
drewhk: If you extend GraphStage directly, then you don't need to deal with the materialized value (which is Unit).

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
drewhk: I am not sure I agree. What would that gain? The current approach gets rid of a `match` in favor of a polymorphic call.

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
```diff
@@ -0,0 +1,108 @@
...
+        else acknowledgementReceived = true
+      })
+
+      private val deathWatchCallback: AsyncCallbackNoArgs =
+        getAsyncCallback(() ⇒ completeStage())
+
+      implicit val internalRefCallback = new akka.actor.MinimalActorRef {
+        override def provider: ActorRefProvider =
+          throw new UnsupportedOperationException("ActorRefBackpressureSink MinimalActor does not provide")
+        override def path: ActorPath =
```
drewhk: shouldn't this be a val?

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
drewhk: I would also make dependencies explicit by defining this class outside of the logic and taking the deps (Materializer, callbacks, etc.) as constructor arguments.

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
```diff
@@ -0,0 +1,108 @@
...
+          ActorMaterializer.downcast(interpreter.materializer).supervisor.path / "ActorRefWithAckMinimalActor"
+        override def !(message: Any)(implicit sender: ActorRef = Actor.noSender): Unit = {
+          message match {
+            case msg if msg == ackMessage ⇒ callback.invoke()
+            case _                        ⇒ //ignore all other messages
+          }
+        }
+        override def sendSystemMessage(message: SystemMessage): Unit = message match {
+          case DeathWatchNotification(_, _, _) ⇒ deathWatchCallback.invoke()
```
drewhk: maybe match also on the watched ref, just to be on the safe side?

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
drewhk: +1, we will need this for TCP

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: +1

drewhk:
Looks great! Just a few steps away from merge:
 - fix the small issues
 - make the ref creation code accessible from GraphStage (maybe as an INTERNAL API first)
 - rebase and squash

agolubev:
Thank you both for great review.

agolubev:
seems like it's still an issue with AsyncCallback 
Created PR with test case #18835

drewhk:
@rkuhn fixed quite a bit of bugs in the interpreter and they were merged during the weekend. Can you try rebasing your PR on the latest changes? Maybe your issue has been solved already.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefBackpressureSinkSpec.scala
drewhk: I.e. please use 
    
    case `initMessage` =>

or capitalize the variable name and then you can use

    case InitMessage =>

agolubev:
@drewhk, I did rebase.

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkActor.scala
agolubev: it's allowed, I agree. But we cannot request more than buffer size (as waiting for actorRef acknowledgment)

drewhk:
@agolubev I guess this needs to wait until after M1 because solving #18835 is not exactly trivial and we don't want to delay the milestone. This will be one of the first things we will do though.

agolubev:
@drewhk, yeah, it's not a problem. Still, let me finish changes per reviews

agolubev:
All done for PR. 
Waiting for #18835 to merge this in

drewhk:
This depends on me to provide the necessary feature.

drewhk:
@agolubev I reused your ActorRef code, changed slightly here: https://github.com/akka/akka/pull/18918

agolubev:
That's cool :)

agolubev:
@drewhk, fix for #18835 - works for this ticket. Thanks!
I rebased and squashed my changes. Deleted pending tests as well.

drewhk:
Thanks, though I will need to modify the StageActorRef feature. So this still needs to wait :(

agolubev:
@drewhk, can you give ticket number for StageActorRef feature?

rkuhn:
This LGTM: please rebase on current release-2.3-dev and use the StageActorRef that is in there, then we can merge this.

agolubev:
build failed with
```
[info] groupBy
[info] - must work in normal user scenario *** FAILED *** (53 milliseconds)
[info]   Vector(Vector("Bcc"), Vector("Cdd", "Cee"), Vector("Aaa", "Abb")) did not equal List(List("Aaa", "Abb"), List("Bcc"), List("Cdd", "Cee")) (FlowGroupBySpec.scala:120)
```
@rkuhn, PR ready for review/merge

rkuhn:
Thanks again, @agolubev !

agolubev:
Great! Thanks

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkStage.scala
```diff
@@ -0,0 +1,93 @@
...
+  override val shape: SinkShape[In] = SinkShape(in)
+
+  val maxBuffer = module.attributes.getAttribute(classOf[InputBuffer], InputBuffer(16, 16)).max
+  require(maxBuffer > 0, "Buffer size must be greater than 0")
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) {
+      implicit var self: StageActorRef = _
+
+      val buffer: util.Deque[In] = new util.ArrayDeque[In]()
```
ktoso: this is JDK7 API, can't use it in streams yet :warning:

ktoso: Oh it's not... mis remembered things; nevermind

akka-stream-tests/src/test/scala/akka/stream/scaladsl/ActorRefBackpressureSinkSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+}
+
+class ActorRefBackpressureSinkSpec extends AkkaSpec {
+  import ActorRefBackpressureSinkSpec._
+  implicit val mat = ActorMaterializer()
+
+  def createActor[T](c: Class[T]) =
+    system.actorOf(Props(c, testActor).withDispatcher("akka.test.stream-dispatcher"))
+
+  "An ActorRefSink" must {
```
ktoso: `An ActorRefSinkWithAck`?

ktoso: Or more like the impl class `ActorRefBackpressureSink`

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkStage.scala
```diff
@@ -0,0 +1,93 @@
...
+      var acknowledgementReceived = false
+      var completeReceived = false
+
+      override def keepGoingAfterAllPortsClosed: Boolean = true
+
+      private val callback: AsyncCallback[Unit] = getAsyncCallback((_: Unit) ⇒ {
+        if (!buffer.isEmpty) sendData()
+        else acknowledgementReceived = true
+      })
```
ktoso: Is getting the async callback before preStart always safe now? I may be mis-remembering that it's not. // @rkuhn

ktoso:
Was late with the review but still wanted to go through :)
Looks very good, thanks @agolubev!

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkStage.scala
rkuhn: Obtaining the callback is safe, sharing it with the outside world is not.

But this made me look at this again: why do we dispatch asynchronously from the StageActorRef again? We should be able to just do The Things from within `receive`, no?

agolubev: that's correct.  getStageActorRef makes ```receive``` safe for graphStage state change. I will remove AsyncCallback

agolubev:
changes per post merge review are here: #19163

Issue #18687, opened 2015-10-09
add filterNot combinator


rkuhn:

hepin1989:
this should be closed now.

ktoso:
Thanks!

PR #18775, opened 2015-10-25
+str #17967 add filterNot combinator

agolubev:
Apologies, it's #18687. 
This is sugar I believe, so perhaps we don't need to add this to keep API simple.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
@@ -492,6 +492,20 @@ trait FlowOps[+Out, +Mat] {
   def filter(p: Out ⇒ Boolean): Repr[Out, Mat] = andThen(Filter(p.asInstanceOf[Any ⇒ Boolean]))
 
   /**
+   * Only pass on those elements that NOT satisfy the given predicate.
+   *
+   * '''Emits when''' the given predicate returns false for the element
+   *
+   * '''Backpressures when''' the given predicate returns false for the element and downstream backpressures
+   *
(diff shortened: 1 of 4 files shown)


ktoso:
OK TO TEST

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -492,6 +492,19 @@ trait FlowOps[+Out, +Mat] {
...
+   *
+   * '''Emits when''' the given predicate returns false for the element
+   *
+   * '''Backpressures when''' the given predicate returns false for the element and downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def filterNot(p: Out ⇒ Boolean): Repr[Out, Mat] = filter(!p(_))
```
hepin1989: This will shows up with ```filter``` whn debug,not ```filterNot```.

drewhk: good point!

drewhk: `via(Flow[Out].filter(!p(_)).withAttributes(...))` will fix it without the need for a new stage

ktoso: @drewhk :+1:

rkuhn:
LGTM after the proposed improvement. Thanks @agolubev !

drewhk:
LGTM

Issue #18766, opened 2015-10-23
Transparent exponential back off supervisor


henrymai:
The existing back off supervisor (`akka.pattern.BackoffSupervisor`) requires the child to be stopped in order for the supervisor to restart the child. This seems inconsistent with the behavior of normal supervision, where the child will be automatically restarted upon an exception occurring (subject to the decider on the SupervisorStrategy).

This requires the child actor being supervised to be aware of this semantic, which means that it has to catch any exceptions that occur and cause itself to be terminated in order for it to be restarted.
This seems cumbersome and loses some of the niceties provided by normal supervision such as automatic logging of exceptions.

I ended up creating my own backoff supervisor in order to get the behavior mentioned above and I wanted to check in with you guys to see if this is useful:
https://github.com/henrymai/akka-transparent-exponential-backoff-supervisor

patriknw:
This looks like a duplicate of https://github.com/akka/akka/issues/18487

patriknw:
closing as duplicate, you are welcome to contribute with a pull request

patriknw:
I'm opening this again, since it seems that it should not be aligned with existing BackoffSupervisor.

henrymai:
https://github.com/akka/akka/pull/18776 has been merged. Closing this issue.

Issue #18487, opened 2015-09-16
Provide Supervisor(strategy, childProps) to ease Sharding entity supervision


ktoso:
We should provide a `Supervisor(strategy, childProps)` that works similar to `BackoffSupervisor` however applies the strategy right away. It is useful when you need to configure supervision, yet "someone else is starting the actor for you, so you need to introduce this Supervisor parent", a common example being Cluster Sharding.

More details here: https://github.com/akka/akka/pull/18445#issuecomment-140698857

patriknw:
at the same time we can also make `BackoffSupervisor` a bit more flexible by making it possible to install a strategy into that also, so that exceptions can result in stop-backoff-start instead of immediate restart

patriknw:
See also https://github.com/henrymai/akka-transparent-exponential-backoff-supervisor that was shared via #18766

masterzen:
:+1: for adding a possibility to add a supervision strategy to a cluster sharding entity.

Still, one of the issue with `BackoffSupervisor` (and other parent-masquerading-child patterns) with respect to Cluster Sharding is that it prevents using the Passivate mechanism.

Because the entity running as the child actor of the `BackoffSupervisor` will send its `Passivate` message, instead to the `Shard`, resulting in the entity receiving back the `Passivate`. The only solution is to infer the path to the `Shard` from its own entity path and use an `ActorSelection` or create a specific `BackoffSupervisor` that knows how to forward the `Passivate` message upstream.

sdab75:
We tried testing the supervision functionality as described in the 2.4.0 documentation and in fact we ried running partiknw example and we don't see exception getting propagated to the supervisor actor. My project with few modifications checked-in at  this location (https://github.com/sdab75/akka-cluster-sharding-test/blob/master/src/main/java/akka/cluster/sharding/ClusterShardingTest.java). Since it was documented we thought this working. Can some one help us understand if this fix was merged in the 2.4.0 branch or not.

PR #18776, opened 2015-10-25
Add TransparentExponentialBackoffSupervisor

henrymai:
See this link for details:
https://github.com/akka/akka/issues/18766

### DIFF
--- a/akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
+++ b/akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
@@ -61,6 +61,26 @@ object BackoffSupervisor {
 
   private case object StartChild extends DeadLetterSuppression
   private case class ResetRestartCount(current: Int) extends DeadLetterSuppression
+
+  /**
+   * INTERNAL API
+   *
+   * Calculates an exponential back off delay.
+   */
+  private[akka] def calculateDelay(
+    restartCount: Int,
+    minBackoff: FiniteDuration,
+    maxBackoff: FiniteDuration,
+    randomFactor: Double): FiniteDuration = {
+    val rnd = 1.0 + ThreadLocalRandom.current().nextDouble() * randomFactor
+    if (restartCount >= 30) // Duration overflow protection (> 100 years)
+      maxBackoff
+    else
+      maxBackoff.min(minBackoff * math.pow(2, restartCount)) * rnd match {
+        case f: FiniteDuration ⇒ f
+        case _                 ⇒ maxBackoff
+      }
+  }
 }
 
 /**
@@ -121,15 +141,7 @@ final class BackoffSupervisor(
   def receive = {
     case Terminated(ref) if child.contains(ref) ⇒
       child = None
-      val rnd = 1.0 + ThreadLocalRandom.current().nextDouble() * randomFactor
-      val restartDelay =
-        if (restartCount >= 30) // Duration overflow protection (> 100 years)
-          maxBackoff
-        else
-          maxBackoff.min(minBackoff * math.pow(2, restartCount)) * rnd match {
-            case f: FiniteDuration ⇒ f
-            case _                 ⇒ maxBackoff
-          }
+      val restartDelay = calculateDelay(restartCount, minBackoff, maxBackoff, randomFactor)
       context.system.scheduler.scheduleOnce(restartDelay, self, StartChild)
       restartCount += 1
 
--- a/akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
+++ b/akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
@@ -1,0 +1,202 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.contrib.pattern
+
+import akka.actor._
+import akka.actor.OneForOneStrategy
+import akka.actor.SupervisorStrategy._
+import scala.concurrent.duration._
+
+object TransparentExponentialBackoffSupervisor {
+  private case class ScheduleRestart(childRef: ActorRef) extends DeadLetterSuppression
+  private case object StartChild extends DeadLetterSuppression
+  private case class ResetRestartCount(lastNumRestarts: Int) extends DeadLetterSuppression
+
+  /**
+   * Props for creating a [[TransparentExponentialBackoffSupervisor]] with a decider.
+   *
+   * @param childProps the [[akka.actor.Props]] of the child to be supervised.
+   * @param childName the name of the child actor.
+   * @param minBackoff the min time before the child is restarted.
+   * @param maxBackoff the max time (upperbound) for a child restart.
+   * @param randomFactor a random delay factor to add on top of the calculated exponential
+   *   back off.
+   *   The calculation is equivalent to:
+   *   {{{
+   *     final_delay = min(
+   *       maxBackoff,
+   *       (random_delay_factor * calculated_backoff) + calculated_backoff)
+   *   }}}
+   * @param decider an [[akka.actor.SupervisorStrategy.Decider]] to specify how the supervisor
+   *   should behave for different exceptions. If no cases are matched, the default decider of
+   *   [[akka.actor.Actor]] is used. When the [[akka.actor.SupervisorStrategy.Restart]] directive
+   *   is returned by the decider, this supervisor will apply an exponential back off restart.
+   */
+  def propsWithDecider(
+    childProps: Props,
+    childName: String,
+    minBackoff: FiniteDuration,
+    maxBackoff: FiniteDuration,
+    randomFactor: Double)(decider: Decider): Props = {
+    Props(
+      new TransparentExponentialBackoffSupervisor(
+        childProps,
+        childName,
+        Some(decider),
+        minBackoff,
+        maxBackoff,
+        randomFactor))
+  }
+
+  /**
+   * Props for creating a [[TransparentExponentialBackoffSupervisor]] using the
+   * default [[akka.actor.Actor]] decider.
+   *
+   * @param childProps the [[akka.actor.Props]] of the child to be supervised.
+   * @param childName the name of the child actor.
+   * @param minBackoff the min time before the child is restarted.
+   * @param maxBackoff the max time (upperbound) for a child restart.
+   * @param randomFactor a random delay factor to add on top of the calculated exponential
+   *   back off.
+   *   The calculation is equivalent to:
+   *   {{{
+   *     final_delay = min(
+   *       maxBackoff,
+   *       (random_delay_factor * calculated_backoff) + calculated_backoff)
+   *   }}}
+   */
+  def props(
+    childProps: Props,
+    childName: String,
+    minBackoff: FiniteDuration,
+    maxBackoff: FiniteDuration,
+    randomFactor: Double): Props = {
+    Props(
+      new TransparentExponentialBackoffSupervisor(
+        childProps,
+        childName,
+        None,
+        minBackoff,
+        maxBackoff,
+        randomFactor))
+  }
+}
+
+/**
+ * A supervising actor that restarts a child actor with an exponential back off.
+ *
+ * This explicit supervisor behaves similarly to the normal implicit supervision where
+ * if an actor throws an exception, the decider on the supervisor will decide when to
+ * `Stop`, `Restart`, `Escalate`, `Resume` the child actor.
+ *
+ * When the `Restart` directive is specified, the supervisor will delay the restart
+ * using an exponential back off strategy (bounded by minBackoff and maxBackoff).
+ *
+ * This supervisor is intended to be transparent to both the child actor and external actors.
+ * Where external actors can send messages to the supervisor as if it was the child and the
+ * messages will be forwarded. And when the child is `Terminated`, the supervisor is also
+ * `Terminated`.
+ * Transparent to the child means that the child does not have to be aware that it is being
+ * supervised specifically by the [[TransparentExponentialBackoffSupervisor]]. Just like it does
+ * not need to know when it is being supervised by the usual implicit supervisors.
+ * The only caveat is that the `ActorRef` of the child is not stable, so any user storing the
+ * `sender()` `ActorRef` from the child response may eventually not be able to communicate with
+ * the stored `ActorRef`. In general all messages to the child should be directed through the
+ * [[TransparentExponentialBackoffSupervisor]].
+ *
+ * An example of where this supervisor might be used is when you may have an actor that is
+ * responsible for continuously polling on a server for some resource that sometimes may be down.
+ * Instead of hammering the server continuously when the resource is unavailable, the actor will
+ * be restarted with an exponentially increasing back off until the resource is available again.
+ *
+ * '''***
+ * This supervisor should not be used with `Akka Persistence` child actors.
+ * `Akka Persistence` actors, currently, shutdown unconditionally on `persistFailure()`s rather
+ * than throw an exception on a failure like normal actors.
+ * [[akka.pattern.BackoffSupervisor]] should be used instead for cases where the child actor
+ * terminates itself as a failure signal instead of the normal behavior of throwing an exception.
+ * ***'''
+ */
+class TransparentExponentialBackoffSupervisor(
+  props: Props,
+  childName: String,
+  decider: Option[Decider],
+  minBackoff: FiniteDuration,
+  maxBackoff: FiniteDuration,
+  randomFactor: Double)
+  extends Actor
+  with Stash
+  with ActorLogging {
+
+  import TransparentExponentialBackoffSupervisor._
+  import context._
+
+  override val supervisorStrategy = OneForOneStrategy() {
+    case ex ⇒
+      val defaultDirective: Directive =
+        super.supervisorStrategy.decider.applyOrElse(ex, (_: Any) ⇒ Escalate)
+      val maybeDirective: Option[Directive] = decider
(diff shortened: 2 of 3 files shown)


patriknw:
Can this be worked in to the `BackoffSupervisor` in a backwards compatible way? I don't think we should provide two classes for almost the same thing.

patriknw:
Refs #18487

henrymai:
I don't think there's a straight forward way to work my code into the `BackoffSupervisor` because it works in a fundamentally different way than `BackoffSupervisor`. Mine relies on the normal implicit supervision behavior where it uses a `SupervisorStrategy.Decider` to decide when to restart a child actor, whereas the existing `BackoffSupervisor` relies on the child being terminated in order to know to restart the child.

There might be a way for me to do it the other way around, where I can provide a different props method that will pass a flag into the constructor, so that the actor can mimic old behavior as part of my supervisor. However, I don't see much value in doing so, because the old behavior is strictly less desirable than the new behavior. It would entail adding extra cruft into my implementation to support a behavior that is unwanted. Additionally the message interface that I expose is also different, in that I do not want to allow explicit retrieval of the child `ActorRef`.

I would prefer to just to keep these two separate and deprecate the existing `BackoffSupervisor` over time instead of attempting to merge the two together. It is cleaner this way and won't break any existing users of `BackoffSupervisor`.

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
```diff
@@ -0,0 +1,206 @@
...
+        .map(_.applyOrElse(ex, (_: Any) ⇒ defaultDirective))
+
+      // Get the directive from the specified decider or fallback to
+      // the default decider.
+      // Whatever the final Directive is, we will translate all Restarts
+      // to our own Restarts, which involves stopping the child.
+      maybeDirective.getOrElse(defaultDirective) match {
+        case Restart ⇒
+          self ! RestartChild
+          Stop
```
rkuhn: This Stop may overtake the RestartChild in the mailbox in the presence of other messages, leading to the termination of this supervisor actor.

rkuhn:
I agree with @henrymai, this implementation approach is superior to the existing one.

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
```diff
@@ -0,0 +1,206 @@
...
+
+  // Initialize by starting up and watching the child
+  self ! RestartChild
+
+  def receive = {
+    case RestartChild ⇒
+      val childRef = actorOf(props)
+      watch(childRef)
+      unstashAll()
+      become(watching(childRef, 0))
```
ktoso: Nice, I like this impl.

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
```diff
@@ -0,0 +1,206 @@
...
+      stop(self)
+    case Terminated(other) ⇒
+      log.debug(s"Ignoring unexpected Terminated message from: $other")
+    case msg ⇒
+      childRef.forward(msg)
+  }
+
+  def calculateDelay(numRestarts: Int): FiniteDuration = {
+    // Source:
+    // https://github.com/akka/akka/blob/v2.4.0/akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
```
ktoso: Seems we'll need to keep both impls for a bit, lets put the back off calculation in one spot then somehow?

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
henrymai: Yea, I was afraid that might be the case.
But thankfully, I actually had an alternate implementation before that did a `Resume` here instead of a `Stop` (still sending a `RestartChild` message) and then in the `RestartChild` handler, it would terminate the child before the become(waitingToRestart(childRef, numRestarts)).
I'll switch it to this implementation later tonight, but I was wondering if you could elaborate a little more on why the `Stop` can overtake the `RestartChild` message in the presence of other messages (just for my own edification).

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
henrymai: Sure, I can refactor the existing impl in the `BackoffSupervisor` into a function in the companion object and make my supervisor make use of that function.

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
rkuhn: Actually, thinking about it, we might accidentally ensure that this cannot happen: Stop means a Terminate sysmsg to the child actor, which triggers a ChildTerminated sysmsg back to us, but processing that will just enqueue Terminated as a normal message at the back of the mailbox behind the RestartChild message—this is done so that any message from the child actor that was emitted before termination is also processed before the Terminated message and it saves the day here.

Still, it would be more obvious if you just switched the behavior to `waitingToRestart` right here and got rid of the RestartChild message altogether—initial creation can just happen in `preStart`.

henrymai: > but processing that will just enqueue Terminated as a normal message at the back of the mailbox behind the RestartChild message
Right, that's what I thought from reading the akka code.

> Still, it would be more obvious if you just switched the behavior to waitingToRestart right here and got rid of the RestartChild message altogether—initial creation can just happen in preStart.

The problem with switching the behavior at that location is that I don't have access to numRestarts at that point (unless I make it a var). It is my preference to not use vars whenever I can avoid it.

rkuhn: This—in a nutshell—is the reason for making all lifecycle events (like child failure) normal messages in Akka Typed. I’d probably use a `var` here.

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
ktoso: Yup, exactly

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
henrymai: I just updated the patch to address this concern. Still managed to avoid using a var with the same lines of code :)

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
henrymai: Just updated the patch to address this concern.

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
```diff
@@ -61,6 +61,21 @@ object BackoffSupervisor {
 
   private case object StartChild extends DeadLetterSuppression
   private case class ResetRestartCount(current: Int) extends DeadLetterSuppression
+
+  def calculateDelay(
```
patriknw: this should be `private[akka]` and marked as INTERNAL API

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
```diff
@@ -0,0 +1,192 @@
...
+
+import akka.actor._
+import akka.actor.OneForOneStrategy
+import akka.actor.SupervisorStrategy._
+import scala.concurrent.duration._
+
+object TransparentExponentialBackoffSupervisor {
+  private case class ScheduleRestart(childRef: ActorRef)
+  private case object StartChild
+  private case class ResetRestartCount(lastNumRestarts: Int)
```
patriknw: DeadLetterSuppression on these?

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
```diff
@@ -0,0 +1,192 @@
...
+   *   [[akka.actor.Actor]] is used. When the [[akka.actor.SupervisorStrategy.Restart]] directive
+   *   is returned by the decider, this supervisor will apply an exponential back off restart.
+   */
+  def propsWithDecider(
+    childProps: Props,
+    minBackoff: FiniteDuration,
+    maxBackoff: FiniteDuration,
+    randomFactor: Double)(decider: Decider): Props = {
+    Props(
+      classOf[TransparentExponentialBackoffSupervisor],
```
patriknw: use `new TransparentExponentialBackoffSupervisor(...`

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
```diff
@@ -0,0 +1,192 @@
...
+ * if an actor throws an exception, the decider on the supervisor will decide when to
+ * `Stop`, `Restart`, `Escalate`, `Resume` the child actor.
+ *
+ * When the `Restart` directive is specified, the supervisor will delay the restart
+ * using an exponential back off strategy (bounded by minBackoff and maxBackoff).
+ *
+ * This supervisor is intended to be transparent to both the child actor and external actors.
+ * Where external actors can send messages to the supervisor as if it was the child and the
+ * messages will be forwarded. And when the child is `Terminated`, the supervisor is also
+ * `Terminated`.
```
patriknw: why can't we use the same semantics as the old `BackoffSupervisor` so that this can be used with Akka Persistence, i.e. when the child actor is terminated exponential back-off kicks in?

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
```diff
@@ -0,0 +1,192 @@
...
+  maxBackoff: FiniteDuration,
+  randomFactor: Double)
+  extends Actor
+  with Stash
+  with ActorLogging {
+
+  import TransparentExponentialBackoffSupervisor._
+  import context._
+
+  override val supervisorStrategy = OneForOneStrategy(maxNrOfRetries = -1) {
```
patriknw: skip `maxNrOfRetries` parameter?

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
rkuhn: @henrymai has a good point there: either termination is interpreted as a failure signal (which it normally is not, apart from BackoffSupervisor) or it is just that—termination (as it is here). I think these two are different enough to warrant different treatment.

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
```diff
@@ -0,0 +1,192 @@
...
+      // Get the directive from the specified decider or fallback to
+      // the default decider.
+      // Whatever the final Directive is, we will translate all Restarts
+      // to our own Restarts, which involves stopping the child.
+      maybeDirective.getOrElse(defaultDirective) match {
+        case Restart ⇒
+          val childRef = sender
+          become({
+            case Terminated(`childRef`) ⇒
+              unwatch(childRef)
```
patriknw: no need to unwatch when `Terminated` has been received (that is done automatically)

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
```diff
@@ -0,0 +1,192 @@
...
+              unbecome()
+              self ! ScheduleRestart(childRef)
+            case Terminated(other) ⇒
+              log.debug(s"Received unexpected Terminated message from: $other")
```
patriknw: how can this happen? better to crash if it is not expected?

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
```diff
@@ -0,0 +1,192 @@
...
+      } else {
+        log.debug(s"Last restart count [$last] does not match the current count [$numRestarts]")
+      }
+    case Terminated(`childRef`) ⇒
+      log.debug(s"Terminating, because child [$childRef] terminated itself")
+      stop(self)
+    case Terminated(other) ⇒
+      log.debug(s"Ignoring unexpected Terminated message from: $other")
+    case msg ⇒
+      childRef.forward(msg)
```
patriknw: another thing that was pointed out in the ticket is that we should support the `Passivate` message for cluster sharding. That is sent to the parent. I think we can forward all messages from the child (`sender()`) to the parent.

patriknw:
Can't we try harder to align the two backoff supervisors? I think it will be confusing for the users with two slightly different implementations. This implementation is not compatible with Akka Persistence (which must unconditionally stops the actor when there is a journal failure).

One idea is that `Restart` means ordinary restart and `Stop` means stop-backoff-start. Then we don't change semantics of restart and the needed semantics for Akka Persistence comes naturally.

henrymai:
Hi Patrik,
Thanks for the review.

> I think it will be confusing for the users with two slightly different implementations.

Can we solve that through documentation?

> This implementation is not compatible with Akka Persistence (which must unconditionally stops the actor when there is a journal failure).

My implementation is targeting the case of standard actor and supervision semantics. Is there a reason why Akka Persistence went the non standard route of shutting down the child rather than throw a specific `Exception`/`Throwable`; allowing a supervising actor to decide to `Stop` for that specific exception?

> One idea is that Restart means ordinary restart and Stop means stop-backoff-start. Then we don't change semantics of restart and the needed semantics for Akka Persistence comes naturally.

Unless I'm mistaken, `persistFailure()` will issue a `context.stop(self)`, so it won't actually end up going through the SupervisorStrategy.decider. Meaning even if we decided to change the semantics of this patch to perform a "stop-backoff-start" for the `Stop` directive, we won't actually end up going through that path for Akka Persistence.

Is it not possible to just allow existing Akka Persistence users to continue using the `BackoffSupervisor` and explicitly document that the `TransparentExponentialBackoffSupervisor` is not suitable for usage with Akka Persistence?

rkuhn:
As I said, @henrymai has a valid point here, the signaling mechanism is different so I don’t see how or why we should unify these. In particular I don’t like that self-stopping is part of the usage contract for BackoffSupervisor—that might work for specific cases (in particular “immortal” actors that come back after termination) but it is not the best general solution.

patriknw:
@henrymai the reason persistent actors do `context.stop(self)` is that we can't enforce that users have installed a proper supervision strategy that performs the stopping (and stopping is unconditional as I mentioned earlier).

Ok, I'm in minority in thinking it will be confusing. Please work on the naming and documentation to make it clear then that there are two different utilities for this.

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
henrymai: Addressed in the latest patch

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
henrymai: Doesn't actually ever happen. Addressed in the latest patch.

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
henrymai: Addressed in the latest patch

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
henrymai: Addressed in the latest patch

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
henrymai: Addressed in the latest patch

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
henrymai: Addressed in the latest patch

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
henrymai: Addressed in the latest patch

henrymai:
@rkuhn @patriknw is there anything else that needs to be done for this to be merged?

rkuhn:
Thanks for the contribution, @henrymai ! (and sorry for the long review delay)

PR #18777, opened 2015-10-26
=str - Minor touchups to GraphInterpreter and GraphAssembly

viktorklang:


akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -85,13 +88,14 @@ private[stream] object GraphInterpreter {
    * corresponding segments of these arrays matches the exact same order of the ports in the [[Shape]].
    *
    */
-  final case class GraphAssembly(stages: Array[GraphStageWithMaterializedValue[_, _]],
+  final case class GraphAssembly(stages: Array[GraphStageWithMaterializedValue[_, _]], // FIXME why is this a case class?
```
drewhk: no need to be a case class at all.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -24,6 +26,7 @@ private[stream] object GraphInterpreter {
   final val NoEvent = -1
   final val Boundary = -1
 
+  final val Initialized = 0
```
drewhk: actually all ports should be initialized to `InReady` so this is more like `Uninitialized`. Btw, do you use this value somewhere?

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -175,6 +179,43 @@ private[stream] object GraphInterpreter {
...
+
+  object GraphAssembly {
+    /**
+     * INTERNAL API
+     */
+    final def apply(inlets: immutable.Seq[Inlet[_]],
+                    outlets: immutable.Seq[Outlet[_]],
+                    stages: GraphStageWithMaterializedValue[_, _]*): GraphAssembly = {
+      // add the contents of an iterator to an array starting at idx
+      @tailrec def add[T](i: Iterator[T], a: Array[T], idx: Int): Array[T] =
```
drewhk: A while loop instead?

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -175,6 +179,43 @@ private[stream] object GraphInterpreter {
         outOwners.mkString("[", ",", "]") +
         ")"
   }
+
+  object GraphAssembly {
+    /**
+     * INTERNAL API
+     */
+    final def apply(inlets: immutable.Seq[Inlet[_]],
```
drewhk: useful one.

drewhk:
LGTM. Please don't merge before my PR though.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
viktorklang: Good point, this is a remnant, I'll remove it.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
viktorklang: Perfect, I'll fix.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
viktorklang: It's the same LoC, but yeah, I'll switch

drewhk: I used while loops everywhere else (since everything is mutated anyway) so it will be more consistent. I don't have a strong opinion though.

viktorklang: Nah, retracting that, the tailrec is cleaner

viktorklang: Keeping mutability to a minimum is not optional ;)

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -85,13 +87,14 @@ private[stream] object GraphInterpreter {
    * corresponding segments of these arrays matches the exact same order of the ports in the [[Shape]].
    *
    */
-  final case class GraphAssembly(stages: Array[GraphStageWithMaterializedValue[_, _]],
-                                 ins: Array[Inlet[_]],
-                                 inOwners: Array[Int],
-                                 outs: Array[Outlet[_]],
-                                 outOwners: Array[Int]) {
+  final class GraphAssembly(val stages: Array[GraphStageWithMaterializedValue[_, _]],
```
viktorklang: @drewhk Ok, made it a non-case class

viktorklang:
OK, this is "done" from my PoV.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: unless if it is performance critical and internal ;)

viktorklang: @drewhk Absolutely. However, tailrec does not carry any overhead over while ;-)

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -512,28 +555,28 @@ private[stream] final class GraphInterpreter(
...
-    portStates(connection) = portStates(connection) | OutClosed
-    if ((currentState & InClosed) == 0) {
-      if ((currentState & Pushing) != 0) {} // FIXME: Fold into previous condition
-      else if (connectionSlots(connection) != Empty)
-        enqueue(connection)
-      else
-        enqueue(connection)
+    portStates(connection) = currentState | OutClosed
+
+    if ((currentState & (InClosed | Pushing)) == 0) {
```
viktorklang: @drewhk I think this is way easier on the eye :)

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: I don't fully trust the scala compiler ;) I also find while sligthly more easier to "get" when arrays are involved.

viktorklang: It's sad that arrays do not have methods to copy in stuff, so you wouldn't have to write these methods in the first place :S

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -175,6 +178,42 @@ private[stream] object GraphInterpreter {
...
+      // add the contents of an iterator to an array starting at idx
+      @tailrec def add[T](i: Iterator[T], a: Array[T], idx: Int): Array[T] =
+        if (i.hasNext) {
+          a(idx) = i.next()
+          add(i, a, idx + 1)
+        } else a
+
+      // fill array slots with Boundary
+      def markBoundary(owners: Array[Int], from: Int, to: Int): Array[Int] = {
+        Arrays.fill(owners, from, to, Boundary)
```
hepin1989: System.ArrayCopy?

viktorklang: Arraycopy is from copying *between* arrays

hepin1989: Oh,I see,you are fill the owners with Boundary:)

viktorklang: Yes, sadly Arrays.fill does not return the Array, otherwise this method would not really be needed.

rkuhn:
Thanks!

Issue #18674, opened 2015-10-08
remove FlowGraph.Builder.edge (and friends)


rkuhn:
One DSL to rule them all!

rkuhn:
earlier parts done in #18700

PR #18779, opened 2015-10-26
!str #18674 remove FlowGraph.Builder.add(Sink|Source)

rkuhn:


ktoso:
LGTM, fits in in making the api surface area smaller I guess.

Refs #18674

Issue #16813, opened 2015-02-10
Http, client: Provide high-level support for response decompression


jrudolph:
See spray/spray#987

In spray, you could add a `decode(...)` clause to your pipeline to add support for a single decoder. As noted in the ticket it would make sense to provide also an automatic solution that tries all known decodes.

/cc @sirthias

RichardBradley:
In case it helps anyone, I am using the following code which works around this issue by trying all known coders and decoders:

```
 /**
   * This method takes a route and wraps it to support request/response compression.
   */
  def handleCompression(route: Route): Route = {
    decompressRequest(Gzip, Deflate, NoCoding) {
      compressResponse(NoCoding, Gzip, Deflate) {
        route
      }
    }
  }
```

See also #16998

2beaucoup:
This ticket discussed generic *client side* response decompression so I think it's only remotely related to your referenced issue.

RichardBradley:
Sorry, I misunderstood the issue here. There is a parallel problem for server-side handling of decompression of requests and compression of responses.

13h3r:
+1

13h3r:
I am start looking how can I implement this.
There are few questions I want to ask:
1. In which abstraction should we implement decoding? IMO we should make it like a Flow[Response, Response]. This flow should check headers and replace entity in response if needed.
2. At which level we should put this decoding? Sure we should not push this into Connection Level API. Sure we should make this available in Host Level API. What do you think about Request Level API? Should we push it into singleRequest for example?
3. For the first round I want just implement decoding without tests and submit this code for feedback. Is it ok?

sirthias:
@13h3r I think it makes sense to approach this together with request encoding.
For example, you could extend the existing `akka.http.scaladsl.client.RequestBuilding` trait with something like

```scala
def encodeRequest(coder: Coder): RequestTransformer
```

Then you could add a `ResponseTransformation` trait (and object) containing something like

```scala
def decodeResponse: ResponseTransformer
```

Making the decoding logic available as a pipeline stage (i.e. `Flow[HttpResponse, HttpResponse]`), as you say, would probably also make sense.

2beaucoup:
@sirthias Speaking of `RequestBuilding`: Does it have a future? It's currently used in tests only and I wonder if another way of doing things (`RequestTransformer`/`ResponseTransformer` sound like stream abstractions) won't confuse users. What's the advantage compared to transforming/wrapping the `outgoingConnection` flow?

[...]

PR #18783, opened 2015-10-26
Naive implementation of client negotiable decoder

13h3r:
Related issue #16813. This PR is for feedback.
## Implementation

I am not very familiar with HTTP details and I want to confirm that idea of universal negotiable decoder is viable and correct
## API

I am not sure about creating `ResponseBuilding` as analog of `RequestBuilding` (suggested [here](https://github.com/akka/akka/issues/16813#issuecomment-150221382)). My idea is to provide universal building block in terms of functions and flows. Should we push this decoder into _Request Level API_ flow?
## Tests

I really have no idea what to test in this implementation. All tests looks like an implementation duplication.  Only viable idea is to build whole pipeline of response parsing and decoding.

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/coding/NegotiableDecoder.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/coding/NegotiableDecoder.scala
@@ -1,0 +1,28 @@
+/**
+  * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+  */
+package akka.http.scaladsl.coding
+
+import akka.http.scaladsl.model.HttpResponse
+import akka.http.scaladsl.model.headers.{HttpEncoding, HttpEncodings}
+import akka.stream.scaladsl.Flow
+
+/**
+  * Provides high level decoding for HTTP client by analyzing Content-Encoding header
+  * and try to find appropriate decoder. Gzip and Deflate decoders are supported.
+  * In case of unknown encoding returns response as is.
+  */
+object NegotiableDecoder {
+  private def findCodec(encoding: HttpEncoding) = encoding match {
+    case HttpEncodings.gzip | HttpEncodings.`x-zip` => Gzip
+    case HttpEncodings.deflate => Deflate
+    case _ => NoCoding
+  }
+
+  def decode(resp: HttpResponse) = findCodec(resp.encoding).decode(resp)
+
+  val flow = Flow[HttpResponse].map(decode)
+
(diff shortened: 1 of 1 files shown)


ktoso:
OK TO TEST

akka-http/src/main/scala/akka/http/scaladsl/coding/NegotiableDecoder.scala
```diff
@@ -0,0 +1,17 @@
...
+  private def findCodec(encoding: HttpEncoding) = encoding match {
+    case HttpEncodings.gzip | HttpEncodings.`x-zip` => Gzip
+    case HttpEncodings.deflate => Deflate
+    case _ => NoCoding
+  }
+
+  def decode(resp: HttpResponse) = findCodec(resp.encoding).decode(resp)
+
+  def flow() = Flow[HttpResponse].map(decode)
+}
```
ktoso: In general I guess it'll be something around that style, yeah.
We should also allow picking a specific one in the same API.
Maybe we could make it more fluent, as in `via.(Coding.decode: Flow)`?

drewhk: The flow does not need to be a def, it is supposedly reusable.

13h3r:
ping

akka-http/src/main/scala/akka/http/scaladsl/coding/NegotiableDecoder.scala
```diff
@@ -0,0 +1,25 @@
+package akka.http.scaladsl.coding
```
ktoso: needs license header (see other files)

ktoso:
Hi @13h3r, sorry I didn't review this last week (much travel)!
As for testing it, have you seen [CodingDirectivesExamplesSpec](https://github.com/akka/akka/blob/release-2.3-dev/akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/directives/CodingDirectivesExamplesSpec.scala) it shows how such code can be tested (see `compress("Hello", Gzip)`). You'll need to write a tests that requests a compressed data (use a server that responds with compressed "Hello"), and on the client side you'd apply the decompression you're writing, most likely like so: `httpResponses.via(NegotiableDecoder.flow).runWith(Sink.head)` and then assert on the entity).

Hope this helps! Feel free to ask for help if stuck, I'm available again so should be able to help out :)

ktoso:
Hi @13h3r, any interest to take this further? :)

13h3r:
Hey, @ktoso. I have an interest, but right now don't have a lot of time :( I will continue this later. Thanks for ideas and help.

Issue #18737, opened 2015-10-19
create Streams 1.1 migration guide


rkuhn:
Include migration guide for https://github.com/akka/akka/pull/18700

drewhk:
AsyncStage deprecation
`isHoldingUpstream` returns false while handling completion

2m:
`Source(delay, interval, tick)` to `Source.tick(...)` in #18792

rkuhn:
removal of AsyncStage

viktorklang:
also #18806

drewhk:
remaining is: https://github.com/akka/akka/issues/18841

PR #18792, opened 2015-10-27
=str #15707 name unnamed modules

### LINKED ISSUE #15707: Revisit stream actor naming and hierarchy

2m:
- give name attribute to TickSource and nested modules
- reflow module toString output
- give toString to flow

Fixes #15707

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
@@ -335,12 +335,10 @@ private[akka] object StreamLayout {
     override def toString =
       s"""
         | Module: ${this.attributes.nameOrDefault("unnamed")}
-        | Modules: ${subModules.iterator.map(m ⇒ "   " + m.attributes.nameOrDefault(m.getClass.getName)).mkString("\n")}
-        | Downstreams:
-        | ${downstreams.iterator.map { case (in, out) ⇒ s"   $in -> $out" }.mkString("\n")}
-        | Upstreams:
-        | ${upstreams.iterator.map { case (out, in) ⇒ s"   $out -> $in" }.mkString("\n")}
-      """.stripMargin
+        | Modules: ${subModules.iterator.map(m ⇒ "\n   " + m.attributes.nameOrDefault(m.getClass.getName)).mkString("")}
+        | Downstreams: ${downstreams.iterator.map { case (in, out) ⇒ s"\n   $in -> $out" }.mkString("")}
+        | Upstreams: ${upstreams.iterator.map { case (out, in) ⇒ s"\n   $out -> $in" }.mkString("")}
+        |""".stripMargin
   }
 }
 
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
@@ -237,6 +237,8 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
 
   /** Converts this Scala DSL element to it's Java DSL counterpart. */
   def asJava: javadsl.Flow[In, Out, Mat] = new javadsl.Flow(this)
+
+  override def toString = s"""Flow(${module})"""
 }
 
 object Flow {
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
@@ -204,8 +204,8 @@ object Source {
    * element is produced it will not receive that tick element later. It will
    * receive new tick elements as soon as it has requested more elements.
    */
-  def apply[T](initialDelay: FiniteDuration, interval: FiniteDuration, tick: T): Source[T, Cancellable] =
-    fromGraph(new TickSource[T](initialDelay, interval, tick))
+  def tick[T](initialDelay: FiniteDuration, interval: FiniteDuration, tick: T): Source[T, Cancellable] =
+    fromGraph(new TickSource[T](initialDelay, interval, tick).withAttributes(DefaultAttributes.tickSource))
 
   /**
    * Create a `Source` with one element.
--- a/akka-docs-dev/rst/java/code/docs/MigrationsJava.java
+++ b/akka-docs-dev/rst/java/code/docs/MigrationsJava.java
@@ -1,12 +1,16 @@
 package docs;
 
+import akka.actor.Cancellable;
 import akka.japi.Pair;
 import akka.japi.function.Function;
 import akka.stream.*;
 import akka.stream.javadsl.*;
 import scala.Option;
+import scala.concurrent.duration.FiniteDuration;
 import scala.concurrent.Promise;
 import scala.runtime.BoxedUnit;
+
+import java.util.concurrent.TimeUnit;
 
 public class MigrationsJava {
 
@@ -110,6 +114,11 @@ public class MigrationsJava {
         Source<Integer, Promise<Option<Integer>>> src = Source.<Integer>maybe();
         // Complete the promise with an empty option to emulate the old lazyEmpty
         promise.trySuccess(scala.Option.empty());
+
+        final Source<String, Cancellable> sourceUnderTest = Source.tick(
+          FiniteDuration.create(0, TimeUnit.MILLISECONDS),
+          FiniteDuration.create(200, TimeUnit.MILLISECONDS),
+          "tick");
         //#source-creators
 
         //#empty-flow
--- a/akka-docs-dev/rst/java/migration-guide-1.0-2.x-java.rst
+++ b/akka-docs-dev/rst/java/migration-guide-1.0-2.x-java.rst
@@ -213,17 +213,16 @@ Source constructor name changes
 zero elements by providing an ``Option``. This is different from ``lazyEmpty`` which only allowed completion to be
 sent, but no elements.
 
-The ``apply()`` and ``from()`` overloads on ``Source`` that provide a tick source (``Source(delay,interval,tick)``)
-are replaced by the named method ``Source.tick()`` to reduce the number of overloads and to make the function more
+The ``from()`` overload on ``Source`` that provide a tick source (``Source.from(delay,interval,tick)``)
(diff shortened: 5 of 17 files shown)


akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -241,6 +241,9 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
 
   /** Converts this Scala DSL element to it's Java DSL counterpart. */
   def asJava: javadsl.Flow[In, Out, Mat] = new javadsl.Flow(this)
+
+  override def toString =
```
hepin1989: val?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -241,6 +241,9 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
...
+    s"""|Flow(${module.toString})""".stripMargin
```
drewhk: is this such a good idea? This can look really ugly.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -251,7 +251,7 @@ private[akka] object StreamLayout {
          * materialized value computation of its submodules.
          */
         Atomic(this),
-        Attributes.none)
+        Attributes.name("nested"))
```
drewhk: why?

viktorklang: Reuse same instance at the very least

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: why stripmargin?

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -212,7 +212,7 @@ object Source {
    * receive new tick elements as soon as it has requested more elements.
    */
   def apply[T](initialDelay: FiniteDuration, interval: FiniteDuration, tick: T): Source[T, Cancellable] =
-    fromGraph(new TickSource[T](initialDelay, interval, tick))
+    fromGraph(new TickSource[T](initialDelay, interval, tick).withAttributes(DefaultAttributes.iterableSource))
```
viktorklang: This should be named tickSource, tho?

drewhk: Yes, this apply should be renamed to tickSource. Please add the change to the list https://github.com/akka/akka/issues/18737 so we don't forget to add it to the migration guide.

drewhk: I mean pease rename *and* add it to that ticket.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
2m: What do you mean ugly? It can get pretty long, but if someone is printing out flows, then it probably means she is debugging. In that case verbose output is helpful. Here is the output of one flow created from graph, and another from ops:
```
Flow(
 Module: nested
 Modules: 
   akka.stream.impl.StreamLayout$CompositeModule
 Downstreams: 
   Flow.out -> Merge.in0
   Merge.out -> Broadast.in
   Broadcast.out0 -> Flow.in
   Flow.out -> SubscriberSink.in
 Upstreams: 
   Merge.in0 -> Flow.out
   Broadast.in -> Merge.out
   Flow.in -> Broadcast.out0
   SubscriberSink.in -> Flow.out
)
Flow(
 Module: filter
 Modules: 
   filter
   map
   take
 Downstreams: 
   Flow.out -> Flow.in
   Flow.out -> Flow.in
 Upstreams: 
   Flow.in -> Flow.out
   Flow.in -> Flow.out
)
```

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
2m: I would rather not. String representation could get rather long, so it should probably be better to create these long strings only when needed.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
2m: Otherwise some nested modules have no name attribute which trickles down to `unknown-op` in actor names.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -241,6 +241,8 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
 
   /** Converts this Scala DSL element to it's Java DSL counterpart. */
   def asJava: javadsl.Flow[In, Out, Mat] = new javadsl.Flow(this)
+
+  override def toString = s"""Flow(${module.toString})"""
```
viktorklang: isn't

~~~scala
override def toString = s"""Flow(${module.toString})"""
~~~

equivalent to

~~~scala
override def toString = s"""Flow(${module})"""
~~~

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -335,12 +336,10 @@ private[akka] object StreamLayout {
...
       s"""
         | Module: ${this.attributes.nameOrDefault("unnamed")}
-        | Modules: ${subModules.iterator.map(m ⇒ "   " + m.attributes.nameOrDefault(m.getClass.getName)).mkString("\n")}
-        | Downstreams:
-        | ${downstreams.iterator.map { case (in, out) ⇒ s"   $in -> $out" }.mkString("\n")}
-        | Upstreams:
-        | ${upstreams.iterator.map { case (out, in) ⇒ s"   $out -> $in" }.mkString("\n")}
-      """.stripMargin
+        | Modules: ${subModules.iterator.map(m ⇒ "\n   " + m.attributes.nameOrDefault(m.getClass.getName)).mkString("")}
+        | Downstreams: ${downstreams.iterator.map { case (in, out) ⇒ s"\n   $in -> $out" }.mkString("")}
```
viktorklang: Are these useful?
I saw things like:
Flow.in -> Flow.out
Flow.in -> Flow.out
…

Could someone expand on the utility?

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
```diff
@@ -49,6 +49,7 @@ private[stream] object Stages {
     val processor = name("processor")
     val processorWithKey = name("processorWithKey")
     val identityOp = name("identityOp")
+    val nestedModule = name("nested")
```
viktorklang: No other has "Module" as a part of the name so we can skip that part?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
2m: For the linear flows `upstreams` and `downstreams` sections might not be too useful, but it is very much useful in nested graphs shaped as a flow.

2m: Eg:
```
Flow(
 Module: nested
 Modules: 
   akka.stream.impl.StreamLayout$CompositeModule
 Downstreams: 
   Flow.out -> Merge.in0
   Merge.out -> Broadast.in
   Broadcast.out0 -> Flow.in
   Flow.out -> SubscriberSink.in
 Upstreams: 
   Merge.in0 -> Flow.out
   Broadast.in -> Merge.out
   Flow.in -> Broadcast.out0
   SubscriberSink.in -> Flow.out
)
```

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -251,7 +252,7 @@ private[akka] object StreamLayout {
          * materialized value computation of its submodules.
          */
         Atomic(this),
-        Attributes.none)
+        DefaultAttributes.nested)
```
rkuhn: Do we really want this to introduce a name component?

rkuhn:
LGTM apart from `nested`

drewhk:
does not compile right now

drewhk:
@2m this does not seem to compile, can you please rebase?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
2m: If `Attributes.none` is used, then actors get `unknown-operation` in their names. I think it is better to be exclusive about it and let actor name reflect how the module was constructed, that materialized to the particular actor.

rkuhn: Hmm, not sure I follow: Actor names are constructed from all Name attributes that are found in the list and only if none are found does it get called `unknown-operation`. Nesting modules should not add to the names generated, and not adding to the Attributes list should achieve that, no?

2m:
I am on it. DslConsistencySpec now.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
2m: Okay. I see now. I have reverted this to `Attributes.none`. It also means that this together with changes in [here](https://github.com/akka/akka/commit/162bd9b8e365b5e545bc9f555e079efb611cf20c#diff-c4bb0aec5f8c2431f9bf843216660886L187) produces lots of more actor names with "uknown-operation" in its names. Lets merge this PR as it morphed into rename `Source.from` to `Source.tick` and fix remaining "unknown-operation" in other ticket.

drewhk:
I merged the migration guides. Can you add the migration examples for the tick-source changes? They need just a few lines, I included everything except the new examples which could not be compiled without your changes.

drewhk:
Thanks, LGTM

PR #18796, opened 2015-10-27
use .head when could not inference whether the seq type is inde…

He-Pin:
use .head when could not inference whether the seq type is indexed or not.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/Shape.scala
+++ b/akka-stream/src/main/scala/akka/stream/Shape.scala
@@ -270,7 +270,7 @@ final case class BidiShape[-In1, +Out1, -In2, +Out2](in1: Inlet[In1 @uncheckedVa
   override def copyFromPorts(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]): Shape = {
     require(inlets.size == 2, s"proposed inlets [${inlets.mkString(", ")}] do not fit BidiShape")
     require(outlets.size == 2, s"proposed outlets [${outlets.mkString(", ")}] do not fit BidiShape")
-    BidiShape(inlets(0), outlets(0), inlets(1), outlets(1))
+    BidiShape(inlets.head, outlets.head, inlets(1), outlets(1))
   }
   def reversed: Shape = copyFromPorts(inlets.reverse, outlets.reverse)
   //#implementation-details-elided
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/BidiFlow.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/BidiFlow.scala
@@ -54,9 +54,9 @@ final class BidiFlow[-I1, +O1, -I2, +O2, +Mat](private[stream] override val modu
     val outs = copy.shape.outlets
     new BidiFlow(module
       .compose(copy, combine)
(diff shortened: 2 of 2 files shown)


akka-stream/src/main/scala/akka/stream/Shape.scala
```diff
@@ -270,7 +270,7 @@ final case class BidiShape[-In1, +Out1, -In2, +Out2](in1: Inlet[In1 @uncheckedVa
   override def copyFromPorts(inlets: immutable.Seq[Inlet[_]], outlets: immutable.Seq[Outlet[_]]): Shape = {
     require(inlets.size == 2, s"proposed inlets [${inlets.mkString(", ")}] do not fit BidiShape")
     require(outlets.size == 2, s"proposed outlets [${outlets.mkString(", ")}] do not fit BidiShape")
-    BidiShape(inlets(0), outlets(0), inlets(1), outlets(1))
+    BidiShape(inlets.head, outlets.head, inlets(1), outlets(1))
```
drewhk: maybe also `inlets.tail.head` instead of `inlets(1)`?

drewhk:
OK TO TEST

akka-stream/src/main/scala/akka/stream/Shape.scala
hepin1989: but sometime it's indexed seq.

drewhk: ok that is right.

viktorklang: I'd rather have it consistent (.apply), reason:
apply(0) should be cheap no matter what, and apply(1) should not be expensive either.

hepin1989: for List ```apply(0)``` will not be quick or cheap than just head.
```
  def apply(n: Int): A = {
    val rest = drop(n)
    if (n < 0 || rest.isEmpty) throw new IndexOutOfBoundsException("" + n)
    rest.head
  }
```

hepin1989: change to head will make it cheap for both indexed/nonIndexed .

viktorklang: For immutable.List:

~~~scala
override def drop(n: Int): List[A] = {
    var these = this
    var count = n
    while (!these.isEmpty && count > 0) {
      these = these.tail
      count -= 1
    }
    these
  }
~~~

so drop(0) == this, then this.head? (I mean cheap as no allocations)

And for vector it is:

~~~scala
def apply(index: Int): A = {
    val idx = checkRangeConvert(index)
    getElem(idx, idx ^ focus)
  }
~~~

(note that `head` on Vector is)

~~~scala
  override /*IterableLike*/ def head: A = {
    if (isEmpty) throw new UnsupportedOperationException("empty.head")
    apply(0)
  }
~~~

drewhk:
I guess this PR is not really needed? @viktorklang

viktorklang:
@drewhk It's an interesting proposal but given the implementations I don't think it'll be much of a win (compared to consistency).

drewhk:
my opinion as well. Closing it. @hepin1989 thanks a lot still!

hepin1989:
:) nice to hear.

Issue #17895, opened 2015-07-02
LoggingAdapter need support more args


hepin1989:
maybe using a generator to generate some code.to support more arity for `debug,info,error,warning`,
current the arity is limited to 4.

ktoso:
Sounds good to me, since we now have the boilerplate plugin on master in akka-user anyway we could make use of it for this.

2beaucoup:
Is it possible to switch to varargs?

ktoso:
varargs costs an array allocation - and this api tries to avoid additional allocations :)

2beaucoup:
I see. The boilerplate seems like the way to go.

patriknw:
Would it be possible to use varargs for cases with more than 4 args?

hepin1989:
I think it may be not need to be 22 args,but I think 5 or 6 at least.

patriknw:
yes, I think we should explore alternatives before boilerplating it, allocation overhead would be alright for the rare cases of > 4

kukido:
Could it be something like:

```scala
  def error(cause: Throwable, template: String, arg1: Any, arg2: Any, arg3: Any, arg4: Any, arg: Any*): Unit = {
    if (isErrorEnabled) {
      notifyError(cause, format(template, Seq(arg1, arg2, arg3, arg4) ++ arg))
    }
  }
```

patriknw:
That was what I had in mind. If the overloading works out I think that would be a good solution.

kukido:
@patriknw I believe now I have enough time to finish this item, which branch I should I work from?

ktoso:
@kukido, please work on the `master` branch for `2.4.x` features :-)

drewhk:
Moved due next version due to various binary compat issues.

hepin1989:
@kukido Ping:),will you take this up again?

patriknw:
@hepin1989 @2beaucoup @kukido I looked closer at the logging with many args. We actually already have support for this, but it's not documented. You can use the method with one arg and pass in an array there, and it will be unwrapped into args. We have a test for it here: https://github.com/akka/akka/blob/master/akka-actor-tests/src/test/java/akka/event/LoggingAdapterTest.java#L64-L72

I'll update the documentation tomorrow.

patriknw:
It was actually documented in reference docs, but I have improved the API docs in PR https://github.com/akka/akka/pull/22386

PR #18803, opened 2015-10-28
+act #17895 varargs support for LoggingAdapter

kukido:


kukido:
Somehow when I run `akka-actor-tests > test` the `LoggingAdapterTest` is not invoked. With manual run it goes through fine.

```
[DEBUG] [10/27/2015 21:51:53.875] [main] [EventStream(akka://test-system)] logger log1-Logging$DefaultLogger started
[DEBUG] [10/27/2015 21:51:53.879] [main] [EventStream(akka://test-system)] Default Loggers started
[ERROR] [10/27/2015 21:51:53.958] [main] [LoggingAdapterTest(akka://test-system)] One arg message: the arg
[ERROR] [10/27/2015 21:51:53.973] [main] [LoggingAdapterTest(akka://test-system)] Two args message: an arg, another arg
java.lang.IllegalStateException: This state is illegal
	at akka.event.LoggingAdapterTest$1.<init>(LoggingAdapterTest.java:40)
	at akka.event.LoggingAdapterTest.mustFormatMessage(LoggingAdapterTest.java:35)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:117)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:234)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:74)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:144)

[WARN] [10/27/2015 21:51:53.980] [main] [LoggingAdapterTest(akka://test-system)] Args as array of primitives: 10, 20, 30
[INFO] [10/27/2015 21:51:53.999] [main] [LoggingAdapterTest(akka://test-system)] Args as array of objects: A String, Tue Oct 27 21:51:53 PDT 2015, 6651ab8c-a9e0-45db-b915-d7ce98432bb0
[DEBUG] [10/27/2015 21:51:54.001] [main] [LoggingAdapterTest(akka://test-system)] Four args message: 1, 2, 3, 4
[DEBUG] [10/27/2015 21:51:54.002] [main] [LoggingAdapterTest(akka://test-system)] Six args message: 1, 2, 3, 4, 5, 6

Process finished with exit code 0
```

akka-actor/src/main/scala/akka/event/Logging.scala
```diff
@@ -976,6 +982,12 @@ trait LoggingAdapter {
    * @see [[LoggingAdapter]]
    */
   def warning(template: String, arg1: Any, arg2: Any, arg3: Any, arg4: Any): Unit = { if (isWarningEnabled) notifyWarning(format(template, arg1, arg2, arg3, arg4)) }
+  /**
+   * Message template with varargs replacement arguments.
+   * @see [[LoggingAdapter]]
+   */
+  @annotation.varargs
+  def warning(template: String, arg1: Any, arg2: Any, arg3: Any, arg4: Any, args: Any*): Unit = { if (isWarningEnabled) notifyWarning(format(template, Seq(arg1, arg2, arg3, arg4) ++ args:_*)) }
```
patriknw: please start this and others on a new line
(we should try to use < 120 char lines)

patriknw:
OK TO TEST

patriknw:
Thanks for contributing! Looks great!

patriknw:
LGTM

kukido:
Updated all methods with new line, to be consistent.

ktoso:
LGTM, thanks for contributing!

ktoso:
MiMa freaked out about these methods, which sees a bit weird since it's only adding methods...?
Is there a real binary compatibility issue here? 

```
[info] akka-actor: found 10 potential binary incompatibilities while checking against com.typesafe.akka:akka-actor_2.11:2.3.11  (filtered 236)
[error]  * method log(Int,java.lang.String,java.lang.Object,java.lang.Object,java.lang.Object,java.lang.Object,scala.collection.Seq)Unit in trait akka.event.LoggingAdapter does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("akka.event.LoggingAdapter.log")
[error]  * method 
```

kukido:
Yes, the methods were added. I mean, the message is correct, the methods were not in 2.3. Now, the question is - why does it complain about 10 methods added, while I added only 5 with new signature- (log, info, warning, error, debug). Where can I see the complete log?

kukido:
Ok, got the log, and I see why it is 10, not 5. It is because of the annotation to generate Java-specific method - so there are two for each signature. Makes sense.
I added those methods to ProblemFilters, per MiMa suggestion.

rkuhn:
No: the warnings flagged by MiMa are actual incompatibilities. Adding methods to a trait is a binary incompatible action because existing subclasses won’t have them.

rkuhn:
In short: MiMa warnings can only ever be ignored for code that cannot possibly be called or otherwise accessed directly by client code. There is no such thing as a non-problematic MiMa warning, each and every one needs careful investigation—and sometimes even INTERNAL API cannot be changed because the compiler generates accesses to it even for well-behaved client code.

patriknw:
Isn't LoggingAdapter an interface that can be implemented by user code? Adding more methods would break existing implementations, at least on the source compatiblity level.

kukido:
@rkuhn well, the code is definitely available to clients, I see the problem now. We can introduce another trait, with varargs only?

akka-actor/src/main/scala/akka/event/Logging.scala
```diff
@@ -903,152 +903,251 @@ trait LoggingAdapter {
...
+  def error(template: String, arg1: Any, arg2: Any, arg3: Any, arg4: Any): Unit = {
+    if (isErrorEnabled) notifyError(format(template, arg1, arg2, arg3, arg4))
+  }
+  /**
+   * Message template with varargs replacement arguments.
+   * @see [[LoggingAdapter]]
+   */
+  @annotation.varargs
+  def error(template: String, arg1: Any, arg2: Any, arg3: Any, arg4: Any, args: Any*): Unit = {
+    if (isErrorEnabled) notifyError(format(template, Seq(arg1, arg2, arg3, arg4) ++ args: _*))
```
rkuhn: Please manually allocate the right array size and use `System.arrayCopy` to populate it from `args` (using `args.asInstanceOf[WrappedArray[Any]].array`).

rkuhn:
Yes, adding a derived trait would work AFAICS.

kukido:
Now I'm having some doubts about this change. I looked at github, on how the code is used and it is mostly `LoggingAdapter log = Logging.getLogger(getContext().system(), this)`, where `object Logging` works with `LoggingAdapter`, when we ask for a new logger, then `class BusLogging extends LoggingAdapter` is returned as implementation. So even if we add `LoggingAdapterExt` to `BusLogging`, it will not be enough, as we would have to change `Logging` for the change to be effective, and it is definitely not binary compatible.

rkuhn:
Well, this comes with binary compatibility constraints: you’ll have to extend the whole hierarchy, including ActorLoggingWithVarargs.

patriknw:
and then someone has to ask how important this is?

kukido:
That means there will be `LoggingWithVarargs`, `BusLoggingWithVarargs` and maybe some more. Do we want to do that? Looks heavy-handed, IMHO.
How about adding `LoggingAdapterWithVarargs` and implicit conversion from `LoggingAdapter` to `LoggingAdapterWithVarargs` aka "Pimp My Library" pattern? Then whoever wants to add varargs functionality, they can do that simply by adding an implicit. :speak_no_evil:
An alternative would be just to add the sample code from above to a cookbook, and let people add it to their code.

rkuhn:
refs #17895 

An implicit conversion will not work unless the vararg method has a different name than the original—which might also make sense to keep things understandable. This could then also be offered for Java as a wrapper (just like PrintStream over OutputStream).

kukido:
Varargs is just syntactic sugar, as the library already supports Object[] as an argument. 
```
Date now = new Date();
UUID uuid = UUID.randomUUID();
Object[] objArgs = {"A String", now, uuid};
log.info("Args as array of objects: {}, {}, {}", objArgs);
expectLog(InfoLevel(), "Args as array of objects: A String, " + now.toString() + ", " + uuid.toString());
```
The question is, where do we want it, in the library or on a client.

patriknw:
a wrapper sounds good to me

kukido:
`LoggingAdapterExt` is the wrapper. I have to add implicit for Scala, but for Java it would be something like:
```
        final LoggingAdapter log = Logging.getLogger(system, this);
        final LoggingAdapterExt logext = new LoggingAdapterExt(log);
        new LogJavaTestKit(system) {{
            system.eventStream().subscribe(getRef(), LogEvent.class);
            logext.debug("Six args message: {}, {}, {}, {}, {}, {}", 1, 2, 3, 4, 5, 6);
            expectLog(DebugLevel(), "Six args message: 1, 2, 3, 4, 5, 6");

            int[] primitiveArgs = {10, 20, 30};
            logext.warning("Args as array of primitives: {}, {}, {}", primitiveArgs);
            expectLog(WarningLevel(), "Args as array of primitives: 10, 20, 30");
        }};
```

kukido:
WRT implicit, should I create a package object and then add it there? What is the best practice?
`implicit def loggingAdapterToLoggingAdapterEx(adapter: LoggingAdapter): LoggingAdapterExt = new LoggingAdapterExt(adapter)` 
@rkuhn @patriknw

drewhk:
bah, javadoc failed

akka-actor/src/main/scala/akka/event/Logging.scala
```diff
@@ -1233,9 +1298,84 @@ object NoLogging extends LoggingAdapter {
...
-  final protected override def notifyInfo(message: String): Unit = ()
-  final protected override def notifyDebug(message: String): Unit = ()
+  final protected[akka] override def notifyError(message: String): Unit = ()
+  final protected[akka] override def notifyError(cause: Throwable, message: String): Unit = ()
+  final protected[akka] override def notifyWarning(message: String): Unit = ()
+  final protected[akka] override def notifyInfo(message: String): Unit = ()
+  final protected[akka] override def notifyDebug(message: String): Unit = ()
+}
+
+class LoggingAdapterExt(delegate: LoggingAdapter) extends LoggingAdapter {
```
ktoso: Looks as if we've traded allocating the array into allocating the wrapper for each call - not a huge win as initially planned.

Would it be possible to make it an `implicit class` and `extends AnyVal` to avoid the allocations?

ktoso: Semantics wise I understand it's doing what it should (does not conflict with the varargs version)?

kukido: I tried that approach. Getting compiler error on `@annotation.varargs`:

```
[error] (akka-actor/compile:compile) java.lang.AssertionError: assertion failed: no extension method found for:
[error]
[error]   method debug:(template: String, args: Any*)Unit
[error]
[error]  Candidates:
[error]
[error]
[error]
[error]  Candidates (signatures normalized):
[error]
[error]
[error]
[error]  Eligible Names: debug$extension0,debug$extension1"
[error] Total time: 2 s, completed Nov 11, 2015 11:37:17 AM
```

Without the annotation Java code does not work.

screen shot 2015-11-11 at 11 40 25 am

And Java code looks ugly (I added implicit class to a companion object for LoggingAdapter):

```
final LoggingAdapter log = Logging.getLogger(system, this);
final LoggingAdapter.LoggingAdapterExt logext = new LoggingAdapter.LoggingAdapterExt(log);
```

I think I'm hitting a threshold on that ticket, in terms of value/effort. Maybe we better put it into a backlog until we have a release where we can add backward incompatible changes and just add the method into LoggingAdapter as suggested initially?

drewhk: I would opt for backlogging it, yes. I refile the ticket for next major.

drewhk:
I close this PR for now. Anyone can reopen it if it still feels relevant.

Issue #11214, opened 2011-09-21
CAMEL: Akka-Camel Message is not wrapping Camel Jetty Message's .getAttachment method (for multi-part form support)


krasserm:
imported from https://www.assembla.com/spaces/akka/tickets/1214

(taken from akka-modules)

This is the suggested way of doing multi-part requests on Camel's Jetty Component:

h1. *Multi-part Form support*
 
From Camel 2.3.0, camel-jetty support to multipart form post out of box. The submitted form-data are mapped into the message header. Camel-jetty creates an attachment for each uploaded file. The file name is mapped to the name of the attachment. The content type is set as the content type of the attachment file name. You can find the example here.

Note: getName() functions as shown below in versions 2.5 and higher. In earlier versions you receive the temporary file name for the attachment instead

<pre><code>// Set the jetty temp directory which store the file for multi part form
// camel-jetty will clean up the file after it handled the request.
// The option works rightly from Camel 2.4.0
getContext().getProperties().put("CamelJettyTempDir", "target");

from("jetty://http://localhost:{{port}}/test").process(new Processor() {

    public void process(Exchange exchange) throws Exception {
        Message in = exchange.getIn();
        assertEquals("Get a wrong attachement size", 1, in.getAttachments().size());
        // The file name is attachment id
        DataHandler data = in.getAttachment("NOTICE.txt");

        assertNotNull("Should get the DataHandle NOTICE.txt", data);
        // This assert is wrong, but the correct content-type (application/octet-stream)
        // will not be returned until Jetty makes it available - currently the content-type
        // returned is just the default for FileDataHandler (for the implentation being used)
        //assertEquals("Get a wrong content type", "text/plain", data.getContentType());
        assertEquals("Got the wrong name", "NOTICE.txt", data.getName());

        assertTrue("We should get the data from the DataHandle", data.getDataSource()
            .getInputStream().available() > 0);

        // The other form date can be get from the message header
        exchange.getOut().setBody(in.getHeader("comment"));
    }

});
</code></pre>

However we're missing the getAttachment and Attachment method family in general in Akka-Camel's Message hence it's impossible to achieve the desired functionality through mentioned component.

rkuhn:
Has this been solved in the rewrite of the camel module? /cc @RayRoestenburg

PR #18815, opened 2015-10-30
Add attachments in CamelMessage

visox:
As taken from JCamelMessage
And updated test

An old issue
https://github.com/akka/akka/issues/11214

### DIFF
--- a/akka-camel/src/main/scala/akka/camel/CamelMessage.scala
+++ b/akka-camel/src/main/scala/akka/camel/CamelMessage.scala
@@ -5,9 +5,11 @@
 package akka.camel
 
 import java.util.{ Map ⇒ JMap, Set ⇒ JSet }
+import javax.activation.DataHandler
 import org.apache.camel.{ CamelContext, Message ⇒ JCamelMessage, StreamCache }
 import akka.AkkaException
 import scala.reflect.ClassTag
+import scala.runtime.ScalaRunTime
 import scala.util.Try
 import scala.collection.JavaConversions._
 import akka.dispatch.Mapper
@@ -15,10 +17,14 @@ import akka.dispatch.Mapper
 /**
  * An immutable representation of a Camel message.
  */
-case class CamelMessage(body: Any, headers: Map[String, Any]) {
-  def this(body: Any, headers: JMap[String, Any]) = this(body, headers.toMap) //for Java
+class CamelMessage(val body: Any, val headers: Map[String, Any], val attachments: Map[String, DataHandler]) extends Serializable with Product {
+  def this(body: Any, headers: JMap[String, Any]) = this(body, headers.toMap, Map.empty[String, DataHandler]) //Java
+  def this(body: Any, headers: JMap[String, Any], attachments: JMap[String, DataHandler]) = this(body, headers.toMap, attachments.toMap) //Java
+  def this(body: Any, headers: Map[String, Any]) = this(body, headers.toMap, Map.empty[String, DataHandler])
 
-  override def toString: String = "CamelMessage(%s, %s)" format (body, headers)
+  def copy(body: Any = this.body, headers: Map[String, Any] = this.headers): CamelMessage = CamelMessage(body, headers, this.attachments)
+
+  override def toString: String = "CamelMessage(%s, %s, %s)" format (body, headers, attachments)
 
   /**
    * Returns those headers from this message whose name is contained in <code>names</code>.
@@ -112,7 +118,7 @@ case class CamelMessage(body: Any, headers: Map[String, Any]) {
   /**
    * Java API: Returns a new CamelMessage with a new body, while keeping the same headers.
    */
-  def withBody[T](body: T) = this.copy(body = body)
+  def withBody[T](body: T): CamelMessage = copy(body = body)
   /**
    * Creates a CamelMessage with current <code>body</code> converted to type <code>T</code>.
    * The CamelContext is accessible in a [[akka.camel.javaapi.UntypedConsumerActor]] and [[akka.camel.javaapi.UntypedProducerActor]]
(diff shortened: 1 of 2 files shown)


ktoso:
OK TO TEST

visox:
@ktoso @rkuhn can you help me or link me some stuff about binary compatibility ?

rkuhn:
The errors are in the log:
~~~
[info] akka-camel: found 4 potential binary incompatibilities while checking against com.typesafe.akka:akka-camel_2.11:2.3.12  (filtered 8)
[error]  * method from(org.apache.camel.Message,scala.collection.immutable.Map)akka.camel.CamelMessage in object akka.camel.CamelMessage does not have a correspondent in new version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("akka.camel.CamelMessage.from")
[error]  * method apply(java.lang.Object,scala.collection.immutable.Map)akka.camel.CamelMessage in object akka.camel.CamelMessage does not have a correspondent in new version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("akka.camel.CamelMessage.apply")
[error]  * method copy(java.lang.Object,scala.collection.immutable.Map)akka.camel.CamelMessage in class akka.camel.CamelMessage does not have a correspondent in new version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("akka.camel.CamelMessage.copy")
[error]  * method this(java.lang.Object,scala.collection.immutable.Map)Unit in class akka.camel.CamelMessage's type has changed; was (java.lang.Object,scala.collection.immutable.Map)Unit, is now: (java.lang.Object,java.util.Map)Unit
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("akka.camel.CamelMessage.this")
~~~
This is what I meant in the ticket: you cannot add fields to a case class normally, that breaks binary compatibility. I guess you’ll end up duplicating all methods the compiler auto-generates for a case class, which means also removing the `case` keyword from the definition. Another possibility would be to create a subclass that has attachments, but that subclass cannot be a `case class` then; users will either have to downcast (after a type test) or we add the test&cast to new accessor methods on the existing class.

visox:
@rkuhn thanks, will see if i can make it work.

johanandren:
LGTM

visox:
@ktoso any idea when someone could have the time for this pr ? i realise there are more important thinks so i am just asking for the time window.

ktoso:
Release wise it'll land in 2.4.1 in any case, so that's "near future" until you can use it :) 
Review wise I'll try very soon.

visox:
Thanks, yes we need it in our project and having a local akka version is probably not the right way :)

akka-camel/src/main/scala/akka/camel/CamelMessage.scala
```diff
@@ -128,6 +129,37 @@ case class CamelMessage(body: Any, headers: Map[String, Any]) {
    */
   def withBodyAs[T](clazz: Class[T])(implicit camelContext: CamelContext): CamelMessage = copy(body = getBodyAs(clazz, camelContext))
 
+  /**
+   * Java API: Creates a new CamelMessageWithAttachments with given <code>attachments</code>. A copy of the attachments map is made.
+   */
+  def withAttachments(attachments: JMap[String, DataHandler]): CamelMessageWithAttachments = new CamelMessageWithAttachments(this.body, this.headers, attachments.toMap)
+
+  /**
+   * Creates a new CamelMessageWithAttachments with given <code>attachments</code>.
```
rkuhn: This should be marked `Scala API:`, symmetrical with its Java counterpart.

akka-camel/src/main/scala/akka/camel/CamelMessage.scala
```diff
@@ -128,6 +129,37 @@ case class CamelMessage(body: Any, headers: Map[String, Any]) {
...
+   */
+  def withAttachments(attachments: Map[String, DataHandler]): CamelMessageWithAttachments = new CamelMessageWithAttachments(this.body, this.headers, attachments)
+}
+
+class CamelMessageWithAttachments(override val body: Any, override val headers: Map[String, Any], val attachments: Map[String, DataHandler]) extends CamelMessage(body, headers) {
```
rkuhn: instead of `override val` please make these pure constructor parameters that are passed through to the superclass; our convention is to prefix the name with an underscore to ensure that it is not unintentionally used in the class body (which would lead to an additional field being created within the class)

rkuhn:
Thinking about this some more: the subclass would also need to override the `copy` method so that copying a CamelMessage does not lose its annotations. And then there is the non-nicety of user code having to downcast to CamelMessageWithAttachments, which may or may not work dynamically.

Adriaan just pointed me towards this [earlier discussion](https://groups.google.com/d/msg/sbt-dev/3VJEnuO9ktM/cL93sx2d86kJ) which links to a commit that does part of the case class desugaring for sbt’s IncOptions. This may be a better solution if it can be pulled off:

* remove the `case` modifier
* add `val` to the fields
* add attachments field
* add secondary constructor that sets empty map for attachments
* add `copy` method exactly as the old class had (take a look at the bytecode with javap if you are unsure)
* do not add `copy` method which includes attachments, that is better served by `withAttachments`
* add `apply` and `unapply` to the companion object just like they were defined before (see bytecode)
* add `((Any, Map[String, Any]) => CamelMessage)` supertype to companion object (yes, case classes do that, too ;-) )

I hope this should be it, but MiMa will tell you for sure. If this works then we’ll have a seamlessly usable upgrade that is still compatible with old code. Sorry for the hassle!

visox:
thanks, see you with the fixes.

visox:
`This may be a better solution if it can be pulled off: ....`

It was my first intention to modify the existing case class in this way, you are looking further, will see what can be done.

Should be definitely better then a new extension class (as it is now).

rkuhn:
Thanks for exploring this, I don’t think we have done such a change before and it will be good to know a scheme that works.

visox:
@rkuhn Take a look if this is what you had in your mind.

So now its just one (non case) class with additional attachment field and few functions that allow us to stay binary compatible.

I should probably add comments, but i would like to see a opinion on this first.

akka-camel/src/main/scala/akka/camel/CamelMessage.scala
```diff
@@ -112,7 +118,7 @@ case class CamelMessage(body: Any, headers: Map[String, Any]) {
   /**
    * Java API: Returns a new CamelMessage with a new body, while keeping the same headers.
    */
-  def withBody[T](body: T) = this.copy(body = body)
+  def withBody[T](body: T) = copy(body = body)
```
rkuhn: Please add explicit return type while we are at it.

rkuhn:
Yes, this is looking good, thanks!

visox:
@rkuhn @ktoso If nothing else, it can me merged.

rkuhn:
One last step: please squash to one commit with summary line “+cam #11214 add attachments to CamelMessage”, then we can click the button.

visox:
@rkuhn go for it

rkuhn:
Thanks again @visox !

Issue #18808, opened 2015-10-29
Instead of separate fanoutPublisher, just add a boolean parameter


drewhk:
to Sink.publisher. 

This makes it easier to discover the fanout pubisher (and be aware of the non-reusability aspect) without changing any internals.

viktorklang:
@drewhk How should we deal with initialBufferSize and maximumBufferSize? Ignore them if not a fanout buffer?

Thoughts?

viktorklang:
How about this instead (pseudo):

~~~scala
Sink.publisher(maxNumberOfConcurrentSubscribers: Int, maxInterSubscriberDemandDivergence: Int): X = maxNumberOfConcurrentSubscribers match {
   case n if n < 1 => ErrorPublisher("No Pub for you!")
   case 1 => new PublisherSink()
   case n => new FanoutPublisherSink(n, maxInterSubscriberDemandDivergence / 2, maxInterSubscriberDemandDivergence)
}
~~~

viktorklang:
See #18819

PR #18819, opened 2015-10-30
-str - 18808 - Removes Sink.fanoutPublisher

viktorklang:
Sink.publisher now takes a max number of Subscribers and
the elasticity between concurrent Subscribers.

### DIFF
--- a/akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
+++ b/akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
@@ -41,7 +41,7 @@ class ReactiveStreamsDocSpec extends AkkaSpec {
 
   val impl = new Fixture {
     override def tweets: Publisher[Tweet] =
-      TwitterStreamQuickstartDocSpec.tweets.runWith(Sink.publisher(1))
+      TwitterStreamQuickstartDocSpec.tweets.runWith(Sink.publisher(false))
 
     override def storage = TestSubscriber.manualProbe[Author]
 
@@ -92,7 +92,7 @@ class ReactiveStreamsDocSpec extends AkkaSpec {
 
     //#source-publisher
     val authorPublisher: Publisher[Author] =
-      Source(tweets).via(authors).runWith(Sink.publisher(1))
+      Source(tweets).via(authors).runWith(Sink.publisher(fanout = false))
 
     authorPublisher.subscribe(storage)
     //#source-publisher
@@ -108,7 +108,7 @@ class ReactiveStreamsDocSpec extends AkkaSpec {
     //#source-fanoutPublisher
     val authorPublisher: Publisher[Author] =
       Source(tweets).via(authors)
-        .runWith(Sink.publisher(maxNumberOfSubscribers = Int.MaxValue))
+        .runWith(Sink.publisher(fanout = true))
 
     authorPublisher.subscribe(storage)
     authorPublisher.subscribe(alert)
--- a/akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
+++ b/akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
@@ -41,7 +41,7 @@ class ReactiveStreamsDocSpec extends AkkaSpec {
 
   val impl = new Fixture {
     override def tweets: Publisher[Tweet] =
-      TwitterStreamQuickstartDocSpec.tweets.runWith(Sink.publisher)
+      TwitterStreamQuickstartDocSpec.tweets.runWith(Sink.publisher(1))
 
     override def storage = TestSubscriber.manualProbe[Author]
 
@@ -92,7 +92,7 @@ class ReactiveStreamsDocSpec extends AkkaSpec {
 
     //#source-publisher
     val authorPublisher: Publisher[Author] =
-      Source(tweets).via(authors).runWith(Sink.publisher)
+      Source(tweets).via(authors).runWith(Sink.publisher(1))
 
     authorPublisher.subscribe(storage)
     //#source-publisher
@@ -108,7 +108,7 @@ class ReactiveStreamsDocSpec extends AkkaSpec {
     //#source-fanoutPublisher
     val authorPublisher: Publisher[Author] =
       Source(tweets).via(authors)
-        .runWith(Sink.fanoutPublisher(initialBufferSize = 8, maximumBufferSize = 16))
+        .runWith(Sink.publisher(maxNumberOfSubscribers = Int.MaxValue))
 
     authorPublisher.subscribe(storage)
     authorPublisher.subscribe(alert)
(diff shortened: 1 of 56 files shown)


viktorklang:
@akka/akka-committers I'm not really happy with the result, but would love to hear feedback and suggestions

drewhk:
I don't like that the single-use publisher now needs two parameters, one of which is unnecessary.

viktorklang:
@drewhk what do you suggest?

viktorklang:
@drewhk remove the second parameter altogether and hard-code a default? Or perhaps take the inputBufferSize?

drewhk:
In Scala the case class version could be used:

    Sink.publisher(SingleSubscriber) // SingleSubscriber can be default param
    or
    Sink.publisher(Fanout(16)) // case class can have default param 

I am not sure about the Java side

viktorklang:
@drewhk I'm trying the inputBuffer approach. Most people will have no idea what to use for the second parameter anyway.

viktorklang:
@drewhk Pushed an update with the inputBuffer solution, I kinda like it!

drewhk:
I don't see the update

viktorklang:
Thanks for the heads up, it seems as if I had a network glitch. Should be
there now.

On Mon, Nov 2, 2015 at 12:44 PM, drewhk <notifications@github.com> wrote:

> I don't see the update
>
> —
> <https://github.com/akka/akka/pull/18819#issuecomment-152996200>.
>



-- 
Cheers,
√

akka-docs-dev/rst/scala/code/docs/stream/ReactiveStreamsDocSpec.scala
```diff
@@ -92,7 +92,7 @@ class ReactiveStreamsDocSpec extends AkkaSpec {
 
     //#source-publisher
     val authorPublisher: Publisher[Author] =
-      Source(tweets).via(authors).runWith(Sink.publisher)
+      Source(tweets).via(authors).runWith(Sink.publisher(1))
```
drewhk: cannot we have a default of one?

viktorklang: I don't want to do that for 2 reasons:

1) that would mean we'd have to use overloading for Java
2) that would also mean that given the path of least resistance, people would use the parameterless one without thinking and then would wonder why they can only attach a single subscriber.

drewhk:
I like the fanout parameters coming from the buffer settings. I am not sure though that an explicit number of maximum subscribers is that helpful. Why not just one or quasi-infinite?

viktorklang:
@drewhk 

>I like the fanout parameters coming from the buffer settings. I am not sure though that an explicit number of maximum subscribers is that helpful. Why not just one or quasi-infinite?

That's a great question, the reason why it isn't a boolean is because I couldn't come up with a single reason a user would understand why they'd not always allow an unbounded number of Subscribers.

Which led me to the following dilemma:

A) Rewrite the FanoutProcessorImpl to be as fast as VirtualProcessor for a single Subscriber and then drop the boolean parameter altogether
B) Give the user a conscious decision that the number of Subscribers that it can support is a meaningful one

Now, I can completely get behind the optimization idea, but I don't think we have the time for that.
In the future, assuming that such an optimization is made, can then introduce a parameterless version of `publisher` and then deprecate the one with the int parameter.

drewhk:
> Now, I can completely get behind the optimization idea, but I don't think we have the time for that.
In the future, assuming that such an optimization is made, can then introduce a parameterless version of publisher and then deprecate the one with the int parameter.

I am very skeptical about being able to compete with VirtualProcessor with a Fanout implementation though. Also, I am not sure that there are many use-cases for saying, "there cannot be more than 5 Subscribers", as these are usually well covered by the Graph DSL.

viktorklang:
@drewhk 

> Also, I am not sure that there are many use-cases for saying, "there cannot be more than 5 Subscribers", as these are usually well covered by the Graph DSL.

You'll have to assume that the resulting Publisher will be passed into some other RS implementation, and as such the Graph DSL does not apply anymore.

drewhk:
Unless they provide a Subscriber of course. Also, RS implementations either assume no reusability of the Publisher or complete reusability, I am not aware of "at most N" constraints.

viktorklang:
That's not the full picture. There are Publishers that run independently
for every Subscriber, there are ones which broadcast and there are those
who are free to do anything they want as long as they follow the spec :-)

-- 
Cheers,
√

viktorklang:
I'm open to switch to a 'broadcast: Boolean' parameter if you think that'll
help

-- 
Cheers,
√

drewhk:
> That's not the full picture. There are Publishers that run independently
for every Subscriber, there are ones which broadcast and there are those
who are free to do anything they want as long as they follow the spec :-)

Yes of course. I just want to avoid forcing users to a choice which they probably have hard time to get right first, and later they will devolve to using Int.MaxValue anyway.

viktorklang:
Ok, so let's switch to a broadcast boolean

-- 
Cheers,
√

drewhk:
please rebase

drewhk:
Also please add a section to the migration guide

viktorklang:
@drewhk rebased, switched to a boolean and updated migration guide. I kept the commit which turned it into Int if we ever want to revert.

akka-docs-dev/rst/migration-guide-1.0-2.x.rst
```diff
@@ -161,6 +161,23 @@ Update procedure
...
+To make the decision whether to support fanout or not an active one, the aforementioned methods have been
+replaced with a single method: ``Sink.publisher(fanout: Boolean)``.
+
+Update procedure
+----------------
+
+1. Replace all occurences of ``Sink.publisher`` with ``Sink.publisher(false)``
+2. Replace all occurences of ``Sink.fanoutPublisher`` with ``Sink.publisher(true)``
+
+TODO: code example
```
drewhk: Please prepare it like here: https://github.com/akka/akka/pull/18834
I will quickly push an update with some Java changes and merge, so you can rebase on top of it.

viktorklang: Why did you ask me to rebase if I'm going to have to rebase again? We need to plan the merges to avoid wasting time.

drewhk: (pushed)

viktorklang: Pushed what where?

drewhk: You have not created Scala and Java code snippets:
 - how it looked before (check in a separate project against 1.0 to see if they really compile)
 - how they should look now

I referred to my PR where there are examples how to do this.

viktorklang: @drewhk My point was: In the current release-2.3-dev there are no migration code examples. I suggest you either A) merge this against the current release-2.3-dev branch and then I can create code examples for you to include on your PR so they all go in at once, or merge your PR and I'll rebase on top of that.

drewhk: Ok, I will merge it soon.

drewhk:
PLS BUILD

drewhk:
refs #18808

Issue #18735, opened 2015-10-18
FAILED: RecipeKeepAlive


ktoso:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3923/consoleFull

```
info] Recipe for injecting keepalive messages
[info] - must work *** FAILED *** (43 milliseconds)
[info]   java.lang.AssertionError: assertion failed: expected OnNext(ByteString(11)), found OnNext(ByteString(3))
[info]   at scala.Predef$.assert(Predef.scala:179)
[info]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:339)
[info]   at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:315)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:718)
[info]   at akka.stream.testkit.TestSubscriber$ManualProbe.expectNext(StreamTestKit.scala:252)
[info]   at docs.stream.cookbook.RecipeKeepAlive$$anonfun$1$$anonfun$apply$mcV$sp$1.apply$mcV$sp(RecipeKeepAlive.scala:55)
[info]   at docs.stream.cookbook.RecipeKeepAlive$$anonfun$1$$anonfun$apply$mcV$sp$1.apply(RecipeKeepAlive.scala:11)
[info]   at docs.stream.cookbook.RecipeKeepAlive$$anonfun$1$$anonfun$apply$mcV$sp$1.apply(RecipeKeepAlive.scala:11)
[info]   at org.scalatest.Transformer$$anonfun$apply$1.apply$mcV$sp(Transformer.scala:22)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.WordSpecLike$$anon$1.apply(WordSpecLike.scala:953)
[info]   at org.scalatest.Suite$class.withFixture(Suite.scala:1122)
[info]   at akka.stream.testkit.AkkaSpec.withFixture(AkkaSpec.scala:55)
[info]   at org.scalatest.WordSpecLike$class.invokeWithFixture$1(WordSpecLike.scala:950)
[...]

PR #18826, opened 2015-11-01
!str,htc replace and remove OneBoundedInterpreter

rkuhn:
main work by @drewhk with contributions from @2m and @rkuhn

This work uncovered many well-hidden bugs in existing Stages, in
particular StatefulStage. These were hidden by the behavior of
OneBoundedInterpreter that normally behaves more orderly than it
guarantees in general, especially with respect to the timeliness of
delivery of upstream termination signals; the bugs were then that
internal state was not flushed when onComplete arrived “too early”.


rkuhn:
Migration guide is not yet finished, I just took what @drewhk had written by last Friday.

rkuhn:
test failure already tracked as #18735

rkuhn:
After test with `eventLimit = 1` was successful, now pushed the Real Thing.

rkuhn:
Manually merged this before it grows any more conflicts.

Issue #18756, opened 2015-10-21
FAILED: GraphZipSpec


rkuhn:
possibly flaky: https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/3901/consoleFull

~~~
[info] Zip
[info] - must work in the happy case *** FAILED *** (3 seconds, 32 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsg while waiting for OnComplete
[info]   at scala.Predef$.assert(Predef.scala:179)
[info]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:338)
[info]   at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:315)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:718)
[info]   at akka.stream.testkit.TestSubscriber$ManualProbe.expectComplete(StreamTestKit.scala:319)
[info]   at akka.stream.scaladsl.GraphZipSpec$$anonfun$1$$anonfun$apply$mcV$sp$1$$anonfun$apply$mcV$sp$7.apply(GraphZipSpec.scala:48)
[info]   at akka.stream.scaladsl.GraphZipSpec$$anonfun$1$$anonfun$apply$mcV$sp$1$$anonfun$apply$mcV$sp$7.apply(GraphZipSpec.scala:25)
[info]   at akka.stream.testkit.Utils$.assertAllStagesStopped(Utils.scala:26)
[info]   at akka.stream.scaladsl.GraphZipSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply$mcV$sp(GraphZipSpec.scala:25)
[info]   at akka.stream.scaladsl.GraphZipSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply(GraphZipSpec.scala:25)
[info]   at akka.stream.scaladsl.GraphZipSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply(GraphZipSpec.scala:25)
~~~

ktoso:
Another one

```
nfo] - must work in the happy case *** FAILED *** (3 seconds, 37 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsg while waiting for OnComplete
[info]   at scala.Predef$.assert(Predef.scala:179)
[info]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:338)
[...]

PR #18843, opened 2015-11-03
=str #18756: Fix Zip completion

drewhk:
refs #18756

### DIFF
--- a/akka-stream/src/main/boilerplate/akka/stream/scaladsl/ZipWithApply.scala.template
+++ b/akka-stream/src/main/boilerplate/akka/stream/scaladsl/ZipWithApply.scala.template
@@ -42,7 +42,7 @@ class ZipWith1[[#A1#], O] (zipper: ([#A1#]) ⇒ O) extends GraphStage[FanInShape
       }
 
       override def onUpstreamFinish(): Unit = {
-        if (!isAvailable(out) || !isAvailable(in0)) completeStage()
+        if (!isAvailable(in0)) completeStage()
       }
 
     })#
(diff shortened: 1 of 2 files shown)


akka-stream/src/main/boilerplate/akka/stream/scaladsl/ZipWithApply.scala.template
```diff
@@ -41,10 +41,6 @@ class ZipWith1[[#A1#], O] (zipper: ([#A1#]) ⇒ O) extends GraphStage[FanInShape
         if (pending == ##0) pushAll()
       }
 
-      override def onUpstreamFinish(): Unit = {
-        if (!isAvailable(out) || !isAvailable(in0)) completeStage()
```
drewhk: yeah, pretty stupid.

drewhk: Yeah pretty stupid.

drewhk:
Explanation:

The Zip op is implemented in a fully synchronous fashion with minimal buffering. This implies a two phase operation:
 - phase one: downstream demand is not present, upstreams are not pulled
 - phase two: downstream demand is present, awaiting all upstreams

The original closing logic tried to parry a situation which simply cannot exists: it wanted to handle the case when all upstream ports have elements but there is not yet downstream demand. Since we only pull the upstreams (all of them) exactly when we get a downstream demand, this will never happen. 

In fact, any time an upstream close event happens it is safe to close Zip:
 - if there is no downstream demand then it is safe to close since the upstream ports have been not yet pulled, i.e. a full zipping will not be possible (since we just lost one port)
 - if there is downstream demand then the only problem would be if we close although all ports have buffered elements ready to be emitted as a touple. This cannot happen though since the last push would have already emitted that before.

drewhk:
There was just one problem with my logic up there: If an upstream complete is completed but is *also* having an element, then we must not complete. So the original logic wan't that totally stupid actually ;)

Issue #17924, opened 2015-07-07
sbt error during import of release-2.3-dev into idea


2beaucoup:
"Settings logger used after project was loaded."

```
[info] The new value will be used by no settings or tasks.
[info] Reapplying settings...
[warn] Java 8 installation has not been found. Java 8 tasks will not be run.
[info] Set current project to akka (in build file:/[...]/akka/)
[info] Defining akka/*:artifactPath
[info] The new value will be used by no settings or tasks.
[info] Reapplying settings...
java.lang.RuntimeException: Settings logger used after project was loaded.
	at scala.sys.package$.error(package.scala:27)
	at sbt.LogManager$$anon$1$$anonfun$slog$1.apply(LogManager.scala:111)
	at sbt.LogManager$$anon$1$$anonfun$slog$1.apply(LogManager.scala:111)
	at scala.Option.getOrElse(Option.scala:120)
	at sbt.LogManager$$anon$1.slog(LogManager.scala:111)
	at sbt.LogManager$$anon$1.log(LogManager.scala:116)
	at sbt.Logger$class.warn(Logger.scala:123)
	at sbt.LogManager$$anon$1.warn(LogManager.scala:110)
	at akka.AkkaBuild$$anonfun$defaultSettings$18$$anonfun$apply$305.apply(AkkaBuild.scala:1110)
	at akka.AkkaBuild$$anonfun$defaultSettings$18$$anonfun$apply$305.apply(AkkaBuild.scala:1110)
	at scala.Option.orElse(Option.scala:257)
	at akka.AkkaBuild$$anonfun$defaultSettings$18.apply(AkkaBuild.scala:1110)
	at akka.AkkaBuild$$anonfun$defaultSettings$18.apply(AkkaBuild.scala:1097)
	at scala.Function1$$anonfun$compose$1.apply(Function1.scala:47)
	at scala.Function1$$anonfun$compose$1.apply(Function1.scala:47)
	at sbt.EvaluateSettings$MixedNode.evaluate0(INode.scala:175)
	at sbt.EvaluateSettings$INode.evaluate(INode.scala:134)
	at sbt.EvaluateSettings$$anonfun$sbt$EvaluateSettings$$submitEvaluate$1.apply$mcV$sp(INode.scala:66)
	at sbt.EvaluateSettings.sbt$EvaluateSettings$$run0(INode.scala:75)
	at sbt.EvaluateSettings$$anon$3.run(INode.scala:71)
[...]

PR #18848, opened 2015-11-04
=str #17924: Make IntelliJ IDEA sbt import work again

drewhk:
This does not solve https://github.com/akka/akka/issues/17924 completely, since I don't know what the commented out line did. 

@jrudolph if that was a feature that you don't want to see gone, then please fix it as I have no clue what it does and why. SBT integration with IDEA is rather important and I don't want to use gen-idea again.

### DIFF
--- a/project/AkkaBuild.scala
+++ b/project/AkkaBuild.scala
@@ -1123,7 +1123,6 @@ object AkkaBuild extends Build {
         Try("/usr/libexec/java_home -v 1.8".!!).toOption.toSeq // OS/X method
         )
         .find(isJavaHome)
-        .orElse { sLog.value.warn("Java 8 installation has not been found. Java 8 tasks will not be run."); None }
         .map(dirName => file(dirName.trim))
     },
 


ktoso:
LGTM from my side.
I guess this is to speed up the build slightly.

drewhk:
This resurfaced while updating to IDEA 15 and the new project simply refused to create scala files for me. I didn't want to use gen-idea again though.

2beaucoup:
the original problem of the ticket was https://github.com/drewhk/akka/blob/wip-17924-make-intellij-import-work-drewhk/project/AkkaBuild.scala#L1127

drewhk:
Hm, you linked to line 1110 in the ticket, and that is what I commented out. I can try L1127.

2beaucoup:
This line seems to have moved in release-2.3-dev. Should have linked to a a version hash instead.

drewhk:
Ah I see. But commenting out made the import pass interestingly :)

drewhk:
will investigate this later then

2beaucoup:
Do you have `AKKA_BUILD_JAVA8_HOME` set or does one of the paths match on your machine?

drewhk:
no

drewhk:
Now it removes the *real* offending line.

drewhk:
And I fixed the Zip issue already here: https://github.com/akka/akka/pull/18904

Issue #18040, opened 2015-07-19
lots of logging and unhandled ResumeReading in akka-http


fommil:
I'm seeing these messages in the logs

```
14:26:16.012 DEBUG akka://ENSIME/system/IO-TCP/selectors/$a/6 a.a.LocalActorRef - unhandled message from Actor[akka://ENSIME/system/IO-TCP-STREAM/server-1-%2F127.0.0.1%3A0/$e#425259929]: ResumeReading
```

our vanilla use of streams here:

https://github.com/ensime/ensime-server/blob/master/server/src/main/scala/org/ensime/server/WebServer.scala

fommil:
actually, it's more noisy than I thought... a single

```
 curl -X POST -H 'content-type: application/json' --data '{"typehint":"ConnectionInfoReq"}' "http://localhost:$(cat .ensime_cache/http)/rpc"
```

will produce

```
20:27:45.124 DEBUG akka://ENSIME/system/IO-TCP/selectors/$a/0 a.i.TcpListener - New connection accepted
20:27:45.126 DEBUG akka://ENSIME/system/IO-TCP/selectors/$a/0 a.i.TcpListener - New connection accepted
20:27:45.147 DEBUG akka://ENSIME/system/IO-TCP/selectors/$a/0 a.i.TcpListener - New connection accepted
20:27:45.159 DEBUG akka://ENSIME/system/IO-TCP/selectors/$a/0 a.i.TcpListener - New connection accepted
20:28:04.510 DEBUG akka://ENSIME/system/IO-TCP/selectors/$a/11 a.a.LocalActorRef - unhandled message from Actor[akka://ENSIME/system/IO-TCP-STREAM/server-1-%2F127.0.0.1%3A0/$j#557891595]: ResumeReading
20:28:04.510 DEBUG akka://ENSIME/system/IO-TCP/selectors/$a/10 a.a.LocalActorRef - unhandled message from Actor[akka://ENSIME/system/IO-TCP-STREAM/server-1-%2F127.0.0.1%3A0/$i#-2088821442]: ResumeReading
20:28:04.510 DEBUG akka://ENSIME/system/IO-TCP/selectors/$a/9 a.a.LocalActorRef - unhandled message from Actor[akka://ENSIME/system/IO-TCP-STREAM/server-1-%2F127.0.0.1%3A0/$h#-2113035534]: ResumeReading
[...]

PR #18853, opened 2015-11-04
=str #18040 unhandled ResumeReading

agolubev:
issue was that it was trying to send ResumeReading message when it's in half closed state.
Almost all warnings (maybe except one) with "unhandled message ResumeReading" must gone from build logs

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/io/TcpConnectionStream.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/io/TcpConnectionStream.scala
@@ -70,15 +70,13 @@ private[akka] abstract class TcpStreamActor(val settings: ActorMaterializerSetti
 
     def handleRead: Receive = {
       case Received(data) ⇒
-        if (closed) connection ! ResumeReading
-        else {
+        if (closed) {
+          connection ! ResumeReading
+        } else {
           pendingElement = data
           readPump.pump()
         }
-      case ConfirmedClosed ⇒
-        cancel()
-        readPump.pump()
-      case PeerClosed ⇒
+      case ConfirmedClosed | PeerClosed ⇒
         cancel()
         readPump.pump()
     }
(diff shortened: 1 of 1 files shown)


akka-stream/src/main/scala/akka/stream/impl/io/TcpConnectionStream.scala
```diff
@@ -91,7 +89,6 @@ private[akka] abstract class TcpStreamActor(val settings: ActorMaterializerSetti
       if (!closed) {
         closed = true
         pendingElement = null
-        if (!tcpOutputs.isFlushed && (connection ne null)) connection ! ResumeReading
```
drewhk: this will stop draining from the input, which has the side-effect of potentially missing close events. The idea is that cancellation turns the TCP read side to automatic drain-and-discard to handle these cases.

agolubev: Let me look into this. The weird thing is that all test pass with this change.

agolubev: @drewhk, I thought about this a little bit. So here is short list:
1. One line was removed that is sending ResumeReading to Connection, causing next read cycle.
2. Method is doing closed=true so this is changing state (of TcpConnection) to drain-and-discard mode I believe
3. This change is not causing switching to another Receive (like actor.become(newReceive)) so we should not miss any close events (in TcpStreamActor that we get fromTcp Connection)
4. The only case I can imagine is that we are closing TcpStreamActor from stream side with expectation that will trigger reading from TcpConnection
5. this is the only place in tcpInputs where we care about tcpOutputs (isFlushed). So this sounds like if input is closing on Steam side and for some reason we haven't sent all data to TcpConnection - we need to Resume Reading. Sounds weired.

Anyway, if you can give steps to reproduce I will create unit test and will go from there

drewhk: I don't remember why that condition is there, but `isFlushed` means that the TCP write side has been fully closed with no writes/bytes pending.

drewhk:
@agolubev I would opt for closing this. The reason is that I am not exactly sure how this worked, and since I want to get rid of the current TCP implementation and replace it with a GraphStage based one I am not sure if we want to put extra effort into this.

agolubev:
Yeah, this issue will be visible from build logs anyway

Issue #18817, opened 2015-10-30
Finalize Java Api for GraphStage


2m:
Main problem as I can see so far is creating [In/Out handlers](https://github.com/akka/akka/blob/5f89029b7e60919bc9e5b325dbf379e4cdcc87e8/akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala#L783-L826) from Java as they are now traits with default implementations.

What should be done:
  Remove InHandler's GraphStageLogic var
  Remove OutHandler's GraphStageLogic var
  Make the callback methods on InHandler and OutHandler take a GraphStageLogic parameter
  Make GraphStageLogic.{ EagerTerminateInput, IgnoreTerminateInput, ConditionalTerminateInput, TotallyIgnorantInput, EagerTerminateOutput, IgnoreTerminateOutput, ConditionalTerminateOutput } objects wherever possible and final when not intended for user inheritance.

viktorklang:
@2m The fix sounds like it is to make them abstract classes (since they have vars etc I think that's the sanest solution anyway.

drewhk:
Hm, that makes it impossible to have the same class extend both :( (one
more object header per PushPullStage)

drewhk:
Strictly speaking, there is no reason for them to have default implementations other than convenience (the finish methods default to finishing the stage -- so the stage must be injected to the handlers, therefore the var). I.e. one option is to have the Java side suffer by always having to implement all callbacks ;)

viktorklang:
May I ask why they are separate types at all?

drewhk:
Because they encapsulate different set of callbacks. For an input port you have these events:
 - onPush()
 - onUpstreamFinish()
 - onUpstreamFailure()

and for an output port you have these events:
 - onPull()
 - onDownstreamFinish()

These interfaces collect these callbacks together. You can't use the same type for both ports (at least not without serious hackyness).

viktorklang:
@drewhk have you considered putting the `GraphStageLogic` as a parameter to the callbacks instead of using the var?

viktorklang:
Added the following to the description:

~~~txt
What should be done:
Remove InHandler's GraphStageLogic var
Remove OutHandler's GraphStageLogic var
Make the callback methods on InHandler and OutHandler take a GraphStageLogic parameter
[...]

PR #18861, opened 2015-11-05
=str #18817 add Java API for GraphStage In/OutHandler

drewhk:
Also added a simple stage in Java and a test for it.

### DIFF
--- a/akka-stream-tests/src/test/java/akka/stream/stage/JavaIdentityStage.java
+++ b/akka-stream-tests/src/test/java/akka/stream/stage/JavaIdentityStage.java
@@ -1,0 +1,49 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.stage;
+
+import akka.stream.Attributes;
+import akka.stream.FlowShape;
+import akka.stream.Inlet;
+import akka.stream.Outlet;
+
+public class JavaIdentityStage<T> extends GraphStage<FlowShape<T, T>> {
+  private Inlet<T> _in = Inlet.create("Identity.in");
+  private Outlet<T> _out = Outlet.create("Identity.out");
+  private FlowShape<T, T> _shape = FlowShape.of(_in, _out);
+
+  public Inlet<T> in() { return _in; }
+  public Outlet<T> out() { return _out; }
+
+
+  @Override
+  public GraphStageLogic createLogic(Attributes inheritedAttributes) {
+    return new GraphStageLogic(shape()) {
+
+      {
+        setHandler(in(), new AbstractInHandler() {
+          @Override
+          public void onPush() {
+            push(out(), grab(in()));
+          }
+
+        });
+
+        setHandler(out(), new AbstractOutHandler() {
+          @Override
+          public void onPull() {
+            pull(in());
+          }
+        });
+
+      }
+
+    };
+  }
+
+  @Override
+  public FlowShape<T, T> shape() {
+    return _shape;
+  }
+}
--- a/akka-stream-tests/src/test/java/akka/stream/stage/StageTest.java
+++ b/akka-stream-tests/src/test/java/akka/stream/stage/StageTest.java
@@ -1,0 +1,47 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.stage;
+
+import akka.stream.StreamTest;
+import akka.stream.javadsl.AkkaJUnitActorSystemResource;
+import akka.stream.javadsl.Sink;
+import akka.stream.javadsl.Source;
+import akka.stream.testkit.AkkaSpec;
+
+import org.junit.ClassRule;
+import org.junit.Ignore;
+import org.junit.Test;
+import static org.junit.Assert.assertEquals;
+import scala.concurrent.Await;
+import scala.concurrent.Future;
+import scala.concurrent.duration.Duration;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class StageTest extends StreamTest {
+  public StageTest() {
+    super(actorSystemResource);
+  }
+
+  @ClassRule
+  public static AkkaJUnitActorSystemResource actorSystemResource = new AkkaJUnitActorSystemResource("FlowTest",
+    AkkaSpec.testConf());
+
+  @Test
+  public void javaStageUsage() throws Exception {
+    final java.lang.Iterable<Integer> input = Arrays.asList(0, 1, 2, 3, 4, 5);
+    final Source<Integer, ?> ints = Source.from(input);
+    final JavaIdentityStage<Integer> identity = new JavaIdentityStage<Integer>();
+
+    final Future<List<Integer>> result =
+      ints
+        .via(identity)
+        .via(identity)
+        .grouped(1000)
+        .runWith(Sink.<List<Integer>>head(), materializer);
+
+    assertEquals(Arrays.asList(0, 1, 2, 3, 4, 5), Await.result(result, Duration.create(3, "seconds")));
+  }
+}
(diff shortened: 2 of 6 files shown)


akka-stream/src/main/scala/akka/stream/Shape.scala
```diff
@@ -45,7 +45,19 @@ sealed abstract class OutPort { self: Outlet[_] ⇒
  * express the internal structural hierarchy of stream topologies).
  */
 object Inlet {
-  def apply[T](toString: String): Inlet[T] = new Inlet[T](toString)
+  /**
+   * Scala API
+   *
+   * Creates a new Inlet with the given name.
+   */
+  def apply[T](name: String): Inlet[T] = new Inlet[T](toString)
```
viktorklang: toString?

viktorklang: Also, in the scaladoc/javadoc, I think it is important to describe the significance of the name (if any)

akka-stream/src/main/scala/akka/stream/Shape.scala
```diff
@@ -62,7 +74,20 @@ final class Inlet[T] private (override val toString: String) extends InPort {
...
  */
 object Outlet {
-  def apply[T](toString: String): Outlet[T] = new Outlet[T](toString)
+
+  /**
+   * Scala API
+   *
+   * Creates a new Outlet with the given name.
+   */
+  def apply[T](name: String): Outlet[T] = new Outlet[T](toString)
```
viktorklang: bug (toString is from object Outlet)

akka-stream/src/main/scala/akka/stream/Shape.scala
drewhk: Why toString? It is the name of the port. What does toString even mean for a parameter? It if course makes sense in the case class, but I don't think it is meaningful for the user of the API.

akka-stream/src/main/scala/akka/stream/Shape.scala
drewhk: yep, thans!

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -351,6 +386,11 @@ private[stream] final class GraphInterpreter(
     } else _Name
 
   /**
+   * INTERNAL API
+   */
+  private[stream] def nonNull = this
```
viktorklang: No return type?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -95,14 +95,15 @@ object GraphStageLogic {
    */
   class ConditionalTerminateInput(predicate: () ⇒ Boolean) extends InHandler {
     override def onPush(): Unit = ()
-    override def onUpstreamFinish(): Unit = if (predicate()) inOwnerStageLogic.completeStage()
+    override def onUpstreamFinish(): Unit =
```
viktorklang: Why isn't the GraphStageLogic a parameter to this method instead?

akka-stream/src/main/scala/akka/stream/Shape.scala
viktorklang: It's object Inlet's toString method you're invoking here

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: for posterity my argument from gitter:

> The point of the current proposal is to enable the common defaults to be shared and save their allocations. Passing in the stage would enable that but user code would still have to close over the stage as well in order to do anything interesting, therefore the cluttered API does not save allocations or object size for them. To my mind the syntax overhead does not pull is weight.

viktorklang: @rkuhn The nested classes approach seems to work just fine: https://github.com/akka/akka/pull/18862

drewhk: But you cannot preallocate certain handlers. I.e. that wastes memory, this one does not.

2m:
LGTM. But this PR should unignore these two Java tests aswell:

https://github.com/akka/akka/commit/162bd9b8e365b5e545bc9f555e079efb611cf20c#diff-3b98156e098f2c6b236ff08f1b9d0598R183
https://github.com/akka/akka/commit/162bd9b8e365b5e545bc9f555e079efb611cf20c#diff-98bf7f3f099ab230ec2d012082452b71R111

drewhk:
Do they work without modification?

drewhk:
@2m Those cannot be unignored as they don't pass. Can you please open a ticket about what needs to be done there? We will do it after M1.

akka-stream-tests/src/test/java/akka/stream/stage/JavaIdentityStage.java
```diff
@@ -0,0 +1,46 @@
+package akka.stream.stage;
```
ktoso: license header

akka-stream-tests/src/test/java/akka/stream/stage/StageTest.java
```diff
@@ -0,0 +1,44 @@
+package akka.stream.stage;
```
ktoso: license header?

2m:
Ok. I'll create a ticket.

akka-stream-tests/src/test/java/akka/stream/stage/JavaIdentityStage.java
```diff
@@ -0,0 +1,46 @@
...
+
+import akka.stream.Attributes;
+import akka.stream.FlowShape;
+import akka.stream.Inlet;
+import akka.stream.Outlet;
+
+public class JavaIdentityStage<T> extends GraphStage<FlowShape<T, T>> {
+  private Inlet<T> _in = Inlet.of("Identity.in");
```
ktoso: here `of` reads weirdly... create would sound better

akka-stream-tests/src/test/java/akka/stream/stage/JavaIdentityStage.java
```diff
@@ -0,0 +1,46 @@
...
+        });
+
+        setHandler(out(), new AbstractOutHandler() {
+          @Override
+          public void onPull() {
+            pull(in());
+          }
+        });
+
+      }
```
ktoso: looks good

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -222,6 +222,38 @@ private[stream] object GraphInterpreter {
...
+  /**
+   * INTERNAL API
+   */
+  private val _currentInterpreter = new ThreadLocal[Array[AnyRef]] {
+    /*
+     * Using an Object-array avoids holding on to the GraphInterpreter class
+     * when this accidentally leaks onto threads that are not stopped when this
+     * class should be unloaded.
+     */
+    override def initialValue = new Array(1)
```
ktoso: Sneaky, I get it - we avoid hitting the thread-local, but just touch the pulled out array after it's pulled out... nice.

akka-stream-tests/src/test/java/akka/stream/stage/JavaIdentityStage.java
drewhk: Well, it is also quite weird to have `create` and `of` depending which sounds better in a certain context. Let's have one, consistent Java creator name then.

ktoso:
LGTM AFAICS, license header would be good to add though.
Naming bikeshed `of/create` is not urgent.

akka-stream-tests/src/test/java/akka/stream/stage/JavaIdentityStage.java
ktoso: I prefer "what reads best" in general for Java. APIs like guava have been doing this for years - to make using Java libraries bearable, they need to sound "fluent". Maybe that's a discussion for the entire API though, not this PR :)

PR #18862, opened 2015-11-05
=str #18817 make InHandler and OutHandler member classes and add Java API

drewhk:
Alternative to https://github.com/akka/akka/pull/18861 using member types.

### DIFF
--- a/akka-stream-tests/src/test/java/akka/stream/stage/JavaIdentityStage.java
+++ b/akka-stream-tests/src/test/java/akka/stream/stage/JavaIdentityStage.java
@@ -1,0 +1,45 @@
+package akka.stream.stage;
+
+import akka.stream.Attributes;
+import akka.stream.FlowShape;
+import akka.stream.Inlet;
+import akka.stream.Outlet;
+
+public class JavaIdentityStage<T> extends GraphStage<FlowShape<T, T>> {
+  private Inlet<T> _in = Inlet.of("Identity.in");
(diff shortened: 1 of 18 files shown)


akka-stream-tests/src/test/java/akka/stream/stage/JavaIdentityStage.java
```diff
@@ -0,0 +1,45 @@
...
+  private Inlet<T> _in = Inlet.of("Identity.in");
+  private Outlet<T> _out = Outlet.of("Identity.out");
+  private FlowShape<T, T> _shape = FlowShape.of(_in, _out);
+
+  public Inlet<T> in() { return _in; }
+  public Outlet<T> out() { return _out; }
+
+
+  @Override
+  public GraphStageLogic createLogic(Attributes inheritedAttributes) {
```
drewhk: Even though it is Java, this stage looks almost as compact as the Scala version.

drewhk:
We can revisit this vesion later

akka-stream-tests/src/test/java/akka/stream/stage/JavaIdentityStage.java
rkuhn: There is no difference with #18861, though.

drewhk: Of course not. The difference is the lack of ThreadLocal, but that is internal.

PR #18295, opened 2015-08-22
+doc show warning if browsing outdated docs

ktoso:
Shows warning (and links to latest docs) when browsing "old" versions of docs.
Also, more **bold** warning if browsing 2.2 docs.

TODO:
- [x] needs better text
- [x] 2.2 warning should link to 2.3.12 right away
- [x] warnings for 2.1 and 2.0 as well
- [x] "close" button, to close the warning without timeout
- [x] some styling?
- [x] support streams versions
- [x] merge and release old docs with this impr

Preview of unsupported:

Preview of just old docs:

Maybe this one shoudl be yellow?

PR #18868, opened 2015-11-05
+doc #18295 add close button with 1day delay to reappear

ktoso:
Port of docs warnings to `release-2.3` in case we'll release a backports release of 2.3


ktoso:
Failure was `ProcessorChannelSpec`, so something we've removed since...

Issue #18870, opened 2015-11-06
InputStreamSink, OutputSreamSource - small optimizations


agolubev:
- remove redundant usage of Future along with getAsyncCallback
- move vars into classes (internal) that are using them (to avoid creating redundant ref objects).

PR #18871, opened 2015-11-06
=str #18870 optimizations of InputStreamSink, OutpuSreamSource

agolubev:


2m:
OK TO TEST

drewhk:
LGTM

viktorklang:
hat was the reason for invoking them in a Future-callback before?

agolubev:
I thought that callback.invoke is synchronous call, but it's not. The same was requested as result of code review for PR #18773. Hm, but cannot find exact comment for this.

2beaucoup:
The `detachedChunk.size` [here](https://github.com/agolubev/akka/blob/agolubev-%2318870-Refactoring_OutputStreamSource_InutStreamSink/akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSinkStage.scala#L138) may not be what we're looking for...

agolubev:
ahh, my bad, I'll fix it

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSinkStage.scala
```diff
@@ -33,34 +33,24 @@ private[akka] object InputStreamSinkStage {
...
-          case ReadElementAcknowledgement ⇒
-            sendPullIfAllowed()
-          case Close ⇒
-            completeStage()
+        getAsyncCallback {
+          case ReadElementAcknowledgement ⇒ sendPullIfAllowed()
+          case Close                      ⇒ completeStage()
         }
 
+      override def wakeUp(msg: AdapterToStageMessage): Unit = if (!isClosed(in)) callback.invoke(msg)
```
rkuhn: this is wrong because it evaluates `isClosed(in)` from a different execution context, and it is also not needed: simply pass `callback.invoke _` to the handler’s constructor

rkuhn:
Thanks again, @agolubev! LGTM after fixing.

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamSinkStage.scala
```diff
@@ -145,7 +140,8 @@ private[akka] class InputStreamAdapter(sharedBuffer: BlockingQueue[StreamToAdapt
   }
 
   private[this] def readBytes(a: Array[Byte], begin: Int, length: Int): Int = {
-    val availableInChunk = detachedChunk.size - skipBytes
+    require(detachedChunk.isInstanceOf[Some[ByteString]], "Chunk must be pulled from shared buffer")
```
johanandren: Why is this done with an instanceOf-check, would have expected .nonEmpty, is it an optimization?

hepin1989: I don't think this should be more optimizaed than if check,maybe just follow his heart.

johanandren:
LGTM

drewhk:
LGTM. Please rebase and I will merge it.

agolubev:
@drewhk, rebase done, thanks.

drewhk:
I will rebuild this, so it has a pass with the new fuzzer: https://github.com/akka/akka/pull/18904

drewhk:
PLS BUILD

PR #18873, opened 2015-11-06
=str small improvement in GraphInterpreter.currentInterpreter

rkuhn:
reduce the number of hash table lookups by a factor of three

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
@@ -248,12 +248,6 @@ private[stream] object GraphInterpreter {
   private[stream] def currentInterpreterOrNull: GraphInterpreter =
     _currentInterpreter.get()(0).asInstanceOf[GraphInterpreter]
 
-  /**
-   * INTERNAL API
-   */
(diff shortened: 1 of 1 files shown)


drewhk:
Sneaky. *Now* I see the purpose of that array!

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -518,7 +519,7 @@ private[stream] final class GraphInterpreter(
         eventsRemaining -= 1
       }
     } finally {
-      setCurrentInterpreter(previousInterpreter)
+      currentInterpreterHolder(0) = previousInterpreter
```
2m: After this change `setCurrentInterpreter`, `currentInterpreterOrNull` and `currentInterpreter` is not used anymore.

rkuhn: Oh, right, the setter is indeed not used anymore, but I anticipate that we will need both getters in the immediate future—when running interpreters within interpreters for fused stream-of-stream stages.

ktoso:
LGTM, very nice trick :-) Good to have Bjorn around to pitch those.

Issue #18839, opened 2015-11-03
SharedLevelDbJournal - simplify store initialization


talp-traiana:
In order to use the SharedLevelDbJournal, the setStore method has to be called. This has to be done after the actor system creation but before any PersistentActors are created. However, the documentation isn't very clear about this, and in any case it's just boilerplate. As such, I think this should be done automatically during the plugin initialization (based on config).

(see https://groups.google.com/forum/#!topic/akka-user/bJwhzGkwMI4 )

patriknw:
@talp-traiana I actually have the need to use a journal proxy for another journal than leveldb, so I also started thinking about this. AsyncWriteProxy is complicating things a lot, as far as I can see. I have instead tried to implement a journal proxy, that forwards at the journal message protocol level. A lot easier, and would work with any journal.

I can add the automatic discovery of the target journal to this utility also.

talp-traiana:
@patriknw Yeah, that sounds like it could simplify things quite a bit there. I've been going over how the LeveldbReadJournal works and whether it should work with a SharedLeveldbJournal, and the AsyncWriteProxy is making it hard to figure out...

Since I started working on this yesterday, I just want to make sure - there's no point in me continuing to work on this, right? Is there anything else I can help with?

patriknw:
Contributions are very much welcome, and we have plenty of issues that would be great if you can help out with. If you are interested in persistence, take a look at this list: https://github.com/akka/akka/issues?q=is%3Aopen+is%3Aissue+milestone%3A2.4.x+label%3At%3Apersistence

For the journal proxy, I think I'm rather close to a working solution. I'll create a pull request in a few days, and it would be great if you can review it to see that it meets your expectations. I'll ping you.

PR #18876, opened 2015-11-06
+per #18839 Add better journal proxy

patriknw:
Not ready yet. Missing docs and test.
- JournalProxy works on the journal and snapshot store
  message protocol level and is therefore much simpler
  than the previous AsyncWriteProxy approach
- JournalProxy will work out of the box with any journal,
  I have tried with Cassandra
- should in the end phase out AsyncWriteProxy and
  SharedLeveldbJournal, we need deprecation phase
- rewrote the InmemJournal to remove the dependency to
  AsyncWriteProxy, which I can't see why it is needed
  for the InmemJournal

### DIFF
--- a/akka-persistence/src/main/resources/reference.conf
+++ b/akka-persistence/src/main/resources/reference.conf
@@ -243,6 +243,34 @@ akka.persistence.journal.leveldb-shared {
     }
 }
 
+akka.persistence.journal.proxy {
+  # Class name of the plugin.
+  class = "akka.persistence.journal.JournalProxy"
+  # Dispatcher for the plugin actor.
+  plugin-dispatcher = "akka.actor.default-dispatcher"
+  # Set this to on in the configuration of the ActorSystem
+  # that will host the target journal
+  start-target-journal = off
+  # The journal plugin config path to use for the target journal
+  target-journal-plugin = ""
+  # Initialization timeout of target lookup
+  init-timeout = 10s
+}
+
+akka.persistence.snapshot-store.proxy {
+  # Class name of the plugin.
+  class = "akka.persistence.journal.JournalProxy"
+  # Dispatcher for the plugin actor.
+  plugin-dispatcher = "akka.actor.default-dispatcher"
+  # Set this to on in the configuration of the ActorSystem
+  # that will host the target snapshot-store
+  start-target-snapshot-store = off
+  # The journal plugin config path to use for the target snapshot-store
+  target-snapshot-store-plugin = ""
+  # Initialization timeout of target lookup
+  init-timeout = 10s
+}
+
 # LevelDB persistence requires the following dependency declarations:
 #
 # SBT:
--- a/akka-persistence/src/main/scala/akka/persistence/journal/JournalProxy.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/journal/JournalProxy.scala
@@ -1,0 +1,181 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.persistence.journal
+
+import akka.util.Helpers.Requiring
+import scala.concurrent.duration._
+import akka.actor.Actor
+import akka.actor.Stash
+import scala.concurrent.duration.FiniteDuration
+import akka.actor.ActorRef
+import akka.persistence.JournalProtocol
+import akka.actor.ActorSystem
+import akka.persistence.Persistence
+import scala.util.control.NoStackTrace
+import java.util.concurrent.TimeoutException
+import akka.persistence.AtomicWrite
+import akka.persistence.NonPersistentRepr
+import akka.persistence.DeleteMessagesFailure
+import akka.actor.ActorLogging
+import com.typesafe.config.Config
+import akka.actor.Address
+import akka.actor.ActorIdentity
+import akka.actor.RootActorPath
+import akka.actor.Identify
+import akka.actor.ReceiveTimeout
+import akka.actor.ExtendedActorSystem
+import akka.persistence.SaveSnapshotFailure
+import akka.persistence.DeleteSnapshotFailure
+import akka.persistence.DeleteSnapshotsFailure
+import akka.persistence.SnapshotProtocol
+
+object JournalProxy {
+  final case class TargetLocation(address: Address)
+  private case object InitTimeout
+
+  def setTargetLocation(system: ActorSystem, address: Address): Unit = {
+    Persistence(system).journalFor(null) ! TargetLocation(address)
+    if (system.settings.config.getString("akka.persistence.snapshot-store.plugin") != "")
+      Persistence(system).snapshotStoreFor(null) ! TargetLocation(address)
+  }
+
+  private sealed trait PluginType {
+    def qualifier: String
+  }
+  private case object Journal extends PluginType {
+    override def qualifier: String = "journal"
+  }
+  private case object SnapshotStore extends PluginType {
+    override def qualifier: String = "snapshot-store"
+  }
+}
+
+// FIXME document me
+final class JournalProxy(config: Config) extends Actor with Stash with ActorLogging {
+  import JournalProxy._
+  import JournalProtocol._
+  import SnapshotProtocol._
+
+  private val pluginId = self.path.name
+  private val pluginType: PluginType = pluginId match {
+    case "akka.persistence.journal.proxy"        ⇒ Journal
+    case "akka.persistence.snapshot-store.proxy" ⇒ SnapshotStore
+    case other ⇒
+      throw new IllegalArgumentException("Unknown plugin type: " + other)
+  }
+
+  private val timeout: FiniteDuration = config.getDuration("init-timeout", MILLISECONDS).millis
+  private val targetPluginId: String = {
+    val key = s"target-${pluginType.qualifier}-plugin"
+    config.getString(key).requiring(_ != "", s"$pluginId.$key must be defined")
+  }
+  private val startTarget: Boolean = config.getBoolean(s"start-target-${pluginType.qualifier}")
+
+  override def preStart(): Unit = {
+    if (startTarget) {
+      val target = pluginType match {
+        case Journal ⇒
+          log.info("Starting target journal [{}]", targetPluginId)
+          Persistence(context.system).journalFor(targetPluginId)
+        case SnapshotStore ⇒
+          log.info("Starting target snapshot-store [{}]", targetPluginId)
+          Persistence(context.system).snapshotStoreFor(targetPluginId)
+      }
+      context.become(active(target, targetAtThisNode = true))
+    } else {
+      context.system.scheduler.scheduleOnce(timeout, self, InitTimeout)(context.dispatcher)
+    }
+  }
+
+  private val selfAddress: Address =
+    context.system.asInstanceOf[ExtendedActorSystem].provider.getDefaultAddress
+
+  private def timeoutException() = new TimeoutException(s"Target ${pluginType.qualifier} not initialized. " +
+    "Use `JournalProxy.setTargetLocation`")
+
+  def receive = init
+
+  def init: Receive = {
+    case TargetLocation(address) ⇒
+      context.setReceiveTimeout(1.second) // for retries
+      context.become(identifying(address))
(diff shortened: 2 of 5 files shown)


akka-persistence/src/main/scala/akka/persistence/journal/JournalProxy.scala
```diff
@@ -0,0 +1,181 @@
...
+import akka.persistence.SaveSnapshotFailure
+import akka.persistence.DeleteSnapshotFailure
+import akka.persistence.DeleteSnapshotsFailure
+import akka.persistence.SnapshotProtocol
+
+object JournalProxy {
+  final case class TargetLocation(address: Address)
+  private case object InitTimeout
+
+  def setTargetLocation(system: ActorSystem, address: Address): Unit = {
```
patriknw: this is the only user api

akka-persistence/src/main/resources/reference.conf
```diff
@@ -243,6 +243,34 @@ akka.persistence.journal.leveldb-shared {
...
+akka.persistence.journal.proxy {
+  # Class name of the plugin.
+  class = "akka.persistence.journal.JournalProxy"
+  # Dispatcher for the plugin actor.
+  plugin-dispatcher = "akka.actor.default-dispatcher"
+  # Set this to on in the configuration of the ActorSystem
+  # that will host the target journal
+  start-target-journal = off
+  # The journal plugin config path to use for the target journal
+  target-journal-plugin = ""
```
patriknw: user must define this, e.g. `cassandra-journal` or `akka.persistence.journal.leveldb`

patriknw:
Refs #18839

ktoso:
Awesome, will review soon :)

patriknw:
@talp-traiana I was not able to complete this today. Docs and test are missing. It would be great if you have time to review anyway. If you want to help out, it would be great if you can add a test and update documentation (replace everything related to SharedLeveldbJournal with this one).

I have tried it with Cassandra, so it's not completely untested from my side.

akka-persistence/src/main/resources/reference.conf
```diff
@@ -243,6 +243,34 @@ akka.persistence.journal.leveldb-shared {
...
 }
 
+akka.persistence.journal.proxy {
+  # Class name of the plugin.
+  class = "akka.persistence.journal.JournalProxy"
+  # Dispatcher for the plugin actor.
+  plugin-dispatcher = "akka.actor.default-dispatcher"
+  # Set this to on in the configuration of the ActorSystem
+  # that will host the target journal
+  start-target-journal = off
```
patriknw: user must set this to `on` at one node

akka-persistence/src/main/scala/akka/persistence/journal/JournalProxy.scala
```diff
@@ -0,0 +1,181 @@
...
+      sendIdentify(address)
+  }: Receive).orElse(init)
+
+  def active(targetJournal: ActorRef, targetAtThisNode: Boolean): Receive = {
+    case TargetLocation(address) ⇒
+      if (targetAtThisNode && address != selfAddress)
+        becomeIdentifying(address)
+    case InitTimeout ⇒
+    case msg ⇒
+      targetJournal.forward(msg)
```
rkuhn: PersistentActor assumes that messages to the journal don’t get lost: shouldn’t this protocol be a bit more advanced, with resends etc.? The failure mode would currently be that the persistent actor simply hangs forever, right?

patriknw: Good point. I'll see if I can use ask instead. (we can't resend, since there is no deduplication)

Long term, we have another ticket that would solve this at the right place: https://github.com/akka/akka/issues/15303

It should probably also watch the target journal.

akka-persistence/src/main/scala/akka/persistence/journal/JournalProxy.scala
ktoso: :+1:

akka-persistence/src/main/resources/reference.conf
ktoso: maybe add this comment in the conf file ("e.g. akka.persistence.journal.leveldb"), so it's clear we want the path, not the class of the plugin

akka-persistence/src/main/resources/reference.conf
```diff
@@ -243,6 +243,34 @@ akka.persistence.journal.leveldb-shared {
...
+  # Dispatcher for the plugin actor.
+  plugin-dispatcher = "akka.actor.default-dispatcher"
+  # Set this to on in the configuration of the ActorSystem
+  # that will host the target snapshot-store
+  start-target-snapshot-store = off
+  # The journal plugin config path to use for the target snapshot-store
+  target-snapshot-store-plugin = ""
+  # Initialization timeout of target lookup
+  init-timeout = 10s
+}
```
ktoso: :+1:

akka-persistence/src/main/scala/akka/persistence/journal/JournalProxy.scala
```diff
@@ -0,0 +1,181 @@
...
+import akka.persistence.SnapshotProtocol
+
+object JournalProxy {
+  final case class TargetLocation(address: Address)
+  private case object InitTimeout
+
+  def setTargetLocation(system: ActorSystem, address: Address): Unit = {
+    Persistence(system).journalFor(null) ! TargetLocation(address)
+    if (system.settings.config.getString("akka.persistence.snapshot-store.plugin") != "")
+      Persistence(system).snapshotStoreFor(null) ! TargetLocation(address)
```
ktoso: so this sets both, snapshot store and journal to the same address? I guess this is indeed the most popular use case for this utility.
In theory could be split up, but let's keep as is - document what it sets on the methods scaladoc maybe

akka-persistence/src/main/scala/akka/persistence/journal/JournalProxy.scala
```diff
@@ -0,0 +1,181 @@
...
+    } else {
+      context.system.scheduler.scheduleOnce(timeout, self, InitTimeout)(context.dispatcher)
+    }
+  }
+
+  private val selfAddress: Address =
+    context.system.asInstanceOf[ExtendedActorSystem].provider.getDefaultAddress
+
+  private def timeoutException() = new TimeoutException(s"Target ${pluginType.qualifier} not initialized. " +
+    "Use `JournalProxy.setTargetLocation`")
```
ktoso: should include the timeout (`"after ... ms"`)

akka-persistence/src/main/scala/akka/persistence/journal/JournalProxy.scala
```diff
@@ -0,0 +1,181 @@
...
+
+  def receive = init
+
+  def init: Receive = {
+    case TargetLocation(address) ⇒
+      context.setReceiveTimeout(1.second) // for retries
```
ktoso: OK

akka-persistence/src/main/scala/akka/persistence/journal/JournalProxy.scala
```diff
@@ -0,0 +1,181 @@
...
+      context.become(identifying(address))
+    case InitTimeout ⇒
+      log.info("Initialization timeout, Use `JournalProxy.setTargetLocation`")
```
ktoso: log what the timeout was (10s?)

akka-persistence/src/main/scala/akka/persistence/journal/JournalProxy.scala
```diff
@@ -0,0 +1,181 @@
...
+    case req: SnapshotProtocol.Request ⇒ req match { // exhaustive match
+      case LoadSnapshot(persistenceId, criteria, toSequenceNr) ⇒
+        sender() ! LoadSnapshotResult(None, toSequenceNr)
+      case SaveSnapshot(metadata, snapshot) ⇒
+        sender() ! SaveSnapshotFailure(metadata, timeoutException)
+      case DeleteSnapshot(metadata) ⇒
+        sender() ! DeleteSnapshotFailure(metadata, timeoutException)
+      case DeleteSnapshots(persistenceId, criteria) ⇒
+        sender() ! DeleteSnapshotsFailure(criteria, timeoutException)
+    }
```
ktoso: nice; takes care of the "but my PersistentActor is hanging and does nothing" nicely.

akka-persistence/src/test/scala/akka/persistence/journal/SteppingInmemJournal.scala
```diff
@@ -97,7 +97,10 @@ final class SteppingInmemJournal extends InmemJournal {
       val promise = Promise[Try[Unit]]()
       val future = promise.future
       doOrEnqueue { () ⇒
-        promise.completeWith(super.asyncWriteMessages(Seq(message)).map(_.head))
+        promise.completeWith(super.asyncWriteMessages(Seq(message)).map {
+          case Nil       ⇒ AsyncWriteJournal.successUnit
+          case head :: _ ⇒ head
+        })
```
ktoso: :+1:

akka-persistence/src/main/scala/akka/persistence/journal/inmem/InmemJournal.scala
```diff
@@ -5,28 +5,45 @@
...
+import akka.persistence.AtomicWrite
 
 /**
  * INTERNAL API.
  *
  * In-memory journal for testing purposes only.
  */
-private[persistence] class InmemJournal extends AsyncWriteProxy {
-  import AsyncWriteProxy.SetStore
+private[persistence] class InmemJournal extends AsyncWriteJournal with InmemMessages {
```
ktoso: Much cleaner now, good

ktoso:
Looks good! Looking forward to it's polished version :)

talp-traiana:
@patriknw 
I went over the code and it looks good overall. Just a couple of comments:

In general, I find that akka-persistence sometimes uses the term journal to refer to a journal, and other times for journal+snapshots. This causes confusion regarding what functionality is relevant where (for example, does a persistence query plugin read from the snapshot store?). Because of this, I think maybe the name should be changed from JournalProxy, in order to make it clearer that it's a proxy for snapshots as well (PersistencePluginProxy?).
I think there should be a way to initialize the target location via configuration, without having to call setTargetLocation manually. It's not too complicated to implement (just check whether a certain config key exists and send the TargetLocation message to self), and it would make it possible to switch between using a proxy and not using a proxy without code changes.
In any case, I'm still trying to set up a sample project so I can actually test it (especially compatibility with persistence query plugins, although it looks like it should work). Once I get it up and running I hope I'll have some more feedback.

One last thing - I might work a little on the documentation, and was wondering if there was a style guide for it or something...

Tal

patriknw:
I agree about the naming. PersistencePluginProxy sounds good. It's for both journal and snapshots.

The config is also a good idea. An optional config property should do the trick. It must still be possible to leave that out and set the target after the fact (that is the typical problem for multi-node testing).

Regarding docs. No style guide really, try to look around and use similar style as other things (and improve over existing state). I was thinking about replacing this section: http://doc.akka.io/docs/akka/2.4.0/scala/persistence.html#Shared_LevelDB_journal

You are welcome to open a new pull request with this commit as a base. I have limited time, so I appreciate your help. Let me know if you have any questions.

talp-traiana:
@patriknw 
I opened a PR with documentation for this (18876), let me know what you think.
I was also thinking, if you'd like, I could also do the 2 changes we discussed (renaming the plugin, and adding configuration for setting the target).

patriknw:
That's great, please go ahead with the additional changes.

Perhaps you can take a stab on trying to use `ask` also so that all requests send back a reply, also when there is no reply from the target journal actor. (timeout) The tricky part with that is to keep track of how the `sender` and persistent actor ref is passed around.

talp-traiana:
Sure, I'll start on all those changes. Should I push them to the same PR as the docs, or create a new one?

patriknw:
push to same pr, use separate commits to simplify life for reviewers (if there are need for incremental reviews).
Before merge we will squash.

patriknw:
Superseded by https://github.com/akka/akka/pull/19008

PR #18877, opened 2015-11-06
=str #16927 preserve header order in body parts

### LINKED ISSUE #16927: Preserve header order in BodyPartParser
When parsing a Multipart in akka-http-core-experimental 1.0-M3, Akka's parser reverses the order of the headers of each BodyPart: the order of bodyPart.headers is the opposite of the order in the incoming bytestream. It should preserve the order.

The culprit appears to be akka.http.engine.parsing.BodyPartParser, where `parseHeaderLines` prepends the header onto the list with `h :: headers` (line 157), and the resulting `List[HttpHeader]` becomes BodyPartStart.headers. This is used unchanged by MultipartUnmarshallers (lines 81, 92) to build the body part.

I think BodyPartParser needs the same fix as HttpMessageParser had in #16485, which appends the headers to a ListBuffer instead of prepending them to a List.

ktoso:
Resolves #16927

Preserves order of headers inside the parts (as does the parser for top level headers).

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala
@@ -14,6 +14,8 @@ import akka.http.scaladsl.model._
 import akka.http.impl.util._
 import headers._
 
+import scala.collection.mutable.ListBuffer
+
 /**
  * INTERNAL API
  *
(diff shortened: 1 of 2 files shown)


ktoso:
// cc @sirthias @jrudolph

sirthias:
Yup. LGTM!

The only thing: I think the commit message should start with `=htp` rather than `!str`, right?

ktoso:
> The only thing: I think the commit message should start with =htp rather than !str, right?

True, it's not API breaking, it may break if someone depended on the order but not really a case for `!`, thanks :)

rkuhn:
LGTM

ktoso:
Fixed commit message and merged

kw217:
Thank you!

kswmsw:
Thank you!

ktoso:
You're welcome :) Will be in M2 (expected late this week)

Issue #18555, opened 2015-09-25
add `throttle` combinator


rkuhn:
Even with Source.zipWith there is still considerable boilerplate in defining a throttling stage, we should make that available cleanly.

drewhk:
You can use the new GraphStage for that. I would also recommend adding a keepalive-inject next to it (injects a predefined element if: for x seconds there was no traffic, and there is downstream demand).

drewhk:
maybe we should also document that the throttle can be efficiently combined with a buffer to add policies:
 - backpressure on violation
 - drop on violation
 - fail on violation

drewhk:
Also, if it was not clear, I mentioned the keepalive-idle-inject thing here because it provides a minimum rate, while throttle sets a maximum rate.

rkuhn:
Yes, that makes perfect sense, thansk!

agolubev:
keepAlive combinator is added in here #18735

PR #18887, opened 2015-11-09
=str #18555 add throttle combinator

agolubev:
Ref #18555

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/ThrottleMode.scala
+++ b/akka-stream/src/main/scala/akka/stream/ThrottleMode.scala
@@ -1,0 +1,28 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream
+
+/**
+ * Represents a mode that decides how to deal exceed rate for Throttle combinator
+ */
+sealed abstract class ThrottleMode
+
+object ThrottleMode {
+
+  /**
+   *  Tells throttle to make pauses before emitting messages to meet throttle rate
+   */
+  case object Shaping extends ThrottleMode
+
+  /**
+   * Makes throttle fail with exception when upstream is faster than throttle rate
+   */
+  case object Enforcing extends ThrottleMode
+
+}
+
+/**
+ * Exception that is thrown when rated controlled by stream is exceeded
+ */
+class RateExceededException(msg: String) extends RuntimeException(msg)
--- a/akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
@@ -1,0 +1,81 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.impl
+
+import akka.stream.ThrottleMode.{ Enforcing, Shaping }
+import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
+import akka.stream.stage._
+import akka.stream._
+
+import scala.concurrent.duration.{ FiniteDuration, _ }
+import scala.util.control.NonFatal
+
+/**
+ * INTERNAL API
+ */
+private[stream] class Throttle[T](cost: Int,
+                                  per: FiniteDuration,
+                                  maximumBurst: Int,
+                                  costCalculation: (T) ⇒ Int,
+                                  mode: ThrottleMode)
+  extends SimpleLinearGraphStage[T] {
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new TimerGraphStageLogic(shape) {
+    var willStop = false
+    var lastTokens: Long = 0
+    var previousTime: Long = 0
+
+    val speed = ((cost.toDouble / per.toMillis) * 1024 * 1024).toLong
+    val timerName: String = "ThrottleTimer"
+
+    var currentElement: Option[T] = None
+
+    setHandler(in, new InHandler {
+      val scaledMaximumBurst = scale(maximumBurst)
+
+      override def onUpstreamFinish(): Unit =
+        if (isAvailable(out) && isTimerActive(timerName)) willStop = true
+        else completeStage()
+
+      override def onPush(): Unit = {
+        val timeElapsed = now() - previousTime
+        val currentTokens = Math.min(timeElapsed * speed + lastTokens, scaledMaximumBurst)
+        val elem = grab(in)
+        val elementCost = scale(costCalculation(elem))
+        if (currentTokens < elementCost)
+          mode match {
+            case Shaping ⇒
+              currentElement = Some(elem)
+              scheduleOnce(timerName, ((elementCost - currentTokens) / speed).millis)
+            case Enforcing ⇒ failStage(new RateExceededException("Maximum throttle throughput exceeded"))
+          }
+        else {
+          lastTokens = currentTokens - elementCost
+          previousTime = now()
+          push(out, elem)
+        }
+      }
+    })
+
+    override protected def onTimer(key: Any): Unit = {
+      push(out, currentElement.get)
+      currentElement = None
+      previousTime = now()
+      lastTokens = 0
+      if (willStop) completeStage()
+    }
+
+    setHandler(out, new OutHandler {
+      override def onPull(): Unit = pull(in)
+    })
+
+    override def preStart(): Unit = previousTime = now()
+
+    private def now(): Long = System.currentTimeMillis()
+
+    private def scale(e: Int): Long = e.toLong << 20
+  }
+
+  override def toString = "Throttle"
+}
--- a/akka-stream/src/main/scala/akka/stream/impl/Timers.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/Timers.scala
@@ -1,3 +1,6 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
 package akka.stream.impl
 
 import java.util.concurrent.{ TimeUnit, TimeoutException }
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
@@ -1064,6 +1064,62 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
     new Flow(delegate.keepAlive(maxIdle, () ⇒ injectedElem.create()))
 
   /**
+   * Sends elements downstream with speed limited to `elements/per`. In other words, this stage set the maximum rate
+   * for emitting messages. This combinator works for streams where all elements have the same cost or length.
+   *
+   * Throttle implements the token bucket model. There is a bucket with a given token capacity (burst size or maximumBurst).
+   * Tokens drops into the bucket at a given rate and can be `spared` for later use up to bucket capacity
+   * to allow some burstyness. Whenever stream wants to send an element, it takes as many
+   * tokens from the bucket as number of elements. If there isn't any, throttle waits until the
+   * bucket accumulates enough tokens.
+   *
+   * Parameter `mode` manages behaviour when upstream is faster than throttle rate:
+   *  - [[akka.stream.ThrottleMode.Shaping]] makes pauses before emitting messages to meet throttle rate
+   *  - [[akka.stream.ThrottleMode.Enforcing]] fails with exception when upstream is faster than throttle rate
+   *
+   * '''Emits when''' upstream emits an element and configured time per each element elapsed
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def throttle(elements: Int, per: FiniteDuration, maximumBurst: Int,
+               mode: ThrottleMode): javadsl.Flow[In, Out, Mat] =
+    new Flow(delegate.throttle(elements, per, maximumBurst, mode))
+
+  /**
+   * Sends elements downstream with speed limited to `cost/per`. Cost is
+   * calculating for each element individually by calling `calculateCost` function.
+   * This combinator works for streams when elements have different cost(length).
+   * Streams of `ByteString` for example.
+   *
+   * Throttle implements the token bucket model. There is a bucket with a given token capacity (burst size or maximumBurst).
+   * Tokens drops into the bucket at a given rate and can be `spared` for later use up to bucket capacity
+   * to allow some burstyness. Whenever stream wants to send an element, it takes as many
+   * tokens from the bucket as element cost. If there isn't any, throttle waits until the
+   * bucket accumulates enough tokens. Elements that costs more than the allowed burst will be delayed proportionally
+   * to their cost minus available tokens, meeting the target rate.
+   *
+   * Parameter `mode` manages behaviour when upstream is faster than throttle rate:
+   *  - [[akka.stream.ThrottleMode.Shaping]] makes pauses before emitting messages to meet throttle rate
+   *  - [[akka.stream.ThrottleMode.Enforcing]] fails with exception when upstream is faster than throttle rate. Enforcing
+   *  cannot emit elements that cost more than the maximumBurst
+   *
+   * '''Emits when''' upstream emits an element and configured time per each element elapsed
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def throttle(cost: Int, per: FiniteDuration, maximumBurst: Int,
+               costCalculation: function.Function[Out, Integer], mode: ThrottleMode): javadsl.Flow[In, Out, Mat] =
+    new Flow(delegate.throttle(cost, per, maximumBurst, costCalculation.apply, mode))
+
+  /**
    * Delays the initial element by the specified duration.
    *
    * '''Emits when''' upstream emits an element if the initial delay already elapsed
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
@@ -8,9 +8,9 @@ import akka.stream.Attributes._
 import akka.stream._
 import akka.stream.impl.Stages.{ DirectProcessor, StageModule, SymbolicGraphStage }
 import akka.stream.impl.StreamLayout.{ EmptyModule, Module }
-import akka.stream.impl.fusing.{ DropWithin, GroupedWithin, TakeWithin, MapAsync, MapAsyncUnordered }
-import akka.stream.impl.{ ReactiveStreamsCompliance, ConstantFun, Stages, StreamLayout, Timers }
-import akka.stream.stage.AbstractStage.{ PushPullGraphStageWithMaterializedValue, PushPullGraphStage }
+import akka.stream.impl._
+import akka.stream.impl.fusing.{ DropWithin, GroupedWithin, MapAsync, MapAsyncUnordered, TakeWithin }
+import akka.stream.stage.AbstractStage.{ PushPullGraphStage, PushPullGraphStageWithMaterializedValue }
 import akka.stream.stage._
 import org.reactivestreams.{ Processor, Publisher, Subscriber, Subscription }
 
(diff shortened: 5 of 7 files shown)


akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -1064,6 +1064,29 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
...
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   *
+   * @see [[#buffer]], [[akka.stream.scaladsl.Source#tick]].
+   */
+  def throttle(initialDelay: FiniteDuration, interval: FiniteDuration): javadsl.Flow[In, Out, Mat] =
```
ktoso: I was veeeery tempted to implement this one and make it: `throttle(ops = 10, per = 1.second)`, I think that'd be awesome.
What do others think about such API?

ktoso:
OK TO TEST

ktoso:
Thanks @agolubev !

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
agolubev: @ktoso, does this mean 10 messages per one second?

agolubev: Hm, this ticket may be more interesting than it is now.

ktoso: > does this mean 10 messages per one second?

yes, at most 10 messages per second to be signalled downstream. Then overflow should be configurable somehow.

drewhk: hm. I would rather see a token bucket:

   throttle(elements = 10, per = 1.second, maximumBurst = 100)

And a sized version (for ByteStrings for example):

   throttle(cost = 10, per = 1.second, maximumBurst = 100, costCalculation: (T) => Int)

To implement this, actually no timer is required while there are tokens available. You need to track how many tokens remained from the previous round. So the pseudoalgorithm for an incoming message is something like:

     timeElapsed = currentTime - previousTime
     rate = (cost / time)
     currentTokens = min(timeElapsed * rate + lastTokens, maximumBurst)
     elementCost = costCalculation(currentElement) 
     if (elementCost > currentTokens) 
       schedule timer((elementCost - currentTokens) / rate)
    else 
       lastTokens = currentTokens - elementCost
       previousTime = now
       push(elem)
       pull()

     onTimer(elem) = 
       previousTime = now
       lastTokens = 0 // Since we scheduled to the time when the element consumes all tokens
       push(elem)
       pull()

This scheme uses a timer only when actual delaying is needed. It also allows bursts at highest speed before throttling kicks in (token bucket semantics) but allows elements larger than the burst itself (i.e. setting burst to 0 will revert it to delaying elements according to target rate).

I would also recommend internally to precalculate the token refresh rate (cost / time) in 1024th (or even 1024 * 1024th) fragments. I.e.:

    val rate = ((cost / time.toDouble) * 1024 * 1024).toLong
    elementCost = costCalculation(currentElement)  << 20

This gives more accurate timings because it keeps track of tokens with 1/1048576 resolution, therefore not accumulating as much clipping errors.

hepin1989: +1 for token bucket.in fact one of my frinds in suning is using this with akka to implement throttling.

drewhk: Also, I think as an advanced use-case, we should have a version that takes a mode:

    throttle(cost = 10, per = 1.second, maximumBurst = 100, costCalculation: (T) => Int, mode: RateLimitMode)

The modes can be:

   - Shaping: this is the backpressured variant. 
   - Enforcing: this variant errors if the limit is reached

The following should work:

     shaping.via(enforcing) should not fail (i.e. if user shapes its traffic according to SLA then server will not kick him out)

There are fun uses of this btw:

    // 256kBytes/sec on average, with a 256kByte allowed initial burst
    tcpBytes.throttle(cost = 256 * 1024, 1.second, burst = 256 * 1024(byteString) => byteString.size)
      .via(parser)
      .throttle(cost = 100, 1.minute, burst = 10) // and then allowing only 100 API calls/minute (with some burstyness allowed)

Which can work for large requests where we want to limit both the datarate and the count of requests.

agolubev: Great, thanks for ideas, will add this shortly.

agolubev:
@ktoso, stable version is pushed

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -0,0 +1,123 @@
...
+import scala.util.control.NonFatal
+
+/**
+ * INTERNAL API
+ */
+private[stream] trait ThrottleTimeCalculationHelper {
+  def now(): Long = System.currentTimeMillis()
+
+  def calculateSpeed(counter: Double, time: FiniteDuration): Long =
+    ((counter / time.toMillis) * 1024 * 1024).toLong
```
drewhk: you probably want a floating-point division there...

drewhk: Ah counter is a Double...

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -0,0 +1,123 @@
...
+import akka.stream.stage._
+import akka.stream._
+
+import scala.concurrent.duration.{ FiniteDuration, _ }
+import scala.util.control.NonFatal
+
+/**
+ * INTERNAL API
+ */
+private[stream] trait ThrottleTimeCalculationHelper {
```
drewhk: why is this trait needed?

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -0,0 +1,123 @@
...
+    setHandler(in, new InHandler {
+      val scaledMaximumBurst = scale(maximumBurst)
+
+      override def onUpstreamFinish(): Unit =
+        if (isAvailable(out) && isTimerActive(timerName)) willStop = true
+        else completeStage()
+
+      override def onPush(): Unit = {
+        val timeElapsed = now() - previousTime
+        val speed = calculateSpeed(cost, per)
```
drewhk: You can extract this from the method so it is only calculated once, not per `onPush`

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -0,0 +1,123 @@
...
+ * INTERNAL API
+ */
+private[stream] class EqualCostElementsThrottle[T](elements: Int,
+                                                   per: FiniteDuration,
+                                                   maximumBurst: Int,
+                                                   mode: ThrottleMode)
+  extends AbstractThrottle[T](elements, per, maximumBurst, mode) {
+  override def calculateElementCost(elem: T): Long = scale(1)
+
+  override val timerName: String = "EqualCostElementsTimer"
```
drewhk: I don't think we need different timer names in the two subclasses.

drewhk: btw, why do we really need these subclasses?

akka-stream/src/main/scala/akka/stream/ThrottleMode.scala
```diff
@@ -0,0 +1,23 @@
...
+ * Represents a mode that decides how to deal exceed rate for Throttle combinator
+ */
+sealed abstract class ThrottleMode
+
+object ThrottleMode {
+
+  /**
+   *  Tells throttle to make pauses before emitting messages to meet throttle rate
+   */
+  case object ShapingMode extends ThrottleMode
```
drewhk: since it is in object `ThrottleMode` I would only call them `Shaping` and `Enforcing`

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -0,0 +1,123 @@
...
+                currentElement = Some(elem)
+                scheduleOnce(timerName, ((elementCost - currentTokens) / speed).millis)
+              case EnforcingMode ⇒ failStage(enforcedModeException)
+            }
+          else {
+            lastTokens = currentTokens - elementCost
+            previousTime = now()
+            push(out, elem)
+          }
+        } catch { case NonFatal(ex) ⇒ failStage(ex) } // for calculatingCost
```
drewhk: no need, the stage environment will automatically fail the stage on escaped exceptions.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -0,0 +1,123 @@
...
+      if (willStop) completeStage()
+    }
+
+    setHandler(out, new OutHandler {
+      override def onPull(): Unit = pull(in)
+    })
+
+    override def preStart(): Unit = previousTime = now()
+
+    override def keepGoingAfterAllPortsClosed: Boolean = true
```
drewhk: I don't think this is needed. This would mean that you would do something even after both the input and outputs have been closed. Since after the output has been cancelled there is really nothing this stage can do, this override should go.

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -1064,6 +1064,57 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
     new Flow(delegate.keepAlive(maxIdle, () ⇒ injectedElem.create()))
 
   /**
+   * Sends elements downstream with maximum speed of `elements/per`. In other words, this stage set the maximum rate
```
drewhk: Nitpick: it does not send with maximum speed, it limits the maximum speed.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
agolubev: just for having time conversion logic separate. Still can merge it to AbstractThrottle

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -1064,6 +1064,57 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
...
+   * for emitting messages. This combinator works for streams where all elements have the same cost or length.
+   *
+   * If there were no messages for sometime, throttle can emit `maximumBurst` number of elements at once, without
+   * waiting.
+   *
+   * Parameter `mode` manages behaviour when upstream is faster than throttle rate:
+   *  - [[akka.stream.ThrottleMode.ShapingMode]] makes pauses before emitting messages to meet throttle rate
+   *  - [[akka.stream.ThrottleMode.EnforcingMode]] fails with exception when upstream is faster than throttle rate. This
+   *  can be used for http servers to control SLA
```
drewhk: It is more like a quota than an SLA. But don't mention it here, HTTP quotas are more complex as most of the time there are multiple frontends which this stage does not handle (throttling is local to the stream it is applied to).

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -0,0 +1,123 @@
...
+private[stream] abstract class AbstractThrottle[T](cost: Int,
+                                                   per: FiniteDuration,
+                                                   maximumBurst: Int,
+                                                   mode: ThrottleMode)
+  extends SimpleLinearGraphStage[T] with ThrottleTimeCalculationHelper {
+
+  def calculateElementCost(elem: T): Long
+  val timerName: String
+
+  val enforcedModeException = new IllegalStateException("Maximum throttle throughput exceeded")
```
drewhk: this messes up the stacktrace. No need to cache this exception. Also, I would prefer a domain-specific exception for this: RateExceededException.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
agolubev: timer names was actually the only reason to introduce this along with some api simplification.

akka-stream/src/main/scala/akka/stream/ThrottleMode.scala
```diff
@@ -0,0 +1,28 @@
...
+   * Makes throttle fail with exception when upstream is faster than throttle rate
+   */
+  case object Enforcing extends ThrottleMode
+
+}
+
+/**
+ * Exception that is thrown when rated controlled by stream is exceeded
+ */
+class RateExceededException(mag: String) extends RuntimeException
```
ktoso: `mag` => `msg`
Message is lost here, should be:

```scala
final class RateExceededException(msg: String) extends RuntimeException(msg)
```

akka-stream/src/main/scala/akka/stream/impl/Timers.scala
```diff
@@ -1,3 +1,6 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
 package akka.stream.impl
```
ktoso: thanks!

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -0,0 +1,85 @@
...
+
+    var currentElement: Option[T] = None
+    def calculateElementCost(elem: T): Long = scale(costCalculation(elem))
+
+    setHandler(in, new InHandler {
+      val scaledMaximumBurst = scale(maximumBurst)
+
+      override def onUpstreamFinish(): Unit =
+        if (isAvailable(out) && isTimerActive(timerName)) willStop = true
+        else completeStage()
```
ktoso: won't this cause trouble when I'd `(1, per = 1.hour)`? It'll only propagate the completion event downstream after one hour has passed. I think we need to terminate eagerly once upstream finishes.

ktoso: Strike that, it's completion so we may have buffered elements... okey

drewhk: yes, if someone asked for 1 per hour, then they didn't ask for dropping it, but to delay it so it does not get faster than one per hour. People can add idle timeouts, completion timeouts, whatnot to control the overall time the flow takes.

ktoso: yeah and for dropping behaviour one can prefix with a dropping buffer.

ktoso:
I'll continue reviewing in a bit, I'm not completely sure on the semantics yet; in general looks very good though :)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1076,6 +1076,56 @@ trait FlowOps[+Out, +Mat] {
...
+   *
+   * '''Emits when''' upstream emits an element and configured time per each element elapsed
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def throttle(elements: Int, per: FiniteDuration, maximumBurst: Int,
```
ktoso: Tempted to say `maximumBurst = 0` by default?

drewhk: I don't think it is necessary. Having a small burst allowance reduces latency for short payloads because they won't be delayed. A `maximumBurst` of zero essentially means that every packet, including the first suffers a delay.

ktoso: I think I didn't grasp the exact behaviour of the burst yet... Seems a bit weird to me.
I understand throttle as "until I can signal elements during this 'time slot', keep pushing through, if not then don't.", burst does not fit my mental model here somehow; I'll need to dive into the code more here.

drewhk: It is the token bucket model. You have a bucket with a given capacity (burst size). Tokens drops into the bucket at a given rate (desired long term output rate). Whenever you want to send a packet, you take as many tokens from the bucket as the packets cost. If there isn't any, you wait until the bucket accumulates enough. The bucket starts full.

agolubev: maximumBurst is like how many messages you can send downstream at once in case a lot of time passed since last message went through

agolubev:
Interesting: 
- logic code to testing code ratio (lines of code) is 1/3 
- logic code to all other changes (in lines including docs, API, tests) 1/5

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -0,0 +1,85 @@
...
+
+    setHandler(out, new OutHandler {
+      override def onPull(): Unit = pull(in)
+    })
+
+    override def preStart(): Unit = previousTime = now()
+
+    def now(): Long = System.currentTimeMillis()
+
+    def calculateSpeed(counter: Double, time: FiniteDuration): Long =
```
drewhk: This is only used in one place, please inline it.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -0,0 +1,85 @@
...
+      if (willStop) completeStage()
+    }
+
+    setHandler(out, new OutHandler {
+      override def onPull(): Unit = pull(in)
+    })
+
+    override def preStart(): Unit = previousTime = now()
+
+    def now(): Long = System.currentTimeMillis()
```
drewhk: make this private.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -0,0 +1,85 @@
...
+
+    def calculateSpeed(counter: Double, time: FiniteDuration): Long =
+      ((counter / time.toMillis) * 1024 * 1024).toLong
+
+    def scale(e: Int): Long = e.toLong << 20
```
drewhk: private

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -1064,6 +1064,54 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
...
+   *
+   * '''Emits when''' upstream emits an element and configured time per each element elapsed
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def throttle(cost: Int, per: FiniteDuration, maximumBurst: Int,
```
drewhk: please also note that `Enforcing` cannot emit elements that cost more than the maximumBurst. I.e. if the maximumBurst is set to 0 it cannot work with Enforcing at all. (Maybe put a `require` to check this so it fails early).

drewhk:
Wonderful, LGTM! Minor nitpicks, but after that I merge.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
ktoso: Token bucket I get, thanks. maximum burst needs to be explained better I think, wording "at once" is confusing IMHO, it's like "how much can be sent down consuming accumulated tokens when for a long time no elements were arriving".

drewhk: we can explicitly say `tokenCapacity` and then explain that tokens can be "spared" for later use up to capacity to allow some burstyness. We should probably also explain in the cost version that elements costing more than the allowed burst will be delayed proportionally to their cost minus available tokens, meeting the target rate.

ktoso: :+1: yep, please add such docs and I'm happy with this impl :)

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
agolubev: ok, but initial idea was to have all calculations in one place to have to make it more readable.

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
agolubev: good point. will add this

agolubev: will add require that cost must also be > 0

drewhk: hm, zero cost might be still allowed

agolubev: in current implementation it will not work because of line 
```scheduleOnce(timerName, ((elementCost - currentTokens) / speed).millis)```

drewhk: but that is rate. I was talking about `cost`. Since ` if (currentTokens < elementCost)` will always evaluate to false if cost is zero, the element will be emitted immediately.

drewhk:
Thanks @agolubev !

agolubev:
Great! :smile:

Issue #16468, opened 2014-12-05
Enable (late?) programmatic access to `max-content-length` limit


huntc:
`akka.server.parsing.max-content-length` is not required when dealing with streams unless strict entities are required. For example, having this setting apply does not make sense when loading large files via multipart file data. The size of the file to consume in this instance becomes an application concerns, perhaps associated with the amount of disk space available.

sirthias:
Coming back to issue now (and aiming for a solution in the immediate future) I'd summarize the discussion
so far as follows:

1. We want to enable (late) programmatic definition of a length-limit for request entities,
   e.g. with something like ``request.withEntityLengthLimit(bytes)``.
2. By default the configured, rather aggressive (small) limit should be "attached" by the HTTP-layer.
3. The user should be able to increase or decrease the limit on their side.
4. The limit should apply to chunked requests just as well as unchunked ones.
5. A "limit-exceeded" error would be raised by error closing the request entity stream. It also means that
   the connection is burned for subsequent requests because we stop reading from the incoming side.
   After the response has been sent the connection must be (hard-)closed.

The requirements have these consequences:

1. We cannot immediately decide (in the HTTP layer) whether an incoming request has an exceedingly-large entity
   or not. We need to dispatch the request to the user in all cases because the user might apply a large enough
   limit after having received the request.
2. The best time for applying the configured or programmatically set limit appears to be the time the request
   entity stream is materialized. It can only be materialized once anyway and materialization happens on the
   user side, so the user has a chance to apply whatever limit before materialization.
3. If the entity is unchunked the decision whether the entity is small enough can be made right at
   materialization time (because we have the `Content-Length` header). For chunked requests we need to
   inject a counting stage which checks for length overrun as the entity is being consumed.

Implementation proposal:

The core problem to solve here is how to surface an adaptable limit to the user without introducing mutability
and without disrupting the established and otherwise quite clean entity model API.
And we need to uphold this property:

```scala
val requestA = HttpRequest(entity = HttpEntity.Default(20MB))
  .withEntityLengthLimit(10MB)
val requestB = requestA.withEntityLengthLimit(30MB)
requestB.entity.toStrict(...).get // ok
requestA.entity.toStrict(...).get // throws
```

I currently don't see a perfect solution here (maybe someone else does) but could imagine that we can introduce
one small constraint that is acceptable and allows for 98% of use cases to be fine:
[...]

PR #18888, opened 2015-11-09
+htp #16468 add `withSizeLimit` modifier to `HttpEntity`

sirthias:
This is how I thought #16468 might be solvable,
however I'm stuck due to my limited understanding of how the `StreamLayout.Module` mechanics work under the hood.

@rkuhn @drewhk @ktoso: Could one of you guys take a peek and let me know if the approach I took here is salvageable or will never fly no matter what?

Any tips on how to solve this best would be much appreciated.

/cc @jrudolph


sirthias:
I should probably also explain, what the problem with the current approach is:
Due to materialization using a `HashMap` with the Module as key my `RevertableModule` approach sometimes causes NoSuchElementExceptions in the process.
I guess that's why `Module` has `hashCode` and `equals` marked final.

rkuhn:
Sorry for the delay, I only now got around to reading this PR. This approach seems overly complicated to me, I was thinking more about creating a Source module that really just contains a `var` for the size limit and the stage logic accesses this variable during materialization to configure itself. Accessing this field would be a pain if the Source has already been transformed by earlier directives, but even that would be possible by walking the module tree—there can be only one such Source in there.

However, all of this looks terribly complicated and I am not convinced of the value of this feature. The principled solution would be that the HttpRequest does not contain a `Source[ByteString]` but more pieces that model this, but that would be too clunky for all those cases where it is not needed. Why not leave things like they are and only allow late narrowing? We could add a directive that just passes through requests and for all URLs not given as parameters it applies an additional limit.

ktoso:
We talked this one over and using attributes to relay the information seems
feesable and most likely simpler than the resettable behaviour. :)

rkuhn:
Hmm, that thought had crossed my mind as well, but I initially concluded that it would a rather dirty hack—now I’m not so sure anymore, it might actually work. We’ll see when it’s done :-)

sirthias:
The approach we discussed on Monday, using attributes to channel the bound to the limiting stage before materialization, appears to working well so far. Apart from actually doing what it's supposed to do it also has the benefit that later transformations on the source are no problem and we can get rid of the `Option` in the result of `withSizeLimit`. So overall much, much better.
Thanks, @drewhk, for the idea!

drewhk:
Glad to be of help. This is also the first step towards exploring the attributes feature which is currently a bit rough, but we have no data so far on the relevant use cases.

sirthias:
Ok, finally got this done in what I think is now a nice solution.
As mentioned before the attribute approach suggested by @drewhk worked out nicely. Only a few small tweaks were required in akka-stream to get things done.
Improvement ideas of course, and as always, much appreciated.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -107,7 +106,8 @@ private[http] object HttpServerBluePrint {
 
     val rendererPipeline =
       Flow[ResponseRenderingContext]
-        .via(Flow[ResponseRenderingContext].transform(() ⇒ new ErrorsTo500ResponseRecovery(log)).named("recover")) // FIXME: simplify after #16394 is closed
+        .via(Flow[ResponseRenderingContext].transform(() ⇒ new ErrorResponseRecovery(log,
+          settings)).named("recover")) // FIXME: simplify after #16394 is closed
```
2beaucoup: Looks like #16394 is fixed.

akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
```diff
@@ -72,4 +72,8 @@ object EntityStreamException {
   def apply(summary: String, detail: String = ""): EntityStreamException = apply(ErrorInfo(summary, detail))
 }
 
+case class EntityStreamSizeException(limit: Long, actualSize: Option[Long] = None) extends RuntimeException {
+  override def toString = s"EntityStreamSizeException($limit, $actualSize)"
```
2beaucoup: Is this needed?

ktoso:
Great, will review mid-day tomorrow, thanks!

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
```diff
@@ -67,9 +67,6 @@ sealed trait HttpMessage extends jm.HttpMessage {
   /** Returns a copy of this message with the list of headers transformed by the given function */
   def mapHeaders(f: immutable.Seq[HttpHeader] ⇒ immutable.Seq[HttpHeader]): Self = withHeaders(f(headers))
 
-  /** Returns a copy of this message with the entity transformed by the given function */
-  def mapEntity(f: HttpEntity ⇒ MessageEntity): Self = withEntity(f(entity))
```
2beaucoup: Why was this removed?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -354,4 +389,51 @@ object HttpEntity {
...
   object LastChunk extends LastChunk("", Nil)
+
+  /**
+   * Turns the given source into one that respects the `withSizeLimit` calls when used as a parameter
+   * to entity constructors.
+   */
+  def limitable[Out: SizeOfInBytes, Mat](source: Source[Out, Mat]): Source[Out, Mat] =
+    source.transform { () ⇒
+      new PushStage[Out, Out] {
+        val sizeOf = implicitly[SizeOfInBytes[Out]]
```
ktoso: `implicit ` in param list for better discoverability that it's using implicits?

sirthias: Does the context-bound on `Out` suffice?
Or maybe context-bounds are against the Akka style guide? (I don't quite remember anything here, but maybe things have changed as well...)

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
sirthias: Because it clashes with the `mapEntity` methods on `HttpRequest` and `HttpResponse` and wasn't usable anyway.
Actually I wasn't even sure how things compiled here in the first place.

akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
sirthias: Unfortunately yes. The compiler doesn't add a `toString` because `RuntimeException` already comes with a non-default implementation.

```scala
scala> case class Foo(i: Int) extends RuntimeException
defined class Foo

scala> Foo(42)
res0: Foo = Foo
```

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
sirthias: Ah, yes. Good catch.
I'll apply the simplification then as well, in a separate commit.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
drewhk: given that similar implicit magics have been removed recently, I opt against it.

drewhk:
LGTM, looks cool! I recommend one more test, namely, when the user attaches more attributes of his own it should not affect the Http sizeLimit attribute. You can do that by calling .named() for example, since that attaches a name attribute.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
sirthias: Ok, I'll simplify here.

sirthias:
@drewhk Yup, good idea. I'll add the test.

akka-http-core/src/main/scala/akka/http/scaladsl/model/ErrorInfo.scala
2beaucoup: I see. Thx for clarifying.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -74,6 +74,23 @@ sealed trait HttpEntity extends jm.HttpEntity {
    */
   def withContentType(contentType: ContentType): HttpEntity
 
+  /**
+   * Apply the given size limit to this entity by returning a new entity instance which automatically verifies that the
+   * data stream encapsulated by this instance produces at most ``maxBytes`` data bytes. In case this verification fails
+   * the respective stream will be terminated with an ``EntityStreamException`` either directly at materialization
+   * time (if the Content-Length is known) or whenever more data bytes than allowed have been read.
+   *
+   * When called on ``Strict`` entities the method will return the entity itself if the length is within the bound,
```
ktoso: Just a single ` here, we're in scaladoc - not rst :)

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
ktoso: Yup, thanks!

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
sirthias: Right, I always mix up all those markup styles.
These tools should simply "do the right thing" ... :)

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -354,4 +389,51 @@ object HttpEntity {
...
+            case None ⇒
+          }
+
+        def onPush(elem: Out, ctx: stage.Context[Out]): stage.SyncDirective = {
+          bytesLeft -= sizeOf(elem)
+          if (bytesLeft >= 0) ctx.push(elem)
+          else ctx.fail(EntityStreamSizeException(maxBytes))
+        }
+      }
+    }
```
ktoso: perhaps add a `.named` to it?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
ktoso: good to know, thanks for buy-scouting here :)

akka-stream/src/main/scala/akka/stream/Attributes.scala
```diff
@@ -67,12 +81,25 @@ final case class Attributes(attributeList: List[Attributes.Attribute] = Nil) {
...
    * Get the last (most specific) attribute of a given type parameter T `Class` or subclass thereof.
    */
   def get[T <: Attribute: ClassTag] =
     getAttribute(classTag[T].runtimeClass.asInstanceOf[Class[T]])
 
   /**
+   * Get the first (least specific) attribute of a given type parameter T `Class` or subclass thereof.
+   */
+  def getFirst[T <: Attribute: ClassTag] =
+    getFirstAttribute(classTag[T].runtimeClass.asInstanceOf[Class[T]])
```
ktoso: ok

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -213,6 +236,9 @@ object HttpEntity {
     def withContentType(contentType: ContentType): Default =
       if (contentType == this.contentType) this else copy(contentType = contentType)
 
+    def withSizeLimit(maxBytes: Long): Default =
+      copy(data = data withAttributes Attributes(SizeLimit(maxBytes, Some(contentLength))))
```
jrudolph: Only checking the contentLength late means we may introduce strange effects if someone uses 

```scala
entity.withSizeLimit(abc).copy(contentLength = newCL, data = entity.data.transform(xyz))
```

What's the reason not to fail early if possible?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpMessage.scala
jrudolph: Yes,it would only be useful if you would be working on HttpMessage which seems to happen rarely enough.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
sirthias: Yes, good idea! Will do.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
sirthias: > What's the reason not to fail early if possible?

Because someone might want to raise the limit to a higher value later.
With this PR we want to give to user the ability to override previously defined limits as often as necessary, to higher or lower values, and only have the last defined limit be applied at materialization time.

jrudolph: Ok, but as shown in the line above this doesn't work if one changes the data after the limit. Relatedly, I also wonder how this feature plays together with any kinds of data transformations?

drewhk: It always controls the wire level cap if I understand it correctly. I.e. even if you attack a Sink.ignore, which essentially would be "safe" with arbitrary large data, it will still cap, because the affected stage is down below. I think this is actually a *good* thing, but needs to be documented.

sirthias: Yes, documentation is required.
If you transform the entity in a way that changes the Content-Length and then apply another limit then this new limit will be applied against the new Content-Length, which is the right thing to do IMO.
If you transform the entity in a way that changes the Content-Length and *don't* apply another limit then the *previous* limit (e.g. the configured one) will be applied against the *previous* Content-Length, which is also ok I think.
So, all in all I think this feature is actually behaving nicely overall.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
sirthias: Trying this I realized that there is currently no easy way to attach a name.
I'd have to package the transformation into Flow which I then apply a `named` to and link it in with `via`, which would result in even more intermediate overhead stages and allocations.
Would you deem the cost/benefit ratio in favor of still adding `named` then?

drewhk: > I'd have to package the transformation into Flow 

It is done anyway, since the above API is just sugar over that:

```
via(new PushPullGraphStage((attr) ⇒ mkStage(), Attributes.none))
```

sirthias: Yes, I know, but `PushPullGraphStage` is `private[stream]`.
Some more sugar is needed for users like us here, I think.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
ktoso: Current behaviour makes sense to me; let's document this explicitly on limited though (and also in reference docs).

akka-http-core/src/test/scala/akka/http/impl/engine/client/LowLevelOutgoingConnectionSpec.scala
```diff
@@ -346,14 +302,225 @@ class LowLevelOutgoingConnectionSpec extends AkkaSpec("akka.loggers = []\n akka.
...
+                entity.toStrict(100.millis).awaitResult(100.millis).data.utf8String shouldEqual entityBase.take(bytes)
+            }
+
+          def expectSizeErrorInEntityOfType[T <: HttpEntity: ClassTag](limit: Int, actualSize: Option[Long] = None) =
+            inside(response) {
+              case HttpResponse(_, _, entity: T, _) ⇒
+                def gatherBytes = entity.dataBytes.runFold(ByteString.empty)(_ ++ _).awaitResult(100.millis)
+                (the[Exception] thrownBy gatherBytes).getCause shouldEqual EntityStreamSizeException(limit, actualSize)
+            }
+        }
```
ktoso: very nice assertions btw :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -295,11 +295,24 @@ private[http] object HttpServerBluePrint {
 
     override def onUpstreamFailure(error: Throwable, ctx: Context[ResponseRenderingContext]) =
       error match {
+        case EntityStreamSizeException(limit, contentLength) ⇒
+          val status = StatusCodes.RequestEntityTooLarge
+          val summary = contentLength match {
+            case Some(cl) ⇒ s"Request Content-Length of $cl bytes exceeds the configured limit of $limit bytes"
+            case None     ⇒ s"Aggregated data length of request entity exceeds the configured limit of $limit bytes"
+          }
+          val info = ErrorInfo(summary, "Consider increasing the value of akka.http.server.parsing.max-content-length")
```
2beaucoup: This is only true if `limit` == `max-content-length` setting.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
drewhk: Why would you need that wrapper directly? What is missing?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
sirthias: Yes, that's right. It is meant as a helper for people who are surprised by the error.
For these the comment should be helpful. If I have already applied a custom limit myself I probably know how to change things... at least that's my reasoning... :)

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
sirthias: I'd like to attach a name to this transformation stage with as little overhead as possible.

drewhk: What overhead? Why not cache the flow then?

sirthias: I'm looking for a simple, low-overhead (both in terms of LOCs and runtime overhead) way to mark the transform stage with a name, that's all. Currently there doesn't appear to be one, unless I widen the `private[stream]` marker on `PushPullGraphStage` to `private[akka]`. If you think that that's the way to go I'm happy to make this change.

Otherwise I think I have to do something like

```scala
source.via(Flow[Out].transform { () ⇒
  new PushStage[Out, Out] {
    ...
  }
}.named("limit"))
```

which works but isn't really "nice", neither from a source nor from a runtime perspective (2 nested `via` constructs).

drewhk: What is wrong with two nested ´via`s? i don't get it. What you included as a snippet is the proper way to name a stage. Since every immediate step of a Flow is reusable, you need the .via() on the source, otherwise the name would appliy to the whole assembled Source. The first via (from transform) will not have any more overhead than creating a Flow direcly, since you start from a cached empty flow which will simply delegate to a new Flow. The second via, attaching to the Source has exactly the same cost as calling `.map` since that is implemented in terms of via as well.

sirthias: Ok, got it.
I'll do it as indicated above.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
ktoso: Definitely want to keep this msg like so :)

2beaucoup: Yeah, I guess it's OK as the actual applied limit is part of the summary. Folks will find out that `akka.http.server.parsing.max-content-length` isn't what they're looking for when it's different from `limit`.

sirthias:
Ok, applied all feedback so far and also added a new ref doc section.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -107,7 +106,7 @@ private[http] object HttpServerBluePrint {
 
     val rendererPipeline =
       Flow[ResponseRenderingContext]
-        .via(Flow[ResponseRenderingContext].transform(() ⇒ new ErrorsTo500ResponseRecovery(log)).named("recover")) // FIXME: simplify after #16394 is closed
+        .recover(errorResponseRecovery(log, settings))
```

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -95,6 +118,7 @@ sealed trait HttpEntity extends jm.HttpEntity {
 /* An entity that can be used for body parts */
 sealed trait BodyPartEntity extends HttpEntity with jm.BodyPartEntity {
   def withContentType(contentType: ContentType): BodyPartEntity
+  def withSizeLimit(maxBytes: Long): BodyPartEntity
```
ktoso: That's a user facing API method as well right? More rarely used perhaps but still.
Copying the comment from is a bit of a pain, so let's at least add a `See [[HttpEntity#withSizeLimit]] for detailed documentation.`?

ktoso: (doing last round of review right now and let's merge then, sorry this is dragging out so slow @sirthias )

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -115,12 +139,13 @@ sealed trait RequestEntity extends HttpEntity with jm.RequestEntity with Respons
  */
 sealed trait ResponseEntity extends HttpEntity with jm.ResponseEntity {
   def withContentType(contentType: ContentType): ResponseEntity
-
-  override def transformDataBytes(transformer: Flow[ByteString, ByteString, Any]): ResponseEntity
+  def withSizeLimit(maxBytes: Long): ResponseEntity
```
ktoso: See https://github.com/akka/akka/pull/18888/files#r44797512, let's link to the one with docs?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
sirthias: Ok, will do.

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -354,4 +395,57 @@ object HttpEntity {
...
+            case None ⇒
+          }
+
+        def onPush(elem: Out, ctx: stage.Context[Out]): stage.SyncDirective = {
+          bytesLeft -= sizeOf(elem)
+          if (bytesLeft >= 0) ctx.push(elem)
+          else ctx.fail(EntityStreamSizeException(maxBytes))
+        }
+      }
+    }.named("limit"))
```
ktoso: thanks :)

ktoso: Though technically `limitable`; The reason I'm picky here is because we're considering a `limit` op on Flow ops ( https://github.com/akka/akka/issues/18921 ), so the name "limit" used here would be a bit misleading (better fix now, than when adding limit IMHO)

sirthias: Ok, no problem.
(Let me know when you are through then I'll immediately push.)

ktoso:
I'm tempted to also have a directive which maps the request entity by modifying the limit, WDYT?

ktoso:
Otherwise LGTM!

sirthias:
Yes, a high-level directive around `withSizeLimit` will definitely be valuable.
I'd see this outside of this ticket/PR though, as it will require it's own docs, tests, etc.

ktoso:
Sounds good to me, let's merge this one first. Added some minor comments in general looks good!

sirthias:
Ok, updated again with all your feedback in.

ktoso:
Ok, I'll let validation do it's thing and merge (just in case, maybe javadoc freaks out or something) :+1: Thanks a lot!

ktoso:
Follow up ticket for the directive: https://github.com/akka/akka/issues/18929

ktoso:
Merged – have a nice weekend @sirthias :-)

drewhk:
Yaay! (even if the cheers comes a bit late)

PR #18889, opened 2015-11-09
fix includecode

aruediger:


ktoso:
Ouch, good catch - thanks @2beaucoup!

Issue #18878, opened 2015-11-06
Document gotcha about forgetting to use ~ in http routing dsl 


divijan:
I accidentally did something like 

    get {
      path("") {
      ...
      }
    }
    path("hello") {
    ...
    }

which compiled but didn't produce the result I expected, because I missed the '~' between the directives. I suggest this shouldn't compile because it would produce dumb but hard to track down bugs. As a side question, why does it compile and what are the semantics of such a construct if any? (note that if I put just a space between the brace and the next directive, I *will* get an error)

ktoso:
Hi @divijan and thanks for reporting.
I don't think this is something we can do something about without going into macros/compiler-plugins or other areas which we can't go into with Akka.

Short answer to "why?" is: "Because that's valid Scala syntax".

Let me minimise the example to explain the issue you faced:

```scala
get {
  path("") {
  ...
  }
} ~
path("hello") {
...
}
// is exactly:
get(path("")(...)).~(path("hello"))
```

This is exactly how Scala "sees" the code, this is the "right way".
The mistake you point out is equivalent to the following: 

```scala
get {
  path("") {
  ...
  }
} 
path("hello") {
...
}
// is exactly:
get(path("")(...));
path("hello")
```

Scala sees it as separate method calls, and yeah if you do:

```scala
val routes = get {
  path("") {
  ...
  }
} 
path("hello") {
...
}
// then it means exactly:

```scala
val routes = get(path("")(...));
path("hello")
```

There's no programmatic way (other than macros or compiler plugins or a completely separate linter tool, which we do not want to apply here (really!)). So for now I think we need to mark this as will not fix.

I do however understand your point and it's valuable feedback.
We actually have a tool at typesafe which is trying to address these things, but it is in it's early stages still.
I asked the team here: https://github.com/scala/scala-abide/issues/67 if it would be possible to write such check (which would make the thing not compile if ~ missing).

----
Forgot one point you're asking about, adding it here: The reason the newline matters is this:

```scala
val whoops = get {}
post {} // compiles but "shouldn't"
// translates to:
val whoops = get();
post();

// ---- 
val aha = get {} post {} // does not compile
// translates to:
val aha = get().post(); // does not compile, get's return type does not have a post method
```

This is because how parsing works in scala. In `whoops` it assumes "oh ok, end of method chain, I'll assign it to whoops", and in `aha` it thinks "OK, so you want to call a method called post... oh there isn't one!".

Hope this helps.
[...]

PR #18896, opened 2015-11-09
=doc #18878 Document gotcha about missing tilde

johanandren:
Refs #18878


ktoso:
Excellent :) LGTM

ktoso:
Note or warning... WDYT?

johanandren:
I think Note is enough, it's an annoyance rather than something dangerous

ktoso:
true, LGTM :)

PR #18897, opened 2015-11-09
Don't print parsing exception stackstraces to the console

### LINKED ISSUE #18825: HTTP parsing exceptions should not be printed to the console
see HttpHeaderSpec’s test output for Host, Referer and Set-Cookie

assigned to http:next for @sirthias and @jrudolph to verify that this will not spill into production logs (since I cannot find the code that does the printing—perhaps it is within parboiled?)

aruediger:
fixes #18825


ktoso:
Ah yes, excellent :) Thanks @2beaucoup!

ktoso:
FYI @sirthias @jrudolph so you can also remove it from parboiled2 itself.

PR #18903, opened 2015-11-10
Accept multipart boundary parameters with internal spaces

### LINKED ISSUE #18881: BodyPartParser rejects multipart boundary parameters with internal spaces
Attempting to unmarshal a multipart entity with a content type that has a boundary parameter with an internal space (e.g. "simple boundary")  results in an IllegalArgumentException. This appears to be because the BodyPartParser enforces a requirement that the boundary match a "boundaryCharNoSpace" pattern. However, this is incorrect per RFC2046 5.1.1 - spaces may appear anywhere in a boundary except the last character (and in fact "simple boundary" is given as an example of a valid boundary in that RFC).

aruediger:
fixes #18881

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala
@@ -31,8 +31,8 @@ private[http] final class BodyPartParser(defaultContentType: ContentType,
 
   require(boundary.nonEmpty, "'boundary' parameter of multipart Content-Type must be non-empty")
   require(boundary.charAt(boundary.length - 1) != ' ', "'boundary' parameter of multipart Content-Type must not end with a space char")
-  require(boundaryCharNoSpace matchesAll boundary,
-    s"'boundary' parameter of multipart Content-Type contains illegal character '${boundaryCharNoSpace.firstMismatch(boundary).get}'")
+  require(boundaryChar matchesAll boundary,
+    s"'boundary' parameter of multipart Content-Type contains illegal character '${boundaryChar.firstMismatch(boundary).get}'")
 
   sealed trait StateResult // phantom type for ensuring soundness of our parsing method setup
 
@@ -252,7 +252,7 @@ private[http] final class BodyPartParser(defaultContentType: ContentType,
 
 private[http] object BodyPartParser {
   // http://tools.ietf.org/html/rfc2046#section-5.1.1
-  val boundaryCharNoSpace = CharPredicate.Digit ++ CharPredicate.Alpha ++ "'()+_,-./:=?"
+  val boundaryChar = CharPredicate.Digit ++ CharPredicate.Alpha ++ "'()+_,-./:=? "
 
   private object BoundaryHeader extends HttpHeader {
     def name = ""
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/MultipartUnmarshallersSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/MultipartUnmarshallersSpec.scala
@@ -16,7 +16,7 @@ import akka.stream.scaladsl._
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.util.FastFuture._
 import akka.http.impl.util._
-import headers._
+import akka.http.scaladsl.model.headers._
 import MediaTypes._
 
 class MultipartUnmarshallersSpec extends FreeSpec with Matchers with BeforeAndAfterAll with ScalatestUtils {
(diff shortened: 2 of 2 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala
```diff
@@ -252,7 +252,7 @@ private[http] final class BodyPartParser(defaultContentType: ContentType,
 
 private[http] object BodyPartParser {
   // http://tools.ietf.org/html/rfc2046#section-5.1.1
-  val boundaryCharNoSpace = CharPredicate.Digit ++ CharPredicate.Alpha ++ "'()+_,-./:=?"
+  val boundaryChar = CharPredicate.Digit ++ CharPredicate.Alpha ++ "'()+_,-./:=? "
```
ktoso: Was this used in some other place as well or only for the boundary? (did we impact something else by this change?)

2beaucoup: it's exclusively used [here](https://github.com/akka/akka/pull/18903/files#diff-1e81ca1591fa3a9265369f2153fa2f93R35)

ktoso: Thanks for confirming!

akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/MultipartUnmarshallersSpec.scala
```diff
@@ -261,21 +279,24 @@ class MultipartUnmarshallersSpec extends FreeSpec with Matchers with BeforeAndAf
...
-      //            |
-      //            |test@there.com
-      //            |--XYZABC--""".stripMargin).as[MultipartFormData]
-      //        msg shouldEqual "Illegal multipart/form-data content: unnamed body part (no Content-Disposition header or no 'name' parameter)"
-      //      }
+      "multipartFormDataUnmarshaller should reject" - {
+        "illegal multipart content" in {
+          Await.result(
+            Unmarshal(HttpEntity(`multipart/form-data` withBoundary "XYZABC", "--noboundary--"))
+              .to[Multipart.FormData].failed, 1.second).getMessage shouldEqual "Unexpected end of multipart entity"
```
ktoso: slightly different but valid message I see

akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/MultipartUnmarshallersSpec.scala
```diff
@@ -261,21 +279,24 @@ class MultipartUnmarshallersSpec extends FreeSpec with Matchers with BeforeAndAf
               RawHeader("Content-Additional-1", "anything"),
               RawHeader("Content-Additional-2", "really-anything")))) // verifies order of headers is preserved
       }
-      // TODO: reactivate after multipart/form-data unmarshalling integrity verification is implemented
```
ktoso: Thanks!

ktoso:
OK TO TEST

ktoso:
LGTM

akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/MultipartUnmarshallersSpec.scala
2beaucoup: Yeah. Just came across these two and reenabled them as well. Looks like the current stage base multipart parser doesn't provide the "Missing start boundary" message anymore. Same for "unnamed body part (no Content-Disposition header or no 'name' parameter)" -> "multipart/form-data part must contain `Content-Disposition` header with `name` parameter". (/cc @sirthias)

2beaucoup:
Failure looks strange. This succeeds on my machine. Maybe I should save the reactivation for later...

ktoso:
Yeah, a bit unexpected, let's tackle that separately.

```scala
[info]     - illegal form-data content *** FAILED *** (13 milliseconds)
[info]       "[Unexpected end of multipart entity]" did not equal "[multipart/form-data part must contain `Content-Disposition` header with `name` parameter]" (MultipartUnmarshallersSpec.scala:296)
[info]       org.scalatest.exceptions.TestFailedException:
[info]       at org.scalatest.MatchersHelper$.newTestFailedException(MatchersHelper.scala:160)
[info]       at org.scalatest.Match
```

ktoso:
Can you comment the tests again and open a ticket for it? Thanks @2beaucoup !

2beaucoup:
done (#18908)

ktoso:
LGTM, will merge after validation

Issue #18798, opened 2015-10-27
Add fuzzing mode to the interpreter


drewhk:
The solution is to expose a configuration option `akka.stream.interpreter.fuzzing-mode` that defaults to off, and when it is on
 - the interpreter should ignore the given eventLimit parameter and always use *1* instead
 - on dequeue() internally, it should pick a random slot for the to be dequeued connection id, put it in the random slot, and dequeue the previous connection id from that random slot instead.
 - the ActorGraphInterpreter should sometimes suspend itself even when there is an external event to be processed

These three conditions together will ensure a great depth of fuzzing of both internal and external events. I think this will be safe, since order dependent events like complete-after-push are encoded in the state bitfield instead in the FIFO queue of connection IDs, but I might be wrong about this.

Some tests will need this setting off:
 - low level interpreter tests, mostly in `akka.stream.impl.fusing` which depend on ordering
 - benchmarks

drewhk:
at the same time, add a `event-limit-max` parameter and apply it in `ActorGraphInterpreter`

drewhk:
Preliminary results: only these tests failed for me:

     akka.stream.scaladsl.FlowConflateSpec
     akka.stream.scaladsl.FlowZipSpec
     akka.stream.scaladsl.GraphZipSpec
     akka.stream.scaladsl.FlowIdleInjectSpec

2m:
LGTM

PR #18904, opened 2015-11-10
+str #18798: Fuzzing mode for the interpreter

drewhk:
_PHEAR ME_

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/GraphInterpreterPortsSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/GraphInterpreterPortsSpec.scala
@@ -384,6 +384,8 @@ class GraphInterpreterPortsSpec extends AkkaSpec with GraphInterpreterSpecKit {
     "ignore pull while completing" in new PortTestSetup {
       out.complete()
       in.pull()
+      // While the pull event is not enqueued at this point, we should still report the state correctly
+      in.hasBeenPulled should be(true)
 
       stepAll()
 
@@ -648,6 +650,8 @@ class GraphInterpreterPortsSpec extends AkkaSpec with GraphInterpreterSpecKit {
 
       in.cancel()
       out.push(0)
+      // While the push event is not enqueued at this point, we should still report the state correctly
+      out.isAvailable should be(false)
 
       stepAll()
 
@@ -1066,6 +1070,7 @@ class GraphInterpreterPortsSpec extends AkkaSpec with GraphInterpreterSpecKit {
     "ignore pull while failing" in new PortTestSetup {
       out.fail(TE("test"))
       in.pull()
+      in.hasBeenPulled should be(true)
 
       stepAll()
 
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowExpandSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowExpandSpec.scala
@@ -7,8 +7,7 @@ import scala.concurrent.Await
 import scala.concurrent.duration._
 import scala.concurrent.forkjoin.ThreadLocalRandom
 
-import akka.stream.ActorMaterializer
-import akka.stream.ActorMaterializerSettings
+import akka.stream.{ ActorMaterializer, ActorMaterializerSettings }
 
 import akka.stream.testkit._
 
@@ -22,6 +21,9 @@ class FlowExpandSpec extends AkkaSpec {
   "Expand" must {
 
     "pass-through elements unchanged when there is no rate difference" in {
+      // Shadow the fuzzed materializer (see the ordering guarantee needed by the for loop below).
+      implicit val materializer = ActorMaterializer(settings.withFuzzing(false))
+
       val publisher = TestPublisher.probe[Int]()
       val subscriber = TestSubscriber.probe[Int]()
 
@@ -51,6 +53,9 @@ class FlowExpandSpec extends AkkaSpec {
       }
 
       publisher.sendNext(-42)
+
+      // The request below is otherwise in race with the above sendNext
+      subscriber.expectNoMsg(500.millis)
       subscriber.requestNext(-42)
 
       subscriber.cancel()
@@ -68,6 +73,9 @@ class FlowExpandSpec extends AkkaSpec {
 
       publisher.sendNext(2)
       publisher.sendComplete()
+
+      // The request below is otherwise in race with the above sendNext(2) (and completion)
+      subscriber.expectNoMsg(500.millis)
 
       subscriber.requestNext(2)
       subscriber.expectComplete()
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMapAsyncSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMapAsyncSpec.scala
@@ -247,7 +247,7 @@ class FlowMapAsyncSpec extends AkkaSpec with ScalaFutures {
       }
 
       try {
-        val N = 100000
+        val N = 10000
         Source(1 to N)
           .mapAsync(parallelism)(i ⇒ deferred())
           .runFold(0)((c, _) ⇒ c + 1)
--- a/akka-stream/src/main/resources/reference.conf
+++ b/akka-stream/src/main/resources/reference.conf
@@ -40,6 +40,17 @@ akka {
 
       # Maximum number of elements emitted in batch if downstream signals large demand
       output-burst-limit = 1000
+
+      debug {
+        # Enables the fuzzing mode which increases the chance of race conditions
+        # by aggressively reordering events and making certain operations more
+        # concurrent than usual.
+        # This setting is for testing purposes, NEVER enable this in a production
+        # environment!
+        # To get the best results, try combining this setting with a throughput
+        # of 1 on the corresponding dispatchers.
+        fuzzing-mode = off
+      }
     }
 
     # Fully qualified config path which holds the dispatcher configuration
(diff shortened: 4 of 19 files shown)


akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowExpandSpec.scala
```diff
@@ -22,6 +21,9 @@ class FlowExpandSpec extends AkkaSpec {
   "Expand" must {
 
     "pass-through elements unchanged when there is no rate difference" in {
+      // Shadow the fuzzed materializer (see the ordering guarantee needed by the for loop below).
+      implicit val materializer = ActorMaterializer(settings.withFuzzing(false))
```
viktorklang: Why is it a setting rather than a type of Materializer? Now we're requiring all Materializers to respect this setting, right?

drewhk:
@ktoso @sirthias @jrudolph Now you can also experience the FUN (dwarffortresswiki.org/index.php/Fun) by flipping a simple switch ;)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowExpandSpec.scala
drewhk: No, that is an ActorMaterializerSetting type.

viktorklang: Ah, gotcha! :)

akka-stream/src/main/resources/reference.conf
```diff
@@ -40,6 +40,17 @@ akka {
 
       # Maximum number of elements emitted in batch if downstream signals large demand
       output-burst-limit = 1000
+
+      interpreter {
```
viktorklang: Isn't "interpreter" an implementation detail/name?
I.e. is this a good name for a section in the official configuration.

akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
```diff
@@ -308,6 +313,9 @@ final class ActorMaterializerSettings(
     })
   }
 
+  def withFuzzing(enable: Boolean): ActorMaterializerSettings =
+    copy(fuzzingMode = enable)
+
   def withDebugLogging(enable: Boolean): ActorMaterializerSettings =
     copy(debugLogging = enable)
```
viktorklang: Could we make a quick boyscout and only copy if the new setting deviates from the current setting? (goes for all the `with` methods)

viktorklang:
Awesome PR @drewhk, I think this is going to be able to catch a lot of subtle bugs before they make it to a release :D

akka-stream/src/main/resources/reference.conf
drewhk: you are right, maybe it should be a "debug" section instead. (could be pulled up one section of course instead of debug, but I would like to isolate this section from prod settings more)

akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
drewhk: yep, of course.

drewhk:
There were 2 bugs caught already and 1 almost-bug (zip sometimes needed an extra pull round to signal completion)

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/GraphInterpreterPortsSpec.scala
```diff
@@ -384,6 +384,8 @@ class GraphInterpreterPortsSpec extends AkkaSpec with GraphInterpreterSpecKit {
     "ignore pull while completing" in new PortTestSetup {
       out.complete()
       in.pull()
+      // While the pull event is not enqueued at this point, we should still report the state correctly
+      in.hasBeenPulled should be(true)
```
drewhk: @rkuhn and I have thought that no more logic bugs are present in the interpreter ;)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMapAsyncSpec.scala
```diff
@@ -247,7 +247,7 @@ class FlowMapAsyncSpec extends AkkaSpec with ScalaFutures {
       }
 
       try {
-        val N = 100000
+        val N = 10000
```
drewhk: had to decrease otherwise it times out on Jenkins (takes ~5 seconds). I think with the fuzzer less iterations are enough to exercise this logic well, so I believe this is a safe modification.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/GraphInterpreterPortsSpec.scala
ktoso: :)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMapAsyncSpec.scala
ktoso: sounds OK, and there's nightlies as well.

akka-stream/src/main/resources/reference.conf
```diff
@@ -40,6 +40,17 @@ akka {
...
+      debug {
+        # Enables the fuzzing mode which increases the chance of race conditions
+        # by aggressively reordering events and making certain operations more
+        # concurrent than usual.
+        # This setting is for testing purposes, NEVER enable this in a production
+        # environment!
+        # To get the best results, try combining this setting with a throughput
+        # of 1 on the corresponding dispatchers.
+        fuzzing-mode = off
+      }
```
ktoso: :+1:

ktoso:
LGTM

akka-stream/src/main/scala/akka/stream/impl/Timers.scala
```diff
@@ -196,13 +196,18 @@ private[stream] object Timers {
...
 
       setHandler(out, new OutHandler {
         override def onPull(): Unit = {
           if (isAvailable(in)) {
             push(out, grab(in))
-            pull(in)
+            if (isClosed(in)) completeStage()
+            else pull(in)
           } else {
             if (nextDeadline.isOverdue()) {
```
rkuhn: please collapse with previous line to make the tristate behavior more obvious

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -633,8 +639,8 @@ private[stream] final class GraphInterpreter(
 
   private[stream] def pull(connection: Int): Unit = {
     val currentState = portStates(connection)
+    portStates(connection) = currentState ^ PullStartFlip
     if ((currentState & OutClosed) == 0) {
-      portStates(connection) = currentState ^ PullStartFlip
```
rkuhn: yep, dat’s right; it’s so obvious now ;-)

rkuhn:
LGTM

akka-stream/src/main/scala/akka/stream/impl/Timers.scala
drewhk: I am not sure. This is two main states (original input available or not) with two sub-cases. Of course this can be organized in various permutatons, so this is more about signaling intent. Why do you think a 3-state approach where one state have two sub-cases is better?

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: yeah, it is one of those visibility bugs :) Took a while to figure it out, even though I had the event trace of the failure. Btw, the reason why it was undetected is because it was usually happening like this:

 - since completion is usually reported early, the upstream boundary basically always finished with a push-then-complete
 - once the push and the complete was enqueued, the downstream pull could only arrive after the complete
 - this means, that push arrived first, resulting in a swallowed pull from buffer (because buffer is detached)
 - then complete arrived immediately, impossible to overtake by a downstream pull (because FIFO)
 - pull arrived after complete: either being ignored, or isHoldingUpstream returning false already

The reorder made it possible to observe the sequence of:
 - push-complete enqueued
 - pull enqueued
 - push arrives
 - pull arrives (reordered wrt complete)
 
and the rest of it is clear now. The nice thing about this fuzzer is that while we can still rely on the nice fairness properties of the FIFO queue, this fuzzer forces us to **not** rely it on being FIFO for the actual logics.

akka-stream/src/main/scala/akka/stream/impl/Timers.scala
rkuhn: Trying it on for size:
~~~scala
if (isAvailable(in)) {
  push(out, grab(in))
  if (isClosed(in)) completeStage()
  else pull(in)
} else if (nextDeadline.isOverdue()) {
  nextDeadline = Deadline.now + timeout
  push(out, inject())
} else scheduleOnce(IdleTimer, nextDeadline.timeLeft)
~~~

rkuhn: This is more organized according to the 3.5 possible outcomes and their conditions. But I agree that neither is a clear winner.

drewhk: Depends on what do you consider a substate of what. I have no strong opinion on either. Maybe we need a tie-breaker. @viktorklang @ktoso ?

viktorklang: I'd probably prefer:

if (isAvailable(in) {
} else { //deal with deadlines
}

drewhk: then it is decided.

drewhk:
refs #18798

Issue #18635, opened 2015-10-02
Migration guide: document shutdown() => terminate()


ktoso:
The migration guide is not covering `shutdown()`'s deprecation, came up on gitter: 

:point_up: [October 2, 2015 9:45 PM](https://gitter.im/akka/akka?at=560eded776931cb7479ec7bb)

PR #18907, opened 2015-11-10
=doc #18635 Added section about `ActorSystem.shutdown` being deprecated

johanandren:
Refs #18635

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/ActorSystem.scala
+++ b/akka-actor/src/main/scala/akka/actor/ActorSystem.scala
@@ -427,12 +427,17 @@ abstract class ActorSystem extends ActorRefFactory {
    * will recursively stop all its child actors, then the system guardian
    * (below which the logging actors reside) and the execute all registered
    * termination handlers (see [[ActorSystem#registerOnTermination]]).
+   * Be careful to not schedule any operations on completion of the returned future
+   * using the `dispatcher` of this actor system as it will have been shut down before the
+   * future completes.
    */
   def terminate(): Future[Terminated]
 
   /**
    * Returns a Future which will be completed after the ActorSystem has been terminated
-   * and termination hooks have been executed.
+   * and termination hooks have been executed. Be careful to not schedule any operations
+   * on the `dispatcher` of this actor system as it will have been shut down before this
+   * future completes.
    */
   def whenTerminated: Future[Terminated]
 
--- a/akka-docs/rst/project/migration-guide-2.3.x-2.4.x.rst
+++ b/akka-docs/rst/project/migration-guide-2.3.x-2.4.x.rst
@@ -646,3 +646,30 @@ Now:
 
 The Java API remains unchanged and has simply gained the 2nd overload which allows ``ActorSelection`` to be
 passed in directly (without converting to ``ActorPath``).
+
+
+Actor system shutdown
+---------------------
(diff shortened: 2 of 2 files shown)


drewhk:
LGTM

akka-docs/rst/project/migration-guide-2.3.x-2.4.x.rst
```diff
@@ -645,4 +645,17 @@ Now:
...
+Actor system shutdown
+---------------------
+``ActorSystem.shutdown``, ``ActorSystem.awaitTermination`` and ``ActorSystem.isTerminated`` has been
+deprecated in favor of ``ActorSystem.terminate`` and ``ActorSystem.whenTerminated```. Both returns a
+``Future[Terminated]`` value that will complete when the actor system has terminated.
+
+To get the same behavior as ``ActorSystem.awaitTermination`` block and wait for ``Future[Terminated]`` value
+with ``Await.result`` from the Scala standard library.
+
+    Await.result(system.whenTerminated, Duration.Inf)
```
ktoso: `terminate` also returns a future now right? Maybe show it instead:

```
  def terminate(): Future[Terminated]
```

Terminating is a one liner kind of now.

`whenTerminated` is more for when different parts of the app want to do stuff on shutdown. You can show it as well here :) Add a note to NOT use an `context.dispatcher` when `.map()` in on the Future, since that dispatcher may have been shut down already, since we're shutting down, right :)

// I see this note is missing from the scaladoc on terminate / whenTerminated - maybe add it there as well?

https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/actor/ActorSystem.scala#L431

johanandren: Good catch, I have actually seen someone doing exactly that recently. Will update Scaladoc as well!

akka-actor/src/main/scala/akka/actor/ActorSystem.scala
```diff
@@ -427,12 +427,17 @@ abstract class ActorSystem extends ActorRefFactory {
    * will recursively stop all its child actors, then the system guardian
    * (below which the logging actors reside) and the execute all registered
    * termination handlers (see [[ActorSystem#registerOnTermination]]).
+   * Be careful to not schedule any operations on completion of the returned future
+   * using the `dispatcher` of this actor system as it will have been shut down before the
+   * future completes.
```
ktoso: LGTM!

akka-docs/rst/project/migration-guide-2.3.x-2.4.x.rst
```diff
@@ -645,4 +645,34 @@ Now:
...
+To get the same behavior as ``ActorSystem.awaitTermination`` block and wait for ``Future[Terminated]`` value
+with ``Await.result`` from the Scala standard library.
+
+To trigger a termination and wait for it to complete:
+
+    Await.result(system.terminate(), Duration.Inf)
+
+To wait for termination that is triggered elsewhere:
+
+    Await.result(system.whenTerminated, Duration.Inf)
```
ktoso: Maybe 

```scala

Await.result(system.whenTerminated, 10.seconds)
```

We try to not show `Inf` anywhere

ktoso:
@johanandren last nitpick https://github.com/akka/akka/pull/18907/files#r44420781 and let's merge. Please squash and merge :)

Issue #18899, opened 2015-11-09
FAILED javadoc on jenkins in akka-contrib


ktoso:
```
[error] /localhome/jenkinsakka/workspace/akka-multi-node-nightly/akka-contrib/target/java/akka/contrib/pattern/ReliableProxy.java:41: warning: empty <code> tag
[error]  * <code></code>SubscribeTransitionCallBack<code></code> and <code></code>UnsubscribeTransitionCallBack<code></code>
[error]          ^
[error] /localhome/jenkinsakka/workspace/akka-multi-node-nightly/akka-contrib/target/java/akka/contrib/pattern/ReliableProxy.java:41: warning: empty <code> tag
[error]  * <code></code>SubscribeTransitionCallBack<code></code> and <code></code>UnsubscribeTransitionCallBack<code></code>
[error]                                                  ^
[error] /localhome/jenkinsakka/workspace/akka-multi-node-nightly/akka-contrib/target/java/akka/contrib/pattern/ReliableProxy.java:41: warning: empty <code> tag
[error]  * <code></code>SubscribeTransitionCallBack<code></code> and <code></code>UnsubscribeTransitionCallBack<code></code>
[error]                                                                    ^
[error] /localhome/jenkinsakka/workspace/akka-multi-node-nightly/akka-contrib/target/java/akka/contrib/pattern/ReliableProxy.java:41: warning: empty <code> tag
[error]  * <code></code>SubscribeTransitionCallBack<code></code> and <code></code>UnsubscribeTransitionCallBack<code></code>
[error]                                                                                                              ^
[error] /localhome/jenkinsakka/workspace/akka-multi-node-nightly/akka-contrib/target/java/akka/contrib/pattern/ReliableProxy.java:81: warning: empty <code> tag
[error]  * param:  targetPath is the <code></code>ActorPath<code></code> to the actor to which all messages will be forwarded.
[error]                                    ^
[error] /localhome/jenkinsakka/workspace/akka-multi-node-nightly/akka-contrib/target/java/akka/contrib/pattern/ReliableProxy.java:81: warning: empty <code> tag
[error]  * param:  targetPath is the <code></code>ActorPath<code></code> to the actor to which all messages will be forwarded.
[error]                                                          ^
[error] /localhome/jenkinsakka/workspace/akka-multi-node-nightly/akka-contrib/target/java/akka/contrib/pattern/ReliableProxy.java:82: warning: empty <code> tag
[...]

PR #18909, opened 2015-11-11
=doc Fixes for genjavadoc

kukido:
- removed double backticks
- removed javadoc link where the class is not available in Java source
- formatted documentation for proper sphinx syntax

### DIFF
--- a/akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
+++ b/akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
@@ -29,9 +29,9 @@ object TransparentExponentialBackoffSupervisor {
    *       maxBackoff,
    *       (random_delay_factor * calculated_backoff) + calculated_backoff)
    *   }}}
-   * @param decider an [[akka.actor.SupervisorStrategy.Decider]] to specify how the supervisor
+   * @param decider an `Decider` to specify how the supervisor
    *   should behave for different exceptions. If no cases are matched, the default decider of
-   *   [[akka.actor.Actor]] is used. When the [[akka.actor.SupervisorStrategy.Restart]] directive
+   *   [[akka.actor.Actor]] is used. When the `Restart` directive
    *   is returned by the decider, this supervisor will apply an exponential back off restart.
    */
   def propsWithDecider(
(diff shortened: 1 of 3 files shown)


akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
```diff
@@ -29,9 +29,9 @@ object TransparentExponentialBackoffSupervisor {
    *       maxBackoff,
    *       (random_delay_factor * calculated_backoff) + calculated_backoff)
    *   }}}
-   * @param decider an [[akka.actor.SupervisorStrategy.Decider]] to specify how the supervisor
+   * @param decider an `Decider` to specify how the supervisor
```
johanandren: without "akka" in front it becomes "a `Decider`" rather than 'an'

johanandren: Also, this way we loose navigation in the docs, I think I have used `[[a.b.Singleton$]]` previosly to make it work with both javadoc and scaladoc.

johanandren:
LGTM except for loosing the interdoc links, so maybe fix that?

johanandren:
Refs #18899 right?

ktoso:
OK TO TEST

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
ktoso: I agree with @johanandren here, but also want to merge this quickly to unstuck validation.

ktoso:
Resolves https://github.com/akka/akka/issues/18899, yup

ktoso:
I'll merge and fix what Johan noticed - need to un-stuck validation.

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
kukido: `$` did work for `Restart`, because there's a restart class. 
It did not work with `Decider`. Because `Decider` is just a type, it is converted into Java as is - `scala.PartialFunction<java.lang.Throwable, akka.actor.SupervisorStrategy.Directive> decider` so it cannot be found by javadoc.

kukido: Although with `$` javadoc does not throw an error, it does not pick up the object.

> When the SupervisorStrategy.Restart$ directive is returned by the decider, this supervisor will apply an exponential back off restart.

Meh.

johanandren: Shame. Sorry to make you waste your time on it!

Issue #18835, opened 2015-11-03
GraphStage does not call AsyncCallback method after got onComplete


agolubev:
this is use case from #17967 ticket.
See test case below:
```scala
  class TestSinkStage() extends SinkStage[Int, () ⇒ Future[Int]]("TestSink") {

    override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, () ⇒ Future[Int]) = {
      val buffer = new util.LinkedList[Int]()
      val logic = new GraphStageLogic(shape) {
        private val callback: AsyncCallback[Promise[Int]] = getAsyncCallback((p: Promise[Int]) ⇒ {
          p.success(buffer.pop())
          if (buffer.isEmpty) completeStage()
        })

        def pull(): Future[Int] = {
          val p = Promise[Int]()
          callback.invoke(p)
          p.future
        }

        override def preStart() = pull(in)

        setHandler(in, new InHandler {
          override def onPush(): Unit = {
            buffer.add(grab(in))
            if (buffer.size() < 3) pull(in)
          }

          override def onUpstreamFinish(): Unit = if (buffer.isEmpty()) completeStage()
        })
      }
      (logic, logic.pull)
    }
  }

  "GraphStage asyncCallback" must {
    "must work after onComplete received" in {
      val (probe, pull) = TestSource.probe[Int]
        .toMat(Sink.fromGraph(new TestSinkStage()))(Keep.both)
        .run()
      probe.sendNext(1)
      probe.sendNext(2)
      probe.sendComplete() //if comment out this line - test passes
      // test fails here as AsyncCallback never executes after received onComplete
      Await.result(pull(), 300.millis) shouldBe (1)
      Await.result(pull(), 300.millis) shouldBe (2)
    }
  }
```

rkuhn:
There is nothing we can do on the library side here: only your stage knows that it is not yet finished, but you call `completeStage()` anyway (in `onUpstreamFinish`).

drewhk:
Well, a GraphStage will receive the callbacks until all of its ports are closed when it is implicitly shut down. I was tempted to add an overridable parameter that allows keeping alive a stage even if there are no open ports, but that felt quite risky (easy to leak actors and stages).

rkuhn:
Also: closing all ports means that nothing can be done anymore afterwards—the incoming async notification cannot have an observable effect within the stream. I thought about this as well while working on GraphStageLogic and this was my conclusion.

[...]

PR #18914, opened 2015-11-11
+str #18835: Allow stages to stay alive even after all ports closed

drewhk:
Turns out it wasn't as problematic because since last time @rkuhn's refactoring that introduced a two-phase completion via the `afterStageHasRun` which is actually correctly handled with async channels as well. 

@rkuhn please double check if this is correct.
@agolubev this will unlock https://github.com/akka/akka/pull/18773 for you

refs #18835

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/KeepGoingStageSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/KeepGoingStageSpec.scala
@@ -1,0 +1,201 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.impl.fusing
+
+import akka.actor.{ NoSerializationVerificationNeeded, ActorRef }
+import akka.stream.scaladsl.{ Keep, Source, Sink }
+import akka.stream.{ Attributes, Inlet, SinkShape, ActorMaterializer }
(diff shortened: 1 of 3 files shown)


drewhk:
Hm, seems like some tests fail, interestingly.

drewhk:
Ok, I see why GraphStageLogicSpec fails. I forced now the stage to close earlier, although emits were still pending.

drewhk:
stand back, fix is on the way

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/KeepGoingStageSpec.scala
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        setHandler(shape.inlet, new InHandler {
+          override def onPush(): Unit = pull(shape.inlet)
+
+          // Ignore finish
+          override def onUpstreamFinish(): Unit = listener.foreach(_ ! UpstreamCompleted)
+        })
+
+        override def keepGoingAfterAllPortsClosed: Boolean = keepAlive
```
ktoso: bikeshed: `keepAliveAfter...` sounds a bit more like what it is, but we have `keepAlive` as method so to avoid confusion maybe keepGoing is OK... (thinking out loud, no need to change I guess)

ktoso:
LGTM

PR #18920, opened 2015-11-12
+rem #18353: Prune reliable deliver actors (fwd)

drewhk:
(cherry picked from commit 6643f56)


drewhk:
wot, forgot to check in my MiMa filters...

PR #18924, opened 2015-11-13
Show where connections are coming from in stream-io.rst

aruediger:


drewhk:
OK TO TEST

drewhk:
LGTM

2beaucoup:
added in `StreamTcpDocTest.java` as well.

2beaucoup:
Here's the sample adjusted so that it shows how to get the binding *and* handle each connection:
```scala
val connections: Source[IncomingConnection, Future[ServerBinding]] =
  Tcp().bind("127.0.0.1", 8888)

def connectionHandler(connection: IncomingConnection): Unit = {
  println(s"New connection from: ${connection.remoteAddress}")

  import akka.stream.io.Framing
  val echo = Flow[ByteString]
    .via(Framing.delimiter(
      ByteString("\n"),
      maximumFrameLength = 256,
      allowTruncation = true))
    .map(_.utf8String)
    .map(_ + "!!!\n")
    .map(ByteString(_))

  connection.handleWith(echo)
}

val binding = connections.toMat(Sink.foreach(connectionHandler))(Keep.left).run

binding.map { b =>
  b.unbind() onComplete {
    case _ => // ...
  }
}
```
The current sample only shows how to achieve one of these tasks with `to(Sink.ignore).run()` + `runForeach` but the combination of these two isn't that obvious IMO.

ktoso:
LGTM, thanks!

2beaucoup:
hmm, ok.

ktoso:
Whoops, did you mean to include the snippet from the comment instead of the current sample? @2beaucoup

2beaucoup:
> The current sample only shows how to achieve one of these tasks with to(Sink.ignore).run() + runForeach but the combination of these two isn't that obvious IMO.

I just wanted to discuss if the provided sample could be an improvement to the current version.

ktoso:
I'm sorry I missed that intent somehow! I see what you mean with the example now – would be happy to accept the above proposed sample to bind it all together nicely. Can you submit it in another PR? Thanks in advance!

PR #18926, opened 2015-11-13
+htc #18925 Add parallelism param to IncConn.handleWithAsyncHandler

hseeberger:
Also fix some doco.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/javadsl/IncomingConnection.scala
+++ b/akka-http-core/src/main/scala/akka/http/javadsl/IncomingConnection.scala
@@ -42,15 +42,19 @@ class IncomingConnection private[http] (delegate: akka.http.scaladsl.Http.Incomi
 
   /**
    * Handles the connection with the given handler function.
-   * Returns the materialization result of the underlying flow materialization.
    */
   def handleWithSyncHandler(handler: Function[HttpRequest, HttpResponse], materializer: Materializer): Unit =
     delegate.handleWithSyncHandler(handler.apply(_).asInstanceOf[sm.HttpResponse])(materializer)
 
   /**
    * Handles the connection with the given handler function.
-   * Returns the materialization result of the underlying flow materialization.
    */
   def handleWithAsyncHandler(handler: Function[HttpRequest, Future[HttpResponse]], materializer: Materializer): Unit =
     delegate.handleWithAsyncHandler(handler.apply(_).asInstanceOf[Future[sm.HttpResponse]])(materializer)
+
+  /**
+   * Handles the connection with the given handler function.
+   */
+  def handleWithAsyncHandler(handler: Function[HttpRequest, Future[HttpResponse]], parallelism: Int, materializer: Materializer): Unit =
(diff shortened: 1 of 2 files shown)


hseeberger:
See 18925.

drewhk:
OK TO TEST

drewhk:
I am a bit worried about this. The parallelism here does not mean what a user might think it means. Maybe we should find a different name.

drewhk:
LGTM

2beaucoup:
prefix should be `+htp #18925` (or maybe `!`)

hseeberger:
@2beaucoup, why "htp"? And why should this be a breaking change?

drewhk:
Btw, shouldn't the Java API changed as well?

drewhk:
OK TO TEST

2beaucoup:
> @2beaucoup, why "htp"?

It's "htc" actually. (htp -> akka-http, htc -> akka-http-core)

> And why should this be a breaking change?

You are right. Should be safe.

hseeberger:
@2beaucoup, I followed the instructions under https://github.com/akka/akka/blob/master/CONTRIBUTING.md. Where are your prefixing rules documented?

hseeberger:
@drewhk, I knew you would ask for that ;-)
Will add the Java stuff. Sigh!

hseeberger:
OK, added Java stuff and changed the prefix to htc.

2beaucoup:
> Where are your prefixing rules documented?

Currently only in the commit history. ;)
Now we just have to get rid of that nasty double '#'...

akka-http-core/src/main/scala/akka/http/javadsl/IncomingConnection.scala
```diff
@@ -42,15 +42,19 @@ class IncomingConnection private[http] (delegate: akka.http.scaladsl.Http.Incomi
...
 
   /**
    * Handles the connection with the given handler function.
-   * Returns the materialization result of the underlying flow materialization.
    */
   def handleWithAsyncHandler(handler: Function[HttpRequest, Future[HttpResponse]], materializer: Materializer): Unit =
     delegate.handleWithAsyncHandler(handler.apply(_).asInstanceOf[Future[sm.HttpResponse]])(materializer)
+
+  /**
+   * Handles the connection with the given handler function.
```
2beaucoup: This should mention the new param.

hseeberger: The other params aren't documented neither.

drewhk: boy scout ;)

hseeberger:
OK, all unnecessary #s deleted.

ktoso:
LGTM, thanks Heiko

ktoso:
AFAR this is binary incompatible, but we don't mind during Milestones and there's other changing stuff anyway during the rush towards 2.0 ;-)

PR #18927, opened 2015-11-13
Fix version for stream and http

aruediger:


drewhk:
OK TO TEST

drewhk:
Thanks, LGTM

ktoso:
LGTM AFAICS, the "substitute version nr during release" works on obtaining the nr from sbt so this should work, right?

ktoso:
OK TO TEST

drewhk:
yes it should work.

Issue #18913, opened 2015-11-11
Move the "pretty duration" printer from akka.testkit.metrics.reporter to a main package


drewhk:

ktoso:
I'll eventually get to it, but it's a simple task someone may want to use to get into the codebase as first contribution as well :)

lolski:
@ktoso I'll do it

ktoso:
Great, thanks :)

PR #18931, opened 2015-11-14
moving PrettyDuration and PrettyDurationSpec to akka.util

lolski:
https://github.com/akka/akka/issues/18913

### DIFF
--- a/akka-actor/src/main/scala/akka/util/PrettyDuration.scala
+++ b/akka-actor/src/main/scala/akka/util/PrettyDuration.scala
@@ -1,7 +1,7 @@
 /**
  * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
  */
-package akka.testkit.metrics.reporter
+package akka.util
 
 import java.util.Locale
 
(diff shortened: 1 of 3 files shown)


johanandren:
OK TO TEST

johanandren:
OK TO TEST

johanandren:
PLS BUILD

akka-actor/src/main/scala/akka/util/PrettyDuration.scala
```diff
@@ -1,7 +1,7 @@
 /**
  * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
  */
-package akka.testkit.metrics.reporter
+package akka.util
```
patriknw: should we add a Java API also, since it's now public?

ktoso: :+1: a simple `def create(...)` will do (and document as `/** Java API */`)

patriknw: or?

```
object PrettyDuration {
   /** JAVA API */
   format(duration: Duration, includeNanos: Boolean, precision: Int): String =
```

ktoso: Good one, I concur.

lolski: Got it

lolski: Should we add a default argument, e.g. `def format(duration: Duration, includeNanos: Boolean = false, precision: Int = 4)`? In this example, both arguments are identical to the implementations of `PrettyPrintableDuration.pretty`

ktoso: It's Java API, can't have default arguments :-)
But yes, let's do an overload with `def format(duration) = format(duration, includeNanos = false, precision = 4)`

ktoso:
OK TO TEST

ktoso:
LGTM, please squash into 1 commit with message according to our guidelines: https://github.com/akka/akka/blob/master/CONTRIBUTING.md#creating-commits-and-writing-commit-messages
So in this case `+act #18913 moved PrettyDuration to akka-actor`

Thanks!

lolski:
Done!

ktoso:
LGTM, thanks; will merge after validation.

drewhk:
Failure seems to be unrelated

johanandren:
PLS BUILD

lolski:
Added a few lines of docs for the function as @patriknw suggested

ktoso:
Looks good to me, thanks!

Issue #18916, opened 2015-11-11
Move SynchronousFileSource to Source.file


ktoso:
As discussed with @viktorklang over dinner.
I agree, discoverability is dog and I'd like to move the file sink / source directly onto the types (and kill the implicit conversion - no one ever found it, and `SynchronousFileSource` sounds like leaking internal details).

Also, every 2nd hello world in akka streams is using a file.

drewhk:
+1

PR #18934, opened 2015-11-14
!str - 18916 - Source.file and Sink.file

viktorklang:
- Removes `Synchronous` from the names and descriptions of File I/O as it leaks impl details
- Removes the factories for FileSource and FileSink and puts them in Source and Sink respectively

For discussion, awaiting the creation of migration docs for M2

Related: https://github.com/akka/akka/issues/18916


akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -369,4 +373,17 @@ object Source {
...
+   * Creates a Source from a Files contents.
+   * Emitted elements are `chunkSize` sized [[akka.util.ByteString]] elements,
+   * except the final element, which will be up to `chunkSize` in size.
+   *
+   * You can configure the default dispatcher for this Source by changing the `akka.stream.blocking-io-dispatcher` or
+   * set it for a given Source by using [[ActorAttributes]].
+   *
+   * It materializes a [[Future]] containing the number of bytes read from the source file upon completion.
+   */
+  def file(f: File, chunkSize: Int = 8192): Source[ByteString, Future[Long]] =
```
ktoso: since we change API here in this PR, maybe let's move chunkSize into an attribute right away?

ktoso: (we wanted the default to be taken from config)

viktorklang: Or perhaps remove the default and force people to make a decision?

ktoso: I don't think taking away the default is helping... 8192 is a default used in JDK impls as well as other IO things in scala (i.e. infameous io.Source). When picking this value I benchmarked various settings and this is pretty good as a default I found (can look for benchmarks if you're interested, but then again - it's jsut a default)

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -287,8 +286,8 @@ object HttpEntity {
     def dataBytes: Source[ByteString, Any] = data
 
     /**
-      * See [[HttpEntity#withSizeLimit]].
```
drewhk: Note: IntelliJ Scala plugin changed comment formatting to this style, and I don't know how to change it back. I correct these by hand but @sirthias might have missed it. Can the scalariform plugin fix these, btw?

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
drewhk: Why an attribute? You want late control? Don't forget that attributes will apply to *all* enclused Sinks/Sources, so it might not be what you want.

drewhk:
LGTM, needs migration guide additions tho

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
viktorklang: I mean, either 8k is a good value in which case perhaps we should just have it non-configurable?

viktorklang:
@drewhk Added migration notes

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
ktoso: Yes, scalariform can fix those: https://github.com/daniel-trinh/scalariform#placescaladocasterisksbeneathsecondasterisk

Perhaps we should upgrade to the latest one and enable this setting btw (not really sure what's supported in the version we're on)

akka-stream/src/main/scala/akka/stream/io/Implicits.scala
```diff
@@ -29,20 +29,12 @@ object Implicits {
...
 
   // ---- Sinks ----
 
-  implicit final class AddSynchronousFileSink(val s: Sink.type) extends AnyVal {
-    def synchronousFile: SynchronousFileSink.type = SynchronousFileSink
-  }
-
   implicit final class AddOutputStreamSink(val s: Sink.type) extends AnyVal {
     def outputStream: OutputStreamSink.type = OutputStreamSink
   }
```
ktoso: create a ticket about removing those as well so we can get rid of `Implicits`?

viktorklang: Ok: https://github.com/akka/akka/issues/18943

ktoso: Thanks!

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
ktoso: I'm ok with keeping it here or dropping then, with a slight preference on keeping it in the API still.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -218,10 +222,22 @@ object Sink {
...
+   * Creates a Sink which writes incoming [[ByteString]] elements to the given file and either overwrites
+   * or appends to it.
+   *
+   * Materializes a [[Future]] that will be completed with the size of the file (in bytes) at the streams completion.
+   *
+   * This source is backed by an Actor which will use the dedicated `akka.stream.blocking-io-dispatcher`,
+   * unless configured otherwise by using [[ActorAttributes]].
+   */
+  def file(f: File, append: Boolean = false): Sink[ByteString, Future[Long]] =
+    new Sink(new FileSink(f, append, DefaultAttributes.fileSink, shape("FileSink")))
```
ktoso: if we want to keep the bufSize then it's missing here

akka-docs-dev/rst/java/migration-guide-1.0-2.x-java.rst
```diff
@@ -420,3 +420,21 @@ And use of query parameters from ``Uri`` that looked like this:
...
+
+Replace `SynchronousFileSource.create(` with `Source.file(`
+
+Replace `SynchronousFileSource` with `Source`.
+
+Replace `SynchronousFileSink.create(` with `Sink.file(`
+
+Replace `SynchronousFileSink.appendTo(f)` with `Sink.file(f, true)`
+
+Replace `SynchronousFileSink` with `Sink`
```
ktoso: When would one apply this?

// inconsistent use of `.` at end of sentences :)

ktoso:
Done reviewing here, had some minor nitpicks; Happy to merge once cleaned up and we know if we keep the buf size or not.

akka-docs-dev/rst/java/migration-guide-1.0-2.x-java.rst
viktorklang: if one has used SynchronousFileSink as a type in a method signature or as a val/var type

ktoso: This won't happen, users only touch the `SynchronousFileSink` object (factory) which returns `Sink`.
The class is internals, see here: https://github.com/akka/akka/pull/18934/files#diff-a5569b40beb4dc7fecd384b9d651c2a1L28 users won't have it in their code.

viktorklang: Fair enough :)

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
viktorklang: :D

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
viktorklang: I guess my point is that the performance will depend on the host OS and host HW as well as the size of the file, I gues, so putting the buf value in the code (for performance reasons).

I propose removing the option and have Akka be smarter about the sizing.

ktoso: Smarter how? One would have to "explore different buffer sizes and try to find which one works", but on a machine which is impacted by various factors that may be a) nondeterministic b) too slow to figure out the best size for not too large streams.

Giving the option in the param at least allows people who want to tweak it to tweak it. Not sure if it's hurting to keep it - do you mean because "one more param => more complexity"? It's 2 params so... but that bad yet :)

viktorklang: OTOH, it may be required to be able to reason about the max size "in flight". So I'll keep it.
But with that in mind, shouldn't we always require the user to specify it?

ktoso: > But with that in mind, shouldn't we always require the user to specify it?
:-1: because the hello world should be dead simple yet it's already quite complex: `Source.file(thing).via(Framing.delimiter("\n")).runForeach(println)`

viktorklang: @ktoso Alright, I'll keep the default hardcoded to 8192. Speaking of the demo tho, `Framing` is a terrible name, it's not really about framing, it's about parsing (IMO).

~~~scala
Source.file(thing).via(Parsing.newLines).runForeach(println)
~~~

ktoso: I really want `Framing.lines` or `Parsing.lines`, convince the guys please :wink: (why isn't if framing..? makes sense to me)

viktorklang: @ktoso As a developer, when have you used any code that has the word "framing" in it? (besides the case in point here)

viktorklang:
@ktoso This should be ready to go.

akka-docs-dev/rst/scala/migration-guide-1.0-2.x-scala.rst
```diff
@@ -450,3 +450,15 @@ And use of query parameters from ``Uri`` that looked like this:
...
+
+SynchronousFileSource and SynchronousFileSink
+============================================
+
+Both have been replaced by `Source.file(…)` and `Sink.file(…)`.
+
+Update procedure
+----------------
+
+Replace `SynchronousFileSource(` and `SynchronousFileSource.apply(` with `Source.file(`
```
drewhk: This needs code samples (other sections have those).

drewhk: Also, please add one sentence about the rationale (nothing fancy)

drewhk:
LGTM after minor tweaks on the migration guide.

viktorklang:
@ktoso Updated migration docs

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
ktoso: Not that often I guess ;) Let's tackle that separately

ktoso:
LGTM, thanks @viktorklang!

viktorklang:
@ktoso wait, I'll add a better description too

ktoso:
:+1:

viktorklang:
@ktoso done

viktorklang:
@ktoso Could you merge when it is green so I can tackle the InputStream/OutputStream ones?

ktoso:
That's the plan, yes

ktoso:
Merged, thanks @viktorklang

PR #18941, opened 2015-11-16
!htc allow configuration of cookie parsing mode, fixes #16828

### LINKED ISSUE #16828: Support partial parsing of cookie header
From spray/spray#737:

> Following up on the discussion in #702 I propose to improve the header parser for Cookie headers in the following way:
  * parse all cookies in the header value independently
  * create a Cookie header holding all cookies that have been successfully parsed
  * create a RawHeader("Cookie", ...) holding the raw values of all invalid cookies

The rationale behind this request is to make it easy to access particular cookies even in the presence of other, malformed cookies (which e.g. often happens during testing on localhost).

/cc @sirthias

jrudolph:
Here's the implementation of what was outlined in the discussion in #16828.

The new rule for parsing cookies is this: every cookie is treated separately and a cookie that cannot be parsed using the configured parsing mode is silently discarded.

There are now two modes that can be configured:
- rfc6265 which corresponds to the existing implementation
- raw which will accept almost everything (including any combination of whitespace, dquotes, commas, or unicode characters)

This enables two new things:
- you can use existing infrastructure to read your (well-formed, RFC 6265 compliant) cookie values regardless of the existence of other cookie values that would have previously prevented the cookie to be parsed at all
- you can relax parsing rules further to enable access to other cookies out there in the wild

Fixes #16828.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala
@@ -4,6 +4,8 @@
 
 package akka.http.impl.engine.parsing
 
+import akka.http.ParserSettings
+
 import scala.annotation.tailrec
 import akka.event.LoggingAdapter
 import akka.parboiled2.CharPredicate
@@ -274,7 +276,8 @@ private[http] object BodyPartParser {
     maxHeaderCount: Int,
     illegalHeaderWarnings: Boolean,
     headerValueCacheLimit: Int,
-    uriParsingMode: Uri.ParsingMode) extends HttpHeaderParser.Settings {
+    uriParsingMode: Uri.ParsingMode,
+    cookieParsingMode: ParserSettings.CookieParsingMode) extends HttpHeaderParser.Settings {
     require(maxHeaderNameLength > 0, "maxHeaderNameLength must be > 0")
     require(maxHeaderValueLength > 0, "maxHeaderValueLength must be > 0")
     require(maxHeaderCount > 0, "maxHeaderCount must be > 0")
@@ -289,5 +292,6 @@ private[http] object BodyPartParser {
     maxHeaderCount = 64,
     illegalHeaderWarnings = true,
     headerValueCacheLimit = 8,
-    uriParsingMode = Uri.ParsingMode.Relaxed)
+    uriParsingMode = Uri.ParsingMode.Relaxed,
+    cookieParsingMode = ParserSettings.CookieParsingMode.RFC6265)
 }
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/CharacterClasses.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/CharacterClasses.scala
@@ -60,7 +60,11 @@ private[http] object CharacterClasses {
   val `token68-start` = ALPHA ++ DIGIT ++ "-._~+/"
 
   // https://tools.ietf.org/html/rfc6265#section-4.1.1
-  val `cookie-octet` = CharPredicate('\u0021', '\u0023' to '\u002b', '\u002d' to '\u003a', '\u003c' to '\u005b', '\u005d' to '\u007e')
+  val `cookie-octet-rfc-6265` = CharPredicate('\u0021', '\u0023' to '\u002b', '\u002d' to '\u003a', '\u003c' to '\u005b', '\u005d' to '\u007e')
+  val `cookie-separator` = CharPredicate(akka.parboiled2.EOI, ';')
+  val `cookie-octet-raw` =
+    CharPredicate('\u0020' to '\u007e') ++
+      CharPredicate((x: Char) ⇒ x > 0x7f && java.lang.Character.isDefined(x)) -- `cookie-separator`
   val `av-octet` = CharPredicate('\u0020' to '\u003a', '\u003c' to '\u007e') // http://www.rfc-editor.org/errata_search.php?rfc=6265
 
   // http://tools.ietf.org/html/rfc5988#section-5
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
@@ -215,17 +215,35 @@ private[parser] trait CommonRules { this: Parser with StringBuilding ⇒
     `challenge-or-credentials` ~> ((scheme, params) ⇒ GenericHttpCredentials(scheme, params.toMap))
   }
 
-  // ******************************************************************************************
-  // https://tools.ietf.org/html/rfc6265#section-4.1.1
-  // ******************************************************************************************
-  def `cookie-pair` = rule {
-    `cookie-name` ~ ws('=') ~ `cookie-value` ~> (HttpCookiePair(_: String, _: String))
+  /**
+   * Either `Some(cookiePair)` if the cookie pair is parsable using the giving cookie parsing mode
+   * or None, otherwise.
+   */
+  def `optional-cookie-pair`: Rule1[Option[HttpCookiePair]] = rule {
+    (`cookie-pair` ~ &(`cookie-separator`) ~> (Some(_: HttpCookiePair))) |
+      // fallback that parses and discards everything until the next semicolon
+      (zeroOrMore(!`cookie-separator` ~ ANY) ~ &(`cookie-separator`) ~ push(None))
+  }
+
+  def `cookie-pair`: Rule1[HttpCookiePair] = rule {
+    `cookie-name` ~ ws('=') ~ `cookie-value` ~> (createCookiePair _)
   }
 
   def `cookie-name` = rule { token }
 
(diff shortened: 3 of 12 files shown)


akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
```diff
@@ -95,7 +95,15 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
...
   // https://tools.ietf.org/html/rfc6265#section-4.2
   def `cookie` = rule {
-    oneOrMore(`cookie-pair`).separatedBy((ch(';') | ch(',')) ~ OWS) ~ EOI ~> (`Cookie`(_))
+    oneOrMore(`optional-cookie-pair`).separatedBy(';' ~ OWS) ~ EOI ~> { pairs ⇒
+      val validPairs = pairs.collect { case Some(p) ⇒ p }
+      `Cookie` {
+        if (validPairs.nonEmpty) validPairs
+        // Parsing infrastructure requires to return an HttpHeader value here but it is not possible
+        // to create a Cookie header without elements, so we create a dummy instance.
+        else HttpCookiePair("__dummy__", "__No_more_cookies_left_after_discarding_invalid_ones__") :: Nil
```
jrudolph: This is very ugly. An alternative would be to throw from here and let the parsing infrastructure create a `RawHeader` instead in a case like this. WDYT, @sirthias?

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala
```diff
@@ -289,5 +292,6 @@ private[http] object BodyPartParser {
     maxHeaderCount = 64,
```
ktoso: I noticed a TODO above this line `// TODO: load from config` - are we blocked on resolving it (i.e. no actor system here it seems) by something that needs discussion or just wasn't done yet?

akka-http-core/src/main/scala/akka/http/impl/model/parser/CharacterClasses.scala
```diff
@@ -60,7 +60,11 @@ private[http] object CharacterClasses {
   val `token68-start` = ALPHA ++ DIGIT ++ "-._~+/"
 
   // https://tools.ietf.org/html/rfc6265#section-4.1.1
-  val `cookie-octet` = CharPredicate('\u0021', '\u0023' to '\u002b', '\u002d' to '\u003a', '\u003c' to '\u005b', '\u005d' to '\u007e')
+  val `cookie-octet-rfc-6265` = CharPredicate('\u0021', '\u0023' to '\u002b', '\u002d' to '\u003a', '\u003c' to '\u005b', '\u005d' to '\u007e')
+  val `cookie-separator` = CharPredicate(akka.parboiled2.EOI, ';')
+  val `cookie-octet-raw` =
+    CharPredicate('\u0020' to '\u007e') ++
+      CharPredicate((x: Char) ⇒ x > 0x7f && java.lang.Character.isDefined(x)) -- `cookie-separator`
```
ktoso: good trick, I'll remember that one

ktoso: (sanity checked values with rfc, seems good)

akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
```diff
@@ -215,17 +215,35 @@ private[parser] trait CommonRules { this: Parser with StringBuilding ⇒
...
 
+  /**
+   * Either `Some(cookiePair)` if the cookie pair is parsable using the giving cookie parsing mode
+   * or None, otherwise.
+   */
+  def `optional-cookie-pair`: Rule1[Option[HttpCookiePair]] = rule {
+    (`cookie-pair` ~ &(`cookie-separator`) ~> (Some(_: HttpCookiePair))) |
+      // fallback that parses and discards everything until the next semicolon
+      (zeroOrMore(!`cookie-separator` ~ ANY) ~ &(`cookie-separator`) ~ push(None))
+  }
```
ktoso: very nice rule :)

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
ktoso: I have dived a bit into how this piece works and I think I'd favour the throw a stacktrace less exception here actually - since there seems to be a place we can handle errors such as this already - inside the parseFull, so it'd be transparent to others?

If I understand the flow correctly (not sure due to one macro along the way) if we'd throw then we can handle a 
`case Left(e) if isSpecial(e)  ⇒ Right(RawHeader("", "")) // what header would it be...? again a special one on this layer hm :<` inside https://github.com/akka/akka/pull/18941/files?diff=split#diff-09a49db0c79cbd1ad03826f084fbd345R88 right?

I'm not clear about what RawHeader could be put there as a placeholder though, WDYM with that part of the proposal? Simply an "empty raw header"? (Related question - where is this __dummy__ filtered out, or we don't AFAICS (currently)?)

Do you think throwing would be performance wise acceptable? Should be measured I guess in any case..

// made me realise I've never used `DynamicRuleDispatch` so read up about it, interesting!

ktoso: @sirthias's feedback here would of course be worth a ton of gold... ;-)

jrudolph: No, it's not filtered out right now. 

I guess the exception makes more sense. Without adding any extra handling the surrounding logic creates a `RawHeader` with the original content.

sirthias: I agree that throwing an exception is probably the way to go.
Basically what we want to do here is "fail late", i.e. abort parsing into a ModeledHeader after having already decided that we (probably) can.
Actually it might even suffice to simply `throw HeaderParser.RuleNotFoundException`, which should fall back to creating a `RawHeader`. Could you try that, Johannes?

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala
sirthias: I don't think there is anything missing here.
It's simply another settings class that needs to be introduced, along with the proper defaults and docs in `reference.conf`. Basically just "work"... :)

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
jrudolph: Sounds good. I'll try that.

jrudolph: > Actually it might even suffice to simply throw HeaderParser.RuleNotFoundException

No, this doesn't work and fails even worse because the error handling logic cannot deal with exceptions without messages (fails in `HeaderParser.failure` -> `ErrorInfo.fromCompoundString`).

I now throw a new dedicated exception which should do the trick.

jrudolph:
Updated with throwing exceptions when all cookie values are invalid.

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
sirthias: Ok, bummer. However, the dedicated Exception is probably the cleaner solution anyway...

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala
ktoso: Gotcha, if left unaddressed please open a ticket for it - exposing config from previously hardcoded is a nice simple task  to pick up :) // cc @jrudolph

akka-http-core/src/main/scala/akka/http/impl/model/parser/HeaderParser.scala
```diff
@@ -70,6 +84,7 @@ private[http] class HeaderParser(val input: ParserInput, settings: HeaderParser.
  */
 private[http] object HeaderParser {
   object RuleNotFoundException extends SingletonException
+  object EmptyCookieException extends SingletonException("Cookie header contained no parsable cookie values.")
```
ktoso: :+1:

jrudolph:
Can this be merged?

ktoso:
Yes, sorry I didn't right away.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/BodyPartParser.scala
ktoso: Made it a ticket https://github.com/akka/akka/issues/18958

jrudolph:
No problem :)

Issue #18898, opened 2015-11-09
CustomHeaders don't work as nice with pattern matching as they could


ktoso:
Since they're missing the `unapply(RawHeader): Option[MyHeader]` they're suprisingly broken when people try using them in directives like `headerValuePF`.

We should probably provide a companion trait to be extended by users in the companion class to make this consistent.

```scala
    "CustomHeader should be able to match from RawHeader" in {
      final class ApiTokenHeader(token: String) extends CustomHeader {
        override def value: String = token
        override def name: String = ApiTokenHeader.name
      }
      object ApiTokenHeader {
        val name = "apiKey"
        val lowercaseName = name.toLowerCase(Locale.ROOT)

        def apply(token: String) = new ApiTokenHeader(token)
        def unapply(raw: RawHeader): Option[ApiTokenHeader] = {
          println("raw = " + raw.lowercaseName)
          val res =
            if (raw.lowercaseName == "apikey") Some(ApiTokenHeader(raw.value))
            else None
          println("res = " + res)
          res
        }
        //        def unapply(header: ApiTokenHeader): Option[ApiTokenHeader] =
        //          Some(header)
        //        implicit def r(raw: RawHeader) = new {
        //          def unapply(api: ApiTokenHeader): Option[RawHeader] =
        //            Some(RawHeader(api.name, api.value))
        //        }
      }

      //      val ApiTokenHeader(t1) = ApiTokenHeader("token")
      //      val RawHeader(k1, v1) = ApiTokenHeader("token")
      //      val RawHeader(k2, v2) = RawHeader("apiKey", "token")
      //      val ApiTokenHeader(v3) = RawHeader("apiKey", "token")

      val createUserApiKeys = new {
        def contains(any: Any) = true
        def mkString = "<fictional>"
      }

      val allowedApiKeys = new {
        def contains(any: Any) = true
        def mkString = "<fictional>"
      }
      case object ApiKeyAllowCreateUser
      case object ApiKeyValid

      def extractApiKey = headerValuePF {
        case ApiTokenHeader(token) if createUserApiKeys.contains(token) ⇒ ApiKeyAllowCreateUser
        case ApiTokenHeader(token) if allowedApiKeys.contains(token) ⇒ ApiKeyValid
        case ApiTokenHeader(token) ⇒ println(s"apiTokenHeader matched, but wasn't valid")
[...]

PR #18942, opened 2015-11-16
+htc #18898 modeledCustomHeader to ease matching on headers

ktoso:
Resolves https://github.com/akka/akka/issues/18898

Though I'm not completely happy yet; the problem is about not having "an additional" type, like now, but "one for all custom headers".
The problem is that custom headers are "abused" (well not really, just used in "special cases" by HTTP internally, and making that compile with the companion approach proved to be very tricky (burnt some time on it today).

Would like some initial feedback, and also opinions if we should keep 2 "custom headers" (I could make the bare one be extended by the Modeled one, and we internally keep using the bare; and tell users to use the Modeled one.

// cc @jrudolph @sirthias @rkuhn

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
@@ -9,8 +9,10 @@ import java.net.InetSocketAddress
 import java.security.MessageDigest
 import java.util
 
+import akka.event.Logging
+
 import scala.reflect.ClassTag
-import scala.util.Try
+import scala.util.{ Failure, Success, Try }
 import scala.annotation.tailrec
 import scala.collection.immutable
 
@@ -48,6 +50,10 @@ sealed trait ModeledHeader extends HttpHeader with Serializable {
 
 /**
  * Superclass for user-defined custom headers defined by implementing `name` and `value`.
+ *
+ * Prefer to extend [[ModeledCustomHeader]] and [[ModeledCustomHeaderCompanion]] instead if
+ * planning to use the defined header in match clauses (e.g. in the routing layer of Akka HTTP),
+ * as they allow the custom header to be matched from [[RawHeader]] and vice-versa.
  */
 abstract class CustomHeader extends jm.headers.CustomHeader {
   /** Override to return true if this header shouldn't be rendered */
@@ -55,6 +61,43 @@ abstract class CustomHeader extends jm.headers.CustomHeader {
 
   def lowercaseName: String = name.toRootLowerCase
   final def render[R <: Rendering](r: R): r.type = r ~~ name ~~ ':' ~~ ' ' ~~ value
+}
+
+/**
+ * To be extended by companion object of a custom header extending [[ModeledCustomHeader]].
+ * Implements necessary apply and unapply methods to make the such defined header feel "native".
+ */
+abstract class ModeledCustomHeaderCompanion[H <: ModeledCustomHeader[H]] {
+  def name: String
+  def lowercaseName: String = name.toRootLowerCase
+
+  def parse(value: String): Try[H]
+
+  def apply(value: String): H =
+    parse(value) match {
+      case Success(parsed) ⇒ parsed
+      case Failure(ex)     ⇒ throw new IllegalArgumentException(s"Unable to construct custom header by parsing: '$value'", ex)
+    }
+
+  def unapply(h: HttpHeader): Option[String] = h match {
+    case _: RawHeader    ⇒ if (h.lowercaseName == lowercaseName) Some(h.value) else None
+    case _: CustomHeader ⇒ if (h.lowercaseName == lowercaseName) Some(h.value) else None
+    case _               ⇒ None
+  }
+
+}
+
+/**
+ * Support class for building user-defined custom headers defined by implementing `name` and `value`.
+ * By implementing a [[ModeledCustomHeader]] instead of [[CustomHeader]] directly, all needed unapply
+ * methods are provided for this class, such that it can be pattern matched on from [[RawHeader]] and
+ * the other way around as well.
+ */
+abstract class ModeledCustomHeader[H <: ModeledCustomHeader[H]] extends CustomHeader { this: H ⇒
+  def companion: ModeledCustomHeaderCompanion[H]
+
+  final override def name = companion.name
+  final override def lowercaseName: String = name.toRootLowerCase
 }
 
 import akka.http.impl.util.JavaMapping.Implicits._
@@ -138,6 +181,10 @@ final case class `If-Range`(entityTagOrDateTime: Either[EntityTag, DateTime]) ex
 final case class RawHeader(name: String, value: String) extends jm.headers.RawHeader {
   val lowercaseName = name.toRootLowerCase
   def render[R <: Rendering](r: R): r.type = r ~~ name ~~ ':' ~~ ' ' ~~ value
+}
+object RawHeader {
+  def unapply[H <: HttpHeader](customHeader: H): Option[(String, String)] =
+    Some(customHeader.name -> customHeader.value)
 }
 
 // http://tools.ietf.org/html/rfc7231#section-5.3.2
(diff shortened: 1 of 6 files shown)


ktoso:
Whoops test will fail, sorry

ktoso:
Note to self, promised to deliver in M2.

ktoso:
This is now complete (I think – please review) and up for M2.

drewhk:
LGTM I guess

ktoso:
Thanks @drewhk!
Can you take a look at it @sirthias / @jrudolph ?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -57,6 +61,60 @@ abstract class CustomHeader extends jm.headers.CustomHeader {
...
 }
 
+abstract class ModeledCustomHeaderCompanion[H <: ModeledCustomHeader[H]] {
+  def name: String
+  def lowercaseName: String = name.toRootLowerCase
+
+  def parse(value: String): Option[H]
+
+  def apply(value: String): H =
+    parse(value).getOrElse(throw new RuntimeException("Unable to construct custom header by parsing value:" + value))
```
sirthias: I think, as a user, I'd expect a space char here after the `:` in the error message.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -57,6 +61,60 @@ abstract class CustomHeader extends jm.headers.CustomHeader {
...
+//  def name: String = companion.name
+//  def lowercaseName: String = name.toRootLowerCase
+//  final def render[R <: Rendering](r: R): r.type = r ~~ name ~~ ':' ~~ ' ' ~~ value
+//}
+abstract class CustomHeaderCompanion[H <: CustomHeader] {
+  def name: String
+  def lowercaseName: String = name.toRootLowerCase
+
+  def parse(value: String): Option[H]
+  def apply(value: String): H = parse(value).getOrElse(throw new Exception("Whoops"))
```
sirthias: Maybe `RuntimeException` rather than `Exception`?
And an error message that shows the offending input?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -57,6 +61,60 @@ abstract class CustomHeader extends jm.headers.CustomHeader {
...
+  def unapply(raw: RawHeader): Option[H] =
+    if (raw.lowercaseName == lowercaseName) parse(raw.value) else None
+
+  def unapply(value: String): Option[H] =
+    parse(value)
+  //  def unapply(header: H): Option[RawHeader] = Some(RawHeader(header.name, header.value))
```
sirthias: left-over commenting-out?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -57,6 +61,60 @@ abstract class CustomHeader extends jm.headers.CustomHeader {
...
+//abstract class CustomHeader[Self <: CustomHeader[Self]] extends jm.headers.CustomHeader {
+//  def companion: CustomHeaderCompanion[Self]
+//
+//  /** Override to return true if this header shouldn't be rendered */
+//  def suppressRendering: Boolean = false
+//
+//  def name: String = companion.name
+//  def lowercaseName: String = name.toRootLowerCase
+//  final def render[R <: Rendering](r: R): r.type = r ~~ name ~~ ':' ~~ ' ' ~~ value
+//}
```
ktoso: note to self remove this

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
ktoso: Ouch, that's a very crappy error message I left here. Thanks for noticing!

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
ktoso: note to self, remove this

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -135,9 +193,21 @@ final case class `If-Range`(entityTagOrDateTime: Either[EntityTag, DateTime]) ex
...
+final class RawHeader(val name: String, val value: String) extends jm.headers.RawHeader {
   val lowercaseName = name.toRootLowerCase
   def render[R <: Rendering](r: R): r.type = r ~~ name ~~ ':' ~~ ' ' ~~ value
+
+  override def equals(obj: Any): Boolean = obj match {
+    case RawHeader(n: Any @unchecked, v) ⇒ (n.toRootLowerCase == lowercaseName) && (v == value)
+    case _                               ⇒ false
+  }
+  override def hashCode(): Int = super.hashCode()
+}
```
ktoso: I hope I got that right?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -135,9 +193,21 @@ final case class `If-Range`(entityTagOrDateTime: Either[EntityTag, DateTime]) ex
 }
 
 // FIXME: resurrect SSL-Session-Info header once akka.io.SslTlsSupport supports it
-final case class RawHeader(name: String, value: String) extends jm.headers.RawHeader {
+final class RawHeader(val name: String, val value: String) extends jm.headers.RawHeader {
```
sirthias: Do we *have* to drop the `case class` here?
It seems we are loosing the `toString` implementation and `copy` method as well, no?

ktoso: True, thanks. I'll see if possible to keep it a case class and just add the apply/unapply manually.

ktoso: I was able to keep it a case class :+1:

ktoso:
All comments addressed and error reporting improved. RawHeader stays a case class.

akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/directives/BasicDirectivesExamplesSpec.scala
```diff
@@ -233,6 +234,49 @@ class BasicDirectivesExamplesSpec extends RoutingSpec {
       status shouldEqual StatusCodes.BadGateway
     }
   }
+  "1mapResponse-advanced-json" in {
```
ktoso: this was supposed to be a separate commit, removing

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -138,6 +181,11 @@ final case class `If-Range`(entityTagOrDateTime: Either[EntityTag, DateTime]) ex
 final case class RawHeader(name: String, value: String) extends jm.headers.RawHeader {
   val lowercaseName = name.toRootLowerCase
   def render[R <: Rendering](r: R): r.type = r ~~ name ~~ ':' ~~ ' ' ~~ value
+  override def hashCode(): Int = super.hashCode()
```
sirthias: What's the purpose of this now?

ktoso: can be removed, thanks

akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/directives/BasicDirectivesExamplesSpec.scala
ktoso: (in a separate PR)

ktoso:
Ok, _now_ in a state I'm happy with.

sirthias:
Cool, LGTM now!

Issue #16841, opened 2015-02-10
Add `uploadedFile` directive/unmarshaller to handle file upload


jrudolph:
@stanpalatnik in spray/spray#252:

>  Create an `uploadedFile` directive that unmarshals the request entity and parses all information about any files posted. 

> Returns: 

> An array of uploaded file content represented as a `FormFile` similarly to how Play does it: http://www.playframework.com/documentation/2.0.2/JavaFileUpload

> As discussed here: https://groups.google.com/d/msg/spray-user/qjzHJMFmXoo/aMvLfAJvUpMJ

/cc @sirthias

nykolaslima:
+1

yanivby:
+1

Mullefa:
+1

ktoso:
// cc @johanandren could be this ticket.

johanandren:
After a discussion with @ktoso the scope is a bit narrower than the initially proposed, signature will be like this ```def uploadedFile(destination: File): Directive1[Try[FormFile]]``` - accepting just one file rather than a list of files

yanivby:
Thank you for your work on this, @johanandren.
Will this directive support uploading of large files, meaning, will the directive provide a stream of the uploaded content rather than holding all of it in memory?

ktoso:
Of course :-) It boils down to: `entityBytes to Sink.file() run()`, so streaming right from the network into the file.

2beaucoup:
https://github.com/akka/akka/issues/17665#issuecomment-139176589 also contains some API suggestions

ktoso:
Initial directives merged, thanks @johanandren!
Please give the new directives a ride in M2 soon and provide feedback  :)

nykolaslima:
Awesome! Thanks @ktoso and @johanandren

yanivby:
Great!
So I assume it can be streamed from the network into any target stream, not
just files on disk? For example, in my use case I would like to upload
files directly into MongoDB GridFS (using ReactiveMongo which provides
a play.api.libs.iteratee.Iteratee object to consume the stream).

On Mon, Nov 16, 2015 at 6:15 PM, Konrad Malawski <notifications@github.com>
wrote:

> Of course :-) It boils down to: entityBytes to Sink.file() run(), so
> streaming right from the network into the file.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/akka/akka/issues/16841#issuecomment-157084604>.
>



-- 
Thanks,
Yaniv

[...]

Issue #17665, opened 2015-06-05
Document and provide examples how to do multipart/formdata uploads on the client side


jrudolph:
Low-level and with marshalling.

/cc @sirthias

jrudolph:
Here's a basic example I hacked together: https://gist.github.com/jrudolph/08d0d28e1eddcd64dbd0

nykolaslima:
Hi @jrudolph.
I'm pretty new to akka but in my opinion this example that you hacked is too complex.
File upload should be a simple feature, it's made with just a few lines in the majority of the other frameworks. I believe that we should have a simple example, that shows how to receive the entity body and upload it to somewhere. We could also have more complex examples, but this basic one is much needed IMHO.

jrudolph:
@nykolaslima thanks for the comment. We are always looking for suggestions how to simplify things. Can you give an example about how you would expect it to look like (or how other client libraries do this)? The part of creating a file upload request isn't not as short as it could be but the 10+ lines don't seem too bad:

```scala
  def createFileUploadEntityFromFile(file: File): Future[RequestEntity] = {
    require(file.exists())
    val formData =
      Multipart.FormData(
        Source.single(
          Multipart.FormData.BodyPart(
            "test",
            HttpEntity(MediaTypes.`application/octet-stream`, file.length(), SynchronousFileSource(file, chunkSize = 100000)), // the chunk size here is currently critical for performance
            Map("filename" -> file.getName))))
    Marshal(formData).to[RequestEntity]
  }
 
  def createRequest(target: Uri, file: File): Future[HttpRequest] =
    for {
      e ← createFileUploadEntityFromFile(file)
    } yield HttpRequest(HttpMethods.POST, uri = target, entity = e)
```

I guess it would be nice if there were
 * a `FormData.BodyPart` constructor that takes a file as an argument
 * a `FormData` constructor that takes one or more files as an argument
 * a `FormData.toRequestEntity` method that would allow marshalling directly from the model

This example could then be written as

```scala
  def createFileUploadEntityFromFile(file: File): Future[RequestEntity] = {
    require(file.exists())
    Multipart.FormData.fromFile(file, chunkSize = ...).toRequestEntity
  }
 
  def createRequest(target: Uri, file: File): Future[HttpRequest] =
    for {
      e ← createFileUploadEntityFromFile(file)
    } yield HttpRequest(HttpMethods.POST, uri = target, entity = e)
```

nykolaslima:
[...]

PR #18946, opened 2015-11-17
Wip #16841 uploaded file directive

johanandren:
A new very basic directive for accepting files through multipart uploads. Streams the file onto a user specified location on disk.

Refs #16841

### DIFF
--- a/akka-docs-dev/rst/scala/http/routing-dsl/directives/file-upload-directives/fileUpload.rst
+++ b/akka-docs-dev/rst/scala/http/routing-dsl/directives/file-upload-directives/fileUpload.rst
@@ -1,0 +1,25 @@
+.. _-fileUpload-:
+
+fileUpload
+==========
+
+Signature
+---------
+
+.. includecode2:: /../../akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
+   :snippet: fileUpload
+
+Description
+-----------
+Simple access to the stream of bytes for a file uploaded as a multipart form together with metadata
+about the upload as extracted value.
+
+If there is no field with the given name the request will be rejected, if there are multiple file parts
+with the same name, the first one will be used and the subsequent ones ignored.
+
+
+Example
+-------
+
+.. includecode2:: ../../../../code/docs/http/scaladsl/server/directives/FileUploadDirectivesExamplesSpec.scala
+   :snippet: fileUpload
--- a/akka-docs-dev/rst/scala/http/routing-dsl/directives/file-upload-directives/uploadedFile.rst
+++ b/akka-docs-dev/rst/scala/http/routing-dsl/directives/file-upload-directives/uploadedFile.rst
@@ -1,0 +1,31 @@
+.. _-uploadedFile-:
+
+uploadedFile
+============
+
+Signature
+---------
+
+.. includecode2:: /../../akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
+   :snippet: uploadedFile
+
+Description
+-----------
+Streams the contents of a file uploaded as a multipart form into a temporary file on disk and provides the file and
+metadata about the upload as extracted value.
+
+If there is an error writing to disk the request will be failed with the thrown exception, if there is no field
+with the given name the request will be rejected, if there are multiple file parts with the same name, the first
+one will be used and the subsequent ones ignored.
+
+.. note::
+   This directive will stream contents of the request into a file, however one can not start processing these
+   until the file has been written completely. For streaming APIs it is preferred to use the :ref:`-fileUpload-`
+   directive, as it allows for streaming handling of the incoming data bytes.
+
+
+Example
+-------
+
+.. includecode2:: ../../../../code/docs/http/scaladsl/server/directives/FileUploadDirectivesExamplesSpec.scala
+   :snippet: uploadedFile
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
@@ -1,0 +1,91 @@
+/*
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.http.scaladsl.server.directives
+
+import java.io.File
+
+import akka.http.scaladsl.server.{ MissingFormFieldRejection, Directive1 }
+import akka.http.scaladsl.model.{ ContentType, Multipart }
+import akka.util.ByteString
+import scala.concurrent.Future
+import scala.util.{ Failure, Success, Try }
+import akka.stream.scaladsl._
+
+trait FileUploadDirectives {
+
+  import BasicDirectives._
+  import RouteDirectives._
+  import FutureDirectives._
+  import MarshallingDirectives._
+
+  /**
+   * Streams the bytes of the file submitted using multipart with the given file name into a temporary file on disk.
+   * If there is an error writing to disk the request will be failed with the thrown exception, if there is no such
+   * field the request will be rejected, if there are multiple file parts with the same name, the first one will be
+   * used and the subsequent ones ignored.
+   */
+  def uploadedFile(fieldName: String): Directive1[(FileInfo, File)] =
+    extractRequestContext.flatMap { ctx ⇒
+      import ctx.executionContext
+      import ctx.materializer
+
+      fileUpload(fieldName).flatMap {
+        case (fileInfo, bytes) ⇒
+
+          val destination = File.createTempFile("akka-http-upload", ".tmp")
+          val uploadedF: Future[(FileInfo, File)] = bytes.runWith(Sink.file(destination))
+            .map(_ ⇒ (fileInfo, destination))
+
+          onComplete[(FileInfo, File)](uploadedF).flatMap {
+
+            case Success(uploaded) ⇒
+              provide(uploaded)
+
+            case Failure(ex) ⇒
+              destination.delete()
+              failWith(ex)
+
+          }
+      }
+    }
+
+  /**
+   * Collects each body part that is a multipart file as a tuple containing metadata and a `Source`
+   * for streaming the file contents somewhere. If there is no such field the request will be rejected,
+   * if there are multiple file parts with the same name, the first one will be used and the subsequent
+   * ones ignored.
+   */
+  def fileUpload(fieldName: String): Directive1[(FileInfo, Source[ByteString, Any])] =
+    entity(as[Multipart.FormData]).flatMap { formData ⇒
+      extractRequestContext.flatMap { ctx ⇒
+        implicit val mat = ctx.materializer
+        implicit val ec = ctx.executionContext
+
+        val onePartSource: Source[(FileInfo, Source[ByteString, Any]), Any] = formData.parts
+          .filter(part ⇒ part.filename.isDefined && part.name == fieldName)
+          .map(part ⇒ (FileInfo(part.name, part.filename.get, part.entity.contentType), part.entity.dataBytes))
+          .take(1)
+
+        val onePartF = onePartSource.runWith(Sink.headOption[(FileInfo, Source[ByteString, Any])])
+
+        onSuccess(onePartF)
+      }
+
+    }.flatMap {
+      case Some(tuple) ⇒ provide(tuple)
+      case None        ⇒ reject(MissingFormFieldRejection(fieldName))
+    }
+
+}
+
+object FileUploadDirectives extends FileUploadDirectives
+
+/**
+ * Additional metadata about the file being uploaded/that was uploaded using the [[FileUploadDirectives]]
+ *
+ * @param fieldName Name of the form field the file was uploaded in
+ * @param fileName User specified name of the uploaded file
+ * @param contentType Content type of the file
+ */
+final case class FileInfo(fieldName: String, fileName: String, contentType: ContentType)
--- a/akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/directives/FileUploadDirectivesExamplesSpec.scala
+++ b/akka-docs-dev/rst/scala/code/docs/http/scaladsl/server/directives/FileUploadDirectivesExamplesSpec.scala
@@ -1,0 +1,81 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package docs.http.scaladsl.server.directives
+
+import java.io.File
+
+import akka.http.scaladsl.model.{ MediaTypes, HttpEntity, Multipart, StatusCodes }
+import akka.stream.io.Framing
+import akka.util.ByteString
+import docs.http.scaladsl.server.RoutingSpec
+
+import scala.concurrent.Future
+import scala.util.{ Success, Failure }
+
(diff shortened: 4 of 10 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -0,0 +1,56 @@
...
+  import BasicDirectives._
+  import FutureDirectives._
+  import MarshallingDirectives._
+
+  /**
+   * Collects the bytes of the first file part from the request into the given destination file
+   *
+   * @param destination Will be overwritten if not empty
+   */
+  def uploadedFile(destination: File): Directive1[Try[FormFile]] =
```
2beaucoup: How about a more flexible
```scala
def upload(maxFiles: Int = 1): Directive1[Source[(FileInfo, Source[ByteString])]]
```
?

We could then build shortcuts like the above around that.

related: #17665

johanandren: So, that wouldn't actually store anything on disk, just collect the metadata and the stream for each of the parts?

2beaucoup: Yeah, maximum flexibility! ;) That could serve as a base for higher level directives that write files.

ktoso: Interesting idea @2beaucoup! I wonder if both would be useful? This being `uploadTo(file) { f => }` then?

// cc @sirthias @jrudolph

johanandren: On one hand I like that, one super simple and one very flexible but on the other hand, that is not very far off from what we already have - basically a filter and a map on the existing stream of body parts? So, bigger api surface for little gain? 

I think the super-simple one is great to keep the threshold for newcomers low and make "the simple things very easy", so we should definitely keep it or something close to it.

2beaucoup: My suggestion was to keep both, with the simple one being based on the flexible one. ;)

ktoso: > My suggestion was to keep both, with the simple one being based on the flexible one. ;)

:+1: 

And in docs we should mention how to get the `dataBytes`, and hint at "you can get the request body as a stream, isn't that amazing" ;)

johanandren:
Some more doc fixes to do, but wanted to push working code to see if there is any more api-feedback first.

johanandren:
PLS BUILD

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -0,0 +1,75 @@
...
+  import BasicDirectives._
+  import FutureDirectives._
+  import MarshallingDirectives._
+
+  /**
+   * Collects the bytes of the first file part from the request into the given destination file
+   *
+   * @param destination Will be overwritten if not empty
+   */
+  def uploadedFile(destination: File): Directive1[Try[UploadedFile]] =
```
johanandren: I'm starting to think it is a bit weird to not talk about field name here, if we look at the Play API for the same thing for example that is kind of central: https://www.playframework.com/documentation/2.0.2/JavaFileUpload

Maybe it should be `uploadedFile(fieldName: String, destination: File)` rather?

2beaucoup: makes sense!

2beaucoup:
Looks like some of my comments were force-pushed away. ;(

`ctx` in `uploadedFiles` is unused and `uploadedFile` should extract a plain `File` so we can get rid of `UploadedFile`.

johanandren:
Yeah, had to rebase on top of the Source.file changes, sorry.

About the using regular `File` instead of `UploadedFile`: The reason I added `UploadedFile` was to keep metadata such as original file name and mime type of the file, which you would otherwise loose when choosing your own file (and which probably aren't entirely safe to just trust but will require some further processing in actual apps)

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -0,0 +1,73 @@
...
+        onComplete(uploaded.runWith(Sink.head[UploadedFile]))
+      }
+    }
+
+  /**
+   * Collects each body part that is a multipart file as a tuple containing metadata and a `Source`
+   * for streaming the file contents somewhere.
+   * @param maxFiles Accept at most this many uploaded files, any additional files are ignored
+   */
+  def uploadedFiles(maxFiles: Int): Directive1[Source[(FileInfo, Source[ByteString, Any]), Any]] =
```
2beaucoup: What about
```scala 
def uploadedFiles(fieldNames: String*)
``` 
?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
2beaucoup: We could expect the destination to be a folder and reuse the `part.fileName` as the destination file. Another option would be:
```scala
def upload(fieldNames: String*, destination: FileInfo => File): Directive1[Map[FileInfo, Try[File]]]
```

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
ktoso: :+1:

2beaucoup:
What we should keep in mind is that when the `multiple` attribute is used in multpart forms (for selecting multiple files at once for one input field) then there are multiple parts with the same `name` but different `filename`s.

johanandren:
I think we should let the API look like it does in the last commit. I'm thinking the following use cases:

1. you are new to the api and just want to accept a simple file upload to your server - `uploadedFile()` is super simple and provides you with all the metadata you need
2. you want to do a more complex form where it is possible to upload multiple files (possibly with the `multiple` attribute) - `uploadedFiles()` then is for you, this also allows streaming the contents somewhere else etc.
3. You want to do something even more complex - you will have to use the "raw" multipart handling, but you can look at the sources of `FileUploadDirectives.scala` to get inspiration.

2beaucoup:
I'd vote for at least being explicit in what param name should be used instead of just using the first one.

Removing the extraction and just rejecting if something went wrong would simplify things further.

The current interface makes it hard to compute the `destination` from the uploaded `part.filename` which might be a requirement even for the simple case.

I'd therefore vote for:

```scala
// simple:
def uploadFile(fieldName: String, destinationFolder: File): Directive0 // using part.filename
// or
def uploadFile(fieldName: String, destination: FileInfo => File): Directive0

// advanced:
def uploadFiles(fieldNames: String*): Directive1[Source[(FileInfo, Source[ByteString, Any]), Any]]
```

Maybe @sirthias and @jrudolph have something to add here?

johanandren:
How about three different levels of it, all including the field names:
```Scala
// super-simple use case: creates a tmp-file with the contents, 
// if there is a failure the file is removed to not leak disk space
// rejects on error or if there is no such field - this is on par with what play, rails and php provides
def uploadedFile(fieldName: String): Directive1[UploadedFile]

// semi-advanced: 
// I still think you want the metadata extracted, if we do not extract it the user would be 
// encouraged to side-effect-saving the file info to get access to it
// rejects on error or if there is no such field
def uploadedFile(fieldName: String, destination: FileInfo => File): Directive1[UploadedFile]

// advanced use case but still filter on field names, should allow most use cases
// but you always know what fields you have in your forms in advance
// also, all error handling is left to the user
def uploadedFiles(fieldNames: String*): Directive1[Source[(FileInfo, Source[ByteString, Unit], Unit]
```

johanandren:
Or maybe complete with BadRequest/InternalServerError is better than rejecting.

ktoso:
@2beaucoup @johanandren 

As for the `uploadFile`: 
- Rejecting is good – that's what `MissingFormFieldRejection` is for. If users want to 500 they can from handling that rejection.
- `(fieldName: String, destinationFolder: File) // using part.filename` the destinationFolder seems rather meh; wouldn't this encourage usage in which I can easily end up with a corrupt file? Since multiple requests, could concurrently write to the same file - if we base the name on part.filename. I'd vote against using the filename from the req, as real filename on the disk. Thus, we need to provide the user UploadedFile with metadata about the file ("real name").

As for streams:
- getting a stream must be as simple as doing the `request > File`, since we do want to encourage "streams all the way through"
- most likely different name for this would be better; how about `fileUpload("formFieldName") { source =>`, since "fileUpload" sounds like "we'll be uploading a file" and "uploadedFile" is past tense so it implies "there is this file, and you can read it".
  - let's not do `fileUpload[s]` for now (stream of streams)

2beaucoup:
```scala
def fileUpload(fieldName: String, tempFolder: File): Directive1[(FileInfo, File)]
def fileUpload(fieldName: String, destination: FileInfo => File): Directive1[FileInfo]
def fileUpload(fieldName: String): Directive1[(FileInfo, Source[ByteString, Any])]
```
?

johanandren:
Yeah, after discussion on gitter it will be 
```
def uploadedFile(fieldName: String): Directive1[UploadedFile]
def fileUpload(fieldName: String): Directive1[(FileInfo, Source[ByteString, Any])]
```

for now, and we'll see if there is much request for more advanced ones and add those later

2beaucoup:
> def uploadedFile(fieldName: String): Directive1[UploadedFile]

But *where* is it stored?

johanandren:
`File.createTempFile`

2beaucoup:
```java
public static File createTempFile(String prefix, String suffix)
```

or

```java
public static File createTempFile(String prefix, String suffix, File directory)
```

? ;)

A hard coded destination might a bit too inflexible, but should work for now.

johanandren:
The one with the default system tmp file, prefixing the files with "akka-http-upload" and suffixing with ".tmp"

ktoso:
> A hard coded destination might a bit too inflexible, but should work for now.

Let's follow up on that as separate ticket and see if there's user demand for it :)
In this PR let's do the simplest possible thing, the signatures mentioned in: https://github.com/akka/akka/pull/18946#issuecomment-157685268 from there on we're free to add more afterwards (adding is simple, removing is not :wink:)

2beaucoup:
@johanandren, @ktoso agreed.

One last (promised!) thought: Why not avoid the introduction of `UploadedFile`? `uploadedFile` could extract a `(FileInfo, File)` tuple (analog to `fileUpload` extracting a `(FileInfo, Source[ByteString, Any])` tuple). The size would be accessable via the `File`.

Over and out. :)

ktoso:
> One last (promised!) thought: Why not avoid the introduction of UploadedFile? uploadedFile could extract a (FileInfo, File) tuple (analog to fileUpload extracting a (FileInfo, Source[ByteString, Any]) tuple). The size would be accessable via the File.

I agree :+1:

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -0,0 +1,89 @@
...
+trait FileUploadDirectives {
+
+  import BasicDirectives._
+  import RouteDirectives._
+  import FutureDirectives._
+  import MarshallingDirectives._
+
+  /**
+   * Streams the bytes of the file submitted using multipart with the given file name into a temporary file on disk.
+   * If there is an error writing to disk the request will be rejected, if there is no such field the request will
```
ktoso: Last nitpick: this is not implemented as rejecting, it uses `failWith` which is "will be failed with the cause exception".
Reject is "this route does not match".

akka-docs-dev/rst/scala/http/routing-dsl/directives/file-upload-directives/fileUpload.rst
```diff
@@ -0,0 +1,21 @@
...
+Signature
+---------
+
+.. includecode2:: /../../akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
+   :snippet: fileUpload
+
+Description
+-----------
+Simple access to the stream of bytes for a file uploaded as a multipart form together with metadata
+about the upload as extracted value.
```
ktoso: Paste the part about failure handling here

akka-docs-dev/rst/scala/http/routing-dsl/directives/file-upload-directives/uploadedFile.rst
```diff
@@ -0,0 +1,21 @@
...
+Signature
+---------
+
+.. includecode2:: /../../akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
+   :snippet: uploadedFile
+
+Description
+-----------
+Streams the contents of a file uploaded as a multipart form into a temporary file on disk and provides the file and
+metadata about the upload as extracted value.
```
ktoso: Paste the part about error handling here

ktoso: Add a 
```
.. _note: 
  This directive will stream contents of the request into a file, however one can not start processing these until the file has been written completely. For streaming APIs it is prefered to use the :ref:`-fileUpload-` directive, as it allows for streaming handling of the incoming data bytes.
```

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -0,0 +1,89 @@
...
+      case Some(tuple) ⇒ provide(tuple)
+      case None        ⇒ reject(MissingFormFieldRejection(fieldName))
+    }
+
+}
+
+object FileUploadDirectives extends FileUploadDirectives
+
+/**
+ * @param fieldName Name of the form field the file was uploaded in
```
ktoso: Needs a comment describing when it's passed out, not just the params I'd say

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileUploadDirectives.scala
```diff
@@ -0,0 +1,89 @@
...
+  import FutureDirectives._
+  import MarshallingDirectives._
+
+  /**
+   * Streams the bytes of the file submitted using multipart with the given file name into a temporary file on disk.
+   * If there is an error writing to disk the request will be rejected, if there is no such field the request will
+   * be rejected, if there are multiple file parts with the same name, the first one will be used and the subsequent
+   * ones ignored.
+   */
+  def uploadedFile(fieldName: String): Directive1[(FileInfo, File)] =
```
ktoso: these should also work with symbols, to be consistent with the rest of HTTP.
I'll provide a sample how to do that

ktoso: Ignore that :) The `FileAndResourceDirectives` are not defined using symbol/string anyway, only paths are so we're OK here.

ktoso:
LGTM, found last minor things and let's merge then

ktoso:
Thanks!

PR #18949, opened 2015-11-17
Documentation and changes to JournalProxy (PR #18876)

ghost:
- Add note in Shared LevelDB saying Journal Proxy should be used instead.
- Still refers to the plugin as JournalProxy (not PersistencePluginProxy).
- Documented the ability to set target via configuration (configuration key was left TBD).


akka-docs/rst/java/lambda-persistence.rst
```diff
@@ -892,6 +892,10 @@ backup node.
   A shared LevelDB instance is a single point of failure and should therefore only be used for testing
   purposes. Highly-available, replicated journal are available as `Community plugins`_.
 
+.. note::
+
+  This plugin has been supplanted by :ref:`Journal Proxy <journal-proxy>`.
```
patriknw: We should also start a migration guide migration-guide-2.4.x-2.5.x.rst, use https://github.com/akka/akka/blob/master/akka-docs/rst/project/migration-guide-2.3.x-2.4.x.rst as template. Nothing fancy, just mentioning what you wrote here in the note.

akka-docs/rst/java/lambda-persistence.rst
```diff
@@ -932,6 +936,37 @@ directory. This can be changed by configuration where the specified path can be
...
+.. _journal-proxy:
+
+Journal Proxy
+-------------
+
+A journal proxy allows sharing of journals and snapshot stores across multiple actor systems (on the same or on
+different nodes). This, for example, allows persistent actors to failover to a backup node and continue using the
+shared journal instance from the backup node. The journal proxy works by forwarding all the journal/snapshot store
+messages to a single, shared, persistence plugin instance, and therefor supports any use case supported by the proxied
+plugin (for example, :ref:`Persistence Query <persistence-query-scala>`).
```
patriknw: is this the right link? I don't think query plugins can be proxied.

akka-docs/rst/java/persistence.rst
```diff
@@ -918,6 +922,37 @@ directory. This can be changed by configuration where the specified path can be
 Note that it is not mandatory to specify a snapshot store plugin. If you don't use snapshots
 you don't have to configure it.
 
+.. _journal-proxy:
```
patriknw: I think those must be unique, so include the `java` as we have done in other places

patriknw:
OK TO TEST

patriknw:
looking great, thanks

akka-docs/rst/java/lambda-persistence.rst
talp-traiana: I've tested the LevelDB persistence query with the journal and it works (I also checked the code, and all queries forward messages to the journal).
I'm not sure that's actually what was intended for PQ, but I would argue that if it doesn't communicate with the persistence layer via the plugin then there's a pretty strong dependency on the exact data format the journal uses.

patriknw: ah, that's just because the leveldb query is using the write journal to serve the queries, I don't think that will be a normal solution, so we should not promise anything about that in the docs

talp-traiana:
TODO:
1. Handle future timeouts properly
2. Only do initialization-by-config if startTarget is false

talp-traiana:
@patriknw 
I did some more testing, and ran into an issue I thought you might have some ideas about.
It appears that many of the journal messages contain the ActorRef to respond to as part of the message. This causes the ask futures to time-out since they're not the ones that get the response.

There are 3 ways around this that I can think of, but I don't like any of them:
1. Change AsyncWriteJournal and the journal protocol messages to always reply to sender. Don't like it, because I assume there's a good reason it was implemented this way in the first place.
2. Add a mechanism that keeps tracks of requests and responses in the proxy. This seems like unnecessary complication of the proxy code, for something that seems pretty simple.
3. Extend AskSupport with an ask that takes an (ActorRef => Any), and inject ask's PromiseActorRef into the message. The downside to this is that it's messing with something very general (the ask mechanism) for a very specific use-case. It's also possible to just duplicate the relevant ask code, but then we get code duplication...

What do you think?

patriknw:
@talp-traiana 

> What should we do regarding "regular" messages that are not part of the protocol (ActorIdentity, etc.)?

just forward them, there is not anything else we can do

> sender

I think the reason is that we need to keep track of two things. 1) the PersistentActor ref to send the replies to, 2) the original sender of the application, i.e. the sender of the command in the PersistentActor

Can you take a look at https://github.com/akka/akka/pull/18243 and see if we can use that? It was recently merged to master.

akka-persistence/src/main/scala/akka/persistence/journal/PersistencePluginProxy.scala
```diff
@@ -84,6 +68,17 @@ final class JournalProxy(config: Config) extends Actor with Stash with ActorLogg
       }
       context.become(active(target, targetAtThisNode = true))
     } else {
+      val targetAddressKey = s"target-${pluginType.qualifier}-address"
+      if (config.hasPath(targetAddressKey)) {
```
patriknw: We normally define the property in the `reference.conf` anyway, with a placeholder value, such as empty string `""`.
In that way we can document the property and it is more visible that it exists.

akka-persistence/src/main/scala/akka/persistence/journal/PersistencePluginProxy.scala
```diff
@@ -136,8 +131,57 @@ final class JournalProxy(config: Config) extends Actor with Stash with ActorLogg
       if (targetAtThisNode && address != selfAddress)
         becomeIdentifying(address)
     case InitTimeout ⇒
-    case msg ⇒
-      targetJournal.forward(msg)
+
+    case msg @ WriteMessages(messages, _, actorInstanceId) ⇒
+      implicit val ec = context.system.dispatcher
+      val sender = context.sender()
```
patriknw: use another name than `sender`, e.g. `replyTo`

akka-persistence/src/main/scala/akka/persistence/journal/PersistencePluginProxy.scala
```diff
@@ -136,8 +131,57 @@ final class JournalProxy(config: Config) extends Actor with Stash with ActorLogg
...
+      val f = ask(targetJournal, msg)(1.second)
+      f onComplete {
```
patriknw: use `recover` + `pipeTo` as in the other cases

akka-persistence/src/main/scala/akka/persistence/journal/PersistencePluginProxy.scala
```diff
@@ -136,8 +131,57 @@ final class JournalProxy(config: Config) extends Actor with Stash with ActorLogg
...
+      val sender = context.sender()
+      val f = ask(targetJournal, msg)(2.seconds)
+      f recover { case t ⇒ DeleteSnapshotFailure(metadata, t) } pipeTo sender
+    case msg @ DeleteSnapshots(_, criteria) ⇒
+      implicit val ec = context.system.dispatcher
+      val sender = context.sender()
+      val f = ask(targetJournal, msg)(2.seconds)
+      f recover { case t ⇒ DeleteSnapshotsFailure(criteria, t) } pipeTo sender
+
+    case msg ⇒ // is this necessary? what other messages can we get?
```
patriknw: for custom plugin defined messages, let's just forward them (we can't know if there will be a reply)

akka-persistence/src/main/scala/akka/persistence/journal/PersistencePluginProxy.scala
```diff
@@ -136,8 +131,57 @@ final class JournalProxy(config: Config) extends Actor with Stash with ActorLogg
...
+          sender ! WriteMessagesFailed(t)
+          messages foreach {
+            case a: AtomicWrite       ⇒ a.payload.foreach { p ⇒ sender ! WriteMessageFailure(p, t, actorInstanceId) }
+            case r: NonPersistentRepr ⇒ sender ! LoopMessageSuccess(r.payload, actorInstanceId)
+          }
+      }
+    case msg @ DeleteMessagesTo(_, toSeqNr, _) ⇒
+      implicit val ec = context.system.dispatcher
+      val sender = context.sender()
+      val f = ask(targetJournal, msg)(2.seconds)
```
patriknw: define the timeout in a config property?

patriknw: and use a longer default, e.g. 10 seconds (some journals are slow when starting up in testing mode, e.g. embedded cassandra)

talp-traiana:
Hmm, I rebased on master to get #18243, and made a mess of this PR. Should I close this and open a new one?

patriknw:
do whatever is easiest for you

talp-traiana:
Replaced by #19008

akka-persistence/src/main/scala/akka/persistence/journal/PersistencePluginProxy.scala
talp-traiana: From what I understand, pipeTo only returns 1 message whereas when this future fails we may need to send the actor several messages. I'm open to suggestions on how to make it cleaner, though.

patriknw: ah, I missed that, you are good with `onComplete` then

PR #18951, opened 2015-11-17
+str - Adds Sink.last and Sink.lastOption to mirror Sink.head and Sin…

viktorklang:
…k.headOption

Thoughts?


akka-stream-tests/src/test/scala/akka/stream/scaladsl/LastSinkSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+import akka.stream.testkit._
+import akka.stream.testkit.Utils._
+
+class LastSinkSpec extends AkkaSpec with ScriptedTest {
+
+  val settings = ActorMaterializerSettings(system)
+
+  implicit val materializer = ActorMaterializer(settings)
+
+  "A Flow with Sink.head" must {
```
drewhk: Sink.last

drewhk:
I think it makes sense. It can be generalized to takeLast if we want to.

drewhk:
LGTM

viktorklang:
Fixed

On Wed, Nov 18, 2015 at 10:43 AM, drewhk <notifications@github.com> wrote:

> LGTM
>
> —
> <https://github.com/akka/akka/pull/18951#issuecomment-157658419>.
>



-- 
Cheers,
√

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -142,6 +142,62 @@ private[akka] final class HeadOptionSink[In](val attributes: Attributes, shape:
...
+      promise.tryFailure(t)
+      subscription = null
+    }
+
+    override def onComplete(): Unit = {
+      promise.trySuccess(Option(prev))
+      prev = null.asInstanceOf[In]
+      subscription = null
+    }
+  }
```
ktoso: raw Subscriber implementation would be good to TCK test (we do still bump into silly mistakes with them ince in a while) though not sure how many tests such one would pass (semantics wise), have you had a look?

viktorklang: That's a good comment. I think in general we want to cut down on the RS
implementations, instead opting for reuse. I can take a stab at having it
TCKed.

-- 
Cheers,
√

ktoso:
I like the operation – LGTM; in general would try to avoid raw subscriber implementations though, let's see if it can be TCKified

takeLast would also be useful, +1

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
rkuhn: instead of adding this (and applying the TCK) I’d strongly prefer a GraphStage implementation

viktorklang: Ill take a stab at that and also fix HeadOptionSubscriber

viktorklang: Created my first 2 GraphStages *proud*, I also created a sinkhole. I win.

rkuhn: :-)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/LastSinkSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+      val f = Source(p).runWith(Sink.last)
+      val proc = p.expectSubscription()
+      proc.expectRequest()
+      proc.sendComplete()
+      Await.ready(f, 100.millis)
+      f.value.get match {
+        case Failure(e: NoSuchElementException) ⇒ e.getMessage should be("last of empty stream")
+        case x                                  ⇒ fail("expected NoSuchElementException, got " + x)
+      }
+    }
```
rkuhn: Hmm, this could also be written as
~~~sclaa
intercept[NoSuchElementException](Source.empty.runWith(Sink.last).futureValue).getMessage should be("last of empty stream")
~~~
What is the advantage of going so low-level?

viktorklang: TBH it's copy-pasta from the Head* as they should behave the same.

I can boyscout both.

rkuhn:
LGTM, nice work! Please merge once ready.

viktorklang:
@rkuhn Thanks, it's ready, I'll merge when the tests pass.

Issue #18722, opened 2015-10-14
Sharding coordinator state is not updated properly after leader downing


mkotsur:
Related to [this mailing list question](https://groups.google.com/forum/#!topic/akka-user/rLILA4TOo6E).

We've observed this in _distributed data_ mode only, it works as expected in _persistence_ mode. However, it's important for us to get it working with _ddata_.

#### Environment:
* Akka 2.4.0
* `state-store-mode = "ddata";`
* `auto-down-unreachable-after = 10s; rebalance-threshold = 3; remember-entities = off;`
* akka-persistence dependency is explicitly disabled.

The playground code is [available on github](https://github.com/mkotsur/akkaClusterPoc/).

#### STR:

The logs are [available here](https://www.dropbox.com/s/ti5pxjboq1h3ms4/logs.zip?dl=1).

* First, I start a single node (port 2551), and create 6 actors there, they all fall into different shards. This works well. See `ddata.node1.txt@[09:58:27...09:59:13]`
* Then the second node joins (port 2552) and some shards move to it. That works also well. See `ddata.node1.txt@[09:59:27...09:59:43]` and `ddata.node2.txt@[09:59:27...09:59:43]`. Shards `a` and `b` are on node 2 now, the rest is still on node 1.
* Now shutting down the first node.

```
2015-10-14 10:00:14,111 DEBUG a.c.sharding.DDataShardCoordinator ClusterSystem-akka.actor.default-dispatcher-15 - ShardRegion terminated: [Actor[akka.tcp://ClusterSystem@127.0.0.1:2551/system/sharding/ping#606042766]]
[...]

PR #18960, opened 2015-11-18
=cls #18722 fix DDataShardCoordinator init

patriknw:
- the become logic was wrong when watchStateActors triggers an immediate
  state update


patriknw:
Refs #18722

Issue #18921, opened 2015-11-12
enhancement: "Draining a stream to a strict collection" / Sink.toSeq / Stream.limit


RichardBradley:
See lots of discussion on the aborted PR #18267

The ["Draining a stream to a strict collection" recipe](http://doc.akka.io/docs/akka-stream-and-http-experimental/2.0-M1/scala/stream-cookbook.html#Draining_a_stream_to_a_strict_collection) could do with replacing with a builtin `Sink.toSeq` sink helper, for a number of reasons:

 * magic numbers: avoiding users needing to guess in advance how long their stream might be
 * correctness: enforcing that the stream ends (the recipe linked above will ignore without error any elements after "MaxAllowedSeqSize")
 * ease of use for a common operation
 * performance

There is some controversy over whether "draining a stream to a strict collection" is a legitimate thing to want to do (see e.g. @drewhk's and @ktoso's comments on #18267), but after much back-and-forth which I won't try to summarise here, @rkuhn said:

> Sorry, I’m late to the party: it is obvious that collecting a finite stream into a sequence is a popular operation, I think we should support it. I’d recommend keeping the concerns orthogonal:
>
>    val single = source.limit(1).runWith(Sink.head)
>    val seq = source.limit(65536).within(5.seconds).runWith(Sink.toSeq)
>
> This implies that Sink.single is superfluous as shown above (limit will have to assert completion before emitting the final element in order to live up to its expectations). Does that work for everyone?
> 
> Side remark: instead of having an unusual alias in the Java DSL we should instead recognize this difference in the consistency spec.

rkuhn:
Yes, thanks.

lolski:
Can I give a shot?

drewhk:
Yes, go ahead!

lolski:
Before I commence, what is the status of @RichardBradley's PR here: https://github.com/akka/akka/pull/18267. Since it would overlap with my work, I would like to check if it is still actively being developed

lolski:
@rkuhn @drewhk @ktoso 
the problem I see with `source.limit(65536).within(5.seconds).runWith(Sink.toSeq)` is that there's no stopping unsuspecting users into calling `source.runWith(sink.toSeq)` since it's not enforced by the type.

If we enforce the parameter i.e, `Sink.toStrict(maxCount: Int, atMost: FiniteDuration)` I think it will be safer overall. What do you guys think?

RichardBradley:
[...]

PR #18965, opened 2015-11-19
toSeq and limit WIP

lolski:
See https://github.com/akka/akka/issues/18921

Two talking points about `toSeq[T]` (I have yet to implement `limit`):
1. `grouped(MAX_INT)` and `Sink.headOption` is used to collect and map the result to `Seq[T]`. An alternative implementation I can think of is to implement a separate stage e.g. `ToSeqStage` if the former approach is "less preferred".
2. I wonder if it is wise to use `ExecutionContexts.sameThreadExecutionContext` as I do here.

This seems like quite a rough, early-stage PR, so feel free to beat the hell out of it.

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitSpec.scala
@@ -43,5 +43,18 @@ class FlowLimitSpec extends AkkaSpec {
 
       result should be(true)
     }
+
+    "throw a StreamLimitReachedException when n < 0" in {
+      val input = (1 to 6)
+      val n = -1
+
+      val future = Source(input).limit(n).grouped(Integer.MAX_VALUE).runWith(Sink.head)
+      val result = Try(Await.result(future, 300.millis)) match {
+        case Failure(t: StreamLimitReachedException) ⇒ true
+        case _                                       ⇒ false
+      }
+
+      result should be(true)
+    }
   }
 }
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitSpec.scala
@@ -1,11 +1,9 @@
 package akka.stream.scaladsl
 
-import akka.stream.testkit.Utils._
 import akka.stream.{ StreamLimitReachedException, ActorMaterializer, ActorMaterializerSettings }
 import akka.stream.testkit.AkkaSpec
 import scala.concurrent.Await
 import scala.concurrent.duration._
-import scala.util.{ Failure, Try }
 
 class FlowLimitSpec extends AkkaSpec {
 
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitSpec.scala
@@ -1,14 +1,48 @@
 package akka.stream.scaladsl
 
+import akka.stream.testkit.Utils._
+import akka.stream.{ StreamLimitReachedException, ActorMaterializer, ActorMaterializerSettings }
 import akka.stream.testkit.AkkaSpec
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import scala.util.{ Failure, Try }
 
-/**
-  * Created by lolski on 11/25/15.
-  */
 class FlowLimitSpec extends AkkaSpec {
-  // Source(size = n).limit(n) == ???
-  "???" must {
+  import system.dispatcher
 
+  val settings = ActorMaterializerSettings(system)
+    .withInputBuffer(initialSize = 2, maxSize = 16)
+
+  implicit val mat = ActorMaterializer(settings)
+
+  // Source(size = n).limit(n) == ???
+  "Limit" must {
+    "produce output that is identical to the input when n = input.length" in assertAllStagesStopped {
+      val input = (1 to 6)
+      val n = input.length
+      val future = Source(input).limit(n).grouped(Integer.MAX_VALUE).runWith(Sink.head)
+      val result = Await.result(future, 300.millis)
+      result == input.toSeq
+    }
+  }
+
+  "produce output that is identical to the input when n > input.length" in assertAllStagesStopped {
+    val input = (1 to 6)
+    val n = input.length + 2 // n > input.length
+    val future = Source(input).limit(n).grouped(Integer.MAX_VALUE).runWith(Sink.head)
+    val result = Await.result(future, 300.millis)
+    result == input.toSeq
+  }
+
+  "produce n messages before throwing a StreamLimitReachedException when n < input.size" in {
+    val input = (1 to 6)
+    val n = input.length - 2 // n < input.length
+    val future = Source(input).limit(n).grouped(Integer.MAX_VALUE).runWith(Sink.head)
+    val result = Try(Await.result(future, 300.millis))
+    result match {
+      case Failure(t: StreamLimitReachedException) if t.n == n ⇒ true
+      case _ ⇒ false
+    }
   }
 
   // Source(size = m).limit(n) where m < n == ???
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitSpec.scala
@@ -8,21 +8,19 @@ import scala.concurrent.duration._
 import scala.util.{ Failure, Try }
 
 class FlowLimitSpec extends AkkaSpec {
-  import system.dispatcher
 
   val settings = ActorMaterializerSettings(system)
     .withInputBuffer(initialSize = 2, maxSize = 16)
 
   implicit val mat = ActorMaterializer(settings)
 
-  // Source(size = n).limit(n) == ???
   "Limit" must {
     "produce output that is identical to the input when n = input.length" in assertAllStagesStopped {
       val input = (1 to 6)
       val n = input.length
       val future = Source(input).limit(n).grouped(Integer.MAX_VALUE).runWith(Sink.head)
       val result = Await.result(future, 300.millis)
-      result == input.toSeq
+      result should be(input.toSeq)
     }
   }
 
@@ -31,22 +29,19 @@ class FlowLimitSpec extends AkkaSpec {
     val n = input.length + 2 // n > input.length
     val future = Source(input).limit(n).grouped(Integer.MAX_VALUE).runWith(Sink.head)
     val result = Await.result(future, 300.millis)
-    result == input.toSeq
+    result should be(input.toSeq)
   }
 
   "produce n messages before throwing a StreamLimitReachedException when n < input.size" in {
     val input = (1 to 6)
     val n = input.length - 2 // n < input.length
+
     val future = Source(input).limit(n).grouped(Integer.MAX_VALUE).runWith(Sink.head)
-    val result = Try(Await.result(future, 300.millis))
-    result match {
+    val result = Try(Await.result(future, 300.millis)) match {
       case Failure(t: StreamLimitReachedException) if t.n == n ⇒ true
       case _ ⇒ false
     }
+
+    result should be(true)
   }
-
-  // Source(size = m).limit(n) where m < n == ???
-  // Source(size = m).limit(n) where m > n == ???
-
-  // Source(size = n).limit(n) where one of the element throws an exception
 }
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitSpec.scala
@@ -32,16 +32,15 @@ class FlowLimitSpec extends AkkaSpec {
     }
 
     "produce n messages before throwing a StreamLimitReachedException when n < input.size" in {
+      // TODO: check if it actually produces n messages
       val input = (1 to 6)
       val n = input.length - 2 // n < input.length
 
       val future = Source(input).limit(n).grouped(Integer.MAX_VALUE).runWith(Sink.head)
-      val result = Try(Await.result(future, 300.millis)) match {
-        case Failure(t: StreamLimitReachedException) if t.n == n ⇒ true
-        case _ ⇒ false
-      }
 
-      result should be(true)
+      a[StreamLimitReachedException] shouldBe thrownBy {
+        Await.result(future, 300.millis)
+      }
     }
 
     "throw a StreamLimitReachedException when n < 0" in {
@@ -49,12 +48,9 @@ class FlowLimitSpec extends AkkaSpec {
       val n = -1
 
       val future = Source(input).limit(n).grouped(Integer.MAX_VALUE).runWith(Sink.head)
-      val result = Try(Await.result(future, 300.millis)) match {
-        case Failure(t: StreamLimitReachedException) ⇒ true
-        case _                                       ⇒ false
+      a[StreamLimitReachedException] shouldBe thrownBy {
+        Await.result(future, 300.millis)
       }
-
-      result should be(true)
     }
   }
 }
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitSpec.scala
@@ -13,6 +13,14 @@ class FlowLimitSpec extends AkkaSpec {
   implicit val mat = ActorMaterializer(settings)
 
   "Limit" must {
+    "produce empty sequence when source is empty and n = 0" in {
+      val input = Range(0, 0, 1)
+      val n = input.length
+      val future = Source(input).limit(n).grouped(Integer.MAX_VALUE).runWith(Sink.headOption)
+      val result = Await.result(future, 300.millis)
+      result should be(None)
+    }
+
     "produce output that is identical to the input when n = input.length" in {
       val input = (1 to 6)
       val n = input.length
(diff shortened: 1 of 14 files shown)


akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -98,6 +98,16 @@ object Sink {
   def headOption[T]: Sink[T, Future[Option[T]]] = new Sink(new HeadOptionSink[T](DefaultAttributes.headSink, shape("HeadOptionSink")))
 
   /**
+   * TODO: document usage
+   *
+   */
+  def toSeq[T]: Sink[T, Future[Seq[T]]] = {
```
ktoso: remember to mirror it in https://github.com/akka/akka/blob/release-2.3-dev/akka-stream/src/main/scala/akka/stream/javadsl/Sink.scala afterwards (and tests :))

RichardBradley:
Why not start on top of #18267 ?

lolski:
@RichardBradley Yes, I think it is better. How to do that though? Should I clone your akka fork and work there?

RichardBradley:
> How to do that though?

Add my repos as a upstream "remote" in git and merge in my `Sink.toSeq-Sink.single` branch (or rebase on top of it, or restart from there etc.).
i.e.
```
git clone git@github.com:lolski/akka.git && cd akka
git checkout lolski-issue-18921
git remote add RichardBradley git@github.com:RichardBradley/akka.git
git fetch RichardBradley
git merge RichardBradley/Sink.toSeq-Sink.single
# or: git rebase RichardBradley/Sink.toSeq-Sink.single
# or: git reset --hard RichardBradley/Sink.toSeq-Sink.single
```

lolski:
Thanks for the tip @RichardBradley

lolski:
Added a WIP implementation of a `limit` operator that should at least cover the happy path. Basically, limit will simply pass the element downstream but will error out if upstream has more elements than specified.

Meanwhile, I'm currently merging the other PR as well as thinking of corner cases that might happen.

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
```diff
@@ -164,6 +165,11 @@ private[stream] object Stages {
     override def create(attr: Attributes): Stage[T, immutable.Seq[T]] = fusing.Grouped(n)
   }
 
+  final case class Limit[T](n: Int, attributes: Attributes = limit) extends SymbolicStage[T, T] {
+    require(n > 0, "n must be greater than 0")
```
drewhk: I think we can allow zero here. That will still allow completion without any elements.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -317,6 +317,26 @@ private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immut
...
  */
+private[akka] final case class Limit[T](n: Int) extends PushPullStage[T, T] {
+  private var left = n
+
+  override def onPush(elem: T, ctx: Context[T]): SyncDirective = {
+    left -= 1
+    if (left >= 0) ctx.push(elem)
+    else {
+      if (ctx.isFinishing) ctx.push(elem)
+      else ctx.fail(new Exception("more elements not allowed"))
```
drewhk: this needs to be probably an OutOfBoundException. Also, it should contain a more informative text like "More than $n elements arrived."

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -572,6 +572,11 @@ trait FlowOps[+Out, +Mat] {
   def grouped(n: Int): Repr[immutable.Seq[Out], Mat] = andThen(Grouped(n))
 
   /**
+   * TODO: description
+   */
+  def limit[T](n: Int): Repr[Out, Mat] = andThen(Limit(n))
```
drewhk: need Java DSL as well.

akka-stream/src/main/scala/akka/stream/impl/Stages.scala
ktoso: good catch :+1:

drewhk:
Looks good, minor nitpicks only.

lolski:
Good to hear that there's nothing inherently wrong with the approach I chose. I'll be working on the rest then.

lolski:
The changes in the last few commits include:

1. A few unit tests for `limit` and `toSeq` (in `FlowLimitSpec` and `ToSeqSinkSpec`, respectively)
2. A Java DSL (UNTESTED)
3. `fusing.Limit` is now based on `PushStage` instead of `PushPullStage`
4. `limit` can take an integer from 0 to MAX_INT, otherwise will throw an `IllegalArgumentException`
5. A custom exception that will be thrown when the Stream max limit is reached, `StreamLimitReachedException`.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
viktorklang: I don't like the `toSeq` name: `Source(1 to 10) via flow to Sink.toSeq` <-- toto?
There are no other Sink starts with `to` so there's no prior art to refer to.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: Why does it take an `Int`? (`take` takes a Long)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -317,6 +317,22 @@ private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immut
...
  * INTERNAL API
  */
+private[akka] final case class Limit[T](n: Int) extends PushStage[T, T] {
+  private var left = n
+
+  override def onPush(elem: T, ctx: Context[T]): SyncDirective = {
+    left -= 1
+    if (left >= 0) ctx.push(elem)
+    else {
+      if (ctx.isFinishing) ctx.push(elem)
```
viktorklang: pushAndFinish?

viktorklang: Btw, should it really push it if left < 0?

rkuhn: This branch is never taken anyway since nobody calls `ctx.absorbTermination()`. The proper fix is to just remove this line.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
rkuhn: how about calling it `collect`? it does return a “collect-ion” as well …

rkuhn:
Before pushing the next update please make sure that DslConsistencySpec passes—that will tell you about all the places where these methods need to be reflected on the Java side of things.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
viktorklang: I think that would clash with GlowOps.collect. Perhaps collection?

-- 
Cheers,
√

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
lolski: Yes, you are correct. I'm just about to push the change.

Also, when `n < 0`, currently I'm following the convention of adding a `require(n >= 0, ...)` on `akka.stream.impl.Limit` class so that it throws an `IllegalArgumentException`. But we can also terminate the stream with `StreamLimitReachedException` instead

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -317,6 +317,20 @@ private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immut
 /**
  * INTERNAL API
  */
+
+private[akka] final case class Limit[T](n: Int) extends PushStage[T, T] {
+  private var left = n
+
+  override def onPush(elem: T, ctx: Context[T]): SyncDirective = {
```
viktorklang: Very nice!

I'd probably encode it like this to avoid decrementing when it already known to be exceeded.

~~~scala
override def onPush(elem: T, ctx: Context[T]): SyncDirective =
  if (left > 0) {
    left -= 1
    ctx.push(elem)
   } else ctx.fail(new StreamLimitReachedException(n))
~~~

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
lolski: Yes, I think you are correct, we should make it take a `Long`

Originally, the reason was because `limit` is meant to be used with `toSeq` which is only indexable until `Integer.MAX_VALUE` anyway. But I guess people will simply need to `limit(Int.MAX)` if they really want to drain a stream using `toSeq`.

lolski:
Also, if someone can make a remark of my `toSeq` implementation:
```def toSeq[T]: Sink[T, Future[Seq[T]]] = {
    Flow[T].grouped(Integer.MAX_VALUE).toMat(Sink.headOption)(Keep.right) mapMaterializedValue { e ⇒
      e.map(_.getOrElse(Seq.empty[T]))(ExecutionContexts.sameThreadExecutionContext)
    }
  }```

The other option is of course to add a new Sink `ToSeqSink`, and have toSeq implemented on top of that, i.e. `def toSeq = new Sink(new ToSeqSink(...))`

File: akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitSpec.scala [hunk 7]
```diff
@@ -0,0 +1,60 @@
...
+
+      result should be(true)
+    }
+
+    "throw a StreamLimitReachedException when n < 0" in {
+      val input = (1 to 6)
+      val n = -1
+
+      val future = Source(input).limit(n).grouped(Integer.MAX_VALUE).runWith(Sink.head)
+      val result = Try(Await.result(future, 300.millis)) match {
```
drewhk: you can use:
```
a[StreamLimitReachedException] shouldBe thrownBy {
   Await.result(...)
}
```

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: +1

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -365,6 +365,11 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
     new Flow(delegate.grouped(n).map(_.asJava)) // TODO optimize to one step
 
   /**
+   * TODO: description
```
drewhk: when you document this please mention here `take` (which does not fail on the limit) and at the `take` docs mention this as the failing counterpart.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
drewhk: `Sink.seq`? `src.to(Sink.seq)`

rkuhn: Sold, to the man without the funny mustache!

drewhk: ;-{)

drewhk:
I think limit should also have a version that accepts a cost function. For example ByteStrings can have their length as their cost, so I can limit incoming bytes.

viktorklang:
misclick

lolski:
In that case, would that make sense to have another version of `take` which accepts a cost function as well?

rkuhn:
Hmm, I’m not sure about overloading things, `take` should just mean one thing and since the name is used for collections as well we should stay within those bounds. `limit` is different enough that it might work, but I would actually prefer `weightedLimit` to make the difference in semantics clear.

viktorklang:
+1 to what Roland said.

-- 
Cheers,
√

drewhk:
weightedLimit makes sense. With take it won't be useful without a splitting function (for example if I take 100 bytes on a stream of ByteStrings then I expect it to be 100 bytes independently of ByteString boundaries.)

lolski:
This update includes:
1. Renaming `Sink.toSeq` to `Sink.seq` (as per https://github.com/akka/akka/pull/18965#discussion_r46003588). If we're striving for consistency with the library I think we should stick with `toSeq`. That's also my personal take, since I don't find `to(Sink.toSeq)` weird :-)
2. `limit(n)` now takes a `Long` instead of an `Int
3. Adding a weighted limit function with the following remarks:
- My proposal is to call it `limitWeighted` instead of `weightedLimit` since it's more consistent (e.g. like `take` and `takeWithin`) and of course more IDE completion friendly
- Currently the cost function evaluates to `Long`, but I'm planning to add another version that evaluates to `Double`.

rkuhn:
OK TO TEST

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+    "produce output that is identical to the input when n > input.length" in {
+      val input = (1 to 6)
+      val n = input.length + 2 // n > input.length
+      val future = Source(input).limit(n).grouped(Integer.MAX_VALUE).runWith(Sink.head)
+      val result = Await.result(future, 300.millis)
+      result should be(input.toSeq)
+    }
+
+    "produce n messages before throwing a StreamLimitReachedException when n < input.size" in {
+      // TODO: check if it actually produces n messages
```
rkuhn: doing this means using `recover` after the `limit` stage so that `grouped` actually gets a chance to emit things

rkuhn:
Looking good so far, please copy `seq` to `javadsl.Sink` and the limit* methods to the Java counterparts (there now are four of them: Source, Flow, SubSource, SubFlow) and add docs (including the operator overview listing in the reference docs).

lolski:
New changes include:
1. Finalizing `seq` and `limit*` JavaDSL methods
2. Adding docs

With regards to the cost function for `limitWithin`, is it ok to utilize `scala.math.Numeric` so we can make it generic enough to accept all numeric types (i.e. `Int`, `Long`, `Float`, and `Double`)?

viktorklang:
I'd skip limitWithin for now.

-- 
Cheers,
√

lolski:
you mean limitWeighted?

viktorklang:
right, thanks for noticing!

-- 
Cheers,
√

rkuhn:
Awesome, thanks!

rkuhn:
Concerning using Numeric: let’s give that some more consideration between 2.0 and 2.4 (i.e. within the next 4–6 weeks; please create a ticket for that.

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -404,6 +404,57 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
...
+   * upstream failure `StreamLimitException` downstream.
+   *
+   * Due to input buffering some elements may have been
+   * requested from upstream publishers that will then not be processed downstream
+   * of this step.
+   *
+   * The stream will be completed without producing any elements if `n` is zero
+   * or negative.
+   *
+   * '''Emits when''' the specified number of elements to take has not yet been reached
```
RichardBradley: Shouldn't this be more like:

"emits when" the upstream emits and the specified number ...

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -404,6 +404,57 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
...
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' the defined number of elements has been taken or upstream completes
+   *
+   * '''Cancels when''' the defined number of elements has been taken or downstream cancels
```
RichardBradley: It is not clear from this comment whether this stage will complete or cancel when the defined number of elements has been taken -- from this comment, it sounds like it does both or either?

akka-stream/src/main/scala/akka/stream/javadsl/Sink.scala
```diff
@@ -136,6 +136,19 @@ object Sink {
       _.map(akka.japi.Option.fromScalaOption)(ExecutionContexts.sameThreadExecutionContext)))
 
   /**
+   * A `Sink` that keeps on collecting incoming elements until upstream terminates.
+   * As upstream may be unbounded, `Flow[T].take` or the stricter ``Flow[T].limit` (and their variants)
+   * may be used to ensure boundedness.
+   * Materializes into a Future` of `Seq[T]` containing all the collected elements.
```
RichardBradley: the backticks look unbalanced in this comment -- does it render OK in the Scaladoc?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLimitWeightedSpec.scala
```diff
@@ -0,0 +1,63 @@
+package akka.stream.scaladsl
```
ktoso: missing copyright everywhere here, I'll add in a PR

ktoso: Fixed: https://github.com/akka/akka/pull/19300/files

lolski: I will also add the copyright for `StreamLimitReachedException.scala` in the PR for https://github.com/akka/akka/issues/19293 later

Issue #18311, opened 2015-08-24
HTTP client not properly handling connection failures


suiryc:
Using the examples given in the documentation, on an URL where nothing is listening (and thus triggering a connection failure right away - akka.stream.StreamTcpException), with akka 2.3.12 and akka-http 1.0.

With the [request-level API super pool](http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0/scala/http/client-side/request-level.html#Flow-Based_Variant), trying a lot of requests in a short time (just do a loop of a thousands requests injected in the pool) freezes the pool after some time: sometimes after a few requests, sometimes after a few hundreds ones.
The same may happen with host-level pools.

With the [connection-level API](http://doc.akka.io/docs/akka-stream-and-http-experimental/1.0/scala/http/client-side/connection-level.html#Opening_HTTP_Connections), it appears resources created by the stream (mainly the actors) remain in the system and use (more and more) memory without ever being terminated (and reclaimed by GC).
Not an issue if you do only a few requests, but it you have a system running 24/24 for a long time, it may become one.

If pointing to a valid URL (properly listening - even if responding error codes -, and not forgetting to consume the response on the client side), everything goes fine: no pool frozen, and no remaining actors/memory used.

Is there something I can/must do when facing connection issues ? Or is this something to fix inside akka-http/akka-stream ?

jrudolph:
To me it looks like a bug but it would be nice if you could provide us with a standalone reproducer that shows that something is broken inside of akka-http.

/cc @sirthias

suiryc:
Some [gist code](https://gist.github.com/suiryc/72aa6212a174d9ece44d).

I tried to keep it simple.

When testing `sendPoolFlow`, as you can see, injecting a bunch of requests, which I believed should be buffered by the Source actor - up to 10000 - and processed when possible, first output:
```
[ERROR] [08/25/2015 20:16:09.300] [test-client-akka.actor.default-dispatcher-14] [ActorSystem(test-client)] Outgoing request stream error (akka.stream.StreamTcpException)
1: akka.stream.StreamTcpException: Connection failed.
...
65: akka.stream.StreamTcpException: Connection failed.
```
then nothing anymore. A second run goes up to 37, then again nothing anymore. Depending on the computer, it can freeze as soon as ~10 requests, and sometimes go to 100-200 requests.

[...]

Issue #18928, opened 2015-11-13
BidiFlow.reverse is broken


drewhk:
It simply replaces the shape, but this reorders ports compared to what order they were enumerated before. This breaks one of the internal assumptions and leads to class-cast execeptions. Basically it rotates the exterior of the stage while keeping the interior unrotated.

/cc @rkuhn

drewhk:
Solution is simply:
```
  def reversed: BidiFlow[I2, O2, I1, O1, Mat] = {
    BidiFlow.fromGraph(FlowGraph.create(this) { implicit b ⇒
      reversed ⇒
        BidiShape(reversed.in2, reversed.out2, reversed.in1, reversed.out1)
    })
  }
```

The import autaomatically handles the remapping that is needed.

drewhk:
this will be included in my PR here: https://github.com/akka/akka/pull/18918

viktorklang:
Nice catch, @drewhk

ktoso:
Fix now ripped out from @drewhk's Tls work and part of my PR https://github.com/akka/akka/pull/18973 here

PR #18973, opened 2015-11-19
Failing Connection handling fixes in Http Client

ktoso:
The 2 problems I found:
- [x] The SlotProcessors may get "stuck". It happens when an error (connection error) comes back while still in the `waitingForSubscribePending` state instead of `running` (when becoming unconnected). Then no even would be emitted other than the `Disconnected`, so no error was signalled and the future not completed, making the pipeline get "stuck". This happens only rarely, so during tests I guess it slipped through, but in use cases where we send a lot of requests to a not existing host it was always reproducible. This is fixed now.
- [x] Actors leaking. I believe I know where, but somehow can't figure out why this can be happening.

More about issue 2: The actors leaking seem to be:

```
           ⌊-> flow-14-31-InternalConnectionFlow RepointableActorRef class akka.stream.impl.fusing.ActorGraphInterpreter status=0 no children
```

Which I think to be the `connectionFlow` passed into the `SlotProcessor` from all the way up in the `PoolInterfaceActor`, this guy:

``` scala
    val connectionFlow =
      if (httpsContext.isEmpty) Http().outgoingConnection(host, port, None, settings.connectionSettings, setup.log)
      else Http().outgoingConnectionTls(host, port, None, settings.connectionSettings, httpsContext, setup.log)

    val poolFlow = PoolFlow(
      Flow[HttpRequest]
        .log("BeforeConnectionFlow").withAttributes(ActorAttributes.createLogLevels(Logging.DebugLevel, Logging.DebugLevel, Logging.DebugLevel))
        .viaMat(connectionFlow)(Keep.right)
```

Other notes, if the HttpConnection fails as usual, then the:

``` scala
      .via(poolFlow)
      .log("AfterPoolFlow")
```

fails properly... but here we have a situation where we get the failure via:

``` scala
    case OnNext(ResponseContext(rc, responseTry)) ⇒
      log.debug("pool got response = " + ResponseContext(rc, responseTry))
      rc.responsePromise.complete(responseTry)
      activateIdleTimeoutIfNecessary()
```

instead, and I'm not sure what one should do then. Having that said, the actors leak regardless of if this or the other (i.e. "proper") failure happens. So I think there's something else wrong. I'm somewhat stuck and out of ideas currently and would appreciate some help if you have an idea what might cause the problem maybe - cc @sirthias @jrudolph @drewhk @rkuhn.

Thanks in advance, I'll continue digging into it but need to make the other PR ready for tomorrows M2 as well.

Will eventually resolve https://github.com/akka/akka/issues/18311


ktoso:
Got somewhat stuck with cleaning up, will finish it today.

drewhk:
This does not include the fusing TCP right?

ktoso:
Don't review here yet, I'm cleaning up the PR and almost there with it.
It will include your Fused TCP:

```
* a0f0e60 - (HEAD -> wip-oom-connection-level-ktoso) WIP cleaning up (12 hours ago) <Konrad Malawski>
* a981074 - =htp Fixes early error also being signalled as error via event (2 days ago) <Konrad Malawski>
* 56eccbf - -str remove DelayedInitProcessor, no longer needed (2 days ago) <Konrad Malawski>
* 9f3706e - =str: Make TCP fusable (first attempts) (2 days ago) <Endre Sándor Varga>
```

ktoso:
This is now cleaned up and includes a test to check for the leaks as well.
It also moved the `SeqActorName` around, since saw the need (i.e. it being reinvented in many places ad-hoc inside Streams, though it was only available in HTTP).

Please review.

ktoso:
Found some tests that relied on the stream supervisor being "/user/$a" :wink: 
Now it's more adequately called `StreamSupervisor-0`; fixed the tests

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
```diff
@@ -87,15 +91,15 @@ private object PoolSlot {
   private class SlotProcessor(slotIx: Int, connectionFlow: Flow[HttpRequest, HttpResponse, Any],
```
drewhk: I hope this can be soon rewritten as a GraphStage

akka-http/src/main/scala/akka/http/scaladsl/server/Directives.scala
```diff
@@ -4,8 +4,15 @@
 
 package akka.http.scaladsl.server
 
+import java.io.File
+
+import akka.http.scaladsl.model.Multipart
+import akka.stream.scaladsl.{ Source, Sink }
+import akka.util.ByteString
```
drewhk: these imports are probably not needed, since the file didn't change below.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
ktoso: I think it's possible, didn't go at it right away though. I'll make that a separate ticket.

akka-stream-testkit/src/test/scala/akka/stream/testkit/StreamTestDefaultMailbox.scala
```diff
@@ -28,8 +26,7 @@ private[akka] final case class StreamTestDefaultMailbox() extends MailboxType wi
           val actorClass = r.underlying.props.actorClass
           assert(actorClass != classOf[Actor], s"Don't use anonymous actor classes, actor class for $r was [${actorClass.getName}]")
           // StreamTcpManager is allowed to use another dispatcher
-          val specialCases: Set[Class[_]] = Set(classOf[StreamTcpManager], classOf[TcpListenStreamActor])
```
drewhk: Yeah, that was so nice to delete ;)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLogSpec.scala
```diff
@@ -28,7 +28,7 @@ class FlowLogSpec extends AkkaSpec("akka.loglevel = DEBUG") with ScriptedTest {
 
   "A Log" must {
 
-    val supervisorPath = "akka://FlowLogSpec/user/$a"
+    val supervisorPath = "akka://FlowLogSpec/user/StreamSupervisor-0"
```
drewhk: good

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
```diff
@@ -215,6 +212,9 @@ private[akka] object StreamSupervisor {
   def props(settings: ActorMaterializerSettings, haveShutDown: AtomicBoolean): Props =
     Props(new StreamSupervisor(settings, haveShutDown)).withDeploy(Deploy.local)
 
+  private val actorName = SeqActorName("StreamSupervisor")
+  def nextName(): String = actorName.next()
+
   final case class Materialize(props: Props, name: String)
     extends DeadLetterSuppression with NoSerializationVerificationNeeded
```
drewhk: I think FlowOps.deprecatedAndThenMat can be now removed (but not yet deprecatedAndThen since stream-o-streams needs those)

akka-stream/src/main/scala/akka/stream/impl/io/TcpStages.scala
```diff
@@ -0,0 +1,320 @@
...
+                                            val halfClose: Boolean,
+                                            val idleTimeout: Duration)
+  extends GraphStageWithMaterializedValue[SourceShape[StreamTcp.IncomingConnection], Future[StreamTcp.ServerBinding]] {
+
+  val out: Outlet[StreamTcp.IncomingConnection] = Outlet("IncomingConnections.out")
+  val shape: SourceShape[StreamTcp.IncomingConnection] = SourceShape(out)
+
+  private val BindTimer = "BindTimer"
+
+  // TODO: Timeout on bind
```
drewhk: that should be rather easy. Set the timer, and in onTimer fail the stage. If Bound is received, cancel the timer.

ktoso: I can give it a shot right away then

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -137,6 +138,32 @@ object GraphStageLogic {
...
+   */
+  class StageActorRef(asyncCallback: AsyncCallback[(ActorRef, Any)], override val path: ActorPath) extends akka.actor.MinimalActorRef {
+    override def provider: ActorRefProvider =
+      throw new UnsupportedOperationException("GraphStage MinimalActor does not provide")
+
+    override def !(message: Any)(implicit sender: ActorRef = Actor.noSender): Unit = {
+      asyncCallback.invoke((sender, message))
+    }
+
+    // FIXME: This ActorRef is not watchable!
```
drewhk: unfortunately this needs to be fixed. Tcp actors watch these ActorRefs and clean up in the case these go away. I.e. this cannot be postponed to another ticket.

drewhk:
I like the fusable TCP stuff so much :)

akka-stream/src/main/scala/akka/stream/impl/io/TcpStages.scala
ktoso: So - akka.io.Tcp's Bind does not have a bind timeout actually. Shall we add it to both (separate ticket) or leave as is?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: Argh, how could I have forgotten about that bit. Thanks for the ping.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowLogSpec.scala
ktoso: Right; related thought: should it not live under `/system`? For Persistence and Cluster actors like this live in the system namespace, should we move it (separate ticket)?

drewhk: How would that work? The ActorMaterializer wraps an ActorRefFactory, i.e. it will use `actorOf` on that context, or in other words, it will create this supervisor actor as the child of the actor that created the materializer or the top level system.

ktoso: Hmm true... So if I'd use an Actor's context as factory the supervisor would be living in that part of the actor tree. Makes sense.

ktoso:
The StageActorRef is now watchable

rkuhn:
awesome! will review next.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterfaceActor.scala
```diff
@@ -106,7 +116,7 @@ private class PoolInterfaceActor(hcps: HostConnectionPoolSetup,
         // if we can't dispatch right now we buffer and dispatch when demand from the pool arrives
         if (inputBuffer.isFull) {
           x.responsePromise.failure(
-            new RuntimeException(s"Exceeded configured max-open-requests value of [${inputBuffer.size}]"))
+            new RuntimeException(s"Exceeded configured max-open-requests value of [${inputBuffer.size}]")) // TODO maybe named exception?
```
rkuhn: BufferOverflowException?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterfaceActor.scala
```diff
@@ -26,7 +27,7 @@ private object PoolInterfaceActor {
 
   case object Shutdown extends DeadLetterSuppression
 
-  val name = new SeqActorName("PoolInterfaceActor")
+  val names = SeqActorName("PoolInterfaceActor")
```
rkuhn: I’m slightly in favor of the singular form, but in any case it should be used uniformly.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
```diff
@@ -107,20 +111,25 @@ private object PoolSlot {
     def waitingForSubscribePending: Receive = {
       case SubscribePending ⇒
         exposedPublisher.takePendingSubscribers() foreach (s ⇒ self ! ActorPublisher.Internal.Subscribe(s))
+        log.debug("become unconnected, from subscriber pending")
         context.become(unconnected)
     }
 
     val unconnected: Receive = {
-      case OnNext(rc: RequestContext) ⇒
+      case m @ OnNext(rc: RequestContext) ⇒
```
rkuhn: debug left-over?

ktoso: true, will remove the `m @`'s

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
ktoso: Let's do that in a separate PR, I'm not sure it's a simple "delete this method"

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -699,18 +790,63 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount:
...
+  final protected def getAsyncCallback[T](handler: T ⇒ Unit): AsyncCallback[T] = {
     new AsyncCallback[T] {
       override def invoke(event: T): Unit =
         interpreter.onAsyncInput(GraphStageLogic.this, event, handler.asInstanceOf[Any ⇒ Unit])
     }
   }
 
+  // TODO rethink if it's worth it; move the Updater to an object; etc
+  // via atomics since users may do something weird - try to Future { init } or something...
+  private final val _setStageActorRef = AtomicReferenceFieldUpdater.newUpdater(classOf[GraphStageLogic], classOf[StageActorRef], "_stageActorRefDoNotCallMeDirectly")
```
ktoso: I tried to move the updater to an object... and, wat - seems to be impossible due to the Updater being confused about some private volatile field? (even if I make the `_stageActorRefDoNotCallMeDirectly ` public, so it's not that which it is complaining about). Is it impossible to use those in Scala from a companion into the class for some reason?

// to be discussed anyway, if we need this or not. Also I was trying to make it possible to not be forced into using `implicit var` like @drewhk has to now in the TcpStages.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/StageActorRefSpec.scala
```diff
@@ -0,0 +1,176 @@
...
+      watch(stageRef)
+
+      stageRef ! Add(1)
+      source.success(None)
+
+      res.futureValue should ===(1)
+      expectTerminated(stageRef)
+    }
+
+    "complete the stage when PoisonPill-ed" in {
```
rkuhn: this title is outdated

akka-stream/src/main/scala/akka/stream/impl/io/TcpStages.scala
rkuhn: `bind` should return quickly in any case, leave out and add a ticket, please

akka-stream-tests/src/test/scala/akka/stream/scaladsl/StageActorRefSpec.scala
ktoso: thanks

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterfaceActor.scala
ktoso: ok

akka-stream/src/main/scala/akka/stream/impl/io/TcpStages.scala
```diff
@@ -0,0 +1,320 @@
...
+  override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, Future[ServerBinding]) = {
+    val bindingPromise = Promise[ServerBinding]
+
+    val logic = new TimerGraphStageLogic(shape) {
+      implicit var self: StageActorRef = _
```
rkuhn: This should be a `val`.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterfaceActor.scala
ktoso: The one in JDK does not take a message, I'll have to roll our own (or use the one from OverflowStrategy)?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -699,18 +790,63 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount:
...
+  @volatile private var _stageActorRefDoNotCallMeDirectly: StageActorRef = _ // DO NOT TOUCH!
+  final implicit def stageActorRef: ActorRef =
+    if (_stageActorRefDoNotCallMeDirectly ne null) _stageActorRefDoNotCallMeDirectly // TODO rethink this
+    else throw StageActorRefNotInitializedException()
+
+  case class StageActorRefNotInitializedException() extends RuntimeException("You must first call getStageActorRef, to initialize the Actors behaviour")
```
rkuhn: I’d vote for removing all this, getting rid of the additional field and letting users store the ref they get back within their logic—that also places the cost where the demand is.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -699,18 +790,63 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount:
...
+   *
+   * The [[StageActorRef]]'s lifecycle is bound to the Stage, in other words when the Stage is finished,
+   * the Actor will be terminated as well. The entity backing the [[StageActorRef]] is not a real Actor,
+   * but the [[GraphStageLogic]] itself, therefore it does not react to [[PoisonPill]].
+   *
+   * @param receive callback that will be called upon receiving of a message by this special Actor
+   * @return minimal actor with watch method
+   */
+  // FIXME: I don't like the Pair allocation :(
+  final protected def initStageActorRef(receive: ((ActorRef, Any)) ⇒ Unit): StageActorRef = {
```
rkuhn: … and then rename this to `getStageActorRef`

akka-stream/src/main/scala/akka/stream/impl/io/TcpStages.scala
ktoso: not possible currently; I tried to fix that, but currently `initStageActorRef` can only be called after the construction has run.

I explored a number of ideas, one would be to have a repointable ActorRef here which buffers for a while... 
Other ideas involved much changes so I dropped those.

// cc @drewhk

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -137,6 +143,91 @@ object GraphStageLogic {
...
+  }
+
+  /**
+   * Minimal actor to work with other actors and watch them in a synchronous ways
+   */
+  final class StageActorRef(val log: LoggingAdapter, asyncCallback: AsyncCallback[(ActorRef, Any)], override val path: ActorPath)
+    extends akka.actor.MinimalActorRef with MinimalDeathWatch {
+
+    override def provider: ActorRefProvider =
+      throw new UnsupportedOperationException("GraphStage MinimalActor does not provide")
```
rkuhn: well, we said that it should in fact provide ;-)

akka-stream/src/main/scala/akka/stream/impl/io/TcpStages.scala
ktoso: Another idea I had was to make the `self` always available, so no-one has to do this `self = init` code.
This lead me to the idea that maybe we need to separate the StageActorRef into a separate trait, as in `TimerGraphStageLogic with StageRef`

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: Now I'm confused – the provider I'd need for registering in the provider, same as in the PromiseActorRef.
We're not doing that in this PR, but will be in https://github.com/akka/akka/issues/19014

What am I missing about the provider? (passing it in is not a big problem)

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -137,6 +143,91 @@ object GraphStageLogic {
   private object DoNothing extends (() ⇒ Unit) {
     def apply(): Unit = ()
   }
+
+  trait MinimalDeathWatch {
+    this: MinimalActorRef ⇒
+    private var watchedBy = ActorCell.emptyActorRefSet
```
rkuhn: This is not going to work: watching/unwatching happens concurrently from the outside, this needs to be threadsafe.

ktoso: Ouch, of course. Must have thought somehow that it loops through async input, but of course doesn't will fix.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: okey, happy to remove it in fact.

akka-stream/src/main/scala/akka/stream/impl/io/TcpStages.scala
rkuhn: Once #19015 is fixed there is no difference between constructor and `preStart` anymore.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: I just don’t like UOE when we could actually easily implement this method.

ktoso: ok, cool. Will impl then, it's easy to fill in.
Thought there's something else this breaks

drewhk: The exception message was a joke and it was commented in the other PR actually :) (i.e. this was there as a stub, sorry for that)

ktoso: lol, ok! :-) Added it now.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: yes, this should be an AtomicReference and either contain a list of watchers or a tombstone. `sendTerminated` should do a `getAndSet` to install the tombstone and notify watchers, while addWatcher should do `compareAndSet` to add watchers, or simply send `DeathWatchNotification` if there is a tombstone already.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: we should simply disallow calling the getActorRef method from outside the stage.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: ok renamed right away then (I liked init more somehow, but not married to the name ;))

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: note, so the reason we need a `var` internally here for it is to guarantee "we always return the same actor ref"

rkuhn: my vote is to remove this feature: if someone calls this twice they should get different refs; of course this entails that the ref itself needs a `become` method

akka-stream/src/main/scala/akka/stream/impl/io/TcpStages.scala
ktoso: (note for reviewers - 19015 deemed not doable)

ktoso:
Thanks for the reviews; Dropping the EarlyRefs proposal and back to the previous commit with some fixups.

akka-stream/src/main/scala/akka/stream/impl/io/TcpStages.scala
ktoso: https://github.com/akka/akka/issues/19024

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterfaceActor.scala
ktoso: Re-ised ours from OverflowStrategies, makes sense.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterfaceActor.scala
ktoso: ok, singular it is.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: thanks, done

akka-http/src/main/scala/akka/http/scaladsl/server/Directives.scala
drewhk: still didn't change ;)

ktoso: right thanks ;) fixing now.

akka-stream/src/main/scala/akka/stream/impl/ActorMaterializerImpl.scala
drewhk: tcp streams was the only use-case AFAIK, I checked the usages.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -137,6 +144,129 @@ object GraphStageLogic {
   private object DoNothing extends (() ⇒ Unit) {
     def apply(): Unit = ()
   }
+
+  trait MinimalDeathWatch {
```
drewhk: fold this trait into the class and remove the ``watchee`` parameters if possible.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -137,6 +144,129 @@ object GraphStageLogic {
...
+    private[this] val _watchedBy = new AtomicReference[Set[ActorRef]](ActorCell.emptyActorRefSet)
+
+    def log: LoggingAdapter
+
+    override def isTerminated = _watchedBy.get() == StageTerminatedTombstone
+
+    //noinspection EmptyCheck
+    protected def sendTerminated(): Unit = {
+      val watchedBy = _watchedBy.getAndSet(StageTerminatedTombstone)
+      val wasAlreadyTerminated = watchedBy == StageTerminatedTombstone
```
drewhk: no need for this val anymore

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: :+1: okey

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: if I'd get into `sendTerminated()` concurrently somehow, watchedBy could be StageTerminatedTombstone already, which is `null` => NPE.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -137,6 +144,129 @@ object GraphStageLogic {
...
+    /**
+     * Special `become` allowing to swap the behaviour of this StageActorRef.
+     * Unbecome is not available.
+     */
+    def become(receive: StageActorRef.Receive): Unit = {
+      behaviour = receive
+    }
+  }
+  object StageActorRef {
+    type Receive = ((ActorRef, Any)) ⇒ Unit
```
drewhk: don't take a pair here. It complicates the API (especially Java) and later we might be able to optimize the pair away (once built-in sender() is removed for example).

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: well, I guess only "our before post stop" calls stop() so should be safe perhaps

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -699,18 +829,71 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount:
...
+  final protected def getAsyncCallback[T](handler: T ⇒ Unit): AsyncCallback[T] = {
     new AsyncCallback[T] {
       override def invoke(event: T): Unit =
         interpreter.onAsyncInput(GraphStageLogic.this, event, handler.asInstanceOf[Any ⇒ Unit])
     }
   }
 
+  // TODO rethink if it's worth it; move the Updater to an object; etc
+  private final val _setStageActorRef = AtomicReferenceFieldUpdater.newUpdater(classOf[GraphStageLogic], classOf[StageActorRef], "_stageActorRefDoNotCallMeDirectly")
+  @volatile private var _stageActorRefDoNotCallMeDirectly: StageActorRef = _ // DO NOT TOUCH!
```
drewhk: why is this needed?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: It has to be the type as in `final protected def getStageActorRef(receive: ((ActorRef, Any)) ⇒ Unit): StageActorRef = {` though. Unless we want to not have the sender at all?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -699,18 +829,71 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount:
...
+
   // Internal hooks to avoid reliance on user calling super in preStart
+  /** INTERNAL API */
   protected[stream] def beforePreStart(): Unit = ()
 
   // Internal hooks to avoid reliance on user calling super in postStop
-  protected[stream] def afterPostStop(): Unit = ()
+  /** INTERNAL API */
+  protected[stream] def afterPostStop(): Unit = {
+    val ref = _setStageActorRef.getAndSet(this, null)
```
drewhk: Why a getAndSet? Is don't think there is any concurrent access here.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: alternative would be to go AtomicReference around it to be safe.
Or we don't care and keep it a normal var, and if users screw up it may throw the exception. 

Now that you mention it Roland seemed to prefer that as well... Shall I remove all volatile/atomic from here?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: here true; so shall we remove atomic/volatile from the entire field handling?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: Case for concurrent access is: users being weird and calling getStageActorRef from a future or something;
Let's say that's their fault and I remove the volatiles here then.

ktoso:
Addressed most comments in https://github.com/ktoso/akka/commit/450a6d2d29260db803f2321a4bb418b4363b48ab @drewhk

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: but you already aquired watchedBy in a local val, there is problem with reentrancy here. You are deriving a boolean from a value that already is on your stack.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: No, you only need to register a prox function instead of the callback that just unwraps the pair. It helps people to directly call `match` on the message and not the pair.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: no, we should not support that at all. We don't support it for any other method, why would we do it for this one?

ktoso: Okey, removed in https://github.com/ktoso/akka/commit/450a6d2d29260db803f2321a4bb418b4363b48ab

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: I'm not sure I see your point, it's a normal nullcheck, because the Tombstone could have been set already. Or we extend the thinking about "it is only ever called from stop(), from beforePostStop() which is safe", then sure - no need to nullcheck. Shall I remove it?

ktoso:
Now also resolves #18928

ktoso:
PLS BUILD

ktoso:
Rebased.

ktoso:
Solves https://github.com/akka/akka/issues/18982

Issue #18830, opened 2015-11-02
Http: replace bypass with graph stage


ktoso:
Note to remember to replace bypasses in Http with a GraphStage, should make it faster *and* simpler to reason about.

drewhk:
@ktoso they have been replaced right? I mean there are no FlexiXXX anymore.

ktoso:
Not quite, there still is this quite elaborate dance concerning these:

```
val bypassFanout = b.add(Broadcast[RequestOutput](2).named("bypassFanout"))
val bypassMerge = b.add(new BypassMerge(settings, log))
val bypassOneHundredContinueInput = bypassMerge.in1
val bypassApplicationInput = bypassMerge.in2
```

We chatted with @sirthias and think it could be replaced with one graph stage instead, which should make things a bit simpler there.

My notes (to self, and FYI) why the bypass is needed:

- when responding to a HEAD request response may not contain response body, and similar situations like that
- 100-continue, since we generate 2 responses, 1st to accept the continue,
- short circuiting when parsing etc fails and we gen a response without invoking user code

drewhk:
You mean replacing larger constructs with a single GraphStage? Ok that might make sense if it simplify things. Where it does not, don't bother though, the goal is to fuse them anyway.

ktoso:
Yeah, it's more for simplifying the logic if "oh, it was a HEAD request, so when writing the response I have to ...". This logic can then be contained in one stage, instead of being spread all over the place using the bypass fanout/merge

jrudolph:
One of the main advantages would be that you could prevent the race conditions between bypass and user-handler more easily. Fusing alone may make it more deterministic but it cannot probably give you any additional hard guarantees. With graph stages you can maintain one shared state before and after the user-handler.

sirthias:
Just as a heads up:
I'm currently in the process of doing this refactoring for the server side, as it makes the solution to #16898 much easier. Doesn't appear to be too hard so far.

ktoso:
Cool, thanks for the heads up!

PR #18979, opened 2015-11-20
htp: remove server bypass, add early response detection

### LINKED ISSUE #16898: Add detection and clean handling of early responses
As the discussion in #16893 shows we should add proper server-side early response handling by:

1. Detecting when a user app sends an early response.
An early response is one that the HTTP layer receives from the user before we have dispatched the end of the request entity stream to the application.
An early response can by definition never be a "successful" response because the server application cannot determine that the request was successfully received in its entirety before having seen end-of-stream on the request entity. Therefore an early 2xx response is fishy.

2. When we receive an early response we should:
- log a warning if the response is a 2xx response
- add a `Connection: close` header
- potentially switch to a more aggressive TCP closing mode (#16899)

sirthias:
Closes #16898 and solves the first (server-side) half of #18830.

/cc @jrudolph

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -31,22 +31,22 @@ import akka.util.ByteString
  * INTERNAL API
  *
  *
- * HTTP pipeline setup:
+ * HTTP pipeline setup (without the underlying SSL/TLS (un)wrapping and the websocket switch):
  *
- *                     +-------------+                 +-------------+                   +-----------+
- *     HttpRequest     |  request-   |     Request-    |             |      Request-     | request-  |     ByteString
- * | <-----------------+ Preparation <-----------------+             <-------------------+  Parsing  <---------------
- * |                   |             |     Output      |             |      Output       |           |
- * |                   +-------------+                 |             |                   +-----------+
- * |                                                   |             |
- * | Application-                                      | controller- |
- * | Flow                                              |    Stage    |
- * |                                                   |             |
- * |                                                   |             |                   +-----------+
- * |                  HttpResponse                     |             |     Response-     | renderer- |     ByteString
- * v -------------------------------------------------->             +-------------------> Pipeline  +-------------->
- *                                                     |             |     Rendering-    |           |
- *                                                     +-------------+     Context       +-----------+
+ *                 +----------+          +-------------+          +-------------+             +-----------+
+ *    HttpRequest  |          |   Http-  |  request-   | Request- |             |   Request-  | request-  | ByteString
+ *  | <------------+          <----------+ Preparation <----------+             <-------------+  Parsing  <-----------
+ *  |              |          |  Request |             | Output   |             |   Output    |           |
+ *  |              |          |          +-------------+          |             |             +-----------+
+ *  |              |          |                                   |             |
+ *  | Application- | One2One- |                                   | controller- |
+ *  | Flow         |   Bidi   |                                   |    Stage    |
+ *  |              |          |                                   |             |
+ *  |              |          |                                   |             |             +-----------+
+ *  | HttpResponse |          |           HttpResponse            |             |  Response-  | renderer- | ByteString
+ *  v ------------->          +----------------------------------->             +-------------> Pipeline  +---------->
+ *                 |          |                                   |             |  Rendering- |           |
+ *                 +----------+                                   +-------------+  Context    +-----------+
  */
 private[http] object HttpServerBluePrint {
   def apply(settings: ServerSettings, remoteAddress: Option[InetSocketAddress], log: LoggingAdapter)(implicit mat: Materializer): Http.ServerLayer = {
@@ -81,7 +81,7 @@ private[http] object HttpServerBluePrint {
       rootParser.createShallowCopy().stage).named("rootParser")
       .map(establishAbsoluteUri)
 
-    val requestPreparation =
+    val requestPreparationFlow =
       Flow[RequestOutput]
         .splitWhen(x ⇒ x.isInstanceOf[MessageStart] || x == MessageEnd)
         .via(headAndTailFlow)
@@ -105,58 +105,58 @@ private[http] object HttpServerBluePrint {
         // `buffer` will ensure demand and therefore make sure that completion is reported eagerly.
         .buffer(1, OverflowStrategy.backpressure)
 
-    val rendererPipeline =
+    val responseRenderingFlow =
       Flow[ResponseRenderingContext]
         .via(Flow[ResponseRenderingContext].transform(() ⇒ responseRendererFactory.newRenderer).named("renderer"))
         .flatMapConcat(ConstantFun.scalaIdentityFunction)
         .via(Flow[ResponseRenderingOutput].transform(() ⇒ errorLogger(log, "Outgoing response stream error")).named("errorLogger"))
 
-    BidiFlow.fromGraph(FlowGraph.create(requestParsingFlow, rendererPipeline)(Keep.none) { implicit b ⇒
-      (requestParsing, renderer) ⇒
-        import FlowGraph.Implicits._
+    BidiFlow.fromGraph(FlowGraph.create() { implicit b ⇒
+      import FlowGraph.Implicits._
 
-        // HTTP
-        val requestPrep = b.add(requestPreparation)
-        val controllerStage = b.add(new ControllerStage(settings, log))
-        val csRequestParsingIn = controllerStage.in1
-        val csRequestPrepOut = controllerStage.out1
-        val csHttpResponseIn = controllerStage.in2
-        val csResponseCtxOut = controllerStage.out2
-        requestParsing.outlet ~> csRequestParsingIn
-        csResponseCtxOut ~> renderer.inlet
-        csRequestPrepOut ~> requestPrep
+      // HTTP
+      val requestParsing = b.add(requestParsingFlow)
+      val requestPreparation = b.add(requestPreparationFlow)
+      val responseRendering = b.add(responseRenderingFlow)
+      val controllerStage = b.add(new ControllerStage(settings, log))
+      val csRequestParsingIn = controllerStage.in1
+      val csRequestPrepOut = controllerStage.out1
+      val csHttpResponseIn = controllerStage.in2
+      val csResponseCtxOut = controllerStage.out2
+      requestParsing.outlet ~> csRequestParsingIn
+      csResponseCtxOut ~> responseRendering.inlet
+      csRequestPrepOut ~> requestPreparation
 
-        // One2OneBidi
-        val one2one = b.add(new One2OneBidi[HttpRequest, HttpResponse](settings.pipeliningLimit))
-        requestPrep.outlet ~> one2one.in1
-        one2one.out2 ~> csHttpResponseIn
+      // One2OneBidi
+      val one2one = b.add(new One2OneBidi[HttpRequest, HttpResponse](settings.pipeliningLimit))
+      requestPreparation.outlet ~> one2one.in1
+      one2one.out2 ~> csHttpResponseIn
 
-        // Websocket
-        val http = FlowShape(requestParsing.inlet, renderer.outlet)
-        val switchTokenBroadcast = b.add(Broadcast[ResponseRenderingOutput](2))
-        val switchToWebsocket = b.add(Flow[ResponseRenderingOutput]
-          .collect { case _: ResponseRenderingOutput.SwitchToWebsocket ⇒ SwitchToWebsocketToken })
-        val websocket = b.add(ws.websocketFlow)
-        val protocolRouter = b.add(WebsocketSwitchRouter)
-        val protocolMerge = b.add(new WebsocketMerge(ws.installHandler, settings.websocketRandomFactory, log))
-        val wsSwitchTokenMerge = b.add(WsSwitchTokenMerge)
-        switchTokenBroadcast ~> switchToWebsocket
-        protocolRouter.out0 ~> http ~> switchTokenBroadcast ~> protocolMerge.in0
-        protocolRouter.out1 ~> websocket ~> protocolMerge.in1
-        switchToWebsocket ~> wsSwitchTokenMerge.in1
-        wsSwitchTokenMerge.out ~> protocolRouter.in
+      // Websocket
+      val http = FlowShape(requestParsing.inlet, responseRendering.outlet)
+      val switchTokenBroadcast = b.add(Broadcast[ResponseRenderingOutput](2))
+      val switchToWebsocket = b.add(Flow[ResponseRenderingOutput]
+        .collect { case _: ResponseRenderingOutput.SwitchToWebsocket ⇒ SwitchToWebsocketToken })
+      val websocket = b.add(ws.websocketFlow)
+      val protocolRouter = b.add(WebsocketSwitchRouter)
+      val protocolMerge = b.add(new WebsocketMerge(ws.installHandler, settings.websocketRandomFactory, log))
+      val wsSwitchTokenMerge = b.add(WsSwitchTokenMerge)
+      switchTokenBroadcast ~> switchToWebsocket ~> wsSwitchTokenMerge.in1
+      protocolRouter.out0 ~> http ~> switchTokenBroadcast ~> protocolMerge.in0
+      protocolRouter.out1 ~> websocket ~> protocolMerge.in1
+      wsSwitchTokenMerge.out ~> protocolRouter.in
 
-        // SSL/TLS
-        val unwrapTls = b.add(Flow[SslTlsInbound] collect { case x: SessionBytes ⇒ x.bytes })
-        val wrapTls = b.add(Flow[ByteString].map[SslTlsOutbound](SendBytes))
-        unwrapTls ~> wsSwitchTokenMerge.in0
-        protocolMerge.out ~> wrapTls
+      // SSL/TLS
+      val unwrapTls = b.add(Flow[SslTlsInbound] collect { case x: SessionBytes ⇒ x.bytes })
+      val wrapTls = b.add(Flow[ByteString].map[SslTlsOutbound](SendBytes))
+      unwrapTls ~> wsSwitchTokenMerge.in0
+      protocolMerge.out ~> wrapTls
 
-        BidiShape[HttpResponse, SslTlsOutbound, SslTlsInbound, HttpRequest](
-          one2one.in2,
-          wrapTls.outlet,
-          unwrapTls.inlet,
-          one2one.out1)
+      BidiShape[HttpResponse, SslTlsOutbound, SslTlsInbound, HttpRequest](
+        one2one.in2,
+        wrapTls.outlet,
+        unwrapTls.inlet,
+        one2one.out1)
     })
   }
 
@@ -169,6 +169,7 @@ private[http] object HttpServerBluePrint {
     val shape = new BidiShape(requestParsingIn, requestPrepOut, httpResponseIn, responseCtxOut)
 
     def createLogic(effectiveAttributes: Attributes) = new GraphStageLogic(shape) {
+      val pullHttpResponseIn = () ⇒ pull(httpResponseIn)
       var openRequests = immutable.Queue[RequestStart]()
       var oneHundredContinueResponsePending = false
       var pullSuppressed = false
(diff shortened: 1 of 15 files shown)


ktoso:
Awesome, thanks - will review soon, out of home now. Have a nice weekend in any case! :)

jrudolph:
Nice!

sirthias:
One benefit of this refactoring: The server-layer is now, finally, fully reusable!

drewhk:
Can we actually cache it then in the relevant places?

sirthias:
Usually it is not really reused anywhere. You only benefit from the reusability if you bind, unbind, and rebind, which should be a rare use case.

jrudolph:
Actually, the websocket part is probably not reusable.

sirthias:
Ah, ok, that sucks. I thought the ugly hack involving the ActorRef to the OneHundredContinue trigger was all that prevented reusability.

analytically:
Impressive work!

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -96,208 +87,222 @@ private[http] object HttpServerBluePrint {
...
+        val csRequestParsingIn = controllerStage.in1
+        val csRequestPrepOut = controllerStage.out1
+        val csHttpResponseIn = controllerStage.in2
+        val csResponseCtxOut = controllerStage.out2
+        requestParsing.outlet ~> csRequestParsingIn
+        csResponseCtxOut ~> renderer.inlet
+        csRequestPrepOut ~> requestPrep
+
+        // One2OneBidi
+        val one2one = b.add(new One2OneBidi[HttpRequest, HttpResponse](16)) // TODO: replace hard-coded value by reintroducing `pipeline-limit` setting from spray
```
ktoso: Could you try to address this TODO as part of this PR already @sirthias?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -96,208 +87,222 @@ private[http] object HttpServerBluePrint {
...
-
-    override val shape = new FanInShape3(bypassInput, oneHundredContinue, applicationInput, out)
-
-    override def createLogic(effectiveAttributes: Attributes) = new GraphStageLogic(shape) {
-      var requestStart: RequestStart = _
-
-      setHandler(bypassInput, new InHandler {
-        override def onPush(): Unit = {
-          grab(bypassInput) match {
+  class ControllerStage(settings: ServerSettings, log: LoggingAdapter)
```
ktoso: reads much nicer indeed :+1:

drewhk:
Please enable `akka.stream.materializer.debug.fuzzing-mode=on` on the relevant akka-http projects to see if these changes cause any problems. Probably another ticket.

ktoso:
Failure was unrelated (AcknowledgeSink for which we have a ticket), LGTM.
Would you like to amend the remaining TODO before we merge here @sirthias ?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
sirthias: Ok, will do.

sirthias:
@drewhk You mean change the config in all tests?

drewhk:
you should put it into /test/resources/reference.conf so that it applies to all tests and then disable it in tests where it is provably wrong to enable it.

sirthias:
@drewhk I see, thanks. Will do.

sirthias:
@drewhk With `fuzzing-mode=on` I get sporadic future timeout errors in 2 specs of the http-core tests in about 50% of the cases (i.e. about half the time all http-core tests pass on my machine).
So, there does seem to be something to investigate. I'll therefore create a separate ticket for enabling fuzzing-mode in http tests and investigate errors, so we can get this thing here merged first.

drewhk:
Ok good! It is a quite cool feature, it would be nice to have it enabled everywhere :)

sirthias:
Ok, rebased, added `pipelining-limit` server config setting and prepared for enabling fuzzing mode.
Also added a pretty ascii art showing the server-side HTTP pipeline setup.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -96,208 +105,222 @@ private[http] object HttpServerBluePrint {
...
     val rendererPipeline =
       Flow[ResponseRenderingContext]
-        .recover(errorResponseRecovery(log, settings))
         .via(Flow[ResponseRenderingContext].transform(() ⇒ responseRendererFactory.newRenderer).named("renderer"))
         .flatMapConcat(ConstantFun.scalaIdentityFunction)
         .via(Flow[ResponseRenderingOutput].transform(() ⇒ errorLogger(log, "Outgoing response stream error")).named("errorLogger"))
 
-    BidiFlow.fromGraph(FlowGraph.create(requestParsingFlow, rendererPipeline, oneHundredContinueSource)((_, _, _) ⇒ ()) { implicit b ⇒
-      (requestParsing, renderer, oneHundreds) ⇒
+    BidiFlow.fromGraph(FlowGraph.create(requestParsingFlow, rendererPipeline)(Keep.none) { implicit b ⇒
```
rkuhn: passing parameters here is only for the purpose of using their materialized values—it would seem that requestParsing and rendererPipeline might as well be imported like all the others below

drewhk: +1

drewhk:
@sirthias @jrudolph can you tell which parts are still not reusable? We might be able to fix those now or very soon, but I don't know what is the current non-reusable part and why.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -96,208 +105,222 @@ private[http] object HttpServerBluePrint {
...
 
-    // we need to make sure that only one element per incoming request is queueing up in front of
-    // the bypassMerge.bypassInput. Otherwise the rising backpressure against the bypassFanout
-    // would eventually prevent us from reading the remaining request chunks from the transportIn
-    val bypass = Flow[RequestOutput].collect {
-      case r: RequestStart      ⇒ r
-      case m: MessageStartError ⇒ m
-    }
-
     val rendererPipeline =
```
2beaucoup: minor nitpick: rename `requestParsingFlow`/`rendererPipeline` to `requestParsing`/`responseRendering`?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -96,208 +105,222 @@ private[http] object HttpServerBluePrint {
...
-        protocolRouter.out1 ~> websocket ~> protocolMerge.in1
-
-        // protocol switching
         val wsSwitchTokenMerge = b.add(WsSwitchTokenMerge)
-        // feed back switch signal to the protocol router
-        switchSource ~> wsSwitchTokenMerge.in1
+        switchTokenBroadcast ~> switchToWebsocket
+        protocolRouter.out0 ~> http ~> switchTokenBroadcast ~> protocolMerge.in0
+        protocolRouter.out1 ~> websocket ~> protocolMerge.in1
+        switchToWebsocket ~> wsSwitchTokenMerge.in1
```
rkuhn: please fold into three lines above: `switchTokenBroadcast ~> switchToWebsocket ~> wsSwitchTokenMerge.in1`

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
sirthias: I see. I was wondering about the difference myself.
Thanks for the clarification, will simplify.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
sirthias: Yup, that's better!

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
sirthias: Yes, will do.

sirthias:
@drewhk The HTTP part itself should be fully reusable.
@jrudolph said that there might be non-reusable stuff in the websocket graph, but I don't know/remember which things that comprises.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -96,208 +105,222 @@ private[http] object HttpServerBluePrint {
...
+            case NonFatal(e) ⇒
+              log.error(e, "Internal server error, sending 500 response")
+              emitErrorResponse(HttpResponse(StatusCodes.InternalServerError))
           }
       })
 
-      def finishWithError(status: StatusCode, info: ErrorInfo): Unit = {
+      setHandler(responseCtxOut, new OutHandler {
+        def onPull(): Unit = if (!hasBeenPulled(httpResponseIn)) pull(httpResponseIn)
+        override def onDownstreamFinish() = cancel(httpResponseIn)
```
rkuhn: if there are outstanding requests then this will deadlock the stage, right?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -96,208 +105,222 @@ private[http] object HttpServerBluePrint {
...
+          val close = requestStart.closeRequested ||
+            requestStart.expect100Continue && oneHundredContinueResponsePending ||
+            isClosed(requestParsingIn) && openRequests.isEmpty ||
+            isEarlyResponse
+          push(responseCtxOut, ResponseRenderingContext(response, requestStart.method, requestStart.protocol, close))
+          if (close) complete(responseCtxOut)
         }
+        override def onUpstreamFinish() =
+          if (openRequests.isEmpty && isClosed(requestParsingIn)) completeStage()
+          else complete(responseCtxOut)
```
rkuhn: there won’t be any more responses, so why not complete the stage right away?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -96,208 +105,222 @@ private[http] object HttpServerBluePrint {
...
+          else pull(requestParsingIn)
+        override def onDownstreamFinish() = cancel(requestParsingIn)
+      })
+
+      setHandler(httpResponseIn, new InHandler {
+        def onPush(): Unit = {
+          val response = grab(httpResponseIn)
+          val requestStart = openRequests.head
+          openRequests = openRequests.tail
+          val isEarlyResponse = messageEndPending && openRequests.isEmpty
```
rkuhn: this means that in the face of pipelining early responses are not detected; fixing this would be highly welcome :-) (since we’ll want to piggy-back on the code in order to consume non-materialized request entities at this point as well)

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -96,208 +105,222 @@ private[http] object HttpServerBluePrint {
...
+
+            case NonFatal(e) ⇒
+              log.error(e, "Internal server error, sending 500 response")
+              emitErrorResponse(HttpResponse(StatusCodes.InternalServerError))
           }
       })
 
-      def finishWithError(status: StatusCode, info: ErrorInfo): Unit = {
+      setHandler(responseCtxOut, new OutHandler {
+        def onPull(): Unit = if (!hasBeenPulled(httpResponseIn)) pull(httpResponseIn)
```
rkuhn: This is fishy (in general—using hasBeenPulled reeks of not knowing what state we’re in); in particular it is quite hard to predict when this handler will be in effect due to the simultaneous use of `emit`.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -96,208 +105,222 @@ private[http] object HttpServerBluePrint {
...
+          if (isEarlyResponse && response.status.isSuccess)
+            log.warning(
+              """Sending 2xx response before end of request was received...
+                |Note that the connection will be closed after this response. Also, many clients will not read early responses!
+                |Consider waiting for the request end before dispatching this response!""".stripMargin)
+          val close = requestStart.closeRequested ||
+            requestStart.expect100Continue && oneHundredContinueResponsePending ||
+            isClosed(requestParsingIn) && openRequests.isEmpty ||
+            isEarlyResponse
+          push(responseCtxOut, ResponseRenderingContext(response, requestStart.method, requestStart.protocol, close))
```
rkuhn: I cannot prove that this is safe here; my recommendation is to `emit` exclusively on this output port, and add a `() => pull(httpResponseIn)` (lifted into a `val`, of course)

jrudolph:
I think the only thing that isn't reusable is the `websocketSetup`. To make that reusable you would have to access materialized values of the Websocket SubscriberSource / PublisherSink stubs during materialization.

akka-stream/src/main/scala/akka/stream/scaladsl/One2OneBidiFlow.scala
```diff
@@ -77,10 +75,7 @@ object One2OneBidiFlow {
 
       setHandler(outOut, new OutHandler {
         override def onPull(): Unit = pull(outIn)
-        override def onDownstreamFinish(): Unit = {
-          cancel(outIn)
-          cancel(inIn) // short-cut to speed up cleanup of upstream
-        }
+        override def onDownstreamFinish(): Unit = cancel(outIn)
```
rkuhn: point 4) above is not true, it will signal UnexpectedOutputException in this case

`GraphStageLogic.complete(...)` should get the same treatment of Emitting states that `completeStage()` has, I suspect that this may be the cause of some problems you see

`pullsSuppressed` can only ever be 0 or 1, i.e. it should be a boolean

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -29,6 +29,24 @@ import akka.util.ByteString
...
+ * |                   +-------------+                 |             |                   +-----------+
+ * |                                                   |             |
+ * | Application-                                      | controller- |
+ * | Flow                                              |    Stage    |
+ * |                                                   |             |
+ * |                                                   |             |                   +-----------+
+ * |                  HttpResponse                     |             |     Response-     | renderer- |     ByteString
+ * v -------------------------------------------------->             +-------------------> Pipeline  +-------------->
+ *                                                     |             |     Rendering-    |           |
+ *                                                     +-------------+     Context       +-----------+
```
rkuhn: I know it is getting big, but not including the One2One stage, websockets and TLS is slightly confusing; maybe there should at least be some indication that something is missing.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
sirthias: I'd expect the cancellation to propagate through the application and the incoming side back to the TCP layer.
But I guess we could also `completeStage()` here since it's unlikely that there is still interesting stuff going on on the incoming side.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
sirthias: Because in certain cases (involving early responses!) the *entity* of the last response might depend on the entity of the last request still being read from the network. A simple example would be the classic "ping/pong"  where you stuff the request entity directly into the response. We actually have a test for exactly this setup.
If we complete the stage here we kill the incoming side as well and therefore the entity of the last response.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
sirthias: If we already have another request queued then the response cannot be an "early" one anymore.
The `messageEndPending` here always concerns the last request in the queue. For all earlier ones we must have already seen the end of the request entity, otherwise we wouldn't have been able to read (and queue) the next request!

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
sirthias: True. Thinking about it again there is currently potential for a "push-without-preceding-pull" error.
I'll straighten this out.
Thanks for the cue!

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
sirthias: Yes, thanks!

akka-stream/src/main/scala/akka/stream/scaladsl/One2OneBidiFlow.scala
sirthias: > point 4) above is not true, it will signal UnexpectedOutputException in this case

Yes, true. Appears to be a left-over from an earlier implementation.
I'll simply remove point 4) from the scaladoc.

> pullsSuppressed can only ever be 0 or 1, i.e. it should be a boolean

Yes. I always keep forgetting that there can only every be one pull and one push pending.

> GraphStageLogic.complete(...) should get the same treatment of Emitting states that completeStage() has, I suspect that this may be the cause of some problems you see

I don't understand what you mean by that. Can you clarify?

rkuhn: I mean that it should check the currently installed handler and see if it is Emitting, scheduling the completion for later in this case (see https://github.com/akka/akka/blob/release-2.3-dev/akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala#L393).

sirthias: Ah, yes, that makes sense.
Should I make the change here or postpone to an extra ticket?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
sirthias: Coming back to this:
Wouldn't an `emit(..., () => pull(httpResponseIn))` create a potential OOME here?
After all we'd be pulling from the upstream at full speed without any backpressure from downstream allowing to come in!

akka-stream/src/main/scala/akka/stream/scaladsl/One2OneBidiFlow.scala
rkuhn: Please make it here, it might be that this PR is affected by the lack of this feature.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
rkuhn: No, that is precisely what the callback is for: it is only executed once we have emitted to the Outlet, not before.

sirthias: Ok, I see. Thx!

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
drewhk: No worries, once you enable the fuzzer it will find many of these ;) (talking from experience)

akka-stream/src/main/scala/akka/stream/scaladsl/One2OneBidiFlow.scala
drewhk: > Yes. I always keep forgetting that there can only every be one pull and one push pending.

`PushPullStage` enforces this mostly at compile time with proper types, but `GraphStage` does not. Btw this simple invariant is what allows us to make these interpreters allocation free.

sirthias:
Ok, applied all review feedback so far.
Ready for next round.

sirthias:
@ktoso Does this make it into M2 as well?

drewhk:
I would like to. Any objections?

ktoso:
Yes let's include it as well

ktoso:
I'll give the last commits a look now before merging.

ktoso:
Last commits look good, thanks @sirthias

jrudolph:
I just figured that passing the RemoteAddress to the request parser is another reason the serverblueprint is not reusable. I created #19050 to track the effort.

PR #18981, opened 2015-11-20
=act move lookupAndSetField to Reflect helper.

andreaTP:
Is there any reason I'm not aware of to keep lookupAndSetField within the ActorCell class?
It looks like a perfect match for akka.util.Reflect!

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/ActorCell.scala
+++ b/akka-actor/src/main/scala/akka/actor/ActorCell.scala
@@ -17,6 +17,7 @@ import scala.concurrent.duration.Duration
 import java.util.concurrent.ThreadLocalRandom
 import scala.util.control.NonFatal
 import akka.dispatch.MessageDispatcher
+import akka.util.Reflect
 
 /**
  * The actor context - the view of the actor cell from the actor.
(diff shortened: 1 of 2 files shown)


ktoso:
OK TO TEST

akka-actor/src/main/scala/akka/actor/ActorCell.scala
```diff
@@ -17,6 +17,7 @@ import scala.concurrent.duration.Duration
 import java.util.concurrent.ThreadLocalRandom
 import scala.util.control.NonFatal
 import akka.dispatch.MessageDispatcher
+import akka.util.Reflect.lookupAndSetField
```
ktoso: Can we change the actual lines using this method to `Reflect.lookupAndSetField`?
It's more explicit then that where we're calling into then

ktoso:
After my comment is addressed LGTM :)

andreaTP:
fixed, the AbstractFSM modification is just a trailing space removed... should I  remove it?

ktoso:
That would be nice @andreaTP, yes thanks! Will cause less merge trouble if we end up having to do weird cherry picks eventually.

Needs another pair of eyes to look at it and we can merge if reviewer is happy // cc @drewhk @rkuhn @patriknw

andreaTP:
ok, @ktoso , fixed, probably that space needs to be fixed anyway, maybe you could do that with a simple commit?
It tends to be fixed while compiling and then you cannot do anymore ```git add .``` XD

Thanks!

patriknw:
LGTM

ktoso:
LGTM

andreaTP:
Great! Thanx!

Issue #18556, opened 2015-09-25
add efficient `delay` combinator


rkuhn:
Using mapAsync and pattern.after is rather limiting and inefficient, this stage should only have one outstanding timer at all times.

drewhk:
You can use GraphStage now for this. Look at Timeouts object for various timeout stages, this can be implemented in a similar fashion.

drewhk:
@agolubev working on this

agolubev:
ready for review

rkuhn:
fixed in #18985

PR #18985, opened 2015-11-21
=str #18556 add delay combinator

agolubev:
Ref #18556 
Seems like trivial. 
Also removed redundant code like `override def onUpstreamFinish(): Unit = completeStage()`

### DIFF
--- a/akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
+++ b/akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
@@ -266,6 +266,16 @@ object TestSubscriber {
     /**
      * Fluent DSL
      *
+     * Expect a stream element during specified time, then timeout.
+     */
+    def expectNext(d: FiniteDuration, element: I): Self = {
+      probe.expectMsg(d, OnNext(element))
+      self
+    }
+
+    /**
+     * Fluent DSL
+     *
      * Expect multiple stream elements.
      */
     @annotation.varargs def expectNext(e1: I, e2: I, es: I*): Self =
--- a/akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
+++ b/akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
@@ -266,7 +266,7 @@ object TestSubscriber {
     /**
      * Fluent DSL
      *
-     * Expect a stream element during specified time, then timeout.
+     * Expect a stream element during specified time or timeout.
      */
     def expectNext(d: FiniteDuration, element: I): Self = {
       probe.expectMsg(d, OnNext(element))
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowDelaySpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowDelaySpec.scala
@@ -1,0 +1,42 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.scaladsl
+
+import akka.stream.ActorMaterializer
+import akka.stream.testkit.scaladsl.TestSink
+import akka.stream.testkit.{ AkkaSpec, TestPublisher, TestSubscriber }
+import akka.stream.testkit.Utils._
+
+import scala.concurrent.duration._
+
+class FlowDelaySpec extends AkkaSpec {
+
+  implicit val materializer = ActorMaterializer()
+
+  "A Delay" must {
+
+    "deliver element after time passed" in {
+      Source(1 to 3).delay(300.millis).runWith(TestSink.probe[Int])
+        .request(3)
+        .expectNoMsg(100.millis)
+        .expectNext(1)
+        .expectNoMsg(100.millis)
+        .expectNext(2)
+        .expectNoMsg(100.millis)
+        .expectNext(3)
+        .expectComplete()
+    }
+
+    "deliver buffered elements onComplete before the timeout" in assertAllStagesStopped {
+      val c = TestSubscriber.manualProbe[Int]()
+      Source(1 to 3).delay(300.millis).to(Sink(c)).run()
+      val cSub = c.expectSubscription()
+      c.expectNoMsg(200.millis)
+      cSub.request(100)
+      (1 to 3) foreach { n ⇒ c.expectNext(n) }
+      c.expectComplete()
+      c.expectNoMsg(200.millis)
+    }
+  }
+}
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowDelaySpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowDelaySpec.scala
@@ -3,40 +3,116 @@
  */
 package akka.stream.scaladsl
 
-import akka.stream.ActorMaterializer
+import akka.stream.Attributes._
+import akka.stream.testkit.Utils._
 import akka.stream.testkit.scaladsl.TestSink
 import akka.stream.testkit.{ AkkaSpec, TestPublisher, TestSubscriber }
-import akka.stream.testkit.Utils._
+import akka.stream.{ DelayOverflowStrategy, ActorMaterializer }
 
+import scala.concurrent.Await
 import scala.concurrent.duration._
+import scala.util.control.NoStackTrace
 
 class FlowDelaySpec extends AkkaSpec {
 
   implicit val materializer = ActorMaterializer()
 
   "A Delay" must {
+    "deliver elements with some time shift" in {
+      Await.result(
+        Source(1 to 10).delay(1.seconds).grouped(100).runWith(Sink.head),
+        1200.millis) should ===(1 to 10)
+    }
 
-    "deliver element after time passed" in {
+    "deliver element after time passed from actual receiving element" in {
       Source(1 to 3).delay(300.millis).runWith(TestSink.probe[Int])
-        .request(3)
-        .expectNoMsg(100.millis)
-        .expectNext(1)
-        .expectNoMsg(100.millis)
-        .expectNext(2)
-        .expectNoMsg(100.millis)
-        .expectNext(3)
+        .request(2)
+        .expectNoMsg(200.millis) //delay
+        .expectNext(200.millis, 1) //delayed element
+        .expectNext(100.millis, 2) //buffered element
+        .expectNoMsg(200.millis)
+        .request(1)
+        .expectNext(3) //buffered element
         .expectComplete()
     }
 
-    "deliver buffered elements onComplete before the timeout" in assertAllStagesStopped {
+    "deliver elements with delay for slow stream" in assertAllStagesStopped {
       val c = TestSubscriber.manualProbe[Int]()
-      Source(1 to 3).delay(300.millis).to(Sink(c)).run()
+      val p = TestPublisher.manualProbe[Int]()
+
+      Source(p).delay(300.millis).to(Sink(c)).run()
       val cSub = c.expectSubscription()
-      c.expectNoMsg(200.millis)
+      val pSub = p.expectSubscription()
       cSub.request(100)
-      (1 to 3) foreach { n ⇒ c.expectNext(n) }
-      c.expectComplete()
+      pSub.sendNext(1)
       c.expectNoMsg(200.millis)
+      c.expectNext(1)
+      pSub.sendNext(2)
+      c.expectNoMsg(200.millis)
+      c.expectNext(2)
+      pSub.sendComplete()
+      c.expectComplete()
+    }
+
+    "drop tail for internal buffer if it's full in DropTail mode" in assertAllStagesStopped {
+      Await.result(
+        Source(1 to 20).delay(1.seconds, DelayOverflowStrategy.dropTail).withAttributes(inputBuffer(16, 16))
+          .grouped(100)
+          .runWith(Sink.head),
+        1200.millis) should ===((1 to 15).toList :+ 20)
+    }
+
+    "drop head for internal buffer if it's full in DropHead mode" in assertAllStagesStopped {
+      Await.result(
+        Source(1 to 20).delay(1.seconds, DelayOverflowStrategy.dropHead).withAttributes(inputBuffer(16, 16))
+          .grouped(100)
+          .runWith(Sink.head),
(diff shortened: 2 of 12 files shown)


ktoso:
OK TO TEST

drewhk:
This is not necessarily what I would consider to be delay semantics, this is more like a minimum-time-between operation. To demonstrate what I mean:

 - assume original stream emits elements A and B at times 1 and 5
 - delay is 10
 - this op will emit A at 11 and B at 21 (ignoring element passing costs), since it schedules the 10 seconds delay for B *after* A has been emitted, and not when B was actually received.

Now the delay semantics that require an element to be scheduled to be around T + Delta (assuming backpressure does not interfere) can be only obtained by an unbounded internal buffer. This is quite costly, and backpressure might mess up the accurate delaying anyway, nevertheless, this is the more intuitive delay operation.

I am not sure what are the right semantics to aim for here. @rkuhn @viktorklang any opinions?

rkuhn:
I agree with @drewhk, the delay needs to start ticking upon reception of each element. The buffer of course needs to be bounded, which then means that the user needs to select how to react to when it runs full. The choices are

* emit early
* drop
* fail
* back-pressure

The last one may or may not make sense, depending on the use-case, but it does have clearly defined semantics (and it also degenerates into a rate limiter plus delay).

drewhk:
I am not sure we need to provide more options than early and backpressure, since a buffer can be always attached in front of it. In fact this might mean that a simple buffer of one might be just enough.

rkuhn:
good point

agolubev:
Make sense, I was looking here: http://reactivex.io/documentation/operators/delay.html
But current ticket is about minimum rate for emitting elements. 
Ok, I'll will rework this

drewhk:
> But current ticket is about minimum rate for emitting elements. 

The ticket is about delaying... What do I miss here?

> Make sense, I was looking here: http://reactivex.io/documentation/operators/delay.html

Interestinly enough, their diagram illustrates what me and @rkuhn proposed, while their description directly contradicts that: "The Delay operator modifies its source Observable by pausing for a particular increment of time (that you specify) before emitting each of the source Observable’s items.". If their diagram would depict what they just said then every marble would be at least delay space apart, which they aren't.

drewhk:
Ah, because in Rx they don't need to request for next. But that means they will need many timers

agolubev:
@drewhk, actually the question is should delay starting point be when:
1) last element pushed (when A sends downstream in you example)
2) when received current element to push (when we get B)
The second one is what PR is doing now.

drewhk:
@agolubev we proposed 2, but with the difference that you need to request elements in advance, othewise you would never learn their true arrival time.

drewhk:
To give a small test case:

```
Source(1 to 10).delay(10.seconds).completionTimeout(11.seconds)
```

should not throw, since every element is delayed compared to its true arrival (i.e. just as if the delay operation would not be there). With this PR it would take 100 seconds to finish because there would be 10 seconds delay between *each* element.

agolubev:
Ok, got it. 
>In fact this might mean that a simple internal buffer of one might be just enough.

Per my understanding we need to store elements during delay time somewhere. So not sure how to operate external (upstream?) buffer in this case. Are there any example in existing code base I can refer to?

agolubev:
I'm thinking about having buffer functionality (with overflow strategy) separated. And then use it in Buffer detached state and in delay GraphStage

drewhk:
Unfortunately I don't think a separate buffer will work (I was wrong with that), so probably the best is to have a simple Array that can hold up to N elements with their arrival timestamp while the next element is being delayed. Then use the drop strategy to decide what to do (as @rkuhn proposed).

agolubev:
@drewhk, please take a look when you have time. 
- I added strategies for delay
- now like API more, still not sure if need to add buffer size as a parameter
- added trait to mark OverflowStrategies as Delay related, this change failed few tests with OverflowSrategy  serialization check failed. Also this change caused warning (fixed with suppressing this) for match expression with OverflowStrategy.

agolubev:
Build failed as failed Auto-merging akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -266,6 +266,16 @@ object TestSubscriber {
     /**
      * Fluent DSL
      *
+     * Expect a stream element during specified time, then timeout.
```
drewhk: or instead of then

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowDelaySpec.scala
```diff
@@ -0,0 +1,118 @@
...
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import scala.util.control.NoStackTrace
+
+class FlowDelaySpec extends AkkaSpec {
+
+  implicit val materializer = ActorMaterializer()
+
+  "A Delay" must {
```
drewhk: add also a test of `Source(1 to 10).initialDelay()` which will delay the initial element of the source (but not others). Then you can see if your `delay` combinator properly adds time to that initial delay.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -857,13 +858,89 @@ private[stream] class GroupedWithin[T](n: Int, d: FiniteDuration) extends GraphS
...
+            grabElementAndSchedule(true)
+          case DelayOverflowStrategy.DropNew ⇒
+            grab(in)
+            if (!isTimerActive(timerName)) scheduleOnce(timerName, d)
+          case DelayOverflowStrategy.DropBuffer ⇒
+            buffer.clear()
+            grabElementAndSchedule(true)
+          case DelayOverflowStrategy.Fail ⇒
+            failStage(new DelayOverflowStrategy.Fail.BufferOverflowException(s"Buffer overflow for delay combinator (max capacity was: $maxBuffer)!"))
+          case DelayOverflowStrategy.Backpressure ⇒ //do nothing here
```
drewhk: this case is a bug, so this should throw an exception indicating that this should not happen.

akka-stream/src/main/scala/akka/stream/OverflowStrategy.scala
```diff
@@ -6,42 +6,45 @@ package akka.stream
...
+sealed abstract class OverflowStrategy extends Serializable
+sealed trait DelayOverflowStrategy extends Serializable
 
-object OverflowStrategy {
+private[akka] trait BaseOverflowStrategy {
 
   /**
    * INTERNAL API
    */
-  private[akka] final case object DropHead extends OverflowStrategy
```
drewhk: why remove the final?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -857,13 +858,89 @@ private[stream] class GroupedWithin[T](n: Int, d: FiniteDuration) extends GraphS
...
+        }
+        else {
+          grabElementAndSchedule(strategy != DelayOverflowStrategy.Backpressure)
+        }
+      }
+
+      def grabElementAndSchedule(pullCondition: Boolean): Unit = {
+        buffer.enqueue((System.currentTimeMillis(), grab(in)))
+        if (!isTimerActive(timerName)) scheduleOnce(timerName, d)
```
drewhk: why do we need to schedule a timer here?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -857,13 +858,89 @@ private[stream] class GroupedWithin[T](n: Int, d: FiniteDuration) extends GraphS
...
+            if (!isTimerActive(timerName)) scheduleOnce(timerName, d)
+          case DelayOverflowStrategy.DropBuffer ⇒
+            buffer.clear()
+            grabElementAndSchedule(true)
+          case DelayOverflowStrategy.Fail ⇒
+            failStage(new DelayOverflowStrategy.Fail.BufferOverflowException(s"Buffer overflow for delay combinator (max capacity was: $maxBuffer)!"))
+          case DelayOverflowStrategy.Backpressure ⇒ //do nothing here
+        }
+        else {
+          grabElementAndSchedule(strategy != DelayOverflowStrategy.Backpressure)
```
drewhk: it should basically alway pull to the buffer, except when it is full && Backpressure.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -857,13 +858,89 @@ private[stream] class GroupedWithin[T](n: Int, d: FiniteDuration) extends GraphS
...
+        }
+      }
```
drewhk: shouldn't a timer be fired if the buffer was empty just before the element arrived?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
agolubev: yeah - got it - so will have backpressure only on full buffer

akka-stream/src/main/scala/akka/stream/OverflowStrategy.scala
agolubev: IntelliJ told it's redundant, it seemed for me logical

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
agolubev: refactor this a little. now it's scheduing timer when buffer is not full

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
agolubev: not sure I'm following you. Timer is pushing elements and schedule next time if buffer is not empty. So not sure why need to call onTimer before element arrived.

akka-stream/src/main/scala/akka/stream/OverflowStrategy.scala
hepin1989: @agolubev I asked the question once and akka team said that they mark it final for internal thing,

agolubev: Yeah, will look deeply into this on weekend.

drewhk: no I didn't see it is a case object. I thought it is a case class.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -857,13 +858,89 @@ private[stream] class GroupedWithin[T](n: Int, d: FiniteDuration) extends GraphS
...
+
+      override def onUpstreamFinish(): Unit = {
+        if (isAvailable(out) && isTimerActive(timerName)) willStop = true
+        else completeStage()
+      }
+    })
+
+    setHandler(out, new OutHandler {
+      override def onPull(): Unit = {
+        if (!isTimerActive(timerName) && !buffer.isEmpty && nextElementWaitTime() < 0)
```
drewhk: I would give `nextElementWaitTime` a small threshold since it is not worth to fire a timer of 10ms resolution when the waiting time is less than 10ms.

drewhk:
LGTM. Please rebase.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
viktorklang: The problem is that you can't make any assumption on the Scheduler implementation, unless you want to introspect the config for the LARS and check its tick-duration?

drewhk: I still find it overkill to schedule timers for such a short delay. In worst case, you can schedule a timer for every message even if they arrived basically subsequently. We can accept an explicit resolution parameter, but the default of 10ms is not arbitrary since this is what you can reasonably expect from most OSs. Of course multimedia timers might be more precise, but that is a bit out of scope.

viktorklang: Where can I find the reference for the 10ms reasonability? I thought high-precision times could have sub-ms precision?

drewhk: > For example, for Windows running on an x86 processor, the default interval between system clock ticks is typically about 15 milliseconds,

(https://msdn.microsoft.com/en-us/library/windows/hardware/dn265247(v=vs.85).aspx)

To get more resolution there are other Timer versions for higher resolution but you cannot rely on this on the JVM. (at least for a while)

viktorklang: Speaking of which, this code in this PR ought to use nanoTime over currentTimeMillis for resolution purposes.

drewhk: Also: 
> Practical clocks tick at a finite rate, with rates of 100 hertz and 1000 hertz being common. The inverse of this tick rate is the clock resolution, also called the clock granularity, which in either case is expressed as a time duration, being 10 milliseconds and 1 millisecond respectively for these common rates

(timer_settimer manpage: http://linux.die.net/man/3/timer_settime)

drewhk: yes nanoTime is a good idea.

viktorklang: I think a good compromise is to cap the precision of delay to 10ms and document that properly, and then we'll see if anyone complains.

drewhk: The nice thing is that delay will not accumulate delay errors though, since eventually the error gets larger than the 10ms threshold, which will trigger a timer that will just automatically compensate previous errors (since we know the exact arrival time of elements).

viktorklang: Hehe, it will *converge* on correctness if the stream is long enough ;)

agolubev:
@drewhk, ready for review
build failed with CONFLICT (content): Merge conflict in akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala

PR #18986, opened 2015-11-21
=htc add test to verify URI against a very long path

ktoso:
Was reported in Spray gitter channel, spray's URI throws an StackOverflowException for such URI.
Adding the test to make our URI survives it (it does), and that it stays this way.


ktoso:
See https://gitter.im/spray/spray?at=5650a0e7051b93d13c8d659d

Issue #18996, opened 2015-11-23
[akka] japi.PFBuilder matchAny should take declared input type I, not Object


jypma:
Currently PFBuilder.matchAny takes a lambda with Object as input argument. This loses type information for partial functions that are not akka receive functions, e.g. exception handlers:
```
new PFBuilder<Throwable,HttpResponse>()
            .matchAny(x -> {
                log.error("Unhandled exception", x.getCause()); // compile error, since x is Object
```

https://github.com/akka/akka/pull/18995 fixes this by declaring `matchAny`'s lambda to take `I` instead of `Object`.

jypma:
New PR created against master.

PR #18998, opened 2015-11-23
PFBuilder matchAny should take declared input type I

jypma:
Currently PFBuilder.matchAny takes a lambda with Object as input argument.
This loses type information for partial functions that are not akka
receive functions, e.g. exception handlers (PF<Exception,RouteResult).
With this change, the fallback handler still has the guaranteed type I
visible on its lambda.

Fixes https://github.com/akka/akka/issues/18996 .

### DIFF
--- a/akka-actor-tests/src/test/java/akka/japi/pf/PFBuilderTest.java
+++ b/akka-actor-tests/src/test/java/akka/japi/pf/PFBuilderTest.java
@@ -1,0 +1,24 @@
+/**
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.japi.pf;
+
+import org.junit.Test;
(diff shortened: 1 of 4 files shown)


ktoso:
OK TO TEST

ktoso:
I hope this change is not binary incompatible, we'll see (tooling will tell us)

jypma:
It shouldn't be. Anything that's `<I>` with no type bounds becomes `Object` at the class level anyways. 

Besides, `PFBuilder` does state it's "EXPERIMENTAL API" in the header :-)

ktoso:
`EXPERIMENTAL API` That's a life saver ;-)

ktoso:
LGTM as far as I can see.
Might want to include in 2.4.1 // cc @patriknw

patriknw:
LGTM

PR #19004, opened 2015-11-24
=doc,htc better example docs on mapping error response to empty JSON …

ktoso:
better example docs on mapping error response to empty JSON body.

Came out of helping a customer solve this issue.


drewhk:
LGTM

ktoso:
Thanks :)

PR #19006, opened 2015-11-24
groupBy shall emit SubFlow

### LINKED ISSUE #19005: groupBy/splitWhen/splitAfter shall emit SubFlow
Emitting a stream of Sources is just too powerful, which also means boilerplate code for the users. Instead we should have the DSL drop down into substream mode after this operator, staying in that mode until substreams are merged back or closed with a Sink.

rkuhn:
fixes #19005 

will squash/rebase after review

### DIFF
--- a/akka-docs-dev/rst/java/migration-guide-1.0-2.x-java.rst
+++ b/akka-docs-dev/rst/java/migration-guide-1.0-2.x-java.rst
@@ -322,6 +322,62 @@ Update procedure
 *There is no simple update procedure. The affected stages must be ported to the new ``GraphStage`` DSL manually. Please
 read the* ``GraphStage`` *documentation (TODO) for details.*
 
+GroupBy, SplitWhen and SplitAfter now return SubFlow or SubSource
+=================================================================
+
+Previously the ``groupBy``, ``splitWhen``, and ``splitAfter`` combinators
+returned a type that included a :class:`Source` within its elements.
+Transforming these substreams was only possible by nesting the respective
+combinators inside a ``map`` of the outer stream. This has been made more
+convenient and also safer by dropping down into transforming the substreams
+instead: the return type is now a :class:`SubSource` (for sources) or a
+:class:`SubFlow` (for flows) that does not implement the :class:`Graph`
+interface and therefore only represents an unfinished intermediate builder
+step.
+
+Update Procedure
+----------------
+
+The transformations that were done on the substreams need to be lifted up one
+level. This only works for cases where the processing topology is homogenous
+for all substreams.
+
+Example
+^^^^^^^
+
+::
+
+  Flow.<Integer> create()
+    // This no longer works!
+    .groupBy(i -> i % 2)
+    // This no longer works!
+    .map(pair -> pair.second().map(i -> i + 3))
+    // This no longer works!
+    .flatten(FlattenStrategy.concat())
+
+This is implemented now as
+
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/MigrationsJava.java#group-flatten
+
+Example 2
+^^^^^^^^^
+
+::
+
+  Flow.<String> create()
+    // This no longer works!
+    .groupBy(i -> i)
+    // This no longer works!
+    .map(pair ->
+      pair.second().runFold(new Pair<>(pair.first(), 0),
+                            (pair, word) -> new Pair<>(word, pair.second() + 1)))
+    // This no longer works!
+    .mapAsyncUnordered(4, i -> i)
+
+This is implemented now as
+
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/MigrationsJava.java#group-fold
+
 Semantic change in ``isHoldingUpstream`` in the DetachedStage DSL
 =================================================================
 
@@ -437,7 +493,7 @@ should be replaced by:
 .. includecode:: code/docs/MigrationsJava.java#query-param
 
 SynchronousFileSource and SynchronousFileSink
-============================================
+=============================================
 
 Both have been replaced by ``Source.file(…)`` and ``Sink.file(…)`` due to discoverability issues
 paired with names which leaked internal implementation details.
--- a/akka-docs-dev/rst/java/stream-cookbook.rst
+++ b/akka-docs-dev/rst/java/stream-cookbook.rst
@@ -107,23 +107,28 @@ Implementing reduce-by-key
 elements.
 
 The "hello world" of reduce-by-key style operations is *wordcount* which we demonstrate below. Given a stream of words
-we first create a new stream ``wordStreams`` that groups the words according to the ``i -> i`` function, i.e. now
+we first create a new stream that groups the words according to the ``i -> i`` function, i.e. now
 we have a stream of streams, where every substream will serve identical words.
 
-To count the words, we need to process the stream of streams (the actual groups containing identical words). By mapping
-over the groups and using ``fold`` (remember that ``fold`` automatically materializes and runs the stream it is used
-on) we get a stream with elements of ``Future[String,Int]``. Now all we need is to flatten this stream, which
-can be achieved by calling ``mapAsync`` with ``i -> i`` identity function.
+To count the words, we need to process the stream of streams (the actual groups
+containing identical words). ``groupBy`` returns a :class:`SubSource`, which
+means that we transform the resulting substreams directly. In this case we use
+the ``fold`` combinator to aggregate the word itself and the number of its
+occurrences within a :class:`Pair<String, Integer>`. Each substream will then
+emit one final value—precisely such a pair—when the overall input completes. As
+a last step we merge back these values from the substreams into one single
+output stream.
 
-There is one tricky issue to be noted here. The careful reader probably noticed that we put a ``buffer`` between the
-``mapAsync()`` operation that flattens the stream of futures and the actual stream of futures. The reason for this is
-that the substreams produced by ``groupBy()`` can only complete when the original upstream source completes. This means
-that ``mapAsync()`` cannot pull for more substreams because it still waits on folding futures to finish, but these
-futures never finish if the additional group streams are not consumed. This typical deadlock situation is resolved by
-this buffer which either able to contain all the group streams (which ensures that they are already running and folding)
-or fails with an explicit failure instead of a silent deadlock.
+One noteworthy detail pertains to the ``MAXIMUM_DISTINCT_WORDS`` parameter:
+this defines the breadth of the merge operation. Akka Streams is focused on
+bounded resource consumption and the number of concurrently open inputs to the
+merge operator describes the amount of resources needed by the merge itself.
+Therefore only a finite number of substreams can be active at any given time.
+If the ``groupBy`` operator encounters more keys than this number then the
+stream cannot continue without violating its resource bound, in this case
+``groupBy`` will terminate with a failure.
 
-.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKey.java#word-count
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKeyTest.java#word-count
 
 By extracting the parts specific to *wordcount* into
 
@@ -133,13 +138,14 @@ By extracting the parts specific to *wordcount* into
 
 we get a generalized version below:
 
-.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKey.java#reduce-by-key-general
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKeyTest.java#reduce-by-key-general
 
-.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKey.java#reduce-by-key-general2
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKeyTest.java#reduce-by-key-general2
 
 .. note::
-  Please note that the reduce-by-key version we discussed above is sequential, in other words it is **NOT** a
-  parallelization pattern like mapReduce and similar frameworks.
+  Please note that the reduce-by-key version we discussed above is sequential
+  in reading the overall input stream, in other words it is **NOT** a
+  parallelization pattern like MapReduce and similar frameworks.
 
 Sorting elements to multiple groups with groupBy
 ------------------------------------------------
@@ -150,12 +156,12 @@ Sometimes we want to map elements into multiple groups simultaneously.
 To achieve the desired result, we attack the problem in two steps:
 
 * first, using a function ``topicMapper`` that gives a list of topics (groups) a message belongs to, we transform our
-  stream of ``Message`` to a stream of ``(Message, Topic)`` where for each topic the message belongs to a separate pair
+  stream of ``Message`` to a stream of :class:`Pair<Message, Topic>`` where for each topic the message belongs to a separate pair
   will be emitted. This is achieved by using ``mapConcat``
 * Then we take this new stream of message topic pairs (containing a separate pair for each topic a given message
   belongs to) and feed it into groupBy, using the topic as the group key.
 
-.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeMultiGroupBy.java#multi-groupby
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeMultiGroupByTest.java#multi-groupby
 
 Working with Graphs
 ===================
--- a/akka-docs-dev/rst/scala/code/docs/MigrationsScala.scala
+++ b/akka-docs-dev/rst/scala/code/docs/MigrationsScala.scala
@@ -116,6 +116,21 @@ class MigrationsScala extends AkkaSpec {
         Flow[Source[Int, Any]].flatMapConcat(identity)
         //#flatMapConcat
 
+        //#group-flatten
+        Flow[Int]
+          .groupBy(2, _ % 2) // the first parameter sets max number of substreams
+          .map(_ + 3)
+          .concatSubstreams
+        //#group-flatten
+
+        val MaxDistinctWords = 1000
+        //#group-fold
+        Flow[String]
+          .groupBy(MaxDistinctWords, identity)
+          .fold(("", 0))((pair, word) => (word, pair._2 + 1))
+          .mergeSubstreams
+        //#group-fold
+
         //#port-async
         class MapAsyncOne[In, Out](f: In ⇒ Future[Out])(implicit ec: ExecutionContext)
           extends GraphStage[FlowShape[In, Out]] {
--- a/akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeReduceByKey.scala
+++ b/akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeReduceByKey.scala
@@ -1,10 +1,10 @@
 package docs.stream.cookbook
 
-import akka.stream.OverflowStrategy
+import akka.stream.{ Graph, FlowShape, Inlet, Outlet, Attributes, OverflowStrategy }
 import akka.stream.scaladsl._
-
 import scala.concurrent.{ Await, Future }
 import scala.concurrent.duration._
+import akka.stream.stage.{ GraphStage, GraphStageLogic }
 
 class RecipeReduceByKey extends RecipeSpec {
 
@@ -17,22 +17,15 @@ class RecipeReduceByKey extends RecipeSpec {
       def words = Source(List("hello", "world", "and", "hello", "universe", "akka") ++ List.fill(1000)("rocks!"))
 
       //#word-count
-      // split the words into separate streams first
-      val wordStreams: Source[(String, Source[String, Unit]), Unit] = words.groupBy(identity)
-
-      // add counting logic to the streams
-      val countedWords: Source[Future[(String, Int)], Unit] = wordStreams.map {
-        case (word, wordStream) =>
-          wordStream.runFold((word, 0)) {
-            case ((w, count), _) => (w, count + 1)
-          }
-      }
-
-      // get a stream of word counts
-      val counts: Source[(String, Int), Unit] =
-        countedWords
-          .buffer(MaximumDistinctWords, OverflowStrategy.fail)
-          .mapAsync(4)(identity)
+      val counts: Source[(String, Int), Unit] = words
+        // split the words into separate streams first
+        .groupBy(MaximumDistinctWords, identity)
+        // add counting logic to the streams
+        .fold(("", 0)) {
+          case ((_, count), word) => (word, count + 1)
+        }
+        // get a stream of word counts
+        .mergeSubstreams
       //#word-count
 
       Await.result(counts.grouped(10).runWith(Sink.head), 3.seconds).toSet should be(Set(
@@ -54,15 +47,17 @@ class RecipeReduceByKey extends RecipeSpec {
         groupKey: (In) => K,
         foldZero: (K) => Out)(fold: (Out, In) => Out): Flow[In, (K, Out), Unit] = {
 
-        val groupStreams = Flow[In].groupBy(groupKey)
-        val reducedValues = groupStreams.map {
-          case (key, groupStream) =>
-            groupStream.runFold((key, foldZero(key))) {
-              case ((key, aggregated), elem) => (key, fold(aggregated, elem))
-            }
-        }
-
-        reducedValues.buffer(maximumGroupSize, OverflowStrategy.fail).mapAsync(4)(identity)
+        Flow[In]
+          .groupBy(maximumGroupSize, groupKey)
+          .fold(Option.empty[(K, Out)]) {
+            case (None, elem) =>
+              val key = groupKey(elem)
+              Some((key, fold(foldZero(key), elem)))
+            case (Some((key, out)), elem) =>
+              Some((key, fold(out, elem)))
+          }
+          .map(_.get)
+          .mergeSubstreams
       }
 
       val wordCounts = words.via(reduceByKey(
--- a/akka-docs-dev/rst/scala/stream-testkit.rst
+++ b/akka-docs-dev/rst/scala/stream-testkit.rst
@@ -4,60 +4,89 @@
 Testing streams
 ###############
 
-Verifying behaviour of Akka Stream sources, flows and sinks can be done using various code patterns and libraries. Here we will discuss testing these elements using:
+Verifying behaviour of Akka Stream sources, flows and sinks can be done using
+various code patterns and libraries. Here we will discuss testing these
+elements using:
 
 - simple sources, sinks and flows;
 - sources and sinks in combination with :class:`TestProbe` from the :mod:`akka-testkit` module;
 - sources and sinks specifically crafted for writing tests from the :mod:`akka-stream-testkit` module.
 
-It is important to keep your data processing pipeline as separate sources, flows and sinks. This makes them easily testable by wiring them up to other sources or sinks, or some test harnesses that :mod:`akka-testkit` or :mod:`akka-stream-testkit` provide.
+It is important to keep your data processing pipeline as separate sources,
+flows and sinks. This makes them easily testable by wiring them up to other
+sources or sinks, or some test harnesses that :mod:`akka-testkit` or
+:mod:`akka-stream-testkit` provide.
 
 Built in sources, sinks and combinators
 =======================================
 
-Testing a custom sink can be as simple as attaching a source that emits elements from a predefined collection, running a constructed test flow and asserting on the results that sink produced. Here is an example of a test for a sink:
+Testing a custom sink can be as simple as attaching a source that emits
+elements from a predefined collection, running a constructed test flow and
+asserting on the results that sink produced. Here is an example of a test for a
+sink:
 
 .. includecode:: code/docs/stream/StreamTestKitDocSpec.scala#strict-collection
 
-The same strategy can be applied for sources as well. In the next example we have a source that produces an infinite stream of elements. Such source can be tested by asserting that first arbitrary number of elements hold some condition. Here the :code:`grouped` combinator and :code:`Sink.head` are very useful.
+The same strategy can be applied for sources as well. In the next example we
+have a source that produces an infinite stream of elements. Such source can be
+tested by asserting that first arbitrary number of elements hold some
+condition. Here the ``grouped`` combinator and ``Sink.head`` are very useful.
 
 .. includecode:: code/docs/stream/StreamTestKitDocSpec.scala#grouped-infinite
 
-When testing a flow we need to attach a source and a sink. As both stream ends are under our control, we can choose sources that tests various edge cases of the flow and sinks that ease assertions.
+When testing a flow we need to attach a source and a sink. As both stream ends
+are under our control, we can choose sources that tests various edge cases of
+the flow and sinks that ease assertions.
 
 .. includecode:: code/docs/stream/StreamTestKitDocSpec.scala#folded-stream
 
 TestKit
 =======
 
-Akka Stream offers integration with Actors out of the box. This support can be used for writing stream tests that use familiar :class:`TestProbe` from the :mod:`akka-testkit` API.
+Akka Stream offers integration with Actors out of the box. This support can be
+used for writing stream tests that use familiar :class:`TestProbe` from the
+:mod:`akka-testkit` API.
 
-One of the more straightforward tests would be to materialize stream to a :class:`Future` and then use :code:`pipe` pattern to pipe the result of that future to the probe.
+One of the more straightforward tests would be to materialize stream to a
+:class:`Future` and then use ``pipe`` pattern to pipe the result of that future
+to the probe.
 
 .. includecode:: code/docs/stream/StreamTestKitDocSpec.scala#pipeto-testprobe
 
-Instead of materializing to a future, we can use a :class:`Sink.actorRef` that sends all incoming elements to the given :class:`ActorRef`. Now we can use assertion methods on :class:`TestProbe` and expect elements one by one as they arrive. We can also assert stream completion by expecting for :code:`onCompleteMessage` which was given to :class:`Sink.actorRef`.
+Instead of materializing to a future, we can use a :class:`Sink.actorRef` that
+sends all incoming elements to the given :class:`ActorRef`. Now we can use
+assertion methods on :class:`TestProbe` and expect elements one by one as they
+arrive. We can also assert stream completion by expecting for
+``onCompleteMessage`` which was given to :class:`Sink.actorRef`.
 
 .. includecode:: code/docs/stream/StreamTestKitDocSpec.scala#sink-actorref
 
-Similarly to :class:`Sink.actorRef` that provides control over received elements, we can use :class:`Source.actorRef` and have full control over elements to be sent.
+Similarly to :class:`Sink.actorRef` that provides control over received
+elements, we can use :class:`Source.actorRef` and have full control over
+elements to be sent.
 
 .. includecode:: code/docs/stream/StreamTestKitDocSpec.scala#source-actorref
 
 Streams TestKit
 ===============
 
-You may have noticed various code patterns that emerge when testing stream pipelines. Akka Stream has a separate :mod:`akka-stream-testkit` module that provides tools specifically for writing stream tests. This module comes with two main components that are :class:`TestSource` and :class:`TestSink` which provide sources and sinks that materialize to probes that allow fluent API.
+You may have noticed various code patterns that emerge when testing stream
+pipelines. Akka Stream has a separate :mod:`akka-stream-testkit` module that
+provides tools specifically for writing stream tests. This module comes with
+two main components that are :class:`TestSource` and :class:`TestSink` which
+provide sources and sinks that materialize to probes that allow fluent API.
 
 .. note::
 
    Be sure to add the module :mod:`akka-stream-testkit` to your dependencies.
 
-A sink returned by :code:`TestSink.probe` allows manual control over demand and assertions over elements coming downstream.
+A sink returned by ``TestSink.probe`` allows manual control over demand and
+assertions over elements coming downstream.
 
 .. includecode:: code/docs/stream/StreamTestKitDocSpec.scala#test-sink-probe
 
-A source returned by :code:`TestSource.probe` can be used for asserting demand or controlling when stream is completed or ended with an error.
+A source returned by ``TestSource.probe`` can be used for asserting demand or
+controlling when stream is completed or ended with an error.
 
 .. includecode:: code/docs/stream/StreamTestKitDocSpec.scala#test-source-probe
 
--- a/akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKeyTest.java
+++ b/akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKeyTest.java
@@ -15,14 +15,21 @@ import akka.japi.function.Function;
 import akka.japi.function.Function2;
 import akka.testkit.JavaTestKit;
 import org.junit.AfterClass;
+import org.junit.Assert;
 import org.junit.BeforeClass;
 import org.junit.Test;
+
+import scala.concurrent.Await;
 import scala.concurrent.Future;
 import scala.runtime.BoxedUnit;
 
 import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 
-public class RecipeReduceByKey extends RecipeTest {
+public class RecipeReduceByKeyTest extends RecipeTest {
   static ActorSystem system;
 
   @BeforeClass
@@ -46,27 +53,24 @@ public class RecipeReduceByKey extends RecipeTest {
 
         //#word-count
         final int MAXIMUM_DISTINCT_WORDS = 1000;
-
-        // split the words into separate streams first
-        final Source<Pair<String, Source<String, BoxedUnit>>, BoxedUnit> wordStreams = words
-          .groupBy(i -> i);
-
-        // add counting logic to the streams
-        Source<Future<Pair<String, Integer>>, BoxedUnit> countedWords = wordStreams.map(pair -> {
-          final String word = pair.first();
-          final Source<String, BoxedUnit> wordStream = pair.second();
-          return wordStream.runFold(
-            new Pair<>(word, 0),
-            (acc, w) -> new Pair<>(word, acc.second() + 1), mat);
-        });
-
-        // get a stream of word counts
-        final Source<Pair<String, Integer>, BoxedUnit> counts = countedWords
-          .buffer(MAXIMUM_DISTINCT_WORDS, OverflowStrategy.fail())
-          .mapAsync(4, i -> i);
+        
+        final Source<Pair<String, Integer>, BoxedUnit> counts = words
+            // split the words into separate streams first
+          .groupBy(MAXIMUM_DISTINCT_WORDS, i -> i)
+          // add counting logic to the streams
+          .fold(new Pair<>("", 0), (pair, elem) -> new Pair<>(elem, pair.second() + 1))
+          // get a stream of word counts
+          .mergeSubstreams();
         //#word-count
 
-        counts.runWith(Sink.ignore(), mat);
+        final Future<List<Pair<String, Integer>>> f = counts.grouped(10).runWith(Sink.head(), mat);
+        final Set<Pair<String, Integer>> result = Await.result(f, getRemainingTime()).stream().collect(Collectors.toSet());
+        final Set<Pair<String, Integer>> expected = new HashSet<>();
+        expected.add(new Pair<>("hello", 2));
+        expected.add(new Pair<>("world", 1));
+        expected.add(new Pair<>("and", 1));
+        expected.add(new Pair<>("akka", 1));
+        Assert.assertEquals(expected, result);
       }
     };
   }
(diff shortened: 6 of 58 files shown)


akka-docs-dev/rst/java/stream-cookbook.rst
```diff
@@ -107,23 +107,30 @@ Implementing reduce-by-key
...
-``mapAsync()`` operation that flattens the stream of futures and the actual stream of futures. The reason for this is
-that the substreams produced by ``groupBy()`` can only complete when the original upstream source completes. This means
-that ``mapAsync()`` cannot pull for more substreams because it still waits on folding futures to finish, but these
-futures never finish if the additional group streams are not consumed. This typical deadlock situation is resolved by
-this buffer which either able to contain all the group streams (which ensures that they are already running and folding)
-or fails with an explicit failure instead of a silent deadlock.
-
-.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKey.java#word-count
+To count the words, we need to process the stream of streams (the actual groups
+containing identical words). ``GroupBy`` returns a :class:`SubSource`, which
```
drewhk: weird capitalization of GroupBy since in code it is small-caps. I usually try to refactor these kind of sentences so that the first word is not code to sidestep this issue.

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeReduceByKey.scala
```diff
@@ -17,22 +17,15 @@ class RecipeReduceByKey extends RecipeSpec {
...
-      // get a stream of word counts
-      val counts: Source[(String, Int), Unit] =
-        countedWords
-          .buffer(MaximumDistinctWords, OverflowStrategy.fail)
-          .mapAsync(4)(identity)
+      val counts: Source[(String, Int), Unit] = words
+        // split the words into separate streams first
+        .groupBy(identity)
+        // add counting logic to the streams
+        .fold(("", 0)) {
```
drewhk: another reason why a non-terminal `fold` is useful. Thanks @viktorklang for doing the necessary refactor!

akka-stream-tck/src/test/scala/akka/stream/tck/GroupByTest.scala
```diff
@@ -17,7 +17,12 @@ class GroupByTest extends AkkaPublisherVerification[Int] {
     if (elements == 0) EmptyPublisher[Int]
     else {
       val futureGroupSource =
-        Source(iterable(elements)).groupBy(elem ⇒ "all").map { case (_, group) ⇒ group }.runWith(Sink.head)
+        Source(iterable(elements))
+          .groupBy(elem ⇒ "all")
+          .prefixAndTail(0)
```
drewhk: shouldn't this be called `lift` or something? Also, this method should take some kind of licence ;)

rkuhn: That is precisely the reason for:

* not documenting this purpose
* making it involve some boilerplate
* and in general being non-obvious

:-)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowGroupBySpec.scala
```diff
@@ -13,8 +13,21 @@ import akka.stream.testkit.Utils._
...
+import org.scalatest.concurrent.ScalaFutures
+import org.scalactic.ConversionCheckedTripleEquals
+import org.scalatest.concurrent.PatienceConfiguration.Timeout
 
-class FlowGroupBySpec extends AkkaSpec {
+object FlowGroupBySpec {
+  import language.higherKinds
+
+  implicit class Lift[M](val f: SubFlow[Int, M, Source[Int, M]#Repr, RunnableGraph[M]]) extends AnyVal {
+    def lift(key: Int ⇒ Int) = f.prefixAndTail(1).map(p ⇒ key(p._1.head) -> (Source.single(p._1.head) ++ p._2)).mergeBack(1)
```
drewhk: here is my lift ;)

akka-stream/src/main/scala/akka/stream/impl/SubFlowImpl.scala
```diff
@@ -0,0 +1,35 @@
...
+ */
+package akka.stream.impl
+
+import akka.stream._
+import akka.stream.scaladsl._
+
+import language.higherKinds
+
+object SubFlowImpl {
+  trait MergeBack[In, F[+_]] {
```
drewhk: `F[+_]` lol wat?

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -733,12 +732,20 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
...
+   * The object returned from this method is not a normal [[Flow]],
+   * it is a [[SubFlow]]. This means that after this combinator all transformations
+   * are applied to all encountered substreams in the same fashion. Substream mode
+   * is exited either by closing the substream (i.e. connecting it to a [[Sink]])
+   * or by merging the substreams back together; see the `to` and `mergeBack` methods
+   * on [[SubFlow]] for more information.
+   *
+   * It is important to note that the substreams also propagate back-pressure as
+   * any other stream, which means that blocking one substream will block the `groupBy`
+   * operator itself&mdash;and thereby all substreams&mdash; once all internal or
```
drewhk: disagree with the `&mdash`. While it might improve scaladocs, it is not that uncommon to read these comments in code, where it is annoying.

akka-stream/src/main/scala/akka/stream/javadsl/SubFlow.scala
```diff
@@ -0,0 +1,865 @@
...
+   * Flatten the sub-flows back into the super-flow, performing a merge of the
+   * given breadth. For concatenation of the flows use a breadth of 1 but beware
+   * that this will lead to deadlocks if the sub-flows need to make progress
+   * concurrently (as is generally the case when using `groupBy` on an unordered
+   * input stream).
+   *
+   * WARNING: currently only a breadth of 1 is actually implemented! Beware of deadlocks!
+   */
+  // FIXME remove doc warning once fixed
+  def mergeBack(breadth: Int): Flow[In, Out, Mat] =
```
drewhk: shouldn't we create a `concat()` method? We tried to be rather restrictive in the past and making certain dangerous methods/combinations harder to access but then later we reverted many of them. `concat()` conveys intent more than `mergeBack(1)`

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeReduceByKey.scala
viktorklang: The Klang abides :)

akka-docs-dev/rst/java/migration-guide-1.0-2.x-java.rst
```diff
@@ -422,7 +422,7 @@ should be replaced by:
 .. includecode:: code/docs/MigrationsJava.java#query-param
 
 SynchronousFileSource and SynchronousFileSink
-============================================
+=============================================
```
viktorklang: GAME CHANGER

akka-stream/src/main/scala/akka/stream/javadsl/SubFlow.scala
```diff
@@ -0,0 +1,865 @@
...
+   *
+   * By default element and completion signals are logged on debug level, and errors are logged on Error level.
+   * This can be adjusted according to your needs by providing a custom [[Attributes.LogLevels]] attribute on the given Flow.
+   *
+   * Uses an internally created [[LoggingAdapter]] which uses `akka.stream.Log` as it's source (use this class to configure slf4j loggers).
+   */
+  def log(name: String): SubFlow[In, Out, Mat] =
+    this.log(name, ConstantFun.javaIdentityFunction[Out], null)
+
+}
```
drewhk: eh, +865 lines of duplication... Would be nice at least to have some boilerplate plugin like magic to generat these...

akka-stream/src/main/scala/akka/stream/javadsl/SubSource.scala
```diff
@@ -0,0 +1,865 @@
...
+   *
+   * By default element and completion signals are logged on debug level, and errors are logged on Error level.
+   * This can be adjusted according to your needs by providing a custom [[Attributes.LogLevels]] attribute on the given Flow.
+   *
+   * Uses an internally created [[LoggingAdapter]] which uses `akka.stream.Log` as it's source (use this class to configure slf4j loggers).
+   */
+  def log(name: String): SubSource[Out, Mat] =
+    this.log(name, ConstantFun.javaIdentityFunction[Out], null)
+
+}
```
drewhk: ah, 1700 more lines, aaargh :(

akka-stream/src/main/scala/akka/stream/impl/SubFlowImpl.scala
rkuhn: It’s just a monoid in the category of endofunctors … (no, it is actually a covariant type constructor).

viktorklang: It's a type constructor named F with a single type parameter which is covariant.

:sunglasses:

2beaucoup:
There are a lot of `prefixAndTail(1)`s here. Maybe introduce `headAndTail` which returns `(Out, Source[U, Unit])` instead of `(immutable.Seq[Out], Source[U, Unit])` and which spares users the hassle of sequence matching? ;)

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
rkuhn: What do you propose? (and there is an extraneous space character on this line as well … )

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -915,8 +913,36 @@ trait FlowOps[+Out, +Mat] {
...
+        setHandler(out, eagerTerminateOutput)
+        override def preStart(): Unit = pull(in)
+      }
+    }
+    val merge = new SubFlowImpl.MergeBack[Out, Repr] {
+      override def apply[T](flow: Flow[Out, T, Unit], breadth: Int): Repr[T] =
+        deprecatedAndThen[Source[Out, Unit]](GroupBy(f.asInstanceOf[Any ⇒ Any]))
+          .map(_.via(flow).via(makeEager) /* needed to start generating some demand */ )
+          .map(s ⇒ Source(s.runWith(Sink.publisher(false)))) // need to be eager as a workaround for now
+          .buffer(4, OverflowStrategy.Backpressure) // needed to make reduceByKey recipe work
```
drewhk: ? That is weird.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -915,8 +913,36 @@ trait FlowOps[+Out, +Mat] {
...
+      override def createLogic(attr: Attributes) = new GraphStageLogic(shape) {
+        passAlong(in, out, doFinish = true, doFail = true)
+        setHandler(out, eagerTerminateOutput)
+        override def preStart(): Unit = pull(in)
+      }
+    }
+    val merge = new SubFlowImpl.MergeBack[Out, Repr] {
+      override def apply[T](flow: Flow[Out, T, Unit], breadth: Int): Repr[T] =
+        deprecatedAndThen[Source[Out, Unit]](GroupBy(f.asInstanceOf[Any ⇒ Any]))
+          .map(_.via(flow).via(makeEager) /* needed to start generating some demand */ )
```
drewhk: because of the mapConcat right? Otherwise it will not start consuming all subgroups?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: Ah, to feed enough makeEager into the map above?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -915,8 +913,36 @@ trait FlowOps[+Out, +Mat] {
...
+          .map(s ⇒ Source(s.runWith(Sink.publisher(false)))) // need to be eager as a workaround for now
+          .buffer(4, OverflowStrategy.Backpressure) // needed to make reduceByKey recipe work
+          .flatMapConcat(identity)
+          .asInstanceOf[Repr[T]]
+      /*
+       * FIXME remove all those commented workarounds above by implementing flatMapMerge(breadth)
```
drewhk: I see.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: Nope, it is an artifact of using the totally lazy ConcatAll for now: it will not pull any more elements (nested streams) out of the prematerialization `map()` than what its input buffer demands (4) while the sample has 5 substreams … We must fix this for real before doing a release, which means actually implementing a flatMapMerge with configurable breadth.

akka-stream/src/main/scala/akka/stream/impl/SubFlowImpl.scala
drewhk: got it.

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
drewhk: can't you just paste the proper UTF-8 character? Or does Javadoc trip on that?

akka-docs-dev/rst/java/migration-guide-1.0-2.x-java.rst
drewhk: Well, now it will output proper HTML ;) (yes, RST actually does care of the length of those bars)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: that's what I said above. This buffer makes sure that all subgroups pass the map above, that attaches the makeEager stage.

viktorklang: I am concerned by `breadth` as it becomes "dynamically typed" and depending on the Flows before it.
What other options are there?

Case in point:

groupBy(…).groupBy(…).mergeBack(2)

and then someone adds or removes a groupBy but forgets to change the magic number.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -915,8 +913,36 @@ trait FlowOps[+Out, +Mat] {
...
+        override def preStart(): Unit = pull(in)
+      }
+    }
+    val merge = new SubFlowImpl.MergeBack[Out, Repr] {
+      override def apply[T](flow: Flow[Out, T, Unit], breadth: Int): Repr[T] =
+        deprecatedAndThen[Source[Out, Unit]](GroupBy(f.asInstanceOf[Any ⇒ Any]))
+          .map(_.via(flow).via(makeEager) /* needed to start generating some demand */ )
+          .map(s ⇒ Source(s.runWith(Sink.publisher(false)))) // need to be eager as a workaround for now
+          .buffer(4, OverflowStrategy.Backpressure) // needed to make reduceByKey recipe work
+          .flatMapConcat(identity)
```
viktorklang: Use proper identity, the one of constant fun.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -959,8 +997,20 @@ trait FlowOps[+Out, +Mat] {
...
    * See also [[FlowOps.splitAfter]].
    */
-  def splitWhen[U >: Out](p: Out ⇒ Boolean): Repr[Source[U, Unit], Mat] =
-    deprecatedAndThen(Split.when(p.asInstanceOf[Any ⇒ Boolean]))
+  def splitWhen(p: Out ⇒ Boolean): SubFlow[Out, Mat, Repr, Closed] = {
+    val merge = new SubFlowImpl.MergeBack[Out, Repr] {
+      override def apply[T](flow: Flow[Out, T, Unit], breadth: Int): Repr[T] =
+        deprecatedAndThen[Source[Out, Unit]](Split.when(p.asInstanceOf[Any ⇒ Boolean]))
+          .map(_.via(flow))
+          .flatMapConcat(identity)
```
viktorklang: Find and replace use of the wrong identity :)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: @viktorklang Nothing about the structure is dynamic here, the breadth parameter only determines how many substreams are merged concurrently. Your “case in point” would not compile in a context where a normal stream is expected because every `groupBy` needs to be matched by exactly one `mergeBack`.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: This code is broken in more important ways and will be replaced in the next PR :-)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: Ah, ok, then if it is how many streams are merged in concurrently I think it would be nice to have a cohesive name for that parameter value, as it is related to mapAsync/mapAsyncUnordered etc. so mergeBack(1) == concat

drewhk:
Looks very nice so far!

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
clockfly: +1 on Viktor.
mergeBack(1) don't looks good, it may create confusion with merge.
Maybe using name flatten or concat is better.

rkuhn: If you see `mergeBack(1)` in code you are almost always looking at a potential deadlock, so I disagree with the notion that there may be confusion here—what we do is to make the reasonable thing simple and discoverable.

OTOH I am really having doubts on this limit: instead of having people choose a limit beyond which the stream just deadlocks, why don’t we keep the mergeBack unlimited and have people think about limits when invoking groupBy?

akka-docs-dev/rst/java/stream-cookbook.rst
rkuhn: fixed

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
rkuhn: Using the UTF-8 character “works for me” (tm) on Chrome, and I’ll probably just do that, but old browsers will choke on this (and display a replacement glyph or something weird).

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: Having an Int-param on groupBy to denominate max number of concurrent groups sounds like a stroke of genius.

akka-stream/src/main/scala/akka/stream/javadsl/SubFlow.scala
rkuhn: Okay, so here’s what I’ll do:

* remove the parameter and rename to `mergeSubstreams`
* add `concatSubstreams` method
* add `maximumSubstreams` parameter to `groupBy`

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: Let’s just hope that the genius doesn’t die of a stroke soon … ;-)

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
drewhk: let's just say "ancient, no longer supported browsers" instead of "old browsers" ;) I don't think we need to worry about that. If someone reports this problem we will fix it, but I highly doubt that this will be a problem nowadays.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: Not to take away the *genius* marker from @rkuhn but we actually discussed this in the past, it just never have been applied (if I recall because we wanted to fix stream-of-streams stuff properly, once and for all, so we postponed it).

akka-stream/src/main/scala/akka/stream/javadsl/SubFlow.scala
drewhk: `concatSubstreams` should take a `String` parameter that must have the exact value of "I know what I am doing." ;)

viktorklang: It's not `maximumSubstreams` though. It's `maxConcurrentSubstreams`

On Wed, Nov 25, 2015 at 12:24 PM, Roland Kuhn <notifications@github.com>
wrote:

> In akka-stream/src/main/scala/akka/stream/javadsl/SubFlow.scala
> <https://github.com/akka/akka/pull/19006#discussion_r45855773>:
>
> > +class SubFlow[-In, +Out, +Mat](delegate: scaladsl.SubFlow[Out, Mat, scaladsl.Flow[In, Out, Mat]#Repr, scaladsl.Sink[In, Mat]]) {
> > +
> > +  /** Converts this Flow to its Scala DSL counterpart */
> > +  def asScala: scaladsl.SubFlow[Out, Mat, scaladsl.Flow[In, Out, Mat]#Repr, scaladsl.Sink[In, Mat]] @uncheckedVariance = delegate
> > +
> > +  /**
> > +   * Flatten the sub-flows back into the super-flow, performing a merge of the
> > +   * given breadth. For concatenation of the flows use a breadth of 1 but beware
> > +   * that this will lead to deadlocks if the sub-flows need to make progress
> > +   * concurrently (as is generally the case when using `groupBy` on an unordered
> > +   * input stream).
> > +   *
> > +   * WARNING: currently only a breadth of 1 is actually implemented! Beware of deadlocks!
> > +   */
> > +  // FIXME remove doc warning once fixed
> > +  def mergeBack(breadth: Int): Flow[In, Out, Mat] =
>
> Okay, so here’s what I’ll do:
>
>    - remove the parameter and rename to mergeSubstreams
>    - add concatSubstreams method
>    - add maximumSubstreams parameter to groupBy
>
> —
> <https://github.com/akka/akka/pull/19006/files#r45855773>.
>



-- 
Cheers,
√

drewhk: Since groupBy is not able to close any group before the primary stream is closed, it is actually `maximumSubstreams`.

viktorklang: Indeed! I stand corrected.

rkuhn:
documented, squashed and rebased

akka-docs-dev/rst/java/migration-guide-1.0-2.x-java.rst
ktoso: Oh I didn't realise it actually breaks when not aligned. I always fix them because of my OCD anyway :wink:

akka-docs-dev/rst/java/stream-cookbook.rst
```diff
@@ -107,23 +107,30 @@ Implementing reduce-by-key
...
+bounded resource consumption and the number of concurrently open inputs to the
+merge operator describes the amount of resources needed by the merge itself.
+Therefore only a finite number of substreams can be active at any given time.
+If the ``groupBy`` operator emits more than this number then the stream cannot
+continue: in order to write the new word to its new substream there needs to be
+demand on that substream and ``mergeBack`` will not create this demand because
+it cannot consume the resulting elements within its resource bounds. The
+consequence is that the stream will deadlock.
+
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKeyTest.java#word-count
```
ktoso: tempted to add a 1 sentence summary in a warning block here?

akka-docs-dev/rst/java/stream-cookbook.rst
```diff
@@ -133,12 +140,13 @@ By extracting the parts specific to *wordcount* into
...
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKeyTest.java#reduce-by-key-general
 
-.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKey.java#reduce-by-key-general2
+.. includecode:: ../../../akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKeyTest.java#reduce-by-key-general2
 
 .. note::
-  Please note that the reduce-by-key version we discussed above is sequential, in other words it is **NOT** a
+  Please note that the reduce-by-key version we discussed above is sequential
+  in reading the overall input stream, in other words it is **NOT** a
   parallelization pattern like mapReduce and similar frameworks.
```
ktoso: MapReduce or map-reduce?

akka-docs-dev/rst/java/stream-cookbook.rst
```diff
@@ -150,12 +158,12 @@ Sometimes we want to map elements into multiple groups simultaneously.
 To achieve the desired result, we attack the problem in two steps:
 
 * first, using a function ``topicMapper`` that gives a list of topics (groups) a message belongs to, we transform our
-  stream of ``Message`` to a stream of ``(Message, Topic)`` where for each topic the message belongs to a separate pair
+  stream of ``Message`` to a stream of :class:`Pair<Message, Topic>`` where for each topic the message belongs to a separate pair
```
ktoso: good catch, thanks

akka-docs-dev/rst/scala/code/docs/MigrationsScala.scala
```diff
@@ -116,9 +116,24 @@ class MigrationsScala extends AkkaSpec {
         Flow[Source[Int, Any]].flatMapConcat(identity)
         //#flatMapConcat
 
+        //#group-flatten
+        Flow[Int]
+          .groupBy(2, _ % 2) // the first parameter sets max number of substreams
+          .map(_ + 3)
+          .concatSubstreams
```
ktoso: I'm late to the naming party but wanted to say it's a great name :+1:

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKeyTest.java
```diff
@@ -46,27 +53,24 @@ public void work() throws Exception {
...
-        // get a stream of word counts
-        final Source<Pair<String, Integer>, BoxedUnit> counts = countedWords
-          .buffer(MAXIMUM_DISTINCT_WORDS, OverflowStrategy.fail())
-          .mapAsync(4, i -> i);
+        
+        final Source<Pair<String, Integer>, BoxedUnit> counts = words
+            // split the words into separate streams first
+          .groupBy(MAXIMUM_DISTINCT_WORDS, i -> i)
+          // add counting logic to the streams
+          .fold(new Pair<>("", 0), (pair, elem) -> new Pair<>(elem, pair.second() + 1))
```
ktoso: In theory we have `Pair.create` so you don't have to do the diamond.

akka-stream-testkit/src/main/scala/akka/stream/testkit/StreamTestKit.scala
```diff
@@ -42,7 +42,7 @@ object TestPublisher {
   }
 
   /**
-   * Publisher that signals error to subscribers immediately, before handing out subscription.
+   * Publisher that signals error to subscribers immediately after handing out subscription.
```
ktoso: good catch

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -757,9 +764,11 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
    *
    * '''Cancels when''' downstream cancels and all substreams cancel
    *
+   * @param maxSubstreams configures the maximum number of substreams (keys)
+   *        that are supported; if more distinct keys are encountere then the stream fails
```
ktoso: s/encountere/encountered

akka-docs-dev/rst/java/stream-cookbook.rst
```diff
@@ -107,23 +107,30 @@ Implementing reduce-by-key
...
+output stream.
+
+One noteworthy detail pertains to the ``MAXIMUM_DISTINCT_WORDS`` parameter:
+this defines the breadth of the merge operation. Akka Streams is focused on
+bounded resource consumption and the number of concurrently open inputs to the
+merge operator describes the amount of resources needed by the merge itself.
+Therefore only a finite number of substreams can be active at any given time.
+If the ``groupBy`` operator emits more than this number then the stream cannot
+continue: in order to write the new word to its new substream there needs to be
+demand on that substream and ``mergeBack`` will not create this demand because
```
viktorklang: Strongly against `mergeBack` as a name, especially if the method does not exist ;)

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeReduceByKey.scala
```diff
@@ -54,15 +47,17 @@ class RecipeReduceByKey extends RecipeSpec {
...
+        Flow[In]
+          .groupBy(maximumGroupSize, groupKey)
+          .fold(Option.empty[(K, Out)]) {
+            case (None, elem) =>
+              val key = groupKey(elem)
+              Some((key, fold(foldZero(key), elem)))
+            case (Some((key, out)), elem) =>
+              Some((key, fold(out, elem)))
+          }
+          .map(_.get)
```
viktorklang: ?

akka-docs-dev/rst/scala/stream-testkit.rst
```diff
@@ -4,60 +4,89 @@
...
+Testing a custom sink can be as simple as attaching a source that emits
+elements from a predefined collection, running a constructed test flow and
+asserting on the results that sink produced. Here is an example of a test for a
+sink:
 
 .. includecode:: code/docs/stream/StreamTestKitDocSpec.scala#strict-collection
 
-The same strategy can be applied for sources as well. In the next example we have a source that produces an infinite stream of elements. Such source can be tested by asserting that first arbitrary number of elements hold some condition. Here the :code:`grouped` combinator and :code:`Sink.head` are very useful.
+The same strategy can be applied for sources as well. In the next example we
+have a source that produces an infinite stream of elements. Such source can be
```
viktorklang: Should we consistently use `unbounded` iso `infinite`?

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -533,6 +533,14 @@ final class Source[+Out, +Mat](delegate: scaladsl.Source[Out, Mat]) extends Grap
...
    * Transform this stream by applying the given function to each of the elements
    * as they pass through this processing step.
+   *
+   * '''Emits when''' the mapping function returns an element
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
```
ktoso: so... about that genjavadoc improvement... :stuck_out_tongue_winking_eye:

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -541,16 +549,38 @@ final class Source[+Out, +Mat](delegate: scaladsl.Source[Out, Mat]) extends Grap
...
+   * '''Completes when''' upstream completes or upstream failed with exception pf can handle
+   *
+   * '''Cancels when''' downstream cancels
    */
   def recover[T >: Out](pf: PartialFunction[Throwable, T]): javadsl.Source[T, Mat] =
     new Source(delegate.recover(pf))
 
   /**
-   * Transform each input element into a sequence of output elements that is
+   * Transform each input element into an `Iterable` of output elements that is
```
ktoso: technically into a `java.util.List[T]]`

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -541,16 +549,38 @@ final class Source[+Out, +Mat](delegate: scaladsl.Source[Out, Mat]) extends Grap
...
    * then flattened into the output stream.
    *
-   * The returned list MUST NOT contain `null` values,
+   * Make sure that the `Iterable` is immutable or at least not modified after
+   * being used as an output sequence. Otherwise the stream may fail with
+   * `ConcurrentModificationException` or other more subtle errors may occur.
+   *
+   * The returned `Iterable` MUST NOT contain `null` values,
```
ktoso: java.util.List[T]], or change the signature to be Iterable?

akka-stream/src/main/scala/akka/stream/javadsl/SubFlow.scala
```diff
@@ -0,0 +1,868 @@
...
+   *     |  +------+        +------+  |
+   *     |  |      |        |      |  |
+   * In ~~> | this | ~Out~> | flow | ~~> T
+   *     |  |      |        |      |  |
+   *     |  +------+        +------+  |
+   *     +----------------------------+
+   * }}}
+   *
+   * The materialized value of the combined [[Flow]] will be the materialized
+   * value of the current flow (ignoring the other Flow’s value), use
```
ktoso: other => inner?

akka-stream/src/main/scala/akka/stream/javadsl/SubFlow.scala
ktoso: yeah, this is becoming really crazy... :( 
Make it a ticket, if I'm a bit more free I can try out something.

ktoso: https://github.com/akka/akka/issues/19019

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowConcatAllSpec.scala
```diff
@@ -44,7 +44,13 @@ class FlowConcatAllSpec extends AkkaSpec {
 
     "work together with SplitWhen" in {
       val subscriber = TestSubscriber.manualProbe[Int]()
-      Source(1 to 10).splitWhen(_ % 2 == 0).flatMapConcat(ConstantFun.scalaIdentityFunction).runWith(Sink(subscriber))
+      Source(1 to 10)
+        .splitWhen(_ % 2 == 0)
+        .prefixAndTail(0)
+        .map(_._2)
+        .concatSubstreams
+        .flatMapConcat(ConstantFun.scalaIdentityFunction)
```
viktorklang: I find this code confusing. :( (i.e. the double concats)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1245,7 +1294,26 @@ trait FlowOps[+Out, +Mat] {
...
+   *     |  +------+        +------+  |
+   *     |  |      |        |      |  |
+   * In ~~> | flow | ~Out~> | sink |  |
+   *     |  |      |        |      |  |
+   *     |  +------+        +------+  |
+   *     +----------------------------+
+   * }}}
+   * The materialized value of the combined [[Sink]] will be the materialized
+   * value of the current flow (ignoring the given Sink’s value), use
+   * [[Flow#toMat[Mat2* toMat]] if a different strategy is needed.
```
ktoso: whoa, this ref works?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1259,29 +1327,122 @@ trait FlowOps[+Out, +Mat] {
...
+  /**
+   * Attaches the given [[Sink]] to this [[Flow]], meaning that elements that passes
+   * through will also be sent to the [[Sink]].
+   *
+   * @see [[#alsoTo]]
+   */
+  def alsoToMat[Mat2, Mat3](that: Graph[SinkShape[Out], Mat2])(matF: (Mat, Mat2) ⇒ Mat3): ReprMat[Out, Mat3] =
+    viaMat(alsoToGraph(that))(matF)
+
+  private[akka] def transformMaterializing[T, M](mkStageAndMaterialized: () ⇒ (Stage[Out, T], M)): ReprMat[T, M] =
```
ktoso: INTERNAL API?

akka-stream/src/main/scala/akka/stream/scaladsl/SubFlow.scala
```diff
@@ -0,0 +1,40 @@
...
+import akka.stream.impl.Stages.StageModule
+import language.higherKinds
+import scala.annotation.unchecked.uncheckedVariance
+
+/**
+ * A “stream of streams” sub-flow of data elements, e.g. produced by `groupBy`.
+ * SubFlows cannot contribute to the super-flow’s materialized value since they
+ * are materialized later, during the runtime of the flow graph processing.
+ */
+trait SubFlow[+Out, +Mat, +F[+_], C] extends FlowOps[Out, Mat] {
```
ktoso: (<3 scala's code reusability)

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeReduceByKey.scala
rkuhn: WDYM?

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKeyTest.java
rkuhn: But `Pair.create` does not look nicer or save characters, so I prefer the obvious notation.

akka-docs-dev/rst/scala/stream-testkit.rst
rkuhn: good point, please track as a separate ticket

akka-stream/src/main/scala/akka/stream/javadsl/SubFlow.scala
rkuhn: no, the current is inner as well

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowConcatAllSpec.scala
rkuhn: Hmm, I don’t see the problem: there are two layers of substreaming, so there must be two layers of flattening as well. What is confusing?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: no, of course not (yes, it does work for ScalaDoc, but not for JavaDoc)

rkuhn: which is totally okay since they are separate, but thanks for the poking: the Java version had not yet been outfitted with the drawing!

2beaucoup:
> There are a lot of prefixAndTail(1)s here. Maybe introduce headAndTail which returns (Out, Source[U, Unit]) instead of (immutable.Seq[Out], Source[U, Unit]) and which spares users the hassle of sequence matching? ;)

WDYT @rkuhn?

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKeyTest.java
2beaucoup: If it only would have been named `Pair.of`... :(

rkuhn:
`prefixAndTail` is the single most dangerous combinator in the whole library that remains after these changes, so making it any more attractive is not a good idea: if people really need to formulate dynamic stream processing topologies then they’ll need to go through some undocumented boilerplate rituals. We do want to make the alternatives as easily accessible as possible, though, which this PR is one part of.

akka-docs-dev/rst/java/migration-guide-1.0-2.x-java.rst
drewhk: you don't use the RST plugin in IDEA? It shows misaligned header lines.

ktoso: I? Sure I do; so I always fix those. But never realised it's a real rendering problem that I'm fixing ;)

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeReduceByKey.scala
viktorklang: Option.get is always a smell, if you want it to fail on None use getOrElse and throw a more meaningful exception?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowConcatAllSpec.scala
viktorklang: Since `concatSubstreams` and `flatMapConcat(identity)` are isomorphic, it looks weird to have both, and to have them adjacent, don't you agree?

drewhk: But what is your suggestion? They are different operations with different safety and operational requirements.

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeReduceByKey.scala
drewhk: In this case it emulates `reduce` which throws on an empty collection. Here it throws on an empty stream.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowConcatAllSpec.scala
viktorklang: Looking at the code above, how do I as the user or reader of the code by any means elicit these safety and operational requirements?

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeReduceByKey.scala
rkuhn: What you guys are overlooking is that this cannot ever be empty, by construction. `reduce` is more appropriate but does not exist yet.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowConcatAllSpec.scala
drewhk: You ask questions, but don't provide an answer. You kind of suggest that there is, but you don't give it. If you have an answer, then please state it as a constructive recommendation, not an open ended question.

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeReduceByKey.scala
viktorklang: Yes, if you re-read what I wrote, you'll notice that I have nothing against throwing exceptions but what I was against was A) the use of Option.get and B) the generated exception from using Option.get

Option.get should not be encouraged, and in this case the error message ought to be different than: java.util.NoSuchElementException: None.get

drewhk: I know it cannot be empty, I wrote this sample :) I just pointed out that it does exactly what reduce: works only with non-empty collections. I did not imply that the substream can be empty.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowConcatAllSpec.scala
rkuhn: These operations are not isomorphic, they even have different types. And users will not—they shall not—use prefixAndTail unless they know what they are doing.

viktorklang: No, that was an honest question. As the designers of this API, we need to be able to answer it, and right now I can't if someone were to ask me. So I'm asking the author.

drewhk:
```
.prefixAndTail(0)
.map(_._2)
.concatSubstreams
```
is a funky way to way to lift the substreams to actual elements of the stream. It is funky because it is dangerous therefore we don't provide sugar for it. The additional `flatMapConcat` is part of the integration test which is to see if splitWhen and concatAll works together.

viktorklang: @drewhk Thanks! Alright, so if it is code we don't want people to write anyway then it doesn't really count towards it that it looks confusing. Thanks for educating me!

rkuhn: Right, the only reason for this code to exist is that I changed the API without changing what the test does, but the test does no longer use the API in a realistic or idiomatic way. Nevertheless it does test all those important properties.

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeReduceByKey.scala
rkuhn: @viktorklang we’re arguing over the empty set, especially in view of #19020.

viktorklang: @rkuhn :+1:

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeReduceByKeyTest.java
hepin1989: new api in jdk 9 will make use of the `of`
http://openjdk.java.net/jeps/269

ktoso: I am a big fan of `of`. I think we should add it (once streams are on master, that'll be much simpler)

hepin1989: And more `of` here https://github.com/jOOQ/jOOL

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
ktoso: great, the drawings are very helpful :+1:

ktoso:
Needs a rebase and validation failed on compiling java tests:

```
[error] /localhome/jenkinsakka/workspace/pr-validator-per-commit-jenkins/akka-stream-tests/src/test/java/akka/stream/javadsl/SourceTest.java:67: error: method mapConcat in class Source<Out,Mat> cannot be applied to given types;

```

rkuhn:
Yes, I stashed that and could potentially finish it now that FlattenMerge is there, but I’m fusing all the things right now :-)

ktoso:
Roger that, thanks for the update (going through all PRs today)

PR #19008, opened 2015-11-25
per #18839 Add better journal proxy

ghost:
Replaces (and includes) #18949 and #18876

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/journal/PersistencePluginProxy.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/journal/PersistencePluginProxy.scala
@@ -3,34 +3,17 @@
  */
 package akka.persistence.journal
 
-import akka.util.Helpers.Requiring
-import scala.concurrent.duration._
-import akka.actor.Actor
-import akka.actor.Stash
-import scala.concurrent.duration.FiniteDuration
-import akka.actor.ActorRef
-import akka.persistence.JournalProtocol
-import akka.actor.ActorSystem
-import akka.persistence.Persistence
-import scala.util.control.NoStackTrace
+import java.net.URISyntaxException
 import java.util.concurrent.TimeoutException
-import akka.persistence.AtomicWrite
-import akka.persistence.NonPersistentRepr
-import akka.persistence.DeleteMessagesFailure
-import akka.actor.ActorLogging
-import com.typesafe.config.Config
-import akka.actor.Address
-import akka.actor.ActorIdentity
-import akka.actor.RootActorPath
-import akka.actor.Identify
-import akka.actor.ReceiveTimeout
-import akka.actor.ExtendedActorSystem
-import akka.persistence.SaveSnapshotFailure
-import akka.persistence.DeleteSnapshotFailure
-import akka.persistence.DeleteSnapshotsFailure
-import akka.persistence.SnapshotProtocol
 
-object JournalProxy {
+import akka.actor._
+import akka.persistence.{ AtomicWrite, DeleteMessagesFailure, DeleteSnapshotFailure, DeleteSnapshotsFailure, JournalProtocol, NonPersistentRepr, Persistence, SaveSnapshotFailure, SnapshotProtocol }
+import akka.util.Helpers.Requiring
+import com.typesafe.config.Config
+
+import scala.concurrent.duration._
+
+object PersistencePluginProxy {
   final case class TargetLocation(address: Address)
   private case object InitTimeout
 
@@ -38,6 +21,13 @@ object JournalProxy {
     Persistence(system).journalFor(null) ! TargetLocation(address)
     if (system.settings.config.getString("akka.persistence.snapshot-store.plugin") != "")
       Persistence(system).snapshotStoreFor(null) ! TargetLocation(address)
+  }
+
+  def start(system: ActorSystem): Unit = {
+    Persistence(system).journalFor(null)
+    if (system.settings.config.getString("akka.persistence.snapshot-store.plugin") != "")
+      Persistence(system).snapshotStoreFor(null)
+
   }
 
   private sealed trait PluginType {
@@ -51,9 +41,23 @@ object JournalProxy {
   }
 }
 
-// FIXME document me
-final class JournalProxy(config: Config) extends Actor with Stash with ActorLogging {
-  import JournalProxy._
+/**
+ * PersistencePluginProxyExtensionImpl is an `Extension` that enables initialization of the `PersistencePluginProxy`
+ * via configuration, without requiring any code changes or the creation of any actors.
+ * @param system The actor system to initialize the extension for
+ */
+class PersistencePluginProxyExtensionImpl(system: ActorSystem) extends Extension {
+  PersistencePluginProxy.start(system)
+}
+
+object PersistencePluginProxyExtension extends ExtensionId[PersistencePluginProxyExtensionImpl] with ExtensionIdProvider {
+  override def createExtension(system: ExtendedActorSystem): PersistencePluginProxyExtensionImpl = new PersistencePluginProxyExtensionImpl(system)
+  override def lookup(): ExtensionId[_ <: Extension] = PersistencePluginProxyExtension
+  override def get(system: ActorSystem): PersistencePluginProxyExtensionImpl = super.get(system)
+}
+
+final class PersistencePluginProxy(config: Config) extends Actor with Stash with ActorLogging {
+  import PersistencePluginProxy._
   import JournalProtocol._
   import SnapshotProtocol._
 
@@ -65,7 +69,7 @@ final class JournalProxy(config: Config) extends Actor with Stash with ActorLogg
       throw new IllegalArgumentException("Unknown plugin type: " + other)
   }
 
-  private val timeout: FiniteDuration = config.getDuration("init-timeout", MILLISECONDS).millis
+  private val initTimeout: FiniteDuration = config.getDuration("init-timeout", MILLISECONDS).millis
   private val targetPluginId: String = {
     val key = s"target-${pluginType.qualifier}-plugin"
     config.getString(key).requiring(_ != "", s"$pluginId.$key must be defined")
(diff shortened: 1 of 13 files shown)


patriknw:
OK TO TEST

patriknw:
@talp-traiana what is the status here, did you find a solution for the ask+sender problem?

talp-traiana:
I'm doing it right now, using the PR you suggested. For now it's the same content as the old PR, just rebased.

patriknw:
great, ping me when it's ready to be reviewed

talp-traiana:
@patriknw 
Latest commit should take care of all the comments from the previous PR (including the ask changes)

patriknw:
LGTM, awesome contribution, @talp-traiana 
Thanks!

talp-traiana:
Great!
I'll start working on some tests. Is there anything else to do here?

patriknw:
some tests would be great, I think that is the remaining part

talp-traiana:
@patriknw 
I came across another issue during testing, although I'm not sure if it's really a problem and if there's any way to fix it.
As far as I can tell, the persistence plugin (the proxy, in this case) does not get created until a persistent actor is created. This means that the "main" node (the one holding the target) *has* to create a persistent actor, or the actor selection will not return any results.
So the question(s) is:
1. Is there a way to force the creation of the plugin (via configuration)?
2. If there is a way, do we want to do it in all the actor systems or just the "main" one? On the one hand, there's no reason to do it in all of them. On the other hand, we don't want Akka to check the plugin's "start-target-journal" flag...
3. If there isn't a way, then is there a better way than creating a persistent actor? Some sort of programmatic way to manually instantiate the plugin... In any case, whether it's creating an actor or some other way, this should be added to the docs.

patriknw:
I see what you mean. You want `Persistence(system).journalFor(null)` to be called on system startup. 
One thought I had was to do that in the `Persistence` extension but we already have some things marked lazy with comment "Lazy, so user is not forced to configure defaults when she is not using them.", so that is a bad idea.

I see no other way than that you add a `def start(): Unit` method in `object PersistencePluginProxy`, which must be called by user code (on all nodes should be alright).

talp-traiana:
So I've run into sort of a wall with ask - it only returns the first response, and that's not enough for WriteMessages or ReplayMessages.
The solutions I can think of right now are:
1. Manage the state for each request in the proxy (and keep track of timeouts, etc.), or
2. Create some sort of pipe that's similar to ask but can handle multiple messages. This raises the issue of when to shut down the spawned actor - I'm thinking of maybe a predicate that would tell the actor when to stop itself.
Both solutions look like they could get a little hairy, but I think out of the 2 I prefer the second one.

What do you think? Do you have any other ideas?
In any case, I'll probably start messing around with it tomorrow and see where it goes...

akka-persistence/src/main/scala/akka/persistence/journal/PersistencePluginProxy.scala
```diff
@@ -0,0 +1,223 @@
...
+      val f = askExt(targetJournal, { askSender: ActorRef ⇒ msg.copy(persistentActor = askSender) })(msgTimeout)
+      f onComplete {
+        case Success(v) ⇒ replyTo ! v
+        case Failure(t) ⇒
+          replyTo ! WriteMessagesFailed(t)
+          messages foreach {
+            case a: AtomicWrite       ⇒ a.payload.foreach { p ⇒ replyTo ! WriteMessageFailure(p, t, actorInstanceId) }
+            case r: NonPersistentRepr ⇒ replyTo ! LoopMessageSuccess(r.payload, actorInstanceId)
+          }
+      }
```
patriknw: @talp-traiana is it here you have problems with ask and multiple messages. I don't fully understand the problem. Let's discuss tomorrow.

talp-traiana:
@patriknw 
It's part of the issue.

The main problem is that ask is intended for a "protocol" where a request has 1 response. So ask spawns an actor, waits for the response, forwards it, and kills the actor. However, in the journal protocol, some requests result in several responses, so the ask pattern breaks down.
In the case of a timeout, we send the appropriate errors messages to the persistent actor.

So in our case, consider an actor trying to recover - after sending a ReplayMessages to the journal, it responds with all the events in the journal followed by a RecoveryComplete. However, only the first message arrives at the persistent actor and the rest end up as dead letters.

patriknw:
@talp-traiana I see what you mean. Ask cannot work for this. I'm so sorry that I didn't realized this from the beginning.

Let's take a step back. The purpose of this tool is for testing and it would have been nice to not be silent in case of connectivity problems, but it's not intended for production. Therefore I don't think we should implement anything complicated here.

We have another ticket for making [PersistentActor resilient against missing messages from the Journal](https://github.com/akka/akka/issues/15303). That would be the right place to implement it, and it would work for the plugin proxy also. That is not completely trivial to implement without sacrificing performance, so it's not something we should do hastliy.

For the plugin proxy I suggest that you add `watch` and if the journal terminates `become initTimedOut`. Then it will at least reply with failures for new requests.

In `active` you can keep the ask for the single reply requests, but the others should just be forwarded.

That is good enough.

talp-traiana:
OK, sounds good. Now the only question is whether it's worth it to keep the ask in single-reply cases, or forward those as well.
Some arguments for ditching ask altogether:
- Makes the proxy much simpler, without any special cases.
- Makes the proxy resilient to protocol changes.
- If and when the PA resilience issue is resolved it would make the ask pointless.

patriknw:
I don't mind ditching ask altogether.

talp-traiana:
@patriknw 
OK, so I think this should cover everything we talked about (watching the target, forwarding messages, initialization, documentation, tests)
Let me know if there's anything else...

akka-persistence/src/main/scala/akka/persistence/journal/PersistencePluginProxy.scala
```diff
@@ -0,0 +1,186 @@
...
+  final case class TargetLocation(address: Address)
+  private case object InitTimeout
+
+  def setTargetLocation(system: ActorSystem, address: Address): Unit = {
+    Persistence(system).journalFor(null) ! TargetLocation(address)
+    if (system.settings.config.getString("akka.persistence.snapshot-store.plugin") != "")
+      Persistence(system).snapshotStoreFor(null) ! TargetLocation(address)
+  }
+
+  def start(system: ActorSystem): Unit = {
```
patriknw: One thought. We could make this an Akka [Extension](http://doc.akka.io/docs/akka/2.4.0/scala/extending-akka.html) instead. Then those methods would not need to have a `ActorSystem` parameter and it is possible to configure an extension to be loaded at system startup, as alternative to calling this start() method.

In the extension we would do `Persistence(system).journalFor(null)` in the constructor of the extension.

WDYT?

akka-persistence/src/main/scala/akka/persistence/journal/PersistencePluginProxy.scala
```diff
@@ -0,0 +1,186 @@
...
+      context.become(active(target, address == selfAddress))
+    case _: ActorIdentity ⇒ // will retry after ReceiveTimeout
+    case ReceiveTimeout ⇒
+      sendIdentify(address)
+  }: Receive).orElse(init)
+
+  def active(targetJournal: ActorRef, targetAtThisNode: Boolean): Receive = {
+    case TargetLocation(address) ⇒
+      if (targetAtThisNode && address != selfAddress)
+        becomeIdentifying(address)
```
patriknw: to be on the safe side I think you need to handle `Terminated` also in the other Receive states, in case we don't get a Terminated from an "old" journal. In the other places it can just be ignored (but it must be handled).

akka-persistence/src/main/scala/akka/persistence/journal/PersistencePluginProxy.scala
talp-traiana: I don't really know much about extensions, but it sounds good. I'll read up on it, and give it a shot.
It won't be till next week, though.

akka-persistence/src/main/scala/akka/persistence/journal/PersistencePluginProxy.scala
talp-traiana: Will do.

talp-traiana:
@patriknw 
I added the extension (and the test uses it), and handled the Terminated messages in all states.
Anything else?

patriknw:
The build failure is the binary compatibility check, please add the following to [Mima.scala](https://github.com/akka/akka/blob/master/project/MiMa.scala#L583)

    FilterAnyProblem("akka.persistence.journal.inmem.InmemJournal")

It's alright to ignore these incompatibilities because `InmemJournal` is INTERNAL API (internal testing factility).

You can try that it works by using `sbt mimaReportBinaryIssues`

patriknw:
After fixing the mima issue this is ready to be merged. Please squash it to one commit (or two if you want to keep my initial commit).

talp-traiana:
@patriknw 
I've had to also ignore problems with InmemStore as well as InmemJournal. That's OK, right?

patriknw:
yes, that's fine. Thanks

talp-traiana:
@patriknw 
OK, squashed and rebased on master.

talp-traiana:
@patriknw 
Just wanted to know what the status on this is - is it waiting for anything?

patriknw:
Sorry for the delay. Merging it now. Thanks a lot for an excellent contribution. You are always welcome to contribute with more things.

patriknw:
Refs #18839

talp-traiana:
Thanks!
I'll have a look around and see if there's anything else I feel I'm up to (suggestions would be welcome ^_^ )

patriknw:
Thanks

Persistence related issues are tagged with `t:persistence`: https://github.com/akka/akka/issues?q=is%3Aopen+is%3Aissue+label%3At%3Apersistence+milestone%3A2.4.x

`community-contrib` is another tag that might contain something: https://github.com/akka/akka/issues?utf8=✓&q=is%3Aopen+is%3Aissue+label%3Acommunity-contrib

Issue #18978, opened 2015-11-20
After sharding handoff the shard allocation should be lazy


patriknw:
I'm pretty sure it was lazy (first message) originally, but when checking the ShardCoordinator it looks like it's not. Perhaps it was changed due to the remember-entities feature. I think it should be lazy when remember-entities is not used.

Also verify that we don't grow the State when shards are deallocated (`unallocatedShards`).

rkuhn:
This should be closed, right?

ktoso:
No, it's missing a full blown integration test that will be added next week.

rkuhn:
@ktoso please create a new ticket for the integration test

ktoso:
Okey, here it is: https://github.com/akka/akka/issues/19033

patriknw:
No, you are mixing it up with the other issue about BackoffSupervisor

ktoso:
Ouch, yes indeed; sorry about that

PR #19010, opened 2015-11-25
=cls #18978 Lazy startup of shards when rememberEntities=false

patriknw:
- and don't populate the unallocatedShards Set in the State
  when rememberEntities=false

### DIFF
--- a/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
+++ b/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
@@ -257,7 +257,15 @@ object ShardCoordinator {
       // shards for each region
       regions: Map[ActorRef, Vector[ShardId]] = Map.empty,
       regionProxies: Set[ActorRef] = Set.empty,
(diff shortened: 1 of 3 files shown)


patriknw:
Refs #18978

patriknw:
Please review so that this can be included in 2.4.1 today.

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -809,7 +828,8 @@ class DDataShardCoordinator(typeName: String, settings: ClusterShardingSettings,
   // This state will drop all other messages since they will be retried
   def waitingForState: Receive = ({
     case g @ GetSuccess(CoordinatorStateKey, _) ⇒
-      state = g.get(CoordinatorStateKey).value
+      println(s"# GetSuccess $self ${settings.rememberEntities}") // FIXME
```
ktoso: remove println please

patriknw: oh, thanks

ktoso:
LGTM, let's remove the println and merge

drewhk:
LGTM

Issue #18639, opened 2015-10-04
JournalSpec reject non-serializable events


t3hnar:
It is not possible to satisfy `JournalSpec reject non-serializable events` eventstore journal, as it expects events to have incremented sequence number.

So basically the case with `6L` missed and then `7L` is written is not valid for this particular journal.
I'm still wondering whether it is expected to be a valid case for other journals?

```scala
      probe.expectMsgPF() {
        case WriteMessageRejected(PersistentImpl(payload, 7L, Pid, _, _, Actor.noSender, WriterUuid), _, _) ⇒
          payload should be(notSerializableEvent)
      }
      probe.expectMsgPF() {
        case WriteMessageSuccess(PersistentImpl(payload, 8L, Pid, _, _, Actor.noSender, WriterUuid), _) ⇒ payload should be(s"b-8")
      }
```

patriknw:
There will be holes in the sequence numbers when events are rejected. Akka Persistence itself has no problems with holes. However, it should not be mandatory for a plugin to implement rejections, it should be allowed to fail in stead. Serialization is very specific business of the journal and we should not mandate anything around that.

We should relax this part of the TCK. Thanks for reporting.

patriknw:
@ktoso Would you be able to fix this TCK adjustment, so that it's included in 2.4.1?

ktoso:
I'll take care of it

ktoso:
It is now possible to skip these tests.

PR #19011, opened 2015-11-25
=per,tck #18639 makes testing for rejections optional

ktoso:
This solves https://github.com/akka/akka/issues/18639 and allows us to support "OPTIONAL" tests a bit nicer.


ktoso:
Output looks like this:

```
[info] A Journal
[info] - may reject non-serializable events !!! CANCELED !!! (20 milliseconds)
[info]   This journal does not reject non-serializable events. This is legal. (MatchersHelper.scala:160)
[info]   akka.persistence.scalatest.MayVerb$TestCanceledByFailure:
[info]   at akka.persistence.journal.JournalSpec$$anonfun$2$$anonfun$apply$mcV$sp$15$$anonfun$apply$mcV$sp$16.apply$mcV$sp(JournalSpec.scala:190)
```

Sadly not showing any stacktrace is not possible using ScalaTest infra currently. It has to be at least one line.

akka-persistence-tck/src/main/scala/akka/persistence/journal/JournalSpec.scala
```diff
@@ -181,8 +183,11 @@ abstract class JournalSpec(config: Config) extends PluginSpec(config) {
       journal ! ReplayMessages(0, Long.MaxValue, Long.MaxValue, pid, receiverProbe.ref)
       receiverProbe.expectMsg(RecoverySuccess(highestSequenceNr = 5L))
     }
+  }
 
-    "reject non-serializable events" in {
+  "A Journal" may {
+    "reject non-serializable events" in optional("This journal does not reject non-serializable events. This is legal.") {
+      1 should ===(2)
```
patriknw: `1 should ===(2)`?

akka-persistence-tck/src/main/scala/akka/persistence/journal/JournalSpec.scala
```diff
@@ -181,8 +183,11 @@ abstract class JournalSpec(config: Config) extends PluginSpec(config) {
       journal ! ReplayMessages(0, Long.MaxValue, Long.MaxValue, pid, receiverProbe.ref)
       receiverProbe.expectMsg(RecoverySuccess(highestSequenceNr = 5L))
     }
+  }
 
-    "reject non-serializable events" in {
+  "A Journal" may {
+    "reject non-serializable events" in optional("This journal does not reject non-serializable events. This is legal.") {
```
patriknw: how does this work? how can a journal implementation define that it supports this feature or not?

ktoso: if it passes the test, it's green. If it doesn't, the test "skipped" (canceled in scalatest terminology), see how that looks like in the comment I made on the PR

akka-persistence-tck/src/main/scala/akka/persistence/journal/JournalSpec.scala
ktoso: Ah yeah, will remove ofc; this was used to generate: https://github.com/akka/akka/pull/19011#issuecomment-159589805

akka-persistence-tck/src/main/scala/akka/persistence/journal/JournalSpec.scala
ktoso: If we want to go down the route of enabling features to be tested it could be made a bit simpler; Do we?

patriknw: Then I don't think the test is worth much, since the plugin maintainer will not see test failures no matter how things are implemented (or broken by regression), apart from parsing logs.

I think it's better to have a method that the plugin can override to say that it doesn't support rejections. I think we already have something like that for atomic `persistAll`.

ktoso: Ok, so overrideable "feature flags" :) will do.

patriknw: thanks

ktoso:
Capability Flags now implemented.

Looks like this:

```scala
  override def supportsRejectingNonSerializableObjects = true
```

```
[info] A Journal optionally
[info] + CapabilityFlag `supportsRejectingNonSerializableObjects` was turned `on`.
[info] - may reject non-serializable events (17 milliseconds)
```

```
[info] A Journal optionally
[info] + CapabilityFlag `supportsRejectingNonSerializableObjects` was turned `off`. To enable the related tests override it with `CapabilityFlag.on` (or `true` in Scala).
```

ktoso:
Please review @patriknw

.java-version
```diff
@@ -1 +1 @@
-1.6
+1.8
```
ktoso: (I actually use this file - for jenv ( http://www.jenv.be/ ), but if people mind I'm ok to remove it from the repo (and add to my local gitignore)

patriknw:
LGTM, thanks

ktoso:
Added docs and akka-docs should now pass as well.

Issue #19016, opened 2015-11-25
BackoffSupervisor can't be used with ClusterSharding passivation


masterzen:
Despite 3f3642337eeaf90c66e550ff6848723ec3527145 that implements forwarding back messages from the child to the parent of the `BackoffSupervisor` (which is useful for passivation), this can't work because:
1. The actor (protected by a `BackoffSupervisor`) needs to stop
2. it sends a `Passivate` message to its parent `BackoffSupervisor`
3. the parent `BackoffSupervisor` forwards it to the `Shard`
4. the `Shard` sees the `sender()` as the real actor  and not the `BackoffSupervisor` and finally executes calls [passivate](https://github.com/akka/akka/blob/master/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/Shard.scala#L155)
5. which in turn ends in this code with `entity` as the original passivation starter actor (the child of the `BackoffSupervisor`):

``` scala
  def passivate(entity: ActorRef, stopMessage: Any): Unit = {
    idByRef.get(entity) match {
      case Some(id) if !messageBuffers.contains(id) ⇒
        log.debug("Passivating started on entity {}", id)

        passivating = passivating + entity
        messageBuffers = messageBuffers.updated(id, Vector.empty)
        entity ! stopMessage

      case _ ⇒ //ignored
    }
  }
```

The forwarded `sender()` can't be in the `idByRef` Map, since it's not the direct child of the `Shard` but the child of the `BackoffSupervisor`, so passivation is not started.

A related issue, if this aforementioned one is solved is that the real actor receiving the passivation return message will have to stop itself, but also to stop its ancestor (the `BackoffSupervisor`) otherwise this one will restart it.

In a word, passivation needs to be built-in in the `BackoffSupervisor` (or in a Cluster Sharding specific supervisor system), for it to work correctly.

ktoso:
(note to self - keeping open until test added)

nvollmar:
strange to see this in the release notes when the pull request has been merged 3 years ago already?

chbatey:
The PR was #25933

nvollmar:
Right, thanks for the hint. Did I understand correctly that we now should set `.withFinalStopMessage(_ == StopMessage)` to get passivation working properly?

nvollmar:
Right, thanks for the hint. 

[...]

PR #19027, opened 2015-11-27
=act #19016 use the BackoffSupervisor as sender for parent msg

patriknw:
Intermediate fix for using BackoffSupervisor with sharding passivation. I'm short of time and therefore no tests. WIll not close the ticket until tests are added or another solution implemented.

Nevertheless this is more correct than the previous forward, since the child ref should not really be exposed to the outside when using BackoffSupervisor.

As far as I can see this should at least work with Cluster Sharding passivation when using PoisonPill as stop message.

### DIFF
--- a/akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
+++ b/akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
@@ -148,8 +148,8 @@ object BackoffSupervisor {
  * message to this actor and it will reply with [[akka.pattern.BackoffSupervisor.CurrentChild]]
  * containing the `ActorRef` of the current child, if any.
  *
- * The `BackoffSupervisor` forwards all messages from the child to the parent of the
- * `BackoffSupervisor`.
+ * The `BackoffSupervisor`delegates all messages from the child to the parent of the
+ * `BackoffSupervisor`, with the supervisor as sender.
  *
  * The `BackoffSupervisor` forwards all other messages to the child, if it is currently running.
  *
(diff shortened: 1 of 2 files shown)


patriknw:
Refs #19016

johanandren:
LGTM (except for missing tests)

ktoso:
LGTM, I agree adding a test would be nice.
We can add it separately if adding one would make it slip (please add a ticket then).

[Merge + ticket] or [add test + merge] :-)

ktoso:
Duh, I should read PR descriptions :wink: 
LGTM

patriknw:
I will not be able to test things today, but since this parent forwarding didn't exist in 2.4.0, I think this is better than just leaving the forward in. Sorry for not being able to solve the full thing.

ktoso:
Yup, not a problem; let's keep the ticket open and follow up next week.

drewhk:
LGTM

ahjohannessen:
This is not perfect for all scenarios, e.g. if the "real" parent is stashing while shutting the "real" child down due to a "passivate" message, then the backoff supervisor will get a poison pill (if that is used) and possibly not allowing the child to process its mailbox completely. If a custom stop message is used then it is not intuitive that the child needs to kill its parent (backoff supervisor) in order to be stopped itself.

patriknw:
yes, this was an intermediate fix, more thoughts should be put into this. The ticket is still open for that reason, please comment there: https://github.com/akka/akka/issues/19016
Ideas of better solution or PR of it is welcome.

Issue #18807, opened 2015-10-29
Add range Source for Java DSL


drewhk:
Since Java has no range unlike scala: `Source(1 to N)` or `Source(() => Iterator from 0)`

viktorklang:
I liked this implementation: http://stackoverflow.com/a/6828887/227803

agolubev:
Ready for review

drewhk:
Fixed by @agolubev

PR #19038, opened 2015-11-28
+str #18807 add Source.range for Java DSL

agolubev:


ktoso:
OK TO TEST

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -111,6 +112,18 @@ object Source {
...
 
   /**
+   * Creates [[Source]] with `start` as the first element and each next element as `previous + 1` until
+   * it reaches `end`. It allows to create `Source` out of range as simply as on scala `Source(1 to N)`
+   */
+  def range(start: Int, end: Int): javadsl.Source[Integer, Unit] = {
+    require(start <= end, "start must be less or equal than end")
+    from(new util.AbstractList[Integer]() {
+      override def get(index: Int) = start + index
+      override def size() = end - start + 1
```
viktorklang: Perhaps add:

`override def toString: String = s"$start to $end"`

ktoso: Yes, good idea, but via `.named(s"Range($start to $end)")` perhaps instead?

viktorklang: To be symmetric with something like Source.apply(() => Iterator), it should be a toString override.

ktoso: Random thought... I seem to remember Scala's Range was "pretty heavy" though can't recall completely in what way...? So since `javadsl.Source.range(1, 10)` wants to be the equivalent of `scaladsl.Source(1 to 10)`, could we dispel my worry that if we don't provide a `range` in Scala it would perform worse? (I.e. do we not want to provide `range` for scala as well?)

viktorklang: I felt the doc also needs to be more explicit if 'end' is inclusive or
exclusive.

-- 
Cheers,
√

viktorklang: I was just going to propose to use s.c.i.Iterable.range but then I saw in
horror what it produces.

On Sun, Nov 29, 2015 at 12:44 AM, Konrad Malawski <notifications@github.com>
wrote:

> In akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
> <https://github.com/akka/akka/pull/19038#discussion_r46085788>:
>
> > @@ -111,6 +112,18 @@ object Source {
> >    }
> >
> >    /**
> > +   * Creates [[Source]] with `start` as the first element and each next element as `previous + 1` until
> > +   * it reaches `end`. It allows to create `Source` out of range as simply as on scala `Source(1 to N)`
> > +   */
> > +  def range(start: Int, end: Int): javadsl.Source[Integer, Unit] = {
> > +    require(start <= end, "start must be less or equal than end")
> > +    from(new util.AbstractList[Integer]() {
> > +      override def get(index: Int) = start + index
> > +      override def size() = end - start + 1
>
> Random thought... I seem to remember Scala's Range was "pretty heavy"
> though can't recall completely in what way...? So since javadsl.Source.range(1,
> 10) wants to be the equivalent of scaladsl.Source(1 to 10), could we
> dispel my worry that if we don't provide a range in Scala it would
> perform worse? (I.e. do we not want to provide range for scala as well?)
>
> —
> <https://github.com/akka/akka/pull/19038/files#r46085788>.
>



-- 
Cheers,
√

ktoso: Yup. Thus, I'd argue we need range for scala-dsl too.

viktorklang: s.c.i.Range is not bad tho?

ktoso: Oh, so perhaps I mixed up those two... Will need to double check

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -111,6 +112,18 @@ object Source {
   }
 
   /**
+   * Creates [[Source]] with `start` as the first element and each next element as `previous + 1` until
+   * it reaches `end`. It allows to create `Source` out of range as simply as on scala `Source(1 to N)`
+   */
+  def range(start: Int, end: Int): javadsl.Source[Integer, Unit] = {
```
drewhk: maybe a step parameter?

viktorklang: +1 on `step` overload

drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
viktorklang: s.c.i.RAnge works really well. Please update to delegate to the scaladsl's immutable.Iterable-apply and use s.c.i.Range

agolubev:
String rendering test case "should not open a second connection if there is an idle one available" failed.
Don't think it relates to Source.range.
@drewhk PR is ready for review

ktoso:
What about the `toString` or `named` here - as discused in https://github.com/akka/akka/pull/19038#discussion_r46084517

agolubev:
Yeah, will add this as default prints all the range:
```
scala> Range.inclusive(1,20,1)
res0: scala.collection.immutable.Range.Inclusive = Range(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20)

scala> res0.toString
res1: String = Range(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20)
```

johanandren:
PLS BUILD

agolubev:
hm build failed with strange exception VirtualProcessorTest:
```
required_spec309_requestZeroMustSignalIllegalArgumentException *** FAILED ***
[info]   java.lang.AssertionError: Expected onError(java.lang.IllegalArgumentException) within 500 ms
```
perhaps timing issue

rkuhn:
LGTM thanks!

PR #19052, opened 2015-11-30
=htc restructure HttpServerBlueprint as stack of BidiFlows for better readability

jrudolph:
The main benefit is the split up of the big graph building code block into easier to digest pieces.

/cc @sirthias

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -7,7 +7,6 @@ package akka.http.impl.engine.server
 import java.net.InetSocketAddress
 import java.util.Random
 import scala.collection.immutable
-import akka.stream.scaladsl.One2OneBidiFlow.One2OneBidi
 import org.reactivestreams.{ Publisher, Subscriber }
 import scala.util.control.NonFatal
 import akka.event.LoggingAdapter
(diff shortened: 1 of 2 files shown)


ktoso:
LGTM, I like it!

jrudolph:
I added a commit to convert the Websocket graph in HttpServerBlueprint by a single simpler GraphStage. 

There is now one test failing I didn't manage to fix, though. It's in TlsEndpointVerificationSpec. In this scenario SSL fails with a certificate error and we want to propagate the exception to the HTTP user handler. This now doesn't work any more because there's the usual race condition between cancelling one side and reporting the error on the other side.

It seems with the old solution something prevented that cancelling the network side would propagate to HTTP. I guess the reason was that in the old solution I had an extra broadcast (the switchTokenBroadcast) where only one leg was cancelled.

In the new GraphStage, I need to prevent that cancelling `toNet` propagates to `fromHttp`. How can I do this?

jrudolph:
Aha, I figured that it wasn't the cancellation which was propagating but it was the error that triggered a shutdown of the stage and so triggered the cancellation of the other end. I fixed that.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -49,39 +47,45 @@ import akka.util.ByteString
...
-    // will not be mutated, all "shared copy" parsers copy on first-write into the header cache
-    val rootParser = new HttpRequestParser(parserSettings, rawRequestUriHeader,
-      HttpHeaderParser(parserSettings) { info ⇒
-        if (parserSettings.illegalHeaderWarnings)
-          logParsingError(info withSummaryPrepended "Illegal request header", log, parserSettings.errorLoggingVerbosity)
-      })
+  def apply(settings: ServerSettings, remoteAddress: Option[InetSocketAddress], log: LoggingAdapter)(implicit mat: Materializer): Http.ServerLayer =
+    stack(settings, remoteAddress, log)
+
+  def stack(settings: ServerSettings, remoteAddress: Option[InetSocketAddress], log: LoggingAdapter)(implicit mat: Materializer): Http.ServerLayer = {
```
sirthias: Nice, this looks great!

jrudolph:
@ktoso yes, exactly, the stack just represents what the diagram shows horizontally.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -49,39 +47,45 @@ import akka.util.ByteString
...
+    One2OneBidiFlow[HttpRequest, HttpResponse](settings.pipeliningLimit).reversed atop
+      requestPreparation(settings, remoteAddress) atop
+      controller(settings, log) atop
+      parsingRendering(settings, log) atop
+      websocketSupport(settings, log) atop
+      unwrapTls
+  }
+  def unwrapTls: BidiFlow[ByteString, SslTlsOutbound, SslTlsInbound, ByteString, Unit] =
```
rkuhn: this should be a `val`

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -49,39 +47,45 @@ import akka.util.ByteString
...
-    val rootParser = new HttpRequestParser(parserSettings, rawRequestUriHeader,
-      HttpHeaderParser(parserSettings) { info ⇒
-        if (parserSettings.illegalHeaderWarnings)
-          logParsingError(info withSummaryPrepended "Illegal request header", log, parserSettings.errorLoggingVerbosity)
-      })
+  def apply(settings: ServerSettings, remoteAddress: Option[InetSocketAddress], log: LoggingAdapter)(implicit mat: Materializer): Http.ServerLayer =
+    stack(settings, remoteAddress, log)
+
+  def stack(settings: ServerSettings, remoteAddress: Option[InetSocketAddress], log: LoggingAdapter)(implicit mat: Materializer): Http.ServerLayer = {
+    One2OneBidiFlow[HttpRequest, HttpResponse](settings.pipeliningLimit).reversed atop
```
drewhk: This needs a better name... It is a limiter basically, but currently this is not clear from its name.

rkuhn:
LGTM

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
ktoso: I'll fix that in my PR, would like to merge this right now since no other comments on the PR (to spare me basing a PR on a PR etc)

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
jrudolph: How about `userHandlerGuard`?

drewhk: No because now it lives in akka.stream. It is more general than that.

jrudolph: Yes, I agree it makes sense to rename it in both contexts :)

PR #19053, opened 2015-11-30
 Extract resolving remote ShardRegionActor address to a function

kpbochenek:
I wanted to query all actors under given shardTypeName because this should be possible with akka 2.4.1
I know this can be terrible idea when we have a lot of actors but assuming there won't be many of them this should be fine.
Querying ShardRegion only gives us Set[Address] but I was confused how to convert these Addresses to proper ShardRegion actorRefs.
Luckily I found in a test a way to construct ActorPath and then rest was easy.
I think this could be worth extracting.

I though that maybe this could be good addition as a snippet to get all ShardActor Ids but I had no clue where to put it.

```
      import akka.pattern.ask
      import akka.cluster.sharding.ShardRegion.GetCurrentRegions
      import akka.cluster.sharding.ShardRegion.CurrentRegions
      import akka.cluster.sharding.ShardRegion.GetShardRegionState
      import akka.cluster.sharding.ShardRegion.CurrentShardRegionState

      val shardTypeName: String = "???"
      val currentRegions = (ClusterSharding(system).shardRegion(shardTypeName) ? GetCurrentRegions).mapTo[CurrentRegions].map(_.regions)

      def addressToFutureShardIds(region: Address): Future[Set[String]] = {
        val shardState = (ClusterSharding(system).remoteShardRegion(region, shardTypeName) ? GetShardRegionState).mapTo[CurrentShardRegionState].map(_.shards)
        shardState.map(_.flatMap(_.entityIds))
      }

      def toShardIds(regions: Set[Address]): Future[Set[String]] = Future.sequence(regions.map(addressToFutureShardIds)).map(_.flatten)

      val futureIds: Future[Set[String]] = currentRegions.flatMap(toShardIds)
      futureIds.onSuccess { case msg => println(s"Shard[$shardTypeName] contains IDS: $msg") }
```

What do you think?
(damn it picked up my previous PR I will try to sort this out)


akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ClusterSharding.scala
```diff
@@ -7,17 +7,7 @@ import java.net.URLEncoder
...
-import akka.actor.ActorSystem
-import akka.actor.Deploy
-import akka.actor.ExtendedActorSystem
-import akka.actor.Extension
-import akka.actor.ExtensionId
-import akka.actor.ExtensionIdProvider
-import akka.actor.NoSerializationVerificationNeeded
-import akka.actor.PoisonPill
-import akka.actor.Props
+import akka.actor._
```
hepin1989: may this one is not good:)

kpbochenek:
@hepin1989 Intellij import optimization :P Fixed.

Also added snippet mentioned above into docs/pattern section but not sure if that is the best place.

ktoso:
OK TO TEST

johanandren:
This kind of functionality was intentionally left out of the sharding stats changes in 2.4.1, I'm very unsure this is a good idea. WDYT @patriknw ?

patriknw:
Yes, this is exactly what we don't want to provide as a convenience that will be abused (transferring lots of data over the wire). Instead you should start dedicated actors on each node that interact with the local region.

ktoso:
I also vote against exposing this from sharding directly: :-1: 
Reasons mentioned both by @patriknw and @kpbochenek yourself – it's not a good idea in general anyway. If you need it, I'd rather recommend having a dedicated master or region-master actors which maintain such lists as their persistent state.

ktoso:
Would you mind if I close this PR @kpbochenek?
Instead perhaps we could document what to do when you have such need  – would you be able to help us out with adding such section to the docs?

kpbochenek:
Got your point, close if you want.

ktoso:
Closed, thanks a lot for contributing and the discussion!

PR #19055, opened 2015-11-30
=docs style currently deployed

ktoso:
Slight button improvements. Also not `%` based size.

red:

blue:

hover:


ktoso:
Merging as it's the status quo.

PR #19057, opened 2015-12-01
=htc slight renaming, refs #19052

jrudolph:
Refactorings that didn't make it into #19052 any more.


ktoso:
Thanks a lot for following up quickly on that! :-)

ktoso:
LGTM

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -324,6 +325,14 @@ private[http] object HttpServerBluePrint {
     }
   }
 
+  /**
+   * Ensures that the user handler
+   *  - produces exactly one response per request
+   *  - has not more than ``pipeliningLimit`` responses outstanding
+   */
+  def userHandlerGuard(pipeliningLimit: Int): BidiFlow[HttpResponse, HttpResponse, HttpRequest, HttpRequest, Unit] =
+    One2OneBidiFlow[HttpRequest, HttpResponse](pipeliningLimit).reversed
```
drewhk: Ah, I meant renaming the actual class.

Issue #19050, opened 2015-11-30
Make HttpServerBlueprint reusable and actually reuse it


jrudolph:
As discussed at https://github.com/akka/akka/pull/18979#issuecomment-159315798

Currently, we think the performance benefit is small under the assumption that the (then-reusable) pre-materialization cost is small in relation to the actual materialization. IIUC this may change in the future if part of the fused infrastructure can be prepared before materialization.

Current obstacles:
 - [ ] Dynamically supplying a websocket user-handler: this requires to feed the materialized value of the provided user-handler stubs into the HTTP renderer (which later instantiates the WS user-handler). The suggested solution of accessing the materialized value during building a graph is still quite complicated and prone to race conditions.
 - [x] The remote address of the peer is currently passed to the parser during creation, removing this obstacle would need a method to parameterize materialization. The suggested workaround of materializing a function or promise (or any kind of cell) to pass the value in *after* materialization is again prone to race conditions (or would complicate the target stage with extra waiting code to receive the value).
- [ ] Actually reuse the blueprint // added @ktoso 

/cc @sirthias

drewhk:
Can't the remote address passed in as an Attribute? Not the most elegant, I admit, and has its own issues, but as a first approach it might work.

ktoso:
Thanks a lot for the hints @jrudolph – I'll start investigating this right away, and for this week in general.

ktoso:
Point 2) addressed in: https://github.com/akka/akka/issues/19060
Point 1) ready to be worked on with the security blancket of the Autobahn tests now running on jenkins as well: https://github.com/akka/akka/issues/18563

ktoso:
AFAIR this has been done @drewhk @rkuhn ?

drewhk:
It is reusable, but it is not reused.

ktoso:
This ticket was only about make reusable AFAICS, closing then.

drewhk:
Then make a ticket about reusing it. and put it into 2.4.2 since this is the most relevant optimization.

ktoso:
Ok, made it so

ktoso:
Realised we have another ticket about the exact same thing: https://github.com/akka/akka/issues/19232
Closing this one

PR #19060, opened 2015-12-01
[WIP] =htc #19050 reusable http blueprint: attribute injected rem-addr

ktoso:
Moves the remote address injection to Attributes, which allows that part of the HTTP blueprint to become reusable. Need to decide if/how we expose this stage. May be useful enough to include on FlowOps I think actually.

Addresses point two of: https://github.com/akka/akka/issues/19050

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
@@ -1,0 +1,24 @@
+/*
+ * Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.impl.engine.server
+
+import java.net.InetSocketAddress
+
+import akka.stream.Attributes
+
+/**
+ * INTERNAL API
+ * Internally used attributes set in the HTTP pipeline.
+ * May potentially be opened up in the future.
+ */
+private[akka] object HttpAttributes {
+  import Attributes._
+
+  private[akka] final case class RemoteAddress(address: Option[InetSocketAddress]) extends Attribute
+
+  private[akka] def remoteAddress(address: Option[InetSocketAddress]) =
+    Attributes(RemoteAddress(address))
+
+}
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -47,16 +47,16 @@ import akka.util.ByteString
  *                 +----------+                                   +-------------+  Context    +-----------+
  */
 private[http] object HttpServerBluePrint {
-  def apply(settings: ServerSettings, remoteAddress: Option[InetSocketAddress], log: LoggingAdapter)(implicit mat: Materializer): Http.ServerLayer =
-    stack(settings, remoteAddress, log)
+  def apply(settings: ServerSettings, remoteAddress: Option[InetSocketAddress], log: LoggingAdapter)(implicit mat: Materializer): Http.ServerLayer = {
+    val theStack =
+      userHandlerGuard(settings.pipeliningLimit) atop
+        requestPreparation(settings) atop
+        controller(settings, log) atop
+        parsingRendering(settings, log) atop
+        websocketSupport(settings, log) atop
+        unwrapTls
 
-  def stack(settings: ServerSettings, remoteAddress: Option[InetSocketAddress], log: LoggingAdapter)(implicit mat: Materializer): Http.ServerLayer = {
-    userHandlerGuard(settings.pipeliningLimit) atop
-      requestPreparation(settings, remoteAddress) atop
-      controller(settings, log) atop
-      parsingRendering(settings, log) atop
-      websocketSupport(settings, log) atop
-      unwrapTls
+    theStack.withAttributes(HttpAttributes.remoteAddress(remoteAddress))
   }
 
   val unwrapTls: BidiFlow[ByteString, SslTlsOutbound, SslTlsInbound, ByteString, Unit] =
@@ -83,27 +83,12 @@ private[http] object HttpServerBluePrint {
   def controller(settings: ServerSettings, log: LoggingAdapter): BidiFlow[HttpResponse, ResponseRenderingContext, RequestOutput, RequestOutput, Unit] =
     BidiFlow.fromGraph(new ControllerStage(settings, log)).reversed
 
-  def requestPreparation(settings: ServerSettings, remoteAddress: Option[InetSocketAddress])(implicit mat: Materializer): BidiFlow[HttpResponse, HttpResponse, RequestOutput, HttpRequest, Unit] = {
-    import settings._
-
+  def requestPreparation(settings: ServerSettings)(implicit mat: Materializer): BidiFlow[HttpResponse, HttpResponse, RequestOutput, HttpRequest, Unit] = {
     val prepareRequest =
       Flow[RequestOutput]
         .splitWhen(x ⇒ x.isInstanceOf[MessageStart] || x == MessageEnd)
         .via(headAndTailFlow)
-        .map {
-          case (RequestStart(method, uri, protocol, hdrs, createEntity, _, _), entityParts) ⇒
-            val effectiveMethod = if (method == HttpMethods.HEAD && transparentHeadRequests) HttpMethods.GET else method
-            val effectiveHeaders =
-              if (settings.remoteAddressHeader && remoteAddress.isDefined)
-                headers.`Remote-Address`(RemoteAddress(remoteAddress.get)) +: hdrs
-              else hdrs
-
-            val entity = createEntity(entityParts) withSizeLimit parserSettings.maxContentLength
-            HttpRequest(effectiveMethod, uri, effectiveHeaders, entity, protocol)
-          case (_, src) ⇒ src.runWith(Sink.ignore)
-        }.collect {
-          case r: HttpRequest ⇒ r
-        }
+        .via(requestStartOrRunIgnore(settings))
         // FIXME #16583 / #18170
         // `buffer` is needed because of current behavior of collect which will queue completion
         // behind an ignored (= not collected) element if there is no demand.
@@ -112,6 +97,35 @@ private[http] object HttpServerBluePrint {
 
     BidiFlow.fromFlows(Flow[HttpResponse], prepareRequest)
   }
+
+  def requestStartOrRunIgnore(settings: ServerSettings)(implicit mat: Materializer): Flow[(ParserOutput.RequestOutput, Source[ParserOutput.RequestOutput, Unit]), HttpRequest, Unit] =
+    Flow.fromGraph(new FlowStage[(RequestOutput, Source[RequestOutput, Unit]), HttpRequest, Unit]("RequestStartThenRunIgnore") {
+      override def createLogicAndMaterializedValue(inheritedAttributes: Attributes) = (new GraphStageLogic(shape) {
+        val remoteAddress = inheritedAttributes.get[HttpAttributes.RemoteAddress].flatMap(_.address)
+
+        setHandler(in, new InHandler {
+          override def onPush(): Unit = grab(in) match {
+            case (RequestStart(method, uri, protocol, hdrs, createEntity, _, _), entityParts) ⇒
+              val effectiveMethod = if (method == HttpMethods.HEAD && settings.transparentHeadRequests) HttpMethods.GET else method
+              val effectiveHeaders =
+                if (settings.remoteAddressHeader && remoteAddress.isDefined)
+                  headers.`Remote-Address`(RemoteAddress(remoteAddress.get)) +: hdrs
+                else hdrs
+
+              val entity = createEntity(entityParts) withSizeLimit settings.parserSettings.maxContentLength
+              push(out, HttpRequest(effectiveMethod, uri, effectiveHeaders, entity, protocol))
+
+            case (wat, src) ⇒
+              src.runWith(Sink.ignore)
+              pull(in)
+          }
+        })
+
+        setHandler(out, new OutHandler {
+          override def onPull(): Unit = pull(in)
+        })
+      }, ())
+    })
 
   def parsing(settings: ServerSettings, log: LoggingAdapter): Flow[ByteString, RequestOutput, Unit] = {
     import settings._
--- a/akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
+++ b/akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
@@ -73,6 +73,16 @@ abstract class SinkStage[In, M](name: String) extends GraphStageWithMaterialized
   override val shape: SinkShape[In] = SinkShape(in)
 }
 
+/**
+ * A FlowStage represents a reusable graph stream processing stage. A FlowStage consists of a [[akka.stream.Shape]] which describes
+ * its input and output ports.
+ */
+abstract class FlowStage[In, Out, M](name: String) extends GraphStageWithMaterializedValue[FlowShape[In, Out], M] {
+  val in: Inlet[In] = Inlet[In](name + ".in")
+  val out: Outlet[Out] = Outlet[Out](name + ".out")
+  override val shape: FlowShape[In, Out] = FlowShape(in, out)
+}
+
 private object TimerMessages {
   final case class Scheduled(timerKey: Any, timerId: Int, repeating: Boolean) extends DeadLetterSuppression
   final case class Timer(id: Int, task: Cancellable)


akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
```diff
@@ -0,0 +1,20 @@
...
+/** Internal attributes used by the HTTP pipeline */
+object HttpAttributes {
+  import Attributes._
+
+  final case class RemoteAddress(address: Option[InetSocketAddress]) extends Attribute
+
+  def remoteAddress(address: Option[InetSocketAddress]) =
+    Attributes(RemoteAddress(address))
+
+}
```
ktoso: technically don't need to be private, may even be useful I guess.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -83,24 +86,27 @@ private[http] object HttpServerBluePrint {
   def controller(settings: ServerSettings, log: LoggingAdapter): BidiFlow[HttpResponse, ResponseRenderingContext, RequestOutput, RequestOutput, Unit] =
     BidiFlow.fromGraph(new ControllerStage(settings, log)).reversed
 
-  def requestPreparation(settings: ServerSettings, remoteAddress: Option[InetSocketAddress])(implicit mat: Materializer): BidiFlow[HttpResponse, HttpResponse, RequestOutput, HttpRequest, Unit] = {
+  //  def requestPreparation(settings: ServerSettings, remoteAddress: Option[InetSocketAddress])(implicit mat: Materializer): BidiFlow[HttpResponse, HttpResponse, RequestOutput, HttpRequest, Unit] = {
```
ktoso: note to self, remove this

akka-http-tests/src/test/scala/akka/http/scaladsl/server/DontLeakActorsOnFailingConnectionSpecs.scala
```diff
@@ -13,15 +11,12 @@ import akka.event.Logging
...
 import akka.stream.scaladsl.{ Sink, Source }
 import akka.testkit.TestProbe
 import com.typesafe.config.ConfigFactory
 import org.scalatest.{ Matchers, BeforeAndAfterAll, WordSpecLike }
-import sun.security.provider.ConfigFile
 
-import scala.concurrent.duration.FiniteDuration
-import scala.concurrent.{ Await, Future }
 import scala.concurrent.duration._
 import scala.util.{ Failure, Success, Try }
```
ktoso: will remove from PR

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1340,6 +1341,38 @@ trait FlowOps[+Out, +Mat] {
...
 
+  // TODO find this buddy a home (sounds useful enough to expose?)
+
+  def exposeAttribute[A <: Attribute]()(implicit t: ClassTag[A]): Repr[(Out, Option[A]), Mat] =
+    exposeAttribute(ConstantFun.scalaTuplify[Out, Option[A]])(t)
+
+  def exposeAttribute[A <: Attribute, E](exposeAs: (Out, Option[A]) ⇒ E)(implicit t: ClassTag[A]): Repr[E, Mat] =
+    this.via(Flow.fromGraph(new ExposeAttributeStage[Out, A, E, Mat](t.runtimeClass.asInstanceOf[Class[A]], exposeAs)))
+
+  class ExposeAttributeStage[In, A <: Attribute, E, M](attrClazz: Class[A], exposeFn: (In, Option[A]) ⇒ E)
```
ktoso: TODO: will move to a proper place once we decide if we keep the op here or should hide it.

drewhk: hm, this stage have a hacky feel about it...

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
viktorklang: Except that it isn't private.
Since this sort of smells like a hack (secret parameter), I think it ought to stay private until the solution is recognized as workable in the general sense. Wdyt?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -83,24 +86,27 @@ private[http] object HttpServerBluePrint {
...
         .map {
-          case (RequestStart(method, uri, protocol, hdrs, createEntity, _, _), entityParts) ⇒
+          case ((RequestStart(method, uri, protocol, hdrs, createEntity, _, _), entityParts), remoteAddress) ⇒
             val effectiveMethod = if (method == HttpMethods.HEAD && transparentHeadRequests) HttpMethods.GET else method
+            val addr = remoteAddress.flatMap(_.address)
             val effectiveHeaders =
-              if (settings.remoteAddressHeader && remoteAddress.isDefined)
-                headers.`Remote-Address`(RemoteAddress(remoteAddress.get)) +: hdrs
+              if (settings.remoteAddressHeader && addr.isDefined)
+                headers.`Remote-Address`(RemoteAddress(addr.get)) +: hdrs
```
viktorklang: Since you are using `.get` anyway, how about discarding the flatMap?
~~~scala
if (settings.remoteAddressHeader && remoteAddress.isDefined)
  headers.`Remote-Address`(RemoteAddress(remoteAddress.get.address)) +: hdrs
~~~

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: The solution leaves a lot to be desired.

Perhaps we can restate the problem?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpAttributes.scala
ktoso: Yeah, I PRed early to see if this option is deemed ok to pursue at all or not; they'd be private then.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
ktoso: The problem is "have http pipeline reusable" this address is injected though, so we need a way to do that, and pull it out once the stream runs. Attributes are a way of doing that; another one would be inventing what "parameterized materialization" would mean, and where the value could be pulled out from then.

Specifically it's about:

>  The remote address of the peer is currently passed to the parser during creation, removing this obstacle would need a method to parameterize materialization. The suggested workaround of materializing a function or promise (or any kind of cell) to pass the value in after materialization is again prone to race conditions (or would complicate the target stage with extra waiting code to receive the value).

from: https://github.com/akka/akka/issues/19050

Other ideas?

drewhk: Not really. Materialized value is the other option, but it is even more invasive. Attributes *are* parametrized materialization, so I don't see it as a total misuse, but I don't like a generalized stage to inject attributes into a stream. It feels wrong somehow.

ktoso: > Attributes are parametrized materialization

Yeah I was just thinking that after I wrote the above.

> don't like a generalized stage to inject attributes into a stream. It feels wrong somehow.

I see what you mean here... We could start out by hiding it as Http specific trick for now, until we find something better?

ktoso: OR... I can actually try making the collect which uses this value a GraphStage – then it can properly access the value.
I'll try that and see how it gors.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
ktoso: the inner `address` is also an Option.

ktoso: outer: "the Attribute was there (here = always actually"
inner: "the stack() method was called with remoteAddress: Option[...]"

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
jrudolph: @ktoso Isn't the affected combinator only a `map`? This should be quite easily be convertible into a PushStage and also combine the subsequent `collect` functionality into it as well.

(Still it's somewhat of a bummer that a generic solution seems so hard to achieve.)

ktoso: > @ktoso Isn't the affected combinator only a map? This should be quite easily be convertible into a PushStage and also combine the subsequent collect functionality into it as well.

Yeah, I'm trying exactly that right now (I'm making it a GraphStage, since the least overhead).

ktoso:
Pushed the different impl, please have a look guys.
It is injecting attributes, and handling in a custom stage instead of `map + collect`

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1340,6 +1341,8 @@ trait FlowOps[+Out, +Mat] {
         FlowShape(bcast.in, bcast.out(0))
     })(matF)
 
+  final class AttributeMissingException(msg: String) extends RuntimeException(msg)
```
viktorklang: Not used :)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -19,6 +19,7 @@ import scala.collection.immutable
 import scala.concurrent.Future
 import scala.concurrent.duration.{ Duration, FiniteDuration }
 import scala.language.higherKinds
+import scala.reflect.ClassTag
```
viktorklang: Not used

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -3,7 +3,7 @@
  */
 package akka.stream.scaladsl
 
-import akka.event.LoggingAdapter
+import akka.event.{ Logging, LoggingAdapter }
```
viktorklang: Not used.

akka-stream/src/main/scala/akka/stream/impl/ConstantFun.scala
```diff
@@ -17,9 +17,13 @@ private[akka] object ConstantFun {
...
 
+  private[this] val ScalaTuplifyFunction = (l: Any, r: Any) ⇒ (l, r)
+
   def javaCreatePairFunction[A, B]: JFun2[A, B, JPair[A, B]] = JavaPairFunction.asInstanceOf[JFun2[A, B, JPair[A, B]]]
 
   def javaIdentityFunction[T]: JFun[T, T] = JavaIdentityFunction.asInstanceOf[JFun[T, T]]
 
   def scalaIdentityFunction[T]: T ⇒ T = ScalaIdentityFunction.asInstanceOf[T ⇒ T]
+
+  def scalaTuplify[L, R]: (L, R) ⇒ (L, R) = ScalaTuplifyFunction.asInstanceOf[(L, R) ⇒ (L, R)]
```
viktorklang: Not used :)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AttributesSpec.scala
```diff
@@ -3,12 +3,10 @@
  */
 package akka.stream.scaladsl
 
-import akka.stream.ActorMaterializer
-import akka.stream.ActorMaterializerSettings
-import akka.stream.Attributes
+import akka.event.Logging
```
viktorklang: Nothing used here :)

akka-stream/src/main/scala/akka/stream/impl/ConstantFun.scala
ktoso: lol, it's a WIP, please check the Stage :wink:

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -51,12 +51,15 @@ private[http] object HttpServerBluePrint {
     stack(settings, remoteAddress, log)
 
   def stack(settings: ServerSettings, remoteAddress: Option[InetSocketAddress], log: LoggingAdapter)(implicit mat: Materializer): Http.ServerLayer = {
-    userHandlerGuard(settings.pipeliningLimit) atop
-      requestPreparation(settings, remoteAddress) atop
-      controller(settings, log) atop
-      parsingRendering(settings, log) atop
-      websocketSupport(settings, log) atop
-      unwrapTls
+    val theStack =
```
viktorklang: This is undeniably beautiful

ktoso: very true, like legos being plucked together :smile:

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -51,12 +51,15 @@ private[http] object HttpServerBluePrint {
...
+      userHandlerGuard(settings.pipeliningLimit) atop
+        requestPreparation(settings) atop
```
viktorklang: Why can't you pass the `remoteAddress` as a parameter to requestPreparation?

ktoso: That's exactly what we need to get _away_ from, in order to be able to cache the http blueprint.
// I don't know yet how we'll handle the `settings`.

viktorklang: Have we considered passing `remoteAddress` as a prepended part of the payload so the parser can extract it from the bytestream? Or is that stupid?

ktoso: That sounds somewhat more hacky and disappointing in a way, IMHO Attributes _are_ the side channel we have built for these kinds of situations. Putting it into the Flow is something impls without an infra around the streams like we have would start out with I guess.

viktorklang: @ktoso TBH the attribute sounds like a hack to me. It's like using ThreadLocals to pass parameters to a method call.

viktorklang: But I guess as long as the RemoteAddress Attribute is never exposed to any user we are always free to change how it's implemented under the hood.

ktoso: Right, and if we want to _really hide it_, we can by making:

```
final case class Attributes(/* this: */attributeList: List[Attributes.Attribute] = Nil) {
```

a `def` instead that hides any internal Attributes. 

That's the only place where currently someone may discover this internal attribute.

jrudolph: We could also turn this logic around and say the Attribute is another actual user API to get the peer address. It's not so much different from the hack the `Remote-Address` header is on the HTTP level to get metadata to the user.

viktorklang: or you zip it with a Source.maybe and send the address there?

-- 
Cheers,
√

ktoso: I still feel that the address is "a parameter to this materialization", which is what attributes are - it is the same blueprint, just parameterized with a different "setting" if you will.
Making it directly available from "the address Source" feels awkard, one has to start thinking how many times it has to emit etc... can it be maybe or must it be repeat etc.

ktoso: > We could also turn this logic around and say the Attribute is another actual user API to get the peer address. It's not so much different from the hack the Remote-Address header is on the HTTP level to get metadata to the user.

Yeah, very true actually. We have mechanisms like that with injecting headers already.

viktorklang: Yes, I understand that, and I can sympathize with the situation, but don't you agree that it is almost identical to using ThreadLocals to pass parameters to a method?

I'd recommend trying out the Source.maybe + zipWith/zip solution so we at least picked the least of evils.

jrudolph: > I'd recommend trying out the Source.maybe + zipWith/zip solution so we at least picked the least of evils.

This, however, conflates the run times. The address is already available before materialization, so why would you defer passing it to running the stream? That's like writing a console program and, instead of passing arguments directly, only pass a socket descriptor and then later try to read the parameters from the socket :)

viktorklang: @jrudolph Perhaps I'm missing context. When you say "THE address" do you mean that the address is *always* the same (in which case it ought to be possible to provide when the graph is created, not when it is materialized) or do you mean "AN address" which means that it needs to be provided *after* graph construction but *before* materialization?

jrudolph: I mean "AN address" that needs to supplied after compilation but before running.

viktorklang: Great, then we are on the same page. Then do we agree that using Source.maybe *would* also solve the problem of injecting the *current* Address but without having to resort to attributes?

ktoso: Well ok, I can give the Source approach a shot so we can compare the two, but still feel it's wrong.

viktorklang: I agree, but I'd rather choose the lesser of 2 evils 😊

-- 
Cheers,
√

jrudolph: @viktorklang we agree that your approach is possible but I don't find it better :)

viktorklang: I can sympathize with that 😊

-- 
Cheers,
√

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -658,26 +662,29 @@ object Http extends ExtensionId[HttpExt] with ExtensionIdProvider {
...
      * Handles the connection with the given flow, which is materialized exactly once
      * and the respective materialization result returned.
      */
-    def handleWith[Mat](handler: Flow[HttpRequest, HttpResponse, Mat])(implicit fm: Materializer): Mat =
-      flow.joinMat(handler)(Keep.right).run()
+    def handleWith[Mat](handler: Flow[HttpRequest, HttpResponse, Mat])(implicit fm: Materializer): Mat = {
+      val fillMeInAndUserMat = flow.joinMat(handler)(Keep.both).run()
+      fillMeInAndUserMat._1.success(Some(Some(remoteAddress)))
+      fillMeInAndUserMat._2
+    }
```
ktoso: Stuff like this would have to be done everywhere we materialize the server. It's hard to keep it in one place.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -83,34 +84,38 @@ private[http] object HttpServerBluePrint {
   def controller(settings: ServerSettings, log: LoggingAdapter): BidiFlow[HttpResponse, ResponseRenderingContext, RequestOutput, RequestOutput, Unit] =
     BidiFlow.fromGraph(new ControllerStage(settings, log)).reversed
 
-  def requestPreparation(settings: ServerSettings, remoteAddress: Option[InetSocketAddress])(implicit mat: Materializer): BidiFlow[HttpResponse, HttpResponse, RequestOutput, HttpRequest, Unit] = {
-    import settings._
+  def requestPreparation(settings: ServerSettings)(implicit mat: Materializer): BidiFlow[HttpResponse, HttpResponse, RequestOutput, HttpRequest, Promise[Option[Option[InetSocketAddress]]]] = {
+    // yes that's the type we actually need... so much can go wrong here, proper completion is (Some(Some(addr)), otherwise one has to Some(None)...
+    // becaue of completion semantics of success(None) on Source.maybe
+    val remoteAddressSource: Source[Option[InetSocketAddress], Promise[Option[Option[InetSocketAddress]]]] =
+      Source.maybe[Option[InetSocketAddress]]
```
ktoso: Here's the maybe source; Notice the type of the Promise btw (since `success(None)` == `complete`, and `success(Some(None))` == "no remote address given")

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -83,34 +84,38 @@ private[http] object HttpServerBluePrint {
...
 
     val prepareRequest =
       Flow[RequestOutput]
         .splitWhen(x ⇒ x.isInstanceOf[MessageStart] || x == MessageEnd)
         .via(headAndTailFlow)
+        .zipWithMat(remoteAddressSource)({ case ((o, src), addr) ⇒ (o, src, addr) })(Keep.right)
```
ktoso: I think that's not correct, it would need to be `remoteAddresSource.cache()` AFAICS (for simple things this already works though and is enough for the discussion at hand)

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -83,34 +84,38 @@ private[http] object HttpServerBluePrint {
...
         }
+        .collect { case r: HttpRequest ⇒ r }
         // FIXME #16583 / #18170
         // `buffer` is needed because of current behavior of collect which will queue completion
         // behind an ignored (= not collected) element if there is no demand.
         // `buffer` will ensure demand and therefore make sure that completion is reported eagerly.
         .buffer(1, OverflowStrategy.backpressure)
 
-    BidiFlow.fromFlows(Flow[HttpResponse], prepareRequest)
+    BidiFlow.fromFlowsMat(Flow[HttpResponse], prepareRequest)((_, r) ⇒ r)
```
ktoso: Keep.right, I know

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
ktoso: Well, to at least show the other evil, here it is: https://github.com/ktoso/akka/commit/98a068ab12552c253ddbc82610eb2f725b8ef44a (Source.maybe + completing it with the remote address)

It looks and feels very weird and it's hard to track down all the places where one has to complete the maybe source to get the server running. Technically `serverLayer` is even user-api AFAIR, so we'd need to document "yeah and then you need to get the materialized value and complete it" which sounds _really weird_.

It of course also is slower, since we're halting any processing until stream is materialized -> promise is completed -> future completes, triggering the zip. We _have_ the value, and setting it early, as attribute, makes a lot of sense from a usability, maintenance as well as performance perspective. 

Thus, I'm suggesting to stick with the `attribute + graphStage` as implemented in https://github.com/ktoso/akka/commit/99e40f918b482a89f56ff5c789d0a75ef7f09563 and one question which should be answered then is: do we make accessing that attribute an actual user API (as @jrudolph suggested), or do we hide it (needs `InternalAttribute` marker interfaces, but is doable ofc).

drewhk: Yes, I just wanted to mention that the main drawback of using Source.maybe is that it will pollute the materialized value. The correct way for parametrized materialization would be of course to have yet another type parameter (instead of the dynamically typed attribute), and then combinators like `viaAtr` and `viaMatAtr`, etc. People would crucify us though :)

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -658,26 +662,29 @@ object Http extends ExtensionId[HttpExt] with ExtensionIdProvider {
...
 
     /**
      * Handles the connection with the given handler function.
      */
     def handleWithSyncHandler(handler: HttpRequest ⇒ HttpResponse)(implicit fm: Materializer): Unit =
-      handleWith(Flow[HttpRequest].map(handler))
+      handleWith(Flow[HttpRequest].map(handler)) // TODO "fill in" remoteAddress, otherwise server never starts
```
viktorklang: Doesn't `handleWith` do this?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
viktorklang: I'd probably add the Option[InetAddress] to RequestStart instead of creating yet another tuple.
.zipWithMat(remoteAddressSource)((hAt, addr) => (hAt._1.withRemoteAddress(addr), hAt._2))(Keep.right)

viktorklang: Also, I just realized that it can be even further zimplified by folding it into the following map-operation, you can do all that work in zipWith

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
ktoso: right, here it is taken care of.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
viktorklang: Agreed, I tried to do what you did locally, but I think both the attribute AND the Source.maybe solution suffers from the same problem: it is too easy to forget to set it.

ktoso: True, one has to set it somewhere.Where one does it is inverted though: with the attribute, we need to set it _in_ the blueprint, with the materialized value it has to be set at "materialization-site". The second being in more places.

viktorklang: I don't understand why that's the case, the place where the remoteAddress is know is the same?

jrudolph: You can set an attribute on a Flow but you can deal with the materialized value only after materialization.

jrudolph: On the other hand, you could also map the materialized value Promise at the same point where you would set the attribute. 

Still, as @ktoso said, using the materialized value makes the API more error prone for all consumers who don't want to set the RemoteAddress which is an optional feature.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -602,7 +606,7 @@ object Http extends ExtensionId[HttpExt] with ExtensionIdProvider {
    *                +------+
    * }}}
    */
-  type ServerLayer = BidiFlow[HttpResponse, SslTlsOutbound, SslTlsInbound, HttpRequest, Unit]
+  type ServerLayer = BidiFlow[HttpResponse, SslTlsOutbound, SslTlsInbound, HttpRequest, Promise[Option[Option[InetSocketAddress]]]]
```
jrudolph: The materialized value needs to be documented everywhere it turns up.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
viktorklang: >On the other hand, you could also map the materialized value Promise at the same point where you would set the attribute.

Indeed. I wonder what the impact of that would be.

>Still, as @ktoso said, using the materialized value makes the API more error prone for all consumers who don't want to set the RemoteAddress which is an optional feature.

Absolutely. But using the Attribute means you first of all have to know that it exists and that you may want to set it. Right now it becomes a hidden setting. Do you agree with that assessment?

ktoso: > Absolutely. But using the Attribute means you first of all have to know that it exists and that you may want to set it. Right now it becomes a hidden setting. Do you agree with that assessment?

Right, that is certainly true: someone needs to set it, correct. We need to do this for each materialization (connection).
It is different in semantics though (and I do think we prefer the attribute-way semantics):

- attribute when I'd forget to set, we're in the (pseudocode) `None[Option[InetSocketAddress]], and I can happily continue without it.
  - forgetting to complete the materialized promise causes the server to lock-up waiting for it. It's tricky to "oh by default we'll emit None", since the completion must happen after materialization, so there must be a delay by construction.

A number of other downsides of the element injecting were mentioned in this PR as well (inherent measurable delay, increasing the burden of the server's type signature) and explained the both approaches to both @rkuhn, @drewkh and @jrudolph which also think this is the lesser evil. 

So I hope your argument of picking the lesser of two evils is satisfied and we can move on - there's loads of more work here, and likely more tricky than this simple one :-) I got the WS TCK running on jenkins, so feel more comfortable to tackle part 2 of this effort now, which relates to making the user provided WS accessible in a reusable blueprint – _that_ will be way more tricky than this was I think.

viktorklang:
>A number of other downsides of the element injecting were mentioned in this PR as well (inherent measurable delay, increasing the burden of the server's type signature) and explained the both approaches to both @rkuhn, @drewkh and @jrudolph which also think this is the lesser evil.

I can definitely accept the RemoteAddressAttribute proposal, as long as it stays hidden from the user so we are free to change it down the line without impacting compatibility. :)

ktoso:
Okey. So current state (ouch I see something failed :<) contains that style, and only `private[akka]` attribute. We can go the extra mile to hide it from ever being seen if we really want to, as explained in https://github.com/akka/akka/pull/19060#discussion_r46353772 do we need to do that?

viktorklang:
Nah, even if they find it they are not "allowed" to see the type. And we
should mark it as INTERNAL API so it could go away without notice.

On Thu, Dec 3, 2015 at 9:13 PM, Konrad Malawski <notifications@github.com>
wrote:

> Okey. So current state (ouch I see something failed :<) contains that
> style, and only private[akka] attribute. We can go the extra mile to hide
> it from ever being seen if we really want to, as explained in #19060
> (comment) <https://github.com/akka/akka/pull/19060#discussion_r46353772>
> do we need to do that?
>
> —
> <https://github.com/akka/akka/pull/19060#issuecomment-161769706>.
>



-- 
Cheers,
√

ktoso:
Yup, agree on that as well. I'll figure out what failed here in the morning then and push towards merging soon-ish.

rkuhn:
I can’t find the discussion which is partially quoted here, but I’d like to clarify some fundamental points:

* the materialized value is something that the engine emits that is “about to be running”, it may contain per-materialization information or levers for future control (cancellation for example)
* the stream itself contains data elements, it is the business domain that we are talking about here
* the attributes give inputs for the materialization process, they configure the materializer and the stages

In this sense passing the remote address via attributes is the only logical choice and I see no reason to qualify this as “internal hack that we may want to change later”. In particular my vote is to make this attribute user API and document it accordingly.

viktorklang:
The "hack part", to me, is that it is a *hidden* parameter, which is very
much easy to forget or not know—i.e. how do I know which hidden attributes
apply to X where is is some Graph[_]? It's similar to using ThreadLocals to
pass parameters to methods rather than using method parameters.

> the attributes give inputs for the materialization process, they
configure the materializer and the stages

If the attributes to not apply to all stages, how do I know which ones
apply at which point, and does my code silently break if my graph subtly
changes?

rkuhn:
Exposing this parameter explicitly has an enormous cost, it would require even more type parameters plus extractor functions (or a Reader monad) mirroring our materialized value combinators. This is clearly not tolerable, so implying that “there is a proper solution” does not really help.

The line between stream elements and attributes should be quite clear: an Attribute is something optional, it may be set or it may not, while stream connections are statically (or in some cases at runtime) mandated to be fulfilled.

As Konrad has pointed out, materialized values are not appropriate for configuration because users cannot be expected to perform some required actions on them—the main issue is that these actions would necessarily happen asynchronously and there is no reasonable maximum waiting time before concluding that the user has forgotten to do it.

To your question on “silent breakage”: why should any code that does not know that an HttpServerBlueprint is in involved ever care about this attribute? The advantage of this approach is precisely that the knowledge of such an implementation detail can remain encapsulated. The rules for attribute scoping even make it composable when multiple HttpServerBlueprints are involved.

viktorklang:
Something being optional is completely fine. But that assumes that the user
knows that it exists. What's the discoverability story?
(I'm asking this because if the chosen solution is deemed to be good enough
to inspire more similar use-cases, the story of discoverability needs to be
told.)

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -113,6 +98,33 @@ private[http] object HttpServerBluePrint {
     BidiFlow.fromFlows(Flow[HttpResponse], prepareRequest)
   }
 
+  def requestStartThenRunIgnore(settings: ServerSettings)(implicit mat: Materializer): Flow[(ParserOutput.RequestOutput, Source[ParserOutput.RequestOutput, Unit]), HttpRequest, Unit] =
```
viktorklang: Isn't it "requestStartOrRunIgnore"?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -83,27 +83,12 @@ private[http] object HttpServerBluePrint {
...
-
-            val entity = createEntity(entityParts) withSizeLimit parserSettings.maxContentLength
-            HttpRequest(effectiveMethod, uri, effectiveHeaders, entity, protocol)
-          case (_, src) ⇒ src.runWith(Sink.ignore)
-        }.collect {
-          case r: HttpRequest ⇒ r
-        }
+        .via(requestStartThenRunIgnore(settings))
         // FIXME #16583 / #18170
         // `buffer` is needed because of current behavior of collect which will queue completion
```
viktorklang: Is the buffer still needed without the collect?

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -113,6 +98,33 @@ private[http] object HttpServerBluePrint {
...
+            case (_, src) ⇒ src.runWith(Sink.ignore)
+          }
+        })
+
+        setHandler(out, new OutHandler {
+          override def onPull(): Unit = pull(in)
+        })
+      } -> ()
+    })
+
```
viktorklang: @ktoso Kudos on this version of the code, looks really clean

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
ktoso: right, thanks!

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
ktoso: Good question, I'll see if I'll be able to remove it

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
ktoso: thanks :smile:

ktoso:
// lol, failure was missing sbt-jar (because sbt plugin update, which I fixed last night, need to re-run):

ktoso:
PLS BUILD

ktoso:
Failure fixed, missed one `pull()` [call](https://github.com/akka/akka/pull/19060/files#diff-623ef36775342f2db6b86855f7ffea64R120).

ktoso:
I really want to merge this to not go in the weekend with an empty sad feeling, so doing so now since this PR has been discussed at length.
Let's talk about the discoverability story when we meet next week perhaps? :)

rkuhn:
Yes, indeed! No reasons for empty sad feelings :-)

ktoso:
True, :christmas_tree: is coming ;-)

PR #19061, opened 2015-12-01
=pro #19054 update to sbt-dependency-graph 0.8.0

### LINKED ISSUE #19054: Comply with new sbt-dependency-graph(0.8.0)
Hi,

I have in my .sbt/0.13/plugins/plugins.sbt a fresh version of sbt-dependency-graph that is picked up by sbt on akka project.
Unfortunately akka relies on older version(0.7.5), we could align to the newest one.
These changes introduced issue: 
https://github.com/jrudolph/sbt-dependency-graph/commit/7a5126ef875ee24f364c7b0be591893fd5d80d4a

jrudolph:
Fixes #19054.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,1 @@
 enablePlugins(akka.UnidocRoot, akka.TimeStampede, akka.UnidocWithPrValidation)
-
-net.virtualvoid.sbt.graph.Plugin.graphSettings
--- a/project/ValidatePullRequest.scala
+++ b/project/ValidatePullRequest.scala
@@ -4,8 +4,9 @@
 package akka
 
 import com.typesafe.tools.mima.plugin.MimaKeys.reportBinaryIssues
-import net.virtualvoid.sbt.graph.backend.IvyReport
-import net.virtualvoid.sbt.graph.ModuleId
+import net.virtualvoid.sbt.graph.backend.SbtUpdateReport
+import net.virtualvoid.sbt.graph.DependencyGraphKeys._
+import net.virtualvoid.sbt.graph.ModuleGraph
 import org.kohsuke.github._
 import sbtunidoc.Plugin.UnidocKeys.unidoc
 import sbt.Keys._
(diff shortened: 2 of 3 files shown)


build.sbt
```diff
@@ -1,3 +1 @@
 enablePlugins(akka.UnidocRoot, akka.TimeStampede, akka.UnidocWithPrValidation)
-
-net.virtualvoid.sbt.graph.Plugin.graphSettings
```
ktoso: cool

project/ValidatePullRequest.scala
```diff
@@ -101,7 +101,7 @@ object ValidatePullRequest extends AutoPlugin {
           if (f1.exists()) f1 else f2
         }
 
-        val deps = IvyGraphMLDependencies.graph(ivyReportFile.getAbsolutePath)
+        val deps = IvyReport.fromReportFile(ivyReportFile.getAbsolutePath)
```
ktoso: great, thanks!

ktoso:
LGTM, I'll also investigate the PR Validation output to make sure it works as expected still (likely does, but I'll check)

jrudolph:
I wonder if the logic could be updated to use the new [SbtUpdateReport backend](https://github.com/jrudolph/sbt-dependency-graph/blob/master/src/main/scala/net/virtualvoid/sbt/graph/backend/SbtUpdateReport.scala). Is there any special reason why you (re-)implemented the logic to lookup the ivy report file? Would it be reasonable to depend on `moduleGraph` (for each configuration) directly?

jrudolph:
Yes, please validate the validation. I wouldn't know if I broke something.

jrudolph:
I added two additional commits. The first one changes the graph to be generated from sbt's `update` result directly. The second one moves some of the keys into Global/Build scope which prevents re-executing them for every project if the validation is run by project aggregation. Is aggregation used for validation?

jrudolph:
It seems validation uses aggregation, indeed, so the last commit will prevent `git` and the access to github.com to run for every subproject. This probably won't speed up things a lot but removes a bit of clutter from the log.

ktoso:
> Is there any special reason why you (re-)implemented the logic to lookup the ivy report file? 

No, I think I simply didn't notice it I guess.

Like the commits here a lot, will look at how validation behaves not but changes seem valid to me, thanks!

ktoso:
Looks good (verified output), thanks @jrudolph!

PR #19063, opened 2015-12-01
!str #15089 add flatMapMerge

### LINKED ISSUE #15089: add flattening for streams: merge
Merge elements from a configurable maximum number of simultaneously open substreams, using a configurable merge strategy (see #15079). Fail as soon as a substream fails.

rkuhn:
fixes #15089


akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowFlattenMergeSpec.scala
```diff
@@ -0,0 +1,129 @@
...
+import org.scalactic.ConversionCheckedTripleEquals
+import akka.stream.testkit.TestPublisher
+import org.scalatest.exceptions.TestFailedException
+import akka.stream.testkit.scaladsl.TestSink
+
+class FlowFlattenMergeSpec extends AkkaSpec with ScalaFutures with ConversionCheckedTripleEquals {
+  implicit val mat = ActorMaterializer()
+  import system.dispatcher
+
+  def src(i: Int) = Source((10 * i) until (10 * i + 10))
```
drewhk: i would remove the multiplication. `Source(List(src(0), src(10), src(20), src(30)))` is easier for me to mentally relate to the actual ranges of 10.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowFlattenMergeSpec.scala
```diff
@@ -0,0 +1,129 @@
...
+  def blocked = Source(Promise[Int].future)
```
drewhk: you can use Source.maybe

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowFlattenMergeSpec.scala
```diff
@@ -0,0 +1,129 @@
...
+      val sink = Source(List(Source(p1), Source(p2)))
+        .flatMapMerge(5, identity)
+        .runWith(TestSink.probe)
+      sink.request(1)
+      p1.expectRequest()
+      p2.expectRequest()
+      sink.cancel()
+      p1.expectCancellation()
+      p2.expectCancellation()
+    }
```
ktoso: maybe add a test to see make sure `flatMapMerge(5, throw Boom)` is handled correctly?

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  override def createLogic(attr: Attributes) = new GraphStageLogic(shape) {
+
+    var sources = Vector.empty[LocalSource[T]]
+    var haveData = 0
+
+    val readyFilter = (src: LocalSource[T]) ⇒ src.elem != null
+    def pushOut(): Unit = {
+      val idx = ThreadLocalRandom.current.nextInt(haveData)
+      val src = sources.iterator.filter(readyFilter).drop(idx).next()
```
drewhk: hm. This is one option, another is to have a queue of ready sources.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowFlattenMergeSpec.scala
rkuhn: okay

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowFlattenMergeSpec.scala
rkuhn: yeah, but it doesn’t really make a difference

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowFlattenMergeSpec.scala
rkuhn: good point

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
rkuhn: So, do you have an opinion on which one it should be?

rkuhn: Hmm, trying out a fixed-size queue triggered the thought that we might want to gracefully deal with `breadth == Int.MaxValue`, which means not allocating an 8GB or 16GB array. OTOH dropping from that iterator also becomes O(n), so a queue is better. But it needs to be dynamic, so I’m implementing that now.

rkuhn:
it is ready now

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -0,0 +1,162 @@
...
+import java.{ util ⇒ ju }
+
+class FlattenMerge[T, M](breadth: Int) extends GraphStage[FlowShape[Source[T, M], T]] {
+  private val in = Inlet[Source[T, M]]("flatten.in")
+  private val out = Outlet[T]("flatten.out")
+  override val shape = FlowShape(in, out)
+
+  override def createLogic(attr: Attributes) = new GraphStageLogic(shape) {
+
+    var sources = Set.empty[LocalSource[T]]
```
viktorklang: Do not care if it is a TreeSet or HashSet? (also: LocalSource doesn't implement equals or hashCode so I assume you want identity based equality?)

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -0,0 +1,162 @@
...
+    def activeSources = sources.size
+
+    private trait Queue {
```
viktorklang: Put these in companion object?

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -0,0 +1,162 @@
...
+          tail += 1
+        }
+      def dequeue(): LocalSource[T] = {
+        val ret = queue(head & Mask)
+        head += 1
+        ret
+      }
+    }
+
+    private class DynamicQueue(queue: ju.Queue[LocalSource[T]]) extends Queue {
```
viktorklang: May as well extend ju.LinkedList to avoid the indirection (and It's only used in one place anyway)

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -0,0 +1,162 @@
...
+import akka.stream.scaladsl._
+import akka.stream.actor.ActorSubscriberMessage
+import akka.stream.actor.ActorSubscriberMessage._
+import akka.stream.actor.ActorPublisherMessage
+import akka.stream.actor.ActorPublisherMessage._
+import scala.concurrent.forkjoin.ThreadLocalRandom
+
+import java.{ util ⇒ ju }
+
+class FlattenMerge[T, M](breadth: Int) extends GraphStage[FlowShape[Source[T, M], T]] {
```
viktorklang: How much of this implementation could be "reused" for `interleave`?

rkuhn: Do we really want a dynamic interleave? I thought that was only for static topologies …

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
rkuhn: yes, identity it is

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
rkuhn: nope, needs to set a private field in the logic

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
rkuhn: good point

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
viktorklang: True, I don't have a good use case but it would indeed be symmetric to have one.

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -0,0 +1,173 @@
...
+    var sources = Set.empty[LocalSource[T]]
+    def activeSources = sources.size
+
+    private trait Queue {
+      def hasData: Boolean
+      def enqueue(src: LocalSource[T]): Unit
+      def dequeue(): LocalSource[T]
+    }
+
+    private class FixedQueue extends Queue {
```
drewhk: Nice approach. I had a similar idea about having a circular linked list of arrays as a reusable dynamic, but cache friendly queue. Not for this ticket though, but it seems to be useful in many places to be investigated further.

drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1009,10 +1016,23 @@ trait FlowOps[+Out, +Mat] {
...
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes and all consumed substreams complete
+   *
+   * '''Cancels when''' downstream cancels
    */
-  def flatMapConcat[T](f: Out ⇒ Source[T, _]): Repr[T, Mat] =
-    deprecatedAndThen(ConcatAll(f.asInstanceOf[Any ⇒ Source[Any, _]]))
+  def flatMapMerge[T, M](breadth: Int, f: Out ⇒ Source[T, M]): Repr[T, Mat] = map(f).via(new FlattenMerge[T, M](breadth))
```
ktoso: I guess you have a reason - but why don't we provide identity as a default for `f`?

ktoso:
LGTM, great :)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: Then you need to add an implicit evidence that Out <:< T

ktoso: Oh I see, thanks!

Issue #16991, opened 2015-03-04
Static file/resource serving should allow file types to specify encoding header


sirthias:
Currently the `FileAndResourceDirectives` automatically derive the response `Content-Type` from the registered file extensions.

However, there are pre-compressed file types like `.svgz`, `.htmlz`, `.js.gz` or `.css.gz` that, in addition to the `Content-Type` header, require the addition of a `Content-Encoding: gzip` header.

We should extend the static file/resource serving infrastructure to fully support such file types.

PR #19066, opened 2015-12-01
http: refactor MediaType / ContentType model and content negotiation

### LINKED ISSUE #19034: http: content negotiation doesn't properly handle certain edge cases
These tests currently fail when added to the respective specs:

`ContentNegotiationSpec`:

```scala
"Accept: text/*, application/json;q=0.8, text/plain;q=0.5" test { accept ⇒
  accept(`text/plain`, `application/json`) should select(`application/json`, `UTF-8`)
}

"Accept-Charset: UTF-8, *;q=0.8, us;q=0.1" test { accept ⇒
  accept(`text/plain` withCharset `US-ASCII`,
    `text/plain` withCharset `ISO-8859-1`) should select(`text/plain`, `ISO-8859-1`)
}
```

`CodingDirectiveSpec`:

```scala
Get("/") ~> `Accept-Encoding`(HttpEncodingRange.`*`, deflate withQValue 0.2) ~> {
  encodeResponseWith(Deflate, Gzip) { yeah }
} ~> check {
  response should haveContentEncoding(gzip)
  strictify(responseEntity) shouldEqual HttpEntity(ContentType(`text/plain`, `UTF-8`), yeahGzipped)
}
```

I'm going to submit a fix soon.

/cc @jrudolph @2beaucoup

sirthias:
Closes #19034.

From an initially seemingly simple fix this thing has turned into quite a major refactoring, cleaning up some of the weaker parts of the HTTP model. The core problem was the `ContentType` type, which has quite a central role but whose exact semantics had become more and more blurry over time, as it was used for more and more related, but not quite identical, things.

This PR completely revamps the `MediaType` and `ContentType` design, moving runtime checks into the type system and providing a more structured and safer API.

Unfortunately, because these two types are quite core, this patch is now somewhat big and probably a little harder to review. Let me know if there is something that I can do to ease the process.

/cc @jrudolph


ktoso:
Sounds very exciting from the description already – I'll dive right into it.
(I myself was not always sure about MediaType / ContentType use in code, so we'll see how I feel now :wink: )

akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntities.java
```diff
@@ -28,8 +28,8 @@ public static HttpEntityStrict create(ByteString bytes) {
         return HttpEntity$.MODULE$.apply(bytes);
     }
 
-    public static HttpEntityStrict create(ContentType contentType, String string) {
-        return HttpEntity$.MODULE$.apply((akka.http.scaladsl.model.ContentType) contentType, string);
+    public static HttpEntityStrict create(ContentType.NonBinary contentType, String string) {
```
ktoso: oh nice idea :+1:

akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
```diff
@@ -40,48 +40,48 @@
...
      * Returns the content-type of this entity
      */
-    public abstract ContentType contentType();
+    ContentType contentType();
 
     /**
      * The empty entity.
      */
-    public static final HttpEntityStrict EMPTY = HttpEntity$.MODULE$.Empty();
+    HttpEntityStrict EMPTY = HttpEntity$.MODULE$.Empty();
```
ktoso: sure about dropping the static?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpEncoding.scala
```diff
@@ -22,7 +22,7 @@ object HttpEncodingRange {
   case class `*`(qValue: Float) extends HttpEncodingRange {
     require(0.0f <= qValue && qValue <= 1.0f, "qValue must be >= 0 and <= 1.0")
     final def render[R <: Rendering](r: R): r.type = if (qValue < 1.0f) r ~~ "*;q=" ~~ qValue else r ~~ '*'
-    def matches(encoding: HttpEncoding) = qValue > 0f
+    def matches(encoding: HttpEncoding) = true
```
ktoso: :+1:

akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
```diff
@@ -31,74 +34,82 @@ object ContentTypeRange {
...
-  val `text/plain` = ContentType(MediaTypes.`text/plain`)
-  val `text/plain(UTF-8)` = ContentType(MediaTypes.`text/plain`, HttpCharsets.`UTF-8`)
-  val `text/html` = ContentType(MediaTypes.`text/html`)
-  val `text/xml` = ContentType(MediaTypes.`text/xml`)
-
-  val `application/x-www-form-urlencoded` = ContentType(MediaTypes.`application/x-www-form-urlencoded`)
-  val `multipart/form-data` = ContentType(MediaTypes.`multipart/form-data`)
+  val `text/plain(UTF-8)` = MediaTypes.`text/plain` withCharset HttpCharsets.`UTF-8`
+  val `text/html(UTF-8)` = MediaTypes.`text/html` withCharset HttpCharsets.`UTF-8`
+  val `text/xml(UTF-8)` = MediaTypes.`text/xml` withCharset HttpCharsets.`UTF-8`
```
ktoso: what about json above here?

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
```diff
@@ -884,6 +884,7 @@ class HttpServerSpec extends AkkaSpec("akka.loggers = []\n akka.loglevel = OFF")
       "the config setting applied before another attribute (default entity)" in new LengthVerificationTest(maxContentLength = 10) {
         def nameDataSource(name: String): RequestEntity ⇒ RequestEntity = {
           case x: HttpEntity.Default ⇒ x.copy(data = x.data named name)
+          case _                     ⇒ ???
```
ktoso: to get rid of the warning I guess? ok :wink:

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
```diff
@@ -5,272 +5,253 @@
...
-  def isAudio = false
-  def isImage = false
-  def isMessage = false
-  def isMultipart = false
-  def isText = false
-  def isVideo = false
+  def custom(value: String, binary: Boolean, compressible: Boolean = true,
+             fileExtensions: List[String] = Nil): MediaType = {
+    val parts = value.split('/')
+    if (parts.length != 2) throw new IllegalArgumentException(value + " is not a valid media-type")
```
ktoso: Can be a bit more helpful here: 

`" is not a valid media-type. Should be 2 `/` separated parts."` (open for better wording ideas here hm)

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
```diff
@@ -5,272 +5,253 @@
...
+ * is not generally the case for non-binary (i.e. character-based) content, which also requires the definition
+ * of a specific character encoding ([[HttpCharset]]).
+ * Therefore [[MediaType]] instances are frequently encountered as a member of a [[ContentType]], which
+ * groups a [[MediaType]] with a potentially required [[HttpCharset]] to hold everything required for being
+ * able to interpret an [[HttpEntity]].
+ *
+ * MediaTypes come in three basic forms:
+ *
+ * 1. Binary: These do not need an additional [[HttpCharset]] to be able to form a [[ContentType]]. Therefore
+ *    they can be implicitly converted to the latter.
```
ktoso: Great explanation and makes a lot of sense, thanks.

akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
ktoso: Okey, I see – so now, ContentType, which is not binary, MUST have a charset. Sounds right :+1:

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaRange.scala
```diff
@@ -0,0 +1,163 @@
...
+  }
+  val `text/*` = new PredefinedMediaRange("text/*") {
+    def matches(mediaType: MediaType) = mediaType.isText
+    override def isText = true
+    def specimen = MediaTypes.`text/plain`
+  }
+  val `video/*` = new PredefinedMediaRange("video/*") {
+    def matches(mediaType: MediaType) = mediaType.isVideo
+    override def isVideo = true
+    def specimen = MediaTypes.`video/mp4`
```
ktoso: just for my reference: how do we pick a specimen – any will do?

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
```diff
@@ -410,40 +366,34 @@ object MediaTypes extends ObjectRegistry[(String, String), MediaType] {
...
+  val `image/x-pict`      = img("x-pict", compressible, "pct")
+  val `image/x-quicktime` = img("x-quicktime", uncompressible, "qif", "qti", "qtif")
+  val `image/x-rgb`       = img("x-rgb", compressible, "rgb")
+  val `image/x-xbitmap`   = img("x-xbitmap", compressible, "xbm")
+  val `image/x-xpixmap`   = img("x-xpixmap", compressible, "xpm")
+  val `image/webp`        = img("webp", uncompressible, "webp")
 
   val `message/http`            = msg("http")
   val `message/delivery-status` = msg("delivery-status")
   val `message/rfc822`          = msg("rfc822", "eml", "mht", "mhtml", "mime")
```
ktoso: Looked through the changes here for typos, seems OK.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpEncoding.scala
```diff
@@ -55,7 +55,6 @@ object HttpEncodings extends ObjectRegistry[String, HttpEncoding] {
   val deflate          = register("deflate")
   val gzip             = register("gzip")
   val identity         = register("identity")
-  val `identity;q=MIN` = identity.withQValue(Float.MinPositiveValue)
```
ktoso: OK, so `identity;q=MIN` is gone completely. The most general thing being identity now, right?

akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
ktoso: Thought – so for making requests we certainly will want to specify UTF-8 anyway always, but when I'm using the low-level API, and pattern matching on a response... wouldn't there `text/xml` be useful, in case I really don't care what encoding it is in ("bah, ok windows-1250. but I just pipe it to a file anyway").?

akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
sirthias: Nope, that's wrong.
Good catch!

sirthias: Ah, wait. `HttpEntity` is an `interface`, therefore the `static` is redundant, am I wrong?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpEncoding.scala
sirthias: This `identity;q=MIN` thing was only a "convenience" thing anyway, with almost zero real-life usage, however.
`gzip` and `deflate` are the only things that really matter.

akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
sirthias: If you want `text/xml` without a specified charset, then the `MediaType` is the thing:

```scala
MediaTypes.`text/xml`
```

A `ContentType` must contain everything required for being able to interpret raw bytes (things like compression and transfer-encoding notwithstanding).

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerSpec.scala
sirthias: Yes, I'll add a comment to make this clear.

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
sirthias: Yes, good idea!

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaRange.scala
sirthias: Yes, any will do.
Bascially, if someone says they accept `video/*` they should also accept `video/mp4`.
Very rarely used here, I'm sure, but required for completeness.

akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
hepin1989: @sirthias yes,`public static` is not needed in java interface.

sirthias: Thanks for confirming, @hepin1989, I must admit that I am becoming more and more unsure of myself once dropped down to java-land...

hepin1989: I must say I hate java code cause it don't have defined side variance,and I know your pain,cause we have a hyper production codebase too:(

ktoso: Hah, right – thanks guys. It's been a while since I've put a constant in a Java interface. ;-)

akka-http-core/src/main/scala/akka/http/scaladsl/model/ContentType.scala
ktoso: Got it, thanks

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaRange.scala
ktoso: Okey, thanks

sirthias:
Ok, applied all feedback so far.
Thanks again, @2beaucoup, for some excellent improvements ideas!

sirthias:
Updated with all feedback so far.

ktoso:
I skimmed the PR again and looks good.
If there's some detail missing we'll likely find out through user feedback anyway.

LGTM –  I'd merge unless someone wants to review in depth again?

2beaucoup:
2 final cents from my side:

As we want to make it hard to implement homegrown content negotiation I'd vote for making it private.

I (still) don't fully grasp the `specimen` stuff. Perhaps @sirthias could provide some quick hints why we need it?

Don't have the time for a complete review right now, but the content negotiation stuff (last commit) looks :+1:.

sirthias:
@ktoso Hold off a little bit more with merging, I'm going to look into the specimen stuff once more.
Maybe we can get rid of it completely by restructuring a few things somewhere else.
I agree with @2beaucoup in that is does have a certain smell to it...

Also, I think that André's point, that we could make everything around Content-Negotiation private until someone presents a valid argument for opening it up, is convincing. I'll do that change as well...

2beaucoup:
Excellent Mathias!

ktoso:
Great, thanks so much guys :-)
The dude abides.

akka-http/src/main/scala/akka/http/scaladsl/server/ContentNegotation.scala
```diff
@@ -0,0 +1,265 @@
...
+   * If there are several best encoding alternatives that the client has equal preference for
+   * the order of the arguments is used as a tie breaker (first one wins).
+   *
+   * If none of the given alternatives is acceptable to the client the methods return `None`.
+   */
+  def pickEncoding(alternatives: List[HttpEncoding]): Option[HttpEncoding] =
+    if (alternatives.nonEmpty) {
+      val (bestAlternative, bestQVneg) = alternatives.map(l ⇒ l → -qValueFor(l)).sortBy(_._2).head
+      if (bestQVneg != 0f) Some(bestAlternative) else None
+    } else None
```
2beaucoup: `if`-less version:
```scala
def pickEncoding(alternatives: List[HttpEncoding]): Option[HttpEncoding] =
    alternatives
      .map(l ⇒ l → qValueFor(l))
      .sortBy(-_._2)
      .find(_._2 > 0f)
      .map(_._1)
```

2beaucoup: even shorter:
```scala
.map(l ⇒ l → qValueFor(l))
.sortBy(-_._2)
.collectFirst { case (l, q) if q > 0f ⇒ l }
```

akka-http/src/main/scala/akka/http/scaladsl/server/ContentNegotation.scala
```diff
@@ -0,0 +1,265 @@
...
+  def pickContentType(alternatives: List[Alternative]): Option[ContentType] =
+    if (alternatives.nonEmpty) {
+      val (bestAlternative, bestQVneg) = alternatives.map(alt ⇒ (alt, -qValueFor(alt))).sortBy(_._2).head
+      if (bestQVneg != 0f) {
+        bestAlternative match {
+          case Alternative.ContentType(ct) ⇒ Some(ct)
+          case Alternative.MediaType(mt)   ⇒ csn.pickBest.map(mt.withCharset)
+        }
+      } else None
+    } else None
```
2beaucoup: would work here too:
```scala
def pickContentType(alternatives: List[Alternative]): Option[ContentType] =
    alternatives
      .map(l ⇒ l → qValueFor(alt))
      .sortBy(-_._2)
      .collectFirst { case (l, q) if q > 0f => l}
      .flatMap {
        case Alternative.ContentType(ct) ⇒ Some(ct)
        case Alternative.MediaType(mt)   ⇒ csn.pickBest.map(mt.withCharset)
      }
```

akka-http/src/main/scala/akka/http/scaladsl/server/ContentNegotation.scala
sirthias: Yes, that's better. I tried to reduce the number of intermediate results and function instances created but your short version is much more readable.

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
```diff
@@ -280,120 +261,95 @@ object MediaTypes extends ObjectRegistry[(String, String), MediaType] {
...
+  val `application/x-compress`                                                    = abin("x-compress", uncompressible, "z")
+  val `application/x-compressed`                                                  = abin("x-compressed", uncompressible, "gz")
+  val `application/x-debian-package`                                              = abin("x-debian-package", compressible, "deb")
+  val `application/x-dvi`                                                         = abin("x-dvi", compressible, "dvi")
+  val `application/x-font-truetype`                                               = abin("x-font-truetype", compressible, "ttf")
+  val `application/x-font-opentype`                                               = abin("x-font-opentype", compressible, "otf")
+  val `application/x-gtar`                                                        = abin("x-gtar", uncompressible, "gtar")
+  val `application/x-gzip`                                                        = abin("x-gzip", uncompressible, "gzip")
+  val `application/x-latex`                                                       = abin("x-latex", compressible, "latex", "ltx")
```
analytically: Latex consists of plain text and do not contain hidden formatting codes or binary instructions, so this should be awoc?

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
```diff
@@ -280,120 +261,95 @@ object MediaTypes extends ObjectRegistry[(String, String), MediaType] {
...
-  val `application/x-x509-ca-cert`                                                = app("x-x509-ca-cert", compressible, binary, "der")
-  val `application/x-xpinstall`                                                   = app("x-xpinstall", uncompressible, binary, "xpi")
-  val `application/xhtml+xml`                                                     = app("xhtml+xml", compressible, openEncoding)
-  val `application/xml-dtd`                                                       = app("xml-dtd", compressible, openEncoding)
-  val `application/xml`                                                           = app("xml", compressible, openEncoding)
-  val `application/zip`                                                           = app("zip", uncompressible, binary, "zip")
+  private[http] val NoMediaType = MediaType.customBinary("none", "none", compressible = false)
+
+  val `application/atom+xml`                                                      = awoc("atom+xml", "atom")
+  val `application/base64`                                                        = abin("base64", compressible, "mm", "mme")
```
analytically: base64 considered binary?

2beaucoup:
Should we add the media types mentioned in #16991?

akka-http/src/main/scala/akka/http/scaladsl/server/ContentNegotation.scala
```diff
@@ -0,0 +1,265 @@
...
+      case x ⇒ mtn.qValueFor(x.mediaType)
+    }
+
+  /**
+   * Picks the best of the given content alternatives given the preferences
+   * the client indicated in the request's `Accept` and `Accept-Charset` headers.
+   * See http://tools.ietf.org/html/rfc7231#section-5.3.2 ff for details on the negotiation logic.
+   *
+   * If there are several best alternatives that the client has equal preference for
+   * the order of the arguments is used as a tie breaker (first one wins).
```
2beaucoup: the order of `alternatives`?

akka-http/src/main/scala/akka/http/scaladsl/server/ContentNegotation.scala
```diff
@@ -0,0 +1,265 @@
...
+  def hasMatchingFor(encoding: HttpEncoding): Boolean =
+    acceptedEncodingRanges.exists(_ matches encoding)
+
+  /**
+   * Picks the best of the given encoding alternatives given the preferences
+   * the client indicated in the request's `Accept-Encoding` headers.
+   * See http://tools.ietf.org/html/rfc7231#section-5.3.4 for details on the negotiation logic.
+   *
+   * If there are several best encoding alternatives that the client has equal preference for
+   * the order of the arguments is used as a tie breaker (first one wins).
```
2beaucoup: Looks like the `sortBy` below keeps the order of `alternatives` ISO `acceptedEncodingRanges`. Is this tested?

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
sirthias: Yes, right. Should be `awoc`. Thanks for looking through this here, Mathias!

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
sirthias: Nope, should be `awoc` too.
Thanks!

akka-http/src/main/scala/akka/http/scaladsl/server/ContentNegotation.scala
sirthias: "the order of the given alternatives" is better, yes. Thx!

akka-http/src/main/scala/akka/http/scaladsl/server/ContentNegotation.scala
sirthias: The `sortBy` in the `pick...` methods doesn't change the order of the alternatives if they are mapped to the same qValue, yes. This is what we want here. And it's also tested.

sirthias:
@2beaucoup Good idea, to tackle #16991 right in the context of this PR. Makes sense. I'll quickly add a commit.

akka-http/src/main/scala/akka/http/scaladsl/server/ContentNegotation.scala
2beaucoup: Strike that. I misinterpreted this statement as "`Accept` header order wins". Maybe replace "arguments" with "aletrnatives" here too to make it more clear.

So the stable sort in `acceptedXXX` doesn't really matter, right?

2beaucoup:
Wow, this salami tactic (aka "kleiner Finger, ganze Hand") thing really works! ;)

akka-http/src/main/scala/akka/http/scaladsl/server/ContentNegotation.scala
sirthias: > So the stable sort in acceptedXXX doesn't really matter, right?

No any more it doesn't, no.
In the first version I followed the spec by the book, which says that the order of the ranges in the Accept* headers should break ties. However, if we implement it this way then the application developer has no way to establish her own precedence, which is something I'd value higher than strictly following the spec in this matter.

So, if the client says he'll accept both `deflate` and `gzip` with equal preference then we can decide what to pick via our alternative ordering. If the client really wants `gzip` rather than `deflate` he needs to signal that preference with q-values.

sirthias:
:)
It does here because #16991 requires breaking the API once more, which now is a good time for (and later probably isn't). And also: I need svgz support myself.... :)

akka-http/src/main/scala/akka/http/scaladsl/server/ContentNegotation.scala
2beaucoup: This logic seems to not contradict the spec so :+1:.

sirthias:
Ok, applied all feedback so far and added a commit closing #16991.

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
```diff
@@ -91,6 +92,30 @@ class FileAndResourceDirectivesSpec extends RoutingSpec with Inspectors with Ins
         }
       } finally file.delete
     }
+
+    "support precompressed files with registered MediaType" in {
+      val file = File.createTempFile("akkaHttpTest", ".svgz")
+      try {
+        writeAllText("123", file)
+        Get() ~> getFromFile(file) ~> check {
+          mediaType shouldEqual `image/svgz`
```
2beaucoup: This looks redundant. Souldn't it rather be `image/svg`?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -186,6 +181,11 @@ trait FileAndResourceDirectives {
 }
 
 object FileAndResourceDirectives extends FileAndResourceDirectives {
+  private val withRangeSupportAndPrecompressedMediaTypeSupportAndExtractSettings =
```
2beaucoup: As this is internal why not simply call it `wrap`?

akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectivesSpec.scala
sirthias: mediaType shouldEqual `image/svg` will also match as they both have the same value string "image/svg".
But I agree that

```scala
mediaType shouldEqual `image/svg`
```

would make the intent a bit clearer.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
sirthias: Because with `wrap` I'd have to look at the source to see what exactly it's doing...

akka-http/src/main/scala/akka/http/scaladsl/server/directives/CodingDirectives.scala
```diff
@@ -117,37 +128,21 @@ object CodingDirectives extends CodingDirectives {
...
-            case _ ⇒ None
-          }
-          find(withIdentity.sortBy(e ⇒ (-e.qValue, withIdentity.indexOf(e))).toList)
-      }
-      acceptedEncoder match {
+    BasicDirectives.extractRequest.flatMap { request ⇒
+      val negotiator = EncodingNegotiator(request.headers)
+      val encodings: List[HttpEncoding] = encoders.map(_.encoding)(collection.breakOut)
+      val bestEncoder =
+        if (encodings.nonEmpty)
```
2beaucoup: Unnecessary?

sirthias: Yes.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
2beaucoup: I *had* to look at the source to see what it's doing. ;)

akka-http/src/main/scala/akka/http/scaladsl/server/directives/CodingDirectives.scala
```diff
@@ -117,37 +128,21 @@ object CodingDirectives extends CodingDirectives {
...
+          negotiator.pickEncoding(encodings).flatMap(be ⇒ encoders.find(_.encoding == be))
+        else None
+      bestEncoder match {
         case Some(encoder) ⇒ mapResponse(encoder.encode(_))
-        case _             ⇒ reject(UnacceptedResponseEncodingRejection(encoders.map(_.encoding).toSet))
+        case _ ⇒
+          if (encoders.contains(NoCoding) && !negotiator.hasMatchingFor(HttpEncodings.identity)) pass
```
2beaucoup: This checks if identity wasn't explicitly forbidden (q=0), right?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
sirthias: You *always* look at the source to see what it's doing!
But now, for this method, you'll never have to again because the name is so descriptive... :)

akka-http/src/main/scala/akka/http/scaladsl/server/directives/CodingDirectives.scala
sirthias: Yes, exactly.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
2beaucoup: Yes, in fact this is a hot candidate for most descriptive method name, ever! :)

sirthias:
So, updated once more.
Can we please get this thing merged before @2beaucoup finds even more well-grounded improvement potential? ;-)

akka-http/src/main/scala/akka/http/scaladsl/server/directives/CodingDirectives.scala
```diff
@@ -107,6 +108,16 @@ trait CodingDirectives {
    */
   def decodeRequest: Directive0 =
     decodeRequestWith(DefaultCoders: _*)
+
+  /**
+   * Inspects the response entity and adds a `Content-Encoding: gzip` response header if
+   * the entities media-type is precompressed with gzip and no `Content-Encoding` header is present yet.
```
2beaucoup: `Content-Encoding` is stackable. I think we could just add `gzip`anyway.

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
```diff
@@ -5,445 +5,411 @@
...
-  /**
-   * Tries to parse a ``MediaType`` value from the given String. Returns ``Right(mediaType)`` if successful and
-   * ``Left(errors)`` otherwise.
-   */
-  def parse(value: String): Either[List[ErrorInfo], MediaType] =
-    ContentType.parse(value).right.map(_.mediaType)
+  sealed abstract class Compressibility(val compressible: Boolean)
+  case object Compressible extends Compressibility(compressible = true)
+  case object NotCompressible extends Compressibility(compressible = false)
+  case object Gzipped extends Compressibility(compressible = false)
```
ktoso: `Gzipped` or `Compressed`? What if I roll my own compression (like our coding.Gzip)

akka-http/src/main/scala/akka/http/scaladsl/server/directives/CodingDirectives.scala
sirthias: Yes, it's stackable. But the order matters. Since someone has already added a `Content-Encoding` header, do we now add `gzip` to the front or the back?
I'd say, if someone added a `Content-Encoding` we simply assume she knows what she is doing and bail out...

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
sirthias: No, is not "generally compressed" here, it's "compressed with gzip".
The reason is that we want to automatically add a `Content-Encoding: gzip` header, when we serve a file with a matching file extension.

ktoso: Right, I see.

I guess when rolling my own MediaType (and it's compressed using something) I could use `NotCompressible`, but somehow feels that this could be useful to allow users to extend (i.e. drop the `sealed` off `Compressability`) - let's say I want to add my special content-encoding, in a similar way like we do with the built-in gzip.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/CodingDirectives.scala
2beaucoup: Ok. I once experimented with an impl that for nested coding directives peeled off `Content-Encoding` values while decoding, and stacked them for encoding: https://github.com/2beaucoup/akka/commit/b32aeb5ebcc4b433bc8803b0dbcd921da078651b

The current impl just replaces the whole header so this looks ok.

You're off the hook now. ;)

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
sirthias: You can do whatever you want and simply use `NotCompressible`.
You just won't get any special support with the FileAndResourceDirectives adding certain headers for your media-type.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/CodingDirectives.scala
sirthias: > You're off the hook now. ;)

Phew... :)

akka-http-core/src/main/scala/akka/http/scaladsl/model/MediaType.scala
ktoso: Yeah, I could inspect the ContentType as a whole and add my headers... got it. Thanks!

ktoso:
> Can we please get this thing merged before @2beaucoup finds even more well-grounded improvement potential? ;-)

Yeah, I know, right? It's _horrible_! :wink: (joking of course, huge thanks for all the feedback @2beaucoup - you're awesome).

I'll merge once validation passes then, thanks!

2beaucoup:
:+1: Nice work @sirthias!

ktoso:
Merged :fireworks:

sirthias:
Great, thanks, Konrad!

PR #19078, opened 2015-12-03
+htc,tes #18563 move scripts autobahn jenkins scripts to project

### LINKED ISSUE #18563: Include Autobahn WS tests in the build

ktoso:
This resolves #18563 by moving the scripts that make it work, to the repo itself, so it's easier to maintain (than in jenkins in the script windows).

### DIFF
--- a/akka-http-core/src/test/scripts/autobahn-jenkins/step-1-client-tests.sh
+++ b/akka-http-core/src/test/scripts/autobahn-jenkins/step-1-client-tests.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+### CLIENT TESTS ###
+
+CLIENT_CID=autobahn-client.cid
+
+echo "~~~ Docker state before running CLIENT tests ~~~"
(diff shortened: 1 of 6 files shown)


ktoso:
The build currently works on HEAD `release-2.3-dev`, so I'll merge this to make sure it's "really done" (i.e. the calling of the scripts works), but please review and if something is weird I can change in another PR.

PR #19087, opened 2015-12-04
+ act #17895 bump the arity of debug,info,error,warning up to 10

He-Pin:
cause the first attack by https://github.com/akka/akka/pull/18803 is closed by BC,and I want the feature really,so I open up this one.

only bump the arity up to 10,which is OK I think,so not bump that to more like 22.the more one could comes in later major release.

I think this will not cause any binary compatibility issue,not using the sbt-boilerplate here.

### DIFF
--- a/akka-actor/src/main/scala/akka/event/Logging.scala
+++ b/akka-actor/src/main/scala/akka/event/Logging.scala
@@ -924,6 +924,36 @@ trait LoggingAdapter {
    * @see [[LoggingAdapter]]
    */
   def error(cause: Throwable, template: String, arg1: Any, arg2: Any, arg3: Any, arg4: Any): Unit = { if (isErrorEnabled) notifyError(cause, format(template, arg1, arg2, arg3, arg4)) }
+  /**
+   * Message template with 5 replacement arguments.
+   * @see [[LoggingAdapter]]
+   */
+  def error(cause: Throwable, template: String, arg1: Any, arg2: Any, arg3: Any, arg4: Any, arg5: Any): Unit = { if (isErrorEnabled) notifyError(cause, format(template, arg1, arg2, arg3, arg4, arg5)) }
+  /**
+   * Message template with 6 replacement arguments.
+   * @see [[LoggingAdapter]]
+   */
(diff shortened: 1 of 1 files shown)


hepin1989:
@kukido Are you OK with this ?I think if this could be merged in,then you could continue you work to support more than 10 arities one for the next major release.

ktoso:
OK TO TEST

ktoso:
I can see the need for it and it should be binary compatible this way hm... Let's sanity check with MiMa (PR validation does so)

ktoso:
Sadly [computer says no](https://www.youtube.com/watch?v=AJQ3TM-p2QI)

so... the scenario would be, AFAICS:
X extended trait LoggingAdapter {
this translates to implementing the LoggingAdapter interface
and bridge methods being generated
new akka comes in, has more methods in LoggingAdapter
X's class "implements LoggingAdapter", yet does not have the bridges generated
fails, because it did not implement all methods from that interface
-> if X recompiled it would work; but that's not binary compat anymore

hepin1989:
close it for I think I can't make it binary compatibility without wrapper and  implicit conversion,and on the java side will looks very ugly too,even on the scala side,it will be wired too.
this could be implemented with another new `ActorLogging` which return a custom implement of LoggingAdapter,but that will be another story,

ktoso:
Indeed, and thanks for the continued research into this!
I agree it's a nice thing to improve but will have to wait until we break comat

hepin1989:
@ktoso yes ,I implement it that way locally,thanks for point out the BC,I am new at this area.

kukido:
@hepin1989 if you look into history of my pull request, there will be all iterations for the process. Including the last one, with implicit conversion for Scala and wrapper class for Java, they just don't look right. If you are coding in Scala, why don't you just create your own implicit class with the required methods, before the changes are made to the codebase?

https://github.com/kukido/akka/commit/dc2d7659990c591b86b3f3b740e2bc1e040fcae7

hepin1989:
@kukido yes,I think we both go into the dead end.
I will just create a default implement locally:)
and the real problem is we have a hyper code base,anyway ,I think an implement of the LoggingAdapter will be a better way to use it locally.
BC is dog:),thank you :+1:

hepin1989:
and I have to admire that I have not look much into the BC before,sorry.

hepin1989:
currently use` ActorLoggingExt` and `LoggingAdapterExt` locally:)

PR #19088, opened 2015-12-04
+act #19085 adds Java API ByteString#createBuilder

### LINKED ISSUE #19085: Provide API for Java users to create ByteStringBuilders.
```java
ByteStringBuilder builder = ByteString.newBuilder();
```

doesn't compile:

```
[error] /home/jroper/src/reactive-services-platform/jackson/src/main/java/com/typesafe/service/javadsl/jackson/JacksonExceptionSerializer.java:48: non-static method newBuilder() cannot be referenced from a static context
[error] ByteString.newBuilder
```

because the `ByteString` class extends `scala.collection.generic.HasNewBuilder`, which already defines the `newBuilder` method, so the `newBuilder` on `ByteString$` can't be duplicated on the `ByteString` class as a static method.  So, currently the only way for Java users to create a `ByteStringBuilder` is on the `ByteString$` module instance:

```java
ByteStringBuilder builder = ByteString$.MODULE$.newBuilder();
```

ktoso:
May be good enough to fix https://github.com/akka/akka/issues/19085 actually

// cc @jroper @patriknw


patriknw:
LGTM, thanks

Issue #19064, opened 2015-12-01
Framing.lengthField with LITTLE_ENDIAN byte order issue


maciej:
`Framing.lengthField` fails to a stream frame when using a 4 byte `LITTLE_ENDIAN` length field (or I've got something completely wrong regarding endianness!) resulting in 4 byte frames produced. It works well when using `ByteOrder.BIG_ENDIAN`, though.

Here is the link to the example code and its output: https://gist.github.com/maciej/310e574a62f4d2b80fef

I'm using Akka 2.4.1, Streams 2.0-M2, Oracle JDK 8u60, OSX 10.11.1.

ktoso:
Hej Maciej, thanks for reporting; we'll have a look.

ktoso:
Fixed, thanks @jrudolph !

maciej:
:+1: thanks @jrudolph!

PR #19090, opened 2015-12-04
=str #19064 fix little endian conversion for field length == 4

jrudolph:


jrudolph:
Fixes #19064.

jrudolph:
Had the same error in the implementation and test :)

Issue #18988, opened 2015-11-22
Flaky test in AcknowledgeSinkSpec


agolubev:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/4276/consoleFull
also saw this locally
```
[info] An AcknowledgeSink
[info] - must send the elements as result of future *** FAILED *** (3 seconds, 29 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsg while waiting for None
[info]   at scala.Predef$.assert(Predef.scala:179)
[info]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:338)
[info]   at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:315)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:718)
[info]   at akka.stream.scaladsl.AcknowledgeSinkSpec$$anonfun$1$$anonfun$apply$mcV$sp$1$$anonfun$apply$mcV$sp$12.apply(AcknowledgeSinkSpec.scala:35)
[info]   at akka.stream.scaladsl.AcknowledgeSinkSpec$$anonfun$1$$anonfun$apply$mcV$sp$1$$anonfun$apply$mcV$sp$12.apply(AcknowledgeSinkSpec.scala:29)
[info]   at akka.stream.testkit.Utils$.assertAllStagesStopped(Utils.scala:26)
[info]   at akka.stream.scaladsl.AcknowledgeSinkSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply$mcV$sp(AcknowledgeSinkSpec.scala:29)
[info]   at akka.stream.scaladsl.AcknowledgeSinkSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply(AcknowledgeSinkSpec.scala:29)
[info]   at akka.stream.scaladsl.AcknowledgeSinkSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply(AcknowledgeSinkSpec.scala:29)
```

ktoso:
Thanks for reporting

johanandren:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/4397/consoleFull

johanandren:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/4426/consoleFull

Issue #18821, opened 2015-10-31
Fix Sink.queue termination


rkuhn:
When the stream completes while no request is pending the next request will just time out because the Actor is already dead.

While switching from and ask-model to a proper atomic queue this Sink should also be rewritten as a GraphStage.

agolubev:
working on it

ktoso:
Thanks!

PR #19097, opened 2015-12-05
=str #18821 fix Sink.queue termination

agolubev:
Sink.queue rewritten with GraphStage


agolubev:
Ref #18821

ktoso:
Thanks!

ktoso:
OK TO TEST

rkuhn:
fixes #18988

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -247,3 +224,78 @@ private[akka] final class HeadOptionStage[T] extends SinkStage[T, Future[Option[
     }, p.future)
   }
 }
+
+/**
+ * INTERNAL API
+ */
+private[akka] class QueueSink[T]() extends SinkStage[T, SinkQueue[T]]("queueSink") {
```
rkuhn: This Sink is currently not reusable (i.e. if it is included in a graph that is included twice in a larger graph then things will break) due to state being kept inside the GraphStage—all state must be in the GraphStageLogic instead.

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -247,3 +224,78 @@ private[akka] final class HeadOptionStage[T] extends SinkStage[T, Future[Option[
...
+  trait RequestElementCallback {
+    def requestElement(p: Promise[Option[T]]): Unit
+  }
+
+  type Received[E] = Try[Option[E]]
+
+  val maxBuffer = module.attributes.getAttribute(classOf[InputBuffer], InputBuffer(16, 16)).max
+  require(maxBuffer > 0, "Buffer size must be greater than 0")
+  val requestInProgress = new AtomicBoolean()
+  val buffer = new LinkedBlockingQueue[Received[T]](maxBuffer + 1)
```
rkuhn: Please do not use blocking as a means of conveying back-pressure: wrapping `take` in a Future is not a nice thing to do.

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -247,3 +224,78 @@ private[akka] final class HeadOptionStage[T] extends SinkStage[T, Future[Option[
...
+  val logicInitialized = new CountDownLatch(1)
```
rkuhn: Similar here: start-up needs to be fully async, but not blocking.

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -247,3 +224,78 @@ private[akka] final class HeadOptionStage[T] extends SinkStage[T, Future[Option[
...
+
+    val stageLogic = new GraphStageLogic(shape) with RequestElementCallback {
+      override def keepGoingAfterAllPortsClosed = true
+      override def requestElement(p: Promise[Option[T]]): Unit = {
+        implicit val ex = interpreter.materializer.executionContext
+        Future(buffer.take()).onComplete {
+          case Success(opt) ⇒ callback.invoke((opt, p))
+          case Failure(t)   ⇒ p.failure(t) //timeout
+        }
+      }
```
rkuhn: Instead of this mix of different executions within the GraphStageLogic I’d propose to separate things out:

* have an external `val getter = Promise[(Promise[T] => Unit)]` that is fulfilled with an AsyncCallback from `preStart`
* `queue.pull()` then creates a `Promise[T]`, puts it into `getter.foreach(_(p))(sameThreadExecutionContext)` and returns the associated Future

agolubev:
@rkuhn, thanks for the feedback. I think I managed to make changes during weekend.

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
agolubev: yeah, that smells bad. Will fix it.

agolubev:
Build failed with: Merge conflict in akka-stream-tests/src/test/scala/akka/stream/scaladsl/QueueSinkSpec.scala

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -247,3 +222,86 @@ private[akka] final class HeadOptionStage[T] extends SinkStage[T, Future[Option[
...
+
+      def sendDownstream(promise: Requested[T]): Unit =
+        buffer.dequeue() match {
+          case Success(opt) ⇒ opt match {
+            case Some(e) ⇒ //emit element
+              promise.success(Some(e))
+            case None ⇒ //completed
+              promise.success(None)
+              completeStage()
+          }
```
rkuhn: how about
~~~scala
case Success(s: Some[_]) => promise.success(s)
case Success(None) => promise.success(None); completeStage()
~~~

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -247,3 +222,86 @@ private[akka] final class HeadOptionStage[T] extends SinkStage[T, Future[Option[
...
+            failStage(t)
+        }
+
+      def enqueueAndNotify(requested: Received[T]): Unit = {
+        buffer.enqueue(requested)
+        currentRequest match {
+          case Some(p) ⇒ sendDownstream(p)
+          case None    ⇒ //do nothing
+        }
+        currentRequest = None
```
rkuhn: It’s clearer (and a tiny bit faster) to move this write into the Some-case above.

rkuhn:
Please rebase as part of the two cosmetic fixes and then we can merge. Thanks again!

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
viktorklang: Or avoid to re-allocate the Try and do:

~~~scala
val e = buffer.dequeue()
promise.complete(e)
e match {
  case Success(None) => completeStage()
  case Failure(t) => failStage(t)
}

drewhk:
Needs a rebase, too.

agolubev:
hoping to see strange test fail on build server as well

agolubev:
build failed for strange reason I saw locally as well:
QueueSinkSpec:
[info] - must allow to have only one future waiting for result in each point of time *** FAILED *** (332 milliseconds)
[info]   Expected exception java.lang.IllegalStateException to be thrown, but java.util.concurrent.TimeoutException was thrown. (QueueSinkSpec.scala:35)
[info]   at org.scalatest.Assertions$class.newAssertionFailedException(Assertions.scala:496)
[info]   at org.scalatest.words.ResultOfAnTypeInvocation.shouldBe(ResultOfAnTypeInvocation.scala:71)
[info]   at akka.stream.scaladsl.QueueSinkSpec$$anonfun$1$$anonfun$apply$mcV$sp$3$$anonfun$apply$mcV$sp$10.apply(QueueSinkSpec.scala:41)

Issue here is that we are sending first promise to QueueSink it stores in stage because no data is available. Then second request comes and it fail second request promise because one request is already waiting for data. 
And now the strange part: for some reason from time to time promise2 is not returning result for Await.result(future2, 300.millis). I tried timeout like 1 sec - the same behavior.
I'm thinking - maybe because of sameThreadExecutionContext promise completion tryin to use the same thread that is blocked by Await?

Another observation: for failed cases both queue pull requests running before stage started (before promise is completed with async callback function).

Any thoughts are welcome.

rkuhn:
This is an ordering issue

* the first pull arrives before the request callback promise is fulfilled and is therefore enqueued
* the second pull arrives afterwards and is executed immediately
* the first pull is actually executed by the GraphStageLogic after fulfilling the promise, and this can lose the race against the other main test thread

The only solution I can see is to use an AtomicReference instead of a Promise:

* initially it contains Nil
* when pulling, try to replace existing list with one that has new pull prepended; if not a list any longer, just use the callback
* from preStart, getAndSet the callback in exchange for the enqueued pulls (reversed list) which are then immediately processed

agolubev:
Thanks for feedback. Fixed accordingly.

agolubev:
PR is ready for review

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -257,3 +233,89 @@ private[akka] final class HeadOptionStage[T] extends GraphStageWithMaterializedV
     }, p.future)
   }
 }
+
+/**
+ * INTERNAL API
+ */
+private[akka] class QueueSink[T]() extends GraphStageWithMaterializedValue[SinkShape[T], SinkQueue[T]] {
+  trait RequestElementCallback[E] {
+    val requestElement = new AtomicReference[Either[List[E], E ⇒ Unit]](Left(Nil))
```
drewhk: instead of Either please use case classes with proper names (helps understanding the logic)

rkuhn: for these cases I’d actually just use AnyRef and store the different things directly—we need pattern matching to find out the difference in any case

agolubev:
ready for review

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -257,3 +233,89 @@ private[akka] final class HeadOptionStage[T] extends GraphStageWithMaterializedV
...
+    }
+
+    (stageLogic, new SinkQueue[T] {
+      override def pull(): Future[Option[T]] = {
+        val ref = stageLogic.requestElement
+        val p = Promise[Option[T]]
+        ref.get() match {
+          case list: List[_] ⇒
+            if (!ref.compareAndSet(list, p :: list))
+              ref.get().asInstanceOf[Requested[T] ⇒ Unit](p)
```
rkuhn: This should be done via a pattern match: if some other thread concurrently tried to pull()—which is forbidden—then the error will be a ClassCastException instead of the more informative IllegalStateException.

rkuhn:
Apart from that final nitpick it is ready.

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -257,3 +233,94 @@ private[akka] final class HeadOptionStage[T] extends GraphStageWithMaterializedV
...
+        type PF = scala.PartialFunction[Any, Unit]
+        val callback: PF = { case f: Function1[_, _] ⇒ f.asInstanceOf[Requested[T] ⇒ Unit](p) }
+        val detectedConcurrency: PF = {
+          case _: List[_] ⇒ throw new IllegalStateException("Concurrent call of SinkQueue.pull() is detected")
+        }
+        val list: PF = {
+          case l: List[_] ⇒ if (!ref.compareAndSet(l, p :: l)) (detectedConcurrency orElse callback)(ref.get())
+        }
+
+        (list orElse callback)(ref.get())
```
rkuhn: Ouch, these are at least a handful of allocations too many ;-)
~~~scala
ref.get() match {
  case f: Function1[_, _] => f.asInstanceOf[Requested[T] => Unit](p)
  case l: List[_] =>
    if (!ref.compareAndSet(l, p :: l))
      ref.get() match {
        case f: Function1[_, _] => f.asInstanceOf[Requested[T] => Unit](p)
        case _ => throw new IllegalStateException("...")
      }
}
~~~

agolubev: yeah, make sense.
internal match should be like list first and than function
```
ref.get() match {
                case _: List[_]         ⇒ throw new IllegalStateException("Concurrent call of SinkQueue.pull() is detected")
                case f: Function1[_, _] ⇒ f.asInstanceOf[Requested[T] ⇒ Unit](p)
              }
```
as function will match the list as well

```
scala> val any:AnyRef = Nil
any: AnyRef = List()

scala> any match{
     | case _:Function1[_,_]=> println("function")
     | case _:List[_]=>println("list")
     | }
<console>:11: warning: unreachable code
              case _:List[_]=>println("list")
                                     ^
function
```

agolubev:
@rkuhn, PR is ready for review

rkuhn:
merging manually

rkuhn:
done (finally—slow internet on a train)

rkuhn:
Thanks again, @agolubev !

PR #19100, opened 2015-12-05
=htp #19099 make sure headers are always written on `finish` in `GzipCompressor`

### LINKED ISSUE #19099: Http: `encodeResponse` directive produces corrupted gzip data on empty chunked response stream
Usage of `encodeResponse` directive with empty chunked response stream leads to corrupted gzip data, due to gzip header never being sent.

Corresponding PR in Spray: https://github.com/spray/spray/pull/1090

```scala
implicit val system = ActorSystem("http-streaming")
implicit val flowBuilder = ActorMaterializer()

val result = HttpEntity.Chunked.fromData(MediaTypes.`text/html`, Source.empty)
val route = encodeResponse & path("stream") & complete(result)

Http().bindAndHandle(route, "localhost", 8080)
```

```
$ http :8080/stream
HTTP/1.1 200 OK
Content-Encoding: gzip
Content-Type: text/html
Date: Sat, 05 Dec 2015 22:15:02 GMT
Server: akka-http/2.3.12
Transfer-Encoding: chunked


http: error: ContentDecodingError: ('Received response with content-encoding: gzip, but failed to decode it.', error('Error -3 while decompressing: incorrect header check',))
```

ataraxer:
Closes #19099

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/coding/CodecSpecSupport.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/coding/CodecSpecSupport.scala
@@ -22,9 +22,11 @@ trait CodecSpecSupport extends Matchers with BeforeAndAfterAll { self: Suite ⇒
     i
   }
 
+  lazy val emptyTextBytes = ByteString(emptyText, "UTF8")
   lazy val smallTextBytes = ByteString(smallText, "UTF8")
   lazy val largeTextBytes = ByteString(largeText, "UTF8")
 
+  val emptyText = ""
   val smallText = "Yeah!"
   val largeText =
     """Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/coding/CoderSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/coding/CoderSpec.scala
@@ -32,6 +32,18 @@ abstract class CoderSpec extends WordSpec with CodecSpecSupport with Inspectors 
   def extraTests(): Unit = {}
 
   s"The ${Coder.encoding.value} codec" should {
+    "produce valid data on immediate finish" in {
+      streamDecode(Coder.newCompressor.finish()) should readAs(emptyText)
+    }
+    "properly encode an empty string" in {
+      streamDecode(ourEncode(emptyTextBytes)) should readAs(emptyText)
+    }
+    "properly decode an empty string" in {
+      ourDecode(streamEncode(emptyTextBytes)) should readAs(emptyText)
+    }
+    "properly round-trip encode/decode an empty string" in {
+      ourDecode(ourEncode(emptyTextBytes)) should readAs(emptyText)
+    }
     "properly encode a small string" in {
       streamDecode(ourEncode(smallTextBytes)) should readAs(smallText)
     }
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/CodingDirectivesSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/directives/CodingDirectivesSpec.scala
@@ -214,6 +214,19 @@ class CodingDirectivesSpec extends RoutingSpec with Inside {
         Gzip.decode(bytes).awaitResult(1.second) should readAs(text)
       }
     }
+    "correctly encode the chunk stream produced by an empty chunked response" in {
+      val emptyChunkedEntity = HttpEntity.Chunked(MediaTypes.`text/plain`, Source.empty)
+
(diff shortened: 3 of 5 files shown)


ktoso:
OK TO TEST

jrudolph:
Thanks for the PR, @ataraxer. Fix and test LGTM. We should also add a test to `CoderSpec` which has the additional benefit that all the encoders will be tested against similar kind of bugs.

jrudolph:
I've seen you've added tests to spray in the equivalent of CoderSpec so maybe you could just copy them over.

ataraxer:
@jrudolph unfortunately it turns out those particular tests aren't reliable in this particular case due to `ourGzip` explicitly calling `compress`. I'll modify them to call `finish` directly and carry over to this PR.

ataraxer:
Aaaand done.

While working on tests I've also discovered that empty string encoding is broken because `compressAndFinish` calls `newTempBuffer` with size of an input as an argument, leading to empty buffer being created and deflator not being able to produce any output via it (imperative programming sucks :c ).

I've also noticed that there is no limit on buffer size whatsoever, which may lead to excessive memory footprint on compression of big entities, which would effectively double the amount of used memory used during compression.

ataraxer:
Spray seems to be OK on those issues, since it uses fixed-size buffer.

jrudolph:
> While working on tests I've also discovered that empty string encoding is broken because compressAndFinish calls newTempBuffer with size of an input as an argument, leading to empty buffer being created and deflator not being able to produce any output via it (imperative programming sucks :c ).

Ah, nice. I yesterday tried those test myself quickly and saw that something was wrong.

> I've also noticed that there is no limit on buffer size whatsoever, which may lead to excessive memory footprint on compression of big entities, effectively doubling the amount of memory used during compression.

Can you explain how this would happen? The Gzip and Deflate compressors both used the default value which isn't different from what you've specified here. The decompressors use a configurable maximum number of bytes which I think is a more flexible solution.

ataraxer:
> Can you explain how this would happen? The Gzip and Deflate compressors both used the default value which isn't different from what you've specified here. The decompressors use a configurable maximum number of bytes which I think is a more flexible solution.

Indeed, my concern might have been a bit superficial: it only appears to be the case when `Encoder.encode` is invoked directly on `ByteString` (`encode => compressAndFinish => newTempBuffer(input.size)` leading to buffer size of up to `Int.MaxValue`) which only happens in tests as far as I can see. However it is still a public API so it might be a good idea to keep buffer size in check anyway.

jrudolph:
> However it is still a public API so it might be a good idea to keep buffer size in check anyway.

Yes, but if we want to do it we should do it consistently in the same way as it is done for the decoding side. I'd say let's keep it the way it is and only introduce the lower bound for now to fix the bug.

Then we can create another ticket to figure out if and how we want to fix the upper bound issue on the encoding side. I agree it can be somewhat surprising but we could even document ourselves out of it by specifying that an encoding call may temporarily allocate additional memory in the size of the incoming buffer which is still a predictable solution (in contrast to the former solution on the decoding side which lead to the introduction of the buffer bound).

ataraxer:
> Yes, but if we want to do it we should do it consistently in the same way as it is done for the decoding side. I'd say let's keep it the way it is and only introduce the lower bound for now to fix the bug. Then we can create another ticket to figure out if and how we want to fix the upper bound issue on the encoding side.

Sounds good to me.

ataraxer:
Removed the upper bound and created a separate issue: #19119.

ataraxer:
ping

ktoso:
Thanks for the fix and the other ticket, @ataraxer.
LGTM so I'll merge right away then (last days was team meeting, so we we're not really online, thanks for your patience!)

jrudolph:
Great, thanks @ataraxer.

jrudolph:
Oops, this also targeted the wrong branch.

ktoso:
Ouch nooo. I didn't notice..
Ill more it to the right one tomorrow, then.

ktoso:
I'm doing the porting now.

ataraxer:
> Oops, this also targeted the wrong branch.

Sorry, it is a bit unclear which branch should be used for akka-http/akka-streams PR's since they are not present in the master for some reason and ...well, there are a lot of branches out there :\

ataraxer:
Also I would like to remind that original Spray issue is still open: https://github.com/spray/spray/pull/1090

ktoso:
> Sorry, it is a bit unclear which branch should be used for akka-http/akka-streams PR's since they are not present in the master for some reason and ...well, there are a lot of branches out there :\

Hey, thanks for your contribs and sorry we didn't notice right away. 
The branches you care about are (at the moment I'm writing this):

- `release-2.3-dev` = akka streams and http
- `master` = akka 2.4 (currently)
- `release-2.3` = akka 2.3 maintanance

We'll soon merge streams into master, we'll let the community know about it

ataraxer:
Thanks for the clarifications! (:

PR #19101, opened 2015-12-06
+htc add `text/csv(UTF-8)` ContentType

ktoso:
"CSV files considered the ultimate integration pattern" :wink:

Adding the ContentType here is useful when someone wants to render a raw response which we know is a csv (i.e. fed from a file)


johanandren:
LGTM ;)

ktoso:
Thanks :wink:

Issue #18736, opened 2015-10-19
Make Intersperse a GraphStage instead of StatefulStage


ktoso:
It's possible to make it a PushPullStage: https://github.com/akka/akka/pull/18734/files#diff-3203199ba389ee802eef94486f26dc88R242

lolski:
I'll give it a try

ktoso:
Cool, thanks! Marking as in progress

drewhk:
@lolski I recommend to go for a `GraphStage` directly instead. Can you do that?

ktoso:
+1 to Endre's GraphStage point, as I've mentioned about this ticket already over the weekend: :point_up: [November 29, 2015 1:56 PM](https://gitter.im/akka/dev?at=565af5f549e74fad21eb96f8)

lolski:
No problem, I'll give it a try (and maybe ask a couple of questions along the way).

PR #19102, opened 2015-12-06
GraphStage-based intersperse

lolski:
Please see https://github.com/akka/akka/issues/18736

This intersperse implementation is based on GraphStage. It has passed the `akka.stream.scaladsl.FlowIntersperseSpec`.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -247,41 +247,35 @@ final case class Intersperse[T](start: Option[T], inject: T, end: Option[T]) ext
   override val shape = FlowShape(in, out)
 
   override def createLogic(attr: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
-    var first = true // is this the first element?
-
-    setHandler(in, new InHandler {
+    val startInHandler = new InHandler {
       override def onPush(): Unit = {
-        if (first) {
-          first = false
-          // emit: start if defined, elem
-          if (start.isDefined) emitMultipe(out, Iterator(start.get, grab(in)))
-          else emit(out, grab(in))
-        } else {
-          // emit: inject, elem
-          emitMultipe(out, Iterator(inject, grab(in)))
-        }
+        // if else (to avoid using Iterator[T].flatten in hot code)
+        if (start.isDefined) emitMultipe(out, Iterator(start.get, grab(in)))
+        else emit(out, grab(in))
+        setHandler(in, restInHandler) // switch handler
       }
 
       override def onUpstreamFinish(): Unit = {
-        if (first) {
-          // this branch is only called if upstream is an empty one
-          // emit: start if defined, end if defined, otherwise emit nothing
-          if (start.isDefined && end.isDefined) emitMultipe(out, Iterator(start.get, end.get))
-          else if (start.isDefined && !end.isDefined) emit(out, start.get)
-          else if (!start.isDefined && end.isDefined) emit(out, end.get)
-          else { /* emit nothing */ }
-        } else {
-          // emit: end if defined, otherwise emit nothing
-          if (end.isDefined) emit(out, end.get)
-          else { /* emit nothing */ }
-        }
+        emitMultipe(out, Iterator(start, end).flatten)
         completeStage()
       }
-    })
+    }
 
-    setHandler(out, new OutHandler {
+    val restInHandler = new InHandler {
+      override def onPush(): Unit = emitMultipe(out, Iterator(inject, grab(in)))
+
+      override def onUpstreamFinish(): Unit = {
+        if (end.isDefined) emit(out, end.get)
+        completeStage()
+      }
+    }
+
+    val outHandler = new OutHandler {
       override def onPull(): Unit = pull(in)
-    })
+    }
+
+    setHandler(in, startInHandler)
+    setHandler(out, outHandler)
   }
 }
 
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -250,19 +250,19 @@ final case class Intersperse[T](start: Option[T], inject: T, end: Option[T]) ext
     val startInHandler = new InHandler {
       override def onPush(): Unit = {
         // if else (to avoid using Iterator[T].flatten in hot code)
-        if (start.isDefined) emitMultipe(out, Iterator(start.get, grab(in)))
+        if (start.isDefined) emitMultiple(out, Iterator(start.get, grab(in)))
         else emit(out, grab(in))
         setHandler(in, restInHandler) // switch handler
       }
 
       override def onUpstreamFinish(): Unit = {
-        emitMultipe(out, Iterator(start, end).flatten)
+        emitMultiple(out, Iterator(start, end).flatten)
         completeStage()
       }
     }
 
     val restInHandler = new InHandler {
-      override def onPush(): Unit = emitMultipe(out, Iterator(inject, grab(in)))
+      override def onPush(): Unit = emitMultiple(out, Iterator(inject, grab(in)))
 
       override def onUpstreamFinish(): Unit = {
         if (end.isDefined) emit(out, end.get)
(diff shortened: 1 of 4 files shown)


akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -239,43 +239,49 @@ private[akka] final case class Fold[In, Out](zero: Out, f: (Out, In) ⇒ Out, de
...
+  override def createLogic(attr: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    var first = true // is this the first element?
 
-  def running = new StageState[T, T] {
-    override def onPush(elem: T, ctx: Context[T]): SyncDirective =
-      emit(Iterator(inject, elem), ctx)
-  }
+    setHandler(in, new InHandler {
+      override def onPush(): Unit = {
+        if (first) {
```
hepin1989: could we reset the InHandler?and then we do not need `first`.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -239,43 +239,49 @@ private[akka] final case class Fold[In, Out](zero: Out, f: (Out, In) ⇒ Out, de
...
-      case _ ⇒
-        terminationEmit(Iterator(), ctx)
-    }
+      override def onUpstreamFinish(): Unit = {
+        if (first) {
+          // this branch is only called if upstream is an empty one
+          // emit: start if defined, end if defined, otherwise emit nothing
+          if (start.isDefined && end.isDefined) emitMultipe(out, Iterator(start.get, end.get))
+          else if (start.isDefined && !end.isDefined) emit(out, start.get)
+          else if (!start.isDefined && end.isDefined) emit(out, end.get)
```
johanandren: Could you maybe put the options in the iterable and just flatten it to achieve this with one line or so?

ktoso:
OK TO TEST

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -239,43 +239,49 @@ private[akka] final case class Fold[In, Out](zero: Out, f: (Out, In) ⇒ Out, de
...
-  def running = new StageState[T, T] {
-    override def onPush(elem: T, ctx: Context[T]): SyncDirective =
-      emit(Iterator(inject, elem), ctx)
-  }
+    setHandler(in, new InHandler {
+      override def onPush(): Unit = {
+        if (first) {
+          first = false
+          // emit: start if defined, elem
+          if (start.isDefined) emitMultipe(out, Iterator(start.get, grab(in)))
```
ktoso: @lolski we've fixed a typo here, the method is now `emitMultiple`, could you rebase this PR please?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: yep, that should work.
pseudocode:

```scala
setHandler(in, {
  // handle first push
  setHandler(in, { /*handle other pushes*/ }
}
```

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -239,43 +239,49 @@ private[akka] final case class Fold[In, Out](zero: Out, f: (Out, In) ⇒ Out, de
...
-      case Some(e) ⇒
-        terminationEmit(Iterator(end.get), ctx)
-      case _ ⇒
-        terminationEmit(Iterator(), ctx)
-    }
+      override def onUpstreamFinish(): Unit = {
+        if (first) {
+          // this branch is only called if upstream is an empty one
+          // emit: start if defined, end if defined, otherwise emit nothing
+          if (start.isDefined && end.isDefined) emitMultipe(out, Iterator(start.get, end.get))
```
ktoso: oh right, disregard my comment in https://github.com/akka/akka/pull/19102/files#r46811292 
since this will cause us to need the `first` marker anyway

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: nice idea

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -239,43 +239,49 @@ private[akka] final case class Fold[In, Out](zero: Out, f: (Out, In) ⇒ Out, de
...
+        } else {
+          // emit: end if defined, otherwise emit nothing
+          if (end.isDefined) emit(out, end.get)
+          else { /* emit nothing */ }
+        }
+        completeStage()
+      }
+    })
+
+    setHandler(out, new OutHandler {
```
ktoso: Thought: we have "typical" behaviours cached as objects, such as:

```
object EagerTerminateOutput extends OutHandler {
    override def onPull(): Unit = ()
  }
```

Why not add one that does only `pull` as predefined one as well?

// cc @drewhk @rkuhn

ktoso: Meh, nevermind, not doable since it needs to take in the `in` param, so it wouldn't be an object.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
johanandren: Wouldn't there just be a different onUpstreamFinish in the first handler?

ktoso: Hm, that's true actually - the coffee didn't kick in still it seems :wink:

lolski:
@hepin1989 @johanandren good advice guys, thanks! @ktoso Woah, I didn't even notice. I will do the rebase ASAP.

lolski:
@ktoso now it's rebased :)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -239,43 +239,43 @@ private[akka] final case class Fold[In, Out](zero: Out, f: (Out, In) ⇒ Out, de
...
-  override def onUpstreamFinish(ctx: Context[T]): TerminationDirective = {
-    end match {
-      case Some(e) if needsToEmitStart ⇒
-        terminationEmit(Iterator(start.get, end.get), ctx)
-      case Some(e) ⇒
-        terminationEmit(Iterator(end.get), ctx)
-      case _ ⇒
-        terminationEmit(Iterator(), ctx)
+      override def onUpstreamFinish(): Unit = {
+        if (end.isDefined) emit(out, end.get)
```
johanandren: Don't you need to complete stage after emitting here?

lolski: Yes, it will call `completeStage` after the optional `emit` https://github.com/lolski/akka/blob/lolski-issue-18736/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala#L269, no?

johanandren: Sorry, my brain inferred an else before the next line.

I'm not a 100% sure about this, because emit replaces the handler temporarily until it has emitted all elements and then reinstates it, but I think now it will be completed before the new handler can do anything. Is there test coverage for this that says that it behaves as expected?

lolski: Not sure. I'll try to skim the docs and see what I can find.

If any `GraphStage` experts can chime in it would be great.

lolski: As per @rkuhn comment https://gitter.im/akka/dev?at=5667fcdf868b8da62a258921, I think we can conclude that this is safe.

rkuhn: Yes, this must work, if it does not then emit/completeStage has a bug.

lolski:
Added stage completion test on `FlowIntersperseSpec`.

lolski:
I think the error is unrelated

rkuhn:
LGTM, thanks!

PR #19104, opened 2015-12-06
Implement Operation Fusing

rkuhn:
# IT FUSES!

Some tests fail when aggressively fusing everything everywhere, most notably TCP does not seem to like this (which indicates some implementation trouble), but all features work, including materialized value sources (which are fused now as well).

The solution is not particularly smart or efficient, but getting something to work is the first step. Fusing the HttpServerBlueprint (into one GraphModule and three non-fusable stages) takes roughly 170µs on my machine, which I deem to be good enough for starters.

Next up would be to see how this changes the HTTP benchmark numbers, but for that we need working TCP support first.


drewhk:
Tcp is the only stage now that uses ActorRef support so it can be very well the source of the fuse unfriendliness (i.e. likely a bug there.). Also, Tcp silently attaches a Detacher stage for read-ahead, that can be also the source of problems (unlikely though).

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
```diff
@@ -0,0 +1,362 @@
...
+  private def descend(m: Module,
+                      inheritedAttributes: Attributes,
+                      struct: BuildStructuralInfo,
+                      openGroup: ju.Set[Module],
+                      indent: String): MaterializedValueNode = {
+    def log(msg: String): Unit = println(indent + msg)
+    val async = m match {
+      case _: GraphStageModule ⇒ m.attributes.contains(AsyncBoundary)
+      case _ if m.isAtomic     ⇒ true
+      case _                   ⇒ m.attributes.contains(AsyncBoundary)
```
2m: Should we check here for `Dispatcher` attribute as well? Setting a custom dispatcher effectively means async boundary.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -23,14 +29,14 @@ private[akka] object StreamLayout {
   // compile-time constant
   final val Debug = false
 
-  final def validate(m: Module, level: Int = 0, doPrint: Boolean = false, idMap: mutable.Map[AnyRef, Int] = mutable.Map.empty): Unit = {
+  final def validate(m: Module, level: Int = 0, doPrint: Boolean = false, idMap: ju.Map[AnyRef, Integer] = new ju.HashMap): Unit = {
     val ids = Iterator from 1
     def id(obj: AnyRef) = idMap get obj match {
-      case Some(x) ⇒ x
-      case None ⇒
+      case null ⇒
```
drewhk: yay! :)

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -98,10 +109,24 @@ private[akka] object StreamLayout {
...
-  case class Combine(f: (Any, Any) ⇒ Any, dep1: MaterializedValueNode, dep2: MaterializedValueNode) extends MaterializedValueNode
-  case class Atomic(module: Module) extends MaterializedValueNode
-  case class Transform(f: Any ⇒ Any, dep: MaterializedValueNode) extends MaterializedValueNode
+  sealed trait MaterializedValueNode {
+    /*
+     * These nodes are used in hash maps and therefore must have efficient implementations
+     * of hashCode and equals. There is no value in allowing aliases to be equal, so using
+     * reference equality.
+     */
+    override def hashCode: Int = super.hashCode
```
drewhk: good

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -229,6 +254,39 @@ private[akka] object StreamLayout {
     }
 
     /**
+     * Creates a new Module which is `this` Module composed with `that` Module.
+     *
+     * The difference to compose(that) is that this version completely ignores the materialized value
+     * computation of `that` while the normal version executes the computation and discards its result.
+     * This means that this version must not be used for user-provided `that` modules, it can only be
+     * used where we know that there is no meaningful computation to be done (like for
+     * MaterializedValueSource).
```
drewhk: maybe add something like: because in user facing DSLs we need to call mapMaterialized for possible side-effects

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -236,7 +294,7 @@ private[akka] object StreamLayout {
       if (Debug) validate(this)
 
       CompositeModule(
-        subModules = Set(this),
+        Set(this),
```
drewhk: why remove the named parameter

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -321,7 +379,8 @@ private[akka] object StreamLayout {
     override val downstreams: Map[OutPort, InPort],
     override val upstreams: Map[InPort, OutPort],
     override val materializedValueComputation: MaterializedValueNode,
-    attributes: Attributes) extends Module {
+    attributes: Attributes,
+    extra: Any = null) extends Module {
```
drewhk: extra will need a rename and type as we discussed, but can be another ticket

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -596,10 +516,10 @@ private[stream] object MaterializerSession {
 private[stream] abstract class MaterializerSession(val topLevel: StreamLayout.Module, val initialAttributes: Attributes) {
   import StreamLayout._
 
-  private var subscribersStack: List[mutable.Map[InPort, Subscriber[Any]]] =
-    mutable.Map.empty[InPort, Subscriber[Any]].withDefaultValue(null) :: Nil
-  private var publishersStack: List[mutable.Map[OutPort, Publisher[Any]]] =
-    mutable.Map.empty[OutPort, Publisher[Any]].withDefaultValue(null) :: Nil
+  private var subscribersStack: List[ju.Map[InPort, Subscriber[Any]]] =
+    new ju.HashMap[InPort, Subscriber[Any]] :: Nil
```
drewhk: good

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: It is already replaced with FusedModule, but not yet pushed.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -671,20 +591,27 @@ private[stream] abstract class MaterializerSession(val topLevel: StreamLayout.Mo
   protected def mergeAttributes(parent: Attributes, current: Attributes): Attributes =
     parent and current
 
+  private val matValSrc: ju.Map[MaterializedValueNode, List[MaterializedValueSource[Any]]] = new ju.HashMap
+  def registerSrc(ms: MaterializedValueSource[Any]): Unit = {
+    if (MaterializerSession.Debug) println(s"registering source $ms")
+    matValSrc.get(ms.computation) match {
+      case null ⇒ matValSrc.put(ms.computation, ms :: Nil)
```
drewhk: are these kind of `match` statements properly optimized? I.e. is it equivalent to an `if`?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -671,20 +591,27 @@ private[stream] abstract class MaterializerSession(val topLevel: StreamLayout.Mo
...
     for (submodule ← module.subModules) {
       val subEffectiveAttributes = mergeAttributes(effectiveAttributes, submodule.attributes)
       submodule match {
-        case mv: MaterializedValueSource[_] ⇒
-          val pub = new MaterializedValuePublisher
-          materializedValuePublishers ::= pub
-          materializedValues.put(mv, ())
-          assignPort(mv.shape.outlet, pub)
+        case GraphStageModule(shape, attributes, mv: MaterializedValueSource[_]) ⇒
+          val copy = mv.copy.asInstanceOf[MaterializedValueSource[Any]]
```
drewhk: what is the purpose of the copy here?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -694,38 +621,54 @@ private[stream] abstract class MaterializerSession(val topLevel: StreamLayout.Mo
...
+  private def resolveMaterialized(matNode: MaterializedValueNode, matVal: ju.Map[Module, Any], indent: String): Any = {
+    if (MaterializerSession.Debug) println(indent + matNode)
+    val ret = matNode match {
+      case Atomic(m)          ⇒ matVal.get(m)
+      case Combine(f, d1, d2) ⇒ f(resolveMaterialized(d1, matVal, indent + "  "), resolveMaterialized(d2, matVal, indent + "  "))
+      case Transform(f, d)    ⇒ f(resolveMaterialized(d, matVal, indent + "  "))
+      case Ignore             ⇒ ()
+    }
+    if (MaterializerSession.Debug) println(indent + s"result = $ret")
+    matValSrc.remove(matNode) match {
```
drewhk: is there a reason why we remove here? (and not just check)

akka-stream/src/main/scala/akka/stream/impl/Timers.scala
```diff
@@ -105,6 +105,8 @@ private[stream] object Timers {
     val in2 = Inlet[O]("in2")
     val out1 = Outlet[I]("out1")
     val out2 = Outlet[O]("out2")
+
+    override def initialAttributes = Attributes.name("IdleBidi")
```
drewhk: The Bidi is not idle :) IdleTimeoutBidi maybe (also rename class)

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -372,8 +374,10 @@ private[stream] class ActorGraphInterpreter(
       inputs(id).onComplete()
       runBatch()
     case OnSubscribe(id: Int, subscription: Subscription) ⇒
+      if (GraphInterpreter.Debug) println(s"${interpreter.Name}  onSubscribe id=$id")
       subscribesPending -= 1
       inputs(id).onSubscribe(subscription)
+      runBatch()
```
drewhk: good

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
```diff
@@ -21,12 +38,13 @@ object GraphStages {
    * INERNAL API
    */
   private[stream] abstract class SimpleLinearGraphStage[T] extends GraphStage[FlowShape[T, T]] {
-    val in = Inlet[T]("in")
-    val out = Outlet[T]("out")
+    val in = Inlet[T](Logging.simpleName(this) + ".in")
```
drewhk: nice

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
```diff
@@ -124,4 +146,30 @@ object GraphStages {
       (logic, cancellable)
     }
   }
+
+  /**
+   * INTERNAL API.
+   *
+   * This source is not reusable, it is only created internally.
+   */
+  private[stream] class MaterializedValueSource[T](val computation: MaterializedValueNode, val out: Outlet[T]) extends GraphStage[SourceShape[T]] {
```
drewhk: Looks ok. Maybe there are cheaper ways to do this, but since these sources are rare, and the cost is only paid when it is used, I think this is fine.

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
```diff
@@ -0,0 +1,362 @@
...
+import akka.stream._
+import akka.stream.impl.StreamLayout._
+import akka.stream.impl.fusing.GraphStages.MaterializedValueSource
+import akka.stream.Attributes.AsyncBoundary
+import scala.util.control.NonFatal
+import akka.stream.stage.GraphStageWithMaterializedValue
+import scala.annotation.tailrec
+import java.util.Arrays
+
+object Fusing {
```
drewhk: entering now this beast...

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
```diff
@@ -0,0 +1,362 @@
...
+
+  final val Debug = false
+
+  def aggressive[S <: Shape, M](g: Graph[S, M]): FusedGraph[S, M] = {
```
drewhk: what does aggresive mean? (i.e. docs)

jrudolph:
Here are the result of a benchmark I did against this branch with a few config changes as shown in [this branch](https://github.com/spray/akka/commits/w/fusing-benchmark).

Throughput is about 12-15k RPS.

I think isn't surprising as per-request stream-of-streams materialization still easily dwarfs everything else (50% time spent alone in FlattenMerge). I'm not sure about this but I would guess that doing the fusing alone would actually introduce a bit of overhead to materialization if you cannot reuse complete pre-fused flow-graphs.

Here are the flames ([see here for the interactive svg](http://jrudolph.github.io/akka-http-fused-in-flames.svg)):



Notable:
 * 28 % spent in fusing inside of FlattenMerge
 * most of this time is spent in StreamLayout.validate

[In another experiment](https://github.com/spray/akka/commit/1255efdeaccdff1bf5a5a658ccae085762dc62c9) I added a patch that removes streaming entity support from HTTP to get rid of the stream-of-streams parts. In that case the fusing speedups can be realized so that RPS goes up to ~ 90-110k RPS from what was 40-60k before fusing.

There seems to a bug because the connection seem to be stuck in the beginning for several seconds before data starts flowing (is that the remaining TCP problem mentioned somewhere?). The bug can be reproduced by running `test:run-main akka.http.scaladsl.TestServer` and trying to `curl http://localhost:9001/ping` which only finishes after ~ 7.5 seconds. So, the above RPS numbers need to be taken with a grain of salt because I had to extrapolate to what to expect without the initial hold-up.

rkuhn:
Thanks a lot, @jrudolph! The costly `validate` call should of course be removed, it is in there only to easy current development (i.e. blow up early), other than that this looks very encouraging!

drewhk:
What I find nicest is that there is practically no meaningful overhead from the graphInterpreter here. The only interpreter methods that show up are getting their overhead transitively from fuse, materialize or creating graphs, or Dispatcher.dispatch. We are far from hitting that point where it is the bottleneck itself.

drewhk:
Thanks @jrudolph !

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
drewhk: Alternatively, we can log a warning in those cases: "Have you forgot an AsyncBoundary somewhere?".

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
```diff
@@ -0,0 +1,362 @@
...
+    GraphModule(new GraphInterpreter.GraphAssembly(stages, attributes, ins, inOwners, outs, outOwners), shape, Attributes.none, matValIDs)
+  }
+
+  @tailrec private def copyToArray[T](it: ju.Iterator[T], array: Array[T], idx: Int): Int =
+    if (it.hasNext) {
+      array(idx) = it.next()
+      copyToArray(it, array, idx + 1)
+    } else idx
+
+  private def descend(m: Module,
```
drewhk: I would pretty much apprieciate a large internal doc block just like in the GraphInterpreter file explaining the overall phases, data structures and principles used. Then with some smaller comment blocks on the actual pieces. This can be a hell to maintain otherwise.

drewhk:
btw, make sure that benchmarks are not run with `fuzzing-mode-enabled`!!!

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
```diff
@@ -0,0 +1,362 @@
...
+      m match {
+        case CopiedModule(shape, _, copyOf) ⇒
+          val ret = descend(copyOf, attributes, struct, localGroup, indent + "  ")
+          struct.rewire(copyOf.shape, shape, indent)
+          ret
+        case _ ⇒
+          struct.enterMatCtx()
+          val subMat: Predef.Map[Module, MaterializedValueNode] =
+            m.subModules.map(sub ⇒ sub -> descend(sub, attributes, struct, localGroup, indent + "  "))(collection.breakOut)
+          val down = m.downstreams.toSet -- m.subModules.map(_.downstreams).reduce(_ ++ _)
```
drewhk: maybe enough to (m.subModules.iterator) there? And `_.downstreams.iterator`.

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
```diff
@@ -0,0 +1,362 @@
...
+
+  private def descend(m: Module,
+                      inheritedAttributes: Attributes,
+                      struct: BuildStructuralInfo,
+                      openGroup: ju.Set[Module],
+                      indent: String): MaterializedValueNode = {
+    def log(msg: String): Unit = println(indent + msg)
+    val async = m match {
+      case _: GraphStageModule ⇒ m.attributes.contains(AsyncBoundary)
+      case _ if m.isAtomic     ⇒ true
```
drewhk: Maybe explain in a comment that this arm means "is atomic and it is *not* a GraphStageModule". I first read it as atomic module => async.

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
```diff
@@ -0,0 +1,362 @@
...
+            if (Debug) require(find(ms.computation, m.materializedValueComputation), s"mismatch:\n  ${ms.computation}\n  ${m.materializedValueComputation}")
+            val replacement = CopiedModule(c.shape, c.attributes, new MaterializedValueSource[Any](newMat, ms.out).module)
+            struct.replace(c, replacement, localGroup)
+          }
+          newMat
+      }
+    }
+  }
+
+  private def find(m1: MaterializedValueNode, m2: MaterializedValueNode): Boolean =
```
drewhk: give better names to arguments since the case is not symmetric (m1 and m2 are treated differently, therefore they have a clear role)

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
```diff
@@ -0,0 +1,362 @@
...
+      group
+    }
+
+    def addModule(m: Module, group: ju.Set[Module], inheritedAttributes: Attributes, indent: String): Atomic = {
+      val copy = CopiedModule(m.shape.deepCopy(), inheritedAttributes, realModule(m))
+      if (Debug) println(indent + s"adding copy ${hash(copy)} ${printShape(copy.shape)} of ${printShape(m.shape)}")
+      group.add(copy)
+      modules.add(copy)
+      copy.shape.outlets.foreach(o ⇒ outGroup.put(o, group))
+      m.shape.inlets.iterator.zip(copy.shape.inlets.iterator).foreach { p ⇒ addMapping(p._1, p._2, newIns) }
```
drewhk: joint while loop to eliminate pair allocation

drewhk:
Not fully grasping yet the algorithm, but found no gaping holes so far.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: yes, thanks

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: because it didn’t make sense for me to special-case this single parameter-passing: other parameters are also not named, and in fact we only routinely name boolean parameters

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: yes, they are (just decompiled to check)

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: every inclusion of a MaterializedValueSource must have its own Promise—they are not reusable

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: I thought keeping that map small should be beneficial for lookup speed.

akka-stream/src/main/scala/akka/stream/impl/Timers.scala
rkuhn: Boy-scout all the names!

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
ktoso: very good to know, thanks

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: makes sense

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: Ok, I guessed right, but I don't yet see how the copy gets *known* where it needs to be.

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
rkuhn: yes, definitely

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
rkuhn: A foldLeft is actually clearer and should also be faster than the current one.

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
rkuhn: yep

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
rkuhn: done

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
rkuhn: yes, I first wanted to get it to work, but can optimize now

rkuhn:
@drewhk I pushed the latest (mostly docs).

rkuhn:
Making Source.single fusable had some interesting side-effects, most notably requiring runBatch from preStart and rethinking the calculation of the eventLimit—which was zero in this case ;-)

rkuhn:
IT DOES FUSE NOW!

Will merge once reviewed and validated.

jrudolph:
> What I find nicest is that there is practically no meaningful overhead from the graphInterpreter here.

It's a bit hard to see here because with materialization still being big the details may not be shown (that's basically a feature of sampling). In my experiment it seems the interpretation overhead is about 20% which I think is still quite good. Half of that overhead is just the cost of virtual/interface calls at the megamorphic call sites in the interpreter which cannot be JIT optimized (because they really are dynamic in nature).

drewhk:
Yes, I agree with all of that. What I wanted to say that it is not yet the bottleneck we have to worry about. Hopefully its allocation-free design also gives some advantage in load tests in longer runs.

Issue #18758, opened 2015-10-21
Add named exception to detect when a cluster node has been quarantined by others


garyiwu:
When a cluster node gets auto-downed/quarantined, it has to restart its own actor system before it can rejoin the cluster.  To do this we need a way to reliably detect when a node has been quarantined by other nodes.  

Currently, the only way to do this appears to be doing string matching on the exception message of an `AssociationError`, which is fragile.  Ideally there should be a named exception that we can check, so that the node can restart itself as appropriate.

jdevelop:
:+1:

jdevelop:
I faced the same problem and came out with the following solution:

- add the special case class **AssociationHandle.ActorSystemQuarantinedEvent** that will contain the local address of the system being quarantined, and remote address of the system that has quarantined the current one.
- upon receive of the quarantine message in the Endpoint - send a new **AssociationHandle.ActorSystemQuarantinedEvent** to the event stream of the current ActorSystem. 

Hence any actor could subscribe to those events and in case of quarantine - do something, e.g - trigger system restart.

I found this pretty flexible in terms of usage, however need somebody to review the approach and the proposed patch. Thanks.

jdevelop:
Perhaps this one could be mark as fixed?

rkuhn:
Indeed, thanks!

PR #19107, opened 2015-12-07
#18758 Send appropriate events on remote actor system shutdown and quarantine

jdevelop:
Publish appropriate events to the current ActorSystem event stream upon remote ActorSystem shutdown or when current ActorSystem is quarantined by the remote ActorSystem.

### DIFF
--- a/akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteRestartedQuarantinedSpec.scala
+++ b/akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteRestartedQuarantinedSpec.scala
@@ -100,6 +100,7 @@ abstract class RemoteRestartedQuarantinedSpec
       runOn(second) {
         val addr = system.asInstanceOf[ExtendedActorSystem].provider.getDefaultAddress
         val firstAddress = node(first).address
+        system.eventStream.subscribe(testActor, classOf[ThisActorSystemQuarantinedEvent])
 
         val (_, ref) = identifyWithUid(first, "subject")
 
@@ -112,6 +113,10 @@ abstract class RemoteRestartedQuarantinedSpec
               ref ! "boo!"
             }
           }
+        }
+
+        expectMsgPF(10 seconds) {
+          case ThisActorSystemQuarantinedEvent(local, remote) ⇒
         }
 
         enterBarrier("still-quarantined")
--- a/akka-remote/src/main/scala/akka/remote/RemotingLifecycleEvent.scala
+++ b/akka-remote/src/main/scala/akka/remote/RemotingLifecycleEvent.scala
@@ -3,6 +3,7 @@
  */
 package akka.remote
 
+import akka.event.Logging.LogLevel
 import akka.event.{ LoggingAdapter, Logging }
 import akka.actor.{ ActorSystem, Address }
 
@@ -85,6 +86,12 @@ final case class QuarantinedEvent(address: Address, uid: Int) extends RemotingLi
     s"Association to [$address] having UID [$uid] is irrecoverably failed. UID is now quarantined and all " +
       "messages to this UID will be delivered to dead letters. Remote actorsystem must be restarted to recover " +
       "from this situation."
+}
+
+@SerialVersionUID(1L)
+final case class ThisActorSystemQuarantinedEvent(localAddress: Address, remoteAddress: Address) extends RemotingLifecycleEvent {
+  override def logLevel: LogLevel = Logging.WarningLevel
+  override val toString: String = s"The remote system ${remoteAddress} has quarantined this system ${localAddress}."
 }
 
 /**
--- a/akka-docs/rst/java/remoting.rst
+++ b/akka-docs/rst/java/remoting.rst
@@ -377,6 +377,9 @@ contains the addresses the remoting listens on.
 
 To be notified  when the remoting subsystem has been shut down, listen to ``RemotingShutdownEvent``.
 
+To be notified when the current system is quarantined by the remote system, listen to ``ThisActorSystemQuarantinedEvent``,
+which includes the addresses of local and remote ActorSystems.
+
 To intercept generic remoting related errors, listen to ``RemotingErrorEvent`` which holds the ``Throwable`` cause.
 
 Remote Security
--- a/akka-docs/rst/scala/remoting.rst
+++ b/akka-docs/rst/scala/remoting.rst
@@ -379,6 +379,9 @@ holds the direction of the association (inbound or outbound), the addresses of t
 To be notified  when the remoting subsystem is ready to accept associations, listen to ``RemotingListenEvent`` which
 contains the addresses the remoting listens on.
 
+To be notified when the current system is quarantined by the remote system, listen to ``ThisActorSystemQuarantinedEvent``,
+which includes the addresses of local and remote ActorSystems.
+
 To be notified  when the remoting subsystem has been shut down, listen to ``RemotingShutdownEvent``.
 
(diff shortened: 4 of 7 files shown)


johanandren:
OK TO TEST

johanandren:
Refs #18758

johanandren:
Should have test coverage?

johanandren:
And probably documentation as well.

johanandren:
PLS BUILD

jdevelop:
Anyone has an idea of how can I emulate that disassociation event in a sort of integration test? Or setting up an instance of Endpoint and sending a disassociation message to it would be enough? Please advice.

johanandren:
Maybe ```NodeLeavingAndExitingAndBeingRemovedSpec``` in akka-cluster can be useful to look at?

johanandren:
PLS BUILD

akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteRestartedQuarantinedSpec.scala
```diff
@@ -114,6 +115,10 @@ abstract class RemoteRestartedQuarantinedSpec
           }
         }
 
+        expectMsgPF(10 seconds) {
+          case AssociationHandle.ActorSystemQuarantinedEvent(local, remote) ⇒
+        }
```
jdevelop: I have added this expectation just to test that the message is received upon quarantine. Not sure if any further validation is necessary.

jdevelop:
@johanandren I realized that there is the multi-jvm test for the quarantined event, so I altered it to watch for the new event introduced.

johanandren:
Just realized the ticket was [triaged] but there was little discussion about how this should be achieved, so I'll stop feedbacking it for now until someone from the Akka-team says that this way to solve the ticket is a good idea, so that you do not have to spend a lot of time fixing it up in vain.

jdevelop:
Okay, I'll add some description to the ticket and will see. Thanks for the help!

akka-remote/src/main/scala/akka/remote/transport/Transport.scala
```diff
@@ -179,6 +179,9 @@ object AssociationHandle {
   case object Shutdown extends DisassociateInfo
   case object Quarantined extends DisassociateInfo
 
+  case class ActorSystemQuarantinedEvent(localAddress: Address, remoteAddress: Address)
+  case class RemoteShutdownEvent(localAddress: Address, remoteAddress: Address)
```
rkuhn: These are user API and should live directly in `akka.remote`, plus it would be helpful to add ScalaDoc to them.

rkuhn:
Thanks for the initiative! I think this is the right way to go about this, please add docs [here](http://doc.akka.io/docs/akka/2.4.1/scala/remoting.html#Remote_Events) and we can merge it.

jdevelop:
@rkuhn ok, thanks for the hint! Please take a look at the updates in my PR.

akka-remote/src/main/scala/akka/remote/RemotingLifecycleEvent.scala
```diff
@@ -87,6 +87,9 @@ final case class QuarantinedEvent(address: Address, uid: Int) extends RemotingLi
       "from this situation."
 }
 
+@SerialVersionUID(1L)
+final case class ActorSystemQuarantinedEvent(localAddress: Address, remoteAddress: Address)
```
rkuhn: Please also extend RemotingLifecycleEvent and add a descriptive `toString` message. In order to distinguish the meaning from QuarantinedEvent this one should be called something like WeHaveBeenQuarantinedEvent (my naming skills are not perfect today, so if you have a better idea … ). I think it should be logged at warning level and replace the old unstructured log message.

jdevelop: @rkuhn Done, pls take a look

rkuhn:
Looking good, apart from that now there should be two warnings logged when quarantined, right?

jdevelop:
@rkuhn I'm not sure about the logging. If you mean the following code

    case AssociationHandle.Quarantined ⇒
      context.system.eventStream.publish(ThisActorSystemQuarantinedEvent(localAddress, remoteAddress))
      throw InvalidAssociation(
        localAddress,
        remoteAddress,
        InvalidAssociationException("The remote system has quarantined this system. No further associations " +
          "to the remote system are possible until this system is restarted."))

it seems that this exception would cause the actor to terminate, because it's not handled anywhere. So perhaps this one is required for the correct workflow.

rkuhn:
Hmm, trying to figure out what is logged where (and how that other log message gets printed that appears in the test case above your addition) I am becoming less sure of this approach: should this new event not be published from the EndpointManager’s supervisorStrategy instead? /cc @drewhk

jdevelop:
@rkuhn the new event is not supposed to be a drop-in replacement for the current workflow of supervisor actor. The only goal is to get the notification that the current system is quarantined, to the user may take some countermeasurements (restart system, restart app etc). That's why I didn't remove the current "throw an error and let the supervision strategy to deal with it) but added some notification to the event stream.

rkuhn:
Right, but the supervisorStrategy also publishes to the eventStream, so it might be more appropriate to hook it in there. I’ll wait until Endre chimes in.

drewhk:
I would also like to hook into the supervision in the Remoting class instead (since that is where these exceptions bubble up) and put the notification there.

jdevelop:
@drewhk , @rkuhn ok, so just to confirm - the event should be propagated not in the Endpoint, but in Remoting. Anything else?

rkuhn:
Yes, that is the only change, thanks!

jdevelop:
@rkuhn , @drewhk 
I moved the event notification to the Remoting supervisor, also I added the disassociation info to  InvalidAssociation class, so it will be possible to see what exactly went wrong (instead of picking up a string message). Please review.

rkuhn:
OK TO TEST

jdevelop:
@rkuhn  I wonder if this failure means something about my changes:

    java.lang.RuntimeException: akka-remote: Binary compatibility check failed!
	at scala.sys.package$.error(package.scala:27)
	at com.typesafe.tools.mima.plugin.SbtMima$$anonfun$reportErrors$1.apply(SbtMima.scala:77)
 	at com.typesafe.tools.mima.plugin.SbtMima$$anonfun$reportErrors$1.apply(SbtMima.scala:62)
	at scala.collection.immutable.HashMap$HashMap1.foreach(HashMap.scala:224)
	at scala.collection.immutable.HashMap$HashTrieMap.foreach(HashMap.scala:403)
	at com.typesafe.tools.mima.plugin.SbtMima$.reportErrors(SbtMima.scala:62)
	at com.typesafe.tools.mima.plugin.MimaPlugin$$anonfun$mimaReportSettings$4.apply(MimaPlugin.scala:32)
	at com.typesafe.tools.mima.plugin.MimaPlugin$$anonfun$mimaReportSettings$4.apply(MimaPlugin.scala:31)
	at scala.Function1$$anonfun$compose$1.apply(Function1.scala:47)
	at sbt.$tilde$greater$$anonfun$$u2219$1.apply(TypeFunctions.scala:40)

Is it something I have to fix?

drewhk:
It is a false positive about binary compatibility. A problem filter needs to be added to the build.

jdevelop:
@drewhk still not clear - should I do something about this?

rkuhn:
Thanks @jdevelop !

jdevelop:
Perhaps somebody could close the issue #18758

Issue #16395, opened 2014-11-26
`HttpEntity.Strict` toString prints char int values instead of entity text


RichardBradley:
`HttpEntity.toString` is difficult to read in the Akka HTTP preview releases, e.g.

```
HttpResponse(400 Bad Request,List(Date: Tue, 25 Nov 2014 17:26:27 GMT, Server: akka-http/2.3.6),HttpEntity.Strict(application/json; charset=UTF-8,ByteString(123, 10, 32, 32, 34, 114, 101, 113, 117, 101,…
```

In Spray 1.3.1, the entity body was printed as a string: [HttpEntity.NonEmpty.toString](https://github.com/spray/spray/blob/1ce512cbd17380655fe1756b524c7f19dc9a3de3/spray-http/src/main/scala/spray/http/HttpEntity.scala#L81)

Please could we have something similar in Akka HTTP? It makes testing much harder if the error responses from your web server are printed as int char values instead of as an error message.

(Perhaps this needs to be overridden in `HttpEntity.Strict` based on the ContentType rather than in `ByteString` itself, in case the body is a binary type.)

RichardBradley:
Is there any chance of seeing this in the HTTP 1.0 milestone?

I've just had another painful debugging session where my error message looks like
```
Cannot parse response from remote server: HttpEntity.Strict(application/json,ByteString(123, 10, 32, 32, 34, 101, 114, 114, 111, 114, 34, 58, 32, 34, ...
```

Converting this output to a readable string is surprisingly difficult if the ByteString is very long, as the scala compiler won't accept an `int[]` literal over a certain size: "Method code too large".

I'll see if I can make some time to submit a PR for this issue in the next few days.

sirthias:
This ticket is currently not a high-prio ticket for us, so now, there probably won't be any work on it from our side in the immediate future.
However, it appears to be a pretty self-contained thing (at least if we restrict the fix to `HttpEntity.Strict` instead of `ByteString` for now), so I guess you'll have a good change of getting a contribution PR merged, if you like... :)

ktoso:
Isn't this more about making `logRequest` and friends print something nicely, instead of the raw byte string?
We do not want to change the toString on it itself.

ktoso:
Note, there was this very cool toString by @jrudolph that we might want to put on ByteString in any case.

catalin-ursachi:
[...]

PR #19117, opened 2015-12-07
=htc #16395 print decoded text for non-binary strict entities

catalin-ursachi:
See #16395

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
@@ -238,7 +238,7 @@ object HttpEntity {
       if (contentType == this.contentType) this else copy(contentType = contentType)
 
     /**
-     * See [[HttpEntity#withSizeLimit]].#
+     * See [[HttpEntity#withSizeLimit]].
      */
     def withSizeLimit(maxBytes: Long): UniversalEntity =
       if (data.length <= maxBytes) this
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
@@ -24,6 +24,8 @@ import akka.http.scaladsl.util.FastFuture
 import akka.http.javadsl.{ model ⇒ jm }
 import akka.http.impl.util.JavaMapping.Implicits._
 
+import scala.util.control.NonFatal
+
 /**
  * Models the entity (aka "body" or "content) of an HTTP message.
  */
@@ -236,7 +238,7 @@ object HttpEntity {
       if (contentType == this.contentType) this else copy(contentType = contentType)
 
     /**
-     * See [[HttpEntity#withSizeLimit]].
+     * See [[HttpEntity#withSizeLimit]].#
      */
     def withSizeLimit(maxBytes: Long): UniversalEntity =
       if (data.length <= maxBytes) this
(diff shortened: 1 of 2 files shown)


typesafehub-validator:
Hi @catalin-ursachi,

johanandren:
OK TO TEST

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -239,6 +240,21 @@ object HttpEntity {
...
 
     override def productPrefix = "HttpEntity.Strict"
+
+    override def toString = {
+      val dataAsString = contentType match {
+        case _: Binary ⇒
+          data.toString()
+        case nb: NonBinary ⇒
+          val decodedString = data.decodeString(nb.charset.value)
+          if (decodedString.length > 4000)
```
2beaucoup: FTR: `Bytestring.toString` [uses 100](https://github.com/akka/akka/commit/d8ef6b358a8ce53a796ae43042c6b177bd5209de).

catalin-ursachi: Indeed. However, I would argue that it's useful to print significantly more of a (typically) human-readable text entity, than of a (typically unparseable) raw byte-only entity.

catalin-ursachi:
The Jenkins build got the following test failure in FlowConcatAllSpec:

```
[info] - must on cancellation cancel the current open substream and the master stream *** FAILED *** (3 seconds, 51 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during fishForMessage, hint:
```

I can't reproduce it locally, and it seems entirely unrelated to my change, so it's probably just a spurious timeout.

johanandren:
PLS BUILD

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
johanandren: Should it maybe not decode the entire body first but limit the bytes it will make a string out of, imagine a 10mb body, copied into a string just to take the first 4000 of it.

2beaucoup: Thought about this as well. It gets tricky with encodings with more than one byte per char so the last char can get truncated. But maybe that's ok here.

catalin-ursachi: If we were to use 4096 as the byte cut-off, wouldn't that ensure we don't get any mid-character truncation for most (all?) encodings?

johanandren: I'm thinking just 4 extra bytes would cover the longest UTF-8 and UTF-32, is there any "wider" encoding commonly used?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -239,6 +240,21 @@ object HttpEntity {
...
+            decodedString.take(4000) + s" ... (and ${decodedString.length - 4000} more chars)"
```
ktoso: put `4000` in a value, instead of hardcoding it in 3 places please

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
ktoso: Let's take `4096 `

ktoso:
LGTM after comments addressed.

FYI @jrudolph @sirthias

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
2beaucoup: UTF-8 uses one to four bytes per code point.

catalin-ursachi: Hm; so some truncation seems inevitable then, short of making this bit of code more complicated than seems warranted. Should we just treat that as acceptable?

ktoso: Yeha I think that it's good enough. Would like it to be a nice 2-based number though ;)

2beaucoup: `data.take(maxBytes).decodeString(nb.charset.value).dropRight(1)`?

catalin-ursachi: We would, then, become unable to report the number of bytes remaining (since we didn't know how many bytes the last character contained). Is that fine nevertheless?

2beaucoup: I'd just report the total size here.

catalin-ursachi: Great! :)

ktoso:
Looks good, thanks @catalin-ursachi!

rkuhn:
I squashed and merged this one manually. Thanks @catalin-ursachi !

Issue #18567, opened 2015-09-25
Support parsing param as(CsvInt)


ktoso:
It's a common scenario for when passing around a list of IDs and we already support `as(CsvString)`.

Suggested by @2boc https://github.com/akka/akka/pull/18520#commitcomment-13335926

PR #19118, opened 2015-12-07
!htp #18567 Added unmarshallers for CSV integer sequences

catalin-ursachi:
See #18567

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/PredefinedFromStringUnmarshallers.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/PredefinedFromStringUnmarshallers.scala
@@ -9,28 +9,16 @@ import scala.collection.immutable
 trait PredefinedFromStringUnmarshallers {
 
   implicit val byteFromStringUnmarshaller: Unmarshaller[String, Byte] =
-    Unmarshaller.strict[String, Byte] { string ⇒
-      try string.toByte
-      catch numberFormatError(string, "8-bit signed integer")
-    }
+    Unmarshaller.strict(byteFromString)
 
   implicit val shortFromStringUnmarshaller: Unmarshaller[String, Short] =
-    Unmarshaller.strict[String, Short] { string ⇒
-      try string.toShort
-      catch numberFormatError(string, "16-bit signed integer")
-    }
+    Unmarshaller.strict(shortFromString)
 
   implicit val intFromStringUnmarshaller: Unmarshaller[String, Int] =
-    Unmarshaller.strict[String, Int] { string ⇒
-      try string.toInt
-      catch numberFormatError(string, "32-bit signed integer")
-    }
+    Unmarshaller.strict(intFromString)
 
   implicit val longFromStringUnmarshaller: Unmarshaller[String, Long] =
-    Unmarshaller.strict[String, Long] { string ⇒
-      try string.toLong
-      catch numberFormatError(string, "64-bit signed integer")
-    }
+    Unmarshaller.strict(longFromString)
 
   val HexByte: Unmarshaller[String, Byte] =
     Unmarshaller.strict[String, Byte] { string ⇒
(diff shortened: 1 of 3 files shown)


johanandren:
OK TO TEST

johanandren:
PLS BUILD

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/PredefinedFromStringUnmarshallers.scala
```diff
@@ -78,11 +66,39 @@ trait PredefinedFromStringUnmarshallers {
...
+    CsvStringSeq.map(_.map(byteFromString))
+
+  val CsvShortSeq: Unmarshaller[String, immutable.Seq[Short]] =
+    CsvStringSeq.map(_.map(shortFromString))
+
+  val CsvIntSeq: Unmarshaller[String, immutable.Seq[Int]] =
+    CsvStringSeq.map(_.map(intFromString))
+
+  val CsvLongSeq: Unmarshaller[String, immutable.Seq[Long]] =
+    CsvStringSeq.map(_.map(longFromString))
```
ktoso: Perhaps we could go with `Csv[String]` and `Csv[Long]` instead of having a `val` for each of the types?

ktoso: We'd take in a marshaller of `String => T`, this way it could even be used to make `Csv[Person]` if a `String => Person` marshaller is available.

2beaucoup: :+1:

catalin-ursachi: The problem is with this bit:
```scala
  val CsvStringSeq: Unmarshaller[String, immutable.Seq[String]] =
     Unmarshaller.strict[String, immutable.Seq[String]] { string ⇒
       string.split(",").toList
     }
```
This is not, in fact a proper CSV unmarshaller; the only CSV sequences it can decode are those that don't include commas. So, for example, a CSV-encoded sequence of Point objects might look like:

`"5,3","0,0","0,1"`

But the unmarshaller above would split it to:

```
"5
3"
"0
0"
"0
1"
```

This is obviously undesirable, so even having a so-called 'CsvString' unmarshaller that does the above is questionable, but perhaps tolerable if we expect the user to exercise due caution. Having 'CsvIntSeq' is no more problematic, as Ints certainly don't contain commas.

Having a generic CSV unmarshaller that can accept any arbitrary element unmarshaller, but that will in fact fail for a wide range of element formats, is probably a very bad idea.

So, short of adding a full-fledged CSV unmarshalling library - probably not _that_ much effort - I'm not sure it's possible to take it far beyond this.

2beaucoup: > This is not, in fact, a proper CSV unmarshaller; ...

Let's just call it "good enough for typical form data" then. ;) If you need fully fledged csv parsing then you can still plug in your custom unmarshaller, right? I don't think that form fields are a good way to submit nested values, though.

ktoso: +1, this is not meant to "solve all the things". This is meant to solve the typical: `ids=1,2,3,4` or `names=bob,me,myself`, the numeric one allowing to be Int/Long/... etc

rkuhn:
LGTM, thanks @catalin-ursachi !

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/PredefinedFromStringUnmarshallers.scala
2beaucoup: Was @ktoso's proposal dismissed?

rkuhn: Well, I read @ktoso’s latest comment as “yeah, I’m fine with the current one”, plus I agree with that sentiment.

2beaucoup: I'm referring to the first two comments. The discussion derailed a little after them.

PR #19121, opened 2015-12-08
=str remove unused buffer pool

2m:


drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
```diff
@@ -87,8 +87,6 @@ private[akka] class InputStreamPublisher(is: InputStream, bytesReadPromise: Prom
 
   /** BLOCKING I/O READ */
   def loadChunk() = try {
-    val arr = Array.ofDim[Byte](chunkSize)
```
viktorklang: This is *NOT* safe, see that the code uses ByteString1C which points to the array itself: https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/util/ByteString.scala#L106

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
```diff
@@ -37,7 +37,7 @@ private[akka] class InputStreamPublisher(is: InputStream, bytesReadPromise: Prom
 
   import InputStreamPublisher._
 
-  val buffs = new DirectByteBufferPool(chunkSize, maxBuffer)
```
viktorklang: Deleting this makes a ton of sense since it isn't used though!

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
viktorklang: As for the code:

~~~scala
if (readBytes == chunkSize) availableChunks :+= ByteString1C(arr)
         else availableChunks :+= ByteString1C(arr).take(readBytes)
~~~

I'd probably write it as:

~~~scala
availableChunks :+= ByteString1C(arr).take(readBytes)
~~~

Because a `take` of the full length (i.e. `readBytes` == `chunkSize`) returns the ByteString instance itself (no-op), and anything where `readBytes` is smaller than `chunkSize` will do the right thing.

viktorklang: Also, the following line: `log.debug("No more bytes available to read (got `-1` or `0` from `read`), marking final bytes of file @ " + eofReachedAtOffset)` is not true since the match only checks for `-1`.

A read of `0` should and could most definitely be optimized by the match-case.

drewhk: Ah, so wrapping an array in a ByteString does not copy the array...

viktorklang: @drewhk Yes, it copies the array. `ByteString1C` however doesn't. (ByteString1C is an internal API)

drewhk: ok, I didn't see that ByteString1C was used. Isn't then simple to just use ByteString.apply on the shared array, which will do the proper copying in turn.

viktorklang: @drewhk The difference is that we'd retain the shared array between reads with 0 additional benefit (besides not using an internal API)

2m: I see. I will add a comment here instead.

viktorklang:
:+1:

ktoso:
LGTM

PR #19124, opened 2015-12-08
=str remove unneeded test marker

ktoso:
Remove the xoxo marker (which we used to run only this test) when debugging it.


viktorklang:
:+1: :shipit:

ktoso:
:ship: it

Issue #19058, opened 2015-12-01
Make Idle Timeout log DEBUG instead of exception in HTTP


ktoso:
As discused in https://github.com/akka/akka/issues/18982#issuecomment-160931684

kkwak:
Would it be possible to omit the stacktrace when moving it to INFO?

danielwegener:
I wonder if this should not even be a DEBUG log. I think it is a pretty normal behaviour that a http connection times out and since info is the default log level - well I would not bother a user with such information that occurs for possibly each user-agent.

ktoso:
Yeah DEBUG might be good. Would one of you want to contribute this fix? :)

danielwegener:
Yeah - I like low hanging fruits :grapes:  :)

quick fix without test (sbt test does not run on this machine right now) c2bc58 , however, I'm wondering if the timeout exception from the IdleBidi should be explicitly marked as TCP socket timeout exception (maybe with the original TimeoutException as cause). Otherwise we cannot really decide if the TimeoutException comes from the TCP Connection or some other point in the flow (might even come from the users route)

ktoso:
Hm, it may be good to lift that timeout error into a special internal exception type, and only that one handle like you did, you'll need to poke around a bit to make sure it is done properly (in both directions)

ktoso:
The conversion you can do with `.recover { case _ : TimeoutException => throw HttpTimeoutException }` right after the idletimeout

ktoso:
Revisit for 2.4.3

MasseGuillaume:
@ktoso 

It's still logged as an `ERROR`

```
webappJVM [ERROR] [04/24/2016 11:27:05.350] [scalakata-playground-akka.actor.default-dispatcher-30] [akka://scalakata-playground/user/StreamSupervisor-0/flow-600-0-unknown-operation] Error in stage [recover]: No elements passed in the last 25 seconds. (akka.http.impl.engine.HttpConnectionTimeoutException)
webappJVM [ERROR] [04/24/2016 11:27:05.359] [scalakata-playground-akka.actor.default-dispatcher-30] [akka://scalakata-playground/user/StreamSupervisor-0/flow-602-0-unknown-operation] Error in stage [recover]: No elements passed in the last 25 seconds. (akka.http.impl.engine.HttpConnectionTimeoutException)
[...]

PR #19130, opened 2015-12-08
=http #19058 Make idle timeout log DEBUG instead of exception

danielwegener:
Solves #19058

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionTimeoutException.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionTimeoutException.scala
@@ -1,0 +1,9 @@
+/*
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.impl.engine
+
+import scala.util.control.NoStackTrace
+
+class HttpConnectionTimeoutException(msg: String) extends RuntimeException(msg) with NoStackTrace
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -11,6 +11,7 @@ import org.reactivestreams.{ Publisher, Subscriber }
 import scala.util.control.NonFatal
 import akka.event.LoggingAdapter
 import akka.http.ServerSettings
+import akka.http.impl.engine.HttpConnectionTimeoutException
 import akka.http.impl.engine.parsing.ParserOutput._
 import akka.http.impl.engine.parsing._
 import akka.http.impl.engine.rendering.{ HttpResponseRendererFactory, ResponseRenderingContext, ResponseRenderingOutput }
@@ -156,10 +157,16 @@ private[http] object HttpServerBluePrint {
 
     val responseRendererFactory = new HttpResponseRendererFactory(serverHeader, responseHeaderSizeHint, log)
 
+    val errorHandler: Throwable ⇒ Unit = {
+      // idle timeouts should not result in errors in the log. See 19058.
+      case timeout: HttpConnectionTimeoutException ⇒ log.debug(s"Closing HttpConnection due to timeout: ${timeout.getMessage}")
+      case t                                       ⇒ log.error(t, "Outgoing response stream error")
+    }
+
     Flow[ResponseRenderingContext]
       .via(Flow[ResponseRenderingContext].transform(() ⇒ responseRendererFactory.newRenderer).named("renderer"))
       .flatMapConcat(ConstantFun.scalaIdentityFunction)
-      .via(Flow[ResponseRenderingOutput].transform(() ⇒ errorLogger(log, "Outgoing response stream error")).named("errorLogger"))
+      .via(Flow[ResponseRenderingOutput].transform(() ⇒ errorHandling(errorHandler)).named("errorLogger"))
   }
 
   class ControllerStage(settings: ServerSettings, log: LoggingAdapter)
--- a/akka-http-core/src/main/scala/akka/http/impl/util/package.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/package.scala
@@ -93,11 +93,11 @@ package object util {
     }
   }
 
-  private[http] def errorLogger[T](log: LoggingAdapter, msg: String): PushStage[T, T] =
+  private[http] def errorHandling[T](handler: Throwable ⇒ Unit): PushStage[T, T] =
     new PushStage[T, T] {
       override def onPush(element: T, ctx: Context[T]): SyncDirective = ctx.push(element)
       override def onUpstreamFailure(cause: Throwable, ctx: Context[T]): TerminationDirective = {
-        log.error(cause, msg)
+        handler(cause)
         super.onUpstreamFailure(cause, ctx)
       }
     }
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -12,6 +12,7 @@ import javax.net.ssl.{ SSLContext, SSLParameters }
 import akka.actor._
 import akka.event.LoggingAdapter
 import akka.http._
+import akka.http.impl.engine.HttpConnectionTimeoutException
 import akka.http.impl.engine.client._
 import akka.http.impl.engine.server._
 import akka.http.impl.engine.ws.WebsocketClientBlueprint
@@ -27,7 +28,7 @@ import akka.stream.scaladsl._
 import com.typesafe.config.Config
 
 import scala.collection.immutable
-import scala.concurrent.{ ExecutionContext, Future, Promise }
+import scala.concurrent.{ ExecutionContext, Future, Promise, TimeoutException }
 import scala.util.Try
 import scala.util.control.NonFatal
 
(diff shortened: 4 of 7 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionTimeoutException.scala
```diff
@@ -0,0 +1,5 @@
+package akka.http.impl.engine
+
+import scala.util.control.NoStackTrace
+
+class HttpConnectionTimeoutException(msg: String) extends RuntimeException(msg) with NoStackTrace
```
ktoso: Needs license header:

```
/**
  * Copyright (C) 2009-2014 Typesafe Inc. <http://www.typesafe.com>
  */
```

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -154,10 +155,16 @@ private[http] object HttpServerBluePrint {
 
     val responseRendererFactory = new HttpResponseRendererFactory(serverHeader, responseHeaderSizeHint, log)
 
+    val errorHandler: Throwable ⇒ Unit = {
+      // socket timeouts should not result in errors in the log. See 19058.
```
ktoso: It's not really "socket timeout", its "idle timeout" - socket can be perfectly fine, just nothing happening on it.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -154,10 +155,16 @@ private[http] object HttpServerBluePrint {
...
+      case timeout: HttpConnectionTimeoutException ⇒ log.debug(s"Closing HttpConnection due to timeout: ${timeout.getMessage}")
+      case t                                       ⇒ log.error(t, "Outgoing response stream error")
```
ktoso: why not don't to anything to other errors here? So it bubbles through downstream, at which point it'll log (like it does now?)

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -154,10 +155,16 @@ private[http] object HttpServerBluePrint {
 
     val responseRendererFactory = new HttpResponseRendererFactory(serverHeader, responseHeaderSizeHint, log)
 
+    val errorHandler: Throwable ⇒ Unit = {
+      // socket timeouts should not result in errors in the log. See 19058.
+      case timeout: HttpConnectionTimeoutException ⇒ log.debug(s"Closing HttpConnection due to timeout: ${timeout.getMessage}")
```
ktoso: Good :)

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
```diff
@@ -101,11 +101,11 @@ package object util {
     }
   }
 
-  private[http] def errorLogger[T](log: LoggingAdapter, msg: String): PushStage[T, T] =
+  private[http] def errorHandling[T](handler: Throwable ⇒ Unit): PushStage[T, T] =
```
ktoso: good

akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionTimeoutException.scala
ktoso: very good about the with NoStackTrace btw.
I wonder if it could be made an object even hmm

akka-stream/src/main/scala/akka/stream/impl/Timers.scala
```diff
@@ -3,8 +3,9 @@
  */
 package akka.stream.impl
 
-import java.util.concurrent.{ TimeUnit, TimeoutException }
+import java.util.concurrent.TimeUnit
 
+import scala.concurrent.TimeoutException
 import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
 import akka.stream._
 import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler, TimerGraphStageLogic }
```
ktoso: would be nice to not touch this file, since you're not changing anything in it.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
danielwegener: I just wanted to keep the current behaviour (see `errorLogger(log, "Outgoing response stream error")).named("errorLogger"))`)

ktoso: Okey, got it (was reviewing top down, without looking how this val was used).
Indeed looks OK

akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
```diff
@@ -175,10 +175,13 @@ class ClientServerSpec extends WordSpec with Matchers with BeforeAndAfterAll {
...
-            intercept[StreamTcpException] {
-              Await.result(clientsResponseFuture, 2.second)
-            }
+            EventFilter.debug("Closing HttpConnection due to timeout: No elements passed in the last 300 milliseconds.", occurrences = 1) intercept ({
+              // waiting for the timeout to happen on the client
+              intercept[StreamTcpException] {
+                Await.result(clientsResponseFuture, 2.second)
+              }
+            }, system)
+
```
ktoso: thanks for the test, looks OK

akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionTimeoutException.scala
danielwegener: That was bady copy&paste programming :) TokenSourceActor.scala has no license header either

ktoso: Whoop that's bad :) Can you please add it there too? Thanks!

akka-stream/src/main/scala/akka/stream/impl/Timers.scala
danielwegener: Aye, will be reverted!

akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionTimeoutException.scala
danielwegener: Since it exists since 2014 (45ddb7b), does it need a 2009-2014 header? (nitpicking :))

ktoso: Technically it should be "2009-2015", and we'll update the headers to be "2009-2016" once new years hits the calendar.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
danielwegener: I think this is because failures in the connection flow are at some point swallowed so they do not kill the binding flow (think i have seen such a line)

ktoso:
OK TO TEST

ktoso:
I poked around the logging and I don't think we have a simple way to silence the output while simultaneously. Or did we do such a thing somewhere already // cc @rkuhn @patriknw @drewhk @2m (See this question https://github.com/akka/akka/pull/19130#commitcomment-14854418 )?

rkuhn:
`EventFilter(...).intercept { ... }` is the feature you are looking for, especially in combination with the `occurrences` parameter which allows you to assert that these messages actually are generated.

danielwegener:
Yeah thats what I am using - but to expect a debug messages, I had to change `akka.loglevel` from `ERROR` to `DEBUG`. The intercept/expect works fine but all other log messages (debug, info, and warning) are written to stdout. Thats what I ment with 'polluting the sbt test output'. I also thought about installing another global EventFilter for `ClientServerSpec` that silently consumes all other log messages withlLevel<ERROR that do _not_ match the expected log-message - but that feels a bit hacky.

I would also be fine to remove the debug-logging test again - or keep the verbose sbt test output. WDYT?

danielwegener:
(Or `TestEventListener` could lern to respect `akka.stdout-loglevel`  )

akka-http-core/src/main/scala/akka/http/impl/engine/HttpConnectionTimeoutException.scala
rkuhn: Well, the 2009 part was the question, and I don’t see a reason to start the range with an earlier year than the inception of the file it heralds (i.e. it should only say `2015`, no range).

ktoso: Hm okey; may be a bit difficult to apply retroactively but we can do so from now on I guess.

rkuhn:
Concerning the test: I’d rather assert that no ERROR is logged anymore and ignore (i.e. disable) the DEBUG output.

Otherwise LGTM.

danielwegener:
Uhm - Using `EventFilter` to test that something is _not_ logged does not really work, does it? I'd rather remove the EventFilter assert then and see it as positive outcome that there is one less `[ERROR]`-line on sbt test output.

rkuhn:
Does it not do the right thing if you set `occurrences=0`? Otherwise I'm fine with just removing the check.

danielwegener:
Squashed and rebased.

rkuhn:
Thanks @danielwegener !

ktoso:
Great, thank you!

jrudolph:
I'm still getting those exceptions in 2.4.2.

```
backend [ERROR] [02/29/2016 14:48:57.938] [default-akka.actor.default-dispatcher-63] [akka://default/user/StreamSupervisor-0/flow-90-0-unknown-operation] Error in stage [recover]: No elements passed in the last 1 minute. (akka.http.impl.engine.HttpConnectionTimeoutException)
```

I wonder if I'm doing something strange or if it really is still happening, I'll have another try at reproducing it from the akka repository directly.

ngbinh:
we are seeing the same thing on akka `2.4.2`

jrudolph:
Thanks for confirming, @ngbinh.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -71,7 +72,8 @@ class HttpExt(config: Config)(implicit system: ActorSystem) extends akka.actor.E
     connections.map {
       case Tcp.IncomingConnection(localAddress, remoteAddress, flow) ⇒
         val layer = serverLayer(settings, Some(remoteAddress), log)
-        IncomingConnection(localAddress, remoteAddress, layer atop tlsStage join flow)
+        val flowWithTimeoutRecovered = flow.recover { case t: TimeoutException ⇒ throw new HttpConnectionTimeoutException(t.getMessage) }
```
jrudolph: It still logs an error because the code throws out of the `recover` blog. What would be needed is a `mapFailure` combinator instead.

Issue #19072, opened 2015-12-02
errors from FlattenMerge during WS client Autobahn tests


ktoso:
I'm setting up Autobahn tests on jenkins, and on the first run noticed some errors being printed from FlattenMerge.
It would seem they're harmless, the tests seem to have passed, though perhaps worth investigating nevertheless:

```
OK                64 ms 9.2.1   - Send binary message message with payload of length 64 * 2**10 (64k). ...
OK                92 ms 9.2.2   - Send binary message message with payload of length 256 * 2**10 (256k). ...
[ERROR] [12/02/2015 19:23:54.513] [default-akka.test.stream-dispatcher-10] [akka://default/user/StreamSupervisor-0/flow-2416-31-ws-message-api-ws-prepare-messages] Error during postStop in [akka.stream.impl.fusing.FlattenMerge@182939a5]
java.lang.IllegalStateException: not yet initialized: only setHandler is allowed in GraphStageLogic constructor
	at akka.stream.stage.GraphStageLogic.interpreter(GraphStage.scala:321)
	at akka.stream.stage.GraphStageLogic$$anon$3.invoke(GraphStage.scala:837)
	at akka.stream.impl.fusing.LocalSink$Logic$1$$anonfun$4.apply(StreamOfStreams.scala:167)
	at akka.stream.impl.fusing.LocalSink$Logic$1$$anonfun$4.apply(StreamOfStreams.scala:167)
	at akka.stream.impl.fusing.LocalSinkSubscription.cancel(StreamOfStreams.scala:142)
	at akka.stream.impl.fusing.FlattenMerge$$anon$1$$anonfun$postStop$1.apply(StreamOfStreams.scala:129)
	at akka.stream.impl.fusing.FlattenMerge$$anon$1$$anonfun$postStop$1.apply(StreamOfStreams.scala:128)
	at scala.collection.immutable.Set$Set1.foreach(Set.scala:74)
	at akka.stream.impl.fusing.FlattenMerge$$anon$1.postStop(StreamOfStreams.scala:128)
	at akka.stream.impl.fusing.GraphInterpreter.finalizeStage(GraphInterpreter.scala:630)
	at akka.stream.impl.fusing.GraphInterpreter.afterStageHasRun(GraphInterpreter.scala:607)
	at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:515)
[...]

PR #19098, opened 2015-12-05
=str - Adds finals and reorganizes some StreamOfStreams code

viktorklang:

PR #19132, opened 2015-12-09
=str #19072 fix FlattenMerge init race

rkuhn:


akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
```diff
@@ -190,7 +190,8 @@
     public static MediaType custom(String value, boolean binary, boolean compressible) {
         akka.http.scaladsl.model.MediaType.Compressibility comp = compressible ?
                 akka.http.scaladsl.model.MediaType.Compressible$.MODULE$ : akka.http.scaladsl.model.MediaType.NotCompressible$.MODULE$;
-        return akka.http.scaladsl.model.MediaType.custom(value, binary, comp , List.<String>empty());
+        return akka.http.scaladsl.model.MediaType.custom(value, binary, comp , 
+            akka.http.scaladsl.model.MediaType.custom$default$4());
```
rkuhn: `List.<String> empty()` is ambiguous in the eyes of the JDT … (/cc @sirthias)

sirthias: What's the JDT? And what kind of ambiguity does it see?

drewhk: It is the Eclipse Java compiler.

rkuhn: I don’t know what it “sees”, but it fails compilation.

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -145,29 +144,56 @@ private[fusing] final class LocalSinkSubscription[T](sub: ActorPublisherMessage
...
+  private var subF: Future[LocalSinkSubscription[T]] = _
+  private var sub: LocalSinkSubscription[T] = _
+
+  var elem: T = null.asInstanceOf[T]
+
+  def isActive: Boolean = sub ne null
+  def deactivate(): Unit = sub = null
+  def activate(f: Future[LocalSinkSubscription[T]]): Unit = {
+    subF = f
+    f.foreach(s ⇒ sub = s)(sameThreadExecutionContext)
```
viktorklang: This looks extremely shady

rkuhn: Good, I hate sunburns. Shall I add a comment that sheds some artificial light?

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -145,29 +144,56 @@ private[fusing] final class LocalSinkSubscription[T](sub: ActorPublisherMessage
...
  */
-private[fusing] final class LocalSource[T](var sub: LocalSinkSubscription[T] = null, var elem: T = null.asInstanceOf[T])
+private[fusing] final class LocalSource[T] {
+  private var subF: Future[LocalSinkSubscription[T]] = _
+  private var sub: LocalSinkSubscription[T] = _
+
+  var elem: T = null.asInstanceOf[T]
+
+  def isActive: Boolean = sub ne null
+  def deactivate(): Unit = sub = null
```
viktorklang: Should this also null out the subF so it doesn't hang onto a dead ref?

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
viktorklang: Absolutely. What guarantees that `sub` is visible to the other threads when the completing thread has executed the callback?

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
rkuhn: Yes, it should, thanks!

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -145,29 +144,56 @@ private[fusing] final class LocalSinkSubscription[T](sub: ActorPublisherMessage
...
+  def activate(f: Future[LocalSinkSubscription[T]]): Unit = {
+    subF = f
+    f.foreach(s ⇒ sub = s)(sameThreadExecutionContext)
+  }
+
+  def pull(): Unit = {
+    if (sub eq null)
+      throw new IllegalStateException("not yet initialized, subF=" + subF.value)
```
ktoso: Message could be made a bit less dodgy? `subF=...` is very weird.

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
viktorklang: :+1:

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -145,29 +144,56 @@ private[fusing] final class LocalSinkSubscription[T](sub: ActorPublisherMessage
...
+    sub.pull()
+  }
+
+  def cancel(): Unit = subF.foreach(_.cancel())(sameThreadExecutionContext)
```
viktorklang: Note that `cancel` does not modify `subF` nor `sub`.
Also I believe there to be a potential race-condition between `cancel` and `activate` (as callbacks can run in any order).

rkuhn: No, LocalSource is only used from within the FlattenMerge logic, which is a sequential place.

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
rkuhn: okay …

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
viktorklang: Now yes, but if those classes are ever reused by any other logic now all bets are off?

rkuhn:
merged so that #19098 can work on top of it

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
rkuhn: This is why this is an implementation detail of FlattenMerge, which you are about to properly express by moving things into a tighter namespace (thanks!).

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -145,29 +144,70 @@ private[fusing] final class LocalSinkSubscription[T](sub: ActorPublisherMessage
...
+  def activate(f: Future[LocalSinkSubscription[T]]): Unit = {
+    subF = f
+    /*
+     * The subscription is communicated to the FlattenMerge stage by way of completing
+     * the future. Encoding it like this means that the `sub` field will be written
+     * either by us (if the future has already been completed) or by the LocalSink (when
+     * it eventually completes the future in its `preStart`). The important part is that
+     * either way the `sub` field is populated before we get the first `OnNext` message
+     * and the value is safely published in either case as well (since AsyncCallback is
+     * based on an Actor message send).
```
ktoso: much needed comment, thanks for adding it :+1:

PR #19138, opened 2015-12-09
=str - 18988 - Make AcknowledgeSinkSpec more robust

viktorklang:
#18988

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
@@ -21,18 +21,15 @@ class AcknowledgeSinkSpec extends AkkaSpec {
 
   val noMsgTimeout = 300.millis
 
-  def assertSuccess(value: Any, fb: Future[Option[Any]]): Unit =
-    Await.result(fb, 1.second) should be(Some(value))
-
   "An AcknowledgeSink" must {
 
     "send the elements as result of future" in assertAllStagesStopped {
-      val queue = Source(List(1, 2, 3)).runWith(Sink.queue(3))
-      assertSuccess(1, queue.pull())
-      assertSuccess(2, queue.pull())
-      assertSuccess(3, queue.pull())
-      queue.pull().pipeTo(testActor)
-      expectMsg(None)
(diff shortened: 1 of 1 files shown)


johanandren:
LGTM!

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
```diff
@@ -21,18 +21,15 @@ class AcknowledgeSinkSpec extends AkkaSpec {
...
-      assertSuccess(2, queue.pull())
-      assertSuccess(3, queue.pull())
-      queue.pull().pipeTo(testActor)
-      expectMsg(None)
+      val expected = List(Some(1), Some(2), Some(3), None)
+      val queue = Source(expected.flatten).runWith(Sink.queue(3))
+      expected foreach { v ⇒
+        queue.pull() pipeTo testActor
+        expectMsg(v)
+      }
```
rkuhn: AFAICS this does not make the test more reliable—it already expresses the right thing—but it is certainly an improvement.

rkuhn:
LGTM, but I’m not yet convinced that this fixes #18988—but that test should have been fixed already by #19097.

rkuhn:
Or rather: it would be fixed if that other PR was already merged …

akka-stream-tests/src/test/scala/akka/stream/scaladsl/AcknowledgeSinkSpec.scala
viktorklang: It changes the timeout.

-- 
Cheers,
√

rkuhn: Yes, but the timeout was not the issue: until #19097 is merged this will fail sporadically.

viktorklang: Ok, so now it is both fixed and the test more robust. sounds like winwin to
me!  😊

-- 
Cheers,
√

rkuhn: Yes :-)

Issue #19021, opened 2015-11-26
add unfold / unfoldM Source generators to akka stream


hochgi:
as discussed on the mailing list: 
https://groups.google.com/forum/#!topic/akka-user/2HkQIJIzvXU

unfold & unfoldM are very useful, and it would be great to have it implemented.

also, there's a POC implementation which seems to be working:
https://gist.github.com/hochgi/cbe5ffc6cf2915e31091

etaty:
https://groups.google.com/forum/#!topic/akka-user/T8WurGezKQk

So i am probably going to create a pull request for that with GraphStage[SourceShape[Out]] in the next days

etaty:
So there is a version using GraphStage[SourceShape[Out]]

https://gist.github.com/etaty/2eb5848f9cc3b10c9a1f

PR #19143, opened 2015-12-10
+str #19021 Add unfold (and variants) generators to Source

hochgi:
see #19021

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
@@ -219,6 +219,44 @@ class SourceSpec extends AkkaSpec {
     }
   }
 
+  "Unfold Source" must {
+    val expected = List(9227465, 5702887, 3524578, 2178309, 1346269, 832040, 514229, 317811, 196418, 121393, 75025, 46368, 28657, 17711, 10946, 6765, 4181, 2584, 1597, 987, 610, 377, 233, 144, 89, 55, 34, 21, 13, 8, 5, 3, 2, 1, 1, 0)
+
+    "generate a finite fibonacci sequence" in {
+      import GraphDSL.Implicits._
+
+      val source = Source.unfold((0, 1)) {
+        case (a, _) if a > 10000000 ⇒ None
+        case (a, b)                 ⇒ Some((b, a + b) → a)
+      }
+      val result = Await.result(source.runFold(List.empty[Int]) { case (xs, x) ⇒ x :: xs }, 1.second)
+      result should ===(expected)
+    }
+
+    "generate a finite fibonacci sequence asynchronously" in {
+      import GraphDSL.Implicits._
+      import scala.concurrent.Future
+      import scala.concurrent.ExecutionContext.Implicits.global
+
+      val source = Source.unfoldAsync((0, 1)) {
+        case (a, _) if a > 10000000 ⇒ Future.successful(None)
+        case (a, b)                 ⇒ Future(Some((b, a + b) → a))
+      }
+      val result = Await.result(source.runFold(List.empty[Int]) { case (xs, x) ⇒ x :: xs }, 1.second)
+      result should ===(expected)
+    }
+
+    "generate an infinite fibonacci sequence" in {
+      import GraphDSL.Implicits._
+
+      val source = Source.unfoldInf((0, 1)) {
+        case (a, b) ⇒ (b, a + b) → a
+      }
+      val result = Await.result(source.take(36).runFold(List.empty[Int]) { case (xs, x) ⇒ x :: xs }, 1.second)
+      result should ===(expected)
+    }
+  }
+
   "Iterator Source" must {
     "properly iterate" in {
       val result = Await.result(Source(() ⇒ Iterator.iterate(false)(!_)).grouped(10).runWith(Sink.head), 1.second)
--- a/akka-stream/src/main/scala/akka/stream/impl/Unfold.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/Unfold.scala
@@ -1,0 +1,79 @@
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.scaladsl
+
+import akka.stream.stage.{ OutHandler, GraphStageLogic, GraphStage }
+import akka.stream._
+
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.util.{ Failure, Success, Try }
+
+/**
+ * Unfold `GraphStage` class
+ * @param s initial state
+ * @param f unfold function
+ * @tparam S state
+ * @tparam E element
+ */
+private[akka] class Unfold[S, E](s: S, f: S ⇒ Option[(S, E)]) extends GraphStage[SourceShape[E]] {
+
+  val out: Outlet[E] = Outlet("Unfold")
+
+  override val shape: SourceShape[E] = SourceShape(out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = {
+    new GraphStageLogic(shape) {
+      private[this] var state = s
+
+      setHandler(out, new OutHandler {
+        override def onPull(): Unit = f(state) match {
+          case None ⇒ complete(out)
+          case Some((newState, v)) ⇒ {
+            push(out, v)
+            state = newState
+          }
+        }
+      })
+    }
+  }
+}
+
+/**
+ * UnfoldAsync `GraphStage` class
+ * @param s initial state
+ * @param f unfold function
+ * @tparam S state
+ * @tparam E element
+ */
+private[akka] class UnfoldAsync[S, E](s: S, f: S ⇒ Future[Option[(S, E)]]) extends GraphStage[SourceShape[E]] {
+
+  val out: Outlet[E] = Outlet("UnfoldAsync")
+
+  override val shape: SourceShape[E] = SourceShape(out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = {
+    new GraphStageLogic(shape) {
+      private[this] var state = s
+
+      private[this] var asyncHandler: Function1[Try[Option[(S, E)]], Unit] = _
+
+      override def preStart() = {
+        val ac = getAsyncCallback[Try[Option[(S, E)]]] {
+          case Failure(ex)   ⇒ fail(out, ex)
+          case Success(None) ⇒ complete(out)
+          case Success(Some((newS, elem))) ⇒ {
+            push(out, elem)
+            state = newS
+          }
+        }
+        asyncHandler = ac.invoke
+      }
+
+      setHandler(out, new OutHandler {
+        override def onPull(): Unit =
+          f(state).onComplete(asyncHandler)(akka.dispatch.ExecutionContexts.sameThreadExecutionContext)
+      })
+    }
+  }
+}
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
@@ -143,6 +143,26 @@ object Source {
     new Source(scaladsl.Source.repeat(element))
 
   /**
+   * create a `Source` that will unfold a value of type `S` into
+   * a pair of the next state `S` and output elements of type `E`.
+   */
+  def unfold[S, E](s: S, f: function.Function[S, Option[(S, E)]]): Source[E, Unit] =
+    new Source(scaladsl.Source.unfold(s)((s: S) ⇒ f.apply(s)))
+
+  /**
+   * same as unfold, but uses an async function to generate the next state-element tuple.
+   */
+  def unfoldAsync[S, E](s: S, f: function.Function[S, Future[Option[(S, E)]]]): Source[E, Unit] =
+    new Source(scaladsl.Source.unfoldAsync(s)((s: S) ⇒ f.apply(s)))
+
+  /**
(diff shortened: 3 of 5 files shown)


ktoso:
OK TO TEST

johanandren:
I don't think `Unfold` should live in the `scaladsl` package, all other stages lives under `impl` or `fusing`. And it needs some test coverage.

LGTM except for that

hochgi:
@johanandren I really didn't know where to put it. where do you think would be best? and should I add it as a new file or append to e.g: `Graph.scala`

johanandren:
I think you should put it under `akka.stream.impl` and make it `private[akka]` so that we have a chance to move it later rather than it becoming a public api - the factory methods are basically the public api of it.

hochgi:
it seems like the tests failed on something not related: `groupBy`. anyway, all the tests passed for me.

rkuhn:
Indeed, the FlowGroupBySpec has some wrong ordering assumptions built in, now that `mergeSubstreams` actually is a merge …

johanandren:
PLS BUILD

johanandren:
Ah, or FlowGroupBy is consistently broken, not just racing? (Sorry for triggering another failed build in that case). Ticket for failing #19145

akka-stream/src/main/scala/akka/stream/impl/Unfold.scala
```diff
@@ -0,0 +1,77 @@
...
+ */
+private[akka] class UnfoldAsync[S, E](s: S, f: S ⇒ Future[Option[(S, E)]])(implicit ec: ExecutionContext) extends GraphStage[SourceShape[E]] {
+
+  val out: Outlet[E] = Outlet("UnfoldAsync")
+
+  override val shape: SourceShape[E] = SourceShape(out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = {
+    new GraphStageLogic(shape) {
+      private[this] var state = s
```
johanandren: Why is it `private[this]` here but not in `Unfold`?

akka-stream/src/main/scala/akka/stream/impl/Unfold.scala
```diff
@@ -0,0 +1,77 @@
...
+
+      private[this] val asyncHandler: Function1[Try[Option[(S, E)]], Unit] = {
+        val ac = getAsyncCallback[Try[Option[(S, E)]]] {
```
johanandren: It's not safe to call `getAsyncCallback` in the constructor (could create race conditions), move it to `preStart`.

rkuhn:
FlowGroupBySpec is definitely broken, `.groupBy` should be fine. I’m working on it.

hochgi:
@johanandren thanks for feedback. I applied the changes you suggested :)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
```diff
@@ -219,6 +219,47 @@ class SourceSpec extends AkkaSpec {
...
+
+    "generate a finite fibonacci sequence asynchronously" in {
+      import GraphDSL.Implicits._
+      import scala.concurrent.Future
+      import scala.concurrent.ExecutionContext.Implicits.global
+
+      val source = Source.unfoldAsync(0 → 1) {
+        case (a, _) if a > 10000000 ⇒ Future.successful(None)
+        case (a, b) ⇒ Future {
+          Thread.sleep(10)
```
johanandren: Why do we need a sleep here, the operations are anyways serialized no?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
```diff
@@ -219,6 +219,47 @@ class SourceSpec extends AkkaSpec {
     }
   }
 
+  "Unfold Source" must {
+    val expected = List(9227465, 5702887, 3524578, 2178309, 1346269, 832040, 514229, 317811, 196418, 121393, 75025, 46368, 28657, 17711, 10946, 6765, 4181, 2584, 1597, 987, 610, 377, 233, 144, 89, 55, 34, 21, 13, 8, 5, 3, 2, 1, 1, 0)
+
+    "generate a finite fibonacci sequence" in {
+      import GraphDSL.Implicits._
+
+      val source = Source.unfold(0 → 1) {
```
johanandren: Maybe just a matter of taste, but I find the arrow-syntax for tuples really confusing here since it isn't a key -> value relation but rather (n, nextN).

akka-stream/src/main/scala/akka/stream/impl/Unfold.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  val out: Outlet[E] = Outlet("UnfoldAsync")
+
+  override val shape: SourceShape[E] = SourceShape(out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = {
+    new GraphStageLogic(shape) {
+      private[this] var state = s
+
+      private[this] var asyncHandler: Function1[Try[Option[(S, E)]], Unit] = _
```
johanandren: Another matter of taste, why not keep the `AsyncCallback` instead of lifting a method to a function (one extra allocation?)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
```diff
@@ -219,6 +219,47 @@ class SourceSpec extends AkkaSpec {
...
+      result should ===(expected)
+    }
+
+    "generate a finite fibonacci sequence asynchronously" in {
+      import GraphDSL.Implicits._
+      import scala.concurrent.Future
+      import scala.concurrent.ExecutionContext.Implicits.global
+
+      val source = Source.unfoldAsync(0 → 1) {
+        case (a, _) if a > 10000000 ⇒ Future.successful(None)
```
johanandren: Do we really need to do this 10 000 000 times before we are sure it works as expected?

akka-stream/src/main/scala/akka/stream/impl/Unfold.scala
rkuhn: This is actually an optimization: instead of allocating an eta expansion object every time, we just allocate it once.

johanandren: Ah, I see. (Thanks for clarifying)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
hochgi: it's not 10M times, it's the fibonacci number under 10M, which means it's actually 36 times.

johanandren: Sorry, my bad, of course!

johanandren:
LGTM!

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
hochgi: you're right. not needed, just to clarify, but i'll remove it :)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
hochgi: consider the line: `case (a, b) ⇒ (b → (a + b)) → a` with the regular tuple syntax, it would look like: `case (a, b) ⇒ ((b,(a+b)),a)`, the arrow syntax saves parenthesis in an already heavy parenthesis expression. for the rest of the test, it's more a matter of uniformity. In any case, if you still think I should change the code to use the regular tuple syntax, I will change it.

rkuhn: I’m with @johanandren on this one (this particular line). The line below would look like
~~~scala
case (a, b) => Some((b, a + b) -> a)
~~~

rkuhn: The rationale for changing the seed value’s presentation is that it currently does not match the extraction in the following lines (which is the confusing part).

akka-stream/src/main/scala/akka/stream/impl/Unfold.scala
```diff
@@ -0,0 +1,79 @@
...
+            push(out, elem)
+            state = newS
+          }
+        }
+        asyncHandler = ac.invoke
+      }
+
+      setHandler(out, new OutHandler {
+        override def onPull(): Unit =
+          f(state).onComplete(asyncHandler)(ec)
```
rkuhn: invoking an AsyncCallback is quick and non-blocking, so we might as well use the ExecutionContexts.sameThreadExecutionContext and unburden the user of the API.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
hochgi: OK, fixed :)

akka-stream/src/main/scala/akka/stream/impl/Unfold.scala
hochgi: it makes sense. done :)

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -240,6 +240,64 @@ object Source {
...
+
+      val uzip = b.add(UnzipWith(f))
+      val cnct = b.add(Concat[S]())
+      val init = Source.single(s)
+
+      init ~> cnct ~> uzip.in
+      cnct <~ uzip.out0
+
+      SourceShape(uzip.out1)
+    })
```
rkuhn: Please add `DefaultAttributes.*` entries for these three new sources and use them here (to name things).

rkuhn:
Test failure is unrelated and fixed now, please add the `named()` attributes and squash everything into a single commit, then `git push -f` and after validation we can merge.

rkuhn:
Great, thanks @hochgi !

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -143,6 +143,26 @@ object Source {
     new Source(scaladsl.Source.repeat(element))
 
   /**
+   * create a `Source` that will unfold a value of type `S` into
+   * a pair of the next state `S` and output elements of type `E`.
```
ktoso: Why lower case? looks weird.

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -240,6 +240,64 @@ object Source {
...
   /**
+   * create a `Source` that will unfold a value of type `S` into
+   * a pair of the next state `S` and output elements of type `E`.
+   *
+   * for example, all the fibonacci numbers under 10M:
+   *
+   * {{{
+   *   Source.unfold(0 → 1){
+   *    case (a,_) if a > 10000000 ⇒ None
+   *    case (a,b) ⇒ Some((b → (a + b)) → a)
```
ktoso: `case (a, b)` to be consistent with our usual formatting?

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -240,6 +240,64 @@ object Source {
   }
 
   /**
+   * create a `Source` that will unfold a value of type `S` into
+   * a pair of the next state `S` and output elements of type `E`.
+   *
+   * for example, all the fibonacci numbers under 10M:
+   *
+   * {{{
+   *   Source.unfold(0 → 1){
```
ktoso: with space before `{` like so  `Source.unfold(0 → 1) {` please

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -240,6 +240,64 @@ object Source {
...
+   *      Some((b → (a + b)) → a)
+   *    }
+   *   }
+   * }}}
+   */
+  def unfoldAsync[S, E](s: S)(f: S ⇒ Future[Option[(S, E)]]): Source[E, Unit] =
+    Source.fromGraph(new UnfoldAsync(s, f)).withAttributes(DefaultAttributes.unfoldAsync)
+
+  /**
+   * simpler unfold, for infinite sequences.
```
ktoso:
```
Simpler [[unfold]], for infinite sequences.
```

maybe?

akka-stream/src/main/scala/akka/stream/impl/Unfold.scala
```diff
@@ -0,0 +1,79 @@
...
+  }
+}
+
+/**
+ * UnfoldAsync `GraphStage` class
+ * @param s initial state
+ * @param f unfold function
+ * @tparam S state
+ * @tparam E element
+ */
```
ktoso: good docs here, thanks!

ktoso:
Very nice, thanks a lot @hochgi!
Small update of the scaladoc would be nice where I commented (and apply to similar docs on these methods), thanks in advance!

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
hochgi: meaning? how does the usual formatting look like? how should I change this?

ktoso: Like I wrote :) with a space between element and the `,`: 

```scala
case (a, b) ⇒ Some((b → (a + b)) → a)
```

PR #19146, opened 2015-12-10
Port commits to the right branch (http)

ktoso:
Ports https://github.com/akka/akka/pull/19100 to the right branch

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
@@ -55,7 +55,7 @@ class DeflateCompressor extends Compressor {
     res
   }
 
-  private def newTempBuffer(size: Int = 65536): Array[Byte] =
+  private def newTempBuffer(size: Int = 65536): Array[Byte] = {
     // The default size is somewhat arbitrary, we'd like to guess a better value but Deflater/zlib
     // is buffering in an unpredictable manner.
     // `compress` will only return any data if the buffered compressed data has some size in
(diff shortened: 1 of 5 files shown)


ktoso:
Will merge after validation.

ktoso:
Yep... needs adjustments, I'll do those.

ktoso:
FYI @ataraxer :)

akka-http/src/main/scala/akka/http/scaladsl/coding/Deflate.scala
```diff
@@ -63,10 +63,15 @@ class DeflateCompressor extends Compressor {
     // `flush` and `finish` will return any size depending on the previous input.
     // This value will hopefully provide a good compromise between memory churn and
     // excessive fragmentation of ByteStrings.
-    new Array[Byte](size)
+    // We also make sure that buffer size stays within a reasonable range, to avoid
+    // draining deflator with too small buffer.
+    new Array[Byte](math.max(size, MinBufferSize))
+  }
```
2beaucoup: no need for these

ataraxer: This change actually fixes the empty string encoding case.

2beaucoup: These braces don't seem to fix anything. ;)

ataraxer: Ah, *these* :)
IMO dropping braces is OK for one-liners, but when there's a multi line comment directly above — it makes it a bit confusing.

ktoso: Yeah, technically not needed but I don't really mind them here ;-)

ktoso:
Failure was unrelated (groupBy, which Roland resolved AFAIR).

PR #19153, opened 2015-12-12
=str - 19149 - Corrects the behavior of FlowOps.sliding

### LINKED ISSUE #19149: stream (2.0-M2): bug in Flow.sliding
Follow up on: https://groups.google.com/forum/#!topic/akka-user/fhlCbYJ8kOc

Flow.sliding does not correspond exactly with the Seq.sliding behaviour, as was apparently intended. A  minimised example for this bug is:

```scala
Source(Stream.from(0)).take(2).sliding(1, 3)
```
This will place two elements on the stream: {0} and {0, 1}. The correct behaviour would only have placed the former on the stream.

Because I had it lying around by now, a test that should fail for all generated test-cases, but right now succeeds for all of them. Perhaps this will prove useful:
```scala
import akka.actor.ActorSystem
import akka.testkit.TestKit
import org.scalacheck.{Gen, Shrink}
import org.scalatest.concurrent.{Eventually, ScalaFutures}
import org.scalatest.prop.GeneratorDrivenPropertyChecks
import org.scalatest.{FlatSpecLike, Matchers}

/**
 * Created by merlijn on 10/12/15.
 */
class SlidingWindowTest extends TestKit(ActorSystem("test")) with FlatSpecLike with Matchers with GeneratorDrivenPropertyChecks with Eventually with ScalaFutures {

  implicit def noShrink[T]: Shrink[T] = Shrink.shrinkAny[T]

  def test(count: Int, window: Int, step: Int): Future[Seq[Seq[Int]]] = Source(Stream.from(0))
    .take(count)
    .sliding(window, step)
    .runFold(Seq.empty[Seq[Int]]) { case (total, slice) =>
      total :+ slice
    }

  "Sliding" should "never pass this test" in {
    val windowStep = for {
      window <- Gen.choose(1, 10)
      step <- Gen.choose(window + 2, 20)
    } yield (window, step)

    forAll(windowStep) {
      case (window, step) =>

        val incompatibleElementCount = for {
          n <- Gen.choose(1, 20)
          count <- Gen.choose(step * n + window + 1, step * (n + 1) - 1)
        } yield count

        forAll(incompatibleElementCount) { count =>
          whenReady(test(count, window, step)) { testResult =>
            testResult.last.size should be > (window) //For correct behaviour, this test should pass by reversing the assertion: testResult.last.size should be <= (window)
          }
        }
    }
  }
}
```

In essence, the problem pops up (as far as I observed) when the next step in sliding exceeds the last available element in the stream, and some of the elements of the terminated stream were not included in the previous window. In that case, it will create a new window with the same lowerBound as the previous window, and include all elements up to the last one.

viktorklang:
Generatively attempts to verify that it behaves exactly like the
collections library

Fixes #19149

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
@@ -3,128 +3,54 @@
  */
 package akka.stream.scaladsl
 
-import akka.actor.ActorSystem
 import akka.stream.testkit.Utils._
-import akka.stream.testkit.scaladsl.TestSink
 import akka.stream.{ ActorMaterializer, ActorMaterializerSettings }
 import akka.stream.testkit._
-import com.typesafe.config.ConfigFactory
+import org.scalacheck.Gen
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+import akka.pattern.pipe
+
 import scala.concurrent.Await
-import scala.concurrent.duration._
 
-class FlowSlidingSpec extends AkkaSpec {
+class FlowSlidingSpec extends AkkaSpec with GeneratorDrivenPropertyChecks {
   import system.dispatcher
-  val settings = ActorMaterializerSettings(system)
-    .withInputBuffer(initialSize = 2, maxSize = 16)
-
-  implicit val mat = ActorMaterializer(settings)
-
+  implicit val mat = ActorMaterializer(ActorMaterializerSettings(system))
   "Sliding" must {
-    "work with n = 3, step = 1" in assertAllStagesStopped {
-      Source(1 to 6).sliding(n = 3, step = 1).runForeach(testActor ! _) onSuccess {
-        case _ ⇒ testActor ! "done"
+    import org.scalacheck.Shrink.shrinkAny
+    def check(gen: Gen[(Int, Int, Int)]): Unit =
+      forAll(gen, MinSize(1000), MaxSize(1000)) {
+        case (len, win, step) ⇒
+          val af = Source(() ⇒ Iterator.from(0).take(len)).sliding(win, step).runFold(Seq.empty[Seq[Int]])(_ :+ _)
+          val cf = Source(() ⇒ Iterator.from(0).take(len).sliding(win, step)).runFold(Seq.empty[Seq[Int]])(_ :+ _)
+          Await.result(af, remaining) should be(Await.result(cf, remaining))
       }
-      expectMsg(Vector(1, 2, 3))
-      expectMsg(Vector(2, 3, 4))
-      expectMsg(Vector(3, 4, 5))
-      expectMsg(Vector(4, 5, 6))
-      expectMsg("done")
+
+    "behave just like collections sliding with step < window" in assertAllStagesStopped {
+      check(for {
+        len ← Gen.choose(0, 31)
+        win ← Gen.choose(1, 61)
+        step ← Gen.choose(1, win - 1)
+      } yield (len, win, step))
     }
 
-    "work with n = 3, step = 1, 7 elements" in assertAllStagesStopped {
-      Source(1 to 7).sliding(n = 3, step = 1).runForeach(testActor ! _) onSuccess {
-        case _ ⇒ testActor ! "done"
-      }
-      expectMsg(Vector(1, 2, 3))
-      expectMsg(Vector(2, 3, 4))
-      expectMsg(Vector(3, 4, 5))
-      expectMsg(Vector(4, 5, 6))
-      expectMsg(Vector(5, 6, 7))
-      expectMsg("done")
+    "behave just like collections sliding with step == window" in assertAllStagesStopped {
+      check(for {
+        len ← Gen.choose(0, 31)
+        win ← Gen.choose(1, 61)
+        step ← Gen.const(win)
+      } yield (len, win, step))
     }
 
-    "work with n = 3, step = 2" in assertAllStagesStopped {
-      Source(1 to 6).sliding(n = 3, step = 2).runForeach(testActor ! _) onSuccess {
-        case _ ⇒ testActor ! "done"
-      }
-      expectMsg(Vector(1, 2, 3))
-      expectMsg(Vector(3, 4, 5))
-      expectMsg(Vector(5, 6))
-      expectMsg("done")
-    }
-
-    "work with n = 3, step = 2, complete group" in assertAllStagesStopped {
-      Source(1 to 7).sliding(n = 3, step = 2).runForeach(testActor ! _) onSuccess {
-        case _ ⇒ testActor ! "done"
-      }
-      expectMsg(Vector(1, 2, 3))
-      expectMsg(Vector(3, 4, 5))
-      expectMsg(Vector(5, 6, 7))
-      expectMsg("done")
-    }
-
-    "work with n = 3, step = 3" in assertAllStagesStopped {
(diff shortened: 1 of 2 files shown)


akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -318,55 +318,35 @@ private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immut
...
-    if (ctx.isFinishing) {
-      val emit = buf.result()
-      if (emit.size == n) {
-        ctx.finish()
-      } else {
-        ctx.pushAndFinish(emit)
-      }
-    } else ctx.pull()
+    if (!ctx.isFinishing) ctx.pull()
+    else if (buf.size == n || step > n) ctx.finish()
```
viktorklang: This line fixes the bug, the rest is testing and code cleanup.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
```diff
@@ -3,128 +3,39 @@
...
-    "work with n = 3, step = 6" in assertAllStagesStopped {
-      Source(1 to 12).sliding(n = 3, step = 6).runForeach(testActor ! _) onSuccess {
-        case _ ⇒ testActor ! "done"
+      forAll(gen) {
+        case (len, win, step) ⇒
+          val done = s"done: $len $win $step"
+          val as = Source(() ⇒ Iterator.from(0).take(len)).sliding(win, step)
+          val cs = Source(() ⇒ Iterator.from(0).take(len).sliding(win, step))
+          as.zipWith(cs)(_ should be(_)).runFold(done)(Keep.left) pipeTo testActor
+          expectMsg(done)
```
rkuhn: As nice as this looks there must be something wrong with it because it should fail for len 1, win 2, step 3.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
rkuhn: But this line is not free from bugs: when the buffer is not yet full enough, we must emit regardless of whether `step > n` or not.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -318,55 +318,32 @@ private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immut
...
-    bufferedElements += 1
-    if (bufferedElements < n) {
+    buf :+= elem
+    if (buf.size < n) {
+      ctx.pull()
+    } else if (buf.size == n) {
+      ctx.push(buf)
+    } else if (step > n) {
+      if (buf.size == step)
+        buf = Vector.empty
```
rkuhn: This is pretty wasteful for large `step` values: we would build up a huge vector for no reason.

viktorklang: My thinking was to defer opting it until it becomes a graphstage.
Right now it is buggy and I'd rather have an impl which is easy to reason
about than a buggy fast one. This PR is what my very limited time can offer
right now :-/

-- 
Cheers,
√

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
viktorklang: I can try it with that specifically but ScalaCheck tends not to lie.
I'll also add tests to verify that it throws when it should.

-- 
Cheers,
√

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
viktorklang: What do you mean in terms of code? (I can try that)

-- 
Cheers,
√

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
rkuhn: Well, I tried it in the REPL and got different results :-)

viktorklang: Then I definitely need to tweak ScalaCheck to be more thurough!

-- 
Cheers,
√

viktorklang: I found the error, the failure is always in that one of them is too short, and of course zipWith just completes. So I have fixed the scalacheck both to generate more values but also verify lengths.

Pushing my (failing) branch now.

viktorklang:
@rkuhn Alright, I both fixed the bug, made the test better and have some awesome cheese.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
viktorklang: Also, speaking of building up a huge vector for no reason, look at the occurrences of `buf.result()` in the code I replaced. *shrug*

rkuhn: Yes, I know: boy-scout rule :-) Anyway, I agree with first getting it right.

viktorklang: At least the newly pushed branch seems right, and I split the checks out in 3 different ones, testing step less than window, equals to window and greater than window.

Now we have a decent test suite, which will hopefully make it easier to migrate the code to GraphStage and optimize it down the road.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -318,55 +318,32 @@ private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immut
...
-    if (ctx.isFinishing) {
-      val emit = buf.result()
-      if (emit.size == n) {
-        ctx.finish()
-      } else {
-        ctx.pushAndFinish(emit)
-      }
-    } else ctx.pull()
+    if (!ctx.isFinishing) ctx.pull()
+    else if (buf.size >= n) ctx.finish()
```
viktorklang: This line is the bugfix

rkuhn:
LGTM, thanks!

viktorklang:
@rkuhn

I just realized that step may end up 0 here:
https://github.com/akka/akka/pull/19153/files#diff-17e38cb6b9c5bd357a7eb5abb6de7a9fR32

should be Gen.choose(1, math.max(1, win - 1))

-- 
Cheers,
√

rkuhn:
That’s not a problem, Gen.choose(1, 0) will just not emit anything.

viktorklang:
Ok, surprising but good

-- 
Cheers,
√

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSlidingSpec.scala
drewhk: btw, this is what I don't like about ScalaCheck - the list of cases being checked is not visible explicitly. Maybe an IDE plugin to generate test files using ScalaCheck like rules what I would be happy with.

rkuhn: On the contrary, the list is very succinct and visible: assume that all cases covered by the generators will (eventually) be tested. This is so much easier for humans to understand and verify than an incomplete listing of manually enumerated corner cases.

Issue #19017, opened 2015-11-25
DistributedPubSub - race condition regarding children/topic actor liveness


Adebski:
When new subscriber arrives to `DistributedPubSubMediator` for given topic the child/topic actor is looked up using  `context.child(encTopic)`, if actor exists message is forwarded, if not the child actor is created, registered in `registry` and then message is forwarded. 

Topic actors stop themselves when all their subscribers are either terminated (death watch by topic actor on subscriber) or they just unsubscribe themselves from the topic.

There could be a situation where
* `Subscribe` message is sent to mediator from `SubscriberA`
* At the same time `SubscriberB` is terminated and this message arrives to topic actor.
* Topic actor stops itself using `context.stop(self)`
* At the same time parent/mediator actor performs lookup of child and decided that topic actor is alive and forwards him `Subscribe` message
* `Subscribe` message goes to dead letters because topic actor is already dead, mediator does not inform subscriber that something went wrong and/or do not try to retry sending Subscribe message to new version of topic actor.

According to http://stackoverflow.com/questions/23629159/get-or-create-child-akka-actor-and-ensure-liveness the solution would be to send kind-of Terminated message to mediator so `Subscribe` and `Terminated` messages are processed serially, also watching out for the fact that stopping actor is async operation so even if we invoke `context.stop(child)` from context of parent there is no guarantee that path is immediately.

ktoso:
Thanks for reporting. We'll follow up here shortly.

// cc @patriknw

ktoso:
Having that said, the Akka pub-sub has the same guarantees as usual Actor messaging - messages can get lost, if you want to guarantee they don't, you need to use `AtLeastOnceDelivery` or build ACKs into your protocol. It seems that's what you're after here?

Adebski:
I was not thinking about total guarantee, more like reducing the possibility of such by handling the termination of a child actor in a way that eliminates this particular factor from a chance of losing a message.

patriknw:
I agree with @Adebski
This is about subscription messages, which we don't handle in the best possible way.

Adebski:
Ok, my idea is to store some additional state inside mediator that basically says to us "for how many subscribers we sent `Subscribe` to `Topic` actor but not received confirmation". If we have this kind of information we can do something like:

* when `Prune` message arrives to Topic actor and it sees that for this point in time it has no subscribers it sends some new kind of message to the parent (e.g. `PruneNotification`)
[...]

PR #19154, opened 2015-12-12
Race-condition problem in distributed-pub-sub fixed by introducing passivate-like protocol for mediator and Topic actors.

Adebski:
Should also fix a problem when there would be `Prune` message queued in `Topic` actor mailbox and `Subscribe` message is sent to `Topic` by `DistributedPubSubMediator` when `pruneDeadline` is overdue.

Due to the nature of the issue unfortunately there are no new test cases for this racy situation - I do not  see any easy way to "guide" involved actors (`DistributedPubSubMediator`, `Topic`, `Group`) to deterministically recreate situations where in old code we would lose messages.

### DIFF
--- a/akka-cluster-tools/src/main/resources/reference.conf
+++ b/akka-cluster-tools/src/main/resources/reference.conf
@@ -33,7 +33,6 @@ akka.cluster.pub-sub {
   # If not specified default dispatcher is used.
   # If specified you need to define the settings of the actual dispatcher.
   use-dispatcher = ""
-
 }
 # //#pub-sub-ext-config
 
--- a/akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
+++ b/akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
@@ -10,19 +10,7 @@ import scala.concurrent.duration._
 import java.util.concurrent.ThreadLocalRandom
 import java.net.URLEncoder
 import java.net.URLDecoder
-import akka.actor.Actor
-import akka.actor.ActorContext
-import akka.actor.ActorLogging
-import akka.actor.ActorPath
-import akka.actor.ActorRef
-import akka.actor.ActorSystem
-import akka.actor.Address
-import akka.actor.ExtendedActorSystem
-import akka.actor.Extension
-import akka.actor.ExtensionId
-import akka.actor.ExtensionIdProvider
-import akka.actor.Props
-import akka.actor.Terminated
+import akka.actor._
 import akka.cluster.Cluster
 import akka.cluster.ClusterEvent._
 import akka.cluster.Member
@@ -38,10 +26,7 @@ import akka.routing.ConsistentHashingRoutingLogic
 import akka.routing.BroadcastRoutingLogic
 import scala.collection.immutable.TreeMap
 import com.typesafe.config.Config
-import akka.actor.NoSerializationVerificationNeeded
-import akka.actor.Deploy
 import akka.dispatch.Dispatchers
-import akka.actor.DeadLetterSuppression
 
 object DistributedPubSubSettings {
   /**
@@ -251,12 +236,48 @@ object DistributedPubSubMediator {
     @SerialVersionUID(1L)
     final case class SendToOneSubscriber(msg: Any)
 
+    /**
+     * Messages used to encode protocol to make sure that we do not send Subscribe/Unsubscribe message to
+     * child (mediator -&gt; topic, topic -&gt; group) during a period of transition. Protects from situations like:
+     *
+     * Sending Subscribe/Unsubscribe message to child actor after child has been terminated
+     * but Terminate message did not yet arrive to parent.
+     *
+     * Sending Subscribe/Unsubscribe message to child actor that has Prune message queued and pruneDeadline set.
+     *
+     * In both of those situation parent actor still thinks that child actor is alive and forwards messages to it resulting in lost ACKs.
+     */
+    trait ChildActorTerminationProtocol
+
+    /**
+     * Passivate-like message sent from child to parent, used to signal that sender has no subscribers and no child actors.
+     */
+    case object NoMoreSubscribers extends ChildActorTerminationProtocol
+
+    /**
+     * Sent from parent to child actor to signalize that messages are being buffered. When received by child actor
+     * if no [[Subscribe]] message has been received after sending [[NoMoreSubscribers]] message child actor will stop itself.
+     */
+    case object TerminateRequest extends ChildActorTerminationProtocol
+
+    /**
+     * Sent from child to parent actor as response to [[TerminateRequest]] in case [[Subscribe]] message arrived
+     * after sending [[NoMoreSubscribers]] but before receiving [[TerminateRequest]].
+     *
+     * When received by the parent buffered messages will be forwarded to child actor for processing.
+     */
+    case object NewSubscriberArrived extends ChildActorTerminationProtocol
+
     @SerialVersionUID(1L)
     final case class MediatorRouterEnvelope(msg: Any) extends RouterEnvelope {
       override def message = msg
     }
 
     def encName(s: String) = URLEncoder.encode(s, "utf-8")
+
+    def mkKey(ref: ActorRef): String = mkKey(ref.path)
+
+    def mkKey(path: ActorPath): String = path.toStringWithoutAddress
 
     trait TopicLike extends Actor {
       import context.dispatcher
@@ -286,7 +307,15 @@ object DistributedPubSubMediator {
         case Terminated(ref) ⇒
           remove(ref)
         case Prune ⇒
-          for (d ← pruneDeadline if d.isOverdue) context stop self
+          for (d ← pruneDeadline if d.isOverdue) {
+            pruneDeadline = None
+            context.parent ! NoMoreSubscribers
+          }
+        case TerminateRequest ⇒
+          if (subscribers.isEmpty && context.children.isEmpty)
+            context stop self
+          else
+            context.parent ! NewSubscriberArrived
         case msg ⇒
           subscribers foreach { _ forward msg }
       }
@@ -303,28 +332,46 @@ object DistributedPubSubMediator {
       }
     }
 
-    class Topic(val emptyTimeToLive: FiniteDuration, routingLogic: RoutingLogic) extends TopicLike {
+    class Topic(val emptyTimeToLive: FiniteDuration, routingLogic: RoutingLogic) extends TopicLike with PerGroupingBuffer {
       def business = {
         case msg @ Subscribe(_, Some(group), _) ⇒
           val encGroup = encName(group)
-          context.child(encGroup) match {
-            case Some(g) ⇒ g forward msg
-            case None ⇒
-              val g = context.actorOf(Props(classOf[Group], emptyTimeToLive, routingLogic), name = encGroup)
-              g forward msg
-              context watch g
-              context.parent ! RegisterTopic(g)
+          bufferOr(mkKey(self.path / encGroup), msg, sender()) {
+            context.child(encGroup) match {
+              case Some(g) ⇒ g forward msg
+              case None    ⇒ newGroupActor(encGroup) forward msg
+            }
           }
           pruneDeadline = None
         case msg @ Unsubscribe(_, Some(group), _) ⇒
-          context.child(encName(group)) match {
-            case Some(g) ⇒ g forward msg
-            case None    ⇒ // no such group here
+          val encGroup = encName(group)
+          bufferOr(mkKey(self.path / encGroup), msg, sender()) {
+            context.child(encGroup) match {
+              case Some(g) ⇒ g forward msg
+              case None    ⇒ // no such group here
+            }
           }
         case msg: Subscribed ⇒
           context.parent forward msg
         case msg: Unsubscribed ⇒
           context.parent forward msg
+        case NoMoreSubscribers ⇒
+          val key = mkKey(sender())
+          initializeGrouping(key)
+          sender() ! TerminateRequest
+        case NewSubscriberArrived ⇒
+          val key = mkKey(sender())
+          forwardMessages(key, sender())
+        case Terminated(ref) ⇒
+          val key = mkKey(ref)
+          recreateAndForwardMessagesIfNeeded(key, newGroupActor(ref.path.name))
+      }
+
+      def newGroupActor(encGroup: String): ActorRef = {
+        val g = context.actorOf(Props(classOf[Group], emptyTimeToLive, routingLogic), name = encGroup)
+        context watch g
+        context.parent ! RegisterTopic(g)
+        g
       }
     }
 
@@ -424,7 +471,7 @@ trait DistributedPubSubMessage extends Serializable
  * [[DistributedPubSubMediator.SubscribeAck]] and [[DistributedPubSubMediator.UnsubscribeAck]]
  * replies.
  */
-class DistributedPubSubMediator(settings: DistributedPubSubSettings) extends Actor with ActorLogging {
+class DistributedPubSubMediator(settings: DistributedPubSubSettings) extends Actor with ActorLogging with PerGroupingBuffer {
 
   import DistributedPubSubMediator._
   import DistributedPubSubMediator.Internal._
@@ -521,17 +568,27 @@ class DistributedPubSubMediator(settings: DistributedPubSubSettings) extends Act
 
     case msg @ Subscribe(topic, _, _) ⇒
       // each topic is managed by a child actor with the same name as the topic
+
       val encTopic = encName(topic)
-      context.child(encTopic) match {
-        case Some(t) ⇒ t forward msg
-        case None ⇒
-          val t = context.actorOf(Props(classOf[Topic], removedTimeToLive, routingLogic), name = encTopic)
-          t forward msg
-          registerTopic(t)
+
+      bufferOr(mkKey(self.path / encTopic), msg, sender()) {
+        context.child(encTopic) match {
+          case Some(t) ⇒ t forward msg
+          case None    ⇒ newTopicActor(encTopic) forward msg
+        }
       }
 
     case msg @ RegisterTopic(t) ⇒
       registerTopic(t)
+
+    case NoMoreSubscribers ⇒
+      val key = mkKey(sender())
+      initializeGrouping(key)
+      sender() ! TerminateRequest
+
+    case NewSubscriberArrived ⇒
+      val key = mkKey(sender())
+      forwardMessages(key, sender())
 
     case GetTopics ⇒ {
       sender ! CurrentTopics(getCurrentTopics())
@@ -541,9 +598,12 @@ class DistributedPubSubMediator(settings: DistributedPubSubSettings) extends Act
       ref ! ack
 
     case msg @ Unsubscribe(topic, _, _) ⇒
-      context.child(encName(topic)) match {
-        case Some(t) ⇒ t forward msg
-        case None    ⇒ // no such topic here
+      val encTopic = encName(topic)
+      bufferOr(mkKey(self.path / encTopic), msg, sender()) {
+        context.child(encTopic) match {
+          case Some(t) ⇒ t forward msg
+          case None    ⇒ // no such topic here
+        }
       }
 
     case msg @ Unsubscribed(ack, ref) ⇒
@@ -585,6 +645,7 @@ class DistributedPubSubMediator(settings: DistributedPubSubSettings) extends Act
           put(key, None)
         case _ ⇒
       }
+      recreateAndForwardMessagesIfNeeded(key, newTopicActor(a.path.name))
 
     case state: CurrentClusterState ⇒
       nodes = state.members.collect {
@@ -669,9 +730,9 @@ class DistributedPubSubMediator(settings: DistributedPubSubSettings) extends Act
     context.watch(ref)
   }
 
-  def mkKey(ref: ActorRef): String = mkKey(ref.path)
+  def mkKey(ref: ActorRef): String = Internal.mkKey(ref)
 
-  def mkKey(path: ActorPath): String = path.toStringWithoutAddress
+  def mkKey(path: ActorPath): String = Internal.mkKey(path)
 
   def myVersions: Map[Address, Long] = registry.map { case (owner, bucket) ⇒ (owner -> bucket.version) }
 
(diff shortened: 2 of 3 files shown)


akka-cluster-tools/src/main/resources/reference.conf
```diff
@@ -34,6 +34,11 @@ akka.cluster.pub-sub {
   # If specified you need to define the settings of the actual dispatcher.
   use-dispatcher = ""
 
+  # Maximum number of Subscribe/Unsubscribe messages buffered by mediator actor in situation where topic actor is in process of temrination.
```
rkuhn: `s/temrination/termination/`

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -251,6 +253,35 @@ object DistributedPubSubMediator {
...
+    /*
+    Messages used to encode protocol to make sure that we do not send Subscribe/Unsubscribe message to 
+    child (mediator -> topic, topic -> group) during a period of transition. Protects from situations like:
+    
+    Sending Subscribe/Unsubscribe message to child actor after child has been terminated but Terminate message did not yet arrive to parent.
+    
+    Sending Subscribe/Unsubscribe message to child actor that has Prune message queued and pruneDeadline set.
+    
+    In both of those situation parent actor still thinks that child actor is alive and forwards messages to it resulting in lost ACKs.
+     */
```
rkuhn: We use the following style:
~~~
/*
 * <text goes here>
 */
~~~

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -258,6 +289,33 @@ object DistributedPubSubMediator {
 
     def encName(s: String) = URLEncoder.encode(s, "utf-8")
 
+    def mkKey(ref: ActorRef): String = mkKey(ref.path)
+
+    def mkKey(path: ActorPath): String = path.toStringWithoutAddress
+
+    trait PerGroupingBuffer {
+      var buffers: Map[String, Vector[(Any, ActorRef)]] = Map.empty
```
rkuhn: this should be private

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -258,6 +289,33 @@ object DistributedPubSubMediator {
...
+
+      def totalBufferSize = buffers.foldLeft(0) { (sum, entity) ⇒ sum + entity._2.size }
```
rkuhn: and then this can be done via a counter that is kept up-to-date, since folding creates lots of garbage and wastes cycles

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -303,28 +369,52 @@ object DistributedPubSubMediator {
...
-            case None ⇒
-              val g = context.actorOf(Props(classOf[Group], emptyTimeToLive, routingLogic), name = encGroup)
-              g forward msg
-              context watch g
-              context.parent ! RegisterTopic(g)
+          bufferOr(bufferLimit, mkKey(self.path / encGroup), msg, sender(), msg ⇒ unhandled(msg)) {
+            context.child(encGroup) match {
+              case Some(g) ⇒ g forward msg
+              case None ⇒
+                newGroupActor(encGroup) forward msg
```
rkuhn: please join this onto the previous line

rkuhn: OTOH: where has `context.parent ! RegisterTopic(g)` gone?

rkuhn: oh, forget it: I mixed up with the definition of newTopicActor below

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -303,28 +369,52 @@ object DistributedPubSubMediator {
...
+        case NoMoreSubscribers ⇒
+          val key = mkKey(sender())
+          buffers += key -> Vector.empty
+          sender() ! TerminateRequest
+        case NewSubscriberArrived ⇒
+          val key = mkKey(sender())
+          clearGrouping(key, sender())
+        case Terminated(ref) ⇒
+          val key = mkKey(ref)
+          if (buffers.contains(key)) {
```
rkuhn: The buffer should typically contain an empty Vector here, in which case we wouldn’t want a new Actor to be created, right?

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -585,6 +688,10 @@ class DistributedPubSubMediator(settings: DistributedPubSubSettings) extends Act
           put(key, None)
         case _ ⇒
       }
+      if (buffers.contains(key)) {
```
rkuhn: same here: should check buffer contents before spawning a new Actor

rkuhn:
OK TO TEST

rkuhn:
Thanks @Adebski, looking pretty good already!

drewhk:
Javadoc error

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -251,6 +253,35 @@ object DistributedPubSubMediator {
     @SerialVersionUID(1L)
     final case class SendToOneSubscriber(msg: Any)
 
+    /*
+    Messages used to encode protocol to make sure that we do not send Subscribe/Unsubscribe message to 
+    child (mediator -> topic, topic -> group) during a period of transition. Protects from situations like:
+    
+    Sending Subscribe/Unsubscribe message to child actor after child has been terminated but Terminate message did not yet arrive to parent.
```
drewhk: break long line

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
drewhk: yes, incremental updates are preferred.

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -258,6 +289,33 @@ object DistributedPubSubMediator {
...
+
+    trait PerGroupingBuffer {
+      var buffers: Map[String, Vector[(Any, ActorRef)]] = Map.empty
+
+      def totalBufferSize = buffers.foldLeft(0) { (sum, entity) ⇒ sum + entity._2.size }
+
+      def bufferOr(limit: Int, grouping: String, message: Any, originalSender: ActorRef, discard: Any ⇒ Unit)(action: ⇒ Unit): Unit = {
+        buffers.get(grouping) match {
+          case None                                      ⇒ action
+          case Some(messages) if totalBufferSize < limit ⇒ buffers = buffers.updated(grouping, messages :+ ((message, originalSender)))
```
drewhk: break long line

Adebski:
About the comment with old `totalBufferSize` - the same "technique" is used in `ShardRegion` actor so it is also potentially place where simple int counter could be introduced.

rkuhn:
Yes, that is probably true. It would be great if you could apply that over there as well.

Adebski:
Do you want separate issue for that or can I do it in this one?

rkuhn:
It would be better as a separate PR, thanks!

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -96,13 +85,18 @@ object DistributedPubSubSettings {
...
  */
 final class DistributedPubSubSettings(
   val role: Option[String],
   val routingLogic: RoutingLogic,
   val gossipInterval: FiniteDuration,
   val removedTimeToLive: FiniteDuration,
-  val maxDeltaElements: Int) extends NoSerializationVerificationNeeded {
+  val maxDeltaElements: Int,
+  val mediatorBufferSize: Int,
+  val perTopicActorBufferSize: Int) extends NoSerializationVerificationNeeded {
```
patriknw: this is not binary backwards compatible, you need to add a constructor that matches the old constructor

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -251,6 +253,36 @@ object DistributedPubSubMediator {
...
+     */
+    case object TerminateRequest
+
+    /**
+     * Sent from child to parent actor as response to [[TerminateRequest]] in case [[Subscribe]] message arrived
+     * after sending [[NoMoreSubscribers]] but before receiving [[TerminateRequest]].
+     *
+     * When received by the parent buffered messages will be forwarded to child actor for processing.
+     */
+    case object NewSubscriberArrived
```
patriknw: these should not be public api, place them inside the `Internal` object

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -258,6 +290,47 @@ object DistributedPubSubMediator {
...
+          case None ⇒ action
+          case Some(messages) if totalBufferSize < limit ⇒
+            buffers = buffers.updated(grouping, messages :+ ((message, originalSender)))
+            totalBufferSize += 1
+          case Some(_) ⇒ discard(message)
+        }
+      }
+
+      def recreateAndForwardMessagesIfNeeded(grouping: String, recipient: ⇒ ActorRef): Unit = {
+        buffers.get(grouping).filter(_.nonEmpty).foreach(forwardMessages(_, recipient))
```
patriknw: why filter?

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -258,6 +290,47 @@ object DistributedPubSubMediator {
...
+        buffers -= grouping
+      }
+
+      def forwardMessages(grouping: String, recipient: ActorRef): Unit = {
+        buffers.get(grouping).foreach(forwardMessages(_, recipient))
+        buffers -= grouping
```
patriknw: I'd prefer to update the `totalBufferSize` at the same place as where the `buffers` is updated

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -303,28 +384,46 @@ object DistributedPubSubMediator {
...
         case msg @ Subscribe(_, Some(group), _) ⇒
           val encGroup = encName(group)
-          context.child(encGroup) match {
-            case Some(g) ⇒ g forward msg
-            case None ⇒
-              val g = context.actorOf(Props(classOf[Group], emptyTimeToLive, routingLogic), name = encGroup)
-              g forward msg
-              context watch g
-              context.parent ! RegisterTopic(g)
+          bufferOr(bufferLimit, mkKey(self.path / encGroup), msg, sender(), msg ⇒ unhandled(msg)) {
```
patriknw: if you use
`trait PerGroupingBuffer { slf: Actor =>`
You can use `unhandled` inside `PerGroupingBuffer` instead of `msg => unhandled`
Otherwise the function should be defined as a `val` to avoid allocations.

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -303,28 +384,46 @@ object DistributedPubSubMediator {
...
           context.parent forward msg
         case msg: Unsubscribed ⇒
           context.parent forward msg
+        case NoMoreSubscribers ⇒
+          val key = mkKey(sender())
+          initializeGrouping(key)
+          sender() ! TerminateRequest
+        case NewSubscriberArrived ⇒
+          val key = mkKey(sender())
+          forwardMessages(key, sender())
```
patriknw: how can this case happen? aren't everything buffered from `! TerminateRequest` until `Terminated`?

akka-cluster-tools/src/main/resources/reference.conf
```diff
@@ -34,6 +34,11 @@ akka.cluster.pub-sub {
   # If specified you need to define the settings of the actual dispatcher.
   use-dispatcher = ""
 
+  # Maximum number of Subscribe/Unsubscribe messages buffered by mediator actor in situation where topic actor is in process of termination.
+  mediator-buffer-size = 1000
+
+  # Maximum number of Subscribe/Unsubscribe messages buffered by each topic actor in situation where group actor is in process of termination.
+  per-topic-actor-buffer-size = 200
```
patriknw: I don't think we need a limited buffer size for this. We know that the stopping will be quick enough to buffer everything.

patriknw:
great contribution, @Adebski 
a little more polish and it will shine

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
Adebski: Those messages are already inside `private[akka] object Internal` as far as I can see

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
Adebski: this method is called when Terminate message finally arrives to the Mediator/Topic. When that happens if there is grouping and there are messages buffered we have to recreate Topic/Group actor and forward the messages so `.filter(_nonEmpty)` return Some(messages) only if there is at least one message buffered. 

If there is no grouping or we did not buffer any messages we do not have to recreate the actor. This solves the problem mentioned in previous "iteration" - that only checking if there is a grouping for given key is not enough in case when we want to make sure we do not have to recreate child actor.

akka-cluster-tools/src/main/resources/reference.conf
Adebski: You are referring to both of the values or only the value for topic actor?

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
Adebski: There is one situation where it could happen. Imagine this:

* Topic actor sends `NoMoreSubscribers` to parent
* before `NoMoreSubscribers` arrives to mediator mediator handles `Subscribe` message and forwards it to `Topic`
* `Subscribe` message arrives to `Topic` actor
* `NoMoreSubscribers` arrives to `Mediator`, we start buffering the messages 

now if mediator would only stop `Topic` actor using `context.stop` or send `PoisonPill` subscriber could think that his subscription is registered but we would kill `Topic` actor anyway because we can't do any action when `PoisonPill` is received. 

This scheme works for `cluster-sharding` due to the fact that persistent actors are persistent so even if we terminate it on recreation the state will be restored, in this case we would lose information about subscription that was ACKed just before topic actor was killed.

That is why child actors are not stopped from the parent but additional message is sent to them -  `TerminateRequest`. If child actor sees that it has subscribers/childs it knows that some important message arrived before parent started buffering but after `NoMoreSubscribers` was sent. In that case `NewSubscriberArrived` is sent to parent so it can forward buffered messages to old child actor.

patriknw: ah, now I see the need for NewSubscriberArrived, thanks for clarifying

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
patriknw: ah, the `filter` is on the `Option`, I misread that

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
patriknw: ah, sorry, missed full context

akka-cluster-tools/src/main/resources/reference.conf
patriknw: both values

Adebski: Ok, so there won't be a need to change config class and it will also simplify the logic a bit.

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
Adebski: Moved counter decrease to .lambda passed to foreach

Adebski:
@patriknw - any update on review? As for failing tests I can see in the logs that NPE is thrown on Javadoc generation

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -251,6 +238,36 @@ object DistributedPubSubMediator {
     @SerialVersionUID(1L)
     final case class SendToOneSubscriber(msg: Any)
 
+    /*
+     * Messages used to encode protocol to make sure that we do not send Subscribe/Unsubscribe message to 
+     * child (mediator -> topic, topic -> group) during a period of transition. Protects from situations like:
```
patriknw: the javadoc problems might be because of this, plesase encode `>` as `&gt;`

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -251,6 +238,36 @@ object DistributedPubSubMediator {
...
+     * 
+     * Sending Subscribe/Unsubscribe message to child actor after child has been terminated 
+     * but Terminate message did not yet arrive to parent.
+     *
+     * Sending Subscribe/Unsubscribe message to child actor that has Prune message queued and pruneDeadline set.
+     *
+     * In both of those situation parent actor still thinks that child actor is alive and forwards messages to it resulting in lost ACKs.
+     */
+    /**
```
patriknw: hm, should't the above comment be included in this?

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
```diff
@@ -251,6 +238,36 @@ object DistributedPubSubMediator {
...
+     * Passivate-like message sent from child to parent, used to signal that sender has no subscribers and no child actors.
+     */
+    case object NoMoreSubscribers
+
+    /**
+     * Sent from parent to child actor to signalize that messages are being buffered. When received by child actor
+     * if no [[Subscribe]] message has been received after sending [[NoMoreSubscribers]] message child actor will stop itself.
```
patriknw: the javadoc problems might also be because of these links, (objects in objects). I suggest that you just change to `Subscribe` with single back-tick, same with other such links here and below

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
Adebski: My intention was to describe why NoMoreSubscribers, TerminateRequest, NewSubscriberArrived were added in the first place, putting this comment inside java/scaladoc would seem like it only describes this particular message.

Maybe those three messages could be enclosed in some external object and this comment could be included in java/scaladoc for this enclosing object.

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
Adebski: As for the javadoc problems I tried to invoke `genjavadoc:doc` task on my local machine and the problem seems to be the placement of `PerGroupingBuffer`.

I am not sure if that is the direct reason but because `DistributedPubSubMediator` has to have `PerGroupingBufer` there is `import akka.cluster.pubsub.DistributedPubSubMediator.Internal.PerGroupingBuffer` and looking at the stacktrace 

```Scala```
error]   symbol:   class Internal
[error]   location: class DistributedPubSubMediator
[error] /home/adebski/Nauka/akka/akka-cluster-tools/target/java/akka/cluster/pubsub/DistributedPubSubMediator.java:69: error: cyclic inheritance involving DistributedPubSubMediator
[error] public  class DistributedPubSubMediator implements akka.actor.Actor, akka.actor.ActorLogging, akka.cluster.pubsub.DistributedPubSubMediator.Internal.PerGroupingBuffer {
[error]         ^
[error] /home/adebski/Nauka/akka/akka-cluster-tools/target/java/akka/cluster/pubsub/DistributedPubSubMediator.java:263: error: cannot find symbol
[error]     public static final Internal$ MODULE$ = null;
```
it may be it. 

I can just move this trait outside DistributedPubSubMediator companion object or even to separate file.

drewhk:
This can be merged once it compiles

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
patriknw: you could add a marker trait, that these three extends

akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala
patriknw: did you remove the `[[ ... ]]` links, as I suggested? These are not public classes anyway

Adebski: Yes - I even removed all additional comments made by me and it only worked after I moved mentioned trait outside `DistributedPubSubMediator` companion object

Adebski:
To fix binary compatibility errors I restored `mkKey` methods to `DistributedPubSubMediator` class but now they forward to implementation in `Internal` object so it can be reused from places

patriknw:
LGTM, great work @Adebski 
Please squash to one commit and we will merge this

patriknw:
Refs #19017

Adebski:
Squashed to one commit.

patriknw:
Thanks for contributing, @Adebski

PR #19155, opened 2015-12-12
 #18962 MergeSorted and read() completion handling

### LINKED ISSUE #18962: add MergeSorted
as actual test and use-case of GraphStageLogic’s `read` support (and with the need to add an `onClose` handler to that)

rkuhn:
fixes #18962

also fixes some bugs in `read` and `passAlong`


akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -418,23 +422,25 @@ private[http] object HttpServerBluePrint {
...
-                case Left(frameHandler) ⇒ frameHandler
-                case Right(messageHandler) ⇒
-                  Websocket.stack(serverSide = true, maskingRandomFactory = websocketRandomFactory, log = log).join(messageHandler)
-              }
-              installHandler(frameHandler)
-              websocketHandlerWasInstalled = true
-          }
-          else
-            read(fromWs)(push(toNet, _))
+          if (isHttp) read(fromHttp)(httpToNet, shutdown)
```
viktorklang: Will it ever make sense to make this less hard-coded? (Upgrade to more than WS I mean)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphMergeSortedSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  override def fixture(b: GraphDSL.Builder[_]): Fixture = new Fixture(b) {
+    val merge = b.add(new MergeSorted[Outputs])
+
+    override def left: Inlet[Outputs] = merge.in0
+    override def right: Inlet[Outputs] = merge.in1
+    override def out: Outlet[Outputs] = merge.out
+  }
+
+  implicit val patience = PatienceConfig(1.second)
```
viktorklang: I'd probably put the `implicit val shrink = Shrink(_ ⇒ Stream.empty)` here iso the `(implicitly,…)`-dance

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphMergeSortedSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+import akka.stream.testkit.TwoStreamsSetup
+import org.scalacheck.Gen
+import scala.util.Random
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+import org.scalatest.concurrent.ScalaFutures
+import scala.concurrent.duration._
+import org.scalactic.ConversionCheckedTripleEquals
+import org.scalacheck.Shrink
+
+class GraphMergeSortedSpec extends TwoStreamsSetup with GeneratorDrivenPropertyChecks with ScalaFutures with ConversionCheckedTripleEquals {
```
viktorklang: ScalaCheck

ALL THE THINGS!!

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -1058,6 +1059,33 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
...
+   *
+   * '''Emits when''' one of the inputs has an element available
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' all upstreams complete
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def mergeSorted[U >: Out, M](that: Graph[SourceShape[U], M], comp: Comparator[U]): javadsl.Flow[In, U, Mat] =
```
viktorklang: I'm not head-over-heels with the proposed name. Not only is "mergesort" something completely different, but the resulting stream. Also, wat is important to remember is that the resulting stream is only sorted if the merged streams were sorted.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1352,6 +1352,31 @@ trait FlowOps[+Out, +Mat] {
...
+   *
+   * '''Completes when''' all upstreams complete
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def mergeSorted[U >: Out, M](that: Graph[SourceShape[U], M])(implicit ord: Ordering[U]): Repr[U] =
+    via(mergeSortedGraph(that))
+
+  protected def mergeSortedGraph[U >: Out, M](that: Graph[SourceShape[U], M])(implicit ord: Ordering[U]): Graph[FlowShape[Out @uncheckedVariance, U], M] =
+    GraphDSL.create(that) { implicit b ⇒
```
viktorklang: Given how many methods we now have with almost exactly this piece of code/scaffolding I would propose to attempt some reuse :)

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -227,6 +227,55 @@ final class MergePreferred[T] private (val secondaryPorts: Int, val eagerClose:
...
+ *
+ * '''Emits when''' both inputs have an element available
+ *
+ * '''Backpressures when''' downstream backpressures
+ *
+ * '''Completes when''' all upstreams complete
+ *
+ * '''Cancels when''' downstream cancels
+ */
+class MergeSorted[T: Ordering] extends GraphStage[FanInShape2[T, T, T]] {
```
viktorklang: final

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -227,6 +227,55 @@ final class MergePreferred[T] private (val secondaryPorts: Int, val eagerClose:
...
+  private val left = Inlet[T]("left")
```
viktorklang: left.in?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -816,19 +824,28 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount:
...
+   * given outlet before pulling for more data. `doFinish` and `doFail` control whether
    * completion or failure of the given inlet shall lead to stage termination or not.
-   */
-  final protected def passAlong[Out, In <: Out](from: Inlet[In], to: Outlet[Out], doFinish: Boolean, doFail: Boolean): Unit =
+   * `doPull` instructs to perform one initial pull on the `from` port.
+   */
+  final protected def passAlong[Out, In <: Out](from: Inlet[In], to: Outlet[Out],
+                                                doFinish: Boolean = true, doFail: Boolean = true,
+                                                doPull: Boolean = false): Unit = {
+    val puller = () ⇒ tryPull(from)
```
viktorklang: These kinds allocations and the indirection they cause is still giving me the heebie-jeebies

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
rkuhn: We’ll burn that bridge when we get there …

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphMergeSortedSpec.scala
rkuhn: Yes, but I wanted to wait with that cosmetic fix until review comments are in ;-)

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
rkuhn: I actually quite like the name because it matches “merge sorted streams” as well as that the result of that operation is a sorted stream (i.e. if the inputs were sorted).

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: Yes, I might look into that now …

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
rkuhn: no, why?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: It is the best possible solution, how else could it be done?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: Unfortunately the types and how to access the ports vary such that any syntactic gains would have runtime cost.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: well, admittedly I just found a solution with one less allocation—happy now? ;-)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphMergeSortedSpec.scala
viktorklang: Yeah, right… ;)

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
viktorklang: Yes, but do you acknowledge the "mergesort" name clash? :)
Before we burn the shed, how about "mergePreSorted"?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: Fair enough!

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -816,19 +824,30 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount:
...
-      val puller = () ⇒ tryPull(from)
+  final protected def passAlong[Out, In <: Out](from: Inlet[In], to: Outlet[Out],
+                                                doFinish: Boolean = true, doFail: Boolean = true,
+                                                doPull: Boolean = false): Unit = {
+    class PassAlongHandler extends InHandler with (() ⇒ Unit) {
+      override def apply(): Unit = tryPull(from)
       override def onPush(): Unit = {
         val elem = grab(from)
-        emit(to, elem, puller)
+        emit(to, elem, this)
```
viktorklang: :)

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
rkuhn: Well, not sure actually. The interpretation of “mergeSorted = merged by way of merge-sort” seems a bit far-fetched to me, while “merge, but sorted” is much shorter and closer at hand. And if the name does not reveal the semantics there’s always the documentation to refer to.

`mergePreSorted` does convey that something is pre-sorted here, but it does not convey the merge semantics themselves. `mergeSortedStreamsInSortedFashion` would be really precise, but people would just crucify us for picking such a name, I think ;-)

rkuhn: In fact, what we do is actually a kind of merge-sort.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphMergeSortedSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+    override def out: Outlet[Outputs] = merge.out
+  }
+
+  implicit val patience = PatienceConfig(1.second)
+  implicit def noShrink[T] = Shrink[T](_ ⇒ Stream.empty) // do not shrink failures, it only destroys evidence
+
+  "MergeSorted" must {
+
+    "work in the nominal case" in {
+      val gen = Gen.listOf(Gen.oneOf(false, true))
```
drewhk: Is it really worth to pull in ScalaCheck for these simple generators? I would prefer rather to see explicitly the list of corner cases being listed and looped over.

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
drewhk: Also, mergeSort exploits streaming sorted merges, so the naming is even more appropriate from that angle.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -318,6 +318,55 @@ final class Interleave[T] private (val inputPorts: Int, val segmentSize: Int, va
...
+    setHandler(out, eagerTerminateOutput)
+
+    var other: T = _
+    def nullOut(): Unit = other = null.asInstanceOf[T]
+
+    def dispatch(l: T, r: T): Unit =
+      if (l < r) { other = r; emit(out, l, readL) }
+      else { other = l; emit(out, r, readR) }
+
+    val dispatchR = dispatch(other, _: T)
```
drewhk: private vals?

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
viktorklang: `mergeSortPreSorted` ;-)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1372,19 +1370,42 @@ trait FlowOps[+Out, +Mat] {
...
       r ⇒
-        import GraphDSL.Implicits._
         val merge = b.add(Merge[U](2))
         r ~> merge.in(1)
         FlowShape(merge.in(0), merge.out)
     }
 
   /**
+   * Merge the given [[Source]] to this [[Flow]], taking elements as they arrive from input streams,
+   * picking always the smallest of the available elements (waiting for one element from each side
```
drewhk: this point needs to be more explained. (We might want a gap-less version in the future as well)

drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
viktorklang: There is a crucial difference though, mergeSort sorts, this only sort-merges presorted.

drewhk: sorting merged streams predates merge-sort. It is merge sort that got the name from those merges and not the other way around.

viktorklang: Merge-sort should really be called sort-merge because you have to sort before you merge. #justsaying

drewhk: no need to sort, because one-element collections are always sorted, so you can argue that it starts with merge ;)

viktorklang: @drewhk LOL! Fair enough :)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphMergeSortedSpec.scala
rkuhn: No “pulling in” necessary; and ScalaCheck is a lot better than us at finding those corner cases—as witnessed in the sliding window operator fix Viktor did over the weekend. I think we should have collections-equivalence tests of this kind for all collection-like operators.

viktorklang: :+100:

drewhk: "pulling in" as in using it here. In more complex cases I almost agree with you, but  Gen.listOf(Gen.oneOf(false, true)) does not seem to do much here. My other problem is with ScalaCheck is that it can gives false confidence (see the sliding implementation). I would prefer to see the list of cases being checked than a black-box. Especially since these corner-cases allow in a possible rewrite to analyze what issues are to be faced.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
rkuhn: they are not exposed on any named type, so adding this keyword does not really change anything

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: WDYM? Gap-less cannot be ensured in general, it would be really specific and would need to reorder the streams—which implies buffering. “Picking the smallest” is exactly what it does, can you suggest a concrete improvement?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1372,19 +1370,42 @@ trait FlowOps[+Out, +Mat] {
...
+   * to be available).
+   *
+   * '''Emits when''' one of the inputs has an element available
```
rkuhn: This needs fixing.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: got it: added note on that it always needs to wait, even when not intuitively necessary

drewhk: yeah, the point was that the intuition leads to the wrong conclusion, therefore it needs extra communication from our side.

PR #19158, opened 2015-12-13
=htc,doc more docs on EntityStreamSizeException

ktoso:
This exception is something newcomers will run into quite early, when trying a file upload or something else when trying to stream some data (i.e. large entity), so it must be very clearly documented both in scaladoc and exception message as well I'd argue.


drewhk:
LGTM

ktoso:
Failure unrelated and I'll look into it (ConnectionPoolSpec)

Issue #19159, opened 2015-12-13
add fallback for "UTF8" encoding when the request encompasses an Accept-Charset header with a decimal priority


ktoso:

wlk:
This was fixed in #19160

ktoso:
Thanks for the pointer @wlk

PR #19135, opened 2015-12-09
Fix content type charset wildcard

thyandrecardoso:
This pull request adds a fallback for "UTF8" encoding when the request encompasses an `Accept-Charset` header with a decimal priority (<1.0 and >0.0). 

Previously, requests completing with binary types (an image for example), and coming with headers like `Accept-Charset *;q=0.6`, would fail with `UnacceptableResponseContentTypeException`. This fixes that.

PR #19160, opened 2015-12-13
=htc #19159 Adds falback for wildcard charset with decimal priority

ktoso:
Ports the PR https://github.com/akka/akka/pull/19135 by @belerophon to the right branch.

I investigated the code, and since @sirthias's recent rewrite of the ContentType models we're all good here, and the change proposed by that PR is not required. I am porting the PR however as the test is good to have.

Please sanity check if I missed something.


ktoso:
Resolves #19159

sirthias:
Yup, LGMT.

ktoso:
Thanks!

PR #19161, opened 2015-12-13
=htp stop case class extraction from throwing IllegalArgumentException

aosagie:
Hi. I noticed I was getting exceptions when using "as" to extract to a case class in which its requirements failed. This was contrary to the "ValidationRejection" the docs stated would happen.

I made several attempts not to use a nested method (i.e. validatedMap):
1) Passing an implicit tupler to the "as" method changed the API
2) Using a magnet that combined "ConstructFromTuple" and "Tupler" necessitated implicit conversion methods for all Tuples1-22.
3) Calling "val tupler = implicitly[Tupler[A]]"  within the "as" method prevented the compiler from knowing  "Tuple1[A]" and "tupler.Out" was the same thing.

Anyway, this is my first contribution to akka. Let me know if I messed anything up. Thanks!

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/BasicRouteSpecs.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/BasicRouteSpecs.scala
@@ -119,6 +119,17 @@ class BasicRouteSpecs extends RoutingSpec {
         responseAs[String] shouldEqual "Person(john,38)"
       }
     }
+    "reject if case class requirements fail" in {
+      case class MyValidNumber(i: Int) {
+        require(i > 10)
+      }
+
+      val abcPath = path("abc" / IntNumber).as(MyValidNumber)(echoComplete)
+
+      Get("/abc/5") ~> abcPath ~> check {
+        rejection shouldBe a[ValidationRejection]
+      }
+    }
   }
   "Dynamic execution of inner routes of Directive0" should {
     "re-execute inner routes every time" in {
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/Directive.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/Directive.scala
@@ -9,6 +9,7 @@ import akka.http.scaladsl.server.directives.RouteDirectives
 import akka.http.scaladsl.server.util._
 import akka.http.scaladsl.util.FastFuture
 import akka.http.scaladsl.util.FastFuture._
+import akka.http.impl.util._
 
 /**
  * A directive that provides a tuple of values of type `L` to create an inner route.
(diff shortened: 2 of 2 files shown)


ktoso:
OK TO TEST

ktoso:
Very interesting idea!
I have mixed feeling somehow, on one hand it's pretty cool and `require` used in case classes is always for validation I think, on the other one it may be somewhat surprising (and slow down routing in failure cases) hm.

Idea: perhaps this should by `tryAs[...]`, such that the "fail fast" scenario of `as` remains unchanged.

WDYT @sirthias @jrudolph

sirthias:
@aosagie Thanks for noticing, reporting and proposing a fix for this problem!
It seems to me that not having case class extraction catch IllegalArgumentExceptions was indeed an oversight in the port from spray.
In spray case class extraction works exactly as you expect it to, i.e. IllegalArgumentExceptions are caught and immediately converted into ValidationRejections.
Not dealing with them at all, as is appears to be the case right now, is definitely not right.

@ktoso Having this fix in would remove one potential stumbling point in the migration from spray, so initially I'd was in favor of merging as is, but in fact I think I prefer your `tryAs` idea in the end.
The difference between `as` and `tryAs` would be that in the `as` case the IAE would bubble up to the `ExceptionHandler`, whereas the `tryAs` would reject and thus continue traversal of the routing tree.
One could easily argue that a failing `as` really is an exceptional situation that should be dealt with by the `ExceptionHandler` and not be simply masked by a potentially succeeding route that happens to be chained in later.

So, after a bit of thought I think I'd vote for doing this:

1. Implement `tryAs` along the lines of the what @aosagie nicely already implemented.
    (Maybe we could simplify the implementation to something like this though (untested):

    ```scala
    def tryAs[A](constructor: ConstructFromTuple[L, A]): Directive1[A] =
      ExecutionDirectives.rejectOnIAE(tmap(constructor))

    object ExecutionDirectives extends ExecutionDirectives {
      private val rejectOnIAE: Directive0 =
        handleExceptions {
          ExceptionHandler {
            case ex: IllegalArgumentException ⇒ reject(ValidationRejection(e.getMessage.nullAsEmpty, Some(e)))
          }
        }
    }
    ```
    )

2. Change `as` to catch the IAE and rethrow as an `IllegalRequestException`, which the default `ExceptionHandler` would properly catch and transform.

With this we should have case class extraction properly covered.

WDYT, @aosagie, @ktoso, @jrudolph ?

drewhk:
Maybe rename then `as` to `strictlyAs` and use `tryAs` as you proposed (and of course document the difference).

rkuhn:
Hmm, given the intended semantics of [IllegalArgumentException](http://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html) I think that the PR is good as proposed: only that exception will trigger rejection while others bubble up. There should probably be a well-documented and easy way to make this rejection fail the routing on the spot (and if that is too wordy then strictAs sounds like a good workaround).

2beaucoup:
I also like it as is. This ensures consistency with the marshalling directives (see #15792).

ktoso:
Great. Let's go ahead with merging this one, then let's add `strictlyAs` and document the difference.
Would you like to contribute the `strictlyAs` directive as well @aosagie? That would be very awesome of you.

LGTM

ktoso:
Resolves https://github.com/akka/akka/issues/19167

aosagie:
@ktoso Sure, will do

ktoso:
Ouch – I just noticed, wrong branch :scream:
I'll port it to the right one, everything else about the PR was good, don't worry about it @aosagie.
Please target the next PR onto `release-2.3-dev`, that's where akka streams / http live, thanks!

PR #19166, opened 2015-12-14
=str - Adds minor touchups to Unfold & UnfoldAsync

viktorklang:
- Makes the classes final
- Sprinkles pixiedust
- Switches to expectMsg in tests

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
@@ -3,17 +3,20 @@
  */
 package akka.stream.scaladsl
 
-import scala.concurrent.Await
+import akka.testkit.DefaultTimeout
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.time.{ Span, Millis }
+import scala.concurrent.{ Future, Await }
 import scala.concurrent.duration._
-import scala.util.{ Success, Failure }
+import scala.util.Failure
 import scala.util.control.NoStackTrace
 import akka.stream.{ SourceShape, ActorMaterializer }
 import akka.stream.testkit._
-import akka.stream.impl.{ PublisherSource, ReactiveStreamsCompliance }
 
-class SourceSpec extends AkkaSpec {
+class SourceSpec extends AkkaSpec with DefaultTimeout with ScalaFutures {
 
   implicit val materializer = ActorMaterializer()
+  implicit val config = PatienceConfig(timeout = Span(timeout.duration.toMillis, Millis))
 
   "Single Source" must {
     "produce element" in {
@@ -212,10 +215,9 @@ class SourceSpec extends AkkaSpec {
 
   "Repeat Source" must {
     "repeat as long as it takes" in {
-      import GraphDSL.Implicits._
-      val result = Await.result(Source.repeat(42).grouped(10000).runWith(Sink.head), 1.second)
-      result.size should ===(10000)
-      result.toSet should ===(Set(42))
+      val f = Source.repeat(42).grouped(1000).runWith(Sink.head)
+      f.futureValue.size should ===(1000)
+      f.futureValue.toSet should ===(Set(42))
     }
   }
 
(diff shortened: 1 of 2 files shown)


akka-stream/src/main/scala/akka/stream/impl/Unfold.scala
```diff
@@ -16,13 +16,11 @@ import scala.util.{ Failure, Success, Try }
  * @tparam S state
  * @tparam E element
  */
-private[akka] class Unfold[S, E](s: S, f: S ⇒ Option[(S, E)]) extends GraphStage[SourceShape[E]] {
-
-  val out: Outlet[E] = Outlet("Unfold")
-
+private[akka] final class Unfold[S, E](s: S, f: S ⇒ Option[(S, E)]) extends GraphStage[SourceShape[E]] {
```
ktoso: Could you also add `INTERNAL API` to the comment while here?

ktoso:
LGTM

akka-stream/src/main/scala/akka/stream/impl/Unfold.scala
```diff
@@ -36,36 +30,27 @@ private[akka] class Unfold[S, E](s: S, f: S ⇒ Option[(S, E)]) extends GraphSta
...
- * @param s initial state
- * @param f unfold function
- * @tparam S state
- * @tparam E element
- */
-private[akka] class UnfoldAsync[S, E](s: S, f: S ⇒ Future[Option[(S, E)]]) extends GraphStage[SourceShape[E]] {
-
-  val out: Outlet[E] = Outlet("UnfoldAsync")
-
+  * INTERNAL API
```
viktorklang: @ktoso Done

ktoso: thanks!

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
```diff
@@ -223,44 +223,53 @@ class SourceSpec extends AkkaSpec {
...
-
+      import system.dispatcher
       val source = Source.unfold((0, 1)) {
         case (a, _) if a > 10000000 ⇒ None
         case (a, b)                 ⇒ Some((b, a + b) → a)
       }
-      val result = Await.result(source.runFold(List.empty[Int]) { case (xs, x) ⇒ x :: xs }, 1.second)
-      result should ===(expected)
+      source.runFold(List.empty[Int]) { case (xs, x) ⇒ x :: xs } pipeTo testActor
+      expectMsg(expected)
```
drewhk: why? The error reporting by `should ===` is usually better, and `Await` will rethrow the onError value here, which is easier to parse than it being wrapped in a Status.Failure that was got instead of an expected message.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
```diff
@@ -223,44 +223,53 @@ class SourceSpec extends AkkaSpec {
...
+    }
+
+    "terminate with a failure if there is an exception thrown" in {
+      import system.dispatcher
+      val t = new RuntimeException("expected")
+      val source = Source.unfold((0, 1)) {
+        case (a, _) if a > 10000000 ⇒ throw t
+        case (a, b)                 ⇒ Some((b, a + b) → a)
+      }
+      source.runFold(List.empty[Int]) { case (xs, x) ⇒ x :: xs } pipeTo testActor
```
drewhk: same here

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
```diff
@@ -223,44 +223,53 @@ class SourceSpec extends AkkaSpec {
...
+      import system.dispatcher
 
       val source = Source.unfoldAsync((0, 1)) {
         case (a, _) if a > 10000000 ⇒ Future.successful(None)
         case (a, b)                 ⇒ Future(Some((b, a + b) → a))
       }
-      val result = Await.result(source.runFold(List.empty[Int]) { case (xs, x) ⇒ x :: xs }, 1.second)
-      result should ===(expected)
+      source.runFold(List.empty[Int]) { case (xs, x) ⇒ x :: xs } pipeTo testActor
+      expectMsg(expected)
```
drewhk: same here, and also in subsequent cases

drewhk:
I am not sure about the test code changes. At least I would like to see an actual error output from both variants before decide which one to use.

viktorklang:
do you have that requirement for all tests written?

-- 
Cheers,
√

viktorklang:
because if expectMsg is bad at reporrting failures then that is a bug.

-- 
Cheers,
√

drewhk:
I meant the `Await + should=== ` rewrite to `pipeTo+expectMsg()`.

drewhk:
What is wrong with the original `Await + should===`? I don't get it. Also, scalatest displays more nicely when collections mismatch for example. I am not sure we would like to reimplement that logic.

viktorklang:
I dislike Await with a passion.
not only does one have to manually add the timeout everywhere and either
manually dilate it or tune it per test, as well as Await.result adds more
visual clutter.

futureValue seems like an improvement but I wouldd suggest a bit of test
suite TLC in 2016.

-- 
Cheers,
√

drewhk:
Please show the output error report in both cases when the test fails. Until then I am not convinced that those test changes are an improvement. And if you feel the output of that variant is a bug of expectMsg then please fix expectMsg first. `futureValue` is probably a better alternative though (given the options we have right now).

viktorklang:
@drewhk I switched to ScalaFutures. Pushed

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
```diff
@@ -223,44 +225,46 @@ class SourceSpec extends AkkaSpec {
...
-      val source = Source.unfoldInf((0, 1)) {
-        case (a, b) ⇒ (b, a + b) → a
-      }
-      val result = Await.result(source.take(36).runFold(List.empty[Int]) { case (xs, x) ⇒ x :: xs }, 1.second)
-      result should ===(expected)
+    "generate an unbounded fibonacci sequence" in {
+      Source.unfoldInf((0, 1))({ case (a, b) ⇒ (b, a + b) → a })
+        .take(36)
+        .runFold(List.empty[Int]) { case (xs, x) ⇒ x :: xs }
+        .futureValue should be(expected)
```
drewhk: isn't the `===` is the type safe one? Or it does not work with `futureValue`? (I haven't used it before)

viktorklang: I have no idea, in any case, my suggestion on TLC for the test suite still stands. Lots of gardening required.

viktorklang:
Merge?

rkuhn:
Hmm, I have done some of these already yesterday, but the added test case is good.

viktorklang:
@rkuhn Any reason why this wasn't merged?

ktoso:
Doesn't seem to me like consensus was reached above though?

rkuhn:
The reason was that it had conflicting changes—I now cherry-picked it and fixed it up manually. Will be part of an upcoming PR.

viktorklang:
Thanks!

On Tue, Dec 22, 2015 at 6:35 PM, Roland Kuhn <notifications@github.com>
wrote:

> Closed #19166 <https://github.com/akka/akka/pull/19166>.
>
> —
> <https://github.com/akka/akka/pull/19166#event-498733294>.
>



-- 
Cheers,
√

Issue #19067, opened 2015-12-02
AIOOB exception during shutdown of stage


jrudolph:
Don't have a reproducer right now:

```
[info] java.lang.ArrayIndexOutOfBoundsException: -1
[info] 	at akka.stream.impl.fusing.GraphInterpreter.processEvent(GraphInterpreter.scala:541)
[info] 	at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:509)
[info] 	at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$tryAbort(ActorGraphInterpreter.scala:438)
[info] 	at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$runBatch(ActorGraphInterpreter.scala:423)
[info] 	at akka.stream.impl.fusing.ActorGraphInterpreter$$anonfun$receive$1.applyOrElse(ActorGraphInterpreter.scala:346)
[info] 	at akka.actor.Actor$class.aroundReceive(Actor.scala:467)
[info] 	at akka.stream.impl.fusing.ActorGraphInterpreter.aroundReceive(ActorGraphInterpreter.scala:292)
[info] 	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:516)
[info] 	at akka.actor.ActorCell.invoke(ActorCell.scala:487)
[info] 	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:238)
[info] 	at akka.dispatch.Mailbox.run(Mailbox.scala:220)
[info] 	at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(AbstractDispatcher.scala:397)
[info] 	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
[info] 	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
[info] 	at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
[info] 	at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
```

drewhk:
The context is important here. What did you do when it happened?

jrudolph:
[...]

PR #19175, opened 2015-12-14
=str #19067: Fixed error reporting after stage is closed.

drewhk:


akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -663,10 +663,13 @@ private[stream] final class GraphInterpreter(
   private[stream] def fail(connection: Int, ex: Throwable): Unit = {
     val currentState = portStates(connection)
     if (Debug) println(s"$Name   fail($connection, $ex) [$currentState]")
-    portStates(connection) = currentState | (OutClosed | InFailed)
-    if ((currentState & InClosed) == 0) {
+    portStates(connection) = currentState | OutClosed
+    if ((currentState & (InClosed | OutClosed)) == 0) {
```
drewhk: This was the actual issue. We overloaded the internal queue since we tried to enqueue after the port has been already closed.

drewhk:
refs #19067

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -663,10 +663,13 @@ private[stream] final class GraphInterpreter(
   private[stream] def fail(connection: Int, ex: Throwable): Unit = {
     val currentState = portStates(connection)
     if (Debug) println(s"$Name   fail($connection, $ex) [$currentState]")
-    portStates(connection) = currentState | (OutClosed | InFailed)
-    if ((currentState & InClosed) == 0) {
+    portStates(connection) = currentState | OutClosed
```
rkuhn: please move this into the `else` branch

rkuhn:
LGTM (although now SSL failures are more chatty)

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: ok, makes sense

drewhk:
Why are they more chatty?

rkuhn:
as pointed out by PR validation failure :-)

drewhk:
? It didn't fail for me locally

drewhk:
Ah, in http. Darn.

drewhk:
The failure is the known one. That will be fixed next.

Issue #19133, opened 2015-12-09
Unexpected DeathPactException in remoting


ktoso:
We should investigate this.

INFO (11/23) 13:34:07 [stdout]: [ERROR] [11/23/2015 13:34:07.416] [RelaySystem-akka.remote.default-remote-dispatcher-6] [Remoting] Monitored actor [Actor[akka://RelaySystem/system/endpointManager/reliableEndpointWriter-akka.tcp%3A%2F%2FServerSystem%4010.3.1.102%3A2553-0/endpointWriter#-1184433023]] terminated (akka.actor.DeathPactException)

This was on 2.3.14

patriknw:
I think I see how this can happen. ReliableDeliverySupervisor doesn't handle `Terminated` in its `idle` state, because then it is supposed to not have any writer child actor. However, it can enter the `idle` state with a non-terminated writer if it receives an `Ungate` message in `gated` state.

`Ungated` is sent from the `EndpointManager` at two places.

I think we need to keep track of when writer is terminated and only act on the `Ungate` message when the writer has been terminated.

Does that sound right, @drewhk ?

drewhk:
Hm, but how can be there a writer if the connection is gated?

patriknw:
I guess it is in progress of being terminated. Stopping is async, you know.

PR #19178, opened 2015-12-14
=rem #19133 avoid DeathPactException race condition

patriknw:
- DeathPactException could occur if the ReliableDeliverySupervisor
  was gated but not yet received Terminated and got an Ungate message
  from the EndpointManager and thereby entered idle state, followed by
  receiving the Terminated message, which is not handled in idle

### DIFF
--- a/akka-remote/src/main/scala/akka/remote/Endpoint.scala
+++ b/akka-remote/src/main/scala/akka/remote/Endpoint.scala
@@ -213,7 +213,7 @@ private[remote] class ReliableDeliverySupervisor(
       if (bufferWasInUse) {
         if ((resendBuffer.nacked.nonEmpty || resendBuffer.nonAcked.nonEmpty) && bailoutAt.isEmpty)
           bailoutAt = Some(Deadline.now + settings.InitialSysMsgDeliveryTimeout)
-        context.become(gated)
+        context.become(gated(writerTerminated = false, earlyUngateRequested = false))
         currentHandle = None
         context.parent ! StoppedReading(self)
         Stop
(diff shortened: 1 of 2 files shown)


akka-remote/src/main/scala/akka/remote/Endpoint.scala
```diff
@@ -312,10 +313,15 @@ private[remote] class ReliableDeliverySupervisor(
...
-    case Terminated(_) ⇒
+  def gated(writerTerminated: Boolean): Receive = {
+    case Terminated(_) if !writerTerminated ⇒
       context.system.scheduler.scheduleOnce(settings.RetryGateClosedFor, self, Ungate)
+      context.become(gated(writerTerminated = true))
     case IsIdle ⇒ sender() ! Idle
+    case Ungate if !writerTerminated ⇒
+      // Ungate was sent from EndpointManager, but we must wait for Terminated first.
+      // Try again soon, it will be Terminated in just a moment.
+      context.system.scheduler.scheduleOnce(100.millis, self, Ungate)
```
rkuhn: we could also store the fact that Ungate was received in another boolean flag and then react accordingly on line 319

patriknw: good suggestion, I like that

patriknw:
@rkuhn updated according to your suggestion: 7af7374

akka-remote/src/main/scala/akka/remote/Endpoint.scala
```diff
@@ -3,6 +3,7 @@
  */
 package akka.remote
 
+import scala.concurrent.duration._
```
patriknw: ah, then this is not needed

akka-remote/src/main/scala/akka/remote/Endpoint.scala
```diff
@@ -312,10 +313,18 @@ private[remote] class ReliableDeliverySupervisor(
       writer forward s
   }
 
-  def gated: Receive = {
-    case Terminated(_) ⇒
-      context.system.scheduler.scheduleOnce(settings.RetryGateClosedFor, self, Ungate)
+  def gated(writerTerminated: Boolean, earlyUngateRequested: Boolean): Receive = {
+    case Terminated(_) if !writerTerminated ⇒
+      if (earlyUngateRequested)
+        self ! Ungate
```
patriknw: good to review my own mistakes, this should of course also do the become

patriknw:
@rkuhn fixed mistake in previous commit, the new one is: ff09ad0

rkuhn:
LGTM

akka-remote/src/main/scala/akka/remote/Endpoint.scala
```diff
@@ -312,12 +312,19 @@ private[remote] class ReliableDeliverySupervisor(
...
+      else
+        context.system.scheduler.scheduleOnce(settings.RetryGateClosedFor, self, Ungate)
+      context.become(gated(writerTerminated = true, earlyUngateRequested))
     case IsIdle ⇒ sender() ! Idle
     case Ungate ⇒
-      if (resendBuffer.nonAcked.nonEmpty || resendBuffer.nacked.nonEmpty) {
+      if (!writerTerminated) {
+        // Ungate was sent from EndpointManager, but we must wait for Terminated first.
+        context.become(gated(writerTerminated, earlyUngateRequested = true))
```
drewhk: maybe write `writerTerminated = false` which is the only value that is possible here, but more explicit when reading it.

akka-remote/src/main/scala/akka/remote/Endpoint.scala
drewhk: well, currently both if branches take the context become at L231

drewhk: ah, obsolete comment

drewhk:
LGTM

akka-remote/src/main/scala/akka/remote/Endpoint.scala
patriknw: ok

patriknw:
Refs #19133

PR #19181, opened 2015-12-14
=htp add explicit type annotations to public implicits where still missing

sirthias:
For some reason I'm still having implicit resolution troubles with the `formFieldDirectives`.
So I temporarily added [wartremover](https://github.com/puffnfresh/wartremover) to identify all the remaining implicits that don't carry an explicit type annotation.

There were quite a few, which this PR fixes.
Even with it doesn't immediately fix any known problems it might prevent us from more future issues.

/cc @jrudolph

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/server/directives/PathDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/server/directives/PathDirectives.scala
@@ -35,7 +35,7 @@ trait PathDirectives extends PathMatchers with ImplicitPathMatcherConstruction w
    * If matched the value extracted by the PathMatcher is extracted on the directive level.
    */
   def rawPathPrefix[L](pm: PathMatcher[L]): Directive[L] = {
-    implicit def LIsTuple = pm.ev
+    implicit val LIsTuple = pm.ev
     extract(ctx ⇒ pm(ctx.unmatchedPath)).flatMap {
       case Matched(rest, values) ⇒ tprovide(values) & mapRequestContext(_ withUnmatchedPath rest)
       case Unmatched             ⇒ reject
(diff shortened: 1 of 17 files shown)


akka-http/src/main/scala/akka/http/scaladsl/server/directives/PathDirectives.scala
```diff
@@ -83,7 +83,7 @@ trait PathDirectives extends PathMatchers with ImplicitPathMatcherConstruction w
    * order, i.e. `pathSuffixTest("baz" / "bar")` would match `/foo/bar/baz`!
    */
   def pathSuffixTest[L](pm: PathMatcher[L]): Directive[L] = {
-    implicit def LIsTuple = pm.ev
+    implicit val LIsTuple = pm.ev
```
ktoso: why is `val` better here?

ktoso:
Good idea, thanks @sirthias!

ktoso:
PLS BUILD

jrudolph:
+1 Good idea to use wartremover to find the remaining places. Hopefully the problems are gone now for good.

drewhk:
LGTM

2beaucoup:
:+1:

akka-http/src/main/scala/akka/http/scaladsl/server/directives/PathDirectives.scala
sirthias: Because it doesn't end up generating a method which reduces JAR size by a tiny bit... :)

ktoso: Jaj, true true :)

rkuhn:
Failure is unrelated, thanks @sirthias !

Issue #19176, opened 2015-12-14
Add Partition Graph Stage


israel:
Add a partition stage from 1 inlet to n outlets 
![drawing](https://cloud.githubusercontent.com/assets/239588/11787502/aa1c6cae-a294-11e5-9e14-51d20f486348.png)

Excuse my drawing skills... :-)

wlk:
it seems that this can be closed? can you confirm?

ktoso:
Right, closing

PR #19190, opened 2015-12-15
+str #19176 added a Partition Graph Stage enabling to target elements…

israel:


akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,91 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+ *   when one of the downstreams cancel
+ */
+
+class Partition[T](outputPorts: Int, partitioner: T ⇒ Int) extends GraphStage[UniformFanOutShape[T, T]] {
+
+  val in: Inlet[T] = Inlet[T]("Partition.in")
+  val out: Seq[Outlet[T]] = Seq.tabulate(outputPorts)(i ⇒ Outlet[T]("Partition.out" + i))
+  override val shape: UniformFanOutShape[T, T] = UniformFanOutShape[T, T](in, out: _*)
+  private val outPulled = Array.fill[Boolean](outputPorts)(false)
+  private val outPending = mutable.Map.empty[Int, T]
```
rkuhn: mutable state must be kept inside the GraphStageLogic, otherwise this module is not reusable

drewhk: (and not threadsafe either)

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,91 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+ * '''Cancels when'''
+ *   when one of the downstreams cancel
+ */
+
+class Partition[T](outputPorts: Int, partitioner: T ⇒ Int) extends GraphStage[UniformFanOutShape[T, T]] {
+
+  val in: Inlet[T] = Inlet[T]("Partition.in")
+  val out: Seq[Outlet[T]] = Seq.tabulate(outputPorts)(i ⇒ Outlet[T]("Partition.out" + i))
+  override val shape: UniformFanOutShape[T, T] = UniformFanOutShape[T, T](in, out: _*)
+  private val outPulled = Array.fill[Boolean](outputPorts)(false)
```
rkuhn: This can be removed since there is already `isAvailable(out(idx))` in the GraphStageLogic API.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
rkuhn: This map is used as a buffer of size one, so better just use a pair of variables (outlet and element).

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,91 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+          if (!isClosed(out(idx))) {
+            push(out(idx), elem)
+            outPulled(idx) = false
+            if (outPulled.exists(x ⇒ x) && !hasBeenPulled(in))
+              pull(in)
+          }
+        } else {
+          outPending.put(idx, elem)
+        }
+      }
```
rkuhn: upstream completion should not terminate the stage while outputs are pending

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,91 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+                outPulled(idx) = true
+                if (outPending.isEmpty && !hasBeenPulled(in)) {
+                  pull(in)
+                }
+            }
+          }
+
+          override def onDownstreamFinish(): Unit = {
+            completeStage()
+          }
```
rkuhn: downstream completion should not terminate the stage while other outputs are still open

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,91 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+      }
+    })
+
+    out.zipWithIndex.foreach {
+      case (o, idx) ⇒
+        setHandler(o, new OutHandler {
+          override def onPull() = {
+            outPending.get(idx) match {
+              case Some(elem) ⇒
+                if (!isClosed(o)) {
```
rkuhn: isClosed cannot be true during onPull

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,91 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+ *
+ * '''Backpressures when''' one of the outputs backpressure
+ *
+ * '''Completes when''' upstream completes
+ *
+ * '''Cancels when'''
+ *   when one of the downstreams cancel
+ */
+
+class Partition[T](outputPorts: Int, partitioner: T ⇒ Int) extends GraphStage[UniformFanOutShape[T, T]] {
```
rkuhn: we make all these classes `final` for now—removing that modifier is easier than adding it :-)

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,91 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+   * @param partitioner function deciding which output each element will be targeted
+   */
+  def apply[T](outputPorts: Int, partitioner: T ⇒ Int): Partition[T] = new Partition(outputPorts, partitioner)
+}
+
+/**
+ * Fan-out the stream to several streams. emitting an incoming upstream element to one downstream consumer according
+ * to the partitioner function applied to the element
+ *
+ * '''Emits when''' all of the outputs stops backpressuring and there is an input element available
```
rkuhn: Hmm, it should probably be “emits when the currently chosen output has demand”.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,91 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+ *
+ * '''Backpressures when''' one of the outputs backpressure
```
rkuhn: “ … when the currently chosen output back-pressures.”

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,91 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+ *
+ * '''Completes when''' upstream completes
+ *
+ * '''Cancels when'''
+ *   when one of the downstreams cancel
```
rkuhn: I think it would be more useful with “when all of the downstreams cancel”, e.g. when attaching a `take(5)` to one of the outputs.

akka-stream/src/main/scala/akka/stream/javadsl/Graph.scala
```diff
@@ -136,6 +136,40 @@ object Broadcast {
...
+ */
+object Partition {
+  /**
+   * Create a new `Partition` stage with the specified input type.
+   *
+   * @param outputCount number of output ports
+   * @param partitioner function deciding which output each element will be targeted
+   */
+  def create[T](outputCount: Int, partitioner: function.Function[T, Int]): Graph[UniformFanOutShape[T, T], Unit] =
+    scaladsl.Partition(outputCount, partitioner = (t: T) ⇒ partitioner.apply(t))
```
rkuhn: This explicit style is fine (and I’m not saying that it should be changed), but it is equivalent to just saying
~~~scala
scaladsl.Partition(outputCount, partitioner.apply)
~~~

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,91 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+          }
+
+          override def onDownstreamFinish(): Unit = {
+            completeStage()
+          }
+        })
+    }
+  }
+
+  override def toString = "Partition"
```
rkuhn: please add the number of output ports as well

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+  val settings = ActorMaterializerSettings(system)
+    .withInputBuffer(initialSize = 2, maxSize = 16)
+
+  implicit val materializer = ActorMaterializer(settings)
+
+  "A partition" must {
+    import GraphDSL.Implicits._
+
+    "partition to three subscribers" in assertAllStagesStopped {
+      val c1 = TestSubscriber.manualProbe[Int]()
```
drewhk: avoid manualProbe. It is rarely needed and it invites to do the wrong stuff. Use the normal probe instead.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+        partition.out(1) ~> Flow[Int].buffer(16, OverflowStrategy.backpressure) ~> Sink(c2)
+        ClosedShape
+      }).run()
+
+      val p1Sub = p1.expectSubscription()
+      val sub1 = c1.expectSubscription()
+      val sub2 = c2.expectSubscription()
+      sub1.request(3)
+      sub2.request(3)
+      p1.expectRequest(p1Sub, 16)
```
drewhk: no need for these if you use the normal probe instead of the manual one.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
drewhk: I highly recommend the detailed state diagrams in the docs: http://doc.akka.io/docs/akka-stream-and-http-experimental/2.0-M2/scala/stream-customize.html#custom-processing-with-graphstage

They enumerate the roles of isAvailable, hasBeenPulled, isClosed queries, and all available internal and external state transitions. It is fully tested and can be considered as a full specification of the port API.

rkuhn: Well, I think `isClosed` can actually be true in `onPush()`, no?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+import scala.concurrent.duration._
+
+class GraphPartitionSpec extends AkkaSpec {
+
+  val settings = ActorMaterializerSettings(system)
+    .withInputBuffer(initialSize = 2, maxSize = 16)
+
+  implicit val materializer = ActorMaterializer(settings)
+
+  "A partition" must {
```
rkuhn: In general there should be two kinds of tests:

* functional tests that do not use any testkit functionality and just verify that the right elements eventually end up in the right places
* specific tests around completion and cancellation that can only be done using `TestSink.probe` and `TestSource.probe`

rkuhn:
OK TO TEST

rkuhn:
Thanks for the contribution @israel ! We’ll work with you to see it through; an interesting second step would be to create a FlowOps combinator that makes use of this, potentially modeled after groupBy (i.e. using SubFlow), but that does not have to be done in this PR.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
drewhk: No, it should not, except if you closed it yourself. I.e. you called completeStage or complete on that port. Then of course it will report isClosed.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: done

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: this is by design, the other option is to discard elements targeted for the closed outputs. or to have both options set by a flag?

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: done

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: done

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: done

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: done

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: done

akka-stream/src/main/scala/akka/stream/javadsl/Graph.scala
israel: thx

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: done

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
rkuhn: Eager termination might be offered by a flag, but it should not be the default: something like `.take(4)` on a port is usually not indicative that the other ports shall finish as well. Elements for closed ports should just be dropped.

drewhk: Yes, Broadcast, Unzip, Balance proceed even if ports are cancelled. GroupBy also allows individiual substream cancellation, and so does splitWhen. I.e. this behavior is the "de facto" standard, optionally with additional flags.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,94 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+          override def onPull() = {
+            outPending match {
+              case Some((i, elem)) if (i == idx) ⇒
+                push(o, elem)
+                outPending = None
+                if (!hasBeenPulled(in)) {
+                  pull(in)
+                }
+              case None ⇒
+                if (outPending.isEmpty && !hasBeenPulled(in)) {
```
rkuhn: `outPending.isEmpty` is always `true` here

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,94 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    private var outPending: Option[(Int, T)] = None
+
+    setHandler(in, new InHandler {
+      override def onPush() = {
+        val elem = grab(in)
+        val idx = partitioner(elem)
+        if (isAvailable(out(idx))) {
+          if (!isClosed(out(idx))) {
```
rkuhn: This is the wrong order: if `isClosed` then nothing should happen, independent of what `isAvailable` says.

rkuhn: Oh, wrong: it should `pull(in)` when hitting a closed port.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,94 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+
+    out.zipWithIndex.foreach {
+      case (o, idx) ⇒
+        setHandler(o, new OutHandler {
+          override def onPull() = {
+            outPending match {
+              case Some((i, elem)) if (i == idx) ⇒
+                push(o, elem)
+                outPending = None
+                if (!hasBeenPulled(in)) {
```
rkuhn: This needs to be `if (!isClosed(in))` instead.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,94 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+                  pull(in)
+                }
```
rkuhn: … and if it was closed, then we need to `completeStage()`

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,94 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+              case None ⇒
+                if (outPending.isEmpty && !hasBeenPulled(in)) {
+                  pull(in)
+                }
+            }
+          }
+
+          override def onDownstreamFinish(): Unit = {
+            if (out.forall(isClosed(_)))
```
rkuhn: I’d rather replace this with a counter to avoid the traversal and the lambda allocation.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,94 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+              completeStage()
+            else {
+              outPending.foreach { case (i, _) if (i == idx) ⇒ outPending = None }
```
rkuhn: This would be more efficient with a pattern match instead.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -466,6 +466,94 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+    private var outPending: Option[(Int, T)] = None
+
+    setHandler(in, new InHandler {
+      override def onPush() = {
+        val elem = grab(in)
+        val idx = partitioner(elem)
+        if (isAvailable(out(idx))) {
+          if (!isClosed(out(idx))) {
+            push(out(idx), elem)
+            if (out.exists(x ⇒ isAvailable(x)) && !hasBeenPulled(in))
```
rkuhn: This traversal should be replaced with a counter as well: keep track of how many outputs have demand outstanding and pull when greater zero.

rkuhn: ah: but don’t pull if `outPending.nonEmpty`!

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: why it should always pull in such case? suppose we have 3 outs. first pulled then we pulled and got pushed and meanwhile the first is closed. none of the two other outs pulled... why should we pull again? I guess I should check if any of the other outs isAvailable and only then pull again.

ktoso:
refs https://github.com/akka/akka/issues/19176

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: re: the counter of how many outputs are pending, how can I know it current out was pending so I can subtract 1 from the counter. I understand that isPending is not relevant here so it seems I have to go back to pending flag for each out. (?)

rkuhn: You can initialize a variable with the number of output ports and subtract one for every `onDownstreamFinish` you get.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
rkuhn: Yes, that exception is correct.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: but how can I tell that the current leaving outport was actually pending? suppose a scenario where it was just pushed and then finished without pulling again (thus without raising the counter by one) why should I subtract one?

rkuhn:
@israel do you intend to keep working on this? If not, then we’ll have to close due to inactivity—you can always open again later.

israel:
@rkuhn Sure. I was just waiting for your response on https://github.com/akka/akka/pull/19190#discussion_r48198969 and you guys went on vacation ... :-)
I am ready to continue

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
rkuhn: This case here is only about terminating when all outputs are closed, whether they are pending or not (AFAICS).

rkuhn:
Ah, sorry, missed that one. Thanks for your prompt answer!

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: Maybe I am missing something, but, besides terminating when all outputs are closed, since I have a outPending counter, I have to update it when an OUT is closing and since I don't have a Boolean per OUT anymore, but rather a global counter, when an OUT is leaving I don't know whether it was pending or not since at this stage I cannot use isAvailable(out) ... hence I have to go back to a flag per out instead of the global pending counter.

rkuhn: Sounds like you should try out some approaches :-)

rkuhn:

Issue #19086, opened 2015-12-04
Add a helper method to Cluster extension to get full ActorPath


kosii:
Currently there is no other way to get the full path with address of an actor identified by `actorRef` than     

`ActorPath.fromString(actorRef.path.toStringWithAddress(Cluster(system).selfAddress))`

By adding a helper method to the Cluster extension class, we could do something much simpler, like this:

`Cluster(system).getPath(actorRef)`

Or if there is a way to programmatically get the `actorRef`'s system through some private API, then it would also be possible adding a method to the companion object:

`Cluster.getPath(actorRef)`

With the second method it would be possible to make sure that the user doesn't pass a wrong ActorSystem as parameter (I'm not sure it's a huge issue, in most cases there is only one system, right?)

What do you think, which one do you prefer?

ktoso:
Thanks for reporting

How about:

`Cluster(system)` // because consistently everything "in the extension", and you consciously obtain the extension
`.pathOf(actorRef)` // because somewhat nicer to read?

Somewhat related to: https://github.com/akka/akka/issues/17690

kosii:
Though I've read the other issue, and the conclusion was that the team can't add all methods that are nice to have. Is this feature nice enough?

PR #19192, opened 2015-12-15
Cluser extension helper to get full path of an actor ref

kosii:
#19086

### DIFF
--- a/akka-cluster/src/main/scala/akka/cluster/Cluster.scala
+++ b/akka-cluster/src/main/scala/akka/cluster/Cluster.scala
@@ -356,6 +356,20 @@ class Cluster(val system: ExtendedActorSystem) extends Extension {
     else
       clusterDaemons ! InternalClusterAction.AddOnMemberRemovedListener(callback)
   }
+
+  /**
+   * Generate the remote actor path by replacing the Address in the RootActor Path for the given
+   * ActorRef with the cluster's `selfAddress`, unless address' host is already defined
+   */
+  def remotePathOf(actorRef: ActorRef): ActorPath = {
+    val path = actorRef.path
+    if (path.address.host.isDefined) {
+      path
+    } else {
+      path.root.copy(selfAddress) / path.elements withUid path.uid
+    }
+  }
+
   // ========================================================
   // ===================== INTERNAL API =====================
   // ========================================================
--- a/akka-cluster/src/test/scala/akka/cluster/ClusterSpec.scala
+++ b/akka-cluster/src/test/scala/akka/cluster/ClusterSpec.scala
@@ -135,5 +135,12 @@ class ClusterSpec extends AkkaSpec(ClusterSpec.config) with ImplicitSender {
       }
     }
 
+    "allow to resolve remotePathOf any actor" in {
+      val remotePath = cluster.remotePathOf(testActor)
+
+      testActor.path.address.host should ===(None)
+      cluster.remotePathOf(testActor).uid should ===(testActor.path.uid)
+      cluster.remotePathOf(testActor).address should ===(selfAddress)
+    }
   }
 }


akka-cluster/src/main/scala/akka/cluster/Cluster.scala
```diff
@@ -356,6 +356,15 @@ class Cluster(val system: ExtendedActorSystem) extends Extension {
...
       clusterDaemons ! InternalClusterAction.AddOnMemberRemovedListener(callback)
   }
+
+  /**
+   * Generate the actor path by replacing the Address in the RootActor Path for the given
+   * ActorRef with the cluster's `selfAddress`, unless an address is already included
+   */
+  def pathOf(actorRef: ActorRef): ActorPath = {
+    ActorPath.fromString(actorRef.path.toStringWithAddress(selfAddress))
+  }
```
patriknw: if we were to add such an utility I think it should be on `ActorPath` itself, e.g. `path.withAddress(Cluster(system).selfAddress)`

rkuhn: Hmm, but then it would not be convenient (i.e. what you wrote) or not obvious that it uses the Cluster extension—and that is not even possible when doing it in akka-actor.

akka-cluster/src/main/scala/akka/cluster/Cluster.scala
```diff
@@ -356,6 +356,15 @@ class Cluster(val system: ExtendedActorSystem) extends Extension {
     else
       clusterDaemons ! InternalClusterAction.AddOnMemberRemovedListener(callback)
   }
+
+  /**
+   * Generate the actor path by replacing the Address in the RootActor Path for the given
+   * ActorRef with the cluster's `selfAddress`, unless an address is already included
+   */
+  def pathOf(actorRef: ActorRef): ActorPath = {
+    ActorPath.fromString(actorRef.path.toStringWithAddress(selfAddress))
```
rkuhn: This should not use a String as intermediate format, in particular when the passed-in ref is already a remote one. ActorPath has utilities for getting and replacing the contained Address already, IIRC.

akka-cluster/src/main/scala/akka/cluster/Cluster.scala
patriknw: sorry, I didn't think about the purpose of this utility and the name did not guide me.
We should name it something like `fillInMissingAddress`.
I think the parameter should be a `ActorPath` instead of an `ActorRef`.

rkuhn: Well, the point is to be actually nicer (and shorter) than the current way; what about `remotePathOf(ActorRef)`?

patriknw: deal, `remotePathOf` is nice

ktoso: me likes :+1:

akka-cluster/src/main/scala/akka/cluster/Cluster.scala
kosii: I'm not sure I found the best possible methods this time, but I tried to improve the code

rkuhn:
OK TO TEST

akka-cluster/src/main/scala/akka/cluster/Cluster.scala
rkuhn: Looking good!

kosii:
I squashed the commits

rkuhn:
merged manually after fixing the commit message typo

Thanks a lot @kosii !

kosii:
my pleasure!

PR #19194, opened 2015-12-16
WIP - DON'T MERGE - Set Source.inputStream() default initial and max buffer length to 1

### LINKED ISSUE #19193: Source.inputStream must emit bytes eagerly without batching
The default behaviour of `Source.inputStream()` is to block until the internal buffer is filled up.

Since we won't be able to tell the length of the data in advance, the sensible default behaviour is to send data as soon as it's available from the Source, instead of waiting until the buffer is filled.

This can be achieved at the moment with the following snippet - (thanks @2m):

```
Source.inputStream(() => input).withAttributes(Attributes.inputBuffer(1, 1))
```

We should have the `Attributes.inputBuffer(1, 1)` as the default behaviour.

fsat:
This is provide default behaviour where data flows down from the Source as soon as it's available since the length of the data from InputStream is not known beforehand.

Fixes #19193.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
@@ -324,7 +324,9 @@ object Source {
    * It materializes a [[Future]] containing the number of bytes read from the source file upon completion.
    */
   def inputStream(in: function.Creator[InputStream], chunkSize: Int): javadsl.Source[ByteString, Future[java.lang.Long]] =
-    new Source(scaladsl.Source.inputStream(() ⇒ in.create(), chunkSize)).asInstanceOf[Source[ByteString, Future[java.lang.Long]]]
+    new Source(scaladsl.Source.inputStream(() ⇒ in.create(), chunkSize))
+      .withAttributes(Attributes.inputBuffer(initial = 1, max = 1))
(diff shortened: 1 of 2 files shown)


drewhk:
OK TO TEST

drewhk:
Hmm, why does this matter?

drewhk:
I am not sure this is the right approach. Does the inputstream source reinterpret these buffer values?

fsat:
@drewhk - I guess this matters because we can't tell what's the length of the data from InputStream beforehand, so if the data is shorter then the default max size, the stream will block.

According to @rkuhn the proper way to do this is to fix the underlying `Source.inputStream()` implementation, which really means this PR may no longer be valid. If you do feel this is the case, please close this PR.

rkuhn:
Yes, we should really fix the root cause instead. Do you have time to work on this @fsat?

fsat:
@rkuhn - I'd like to give it a try. 

Would it be correct to set initialSize and maxSize of 1 instead of reading from materializer settings within the `InputStreamSource` itself:

https://github.com/akka/akka/blob/release-2.3-dev/akka-stream/src/main/scala/akka/stream/impl/io/IOSources.scala#L59

rkuhn:
Martynas has beaten you to it, see #19208.

Issue #18741, opened 2015-10-20
FAILED: ClusterClientSpec on release-2.3


patriknw:
https://jenkins.akka.io:8498/job/akka-release-2.3/219/consoleFull

Note that this is on the release-2.3 branch

patriknw:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/4570/consoleFull

bantonsson:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/4609/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/akka-release-2.3/225/consoleFull

patriknw:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/5138/consoleFull

PR #19199, opened 2015-12-16
=rem #19133 avoid DeathPactException race condition (for validation)

patriknw:
- DeathPactException could occur if the ReliableDeliverySupervisor
  was gated but not yet received Terminated and got an Ungate message
  from the EndpointManager and thereby entered idle state, followed by
  receiving the Terminated message, which is not handled in idle

(cherry picked from commit b6b498bd2c38e6aa90e507ec083a6a21fb30e25f)


patriknw:
Refs #19133

patriknw:
Failure unrealated https://github.com/akka/akka/issues/18741

patriknw:
PLS BUILD

PR #19202, opened 2015-12-16
!htc #15674 add Java side of Multipart model

### LINKED ISSUE #15674: Add Java side of multipart model
It seems I forgot those in the initial version.

/cc @sirthias

sirthias:
Closes #15674.

/cc @jrudolph @2beaucoup


akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
```diff
@@ -92,5 +92,41 @@
...
+    interface Chunked extends RequestEntity, ResponseEntity {
+        Source<ChunkStreamPart, ?> getChunks();
+    }
+
+    /**
+     * Represents an entity without a predetermined content-length to use in a BodyParts.
+     */
+    interface IndefiniteLength extends BodyPartEntity {
+    }
 }
```
ktoso: a welcome change, to group them here

akka-http-core/src/main/java/akka/http/javadsl/model/Multipart.java
```diff
@@ -0,0 +1,145 @@
...
+ */
+public interface Multipart {
+
+    MediaType.Multipart getMediaType();
+
+    Source<? extends Multipart.BodyPart, Object> getParts();
+
+    /**
+     * Converts this content into its strict counterpart.
+     * The given ``timeout`` denotes the max time that an individual part must be read in.
```
ktoso: just a single `, not in rst here ;)

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -263,6 +265,9 @@ object HttpEntity {
 
       s"$productPrefix($contentType,$dataAsString)"
     }
+
+    /** Java API */
+    def getData = data
```
ktoso: `override def` please

akka-http-core/src/main/java/akka/http/javadsl/model/Multipart.java
sirthias: Yup, will fix. Everywhere.

sirthias:
Applied feedback, fixed validation error and added commit replacing all double backticks in comments in htc and htp with single backticks.

ktoso:
LGTM, having a test which touches those would be nice, but those will come together with example docs of how to use this in the java dsl I think.

akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
```diff
@@ -92,5 +92,41 @@
...
+     */
+    interface CloseDelimited extends ResponseEntity {
+    }
+
+    /**
+     * Represents an entity transferred using `Transfer-Encoding: chunked`. It consists of a
+     * stream of {@link ChunkStreamPart}.
+     */
+    interface Chunked extends RequestEntity, ResponseEntity {
+        Source<ChunkStreamPart, ?> getChunks();
```
rkuhn: Please use BoxedUnit instead of `?`, type checking can be really nasty with existentials, also in Java.

akka-http-core/src/main/java/akka/http/javadsl/model/Multipart.java
```diff
@@ -0,0 +1,145 @@
...
+        Option<ContentDispositionType> getDispositionType();
+
+        Future<? extends Multipart.BodyPart.Strict> toStrict(long timeoutMillis, Materializer materializer);
+
+        interface Strict extends Multipart.BodyPart {
+            HttpEntity.Strict getEntity();
+        }
+    }
+
+    interface Strict extends Multipart {
```
rkuhn: I’d move this before `BodyPart`: it took me some seconds to see that this is a MultiPart and not a MultiPart.BodyPart subtype.

akka-http-core/src/main/scala/akka/http/scaladsl/model/Multipart.scala
```diff
@@ -43,20 +53,48 @@ sealed trait Multipart {
         .flatMapConcat(ConstantFun.scalaIdentityFunction)
     HttpEntity.Chunked(mediaType withBoundary boundary withCharset charset, chunks)
   }
+
+  /** Java API */
```
rkuhn: Please never use this comment: it is the most useless thing that can show up in the method summary of JavaDoc, and it is also asymmetric with the “Scala API”.

rkuhn: I think for now we’ll just have to remove these lines—at some point we’ll have to have ScalaDoc on all public methods!

rkuhn:
LGTM apart from comments

akka-http-core/src/main/java/akka/http/javadsl/model/Multipart.java
```diff
@@ -0,0 +1,145 @@
...
+     * Converts this content into its strict counterpart.
+     * The given `timeout` denotes the max time that an individual part must be read in.
+     * The Future is failed with an TimeoutException if one part isn't read completely after the given timeout.
+     */
+    Future<? extends Multipart.Strict> toStrict(long timeoutMillis, Materializer materializer);
+
+    /**
+     * Creates an entity from this multipart object.
+     */
+    RequestEntity toEntity(HttpCharset charset, String boundary);
```
2beaucoup: `HttpEntity`?

akka-http-core/src/main/scala/akka/http/scaladsl/model/Multipart.scala
jrudolph: Did Akka's policies wrt to this change at some point? In the optimal case, there wouldn't be any JavaDoc for this Scala file and the methods would completely be hidden in ScalaDoc. From the Scala side these methods are just inevitable and useless boilerplate. Wasn't that your original reasoning why we had to add those comments?

akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
```diff
@@ -92,5 +92,41 @@
      * Use getDataBytes and stream processing instead if the expected data is big or
      * is likely to take a long time.
      */
-    Future<HttpEntityStrict> toStrict(long timeoutMillis, Materializer materializer);
+    Future<HttpEntity.Strict> toStrict(long timeoutMillis, Materializer materializer);
+
+    /**
+     * The entity type which consists of a predefined fixed ByteString of data.
+     */
+    interface Strict extends UniversalEntity {
```
jrudolph: Nice, makes sense to move them in.

jrudolph:
LGTM

akka-http-core/src/main/scala/akka/http/scaladsl/model/Multipart.scala
rkuhn: Oh, right, I forgot about the slightly baroque way the Scala and Java APIs inherit each other in HTTP, please disregard the removal part. But who will add ScalaDoc to all the other methods?

akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
sirthias: Ok.

akka-http-core/src/main/java/akka/http/javadsl/model/Multipart.java
sirthias: Sure, no problem.

akka-http-core/src/main/java/akka/http/javadsl/model/Multipart.java
sirthias: I think `HttpEntity` would be too broad.
On the Scala side this is `MessageEntity`, which is an alias to `RequestEntity`, which is also a `ResponseEntity` (because it can also be used in responses).
Since we cannot have type aliases in Java, `RequestEntity` is the most specific one we can put here.

akka-http-core/src/main/scala/akka/http/scaladsl/model/Multipart.scala
sirthias: The Scala side inherits from the Java side, yes.
What would you consider baroque about that? (Apart from having a Java-side in the first place, for everything we do.)
I will add ScalaDoc to all the methods here.

rkuhn: I mostly meant that we’re not historically used to this scheme: Akka previously only had shared APIs, Streams have completely separate APIs, but HTTP has somewhat related APIs where the inheritance is going in the opposite direction as compared to Actor vs. UntypedActor. Anyway, this is good, we might even patch genjavadoc at some point so that it does not emit the Java API marked methods.

sirthias: Ok, I see. Thx!

ktoso:
Awaiting on small fixup of https://github.com/akka/akka/pull/19202#discussion_r47809131 and will merge.

sirthias:
Will so it tonight... (currently on the road)

--

sirthias:
Applied review feedback and rebased to resolve conflict.

ktoso:
Thanks a lot! Will merge after validation.

Issue #19196, opened 2015-12-16
Memory Leak within Akka HTTP low level API


fsat:
We've encountered memory leak when issuing HTTP request against an HTTP endpoint that is either down or not yet available. This behaviour is observed using Akka HTTP low level API.

The affected version is 2.0-M2.

We have a reproducer to show this behaviour:

https://github.com/fsat/akka-stream-http-test-stuff

The setup of the reproducer can be found at https://github.com/fsat/akka-stream-http-test-stuff/blob/master/README.md

rkuhn:
I suspect that the single-use outgoing connection fails to properly shut down all its actors, which should be simple to verify with a smaller test case using `assertAllStagesStopped`.

fsat:
Checked again with the reproducer after building `release-2.3-dev` with merged change - everything looks good, OOM has been fixed:

<img width="473" alt="akka-http-tcp-manager-fix" src="https://cloud.githubusercontent.com/assets/1665611/11887029/48e7e92a-a584-11e5-816c-1cb1518b1434.png">

rkuhn:
Thanks@

PR #19206, opened 2015-12-16
#19196 fix StageActorRef termination watch leak

rkuhn:
@drewhk One question remains: the leak actually was caused by the TcpManager actor retaining the GraphInterpreters via its watchedBy set—why is the TcpManager being watched, I wouldn’t expect it to ever terminate?

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
+++ b/akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
@@ -138,8 +138,16 @@ object GraphStageLogic {
     private[this] var behaviour = initialReceive
 
     /** INTERNAL API */
-    private[akka] def internalReceive(pack: (ActorRef, Any)): Unit =
-      behaviour(pack)
+    private[akka] def internalReceive(pack: (ActorRef, Any)): Unit = {
+      pack._2 match {
+        case Terminated(ref) ⇒
+          if (watching contains ref) {
+            watching -= ref
+            behaviour(pack)
+          }
+        case _ ⇒ behaviour(pack)
+      }
+    }
 
     override def !(message: Any)(implicit sender: ActorRef = Actor.noSender): Unit = {
       message match {
@@ -167,6 +175,7 @@ object GraphStageLogic {
       behaviour = receive
     }
 
+    private[this] var watching = ActorCell.emptyActorRefSet
     private[this] val _watchedBy = new AtomicReference[Set[ActorRef]](ActorCell.emptyActorRefSet)
 
     override def isTerminated = _watchedBy.get() == StageTerminatedTombstone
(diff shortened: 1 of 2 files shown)


rkuhn:
fixes #19196

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -138,8 +138,16 @@ object GraphStageLogic {
...
 
     /** INTERNAL API */
-    private[akka] def internalReceive(pack: (ActorRef, Any)): Unit =
-      behaviour(pack)
+    private[akka] def internalReceive(pack: (ActorRef, Any)): Unit = {
+      pack._2 match {
+        case Terminated(ref) ⇒
+          if (watching contains ref) {
+            watching -= ref
+            behaviour(pack)
```
viktorklang: So you *don't* want to run `behavior(pack)` is the ref is not in the watching set?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -167,6 +175,7 @@ object GraphStageLogic {
       behaviour = receive
     }
 
+    private[this] var watching = Set.empty[ActorRef]
```
viktorklang: Isn't the default set a HashSet? (I'm thinking a TreeSet might be more space efficient)

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -885,8 +907,10 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount:
   // Internal hooks to avoid reliance on user calling super in postStop
   /** INTERNAL API */
   protected[stream] def afterPostStop(): Unit = {
-    if (_stageActorRef ne null) _stageActorRef.stop()
-    _stageActorRef = null
+    if (_stageActorRef ne null) {
```
viktorklang: :+1:

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -219,18 +237,22 @@ object GraphStageLogic {
...
 
-    def watch(actorRef: ActorRef): Unit =
+    def watch(actorRef: ActorRef): Unit = {
+      watching += actorRef
       actorRef.asInstanceOf[InternalActorRef].sendSystemMessage(Watch(actorRef.asInstanceOf[InternalActorRef], this))
+    }
 
-    def unwatch(actorRef: ActorRef): Unit =
+    def unwatch(actorRef: ActorRef): Unit = {
+      watching -= actorRef
```
viktorklang: So these can never be called concurrently?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: Could be the cached `ActorCell.emptyActorRefSet`

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
ktoso: I guess the assumption is those are called from within the Stage, so it's synchronous and safe.
// spelling it out for my understanding as well

ktoso:
Fix makes sense, thanks!

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: Yes, these are the semantics from normal ActorCell.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: Yes, all modifications of the `watching` set happen from within the GraphStage.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
rkuhn: The default set is rather minimal for up to four entries and I switched to the cached one.

rkuhn:
git failure …

rkuhn:
PLS BUILD

drewhk:
I don't think it is a bad thing to watch the TcpManager actor. If it goes away because of a bug then all listener actors will just hang around, while currently they would kill themselves failing early.

drewhk:
LGTM otherwise

rkuhn:
Okay, that’s some reason—better be safe than sorry (even if unlikely).

drewhk:
I am not against removing it though if it makes things faster or eat less memory.

rkuhn:
Nah, let’s keep the change minimal right now.

PR #19208, opened 2015-12-17
=str #19193 remove buffer from InputStreamSource

### LINKED ISSUE #19193: Source.inputStream must emit bytes eagerly without batching

2m:
Fixes #19193

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
@@ -17,19 +17,17 @@ import scala.concurrent.Promise
 /** INTERNAL API */
 private[akka] object InputStreamPublisher {
 
-  def props(is: InputStream, completionPromise: Promise[Long], chunkSize: Int, initialBuffer: Int, maxBuffer: Int): Props = {
+  def props(is: InputStream, completionPromise: Promise[Long], chunkSize: Int): Props = {
     require(chunkSize > 0, s"chunkSize must be > 0 (was $chunkSize)")
-    require(initialBuffer > 0, s"initialBuffer must be > 0 (was $initialBuffer)")
-    require(maxBuffer >= initialBuffer, s"maxBuffer must be >= initialBuffer (was $maxBuffer)")
 
-    Props(classOf[InputStreamPublisher], is, completionPromise, chunkSize, initialBuffer, maxBuffer).withDeploy(Deploy.local)
+    Props(classOf[InputStreamPublisher], is, completionPromise, chunkSize).withDeploy(Deploy.local)
   }
 
   private final case object Continue extends DeadLetterSuppression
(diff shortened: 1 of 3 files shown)


akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
```diff
@@ -96,23 +59,20 @@ private[akka] class InputStreamPublisher(is: InputStream, bytesReadPromise: Prom
...
+        onCompleteThenStop()
 
       case _ ⇒
         readBytesTotal += readBytes
-        if (readBytes == chunkSize) availableChunks :+= ByteString1C(arr)
-        else availableChunks :+= ByteString1C(arr).take(readBytes)
 
-      // valid read, continue
+        // emit immediately, as this is the only chance to do it before we might block again
+        onNext(ByteString1C(arr).take(readBytes))
```
rkuhn: Isn’t it cheaper to keep using the same array of chunkSize for reading and copying into an appropriately-sized one before emitting? Especially for workloads where this change matters readBytes will frequently be significantly less than chunkSize, wasting a lot of space.

rkuhn:
Thanks @2m !

/cc @fsat

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
drewhk: Using the same array might be also more cache efficient. Not that it matters for IO though...

drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
2m: Let's see. With the shared array the worst case scenario is if the `chunkSize` is such that `is.read(..)` almost always reads fully, then the space wasted is the size of the shared array.

With no shared array wasted space is for every read of size `chunkSize - readBytes`. Which could be small but also does not have a limit.

It really looks like shared array is less wasteful.

2m:
Changed impl to have a shared buffer.

@viktorklang I remember we had a discussion against this before, but in the current light it looks like shared buffer approach produces at least limited memory waste.

drewhk:
The previous discussion was about shared array vs ByteString1C (which does not do a copy of the array).

PR #18654, opened 2015-10-06
=htc #18121 enable UTF-8 decoding for HTTP header values

sirthias:
If we detect a potential UTF-8 byte sequence we attempt decoding, otherwise we leave the bytes as they are.

Closes #18121.

/cc @jrudolph

PR #19209, opened 2015-12-17
=htc #16821 support UTF-8 characters in field values

### LINKED ISSUE #16821: Ensure that UTF-8 filenames in multipart/formdata file uploads are unmarshalled properly
See spray/spray#840

/cc @sirthias

jrudolph:
This alone seems to fix the immediate issue that we cannot parse UTF8 encoded filenames in `form-data/multipart`. The question is whether this change is not too broad (i.e. what other parsing rules are impacted) and e.g. whether we should guard against consuming half surrogate pairs in the parser.

The other issue is that we still cannot configure a specific charset to do header parsing and so we cannot introduce a configurable charset for multipart parsing as suggested in #16821. On the other hand, UTF8 should be common enough today that with this simple change we may already enable 90% of the use cases. If you control both the page containing the form and the backend receiving the data it should be no problem at all as you can make sure that the HTML page is encoded with UTF8.

Fixes #16821.

/cc @sirthias

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/model/parser/CommonRules.scala
@@ -42,7 +42,7 @@ private[parser] trait CommonRules { this: Parser with StringBuilding ⇒
 
   def qdtext = rule { `qdtext-base` | `obs-text` }
 
-  def `obs-text` = rule { "\u0080" - "\u00FF" }
+  def `obs-text` = rule { "\u0080" - "\uFFFE" }
 
   def `quoted-pair` = rule { '\\' ~ (`quotable-base` | `obs-text`) ~ appendSB() }
 
--- a/akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala
@@ -589,9 +589,10 @@ class HttpHeaderSpec extends FreeSpec with Matchers {
       parse("Foo", "ba\u0000r") shouldEqual ParsingResult.Error(ErrorInfo(
         "Illegal HTTP header value: Invalid input '\\u0000', expected field-value-char, FWS or 'EOI' (line 1, column 3)",
         "ba\u0000r\n  ^"))
-      parse("Flood-Resistant-Hammerdrill", "árvíztűrő ütvefúrógép") shouldEqual ParsingResult.Error(ErrorInfo(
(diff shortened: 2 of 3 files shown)


sirthias:
This change makes me a bit anxious as to its consequences.
I somehow expected https://github.com/akka/akka/pull/18654 to also fix this issue here.
Do you know why it doesn't (if it doesn't)?

jrudolph:
It's exactly because this change is missing, i.e. there were no parsing rules which would accept chars > 0xff.

jrudolph:
https://tools.ietf.org/html/rfc7230#section-3.2.4 says:

> Historically, HTTP has allowed field content with text in the ISO-8859-1 charset [ISO-8859-1], supporting other charsets only through use of [RFC2047] encoding.  In practice, most HTTP header  field values use only a subset of the US-ASCII charset [USASCII].  Newly defined header fields SHOULD limit their field values to US-ASCII octets.  A recipient SHOULD treat other octets in field content (obs-text) as opaque data.

So, that's exactly what `obs-text` is meant for: it collects data which cannot be otherwise parsed. Our current rule is wrong because we operate on UTF-16 characters and not on bytes. That's because we do UTF8 decoding *before* running the parser (so #18654 should have been the change to make you anxious :) ). A full solution would require to run the header parser infrastructure on arbitrary bytes so that we could decide later on how to decode particular fields which need to be interpreted with anyting else than USASCII. But this would be a far greater change and I'm not sure if it's worth the effort.

sirthias:
Ok, got it now.
Then this patch should indeed work as expected if we slightly constrain `obs-text` to:

```scala
def `obs-text` = rule { "\u0080" - "\uFFFE" }
```

to not also cover `EOI`.

In its current form pb2 simply cannot deal with text input potentially containing `'\uFFFF'` chars or "half-chars" (surrogate code points) with that value. However, since `'\uFFFF'` is a non-character (i.e. there are no valid characters that are represented by that value) and there are also no surrogate code points with that value (according to [this](http://docs.oracle.com/javase/6/docs/api/java/lang/Character.html#unicode) all surrogate code points fall in the range `0xD800 - 0xDFFF`) the change should not only be effective but also safe.

So my anxiety is gone... :)

rkuhn:
Phew, that’s good to hear—seeing @sirthias frightened by a change request is scary! :-)

drewhk:
LGTM but I have not much clue :)

jrudolph:
Ok, updated. For some reason the hammerdrill is now not only flood resistant but also parsing resistant. I guess that's a good thing.

ktoso:
Glad to have @sirthias to review such changes ;)
LGTM, thanks! Will merge after validation

Issue #18351, opened 2015-08-28
Implement TLS session renegotiation support for client certification authentication


jrudolph:
If someone knows a specification for this behavior please comment.

The scheme seems to be like this:
 * start TLS without requiring a client certificate
 * receive HTTP request
 * decide that an HTTP request is needed
 * send a TLS session renegotiation request to the client requiring a client certificate
 * wait for the client to provide the certificate
 * validate the certificate
 * send out the response

On the API level, we could introduce a special response header, that contains a client certificate validation function that when it is encountered in the renderer, first runs the TLS renegotiation, then calls the validation function with the new certificate, and only then sends out the response. In any case, it's tricky to get right because of the messages that need to be send around from the renderer to the parser and the other way round.

Part of #13749.

bblfish:
The above scheme makes sense. It is also good to have a certificate validation function, as that would allow schemes like WebID-TLS to implement more flexible validation than CA based ones. 

Btw. I am not sure what the parser needs to do here. The parser just needs to parse the certificate. 
After that validation gets going. The basic validation is that one tests to see if the Certificate is signed by a known CA. If that fails one might want to search for the CAs public key ( eg: by looking at the Issuer Alternative Name ). If that fails one might want to just verify the ownership of the public key by doing what is described in [WebID TLS](http://webid.info/spec). All of those things combined would reduce the reliance on CAs. 

What may be useful at the API level is to return the Certificate wrapped in some Verification object, which says what types of verifications the Certificate has undergone, because that is what would tell one what one can extract from it, and also how reliable it is.

jrudolph:
It seems this should be quite easy to implement by now if the validation can be configured at bind-time by providing the right SSLContext configuration to actually validate a certificate against a CA. In that case the only thing which needs to be done is issuing a `NegotiateNewSession` to request client authentication before sending the actual response. The SslTlsSupport will keep back the response until the new session (including the client authentication) was established and validated.

However, this only allows client authentication but not per-request authorization i.e. actually inspecting the client principal to figure out whether this generally accepted principal is also allowed to access a specific resource. To enable this we would need to access the new session between receiving the request and sending a response. Is this possible with the current SslTlsSupport implementation? I.e. will it send out a `SessionBytes(newSession, ByteString.empty)` after each successful handshake even if no other data is incoming at this point?

jrudolph:
[...]

PR #19210, opened 2015-12-17
+htc #18349 emit Tls-Session-Info header when configured on both client and server

### LINKED ISSUE #18349: Provide TLS session details in custom TlsSessionInfo header
See also #13749

jrudolph:
Fixes #18349

This works but needs documentation and Java-side.

/cc @sirthias

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/headers/TlsSessionInfo.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/headers/TlsSessionInfo.java
@@ -1,0 +1,25 @@
+/**
+ * Copyright (C) 2009-2016 Typesafe Inc. <http://www.typesafe.com>
+ */
+
+package akka.http.javadsl.model.headers;
+
+import javax.net.ssl.SSLSession;
+
+/**
+ * Model for the synthetic `Tls-Session-Info` header which carries the SSLSession of the connection
+ * the message carrying this header was received with.
+ *
+ * This header will only be added if it enabled in the configuration by setting
+ * <code>akka.http.[client|server].parsing.tls-session-info-header = on</code>.
+ */
+public abstract class TlsSessionInfo extends CustomHeader {
+    /**
+     * @return the SSLSession this message was received over.
+     */
+    public abstract SSLSession getSession();
+
+    public static TlsSessionInfo create(SSLSession session) {
+        return akka.http.scaladsl.model.headers.Tls$minusSession$minusInfo$.MODULE$.apply(session);
+    }
+}
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
@@ -100,8 +100,8 @@ private[http] object OutgoingConnectionBlueprint {
       val wrapTls = b.add(Flow[ByteString].map(SendBytes))
       terminationMerge.out ~> requestRendering ~> logger ~> wrapTls
 
-      val unwrapTls = b.add(Flow[SslTlsInbound].collect { case SessionBytes(_, bytes) ⇒ bytes })
-      unwrapTls ~> responseParsingMerge.in0
+      val collectSessionBytes = b.add(Flow[SslTlsInbound].collect { case s: SessionBytes ⇒ s })
+      collectSessionBytes ~> responseParsingMerge.in0
 
       methodBypassFanout.out(0) ~> terminationMerge.in0
 
@@ -113,7 +113,7 @@ private[http] object OutgoingConnectionBlueprint {
       BidiShape(
         methodBypassFanout.in,
         wrapTls.out,
-        unwrapTls.in,
+        collectSessionBytes.in,
         terminationFanout.out(1))
     })
 
@@ -154,8 +154,8 @@ private[http] object OutgoingConnectionBlueprint {
    * 2. Read from the dataInput until exactly one response has been fully received
    * 3. Go back to 1.
    */
-  class ResponseParsingMerge(rootParser: HttpResponseParser) extends GraphStage[FanInShape2[ByteString, HttpMethod, List[ResponseOutput]]] {
-    private val dataInput = Inlet[ByteString]("data")
+  class ResponseParsingMerge(rootParser: HttpResponseParser) extends GraphStage[FanInShape2[SessionBytes, HttpMethod, List[ResponseOutput]]] {
+    private val dataInput = Inlet[SessionBytes]("data")
     private val methodBypassInput = Inlet[HttpMethod]("method")
     private val out = Outlet[List[ResponseOutput]]("out")
 
@@ -174,7 +174,7 @@ private[http] object OutgoingConnectionBlueprint {
         override def onPush(): Unit = {
           val method = grab(methodBypassInput)
           parser.setRequestMethodForNextResponse(method)
-          val output = parser.onPush(ByteString.empty)
+          val output = parser.parseBytes(ByteString.empty)
           drainParser(output)
         }
         override def onUpstreamFinish(): Unit =
@@ -185,7 +185,7 @@ private[http] object OutgoingConnectionBlueprint {
       setHandler(dataInput, new InHandler {
         override def onPush(): Unit = {
           val bytes = grab(dataInput)
-          val output = parser.onPush(bytes)
+          val output = parser.parseSessionBytes(bytes)
           drainParser(output)
         }
         override def onUpstreamFinish(): Unit =
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
@@ -4,6 +4,10 @@
 
 package akka.http.impl.engine.parsing
 
+import javax.net.ssl.SSLSession
+
+import akka.stream.io.{ SessionBytes, SslTlsInbound }
+
 import scala.annotation.tailrec
 import scala.collection.mutable.ListBuffer
 import akka.parboiled2.CharUtils
@@ -30,11 +34,17 @@ private[http] abstract class HttpMessageParser[Output >: MessageOutput <: Parser
   private[this] var completionHandling: CompletionHandling = CompletionOk
   private[this] var terminated = false
 
+  private[this] var lastSession: SSLSession = null // used to prevent having to recreate header on each message
+  private[this] var tlsSessionInfoHeader: `Tls-Session-Info` = null
+  def initialHeaderBuffer: ListBuffer[HttpHeader] =
+    if (settings.includeTlsSessionInfoHeader && tlsSessionInfoHeader != null) ListBuffer(tlsSessionInfoHeader)
+    else ListBuffer()
+
   def isTerminated = terminated
 
-  val stage: PushPullStage[ByteString, Output] =
-    new PushPullStage[ByteString, Output] {
-      def onPush(elem: ByteString, ctx: Context[Output]) = handleParserOutput(self.onPush(elem), ctx)
+  val stage: PushPullStage[SessionBytes, Output] =
+    new PushPullStage[SessionBytes, Output] {
+      def onPush(input: SessionBytes, ctx: Context[Output]) = handleParserOutput(self.parseSessionBytes(input), ctx)
       def onPull(ctx: Context[Output]) = handleParserOutput(self.onPull(), ctx)
       private def handleParserOutput(output: Output, ctx: Context[Output]): SyncDirective =
         output match {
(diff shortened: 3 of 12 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -88,7 +88,7 @@ private[http] object OutgoingConnectionBlueprint {
       val wrapTls = b.add(Flow[ByteString].map(SendBytes))
       terminationMerge.out ~> requestRendering ~> logger ~> wrapTls
 
-      val unwrapTls = b.add(Flow[SslTlsInbound].collect { case SessionBytes(_, bytes) ⇒ bytes })
+      val unwrapTls = b.add(Flow[SslTlsInbound].collect { case s: SessionBytes ⇒ s })
```
ktoso: not really `unwrap` now it seems though?

drewhk:
Looks ok so far

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
jrudolph: Yes, I guess this needs a slight name change.

akka-http-core/src/test/scala/akka/http/impl/engine/client/TlsEndpointVerificationSpec.scala
```diff
@@ -21,7 +21,9 @@ import javax.net.ssl.SSLException
 
 class TlsEndpointVerificationSpec extends AkkaSpec("""
     akka.loglevel = DEBUG
-    akka.io.tcp.trace-logging = off""") with ScalaFutures {
+    akka.io.tcp.trace-logging = off
+    akka.http.client.parsing.tls-session-info-header = on
+    akka.http.server.parsing.tls-session-info-header = on""") with ScalaFutures {
```
ktoso: Could also be `akka.http.parsing.tls-session-info-header = on`, right? 
Due to how we resolve config (sanity checking that it works here too).

ktoso:
Looks good so far, awaiting Java parts :)

rkuhn:
Awesome initiative, would be great to have in 2.0!

jrudolph:
I added commits implementing requesting client certificates by renegotiation (#18351). This works but I think one test is still failing and the API is a bit rough (renegotiation is not really a mainstream feature, though).

The idea is that if you need a client certificate you return a special HttpResponse containing a `RequestClientCertificate` header. This response is never rendered but instead renegotiation is started. After a new session was established the session info is given to a function the user passed through `RequestClientCertificate` to generate the final response. That way you can return a response that is tailored to the actual peer principal.

See it here in action:

https://github.com/akka/akka/pull/19210/files#diff-4b8ee14211ee8f9a69f3803c9b846421R40

The API is not really nice but can be made infinitely better by creating a directive for it on the higher level.

An issue is that this feature currently adds a real cost to any response processing as we need to look for that special header. So, it would probably make sense to add other distinguishing elements to the HttpResponse which are cheaper to check for.

akka-http-core/src/test/scala/akka/http/scaladsl/RenegotiatingHttpsTestServer.scala
```diff
@@ -0,0 +1,75 @@
...
+
+        req.header[headers.`Tls-Session-Info`] match {
+          case Some(infoHeader) if infoHeader.info.peerPrincipal.isDefined ⇒
+            responseForPrincipal(infoHeader.info.peerPrincipal.get)
+          case _ ⇒
+            HttpResponse(headers = RequestClientCertificate { info ⇒
+              info.peerPrincipal match {
+                case Some(p) ⇒ responseForPrincipal(p)
+                case None    ⇒ HttpResponse(StatusCodes.Unauthorized, entity = "401 Requested peer certificate but didn't get any")
+              }
```
rkuhn: I’m not sure about this continuation: why not just use the same route again with new info next time? (or maybe I’m misunderstanding)

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -380,12 +400,12 @@ private[http] object HttpServerBluePrint {
   }
 
   private case class ProtocolSwitchShape(
```
rkuhn: ah, this will need to be adapted heavily: once #19243 is merged the websocket setup is gone and this stage is a lot cleaner

rkuhn:
For my limited understanding: on the client side no other changes need to be made since the engine is supposed to renegotiate, right? Will a new request be sent as well? The current continuation approach seems to imply that it is not possible to put the renegotiation trigger in a nice directive because the final result needs to be decided before pulling the trigger, do I understand this correctly?

jrudolph:
Right. We could write a directive but it wouldn't follow the usual semantics because it would be somewhat detached from the rest of the routing tree. The reason is that we cannot capture the complete "routing continuation" at some branch of the routing tree, i.e. route alternatives, rejection handlers, and exception handlers. So, the inner route (which would form the continuation in the RequestClientCertificate header) would need to be sealed separately and we would have to document that route alternatives, rejection handlers, and exception handlers wrapping that directive wouldn't work.

So, that's a good argument for the alternative to just rerun the complete request (changed to include the newly negotiated Tls credentials) through the user handler.

rkuhn:
Does this imply that the client is not expected to send the request again after renegotiation?

jrudolph:
Yes, exactly, that's how it seems to work.

rkuhn:
But that means that the request was not actually transmitted over the session that we require—which is very weird from a security perspective.

jrudolph:
Yes, and there have been security issues around this in the past. 

IIUC this has been fixed at some point in TLS by ensuring cryptographically that all sessions on a connection have been established by the same peer ([RFC 5746](http://tools.ietf.org/html/rfc5746)). IIRC you came to the same conclusion at some point but I couldn't find the comment.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/TlsSessionInfo.scala
```diff
@@ -0,0 +1,36 @@
+package akka.http.scaladsl.model.headers
+
+import java.security.Principal
+import java.security.cert.Certificate
+import javax.net.ssl.{ SSLPeerUnverifiedException, SSLSession }
+
+/** Allows access to an SSLSession with Scala types */
+trait TlsSessionInfo {
```
viktorklang: not sealed?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -760,6 +760,17 @@ final case class `Set-Cookie`(cookie: HttpCookie) extends jm.headers.SetCookie w
   protected def companion = `Set-Cookie`
 }
 
+object `Tls-Session-Info` {
+  def apply(session: SSLSession): `Tls-Session-Info` = `Tls-Session-Info`(TlsSessionInfo(session))
+}
+/** Virtual header that provides information about the SSL session the message was received over. */
+case class `Tls-Session-Info`(info: TlsSessionInfo) extends CustomHeader {
```
viktorklang: final

jrudolph:
I rebased the original PR, fixed a few things, and added the Java side. I removed the bits implementing #18351 to be resurrected in a separate PR.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -432,14 +435,14 @@ private[http] object HttpServerBluePrint {
...
-          override def onPush(): Unit = sourceOut.push(grab(fromNet))
+          override def onPush(): Unit = sourceOut.push(grab(fromNet).bytes)
         })
         sourceOut.setHandler(new OutHandler {
           override def onPull(): Unit = {
             if (!hasBeenPulled(fromNet)) pull(fromNet)
             cancelTimeout(timeoutKey)
             sourceOut.setHandler(new OutHandler {
-              override def onPull(): Unit = pull(fromNet)
+              override def onPull(): Unit = if (!hasBeenPulled(fromNet)) pull(fromNet)
```
jrudolph: For some reasons I saw tests failing here. Not sure why this PR would have changed the behavior.

akka-http-core/src/main/java/akka/http/javadsl/model/headers/TlsSessionInfo.java
```diff
@@ -0,0 +1,22 @@
...
+
+import javax.net.ssl.SSLSession;
+
+/**
+ * Model for the synthetic `Tls-Session-Info` header which carries the SSLSession of the connection
+ * the message carrying this header was received with.
+ *
+ * This header will only be added if it enabled in the configuration by setting
+ *
+ *     akka.http.(client|server).parsing.tls-session-info-header = on
```
ktoso: This indent causes it to render as <pre/>? Never seen that before hm...

I thought it's:

```
'''
akka.http.(client|server).parsing.tls-session-info-header = on
'''
```

We usually use `[client|server]`.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
```diff
@@ -46,6 +56,13 @@ private[http] abstract class HttpMessageParser[Output >: MessageOutput <: Parser
         if (self.onUpstreamFinish()) ctx.finish() else ctx.absorbTermination()
     }
 
+  final def onPush(input: SessionBytes): Output = {
```
ktoso: add `override` keyword here, so it's clear which one we want to be called, and the other one we just delegate to

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -57,16 +57,19 @@ private[http] object HttpServerBluePrint {
         requestPreparation(settings) atop
         controller(settings, log) atop
         parsingRendering(settings, log) atop
-        new ProtocolSwitchStage(settings, log) atop
-        unwrapTls
+        websocketSupport(settings, log) atop
+        tlsSupport
```

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -760,6 +761,27 @@ final case class `Set-Cookie`(cookie: HttpCookie) extends jm.headers.SetCookie w
   protected def companion = `Set-Cookie`
 }
 
+/**
+ * Model for the synthetic `Tls-Session-Info` header which carries the SSLSession of the connection
+ * the message carrying this header was received with.
+ *
+ * This header will only be added if it enabled in the configuration by setting
+ *
+ *     akka.http.(client|server).parsing.tls-session-info-header = on
```
ktoso: same question as above

akka-http-core/src/main/java/akka/http/javadsl/model/headers/TlsSessionInfo.java
jrudolph: Thanks for catching that. Will fix :)

akka-http-core/src/main/java/akka/http/javadsl/model/headers/TlsSessionInfo.java
```diff
@@ -0,0 +1,22 @@
+package akka.http.javadsl.model.headers;
```
ktoso: license header missing :<

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
ktoso: Yeah, I don't see why this PR specifically would have changed behaviour here hm... 
Ok to keep so for now. In general use of `hasBeenPulled` seems like a warning sign to me, maybe add a TODO at least here to research this at some point?

ktoso:
LGTM as far as I understand it.
Should we open a ticket for a directive for it right away and see how many people would actually want it?

jrudolph:
Thanks for the review, @ktoso. Fixed the issues you found.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpMessageParser.scala
jrudolph: This didn't override anything but I changed the names so it's clearer which method calls which.

jrudolph:
Botched the Javadoc formatting. Once again updated...

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -760,6 +761,29 @@ final case class `Set-Cookie`(cookie: HttpCookie) extends jm.headers.SetCookie w
...
 
+/**
+ * Model for the synthetic `Tls-Session-Info` header which carries the SSLSession of the connection
+ * the message carrying this header was received with.
+ *
+ * This header will only be added if it enabled in the configuration by setting
+ *
+ * ```
+ * akka.http.[client|server].parsing.tls-session-info-header = on
+ * ```
```
ktoso: Hm wait, previous was right I think - the `{{{}}}}`

ktoso: ScalaDoc – confusing people with syntax since 2004 :wink:

jrudolph: I think I tried and both worked :)

ktoso: Oh that's very cool then, thanks for checking (intellij didn't highlight it for me and the other one it did)

ktoso:
LGTM, about time to merge this one :)
Thanks again @jrudolph

jrudolph:
Yay!

Issue #19201, opened 2015-12-16
wrong documentation for max-pool-size-*


patriknw:
```
        # Minimum number of threads to cap factor-based max number to
        # (if using a bounded task queue)
        max-pool-size-min = 8

        # Max no of threads (if using a bounded task queue) is determined by
        # calculating: ceil(available processors * factor)
        max-pool-size-factor  = 3.0

        # Max number of threads to cap factor-based max number to
        # (if using a  bounded task queue)
        max-pool-size-max = 64
```

This has always made me think that max-pool-size-* properties does not matter if you use an unbounded task queue. That is wrong. The pool will be capped by the `maximumPoolSize`

I think we should automatically set the `maximumPoolSize` to at least the value of `corePoolSize`, so that a fixed size pool can be configured by the `core-pool-size-min` and `core-pool-size-max` settings.

richdougherty:
I agree that all the `(if using a bounded task queue)` comments are confusing. They probably mean `this is a useful setting if you want to bound the size of the thread pool`.

PR #19211, opened 2015-12-17
=act #19201 improve configuration of thread-pool-executor

patriknw:
- The old implementation would cap the pool size (both corePoolSize
  and maximumPoolSize) to max-pool-size, which is very confusing
  becuase maximumPoolSize is only used when the task queue is bounded.
- That resulted in configuring core-pool-size-min and core-pool-size-max
  was not enough, because it could be capped by the default max-pool-size.
- The new behavior is simply that maximumPoolSize is adjusted to not be
  less than corePoolSize, but otherwise the config properties match the
  underlying ThreadPoolExecutor implementation.
- Added a convenience fixed-pool-size property.

### DIFF
--- a/akka-docs/rst/java/cluster-sharding.rst
+++ b/akka-docs/rst/java/cluster-sharding.rst
@@ -349,11 +349,11 @@ Inspecting cluster sharding state
 ---------------------------------
 Two requests to inspect the cluster state are available:
 
-`ClusterShard.getShardRegionStateInstance` which will return a `ClusterShard.ShardRegionState` that contains
-the `ShardId`s running in a Region and what `EntityId`s are alive for each of them.
+``ClusterShard.getShardRegionStateInstance`` which will return a ``ClusterShard.ShardRegionState`` that contains
+the identifiers of the shards running in a Region and what entities are alive for each of them.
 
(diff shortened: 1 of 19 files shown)


patriknw:
Refs #19201

drewhk:
LGTM

akka-docs/rst/java/cluster-sharding.rst
```diff
@@ -349,11 +349,11 @@ Inspecting cluster sharding state
 ---------------------------------
 Two requests to inspect the cluster state are available:
 
-`ClusterShard.getShardRegionStateInstance` which will return a `ClusterShard.ShardRegionState` that contains
-the `ShardId`s running in a Region and what `EntityId`s are alive for each of them.
+``ClusterShard.GetShardRegionState`` which will return a ``ClusterShard.ShardRegionState`` that contains
```
ktoso: sure, for java?

patriknw: ah, I just copied, will fix

ktoso:
LGTM after comment on docs addressed

rkuhn:
LGTM

patriknw:
fixed the docs

PR #19212, opened 2015-12-17
Fusable SplitWhen

drewhk:
Second commit are changes I did during debugging failing HTTP tests and are not part of the intended solution.


akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -198,7 +198,8 @@ private[http] object OutgoingConnectionBlueprint {
 
       val getNextData = () ⇒ {
         waitingForMethod = false
-        pull(dataInput)
+        if (!isClosed(dataInput)) pull(dataInput)
```
drewhk: I am not sure if this change is correct, but this line sometimes results in trying to call a closed port otherwise.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSplitWhenSpec.scala
```diff
@@ -85,6 +87,16 @@ class FlowSplitWhenSpec extends AkkaSpec {
       }
     }
 
+    "not emit substreams if the parent stream is empty" in assertAllStagesStopped {
+
+      Await.result(
+        Source.empty[Int]
+          .splitWhen(_ ⇒ true).lift
+          .mapAsync(1)(_.runWith(Sink.headOption)).grouped(10).runWith(Sink.headOption),
+        3.seconds) should ===(None)
```
drewhk: If an empty substream is emitted this reqults in a `Some(None)`. A bit tricky test, but nice and compact.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1021,12 +1021,12 @@ trait FlowOps[+Out, +Mat] {
   def splitWhen(p: Out ⇒ Boolean): SubFlow[Out, Mat, Repr, Closed] = {
     val merge = new SubFlowImpl.MergeBack[Out, Repr] {
       override def apply[T](flow: Flow[Out, T, Unit], breadth: Int): Repr[T] =
-        deprecatedAndThen[Source[Out, Unit]](Split.when(p.asInstanceOf[Any ⇒ Boolean]))
-          .map(_.via(flow))
+        via(Split.when(p))
+          .map(_.buffer(16, OverflowStrategy.backpressure).via(flow))
```
drewhk: this buffer is in there only for debug purposes, should not be in the final solution.

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -413,4 +414,226 @@ final class PrefixAndTail[T](n: Int) extends GraphStage[FlowShape[T, (immutable.
...
+ */
+final class Split[T](decision: Split.SplitDecision, p: T ⇒ Boolean) extends GraphStage[FlowShape[T, Source[T, Unit]]] {
+  val in: Inlet[T] = Inlet("Split.in")
+  val out: Outlet[Source[T, Unit]] = Outlet("Split.out")
+  override val shape: FlowShape[T, Source[T, Unit]] = FlowShape(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new TimerGraphStageLogic(shape) {
+    import SubSource._
+    import Split._
+    private val SubscriptionTimer = "SubstreamSubscriptionTimer"
```
rkuhn: you don’t seem to actually schedule this timer

drewhk: oh right.

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -413,4 +414,226 @@ final class PrefixAndTail[T](n: Int) extends GraphStage[FlowShape[T, (immutable.
...
+
+      val onSubstreamPull = getAsyncCallback[Unit] { _ ⇒
+        if (!ignoreAsync) {
+          if (hasInitialElement) {
+            substream.pushSubstream(firstElem)
+            firstElem = null.asInstanceOf[T]
+            if (willCompleteAfterInitialElement) {
+              substream.completeSubstream()
+              ignoreAsync = true
+              completeStage()
```
rkuhn: setting `ignoreAsync = true` should be extraneous in all cases where it is followed by stage termination, right?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSplitWhenSpec.scala
```diff
@@ -173,126 +185,158 @@ class FlowSplitWhenSpec extends AkkaSpec {
...
-    val exc = TE("test")
-    val publisher = Source(publisherProbeProbe)
-      .splitWhen(elem ⇒ if (elem == 3) throw exc else elem % 3 == 0)
-      .lift
-      .withAttributes(ActorAttributes.supervisionStrategy(resumingDecider))
-      .runWith(Sink.publisher(false))
-    val subscriber = TestSubscriber.manualProbe[Source[Int, Unit]]()
-    publisher.subscribe(subscriber)
+    "fail substream if materialized twice" in assertAllStagesStopped {
+      // TODO: I don't know how to test this
```
drewhk: ok, I figured out

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSplitWhenSpec.scala
```diff
@@ -173,126 +185,158 @@ class FlowSplitWhenSpec extends AkkaSpec {
...
+      pending
+    }
 
-    val upstreamSubscription = publisherProbeProbe.expectSubscription()
+    "fail stream if substream not materialized in time" in assertAllStagesStopped {
+      // TODO: I don't know how to test this
```
drewhk: and this one too

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
drewhk: Yes. I just searched for `completeSubstream` and `failSubstream` and put these guards there.

drewhk:
Let's see what Jenkins has to say about this.

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
rkuhn: I’d prefer to keep it only where it has an effect, so as to not confuse the future reader.

drewhk: Of course. But let's fix all the bugs firts, until then I don't dare to touch too many things at once :)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSplitAfterSpec.scala
```diff
@@ -115,6 +116,12 @@ class FlowSplitAfterSpec extends AkkaSpec {
       }
     }
 
+    "work with single elem splits" in assertAllStagesStopped {
+      Await.result(
+        Source(1 to 10).splitAfter(_ ⇒ true).lift.mapAsync(1)(_.runWith(Sink.head)).grouped(10).runWith(Sink.head),
```
rkuhn: why not just use concatSubstreams?

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
rkuhn: Yes, this change is correct: getNextData can be called from methodBypassInput’s `onPush()` in which case a close event may already be pending but not yet delivered to the dataInput.

rkuhn: the same applies in getNextMethod as well

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSplitAfterSpec.scala
drewhk: because Sink.head fails on empty substreams.

rkuhn:
~~~scala
Source(1 to 10)
  .splitAfter(_ ⇒ true)
  .concatSubstreams
  .grouped(20)
  .runWith(Sink.head)
  .futureValue should ===(1 to 10)
~~~

rkuhn:
So, what’s up with the extra buffers? In the past the tests were just missing a `.request(1)` before expecting completion or failure in some places.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowSplitAfterSpec.scala
drewhk: No, that is not the same. Your code will pass if `splitAfter` would return empty substreams. My code does not, because it runs the `Sink.head` on the individual substreams

rkuhn: now _that_ definitely wants to be documented in a comment

drewhk: Yes, I already took note :)

drewhk:
ok, so tests passed Jenkins too with the extra buffers.

akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala
```diff
@@ -26,6 +26,7 @@ import akka.http.scaladsl.model._
 class ConnectionPoolSpec extends AkkaSpec("""
     akka.loggers = []
     akka.loglevel = OFF
+    akka.io.tcp.windows-connection-abort-workaround-enabled = on
```
drewhk: this should be `auto`, hence the failure on Jenkins

2beaucoup: this settings seems removed

jrudolph:
You can probably just comment the part in `MultipartUnmarshallers` that uses the IteratorInterpreter. I think it is just an optimization for Strict entities to avoid having to go through full materialization if all the data is already available.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/ParserOutput.scala
```diff
@@ -64,7 +65,11 @@ private[http] object ParserOutput {
   sealed abstract class EntityCreator[-A <: ParserOutput, +B >: HttpEntity.Strict <: HttpEntity] extends (Source[A, Unit] ⇒ B)
 
   final case class StrictEntityCreator(entity: HttpEntity.Strict) extends EntityCreator[ParserOutput, HttpEntity.Strict] {
-    def apply(parts: Source[ParserOutput, Unit]) = entity
+    def apply(parts: Source[ParserOutput, Unit]) = {
+      // We might need to drain stray empty tail streams which will be read by no one.
+      parts.runWith(Sink.ignore)(GraphInterpreter.currentInterpreter.subFusingMaterializer)
```
jrudolph: This now shows up with 15% in the profile.

drewhk: Unfortunately this was a bug, even in the non-fused version, it just needed a much rare race, mostly avoided by the buffers+prefetch. Currently this is deterministic due to lack of buffers and prefetch. This can be probably hacked to be faster, but I think there is a superior solution that you already hinted at in another PR.

I think the proper solution is to put the substream streaming right into the parser, avoiding the `splitWhen` dance altogether. The benefits would be immense:
 - not creating accidental empty substreams just to drop them (and materialize them), since the parser knows when to emit streams in the first place
 - getting rid of four stages: `splitWhen`, `prefixAndTail`, `collect`, and `concatSubstreams`. This means less memory, and less execution cost.
 - getting rid of two async callbacks: Currently one in `splitWhen`, `prefixAndTail` and `concatSubstreams`. These would be replaced with only one in the parser
 - getting rid of the materialization in `concatSubstreams`. This one would be huge alone.
 - strict entities would not need any substreams at all. This means that the cost of substreams would be only paid for entities that need it, strict entities would not involve any substreams at all (and these are very common for ordinary requests, so this is a hot path).
 - it would be possible to plant an `AsyncCallback` in non-strict entities that allow cancellation of the entity stream without materialization with `runWith(Sink.cancel)`. This can be very nice for automatic substream cleanup

One cost of this change would be that it complicates the parser. Creating a stream-of-streams `GraphStage` is not a simple task, as it is shown by `SplitWhen`. On the other hand this change would simplify the overall HTTP architecture quite a bit.

The other objection can be philosophical: creating custom stream-of-stream stages for each use case is something to be avoided in general, and not something we should advertise or practice, but the current solution here is both costly and also not "pure". I think until we find some more general, philosophically desirable solution, the performance benefits of this change waaaay outweigh the costs of purity here.

/cc @sirthias @rkuhn @ktoso

rkuhn: Thanks for the analysis, @drewhk, although it is slightly outdated in one regard: creating a substream in a GraphStage is now—while not completely trivial—quite simple actually (pending a PR from my side; example is the new impl of PrefixAndTail):
~~~scala
    private def subHandler = new OutHandler {
        cancelTimer(SubscriptionTimer)
        pull(in)
        tailSource.setHandler(new OutHandler {
        })
      }
    }

    private def openSubstream(): Source[T, Unit] = {
      val timeout = ActorMaterializer.downcast(interpreter.materializer).settings.subscriptionTimeoutSettings.timeout
      tailSource = new SubSourceOutlet[T]("TailSource")
      tailSource.setHandler(subHandler)
      scheduleOnce(SubscriptionTimer, timeout)
      builder = null
      Source.fromGraph(tailSource.source)
    }
~~~
The only sightly complex aspect is to not forget about the substream subscription timeout.

So, for me the conclusion is pretty clear, we’ll need to move this into the parser after 2.0.

drewhk: Totally awesome!!!

rkuhn:
taking over

PR #19229, opened 2015-12-18
=cls improve the graceful shutdown example

patriknw:


drewhk:
Hm, 3 seconds are enough? How to determine a safe value?

drewhk:
otherwise LGTM

ktoso:
LGTM, I'd also give perhaps 10s in the example?

patriknw:
Nothing bad happens if its shutdown too early, apart from a bit longer take-over for the singletons. I'll increase it anyway, thanks.

Issue #19228, opened 2015-12-18
Bump default TLS version for remoting


ktoso:
It's currently defaulting to `protocol = "TLSv1"` (because JDK6) but on master we don't care about 6 anymore. We should default to `TLSv1.2`, see also: https://github.com/ktoso/ssl-config/blob/master/core/src/main/resources/reference.conf#L25

The setting is `akka.remote.netty.ssl.security.protocol`

drewhk:
absolutely

PR #19230, opened 2015-12-18
=rem #19228 make the default TLS version 1.2

ktoso:
TLS 1.2 should be preferred and we always have it available by virtue of requiring JDK8 for Akka 2.4.x.
See also https://blogs.oracle.com/java-platform-group/entry/java_8_will_use_tls


ktoso:
Resolves #19228

drewhk:
LGTM

ktoso:
Found it in some more places, should now pass tests as well (tests checked which version is used)

PR #19243, opened 2015-12-20
fusable SplitWhen and reusable HTTP blueprint

rkuhn:
plus further optimizations


rkuhn:
@drewhk this is the collection of all work, I squashed and added credits to the commit message.

akka-http-core/src/main/scala/akka/http/scaladsl/model/Multipart.scala
```diff
@@ -187,10 +187,7 @@ object Multipart {
 
   private def strictify[BP <: Multipart.BodyPart, BPS <: Multipart.BodyPart.Strict](parts: Source[BP, Any])(f: BP ⇒ Future[BPS])(implicit fm: Materializer): Future[Vector[BPS]] = {
     import fm.executionContext
-    // TODO: move to Vector `:+` when https://issues.scala-lang.org/browse/SI-8930 is fixed
-    parts.runFold(new VectorBuilder[Future[BPS]]) {
-      case (builder, part) ⇒ builder += f(part)
-    }.fast.flatMap(builder ⇒ FastFuture.sequence(builder.result()))
+    parts.mapAsync(Int.MaxValue)(f).runWith(Sink.seq).fast.map(_.toVector)
```
rkuhn: @sirthias attempting to consume coupled sources strictly in sequence is likely to lead to deadlock; do you see a problem with allowing all parts to be consumed in parallel? Is this a possible remote DoS scenario?

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -675,7 +694,7 @@ private[stream] final class GraphInterpreter(
     val currentState = portStates(connection)
     if (Debug) println(s"$Name   complete($connection) [$currentState]")
     portStates(connection) = currentState | OutClosed
-    if ((currentState & (InClosed | Pushing | Pulling)) == 0) enqueue(connection)
+    if ((currentState & (InClosed | Pushing | Pulling | OutClosed)) == 0) enqueue(connection)
```
rkuhn: @drewhk this one and the following one are bugfixes: enqueueing the connection again may lead to queue overflow if the previous completion has not yet been processed

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -547,38 +548,38 @@ private[akka] final case class MapAsync[In, Out](parallelism: Int, f: In ⇒ Fut
       inheritedAttributes.getAttribute(classOf[SupervisionStrategy])
         .map(_.decider).getOrElse(Supervision.stoppingDecider)
 
-    val buffer = FixedSizeBuffer[Try[Out]](parallelism)
+    val buffer = new BoundedBuffer[Holder[Try[Out]]](parallelism)
```
rkuhn: This is necessary to allow `Int.MaxValue` as the parallelism bound.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: ok, I fixed a similar one in fail. Can you double-check that method, too?

rkuhn: Yes, I already did :-)

akka-http-core/src/main/scala/akka/http/scaladsl/model/Multipart.scala
sirthias: I'm not quite clear on the deadlock risk here. Can you explain?

rkuhn: Overlapping byte ranges may well be separated by more bytes than we are willing to buffer, which means that draining the first one is impossible without concurrently draining the other.

rkuhn: Hmm, I think I got that wrong for byte ranges since we make sure that they don’t overlap. But this code is not only for those, and a general `strictify` method must assume the streams to be coupled.

rkuhn:
Fixed multipart byte-range ordering, trying again …

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
ktoso: Right, thanks for fixing right away!

akka-http/src/main/scala/akka/http/scaladsl/server/directives/RangeDirectives.scala
```diff
@@ -73,12 +75,31 @@ trait RangeDirectives {
...
+                val flow = StreamUtils.sliceBytesTransformer(range.start, range.length)
+                bcast ~> flow.buffer(16, OverflowStrategy.backpressure).prefixAndTail(0).map {
+                  case (_, bytes) ⇒
+                    Multipart.ByteRanges.BodyPart(range.contentRange(length), HttpEntity(entity.contentType, range.length, bytes))
+                } ~> merge
+              }
+              entity.dataBytes ~> bcast
+              SourceShape(merge.out)
+            }
         }
```
2beaucoup: @rkuhn Perhaps we can generalize `Split` (combine `splitBefore` and `splitAfter`) to emit multiple sources and get rid of `sliceBytesTransformer`?

rkuhn: Yes, indeed; please file tickets for all such ideas!

2beaucoup: see #19253

akka-http-core/src/main/scala/akka/http/scaladsl/model/Multipart.scala
sirthias: Ok, so the deadlock would occur on `Multipart` instances that are created by the user and which have overlapping part streams, I see. Not a very probably scenario but not impossible, true.
The `mapAsync` is definitely the more generally correct solution here, thanks!

PR #19245, opened 2015-12-21
=cls increase the delay in the graceful shutdown example

patriknw:
as discussed in https://github.com/akka/akka/pull/19229, but it was merged before I changed


ktoso:
LGTM, thanks!

Issue #19240, opened 2015-12-19
rework request parser to directly emit entities


rkuhn:
The current split/prefixAndTail/concat dance is not only costly, it also is inherently broken due to the ease with which substreams are dropped (ignored, because presumably empty), thereby deadlocking the whole stream. For a discussion see [this comment](https://github.com/akka/akka/pull/19212#discussion_r48040654).

This will also mean that MultipartUnmarshallers will be vastly simplified and optimized.

2beaucoup:
:+1:

sirthias:
Yes, that is certainly the way to go, now that we have the cool new `GraphStage`, which was not available at the time the current HTTP stream structure was built.
Looking forward to the new perf numbers once this change is in!

drewhk:
Is this ticket really fully closed?

rkuhn:
I don’t think so.

ktoso:
No it isn't, closing it must have been an accident, sorry

jrudolph:
The problem that triggered this issue is now solved in another way which is cleaner IMO. Now we have two stages that split the work:

 * the parser is (mostly) responsible for interpreting the bytes
 * the preparation stage does the remaining work to present the data to the user

In any case, we don't created useless empty streams or streams for strict entities any more.

This split has the added benefit that it would allow a later, alternative entity-less and stream-of-streams-less API (more akin to spray's messaging API) that would deliver streamed messages as a sequence of dependent messages. As long as materialization is so slow this may be the only way to achieve reasonable performance also for streamed requests.

ataraxer:
The client side has been taken care of in #19528.

drewhk:
Hmm, is this solved now? The solution turned out to be somewhat different than what the original description states, but now both server and client are covered with an alternate solution. Is there something missing?

rkuhn:
Yes, I think this is solved now.

drewhk:
great! (moved to 2.4.2)

PR #19255, opened 2015-12-21
Fix performance of HTTP server (for strict entities at least)

jrudolph:
i.e. only for Strict entities where we can now completely get rid of per-request materialization.

This includes a fix for #19240 and also the corresponding issue on the rendering side.

This is just a quick try which makes a few tests fail probably because some boundary conditions need to be fixed properly.

With this patch performance is up to ~ 90k RPS (from 15k without the patch) on my machine.

/cc @sirthias

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpResponseRendererFactory.scala
@@ -6,6 +6,7 @@ package akka.http.impl.engine.rendering
 
 import akka.http.impl.engine.ws.{ FrameEvent, UpgradeToWebsocketResponseHeader }
 import akka.http.scaladsl.model.ws.Message
+import akka.stream.{ Outlet, Inlet, Attributes, FlowShape }
 
 import scala.annotation.tailrec
 import akka.event.LoggingAdapter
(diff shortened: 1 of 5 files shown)


jrudolph:
Cleaned up and rebased (also I added an issue for the second part: #19352).

It is interesting that you don't need to actually touch the stage functionality (on both sides). The only thing that is needed is some slight additions to the already existing GraphStages that handle the substreams properly. I wonder if some form of this would even be generic enough to be included in GraphStages themselves as helper methods.

On the rendering side the only thing needed would be an `emit(Source)` which would transfer all elements from the Source to the given Outlet.

On the parsing side, something similar is needed, i.e. which creates a substream, wraps that substream (in an HttpEntity in that case) and pushes it and then transfers input elements to that substream as long as (or until) some predicate holds.

jrudolph:
Here is the current benchmark with and without this PR:

RPS | Fusing enabled | What
------ |  -------------------- | ---------
23k  | no | akka-http 2.0.1
23k  | yes | akka-http 2.0.1
67k  | no | This PR
115k  | yes | This PR

It turns out that fusing has no effect in the original version because basically all the time is spent in materialization. Only after per-request materialization was removed the benefits of fusing can actually reap their benefits.

jrudolph:
The test failure was in `akka.stream.scaladsl.QueueSinkSpec`, so most likely unrelated.

ktoso:
Those look very promising, and also seem to overlap with what Roland observed before xmas when benchmarking (that things did not make much of a difference then). Thanks a lot!
I'll want to read the first commit here in detail before merging, will do that soon :)

ktoso:
PLS BUILD

ktoso:
Promising to review (and hope to merge) today.

ktoso:
Finally read through the entire thing, looks good! Thanks a lot and sorry for the delay @jrudolph!

jrudolph:
Cool, thanks for reviewing and merging, @ktoso.

Issue #19256, opened 2015-12-22
InputStreamPublisher should immediately emit data only when there's available demand


fsat:
I have encountered the following exception when using `StreamConverters.fromInputStream`:

```
2015-12-22T04:51:10Z 6135c7857c85 ERROR ActorSystemImpl ActorSystem(conductr) - Outgoing response stream error
java.lang.IllegalStateException: onNext is not allowed when the stream has not requested elements, totalDemand was 0
	at akka.stream.actor.ActorPublisher$class.onNext(ActorPublisher.scala:182) ~[com.typesafe.akka.akka-stream-experimental_2.11-2.0.jar:na]
	at akka.stream.impl.io.InputStreamPublisher.onNext(InputStreamPublisher.scala:30) ~[com.typesafe.akka.akka-stream-experimental_2.11-2.0.jar:na]
	at akka.stream.impl.io.InputStreamPublisher.readAndEmit(InputStreamPublisher.scala:67) ~[com.typesafe.akka.akka-stream-experimental_2.11-2.0.jar:na]
	at akka.stream.impl.io.InputStreamPublisher.readAndSignal(InputStreamPublisher.scala:49) ~[com.typesafe.akka.akka-stream-experimental_2.11-2.0.jar:na]
	at akka.stream.impl.io.InputStreamPublisher$$anonfun$receive$1.applyOrElse(InputStreamPublisher.scala:43) ~[com.typesafe.akka.akka-stream-experimental_2.11-2.0.jar:na]
	at akka.actor.Actor$class.aroundReceive(Actor.scala:467) ~[com.typesafe.akka.akka-actor_2.11-2.3-bin-rp-15v09p03i03.jar:na]
	at akka.stream.impl.io.InputStreamPublisher.akka$stream$actor$ActorPublisher$$super$aroundReceive(InputStreamPublisher.scala:30) ~[com.typesafe.akka.akka-stream-experimental_2.11-2.0.jar:na]
	at akka.stream.actor.ActorPublisher$class.aroundReceive(ActorPublisher.scala:309) ~[com.typesafe.akka.akka-stream-experimental_2.11-2.0.jar:na]
	at akka.stream.impl.io.InputStreamPublisher.aroundReceive(InputStreamPublisher.scala:30) ~[com.typesafe.akka.akka-stream-experimental_2.11-2.0.jar:na]
[...]

PR #19262, opened 2015-12-22
=htc,str #19256 InputStreamPublisher could emit without pending demnd

ktoso:
Fixes the bug found in https://github.com/akka/akka/issues/19256

This breaks HTTP for serving resources.
We need to decide when to roll an 2.0.1 as this breaks existing apps...

// I think I'll also want to look at https://github.com/typesafehub/ssl-config/issues/22 before rolling 2.0.1. Anything else?

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSourceSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSourceSpec.scala
@@ -3,17 +3,18 @@
  */
 package akka.stream.io
 
-import java.io.InputStream
+import java.io.{ OutputStream, PipedOutputStream, PipedInputStream, InputStream }
 import java.util.concurrent.CountDownLatch
 
-import akka.stream.scaladsl.Sink
-import akka.stream.scaladsl.StreamConverters
+import akka.stream.scaladsl.{ Source, Sink, StreamConverters }
 import akka.stream.testkit._
 import akka.stream.testkit.Utils._
 import akka.stream.testkit.scaladsl.TestSink
-import akka.stream.{ ActorMaterializer, ActorMaterializerSettings }
+import akka.stream.{ Attributes, OverflowStrategy, ActorMaterializer, ActorMaterializerSettings }
 import akka.util.ByteString
 import org.scalatest.concurrent.ScalaFutures
+import scala.concurrent.Await
+import scala.concurrent.duration._
 
 class InputStreamSourceSpec extends AkkaSpec(UnboundedMailboxConfig) with ScalaFutures {
 
@@ -21,6 +22,17 @@ class InputStreamSourceSpec extends AkkaSpec(UnboundedMailboxConfig) with ScalaF
   implicit val materializer = ActorMaterializer(settings)
 
   "InputStreamSource" must {
+
+    "not signal when no demand" in {
+      val f = StreamConverters.fromInputStream(() ⇒ new InputStream {
+        override def read(): Int = 42
+      })
+
+      Await.result(f
+        .takeWithin(5.seconds)
+        .runForeach(it ⇒ ()), 10.seconds)
+    }
+
     "read bytes from InputStream" in assertAllStagesStopped {
       val f = StreamConverters.fromInputStream(() ⇒ new InputStream {
         @volatile var buf = List("a", "b", "c").map(_.charAt(0).toInt)
--- a/akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
@@ -50,7 +50,7 @@ private[akka] class InputStreamPublisher(is: InputStream, bytesReadPromise: Prom
       if (totalDemand > 0) self ! Continue
     }
 
-  def readAndEmit() = try {
+  def readAndEmit(): Unit = if (totalDemand > 0) try {
     // blocking read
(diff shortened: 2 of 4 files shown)


ktoso:
// cc @fsat @huntc fix for the 2.0 bug.
// cc @rkuhn to decide when we should release (collect a few more this week and then?)... ping me on chat and I'll take it from there.

ktoso:
// cc @sirthias @jrudolph FYI, fix really simple and reproducer worked to repr each time.

ktoso:
FilePublisher does not suffer from this btw, I re-read it.

rkuhn:
LGTM

We should publish 2.0.1 tomorrow morning, gathering up some more fixes along the way.

ktoso:
Ok, I'll take care of it.

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSourceSpec.scala
```diff
@@ -21,6 +22,18 @@ class InputStreamSourceSpec extends AkkaSpec(UnboundedMailboxConfig) with ScalaF
...
   "InputStreamSource" must {
+
+    "not signal when no demand" in {
+      val f = StreamConverters.fromInputStream(() ⇒ new InputStream {
+        override def read(): Int = 42
+      })
+
+      val it = f
+        .takeWithin(5.seconds)
+        .runForeach(it ⇒ ())
```
drewhk: don't shadow `it`

drewhk:
Can we add a test case for resourceDirectives as well?

drewhk:
LGTM

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSourceSpec.scala
ktoso: right, good catch thanks

ktoso:
> Can we add a test case for resourceDirectives as well?

I was thinking about that... how would I mock such a resource that is long enough to reliably fail... I can play around and commit a very large file into the `resources/` path and then use hm...

ktoso:
Fixed the shadowing; not sure about testing in the resources directives, maybe I'm too tired now. Will keep open until morning.

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
```diff
@@ -50,7 +50,7 @@ private[akka] class InputStreamPublisher(is: InputStream, bytesReadPromise: Prom
       if (totalDemand > 0) self ! Continue
     }
 
-  def readAndEmit() = try {
+  def readAndEmit(): Unit = try if (totalDemand > 0) {
```
viktorklang: isn't it: `if (totalDemand > 0) try {` ?

rkuhn: What would be the difference? (assuming that the comparison cannot possibly result in throwing an exception)

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSourceSpec.scala
```diff
@@ -21,6 +22,17 @@ class InputStreamSourceSpec extends AkkaSpec(UnboundedMailboxConfig) with ScalaF
...
+
+    "not signal when no demand" in {
+      val f = StreamConverters.fromInputStream(() ⇒ new InputStream {
+        override def read(): Int = 42
+      })
+
+      Await.result(f
+        .takeWithin(5.seconds)
+        .runForeach(it ⇒ ()), 10.seconds)
+    }
```
2beaucoup: is an assert missing here?

ktoso: not really, this tests that it did not blow up – during the 5 seconds it was operating

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
ktoso: I'm happy to shuffle it around :wink:

ktoso:
Note to self... What about the tck on this publisher hmmmm

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSourceSpec.scala
```diff
@@ -3,17 +3,18 @@
...
+import akka.stream.scaladsl.{ Source, Sink, StreamConverters }
 import akka.stream.testkit._
 import akka.stream.testkit.Utils._
 import akka.stream.testkit.scaladsl.TestSink
-import akka.stream.{ ActorMaterializer, ActorMaterializerSettings }
+import akka.stream.{ Attributes, OverflowStrategy, ActorMaterializer, ActorMaterializerSettings }
 import akka.util.ByteString
 import org.scalatest.concurrent.ScalaFutures
+import scala.concurrent.Await
+import scala.concurrent.duration._
```
2beaucoup: are these needed?

ktoso: It could be via patience config but this is OK I believe (need to await longer than I allow it to run)

ktoso:
Added TCK test for InputStreamPublisher, it would not have found this one, it's rather specific... but rather safe than sorry

ktoso:
Failure was unrelated and known (ConnectionPoolSpec)

PR #19198, opened 2015-12-16
Early pull zip, concat and merge preferred stages

2m:

PR #19270, opened 2015-12-23
Eager pull fan-in stages.

ktoso:
In order to bring in https://github.com/akka/akka/pull/19198

### DIFF
--- a/akka-stream/src/main/boilerplate/akka/stream/scaladsl/ZipWithApply.scala.template
+++ b/akka-stream/src/main/boilerplate/akka/stream/scaladsl/ZipWithApply.scala.template
@@ -31,13 +31,22 @@ class ZipWith1[[#A1#], O] (zipper: ([#A1#]) ⇒ O) extends GraphStage[FanInShape
   ]
 
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
-    var pending = 1
+    var pending = ##0
     // Without this field the completion signalling would take one extra pull
     var willShutDown = false
 
     private def pushAll(): Unit = {
       push(out, zipper([#grab(in0)#]))
       if (willShutDown) completeStage()
+      else {
+        [#pull(in0)#
+        ]
+      }
+    }
+
+    override def preStart(): Unit = {
+      [#pull(in0)#
+      ]
     }
 
     [#setHandler(in0, new InHandler {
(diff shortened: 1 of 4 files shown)


ktoso:
Got reproducable failures locally... don't want to include it in 2.0.1

akka-stream/src/main/boilerplate/akka/stream/scaladsl/ZipWithApply.scala.template
```diff
@@ -56,17 +65,13 @@ class ZipWith1[[#A1#], O] (zipper: ([#A1#]) ⇒ O) extends GraphStage[FanInShape
 
     setHandler(out, new OutHandler {
       override def onPull(): Unit = {
-        pending = shape.inlets.size
-        if (willShutDown) completeStage()
-        else {
-            [#pull(in0)#
-            ]
-        }
+        pending += shape.inlets.size
```
viktorklang: `inlets` is an immutable.Seq so `size` could most definitely be `O(n)` so perhaps not do that for every one of these calls?

viktorklang: But looking at the FanInShape *impl* it does look like a Vector :)

akka-stream/src/main/boilerplate/akka/stream/scaladsl/ZipWithApply.scala.template
```diff
@@ -56,17 +65,13 @@ class ZipWith1[[#A1#], O] (zipper: ([#A1#]) ⇒ O) extends GraphStage[FanInShape
...
+        if (pending == ##0) pushAll()
       }
     })
   }
 
-  override def toString = "Zip"
+  override def toString = "ZipWith1"
```
viktorklang: ZipWith##0 ?

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -643,7 +640,16 @@ object Concat {
   /**
    * Create a new `Concat`.
    */
-  def apply[T](inputPorts: Int = 2): Concat[T] = new Concat(inputPorts)
+  def apply[T](inputCount: Int = 2) = GraphDSL.create() { implicit builder ⇒
+    import GraphDSL.Implicits._
+    val concat = builder.add(new Concat[T](inputCount))
+    val ds = concat.inSeq.map { inlet ⇒
+      val detacher = builder.add(new Detacher[T])
```
viktorklang: Isn't it possible to cache the Detacher instance and add it for every element?

akka-stream/src/main/boilerplate/akka/stream/scaladsl/ZipWithApply.scala.template
ktoso: Why? We want that number to be ZipWith2, ZipWith3, ZipWith4 etc.
`##0` hardcodes it to 0.

ktoso:
@2m is reviving his PR back in https://github.com/akka/akka/pull/19198 now,
closing this one.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
rkuhn: This is already done, it is called `def detacher[T]` now, which should be used here.

Issue #19260, opened 2015-12-22
Add ; charset=UTF-8 to application/json response headers


raam86:
Although the default encoding is utf-8 it is not being added to the response header rendering: 
  "×œ× ×™×•×ª×¨ ×ž-10"
instead of 
"לא יותר מ-10"

It is a minor issue since it only impacts application/json get requests in the browser (as in put the URI in the url field and hit enter)

ktoso:
Thanks for reporting, it seems like a bug, it definitely has the charset defined, it must be something off during rendering...

Could you try out defining and using the following instead of the default one:

```
val `application/json(UTF-8)` = MediaTypes.`application/json` withCharsetRange HttpCharsets.`UTF-8`
```

and ping this ticket if this made a difference?

Thanks in advance.

raam86:
Hi I've tried the following: 
```
ContentType(MediaTypes.`application/json`,HttpCharsets.`UTF-8`) 
```
with no difference 

Also you offer throws:
```
Error:(124, 65) value withCharsetRange is not a member of akka.http.scaladsl.model.MediaType
  val `application/json(UTF-8)` = MediaTypes.`application/json` withCharsetRange HttpCharsets.`UTF-8`
```
and using: 
```
 MediaTypes.`application/json`.withCharset(HttpCharsets.`UTF-8`)
```
Won't add the header either.

Some debugging shows that `application/json` doesn't have a `definedCharset` (in file akka.http.scaladsl.model.ContentType, line 69) and the constructor won't add it 

I
                                                                ^

ktoso:
Thanks a lot for checking, will have to dig into that.
When I skimmed it it seemed to me it has the charset defined:

```
val `application/json`                                                          = awfc("json", HttpCharsets.`UTF-8`, "json")
```

Sanity check, you're on `2.0` right?

raam86:
`val akkaStreamV = "2.0-M1"`

 in build.sbt

ktoso:
That's not supported. Please update to `2.0` :-)

raam86:
Doing it as I write. Will update.

ktoso:
I say that because we did a major rework of content types recently, I think the new version may be bug-free in this case :)
Thanks for reporting and trying to update, let us know how it went (if it's fixed on 2.0)

raam86:
Updated to v 2.0 and still same result. 

btw Source.fromPublisher was a little confusing. Where can I ask why it couldn't be added as another apply?

[...]

PR #19272, opened 2015-12-23
Added test to check if charset is set on application/json content type

raam86:
I looks like part of the problem is with that application/json is WithFIxedCharset 
Firstly `application/json` can have utf-16 and utf-32 charsets and secondly I cannot force charsets.

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/rendering/ResponseRendererSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/rendering/ResponseRendererSpec.scala
@@ -156,6 +156,24 @@ class ResponseRendererSpec extends FreeSpec with Matchers with BeforeAndAfterAll
               |
               |""", close = false)
       }
+
+      "add charset to application/json" in new TestSetup() {
+
+        ResponseRenderingContext(
+          requestMethod = HttpMethods.HEAD,
+          response = HttpResponse(
+            headers = List(Age(30)),
+            entity = HttpEntity.Default(ContentTypes.`application/json`, 100, Source.empty))) should renderTo(
+            """HTTP/1.1 200 OK
+            |Age: 30
+            |Server: akka-http/1.0.0
+            |Date: Thu, 25 Aug 2011 09:10:29 GMT
+            |Content-Type: application/json ; charset=UTF-8
+            |Content-Length: 100
+            |
+            |""", close = false)
+
+      }
     }
 
     "a response with a Strict body," - {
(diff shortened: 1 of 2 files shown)


raam86:
Relevant bug 
https://github.com/akka/akka/issues/19260#issuecomment-166866370

ktoso:
Thanks for the PR, initially I was thinking our rendering could have been wrong, but in light of @jrudolph's findings:

> Here's the bug report for chromium: https://code.google.com/p/chromium/issues/detail?id=438464
and for firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1166558

It indeed seems it is browser bugs... We'll need to think about what to do here. Technically it is not an Akka but, and Akka is abiding the RFC properly it seems.

Pinging @sirthias for feedback if he has time.

raam86:
I am not sure it is wrong per se, however there is an underlying issue where application/json can also have utf-32 charset but it is not possible to set since it's WithFixedCharset.

sirthias:
The current JSON spec says [this](http://tools.ietf.org/html/rfc7158#section-8.1):

> JSON text SHALL be encoded in UTF-8, UTF-16, or UTF-32.  The default
> encoding is UTF-8, and JSON texts that are encoded in UTF-8 are
> interoperable in the sense that they will be read successfully by the
> maximum number of implementations; there are many implementations
> that cannot successfully read texts in other encodings (such as
> UTF-16 and UTF-32).

So, while other unicode encoding are allowed we've not seen any actual use of these in all of spray's (akka-http's) life so far. That's why we went for the fixed UTF-8 encoding because that is the one that people will want to use in 99,999% of the time.

Given the browser pecularities we've decided to render an explicit `charset` parameter within the Content-Type, so that there are no ambiguities (even in the presence of incorrect default charsets on the peer's side). So, I'd say that akka-http does the right thing with regard to JSON de-/encoding at this point.

As to this PR, I think @raam86 is right in improving test coverage here. We should indeed add a test demonstrating that `application/json` content types are rendered in the way we want.
However, I could imagine that the `HttpHeaderSpec` would be the better (and easier) place to add it.
Basically we only need one more test case here: https://github.com/akka/akka/blob/release-2.3-dev/akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala#L194

raam86:
Can one of the repo owners verify this patch?

ktoso:
OK TO TEST

jrudolph:
LGTM

ktoso:
Thanks a lot @raam86!

ktoso:
LGTM

ktoso:
I squashed it into one commit and merged manually, thanks again @raam86!

Issue #19246, opened 2015-12-21
BackoffSupervisor: automatic and manual reset


RayRoestenburg:
(Description copied from post on akka user)
There seems to be no way for the child to tell the `BackoffSupervisor` that everything is ok again, the back-off is automatically reset based on a duration of no errors:

https://github.com/akka/akka/blob/v2.4.1/akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala#L204
https://github.com/akka/akka/blob/v2.4.1/akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala#L207-L208

The `minBackoff` is both used as minimal back-off duration and as this automatic reset duration. Which actually makes the `minBackoff` hard to set correctly, since it depends on the types of errors (and the speed at which the errors occur) in the child, and it makes it a bad idea to set it too low, since the back-off can then get stuck in repeating per `minBackoff` if the child does not crash fast enough the first time (within `minBackoff`), which means it will not properly back-off.

The user should be able to choose between an automatic reset and a manual reset (where the child sends a message to the supervisor to indicate success).

Also, there are currently two implementations of back-off supervisor, which should both reside in `akka.pattern`, the `BackoffSupervisor` props methods can create the correct implementation.

Discussion thread here: https://groups.google.com/d/msg/akka-user/rWNMM62fy9M/TT_5n4YDCgAJ

ktoso:
Thanks for collecting the info in a ticket

RayRoestenburg:
You're welcome. I'm starting on this today since I have some time before the holidays..

ktoso:
Great, thanks. I'm online most of december so will be will be able to review :)

RayRoestenburg:
Cool :-)

RayRoestenburg:
@ktoso so I'm thinking of making the following changes, which I want to run by you (let me know if you disagree with any of this so the review will cause a little less rework).

This is mostly about hiding both current `BackoffSupervisor` and `TransparentExponentialBackoffSupervisor` behind a 
`BackoffSupervisor` object that creates a back-off supervisor as required.

- The back-off supervisor is put in front of the child and behaves exactly like the child actor, so I want to drop the 
`GetCurrentChild` and `CurrentChild` messages (unless there is some use case for it I can't immediately think of?).

- The `BackoffSupervisor` takes a `SupervisorStrategy`, the `TransparentExponentialBackoffSupervisor` a `Decider`.
[...]

PR #19281, opened 2015-12-23
#19246 Auto and manual reset and `BackoffOptions`.

RayRoestenburg:
Also moved `TransparantExponentialBackoffSupervisor` to `akka.pattern` (renamed to `BackoffOnRestartSupervisor`)

I'm going to add more tests, but please have a look if you agree with the direction I've taken.
also I haven't contributed for a while, so if I break any rules, let me know.


RayRoestenburg:
Oh and I also still have to update the docs.

ktoso:
OK TO TEST

ktoso:
Great, thanks – will give it a look soon :)

ktoso:
Refs #19246

akka-actor-tests/src/test/scala/akka/pattern/BackoffSupervisorSpec.scala
```diff
@@ -70,5 +85,39 @@ class BackoffSupervisorSpec extends AkkaSpec with ImplicitSender {
         expectMsgType[BackoffSupervisor.CurrentChild].ref.get should !==(c1)
       }
     }
+
+    "support manual reset" in {
```
henrymai: Maybe parameterize this whole block so that both onStop() and onFailure() can both be tested

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
```diff
@@ -0,0 +1,115 @@
...
+  def waitingToStart(numRestarts: Int, scheduleCounterReset: Boolean): Receive = {
+    case StartChild ⇒
+      val childRef = actorOf(childProps, childName)
+      child = Some(childRef)
+      watch(childRef)
+      if (scheduleCounterReset) {
+        system.scheduler.scheduleOnce(minBackoff, self, ResetRestartCount(numRestarts + 1))
+      }
+      become(watching(childRef, numRestarts + 1))
+    case _ ⇒
```
henrymai: Remove this case

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  // Steady state
+  def watching(childRef: ActorRef, numRestarts: Int): Receive = {
+    case ScheduleRestart(`childRef`) ⇒
+      val delay = akka.pattern.BackoffSupervisor.calculateDelay(
+        numRestarts, minBackoff, maxBackoff, randomFactor)
+      system.scheduler.scheduleOnce(delay, self, StartChild)
+      become(waitingToStart(numRestarts, true))
```
henrymai: For the case that reset is ManualReset, do we still actually want to schedule the automatic counter reset?

RayRoestenburg: FYI AFK till 28th will respond then.

akka-actor-tests/src/test/scala/akka/pattern/BackoffOnRestartSupervisorSpec.scala
```diff
@@ -66,7 +65,7 @@ class TransparentExponentialBackoffSupervisorSpec extends AkkaSpec {
     val child = probe.lastSender
   }
 
-  "TransparentExponentialBackoffSupervisor" must {
+  "BackoffOnRestartSupervisor" must {
```
patriknw: good name

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
```diff
@@ -0,0 +1,115 @@
...
+        log.debug(s"Last restart count [$last] matches current count; resetting")
+        become(watching(childRef, 0))
+      } else {
+        log.debug(s"Last restart count [$last] does not match the current count [$numRestarts]")
+      }
+
+    case BackoffSupervisor.Reset ⇒
+      reset match {
+        case ManualReset ⇒ become(watching(childRef, 0))
+        case msg         ⇒ context.system.deadLetters.forward(msg)
```
patriknw: perhaps use `unhandled` instead?

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
```diff
@@ -0,0 +1,115 @@
...
+        system.scheduler.scheduleOnce(minBackoff, self, ResetRestartCount(numRestarts + 1))
+      }
+      become(watching(childRef, numRestarts + 1))
+    case _ ⇒
+  }
+
+  // Steady state
+  def watching(childRef: ActorRef, numRestarts: Int): Receive = {
+    case ScheduleRestart(`childRef`) ⇒
+      val delay = akka.pattern.BackoffSupervisor.calculateDelay(
```
patriknw: `akka.pattern` not needed

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
```diff
@@ -0,0 +1,115 @@
...
+  reset: BackoffReset,
+  randomFactor: Double,
+  strategy: OneForOneStrategy)
+  extends Actor
+  with ActorLogging {
+
+  import BackoffOnRestartSupervisor._
+  import context._
+
+  private var child: Option[ActorRef] = None
```
patriknw: is this field needed?

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
```diff
@@ -0,0 +1,233 @@
...
+   * the stored `ActorRef`. In general all messages to the child should be directed through this actor.
+   *
+   * An example of where this supervisor might be used is when you may have an actor that is
+   * responsible for continuously polling on a server for some resource that sometimes may be down.
+   * Instead of hammering the server continuously when the resource is unavailable, the actor will
+   * be restarted with an exponentially increasing back off until the resource is available again.
+   *
+   * '''***
+   * This supervisor should not be used with `Akka Persistence` child actors.
+   * `Akka Persistence` actors, currently, shutdown unconditionally on `persistFailure()`s rather
```
patriknw: the word "currently" is not needed, we have no plans on changing and software is always changing

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
```diff
@@ -0,0 +1,233 @@
...
+   * than throw an exception on a failure like normal actors.
+   * [[akka.pattern.BackoffSupervisor]] should be used instead for cases where the child actor
```
patriknw: the link to `akka.pattern.BackoffSupervisor` is probably obsolete

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
```diff
@@ -0,0 +1,233 @@
...
+   *
+   * The `BackoffSupervisor`delegates all messages from the child to the parent of the
+   * `BackoffSupervisor`, with the supervisor as sender.
+   *
+   * The `BackoffSupervisor` forwards all other messages to the child, if it is currently running.
+   *
+   * The child can stop itself and send a [[akka.actor.PoisonPill]] to the parent supervisor
+   * if it wants to do an intentional stop.
+   *
+   * Exceptions in the child are handled with the given `supervisionStrategy`. A
```
patriknw: perhaps some hint about `withSupervisorStrategy` instead?

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
```diff
@@ -0,0 +1,233 @@
...
+   */
+  def withDefaultStoppingStrategy: BackoffOptions
+
+  /**
+   * Returns the props to create the back-off supervisor.
+   */
+  private[akka] def props: Props
+}
+
+private case class BackoffOptionsImpl(
```
patriknw: `final` for case classes

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
```diff
@@ -96,6 +105,24 @@ object BackoffSupervisor {
...
+  /**
+   * Java API: Send this message to the [[BackoffSupervisor]] and it will reply with
+   * [[BackoffSupervisor.RestartCount]] containing the current restart count.
+   */
+  def getRestartCount = GetRestartCount
+
+  final case class RestartCount(count: Int)
+
   private case object StartChild extends DeadLetterSuppression
   private case class ResetRestartCount(current: Int) extends DeadLetterSuppression
```
patriknw: final?

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
```diff
@@ -120,58 +147,42 @@ object BackoffSupervisor {
...
- *
- * The child can stop itself and send a [[akka.actor.PoisonPill]] to the parent supervisor
- * if it wants to do an intentional stop.
- *
- * Exceptions in the child are handled with the given `supervisionStrategy`. A
- * `Restart` will perform a normal immediate restart of the child. A `Stop` will
- * stop the child, but it will be started again after the back-off duration.
- */
-final class BackoffSupervisor(
+private final class BackoffSupervisor(
```
patriknw: I don't think we can make it private because of backwards compatibility.
Some link in the docs of this class would be nice.

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
```diff
@@ -201,12 +212,25 @@ final class BackoffSupervisor(
...
+      reset match {
+        case AutoReset(resetBackoff) ⇒
+          val _ = context.system.scheduler.scheduleOnce(resetBackoff, self, ResetRestartCount(restartCount))
+        case _ ⇒
+      }
+
+    case Reset ⇒
+      reset match {
+        case ManualReset ⇒ restartCount = 0
+        case msg         ⇒ context.system.deadLetters.forward(msg)
```
patriknw: unhandled

patriknw:
only minor nitpicks, great work!

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
```diff
@@ -1,203 +0,0 @@
-/**
```
henrymai: Can you have a the movement of this file from:

`akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala` to
`akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala`

be in a separate commit that comes before this commit (in the same pull request)?
This way we can see the changes between the two versions more easily.

RayRoestenburg: Sure no problem will fix tomorrow

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
RayRoestenburg: I'll change it to 
```
case msg => unhandled(msg)
```
So it gets reported and not blackholed.

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
RayRoestenburg: Ah right, I'll change to 
```scala
become(waitingToStart(numRestarts, reset != ManualReset))
```

akka-actor-tests/src/test/scala/akka/pattern/BackoffOnRestartSupervisorSpec.scala
RayRoestenburg: thanks

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
RayRoestenburg: Ok

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
RayRoestenburg: Ok

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
RayRoestenburg: Ah, I could just use 
```scala
child(childName)
```

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
RayRoestenburg: Ok

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
RayRoestenburg: Good catch

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
RayRoestenburg: Ok

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
RayRoestenburg: Ah missed that one, Ok.

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
RayRoestenburg: Ok.

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
RayRoestenburg: Ok.

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
RayRoestenburg: Ok.

akka-actor-tests/src/test/scala/akka/pattern/BackoffSupervisorSpec.scala
RayRoestenburg: I'll add more tests.

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
```diff
@@ -0,0 +1,115 @@
...
+        log.debug(s"Last restart count [$last] does not match the current count [$numRestarts]")
+      }
+
+    case BackoffSupervisor.Reset ⇒
+      reset match {
+        case ManualReset ⇒ become(watching(childRef, 0))
+        case msg         ⇒ unhandled(msg)
+      }
+    case BackoffSupervisor.GetCurrentChild ⇒
+      sender() ! BackoffSupervisor.CurrentChild(child(childName))
```
henrymai: Why not just: CurrentChild(Some(childRef))

- The child can always end up being shutdown after a response to GetCurrentChild.
- We should generally not be in the watching() state when the child is shutdown. See the sequence of lns 51 - 52, 82
- Which leads to another point, which is that you may want to handle the GetCurrentChild message in the waitingToStart() state and just respond with CurrentChild(None) ?

RayRoestenburg: Good points.

RayRoestenburg:
Tests seem to fail on generated javadoc on ci, I'll have to look into that as well.

henrymai:
Looks good. Just some nits and still needs a test for "support manual reset" for BackoffOptions.onFailure.

RayRoestenburg:
@henrymai I refactored some more, added some tests, and found there were some issues, now converging both to the same style, which I think is easier to read, (and tests pass, which is important). And I also just factored out the common behavior.
There was an annoying behavior in the first message getting black-holed on the BackoffOnRestartSupervisor because of `self ! StartChild` in `preStart`.

RayRoestenburg:
I did some extra cleanup and refactoring, @patriknw I'm assuming adding a private trait does not break binary compatibility?

RayRoestenburg:
I've also changed the docs so genjavadoc doesn't complain.

RayRoestenburg:
I think this should be ready now. I did refactor the code for some internal reuse gains.

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
```diff
@@ -96,8 +104,26 @@ object BackoffSupervisor {
     def getRef: Optional[ActorRef] = Optional.ofNullable(ref.orNull)
   }
 
-  private case object StartChild extends DeadLetterSuppression
-  private case class ResetRestartCount(current: Int) extends DeadLetterSuppression
+  final case object Reset
```
ktoso: Needs the same docs as the java API below.

```
Send this message to the [[BackoffSupervisor]] and it will reset the back-off.
This should be used in conjunction with `withManualReset` in [[BackoffOptions]].
```

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
```diff
@@ -96,8 +104,26 @@ object BackoffSupervisor {
...
+
+  /**
+   * Java API: Send this message to the [[BackoffSupervisor]] and it will reset the back-off.
+   * This should be used in conjunction with `withManualReset` in [[BackoffOptions]].
+   */
+  def reset = Reset
+
+  final case object GetRestartCount
```
ktoso: Needs the same docs as the Java API

```
Send this message to the [[BackoffSupervisor]] and it will reply with
[[BackoffSupervisor.RestartCount]] containing the current restart count.
```

ktoso:
Failure seems like missing imports 

```
[error] /localhome/jenkinsakka/workspace/pr-validator-per-commit-jenkins/akka-docs/rst/scala/code/docs/pattern/BackoffSupervisorDocSpec.scala:91: not found: value OneForOneStrategy
[error]           OneForOneStrategy() {
[error]           ^
[error] /localhome/jenkinsakka/workspace/pr-validator-per-commit-jenkins/akka-docs/rst/scala/code/docs/pattern/BackoffSupervisorDocSpec.scala:92: not found: value SupervisorStrategy
[error]             case _: MyException ⇒ SupervisorStrategy.Restart
[error]                                   ^
[error] /localhome/jenkinsakka/workspace/pr-validator-per-commit-jenkins/akka-docs/rst/scala/code/docs/pattern/BackoffSupervisorDocSpec.scala:93: not found: value SupervisorStrategy
[error]             case _              ⇒ SupervisorStrategy.Escalate
[error]                                   ^
```

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
```diff
@@ -0,0 +1,234 @@
...
+ * val options = BackoffOptions.onFailure(childProps, childName, minBackoff, maxBackoff, randomFactor)
+ *               .withManualReset
+ * context.actorOf(BackoffSupervisor.props(options), name)
+ * }}}
+ * or capture the result of a method in a variable:
+ * {{{
+ * var options = BackoffOptions.onFailure(childProps, childName, minBackoff, maxBackoff, randomFactor)
+ * options = options.withManualReset
+ * context.actorOf(BackoffSupervisor.props(options), name)
+ * }}}
```
ktoso: let's not show the `var` style :) People can figure it out if they need to, it's not something we want to encourage though I think

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
```diff
@@ -0,0 +1,234 @@
...
+ */
+trait BackoffOptions {
+  /**
+   * Returns a new BackoffOptions with automatic back-off reset.
+   * The back-off algorithm is reset if the child does not crash witin the specified `resetBackoff`.
```
ktoso: s/witin/within/

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
RayRoestenburg: Ok.

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
RayRoestenburg: Ok.

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
RayRoestenburg: Ok.

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
```diff
@@ -0,0 +1,234 @@
...
+   * @param resetBackoff The back-off is reset if the child does not crash within this duration.
+   */
+  def withAutoReset(resetBackoff: FiniteDuration): BackoffOptions
+
+  /**
+   * Returns a new BackoffOptions with manual back-off reset. The back-off is only reset
+   * if the child sends a `BackoffSupervisor.Reset` to its parent (the backoff-supervisor actor).
```
ktoso: [[BackoffSupervisor.Reset]] (so it's navigable in scaladoc)

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
RayRoestenburg: Ok.

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
```diff
@@ -0,0 +1,234 @@
...
+  def withDefaultStoppingStrategy = copy(supervisorStrategy = OneForOneStrategy()(SupervisorStrategy.stoppingStrategy.decider))
+
+  def props = {
+    require(minBackoff > Duration.Zero, "minBackoff must be > 0")
+    require(maxBackoff >= minBackoff, "maxBackoff must be >= minBackoff")
+    require(0.0 <= randomFactor && randomFactor <= 1.0, "randomFactor must be between 0.0 and 1.0")
+    backoffReset match {
+      case AutoReset(resetBackoff) ⇒
+        require(minBackoff <= resetBackoff && resetBackoff <= maxBackoff)
+      case _ ⇒
```
ktoso: ok to ignore here, but let's add `// ignore` so that it's obvious that it's intentional

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
```diff
@@ -0,0 +1,234 @@
...
+        Props(new BackoffOnRestartSupervisor(childProps, childName, minBackoff, maxBackoff, backoffReset, randomFactor, supervisorStrategy))
+      case StopImpliesFailure ⇒
+        Props(new BackoffSupervisor(childProps, childName, minBackoff, maxBackoff, backoffReset, randomFactor, supervisorStrategy))
+    }
+  }
+}
+
+private sealed trait BackoffType
+private final case object StopImpliesFailure extends BackoffType
+private final case object RestartImpliesFailure extends BackoffType
```
ktoso: like those names :+1:

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
```diff
@@ -78,6 +80,12 @@ object BackoffSupervisor {
   }
 
   /**
+   * Props for creating a [[BackoffSupervisor]] actor from [[BackoffOptions]].
+   * @param options the [[BackoffOptions]] that specify how to construct a backoff-supervisor.
+   */
+  def props(options: BackoffOptions): Props = options.props
```
ktoso: :+1:

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
RayRoestenburg: `[[` gave me trouble with genjavadoc, does a single `[` work for scaladoc?

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
```diff
@@ -121,59 +147,45 @@ object BackoffSupervisor {
...
-  extends Actor {
+  strategy: SupervisorStrategy)
+  extends Actor with HandleBackoff {
 
   import BackoffSupervisor._
   import context.dispatcher
 
-  private var child: Option[ActorRef] = None
-  private var restartCount = 0
+  // to keep binary compatibility with 2.4.1
```
ktoso: good, thanks

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
ktoso: hah, just noticed that `[[]] ()` above was rendered as markdown by github in above comment ;-)

Hm, I see `[[]]` may be troublesome because inner object; Ignore this comment then, not a big issue anyway

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
```diff
@@ -184,29 +196,57 @@ final class BackoffSupervisor(
...
+    }
 
+  def handleBackoff: Receive = {
     case StartChild ⇒
       startChild()
-      context.system.scheduler.scheduleOnce(minBackoff, self, ResetRestartCount(restartCount))
+      reset match {
+        case AutoReset(resetBackoff) ⇒
+          val _ = context.system.scheduler.scheduleOnce(resetBackoff, self, ResetRestartCount(restartCount))
+        case _ ⇒
```
ktoso: add an `// ignore` so we know at a glance that it's not accidental omission

RayRoestenburg:
Hm, crap. what is the task to run locally to check this BTW?

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
```diff
@@ -184,29 +196,57 @@ final class BackoffSupervisor(
...
+      }
+
+    case Reset ⇒
+      reset match {
+        case ManualReset ⇒ restartCount = 0
+        case msg         ⇒ unhandled(msg)
```
ktoso: good (was thinking if logging warnings would be better, but unhandled is OK)

akka-contrib/src/main/scala/akka/contrib/pattern/TransparentExponentialBackoffSupervisor.scala
ktoso: Good tip indeed, thanks @henrymai !

akka-docs/rst/general/supervision.rst
```diff
@@ -203,25 +203,26 @@ terminates a child by way of the ``system.stop(child)`` method or sending a
...
 and we need to give it some time to start-up again. One of the prime examples when this is useful is
-when a :ref:`PersistentActor <persistence-scala>` fails with an persistence failure - which indicates that
+when a :ref:`PersistentActor <persistence-scala>` fails (by stopping) with a persistence failure - which indicates that
 the database may be down or overloaded, in such situations it makes most sense to give it a little bit of time
-to recover before the peristent actor is restarted.
+to recover before the peristent actor is started.
 
-The following Scala snippet shows how to create a backoff supervisor which will start the given echo actor
-in increasing intervals of 3, 6, 12, 24 and finally 30 seconds:
+.. [#] A failure can be indicated in two different ways; by an actor terminating or crashing.
```
ktoso: slippery wording here - technically when it crashes and stops "it terminates". Instead of terminating say "stops" here?

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
RayRoestenburg: Ok.

akka-actor/src/main/scala/akka/pattern/BackoffSupervisor.scala
RayRoestenburg: Thanks

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
RayRoestenburg: Ok.

akka-actor/src/main/scala/akka/pattern/BackoffOptions.scala
RayRoestenburg: alright.

akka-docs/rst/general/supervision.rst
RayRoestenburg: Ok.

ktoso:
> Hm, crap. what is the task to run locally to check this BTW?

Either of:

```
sbt shell
> test
# or
> project akka-docs
> test
```

RayRoestenburg:
@ktoso Aha thanks.

akka-actor-tests/src/test/scala/akka/pattern/BackoffSupervisorSpec.scala
```diff
@@ -47,27 +64,92 @@ class BackoffSupervisorSpec extends AkkaSpec with ImplicitSender {
...
+          expectMsg(BackoffSupervisor.RestartCount(0))
+        }
+
+        val stoppingStrategy = OneForOneStrategy() {
+          case _: TestException ⇒ SupervisorStrategy.Stop
+        }
+        val restartingStrategy = OneForOneStrategy() {
+          case _: TestException ⇒ SupervisorStrategy.Restart
+        }
+        assertManualReset(create(onStopOptions(ManualChild.props(testActor)).withManualReset.withSupervisorStrategy(stoppingStrategy)))
```
ktoso: break into multiple lines

akka-actor-tests/src/test/scala/akka/pattern/BackoffSupervisorSpec.scala
```diff
@@ -47,27 +64,92 @@ class BackoffSupervisorSpec extends AkkaSpec with ImplicitSender {
...
+        val supervisor = create(onStopOptions().withDefaultStoppingStrategy)
+        supervisor ! BackoffSupervisor.GetCurrentChild
+        val c1 = expectMsgType[BackoffSupervisor.CurrentChild].ref.get
+        watch(c1)
+        c1 ! "boom"
+        expectTerminated(c1)
+        awaitAssert {
+          supervisor ! BackoffSupervisor.GetCurrentChild
+          // new instance
+          expectMsgType[BackoffSupervisor.CurrentChild].ref.get should !==(c1)
```
ktoso: maybe also check that the restarts count was incremented by sending `GetRestartCount`?

ktoso:
Looks awesome, thanks for the contrib @RayRoestenburg!
I think I ran out of nitpicks, and the core looks good.

Let's get it to compile and pass tests then :)

RayRoestenburg:
@ktoso Ok will do. Hm, I forgot Java does not like companion objects, so I renamed the `BackoffOptions` object to `Backoff`, so you get `Backoff.onStop` and `Backoff.onFailure` (which is not bad).

akka-actor-tests/src/test/scala/akka/pattern/BackoffSupervisorSpec.scala
RayRoestenburg: Ok.

akka-actor-tests/src/test/scala/akka/pattern/BackoffSupervisorSpec.scala
RayRoestenburg: Ok.

ktoso:
It should work just fine with a companion as long as it's a real one – as in `class + object`, `trait + object` indeed causes trouble AFAIR

RayRoestenburg:
Yeah that's what I meant, it was `trait + object`, so not a real one..

ktoso:
Ok, I'll give the current names a look then :)

ktoso:
`Backoff.on...` reads nice in any case :) LGTM from my side.

RayRoestenburg:
Yeah accidentally a better name :smile:. I'm really bad at naming most of the time. Cool. now lets see if it finally passes all the docs etc..

RayRoestenburg:
@ktoso ehh, help?

> akka-actor: Binary compatibility check failed!

Where can I get more information?

ktoso:
You can run `mimaReportBinaryIssues` to see the problem. If it's just internals we can add a filter, if it's actual API we need to work around it such that it's not breaking binary compat.

RayRoestenburg:
@ktoso thanks I'll have a look.

ktoso:
Failures are:

```
[error]  * synthetic method akka$pattern$BackoffSupervisor$$child_=(scala.Option)Unit in class akka.pattern.BackoffSupervisor does not have a correspondent in new version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("akka.pattern.BackoffSupervisor.akka$pattern$BackoffSupervisor$$child_=")
[error]  * synthetic method akka$pattern$BackoffSupervisor$$restartCount()Int in class akka.pattern.BackoffSupervisor does not have a correspondent in new version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("akka.pattern.BackoffSupervisor.akka$pattern$BackoffSupervisor$$restartCount")
[error]  * synthetic method akka$pattern$BackoffSupervisor$$restartCount_=(Int)Unit in class akka.pattern.BackoffSupervisor does not have a correspondent in new version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("akka.pattern.BackoffSupervisor.akka$pattern$BackoffSupervisor$$restartCount_=")
[error]  * synthetic method akka$pattern$BackoffSupervisor$$child()scala.Option in class akka.pattern.BackoffSupervisor does not have a correspondent in new version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("akka.pattern.BackoffSupervisor.akka$pattern$BackoffSupervisor$$child")
[error]  * class akka.pattern.BackoffSupervisor#ResetRestartCount is declared final in new version
[error]    filter with: ProblemFilters.exclude[FinalClassProblem]("akka.pattern.BackoffSupervisor$ResetRestartCount")
```

I'm looking into it (which ones we can ignore and which not)

RayRoestenburg:
@ktoso I'm also looking into it, I can declare the `ResetRestartCount` as not final (easy), the others are a result of refactoring into `HandleBackoff`, the `child` and `restartCount` were public vars in the previous version (since it's an `Actor` no one can touch them so that makes sense why it was that way) so maybe those can be filtered out? Otherwise I'll have to leave them in as dummies and change the names internally, which is not that nice.

RayRoestenburg:
Hm, doesn't seem to be as easy as that, even if I add those vars back in. @ktoso let me now if you can filter this out.

ktoso:
The class needs to be un-final-ized yeah. 
The rest, in fact, we can ignore all these since they are about BackoffSupervisor which is an `Actor` and is `final` anyway.

Please add the filters as the above error message lists them to `project/MiMa.scala`, like so:

```
ProblemFilters.exclude[MissingMethodProblem]("akka.remote.InvalidAssociation.this"),

        // #19281 BackoffSupervisor updates
        ProblemFilters.exclude[FinalClassProblem]("akka.pattern.BackoffSupervisor$ResetRestartCount")
          ...
```

Thanks!

RayRoestenburg:
@ktoso done. I've made the `ResetRestartCount` non-final I think that is what you meant right? All others are ignored using the `ProblemFilters`.

ktoso:
Yup, thanks!

ktoso:
Patrik reviewed and liked it and LGTM from my side as well so merging.
Really excellent improvement, thanks Ray!

RayRoestenburg:
@ktoso you're welcome!

Issue #19274, opened 2015-12-23
Akka Cluster can get stuck joining re-started node when


ktoso:
I debugged this from some logs we gathered with @patriknw.

We believe this is some extremely rare condition in which (maybe, we're not sure yet) a node which is removed from the cluster and restarts, begins JOINING, gets accepted, the `InitJoinAck` is sent but lands in `deadLetters` and then the node wants to join again... and the cluster lets it, wanting to replace the old node with the same address with this one and somehow gets stuck in this state:

The problem looks like this (IPs and details redacted):
```
2015-12-20 03:40:27,403 INFO  [Cluster(akka://ExampleService)] [sourceThread=ExampleService-dispatchers.cluster-14, akkaSource=Cluster(akka://ExampleService), akkaTimestamp=03:40:27.403UTC] - Cluster Node [akka.tcp://ExampleService@2.2.2.2:2552] - Node [akka.tcp://ExampleService@3.3.3.3:2552] is JOINING, roles [a, b, c]
2015-12-20 03:41:21,981 WARN  [a.remote.ReliableDeliverySupervisor] [sourceThread=ExampleService-akka.remote.default-remote-dispatcher-5, akkaTimestamp=03:41:21.978UTC, akkaSource=akka.tcp://ExampleService@2.2.2.2:2552/system/endpointManager/reliableEndpointWriter-akka.tcp%3A%2F%2FExampleService%403.3.3.3%3A2552-16] - Association with remote system [akka.tcp://ExampleService@3.3.3.3:2552] has failed, address is now gated for [5000] ms. Reason: [Disassociated]
2015-12-20 03:41:21,982 WARN  [a.remote.ReliableDeliverySupervisor] [sourceThread=ExampleService-akka.remote.default-remote-dispatcher-5, akkaTimestamp=03:41:21.982UTC, akkaSource=akka.tcp://ExampleService@2.2.2.2:2552/system/endpointManager/reliableEndpointWriter-akka.tcp%3A%2F%2FExampleService%403.3.3.3%3A2552-17] - Association with remote system [akka.tcp://ExampleService@3.3.3.3:2552] has failed, address is now gated for [5000] ms. Reason: [Disassociated]
[...]

PR #19295, opened 2015-12-26
=clu #19274 failure detection of joining/down member status

patriknw:
- Failure detection heartbeating was not performed to joining
  nodes, since it was expected that they will become Up first.
- If a joining node is downed before it is changed to Up failure
  detection will not be performed for that node. That resulted in
  the downed node will not be removed from membership, since the
  unreachability signal is used as confirmation that the node is
  actually stopped before removing it.


patriknw:
Refs #19274

ktoso:
LGTM, very good test too, that's exactly the case we had. Thanks a lot @patriknw!

johanandren:
LGTM!

Issue #18890, opened 2015-11-09
Polish Source.queue


ktoso:
It's not quite complete yet, open questions:

- how to signal completion (there's only offer, we need some form of complete as well)
- how to signal error (there's only offer, we need some form of complete as well)
- offer returns a Future[Boolean] but it can return spurious true because early termination
-  it is very unclear how the user is supposed to encode their sender, should they do some async-tailcall-based thing?
- creating the SourceQueue should not take a timeout, and especially not one with an arbitrary timeout value (5 seconds in the current source tree), if anything the entity which calls `offer` needs to specify any timeouts. It can also be argued that this functionality does not need to be embedded within the SourceQueue at all.

viktorklang:
Also, there is a bit of possible overlap with `Source.maybe`

viktorklang:
Updated the Issue description to also consider timeouts

agolubev:
starting this if no objections.
This should be migrated to graph stage.

hepin1989:
+1,we are migrating to GraphStage anyway.

drewhk:
@agolubev go ahead!

PR #19296, opened 2015-12-26
Agolubev #18890 polish source.queue

agolubev:
Ref #18890
Have one concern about call async callback when stage is about to close. Is there a mechanism that does not allow to schedule such call when onComplete received?
I mean potentially getAsyncCallback can be called and scheduled right before stage is completed, so scheduled call will never happen


agolubev:
Also let me know if StreamCallbackStatus types are good enough so I'll migrate Sink.queue to them as well. They are more descriptive I believe.

ktoso:
OK TO TEST

agolubev:
Build failed with  TlsSpec test.

agolubev:
PR is ready for review

agolubev:
build failed with InputStreamSinkSpec
```
[info] - must fill up buffer by default *** FAILED *** (38 milliseconds)
[info]   Array(77, 42, -94) did not equal ByteString(-82, 117, -102) (InputStreamSinkSpec.scala:149)
[info]   at akka.stream.io.InputStreamSinkSpec$$anonfun$1$$anonfun$apply$mcV$sp$11$$anonfun$apply$mcV$sp$12.apply$mcV$sp(InputStreamSinkSpec.scala:149)
[info]   at akka.stream.io.InputStreamSinkSpec$$anonfun$1$$anonfun$apply$mcV$sp$11$$anonfun$apply$mcV$sp$12.apply(InputStreamSinkSpec.scala:138)
[info]   at akka.stream.io.InputStreamSinkSpec$$anonfun$1$$anonfun$apply$mcV$sp$11$$anonfun$apply$mcV$sp$12.apply(InputStreamSinkSpec.scala:138)
[info]   at akka.stream.testkit.Utils$.assertAllStagesStopped(Utils.scala:26)
[info]   at akka.stream.io.InputStreamSinkSpec$$anonfun$1$$anonfun$apply$mcV$sp$11.apply$mcV$sp(InputStreamSinkSpec.scala:138)
[info]   at akka.stream.io.InputStreamSinkSpec$$anonfun$1$$anonfun$apply$mcV$sp$11.apply(InputStreamSinkSpec.scala:138)
```

johanandren:
PLS BUILD

akka-stream/src/main/scala/akka/stream/Queue.scala
```diff
@@ -35,3 +43,29 @@ trait SinkQueue[T] {
...
+  /**
+   * Type is used to indicate that stream is failed before or during call to the stream
+   * @param cause - exception that stream failed with
+   */
+  case class Failure[T](cause: Throwable) extends StreamCallbackStatus[T]
+
+  /**
+   * Type is used to indicate that stream is completed before call
+   */
+  case class StreamCompleted[T]() extends StreamCallbackStatus[T]
```
johanandren: Can be an object using `Nothing` (just like for example `None`)

akka-stream/src/main/scala/akka/stream/Queue.scala
```diff
@@ -35,3 +43,29 @@ trait SinkQueue[T] {
...
+/**
+ * Contains types that is used as return types for async callbacks to streams
+ */
+object StreamCallbackStatus {
+
+  /**
+   * This class/message type is used to indicate success async call to stream.
+   * @param result - result of async calls to stream
+   */
+  case class Success[T](result: T) extends StreamCallbackStatus[T]
```
johanandren: Final on these?

akka-stream/src/main/scala/akka/stream/Queue.scala
```diff
@@ -35,3 +43,29 @@ trait SinkQueue[T] {
...
+
+  /**
+   * Type is used to indicate that stream is failed before or during call to the stream
+   * @param cause - exception that stream failed with
+   */
+  case class Failure[T](cause: Throwable) extends StreamCallbackStatus[T]
```
johanandren: Could skip the type parameter by using `Nothing`

akka-stream/src/main/scala/akka/stream/Queue.scala
```diff
@@ -35,3 +43,29 @@ trait SinkQueue[T] {
    */
   def pull(): Future[Option[T]]
 }
+
+sealed trait StreamCallbackStatus[T]
```
johanandren: If we only use it with a boolean, do we really need a type parameter?

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -0,0 +1,132 @@
...
+
+      private def enqueueAndSuccess(elem: T, promise: Offered): Unit = {
+        buffer.enqueue(elem)
+        promise.success(StreamCallbackStatus.Success(true))
+      }
+
+      private def receiveElem(elem: T, promise: Offered): Unit = {
+        if (!buffer.isFull) {
+          enqueueAndSuccess(elem, promise)
+        } else (overflowStrategy: @unchecked) match {
```
johanandren: Why `@unchecked` here are there strategies not covered?

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -0,0 +1,132 @@
...
+            }
+          } else pulled = true
+        }
+      })
+    }
+
+    (stageLogic, new SourceQueue[T] {
+      override def watchCompletion() = stageLogic.completion.future
+      override def offer(element: T): Future[StreamCallbackStatus[Boolean]] = {
+        val p = Promise[StreamCallbackStatus[Boolean]]
```
johanandren: `()` ?

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -1338,3 +1338,36 @@ abstract class AbstractOutHandler extends OutHandler
  * (completing when upstream completes, failing when upstream fails, completing when downstream cancels).
  */
 abstract class AbstractInOutHandler extends InHandler with OutHandler
+
+private[akka] trait CallbackWrapper[T] {
```
johanandren: Is this really general enough to not live together with the `QueueSource` but with the generic handler code? If so it should not contain an error naming `SourceQueue.offer` below.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -1338,3 +1338,36 @@ abstract class AbstractOutHandler extends OutHandler
...
+  trait CallbackState
```
johanandren: sealed and then final concrete types

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
johanandren: In addition I think it needs scaladoc explaining purpose and how it works if it should be re-usable.

agolubev: > If so it should not contain an error naming SourceQueue.offer below. 

Ah, you are right

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
agolubev: I'll make them private as thy not supposed to be used outside of given trait

akka-stream/src/main/scala/akka/stream/Queue.scala
agolubev: Yeah, will add

akka-stream/src/main/scala/akka/stream/Queue.scala
agolubev: I'm going to add this to Sink.queue. It's currently Future[Option[Message]] which is not descriptive enough I beleive

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
agolubev: reworked OverflowStatgy a little, it's not perfect, still unchecked problem is solved

agolubev:
Ready for review

agolubev:
Ouch, did something wrong

agolubev:
Let me know if need to rallback and rebase.

akka-stream/src/main/scala/akka/stream/OverflowStrategy.scala
```diff
@@ -3,48 +3,55 @@
...
    */
-  private[akka] case object Backpressure extends OverflowStrategy with DelayOverflowStrategy
-
+  private[akka] case object Backpressure extends OverflowStrategy
   /**
    * INTERNAL API
    */
-  private[akka] case object Fail extends OverflowStrategy with DelayOverflowStrategy {
+  private[akka] case object Fail extends OverflowStrategy {
     final case class BufferOverflowException(msg: String) extends RuntimeException(msg)
```
rkuhn: This type will need to be presented at the top-level, not hidden in `private[akka]` context, because users will want to match on it.

akka-stream/src/main/scala/akka/stream/Queue.scala
```diff
@@ -12,13 +12,21 @@ trait SourceQueue[T] {
 
   /**
    * Method offers next element to a stream and returns future that:
-   * - competes with true if element is consumed by a stream
-   * - competes with false when stream dropped offered element
-   * - fails if stream is completed or cancelled.
+   * - competes with `Success(true)` if element is consumed by a stream
```
rkuhn: `s/competes/completes/` (also below)

akka-stream/src/main/scala/akka/stream/Queue.scala
```diff
@@ -12,13 +12,21 @@ trait SourceQueue[T] {
...
+   * - competes with `Success(false)` when stream dropped offered element
+   * - competes with `StreamCompleted` when stream is completed during future is active
+   * - competes with `Failure(f)` when stream is failed
```
rkuhn: this should not be possible, only cancelation is signaled upstream

akka-stream/src/main/scala/akka/stream/Queue.scala
```diff
@@ -35,3 +43,29 @@ trait SinkQueue[T] {
    */
   def pull(): Future[Option[T]]
 }
+
+sealed trait StreamCallbackStatus[+T]
```
rkuhn: please make it an `abstract class` instead of a `trait` because this package is to be consumed by Scala and Java alike

akka-stream/src/main/scala/akka/stream/impl/ActorRefSourceActor.scala
```diff
@@ -58,7 +59,7 @@ private[akka] class ActorRefSourceActor(bufferSize: Int, overflowStrategy: Overf
         log.debug("Dropping element because there is no downstream demand: [{}]", elem)
       else if (!buffer.isFull)
         buffer.enqueue(elem)
-      else (overflowStrategy: @unchecked) match {
+      else overflowStrategy match {
```
rkuhn: good!

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -1342,3 +1342,47 @@ abstract class AbstractOutHandler extends OutHandler
...
+
+  def stopCallback(f: T ⇒ Unit): Unit = {
+    callbackState.set(Stopped(f))
+  }
+
+  def initCallback(f: T ⇒ Unit): Unit = {
+    val list = (callbackState.getAndSet(Initialized(f)): @unchecked) match {
+      case NotInitialized(l) ⇒ l
+    }
+    list.reverse.foreach(f)
```
rkuhn: I’m afraid to say it, but this won’t work without a lock: while traversing this list the client could concurrently already add new callbacks that go directly to the GraphStage, bypassing this queue and thereby mixing up the ordering. Adding a lock here and around the Initialized case of `callback` below would fix it, but we hate that solution.

This problem pops up everywhere when switching between two enqueueing solutions while requiring ordering to be maintained, for that reason we also have a lock in RepointableActorCell.

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -387,7 +387,7 @@ object Source {
    */
   def actorRef[T](bufferSize: Int, overflowStrategy: OverflowStrategy): Source[T, ActorRef] = {
     require(bufferSize >= 0, "bufferSize must be greater than or equal to 0")
-    require(overflowStrategy != OverflowStrategy.Backpressure, "Backpressure overflowStrategy not supported")
+    require(overflowStrategy != OverflowStrategies.Backpressure, "Backpressure overflowStrategy not supported")
```
rkuhn: now that we have different strategy classifications we can easily push this runtime check to a type check instaed

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -1342,3 +1342,47 @@ abstract class AbstractOutHandler extends OutHandler
...
+  }
+
+  def callback(arg: T): Unit = {
+    callbackState.get() match {
+      case Initialized(cb) ⇒ cb(arg)
+      case list @ NotInitialized(l) ⇒
+        if (!callbackState.compareAndSet(list, NotInitialized(arg :: l)))
+          callbackState.get() match {
+            case NotInitialized(_) ⇒ throw new IllegalStateException("Concurrent call of CallbackWrapper is detected")
```
rkuhn: This is too limiting for the general case, but it is desirable for QueueSink. I think I’d rather remove it because concurrent use goes unnoticed in almost all cases anyway.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -1342,3 +1342,47 @@ abstract class AbstractOutHandler extends OutHandler
...
+ * This trait wraps callback for `GraphStage` stage instances and handle gracefully cases when stage is
+ * not yet initialized or already finished.
+ *
+ * While `GraphStage` has not initialized it adds all requests to list.
+ * As soon as `GraphStage` is started it stops collecting requests (pointing to real callback
+ * function) and run all the callbacks from the list
+ *
+ * Supposed to be used by GraphStages that share call back to outer world
+ */
+private[akka] trait CallbackWrapper[T] {
```
rkuhn: Why not make it extend `Callback[T]` and rename `callback` to `invoke`? I do agree with the `private[akka]` scope at this point, and it should also get an `INTERNAL API` header in the comment.

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  type Offered = Promise[StreamCallbackStatus[Boolean]]
+
+  val out = Outlet[T]("queueSource.out")
+  override val shape: SourceShape[T] = SourceShape.of(out)
+
+  override def createLogicAndMaterializedValue(inheritedAttributes: Attributes) = {
+    val buffer = if (maxBuffer == 0) null else FixedSizeBuffer[T](maxBuffer)
+    var pendingOffer: Option[(T, Offered)] = None
+    var pulled = false
```
rkuhn: these should go into the logic, that will save several object allocations (for variable holders)

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -0,0 +1,131 @@
...
+
+    val stageLogic = new GraphStageLogic(shape) with CallbackWrapper[(T, Offered)] with SourceCompletion {
+      def bufferOverflowException = new Fail.BufferOverflowException(s"Buffer overflow (max capacity was: $maxBuffer)!")
+
+      override def preStart(): Unit = initCallback(callback.invoke)
+      override def postStop(): Unit = stopCallback {
+        case (elem, promise) ⇒ promise.failure(new IllegalStateException("Stream is terminated. SourceQueue is detached"))
+      }
```
rkuhn: Isn’t something like this missing for SinkQueue as well?

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -0,0 +1,131 @@
...
+import akka.stream.stage._
+import scala.concurrent.{ Future, Promise }
+
+/**
+ * INTERNAL API
+ */
+private[akka] class QueueSource[T](maxBuffer: Int, overflowStrategy: OverflowStrategy) extends GraphStageWithMaterializedValue[SourceShape[T], SourceQueue[T]] {
+  trait SourceCompletion {
+    val completion = Promise[Unit]
+  }
```
rkuhn: This should be just a member in the logic instead of a trait.

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -0,0 +1,131 @@
...
+              case Some(_) ⇒
+                promise.failure(new IllegalStateException("You have to wait for previous offer to be resolved to send another request"))
+              case None ⇒
+                pendingOffer = Some((elem, promise))
+            }
+        }
+      }
+
+      private val callback: AsyncCallback[(T, Offered)] = getAsyncCallback(tuple ⇒ {
+        val (elem, promise) = tuple
```
rkuhn: This is more efficiently formulated as
~~~scala
getAsyncCallback {
  case (elem, promise) =>
    ...
}
~~~

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -0,0 +1,131 @@
...
+          receiveElem(elem, promise)
+          if (pulled) {
+            push(out, buffer.dequeue())
+            pulled = false
+          }
+        } else if (pulled) {
+          push(out, elem)
+          pulled = false
+          promise.success(StreamCallbackStatus.Success(true))
+        } else pendingOffer = Some((elem, promise))
```
rkuhn: ah, forget the above if you just use `tuple` here (to avoid the extra allocation)

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -0,0 +1,131 @@
...
+      override def postStop(): Unit = stopCallback {
+        case (elem, promise) ⇒ promise.failure(new IllegalStateException("Stream is terminated. SourceQueue is detached"))
+      }
+
+      private def enqueueAndSuccess(elem: T, promise: Offered): Unit = {
+        buffer.enqueue(elem)
+        promise.success(StreamCallbackStatus.Success(true))
+      }
+
+      private def receiveElem(elem: T, promise: Offered): Unit = {
```
rkuhn: please rename to `bufferElem` to make it clear that this is only for buffering cases

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -0,0 +1,131 @@
...
+            enqueueAndSuccess(elem, promise)
+          case DropBuffer ⇒
+            buffer.clear()
+            enqueueAndSuccess(elem, promise)
+          case DropNew ⇒
+            promise.success(StreamCallbackStatus.Success(false))
+          case Fail ⇒
+            promise.success(StreamCallbackStatus.Failure(bufferOverflowException))
+            completion.failure(bufferOverflowException)
+            failStage(bufferOverflowException)
```
rkuhn: please reuse the `bufferOverflowException` object, it is costly to create

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -0,0 +1,131 @@
...
+          case DropTail ⇒
+            buffer.dropTail()
+            enqueueAndSuccess(elem, promise)
+          case DropBuffer ⇒
+            buffer.clear()
+            enqueueAndSuccess(elem, promise)
+          case DropNew ⇒
+            promise.success(StreamCallbackStatus.Success(false))
+          case Fail ⇒
+            promise.success(StreamCallbackStatus.Failure(bufferOverflowException))
```
rkuhn: I see now that this makes sense, but the documentation should be clarified that only failure to enqueue is signaled like that—it currently speaks about “stream failure” in general, which is not true.

rkuhn:
Thanks for pushing this! Please note that we’ll switch development to the `master` branch very soon, so you might need to then rebase this PR once finished (unless you finish it within the next two hours :-) ).

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
agolubev: I think so.

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
agolubev: Yeah, thanks

agolubev:
>so you might need to then rebase this PR once finished (unless you finish it within the next two hours :-) ).

That's good news. Will rebase it when it's ready

rkuhn:

akka-stream/src/main/scala/akka/stream/OverflowStrategy.scala
agolubev: actually it was intentionally hidden before my this PR and I tried to preserve this approach. 
Strategies case objects are hidden but OverflowStrategy.<strategyName> are visible.
That works in case we are using stream from external code.
Still if developer wants to create own Stage that handle strategies - this will need to be public.
Let me know if we are going this way.

rkuhn: For the exception type this reasoning does not apply: the exception will possibly fail the stream, so the user must be able to access this type to pattern match on it.

akka-stream/src/main/scala/akka/stream/Queue.scala
agolubev: changed this to "when host stage is failed". queue can fail when OverflowStrategy is Fail

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
agolubev: Build is showing warning as promise must be visible to external world (in SourceQueue instance)
```
[warn] /Users/alexandergolubev/Projects/akka/akka-stream/src/main/scala/akka/stream/impl/Sources.scala:121: reflective access of structural type member value completion should be enabled
[warn] by making the implicit value scala.language.reflectiveCalls visible.
[warn] This can be achieved by adding the import clause 'import scala.language.reflectiveCalls'
[warn] or by setting the compiler option -language:reflectiveCalls.
[warn] See the Scala docs for value scala.language.reflectiveCalls for a discussion
[warn] why the feature should be explicitly enabled.
[warn]       override def watchCompletion() = stageLogic.completion.future
```

akka-stream/src/main/scala/akka/stream/OverflowStrategy.scala
agolubev: thank you for clarification will fix this

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
agolubev: using DelayOverflowStrategy will add EmitEarly to strategy list which is not applicable to ActorRefSource I believe.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
agolubev: thanks for the reference, will redo this.

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
rkuhn: Yes, the solution is to turn the anonymous logic into a local class.

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
rkuhn: I meant that you can easily introduce another base type to mark out only the allowed values.

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
agolubev: got it.

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
agolubev: Ah, this is fairly small - will add this to current PR

PR #19297, opened 2015-12-27
fix error in distributed pub sub documentation

bdrupieski:
Both cluster-singleton.rst and distributed-pub-sub.rst say "To use the Cluster Singleton you must add the following dependency in your project." in the Dependencies header. This is confusing for distributed-pub-sub.rst because it does not refer to the Cluster Singleton. I changed the reference of "the Cluster Singleton" to "Distributed Publish Subscribe".


ktoso:
LGTM, thanks!

PR #19298, opened 2015-12-27
=clu #19274 failure detection of joining/down member status (for validation)

patriknw:
- Failure detection heartbeating was not performed to joining
  nodes, since it was expected that they will become Up first.
- If a joining node is downed before it is changed to Up failure
  detection will not be performed for that node. That resulted in
  the downed node will not be removed from membership, since the
  unreachability signal is used as confirmation that the node is
  actually stopped before removing it.

(cherry picked from commit 4d64901228d3c78558cde4f48a1df20d73a4e5d6)


ktoso:
LGTM, thanks @patriknw !

ktoso:
Resolves #19274

Issue #19299, opened 2015-12-28
Performance of Flow.flatMapMerge


hbf:
I've observed that `Flow.flatMapMerge` is slower than creating the involved streams explicitly and using a `Merge`.

I cooked this down to this:

```java
public static void main(String[] argv) throws Exception {
    final ActorSystem system = ActorSystem.create("testing");
    final Materializer mat = ActorMaterializer.create(system);

    long before = System.currentTimeMillis();
    final Source<Integer, BoxedUnit> source = true? // (*)
            Source.single(BoxedUnit.UNIT).flatMapMerge(1, __ -> createSource()) :
            createSource();

    Await.result(Source.fromGraph(source).toMat(Sink.ignore(), Keep.right()).run(mat), Inf());
    System.out.println("Took " + (System.currentTimeMillis()-before)/1000.0f);

    system.shutdown();
}

private static Source<Integer, BoxedUnit> createSource() {
    return Source.repeat(1).take(30000000);
}
```

Running time:

 * 21.2s (code above)
 * 2.9s (code above with `true` changed to `false` in line `*`)

Is this slow down expected (maybe because fusion cannot take place)? Is a workaround know? (In my use case, the stream's are not known at compile time.)

ktoso:
Thanks for reporting @hbf!
We've not yet benchmarked that one, would you be able to help us out to get started on optimisations there by implementing a benchmark (using JMH), in akka-bench-jmh-dev? Here's an example benchmark: https://github.com/akka/akka/blob/release-2.3-dev/akka-bench-jmh-dev/src/main/scala/akka/stream/GraphBuilderBenchmark.scala

// here's a read-me how to run those: https://github.com/akka/akka/tree/release-2.3-dev/akka-bench-jmh
// it's using: https://github.com/ktoso/sbt-jmh

Thanks for the help in advance!

hbf:
Sure. Have a a look here and let me know what you think: https://github.com/akka/akka/pull/19308

hbf:
@ktoso Would you have a guess what needs to be done to make this faster? I'm not acquainted with Akka internals but maybe I'd be happy to spend some time trying to improve this. (Note: a quick look showed that the `flatMapMerge`-version generates about 2-3 times more churn, in particular `AsyncInput` instances.)

ktoso:
I'd have do the same as anyone trying to optimise this, fire up a profiler and track hot-paths :)
This is not my priority this week, so any help would be appreciated :)

[...]

PR #19308, opened 2015-12-29
Benchmark for the flatMapMerge operator

hbf:
Add a JMH benchmark for `Flow.flatMapMerge()`.

The benchmark processes 100k elements that – depending on the benchmark parameter `NumberOfStreams` – come from:
- Case `NumberOfStreams`=0: a single source. In this case, `flatMapMerge()` is _not_ used. This is the base line against which the other tests compare.
- Case `NumberOfStreams`>0: the 100k elements come from `NumberOfStreams`-many sources, each of which produces (100k/`NumberOfStreams`)-many elements. The sources are merged together via `flatMapMerge()`.

Notice that in the latter case, the sources are pre-fused to make the comparison as fair as possible.

On my laptop, `sbt 'akka-bench-jmh-dev/run -i 10 -wi 4 -f1 .*FlatMapMergeBenchmark.*'` gives:

```
[info] Benchmark                                           (NumberOfStreams)   Mode  Cnt     Score     Error   Units
[info] FlatMapMergeBenchmark.flat_map_merge_100k_elements                  0  thrpt   10  8455.581 ± 503.256  ops/ms
[info] FlatMapMergeBenchmark.flat_map_merge_100k_elements                  1  thrpt   10   668.567 ±  18.370  ops/ms
[info] FlatMapMergeBenchmark.flat_map_merge_100k_elements                 10  thrpt   10   641.890 ±  14.682  ops/ms
```

### DIFF
--- a/akka-bench-jmh-dev/src/main/scala/akka/stream/FlatMapMergeBenchmark.scala
+++ b/akka-bench-jmh-dev/src/main/scala/akka/stream/FlatMapMergeBenchmark.scala
@@ -1,0 +1,55 @@
+/**
+  * Copyright (C) 2014 Typesafe Inc. <http://www.typesafe.com>
+  */
+
+package akka.stream
+
+import akka.actor.ActorSystem
+import akka.stream.scaladsl._
+import java.util.concurrent.TimeUnit
+import org.openjdk.jmh.annotations._
+import scala.concurrent._
+import scala.concurrent.duration.Duration.Inf
+
+@State(Scope.Benchmark)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Array(Mode.Throughput))
+class FlatMapMergeBenchmark {
+  implicit val system = ActorSystem("FlatMapMergeBenchmark")
+  val materializerSettings = ActorMaterializerSettings(system).withDispatcher("akka.test.stream-dispatcher")
+  implicit val materializer = ActorMaterializer(materializerSettings)
+
+  val NumberOfElements = 100000
+
+  @Param(Array("0", "1", "10"))
+  val NumberOfStreams = 0
+
+  var graph: RunnableGraph[Future[Unit]] = _
+
+  def createSource(count: Int): Graph[SourceShape[Int], Unit] = akka.stream.Fusing.aggressive(Source.repeat(1).take(count))
+
+  @Setup
+  def setup() {
+    val source = NumberOfStreams match {
+      // Base line: process NumberOfElements-many elements from a single source without using flatMapMerge
+      case 0 => createSource(NumberOfElements)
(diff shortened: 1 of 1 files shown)


ktoso:
OK TO TEST

hbf:
The Jenkins log says

```
FAILED ***
[error] Failed: Total 539, Failed 1, Errors 0, Passed 538, Pending 20
[error] Failed tests:
[error] 	akka.http.impl.engine.ws.WebsocketServerSpec
```

Is that related to this commit?

ktoso:
The failure is unrelated, don't worry

akka-bench-jmh-dev/src/main/scala/akka/stream/FlatMapMergeBenchmark.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  var graph: RunnableGraph[Future[Unit]] = _
+
+  def createSource(count: Int): Graph[SourceShape[Int], Unit] = akka.stream.Fusing.aggressive(Source.repeat(1).take(count))
+
+  @Setup
+  def setup() {
+    val source = NumberOfStreams match {
+      // Base line: process NumberOfElements-many elements from a single source without using flatMapMerge
+      case 0 => createSource(NumberOfElements)
```
ktoso: our baseline should be the "manually constructed equivalent" though, not a plain Source here.

hbf: What do you suggest to do? I agree the word "base line" is not really appropriate here. It's rather a lower bound.

ktoso: It's a useful beginning for looking into the merge's performance in any case, thanks!
Let's merge and if you have the time and would like to dig around with a profiler more insights or fix PR would be certainly welcome :)

We'll spend January on performance with the entire team btw, so this bench might come in handy as a starting point. Thanks!

ktoso:
Oh, one more thing @hbf,
I just noticed that you have signed the 2.1 version of the Typesafe CLA, but the current version is 2.2.
Would you mind confirming the 2.2 version via this site so I can merge this?http://www.typesafe.com/contribute/cla

Thanks!

ktoso:
PLS BUILD

akka-bench-jmh-dev/src/main/scala/akka/stream/FlatMapMergeBenchmark.scala
hbf: Sounds great, thanks @ktoso !

hbf:
I confirmed version 2.2. You should be good to go.

ktoso:
Thanks! Let's get the benchmark, and if you have time to investigate it more that'd be very awesome, thanks in advance!

ktoso:
Refs https://github.com/akka/akka/issues/19299

PR #19313, opened 2015-12-30
+htp #19311 allow handling a route with Future[RouteResult]

### LINKED ISSUE #19311: Future[RouteResult] to Route conversion to ease handling routes with actors
Which allows to be still type-safe yet implement a similar thing like previously one would have done it with spray per-request-actors.

This came up during a discussion with @briantopping on gitter: :point_up: [December 30, 2015 10:24 AM](https://gitter.im/akka/akka?at=5683a2d435e1a316162ea96d)

ktoso:
Resolves #19311

Please have a look @sirthias @jrudolph, does this change make sense to you?
Ticket and gitter discussion explains more, but in short: it's to make handling requests easier with Actors (one example would be people migrating from Spray, where this pattern was popular)

### DIFF
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/server/BasicRouteSpecs.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/server/BasicRouteSpecs.scala
@@ -9,6 +9,8 @@ import model.HttpMethods._
 import model.StatusCodes
 import akka.testkit.EventFilter
 
+import scala.concurrent.Future
+
 class BasicRouteSpecs extends RoutingSpec {
 
   "routes created by the concatenation operator '~'" should {
@@ -189,4 +191,31 @@ class BasicRouteSpecs extends RoutingSpec {
       } ~> check { status shouldEqual StatusCodes.MethodNotAllowed }
     }
   }
+
+  "RouteResult creation by user" should {
+    "allow handling route with Future[RouteResult]" in {
+      val routes = {
+        def askSomeone(i: Int, ctx: RequestContext): Future[RouteResult] =
+          if (i == 42) ctx.complete(s"Done: $i")
+          else ctx.reject(ValidationRejection(s"Nein, $i!"))
+
+        parameter("q".as[Int]) { q ⇒
+          extractRequestContext { ctx ⇒
+            askSomeone(q, ctx)
+          }
+        }
+      }
+
+      Get("/?q=42") ~> routes ~> check {
+        val theResponse = responseAs[String]
+        info("theResponse = " + theResponse)
+      }
+
+      Get("/?q=9") ~> routes ~> check {
+        handled should ===(false)
+        rejections should ===(List(ValidationRejection("Nein, 9!")))
+      }
+    }
+  }
+
 }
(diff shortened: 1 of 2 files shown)


sirthias:
Hmm.
So you want this to work:

```scala
get {
  someRouteResultFuture
}
```

Note that this already does work:

```scala
get { _ =>
  someRouteResultFuture
}
```

If the goal is to properly integrate an actor backend I wouldn't recommend this solution, as it requires the backend to produce a `RouteResult` instance which requires the backend to add a dependency onto the http module, which violates the architectural layer separation we usually try to advocate.

Would the usual actor ask approach work in your case, @briantopping?:

```scala
get {
   complete(actorRef.ask(cmd).mapTo[MyDomainResponse])
}
```

Here you would provide a proper marshaller for `MyDomainResponse` and you should be set.
If you also want the backend to have control over rejections you could do that via custom exceptions and a respective ExceptionHandler.
Or, if you really want to go "all dirty" (i.e. have the http layer leak into your backend):

```scala
get { ctx =>
   actorRef.ask(MyCmd(ctx)).mapTo[RouteResult]
}
```

So, I'm not sure this patch really anything valuable to a larger fraction of our users.

ktoso:
Thanks for the analysis @sirthias! After having stepped back from this PR for a few (holi)days I agree it's not really helping.

What this shows we're lacking docs showing users coming from spray how they should do this (essentially your comment / the migration guide).

briantopping:
Hi @sirthias, welcome back! Happy new year.

What I had done to get me around the issue is to grab the magic out of the PR @ktoso made here and use that inline:

```
  implicit def routeResult2Route(f: Future[RouteResult]): Route =
    new StandardRoute {
      override def apply(ctx: RequestContext): Future[RouteResult] = f
    }

  private def dynRoute() = {
    import akka.http.scaladsl.server.Directives._
    pathPrefix(backends) { ref =>
      extractRequestContext { req =>
        implicit val timeout = Timeout(2.second)
        akka.pattern.ask(ref, req).mapTo[RouteResult]
      }
    }
  }
```

I don't remember all the little details of this now, but I think the problem with using `complete` is the remote route may reject or error and the semantics shouldn't be any different than if the remote were actually hosting the HTTP endpoint. Apologies if your solution describes that and and I'm missing it!

sirthias:
Happy New Year to you too, Brian!
Yes, I figured that enabling control over rejections and errors from your "backend" actor was actually the underlying reason for your request. As long as this actor doesn't actually live in your backend, but rather in some other (intermediate) layer architecturally still belonging to the API this is probably ok.

However, as indicated in my last reply above, we don't really want to make it easy to have the API layer leak into the backend by exposing things like rejections to the backend. Rejections are a concept that is part of the API layer. In the long run users should be better of by keeping things architecturally clean and uncoupled as much as possible.

That's why I'd vote for keeping things as they are and simply pointing advanced users who know what they are doing to Konrad's solution. However, I think that you shouldn't actually need the added `implicit def routeResult2Route`. If I am not mistaken your `dynRoute` should also work like this:

```scala
import akka.http.scaladsl.server.Directives._
def dynRoute() =
  pathPrefix(backends) { ref => ctx =>
    akka.pattern.ask(ref, ctx).mapTo[RouteResult]
  }
```

briantopping:
Thanks, that's what I had to start with, but I didn't have the implicit conversion until @ktoso was able to sleuth that. The extra cruft was WIP. Cheers!

Issue #19302, opened 2015-12-28
"1.3.10 Using Akka with SBT" refers to a missing "Akka/SBT template" project


kryger:
The documentation reads:

> The simplest way to get started with Akka and SBT is to check out the Akka/SBT template project.

This is linked to https://www.typesafe.com/resources/getting-started/typesafe-stack/downloading-installing.html#template-projects-for-scala-akka-and-play , but this redirects to Typesafe Activator. I found the "Akka/Scala/SBT template" project here: https://github.com/typesafehub/akka-scala-sbt.g8 , but it hasn't been updated in four years so I'm assuming it's outdated and deprecated.

I suspect using the template is no longer the recommended method to get started, but if the Activator is the way to go the docs should be rewritten to explain this clearly.

I would've suggested a PR but I have no idea what would be the correct way to fix it (i.e. the docs text, the link, or both?)

ktoso:
Hi Lukasz,
where specifically is it mentioned in the docs? I could not find that sentence quickly grepping the site or docs.

You're right, it should point to: https://www.typesafe.com/activator/templates#filter:akka instead, thanks!

Any g8 templates are deprecated, we should remove them, thanks for the ping.

kryger:
Ahoy there,
I found it in the Scala PDF version of the docs, but it's also in the HTML version here: http://doc.akka.io/docs/akka/2.4.1/intro/getting-started.html#Using_Akka_with_SBT

ktoso:
You can fix it here: https://github.com/akka/akka/edit/master/akka-docs/rst/intro/getting-started.rst and send a PullRequest with the fix.

Please name the commit like this: `=doc #19302 fix activator link in getting started docs`

ktoso:
Thanks in advance for the help!

ktoso:
@kryger you've submitted a number of fixes recently, but not for this one AFAIR. Are you still interesting in submiting a PR fixing this sentence?

kryger:
I'll try to get round to it tomorrow, sadly my day job takes precedence
over GitHub ;)

ktoso:
Right, of course :-) I thought you may be on a holiday break and hacking open source stuffs :-)
Thanks a lot in advance, marking it as in progress then :)

ktoso:
Thanks!

PR #19314, opened 2015-12-30
=doc #19302 fix activator link in getting started docs

kryger:
Fixed the docs as per discussion in the other issue. Wasn't sure which SBT template specifically is considered "canonical" so linked to the templates index page.


ktoso:
Thanks a lot for noticing and the fix!

ktoso:
Resolves #19302

PR #19317, opened 2015-12-30
make FlowOps.reprFlatten safe and precise

rkuhn:
Finally I figured out how to teach scalac that `FlowOps#Repr` is some kind of “my-type”. This implementation removes some casts and actually requires FlowOps subtypes to adhere to the contract.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
@@ -349,19 +349,13 @@ trait FlowOps[+Out, +Mat] {
   import akka.stream.impl.Stages._
   import GraphDSL.Implicits._
 
-  type Repr[+O] <: FlowOps[O, Mat]
+  type Repr[+O] <: FlowOps[O, Mat] {
+    type Repr[+OO] = FlowOps.this.Repr[OO]
+    type Closed = FlowOps.this.Closed
+  }
 
   // result of closing a Source is RunnableGraph, closing a Flow is Sink
   type Closed
-
-  /*
-   * Repr is actually self-bounded, but that would be a cyclic type declaration that is illegal in Scala.
-   * Therefore we need to help the compiler by specifying that Repr expressions can be flattened.
-   */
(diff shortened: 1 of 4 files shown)


akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -349,20 +349,14 @@ trait FlowOps[+Out, +Mat] {
...
+  type Repr[+O] <: FlowOps[O, Mat] {
+    type Repr[+O] = FlowOps.this.Repr[O]
+    type Closed = FlowOps.this.Closed
+  }
 
   // result of closing a Source is RunnableGraph, closing a Flow is Sink
   type Closed
 
-  /*
-   * Repr is actually self-bounded, but that would be a cyclic type declaration that is illegal in Scala.
```
rkuhn: This part was wrong: it is not self-bounded, it only needs to produce a type that has the same Repr type constructor. With the fix above scalac will actually verify that implementations are correct.

rkuhn:
Oh, and this fix also removes some method calls (“identity functions”).

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -922,10 +922,9 @@ object GraphDSL extends GraphApply {
 
       def to[Mat2](sink: Graph[SinkShape[Out], Mat2]): Closed =
         super.~>(sink)(b)
-
     }
 
-    private class DisabledPortOps[Out, Mat](msg: String) extends PortOpsImpl[Out](null, null) {
+    private class DisabledPortOps[Out](msg: String) extends PortOpsImpl[Out](null, null) {
```
viktorklang: Nice Boyscouting

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: :+1:

Issue #18565, opened 2015-09-25
add combinator for termination notification


rkuhn:
For context see [this discussion](https://groups.google.com/d/msg/akka-user/l8ACEw6ndsA/ZK2UCmqNBQAJ).

[...] Perhaps we should offer a combinator whose only purpose is to complete a Future upon termination, but otherwise it just sends all elements downstream, like

    def completionProbe[M](combine: [Mat, Future[Unit]] => M): Repr[Out, M]

Then this problem could be solved independent of the chosen Source:

    val (completion, sinkFuture) = source.completionProbe(Keep.right).toMat(sink)(Keep.both).run()

For this combinator (and possibly others of its kind) I would not want to provide a default scheme for treating the materialized value since adding such a value is its only purpose.

drewhk:
Isn't the issue with this is that the completion of this stage does not translate to the completion of the downstream?

viktorklang:
@drewhk Great point.

WARNING: PREMATURE IDEA BELOW

What if Sinks are "simply" result-projectors instead of doing "actual" work.

In that case the "Writing of a File" would be a Flow[ByteString, (Try[Unit], Long), Unit] which gives:

~~~scala
Source(file) via WriteBytes(file) to Sink.head
~~~

rkuhn:
@drewhk That is not an issue here since the user is just interested in whether the stream has successfully flowed past this point in its entirety, which has well-defined semantics and is probably useful elsewhere.

@viktorklang Your proposal would ruin the necessary write-count value that the sink should emit regardless of whether the stream terminated successfully or not. I do maintain that my proposal has the right structure, but please do keep beating on it :-)

rkuhn:
Ah, and the name needs some love, as you noted on the mailing list.

viktorklang:
@rkuhn since it is a Flow it obeys the same rules as a Processor, in which case it can recover from a failure and signal (Try[Unit], Long) and then complete ;)

rkuhn:
I take the smiley to imply that you acknowledge that providing a generic solution is better than complicating every single Sink.

drewhk:
I think @viktorklang's idea is actually quite nice. Maybe not expressed as a Flow, but the idea of providing some baseline support for sink completion is appealing.

kutchar:
Since we are on this topic, I think there are a few other things that need to be addressed:
1. Upstream failure handling. For example, if a file source fails, then the failure should be propagated downstream.

[...]

PR #19319, opened 2015-12-31
Agolubev #18565 termination notification combinator

agolubev:
Ref #18565
Two commits are here. 
- First is about having `def watchTermination[M](combine: [Mat, Future[Unit]] => M): Repr[Out, M]`
- Second is having watchTermination (with Keep.right) and watchTerminationMat (with combine function). I know  that default is Keep.left but with second approach it's more elegant (and still can us Mat version if we need both Mat values).
- watch termination from previous stage and cancellation from next stage

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
@@ -3,11 +3,15 @@
  */
 package akka.stream.impl.fusing
 
+import java.io.InputStream
+import java.util.concurrent.LinkedBlockingDeque
 import java.util.concurrent.atomic.AtomicBoolean
 import akka.actor.Cancellable
 import akka.dispatch.ExecutionContexts
 import akka.event.Logging
 import akka.stream._
+import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
+import akka.stream.impl.io.InputStreamSinkStage.{ Failed, Data }
 import akka.stream.stage._
 import scala.concurrent.{ Future, Promise }
 import scala.concurrent.duration.FiniteDuration
@@ -69,7 +73,6 @@ object GraphStages {
     override val shape = FlowShape(in, out)
 
     override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
-      var initialized = false
 
       setHandler(in, new InHandler {
         override def onPush(): Unit = {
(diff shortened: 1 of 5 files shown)


viktorklang:
Isn't it important to distinguish that this combinator checks the termination of *previous* stages?

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
```diff
@@ -117,6 +120,51 @@ object GraphStages {
...
 
+  private class TerminationWatcher[T] extends GraphStageWithMaterializedValue[FlowShape[T, T], Future[Unit]] {
+    val name = "TerminationWatcher"
+    val in = Inlet[T](name + ".in")
+    val out = Outlet[T](name + ".out")
+    override def initialAttributes = Attributes.name(name)
+    override val shape = FlowShape(in, out)
+
+    override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, Future[Unit]) = {
+      trait SharingTerminationFuture {
```
viktorklang: Not used?

agolubev: without this there is warning
[warn] /Users/alexandergolubev/Projects/akka/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala:159: reflective access of structural type member value finishFuture should be enabled
[warn] by making the implicit value scala.language.reflectiveCalls visible.
[warn] This can be achieved by adding the import clause 'import scala.language.reflectiveCalls'
[warn] or by setting the compiler option -language:reflectiveCalls.
[warn] See the Scala docs for value scala.language.reflectiveCalls for a discussion
[warn] why the feature should be explicitly enabled.
[warn]       (logic, logic.finishFuture)
[warn]                     ^
[info] /Users/alexandergolubev/Projects/akka/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala:159: method invocation uses reflection
[info]       (logic, logic.finishFuture)
[info]                     ^
[warn] one warning found

viktorklang: Ah, thanks, I expanded the wrong code section.

I'd do this instead:

~~~scala

      val finishPromise = Promise[Unit]
      val logic = new GraphStageLogic(shape) with SharingTerminationFuture {

          override def onPush(): Unit = push(out, grab(in))

          override def onUpstreamFinish(): Unit = {
            finishPromise.success(())
            super.onUpstreamFinish()
          }

          override def onUpstreamFailure(ex: Throwable): Unit = {
            finishPromise.failure(ex)
            super.onUpstreamFailure(ex)
          }
        })
          override def onDownstreamFinish(): Unit = {
            finishPromise.success(())
            super.onDownstreamFinish()
          }
        })
      }
      (logic, finishPromise.future)
~~~

agolubev: yeah, makes sense

ktoso:
OK TO TEST

rkuhn:

agolubev:
>Isn't it important to distinguish that this combinator checks the termination of previous stages?

Mentioned this in scaladoc for watchTermination method

PR #19325, opened 2016-01-01
!peq #19324 update persistence-query to stream 2.0.1

patriknw:
I'm not sure when the right time is to merge this, but here are the needed changes anyway.

### DIFF
--- a/akka-persistence-query/src/test/scala/akka/persistence/query/journal/leveldb/AllPersistenceIdsSpec.scala
+++ b/akka-persistence-query/src/test/scala/akka/persistence/query/journal/leveldb/AllPersistenceIdsSpec.scala
@@ -45,10 +45,12 @@ class AllPersistenceIdsSpec extends AkkaSpec(AllPersistenceIdsSpec.config)
       expectMsg("c1-done")
 
       val src = queries.currentPersistenceIds()
-      src.runWith(TestSink.probe[String])
-        .request(5)
-        .expectNextUnordered("a", "b", "c")
-        .expectComplete()
+      val probe = src.runWith(TestSink.probe[String])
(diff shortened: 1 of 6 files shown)


ktoso:
LGTM, I think as soon as we can. It's experimental so can track the experimental streams.
Changes in january should be rather minimal.

akka-persistence-query/src/test/scala/akka/persistence/query/journal/leveldb/AllPersistenceIdsSpec.scala
```diff
@@ -45,10 +45,12 @@ class AllPersistenceIdsSpec extends AkkaSpec(AllPersistenceIdsSpec.config)
       expectMsg("c1-done")
 
       val src = queries.currentPersistenceIds()
-      src.runWith(TestSink.probe[String])
-        .request(5)
-        .expectNextUnordered("a", "b", "c")
-        .expectComplete()
+      val probe = src.runWith(TestSink.probe[String])
+      probe.within(10.seconds) {
```
patriknw: I created an issue for this: https://github.com/akka/akka/issues/19326

patriknw:
yeah, the only thing I'm worried about is that we get a mix of incompatible persistence plugins out there. We will change one more time when stream is merged into master (2.4.x).

patriknw:
fixed code in docs also

rkuhn:
We won’t release an Akka version before Streams are in, so we can merge this ASAP—it does need a trivial rebase that I can do later today unless you beat me to it.

patriknw:
alright, rebased

patriknw:
Refs #19324

Issue #19323, opened 2015-12-31
Resource resolution in BundleDelegatingClassLoader is non-deterministic


briantopping:
After working briefly with `BundleDelegatingClassLoader`, it became apparent that it's resource resolution is non-deterministic because it uses `Set` instead of `Seq`. Any change to dependencies or key hash will result in a different order of resolution for classes or resources. 

Should fix #18846. PR to follow...

viktorklang:
:+1:

briantopping:
First part of this PR complete as a deterministic breadth-first discovery of transitive dependencies. It's only called on bundle startup so the performance hit of replacing `Set` with `Seq` shouldn't create any issues (and may improve performance once loaded due to the better ordering of the bundle list by distance). This doesn't solve all the issues and I don't think some of them are easily solvable, so I'd like to better document what's going on as a second part of the PR. 

Refresher on OSGi: A Bundle is a Java JAR that has additional MANIFEST.MF entries documenting dependencies and how to interact with those dependencies. For this discussion, we mostly care about the `Import-Package` header. Unfortunately, [bndtools](http://bnd.bndtools.org) (which tools link against for generating manifest entries) sorts the imports alphabetically, not by graph distance. This isn't a huge disadvantage WRT monolithic JVM apps since `ClassLoader` makes no guarantees on search result ordering. In turn, if `BundleDelegatingClassLoader` were to start doing so, it doesn't mean we should rely on it. 

So with that in mind, there's an important subtlety of the use case with OSGi: As bundles are composed into applications, it's pretty easy to unexpectedly pick up configuration from the transitive closure of direct dependencies when using `BundleDelegatingClassLoader`. Once again, this isn't different than a monolithic app, but the ease of mixing and matching bundles makes it a lot more likely. There is certainly a need for prioritization of configuration, but it's probably adequate to document the use of `akka.osgi.ActorSystemActivator#getActorSystemConfiguration` with some examples and use cases. Maybe

Sound reasonable so far? :)

briantopping:
Here's a start of doc, interested in feedback. TMI? 

I think the next steps are generating `ActorSystemActivator` examples at various degrees of usage, both with and without Blueprint, which is a pretty common deployer.

[Akka in OSGi — Akka Documentation.pdf](https://github.com/akka/akka/files/76800/Akka.in.OSGi.Akka.Documentation.pdf)

briantopping:
[...]

PR #19330, opened 2016-01-02
=osgi #19323 Order classpath resolution in `BundleDelegatingClassLoader` as a breadth-first search

briantopping:
This may still be inadequate, but should reduce some of the uncertainty around using `BundleDelegatingClassLoader`. Let's add to this PR with much stronger documentation after some discussion.

### DIFF
--- a/akka-osgi/src/main/scala/akka/osgi/BundleDelegatingClassLoader.scala
+++ b/akka-osgi/src/main/scala/akka/osgi/BundleDelegatingClassLoader.scala
@@ -35,10 +35,10 @@ object BundleDelegatingClassLoader {
  */
 class BundleDelegatingClassLoader(bundle: Bundle, fallBackClassLoader: ClassLoader) extends ClassLoader(fallBackClassLoader) {
 
-  private val bundles = findTransitiveBundles(bundle).toList
+  private val bundles = findTransitiveBundles(bundle)
 
   override def findClass(name: String): Class[_] = {
-    @tailrec def find(remaining: List[Bundle]): Class[_] = {
+    @tailrec def find(remaining: Seq[Bundle]): Class[_] = {
       if (remaining.isEmpty) throw new ClassNotFoundException(name)
       else Try { remaining.head.loadClass(name) } match {
         case Success(cls) ⇒ cls
@@ -49,7 +49,7 @@ class BundleDelegatingClassLoader(bundle: Bundle, fallBackClassLoader: ClassLoad
   }
 
   override def findResource(name: String): URL = {
-    @tailrec def find(remaining: List[Bundle]): URL = {
+    @tailrec def find(remaining: Seq[Bundle]): URL = {
       if (remaining.isEmpty) getParent.getResource(name)
       else Option { remaining.head.getResource(name) } match {
         case Some(r) ⇒ r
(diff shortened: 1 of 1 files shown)


ktoso:
OK TO TEST

ktoso:
Refs #19323

akka-osgi/src/main/scala/akka/osgi/BundleDelegatingClassLoader.scala
```diff
@@ -66,8 +66,8 @@ class BundleDelegatingClassLoader(bundle: Bundle, fallBackClassLoader: ClassLoad
     java.util.Collections.enumeration(resources.asJava)
   }
 
-  private def findTransitiveBundles(bundle: Bundle): Set[Bundle] = {
-    @tailrec def process(processed: Set[Bundle], remaining: Set[Bundle]): Set[Bundle] = {
+  private def findTransitiveBundles(bundle: Bundle): Seq[Bundle] = {
+    @tailrec def process(processed: Seq[Bundle], remaining: Seq[Bundle]): Seq[Bundle] = {
```
rkuhn: `processed` should remain a `Set` because it is used only for membership tests

rkuhn: nope, discard my comment

rkuhn:
I cannot reconcile the PR title with the code change: it seems to me that it does a breadth-last approach, i.e. the most deeply nested bundle will now be scanned first and the initial (i.e. top-level) bundle is scanned last.

akka-osgi/src/main/scala/akka/osgi/BundleDelegatingClassLoader.scala
```diff
@@ -76,20 +76,20 @@ class BundleDelegatingClassLoader(bundle: Bundle, fallBackClassLoader: ClassLoad
...
-                wiring.getRequiredWires(BundleRevision.PACKAGE_NAMESPACE).asScala.toList
+              val requiredWires: Seq[BundleWire] =
+                wiring.getRequiredWires(BundleRevision.PACKAGE_NAMESPACE).asScala.toSeq
               requiredWires.flatMap {
                 wire ⇒ Option(wire.getProviderWiring) map { _.getBundle }
-              }.toSet
+              }
             }
-          process(processed + b, rest ++ (direct -- processed))
+          process(Seq(b) ++ processed, rest ++ direct.diff(processed))
```
rkuhn: Since `processed` is only prepended to, why not just call it a `List` and use `::`?

briantopping: I think that's a mistake on my part, it should be appending. I'll follow up on the issue itself in a minute.

briantopping:
Yes, I think your prepend comment is correct (but it is at least *ordered* now :)). 

Next steps ought to be getting a unit test together for this. It shouldn't be difficult and would help with reasoning about it.

ktoso:
Seems like this one as a minor comment to address before merging, right?

drewhk:
I am closing this temporarily. Please reopen when there is new progress.

briantopping:
Yes, that's fine. There's only four files, I've copied them locally for the time being and am deciding whether to continue working on `akka-osgi` or create a separate project.

Issue #19293, opened 2015-12-26
Fix remaining issues in Sink.seq, Flow.limit and Flow.limitWithin


lolski:
As found by @RichardBradley, PR https://github.com/akka/akka/pull/18965 has some lingering issue that needs to be fixed. To quote https://github.com/akka/akka/issues/18921#issuecomment-166253591:

- Loads of `stream.grouped(n).runWith(head)` expressions in Akka (mainly in test) can now be replaced with `stream.runWith(seq)` -- see #18267
- The current implementation of `Sink.seq` will silently do the wrong thing if the stream is longer than `INT.MAX_VALUE`. I think this deserves fixing, or at the very least a note in the docs. Notably, `limit` and other length operators take a `Long`, so Akka does seem to allow streams longer than `Int.MAX_VALUE`.
- The `stream-cookbook.rst` hasn't been updated, and still suggests the old recipe. See #18267 for a fix.
- Possible incorrect "emits when" description for `limit` and `limitWithin`
- `Sink.seq` (Scala DSL) should return `immutable.Seq` rather than `Seq`

ktoso:
> Loads of "stream.grouped(n).runWith(head)" expressions in Akka (mainly in test) can now be replaced with "stream.runWith(seq)" -- see #18267

Please be aware of https://github.com/akka/akka/pull/18503 which is there since a long time (test kit support for `toStrict`)

lolski:
@ktoso I assume we want to replace those calls with `seq` as well?

lolski:
Added another issue where `Seq` is returned instead of `immutable.Seq`

ktoso:
Are you planning to pick those up @lolski? Thanks in advance if so :)

lolski:
@ktoso Yup, already about half-way on it.

Issue #19094, opened 2015-12-05
Http client causes 'Cannot pull port twice' error


daniel-trinh:
I'm not 100% sure pipelining is the cause, although this started occuring when I set akka.http.client.host-connection-pool.pipelining-limit to something greater than 1.

A multi build project has been set up that can reproduce this, with two projects in the link below, one as a dummy echo server to receive http requests, and another sub project that uses akka http client to send requests to the echo server.

https://github.com/daniel-trinh/echo_server

This problem also occurs on sending requests to a server unrelated to akka-http, so I imagine this would be reproducible with any server that can accept requests from the client.

cc @ktoso
```
[ERROR] [12/04/2015 23:02:31.610] [vulcan-loader-akka.actor.default-dispatcher-14] [ActorSystem(vulcan-loader)] Outgoing request stream error
java.lang.IllegalArgumentException: requirement failed: Cannot pull port twice
	at scala.Predef$.require(Predef.scala:219)
	at akka.stream.stage.GraphStageLogic.pull(GraphStage.scala:408)
	at akka.http.impl.engine.client.PoolConductor$SlotSelector$$anon$1$$anonfun$1.apply$mcV$sp(PoolConductor.scala:149)
	at akka.stream.stage.GraphStageLogic$Emitting.followUp(GraphStage.scala:741)
	at akka.stream.stage.GraphStageLogic$EmittingSingle.onPull(GraphStage.scala:790)
	at akka.stream.impl.fusing.GraphInterpreter.processEvent(GraphInterpreter.scala:553)
	at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:509)
	at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$runBatch(ActorGraphInterpreter.scala:408)
	at akka.stream.impl.fusing.ActorGraphInterpreter$$anonfun$receive$1.applyOrElse(ActorGraphInterpreter.scala:350)
	at akka.actor.Actor$class.aroundReceive(Actor.scala:467)
	at akka.stream.impl.fusing.ActorGraphInterpreter.aroundReceive(ActorGraphInterpreter.scala:292)
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:516)
	at akka.actor.ActorCell.invoke(ActorCell.scala:487)
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:238)
[...]

PR #19332, opened 2016-01-03
WIP lolski issue 19293

lolski:
See https://github.com/akka/akka/issues/19293

Contains fixes for most of the issues except a few javadsl counterparts and this:
- The current implementation of Sink.seq will silently do the wrong thing if the stream is longer than INT.MAX_VALUE. I think this deserves fixing, or at the very least a note in the docs. Notably, limit and other length operators take a Long, so Akka does seem to allow streams longer than Int.MAX_VALUE.
- Should we rename `Sink.seq` in Java DSL to `Sink.toList`?


ktoso:
OK TO TEST

akka-docs-dev/rst/scala/code/docs/stream/BidiFlowDocSpec.scala
```diff
@@ -160,14 +160,14 @@ class BidiFlowDocSpec extends AkkaSpec with ConversionCheckedTripleEquals {
...
-      val f = Source((0 to 9).map(Ping)).via(flow).grouped(20).runWith(Sink.head)
+      val f = Source((0 to 9).map(Ping)).via(flow).limit(20).runWith(Sink.seq)
       Await.result(f, 1.second) should ===((0 to 9).map(Pong))
     }
 
     "work when accumulated" in {
       val stack = codec.atop(framing)
       val flow = stack.atop(accumulate).atop(stack.reversed).join(Flow[Message].map { case Ping(id) => Pong(id) })
-      val f = Source((0 to 9).map(Ping)).via(flow).grouped(20).runWith(Sink.head)
+      val f = Source((0 to 9).map(Ping)).via(flow).limit(20).runWith(Sink.seq)
```
ktoso: OK

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeSeq.scala
```diff
@@ -0,0 +1,56 @@
...
+      val unsafe: Future[Seq[Message]] = myData.runWith(Sink.seq) // dangerous!
+      //#draining-to-seq-unsafe
+
+      Await.result(unsafe, 3.seconds) should be(result)
+    }
+
+    "work together with limit(n)" in {
+      val result = List("1", "2", "3")
+      val myData = Source(result)
+      val MaxAllowedSeqSize = 100
```
ktoso: move this into the code snippet please

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeSeq.scala
```diff
@@ -0,0 +1,56 @@
...
+      val safe1: Future[immutable.Seq[Message]] = myData.limit(MaxAllowedSeqSize).runWith(Sink.seq)
+      //#draining-to-seq-safe
+
+      Await.result(safe1, 3.seconds) should be(result)
+    }
+
+    "work together with take(n)" in {
+      val result = List("1", "2", "3")
+      val myData = Source(result)
+      val MaxAllowedSeqSize = 100
```
ktoso: move this into the code snippet please

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeSeq.scala
```diff
@@ -0,0 +1,56 @@
...
+import scala.concurrent.duration._
+
+class RecipeSeq extends RecipeSpec {
+
+  "Recipe for draining a stream into a strict collection" must {
+
+    "work" in {
+      val result = immutable.Seq[Message]("1", "2", "3")
+      val myData = Source(result)
+      val MaxAllowedSeqSize = 100
```
ktoso: not used, please remove

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeSeq.scala
```diff
@@ -0,0 +1,56 @@
...
+
+    "work together with limit(n)" in {
+      val result = List("1", "2", "3")
+      val myData = Source(result)
+      val MaxAllowedSeqSize = 100
+
+      //#draining-to-seq-safe
+      // OK. Future will fail with a `StreamLimitReachedException`
+      // if the number of incoming elements is larger than MaxAllowedSeqSize
+      val safe1: Future[immutable.Seq[Message]] = myData.limit(MaxAllowedSeqSize).runWith(Sink.seq)
```
ktoso: please add trailing newline here, otherwise looks weird (concated snippets) in docs AFAIR

akka-http-core/src/test/scala/akka/http/impl/engine/parsing/ResponseParserSpec.scala
```diff
@@ -204,10 +204,10 @@ class ResponseParserSpec extends FreeSpec with Matchers with BeforeAndAfterAll {
...
       "response with additional transfer encodings" in new Test {
         Seq("""HTTP/1.1 200 OK
-          |Transfer-Encoding: fancy, chunked
-          |Cont""", """ent-Type: application/pdf
-          |
-          |""") should generalMultiParseTo(
+              |Transfer-Encoding: fancy, chunked
+              |Cont""", """ent-Type: application/pdf
+                          |
+                          |""") should generalMultiParseTo(
```
ktoso: please align :)

akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTestResultComponent.scala
```diff
@@ -25,6 +25,7 @@ trait RouteTestResultComponent {
    * A receptacle for the response or rejections created by a route.
    */
   class RouteTestResult(timeout: FiniteDuration)(implicit fm: Materializer) {
+    import fm.executionContext
```
ktoso: why wasnt it needed before? Seems only change below should not require adding anything

akka-stream-tests/src/test/java/akka/stream/javadsl/FlowGraphTest.java
```diff
@@ -82,7 +82,7 @@ public void mustBeAbleToUseMerge() throws Exception {
 
     // collecting
     final Publisher<String> pub = source.runWith(publisher, materializer);
-    final Future<List<String>> all = Source.fromPublisher(pub).grouped(100).runWith(Sink.<List<String>>head(), materializer);
+    final Future<List<String>> all = Source.fromPublisher(pub).limit(100).runWith(Sink.<String>seq(), materializer);
```
ktoso: shouldn't this use the `Sink.toList`? It's Java.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -144,9 +145,9 @@ object Sink {
    *
    * See also [[Flow.limit]], [[Flow.limitWeighted]], [[Flow.take]], [[Flow.takeWithin]], [[Flow.takeWhile]]
    */
-  def seq[T]: Sink[T, Future[Seq[T]]] = {
+  def seq[T]: Sink[T, Future[immutable.Seq[T]]] = {
```
ktoso: good, thanks

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -617,16 +617,13 @@ trait FlowOps[+Out, +Mat] {
...
-   * or negative.
-   *
-   * '''Emits when''' the specified number of elements to take has not yet been reached
+   * '''Emits when''' upstream emits and the number of emitted elements has not reached max
    *
    * '''Backpressures when''' downstream backpressures
    *
-   * '''Completes when''' the defined number of elements has been taken or upstream completes
+   * '''Completes when''' upstream completes and the number of emitted elements has not reached max
    *
```
ktoso: For this one we actually should document when it errors then, shouldn't we?

akka-stream/src/main/scala/akka/stream/StreamLimitReachedException.scala
```diff
@@ -1,3 +1,6 @@
 package akka.stream
 
+/**
+ * Copyright (C) 2015 Typesafe Inc. <http://www.typesafe.com>
+ */
```
ktoso: Please move to line `1` – above the `package akka.stream` line

akka-samples/akka-docs-java-lambda/src/test/java/docs/stream/cookbook/RecipeByteStrings.java
```diff
@@ -96,7 +96,7 @@ public SyncDirective emitChunkOrPull(Context<ByteString> ctx) {
           rawBytes.transform(() -> new Chunker(CHUNK_LIMIT));
         //#bytestring-chunker2
 
-        Future<List<ByteString>> chunksFuture = chunksStream.grouped(10).runWith(Sink.head(), mat);
+        Future<List<ByteString>> chunksFuture = chunksStream.runWith(Sink.toList(mat.executionContext), mat);
```
ktoso: /me slightly confused... We don't seem to have toList though

lolski: No, we do not. I must have missed this.

akka-stream-tests/src/test/java/akka/stream/javadsl/FlowGraphTest.java
lolski: `Sink.seq` is the name for both the javadsl and scaladsl.

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeSeq.scala
ktoso: I mean this should be visible in the snippet. So move it into the `  //#draining-to-seq-safe` delimited block.

lolski: @ktonga ah, got it

lolski: @ktoso got it

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
lolski: done

akka-docs-dev/rst/scala/code/docs/stream/cookbook/RecipeSeq.scala
lolski: done

ktoso:
PLS BUILD

lolski:
The error was caused by a failure in a test in `FlowScanSpec`, i.e., this line:

`Source.single(1).concat(Source.maybe[Int]).scan(0)(_ + _).limit(2).runWith(Sink.seq)`

A very short investigation last week shows that the test will pass if the call to `concat` is removed. Otherwise for some reason upstream won't push the element. This seems to be the case even with `grouped(n)`/`Sink.head` if `n > actualElementCount`.

I will investigate this issue after I've finished with the current ones.

lolski:
Basing `Sink.seq` on `GraphStageWithMaterializedValue` which is backed by a `Vector` buffer.

In this implementation, the method will keep on inserting element into the buffer until it finishes or runs out of memory.

Unlike the previous implementation based on `grouped`/`headOption`, it won't fail silently when dealing with elements larger than `Integer.MAX_VALUE`.

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -238,6 +239,42 @@ private[akka] final class HeadOptionStage[T] extends GraphStageWithMaterializedV
...
+  val in = Inlet[T]("seq.in")
+
+  override val shape: SinkShape[T] = SinkShape.of(in)
+
+  override def createLogicAndMaterializedValue(inheritedAttributes: Attributes) = {
+    val p: Promise[immutable.Seq[T]] = Promise()
+    val logic = new GraphStageLogic(shape) {
+      val buf = Vector.newBuilder[T]
+
+      override def preStart(): Unit = pull(in)
```
rkuhn: I think this should be a `tryPull()`, it is possible that the upstream completed already in its preStart hook before this one runs.

rkuhn: @drewhk thinking more about this I’d say that the current `pull` is actually useless as it can always fail spuriously

ktoso: pull() in preStart is actually also what causes the https://github.com/akka/akka/issues/19094 occassionally. That's as far as I got into researching that ticket.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -144,11 +145,7 @@ object Sink {
    *
    * See also [[Flow.limit]], [[Flow.limitWeighted]], [[Flow.take]], [[Flow.takeWithin]], [[Flow.takeWhile]]
    */
-  def seq[T]: Sink[T, Future[Seq[T]]] = {
-    Flow[T].grouped(Integer.MAX_VALUE).toMat(Sink.headOption)(Keep.right) mapMaterializedValue { e ⇒
-      e.map(_.getOrElse(Seq.empty[T]))(ExecutionContexts.sameThreadExecutionContext)
-    }
-  }
+  def seq[T]: Sink[T, Future[immutable.Seq[T]]] = Sink.fromGraph(new SeqStage[T]).withAttributes(DefaultAttributes.seqSink)
```
rkuhn: The attributes should be included via `GraphStage.initialAttributes` override.

rkuhn:
Great work, @lolski ! LGTM after the minor nitpicks.

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
drewhk: Hmm, shouldn't we ensure that preStart is called before any external events has been consumed, i.e. pulling is always valid?

rkuhn:

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
rkuhn: No, the preStart of another stage may have made it invalid without any external event.

drewhk: This is not true, you have to explicitly execute the enqueued event, otherwise it is not yet visible by another stage.

rkuhn: Ah, I was not sure that that is always true. I have observed completion to leak through before onUpstreamFinish is invoked, and this may be the cause of some of the current bugs.

drewhk: that is a bug then. Things should not be visible before the appropriate handler has been called. I.e. preStart should be able to pull without problems (as onUpstreamFinish cannot be called before preStart).

PR #19335, opened 2016-01-03
=str - Significant simplification of the implementation of Source.unf…

viktorklang:
…oldInf

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
@@ -260,7 +260,7 @@ object Source {
    * }}}
    */
   def unfold[S, E](s: S)(f: S ⇒ Option[(S, E)]): Source[E, Unit] =
-    Source.fromGraph(new Unfold(s, f)).withAttributes(DefaultAttributes.unfold)
+    Source.fromGraph(new Unfold(s, f))
 
   /**
    * Same as [[unfold]], but uses an async function to generate the next state-element tuple.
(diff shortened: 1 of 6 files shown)


akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -289,20 +289,8 @@ object Source {
...
-      val init = Source.single(s)
-
-      init ~> cnct ~> uzip.in
-      cnct <~ Flow[S].buffer(2, OverflowStrategy.backpressure) <~ uzip.out0
-
-      SourceShape(uzip.out1)
-    }).withAttributes(DefaultAttributes.unfoldInf)
-  }
+  def unfoldInf[S, E](s: S)(f: S ⇒ (S, E)): Source[E, Unit] =
+    unfold(s)(s ⇒ Some(f(s))).withAttributes(DefaultAttributes.unfoldInf)
```
viktorklang: I guess at this point the question is of the utility of the unfoldInf method itself.

viktorklang:
Test Failure is unrelated, issue here: https://github.com/akka/akka/issues/19342

johanandren:
PLS BUILD

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
drewhk: +1 for removing the inf version.

ktoso: OK to remove IMO too.

viktorklang: @drewhk When I remove `unfoldInf` and relace the usage thereof in `FusingSpec` it fails, because it is introspecting the structure of the graph, and the name-attribute of `Unfold` gets hidden in a CopiedModule where it's `attributes` do not match `copyOf.attributes`, so if I do `unfold(…).named("unfold").named("unfold")` the test passes.

I'm not sure whether the FusingSpec `def singlePath` makes improper assumptions or whether `CopiedModuled` has a bug. thoughts?

rkuhn:

Issue #19336, opened 2016-01-03
Akka Streams: FlowOpsMat Repr should be elevated to at least FlowOpsMat


timcharper:
In [FlowOpsMat](https://github.com/akka/akka/blob/akka-stream-and-http-experimental-2.0/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala#L1602), `Repr` is left as it was inherited from [FlowOps](https://github.com/akka/akka/blob/akka-stream-and-http-experimental-2.0/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala#L339).

I think if you are using an interface that implements FlowOpsMat, it's safe to assume that the `Repr` will implement FlowOpsMat, also.

This is an issue because if you implement a class which contains an object that implements `FlowOpsMat`, and then call `map` on the generic object, what you get back is a `FlowOps` interface and can't call any materialization methods.

timcharper:
I marked the pull request as API breaking, since it is possible (but I think extremely unlikely) for this to cause code to no longer compile.

There's potentially even more remote possibility this could lead to a binary compatibility issue, but not directly, but I can't come up with a concrete example of how it would break.

timcharper:
I'm working around it, for now; perhaps seeing this workaround will provide any missing context as to why I feel this is an issue.

https://github.com/timcharper/acked-stream/blob/akka-issue-19336/src/main/scala/com/timcharper/acked/AckedFlowOps.scala#L581

rkuhn:
The provided example does not show the problem, though, because the type of `wrappedRepr` is `Flow` (which works fine) and not `FlowOpsMat`, right?

I just pushed an update to #19317 that should fix this issue (and a few more).

timcharper:
WrappedRepr is at least a FlowOpsMat

rkuhn:
Yes, but the value that is mapped (`wrappedRepr`) is a `Flow` and not a `WrappedRepr`.

timcharper:
I'm sorry Roland, I'm really trying hard to figure out where the miscommunication is happening.

[...]

PR #19337, opened 2016-01-03
!str #19336 make FlowOpsMat#Repr more narrow

timcharper:
Rational for this change is explained in the referenced issue, https://github.com/akka/akka/issues/19336


ktoso:
OK TO TEST

ktoso:
Reasoning and fix look sound to me, LGTM and ping to @rkuhn for review.

rkuhn:
I agree with the goal but the fix should be aligned with my open PR, I can try to expand that tomorrow.

timcharper:
Just went and read your [changes](https://github.com/akka/akka/pull/19317). Neat, @rkuhn !  I also had to do that flattening and I was thinking about a way to avoid it. Super exciting!

timcharper:
Resolved by https://github.com/akka/akka/pull/19317

PR #19349, opened 2016-01-05
+htc #19237 entity#withSizeLimit for Java + add withoutSizeLimit

### LINKED ISSUE #19237: HttpEntity change interface: add withNoLimit
HttpEntity has a method: `HttpEntity.withSizeLimit`, but it has no way to indicate that there should be no limit. It is possible to configure `infinite` in the `application.conf`, but then it would apply to all entities.

A workaround is to set `HttpEntity.withSizeLimit(Long.MaxValue)`, which is near infinite. Having `HttpEntity.withNoLimit()` and/or `HttpEntity.withSizeLimit(-1)` would make it more explicit.

For the discussion see: https://groups.google.com/forum/m/#!topic/akka-user/DIZ2WY97dnA

ktoso:
Resolves #19237

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
@@ -85,6 +85,45 @@ public interface HttpEntity {
     Source<ByteString, Object> getDataBytes();
 
     /**
+     * Apply the given size limit to this entity by returning a new entity instance which automatically verifies that the
+     * data stream encapsulated by this instance produces at most `maxBytes` data bytes. In case this verification fails
+     * the respective stream will be terminated with an `EntityStreamException` either directly at materialization
+     * time (if the Content-Length is known) or whenever more data bytes than allowed have been read.
+     *
+     * When called on `Strict` entities the method will return the entity itself if the length is within the bound,
+     * otherwise a `Default` entity with a single element data stream. This allows for potential refinement of the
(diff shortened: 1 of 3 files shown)


akka-http-core/src/main/java/akka/http/javadsl/model/HttpEntity.java
```diff
@@ -85,6 +85,45 @@
...
+     * Content-Length and then another limit is applied then this new limit will be evaluated against the new
+     * Content-Length. If the entity is transformed in a way that changes the Content-Length and no new limit is applied
+     * then the previous limit will be applied against the previous Content-Length.
+     *
+     * Note that the size limit applied via this method will only have any effect if the `Source` instance contained
+     * in this entity has been appropriately modified via the `HttpEntity.limitable` method. For all entities created
+     * by the HTTP layer itself this is always the case, but if you create entities yourself and would like them to
+     * properly respect limits defined via this method you need to make sure to apply `HttpEntity.limitable` yourself.
+     */
+    HttpEntity withSizeLimit(long maxBytes);
```
ktoso: moved here to make it accessible to JavaDSL as well.

ktoso:
Please review when you have a second @sirthias @jrudolph

johanandren:
LGTM!

ktoso:
WDYT here @jrudolph ?

jrudolph:
LGTM

ktoso:
Thanks Johannes :)

Issue #19341, opened 2016-01-04
FAILED: akka.pattern.AskSpec


johanandren:
https://jenkins.akka.io:8498/job/akka-nightly/2930/consoleFull

500 ms timeout in "must work when reply uses actor selection"

PR #19351, opened 2016-01-05
=act #19341 increase timeout of askspec testcase

johanandren:
Fix for #19341, just increased the failing timeout, nothing wrong with the test.


ktoso:
3sec probably enough IMHO. LGTM, please merge after validation.

johanandren:
Yeah, 3s is what I've seen before, but the test case above had 5 so followed that.

ktoso:
Ok, sounds good. LGTM :)

Issue #17911, opened 2015-07-04
properly represent IPv6 addresses


rkuhn:
Currently we just grab any InetAddress from the socket and use its string representation, which may end up looking like `fe80::1:2:3:4%17` (the `%17` is the scope identifier for non-global addresses). Using this address directly within a URI will lead to interpreting the whole authority part as a blob without discerning host and port, RFC2732 specifies that IPv6 addresses must be enclosed in brackets within URIs.

Since `InetAddress.getByName` supports the bracket-enclosed syntax for literal addresses the proper fix is to mandate that `Transport.listen` must return an `Address` with a valid host part (i.e. `NettyTransport.addressFromSocketAddress` must treat IPv4 and IPv6 differently). Perhaps `Address` should guard against common problems by rejecting `host` portions that contain `:` but do not start and end with brackets.

rkuhn:
We should also document that `akka.remote.netty.tcp.host` must follow this format.

BTW: this was noticed only on JDK8 which seems to use different means of obtaining the local host address—JDK6 returned 127.0.0.1.

ktoso:
Sounds like something to do for `2.4.0`, right?

ktoso:
Oh, or M3, I don't mind either way I guess, has to be done.

patriknw:
See also https://groups.google.com/d/msg/akka-user/K9eOHzbUQ8I/1yaqYhtEAwAJ

ktoso:
Actually not fixed, @johanandren is on it.

johanandren:
I started on a fix, but got stuck with problems because the host name encoded in actor path and then decoded later on the remote.

PR #19353, opened 2016-01-05
=act,rem #17911 handle IPv6 Addresses better

johanandren:
Refs #17911 
- NettyTransport.listen will autowrap IPv6 addresses in brackets if needed
- Address asserts that IPv6 addresses are wrapped in brackets

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/actor/AddressSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/actor/AddressSpec.scala
@@ -1,0 +1,32 @@
+/**
+ * Copyright (C) 2009-2016 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.actor
+
+import org.scalatest.{ WordSpec, ShouldMatchers }
+
+@org.junit.runner.RunWith(classOf[org.scalatest.junit.JUnitRunner])
+class AddressSpec extends WordSpec with ShouldMatchers {
+
+  "The Address factory methods" should {
+
+    "fail if ipv6 address is not wrapped in brackets" in {
+      intercept[IllegalArgumentException] {
+        Address("tcp", "system", "0:0:0:0:0:0:0:1", 2551)
+      }.getMessage should include("must be wrapped with []")
+    }
+
+    "not fail if ipv6 address is wrapped in brackets" in {
+      Address("tcp", "system", "[0:0:0:0:0:0:0:1]", 2551)
+
+    }
+
+    "not fail with an ipv4 address" in {
+      Address("tcp", "system", "192.168.0.1", 2551)
+    }
+
+    "not fail with a hostname address" in {
+      Address("tcp", "system", "example.com", 2551)
+    }
+  }
+}
--- a/akka-actor/src/main/scala/akka/actor/Address.scala
+++ b/akka-actor/src/main/scala/akka/actor/Address.scala
@@ -75,7 +75,11 @@ object Address {
   /**
    * Constructs a new Address with the specified protocol, system name, host and port
    */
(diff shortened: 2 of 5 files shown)


akka-actor-tests/src/test/scala/akka/actor/AddressSpec.scala
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (C) 2009-2016 Typesafe Inc. <http://www.typesafe.com>
```
ktoso: *2016, here we go!* :+1:

ktoso:
Good timing with that one :)

akka-actor-tests/src/test/scala/akka/actor/AddressSpec.scala
```diff
@@ -0,0 +1,32 @@
...
+
+@org.junit.runner.RunWith(classOf[org.scalatest.junit.JUnitRunner])
+class AddressSpec extends WordSpec with ShouldMatchers {
+
+  "The Address factory methods" should {
+
+    "fail if ipv6 address is not wrapped in brackets" in {
+      intercept[AssertionError] {
+        Address("tcp", "system", "0:0:0:0:0:0:0:1", 2551)
+      }
```
ktoso: `.getMessage should include "must be wrapped with []"`?

akka-actor/src/main/scala/akka/actor/Address.scala
```diff
@@ -75,7 +75,11 @@ object Address {
   /**
    * Constructs a new Address with the specified protocol, system name, host and port
    */
-  def apply(protocol: String, system: String, host: String, port: Int) = new Address(protocol, system, Some(host), Some(port))
+  def apply(protocol: String, system: String, host: String, port: Int) = {
+    assert(!host.contains(":") || host.startsWith("["), s"IPv6 address $host must be wrapped with [] to be safe")
```
ktoso: assert is **deadly** :fire: Please use `require` instead.

// `assert` throws `AssertionError`, which is an `Error`, not exception, which can cause the entire `ActorSystem` to blow up, because `Error` is not `NonFatal`

akka-remote/src/main/scala/akka/remote/transport/netty/NettyTransport.scala
```diff
@@ -238,16 +238,32 @@ private[transport] object NettyTransport {
...
-      hostName.getOrElse(sa.getAddress.getHostAddress), port.getOrElse(sa.getPort))) // perhaps use getHostString in jdk 1.7
-    case _ ⇒ None
+                               hostName: Option[String], port: Option[Int]): Option[Address] = {
+
+    addr match {
+      case sa: InetSocketAddress ⇒
+        Some(Address(
+          schemeIdentifier,
+          systemName,
+          ipv6SafeHostname(hostName.getOrElse(sa.getHostString)),
```
ktoso: we're JDK8 now here, so please `getHostString` :-) as hinted in comment `// perhaps use getHostString in jdk 1.7`

ktoso:
Minor fixups needed :)

akka-remote/src/main/scala/akka/remote/transport/netty/NettyTransport.scala
johanandren: Isn't that what I am doing?

ktoso: Too long day :( Yep, sorry.

johanandren:
Minor fixups done! :)

ktoso:
LGTM, thanks!

Issue #19357, opened 2016-01-05
Grammatical errors in CONTRIBUTING.md


kubukoz:
I found some sentences that should probably be changed as follows:

`line 13`
We encourage changes that make~~s~~ it easier
to
We encourage changes that make it easier

`line 117`
This can be done ~~in~~ either in the documentation or in the build file next to each dependency
to
This can be done either in the documentation or in the build file next to each dependency

`line 121`
This can be done by either prefix the name with wip- as in ‘Work In Progress’, or use hierarchical names
to
This can be done by either prefix**ing** the name with wip- as in ‘Work In Progress’, or us**ing** hierarchical names

`line 129`
or work in a branch for long time doing merges/rebases etc
to
or work in a branch for **a** long time doing merges/rebases etc

I will create a PR within a few minutes ;)

kubukoz:
Also, some commas are missing. I'll add them if I encounter their rightful positions.

PR #19358, opened 2016-01-05
=doc #19357 fix grammar in CONTRIBUTING.md

kubukoz:
Replaces:

We encourage changes that make~~s~~ it easier
->
We encourage changes that make it easier

This can be done ~~in~~ either in the documentation or in the build file next to each dependency
->
This can be done either in the documentation or in the build file next to each dependency

This can be done by either prefix the name with wip- as in ‘Work In Progress’, or use hierarchical names
->
This can be done by either prefix**ing** the name with wip- as in ‘Work In Progress’, or us**ing** hierarchical names

or work in a branch for long time doing merges/rebases etc
->
or work in a branch for **a** long time doing merges/rebases etc

Also added a comma in `line 121`: `If so, then it is preferable`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,7 @@ In case of questions about the contribution process or for discussion of specifi
 
 # Typesafe Project & Developer Guidelines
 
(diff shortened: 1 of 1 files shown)


ktoso:
Thanks a lot, these must have been there since forever :)

ktoso:
Refs #19357

Issue #19331, opened 2016-01-03
`FileAndResourceDirectives#getFromResourceDirectory` fails under OSGi


briantopping:
`FileAndResourceDirectives#getFromResourceDirectory` with a classloader set to that of a bundle eventually calls https://github.com/akka/akka/blob/a19050dbcbb7b28a7533a5db1019d9b521ffc210/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala#L206-L227 with a URL using a `bundle` scheme under Karaf. 

Under OSGi, a bundle does not have reliable access to the filesystem to open a bundle in the way that a JAR is being opened here. I may not be able to fix this one, but I am very curious to understand why registered URL handlers were avoided in this manner.  

What can I do to help here?

rkuhn:
I have no clue. @sirthias @jrudolph why does this not just use `ClassLoader.getResourceAsStream`? @briantopping would that fix it? (I assume so, but my understanding of OSGi is that it is broken in interesting ways ;-) )

briantopping:
Thanks Roland, that would definitely fix it in a manner that is impervious to any quirks that exist across implementations.

2beaucoup:
We used `classLoader.getResource` but switched to the current impl while fixing #17495.

Maybe we should use URL handlers just for the default case?

briantopping:
Oh interesting, makes sense. Yes, default case would be fine I'm sure. Shall I cook up a PR? Or I can test on this end once one is put together. Thanks!

briantopping:
I've been having weird problems with the build all day and haven't been able to test the patch I have together. I believe changing the `file` case in https://github.com/akka/akka/blob/a19050dbcbb7b28a7533a5db1019d9b521ffc210/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala#L208-L225 to the default case will solve the issue and should solve it for any other interesting cases that arise. :)

rkuhn:
@briantopping what kind of problems? If you are reasonably sure of the patch, you can also submit it as a PR and have the validator test it, but we should definitely sort out any build issues that prevent contributors from contributing!

ktoso:
Let me know what exactly blows up for you @briantopping and should be able to help out. Not sure why you compile using `+compile` during development?
PR very welcome here ofc, should retain current semantics etc :)

briantopping:
[...]

PR #19363, opened 2016-01-05
=htp #19331 Fix #17495 for default stream types

### LINKED ISSUE #17495: FileAndResourceDirectives.getFromResourceDirectory with sub directories may list files

briantopping:


akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -222,7 +222,16 @@ object FileAndResourceDirectives extends FileAndResourceDirectives {
             ResourceFile(url, entry.getSize, entry.getTime)
           }
         } finally jar.close()
-      case _ ⇒ None
+      case _ ⇒
+        val conn = url.openConnection()
+        try {
+          conn.setUseCaches(false) // otherwise the JDK will keep the JAR file open when we close!
```
ktoso: we're not in the JAR case here... what am I missing?

briantopping: Under OSGi, the URL scheme is `bundle`, can't really make a case for every scheme, not sure what schemes any given container might load. Ideas?

ktoso: so it's for when we're inside a jar, and ResourceFile was applied again... I see

ktoso:
OK TO TEST

briantopping:
Thanks Konrad

ktoso:
No prob ;) Will look more into it tomorrow. Handling any schema with `openConnection` is a bit scary (well technically it could be that some crazy person passes in `http://` even if in theory we shouldn't and we wouldn't... hm)

briantopping:
This is actually the old code modified to handle the new return monad. I don't think the scheme is ever provided by the user, rather it is generated by the immediate caller: https://github.com/briantopping/akka/blob/2aa32b3144dc0cde8ba1e1e422a7d62a5df688a9/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala#L86-L86.

2beaucoup:
Why the ref to #17495 in the commit message? Does this still prevent directory listing in the bundle case?

A test would be nice although I guess this is only reproducible in an OSGI env.

briantopping:
Not sure I understand, #17495 is the commit that broke functionality under OSGi and any other container that creates custom URL schemes. Should I have referenced it differently?

Yes, it does prevent directory listing in the bundle case, but the original change would not have been accepted in the first place if it got proper testing and akka-http is currently broken for all OSGi users.

Tests would be good, but given that containers are likely to have URL handlers that cannot handle directory listing, there's simply no way to write the tests.

I've been avoiding asking the obvious question of whether we need directory listing at this much expense, but I think it's relevant to be asking.

cc: @jrudolph @sirthias

2beaucoup:
Just to clarify: #17495 was about preventing accidential directory listings. Therefore I asked if this will be protected with your fix cause you referenced it in the commit.

The comment about testing was about loading custom URI schemes and had no relation to the listings.

briantopping:
I guess I'm really confused now, apologies. #17495 broke the ability for any resources at all to be read in an OSGi container due to the explicit checking of URL schemes and rejecting all schemes that it didn't understand. Custom URL schemes are very common in application server environments, there's really no way to know how many there will be. I don't know anything about the directory listings TBH, I just know that #17495 caused one or more classes of deployment containers to fail and there's no way to know how many of them failed. 

Wouldn't you agree that directory listing functionality is secondary to making resources completely inaccessible while running in affected deployment containers? What am I missing here?

2beaucoup:
I strongly agree that we have a problem here and the fix basically LGTM.

#17495 fixed a security problem but unfortunately introduced a regression in `FileAndResourceDirectives.getFromResource`. All I (as the one responsible for the breakage) wanted to make sure is:
- to verify that this PR actually fixes loading resources in a container context. Unit testing this looks hard.
- that #17495 isn't reintroduced for custom schemes. Maybe you can try manually by referring to a bundle directory without a trailing slash.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
2beaucoup: simply rename "JAR" -> "connection" ?

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -222,7 +222,16 @@ object FileAndResourceDirectives extends FileAndResourceDirectives {
...
+          val len = conn.getContentLength
+          val lm = conn.getLastModified
+          Some(ResourceFile(url, len, lm))
+        } catch {
+          case _: Throwable ⇒ None
```
2beaucoup: Dunno if we should swallow everything here.

To defend against a possible `NullPointerException` and detect directories (`getInputStream == null`) you could use:
```scala
for {
  c <- Option(url.openConnection())
  _ = c.setUseCaches(false) // otherwise the JDK will keep the resource open when we close!
  is <- Option(c.getInputStream)
} yield {
  is.close()
  ResourceFile(url, c.getContentLength, c.getLastModified)
}
```

jrudolph:
> Wouldn't you agree that directory listing functionality is secondary to making resources completely inaccessible while running in affected deployment containers?

#17495 was actually about the functionality *not to* list directories inadvertently. The directory listing functionality was provided by the JDK for the `jar` URL connection handler automatically but we don't want to leak directory listings. What @2beaucoup means is that we now need to verify that other URL scheme handlers like the `bundle` one don't have a similar "feature" which would leak bundle content metadata inadvertently.

briantopping:
Hi thanks for the explanations. I misunderstood the the initial issue that directory listings were being added, not that security issues were being addressed. 

@jrudolph I think the problem is there's no way to contain the set of "other URL handlers". How many application containers are there? Are they documented in one place for each container? How deep do we have to dig to discover them? How do we keep up with new ones that are introduced? 

Could it be we're chasing how to handle an unbounded set of URL handlers to solve the case of the directory listings being insecure when maybe the original code was correct and directory listings are the problem that should be addressed? It seems like this may be the path of reduced overall complexity.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
briantopping: Appreciate the input. My caution with your approach is we agree that it's very difficult to test this and the old code worked for quite some time. 

Considering the NPE case, I don't see where the developer or end-user can introduce a URL to the call chain. As best I can tell, they are only generated by the JRE for known good entities. 

That said, is there some manner that your code suggestion could replace what was done in #17495 and generically cover custom URL scheme handlers in the process?

2beaucoup: I was referring to catching `Throwable` here, which wasn't there before and is considered bad practice.

The example above is basically equivalent to what we had before #17495 but additionally filters out directories (comment added).

If you can verify that it works as is then all is well.

2beaucoup:
> I think the problem is there's no way to contain the set of "other URL handlers".

No one suggested doing that. I proposed a generic solution [above](https://github.com/akka/akka/pull/19363#discussion_r49053825), but can't verify that it works since I don't have the test environment available.

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
briantopping: Not sure I understand, which line are we referring to here?

2beaucoup: The above (#228). New version would be
```
conn.setUseCaches(false) // otherwise the JDK will keep the connection open when we close!
```

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
briantopping: Totally understand, I'm a fan of the process and detail that goes into these decisions. 

The wildcard catch matches the previous code in that the semantics changed from a default value of `0` for the extracted tuple of `length` and `lastModified` to a monad that required provision. The try-finally pattern of the old code had an implicit catch of all values that are ignored. In order to generate the result for the new monad, the implicit catch was made explicit.

I totally appreciate this is frowned upon, but introducing new semantics also introduces risk, which I like to avoid as a new contributor. :)

briantopping:
> No one suggested doing that. I proposed a generic solution above, but can't verify that it works since I don't have the test environment available.

Please let me know if you'd like me to test any PRs you have!

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
2beaucoup: > The wildcard catch matches the previous code in that the semantics changed from a default value of 0 for the extracted tuple of length and lastModified to a monad that required provision. The try-finally pattern of the old code had an implicit catch of all values that are ignored.

Were are the default `0` values and the implicit catch in the [old version](https://github.com/2beaucoup/akka/blob/aa8da46068795a1110ed3ce07cf648fedaf1000d/akka-http-scala/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala#L92-L99)?

briantopping: I guess there's a limit to this. Maybe we can close this issue if it's not sufficient.

2beaucoup: Brian, I don't meant to be annoying, here. I just asked you to remove the two lines above (232, 233) to restore the original state of this code without the `catch`. That lets the `ExceptionHandler` do its thing and prevents catching stuff like `OutOfMemoryException` by accident.

briantopping: My bad, I forgot the compiler was smart enough to handle the case where the block either generates a value or throws and exception and doesn't need an explicit value in the latter.

I'll squash these shortly.

2beaucoup:
@briantopping:
> Please let me know if you'd like me to test any PRs you have!

See #19380.

rkuhn:

briantopping:
Moved to #19415

PR #19369, opened 2016-01-06
=htc disable fuzzing-mode by default in TestServer

ktoso:
So I won't catch myself again and again measuring something with fuzzing on ;)


ktoso:
I don't think anyone will mind this, merging.

// cc @2m @johanandren

ktoso:
Hah, actually noticed now that the exception message contains wrong setting key. Good to review things.

ktoso:
Also fixes setting key reported in warning when fuzzing enabled (valid key is: https://github.com/akka/akka/blob/release-2.3-dev/akka-stream/src/main/resources/reference.conf#L57 )

ktoso:
Pleave sanity chekc @2m @johanandren

PR #19370, opened 2016-01-07
=str include port name when require fails in GraphStage

ktoso:
Had this idea while looking at the stack-trace in https://github.com/akka/akka/issues/19094

Having port names would simplify "where to look" when these exceptions show up as it's not clear from the stacktrace which port, since we have 2 inlets, and arrived at pull() from emitting (via `followUp()`), so not sure which port exactly is being pulled.

Example of such stacktrace:

```
    at akka.stream.stage.GraphStageLogic.pull(GraphStage.scala:408)
    at akka.http.impl.engine.client.PoolConductor$SlotSelector$$anon$1$$anonfun$1.apply$mcV$sp(PoolConductor.scala:149)
    at akka.stream.stage.GraphStageLogic$Emitting.followUp(GraphStage.scala:741)
    at akka.stream.stage.GraphStageLogic$EmittingSingle.onPull(GraphStage.scala:790)
```

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
+++ b/akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
@@ -416,8 +416,8 @@ abstract class GraphStageLogic private[stream] (val inCount: Int, val outCount: 
       interpreter.pull(conn(in))
     } else {
       // Detailed error information should not add overhead to the hot path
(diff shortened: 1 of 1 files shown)


johanandren:
LGTM!

ktoso:
I'll merge this to help us find the problem.

PR #19372, opened 2016-01-07
Revert "=act,rem #17911 handle IPv6 Addresses better"

ktoso:
Reverts akka/akka#19353

As @johanandren points out after a manual test, this wasn't quite enough just yet.


ktoso:
Refs #17911

Issue #19251, opened 2015-12-21
Revert Java 6 modification to Multipart JavaDSL


ktoso:
Once we drop JDK6 we can revert this commit: https://github.com/akka/akka/pull/19250

PR #19375, opened 2016-01-07
Revert "!htc fix multipart model (java6 had trouble)"

ktoso:
Resolves https://github.com/akka/akka/issues/19251

Since the upcoming HTTP release is aiming to use Java 8, we can revert this change.
See the above issue for details.

Reverts akka/akka#19250


drewhk:
Ok, I don't really get this, but I trust you. Is there a reason for not merging yet? Btw, given recent discussions, this is not binary compatible.

ktoso:
This change should only be applied where we are Java 8, I wasnt sure when/how we move branches, so technically this should not be comitted where we'll maintain Java 6.

ktoso:
FYI @rkuhn

rkuhn:
`release-2.3-dev` is basically Java 8 now as we won’t cut another 2.0 release from it

ktoso:
Ok, got it :) merging is good then.

Issue #19381, opened 2016-01-08
silence heartbeat logging in cluster client


ktoso:
Silence heartbeat logging by default, rarely are people interested in them.
We want to keep it though, as it helps debugging heartbeat problems.

PR #19382, opened 2016-01-08
=clt silence heartbeat logging in cluster client

ktoso:
Resolves https://github.com/akka/akka/issues/19381

### DIFF
--- a/akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
+++ b/akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
@@ -569,6 +569,7 @@ final class ClusterReceptionist(pubSubMediator: ActorRef, settings: ClusterRecep
   import settings._
 
   val cluster = Cluster(context.system)
+  val verboseHeartbeat = cluster.settings.Debug.VerboseHeartbeatLogging
   import cluster.selfAddress
 
   require(role.forall(cluster.selfRoles.contains),
@@ -620,7 +621,7 @@ final class ClusterReceptionist(pubSubMediator: ActorRef, settings: ClusterRecep
       pubSubMediator.tell(msg, tunnel)
 
     case Heartbeat ⇒
-      log.debug("Heartbeat from client [{}]", sender().path)
+      if (verboseHeartbeat) log.debug("Heartbeat from client [{}]", sender().path)
       sender() ! HeartbeatRsp
 
     case GetContacts ⇒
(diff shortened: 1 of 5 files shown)


ktoso:
Please have a look @patriknw

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -470,7 +471,12 @@ object ClusterReceptionistSettings {
 final class ClusterReceptionistSettings(
   val role: Option[String],
   val numberOfContacts: Int,
-  val responseTunnelReceiveTimeout: FiniteDuration) extends NoSerializationVerificationNeeded {
+  val responseTunnelReceiveTimeout: FiniteDuration,
+  val verboseHeartbeatLogging: Boolean) extends NoSerializationVerificationNeeded {
+
+  def this(role: Option[String], numberOfContacts: Int, responseTunnelReceiveTimeout: FiniteDuration) {
```
ktoso: for bin compat only...

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -620,7 +627,7 @@ final class ClusterReceptionist(pubSubMediator: ActorRef, settings: ClusterRecep
       pubSubMediator.tell(msg, tunnel)
 
     case Heartbeat ⇒
-      log.debug("Heartbeat from client [{}]", sender().path)
+      if (verboseHeartbeatLogging) log.debug("Heartbeat from client [{}]", sender().path)
```
patriknw: In addition to this we should silence the real Cluster heartbeats also. The debug logging in this file: https://github.com/akka/akka/blob/master/akka-cluster/src/main/scala/akka/cluster/ClusterHeartbeat.scala

Therefore the configuration property should be located in akka-cluster/reference.conf, but I think we can use the same setting for the ClusterClient.

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
patriknw: When you add the config property to ClusterSettings I think it's not necessary to mess with the ClusterReceptionistSettings, instead you can use `Cluster(context.system).settings.VerboseHeartbeatLogging directly inside the ClusterReceptionist actor.

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
ktoso: Right, I was thinking if we want to silence both actually. Sounds good, I'll do that

ktoso:
Updated, all cluster heartbeats debug logs are not disabled by default. Config option is `akka.cluster.debug.verbose-heartbeat-logging`

johanandren:
LGTM!

patriknw:
LGTM

Issue #18577, opened 2015-09-28
Notification when ClusterClient can't reconnect


patriknw:
The ClusterClient should stop itself if the reconnect attempts with know contact points are not successful after a configured timeout. Then it can be watched to know that it failed and must be started again with new contact points.

The use case is that the server cluster is completely restarted and bound to new addresses. Then the ClusterClient will never be able to reconnect with the old contact points. Application might be able to lookup new contact points from some service discovery facility and start new ClusterClient with fresh information.

huntc:
Thanks for raising this - so just to clarify - there's presently no way of detecting that a cluster client cannot connect to its list of previous endpoints?

patriknw:
I don't see a nice workaround. The app can of course send some kind of heartbeats or ack messages and when those don't arrive it can interpret that as this connection issue.

PR #19384, opened 2016-01-08
=clu #18577 Stop cluster client after duration of no contact

johanandren:
When the cluster client has been in the "establishing" state for more than a configurable timeout it will log failure to reconnect and stop itself.

Refs #18577

### DIFF
--- a/akka-cluster-tools/src/main/resources/reference.conf
+++ b/akka-cluster-tools/src/main/resources/reference.conf
@@ -112,6 +112,15 @@ akka.cluster.client {
   # immediately if the location of the singleton is unknown.
   # Maximum allowed buffer size is 10000.
   buffer-size = 1000
+
+  # If connection to the receiptionist is lost and the client has not been
+  # able to acquire a new connection for this long the client will stop itself.
+  # This duration makes it possible to watch the cluster client and react on a more permanent
+  # loss of connection with the cluster, for example by accessing some kind of
+  # service registry for an updated set of initial contacts to start a new cluster client with.
+  # If this is not wanted it can be set to "off" to disable the timeout and retry
+  # forever.
+  reconnect-timeout = off
 }
 # //#cluster-client-config
 
--- a/akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
+++ b/akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
@@ -58,7 +58,11 @@ object ClusterClientSettings {
       refreshContactsInterval = config.getDuration("refresh-contacts-interval", MILLISECONDS).millis,
       heartbeatInterval = config.getDuration("heartbeat-interval", MILLISECONDS).millis,
       acceptableHeartbeatPause = config.getDuration("acceptable-heartbeat-pause", MILLISECONDS).millis,
-      bufferSize = config.getInt("buffer-size"))
+      bufferSize = config.getInt("buffer-size"),
+      reconnectTimeout = config.getString("reconnect-timeout") match {
+        case "off" ⇒ None
+        case _     ⇒ Some(config.getDuration("reconnect-timeout", MILLISECONDS).millis)
+      })
   }
 
   /**
@@ -96,6 +100,10 @@ object ClusterClientSettings {
  *   When the buffer is full old messages will be dropped when new messages are sent via the
  *   client. Use 0 to disable buffering, i.e. messages will be dropped immediately if the
  *   location of the receptionist is unavailable.
+ * @param reconnectTimeout If the connection to the receptionist is lost and cannot
+ *   be re-established within this duration the cluster client will be stopped. This makes it possible
+ *   to watch it from another actor and possibly acquire a new list of initialContacts from some
+ *   external service registry
  */
 final class ClusterClientSettings(
   val initialContacts: Set[ActorPath],
@@ -103,9 +111,23 @@ final class ClusterClientSettings(
   val refreshContactsInterval: FiniteDuration,
   val heartbeatInterval: FiniteDuration,
   val acceptableHeartbeatPause: FiniteDuration,
-  val bufferSize: Int) extends NoSerializationVerificationNeeded {
+  val bufferSize: Int,
+  val reconnectTimeout: Option[FiniteDuration]) extends NoSerializationVerificationNeeded {
 
   require(bufferSize >= 0 && bufferSize <= 10000, "bufferSize must be >= 0 and <= 10000")
+
+  /**
+   * For binary/source compatibility
+   */
+  def this(
+    initialContacts: Set[ActorPath],
+    establishingGetContactsInterval: FiniteDuration,
+    refreshContactsInterval: FiniteDuration,
+    heartbeatInterval: FiniteDuration,
+    acceptableHeartbeatPause: FiniteDuration,
+    bufferSize: Int) =
+    this(initialContacts, establishingGetContactsInterval, refreshContactsInterval, heartbeatInterval,
+      acceptableHeartbeatPause, bufferSize, None)
 
   /**
    * Scala API
@@ -135,15 +157,19 @@ final class ClusterClientSettings(
   def withBufferSize(bufferSize: Int): ClusterClientSettings =
     copy(bufferSize = bufferSize)
 
+  def withReconnectTimeout(reconnectTimeout: Option[FiniteDuration]): ClusterClientSettings =
+    copy(reconnectTimeout = reconnectTimeout)
+
   private def copy(
     initialContacts: Set[ActorPath] = initialContacts,
     establishingGetContactsInterval: FiniteDuration = establishingGetContactsInterval,
     refreshContactsInterval: FiniteDuration = refreshContactsInterval,
     heartbeatInterval: FiniteDuration = heartbeatInterval,
     acceptableHeartbeatPause: FiniteDuration = acceptableHeartbeatPause,
-    bufferSize: Int = bufferSize): ClusterClientSettings =
+    bufferSize: Int = bufferSize,
+    reconnectTimeout: Option[FiniteDuration] = reconnectTimeout): ClusterClientSettings =
     new ClusterClientSettings(initialContacts, establishingGetContactsInterval, refreshContactsInterval,
-      heartbeatInterval, acceptableHeartbeatPause, bufferSize)
+      heartbeatInterval, acceptableHeartbeatPause, bufferSize, reconnectTimeout)
 }
 
 object ClusterClient {
@@ -172,6 +198,7 @@ object ClusterClient {
   private[akka] object Internal {
     case object RefreshContactsTick
     case object HeartbeatTick
+    case object ReconnectTimeout
   }
 }
 
@@ -257,27 +284,37 @@ final class ClusterClient(settings: ClusterClientSettings) extends Actor with Ac
   def receive = establishing
 
   def establishing: Actor.Receive = {
-    case Contacts(contactPoints) ⇒
-      if (contactPoints.nonEmpty) {
-        contacts = contactPoints.map(context.actorSelection)
-        contacts foreach { _ ! Identify(None) }
-      }
-    case ActorIdentity(_, Some(receptionist)) ⇒
-      log.info("Connected to [{}]", receptionist.path)
-      scheduleRefreshContactsTick(refreshContactsInterval)
-      sendBuffered(receptionist)
-      context.become(active(receptionist))
-      failureDetector.heartbeat()
-    case ActorIdentity(_, None) ⇒ // ok, use another instead
-    case HeartbeatTick ⇒
-      failureDetector.heartbeat()
-    case RefreshContactsTick ⇒ sendGetContacts()
-    case Send(path, msg, localAffinity) ⇒
-      buffer(DistributedPubSubMediator.Send(path, msg, localAffinity))
-    case SendToAll(path, msg) ⇒
-      buffer(DistributedPubSubMediator.SendToAll(path, msg))
-    case Publish(topic, msg) ⇒
-      buffer(DistributedPubSubMediator.Publish(topic, msg))
+    val connectTimerCancelable = settings.reconnectTimeout.map { timeout ⇒
+      context.system.scheduler.scheduleOnce(timeout, self, ReconnectTimeout)
+    }
+
+    {
+      case Contacts(contactPoints) ⇒
+        if (contactPoints.nonEmpty) {
+          contacts = contactPoints.map(context.actorSelection)
+          contacts foreach { _ ! Identify(None) }
+        }
+      case ActorIdentity(_, Some(receptionist)) ⇒
+        log.info("Connected to [{}]", receptionist.path)
+        scheduleRefreshContactsTick(refreshContactsInterval)
+        sendBuffered(receptionist)
+        context.become(active(receptionist))
+        connectTimerCancelable.foreach(_.cancel())
+        failureDetector.heartbeat()
+      case ActorIdentity(_, None) ⇒ // ok, use another instead
+      case HeartbeatTick ⇒
+        failureDetector.heartbeat()
+      case RefreshContactsTick ⇒ sendGetContacts()
+      case Send(path, msg, localAffinity) ⇒
+        buffer(DistributedPubSubMediator.Send(path, msg, localAffinity))
+      case SendToAll(path, msg) ⇒
+        buffer(DistributedPubSubMediator.SendToAll(path, msg))
+      case Publish(topic, msg) ⇒
+        buffer(DistributedPubSubMediator.Publish(topic, msg))
+      case ReconnectTimeout ⇒
+        log.warning("Receptionist reconnect not successful within {} stopping cluster client", settings.reconnectTimeout)
+        context.stop(self)
+    }
   }
 
   def active(receptionist: ActorRef): Actor.Receive = {
--- a/akka-cluster-tools/src/multi-jvm/scala/akka/cluster/client/ClusterClientStopSpec.scala
+++ b/akka-cluster-tools/src/multi-jvm/scala/akka/cluster/client/ClusterClientStopSpec.scala
@@ -1,0 +1,111 @@
+/**
+ * Copyright (C) 2009-2016 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.cluster.client
+
+import akka.actor.{ Actor, Props, Terminated }
+import akka.cluster.Cluster
+import akka.cluster.pubsub.{ DistributedPubSub, DistributedPubSubMediator }
+import akka.remote.testconductor.RoleName
+import akka.remote.testkit.{ STMultiNodeSpec, MultiNodeSpec, MultiNodeConfig }
+import akka.testkit.{ EventFilter, ImplicitSender }
+import com.typesafe.config.ConfigFactory
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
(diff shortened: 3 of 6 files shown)


akka-cluster-tools/src/main/resources/reference.conf
```diff
@@ -111,6 +111,13 @@ akka.cluster.client {
   # immediately if the location of the singleton is unknown.
   # Maximum allowed buffer size is 10000.
   buffer-size = 1000
+
+  # If the connection has lost connection to the receiptionist and has not been
+  # able to acquire a new connection for this long the client will stop itself.
+  # This makes it possible to watch the cluster client and react on a more permanent
+  # loss of connection with the cluster, for example by accessing some kind of
+  # service registry for an updated set of initial contacts to start a new cluster client with.
+  reconnect-timeout = 3m
```
johanandren: This is totally arbitrary, what is a good timeout default?

johanandren:
Ping @ktoso @2m @patriknw

akka-cluster-tools/src/main/resources/reference.conf
ktoso: "If the connection has lost connection to"? :)

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -58,7 +58,8 @@ object ClusterClientSettings {
       refreshContactsInterval = config.getDuration("refresh-contacts-interval", MILLISECONDS).millis,
       heartbeatInterval = config.getDuration("heartbeat-interval", MILLISECONDS).millis,
       acceptableHeartbeatPause = config.getDuration("acceptable-heartbeat-pause", MILLISECONDS).millis,
-      bufferSize = config.getInt("buffer-size"))
+      bufferSize = config.getInt("buffer-size"),
+      reconnectTimeout = config.getDuration("reconnect-timeout", MILLISECONDS).millis)
   }
```
ktoso: will fail binary compat check, won't it?

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -257,27 +267,35 @@ final class ClusterClient(settings: ClusterClientSettings) extends Actor with Ac
...
+        failureDetector.heartbeat()
+      case RefreshContactsTick ⇒ sendGetContacts()
+      case Send(path, msg, localAffinity) ⇒
+        buffer(DistributedPubSubMediator.Send(path, msg, localAffinity))
+      case SendToAll(path, msg) ⇒
+        buffer(DistributedPubSubMediator.SendToAll(path, msg))
+      case Publish(topic, msg) ⇒
+        buffer(DistributedPubSubMediator.Publish(topic, msg))
+      case ReconnectFailureTick ⇒
+        log.info(s"Receptionist reconnect not successful within ${settings.reconnectTimeout.toSeconds}s stopping cluster client")
```
ktoso: sounds like WARN though?

ktoso: Please use `warn("blabla {}", theThing)` syntax, it avoids string concatenation if not log level is disabled.

johanandren: Was my first thought as well, but then I looked at other "connection loss" in this class and they use info. Although, I guess they aren't as severe. Will change.

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -257,27 +267,35 @@ final class ClusterClient(settings: ClusterClientSettings) extends Actor with Ac
...
+        failureDetector.heartbeat()
+      case RefreshContactsTick ⇒ sendGetContacts()
+      case Send(path, msg, localAffinity) ⇒
+        buffer(DistributedPubSubMediator.Send(path, msg, localAffinity))
+      case SendToAll(path, msg) ⇒
+        buffer(DistributedPubSubMediator.SendToAll(path, msg))
+      case Publish(topic, msg) ⇒
+        buffer(DistributedPubSubMediator.Publish(topic, msg))
+      case ReconnectFailureTick ⇒
+        log.warning("Receptionist reconnect not successful within {}s stopping cluster client", settings.reconnectTimeout.toSeconds)
```
ktoso: use `settings.reconnectTimeout` and no need for the `s` in the string then. It formats properly.

ktoso: stop the cancelable here – helps avoiding dead letters

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -257,27 +267,35 @@ final class ClusterClient(settings: ClusterClientSettings) extends Actor with Ac
...
-    case HeartbeatTick ⇒
-      failureDetector.heartbeat()
-    case RefreshContactsTick ⇒ sendGetContacts()
-    case Send(path, msg, localAffinity) ⇒
-      buffer(DistributedPubSubMediator.Send(path, msg, localAffinity))
-    case SendToAll(path, msg) ⇒
-      buffer(DistributedPubSubMediator.SendToAll(path, msg))
-    case Publish(topic, msg) ⇒
-      buffer(DistributedPubSubMediator.Publish(topic, msg))
+    val cancelable = context.system.scheduler.scheduleOnce(settings.reconnectTimeout, self, ReconnectFailureTick)
```
ktoso: rename to reconnectTimerCancelable?

akka-cluster-tools/src/main/resources/reference.conf
ktoso: Maybe 1m? Also rather arbitrary number though

akka-cluster-tools/src/multi-jvm/scala/akka/cluster/client/ClusterClientStopSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+
+        watch(client)
+
+        expectMsgType[Terminated](10.seconds)
+        EventFilter.warning(start = "Receptionist reconnect not successful within", occurrences = 1)
+
+      }
+
+      runOn(first, second) {
+        enterBarrier("was-in-contact")
```
ktoso: different barrier name?

akka-cluster-tools/src/multi-jvm/scala/akka/cluster/client/ClusterClientStopSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+      }
+
+      enterBarrier("cluster-started")
+    }
+
+    "stop if re-establish fails for too long time" in within(20.seconds) {
+      runOn(client) {
+        val client = system.actorOf(ClusterClient.props(
+          ClusterClientSettings(system).withInitialContacts(initialContacts)), "client1")
+        client ! ClusterClient.Send("/user/testService", "hello", localAffinity = true)
```
ktoso: don't shadow the `client` role

akka-cluster-tools/src/multi-jvm/scala/akka/cluster/client/ClusterClientStopSpec.scala
ktoso: OH WAIT, now I understand the test :)

ktoso:
Make MiMa happy and LGTM :0

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
johanandren: Its a `scheduleOnce` so when we got here it will not happen again?

akka-cluster-tools/src/multi-jvm/scala/akka/cluster/client/ClusterClientStopSpec.scala
johanandren: Oups. Will fix.

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
ktoso: Right thanks :)

akka-cluster-tools/src/main/resources/reference.conf
```diff
@@ -111,6 +111,13 @@ akka.cluster.client {
   # immediately if the location of the singleton is unknown.
   # Maximum allowed buffer size is 10000.
   buffer-size = 1000
+
+  # If connection to the receiptionist is lost and the client has not been
+  # able to acquire a new connection for this long the client will stop itself.
+  # This makes it possible to watch the cluster client and react on a more permanent
+  # loss of connection with the cluster, for example by accessing some kind of
+  # service registry for an updated set of initial contacts to start a new cluster client with.
+  reconnect-timeout = 1m
```
patriknw: 60s to be consistent with the other settings here

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -172,6 +194,7 @@ object ClusterClient {
   private[akka] object Internal {
     case object RefreshContactsTick
     case object HeartbeatTick
+    case object ReconnectFailureTick
```
patriknw: it's not really a tick (periodic thing), perhaps `ReconnectFailureTimeout`?

akka-cluster-tools/src/multi-jvm/scala/akka/cluster/client/ClusterClientStopSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+      runOn(client) {
+        val c = system.actorOf(ClusterClient.props(
+          ClusterClientSettings(system).withInitialContacts(initialContacts)), "client1")
+        c ! ClusterClient.Send("/user/testService", "hello", localAffinity = true)
+        expectMsgType[String](3.seconds) should be("hello")
+        enterBarrier("was-in-contact")
+
+        watch(c)
+
+        expectMsgType[Terminated](10.seconds)
```
patriknw: `expectTerminated(c, 10.seconds)`

patriknw:
Please add something to the documentation also (including a mention of the use case as described in the issue).

patriknw:
Is this change of behavior alright to deliver in a minor version, or do we have to use default values that "disable" the feature? Personally I think it is alright, but someone less pragmatic might object.

akka-cluster-tools/src/main/resources/reference.conf
hepin1989: the default should be `off` I think.and have a callback like `whenReconnectTimeOut(...)` to be called before it is stop should be nicer.

johanandren: I can accept ``off`` by default, but why is a custom specific callback better than watching the actor and reacting on ``Terminate``, which is entirely generic and requires no extra API?

hepin1989: @johanandren If you have a monitor service,then you could update the status,that some error occur.

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
hepin1989: this should be a Timeout,yes.

akka-cluster-tools/src/main/resources/reference.conf
patriknw: Callbacks are a mess, users will get the threading wrong. If we would need something more than the stop signal it whould be a message. This is actors after all. In my opinion we don't need anything more than the stop signal for the use case presented here.

ktoso: Agree with @patriknw here. Watching and "oh it terminated" is the actor way here ;)

patriknw: let's be conservative and use `off` as default

hepin1989: Actor and Message everywhere

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -257,27 +284,37 @@ final class ClusterClient(settings: ClusterClientSettings) extends Actor with Ac
...
+      case RefreshContactsTick ⇒ sendGetContacts()
+      case Send(path, msg, localAffinity) ⇒
+        buffer(DistributedPubSubMediator.Send(path, msg, localAffinity))
+      case SendToAll(path, msg) ⇒
+        buffer(DistributedPubSubMediator.SendToAll(path, msg))
+      case Publish(topic, msg) ⇒
+        buffer(DistributedPubSubMediator.Publish(topic, msg))
+      case ReconnectTimeout ⇒
+        log.warning("Receptionist reconnect not successful within {} stopping cluster client", settings.reconnectTimeout)
+        context.stop(self)
```
hepin1989: how about publish an Event about this here?

akka-cluster-tools/src/main/resources/reference.conf
```diff
@@ -112,6 +112,13 @@ akka.cluster.client {
   # immediately if the location of the singleton is unknown.
   # Maximum allowed buffer size is 10000.
   buffer-size = 1000
+
+  # If connection to the receiptionist is lost and the client has not been
+  # able to acquire a new connection for this long the client will stop itself.
+  # This makes it possible to watch the cluster client and react on a more permanent
+  # loss of connection with the cluster, for example by accessing some kind of
+  # service registry for an updated set of initial contacts to start a new cluster client with.
+  reconnect-timeout = off
```
patriknw: The description should mention that this is a duration (e.g. 60s)

akka-docs/rst/java/cluster-client.rst
```diff
@@ -146,3 +146,24 @@ a parameter to the ``ClusterClient.props`` factory method, i.e. each client can
...
+if more than a configurable amount of heartbeats are missed the client will try to reconnect to its known
+set of contacts to find a receptionist it can access.
+
+** When the cluster cannot be reached at all **
+It is possible to make the cluster client stop entirely if it cannot find a receptionist it can talk to
+within a configurable interval. This is configured with the ``reconnect-timeout``, which defaults to ``off``.
+This can be useful when initial contacts are provided from some kind of service registry, cluster node addresses
+are entirely dynamic and the entire cluster might shut down or crash, be restarted on new addresses. Since the
+client will be stopped in that case a monitoring actor can watch it and upon ``Terminate`` a new set of initial
+contacts can be fetched and a new cluster client started.
```
patriknw: good

patriknw:
perhaps clarify the reference.conf description, but otherwise LGTM

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -58,7 +58,11 @@ object ClusterClientSettings {
       refreshContactsInterval = config.getDuration("refresh-contacts-interval", MILLISECONDS).millis,
       heartbeatInterval = config.getDuration("heartbeat-interval", MILLISECONDS).millis,
       acceptableHeartbeatPause = config.getDuration("acceptable-heartbeat-pause", MILLISECONDS).millis,
-      bufferSize = config.getInt("buffer-size"))
+      bufferSize = config.getInt("buffer-size"),
+      reconnectTimeout = config.getString("reconnect-timeout") match {
+        case "off" ⇒ None
```
rkuhn: Another possibility would be to use Duration.Undefined.

akka-docs/rst/java/cluster-client.rst
```diff
@@ -146,3 +146,24 @@ a parameter to the ``ClusterClient.props`` factory method, i.e. each client can
...
+by ``establishing-get-contacts-interval``) try to contact those until it gets in contact with one of them.
+While running, the list of contacts are continuously updated with data from the receptionists (again, with an
+interval configurable with ``refresh-contacts-interval``), so that if there are more receptionists in the cluster
+than the initial contacts provided to the client the client will learn about them.
+
+While the client is running it will detect failures in its connection to the receptionist by heartbeats
+if more than a configurable amount of heartbeats are missed the client will try to reconnect to its known
+set of contacts to find a receptionist it can access.
+
+** When the cluster cannot be reached at all **
```
rkuhn: Please make this a heading (of the appropriate level).

akka-persistence-query/src/test/scala/akka/persistence/query/journal/leveldb/EventsByPersistenceIdSpec.scala
```diff
@@ -120,7 +120,7 @@ class EventsByPersistenceIdSpec extends AkkaSpec(EventsByPersistenceIdSpec.confi
       expectMsg(s"${2L}-deleted")
 
       val src = queries.currentEventsByPersistenceId("h", 0L, Long.MaxValue)
-      src.map(_.event).runWith(TestSink.probe[Any]).request(1).expectNext("h-3") expectComplete()
+      src.map(_.event).runWith(TestSink.probe[Any]).request(1).expectNext("h-3") expectComplete ()
```
rkuhn: I'd rather add the dot ;-)

rkuhn:
LGTM

akka-cluster-tools/src/main/resources/reference.conf
drewhk: More generally, it should mention accepted values: duration and `off`

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
```diff
@@ -257,27 +284,37 @@ final class ClusterClient(settings: ClusterClientSettings) extends Actor with Ac
...
-      buffer(DistributedPubSubMediator.Send(path, msg, localAffinity))
-    case SendToAll(path, msg) ⇒
-      buffer(DistributedPubSubMediator.SendToAll(path, msg))
-    case Publish(topic, msg) ⇒
-      buffer(DistributedPubSubMediator.Publish(topic, msg))
+    val connectTimerCancelable = settings.reconnectTimeout.map { timeout ⇒
+      context.system.scheduler.scheduleOnce(timeout, self, ReconnectTimeout)
+    }
+
+    {
```
drewhk: hm, not a big fan of these...

drewhk:
LGTM

akka-persistence-query/src/test/scala/akka/persistence/query/journal/leveldb/EventsByPersistenceIdSpec.scala
johanandren: Wasn't me, was the auto-code-formatter-thingie. But I'll undo.

akka-cluster-tools/src/main/scala/akka/cluster/client/ClusterClient.scala
johanandren: Thought about it for a bit, I like that it is clear that it is disabled rather than not defined - there is _no_ timeout. If you feel it is important I'll change it.

rkuhn: I was just talking about the internal representation, `reconnectTimeout.isDefined` would then be `reconnectTimeout.isFinite` instead, but this is not an important point

PR #19399, opened 2016-01-11
=pro check binary compatibility

2m:


project/AkkaBuild.scala
```diff
@@ -1332,12 +1329,20 @@ object AkkaBuild extends Build {
       ProblemFilters.exclude[MissingTypesProblem]("akka.remote.ReliableDeliverySupervisor$GotUid$"),
       ProblemFilters.exclude[MissingMethodProblem]("akka.remote.ReliableDeliverySupervisor#GotUid.apply")
     )
+
+    val akkaStream = Seq(
```
drewhk: Hmm. These will be forward ported hence the current exclude right?

project/AkkaBuild.scala
```diff
@@ -41,6 +41,7 @@ object AkkaBuild extends Build {
   val requestedScalaVersion = System.getProperty("akka.scalaVersion", "2.10.5")
   val Seq(scalaEpoch, scalaMajor) = """(\d+)\.(\d+)\..*""".r.unapplySeq(requestedScalaVersion).get.map(_.toInt)
   val streamAndHttpVersion = "2.0.1"
+  val streamAndHttpBinCompVersion = "2.0"
```
drewhk: Probably does not matter, but shouldn't 2.0.1 be the baseline (given that 2.0 was flawed)?

drewhk:
LGTM apart from nitpicks

project/AkkaBuild.scala
rkuhn: Yes, I agree, and then the list of exceptions should be empty to start with.

project/AkkaBuild.scala
2m: These are incompatibilities between `2.0` and current `stream-http-2.0` (currently same as `2.0.1`) therefore these incompatibilities need to be filtered out.

project/AkkaBuild.scala
2m: I do not think this matters, as 2.0 was not flawed regarding API.

2m:
Changed to `2.0.1`. Failure was:

```
WebsocketServerSpec
[info]   - for echoing user handler *** FAILED *** (75 milliseconds)
[info]     java.lang.AssertionError: expected OnNext, found OnError(java.lang.IllegalArgumentException: requirement failed: Cannot pull port twice
```

ktoso:
The pulling port twice issue is tracked here: https://github.com/akka/akka/issues/19094

ktoso:
LGTM

PR #19400, opened 2016-01-11
Akka Streams and HTTP => master

ktoso:
Also known as: Return of the Jedi.

This is the result of:

```
git filter-branch --index-filter 'git rm --cached -qr -- . && git reset -q $GIT_COMMIT -- akka-http akka-http-core akka-http-marshallers-java akka-http-marshallers-scala akka-http-parent akka-http-testkit akka-http-tests akka-http-tests-java8 akka-stream akka-stream-and-http akka-stream-tck akka-stream-testkit akka-stream-tests' --prune-empty -- --all
```

This produced a repo with only the above listed directories.

Next I `rebased` it onto `master`. Almost no conflicts happened, as the only directories we have here, do not exist on the master branch – with the notable exception of `akka-stream/src/main/scala/akka/persistence/stream/PersistentPublisher.scala` and friends.

Then I merged it into `master`. Then I branched off to `master-with-streams` from the master with merged `release-2.3-omg` branch, which includes all streams.

And that's this PR.

I'll be validating if everything is OK, but letting PR validation do the same already.


ktoso:
`akka-docs-dev` was missing. Have to re run.

PR #19401, opened 2016-01-11
Remove old akka-stream subproject, replacign with real one

ktoso:
Makes life slightly easier for the streams => master move.


ktoso:
Objections should be filed in paper, and will be looked at within 2 working weeks.

// joking. This should not be harmful to anyone, was mostly an experimentl.

ktoso:
FYI @rkuhn @patriknw

rkuhn:
Yes, indeed.

PR #19403, opened 2016-01-11
Akka Streams and HTTP => master

ktoso:
This brings all the directories with commit history from `release-2.3-dev`, including: akka-http akka-http-core akka-http-marshallers-java akka-http-marshallers-scala akka-http-parent akka-http-testkit akka-http-tests akka-http-tests-java8 akka-stream akka-stream-and-http akka-stream-tck akka-stream-testkit akka-stream-tests akka-docs-dev akka-parsing akka-bench-jmh-dev

This one is really correct I believe, so last image pun for the time being:

p1829_p_v7_aa_jpg__480x720_


danielwegener:
:clap: :+1:

drewhk:
Welcome home streams, welcome home...

rkuhn:
The years of wandering are over …

hepin1989:
:clap: ,finally!

reid-spencer:
Nice to see this. Way to go Akka team.

jrudolph:
Yay! Finally in master!

I wonder, though, why you rewrote all history? There are lots of reasons why you wouldn't do it:
 * everyone expects two years of history to be immutable
 * `gitk --merge` is now useless for fixing conflicts in old and open branches against the new master
 * Are you sure that all rewritten commits still build in the same way as before? Otherwise, bisecting may be broken? (I checked and none of the old commits build any more because the build file is missing)
 * What about all the links in issues and PRs? They now all point to a dead branch. Also the backlinks from commits to PRs don't work any more.
 * Last and least: I liked c0ffee42 and def1a7e0 and am a little sad to see them expelled from history...

ktoso:
Our thinking was that we want to be able to use git blame and track where a line came from.
The other way proposed was to "copy and paste into master" which would have dropped all history.

Goal was: to cleanup/filter-branch: to avoid any conflicts (e.g. lots of in the build and files which were touched accidentally in wrong directories - these were such incidents).

I guess you are proposing a 3rd way, being "delete all other files from release-2.3-dev, commit, and then merge, to not have any conflicts" (there are occasional touches of wrong files on the -dev branch, so we end up with conflicts otherwise which we wanted to avoid).

c0ffee42598e6ec55df6fa91b929f0bee01a2e81 is c77f4e28f6057237549294be91fd6e8e9046973b but I understand your point about the links.

Technically we can try the 3rd way again now. Do we think we should?

jrudolph:
The third way would be to take the end result of what you have here and commit it in the merge commit from the streams branch into master.

See https://github.com/spray/akka/tree/merged-master

I did it roughly like that:

```
git checkout origin/master
# remove the rewritten history
git reset --hard HEAD~1
# now try to merge
git merge origin/stream-http-2.0
# woah, lots of conflicts, so clean working directory
rm *
# replace the working directory by the state of merged master
git checkout origin/master .
# add everything back
git add -A
# we have a lift-off
git commit
```

jrudolph:
What you get is this merge commit: https://github.com/akka/akka/commit/7a8613fb6965a81dbcef9582007aef242def995f which looks as if the merge is not yet complete? It seems you added some files in akka-docs-dev along the way and the build wasn't yet merged, right?

rkuhn:
Hmm, I’m not sure I understand all the points: all the old commits are still there under the tag https://github.com/akka/akka/tree/final_state_of_release-2.3-dev_before_move_to_master and all links that were reasonable should still work. If github decides to not display some links after the removal of a branch then that is not something we can fix, feature branches will be deleted when the feature is merged.

The solution we chose duplicates all history, one set is complete (and buildable), the other is only there to allow `git blame` to be used on the master branch. Merging release-2.3-dev into master would not have been desirable because it contains things that are not supposed to be on master—and also not in its history.

But I may of course be missing things.

jrudolph:
> because it contains things that are not supposed to be on master—and also not in its history.

Yes, the only reason to rewrite history is if you want to redefine history. I question that it is worthwhile to have duplicate histories and all the confusion that this will bring just to exclude something from the history of master (without knowing what you want to exclude part of the history or why). What do you gain by excluding things from master's history but keep it in the repository?

> all links that were reasonable should still work

No, you won't be able to blame a line in the new master and put it into github to find out which PR it belonged to. For these kind of things you will always have to do a manual search to find the corresponding commit in the real history.

ktoso:
I agree that I messed this up, I'm sorry. We can still straighten things up. I'm sad we didn't realise this sooner.

rkuhn:
I still don’t agree—what is the proposed solution in plain English, and why is it better?

rkuhn:
@jrudolph The only scenario that was brought up so far is of dubious utility: I never used that link (didn’t even know that it existed) and we have a convention in place that allows very straight-forward identification of the issue and thereby the PR.

2beaucoup:
Have to agree with @jrudolph here. It is already hard enough to investigate where things came from. And it's not just the PRs that point to old commits now. There are a lot of comments on commits too.

I guess this also has an impact on repo size.

But maybe it's not too late.

rkuhn:
So, to clarify, we did actually consider all options before doing this:

* My initial proposal was to just dump the current state of the directories in one commit onto master; this would lose all history for further development, `git blame` would be useless. But it is the simplest solution.
* I had previously discarded the possibility of merging `release-2.3-dev` into `master`, because no matter how it is done it would lead to the wrong history—a coupling of the release-2.3 and master branches that is not desirable and also not reflective of the true history.
* I had also discarded the possibility of rebasing `release-2.3-dev` onto `master`, which would have been the proper way of getting this “feature branch” in—it is what we always do. But not only would it take an absolutely unreasonable amount of effort, it would also lead to all commits having new hash values, so it also is not a magical silver bullet.
* So we ended up doing the first option, but with as much history as we can reasonably preserve—the use-case being `git blame` so that we can see the context of a given change.

As far as I can see, and taking into account everything that was raised so far, our solution supports all use-cases. All original and immutable history is preserved, and due to our commit message naming convention we also have an easy way of finding our way back from the copied history to its original, including all issue and PR discussions. The only thing I broke—very much deliberately—is the `release-2.3-dev` branch; feature branches are deleted once the feature is merged, and this one is no exception. There is [a tag](https://github.com/akka/akka/tree/final_state_of_release-2.3-dev_before_move_to_master) that points to the final state of the old branch.

rkuhn:
@2beaucoup no comments on commits were harmed during the operation, unless I am severely misunderstanding something here.

jrudolph:
I don't doubt that there are always manual or semi-manual ways to make the duplicate history work. I just don't understand why you would want to duplicate all commits. Do you want me to list all reasons why it is good practice not to rewrite history?

https://sethrobertson.github.io/GitBestPractices/#pubonce
http://tech.novapost.fr/git-history-matters-en.html

The proposed solution is to merge stream-http-2.0 into master. Regarding the practical approach I would start with a branch from stream-http-2.0 into which I would merge master first. Because this will create lots of conflicts at least in the build file, this initial merge doesn't need to be complete. You could then build up from this branch to get it to the same state as stream-http-2.0. Once you can actually use this branch to develop things, create a PR to merge it into master.

2beaucoup:
> @2beaucoup no comments on commits were harmed during the operation, unless I am severely misunderstanding something here.

They're not gone. Only hard to discover. I have to know that there's a duplicated commit on ``release-2.3-dev` which has them.

rkuhn:
@jrudolph Please read my post, it is not helpful to reiterate the same point without addressing the reason I give. And please stop suggesting that we were rewriting history, this is simply not true, nothing was altered.

jrudolph:
> I had previously discarded the possibility of merging release-2.3-dev into master, because no matter how it is done it would lead to the wrong history—a coupling of the release-2.3 and master branches that is not desirable and also not reflective of the true history.

It would be exactly reflective of the true history. master and release-2.3-dev were developed separately for a long time and in the end they were merged. I don't agree with the idea that one has to protect from "wrong history". Yes, there may be things deep in the history that are not yet any more part of the current tree. So what? You don't see them if you don't go looking for them. 

What people will find when they go back in the proposed history of master is a series of commits that don't represent any real state of the repository. A state that cannot even be built.

What I want is going back in the history of the master branch and find the actual states of development at that point in time. That's what the history is for.

(Wrt "rewriting": in the long view, only master is relevant and should contain the original commits. If it doesn't contain the original commits but some synthesized version of them, I call it rewriting the original history that led to that state.)

2beaucoup:
> And please stop suggesting that we were rewriting history, this is simply not true, nothing was altered.

But with the original history "hidden" in the branch while putting the duplicated history front and center in master newcomers will assume that that's where changes originally came from.

"Rewriting history" might technically not be correct but the real world impact might be the same.

rkuhn:
@jrudolph No, what you are missing is that master and release-2.3 had already diverged when we branched off release-2.3-dev.

Again, what we have done here is exactly what we do for every PR: we squash and rebase before we merge. This time we didn’t do the squash part because we wanted to keep more history, that is all. No alternatives have been presented that would be “more correct” in the view of the actual history of the code.

The only consequence I could draw from this discussion is that we should not have kept the history, but just lumped it into one commit as per my initial plan. But I am quite certain that that would have led to complaints as well, so we will stick with what we have right now. We have far more important things to worry about.

Concerning “rewriting history”: we do that every day. But once something has hit `master` it is sacrosanct. This has happened for Streams & HTTP this morning.

jrudolph:
I disagree that merging a branch that ran for two years should be treated the same way as a single PR. I know that the branch had diverged earlier but this is no reason not to include the original history into master. But I'll stop arguing here.

darionyaphet:
Stream & HTTP :+1:

jrudolph:
Here's a workaround for all others that want to keep the illusion of the one and only real history ;) : https://github.com/jrudolph/akka/blob/master-with-original-stream-history/README.md

rkuhn:
Nice, I didn’t know that this existed! Thanks for sharing, @jrudolph, with this everyone should be able to see the history they like.

jrudolph:
By which reasoning were commits selected that got in `stream-http-2.0`? There's now e.g. https://github.com/spray/akka/commit/c74149952b671d2fcf81b1c78a41e7f2d17ff8d0 which isn't contained in any branches any more?

Or, in other words how does final_state_of_release-2.3-dev_before_move_to_master differ from `stream-http-2.0`?

ktoso:
It is on master:

```
master u+7!akka $> branch --contains $(log --oneline --grep 'Tls-Session-Info header when configured on ' | awk '{print $1}' | head -n1)
* master
  wip-add-streams-to-build-ktoso
  wip-akka-scalariform-latest
  wip-mediatype-java-problem
```

`stream-http-2.0` was cut from `2.0.1` AFAIR.

And yes, that commit should be backported to `2.0.3`, last week was tough for me (in terms of private life) so I wasn't able to do that then.

rkuhn:
The “final…” tag is what was transported into master, `stream-http-2.0` does not include anything past the 2.0.1 release. We will backport some of the things, but not all of them, and right now we are concentrating on getting 2.4.2-RC1 out.

rkuhn:
further info: the 2.0.x releases should remain binary compatible (unless we find a good reason to break that)

jrudolph:
I see. Thanks for the clarification @ktoso and @rkuhn. No hurry from my side, I was just confused about where those commits went.

I based my merge to master on stream-http-2.0 so I had to update the replacement ref to update the merge to final_state_of_release-2.3-dev_before_move_to_master. Whoever has followed my instructions from above now needs to re-fetch the replacement reference (add `-f` to the fetch-command).

ktoso:
I prepped useful aliases for grepping which work for our usual workflow (that involves cherry-picking across branches for back/forward porting):

```
#$HOME/.gitconfig
	ffind-branches = "!f() { git branch --contains $(git ffind-commits $@ | awk '{print $1}' | head -n1); }; f"
	ffind-commits  = "!f() { git log --oneline --grep \"$*\"; }; f"
```

Issue #19395, opened 2016-01-11
Akka streams operation: dedupe


Horusiath:
This is the proposal of a new graph stage - reflected also in `FlowOps` DSL - which should filter out consecutive duplicates from the stream of events, working on similar rules as Rx [distinctUntilChanged](http://rxmarbles.com/#distinctUntilChanged) method works.

I'd like to contribute with this feature.

drewhk:
Sure, nice contribution! The only thing we need to decide whether this is provided as a Flow, or added to the DSL as a method. (We usually don't like to add new DSL methods to avoid a mess there. Flows are reusable units, they can be organized into packages, etc.)

viktorklang:
@Horusiath Thanks for picking up the ball here!

I found this in my git stash, which may serve as inspiration:

~~~scala
private[akka] final class Deduplicate[T](equiv: Equiv[T]) extends GraphStage[FlowShape[T, T]] {
  override val shape = FlowShape(Inlet[T]("deduplicate.in"), Outlet[T]("deduplicate.out"))
  val in = shape.in
  val out = shape.out
  override def initialAttributes: Attributes = DefaultAttributes.deduplicate
  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
    setHandler(in, new InHandler {
      private[this] var lastPushed: T = null.asInstanceOf[T]
      override def onPush(): Unit = {
        val last = lastPushed
        val newIn = grab(in)
        if (last == null || !equiv.equiv(last, newIn)) {
          push(out, newIn)
          lastPushed = newIn
        } else pull(in)
      }
    })

    setHandler(out, new OutHandler {
      override def onPull(): Unit = pull(in)
    })
  }
  override def toString = "Deduplicate"
}
~~~

rkuhn:
There was a lot of discussion in #19408, with the conclusion being that this kind of deduplication is too narrow a use-case to justify a combinator on FlowOps. More generic ones would be counting the running sequences of equivalent elements (compressed to `(Out, Long)`) and full hysteresis support.

2beaucoup:
Is this solved by `statefulMapConcat`?

drewhk:
yes. Also, if someone wants to make a ready-made version of this, the right place would be: https://github.com/akka/akka-stream-contrib

drewhk:
Added a ticket there instead.

cnnickolay:
An alternative solution with using sliding

```
[...]

PR #19405, opened 2016-01-12
#19395 : Introduced dedupe graph stage and flow ops method

Horusiath:
## Motivation

As mentioned in #19395 I've seen a missing, but somewhat useful feature of removing consecutive duplicates from the stream of events. This is not same as distinct/unique operation over the standard collections, since this would require potential buffering of the whole stream.

Example:

``` scala
Source(List(1, 1, 2, 2, 2, 1, 1, 1)).dedupe   // 1, 2, 1
```
## Changes

This PR introduces enhancement to akka Flow DSL in form of `dedupe` operation and correlated `Dedupe` graph stage. Correlated specs attached.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -7,6 +7,7 @@ import akka.event.Logging.LogLevel
 import akka.event.{ LogSource, Logging, LoggingAdapter }
 import akka.stream.Attributes.{ InputBuffer, LogLevels }
 import akka.stream.DelayOverflowStrategy.EmitEarly
+import akka.stream.impl.Stages.DefaultAttributes
 import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
 import akka.stream.impl.{ FixedSizeBuffer, BoundedBuffer, ReactiveStreamsCompliance }
 import akka.stream.stage._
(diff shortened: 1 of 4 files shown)


drewhk:
Please retarget branch to master as 2.3-dev has been just deleted.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -283,6 +284,39 @@ final case class Intersperse[T](start: Option[T], inject: T, end: Option[T]) ext
 /**
  * INTERNAL API
  */
+private[akka] final case class Dedupe[T]() extends GraphStage[FlowShape[T, T]] {
```
viktorklang: Please use either (T,T) => Boolean or a scala.math.Equiv[T] instance here so we don't have to rely on "universal equality"

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -283,6 +284,39 @@ final case class Intersperse[T](start: Option[T], inject: T, end: Option[T]) ext
...
+  private val in = Inlet[T]("Dedupe.in")
+  private val out = Outlet[T]("Dedupe.out")
+
+  override def shape: FlowShape[T, T] = FlowShape(in, out)
+  override def initialAttributes = DefaultAttributes.dedupe
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    var last: Any = null
+    val inHandler = new InHandler {
```
viktorklang: No need to increase the size of the GraphStageLogic class by having these as fields, use setHandler directly to the new InHandler/OutHandler.

PR #19408, opened 2016-01-12
#19395: added dedupe graph stage

Horusiath:
## Motivation

As mentioned in #19395 I've seen a missing, but somewhat useful feature of removing consecutive duplicates from the stream of events. This is not same as distinct/unique operation over the standard collections, since this would require potential buffering of the whole stream.

Example:

``` scala
Source(List(1, 1, 2, 2, 2, 1, 1, 1)).dedupe   // 1, 2, 1
```
## Changes

This PR introduces enhancement to akka Flow DSL in form of `dedupe` operation and correlated `Dedupe` graph stage. Correlated specs attached.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -281,6 +281,36 @@ final case class Intersperse[T](start: Option[T], inject: T, end: Option[T]) ext
 }
 
 /**
+  * INTERNAL API
+  */
+private[akka] final case class Deduplicate[T](implicit equiv: Equiv[T]) extends GraphStage[FlowShape[T, T]] {
+  private val in = Inlet[T]("Deduplicate.in")
+  private val out = Outlet[T]("Deduplicate.out")
+
+  override def shape: FlowShape[T, T] = FlowShape(in, out)
+  override def initialAttributes = DefaultAttributes.deduplicate
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    var last: T = null.asInstanceOf[T]
+
+    setHandler(in, new InHandler {
+      override def onPush(): Unit = {
+        val elem = grab(in)
+        if (equiv.equiv(elem, last)) {
+          last = elem
+          push(out, elem)
+        } else pull(in)
+      }
+    })
+    setHandler(out, new OutHandler {
+      override def onPull(): Unit = pull(in)
+    })
+  }
+
+  override def toString = "Deduplicate"
+}
+
+/**
  * INTERNAL API
  */
 private[akka] final case class Grouped[T](n: Int) extends PushPullStage[T, immutable.Seq[T]] {
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowDeduplicateSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowDeduplicateSpec.scala
@@ -1,0 +1,29 @@
+/**
+  * Copyright (C) 2016 Typesafe Inc. <http://www.typesafe.com>
+  */
+package akka.stream.scaladsl
+
+import akka.stream.testkit._
+import akka.stream.testkit.Utils.assertAllStagesStopped
+import akka.stream.testkit.scaladsl.{ TestSource, TestSink }
+import akka.stream.{ ActorMaterializer, ActorMaterializerSettings }
+
+class FlowDeduplicateSpec extends AkkaSpec {
+
+  implicit val materializer = ActorMaterializer()
+
+  "A Deduplicate" must {
+
+    "remove consecutive duplicates" in {
+      val input = List(1, 1, 1, 2, 2, 1, 1, 3)
+      val probe = TestSubscriber.manualProbe[Int]()
+      Source(input).deduplicate.runWith(Sink.fromSubscriber(probe))
+
+      val subscription = probe.expectSubscription()
+      subscription.request(input.size)
+
+      probe.expectNext(1, 2, 1, 3)
+      probe.expectComplete()
+    }
+  }
+}
--- a/akka-stream/src/main/scala/akka/stream/impl/Stages.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/Stages.scala
@@ -43,6 +43,7 @@ private[stream] object Stages {
     val scan = name("scan")
     val fold = name("fold")
     val intersperse = name("intersperse")
+    val deduplicate = name("deduplicate")
     val buffer = name("buffer")
     val conflate = name("conflate")
     val expand = name("expand")
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
@@ -1352,6 +1352,25 @@ trait FlowOps[+Out, +Mat] {
    * '''Cancels when''' downstream cancels
    */
   def initialDelay(delay: FiniteDuration): Repr[Out] = via(new Timers.DelayInitial[Out](delay))
+  /**
+   *  Filters our consecutive duplicated elements from the stream.
+   *
+   * Examples:
+   *
+   * {{{
+   *   var nums = Source(List(1,1,1,2,2,1,1))
+   *   nums.deduplicate   // 1, 2, 1
+   * }}}
+   *
+   * '''Emits when''' upstream emits element distinct from element emitted previously
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   */
+  def deduplicate(implicit equiv: Equiv[Out]): Repr[Out] = via(new Deduplicate(equiv))
 
   /**
    * Logs elements flowing through the stream as well as completion and erroring.


akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowDedupeSpec.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  "A Dedupe" must {
+
+    "remove consecutive duplicates" in {
+      val input = List(1, 1, 1, 2, 2, 1, 1, 3)
+      val probe = TestSubscriber.manualProbe[Int]()
+      Source(input).dedupe.runWith(Sink.fromSubscriber(probe))
+
+      val subscription = probe.expectSubscription()
+      subscription.request(Long.MaxValue)
```
viktorklang: `input.size` iso Long.MaxValue?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowDedupeSpec.scala
```diff
@@ -0,0 +1,32 @@
...
+
+      probe.expectNext(1)
```
viktorklang: replace these with: `input.distinct.foreach(probe.expectNext)` ?

Horusiath: I believe that `input.distinct` (which will be `1,2,3`) is not the same as dedupe (`1,2,1,3`). This is major difference between these two operations: from stream point of view dedupe prefers local uniqueness of elements instead of global one.

viktorklang: Ah, of course, sorry, I misremembered what the collection op is named. (and there doesn't seem to be an equivalent on scala.collection)

drewhk: Well, `dedupe` can be a misleading name since in many domains it means exactly what scala `distinct` means. Btw, no need to explicitly request, AFAIK there is a varargs version of `requestNext`.

viktorklang: I prefer `norepeat` (also is nicely contrasted with Source.repeat)

viktorklang: On further analysis I think it ought to be named `destutter`:

https://wpollock.com/lister.php?file=AJava/Destutter.java&nodir&linenums&dl
https://books.google.com/books?id=nabtAQAAQBAJ&pg=PA63&lpg=PA63&dq=%22destutter%22&source=bl&ots=yZp5XRI5lZ&sig=L7p4F1mxiJ6ot52MpOzhezo9K28&hl=en&sa=X&ved=0ahUKEwim1P-JoqTKAhWJtxoKHU-TAvEQ6AEILTAD#v=onepage&q=%22destutter%22&f=false
https://gist.github.com/invisiblefunnel/1248809

and more

drewhk: Also, a cusomizable equivalence function is needed.

drewhk: destutter sounds good.

ktoso: destutter sounds ok. dedupe pretty bad, so for that rename :+1:

2beaucoup: :+1: for `destutter`

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -281,6 +281,36 @@ final case class Intersperse[T](start: Option[T], inject: T, end: Option[T]) ext
...
+  override def shape: FlowShape[T, T] = FlowShape(in, out)
+  override def initialAttributes = DefaultAttributes.destutter
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    var last: T = null.asInstanceOf[T]
+
+    setHandler(in, new InHandler {
+      override def onPush(): Unit = {
+        val elem = grab(in)
+        if (ęquiv.equiv(elem, last)) {
```
2beaucoup: s/ęquiv/equiv

rkuhn:
Thanks for the contribution @Horusiath! I thought quite a bit about this and am still a bit sad that I have not yet been able to find a simple formulation that allows the user to write a more general hysteresis. A simple equivalence does not work because state would need to be kept as well.

In case we go with this less powerful combinator (because of general usefulness) then I’m still skeptical about the name, stuttering to my knowledge has a stronger connotation with a medical condition than a technical term.

rkuhn:
Thinking about the semantics of the current proposal I think that `.deduplicate` is the most correct description of the functionality.

Horusiath:
@rkuhn I'm feeling, that we've made a circle here ;) All names have been changed to deduplicate.

2beaucoup:
@rkuhn I had already [discussed](https://gitter.im/akka/dev?at=568ce8219a5f8fe839216979) naming of this with @viktorklang on gitter.

"deduplicate", "distinct", "unique", etc. normally work across the entire collection.

"noRepeat", "filterRepeated" are good alternatives but I think "destutter" is a good fit that (as @viktorklang [showed above](https://github.com/akka/akka/pull/19408#discussion_r49447437)) already has some usages. I understand that it's also used as a medical term but in a technical context I think it's  unambiguous.

rkuhn:
@2beaucoup I’ve seen the links to prior usage, but just as I dislike case law I also think that precedent alone is not a good justification—humans err. When an average English reader sees this term, they will undoubtedly think of the medical meaning first, even if some of them later realize that it has technical significance as well.

The counterpoint to “deduplicate” based on collections semantics is valid, it might have saved @Horusiath a round of renames to link to it :-) I quite like the name “debounce” because it has no other meaning, but unfortunately it describes a different (but similar) operation. Shall we settle for a boring but precise `removeRepetitions` then? This will not be a very common operation, and even a long name like this is still quite a bit shorter than implementing it with `scan` and `collect`.

viktorklang:
`distinctConsecutive` is another option

On Wed, Jan 13, 2016 at 9:18 AM, Roland Kuhn <notifications@github.com>
wrote:

> @2beaucoup <https://github.com/2beaucoup> I’ve seen the links to prior
> usage, but just as I dislike case law I also think that precedent alone is
> not a good justification—humans err. When an average English reader sees
> this term, they will undoubtedly think of the medical meaning first, even
> if some of them later realize that it has technical significance as well.
>
> The counterpoint to “deduplicate” based on collections semantics is valid,
> it might have saved @Horusiath <https://github.com/Horusiath> a round of
> renames to link to it :-) I quite like the name “debounce” because it has
> no other meaning, but unfortunately it describes a different (but similar)
> operation. Shall we settle for a boring but precise removeRepetitions
> then? This will not be a very common operation, and even a long name like
> this is still quite a bit shorter than implementing it with scan and
> collect.
>
> —
> <https://github.com/akka/akka/pull/19408#issuecomment-171212839>.
>



-- 
Cheers,
√

2beaucoup:
> When an average English reader sees this term, they will undoubtedly think of the medical meaning first, even if some of them later realize that it has technical significance as well.

Cause it's the perfect analogy. ;) Last thought from my side: What's the problem here in your opinion? That people could be offended?

hepin1989:
`destutter` +1 for that.

viktorklang:
If it's insensitive to use that name then we should most definitively not
use it.

`distinctConsecutive` is probably the most semantically accurate name
though (contrasted with `distinct` for full deduplication)

On Wed, Jan 13, 2016 at 9:41 AM, kerr <notifications@github.com> wrote:

> destutter +1 for that.
>
> —
> <https://github.com/akka/akka/pull/19408#issuecomment-171216694>.
>



-- 
Cheers,
√

hepin1989:
`distinctConsecutive` is a good name ,just a little longer.but with autocomplete ,that should not be a problem.but should not use `deduplicate` or `distinct` for this one.

drewhk:
> When an average English reader sees this term, they will undoubtedly think of the medical meaning first, even if some of them later realize that it has technical significance as well.

I don't believe this is true. When I look at code I definitely assume that it has technical meaning, how would I assume any other way? It feels unrealistic that when looking at code completion offers I would think about a medical condition. Technical and scientific practice is full of these mnemonics which *allude* to the actual meaning and not directly convey them because that would be too long or less catchy.

rkuhn:
Hmm, but talking about discoverability and intuitive recognition: “destutter” has some connotations that may guide a reader, but I posit that most of them will need to read the ScalaDoc in order to be sure of the precise semantics—another intuitive interpretation is to smooth out the timing between elements (this comes from the engineering meaning of the word, where it refers to a short transient failure of an engine). `distinctConsecutive` describes a property of the resulting stream, but in a somewhat backwards way, it takes more parsing steps to understand than `removeRepetitions` which unmistakably describes what the operation does. I don’t think it is a coincidence that most other operators use a verb in imperative form to describe their function, it is easier to grasp than a description of a property of the resulting stream.

If we go ahead with `destutter` I would like to have confirmation by some native English speakers that this evokes the right associations—I simply don’t trust a bunch of Germans, Swedes, Hungarians, … to get this right :-)

2beaucoup:
> I simply don’t trust a bunch of Germans, Swedes, Hungarians, … to get this right :-)

Haha! +1 :)

drewhk:
>  but I posit that most of them will need to read the ScalaDoc in order to be sure of the precise semantics

This is almost always the case. Apart from a few trivial examples there are really no self-describing methods names (just think about `materialize` for a few seconds). These names are mnemonics in most of the cases.

viktorklang:
My objection to 'removeRepetitions' is that it could mean collapsing
1,2,3,1,2,3 to 1,2,3

distinctConsecutive is more precise.

-- 
Cheers,
√

drewhk:
My issue is that `distinctConsecutive` might be more discoverable, but I bet that I would forget its name all the time and would need to find it again. `destutter` is much easier to remember for me, it is just so catchy.

drewhk:
Btw, that is one of the side effects of removing the `apply` overloads that now I don't remember the exact names that need to be invoked (even though they are quite logical). There are always tradeoffs.

viktorklang:
If 'destutter' is deemed safe to use I'd prefer it too.

-- 
Cheers,
√

rkuhn:
There was another question I raised and which so far has not seen any response; it might also have impact on the naming discussion. The operation we are discussing is a very simple form of hysteresis. What if we had a combinator that allowed this to be expressed in general?
~~~scala
FlowOps {
  ...
  def statefulMapConcat[T](f: () => Out => Iterable[T]): Repr[T]
}

Flow[T].statefulMapConcat(() => {
  var last: T = null
  elem =>
    if (last eq null || !equiv.equiv(elem, last)) { last = elem; List(elem) }
    else Nil
})

// which could be provided in the DSL package as
Flow[T].statefulMapConcat(destutter)
~~~

drewhk:
I expect destutter to be used for logging, etc, where you might want to add as little overhead as possible, so the List version is a bit heavyweight for my tastes. Although that is a nice combinator, but I wouldn't use it for this.

There is another generalization possible, which keeps a FIFO of recently seen elements and omits those new ones that are in the FIFO. In that world `destutter` is a FIFO of size 1. I would still keep this variant as a separat implementation, it is a very useful combinator.

rkuhn:
For logging I would want to keep track of how many events were collapsed into one, which means that the currently discussed operator is not the right answer. That was another line along which I started thinking, that the more generic operator for this purpose would emit tuples of `(Out, Int)`. My feeling is that the currently discussed `destutter` may be too specialized to be offered on FlowOps, but I’m open to arguments.

drewhk:
Hm, true, just unconditionally dropping, without allowing to either inject elements ("duplicates have been omitted") or modifying elements it is not as useful.

viktorklang:
@rkuhn

Why stop there?

~~~scala
def godMode[Z, T](z: Z) f: (Z, Out) => (Z, Iterable[T])): Repr[T]
~~~

And now you can represent virtually any stream combinator...

rkuhn:
@viktorklang what I propose enables you to write that wasteful combinator yourself :-)

viktorklang:
@rkuhn But we already have that, it's called GraphStage :)

You could also view `distinctConsecutive` as a `zipWithPrevious: Repr[(Out,Out)]` and do: `flow.zipWithPrevious.filter(_ != _)`

viktorklang:
@rkuhn I prefer having the semantic operation, because then we can optimize it internally, having to lift things into structure means allocation and less opportunity for optimization.

rkuhn:
I prefer not having a too narrow operator—is this operation used often enough to justify an AST node?

Or should this be replaced by `countConsecutiveRepeatedElements`?

viktorklang:
@rkuhn Speaking of which, I think we need to get on top of the AST nodes, it seems like a lot of ops are simply GraphStages without any AST nodes.

drewhk:
well, another variant could also make sense, were the deduplication emits the *last* element in the equivalent range (which might be not equivalent in some other criteria). I think the current approach is just too limiting, let's do something more flexible.

rkuhn:
I transported this conclusion to the ticket, closing this PR for now. Thanks for inspiring the discussion @Horusiath !

PR #19411, opened 2016-01-12
str #19338: NPE in ActorGraphInterpreter due to race with publisher shutdown and subscription

### LINKED ISSUE #19338: NPE from ActorGraphInterpreter
This is probably the fault of my code, but the exception doesn't leave me any wiser as to what the problem could be:

```
java.lang.NullPointerException: null
	at akka.stream.impl.fusing.ActorGraphInterpreter$ActorOutputBoundary.subscribePending(ActorGraphInterpreter.scala:262) ~[akka-stream-experimental_2.11-2.0.1.jar:na]
	at akka.stream.impl.fusing.GraphInterpreterShell.receive(ActorGraphInterpreter.scala:421) ~[akka-stream-experimental_2.11-2.0.1.jar:na]
	at akka.stream.impl.fusing.ActorGraphInterpreter$$anonfun$receive$1.applyOrElse(ActorGraphInterpreter.scala:547) ~[akka-stream-experimental_2.11-2.0.1.jar:na]
	at akka.actor.Actor$class.aroundReceive(Actor.scala:480) ~[akka-actor_2.11-2.4.1.jar:na]
	at akka.stream.impl.fusing.ActorGraphInterpreter.aroundReceive(ActorGraphInterpreter.scala:495) ~[akka-stream-experimental_2.11-2.0.1.jar:na]
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:526) [akka-actor_2.11-2.4.1.jar:na]
	at akka.actor.ActorCell.invoke(ActorCell.scala:495) [akka-actor_2.11-2.4.1.jar:na]
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:257) [akka-actor_2.11-2.4.1.jar:na]
	at akka.dispatch.Mailbox.run(Mailbox.scala:224) [akka-actor_2.11-2.4.1.jar:na]
	at akka.dispatch.Mailbox.exec(Mailbox.scala:234) [akka-actor_2.11-2.4.1.jar:na]
```

drewhk:
Fixes #19338

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/ActorGraphInterpreterSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/ActorGraphInterpreterSpec.scala
@@ -3,11 +3,13 @@
  */
 package akka.stream.impl.fusing
 
+import java.util.concurrent.CountDownLatch
+
 import akka.stream._
 import akka.stream.scaladsl._
-import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
-import akka.stream.testkit.AkkaSpec
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
 import akka.stream.testkit.Utils._
+import akka.stream.testkit.{AkkaSpec, TestPublisher, TestSubscriber}
 import akka.testkit.EventFilter
 
 import scala.concurrent.Await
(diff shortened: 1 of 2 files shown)


akka-stream-tests/src/test/scala/akka/stream/impl/fusing/ActorGraphInterpreterSpec.scala
```diff
@@ -256,5 +258,65 @@ class ActorGraphInterpreterSpec extends AkkaSpec {
...
+
+          override def preStart(): Unit = {
+            latch.await()
+            fail(shape.out, te)
+          }
+
+          override def postStop(): Unit = testActor ! "I am dead"
+
+          setHandler(shape.out, GraphStageLogic.IgnoreTerminateOutput) //We fail in preStart anyway
+          setHandler(shape.in, GraphStageLogic.EagerTerminateInput)
```
rkuhn: why not use `ignoreTerminateOutput` etc.?

drewhk: because I forgot they are there :)

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/ActorGraphInterpreterSpec.scala
```diff
@@ -256,5 +258,65 @@ class ActorGraphInterpreterSpec extends AkkaSpec {
...
+
+      val downstream = TestSubscriber.probe[Int]()
+      val upstream = TestPublisher.probe[Int]()
+
+      Source.fromPublisher(upstream).via(failyStage).runWith(Sink.fromSubscriber(downstream))
+      latch.countDown()
+
+      // If an NPE would happen due to unset exposedPublisher (see #19338) this would receive the "I am dead" message
+      // do to postStop being executed earlier than expected.
+      expectNoMsg(500.millis)
```
rkuhn: what you could do is to use another output port from which we expect a single element that comes from the source, then we’d see the NPE bubbling up there, right?

rkuhn:
LGTM after considering my alternative test proposal

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/ActorGraphInterpreterSpec.scala
drewhk: I didn't want a graph though, but yes that is another option.

rkuhn: well, it would be non-racy and it would not rely on some sleep period, so I’d consider that an improvement

drewhk: ok, I give it a try

patriknw:
perhaps add a `require` (or similar) at the place where the NPE occured, telling that exposed publisher is expected first

rkuhn:
That would not add any information but would have a runtime cost, would it not?

drewhk:
Ok, although in this case the NPE was fully informative -- the issue was figuring out how that could happen. Also, the message would be misleading here, since it already tells what might be the issue ("exposed publisher expected first") but it is actually wrong: exposed publisher was there, but not set.

patriknw:
Ok, I agree

rkuhn:
merge at will!

ktoso:
LGTM

drewhk:
port forward!

PR #19414, opened 2016-01-12
htp #19094 fix double pull in PoolConductor

rkuhn:
lacking a test case, just for discussion

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
@@ -75,8 +75,8 @@ private[http] object OutgoingConnectionBlueprint {
         case (Seq(MessageEnd), remaining) ⇒
           SubSource.kill(remaining)
           false
-        case _ ⇒
-          true
+        case (seq, _) ⇒
+          seq.nonEmpty
       }
       .map {
         case (Seq(ResponseStart(statusCode, protocol, headers, createEntity, _)), entityParts) ⇒
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
@@ -140,15 +140,16 @@ private object PoolConductor {
               slotStates(slotIx) = slotStateAfterDisconnect(slotStates(slotIx), failed)
           }
           pull(slotIn)
-          val tryPull = nextSlot == -1
+          val wasBlocked = nextSlot == -1
           nextSlot = bestSlot()
-          if (tryPull) tryPullCtx()
+          val nowUnblocked = nextSlot != -1
+          if (wasBlocked && nowUnblocked) pull(ctxIn) // get next request context
         }
       })
 
       setHandler(out, eagerTerminateOutput)
 
-      val tryPullCtx = () ⇒ if (nextSlot != -1) pull(ctxIn)
+      val tryPullCtx = () ⇒ if (nextSlot != -1 && !hasBeenPulled(ctxIn)) pull(ctxIn)
 
       override def preStart(): Unit = {
         pull(ctxIn)
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/ws/Websocket.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/ws/Websocket.scala
@@ -82,7 +82,7 @@ private[http] object Websocket {
       }
     }
 
-    /** Collects user-level API messages from MessageDataParts */
+    /* Collects user-level API messages from MessageDataParts */
     val collectMessage: Flow[MessageDataPart, Message, Unit] =
       Flow[MessageDataPart]
         .prefixAndTail(1)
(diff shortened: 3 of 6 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolConductor.scala
```diff
@@ -140,15 +140,15 @@ private object PoolConductor {
               slotStates(slotIx) = slotStateAfterDisconnect(slotStates(slotIx), failed)
           }
           pull(slotIn)
-          val tryPull = nextSlot == -1
+          val wasBlocked = nextSlot == -1
           nextSlot = bestSlot()
-          if (tryPull) tryPullCtx()
+          if (wasBlocked && nextSlot != -1) pull(ctxIn)
```
drewhk: I am still not fully understanding what this intends to do. I get it that it checks for the transition of nextSlot == -1 to nextSlot != -1, but I don't get why this is a useful event to decide when to pull `ctxIn`?

rkuhn: The slot selector only blocks if the currently best slot back-pressures or if there is no available slot—as soon as one becomes available it should pull again because it will be able to emit the resulting request as well.

drewhk: ah right, this is the slot port. Maybe name `nextSlot != -1` to `nowUnblocked` and add a comment:

    if (wasBlocked && nowUnblocked) pull(ctxIn) // Get next request context

Not to nitpick, but this is mostly alien code to me so any help is nice to have :)

rkuhn: good point, will do

rkuhn:
fixes #19094 

fixes #19417

akka-http-core/src/test/scala/akka/http/impl/engine/ws/BypassRouterSpec.scala
```diff
@@ -0,0 +1,49 @@
...
+        case HttpRequest(_, _, headers, _, _) ⇒
+          val upgrade = headers.collectFirst { case u: UpgradeToWebsocket ⇒ u }.get
+          upgrade.handleMessages(Flow.apply, None)
+      }, interface = "localhost", port = 8080)
+      val binding = Await.result(bindingFuture, 3.seconds)
+
+      val (future, t) = Http().singleWebsocketRequest(
+        WebsocketRequest("ws://127.0.0.1:8080"),
+        Flow.fromSinkAndSourceMat(
+          Sink.ignore.mapMaterializedValue(_.map(_ ⇒ "hallo")),
```
drewhk: I would check the number of messages received

drewhk: Btw, a Sink.count would be useful.

drewhk:
LGTM

akka-http-core/src/test/scala/akka/http/impl/engine/ws/BypassRouterSpec.scala
ktoso: Yeah, comes up sometimes - ticketify

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -75,8 +75,8 @@ private[http] object OutgoingConnectionBlueprint {
         case (Seq(MessageEnd), remaining) ⇒
           SubSource.kill(remaining)
           false
-        case _ ⇒
-          true
+        case (seq, _) ⇒
+          seq.nonEmpty
```
ktoso: I see

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Websocket.scala
```diff
@@ -172,13 +172,12 @@ private[http] object Websocket {
           }
         }
       })
-      val pullIn = () ⇒ pull(in)
+      val pullIn = () ⇒ tryPull(in)
```
ktoso: Hm, try pull is `if (!closed(in))`, I thought we'll want to pull `if (!hasBeenPulled(in))` - why try here?

ktoso:
LGTM, had one question for my better understanding of the fix

akka-http-core/src/test/scala/akka/http/impl/engine/ws/BypassRouterSpec.scala
rkuhn: #19422 (will use `fold` here)

akka-http-core/src/main/scala/akka/http/impl/engine/ws/Websocket.scala
rkuhn: This case was not a double-pull, it was a pull-when-closed that happens because we unconditionally execute the pull even when the stage was completed while the emit was running.

ktoso: Cool, I see

PR #19415, opened 2016-01-12
=htp #19331 Make FileAndResourceDirectives work in OSGi environments

briantopping:
Reopened against master


briantopping:
This has gotten to be a very confusing patch. @2beaucoup is pushing very hard for his own riff off the debugging I did for this, but doesn't have any unit tests and changes the API of the code (no longer "=" AFAICT). I don't have time to test his work due to a JDK crash I am knee deep in. 

I believe this is the safest path to getting this fixed. Since OSGi environments are broken as a result of #17495, we really need to get this fixed before any release.

2beaucoup:
> @2beaucoup is pushing very hard for his own riff off the debugging I did for this...

Seriously? C'mon Brian. I just tried to help.

briantopping:
We're all trying to help.

2beaucoup:
See https://github.com/playframework/playframework/commit/195f019f8cd68fc0900f6604f76f9619ca5cf5d9.

briantopping:
Also see http://stackoverflow.com/questions/6244993/no-access-to-bundle-resource-file-osgi#answer-6251422. Testing with Felix is not testing with other containers such as Equinox.

Even outside OSGi, there's no requirement that the `getResource` family of API does not return locally unique schemes that do the right thing when treated opaquely. 

There's no way to code to all these schemes, see what I mean? I strongly believe the rollback should be to #17495, not opening up a can of worms with potentially dozens of scheme handlers, one at a time as each poor user finds Akka-HTTP doesn't behave in his environment. Others may differ, but directory listings are not worth these headaches for so many.

Maybe directory listings should be an extension and deal with these issues separately.

ktoso:
Please calm down guys. We're all trying to improve things together, no reason to get offensive.
Sorry I was not able to look into these PRs yet, but the streams migration took first priority (and other not so happy life events), will catch up here soon and figure out what we can do/merge here. 
Thanks for your help both, let's be nice to each other please :)

ktoso:
PLS BUILD

ktoso:
LGTM

akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala
```diff
@@ -222,7 +222,14 @@ object FileAndResourceDirectives extends FileAndResourceDirectives {
             ResourceFile(url, entry.getSize, entry.getTime)
           }
         } finally jar.close()
-      case _ ⇒ None
+      case _ ⇒
+        val conn = url.openConnection()
```
jypma: You probably want to do a `url.setConnectTimeout` and `url.setReadTimeout`, since the default is possibly infinite.

And maybe have a unit test for this?

2beaucoup: `url` is the result of `lassLoader.getResource` so it should point to a local resource. Dunno if that makes it safe though.

jypma: Less likely, yes. But e.g. a `URLClassLoader` for a remote JAR file could still have network activity as a result to requesting a resource, depending on how it's implemented. Of course, I don't know if it would then actually apply these timeouts to its own connections.

briantopping: @jypma There's infinite opportunities for this patch to not work. Too many containers out there, too many branches. Even if there was an OSGi test framework for SBT, it's just one of many containers that could depend on custom URLs (eclipse, for example).

The problem is the code itself, it shouldn't be trying to look at opaque URLs. I understand that there's a security hole in directory listings, maybe those should be removed.

ktoso:
After quite some back and forth and one vacation later I say we merge this one.
Thanks for your contribs related to osgi, we certainly welcome any further ones if there's space for improvement.

ktoso:
Refs #19331

briantopping:
Thanks Konrad, welcome back :)

ktoso:
It's good to be back. ただいま。

PR #19424, opened 2016-01-13
str #19338: NPE in ActorGraphInterpreter (fwd port)

drewhk:
(cherry picked from commit 161e01e)

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
@@ -269,12 +269,12 @@ private[stream] object ActorGraphInterpreter {
       }
 
     def exposedPublisher(publisher: ActorPublisher[Any]): Unit = {
+      exposedPublisher = publisher
       upstreamFailed match {
         case _: Some[_] ⇒
           publisher.shutdown(upstreamFailed)
         case _ ⇒
           if (upstreamCompleted) publisher.shutdown(None)
-          else exposedPublisher = publisher
       }
     }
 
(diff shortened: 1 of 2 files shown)


akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -269,12 +269,12 @@ private[stream] object ActorGraphInterpreter {
       }
 
     def exposedPublisher(publisher: ActorPublisher[Any]): Unit = {
+      exposedPublisher = publisher
```
viktorklang: Is there any risk for overwriting? (if exposePublisher is called twice?)

drewhk: Not really.

viktorklang: Shouldn't we guard against that invariant then?

drewhk: Why? There are thousands of such lines where you can check the validity of a previous value (null here) and if I add an assert here, then why not add assertions there? I.e. where do you draw the line?

viktorklang: Should this method do anything if it is invoked twice?

drewhk: It simply shouldn't be called twice.

viktorklang: And how does a reader of this code know that?

rkuhn: We only send one ExposedPublisher message, which is documented accordingly. Endre is right that this is not useful to express in code—just like in many other places.

drewhk: I can add a comment if that helps.

viktorklang: Comment would be nice! :)

PR #19425, opened 2016-01-13
Wip 19094 fix double pull pool conductor rk

rkuhn:
forward port

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
@@ -532,7 +532,9 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
         if (shell.isInitialized) {
           // yes, this steals another shell’s Resume, but that’s okay because extra ones will just not do anything
           finishShellRegistration()
-        } else tryInit(shell)
+        } else if (!tryInit(shell)) {
+          if (activeInterpreters.isEmpty) finishShellRegistration()
+        }
     }
 
   override def preStart(): Unit = {
@@ -564,5 +566,9 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
       }
   }
 
-  override def postStop(): Unit = activeInterpreters.foreach(_.tryAbort(AbruptTerminationException(self)))
+  override def postStop(): Unit = {
+    val ex = AbruptTerminationException(self)
+    activeInterpreters.foreach(_.tryAbort(ex))
+    newShells.foreach(s ⇒ if (tryInit(s)) s.tryAbort(ex))
+  }
 }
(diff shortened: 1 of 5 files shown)


akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -532,7 +532,9 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
         if (shell.isInitialized) {
           // yes, this steals another shell’s Resume, but that’s okay because extra ones will just not do anything
           finishShellRegistration()
-        } else tryInit(shell)
+        } else if (!tryInit(shell)) {
```
viktorklang: Any reason for not: 
~~~scala
} else if (!tryInit(shell) && activeInterpreters.isEmpty) finishShellRegistration()
~~~

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -564,5 +566,8 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
       }
   }
 
-  override def postStop(): Unit = activeInterpreters.foreach(_.tryAbort(AbruptTerminationException(self)))
+  override def postStop(): Unit = {
+    activeInterpreters.foreach(_.tryAbort(AbruptTerminationException(self)))
```
viktorklang: Allocate the exception once?

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
rkuhn: Yes: I hate hiding side-effects among simple tests. Sometimes it is hard to avoid, but here I think this is a lot clearer.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
rkuhn: This is a very good point, somehow I was blind :-)

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
viktorklang: Good point. I can't help to think that the failure-behavior ought to be possible to embed within tryInit itself. It does seem scattered throughout the class as is.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
viktorklang: I sense allocations like a disturbance in the Force.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
rkuhn: Well, this is not failure handling, it is the detection of when this actor ought to shut itself down—which is not related to any single tryInit call in itself.

viktorklang: Fair enough!

ktoso:
PLS BUILD

ktoso:
Should perhaps survive the merged build on jenkins without need to rebase here hm

PR #19434, opened 2016-01-13
fix github links for stream-http-2.0 branch

rkuhn:


patriknw:
I think there is a mention in CONTRIBUTING also

ktoso:
Thanks a lot for noticing and please don't merge yet, will make my life a bit easier with the bigger PR ;)

ktoso:
Ah, that's on 2.0 – LGTM then :)

rkuhn:
The MiMa failure from Jenkins does not make sense at all, the code it complains about is in another PR that is not yet merged, and checking out the current PR’s very commit I cannot reproduce any of it (i.e. the code really is not there)—weird.

Issue #19418, opened 2016-01-12
[akka-http] scalac bug when using members of MediaTypes in a map


rklaehn:
The members of `akka.http.scaladsl.model.MediaTypes` have different types. Some are of type `MediaType.WithOpenCharset` while others are of type `MediaType.WithFixedCharset`. Using them in isolation works fine, but when scalac is asked to calculate the LUB of both types, it blows up in the most peculiar way. This can be worked around by declaring the type explicitly.

```
Welcome to Scala version 2.11.7 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_45).
Type in expressions to have them evaluated.
Type :help for more information.

scala> import akka.http.scaladsl.model.{MediaType, MediaTypes}
import akka.http.scaladsl.model.{MediaType, MediaTypes}

scala> val a = MediaTypes.`application/xml`
a: akka.http.scaladsl.model.MediaType.WithOpenCharset = application/xml

scala> val b = MediaTypes.`application/json`
b: akka.http.scaladsl.model.MediaType.WithFixedCharset = application/json

scala> val boom = Seq(a, b)
error: missing or invalid dependency detected while loading class file 'MediaType.class'.
Could not access type WithFixedCharset in trait akka.http.javadsl.model.MediaType,
because it (or its dependencies) are missing. Check your build definition for
missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
A full rebuild may help if 'MediaType.class' was compiled against an incompatible version of akka.http.javadsl.model.MediaType.

scala> val works = Seq[MediaType](a, b)
works: Seq[akka.http.scaladsl.model.MediaType] = List(application/xml, application/json)
```

Here is a repo with a tiny reproducer https://github.com/rklaehn/invaliddependency

ktoso:
Thanks for digging this up! Related scalac issue: https://issues.scala-lang.org/browse/SI-9621
Delaying decision if we need a workaround until we get an estimate there

ktoso:
Root of the issue is sadly a long-standing one, we should do a workaround I suggest.
Workaround seems to be: avoiding nested interface and making it a top level one. Yes, sadly not as nice, but at least it'll compile - see details in https://issues.scala-lang.org/browse/SI-9621 root cause: https://issues.scala-lang.org/browse/SI-1409

// FYI @sirthias @jrudolph

jrudolph:
[...]

PR #19453, opened 2016-01-14
=htc #19418 avoid LUB calculation problem, see SI-9621

ktoso:
Resolves https://github.com/akka/akka/issues/19418

// port to master is here https://github.com/akka/akka/issues/19454


drewhk:
Ok, that was painless. LGTM

ktoso:
duh should have checked MiMa

ktoso:
Problem is not from this PR actually:

```scala
[error]  * class akka.stream.impl.fusing.GraphStages#Breaker is declared final in new version
[error]    filter with: ProblemFilters.exclude[FinalClassProblem]("akka.stream.impl.fusing.GraphStages$Breaker")

```

rkuhn:
This mima thing is really worrisome: Breaker was not in there before, so what exactly is it complaining about? It seems to me that on Jenkins (and only there!) the comparison is made not against 2.0.1 but …

WAIT A SECOND

Please remove all akka-stream artifacts for version 2.0.1 from the maven and ivy caches, they are corrupted due to my mistake when basing the stream-http-2.0 branch on the 2.0.1 tag. That will fix it.

rkuhn:
do we publish nightlies? in that case please check the snapshots repo as well …

sorry about this!

ktoso:
Hmm, I don't quite get how the wrong branch poisoned this...?
I'll clean chackes.

ktoso:
I think I cleaned everything now, will check where else I need to apply the LUB fix and merge / release

ktoso:
Additional fix in ContentTypes now added. Awaiting validation

ktoso:
because removed branch

ktoso:
I need to figure out how to add links to these FAILed messages again...

Issue #19447, opened 2016-01-14
Remove Java 6 Compatibility


2beaucoup:

drewhk:
A bit more info would be nice :)

ktoso:
`akka.http.impl.util.Java6Compat` :smile:

2beaucoup:
among others... ;)

2beaucoup:
I'll update the description once I'm through this.

ktoso:
Thanks @2beaucoup !

2beaucoup:
Should I remove `HandlerX`/`AsyncHandlerX` in `akka.http.javadsl.server` and replace usages with lambdas?

ktoso:
Usages for sure can be replaced with lambdas in all places.

Now thinking if we remove the interfaces too then. The only thing they help with is not having to write the `RequestContext, RouteResult` part of the handler functions, so Handler1 is 1 param etc. In case anyone sees them explicitly I think they may be helpful to keep - WDYT @jrudolph?

Note: We're also considering the other javadsl (closer to scaladsl), nothing decided yet yet though.

Issue #18860, opened 2015-11-05
Use @SafeVarargs when on Java 8 


2m:
Add back annotations that got removed in https://github.com/akka/akka/pull/18852 and other places where appropriate.

ktoso:
I'll just do it then ;)

ktoso:
In fact, it's been done already in aed6771

PR #19455, opened 2016-01-14
Remove java6 compat

aruediger:


akka-actor/src/main/java/akka/japi/JAPI.java
```diff
@@ -4,6 +4,7 @@
 
 public class JAPI {
 
+  @SafeVarargs
```
ktoso: :+1:

akka-http-core/src/main/scala/akka/http/impl/util/SettingsCompanion.scala
```diff
@@ -49,7 +49,7 @@ private[http] abstract class SettingsCompanion[T](protected val prefix: String)
 private[http] object SettingsCompanion {
   lazy val configAdditions: Config = {
     val localHostName =
-      try new InetSocketAddress(InetAddress.getLocalHost, 80).getHostStringJava6Compatible
+      try new InetSocketAddress(InetAddress.getLocalHost, 80).getHostString
```
ktoso: :smile:

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -770,12 +770,7 @@ trait DefaultSSLContextCreation {
     defaultParams.setCipherSuites(cipherSuites)
 
     // hostname!
-    if (!Java6Compat.trySetEndpointIdentificationAlgorithm(defaultParams, "https")) {
-      log.info("Unable to use JDK built-in hostname verification, please consider upgrading your Java runtime to " +
-        "a more up to date version (JDK7+). Using Typesafe ssl-config hostname verification.")
-      // enabling the JDK7+ solution did not work, however this is fine since we do handle hostname
-      // verification directly in SslTlsCipherActor manually by applying an ssl-config provider verifier
-    }
+    defaultParams.setEndpointIdentificationAlgorithm("https")
```
ktoso: thanks

ktoso:
OK TO TEST

ktoso:
Went through in detail, very nice cleanups – thanks a lot @2beaucoup!
LGTM, will merge once validates.

ktoso:
Resolves #19447 
Resolves #18860

drewhk:
Wonderful!!!

2beaucoup:
Fixed the commit msg and did some more minor cleanups.

rkuhn:
LGTM

2beaucoup:
Java 8 All The Things!!!

akka-stream/src/main/scala/akka/stream/impl/io/FileSubscriber.scala
```diff
@@ -28,17 +32,13 @@ private[akka] class FileSubscriber(f: File, bytesWrittenPromise: Promise[Long],
...
-  private var raf: RandomAccessFile = _
   private var chan: FileChannel = _
 
   private var bytesWritten: Long = 0
 
   override def preStart(): Unit = try {
-    raf = new RandomAccessFile(f, "rw") // best way to express this in JDK6, OpenOption are available since JDK7
-    chan = raf.getChannel
-
-    // manually supporting appending to files - in Java 7 we could use OpenModes: FileChannel.open(f, openOptions.asJava)
```
ktoso: here I think it would be nicer to expose the OpenOptions to users and drop the boolean flag. I'll make it a separate ticket

ktoso:
Failure was unrelated IMHO - https://github.com/akka/akka/issues/19464

ktoso:
LGTM

akka-stream/src/main/scala/akka/stream/impl/io/FileSubscriber.scala
2beaucoup: Agreed. Would make sense here. I wonder if we always have to add `WRITE`. I just set `APPEND` here and the test passes.

2beaucoup: @ktoso would the ticket be #19483?

Do you have any idea if `WRITE` is manadatory here? Docs suggest so.

ktoso: Yeah, we can make the options fix as part of that ticket (and then allow making the offset somehow).
I'm not sure, but APPEND implies unix file permission WRITE so I guess it may not be needed.

ktoso: Or maybe it doesn't... since chattr could be used hm to restrict "only append". Let's just set append if it works for you and we'll know about linux on jenkins

2beaucoup:
refs #19455

ktoso:
Refs itself? :)

2beaucoup:
Fixed already. :stuck_out_tongue:

Issue #19443, opened 2016-01-14
create StreamConverters form java.util.stream.Stream


rkuhn:

drewhk:
Do you remember what where the conversion names we came up with?

rkuhn:
something like `asJavaStream` and `fromJavaStream`, I think

drewhk:
putting it back

drewhk:
here is the in-progress PR: https://github.com/akka/akka/pull/19456

drewhk:
Anyone, feel free to take my PR and finish this ticket.

agolubev:
taking it if no objections

rkuhn:
sure, thanks @agolubev !

agolubev:
@drewhk, it seems for me that Sink.javaCollector (from here: https://github.com/akka/akka/pull/19456/commits/72a60674e782426d66de7e43bb75a57e352bd381#diff-fb0fa8a4e622c766a811e2af3a9a3289R226) will  not be so useful in case we implement materialization to java.util.stream.Stream (that's what I'm going to do). You suggested this in comment here: https://github.com/akka/akka/pull/19456#issuecomment-171686122

Do you want to keep Sink.javaCollector anyway?

PR #19456, opened 2016-01-14
#19443: Java Stream support

drewhk:
*\* PREVIEW **

Added support for conversion from Stream and using Collectors (also in unordered parallel mode).

Need some feedback on what we would need. (like do we need to expose a blocking Sink.asStream)

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
@@ -186,6 +186,12 @@ object Source {
       override def toString: String = "() => Iterator"
     })
 
+  def fromJavaStream[T, S <: java.util.stream.BaseStream[T, S]](stream: () ⇒ java.util.stream.BaseStream[T, S]): Source[T, Unit] = {
+
+    import scala.collection.JavaConverters._
+    Source.fromIterator(() ⇒ stream().iterator().asScala)
+  }
+
   /**
    * A graph with the shape of a source logically is a source, this method makes
    * it so also in type.
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/SinkSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/SinkSpec.scala
@@ -3,12 +3,16 @@
  */
 package akka.stream.scaladsl
 
+import java.util.function.ToIntFunction
+import java.util.stream.Collectors
+
 import akka.stream._
-import akka.stream.testkit.TestPublisher.ManualProbe
 import akka.stream.testkit._
+import akka.testkit.DefaultTimeout
+import org.scalatest.concurrent.ScalaFutures
 import scala.concurrent.Future
 
-class SinkSpec extends AkkaSpec {
+class SinkSpec extends AkkaSpec with DefaultTimeout with ScalaFutures {
 
   import GraphDSL.Implicits._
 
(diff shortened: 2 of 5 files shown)


akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -186,6 +186,12 @@ object Source {
       override def toString: String = "() => Iterator"
     })
 
+  def fromJavaStream[T, S <: java.util.stream.BaseStream[T, S]](stream: () ⇒ java.util.stream.BaseStream[T, S]): Source[T, Unit] = {
```
rkuhn: didn’t we want to put this in the StreamConverters object?

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -186,6 +186,12 @@ object Source {
...
+
+    import scala.collection.JavaConverters._
+    Source.fromIterator(() ⇒ stream().iterator().asScala)
```
rkuhn: this should probably be documented to benefit from an asyncBoundary in certain cases (once you get to writing docs, that is :-) )

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
drewhk: The reason why I put here, because it is a natural place on Source for the `javadsl` and I wanted to be symmetric.

rkuhn:
We discussed this (painfully, many times): Source and Sink are for the very basic functions, we group others by theme (like StreamConverters and FileIO). These are stream converters, and we said so back then. You even liked that.

Now, concerning Sinks: the Collector-based one looks good, we should also provide one that materializes to a Spliterator so that a Java 8 Stream can easily be created from it—under the covers it will need to use Sink.queue/Await or similar.

drewhk:
> These are stream converters, and we said so back then. You even liked that.

That was for InputStream and OutputStream, this is a different thing, since this is fundamental for Java Collections.

drewhk:
Alternative (or additionally) to materializing to a apliterator is to add a `toJavaStream` to `javadsl.Source` that does not materialize immediately (but needs to grab a materializer) but returns a custom `Stream` subclass that lazily builds the underlying Source while possible, only blocking on either a terminal operation (Collector) or when an iterator is explicitly requested.

rkuhn:
That is interesting, but only if we commit to that a Materializer must always be thread-safe.

drewhk:
That is not necessary in this case. A Stream is only allowed to be consumed once (unlike a Source), and if the materializer is not thread-safe then the creator of the Stream must provide an exclusive one.

rkuhn:
No, what I mean is that the Stream instance may then be passed to another execution context and materialized there, which means that the materializer that is captured needs to be fine with that.

drewhk:
well, that is weaker than thread-safe (as in concurrent calls to `materialize`) as it only requires that it should well behave if materialization is done on a different thread from where the materializer has been created. I am not sure that is an issue. Anyway, we do not need to create such a Stream interface yet (although it seems quite straightforward so far), we can do it later.

rkuhn:
Right, that is what I meant. Currently that is safe, the question is whether we should require this feature from all future Materializer implementations.

Actually it needs to be safe because graph stages may already use the materializer concurrently. Given that, I think such a Source-Stream should be fine.

drewhk:
I am reopening this when I actually do something with it. I think this is less priority than the rest of Java8 changes since this is just added functionality.

PR #19458, opened 2016-01-14
#19457: emitMultiple should handle empty iterator

### LINKED ISSUE #19457: emitMultiple in GraphStage does not call andThen if iterator is empty
https://github.com/akka/akka/blob/7ee50d18ae9793e21960e6d4d00f6fb3ed0cf4fa/akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala#L760

`emitMultiple` does check `if (element.hasNext)` and if it is false it does nothing but should call `andThen`.

drewhk:
Fixes #19457

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/impl/GraphStageLogicSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/impl/GraphStageLogicSpec.scala
@@ -13,9 +13,10 @@ import akka.stream.testkit.scaladsl.TestSink
 import akka.stream.impl.fusing._
 import akka.stream.impl.fusing.GraphInterpreter._
 import org.scalactic.ConversionCheckedTripleEquals
+import org.scalatest.concurrent.ScalaFutures
 import scala.concurrent.duration.Duration
 
-class GraphStageLogicSpec extends AkkaSpec with GraphInterpreterSpecKit with ConversionCheckedTripleEquals {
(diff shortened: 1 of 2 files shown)


johanandren:
LGTM!

rkuhn:
LGTM

akka-stream-tests/src/test/scala/akka/stream/impl/GraphStageLogicSpec.scala
```diff
@@ -67,6 +68,18 @@ class GraphStageLogicSpec extends AkkaSpec with GraphInterpreterSpecKit with Con
     }
   }
 
+  object emitEmptyIterable extends GraphStage[SourceShape[Int]] {
+    val out = Outlet[Int]("out")
```
ktoso: for discussion in the future – we have 2 different styles now, sometimes "out" and sometimes "StageName-out", we should settle on one style, (not for this issue ofc)

ktoso:
LGTM, oh Endre merged first :) Good

akka-stream-tests/src/test/scala/akka/stream/impl/GraphStageLogicSpec.scala
drewhk: this is just a test class and I was lazy. StageName.out is the official one.

ktoso: Ah, right, doesnt matter that much in test.

PR #19461, opened 2016-01-14
+str #19176 Added Partition Graph Stage with tests

israel:


akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec
```diff
@@ -0,0 +1,173 @@
+/**
```
drewhk: this file is missing a `.scala` extension!

drewhk:
I will enable validation as soon as the test gets a proper `.scala` extension (to avoid an unnecessary validation run).

johanandren:
OK TO TEST

drewhk:
failure is known and unrelated

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
```diff
@@ -0,0 +1,173 @@
...
+
+  val settings = ActorMaterializerSettings(system)
+    .withInputBuffer(initialSize = 2, maxSize = 16)
+
+  implicit val materializer = ActorMaterializer(settings)
+
+  "A partition" must {
+    import GraphDSL.Implicits._
+
+    "partition to three subscribers" in assertAllStagesStopped {
```
drewhk: you can simplify this test and feed the input port from an Iterable and drain the results to collections using Sink.seq.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
```diff
@@ -0,0 +1,173 @@
...
+      }).run()
+
+      c2.request(2)
+      c1.request(2)
+      c3.request(1)
+      c2.expectNext(1)
+      c2.expectNext(2)
+      c3.expectNext(3)
+      c1.expectNext(4)
+      c1.expectNext(5)
```
drewhk: expect completion

drewhk: or rewrite the test to drain to sequences instead

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
```diff
@@ -0,0 +1,173 @@
...
+      }).run()
+
+      c1.request(1)
+      c2.request(4)
+      c1.expectNext("another")
+      c2.expectNext("this")
+      c2.expectNext("is")
+      c2.expectNext("just")
+      c2.expectNext("test")
+
```
drewhk: expect completion

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
```diff
@@ -0,0 +1,173 @@
...
+        partition.out(1) ~> merge.in(1)
+        merge.out ~> Sink.fromSubscriber(probe)
+
+        ClosedShape
+      }).run()
+
+      val subscription = probe.expectSubscription()
+
+      var collected = Set.empty[Int]
+      for (_ ← 1 to 7) {
```
drewhk: No need for this, there is also a version of expect that takes a collection, or even better, drain it with a Sink.seq

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -467,6 +467,106 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    private var outPending: Any = null
+    private var downstreamRunning = outputPorts
+
+    setHandler(in, new InHandler {
+      override def onPush() = {
+        val elem = grab(in)
+        val idx = partitioner(elem)
+        if (idx >= outputPorts)
+          failStage(new IndexOutOfBoundsException(s"partitioner must return an index in the range [0,${outputPorts - 1}]. returned: [$idx] for input [$elem]."))
```
drewhk: forgot to check for < 0

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -467,6 +467,106 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+        else if (!isClosed(out(idx))) {
+          if (isAvailable(out(idx))) {
+            push(out(idx), elem)
+            if (out.exists(isAvailable(_)))
```
drewhk: it is cheaper to keep a counter to track this. Alternative is to always pull the input (prefetch). We usually avoid prefetch in stages (the async boundary does it when necessary automatically) but this stage ends up "prefetching" anyway since you never know which output the next element will belong to (alternative would be out.forAll(isAvailable(_)) but that is unnecessarily conservative). 

I would try out a version where you simply always prefetch, i.e. you call pull here unconditionally (or at the point where you actually grab the elem)

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -467,6 +467,106 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
...
+              pull(in)
+          } else
+            outPending = (idx -> elem)
```
drewhk: please add two variables instead, this unnecessarily allocates a Pair

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
drewhk: Also, since the current test exercises a relatively strict event sequence, using collections would allow the fuzzer to explore more possible sequences.

drewhk:
Will be in merge shape after these minor fixes

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
israel: I don't understand.
I need to check that elements end up in the correct OUTs, not just to drain the result into one sequences.

drewhk: you can attach a sink on each of the output ports. That is not one sequence, but one per *each port*. Makes this test code more compact and allows more fuzzing opportunities.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
israel: I removed completion checks from the tests due to https://github.com/akka/akka/pull/19190#discussion_r47753942
I will bring it back if you want.

drewhk: I think you misunderstood that comment. What @rkuhn said is more or less the same as I recommend:
 - for purely functional testing, avoid probes
 - for tests, that *explicitly test* tricky completion issues (like when does a certain port complete relative to another port) which are part of the semantics of the stage, or ordering issues which are also part of the semantics, use probes

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: I have discussed this with @rkuhn in the previous PR. I cannot keep a counter to track this since I cannot update it when an OUT is leaving in onDownstreamFinished since in such occasion, I don't know if the finished OUT was pending or not.
As for prefetching, I cannot prefetch after grabbing, since if I end up putting it in outPending, I must not pull again.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: done

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
drewhk: Ok, thanks for the explanation.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: done

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
israel: done

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
drewhk: this is roughly how it would look like:

```scala
     val (s1, s2, s3) = RunnableGraph.fromGraph(GraphDSL.create(Sink.seq, Sink.seq, Sink.seq)(Tuple3.apply) { implicit b => (sink1, sink2, sink3) ⇒
       val partition = b.add(Partition[Int](3, {
          case g if (g > 3)  ⇒ 0
          case l if (l < 3)  ⇒ 1
          case e if (e == 3) ⇒ 2
        }))
        Source(List(1, 2, 3, 4, 5)) ~> partition.in
        partition.out(0) ~> sink1
        partition.out(1) ~> sink2
        partition.out(2) ~> sink3
        ClosedShape
      }).run()
   s.futureValue should ===(List(1, 2)) // This needs the ScalaFutures trait mixed in the test from scalatest
```

rkuhn:
fixes #19176 

merge?

drewhk:
Well, I liked one test case to be refactored a bit, but if it does not matter.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
israel: I understand now. I can change if you want.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
israel: My bad. I have added all the completion checks

akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala
drewhk: I would appreciate it yes. It is somewhat easier to read and provides better test coverage.

israel: done. in https://github.com/akka/akka/pull/19592

Issue #19466, opened 2016-01-14
Failure using `Builder` to create `ByteString`


briantopping:
https://github.com/briantopping/akka/tree/wip-ByteString-builder has an extra test in `akka/util/ByteStringSpec.scala`. It demonstrates that when a `Builder` is created for a `ByteString`, a `Vector` is produced. Which is fine, but the `result` should return a `ByteString`. Instead, it returns a `Vector`.

Is this correct behavior? Should this test be turned into a PR?

luben:
Putting a little bit more into it.

ByteString implements Traversable through IndexedSeq <: GenericTraversableTemplate. One of the the trait methods is "genericBuilder" that should return new builder for the collection. Currently ByteString does not implement this method and inherits the implementation from GenericTranversableTemplate that gets it from companion.newBuilder but ByteString do not provide also the "companion" method so it gets it from IndexedSeq that defaults to returning VectorBuilder (as the most generic IndexedSeq).

This represents a problem when trying to generically serialize/deserialize scala Traversable collections: deserializing through the genericBuilder of ByteString results in Vector[Byte] that is not expected.

The right solution is to add "companion" method to the ByteString sealed abstract class that returns the ByteString companion object. Another option is to implement directly the "genericBuilder" method.

briantopping:
Awesome, thanks! My brain is pretty fried from learning the guts of Kryo this week :) 

I think you did most of the work here, do you want to gen the PR or shall I take a swing at it?

luben:
Yes, I'll do it.

luben:
I took a closer look - it looks like a hard task to reconcile the standard library expectations and the ByteString implementation. The main difficulties:

- the companion method should return GenericCompanion[IndexedSeq] and it expect the collection itself to be parametic over the item type.

- in the similar way the genericBuilder requires Builder of generic collection which ByteString isn't

You can give it a try if you have an idea how to solve it.

patriknw:
@luben sounds like there were no solution here, shall I close the ticket?

briantopping:
This may have been a PBKAC on my part, closing.

luben:
Yes, sorry for not closing it earlier. We took care of the exception on the akka-kryo-serialization library side creating a specialized serializer for ByteString

patriknw:
thanks

PR #19468, opened 2016-01-15
=act #19466 add failing test for `ByteString`

briantopping:
This obviously needs a fix before it can be committed. Please also see romix/akka-kryo-serialization#69 for more background here and a workaround that @luben very kindly solved and put together. :+1:

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/util/ByteStringSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/util/ByteStringSpec.scala
@@ -16,6 +16,7 @@ import org.scalacheck.{ Arbitrary, Gen }
 import org.scalatest.{ Matchers, WordSpec }
 import org.scalatest.prop.Checkers
 
+import scala.collection.{ immutable, mutable }
 import scala.collection.mutable.Builder
 
 class ByteStringSpec extends WordSpec with Matchers with Checkers {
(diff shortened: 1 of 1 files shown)


johanandren:
Refs #19466

akka-actor-tests/src/test/scala/akka/util/ByteStringSpec.scala
```diff
@@ -617,5 +618,10 @@ class ByteStringSpec extends WordSpec with Matchers with Checkers {
         }
       }
     }
+    "return the correct class from Builder.result" in {
+      val builder = ByteString().genericBuilder[String]
```
viktorklang: Is there any reason why this method should not throw an UnsupportedOperationException?

Use ByteString.newBuilder instead?

ktoso:
For reference, the origin of this issue is resolved:
The original problem case was kryo serialization but it now handles `ByteString` as first class citizen thanks to @luben's fix here: https://github.com/romix/akka-kryo-serialization/pull/70

briantopping:
Did Kryo work around it or is it a base implementation problem?

viktorklang:
@briantopping What should a `genericBuilder` for ByteString look like?

briantopping:
I see.. thanks!

luben:
Just adding a context for the history:

The akka-kryo-serialization has a generic serializer/deserializer for the collections that implement Traversable[T] trait. This serializer gets picked for ByteString because it implements the trait through IndexedSeq[T]. On deserializing the object it uses the genericBuilder method and in the case with ByteString it get a VectorBuilder instance because this is inherited from the immutable.IndexedSeq.

So I don't think ByteString should raise Unsupported Operation exception - it is part if the Traversable trait that it implements. The problem was solver on the akka-kryo-serialization side by implementing specialized serializer/deserializer for ByteString - it has a better performance as well.

ktoso:
Right, and thanks a lot for the fix in the lib and chiming in here @luben!

Issue #19464, opened 2016-01-14
FAILED: StreamTestKitSpec TestSink Probe


ktoso:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/4828/console

```
[info] A TestSink Probe
[info] - must #toStrict (172 milliseconds)
[info] - must #toStrict with failing source *** FAILED *** (47 milliseconds)
[info]   "toStrict received OnError while draining stream! Accumulated elements: List(1, 2)" did not include substring "Boom!" (StreamTestKitSpec.scala:40)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.MatchersHelper$.newTestFailedException(MatchersHelper.scala:160)
[info]   at org.scalatest.Matchers$ShouldMethodHelper$.shouldMatcher(Matchers.scala:6231)
[info]   at org.scalatest.Matchers$AnyShouldWrapper.should(Matchers.scala:6265)
[info]   at akka.stream.testkit.StreamTestKitSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply$mcV$sp(StreamTestKitSpec.scala:40)
[info]   at akka.stream.testkit.StreamTestKitSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply(StreamTestKitSpec.scala:24)
[info]   at akka.stream.testkit.StreamTestKitSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply(StreamTestKitSpec.scala:24)
[info]   at org.scalatest.Transformer$$anonfun$apply$1.apply$mcV$sp(Transformer.scala:22)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.WordSpecLike$$anon$1.apply(WordSpecLike.scala:953)
[info]   at org.scalatest.Suite$class.withFixture(Suite.scala:1122)
[info]   at akka.stream.testkit.AkkaSpec.withFixture(AkkaSpec.scala:58)
[info]   at org.scalatest.WordSpecLike$class.invokeWithFixture$1(WordSpecLike.scala:950)
[...]

Issue #19441, opened 2016-01-14
convert from Scala Option to Optional in javadsl packages


rkuhn:
using scala-java8-compat library

hepin1989:
+1

rkuhn:
We’re doing these changes for the still-experimental Streams & HTTP only, to see how difficult and breaking it is. We’ll work on the rest of Akka afterwards.

ktoso:
This should be closed when step 2 (applying this in all HTTP classes is completed)

2beaucoup:
Will this be done for functions as well?

ktoso:
Functions is a trickier question I think... since java has Function and BiFunction, but from 3params onwards you're on your own... Perhaps it's more consistent to use our Function types everywhere // cc @rkuhn @drewhk

drewhk:
> This should be closed when step 2 (applying this in all HTTP classes) is completed

@ktoso wdym? I though I changed all relevant HTTP classes.

rkuhn:
There are a few left in
~~~
akka-http-core/src/main/java/akka/http/javadsl/HttpsContext.java
akka-http-core/src/main/scala/akka/http/javadsl/Http.scala
akka-http/src/main/scala/akka/http/javadsl/server/directives/BasicDirectives.scala
akka-http/src/main/scala/akka/http/javadsl/server/values/*
~~~

2m:
Fixed in #19441

PR #19476, opened 2016-01-15
Wip 19441 java optional in javadsl drewhk

drewhk:


drewhk:
Migration guide is coming next

drewhk:
what? This compiled locally what is the isse now...

patriknw:
Here is another one: https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/Attributes.scala#L69

drewhk:
Hm, that was not a japi.Option hence I missed it. Thanks for noticing, this was broken even in 2.0.x (but fixing this would not be bin compat)

patriknw:
LGTM

drewhk:
PLS BUILD

drewhk:
Validation is down?

ktoso:
hm let's restart jenkins then, willdo

drewhk:
I don't think the test failure is related.

rkuhn:
LGTM (just skimmed)

akka-stream/src/main/scala/akka/stream/impl/Emit.scala
```diff
@@ -1,33 +0,0 @@
...
- * Copyright (C) 2014 Typesafe Inc. <http://www.typesafe.com>
- */
-package akka.stream.impl
-
-import scala.collection.immutable
-
-/**
- * INTERNAL API
- */
-private[akka] trait Emit { this: ActorProcessorImpl with Pump ⇒
```
ktoso: was unused?

ktoso:
LGTM, fix for unrelated failure on it's way in already: https://github.com/akka/akka/issues/19464

ktoso:
Refs #19441

akka-stream/src/main/scala/akka/stream/impl/Emit.scala
drewhk: this one yes, the others are sadly used by groupBy and TLS.

PR #19479, opened 2016-01-15
htc: #16819 implement server-side request timeouts

### LINKED ISSUE #16819: Implement request timeouts on the server side

sirthias:
Closes #16819.

Also contains some improvements in header rendering; most importantly a distinction between request and response headers, which prevents that request headers are rendered into responses and vice versa (which is something that has been bothering me for quite some time already).

/cc @jrudolph @2beaucoup

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/HttpHeader.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/HttpHeader.java
@@ -34,4 +34,14 @@ public abstract class HttpHeader {
      * Returns !is(nameInLowerCase).
      */
     public abstract boolean isNot(String nameInLowerCase);
+
+    /**
+     * Returns true iff the header is to be rendered in requests.
+     */
+    public abstract boolean renderInRequests();
+
+    /**
+     * Returns true iff the header is to be rendered in responses.
+     */
+    public abstract boolean renderInResponses();
 }
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -273,13 +273,16 @@ akka.http {
     max-chunk-ext-length       = 256
     max-chunk-size             = 1m
 
-    # Maximum content length which should not be exceeded by incoming HttpRequests.
-    # For file uploads which use the entityBytes Source of an incoming HttpRequest it is safe to
-    # set this to a very high value (or to `infinite` if feeling very adventurous) as the streaming
-    # upload will be back-pressured properly by Akka Streams.
-    # Please note however that this setting is a global property, and is applied to all incoming requests,
-    # not only file uploads consumed in a streaming fashion, so pick this limit wisely.
-    max-content-length         = 8m
+    # Default maximum content length which should not be exceeded by incoming request entities.
+    # Can be changed at runtime (to a higher or lower value) via the `HttpEntity::withSizeLimit` method.
+    # Note that it is not necessarily a problem to set this to a high value as all stream operations
+    # are always properly backpressured.
+    # Nevertheless you might want to apply some limit in order to prevent a single client from consuming
+    # an excessive amount of server resources.
+    #
+    # Set to `infinite` to completely disable entity length checks. (Even then you can still apply one
+    # programmatically via `withSizeLimit`.)
+    max-content-length = 8m
 
     # Sets the strictness mode for parsing request target URIs.
     # The following values are defined:
--- a/akka-http-core/src/main/resources/reference.conf
+++ b/akka-http-core/src/main/resources/reference.conf
@@ -18,6 +18,18 @@ akka.http {
     # Set to `infinite` to completely disable idle connection timeouts.
     idle-timeout = 60 s
 
+    # Defines the default time period within which the application has to
+    # produce an HttpResponse for any given HttpRequest it received.
+    # The timeout begins to run when the *end* of the request has been
+    # received, so even potentially long uploads can have a short timeout.
+    # Set to `infinite` to completely disable request timeout checking.
+    #
+    # If this setting is not `infinite` the HTTP server layer attaches a
+    # `Timeout-Access` header to the request, which enables programmatic
+    # customization of the timeout period and timeout response for each
+    # request individually.
+    request-timeout = 20 s
+
     # The time period within which the TCP binding process must be completed.
     # Set to `infinite` to disable.
     bind-timeout = 1s
--- a/akka-http-core/src/main/scala/akka/http/ServerSettings.scala
+++ b/akka-http-core/src/main/scala/akka/http/ServerSettings.scala
@@ -46,7 +46,10 @@ final case class ServerSettings(
 
 object ServerSettings extends SettingsCompanion[ServerSettings]("akka.http.server") {
   final case class Timeouts(idleTimeout: Duration,
+                            requestTimeout: Duration,
                             bindTimeout: FiniteDuration) {
+    require(idleTimeout > Duration.Zero, "idleTimeout must be infinite or > 0")
+    require(requestTimeout > Duration.Zero, "requestTimeout must be infinite or > 0")
     require(bindTimeout > Duration.Zero, "bindTimeout must be > 0")
   }
   implicit def timeoutsShortcut(s: ServerSettings): Timeouts = s.timeouts
@@ -55,6 +58,7 @@ object ServerSettings extends SettingsCompanion[ServerSettings]("akka.http.serve
     c.getString("server-header").toOption.map(Server(_)),
     Timeouts(
       c getPotentiallyInfiniteDuration "idle-timeout",
+      c getPotentiallyInfiniteDuration "request-timeout",
       c getFiniteDuration "bind-timeout"),
     c getInt "max-connections",
     c getInt "pipelining-limit",
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
@@ -78,8 +78,8 @@ private[http] class HttpRequestRendererFactory(userAgentHeader: Option[headers.`
           case x: `Raw-Request-URI` ⇒ // we never render this header
             renderHeaders(tail, hostHeaderSeen, userAgentSeen, transferEncodingSeen)
 
-          case x: CustomHeader ⇒
-            if (!x.suppressRendering) render(x)
+          case x: CustomHeader if x.renderInRequests ⇒
+            render(x)
             renderHeaders(tail, hostHeaderSeen, userAgentSeen, transferEncodingSeen)
 
           case x: RawHeader if (x is "content-type") || (x is "content-length") || (x is "transfer-encoding") ||
@@ -88,7 +88,8 @@ private[http] class HttpRequestRendererFactory(userAgentHeader: Option[headers.`
             renderHeaders(tail, hostHeaderSeen, userAgentSeen, transferEncodingSeen)
 
           case x ⇒
-            render(x)
+            if (x.renderInRequests) render(x)
+            else log.warning("HTTP header '{}' is not allowed in requests", x)
             renderHeaders(tail, hostHeaderSeen, userAgentSeen, transferEncodingSeen)
         }
 
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -5,12 +5,19 @@
 package akka.http.impl.engine.server
 
 import java.net.InetSocketAddress
-import java.util.Random
-import akka.stream.impl.fusing.GraphInterpreter
+import java.util.concurrent.atomic.AtomicReference
+import scala.concurrent.{ Promise, Future }
+import scala.concurrent.duration.{ Deadline, FiniteDuration, Duration }
 import scala.collection.immutable
-import org.reactivestreams.{ Publisher, Subscriber }
 import scala.util.control.NonFatal
+import akka.actor.Cancellable
+import akka.japi.Function
 import akka.event.LoggingAdapter
+import akka.util.ByteString
+import akka.stream._
+import akka.stream.io._
+import akka.stream.scaladsl._
+import akka.stream.stage._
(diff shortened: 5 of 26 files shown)


akka-http-core/src/main/java/akka/http/javadsl/model/HttpHeader.java
```diff
@@ -34,4 +34,14 @@
...
+
+    /**
+     * Returns true iff the header is to be rendered in requests.
+     */
+    public abstract boolean renderInRequests();
+
+    /**
+     * Returns true iff the header is to be rendered in responses.
+     */
+    public abstract boolean renderInResponses();
```
ktoso: thanks a lot for moving to this from `suppressRendering` makes a lot of sense!

ktoso:
(Sorry I couldn't review right away on friday, family things happened. I'll be on it as soon as I can)

akka-http-core/src/main/resources/reference.conf
```diff
@@ -18,6 +18,18 @@ akka.http {
...
+    # produce an HttpResponse for any given HttpRequest it received.
+    # The timeout begins to run when the *end* of the request has been
+    # received, so even potentially long uploads can have a short timeout.
+    # Set to `infinite` to completely disable request timeout checking.
+    #
+    # If this setting is not `infinite` the HTTP server layer attaches a
+    # `Timeout-Access` header to the request, which enables programmatic
+    # customization of the timeout period and timeout response for each
+    # request individually.
+    request-timeout = 20 s
```
ktoso: which is also the default in spray, good. thanks

akka-http-core/src/main/resources/reference.conf
```diff
@@ -273,13 +285,16 @@ akka.http {
...
 
-    # Maximum content length which should not be exceeded by incoming HttpRequests.
-    # For file uploads which use the entityBytes Source of an incoming HttpRequest it is safe to
-    # set this to a very high value (or to `infinite` if feeling very adventurous) as the streaming
-    # upload will be back-pressured properly by Akka Streams.
-    # Please note however that this setting is a global property, and is applied to all incoming requests,
-    # not only file uploads consumed in a streaming fashion, so pick this limit wisely.
-    max-content-length         = 8m
+    # Default maximum content length which should not be exceeded by incoming request entities.
+    # Can be changed at runtime (to a higher or lower value) via the `HttpEntity::withSizeLimit` method.
```
ktoso: good catch, thanks for adding the note

akka-http-core/src/main/scala/akka/http/ServerSettings.scala
```diff
@@ -46,7 +46,10 @@ final case class ServerSettings(
 
 object ServerSettings extends SettingsCompanion[ServerSettings]("akka.http.server") {
   final case class Timeouts(idleTimeout: Duration,
+                            requestTimeout: Duration,
                             bindTimeout: FiniteDuration) {
+    require(idleTimeout > Duration.Zero, "idleTimeout must be infinite or > 0")
+    require(requestTimeout > Duration.Zero, "requestTimeout must be infinite or > 0")
     require(bindTimeout > Duration.Zero, "bindTimeout must be > 0")
   }
```
ktoso: note to self: In some modules we moved away from case classes for "config types" as adding a field is binary-hard.
Perhaps this is something we should consider before calling http-core stable in 2.4.2?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -464,42 +436,80 @@ final case class Cookie(cookies: immutable.Seq[HttpCookiePair]) extends jm.heade
...
+ * INTERNAL API
+ */
+private[headers] object EmptyCompanion extends ModeledCompanion[EmptyHeader.type]
+/**
+ * INTERNAL API
+ */
+private[http] object EmptyHeader extends SyntheticHeader {
+  def renderValue[R <: Rendering](r: R): r.type = r
+  protected def companion: ModeledCompanion[EmptyHeader.type] = EmptyCompanion
+}
```
ktoso: good move, thanks

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -57,9 +60,6 @@ sealed trait ModeledHeader extends HttpHeader with Serializable {
  * as they allow the custom header to be matched from [[RawHeader]] and vice-versa.
  */
 abstract class CustomHeader extends jm.headers.CustomHeader {
-  /** Override to return true if this header shouldn't be rendered */
-  def suppressRendering: Boolean = false
-
```
ktoso: So this now makes people to explicitly say if it should be rendered in request/response right?
Do we need to make everyone think about this?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -49,6 +47,11 @@ sealed trait ModeledHeader extends HttpHeader with Serializable {
   protected def companion: ModeledCompanion[_]
 }
 
+private[headers] sealed trait RequestHeader extends ModeledHeader { override def renderInRequests = true }
+private[headers] sealed trait ResponseHeader extends ModeledHeader { override def renderInResponses = true }
+private[headers] sealed trait RequestResponseHeader extends RequestHeader with ResponseHeader
+private[headers] sealed trait SyntheticHeader extends ModeledHeader
```
ktoso: those are *very* nice now :)

akka-http-core/src/main/scala/akka/http/ServerSettings.scala
sirthias: Hmm.. I see.
I'm not really qualified with regard to issues surrounding binary compat, so your judgement/experience will have to lead the way.

ktoso: I remember that this caused some pains when making 2.3/2.4 compatible... I imagine there will be still some movement and additions in http settings so we should perhaps keep it nimble, will check if I remember correctly with the guys on monday.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
sirthias: Yes and yes.
I think it's good to force people to clearly define what their custom header is all about.
Adding a custom header shouldn't be an all too common thing. Deciding on whether and where you want it to be rendered should be good, also for documentation in your code.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
sirthias: Yes, it really does help with inline documentation, even for us! :)

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
ktoso: Yup, agreed :+1:

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -500,4 +501,24 @@ object HttpEntity {
   private object SizeLimit {
     val Disabled = -1 // any negative value will do
   }
+
+  /**
+   * INTERNAL API
+   */
+  def captureTermination[T <: HttpEntity](entity: T): (T, Future[Unit]) =
```
ktoso: `private[http]`?

sirthias: Yup. Missed that one.

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
```diff
@@ -78,8 +78,8 @@ private[http] class HttpRequestRendererFactory(userAgentHeader: Option[headers.`
           case x: `Raw-Request-URI` ⇒ // we never render this header
             renderHeaders(tail, hostHeaderSeen, userAgentSeen, transferEncodingSeen)
 
-          case x: CustomHeader ⇒
-            if (!x.suppressRendering) render(x)
+          case x: CustomHeader if x.renderInRequests ⇒
```
ktoso: Thought about this move from inner if to guard for a while, but I agree. If we get `renderInRequests false` here then someone is doing it wrong and we don't need to optimise for that case.

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
```diff
@@ -88,7 +88,8 @@ private[http] class HttpRequestRendererFactory(userAgentHeader: Option[headers.`
             renderHeaders(tail, hostHeaderSeen, userAgentSeen, transferEncodingSeen)
 
           case x ⇒
-            render(x)
+            if (x.renderInRequests) render(x)
+            else log.warning("HTTP header '{}' is not allowed in requests", x)
```
ktoso: :+1: for the warning

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
sirthias: Yes. And custom headers are really rare anyway.

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -173,6 +184,106 @@ private[http] object HttpServerBluePrint {
...
+        for (old ← oldSetup.fast) {
+          if ((old.scheduledTask eq null) || old.scheduledTask.cancel()) {
+            val newHandler = if (handler eq null) old.handler else handler
+            val newTimeout = if (timeout eq null) old.timeout else timeout
+            val newScheduling = newTimeout match {
+              case x: FiniteDuration ⇒ schedule(old.timeoutBase + x - Deadline.now, newHandler)
+              case _                 ⇒ null // don't schedule a new timeout
+            }
+            newSetup.success(new TimeoutSetup(old.timeoutBase, newScheduling, newTimeout, newHandler))
+          } // else: too late, the previously set timeout cannot be cancelled anymore
```
sirthias: Just realized that we have the risk for a memory leak here:
If another `update` call comes in after the previous one has failed here it will sit on the never-to-be-fulfilled promise forever.
I'll get that fixed.

ktoso: good catch, thanks. I'm out of power for tonight, will review more tomorrow.

sirthias:
Applied review feedback so far.

akka-http-core/src/main/resources/reference.conf
```diff
@@ -273,13 +285,16 @@ akka.http {
...
+    # Note that it is not necessarily a problem to set this to a high value as all stream operations
+    # are always properly backpressured.
+    # Nevertheless you might want to apply some limit in order to prevent a single client from consuming
+    # an excessive amount of server resources.
+    #
+    # Set to `infinite` to complete disable entity length checks. (Even then you can still apply one
```
2beaucoup: s/complete/completely

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
2beaucoup: Would it make sense to put all the `SyntheticHeader`s in one place?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -629,6 +667,13 @@ final case class Referer(uri: Uri) extends jm.headers.Referer with ModeledHeader
   def getUri: akka.http.javadsl.model.Uri = uri.asJava
 }
 
+object `Timeout-Access` extends ModeledCompanion[`Timeout-Access`]
+final case class `Timeout-Access`(timeoutAccess: akka.http.scaladsl.TimeoutAccess)
```
2beaucoup: Is `TimeoutAccess` ambiguous here or can we drop the namespace?

ktoso:
LGTM AFAICS but needs a rebase :(

drewhk:
Likely conflicts with the move to `Optional`

akka-http-core/src/main/resources/reference.conf
sirthias: Yes, right. Thanks!

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
sirthias: Yeah, I thought about that as well but didn't want to introduce even more disruption with this PR.
And, from the user perspective it's nice to have to remember only a single "rule" on how to find a header:
just look in the `...headers` package.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
sirthias: We could drop the namespace at the expense of another import.
So I opted for saving one line... :)

sirthias:
Ok, applied all feedback and rebased.

ktoso:
Thanks a lot @sirthias! I'll merge after validation confirms we're good here :shipit:

ktoso:
Failure was missing `javadsl.BidiFlow.identity` method... do you have the time to fix that quickly now @sirthias

sirthias:
Thanks, @ktoso, for the offer.
It'd be great, if you could fix the BidiFlow problem...

ktoso:
Added the method in https://github.com/akka/akka/pull/19526

PR #19484, opened 2016-01-16
Corrected several misspellings of the word "losing".

bwmcadams:
Documentation spellings.


ktoso:
Good catch, thanks Brendan :)

PR #19365, opened 2016-01-06
+str #19020 reduce combinator

agolubev:
Ref #19020

Issue #19020, opened 2015-11-26
add reduce combinator


rkuhn:
see wordcount for the obvious need

rkuhn:
when this is fixed the new combinator shall be used in RecipeReduceByKey for Scala and Java

drewhk:
this is as a substreams flattening op? Seems like a nice idea, but needs some care to be deadlock free. Btw, can we carry around the maximum substream count in the SubstreamOps thing? For example if groupBy is called with a maximum width of 5, then subsequent flatteners can exploit that knowledge.

rkuhn:
no, this op is like `fold`

drewhk:
But `ReduceByKey` does flatten the results of fold, this is why I asked.

rkuhn:
the new ReduceByKey:
~~~scala
Flow[String]
  .groupBy(maxWords, identity)
  .map(_ -> 1)
  .reduce((l, r) => (l._1, l._2 + r._2))
  .mergeSubstreams
~~~

drewhk:
but how is this diffeent from fold?

rkuhn:
~~~scala
        Flow[In]
          .groupBy(maximumGroupSize, groupKey)
          .fold(Option.empty[(K, Out)]) {
            case (None, elem) =>
              val key = groupKey(elem)
              Some((key, fold(foldZero(key), elem)))
            case (Some((key, out)), elem) =>
              Some((key, fold(out, elem)))
          }
          .map(_.get)
          .mergeSubstreams
~~~
Lots of added complexity for the unnecessary zero value.

drewhk:
Ah, ok.

agolubev:
Started this

PR #19485, opened 2016-01-16
Agolubev #19020 reduce combinator

agolubev:
Ref #19020 and ex PR #19365

### DIFF
--- a/akka-docs/rst/java/code/docs/stream/javadsl/cookbook/RecipeReduceByKeyTest.java
+++ b/akka-docs/rst/java/code/docs/stream/javadsl/cookbook/RecipeReduceByKeyTest.java
@@ -55,8 +55,10 @@ public class RecipeReduceByKeyTest extends RecipeTest {
         final Source<Pair<String, Integer>, NotUsed> counts = words
             // split the words into separate streams first
           .groupBy(MAXIMUM_DISTINCT_WORDS, i -> i)
+          //transform each element to pair with number of words in it
+          .map(i -> new Pair<>(i, 1))
           // add counting logic to the streams
-          .fold(new Pair<>("", 0), (pair, elem) -> new Pair<>(elem, pair.second() + 1))
+          .reduce((left, right) -> new Pair<>(left.first(), left.second() + right.second()))
           // get a stream of word counts
           .mergeSubstreams();
         //#word-count
@@ -77,17 +79,13 @@ public class RecipeReduceByKeyTest extends RecipeTest {
   static public <In, K, Out> Flow<In, Pair<K, Out>, NotUsed> reduceByKey(
       int maximumGroupSize,
       Function<In, K> groupKey,
-      Function<K, Out> foldZero,
-      Function2<Out, In, Out> fold,
-      Materializer mat) {
+      Function<In, Out> map,
+      Function2<Out, Out, Out> reduce) {
 
     return Flow.<In> create()
-      .groupBy(maximumGroupSize, i -> i)
-      .fold((Pair<K, Out>) null, (pair, elem) -> {
-        final K key = groupKey.apply(elem);
-        if (pair == null) return new Pair<>(key, fold.apply(foldZero.apply(key), elem));
-        else return new Pair<>(key, fold.apply(pair.second(), elem));
-      })
+      .groupBy(maximumGroupSize, groupKey)
+      .map(i -> new Pair<>(groupKey.apply(i), map.apply(i)))
+      .reduce((left, right) -> new Pair<>(left.first(), reduce.apply(left.second(), right.second())))
       .mergeSubstreams();
   }
   //#reduce-by-key-general
@@ -104,9 +102,8 @@ public class RecipeReduceByKeyTest extends RecipeTest {
         Source<Pair<String, Integer>, NotUsed> counts = words.via(reduceByKey(
           MAXIMUM_DISTINCT_WORDS,
           word -> word,
-          key -> 0,
-          (count, elem) -> count + 1,
-          mat));
+          word -> 1,
+          (left, right) -> left + right));
 
         //#reduce-by-key-general2
         final Future<List<Pair<String, Integer>>> f = counts.grouped(10).runWith(Sink.head(), mat);
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -1071,3 +1071,38 @@ private[stream] final class DropWithin[T](timeout: FiniteDuration) extends Simpl
 
   override def toString = "DropWithin"
 }
+
+/**
+ * INTERNAL API
+ */
+private[stream] final class Reduce[T](f: (T, T) ⇒ T) extends SimpleLinearGraphStage[T] {
+  override def initialAttributes: Attributes = DefaultAttributes.reduce
+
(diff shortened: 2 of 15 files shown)


agolubev:
ready for review

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -974,3 +975,38 @@ private[stream] class DropWithin[T](timeout: FiniteDuration) extends SimpleLinea
...
+    var aggregator: T = _
+
+    setHandler(in, new InHandler {
+      override def onPush(): Unit = {
+        aggregator = grab(in)
+        pull(in)
+        setHandler(in, rest)
+      }
+    })
+    val rest = new InHandler {
```
hepin1989: move this inside the upper one will save a pre allocation right?

ktoso: ok to keep here however does not need to be a val, `def` it can be

agolubev: is val taking more time or more memory? stress here it's only one instance.

agolubev: is there some reference materials(talk?) for such performance nit-picks

akka-docs/rst/java/code/docs/stream/javadsl/cookbook/RecipeReduceByKeyTest.java
```diff
@@ -51,12 +51,14 @@ public void work() throws Exception {
...
+
         final Source<Pair<String, Integer>, BoxedUnit> counts = words
-            // split the words into separate streams first
+          // split the words into separate streams first
           .groupBy(MAXIMUM_DISTINCT_WORDS, i -> i)
+          //transform each element to pair with number of words in it
+          .map(i -> new Pair<String, Integer>(i, 1))
           // add counting logic to the streams
-          .fold(new Pair<>("", 0), (pair, elem) -> new Pair<>(elem, pair.second() + 1))
+          .<Pair<String, Integer>> reduce((left, right) -> new Pair<>(left.first(), left.second() + right.second()))
```
rkuhn: This type ascription should not be necessary: why is the reduce method generic?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -704,10 +704,29 @@ trait FlowOps[+Out, +Mat] {
...
+   *
+   * '''Backpressures when''' downstream backpressures
+   *
+   * '''Completes when''' upstream completes
+   *
+   * '''Cancels when''' downstream cancels
+   *
+   * See also [[FlowOps.fold]]
+   */
+  def reduce[T >: Out](f: (T, T) ⇒ T): Repr[T] = via(new Reduce[T](f))
```
rkuhn: This type signature—while consistent with `scala.collection.Seq#reduce`—is surprising: why would reducing be able to change the type of the stream?

@retronym @adriaanm is there a good reason to take this route?

akka-docs/rst/java/code/docs/stream/javadsl/cookbook/RecipeReduceByKeyTest.java
agolubev: I think that this is because Java can not really understand [T >: Out]:  ```def reduce[T >: Out](f: function.Function2[T, T, T]): SubSource[T, Mat] ``` 
follow API works without generic in Java: ```def reduce[T >: Out](f: function.Function2[Out, Out, T]): SubSource[T, Mat] ``` 
However List.reduce is ```def reduce[A1 >: A](op: (A1, A1) ⇒ A1): A1 ```

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: not really. In this case it being a val means an extra reference stored although this instance is not referenced anymore by the stage itself, so it can be a def. Not much of an issue though, but since it is cheap to fix let's make it a def.

drewhk:
OK TO TEST

drewhk:
LGTM otherwise, after figuring out the desired type signature.

agolubev:
build failed with 
```
[info] - must #toStrict with failing source *** FAILED *** (48 milliseconds)
[info]   "toStrict received OnError while draining stream! Accumulated elements: List(1, 2)" did not include substring "Boom!" (StreamTestKitSpec.scala:40)
[info]   at org.scalatest.Matchers$ShouldMethodHelper$.shouldMatcher(Matchers.scala:6231)
[info]   at org.scalatest.Matchers$AnyShouldWrapper.should(Matchers.scala:6265)
[info]   at akka.stream.testkit.StreamTestKitSpec$$anonfun$1$$anonfun$apply$mcV$sp$2.apply$mcV$sp(StreamTestKitSpec.scala:40)
```

rkuhn:
This is known (and fixed in #19487).

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: Stupid idea: let's take the stricter signature and change it if a use-case comes up. AFAIK this change would be binary compatible due to erasure and the new signature would be strictly more general and hence source compatible. Do I miss some dark secret here which is only known by the selected few?

rkuhn: Unfortunately the stricter signature requires us to use `@uncheckedVariance` on `f` if we remove the T type parameter. I guess this is the reason for the signature of `Seq#reduce` as well—and I consider this a shortcoming of Scala’s declaration site variance approach (because it is impossible to prove to the compiler that this is not actually a problem here, although it could potentially be a problem elsewhere; the alternative of saying “shut up, compiler!” is not nice because it disables the check completely).

So, I’d say that we accept this potential broadening in the Scala API but restrict the Java API because there we can express what we actually want. That has the benefit of removing one potentially failing type inference for javac ;-)

drewhk: sounds ok to me. So @agolubev if you fix this I think we can merge this finally!

agolubev: Ok, so I'll switch to:
```def reduce[T >: Out](f: (Out, Out) ⇒ T): Repr[T]``` for Java API

rkuhn: No, please completely remove `T` from that signature, otherwise the benefit for Java users is non-existent.

drewhk:
also, please don't forget to squash this. I am eager to merge this soon :)

rkuhn:
I can take this over (making the last small change) and push over the line—I’d like to have this in today.

drewhk:
grreat!

agolubev:
@rkuhn, I will finalize this shortly if you haven't started it

agolubev:
@rkuhn, PR is updated and rebased.

rkuhn:
Thanks a lot @agolubev !

BTW: do you have a twitter handle?

agolubev:
@rkuhn, my twitter handle is:  alex_golubev

PR #19486, opened 2016-01-16
WIP Sink.seq remaining issues

lolski:
This is https://github.com/akka/akka/pull/19332 but is targeted at master.

Last issue:
- Should we rename `Sink.seq` in Java DSL to `Sink.toList`? Or should both be called `Sink.seq`?

Also, @rkuhn regarding using `tryPull`: I am currently keeping the `pull` call in `SeqStage` as per @drewhk's comment here https://github.com/akka/akka/pull/19332#discussion_r49429246 and here https://github.com/akka/akka/pull/19332#discussion_r49431032


ktoso:
what I would recommend to clean up this PR is:

```
rebase --interactive 03139e0 # squash all your commits into one, we'll want that anyway before merging
log # remember the commit id of that commit
checkout master
checkout -b my-pr
cherry-pick [commit id known since the log above]
push # and PR that branch
```

drewhk:
This needs to be properly rebased against master. I am closing this PR for now, but only to keep our PR queue clean: i.e. we are still eager to have this in! Please reopen once it is ready for final review.

lolski:
Ah I get it now, sorry I got mixed up in how to do it.

PS: I'll get it done next week, currently occupied with work.

PR #19487, opened 2016-01-16
Wip 19014 make stage actor ref location transparent rk

### LINKED ISSUE #19014: StageActorRef is local only currenctly
We decided to keep the `StageActorRef` currently a "local only, special ActorRef".
We can make it work with remoting (register it in the ActorRefProvider), but opted not to do so currently.

This ticket is about making it not "local only".

rkuhn:
This new light-weight ActorRef supports running a non-blocking
side-effect upon message send, which is used to dispatch an async
callback to a GraphStageLogic, or it can be used to make the Akka Typed
adapters more efficient. The FunctionRef is registered with its parent,
and it is not user-level API (hence only accessible by downcasting the
ActorContext).

fixes #19014

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/actor/FunctionRefSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/actor/FunctionRefSpec.scala
@@ -1,0 +1,100 @@
+/**
+ * Copyright (C) 2016 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.actor
+
+import akka.testkit.AkkaSpec
+import akka.testkit.ImplicitSender
+import scala.concurrent.duration._
+import akka.testkit.EventFilter
+
+object FunctionRefSpec {
+
+  case class GetForwarder(replyTo: ActorRef)
+  case class DropForwarder(ref: FunctionRef)
+  case class Forwarded(msg: Any, sender: ActorRef)
+
+  class Super extends Actor {
+    def receive = {
+      case GetForwarder(replyTo) ⇒
+        val cell = context.asInstanceOf[ActorCell]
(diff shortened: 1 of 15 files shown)


akka-actor-tests/src/test/scala/akka/actor/FunctionRefSpec.scala
```diff
@@ -0,0 +1,100 @@
...
+
+  case class GetForwarder(replyTo: ActorRef)
+  case class DropForwarder(ref: FunctionRef)
+  case class Forwarded(msg: Any, sender: ActorRef)
+
+  class Super extends Actor {
+    def receive = {
+      case GetForwarder(replyTo) ⇒
+        val cell = context.asInstanceOf[ActorCell]
+        val ref = cell.addFunctionRef((sender, msg) ⇒ replyTo ! Forwarded(msg, sender))
```
drewhk: nice feature!

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkStage.scala
```diff
@@ -28,7 +28,7 @@ private[akka] class ActorRefBackpressureSinkStage[In](ref: ActorRef, onInitMessa
 
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
     new GraphStageLogic(shape) {
-      implicit var self: StageActorRef = _
+      implicit def self: ActorRef = stageActor.ref
```
ktoso: great :+1:

ktoso:
LGTM, very nice feature (and use in Typed)!

rkuhn:
now with optimized BC …

Issue #19467, opened 2016-01-15
Akka HTTP WebSocket connections reset


jroper:
In our tests, I'm frequently seeing Akka HTTP resetting TCP connections (that is, sending the RST flag mid connection, not sending a FIN flag).

Here's a typical connection, Akka HTTP is on port 19001, the client sends an HTTP request which is a WebSocket request, Akka HTTP sends a response back accepting it, and then Akka HTTP immediately resets the connection (notice the R flag):

```
13:26:42.273419 IP localhost.34368 > localhost.19001: Flags [S], seq 2355686127, win 43690, options [mss 65495,sackOK,TS val 25598901 ecr 0,nop,wscale 7], length 0
E..<qZ@.@.._.........@J9.h...........0.........
............
13:26:42.273439 IP localhost.19001 > localhost.34368: Flags [S.], seq 2135811301, ack 2355686128, win 43690, options [mss 65495,sackOK,TS val 25598901 ecr 25598901,nop,wscale 7], length 0
E..<..@.@.<.........J9.@.M...h.......0.........
............
13:26:42.273452 IP localhost.34368 > localhost.19001: Flags [.], ack 1, win 342, options [nop,nop,TS val 25598901 ecr 25598901], length 0
E..4q[@.@..f.........@J9.h...M.....V.(.....
........
13:26:42.285304 IP localhost.34368 > localhost.19001: Flags [P.], seq 1:207, ack 1, win 342, options [nop,nop,TS val 25598904 ecr 25598901], length 206
E...q\@.@............@J9.h...M.....V.......
........GET /stream HTTP/1.1
Upgrade: websocket
Connection: Upgrade
Sec-WebSocket-Key: Rs0VY7XccMnGDIOrFTPykg==
Host: localhost:19001
Sec-WebSocket-Origin: http://localhost:19001
Sec-WebSocket-Version: 13

13:26:42.285320 IP localhost.19001 > localhost.34368: Flags [.], ack 207, win 350, options [nop,nop,TS val 25598904 ecr 25598904], length 0
E..4.S@.@..n........J9.@.M...h.....^.(.....
........
13:26:42.344631 IP localhost.19001 > localhost.34368: Flags [P.], seq 1:192, ack 207, win 350, options [nop,nop,TS val 25598919 ecr 25598904], length 191
E....T@.@...........J9.@.M...h.....^.......
........HTTP/1.1 101 Switching Protocols
Upgrade: websocket
[...]

PR #19488, opened 2016-01-16
multiple HTTP logic fixes

### LINKED ISSUE #19398: Websocket server leaks inner stream when peer closes connection (without sending close frame first)
As reported in https://groups.google.com/forum/#!topic/akka-user/8tozN2SX--g

This is a regression probably introduced in https://github.com/akka/akka/commit/cdee221b1565f18f809312daf028f8632728bdd3. Completion from `ProtocolSwitchStage.fromNet` triggers a completion of the whole stage. Previously, this also closed `toWS` and cancelled `fromWS`. These were replaced by a dedicated substream in cdee221b1565f18f809312daf028f8632728bdd3 which now needs to be completed/cancelled manually as well when the stage shuts down.

In this case it can probably quite simply be fixed. The question remains whether this is a more general problem with substreams created in a GraphStage. One one hand, it can certainly make sense to keep substreams around longer than the stage that created them but it seems to be prone to leaks like seen here.

### LINKED ISSUE #19451: fix WebsocketIntegrationSpec to use random free port instead of 8080

rkuhn:
fixes #19451 
fixes #19398


rkuhn:
Let’s wait for validation on this one.

rkuhn:
fixes #19467 
fixes #19503

PR #19502, opened 2016-01-18
 #19444 simplify signature of FlowOps.expand

### LINKED ISSUE #19444: simplify the signature of FlowOps.expand
Instead of including an initial mapping step it is easier to use
~~~scala
def expand[T](f: Out => Iterator[T]): Repr[T]
~~~
which allows the most common case to be expressed as `.expand(Iterator.continually)`. It also allows interesting semantic extensions by having the Iterator stop emitting when it runs out of extrapolation power—in which case it waits for the next upstream element before proceeding.

rkuhn:
Any previous usage can be represented in this fashion while typically
saving one function allocation and also saving the function literal’s
syntax overhead (for the seed). Plus a new feature: the provided
iterator does not have to be infinite, limiting how far expand is
allowed to go.

fixes #19444


akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowExpandSpec.scala
```diff
@@ -125,6 +125,26 @@ class FlowExpandSpec extends AkkaSpec {
       publisher.expectRequest()
 
     }
+
+    "work properly with finite extrapolations" in {
+      val (source, sink) =
+        TestSource.probe[Int]
+          .expand(i ⇒ Iterator.from(0).map(i -> _).take(3))
```
drewhk: this is a nice feature

drewhk:
LGTM

johanandren:
Migration docs?

LGTM except for that.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/GraphInterpreterSpecKit.scala
```diff
@@ -298,14 +298,18 @@ trait GraphInterpreterSpecKit extends AkkaSpec {
...
 
-  abstract class OneBoundedSetup[T](ops: Array[GraphStageWithMaterializedValue[Shape, Any]]) extends Builder {
-
-    def this(ops: Iterable[Stage[_, _]]) = {
-      this(ops.map { op ⇒
-        new PushPullGraphStage[Any, Any, Any](
-          (_) ⇒ op.asInstanceOf[Stage[Any, Any]],
-          Attributes.none)
-      }.toArray.asInstanceOf[Array[GraphStageWithMaterializedValue[Shape, Any]]])
+  implicit class ToGraphStage[I, O](stage: Stage[I, O]) {
```
viktorklang: final?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -476,47 +477,54 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
 /**
  * INTERNAL API
  */
-private[akka] final case class Expand[In, Out, Seed](seed: In ⇒ Seed, extrapolate: Seed ⇒ (Out, Seed)) extends DetachedStage[In, Out] {
-  private var s: Seed = _
-  private var started: Boolean = false
-  private var expanded: Boolean = false
+private[akka] final class Expand[In, Out](extrapolate: In ⇒ Iterator[Out]) extends GraphStage[FlowShape[In, Out]] {
+  private val in = Inlet[In]("expand.in")
```
viktorklang: TODO: decide whether these names should be capitalized or not (they currently come in many forms)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -476,47 +477,54 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
 /**
  * INTERNAL API
  */
-private[akka] final case class Expand[In, Out, Seed](seed: In ⇒ Seed, extrapolate: Seed ⇒ (Out, Seed)) extends DetachedStage[In, Out] {
-  private var s: Seed = _
-  private var started: Boolean = false
-  private var expanded: Boolean = false
+private[akka] final class Expand[In, Out](extrapolate: In ⇒ Iterator[Out]) extends GraphStage[FlowShape[In, Out]] {
```
viktorklang: I'm sure I'm ignorant here, but how does this method differ from flatMapConcat now?

rkuhn: `expand` detaches demand and only expands when needed

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/GraphInterpreterSpecKit.scala
rkuhn: why? it’s an implicit conversion in an internal test tool

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: I like capitalized better, at least for me it is easier to parse that way.

drewhk:
Now there are migration docs, so please add a section.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/GraphInterpreterSpecKit.scala
viktorklang: Ah, I got confused by the "Kit" naming

drewhk:
merging after the minor issues were addressed.

rkuhn:
This needs to go in next, after validation has passe.

johanandren:
LGTM!

Issue #18744, opened 2015-10-20
akka-jmh-bench included in 2.4.0 scaladoc


viktorklang:
Why?

http://doc.akka.io/api/akka/2.4.0/?_ga=1.16514806.1579561034.1353497989#akka.persistence.PersistPersistentActor

ktoso:
AFAIR we fixed this after releasing 2.4 will find the fix PR later.

patriknw:
I don't think this is fixed. Snapshot docs also includes benchmarks: http://doc.akka.io/api/akka/snapshot/#akka.persistence.PersistPersistentActor

ktoso:
Will look into it this cleanup-day.

PR #19504, opened 2016-01-18
=pro #18744 Disable generation of javadoc for akka-bench-jmh

johanandren:
Refs #18744

### DIFF
--- a/project/AkkaBuild.scala
+++ b/project/AkkaBuild.scala
@@ -492,6 +492,10 @@ object AkkaBuild extends Build {
     publish := ()
   )
 
+  val dontPublishDocsSettings = Seq(
+    sources in doc in Compile := List()
+  )
+
   override lazy val settings =
     super.settings ++
     buildSettings ++
--- a/akka-bench-jmh/build.sbt
+++ b/akka-bench-jmh/build.sbt
@@ -7,5 +7,5 @@ disablePlugins(Unidoc)
 AkkaBuild.defaultSettings
 
 AkkaBuild.dontPublishSettings
-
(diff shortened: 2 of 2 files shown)


project/AkkaBuild.scala
```diff
@@ -119,6 +119,8 @@ object AkkaBuild extends Build {
       persistence, distributedData,
       testkit
     ).map(_ % "compile;compile->test;provided->provided")
+  ).settings(
+    sources in doc in Compile := List()
```
ktoso: please remove here and in stead in `akka-bench-jmh/build.sbt` add this line: `AkkaBuild.dontPublishSettings`

ktoso:
Valid fix, but not the style we're doing nowadays :)
LGTM after my comment applied

project/AkkaBuild.scala
johanandren: That's already set (and does not work). Hmm.

ktoso: Oh, that might be of interest to @2m

ktoso: Duh I'm so blind, sorry.

ktoso: How about we add to AkkaBuild this:

```scala
  val dontPublishDocsSettings = Seq(
    sources in doc in Compile := List()
  )
```

and that we include in build.snt

johanandren: :+1:

drewhk:
unrelated failure, LGTM

patriknw:
LGTM

Issue #19428, opened 2016-01-13
Enable unidoc for http-core


ktoso:
Disabled to get the build running in: https://github.com/akka/akka/pull/19416
It needs to respect this filtering: `sources in(JavaUnidoc, unidoc) ~= (_.filterNot(_.getPath.contains("Access$minusControl$minusAllow$minusOrigin")))` but doesnt

2m:
Fixed in #19505

PR #19505, opened 2016-01-18
Various streams build improvements

2m:
- remove settings duplication between AkkaBuild and project specific .sbt files
- do not publish test projects
- run akka-http-core tests
- fix lambda sample project versions
- remove obsolete projects (streamAndHttp, httpParent, docsDev)
- exclude parsing project from unidoc
- enable unidoc for http-core #19428
- update to the latest sbt

### DIFF
--- a/akka-docs/_sphinx/exts/includecode2.py
+++ b/akka-docs/_sphinx/exts/includecode2.py
@@ -43,7 +43,7 @@ class IncludeCode2(Directive):
         encoding = self.options.get('encoding', env.config.source_encoding)
         codec_info = codecs.lookup(encoding)
         try:
-            f = codecs.StreamReaderWriter(open(fn, 'U'),
+            f = codecs.StreamReaderWriter(open(fn, 'Ub'),
                     codec_info[2], codec_info[3], 'strict')
             lines = f.readlines()
             f.close()
--- a/akka-http-core/build.sbt
+++ b/akka-http-core/build.sbt
@@ -2,13 +2,8 @@ import akka._
 import com.typesafe.tools.mima.plugin.MimaKeys
 
 AkkaBuild.defaultSettings
-
+AkkaBuild.experimentalSettings
 Formatting.formatSettings
-
 OSGi.httpCore
-
 Dependencies.httpCore
-
 MimaKeys.previousArtifacts := akkaStreamAndHttpPreviousArtifacts("akka-http-core").value
-
-disablePlugins(Unidoc) // TODO remove me
--- a/akka-http/src/main/scala/akka/http/impl/server/RequestContextImpl.scala
+++ b/akka-http/src/main/scala/akka/http/impl/server/RequestContextImpl.scala
@@ -40,7 +40,7 @@ private[http] final case class RequestContextImpl(underlying: ScalaRequestContex
     case MarshallerImpl(m) ⇒
       implicit val marshaller = m(underlying.executionContext)
       underlying.complete(value)
-    case _ ⇒ throw new IllegalArgumentException("Unsupported marshaller: $marshaller")
+    case _ ⇒ throw new IllegalArgumentException(s"Unsupported marshaller: $marshaller")
   }
   def complete(response: jm.HttpResponse): RouteResult = underlying.complete(response.asScala)
 
--- a/project/AkkaBuild.scala
+++ b/project/AkkaBuild.scala
@@ -45,15 +45,6 @@ object AkkaBuild extends Build {
       Protobuf.settings ++ Seq(
       parallelExecution in GlobalScope := System.getProperty("akka.parallelExecution", parallelExecutionByDefault.toString).toBoolean,
       Dist.distExclude := Seq(actorTests.id, docs.id, samples.id, osgi.id),
-
-      // FIXME problem with scalaunidoc:doc, there must be a better way
-      unidocProjectFilter in (ScalaUnidoc, unidoc) := inAnyProject -- inProjects(protobuf, samples,
-        sampleCamelJava, sampleCamelScala, sampleClusterJava, sampleClusterScala, sampleFsmScala, sampleFsmJavaLambda,
-        sampleMainJava, sampleMainScala, sampleMainJavaLambda, sampleMultiNodeScala,
-        samplePersistenceJava, samplePersistenceScala, samplePersistenceJavaLambda,
-        sampleRemoteJava, sampleRemoteScala, sampleSupervisionJavaLambda,
-        sampleDistributedDataScala, sampleDistributedDataJava),
-
       S3.host in S3.upload := "downloads.typesafe.com.s3.amazonaws.com",
       S3.progress in S3.upload := true,
       mappings in S3.upload <<= (Release.releaseDirectory, version) map { (d, v) =>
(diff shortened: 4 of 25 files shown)


drewhk:
nice, LGTM

akka-docs/_sphinx/exts/includecode2.py
```diff
@@ -43,7 +43,7 @@ def run(self):
         encoding = self.options.get('encoding', env.config.source_encoding)
         codec_info = codecs.lookup(encoding)
         try:
-            f = codecs.StreamReaderWriter(open(fn, 'U'),
+            f = codecs.StreamReaderWriter(open(fn, 'Ub'),
```
ktoso: Ub?

akka-http-core/build.sbt
```diff
@@ -2,13 +2,11 @@ import akka._
 import com.typesafe.tools.mima.plugin.MimaKeys
 
 AkkaBuild.defaultSettings
-
+AkkaBuild.experimentalSettings
```
ktoso: I do somewhat enjoy the newlines between them actually...

akka-http/src/main/scala/akka/http/impl/server/RequestContextImpl.scala
```diff
@@ -40,7 +40,7 @@ private[http] final case class RequestContextImpl(underlying: ScalaRequestContex
     case MarshallerImpl(m) ⇒
       implicit val marshaller = m(underlying.executionContext)
       underlying.complete(value)
-    case _ ⇒ throw new IllegalArgumentException("Unsupported marshaller: $marshaller")
+    case _ ⇒ throw new IllegalArgumentException(s"Unsupported marshaller: $marshaller")
```
ktoso: good catch

project/AkkaBuild.scala
```diff
@@ -213,103 +202,34 @@ object AkkaBuild extends Build {
     dependencies = Seq(persistence % "compile;provided->provided;test->test", testkit % "compile;test->test")
   )
 
-  lazy val streamAndHttp = Project(
```
ktoso: please keep this project, I use it a lot. (to test both streams and http with one test command)

project/AkkaBuild.scala
```diff
@@ -67,8 +58,7 @@ object AkkaBuild extends Build {
       slf4j, agent, persistence, persistenceQuery, persistenceTck, kernel, osgi, docs, contrib, samples, multiNodeTestkit, benchJmh, typed, protobuf,
       // streamAndHttp, // does not seem to work
       stream, streamTestkit, streamTests, streamTestsTck,
-      httpCore, http, httpSprayJson, httpXml, httpJackson, httpTests, httpTestkit,
-      docsDev // TODO merge with `docs`
```
ktoso: thanks

project/AkkaBuild.scala
```diff
@@ -489,7 +390,8 @@ object AkkaBuild extends Build {
 
   val dontPublishSettings = Seq(
     publishSigned := (),
-    publish := ()
+    publish := (),
+    publishArtifact in Compile := false
   )
 
   override lazy val settings =
```
ktoso: nice cleanup

project/AkkaBuild.scala
```diff
@@ -213,103 +202,34 @@ object AkkaBuild extends Build {
...
-              new sbt.testing.AnnotatedFingerprint {
-                def annotationName = "org.junit.runner.RunWith"
-                def isModule = false
-              }
-            Seq(new TestDefinition("AllJavaTests", pseudoJUnitRunWithFingerprint, false, Array.empty))
-          },
-          // don't ignore Suites which is the default for the junit-interface
-          testOptions += Tests.Argument(TestFrameworks.JUnit, "--ignore-runners="),
-          mainClass in run in Test := Some("akka.http.javadsl.SimpleServerApp")
-        )
```
2beaucoup: Yeah! :)

ktoso: :+1:

project/Doc.scala
```diff
@@ -120,7 +120,7 @@ object UnidocRoot extends AutoPlugin {
 
   override lazy val projectSettings =
     CliOptions.genjavadocEnabled.ifTrue(scalaJavaUnidocSettings).getOrElse(scalaUnidocSettings) ++
-    settings(Seq(AkkaBuild.samples), Seq(AkkaBuild.remoteTests, AkkaBuild.benchJmh))
+    settings(Seq(AkkaBuild.samples), Seq(AkkaBuild.remoteTests, AkkaBuild.benchJmh, AkkaBuild.parsing, AkkaBuild.protobuf))
```
ktoso: are we sure?

ktoso: cc @rkuhn

rkuhn: If my single-glimpse understanding is correct then this will ignore parsing and protobuf for doc purposes, which should be fine.

akka-docs/_sphinx/exts/includecode2.py
2m: Yes. The added `b` fixes a problem when Python 3 is used when generating the docs.

ktoso: Ok I see, for reference:

```
Note Underlying encoded files are always opened in binary mode. No automatic conversion of '\n' is done on reading and writing. The mode argument may be any binary mode acceptable to the built-in open() function; the 'b' is automatically added.
```

project/Doc.scala
ktoso: Yes, it's excluding. Wasn't sure about parsing. Thanks!

ktoso:
LGTM

project/AkkaBuild.scala
2m: I am not sure it is worth having it and adding additional complexity to the build. Also this project was not solving the problem completely, because one still needs to run `akka-docs/test` and jmh tests separately to see if everything depending on streams compile and run.

ktoso: add docs to this aggregate :)

ktoso: (we really want to keep this, unless you want to make the spray team unhappy)

2m:
Amended the commit by moving

```
sources in(Genjavadoc, doc) ~= (_.filterNot(_.getPath.contains("Access$minusControl$minusAllow$minusOrigin")))
```

out of the `build.sbt` of `akka-http-core` to the `Unidoc` autoplugin. This was required as this setting needs to be wrapped by `UnidocRoot.CliOptions.genjavadocEnabled.ifTrue(...)` which is not quite possible in the `.sbt` file.

Noticed this when I first ran sbt without the `-Dakka.genjavadoc.enabled=true`

project/AkkaBuild.scala
rkuhn: Hmm, not so sure, actually: adding the docs to this aggregate builds everything anyway, and running all HTTP tests is just two commands otherwise. Can this be solved differently? Perhaps adding a task that depends on those tests we want to execute?

rkuhn:
I’d say let’s merge this, and @ktoso can add convenience commands/tasks for aggregating HTTP tasks where needed.

ktoso:
Ok, so FYI we'll go ahead with removing that project and @2m will make it possible to use pr validation locally, it's better because it automatically figures out which projects need re-building :+1:

PR #19507, opened 2016-01-18
+str #16610 adding batch & batchWeighted

hochgi:
main code is written (modelled after the existing `conflate`).
~~still~~-~~need~~-~~to~~-~~figure~~-~~out~~-~~why~~-~~the~~-~~test~~-~~is~~-~~failing.~~
C&C are much appreciated.

### DIFF
--- a/akka-stream-tests/src/test/java/akka/stream/javadsl/FlowTest.java
+++ b/akka-stream-tests/src/test/java/akka/stream/javadsl/FlowTest.java
@@ -534,6 +534,61 @@ public class FlowTest extends StreamTest {
   }
 
   @Test
+  public void mustBeAbleToUseBatch() throws Exception {
+    final JavaTestKit probe = new JavaTestKit(system);
+    final List<String> input = Arrays.asList("A", "B", "C");
+    final Flow<String, String, NotUsed> flow = Flow.of(String.class).batch(3L, new Function<String, String>() {
+      @Override
+      public String apply(String s) throws Exception {
+        return s;
+      }
+    }, new Function2<String, String, String>() {
+      @Override
+      public String apply(String aggr, String in) throws Exception {
+        return aggr + in;
+      }
+    });
+    Future <String> future = Source.from(input).via(flow).runFold("", new Function2<String, String, String>() {
+      @Override
+      public String apply(String aggr, String in) throws Exception {
+        return aggr + in;
+      }
+    }, materializer);
+    String result = Await.result(future, probe.dilated(FiniteDuration.create(3, TimeUnit.SECONDS)));
+    assertEquals("ABC", result);
+  }
+
+  @Test
+  public void mustBeAbleToUseBatchWeighted() throws Exception {
+    final JavaTestKit probe = new JavaTestKit(system);
+    final List<String> input = Arrays.asList("A", "B", "C");
+    final Flow<String, String, NotUsed> flow = Flow.of(String.class).batchWeighted(3L, new Function<String, Object>() {
+      @Override
+      public Object apply(String s) throws Exception {
+        return 1L;
+      }
+    }, new Function<String, String>() {
+      @Override
+      public String apply(String s) throws Exception {
+        return s;
+      }
+    }, new Function2<String, String, String>() {
+      @Override
+      public String apply(String aggr, String in) throws Exception {
+        return aggr + in;
+      }
+    });
+    Future <String> future = Source.from(input).via(flow).runFold("", new Function2<String, String, String>() {
+      @Override
+      public String apply(String aggr, String in) throws Exception {
+        return aggr + in;
+      }
+    }, materializer);
+    String result = Await.result(future, probe.dilated(FiniteDuration.create(3, TimeUnit.SECONDS)));
+    assertEquals("ABC", result);
+  }
+
+  @Test
   public void mustBeAbleToUseExpand() throws Exception {
     final JavaTestKit probe = new JavaTestKit(system);
     final List<String> input = Arrays.asList("A", "B", "C");
--- a/akka-stream/src/main/scala/akka/stream/impl/ConstantFun.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/ConstantFun.scala
@@ -20,4 +20,8 @@ private[akka] object ConstantFun {
   def javaIdentityFunction[T]: JFun[T, T] = JavaIdentityFunction.asInstanceOf[JFun[T, T]]
 
   def scalaIdentityFunction[T]: T ⇒ T = conforms
+
+  def returnZero[T](t: T): Long = 0L
+
+  def returnOne[T](t: T): Long = 1L
 }
--- a/akka-stream/src/main/scala/akka/stream/impl/ConstantFun.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/ConstantFun.scala
@@ -21,7 +21,7 @@ private[akka] object ConstantFun {
 
   def scalaIdentityFunction[T]: T ⇒ T = conforms
 
-  def returnZero[T](t: T): Long = 0L
+  val zeroLong = (_: Any) ⇒ 0L
 
-  def returnOne[T](t: T): Long = 1L
+  val oneLong = (_: Any) ⇒ 1L
 }
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -478,6 +478,9 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
   override def restart(): Conflate[In, Out] = copy()
 }
 
+/**
+ * INTERNAL API
+ */
 private[akka] final case class Batch[In, Out](max: Long, costFn: In ⇒ Long, seed: In ⇒ Out, aggregate: (Out, In) ⇒ Out)
   extends GraphStage[FlowShape[In, Out]] {
 
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -478,27 +478,29 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
   override def restart(): Conflate[In, Out] = copy()
 }
 
-private[akka] sealed abstract class AbstractBatch[In, Out](max: Long, costFn: In ⇒ Long, seed: In ⇒ Out,
-                                                           aggregate: (Out, In) ⇒ Out, val in: Inlet[In],
-                                                           val out: Outlet[Out]) extends GraphStage[FlowShape[In, Out]] {
+private[akka] final case class Batch[In, Out](max: Long, costFn: In ⇒ Long, seed: In ⇒ Out, aggregate: (Out, In) ⇒ Out)
+  extends GraphStage[FlowShape[In, Out]] {
+
+  val in = Inlet[In]("Batch.in")
+  val out = Outlet[Out]("Batch.out")
 
   override val shape: FlowShape[In, Out] = FlowShape.of(in, out)
 
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
-    private var agg: Any = null
+    private var agg: Out = null.asInstanceOf[Out]
     private var left: Long = max
-    private var pending: Any = null
+    private var pending: In = null.asInstanceOf[In]
 
     private def flush(): Unit = {
-      push(out, agg.asInstanceOf[Out])
+      push(out, agg)
       left = max
       if (pending != null) {
-        val elem = pending.asInstanceOf[In]
+        val elem = pending
         agg = seed(elem)
         left -= costFn(elem)
-        pending = null
+        pending = null.asInstanceOf[In]
       } else {
-        agg = null
+        agg = null.asInstanceOf[Out]
       }
     }
 
@@ -516,7 +518,7 @@ private[akka] sealed abstract class AbstractBatch[In, Out](max: Long, costFn: In
           pending = elem
         } else {
           left -= cost
-          agg = aggregate(agg.asInstanceOf[Out], elem)
+          agg = aggregate(agg, elem)
         }
 
         if (isAvailable(out)) flush()
@@ -531,20 +533,16 @@ private[akka] sealed abstract class AbstractBatch[In, Out](max: Long, costFn: In
     setHandler(out, new OutHandler {
 
       override def onPull(): Unit = {
-        //if upstream finished, we still might emit up to 2 more elements (whatever agg is + possibly a pending heavy element)
-        if (isClosed(in)) {
-          if (agg == null) completeStage()
+        if (agg == null) {
+          if (isClosed(in)) completeStage()
+          else if (!hasBeenPulled(in)) pull(in)
+        } else if (isClosed(in)) {
+          push(out, agg)
+          if (pending == null) completeStage()
           else {
-            push(out, agg.asInstanceOf[Out])
-            if (pending == null) completeStage()
-            else {
-              agg = seed(pending.asInstanceOf[In])
-              pending = null
-            }
+            agg = seed(pending)
+            pending = null.asInstanceOf[In]
           }
-        } else if (agg == null) {
-          if (!hasBeenPulled(in))
-            pull(in)
         } else {
           flush()
           if (!hasBeenPulled(in)) pull(in)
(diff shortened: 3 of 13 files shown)


drewhk:
and here is the ticket actually: https://github.com/akka/akka/issues/16610

drewhk:
OK TO TEST

hochgi:
Question for reviewers: is `aggregate` is the best name for the operation?
other alternatives I have thought of were: `accumulate` and `chunk`.
but you are welcome to suggest other names :)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -476,6 +476,86 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
 /**
  * INTERNAL API
  */
+private[akka] final case class AggregateWeighted[In, Out](max: Long, costFn: In ⇒ Long, seed: In ⇒ Out,
+                                                          aggregate: (Out, In) ⇒ Out,
+                                                          decider: Supervision.Decider) extends DetachedStage[In, Out] {
```
drewhk: Can you reimplement this as a `GraphStage`? The DetachedStage will be phased out eventually. The differences are not much (explained in the migration guide with example).

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -476,6 +476,86 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
...
+  private var agg: Any = null
+  private var left: Long = max
+  private var pending: Any = null
+
+  private[this] def flush(ctx: DetachedContext[Out]) = {
```
drewhk: add the return type to the signature.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -476,6 +476,86 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
...
+      agg = aggregate(agg.asInstanceOf[Out], elem)
+    }
+
+    if (!ctx.isHoldingDownstream && pending == null) ctx.pull()
+    else if (!ctx.isHoldingDownstream) ctx.holdUpstream()
+    else flush(ctx)
+  }
+
+  override def onPull(ctx: DetachedContext[Out]): DownstreamDirective = {
+    //if ctx.isFinishing, we still might emit up to 2 more elements (whatever agg is + possibly a pending heavy element)
```
drewhk: this would be much simpler to handle from a `GraphStage`

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -476,6 +476,86 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
...
+
+  private[this] def flush(ctx: DetachedContext[Out]) = {
+    val result = agg.asInstanceOf[Out]
+    agg = null
+    left = max
+    if (pending != null) {
+      val elem = pending.asInstanceOf[In]
+      agg = seed(elem)
+      left -= costFn(elem)
+      pending = null
```
drewhk: In the `GraphStage` version there would be no need for this pending variable as all input ports have a built-in buffer of size one where you need to `grab()` the element from.

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -810,6 +810,58 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
...
+   * duplicate elements.
+   *
+   * '''Emits when''' downstream stops backpressuring and there is an aggregated element available
+   *
+   * '''Backpressures when''' there are `max` aggregated elements and 1 pending element and downstream backpressures
+   *
+   * '''Completes when''' upstream completes and there is no aggregated/pending element waiting
+   *
+   * '''Cancels when''' downstream cancels
+   *
```
drewhk: maybe add a reference to conflate, which is similar, but without an upper limit. Alternative solution is to express conflate in terms of this one, making the cost function constant zero, and making the max Long.MaxValue. Another possible improvement is to make `seed` default to the identity function. @rkuhn you work on something similar with conflate, do you have any other opinion?

drewhk:
I would prefer the name `batch` for this operation. I think we also need one Java smoke test case as well (a simple one, since it delegates to the scala version it is only enough to test it with a single simple case).

drewhk:
Otherwise looks great!

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hochgi: OK, only used it since `Conflate` uses it. Should I also change `Conflate`?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hochgi: it's an internal `private[this]` helper, but OK :)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: I am more thinking about replacing conflate with this one actually.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hochgi: actually I still need to grab it, and compute the element weight, and if it's too heavy, I still would have to store it as pending.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hochgi: :+1:

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: that true.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -473,6 +473,177 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
...
+
+      override def onUpstreamFinish(): Unit = {
+        if (agg == null) completeStage()
+      }
+    })
+
+    setHandler(out, new OutHandler {
+
+      override def onPull(): Unit = {
+        //if upstream finished, we still might emit up to 2 more elements (whatever agg is + possibly a pending heavy element)
```
drewhk: I think you can greatly simplify this code by using the `emit` helper available on GraphStage

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -473,6 +473,177 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
...
+
+//private[akka] final case class Conflate[I, O](seed: I ⇒ O, aggregate: (O, I) ⇒ O)
+//  extends AbstractBatch(Long.MaxValue, { _: I ⇒ 0L }, seed, aggregate, Inlet[I]("Conflate.in"), Outlet[O]("Conflate.out")) {
+//  override def initialAttributes = Attributes.name("Conflate")
+//}
+
+/**
+ * INTERNAL API
+ */
+private[akka] final case class AggregateWeighted[In, Out](max: Long, costFn: In ⇒ Long, seed: In ⇒ Out,
```
rkuhn: I guess this is just dead code, right?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -473,6 +473,177 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
...
+          if (!hasBeenPulled(in)) pull(in)
+        }
+      }
+    })
+  }
+}
+
+private[akka] final case class BatchWeighted[I, O](max: Long, costFn: I ⇒ Long, seed: I ⇒ O, aggregate: (O, I) ⇒ O)
+  extends AbstractBatch(max, costFn, seed, aggregate, Inlet[I]("BatchWeighted.in"), Outlet[O]("BatchWeighted.out")) {
+  override def initialAttributes = Attributes.name("BatchWeighted")
```
rkuhn: the default should go into `akka.stream.impl.Stages.DefaultAttributes`.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -473,6 +473,177 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
...
+}
+
+private[akka] final case class Batch[I, O](max: Long, seed: I ⇒ O, aggregate: (O, I) ⇒ O)
+  extends AbstractBatch(max, { _: I ⇒ 1L }, seed, aggregate, Inlet[I]("Batch.in"), Outlet[O]("Batch.out")) {
```
rkuhn: this function literal should be replaced by a preallocated constant function

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hochgi: yup, forgot to remove it. BTW, regarding `Conflate`, would you want me to also implement it in terms of `Batch`, (see commented out code)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -473,6 +473,177 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
...
+private[akka] sealed abstract class AbstractBatch[In, Out](max: Long, costFn: In ⇒ Long, seed: In ⇒ Out,
+                                                           aggregate: (Out, In) ⇒ Out, val in: Inlet[In],
+                                                           val out: Outlet[Out]) extends GraphStage[FlowShape[In, Out]] {
+
+  override val shape: FlowShape[In, Out] = FlowShape.of(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    private var agg: Any = null
+    private var left: Long = max
+    private var pending: Any = null
```
rkuhn: please use the In and Out types here instead of Any

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -473,6 +473,177 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
...
+          agg = seed(elem)
+        } else if (left < cost) {
+          pending = elem
+        } else {
+          left -= cost
+          agg = aggregate(agg.asInstanceOf[Out], elem)
+        }
+
+        if (isAvailable(out)) flush()
+        if (pending == null) pull(in)
```
rkuhn: this should also check if `left >= 0`, no need to pull when that is false

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
rkuhn: I don’t agree, `emit` by design does not allow the user to react to downstream demand, and this stage needs that.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -473,6 +473,177 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
...
+      override def onUpstreamFinish(): Unit = {
+        if (agg == null) completeStage()
+      }
+    })
+
+    setHandler(out, new OutHandler {
+
+      override def onPull(): Unit = {
+        //if upstream finished, we still might emit up to 2 more elements (whatever agg is + possibly a pending heavy element)
+        if (isClosed(in)) {
```
rkuhn: This would be simpler if it was turned around: first act on `agg` contents, then try to pull if appropriate (which is not try if closed—then terminate).

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -473,6 +473,177 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
   override def restart(): Conflate[In, Out] = copy()
 }
 
+private[akka] sealed abstract class AbstractBatch[In, Out](max: Long, costFn: In ⇒ Long, seed: In ⇒ Out,
+                                                           aggregate: (Out, In) ⇒ Out, val in: Inlet[In],
+                                                           val out: Outlet[Out]) extends GraphStage[FlowShape[In, Out]] {
```
rkuhn: I would prefer this to be a concrete class as the presence of a weight function does not seem important enough to record in the graph AST. The difference is then made just in the API methods.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: the case I mentioned is when upstream is finished, and there is a heavy element pending. That needs two emits, I don't see any need to react to downstream demand apart from emitting those.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -473,6 +473,177 @@ private[akka] final case class Conflate[In, Out](seed: In ⇒ Out, aggregate: (O
...
+      val elem = pending.asInstanceOf[In]
+      agg = seed(elem)
+      left -= costFn(elem)
+      pending = null
+    }
+    ctx.pushAndPull(result)
+  }
+
+  override def onPush(elem: In, ctx: DetachedContext[Out]): UpstreamDirective = {
+    val cost = costFn(elem)
```
rkuhn: For aggregating ByteStrings this is appropriate, but when compressing multiple sets by performing a union it would be more interesting to apply the cost function to the resulting aggregated value. @drewhk WDYT?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -911,11 +911,67 @@ trait FlowOps[+Out, +Mat] {
...
+   * '''Cancels when''' downstream cancels
+   *
+   * See also [[FlowOps.conflate]], [[FlowOps.batch]]
+   *
+   * @param max maximum weight of elements to batch before backpressuring upstream (must be positive non-zero)
+   * @param costFn a function to compute a single element weight
+   * @param seed Provides the first state for a batched value using the first unconsumed element as a start
+   * @param aggregate Takes the currently batched value and the current pending element to produce a new batch
+   */
+  def batchWeighted[S](max: Long, costFn: Out ⇒ Long, seed: Out ⇒ S)(aggregate: (S, Out) ⇒ S): Repr[S] =
```
rkuhn: I think this can fully replace conflate’s use-cases. Maybe we should make `conflate == batchWeighted(1, _ => 0, identity)`?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: that only works if the application is non-destroying or undoable (for example immutable Sets). That somewhat limits that version.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: yep, that is what I thought, too.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hochgi: it's not just the weight function:
- I overridden the `initialAttributes` in the concrete classes.
- If `Conflate` will be implemented in terms of `Batch`, we also need `max`
- most important (since the others could lie in the API function) is the received `Inlet` & `Outlet`. I want to pass these with a good name, and having an abstract class seemed to me the best option

so, should I change it? if so, what names should I supply the `Inlet`&`Outlet` in the abstract class with?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hochgi: actually, AFAICS, if `pending == null`, then `left` is in fact non-negative. and the check is redundant.

rkuhn: no: when agg was null the cost might have been greater than left, and flush() can also leave left negative

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
rkuhn: What I meant with the AST node is that there should be just one class, with one naming of the ports. This would then be Batch, I think, and it would do all the weighting etc. because there is no unweighted implementation.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hochgi: @drewhk are you suggesting something like:

    if (isClosed(in)) {
      if (agg == null)
        completeStage()
      else if (pending == null)
        emit(out, agg.asInstanceOf[Out],() => completeStage())
      else
        emitMultiple(out,List(agg.asInstanceOf[Out],seed(pending.asInstanceOf[In])),() => completeStage())
    }

(never used `emit` before)

rkuhn: I’m not in favor of mixing emit with signal-driven code, that is brittle and hard to review. We might consider doing that later, when we have more time and a better understanding.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hochgi: That's true. but since I can still store a pending element, shouldn't I be eager about it?
Fetching the element at this state would indeed just store it as pending, but I think it might be considered as an optimization especially if the publisher may be slower. WDYT?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hochgi: OK.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: hm, ok, I don't think it is particularly bad, the only issue is if you try to push from outside emit. We might want to add a flag in the interpreter to signal that it is in "auto" mode and throw an exception to the user explaining the issue.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hochgi: even if it's "undoable", aggregate may allocate lots of memory (e.g. `String` concatenation), which would just go to waste if the cost is bigger than the limit.

rkuhn: What we’d actually need is `(Out, S) => (S, Option[S])` so that the user can control all that, but that signature is sooo ugly—and we should definitely provide the normal aggregation logic presented here. We can add the more general combinator later if needed.

drewhk: yes, don't overgeneralize it for now.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hochgi: compiler won't let me :(
(`Out` could be a value type)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hochgi: where do I put it?

rkuhn: in the companion object, I’d say

rkuhn: or in ConstantFun

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
rkuhn: ? value type does not play a role, we have properly typed these fields in all other stages as well

rkuhn: Ah, I see: you need to use `null.asInstanceOf[In]`. Having only that one cast is a lot better than casting the elements all the time.

drewhk: Btw, can't we add something like:

```
def yesIReallyWantANullHere[T]: T = null.asInstanceOf[T]
```

hochgi: Oh... never thought casting a null would be valid.
yeah, I guess you're right :)
(much better than what I was trying: `var agg: Out = _`)

rkuhn: The underscore should work up here, but not further down.

hochgi: it works, but how can I differentiate between uninitialized value and legit values in `AnyVal`s?
e.g:
```
scala> var i: Int = _
i: Int = 0
```

drewhk: I would use a null there explicitly.

rkuhn: in these (generic) cases everything is boxed into reference types anyway, and the underscore is equivalent to `null.asInstanceOn[In]`

drewhk: yes, of course if you have `var i: T = _` then it will be a null.

hochgi: are you sure? I tried the following in the REPL:
```
scala> class A[T] {
     | var a : T = _
     | }
defined class A

scala> new A[Int]
res0: A[Int] = A@123772c4

scala> res0.a
res1: Int = 0
```

rkuhn: This is called “compiler magic” ;-)

akka-stream-tests/src/test/java/akka/stream/javadsl/FlowTest.java
```diff
@@ -533,6 +533,61 @@ public String apply(String aggr, String in) throws Exception {
   }
 
   @Test
+  public void mustBeAbleToUseBatch() throws Exception {
+    final JavaTestKit probe = new JavaTestKit(system);
+    final List<String> input = Arrays.asList("A", "B", "C");
+    final Flow<String, String, BoxedUnit> flow = Flow.of(String.class).batch(3L, new Function<String, String>() {
```
rkuhn: NotUsed is the new BoxedUnit

rkuhn: and we can happily use lambdas now—we just have not rewritten all the existing tests

akka-stream/src/main/scala/akka/stream/impl/ConstantFun.scala
```diff
@@ -20,4 +20,8 @@ private[akka] object ConstantFun {
   def javaIdentityFunction[T]: JFun[T, T] = JavaIdentityFunction.asInstanceOf[JFun[T, T]]
 
   def scalaIdentityFunction[T]: T ⇒ T = conforms
+
+  def returnZero[T](t: T): Long = 0L
+
+  def returnOne[T](t: T): Long = 1L
```
rkuhn: Using these will require one allocation each, what I meant was
~~~scala
val zeroLong = (_: Any) => 0L
val oneLong = (_: Any) => 1L
~~~

rkuhn:
Looking really good now!

rkuhn:
Thanks @hochgi !

PR #19509, opened 2016-01-18
Minor streams touchups

viktorklang:
I found these in my stash.


akka-stream/src/main/scala/akka/stream/Attributes.scala
```diff
@@ -143,6 +143,7 @@ final case class Attributes(attributeList: List[Attributes.Attribute] = Nil) {
       if (i.hasNext)
         i.next() match {
           case Name(n) ⇒
+            // FIXME this URLEncode is a bug IMO, if that format is important then that is how it should be store in Name
```
viktorklang: This is up for discussion, but it seems extremely wasteful to do this again and again and again…

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -792,8 +792,8 @@ trait FlowOps[+Out, +Mat] {
    */
   def groupedWithin(n: Int, d: FiniteDuration): Repr[immutable.Seq[Out]] = {
     require(n > 0, "n must be greater than 0")
-    require(d > Duration.Zero)
-    via(new GroupedWithin[Out](n, d).withAttributes(name("groupedWithin")))
+    require(d > Duration.Zero) // FIXME shouldn't the requirements of GroupedWithin be verified *inside* GroupedWithin?
```
viktorklang: Worth deciding upon

viktorklang:
retest this please

ktoso:
PLS BUILD

akka-stream/src/main/scala/akka/stream/Attributes.scala
rkuhn: good point, I think you are right; ticket?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: yes, and yes

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -854,15 +854,21 @@ private[stream] class GroupedWithin[T](n: Int, d: FiniteDuration) extends GraphS
   }
 }
 
-private[stream] class Delay[T](d: FiniteDuration, strategy: DelayOverflowStrategy) extends SimpleLinearGraphStage[T] {
-
+private[stream] final class Delay[T](d: FiniteDuration, strategy: DelayOverflowStrategy) extends SimpleLinearGraphStage[T] {
+  override def initialAttributes: Attributes = DefaultAttributes.delay
+  private[this] val timerName = "DelayedTimer"
```
rkuhn: maybe put it into the companion object instead

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -854,15 +854,21 @@ private[stream] class GroupedWithin[T](n: Int, d: FiniteDuration) extends GraphS
...
+        case None                        ⇒ throw new IllegalStateException(s"Couldn't find InputBuffer Attribute for $this")
+        case Some(InputBuffer(min, max)) ⇒ max
+      }
+
     val buffer = FixedSizeBuffer[(Long, T)](size) // buffer has pairs timestamp with upstream element
-    val timerName = "DelayedTimer"
     var willStop = false
 
     setHandler(in, handler = new InHandler {
+      //FIXME rewrite into distinct strategies f
```
drewhk: what do you mean here?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: While there are not clear cut answers here, I think it is worth to check most of the things here. My reasoning is that if you check at stage initialization time, you basically allow creating an invalid blueprint that will always result in a faulty stage. Error reports should happen before construction time if the errors are such that would guarantee a faulty blueprint as a result.

rkuhn: No, it should be done in the constructor of GroupedWithin, which happen “here” as well.

drewhk: of course I am wrong since you were talking about `GraphStage` and somehow I thought you are talking about `GraphStageLogic`

drewhk:
LGTM apart from my Delay where I don't understand the comment

akka-stream/src/main/scala/akka/stream/Attributes.scala
viktorklang: Done: https://github.com/akka/akka/issues/19521

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
viktorklang: Technically, making it a def *should* put the string literal in the constant pool for the classfile.
I'll change it to a def.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
viktorklang: I'll clean up the comment, I mean that we shouldn't go through that if-table for every element when we're full. That `DelayOverflowStrategy` won't change.

drewhk: So the polymorphic call is cheaper than the match? Probably so.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: Addressed all these

viktorklang:
@drewhk @rkuhn Updated with latest fixes

drewhk:
ok, merging after validation

rkuhn:
Login failure on jenkins (a5).

Issue #19449, opened 2016-01-14
Create a migration guide for 2.0-2.4


drewhk:
(streams and http of course)

johanandren:
I'll close this and actually filling it with info moves along in other tickets/PRs

PR #19515, opened 2016-01-19
=doc #19449 new migration guide for streams 2.0 to 2.4

johanandren:
Refs #19449 

Just made it empty for now so we have a common file to fill out for the individual PRs


ktoso:
great, thanks

drewhk:
LGTM :)

drewhk:
well, before, we had two such guides, one for Java and one for Scala. What is the approach now?

johanandren:
How about we start out with all-in-one since that is how the other akka migration guides are, and split if it becomes to big a guide?

drewhk:
sounds like a good compromise. Won't expect that much stuff as in the previous one anyway.

Issue #19452, opened 2016-01-14
Add `Flow.auto` and `Sink.contramap`.


aloiscochard:
I would like to suggest a few combinators, I already have a PR ready with those:

Flow.auto (name taken from an other streaming lib I was using, I'm open to a suggestion that fit better the naming scheme of the project)
```
  /**
   * Helper to create `Flow` from a function.
   */
  def auto[A, B](f: A ⇒ B): Flow[A, B, Unit] = apply[A].map(f)
```

Sink.contramap
```
  def contramap[In2](f: In2 ⇒ In): Sink[In2, Mat] = Flow.auto(f).toMat(this) { (_, m) ⇒ m }
```

drewhk:
So the benefit of `auto` is basically type inference?

aloiscochard:
@drewhk yes exactly, forgot to explain the motivation on that one...

I can of course rewrite `contramap` without using `auto` if you feel the latter is unwanted.

ktoso:
"`auto`" though under various different names has come up a number of times already.
Note that it used in isolation the type A needs to be spelled out anyway, as in `Source(a: A => a.toB)`,
yet it's pretty useful when lifting a method: `Source(method _)`

rkuhn:
`contramap` should really be `def prepend[T](f: Flow[T, In]): Sink[T]`, should it not? Our algebra is that of stream combinators, which means that a bare function is a “foreign” element.

drewhk:
At lest we should have `prepend`. Of course `flow.to(sink)` is exactly the same as `sink.prepend(flow)`, but it seems like people looking for this functionality in `Sink` and not `Flow`.

aloiscochard:
@rkuhn good point, I can see both as being useful (I'll probably use more often `contramap`, as I use value classes a lot, that's the original reason for this PR).

I suppose we could then implement `contramap` in term of `prepend` then.

Also the naming `contramap` give intuition for the contravariant functor "interface", where one might intuitively expect working with a "pure" function.

References to some definitions of contramap:
- http://docs.typelevel.org/api/scalaz/nightly/index.html#scalaz.Contravariant
- https://github.com/non/cats/blob/master/core/src/main/scala/cats/functor/Contravariant.scala#L10
- https://dwhjames.github.io/datomisca/api/0.7.x/index.html#datomisca.functional.ContraFunctor
- https://hackage.haskell.org/package/contravariant-1.3.3/docs/Data-Functor-Contravariant.html

[...]

PR #19520, opened 2016-01-19
#19452 Add `Flow.fromFunction`, `Sink#contramap` and `Sink#prepend`.

aloiscochard:
As discussed here: https://github.com/akka/akka/issues/19452

Thanks

/cc @rkuhn

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
@@ -34,6 +34,26 @@ final class Sink[-In, +Mat](private[stream] override val module: Module)
   def contramap[In2](f: In2 ⇒ In): Sink[In2, Mat] = Flow.fromFunction(f).toMat(this)(Keep.right)
 
   /**
+   * Connect this `Sink` to a `Flow`.
+   * This function is equivalent to [[Flow#to]].
+   *
+   * The materialized value of the combined [[Sink]] will be the materialized
+   * value of the given flow (ignoring the current Sink’s value), use
+   * [[Sink#prependMat[Mat2* prependMat]] if a different strategy is needed.
+   */
+  def prepend[In2, Mat2](flow: Flow[In2, In, Mat2]): Sink[In2, Mat2] = flow.to(this)
+
+  /**
+   * Connect this `Sink` to a `Flow`.
+   * This function is equivalent to [[Flow#toMat]].
+   *
+   * The `combine` function is used to compose the materialized values of this sink and that
+   * Flow into the materialized value of the resulting Sink.
+   */
+  def prependMat[In2, Mat2, Mat3](flow: Flow[In2, In, Mat2])(combine: (Mat2, Mat) ⇒ Mat3): Sink[In2, Mat3] =
+    flow.toMat(this)(combine)
+
+  /**
    * Connect this `Sink` to a `Source` and run it. The returned value is the materialized value
    * of the `Source`, e.g. the `Subscriber` of a [[Source#subscriber]].
    */
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
@@ -31,6 +31,8 @@ final class Sink[-In, +Mat](private[stream] override val module: Module)
 
   override val shape: SinkShape[In] = module.shape.asInstanceOf[SinkShape[In]]
 
+  def contramap[In2](f: In2 ⇒ In): Sink[In2, Mat] = Flow.fromFunction(f).toMat(this)(Keep.right)
+
   /**
    * Connect this `Sink` to a `Source` and run it. The returned value is the materialized value
    * of the `Source`, e.g. the `Subscriber` of a [[Source#subscriber]].
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
@@ -284,6 +284,11 @@ object Flow {
   def apply[T]: Flow[T, T, NotUsed] = identity.asInstanceOf[Flow[T, T, NotUsed]]
 
   /**
+   * Helper to create `Flow` from a function.
+   */
+  def fromFunction[A, B](f: A ⇒ B): Flow[A, B, NotUsed] = apply[A].map(f)
+
+  /**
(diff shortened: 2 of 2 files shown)


akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -30,6 +30,10 @@ final class Sink[-In, +Mat](private[stream] override val module: Module)
 
   override val shape: SinkShape[In] = module.shape.asInstanceOf[SinkShape[In]]
 
+  def contramap[In2](f: In2 ⇒ In): Sink[In2, Mat] = Flow.fromFunction(f).toMat(this) { (_, m) ⇒ m }
```
drewhk: Use `Keep.right` as the combiner function.

drewhk: also, a `contramapMat` version is needed.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -30,6 +30,10 @@ final class Sink[-In, +Mat](private[stream] override val module: Module)
...
+
+  def prepend[In2, Mat2](flow: Flow[In2, In, Mat2]): Sink[In2, Mat2] = flow.to(this)
```
drewhk: Needs a `prependMat`

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
aloiscochard: @drewhk how would the implementation of  `contramapMat` differ from the actual one? as we are combining a pure function, I can't see why one would like to give something different than `Keep.right` in this case, am I missing something?

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
aloiscochard: Makes sense.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
drewhk: eh, sorry, you are right :)

viktorklang: I'm on the fence for this one. It is a slippery slope to add N contra methods. More choice to do the same thing (vs prepend) is not automatically better. Thoughts?

rkuhn: This discussion has been had on the ticket already.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
drewhk: Please add docs as well, and a reference that this is equivalent to `Flow.to(thisSink)` if someone prefers the strict left-to-right reading.

aloiscochard:
@rkuhn @drewhk I have added `prependMat`, there is also some documentation referring to equivalent `Flow` function.

Thanks to take an other look.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -31,6 +31,28 @@ final class Sink[-In, +Mat](private[stream] override val module: Module)
 
   override val shape: SinkShape[In] = module.shape.asInstanceOf[SinkShape[In]]
 
+  def contramap[In2](f: In2 ⇒ In): Sink[In2, Mat] = Flow.fromFunction(f).toMat(this)(Keep.right)
```
rkuhn: This one needs docs (see `FlowOps#map` for inspiration).

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -31,6 +31,28 @@ final class Sink[-In, +Mat](private[stream] override val module: Module)
...
+
+  /**
+   * Connect this `Sink` to a `Flow`.
+   * This function is equivalent to [[Flow#to]].
+   *
+   * The materialized value of the combined [[Sink]] will be the materialized
+   * value of the given flow (ignoring the current Sink’s value), use
+   * [[Sink#prependMat[Mat2* prependMat]] if a different strategy is needed.
```
rkuhn: These links do not work in JavaDoc after genjavadoc translation, please just use backticks for a code snippet.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -31,6 +31,28 @@ final class Sink[-In, +Mat](private[stream] override val module: Module)
...
+   */
+  def prepend[In2, Mat2](flow: Flow[In2, In, Mat2]): Sink[In2, Mat2] = flow.to(this)
```
rkuhn: I’d vote to invert the materialized value selection on the grounds that the non-Mat methods generally do not change the Mat parameter.

rkuhn:
Once we agree on the signatures this will need to be replicated such that Dsl*ConsistencySpec suites are happy.

Thanks @aloiscochard !

drewhk:
There are a few things missing @aloiscochard but not much. Do you want us to take this PR over and finish?

aloiscochard:
@drewhk I would feel guilty leaving just the documentation to you, but I'm very busy lately so I would appreciate some help :-)

In any case if you don't have time for it, I'll get back to it at some point... just not sure when :-/ sorry for that

drewhk:
Ok, we will take over then.

drewhk:
I am taking this over now

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
drewhk: Hm. What sould 
```
sink.prepend(flow)(Keep.right)
```
do? What does `right` refer to, code order, or the generally imagined "Sink is on the right" rule?

drewhk: well, we can't refer to code order, since `flow.toMat()` expects it in the reverse order. Unless prepend is not expressed in terms of `toMat` and I implement it by working on the module level directly.

In general though, I am less and less convinced that we should add this operator. I vote for cancelling this.

drewhk:
Migrated to new PR: https://github.com/akka/akka/pull/19734

PR #19524, opened 2016-01-19
Option[HttpsContext] => ConnectionContext and "WebSocket"

ktoso:
This will fail on missing dependency until https://github.com/typesafehub/ssl-config released, please have a look at https://github.com/typesafehub/ssl-config/pull/29 (though tests pass locally)

cc @sirthias tired today now, so perhaps I missed some detail still but that's the general direction.
Turns out "removing" HttpsContext entirely wasn't as good since we want to carry along the config values anyway from API to stream.io (and allow users overriding). I have however moved away from the `Option[HttpsContext]` style, and now it's more explicit who _needs_ `HttpsConnectionContext` and also who _can_ take either Http or Https (which is the server - and based on what it got, that's what it starts up as).


ktoso:
Here are my notes on who needs wat:
### Server

Needs to know upfront.

**bind / bindAndHandle** 
  - has context
  - default HTTP
  - if no port given, based on Context 80/443

=> Type: ConnectionContext - based on type HTTP / HTTPS
Note: context should be obtainable Http().defaultServerHttpsContext

### Client

## connections
Needs to know upfront.

**outgoingConnection**
  - no context 

**outgoingConnectionTls**
  - needs https context
  - provides default HTTPS

**outgoingConnection**
  - no context

**newHostConnectionPoolTls**
  - needs https context

=> Tls methods provide default HTTPS config
Type: HttpsConnectionConfig on Tls methods

## request sensitive (adds TLS when needed):
Needs context "just in case", enables when request needs it.

**singleRequest**
  - has context, default HTTPS, may drop it

**singleWebsocketRequest**
  - has context, default HTTPS, may drop it

**singleWebSocketRequest**
  - needs context, "just in case"
  - provides default HTTPS

=> normal methods, Tls methods
=> Tls methods provide default HTTPS config
Type: HttpsConnectionConfig on Tls methods

ktoso:
I also added more overloads and made them consistent for Java, t'was a bit painful but hopefully now nicer to use for Java users.

ktoso:
cc @sirthias @jrudolph @rkuhn please review if you have a sec. You may want to just look at the 2nd commit, as first just fixes consistent casing of the word "WebSocket" (as per spec)

ktoso:
Rebased.

PR #19465, opened 2016-01-14
[WIP] #16651 Add formFieldMap, formFieldMultiMap & formFieldSeq Directives

wjur:
This PR includes formFieldMap, formFieldMultiMap and formFieldSeq Directives. Fixes #16651.

The code lacks of Java API implementation and documentation for that API.
@ktoso, Java API is missing many other directives. Should I work on Java from fields directives or it is enough to polish Scala-related stuff?

(This is a continuation of PR #19042)

PR #19525, opened 2016-01-19
=doc Move formField directives' docs to a correct directory

wjur:
It seems that I placed docs in an incorrect directory in #19465 (maybe the directory was renamed when moving from 2.3-dev to master and I didn't notice that)
Because of that mistake, the current docs are broken for some of formField directives. See: http://doc.akka.io/docs/akka/snapshot/scala/http/routing-dsl/directives/alphabetically.html There is 'formFieldSeq' directive but the link doesn't work.

I see that the docs are in akka-docs instead of akka-docs-dev. In this PR I just simply move some files between the two directories. If I'm still wrong, correct me, please.

I'm sorry for this, @ktoso @rkuhn


ktoso:
Ah right, yes indeed we didn't catch that.
Yes, the docs moved during the mvoe to master.

Thanks for noticing this!


for reference this commit does:
`diff --git 
a/akka-docs-dev/rst/scala/http/routing-dsl/directives/form-field-directives/formFieldMap.rst 
b/akka-docs-dev/rst/scala/http/routing-dsl/directives/form-field-directives/formFieldMap.rst
`

PR #19527, opened 2016-01-20
#19517 publish akka parsing

### LINKED ISSUE #19517: figure out whether akka-parsing needs to be published or removed as dependency
seems like something broke when moving to master

2m:
This was an oversight. akka-parsing was [being published](http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22com.typesafe.akka%22%20AND%20a%3A%22akka-parsing-experimental_2.11%22) in the original build and should continue being so.

Fixes #19517


ktoso:
LGTM, good catch before 2.4.2

Issue #18203, opened 2015-08-13
[akka-http] Cancelling response entity stream with Sink.cancelled doesn't work


kodemaniak:
I am performing a series of HTTP requests and am only interested in the status code of the response. Cancelling the response entity data stream with Sink.cancelled doesnt seem to work, as the connections are exhausted after `max-connections` requests. Example:

```scala
  val future2 = Source(1 to 15)
    .map(i => RequestBuilding.Get("http://akka.io") -> i)
    .via(Http(system).superPool[Int]())
    .map(response => response._1.get.entity.dataBytes.runWith(Sink.cancelled))
    .runForeach(println)
```

Explicitly consuming the stream, however, does work:

```scala
  val future1 = Source(1 to 15)
    .map(i => RequestBuilding.Get("http://akka.io") -> i)
    .via(Http(system).superPool[Int]())
    .map(response => response._1.get.entity.toStrict(1.second).map(_ => response))
    .mapAsync(5)(identity)
    .runForeach(println)
```

jypma:
I'm seeing something similar when prematurely finishing an HTTP response from a downstream `PushStage`, it seems to stop further processing on that connection.

Might this be related to https://github.com/akka/akka/issues/18189 ?

jrudolph:
@kodemaniak what would you expect the behavior to be when the entity data stream is cancelled? Akka-http could either 
 - close / abort the connection
 - automatically consume and ignore the entity stream and then put the connection back into the pool

Consuming the stream means pulling it completely through the network which is something you may want to avoid.

That said, I agree that the connection slot should never be blocked.

Btw. if you are only interested in the status code, there's the HTTP's `HEAD` method which is meant to be the solution for exactly that case (as long as we are talking about the original requests being GET requests).

kodemaniak:
@jrudolph The concrete use case I have is a batch importer that reads a bunch of files, parses them and updates data via REST. However, there might be data present for which there is no corresponding resource available. Since I don't want to check existence of the resource before every request (and quite a few of those are happening), I just try a PUT and remember if it was successful or not. So HEAD doesn't help me here. Now, in the success case the response body is empty (always), but in the error case the REST API returns some JSON with details. However, in my batch process I am not interested in that JSON. Therefore, I basically need a way to say give me the status and ignore the entity.

[...]

Issue #18010, opened 2015-07-16
akka-http: Cancelling an HttpResponse's entity stream should close the TCP connection


mspielberg:
In 1.0-RC4, if a client cancels the entity Source of an HttpResponse (e.g. with take(), or .runWith(Sink.cancelled),), akka-http continues to consume the response from the network, discarding the content, and continues to use the connection for future requests. This can be very expensive for a large response body that the client decides early it is not interested in consuming.

#17403 addresses clients not consuming the response entity at all. This ticket is intended to address the performance problems of using Sink.ignore when the response entity may be very large.

This may stem from the use of splitWhen in OutgoingConnectionBlueprint, which isolates cancellation of the substreams from cancellation of the overarching request stream. If I understand correctly #16870 originally looked to add configuration for how substream cancellation is propagated, but was closed as invalid in favor of another approach not described in that issue.

rkuhn:
duplicate of #18203?

drewhk:
This is because the TCP stream implementation treats cancel() as ignore. This is necessary because:
 - There is no protocol action at the TCP level to "stop reading"
 - if you treat cancel as simply not reading, compared to reading but ignoring it, then you will never get the completion events from NIO.

The only closing action that actually translates to TCP connetion close events are the closing of the write side.

drewhk:
there needs to be some element in the HTTP pipeline then that captures upstream cancellation and turns into downstream completion if this is a desired functionality.

mspielberg:
If it helps, this is a blocker for my organization to adopt akka-http. Our use case involves multi-GB HTTP transfers where cancellation is common (e.g. streaming video). With the current behavior we're seeing network interfaces getting saturated with data that is immediately discarded.

To be clear, the desired behavior is that cancelling the response entity should result in a RST being sent to the server and the entire TCP connection being torn down. If translating cancellation of the write side into completing the read side results in a FIN to the server, this is insufficient since the server may continue sending data into the half-closed TCP connection.

drewhk:
@mspielberg 

> To be clear, the desired behavior is that cancelling the response entity should result in a RST being sent to the server and the entire TCP connection being torn down.

It is not a good idea to map `cancel()` to RST because a Flow cancellation and completion can be in a race to reach the TCP connection flow, which would mean that a normal completion would nondeterministically turn into an abort or half-close then abort

[...]

PR #19528, opened 2016-01-20
Http: HttpEntity stream cancellation

ataraxer:
Here's an attempt to solve issues #18010 and #18203.

First commit can be treated as a simpler version of the solution, which involves adding `eagerClose` flag to `splitWhen` in order to define how substream cancelation should be treated: by draining the remaining elements in order to prevent stalling (current behaviour) or cancelling the stage altogether.

Second one introduces custom stage instead of `splitWhen/prefixAndTail` pipeline, which would allow to add `eagerClose` flag to `outgoingConnection` without adding flag to `splitWhen`/`splitAt` if this change isn't plausible.

I used https://github.com/akka/akka/commit/ced5aa7ddcac1408e08ed159da9a04fecfad18ed as a reference, however there are some slight changes which will be marked in comments to particular lines.

### DIFF
--- a/akka-docs/rst/java/http/client-side/connection-level.rst
+++ b/akka-docs/rst/java/http/client-side/connection-level.rst
@@ -53,6 +53,10 @@ The connection can also be closed by the server.
 An application can actively trigger the closing of the connection by completing the request stream. In this case the
 underlying TCP connection will be closed when the last pending response has been received.
 
+The connection will also be closed if the response entity is cancelled (e.g. by attaching it to ``Sink.cancelled()``)
+or consumed only partially (e.g. by using ``take`` combinator). In order to prevent this behaviour the entity should be
+explicitly drained by attaching it to ``Sink.ignore()``.
+
 
 Timeouts
 --------
--- a/akka-docs/rst/scala/http/client-side/connection-level.rst
+++ b/akka-docs/rst/scala/http/client-side/connection-level.rst
@@ -55,6 +55,10 @@ The connection can also be closed by the server.
 An application can actively trigger the closing of the connection by completing the request stream. In this case the
 underlying TCP connection will be closed when the last pending response has been received.
 
+The connection will also be closed if the response entity is cancelled (e.g. by attaching it to ``Sink.cancelled``)
+or consumed only partially (e.g. by using ``take`` combinator). In order to prevent this behaviour the entity should be
+explicitly drained by attaching it to ``Sink.ignore``.
+
 
 Timeouts
 --------
--- a/akka-docs/rst/scala/stream/migration-guide-2.0-2.4-scala.rst
+++ b/akka-docs/rst/scala/stream/migration-guide-2.0-2.4-scala.rst
@@ -102,6 +102,19 @@ and were accessible via ``settings``. We now made it possible to configure the p
 settings as well, so ``RoutingSettings`` is now ``routingSettings`` and ``ParserSettings`` is
 now accessible via ``parserSettings``.
 
+Client / server behaviour on cancelled entity
+---------------------------------------------
+
+Previously if request or response were cancelled or consumed only partially
+(e.g. by using ``take`` combinator) the remaining data was silently drained to prevent stalling
+the connection, since there could still be more requests / responses incoming. Now the default
+behaviour is to close the connection in order to prevent using excessive resource usage in case
+of huge entities.
+
+The old behaviour can be achieved by explicitly draining the entity:
+
+   response.entity.dataBytes.runWith(Sink.ignore)
+
 Changed Sources / Sinks
 =======================
 
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
@@ -5,7 +5,7 @@
 package akka.http.impl.engine.client
 
 import akka.NotUsed
-import akka.http.scaladsl.settings.ClientConnectionSettings
+import akka.http.scaladsl.settings.{ ClientConnectionSettings, ParserSettings }
 import language.existentials
 import scala.annotation.tailrec
 import scala.collection.mutable.ListBuffer
@@ -16,13 +16,13 @@ import akka.stream._
 import akka.stream.scaladsl._
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.model.headers.Host
-import akka.http.scaladsl.model.{ IllegalResponseException, HttpMethod, HttpRequest, HttpResponse }
+import akka.http.scaladsl.model.{ IllegalResponseException, HttpMethod, HttpRequest, HttpResponse, ResponseEntity }
 import akka.http.impl.engine.rendering.{ RequestRenderingContext, HttpRequestRendererFactory }
 import akka.http.impl.engine.parsing._
 import akka.http.impl.util._
 import akka.stream.stage.GraphStage
 import akka.stream.stage.GraphStageLogic
-import akka.stream.stage.InHandler
+import akka.stream.stage.{ InHandler, OutHandler }
 import akka.stream.impl.fusing.SubSource
 
 /**
@@ -69,24 +69,7 @@ private[http] object OutgoingConnectionBlueprint {
     import ParserOutput._
     val responsePrep = Flow[List[ResponseOutput]]
       .mapConcat(conforms)
-      .splitWhen(x ⇒ x.isInstanceOf[MessageStart] || x == MessageEnd)
-      .prefixAndTail(1)
-      .filter {
-        case (Seq(MessageEnd), remaining) ⇒
-          SubSource.kill(remaining)
-          false
-        case (seq, _) ⇒
-          seq.nonEmpty
-      }
-      .map {
-        case (Seq(ResponseStart(statusCode, protocol, headers, createEntity, _)), entityParts) ⇒
-          val entity = createEntity(entityParts) withSizeLimit parserSettings.maxContentLength
-          HttpResponse(statusCode, headers, entity, protocol)
-        case (Seq(MessageStartError(_, info)), tail) ⇒
-          // Tails can be empty, but still need one pull to figure that out -- never drop tails.
-          SubSource.kill(tail)
-          throw IllegalResponseException(info)
-      }.concatSubstreams
+      .via(new ResponsePrep(parserSettings))
 
     val core = BidiFlow.fromGraph(GraphDSL.create() { implicit b ⇒
       import GraphDSL.Implicits._
@@ -147,6 +130,95 @@ private[http] object OutgoingConnectionBlueprint {
   }
 
   import ParserOutput._
+
+  private final class ResponsePrep(parserSettings: ParserSettings)
+    extends GraphStage[FlowShape[ResponseOutput, HttpResponse]] {
+
+    private val in = Inlet[ResponseOutput]("ResponsePrep.in")
+    private val out = Outlet[HttpResponse]("ResponsePrep.out")
+
+    val shape = new FlowShape(in, out)
+
+    override def createLogic(effectiveAttributes: Attributes) = new GraphStageLogic(shape) with InHandler with OutHandler {
+      private var entitySource: SubSourceOutlet[ResponseOutput] = _
+      private def entitySubstreamStarted = entitySource ne null
+      private def idle = this
+
+      def setIdleHandlers(): Unit = {
+        setHandler(in, idle)
+        setHandler(out, idle)
+      }
+
+      def onPush(): Unit = grab(in) match {
+        case ResponseStart(statusCode, protocol, headers, entityCreator, _) ⇒
+          val entity = createEntity(entityCreator) withSizeLimit parserSettings.maxContentLength
+          push(out, HttpResponse(statusCode, headers, entity, protocol))
+
+        case MessageStartError(_, info) ⇒
+          throw IllegalResponseException(info)
+
+        case other ⇒
+          throw new IllegalStateException(s"ResponseStart expected but $other received.")
+      }
+
+      def onPull(): Unit = {
+        if (!entitySubstreamStarted) pull(in)
+      }
+
+      setIdleHandlers()
+
+      private lazy val waitForMessageEnd = new InHandler {
+        def onPush(): Unit = grab(in) match {
+          case MessageEnd ⇒ setHandler(in, idle)
+          case other      ⇒ throw new IllegalStateException(s"MessageEnd expected but $other received.")
+        }
+      }
+
+      private lazy val substreamHandler = new InHandler with OutHandler {
+        override def onPush(): Unit = grab(in) match {
+          case MessageEnd ⇒
+            entitySource.complete()
+            entitySource = null
+            setIdleHandlers()
+
+          case messagePart ⇒
+            entitySource.push(messagePart)
+        }
+
+        override def onPull(): Unit = pull(in)
+
+        override def onUpstreamFinish(): Unit = {
+          entitySource.complete()
+          completeStage()
+        }
+
+        override def onUpstreamFailure(reason: Throwable): Unit = {
+          entitySource.fail(reason)
+          failStage(reason)
+        }
+
+        override def onDownstreamFinish(): Unit = {
+          entitySource.complete()
+          completeStage()
+        }
+      }
+
+      private def createEntity(creator: EntityCreator[ResponseOutput, ResponseEntity]): ResponseEntity = {
+        creator match {
+          case StrictEntityCreator(entity) ⇒
+            pull(in)
+            setHandler(in, waitForMessageEnd)
+            entity
+
+          case StreamedEntityCreator(creator) ⇒
+            entitySource = new SubSourceOutlet[ResponseOutput]("EntitySource")
+            entitySource.setHandler(substreamHandler)
+            setHandler(in, substreamHandler)
+            creator(Source.fromGraph(entitySource.source))
+        }
+      }
+    }
+  }
 
   /**
    * A merge that follows this logic:
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -104,10 +104,21 @@ private[http] object HttpServerBluePrint {
       val remoteAddress = inheritedAttributes.get[HttpAttributes.RemoteAddress].flatMap(_.address)
       var downstreamPullWaiting = false
       var completionDeferred = false
+      var entitySource: SubSourceOutlet[RequestOutput] = _
 
       // optimization: to avoid allocations the "idle" case in and out handlers are put directly on the GraphStageLogic itself
       override def onPull(): Unit = {
         pull(in)
+      }
+
+      // optimization: this callback is used to handle entity substream cancellation to avoid allocating a dedicated handler
+      override def onDownstreamFinish(): Unit = {
+        if (entitySource ne null) {
+          // application layer has cancelled or only partially consumed response entity:
+          // connection will be closed
+          entitySource.complete()
+          completeStage()
+        }
       }
 
       override def onPush(): Unit = grab(in) match {
@@ -126,7 +137,7 @@ private[http] object HttpServerBluePrint {
 
       setIdleHandlers()
 
-      def setIdleHandlers() {
+      def setIdleHandlers(): Unit = {
         if (completionDeferred) {
           completeStage()
         } else {
@@ -150,15 +161,17 @@ private[http] object HttpServerBluePrint {
         // stream incoming chunks into the request entity until we reach the end of it
         // and then toggle back to "idle"
 
-        val entitySource = new SubSourceOutlet[RequestOutput]("EntitySource")
+        entitySource = new SubSourceOutlet[RequestOutput]("EntitySource")
         // optimization: re-use the idle outHandler
         entitySource.setHandler(this)
 
-        setHandler(in, new InHandler {
+        // optimization: handlers are combined to reduce allocations
+        val chunkedRequestHandler = new InHandler with OutHandler {
           def onPush(): Unit = {
             grab(in) match {
               case MessageEnd ⇒
                 entitySource.complete()
+                entitySource = null
                 setIdleHandlers()
 
               case x ⇒ entitySource.push(x)
@@ -172,8 +185,6 @@ private[http] object HttpServerBluePrint {
             entitySource.fail(ex)
             failStage(ex)
           }
-        })
-        setHandler(out, new OutHandler {
           override def onPull(): Unit = {
             // remember this until we are done with the chunked entity
             // so can pull downstream then
@@ -185,7 +196,10 @@ private[http] object HttpServerBluePrint {
             // when it completes complete the stage
             completionDeferred = true
           }
-        })
+        }
+
+        setHandler(in, chunkedRequestHandler)
+        setHandler(out, chunkedRequestHandler)
         creator(Source.fromGraph(entitySource.source))
       }
 
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -217,8 +217,11 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
                               log: LoggingAdapter = system.log): Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]] =
     _outgoingConnection(host, port, localAddress, settings, connectionContext, log)
 
-  private def _outgoingConnection(host: String, port: Int, localAddress: Option[InetSocketAddress],
-                                  settings: ClientConnectionSettings, connectionContext: ConnectionContext,
+  private def _outgoingConnection(host: String,
+                                  port: Int,
+                                  localAddress: Option[InetSocketAddress],
+                                  settings: ClientConnectionSettings,
+                                  connectionContext: ConnectionContext,
                                   log: LoggingAdapter): Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]] = {
     val hostHeader = if (port == connectionContext.defaultPort) Host(host) else Host(host, port)
     val layer = clientLayer(hostHeader, settings, log)
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
@@ -214,20 +214,25 @@ object Split {
   /** Splits after the current element. The current element will be the last element in the current substream. */
   case object SplitAfter extends SplitDecision
 
-  def when[T](p: T ⇒ Boolean, propagateSubstreamCancel: Boolean = false): Graph[FlowShape[T, Source[T, NotUsed]], NotUsed] =
-    new Split(Split.SplitBefore, p, propagateSubstreamCancel)
+  def when[T](p: T ⇒ Boolean, substreamCancelStrategy: SubstreamCancelStrategy): Graph[FlowShape[T, Source[T, NotUsed]], NotUsed] =
+    new Split(Split.SplitBefore, p, substreamCancelStrategy)
 
-  def after[T](p: T ⇒ Boolean, propagateSubstreamCancel: Boolean = false): Graph[FlowShape[T, Source[T, NotUsed]], NotUsed] =
-    new Split(Split.SplitAfter, p, propagateSubstreamCancel)
+  def after[T](p: T ⇒ Boolean, substreamCancelStrategy: SubstreamCancelStrategy): Graph[FlowShape[T, Source[T, NotUsed]], NotUsed] =
+    new Split(Split.SplitAfter, p, substreamCancelStrategy)
 }
 
 /**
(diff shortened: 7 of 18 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -149,6 +133,68 @@ private[http] object OutgoingConnectionBlueprint {
...
+
+              override def onUpstreamFinish(): Unit = completeStage()
+            })
+
+            creator(Source.fromGraph(entitySource.source))
+        }
+      }
+
+      override def preStart(): Unit = pull(in)
+      setHandler(out, eagerTerminateOutput)
```
ataraxer: https://github.com/akka/akka/commit/ced5aa7ddcac1408e08ed159da9a04fecfad18ed#diff-623ef36775342f2db6b86855f7ffea64R122 uses custom `OutHandler` which just pulls the input, however while testing I've discovered that this leads to port being pulled twice:

```
[ERROR] [01/20/2016 00:58:08.384] [http-streaming-akka.actor.default-dispatcher-4] [ActorSystem(http-streaming)] Outgoing request stream error
java.lang.IllegalArgumentException: requirement failed: Cannot pull port twice
 at scala.Predef$.require(Predef.scala:233)
 at akka.stream.stage.GraphStageLogic.pull(GraphStage.scala:420)
 at akka.http.impl.engine.client.OutgoingConnectionBlueprint$ResponsePrep$$anon$3$$anon$5.onPull(OutgoingConnectionBlueprint.scala:185)
 ...
```

If I understand it correctly this happens because of handler which is set up on `SubSourceOutlet` for entity, which also produces pull's from the same stage.

drewhk:
OK TO TEST

ktoso:
OK TO TEST

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
drewhk: that might be right, yes

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -346,9 +349,12 @@ final class Split[T](decision: Split.SplitDecision, p: T ⇒ Boolean) extends Gr
             closeThis(handler, elem)
             handOver(handler)
           } else {
-            // Drain into the void
-            if (substreamCancelled) pull(in)
-            else substreamSource.push(elem)
+            if (substreamCancelled) {
```
drewhk: is it actually what you want? I though you want to cancel the upstream and complete possible substreams if the master stream is cancelled. Currently it cancels the upstream and everything when a substream is cancelled. But AFAIK substreams *are* the entitiy streams already, so if someone gets to cancel them that means they are properly consumed.

ataraxer: Yes, this is precisely the case that was discussed in both https://github.com/akka/akka/issues/18010:

> To be clear, the desired behavior is that cancelling the response entity should result in a RST being sent to the server and the entire TCP connection being torn down.

and https://github.com/akka/akka/issues/18203:

> In general, I would expect Sink.cancelled to at least free the connection. With respect to your question whether it should abort the connection or just silently consume the entity, I think that probably it would be nice when we could save sending the response over TCP at all, but I have no idea what that means in terms of HTTP (maybe an error on the server side?). Silently consuming the entity on the other hand would also be unexpected when I cancel something. So I would argue that it should close/abort the connection, without knowing the implications for things like keep-alive or pipelining.

To clarify things a bit: I'm not working in the same company as @mspielberg, but we have a very similar use case and cancelling TCP is precisely what we need when response entity is cancelled (e.g. by running with `Sink.cancelled`) or only consumer only partially (e.g. via `take(n)`).

In our case we have an infinite response stream for consuming data from message queue service. This request also acquires a lock on the server side (this has to do with at least once delivery guarantees), which is automatically released when connection is closed. I wrote a client for this service in akka-http which wraps underlying HTTP protocol and exposes data stream in variety of ways for different use cases, the simplest one being a simple `Source[Message]`. The user may choose to consume only a limited amount of messages from this source, which will translate into entity of an underlying chunked `HttpResponse` being cancelled after only required amount of messages has been consumed. The problem is that current implementation keeps draining chunks  in order to prevent connection stalling, since all requests are occupying the same connection, but since the stream is infinite we end up just reading data into /dev/null consuming excessive resources (not good) and also not releasing the lock on the stream and thus preventing anyone else from accessing it (very-very bad).

ataraxer:
I've updated the `ResponsePrep` stage -- now it should pass all tests but one:
```
ConnectionPoolSpec:
 The host-level client infrastructure
 - should open a second connection if the request on the first one is dispatched but not yet completed *** FAILED *** (3 seconds, 163 milliseconds)
```

For some reason chunk which is sent through the first response isn't arriving from TCP output stream. However if `responseEntitySub.request(1)` is removed it starts to appear in logs. I'm still getting used to streaming internals so I might need some help on this one. I will try to investigate further in a meantime.

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
```diff
@@ -346,9 +349,12 @@ final class Split[T](decision: Split.SplitDecision, p: T ⇒ Boolean) extends Gr
...
+              if (eagerClose) completeStage()
```
rkuhn: shouldn’t this rather happen in onDownstreamFinish for the SubSource?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1093,16 +1093,18 @@ trait FlowOps[+Out, +Mat] {
    *
    * See also [[FlowOps.splitAfter]].
    */
-  def splitWhen(p: Out ⇒ Boolean): SubFlow[Out, Mat, Repr, Closed] = {
+  def splitWhen(p: Out ⇒ Boolean, eagerClose: Boolean = false): SubFlow[Out, Mat, Repr, Closed] = {
```
rkuhn: this will be needed for splitAfter as well

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -241,16 +243,17 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
    * Constructs a [[ClientLayer]] stage using the configured default [[ClientConnectionSettings]],
    * configured using the `akka.http.client` config section.
    */
-  def clientLayer(hostHeader: Host): ClientLayer =
-    clientLayer(hostHeader, ClientConnectionSettings(system))
+  def clientLayer(hostHeader: Host, eagerClose: Boolean): ClientLayer =
```
rkuhn: why should this be configurable? HTTP is rather specific in what it needs, so we can just derive this

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
ataraxer: Sure, I just wanted to decide on API and whether those changes are OK or not first.

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -149,6 +133,90 @@ private[http] object OutgoingConnectionBlueprint {
...
+            setHandler(in, new InHandler {
+              override def onPush(): Unit = grab(in) match {
+                case MessageEnd ⇒
+                  entitySource.complete()
+                  setHandler(in, idle)
+                  responseCancelled = false
+
+                case messagePart ⇒
+                  if (responseCancelled) {
+                    if (eagerClose) completeStage()
```
rkuhn: all completions must be accompanied by closing the entitySource as well, else we leak it

rkuhn: This means that onUpstreamFailure also needs to be overridden.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
rkuhn: ah, yes, I see now; for the client this is actually required

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -70,24 +71,7 @@ private[http] object OutgoingConnectionBlueprint {
...
-      .map {
-        case (Seq(ResponseStart(statusCode, protocol, headers, createEntity, _)), entityParts) ⇒
-          val entity = createEntity(entityParts) withSizeLimit parserSettings.maxContentLength
-          HttpResponse(statusCode, headers, entity, protocol)
-        case (Seq(MessageStartError(_, info)), tail) ⇒
-          // Tails can be empty, but still need one pull to figure that out -- never drop tails.
-          SubSource.kill(tail)
-          throw IllegalResponseException(info)
-      }.concatSubstreams
+      .via(new ResponsePrep(parserSettings, eagerClose))
```
rkuhn: Great performance improvement as well!

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
ataraxer: No, since we need to read at least one more message to find out whether substream has been fully consumed (if next message is matching predicate) or interrupted in the middle, since otherwise normal completion of first substream will always lead to master stream being cancelled on `eagerClose`.

rkuhn: Hmm, not sure I completely understand: receiving cancellation from downstream should only happen if the downstream decides that it does not want more data, which means that canceling is the desired semantics in the eagerClose case (which BTW should be renamed to eagerCancel). Normal substream termination due to the next element matching the substream will signal `substreamSource.complete()`, inhibiting further reception of `onDownstreamFinish()`, so in that case all should be good.

The only “interesting” case is if the substream processing logic double-checks the length of the stream and terminates of its own accord just at the point where the stream would otherwise have ended—but I consider that a less important case than to pass along cancellation without needing further upstream data; an open TCP connection binds resources even when no data are flowing.

ataraxer: The problem is that there is no distinction between mid-stream cancelation and completion of a stream by fully consuming it, both are resulting as `onDownstreamFinish` callback (unless I'm missing something?).

This means that if we will complete stage on `onDownstreamFinish` of a substream when `eagerClose` flag is set, the stream will always terminate as soon as first substream completes (disregarding whether it was interrupted or fully consumed).

Here's an example:

```scala
Source(0 to 9000)
  .splitWhen(0 % 100 == 0, eagerClose = true)
  .concatSubstreams
```

Here each substream will be fully consumed, resulting in `onDownstreamFinish` after the last element of a substream is consumed. So if we complete the entire stage in `onDownstreamFinish` the stream will be interrupted after the first 100 elements.

The purpose of `eagerClose` is to avoid situations when substreams are cancelled without fully draining them, which currently leads to elements being drained automatically which may be unwelcome in case of big or even infinite streams or cases when items of a stream are tightly coupled with some resource consuming work (e.g. HTTP). However, if the substream has been fully consumed there is no point in tearing down the stage.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
ataraxer: What do you thing about the signature? Positional elements after lambda look kinda ugly, so maybe an overload like:
```scala
def splitWhen(eagerClose: Boolean)(p: Out => Boolean)
```
should be introduced instead? This also has the advantage of being extendible with other options/flags in the future, while allowing to use a block syntax for lambda:
```scala
Source(0 to 10).splitWhen(eagerClose = true) { item =>
  ...
  ...
  ...
}
```

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
drewhk: It is `onUpstreamFinish` which is signalled when the last element of a substream is consumed. There is no `onDownstreamFinish` on a port which has been closed, which is usually what happens after `onUpstreamFinish`. The only situations where a stage will get a `onDownstreamFinish` if the consumer does not want to read more. See http://doc.akka.io/docs/akka-stream-and-http-experimental/2.0.2/scala/stream-customize.html#Port_states__InHandler_and_OutHandler

ataraxer: > which BTW should be renamed to eagerCancel

I've seen both being used in different stages/combinators, is there a clear distinction between them? `eagerCancel` is for behaviour related to downstream completion and `eagerClose` is for upstream, correct?

> It is onUpstreamFinish which is signalled when the last element of a substream is consumed.

Ahh, I see.

rkuhn: > I've seen both being used in different stages/combinators, is there a clear distinction between them? eagerCancel is for behaviour related to downstream completion and eagerClose is for upstream, correct?

We have renamed that parameter on the Merge operations to be more precise: there is it about emitting the completion signal, here it is about emitting the cancellation signal.

ataraxer: OK. Should it also be renamed on `outgoingConnection`? I think `eagerClose` suits it better, since it will translate into connection being closed.

rkuhn: WDYM? `Tcp.outgoingConnection` has a `halfClose` parameter, which means something different. “close” is not properly defined for a stage, we have _complete_, _fail_, _cancel_ signals.

We still have eagerClose in SslTlsStage (to be rewritten soon) and on Interleave (which should be changed to eagerComplete).

ataraxer:
What do you think about the signature? Positional elements after lambda look kinda ugly, so maybe an overload like:
```scala
def splitWhen(eagerClose: Boolean)(p: Out => Boolean)
```
should be introduced instead? This also has the advantage of being extendable with other options/flags in the future, while allowing to use a block syntax for lambda:
```scala
Source(0 to 10).splitWhen(eagerClose = true) { item =>
  ...
  ...
  ...
}
```

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -149,6 +133,98 @@ private[http] object OutgoingConnectionBlueprint {
...
+        }
+      }
+
+      setHandler(in, idle)
+
+      private def createEntity(creator: EntityCreator[ResponseOutput, ResponseEntity]): ResponseEntity = {
+        creator match {
+          case StrictEntityCreator(entity) ⇒
+            pull(in)
+            setHandler(in, waitForMessageEnd)
```
ataraxer: Is there probably a better way to read exactly one message of expected type/value?

ataraxer:
- renamed `eagerClose -> eagerCancel` for combinators;
- moved `eagerCancel` handling to `onDownstreamFinish`;
- added `eagerCancel` for `splitAfter`;
- added test cases for `eagerCancel`;
- added `onUpstreamFailure` handler for substreams of `ResponsePrep`;
- added better handling of unexpected items in `ResponsePrep` stage states;

Tests for TCP connection being closed on `HttpResponse` entities cancellation are still WIP (looking for a proper suit to put them into).

Still no luck in figuring out what is wrong with `ConnectionPoolSpec` test 😕

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
ktoso: Awesome, thanks :-)

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -149,6 +133,98 @@ private[http] object OutgoingConnectionBlueprint {
...
+
+          case other ⇒
+            throw new IllegalStateException(f"ResponseStart expected but $other received.")
+        }
+      }
+
+      private val waitForMessageEnd = new InHandler {
+        def onPush(): Unit = grab(in) match {
+          case MessageEnd ⇒ setHandler(in, idle)
+          case other      ⇒ throw new IllegalStateException(f"MessageEnd expected but $other received.")
```
ktoso: We use the `s` interpolator usually, it's enough here too

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
ktoso: This style is fine IMO. We don't have a specific method for the scenario given (you may want to look at `read`, but it's not really better than just setting a handler like already done here).

akka-stream/src/main/scala/akka/stream/impl/fusing/StreamOfStreams.scala
ataraxer: In TCP/HTTP terminology connection is being "closed", so maybe for HTTP client flow constructors `eagerClose` would be more clear. On the other hand I see that `eagerCancel` would be more consistent with other stages.

rkuhn:
LGTM @ataraxer ! I agree with the reasoning concerning the order of the arguments in the Scala DSL, where currying provides nice readability; please add the overloads as proposed (reverting the current signatures to their original state) and then we can merge.

ConnectionPoolSpec is a different issue that we’ll need to figure out, it is not related to this PR AFAICS.

ataraxer:
Well, it used to work with original implementation just fine (: I think I might have actually figured it out now.

And what about the "`eagerClose`/`eagerCancel` on `outgoingConnection` controversy"? (: What's the final decision?

ataraxer:
OK, almost there.
##### TODO
- decide on a flag name;
- add flag to connection pools as well;
- tests, scaladoc and documentation for eagerly closed HTTP connections.

ataraxer:
Also, should eager close flag be added to connection pool settings or only be available programmatically?

ataraxer:
ping

johanandren:
Hi @ataraxer I'm afraid we are spread very thin this week, so that is causing the lack of feedback. We'll look at this as soon as we can!

ataraxer:
Ping 2: The Pingening

How about this week? (:

drewhk:
First of all, it will need a rebase :)

ataraxer:
Ping 3: Revenge of the Ping

rkuhn:
LGTM

rkuhn:
@drewhk please merge tomorrow morning after a last look, I don’t trust my brain right now.

ataraxer:
Please note that there are still some open questions above.

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -148,6 +132,104 @@ private[http] object OutgoingConnectionBlueprint {
...
+                entitySource.fail(reason)
+                failStage(reason)
+              }
+            })
+
+            creator(Source.fromGraph(entitySource.source))
+        }
+      }
+    }
+  }
```
drewhk: This stage looks correct. Nice work!

drewhk:
My opinions:
 - first of all, great work!
 - as for the flag, I would call it `propagateSubstreamCancel` for `splitWhen` 
 - as for HTTP, I think we might not need this to be configurable at all! If someone wants to retain the original behavior, he can just attach a Sink.ignore which will not cancel the substream but drain it into the nothingness. Sink.cancel will automatically close the connection, and I gues that is what people expect. In general, the previous behavior (not propagate cancel) was impossible to opt out from, while with reversing the default it is possible to opt out even without a flag, just by using the proper Sink.

drewhk:
Also, if we agree on the default propagate-cancel, we might want to implement it for the server as well.

ataraxer:
Thanks!

Hopefully I would be able to clean things up this evening and take a look at server side as well.

drewhk:
Also, the benefit of the approach I sketched is that compared to a static flag, the appropriate action can be determined case-by-case, simply by using the proper Sink according to some criteria.

ataraxer:
Aaaaand done.

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
ataraxer: Please give it another quick look, I moved things around a bit to avoid unnecessary allocations.

drewhk:
Almost, the migration guide needs to mention:
 - the actual change, what is the behavior now
 - how to obtain the original behavior (a code snippet showing Sink.ignore - preferably pulled out from an actual unit test).

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -151,8 +151,16 @@ private[http] object HttpServerBluePrint {
         // and then toggle back to "idle"
 
         val entitySource = new SubSourceOutlet[RequestOutput]("EntitySource")
-        // optimization: re-use the idle outHandler
-        entitySource.setHandler(this)
```
ataraxer: This optimization can be restored in a similar way to client's `ResponsePrep` stage by moving entity stream to a variable and implementing `onDownstreamFinish` as a member of stage itself. Things look a bit ..."convoluted" on the server side though, so I decided to avoid messing with it for now (:

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
```diff
@@ -151,8 +151,16 @@ private[http] object HttpServerBluePrint {
...
+          override def onPull(): Unit = pull(in)
+          override def onDownstreamFinish(): Unit = {
+            // application layer has cancelled or only partially consumed response entity:
+            // connection will be closed
+            entitySource.complete()
+            completeStage()
+          }
+        })
 
         setHandler(in, new InHandler {
```
ataraxer: btw this handle can also be combined with the `OutHandler` which is being set on `out` port below too.

ataraxer:
Fixed documentation and added tests.

drewhk:
There are Java docs as well ;)

ataraxer:
Done.

drewhk:
LGTM! Great contribution!

ktoso:
I'm very interested in this one, will review today after some more catching up on other things. Looks very promising :)

drewhk:
Ok, let's try to merge this today or latest tomorrow so we get enough Jenkins runs on it until Friday.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1202,25 +1202,38 @@ trait FlowOps[+Out, +Mat] {
...
    * '''Completes when''' upstream completes
    *
-   * '''Cancels when''' downstream cancels and substreams cancel
+   * '''Cancels when''' downstream cancels and substreams cancel if propagateSubstreamCancel=false, downstream
+   * cancels or any substream cancels if propagateSubstreamCancel=true
    *
    * See also [[FlowOps.splitAfter]].
    */
-  def splitWhen(p: Out ⇒ Boolean): SubFlow[Out, Mat, Repr, Closed] = {
+  def splitWhen(propagateSubstreamCancel: Boolean)(p: Out ⇒ Boolean): SubFlow[Out, Mat, Repr, Closed] = {
```
ktoso: Signature is nice this way :+1: Now we're missing a JavaDSL correspondent for it though. 
Please add a similar one to javadsl.Flow

ktoso: Technically we started phasing out booleans with special classes to make the APIs more readable, so in javadsl it would follow that pattern instead of a boolean I think.

akka-docs/rst/java/http/client-side/connection-level.rst
```diff
@@ -53,6 +53,10 @@ The connection can also be closed by the server.
 An application can actively trigger the closing of the connection by completing the request stream. In this case the
 underlying TCP connection will be closed when the last pending response has been received.
 
+Connection will also be closed if response entity has been cancelled (e.g. by attaching it to ``Sink.cancelled()``)
```
ktoso: "The connection will also be closed if the response entity is cancelled" IMHO sounds more right

akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
ktoso: Can you put a TODO with a note for this here?

akka-docs/rst/scala/stream/migration-guide-2.0-2.4-scala.rst
```diff
@@ -102,6 +102,19 @@ and were accessible via ``settings``. We now made it possible to configure the p
 settings as well, so ``RoutingSettings`` is now ``routingSettings`` and ``ParserSettings`` is
 now accessible via ``parserSettings``.
 
+Client / server behaviour on cancelled entity
+---------------------------------------------
+
+Previously if request received by server-side or response received by client-side were
```
ktoso: if *a* request received

ktoso: Maybe its enough to say: Previously if request or response were cancelled or consumer only partially...

akka-docs/rst/scala/stream/migration-guide-2.0-2.4-scala.rst
```diff
@@ -102,6 +102,19 @@ and were accessible via ``settings``. We now made it possible to configure the p
...
+cancelled or consumed only partially (e.g. by using ``take`` combinator) the remaining
+data was silently drained to prevent stalling the connection, since there could still be
+more requests / responses incoming. Now the default behaviour is to close the connection
+in order to prevent using excessive resource usage in case of huge entities.
+
+Old behaviour can be achieved by explicitly draining the entity:
```
ktoso: The old behaviour

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -69,24 +69,7 @@ private[http] object OutgoingConnectionBlueprint {
     import ParserOutput._
     val responsePrep = Flow[List[ResponseOutput]]
       .mapConcat(conforms)
-      .splitWhen(x ⇒ x.isInstanceOf[MessageStart] || x == MessageEnd)
-      .prefixAndTail(1)
```
ktoso: Great to get rid of this one :)

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
ktoso: I gave it another look and looks good

ktoso:
LGTM, had some minor spelling comments but code looks good, thanks a lot @ataraxer!

akka-docs/rst/scala/http/client-side/connection-level.rst
```diff
@@ -55,6 +55,9 @@ The connection can also be closed by the server.
 An application can actively trigger the closing of the connection by completing the request stream. In this case the
 underlying TCP connection will be closed when the last pending response has been received.
 
+Connection will also be closed if response entity has been cancelled (e.g. by attaching it to ``Sink.cancelled``)
+or consumed only partially (e.g. by calling ``take()`` on it). In order to prevent this behaviour entity should be
+explicitly drained by attaching it to ``Sink.ignore``.
```
2beaucoup: Maybe sprinkle some 'a's or 'an's to prefix nouns like 'connection' and 'entity' (2x)? Or just 'Connections'?

ktoso: :+1: had a comment around the same thing https://github.com/akka/akka/pull/19528#discussion_r52298380

ataraxer:
Argh, missed one.

ataraxer:
Oooook, now everything should be in order:

- restored server optimizations;
- fixed docs;
- rebased.

ktoso:
Thanks a lot - great job on this one! Let's merge after validation :)

drewhk:
Great work!

ataraxer:
Thanks!

Will there be another 2.0.x release or this is going to be a part of 2.4.2?

drewhk:
Bugfixes go into 2.0.x but features and non-compatible changes do not. This is unfortunately a breaking change.

ataraxer:
OK, thanks for clarification.

johanandren:
PLS BUILD

ktoso:
Thanks again, great contrib! :)

PR #19535, opened 2016-01-20
Do not verify if scaladoc diagrams have been generated as there is none

2m:
This is valid, as `akka-http-jackson` project has only one object which does not have any inheritance hiearchy.


drewhk:
LGTM

Thanks for the quick fix

ktoso:
I don't really know if this is a valid fix hm... if you say so though :~
So why would the diagram fail - only because 1 object in there? sounds weird

2m:
So we [have a check](https://github.com/akka/akka/blob/master/project/Doc.scala#L45-L68) in the build that checks if any diagrams are generated during `project/doc` and fails the task if it cannot find any diagrams. In this case there are no diagrams generated, because project in question has only one object for which scaladoc does not generate diagrams. Therefore we need to turn the check off.

ktoso:
I see, thanks for explaining @2m !

ktoso:
LGTM, merge once validates please

PR #19548, opened 2016-01-20
#19522 make javadsl.Sink.asPublisher nicer

### LINKED ISSUE #19522: akka.stream.javadsl.Sink.asPublisher() - change boolean param to enum for Java discoverability
According to the Akka Streams 1.0-to-2.0 migration document at [http://doc.akka.io/docs/akka-stream-and-http-experimental/2.0.2/java/migration-guide-1.0-2.x-java.html#Sink_fanoutPublisher___and_Sink_publisher___is_now_a_single_method] *Source.publisher* and *Source.fanoutPublisher* have been replaced with *Source.asPublisher(boolean)*.  A boolean parameter does not convey the meaning of the parameter value, so I am suggesting with reference to @rkuhn  that the boolean be replaced with enum values (or Scala equivalent) such as FANOUT_PUBLISHER, NON_FANOUT_PUBLISHER (or better name) to convey the parameter's meaning.  Wherever a boolean parameter is exposed in the Java DSL this same change should take place.

rkuhn:
fixes #19522

### DIFF
--- a/akka-stream/src/main/java/akka/stream/javadsl/AsPublisher.java
+++ b/akka-stream/src/main/java/akka/stream/javadsl/AsPublisher.java
@@ -1,0 +1,8 @@
+/**
+ * Copyright (C) 2016 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.javadsl;
+
+public enum AsPublisher {
+  WITH_FANOUT, WITHOUT_FANOUT
+}
--- a/akka-docs/rst/java/code/docs/stream/MigrationsJava.java
+++ b/akka-docs/rst/java/code/docs/stream/MigrationsJava.java
@@ -7,6 +7,9 @@ import java.util.stream.Stream;
 
 import akka.japi.Pair;
 import akka.stream.javadsl.*;
+//#asPublisher-import
+import static akka.stream.javadsl.AsPublisher.*;
+//#asPublisher-import
 
 public class MigrationsJava {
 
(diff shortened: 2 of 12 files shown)


rkuhn:
failure was known (SubstreamSubscriptionTimeoutSpec)

johanandren:
Isn't there a few mentions in the `.rst` files also?

rkuhn:
Hmm, I would have hoped that they use compiled code, but I’ll check.

rkuhn:
You were right, of course :-)

akka-stream/src/main/java/akka/stream/javadsl/AsPublisher.java
```diff
@@ -0,0 +1,8 @@
+/**
+ * Copyright (C) 2016 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.javadsl;
+
+public enum AsPublisher {
+  WITH_FANOUT, WITHOUT_FANOUT
+}
```
ktoso: Perhaps not an issue, but was thinking to keep the same style for those in all places.
e.g. the new Http APIs proposal enum is not enough (as needs to be configurable), so I was thinking if exposing the enum directly is a good idea or not (could be an enum backing 2 impls of a interface, but we get those via methods).

johanandren:
LGTM except for that

akka-stream/src/main/java/akka/stream/javadsl/AsPublisher.java
rkuhn: I think we should favor idiomatic language use over our notion of consistency, i.e. enums have a clear use-case and where they don’t fit we do something else.

ktoso: Also true, let's run with this one then :+1:

ktoso:
LGTM

akka-stream/src/main/java/akka/stream/javadsl/AsPublisher.java
drewhk: yeah, enums are fine here (they have private constructor, they serialize, have some valueOf, toString sugar, etc).

drewhk:
LGTM

rkuhn:
oh, forgot to push, so no validation yet—I was starting to wonder :-)

drewhk:
needs inclusion in the migration guide for java

rkuhn:
merged, including migration guide

Issue #18869, opened 2015-11-05
SynchronousFileSink materialized future should fail on error


pkinsky:
If a SynchronousFileSink fails due to an upstream error, the future created on materialization completes successfully with the number of bytes written.
```
scala> val failed = Source.failed[ByteString](new Exception("YOLO"))
scala> failed.runWith(SynchronousFileSink(new File("test.txt"))).onComplete(println)
[ERROR] [11/05/2015 15:46:33.604] [default-akka.stream.default-file-io-dispatcher-23]
[akka://default/user/$a/flow-3-1-failedSource-synchronousFileSink] 
Tearing down SynchronousFileSink(/Users/pkinsky/dev/ws-watcher/test.txt) due to upstream error
java.lang.Exception: YOLO
...
Success(0)
```

I think that the correct behavior here is for the future to fail with the throwable that caused the upstream error. This could be accomplished by modifying the code here, in SynchronousFileSubscriber.scala, to include `bytesWrittenPromise.failure(cause)`

```
    case ActorSubscriberMessage.OnError(cause) ⇒
      log.error(cause, "Tearing down SynchronousFileSink({}) due to upstream error", f.getAbsolutePath)
      context.stop(self)
```

However, some applications may require access to the number of bytes written before an upstream failure. If this is the desired behavior, I believe a note should be added to the documentation (if there's not one there already, I couldn't find anything documenting this behavior).

pkinsky:
quick and dirty workaround: 
```
  def safeFileSink(f: File): Sink[ByteString, Future[Long]] = {
    val fileSink = SynchronousFileSink(f)
    val errorMonitor = Sink.ignore
    Sink(fileSink, errorMonitor)((f1, f2) => f2.flatMap(_ => f1)) { implicit builder =>
      (fileSink, errorMonitor) =>
        import FlowGraph.Implicits._

        val broadcast = builder.add(Broadcast[ByteString](2))

        broadcast.out(0) ~> fileSink
        broadcast.out(1) ~> errorMonitor

        broadcast.in
    }
  }
```

drewhk:
Thanks for reporting!

rkuhn:
Hmm, so what shall we do? I do not agree that this version is necessarily more “safe” than the current behavior because safety is subject to the environment in which it is evaluated.

[...]

Issue #19344, opened 2016-01-04
FAILED: akka.pattern.BackoffSupervisorSpec


johanandren:
Timeout in "must support manual reset"

https://jenkins.akka.io:8498/job/akka-multi-node-nightly2/224/consoleFull

johanandren:
https://jenkins.akka.io:8498/job/akka-nightly/2939/consoleFull

johanandren:
https://jenkins.akka.io:8498/job/akka-nightly/2941/consoleFull

johanandren:
And another one of "must support custom supervision strategy": 

https://jenkins.akka.io:8498/job/akka-multi-node-nightly2/225/

johanandren:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly2/230/consoleFull

johanandren:
"must support manual reset"
https://jenkins.akka.io:8498/job/akka-nightly/2958/consoleFull

johanandren:
https://jenkins.akka.io:8498/job/akka-nightly/2962/consoleFull

johanandren:
https://jenkins.akka.io:8498/job/akka-nightly/2964/consoleFull

```
 must support custom supervision strategy *** FAILED *** (3 seconds, 212 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class akka.pattern.BackoffSupervisor$CurrentChild
[info]   at scala.Predef$.assert(Predef.scala:165)
[info]   at akka.testkit.TestKitBase$class.expectMsgClass_internal(TestKit.scala:435)
[info]   at akka.testkit.TestKitBase$class.expectMsgType(TestKit.scala:408)
[info]   at akka.testkit.TestKit.expectMsgType(TestKit.scala:735)
[info]   at akka.pattern.BackoffSupervisorSpec$$anonfun$1$$anonfun$apply$mcV$sp$4$$anonfun$akka$pattern$BackoffSupervisorSpec$$anonfun$$anonfun$$assertCustomStrategy$1$1.apply$mcV$sp(BackoffSupervisorSpec.scala:85)
[info]   at akka.pattern.BackoffSupervisorSpec$$anonfun$1$$anonfun$apply$mcV$sp$4$$anonfun$akka$pattern$BackoffSupervisorSpec$$anonfun$$anonfun$$assertCustomStrategy$1$1.apply(BackoffSupervisorSpec.scala:82)
[...]

PR #19554, opened 2016-01-21
#19446 #18869 Add IOResult

### LINKED ISSUE #19446: add IOResult type and use it for FileSink/Source
~~~scala
case class IOResult(count: Long, status: Try[Done])
~~~
The meaning of the `count` member needs to be defined by the stage that emits or materializes to such an element, for byte streams it should be the number of bytes, but for a `Sink.count` it could also be the number of elements encountered (or individually scaled by a weight function).

2m:
which is used as a materialized value for Sources/Sinks that work with IO.

Fixes #19446 #18869.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/io/FileSubscriber.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/io/FileSubscriber.scala
@@ -7,15 +7,18 @@ import java.io.File
 import java.nio.channels.FileChannel
 import java.util.Collections
 
+import akka.Done
 import akka.actor.{ Deploy, ActorLogging, Props }
+import akka.stream.io.IOResult
 import akka.stream.actor.{ ActorSubscriberMessage, WatermarkRequestStrategy }
 import akka.util.ByteString
 
 import scala.concurrent.Promise
+import scala.util.{ Failure, Success }
 
 /** INTERNAL API */
 private[akka] object FileSubscriber {
-  def props(f: File, completionPromise: Promise[Long], bufSize: Int, append: Boolean) = {
+  def props(f: File, completionPromise: Promise[IOResult], bufSize: Int, append: Boolean) = {
     require(bufSize > 0, "buffer size must be > 0")
     Props(classOf[FileSubscriber], f, completionPromise, bufSize, append).withDeploy(Deploy.local)
   }
@@ -26,7 +29,7 @@ private[akka] object FileSubscriber {
 }
 
 /** INTERNAL API */
-private[akka] class FileSubscriber(f: File, bytesWrittenPromise: Promise[Long], bufSize: Int, append: Boolean)
+private[akka] class FileSubscriber(f: File, completionPromise: Promise[IOResult], bufSize: Int, append: Boolean)
   extends akka.stream.actor.ActorSubscriber
   with ActorLogging {
 
@@ -43,7 +46,7 @@ private[akka] class FileSubscriber(f: File, bytesWrittenPromise: Promise[Long], 
     super.preStart()
   } catch {
     case ex: Exception ⇒
-      bytesWrittenPromise.failure(ex)
+      completionPromise.success(IOResult(bytesWritten, Failure(ex)))
       cancel()
   }
 
@@ -53,12 +56,13 @@ private[akka] class FileSubscriber(f: File, bytesWrittenPromise: Promise[Long], 
         bytesWritten += chan.write(bytes.asByteBuffer)
       } catch {
         case ex: Exception ⇒
-          bytesWrittenPromise.failure(ex)
+          completionPromise.success(IOResult(bytesWritten, Failure(ex)))
           cancel()
       }
 
-    case ActorSubscriberMessage.OnError(cause) ⇒
-      log.error(cause, "Tearing down FileSink({}) due to upstream error", f.getAbsolutePath)
+    case ActorSubscriberMessage.OnError(ex) ⇒
+      log.error(ex, "Tearing down FileSink({}) due to upstream error", f.getAbsolutePath)
+      completionPromise.success(IOResult(bytesWritten, Failure(ex)))
       context.stop(self)
 
     case ActorSubscriberMessage.OnComplete ⇒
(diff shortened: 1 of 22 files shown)


akka-stream/src/main/scala/akka/stream/io/IOResult.scala
```diff
@@ -0,0 +1,33 @@
+package akka.stream.io
+
+import akka.Done
+import scala.util.{ Failure, Success, Try }
+
+/**
+ * Holds a result of an IO operation.
+ *
+ * @param count A number of IO operations performed.
```
2beaucoup: That's actually the number of bytes read/written, not the number of reads/writes, right?

2m: This class is generic enough and can be used with different sources and sinks which then define what count really means.

2beaucoup: I just found the term "operations" a bit misleading.

2m:
Failure was in `akka-jmh-bench`  which I missed.

akka-stream/src/main/scala/akka/stream/io/IOResult.scala
```diff
@@ -0,0 +1,33 @@
+package akka.stream.io
```
ktoso: missing license header

akka-stream/src/main/scala/akka/stream/io/IOResult.scala
ktoso: Yeah, I think this should always talk about bytes written / read.
It should be final too OR not a case class which is also interesting... Let's go with final though, others can introduce new types if they need to.

johanandren: What about something like "Numeric value depending on context, for example IO operations performed or Bytes processed"?

akka-stream/src/main/scala/akka/stream/impl/io/FileSubscriber.scala
```diff
@@ -66,13 +70,13 @@ private[akka] class FileSubscriber(f: File, bytesWrittenPromise: Promise[Long],
...
         case ex: Exception ⇒
-          bytesWrittenPromise.failure(ex)
+          completionPromise.success(IOResult(bytesWritten, Failure(ex)))
       }
       context.stop(self)
   }
 
   override def postStop(): Unit = {
-    bytesWrittenPromise.trySuccess(bytesWritten)
+    completionPromise.trySuccess(IOResult(bytesWritten, Success(Done)))
```
ktoso: we discussed it somewhere in a ticket, that the completing of the promise should be done after the channel as been closed – as closing it may still fail actually (line 81 here)

akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamSubscriber.scala
```diff
@@ -51,7 +55,7 @@ private[akka] class OutputStreamSubscriber(os: OutputStream, bytesWrittenPromise
   }
 
   override def postStop(): Unit = {
-    bytesWrittenPromise.trySuccess(bytesWritten)
+    completionPromise.trySuccess(IOResult(bytesWritten, Success(Done)))
```
ktoso: same as above, should be able to handle if os.close blows up

ktoso:
Looks good, had some minor comments

akka-stream/src/main/scala/akka/stream/io/IOResult.scala
drewhk: then please document it that this value is context dependend. In general people are expected in bytes (especially since rechunking might happen). Which case needs something else than bytes?

akka-stream/src/main/scala/akka/stream/impl/io/FileSubscriber.scala
2m: I see. But then I should not only move the completion down one line, but also catch any exceptions from `close()` and complete the promise with `IOResult(bytesWritte, Failure(ex))` if exception is thrown, right?

ktoso: Yeah, exactly – thanks!

akka-stream/src/main/scala/akka/stream/io/IOResult.scala
rkuhn: Bytes might not be available (for a future `Sink.count`) or the right abstraction (for text processing).

+1 for explicitly documenting the meaning to be context-dependent.

rkuhn:
LGTM apart from the comments

johanandren:
LGTM except the number-scaladoc!

rkuhn:
ETA?

2m:
Addressed feedback comments. Failure was `akka.pattern.BackoffSupervisorSpec` (#19344) which is relevant to this PR.

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
```diff
@@ -73,8 +76,15 @@ private[akka] class InputStreamPublisher(is: InputStream, bytesReadPromise: Prom
 
   override def postStop(): Unit = {
     super.postStop()
-    bytesReadPromise.trySuccess(readBytesTotal)
 
-    if (is ne null) is.close()
+    try {
+      if (is ne null) is.close()
+    }
+    catch {
```
ktoso: a bit weird formatting, `} catch {`

ktoso:
LGTM

drewhk:
won't merge though

2m:
Ulialia. Will rebase.

2m:
Rebased. A conflict was in migration spec.

rkuhn:
did it pass validation before?

2m:
Validation before the rebase is still running and is nearly finished. I will merge when current validation finishes and will not wait for the last one, as only migration guide has been changed.

https://jenkins.akka.io:8498/view/PR%20Validators/job/pr-validator-per-commit-jenkins/4944/consoleFull

rkuhn:
Thanks!

Issue #19333, opened 2016-01-03
consider removing trait ImplicitMaterializer or making it more discoverable


rkuhn:
Usage of this trait is easily replaced by a single line and it does not show up in content assist unless the user knows to start searching for `Implicit` (which they won’t).

PR #19558, opened 2016-01-21
#19333: Remove ImplicitMaterializer

drewhk:


drewhk:
Fixes #19333

ktoso:
Remove its mention from `akka-docs/rst/scala/http/client-side/request-level.rst` too.

LGTM

drewhk:
ok, thanks

johanandren:
LGTM!

patriknw:
you don't write migration guide as you go?

drewhk:
I actually forgot in this case, so no.

drewhk:
(in general, yes, but in this case just somehow assumed this class to be not used at all :) )

drewhk:
added the missing section: https://github.com/akka/akka/pull/19574

Issue #19549, opened 2016-01-20
Akka HTTP fails with A Merge must have more than 1 input port


relgames:
This sample https://gist.github.com/relgames/38aac67d4020d7a95f37 works fine with Akka HTTP 2.0-M2 but fails with 2.0 and 2.0.2:

```
[2016-01-20 21:20:52,080] [  DEBUG] [spatcher-3] [akka://default/user/PoolInterfaceActor-0] a.h.i.e.c.PoolInterfaceActor: (Re-)starting host connection pool to 127.0.0.1:8082
[2016-01-20 21:20:52,187] [  ERROR] [spatcher-3] [akka://default/user/PoolInterfaceActor-0] a.a.OneForOneStrategy: requirement failed: A Merge must have more than 1 input port
akka.actor.ActorInitializationException: exception during creation
	at akka.actor.ActorInitializationException$.apply(Actor.scala:172) ~[akka-actor_2.11-2.4.0.jar:na]
	at akka.actor.ActorCell.create(ActorCell.scala:605) ~[akka-actor_2.11-2.4.0.jar:na]
	at akka.actor.ActorCell.invokeAll$1(ActorCell.scala:460) ~[akka-actor_2.11-2.4.0.jar:na]
	at akka.actor.ActorCell.systemInvoke(ActorCell.scala:482) ~[akka-actor_2.11-2.4.0.jar:na]
	at akka.dispatch.Mailbox.processAllSystemMessages(Mailbox.scala:282) [akka-actor_2.11-2.4.0.jar:na]
	at akka.dispatch.Mailbox.run(Mailbox.scala:223) [akka-actor_2.11-2.4.0.jar:na]
	at akka.dispatch.Mailbox.exec(Mailbox.scala:234) [akka-actor_2.11-2.4.0.jar:na]
	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260) [scala-library-2.11.7.jar:na]
	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339) [scala-library-2.11.7.jar:na]
	at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979) [scala-library-2.11.7.jar:na]
	at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107) [scala-library-2.11.7.jar:na]
[...]

Issue #19257, opened 2015-12-22
Support Broadcast with just one output


hbf:
My code builds a graph by calling

    builder.add(Broadcast.create(writers.size()));

If `writers.size()` is 1, I obtain 

    java.lang.IllegalArgumentException: requirement failed: A Broadcast must have more than 1 output ports

I understand that a `Broadcast` with 1 output is semantically meaningless. However, it simplifies my code – and I'm sure the code in many other situations – if `Broadcast` allowed one output.

Can we support that case?

(Tested with _com.typesafe.akka:akka-stream-experimental_2.10:2.0_.)

hepin1989:
@hbf you could using `Flow` instead.

drewhk:
We supported it back then, and then we changed to require at least two outputs. You can put an `if ` around that part where you construct your broadcast, simply replacing it with nothing when the outputs are one.

hbf:
@hepin1989, @drewhk Thanks for the suggestion. I understand that I can use a `Flow` here – my point is about making the `Broadcast` API generic: Not allowing the case of single output forces users to do special casing when they write algorithms based on streams.

hepin1989:
@hbf I asked that once,you could see the conversation here https://github.com/akka/akka/issues/19028

hbf:
@hepin1989 Thanks for the link! Ok, looks like this was discussed at length. I may be missing something but I'd have preferred this special case to be optimized away behind the scenes and allow API users to write simple code. – Just my 2 cents, so feel free to close this issue.

hepin1989:
@hbf we are the community,and learn from each other:)

ktoso:
I'll leave it open until the rest of the team comes back and we decide on it. 
I for one am not that happy with not allowing broadcast(1) (due to not optimising it away currently)

lespea:
+1 this would really simplify some of my code

viktorklang:
I think about it like this: if we could have a constraint type on Int which limited it to `greater-than-two` we would.

Should we also support Broadcast(-1) or Broadcast(Int.MinValue)?

hbf:
My thinking is this: `Broadcast(n)` is meaningful for _n_ >= 0, and the semantics are straightforward and most likely intuitively clear to the user of this API. To me, that speaks for allowing these cases. I'd throw an exception if _n_ < 0. 

[...]

PR #19560, opened 2016-01-21
=str #19549 #19257 Allow Broadcast(1) and Merge(1)

johanandren:
Changes the `require` in each, and covers with tests. No optimizations for the 1-to-1 case.

Refs #19549 and #19257

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphMergeSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphMergeSpec.scala
@@ -5,6 +5,7 @@ package akka.stream.scaladsl
 
 import akka.stream._
 
+import scala.concurrent.Await
 import scala.concurrent.duration._
 
 import akka.stream.testkit._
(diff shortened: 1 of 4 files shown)


akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphMergeSpec.scala
```diff
@@ -58,6 +59,18 @@ class GraphMergeSpec extends TwoStreamsSetup {
       probe.expectComplete()
     }
 
+    "work with 1-way merge" in {
```
johanandren: now that wasn't very consistent was it?

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -39,7 +39,8 @@ object Merge {
  * '''Cancels when''' downstream cancels
  */
 final class Merge[T] private (val inputPorts: Int, val eagerComplete: Boolean) extends GraphStage[UniformFanInShape[T, T]] {
-  require(inputPorts > 1, "A Merge must have more than 1 input port")
+  // one input might seem counter intuitive but saves us from special handling in other places
+  require(inputPorts >= 1, "A Merge must have one or more input ports")
```
drewhk: only these two are affected? Balance does not have the check?

drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
johanandren: Good call, it does indeed have the same check.

drewhk:
now really LGTM

ktoso:
Thanks Johan!

Issue #19552, opened 2016-01-21
implement conflate in terms of Batch


rkuhn:
… and remove the Conflate implementation

PR #19564, opened 2016-01-21
=str #19552 Replacing Conflate with Batch

johanandren:
Refs #19552 

Have problems figuring out if and what to do with the supervision strategy that worked for conflate.


akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowConflateSpec.scala
```diff
@@ -1,100 +0,0 @@
...
-
-    "work with a buffer and fold" in {
-      val future = Source(1 to 50)
-        .conflate(seed = i ⇒ i)(aggregate = (sum, i) ⇒ sum + i)
-        .buffer(50, OverflowStrategy.backpressure)
-        .runFold(0)(_ + _)
-      Await.result(future, 3.seconds) should be((1 to 50).sum)
-    }
-
-  }
```
rkuhn: Please keep this spec.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
```diff
@@ -241,10 +242,11 @@ class InterpreterSpec extends AkkaSpec with GraphInterpreterSpecKit {
       lastEvents() should be(Set(OnNext(Vector(3)), OnComplete))
     }
 
-    "implement conflate" in new OneBoundedSetup[Int](Seq(Conflate(
+    "implement batch (conflate)" in new OneBoundedSetup[Int](Batch(
+      0L,
```
rkuhn: Not that it should matter, but this is `1L` in the actual conflate code.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterStressSpec.scala
```diff
@@ -100,10 +101,11 @@ class InterpreterStressSpec extends AkkaSpec with GraphInterpreterSpecKit {
...
     }
 
-    "work with a massive chain of conflates by overflowing to the heap" in new OneBoundedSetup[Int](Vector.fill(chainLength / 10)(Conflate(
+    "work with a massive chain of batches by overflowing to the heap" in new OneBoundedSetup[Int](Vector.fill(chainLength / 10)(Batch(
+      0L,
+      ConstantFun.zeroLong,
       (in: Int) ⇒ in,
-      (agg: Int, in: Int) ⇒ agg + in,
-      Supervision.stoppingDecider))) {
+      (agg: Int, in: Int) ⇒ agg + in)): _*) {
```
rkuhn: this can now be changed to reuse the Batch instance (Conflate was not reusable)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowConflateSpec.scala
johanandren: Ups. Was a bit too quick there thinking that it tested the `Conflate` I just removed.

johanandren:
Stupid ScalaUdpMulticastSpec. Unrelated error.

johanandren:
Unrelated test failure. Everything else green.

johanandren:
PLS BUILD

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
```diff
@@ -387,70 +387,6 @@ class InterpreterSupervisionSpec extends AkkaSpec with GraphInterpreterSpecKit {
       lastEvents() should be(Set(OnNext(25))) // 1 + 4 + 20
     }
 
-    "restart when Conflate `seed` throws" in new OneBoundedSetup[Int](Seq(Conflate(
```
drewhk: given that the functionality is in there, can't these tests be reinstated?

drewhk:
I would like merge but I am not sure why some tests are removed although there is support for directives in the stage itself. Given that there are restart supports there should be some test for it, even if the original ones are no longer appropriate. (otherwise LGTM)

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
johanandren: Looking at `Batch` I see nothing about supervision, doesn't there have to be implemented by the `GraphStage` if a stage should have supervision?

drewhk: Then what are these lines about: https://github.com/akka/akka/pull/19564/files#diff-3203199ba389ee802eef94486f26dc88R521

johanandren: Oh, now I'm just confused.

johanandren: Ok, now I remember what was the problem, I actually asked you about it once already, tried to do what you recommended and failed. 

The thing was that `withAttributes`, which I needed to set the supervisor, returns a `Graph`, so the actual `GraphStage` is wrapped in an anonymous object and because of that I couldn't use it together with the `OneBoundedSetup` in hte test.

johanandren: How about I remove those tests but add coverage for supervision in the `FlowBatchSpec` instead?

drewhk: yes, you can move it there. You don't really need OneBoundedSetup to test the supervision. This is a very low level testing facility, but I don't think this is needed in this case.

johanandren:
PLS BUILD

johanandren:
Updated with @rkuhn s feedback, supervision test coverage, rebased and squashed.

drewhk:
Great, LGTM

PR #19565, opened 2016-01-21
#19537 Increase RS TCK timeouts

### LINKED ISSUE #19537: Increase TCK test timeouts
Currently it is 500 ms which is probably too short.

drewhk:
Fixes #19537

### DIFF
--- a/akka-stream-tests-tck/src/test/scala/akka/stream/tck/Timeouts.scala
+++ b/akka-stream-tests-tck/src/test/scala/akka/stream/tck/Timeouts.scala
@@ -12,8 +12,8 @@ import akka.testkit._
  */
 object Timeouts {
 
-  def publisherShutdownTimeoutMillis: Int = 1000
+  def publisherShutdownTimeoutMillis: Int = 3000
 
-  def defaultTimeoutMillis: Int = 500
+  def defaultTimeoutMillis: Int = 800
 
 }


akka-stream-tests-tck/src/test/scala/akka/stream/tck/Timeouts.scala
```diff
@@ -12,8 +12,8 @@ import akka.testkit._
  */
 object Timeouts {
 
-  def publisherShutdownTimeoutMillis: Int = 1000
+  def publisherShutdownTimeoutMillis: Int = 3000
```
drewhk: this does not affect TCK validation time too much

akka-stream-tests-tck/src/test/scala/akka/stream/tck/Timeouts.scala
```diff
@@ -12,8 +12,8 @@ import akka.testkit._
...
 
-  def defaultTimeoutMillis: Int = 500
+  def defaultTimeoutMillis: Int = 800
```
drewhk: This adds 100s to validation on my machine so I am not happy. The current limit is too tight, increasing it will make validation almost 2 minutes slower... Is there any viable solution?

akka-stream-tests-tck/src/test/scala/akka/stream/tck/Timeouts.scala
ktoso: there's a few "expect no message" but if it's not too painful then sure, better than racy tests :)

akka-stream-tests-tck/src/test/scala/akka/stream/tck/Timeouts.scala
ktoso: Dependency tracking builds would help, also known as the Holy Grail :<

Let's increase for now, but in medium term I think we must invest time on paralleising the build into "stages" and those must be parallel on multiple machines.

ktoso:
Not a pleasant change but let's do it. We have work to do with the build speed in any case.

akka-stream-tests-tck/src/test/scala/akka/stream/tck/Timeouts.scala
drewhk: This value was not really relevant to the total, compared to the one below :(

ktoso: Ah yeah, the bellow one is used for those awaits... This one is only used in one test or so

PR #19566, opened 2016-01-21
!htc #19514 make config classes easier to evolve bin-compat

### LINKED ISSUE #18958: Body Part Parser contains hardcoded config defaults, should be loaded from reference.conf
As mentioned here https://github.com/akka/akka/pull/18941#discussion_r44984789
It's as simple as making the same config object populated form reference.conf instead of defaults in code.

### LINKED ISSUE #19514: HTTP: Move away from case classes for configuration classes
Because we'll have to binary compatible with http-core we need to move away from case classes for configuration objects ASAP.

More info via @patriknw :

>  January 18, 2016 4:28 PM @ktoso re case classes, yeah they are a nightmare to evolve, and settings classes are expected to be expanded over time. I abandoned case classes for all settings classes in cluster tools, e.g. https://github.com/akka/akka/blob/master/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ClusterShardingSettings.scala

ktoso:
Resolves #19514 - configs are now normal classes, easier to bin-compat evolve (could be even easier, but this'll do for now)
Resolves #18958 - Parser config is now parsed once and shared, instead of being parsed in a few places. It also has defaults from reference.conf and not hardcoded


ktoso:
Thanks @jrudolph for reminding me that Unmarshallers have the materializer available, totally forgot about it last evening.

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/MultipartUnmarshallers.scala
```diff
@@ -67,7 +70,7 @@ trait MultipartUnmarshallers {
               FastFuture.failed(new RuntimeException("Content-Type with a multipart media type must have a 'boundary' parameter"))
             case Some(boundary) ⇒
               import BodyPartParser._
-              val parser = new BodyPartParser(defaultContentType, boundary, log)
+              val parser = new BodyPartParser(defaultContentType, boundary, log, ParserSettings(ActorMaterializer.downcast(mat).system)) // TODO we could cache it
```
ktoso: OR, we make each materializer expose config...? // cc @rkuhn

akka-http/src/main/scala/akka/http/scaladsl/server/RequestContext.scala
```diff
@@ -43,6 +44,11 @@ trait RequestContext {
   def settings: RoutingSettings
 
   /**
+   * The default ParserSettings to be used for configuring directives.
+   */
+  def parserSettings: ParserSettings
```
jrudolph: Is this used somewhere? In any case, it would make sense to keep the RequestContext lean. If we need it here at all we should maybe move it into RoutingSettings? Better would be to remove it if it isn't used, though :)

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/MultipartUnmarshallers.scala
jrudolph: Built like this, it isn't possible to override the settings for a particular use site. My (too brief) suggestion was to add an `parserSettingOverride: Option[ParserSetting]` parameter to all the methods that need it in MultipartUnmarshallers so that a user is able to create custom mutipart marshallers that use the custom setting.

akka-http-core/src/main/scala/akka/http/ClientConnectionSettings.scala
```diff
@@ -19,30 +19,59 @@ import akka.http.impl.util._
...
-final case class ClientConnectionSettings(
-  userAgentHeader: Option[`User-Agent`],
-  connectingTimeout: FiniteDuration,
-  idleTimeout: Duration,
-  requestHeaderSizeHint: Int,
-  websocketRandomFactory: () ⇒ Random,
-  socketOptions: immutable.Traversable[SocketOption],
-  parserSettings: ParserSettings) {
+final class ClientConnectionSettings(
+  val userAgentHeader: Option[`User-Agent`],
```
patriknw: this is for scala api only, right?

patriknw:
LGTM

jrudolph:
Is there a link that would explain the binary compatibility issues with case classes and this change would fix it? What kind of operations are now allowed that wouldn't be otherwise? 

AFAICS you still aren't allowed to add or remove a field here because calls to `copy` will break if you change the parameters. So, what we probably need is a public interface (trait) to a private case class and then manual `withXYZ` methods as a binary compatible alternative to the `copy` method. Or am I missing something?

(Maybe there is a compiler improvement that would help with issues like this? Couldn't you lobby for such a change internally?)

jrudolph:
Or, is the plan to evolve the `copy` methods only for binary-incompatible releases but keep it for fixes? In any case, the constructors then needs to be evolved in a controlled way as well.

jrudolph:
Another possibility would be to keep the code as it is and advice any developers that develop "Akka middleware", i.e. library authors that would suffer from binary incompabilities in Akka, just not to use `apply` and `copy` methods if they want to stay compatible without recompilation. This may already work for some share of the use-cases.

patriknw:
Adding new properties is the use case.
Copy, and constructor must be evolved, case classes adds a lot more and it's difficult to add overloads.

Possibly also `apply`, but I think it's best if `apply` is only for the mandatory properties and the optional are set with `copy`.

I think an even safer approach is to make `copy` private, and instead provide `withX` methods. apply can be for the mandatory properties, and optional properties are set using the `withX` methods.

ktoso:
Thanks for reviewing @jrudolph, I'll get to addressing this shortyly

> I think an even safer approach is to make copy private, and instead provide withX methods. apply can be for the mandatory properties, and optional properties are set using the withX methods.

Yeah that's true.

jrudolph:
I created a ticket for scala to discuss a potential compiler side solution here: https://issues.scala-lang.org/browse/SI-9627

ktoso:
Yeah compiler support would be lovely here... The sbt guys invented something to gen bin compatible looks-like case class some time ago, but it did not get enough traction (was external tool).

I also realised the settings need Java API so `case class impl + scaladsl + javadsl` is our best way out of here, thanks for reviewing guys, would not have realised this sooner.

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/MultipartUnmarshallers.scala
ktoso: thanks for catching this

ktoso:
Passed for me locally just now, attempting rebase.

ktoso:
This versions follows @jrudolph's suggestion of being backed by a case class, however interface is evolvable since hidden behind interfaces

ktoso:
Wohoo, all green :-) Please review to check if I missed anything, but that's admittedly pretty hard to visually check - specs and massive copy pasting helped to make sure everything's covered I hope.

Merge at your convenience, I think should be good as-is.

rkuhn:
Thanks Konrad, and now enjoy your vacation!

ktoso:
Enjoying: in progress :-)

johanandren:
Closing as #19606 contains these changes rebased

PR #19568, opened 2016-01-22
Update akka http and stream version for osgi

wbowling:
The osgi manifests were requiring version 1.0 of the stream and http packages instead of version 2.0


ktoso:
Seems valid to me, thanks.
LGTM

rkuhn:
And since we currently do not have any PR validation for HTTP OSGi we might as well just merge it. Thanks @wbowling !

Issue #19392, opened 2016-01-11
Timeouts when using StreamConverters.asInputStream


kostya-sh:
The following sample program eventually fails because of timeout:

    while (true) {
      val source = FileIO.fromFile(new File("file.txt"), 100)
      val inputStream = source.runWith(StreamConverters.asInputStream(5.second))
      val buf = new Array[Byte](200)
      while (inputStream.read(buf) > 0) {}
      inputStream.close()
      println("ok")
    }

file.txt size is 20kb

Timeout exception:

    Exception in thread "main" java.io.IOException: Timeout on waiting for new data
	at akka.stream.impl.io.InputStreamAdapter$$anonfun$read$1.apply$mcI$sp(InputStreamSinkStage.scala:139)
	at akka.stream.impl.io.InputStreamAdapter$$anonfun$read$1.apply(InputStreamSinkStage.scala:124)
	at akka.stream.impl.io.InputStreamAdapter$$anonfun$read$1.apply(InputStreamSinkStage.scala:124)
	at akka.stream.impl.io.InputStreamAdapter.executeIfNotClosed(InputStreamSinkStage.scala:104)
	at akka.stream.impl.io.InputStreamAdapter.read(InputStreamSinkStage.scala:124)
	at akka.stream.impl.io.InputStreamAdapter.read(InputStreamSinkStage.scala:115)
	at ....

Environment: Java 7, scala 2.11.7, akka 2.3.12, akka-streams 2.0.1

viktorklang:
My suspicion: the returned InputStream's close()-method is not synchronous.

kostya-sh:
@viktorklang, timeout happens in the read method (before close has been called) because `InputStreamAdapter.sharedBuffer` is empty and `InputStreamSinkStage` doesn't put anything there for more than 5 seconds. I suspect that communication protocol between InputStreamAdapter and InputStreamSinkStage is not broken.

kostya-sh:
Possible reason: `sharedBuffer.poll()` is called by `grabDataChunk` that is called recursively from `getData`. While this happens `ReadElementAcknowledgement` is never sent to `InputStreamSinkStage`

kodemaniak:
I think I am hitting the same problem, with the following snippet and Streams 2.0.2:

````scala
val source = Source.single(ByteString("test"))
val is = source.runWith(StreamConverters.asInputStream())
val source2 = StreamConverters.fromInputStream(() => is)
[...]

PR #19575, opened 2016-01-22
Fix for #19392 (Timeouts with StreamConvertes.asInputStream)

kodemaniak:
The problem was , that an Ack pulled also if the buffer was full, which sometimes resulted in pushing into the buffer when it was full. Now the remaining capacity is always checked. Additionally there was a bug  with updating the state of the stage in the adapter resulting in timeouts as well.


ktoso:
OK TO TEST

drewhk:
This needs a rebase and a squash. We can take over this.

drewhk:
taking over here: https://github.com/akka/akka/pull/19660

PR #19579, opened 2016-01-22
#19344 Made BackoffSupervisorSpec more robust

### LINKED ISSUE #19344: FAILED: akka.pattern.BackoffSupervisorSpec

drewhk:
I think there is a race with the `GetRestartCount` and the generated `Terminated` at the parent caused by the `DeathWatchNotification` received for the child, but I am not 100% sure. The modification won't make the situation worse though, so let's give it a try.

Fixes #19344


drewhk:
Merging this since it does not change semantics but might help PR validation and nightlies.

Issue #19388, opened 2016-01-08
Akka Http: Parsing of X-Forwarded-For should avoid DNS lookups if possible


cmbaxter:
As mentioned in this forum post:

https://groups.google.com/forum/#!topic/akka-user/e5cl-TjGqNg

I think it's a good idea to try and avoid DNS lookups when parsing the X-Forwarded-For header.  The InetAddress.getByName method is smart enough to see if the value supplied is an ip vs a host, but if it is a host, it will result in a DNS lookup. For our own purposes, we dispose of hosts in the XFF as we consider that fishy as none of our proxy layers would ever append a host and therefore we throw that component of the XFF out.  I don't know if this will be the case for everyone, so my thinking is that maybe you guys can introduce a setting that controls whether host filtering is done on the XFF header values.  If enabled, then I think a simple filter operation on the list of Strings that can consult a regex or multiple (ipv4, ipv6, ipv6 hex compressed) regexes first to drop items that are not deemed to already be IPs.  It can still use InetAdress.getByName after going through the filtering as you are hedging that you won't get a DNS lookup now because all you have left are IPs.  That would keep things pretty simple as opposed to having to take the next step, which is converting the addresses to Array[Byte] and going through InetAddress.getByAddress instead.

I played around with this using the following regexes and things worked well:

      val ipv4 = """^(25[0-5]|2[0-4]\d|[0-1]?\d?\d)(\.(25[0-5]|2[0-4]\d|[0-1]?\d?\d)){3}$""".r
      val ipv6 = "^(?:[0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}$".r
      val ipv6HexComp = "^((?:[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4})*)?)::((?:[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4})*)?)$".r

ktoso:
Thanks for noticing and reporting!

ktoso:
Sounds like a good move in general, but I fear slapping a regex in there – I remember having recently optimised actor path validation how even simple regex can quickly blow up to being very slow (or literally blow up). Parsing IPs not that hard, so we should perhaps parse them using parboiled properly.

Would you be interested in hacking on this @cmbaxter and contributing?

ktoso:
// cc @jrudolph @sirthias

jrudolph:
[...]

PR #19584, opened 2016-01-22
!htt #19388 Removing use of InetAddress.getByName

cmbaxter:
- Remove String apply methods for RemoteAddress (both scala and java)
- Added unit test to show that hostnames in the address are treated as invalid by the parser
- Added a unit test to show that invalid xff is added as a RawHeader instead of model class

### DIFF
--- a/akka-docs/rst/scala/code/docs/http/scaladsl/server/directives/MiscDirectivesExamplesSpec.scala
+++ b/akka-docs/rst/scala/code/docs/http/scaladsl/server/directives/MiscDirectivesExamplesSpec.scala
@@ -8,6 +8,7 @@ import akka.http.scaladsl.model._
 import akka.http.scaladsl.server._
 import headers._
 import docs.http.scaladsl.server.RoutingSpec
+import java.net.InetAddress
 
 class MiscDirectivesExamplesSpec extends RoutingSpec {
 
@@ -17,7 +18,7 @@ class MiscDirectivesExamplesSpec extends RoutingSpec {
     }
 
     // tests:
-    Get("/").withHeaders(`Remote-Address`(RemoteAddress("192.168.3.12"))) ~> route ~> check {
+    Get("/").withHeaders(`Remote-Address`(RemoteAddress(InetAddress.getByName("192.168.3.12")))) ~> route ~> check {
       responseAs[String] shouldEqual "Client's ip is 192.168.3.12"
     }
   }
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/RemoteAddress.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/RemoteAddress.java
@@ -26,9 +26,6 @@ public abstract class RemoteAddress {
     public static RemoteAddress create(InetSocketAddress address) {
         return akka.http.scaladsl.model.RemoteAddress.apply(address);
     }
-    public static RemoteAddress create(String address) {
-        return akka.http.scaladsl.model.RemoteAddress.apply(address);
-    }
     public static RemoteAddress create(byte[] address) {
         return akka.http.scaladsl.model.RemoteAddress.apply(address);
     }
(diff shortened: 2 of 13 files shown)


ktoso:
OK TO TEST

ktoso:
OK TO TEST

drewhk:
Refs #19388

drewhk:
LGTM, but might need a mention in the migration guide.

drewhk:
This does not compile.

cmbaxter:
@drewhk, ya, I noticed that when the build failed about a week ago.  I then went back to the original ticket and commented as to why.  Can you check out my comments and let me know if you have any feedback before I proceed?

https://github.com/akka/akka/issues/19388

drewhk:
I think the approach you outlined in the ticket is the good one. There is only one thing to consider here, is that we will be cutting a 2.4.2 soon (but first an RC2 so not *too* soon) and after that we cannot make binary incompatible changes. So this needs to go in this week. We can take it over if you prefer.

drewhk:
@cmbaxter do you want us to take over this ticket?

cmbaxter:
@drewhk, I apologize again for not getting on this sooner.  Gotta pay the bills with the day job first :)

I want to take a stab at this again right now.  I need to refresh my feature branch from master and then see how much more work it's going to be to fix the logic from x-real-ip.  If I can't get this fully done today then I will pass the torch to you guys.  I'll write back shortly letting you know when I am with this.  Hopefully I will have a new PR done for today.

cmbaxter:
@drewhk, when I pulled down the latest master and integrated it into my feature branch, I see a test in akka-http-core is currently failing.  It's the `	akka.http.impl.engine.client.HighLevelOutgoingConnectionSpec` spec.  I don't think this has anything to do with my changes, but this test either didn't exist before or wasn't failing before.  Do I need to be concerned with this before moving forward to fixing the issue with x-real-ip?

ktoso:
hey @cmbaxter, sorry for the failure on master; 
@johanandren has a fix for it already. issue: https://github.com/akka/akka/issues/19741 pr: https://github.com/akka/akka/pull/19742

drewhk:
That is a know failure that somehow sneaked in. we have a fix pending. Once we merged that we can rerun the build here. Until then ignore that failure.

cmbaxter:
@drewhk, @ktoso, okay, I will ignore the warning.

For the code impl, my suggestion was to use the code in the `IpAddressParsing` trait to vet the raw x-real-ip header and get the bytes for if it valid.  I see two possible approaches here.  Let me know which of these you prefer:

* Add an `object IpAddressParsing` that extends that trait and has a method to vet a string ip and returns an `Option[Array[Byte]]` depending on if the address is valid according to the rules.  That trait is currently package private and this object would have to not be so that I can then leverage if in the `extractClientIp` directive to properly vet that raw header value from x-real-ip.  Probably not ideal to have to expose this functionality that is currently intentionally package private, but this is the easier thing to do.
* Setup`X-Real-Ip` as a properly modeled header (no longer a raw header) and then hook the rules into the parsing (similar to the `Remote-Address` header).  This takes more work but doesn't require exposing that ip vetting logic outside of that `parsing` package.

If there are other better options please let me know.  I want to get this done right first without having to go back and forth on code reviews based on the PR.

drewhk:
I personally prefer the second option (proper header).

cmbaxter:
@drewhk, okay, changes have been committed and the PR amended to contain them.  I'll be watching for the CI run whenever that kicks off.

drewhk:
Argh, something is wrong, there are unrelated comments in this PR!

cmbaxter:
@drewhk, there was an actual compilation failure in akka-docs, on a test case in there that used the old String apply from RemoteAddress.  I have that fixed and am ready to commit that fix as well.  Is there something else wrong?  I noticed one of my commits, which was we merging from the head of my fork's feature branch does not comply.  Do I need to fix that too?

cmbaxter:
Okay, so the only test failure there was the previously mentioned `akka.http.impl.engine.client.HighLevelOutgoingConnectionSpec`.  Am I in the clear here?

cmbaxter:
@drewhk, are we good with this PR considering the test fails were not related to my changes?

drewhk:
Ok, so now we have proper parser rules, so the DNS problem is no longer there. LGTM from my side, but we will need a second reviewer. We can take over from here if modifications are needed.

ktoso:
LGTM, thanks a lot! The failure is known and short term wise has been addressed, long term we're working on it (it's related to OutputTruncated)

PR #19586, opened 2016-01-22
Java HTTP Extension test coverage

johanandren:


akka-http-core/src/test/scala/akka/http/javadsl/HttpExtensionApiSpec.scala
```diff
@@ -0,0 +1,393 @@
...
+          .toMat(
+            Sink.head(),
+            Keep.both[HostConnectionPool, CompletionStage[Pair[Try[HttpResponse], NotUsed]]])
+          .run(materializer)
+
+      val response = pair.second.toCompletableFuture.get(1, TimeUnit.SECONDS)
+    }
+
+    "allow access to the default client https context" in {
+      http.defaultClientHttpsContext.isSecure should not equal (null)
```
rkuhn: this (and the next) should probably test against boolean values, which serves the double purpose of not switching between different settings inadvertently

akka-http-core/src/test/scala/akka/http/javadsl/HttpExtensionApiSpec.scala
```diff
@@ -0,0 +1,393 @@
...
+      http.clientLayer(headers.Host.create("example.com"))
+      val connectionSettings = ClientConnectionSettings.create(system)
+      http.clientLayer(headers.Host.create("example.com"), connectionSettings)
+      http.clientLayer(headers.Host.create("example.com"), connectionSettings, loggingAdapter)
+    }
+
+    "create an outgoing connection 1" in {
+      // this one cannot be tested because it wants to run on port 80
+      pending
+      val flow = http.outgoingConnection("example.com")
```
rkuhn: doesn’t it support Uri of the shape `"localhost:12345"`?

rkuhn:
LGTM apart from these nitpicks

akka-http-core/src/test/scala/akka/http/javadsl/HttpExtensionApiSpec.scala
johanandren: I think I tested and that failed, but I'll double check.

johanandren:
`ScalaUdpMulticastSpec`, unrelated fail

johanandren:
Unrelated test failure. (`docs.io.ScalaUdpMulticastSpec`)

johanandren:
PLS BUILD

johanandren:
Missing coverage is client and server layers, which are only compile time right now, I think we can add coverage for those later and get this PR merged.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -712,14 +713,19 @@ object Http extends ExtensionId[HttpExt] with ExtensionIdProvider {
...
      * Asynchronously triggers the shutdown of the host connection pool.
      *
      * The produced [[Future]] is fulfilled when the shutdown has been completed.
      */
-    def shutdown()(implicit ec: ExecutionContext): Future[Unit] = gatewayFuture.flatMap(_.shutdown())
+    def shutdown()(implicit ec: ExecutionContextExecutor): Future[Done] = gatewayFuture.flatMap(_.shutdown())
+
+    private[http] def toJava = new akka.http.javadsl.HostConnectionPool {
+      override def setup = HostConnectionPool.this.setup
+      override def shutdown(executor: ExecutionContextExecutor): CompletionStage[Done] = HostConnectionPool.this.shutdown()(executor).toJava
```
drewhk: why `ExecutionContextExecutor`?

akka-http-core/src/test/scala/akka/http/javadsl/HttpExtensionApiSpec.scala
```diff
@@ -0,0 +1,447 @@
...
+      val probe = TestSubscriber.manualProbe[IncomingConnection]()
+      val binding = http.bind(host, port, connectionContext, serverSettings, materializer)
+        .toMat(Sink.fromSubscriber(probe), Keep.left)
+        .run(materializer)
+      val sub = probe.expectSubscription()
+      val address = waitFor(binding).localAddress
+      sub.cancel()
+    }
+
+    "properly bind a server 4" in {
```
drewhk: why are these tests replicated 4 times? Isn't it better to have one test case that loops four times (maybe calling `info()` to log progress)?

akka-http-core/src/test/scala/akka/http/javadsl/HttpExtensionApiSpec.scala
```diff
@@ -0,0 +1,447 @@
...
+    }
+
+    "properly bind a server 3" in {
+      val (_, host, port) = TestUtils.temporaryServerHostnameAndPort()
+      val probe = TestSubscriber.manualProbe[IncomingConnection]()
+      val binding = http.bind(host, port, connectionContext, serverSettings, materializer)
+        .toMat(Sink.fromSubscriber(probe), Keep.left)
+        .run(materializer)
+      val sub = probe.expectSubscription()
+      val address = waitFor(binding).localAddress
```
drewhk: address is not used

akka-http-core/src/test/scala/akka/http/javadsl/HttpExtensionApiSpec.scala
```diff
@@ -0,0 +1,447 @@
...
+        .map(req ⇒ HttpResponse.create())
+        .asJava
+      val binding = http.bindAndHandle(flow, host, port, serverSettings, connectionContext, loggingAdapter, materializer)
+
+      val (_, completion) = http.outgoingConnection(ConnectHttp.toHost(host, port))
+        .runWith(Source.single(HttpRequest.create("/abc")), Sink.head(), materializer).toScala
+
+      waitFor(completion)
+      waitFor(binding).unbind()
+    }
```
drewhk: these three looks also similar. Either document in the test header the difference (since it is hard to find), or collapse it into one test case.

drewhk:
I am not a big fan of such repeating test cases, but otherwise LGTM.

akka-http-core/src/test/scala/akka/http/javadsl/HttpExtensionApiSpec.scala
johanandren: The only difference is the overload of the method called, not sure how to document that without repeating the entire signature?

akka-http-core/src/test/scala/akka/http/javadsl/HttpExtensionApiSpec.scala
johanandren: same as below, for different overload with four slightly different paths on the inside of the Http extension

akka-http-core/src/test/scala/akka/http/javadsl/HttpExtensionApiSpec.scala
rkuhn: @drewhk This “near-repetition” was the whole purpose of the ticket, but I agree with your point about the test title.

johanandren: "with four parameters"? I think they all differ in number of parameters across the test class.

johanandren:
Crap. Found tests that didn't check the result enough, and when I made them do that I found: #19658

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
johanandren: So that the Java can pass an executor easier, I think at least, following @rkuhn Java 8 API changes here.

akka-http-core/src/test/scala/akka/http/javadsl/HttpExtensionApiSpec.scala
drewhk: >  think they all differ in number of parameters across the test class.

In this case yes, but not in the others where there is only a number.

drewhk: ah, you sneakily changed it. Good!

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
drewhk: I still don't get it. This still needs an Akka/Scala specific class, not an `Executor`. I.e you still need to wrap the `Executor` in an `ExecutorContextExecutor`, which is exactly the same as if an `ExecutionContext` is required. However this does not allow passing in executors of any other kind.

drewhk:
LGTM. May I merge? Was there something here to be backported?

johanandren:
Go ahead. Unsure about backporting.

rkuhn:
Nothing to be back-ported.

PR #19590, opened 2016-01-23
+str #19589 Add autoFlush option to  OutputStream Sinks

### LINKED ISSUE #19589: No flushing in StreamConverters.fromOutputStream
When using `StreamConverters.fromOutputStream` to create a stream from an `OutputStream`, the implementation provides no way to flush after writes. This is a problem, for example, when writing to STDIN of a child process (at least on Windows), where flushing does not automatically take place.

A workaround is to wrap the `OutputStream` in a `PrintStream` using its `autoFlush` feature.
```scala
StreamConverters.fromOutputStream(() => new PrintStream(process.getOutputStream, true))
```

However it would be better if `StreamConverters.fromOutputStream` would provide this option directly.

m-sp:
Fixes #19589


drewhk:
LGTM, but needs rebase

drewhk:
OK TO TEST

m-sp:
@drewhk rebased onto master

johanandren:
LGTM

Issue #19591, opened 2016-01-23
GraphPartitionSpec


israel:
update "partition to three subscribers" test as per @drewhk request (https://github.com/akka/akka/pull/19461#discussion_r50111355) and add a test for partitioner out of bounds case.

rkuhn:
see #19751

PR #19592, opened 2016-01-23
=str #19591 updated 'partition to three subscribers' to use Sink.seq …

israel:
…and added a new test to cover a case where partitioner return an index out of bounds


akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -469,6 +471,8 @@ final class Broadcast[T](private val outputPorts: Int, eagerCancel: Boolean) ext
 
 object Partition {
 
+  case class PartitionOutOfBoundsException(msg:String) extends RuntimeException(msg) with NoStackTrace
```
hepin1989: it's not a good idea for case class.

israel: what is not a good idea? the "case" part?
I "took" it from other places in the project. for example:
https://github.com/akka/akka/blob/a12451e007d2d40564860f5a6a1a675ded8824d5/akka-stream/src/main/scala/akka/stream/scaladsl/One2OneBidiFlow.scala#L14

hepin1989: A normal class is OK I think.take a look at other part?

israel: what part? I am sorry, I don't understand

drewhk: what is wrong with a normal IndexOutOfBoundsException?

israel: It does not compare well in https://github.com/israel/akka/blob/israel-%2319591/akka-stream-tests/src/test/scala/akka/stream/scaladsl/GraphPartitionSpec.scala#L179

israel:
```
scala> val x = new java.lang.IndexOutOfBoundsException("123")
x: IndexOutOfBoundsException = java.lang.IndexOutOfBoundsException: 123

scala> val y = new java.lang.IndexOutOfBoundsException("123")
y: IndexOutOfBoundsException = java.lang.IndexOutOfBoundsException: 123

scala> x == y
res0: Boolean = false

scala> case class MyException(msg:String) extends java.lang.RuntimeException(msg)
defined class MyException

scala> val a = MyException("123")
a: MyException = MyException: 123

scala> val b = MyException("123")
b: MyException = MyException: 123

scala> a == b
res1: Boolean = true

scala> 
```

drewhk: Just use `expectError()` which return the Throwable, then you can look at the message or call an `isInstanceOf`

drewhk:
This is not super urgent, but I think this should use `IndexOutOfBoundsException`. I closing it for now since the PR queue is filling up, but please reopen once there is something new.

drewhk:
OK TO TEST

drewhk:
Failure was unrelated.

drewhk:
PLS BUILD

johanandren:
PLS BUILD

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -508,7 +512,7 @@ final class Partition[T](outputPorts: Int, partitioner: T ⇒ Int) extends Graph
         val elem = grab(in)
         val idx = partitioner(elem)
         if (idx < 0 || idx >= outputPorts)
-          failStage(new IndexOutOfBoundsException(s"partitioner must return an index in the range [0,${outputPorts - 1}]. returned: [$idx] for input [$elem]."))
+          failStage(PartitionOutOfBoundsException(s"partitioner must return an index in the range [0,${outputPorts - 1}]. returned: [$idx] for input [$elem]."))
```
rkuhn: hmm, this should probably not include the element—we might want to perform an audit that ensure that we do not log `elem.tostring` anywhere (because that could actually be HUGE)

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
rkuhn: :+1:

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
israel: do you want to remove the elem print at all or maybe just print first N characters of it?

rkuhn: just calling `toString` may be of arbitrary cost, so we typically just print the class name (which admittedly is of limited use here)

drewhk:
argh, something was wrong with your rebase. Too many commits here.

israel:
you're right. I will just create a new PR...

israel:
I have opened a new corrected PR https://github.com/akka/akka/pull/19751

drewhk:
Ok, this one can be closed. Thanks @israel for your contribution!

PR #19594, opened 2016-01-23
+str #19593 Fix a potential race condition in FileSubscriber.

### LINKED ISSUE #19593: Potential race condition in FileSubscriber
When Akka Stream `FileSubscriber` completes with an error, the `Promise` is completed before closing the channel (and file). This will generate a race condition where a client wanting to delete the file upon the promise completion will fail (on some systems) because the file is still open.

Solution is to _always_ close the channel before completing the promise.

ochrons:
If an error occurs in the stream, the completionPromise is completed
before the file is closed, leading to potential race condition if the
file is then tried to be deleted upon the promise completion, while it's
still open.

Fixes #19593

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/io/FileSubscriber.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/io/FileSubscriber.scala
@@ -46,7 +46,7 @@ private[akka] class FileSubscriber(f: File, completionPromise: Promise[IOResult]
     super.preStart()
   } catch {
     case ex: Exception ⇒
-      completionPromise.success(IOResult(bytesWritten, Failure(ex)))
+      closeAndComplete(IOResult(bytesWritten, Failure(ex)))
       cancel()
   }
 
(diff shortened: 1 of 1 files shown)


akka-stream/src/main/scala/akka/stream/impl/io/FileSubscriber.scala
```diff
@@ -46,7 +46,7 @@ private[akka] class FileSubscriber(f: File, completionPromise: Promise[IOResult]
     super.preStart()
   } catch {
     case ex: Exception ⇒
-      completionPromise.success(IOResult(bytesWritten, Failure(ex)))
+      closeAndComplete(IOResult(bytesWritten, Failure(ex)))
```
hepin1989: how about a finally block?

hepin1989: `Nonfatal`?

drewhk:
OK TO TEST

drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/impl/io/FileSubscriber.scala
```diff
@@ -70,20 +70,26 @@ private[akka] class FileSubscriber(f: File, completionPromise: Promise[IOResult]
...
+    closeAndComplete(IOResult(bytesWritten, Success(Done)))
+    super.postStop()
+  }
+
+  private def closeAndComplete(result: IOResult): Unit = {
     try {
+      // close the channel/file before completing the promise, allowing the
+      // file to be deleted, which would not work (on some systems) if the
+      // file is still open for writing
       if (chan ne null) chan.close()
```
viktorklang: perhaps check  `chan.isOpen`?

ochrons: "If the channel has already been closed then this method returns immediately.", so no need to do that check ourselves.

viktorklang: Agreed :)

johanandren:
LGTM

PR #19595, opened 2016-01-23
=str #18890 fix throttle

agolubev:
- the constant cost function should is reused instead of allocated per use
- use nanoTime insted of millis
- remove one redundant currentTime call

Ref #18890

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
@@ -9,7 +9,6 @@ import akka.stream.stage._
 import akka.stream._
 
 import scala.concurrent.duration.{ FiniteDuration, _ }
-import scala.util.control.NonFatal
 
 /**
  * INTERNAL API
@@ -26,10 +25,10 @@ private[stream] class Throttle[T](cost: Int,
 
   override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new TimerGraphStageLogic(shape) {
     var willStop = false
-    var lastTokens: Long = 0
-    var previousTime: Long = 0
+    var lastTokens: Long = maximumBurst
+    var previousTime: Long = now()
 
-    val speed = ((cost.toDouble / per.toMillis) * 1024 * 1024).toLong
+    val speed = ((cost.toDouble / per.toNanos) * 1073741824).toLong
     val timerName: String = "ThrottleTimer"
 
     var currentElement: Option[T] = None
@@ -42,21 +41,29 @@ private[stream] class Throttle[T](cost: Int,
         else completeStage()
 
       override def onPush(): Unit = {
-        val timeElapsed = now() - previousTime
-        val currentTokens = Math.min(timeElapsed * speed + lastTokens, scaledMaximumBurst)
         val elem = grab(in)
         val elementCost = scale(costCalculation(elem))
-        if (currentTokens < elementCost)
-          mode match {
-            case Shaping ⇒
-              currentElement = Some(elem)
-              scheduleOnce(timerName, ((elementCost - currentTokens) / speed).millis)
-            case Enforcing ⇒ failStage(new RateExceededException("Maximum throttle throughput exceeded"))
-          }
-        else {
-          lastTokens = currentTokens - elementCost
-          previousTime = now()
+
+        if (lastTokens >= elementCost) {
+          lastTokens -= elementCost
           push(out, elem)
+        } else {
+          val currentTime = now()
+          val currentTokens = Math.min((currentTime - previousTime) * speed + lastTokens, scaledMaximumBurst)
+          if (currentTokens < elementCost)
+            mode match {
+              case Shaping ⇒
+                currentElement = Some(elem)
+                val waitTime = (elementCost - currentTokens) / speed
+                previousTime = currentTime + waitTime
+                scheduleOnce(timerName, waitTime.nanos)
+              case Enforcing ⇒ failStage(new RateExceededException("Maximum throttle throughput exceeded"))
+            }
+          else {
+            lastTokens = currentTokens - elementCost
+            previousTime = currentTime
+            push(out, elem)
+          }
         }
       }
     })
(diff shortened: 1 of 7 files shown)


akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -42,20 +41,21 @@ private[stream] class Throttle[T](cost: Int,
         else completeStage()
 
       override def onPush(): Unit = {
-        val timeElapsed = now() - previousTime
-        val currentTokens = Math.min(timeElapsed * speed + lastTokens, scaledMaximumBurst)
+        val currentTime = now()
+        val currentTokens = Math.min((currentTime - previousTime) * speed + lastTokens, scaledMaximumBurst)
```
drewhk: You don't need to get the time here all the time. The procedure should be:
 - calculate cost of element
 - if there are enough tokens, just remove that amount. No token refresh
 - if there are not enough tokens:
  - add those tokens to currentTokens which arrived since the last refresh timestamp
  - update the refresh timestamp
  - check if now tokens are enough or not

I.e. the idea is to calculate the time-based refreshment tokens less frequently, in larger batches. If the burst size is 4-5x larger than the average element then this saves a lot of current time evaluations.

agolubev: Ok, finally got it. Adding tokens not on each push but only when it's not enough in bucket.

agolubev:
pushed change recommended by @drewhk

drewhk:
@rkuhn can you look at this, since you had the ticket filed and I am not sure all of your ideas are in.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -29,7 +28,7 @@ private[stream] class Throttle[T](cost: Int,
     var lastTokens: Long = 0
     var previousTime: Long = 0
```
rkuhn: This should start out at `scaledMaximumBurst` and `now()`—it currently depends on the fact that `now()` is sufficiently greater than zero when the first element arrives.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -42,21 +41,27 @@ private[stream] class Throttle[T](cost: Int,
...
+          lastTokens -= elementCost
           push(out, elem)
+        } else {
+          val currentTime = now()
+          val currentTokens = Math.min((currentTime - previousTime) * speed + lastTokens, scaledMaximumBurst)
+          if (currentTokens < elementCost)
+            mode match {
+              case Shaping ⇒
+                currentElement = Some(elem)
+                scheduleOnce(timerName, ((elementCost - currentTokens) / speed).nanos)
```
rkuhn: since we didn’t update any values, this will not solve the problem for elements that cost more than the maximumBurst allows—these will be deferred indefinitely

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
drewhk: yep, good point.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
rkuhn: Or, no, I misread: in that case the calculated bucket fill level will be incorrect, it will be zero instead of taking into account that the timer will almost always be late.

drewhk: Why? The element is delayed until the time comes where the number of tokens would be exactly the same as needed. In the onTimer then the element is emitted, and tokens set to zero (since all of them were consumed).

drewhk: > it will be zero instead of taking into account that the timer will almost always be late.

Ok, I didn't really care about that drift, but this can be solved by recording here the time and then instead of setting to zero, calculate the tokens that were accumulated since the delay.

rkuhn: exactly; or we just set the negative value right here and don’t modify anything in onTimer since the next element coming in will do the right calculation anyway

drewhk: That one is also true. Assuming that the timer misbehaves for some reason, can a negative number there provide any issues? I believe not, but I would be happy to hear your opinion.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1356,7 +1356,7 @@ trait FlowOps[+Out, +Mat] {
    * '''Cancels when''' downstream cancels
    */
   def throttle(elements: Int, per: FiniteDuration, maximumBurst: Int, mode: ThrottleMode): Repr[Out] =
-    throttle(elements, per, maximumBurst, _ ⇒ 1, mode)
+    throttle(elements, per, maximumBurst, ConstantFun.oneInt, mode)
```
rkuhn: we should add in the ScalaDoc that the resulting rate will be lower than requested unless `maximumBurst.toDouble / elements * per > 2 * schedulerPeriod`

rkuhn:
LGTM apart from comments, thanks @agolubev !

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
drewhk: This is in that raw form is not very informative to the user. Let's add a human digestible sentence as well.

rkuhn:
It is not crucial to include this in RC3, it can wait until Monday (so don’t be worried if it misses the train).

agolubev:
I have everything except formula in ScalaDoc. Can commit in this state if it's Ok.

rkuhn:
OK TO TEST

rkuhn:
Thanks a lot, will merge after validation—I’ll ponder the correct ScalaDoc update for a bit :-)

PR #19606, opened 2016-01-25
!htc #19514 make config classes easier to evolve bin-compat (rebase)

johanandren:
Refs #19514 

Rebase of #19566 onto current master


akka-http-core/src/main/scala/akka/http/javadsl/settings/ServerSettings.scala
```diff
@@ -0,0 +1,69 @@
...
+  def withVerboseErrorMessages(newValue: Boolean): ServerSettings = self.copy(verboseErrorMessages = newValue)
+  def withResponseHeaderSizeHint(newValue: Int): ServerSettings = self.copy(responseHeaderSizeHint = newValue)
+  def withBacklog(newValue: Int): ServerSettings = self.copy(backlog = newValue)
+  def withSocketOptions(newValue: java.lang.Iterable[SocketOption]): ServerSettings = self.copy(socketOptions = newValue.asScala.toList)
+  def withDefaultHostHeader(newValue: Host): ServerSettings = self.copy(defaultHostHeader = newValue.asScala)
+  def withParserSettings(newValue: ParserSettings): ServerSettings = self.copy(parserSettings = newValue.asScala)
+  def withWebsocketRandomFactory(newValue: java.util.function.Supplier[Random]): ServerSettings = self.copy(websocketRandomFactory = () ⇒ newValue.get())
+
+  /** INTERNAL API */
+  protected def self = this.asInstanceOf[ServerSettingsImpl]
```
rkuhn: This should better be expressed as a self-type, removing the need for a cast as well as avoiding exposure in user API.

johanandren: Feels like a very weird design decision to me, that the abstract class knows about the implementation.

akka-http-core/src/main/scala/akka/http/javadsl/settings/ParserSettings.scala
```diff
@@ -0,0 +1,78 @@
...
+  def withHeaderValueCacheLimits(newValue: ju.Map[String, Int]): ParserSettings = self.copy(headerValueCacheLimits = newValue.asScala.toMap)
+  def withIncludeTlsSessionInfoHeader(newValue: Boolean): ParserSettings = self.copy(includeTlsSessionInfoHeader = newValue)
+
+  // special ---
+
+  def withCustomMethods(methods: HttpMethod*): ParserSettings = {
+    val map = methods.map(m ⇒ m.name -> m.asScala).toMap
+    self.copy(customMethods = map.get)
+  }
+  def withCustomStatusCodes(codes: StatusCode*): ParserSettings = {
```
patriknw: `@varargs` annotation these since it is javadsl?

akka-http-core/src/main/scala/akka/http/javadsl/settings/ServerSettings.scala
patriknw: I guess it is about utilizing the generated copy method of the case class

patriknw:
LGTM, check the varargs and see if it can be solved with self type

akka-http-core/src/main/scala/akka/http/javadsl/settings/ServerSettings.scala
rkuhn: If the self-type also leaks some implementation details (i.e. methods with $ in their name, visible from Java) then we’ll have to move the implementation into the actual Impl class (i.e. not touch the copy method from here).

johanandren: Does not compile with selftype because then `self` is not available in scaladsl `ServerSettings` where it also used :(. 

Trying out self type with javap shows no public stuff leaked

johanandren: I can make the self type available with a protected method though.

johanandren: Or just selftype both. Ok. Note to self: need to stop thinking out loud in PR-comments.

rkuhn: Scala’s `protected` is actually `public` in the bytecode; for `self` to work in the Scala part you’ll need to add the same self-type annotation there as well

akka-docs/rst/scala/stream/migration-guide-2.0-2.4-scala.rst
```diff
@@ -79,6 +79,17 @@ In Akka 2.4.x this is formulated like so:
...
+Changes in Akka HTTP
+====================
+
+Routing settings parameter name
+-------------------------------
+
+``RoutingSettings`` were previously the only setting available on ``RequestContext``,
+and were accessible via ``settings``. We now made it possible to configure the parsers
+settings as well, so ``RoutingSettings`` is now ``routingSettings`` and ``ParserSetttings`` is
+now accessible via ``parserSettings``.
```
rkuhn: is this Scala-only or should there be a Java migration guide entry as well?

akka-docs/rst/scala/code/docs/http/scaladsl/server/directives/BasicDirectivesExamplesSpec.scala
```diff
@@ -183,7 +183,7 @@ class BasicDirectivesExamplesSpec extends RoutingSpec {
   }
   "withSettings-0" in compileOnlySpec {
     //#withSettings-0
-    val special = RoutingSettings(system).copy(fileIODispatcher = "special-io-dispatcher")
+    val special = RoutingSettingsImpl(system).copy(fileIODispatcher = "special-io-dispatcher")
```
rkuhn: should this be documented with the …Impl type?

akka-docs/rst/scala/code/docs/http/scaladsl/server/directives/BasicDirectivesExamplesSpec.scala
```diff
@@ -687,7 +687,7 @@ class BasicDirectivesExamplesSpec extends RoutingSpec {
   "extractSettings-examples" in {
     //#extractSettings-examples
     val route =
-      extractSettings { settings: RoutingSettings =>
+      extractSettings { settings: RoutingSettingsImpl =>
```
rkuhn: is this necessary to work?

akka-docs/rst/scala/code/docs/http/scaladsl/server/directives/BasicDirectivesExamplesSpec.scala
rkuhn: Looking into the code: this should probably just be renamed to RoutingSettings and changed to a non-case class later—the HTTP DSL is still marked experimental so that we can do this.

2m:
LGTM

akka-docs/rst/scala/stream/migration-guide-2.0-2.4-scala.rst
johanandren: Scala only it seems

akka-docs/rst/scala/code/docs/http/scaladsl/server/directives/BasicDirectivesExamplesSpec.scala
johanandren: Yes, I'm leaving this be right now.

akka-docs/rst/scala/code/docs/http/scaladsl/server/directives/BasicDirectivesExamplesSpec.scala
johanandren: There is no `RoutingSettings` only `Impl`

akka-docs/rst/scala/code/docs/http/scaladsl/server/directives/BasicDirectivesExamplesSpec.scala
ktoso: Whoops, sorry about forgetting that one. Yeah it should follow the same pattern as the others, Impl case class and java/scala  dsl classes

akka-docs/rst/scala/stream/migration-guide-2.0-2.4-scala.rst
ktoso: Java too, same text and we can add "all settings are convieniently modifiable with prepared `with...(...)` methods now"

-- 
Cheers,
Konrad 'ktoso’ Malawski
Akka @ Typesafe

On 25 January 2016 at 23:50:01, Roland Kuhn (notifications@github.com) wrote:

In akka-docs/rst/scala/stream/migration-guide-2.0-2.4-scala.rst:

> @@ -79,6 +79,17 @@ In Akka 2.4.x this is formulated like so:
>   
>  .. includecode:: ../code/docs/stream/MigrationsScala.scala#expand-state
>   
> +Changes in Akka HTTP
> +====================
> +
> +Routing settings parameter name
> +-------------------------------
> +
> +``RoutingSettings`` were previously the only setting available on ``RequestContext``,
> +and were accessible via ``settings``. We now made it possible to configure the parsers
> +settings as well, so ``RoutingSettings`` is now ``routingSettings`` and ``ParserSetttings`` is
> +now accessible via ``parserSettings``.
is this Scala-only or should there be a Java migration guide entry as well?

akka-docs/rst/scala/code/docs/http/scaladsl/server/directives/BasicDirectivesExamplesSpec.scala
johanandren: Reworking it according to ktosos note right now.

akka-docs/rst/scala/stream/migration-guide-2.0-2.4-scala.rst
johanandren: Butbut, looking at `RequestContext` in javadsl I don't see any `RoutingSetting` s?

rkuhn: Then Konrad has probably not done the same change for Java.

ktoso: I might have missed japi for this one, sorry! It should be 1:1 to scala api

johanandren:
Correct, apparently I'm suffering a very rare neurological condition causing me to write `ParserSettings` whenever I mean `RouterSettings`.

johanandren:
PLS BUILD

johanandren:
(unrelated fail)

PR #19612, opened 2016-01-25
=doc Add intended use section to OSGi doc

rocketraman:
In the discussion at https://github.com/typesafehub/config/issues/79 @rkuhn provided some very useful information about the intended use of Akka within an OSGi environment. I think this information deserves to be part of the formal documentation.

### DIFF
--- a/akka-docs/rst/additional/osgi.rst
+++ b/akka-docs/rst/additional/osgi.rst
@@ -100,6 +100,14 @@ Configuring the OSGi Framework
 To use Akka in an OSGi environment, the container must be configured such that the ``org.osgi.framework.bootdelegation``
 property delegates the ``sun.misc`` package to the boot classloader instead of resolving it through the normal OSGi class space.
 
+Intended Use
+------------
+
+Akka only supports the usage of an ActorSystem strictly confined to a single OSGi bundle, where that bundle contains or imports
+all of the actor system's requirements. This means that the approach of offering an ActorSystem as a service to which Actors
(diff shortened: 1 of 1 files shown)


drewhk:
LGTM

rocketraman:
Note I'm not quite sure what @rkuhn meant be "its dependency subtree" and it would be nice to clarify that language. If he meant bundles that import packages from the bundle containing the actor system at the class level, then the actor system itself would not be available on the subtree bundles, only the ActorSystem class (and perhaps not even that, depending on how the ActorSystem class is accessed). If he meant bundles dependent on an ActorSystem service exposed by that bundle -- well that is the case we are describing is not recommended or supported. So I'd like to do some editing / add some clarification there before merge.

rkuhn:
Thanks @rocketraman! What I meant is that the ActorSystem and everything it needs at runtime is confined to the bundle, all dependencies that it needs are dependencies of that bundle and therefore OSGi should “just work” (if the ActorSystem knows how to access all those goodies by using the right ClassLoader). Another way to say this is that the fact that a bundle uses an ActorSystem should always be an implementation detail that is not exposed to the outside—ActorRefs may be exposed, though.

rocketraman:
@rkuhn Thanks for the clarification! I've updated the text accordingly, as well as rebased onto master. Please review!

drewhk:
LGTM a second time :)

rkuhn:
LGTM, thanks @rocketraman !

Issue #19551, opened 2016-01-21
PersistentActor crashed when internalStash  throwing overflow exception 


MQ-EL:
Just need to drop incoming messages than crashed for persistent actor in some time. May be we can support this feature?

MQ-EL:
When PA recoverying, the nr of bussiness domain msg received up to ```stash-capacity``` sometime , so crashed. then PA start recoverying again -> stash overyflow -> crashed ... I need do more work to prevent PA  restart repeatly in somewhere, just drop bussiness domain msgs in PA is more natual?

MQ-EL:
```scala
  private def recovering(recoveryBehavior: Receive) = new State {
    ...
    override def stateReceive(receive: Receive, message: Any) = message match {
      case ... ⇒ ... 
      case other ⇒
        internalStash.stash()  // throw exception, cause recovery failure
    }
  }
```
Give a chance to catch this exception to ignore incomming bussiness domain message?

patriknw:
Perhaps you can handle the exception in the parent (supervisor) actor? Perhaps use the `BackoffSupervisor`?

That said, I think you are right that we should improve the default behavior here.

MQ-EL:
I used ```BackoffSupervisor``` to prevent some true exception, because it impacts performance to system(db or bussiness system used PA)  when PAs restart, especially when a large nr of PAs restart. However the worst is it will let some live PAs persist failure , which will make more restarted PAs.

 I mean we should avoid PA restart as mush as possible.

may be we can add  config ```ignore-stash-overflow``` in persistent config? and add ```def tryStash(): Boolean ``` in ```Stash``` trait to avoid try catch exception every time when stash full.

patriknw:
yeah, but you should configure the `BackoffSupervisor` to stop the PA in case of such exceptions, it will then start it up again after backoff. That is new feature in 2.4.1.

MQ-EL:
Drop business msgs directly will promote performce, becasue it avoid insignificance recovery time or cpu. 

IMO, it can be used with ```BackoffSupervisor``` together for better protect PAs

patriknw:
I agree

MQ-EL:
May i create a PR to add ```tryStash()``` method & ```ignore-stash-overflow``` config ? Or waiting a better solution?

patriknw:
PR is very much welcome (that is often the quickest way to get things done ;-))

PR #19626, opened 2016-01-27
=per #19551 support stash overflow strategy

MQ-EL:
refs #19551

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
@@ -5,14 +5,14 @@
 package akka.persistence
 
 import java.util.concurrent.atomic.AtomicInteger
+import java.util.UUID
+
 import scala.collection.immutable
 import scala.util.control.NonFatal
-import akka.actor.Stash
-import akka.actor.StashFactory
+import akka.actor.DeadLetter
+import akka.actor.StashOverflowException
 import akka.event.Logging
 import akka.event.LoggingAdapter
-import akka.actor.ActorRef
-import java.util.UUID
 
 /**
  * INTERNAL API
@@ -37,7 +37,7 @@ private[persistence] object Eventsourced {
  * Scala API and implementation details of [[PersistentActor]], [[AbstractPersistentActor]] and
  * [[UntypedPersistentActor]].
  */
-private[persistence] trait Eventsourced extends Snapshotter with Stash with StashFactory with PersistenceIdentity with PersistenceRecovery {
+private[persistence] trait Eventsourced extends Snapshotter with PersistenceStash with PersistenceIdentity with PersistenceRecovery {
   import JournalProtocol._
   import SnapshotProtocol.LoadSnapshotResult
   import Eventsourced._
@@ -147,6 +147,20 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
     log.warning("Rejected to persist event type [{}] with sequence number [{}] for persistenceId [{}] due to [{}].",
       event.getClass.getName, seqNr, persistenceId, cause.getMessage)
   }
+
+  private def stashInternally(currMsg: Any): Unit =
+    try internalStash.stash() catch {
+      case e: StashOverflowException ⇒
+        internalStashOverflowStrategy match {
+          case DiscardToDeadLetterStrategy ⇒
+            val snd = sender()
+            context.system.deadLetters.tell(DeadLetter(currMsg, snd, self), snd)
+          case ReplyToStrategy(response) ⇒
+            sender() ! response
+          case ThrowOverflowExceptionStrategy ⇒
+            throw e
+        }
+    }
 
   private def startRecovery(recovery: Recovery): Unit = {
     changeState(recoveryStarted(recovery.replayMax))
@@ -459,7 +473,8 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
         }
         changeState(recovering(recoveryBehavior))
         journal ! ReplayMessages(lastSequenceNr + 1L, toSnr, replayMax, persistenceId, self)
-      case other ⇒ internalStash.stash()
+      case other ⇒
+        stashInternally(other)
     }
   }
 
@@ -474,7 +489,7 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
    * All incoming messages are stashed.
    */
   private def recovering(recoveryBehavior: Receive) = new State {
-    override def toString: String = s"replay started"
+    override def toString: String = "replay started"
     override def recoveryRunning: Boolean = true
 
     override def stateReceive(receive: Receive, message: Any) = message match {
@@ -496,7 +511,7 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
       case ReplayMessagesFailure(cause) ⇒
         try onRecoveryFailure(cause, event = None) finally context.stop(self)
       case other ⇒
-        internalStash.stash()
+        stashInternally(other)
     }
   }
 
@@ -626,7 +641,7 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
 
     override def stateReceive(receive: Receive, message: Any) =
       if (common.isDefinedAt(message)) common(message)
-      else internalStash.stash()
+      else stashInternally(message)
 
     override def onWriteMessageComplete(err: Boolean): Unit = {
       pendingInvocations.pop() match {
--- a/akka-persistence/src/main/scala/akka/persistence/Persistence.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/Persistence.scala
@@ -9,6 +9,7 @@ import java.util.function.Consumer
 import akka.actor._
 import akka.event.{ Logging, LoggingAdapter }
 import akka.persistence.journal.{ EventAdapters, IdentityEventAdapters }
+import akka.util.Collections.EmptyImmutableSeq
 import akka.util.Helpers.ConfigOps
 import com.typesafe.config.Config
 import scala.annotation.tailrec
@@ -108,6 +109,15 @@ trait PersistenceRecovery {
   //#persistence-recovery
 }
 
+trait PersistenceStash extends Stash with StashFactory {
+  /**
+   * The returned [[StashOverflowStrategy]] object determines how to handle the message failed to stash
+   * when the internal Stash capacity exceeded.
+   */
+  def internalStashOverflowStrategy: StashOverflowStrategy =
+    Persistence(context.system).defaultInternalStashOverflowStrategy
+}
+
 /**
  * Persistence extension provider.
  */
@@ -129,7 +139,6 @@ class Persistence(val system: ExtendedActorSystem) extends Extension {
 
   private def log: LoggingAdapter = Logging(system, getClass.getName)
 
-  private val DefaultPluginDispatcherId = "akka.persistence.dispatchers.default-plugin-dispatcher"
   private val NoSnapshotStorePluginId = "akka.persistence.no-snapshot-store"
 
   private val config = system.settings.config.getConfig("akka.persistence")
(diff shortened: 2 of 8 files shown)


hepin1989:
how about provide `StashOverflowStrategy`,like `Fail`,Drop` etc.

MQ-EL:
We can override onStashOverflow method like onPersistRejected. Let business do what they want to?

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -148,6 +148,16 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
       event.getClass.getName, seqNr, persistenceId, cause.getMessage)
   }
 
+  /**
+    * Called when the size of the internal stash is exceeds the capacity of the Stash
+    * 
+    * @param message the message which failed to enqueue
+    */
+  protected def onStashOverflow(message: Any): Unit = {
```
patriknw: we must protect things while invoking this method in case user code does something bad, e.g. `persist` should not be allowed

patriknw:
OK TO TEST

patriknw:
it will be interesting to see if this is binary compatible, I think not

patriknw:
I like this

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
MQ-EL: two ways:
1. write some warning comment of above method to prompt users. constraint weakly but more flexible, imagine(may be not appropriately):  some users need log this event in custom format.
2. or more strictly, predefine sealed ```StashOverflowStrategy``` trait as @hepin1989 said. and  then provide ```DiscardToDeadLettersStrategy```,  ```ThrowOverflowExceptionStrategy```, ```ReplyRejectMessageStrategy```

How are u feeling?

MQ-EL:
eh, I will restore the fault behavior, which throwing overflow exception, after the above discussion is over.

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
patriknw: Users will do the wrong thing, and we get issues, support cases and such, so I think alternative 2 is appropriate.

MQ-EL:
sorry, i'm late. does ```PersistentView``` need to support this feature. i see it was deprecated

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
MQ-EL: fine

patriknw:
I saw some question about PersistentView, don't know where it is, but you don't have to care about PersistentView. It is being phased out.

MQ-EL:
sorry, I just deleted it(misoperation).

MQ-EL:
I don't understand where the code that BC check failed? :(

patriknw:
`[error]  * method tryStash()Boolean in trait akka.actor.StashSupport does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("akka.actor.StashSupport.tryStash")
`

You cannot add new methods to a trait. I think we will have to introduce a new trait.

you can run it locally with `mimaReportBinaryCompatibility`

There was also an ordinary test failure related to tryStash.

MQ-EL:
Introduce a new trait instead of ```stashsupport```? thus, the old ```StashSupport``` need to change the visibility of some fields. Does this surround the BC check? it is some burden of maintains, isn't?

MQ-EL:
The last solution i think is add try catch clause in every ```stash()```  present in  ```EventSourced``` trait.  But i don't like it.

MQ-EL:
for now, the implement of strategy use ```try catch``` clause, If it is not so eligible, we can change the ```PersistenceStash``` trait by using reflect to get ```capacity``` and ```theStash```.  anybody has some good idea to share?

patriknw:
I don't see a way to add this feature to StashSupport without duplicating pretty much everything. Not nice. Reflection is probably also not wanted. The try catch solution is alright, but there is a risk that it introduces a lot of overhead once the stash capacity is reach (one new exception for each incoming message). Hmm, 

Do you have any thoughts about this @rkuhn ?

rkuhn:
If the exception extends `NoStackTrace` then I don’t see an issue.

patriknw:
Good suggestion, let's change that (but make sure the exc message contains enough info)

MQ-EL:
Good idea, but it lost stack trace forever, even in no-persistence environment. I think the ```stash``` should be changed in 2.5.x, because it does not follow **Reference Transparent**. IMO, Whether the ```StashOverflowException``` is real exception depends on user's domain. 
I think the original ```exc msg``` contains enough info, did u have anything to add @patriknw

MQ-EL:
The overflow strategy should  be moved to Stash in future?

patriknw:
I didn't have anything in mind aside from that you should check it. I see that it includes `self`, so that should be enough.

However, it also contains `toString` of the `currMsg` and that can be too costly, so change that to `currMsg.getClass.getName`

MQ-EL:
done, please review

akka-persistence/src/main/scala/akka/persistence/PersistentActor.scala
```diff
@@ -31,6 +30,9 @@ final case class DeleteMessagesSuccess(toSequenceNr: Long)
  */
 final case class DeleteMessagesFailure(cause: Throwable, toSequenceNr: Long)
 
+@SerialVersionUID(1L)
+case object RefusedToStash
```
patriknw: this would require Java api also, but I don't think we should provide a default class for this at all, better that the application defines the message

akka-persistence/src/main/scala/akka/persistence/PersistentActor.scala
```diff
@@ -99,6 +101,58 @@ object Recovery {
...
+ */
+case object DiscardToDeadLetter extends StashOverflowStrategy {
+  /**
+   * Java API: get the singleton instance
+   */
+  def getInstance = this
+
+  override def handleFailure(message: Any, sender: ActorRef, original: StashOverflowException)
+                            (implicit context: ActorContext, receiver: ActorRef): Unit = {
+    context.system.deadLetters.tell(DeadLetter(message, sender, receiver), sender)
```
patriknw: I understand that you used `ActorContext` here to be able to grab the implicit from the enclosing actor, but I think we should not pass the ActorContext around, even though it's only to our own implementations. 

Either change the param to `system`, or perhaps better remove the abstract `handleFailure` method and handle the implementation with a pattern match at the call site instead.

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
```diff
@@ -105,8 +105,18 @@ trait PersistenceRecovery {
    * To skip recovery completely return `Recovery.none`.
    */
   def recovery: Recovery = Recovery()
-  //#persistence-recovery
 }
+//#persistence-recovery
+
+//#persistence-stash
```
patriknw: this special comment is used from documentation includes, but not needed for this

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
```diff
@@ -105,8 +105,18 @@ trait PersistenceRecovery {
...
+trait PersistenceStash extends Stash with StashFactory {
+  /**
+   * The returned [[StashOverflowStrategy]] object determine how to handle the received message when the
+   * size of Stash exceeds the capacity of the Stash.
+   */
+  def stashOverflowStrategy: StashOverflowStrategy = ThrowOverflowExceptionStrategy
```
patriknw: we should mention this method in the reference documentation, there is already some words about the stash in the docs, perhaps we should move that to a new section and include this

akka-persistence/src/main/scala/akka/persistence/PersistentActor.scala
MQ-EL: Yeh, it will reduce the nr of params

akka-persistence/src/main/scala/akka/persistence/PersistentActor.scala
MQ-EL: Make strategy configable in the reference.conf file?

patriknw: That is a good idea.

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
MQ-EL: I'm sorry for late (china new year). The words in the docs is short of the relation between stash and mailbox, eg. the deffience between inherited user's ```unstash()``` and ```unstashAll()```. IMO, there is a real needed to make some words in the docs for a new stash section. If you don't mind, I will commit some words in this pr in recent days.

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
MQ-EL: Thanks for pointing out

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
```diff
@@ -153,16 +161,18 @@ class Persistence(val system: ExtendedActorSystem) extends Extension {
...
 
   /** Check for default or missing identity. */
   private def isEmpty(text: String) = text == null || text.length == 0
 
   /** Discovered persistence journal and snapshot store plugins. */
-  private val journalPluginExtensionId = new AtomicReference[Map[String, ExtensionId[PluginHolder]]](Map.empty)
-
-  /** Discovered persistence snapshot store plugins. */
-  private val snapshotPluginExtensionId = new AtomicReference[Map[String, ExtensionId[PluginHolder]]](Map.empty)
+  private val pluginExtensionId = new AtomicReference[Map[String, ExtensionId[PluginHolder]]](Map.empty)
```
MQ-EL: AFAIK, the ```journal``` and ```snapshot``` holders was placed into one collection.

akka-persistence/src/main/scala/akka/persistence/PersistentActor.scala
```diff
@@ -99,6 +99,58 @@ object Recovery {
...
+ * @param response the message replying to sender with
+ */
+final case class ReplyToStrategy(response: Any) extends StashOverflowStrategy
+
+/**
+ * Implement this interface in order to configure the stashOverflowStrategy for
+ * the internal stash of persistent actor.
+ * An instance of this class must be instantiable using a no-arg constructor.
+ */
+trait StashOverflowStrategyConfigurator {
```
MQ-EL: Is this configurator's design properly? too complicated?

akka-persistence/src/main/scala/akka/persistence/Persistence.scala
```diff
@@ -108,6 +109,14 @@ trait PersistenceRecovery {
   //#persistence-recovery
 }
 
+trait PersistenceStash extends Stash with StashFactory {
+  /**
+   * The returned [[StashOverflowStrategy]] object determines how to handle the message failed to stash
+   * when the internal Stash capacity exceeded.
+   */
+  def internalStashOverflowStrategy: StashOverflowStrategy
```
MQ-EL: the name added ```internal``` prefix for distinguish to inherited user's stash

akka-persistence/src/main/scala/akka/persistence/PersistentActor.scala
patriknw: We don't want these to be extendable, but I see that you need it to be able to support the `ReplyToStrategy` with a custom message. I'm fine with this.

```
cfg.getString("internal-stash-overflow-strategy" match {
  case "overflow-exception" => ThrowOverflowExceptionStrategy
  case "

patriknw:
LGTM

patriknw:
I think this is done. please squash to one commit, @MQ-EL

MQ-EL:
doesn't need to add reference doc?

patriknw:
Ah, yes, docs must be included

MQ-EL:
Hmm, Does this PR has some releation with ```akka.stream.tck.VirtualProcessorTest```? @patriknw

drewhk:
LGTM

rkuhn:
Thanks @MQ-EL !

Issue #19423, opened 2016-01-13
add statefulMapConcat operator


rkuhn:
This will avoid the implementation of many a GraphStage (see [here](https://github.com/akka/akka/pull/19408#issuecomment-171236560) for previous discussion):
~~~scala
FlowOps {
  ...
  def statefulMapConcat[T](f: () => Out => Iterable[T]): Repr[T]
}
~~~

drewhk:
Or implement unfold in terms of a Flow instead of a Source. termination of the unfold round maps to a pull, and the current source unfold maps to Source.single(seed).unfold().

Same can be done with fold, by introducing a terminal signal that expresses that a new batch to be folded is started.

(Sidenote, these were the two basic operations that I came up in the early days of streams design)

viktorklang:
@drewhk I agree. I'd much rather see a generalized unfold.

drewhk:
Btw, when I say "fold" of course I mean a new name since the current fold maps to something already taken and well understood ("foldBatching" or something like that can be the name of the new one).

viktorklang:
@drewhk Why is that needed when you can add a mapConcat(identity) after?

drewhk:
Fold emits only one result once the upstream finished. What I propose can emit before the upstream is finished. Useful for batch/aggregate multiple upstream elements into less downstream elements.

viktorklang:
@drewhk
Wouldn't unfold (+ mapConcat for the Iterable-case) do just fine?

drewhk:
No, that is the opposite, one-to-many.

viktorklang:
@drewhk
If we generalize unfold to a Flow then it is definitely encodable:
unfold(Seq.empty[T])( (state, out) => if (someCondition(out)) Some((out :: state, Nil)) else … ).mapConcat(identity)

rkuhn:
Guys, what you are discussing is different from what this issue is about, let’s please stay on topic. I’m looking for a combinator that can express “stateful map” as well as “hysteresis” and the above is the simplest that I could come up with. Modeling that as a generalized unfold will require multiple operators and even more function literals to be provided; it is of course possible but that is beside the point.

In fact, my proposal can be the efficient basis for a generalized unfold if you want to write one. The reverse is not true.

agolubev:
@rkuhn, I started this and it turned out that mapConcat can just use implementation of statefulMapConcat. Still it's not clear from API where the state is. 

[...]

PR #19634, opened 2016-01-28
=str #19423 add statefulMapConcat

agolubev:
Ref #19423 
not sure how to fix InterpreterSupervisionSpec."resume when MapConcat throws" - not it's commented out

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
@@ -303,26 +303,6 @@ class InterpreterSupervisionSpec extends AkkaSpec with GraphInterpreterSpecKit {
       lastEvents() should be(Set(OnNext(3)))
     }
 
-    "resume when MapConcat throws" in new OneBoundedSetup[Int](Seq(
-      MapConcat((x: Int) ⇒ if (x == 0) throw TE else List(x, -x), resumingDecider))) {
-      downstream.requestOne()
-      lastEvents() should be(Set(RequestOne))
-      upstream.onNext(1)
-      lastEvents() should be(Set(OnNext(1)))
-      downstream.requestOne()
-      lastEvents() should be(Set(OnNext(-1)))
-
-      downstream.requestOne()
-      lastEvents() should be(Set(RequestOne))
-      upstream.onNext(0) // boom
-      lastEvents() should be(Set(RequestOne))
-
-      upstream.onNext(2)
-      lastEvents() should be(Set(OnNext(2)))
-      downstream.requestOne()
-      lastEvents() should be(Set(OnNext(-2)))
-    }
-
     "restart when Collect throws" in {
       // TODO can't get type inference to work with `pf` inlined
       val pf: PartialFunction[Int, Int] =
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowStatefulMapConcatSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowStatefulMapConcatSpec.scala
@@ -1,0 +1,85 @@
+/**
+ * Copyright (C) 2016 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream.scaladsl
+
+import akka.stream.testkit.scaladsl.TestSink
+import akka.stream.{ ActorMaterializer, ActorAttributes, Supervision, ActorMaterializerSettings }
+import akka.stream.testkit._
+
+import scala.util.control.NoStackTrace
+
+class FlowStatefulMapConcatSpec extends AkkaSpec with ScriptedTest {
+
+  val settings = ActorMaterializerSettings(system)
+    .withInputBuffer(initialSize = 2, maxSize = 16)
+  implicit val materializer = ActorMaterializer(settings)
+  val ex = new Exception("TEST") with NoStackTrace
+
+  "A StatefulMapConcat" must {
+
+    "work in happy case" in {
+      val script = Script(
+        Seq(2) -> Seq(),
+        Seq(1) -> Seq(1, 1),
+        Seq(3) -> Seq(3),
+        Seq(6) -> Seq(6, 6, 6))
+      TestConfig.RandomTestRange foreach (_ ⇒ runScript(script, settings)(_.statefulMapConcat(() ⇒ {
+        var prev: Option[Int] = None
+        x ⇒ prev match {
+          case Some(e) ⇒
+            prev = Some(x)
+            (1 to e) map (_ ⇒ x)
+          case None ⇒
+            prev = Some(x)
+            List.empty[Int]
+        }
+      })))
+    }
+
+    "be able to restart" in {
+      Source(List(2, 1, 3, 4, 1)).statefulMapConcat(() ⇒ {
+        var prev: Option[Int] = None
+        x ⇒ {
+          if (x % 3 == 0) throw ex
+          prev match {
+            case Some(e) ⇒
+              prev = Some(x)
+              (1 to e) map (_ ⇒ x)
+            case None ⇒
+              prev = Some(x)
+              List.empty[Int]
+          }
+        }
+      }).withAttributes(ActorAttributes.supervisionStrategy(Supervision.restartingDecider))
+        .runWith(TestSink.probe[Int])
+        .request(2).expectNext(1, 1)
+        .request(4).expectNext(1, 1, 1, 1)
+        .expectComplete()
+    }
+
+    "be able to resume" in {
+      Source(List(2, 1, 3, 4, 1)).statefulMapConcat(() ⇒ {
+        var prev: Option[Int] = None
+        x ⇒ {
+          if (x % 3 == 0) throw ex
+          prev match {
+            case Some(e) ⇒
+              prev = Some(x)
+              (1 to e) map (_ ⇒ x)
+            case None ⇒
+              prev = Some(x)
+              List.empty[Int]
+          }
+        }
+      }).withAttributes(ActorAttributes.supervisionStrategy(Supervision.resumingDecider))
+        .runWith(TestSink.probe[Int])
+        .request(2).expectNext(1, 1)
+        .requestNext(4)
+        .request(4).expectNext(1, 1, 1, 1)
+        .expectComplete()
+    }
+
+  }
+
+}
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -21,6 +21,7 @@ import scala.util.{ Failure, Success, Try }
 import akka.stream.ActorAttributes.SupervisionStrategy
 import scala.concurrent.duration.{ FiniteDuration, _ }
 import akka.stream.impl.Stages.DefaultAttributes
+import akka.NotUsed
 
 /**
  * INTERNAL API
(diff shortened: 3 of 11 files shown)


akka-stream/src/main/scala/akka/stream/javadsl/SubFlow.scala
```diff
@@ -163,7 +163,36 @@ class SubFlow[-In, +Out, +Mat](delegate: scaladsl.SubFlow[Out, Mat, scaladsl.Flo
...
+    * from the previously calculated collection
+    *
+    * '''Backpressures when''' downstream backpressures or there are still remaining elements from the
+    * previously calculated collection
+    *
+    * '''Completes when''' upstream completes and all remaining elements has been emitted
+    *
+    * '''Cancels when''' downstream cancels
+    */
+  def statefulMapConcat[T](f: function.Creator[function.Function[Out, java.lang.Iterable[T]]]): SubFlow[In, T, Mat] =
```
patriknw: Is the intention to create `stateful` variants of other things also, such as `statefulMap` and `statefulFilter`, or is `statefulMapConcat` the universal `stateful` tool?

agolubev: as far as I can see ```statefulMapConcat``` is covering both ```statefulMap``` and ```statefulFilter```

patriknw: I agree, I was just curious

drewhk:
Needs a rebase

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSupervisionSpec.scala
```diff
@@ -303,7 +303,7 @@ class InterpreterSupervisionSpec extends AkkaSpec with GraphInterpreterSpecKit {
       lastEvents() should be(Set(OnNext(3)))
     }
 
-    "resume when MapConcat throws" in new OneBoundedSetup[Int](Seq(
+    /*"resume when MapConcat throws" in new OneBoundedSetup[Int](Seq(
```
drewhk: this needs to be fixed. The approach should be similar to the one @johanandren used in the case of batch: https://github.com/akka/akka/pull/19564

drewhk:
I am a bit unsure if this is what the original ticket is about. Let's wait on @rkuhn to explain what he meant there.

akka-stream/src/main/scala/akka/stream/javadsl/SubFlow.scala
rkuhn: Yes, this is the Swiss Army Knife of linear operators.

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -797,7 +797,36 @@ final class Source[+Out, +Mat](delegate: scaladsl.Source[Out, Mat]) extends Grap
...
+    *
+    * '''Completes when''' upstream completes and all remaining elements has been emitted
+    *
+    * '''Cancels when''' downstream cancels
+    */
+  def statefulMapConcat[T](f: function.Creator[function.Function[Out, java.lang.Iterable[T]]]): javadsl.Source[T, Mat] =
+    new Source(delegate.statefulMapConcat{ () ⇒ {
+      val fun = f.create()
+      elem ⇒ Util.immutableSeq(fun(elem))
+    }})
```
johanandren: As you are not allowed to mutate values outside of the closure with Java lambdas, would it maybe make more sense with a specific type to pass in here?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -1104,3 +1072,44 @@ private[stream] final class Reduce[T](f: (T, T) ⇒ T) extends SimpleLinearGraph
...
+ * INTERNAL API
+ */
+private[stream] final class StatefulMapConcat[In, Out](f: () ⇒ In ⇒ immutable.Iterable[Out]) extends GraphStage[FlowShape[In, Out]] {
+  val in = Inlet[In]("StatefulMapConcat.in")
+  val out = Outlet[Out]("StatefulMapConcat.out")
+  override val shape = FlowShape(in, out)
+  override def initialAttributes: Attributes = DefaultAttributes.statefulMapConcat
+
+  def createLogic(inheritedAttributes: Attributes) = new GraphStageLogic(shape) {
+    var currentIterator: Iterator[Out] = Iterator.empty
```
rkuhn: This reminds me: please put a `object EmptyIterator extends Iterator[Nothing]` into `akka.util.Collections` and use that here (as we as in EmptyImmutableSeq).

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -1104,3 +1072,44 @@ private[stream] final class Reduce[T](f: (T, T) ⇒ T) extends SimpleLinearGraph
...
+      override def onPush(): Unit = {
+        currentIterator = plainFun(grab(in)).iterator
+        pushPull()
+      }
+      override def onUpstreamFinish(): Unit = onFinish()
+    })
+    setHandler(out, new OutHandler {
+      override def onPull(): Unit = pushPull()
+      override def onDownstreamFinish(): Unit = onFinish()
+    })
```
rkuhn: please fold these two handlers into the logic

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -1104,3 +1072,44 @@ private[stream] final class Reduce[T](f: (T, T) ⇒ T) extends SimpleLinearGraph
...
+    setHandler(in, new InHandler {
+      override def onPush(): Unit = {
+        currentIterator = plainFun(grab(in)).iterator
+        pushPull()
+      }
+      override def onUpstreamFinish(): Unit = onFinish()
+    })
+    setHandler(out, new OutHandler {
+      override def onPull(): Unit = pushPull()
+      override def onDownstreamFinish(): Unit = onFinish()
```
rkuhn: this override is not right: when downstream cancels we need to shut down immediately

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -1104,3 +1072,44 @@ private[stream] final class Reduce[T](f: (T, T) ⇒ T) extends SimpleLinearGraph
...
+ */
+private[stream] final class StatefulMapConcat[In, Out](f: () ⇒ In ⇒ immutable.Iterable[Out]) extends GraphStage[FlowShape[In, Out]] {
+  val in = Inlet[In]("StatefulMapConcat.in")
+  val out = Outlet[Out]("StatefulMapConcat.out")
+  override val shape = FlowShape(in, out)
+  override def initialAttributes: Attributes = DefaultAttributes.statefulMapConcat
+
+  def createLogic(inheritedAttributes: Attributes) = new GraphStageLogic(shape) {
+    var currentIterator: Iterator[Out] = Iterator.empty
+    var finished = false
```
rkuhn: this can be replaced by inspecting `isClosed(in)`

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -1104,3 +1072,44 @@ private[stream] final class Reduce[T](f: (T, T) ⇒ T) extends SimpleLinearGraph
...
+        if (!hasNext && finished) completeStage()
+      } else if (!finished)
+        pull(in)
+      else completeStage()
+
+    def onFinish(): Unit = if (!hasNext) completeStage() else finished = true
+
+    setHandler(in, new InHandler {
+      override def onPush(): Unit = {
+        currentIterator = plainFun(grab(in)).iterator
```
rkuhn: This makes me thing: we only ever need an Iterator, so we might as well change the signature to `() => In => Iterator[Out]`. This would also work better for Java, I think.

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -319,7 +319,35 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
...
+    *
+    * '''Backpressures when''' downstream backpressures or there are still remaining elements from the
+    * previously calculated collection
+    *
+    * '''Completes when''' upstream completes and all remaining elements has been emitted
+    *
+    * '''Cancels when''' downstream cancels
+    */
+  def statefulMapConcat[T](f: function.Creator[function.Function[Out, java.lang.Iterable[T]]]):javadsl.Flow[In, T, Mat] =
+    new Flow(delegate.statefulMapConcat{ () ⇒ {
```
rkuhn: I think it is better to always delegate to the directly corresponding method so that we keep the implementation pieces in exactly one place each.

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -319,7 +319,35 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
    * '''Cancels when''' downstream cancels
    */
   def mapConcat[T](f: function.Function[Out, java.lang.Iterable[T]]): javadsl.Flow[In, T, Mat] =
-    new Flow(delegate.mapConcat { elem ⇒ Util.immutableSeq(f(elem)) })
+    new Flow(delegate.statefulMapConcat { () ⇒ elem ⇒ Util.immutableSeq(f(elem)) })
+
+  /**
+    * Transform each input element into an `Iterable` of output elements that is
+    * then flattened into the output stream. Transformation is stateful
+    * meaning it's taking into account previous messages (hysteresis).
```
rkuhn: The second sentence is a bit too indirect (i.e. where is the state?). What about:

> The transformation is meant to be stateful, which is enabled by creating the transformation function anew for every materialization—the returned function will typically close over mutable objects to store state between invocations. For the stateless variant see `mapConcat`.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowStatefulMapConcatSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+        x ⇒ prev match {
+          case Some(e) ⇒
+            prev = Some(x)
+            (1 to e) map (_ ⇒ x)
+          case None ⇒
+            prev = Some(x)
+            List.empty[Int]
+        }
+      })))
+    }
```
rkuhn: Yes, this is how it is supposed to be used. Please add Java tests as well to make sure that type inference and syntax work out. One issue there is that Java only allows to close over effectively final references, which means that people will have to put their state into mutable objects (like arrays of size 1—unless the state object itself is mutable already).

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
rkuhn: No, we’ll let the Java users complain to the Java gods ;-) (there’s not much we can do)

rkuhn:
OK TO TEST

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: > background: Iterator.empty totally uselessly creates a new instance every time

Seriously? Meh, good to know.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: as in the logic can extend both instead of creating new instances?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: +1 this override is not necessary

drewhk:
I have one issue with this API. The iterator interface allows the user encode state, but he is not given a chance to clean up if necessary, we just stop calling the Iterator.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
rkuhn: yes

rkuhn:
You are talking about stream termination (because in all other cases the iterator will be exhausted). I don’t see how and why we should complicate this method signature to add another hook—in any case it would not be a lambda anymore, which defeats the point. If statefulMapConcat is not enough then it is time to use a GraphStage.

drewhk:
Ok, I would already go for a GraphStage, but hey :)

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
agolubev: Not sure I understand. I created additional function here for converting java Iterator to scala one. But perhaps you are talking about mapConcat?

rkuhn: Oops, sorry, my comment was on the wrong method: I mean that mapConcat should delegate to Scsla’s mapConcat and not its statefulMapConcat.

agolubev:
Pushed all changes according to feedback. Rebased.
Added tests for  Supervision strategies.

rkuhn:
Thanks again @agolubev !

Issue #19547, opened 2016-01-20
stages-overview is incomplete


johanandren:
Missing at least the `unfold.*` methods, should cover all stages there is.

rkuhn:
plus `batch*` as well

PR #19641, opened 2016-01-28
=doc #19547 Better overview of stages

johanandren:
List of sources and sinks that yields a toc, short description of each rather than the tabular that was in the overview before.

### DIFF
--- a/akka-docs/rst/general/stream/stages-overview.rst
+++ b/akka-docs/rst/general/stream/stages-overview.rst
@@ -12,51 +12,58 @@ These built-in sources are available from `akka.stream.scaladsl.Source` and `akk
 
 fromIterator
 ^^^^^^^^^^^^
-Streams the values from an iterator, requesting the next value when there is demand. If the iterator
-performs blocking operations, make sure to run it on a separate dispatcher.
+Stream the values from an iterator, requesting the next value when there is demand.
+
+If the iterator perform blocking operations, make sure to run it on a separate dispatcher.
 
 *emits* when the next value returned from the iterator
+
 *completes* when the iterator reaches it's end
 
 single
 ^^^^^^
-Streams a single object
+Stream a single object
 
 *emits* the value once
+
 *completes* when the single value has been emitted
 
 repeat
 ^^^^^^
-Streams a single object repeatedly
+Stream a single object repeatedly
 
 *emits* the same value repeatedly when there is demand
+
 *completes* never
 
 tick
 ^^^^
-A periodical repeated stream of an arbitrary object. Delay of first tick is specified
+A periodical repetition of an arbitrary object. Delay of first tick is specified
 separately from interval of the following ticks.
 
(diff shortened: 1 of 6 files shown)


johanandren:
PR:ing for feedback about the different layout of the page.

akka-docs/rst/general/stream/stages-overview.rst
```diff
@@ -1,21 +1,373 @@
...
+
+tick
+^^^^
+A periodical repeated stream of an arbitrary object. Delay of first tick is specified
+separately from interval of the following ticks.
+
+*emits* periodically, if there is downstream backpressure ticks are skipped
+*completes* never
+
+fromFuture
```
rkuhn: fromFuture/fromCompletionStage

akka-docs/rst/general/stream/stages-overview.rst
```diff
@@ -1,21 +1,373 @@
...
+Streams the result of a function as long as it returns a ``Some`` or non-empty ``Optional``, the value inside the optional
+consists of a tuple (or ``Pair``) where the first value is a state passed back into the next call to the function allowing
+to pass a state. The first invocation of the provided fold function will receive the ``zero`` state.
+
+Can be used to implement many stateful sources without having to touch the more low level ``GraphStage`` API.
+
+*emits* when there is demand and the unfold function over the previous state returns non empty value
+*completes* when the unfold function returns an empty value
+
+unfoldAsync
```
rkuhn: oh, did I convert that to CompletionStage as well? perhaps it is still missing …

rkuhn:
makes sense to me :+1:

johanandren:
I think I will split up separate ones for Java and Scala to not have to mention parallell types everywhere.

rkuhn:
Yes, good point.

akka-docs/rst/general/stream/stages-overview.rst
```diff
@@ -1,21 +1,373 @@
...
+
+*emits* the value once
+*completes* when the single value has been emitted
+
+repeat
+^^^^^^
+Streams a single object repeatedly
+
+*emits* the same value repeatedly when there is demand
+*completes* never
```
drewhk: these do not render vertically but treated as the same paragraph. Or does Github render differently than the Sphinx output?

rkuhn: no, you are right

drewhk:
I am OK with the layout, and the content is nice, too. LGTM after fixing formatting.

akka-docs/rst/general/stream/stages-overview.rst
johanandren: Ah, good point, the intention was to have them on separate lines. (I haven't looked much at the format of the output yet, will cover all stages first and then look at that)

drewhk:
refs #19547

PR #19644, opened 2016-01-29
#19637 fix BC issue in AsyncWriteJournal

### LINKED ISSUE #19637: Fix BC issue in AsyncWriteJournal
Manifests in:
```
Caused by: java.lang.AbstractMethodError: akka.contrib.persistence.mongodb.MongoJournal.akka$persistence$journal$AsyncWriteJournal$_setter_$akka$persistence$journal$AsyncWriteJournal$$replayDebugEnabled_$eq(Z)V
	at akka.persistence.journal.AsyncWriteJournal$class.$init$(AsyncWriteJournal.scala:50)
	at akka.contrib.persistence.mongodb.MongoJournal.<init>(MongoJournal.scala:17)
	... 16 more
```
when current persistence plugins are being used with 2.4.2-RC1.

BC issue was introduced in https://github.com/akka/akka/commit/0af98e7bee1312d10979dd2a8ac81a77502c1f43

rkuhn:
fixes #19637

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
@@ -47,7 +47,6 @@ trait AsyncWriteJournal extends Actor with WriteJournalBase with AsyncRecovery {
   private def isReplayFilterEnabled: Boolean = replayFilterMode != ReplayFilter.Disabled
   private val replayFilterWindowSize: Int = config.getInt("replay-filter.window-size")
   private val replayFilterMaxOldWriters: Int = config.getInt("replay-filter.max-old-writers")
-  private val replayDebugEnabled: Boolean = config.getBoolean("replay-filter.debug")
 
   private val resequencer = context.actorOf(Props[Resequencer]())
   private var resequencerCounter = 1L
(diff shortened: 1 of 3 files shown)


akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
```diff
@@ -47,7 +47,8 @@ trait AsyncWriteJournal extends Actor with WriteJournalBase with AsyncRecovery {
   private def isReplayFilterEnabled: Boolean = replayFilterMode != ReplayFilter.Disabled
   private val replayFilterWindowSize: Int = config.getInt("replay-filter.window-size")
   private val replayFilterMaxOldWriters: Int = config.getInt("replay-filter.max-old-writers")
-  private val replayDebugEnabled: Boolean = config.getBoolean("replay-filter.debug")
+  // cannot be a val due to binary compatibility
+  private def replayDebugEnabled: Boolean = config.getBoolean("replay-filter.debug")
```
patriknw: would it be possible to make it a `val` inside the `receiveWriteJournal`?
```
final val receiveWriteJournal: Actor.Receive = {
  val replayDebugEnabled: Boolean = config.getBoolean("replay-filter.debug")

  {...
```

rkuhn: Yes, that would be possible, I think. I didn’t consider accessing the config to be expensive, but now some memories are coming back that it actually was beneficial to cache these lookups. Will change.

akka-persistence/src/main/scala/akka/persistence/journal/ReplayFilter.scala
```diff
@@ -30,6 +30,13 @@ private[akka] object ReplayFilter {
     Props(new ReplayFilter(persistentActor, mode, windowSize, maxOldWriters, debugEnabled))
   }
 
+  // for binary compatibility
+  def props(
+    persistentActor: ActorRef,
+    mode: Mode,
+    windowSize: Int,
+    maxOldWriters: Int): Props = props(persistentActor, mode, windowSize, maxOldWriters, debugEnabled = false)
```
patriknw: this is INTERNAL API so this should not be needed (we must have some freedom)

akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
patriknw: thanks

akka-persistence/src/main/scala/akka/persistence/journal/ReplayFilter.scala
rkuhn: Yes, but on the other hand I do not see much value in breaking compatibility gratuitously.

akka-persistence/src/main/scala/akka/persistence/journal/AsyncWriteJournal.scala
```diff
@@ -55,109 +54,114 @@ trait AsyncWriteJournal extends Actor with WriteJournalBase with AsyncRecovery {
   final def receive = receiveWriteJournal.orElse[Any, Unit](receivePluginInternal)
 
   final val receiveWriteJournal: Actor.Receive = {
-    case WriteMessages(messages, persistentActor, actorInstanceId) ⇒
-      val cctr = resequencerCounter
-      resequencerCounter += messages.foldLeft(0)((acc, m) ⇒ acc + m.size) + 1
+    // cannot be a val in the trait due to binary compatibility
+    val replayDebugEnabled: Boolean = config.getBoolean("replay-filter.debug")
```
rkuhn: This is the only change, the rest is just indentation (check [here](https://github.com/akka/akka/pull/19644/files?w=1)).

akka-persistence/src/main/scala/akka/persistence/journal/ReplayFilter.scala
patriknw: maintaining superfluous code is a burden

rkuhn: Yes, of course. All I’m saying is that the cost of breaking compatibility without thinking twice is also not zero—for internal api the cost might be small, which means that a small amount of burden might be justified.

patriknw: yes, but I'm thinking, now, anyway, not a big deal

patriknw:
LGTM

akka-persistence/src/main/scala/akka/persistence/journal/ReplayFilter.scala
```diff
@@ -46,6 +53,10 @@ private[akka] class ReplayFilter(persistentActor: ActorRef, mode: ReplayFilter.M
   import JournalProtocol._
   import ReplayFilter.{ Warn, Fail, RepairByDiscardOld, Disabled }
 
+  // for binary compatibility
+  def this(persistentActor: ActorRef, mode: ReplayFilter.Mode,
+           windowSize: Int, maxOldWriters: Int) = this(persistentActor, mode, windowSize, maxOldWriters, debugEnabled = false)
+
```
2m: Do we needs this? Newly added `props` method calls original `props` and not constructor.

rkuhn: As Patrik argued we do not actually need this, I just removed all exceptions while I was at it because I think that there is also some value in keeping that list short. This is a matter that can be debated, of course.

2m: I still do not understand. Patrik argued about new `props` method. This is new constructor which is not used anywhere. Why do we need it?

2m: I see now. It might be called from an older version of akka.

2m:
LGTM

akka-persistence/src/main/scala/akka/persistence/journal/ReplayFilter.scala
patriknw: None of the changes in this file are really needed since it's internal api, but Roland preferred to make them binary backwards compatible anyway and then all of it should be changed.

Issue #19469, opened 2016-01-15
Akka streams TCP support silently handles failures


jroper:
When Akka streams' TCP stage handler gets an upstream failure, it sends an abort message to the stream, but drops the exception, leaving the user in the dark about why the connection was reset.  This is particularly frustrating if the bug that is causing the failure is not in the end users code, for example, if the bug is in Akka HTTP, eg as in #19467, because there's nothing the user can do, short of diving into Akka streams internals with a debugger, to work out what is going wrong.  Akka streams should log any failure exceptions when it aborts the connection.

jroper:
Problem code is here:

https://github.com/akka/akka/blob/stream-http-2.0/akka-stream/src/main/scala/akka/stream/impl/io/TcpStages.scala#L271

drewhk:
Hm, the only reason why it does not call a failStage right there because it waits first on the `Aborted` response. But then it already loses the original exception. The questions I have here:
 - is it enough to fail the stage (i.e. it will fail its output port if it is open)
 - if not, then which loglevel should it be logged?

So if we know what we want here it is an easy fix.

rkuhn:
Different question: every failure will be logged at ERROR level where it originates, is that not enough?

jroper:
That's not the current behaviour is it?  If you're referring to changing things such that when a GraphStage does the wrong thing, eg pull from a closed stream, then it should be logged then and there, yes, I think that's fine, since that will always be a bug, and being noisy about bugs is good.  Just make sure the stack trace is logged too.

drewhk:
I marked this for possible backport, since this can help debug issues reported on the 2.0.x line. We can remove the tag if we decide to leave this out.

PR #19647, opened 2016-01-29
=str #19469 Log all graph stage exceptions

johanandren:
Refs #19469 

Adds logging to the interpreter, does not touch the TcpLayer

### DIFF
--- a/akka-stream-tests/src/test/resources/reference.conf
+++ b/akka-stream-tests/src/test/resources/reference.conf
@@ -5,6 +5,7 @@ akka {
     serialize-messages = on
     default-dispatcher.throughput = 1 // Amplify the effects of fuzzing
   }
+  akka.actor.warn-about-java-serializer-usage = false
 
   stream.materializer.debug.fuzzing-mode = on
 }
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
@@ -467,7 +467,9 @@ private[stream] final class GraphInterpreter(
         logic.beforePreStart()
         logic.preStart()
       } catch {
-        case NonFatal(e) ⇒ logic.failStage(e)
+        case NonFatal(e) ⇒
+          log.error(e, "Error during preStart in [{}]", assembly.stages(logic.stageId))
+          logic.failStage(e)
       }
       afterStageHasRun(logic)
       i += 1
@@ -534,7 +536,12 @@ private[stream] final class GraphInterpreter(
         catch {
           case NonFatal(e) ⇒
             if (activeStage == null) throw e
-            else activeStage.failStage(e, isInternal = true)
+            else {
+              val stage = assembly.stages(activeStage.stageId)
+
+              log.error(e, "Error in stage [{}]: {}", stage, e.getMessage)
+              activeStage.failStage(e)
+            }
         }
         afterStageHasRun(activeStage)
         eventsRemaining -= 1
@@ -669,7 +676,7 @@ private[stream] final class GraphInterpreter(
       logic.afterPostStop()
     } catch {
       case NonFatal(e) ⇒
-        log.error(e, s"Error during postStop in [${assembly.stages(logic.stageId)}]")
+        log.error(e, s"Error during postStop in [{}]: {}", assembly.stages(logic.stageId), e.getMessage)
     }
   }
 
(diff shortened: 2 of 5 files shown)


johanandren:
Since the TcpLayer will get it's possibly loggable exceptions from the stream, they will already be logged so I can't see a reason to also add logging there.

akka-stream-tests/src/test/resources/reference.conf
```diff
@@ -5,6 +5,7 @@ akka {
     serialize-messages = on
     default-dispatcher.throughput = 1 // Amplify the effects of fuzzing
   }
+  akka.actor.warn-about-java-serializer-usage = false
```
drewhk: where was this relevant?

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
```diff
@@ -550,7 +550,9 @@ class InterpreterSpec extends AkkaSpec with GraphInterpreterSpecKit {
       new InvalidAbsorbTermination)) {
       lastEvents() should be(Set.empty)
 
-      EventFilter[UnsupportedOperationException]("It is not allowed to call absorbTermination() from onDownstreamFinish.", occurrences = 1).intercept {
+      // TODO we get two log entries after more verbose error logging was added to the interpreter
+      // couldn't figure out how to not do that without breaking other expected error logging
+      EventFilter[UnsupportedOperationException]("It is not allowed to call absorbTermination() from onDownstreamFinish.", occurrences = 2).intercept {
```
drewhk: what other expected error logging breaks?

akka-stream-tests/src/test/resources/reference.conf
johanandren: Just really noisy test outputs, is disabled in other akka module tests.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
johanandren: I think it was the `GraphInterpreterFailureModesSpec` and possibly a few more

drewhk: there is only one in FailureModesSpec though. I don't like that double logging here, this means something fishy. How are exactly the two log statements look like?

johanandren: It's this one in addition to the logging I added: https://github.com/akka/akka/blob/wip-19469-better-stream-error-logging-johanandren/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala#L715 but if I remove this then tests failed because not all code paths there goes through the error logging I added here https://github.com/akka/akka/blob/wip-19469-better-stream-error-logging-johanandren/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala#L540

drewhk: The issue is error logging in `postStop()`. My original logging addition ensured that these are logged since they were impossible to propagate through the onError mechanism (we didn't log errors before, since the idea was that onError carries it anyway, but after completion there is no more onError). I am not sure this is actually needed anymore, since we always log now. You can probably remove that log statement and have your "catch-all" handle those cases instead.

What might be valuable though is to change the string to indicate if an error happened after the stage was logged and hence error was not propagated anywhere else.

akka-stream-tests/src/test/resources/reference.conf
patriknw: I guess that is because of String and int messages. Perhaps we should add `java.lang` to our ignored prefixes?

patriknw:
LGTM

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -706,8 +713,6 @@ private[stream] final class GraphInterpreter(
       portStates(connection) = currentState | (OutClosed | InFailed)
       connectionSlots(connection) = Failed(ex, connectionSlots(connection))
       if ((currentState & (Pulling | Pushing)) == 0) enqueue(connection)
-    } else if (isInternal) {
```
drewhk: I guess in this case this means that `isInternal` can be completely eliminated.

rkuhn:
LGTM

drewhk:
LGTM

2m:
LGTM

Issue #19653, opened 2016-01-31
Sink.actorRefWithAck gets stuck when buffer runs full


rkuhn:
~~~scala
scala> class Acker extends Actor { def receive = {case x => println(x); sender() ! "ack" } }
scala> Source.repeat(1).runWith(Sink.actorRefWithAck(as.actorOf(Props[Acker]), "init", "ack", "done", x => x))
init
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
~~~
(from @ALPHA-60 on gitter)

drewhk:
Needs a backport right?

PR #19654, opened 2016-01-31
fix ActorRefWithAck when buffer has run full #19653

rkuhn:


rkuhn:
fixes #19653

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkStage.scala
```diff
@@ -52,10 +52,9 @@ private[akka] class ActorRefBackpressureSinkStage[In](ref: ActorRef, onInitMessa
       }
 
       private def sendData(): Unit = {
-        if (!buffer.isEmpty) {
-          ref ! buffer.poll()
-          acknowledgementReceived = false
-        }
+        if (buffer.size() == maxBuffer) tryPull(in)
```
johanandren: We only pull if the buffer is full?

rkuhn: Yes, because on for that last element did we not yet send a pull in response.

johanandren:
LGTM

akka-stream/src/main/scala/akka/stream/impl/ActorRefBackpressureSinkStage.scala
drewhk: Yes, it frees up a slot, and hence it *unstucks* the upstream (this was a `pushAndPull` in the old interpreter). What got confused me here for a second though is that `.poll()` actually removes the element.

drewhk:
LGTM

PR #19666, opened 2016-02-02
=str #19664 Docs no longer claim that framer delimiter is included

### LINKED ISSUE #19664: Framing.delimiter does not include delimiter
In the docs it says:

> The decoded frames will include the separator sequence. If this is not desired, this Flow can be  augmented with a simple ''map'' operation that removes this separator.

But with this test case you can see that it actually strips out the delimiter:

```Scala
val f = Source(List(ByteString("a\n")))
  .via(Framing.delimiter(ByteString("\n"), 256))
  .runFold(ByteString())(_ ++ _)

Await.result(f, 3.seconds) should be(ByteString("a\n"))
```

johanandren:
Fixes #19664


drewhk:
LGTM

Issue #19622, opened 2016-01-27
ClusterShardingStats node names vary in unpredictable ways


jamesmulcahy:
Lets say I have three nodes, A, B and C.  If I start A, then have B and C join, I get ClusterShardingStats responses which represent three regions:

akka://ClusterName  (A)
akka.tcp://ClusterName@127.0.0.1:2552 (B)
akka.tcp://ClusterName@127.0.0.1:2553 (C)

But now if I fail A, then restart it, the responses end up looking like

akka.tcp://ClusterName@127.0.0.1:2551 (A)
akka.tcp://ClusterName@127.0.0.1:2552 (B)
akka.tcp://ClusterName@127.0.0.1:2553 (C)

It looks to me like the implementation is going to some effort to do the re-write in the original case, but that doesn't persist for all future cases (perhaps when the co-ordinator is on a separate node to the one that is initiating the GetClusterShardingStats) request.

So my questions are:  Is this intentional, and if not, is there anything tricky about making this "predictable" - i.e. always use the fully specified name (as per the post-failover) requests.

As it stands, it makes consuming this data a bit of a pain.

johanandren:
Sounds like a bug to me, the code clearly tries to use the self-address but there must be something wrong with that logic.

johanandren:
Any chance you could try to solve and provide a pull request for this? We are currently very busy with streams and http for 2.4.2 so I can't foresee we have time to fix it until a later Akka version if not.

patriknw:
I agree that it sounds better to fill in the full address. Here is how to do that:
```
def fillLocal(address: Address): Address = {
    // local address might be used if grabbed from actorRef.path.address
    if (address.hasLocalScope && address.system == selfAddress.system) selfAddress
    else address
  }
```

jamesmulcahy:
I was just about to push something myself, but you beat me to it :) Thanks @talp-traiana !

PR #19668, opened 2016-02-02
=clu #19622 Use full address in ClusterShardingStats

ghost:


patriknw:
OK TO TEST

patriknw:
LGTM

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardCoordinator.scala
```diff
@@ -555,7 +555,7 @@ abstract class ShardCoordinator(typeName: String, settings: ClusterShardingSetti
         ShardRegion.ClusterShardingStats(allRegionStats.map {
           case (region, stats) ⇒
             val address: Address =
-              if (region == self) Cluster(context.system).selfAddress
+              if (address.hasLocalScope && address.system == cluster.selfAddress.system) cluster.selfAddress
```
johanandren: address refers to the address you are assigning here, so this does not compile

johanandren:
Refs #19622

talp-traiana:
Ohhh, riiight. I compiled akka-cluster instead of akka-cluster-sharding... Sorry.

johanandren:
Would be a good idea (and pretty easy) to cover this in the tests also, in `ClusterShardingGetStatsSpec` you should be able to just add assertions on the address in the stats messages.

Note that those are multi-jvm tests and you would run it with:
`akka-cluster-sharding/multi-jvm:testOnly akka.cluster.sharding.ClusterShardingGetStatsSpec`

talp-traiana:
@johanandren OK, how's this look?

johanandren:
LGTM!

jamesmulcahy:
LGTM. I've tested this and it fixes the problem I originally encountered.  Thanks @talp-traiana

PR #19669, opened 2016-02-02
Don't leak FunctionRef instances

drewhk:
They were not removed anywhere from the parent in streams code (they are removed in Akka typed).

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
+++ b/akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
@@ -132,14 +132,14 @@ object GraphStageLogic {
                          initialReceive: StageActorRef.Receive) {
 
     private val callback = getAsyncCallback(internalReceive)
+    private def cell = materializer.supervisor match {
+      case ref: LocalActorRef                        ⇒ ref.underlying
+      case ref: RepointableActorRef if ref.isStarted ⇒ ref.underlying.asInstanceOf[ActorCell]
+      case unknown ⇒
+        throw new IllegalStateException(s"Stream supervisor must be a local actor, was [${unknown.getClass.getName}]")
+    }
 
     private val functionRef: FunctionRef = {
-      val cell = materializer.supervisor match {
-        case ref: LocalActorRef                        ⇒ ref.underlying
-        case ref: RepointableActorRef if ref.isStarted ⇒ ref.underlying.asInstanceOf[ActorCell]
-        case unknown ⇒
-          throw new IllegalStateException(s"Stream supervisor must be a local actor, was [${unknown.getClass.getName}]")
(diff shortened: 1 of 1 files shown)


drewhk:
I am not sure if the fix is 100% correct. It eliminates the leak but I don't know if there are some races there (children containers and DeathWatchNotification messages), so @rkuhn you should take a look.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -178,7 +178,11 @@ object GraphStageLogic {
       behaviour = receive
     }
 
-    def stop(): Unit = functionRef.stop()
+    def stop(): Unit = {
+      functionRef.stop()
+      // Can this be here?
+      cell.removeFunctionRef(functionRef)
```
drewhk: and null out the cell.. oh

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -132,14 +132,14 @@ object GraphStageLogic {
                          initialReceive: StageActorRef.Receive) {
 
     private val callback = getAsyncCallback(internalReceive)
+    private val cell = materializer.supervisor match {
```
rkuhn: this should probably be a `def`

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
```diff
@@ -178,7 +178,11 @@ object GraphStageLogic {
       behaviour = receive
     }
 
-    def stop(): Unit = functionRef.stop()
+    def stop(): Unit = {
+      functionRef.stop()
```
rkuhn: this now needs to go, otherwise we call it twice

rkuhn:
LGTM apart from the comments—I don’t smell any race conditions here.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: ah, I see, this it calls it.

akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala
drewhk: right, makes sense

drewhk:
updated

drewhk:
I was not sure about the order between stop and remove, but since the latter calls the former, I expect the order is taken properly care of now.

rkuhn:
Should all be good now.

drewhk:
refs: https://github.com/akka/akka/issues/19670

Issue #19675, opened 2016-02-02
Sink.queue gets stuck when buffer runs full


ALPHA-60:
This issue might be similar to #19653. Below, printing stops after 15 lines:

```
scala> val q = Source(1 to 30).runWith(Sink.queue());
scala> def f(): Unit = q.pull().onComplete{case Success(x) => {println(x); f()}; case e => println(e)}
scala> f()
Some(1)
Some(2)
Some(3)
Some(4)
Some(5)
Some(6)
Some(7)
Some(8)
Some(9)
Some(10)
Some(11)
Some(12)
Some(13)
Some(14)
Some(15)
```

wlk:
related PR: #19676 was merged, this probably should be closed, right?

ktoso:
Correct, thanks for noticing

PR #19676, opened 2016-02-02
fix Sink.queue when buffer has run full #19675

ALPHA-60:
The solution is based on #19654.

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/QueueSinkSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/QueueSinkSpec.scala
@@ -8,14 +8,16 @@ import akka.pattern.pipe
 import akka.stream.{ OverflowStrategy, ActorMaterializer }
 import akka.stream.testkit.Utils._
 import akka.stream.testkit.{ AkkaSpec, _ }
+import org.scalatest.concurrent.ScalaFutures
 
 import scala.concurrent.Await
 import scala.concurrent.duration._
 import scala.util.control.NoStackTrace
 
-class QueueSinkSpec extends AkkaSpec {
+class QueueSinkSpec extends AkkaSpec with ScalaFutures {
   implicit val ec = system.dispatcher
   implicit val materializer = ActorMaterializer()
+  implicit val patience = PatienceConfig(2.second)
 
   val ex = new RuntimeException("ex") with NoStackTrace
 
@@ -110,6 +112,21 @@ class QueueSinkSpec extends AkkaSpec {
       Await.result(queue.pull(), noMsgTimeout) should be(None)
 
       queue.pull().onFailure { case e ⇒ e.isInstanceOf[IllegalStateException] should ===(true) }
+    }
+
+    "keep on sending even after the buffer has been full" in assertAllStagesStopped {
+      val (probe, queue) = Source(1 to 20)
+        .alsoToMat(Flow[Int].take(15).watchTermination()(Keep.right).to(Sink.ignore))(Keep.right)
+        .toMat(Sink.queue())(Keep.both)
+        .run()
+      probe.futureValue should ===(akka.Done)
+      for (i ← 1 to 20) {
+        queue.pull() pipeTo testActor
+        expectMsg(Some(i))
+      }
+      queue.pull() pipeTo testActor
+      expectMsg(None)
+
     }
 
   }
--- a/akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
@@ -277,7 +277,10 @@ private[akka] class QueueSink[T]() extends GraphStageWithMaterializedValue[SinkS
             promise.failure(new IllegalStateException("You have to wait for previous future to be resolved to send another request"))
           case None ⇒
             if (buffer.isEmpty) currentRequest = Some(promise)
-            else sendDownstream(promise)
+            else {
+              if (buffer.used == maxBuffer - 1) tryPull(in)
+              sendDownstream(promise)
+            }
         })
 
       def sendDownstream(promise: Requested[T]): Unit = {


typesafehub-validator:
Hi @ALPHA-60,

ALPHA-60:
fixes #19675

akka-stream-tests/src/test/scala/akka/stream/scaladsl/QueueSinkSpec.scala
```diff
@@ -112,5 +114,26 @@ class QueueSinkSpec extends AkkaSpec {
...
     }
 
+    "keep on sending even after the buffer has been full" in assertAllStagesStopped {
+      val (probe, queue) = Source(1 to 20)
+        .alsoToMat(Flow[Int].take(16).watchTermination()(Keep.right).to(Sink.ignore))(Keep.right)
+        .toMat(Sink.queue())(Keep.both)
+        .run()
+
+      // wait for the buffer to fill up before pulling
+      akka.pattern.after(0.5 seconds, system.scheduler)(queue.pull()) pipeTo testActor
```
ALPHA-60: I couldn't think of a better way to detect a regression. The buffer might still not be full after 0.5 seconds.

johanandren: What about something like @rkuhn did in the test coverage for the other bug? #19654

ALPHA-60: This test case was actually based on @rkuhn's test case coverage that you mentioned.

johanandren: No, maybe that didn't help much. Having a happy path which always requires a wait for 0.5 seconds isn't very desirable. We would want it to pass as soon as the requirement has been fulfilled if possible, and then in the worst case it can wait all that time.

Something like this would therefore be better:
```Scala
within(0.5.seconds) {
  queue.pull() pipeTo testActor
  expectMsg(Some(1))
}
```

rkuhn: @johanandren this would pull immediately, which is not what we want: we need to wait until the buffer has filled up. What is missing here is to wait for the `probe` first, and the proceed.

johanandren: Ah, got it, I thought about it backwards (and additionally missed an `awaitAssert` from that thought of train). Blocking on the probe from source to complete sounds like the thing to do.

itekako: I think I understand, i will try to fix it.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/QueueSinkSpec.scala
```diff
@@ -112,5 +114,20 @@ class QueueSinkSpec extends AkkaSpec {
       queue.pull().onFailure { case e ⇒ e.isInstanceOf[IllegalStateException] should ===(true) }
     }
 
+    "keep on sending even after the buffer has been full" in assertAllStagesStopped {
+      val (probe, queue) = Source(1 to 20)
+        .alsoToMat(Flow[Int].take(15).watchTermination()(Keep.right).to(Sink.ignore))(Keep.right)
+        .toMat(Sink.queue())(Keep.both)
+        .run()
+      probe.futureValue should ===(akka.Done)
```
ALPHA-60: I misunderstood the way probe was used in #19654, and, based on @rkuhn's suggestion from conversation with @johanandren, I believe this is the correct way to test buffer filling up.

Buffer can have at most 15 upstream elements (not stream close/failure indicators) in it, based on the condition in `onPull` method.

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -277,7 +277,10 @@ private[akka] class QueueSink[T]() extends GraphStageWithMaterializedValue[SinkS
             promise.failure(new IllegalStateException("You have to wait for previous future to be resolved to send another request"))
           case None ⇒
             if (buffer.isEmpty) currentRequest = Some(promise)
-            else sendDownstream(promise)
+            else {
+              if (buffer.used == maxBuffer - 1) tryPull(in)
```
ALPHA-60: I hope I correctly understood the guard here. The buffer's `used` property might actually be equal to `maxBuffer`, but in this case, the last enqueued buffer element is either a `Success(None)` or `Failure(ex)` and the `tryPull` is not needed. Enqueuing of these special elements happens in the `onUpstreamFinish` or `onUpstreamFailure` callbacks, so the guard from `onPush` (`buffer.used < maxBuffer  - 1`) entails this guard here. (I found this condition somewhat confusing, until I understood that stream end markers are kept in the buffer together with regular elements sent by upstream, and that they are not enqueued in the `onPull` method).  

What I still don't understand is why we allocate a `maxBuffer + 1` sized buffer.  Which code path would lead to buffer having 17 elements in it?

rkuhn: Indeed, the allocated buffer is one too big, wasting one element but also using modulo instead of masking for the internal management (due to not being a power of two in size).

rkuhn:
OK TO TEST

ALPHA-60:
Hm, the failing test is "akka.stream.scaladsl.One2OneBidiFlowSpec", the "must trigger an `OutputTruncationException` if the wrapped stream cancels early" case. It does not seem related to me at the moment, since I didn't see Sink.queue being used in the spec or in the class tested. I tried merging with master locally, and the test passes. Could it be transient? Is it possible to trigger the build again?

I see these throught the log:
```
[ERROR] [02/11/2016 15:31:19.841] [ClientCancellationSpec-akka.actor.default-dispatcher-2] [akka://ClientCancellationSpec/user/StreamSupervisor-8/flow-2-0-unknown-operation] Error in stage [One2OneBidi]: null (akka.stream.scaladsl.One2OneBidiFlow$OutputTruncationException$)
```

rkuhn:
Yes, the failure is unrelated.

ALPHA-60:
Thanks!

drewhk:
that failure is unrelated. I will fix it tomorrow. It is a race in the test that now in hindsight is rather obvious :)

Issue #19679, opened 2016-02-02
akka.cluster.MemberStatus ought to be a sealed type


retronym:
I recently improved exhaustivity warnings in the Scala 2.12 compiler. Running this against Akka sources, we get this warning:

```
[akka] [warn] /home/jenkins/workspace/scala-2.12.x-integrate-community-build/target-0.9.5/project-builds/akka-622a540c67e9bc2514e3685a0f16c7ad317992b0/akka-cluster/src/main/scala/akka/cluster/Member.scala:142: match may not be exhaustive.
[akka] [warn] It would fail on the following inputs: (Up, _), (_, Up), (_, _)
[akka] [warn]   def highestPriorityOf(m1: Member, m2: Member): Member = (m1.status, m2.status) match {
[akka] [warn] 
```

This will go away if we seal `MemberStatus`.

johanandren:
Sounds reasonable to me. Breaks source compatibility but I cannot see how anyone would have provided their own subtype of `MemberStatus`, must be an oversight on our part.

2m:
Fixed in #19680.

PR #19680, opened 2016-02-02
!clu #19679 make MemberStatus a sealed class

retronym:
My assumption is that it the absence of the sealed modifier
was an oversight. Marking it as sealed will avoid exhausitity
warnings from upcoming Scala compiler version in `highestPriorityOf`.

### DIFF
--- a/akka-cluster/src/main/scala/akka/cluster/Member.scala
+++ b/akka-cluster/src/main/scala/akka/cluster/Member.scala
@@ -178,9 +178,9 @@ object Member {
 /**
  * Defines the current status of a cluster member node
  *
- * Can be one of: Joining, Up, Leaving, Exiting and Down.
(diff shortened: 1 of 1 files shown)


retronym:
For ticket #19679

johanandren:
OK TO TEST

drewhk:
LGTM, although it strictly speaking breaks source compatibility. I guess it is not an issue here, but let's discuss.

patriknw:
It was originally a `sealed trait`. Perhaps the sealed part was missed when changing it to an abstract class. Anyway, I see no reason why it shouldn't be sealed. Looks like mima approves, and source incompatibility should not be a problem, since if someone has extended it they are doing something weird.

johanandren:
LGTM then

2m:
LGTM

PR #19684, opened 2016-02-03
#19662 Reuse connection pool settings if none specified

### LINKED ISSUE #19662: Http.singleRequest creates a new pool each time if no settings are specified.
When you don't specify the `ConnectionPoolSettings`, `Http.singleRequest` will create a new instance of them (using the actor system config). These _are not equal_, so will get their own entries in the hostPoolCache - i.e. the pools aren't reused.

The problem seems to be in `ParserSettings` - it has 2 function parameters (`customMethods` & `customParameters`) which will compare differently for different instances, even if they use the default `_ => None` function. Perhaps they should be maps instead?

As an workaround you can store the default settings yourself, e.g.
```
val first = http.singleRequest(req)
val second = http.singleRequest(req) // Uses a different pool

val settings = ConnectionPoolSettings.default
val third = http.singleRequest(req, settings = settings)
val fourth = http.singleRequest(req, settings = settings) // Uses the connection pool from the third
```
The description of the method simply says
>Fires a single HttpRequest across the (cached) host connection pool for the request's effective URI

so I don't think this is expected.

2m:
Fixes #19662


johanandren:
LGTM!

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -43,6 +43,8 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
 
   override val sslConfig = AkkaSSLConfig(system)
 
+  private[this] val defaultConnectionPoolSettings = ConnectionPoolSettings(system)
```
rkuhn: While this is a good change by itself, we also need to fix equality for ConnectionPoolSettings.

ktoso: As @rkuhn said, we'll have to make those behave like their previous case class counterparts (broken since moving to the "easier to binary compat evolve" PR).

drewhk: Then it is important to add test cases.

2m:
Updated PR. The problem with equality was not missing case class inheritance, but `ParserSettings` which has two function members. These were created everytime new `ParserSettings` was instantiated.

2m:
PLS TEST

akka-http-core/src/test/scala/akka/http/scaladsl/settings/SettingsEqualitySpec.scala
```diff
@@ -0,0 +1,68 @@
...
+    "hold for ServerSettings" in {
+      val s1 = ServerSettings(config)
+      val s2 = ServerSettings(config)
+
+      s1 shouldBe s2
+      s1.toString should startWith("ServerSettings(")
+    }
+  }
+
+}
```
johanandren: :+1:

2m:
PLS BUILD

akka-http-core/src/main/scala/akka/http/impl/settings/ClientConnectionSettingsImpl.scala
```diff
@@ -29,6 +29,8 @@ private[akka] final case class ClientConnectionSettingsImpl(
 
   require(connectingTimeout >= Duration.Zero, "connectingTimeout must be >= 0")
   require(requestHeaderSizeHint > 0, "request-size-hint must be > 0")
+
+  override def productPrefix = "ClientConnectionSettings"
```
ktoso: :-)

akka-http-core/src/test/scala/akka/http/scaladsl/ClientSpec.scala
```diff
@@ -0,0 +1,57 @@
...
+      val resp = Await.result(respFuture, 3.seconds)
+      resp.status shouldBe StatusCodes.OK
+
+      (Http() invokePrivate hostPoolCache()).size shouldBe 1
+
+      val respFuture2 = Http().singleRequest(HttpRequest(POST, s"http://$hostname:$port/"))
+      val resp2 = Await.result(respFuture, 3.seconds)
+      resp2.status shouldBe StatusCodes.OK
+
+      (Http() invokePrivate hostPoolCache()).size shouldBe 1
```
ktoso: I'm not a huge fan of invokePrivate, would rather make it `private[http]`. Anyone have strong opinions on it?

johanandren: I concur.

Issue #19674, opened 2016-02-02
Make sure all HTTP headers are marked request/response correctly


johanandren:
We got another report of a header that is not marked as `ResponseHeader` when it should `Accept-Ranges`. 

I suggest we go through each header and check that it is marked correctly according to spec, shouldn't be too hard as each header has a comment with a link to the spec section for it.

PR #19685, opened 2016-02-03
!htt #19674 Audit HTTP headers and mark correctly as response or request

johanandren:
Refs #19674 

Went through each http header and read the RFC entry on each to figure out if they were correctly marked as request, response or both. I feel that it is possible that some were incorrectly marked as both on purpose, so some eyes from @jrudolph or @sirthias would be great.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
@@ -171,7 +171,7 @@ object `Accept-Ranges` extends ModeledCompanion[`Accept-Ranges`] {
   implicit val rangeUnitsRenderer = Renderer.defaultSeqRenderer[RangeUnit] // cache
 }
 final case class `Accept-Ranges`(rangeUnits: immutable.Seq[RangeUnit]) extends jm.headers.AcceptRanges
-  with RequestHeader {
+  with ResponseHeader {
   import `Accept-Ranges`.rangeUnitsRenderer
   def renderValue[R <: Rendering](r: R): r.type = if (rangeUnits.isEmpty) r ~~ "none" else r ~~ rangeUnits
   protected def companion = `Accept-Ranges`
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -559,7 +559,7 @@ object Link extends ModeledCompanion[Link] {
   def apply(values: LinkValue*): Link = apply(immutable.Seq(values: _*))
   implicit val valuesRenderer = Renderer.defaultSeqRenderer[LinkValue] // cache
 }
-final case class Link(values: immutable.Seq[LinkValue]) extends jm.headers.Link with RequestResponseHeader {
+final case class Link(values: immutable.Seq[LinkValue]) extends jm.headers.Link with ResponseHeader {
```
johanandren: Especially this one I found hard to see from the RFC if it actually is ok as request header, but couldn't find any references online to anyone using it in requests so I changed it.

johanandren:
Thanks @sirthias exactly the kind of feedback I was looking for. Will update according to your comments.

drewhk:
LGTM

viktorklang:
Perhaps add test which verify this?

akka-http-core/src/test/scala/akka/http/scaladsl/model/headers/HeaderSpec.scala
```diff
@@ -5,15 +5,15 @@
 package akka.http.scaladsl.model.headers
 
 import akka.http.impl.util._
-import org.scalatest.{ FreeSpec, MustMatchers }
+import org.scalatest._
 
 import akka.http.scaladsl.model._
 
-class HeaderSpec extends FreeSpec with MustMatchers {
+class HeaderSpec extends FreeSpec with Matchers {
```
johanandren: had to change to get access to the `shouldBe` matcher without too much fuss.

2m:
LGTM

ktoso:
LGTM, thanks @johanandren !

PR #19690, opened 2016-02-04
#19533 One2OneBidi should report truncation also if wrapped flow cancels early

### LINKED ISSUE #19533: FAILED: HighLevelOutgoingConnectionSpec
https://jenkins.akka.io:8498/job/akka-nightly/2979/consoleFull

```
[info] - should catch response stream truncation *** FAILED *** (56 milliseconds)
[info]   Expected exception akka.stream.scaladsl.One2OneBidiFlow$OutputTruncationException$ to be thrown, but no exception was thrown. (HighLevelOutgoingConnectionSpec.scala:75)
[info]   at org.scalatest.Assertions$class.newAssertionFailedException(Assertions.scala:495)
[info]   at org.scalatest.Assertions$.checkExpectedException(Assertions.scala:1582)
[info]   at org.scalatest.words.ResultOfBeWordForAType.thrownBy(ResultOfBeWordForAType.scala:39)
[info]   at akka.http.impl.engine.client.HighLevelOutgoingConnectionSpec$$anonfun$1$$anonfun$apply$mcV$sp$3$$anonfun$apply$mcV$sp$6.apply(HighLevelOutgoingConnectionSpec.scala:89)
[info]   at akka.http.impl.engine.client.HighLevelOutgoingConnectionSpec$$anonfun$1$$anonfun$apply$mcV$sp$3$$anonfun$apply$mcV$sp$6.apply(HighLevelOutgoingConnectionSpec.scala:75)
[info]   at akka.stream.testkit.Utils$.assertAllStagesStopped(Utils.scala:26)
[info]   at akka.http.impl.engine.client.HighLevelOutgoingConnectionSpec$$anonfun$1$$anonfun$apply$mcV$sp$3.apply$mcV$sp(HighLevelOutgoingConnectionSpec.scala:75)
[info]   at akka.http.impl.engine.client.HighLevelOutgoingConnectionSpec$$anonfun$1$$anonfun$apply$mcV$sp$3.apply(HighLevelOutgoingConnectionSpec.scala:75)
[info]   at akka.http.impl.engine.client.HighLevelOutgoingConnectionSpec$$anonfun$1$$anonfun$apply$mcV$sp$3.apply(HighLevelOutgoingConnectionSpec.scala:75)
```

drewhk:
The cause for the test failure was that the terminated connection flow sometimes managed to cancel the One2OneBidi before it consumed the third request. Since all the consumed two requests were served at that point, it didn't report a truncation, although there was potentially one there.

It is impossible to generate 100% correct exceptions in this case, since the cancellation might be in race with an actual completion from the request inputs (i.e. connection cancels after 2 served requests but there are no actual 3rd request, so there is no _real_ truncation). The current implementation always reports a truncation error where there was potentially one.

Fixes #19533

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/One2OneBidiFlowSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/One2OneBidiFlowSpec.scala
@@ -36,6 +36,11 @@ class One2OneBidiFlowSpec extends AkkaSpec with ConversionCheckedTripleEquals {
       a[One2OneBidiFlow.OutputTruncationException.type] should be thrownBy Await.result(test(f), 1.second)
     }
 
+    "trigger an `OutputTruncationException` if the wrapped stream cancels early" in {
+      val f = One2OneBidiFlow[Int, Int](-1) join Flow[Int].take(2)
+      a[One2OneBidiFlow.OutputTruncationException.type] should be thrownBy Await.result(test(f), 1.second)
+    }
+
     "trigger an `UnexpectedOutputException` if the wrapped stream produces out-of-order elements" in new Test() {
       inIn.sendNext(1)
       inOut.requestNext() should ===(1)
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/client/HighLevelOutgoingConnectionSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/client/HighLevelOutgoingConnectionSpec.scala
@@ -6,7 +6,7 @@ package akka.http.impl.engine.client
 
 import scala.concurrent.Await
 import scala.concurrent.duration._
-import akka.stream.{ FlowShape, ActorMaterializer }
+import akka.stream.{ ActorMaterializerSettings, FlowShape, ActorMaterializer }
 import akka.stream.scaladsl._
 import akka.stream.testkit.AkkaSpec
 import akka.http.scaladsl.{ Http, TestUtils }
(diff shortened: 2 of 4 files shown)


akka-stream-tests/src/test/scala/akka/stream/scaladsl/One2OneBidiFlowSpec.scala
```diff
@@ -36,6 +36,11 @@ class One2OneBidiFlowSpec extends AkkaSpec with ConversionCheckedTripleEquals {
       a[One2OneBidiFlow.OutputTruncationException.type] should be thrownBy Await.result(test(f), 1.second)
     }
 
+    "trigger an `OutputTruncationException` if the wrapped stream cancels early" in {
+      val f = One2OneBidiFlow[Int, Int](-1) join Flow[Int].take(2)
+      a[One2OneBidiFlow.OutputTruncationException.type] should be thrownBy Await.result(test(f), 1.second)
```
drewhk: argh, this could be racy and might already consume the third element, then get a completion, and hence trigger the exception *but for the wrong reason* (at least from the viewpoint of this test). I refactor this a bit and make the two cases distinct explicitly.

drewhk:
Hm, I am not sure about the ConnectionPoolSpec, but the LowLevelOutgoingConnectionSpec is definitely caused by my change. The real question here is to what to consider a truncation.

drewhk:
Maybe this should be flag controllable: treat inner cancellation as truncation, or just propagate the cancel silently.

PR #19700, opened 2016-02-05
prohibit concurrent write request from same persistenceId #19694

### LINKED ISSUE #19694: EventSourced must make sure that no two writes overlap, and after restart must wait for write confirmation
Journals have a really hard time dealing with multiple outstanding WriteMessage requests from the same actor: since the interface is Future-based it is very easy to just fire them off, which means that they may actually be executed in the wrong order. This leaves holes in the log in case of intervening outages and in general messes up log consistency.

Put another way: it is really hard to get this right in the journal. I doubt that any of them does it right.

There is partial protection against this in EventSourced in the form of the writeInProgress flag, but this flag is not used consistently, and it cannot sensibly be used in preRestart.

Another problem is posed by a WriteMessage right before a restart, which may still be in flight when the postRestart recovery arrives at the journal. In this case the recovering actor may see an incomplete history and subsequently reuse sequence numbers, with catastrophic consequences.

/cc @patriknw

rkuhn:
- also use scalactic.ConversionCheckedTripleEquals
- also remove usage of the setting `max-message-batch-size`

fixes #19694

### DIFF
--- a/akka-docs/rst/java/persistence.rst
+++ b/akka-docs/rst/java/persistence.rst
@@ -330,15 +330,13 @@ command, i.e. ``onPersistRejected`` is called with an exception (typically ``Uns
 Batch writes
 ------------
 
-In order to optimize throughput a persistent actor internally batches events to be stored under high load before
-writing them to the journal (as a single batch). The batch size dynamically grows from 1 under low and moderate loads
-to a configurable maximum size (default is ``200``) under high load. When using ``persistAsync`` this increases
-the maximum throughput dramatically.
-
-.. includecode:: ../scala/code/docs/persistence/PersistencePluginDocSpec.scala#max-message-batch-size
-
-A new batch write is triggered by a persistent actor as soon as a batch reaches the maximum size or if the journal completed
-writing the previous batch. Batch writes are never timer-based which keeps latencies at a minimum.
+In order to optimize throughput when using ``persistAsync``, a persistent actor
+internally batches events to be stored under high load before writing them to
+the journal (as a single batch). The batch size is dynamically determined by
+how many events are emitted during the time of a journal round-trip: after
+sending a batch to the journal no further batch can be sent before confirmation
+has been received that the previous batch has been written. Batch writes are never
+timer-based which keeps latencies at a minimum.
 
 Message deletion
 ----------------
--- a/akka-persistence/src/main/resources/reference.conf
+++ b/akka-persistence/src/main/resources/reference.conf
@@ -105,7 +105,9 @@ akka.persistence {
       # Dispatcher for message replay.
       replay-dispatcher = "akka.persistence.dispatchers.default-replay-dispatcher"
       
-      # Maximum size of a persistent message batch written to the journal.
+      # Removed: used to be the Maximum size of a persistent message batch written to the journal.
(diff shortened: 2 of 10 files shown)


rkuhn:
also added a hook in the TCK for journals to prepare for each test case—e.g. by purging all info for the used `persistenceId` from the journal so that the test can run as expected (some journals might not want to force users to drop the database all the time)

akka-persistence/src/main/resources/reference.conf
```diff
@@ -105,7 +105,9 @@ akka.persistence {
       # Dispatcher for message replay.
       replay-dispatcher = "akka.persistence.dispatchers.default-replay-dispatcher"
       
-      # Maximum size of a persistent message batch written to the journal.
+      # Removed: used to be the Maximum size of a persistent message batch written to the journal.
+      # Now this setting is without function, PersistentActor will write as many messages
+      # as it has accumulated since the last write.
       max-message-batch-size = 200
```
drewhk: shouldn't you remove the setting here, and instead log a warning if someone sets it in the config?

patriknw: That would be slightly dangerous if someone else reads the property, e.g. a journal plugin implementation.

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -51,6 +51,7 @@ private[persistence] trait Eventsourced extends Snapshotter with Stash with Stas
   private val writerUuid = UUID.randomUUID.toString
 
   private var journalBatch = Vector.empty[PersistentEnvelope]
+  // no longer used, but kept for binary compatibility
```
patriknw: there is also `def maxBatchSizeReached` perhaps be removed, or at least not called any more

patriknw: ah, I see, already removed

patriknw:
There is also a section about "Batch writes" that must updated in the rst docs

drewhk:
LGTM

akka-docs/rst/java/persistence.rst
```diff
@@ -330,15 +330,13 @@ command, i.e. ``onPersistRejected`` is called with an exception (typically ``Uns
...
-
-A new batch write is triggered by a persistent actor as soon as a batch reaches the maximum size or if the journal completed
-writing the previous batch. Batch writes are never timer-based which keeps latencies at a minimum.
+In order to optimize throughput when using ``persistAsync``, a persistent actor
+internally batches events to be stored under high load before writing them to
+the journal (as a single batch). The batch size is dynamically determined by
+how many events are emitted during the time of a journal round-trip: after
+sending a batch to the journal no further batch can be sent before confirmation
+has been received that the previous batch has been written. Batch writes are never
+timer-based which keeps latencies at a minimum.
```
patriknw: @rkuhn we also have a lambda-persistence.rst

rkuhn: Hmm. Do we still need that?

patriknw: what do you mean? do we still need the `UntypedPersistentActor`? I think we will keep it as long as we keep `UntypedActor`.

Issue #19481, opened 2016-01-15
Fix akka-bench and update plugin


ktoso:
(streams imported benchmarks break compilation under latest JMH)

Also update plugin to the latest

agolubev:
I'm having following log on jmh:run 
```
akka-bench-jmh > jmh:run
[info] Compiling 1 Scala source to /Users/alexandergolubev/Projects/akka/akka-bench-jmh/target/classes...
Processing 184 classes from /Users/alexandergolubev/Projects/akka/akka-bench-jmh/target/classes with "reflection" generator
Writing out Java source to /Users/alexandergolubev/Projects/akka/akka-bench-jmh/target/src_managed/jmh and resources to /Users/alexandergolubev/Projects/akka/akka-bench-jmh/target/resource_managed/jmh
@Param annotation is not acceptable on final fields.
   [org.openjdk.jmh.generators.reflection.RFFieldInfo@2c17ecad]

@Param annotation is not acceptable on final fields.
   [org.openjdk.jmh.generators.reflection.RFFieldInfo@4a31409b]

@Param annotation is not acceptable on final fields.
   [org.openjdk.jmh.generators.reflection.RFFieldInfo@684866e2]

@Param annotation is not acceptable on final fields.
   [org.openjdk.jmh.generators.reflection.RFFieldInfo@47208be4]

@Param annotation is not acceptable on final fields.
   [org.openjdk.jmh.generators.reflection.RFFieldInfo@54417529]

@Param annotation is not acceptable on final fields.
   [org.openjdk.jmh.generators.reflection.RFFieldInfo@14b4d2a2]
```

have this fixed in PR #19701 about Substreams Performance

ktoso:
Yeap, that's the only problem that needs to be addressed here, thanks a lot for solving it in your PR over there.

PR #19701, opened 2016-02-06
=str #19299 Performance Flow.flatMapMerge

agolubev:
Ref: #19299 
This PR creates short circuit for Async Requests that are made within the same ActorGraphInterpreter (has the same actor) by different logics. 
Main use case is when substream exchanges events with main stream and vice-versa

### DIFF
--- a/akka-bench-jmh/src/main/scala/akka/http/HttpBenchmark.scala
+++ b/akka-bench-jmh/src/main/scala/akka/http/HttpBenchmark.scala
@@ -53,7 +53,7 @@ class HttpBenchmark {
   }
 
   @TearDown
-  def shutdown():Unit ={
+  def shutdown():Unit = {
     Await.ready(Http().shutdownAllConnectionPools(), 1.second)
     binding.unbind()
     Await.result(system.terminate(), 5.seconds)
--- a/akka-docs/rst/java/code/docs/stream/GraphStageDocTest.java
+++ b/akka-docs/rst/java/code/docs/stream/GraphStageDocTest.java
@@ -443,23 +443,22 @@ public class GraphStageDocTest extends AbstractJavaTest {
   public void demonstrateAnAsynchronousSideChannel() throws Exception{
 
     // tests:
+    TestSubscriber.Probe<Integer> out = TestSubscriber.probe(system);
+    TestPublisher.Probe<Integer> in = TestPublisher.probe(0, system);
+
     CompletableFuture<Done> switchF = new CompletableFuture<>();
     Graph<FlowShape<Integer, Integer>, NotUsed> killSwitch =
       Flow.fromGraph(new KillSwitch<>(switchF));
 
-    ExecutionContext ec = system.dispatcher();
+    Source.fromPublisher(in).via(killSwitch).to(Sink.fromSubscriber(out)).run(mat);
 
-    CompletionStage<Integer> valueAfterKill = switchF.thenApply(in -> 4);
-
-
-    CompletionStage<Integer> result =
-      Source.from(Arrays.asList(1, 2, 3)).concat(Source.fromCompletionStage(valueAfterKill))
-        .via(killSwitch)
-        .runFold(0, (n, sum) -> n + sum, mat);
+    out.request(1);
+    in.sendNext(1);
+    out.expectNext(1);
 
     switchF.complete(Done.getInstance());
 
-    assertEquals(new Integer(6), result.toCompletableFuture().get(3, TimeUnit.SECONDS));
+    out.expectComplete();
   }
 
 
--- a/akka-stream/src/main/resources/reference.conf
+++ b/akka-stream/src/main/resources/reference.conf
@@ -45,16 +45,21 @@ akka {
       # this may cause an initial runtime overhead, but most of the time fusing is
       # desirable since it reduces the number of Actors that are created.
       auto-fusing = on
-      
+
       # Those stream elements which have explicit buffers (like mapAsync, mapAsyncUnordered,
       # buffer, flatMapMerge, Source.actorRef, Source.queue, etc.) will preallocate a fixed
       # buffer upon stream materialization if the requested buffer size is less than this
       # configuration parameter. The default is very high because failing early is better
       # than failing under load.
       #
-      # Buffers sized larger than this will dynamically grow/shrink and consume more memory 
+      # Buffers sized larger than this will dynamically grow/shrink and consume more memory
       # per element than the fixed size buffers.
       max-fixed-buffer-size = 1000000000
+
+      # Maximum number of sync messages that actor can process for stream to substream communication.
+      # Parameter allows to interrupt synchronous processing to get upsteam/downstream messages.
+      # Allows to accelerate message processing that happening withing same actor but keep system responsive.
+      sync-processing-limit = 1000
 
       debug {
         # Enables the fuzzing mode which increases the chance of race conditions
--- a/akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
+++ b/akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
@@ -9,6 +9,7 @@ import java.util.concurrent.atomic.{ AtomicBoolean }
 
 import akka.actor.{ ActorContext, ActorRef, ActorRefFactory, ActorSystem, ExtendedActorSystem, Props }
 import akka.event.LoggingAdapter
+import akka.stream.ActorMaterializerSettings.defaultMaxFixedBufferSize
 import akka.stream.impl._
 import com.typesafe.config.Config
 
@@ -224,7 +225,7 @@ object ActorMaterializerSettings {
    * Create [[ActorMaterializerSettings]] from a Config subsection (Scala).
    */
   def apply(config: Config): ActorMaterializerSettings =
-    ActorMaterializerSettings(
+    new ActorMaterializerSettings(
       initialInputBufferSize = config.getInt("initial-input-buffer-size"),
       maxInputBufferSize = config.getInt("max-input-buffer-size"),
       dispatcher = config.getString("dispatcher"),
@@ -234,7 +235,8 @@ object ActorMaterializerSettings {
       outputBurstLimit = config.getInt("output-burst-limit"),
       fuzzingMode = config.getBoolean("debug.fuzzing-mode"),
       autoFusing = config.getBoolean("auto-fusing"),
-      maxFixedBufferSize = config.getInt("max-fixed-buffer-size"))
+      maxFixedBufferSize = config.getInt("max-fixed-buffer-size"),
+      syncProcessingLimit = config.getInt("sync-processing-limit"))
 
   /**
    * Create [[ActorMaterializerSettings]] from individual settings (Java).
@@ -266,13 +268,14 @@ object ActorMaterializerSettings {
   def create(config: Config): ActorMaterializerSettings =
     apply(config)
 
+  private val defaultMaxFixedBufferSize = 1000
 }
 
 /**
  * This class describes the configurable properties of the [[ActorMaterializer]].
  * Please refer to the `withX` methods for descriptions of the individual settings.
  */
-final class ActorMaterializerSettings(
+final class ActorMaterializerSettings private (
   val initialInputBufferSize: Int,
   val maxInputBufferSize: Int,
   val dispatcher: String,
@@ -282,9 +285,25 @@ final class ActorMaterializerSettings(
   val outputBurstLimit: Int,
   val fuzzingMode: Boolean,
   val autoFusing: Boolean,
-  val maxFixedBufferSize: Int) {
+  val maxFixedBufferSize: Int,
+  val syncProcessingLimit: Int) {
+
+  def this(initialInputBufferSize: Int,
+           maxInputBufferSize: Int,
+           dispatcher: String,
+           supervisionDecider: Supervision.Decider,
+           subscriptionTimeoutSettings: StreamSubscriptionTimeoutSettings,
+           debugLogging: Boolean,
+           outputBurstLimit: Int,
+           fuzzingMode: Boolean,
+           autoFusing: Boolean,
+           maxFixedBufferSize: Int) {
+    this(initialInputBufferSize, maxInputBufferSize, dispatcher, supervisionDecider, subscriptionTimeoutSettings, debugLogging,
+      outputBurstLimit, fuzzingMode, autoFusing, maxFixedBufferSize, defaultMaxFixedBufferSize)
+  }
 
   require(initialInputBufferSize > 0, "initialInputBufferSize must be > 0")
+  require(syncProcessingLimit > 0, "syncProcessingLimit must be > 0")
 
   requirePowerOfTwo(maxInputBufferSize, "maxInputBufferSize")
   require(initialInputBufferSize <= maxInputBufferSize, s"initialInputBufferSize($initialInputBufferSize) must be <= maxInputBufferSize($maxInputBufferSize)")
@@ -299,10 +318,12 @@ final class ActorMaterializerSettings(
     outputBurstLimit: Int = this.outputBurstLimit,
     fuzzingMode: Boolean = this.fuzzingMode,
     autoFusing: Boolean = this.autoFusing,
-    maxFixedBufferSize: Int = this.maxFixedBufferSize) =
+    maxFixedBufferSize: Int = this.maxFixedBufferSize,
+    syncProcessingLimit: Int = this.syncProcessingLimit) = {
     new ActorMaterializerSettings(
       initialInputBufferSize, maxInputBufferSize, dispatcher, supervisionDecider, subscriptionTimeoutSettings, debugLogging,
-      outputBurstLimit, fuzzingMode, autoFusing, maxFixedBufferSize)
+      outputBurstLimit, fuzzingMode, autoFusing, maxFixedBufferSize, syncProcessingLimit)
+  }
 
   /**
    * Each asynchronous piece of a materialized stream topology is executed by one Actor
@@ -367,6 +388,13 @@ final class ActorMaterializerSettings(
     else copy(outputBurstLimit = limit)
 
   /**
+   * Limit for number of messages that can be processed synchronously in stream to substream communication
+   */
+  def withSyncProcessingLimit(limit: Int): ActorMaterializerSettings =
+    if (limit == this.syncProcessingLimit) this
+    else copy(syncProcessingLimit = limit)
+
+  /**
    * Enable to log all elements that are dropped due to failures (at DEBUG level).
    */
   def withDebugLogging(enable: Boolean): ActorMaterializerSettings =
@@ -413,12 +441,13 @@ final class ActorMaterializerSettings(
         s.subscriptionTimeoutSettings == subscriptionTimeoutSettings &&
         s.debugLogging == debugLogging &&
         s.outputBurstLimit == outputBurstLimit &&
+        s.syncProcessingLimit == syncProcessingLimit &&
         s.fuzzingMode == fuzzingMode &&
         s.autoFusing == autoFusing
     case _ ⇒ false
   }
 
-  override def toString: String = s"ActorMaterializerSettings($initialInputBufferSize,$maxInputBufferSize,$dispatcher,$supervisionDecider,$subscriptionTimeoutSettings,$debugLogging,$outputBurstLimit,$fuzzingMode,$autoFusing)"
+  override def toString: String = s"ActorMaterializerSettings($initialInputBufferSize,$maxInputBufferSize,$dispatcher,$supervisionDecider,$subscriptionTimeoutSettings,$debugLogging,$outputBurstLimit,$syncProcessingLimit,$fuzzingMode,$autoFusing)"
 }
 
 object StreamSubscriptionTimeoutSettings {
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
@@ -3,14 +3,15 @@
  */
 package akka.stream.impl.fusing
 
+import java.util
 import java.util.concurrent.TimeoutException
 import akka.actor._
 import akka.event.Logging
 import akka.stream._
+import akka.stream.impl._
 import akka.stream.impl.ReactiveStreamsCompliance._
 import akka.stream.impl.StreamLayout.{ CompositeModule, CopiedModule, Module }
 import akka.stream.impl.fusing.GraphInterpreter.{ DownstreamBoundaryStageLogic, UpstreamBoundaryStageLogic, GraphAssembly }
-import akka.stream.impl.{ ActorPublisher, ReactiveStreamsCompliance }
 import akka.stream.stage.{ GraphStageLogic, InHandler, OutHandler }
 import org.reactivestreams.{ Subscriber, Subscription }
 import scala.concurrent.forkjoin.ThreadLocalRandom
@@ -18,7 +19,6 @@ import scala.util.control.NonFatal
 import akka.stream.impl.ActorMaterializerImpl
 import akka.stream.impl.SubFusingActorMaterializerImpl
 import scala.annotation.tailrec
-import akka.stream.impl.StreamSupervisor
 
 /**
  * INTERNAL API
@@ -316,8 +316,16 @@ private[stream] final class GraphInterpreterShell(
   private var self: ActorRef = _
   lazy val log = Logging(mat.system.eventStream, self)
 
-  lazy val interpreter = new GraphInterpreter(assembly, mat, log, inHandlers, outHandlers, logics,
-    (logic, event, handler) ⇒ self ! AsyncInput(this, logic, event, handler), settings.fuzzingMode)
+  private var enqueueToShortCircuit: (Any) ⇒ Unit = _
+
+  lazy val interpreter: GraphInterpreter = new GraphInterpreter(assembly, mat, log, inHandlers, outHandlers, logics,
+    (logic, event, handler) ⇒ {
+      val asyncInput = AsyncInput(this, logic, event, handler)
+      val currentInterpreter = GraphInterpreter.currentInterpreterOrNull
+      if (currentInterpreter == null || (currentInterpreter.context ne self))
+        self ! asyncInput
+      else enqueueToShortCircuit(asyncInput)
+    }, settings.fuzzingMode, self)
 
   private val inputs = new Array[BatchingActorInputBoundary](shape.inlets.size)
   private val outputs = new Array[ActorOutputBoundary](shape.outlets.size)
@@ -332,18 +340,24 @@ private[stream] final class GraphInterpreterShell(
    * a self-message for fairness with other actors. The basic assumption here is
    * to give each input buffer slot a chance to run through the whole pipeline
    * and back (for the demand).
+   *
+   * Considered use case:
+   *  - assume a composite Sink of one expand and one fold
+   *  - assume an infinitely fast source of data
+   *  - assume maxInputBufferSize == 1
+   *  - if the event limit is greater than maxInputBufferSize * (ins + outs) than there will always be expand activity
+   *  because no data can enter “fast enough” from the outside
    */
-  private val eventLimit = settings.maxInputBufferSize * (assembly.ins.length + assembly.outs.length)
-
+  val shellEventLimit = settings.maxInputBufferSize * (assembly.ins.length + assembly.outs.length)
   // Limits the number of events processed by the interpreter on an abort event.
   // TODO: Better heuristic here
-  private val abortLimit = eventLimit * 2
+  private val abortLimit = shellEventLimit * 2
   private var resumeScheduled = false
 
   def isInitialized: Boolean = self != null
-
-  def init(self: ActorRef, subMat: SubFusingActorMaterializerImpl): Unit = {
+  def init(self: ActorRef, subMat: SubFusingActorMaterializerImpl, enqueueToShortCircuit: (Any) ⇒ Unit, eventLimit: Int): Int = {
     this.self = self
+    this.enqueueToShortCircuit = enqueueToShortCircuit
     var i = 0
     while (i < inputs.length) {
       val in = new BatchingActorInputBoundary(settings.maxInputBufferSize, i)
@@ -360,66 +374,74 @@ private[stream] final class GraphInterpreterShell(
       i += 1
     }
     interpreter.init(subMat)
-    runBatch()
+    runBatch(eventLimit)
   }
 
-  def receive(event: BoundaryEvent): Unit =
-    if (waitingForShutdown) event match {
-      case ExposedPublisher(_, id, publisher) ⇒
-        outputs(id).exposedPublisher(publisher)
-        publishersPending -= 1
-        if (canShutDown) _isTerminated = true
-      case OnSubscribe(_, _, sub) ⇒
-        tryCancel(sub)
-        subscribesPending -= 1
-        if (canShutDown) _isTerminated = true
-      case Abort(_) ⇒
-        tryAbort(new TimeoutException("Streaming actor has been already stopped processing (normally), but not all of its " +
-          s"inputs or outputs have been subscribed in [${settings.subscriptionTimeoutSettings.timeout}}]. Aborting actor now."))
-      case _ ⇒ // Ignore, there is nothing to do anyway
-    }
-    else event match {
+  def receive(event: BoundaryEvent, eventLimit: Int): Int = {
+    resumeScheduled = false
+    if (waitingForShutdown) {
+      event match {
+        case ExposedPublisher(_, id, publisher) ⇒
+          outputs(id).exposedPublisher(publisher)
+          publishersPending -= 1
+          if (canShutDown) _isTerminated = true
+        case OnSubscribe(_, _, sub) ⇒
+          tryCancel(sub)
+          subscribesPending -= 1
+          if (canShutDown) _isTerminated = true
+        case Abort(_) ⇒
+          tryAbort(new TimeoutException("Streaming actor has been already stopped processing (normally), but not all of its " +
+            s"inputs or outputs have been subscribed in [${settings.subscriptionTimeoutSettings.timeout}}]. Aborting actor now."))
+        case _ ⇒ // Ignore, there is nothing to do anyway
+      }
+      eventLimit
+    } else event match {
       // Cases that are most likely on the hot path, in decreasing order of frequency
       case OnNext(_, id: Int, e: Any) ⇒
         if (GraphInterpreter.Debug) println(s"${interpreter.Name}  onNext $e id=$id")
         inputs(id).onNext(e)
-        runBatch()
+        runBatch(eventLimit)
       case RequestMore(_, id: Int, demand: Long) ⇒
         if (GraphInterpreter.Debug) println(s"${interpreter.Name}  request  $demand id=$id")
         outputs(id).requestMore(demand)
-        runBatch()
+        runBatch(eventLimit)
       case Resume(_) ⇒
         if (GraphInterpreter.Debug) println(s"${interpreter.Name}  resume")
-        resumeScheduled = false
-        if (interpreter.isSuspended) runBatch()
+        if (interpreter.isSuspended) runBatch(eventLimit) else eventLimit
       case AsyncInput(_, logic, event, handler) ⇒
         interpreter.runAsyncInput(logic, event, handler)
-        runBatch()
+        if (eventLimit == 1 && interpreter.isSuspended) {
+          sendResume(true)
+          0
+        } else runBatch(eventLimit - 1)
 
       // Initialization and completion messages
       case OnError(_, id: Int, cause: Throwable) ⇒
         if (GraphInterpreter.Debug) println(s"${interpreter.Name}  onError id=$id")
         inputs(id).onError(cause)
-        runBatch()
+        runBatch(eventLimit)
       case OnComplete(_, id: Int) ⇒
         if (GraphInterpreter.Debug) println(s"${interpreter.Name}  onComplete id=$id")
         inputs(id).onComplete()
-        runBatch()
+        runBatch(eventLimit)
       case OnSubscribe(_, id: Int, subscription: Subscription) ⇒
         if (GraphInterpreter.Debug) println(s"${interpreter.Name}  onSubscribe id=$id")
         subscribesPending -= 1
         inputs(id).onSubscribe(subscription)
-        runBatch()
+        runBatch(eventLimit)
       case Cancel(_, id: Int) ⇒
         if (GraphInterpreter.Debug) println(s"${interpreter.Name}  cancel id=$id")
         outputs(id).cancel()
-        runBatch()
+        runBatch(eventLimit)
       case SubscribePending(_, id: Int) ⇒
         outputs(id).subscribePending()
+        eventLimit
       case ExposedPublisher(_, id, publisher) ⇒
         publishersPending -= 1
         outputs(id).exposedPublisher(publisher)
+        eventLimit
     }
+  }
 
   private var _isTerminated = false
   def isTerminated: Boolean = _isTerminated
@@ -430,16 +452,16 @@ private[stream] final class GraphInterpreterShell(
 
   private val resume = Resume(this)
 
-  private def runBatch(): Unit = {
+  def sendResume(sendResume: Boolean): Unit = {
+    resumeScheduled = true
+    if (sendResume) self ! resume
+    else enqueueToShortCircuit(resume)
+  }
+
+  def runBatch(actorEventLimit: Int): Int = {
     try {
-      val effectiveLimit = {
-        if (!settings.fuzzingMode) eventLimit
-        else {
-          if (ThreadLocalRandom.current().nextBoolean()) Thread.`yield`()
-          ThreadLocalRandom.current().nextInt(2) // 1 or zero events to be processed
-        }
-      }
-      interpreter.execute(effectiveLimit)
+      val usingShellLimit = shellEventLimit < actorEventLimit
+      val remainingQuota = interpreter.execute(Math.min(actorEventLimit, shellEventLimit))
       if (interpreter.isCompleted) {
         // Cannot stop right away if not completely subscribed
         if (canShutDown) _isTerminated = true
@@ -449,13 +471,13 @@ private[stream] final class GraphInterpreterShell(
             override def run(): Unit = self ! Abort(GraphInterpreterShell.this)
           })
         }
-      } else if (interpreter.isSuspended && !resumeScheduled) {
-        resumeScheduled = true
-        self ! resume
-      }
+      } else if (interpreter.isSuspended && !resumeScheduled) sendResume(!usingShellLimit)
+
+      if (usingShellLimit) actorEventLimit - shellEventLimit + remainingQuota else remainingQuota
     } catch {
       case NonFatal(e) ⇒
         tryAbort(e)
+        actorEventLimit - 1
     }
   }
 
@@ -498,7 +520,7 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
 
   def tryInit(shell: GraphInterpreterShell): Boolean =
     try {
-      shell.init(self, subFusingMaterializerImpl)
+      currentLimit = shell.init(self, subFusingMaterializerImpl, enqueueToShortCircuit(_), currentLimit)
       if (GraphInterpreter.Debug) println(s"registering new shell in ${_initial}\n  ${shell.toString.replace("\n", "\n  ")}")
       if (shell.isTerminated) false
       else {
@@ -511,9 +533,20 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
         false
     }
 
+  //this limits number of messages that can be processed synchronously during one actor receive.
+  private val eventLimit: Int = _initial.mat.settings.syncProcessingLimit
+  private var currentLimit: Int = eventLimit
+  //this is a var in order to save the allocation when no short-circuiting actually happens
+  private var shortCircuitBuffer: util.LinkedList[Any] = null
+
+  def enqueueToShortCircuit(input: Any): Unit = {
+    if (shortCircuitBuffer == null) shortCircuitBuffer = new util.LinkedList[Any]()
+    shortCircuitBuffer.add(input)
+  }
+
   def registerShell(shell: GraphInterpreterShell): ActorRef = {
     newShells ::= shell
-    self ! Resume
+    enqueueToShortCircuit(Resume)
     self
   }
 
@@ -537,35 +570,57 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
   override def preStart(): Unit = {
     tryInit(_initial)
     if (activeInterpreters.isEmpty) context.stop(self)
+    else if (shortCircuitBuffer != null) shortCircuitBatch()
+  }
+
+  private def shortCircuitBatch(): Unit = {
+    while (!shortCircuitBuffer.isEmpty && currentLimit > 0 && activeInterpreters.nonEmpty)
+      shortCircuitBuffer.poll() match {
+        case b: BoundaryEvent ⇒ processEvent(b)
+        case Resume           ⇒ finishShellRegistration()
+      }
+    if (!shortCircuitBuffer.isEmpty && currentLimit == 0) self ! Resume
+  }
+
+  private def processEvent(b: BoundaryEvent): Unit = {
+    val shell = b.shell
+    if (!shell.isTerminated && (shell.isInitialized || tryInit(shell))) {
+      currentLimit = shell.receive(b, currentLimit)
+      if (shell.isTerminated) {
+        activeInterpreters -= shell
+        if (activeInterpreters.isEmpty && newShells.isEmpty) context.stop(self)
+      }
+    }
   }
 
   override def receive: Receive = {
     case b: BoundaryEvent ⇒
-      val shell = b.shell
-      if (!shell.isTerminated && (shell.isInitialized || tryInit(shell))) {
-        shell.receive(b)
-        if (shell.isTerminated) {
-          activeInterpreters -= shell
-          if (activeInterpreters.isEmpty && newShells.isEmpty) context.stop(self)
-        }
-      }
-    case Resume ⇒ finishShellRegistration()
+      currentLimit = eventLimit
+      processEvent(b)
+      if (shortCircuitBuffer != null) shortCircuitBatch()
+
+    case Resume ⇒
+      currentLimit = eventLimit
+      if (shortCircuitBuffer != null) shortCircuitBatch()
+
     case StreamSupervisor.PrintDebugDump ⇒
-      println(s"activeShells:")
+      val builder = new java.lang.StringBuilder(s"activeShells (actor: $self):\n")
       activeInterpreters.foreach { shell ⇒
-        println("  " + shell.toString.replace("\n", "\n  "))
-        shell.interpreter.dumpWaits()
+        builder.append("  " + shell.toString.replace("\n", "\n  "))
+        builder.append(shell.interpreter.toString)
       }
-      println(s"newShells:")
+      builder.append(s"newShells:")
       newShells.foreach { shell ⇒
-        println("  " + shell.toString.replace("\n", "\n  "))
-        shell.interpreter.dumpWaits()
+        builder.append("  " + shell.toString.replace("\n", "\n  "))
+        builder.append(shell.interpreter.toString)
       }
+      println(builder)
   }
 
   override def postStop(): Unit = {
     val ex = AbruptTerminationException(self)
     activeInterpreters.foreach(_.tryAbort(ex))
+    activeInterpreters = Set.empty[GraphInterpreterShell]
     newShells.foreach(s ⇒ if (tryInit(s)) s.tryAbort(ex))
   }
 }
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
@@ -4,13 +4,13 @@
 package akka.stream.impl.fusing
 
 import java.util.Arrays
+import akka.actor.ActorRef
 import akka.event.LoggingAdapter
 import akka.stream.stage._
 import scala.annotation.tailrec
 import scala.collection.immutable
 import akka.stream._
 import akka.stream.impl.StreamLayout._
-import akka.stream.impl.fusing.GraphStages.MaterializedValueSource
 import scala.concurrent.forkjoin.ThreadLocalRandom
 import scala.util.control.NonFatal
 import java.{ util ⇒ ju }
@@ -193,12 +193,12 @@ private[akka] object GraphInterpreter {
 
     override def toString: String =
       "GraphAssembly\n  " +
-        stages.mkString("[", ",", "]") + "\n  " +
-        originalAttributes.mkString("[", ",", "]") + "\n  " +
-        ins.mkString("[", ",", "]") + "\n  " +
-        inOwners.mkString("[", ",", "]") + "\n  " +
-        outs.mkString("[", ",", "]") + "\n  " +
-        outOwners.mkString("[", ",", "]")
+        stages.mkString("Stages: [", ",", "]") + "\n  " +
+        originalAttributes.mkString("Attributes: [", ",", "]") + "\n  " +
+        ins.mkString("Inlets: [", ",", "]") + "\n  " +
+        inOwners.mkString("InOwners: [", ",", "]") + "\n  " +
+        outs.mkString("Outlets: [", ",", "]") + "\n  " +
+        outOwners.mkString("OutOwners: [", ",", "]")
   }
 
   object GraphAssembly {
@@ -348,7 +348,8 @@ private[stream] final class GraphInterpreter(
   val outHandlers: Array[OutHandler], // Lookup table for the outHandler of the connection
   val logics: Array[GraphStageLogic], // Array of stage logics
   val onAsyncInput: (GraphStageLogic, Any, (Any) ⇒ Unit) ⇒ Unit,
-  val fuzzingMode: Boolean) {
+  val fuzzingMode: Boolean,
+  val context: ActorRef) {
   import GraphInterpreter._
 
   // Maintains additional information for events, basically elements in-flight, or failure.
@@ -523,13 +524,13 @@ private[stream] final class GraphInterpreter(
    * Executes pending events until the given limit is met. If there were remaining events, isSuspended will return
    * true.
    */
-  def execute(eventLimit: Int): Unit = {
+  def execute(eventLimit: Int): Int = {
     if (Debug) println(s"$Name ---------------- EXECUTE $queueStatus (running=$runningStages, shutdown=$shutdownCounters)")
     val currentInterpreterHolder = _currentInterpreter.get()
     val previousInterpreter = currentInterpreterHolder(0)
     currentInterpreterHolder(0) = this
+    var eventsRemaining = eventLimit
     try {
-      var eventsRemaining = eventLimit
       while (eventsRemaining > 0 && queueTail != queueHead) {
         val connection = dequeue()
         try processEvent(connection)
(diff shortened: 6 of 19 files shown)


agolubev:
This one was interesting

agolubev:
seems like build failed is unrelated 
```
[info] - should write a posted file to a temporary file on disk *** FAILED *** (1 second, 62 milliseconds)
[info]   Request was neither completed nor rejected within 1 second (FileUploadDirectivesSpec.scala:35)
[info]   at akka.http.scaladsl.testkit.TestFrameworkInterface$Scalatest$class.failTest(TestFrameworkInterface.scala:24)
[info]   at akka.http.scaladsl.server.RoutingSpec.failTest(RoutingSpec.scala:19)
[info]   at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.failNeitherCompletedNorRejected(RouteTestResultComponent.scala:95)
[info]   at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.rawResponse(RouteTestResultComponent.scala:60)
[info]   at akka.http.scaladsl.testkit.RouteTestResultComponent$RouteTestResult.response(RouteTestResultComponent.scala:41)
[info]   at akka.http.scaladsl.testkit.RouteTest$class.response(RouteTest.scala:59)
[info]   at akka.http.scaladsl.server.RoutingSpec.response(RoutingSpec.scala:19)
[info]   at akka.http.scaladsl.testkit.RouteTest$class.responseAs(RouteTest.scala:68)
[info]   at akka.http.scaladsl.server.RoutingSpec.responseAs(RoutingSpec.scala:19)
[info]   at akka.http.scaladsl.server.directives.FileUploadDirectivesSpec$$anonfun$1$$anonfun$apply$mcV$sp$1$$anonfun$apply$mcV$sp$2.apply$mcZ$sp(FileUploadDirectivesSpec.scala:37)
[info]   at akka.http.scaladsl.server.directives.FileUploadDirectivesSpec$$anonfun$1$$anonfun$apply$mcV$sp$1$$anonfun$apply$mcV$sp$2.apply(FileUploadDirectivesSpec.scala:35)
[info]   at akka.http.scaladsl.server.directives.FileUploadDirectivesSpec$$anonfun$1$$anonfun$apply$mcV$sp$1$$anonfun$apply$mcV$sp$2.apply(FileUploadDirectivesSpec.scala:35)
```

johanandren:
OK TO TEST

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -681,7 +681,7 @@ private[akka] final case class MapAsync[In, Out](parallelism: Int, f: In ⇒ Fut
           case NonFatal(ex) ⇒
             if (decider(ex) == Supervision.Stop) failStage(ex)
         }
-        if (todo < parallelism) tryPull(in)
+        if (todo < parallelism && !hasBeenPulled(in)) tryPull(in)
```
drewhk: interesting, so this was a bug uncovered.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -348,7 +347,8 @@ private[stream] final class GraphInterpreter(
   val outHandlers: Array[OutHandler], // Lookup table for the outHandler of the connection
   val logics: Array[GraphStageLogic], // Array of stage logics
   val onAsyncInput: (GraphStageLogic, Any, (Any) ⇒ Unit) ⇒ Unit,
-  val fuzzingMode: Boolean) {
+  val fuzzingMode: Boolean,
+  val context: AnyRef) {
```
drewhk: I was thinking that maybe the context should be simply the GraphIntrerpreterShell since the only synchronous cross-interpreter communication can be inside the same shell. Then this parameter would be at least typed.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -319,8 +319,15 @@ private[stream] final class GraphInterpreterShell(
...
 
-  lazy val interpreter = new GraphInterpreter(assembly, mat, log, inHandlers, outHandlers, logics,
-    (logic, event, handler) ⇒ self ! AsyncInput(this, logic, event, handler), settings.fuzzingMode)
+  private var canEnqueueToShortCircuit: (AsyncInput) ⇒ Boolean = _
+
+  lazy val interpreter: GraphInterpreter = new GraphInterpreter(assembly, mat, log, inHandlers, outHandlers, logics,
+    (logic, event, handler) ⇒ {
+      val asyncInput = AsyncInput(this, logic, event, handler)
+      val currentInterpreter = GraphInterpreter.currentInterpreterOrNull
+      if (currentInterpreter == null || (currentInterpreter.context ne self) || !canEnqueueToShortCircuit(asyncInput))
```
drewhk: can't we pull up this logic into the `invoke` method of the AsyncCallback instead?

agolubev: Yes, but perhaps need to have canEnqueueToShortCircuit(from ActorGraphIntrpreter as it has queue of such async calls) function there in GraphIntrpreter

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
agolubev: Actually ActorGraphInterpreter can have several shells - initial and for all sub streams (```ActorGraphInterpreter.registerShell```). I believe that  ActorGraphInterpreter as Actor executes all its shells in one thread so we can accelerate messages from initial shell to substream shells and back. I can make context parameter as ActorRef to make it typed. Let me know if my assumptions are not correct.

agolubev:
flat map merge benchmark gave:
```
[info] Benchmark                                           (NumberOfStreams)   Mode  Cnt     Score     Error   Units
[info] FlatMapMergeBenchmark.flat_map_merge_100k_elements                  0  thrpt   10  7891.063 ± 639.963  ops/ms
[info] FlatMapMergeBenchmark.flat_map_merge_100k_elements                  1  thrpt   10  3598.405 ±  90.374  ops/ms
[info] FlatMapMergeBenchmark.flat_map_merge_100k_elements                 10  thrpt   10  3619.143 ± 138.566  ops/ms
```

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
drewhk: Ah, I thought that shell is the one that hosts multiple interpreters. I guess we will need to sort this out a bit more and document, because currently slightly confusing. Anyway, now you know better than me so proceed as you wish :)

ktoso:
Also resolves https://github.com/akka/akka/issues/19481

akka-stream/src/main/resources/reference.conf
```diff
@@ -46,6 +46,10 @@ akka {
       # desirable since it reduces the number of Actors that are created.
       auto-fusing = on
 
+      # Maximum size of buffers used when stream exchange messages with substreams and back
+      # Communication is done in syncronous way to accelerate the process
```
ktoso: synchronous

akka-bench-jmh/src/main/scala/akka/stream/FlatMapMergeBenchmark.scala
```diff
@@ -23,7 +23,7 @@ class FlatMapMergeBenchmark {
   val NumberOfElements = 100000
 
   @Param(Array("0", "1", "10"))
-  val NumberOfStreams = 0
+  var NumberOfStreams = 0
```
ktoso: Thanks for this and the other fixes in akka-bench

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
agolubev: will try this as it supposed to accelerate constructing GraphInterpreter

agolubev:
PR is ready for review. Performance is the same for base case. And better for Stream<->Substream communications

drewhk:
The failure in FlattenTest seems to be a valid failure here.

agolubev:
Yes, looking into this now

agolubev:
Finally fixed all tests. PR is ready for review

agolubev:
ugh =(

ktoso:
Though likely not related to this PR:

```
[info] - Demonstrate an asynchronous side channel *** FAILED *** (28 milliseconds)
[info]   List(1, 2, 3, 4) did not equal List(1, 2, 3) (GraphStageDocSpec.scala:317)
[info]   at docs.stream.GraphStageDocSpec$$anonfun$13.apply$mcV$sp(GraphStageDocSpec.scala:317)
[info]   at docs.stream.GraphStageDocSpec$$anonfun$13.apply(GraphStageDocSpec.scala:272)
[info]   at docs.stream.GraphStageDocSpec$$anonfun$13.apply(GraphStageDocSpec.scala:272)
[info]   at org.scalatest.Transformer$$anonfun$apply$1.apply$mcV$sp(Transformer.scala:22)
[info]   at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
```

akka-stream/src/main/resources/reference.conf
```diff
@@ -45,17 +45,22 @@ akka {
...
       # buffer upon stream materialization if the requested buffer size is less than this
       # configuration parameter. The default is very high because failing early is better
       # than failing under load.
       #
-      # Buffers sized larger than this will dynamically grow/shrink and consume more memory 
+      # Buffers sized larger than this will dynamically grow/shrink and consume more memory
       # per element than the fixed size buffers.
       max-fixed-buffer-size = 1000000000
 
+      # Maximum number of sync messages that actor can process for stream to substream communication.
```
viktorklang: Why not reuse the `throughput` (or derive from it) setting of the dispatcher?

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -510,6 +518,20 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
         false
     }
 
+  //this limits number of messages that can be processed synchronously during one actor receive.
+  private val shortCircuitLimit: Int = _initial.mat.settings.syncProcessingLimit
+  private var shortCircuitBuffer: ByteBuffer[BoundaryEvent] = _
+  private var currentLimit: Int = shortCircuitLimit
+
+  def enqueueToShortCircuit(input: AsyncInput): Boolean = {
+    if (shortCircuitBuffer == null) shortCircuitBuffer = FixedSizeBuffer[BoundaryEvent](shortCircuitLimit)
```
viktorklang: why check this for each enqueue?

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -510,6 +518,20 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
         false
     }
 
+  //this limits number of messages that can be processed synchronously during one actor receive.
+  private val shortCircuitLimit: Int = _initial.mat.settings.syncProcessingLimit
+  private var shortCircuitBuffer: ByteBuffer[BoundaryEvent] = _
+  private var currentLimit: Int = shortCircuitLimit
```
viktorklang: Doesn't ByteBuffer track its capacity already? (i.e. do we need this var?)

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -193,12 +193,12 @@ private[akka] object GraphInterpreter {
...
     override def toString: String =
       "GraphAssembly\n  " +
-        stages.mkString("[", ",", "]") + "\n  " +
-        originalAttributes.mkString("[", ",", "]") + "\n  " +
-        ins.mkString("[", ",", "]") + "\n  " +
-        inOwners.mkString("[", ",", "]") + "\n  " +
-        outs.mkString("[", ",", "]") + "\n  " +
-        outOwners.mkString("[", ",", "]")
+        stages.mkString("Stages: [", ",", "]") + "\n  " +
+        originalAttributes.mkString("Attributs: [", ",", "]") + "\n  " +
```
viktorklang: typo on Attributs

akka-stream/src/main/scala/akka/stream/impl/fusing/IteratorInterpreter.scala
```diff
@@ -145,7 +145,8 @@ private[akka] class IteratorInterpreter[I, O](val input: Iterator[I], val ops: S
       outHandlers,
       logics,
       (_, _, _) ⇒ throw new UnsupportedOperationException("IteratorInterpreter does not support asynchronous events."),
-      fuzzingMode = false)
+      fuzzingMode = false,
+      null)
```
viktorklang: why null?

drewhk: IteratorInterpreter does not have support for asynchronous side channels, hence it does not need a context.

akka-stream/src/main/resources/reference.conf
drewhk: That is a different thing. This is related to how many internal messages it processes *per actor message*. If you have your throughput at 1000, you probably don't want to set this to 1000. I am not sure why this new configuration setting is needed though.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
drewhk: I don't even know why we need a ByteBuffer here. Isn't that a mistake there (probably rogue code completion)?

akka-stream/src/main/scala/akka/stream/impl/fusing/IteratorInterpreter.scala
viktorklang: Yes, but a NullObject is almost always better than `null`.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
drewhk: The buffer is only needed if asynchronous internal side-channels are used. The idea is to trade one branch instead of having a buffer that is not needed in most of the cases.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
viktorklang: Is there a regression test?

akka-stream/src/main/scala/akka/stream/impl/fusing/IteratorInterpreter.scala
drewhk: well, it might, just don't mistakenly compare them to each other and think they represent the same execution island.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
viktorklang: How does `syncProcessingLimit` relate to `eventLimit`?

akka-stream/src/main/resources/reference.conf
viktorklang: I think this should be derived rather than added as yet another config parameter.
It's hard enough to reason about the semantics of configuration changes already IMO.

drewhk: I agree.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -315,8 +312,15 @@ private[stream] final class GraphInterpreterShell(
   private var self: ActorRef = _
   lazy val log = Logging(mat.system.eventStream, self)
 
-  lazy val interpreter = new GraphInterpreter(assembly, mat, log, inHandlers, outHandlers, logics,
-    (logic, event, handler) ⇒ self ! AsyncInput(this, logic, event, handler), settings.fuzzingMode)
+  private var canEnqueueToShortCircuit: (AsyncInput) ⇒ Boolean = _
```
viktorklang: IMO this shouldn't be intilized to `null` use a cached default function.

akka-stream/src/main/scala/akka/stream/impl/fusing/IteratorInterpreter.scala
viktorklang: To be honest, the large number of nulls in this part of the code scares the hell out of me.

drewhk: Why?

viktorklang: (this part of the code referred to the interpreter)
It requires analyzing the full code to then deduce the lifecycle (when it has a value which doesn't blow up, and if it is appropriately guarded wherever needed).

drewhk: This field can be a proper non-null value, just use it carefully that was my point. Other places in the interpreter internally use nulls because that class is performance intensive. The reasons are:
 - Option add a new allocation (we eliminated as many object headers as possible, by for example folding two arrays into one: https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala#L360) and an extra branch everywhere
 - Adding an explicit "invalid" subclass where interfaces are expected is not always possible, plus might make certain sites megamorphic which weren't before
 - even if any of the above might work with minimal performance impact, they would only throw an exception anyway, which is not really more useful than the NPE since it is an assertion error which the user cannot do anything about (internal bug) and I don't need the exception message because the NPE stacktrace identifies the issue immediately.

That said, if you try to make a different approach that does not use nulls, and does not add a significant benchmark impact then we should take it. But currently the interpreter is capable of being completely inlined by JIT (see https://github.com/akka/akka/issues/19299#issuecomment-181402288) and every branch had been very carefully evaluated in the main loop whether it is truly necessary. I mean matches were replaced by explicitly balanced bitfield based chained ifs (https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala#L591), branches were designed so that the common, non-failure path is preferred while the error part contains the extra branches to decode the *exact* error (https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/stage/GraphStage.scala#L346), etc.

The interpreter is the dungeon of streams and it uses patterns that we would not use elsewhere.

agolubev:
>Though likely not related to this PR:

I'll check this - as anything that usin AsyncCall (future in failed test) most probably is related

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
agolubev: will fix. Thanks

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
agolubev: >I don't even know why we need a ByteBuffer here.

It's light and fast queue implementation, initially it was mutable.Queue but profiling slowed it's slow. Any suggestions are welcome though

drewhk: That is OK, but why ByteBuffer? But now I see, you renamed stream.impl.Buffer at L10, but this is super confusing now :). We have FixedSizeBuffer (or called something like that) that is array based and you can use it simply.

agolubev: Ok, will change it to FixedSizeBuffer. Initial intention was to abstract from implementation (as it's enough to have just ByteBuffer methods) - but as soon as it's internal thing it's not so important

agolubev: >How does syncProcessingLimit relate to eventLimit?

eventLimit is about processing connection events and maximum for this limit is "number of connections" * 2. This limit works within one shell.
syncProcessingLimit works across several shells for one actor and mazimum should relate to number of shells in this actor. The reason have it here is that from some cases there is a ping pong (asyncCalls) between stream and substream. So actor receive method can process (synchronously) such events very long time. 

Still idea is very similar to for eventLimit and syncProcessingLimit. We are doing sync messages processing for asyncCalls only for for now (which is futures and stream to substream communications). Potentially we can add other events to shortCircuit queue to get rid of actor async overhead.

viktorklang: @agolubev so `syncProcessingLimit` is isomorphic to `throughput` for the dispatcher? Then I'd reuse that setting, or derive something from it. Because the problem is that the only way you can make an informed decision about it is if you know everything that is executing on the same threadpool.

drewhk: Isomorphic, but you can't reuse it. If you set throughput to 1000 and you would take that setting here as well, then you would execute 1000*1000 potential events before putting back the thread to the thread-pool. I don't think one of the settings easily carries to the other. The previous heuristic was set so that a graph can roughly execute a full push-pull cycle before giving back the actor the execution (so it can get the next external event from the mailbox if throughput allows it).

akka-stream/src/main/scala/akka/stream/impl/fusing/IteratorInterpreter.scala
viktorklang: @drewhk Thanks for the explanation, Endre! (I appreciate it!)


I think it boils down to two things: hard-to-maintain code (i.e. mutable state etc) should be created when performance matters and the harder-to-maintain solution has been benchmarked to be significantly (contextual) better than an easier-to-maintain solution, and in which case there should be both a bench AND comments in the code explaining why the code is there, at the very least to console the maintainer that they're having a rougher time for a reason.

and

`null` should be used when defensible, which ideally also should be documented. :)

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
viktorklang: @drewhk  hence the `or derive something from it.` :-)

akka-stream/src/main/scala/akka/stream/impl/fusing/IteratorInterpreter.scala
drewhk: I have a bench, that is what I used back then to measure this https://github.com/akka/akka/blob/master/akka-bench-jmh/src/main/scala/akka/stream/InterpreterBenchmark.scala (it was slow after I fixed some bugs, and I did all the tricks here to make it fast again. That benchmark hosted the old interpreter as well for comparison). I did a lot of docs (well, could be more, but compared to other internal code, I think it is quite good ;) ): 

https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala#L271
https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala#L68

And there are many more comments in the code itself. I did not explicitly document the use of nulls, but the effort for documentation was there, especially because you asked for more docs before so I delivered :)

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
drewhk: my point was that throughput, while isomorphic, does not give good advice on setting this. It is an orthogonal setting here.

viktorklang: Well yes and no, without fusion throughput controls the number of ops processed in a row, right?

drewhk: Let's discuss actual formulas instead because that makes more sense.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
agolubev: will add tests for mapAsync and mapAsyncUnordered

akka-bench-jmh/src/main/scala/akka/http/HttpBenchmark.scala
```diff
@@ -53,7 +53,7 @@ class HttpBenchmark {
   }
 
   @TearDown
-  def shutdown() = {
+  def shutdown() {
```
viktorklang: Procedure syntax has been deprecated so it would be better to add a Unit return type.

agolubev: Ok, will change through benchmarks

agolubev:
Tests Passed, finally!

akka-stream/src/main/scala/akka/stream/impl/fusing/IteratorInterpreter.scala
hepin1989: FYI http://shipilev.net/blog/2015/black-magic-method-dispatch/

drewhk:
Yay!

akka-docs/rst/java/code/docs/stream/GraphStageDocTest.java
```diff
@@ -442,23 +442,22 @@ public void apply(Done param) throws Exception {
...
-        .via(killSwitch)
-        .runFold(0, (n, sum) -> n + sum, mat);
+    out.request(1);
+    in.sendNext(1);
+    out.expectNext(1);
 
     switchF.complete(Done.getInstance());
 
-    assertEquals(new Integer(6), result.toCompletableFuture().get(3, TimeUnit.SECONDS));
+    out.expectComplete();
```
drewhk: thanks, this was racy before.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -512,6 +520,20 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
         false
     }
 
+  //this limits number of messages that can be processed synchronously during one actor receive.
+  private val shortCircuitLimit: Int = _initial.mat.settings.syncProcessingLimit
```
drewhk: I'm afraid that the default 32 is too short for such a global limit. In the case of HTTP there are 20 something stages fused, so those can be 40 events worst case for a full push-pull cycle. Probably less, but it feels still dangerously close to the "edge".

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -540,27 +562,51 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
     if (activeInterpreters.isEmpty) context.stop(self)
   }
 
+  private def shortCircuitBatch(): Unit = {
+    @tailrec
+    def shortCircuitBatchUtil(): Unit =
```
drewhk: Maybe it's just me, but I would prefer a normal `while` here.

drewhk:
LGTM!

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
agolubev: can we stick to 100 than? Performance is similar on jmh on limit=64 comparing to limit=1000

drewhk: The thing is that most pipelines need some external event eventually, so even if you set the limit to Int.MaxValue it will still stop after a few rounds since buffers might become empty. Where it matters is where internal stages chat a log among each other, like a mapConcat+filter, or when a larger graph is created like Http. I used a heuristic here before, but that had no upper cap (which is likely necessary). Maybe a constant is OK, but I would benchmark HTTP and set a constant which has no visible performance impact.

akka-stream/src/main/resources/reference.conf
rkuhn: From what should this setting be derived? Neither of the boundaries are sensible (0 or infinity), and I don’t see any other landmark values to base this on.

akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
```diff
@@ -207,12 +207,13 @@ object ActorMaterializerSettings {
     subscriptionTimeoutSettings: StreamSubscriptionTimeoutSettings,
     debugLogging: Boolean,
     outputBurstLimit: Int,
+    syncProcessingLimit: Int,
```
rkuhn: nope, binary compatibility; which reminds me that we need to switch on MiMa

viktorklang: Yeah, yesterday ;)

-- 
Cheers,
√

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
rkuhn: What would that buy us? I’d rather have an NPE than a non-working feature.

rkuhn: Another issue, though: the purpose of this function is to actually enqueue if possible, which is a bit misrepresented in the name here. I would suggest `hasEnqueuedToShortCircuit`.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -431,16 +435,19 @@ private[stream] final class GraphInterpreterShell(
 
   private val resume = Resume(this)
 
-  private def runBatch(): Unit = {
+  private var currentLimit = effectiveLimit()
```
rkuhn: shouldn’t this be decremented somewhere? as currently coded this does not make sense to me

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -392,6 +395,7 @@ private[stream] final class GraphInterpreterShell(
       case Resume(_) ⇒
         if (GraphInterpreter.Debug) println(s"${interpreter.Name}  resume")
         resumeScheduled = false
+        currentLimit = effectiveLimit()
```
rkuhn: assuming that it will be used for something, why is it not set in all other cases (apart from AsyncInput) as well?

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
rkuhn: @drewhk This limit is only for AsyncInput, so I think your points are invalid.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -512,6 +520,20 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
...
     }
 
+  //this limits number of messages that can be processed synchronously during one actor receive.
+  private val shortCircuitLimit: Int = _initial.mat.settings.syncProcessingLimit
+  private var shortCircuitBuffer: FixedSizeBuffer[BoundaryEvent] = null
+  private var currentLimit: Int = shortCircuitLimit
+
+  def enqueueToShortCircuit(input: AsyncInput): Boolean = {
+    if (shortCircuitBuffer == null) shortCircuitBuffer = FixedSizeBuffer[BoundaryEvent](shortCircuitLimit)
+    if (!shortCircuitBuffer.isFull && currentLimit > 0) {
```
rkuhn: This won’t work: we cannot execute async callbacks out of order, that would be tremendously confusing for people (and for our own code as well). This means that this buffer cannot be bounded and we must always enqueue.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
rkuhn: :+1:

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -540,27 +562,51 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
     if (activeInterpreters.isEmpty) context.stop(self)
   }
 
+  private def shortCircuitBatch(): Unit = {
+    @tailrec
+    def shortCircuitBatchUtil(): Unit =
+      if (!shortCircuitBuffer.isEmpty) {
+        val input = shortCircuitBuffer.dequeue()
+        processEvent(input)
+        shortCircuitBatchUtil()
```
rkuhn: This is an endless loop if it encounters the right circumstances (buffer being refilled while draining it).

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -540,27 +562,51 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
...
-      println(s"activeShells:")
-      activeInterpreters.foreach { shell ⇒
-        println("  " + shell.toString.replace("\n", "\n  "))
-        shell.interpreter.dumpWaits()
-      }
-      println(s"newShells:")
-      newShells.foreach { shell ⇒
-        println("  " + shell.toString.replace("\n", "\n  "))
-        shell.interpreter.dumpWaits()
+      _initial.mat.synchronized {
```
rkuhn: wat?

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -540,27 +562,51 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
     if (activeInterpreters.isEmpty) context.stop(self)
   }
 
+  private def shortCircuitBatch(): Unit = {
+    @tailrec
+    def shortCircuitBatchUtil(): Unit =
+      if (!shortCircuitBuffer.isEmpty) {
+        val input = shortCircuitBuffer.dequeue()
+        processEvent(input)
```
rkuhn: we should probably stop when the actor was stopped during this call

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -348,7 +348,8 @@ private[stream] final class GraphInterpreter(
   val outHandlers: Array[OutHandler], // Lookup table for the outHandler of the connection
   val logics: Array[GraphStageLogic], // Array of stage logics
   val onAsyncInput: (GraphStageLogic, Any, (Any) ⇒ Unit) ⇒ Unit,
-  val fuzzingMode: Boolean) {
+  val fuzzingMode: Boolean,
+  val context: ActorRef) {
```
rkuhn: This does not seem to be used.

rkuhn:
Sorry for the delay @agolubev. I don’t think that this approach can work, in particular the ordering requirement and the desire to only execute a bounded number of steps seem to be at odds with each other. I don’t have any brilliant ideas right now as well …

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
agolubev: when printing debug information for several shells it's showing several dumps simultaneously so lines are intermixed. Not sure if there is more elegant way.

rkuhn: ah, okay; then a better solution is to construct the whole string an `println` just once

agolubev: got it

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
agolubev: ok

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
agolubev: it will not do endless loop as we have currentLimit. It's refreshes once per actor.receive (line 573)

rkuhn: okay, that happens to make it bounded, but it limits it in a non-intuitive way (and not to the shortCircuitLimit, rather to twice as much)

agolubev: yeah I fixed few issues in "in place" manner - will review and refactor

rkuhn:
I actually did have an idea just now:

* have `GraphInterpreter.execute` return how much is left of the step budget
* enqueue all local “async” events into an unbounded buffer (overflow to mailbox does not help in any case—and there is no back-pressure signal for callbacks)
* after processing external stimulus, draw from the internal callback buffer
    * deliver the async input to one (counting as one step), then run that interpreter with the remaining budget; repeat as long as there are inputs and there is budget
    * if resume is needed, first enqueue a global actor resume signal, then the resume for the specific interpreter (if it was the running part that didn’t complete)
    * the global actor resume will just continue pulling from the callback buffer

This ensures that all callbacks are executed in order as observed by their source, it limits the overall number of steps, and it gets rid of the extra configuration setting.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
agolubev: we have several cases here:
1)  asyncInput targeting for another actor - we use standard actor mechanism
2) asyncInput is for current actor and the limit is not yet reached. We need to enqueue this into shortCircuitBuffer and ran after general handler finished
3) asyncInput is for current actor and it's already in shot circuit loop. Here we can enqueue asyncInput only in case if we haven't reached limit yet. 

My point is that if we are sending AsyncInput to different actors we cannot guarantee order here and shortCircuit does not work for this case. When we only dealing with one actor. We can enqueue messages until limit not reached. Until that moment we supporting message order because we are using shortCircuitBuffer. When limit is reached (once per actor.receive) we know we need to interrupt this thread to let other message come and we are sending all messages via actor mechanism - preserving message order as well.

agolubev:
it's enhancement of current implementation - as you are talking about having global actor resume that continue working with buffer. So less events will go in standard way and more with internal callback buffer

agolubev:
you are talking about "unbounded buffer" - do you have example somewhere in codebase? I mean it must be fast enough and perhaps FixedSizeBuffer is good to use - still it's bounded.

rkuhn:
There is no bound on how many callbacks are enqueued by any given GraphStage—even though we don’t have bad examples in the codebase users can implement a 1000x fanout themselves if they want.

agolubev:
Ok, will use linked list like structure than.

akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
agolubev: @rkuhn, should it be the last parameter? not sure I understand you point

agolubev:
Question:

>have GraphInterpreter.execute return how much is left of the step budget

We are not relaying on separate variable: shortCircuitLimit but should use currentLimit that belongs to shell/interpreter. (As for on actor we can have several shells/interpreters we'll have several limits)

In other words we are iterating over internal callback buffer until one of interpreters says that there is no budget, correct? 
(In case there is still part that did not complete we'll send resume for this interpreter)

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
rkuhn: While processing the inputs from the list it is random chance whether any given callback will be enqueued or not—in particular it is possible that after having sent one callback via the ActorRef the next one will go to the short circuit. This leads to a reordering or callbacks coming from within this same actor. There is no problem with external callbacks since their ordering guarantees are governed by the ActorRef semantics.

akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
rkuhn: After switching on MiMa you’ll see that the old method signature must be kept, you cannot modify it; the only possibility is to add another method overload, but we might just not do that and rely on reading from config + withXXX methods for new properties.

But this point is going to become moot once the event limit is reused to control the fast path as well.

rkuhn:
No, I am arguing that the interpreter limit should be lifted up into the ActorGraphInterpreter and shared by all interpreters that run within it. Otherwise the number of steps executed before polling the mailbox becomes arbitrarily large depending on the dynamic graph topology.

Let’s say limit=10 and we get onNext for one of the interpreters. During event processing an async callback is enqueued for another local interpreter. The first interpreter is finished after having done 7 steps. Passing the async input into the other interpreter is step number 8. Then running that other interpreter will have to be done with a limit of 2, because that is all that is left. If the interpreter needs only 1 step, then there is still 1 step left for delivering another short circuit callback (if there was one).

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
agolubev: > While processing the inputs from the list it is random chance whether any given callback will be enqueued or not—in particular it is possible that after having sent one callback via the ActorRef the next one will go to the short circuit.

It will go short circuit based on two conditions:
1) AsyncInput is for shell that locates in the same ActorGraphInterpreter (same actor)
2) limit of AsyncInput in buffer (per ActorGraphInterpreter.received) is not reached

still there is a chance that buffer was full in one moment of time so this AsyncInput went via ActorRef and next go via short circuit. This can be fixed with some flag or having unbounded queue for short circuit

agolubev:
Got it. I will add this change.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
drewhk: > This won’t work: we cannot execute async callbacks out of order, that would be tremendously confusing for people (and for our own code as well).

Yeah, I actually thought that it is of course possible to reorder since we don't guarantee any ordering, and then realized that in a lot of places we rely on ordering. Fixing that would be actually more costly than simply enforcing it. So let's add a test that checks ordering, and also document this guarantee.

akka-stream/src/main/resources/reference.conf
drewhk: I thought that this is the *overall* processing limit but I was mistaken.

drewhk:
I guess unbounded buffer is a way to go, maybe optimizing it to be a resizable actor based one in the future to reduce allocations.

agolubev:
I'm trying to find out good rule for setting limit to  ActorGraphInterpreter interpreter limit.

How it was in PR:
Initially it locates in GraphInterpreter and calculated as: ```eventLimit = settings.maxInputBufferSize * (assembly.ins.length + assembly.outs.length)``` so it taking into account nature of interpreter.
In Pull request I introduced syncProcessingLimit that works as factor to eventLimit. So maximum number of steps per one ```receive``` call will be syncProcessingLimit*eventLimit 

How it should be:
Theoretically ActorGraphInterpreter interpreter limit should depends on number of ins and outs for initial shell and all other shells. I mean maybe need  to implement similar math that described above but with single ActorGraphInterpreter limit?

drewhk:
Basically there are two kinds of limits that are at play here:
 - the limit of events the *actor* processes (maximum) until it suspends itself
 - the limit of events an *interpreter* processes (maximum) until it gives other interpreters a chance to react on cross-interpreter async events

I believe that the second limit should be calculated from the first. As for the first limit, now I am not sure if sizing it according to the graph is a good idea. For large graphs we need an upper cap anyway, and for smaller graphs the suspend comes naturally by simply needing new asynchronous input from the external world eventually. So my proposal is:

 - have a global limit per actor, a configurable constant, shared by all interpreters in that actor (maybe this should be put on the Shell instead of the actor?)
 - have limits per interpreter (probably dividing the global limit per active interpreters, or something like that). If there is only one interpreter that can run then this limit should come out as exactly the same as the global limit

rkuhn:
Counter-proposal: have the per-interpreter limit be equal to the per-actor limit. Everything is triggered by external stimulus, which will guarantee fairness between interpreters. Another way of looking at this is that a short-circuit callback is counted towards the limit of the submitting interpreter—or just borrowing what that one had left of its own share after being finished.

drewhk:
> Another way of looking at this is that a short-circuit callback is counted towards the limit of the submitting interpreter—or just borrowing what that one had left of its own share after being finished.

That is also how I view it, but I imagined the interpreter still "yielding" to other ones after N steps. If after yielding it happens to be still the only executable one, it continues until it finally exhausts the per-actor limit.

drewhk:
@rkuhn just realized, what you propose is prone to starvation (or I misunderstood it, but then you can clarify). Imagine interpreter A with a stream-of-stream stage and interpreter B hosting a substream of that stage. When B is in fully pulled position, it is not waiting on an *external* stimulus, since now the async callback is an *internal* stimulus. At this point, if interpreter A hosts a detached op and a long enough ping-pong like setup (for example a mapConcat before a conflate with enough elements emitted from mapConcat) that means it can:
 - always process an external stimulus
 - always exhaust the budget while processing that stimulus
and then interpreter B can potentially delayed for arbitrary long time since it does not wait on external stimulus itself, and external stimuli of A does not give a chance to B to progress. While the above setup is not too likely in practice, it makes me a bit uneasy.

agolubev:
build failed with ClusterShardingGetStatsSpec failed. I think this is unrelated. Still locally this spec passed/

agolubev:
ActorGraphInterpreter event limit is implemented. Now all shells of one actor shares this limit and actor interpreter interrupts only when there is no event in short circuit or quota is exceeded. Substream initialization is also a short circuit event. So we end up with - we can guarantee order of events in case events are firing for the same actor by different shells. All such events are adding to short circuit queue. External events and resume events still using actor queue.

Also perhaps need to add additional limit to shell interpreter (to interrupt not only actor but also shell loop). So need to interpreter.execute with limit as parameter - limit can be actor event limit or shell.eventLimit ```shell.eventLimit = settings.maxInputBufferSize * (assembly.ins.length + assembly.outs.length)``` whichever is smaller.

@drewhk let me know if I understand you suggestion correctly.

Still need some help with:
* EventLimit for all shells under akka is now hardcoded to 200. There is no HTTP benchmark to play with this number to find something comparatively small but with no big performance gain when increasing this limit. Could you point out to test or doc snippet of code that can be transferred to benchmark?
* Perhaps ActorMaterializer need some refactoring to allow adding new settings without API changes. I added event limit with no existing code influence still approach maybe need some redesign

agolubev:
Ok, tests passed so PR is ready for review.

akka-stream/src/main/scala/akka/stream/ActorMaterializer.scala
```diff
@@ -284,6 +284,7 @@ final class ActorMaterializerSettings(
   val autoFusing: Boolean,
   val maxFixedBufferSize: Int) {
 
+  var syncProcessingLimit = 200 //defailt value
```
rkuhn: Settings must be immutable, please add this limit in the same fashion as all the others. Binary compatibility is achieved by adding a secondary constructor that sets this limit to 200.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -333,17 +342,16 @@ private[stream] final class GraphInterpreterShell(
    * to give each input buffer slot a chance to run through the whole pipeline
    * and back (for the demand).
    */
-  private val eventLimit = settings.maxInputBufferSize * (assembly.ins.length + assembly.outs.length)
-
+  val shellEventLimit = settings.maxInputBufferSize * (assembly.ins.length + assembly.outs.length)
```
rkuhn: Just thought about this—I had the feeling that it should be “at least N” with N > 1, but then I couldn’t back this up with facts. Now I think I have found an example of why this is the upper limit that we can choose, and if this is correct then it should be added to the explaining comment:

* assume a composite Sink of one `expand` and one `fold`
* assume an infinitely fast source of data
* assume maxInputBufferSize == 1 (i.e. the user when the whole nine yards to get more fairness)
* if the event limit is greater than what we calculate here then there will always be `expand` activity because no data can enter “fast enough” from the outside

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -333,17 +342,16 @@ private[stream] final class GraphInterpreterShell(
...
   // Limits the number of events processed by the interpreter on an abort event.
   // TODO: Better heuristic here
-  private val abortLimit = eventLimit * 2
-  private var resumeScheduled = false
+  private val abortLimit = shellEventLimit * 2
+  var resumeScheduled = false
```
rkuhn: please leave the `var` private, add a public accessor if needed

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -449,13 +465,13 @@ private[stream] final class GraphInterpreterShell(
             override def run(): Unit = self ! Abort(GraphInterpreterShell.this)
           })
         }
-      } else if (interpreter.isSuspended && !resumeScheduled) {
-        resumeScheduled = true
-        self ! resume
-      }
+      } else if (interpreter.isSuspended && !resumeScheduled) sendResume(!usingShellLimit)
```
rkuhn: Very good, I hadn’t even thought of handling it this way!

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -511,9 +530,23 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
         false
     }
 
+  private def adjustEventLimit(shell: GraphInterpreterShell): Unit =
+    if (effectiveEventLimit < shell.shellEventLimit) effectiveEventLimit = shell.shellEventLimit
```
rkuhn: This looks wrong: if a single shell in there has a limit of 1, then the whole actor will use this limit.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -511,9 +530,23 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
...
+
+  //this limits number of messages that can be processed synchronously during one actor receive.
+  private val eventLimit: Int = _initial.mat.settings.syncProcessingLimit
+  private var effectiveEventLimit: Int = Math.max(_initial.shellEventLimit, eventLimit)
+  private var currentLimit: Int = effectiveEventLimit
+  private var shortCircuitBuffer: util.LinkedList[Any] = null
```
rkuhn: perhaps add a comment that this is a `var` in order to save the allocation when no short-circuiting actually happens

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -537,35 +570,60 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
...
+  private def processEvent(b: BoundaryEvent): Unit = {
+    val shell = b.shell
+    if (!shell.isTerminated && (shell.isInitialized || tryInit(shell))) {
+      currentLimit = shell.receive(b, currentLimit)
+      if (shell.isTerminated) {
+        activeInterpreters -= shell
+        if (activeInterpreters.isEmpty && newShells.isEmpty) context.stop(self)
+      }
+    }
+    resumeScheduled = shell.resumeScheduled
```
rkuhn: I think this line should be removed, see below.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -537,35 +570,60 @@ private[stream] class ActorGraphInterpreter(_initial: GraphInterpreterShell) ext
...
-      val shell = b.shell
-      if (!shell.isTerminated && (shell.isInitialized || tryInit(shell))) {
-        shell.receive(b)
-        if (shell.isTerminated) {
-          activeInterpreters -= shell
-          if (activeInterpreters.isEmpty && newShells.isEmpty) context.stop(self)
-        }
-      }
-    case Resume ⇒ finishShellRegistration()
+      resumeScheduled = false
```
rkuhn: This line should be removed because it is not a `Resume` that we have received here. Every shell keeps track of having scheduled its own resumption and the Actor only needs to keep track of having `Resume` in its mailbox whenever the short-circuit queue is non-empty. This means that it is simplest (to understand, document and maintain) to only set to `true` when actually sending the message, and to only set to `false` when actually receiving the message.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
rkuhn: Ah, I see now: it is used as an identifier for the context in which this is running.

rkuhn:
Great work, @agolubev ! This is tricky stuff and needs time to sink in and review, and I think it is only an inch away from being merged.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
agolubev: the idea here is that  in case event limit (that we define as constant in configuration) is less than shell  limit we expanding limit to what's for shell. In this case we avoiding additional Resume messages.

We can have limit as 1 only in case we set limit constant for actor in 1 and event limit for all shells is 1.

rkuhn: Ah, sorry for reading the comparison the wrong way around. But still, I think that the configuration file setting should cap the total limit because this is the only weapon we have against unfairness: a huge fused graph would get very long stretches of CPU that cannot be interrupted. 200 events correspond to ca. 3µs on typical hardware today, high throughput applications may want to set it higher (and then also configure the Actor Dispatcher’s `throughput` setting which defaults to `5` messages only).

The HTTP server blueprint has roughly 50 ports, which at the default inputBuffer settings yields 800 events—maybe we set the default limit to 1000 and remove this adaptation logic?

agolubev: >maybe we set the default limit to 1000 and remove this adaptation logic? 

will change it to 1000 than 
BTW, thanks for specific numbers.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
agolubev: Idea here was to add logic to not send additional Resume message (to continue processing short circuit queue if limit is 0 but there are some events in short circuit queue) when shell already sent Resume. 

I checked that it works without this logic - so will remove it

agolubev:
Fixed everything that was in feedback. Ready for Review

agolubev:
@rkuhn, please check when you have some time.

akka-stream/src/main/scala/akka/stream/impl/fusing/ActorGraphInterpreter.scala
```diff
@@ -57,6 +57,7 @@ private[stream] object ActorGraphInterpreter {
 
   final case class AsyncInput(shell: GraphInterpreterShell, logic: GraphStageLogic, evt: Any, handler: (Any) ⇒ Unit) extends BoundaryEvent
 
+  case class Init(shell: GraphInterpreterShell) extends BoundaryEvent
```
rkuhn: this doesn’t seem to be used

rkuhn:
Apart from that tiny nitpick we are now only a squash and push away from merging—thanks a lot!

agolubev:
Just rebased and squashed. Thanks, Roland.

agolubev:
build failed with
```
[info] Passed: Total 725, Failed 0, Errors 0, Passed 725, Ignored 2, Pending 11

java.lang.RuntimeException: akka-stream: Binary compatibility check failed!
	at scala.sys.package$.error(package.scala:27)
	at com.typesafe.tools.mima.plugin.SbtMima$$anonfun$reportErrors$1.apply(SbtMima.scala:96)
	at com.typesafe.tools.mima.plugin.SbtMima$$anonfun$reportErrors$1.apply(SbtMima.scala:78)
	at scala.collection.immutable.Map$Map1.foreach(Map.scala:109)
...
[error] (akka-stream/*:mimaReportBinaryIssues) akka-stream: Binary compatibility check failed!
[error] Total time: 1397 s, completed Mar 14, 2016 5:17:52 PM
```
not sure what API class was changed in incorrect way

ktoso:
It's in the log, search for `"incompatibilities"`:
https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/5497/consoleFull

```
[info] akka-stream: found 4 potential binary incompatibilities while checking against com.typesafe.akka:akka-stream_2.11:2.4.2  (filtered 15)
[error]  * method execute(Int)Unit in class akka.stream.impl.fusing.GraphInterpreter has a different result type in current version, where it is Int rather than Unit
[error]    filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("akka.stream.impl.fusing.GraphInterpreter.execute")
[error]  * method this(akka.stream.impl.fusing.GraphInterpreter#GraphAssembly,akka.stream.Materializer,akka.event.LoggingAdapter,Array[akka.stream.stage.InHandler],Array[akka.stream.stage.OutHandler],Array[akka.stream.stage.GraphStageLogic],scala.Function3,Boolean)Unit in class akka.stream.impl.fusing.GraphInterpreter does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.impl.fusing.GraphInterpreter.this")
[error]  * method init(akka.actor.ActorRef,akka.stream.impl.SubFusingActorMaterializerImpl)Unit in class akka.stream.impl.fusing.GraphInterpreterShell does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.impl.fusing.GraphInterpreterShell.init")
[error]  * method receive(akka.stream.impl.fusing.ActorGraphInterpreter#BoundaryEvent)Unit in class akka.stream.impl.fusing.GraphInterpreterShell does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.impl.fusing.GraphInterpreterShell.receive")
```

Need to decide if safe to ignore and add the exclusions

agolubev:
Thanks

agolubev:
OK so there methods for two classes in incompatibilities log:
* GraphInterpreterShell is private[stream] final 
* GraphInterpreter is private[stream] final 
and seems like both are not part of public API. 
Let me know if I just need to add exclusions

rkuhn:
Yes, these classes are internal, so the BC exceptions can be added.

agolubev:
Great! Test passed.

rkuhn:
Woohoo! :-)

PR #19703, opened 2016-02-07
add maxFixedBufferSize materializer setting #19702

### LINKED ISSUE #19702: choose between BoundedBuffer and FixedSizeBuffer consistently
introduce a configuration setting that decides beyond which threshold BoundedBuffer shall be used

rkuhn:
fixes #19702


akka-stream/src/main/resources/reference.conf
```diff
@@ -45,6 +45,16 @@ akka {
...
+      
+      # Those stream elements which have explicit buffers (like mapAsync, mapAsyncUnordered,
+      # buffer, flatMapMerge, Source.actorRef, Source.queue, etc.) will preallocate a fixed
+      # buffer upon stream materialization if the requested buffer size is less than this
+      # configuration parameter. The default is very high because failing early is better
+      # than failing under load.
+      #
+      # Buffers larger than this value will use a dynamic approach that has a higher
+      # per-element cost.
+      max-fixed-buffer-size = 1000000000
```
viktorklang: No test to verify that the feature works?

rkuhn: I take that to imply that the feature is good? (there was no discussion yet, so I just threw the proposal into the ring)

viktorklang: What's the effect on performance going from monomorphic to bimorphic given
the call site pollution the interpreter causes?

-- 
Cheers,
√

rkuhn: There would only be an effect if the default is changed to a low value and then some buffers are Bounded while others are Fixed. And BoundedBuffer already becomes bi-morphic if its fixed part overflows and it enters dynamic mode.

viktorklang: Fair enough. You might want to loosen the wording on "Buffers larger than this value will use a dynamic approach that has a higher per-element cost." to something like: "Buffers sized larger than this will dynamically grow/shrink and may consume more memory—per element—than the fixed size buffers."

akka-stream/src/main/scala/akka/stream/impl/Buffers.scala
```diff
@@ -0,0 +1,204 @@
...
+        dequeue()
+      }
+    override def dropHead(): Unit = dequeue()
+    override def dropTail(): Unit = {
+      tail -= 1
+      queue(tail & FixedQueueMask) = null
+    }
+  }
+
+  private final class DynamicQueue(startIdx: Int) extends ju.LinkedList[T] with Buffer[T] {
```
drewhk: I don't really like this implementation. This means that for large buffers we will suddenly put huge pressure on the heap and pay tons of allocation costs.

What I would like to see instead:
 - a top level array pointing to sub-arrays of size let's say 128, 128, 256, 512, ... up to some maximum capacity. On need of growth a new top-level array should be allocated and cirularity restored. On falling below a threshold, buffers should be dropped and circularity restored.

In the above scheme we can set the threshold to much lower and still have well behaved heap and allocation behavior, paying mostly for used buffer space (up to 2x), having better cache friendliness etc.

drewhk:
I like the idea in general, but I don't linke the raw LinkedList based implementation. Maybe a benchmarking round should be preferred before we do this or any other change.

rkuhn:
@drewhk Thanks for your comments, but that is not actually the changed part: the only change is when to use which of the pre-existing solutions (and consolidating them in one source file).

rkuhn:
added tests and improved wording

rkuhn:
failures are unrelated and known

Issue #19704, opened 2016-02-07
UnboundedControlAwareMessageQueueSemantics not working


iouwon:
```Scala
object UnboundedControlAwareMailbox {
  class MessageQueue extends ControlAwareMessageQueueSemantics with UnboundedMessageQueueSemantics {
    val controlQueue: Queue[Envelope] = new ConcurrentLinkedQueue[Envelope]()
    val queue: Queue[Envelope] = new ConcurrentLinkedQueue[Envelope]()
  }
}
```
probably should be

```Scala
object UnboundedControlAwareMailbox {
  class MessageQueue extends UnboundedControlAwareMessageQueueSemantics with UnboundedMessageQueueSemantics {
    val controlQueue: Queue[Envelope] = new ConcurrentLinkedQueue[Envelope]()
    val queue: Queue[Envelope] = new ConcurrentLinkedQueue[Envelope]()
  }
}
```
without it trying to do:
```Scala
extends RequiresMessageQueue[UnboundedControlAwareMessageQueueSemantics]
```
fails with:

>Caused by: java.lang.IllegalArgumentException: produced message queue type [class akka.dispatch.UnboundedControlAwareMailbox$MessageQueue] does not fulfill requirement for actor class [class sap.out.order.SapOrderEndpointConsumer]. Must be a subclass of [interface akka.dispatch.UnboundedControlAwareMessageQueueSemantics].

However, doing extends `RequiresMessageQueue[ControlAwareMessageQueueSemantics]` does work but is not quite right I feel

virsox:
Opened a Pull Request for this issue https://github.com/akka/akka/pull/19705

PR #19705, opened 2016-02-08
+act #19704 Fix UnboundedControlAwareMessageQueueSemantics verification

virsox:
I hope I've followed all the contributions guidelines correctly.

### DIFF
--- a/akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
+++ b/akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
@@ -842,7 +842,7 @@ final case class UnboundedControlAwareMailbox() extends MailboxType with Produce
 }
 
 object UnboundedControlAwareMailbox {
-  class MessageQueue extends ControlAwareMessageQueueSemantics with UnboundedMessageQueueSemantics {
+  class MessageQueue extends UnboundedControlAwareMessageQueueSemantics with java.io.Serializable {
     val controlQueue: Queue[Envelope] = new ConcurrentLinkedQueue[Envelope]()
     val queue: Queue[Envelope] = new ConcurrentLinkedQueue[Envelope]()
   }
@@ -862,7 +862,7 @@ final case class BoundedControlAwareMailbox(capacity: Int, override final val pu
 }
 
 object BoundedControlAwareMailbox {
-  class MessageQueue(val capacity: Int, val pushTimeOut: FiniteDuration) extends BoundedControlAwareMessageQueueSemantics {
+  class MessageQueue(val capacity: Int, val pushTimeOut: FiniteDuration) extends BoundedControlAwareMessageQueueSemantics with java.io.Serializable {
 
     private final val size = new AtomicInteger(0)
     private final val putLock = new ReentrantLock()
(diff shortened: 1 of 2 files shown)


drewhk:
OK TO TEST

drewhk:
refs #19704

drewhk:
I'm afraid this won't be binary compatible though.

virsox:
Sorry, any chance of this being cause by another PR? It seems https://github.com/akka/akka/pull/19564 added some new tests that are failing, but I couldn't connect with this PR

johanandren:
Yes, the failure was unrelated to this PR

johanandren:
PLS BUILD

akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
```diff
@@ -842,7 +842,7 @@ final case class UnboundedControlAwareMailbox() extends MailboxType with Produce
 }
 
 object UnboundedControlAwareMailbox {
-  class MessageQueue extends ControlAwareMessageQueueSemantics with UnboundedMessageQueueSemantics {
+  class MessageQueue extends UnboundedControlAwareMessageQueueSemantics with java.io.Serializable {
```
ktoso: Why making it extend Serializable?

akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
```diff
@@ -862,7 +862,7 @@ final case class BoundedControlAwareMailbox(capacity: Int, override final val pu
 }
 
 object BoundedControlAwareMailbox {
-  class MessageQueue(val capacity: Int, val pushTimeOut: FiniteDuration) extends BoundedControlAwareMessageQueueSemantics {
+  class MessageQueue(val capacity: Int, val pushTimeOut: FiniteDuration) extends BoundedControlAwareMessageQueueSemantics with java.io.Serializable {
```
viktorklang: I assume @ktoso has the same Serializable question here

rkuhn:
It seems to be binary compatible, but I, too, wonder about the Serializable extension.

drewhk:
Isn't it only reported as compatible as these things are wired dynamically?

rkuhn:
Which method call do you expect to fail with a LinkageError after these changes?

drewhk:
One class (MessageQueue) now extends a different superclass. Isn't that a problem? I don't know which method since I have not much recollection of the mailbox selection API anymore.

rkuhn:
The additional type that was inserted into the hierarchy does not add any methods, all type tests will still come back with the same result apart from adding the one `true` that was missing. If there was a method change that causes binary incompatibility then MiMa would flag it.

drewhk:
ok

akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
virsox: The tests send `MessageQueue` as messages, so they fail if the `MessageQueue` is not Serializable. It felt odd to me, but as far as I could track all other `MessageQueues` are also Serializable (by inheritance, e.g. `BoundedDequeBasedMailbox.MessageQueue` extends `LinkedBlockingDeque`, which is Serializable). Also, it makes sense if we consider queues store Messages, which are Serializable as well.

Therefore, for consistency sake, I decided to make `BoundedControlAwareMailbox.MessageQueue` Serializable. However, I must say that I haven't thought about binary compatibility here. If needed, I can remove the extends and create some other tests for this functionality

rkuhn: Hmm, interesting. I guess this is at least consistent then.

drewhk:
So this is basically LGTM. Any objections to merge it?

rkuhn:
LGTM, thanks @virsox !

PR #19718, opened 2016-02-08
Add Typesafe copyright header

naferx:


ktoso:
For new files we tend to add just `2016` but this is a good improvement too, thanks!

naferx:
Ok

PR #19724, opened 2016-02-09
=doc hide auto downing from default sample config in cluster

ktoso:
I checked how it looks in generated docs – looks OK.

### DIFF
--- a/akka-samples/akka-sample-cluster-java/src/main/resources/application.conf
+++ b/akka-samples/akka-sample-cluster-java/src/main/resources/application.conf
@@ -1,3 +1,4 @@
+#//#snippet
 akka {
   actor {
     provider = "akka.cluster.ClusterActorRefProvider"
@@ -15,7 +16,14 @@ akka {
       "akka.tcp://ClusterSystem@127.0.0.1:2551",
       "akka.tcp://ClusterSystem@127.0.0.1:2552"]
 
+    #//#snippet
+    # excluded from snippet
     auto-down-unreachable-after = 10s
+    #//#snippet
+    # auto downing is NOT safe for production deployments.
+    # you may want to use it during development, read more about it in the docs.
+    #
+    # auto-down-unreachable-after = 10s
   }
 }
 
@@ -28,3 +36,4 @@ akka.extensions=["akka.cluster.metrics.ClusterMetricsExtension"]
 # Sigar native library extract location during tests.
 # Note: use per-jvm-instance folder when running multiple jvm on one host. 
 akka.cluster.metrics.native-library-extract-folder=${user.dir}/target/native
+#//#snippet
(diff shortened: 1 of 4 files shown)


akka-samples/akka-sample-cluster-java/src/main/resources/application.conf
```diff
@@ -15,7 +16,14 @@ akka {
       "akka.tcp://ClusterSystem@127.0.0.1:2551",
       "akka.tcp://ClusterSystem@127.0.0.1:2552"]
 
+    //#snippet
+    // excluded from snippet
     auto-down-unreachable-after = 10s
+    //#snippet
+    # auto downing is NOT safe for production deployments.
+    # you may want to use it during development, read more about it in the docs.
```
johanandren: :+1: :+1: :+1:

johanandren:
LGTM

akka-samples/akka-sample-cluster-java/src/main/resources/application.conf
```diff
@@ -1,3 +1,4 @@
+//#snippet
```
patriknw: #//#

akka-samples/akka-sample-cluster-java/src/main/resources/application.conf
```diff
@@ -15,7 +16,14 @@ akka {
       "akka.tcp://ClusterSystem@127.0.0.1:2551",
       "akka.tcp://ClusterSystem@127.0.0.1:2552"]
 
+    //#snippet
+    // excluded from snippet
```
patriknw: that is not a comment

akka-samples/akka-sample-cluster-java/src/main/resources/application.conf
patriknw: :+1:

akka-samples/akka-sample-cluster-java/src/main/resources/application.conf
ktoso: It is actually, but I'll better stick to the `#` syntax here indeed :) 
See: https://github.com/typesafehub/config#features-of-hocon

akka-samples/akka-sample-cluster-java/src/main/resources/application.conf
ktoso: ok, will do

ktoso:
Addressed comments about comment style

patriknw:
LGTM (it was not about style, it was syntax error)

ktoso:
Pretty sure `//` works in typesafe config, see here https://github.com/akka/akka/pull/19724#discussion_r52359064 what was I missing?

patriknw:
I didn't know that. Then I'm sorry for the noise.

ktoso:
No problem at all `#` style is more consistent with what we do everywhere anyway so thanks for noticing!

Issue #19645, opened 2016-01-29
FAILED: BackoffSupervisorSpec


2m:
```
[info] - must support custom supervision strategy *** FAILED *** (3 seconds, 220 milliseconds)
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class akka.pattern.BackoffSupervisor$CurrentChild
[info]   at scala.Predef$.assert(Predef.scala:165)
[info]   at akka.testkit.TestKitBase$class.expectMsgClass_internal(TestKit.scala:435)
[info]   at akka.testkit.TestKitBase$class.expectMsgType(TestKit.scala:408)
[info]   at akka.testkit.TestKit.expectMsgType(TestKit.scala:735)
[info]   at akka.pattern.BackoffSupervisorSpec$$anonfun$1$$anonfun$apply$mcV$sp$4$$anonfun$akka$pattern$BackoffSupervisorSpec$$anonfun$$anonfun$$assertCustomStrategy$1$1.apply$mcV$sp(BackoffSupervisorSpec.scala:85)
```

(Jan 29, 2016 3:17:00 AM) https://jenkins.akka.io:8498/job/akka-nightly/3007/consoleFull
(Jan 28, 2016 7:18:00 AM) https://jenkins.akka.io:8498/job/akka-multi-node-nightly2/246/consoleFull

2m:
This is after a fix attempt: https://github.com/akka/akka/pull/19579

patriknw:
https://jenkins.akka.io:8498/job/akka-nightly/3016/consoleFull

drewhk:
Hm, trying this again.

drewhk:
Well, strictly speaking one of the issues *were* fixed, this failure is a second kind (the previous ticket mixed the two).

patriknw:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/3449/consoleFull

drewhk:
This is totally weird, it looks like the "supervisor" actor somehow does not get the GetCurrentChild message, which is weird since it got it not much before for another send. I.e. the GetCurrentChild is lost somehow and hence no reply comes. Now I need to figure out how this can happen, but so far this is rather hairy.

drewhk:
Argh, was looking at the wrong actor.

PR #19726, opened 2016-02-09
#19645: BackoffOnRestartSupervisor should not ignore commands while child is terminating

drewhk:


akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
```diff
@@ -39,20 +39,37 @@ private class BackoffOnRestartSupervisor(
         // directive match {
```
johanandren: Leftover commented code?

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
```diff
@@ -39,20 +39,37 @@ private class BackoffOnRestartSupervisor(
...
-          become({
-            case Terminated(`childRef`) ⇒
-              unbecome()
-              child = None
-              val restartDelay = BackoffSupervisor.calculateDelay(restartCount, minBackoff, maxBackoff, randomFactor)
-              context.system.scheduler.scheduleOnce(restartDelay, self, BackoffSupervisor.StartChild)
-              restartCount += 1
-            case _ ⇒ // ignore
-          }, discardOld = false)
+          become(waitChildTerminatedBeforeBackoff(childRef) orElse handleBackoff)
```
johanandren: :+1:

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
drewhk: seems like

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
```diff
@@ -39,20 +39,37 @@ private class BackoffOnRestartSupervisor(
...
+  def waitChildTerminatedBeforeBackoff(childRef: ActorRef): Receive = {
+    case Terminated(`childRef`) ⇒
+      become(receive)
+      child = None
+      val restartDelay = BackoffSupervisor.calculateDelay(restartCount, minBackoff, maxBackoff, randomFactor)
+      context.system.scheduler.scheduleOnce(restartDelay, self, BackoffSupervisor.StartChild)
+      restartCount += 1
+
+    case StartChild ⇒
+      become(stashedStartChild(childRef) orElse handleBackoff)
```
johanandren: When would we get this `StartChild`? When it terminates above, it goes into `receive` before scheduling it.

drewhk: It can be an external command for manual starting as far as I understand. But I can be wrong.

johanandren: Instead of explaining it to me, a comment explaining it would be nice.

drewhk: I don't know, maybe it is not meant to be used externally :)

johanandren: I think it is `private[akka]`

drewhk: Then this might not be needed at all.

drewhk:
Failure is the known conflatespec

johanandren:
PLS BUILD

ktoso:
Refs #19645

akka-actor/src/main/scala/akka/pattern/BackoffOnRestartSupervisor.scala
patriknw: do I understand it correctly that the fix is the `orElse handleBackoff`?

drewhk: yes, and also explicitly ignoring the StartChild during this period. The rest of it are external management messages.

drewhk:
Failures unrelated.

johanandren:
LGTM!

patriknw:
LGTM

PR #19736, opened 2016-02-10
=str - removing superfluous try-catch in stream-customize.rst

viktorklang:
@drewhk ping

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/dungeon/DeathWatch.scala
+++ b/akka-actor/src/main/scala/akka/actor/dungeon/DeathWatch.scala
@@ -4,9 +4,9 @@
 
 package akka.actor.dungeon
 
-import akka.dispatch.sysmsg.{Unwatch, Watch, DeathWatchNotification}
-import akka.event.Logging.{Warning, Debug}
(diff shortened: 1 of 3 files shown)


drewhk:
LGTM, thanks for noticing!

ktoso:
LGTM, good catch

viktorklang:
The Eye of Klang

viktorklang:
merge?

drewhk:
only after validation

viktorklang:
Is it running all tests to check a doc class? #wtf ?

ktoso:
We have smart validation, will run only docs and projects which depend on it.

viktorklang:
Ow, Y SO SLO? :(

viktorklang:
Ah, so Function declares that it @throws Exception but onPush/onPull and friends on AbstractInHandler/AbstractOutHandler do not declare that they @throws Exception.

viktorklang:
Question is if that should be done or not, @drewhk ?

drewhk:
I think they can be safely decorated with `throws Exception`. It is the framework anyway that needs to deal with any exceptions, but we are the framework, and we happily take the Exception :)

Issue #19737, opened 2016-02-10
UpgradeToWebSocket  not found in Akka 2.4.2-RC2


jftregor:
Hello,

I'm currently working on a project using  Akka 2.4.1 along with akka-http and akka-stream 2.0.3.
This morning, I tried **Akka 2.4.2-RC2**.

This was _almost_ a success :-)

However, I ran into a strange error as the compilation failed like so:
```
[error] /.../src/main/scala/.../server/akkahttp/route/WebSocketRoute.scala:48: not found: type UpgradeToWebsocket
[error]          optionalHeaderValueByType[UpgradeToWebsocket]() {
[error]                                    ^
```
So I tried to add the specific import
```
import akka.http.scaladsl.model.ws._
import akka.http.scaladsl.model.ws.UpgradeToWebSocket
```
but it did not work. Of course, I cleaned the project (several times) and compiled again but the compilation error remains...

The only workaround I found was:
```
// FIXME: following line worked before akka 2.4.2-RC2
// optionalHeaderValueByType[UpgradeToWebsocket]() {
optionalHeaderValueByType[akka.http.scaladsl.model.ws.UpgradeToWebSocket]() {
...
```
BTW, here are the relevant lines from the `build.sbt` file:
```
val akkaVersion = "2.4.2-RC2"

...
libraryDependencies ++= Seq(
    ...		
    	"com.typesafe.akka" %% "akka-stream"             % akkaVersion,
    	"com.typesafe.akka" %% "akka-http-core"          % akkaVersion,
    	"com.typesafe.akka" %% "akka-http-experimental"               % akkaVersion,
    	"com.typesafe.akka" %% "akka-http-spray-json-experimental"    % akkaVersion,
    	"com.typesafe.akka" %% "akka-http-testkit-experimental"       % akkaVersion,
   ...)
```

Am I missing something ?

Regard

ktoso:
`UpgradeToWebSocket` (as such casing is more in line with the WebSocket spec) :wink: 

I recommend using your IDE's find class feature when in trouble like this :)

ktoso:
Actually I'll add a notice of that to the migration guide, thanks for bringing it up!

jftregor:
You're welcome, I can remove my `FIXME` line now :smiley: 

BTW, the same error occurred for _handleWeb**S**ocketMessages_ directive and _isWeb**S**ocketUpgrade_ assertion from http-testkit...

[...]

PR #19739, opened 2016-02-10
Explain WebSocket name change and remove obsolete docs part

ktoso:
Resolves https://github.com/akka/akka/issues/19737


akka-docs/rst/additional/http.rst
```diff
@@ -1,26 +0,0 @@
...
-=========
-
-If you are using Akka Mist (Akka's old HTTP/REST module) with Akka 1.x and wish to upgrade to 2.x
-there is now a port of Akka Mist to Akka 2.x. You can find it `here <https://github.com/thenewmotion/akka-http>`_.
-
-Other Alternatives
-==================
-
-There are a bunch of other alternatives for using Akka with HTTP/REST. You can find some of them
-among the `Community Projects <http://akka.io/community>`_.
```
ktoso: I thought we may want to remove this section having HTTP merged, right?

johanandren:
LGTM

2m:
LGTM

ktoso:
Would like one more eye on the removal of `akka-docs/rst/additional/http.rst` if we're fine with that.

akka-docs/rst/additional/http.rst
johanandren: Is it not linked from somewhere, toc. etc, that now also needs to be updated?

ktoso: good catch, fixed

PR #19742, opened 2016-02-10
=htc #19741 cannot pull twice bug in OutgoingConnectionBlueprint fixed

### LINKED ISSUE #19741: FAILED OutgoingConnectionBlueprint
`HighLevelOutgoingConnectionSpec` consistently fails on my machine, with a 

```
java.lang.IllegalArgumentException: requirement failed: Cannot pull port (ResponsePrep.in) twice
	at scala.Predef$.require(Predef.scala:219)
	at akka.stream.stage.GraphStageLogic.pull(GraphStage.scala:351)
	at akka.http.impl.engine.client.OutgoingConnectionBlueprint$ResponsePrep$$anon$1.onPull(OutgoingConnectionBlueprint.scala:170)
	at akka.stream.impl.fusing.GraphInterpreter.processEvent(GraphInterpreter.scala:600)
...
```

Most likely introduced in #19528

johanandren:
Fixes #19741

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
@@ -69,7 +69,7 @@ private[http] object OutgoingConnectionBlueprint {
     import ParserOutput._
     val responsePrep = Flow[List[ResponseOutput]]
       .mapConcat(conforms)
-      .via(new ResponsePrep(parserSettings))
+      .via(new PrepareResponse(parserSettings))
 
     val core = BidiFlow.fromGraph(GraphDSL.create() { implicit b ⇒
       import GraphDSL.Implicits._
@@ -131,11 +131,19 @@ private[http] object OutgoingConnectionBlueprint {
 
   import ParserOutput._
 
-  private final class ResponsePrep(parserSettings: ParserSettings)
+  /**
+   * This is essentially a three state state machine, it is either 'idle' - waiting for a response to come in
+   * or has seen the start of a response and is waiting for either chunks followed by MessageEnd if chunked
+   * or just MessageEnd in the case of a strict response.
+   *
+   * For chunked responses a new substream into the response entity is opened and data is streamed there instead
+   * of downstream until end of chunks has been reached.
+   */
+  private[client] final class PrepareResponse(parserSettings: ParserSettings)
     extends GraphStage[FlowShape[ResponseOutput, HttpResponse]] {
 
-    private val in = Inlet[ResponseOutput]("ResponsePrep.in")
-    private val out = Outlet[HttpResponse]("ResponsePrep.out")
+    private val in = Inlet[ResponseOutput]("PrepareResponse.in")
+    private val out = Outlet[HttpResponse]("PrepareResponse.out")
 
     val shape = new FlowShape(in, out)
 
@@ -143,10 +151,15 @@ private[http] object OutgoingConnectionBlueprint {
       private var entitySource: SubSourceOutlet[ResponseOutput] = _
       private def entitySubstreamStarted = entitySource ne null
       private def idle = this
+      private var completionDeferred = false
 
       def setIdleHandlers(): Unit = {
-        setHandler(in, idle)
-        setHandler(out, idle)
+        if (completionDeferred) {
+          completeStage()
+        } else {
+          setHandler(in, idle)
+          setHandler(out, idle)
+        }
       }
 
       def onPush(): Unit = grab(in) match {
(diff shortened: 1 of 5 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -169,8 +169,10 @@ private[http] object OutgoingConnectionBlueprint {
```
rkuhn: original line 165 above: the condition will always be true

rkuhn: oh, forget it

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -169,8 +169,10 @@ private[http] object OutgoingConnectionBlueprint {
...
       private lazy val waitForMessageEnd = new InHandler {
         def onPush(): Unit = grab(in) match {
-          case MessageEnd ⇒ setHandler(in, idle)
-          case other      ⇒ throw new IllegalStateException(s"MessageEnd expected but $other received.")
+          case MessageEnd ⇒
+            pull(in)
+            setHandler(in, idle)
+          case other ⇒ throw new IllegalStateException(s"MessageEnd expected but $other received.")
         }
       }
```
rkuhn: case MessageEnd below: here we need to `pull(in)` if we already ignored a downstream pull in the onPull that is set for `out`. This could be done by checking `isAvailable(out)`.

rkuhn:
Your fix LGTM, plus my fix it becomes a full fix :-)

Please also check that the corresponding stage in the server blueprint does not have the same bugs.

johanandren:
Unit test coverage for all the fixed bugs in this stage I could think of (including Rolands addition and one more - deferring close). As well as covering both this stage and the server blueprint PreRequest stage with tests for entity stream cancellation (#18203)

johanandren:
Turned out I had misunderstood how the cancelation of the stream should affect the connection, now according to the ticket referenced above, with extra coverage in the low-low-level unit tests.

I understand that the idea has been more to black box test the entire request or response pipeline, but I think that these two stages are so error prone and trixy that tight unit tests that exactly describe how they are supposed to behave are in place.

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -131,11 +131,19 @@ private[http] object OutgoingConnectionBlueprint {
...
   import ParserOutput._
 
-  private final class ResponsePrep(parserSettings: ParserSettings)
+  /**
+   * This is essentially a three state state machine, it is either 'idle' - waiting for a response to come in
+   * or has seen the start of a response and is waiting for either chunks followed by MessageEnd if chunked
+   * or just MessageEnd in the case of a strict response.
+   *
+   * For chunked responses a new substream into the response entity is opened and data is streamed there instead
+   * of upstream until end of chunks has been reached.
```
rkuhn: `s/upstream/downstream/`

akka-http-core/src/test/scala/akka/http/impl/engine/client/PrepareResponseSpec.scala
```diff
@@ -0,0 +1,243 @@
...
+      inSub.sendNext(messageEnd)
+      entityProbe.expectComplete()
+
+      // and that demand should go downstream
+      // since the chunk end was consumed by the stage
+      inSub.expectRequest(1)
+
+    }
+
+    "not lose demand that comes in while handling strict entity" in {
```
rkuhn: this is a copy of the previous test, i.e. testing the streaming case

akka-http-core/src/test/scala/akka/http/impl/engine/client/PrepareResponseSpec.scala
```diff
@@ -0,0 +1,243 @@
...
+      inSub.sendNext(messageEnd)
+      entityProbe.expectComplete()
+
+      // and that demand should go downstream
+      // since the chunk end was consumed by the stage
+      inSub.expectRequest(1)
+
+    }
+
+    "cancel entity stream when upstream cancels" in {
```
rkuhn: `s/upstream/downstream/`

akka-http-core/src/test/scala/akka/http/impl/engine/client/PrepareResponseSpec.scala
```diff
@@ -0,0 +1,243 @@
...
+      inSub.expectRequest(1)
+      inSub.sendNext(chunk)
+      entityProbe.expectNext()
+
+      // now before entity stream is completed,
+      // upstream cancels
+      responseSub.cancel()
+
+      inSub.expectCancellation()
+      entityProbe.expectComplete()
```
rkuhn: This is not correct: the stage should ferry the remainder of this response through and then shut down.

akka-http-core/src/test/scala/akka/http/impl/engine/client/PrepareResponseSpec.scala
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    "complete input stream when upstream cancels" in {
```
rkuhn: upstream cannot cancel :-)

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -203,9 +219,25 @@ private[http] object OutgoingConnectionBlueprint {
...
+          // we will always pull upon reaching end anyways
+          // so safe to ignore
+        }
+        override def onDownstreamFinish(): Unit = {
+          // if downstream cancels, make sure we also
+          // cancel the entity source
+          if (entitySource != null) entitySource.complete()
+          completeStage()
+        }
+      }
```
rkuhn: This should not be necessary: the onPull behavior is already encoded above and the handling of cancellation can also be folded into the normal `onPull` in the same way.

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -162,23 +170,31 @@ private[http] object OutgoingConnectionBlueprint {
       }
 
       def onPull(): Unit = {
-        if (!entitySubstreamStarted) pull(in)
+        if (!entitySubstreamStarted && !hasBeenPulled(in)) pull(in)
```
rkuhn: I’d fold this change into the waitForMessageEnd handler instead (by making it extend OutHandler as well and ignoring onPull there)

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -203,9 +219,25 @@ private[http] object OutgoingConnectionBlueprint {
         }
```
rkuhn: this is unnecessary, the substream has already signaled cancellation

johanandren: Not sure what this comment refers to, there is no substream for `Strict` messages.

rkuhn: This comment was placed after onUpstreamFailure in the substreamHandler (at least on my UI :-) ).

johanandren: Ah, got it, remember thinking about that yesterday, but I never touched it.

drewhk:
LGTM

Issue #19732, opened 2016-02-10
A more consistent and succinct way to mark async boundaries


johanandren:
`.withAttributes(Attributes.asyncBoundary)` is verbose given that it is something one would expect to do pretty often. I propose that we introduce a shortcut `{Source,Flow,Sink}.async` just like `.named()` which exists already and remove `Flow.viaAsync(flow)` which will then instead be `Flow.via(flow.async)`

johanandren:
Discussion on gitter: https://gitter.im/akka/dev?at=56bb0e980fda6e254372e251

hepin1989:
the elixir way😃

Issue #19303, opened 2015-12-28
Update session stream parallelism in akka streams docs


matheuslima:
http://doc.akka.io/docs/akka-stream-and-http-experimental/2.0.1/scala/stream-parallelism.html
"Akka Streams processing stages (be it simple operators on Flows and Sources or graph junctions) are executed concurrently by default. This is realized by mapping each of the processing stages to a dedicated actor internally." After fusing stages implementation this concept is out of date.

ktoso:
Must be revised before merging to master.

rkuhn:
hasn’t this been done by @johanandren ?

johanandren:
Duplicated and fixed by #20051

PR #19743, opened 2016-02-10
!str #19732 Concise and consistent way to mark async boundaries

johanandren:
Refs #19732

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/Graph.scala
+++ b/akka-stream/src/main/scala/akka/stream/Graph.scala
@@ -26,5 +26,10 @@ trait Graph[+S <: Shape, +M] {
 
   def named(name: String): Graph[S, M] = withAttributes(Attributes.name(name))
 
+  /**
+   * Put an asynchronous boundary around this `Graph`
+   */
+  def async: Graph[S, M] = addAttributes(Attributes.asyncBoundary)
+
   def addAttributes(attr: Attributes): Graph[S, M] = withAttributes(module.attributes and attr)
 }
--- a/akka-stream/src/main/scala/akka/stream/impl/SubFlowImpl.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/SubFlowImpl.scala
@@ -34,6 +34,8 @@ class SubFlowImpl[In, Out, Mat, F[+_], C](val subFlow: Flow[In, Out, NotUsed],
   override def named(name: String): SubFlow[Out, Mat, F, C] =
     new SubFlowImpl[In, Out, Mat, F, C](subFlow.named(name), mergeBackFunction, finishFunction)
 
+  override def async: Repr[Out] = new SubFlowImpl[In, Out, Mat, F, C](subFlow.async, mergeBackFunction, finishFunction)
+
   override def mergeSubstreamsWithParallelism(breadth: Int): F[Out] = mergeBackFunction(subFlow, breadth)
 
   def to[M](sink: Graph[SinkShape[Out], M]): C = finishFunction(subFlow.to(sink))
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
@@ -1007,14 +1007,14 @@ object GraphDSL extends GraphApply {
     private class PortOpsImpl[+Out](override val outlet: Outlet[Out @uncheckedVariance], b: Builder[_])
       extends PortOps[Out] {
 
-      override def withAttributes(attr: Attributes): Repr[Out] =
-        throw new UnsupportedOperationException("Cannot set attributes on chained ops from a junction output port")
+      override def withAttributes(attr: Attributes): Repr[Out] = throw settingAttrNotSupported
+      override def addAttributes(attr: Attributes): Repr[Out] = throw settingAttrNotSupported
+      override def named(name: String): Repr[Out] = throw settingAttrNotSupported
+      override def async: Repr[Out] = throw settingAttrNotSupported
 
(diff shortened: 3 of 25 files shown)


akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
```diff
@@ -1007,14 +1007,14 @@ object GraphDSL extends GraphApply {
     private class PortOpsImpl[+Out](override val outlet: Outlet[Out @uncheckedVariance], b: Builder[_])
       extends PortOps[Out] {
 
-      override def withAttributes(attr: Attributes): Repr[Out] =
-        throw new UnsupportedOperationException("Cannot set attributes on chained ops from a junction output port")
+      override def withAttributes(attr: Attributes): Repr[Out] = throw settingAttrNotSupported
```
drewhk: why not put the throw in that method?

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -64,6 +64,11 @@ final class Sink[-In, +Mat](private[stream] override val module: Module)
    */
   override def named(name: String): Sink[In, Mat] = withAttributes(Attributes.name(name))
 
+  /**
+   * Put an asynchronous boundary around this `Sink`
+   */
+  override def async(): Sink[In, Mat] = withAttributes(Attributes.asyncBoundary)
```
drewhk: should be `addAtributes`

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -147,6 +147,11 @@ final class Source[+Out, +Mat](private[stream] override val module: Module)
    */
   override def named(name: String): Repr[Out] = withAttributes(Attributes.name(name))
 
+  /**
+   * Put an asynchronous boundary around this `Source`
+   */
+  override def async(): Repr[Out] = withAttributes(Attributes.asyncBoundary)
```
drewhk: same here

johanandren:
PLS BUILD

johanandren:
Actually do not please build! Just saw the feedback.

akka-stream/src/main/scala/akka/stream/scaladsl/Graph.scala
johanandren: Wanted the throw line in the right place, but I guess it will anyways be in the stack trace.

drewhk: ok, I guess it is good as it is.

akka-stream/src/main/scala/akka/stream/Graph.scala
```diff
@@ -26,5 +26,10 @@ trait Graph[+S <: Shape, +M] {
 
   def named(name: String): Graph[S, M] = withAttributes(Attributes.name(name))
```
rkuhn: please change this to addAttributes as well, and remove the `.nest()` from the the implementation of `Flow.withAttributes`

akka-stream/src/main/scala/akka/stream/impl/SubFlowImpl.scala
```diff
@@ -34,6 +34,8 @@ class SubFlowImpl[In, Out, Mat, F[+_], C](val subFlow: Flow[In, Out, NotUsed],
   override def named(name: String): SubFlow[Out, Mat, F, C] =
     new SubFlowImpl[In, Out, Mat, F, C](subFlow.named(name), mergeBackFunction, finishFunction)
 
+  override def async: Repr[Out] = new SubFlowImpl[In, Out, Mat, F, C](subFlow.async, mergeBackFunction, finishFunction)
```
rkuhn: I like this style without the parens, I think it should be used everywhere.

rkuhn:
There should be a few docs that should also be changed. Otherwise LGreatTM!

akka-stream/src/main/scala/akka/stream/Graph.scala
johanandren: Ouch, `Source.addAttributes` calls `withAttributes` as well, I guess I should update that as well?

johanandren: Ah, it combines the existing `module.attributes and attr` and then replaces.

akka-stream/src/main/scala/akka/stream/impl/SubFlowImpl.scala
johanandren: I started out like that but the compiler was complaining about it when I put parens in the Java DSL. But I guess I can just omit the parens there as well.

rkuhn: yes

johanandren:
Weirdly enough there isn't a single mention of `viaAsync` in the docs except for the migration that I added.

johanandren:
I will add docs about it in ticket #19303 though

rkuhn:
There should be doc samples that contain `.withAttributes(Attributes.asyncBoundary)`, though.

drewhk:
This is a genuine failure here.

johanandren:
It's the `.nesting()` @rkuhn told me to remove, so since it shouldn't have multiple levels of nesting anymore I can remove the test?

rkuhn:
No, the test is needed, it is the attribute handling that needs to be fixed.

johanandren:
But if I removed `nested` then it isn't nested any more, and the test checks that it is nested?

rkuhn:
No, the test checks that the attributes work, and what the test does should still work. I’m working on a proposal for a fix.

rkuhn:
included in and superseded by #19754

PR #19744, opened 2016-02-10
do not require partial batch replay in TCK #19728

### LINKED ISSUE #19728: spec bug in Persistence TCK
The current TCK does the following:

* persist single events 1 to 3
* persist an atomic write of 4 and 5

Then the test case `"replay messages using a lower and upper sequence number bound"` requests events 2 to 4, expecting to receive all of them.

The point of AtomicWrites is that replay shall give us either none or all of them and the TCK violates this contract. In fact, it prevents journals from implementing it correctly.

Changing this will declare all existing journal implementations as invalid, but I doubt that any user relies upon the broken behavior. How do we go about this?

rkuhn:
fixes #19728


patriknw:
LGTM

PR #19746, opened 2016-02-10
Unicode alias from → to -> in FSM

bwmcadams:
- Some editors and code reformatters allow you to or automatically do
  reformatting from -> to →, which breaks FSM code in onTransition.
  This allows usage of → in onTransition.

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/actor/FSMTransitionSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/actor/FSMTransitionSpec.scala
@@ -51,7 +51,8 @@ object FSMTransitionSpec {
     }
     onTransition {
       case 0 -> 1 ⇒ target ! ((stateData, nextStateData))
-      case 1 -> 1 ⇒ target ! ((stateData, nextStateData))
+      // Verify the unicode alias of -> works correctly in onTransition
+      case 1 → 1  ⇒ target ! ((stateData, nextStateData))
     }
   }
 
--- a/akka-actor/src/main/scala/akka/actor/FSM.scala
+++ b/akka-actor/src/main/scala/akka/actor/FSM.scala
@@ -115,6 +115,13 @@ object FSM {
   }
 
   /**
+   * Convenience Unicode alias for the (S, S) pair to ->
+   */
+  object → {
+    def unapply[S](in: (S, S)) = Some(in)
+  }
+
+  /**
    * Log Entry of the [[akka.actor.LoggingFSM]], can be obtained by calling `getLog`.
    */
   final case class LogEntry[S, D](stateName: S, stateData: D, event: Any)
@@ -320,6 +327,11 @@ trait FSM[S, D] extends Actor with Listeners with ActorLogging {
    * reminder what the new state is.
    */
   val -> = FSM.->
+
+  /**
+   * Unicode alias for ->, matches a (S, S) pair
+   */
+  val → = FSM.->
 
   /**
    * This case object is received in case of a state timeout.


bwmcadams:
Actually, the alias is from → to -> now that I think of it.

rkuhn:
Looks reasonable, thanks! Would you mind adding a ticket and a test? (the latter might just involve changing two characters in a an existing spec, plus a comment)

bwmcadams:
Sure, try to get to that tonight.

patriknw:
LGTM

bwmcadams:
@rkuhn: Issue opened in #19772 ... comment & test modification made to verify functionality of change.

rkuhn:
OK TO TEST

rkuhn:
Hmm.
~~~
[info] akka-actor: found 2 potential binary incompatibilities while checking against com.typesafe.akka:akka-actor_2.11:2.3.12  (filtered 244)
[error]  * method →()akka.actor.FSM#-># in trait akka.actor.FSM does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("akka.actor.FSM.→")
[error]  * synthetic method akka$actor$FSM$_setter_$→_=(akka.actor.FSM#->#)Unit in trait akka.actor.FSM does not have a correspondent in old version
[error]    filter with: ProblemFilters.exclude[MissingMethodProblem]("akka.actor.FSM.akka$actor$FSM$_setter_$→_=")
~~~
The first one is okay, but the second one is not: I don’t think it is possible to add a `val` to a trait and maintain binary compatibility. Does it work with a `def` as well?

(and please squash to one commit and add the ticket number to the oneline summary, thanks!)

bwmcadams:
Hmm. The existing -> method was a val, I didn't even think of it. It will work with a def as well, it just points to an object containing an extractor.

Will check back in and squash.

bwmcadams:
I take that back: it fails to test with it as a def, the symbol cannot be found.

Is there a better way to resolve the binary compatibility issue?

rkuhn:
Not that I know. You can verify my suspicion by compiling an FSM with Akka 2.4.1 and then running it against the changed version—I think it will fail in the initialization of the trait because the setter is not present.

The extractor needs to be a stable value. Unless there is an annotation that tricks the compiler into accepting the `def` this will require an additional import :-(

bwmcadams:
Testing now. Without binary compatibility, is this something that would be usable in 2.5.x instead?

rkuhn:
We have decided to convert to the “normal” interpretation of what a “major” version is, so we can only break BC in 3.0—and we don’t have a schedule for that yet.

akka-actor/src/main/scala/akka/actor/FSM.scala
```diff
@@ -322,6 +329,11 @@ trait FSM[S, D] extends Actor with Listeners with ActorLogging {
   val -> = FSM.->
 
   /**
+   * Unicode alias for ->, matches a (S, S) pair
+   */
+  val → = FSM.->
```
MQ-EL: we can define new trait contains this method, and then let ``FSM`` inherit from it

rkuhn: No, changing the parent hierarchy of FSM is also an incompatible change, and it will incur the same AbstractMethodError as putting the value here.

MQ-EL: yeh, it is an imcompatible change actually, but MiMA doesn't known it.

drewhk:
If I understand correctly then this PR should be closed?

rkuhn:
No, the extractor is still useful, even if it has to be imported explicitly.

drewhk:
Ok, got it.

akka-actor-tests/src/test/scala/akka/actor/FSMTransitionSpec.scala
```diff
@@ -51,7 +51,8 @@ object FSMTransitionSpec {
     }
     onTransition {
       case 0 -> 1 ⇒ target ! ((stateData, nextStateData))
-      case 1 -> 1 ⇒ target ! ((stateData, nextStateData))
+      // Verify the unicode alias of -> works correctly in onTransition
+      case 1 → 1  ⇒ target ! ((stateData, nextStateData))
```
vasseurjb: Codacy Issue found: [Non ascii characters are not allowed](https://www.codacy.com/app/vasseurjb/akka/file/2046658181/issues/source?bid=3132306&fileBranchId=3132374#l55)

akka-actor/src/main/scala/akka/actor/FSM.scala
vasseurjb: Codacy Issue found: [Field name '?' does not match the regular expression '^[a-z][A-Za-z0-9]*$'](https://www.codacy.com/app/vasseurjb/akka/file/2046658285/issues/source?bid=3132306&fileBranchId=3132374#l334)

drewhk:
Ping :)

rkuhn:
rescheduled for Akka 3.0

akka-actor/src/main/scala/akka/actor/FSM.scala
ktoso: actually we don't need this val – if instead the object is defined inside the class it all works fine and MiMa does not complain since there's no weird setter.

// Revived looking at this since bumping scalariform

PR #19749, opened 2016-02-11
MetricsBasedResizerSpec appears to be at risk of race condition failure

markalanrichards:
Got a test error running the complete suite of tests and other tests seem to have this similar Thread sleep.
Please check and accept if appropriate.
Thanks

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/routing/MetricsBasedResizerSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/routing/MetricsBasedResizerSpec.scala
@@ -212,6 +212,7 @@ class MetricsBasedResizerSpec extends AkkaSpec(ResizerSpec.config) with DefaultT
       msgs.head.open()
 
       router.sendToAll()
+      Thread.sleep(1) // wait for routees to update their mail boxes
       resizer.reportMessageCount(router.routees, router.msgs.size)
       resizer.performanceLog.get(2) should not be empty
     }


akka-actor-tests/src/test/scala/akka/routing/MetricsBasedResizerSpec.scala
```diff
@@ -212,6 +212,7 @@ class MetricsBasedResizerSpec extends AkkaSpec(ResizerSpec.config) with DefaultT
       msgs.head.open()
 
       router.sendToAll()
+      Thread.sleep(1) // wait for routees to update their mail boxes
```
patriknw: I would prefer to put this in an `awaitAssert` instead

```scala
awaitAssert {
  resizer.reportMessageCount(router.routees, router.msgs.size)
  resizer.performanceLog.get(2) should not be empty
}
```

viktorklang: :+1: @patriknw

markalanrichards:
Sure, I'm not too worried whether it is Thread.sleep or awaitAssert, I was following what looked like a habit in that area of the tests: does this mean the other tests will be swapped to awaitAssert too?

patriknw:
if possible, that would be good

markalanrichards:
Great

akka-actor-tests/src/test/scala/akka/routing/MetricsBasedResizerSpec.scala
drewhk: agree

drewhk:
@markalanrichards  do you have time to finish this, or would you prefer if we take over?

rkuhn:
I’m taking this over now.

rkuhn:
Thanks for noticing, I ended up making it deterministic without sleeps :-)

PR #19751, opened 2016-02-11
=str #19591 updated 'partition to three subscribers' to use Sink.seq …

israel:
…and added a new test to cover a case where partitioner return an index out of bounds


drewhk:
OK TO TEST

rkuhn:
Thanks @israel !

PR #19753, opened 2016-02-12
Generate version files

pvlugter:
This means that `ActorSystem.Version` and `akka.version` in config can be set dynamically from sbt using the version setting (without the find-replace in the release script) and work with dbuild.

### DIFF
--- a/project/Version.scala
+++ b/project/Version.scala
@@ -1,0 +1,31 @@
+package akka
+
+import sbt._
+import sbt.Keys._
+
+/**
+ * Generate version.conf and akka/Version.scala files based on the version setting.
+ */
+object Version {
+
+  def versionSettings: Seq[Setting[_]] = inConfig(Compile)(Seq(
+    resourceGenerators <+= generateVersion(resourceManaged, _ / "version.conf",
+      """|akka.version = "%s"
+         |"""),
+    sourceGenerators <+= generateVersion(sourceManaged, _ / "akka" / "Version.scala",
+      """|package akka
+         |
+         |object Version {
+         |  val current: String = "%s"
+         |}
+         |""")
+  ))
+
+  def generateVersion(dir: SettingKey[File], locate: File => File, template: String) = Def.task[Seq[File]] {
+    val file = locate(dir.value)
+    val content = template.stripMargin.format(version.value)
+    if (!file.exists || IO.read(file) != content) IO.write(file, content)
+    Seq(file)
+  }
+
+}
--- a/akka-actor-tests/src/test/scala/akka/config/ConfigSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/config/ConfigSpec.scala
@@ -27,8 +27,8 @@ class ConfigSpec extends AkkaSpec(ConfigFactory.defaultReference(ActorSystem.fin
       {
         import config._
 
-        getString("akka.version") should ===("2.4-SNAPSHOT")
-        settings.ConfigVersion should ===("2.4-SNAPSHOT")
+        getString("akka.version") should ===(ActorSystem.Version)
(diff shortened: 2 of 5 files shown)


bantonsson:
OK TO TEST

patriknw:
LGTM

project/Version.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  def versionSettings: Seq[Setting[_]] = inConfig(Compile)(Seq(
+    resourceGenerators <+= generateVersion(resourceManaged, _ / "version.conf",
+      """|akka.version = "%s"
+         |"""),
+    sourceGenerators <+= generateVersion(sourceManaged, _ / "akka" / "Version.scala",
+      """|package akka
+         |
+         |object Version {
+         |  val current: String = "%s"
```
viktorklang: omit the ": String" part to make it a constant?

rkuhn: then it would also have to be `final`—although I don’t know what the benefit is for strings

rkuhn:
Luckily we have already eradicated most of the findReplace things and this PR looks like a good opportunity to get rid of it entirely, the only remnants are:
~~~
akka-kernel/src/main/dist/config/application.conf:# For more info about config, please visit the Akka Documentation: http://akka.io/docs/akka/2.4-SNAPSHOT/
akka-kernel/src/main/dist/README:This is the Akka 2.4-SNAPSHOT download.
akka-docs/rst/intro/deployment-scenarios.rst:   unzip akka-sample-main-scala-2.4-SNAPSHOT.zip
akka-docs/rst/intro/deployment-scenarios.rst:   chmod u+x akka-sample-main-scala-2.4-SNAPSHOT/bin/akka-sample-main-scala
akka-docs/rst/intro/deployment-scenarios.rst:   akka-sample-main-scala-2.4-SNAPSHOT/bin/akka-sample-main-scala sample.hello.Main
~~~
The docs can be conquered by using `@version@` but the akka-kernel may need its README to be generated (the other one can just link to `http://akka.io/docs/akka/current/`).

rkuhn:
Let’s get this in before Tuesday, it does not need to be in RC3.

project/Version.scala
pvlugter: Yes, could be a constant (with `final` and without type). Should then become an inlined string literal in `ActorSystem.Version`. Shall I make that change?

pvlugter:
@rkuhn, great if the find-replace can be removed altogether :)

project/Version.scala
pvlugter: I just tried out the constant. Certainly becomes a string literal in `ActorSystem$`.

Looks like the only gotcha is incremental compiles, which don't pick up changes. So this:

```
sbt> project akka-actor
sbt> compile
sbt> set version := "1.2.3"
sbt> compile
[info] Compiling 1 Scala source to ...
```
Only the `Version.scala` file is recompiled. Leading to:

```
scala> akka.Version.current
res0: String = 1.2.3

scala> ConfigFactory.load.getString("akka.version")
res1: String = 1.2.3

scala> akka.actor.ActorSystem.Version
res2: String = 2.4-SNAPSHOT

scala> startSystem()
akka.ConfigurationException: Akka JAR version [2.4-SNAPSHOT] does not match the provided config version [1.2.3]
```

viktorklang: Sounds like a good reason to avoid the constant :)

-- 
Cheers,
√

rkuhn: Indeed!

rkuhn:
The find-replace.sh script is still used a bit more—in the samples. We should fix that another time.

ktoso:
Was about to LGTM ;) Good :)

PR #19754, opened 2016-02-12
nicer declarative .async boundaries and remove Module.nest()

### LINKED ISSUE #19732: A more consistent and succinct way to mark async boundaries

rkuhn:
fixes #19732 

This adds the nest() removal on top of Johan’s work.

### DIFF
--- a/akka-stream-tests/src/test/resources/reference.conf
+++ b/akka-stream-tests/src/test/resources/reference.conf
@@ -8,4 +8,5 @@ akka {
   akka.actor.warn-about-java-serializer-usage = false
 
   stream.materializer.debug.fuzzing-mode = on
+  stream.secret-test-fuzzing-warning-disable = 42
 }
--- a/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
@@ -511,7 +511,9 @@ class InterpreterSpec extends AkkaSpec with GraphInterpreterSpecKit {
       downstream.requestOne()
       lastEvents() should be(Set(RequestOne))
 
-      upstream.onComplete()
+      EventFilter[IllegalArgumentException](pattern = ".*Cannot pull closed port.*", occurrences = 1).intercept {
+        upstream.onComplete()
+      }
       val ev = lastEvents()
       ev.nonEmpty should be(true)
       ev.forall {
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowIteratorSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowIteratorSpec.scala
@@ -4,7 +4,6 @@
 package akka.stream.scaladsl
 
 import akka.NotUsed
-
 import scala.collection.immutable
 import scala.concurrent.duration._
 import akka.stream.ActorMaterializer
@@ -14,6 +13,7 @@ import akka.stream.testkit.Utils._
 import org.reactivestreams.Subscription
 import akka.testkit.TestProbe
 import org.reactivestreams.Subscriber
+import akka.testkit.EventFilter
 
 class FlowIteratorSpec extends AbstractFlowIteratorSpec {
   override def testName = "A Flow based on an iterator producing function"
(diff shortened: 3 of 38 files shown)


akka-stream-tests/src/test/resources/reference.conf
```diff
@@ -8,4 +8,5 @@ akka {
   akka.actor.warn-about-java-serializer-usage = false
 
   stream.materializer.debug.fuzzing-mode = on
+  stream.secret-test-fuzzing-warning-disable = 42
```
drewhk: ?

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
```diff
@@ -511,7 +511,9 @@ class InterpreterSpec extends AkkaSpec with GraphInterpreterSpecKit {
       downstream.requestOne()
       lastEvents() should be(Set(RequestOne))
 
-      upstream.onComplete()
+      EventFilter[IllegalArgumentException](pattern = ".*Cannot pull closed port.*", occurrences = 1).intercept {
```
drewhk: why the change?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowIteratorSpec.scala
```diff
@@ -40,7 +40,9 @@ class FlowIterableSpec extends AbstractFlowIteratorSpec {
     sub.request(1)
     c.expectNext(1)
     c.expectNoMsg(100.millis)
-    sub.request(2)
+    EventFilter[IllegalStateException](message = "not two", occurrences = 1).intercept {
```
drewhk: same here

akka-stream-tests/src/test/resources/reference.conf
rkuhn: cleaning up the build output

drewhk:
LGTM after those questions are explalined.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
rkuhn: same as above (and below)

akka-stream-tests/src/test/resources/reference.conf
2beaucoup: A `FuzzingGraphInterpreter` that lives in stream-tests would be nice.

drewhk: WDYM? It is already exposed to users: http://doc.akka.io/docs/akka-stream-and-http-experimental/2.0.3/scala/stream-testkit.html#fuzzing-mode

2beaucoup: I see. I thought that way it would be more obvious that fuzzing is only intended for testing and we could get rid of the warning/setting.

(Correction: I guess stream-testkit ISO stream-tests would be the right place.)

drewhk: There is a big warning in the docs:

"Warning
Never use this setting in production or benchmarks. This is a testing tool to provide more coverage of your code during tests, but it reduces the throughput of streams. A warning message will be logged if you have this setting enabled."

Also in the reference.conf:

"       # Enables the fuzzing mode which increases the chance of race conditions
        # by aggressively reordering events and making certain operations more
        # concurrent than usual.
        # This setting is for testing purposes, NEVER enable this in a production
        # environment!
        # To get the best results, try combining this setting with a throughput
        # of 1 on the corresponding dispatchers."

Since these are the two places where ordinary people can learn about the existence of the feature, I assume that these warnings, and additionally the warning log, is more than enough.

2beaucoup: OK. I just thought that this approach would be cleaner as it removes testing knobs from the API/settings. But the current warnings should work as well.

> Since these are the two places where ordinary people can learn about the existence of the feature...

A third place would be the API itself. ;)

PR #19757, opened 2016-02-12
#19553 Fix One2OneBidi not failing if inner flow completion wins the race with cancellation

drewhk:
Also made relevant tests race-free.


akka-stream/src/main/scala/akka/stream/scaladsl/One2OneBidiFlow.scala
```diff
@@ -80,9 +82,10 @@ object One2OneBidiFlow {
             }
           } else throw new UnexpectedOutputException(element)
         }
-        override def onUpstreamFinish(): Unit =
-          if (pending == 0 && !innerFlowCancelled) complete(outOut)
+        override def onUpstreamFinish(): Unit = {
+          if (pending == 0 && isClosed(inIn) && !innerFlowCancelled) complete(outOut)
```
rkuhn: I doubt the purpose of this whole stage given that there is no way to handle cancellation such that it is both intuitive and correct. What is the point?

drewhk: What do you mean? What is not intuitive or correct about it?

drewhk:
we discussed with @rkuhn The fix is correct (well, let's wait on validation), but the stage is not generic or intuitive for generic use, since the current semantics are very much married to its use-cases in HTTP. Hence, in a second PR I will move this to a private place in HTTP.

rkuhn:
LGTM

drewhk:
hm, failure is real.

PR #19769, opened 2016-02-12
=htt Remove redundant modifiers

naferx:


akka-http-core/src/main/java/akka/http/javadsl/model/headers/LanguageRange.java
```diff
@@ -5,11 +5,11 @@
...
-    public abstract String primaryTag();
-    public abstract float qValue();
-    public abstract boolean matches(Language language);
-    public abstract Iterable<String> getSubTags();
-    public abstract LanguageRange withQValue(float qValue);
+    String primaryTag();
+    float qValue();
+    boolean matches(Language language);
+    Iterable<String> getSubTags();
+    LanguageRange withQValue(float qValue);
```
rkuhn: aren’t these package-private now?

ktoso: they're public (welcome to java (method declarations in in interface)).

I would like to keep the modifiers, as we have many scala contributors, and after some years of Scala one simply forgets what "default" means in some places in java. 
So thanks for the PR but I'd like to keep them (i.e. not accept the PR), thanks!

rkuhn: right, you may take my confusion as a data point :-)

ktoso:
Thanks for the idea / PR but we do want to be explicit about these things as commented above.
Thanks again and please don't let this discourage you from future contribs, thanks a lot in advance!

Issue #19781, opened 2016-02-15
Some elements ignore overriden `InputBuffer` attribute


ALPHA-60:
```
Source.repeat(1).runWith(Sink.queue().withAttributes(inputBuffer(0,0)))
```

should produce an `IllegalArgumentException` based on `Sink.queue`'s buffer size precondition, but it doesn't.

This happens with `Sink.queue`, `Sink.actorRefWithAck`,  `StreamConverters.asInputStream` and `StreamConverters.asOutputStream`.

viktorklang:
Fixed by #19781

PR #19783, opened 2016-02-15
=str #19781 Use overriden `InputBuffer` attribute

ALPHA-60:
- Sink.{queue, actorRefWithAck} and StreamConverters.{asInputStream,
  asOutputStream} now use overriden/inherited `InputBuffer` attribute
- They now use their default attributes as initial attributes.

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
@@ -8,6 +8,7 @@ import java.util.concurrent.TimeoutException
 
 import akka.actor.ActorSystem
 import akka.stream._
+import akka.stream.Attributes.inputBuffer
 import akka.stream.impl.StreamSupervisor.Children
 import akka.stream.impl.io.InputStreamSinkStage
 import akka.stream.impl.{ ActorMaterializerImpl, StreamSupervisor }
@@ -217,6 +218,19 @@ class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
       inputStream.read() should ===(-1)
 
       inputStream.close()
+    }
+  }
+
+  "fail to materialize with zero sized input buffer" in {
+    an[IllegalArgumentException] shouldBe thrownBy {
+      Source.single(byteString)
+        .runWith(StreamConverters.asInputStream(timeout).withAttributes(inputBuffer(0, 0)))
+      /*
+       With Source.single we test the code path in which the sink
+       itself throws an exception when being materialized. If
+       Source.empty is used, the same exception is thrown by
+       Materializer.
+       */
     }
   }
 }
--- a/akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
@@ -8,10 +8,11 @@ import java.util.concurrent.TimeoutException
 
 import akka.actor.ActorSystem
 import akka.stream._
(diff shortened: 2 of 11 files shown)


akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
```diff
@@ -219,4 +220,11 @@ class InputStreamSinkSpec extends AkkaSpec(UnboundedMailboxConfig) {
       inputStream.close()
     }
   }
+
+  "fail to materialize with zero sized input buffer" in {
+    an[IllegalArgumentException] shouldBe thrownBy {
+      Source.single(byteString)
```
ALPHA-60: I used `Source.single` instead of `Source.empty`, because zero sized buffer error would be picked up even without this fix.

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
```diff
@@ -147,5 +148,13 @@ class OutputStreamSourceSpec extends AkkaSpec(UnboundedMailboxConfig) {
       sourceProbe.expectMsg(GraphStageMessages.DownstreamFinish)
       the[Exception] thrownBy outputStream.write(bytesArray) shouldBe a[IOException]
     }
+
+    "fail to materialize with zero sized input buffer" in {
+      an[IllegalArgumentException] shouldBe thrownBy {
+        StreamConverters.asOutputStream(timeout)
+          .withAttributes(inputBuffer(0, 0))
+          .runWith(Sink.head)
```
ALPHA-60: As above, test would pass with `Sink.ignore` even without fixes to OutputStreamSource, so i used Sink.head.

akka-stream-tests/src/test/scala/akka/stream/io/InputStreamSinkSpec.scala
drewhk: please add a comment so this is not forgotten

akka-stream-tests/src/test/scala/akka/stream/io/OutputStreamSourceSpec.scala
drewhk: please add a comment so this is not forgotten

drewhk:
LGTM great work!

ktoso:
LGTM, thanks for noticing and fixing these!

ktoso:
OK TO TEST

ktoso:
Refs #19781

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -289,13 +289,14 @@ final private[stream] class QueueSink[T]() extends GraphStageWithMaterializedVal
   type Requested[E] = Promise[Option[E]]
 
   val in = Inlet[T]("queueSink.in")
+  override def initialAttributes = DefaultAttributes.queueSink
```
rkuhn: since this is here now we can also remove the `withAttributes` from the implementation of `Sink.queue`

rkuhn: (the same goes for the StreamConverters methods)

rkuhn:
git failure

ALPHA-60:
Could I modify the tests commited in #19676 and #19654, so that the Sink buffer size is set explicitly, instead of relying on the fact that default size is 16?

rkuhn:
failure was #19792

rkuhn:
Concerning those tests: good point, it would be good to change them to be explicit about these buffer sizes. Thanks @ALPHA-60 !

PR #19787, opened 2016-02-15
PREVIEW: +htc #18351 implement dynamic TLS session renegotiation

jrudolph:
I finally got to rebase and update the work on #18351 I did originally as part of #19210 to master.

Previously, the idea was to supply a continuation which is run when a new session has been established to create a response. In #19210, @rkuhn noticed that we couldn't write a high-level directive to support this feature because we would lose the "route continuation", i.e. the logic in wrapping directives when the continuation is executed. Therefore, I added another commit which changes the behavior so that now a new request must be supplied which is then run through the user handler flow after the new session was established. The advantage is that this will work transparently with existing handlers. The disadvantage is added complexity to allow injecting an extra request into the flow. Also, it's now easy to introduce a loop where a request without client authentication data triggers a renegotiation but the client doesn't provide any new authentication data. When the same request handling logic as before is run (i.e. the original request was supplied to `RequestClientCertificate`) this will then trigger another renegotiation and so on.

All in all I'm still not convinced that this whole feature is actually worth the added complexity. I tested it against Firefox and Chromium which pop up a UI on renegotiation to select a certificate, so it seems to work fine for the single case I could manually test. We cannot write automated tests against it, however, because AFAIK Java's SSLEngine doesn't provide any way on the client side to figure out that a renegotiation was started with the purpose to request a client certificate. AFAIK there's no spec which would explain how the interaction should be done and if it is secure the way it was implemented here. So far, the W3C guys were the only ones ever asking for this feature. In summary, I wouldn't be disappointed if this PR wouldn't be merged and instead be left here as a demonstration that the feature could be implemented and how in case the trade-offs would change in the future. :)

/cc @sirthias


ktoso:
Slipping from 2.4.2 but will be back to have a look at it very soon!

ktoso:
@drewhk @rkuhn your reviews regarding this PR would be very welcome, thanks in advance :)

drewhk:
I will do it tomorrow.

bblfish:
Thanks, if this goes in I can try writing against it and provide feedback. I want to re-write my server using just akka, and this would allow us to both try the SSL authentication and others, and so provide feedback on what the advantages and inconveniences of the various methods are.

The W3C TAG is working on a [report on client certificates](http://w3ctag.github.io/client-certificates/) in this area that may be interesting to review.
 
There are a number of other ways of doing things the pure HTTP route such as [HTTP-Signature](https://tools.ietf.org/html/draft-cavage-http-signatures-05) but that is still in its infancy, but these have no UI support at all for the moment, and it will presumably take quite a lot of time before these things get full browser support. It would require the Verifiable Credentials Task forces [Use Cases](http://opencreds.org/specs/source/use-cases/) to be accepted as a basis for a W3C working group, for that group to succeed in providing a standard within two years, and for browsers then to implement a credentials user interface.

ktoso:
Status update: I've read this PR a number of times now but always getting hung up on "no idea if it's worth it"...

bblfish:
I don't know the code here well enough. If Client Certificates have made it into browsers even given their low civilian usage, it is probably because very large companies and the US military are using it. So my guess is for akka to be useable by those institutions this could be a deal breaker.  You can tell by how Microsoft and other large corporations have put TLS Client certs back on the agenda at the IETF. I can find the links if needed. There would also be an advantage that it would be possible to follow more closely the various proposals put forward for TLS 3.0 client cert usage.

akka-stream/src/main/scala/akka/stream/impl/io/SslTlsCipherActor.scala
```diff
@@ -432,6 +432,7 @@ private[akka] class SslTlsCipherActor(settings: ActorMaterializerSettings,
         currentSession = session
         corkUser = false
     }
+    outputBunch.enqueue(UserOut, SessionBytes(currentSession, ByteString.empty))
```
rkuhn: This is not guaranteed to be met with enthusiasm by the downstream subscriber: before calling `handshakeFinished` we have already called `flushToUser` which means that we don’t know if there is any more demand outstanding at this point.

rkuhn:
My take on this (having read some of the links above and the code) is that the mechanism is not yet established, much less specified, of how TLS renegotiation on an HTTP connection shall be utilized. Therefore I would see our implementation of such a feature as a basis for experimentation rather than a full-fledged and mature end-user API. As such, we would need to expose not only the current best guess of a use-case (which Johannes has provided), we would need to expose the full range of possibilities:

* a route on the server can trigger a renegotiation with arbitrary changes to the parameters (bounded by what SSLEngine supports)
* it should be possible to encode that the request that triggers this should only be answered after the renegotiation has completed, and it should depend on whether a new session was established or not (in case of failure)
* the client API needs to forward the session information to user code so that it can react appropriately
* the client API should also allow a renegotiation to be triggered

With these it should be possible to use Akka HTTP as a testbed for trying out new schemes.

Given the reasoning above I do not consider this addition a high priority for Akka itself, it would need to be driven by external interest and volunteered resources. The burden of maintaining such a feature after its addition will fall into the core team’s lap when external interest diminishes, which means that we will have to carefully consider the code complexity, quality of tests, comments, and documentation when reviewing proposed code changes.

bblfish:
Perhaps what I should try is see if I can implement [WebID-TLS](https://www.w3.org/2005/Incubator/webid/spec/tls/) with this patch...

bblfish:
Got this to work with my simple server using `bindAndHandleAsync`, after compiling this commit.

bblfish:
I noticed that one has to be careful not to have an infinite recursion of certificate requests. This can happen when doing say a `curl` request without supplying a certificate. 

```
$curl -i -k  https://localhost:8443/src/main/scala/co/os/Solid.scala
```

This lead to requests being sent that had a growing number of `Timeout-Access` links like this:


> [INFO] [03/06/2016 22:18:25.297] [co-os-akka.actor.default-dispatcher-3] [akka://co-os/user/SoLiD/src/main/scala/co/os/Solid.scala/$a] guard received HttpRequest(HttpMethod(GET),http://localhost:8443/src/main/scala/co/os/Solid.scala,List(Tls-Session-Info: [Session-967, TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA], Host: localhost:8443, User-Agent: curl/7.45.0, Accept: */*, Timeout-Access: <function1>, Timeout-Access: <function1>, Timeout-Access: <function1>, Timeout-Access: <function1>, Timeout-Access: <function1>, Timeout-Access: <function1>,


A [simple fix](https://github.com/co-operating-systems/Reactive-SoLiD/commit/d74f52b878dfdec08461f860cd45217c81fcedd6) for this was to add some header to the returned response 
`request ! HttpResponse(headers = RequestClientCertificate(req.addHeader(XCert))`
and then check for it's existence before asking for a new certificate. If it exists, then the request was already sent.

bblfish:
Just discovered this live RFC draft that should be of interest: [Reactive Certificate-Based Client Authentication in HTTP/2](https://tools.ietf.org/html/draft-thomson-http2-client-certs-01)

ktoso:
Closing for now, and market the ticket as obsolete (but please feel free to re-open if anyone has strong feelings about including it).

PR #19788, opened 2016-02-15
-htp #19577 replace overloads of bind with HttpConnect

### LINKED ISSUE #19577: JavaDSL: bind(to())
We now provide outgoingConnection(toHttpsHost()) to avoid having overloads explosion in HTTP.
It would feel more consistent if the same style also works for the server.

This would have to be done for 2.4.2 or dropped.

ktoso:
This was the missing piece to resolve https://github.com/akka/akka/issues/19577 completely.
It simply replaces the overloads that explicitly take the ConnectionContext (possibly HttpsConnectionContext) by using HttpConnect everywhere, which carries along an ConnectionContext – when it's Https, it binds as an HTTPS server.

### DIFF
--- a/akka-actor/src/main/scala/akka/routing/RouterConfig.scala
+++ b/akka-actor/src/main/scala/akka/routing/RouterConfig.scala
@@ -42,7 +42,7 @@ trait RouterConfig extends Serializable {
 
   /**
    * Create the actual router, responsible for routing messages to routees.
- *
+   *
    * @param system the ActorSystem this router belongs to
    */
   def createRouter(system: ActorSystem): Router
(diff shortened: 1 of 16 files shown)


johanandren:
Isn't there testcoverage for the removed methods (which I added while you had vacation)?

johanandren:
LGTM except for that

ktoso:
Right thanks @johanandren. I'm still a bit tired from all the meds it seems :-/ Will fix so we can get it in and release early tomorrow.

ktoso:
Fixed.

ktoso:
Rebased

akka-docs/rst/java/code/docs/http/javadsl/server/HttpServerExampleDocTest.java.orig
```diff
@@ -0,0 +1,215 @@
+/*
```
rkuhn: this file doesn’t really want to be here :-)

ktoso: sad file :(

ktoso: (removed)

rkuhn:
LGTM

PR #19800, opened 2016-02-16
add new Streams Quick Start

rkuhn:
also move IOResult into `akka.stream` package

### DIFF
--- a/akka-docs/rst/java/stream/stream-quickstart.rst
+++ b/akka-docs/rst/java/stream/stream-quickstart.rst
@@ -1,7 +1,141 @@
 .. _stream-quickstart-java:
 
-Quick Start Guide: Reactive Tweets
-==================================
+Quick Start Guide
+=================
+
+A stream usually begins at a source, so this is also how we start an Akka
+Stream. Before we create one, we import the full complement of streaming tools:
+
+.. includecode:: ../code/docs/stream/QuickStartDocTest.java#imports
+
+Now we will start with a rather simple source, emitting the integers 1 to 100:
+
+.. includecode:: ../code/docs/stream/QuickStartDocTest.java#create-source
+
+The :class:`Source` type is parameterized with two types: the first one is the
+type of element that this source emits and the second one may signal that
+running the source produces some auxiliary value (e.g. a network source may
+provide information about the bound port or the peer’s address). Where no
+auxiliary information is produced, the type ``akka.NotUsed`` is used—and a
+simple range of integers surely falls into this category.
+
+Having created this source means that we have a description of how to emit the
+first 100 natural numbers, but this source is not yet active. In order to get
+those numbers out we have to run it:
+
+.. includecode:: ../code/docs/stream/QuickStartDocTest.java#run-source
+
+This line will complement the source with a consumer function—in this example
+we simply print out the numbers to the console—and pass this little stream
+setup to an Actor that runs it. This activation is signaled by having “run” be
+part of the method name; there are other methods that run Akka Streams, and
+they all follow this pattern.
+
+You may wonder where the Actor gets created that runs the stream, and you are
+probably also asking yourself what this ``materializer`` means. In order to get
+this value we first need to create an Actor system:
+
+.. includecode:: ../code/docs/stream/QuickStartDocTest.java#create-materializer
+
+There are other ways to create a materializer, e.g. from an
+:class:`ActorContext` when using streams from within Actors. The
+:class:`Materializer` is a factory for stream execution engines, it is the
+thing that makes streams run—you don’t need to worry about any of the details
+just now apart from that you need one for calling any of the ``run`` methods on
+a :class:`Source`.
+
+The nice thing about Akka Streams is that the :class:`Source` is just a
+description of what you want to run, and like an architect’s blueprint it can
+be reused, incorporated into a larger design. We may choose to transform the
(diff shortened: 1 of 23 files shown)


akka-stream/src/main/scala/akka/stream/IOResult.scala
```diff
@@ -1,7 +1,7 @@
 /**
  * Copyright (C) 2016 Typesafe Inc. <http://www.typesafe.com>
  */
-package akka.stream.io
+package akka.stream
```
ktoso: I don't mind but would like to understand the reason behind this move - why not in `io`?

rkuhn: No, much simpler: I didn’t like the additional required import in the docs sample. And there is no reason why it should be in here.

akka-docs/rst/java/stream/stream-quickstart.rst
```diff
@@ -1,7 +1,140 @@
...
+One of the nice parts of Akka Streams—and something that other stream libraries
+do not offer—is that not only sources can be reused like blueprints, all other
+elements can be as well. We can take the file-writing :class:`Sink`, prepend
+the processing steps necessary to get the :class:`ByteString` elements from
+incoming strings and package that up as a reusable piece as well. Since the
+language for writing these streams always flows from left to right (just like
+plain English), we need a starting point that is like a source but with an
+“open” input. In Akka Streams this is called a :class:`Flow`:
+
+.. includecode:: ../code/docs/stream/QuickStartDocTest.java#transform-sink
```
drewhk: one thing that is not nice about this particular example is that due to it taking a filename as parameter, it is not actually reused here (of course it could be, but for newbies this might be a bit confusing).

akka-docs/rst/java/stream/stream-quickstart.rst
```diff
@@ -1,7 +1,140 @@
...
+Time-Based Processing
+---------------------
+
+Before we start looking at a more involved example we explore the streaming
+nature of what Akka Streams can do. Starting from the ``factorials`` source
+we transform the stream by zipping it together with another stream,
+represented by a :class:`Source` that emits the number 0 to 100: the first
+number emitted by the ``factorials`` source is the factorial of zero, the
+second is the factorial of one, and so on. We combine these two by forming
+strings like ``"3! = 6"``.
```
drewhk: hehe, nice sample :)

akka-docs/rst/java/stream/stream-quickstart.rst
```diff
@@ -1,7 +1,140 @@
...
+elements can be as well. We can take the file-writing :class:`Sink`, prepend
+the processing steps necessary to get the :class:`ByteString` elements from
+incoming strings and package that up as a reusable piece as well. Since the
+language for writing these streams always flows from left to right (just like
+plain English), we need a starting point that is like a source but with an
+“open” input. In Akka Streams this is called a :class:`Flow`:
+
+.. includecode:: ../code/docs/stream/QuickStartDocTest.java#transform-sink
+
+Starting from a flow of strings we convert to :class:`ByteString` and then feed
```
ktoso: we convert *each* ?

akka-docs/rst/java/stream/stream-quickstart.rst
```diff
@@ -1,7 +1,140 @@
...
+.. includecode:: ../code/docs/stream/QuickStartDocTest.java#add-streams
+
+All operations so far have been time-independent and could have been performed
+in the same fashion on strict collections of elements. The next line
+demonstrates that we are in fact dealing with streams that can flow at a
+certain speed: we use the ``throttle`` combinator to slow down the stream to 1
+element per second (the second ``1`` in the argument list is the maximum size
+of a burst that we want to allow—passing ``1`` means that the first element
+gets through immediately and the second then has to wait for one second and so
+on).
```
drewhk: maybe add a note about the `shaping()` parameter

akka-stream/src/main/scala/akka/stream/IOResult.scala
ktoso: Then I don't get it, it is about IO so would have expected it to live in the IO package hm...
*:class:`IOResult` is a type that IO operations return in Akka Streams*

I don't mind enough to battle over it but it seems like a weird move if only for the nicer import hm

akka-docs/rst/java/stream/stream-quickstart.rst
rkuhn: I’m getting at a different kind of reuse here, the one you’re talking about is one step beyond (and people hopefully can grasp that from the reuse of Source already).

akka-docs/rst/java/stream/stream-quickstart.rst
rkuhn: yes

drewhk:
LGTM

akka-docs/rst/java/stream/stream-quickstart.rst
rkuhn: I’d rather keep it as concise as possible, the purpose is explained and the name is quite fitting, so I don’t think this is a hindrance to understanding.

ktoso:
LGTM

darionyaphet:
Nice Work Stream :+1:

PR #19805, opened 2016-02-16
splitting up TLS APIs

drewhk:


akka-stream/src/main/scala/akka/stream/SslTlsOptions.scala
```diff
@@ -0,0 +1,255 @@
...
+case object IgnoreCancel extends IgnoreCancel
+
+/**
+ * see [[TLSClosing]]
+ */
+sealed abstract class IgnoreComplete extends TLSClosing {
+  override def ignoreCancel = false
+  override def ignoreComplete = true
+}
+case object IgnoreComplete extends IgnoreComplete
```
ktoso: those not inside the TLSProtocol?

drewhk: First, it is not protocol, but options, second it would end up as object-in-object which is bad for Java.

ktoso: Ok, got it

drewhk:
What failed even there? Jenkins status ball is grey.

drewhk:
PLS BUILD

rkuhn:
LGTM

akka-http-core/src/main/scala/akka/http/impl/engine/ws/WebSocketClientBlueprint.scala
```diff
@@ -13,8 +13,8 @@ import akka.util.ByteString
 import akka.event.LoggingAdapter
 
 import akka.stream.stage._
-import akka.stream.BidiShape
-import akka.stream.io.{ SessionBytes, SendBytes, SslTlsInbound }
+import akka.stream._
```
viktorklang: why all of akka.stream?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
```diff
@@ -9,12 +9,13 @@ import java.net.InetSocketAddress
 import java.security.MessageDigest
 import java.util
 import javax.net.ssl.SSLSession
+import akka.stream.scaladsl.ScalaSessionAPI
```
viktorklang: Why is this named ScalaSessionAPI when it is already located in the scaladsl package?

akka-stream/src/main/scala/akka/stream/SslTlsOptions.scala
```diff
@@ -0,0 +1,255 @@
+/**
+ * Copyright (C) 2015-2016 Typesafe Inc. <http://www.typesafe.com>
+ */
+package akka.stream
```
viktorklang: I find it weird that we go through all the trouble to have a `javadsl` and a `scaladsl` package to separate the APIs and here we just munge them together in the same namespace—I must've missed what's the rationale for this?

akka-stream/src/main/scala/akka/stream/SslTlsOptions.scala
```diff
@@ -0,0 +1,255 @@
...
+ *
+ *  - [[EagerClose]] means to not ignore signals
+ *  - [[IgnoreCancel]] means to not react to cancellation of the receiving
+ *    side unless the sending side has already completed
+ *  - [[IgnoreComplete]] means to not react to the completion of the sending
+ *    side unless the receiving side has already canceled
+ *  - [[IgnoreBoth]] means to ignore the first termination signal—be that
+ *    cancellation or completion—and only act upon the second one
+ */
+sealed abstract class TLSClosing {
```
viktorklang: sealed does not protect against Java code inheriting from it.

akka-stream/src/main/scala/akka/stream/SslTlsOptions.scala
rkuhn: These are the shared pieces that would just create busywork to be duplicated.

akka-stream/src/main/scala/akka/stream/SslTlsOptions.scala
```diff
@@ -0,0 +1,255 @@
...
+  def ignoreCancel: Boolean
+  def ignoreComplete: Boolean
```
viktorklang: Why introduce these methods AND an ADT to describe the valid options?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
rkuhn: This is like a `.toScala` method that also has the language in the name, users won’t use this type much, I reckon (because we provide all the niceness in different ways already).

akka-stream/src/main/scala/akka/stream/SslTlsOptions.scala
```diff
@@ -0,0 +1,255 @@
...
+   * - `enabledCipherSuites` will be passed to `SSLEngine::setEnabledCipherSuites()`
+   * - `enabledProtocols` will be passed to `SSLEngine::setEnabledProtocols()`
+   * - `clientAuth` will be passed to `SSLEngine::setWantClientAuth()` or `SSLEngine.setNeedClientAuth()`, respectively
+   * - `sslParameters` will be passed to `SSLEngine::setSSLParameters()`
+   *
+   * Please note that passing `clientAuth = None` means that no change is done
+   * on client authentication requirements while `clientAuth = Some(ClientAuth.None)`
+   * switches off client authentication.
+   */
+  case class NegotiateNewSession(
```
viktorklang: final?

akka-stream/src/main/scala/akka/stream/SslTlsOptions.scala
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * Java API: Make a copy of this message with the given `enabledProtocols`.
+     */
+    @varargs
+    def withProtocols(p: String*) = copy(enabledProtocols = Some(p.toList))
+
+    /**
+     * Java API: Make a copy of this message with the given [[TLSClientAuth]] setting.
+     */
+    def withClientAuth(ca: TLSClientAuth) = copy(clientAuth = Some(ca))
```
viktorklang: why copy if not changed? (goes for all of the copy invocations)

akka-stream/src/main/scala/akka/stream/SslTlsOptions.scala
```diff
@@ -0,0 +1,255 @@
...
+
+    /**
+     * Java API: Make a copy of this message with the given [[SSLParameters]].
+     */
+    def withParameters(p: SSLParameters) = copy(sslParameters = Some(p))
+  }
+
+  object NegotiateNewSession extends NegotiateNewSession(None, None, None, None) {
```
viktorklang: not case object?

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
rkuhn: It is a little unconventionally named, though. But I’m not sure that is enough reason to change it at the last moment.

akka-stream/src/main/scala/akka/stream/SslTlsOptions.scala
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * Java API: obtain the default value (which will leave the SSLEngine’s
+     * settings unchanged).
+     */
+    def withDefaults = this
```
viktorklang: why no type?

akka-stream/src/main/scala/akka/stream/SslTlsOptions.scala
```diff
@@ -0,0 +1,255 @@
...
+
+  }
+
+  /**
+   * Send the given [[akka.util.ByteString]] across the encrypted session to the
+   * peer.
+   */
+  case class SendBytes(bytes: ByteString) extends SslTlsOutbound
```
viktorklang: final

akka-stream/src/main/scala/akka/stream/impl/io/TlsModule.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import akka.stream._
+import akka.stream.impl.StreamLayout.{ CompositeModule, Module }
+import akka.stream.TLSProtocol._
+import akka.util.ByteString
+
+/**
+ * INTERNAL API.
+ */
+private[akka] case class TlsModule(plainIn: Inlet[SslTlsOutbound], plainOut: Outlet[SslTlsInbound],
```
viktorklang: final?

akka-stream/src/main/scala/akka/stream/javadsl/TLS.scala
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * This object holds simple wrapping [[akka.stream.scaladsl.BidiFlow]] implementations that can
+ * be used instead of [[TLS]] when no encryption is desired. The flows will
+ * just adapt the message protocol by wrapping into [[SessionBytes]] and
+ * unwrapping [[SendBytes]].
+ */
+object TLSPlacebo {
+
+  def getInstance(): javadsl.BidiFlow[SslTlsOutbound, ByteString, ByteString, SessionBytes, NotUsed] = forJava
```
viktorklang: No scaladoc?

akka-stream/src/main/scala/akka/stream/scaladsl/TLS.scala
```diff
@@ -0,0 +1,133 @@
...
+    })
+}
+
+import java.security.Principal
+import java.security.cert.Certificate
+import javax.net.ssl.{ SSLPeerUnverifiedException, SSLSession }
+
+/** Allows access to an SSLSession with Scala types */
+trait ScalaSessionAPI {
+  def session: SSLSession
```
viktorklang: No scaladoc?

akka-stream/src/main/scala/akka/stream/SslTlsOptions.scala
```diff
@@ -0,0 +1,255 @@
...
+   * Plaintext bytes emitted by the SSLEngine are received over one specific
+   * encryption session and this class bundles the bytes with the SSLSession
+   * object. When the session changes due to renegotiation (which can be
+   * initiated by either party) the new session value will not compare equal to
+   * the previous one.
+   *
+   * The Java API for getting session information is given by the SSLSession object,
+   * the Scala API adapters are offered below.
+   */
+  case class SessionBytes(session: SSLSession, bytes: ByteString) extends SslTlsInbound with scaladsl.ScalaSessionAPI
```
viktorklang: final?

viktorklang:
Cursory review done.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala
viktorklang: Consistency is dog

akka-stream/src/main/scala/akka/stream/SslTlsOptions.scala
viktorklang: Still inconsistent. Remember that the API is going to be used way more than it is written, so saving busywork for a few but creating inconsistencies for many doesn't seem like a great tradeoff IMO.

rkuhn: Which inconsistencies do you mean? Akka Streams is completely built like this. The Akka HTTP model is different, granted, but we cannot retroactively apply that principle to Streams now.

viktorklang: I'm not sure I follow, because I could swear Akka Streams has a scaladsl and a javadsl package?

rkuhn: Yes, and a `akka.stream` package for the shared things—that is a difference to how HTTP works. We don’t have separate Materializer types for Java and Scala, for example. Just like we don’t see any need to use different TLSClientAuth types; what would be the benefit?

viktorklang: Could be fine, but tell me this: what's the mnemonic for the users? (i.e. where should they look?)

rkuhn: check the new QuickStart guide:
~~~java
import akka.stream.*;
import akka.stream.javadsl.*;
~~~
This is how you start a project.

rkuhn:
I merged this manually.

Issue #19748, opened 2016-02-11
Akka Remoting hangs on empty connections instead of timing out


drewhk:
Simple reproducer:

```scala
val remote = Tcp().outgoingConnection("127.0.0.1", 2552)
Source.maybe[ByteString].via(remote).log("FUZZER: ").to(Sink.ignore).run()
```

The connections just stays around in limbo without timing out.

While this one:

```scala
(Source.single(ByteString(0, 0, 0, 1, 0)) ++ Source.maybe).via(remote).log("FUZZER ").to(Sink.ignore).run
```

fails properly with:

```
akka.remote.transport.AkkaProtocolException: Error while decoding incoming Akka PDU of length: 1
Caused by: akka.remote.transport.PduCodecException: Decoding PDU failed.
        at akka.remote.transport.AkkaPduProtobufCodec$.decodePdu(AkkaPduCodec.scala:167)
        at akka.remote.transport.ProtocolStateActor.akka$remote$transport$ProtocolStateActor$$decodePdu(AkkaProtocolTransport.scala:580)
        at akka.remote.transport.ProtocolStateActor$$anonfun$3.applyOrElse(AkkaProtocolTransport.scala:375)
        at akka.remote.transport.ProtocolStateActor$$anonfun$3.applyOrElse(AkkaProtocolTransport.scala:343)
        at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
        at akka.actor.FSM$class.processEvent(FSM.scala:604)
        at akka.remote.transport.ProtocolStateActor.processEvent(AkkaProtocolTransport.scala:269)
        at akka.actor.FSM$class.akka$actor$FSM$$processMsg(FSM.scala:598)
        at akka.actor.FSM$$anonfun$receive$1.applyOrElse(FSM.scala:592)
        at akka.actor.Actor$class.aroundReceive(Actor.scala:467)
        at akka.remote.transport.ProtocolStateActor.aroundReceive(AkkaProtocolTransport.scala:269)
        at akka.actor.ActorCell.receiveMessage(ActorCell.scala:516)
        at akka.actor.ActorCell.invoke(ActorCell.scala:487)
        at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:238)
        at akka.dispatch.Mailbox.run(Mailbox.scala:220)
```

drewhk:
[...]

PR #19811, opened 2016-02-17
=rem #18339 Use explicit handshake timeout (backport+validation)

drewhk:
- instead of using transport failure detector
- add a new config property akka.remote.handshake-timeout, but
  for netty.tcp and netty.ssl the existing netty.tcp.connection-timeout
  setting will be used
- add test of the timeouts
- mima filter for internal ProtocolStateActor
  (cherry picked from commit 94896e8)


drewhk:
Refs https://github.com/akka/akka/issues/19748

drewhk:
Hm, ClusterClientSpec

drewhk:
Hm this fails locally for me as well. I need to check 2.3 branch without this change.

drewhk:
Ok, it seems to fail sometimes without this fix, too.

Issue #19809, opened 2016-02-16
make macro-paradise version configurable as well


rkuhn:
2.12.0-M3 needs a newer one than master currently uses

PR #19815, opened 2016-02-17
=htc make HTTP compile under Scala 2.12.0-M3

ktoso:
Makes Akka HTTP compile under `2.12.0-M3`

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/CacheDirective.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/CacheDirective.scala
@@ -68,7 +68,7 @@ object CacheDirectives {
 
   // http://tools.ietf.org/html/rfc7234#section-5.2.1.4
   case object `no-cache` extends SingletonValueRenderable with RequestDirective with ResponseDirective {
-    @varargs def apply(fieldNames: String*): `no-cache` = apply(immutable.Seq(fieldNames: _*))
+    def apply(fieldNames: String*): `no-cache` = new `no-cache`(immutable.Seq(fieldNames: _*))
   }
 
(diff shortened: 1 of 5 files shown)


drewhk:
LGTM

ktoso:
Though it'll break 2.11 (the macro...) in this state.

ktoso:
Updated PR to survive on both 2.11 and 2.12

2m:
LGTM. Do the akka-parsing changes come from the parboiled2 upstream?

ktoso:
Nope, it's a fix I found on my own / with @jrudolph. It possibly should be also in parboiled2

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/CacheDirective.scala
```diff
@@ -96,12 +96,12 @@ object CacheDirectives {
 
   // http://tools.ietf.org/html/rfc7234#section-5.2.2.6
   final case class `private`(fieldNames: immutable.Seq[String]) extends FieldNamesDirective with ResponseDirective
-  object `private` {
-    @varargs def apply(fieldNames: String*): `private` = apply(immutable.Seq(fieldNames: _*))
+  case object `private` {
```
jrudolph: `case object` is not strictly needed for `private`. In the `no-cache` case the companion object can actually be used as a standalone value.

ktoso: Thanks, good catch - removed `case`

ktoso:
Addressed comments and rebased, will merge after validation

drewhk:
LGTM

2m:
Fixes #19809.

ktoso:
It's time to merge this fellow ;)

Issue #19766, opened 2016-02-12
GZipInputStream leaking off-heap memory


412b:
At least in protobuf serializers (in akka-cluster and akka-cluster-tools, hadn't time to check others) GzipInputStream is not closed at all (and the whole decompress looks just copy-pasted), due to GzipInputStream's nature that leads to off-heap memory leak with **every** deserialized message.

[akka/cluster/protobuf/ClusterMessageSerializer.scala](https://github.com/akka/akka/blob/master/akka-cluster/src/main/scala/akka/cluster/protobuf/ClusterMessageSerializer.scala#L102)
[akka/cluster/pubsub/protobuf/DistributedPubSubMessageSerializer.scala](https://github.com/akka/akka/blob/master/akka-cluster-tools/src/main/scala/akka/cluster/pubsub/protobuf/DistributedPubSubMessageSerializer.scala#L102)

the fix is trivial.
should I submit a PR for it or one can do it just during a break?

drewhk:
It would be nice if you would publish a PR. It is somewhat weird though that the unzipper does not free memory after it finished processing the passed InputStream. Have you actually confirmed the leak? We should add a close anyway though, I just wanted to assess the impact.

drewhk:
Yeah, looking at that class (GZipInputStream) it does not seem to call Inflater.end().

412b:
@drewhk it's life-time long and well-known story in java. GZipInputStream uses Inflater internally and native zlib is used. Inflater frees up memory only on explicit call to end() or during finalization (finalize() contains only call to end()). GZipInputStream extends InflaterInputStream and there Inflater's end() is only used in close(), so GZipInputStream should always be explicitly closed.

Java heap footprint is very small and is not creating enough heap pressure for a GC, so eventually you can end up with malloc's fail for zlib and OutOfMemoryError. That's the worst case scenario.
More real-life scenario (especially with scala as it creates enough heap pressure) is that you'll have long-living garbage (until finalizer's run and finalize() is called in Inflater).

412b:
@drewhk you were faster than my opus typing fingers :)

drewhk:
Thanks for reporting, this is indeed a serious problem, but thankfully easy to fix. We will look at this on Monday latest.

patriknw:
That is indeed a good finding. It's probably my bad, thanks.

patriknw:
This was also backported to 2.3

drewhk:
this has been already backported

PR #19821, opened 2016-02-18
GZIPInputStream is not closed at all  at places (for validation)

patriknw:
which leads to off-heap memory leak with every deserialized message.

GZIPInputStream uses Inflater internally (so also native zlib). Inflater frees up memory only on explicit call to end() or during finalization (finalize() contains only call to end()), so GZIPInputStream should always be explicitly closed.

As native libraries are used a non-scalaish try-finally is used to avoid off-heap memory leak for GZIPInputStream and GZIPOutputStream in case of exceptions.

(cherry picked from commit 35ffc492263a6f758e7f05054ba8264625ae5dae)


patriknw:
Refs #19766

viktorklang:
:+1:

PR #19825, opened 2016-02-18
=htc make basic credentials token method return correct token

aosagie:
"cookie" is an array of bytes. Calling "toString" on that returns an identity hash instead of the actual token.

Apologies for not adding an accompanying test, wasn't too sure on where/how to do it. Somewhere in here perhaps?: https://github.com/akka/akka/blob/10d3af14785970ba584e15effa4b7472fc1f39d5/akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCredentials.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCredentials.scala
@@ -27,9 +27,9 @@ final case class BasicHttpCredentials(username: String, password: String) extend
   }
   def render[R <: Rendering](r: R): r.type = r ~~ "Basic " ~~ cookie
 
-  def scheme: String = "Basic"
-  def token = cookie.toString
-  def params = Map.empty
+  override def scheme: String = "Basic"
+  override def token: String = String.valueOf(cookie)
+  override def params: Map[String, String] = Map.empty
 }
 
 object BasicHttpCredentials {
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCredentials.scala
```diff
@@ -28,7 +28,7 @@ final case class BasicHttpCredentials(username: String, password: String) extend
   def render[R <: Rendering](r: R): r.type = r ~~ "Basic " ~~ cookie
 
   def scheme: String = "Basic"
-  def token = cookie.toString
+  def token = String.valueOf(cookie)
```
viktorklang: Please also add explicit return type, and an `override` modifier.

akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/HttpCredentials.scala
```diff
@@ -28,7 +28,7 @@ final case class BasicHttpCredentials(username: String, password: String) extend
...
   def params = Map.empty
```
viktorklang: Please also add explicit return type, and an `override` modifier.

2beaucoup:
Would be nice to have a test. Could you add it [here](https://github.com/akka/akka/blob/10d3af14785970ba584e15effa4b7472fc1f39d5/akka-http-core/src/test/scala/akka/http/impl/model/parser/HttpHeaderSpec.scala#L131)?
```scala
BasicHttpCredentials("Aladdin", "open sesame").token === "..."
```

aosagie:
@2beaucoup Sure, === just returns a boolean however, so I used shouldEqual.

2beaucoup:
@aosagie :+1: (I meant `should ===(....)`).

ktoso:
LGTM, thanks!

johanandren:
OK TO TEST

johanandren:
LGTM

ktoso:
Thanks @aosagie !

ktoso:
Resolves https://github.com/akka/akka/issues/19839

PR #19832, opened 2016-02-19
add Twitter Chill to list of External Akka Serializers

stefan-mees:
We were running in quite some trouble when using the Akka-kryo Serialization Extension by Roman Levenstein. https://github.com/romix/akka-kryo-serialization/issues/75

Twitters Scala extensions for Kryo `Chill` seems to be much more stable, so i would suggest to add it to the list of External Akka Serializers

### DIFF
--- a/akka-docs/rst/scala/serialization.rst
+++ b/akka-docs/rst/scala/serialization.rst
@@ -212,4 +212,4 @@ External Akka Serializers
 `Akka-kryo by Roman Levenstein <https://github.com/romix/akka-kryo-serialization>`_
 
 
-`Twitter Chill Scala extensions for Kryo <https://github.com/twitter/chill>`_
+`Twitter Chill Scala extensions for Kryo (based on Akka Version 2.3.x but due to backwards compatibility of the Serializer Interface this extension also works with 2.4.x) <https://github.com/twitter/chill>`_
--- a/akka-docs/rst/scala/serialization.rst
+++ b/akka-docs/rst/scala/serialization.rst
@@ -210,3 +210,6 @@ External Akka Serializers
 
 
 `Akka-kryo by Roman Levenstein <https://github.com/romix/akka-kryo-serialization>`_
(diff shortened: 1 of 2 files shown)


typesafehub-validator:
Hi @stefan-mees,

ktoso:
Right, makes sense to include those in the list :+1:

akka-docs/rst/scala/serialization.rst
```diff
@@ -210,3 +210,6 @@ External Akka Serializers
 
 
 `Akka-kryo by Roman Levenstein <https://github.com/romix/akka-kryo-serialization>`_
+
+
+`Twitter Chill Scala extensions for Kryo <https://github.com/twitter/chill>`_
```
ktoso: Could you please also add it to `akka-docs/rst/java/serialization.rst`?

patriknw:
The problem with Chill is that it has always been severely outdated when it comes to Akka versions. Is that fixed now? Is it still on Akka 2.3? https://github.com/twitter/chill/blob/v0.8.0/build.sbt#L157

stefan-mees:
@ktoso added to Java documentation as well

@patriknw you are right they are still on Akka 2.3, but the dependency scope is `provided` which should not cause any dependency trouble with newer Akka versions, as long as the Serialization Provider Interface is backwards compatible

drewhk:
@stefan-mees have you tried it out though?

stefan-mees:
@drewhk yes we are using it in our project with Akka version 2.4.1

drewhk:
Ok, then I don't se anything against including it. Maybe add a short note that it is based on 2.3.x but due to backwards compatibility of 2.4.x it works with that branch as well.

ktoso:
+1 to adding a note like @drewhk suggests, and then let's merge it. Thanks!

ktoso:
Ok rewording, thanks!

Issue #19818, opened 2016-02-17
Flow.transform should be deprected too


hepin1989:
`def transform[T](mkStage: () ⇒ Stage[Out, T]): Repr[T] `
should be deprecated too as `Stage` is deprecated now.

viktorklang:
Nice catch @hepin1989!

Would you be able to open a PR to address this?

agolubev:
@viktorklang, is it only about marking method as deprecated?

ktoso:
Yeah, since we can't remove it now since the module is non experimental.

PR #19835, opened 2016-02-20
=str #19818 Flow.transform should be deprected too

agolubev:
Ref #19818

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
@@ -991,6 +991,7 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
    * This operator makes it possible to extend the `Flow` API when there is no specialized
    * operator that performs the transformation.
    */
+  @deprecated("Use via(GraphStage) instead.", "2.4.3")
   def transform[U](mkStage: function.Creator[Stage[Out, U]]): javadsl.Flow[In, U, Mat] =
     new Flow(delegate.transform(() ⇒ mkStage.create()))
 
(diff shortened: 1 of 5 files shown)


johanandren:
OK TO TEST

johanandren:
LGTM

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -991,6 +991,7 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
    * This operator makes it possible to extend the `Flow` API when there is no specialized
    * operator that performs the transformation.
    */
+  @deprecated("As Stage is deprecated.", "2.4.2")
```
johanandren: Will be as of 2.4.3 though. We can't really deprecate it after the fact?

hepin1989:
should we add a new method
`def transform[T](mkStage: () ⇒ GraphStage[FlowShape[Out,T]]): Repr[T]`

johanandren:
No, `GraphStages` are meant to be re-usable blueprint constructs, we should not endorse writing them in only-usable-once ways.

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
viktorklang: And, I think the message should guide the user about what to do instead.

agolubev: And, I think the message should guide the user about what to do instead.>

Is it FlowOps.via?

johanandren: Maybe "Use via(GraphStage) instead"?

agolubev: Ok

johanandren:
LGTM!

drewhk:
LGTM

Issue #19390, opened 2016-01-09
Create Sink.recent that allows to read latest element received from stream


agolubev:
I'm thinking about something similar to Sink.queue, but instead of SinkQueue `recent` will return monitor object that has latest element received from stream as well as stream status.

```scala
def recent[T](): Sink[T, SinkMonitor[T]]

trait SinkMonitor[T]{
   getState():StreamState[T]
}

sealed trait StreamState[+T]

case object NonInitialized  extends StreamState[Nothing]
case object Initialized  extends StreamState[Nothing]
final case class Recived[+T](result: T) extends StreamState[T]
final case class Failure(cause: Throwable) extends StreamState[Nothing]
case object StreamCompleted  extends StreamState[Nothing]
```
Use case I'm thinking is to send some statistics to such sink, like number of requests or throughput.

let me know if this makes sense.

agolubev:
also such sink can have initialization-termination notifications:
```StreamState.terminationNotification():Future[Unit]``` and ```StreamState.initializationNotification():Future[Unit]```
similarly to #18565 termination notification combinator

rkuhn:
Yes, this sounds useful: materializing to something that has a `get()` method and behaves like an AtomicReference that is updated from the stream.

talp-traiana:
Is anyone working on this, or should I give it a shot?

rkuhn:
no, it’s available as far as I can see, thanks!

talp-traiana:
So I've been thinking about this a bit, and was wondering if it wouldn't be more useful to provide some sort of composite Sink (or Source, or any Flow for that matter) that wraps a sink and allows monitoring of the inner sink.
Something like:
def recent[T, M](inner: Sink[T, M]): Sink[T, (M, SinkMonitor[T])

What do you think?

drewhk:
@talp-traiana well, create a flow, and then decorating a Sink becomes just an application for that Flow.

talp-traiana:
@drewhk Not sure I understand what you mean (or maybe I don't understand what @agolubev was asking for).

The way I understand it, if I have a sink (e.g.: that writes messages to RabbitMQ) and I want to be able to monitor it (know what the last processed message is, check the sink state, etc.), how would I do it with API suggested in the OP? Won't I end up with 2 sinks, and no useful way to connect them (broadcasting messages to both sinks will give different results than wrapping)?

drewhk:
[...]

PR #19498, opened 2016-01-18
+str #18565 termination notification combinator

agolubev:
watchTerminationMat() was introduced as making watchTermination() as Keep.right by default.

PR #19836, opened 2016-02-21
+str #19390 Add FlowMonitor

ghost:
Add flow that materializes into object that provides the status of the flow (last received message, failures, etc.)

Still no tests/docs.

### DIFF
--- a/akka-docs/rst/java/stream/stages-overview.rst
+++ b/akka-docs/rst/java/stream/stages-overview.rst
@@ -1067,3 +1067,15 @@ The stage otherwise passes through elements unchanged.
 
 **completes** when upstream completes
 
+monitor
+^^^^^^^
+Materializes to a ``FlowMonitor`` that monitors messages flowing through or completion of the stage. The stage otherwise
+passes through elements unchanged. Note that the ``FlowMonitor`` inserts a memory barrier every time it processes an
+event, and may therefore affect performance.
+
+**emits** when upstream emits an element
+
+**backpressures** when downstream **backpressures**
+
+**completes** when upstream completes
+
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMonitorSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMonitorSpec.scala
@@ -1,0 +1,73 @@
+/**
+  * Copyright (C) 2015-2016 Lightbend Inc. <http://www.lightbend.com>
+  */
+package akka.stream.scaladsl
+
+import akka.stream.testkit.scaladsl.{ TestSource, TestSink }
+import akka.stream.{ ActorMaterializer, ActorMaterializerSettings }
+import akka.testkit.AkkaSpec
+import akka.stream.FlowMonitorState._
+
+import scala.concurrent.duration._
+
+class FlowMonitorSpec extends AkkaSpec {
+
+  val settings = ActorMaterializerSettings(system)
+
+  implicit val materializer = ActorMaterializer(settings)
+
+  "A FlowMonitor" must {
+    "return Finished when stream is completed" in {
+      val ((source, monitor), sink) =
+        TestSource.probe[Any].monitor()(Keep.both).toMat(TestSink.probe[Any])(Keep.both).run()
+      source.sendComplete()
+      awaitAssert(monitor.state == Finished, 3.seconds)
+      sink.expectSubscriptionAndComplete()
+    }
+
+    "return Finished when stream is cancelled from downstream" in {
+      val ((source, monitor), sink) =
+        TestSource.probe[Any].monitor()(Keep.both).toMat(TestSink.probe[Any])(Keep.both).run()
+      sink.cancel()
+      awaitAssert(monitor.state == Finished, 3.seconds)
+    }
+
+    "return Failed when stream fails, and propagate the error" in {
+      val ((source, monitor), sink) =
+        TestSource.probe[Any].monitor()(Keep.both).toMat(TestSink.probe[Any])(Keep.both).run()
+      val ex = new Exception("Source failed")
+      source.sendError(ex)
+      awaitAssert(monitor.state == Failed(ex), 3.seconds)
+      sink.expectSubscriptionAndError(ex)
+    }
+
+    "return Initialized for an empty stream" in {
+      val ((source, monitor), sink) =
+        TestSource.probe[Any].monitor()(Keep.both).toMat(TestSink.probe[Any])(Keep.both).run()
+      awaitAssert(monitor.state == Initialized, 3.seconds)
+      source.expectRequest()
+      sink.expectSubscription()
+    }
+
+    "return Received after receiving a message" in {
+      val ((source, monitor), sink) =
+        TestSource.probe[Any].monitor()(Keep.both).toMat(TestSink.probe[Any])(Keep.both).run()
+      val msg = "message"
+      source.sendNext(msg)
+      sink.requestNext(msg)
+      awaitAssert(monitor.state == Received(msg), 3.seconds)
+    }
+
+    // Check a stream that processes StreamState messages specifically, to make sure the optimization in FlowMonitorImpl
+    // (to avoid allocating an object for each message) doesn't introduce a bug
+    "return Received after receiving a StreamState message" in {
+      val ((source, monitor), sink) =
+        TestSource.probe[Any].monitor()(Keep.both).toMat(TestSink.probe[Any])(Keep.both).run()
+      val msg = Received("message")
+      source.sendNext(msg)
+      sink.requestNext(msg)
+      awaitAssert(monitor.state == Received(msg), 3.seconds)
+    }
+
+  }
+}
--- a/akka-stream/src/main/scala/akka/stream/FlowMonitor.scala
+++ b/akka-stream/src/main/scala/akka/stream/FlowMonitor.scala
@@ -1,0 +1,63 @@
+/**
+ * Copyright (C) 2015-2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream
+
+import akka.stream.FlowMonitorState.StreamState
+
+/**
+ * Used to monitor the state of a stream
+ *
+ * @tparam T Type of messages passed by the stream
+ */
+trait FlowMonitor[+T] {
+  def state: StreamState[T]
+}
+
+object FlowMonitorState {
+  sealed trait StreamState[+U]
+
+  /**
+   * Stream was created, but no events have passed through it
+   */
+  case object Initialized extends StreamState[Nothing]
+
+  /**
+   * Java API
+   */
+  def initialized[U](): StreamState[U] = Initialized
+
+  /**
+   * Stream processed a message
+   *
+   * @param msg The processed message
+   */
+  final case class Received[+U](msg: U) extends StreamState[U]
+
+  /**
+   * Java API
+   */
+  def received[U](msg: U): StreamState[U] = Received(msg)
+
+  /**
+   * Stream failed
+   *
+   * @param cause The cause of the failure
+   */
+  final case class Failed(cause: Throwable) extends StreamState[Nothing]
+
+  /**
+   * Java API
+   */
+  def failed[U](cause: Throwable): StreamState[U] = Failed(cause)
+
+  /**
+   * Stream completed successfully
+   */
+  case object Finished extends StreamState[Nothing]
+
+  /**
+   * Java API
+   */
+  def finished[U]() = Finished
+}
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
@@ -9,6 +9,7 @@ import java.util.concurrent.atomic.{ AtomicBoolean, AtomicReference }
 import akka.actor.Cancellable
 import akka.dispatch.ExecutionContexts
 import akka.event.Logging
+import akka.stream.FlowMonitorState._
 import akka.stream._
 import akka.stream.scaladsl._
 import akka.stream.impl.Stages.DefaultAttributes
@@ -152,6 +153,72 @@ object GraphStages {
 
   def terminationWatcher[T]: GraphStageWithMaterializedValue[FlowShape[T, T], Future[Done]] =
     TerminationWatcher.asInstanceOf[GraphStageWithMaterializedValue[FlowShape[T, T], Future[Done]]]
+
+  private class FlowMonitorImpl[T] extends AtomicReference[Any](Initialized) with FlowMonitor[T] {
+    override def state = get match {
+      case s: StreamState[_] ⇒ s.asInstanceOf[StreamState[T]]
+      case msg               ⇒ Received(msg.asInstanceOf[T])
+    }
+  }
+
+  private class MonitorFlow[T] extends GraphStageWithMaterializedValue[FlowShape[T, T], FlowMonitor[T]] {
+    val in = Inlet[T]("FlowMonitor.in")
+    val out = Outlet[T]("FlowMonitor.out")
+    val shape = FlowShape.of(in, out)
+
+    override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, FlowMonitor[T]) = {
+      val monitor: FlowMonitorImpl[T] = new FlowMonitorImpl[T]
+
+      val logic: GraphStageLogic = new GraphStageLogic(shape) {
+        setHandler(in, new InHandler {
+          override def onPush(): Unit = {
+            val msg = grab(in)
+            push(out, msg)
+            monitor.set(if (msg.isInstanceOf[StreamState[_]]) Received(msg) else msg)
+          }
+          override def onUpstreamFinish(): Unit = {
+            super.onUpstreamFinish()
+            monitor.set(Finished)
+          }
+          override def onUpstreamFailure(ex: Throwable): Unit = {
+            super.onUpstreamFailure(ex)
+            monitor.set(Failed(ex))
+          }
+        })
+        setHandler(out, new OutHandler {
+          override def onPull(): Unit = pull(in)
+          override def onDownstreamFinish(): Unit = {
+            super.onDownstreamFinish()
+            monitor.set(Finished)
+          }
+        })
+
+        override def toString = "MonitorFlowLogic"
+      }
+
+      (logic, monitor)
+    }
+
+    override def toString = "MonitorFlow"
+  }
+
+  def monitor[T]: GraphStageWithMaterializedValue[FlowShape[T, T], FlowMonitor[T]] =
+    new MonitorFlow[T]
+
+  private object TickSource {
+    class TickSourceCancellable(cancelled: AtomicBoolean) extends Cancellable {
+      private val cancelPromise = Promise[Done]()
+
+      def cancelFuture: Future[Done] = cancelPromise.future
+
+      override def cancel(): Boolean = {
+        if (!isCancelled) cancelPromise.trySuccess(Done)
+        true
+      }
+
+      override def isCancelled: Boolean = cancelled.get()
+    }
+  }
 
   final class TickSource[T](initialDelay: FiniteDuration, interval: FiniteDuration, tick: T)
     extends GraphStageWithMaterializedValue[SourceShape[T], Cancellable] {
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
@@ -6,8 +6,8 @@ package akka.stream.javadsl
 import akka.{ NotUsed, Done }
 import akka.event.LoggingAdapter
 import akka.japi.{ function, Pair }
-import akka.stream.impl.{ ConstantFun, StreamLayout }
-import akka.stream.{ scaladsl, _ }
+import akka.stream.impl.{ConstantFun, StreamLayout}
+import akka.stream._
 import akka.stream.stage.Stage
 import org.reactivestreams.Processor
 import scala.annotation.unchecked.uncheckedVariance
@@ -1678,6 +1678,15 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
    */
   def watchTermination[M]()(matF: function.Function2[Mat, CompletionStage[Done], M]): javadsl.Flow[In, Out, M] =
     new Flow(delegate.watchTermination()((left, right) ⇒ matF(left, right.toJava)))
+
+  /**
+    * Materializes to `FlowMonitor[Out]` that allows monitoring of the the current flow. All events are propagated
+    * by the monitor unchanged. Note that the monitor inserts a memory barrier every time it processes an
+    * event, and may therefor affect performance.
+    * The `combine` function is used to combine the `FlowMonitor` with this flow's materialized value.
+    */
+  def monitor[M]()(combine: function.Function2[Mat, FlowMonitor[Out], M]): javadsl.Flow[In, Out, M] =
+    new Flow(delegate.monitor()(combinerToScala(combine)))
 
   /**
    * Delays the initial element by the specified duration.
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
@@ -10,7 +10,7 @@ import akka.actor.{ ActorRef, Cancellable, Props }
 import akka.event.LoggingAdapter
 import akka.japi.{ Pair, Util, function }
 import akka.stream._
-import akka.stream.impl.{ ConstantFun, StreamLayout }
+import akka.stream.impl.{ConstantFun, StreamLayout, SourceQueueAdapter}
 import akka.stream.stage.Stage
 import org.reactivestreams.{ Publisher, Subscriber }
 import scala.annotation.unchecked.uncheckedVariance
@@ -23,7 +23,6 @@ import scala.compat.java8.OptionConverters._
 import java.util.concurrent.CompletionStage
 import java.util.concurrent.CompletableFuture
 import scala.compat.java8.FutureConverters._
-import akka.stream.impl.SourceQueueAdapter
 
 /** Java API */
 object Source {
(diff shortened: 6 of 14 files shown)


johanandren:
Refs #19390

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -69,6 +66,9 @@ object Flow {
     sink: Graph[SinkShape[I], M1], source: Graph[SourceShape[O], M2],
     combine: function.Function2[M1, M2, M]): Flow[I, O, M] =
     new Flow(scaladsl.Flow.fromSinkAndSourceMat(sink, source)(combinerToScala(combine)))
+
+  def monitorFlow[T](): Flow[T, T, FlowMonitor[T]] =
```
johanandren: I think it should have the same name in all three places. `monitor` is good in my opinion.

johanandren:
Needs test coverage, doc on the public methods and mention in stages-overview for Scala and Java in the docs.

akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala
```diff
@@ -342,4 +338,6 @@ object Sink {
   def queue[T](): Sink[T, SinkQueue[T]] =
     Sink.fromGraph(new QueueSink())
 
+  def monitored[T, M](inner: Sink[T, M]): Sink[T, (M, FlowMonitor[T])] =
```
johanandren: Actually, I don`t think it should be on `Sink` as `flow.to(sink)` becomes a sink.

talp-traiana: So what should I do with it? Leave it as is? Move it someplace? Remove it?

johanandren: I think you should remove it and keep it on `Flow` and instances of flow.

akka-actor/src/main/scala/akka/japi/JavaAPI.scala
```diff
@@ -60,7 +60,7 @@ trait Predicate[T] {
  * Additional tuple types for 3 to 22 values are defined in the `akka.japi.tuple` package, e.g. [[akka.japi.tuple.Tuple3]].
  */
 @SerialVersionUID(1L)
-case class Pair[A, B](first: A, second: B) {
+case class Pair[+A, +B](first: A, second: B) {
```
talp-traiana: I'm not sure about this... It feels like a pretty big change, that could affect a lot of things. On the other hand, I think it is right? I mean, Scala Tuples are covariant, and Pair is immutable so it should be OK...
This also brings up the question of what to do with all the other japi.TupleN types...

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -1779,6 +1779,8 @@ final class Source[+Out, +Mat](delegate: scaladsl.Source[Out, Mat]) extends Grap
   def watchTermination[M]()(matF: function.Function2[Mat, CompletionStage[Done], M]): javadsl.Source[Out, M] =
     new Source(delegate.watchTermination()((left, right) ⇒ matF(left, right.toJava)))
 
+  def monitor: javadsl.Source[Out, Pair[Mat, FlowMonitor[Out]]] = viaMat(Flow.monitor[Out](), Keep.both[Mat, FlowMonitor[Out]])
```
johanandren: "Only on Flow" means, not here either. `Source.via(Flow)` = `Source` so no need to add this.

akka-actor/src/main/scala/akka/japi/JavaAPI.scala
johanandren: Why was this change introduced? This won't fly because the type is for Java, which does not have declaration-site variance.

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1934,6 +1935,8 @@ trait FlowOpsMat[+Out, +Mat] extends FlowOps[Out, Mat] {
   def watchTermination[Mat2]()(matF: (Mat, Future[Done]) ⇒ Mat2): ReprMat[Out, Mat2] =
     viaMat(GraphStages.terminationWatcher)(matF)
 
+  def monitor: ReprMat[Out, (Mat, FlowMonitor[Out])] = viaMat(Flow.monitor[Out])(Keep.both)
```
johanandren: Not sure it should be here either

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -322,6 +320,9 @@ object Flow {
    */
   def fromSinkAndSourceMat[I, O, M1, M2, M](sink: Graph[SinkShape[I], M1], source: Graph[SourceShape[O], M2])(combine: (M1, M2) ⇒ M): Flow[I, O, M] =
     fromGraph(GraphDSL.create(sink, source)(combine) { implicit b ⇒ (in, out) ⇒ FlowShape(in.in, out.out) })
+
+  def monitor[T]: Flow[T, T, FlowMonitor[T]] =
+    fromGraph(new MonitorFlow[T])
```
johanandren: I think it belongs in `FlowOps` which should make it available from the right places (and that's where all the other flow stages are available). This will make it available from instances of `Flow` and `Source` as those inherit FlowOps. 

I was wrong saying "on Flow", adding it on `FlowOps` will make it possible to `Flow[String].monitor` as well as `source.monitor`

akka-actor/src/main/scala/akka/japi/JavaAPI.scala
talp-traiana: javadsl.Flow.monitor wouldn't compile otherwise - Out is covariant in Flow/FlowOps but Pair isn't, so I can't return Pair<Out, FlowMonitor>

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
talp-traiana: I tried putting it in FlowOps, but then I couldn't change the materialized value.
Am I missing something?

talp-traiana:
@johanandren 
Sorry, I'm a little confused on where I should/shouldn't add these methods. Can you go over this list and say which of these should be implemented and which shouldn't?

- akkadsl.Flow object
- akkadsl.Flow (Flow class/instances)
- akkadsl.FlowOps (I don't think this works...)
- akkadsl.FlowOpsMat.monitor
- javadsl.Flow object (dual to the one in akkadsl.Flow)
- javadsl.Flow (class/instances, dual to akkadsl.Flow/FlowOps/FlowOpsMat)
- javadsl.Source (class/instances, dual to akkadsl.FlowOps/FlowOpsMat)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
johanandren: No, sorry, my bad, this _is_ the right place, I misread the folding the github view and thought it was on `RunnableGraph`. It's a flow that materializes a value - `FlowOpsMat` is right. 

Return wouldn't be a tuple but `ReprMat[Out, FlowMonitor]`, we would also want a `monitorMat` following the pattern of the other methods taking a function that allows you to choose how to handle the materialized value so that it can be called using `Keep.{left, right, both}`

johanandren:
Sure, probably my fault with some unfocused reviewing.

* `FlowOpsMat.monitor` as it will put it on instances of `Source` and `Flow` in the Scala API, note that it should materialize to your new `FlowMonitor` and not a tuple. It might be useful to also provide the  `monitorMat` version, I'm not quite sure, this allows to keep a previous materialized value in the flow. One could hovever just do `source.viaMat(Flow[T].monitor)(Keep.both)` for those cases so that's why I'm not sure it is needed.
* `javadsl.Source` and `Flow` - I think you need both those as that is where they are available from the Scala-DSL, there should be automatic tests that compare the two DSLs will fail if I'm incorrect about this.

talp-traiana:
OK, I removed all the methods from the objects, added monitorMat (doing it "the long way" from Java would've been pretty annoying, and most of the other DSL methods have *Mat equivalents, so why not?), and cleaned up a bit. What do you think?

(And I'm still missing docs+tests, I'll probably get to those tomorrow)

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
```diff
@@ -269,6 +270,57 @@ object GraphStages {
   def terminationWatcher[T]: GraphStageWithMaterializedValue[FlowShape[T, T], Future[Done]] =
     TerminationWatcher.asInstanceOf[GraphStageWithMaterializedValue[FlowShape[T, T], Future[Done]]]
 
+  private class FlowMonitorImpl[T] extends FlowMonitor[T] {
+    private val stateRef = new AtomicReference[StreamState[T]](Initialized)
```
drewhk: extend AtomicReference directly

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
```diff
@@ -269,6 +270,57 @@ object GraphStages {
...
+    def setState(state: StreamState[T]): Unit = stateRef.set(state)
+
+    override def state = stateRef.get()
+  }
```
drewhk: it might be possible to save on allocation of a Received for every message if you store it in an AnyRef field instead of type T, and wrapping it only back on read if needed. Roughly:
```scala
get() match {
  case s: StreamState => s
  case elem => Received(elem)
}

```

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
```diff
@@ -269,6 +270,57 @@ object GraphStages {
...
+        setHandler(in, new InHandler {
+          override def onPush(): Unit = {
+            val msg = grab(in)
+            push(out, msg)
+            monitor.setState(akka.stream.FlowMonitor.Received(msg))
+          }
+          override def onUpstreamFinish(): Unit = {
+            super.onUpstreamFinish()
+            monitor.setState(Finished)
+          }
```
drewhk: this must call finishStage otherwise there is a leak

drewhk:
tests are missing, but looks like a nice start!

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
drewhk: well, it trades a branch on read with an allocation on write, but I think it is worth it most of the time.

talp-traiana: What happens if T == StreamState? Are we OK with that?
I mean, it would be a shame to miss out on this optimization for such an edge case, but it still feels a bit problematic.

drewhk: hm, that can be parried by the onPush checking that if the grabbed elem is instance of StreamState and if it is then wrap it in a Received().

johanandren:
Looks good so far. Add tests, scaladoc on the methods (look at the other ones and follow that format) and mentions in stages-overview.rst for Java and Scala and it will be done.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
```diff
@@ -269,6 +270,59 @@ object GraphStages {
...
+        setHandler(in, new InHandler {
+          override def onPush(): Unit = {
+            val msg = grab(in)
+            push(out, msg)
+            monitor.set(if (msg.isInstanceOf[StreamState[_]]) Received(msg) else msg)
+          }
+          override def onUpstreamFinish(): Unit = {
+            super.onUpstreamFinish()
+            monitor.set(Finished)
+            completeStage()
```
talp-traiana: @drewhk 
Now that I look at it, I'm not sure these calls are needed. Doesn't super.onFinish take care of that?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMonitorSpec.scala
```diff
@@ -0,0 +1,68 @@
...
+      val ((source, monitor), sink) =
+        TestSource.probe[Any].monitorMat()(Keep.both).toMat(TestSink.probe[Any])(Keep.both).run()
+      awaitCond(monitor.state == Initialized, 3.seconds)
+      source.expectRequest()
+      sink.expectSubscription()
+    }
+
+    "return Received after receiving a message" in {
+      val ((source, monitor), sink) =
+        TestSource.probe[Any].monitorMat()(Keep.both).toMat(Sink.ignore)(Keep.both).run()
```
talp-traiana: @johanandren @drewhk 
I tried connecting a TestSink.probe instead of Sink.ignore so I could check that the message actually reached the sink, but it didn't work (the message was "stuck" somewhere?). Even when I removed the monitor, and connected the source to the sink directly, still nothing happened.
Any ideas why?

johanandren: Could it be because you forgot to signal demand from the sink before expecting any elements coming through?

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
drewhk: yes, they do. I had here a wrong comment which I deleted immediately after. Have you been fixing my ghost comment :)

akka-docs/rst/java/stream/stages-overview.rst
```diff
@@ -1067,3 +1067,15 @@ The stage otherwise passes through elements unchanged.
 
 **completes** when upstream completes
 
+monitor
+^^^^^^^
+Materializes to a ``FlowMonitor`` that monitors messages flowing through or completion of the stage. The stage otherwise
+passes through elements unchanged. Note that the ``FlowMonitor`` inserts a memory barrier every time it processes an
+event, and may therefor affect performance.
```
johanandren: :+1:  that you mention this!

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
talp-traiana: Yeah, looks like it... I just saw the mail notification and did it. ^_^

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMonitorSpec.scala
talp-traiana: Doh, for some reason I thought the sink probe started with demand already seeded... Will fix it and push a new version.

johanandren:
LGTM

talp-traiana:
I think someone needs to "OK TO TEST" this?

johanandren:
OK TO TEST

talp-traiana:
Thanks. once the tests pass I'll rebase and squash (unless there are any more comments?)

akka-docs/rst/java/stream/stages-overview.rst
drewhk: "therefore"

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMonitorSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+        TestSource.probe[Any].monitorMat()(Keep.both).toMat(TestSink.probe[Any])(Keep.both).run()
+      val ex = new Exception("Source failed")
+      source.sendError(ex)
+      awaitCond(monitor.state == Failed(ex), 3.seconds)
+      sink.expectSubscriptionAndError(ex)
+    }
+
+    "return Initialized for an empty stream" in {
+      val ((source, monitor), sink) =
+        TestSource.probe[Any].monitorMat()(Keep.both).toMat(TestSink.probe[Any])(Keep.both).run()
```
drewhk: you can use Source.maybe and Sink.ignore as well. Doesn't really matter though, just mentioning.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
```diff
@@ -269,6 +270,57 @@ object GraphStages {
...
+
+    override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, FlowMonitor[T]) = {
+      val monitor: FlowMonitorImpl[T] = new FlowMonitorImpl[T]
+
+      val logic: GraphStageLogic = new GraphStageLogic(shape) {
+        setHandler(in, new InHandler {
+          override def onPush(): Unit = {
+            val msg = grab(in)
+            push(out, msg)
+            monitor.set(if (msg.isInstanceOf[StreamState[_]]) Received(msg) else msg)
```
drewhk: a test case for this is missing

drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
talp-traiana: The last test-case doesn't take care of it (FlowMonitorSpec:60)?

drewhk: yes it does, my bad

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1933,6 +1933,22 @@ trait FlowOpsMat[+Out, +Mat] extends FlowOps[Out, Mat] {
     viaMat(GraphStages.terminationWatcher)(matF)
 
   /**
+    * Materializes to `FlowMonitor[Out]` that allows monitoring of the the current flow. All events are propagated
+    * by the monitor unchanged. Note that the monitor inserts a memory barrier every time it processes an
+    * event, and may therefor affect performance.
+    */
+  def monitor: ReprMat[Out, FlowMonitor[Out]] = monitorMat()(Keep.right)
```
agolubev: @drewhk, did we switch gears to allow combinators to have mat Keep.right by default? I remember we discussed this point at https://github.com/akka/akka/pull/19498#discussion_r50118558

>I see your point, but at some point we decided that the rules are:
Keep.right everywhere, except
runWith, foldWith, etc.
This makes the "matless" version not too useful though. would be interesting what others think

that's why watch watchTermination exists only with mat combinator as attribute.

drewhk: No, you are right, I think this is an oversight, at least I missed this here. We should be consistent and use Keep.left.

drewhk: @talp-traiana please change this to Keep.left

talp-traiana: @drewhk If we ignore the materialized value of the flow, then there's no point in adding the monitor. Maybe I should just leave the 'mat' version, like watchTermination?

drewhk: was thinking the same but I don't remember how we did it with watchTermination. I think it makes sense to have only the mat version.

talp-traiana: As far as I can tell from the code, there is only watchTermination, and it behaves like the mat version (i.e.: takes a combiner function)

talp-traiana:
@drewhk So can this be merged, or is there anything else?

drewhk:
I need another LGTM from someone else. We are in a company meeting this week, so things are a bit slower than usual.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMonitorSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+  val settings = ActorMaterializerSettings(system)
+
+  implicit val materializer = ActorMaterializer(settings)
+
+  "A FlowMonitor" must {
+    "return Finished when stream is completed" in {
+      val ((source, monitor), sink) =
+        TestSource.probe[Any].monitor()(Keep.both).toMat(TestSink.probe[Any])(Keep.both).run()
+      source.sendComplete()
+      awaitCond(monitor.state == Finished, 3.seconds)
```
patriknw: I prefer `awaitAssert` since it has better error reporting if it fails

akka-stream/src/main/scala/akka/stream/FlowMonitor.scala
```diff
@@ -0,0 +1,37 @@
...
+  def state: StreamState[T]
+}
+
+object FlowMonitor {
+  sealed trait StreamState[+U]
+
+  /**
+   * Stream was created, but no events have passed through it
+   */
+  case object Initialized extends StreamState[Nothing]
```
patriknw: we need a java api for this one, I'm not sure what naming convention we have in streams, but some kind of method is needed in `object FlowMonitor`:

```
/**
 * JAVA API
*/
def getInitializedInstance = Initialized
```

akka-stream/src/main/scala/akka/stream/FlowMonitor.scala
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * Stream failed
+   * @param cause The cause of the failure
+   */
+  final case class Failed(cause: Throwable) extends StreamState[Nothing]
+
+  /**
+   * Stream completed successfully
+   */
+  case object Finished extends StreamState[Nothing]
```
patriknw: java api

patriknw:
LGTM, aside from a few minor things

drewhk:
Needs a rebase and a few final fixes, then this is ready to merge!

talp-traiana:
@drewhk I fixed Patrik's issues in the latest commit. Is there anything else to fix (other than the rebase)?

drewhk:
Sorry, I missed that. I think this only needs a rebase now.

rkuhn:
fixes #19390

patriknw:
@talp-traiana needs a rebase (sorry for the inconvenience)

patriknw:
LGTM again

patriknw:
@drewhk is this ready?

drewhk:
I think it is yes, LGTM

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowMonitorSpec.scala
```diff
@@ -0,0 +1,70 @@
+package akka.stream.scaladsl
```
ktoso: Sorry for nitpicking but: missing license header

akka-stream/src/main/scala/akka/stream/FlowMonitor.scala
```diff
@@ -0,0 +1,60 @@
+package akka.stream
```
ktoso: here too

akka-stream/src/main/scala/akka/stream/FlowMonitor.scala
ktoso: note that this is addressed already – `initialized()` below

ktoso:
LGTM - nice one.
Just the missing license (we can add it). 
Shall we rebase or would you be able to @talp-traiana ?

talp-traiana:
@ktoso I just pushed the rebase, but hadn't noticed the comments regarding the license.

drewhk:
@talp-traiana see @ktoso's last comments like 
https://github.com/akka/akka/pull/19836#discussion_r58035320
https://github.com/akka/akka/pull/19836#discussion_r58035337

talp-traiana:
@drewhk @ktoso OK, added the license too.

ktoso:
Thanks a lot! We'll merge after validation

drewhk:
Hm:
```
[error]  * method monitor(scala.Function2)akka.stream.scaladsl.FlowOpsMat in trait akka.stream.scaladsl.FlowOpsMat is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]
```

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -1989,6 +1989,15 @@ trait FlowOpsMat[+Out, +Mat] extends FlowOps[Out, Mat] {
   def mapMaterializedValue[Mat2](f: Mat ⇒ Mat2): ReprMat[Out, Mat2]
 
   /**
+    * Materializes to `FlowMonitor[Out]` that allows monitoring of the the current flow. All events are propagated
+    * by the monitor unchanged. Note that the monitor inserts a memory barrier every time it processes an
+    * event, and may therefor affect performance.
+    * The `combine` function is used to combine the `FlowMonitor` with this flow's materialized value.
+    */
+  def monitor[Mat2]()(combine: (Mat, FlowMonitor[Out]) => Mat2): ReprMat[Out, Mat2] =
```
drewhk: this is at the wrong place. Move it to the FlowOps trait instead and then MiMa will be happy. FlowOpsMat cannot be changed due to binary compatibility.

rkuhn: No, the placement is determined by whether or not a combinator changes the materialized value type. MiMa will always complain and we will always need to add exceptions for new combinators.

drewhk: ok, I see. Is this safe because no-one is allowed to extend it?

patriknw: and it is safe to add those exceptions because user code will never extend FlowOps, right?

drewhk: We should update the comment though as it is not very helpful (it can be interpreted in various ways):
```
INTERNAL API: extending this trait is not supported under the binary compatibility rules for Akka.
```

rkuhn: If this comment needs improving then please go ahead—the intention is exactly that nobody is supposed to extend it outside Akka because otherwise we would be unable to ever extend the DSL again.

drewhk: I read it as you are not allowed to extend the trait as in: adding new stuff :) hence the confusion.

rkuhn: Ah, I see, I was so fixated on the `extends` keyword that I didn’t consider this interpretation :-)

What about:
~~~
INTERNAL API: this trait will be changed in binary-incompatible ways for classes that are derived from it! Do not implement this interface outside the Akka code base!

Binary compatibility is only maintained for callers of this trait’s interface.
~~~

rkuhn: (the same change must be made for both FlowOps and FlowOpsMat)

talp-traiana: Sorry, I'm confused. Should I Roland's comment to the FlowOps and FlowOpsMat traits?
And what about the MiMa exception? I did it once, but can't remember how...

patriknw: add to https://github.com/akka/akka/blob/master/project/MiMa.scala#L778
I think you'll find similar examples, and the error message from the build shows what you should add.

talp-traiana: OK, let's see if this works...

talp-traiana:
Well, that didn't work... I tried to run mima locally, and it passed, I guess I'm doing something wrong?

patriknw:
Something else got merged. I have rebased this in https://github.com/akka/akka/pull/20195

Issue #19840, opened 2016-02-22
Introduction section for HTTP


johanandren:
We are missing an easy-to-digest introduction to akka-http.

johanandren:
Feedback from Konrad on what not to forget: 
:point_up: [22 februari 2016 11:02](https://gitter.im/akka/dev?at=56cadcba7a66b5965f694b7d)

Issue #19740, opened 2016-02-10
optionMarshaller - is None returning a 200 what we really want?


ktoso:
Hi @sirthias @jrudolph,
I was reading this blogpost : http://danielasfregola.com/2016/02/07/how-to-build-a-rest-api-with-akka-http/

And realised that when we have an `Option[String]` for example, and we `complete` with a `None` the default `EmptyValue[]` mechanism kicks in and the response will be a `200` response. Is there some deeper reason behind that? I would have thought a 404 would be more appropriate to model the `None` case.

Code in question:

```scala
  implicit def optionMarshaller[A, B](implicit m: Marshaller[A, B], empty: EmptyValue[B]): Marshaller[Option[A], B] =
    Marshaller { implicit ec ⇒
      {
        case Some(value) ⇒ m(value)
        case None        ⇒ FastFuture.successful(Marshalling.Opaque(() ⇒ empty.emptyValue) :: Nil)
      }
    }
```

Let me know if there's a story behind the 200 response there or if we simply should fix it, thanks a lot in advance!

// Yup I'm back from vacation, was great (now a bit sick though), so back to hacking full speed now :-)

sirthias:
I don't think there is generally "correct" semantic for marshalling `None`.
Depending on the application `None` could rendered to a `200 Ok`, `204 No Content`, `404 Not Found` or any other kind of response.

Note that in the case you want a `404` response you can use the `rejectEmptyResponse` directive which was inherited from spray (http://spray.io/documentation/1.2.2/spray-routing/misc-directives/rejectEmptyResponse/) and is part of the `MiscDirectives`.

ktoso:
That's true as well that the right answer is "it depends" here hm...
Instead of changing behaviour I'll document how to change the none case by providing an `EmptyValue` implicitly and mention `rejectEmptyResponse` then.

ktoso:
Closing for now as obsolete

PR #19843, opened 2016-02-22
=doc #19840 Introduction docs for HTTP

johanandren:
Refs #19840 

I have only introduced more than one new example, re-using existing ones, I think the problem is more about discoverability and docs navigation than lack of docs, so I hope this intro-page helps with that.

I think Java will have to wait until the new Java API is in place.

### DIFF
--- a/akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
+++ b/akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
@@ -4,7 +4,7 @@
 
 package docs.http.scaladsl
 
-import akka.actor.{ ActorRef, ActorSystem }
+import akka.actor.{ Props, ActorRef, ActorSystem }
 import akka.event.LoggingAdapter
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.Http.ServerBinding
@@ -13,6 +13,7 @@ import akka.stream.ActorMaterializer
 import akka.stream.scaladsl.{ Flow, Sink }
 import akka.testkit.TestActors
 import org.scalatest.{ Matchers, WordSpec }
+import scala.io.StdIn
 import scala.language.postfixOps
 
 import scala.concurrent.{ ExecutionContext, Future }
@@ -31,7 +32,7 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
 
     implicit val system = ActorSystem()
     implicit val materializer = ActorMaterializer()
-    implicit val ec = system.dispatcher
+    implicit val executionContext = system.dispatcher
 
     val serverSource: Source[Http.IncomingConnection, Future[Http.ServerBinding]] =
       Http().bind(interface = "localhost", port = 8080)
@@ -49,7 +50,8 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
 
     implicit val system = ActorSystem()
     implicit val materializer = ActorMaterializer()
-    implicit val ec = system.dispatcher
+    // needed for the future onFailure in the end
+    implicit val executionContext = system.dispatcher
 
     val handler = get {
       complete("Hello world!")
@@ -60,7 +62,7 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
     val bindingFuture: Future[ServerBinding] =
       Http().bindAndHandle(handler, host, port)
 
-    bindingFuture onFailure {
+    bindingFuture.onFailure {
       case ex: Exception =>
         log.error(ex, "Failed to bind to {}:{}!", host, port)
     }
@@ -74,7 +76,8 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
   "binding-failure-handling" in compileOnlySpec {
     implicit val system = ActorSystem()
     implicit val materializer = ActorMaterializer()
-    implicit val ec = system.dispatcher
+    // needed for the future onFailure in the end
+    implicit val executionContext = system.dispatcher
 
     // let's say the OS won't allow us to bind to 80.
     val (host, port) = ("localhost", 80)
@@ -84,7 +87,7 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
       .to(handleConnections) // Sink[Http.IncomingConnection, _]
       .run()
 
-    bindingFuture onFailure {
+    bindingFuture.onFailure {
       case ex: Exception =>
         log.error(ex, "Failed to bind to {}:{}!", host, port)
     }
@@ -97,7 +100,7 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
   "incoming-connections-source-failure-handling" in compileOnlySpec {
     implicit val system = ActorSystem()
     implicit val materializer = ActorMaterializer()
-    implicit val ec = system.dispatcher
+    implicit val executionContext = system.dispatcher
 
     import Http._
     val (host, port) = ("localhost", 8080)
@@ -119,7 +122,7 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
   "connection-stream-failure-handling" in compileOnlySpec {
     implicit val system = ActorSystem()
     implicit val materializer = ActorMaterializer()
-    implicit val ec = system.dispatcher
+    implicit val executionContext = system.dispatcher
 
     val (host, port) = ("localhost", 8080)
     val serverSource = Http().bind(host, port)
@@ -153,6 +156,7 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
 
     implicit val system = ActorSystem()
     implicit val materializer = ActorMaterializer()
+    implicit val executionContext = system.dispatcher
 
     val serverSource = Http().bind(interface = "localhost", port = 8080)
 
@@ -186,87 +190,110 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
     import akka.http.scaladsl.model.HttpMethods._
     import akka.http.scaladsl.model._
     import akka.stream.ActorMaterializer
+    import scala.io.StdIn
 
-    implicit val system = ActorSystem()
-    implicit val materializer = ActorMaterializer()
+    object WebServer {
 
-    val requestHandler: HttpRequest => HttpResponse = {
-      case HttpRequest(GET, Uri.Path("/"), _, _, _) =>
-        HttpResponse(entity = HttpEntity(ContentTypes.`text/html(UTF-8)`,
-          "<html><body>Hello world!</body></html>"))
+      def main(args: Array[String]) {
+        implicit val system = ActorSystem()
+        implicit val materializer = ActorMaterializer()
+        // needed for the future map/flatmap in the end
+        implicit val executionContext = system.dispatcher
 
-      case HttpRequest(GET, Uri.Path("/ping"), _, _, _) =>
-        HttpResponse(entity = "PONG!")
+        val requestHandler: HttpRequest => HttpResponse = {
+          case HttpRequest(GET, Uri.Path("/"), _, _, _) =>
+            HttpResponse(entity = HttpEntity(ContentTypes.`text/html(UTF-8)`,
+              "<html><body>Hello world!</body></html>"))
 
-      case HttpRequest(GET, Uri.Path("/crash"), _, _, _) =>
-        sys.error("BOOM!")
+          case HttpRequest(GET, Uri.Path("/ping"), _, _, _) =>
+            HttpResponse(entity = "PONG!")
 
-      case _: HttpRequest =>
-        HttpResponse(404, entity = "Unknown resource!")
+          case HttpRequest(GET, Uri.Path("/crash"), _, _, _) =>
+            sys.error("BOOM!")
+
+          case _: HttpRequest =>
+            HttpResponse(404, entity = "Unknown resource!")
+        }
+
+        val bindingFuture = Http().bindAndHandleSync(requestHandler, "localhost", 8080)
+        println(s"Server online at http://localhost:8080/\nPress RETURN to stop...")
+        StdIn.readLine() // let it run until user presses return
+        bindingFuture
+          .flatMap(_.unbind()) // trigger unbinding from the port
+          .onComplete(_ ⇒ system.terminate()) // and shutdown when done
+
+      }
     }
-
-    Http().bindAndHandleSync(requestHandler, "localhost", 8080)
   }
 
   // format: OFF
 
   "high-level-server-example" in compileOnlySpec {
     import akka.http.scaladsl.Http
-    import akka.http.scaladsl.marshallers.xml.ScalaXmlSupport._
     import akka.http.scaladsl.server.Directives._
     import akka.stream.ActorMaterializer
+    import scala.io.StdIn
 
-    implicit val system = ActorSystem()
-    implicit val materializer = ActorMaterializer()
+    object WebServer {
+      def main(args: Array[String]) {
+        implicit val system = ActorSystem()
+        implicit val materializer = ActorMaterializer()
+        // needed for the future flatMap/onComplete in the end
+        implicit val executionContext = system.dispatcher
 
-    val route =
-      get {
-        pathSingleSlash {
-          complete {
-            <html>
-              <body>Hello world!</body>
-            </html>
+        val route =
+          get {
+            pathSingleSlash {
+              complete(HttpEntity(ContentTypes.`text/html(UTF-8)`,"<html><body>Hello world!</body></html>"))
+            } ~
+              path("ping") {
+                complete("PONG!")
+              } ~
+              path("crash") {
+                sys.error("BOOM!")
+              }
           }
-        } ~
-        path("ping") {
-          complete("PONG!")
-        } ~
-        path("crash") {
-          sys.error("BOOM!")
-        }
-      }
 
-    // `route` will be implicitly converted to `Flow` using `RouteResult.route2HandlerFlow`
-    Http().bindAndHandle(route, "localhost", 8080)
+        // `route` will be implicitly converted to `Flow` using `RouteResult.route2HandlerFlow`
+        val bindingFuture = Http().bindAndHandle(route, "localhost", 8080)
+        println(s"Server online at http://localhost:8080/\nPress RETURN to stop...")
+        StdIn.readLine() // let it run until user presses return
+        bindingFuture
+          .flatMap(_.unbind()) // trigger unbinding from the port
+          .onComplete(_ ⇒ system.terminate()) // and shutdown when done
+      }
+    }
   }
 
   "minimal-routing-example" in compileOnlySpec {
     import akka.http.scaladsl.Http
-    import akka.http.scaladsl.marshallers.xml.ScalaXmlSupport._
     import akka.http.scaladsl.server.Directives._
     import akka.stream.ActorMaterializer
+    import scala.io.StdIn
 
-    object Main extends App {
-      implicit val system = ActorSystem("my-system")
-      implicit val materializer = ActorMaterializer()
-      implicit val ec = system.dispatcher
+    object WebServer {
+      def main(args: Array[String]) {
 
-      val route =
-        path("hello") {
-          get {
-            complete {
-              <h1>Say hello to akka-http</h1>
+        implicit val system = ActorSystem("my-system")
+        implicit val materializer = ActorMaterializer()
+        // needed for the future flatMap/onComplete in the end
+        implicit val executionContext = system.dispatcher
+
+        val route =
+          path("hello") {
+            get {
+              complete(HttpEntity(ContentTypes.`text/html(UTF-8)`, "<h1>Say hello to akka-http</h1>"))
             }
           }
-        }
 
-      val bindingFuture = Http().bindAndHandle(route, "localhost", 8080)
+        val bindingFuture = Http().bindAndHandle(route, "localhost", 8080)
 
-      println(s"Server online at http://localhost:8080/\nPress RETURN to stop...")
-      Console.readLine() // for the future transformations
-      bindingFuture
-        .flatMap(_.unbind()) // trigger unbinding from the port
-        .onComplete(_ ⇒ system.terminate()) // and shutdown when done
+        println(s"Server online at http://localhost:8080/\nPress RETURN to stop...")
+        StdIn.readLine() // let it run until user presses return
+        bindingFuture
+          .flatMap(_.unbind()) // trigger unbinding from the port
+          .onComplete(_ ⇒ system.terminate()) // and shutdown when done
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


akka-docs/rst/scala/http/introduction.rst
```diff
@@ -10,23 +10,105 @@ You get to pick the API level of abstraction that is most suitable for your appl
...
+.. includecode:: ../code/docs/http/scaladsl/SprayJsonExampleSpec.scala
+   :include: example
+
+One of the strengths of akka-http is that streaming data is at its heart meaning that both request and response bodies
+can be streamed through the server achieving constant memory usage even for very large requests or responses. Streaming
+responses will be backpressured by the remote client so that the server will not push data faster than the client can
+handle, streaming requests means that the server decides how fast the remote client can push the data of the request
+body.
+
+Example that streams all positive numbers (given infinite time):
```
drewhk: no, it will stream everything until Int.MaxValue and then it would start from Int.MinValue :)

drewhk:
I think we will need more than this, but for now it is already a great improvement!

akka-docs/rst/scala/http/introduction.rst
johanandren: D'oh, good point.

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -380,4 +381,36 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
...
+    implicit val system = ActorSystem()
+    implicit val materializer = ActorMaterializer()
+
+    // streams are re-usable so we can define it here and use it for every request
+    val numbers = Source.fromIterator(() => Iterator.iterate(1)(n => n + 1))
+
+    val route =
+      path("numbers") {
+        get {
+          complete {
```
ktoso: Usually we have `complete()` (with `()`) as it's kind of special (so it doesn't get lost among "just some directives")

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -380,4 +381,36 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
...
   }
+
+  "all-positive-numbers" in compileOnlySpec {
+    //#all-positive-numbers
+    import akka.stream.scaladsl._
+    import akka.util.ByteString
+    import akka.http.scaladsl.Http
+    import akka.http.scaladsl.model.HttpEntity
+    import akka.http.scaladsl.server.Directives._
+    import akka.stream.ActorMaterializer
```
ktoso: :+1: to include all imports, very important here :)

akka-docs/rst/scala/http/introduction.rst
```diff
@@ -10,23 +10,105 @@ You get to pick the API level of abstraction that is most suitable for your appl
 This means that, if you have trouble achieving something using a high-level API, there's a good chance that you can get
 it done with a low-level API, which offers more flexibility but might require you to write more application code.
 
+DSL for HTTP servers
```
ktoso: Let's use the name "Routing DSL" consistently, we refer to it in other places

akka-docs/rst/scala/http/introduction.rst
```diff
@@ -10,23 +10,105 @@ You get to pick the API level of abstraction that is most suitable for your appl
...
+The modules that make up Akka HTTP
+----------------------------------
 Akka HTTP is structured into several modules:
 
 akka-http-core
   A complete, mostly low-level, server- and client-side implementation of HTTP (incl. WebSockets)
+  Details can be found in sections :ref:`http-low-level-server-side-api` and :ref:`http-client-side`
 
 akka-http
   Higher-level functionality, like (un)marshalling, (de)compression as well as a powerful DSL
```
ktoso: We could also say `(recommended)` here perhaps, as that's the one we recommend for writing servers?

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -380,4 +381,36 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
...
+    val numbers = Source.fromIterator(() => Iterator.iterate(1)(n => n + 1))
+
+    val route =
+      path("numbers") {
+        get {
+          complete {
+            HttpEntity(
+              ContentTypes.`text/plain(UTF-8)`,
+              // transform each number to a chunk of bytes
+              numbers.map(n => ByteString(s"$n\n"))
```
ktoso: This should perhaps also include a hello world, since here we went all into streaming ASAP.
Maybe:

```scala
get {
  path("hello") {
    complete("Hello World!")
  }
} ~ path("numbers") {
          complete {
            HttpEntity(
              ContentTypes.`text/plain(UTF-8)`,
              // transform each number to a chunk of bytes
              numbers.map(n => ByteString(s"$n\n"))
```

?

drewhk: some other common tasks I think:
 - ask an actor
 - create a per request actor

akka-docs/rst/scala/http/introduction.rst
```diff
@@ -10,23 +10,105 @@ You get to pick the API level of abstraction that is most suitable for your appl
...
 it done with a low-level API, which offers more flexibility but might require you to write more application code.
 
+DSL for HTTP servers
+--------------------
+The high-level, routing API of Akka HTTP provides a DSL to describe HTTP "routes" and how they should be handled.
+Each route is composed of one or more level of ``Directive`` s that narrows down to handling one specific type of
+request.
+
+For example one route might start with matching the ``path`` of the request, only matching if it is "/hello", then
+narrowing it down to only handle HTTP ``get`` requests and then ``complete`` those with a Scala XML literal, which
```
ktoso: Let's try not using the XML literals as hello world examples as they require additional dependencies and imports

ktoso: "object for which an implicit marshaller is provided" "default marshallers are provided for simple objects like String or ByteString, and you can define your own for example for json etc (links here)

johanandren: Yeah, just went lazy and the examples from the high-level-docs already had it. I agree, additionally the usefulness of those literals is a bit questionable so no need to introduce those here.

akka-docs/rst/scala/http/introduction.rst
```diff
@@ -10,23 +10,105 @@ You get to pick the API level of abstraction that is most suitable for your appl
...
+The ``Route`` created using the Route DSL is then "bound" to a port to start serving HTTP requests:
+
+.. includecode2:: ../code/docs/http/scaladsl/HttpServerExampleSpec.scala
+   :snippet: minimal-routing-example
+
+A common use case is to reply to a request using a model object having the marshaller transform it into JSON. In
+this case shown by two separate routes where one returns an ``Item`` that is marshalled into JSON and one accepts
+JSON that is unmarshalled into an ``Order``.
+
+The logic for the marshalling and unmarshalling is provided by the "spray-json" library (Details how to use that here
```
ktoso: Marshallers to/from JSON can be provided by external libraries using any JSON serialization framework, however Akka ships with `akka-http-spray-json` module that uses `spray-json` (link to its spray json support docs here) for that purpose, so we'll be using it in our docs. It's used as so:

akka-docs/rst/scala/http/introduction.rst
```diff
@@ -10,23 +10,105 @@ You get to pick the API level of abstraction that is most suitable for your appl
...
+:ref:`akka-http-spray-json`) as two methods creating reflection based implementations:
```
ktoso: spray-json is macro based not reflection

akka-docs/rst/scala/http/introduction.rst
ktoso: Given the context I guess the example is OK, I'll really want to replace it with "json streaming" soon though (-:

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
johanandren: Not sure I agree on the hello world, the example is the third in the docs with the first of them being a "hello world" example.

The ask an actor/request per actor is a good idea!

ktoso: Was reviewing top to bottom, I see there's an `minimal-routing-example` too, probably good enough then :)

akka-docs/rst/scala/http/introduction.rst
johanandren: I just looked through the `jsonFormatN` used in the sample, and they definitely do reflection + `Product.productElement(n)`, no macros.

akka-docs/rst/scala/http/introduction.rst
johanandren: I already rewrote it using random numbers, to avoid nitpicky-devs to get stuck on that. :)

akka-docs/rst/scala/http/introduction.rst
johanandren: How about I change the high-level example to just use strings, and keep using it, because I think this holds true there as well?

ktoso: yeap, sounds good. Then the JSON example shows marshallers

akka-docs/rst/scala/http/introduction.rst
```diff
@@ -10,23 +10,105 @@ You get to pick the API level of abstraction that is most suitable for your appl
...
+marshallers must be added (See :ref:`akka-http-spray-json` and :ref:`akka-http-xml-marshalling`)
+
+The ``Route`` created using the Route DSL is then "bound" to a port to start serving HTTP requests:
+
+.. includecode2:: ../code/docs/http/scaladsl/HttpServerExampleSpec.scala
+   :snippet: minimal-routing-example
+
+A common use case is to reply to a request using a model object having the marshaller transform it into JSON. In
+this case shown by two separate routes where one returns an ``Item`` that is marshalled into JSON and one accepts
+JSON that is unmarshalled into an ``Order``.
```
ktoso: Perhaps include a Future[Order] here or in the next example? To show that those compose (and we provide ones for Futures)

akka-docs/rst/scala/http/introduction.rst
ktoso: Hm hm, you're right - no macros, however no reflection though – that's plain case class methods AFAICS

johanandren: `ProductFormats.extractFieldNames` ??

akka-docs/rst/scala/http/introduction.rst
```diff
@@ -10,23 +10,105 @@ You get to pick the API level of abstraction that is most suitable for your appl
...
+
+The logic for the marshalling and unmarshalling is provided by the "spray-json" library (Details how to use that here
+:ref:`akka-http-spray-json`) as two methods creating reflection based implementations:
+
+.. includecode:: ../code/docs/http/scaladsl/SprayJsonExampleSpec.scala
+   :include: example
+
+One of the strengths of akka-http is that streaming data is at its heart meaning that both request and response bodies
```
ktoso: Let's try to use `Akka HTTP` consistently unless refering to modules

akka-docs/rst/scala/http/introduction.rst
johanandren: Yep. Will add an example to `SprayJsonExampleSpec`, I anyways strongly dislike how that example looks with all the trait-mixing. I think it really takes focus of what it is trying to show.

akka-docs/rst/scala/http/introduction.rst
ktoso: That's once for each Format, not "during marshalling" so I always think of spray-json as "no reflection after initial setup" => pretty much no reflection as for the parsing. Or am I missing something since some years :)

johanandren: Nope, that is correct.

akka-docs/rst/scala/http/introduction.rst
```diff
@@ -10,23 +10,115 @@ You get to pick the API level of abstraction that is most suitable for your appl
 This means that, if you have trouble achieving something using a high-level API, there's a good chance that you can get
 it done with a low-level API, which offers more flexibility but might require you to write more application code.
 
-Akka HTTP is structured into several modules:
+Using Akka HTTP
+---------------
+Akka HTTP is provided in separate jar files, to use it make sure to include the following dependencies::
+
+  "com.typesafe.akka" %% "akka-http-core" % @version@" @crossString@
+  "com.typesafe.akka" %% "akka-http-experimental" % @version@" @crossString@
```
2m: A `"` is missing before version.

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -254,16 +249,14 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
...
-            }
+            complete("<h1>Say hello to akka-http</h1>")
           }
         }
 
       val bindingFuture = Http().bindAndHandle(route, "localhost", 8080)
 
       println(s"Server online at http://localhost:8080/\nPress RETURN to stop...")
-      Console.readLine() // for the future transformations
+      StdIn.readLine() // for the future transformations
```
2m: The comment is not really helpful. (I know it has been here before). Lets change it to `wait for user input` or `delay future transformations until user input`. WDYT?

johanandren: Agree, I think I did not even read it when I changed the readline. :)

akka-docs/rst/scala/code/docs/http/scaladsl/SprayJsonExampleSpec.scala
```diff
@@ -47,4 +49,48 @@ class SprayJsonExampleSpec extends WordSpec with Matchers {
...
   }
+
+  "second-spray-json-example" in compileOnlySpec {
+    import akka.actor.ActorSystem
+    import akka.stream.ActorMaterializer
+    import akka.Done
+    import akka.http.scaladsl.server.Route
+    import akka.http.scaladsl.server.Directives._
+    import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
+    import spray.json.DefaultJsonProtocol._
```
2m: Since we have all imports enumerated, I think we should also mention that this example requires spray-json library and marshaling support artifacts to run.

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
2m: `import akka.http.scaladsl.model.ContentTypes` is missing.

akka-docs/rst/scala/code/docs/http/scaladsl/SprayJsonExampleSpec.scala
johanandren: There is a section after this specific example referring to the docs for the spray-json which then has the actual artifact, you don't think that is enough?

akka-docs/rst/scala/http/introduction.rst
```diff
@@ -10,23 +10,115 @@ You get to pick the API level of abstraction that is most suitable for your appl
...
+can be streamed through the server achieving constant memory usage even for very large requests or responses. Streaming
+responses will be backpressured by the remote client so that the server will not push data faster than the client can
+handle, streaming requests means that the server decides how fast the remote client can push the data of the request
+body.
+
+Example that streams random numbers as long as the client accepts them:
+
+.. includecode2:: ../code/docs/http/scaladsl/HttpServerExampleSpec.scala
+   :snippet: random-numbers
+
```
2m: Do you think it would be useful to mention that the example can be tried out simulating a slow client with `curl --limit-rate 50b 127.0.0.1:8080/random` ?

johanandren: :+1:

akka-docs/rst/scala/code/docs/http/scaladsl/SprayJsonExampleSpec.scala
2m: Aha. That is true. It is alright then.

2m:
One more nitpick would be that all three examples have minor differences in the structure. First one is `object Main extends App { ... }` second does nothing with a created route and third one runs the route but does not stop the server as the first one does.

As we are aiming this page to be first contact with Akka HTTP, I would suggest to have examples as runnable as possible and minimize differences to get rid of all unnecessary questions that might rise to the user. My suggestion would be to make other examples follow the structure of the first one.

Otherwise, LGTM!

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -186,33 +187,44 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
...
-    }
+          case HttpRequest(GET, Uri.Path("/crash"), _, _, _) =>
+            sys.error("BOOM!")
+
+          case _: HttpRequest =>
+            HttpResponse(404, entity = "Unknown resource!")
+        }
+
+        val bindingFuture = Http().bindAndHandleSync(requestHandler, "localhost", 8080)
+        StdIn.readLine() // let it run until user presses return
```
ktoso: please make it print that info instead of comment `  StdIn.readline(s"Server online at http://localhost:8080/\nPress RETURN to stop...")`

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -222,11 +234,7 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
     val route =
       get {
         pathSingleSlash {
-          complete {
-            <html>
-              <body>Hello world!</body>
-            </html>
-          }
+          complete("<html><body>Hello world!</body></html>")
```
ktoso: yeah, good.

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -380,4 +389,118 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
...
+        implicit val system = ActorSystem()
+        implicit val materializer = ActorMaterializer()
+        implicit val ec = system.dispatcher
+
+        val auction = system.actorOf(Auction.props, "auction")
+
+        val route =
+          path("auction") {
+            put {
+              parameter("bid".as[Int], "user".as[String]) { (bid, user) =>
```
ktoso: I'd usually drop `.as[String]` since String is what you get without the `.as`

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -380,4 +389,118 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
...
+                // place a bid, fire-and-forget
+                auction ! Bid(user, bid)
```
ktoso: this actually reminds me of a very common pattern that people like to implement... I think it's too complex for this doc but maybe we should do it in a template later on.
It's about `request(id) => map.getOrElseNewActor(id) ? bla`

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -380,4 +389,118 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
...
+              implicit val timeout: Timeout = 5.seconds
+
+              // query the actor for the current auction state
+              val bids: Future[Bids] = (auction ? GetBids).mapTo[Bids]
+              complete(bids)
+            }
+          }
+
+        val bindingFuture = Http().bindAndHandle(route, "localhost", 8080)
+        println(s"Server online at http://localhost:8080/\nPress RETURN to stop...")
```
ktoso: :+1:

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -380,4 +389,118 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
...
+          Iterator.continually(Random.nextInt()))
+
+        val route =
+          path("random") {
+            get {
+              complete(
+                HttpEntity(
+                  ContentTypes.`text/plain(UTF-8)`,
+                  // transform each number to a chunk of bytes
+                  numbers.map(n => ByteString(s"$n\n"))
```
ktoso: ok :)

akka-docs/rst/scala/http/introduction.rst
```diff
@@ -10,23 +10,129 @@ You get to pick the API level of abstraction that is most suitable for your appl
...
+   :include: stream-random-numbers
+
+Connecting to this service with a slow HTTP client would backpressure so that the next random number is produced on
+demand with constant memory usage on the server. This can be seen using curl and limiting the rate
+``curl --limit-rate 50b 127.0.0.1:8080/random``
+
+
+Akka HTTP routes easily interacts with actors. In this example one route allows for placing bids in a fire-and-forget
+style while the second route contains a request-response interaction with an actor. The resulting response is rendered
+as json and returned when the response arrives from the actor.
```
ktoso: for reference this is not actor-per-request, I thought that's the one we wanted to show hm

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -380,4 +389,118 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
...
+                auction ! Bid(user, bid)
+                complete(StatusCodes.Accepted, "bid placed")
+              }
+            }
+            get {
+              implicit val timeout: Timeout = 5.seconds
+
+              // query the actor for the current auction state
+              val bids: Future[Bids] = (auction ? GetBids).mapTo[Bids]
+              complete(bids)
```
ktoso: good! :+1:

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -380,4 +389,118 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
...
+    import akka.pattern.ask
+    import akka.actor.ActorSystem
+    import akka.stream.ActorMaterializer
+    import akka.http.scaladsl.Http
+    import akka.http.scaladsl.server.Directives._
+    import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
+    import spray.json.DefaultJsonProtocol._
+    import scala.io.StdIn
+
+    object WebServer {
```
ktoso: I would prefer `extends Directives with SprayJsonSupport with DefaultJsonProtocol` instead of the imports.
That's how we usually recommend constructing these

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
```diff
@@ -380,4 +389,118 @@ class HttpServerExampleSpec extends WordSpec with Matchers {
...
+
+      case class Bid(userId: String, bid: Int)
+      case object GetBids
+      case class Bids(bids: List[Bid])
+
+      implicit val bidFormat = jsonFormat2(Bid)
```
ktoso: Good, perhaps a comment that this is spray json?

akka-docs/rst/scala/code/docs/http/scaladsl/SprayJsonExampleSpec.scala
```diff
@@ -47,4 +49,57 @@ class SprayJsonExampleSpec extends WordSpec with Matchers {
...
+              val maybeItem: Future[Option[Item]] = fetchItem(id)
+              complete(maybeItem)
+            }
+          } ~
+            post {
+              path("create-order") {
+                entity(as[Order]) { order =>
+                  val saved: Future[Done] = saveOrder(order)
+                  onComplete(saved) { done =>
+                    complete("order created")
```
ktoso: Okey :+1:

akka-docs/rst/scala/code/docs/http/scaladsl/SprayJsonExampleSpec.scala
```diff
@@ -47,4 +49,57 @@ class SprayJsonExampleSpec extends WordSpec with Matchers {
...
+
+        // needed to run the route
+        implicit val system = ActorSystem()
+        implicit val materializer = ActorMaterializer()
+
+        val route: Route =
+          get {
+            pathPrefix("item" / LongNumber) { id =>
+              val maybeItem: Future[Option[Item]] = fetchItem(id)
+              complete(maybeItem)
```
ktoso: That's actually a tricky case... could you put that and return `None` in TestServer to see how it behaves by default and document that? Some people may expect 404 to happen but that's not the case (maybe as `..note:` around that snippet).

ktoso:
LGTM, had some last minor comments. Merge once fixed and squashed :)

akka-docs/rst/scala/http/introduction.rst
johanandren: No, I decided to skip that.

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
johanandren: I actually did this on purpose because I really dislike that cakey style of pulling in dependencies, just clashes with `is-a` vs `has-a` or `needs-a` for me. Didn't know that is our default recommendation (and if it is, I still don't agree with it).

akka-docs/rst/scala/http/introduction.rst
johanandren: Heh. Should add rationale that I wrote on gitter yesterday: It is hard to keep nice and short plus I don't think it's the default we should recommend so that new users cargo-cult copy it without actually needing it (having seen quite a bit of that in the wild)

akka-docs/rst/scala/code/docs/http/scaladsl/SprayJsonExampleSpec.scala
johanandren: Yeah, actually expected that to happen, thought you added such a marshaller that does that the other week?

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
johanandren: Also, It's not quite true to do this in the line after, should be in a `Future.foreach` or else we don't know it actually did start. But that makes the sample noisier so I'm leaving it be as is with just printlns.

akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
johanandren: Asch. Content type becomes "text/plain"

akka-docs/rst/scala/code/docs/http/scaladsl/SprayJsonExampleSpec.scala
johanandren: Updating the sample to handle `None` manually instead. A shame that it isn't how the marshaller works, but maybe that's not possible, or there are reasons why that would be bad?

johanandren: For reference, I found #19740

PR #19849, opened 2016-02-23
=htp #19397,#19842 fix content negotiation for non 2xx and Accept handling

### LINKED ISSUE #19397: request.completeWithStatus(NOT_FOUND) results in 406 Not Acceptable
A handler that completes a request with `request.completeWithStatus(NOT_FOUND)` when the incoming request has `Accept: application/json` results in a `406 Not Acceptable` response and entity "Resource representation is only available with these types: text/plain".

This is also the case when completing with `completeWithStatus(NOT_FOUND.intValue())`.

The correct behavior in this case should be the `NOT_FOUND` status code with no entity.

### LINKED ISSUE #19842: Clarify Accept header handling in comparision to Spray
Hi there,

I would like the behaviour of how akka-http handles the Accept header clarified. Possibly in the spray-migration guide as the behaviour appears to be different.

Given the following routes structure in either Spray / Akka-http :
```scala
pathPrefix("test" / LongNumber) { number =>
   get {
      complete {
          //Imagine some other validation occuring
          if(number > 100) {
              BadRequest -> "Currently we don't support numbers greater than 100"
          } else {
              OK -> service.retrieveJsonFor(number)
          }
      }
   }
}
```
Now say I make the following call
```
GET test/101
Accept: application/json
```
Spray will return
```
400 Bad Request
Content-Type: text/plain
Currently we don't support numbers greater than 100
```
Whereas akka-http will return
```
406 Not Acceptable
Content-Type: text/plain
Resource representation is only available with these types:
application/json
```
This difference should be highlighted somewhere in the documentation at least. The behaviour causes the lose of the BadRequest error message / changing of status codes which is a bit of a gotcha when moving an existing Spray app to akka-http.

Kind regards,

am

ktoso:
The added test replicates the behaviour as shown in the example in #19842

The general thing is that we need to retry at the right point, with trimming the Accept headers from the request. Some logic for this was already implemented, just not applied in the right spot from what I found.

Resolves #19397 
Resolves #19842

Please review if you have a moment to spare @jrudolph :)

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshal.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshal.scala
@@ -9,11 +9,13 @@ import akka.http.scaladsl.server.ContentNegotiator
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.util.FastFuture._
 
+import scala.util.control.NoStackTrace
+
 object Marshal {
   def apply[T](value: T): Marshal[T] = new Marshal(value)
 
-  case class UnacceptableResponseContentTypeException(supported: Set[ContentNegotiator.Alternative])
-    extends RuntimeException
+  final case class UnacceptableResponseContentTypeException(supported: Set[ContentNegotiator.Alternative])
+    extends RuntimeException with NoStackTrace
 }
 
 class Marshal[A](val value: A) {
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
@@ -239,6 +239,12 @@ private object PoolSlot {
   }
 
   final class UnexpectedDisconnectException(msg: String, cause: Throwable) extends RuntimeException(msg, cause) {
+    def this(msg: String) {
+      this(msg, null)
+    }
+  }
+
+  final class UnexpectedDisconnectException(msg: String, cause: Throwable) extends RuntimeException(msg, cause) {
     def this(msg: String) = this(msg, null)
   }
 }
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
@@ -239,12 +239,6 @@ private object PoolSlot {
   }
 
   final class UnexpectedDisconnectException(msg: String, cause: Throwable) extends RuntimeException(msg, cause) {
-    def this(msg: String) {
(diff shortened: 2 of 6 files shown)


akka-http/src/main/scala/akka/http/scaladsl/marshalling/Marshal.scala
```diff
@@ -9,11 +9,13 @@ import akka.http.scaladsl.server.ContentNegotiator
...
 
+import scala.util.control.NoStackTrace
+
 object Marshal {
   def apply[T](value: T): Marshal[T] = new Marshal(value)
 
-  case class UnacceptableResponseContentTypeException(supported: Set[ContentNegotiator.Alternative])
-    extends RuntimeException
+  final case class UnacceptableResponseContentTypeException(supported: Set[ContentNegotiator.Alternative])
+    extends RuntimeException with NoStackTrace
```
johanandren: :+1:

johanandren:
LGTM (with limited level of undstanding of the dsl internals)

patriknw:
LGTM

jrudolph:
LGTM, I wonder if a test involving real TCP / HTTP stack is necessary? Couldn't this be run with a simple directive test using the testkit?

ktoso:
Oh, I missed this has gotten enough reviews already. Thanks for reviewing!
I'll give it a look to see if it's possible to test only the route, without entire server as @jrudolph suggests and merge once I fix or give up ;)

ktoso:
Whoa, I was able to pull it off without a full end-to-end starting the server, it does require some different handling inside `TildaArrow` though.

FYI @jrudolph normal RoutingSpec was not enough (now it is, see additional commit) here as we just end up with `Request was rejected with rejection UnacceptedResponseContentTypeRejection(Set(MediaType(text/plain)))` and that's it, whereas we need the logic that retries with an `Accept */*` which is in the server layer.

In the actual server, we properly `Route.seal()` (and in the RoutingSpec we do not, because we want to assert on the rejections). So the trade off is either not being able to inspect rejections when we use `~!>` – which is the new version of "actually handle the rejections" in the RoutingSpec. I think not many people will actually use that so I think I'm ok with such method name... WDYT?

// PS: I did tests that `ContentNegotiationGivenResponseCodeSpec` does exactly the same now as before with real server, but not including the integration tests as the entire goal was to avoid starting the server.

Would like another round of reviews here before I merge, thanks!

jrudolph:
In other tests, we just use `request ~> Route.seal(route) ~> check` for that purpose. Would that have sufficed here as well?

ktoso:
Duh of course, thanks @jrudolph for pointing that out. I've removed my additional code and just use the Route.seal now, thanks!

Issue #19828, opened 2016-02-19
Calling unstashAll in persistAsync callback handler will barrier this messages, but ``perisist`` not.


MQ-EL:
Hi! 

When call ``unstashAll`` in ``persistAsync`` callback, I find this unstashed messages will be hanged by internal stash until a new incomming message arrived.

But the ``persist`` method doesn't have this problem.

_reproducer:_

config: 
```config
 akka {
  loglevel = INFO
  persistence {
    journal {
      plugin = "akka.persistence.journal.inmem"
    }
    snapshot-store {
      plugin = "akka.persistence.snapshot-store.local"
    }
  }
}  
```
code:
```scala
import akka.actor._
import akka.persistence.PersistentActor

object UnstashAllTest extends App{

  val system = ActorSystem()
  
  val pa = system.actorOf(Props[PA], "testPA")
  
  pa ! "start"
  pa ! "persist"
  Thread.sleep(10000)
  pa ! "end" // if comment it, then "start" message will never be popped from internal stash
}

class PA extends PersistentActor with DiagnosticActorLogging {
  var first = true
  
  override def receiveRecover: Receive = { case _ ⇒ }

  override def receiveCommand: Receive = {
    case "start" if first⇒ 
      stash()
      first = false
      log.info("START")
    case "start" ⇒ 
      log.info("UNSTASHALL SUCCESSFUL")
    case "persist" ⇒
      //change to persist method has no problem
      persistAsync("unstashAll") { case _ ⇒ unstashAll() }
    case "end" ⇒ 
      log.info("END")
  }

  override def persistenceId: String = "testPA"
}
```

The persistent actor should not barriers application messages when all the persist transactions were finished?

MQ-EL:
I think it is a bug. when PA in the _processingCommands_ state, it misses to pop ``internalStash``, but _persistingEvents_ does.

The ``internalStash`` may contains msgs come from user's unstashAll, when in _processingCommands_

patriknw:
sounds like a bug, will you work on it @MQ-EL ?

MQ-EL:
no problem

PR #19850, opened 2016-02-23
=per #19828 pop the internal stash when the writing was finished in t…

MQ-EL:
…he processing commands state


project/MiMa.scala
```diff
@@ -607,7 +607,10 @@ object MiMa extends AutoPlugin {
 
         // #19440
         ProblemFilters.exclude[MissingMethodProblem]("akka.pattern.PipeToSupport.pipeCompletionStage"),
-        ProblemFilters.exclude[MissingMethodProblem]("akka.pattern.FutureTimeoutSupport.afterCompletionStage")
+        ProblemFilters.exclude[MissingMethodProblem]("akka.pattern.FutureTimeoutSupport.afterCompletionStage"),
+        
+        //#19828
+        ProblemFilters.exclude[AbstractMethodProblem]("akka.persistence.Eventsourced#ProcessingState.onWriteMessageComplete")
```
MQ-EL: is it reasonable?

MQ-EL:
Refs #19828

ktoso:
OK TO TEST

project/MiMa.scala
johanandren: `ProcessingState` is private so I think this is safe.

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -183,7 +186,7 @@ private[persistence] trait Eventsourced extends Snapshotter with PersistenceStas
   /** INTERNAL API. */
   override protected[akka] def aroundPreRestart(reason: Throwable, message: Option[Any]): Unit = {
     try {
-      internalStash.unstashAll()
+      unstashInternally(all = true)
```
johanandren: I think these could just as well have been left `internalStash.unstashAll`

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -538,6 +541,8 @@ private[persistence] trait Eventsourced extends Snapshotter with PersistenceStas
    * Common receive handler for processingCommands and persistingEvents
    */
   private abstract class ProcessingState extends State {
+    override def recoveryRunning: Boolean = false
+
```
johanandren: Why was this change needed? It's not referenced anywhere in the other changes as far as I can see.

johanandren: Ah, I see, it was in both the concrete states.

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -604,12 +608,13 @@ private[persistence] trait Eventsourced extends Snapshotter with PersistenceStas
...
-        internalStash.unstashAll()
-      else
-        internalStash.unstash()
+      if (pendingStashingPersistInvocations > 0) changeState(persistingEvents)
+      else unstashInternally(all = err)
+    }
+
+    override def onWriteMessageComplete(err: Boolean): Unit = {
+      pendingInvocations.pop()
+      unstashInternally(all = err)
```
johanandren: So, this is the actual bugfix, right, the rest is just generic cleanup?

akka-persistence/src/test/scala/akka/persistence/PersistentActorStashingSpec.scala
```diff
@@ -108,9 +114,13 @@ abstract class PersistentActorStashingSpec(config: Config) extends PersistenceSp
       persistentActor ! Cmd("a")
       persistentActor ! Cmd("b")
       persistentActor ! Cmd("c")
+      persistentActor ! Cmd("d")
+      persistentActor ! Cmd("e")
```
johanandren: A separate test covering this bug would be nicer than appending it to existing test cases.

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
MQ-EL: If the ``internalStash.unstash/all`` method has been encapsulated, perhaps call it directly in other places would not be a good idea for me.

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
MQ-EL: yes, it is.

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
johanandren: I wouldn't think of it as `unstashAll` being encapsulated but the case of conditional `unstashAll` which this isn't. Not important though, if you feel strongly about this, let it be as is.

patriknw: I agree with @johanandren, I don't see the benefit of that method

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
```diff
@@ -604,12 +608,13 @@ private[persistence] trait Eventsourced extends Snapshotter with PersistenceStas
...
       if (eventBatch.nonEmpty) flushBatch()
 
-      if (pendingStashingPersistInvocations > 0)
-        changeState(persistingEvents)
-      else if (err)
-        internalStash.unstashAll()
-      else
-        internalStash.unstash()
+      if (pendingStashingPersistInvocations > 0) changeState(persistingEvents)
+      else unstashInternally(all = err)
```
patriknw: ok, here is the benefit, alright to keep then

project/MiMa.scala
patriknw: diff of javap of EventSouced looks good

patriknw:
LGTM

akka-persistence/src/test/scala/akka/persistence/PersistentActorStashingSpec.scala
drewhk: +1

drewhk:
LGTM, I would also appreciate a new test case though.

akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala
MQ-EL: Got it, thanks @johanandren @patriknw :)

MQ-EL:
BTW: would [contrib-sharding-config](https://github.com/akka/akka/blob/master/akka-contrib/src/main/resources/reference.conf#L10) need to remove?

patriknw:
good catch about contrib-sharding-config, @MQ-EL 
it's indeed obsolite

patriknw:
PLS REBUILD

akka-persistence/src/test/scala/akka/persistence/PersistentActorStashingSpec.scala
```diff
@@ -112,9 +153,11 @@ abstract class PersistentActorStashingSpec(config: Config) extends PersistenceSp
       expectMsg("c")
       expectMsg("a")
     }
+  }
 
+  def stashWithServralMessages[T <: NamedPersistentActor : ClassTag](): Unit = {
```
johanandren: Typo, should be "Several"

akka-persistence/src/test/scala/akka/persistence/PersistentActorStashingSpec.scala
```diff
@@ -134,18 +179,29 @@ abstract class PersistentActorStashingSpec(config: Config) extends PersistenceSp
...
+    behave like stashWithServralMessages[UserStashManyPersistentActor]()
+    behave like stashUnderFailures[UserStashFailurePersistentActor]()
+  }
+  
+  "Stashing(unstashAll called in handler) in a persistent actor" must {
+    behave like stash[UserStashWithinHandlerPersistentActor]()
+    behave like stashWithServralMessages[UserStashWithinHandlerManyPersistentActor]()
+    behave like stashUnderFailures[UserStashWithinHandlerFailureCallbackPersistentActor]()
+  }
+
```
johanandren: Thanks! :+1:

johanandren:
LGTM

MQ-EL:
In my local,  it is ok. what does it happened?

ktoso:
Binary compat check failed: 

```
[info] akka-persistence: found 1 potential binary incompatibilities while checking against com.typesafe.akka:akka-persistence_2.11:2.4.0  (filtered 14)
[error]  * in current version there is abstract method onWriteMessageComplete(Boolean)Unit in class akka.persistence.Eventsourced#ProcessingState, which does not have a correspondent
[error]    filter with: ProblemFilters.exclude[ReversedAbstractMethodProblem]("akka.persistence.Eventsourced#ProcessingState.onWriteMessageComplete")
```

Since `Eventsourced` is private API, we can add the exception AFAICS.

ktoso:
You can run the checl locally via `mimaReportBinaryIssues` or `validatePullRequest` (also runs tests, only in modules which given PR could affect)

patriknw:
@MQ-EL what is missing here? is it only the mima filter? can you fix that, or would you like me to do it?

MQ-EL:
Sorry for late, did i miss something? I have already add mima filter ,but it seems not effect. I will try fix it today.

MQ-EL:
Hi @patriknw , I found that the version of this pr branch is conflict with master( ```MiMa.scala``` && mima version from 2.1.7 to 2.1.9, which ```abstractMethod``` changed). How  to surround it? recreate pr or you do?

patriknw:
@MQ-EL can't you rebase it? I'm sorry for the inconveniences. I can do that for you if you like?

MQ-EL:
ok, I will rebase it, but wait for tomorrow morning. sorry for that

drewhk:
@patriknw this is mergeable right?

PR #19853, opened 2016-02-23
=all cleanup imports

jrudolph:
I don't know if this is the best time to do conflict-prone changes like this (or if you'd rather fix this while a file is touched otherwise). I had a try at an (almost) automatic unused import cleanup thingy, I could rerun it at any other time if that's what you'd prefer.

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/actor/ConsistencySpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/actor/ConsistencySpec.scala
@@ -3,7 +3,7 @@ package akka.actor
 import language.postfixOps
 
 import akka.testkit.AkkaSpec
-import akka.dispatch.{ ThreadPoolConfig, ThreadPoolConfigBuilder, UnboundedMailbox }
+import akka.dispatch.{ ThreadPoolConfig }
 import scala.concurrent.duration._
 
 object ConsistencySpec {
--- a/akka-actor-tests/src/test/scala/akka/pattern/BackoffOnRestartSupervisorSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/pattern/BackoffOnRestartSupervisorSpec.scala
@@ -8,7 +8,6 @@ import java.util.concurrent.{ TimeUnit, CountDownLatch }
 
 import akka.pattern.TestActor.NormalException
 import akka.testkit.{ ImplicitSender, AkkaSpec, TestProbe, filterException }
-import scala.concurrent.Future
 import scala.concurrent.duration._
 import akka.actor._
 import scala.language.postfixOps
@@ -21,7 +20,6 @@ object TestActor {
 }
 
 class TestActor(probe: ActorRef) extends Actor {
-  import context.dispatcher
 
   probe ! "STARTED"
 
--- a/akka-contrib/src/main/scala/akka/contrib/throttle/TimerBasedThrottler.scala
+++ b/akka-contrib/src/main/scala/akka/contrib/throttle/TimerBasedThrottler.scala
@@ -5,13 +5,11 @@
 package akka.contrib.throttle
 
 import scala.concurrent.duration.{ Duration, FiniteDuration }
-import scala.util.control.NonFatal
 import scala.collection.immutable.{ Queue ⇒ Q }
 import akka.actor.{ ActorRef, Actor, FSM }
 import Throttler._
 import TimerBasedThrottler._
 import java.util.concurrent.TimeUnit
-import akka.AkkaException
 
 /**
  * @see [[akka.contrib.throttle.TimerBasedThrottler]]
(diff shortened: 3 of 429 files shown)


akka-contrib/src/main/scala/akka/contrib/throttle/TimerBasedThrottler.scala
```diff
@@ -1,305 +1,301 @@
-/**
- * Copyright (C) 2009-2016 Typesafe Inc. <http://www.typesafe.com>
- */
-
-package akka.contrib.throttle
-
-import scala.concurrent.duration.{ Duration, FiniteDuration }
-import scala.util.control.NonFatal
```
jrudolph: In case you wonder what happened here. There's a trailing '\r' after "NonFatal" that was removed together with the import which seems to make this file being treated as binary.

akka-actor-tests/src/test/scala/akka/actor/ConsistencySpec.scala
```diff
@@ -3,7 +3,7 @@ package akka.actor
 import language.postfixOps
 
 import akka.testkit.AkkaSpec
-import akka.dispatch.{ ThreadPoolConfig, ThreadPoolConfigBuilder, UnboundedMailbox }
+import akka.dispatch.{ ThreadPoolConfig }
```
patriknw: this is not perfect, but alright I guess

akka-actor-tests/src/test/scala/akka/pattern/BackoffOnRestartSupervisorSpec.scala
```diff
@@ -21,7 +20,6 @@ object TestActor {
 }
 
 class TestActor(probe: ActorRef) extends Actor {
-  import context.dispatcher
```
patriknw: that is interesting, did you find that automatically?

akka-contrib/src/main/scala/akka/contrib/throttle/TimerBasedThrottler.scala
ktoso: ok, thanks for the info :) looks rather weird without that hint :)

patriknw:
looking good

akka-docs/rst/scala/code/docs/stream/io/StreamFileDocSpec.scala
```diff
@@ -27,7 +27,6 @@ class StreamFileDocSpec extends AkkaSpec(UnboundedMailboxConfig) {
 
   {
     //#file-source
-    import akka.stream.scaladsl._
```
ktoso: we likely want to keep that import

akka-docs/rst/scala/code/docs/testkit/TestkitDocSpec.scala
```diff
@@ -127,7 +127,6 @@ class TestkitDocSpec extends AkkaSpec with DefaultTimeout with ImplicitSender {
 
     //#test-behavior
     import akka.testkit.TestActorRef
-    import scala.concurrent.duration._
```
ktoso: likely want to keep that import

akka-actor-tests/src/test/scala/akka/actor/ActorDSLSpec.scala
```diff
@@ -15,7 +15,6 @@ import java.util.concurrent.TimeoutException
 
 class ActorDSLDummy {
   //#import
-  import akka.actor.ActorDSL._
```
ktoso: likely want to keep that import

ktoso:
Found one or two places where the imports should likely stay as they're documentation (around `//#`), otherwise LGTM if validation is happy, thanks!

ktoso:
I guess it's best if we merge this one ASAP btw, rebasing will be no fun ;)

drewhk:
wow, nice tool. What did you use? Can it be integrated with our build somehow (and probably disable it on docs project to keep import in sample code).

akka-actor-tests/src/test/scala/akka/pattern/BackoffOnRestartSupervisorSpec.scala
jrudolph: Yes, `-Ywarn-unused-import` seems to be much better on 2.12.

jrudolph:
@ktoso I manually reverted the changes in the docs. (If this should become a tool at some point, a long-term change would be to actually use those import in those dummy doc tests).

@drewhk This is currently [just a hack](https://gist.github.com/jrudolph/432298193b071604b997) that parses compiler warnings and applies some heuristic to delete not more than necessary hopefully.

ktoso:
LGTM, let's merge after validation :)

ktoso:
OH NO, we made it not merge cleanly already :sob:
Sorry about that, perhaps I'll be able to pick and merge manually?

ktoso:
Rebased and merged (was just something silly) https://github.com/akka/akka/commit/0d8dd3c7114ed5258208aaed3575460a056876bb
Authorship preserved ofc.

ktoso:
Thanks again @jrudolph !

jrudolph:
Cool, thanks for merging, @ktoso.

PR #19858, opened 2016-02-23
=htc #19287 fixes SNI issue on JDK8 in http client

ktoso:
Resolves https://github.com/akka/akka/issues/19287 however I think this SNI applying should be done in ssl-config, after review will move it there and prep a version

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/client/HighLevelOutgoingConnectionSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/client/HighLevelOutgoingConnectionSpec.scala
@@ -91,5 +91,6 @@ class HighLevelOutgoingConnectionSpec extends AkkaSpec with ScalaFutures {
       a[One2OneBidiFlow.OutputTruncationException.type] should be thrownBy Await.result(x, 3.second)
       binding.futureValue.unbind()
     }
+
   }
 }
--- a/akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala
@@ -4,6 +4,8 @@
 package akka.stream.impl.io
 
 import java.nio.ByteBuffer
+import java.util
+import java.util.Collections
 import javax.net.ssl.SSLEngineResult.HandshakeStatus
 import javax.net.ssl.SSLEngineResult.HandshakeStatus._
 import javax.net.ssl.SSLEngineResult.Status._
@@ -154,6 +156,15 @@ private[akka] class TLSActor(settings: ActorMaterializerSettings,
     e.setUseClientMode(role == Client)
     e
   }
+
+  // since setting a custom HostnameVerified (in JDK8, update 60 still) disables SNI
+  // see here: https://docs.oracle.com/javase/8/docs/technotes/guides/security/jsse/JSSERefGuide.html#SNIExamples
+  // resolves: https://github.com/akka/akka/issues/19287
+  private def applySNI(hostname: String, params: SSLParameters): Unit = {
+    val serverName = new SNIHostName(hostname)
+    params.setServerNames(Collections.singletonList(serverName))
+  }
+
   var currentSession = engine.getSession
   applySessionParameters(firstSession)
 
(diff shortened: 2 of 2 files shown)


ktoso:
Please have a look at this SNI/TLS fix @wsargent @jrudolph

akka-http-core/src/test/scala/akka/http/impl/engine/client/HighLevelOutgoingConnectionSpec.scala
```diff
@@ -91,5 +91,13 @@ class HighLevelOutgoingConnectionSpec extends AkkaSpec with ScalaFutures {
       a[One2OneBidiFlow.OutputTruncationException.type] should be thrownBy Await.result(x, 3.second)
       binding.futureValue.unbind()
     }
+
+    "the ticket 19287" in Utils.assertAllStagesStopped {
+      val r = Await.result(Http().singleRequest(HttpRequest(uri = "https://denner-api.detailnet.ch/screen/screen-publications")), 3.seconds)
```
ktoso: will need to remove that test though - no idea yet how to make one which we can simulate locally hm hm

akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala
```diff
@@ -166,7 +174,10 @@ private[akka] class TLSActor(settings: ActorMaterializerSettings,
       case Some(TLSClientAuth.Need) ⇒ engine.setNeedClientAuth(true)
       case _                        ⇒ // do nothing
     }
-    params.sslParameters foreach (p ⇒ engine.setSSLParameters(p))
+    params.sslParameters foreach { p ⇒
+      hostInfo foreach { case (host, _) ⇒ applySNI(host, engine, p) }
```
ktoso: this should be moved to ssl-config IMO

ktoso: (Needs to be applied in `applySessionParameters`, it's not enough to do it just when we obtain the `engine`)

ktoso:
PLS BUILD

ktoso:
failure was broken pulling from github

akka-http-core/src/test/scala/akka/http/impl/engine/client/HighLevelOutgoingConnectionSpec.scala
rkuhn: … and rename the test case as well :-)

akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala
```diff
@@ -154,6 +156,12 @@ private[akka] class TLSActor(settings: ActorMaterializerSettings,
     e.setUseClientMode(role == Client)
     e
   }
+
+  def applySNI(hostname: String, e: SSLEngine, params: SSLParameters): Unit = {
```
rkuhn: the `e` parameter is not needed

akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala
rkuhn: so you propose to move `applySNI` to ssl-config—I’m not sure whether that’s worth it, because we cannot make people remember to apply it

ktoso: Yeah, it has that downside of people needing to apply it in the right place.

ktoso: `WS` does have the same problem nowadays though, so technically it would be useful to share... on the other hand, it is just 2 lines of code so maybe not worth it...

FYI @wsargent

drewhk: you can always have it here and than put it into ssl-config and remove it from here. This is internal code, so it should not matter right?

ktoso: yeah, let's roll with that :)

akka-http-core/src/test/scala/akka/http/impl/engine/client/HighLevelOutgoingConnectionSpec.scala
```diff
@@ -91,5 +91,6 @@ class HighLevelOutgoingConnectionSpec extends AkkaSpec with ScalaFutures {
       a[One2OneBidiFlow.OutputTruncationException.type] should be thrownBy Await.result(x, 3.second)
       binding.futureValue.unbind()
     }
+
```
ktoso: I removed the test, hitting random website in tests does not seem like a good idea.

ktoso:
I'll merge after validation

Issue #18620, opened 2015-10-01
add Source.blocking() for legacy API integration


rkuhn:
This shall construct a Source that defaults to the blocking dispatcher (like file IO) and includes construction as well as cleanup, possibly used like so:
~~~scala
Source.blocking(
  () => new BufferedReader(new FileReader("...")),
  file => Option(file.readLine()),
  file => file.close())
~~~
The important part is that the cleanup after normal and exceptional termination as well as cancellation.

drewhk:
Before attacking the one-liner, I would also recommend adding a more flexible API using the GraphStage internally:

    Source.blockingStage { () => new BlockingStage {
         val file = new BufferedReader(new FileReader("..."))

         def onPull(): Unit = {
            val line = file.readLine
            if (line ne null) push(line) // we can add sugar for emitting multiple elements in one go
            else complete()
         }

         override def postStop() = file.close
      }
    }

rkuhn:
BTW: the `emit` sugar is coming in #18605

hepin1989:
```Source.resource```?

agolubev:
Start working on this.

PR #19860, opened 2016-02-24
18620 add source.blocking() agolubev

agolubev:
Ref #18620

### DIFF
--- a/akka-docs/rst/java/stream/stages-overview.rst
+++ b/akka-docs/rst/java/stream/stages-overview.rst
@@ -160,6 +160,23 @@ Emit each integer in a range, with an option to take bigger steps than 1.
 
 **completes** when the end of the range has been reached
 
+unfoldResource
+^^^^^
+Wrap any resource that can be opened, queried for next element (in a blocking way) and closed using three distinct functions into a source.
+
+**emits** when there is demand and read method returns value
+
+**completes** when read function returns ``None``
+
+unfoldAsyncResource
+^^^^^
+Wrap any resource that can be opened, queried for next element and closed using three distinct functions into a source.
+Functions return ``CompletionStage`` result to achieve asynchronous processing
+
+**emits** when there is demand and ``CompletionStage`` from read function returns value
+
+**completes** when ``CompletionStage`` from read function returns ``None``
+
 queue
 ^^^^^
 Materialize a ``SourceQueue`` onto which elements can be pushed for emitting from the source. The queue contains
--- a/akka-stream-tests/src/test/scala/akka/stream/DslFactoriesConsistencySpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/DslFactoriesConsistencySpec.scala
@@ -34,7 +34,6 @@ class DslFactoriesConsistencySpec extends WordSpec with Matchers {
       (classOf[Boolean],                                classOf[akka.stream.javadsl.AsPublisher]) ::
       (classOf[scala.Function0[_]],                     classOf[akka.japi.function.Creator[_]]) ::
       (classOf[scala.Function0[_]],                     classOf[java.util.concurrent.Callable[_]]) ::
-      (classOf[scala.Function0[_]],                     classOf[akka.japi.function.Creator[_]]) ::
       (classOf[scala.Function1[_, Unit]],               classOf[akka.japi.function.Procedure[_]]) ::
       (classOf[scala.Function1[_, _]],                  classOf[akka.japi.function.Function[_, _]]) ::
       (classOf[akka.stream.scaladsl.Source[_, _]],      classOf[akka.stream.javadsl.Source[_, _]]) ::
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/UnfoldResourceAsyncSourceSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/UnfoldResourceAsyncSourceSpec.scala
@@ -1,0 +1,248 @@
+/**
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.scaladsl
+
+import java.io._
+
+import akka.Done
+import akka.actor.{ NoSerializationVerificationNeeded, ActorSystem }
+import akka.stream.ActorAttributes._
+import akka.stream.Supervision._
+import akka.stream.{ ActorMaterializer, _ }
+import akka.stream.impl.StreamSupervisor.Children
+import akka.stream.impl.{ ActorMaterializerImpl, StreamSupervisor }
+import akka.stream.testkit.TestSubscriber
+import akka.stream.testkit.Utils._
+import akka.stream.testkit.scaladsl.TestSink
+import akka.util.{ ByteString, Timeout }
+import akka.testkit.AkkaSpec
+
+import scala.concurrent.{ Await, Future, Promise }
+import scala.concurrent.duration._
+import scala.util.control.NoStackTrace
+
+class UnfoldResourceAsyncSourceSpec extends AkkaSpec(UnboundedMailboxConfig) {
+
+  val settings = ActorMaterializerSettings(system).withDispatcher("akka.actor.default-dispatcher")
+  implicit val materializer = ActorMaterializer(settings)
+
+  val manyLines = {
+    ("a" * 100 + "\n") * 10 +
+      ("b" * 100 + "\n") * 10 +
+      ("c" * 100 + "\n") * 10 +
+      ("d" * 100 + "\n") * 10 +
+      ("e" * 100 + "\n") * 10 +
+      ("f" * 100 + "\n") * 10
+  }
+  val manyLinesArray = manyLines.split("\n")
+
+  val manyLinesFile = {
+    val f = File.createTempFile("blocking-source-async-spec", ".tmp")
+    new FileWriter(f).append(manyLines).close()
+    f
+  }
+
+  val open: () ⇒ Future[BufferedReader] = () ⇒ Promise.successful(new BufferedReader(new FileReader(manyLinesFile))).future
+  val read: (BufferedReader) ⇒ Future[Option[String]] = reader ⇒ Promise.successful(Option(reader.readLine())).future
+  val close: (BufferedReader) ⇒ Future[Done] =
+    reader ⇒ {
+      reader.close()
+      Promise.successful(Done).future
+    }
+
+  "Unfold Resource Async Source" must {
+    "read contents from a file" in assertAllStagesStopped {
+      val createPromise = Promise[BufferedReader]()
+      val readPromise = Promise[Option[String]]()
+      val closePromise = Promise[Done]()
+
+      val createPromiseCalled = Promise[Done]()
+      val readPromiseCalled = Promise[Done]()
+      val closePromiseCalled = Promise[Done]()
+
+      val resource = new BufferedReader(new FileReader(manyLinesFile))
+      val p = Source.unfoldResourceAsync[String, BufferedReader](() ⇒ {
+        createPromiseCalled.success(Done)
+        createPromise.future
+      },
+        reader ⇒ {
+          readPromiseCalled.success(Done)
+          readPromise.future
+        },
+        reader ⇒ {
+          closePromiseCalled.success(Done)
+          closePromise.future
+        })
+        .runWith(Sink.asPublisher(false))
+      val c = TestSubscriber.manualProbe[String]()
+      p.subscribe(c)
+      val sub = c.expectSubscription()
+
+      sub.request(1)
+      Await.ready(createPromiseCalled.future, 3.seconds)
+      c.expectNoMsg(200.millis)
+      createPromise.success(resource)
+
+      val chunks = manyLinesArray.toList.iterator
+
+      Await.ready(readPromiseCalled.future, 3.seconds)
+      c.expectNoMsg(200.millis)
+      readPromise.success(Option(resource.readLine()))
+      c.expectNext() should ===(chunks.next())
+
+      sub.cancel()
+      Await.ready(closePromiseCalled.future, 3.seconds)
+      resource.close()
+      closePromise.success(Done)
+    }
+
+    "close resource successfully right after open" in assertAllStagesStopped {
+      val createPromise = Promise[BufferedReader]()
+      val readPromise = Promise[Option[String]]()
+      val closePromise = Promise[Done]()
+
+      val createPromiseCalled = Promise[Done]()
+      val readPromiseCalled = Promise[Done]()
+      val closePromiseCalled = Promise[Done]()
+
+      val resource = new BufferedReader(new FileReader(manyLinesFile))
+      val p = Source.unfoldResourceAsync[String, BufferedReader](() ⇒ {
+        createPromiseCalled.success(Done)
+        createPromise.future
+      },
+        reader ⇒ {
+          readPromiseCalled.success(Done)
+          readPromise.future
+        },
+        reader ⇒ {
+          closePromiseCalled.success(Done)
+          closePromise.future
+        })
+        .runWith(Sink.asPublisher(false))
+      val c = TestSubscriber.manualProbe[String]()
+      p.subscribe(c)
+      val sub = c.expectSubscription()
+
+      Await.ready(createPromiseCalled.future, 3.seconds)
+      createPromise.success(resource)
+
+      sub.cancel()
+      Await.ready(closePromiseCalled.future, 3.seconds)
+      resource.close()
+      closePromise.success(Done)
+    }
+
+    "continue when Strategy is Resume and exception happened" in assertAllStagesStopped {
+      val p = Source.unfoldResourceAsync[String, BufferedReader](open,
+        reader ⇒ {
+          val s = reader.readLine()
+          if (s != null && s.contains("b")) throw TE("") else Promise.successful(Option(s)).future
+        }, close).withAttributes(supervisionStrategy(resumingDecider))
+        .runWith(Sink.asPublisher(false))
+      val c = TestSubscriber.manualProbe[String]()
+
+      p.subscribe(c)
+      val sub = c.expectSubscription()
+
+      (0 to 49).foreach(i ⇒ {
+        sub.request(1)
+        c.expectNext() should ===(if (i < 10) manyLinesArray(i) else manyLinesArray(i + 10))
+      })
+      sub.request(1)
+      c.expectComplete()
+    }
+
+    "close and open stream again when Strategy is Restart" in assertAllStagesStopped {
+      val p = Source.unfoldResourceAsync[String, BufferedReader](open,
+        reader ⇒ {
+          val s = reader.readLine()
+          if (s != null && s.contains("b")) throw TE("") else Promise.successful(Option(s)).future
+        }, close).withAttributes(supervisionStrategy(restartingDecider))
+        .runWith(Sink.asPublisher(false))
+      val c = TestSubscriber.manualProbe[String]()
+
+      p.subscribe(c)
+      val sub = c.expectSubscription()
+
+      (0 to 19).foreach(i ⇒ {
+        sub.request(1)
+        c.expectNext() should ===(manyLinesArray(0))
+      })
+      sub.cancel()
+    }
+
+    "work with ByteString as well" in assertAllStagesStopped {
+      val chunkSize = 50
+      val buffer = Array.ofDim[Char](chunkSize)
+      val p = Source.unfoldResourceAsync[ByteString, Reader](open,
+        reader ⇒ {
+          val p = Promise[Option[ByteString]]
+          val s = reader.read(buffer)
+          if (s > 0) p.success(Some(ByteString(buffer.mkString("")).take(s))) else p.success(None)
+          p.future
+        },
+        reader ⇒ {
+          reader.close()
+          Promise.successful(Done).future
+        }).runWith(Sink.asPublisher(false))
+      val c = TestSubscriber.manualProbe[ByteString]()
+
+      var remaining = manyLines
+      def nextChunk() = {
+        val (chunk, rest) = remaining.splitAt(chunkSize)
+        remaining = rest
+        chunk
+      }
+
+      p.subscribe(c)
+      val sub = c.expectSubscription()
+
+      (0 to 121).foreach(i ⇒ {
+        sub.request(1)
+        c.expectNext().utf8String should ===(nextChunk().toString)
+      })
+      sub.request(1)
+      c.expectComplete()
+    }
+
+    "use dedicated blocking-io-dispatcher by default" in assertAllStagesStopped {
+      val sys = ActorSystem("dispatcher-testing", UnboundedMailboxConfig)
+      val materializer = ActorMaterializer()(sys)
+      try {
+        val p = Source.unfoldResourceAsync[String, BufferedReader](open,
+          read, close).runWith(TestSink.probe)(materializer)
+
+        materializer.asInstanceOf[ActorMaterializerImpl].supervisor.tell(StreamSupervisor.GetChildren, testActor)
+        val ref = expectMsgType[Children].children.find(_.path.toString contains "unfoldResourceSourceAsync").get
+        try assertDispatcher(ref, "akka.stream.default-blocking-io-dispatcher") finally p.cancel()
+      } finally shutdown(sys)
+    }
+
+    "fail when create throws exception" in assertAllStagesStopped {
+      val p = Source.unfoldResourceAsync[String, BufferedReader](() ⇒ throw TE(""),
+        read, close).runWith(Sink.asPublisher(false))
+      val c = TestSubscriber.manualProbe[String]()
+      p.subscribe(c)
+
+      c.expectSubscription()
+      c.expectError(TE(""))
+    }
+
+    "fail when close throws exception" in assertAllStagesStopped {
+      val p = Source.unfoldResourceAsync[String, BufferedReader](open,
+        read, reader ⇒ throw TE(""))
+        .runWith(Sink.asPublisher(false))
+      val c = TestSubscriber.manualProbe[String]()
+      p.subscribe(c)
+
+      val sub = c.expectSubscription()
+      sub.request(61)
+      c.expectNextN(60)
+      c.expectError()
+    }
+  }
+  override def afterTermination(): Unit = {
+    manyLinesFile.delete()
+  }
+}
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/UnfoldResourceSourceSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/UnfoldResourceSourceSpec.scala
@@ -1,0 +1,185 @@
+/**
+ * Copyright (C) 2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.scaladsl
+
+import java.io._
+
+import akka.actor.ActorSystem
+import akka.stream.ActorAttributes._
+import akka.stream.Supervision._
+import akka.stream.{ ActorMaterializer, _ }
+import akka.stream.impl.StreamSupervisor.Children
+import akka.stream.impl.{ ActorMaterializerImpl, StreamSupervisor }
+import akka.stream.testkit.TestSubscriber
+import akka.stream.testkit.Utils._
+import akka.stream.testkit.scaladsl.TestSink
+import akka.util.{ ByteString, Timeout }
+import akka.testkit.AkkaSpec
+
+import scala.concurrent.duration._
+import scala.util.control.NoStackTrace
+
+class UnfoldResourceSourceSpec extends AkkaSpec(UnboundedMailboxConfig) {
+
+  val settings = ActorMaterializerSettings(system).withDispatcher("akka.actor.default-dispatcher")
+  implicit val materializer = ActorMaterializer(settings)
+
+  val manyLines = {
+    ("a" * 100 + "\n") * 10 +
+      ("b" * 100 + "\n") * 10 +
+      ("c" * 100 + "\n") * 10 +
+      ("d" * 100 + "\n") * 10 +
+      ("e" * 100 + "\n") * 10 +
+      ("f" * 100 + "\n") * 10
+  }
+  val manyLinesArray = manyLines.split("\n")
+
+  val manyLinesFile = {
+    val f = File.createTempFile("blocking-source-spec", ".tmp")
+    new FileWriter(f).append(manyLines).close()
+    f
+  }
+
+  "Unfold Resource Source" must {
+    "read contents from a file" in assertAllStagesStopped {
+      val p = Source.unfoldResource[String, BufferedReader](() ⇒ new BufferedReader(new FileReader(manyLinesFile)),
+        reader ⇒ Option(reader.readLine()),
+        reader ⇒ reader.close())
+        .runWith(Sink.asPublisher(false))
+      val c = TestSubscriber.manualProbe[String]()
+      p.subscribe(c)
+      val sub = c.expectSubscription()
+
+      val chunks = manyLinesArray.toList.iterator
+
+      sub.request(1)
+      c.expectNext() should ===(chunks.next())
+      sub.request(1)
+      c.expectNext() should ===(chunks.next())
+      c.expectNoMsg(300.millis)
+
+      while (chunks.hasNext) {
+        sub.request(1)
+        c.expectNext() should ===(chunks.next())
+      }
+      sub.request(1)
+
+      c.expectComplete()
+    }
+
+    "continue when Strategy is Resume and exception happened" in assertAllStagesStopped {
+      val p = Source.unfoldResource[String, BufferedReader](() ⇒ new BufferedReader(new FileReader(manyLinesFile)),
+        reader ⇒ {
+          val s = reader.readLine()
+          if (s != null && s.contains("b")) throw TE("") else Option(s)
+        },
+        reader ⇒ reader.close()).withAttributes(supervisionStrategy(resumingDecider))
+        .runWith(Sink.asPublisher(false))
+      val c = TestSubscriber.manualProbe[String]()
+
+      p.subscribe(c)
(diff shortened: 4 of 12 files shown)


johanandren:
OK TO TEST

johanandren:
Needs a rebase.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/BlockingSourceSpec.scala
```diff
@@ -0,0 +1,187 @@
+/**
+ * Copyright (C) 2016 Typesafe Inc. <http://www.typesafe.com>
```
johanandren: Now "Lightbend Inc. <http://www.lightbend.com>"

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -136,3 +139,59 @@ private[akka] final class SourceQueueAdapter[T](delegate: SourceQueue[T]) extend
...
+                                                       readData: (S) ⇒ T,
+                                                       close: (S) ⇒ Unit) extends GraphStage[SourceShape[T]] {
+  val out = Outlet[T]("BlockingStreamSource.out")
+  override val shape = SourceShape(out)
+  override def initialAttributes: Attributes = DefaultAttributes.blockingSource
+
+  def createLogic(inheritedAttributes: Attributes) = new GraphStageLogic(shape) with OutHandler {
+    val decider = inheritedAttributes.get[SupervisionStrategy].map(_.decider).getOrElse(Supervision.stoppingDecider)
+    var blockingStream: S = _
+    var resumingMode = false
```
johanandren: Couldn't this be a var in the `onPull` method instead of a class field?

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -136,3 +139,59 @@ private[akka] final class SourceQueueAdapter[T](delegate: SourceQueue[T]) extend
   def offer(elem: T): CompletionStage[QueueOfferResult] = delegate.offer(elem).toJava
   def watchCompletion(): CompletionStage[Done] = delegate.watchCompletion().toJava
 }
+
+/**
+ * INTERNAL API
+ */
+private[stream] final class BlockingStreamSource[T, S](create: () ⇒ S,
```
viktorklang: I am not sure about this. Any function T => R can be blocking.

Also, "create" and "close" are not symmetrical, either "open" and "close" or "create" and "destroy".

Isn't this feature more about ReSources? I.e. things which need to be opened and closed?
Is there a reason why a Resource couldn't be non-blocking? as in: create: () => Future[S], read: S => Future[T] and S => Future[Done] ?

agolubev: You are suggesting:
* rename blocking to resource
* add non blocking version with () => Future[S], read: S => Future[T] and S => Future[Done] 
I'm Ok with this, still would like to ask ticket author. 
@rkuhn, are you Ok to make mentioned points within this PR?

akka-stream-tests/src/test/scala/akka/stream/DslFactoriesConsistencySpec.scala
```diff
@@ -34,7 +34,6 @@ class DslFactoriesConsistencySpec extends WordSpec with Matchers {
       (classOf[Boolean],                                classOf[akka.stream.javadsl.AsPublisher]) ::
       (classOf[scala.Function0[_]],                     classOf[akka.japi.function.Creator[_]]) ::
       (classOf[scala.Function0[_]],                     classOf[java.util.concurrent.Callable[_]]) ::
-      (classOf[scala.Function0[_]],                     classOf[akka.japi.function.Creator[_]]) ::
```
rkuhn: yup :+1:

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
rkuhn: As the original author I had to re-read the ticket :-) I agree that in addition to the blocking resource there should also be a variant for a non-blocking/async resource.

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -136,3 +139,58 @@ private[akka] final class SourceQueueAdapter[T](delegate: SourceQueue[T]) extend
...
+                                                       readData: (S) ⇒ T,
```
rkuhn: The read function needs to be able to signal stream completion, and since we don’t like `null` this will have to be an `S => Option[T]`.

viktorklang: @rkuhn How about Iterator[T] for batched reads?

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
rkuhn: For the asynchronous variant we could probably get away with a function `S => Future[T]` and a `object EOF extends RuntimeException with NoStackTrace`.

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -136,3 +139,58 @@ private[akka] final class SourceQueueAdapter[T](delegate: SourceQueue[T]) extend
...
+  override def initialAttributes: Attributes = DefaultAttributes.blockingSource
+
+  def createLogic(inheritedAttributes: Attributes) = new GraphStageLogic(shape) with OutHandler {
+    val decider = inheritedAttributes.get[SupervisionStrategy].map(_.decider).getOrElse(Supervision.stoppingDecider)
+    var blockingStream: S = _
+    setHandler(out, this)
+
+    override def preStart(): Unit = blockingStream = create()
+
+    override def onPull(): Unit = {
```
rkuhn: this should be `@tailrec final override def`

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -136,3 +139,58 @@ private[akka] final class SourceQueueAdapter[T](delegate: SourceQueue[T]) extend
...
+      var resumingMode = false
+      try {
+        val data = readData(blockingStream)
+        if (data != null) push(out, data)
+        else closeStage()
+      } catch {
+        case NonFatal(ex) ⇒ decider(ex) match {
+          case Supervision.Stop ⇒ failStage(ex)
```
rkuhn: need to `close(blockingStream)` here as well

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -307,6 +306,29 @@ object Source {
...
 
+  /**
+    * Creates Source that works as dispatcher for legacy blocking IO API
+    *
+    * Example:
+    * {{{
+    * Source.blocking(
+    *   () => new BufferedReader(new FileReader("...")),
+    *   reader => reader.readLine(),
+    *   reader => reader.close())
```
rkuhn: please convert to Java 8 syntax

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -136,3 +139,58 @@ private[akka] final class SourceQueueAdapter[T](delegate: SourceQueue[T]) extend
...
+    var blockingStream: S = _
+    setHandler(out, this)
+
+    override def preStart(): Unit = blockingStream = create()
+
+    override def onPull(): Unit = {
+      var resumingMode = false
+      try {
+        val data = readData(blockingStream)
+        if (data != null) push(out, data)
```
rkuhn: I see—this saves a conversion from `Optional` to `Option` as well … hmm.

viktorklang: Iterator would fix this, no?

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
rkuhn: We need three functions anyway since we need to expose the actual resource, and then it looks funny to have the intuitive single-element read return an Iterator. If a single read operation can result in multiple output elements, then why should an empty iterator imply termination? That’s the same kind of conflated API as … nevermind ;-)

viktorklang: Exactly Dr. Kuhn.

rkuhn:
I’m on the fence concerning the usage of `null`: we do not use it in any other API, and `Source.unfold` uses `Option` for this purpose.

viktorklang:
@rkuhn How about calling it unfoldResource and unfoldResourceAsync?

rkuhn:
That, Mr. Klang, is a very good proposal :+1:

agolubev:
I wish such active discussion happened on #19299

rkuhn:
One of these two is a lot harder than the other :-)

viktorklang:
When I have the flu I can only contribute to naming discussions. The
Bikeshed and all that...

-- 
Cheers,
√

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
agolubev: > The read function needs to be able to signal stream completion, and since we don’t like null this will have to be an S => Option[T]

API will look better but example will be more verbose. Maybe will find out some elegant way to express in example

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
agolubev: > For the asynchronous variant we could probably get away with a function S => Future[T] and a object EOF extends RuntimeException with NoStackTrace.

@rkuhn, We have headOption and lastOption as Future[Option[T]]. Should we stick to this? I don't like idea of throwing exception when the case is not exceptional by nature.

agolubev:
PR is ready for review.

agolubev:
Test passed after rebase

akka-stream-tests/src/test/scala/akka/stream/scaladsl/UnfoldResourceAsyncSourceSpec.scala
```diff
@@ -0,0 +1,247 @@
...
+        c.expectNext().utf8String should ===(nextChunk().toString)
+      })
+      sub.request(1)
+      c.expectComplete()
+    }
+
+    "use dedicated blocking-io-dispatcher by default" in assertAllStagesStopped {
+      val sys = ActorSystem("dispatcher-testing", UnboundedMailboxConfig)
+      val materializer = ActorMaterializer()(sys)
+      implicit val timeout = Timeout(500.millis)
```
patriknw: where is this timeout used? 500 millis might be to short in some build environments
perhaps use `remainingOrDefault`?

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -438,4 +438,40 @@ object Source {
   def queue[T](bufferSize: Int, overflowStrategy: OverflowStrategy): Source[T, SourceQueue[T]] =
     Source.fromGraph(new QueueSource(bufferSize, overflowStrategy).withAttributes(DefaultAttributes.queueSource))
 
+  /**
+   * Creates Source that works as dispatcher for legacy blocking IO API
+   *
+   * Example:
+   * {{{
+   * Source.blocking(
```
patriknw: `blocking` vs `unfoldResource`?

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -307,6 +306,46 @@ object Source {
...
+    * All exceptions thrown by `create` or `close` as well as fails of returned futures will complete the stream.
+    *
+    * `Restart` supervision strategy will close and create resource. Default strategy is `Stop` which means
+    * that stream will be terminated on error in `read` function (or future) by default.
+    *
+    *  Stream calls close and completes when `read` future success with None.
+    */
+  def unfoldResourceAsync[T, S](create: function.Creator[Future[S]],
+    read: function.Function[S, Future[Option[T]]],
+    close: function.Function[S, Future[Done]]): javadsl.Source[T, NotUsed] =
```
patriknw: shouldn't this be `CompletionStage`?

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -438,4 +438,40 @@ object Source {
...
+   * }}}
+   *
+   * You can use supervision strategy to handle exceptions for `read` function. All exceptions thrown by `create`
+   * or `close` will complete the stream.
+   *
+   * `Restart` supervision strategy will close and create blocking IO again. Default strategy is `Stop` which means
+   * that stream will be terminated on error in `read` function by default.
+   *
+   *  Stream calls close and completes when `read` returns None.
+   */
```
patriknw: perhaps say that it is using the io-dispatcher?

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -136,3 +142,144 @@ private[akka] final class SourceQueueAdapter[T](delegate: SourceQueue[T]) extend
...
+        }
+      } catch {
+        case NonFatal(ex) ⇒ decider(ex) match {
+          case Supervision.Stop ⇒
+            close(blockingStream)
+            failStage(ex)
+          case Supervision.Restart ⇒
+            restartState()
+            resumingMode = true
+          case Supervision.Resume ⇒ resumingMode = true
```
patriknw: newline to make it consistent with other cases

patriknw:
no need to update rst docs? should [this page](http://doc.akka.io/docs/akka/2.4.2/scala/stream/stages-overview.html] list all ops?

agolubev:
@patriknw thanks for the feedback. I'll fix this shortly and push code

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
agolubev: Yeah, for some reason I missed this activity of transforming all Futures to CompletionStage for java API

agolubev:
PR is ready for review

akka-docs/rst/java/stream/stages-overview.rst
```diff
@@ -160,6 +160,25 @@ Emit each integer in a range, with an option to take bigger steps than 1.
 
 **completes** when the end of the range has been reached
 
+unfoldResource
+^^^^^
+Creates Source that works as dispatcher for legacy blocking IO API. There are three functions that pass as parameters.
+First one creates/opens resource on stream start. Second one reads next data element from resource each time demand message
+came from downstream. Third function closes/destroys resource.
```
johanandren: Formulated very differently than the other stage descriptions. What about something like "Wrap any resource that can be opened, blockingly queried for next element and closed using three distinct functions into a source.". I think the actual parameters is well enough described in the ScalaDoc/JavaDoc

akka-stream-tests/src/test/scala/akka/stream/scaladsl/UnfoldResourceAsyncSourceSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+          closePromise.future
+        })
+        .runWith(Sink.asPublisher(false))
+      val c = TestSubscriber.manualProbe[String]()
+      p.subscribe(c)
+      val sub = c.expectSubscription()
+
+      sub.request(1)
+      expectMsg(Open)
+      c.expectNoMsg(200.millis)
```
johanandren: Some kind of latch would be better than depending on time passing.

johanandren: So, this is testing that we do not get an element before the createPromise is completed, that could then be done with a latch/promise in reader to signal that execution went there and checked without any timeout.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/UnfoldResourceAsyncSourceSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+      val sub = c.expectSubscription()
+
+      (0 to 19).foreach(i ⇒ {
+        sub.request(1)
+        c.expectNext() should ===(manyLinesArray(0))
+      })
+      sub.cancel()
+    }
+
+    "work with ByteString as well" in assertAllStagesStopped {
```
johanandren: Why is this a separate testcase, how does it differ from any other type that is returned from read?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/UnfoldResourceAsyncSourceSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+      val materializer = ActorMaterializer()(sys)
+      try {
+        val p = Source.unfoldResourceAsync[String, BufferedReader](open,
+          read, close).runWith(TestSink.probe)(materializer)
+
+        materializer.asInstanceOf[ActorMaterializerImpl].supervisor.tell(StreamSupervisor.GetChildren, testActor)
+        val ref = expectMsgType[Children].children.find(_.path.toString contains "unfoldResourceSourceAsync").get
+        try assertDispatcher(ref, "akka.stream.default-blocking-io-dispatcher") finally p.cancel()
+      } finally shutdown(sys)
+    }
```
johanandren: :+1:

akka-stream-tests/src/test/scala/akka/stream/scaladsl/UnfoldResourceAsyncSourceSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+      ("e" * 100 + "\n") * 10 +
+      ("f" * 100 + "\n") * 10
+  }
+  val manyLinesArray = manyLines.split("\n")
+
+  val manyLinesFile = {
+    val f = File.createTempFile("blocking-source-async-spec", ".tmp")
+    new FileWriter(f).append(manyLines).close()
+    f
+  }
```
johanandren: Do we actually need to create a file and read that to cover the behavior, wouldn't just an Iterator do perfectly well?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/UnfoldResourceAsyncSourceSpec.scala
```diff
@@ -0,0 +1,246 @@
...
+        }, close).withAttributes(supervisionStrategy(resumingDecider))
+        .runWith(Sink.asPublisher(false))
+      val c = TestSubscriber.manualProbe[String]()
+
+      p.subscribe(c)
+      val sub = c.expectSubscription()
+
+      (0 to 49).foreach(i ⇒ {
+        sub.request(1)
+        c.expectNext() should ===(if (i < 10) manyLinesArray(i) else manyLinesArray(i + 10))
```
johanandren: In addition to the comment about the file up top, I would have enjoyed test-local data here so I could easily understand the test without having to go up top to see what data the test actually runs over.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/UnfoldResourceSourceSpec.scala
```diff
@@ -0,0 +1,187 @@
...
+      val sub = c.expectSubscription()
+      sub.request(61)
+      c.expectNextN(60)
+      c.expectError(ex)
+    }
+  }
+  override def afterTermination(): Unit = {
+    manyLinesFile.delete()
+  }
+}
```
johanandren: and the comments from the previous spec applies here as well of course :)

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -136,3 +142,145 @@ private[akka] final class SourceQueueAdapter[T](delegate: SourceQueue[T]) extend
...
+ */
+private[stream] final class UnfoldResourceSource[T, S](create: () ⇒ S,
+                                                       readData: (S) ⇒ Option[T],
+                                                       close: (S) ⇒ Unit) extends GraphStage[SourceShape[T]] {
+  val out = Outlet[T]("UnfoldResourceSource.out")
+  override val shape = SourceShape(out)
+  override def initialAttributes: Attributes = DefaultAttributes.unfoldResourceSource
+
+  def createLogic(inheritedAttributes: Attributes) = new GraphStageLogic(shape) with OutHandler {
+    val decider = inheritedAttributes.get[SupervisionStrategy].map(_.decider).getOrElse(Supervision.stoppingDecider)
```
johanandren: could be a def or lazy val since it will probably not happen all the time

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -136,3 +142,145 @@ private[akka] final class SourceQueueAdapter[T](delegate: SourceQueue[T]) extend
...
+          case None    ⇒ closeStage()
+        }
+        case scala.util.Failure(t) ⇒ errorHandler(t)
+      }.invoke _
+
+      onResourceReady {
+        case resource ⇒
+          try { readData(resource).onComplete(cb) } catch errorHandler
+      }
+    }
```
johanandren: Creates a new `errorHandler` and a new `asyncCallback` on each pull?

agolubev:
@johanandren Thanks for the feedback. Will make changes today.

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -307,6 +306,53 @@ object Source {
   def queue[T](bufferSize: Int, overflowStrategy: OverflowStrategy): Source[T, SourceQueue[T]] =
     new Source(scaladsl.Source.queue[T](bufferSize, overflowStrategy).mapMaterializedValue(new SourceQueueAdapter(_)))
 
+  /**
+    * Creates Source that works as dispatcher for legacy blocking IO API
```
johanandren: "Creates a `Source` from a blocking API" (if it is Legacy or IO we don't really know here, right?)

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -307,6 +306,53 @@ object Source {
...
+    * Example:
+    * {{{
+    * Source.unfoldResource(
+    *   () -> new BufferedReader(new FileReader("...")),
+    *   reader -> reader.readLine(),
+    *   reader -> reader.close())
+    * }}}
+    *
+    * You can use supervision strategy to handle exceptions for `read` function. All exceptions thrown by `create`
+    * or `close` will complete the stream.
```
johanandren: *the* supervision strategy

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -307,6 +306,53 @@ object Source {
...
+    *
+    * `Restart` supervision strategy will close and create blocking IO again. Default strategy is `Stop` which means
+    * that stream will be terminated on error in `read` function by default.
+    *
+    *  Stream calls close and completes when `read` returns None.
```
johanandren: weird extra space before line

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -307,6 +306,53 @@ object Source {
...
+    *
+    * You can configure the default dispatcher for this Source by changing the `akka.stream.blocking-io-dispatcher` or
+    * set it for a given Source by using [[ActorAttributes]].
+    */
+  def unfoldResource[T, S](create: function.Creator[S],
+                     read: function.Function[S, Option[T]],
```
johanandren: Java API, should be `Optional` rather than `Option`

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -307,6 +306,53 @@ object Source {
...
+    * `Restart` supervision strategy will close and create resource. Default strategy is `Stop` which means
+    * that stream will be terminated on error in `read` function (or future) by default.
+    *
+    *  Stream calls close and completes when `read` future success with None.
+    *
+    * You can configure the default dispatcher for this Source by changing the `akka.stream.blocking-io-dispatcher` or
+    * set it for a given Source by using [[ActorAttributes]].
+    */
+  def unfoldResourceAsync[T, S](create: function.Creator[CompletionStage[S]],
+    read: function.Function[S, CompletionStage[Option[T]]],
```
johanandren: Same here `Optional`

johanandren:
A whole lot of nitpicking, with some performance concerns in the async version. Great work @agolubev!

akka-stream-tests/src/test/scala/akka/stream/scaladsl/UnfoldResourceAsyncSourceSpec.scala
agolubev: It's the only test case with non string but binay data pass through stream

akka-stream-tests/src/test/scala/akka/stream/scaladsl/UnfoldResourceAsyncSourceSpec.scala
agolubev: This can be any resource like connection or stream. Or anything that we have to create/read/destroy or open/read/close. So I believe it's mainly still about IO.

agolubev: I can actually do java Stream from String or Binary (still similar to file reader) but not the iterator as it's not something to close/destroy.

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
agolubev: yes - will fix this

akka-stream-tests/src/test/scala/akka/stream/scaladsl/UnfoldResourceAsyncSourceSpec.scala
agolubev: make sense - will redo

akka-docs/rst/java/stream/stages-overview.rst
```diff
@@ -160,6 +160,23 @@ Emit each integer in a range, with an option to take bigger steps than 1.
...
+Wrap any resource that can be opened, queried for next element (in a blocking way) and closed using three distinct functions into a source.
+
+**emits** when there is demand and read method returns value
+
+**completes** when read function returns ``None``
+
+unfoldAsyncResource
+^^^^^
+Creates Source that works as dispatcher for legacy blocking IO API. It's similar to ``unfoldResource`` but parameter
+functions return ``CompletionStage`` instead of plain value.
```
johanandren: This should get the same kind of wording as the one above!

agolubev: Ok

agolubev:
Build is telling I need to do rebase. Ok

drewhk:
Commits needs to be squashed though into one, though.

agolubev:
@drewhk, PR is rebased and sqaushed. (was not sure if review process is finished)

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -440,4 +440,46 @@ object Source {
...
+   * Stream calls close and completes when `read` returns None.
+   *
+   * You can configure the default dispatcher for this Source by changing the `akka.stream.blocking-io-dispatcher` or
+   * set it for a given Source by using [[ActorAttributes]].
+   */
+  def unfoldResource[T, S](create: () ⇒ S, read: (S) ⇒ Option[T], close: (S) ⇒ Unit): Source[T, NotUsed] =
+    Source.fromGraph(new UnfoldResourceSource(create, read, close))
+
+  /**
+   * Creates Source that works as dispatcher for legacy blocking API.
```
drewhk: wdym "works as a dispatcher"?

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -440,4 +440,46 @@ object Source {
...
+   * It's similar to `unfoldResource` but returns `Futures` instead of plain values.
```
drewhk: it does not return Futures but takes functions that return Futurs.

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -440,4 +440,46 @@ object Source {
...
+   * or `close` will complete the stream.
+   *
+   * `Restart` supervision strategy will close and create blocking IO again. Default strategy is `Stop` which means
+   * that stream will be terminated on error in `read` function by default.
+   *
+   * Stream calls close and completes when `read` returns None.
+   *
+   * You can configure the default dispatcher for this Source by changing the `akka.stream.blocking-io-dispatcher` or
+   * set it for a given Source by using [[ActorAttributes]].
+   */
```
drewhk: Explain the parameters (i.e. the functions taken as parameters)

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -185,3 +191,143 @@ private[akka] final class SourceQueueAdapter[T](delegate: SourceQueueWithComplet
...
+            restartState()
+            resumingMode = true
+          case Supervision.Resume ⇒
+            resumingMode = true
+        }
+      }
+      if (resumingMode) onPull()
+    }
+
+    override def onDownstreamFinish(): Unit = closeStage()
```
drewhk: why not just override postStop()? That handles then the failure cases, too.

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -185,3 +191,143 @@ private[akka] final class SourceQueueAdapter[T](delegate: SourceQueueWithComplet
...
+
+    private def restartState(): Unit = {
+      close(blockingStream)
+      blockingStream = create()
+    }
+
+    private def closeStage(): Unit =
+      try {
+        close(blockingStream)
```
drewhk: I would move this to `postStop`

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -185,3 +191,143 @@ private[akka] final class SourceQueueAdapter[T](delegate: SourceQueueWithComplet
...
+private[stream] final class UnfoldResourceSourceAsync[T, S](create: () ⇒ Future[S],
+                                                            readData: (S) ⇒ Future[Option[T]],
+                                                            close: (S) ⇒ Future[Done]) extends GraphStage[SourceShape[T]] {
+  val out = Outlet[T]("UnfoldResourceSourceAsync.out")
+  override val shape = SourceShape(out)
+  override def initialAttributes: Attributes = DefaultAttributes.unfoldResourceSourceAsync
+
+  def createLogic(inheritedAttributes: Attributes) = new GraphStageLogic(shape) with OutHandler {
+    lazy val decider = inheritedAttributes.get[SupervisionStrategy].map(_.decider).getOrElse(Supervision.stoppingDecider)
+    var resource = Promise[S]()
```
drewhk: why is this promise needed, can't you do the effects in the AsyncCallback blocks?

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -185,3 +191,143 @@ private[akka] final class SourceQueueAdapter[T](delegate: SourceQueueWithComplet
...
+
+    override def preStart(): Unit = createStream(false)
+
+    private def createStream(withPull: Boolean): Unit = {
+      val cb = getAsyncCallback[Try[S]] {
+        case scala.util.Success(res) ⇒
+          resource.success(res)
+          if (withPull) onPull()
+        case scala.util.Failure(t) ⇒ failStage(t)
+      }.invoke _
```
drewhk: I would rather use `create().onComplete(cb.invoke)` instead on L278.

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -185,3 +191,143 @@ private[akka] final class SourceQueueAdapter[T](delegate: SourceQueueWithComplet
...
+    }.invoke _
+
+    final override def onPull(): Unit = onResourceReady {
+      case resource ⇒
+        try { readData(resource).onComplete(callback) } catch errorHandler
+    }
+
+    override def onDownstreamFinish(): Unit = closeStage()
+
+    private def closeAndThen(f: () ⇒ Unit): Unit = {
```
drewhk: Shouldn't you call `setKeepGoing(true)` to make this work? Otherwise this stage stops on onDownstreamFinish.

agolubev:
@drewhk, thanks for the feedback

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
agolubev: My assumption on behavior is: In case close or open throw exception - stream should failed. 
But with having close called in postStop stream does not failed it completes nomally.
Let me know if we should rethink this.

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
agolubev: Yes, makes sense.

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
agolubev: Opened resource is shared across read and close calls. Getting resource is asynchronous event from asyncCallback (as open function returns future). At least we cannot keep effect inside one asyncCallback.

patriknw:
LGTM
@drewhk is this ready to be merged after squash?

drewhk:
LGTM

agolubev:
rebased and squashed

patriknw:
Thanks for great contribution, as always, @agolubev !

PR #19864, opened 2016-02-24
fix and document VirtualProcessor, fixes #19790

### LINKED ISSUE #19790: FAILED akka.stream.tck.VirtualProcessorTest
```
- required_spec309_requestZeroMustSignalIllegalArgumentException *** FAILED ***
[info]   java.lang.AssertionError: Expected onError(java.lang.IllegalArgumentException) within 800 ms
[info]   at org.testng.Assert.fail(Assert.java:78)
[info]   at org.reactivestreams.tck.TestEnvironment.flopAndFail(TestEnvironment.java:180)
[info]   at org.reactivestreams.tck.TestEnvironment$Receptacle.expectError(TestEnvironment.java:930)
[info]   at org.reactivestreams.tck.TestEnvironment$ManualSubscriber.expectError(TestEnvironment.java:484)
[info]   at org.reactivestreams.tck.TestEnvironment$ManualSubscriber.expectError(TestEnvironment.java:476)
[info]   at org.reactivestreams.tck.TestEnvironment$ManualSubscriber.expectErrorWithMessage(TestEnvironment.java:464)
[info]   at org.reactivestreams.tck.TestEnvironment$ManualSubscriber.expectErrorWithMessage(TestEnvironment.java:459)
[info]   at org.reactivestreams.tck.PublisherVerification$21.run(PublisherVerification.java:862)
[info]   at org.reactivestreams.tck.PublisherVerification.activePublisherTest(PublisherVerification.java:1092)
[info]   at org.reactivestreams.tck.PublisherVerification.required_spec309_requestZeroMustSignalIllegalArgumentException(PublisherVerification.java:858)
[info]   at org.reactivestreams.tck.IdentityProcessorVerification.required_spec309_requestZeroMustSignalIllegalArgumentException(IdentityProcessorVerification.java:350)
```

https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/5250/consoleFull

rkuhn:


akka-http-core/src/main/scala/akka/http/impl/engine/rendering/RenderSupport.scala
```diff
@@ -79,13 +79,13 @@ private object RenderSupport {
     override def onPush(elem: ByteString, ctx: Context[ByteString]): SyncDirective = {
       sent += elem.length
       if (sent > length)
-        throw InvalidContentLengthException(s"HTTP message had declared Content-Length $length but entity data stream amounts to more bytes")
+        ctx fail InvalidContentLengthException(s"HTTP message had declared Content-Length $length but entity data stream amounts to more bytes")
```
rkuhn: these didn’t log ERROR previously, but now they do, so back to the old behavior

akka-http-core/src/test/scala/akka/http/impl/engine/ws/MessageSpec.scala
```diff
@@ -743,7 +743,9 @@ class MessageSpec extends FreeSpec with Matchers with WithMaterializerSpec {
           """
 
         pushInput(header)
-        expectProtocolErrorOnNetwork()
+        EventFilter[ProtocolException](occurrences = 1).intercept {
+          expectProtocolErrorOnNetwork()
+        }
```
rkuhn: keeping that build log clean and green …

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/RenderSupport.scala
ktoso: I'm likely missing something here: I thought that throwing is equivalent to failStage, which I'd have thought would do the same as ctx.fail? 
`ctx.fail` seems cleaner in any case but wanted to see what I'm missing here.

rkuhn: Due to the change of logging exceptions that emanate from stages this was now rather loud in the logs while all other things HTTP are quiet. We should probably have an option to log these errors if they hit the “wire”, but this change just makes the tests less black and the code more consistent.

ktoso: I see, thanks!

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/RenderSupport.scala
```diff
@@ -79,13 +79,13 @@ private object RenderSupport {
...
       ctx.push(elem)
     }
 
     override def onUpstreamFinish(ctx: Context[ByteString]): TerminationDirective = {
       if (sent < length)
-        throw InvalidContentLengthException(s"HTTP message had declared Content-Length $length but entity data stream amounts to ${length - sent} bytes less")
+        ctx fail InvalidContentLengthException(s"HTTP message had declared Content-Length $length but entity data stream amounts to ${length - sent} bytes less")
```
johanandren: is there a reason for using infix notation here when all other method calls on `ctx` are "normal" ones?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -426,104 +424,273 @@ object StreamLayout {
...
+ * request() can only be called after the Subscriber has received the Subscription
+ * and that also means that onNext() will only happen after having transitioned into
+ * the Both state as well. The Publisher state means that if the real
+ * Publisher terminates before we get the Subscriber, we can just forget about the
+ * real one and keep an already finished one around for the Subscriber.
+ *
+ * The Subscription that is offered to the Subscriber must cancel the original
+ * Publisher if things go wrong (like `request(0)` coming in from downstream) and
+ * it must ensure that we drop the Subscriber reference when `cancel` is invoked.
+ */
```
johanandren: :+1: great explanation

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/RenderSupport.scala
rkuhn: only aesthetics for github and the diff: I didn’t want to have to change the end of the line

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -541,8 +708,9 @@ private[stream] object MaterializerSession {
 private[stream] abstract class MaterializerSession(val topLevel: StreamLayout.Module, val initialAttributes: Attributes) {
   import StreamLayout._
 
-  private var subscribersStack: List[ju.Map[InPort, Subscriber[Any]]] =
-    new ju.HashMap[InPort, Subscriber[Any]] :: Nil
+  // the contained maps store either Subscriber[Any] or VirtualPublisher, but the type system cannot express that
```
johanandren: Damn it, tricked me into reading an old blog post about union types by Miles. Very cunning move Dr Kuhn.

rkuhn:
@drewhk merge?

drewhk:
Sorry I had no time so far to look into this. Will do it soon though.

drewhk:
reviewing now

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/RenderSupport.scala
drewhk: I don't know, this failure shouldn't happen in normal circumstances or what am I missing? I.e. why are there log outputs of this error in our tests? If they are expected in a test then they should be intercepted instead. If they are expected as in ("not a big error but hey") then log a warning.

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -179,8 +177,8 @@ private[http] object StreamUtils {
 
   /** A copy of PublisherSink that allows access to the publisher through the cell but can only materialized once */
   private class OneTimePublisherSink[In](attributes: Attributes, shape: SinkShape[In], cell: OneTimeWriteCell[Publisher[In]])
-    extends PublisherSink[In](attributes, shape) {
-    override def create(context: MaterializationContext): (Subscriber[In], Publisher[In]) = {
+      extends PublisherSink[In](attributes, shape) {
+    override def create(context: MaterializationContext): (AnyRef, Publisher[In]) = {
```
drewhk: Why AnyRef? Maybe add a comment there?

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -28,7 +28,7 @@ import java.util.Optional
  */
 private[akka] abstract class SinkModule[-In, Mat](val shape: SinkShape[In]) extends Module {
 
-  def create(context: MaterializationContext): (Subscriber[In] @uncheckedVariance, Mat)
+  def create(context: MaterializationContext): (AnyRef, Mat)
```
drewhk: please comment why

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -426,104 +424,273 @@ object StreamLayout {
...
+          else rec(sub)
+        case _ =>
+          rejectAdditionalSubscriber(sub, "VirtualProcessor")
       }
+
+    if (s == null) {
+      val ex = subscriberMustNotBeNullException
+      try rec(Inert.subscriber)
+      finally throw ex // must throw NPE, rule 2:13
+    } else rec(s.asInstanceOf[Subscriber[Any]])
```
drewhk: this transition looks good

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -426,104 +424,273 @@ object StreamLayout {
...
+              getAndSet(Inert) match {
+                case Inert => // nothing to be done
+                case _     => pub.subscribe(subscriber.asInstanceOf[Subscriber[Any]])
+              }
           }
-        case sub: Subscription ⇒
-          s.cancel() // reject further Subscriptions
+        case _ =>
+          // spec violation
+          tryCancel(s)
```
drewhk: hm, but what if we got the onSubscribe in state `Both`, shouldn't we fail our downstream (who is innocent in this case)?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -426,104 +424,273 @@ object StreamLayout {
...
       }
+
+    if (s == null) {
+      val ex = subscriptionMustNotBeNullException
+      try rec(ErrorPublisher(ex, "failed-VirtualProcessor"))
+      finally throw ex // must throw NPE, rule 2:13
+    } else rec(s)
   }
```
drewhk: apart from that question, these transitions look correct.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -426,104 +424,273 @@ object StreamLayout {
...
+          set(Inert)
+          try tryOnError(s, ex)
+          finally if (t == null) throw ex // must throw NPE, rule 2:13
+        case s: Subscriber[_] => // spec violation
+          getAndSet(Inert) match {
+            case Inert => // nothing to be done
+            case _     => ErrorPublisher(ex, "failed-VirtualProcessor").subscribe(s)
+          }
+        case _ => // spec violation or cancellation race, but nothing we can do
+      }
```
drewhk: These also look correct. I would prefer an extra comment about the treatment of `null`s here.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -426,104 +424,273 @@ object StreamLayout {
...
+      throw ex // must throw NPE, rule 2:13
+    } else {
+      @tailrec def rec(): Unit =
+        get() match {
+          case Both(s) =>
+            try s.onNext(t)
+            catch {
+              case NonFatal(e) =>
+                set(Inert)
+                throw new IllegalStateException("Subscriber threw exception, this is in violation of rule 2:13", e)
```
drewhk: don't we have built-in exceptions for this?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -694,8 +862,14 @@ private[stream] abstract class MaterializerSession(val topLevel: StreamLayout.Mo
...
       val subscriber = subscribers.get(currentLayout.downstreams(out))
-      if (subscriber ne null) publisher.subscribe(subscriber)
+      if (subscriber ne null) doSubscribe(publisher, subscriber)
     }
   }
 
+  private def doSubscribe(publisher: Publisher[_ <: Any], subscriberOrVirtual: AnyRef): Unit =
+    subscriberOrVirtual match {
+      case s: Subscriber[_]       => publisher.subscribe(s.asInstanceOf[Subscriber[Any]])
+      case v: VirtualPublisher[_] => v.registerPublisher(publisher)
```
drewhk: ok, so we have to special case this again :(

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -426,104 +424,273 @@ object StreamLayout {
...
+ * The implementation of `Sink.asPublisher` needs to offer a `Publisher` that
+ * defers to the upstream that is connected during materialization. This would
+ * be trivial if it were not for materialized value computations that may even
+ * spawn the code that does `pub.subscribe(sub)` in a Future, running concurrently
+ * with the actual materialization. Therefore we implement a minimial shell here
+ * that plugs the downstream and the upstream together as soon as both are known.
+ * Using a VirtualProcessor would technically also work, but it would defeat the
+ * purpose of subscription timeouts—the subscription would always already be
+ * established from the Actor’s perspective, regardless of whether a downstream
+ * will ever be connected.
```
drewhk: Yeah. The other option would be to move the special casing to the Actor itself, which would recognize that a VirtualProcessor is attached and would modify its behavior accordingly. But that is rather ugly, so I guess this is what we have then.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
drewhk: well, you can make it extend Subscriber and leave its methods unimplemented/throw.

drewhk:
Looks good so far, with a few questions (like the `tryCancel()` in case `Both` in `onSubscribe`)

A bit sad that we must special case VirtualPublisher again.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: I debated that, but the lack of type-safety was what led me down this alternative path: now it is painfully obvious that this map does not only contain suitable Subscribers but also some weird things.

drewhk: ok, makes sense

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/RenderSupport.scala
rkuhn: The question is whether these should be raised as an exceptional server condition (and therefore logged) or considered validation errors for client input (and not logged). If the server does everything right then in my opinion there shouldn’t be any log output; in particular (and as an idealized goal) it should not be possible for a remote attacker to control the volume of log messages that are generated. My impression was that while not perfect we are actually quite close to this ideal.

rkuhn:
The only reason why it is necessary to treat this specially is the subscription timeouts—thinking about it, we don’t seem to have tests that exercise them for `Sink.asPublisher`.

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
rkuhn: seems strange to comment here—I’ll properly document the interface instead

drewhk: this is where I saw it first :)

patriknw:
@rkuhn is this complete? can we merge?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: State `Both` means that this processor currently running, everything is set up. Receiving more Subscriptions means that someone else has done something wrong and the spec prescribes that we cancel it, but why should the law-abiding citizens be punished?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: Yes, but we cannot use them to signal failures because their main purpose is to avoid passing them into onError—which reminds me of a possibly fatal flaw: if an ActorGraphInterpreter shuts down because of a spec violation then this precise mechanism may lead to non-errored (i.e. leaked) Subscribers. Opened #19978.

rkuhn:
Only added comments, merging.

Issue #19862, opened 2016-02-24
Throttle throws division by zero


drewhk:
```
[akka://default/user/StreamSupervisor-0/flow-1-0-unknown-operation] Error in stage [Throttle]: / by zero
java.lang.ArithmeticException: / by zero
at akka.stream.impl.Throttle$$anon$1$$anon$2.onPush(Throttle.scala:57)
```

drewhk:
Also, throttle started from (effectively) zero tokens in the bucket

agolubev:
@drewhk the problem here is that we can do throttling for up to one second. Because now we have multiplier 2^30 in case we need to support 24h it must be 2^46.
So the question is what minimal speed throttle should support? 1 message in 1h?

drewhk:
There is a fix coming I kust need to finish the PR
2016. márc. 12. 20:32 ezt írta ("Alexander Golubev" <
notifications@github.com>):

pirkkajokela:
I just ran in to this. Is there some set of parameters what work with throttling? (In my use, the exact pace is not that critical, so I could fudge the params if that helps.)

drewhk:
No, this cannot be fixed by parameter fudging. The fix will be in 2.4.3, due next Friday.

PR #19869, opened 2016-02-24
#19862 Fix throttle division by zero

drewhk:
- fix throttle to start with full bucket
- add smarter scaling that allows for range 1/nano - 1/10 days


drewhk:
Fixes #19869

drewhk:
I close this now, since we agreed with Roland to go with his approach. I keep this branch for the record though because the approach here might be still good in some more specialized places.

Roughly, the differences between this and Roland's approach
 - Roland's approach needs an integer division, this one doesn't
 - on the other hand, this one needs to track speed in fixed-point arithmetic to be able to use multiplication, and there is some trickyness involved in setting the right fixpoint.
 - this one also has the drawback that depending on the speed, less bits remain to track costs, i.e it is not possible to use large costs when the speed is very low.

Given that the general applicability and also understandability/maintainability is better with Roland's approach I will implement that one. This branch can be revived if someone needs the algo in a high-perf setting where the fixpoint is known in advance and division is an actual issue.

PR #19875, opened 2016-02-25
Fix common typos

jtescher:
Fixing some common typos. Unsure about typos in method names, I can remove those fixes if you would like to handle them in another PR or some other way.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
@@ -312,7 +312,7 @@ private[stream] object Fusing {
               // need to add the module so that the structural (internal) wirings can be rewritten as well
               // but these modules must not be added to any of the groups
               struct.addModule(copy, new ju.HashSet, inheritedAttributes, indent, shape)
-              struct.registerInteral(newShape, indent)
+              struct.registerInternals(newShape, indent)
 
               copy
           }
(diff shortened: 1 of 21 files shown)


johanandren:
OK TO TEST

johanandren:
LGTM (as far as I can see all the changed methods are private internal ones so isn't a problem with BC)

ktoso:
Checked the methods as well, LGTM - thanks @jtescher !

jtescher:
Awesome!

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
```diff
@@ -560,7 +560,7 @@ private[stream] object Fusing {
      * connections within imported (and not dissolved) GraphModules.
      * See also the comment in addModule where this is partially undone.
      */
-    def registerInteral(s: Shape, indent: Int): Unit = {
+    def registerInternal(s: Shape, indent: Int): Unit = {
       if (Debug) println("  " * indent + s"registerInternals(${s.outlets.map(hash)})")
```
ktoso: Oh wait, by looking at the debug println I think this method should actually be called `registerInternals` - could you fix that up in this PR @jtescher please?

jtescher: Ah sure thing :+1:

jtescher:
@ktoso looking better? Want to re-run tests?

akka-stream/src/main/scala/akka/stream/impl/fusing/Fusing.scala
```diff
@@ -560,7 +560,7 @@ private[stream] object Fusing {
      * connections within imported (and not dissolved) GraphModules.
      * See also the comment in addModule where this is partially undone.
      */
-    def registerInteral(s: Shape, indent: Int): Unit = {
+    def registerInternals(s: Shape, indent: Int): Unit = {
```
ktoso: better, thanks!

ktoso:
Better safe than sorry so let's wait for validation to complete and merge then :)
Thanks again!

ktoso:
Seems we're good here, thanks :)

Issue #19614, opened 2016-01-26
AbstractBoundedNodeQueue also suffers from nepotism


bantonsson:
The `AbstractBoundedNodeQueue` was added in Akka 2.4 and suffers from the same issues as `AbstractNodeQueue` described here #19216

Reported by @MQ-EL in #17547

drewhk:
does this need a backport?

patriknw:
no, the bounded version doesn't exist in 2.3, that's why it was missed in the original fix

drewhk:
ok, thanks.

PR #19876, opened 2016-02-25
=act #19614 Fix AbstractBoundedNodeQueue node value nepotism

MQ-EL:
1. fix node value nepotism problem
2. break BC for throwing illegal exception when user has already used `AbstractBoundedNodeQueue` with the`BalancePool` in the old version(akka).
3. remove CAS operation on the pollNode method. 

not sure about 2&3, it may be worth from my perspective. The previous discussion can be found here #17267

refs: #19614 

I waiting for use it in 2.4.3, so create this PR :)

### DIFF
--- a/akka-actor/src/main/java/akka/dispatch/AbstractBoundedNodeQueue.java
+++ b/akka-actor/src/main/java/akka/dispatch/AbstractBoundedNodeQueue.java
@@ -129,7 +129,7 @@ public abstract class AbstractBoundedNodeQueue<T> {
     }
 
     public final boolean isEmpty() {
-        return peekNode() == null;
+        return getEnq() == getDeq();
     }
 
     /**
@@ -169,6 +169,7 @@ public abstract class AbstractBoundedNodeQueue<T> {
             if (next != null) {
                 if (casDeq(deq, next)) {
                     deq.value = next.value;
+                    deq.setNext(null);
                     next.value = null;
                     return deq;
                 } // else we retry (concurrent consumers)


ktoso:
OK TO TES

ktoso:
OK TO TEST

patriknw:
@MQ-EL would you also be interested in fixing https://github.com/akka/akka/issues/19949, which I think was concluded to be a regression of those nepotism fixes?

MQ-EL:
I write some code for test/benchmark for discuss:
```scala
package akka.dispatch

object ABNQTest extends App {

//  type TestQueue = AbstractBoundedNodeQueue[Value] // akka 2.4.2, bug exists version
//  type TestNode = AbstractBoundedNodeQueue.Node[Value] // akka 2.4.2, bug exists version

  type TestQueue = BANQAfterFix[Value] // this PR version
  type TestNode = BANQAfterFix.Node[Value] // this PR version

  val allNr = 1 to 1000000 * 2
  val nq = new TestQueue(allNr.length) {}
  val batchNr = 0 to 10
  nq.add(Value(-1))
  var firstNode = nq.peekNode() // when it is prompted to old gen, then discard it(set null).

  val cycle = 1000000

  new Thread(new Runnable {
    override def run(): Unit = {
      allNr foreach {
        i ⇒
          batchNr.foreach(i ⇒ nq.add(Value(i)))
          batchNr.foreach(i ⇒ nq.poll())
          if (i > cycle && firstNode != null) {
            firstNode = null
            println("discard first node :-)")
          }
      }
      println("over!!")
    }
  }).start()
}

case class Value(i: Long)

```

this pr version gc log: 
[mpsc.txt](https://github.com/akka/akka/files/167103/mpsc.txt)
the old version gc log: 
[mpsc.txt](https://github.com/akka/akka/files/167104/mpsc.txt)

akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
```diff
@@ -391,7 +391,7 @@ class NodeMessageQueue extends AbstractNodeQueue[Envelope] with MessageQueue wit
  * Discards overflowing messages into DeadLetters.
  */
 class BoundedNodeMessageQueue(capacity: Int) extends AbstractBoundedNodeQueue[Envelope](capacity)
-  with MessageQueue with BoundedMessageQueueSemantics with MultipleConsumerSemantics {
+  with MessageQueue with BoundedMessageQueueSemantics {
```
patriknw: so `MultipleConsumerSemantics` was completely wrong?

@rkuhn what's your opinion about this change?

patriknw:
please review @rkuhn @bantonsson @viktorklang

akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
MQ-EL: may be it is better to do sth. to avoid confusion. fix MPMC bug or remove ```MultipleConsumerSemantics``` tag.

MQ-EL:
I change the commit to focus on nepotism bug only(just want to make things simple).
```peekNode``` is not thread safe, could only be used in a single thread after this commit. 

give me some info(close or keep open this pr)?

patriknw:
This PR was updated, simplified to 2 small changes, please re-review @rkuhn

rkuhn:
LGTM, thanks @MQ-EL !

MQ-EL:
u are welcome :) rkuhn patriknw

viktorklang:
:+1: LGTM

PR #19878, opened 2016-02-25
#19862 Fix division-by zero on low rates

drewhk:
- Fix throttle to start with a full bucket (first commit)
- Fix throttle to support a wide range of rates (see https://github.com/akka/akka/pull/19869 for the earlier approach that used only multiplication but was less flexible)

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/util/TokenBucketSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/util/TokenBucketSpec.scala
@@ -1,0 +1,253 @@
+/**
+ * Copyright (C) 2015-2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.util
+
+import akka.testkit.AkkaSpec
+
+import scala.util.Random
+
+class TokenBucketSpec extends AkkaSpec {
+
+  class TestBucket(_cap: Long, _period: Long) extends TokenBucket(_cap, _period) {
+    var currentTime: Long = 0L
+  }
+
+  "A Token Bucket" must {
+
+    "start full" in {
+      val bucket = new TestBucket(10, 1)
+      bucket.init()
+
+      bucket.offer(1) should ===(0)
+      bucket.offer(1) should ===(0)
+      bucket.offer(1) should ===(0)
+      bucket.offer(7) should ===(0)
+
+      bucket.offer(3) should ===(3)
+    }
+
+    "calculate correctly with different rates and capacities" in {
+      val bucketRate2 = new TestBucket(10, 2)
+      bucketRate2.init()
+
+      bucketRate2.offer(5) should ===(0)
+      bucketRate2.offer(5) should ===(0)
+      bucketRate2.offer(5) should ===(10)
+
+      val bucketRate3 = new TestBucket(8, 3)
+      bucketRate3.init()
+      bucketRate3.offer(5) should ===(0)
+      bucketRate3.offer(5) should ===(6)
+
+      bucketRate3.currentTime = 6
+      bucketRate3.offer(3) should ===(9)
+    }
+
+    "allow sending elements larger than capacity" in {
+      val bucket = new TestBucket(10, 2)
+      bucket.init()
+
+      bucket.offer(5) should ===(0)
+      bucket.offer(20) should ===(30)
+
+      bucket.currentTime = 30
+      bucket.offer(1) should ===(2)
+
+      bucket.currentTime = 34
+      bucket.offer(1) should ===(0)
+      bucket.offer(1) should ===(2)
+    }
+
+    "work with zero capacity" in {
+      val bucket = new TestBucket(0, 2)
+      bucket.init()
+
+      bucket.offer(10) should ===(20)
+
+      bucket.currentTime = 40
+      bucket.offer(10) should ===(20)
+    }
+
+    "not delay if rate is higher than production" in {
+      val bucket = new TestBucket(1, 10)
+      bucket.init()
+
+      for (time ← 0 to 100 by 10) {
+        bucket.currentTime = time
+        bucket.offer(1) should ===(0)
+      }
+
+    }
+
+    "maintain maximum capacity" in {
+      val bucket = new TestBucket(10, 1)
+      bucket.init()
+      bucket.offer(10) should ===(0)
+
+      bucket.currentTime = 100000
+      bucket.offer(20) should ===(10)
+    }
+
+    "work if currentTime is negative" in {
+      val bucket = new TestBucket(10, 1)
+      bucket.currentTime = -100 // Must be set before init()!
+      bucket.init()
+
+      bucket.offer(5) should ===(0)
+      bucket.offer(10) should ===(5)
+
+      bucket.currentTime += 10
+
+      bucket.offer(5) should ===(0)
+    }
+
+    "work if currentTime wraps over" in {
+      val bucket = new TestBucket(10, 1)
+      bucket.currentTime = Long.MaxValue - 5 // Must be set before init()!
+      bucket.init()
+
+      bucket.offer(5) should ===(0)
+      bucket.offer(10) should ===(5)
+
+      bucket.currentTime += 10
+
+      bucket.offer(5) should ===(0)
+    }
+
+    "(attempt to) maintain equal time between token renewal intervals" in {
+      val bucket = new TestBucket(5, 3)
+      bucket.init()
+
+      bucket.offer(10) should ===(15)
+
+      bucket.currentTime = 16
+      // At this point there is no token in the bucket (we consumed it at T15) but the next token will arrive at T18!
+      // A naive calculation would consider that there is 1 token needed hence we need to wait 3 units, but in fact
+      // we only need to wait 2 units, otherwise we shift the whole token arrival sequence resulting in lower rates.
+      //
+      // 0   3   9  12  15  18  21  24  27
+      // +---+---+---+---+---+---+---+---+
+      //                 ^ ^
+      //  emitted here --+ +---- currently here (T16)
+      //
+
+      bucket.offer(1) should ===(2)
+
+      bucket.currentTime = 19
+      // At 18 bucket is empty, and so is at 19. For a cost of 2 we need to wait until T24 which is 5 units.
+      //
+      // 0   3   9  12  15  18  21  24  27
+      // +---+---+---+---+---+---+---+---+
+      //                     ^ ^
+      //      emptied here --+ +---- currently here (T19)
+      //
+      bucket.offer(2) should ===(5)
+
+      // Another case
+      val bucket2 = new TestBucket(10, 3)
+      bucket2.init()
+
+      bucket2.currentTime = 4
+      bucket2.offer(6) should ===(0)
+
+      // 4 tokens remain and new tokens arrive at T6 and T9 so here we have 6 tokens remaining.
+      // We need 1 more, which will arrive at T12
+      bucket2.currentTime = 10
+      bucket2.offer(7) should ===(2)
+    }
+
+    "work with cost of zero" in {
+      val bucket = new TestBucket(10, 1)
+      bucket.init()
+
+      // Can be called any number of times
+      bucket.offer(0)
+      bucket.offer(0)
+      bucket.offer(0)
+
+      bucket.offer(10) should ===(0)
+
+      // Bucket is empty now
+      // Still can be called any number of times
+      bucket.offer(0)
+      bucket.offer(0)
+      bucket.offer(0)
+    }
+
+    "work with very slow rates" in {
+      val T = Long.MaxValue >> 10
+      val bucket = new TestBucket(10, T)
+      bucket.init()
+
+      bucket.offer(20) should ===(10 * T)
+      bucket.currentTime += 10 * T
+
+      // Collect 5 tokens
+      bucket.currentTime += 5 * T
+
+      bucket.offer(4) should ===(0)
+      bucket.offer(2) should ===(T)
+    }
+
+    "behave exactly as the ideal (naive) token bucket if offer is called with perfect timing" in {
+      val Debug = false
+
+      for {
+        capacity ← List(0, 1, 5, 10)
+        period ← List(1, 3, 5)
+        arrivalPeriod ← List(1, 3, 5)
+        startTime ← List(Long.MinValue, -1L, 0L, Long.MaxValue)
+        maxCost ← List(1, 5, 10)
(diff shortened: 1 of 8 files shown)


akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -20,21 +35,32 @@ private[stream] class Throttle[T](cost: Int,
...
+    var recordedTokens: Long = maximumBurst
+    // Tracks when the last registered token arrived, i.e. the ideal time when the *tracked* bucket jumped
+    // from n to n+1 tokens (but not when it was actually calculated).
+    // This needs to be only updated for token *additions* as subtractions have nothing to do with token arrival.
+    // If 5 tokens arrived while 4 were consumed, this needs to be still increased by 5 * nanosBetweenTokens.
+    // Be aware that this timestamp is usually in the past compared to when the ideal bucket had a new token arrival.
+    // We track here only the timestamp of the last *registered* token, i.e. corresponding to the last addition to
+    // recordedTokens (this is due to our on-need, lazy bucket update policy which makes us always slightly be in
+    // the past compared to the ideal bucket, doing large jumps in the future only when update is needed).
+    var arrivalOfLastRecordedToken: Long = System.nanoTime()
```
drewhk: this could be 0 actually, since we update it in preStart() (which is slightly closer in time to the "real" start).

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -20,21 +35,32 @@ private[stream] class Throttle[T](cost: Int,
...
   require(cost > 0, "cost must be > 0")
-  require(per.toMillis > 0, "per time must be > 0")
+  require(per.toNanos > 0, "per time must be > 0")
   require(!(mode == ThrottleMode.Enforcing && maximumBurst < 0), "maximumBurst must be > 0 in Enforcing mode")
+  require(per.toNanos >= cost, "Rates larger than 1 unit / nanosecond are not supported")
+
+  // There is some loss of precision here because of rounding, but this only happens if nanosBetweenTokens is very
+  // small which is usually at rates where that precision is highly unlikely anyway as the overhead of this stage
+  // is likely higher than the required accuracy interval.
+  private val nanosBetweenTokens = (per.toNanos.toDouble / cost).toLong
```
rkuhn: I don’t see how `.toDouble` would have an effect here (apart from introducing rounding errors for weirdly specified rates).

drewhk: yeah, right, refactor blindness here :)

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -79,11 +113,8 @@ private[stream] class Throttle[T](cost: Int,
       override def onPull(): Unit = pull(in)
     })
 
-    override def preStart(): Unit = previousTime = now()
-
-    private def now(): Long = System.nanoTime()
+    override def preStart(): Unit = arrivalOfLastRecordedToken = System.nanoTime()
```
rkuhn: this duplicates the initialization above, I think one is enough

drewhk: I noticed already :). It is actually better to do here.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -42,26 +68,34 @@ private[stream] class Throttle[T](cost: Int,
...
+                arrivalOfLastRecordedToken += elementCost * nanosBetweenTokens
+                scheduleOnce(timerName, waitNanos.nanos)
               case Enforcing ⇒ failStage(new RateExceededException("Maximum throttle throughput exceeded"))
             }
           else {
-            lastTokens = currentTokens - elementCost
-            previousTime = currentTime
+            recordedTokens = refreshedTokens - elementCost
+            // Never set this timestamp to currentNanos directly! This tracks the *ideal* time of the arrival of
+            // the last recorded token, not when we actually calculated it (currentNanos).
```
rkuhn: I think this is incorrect: this only holds unless the bucket was completely full, at which point we cannot fill it further (not even ideally), so we must set the last recorded time to “now”. With the current solution we effectively switch off the burst limit.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -71,7 +105,7 @@ private[stream] class Throttle[T](cost: Int,
     override protected def onTimer(key: Any): Unit = {
       push(out, currentElement.get)
       currentElement = None
-      lastTokens = 0
+      recordedTokens = 0 // This will be recalculated automatically once the next onPush is invoked
```
rkuhn: This should be moved upwards, next to the modification of the last recorded timestamp, to keep related things together.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -42,26 +68,34 @@ private[stream] class Throttle[T](cost: Int,
...
                 currentElement = Some(elem)
-                val waitTime = (elementCost - currentTokens) / speed
-                previousTime = currentTime + waitTime
-                scheduleOnce(timerName, waitTime.nanos)
+                val waitNanos = (elementCost - refreshedTokens) * nanosBetweenTokens
+                // due to the rounding down in tokensArrived this will track the idealized time when the last
+                // recorded integer token arrived, which in this case will be the time once we have all the
+                // tokens available to emit this element (i.e. this is calculated *ahead* as this arrival time
+                // is currently in the future and will be valid after the timer fires).
+                arrivalOfLastRecordedToken += elementCost * nanosBetweenTokens
```
rkuhn: This may still lie in the past if refreshedTokens was capped and only therefore did not contain enough for this element.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
drewhk: Well, refreshedTokens is capped, but that is wrong since I wanted the delta here, not the final. I need the delta and it will be correct.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
drewhk: ok

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
drewhk: (i.e. it is incorrect, but for a different reason)

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
rkuhn: It does not matter where it is done since the bucket is full already, therefore I’d prefer to keep the initialization at the declaration site.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
drewhk: the issue here is that the capping should not happen there. I.e. if there has been enough time elapsed that all elements were consumed from the token *and* additional time has elapsed, then we are still fine. I misplaced the capping here.

drewhk:
Ok, so I think it is better to factor out the token bucket itself and add more fine-grained tests. It will also make it reusable in other contexts.

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
drewhk: It does slightly matter, since preStart is closer to actual startup. It is not that important though.

drewhk:
Anyway, I updated the current one. I will then factor the implementation and add more tests.

drewhk:
Well, the moment when you exit the building, go too far to turn back, and you realize that you messed up again :) Anyway, this needs a better test suite that covers all the edge cases.

drewhk:
Now with a fresh new token bucket :)

akka-actor/src/main/scala/akka/util/TokenBucket.scala
```diff
@@ -0,0 +1,88 @@
...
+   *
+   * If a non-zero value is returned, it is the responsibility of the caller to not call this method before the
+   * returned delay has been elapsed (but can be called later). This class does not check or protect against early
+   * calls.
+   *
+   * @param cost How many tokens the element costs. Can be larger than the capacity of the bucket.
+   * @return
+   */
+  def offer(cost: Long): Long = {
+    require(cost >= 0, "Cost must be non-negative")
```
patriknw: how is it now, does this allocate something?

ktoso: Allocates function I guess:

```
def require(requirement: Boolean, message: => Any) 
```

drewhk: meh, replacing with an `if`

patriknw:
LGTM

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -20,70 +21,62 @@ private[stream] class Throttle[T](cost: Int,
...
+  require(per.toNanos > 0, "per time must be > 0")
   require(!(mode == ThrottleMode.Enforcing && maximumBurst < 0), "maximumBurst must be > 0 in Enforcing mode")
+  require(per.toNanos >= cost, "Rates larger than 1 unit / nanosecond are not supported")
+
+  // There is some loss of precision here because of rounding, but this only happens if nanosBetweenTokens is very
+  // small which is usually at rates where that precision is highly unlikely anyway as the overhead of this stage
+  // is likely higher than the required accuracy interval.
+  private val nanosBetweenTokens = per.toNanos / cost
+  private val timerName: String = "ThrottleTimer"
+  private val tokenBucket = new DefaultTokenBucket(maximumBurst, nanosBetweenTokens)
```
drewhk: oh no, this must go into the logic...

drewhk:
I will rebase https://github.com/akka/akka/pull/20022 on top of this

akka-actor-tests/src/test/scala/akka/util/TokenBucketSpec.scala
```diff
@@ -0,0 +1,253 @@
...
+
+import scala.util.Random
+
+class TokenBucketSpec extends AkkaSpec {
+
+  class TestBucket(_cap: Long, _period: Long) extends TokenBucket(_cap, _period) {
+    var currentTime: Long = 0L
+  }
+
+  "A Token Bucket" must {
```
viktorklang: Looks like this spec would be suited as a ScalaCheck spec?

akka-actor/src/main/scala/akka/util/TokenBucket.scala
```diff
@@ -0,0 +1,88 @@
...
+      delay
+    }
+  }
+
+}
+
+/**
+ * Default implementation of [[TokenBucket]] that uses `System.nanoTime` as the time source.
+ */
+final class DefaultTokenBucket(_cap: Long, _period: Long) extends TokenBucket(_cap, _period) {
```
viktorklang: Why not: NanoTimeTokenBucket? (Default is a role not an implementation :))

akka-stream/src/main/scala/akka/stream/impl/Throttle.scala
```diff
@@ -20,70 +21,63 @@ private[stream] class Throttle[T](cost: Int,
...
-                scheduleOnce(timerName, waitTime.nanos)
-              case Enforcing ⇒ failStage(new RateExceededException("Maximum throttle throughput exceeded"))
-            }
+          if (delayNanos == 0L) push(out, elem)
           else {
-            lastTokens = currentTokens - elementCost
-            previousTime = currentTime
-            push(out, elem)
+            if (enforcing) failStage(new RateExceededException("Maximum throttle throughput exceeded."))
+            currentElement = elem
```
viktorklang: else?

drewhk: ah yes, failStage does not exit the method :)

akka-actor-tests/src/test/scala/akka/util/TokenBucketSpec.scala
drewhk: I don't really see how that helps, the cases below all test very specific cases, so if one of them fails you already have a clue where to start (as they are human curated). The stress test in the bottom would be something more ScalaCheck-like (testing a mutable object against a reference implementation), but that kind of test here takes lot of boilerplate in ScalaCheck (at least the last time I checked).

akka-actor/src/main/scala/akka/util/TokenBucket.scala
drewhk: good point

akka-actor-tests/src/test/scala/akka/util/TokenBucketSpec.scala
viktorklang: It helps because the person who is going to add or fix test cases do not have to derive, manually, the laws governing the implementation, but can rather read them based on the generators. :)

drewhk: The kind of laws that ScalaCheck can express are very limiting. The laws here are expressed in the last test which expresses a behavioral equality (bi-simulation if you like) between the IUT and a reference model. ScalaCheck rather ugly in these kind of tests: https://github.com/rickynils/scalacheck/blob/master/doc/UserGuide.md#stateful-testing

viktorklang: Alright, lets revisit

drewhk: Maybe it is time to write a tool for such cases ourselves? I am yet to figure out how can one be nicely readable and not more boilerplate than rolling your own (most of these tests rely on specific environmental help like dummy clocks, stream probes, etc) but this is something I would be happy to investigate.

patriknw:
LGTM again
squash?

drewhk:
It was deliberately not squashed because one is a bugfix with corresponding test updates and the other is the rewrite. I found it confusing to squash together. WDYT?

patriknw:
sounds good (just checking that it wasn't an oversight), please merge when validated

drewhk:
Fixes #19862

Issue #19834, opened 2016-02-20
Replace all PushStages usages with GraphStages


agolubev:
- [x] Collect in progress by @agolubev
- [x] Drop
- [x] DropWhile in progress by @agolubev
- [x] Filter in progress by @miciek
- [x] LimitWeighted WIP by @agolubev
- [x] Log  in progress by @miciek
- [x] Map  in progress by @miciek
- [x] StartTimed in progress by @vans239,@naferx
- [x] StopTimed in progress by @vans239, @naferx
- [x] TakeWhile in progress by @vans239
- [x] TimedInterval in progress by @vans239, @naferx
- [x] UTF8Encoder WIP by @agolubev
- [x] PrepareForUserHandler WIP by @agolubev
- [x] CheckContentLengthTransformer in progress by @vans239 PR #19913
- [x] Take in progress by @trautonen 
- [x] captureTermination by @leachbj

agolubev:
will take this next if there are no objections.

hepin1989:
I think should make the new GraphStage based PushStage support the supervision strategy:)

drewhk:
I am not sure this is an urgent matter.

agolubev:
@drewhk, I'm not aware about current priorities, but will be more than happy to fix something valuable. So let me know what I should take next.

ktoso:
Marking as triaged, it's something we'd like to do but it's not on the top of the prioriy list.

vans239:
Hi, i'm trying to migrate takeWhile and checkContentLengthTransformer to GraphStage.
I can't find any predefined behavior for SupervisionStrategy. Is there any equivalent of decider for graphStage or it should be solved like in Batch with custom handling?

agolubev:
@vans239 you need to add your own logic to handle SupervisionStrategy like in Batch.

miciek:
@agolubev Can I help with this one?

agolubev:
@miciek go ahead. Just check description about what's already in progress.

hepin1989:
@agolubev you should help track the `WIP` one too:).and how about you add a new abstract GraphStage based Stage,with `SupervisionStrategy` support and then we could easily extends to it?
as @vans239 just said.

vans239:
I'd like to migrate StartTimed, StopTimed, TimedInterval

ktoso:
Thanks, marked it as in progress.

miciek:
I can do `Filter`, `Log` and `Map`.

[...]

PR #19886, opened 2016-02-26
Replaced PushStage based Drop with GraphStage #19834

trautonen:
Replaced `PushStage` based `Drop` with a `GraphStage` implementation using `SimpleLinearGraphStage` as a base.

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterStressSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterStressSpec.scala
@@ -16,6 +16,9 @@ class InterpreterStressSpec extends AkkaSpec with GraphInterpreterSpecKit {
 
   val map = Map((x: Int) ⇒ x + 1, stoppingDecider).toGS
 
+  // GraphStage can be reused
+  val dropOne = Drop(1)
+
   "Interpreter" must {
 
     "work with a massive chain of maps" in new OneBoundedSetup[Int](Vector.fill(chainLength)(map): _*) {
@@ -80,7 +83,7 @@ class InterpreterStressSpec extends AkkaSpec with GraphInterpreterSpecKit {
 
     }
 
-    "work with a massive chain of drops" in new OneBoundedSetup[Int](Vector.fill(chainLength / 1000)(Drop(1))) {
+    "work with a massive chain of drops" in new OneBoundedSetup[Int](Vector.fill(chainLength / 1000)(dropOne): _*) {
       lastEvents() should be(Set.empty)
 
       downstream.requestOne()
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -142,14 +142,25 @@ private[akka] final case class Take[T](count: Long) extends PushPullStage[T, T] 
 /**
  * INTERNAL API
  */
-private[akka] final case class Drop[T](count: Long) extends PushStage[T, T] {
-  private var left: Long = count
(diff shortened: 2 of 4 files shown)


akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -142,14 +142,27 @@ private[akka] final case class Take[T](count: Long) extends PushPullStage[T, T]
...
 
-  override def onPush(elem: T, ctx: Context[T]): SyncDirective =
-    if (left > 0) {
-      left -= 1
-      ctx.pull()
-    } else ctx.push(elem)
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
+    private var left: Long = count
+
+    setHandler(in, new InHandler {
```
johanandren: You can save two allocations here by implementing the `InHandler` and `OutHandler` with in the `GraphStageLogic` itself, and use that `setHandlers(in, out, this)`

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterStressSpec.scala
```diff
@@ -80,7 +80,7 @@ class InterpreterStressSpec extends AkkaSpec with GraphInterpreterSpecKit {
 
     }
 
-    "work with a massive chain of drops" in new OneBoundedSetup[Int](Vector.fill(chainLength / 1000)(Drop(1))) {
+    "work with a massive chain of drops" in new OneBoundedSetup[Int](Vector.fill(chainLength / 1000)(Drop(1)): _*) {
```
johanandren: Since instances of `GraphStageLogic` are reusable you can make one instance of `Drop` and re-use it 1000 times here.

johanandren:
OK TO TEST

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterStressSpec.scala
drewhk: Yes, this is a difference from PushPullStage (which required an explicit external factory)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -142,14 +142,27 @@ private[akka] final case class Take[T](count: Long) extends PushPullStage[T, T]
...
+      override def onPush(): Unit = {
+        if (left > 0) {
+          left -= 1
+          pull(in)
+        } else push(out, grab(in))
```
drewhk: You can do `passAlong(in, out)` which would set the two ports in pass-through mode, removing the branch.

akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterStressSpec.scala
rkuhn: we might be so insidious to cache or even pre-create a few (like `Drop(1)`)

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
rkuhn: that does an allocation, though, so I wouldn’t want to guess which one is preferable without micro-benchmark

rkuhn:
LGTM after folding the two handlers into the logic.

trautonen:
Updated the pull request with suggested changes:
- Implemented `InHandler` and `OutHandler` in `GraphStageLogic`
- Reused same `Drop` stage instance for stress test

johanandren:
LGTM, thanks @trautonen!

johanandren:
Refs #19834

Issue #19887, opened 2016-02-26
Current way of passing settings to sbt projects is obsolete and may produce compatible problems


strobe:
I just found that current approach of passing settings to sbt projects is little obsolete 
and may produce compatible issues with some of sbt plugins at global sbt context.
(for instance ```Cannot add dependency 'org.scala-lang\#scala-compiler;2.11.7' to configuration 'ensime-internal'"`` with ensime-sbt (https://github.com/ensime/ensime-sbt/issues/140)).

Following files affected:
- project/AkkaBuild.scala
- akka-samples/akka-sample-cluster-java/build.sbt
- akka-samples/akka-sample-cluster-scala/build.sbt
- akka-samples/akka-sample-distributed-data-java/build.sbt
- akka-samples/akka-sample-distributed-data-scala/build.sbt
- akka-samples/akka-sample-multi-node-scala/build.sbt

Currently settings passed by `Project(settings = Seq(...))` but it looks like at some cases `settings` should be modified at `Project` construction (probably by AutoPlugins).
Is exist alternative approach of passing settings by method: `Project().settings(Seq(...))`, which append settings sequence to project (sbt documentation currently used this approach http://www.scala-sbt.org/0.13/docs/Using-Plugins.html).

From sbt sources:
```scala
/** Appends settings to the current settings sequence for this project. */
def settings(ss: Def.SettingsDefinition*): Project = copy(settings = (settings: Seq[Def.Setting[_]]) ++ Def.settings(ss: _*))
```

Also Defaults.defaultSettings & Project.defaultSettings has been used at those files, but both is deprecated now (`(@deprecated("Use Defaults.coreDefaultSettings instead, combined with AutoPlugins.", "0.13.2"))`).

(I have PR with fix - will create it soon)

2m:
Fixed in #19888

PR #19888, opened 2016-02-26
sbt projects settings fix: compatible way of passing settings to Project

strobe:
issue: #19887 

I've changed approach of passing settings in following sbt project files:
- project/AkkaBuild.scala
- akka-samples/akka-sample-cluster-java/build.sbt
- akka-samples/akka-sample-cluster-scala/build.sbt
- akka-samples/akka-sample-distributed-data-java/build.sbt
- akka-samples/akka-sample-distributed-data-scala/build.sbt
- akka-samples/akka-sample-multi-node-scala/build.sbt

Main resoans for this PR:
- approach of passing settings by `Project(settings = Seq(...))` semantically differ from `Project().settings(Seq())` - and last one doing it by append setting to project instead of overwriting it completely. Current `Project(settings = Seq(...))` approach may conflict with some sbt plugins, for instance if `ensime-sbt` plugin available in scope sbt throws errors like `Cannot add dependency 'org.scala-lang\#scala-compiler;2.11.7' to configuration 'ensime-internal'"` (see some previous history here: https://github.com/ensime/ensime-sbt/issues/140)
- `Project.defaultSettings` and `Defaults.defaultSettings` is deprecated `(@deprecated("Use Defaults.coreDefaultSettings instead, combined with AutoPlugins.", "0.13.2"))`. And It may conflict with `Project().settings(Seq())` approach of passing project settings.
- as nice benefit we getting ability to use Ensime for Akka development, but it's not only about that it allow us to compile Akka witout `Cannot add dependency 'org.scala-lang'...` errors if ensime-sbt added to `~/.sbt/0.13/plugins.sbt` for working with different projects.

I'm not fully sure that those changes not affect something else, but seems like it works well (`validatePullRequest` passed well on local machine) (hope someone could help me to checking it).

### DIFF
--- a/akka-samples/akka-sample-cluster-java/build.sbt
+++ b/akka-samples/akka-sample-cluster-java/build.sbt
@@ -4,9 +4,11 @@ import com.typesafe.sbt.SbtMultiJvm.MultiJvmKeys.MultiJvm
 val akkaVersion = "2.4-SNAPSHOT"
 
 val project = Project(
-  id = "akka-sample-cluster-java",
-  base = file("."),
-  settings = Project.defaultSettings ++ SbtMultiJvm.multiJvmSettings ++ Seq(
+    id = "akka-sample-cluster-java",
+    base = file(".")
+  )
+  .settings(SbtMultiJvm.multiJvmSettings: _*)
+  .settings(
     name := "akka-sample-cluster-java",
     version := "2.4-SNAPSHOT",
     scalaVersion := "2.11.7",
@@ -45,4 +47,4 @@ val project = Project(
     },
     licenses := Seq(("CC0", url("http://creativecommons.org/publicdomain/zero/1.0")))
   )
-) configs (MultiJvm)
+  .configs (MultiJvm)
--- a/project/AkkaBuild.scala
+++ b/project/AkkaBuild.scala
@@ -36,13 +36,10 @@ object AkkaBuild extends Build {
     version             := "2.4-SNAPSHOT"
   )
 
-  lazy val root = Project(
-    id = "akka",
-    base = file("."),
-    settings = parentSettings ++ Release.settings ++
-      SphinxDoc.akkaSettings ++ Dist.settings ++ s3Settings ++
-      UnidocRoot.akkaSettings ++
-      Protobuf.settings ++ Seq(
+  lazy val rootSettings = parentSettings ++ Release.settings ++
+    SphinxDoc.akkaSettings ++ Dist.settings ++ s3Settings ++
+    UnidocRoot.akkaSettings ++
+    Protobuf.settings ++ Seq(
       parallelExecution in GlobalScope := System.getProperty("akka.parallelExecution", parallelExecutionByDefault.toString).toBoolean,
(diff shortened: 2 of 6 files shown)


johanandren:
OK TO TEST

ktoso:
In general I don't mind the change I guess, just two questions:

- have we made sure the "new style" is used consistently in all projects?
- have you given the startup time a look - if the new style comes with a startup time penalty or not?

Otherwise LGTM I guess, would be nice to get @2m's input first though

akka-samples/akka-sample-cluster-java/build.sbt
```diff
@@ -5,44 +5,44 @@ val akkaVersion = "2.4-SNAPSHOT"
...
-            testResults.overall
-        Tests.Output(overall,
-          testResults.events ++ multiNodeResults.events,
-          testResults.summaries ++ multiNodeResults.summaries)
-    },
-    licenses := Seq(("CC0", url("http://creativecommons.org/publicdomain/zero/1.0")))
-  )
-) configs (MultiJvm)
+  base = file(".")
+).settings(Defaults.coreDefaultSettings ++ SbtMultiJvm.multiJvmSettings ++ Seq(
```
2m: Adding `Defaults.coreDefaultSettings` manually is not required as it is added by the `sbt.plugins.CorePlugin` AutoPlugin which is enabled by default on all projects.

I think this change should be reformatted like so:
```
val project = Project(
    id = "akka-sample-cluster-java",
    base = file(".")
  )
  .settings(SbtMultiJvm.multiJvmSettings: _*)
  .settings(
    name := "akka-sample-cluster-java",
    ...
  )
  .configs(MultiJvm)
```

Here I removed `Defaults.coreDefaultSettings`  and also used `settings` method override which accepts var args. This also has a nice little advantage that it does not change setting identation. :)

project/AkkaBuild.scala
```diff
@@ -412,7 +413,7 @@ object AkkaBuild extends Build {
     ) ++
     resolverSettings
 
-  lazy val baseSettings = Defaults.defaultSettings
+  lazy val baseSettings = Defaults.coreDefaultSettings
 
   lazy val parentSettings = baseSettings ++ Seq(
```
2m: `baseSettings` can be removed all together as it only contains `Defaults.coreDefaultSettings` which is added by the Core AutoPlugin to all projects.

strobe:
@2m
I will make that changes.
thanks for help

2m:
Thank you! Another step towards more modern and clean sbt build definition for Akka. :)

rkuhn:
Yes, indeed, this is looking good, we should make sure that we use this style consistently.

strobe:
@2m @ktoso 
code has been updated

Sorry, but I did squash which I had to do later.

jrudolph:
:+1: I just tested the upcoming new version of sbt-boilerplate with the akka build and it doesn't work for exactly the reason given here. (sbt-boilerplate's `sourceGenerators <+=` and Defaults.defaultSettings' `sourceGenerators := Nil` were applied in the wrong order resulting in an empty value)

ktoso:
Great, thanks a lot everyone!

PR #19889, opened 2016-02-26
add Source.queue termination support, fixes #19877

### LINKED ISSUE #19877: support stream completion in Source.queue

rkuhn:


akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -15,124 +15,167 @@ import scala.compat.java8.FutureConverters._
...
-        } else if (pulled) {
-          push(out, elem)
-          pulled = false
-          promise.success(QueueOfferResult.Enqueued)
-        } else pendingOffer = Some(tuple)
-      })
+        case Failure(ex) ⇒
+          completion.failure(ex)
+          failStage(ex)
+      }
```
johanandren: I would have liked a bit less cyclomatic complexity here, can't it be split into smaller parts?

johanandren:
LGTM!

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -15,124 +15,167 @@ import scala.compat.java8.FutureConverters._
...
-            pulled = false
+      private val callback: AsyncCallback[Input[T]] = getAsyncCallback {
+        case offer @ Offer(elem, promise) ⇒
+          if (maxBuffer != 0) {
+            bufferElem(offer)
+            if (isAvailable(out)) push(out, buffer.dequeue())
+          } else if (isAvailable(out)) {
+            push(out, elem)
+            promise.success(QueueOfferResult.Enqueued)
+          } else if (pendingOffer.isEmpty) pendingOffer = Some(offer)
```
patriknw: I'd prefer a newline here (todays most minor nitpick ;-))

akka-stream/src/main/scala/akka/stream/scaladsl/Queue.scala
```diff
@@ -32,6 +32,23 @@ trait SourceQueue[T] {
 }
 
 /**
+ * This trait adds completion support to [[SourceQueue]].
+ */
+trait SourceQueueWithComplete[T] extends SourceQueue[T] {
```
patriknw: is there a reason why those methods are not added to `SourceQueue`?

patriknw:
no javadsl?

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
rkuhn: I see you are a man of good taste!

akka-stream/src/main/scala/akka/stream/scaladsl/Queue.scala
rkuhn: Yes: the trait is public, so adding methods to it would not in general be binary compatible.

rkuhn:
good catch about the Java DSL, I forgot to change the return types

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
rkuhn: Would empty lines between the case statements be a suitable substitute?

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -15,124 +15,173 @@ import scala.compat.java8.FutureConverters._
 /**
  * INTERNAL API
  */
-final private[stream] class QueueSource[T](maxBuffer: Int, overflowStrategy: OverflowStrategy) extends GraphStageWithMaterializedValue[SourceShape[T], SourceQueue[T]] {
-  type Offered = Promise[QueueOfferResult]
+private[stream] object QueueSource {
+  sealed trait Input[+T]
+  case class Offer[+T](elem: T, promise: Promise[QueueOfferResult]) extends Input[T]
```
viktorklang: final?

akka-stream/src/main/scala/akka/stream/impl/Sources.scala
```diff
@@ -15,124 +15,173 @@ import scala.compat.java8.FutureConverters._
...
+  case object Completion extends Input[Nothing]
+  case class Failure(ex: Throwable) extends Input[Nothing]
```
viktorklang: final?

rkuhn:
sprinkled some `final`—but needs one more LGTM

akka-stream/src/main/scala/akka/stream/scaladsl/Queue.scala
```diff
@@ -32,6 +32,23 @@ trait SourceQueue[T] {
...
+   * Complete the stream normally. Use `watchCompletion` to be notified of this
+   * operation’s success.
+   */
+  def complete(): Unit
+
+  /**
+   * Complete the stream with a failure. Use `watchCompletion` to be notified of this
+   * operation’s success.
+   */
+  def fail(ex: Throwable): Unit
```
viktorklang: Is this operation idempotent?
What happens if one tries to send more elements after calling it?

akka-stream/src/main/scala/akka/stream/scaladsl/Queue.scala
```diff
@@ -32,6 +32,23 @@ trait SourceQueue[T] {
...
 
 /**
+ * This trait adds completion support to [[SourceQueue]].
+ */
+trait SourceQueueWithComplete[T] extends SourceQueue[T] {
+  /**
+   * Complete the stream normally. Use `watchCompletion` to be notified of this
+   * operation’s success.
+   */
+  def complete(): Unit
```
viktorklang: Is this operation idempotent?
What happens if one tries to send more elements after calling it?

viktorklang:
@rkuhn besides a couple of questions, LGTM

akka-stream/src/main/scala/akka/stream/scaladsl/Queue.scala
drewhk: I think they are, and yes it should be documented :)

drewhk:
LGTM

akka-stream/src/main/scala/akka/stream/scaladsl/Queue.scala
viktorklang: The problem is if you call `fail` with different Throwables and get no feedback that it had already been `failed`/completed?

Should the methods return booleans to signal success/failure?

rkuhn: not sure what you are getting at here: there is no feedback, so you can call it any number of times to get back any number of `Unit` values—is that not idempotent enough?

rkuhn: to your second question: sending more elements will result in failed futures being returned

viktorklang: >not sure what you are getting at here: there is no feedback, so you can call it any number of times to get back any number of Unit values—is that not idempotent enough?

My point is that if you *Accidentally* or as a result of a race-condition call fail after complete, or fail after fail, or complete after complete, you have no indicator that you have a logic bug?

rkuhn: Yes, that is correct. I guess we disagree on whether that should be fixed, though: there are many bugs that we don’t detect, beginning with `def !(msg: Any)(sender: ActorRef): Unit`.

viktorklang: that method does not exhibit the problem since it has the exact same behavior no matter which message is sent (null excluded).

rkuhn: Hmm, I don’t follow: if you _accidentally_ send a message after PoisonPill, isn’t that the exact same issue?

viktorklang: @rkuhn One would hope so, but it isn't, because PriorityMailbox etc.

rkuhn: This leaves me even more confused. Do we agree now or don’t we? I don’t think there is anything to fix here because declaring this kind of problem worthy of a fix would be inconsistent with many other occasions where we will not or cannot do the equivalent thing.

viktorklang: We do it for Promise. (tryComplete, trySuccess, tryFailure)
And, technically we could store a Promise locally and do: if (promise.trySuccess(())) sendComplete()

Anyway. I digress.

rkuhn: Okay, so I gather that we do agree: it might be fixable in principle, but it is too much effort because all that is gained is that erroneous user code is flagged as such earlier than it is now (assuming test coverage).

viktorklang: Yeah, we have bigger fish to fry

akka-stream/src/main/scala/akka/stream/scaladsl/Queue.scala
patriknw: when will scala have default methods?

patriknw:
LGTM, please rebase

johanandren:
MiMa failure, plus unrelated `akka.cluster.sharding.PersistentClusterSharding`

Issue #19863, opened 2016-02-24
FAILED: ClusterShardingGetStatsSpec


drewhk:
https://jenkins.akka.io:8498/job/akka-nightly/3083/consoleFull

```
[JVM-3] - must return stats after a node leaves *** FAILED ***
[JVM-3]   0 did not equal 2 (ClusterShardingGetStatsSpec.scala:217)
[JVM-3] [WARN] [02/23/2016 22:43:59.735] [ClusterShardingGetStatsSpec-akka.actor.default-dispatcher-18] [akka.tcp://ClusterShardingGetStatsSpec@localhost:56753/system/sharding/Ping] Trying to register to coordinator at [Some(ActorSelection[Anchor(akka://ClusterShardingGetStatsSpec/), Path(/system/sharding/PingCoordinator/singleton/coordinator)])], but no acknowledgement. Total [4] buffered messages.
```

johanandren:
Possibly related to #19601 that got merged two days ago.

drewhk:
https://jenkins.akka.io:8498/job/akka-nightly/3091/consoleFull

drewhk:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/3535/consoleFull

2m:
(Mar 7, 2016 1:17:00 AM) https://jenkins.akka.io:8498/job/akka-nightly/3120/consoleFull
(Mar 4, 2016 11:19:00 PM) https://jenkins.akka.io:8498/job/akka-multi-node-nightly/3566/consoleFull
(Mar 4, 2016 9:17:00 PM) https://jenkins.akka.io:8498/job/akka-nightly/3113/consoleFull
(Mar 4, 2016 1:17:00 AM) https://jenkins.akka.io:8498/job/akka-nightly/3111/consoleFull
(Mar 3, 2016 1:17:00 AM) https://jenkins.akka.io:8498/job/akka-nightly/3108/consoleFull
(Mar 2, 2016 6:19:00 AM) https://jenkins.akka.io:8498/job/akka-multi-node-nightly/3557/consoleFull
(Mar 1, 2016 11:19:00 PM) https://jenkins.akka.io:8498/job/akka-multi-node-nightly/3554/consoleFull
(Mar 1, 2016 4:19:00 AM) https://jenkins.akka.io:8498/job/akka-multi-node-nightly/3552/consoleFull
[...]

PR #19894, opened 2016-02-26
Fix for #19863 ClusterShardingGetStatsSpec failing

johanandren:
Refs #19863

### DIFF
--- a/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
+++ b/akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
@@ -10,11 +10,7 @@ import akka.util.Timeout
 import akka.pattern.{ ask, pipe }
 import akka.actor._
 import akka.cluster.Cluster
-import akka.cluster.ClusterEvent.ClusterDomainEvent
-import akka.cluster.ClusterEvent.CurrentClusterState
-import akka.cluster.ClusterEvent.MemberEvent
-import akka.cluster.ClusterEvent.MemberRemoved
-import akka.cluster.ClusterEvent.MemberUp
+import akka.cluster.ClusterEvent._
 import akka.cluster.Member
 import akka.cluster.MemberStatus
 
@@ -437,7 +433,9 @@ class ShardRegion(
       else if (matchingRole(m))
         changeMembers(membersByAge - m)
 
-    case _ ⇒ unhandled(evt)
+    case _: MemberEvent ⇒ // these are expected, no need to warn about them
+
+    case _              ⇒ unhandled(evt)
   }
 
   def receiveCoordinatorMessage(msg: CoordinatorMessage): Unit = msg match {
--- a/akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingGetStatsSpec.scala
+++ b/akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingGetStatsSpec.scala
@@ -4,7 +4,7 @@
 package akka.cluster.sharding
 
 import akka.actor._
-import akka.cluster.Cluster
+import akka.cluster.{ MemberStatus, Cluster }
 import akka.cluster.ClusterEvent.CurrentClusterState
 import akka.remote.testconductor.RoleName
 import akka.remote.testkit.{ MultiNodeConfig, MultiNodeSpec, STMultiNodeSpec }
@@ -19,7 +19,7 @@ object ClusterShardingGetStatsSpec {
   case object Pong
 
   class ShardedActor extends Actor with ActorLogging {
-    log.info(self.path.toString)
+    log.info(s"entity started {}", self.path)
     def receive = {
       case Stop    ⇒ context.stop(self)
       case _: Ping ⇒ sender() ! Pong
@@ -49,13 +49,15 @@ object ClusterShardingGetStatsSpecConfig extends MultiNodeConfig {
     akka.loglevel = INFO
     akka.actor.provider = "akka.cluster.ClusterActorRefProvider"
     akka.remote.log-remote-lifecycle-events = off
+    akka.log-dead-letters-during-shutdown = off
     akka.cluster.metrics.enabled = off
     akka.cluster.auto-down-unreachable-after = 0s
     akka.cluster.sharding {
-      coordinator-failure-backoff = 3s
-      shard-failure-backoff = 3s
       state-store-mode = "ddata"
+      updating-state-timeout = 2s
+      waiting-for-state-timeout = 2s
     }
+    akka.actor.warn-about-java-serializer-usage=false
     """))
 
   nodeConfig(first, second, third)(ConfigFactory.parseString(
@@ -99,7 +101,7 @@ abstract class ClusterShardingGetStatsSpec extends MultiNodeSpec(ClusterSharding
     enterBarrier(from.name + "-joined")
   }
 
-  var shardActor = Actor.noSender
+  lazy val region = ClusterSharding(system).shardRegion(shardTypeName)
 
   "Inspecting cluster sharding state" must {
 
@@ -109,17 +111,18 @@ abstract class ClusterShardingGetStatsSpec extends MultiNodeSpec(ClusterSharding
       join(second)
       join(third)
 
-      // make sure all nodes has joined
-      awaitAssert {
-        Cluster(system).sendCurrentClusterState(testActor)
-        expectMsgType[CurrentClusterState].members.size === 3
+      // make sure all nodes are up
+      within(10.seconds) {
+        awaitAssert {
+          Cluster(system).state.members.count(_.status == MemberStatus.Up) should === (4)
+        }
       }
 
       runOn(controller) {
         startProxy()
       }
       runOn(first, second, third) {
-        shardActor = startShard()
+        startShard()
       }
 
       enterBarrier("sharding started")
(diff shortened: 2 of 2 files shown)


akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingGetStatsSpec.scala
```diff
@@ -109,17 +109,18 @@ abstract class ClusterShardingGetStatsSpec extends MultiNodeSpec(ClusterSharding
...
       join(third)
 
-      // make sure all nodes has joined
-      awaitAssert {
-        Cluster(system).sendCurrentClusterState(testActor)
-        expectMsgType[CurrentClusterState].members.size === 3
+      // make sure all nodes are up
+      within(10.seconds) {
+        awaitAssert {
+          Cluster(system).state.members.count(_.status == MemberStatus.Up) === 4
```
johanandren: This and one like it below is the actual fix, the rest is boyscouting

drewhk: so previous one was 3, and also include Joining state?

drewhk: btw, isn't there a reusable assertAllNodesUp() somewhere? @patriknw ?

johanandren: Yes, before it got past here with nodes still Joining and then timed out later because getting all nodes up before sharding got started and in contact took time.

drewhk:
LGTM

akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingGetStatsSpec.scala
patriknw: @johanandren but is this really an assert? shouldn't it be
```Cluster(system).state.members.count(_.status == MemberStatus.Up) should === (4)```

File: akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingGetStatsSpec.scala [hunk 2]
```diff
@@ -130,12 +131,11 @@ abstract class ClusterShardingGetStatsSpec extends MultiNodeSpec(ClusterSharding
...
           val probe = TestProbe()
-          val region = ClusterSharding(system).shardRegion(shardTypeName)
           region.tell(ShardRegion.GetClusterShardingStats(10.seconds.dilated), probe.ref)
           val shardStats = probe.expectMsgType[ShardRegion.ClusterShardingStats]
-          shardStats.regions.size shouldEqual 3
-          shardStats.regions.values.map(_.stats.size).sum shouldEqual 0
-          shardStats.regions.keys.forall(_.hasGlobalScope) should be(true)
+          shardStats.regions.size === 3
+          shardStats.regions.values.map(_.stats.size).sum === 0
+          shardStats.regions.keys.forall(_.hasGlobalScope) === true
```
patriknw: same here?

akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingGetStatsSpec.scala
```diff
@@ -144,23 +144,19 @@ abstract class ClusterShardingGetStatsSpec extends MultiNodeSpec(ClusterSharding
...
           awaitAssert {
             val pingProbe = TestProbe()
             // trigger starting of 2 entities on first and second node
             // but leave third node without entities
-            (1 to 6).filterNot(_ % 3 == 0).foreach(n ⇒ region.tell(Ping(n), pingProbe.ref))
+            List(1, 2, 4, 6).foreach(n ⇒ region.tell(Ping(n), pingProbe.ref))
             pingProbe.receiveWhile(messages = 4) {
               case Pong ⇒ ()
-            }
+            }.size === 4
```
patriknw: what does this do (if it was `should ===`) that is not detected by `receiveWhile(messages = 4)`?

johanandren: I was under the impression that `receiveWhile(messages = 4)` is a max 4, not an assert, but reading the sources of it now I think I was wrong. Seems superfluous given that, will fix.

akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingGetStatsSpec.scala
johanandren: I don't want it to pass here until there are 4 nodes up and seen, failing the assertion throws an exception, so that will make it try again for up to 10 seconds? (Could be another case of not knowing the testkit well enough)

johanandren: Ah, misread your comment, I will double check this, but I am pretty sure just `===` also asserts.

johanandren: Nope, my bad. Will fix.

akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingGetStatsSpec.scala
```diff
@@ -207,15 +208,15 @@ abstract class ClusterShardingGetStatsSpec extends MultiNodeSpec(ClusterSharding
...
           awaitAssert {
             val probe = TestProbe()
-            val region = ClusterSharding(system).shardRegion(shardTypeName)
-            region.tell(ShardRegion.GetClusterShardingStats(10.seconds.dilated), probe.ref)
+            region.tell(ShardRegion.GetClusterShardingStats(20.seconds.dilated), probe.ref)
             val regions = probe.expectMsgType[ShardRegion.ClusterShardingStats].regions
-            regions.size shouldEqual 2
-            regions.values.flatMap(_.stats.values).sum shouldEqual 4
+            regions.size === 2
+            system.log.warning(regions.toString)
```
johanandren: ups. left over logging. should be removed

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
```diff
@@ -437,7 +433,9 @@ class ShardRegion(
       else if (matchingRole(m))
         changeMembers(membersByAge - m)
 
-    case _ ⇒ unhandled(evt)
+    case _: MemberEvent ⇒ // these are expected, no need to warn about them
+
```
ktoso: why the weird newline?

akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingGetStatsSpec.scala
```diff
@@ -19,7 +19,7 @@ object ClusterShardingGetStatsSpec {
   case object Pong
 
   class ShardedActor extends Actor with ActorLogging {
-    log.info(self.path.toString)
+    log.info(s"entity started ${self.path}")
```
ktoso: in logging plz `log.info(s"entity started {}", self.path)`

akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingGetStatsSpec.scala
ktoso: fixed in more places, except here :)

akka-cluster-sharding/src/main/scala/akka/cluster/sharding/ShardRegion.scala
johanandren: Because the two cases above in the block had one I guess

johanandren:
@patriknw not sure what to do about this, one node leaving cluster, and then the other nodes assertAwait that the cluster size changes within 10 seconds before moving on with the test, but it's racing, works most of the time, some times not, is there a better way to do this?

johanandren:
Just increase that timeout? Looking at `ClusterShardingSpec` I see that it has 30 seconds for the case when it crashes a node.

akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingGetStatsSpec.scala
```diff
@@ -207,15 +208,14 @@ abstract class ClusterShardingGetStatsSpec extends MultiNodeSpec(ClusterSharding
...
 
-      runOn(first, second) {
-        within(10.seconds) {
+      runOn(controller) {
+        within(20.seconds) {
           awaitAssert {
             val probe = TestProbe()
-            val region = ClusterSharding(system).shardRegion(shardTypeName)
-            region.tell(ShardRegion.GetClusterShardingStats(10.seconds.dilated), probe.ref)
+            region.tell(ShardRegion.GetClusterShardingStats(20.seconds.dilated), probe.ref)
```
patriknw: could use `remaining` here

akka-cluster-sharding/src/multi-jvm/scala/akka/cluster/sharding/ClusterShardingGetStatsSpec.scala
```diff
@@ -175,29 +173,32 @@ abstract class ClusterShardingGetStatsSpec extends MultiNodeSpec(ClusterSharding
...
-      runOn(first) {
+      runOn(controller) {
         Cluster(system).leave(node(third).address)
       }
 
-      runOn(third) {
-        watch(shardActor)
-        expectTerminated(shardActor, 15.seconds)
+      runOn(controller, first, second) {
+        within(10.seconds) {
```
patriknw: it might take longer than 10 seconds for a node to leave completely, failure detector must trigger

ktoso:
Can we merge this? The test is failing a lot in nightlies

johanandren:
I think it is fine to merge.

Issue #19892, opened 2016-02-26
Add Stream "KillSwitch"


drewhk:
This is an object that provides a Flow, which can be materialized arbitrary times, and a method that allows remotely failing or completing (and of course cancelling) its materialized Flows.

rkuhn:
This is the connected variant of the Breaker that already exists as internal API: while this one cuts all its use-points at the same time, the Breaker materializes to cut points that are broken individually.

PR #19895, opened 2016-02-26
#19892: KillSwitch (externally controllable stream completion)

drewhk:
Refs #19892


drewhk:
Needs Java APIs and probably mention in the manual.

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
```diff
@@ -0,0 +1,119 @@
...
+  /**
+   * Retrurns a typed Flow of a requested type that will be linked to this [[KillSwitch]] instance. By invoking
+   * [[KillSwitch#shutdown()]] or [[KillSwitch#abort()]] all running instances of all provided [[Flow]]s by this
+   * switch will be stopped normally or failed.
+   *
+   * @tparam T Type of the elements the Flow will forward
+   * @return   A reusable [[Flow]] that is linked with the switch. The materialized value provided is this switch itself.
+   */
+  def flow[T]: Flow[T, T, KillSwitch] = _flow.asInstanceOf[Flow[T, T, KillSwitch]]
+
```
drewhk: I think it also makes sense to expose the `Future[Done]`, it can be useful to compose more complex shutdown scenarios.

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
```diff
@@ -0,0 +1,119 @@
...
+        setHandler(shape.in, this)
+        setHandler(shape.out, this)
+
+        override def onPush(): Unit = push(shape.out, grab(shape.in))
+        override def onPull(): Unit = pull(shape.in)
+
+        override def preStart(): Unit = {
+          val cb = getAsyncCallback[Try[Done]](onSwitch)
+          // callback.invoke is a simple actor send, so it is fine to run on the invoking thread
+          shutdownPromise.future.onComplete(cb.invoke)(akka.dispatch.ExecutionContexts.sameThreadExecutionContext)
```
drewhk: I should probably just check the value of the Future first if available.

viktorklang: Why `sameThreadExecutionContext`?

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
viktorklang: Did we consider not exposing the `KillSwitch` type and instead have it be a `Flow[T, T, Promise[Done]]`?
(It would reduce the number of concepts and minimize the new-API surface area)

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
```diff
@@ -0,0 +1,119 @@
...
+import akka.stream.{ Attributes, Outlet, Inlet, FlowShape }
+import akka.stream.stage._
+
+import scala.concurrent.Promise
+import scala.util.{ Failure, Success, Try }
+
+/**
+ * Creates a [[KillSwitch]] that can be used to externally control the completion of various streams.
+ *
+ * @see KillSwitch
```
patriknw: is this the right syntax?
I think it should be `[[KillSwitch]]`

patriknw:
LGTM after javadsl + docs

viktorklang:
@patriknw WDYT about my comments w.r.t. API surface area / new concepts-toll?

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
patriknw: @viktorklang In general I like the idea of reducing concepts, but unless I have misunderstood things your proposal will not support one "global" switch for all materialized instances, since the promise is returned as materialized value instead of being provided by the user.

viktorklang: @patriknw Perhaps I misunderstood, I thought the KillSwitch was provided as a materialized value too?

"+  def flow[T]: Flow[T, T, KillSwitch] = _flow.asInstanceOf[Flow[T, T, KillSwitch]]"

patriknw: it might also be a materialized value, but the test shows
```
val switch = KillSwitch("switch")
TestSource.probe[Int].via(switch.flow)
```

viktorklang: @patriknw Having something which takes a Future[Done] would be more general?

~~~scala
val p = Promise[Done]
TestSource.probe[Int].killSwitch(p.future)
~~~

patriknw: then we had the same thought, so it's must be good ;-)
let's see if Endre has a reason for the explicit `KillSwitch`

viktorklang: @patriknw +1 on that it must be good ;)

drewhk: Not much apart from that I like the more domain specific class compared to a Future[Done]. One option I can do is to use the KillSwitch class but optionally take an external Future[Done].

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
drewhk: Because `cb.invoke` is just a fancy way of `!`

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
viktorklang: I'd want to avoid introducing too many concepts and classes (more API surface area and more to learn for the user),
what do you think about the: x.killSwitch(future: Future[Done]) proposal?

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
viktorklang: Given the comment just above that line, I am the one who should apologize.

drewhk: no worries :)

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
ktoso: true, `@see` is javadoc style

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/**
+ * A [[KillSwitch]] is a provider for [[Flow]]s that can be completed or failed from the outside. A [[Flow]] returned
+ * by the switch can be materialized arbitrary amount of times: every newly materialized [[Flow]] belongs to the switch
+ * from which it was aquired. Multiple [[KillSwitch]] instances are isolated from each other, shutting down or aborting
+ * on instance does not affect the [[Flow]]s provided by another instance.
+ *
+ * After calling [[KillSwitch#shutdown()]] all materialized, running instances of all [[Flow]]s provided by the
+ * [[KillSwitch]] will complete their downstreams and cancel their upstreams (unless if finished or failed already in which
```
ktoso: `unless finished`

ktoso:
Once rebased this can be used in place of the https://github.com/akka/akka/pull/19943#discussion-diff-55432549R113 in Http

drewhk:
Updated, now with unique killswitches added, Java API, more docs and tests.

drewhk:
hm, mima doesn't like me.

akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketIntegrationSpec.scala
```diff
@@ -69,12 +72,35 @@ class WebSocketIntegrationSpec extends AkkaSpec("akka.stream.materializer.debug.
...
+          override def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, Promise[Done]) = {
+            val promise = Promise[Done]
+
+            val logic = new GraphStageLogic(shape) with InHandler with OutHandler {
+              override def onPush(): Unit = push(shape.out, grab(shape.in))
+              override def onPull(): Unit = pull(shape.in)
+
+              override def preStart(): Unit = {
+                val cb = getAsyncCallback[Done](_ ⇒ complete(shape.out))
+                promise.future.onSuccess { case Done ⇒ cb.invoke(Done) }(akka.dispatch.ExecutionContexts.sameThreadExecutionContext)
```
viktorklang: use `foreach` iso onSuccess

akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketIntegrationSpec.scala
```diff
@@ -69,12 +72,35 @@ class WebSocketIntegrationSpec extends AkkaSpec("akka.stream.materializer.debug.
       val binding = Await.result(bindingFuture, 3.seconds)
       val myPort = binding.localAddress.getPort
 
+      val completeOnlySwitch: Flow[ByteString, ByteString, Promise[Done]] = Flow.fromGraph(
+        new GraphStageWithMaterializedValue[FlowShape[ByteString, ByteString], Promise[Done]] {
+          override val shape: FlowShape[ByteString, ByteString] = FlowShape(Inlet("in"), Outlet("out"))
```
viktorklang: Shouldn't those be: "KillSwitch.in" and "KillSwitch.out"?

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
```diff
@@ -0,0 +1,248 @@
...
+   */
+  def singleBidi[T1, T2]: BidiFlow[T1, T1, T2, T2, UniqueKillSwitch] =
+    uniqueKillSwitchBidi.asInstanceOf[BidiFlow[T1, T1, T2, T2, UniqueKillSwitch]]
+
+  private val uniqueKillSwitchFlow: Flow[Any, Any, UniqueKillSwitch] = Flow.fromGraph(UniqueKillSwitch)
+  private val uniqueKillSwitchBidi: BidiFlow[Any, Any, Any, Any, UniqueKillSwitch] = BidiFlow.fromGraph(BidiBreaker)
+
+  abstract class KillableGraphStageLogic(val terminationSignal: Future[Done], _shape: Shape) extends GraphStageLogic(_shape) {
+    override def preStart(): Unit = {
+      val cb = getAsyncCallback[Try[Done]](onSwitch)
```
viktorklang: why create this here if you are going to call completeStage/failStage on the next few lines?

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
```diff
@@ -0,0 +1,248 @@
...
+      terminationSignal.value match {
+        case Some(Success(Done)) ⇒ completeStage()
```
viktorklang: why not  `case Some(result) => onSwitch(result)`

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
```diff
@@ -0,0 +1,248 @@
...
+   * switch will be stopped normally or failed.
+   *
+   * @tparam T Type of the elements the Flow will forward
+   * @return   A reusable [[Flow]] that is linked with the switch. The materialized value provided is this switch itself.
+   */
+  def flow[T]: Flow[T, T, SharedKillSwitch] = _flow.asInstanceOf[Flow[T, T, SharedKillSwitch]]
+
+  override def toString: String = s"KillSwitch($name)"
+
+  private class SharedKillSwitchFlow extends GraphStageWithMaterializedValue[FlowShape[Any, Any], SharedKillSwitch] {
```
viktorklang: not final?

akka-stream/src/main/scala/akka/stream/javadsl/KillSwitch.scala
```diff
@@ -0,0 +1,144 @@
...
+   */
+  def shutdown(): Unit = wrapped.shutdown()
+
+  /**
+   * After calling [[UniqueKillSwitch#abort()]] the running instance of the [[Flow]] that materialized to the
+   * [[UniqueKillSwitch]] will fail its downstream with the provided exception and cancel its upstream
+   * (unless if finished or failed already in which case the command is ignored). Subsequent invocations of
+   * completion commands will be ignored.
+   */
+  def abort(reason: Throwable): Unit = wrapped.abort(reason)
```
viktorklang: What happens if abort is called after a call to shutdown and vice versa?

akka-stream/src/main/scala/akka/stream/javadsl/KillSwitch.scala
```diff
@@ -0,0 +1,144 @@
...
+ * (unless it finished or failed already in which case the command is ignored). Subsequent invocations of
+ * [[SharedKillSwitch#shutdown()]] and [[SharedKillSwitch#abort()]] will be ignored.
+ *
+ * The [[Flow]]s provided by the [[SharedKillSwitch]] do not modify the passed through elements in any way or affect
+ * backpressure in the stream. All provided [[Flow]]s provide the parent [[SharedKillSwitch]] as materialized value.
+ *
+ * This class is thread-safe, the instance can be passed safely among threads and its methods may be invoked concurrently.
+ *
+ */
+final class SharedKillSwitch private[stream] (val wrapped: scaladsl.SharedKillSwitch) {
```
viktorklang: should `wrapped` not be private?

akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketIntegrationSpec.scala
drewhk: this is just a small stage for this test, so I was lazy, but you are right, I will add names there

akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketIntegrationSpec.scala
drewhk: true

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
drewhk: true!

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
drewhk: that is also true

akka-stream/src/main/scala/akka/stream/scaladsl/KillSwitch.scala
drewhk: yep, true

akka-stream/src/main/scala/akka/stream/javadsl/KillSwitch.scala
drewhk: the stage infrastructure is idempotent in the close events, so thankfully nothing.

akka-stream/src/main/scala/akka/stream/javadsl/KillSwitch.scala
drewhk: yes, it should, or called toScala

drewhk:
thanks @viktorklang !

akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketIntegrationSpec.scala
viktorklang: i.e:

~~~scala
promise.future.foreach(getAsyncCallback[Done](_ ⇒ complete(shape.out)).invoke)(akka.dispatch.ExecutionContexts.sameThreadExecutionContext)
~~~

akka-http-core/src/test/scala/akka/http/impl/engine/ws/WebSocketIntegrationSpec.scala
```diff
@@ -85,7 +111,7 @@ class WebSocketIntegrationSpec extends AkkaSpec("akka.stream.materializer.debug.
         .request(10)
         .expectNoMsg(1500.millis)
 
-      breaker.value.get.get.complete()
+      breaker.trySuccess(Done)
```
viktorklang: can this return false?

viktorklang:
My only remaining concern is that this adds almost 700 LoC (including comments of course), wouldn't it be drastically less if all we needed to work with is Promise[Done] (which is already documented and has clear semantics for completion?

drewhk:
One of the reasons it is so long is because the duplication with the Java API which turns out to be not needed as the only specific thing is the returned Flow, but if I return Graph[FlowShape] then all these can be shared, no duplications (TBD). Then the only thing that makes this large is the docs. We discussed the plain Promise approach, but the team agreed that we don't like Promise that much as a users API and it is easier to document the methods of a KillSwitch than to refer to the methods of Promise (which has generic documentation).

viktorklang:
@drewhk what you explained sounds like a good compromise

akka-stream-tests/src/test/java/akka/stream/javadsl/KillSwitchTest.java
```diff
@@ -0,0 +1,101 @@
...
+    downstream.expectNext(1);
+
+    assertEquals(killSwitch, k);
+
+    killSwitch.shutdown();
+
+    upstream.expectCancellation();
+    downstream.expectComplete();
+
+    assertEquals(completionStage.toCompletableFuture().get(), Done.getInstance());
```
patriknw: use `get` with timeout parameters

drewhk: true

akka-stream/src/main/scala/akka/stream/KillSwitch.scala
```diff
@@ -0,0 +1,259 @@
...
+      }
+    }
+
+    private def onSwitch(mode: Try[Done]): Unit = mode match {
+      case Success(_)  ⇒ completeStage()
+      case Failure(ex) ⇒ failStage(ex)
+    }
+  }
+
+  private[stream] object UniqueKillSwitchStage extends GraphStageWithMaterializedValue[FlowShape[Any, Any], UniqueKillSwitch] {
```
patriknw: shouldn't this be a `val` instead of nested object if we care about performance, otherwise you could create a new instance each time and avoid that `asInstanceOf`

akka-stream/src/main/scala/akka/stream/KillSwitch.scala
```diff
@@ -0,0 +1,259 @@
...
+    override val initialAttributes = Attributes.name("breaker")
+    override val shape = FlowShape(Inlet[Any]("breaker.in"), Outlet[Any]("breaker.out"))
+    override def toString: String = "Breaker"
```
patriknw: Breaker vs KillSwitch?

patriknw:
mention in reference documentation?

patriknw:
LGTM (after possible doc addition)

akka-stream/src/main/scala/akka/stream/KillSwitch.scala
drewhk: What do you mean about performance here? There is no need for a new instance, a stage is already a factory.

patriknw: didn't we discover that nested objects involved some volatile (or something)?

I was mostly disliked the `asInstanceOf`, but it's not important

drewhk: > didn't we discover that nested objects involved some volatile (or something)?

That case was nested objects in a class which were used all the time (old interpreter). Probably a val is cheaper here but I don't think this is a hot path, especially since the stage is reusable.

akka-stream/src/main/scala/akka/stream/KillSwitch.scala
drewhk: yep, true

Issue #19844, opened 2016-02-22
Add canonical links to docs


ktoso:
Supposedly if we'd add links to `/current/` google would like us more, like so:

`<link rel="canonical" ..>`

PR #19903, opened 2016-02-28
+doc #19844 rel links to "current" on each page

ktoso:
A bit of sphinx hackery but made it work.

Appears as 

```
    <link rel="canonical" href="http://akka.io/docs/akka/current/general/terminology.html" />
```

Resolves https://github.com/akka/akka/issues/19844

### DIFF
--- a/akka-docs/_sphinx/themes/akka/layout.html
+++ b/akka-docs/_sphinx/themes/akka/layout.html
@@ -23,7 +23,9 @@
 {% block relbar2 %}{% endblock %}
 
 {% block extrahead %}
-  <!-- Akka release version-->
+    <!-- Hint to search engines that the "canonical" page is under "current", which will boost it appearing in search results -->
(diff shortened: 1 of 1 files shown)


ktoso:
As discussed on eng meeting, we abuse the rel a bit, but it'll give the results we want from google and others. Merging.

akka-docs/_sphinx/themes/akka/layout.html
```diff
@@ -23,7 +23,9 @@
 {% block relbar2 %}{% endblock %}
 
 {% block extrahead %}
-  <!-- Akka release version-->
+    <!-- Hint to search engines that the "canonical" page is under "current", which will boost it appearing in search results -->
+    <link rel="canonical" href="http://akka.io/docs/akka/current/{{ pagename }}.html" />
```
rkuhn: This should probably only be in the online version (i.e. in the `if include_analytics` block).

ktoso: Oh, that's the reason for it, yeah you're right then

PR #19905, opened 2016-02-28
Replace errorLogging with GraphStage - resolves #19899

### LINKED ISSUE #19899: Replace errorLogging with GraphStage
If you want to help out a bit there's a nice low hanging fruit in that area though:
The usage of transform+pushstage in: https://github.com/akka/akka/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala#L251
and https://github.com/akka/akka/blob/master/akka-http-core/src/main/scala/akka/http/impl/util/package.scala#L92 
should be rewritten to be a GraphStage

miciek:
Created `ErrorHandling` stage. @ktoso Review?

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
@@ -78,7 +78,7 @@ private[http] object OutgoingConnectionBlueprint {
       val terminationFanout = b.add(Broadcast[HttpResponse](2))
       val terminationMerge = b.add(TerminationMerge)
 
-      val logger = b.add(Flow[ByteString].transform(() ⇒ errorHandling((t: Throwable) ⇒ log.error(t, "Outgoing request stream error"))).named("errorLogger"))
+      val logger = b.add(ErrorHandling[ByteString]((t: Throwable) ⇒ log.error(t, "Outgoing request stream error")).named("errorLogger"))
       val wrapTls = b.add(Flow[ByteString].map(SendBytes))
       terminationMerge.out ~> requestRendering ~> logger ~> wrapTls
 
--- a/akka-http-core/src/main/scala/akka/http/impl/util/package.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/package.scala
@@ -5,6 +5,7 @@
 package akka.http.impl
 
 import akka.NotUsed
+import akka.stream.{ Attributes, Outlet, Inlet, FlowShape }
 
 import language.implicitConversions
 import java.nio.charset.Charset
@@ -88,15 +89,6 @@ package object util {
     }
   }
 
-  private[http] def errorHandling[T](handler: Throwable ⇒ Unit): PushStage[T, T] =
-    new PushStage[T, T] {
-      override def onPush(element: T, ctx: Context[T]): SyncDirective = ctx.push(element)
-      override def onUpstreamFailure(cause: Throwable, ctx: Context[T]): TerminationDirective = {
-        handler(cause)
-        super.onUpstreamFailure(cause, ctx)
-      }
-    }
-
   private[http] def humanReadableByteCount(bytes: Long, si: Boolean): String = {
     val unit = if (si) 1000 else 1024
     if (bytes >= unit) {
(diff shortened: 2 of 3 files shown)


akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -78,7 +78,7 @@ private[http] object OutgoingConnectionBlueprint {
       val terminationFanout = b.add(Broadcast[HttpResponse](2))
       val terminationMerge = b.add(TerminationMerge)
 
-      val logger = b.add(Flow[ByteString].transform(() ⇒ errorHandling((t: Throwable) ⇒ log.error(t, "Outgoing request stream error"))).named("errorLogger"))
+      val logger = b.add(ErrorHandling[ByteString]((t: Throwable) ⇒ log.error(t, "Outgoing request stream error")).named("errorLogger"))
```
miciek: Do we need `named` here?

ktoso:
OK TO TEST

ktoso:
Thanks for the PR, will review soon!

ktoso:
Resolves https://github.com/akka/akka/issues/19899#issuecomment-189853503

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
johanandren: I think it might be nice for debugging purposes

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
```diff
@@ -113,6 +105,28 @@ package util {
...
+      setHandler(in, new InHandler {
+        override def onPush(): Unit = push(out, grab(in))
+
+        override def onUpstreamFailure(ex: Throwable): Unit = {
+          handler(ex)
+          super.onUpstreamFailure(ex)
+        }
+      })
+
+      setHandler(out, new OutHandler {
```
johanandren: Both the `InHandler` and `OutHandler` can be implemented by the anonymous `GraphStageLogic` and then just `setHanders(in, out, this)` to save two allocations on each use (we do that optimization in other graphstages)

johanandren:
LGTM (but the optimization should probably be done, we want the HTTP blueprints as lean as possible)

miciek:
@johanandren Are you talking about something like this?

```
    override def createLogic(attr: Attributes) =
      new GraphStageLogic(shape) with InHandler with OutHandler {
        override def onPush(): Unit = push(out, grab(in))

        override def onUpstreamFailure(ex: Throwable): Unit = {
          handler(ex)
          super.onUpstreamFailure(ex)
        }

        setHandler(in, this)
        setHandler(out, this)
      }
```

I see your point, but don't like it so much comparing to the first approach. But if you guys want, I can commit this optimised version. Let me know

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
```diff
@@ -113,6 +105,27 @@ package util {
   import akka.stream.{ Attributes, Outlet, Inlet, FlowShape }
   import scala.concurrent.duration.FiniteDuration
 
+  private[http] final case class ErrorHandling[T](handler: Throwable ⇒ Unit) extends GraphStage[FlowShape[T, T]] {
```
ktoso: You could extend `SimpleLinearGraphStage` (then remove `in`/`out`/`shape` :-))

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
```diff
@@ -113,6 +105,27 @@ package util {
...
+    private val in = Inlet[T]("ErrorHandling.in")
+    private val out = Outlet[T]("ErrorHandling.out")
+
+    override val shape = FlowShape(in, out)
+
+    override def createLogic(attr: Attributes) =
+      new GraphStageLogic(shape) with InHandler with OutHandler {
```
ktoso: Good, we want the http pipeline to be as lean on allocations as possible, so :+1: on that trick.

ktoso:
LGTM, https://github.com/akka/akka/pull/19905#discussion_r54380142 could be applied though

johanandren:
LGTM

akka-http-core/src/main/scala/akka/http/impl/util/package.scala
miciek: Done :)

johanandren:
Thanks @miciek !

Issue #18804, opened 2015-10-28
Docs: strange example for TestKit in Akka documentation


amorfis:
In docs about testing (http://doc.akka.io/docs/akka/snapshot/scala/testing.html#Asynchronous_Integration_Testing_with_TestKit), there is this example of code:

```scala
class MySpec(_system: ActorSystem) extends TestKit(_system) with ImplicitSender
  with WordSpecLike with Matchers with BeforeAndAfterAll {
 
  def this() = this(ActorSystem("MySpec"))
```
Why is there auxillary parameterless constructor? Couldn't it be simplified to this:

```scala
class MySpec() extends TestKit(ActorSystem("MySpec")) with ImplicitSender
  with WordSpecLike with Matchers with BeforeAndAfterAll {
```

Is there any reason for that auxillary constructor?

rkuhn:
Hmm, I can’t see a good reason either; would you mind trying out and proposing a change?

amorfis:
Do you mean pull request? I can change the tests in the code here on github, but I can't find the docs text. I think it's not here in the akka repository?

2beaucoup:
@amorfis it's here: https://github.com/akka/akka/blob/master/akka-docs/rst/scala/testing.rst#asynchronous-integration-testing-with-classtestkit

PR #19906, opened 2016-02-28
simplifyTestKit doc async sample #18804

richardimaoka:
I tested this locally in my desktop, but assume the change will be auto-tested too?
Thanks

### DIFF
--- a/akka-docs/rst/scala/code/docs/testkit/PlainWordSpec.scala
+++ b/akka-docs/rst/scala/code/docs/testkit/PlainWordSpec.scala
@@ -13,11 +13,9 @@ import org.scalatest.Matchers
 import org.scalatest.BeforeAndAfterAll
 
 //#implicit-sender
-class MySpec(_system: ActorSystem) extends TestKit(_system) with ImplicitSender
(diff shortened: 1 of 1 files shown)


johanandren:
OK TO TEST

akka-docs/rst/scala/code/docs/testkit/PlainWordSpec.scala
```diff
@@ -13,12 +13,10 @@ import org.scalatest.Matchers
 import org.scalatest.BeforeAndAfterAll
 
 //#implicit-sender
-class MySpec(_system: ActorSystem) extends TestKit(_system) with ImplicitSender
+class MySpec() extends TestKit(ActorSystem("MySpec")) with ImplicitSender
```
johanandren: Seems reasonable to me, as the sample does not show how to create an abstract base class for your tests, but an actual concrete test class

johanandren:
Fixes #18804

johanandren:
LGTM

ktoso:
LGTM, the `def this` has confused newcomers in the past indeed. 
Thanks for the contrib!

Issue #19723, opened 2016-02-09
Should streaming IO reword the "once we can use Java 8" section?


skestle:
I've just found the [file streaming API](http://doc.akka.io/docs/akka-stream-and-http-experimental/2.0.2/scala/stream-io.html#Streaming_File_IO) but states:

> Once Akka is free to require JDK8 (from 2.4.x) these implementations will be updated to make use of the new NIO APIs (i.e. AsynchronousFileChannel).

Akka is at 2.4 and on Java 8; this much should be clarified; does the streaming api use `AsynchronousFileChannel`s? (I presume not, otherwise that would be 2.1+ right?)

ktoso:
It's not yet using `AsynchronousFileChannel`

ktoso:
This was solved in #19907

Issue #17269, opened 2015-04-22
Provide JDK7 AsynchronousFileChannel backed FileSource/Sink


ktoso:
As follow up to https://github.com/akka/akka/pull/17211#issuecomment-95279417 and [#15588 - streams: Provide basic streams for files and from InputStream etc.](https://github.com/akka/akka/issues/15588).

We will want to re-visit SynchronousFileSource and SynchronousFileSink and possibly reimplement using AsynchronousFileChannel (it won't necessarily be faster).

We also want to re-visit `SynchronousFileSource`, about:
- can we reimplement it as a stage (yes we can, but:) and it will be fused (needs fusing API exposed nicer to user land, so that users can put fuse blocks etc)
- do we need to read-ahead? if yes can we read-ahead better? (see here for a [discussion on read-ahead](https://github.com/akka/akka/pull/17211#discussion_r28886575))
- compare perf with asyncchanel based (I remember the async one getting to almost the same perf as synchronous)

zaharidichev:
@ktoso I would like to give that one a shot if its free to grab. I wonder however, if benchmarking does not show any significant improvement, what is the rationale for moving to asynchronous channel ?

zaharidichev:
@ktoso I would like to give that one a shot if its free to grab. I wonder however, if benchmarking does not show any significant improvement, what is the rationale for moving to asynchronous channel ?

johanandren:
For reference: PR exploring this #21801

ktoso:
Here it's most important to think and confirm if there are gains from the rewrite or not.
The blocking one will likely have (as found in the other PR) better raw throughput, but then we end up putting it on the IO dispatcher. The async one could remain on the default dispatcher perhaps (?). So another question is about the scalability - what happens if we write to 1000 files at the same time, which impl does better then?

ktoso:
This is being worked on by @agolubev

agolubev:
There is almost a decision to close this PR and stay with sync approach https://github.com/akka/akka/pull/21801 as there is similar performance (sync and async) but async is limited with regards to thread pool configuration (which will be external to Akka).
Ref: https://github.com/akka/akka/pull/23217#issuecomment-323351104

ktoso:
AFAIR we should close this as we went with https://github.com/akka/akka/pull/21801 after tons of benchmarking

PR #19907, opened 2016-02-28
=doc Remove outdated note in Streaming File IO

### LINKED ISSUE #19904: Outdated note in Streaming File IO
It has a note that talks about "we're still on 2.3.x" etc.

http://doc.akka.io/docs/akka/2.4.2/scala/stream/stream-io.html#Streaming_File_IO

Remove / adjust the note.

naferx:
Fixes #19904

### DIFF
--- a/akka-docs/rst/java/stream/stream-io.rst
+++ b/akka-docs/rst/java/stream/stream-io.rst
@@ -99,11 +99,6 @@ Streaming File IO
 Akka Streams provide simple Sources and Sinks that can work with :class:`ByteString` instances to perform IO operations
 on files.
 
-.. note::
-  Since the current version of Akka (``2.3.x``) needs to support JDK6, the currently provided File IO implementations
-  are not able to utilise Asynchronous File IO operations, as these were introduced in JDK7 (and newer).
-  Once Akka is free to require JDK8 (from ``2.4.x``) these implementations will be updated to make use of the
-  new NIO APIs (i.e. :class:`AsynchronousFileChannel`).
 
 Streaming data from a file is as easy as creating a `FileIO.fromFile` given a target file, and an optional
 ``chunkSize`` which determines the buffer size determined as one "element" in such stream:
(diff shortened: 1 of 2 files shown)


akka-docs/rst/java/stream/stream-io.rst
```diff
@@ -100,7 +100,7 @@ Akka Streams provide simple Sources and Sinks that can work with :class:`ByteStr
 on files.
 
 .. note::
-  Since the current version of Akka (``2.3.x``) needs to support JDK6, the currently provided File IO implementations
+  Since the version of Akka (``2.3.x``) needs to support JDK6, the currently provided File IO implementations
   are not able to utilise Asynchronous File IO operations, as these were introduced in JDK7 (and newer).
   Once Akka is free to require JDK8 (from ``2.4.x``) these implementations will be updated to make use of the
```
johanandren: The entire note needs updating, as we are now on 2.4.x and JDK8. We just haven't had time to move to NIO. Maybe we should just remove it.

ktoso: +1 to removing the entire note

naferx:
@johanandren @ktoso Updated the PR according to your suggestions!.

ktoso:
Thanks a lot @naferx!
LGTM

ktoso:
Resolves https://github.com/akka/akka/issues/19942
Resolves #19904

julianvmodesto:
Does the streaming FileIO implementation use NIO's AsynchronousFileChannel?

ktoso:
No it doesn't and I don't think you'd notice as we dedicate a preconfigured threadpool for the blocking ops
In my benchmarks only small performance gains could be won with the Async channel, though I'd definitely welcome a rewrite of the Publisher/Subscriber based FileIO onto a GraphStage with AsyncFileChannel - callback handling is so much simpler now that we have `GraphStage` (we did not have it before).

In summary:
- it doesn't, I don't think you'll notice though - dedicated pools for the blocking ops now hide most of the difference
- it could be rewritten, we currently do not have the capacity to hack on it, if you'd like to help out that'd be awesome :-)

julianvmodesto:
Sorry I deleted my comment ("Does the File IO implementation use NIO's AsynchronousFileChannel?") – deleted because I saw #19723 and #17269.

Appreciate the reply! I'm entirely new to Akka hehe. GraphStage sounds interesting, I'll take a look at it.

PR #19908, opened 2016-02-28
Replaced PushStage based Take with GraphStage #19834

trautonen:
Replaced `PushPullStage` based `Take` with a `GraphStage` implementation using `SimpleLinearGraphStage` as a base.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -124,19 +124,30 @@ private[akka] final case class Recover[T](pf: PartialFunction[Throwable, T]) ext
 /**
  * INTERNAL API
  */
-private[akka] final case class Take[T](count: Long) extends PushPullStage[T, T] {
-  private var left: Long = count
+private[akka] final case class Take[T](count: Long) extends SimpleLinearGraphStage[T] {
+  override def initialAttributes: Attributes = DefaultAttributes.take
 
-  override def onPush(elem: T, ctx: Context[T]): SyncDirective = {
-    left -= 1
-    if (left > 0) ctx.push(elem)
-    else if (left == 0) ctx.pushAndFinish(elem)
-    else ctx.finish() //Handle negative take counts
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) with InHandler with OutHandler {
+    private var left: Long = count
+
+    override def onPush(): Unit = {
+      val leftBefore = left
+      if (leftBefore >= 1) {
+        left = leftBefore - 1
+        push(out, grab(in))
+      }
+      if (leftBefore <= 1) completeStage()
+    }
+
+    override def onPull(): Unit = {
+      if (left > 0) pull(in)
+      else completeStage()
+    }
+
+    setHandlers(in, out, this)
   }
 
-  override def onPull(ctx: Context[T]): SyncDirective =
-    if (left <= 0) ctx.finish()
-    else ctx.pull()
+  override def toString: String = "Take"
 }
 
 /**
--- a/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/impl/fusing/InterpreterSpec.scala
@@ -14,10 +14,13 @@ class InterpreterSpec extends AkkaSpec with GraphInterpreterSpecKit {
   import Supervision.stoppingDecider
 
   /*
-   * These tests were writtern for the previous veryion of the interpreter, the so called OneBoundedInterpreter.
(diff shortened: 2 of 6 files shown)


akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -124,19 +124,33 @@ private[akka] final case class Recover[T](pf: PartialFunction[Throwable, T]) ext
...
+      if (left > 0) {
+        push(out, grab(in))
+      } else if (left == 0) {
+        push(out, grab(in))
+        completeStage()
+      } else {
+        completeStage()
+      }
+    }
+
```
trautonen: _onPush()_ can also be implemented as:

```scala
      left -= 1
      if (left >= 0) push(out, grab(in))
      if (left <= 0) completeStage()
    }
```

Not sure which one is cleaner and is there any performance difference.

ktoso:
OK TO TEST

johanandren:
LGTM

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -124,19 +124,33 @@ private[akka] final case class Recover[T](pf: PartialFunction[Throwable, T]) ext
...
+    override def onPull(): Unit = {
+      if (left <= 0) completeStage()
```
viktorklang: Always put the `mostly-true` case in the first branch :)

~~~scala
if (left > 0) pull(in) else completeStage()
~~~

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -124,19 +124,33 @@ private[akka] final case class Recover[T](pf: PartialFunction[Throwable, T]) ext
 /**
  * INTERNAL API
  */
-private[akka] final case class Take[T](count: Long) extends PushPullStage[T, T] {
-  private var left: Long = count
+private[akka] final case class Take[T](count: Long) extends SimpleLinearGraphStage[T] {
+  override def initialAttributes: Attributes = DefaultAttributes.take
```
viktorklang: :+1:

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
viktorklang: @trautonen I'd probably do something like this:

~~~scala
      val leftBefore = left
      left = leftBefore - 1
      if (leftBefore >= 1) push(out, grab(in))
      if (leftBefore <= 1) completeStage()
    }
~~~

Why? Because I prefer code to be explicit that it doesn't depend on other logic manipulating the `var` somewhere in the callchain of the method. :)

trautonen:
Jenkins job https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/5514/ failed, but not sure if the failure is related to the code changes int his pull request at all? As far as I understand, the job fails due to binary incompatibility, but this pull request does not event change any public APIs.

johanandren:
Fail is binary compatibility check (MiMa), you can run it locally using `mimaReportBinaryIssues`. The class is an internal API so just whitelisting the error in `project/MiMa.scala` is safe.

johanandren:
(When you run it locally it will tell you the statement to whitelist the error)

drewhk:
LGTM

ktoso:
Please add these filters to `MiMa.scala` as @johanandren mentioned, otherwise we can't merge:

```
[error]  * object akka.stream.impl.Stages#Take does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("akka.stream.impl.Stages$Take$")
[error]  * class akka.stream.impl.Stages#Take does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("akka.stream.impl.Stages$Take")
[error]  * the type hierarchy of class akka.stream.impl.fusing.Take is different in current version. Missing types {akka.stream.stage.AbstractStage,akka.stream.stage.PushPullStage}
[error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("akka.stream.impl.fusing.Take")
[error]  * method onPush(java.lang.Object,akka.stream.stage.Context)akka.stream.stage.SyncDirective in class akka.stream.impl.fusing.Take does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.impl.fusing.Take.onPush")
[error]  * method onPull(akka.stream.stage.Context)akka.stream.stage.SyncDirective in class akka.stream.impl.fusing.Take does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.impl.fusing.Take.onPull")
```

trautonen:
Added MiMa filters with a comment to this pull request.

ktoso:
onPush could be microbenchmarked the hell out of it I think but I think this is good to merge already :-)
Thanks a lot, LGTM!

PR #19912, opened 2016-02-29
#19824 Unhandled akka.actor.Status.Failure in RemoteActorRefProvider in state WaitTransportShutdown

### LINKED ISSUE #19824: Unhandled akka.actor.Status.Failure in RemoteActorRefProvider in state WaitTransportShutdown
When RARP is being shutdown in state WaitDaemonShutdown it pipes call result of transport shutdown to self [here](https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala#L57). `pipeTo` transforms `scala.util.Failure` into `akka.actor.Status.Failure` [here](https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/pattern/PipeToSupport.scala#L20)

Which could still be ok, but [here](https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/Remoting.scala#L159) map is used on Future, which is applied only to **Success**, so when we get a failure it has following flow:
[ask] => scala.util.Failure => [pipeTo] => akka.actor.Status.Failure => unhandled message [here](https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala#L61)

Filed match for me under heavy load with VisualVM sampling and profiling at the same time brought me
> [WARN] [02/18/2016 09:36:04.442] [lp-akka-cluster-local-test-remote-dispatcher-29] [akka.tcp://lp-akka-cluster-local-test@XXX:2549/system/remoting-terminator] unhandled event Failure(akka.pattern.AskTimeoutException: Ask timed out on [Actor[akka://lp-akka-cluster-local-test/system/endpointManager#-2071163190]] after [10000 ms]. Sender[null] sent message of type "akka.remote.EndpointManager$ShutdownAndFlush$".) in state WaitTransportShutdown

There are 3 possible ways to fix it:
1. simply replace `map` with `andThen` - simple and fast, currently failure is not propagated anyway
2. add match for `akka.actor.Status.Failure` in RARP
3. use `akka.actor.Status.Success` (some simple case object can fit) and `akka.actor.Status.Failure` as messages for communication - that way failure can be propagated (I'm not sure, if that's needed anyway) and it looks cleaner for me than using `()` as message, option 2 is required for that to work, of course.

412b:
Fixes #19824 

When RARP is being shutdown, it pipes result of the shutdown of transport to self.

remote's shutdown is using `ask` pattern, so it can produce `Status.Failure`, which is not handled in RARP's state `WaitTransportShutdown`.
For fixing it added matching for `Status.Failure` and changed `RemoteTransport`'s shutdown signature to use `akka.Done`, which looks more consistent with other shutdown's and `akka.Done` is more verbose than previously used `Unit`.

### DIFF
--- a/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala
+++ b/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala
@@ -4,6 +4,7 @@
 
 package akka.remote
 
+import akka.Done
 import akka.actor._
 import akka.dispatch.sysmsg._
 import akka.event.{ Logging, LoggingAdapter, EventStream }
@@ -59,8 +60,13 @@ private[akka] object RemoteActorRefProvider {
     }
 
     when(WaitTransportShutdown) {
-      case Event((), _) ⇒
+      case Event(Done, _) ⇒
         log.info("Remoting shut down.")
+        systemGuardian ! TerminationHookDone
+        stop()
+
+      case Event(Status.Failure(ex), _) ⇒
+        log.error(ex, "Remoting shut down with error")
         systemGuardian ! TerminationHookDone
         stop()
     }
--- a/akka-remote/src/main/scala/akka/remote/RemoteTransport.scala
+++ b/akka-remote/src/main/scala/akka/remote/RemoteTransport.scala
@@ -5,6 +5,7 @@
 package akka.remote
 
 import akka.AkkaException
+import akka.Done
 import akka.actor._
 import akka.event.{ LoggingAdapter }
 import scala.collection.immutable
@@ -39,7 +40,7 @@ private[akka] abstract class RemoteTransport(val system: ExtendedActorSystem, va
   /**
    * Shuts down the remoting
    */
-  def shutdown(): Future[Unit]
+  def shutdown(): Future[Done]
 
   /**
    * Address to be used in RootActorPath of refs generated for this transport.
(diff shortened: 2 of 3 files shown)


akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala
```diff
@@ -59,10 +60,15 @@ private[akka] object RemoteActorRefProvider {
...
 
     when(WaitTransportShutdown) {
-      case Event((), _) ⇒
+      case Event(Done, _) ⇒
         log.info("Remoting shut down.")
         systemGuardian ! TerminationHookDone
         stop()
+
+      case Event(Status.Failure(ex), _) ⇒
+        log.error(ex, "Remoting shut down with error")
```
patriknw: would it be interesting to include the `ex.getMessage`?

patriknw: when will this happen? is it ask timeout?
I'm thinking if the log level should be error, warning or info.

412b: @patriknw in general it's not limited to `AskTimeoutException`, it can be as well `IllegalArgumentException` (from `ask`) or any other, which got piped.
error was used as initil variant as `Remoting` uses `notifyError` and publishes `RemoteErrorEvent`, when shutdown goes wrong.

patriknw: ok, as long as it's not something that happens during "rather normal" shutdown I think log.error is alright

patriknw:
OK TO TEST

patriknw:
LGTM

PR #19915, opened 2016-02-29
=htc #19058 prevent errors to be logged for http timeouts

### LINKED ISSUE #19058: Make Idle Timeout log DEBUG instead of exception in HTTP

jrudolph:
Fixes #19058, follow up to #19130.

/cc @danielwegener

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -15,7 +15,7 @@ import akka.http.impl.engine.client._
 import akka.http.impl.engine.server._
 import akka.http.impl.engine.ws.WebSocketClientBlueprint
 import akka.http.impl.settings.{ ConnectionPoolSetup, HostConnectionPoolSetup }
-import akka.http.impl.util.StreamUtils
+import akka.http.impl.util.{ MapError, StreamUtils }
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers.Host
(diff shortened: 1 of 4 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -85,7 +85,7 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
     connections.map {
       case Tcp.IncomingConnection(localAddress, remoteAddress, flow) ⇒
         val layer = serverLayer(settings, Some(remoteAddress), log)
-        val flowWithTimeoutRecovered = flow.recover { case t: TimeoutException ⇒ throw new HttpConnectionTimeoutException(t.getMessage) }
+        val flowWithTimeoutRecovered = flow.via(MapError { case t: TimeoutException ⇒ new HttpConnectionTimeoutException(t.getMessage) })
```
ktoso: Ah! Thanks to @drewhk I see why that's the fix. So if we throw in a stage we log, and if we invoke failure we do not log automagically, and then the proper debug/error logging applies.

Thanks a lot @jrudolph for the fix!

drewhk:
LGTM

ktoso:
LGTM, thanks!

ktoso:
We could consider adding a `mapError` operator actually, WDYT @drewhk @rkuhn ?

danielwegener:
@jrudolph TIL! cheers!

Antwnis:
Cheers! Looking forward for a 2.4.3 release

Issue #19926, opened 2016-03-01
Error in fixed-pool-size example


RayRoestenburg:
fixed-pool-size does not configure the size of the pool, but is used to switch a fixed pool size on or off: http://doc.akka.io/docs/akka/snapshot/scala/dispatchers.html#More_dispatcher_configuration_examples

From reference.conf:

        # Define a fixed thread pool size with this property. The corePoolSize
        # and the maximumPoolSize of the ThreadPoolExecutor will be set to this
        # value, if it is defined. Then the other pool-size properties will not
        # be used.
        fixed-pool-size = off

ktoso:
It does, see more details in https://github.com/akka/akka/pull/19927

ktoso:
Thanks for reporting in any case, could we improve wording somehow?

PR #19927, opened 2016-03-01
#19926 Fix for fixed-pool-size example.

RayRoestenburg:


ktoso:
This is not correct. That setting is something new in 2.4, see logic:

```
if (config.getString("fixed-pool-size") == "off")
      builder
        .setCorePoolSizeFromFactor(config getInt "core-pool-size-min", config getDouble "core-pool-size-factor", config getInt "core-pool-size-max")
        .setMaxPoolSizeFromFactor(config getInt "max-pool-size-min", config getDouble "max-pool-size-factor", config getInt "max-pool-size-max")
    else
      builder.setFixedPoolSize(config.getInt("fixed-pool-size"))
```

ktoso:
Thanks for the PR in any case, but it does not reflect state of HEAD :-)

RayRoestenburg:
@ktoso Ahhh ok, sorry about that, It was on an older version and mixed up the docs (I thought this was there already for a long time).

Issue #17570, opened 2015-05-26
upgrade scalariform


fommil:
your version of sbt-scalariform is old and anyone using a more recent version in their user config will trump it (and reformat a lot of files). I recommend changing https://github.com/akka/akka/blob/master/project/plugins.sbt#L14 to

```
addSbtPlugin("org.scalariform" %% "sbt-scalariform" % "1.4.0")
```

(note change of groupId)

I would have sent a PR but I can't do a clean `sbt compile test:compile it:compile` etc owing to missing snapshots that you depend on.

patriknw:
sounds good

missing snapshots? please elaborate

fommil:
```
sbt.ResolveException: unresolved dependency: com.typesafe.akka#akka-persistence-experimental_2.10;2.4-SNAPSHOT: not found
```

and many more...

patriknw:
ah, that is from the sample projects
yeah, we are breaking out those from the build
until then you have to first do a publishLocal (hm is there a cycle here?)

2m:
Alternatively you can leave out sample builds from the aggregate build by starting sbt like so:
```
sbt -Dakka.build.aggregateSamples=false
```

patriknw:
I don't think it was the samples. At least not for master. _2.10 is not something we use in master (unless it has been forgotten somewhere) and I tried a clean build (removed ivy cache) without problems.

PR #19936, opened 2016-03-01
=pro #17570 upgrade scalariform plugin

jrudolph:
I tried adapting the settings a bit to keep the old style as much as possible with these changes:

``` diff
-      .setPreference(AlignParameters, true)
+      .setPreference(AlignParameters, false)
+      .setPreference(SpacesAroundMultiImports, true)
+      .setPreference(DanglingCloseParenthesis, Prevent)
```

which mostly keeps the existing formatting (aside from those in these 300 files...). Some changes are only due to the fact the old scalariform was broken and didn't format lots of files for quite some time it seems.

On the other hand, it may make sense to try to use more of the default settings right now and do the hard and breaking change right now). Especially, `AlignParameters` actually looks quite nice for case classes if enabled (although it also generally increases conflict risk similar to alignment of case-lines in patterns).


jrudolph:
It seems there's still a bug in scalariform wrt parsing `val ->` in FSM.scala. See https://github.com/daniel-trinh/scalariform/issues/113.

jrudolph:
There's https://github.com/scala-ide/scalariform/issues/202 which, when fixed, would prevent a lot of those changes needed here.

The test failure is also a genuine problem. Pattern matches using FSM's `->` are rewritten to use `→` which isn't defined.

ktoso:
Status update: We'll focus on merging pending stuff and do the plugin upgrade soon

jrudolph:
It may make sense to wait for a while. It may be that the next version of sbt-scalariform fill fix most of the otherwise required changes.

ktoso:
We do want to do this soon however for now I'll close the PR (need to keep their number a bit more managable, 30+ is hard to look at :-))

project/plugins.sbt
```diff
@@ -11,7 +11,8 @@ resolvers += "Bintray Jcenter" at "https://jcenter.bintray.com/"
 addSbtPlugin("com.typesafe.sbt" % "sbt-multi-jvm" % "0.3.8")
 //#sbt-multi-jvm
 
-addSbtPlugin("com.typesafe.sbt" % "sbt-scalariform" % "1.2.0")
+addSbtPlugin("org.scalariform" % "sbt-scalariform" % "1.6.0")
+libraryDependencies += "org.scalariform" %% "scalariform" % "0.2.0-SNAPSHOT"
```
ktoso: not sure if we want to be on snapshot :~

PR #19943, opened 2016-03-03
=htc #15799 implement client-side `Expect: 100-continue` support

### LINKED ISSUE #15799: Client: Implement `Expect / 100-Continue` support

sirthias:
Finally closes #15799 for good.
Sorry for the long delay in getting this done.

This also includes a fix to an oversight that somehow wasn't discovered so far.
The spec requires a response parser to accept and ignore interim 1xx informational responses, which akka-http (and spray) has not been doing up to now.

/cc @jrudolph @2beaucoup

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
@@ -6,8 +6,10 @@ package akka.http.impl.engine.client
 
 import akka.NotUsed
 import akka.http.scaladsl.settings.{ ClientConnectionSettings, ParserSettings }
+import akka.stream.impl.ConstantFun
 import language.existentials
 import scala.annotation.tailrec
+import scala.concurrent.Promise
 import scala.collection.mutable.ListBuffer
 import akka.stream.TLSProtocol._
 import akka.util.ByteString
@@ -15,8 +17,8 @@ import akka.event.LoggingAdapter
 import akka.stream._
 import akka.stream.scaladsl._
 import akka.http.scaladsl.Http
-import akka.http.scaladsl.model.headers.Host
-import akka.http.scaladsl.model.{ IllegalResponseException, HttpMethod, HttpRequest, HttpResponse, ResponseEntity }
+import akka.http.scaladsl.model.headers
+import akka.http.scaladsl.model.{ IllegalResponseException, HttpRequest, HttpResponse, ResponseEntity }
 import akka.http.impl.engine.rendering.{ RequestRenderingContext, HttpRequestRendererFactory }
 import akka.http.impl.engine.parsing._
 import akka.http.impl.util._
@@ -28,6 +30,9 @@ import akka.stream.stage.{ InHandler, OutHandler }
  * INTERNAL API
  */
 private[http] object OutgoingConnectionBlueprint {
+
+  type BypassData = HttpResponseParser.ResponseContext
+
   /*
     Stream Setup
     ============
@@ -38,62 +43,75 @@ private[http] object OutgoingConnectionBlueprint {
                  +-------------------------------------->|  Merge   |                       |
                  | Termination Backchannel          |    +----------+                       |  TCP-
                  |                                  |                                       |  level
-                 |                                  | Method                                |  client
-                 |                +------------+    | Bypass                                |  flow
+                 |                                  | BypassData                            |  client
+                 |                +------------+    |                                       |  flow
     responseOut  |  responsePrep  |  Response  |<---+                                       |
     <------------+----------------|  Parsing   |                                            |
                                   |  Merge     |<------------------------------------------ V
                                   +------------+
   */
-  def apply(hostHeader: Host,
+  def apply(hostHeader: headers.Host,
             settings: ClientConnectionSettings,
             log: LoggingAdapter): Http.ClientLayer = {
     import settings._
 
-    // the initial header parser we initially use for every connection,
-    // will not be mutated, all "shared copy" parsers copy on first-write into the header cache
-    val rootParser = new HttpResponseParser(parserSettings, HttpHeaderParser(parserSettings) { info ⇒
-      if (parserSettings.illegalHeaderWarnings)
-        logParsingError(info withSummaryPrepended "Illegal response header", log, parserSettings.errorLoggingVerbosity)
-    })
-
-    val requestRendererFactory = new HttpRequestRendererFactory(userAgentHeader, requestHeaderSizeHint, log)
-
-    val requestRendering: Flow[HttpRequest, ByteString, NotUsed] = Flow[HttpRequest]
-      .map(RequestRenderingContext(_, hostHeader))
-      .via(Flow[RequestRenderingContext].flatMapConcat(requestRendererFactory.renderToSource).named("renderer"))
-
-    val methodBypass = Flow[HttpRequest].map(_.method)
-
-    import ParserOutput._
-    val responsePrep = Flow[List[ResponseOutput]]
-      .mapConcat(conforms)
-      .via(new PrepareResponse(parserSettings))
-
     val core = BidiFlow.fromGraph(GraphDSL.create() { implicit b ⇒
       import GraphDSL.Implicits._
-      val methodBypassFanout = b.add(Broadcast[HttpRequest](2, eagerCancel = true))
-      val responseParsingMerge = b.add(new ResponseParsingMerge(rootParser))
+
+      val renderingContextCreation = b.add {
+        Flow[HttpRequest] map { request ⇒
+          val sendEntityTrigger =
+            request.headers collectFirst { case headers.Expect.`100-continue` ⇒ Promise[NotUsed]().future }
+          RequestRenderingContext(request, hostHeader, sendEntityTrigger)
+        }
+      }
+
+      val bypassFanout = b.add(Broadcast[RequestRenderingContext](2, eagerCancel = true))
+
+      val terminationMerge = b.add(TerminationMerge)
+
+      val requestRendering: Flow[RequestRenderingContext, ByteString, NotUsed] = {
+        val requestRendererFactory = new HttpRequestRendererFactory(userAgentHeader, requestHeaderSizeHint, log)
+        Flow[RequestRenderingContext].flatMapConcat(requestRendererFactory.renderToSource).named("renderer")
+      }
+
+      val bypass = Flow[RequestRenderingContext] map { ctx ⇒
+        HttpResponseParser.ResponseContext(ctx.request.method, ctx.sendEntityTrigger.map(_.asInstanceOf[Promise[Unit]]))
+      }
+
+      val responseParsingMerge = b.add {
+        // the initial header parser we initially use for every connection,
+        // will not be mutated, all "shared copy" parsers copy on first-write into the header cache
+        val rootParser = new HttpResponseParser(parserSettings, HttpHeaderParser(parserSettings) { info ⇒
+          if (parserSettings.illegalHeaderWarnings)
+            logParsingError(info withSummaryPrepended "Illegal response header", log, parserSettings.errorLoggingVerbosity)
+        })
+        new ResponseParsingMerge(rootParser)
+      }
+
+      val responsePrep = Flow[List[ParserOutput.ResponseOutput]]
+        .mapConcat(ConstantFun.scalaIdentityFunction)
+        .via(new PrepareResponse(parserSettings))
 
       val terminationFanout = b.add(Broadcast[HttpResponse](2))
-      val terminationMerge = b.add(TerminationMerge)
 
       val logger = b.add(MapError[ByteString] { case t ⇒ log.error(t, "Outgoing request stream error"); t }.named("errorLogger"))
       val wrapTls = b.add(Flow[ByteString].map(SendBytes))
-      terminationMerge.out ~> requestRendering ~> logger ~> wrapTls
 
       val collectSessionBytes = b.add(Flow[SslTlsInbound].collect { case s: SessionBytes ⇒ s })
+
+      renderingContextCreation.out ~> bypassFanout.in
+      bypassFanout.out(0) ~> terminationMerge.in0
+      terminationMerge.out ~> requestRendering ~> logger ~> wrapTls
+
+      bypassFanout.out(1) ~> bypass ~> responseParsingMerge.in1
       collectSessionBytes ~> responseParsingMerge.in0
-
-      methodBypassFanout.out(0) ~> terminationMerge.in0
-
-      methodBypassFanout.out(1) ~> methodBypass ~> responseParsingMerge.in1
 
       responseParsingMerge.out ~> responsePrep ~> terminationFanout.in
       terminationFanout.out(0) ~> terminationMerge.in1
 
       BidiShape(
-        methodBypassFanout.in,
+        renderingContextCreation.in,
         wrapTls.out,
         collectSessionBytes.in,
         terminationFanout.out(1))
@@ -104,10 +122,10 @@ private[http] object OutgoingConnectionBlueprint {
 
   // a simple merge stage that simply forwards its first input and ignores its second input
   // (the terminationBackchannelInput), but applies a special completion handling
-  private object TerminationMerge extends GraphStage[FanInShape2[HttpRequest, HttpResponse, HttpRequest]] {
-    private val requests = Inlet[HttpRequest]("requests")
+  private object TerminationMerge extends GraphStage[FanInShape2[RequestRenderingContext, HttpResponse, RequestRenderingContext]] {
+    private val requests = Inlet[RequestRenderingContext]("requests")
     private val responses = Inlet[HttpResponse]("responses")
-    private val out = Outlet[HttpRequest]("out")
+    private val out = Outlet[RequestRenderingContext]("out")
 
     override def initialAttributes = Attributes.name("TerminationMerge")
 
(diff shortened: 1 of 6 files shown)


ktoso:
Awesomeness! Thanks, will review after lunch :)

ktoso:
Should have said "after engineering meeting" I guess ;-)
Back home and onto it!

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -38,62 +42,75 @@ private[http] object OutgoingConnectionBlueprint {
...
       val collectSessionBytes = b.add(Flow[SslTlsInbound].collect { case s: SessionBytes ⇒ s })
-      collectSessionBytes ~> responseParsingMerge.in0
 
-      methodBypassFanout.out(0) ~> terminationMerge.in0
+      renderingContextCreation.out ~> bypassFanout.in
+      bypassFanout.out(0) ~> terminationMerge.in0
+      terminationMerge.out ~> requestRendering ~> logger ~> wrapTls
 
-      methodBypassFanout.out(1) ~> methodBypass ~> responseParsingMerge.in1
+      bypassFanout.out(1) ~> bypass ~> responseParsingMerge.in1
```
ktoso: Oh I see, it took me a bit to see why the dance of Promise.future and then casting back to Promise.
So the bypass branch may complete that promise, and the other side may take action on it...

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -38,62 +42,75 @@ private[http] object OutgoingConnectionBlueprint {
...
+        val requestRendererFactory = new HttpRequestRendererFactory(userAgentHeader, requestHeaderSizeHint, log)
+        Flow[RequestRenderingContext].flatMapConcat(requestRendererFactory.renderToSource).named("renderer")
+      }
 
-    val core = BidiFlow.fromGraph(GraphDSL.create() { implicit b ⇒
-      import GraphDSL.Implicits._
-      val methodBypassFanout = b.add(Broadcast[HttpRequest](2, eagerCancel = true))
-      val responseParsingMerge = b.add(new ResponseParsingMerge(rootParser))
+      val bypass = Flow[RequestRenderingContext] map { ctx ⇒
+        HttpResponseParser.ResponseContext(ctx.request.method, ctx.sendEntityTrigger.map(_.asInstanceOf[Promise[Unit]]))
```
ktoso: hah, learnt something new today – good to know it Promise extends Future, I never realised.

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -38,62 +42,75 @@ private[http] object OutgoingConnectionBlueprint {
...
+        // will not be mutated, all "shared copy" parsers copy on first-write into the header cache
+        val rootParser = new HttpResponseParser(parserSettings, HttpHeaderParser(parserSettings) { info ⇒
+          if (parserSettings.illegalHeaderWarnings)
+            logParsingError(info withSummaryPrepended "Illegal response header", log, parserSettings.errorLoggingVerbosity)
+        })
+        new ResponseParsingMerge(rootParser)
+      }
+
+      val responsePrep = Flow[List[ParserOutput.ResponseOutput]]
+        .mapConcat(conforms)
```
ktoso: use `ConstantFun.identity` instead :)

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
```diff
@@ -137,4 +166,19 @@ private[http] class HttpResponseParser(_settings: ParserSettings, _headerParser:
...
+}
+
+private[http] object HttpResponseParser {
+  /**
+   * @param requestMethod the request's HTTP method
+   * @param oneHundredContinueTrigger if the request contains an `Expect: 100-continue` header this option contains
+   *                                  a promise whose completion either triggers the sending of the (suspended)
+   *                                  request entity or the closing of the connection (for error completion)
+   */
+  private[http] case class ResponseContext(requestMethod: HttpMethod,
```
ktoso: minor: final

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
```diff
@@ -155,5 +166,14 @@ private[http] object HttpRequestRendererFactory {
...
+ * @param hostHeader the host header to render (not necessarily contained in the request.headers)
+ * @param sendEntityTrigger defined when the request has a `Expect: 100-continue` header; in this case the future will
+ *                          be completed successfully when the request entity is allowed to go out onto the wire;
+ *                          if the future is completed with an error the connection is to be closed.
  */
-private[http] final case class RequestRenderingContext(request: HttpRequest, hostHeader: Host)
+private[http] final case class RequestRenderingContext(
+  request: HttpRequest,
+  hostHeader: Host,
+  sendEntityTrigger: Option[Future[Unit]] = None)
```
ktoso: thanks for the good docs here

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
```diff
@@ -102,8 +104,16 @@ private[http] class HttpRequestRendererFactory(userAgentHeader: Option[headers.`
...
 
-    def renderStreamed(body: Source[ByteString, Any]): RequestRenderingOutput =
-      RequestRenderingOutput.Streamed(renderByteStrings(r, body))
+    def renderStreamed(body: Source[ByteString, Any]): RequestRenderingOutput = {
+      val headerPart = Source.single(r.get)
+      val stream = ctx.sendEntityTrigger match {
+        case None ⇒ headerPart ++ body
+        case Some(future) ⇒
+          val barrier = Source.fromFuture(future).drop(1).asInstanceOf[Source[ByteString, Any]]
+          (headerPart ++ barrier ++ body).recoverWith { case HttpResponseParser.OneHundredContinueError ⇒ Source.empty }
```
ktoso: Love how that reads with the barrier before body!

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -38,62 +42,75 @@ private[http] object OutgoingConnectionBlueprint {
...
-        logParsingError(info withSummaryPrepended "Illegal response header", log, parserSettings.errorLoggingVerbosity)
-    })
+    val core = BidiFlow.fromGraph(GraphDSL.create() { implicit b ⇒
+      import GraphDSL.Implicits._
 
-    val requestRendererFactory = new HttpRequestRendererFactory(userAgentHeader, requestHeaderSizeHint, log)
+      val renderingContextCreation = b.add {
+        Flow[HttpRequest] map { request ⇒
+          val sendEntityTrigger =
+            request.headers collectFirst { case headers.Expect.`100-continue` ⇒ Promise[Unit]().future }
```
ktoso: Shall we use `Done` or `NotUsed` as type param instead? It's just a trigger

ktoso:
Went through in detail, looks great – thanks!
Two minor nitpicks, nothing big really.

LGTM, will merge after another LGTM

jrudolph:
Will look through it but didn't get to it yet...

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
sirthias: I'll switch to `NotUsed`...

sirthias:
Ok, updated with Konrads comments addressed.

ktoso:
Thanks a lot, I'm happy to merge. Sounded like @jrudolph wants to have a look as well so waiting for a bit :)

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
```diff
@@ -17,19 +19,20 @@ import ParserOutput._
  */
 private[http] class HttpResponseParser(_settings: ParserSettings, _headerParser: HttpHeaderParser)
   extends HttpMessageParser[ResponseOutput](_settings, _headerParser) {
+  import HttpResponseParser._
   import HttpMessageParser._
   import settings._
 
-  private[this] var requestMethodForCurrentResponse: Option[HttpMethod] = None
+  private[this] var responseForCurrentResponse: Option[ResponseContext] = None
```
jrudolph: Sounds a bit strange :) Was this meant to be called `responseContextForCurrentResponse`?

jrudolph: Or just `contextForCurrentResponse`?

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
```diff
@@ -102,8 +105,16 @@ private[http] class HttpRequestRendererFactory(userAgentHeader: Option[headers.`
...
       if (method.isEntityAccepted && (contentLength > 0 || method.requestEntityAcceptance == Expected)) r ~~ `Content-Length` ~~ contentLength ~~ CrLf else r
 
-    def renderStreamed(body: Source[ByteString, Any]): RequestRenderingOutput =
-      RequestRenderingOutput.Streamed(renderByteStrings(r, body))
+    def renderStreamed(body: Source[ByteString, Any]): RequestRenderingOutput = {
+      val headerPart = Source.single(r.get)
+      val stream = ctx.sendEntityTrigger match {
+        case None ⇒ headerPart ++ body
+        case Some(future) ⇒
+          val barrier = Source.fromFuture(future).drop(1).asInstanceOf[Source[ByteString, Any]]
```
jrudolph: Do we have this as a combinator somewhere? I think I have used the same in some test before.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
```diff
@@ -17,19 +19,20 @@ import ParserOutput._
...
   private[this] var statusCode: StatusCode = StatusCodes.OK
 
   def createShallowCopy(): HttpResponseParser = new HttpResponseParser(settings, headerParser.createShallowCopy())
 
-  def setRequestMethodForNextResponse(requestMethod: HttpMethod): Unit =
-    if (requestMethodForCurrentResponse.isEmpty) requestMethodForCurrentResponse = Some(requestMethod)
+  def setRequestContextForNextResponse(responseContext: ResponseContext): Unit =
```
jrudolph: This name here seems strange as well ;)

jrudolph:
Very nice LGTM. Maybe the names could be improved a bit. Do we know of a server against which this could be tested "for real"?

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
sirthias: Yes, `contextForCurrentResponse` is much better.

akka-http-core/src/main/scala/akka/http/impl/engine/parsing/HttpResponseParser.scala
sirthias: Right. It should be `setContextForNextResponse`.
Will rename.

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
sirthias: +1 on adding a combinator for that.
There is nothing HTTP specific to it.

sirthias:
Fixed the names as suggested by Johannes.

ktoso:
Excellent, thanks :)

akka-http-core/src/main/scala/akka/http/impl/engine/rendering/HttpRequestRendererFactory.scala
viktorklang: Isn't Source.fromFuture(future).drop(1) == Source.empty?

ktoso: Not quite, it's used as a `barrier` (think like stream-ish countdownlatch basically).
Once the future completes, it emits an element, which we drop, and the stream completes which then causes the rest of the concated stream to be emitted: `headerPart ++ barrier ++ body`. So barrier never emits anything, however it prevents `body` from being signalled until barrier is triggered.

viktorklang: @ktoso Ah, isn't this ´killSwitch` as discussed in https://github.com/akka/akka/pull/19895/files#r55548466

viktorklang: There's also Source.maybe:

`Source.maybe[ByteString].mapMaterializedValue(_.completeWith(future.map(_ => Done)))`

ktoso: `maybe` is not helping here, KillSwitch is. We can change once KillSwitch is merged

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
```diff
@@ -38,62 +43,75 @@ private[http] object OutgoingConnectionBlueprint {
...
-        logParsingError(info withSummaryPrepended "Illegal response header", log, parserSettings.errorLoggingVerbosity)
-    })
+    val core = BidiFlow.fromGraph(GraphDSL.create() { implicit b ⇒
+      import GraphDSL.Implicits._
 
-    val requestRendererFactory = new HttpRequestRendererFactory(userAgentHeader, requestHeaderSizeHint, log)
+      val renderingContextCreation = b.add {
+        Flow[HttpRequest] map { request ⇒
+          val sendEntityTrigger =
+            request.headers collectFirst { case headers.Expect.`100-continue` ⇒ Promise[NotUsed]().future }
```
viktorklang: `Promise[NotUsed]().future` will never complete, so any callbacks that are added to it will leak for as long as the Future is referenced. Is this intentional? (See: https://github.com/viktorklang/blog/blob/master/Futures-in-Scala-2.12-part-6.md#missing-utilities-unit--never )

ktoso: Read-on, it's cast to Promise below and actually utilised.

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
viktorklang: Wow, I'd expect a big nice comment explaining the rationale here. Also, this could silently break if Scala Futures implementation changes.

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
ktoso: We were thinking if replaceable with KillSwitch (see comment in that PR)

viktorklang: Then I would expect a //FIXME here so it doesn't get forgotten :)

ktoso: Heh, I'll add.

viktorklang: Thanks Konrad, I appreciate it! :)

akka-http-core/src/main/scala/akka/http/impl/engine/client/OutgoingConnectionBlueprint.scala
sirthias: Yeah, it's a bit of a hack.
But any change in the underlying Future implementation will break several tests, so it won't be silent.

viktorklang: If Future were to change in 2.11.9, it would break all released artifacts
that relied on such a hack. Silently. :-)

-- 
Cheers,
√

sirthias: True. We are in fact relying on non-public features of the std library API here, which are thus not protected by binary compatibility guarantees.
I didn't think of that...

I'll submit a patch to get this fixed.

viktorklang: How about we expedite the killSwitch-feature and use that from day 0?

-- 
Cheers,
√

Issue #19789, opened 2016-02-15
silence Java serialize warning for java.lang.String


patriknw:
I think all `java.lang.*` should be silenced, at least `String` and primitive types.

johanandren:
Is that because it is annoying in the tests? Isn't it still a bad idea, even for primitives and `String`?

patriknw:
Yes, it's annoying in tests, but Java serialization of String should not be too bad. The format will not change and performance should be alright since it is defined specially in the [Java Serialization Protocol](https://docs.oracle.com/javase/7/docs/platform/serialization/spec/protocol.html#8299).

PR #19962, opened 2016-03-07
fix for "silence Java serialize warning for java.lang.String"

wlk:
addresses #19789

### DIFF
--- a/akka-actor/src/main/scala/akka/serialization/Serialization.scala
+++ b/akka-actor/src/main/scala/akka/serialization/Serialization.scala
@@ -246,10 +246,14 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
   val serializerByIdentity: Map[Int, Serializer] =
     Map(NullSerializer.identifier -> NullSerializer) ++ serializers map { case (_, v) ⇒ (v.identifier, v) }
 
-  private def shouldWarnAboutJavaSerializer(serializedClass: Class[_], serializer: Serializer) =
-    settings.config.getBoolean("akka.actor.warn-about-java-serializer-usage") &&
+  private val isJavaSerializationWarningEnabled = settings.config.getBoolean("akka.actor.warn-about-java-serializer-usage")
+
+  private def shouldWarnAboutJavaSerializer(serializedClass: Class[_], serializer: Serializer) = {
+    isJavaSerializationWarningEnabled &&
       serializer.isInstanceOf[JavaSerializer] &&
-      !serializedClass.getName.startsWith("akka.")
+      !serializedClass.getName.startsWith("akka.") &&
+      !serializedClass.getName.startsWith("java.lang.")
+  }
 
 }
 
--- a/akka-actor-tests/src/test/scala/akka/serialization/SerializeSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/serialization/SerializeSpec.scala
@@ -420,12 +420,19 @@ class DefaultSerializationWarningSpec extends AkkaSpec(
   ConfigFactory.parseString("akka.actor.warn-about-java-serializer-usage = on")) {
 
   val ser = SerializationExtension(system)
+  val messagePrefix = "Using the default Java serializer for class.*"
 
   "Using the default Java serializer" must {
 
-    "log a warning" in {
-      EventFilter.warning(message = "Using the default Java serializer for class.*") intercept {
-        ser.serializerFor(classOf[java.lang.Integer])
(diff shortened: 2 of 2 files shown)


wlk:
PLS BUILD

ktoso:
Sorry @wlk, jenkins commands are for owners :)

wlk:
@ktoso ok then the docs are not clear enough, I thought only "OK TO TEST" is for repo owners, will create PR for that as well :)

akka-actor/src/main/scala/akka/serialization/Serialization.scala
```diff
@@ -246,10 +246,17 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
   val serializerByIdentity: Map[Int, Serializer] =
     Map(NullSerializer.identifier -> NullSerializer) ++ serializers map { case (_, v) ⇒ (v.identifier, v) }
 
-  private def shouldWarnAboutJavaSerializer(serializedClass: Class[_], serializer: Serializer) =
+  private def shouldWarnAboutJavaSerializer(serializedClass: Class[_], serializer: Serializer) = {
+    def isJavaPrimitive(packageName: String): Boolean = {
+      val primitives = List("java.lang.Boolean", "java.lang.Byte", "java.lang.Character", "java.lang.Double", "java.lang.Float", "java.lang.Integer", "java.lang.Long", "java.lang.Short", "java.lang.String")
```
ktoso: - should be allocated once somewhere (in object most likely)
- should be a set for more efficient `contains`
- would we be able to obtain those from config? I think so, since system should be available here :) Then we'd allow people to silence the warnings if they "really know about one"

ktoso:
Right, technically speaking PLS BUILD will work AFAIR but after the initial OK if I remember correctly

ktoso:
OK TO TEST

ktoso:
Thanks for picking this up! Let's also add a simple test to see that it considers these.

akka-actor/src/main/scala/akka/serialization/Serialization.scala
wlk: thanks for the comments, rewriting this to take class names from config, could take a while

ktoso: Now that I think about it we should special case the `java.lang.` prefix case (for perf), and open up for configuration but perhaps in another PR, I'll let @patriknw weigh in on that :)

wlk: ok, in this PR I'll just do checking only for prefix `java.lang.`, and we can followup this in the next PR

ktoso: sounds good :+1:

wlk:
renamed issue, seems like it's ready to be retested and merged

akka-actor/src/main/scala/akka/serialization/Serialization.scala
```diff
@@ -246,10 +246,12 @@ class Serialization(val system: ExtendedActorSystem) extends Extension {
   val serializerByIdentity: Map[Int, Serializer] =
     Map(NullSerializer.identifier -> NullSerializer) ++ serializers map { case (_, v) ⇒ (v.identifier, v) }
 
-  private def shouldWarnAboutJavaSerializer(serializedClass: Class[_], serializer: Serializer) =
+  private def shouldWarnAboutJavaSerializer(serializedClass: Class[_], serializer: Serializer) = {
     settings.config.getBoolean("akka.actor.warn-about-java-serializer-usage") &&
       serializer.isInstanceOf[JavaSerializer] &&
```
patriknw: Could you please fix a small thing here, while you are at it. This method is only called once per class, but I'm still somewhat uneasy with performance implications of the config lookup. Please extract `settings.config.getBoolean("akka.actor.warn-about-java-serializer-usage")` to a `private val` in the `Serialization` class.

ktoso: :+1:

ktoso:
Almost there :-)
Failure was unrelated `VirtualProcessorTest ` for which we have a fix pending AFAIR

wlk:
I have fixed last issue, let me know what are the next steps. Should I squash the commits?

ktoso:
Yes, please squash and title it as `=act #19789 do not log serialization warnings for java.lang.*`

patriknw:
LGTM after squash

johanandren:
LGTM after squash

ktoso:
Thanks @wlk !

PR #19974, opened 2016-03-08
add cleaner attribute

He-Pin:
What I add here is for the application managed object.like netty's `ReferenceCounted` object.currently,when we streaming that type of objectes directly to akka stream without a copy,it will cause a leak,when we drop things,like you are using a `PooledByteBufAllocator` etc.

So I add a new attribute and just updated the `Filter` one for demostration.

I think now I could use 

``` scala
Source.fromIterator(() => new AbstractIterator[ByteBuf] {
    def hasNext = true
    def next = pooledAlloc.buffer(100)
  }).toSink(Sink.ignore.withAttributes(Attributes.cleaner(ReferenceCountUtil.release)))
```

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/Attributes.scala
+++ b/akka-stream/src/main/scala/akka/stream/Attributes.scala
@@ -6,8 +6,9 @@ package akka.stream
 import java.util.Optional
 
 import akka.event.Logging
+
 import scala.annotation.tailrec
-import scala.reflect.{ classTag, ClassTag }
+import scala.reflect.{ClassTag, classTag}
 import akka.japi.function
 import akka.stream.impl.StreamLayout._
 import java.net.URLEncoder
@@ -169,7 +170,7 @@ final case class Attributes(attributeList: List[Attributes.Attribute] = Nil) {
 }
 
 /**
- * Note that more attributes for the [[ActorMaterializer]] are defined in [[ActorAttributes]].
+ * Note that more attributes for the [[ActorMaterializer]] are defined in [[akka.stream.ActorAttributes]].
  */
 object Attributes {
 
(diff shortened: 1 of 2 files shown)


ktoso:
I'd rather model the specific action (i.e. route to "discard") instead of modifying filter with attributes as to what it does.

hepin1989:
@ktoso the problem here is to fix the `Filter` or `Drop` and etc,which drop things.

ktoso:
Yeah, I understand but I would not feel comfortable with "hidden cleaner feature" that's hard to discover.
Instead something like these would "feel more like akka streams":

Random API proposals I'd be more comfortable with:

```scala
  class Thingy {
    def include = false
    def dispose = ()
  }

  val disposer: Sink[Thingy, NotUsed] = Flow[Thingy].to(Sink.foreach(_.dispose))

  Flow[Thingy].filter(_.include) // does nothing to filtered out elements
  Flow[Thingy].filter(_.include, disposer) // disposes filtered out elements

  Flow[Thingy].drop(10)
  Flow[Thingy].drop(10, disposer)
  Flow[Thingy].dropWhile(_.include)
  Flow[Thingy].dropWhile(_.include, disposer)
```

etc. 

Only problem is API expansion however the request to do stuff with "dropped" or "filtered out" elements.

The Attribute approach is too confusing when multiple "thing that uses cleaner" are in the same graph IMO, as one looses typesafety there and also it's hard to figure out where a cleaner is actually applied.

johanandren:
I'm not convinced the use case is common enough to warrant the API expansion, after all these are all pretty simple stages and not so hard to create your own versions of with whatever added features you need in some specific use case?

ktoso:
Or we provide an in-line partition; this would help here too.

hepin1989:
@johanandren Yes ,you can,but not that practical.

johanandren:
I agree more with that idea.

viktorklang:
This feature is just not tractable when distributed stream materialization is in place.

hepin1989:
Once you cross the node boundary,yes.And then we need do the clean up where the `cross` taking place.
Need more thought.

patriknw:
I agree with others that this attributes based api is not something we can include, sorry

Issue #19500, opened 2016-01-18
Make encoder and decoder used by akka.stream.io.Framing.simpleFramingProtocol public


pkinsky:
I'd like to make the encoder and decoder used by `akka.stream.io.Framing.simpleFramingProtocol` available. Basically, this means moving the encoder and decoder flows defined inside `simpleFramingProtocol` here outside of the function scope and renaming them to `simpleFramingProtocolEncoder` and `simpleFramingProtocolDecoder`. 

(from Framing.scala)
```scala
  def simpleFramingProtocol(maximumMessageLength: Int): BidiFlow[ByteString, ByteString, ByteString, ByteString, Unit] = {
    val decoder = lengthField(4, 0, maximumMessageLength + 4, ByteOrder.BIG_ENDIAN).map(_.drop(4))
    val encoder = Flow[ByteString].transform(() ⇒ new PushStage[ByteString, ByteString] {

      override def onPush(message: ByteString, ctx: Context[ByteString]): SyncDirective = {
        if (message.size > maximumMessageLength)
          ctx.fail(new FramingException(s"Maximum allowed message size is $maximumMessageLength " +
            s"but tried to send ${message.size} bytes"))
        else {
          val header = ByteString(
            (message.size >> 24) & 0xFF,
            (message.size >> 16) & 0xFF,
            (message.size >> 8) & 0xFF,
            message.size & 0xFF)
          ctx.push(header ++ message)
        }
      }

    })

    BidiFlow.fromFlowsMat(encoder, decoder)(Keep.left)
  }
```

I believe this is a cleaner design that simplifies more lightweight use cases where no bidiFlow is required, for example reading or writing a stream of length-prefixed messages to or from a file or process (via input/output stream), which currently requires fiddling with `Framing.lengthField`, choosing between big and little endian encodings, padding the length-header to ensure it is the correct size, etc (or using the graph DSL to bind `Sink.maybe` and `Sink.ignore` to two of the bidi flow's ports, but that's just inelegant).

I'd like to implement this myself (it's basically just a trivial refactoring, the existing tests exercising `simpleFramingProtocol` should suffice), as I understand it the proper procedure is to file an issue first and wait for approval.

PR #19982, opened 2016-03-09
make simpleFramingProtocol decoder and encoder public

wlk:
addresses #19500

Extracted encoder and decoder outside of `simpleFramingProtocol`

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/scaladsl/Framing.scala
+++ b/akka-stream/src/main/scala/akka/stream/scaladsl/Framing.scala
@@ -74,8 +74,20 @@ object Framing {
    *                             included in this limit.
    */
   def simpleFramingProtocol(maximumMessageLength: Int): BidiFlow[ByteString, ByteString, ByteString, ByteString, NotUsed] = {
-    val decoder = lengthField(4, 0, maximumMessageLength + 4, ByteOrder.BIG_ENDIAN).map(_.drop(4))
-    val encoder = Flow[ByteString].transform(() ⇒ new PushStage[ByteString, ByteString] {
+    BidiFlow.fromFlowsMat(simpleFramingProtocolEncoder(maximumMessageLength), simpleFramingProtocolDecoder(maximumMessageLength))(Keep.left)
+  }
+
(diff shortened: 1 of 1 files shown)


akka-stream/src/main/scala/akka/stream/scaladsl/Framing.scala
```diff
@@ -74,22 +74,29 @@ object Framing {
...
-    })
-
-    BidiFlow.fromFlowsMat(encoder, decoder)(Keep.left)
+    BidiFlow.fromFlowsMat(simpleFramingProtocolEncoder(maximumMessageLength), simpleFramingProtocolDecoder(maximumMessageLength))(Keep.left)
   }
 
+  /**
+    * Protocol decoder that is used by [[Framing#simpleFramingProtocol]]
+    */
+  def simpleFramingProtocolDecoder(maximumMessageLength: Int) = lengthField(4, 0, maximumMessageLength + 4, ByteOrder.BIG_ENDIAN).map(_.drop(4))
```
patriknw: explicit return type, please

patriknw:
LGTM after the return type

wlk:
explicit return type added

ktoso:
LGTM, please squash

ktoso:
OK TO TEST

wlk:
squashed and pushed

can you clarify commit message format I have seen everywhere else:
* IDK what `=` stands for (there are also other symbols)
* was able to guess that `str` is for the `akka-streams` submodule - is this correct?

ktoso:
We're not as strict for using the `=str` prefix nowadays, I personally like it, story is: Historically we've required it to be always present, but it confused contributors so we don't require it nowadays.

The semantics are:

```
prefix = mod ~ pro
mod = "+" | "-" | "=" | "!" # new api, remove api, no api change, breaking api change
pro = prefix(module name) # so str, clu or act would be such examples
```

It's mostly to know what modules a commit touched. But as I said, we're not very strict about them nowadays.

ktoso:
LGTM, thanks!

PR #19984, opened 2016-03-09
Add withoutSizeLimit overrides for Scala API

### LINKED ISSUE #19983: Scala API withoutSizeLimit overrides are missing
effectively switching from Scala DSL to Java DSL when `withoutSizeLimit` is used.

2m:
Fixes #19983

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
@@ -126,6 +126,11 @@ sealed trait RequestEntity extends HttpEntity with jm.RequestEntity with Respons
    */
   def withSizeLimit(maxBytes: Long): RequestEntity
 
+  /**
+   * See [[HttpEntity#withoutSizeLimit]].
+   */
+  def withoutSizeLimit: RequestEntity
+
   def transformDataBytes(transformer: Flow[ByteString, ByteString, Any]): RequestEntity
 }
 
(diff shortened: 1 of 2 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -228,7 +243,7 @@ object HttpEntity {
       if (contentType == this.contentType) this else copy(contentType = contentType)
 
     override def withSizeLimit(maxBytes: Long): UniversalEntity =
-      if (data.length <= maxBytes) this
+      if (data.length <= maxBytes || isKnownEmpty) this
```
2m: @ktoso this is something that I added to make `HttpEntity.Empty.withoutSizeLimit` run without `require` exceptions.

ktoso: Looks good :+1:

ktoso:
LGTM, thanks!

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -126,6 +126,11 @@ sealed trait RequestEntity extends HttpEntity with jm.RequestEntity with Respons
    */
   def withSizeLimit(maxBytes: Long): RequestEntity
 
+  /**
+   * See [[HttpEntity#withoutSizeLimit]].
+   */
+  def withoutSizeLimit: RequestEntity
```
patriknw: add `override` on this and other methods?

patriknw:
LGTM

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
ktoso: We can attempt that as a sweeping change across the http codebase I think. It does not use override, we had some discusions on that, moving towards using it more now.

PR #19989, opened 2016-03-09
Wip small doc fixes rk

### LINKED ISSUE #12772: DOC: sbt needs more memory for running Akka tests
imported from https://www.assembla.com/spaces/akka/tickets/2772

I was getting occasional failures due to the SupervisorHierarchySpec test running too slowly. It has been suggested that I need to increase my memory for building/testing Akka up to 3G. This needs investigating and documenting in the guide for building Akka.

### LINKED ISSUE #15120: BalancingPool doc improvements
The BalancingPool has the property that its routees do not have truly distinct identity: they have different names, but talking to them will not end up at the right actor in most cases. Therefore you cannot use it for workflows that require state to be kept within the routee, you would in this case have to include the whole state in the messages.

With a SmallestMailboxPool you can have a vertically scaling service that can interact in a stateful fashion with other services in the back-end before replying to the original client. The other advantage is that it does not place a restriction on the message queue implementation as BalancingPool does.

### LINKED ISSUE #15130: renovate README
It has gathered some dust, needs touch ups for the various new features developed over the past three years.

### LINKED ISSUE #15313: FAQ: careful with ActorSystem creation
double-creation will not be detected by the ActorSystem itself, other than e.g. remoting failing to bind to its configured port; never use `def system: ActorSystem` to create it, make it a `val`

### LINKED ISSUE #15314: FAQ: cannot exit REPL
point out that ActorSystem must be shutdown, otherwise REPL will wait indefinitely for its thread pools to shut down

rkuhn:
fixes #12772 
fixes #15120 
fixes #15130 
fixes #15313 
fixes #15314

### DIFF
--- a/akka-docs/_sphinx/themes/akka/layout.html
+++ b/akka-docs/_sphinx/themes/akka/layout.html
@@ -25,10 +25,10 @@
 {% block relbar2 %}{% endblock %}
 
 {% block extrahead %}
+  {%- if include_analytics %}
     <!-- Hint to search engines that the "canonical" page is under "current", which will boost it appearing in search results -->
     <link rel="canonical" href="http://akka.io/docs/akka/current/{{ pagename }}.html" />
 
-  {%- if include_analytics %}
     <!--Google Analytics-->
     <script type="text/javascript">
       var _gaq = _gaq || [];
--- a/akka-docs/rst/additional/faq.rst
+++ b/akka-docs/rst/additional/faq.rst
@@ -37,6 +37,16 @@ you have a matching ``stop``, ``terminate``, or ``shutdown`` call implemented.
 In particular you typically want to bind such values to immutable references, i.e.
 ``final ActorSystem system`` in Java or ``val system: ActorSystem`` in Scala.
 
+JVM application or Scala REPL “hanging”
+---------------------------------------
+
+Due to an ActorSystem’s explicit lifecycle the JVM will not exit until it is stopped.
+Therefore it is necessary to shutdown all ActorSystems within a running application or
+Scala REPL session in order to allow these processes to terminate.
+
+Shutting down an ActorSystem will properly terminate all Actors and ActorMaterializers
+that were created within it.
+
 Actors in General
 ^^^^^^^^^^^^^^^^^
 
--- a/akka-docs/rst/additional/faq.rst
+++ b/akka-docs/rst/additional/faq.rst
@@ -25,6 +25,18 @@ Akka is also:
 * a town in Morocco
 * a near-earth asteroid
 
+Resources with Explicit Lifecycle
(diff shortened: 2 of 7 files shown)


akka-docs/_sphinx/themes/akka/layout.html
```diff
@@ -25,10 +25,10 @@
 {% block relbar2 %}{% endblock %}
 
 {% block extrahead %}
+  {%- if include_analytics %}
```
ktoso: thanks

akka-docs/rst/additional/faq.rst
```diff
@@ -25,6 +25,28 @@ Akka is also:
...
+Actors, ActorSystems, ActorMaterializers (for streams), all these types of objects bind
+resources that must be released explicitly. The reason is that Actors are meant to have
+a life of their own, existing independently of whether messages are currently en route
+to them. Therefore you should always make sure that for every creation of such an object
+you have a matching ``stop``, ``terminate``, or ``shutdown`` call implemented.
+
+In particular you typically want to bind such values to immutable references, i.e.
+``final ActorSystem system`` in Java or ``val system: ActorSystem`` in Scala.
+
+JVM application or Scala REPL “hanging”
```
ktoso: yeah, great one to add (do we want to mention it's using non daemon threads? I worded it a bit differently here http://stackoverflow.com/questions/35583739/why-does-this-simple-akka-streams-program-never-terminate/35584236#35584236 perhaps too much tech detail?)

akka-docs/rst/intro/what-is-akka.rst
```diff
@@ -4,9 +4,9 @@
  What is Akka?
 ###############
 
-**Scalable real-time transaction processing**
+**«resilient elastic distributed real-time transaction processing»**
```
ktoso: Resilient?

rkuhn: it is not a sentence, so my syntax analyzer does not grant it capitalization ;-)

akka-docs/rst/additional/faq.rst
rkuhn: For the FAQ I’d like to not go there: it will only lead to people complaining that their app does not do anything, it just stops.

akka-docs/rst/intro/what-is-akka.rst
ktoso: weird but ok:)

akka-docs/rst/additional/faq.rst
ktoso: Ok :+1:

ktoso:
LGTM

patriknw:
LGTM

PR #19990, opened 2016-03-10
19834 replace collect with graph stage

agolubev:
also added generalized class for stage logic to encapsulate supervision catching logic
Ref #19834

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowCollectSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowCollectSpec.scala
@@ -3,15 +3,24 @@
  */
 package akka.stream.scaladsl
 
+import akka.stream.ActorAttributes._
+import akka.stream.Supervision._
+import akka.stream.impl.ConstantFun
+import akka.stream.testkit.Utils.TE
+import akka.stream.testkit.scaladsl.TestSink
+
 import scala.concurrent.forkjoin.ThreadLocalRandom.{ current ⇒ random }
 
-import akka.stream.ActorMaterializerSettings
+import akka.stream.{ ActorMaterializer, ActorMaterializerSettings }
 import akka.testkit.AkkaSpec
-import akka.stream.testkit.ScriptedTest
+import akka.stream.testkit.{ TestSubscriber, ScriptedTest }
+
+import scala.util.control.NoStackTrace
 
 class FlowCollectSpec extends AkkaSpec with ScriptedTest {
 
   val settings = ActorMaterializerSettings(system)
+  implicit val materializer = ActorMaterializer(settings)
 
   "A Collect" must {
 
@@ -21,6 +30,19 @@ class FlowCollectSpec extends AkkaSpec with ScriptedTest {
         Seq(x) -> (if ((x & 1) == 0) Seq((x * x).toString) else Seq.empty[String])
       }: _*)
       TestConfig.RandomTestRange foreach (_ ⇒ runScript(script, settings)(_.collect { case x if x % 2 == 0 ⇒ (x * x).toString }))
+    }
+
+    "restart when Collect throws" in {
+      val pf: PartialFunction[Int, Int] =
+        { case x: Int ⇒ if (x == 2) throw TE("") else x }
+      Source(1 to 3).collect(pf).withAttributes(supervisionStrategy(restartingDecider))
+        .runWith(TestSink.probe[Int])
+        .request(1)
+        .expectNext(1)
+        .request(1)
+        .expectNext(3)
+        .request(1)
+        .expectComplete()
     }
 
   }
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -59,38 +59,89 @@ private[akka] final case class TakeWhile[T](p: T ⇒ Boolean, decider: Supervisi
 /**
  * INTERNAL API
  */
-private[akka] final case class DropWhile[T](p: T ⇒ Boolean, decider: Supervision.Decider) extends PushStage[T, T] {
-  var taking = false
+private[stream] final case class DropWhile[T](p: T ⇒ Boolean) extends GraphStage[FlowShape[T, T]] {
+  val in = Inlet[T]("DropWhile.in")
+  val out = Outlet[T]("DropWhile.out")
+  override val shape = FlowShape(in, out)
+  override def initialAttributes: Attributes = DefaultAttributes.dropWhile
 
-  override def onPush(elem: T, ctx: Context[T]): SyncDirective =
-    if (taking || !p(elem)) {
-      taking = true
-      ctx.push(elem)
-    } else {
-      ctx.pull()
+  def createLogic(inheritedAttributes: Attributes) = new SupervisedGraphStageLogic(inheritedAttributes, shape) with InHandler with OutHandler {
+    override def onPush(): Unit = {
+      val elem = grab(in)
+      withSupervision(() ⇒ p(elem)) match {
+        case Some(flag) if flag ⇒ pull(in)
+        case Some(flag) if !flag ⇒
+          push(out, elem)
+          setHandler(in, rest)
+        case None ⇒ // do nothing
+      }
     }
 
-  override def decide(t: Throwable): Supervision.Directive = decider(t)
+    def rest = new InHandler {
+      def onPush() = push(out, grab(in))
+    }
+
+    override def onResume(t: Throwable): Unit = if (!hasBeenPulled(in)) pull(in)
+    override def onPull(): Unit = pull(in)
+    setHandlers(in, out, this)
+  }
+  override def toString = "DropWhile"
 }
 
-private[akka] object Collect {
+/**
+ * INTERNAL API
+ */
+abstract private[stream] class SupervisedGraphStageLogic(inheritedAttributes: Attributes, shape: Shape) extends GraphStageLogic(shape) {
+  private lazy val decider = inheritedAttributes.get[SupervisionStrategy].map(_.decider).getOrElse(Supervision.stoppingDecider)
+  def withSupervision[T](f: () ⇒ T): Option[T] =
+    try { Some(f()) } catch {
+      case NonFatal(ex) ⇒
+        decider(ex) match {
+          case Supervision.Stop    ⇒ onStop(ex)
+          case Supervision.Resume  ⇒ onResume(ex)
+          case Supervision.Restart ⇒ onRestart(ex)
+        }
+        None
+    }
+
+  def onResume(t: Throwable): Unit
+  def onStop(t: Throwable): Unit = failStage(t)
+  def onRestart(t: Throwable): Unit = onResume(t)
+}
+
+private[stream] object Collect {
   // Cached function that can be used with PartialFunction.applyOrElse to ensure that A) the guard is only applied once,
   // and the caller can check the returned value with Collect.notApplied to query whether the PF was applied or not.
   // Prior art: https://github.com/scala/scala/blob/v2.11.4/src/library/scala/collection/immutable/List.scala#L458
   final val NotApplied: Any ⇒ Any = _ ⇒ Collect.NotApplied
 }
 
-private[akka] final case class Collect[In, Out](pf: PartialFunction[In, Out], decider: Supervision.Decider) extends PushStage[In, Out] {
+/**
+ * INTERNAL API
+ */
+private[stream] final case class Collect[In, Out](pf: PartialFunction[In, Out]) extends GraphStage[FlowShape[In, Out]] {
+  val in = Inlet[In]("Collect.in")
+  val out = Outlet[Out]("Collect.out")
+  override val shape = FlowShape(in, out)
+  override def initialAttributes: Attributes = DefaultAttributes.collect
 
-  import Collect.NotApplied
+  def createLogic(inheritedAttributes: Attributes) = new SupervisedGraphStageLogic(inheritedAttributes, shape) with InHandler with OutHandler {
+    import Collect.NotApplied
+    val wrappedPf = () ⇒ pf.applyOrElse(grab(in), NotApplied)
 
-  override def onPush(elem: In, ctx: Context[Out]): SyncDirective =
-    pf.applyOrElse(elem, NotApplied) match {
-      case NotApplied             ⇒ ctx.pull()
-      case result: Out @unchecked ⇒ ctx.push(result)
+    override def onPush(): Unit = withSupervision(wrappedPf) match {
+      case Some(result) ⇒ result match {
+        case NotApplied             ⇒ pull(in)
+        case result: Out @unchecked ⇒ push(out, result)
+      }
+      case None ⇒ //do nothing
     }
 
-  override def decide(t: Throwable): Supervision.Directive = decider(t)
+    override def onResume(t: Throwable): Unit = if (!hasBeenPulled(in)) pull(in)
+    override def onPull(): Unit = pull(in)
+    setHandlers(in, out, this)
+  }
+  override def toString = "Collect"
 }
 
 /**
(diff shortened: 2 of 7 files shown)


akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -73,24 +73,54 @@ private[akka] final case class DropWhile[T](p: T ⇒ Boolean, decider: Supervisi
   override def decide(t: Throwable): Supervision.Directive = decider(t)
 }
 
-private[akka] object Collect {
+private[stream] class SupervisionedGraphStageLogic(inheritedAttributes: Attributes, shape: Shape) extends GraphStageLogic(shape) {
```
hepin1989: +1 for this:)

Supervised

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -73,24 +73,54 @@ private[akka] final case class DropWhile[T](p: T ⇒ Boolean, decider: Supervisi
...
-  override def onPush(elem: In, ctx: Context[Out]): SyncDirective =
-    pf.applyOrElse(elem, NotApplied) match {
-      case NotApplied             ⇒ ctx.pull()
-      case result: Out @unchecked ⇒ ctx.push(result)
-    }
+    override def onResume(t: Throwable): Unit = if (!hasBeenPulled(in)) pull(in)
+    override def onRestart(t: Throwable): Unit = onResume(t)
 
-  override def decide(t: Throwable): Supervision.Directive = decider(t)
+    override def onPull(): Unit = pull(in)
```
hepin1989: how about add another 
`SupervisedPushPullGraphStageLogic` and `SupervisedPushGraphStageLogic`?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -73,24 +73,54 @@ private[akka] final case class DropWhile[T](p: T ⇒ Boolean, decider: Supervisi
   override def decide(t: Throwable): Supervision.Directive = decider(t)
 }
 
-private[akka] object Collect {
+private[stream] class SupervisionedGraphStageLogic(inheritedAttributes: Attributes, shape: Shape) extends GraphStageLogic(shape) {
+  val decider = inheritedAttributes.get[SupervisionStrategy].map(_.decider).getOrElse(Supervision.stoppingDecider)
```
hepin1989: @rkuhn want this https://github.com/akka/akka/issues/19950.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -59,38 +59,82 @@ private[akka] final case class TakeWhile[T](p: T ⇒ Boolean, decider: Supervisi
 /**
  * INTERNAL API
  */
-private[akka] final case class DropWhile[T](p: T ⇒ Boolean, decider: Supervision.Decider) extends PushStage[T, T] {
+private[stream] final case class DropWhile[T](p: T ⇒ Boolean) extends GraphStage[FlowShape[T, T]] {
+  val in = Inlet[T]("DropWhile.in")
+  val out = Outlet[T]("DropWhile.out")
+  override val shape = FlowShape(in, out)
+  override def initialAttributes: Attributes = DefaultAttributes.dropWhile
   var taking = false
```
viktorklang: This var can't be here, it needs to be inside the `SupervisionedGraphStageLogic`

viktorklang: @patriknw Heads up, this is likely a bug in 2.4.2

rkuhn: Ouch, well spotted!

viktorklang: @rkuhn I don't have any cycles right now, do you want me to open a ticket so we can make sure to get this into 2.4.3?

ktoso: Yes please, we'll pick it up from there

rkuhn: opened #20009

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -59,38 +59,82 @@ private[akka] final case class TakeWhile[T](p: T ⇒ Boolean, decider: Supervisi
...
+    override def onPull(): Unit = pull(in)
+    setHandlers(in, out, this)
+  }
+  override def toString = "DropWhile"
+}
+
+/**
+  * INTERNAL API
+  */
+abstract private[stream] class SupervisionedGraphStageLogic(inheritedAttributes: Attributes, shape: Shape) extends GraphStageLogic(shape) {
```
johanandren: Nitpick: it should be "Supervised" not "Supervisioned"

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -59,38 +59,82 @@ private[akka] final case class TakeWhile[T](p: T ⇒ Boolean, decider: Supervisi
...
-      ctx.pull()
+  def createLogic(inheritedAttributes: Attributes) = new SupervisionedGraphStageLogic(inheritedAttributes, shape) with InHandler with OutHandler {
+    override def onPush(): Unit = execute(() ⇒ {
+      val el = grab(in)
+      if (taking || !p(el)) {
+        taking = true
+        push(out, el)
+      } else
+        pull(in)
+    })
```
johanandren: Something I personally am unsure about, shouldn't the supervision logic _only_ be for the call to `p` here, with the current impl it would do supervision if the stage threw "pulled twice" etc. which isn't really something that can be resumed/restarted?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -59,38 +59,82 @@ private[akka] final case class TakeWhile[T](p: T ⇒ Boolean, decider: Supervisi
...
+    setHandlers(in, out, this)
+  }
+  override def toString = "DropWhile"
+}
+
+/**
+  * INTERNAL API
+  */
+abstract private[stream] class SupervisionedGraphStageLogic(inheritedAttributes: Attributes, shape: Shape) extends GraphStageLogic(shape) {
+  val decider = inheritedAttributes.get[SupervisionStrategy].map(_.decider).getOrElse(Supervision.stoppingDecider)
```
johanandren: could be private or at least protected?

johanandren: Also, there was some discussion about making decider a `lazy val` or a `def` since it is the exceptional case that it is accessed rather than the regular case. Not sure if that can be applied generally here but something to think about at least.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowCollectSpec.scala
```diff
@@ -23,6 +31,20 @@ class FlowCollectSpec extends AkkaSpec with ScriptedTest {
       TestConfig.RandomTestRange foreach (_ ⇒ runScript(script, settings)(_.collect { case x if x % 2 == 0 ⇒ (x * x).toString }))
     }
 
+    "restart when Collect throws" in {
+      val ex = new Exception("test") with NoStackTrace
```
johanandren: There is a test exception available in the testkit already, `akka.stream.testkit.Utils.TE`

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowCollectSpec.scala
```diff
@@ -23,6 +31,20 @@ class FlowCollectSpec extends AkkaSpec with ScriptedTest {
...
+      Source(1 to 3).collect(pf).withAttributes(supervisionStrategy(restartingDecider))
+        .runWith(TestSink.probe[Int])
+        .request(1)
+        .expectNext(1)
+        .request(1)
+        .expectNext(3)
+        .request(1)
+        .expectComplete()
+    }
+
```
johanandren: :+1:

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
johanandren: Wait a minute, it was a `PushStage` in 2.4.2, so it wasn't wrong until now when it became a `GraphStage`?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
agolubev: nice point! - will make it lazy

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
agolubev: Hm, need to fix restart strategy.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
agolubev: Idea here is that `execute` mimic `try catch` so no code will be executed after exception anywhere inside parameter function.
Regarding `pulled twice` problem I'm not sure - perhaps it's inconsistent sate.  I would say that `pulled twice` is like run time exception and need to be fixed in opposite to regular exception that must be handled (according to Java vision at least). 
Let me know if there is some code snippet that shows different approach. Maybe I missed some important discussion on this topic.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -59,38 +59,82 @@ private[akka] final case class TakeWhile[T](p: T ⇒ Boolean, decider: Supervisi
...
-    if (taking || !p(elem)) {
-      taking = true
-      ctx.push(elem)
-    } else {
-      ctx.pull()
+  def createLogic(inheritedAttributes: Attributes) = new SupervisionedGraphStageLogic(inheritedAttributes, shape) with InHandler with OutHandler {
+    override def onPush(): Unit = execute(() ⇒ {
+      val el = grab(in)
+      if (taking || !p(el)) {
+        taking = true
```
hepin1989: how about not using the `tacking`,but reseting the OutHandler.

viktorklang: Good idea!

-- 
Cheers,
√

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -59,38 +59,82 @@ private[akka] final case class TakeWhile[T](p: T ⇒ Boolean, decider: Supervisi
...
+      execute(() ⇒ {
+        pf.applyOrElse(grab(in), NotApplied) match {
+          case NotApplied             ⇒ pull(in)
+          case result: Out @unchecked ⇒ push(out, result)
+        }
+      })
 
-  override def decide(t: Throwable): Supervision.Directive = decider(t)
+    override def onResume(t: Throwable): Unit = if (!hasBeenPulled(in)) pull(in)
+    override def onPull(): Unit = pull(in)
```
hepin1989: I still think we could have a `SupervsedPushGraphStageLogic` and not repeat this `override def onPull(): Unit = pull(in)` again and again.

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -59,38 +59,82 @@ private[akka] final case class TakeWhile[T](p: T ⇒ Boolean, decider: Supervisi
...
+  }
+  override def toString = "DropWhile"
+}
+
+/**
+  * INTERNAL API
+  */
+abstract private[stream] class SupervisionedGraphStageLogic(inheritedAttributes: Attributes, shape: Shape) extends GraphStageLogic(shape) {
+  val decider = inheritedAttributes.get[SupervisionStrategy].map(_.decider).getOrElse(Supervision.stoppingDecider)
+  def execute(f: () ⇒ Unit): Unit =
```
hepin1989: `safeHold`?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -59,38 +59,82 @@ private[akka] final case class TakeWhile[T](p: T ⇒ Boolean, decider: Supervisi
...
+    try { f() } catch {
+      case NonFatal(ex) ⇒ decider(ex) match {
+        case Supervision.Stop    ⇒ onStop(ex)
+        case Supervision.Resume  ⇒ onResume(ex)
+        case Supervision.Restart ⇒ onRestart(ex)
```
hepin1989: The most common thing is `Stop`?I think how about：

1. Resume
2. Restart
3. Stop

？

agolubev: Stop is default

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hepin1989: eg:
```scala
abstract private[stream] class SupervisedPushPullGraphStageLogic[-In, +Out](inheritedAttributes: Attributes, shape: FlowShape[In, Out])
  extends SupervisionedGraphStageLogic(inheritedAttributes, shape) with InHandler with OutHandler {
  setHandlers(shape.in, shape.out, this)
}

abstract private[stream] class SupervisedPushGraphStageLogic[-In, +Out](inheritedAttributes: Attributes, shape: FlowShape[In, Out])
  extends SupervisedPushPullGraphStageLogic(inheritedAttributes, shape) {
  override def onPull() = pull(shape.in)
}
```

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
hepin1989: thank you :)

patriknw:
LGTM

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -59,38 +59,81 @@ private[akka] final case class TakeWhile[T](p: T ⇒ Boolean, decider: Supervisi
...
-      ctx.pull()
+  def createLogic(inheritedAttributes: Attributes) = new SupervisedGraphStageLogic(inheritedAttributes, shape) with InHandler with OutHandler {
+    override def onPush(): Unit = execute(() ⇒ {
+      val elem = grab(in)
+      if (p(elem)) pull(in)
+      else {
+        push(out, elem)
+        setHandler(in, rest)
+      }
+    })
```
johanandren: Won't this allocate a function on each push?

viktorklang: Yes. (And what does `execute` do?)

agolubev: Ok, will rename it to tryExecute (other suggestions?) and create function as val

johanandren: `withSupervision`, `supervised`?

agolubev: I like ```withSupervision``` one

agolubev: should I also add what is suggested in https://github.com/akka/akka/pull/19990/files/e0d4b2e4ecf1a0559ccb477e690b6350368f5eec#r55859774 ?

johanandren: I don't think so, feels like type explosion to save one line of code to me.

agolubev:
Changed according to feedback. Looking for another LGTM for rebase.

ktoso:
OK TO TEST

agolubev:
it was Mima

johanandren:
LGTM

agolubev:
Rebased and squashed

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: pulled twice is an internal error in case of filter and should not be supervised. We should only recover user errors (like a division by zero in the predicate)

agolubev:
Was thinking about @drewhk  comment. ("We should only recover user errors (like a division by zero in the predicate)") So need to fix exception handling. Hold on.

ktoso:
LGTM, needs a rebase now though (likely due to MiMa.scala)

agolubev:
@ktoso, PR is rebased.

drewhk:
LGTM

Issue #19162, opened 2015-12-13
Akka HTTP Java API: NullPointerException when using AccessControlAllowOrigin.create(HttpOriginRange.ALL)


relgames:
I'm writing a service which calls one or several external services and transforms their responses and then sends back to a requestor.
Code in question:

```java
    public Future<RouteResult> login(RequestContext ctx, BaasBoxCtx baasBoxCtx, OAuth auth) {
        log.debug("Got request {}", ctx.request());
        String oauthJson = toJson(auth);

        HttpRequest request = HttpRequest.POST(cfg.baasBoxUrl("/social/facebook"))
                .withEntity(ContentTypes.APPLICATION_JSON, oauthJson)
                .addHeader(baasBoxCtx.getAppCodeHeader());

        return httpClient.request(request)
                .flatMap(func(response -> {
                    Option<HttpHeader> session = response.getHeader(BaasBoxCtx.SESSION_HEADER);

                    log.debug("Got from BaasBox: {}", response);

                    if (session.isDefined()) {
                        return response.entity().getDataBytes()
                                .map(byteString -> {
                                    String facebookJson = byteString.utf8String();
                                    sendToActor(baasBoxCtx, auth, session, facebookJson);
                                    return ctx.complete(HttpResponse.create()
                                            .addHeader(AccessControlAllowOrigin.create(HttpOriginRange.ALL))
                                            .addHeader(session.get())
                                            .addHeader(baasBoxCtx.getAppCodeHeader())
                                            .withEntity(ContentTypes.APPLICATION_JSON, facebookJson)
                                    );
                                })
                                .runWith(Sink.head(), materializer);
                    } else {
                        return Futures.successful(ctx.complete(response));
                    }
                }), ctx.executionContext());
    }
```

It fails with NPE when a request is made:

```
[2015-12-13 20:32:15,970] [  ERROR] [spatcher-8] [akka.actor.ActorSystemImpl(feedigy)] a.a.ActorSystemImpl: Outgoing response stream error
java.lang.NullPointerException: null
	at akka.http.impl.util.Renderer$RenderableRenderer$.render(Rendering.scala:100) ~[akka-http-core-experimental_2.11-2.0-M2.jar:na]
	at akka.http.impl.util.Renderer$RenderableRenderer$.render(Rendering.scala:99) ~[akka-http-core-experimental_2.11-2.0-M2.jar:na]
[...]

PR #20011, opened 2016-03-11
+htc #19162 unholy fix for javadsl initialization issues

ktoso:
Unholy touching of static objects and setting them.

Read https://github.com/akka/akka/issues/19162 for details.
Horrible, but safe yet safe and keeps bin compat.

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/RemoteAddress.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/RemoteAddress.java
@@ -7,6 +7,8 @@ package akka.http.javadsl.model;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.util.Optional;
+
+import akka.http.javadsl.model.headers.HttpEncodingRanges;
 import scala.compat.java8.OptionConverters;
 
 public abstract class RemoteAddress {
@@ -19,7 +21,6 @@ public abstract class RemoteAddress {
      */
     public abstract int getPort();
 
-    public static final RemoteAddress UNKNOWN = akka.http.scaladsl.model.RemoteAddress.Unknown$.MODULE$;
     public static RemoteAddress create(InetAddress address) {
         return akka.http.scaladsl.model.RemoteAddress.apply(address, OptionConverters.toScala(Optional.empty()));
     }
@@ -29,4 +30,13 @@ public abstract class RemoteAddress {
     public static RemoteAddress create(byte[] address) {
         return akka.http.scaladsl.model.RemoteAddress.apply(address);
     }
+
+    /**
+     * @deprecated because of troublesome initialisation order (with regards to scaladsl class implementing this class).
+     *             In some edge cases this field could end up containing a null value.
+     *             Will be removed in Akka 3.x, use {@link HttpEncodingRanges#ALL} instead.
+     */
+    @Deprecated
+    // FIXME: Remove in Akka 3.0
+    public static final RemoteAddress UNKNOWN = RemoteAddresses.UNKNOWN;
 }
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/RemoteAddress.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/RemoteAddress.java
@@ -34,7 +34,7 @@ public abstract class RemoteAddress {
     /**
      * @deprecated because of troublesome initialisation order (with regards to scaladsl class implementing this class).
      *             In some edge cases this field could end up containing a null value.
-     *             Will be removed in Akka 3.x, use {@link HttpEncodingRanges#ALL} instead.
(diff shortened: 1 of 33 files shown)


akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/EntityTag.scala
```diff
@@ -37,4 +39,15 @@ object EntityTagRange {
     require(tags.nonEmpty, "tags must not be empty")
     def render[R <: Rendering](r: R): r.type = r ~~ tags
   }
+
+  {
+    // FIXME: Exists only to fix JavaDSL init problem: #19162. Remove in Akka 3.x when "EMPTY" static fields moved
+    val u = Unsafe.instance
+    def initialize(f: java.lang.reflect.Field) =
+      u.putObject(u.staticFieldBase(f), u.staticFieldOffset(f), `*`)
```
ktoso: can be here, or technically inside the 

```
`*`
```

object, but this seemed somewhat safer (earlier init)

akka-http-core/src/main/java/akka/http/javadsl/model/RemoteAddresses.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+
+package akka.http.javadsl.model;
+
+public abstract class RemoteAddresses {
```
viktorklang: Is this intended to be extended?

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -215,6 +215,16 @@ object HttpEntity {
     if (contentType == Empty.contentType) Empty
     else HttpEntity.Strict(contentType, data = ByteString.empty)
 
+  {
+    // FIXME: Exists only to fix JavaDSL init problem: #19162. Remove in Akka 3.x when "EMPTY" static fields moved
+    val u = Unsafe.instance
```
viktorklang: Is there a reason for having this code duplicated many times? It would seem easier to create a utility for this so that the code becomes a one-liner which is easy to see all occurrences of?

akka-http-core/src/main/java/akka/http/javadsl/model/RemoteAddresses.java
ktoso: No, good catch, will sprinkle some finals around

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
ktoso: Yeah, can do.

akka-http-core/src/main/java/akka/http/javadsl/model/RemoteAddress.java
```diff
@@ -29,4 +30,13 @@ public static RemoteAddress create(InetSocketAddress address) {
     public static RemoteAddress create(byte[] address) {
         return akka.http.scaladsl.model.RemoteAddress.apply(address);
     }
+
+    /**
+     * @deprecated because of troublesome initialisation order (with regards to scaladsl class implementing this class).
+     *             In some edge cases this field could end up containing a null value.
+     *             Will be removed in Akka 3.x, use {@link HttpEncodingRanges#ALL} instead.
```
rkuhn: Copy pasta 😉

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
```diff
@@ -215,6 +217,11 @@ object HttpEntity {
     if (contentType == Empty.contentType) Empty
     else HttpEntity.Strict(contentType, data = ByteString.empty)
 
+  JavaInitialization.initializeStaticFieldWith(
+    Empty,
+    classOf[jm.HttpEntity].getField("EMPTY"),
+    classOf[jm.HttpEntities].getField("EMPTY"))
```
rkuhn: Are these secondary targets really needed? I thought that with the placement of the constants in separate classes the problem should be fixed.

rkuhn:
LGTM apart from the one question

patriknw:
LGTM

akka-http-core/src/main/java/akka/http/javadsl/model/RemoteAddress.java
ktoso: thanks

akka-http-core/src/main/scala/akka/http/scaladsl/model/HttpEntity.scala
ktoso: I think you're right, fixing

ktoso:
Should now pass bin compat.

Issue #19924, opened 2016-03-01
Don't build debug identation string in stream materialization


drewhk:
StreamLayout:669

ktoso:
One of @rkuhn's PRs contains a fix for this.

PR #20016, opened 2016-03-11
simplify materialized value computation tree, fixes #20015

### LINKED ISSUE #20015: compress materialized value computation tree
When performing 10000 map operations in a row, there is no reason for the resulting graph module to have a materialized value computation tree of depth 10000: all composition nodes just do `Keep.left` anyway.

Solving this ticket will take the immediate pressure off our materialization and fusing algorithms to trampoline the tree traversals, which is rather tricky to do. A graph with materialized value computation depth of several thousand levels is considered a special use-case that requires tuning the JVM stack size.

rkuhn:
- also fixes materialized value sources for graphs that import zero or
  one graphs, with and without Fusing

The bulk of this PR is just collateral damage:
- I made the Module trait sealed, allowing only AtomicModules to be created elsewhere
- I added useful toString representations to everything, especially for using the various Debug constants

(The latter was crucial in debugging the issues encountered during this whole endeavor.)

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/FlowModule.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/FlowModule.scala
@@ -5,11 +5,12 @@ package akka.stream.impl
 
 import akka.stream._
 import akka.stream.impl.StreamLayout.Module
+import akka.event.Logging
 
 /**
  * INTERNAL API
  */
-private[stream] trait FlowModule[In, Out, Mat] extends StreamLayout.Module {
+private[stream] trait FlowModule[In, Out, Mat] extends StreamLayout.AtomicModule {
   override def replaceShape(s: Shape) =
     if (s != shape) throw new UnsupportedOperationException("cannot replace the shape of a FlowModule")
     else this
@@ -18,6 +19,6 @@ private[stream] trait FlowModule[In, Out, Mat] extends StreamLayout.Module {
   val outPort = Outlet[Out]("Flow.out")
   override val shape = new FlowShape(inPort, outPort)
 
-  override def subModules: Set[Module] = Set.empty
+  protected def label: String = Logging.simpleName(this)
+  final override def toString: String = f"$label [${System.identityHashCode(this)}%08x]"
 }
-
--- a/akka-stream/src/main/scala/akka/stream/impl/Modules.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/Modules.scala
@@ -6,29 +6,30 @@ package akka.stream.impl
 import akka.NotUsed
 import akka.actor._
 import akka.stream._
-import akka.stream.impl.StreamLayout.Module
+import akka.stream.impl.StreamLayout.AtomicModule
 import org.reactivestreams._
-
 import scala.annotation.unchecked.uncheckedVariance
 import scala.concurrent.Promise
+import akka.event.Logging
 
 /**
  * INTERNAL API
  */
-private[akka] abstract class SourceModule[+Out, +Mat](val shape: SourceShape[Out]) extends Module {
+private[akka] abstract class SourceModule[+Out, +Mat](val shape: SourceShape[Out]) extends AtomicModule {
+
+  protected def label: String = Logging.simpleName(this)
+  final override def toString: String = f"$label [${System.identityHashCode(this)}%08x]"
 
   def create(context: MaterializationContext): (Publisher[Out] @uncheckedVariance, Mat)
 
-  override def replaceShape(s: Shape): Module =
+  override def replaceShape(s: Shape): AtomicModule =
     if (s != shape) throw new UnsupportedOperationException("cannot replace the shape of a Source, you need to wrap it in a Graph for that")
     else this
 
   // This is okay since the only caller of this method is right below.
   protected def newInstance(shape: SourceShape[Out] @uncheckedVariance): SourceModule[Out, Mat]
 
-  override def carbonCopy: Module = newInstance(SourceShape(shape.out.carbonCopy()))
-
-  override def subModules: Set[Module] = Set.empty
+  override def carbonCopy: AtomicModule = newInstance(SourceShape(shape.out.carbonCopy()))
 
   protected def amendShape(attr: Attributes): SourceShape[Out] = {
     val thisN = attributes.nameOrDefault(null)
@@ -52,7 +53,7 @@ private[akka] final class SubscriberSource[Out](val attributes: Attributes, shap
   }
 
   override protected def newInstance(shape: SourceShape[Out]): SourceModule[Out, Subscriber[Out]] = new SubscriberSource[Out](attributes, shape)
-  override def withAttributes(attr: Attributes): Module = new SubscriberSource[Out](attr, amendShape(attr))
+  override def withAttributes(attr: Attributes): AtomicModule = new SubscriberSource[Out](attr, amendShape(attr))
 }
 
 /**
@@ -63,22 +64,26 @@ private[akka] final class SubscriberSource[Out](val attributes: Attributes, shap
  * back-pressure upstream.
  */
 private[akka] final class PublisherSource[Out](p: Publisher[Out], val attributes: Attributes, shape: SourceShape[Out]) extends SourceModule[Out, NotUsed](shape) {
+
+  override protected def label: String = s"PublisherSource($p)"
+
   override def create(context: MaterializationContext) = (p, NotUsed)
 
   override protected def newInstance(shape: SourceShape[Out]): SourceModule[Out, NotUsed] = new PublisherSource[Out](p, attributes, shape)
-  override def withAttributes(attr: Attributes): Module = new PublisherSource[Out](p, attr, amendShape(attr))
+  override def withAttributes(attr: Attributes): AtomicModule = new PublisherSource[Out](p, attr, amendShape(attr))
 }
 
 /**
  * INTERNAL API
  */
 private[akka] final class MaybeSource[Out](val attributes: Attributes, shape: SourceShape[Out]) extends SourceModule[Out, Promise[Option[Out]]](shape) {
+
   override def create(context: MaterializationContext) = {
     val p = Promise[Option[Out]]()
     new MaybePublisher[Out](p, attributes.nameOrDefault("MaybeSource"))(context.materializer.executionContext) → p
   }
   override protected def newInstance(shape: SourceShape[Out]): SourceModule[Out, Promise[Option[Out]]] = new MaybeSource[Out](attributes, shape)
-  override def withAttributes(attr: Attributes): Module = new MaybeSource(attr, amendShape(attr))
+  override def withAttributes(attr: Attributes): AtomicModule = new MaybeSource(attr, amendShape(attr))
 }
 
 /**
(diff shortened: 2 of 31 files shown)


akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
```diff
@@ -38,7 +40,14 @@ final class Flow[-In, +Out, +Mat](private[stream] override val module: Module)
...
   private[stream] def isIdentity: Boolean = this.module eq GraphStages.Identity.module
 
-  override def via[T, Mat2](flow: Graph[FlowShape[Out, T], Mat2]): Repr[T] = viaMat(flow)(Keep.left)
+  override def via[T, Mat2](flow: Graph[FlowShape[Out, T], Mat2]): Repr[T] =
+    if (this.isIdentity) {
+      import Predef.Map.empty
+      import StreamLayout.{ CompositeModule, Ignore, IgnorableMatValComp, Transform, Atomic }
+      val m = flow.module
+      val mat = if (IgnorableMatValComp(m)) Ignore else Transform(_ => NotUsed, Atomic(m))
+      new Flow(CompositeModule(Set(m), m.shape, empty, empty, mat, Attributes.none))
```
rkuhn: This is another optimization: if we always delegate to `viaMat` this will always result in a computation node to be added, which will prevent the computation from being ignored, even if it could be.

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -247,13 +260,24 @@ object StreamLayout {
...
+        {
+          val comp =
+            if (f == scaladsl.Keep.left) {
+              if (IgnorableMatValComp(matComputation2)) matComputation1 else null
+            } else if (f == scaladsl.Keep.right) {
+              if (IgnorableMatValComp(matComputation1)) matComputation2 else null
+            } else null
+          if (comp == null) Combine(f.asInstanceOf[(Any, Any) ⇒ Any], matComputation1, matComputation2)
+          else comp
+        }
```
rkuhn: This is the main change.

hepin1989: `matComputation1/2` is not a good name :).

rkuhn: Hmm, now that you mention it, something with left/right would probably be better, indeed.

hepin1989: this function will not help the user defined `Keep.left` & `Keep.right`,maybe should metion it on doc.like we do special optimize when you use `Keep.left` & `Keep.right`,so you should not write your own ones.

akka-stream/src/main/scala/akka/stream/impl/FlowModule.scala
```diff
@@ -18,6 +18,6 @@ private[stream] trait FlowModule[In, Out, Mat] extends StreamLayout.Module {
   val outPort = Outlet[Out]("Flow.out")
   override val shape = new FlowShape(inPort, outPort)
 
-  override def subModules: Set[Module] = Set.empty
+  protected def label: String
```
viktorklang: A teensy Scaladoc comment explaining what it is used for?

akka-stream/src/main/scala/akka/stream/impl/Modules.scala
```diff
@@ -46,13 +47,15 @@ private[akka] abstract class SourceModule[+Out, +Mat](val shape: SourceShape[Out
  */
 private[akka] final class SubscriberSource[Out](val attributes: Attributes, shape: SourceShape[Out]) extends SourceModule[Out, Subscriber[Out]](shape) {
 
+  override protected def label: String = "SubscriberSource"
```
viktorklang: Is there a reason why `Logging.simpleName(this)` wouldn't be the canonical implementation hereof?

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
```diff
@@ -53,6 +51,10 @@ private[akka] abstract class SinkModule[-In, Mat](val shape: SinkShape[In]) exte
     if ((thatN eq null) || thisN == thatN) shape
     else shape.copy(in = Inlet(thatN + ".in"))
   }
+
+  protected def label: String
+  final override def toString: String = f"$label [${System.identityHashCode(this)}%08x]"
```
viktorklang: @bantonsson doesn't identityHashCode cause inflation of the object header to store the current address of the pointer to the object?

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
viktorklang: +1 @hepin1989

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
```diff
@@ -718,6 +753,8 @@ private[stream] abstract class MaterializerSession(val topLevel: StreamLayout.Mo
     new ju.HashMap[InPort, AnyRef] :: Nil
   private var publishersStack: List[ju.Map[OutPort, Publisher[Any]]] =
     new ju.HashMap[OutPort, Publisher[Any]] :: Nil
+  private var matValSrcStack: List[ju.Map[MaterializedValueNode, List[MaterializedValueSource[Any]]]] =
+    new ju.HashMap[MaterializedValueNode, List[MaterializedValueSource[Any]]] :: Nil
```
viktorklang: Please note that an empty j.u.HashMap of default capacity can occupy 3.5KB

Source: http://java-performance.info/memory-consumption-of-java-data-types-2/

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
```diff
@@ -269,6 +254,21 @@ object GraphStages {
...
 
+  private object TickSource {
+    class TickSourceCancellable(cancelled: AtomicBoolean) extends Cancellable {
+      private val cancelPromise = Promise[Done]()
+
+      def cancelFuture: Future[Done] = cancelPromise.future
+
+      override def cancel(): Boolean = {
+        if (!isCancelled) cancelPromise.trySuccess(Done)
+        true
```
viktorklang: Isn't this a bug?

akka-stream/src/main/scala/akka/stream/javadsl/package.scala
```diff
@@ -6,6 +6,8 @@ package akka.stream
 package object javadsl {
   def combinerToScala[M1, M2, M](f: akka.japi.function.Function2[M1, M2, M]): (M1, M2) ⇒ M =
     f match {
+      case x if x eq Keep.left   ⇒ scaladsl.Keep.left.asInstanceOf[(M1, M2) ⇒ M]
```
viktorklang: doesn't: `case Keep.`left` => ` work?

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
viktorklang: WAit, why wouldn't viaMat simply look at the combiner and see Keep.left and simply go the quick path?

viktorklang: (i.e. I don't like how this complexity creeps upwards)

akka-stream/src/main/scala/akka/stream/impl/Sinks.scala
bantonsson: The inflation bit was a JRockit trick that happened if the object was moved during GC after identity hash code. AFAIK using identity hash code on HotSPot will only prevent biased locking from being used. Which could be bad enough depending on the use case.

viktorklang: @bantonsson Thanks for verifying. identityHashCode gives me the shivers.

rkuhn: While I cannot comment on your bodily reactions I can testify that without including the object identity in the string representation it would be utterly impossible to debug materialization problems. Also, calling toString on graphs is not something people will normally do routinely, but if it they do the effect will be non-existent since we do not use the object monitors at all.

viktorklang: Every time you create a Graph in the REPL? ;-)

akka-stream/src/main/scala/akka/stream/scaladsl/Flow.scala
rkuhn: I see your point; folding this logic into `viaMat` makes that one quite a bit more complex, but it seems better overall.

akka-stream/src/main/scala/akka/stream/impl/Modules.scala
rkuhn: right you are: I expected more variance

akka-stream/src/main/scala/akka/stream/javadsl/package.scala
rkuhn: nope, not a stable identifier (and I don’t want to make the constants public)

akka-stream/src/main/scala/akka/stream/impl/StreamLayout.scala
rkuhn: That’s not default capacity (which is 16) but capacity 100.

akka-stream/src/main/scala/akka/stream/impl/FlowModule.scala
rkuhn: Well, sure could be done, but the only usage is directly on the line below …

akka-stream/src/main/scala/akka/stream/javadsl/package.scala
viktorklang: In theory, would http://www.scala-lang.org/api/current/#scala.annotation.unchecked.uncheckedStable do the trick?

rkuhn: Hmm, looks useful, but I cannot figure out how to make it work.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
rkuhn: Hmm, I don’t know, I just moved this down a few lines … The cancellation logic is quite complex, perhaps more than is healthy.

akka-stream/src/main/scala/akka/stream/javadsl/package.scala
viktorklang: crap :S

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphStages.scala
viktorklang: If we can't tell, then it definitely is :)

rkuhn: filed #20028

patriknw:
LGTM

ktoso:
This PR also fixes https://github.com/akka/akka/issues/19924

akka-stream/src/main/scala/akka/stream/javadsl/package.scala
ktoso: Interesting, I never knew about that one.
Seems though that it only is about "stable type", and not generally speaking "assume this is stable": http://www.scala-lang.org/files/archive/spec/2.11/11-annotations.html#scala-compiler-annotations

(wow so nice we actually can link to spec like that nowadays (regardless of its state, I know I know))

patriknw:
anything remaining here, can we rebase and merge?

rkuhn:
I’ll look into the test failure and rebase after finishing my current bugfix.

Issue #20024, opened 2016-03-14
Throw more specific exception for "Unexpected (early) disconnect"


ktoso:
We're currently throwing RuntimeException in PoolSlot, would be nicer to throw a custom Akka HTTP exception.

PR #20025, opened 2016-03-14
=htc #20024 introduce more specific exception type for connection close

ktoso:
Resolves https://github.com/akka/akka/issues/20024

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
@@ -177,13 +177,13 @@ private object PoolSlot {
       val results: List[ProcessorOut] = {
         if (inflightRequests.isEmpty && firstContext.isDefined) {
           (error match {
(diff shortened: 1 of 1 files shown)


johanandren:
LGTM

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolSlot.scala
```diff
@@ -237,4 +237,10 @@ private object PoolSlot {
         context.stop(self)
     }
   }
+
+  final class UnexpectedDisconnectException(msg: String, cause: Throwable) extends RuntimeException(msg, cause) {
+    def this(msg: String) {
```
patriknw: I think that we more often use this syntax:
```
def this(msg: String) = this(msg, null)
```

ktoso: yep, will use that. I so rarely write secondary constructors that I'm always confused about syntax of them ;)

patriknw:
LGTM, aside from the nitpick

Issue #19780, opened 2016-02-15
Cluster/Remoting: ACKs misaligned (?)


ktoso:
Core issue:

```
February 12th 2016, 04:00:59.000	akka.remote.Remoting	ip-10-0-200-7	ERROR	cluster	Association to [akka.tcp://application@xx.xx.xx.xx:2551] with UID [656067421] irrecoverably failed. Quarantining address. java.lang.IllegalStateException: Error encountered while processing system message acknowledgement buffer: [-1 {}] ack: ACK[7, {}]t
	at akka.remote.ReliableDeliverySupervisor$$anonfun$receive$1.applyOrElse(Endpoint.scala:288) ~[com.typesafe.akka.akka-remote_2.11-2.4.1.jar:na]t
	at akka.actor.Actor$class.aroundReceive(Actor.scala:480) ~[com.typesafe.akka.akka-actor_2.11-2.4.1.jar:na]t	at akka.remote.ReliableDeliverySupervisor.aroundReceive(Endpoint.scala:189) ~[com.typesafe.akka.akka-remote_2.11-2.4.1.jar:na]t
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:526) [com.typesafe.akka.akka-actor_2.11-2.4.1.jar:na]t
	at akka.actor.ActorCell.invoke(ActorCell.scala:495) [com.typesafe.akka.akka-actor_2.11-2.4.1.jar:na]t
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:257) [com.typesafe.akka.akka-actor_2.11-2.4.1.jar:na]t
	at akka.dispatch.Mailbox.run(Mailbox.scala:224) [com.typesafe.akka.akka-actor_2.11-2.4.1.jar:na]t
	at akka.dispatch.Mailbox.exec(Mailbox.scala:234) [com.typesafe.akka.akka-actor_2.11-2.4.1.jar:na]t
	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260) [org.scala-lang.scala-library-2.11.7.jar:na]t
	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339) [org.scala-lang.scala-library-2.11.7.jar:na]t
	at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979) [org.scala-lang.scala-library-2.11.7.jar:na]t
[...]

PR #20033, opened 2016-03-15
=rem #19780 discard in-flight ACKs for wrong system

patriknw:
- fixes issues noticed as
  "Error encountered while processing system message acknowledgement buffer: [-1 {}] ack: ACK[0, {}]"
  when restarting actor system
- include the origin uid in the ACK and discard received ACKs that
  were targed to an old incarnation of the system with same host:port

### DIFF
--- a/akka-cluster/src/multi-jvm/scala/akka/cluster/RestartNodeSpec.scala
+++ b/akka-cluster/src/multi-jvm/scala/akka/cluster/RestartNodeSpec.scala
@@ -3,21 +3,25 @@
  */
 package akka.cluster
 
-import language.postfixOps
 import scala.collection.immutable
-import com.typesafe.config.ConfigFactory
-import org.scalatest.BeforeAndAfter
+import scala.concurrent.duration._
+
+import akka.Done
+import akka.actor.Actor
+import akka.actor.ActorIdentity
+import akka.actor.ActorRef
+import akka.actor.ActorSystem
+import akka.actor.Address
+import akka.actor.Deploy
+import akka.actor.Identify
+import akka.actor.Props
+import akka.actor.RootActorPath
+import akka.actor.Terminated
+import akka.cluster.MemberStatus._
 import akka.remote.testkit.MultiNodeConfig
 import akka.remote.testkit.MultiNodeSpec
 import akka.testkit._
-import scala.concurrent.duration._
-import akka.actor.Address
-import akka.actor.ActorSystem
-import akka.actor.Props
-import akka.actor.Actor
-import akka.actor.RootActorPath
-import akka.cluster.MemberStatus._
-import akka.actor.Deploy
+import com.typesafe.config.ConfigFactory
 
 object RestartNodeMultiJvmSpec extends MultiNodeConfig {
   val first = role("first")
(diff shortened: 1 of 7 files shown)


patriknw:
Refs https://github.com/akka/akka/issues/19780

akka-cluster/src/multi-jvm/scala/akka/cluster/RestartNodeSpec.scala
```diff
@@ -27,6 +31,22 @@ object RestartNodeMultiJvmSpec extends MultiNodeConfig {
   commonConfig(debugConfig(on = false).
     withFallback(ConfigFactory.parseString("akka.cluster.auto-down-unreachable-after = 5s")).
     withFallback(MultiNodeClusterSpec.clusterConfig))
+
+  /**
+   * This was used together with sleep in EndpointReader before deliverAndAck
+   * to reproduce issue with misaligned ACKs when restarting system,
```
ktoso: Thanks a lot for this comment, I see where that is finally

rkuhn:
LGTM

rkuhn:
Great detective work, @patriknw !

ktoso:
LGTM, great work @patriknw!

patriknw:
let me know if you have any more questions and I can try to explain in a hangout?

johanandren:
Will this for break it for people doing rolling upgrades of their cluster (I know we don't promise to not break)?

johanandren:
NVM saw the comment now!

patriknw:
that's why I added the `originUid` field as optional, so if talking to old system it will be as before

johanandren:
LGTM

ktoso:
Needs a rebase now

drewhk:
LGTM, but why should the ACK carry the UID? Shouldn't be the UID available after the handshake finished? I.e. you should probably just drop the ACKs until the UID of the remote becomes known again? (the ACKs will be resent anyway).

patriknw:
@drewhk the problem was that we got an old ACK (arrived after `GotUid`) that was targeted for the previous incarnation of the system. Can be reproduced as described in the [issue](https://github.com/akka/akka/issues/19780#issuecomment-196280119)

drewhk:
Hm, but how could that happen? I saw the reproduction steps, but the Ack arrives on a TCP connection, so if it arrives from an old connection that means that stopping reading from the previous connection has not been done? I ask this because the underlying issue can make other messages to arrive interleaved with the new ones, not just an Ack.

patriknw:
Good question. I can try to debug it more, since I have a reproducer. It's not that easy to follow what is going on here, as we know.

drewhk:
I can take a look, if you want. With a reproducer it should be at least easier.

patriknw:
I'll like to do it, at least until I'm stuck, thanks

drewhk:
great, I won't complain :D

patriknw:
@drewhk I found the reason and have a better solution. This PR is superseded by https://github.com/akka/akka/pull/20093

drewhk:
Great, you are my hero :)

patriknw:
thanks for pushing me in the right direction

PR #20034, opened 2016-03-15
fix endless loop race condition in NodeQueue, #19949

### LINKED ISSUE #19949: Smallest mailbox router pool occasionally eating 100% CPU
I originally posted this on the Akka User groups page here:

https://groups.google.com/forum/#!topic/akka-user/2vA2npqq2aU

It appears that sometimes the smallest mailbox routing logic can get stuck in a do/while loop in `AbstractNodeQueue.peekNode`.  This is sporadic in nature, but if I run enough volume through my server I can eventually get it to do this.  I did not see this in the 2.3.x series, and only noticed it after we upgraded to 2.4.2.  The stack of the stuck thread looks like this:

> "AqutoServices-akka.actor.default-dispatcher-17" #32 prio=5 os_prio=0 tid=0x00007f404c016000 nid=0x37d1 runnable [0x00007f406445b000]
   java.lang.Thread.State: RUNNABLE
	at akka.dispatch.AbstractNodeQueue.peekNode(AbstractNodeQueue.java:42)
	at akka.dispatch.AbstractNodeQueue.count(AbstractNodeQueue.java:70)
	at akka.dispatch.NodeMessageQueue.numberOfMessages(Mailbox.scala:376)
	at akka.dispatch.Mailbox.numberOfMessages(Mailbox.scala:102)
	at akka.actor.dungeon.Dispatch$class.numberOfMessages(Dispatch.scala:34)
	at akka.actor.ActorCell.numberOfMessages(ActorCell.scala:374)
	at akka.routing.SmallestMailboxRoutingLogic.numberOfMessages(SmallestMailbox.scala:132)
	at akka.routing.SmallestMailboxRoutingLogic.selectNext(SmallestMailbox.scala:67)
	at akka.routing.SmallestMailboxRoutingLogic.select(SmallestMailbox.scala:37)
	at akka.routing.Router.route(Router.scala:116)

I'm going to try and switch those actors over to a round robin router pool for now.  I would appreciate any insight you guys might have on this.  Viktor seems to think that it might have to do with: https://github.com/akka/akka/issues/19216

rkuhn:
- also fixed some forgotten copy-pasta between peekNode and pollNode
- also added JavaDoc to all methods, explaining which can be used from
  what thread
- did not fix the JDK8 improvement of using Unsafe instead of inheriting
  from AtomicReference since that inheritance is not a bad thing,
  actually

fixes #19949

### DIFF
--- a/akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
+++ b/akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
@@ -11,11 +11,20 @@ import java.util.concurrent.atomic.AtomicReference;
 /**
  * Lock-free MPSC linked queue implementation based on Dmitriy Vyukov's non-intrusive MPSC queue:
  * http://www.1024cores.net/home/lock-free-algorithms/queues/non-intrusive-mpsc-node-based-queue
+ * 
+ * This queue could be wait-free (i.e. without the spinning loops in peekNode and pollNode) if
+ * it were permitted to return null while the queue is not quite empty anymore but the enqueued
+ * element is not yet visible. This would break actor scheduling, though.
  */
 @SuppressWarnings("serial")
 public abstract class AbstractNodeQueue<T> extends AtomicReference<AbstractNodeQueue.Node<T>> {
-    // Extends AtomicReference for the "head" slot (which is the one that is appended to) since
-    // Unsafe does not expose XCHG operation intrinsically before JDK 8
+
+    /*
+     * Extends AtomicReference for the "head" slot (which is the one that is appended to) since
+     * there is nothing to be gained by going to all-out Unsafe usage—we’d have to do
+     * cache-line padding ourselves.
+     */
+  
     @SuppressWarnings("unused")
     private volatile Node<T> _tailDoNotCallMeDirectly;
 
@@ -25,10 +34,14 @@ public abstract class AbstractNodeQueue<T> extends AtomicReference<AbstractNodeQ
        set(n);
     }
 
-    /*
+    /**
+     * Query the queue tail for the next element without dequeuing it.
+     * 
      * Use this method only from the consumer thread!
      * 
      * !!! There is a copy of this code in pollNode() !!!
+     * 
+     * @return queue node with element inside if there was one, or null if there was none
      */
     @SuppressWarnings("unchecked")
     protected final Node<T> peekNode() {
@@ -43,37 +56,83 @@ public abstract class AbstractNodeQueue<T> extends AtomicReference<AbstractNodeQ
         }
         return next;
     }
-    /*
+    
+    /**
+     * Query the queue tail for the next element without dequeuing it.
+     * 
      * Use this method only from the consumer thread!
+     * 
+     * @return element if there was one, or null if there was none
      */
     public final T peek() {
         final Node<T> n = peekNode();
         return (n != null) ? n.value : null;
     }
 
+    /**
+     * Add an element to the head of the queue.
+     * 
+     * This method can be used from any thread.
+     * 
+     * @param value the element to be added; must not be null
+     */
     public final void add(final T value) {
         final Node<T> n = new Node<T>(value);
         getAndSet(n).setNext(n);
     }
     
+    /**
+     * Add an element to the head of the queue, providing the queue node to be used.
+     * 
+     * This method can be used from any thread.
+     * 
+     * @param n the node containing the element to be added; both must not be null
+     */
     public final void addNode(final Node<T> n) {
         n.setNext(null);
         getAndSet(n).setNext(n);
     }
 
+    /**
+     * Query the queue whether it is empty right now.
+     * 
+     * This method can be used from any thread.
+     * 
+     * @return true if queue was empty at some point in the past
+     */
     public final boolean isEmpty() {
         return Unsafe.instance.getObjectVolatile(this, tailOffset) == get();
     }
 
+    /**
+     * This method returns an upper bound on the queue size at the time it
+     * starts executing. It may spuriously return smaller values (including
+     * zero) if the consumer pulls items out concurrently.
+     * 
+     * This method can be used from any thread.
+     * 
+     * @return an upper bound on queue length at some time in the past
+     */
+    @SuppressWarnings("unchecked")
     public final int count() {
         int count = 0;
-        for(Node<T> n = peekNode();n != null && count < Integer.MAX_VALUE; n = n.next())
+        final Node<T> head = get();
+        for(Node<T> n = ((Node<T>) Unsafe.instance.getObjectVolatile(this, tailOffset)).next();
+            n != null && count < Integer.MAX_VALUE; 
+            n = n.next()) {
           ++count;
+          // only iterate up to the point where head was when starting: this is a moving queue!
+          if (n == head) break;
+        }
         return count;
     }
 
-    /*
+    /**
+     * Pull one item from the queue’s tail if there is one.
+     * 
      * Use this method only from the consumer thread!
+     * 
+     * @return element if there was one, or null if there was none
      */
     public final T poll() {
         final Node<T> next = pollNode();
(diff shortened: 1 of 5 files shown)


akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
```diff
@@ -619,8 +619,11 @@ object UnboundedMailbox {
 
 /**
  * SingleConsumerOnlyUnboundedMailbox is a high-performance, multiple producer—single consumer, unbounded MailboxType,
- * the only drawback is that you can't have multiple consumers,
+ * with the drawback that you can't have multiple consumers,
  * which rules out using it with BalancingPool (BalancingDispatcher) for instance.
+ *
+ * Currently this queue is slower for some benchmarks than the ConcurrentLinkedQueue from JDK 8 that is used by default,
+ * so be sure to measure the performance in your particular setting in order to determine which one to use.
```
patriknw: should we add this to `mailboxes.rst` also?

akka-bench-jmh/src/main/scala/akka/dispatch/NodeQueueBenchmark.scala
```diff
@@ -0,0 +1,72 @@
...
+  def teardown(): Unit = Await.result(sys.terminate(), 5.seconds)
+
+  @TearDown(Level.Invocation)
+  def waitInBetween(): Unit = {
+    val probe = TestProbe()
+    probe.send(ref, Stop)
+    probe.expectMsg(Stop)
+    System.gc()
+    System.gc()
+    System.gc()
```
patriknw: should such things be done in JMH benchmarks?

patriknw:
please review @bantonsson @MQ-EL

patriknw:
LGTM, I leave the review of the queue changes to those how are more into this queue

patriknw:
When reviewed this should also be backported to release-2.3

akka-bench-jmh/src/main/scala/akka/dispatch/NodeQueueBenchmark.scala
bantonsson: @patriknw It's perfectly fine to try to make the heap nice between runs in the bnechmarks.

bantonsson:
LGTM

akka-bench-jmh/src/main/scala/akka/dispatch/NodeQueueBenchmark.scala
patriknw: thanks

akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
```diff
@@ -43,37 +56,83 @@ protected AbstractNodeQueue() {
...
+     * 
+     * @return an upper bound on queue length at some time in the past
+     */
+    @SuppressWarnings("unchecked")
     public final int count() {
         int count = 0;
-        for(Node<T> n = peekNode();n != null && count < Integer.MAX_VALUE; n = n.next())
+        final Node<T> head = get();
+        for(Node<T> n = ((Node<T>) Unsafe.instance.getObjectVolatile(this, tailOffset)).next();
+            n != null && count < Integer.MAX_VALUE;
```
viktorklang: How about: `n != null && count < Integer.MAX_VALUE && n != head`

akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
```diff
@@ -85,18 +144,23 @@ public final T poll() {
...
-        next = tail.next();
-        if (next != null || get() == tail)
-          break;
+      final Node<T> tail = (Node<T>) Unsafe.instance.getObjectVolatile(this, tailOffset);
+      Node<T> next = tail.next();
+      if (next == null && get() != tail) {
+          // if tail != head this is not going to change until producer makes progress
+          // we can avoid reading the head and just spin on next until it shows up
+          do {
+              next = tail.next();
```
viktorklang: Perhaps add a TODO to use http://mail.openjdk.java.net/pipermail/hotspot-compiler-dev/2016-January/020987.html here?

akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
```diff
@@ -11,11 +11,20 @@
 /**
  * Lock-free MPSC linked queue implementation based on Dmitriy Vyukov's non-intrusive MPSC queue:
  * http://www.1024cores.net/home/lock-free-algorithms/queues/non-intrusive-mpsc-node-based-queue
+ * 
+ * This queue could be wait-free (i.e. without the spinning loops in peekNode and pollNode) if
+ * it were permitted to return null while the queue is not quite empty anymore but the enqueued
+ * element is not yet visible. This would break actor scheduling, though.
  */
 @SuppressWarnings("serial")
 public abstract class AbstractNodeQueue<T> extends AtomicReference<AbstractNodeQueue.Node<T>> {
```
viktorklang: Does the bounded version suffer from the same problems?

akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
rkuhn: That results in a count that is strictly less correct :-)

viktorklang: `n != null && ++count < Integer.MAX_VALUE && n != head` :D

viktorklang: Heh, or rather:

`n != null && count++ < Integer.MAX_VALUE && n != head`

rkuhn: This ain’t the IOCCC, there are rules!

viktorklang: Dude, this is High-Performance Java—the only things which are prohibited is eye-gouging and biting. ^^

akka-bench-jmh/src/main/scala/akka/dispatch/NodeQueueBenchmark.scala
ktoso: I don't mind those here, but wonder if `-gc` would not be enough (do the same, but by JMH):

> If you wish to isolate GC effects between iterations you can use the gc option, this is often desirable to help getting more uniform results.

http://psy-lob-saw.blogspot.com/2013/04/writing-java-micro-benchmarks-with-jmh.html

rkuhn: This is between invocations, not between iterations, and that is on purpose: in the interest of somewhat stable measurements we only do things one 100Kmsg burst at a time. Otherwise the measurement is completely dominated by GC performance.

akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
rkuhn: No, the bounded version suffers from sufficiently different problems ;-)

akka-actor/src/main/scala/akka/dispatch/Mailbox.scala
rkuhn: good point, done

akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
MQ-EL: why not use ``` @Contended ```? It work fine as artificially padding. and more importantly, it performance better than CLQ from myselft test result.

akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
```diff
@@ -43,37 +56,83 @@ protected AbstractNodeQueue() {
         }
```
MQ-EL: how about ``` while ((next = tail.next()) == null) ;``` ?

akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
MQ-EL: and ```  tailOffset ``` field

rkuhn: @MQ-EL no matter what I tried, I couldn’t find a way to make this queue outperform the default CLQ in the context of sending to an actor—things might be quite different when testing the queue in isolation. If you find a way to improve actor sending performance, please open a PR!

akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
rkuhn: I prefer readability over conciseness.

akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
rkuhn: Still, your version has different semantics.

akka-actor/src/main/java/akka/dispatch/AbstractNodeQueue.java
rkuhn: Seems way too far out to put into the sources.

akka-bench-jmh/src/main/scala/akka/dispatch/NodeQueueBenchmark.scala
ktoso: I see, thanks :+1:

Issue #19322, opened 2015-12-31
Akka.Remote ReliableDeliverySupervisor bufferWasInUse should be set to false on reset(?)


Aaronontheweb:
Hi from the Akka.NET project!

I've been fixing some bugs in our own Akka.Remote implementation around adaptive backoff and buffering, and in the process I updated our system remoting actors to match what's in the master branch.

Within the `ReliableDeliverySupervisor` I noticed something that seemed odd to me and wanted to bring it up as a possible bug.

The internal field `bufferWasInUse` https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/Endpoint.scala#L207 is used to indicate whether or not anything was ever attempted to be written to the buffer (`tryBuffer`) and plays a role inside the `SupervisorStrategy` of the `ReliableDeliverySupervisor`:

```scala
override val supervisorStrategy = OneForOneStrategy(loggingEnabled = false) {
    case e @ (_: AssociationProblem) ⇒ Escalate
    case NonFatal(e) ⇒
      val causedBy = if (e.getCause == null) "" else s"Caused by: [${e.getCause.getMessage}]"
      log.warning("Association with remote system [{}] has failed, address is now gated for [{}] ms. Reason: [{}] {}",
        remoteAddress, settings.RetryGateClosedFor.toMillis, e.getMessage, causedBy)
      uidConfirmed = false // Need confirmation of UID again
      if (bufferWasInUse) {
        if ((resendBuffer.nacked.nonEmpty || resendBuffer.nonAcked.nonEmpty) && bailoutAt.isEmpty)
          bailoutAt = Some(Deadline.now + settings.InitialSysMsgDeliveryTimeout)
        context.become(gated(writerTerminated = false, earlyUngateRequested = false))
        currentHandle = None
        context.parent ! StoppedReading(self)
        Stop
      } else Escalate
  }
```

I noticed that inside the `reset` method, which gets called if and only if the node we're associated with comes back with a different UID than our previous association, doesn't reset the value of this field back to false even though the buffer itself gets discarded and recreated. Couldn't this have potential unintended "dirtying" side-effects between associations with that remote system?

patriknw:
Good observation. I think it makes sense to clear `bufferWasInUse` together with the `buffer` reset.
cc @drewhk

Aaronontheweb:
Excellent, thanks - I'll see that we do this in the .NET implementation too.

patriknw:
fixed in master and release-2.3

PR #20036, opened 2016-03-15
=rem #19322 reset bufferWasInUse

patriknw:
Refs #19322


johanandren:
LGTM

akka-remote/src/main/scala/akka/remote/Endpoint.scala
```diff
@@ -229,6 +229,7 @@ private[remote] class ReliableDeliverySupervisor(
 
   def reset() {
```
viktorklang: : Unit = // Boy Scout

patriknw: fixed, thanks

PR #20040, opened 2016-03-15
=str #20039 incorrect FixedBufferSpec test case fixed

### LINKED ISSUE #20039: "become non-full after head dropped from full buffer" from the FixedBufferSpec's checks wrong behavior
I'm currently porting these tests to Akka.net and found a copy paste issue. 

The  [become non-full after head dropped from full buffer](https://github.com/akka/akka/blob/master/akka-stream-tests/src/test/scala/akka/stream/impl/FixedBufferSpec.scala#L76) spec doesn't check dropHead but rather dropTail from the previous spec.

johanandren:
Fixes #20039


patriknw:
LGTM

ktoso:
LGTM

PR #20041, opened 2016-03-15
fix ActorPublisher state machine, fixes #20031

### LINKED ISSUE #20031: FileIO publisher invokes onComplete twice
Akka version: 2.4.2
OS: Ubuntu 15.10

`FileIO` violates section 1.7 of the reactive streams spec, *Once a terminal state has been signaled (onError, onComplete) it is REQUIRED that no further signals occur.*

Reproducer:

```scala
import akka.actor._
import akka.util.ByteString
import akka.stream._
import akka.stream.scaladsl._
import org.reactivestreams._

implicit val mat = ActorMaterializer()
val file = new java.io.File(/* insert any existing file here */ "/home/jroper/bike.jpg")
val subscriber = new Subscriber[ByteString] {
  var s: Subscription = _
  def onSubscribe(s: Subscription) = {
    println("onSubscribe")
    this.s = s
    s.request(1)
  }
  def onNext(b: ByteString) = s.request(1)
  def onError(t: Throwable) = println(t)
  def onComplete = println("onComplete")
}

FileIO.fromFile(file).runWith(Sink.fromSubscriber(subscriber))
```

Running it:

```
scala> FileIO.fromFile(file).runWith(Sink.fromSubscriber(subscriber))
res0: akka.NotUsed = NotUsed

scala> onSubscribe
onComplete
onComplete
```

rkuhn:
ref #20031

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
@@ -28,8 +28,8 @@ private[akka] object InputStreamPublisher {
 
 /** INTERNAL API */
 private[akka] class InputStreamPublisher(is: InputStream, completionPromise: Promise[IOResult], chunkSize: Int)
-  extends akka.stream.actor.ActorPublisher[ByteString]
-  with ActorLogging {
+    extends akka.stream.actor.ActorPublisher[ByteString]
+    with ActorLogging {
 
   // TODO possibly de-duplicate with FilePublisher?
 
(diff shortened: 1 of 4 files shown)


akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
```diff
@@ -47,7 +47,7 @@ private[akka] class InputStreamPublisher(is: InputStream, completionPromise: Pro
   def readAndSignal(): Unit =
     if (isActive) {
       readAndEmit()
-      if (totalDemand > 0) self ! Continue
+      if (totalDemand > 0 && isActive) self ! Continue
```
ktoso: Ah, not forgetting that one is tricky as hell

rkuhn: to be fair: this only gets rid of a dead letter, but people don’t really love those somehow …

ktoso:
LGTM AFAICS

akka-stream/src/main/scala/akka/stream/impl/io/InputStreamPublisher.scala
viktorklang: I think it's a name-thing

patriknw:
LGTM

Issue #18500, opened 2015-09-17
Fix scaladoc warnings (release-2.3-dev)


drewhk:
Mostly link problems

drewhk:
There are 150 warnings

agolubev:
@drewhk, as far as I understand it's about 
>[warn] /localhome/jenkinsakka/workspace/pr-validator-per-commit-jenkins/akka-stream/target/java/akka/stream/io/Implicits.java:20: warning - Tag @link: reference not found: Source
and this must be fixed by adding package like: [[Source]] -> [[akka.stream.scaladsl.Source]]

wlk:
There is also a number of warnings that appear when running `sbt doc` locally that are not logged on jenkins, for example:
```
[warn] /home/w/projects/akka/akka-actor/src/main/scala/akka/io/Inet.scala:120: Could not find any member to link for "java.net.Socket#setSendBufferSize".
```
This error basically means that it's unable to find Java class, and in the result there is no clickable link.

Or something like that:
```
[warn] /home/w/projects/akka/akka-actor/src/main/scala/akka/actor/ActorSystem.scala:290: The link target "ActorSystem#actorSelection" is ambiguous. Several members fit the target:
[warn] (path: akka.actor.ActorPath): akka.actor.ActorSelection in class ActorSystem [chosen]
[warn] (path: String): akka.actor.ActorSelection in class ActorSystem
```

In this case the link is clickable, but can lead to wrong version of the method.

wlk:
I'll start off with fixing what is reported in jenkins, and will move my previous comment to new issue

wlk:
I'm fixing those warnings in PR #20000

wlk:
Small update: I have went through the first iteration of fixing link problems.

But after a while I have noticed that warnings reported by jenkins are actually Javadoc warnings, not Scaladoc warnings

I'm preparing another PR where I fix the rest of warnings.

ktoso:
Right :)
Thanks a lot (you know how to get javadocs right? via genjavadoc)

ktoso:
I'm not sure if all is fixed, so keeping open still.

rkuhn:
I’d vote for adding a ticket that makes ScalaDoc or JavaDoc warnings fail PR validation—and then getting that across the PR merge process :-) And I think that this particular ticket should be closed.

/cc @patriknw

patriknw:
I'm all for enforcing no warnings if would be possible to tell that some should be ignored.

[...]

PR #20043, opened 2016-03-15
=doc fix scaladoc warnings #18500

wlk:
Another iteration of fixing Scala/Javadoc warnings, addresses #18500

### DIFF
--- a/akka-http/src/main/scala/akka/http/javadsl/server/RejectionHandler.scala
+++ b/akka-http/src/main/scala/akka/http/javadsl/server/RejectionHandler.scala
@@ -134,7 +134,7 @@ abstract class RejectionHandler {
   def handleUnacceptedResponseEncodingRejection(ctx: RequestContext, supported: jl.Iterable[HttpEncoding]): RouteResult = passRejection()
 
   /**
-   * Callback called to handle rejection created by an [[akka.http.scaladsl.server.authentication.HttpAuthenticator]].
+   * Callback called to handle rejection created by an [[akka.http.javadsl.server.values.HttpBasicAuthenticator]].
    * Signals that the request was rejected because the user could not be authenticated. The reason for the rejection is
    * specified in the cause.
    *
--- a/akka-http/src/main/scala/akka/http/javadsl/server/directives/PathDirectives.scala
+++ b/akka-http/src/main/scala/akka/http/javadsl/server/directives/PathDirectives.scala
@@ -133,8 +133,7 @@ abstract class PathDirectives extends MiscDirectives {
    * }
    * }}}
    *
-   * For further information, refer to:
-   * [[http://googlewebmastercentral.blogspot.de/2010/04/to-slash-or-not-to-slash.html]]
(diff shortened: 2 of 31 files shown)


ktoso:
OK TO TEST

akka-http/src/main/scala/akka/http/javadsl/server/directives/PathDirectives.scala
```diff
@@ -134,7 +134,7 @@ abstract class PathDirectives extends MiscDirectives {
    * }}}
    *
    * For further information, refer to:
-   * [[http://googlewebmastercentral.blogspot.de/2010/04/to-slash-or-not-to-slash.html]]
+   * @see "http://googlewebmastercentral.blogspot.de/2010/04/to-slash-or-not-to-slash.html"
```
ktoso: Are you sure this results in valid scaladoc and javadoc?
I'd think this breaks scaladoc

akka-http/src/main/scala/akka/http/javadsl/server/RejectionHandler.scala
```diff
@@ -134,7 +134,7 @@ abstract class RejectionHandler {
   def handleUnacceptedResponseEncodingRejection(ctx: RequestContext, supported: jl.Iterable[HttpEncoding]): RouteResult = passRejection()
 
   /**
-   * Callback called to handle rejection created by an [[akka.http.scaladsl.server.authentication.HttpAuthenticator]].
+   * Callback called to handle rejection created by an [[akka.http.javadsl.server.values.HttpBasicAuthenticator]].
```
ktoso: good catch!

akka-http/src/main/scala/akka/http/javadsl/server/directives/PathDirectives.scala
wlk: probably should be `@see [["..."]]` checking this now

wlk:
In javadoc generated by "genjavadoc" the links are not clickable anyway, so I have just pasted them directly so at least they won't generate warnings.
In scaladoc, the exptected format is `@see [[..]]` - this makes the link clickable and not generate any warnings

ktoso:
How does the `@see` part show up in scaladoc though? AFAIR it does not parse that, do we have a `"@see"` in the text?

wlk:
this:
```
...
* For further information, refer to:
   * @see [[http://googlewebmastercentral.blogspot.de/2010/04/to-slash-or-not-to-slash.html]]
...
```

shows up as:
see

ktoso:
Hm, weird I'd expect javadoc to understand `@see`, perhaps this should be solved in genjavadoc: https://github.com/typesafehub/genjavadoc/issues/69

patriknw:
LGTM
if you are still looking for how to create external links, I think ordinary `a href` tag works both in scaladoc and javadoc

ktoso:
Failure was unrelated and seems like overall improvement, let's merge this one to keep the queue small.

ktoso:
Thanks @wlk !

Issue #20044, opened 2016-03-15
Akka website margins messed up


Falmarri:
The combination of the negative top margin of the .main container and the overflow:auto are causing the page to be right aligned instead of center aligned.

![image](https://cloud.githubusercontent.com/assets/463948/13796179/79570de2-eac2-11e5-9cbe-15026d212aa8.png)

vs

![image](https://cloud.githubusercontent.com/assets/463948/13796155/5619fdc6-eac2-11e5-85e9-a775f841afc2.png)

ktoso:
Thanks for figuring that out!
Seems like you have the change ready, could you submit a PR?

the css files are here: https://github.com/akka/akka/blob/master/akka-docs/_sphinx/themes/akka/static/docs.css

Falmarri:
@ktoso I'll submit a PR, but this seems to be a chromium specific bug. -18px works on firefox. However -17px shouldn't break anything else

ktoso:
Right, chromium is usually a bit iffy to judge by but such small tweak should be OK.

ktoso:
Thanks again @Falmarri !

PR #20045, opened 2016-03-15
Fix page layout issue on chromium/linux

Falmarri:
https://github.com/akka/akka/issues/20044

Edit: Underlying bug reported here https://bugs.chromium.org/p/chromium/issues/detail?id=595166#c5

### DIFF
--- a/akka-docs/_sphinx/themes/akka/static/docs.css
+++ b/akka-docs/_sphinx/themes/akka/static/docs.css
@@ -3,7 +3,7 @@ a { color: #15A9CE; }
 a:hover { color: #15A9CE; text-decoration: underline; }
 .navbar { margin-bottom: 18px; }
 .navbar-logo { visibility: visible; float: left; padding-top: 0px; }
-.main { position: relative; height: auto; margin-top: -18px; overflow: auto; background: #15A9CE;}
+.main { position: relative; height: auto; margin-top: -18px; overflow: visible; background: #15A9CE;}
 .page-title { position: relative; top: 24px; font-family: 'Source Sans Pro', sans-serif; font-size: 24px; font-weight: 400; color: rgba(255, 255, 255, 1); width: 840px;}
 .main-container { background: #ffffff;  min-height: 600px; padding-top: 20px; margin-top: 28px; }
 .pdf-link { float: right; height: 40px; margin-bottom: -15px; margin-top: -5px; }


ktoso:
Thanks a lot for looking into it and the fix too!

I looked into it and my OCD kicked in slightly :cry: 
If we do that there's going to be 1px of lines not touching at the top bar:
cursor_and_scala_documentation_ _akka_documentation

I also got my linux notebook to check it over there, on Chromium 47 (I guess that's old though) it looks good still, hm. Now I'm not sure if we should apply this... maybe I'm overthinking it over that 1 px though, let's see what the team thinks.

Thanks so much for jumping in with the PR in any case!

Falmarri:
@ktoso yeah this might be a chromium issue. I hadn't seen this issue before and I use the akka docs almost daily. I did recently upgrade my system and I just noticed it today. I'm on arch linux with chromium Version 49.0.2623.87 (64-bit). I agree that 1px is annoying. I can probably fix it cleaner but the change will be slightly bigger. basically change the navbar and the main content to not use opposing margins to try to line up pixels

Falmarri:
@ktoso removing the overflow:auto seems to work in both firefox and chromium. If someone could test on windows/osx it might be as simple as that.

ktoso:
Removing overflow auto breaks the bottom bar (on firefox, on chrome it was ok):

scala_documentation_ _akka_documentation

Falmarri:
Argh.

ktoso:
I'm starting to lean towards left untouched hm... :(

Falmarri:
What firefox version/os? Without the overflow: auto it works fine on Firefox 45.0

Falmarri:
@ktoso what if you change overflow: auto to overflow: visible. That fixes it for me as well.

ktoso:
AFAICS that would work!
I checked that on firefox/osx and chrome/osx and chromium/linux, seems to work (`overflow: visible`)

patriknw:
LGTM

patriknw:
please squash to one commit

Falmarri:
@patriknw done

ktoso:
Technically commit should include ref to `#20044`, but it's a minor thing I'll merge once back home in some minutes.

Falmarri:
I'll amend my commit

ktoso:
Thanks!

Issue #20017, opened 2016-03-11
Buffers and working with rate chapter should mention fusion


Mihai238:
Hi,

According to http://doc.akka.io/docs/akka/2.4.2/java/stream/stream-rate.html akka streams runs stages asynchronously, starting to handle one element immediately after the previous one was delivered to its downstream.

Following code:

```java
public static void main(String[] args) throws InterruptedException {
        ActorSystem system = ActorSystem.create();
        ActorMaterializer mat = ActorMaterializer.create(system);

        Source.from(Arrays.asList(1, 2, 3))
                .map(i -> {System.out.println("A: " + i); return i;})
                .map(i -> {System.out.println("B: " + i); return i;})
                .map(i -> {System.out.println("C: " + i); return i;})
                .runWith(Sink.<Integer>ignore(), mat);
    }
```

produces always the following output on my computer:

```
A: 1
B: 1
C: 1
A: 2
B: 2
C: 2
A: 3
B: 3
C: 3
```

Same output is produced also if I add a `Thread.sleep(1000)` statement in the last `map` operation. There is no `application.conf` file in my classpath. Am I missing something?

Additional information:
Akka Stream version: akka-stream_2.11:2.4.2
Java Version: openjdk 1.8.0_74

drewhk:
By default stages are fused, you need to mark your stages async explicitly, see http://doc.akka.io/docs/akka/2.4.2/scala/stream/stream-flows-and-basics.html#Operator_Fusion for details.

Mihai238:
Oh, I missed that. Thank you! Nevertheless, it would be great if this information would also be stated in `Buffers and working with rate` chapter.

drewhk:
Ok, I renamed the ticket and reopened.

johanandren:
Was actually fixed in 2.4.3 but forgot to mark ticket as such.

PR #20050, opened 2016-03-16
=doc #20017 Update the stream-rate docs to current streams behavior

johanandren:
Refs #20017

### DIFF
--- a/akka-docs/rst/java/code/docs/stream/StreamBuffersRateDocTest.java
+++ b/akka-docs/rst/java/code/docs/stream/StreamBuffersRateDocTest.java
@@ -45,9 +45,9 @@ public class StreamBuffersRateDocTest extends AbstractJavaTest {
   public void demonstratePipelining() {
     //#pipelining
     Source.from(Arrays.asList(1, 2, 3))
-      .map(i -> {System.out.println("A: " + i); return i;})
-      .map(i -> {System.out.println("B: " + i); return i;})
-      .map(i -> {System.out.println("C: " + i); return i;})
+      .map(i -> {System.out.println("A: " + i); return i;}).async()
+      .map(i -> {System.out.println("B: " + i); return i;}).async()
+      .map(i -> {System.out.println("C: " + i); return i;}).async()
       .runWith(Sink.ignore(), mat);
     //#pipelining
   }
@@ -64,12 +64,12 @@ public class StreamBuffersRateDocTest extends AbstractJavaTest {
     //#section-buffer
     final Flow<Integer, Integer, NotUsed> flow1 =
       Flow.of(Integer.class)
-      .map(elem -> elem * 2) // the buffer size of this map is 1
-      .withAttributes(Attributes.inputBuffer(1, 1));
+      .map(elem -> elem * 2).async()
+      .withAttributes(Attributes.inputBuffer(1, 1)); // the buffer size of this map is 1
     final Flow<Integer, Integer, NotUsed> flow2 =
       flow1.via(
         Flow.of(Integer.class)
-        .map(elem -> elem / 2)); // the buffer size of this map is the default
+        .map(elem -> elem / 2)).async(); // the buffer size of this map is the default
     //#section-buffer
   }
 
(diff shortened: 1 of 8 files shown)


akka-docs/rst/java/code/docs/stream/StreamBuffersRateDocTest.java
```diff
@@ -64,12 +64,12 @@ public void demonstrateBufferSizes() {
     //#section-buffer
     final Flow<Integer, Integer, NotUsed> flow1 =
       Flow.of(Integer.class)
-      .map(elem -> elem * 2) // the buffer size of this map is 1
+      .map(elem -> elem * 2).async() // the buffer size of this map is 1
```
drewhk: maybe move the comment where the actual buffer definition is

akka-docs/rst/java/stream/stream-composition.rst
```diff
@@ -285,9 +285,9 @@ Attributes
 We have seen that we can use ``named()`` to introduce a nesting level in the fluid DSL (and also explicit nesting by using
 ``create()`` from :class:`GraphDSL`). Apart from having the effect of adding a nesting level, ``named()`` is actually
 a shorthand for calling ``withAttributes(Attributes.name("someName"))``. Attributes provide a way to fine-tune certain
-aspects of the materialized running entity. For example buffer sizes can be controlled via attributes (see
-:ref:`stream-buffers-scala`). When it comes to hierarchic composition, attributes are inherited by nested modules,
-unless they override them with a custom value.
+aspects of the materialized running entity. For example buffer sizes for asynchronous stagescan be controlled via
```
drewhk: stagescan/stages can

akka-docs/rst/java/stream/stream-rate.rst
```diff
@@ -141,29 +57,124 @@ Rate transformation
...
 .. includecode:: ../code/docs/stream/RateTransformationDocTest.java#expand-drift
 
-Note that all of the elements coming from upstream will go through ``expand`` at least once. This means that the output of this flow is going to report a drift of zero if producer is fast enough, or a larger drift otherwise.
+Note that all of the elements coming from upstream will go through ``expand`` at least once. This means that the
+output of this flow is going to report a drift of zero if producer is fast enough, or a larger drift otherwise.
+
+
+.. _async-stream-buffers-java:
+
+Buffers for asynchronous stages
```
drewhk: hm, why move it down here?

johanandren: My thinking: When every stage was async the internal buffers was _the_ big thing to inform about since everyone would get hit by it, everyone needed to know about it. With fusing it's now more of something to know if you introduce async stages, non-internal buffers however is useful to everyone, so more important/useful -> more prominent location.

drewhk: I get it, but this chapter is mostly about the async buffers

johanandren: Actually a little bit more than half of the rendered page is about the buffer and rate stages and the chapter is titled "Buffers and working with rate".

drewhk: Yes, but my primary goal was when writing this chapter is to inform people about the effects of internal buffers which regularly confused people (compared to the explicit buffers and rate transformation)

johanandren: Should we maybe move the inform-about-effects-about-async-buffers into one section about async (together with the parallelism and pipelining which is the main place we talk about asynchronous boundaries now) and the buffer+rate limiting content into the stages overview?

If we want to help people not be confused I think the place where we introduce `async` should also contain "but you also need to know this", rather than having that in an entirely different location in the docs.

drewhk: > Should we maybe move the inform-about-effects-about-async-buffers into one section about async (together with the parallelism and pipelining which is the main place we talk about asynchronous boundaries now) and the buffer+rate limiting content into the stages overview?

In my eyes these rate and buffer management subjects belong together, they all discuss various detachment concepts. Others might disagree though.

> If we want to help people not be confused I think the place where we introduce async should also contain "but you also need to know this", rather than having that in an entirely different location in the docs.

This leads to overly bloated chapters, just like the "actors" page.

patriknw: can't you just add a cross reference between the chapters?

johanandren: Yup. Will do that in the PR for #20051

johanandren: There actually already is such a cross reference. I'll see if I can make it more noticeable in the `Pipelining and Parallelism` though, that it is something you should understand before using async.

I still think that this is a better order though, when someone uses `async` and get unexpected results I don't think they will go to the docs and read under "Buffers and working with rate". I think readers will get here either because they want to introduce a buffer or handle rate and end up in the top of the chapter, or they will come through the cross references about internal buffers and end up wherever that is in the page.

ktoso: So what's to be decided here? Just if we move it down here or not?
Seems to me keeping it at the top of this page would make understanding conflate simpler, as the protocol is a bit explained here already?

ktoso:
LGTM, moving the section back as Endre suggested sounds good I thought

johanandren:
I have changed the order back with the async buffers up top!

PR #20053, opened 2016-03-16
Allow loggers to use a dedicated dispatcher, fixes #20006

### LINKED ISSUE #20006: Add option to specify dispatcher other than default for logger actors.
Currently there are options to configure dispatchers for some parts of akka backend, e.g. remoting, clustering but (as far as I can see) there is no option to specify non-default dispatcher for actors that are part of logging infrastructure - `akka.loggers`. In some cases this could lead to 

* lost logs if all threads from `default-dispatcher` are blocked and someone decides to kill JVM before some of the threads are free again 
* significant delay in writing logs to desired output

Of course it would be nice if blocking tasks were delegated to some other specific pools instead of default one but because both remoting and clustering parts of akka use this logging infrastructure having an option to specify separate dispatcher could prove useful.

guidomedina:
Allow loggers to use a dedicated dispatcher, fixes #20006

### DIFF
--- a/akka-actor/src/main/resources/reference.conf
+++ b/akka-actor/src/main/resources/reference.conf
@@ -25,6 +25,9 @@ akka {
   # (akka.actor.ActorSystem.Settings, akka.event.EventStream) parameters.
   logging-filter = "akka.event.DefaultLoggingFilter"
 
+  # Specifies the default loggers dispatcher
+  loggers-dispatcher = "akka.actor.default-dispatcher"
+
   # Loggers are created and registered synchronously during ActorSystem
   # start-up, and since they are actors, this timeout is used to bound the
   # waiting time
--- a/akka-actor/src/main/scala/akka/actor/ActorSystem.scala
+++ b/akka-actor/src/main/scala/akka/actor/ActorSystem.scala
@@ -176,6 +176,7 @@ object ActorSystem {
     final val LogLevel: String = getString("akka.loglevel")
     final val StdoutLogLevel: String = getString("akka.stdout-loglevel")
     final val Loggers: immutable.Seq[String] = immutableSeq(getStringList("akka.loggers"))
+    final val LoggersDispatcher: String = getString("akka.loggers-dispatcher")
     final val LoggingFilter: String = getString("akka.logging-filter")
     final val LoggerStartTimeout: Timeout = Timeout(config.getMillisDuration("akka.logger-startup-timeout"))
     final val LogConfigOnStart: Boolean = config.getBoolean("akka.log-config-on-start")
--- a/akka-actor/src/main/scala/akka/event/Logging.scala
+++ b/akka-actor/src/main/scala/akka/event/Logging.scala
@@ -174,7 +174,7 @@ trait LoggingBus extends ActorEventBus {
    */
   private def addLogger(system: ActorSystemImpl, clazz: Class[_ <: Actor], level: LogLevel, logName: String): ActorRef = {
     val name = "log" + Extension(system).id() + "-" + simpleName(clazz)
-    val actor = system.systemActorOf(Props(clazz), name)
+    val actor = system.systemActorOf(Props(clazz).withDispatcher(system.settings.LoggersDispatcher), name)
     implicit def timeout = system.settings.LoggerStartTimeout
     import akka.pattern.ask
     val response = try Await.result(actor ? InitializeLogger(this), timeout.duration) catch {


akka-actor/src/main/resources/reference.conf
```diff
@@ -25,6 +25,17 @@ akka {
   # (akka.actor.ActorSystem.Settings, akka.event.EventStream) parameters.
   logging-filter = "akka.event.DefaultLoggingFilter"
 
+  # Specifies the default loggers dispatcher
+  loggers-dispatcher = "akka.loggers.pinned-dispatcher"
```
patriknw: I don't think we should change the default, it's a tradeoff, some complain about too many threads/thread pools

patriknw:
OK TO TEST

patriknw:
thanks for contributing @guidomedina, that's the way to get things done ;-)

guidomedina:
I would contribute more but I'm in the process of learning Scala.

akka-actor/src/main/resources/reference.conf
guidomedina: Hmmm, too many threads? I understand what you are saying but ... it is an I/O thread, I wanted to make it use akka.io.dispatcher but then I would at being messing with the `selector` functionality whatever that is ;-)

guidomedina: I meant `akka.io.pinned-dispatcher` happy to force push if that's what you want, it would only point to that other dispatcher.

guidomedina: @patriknw

Hmmm, too many threads? I understand what you are saying but ... it is an I/O thread, I wanted to make it use `akka.io.pinned-dispatcher` but then I would be messing with the selector functionality/thread whatever that is ;-)

I can change it to default `akka.io.pinned-dispatcher` just let me know, I would just force push.

patriknw:
what could possibly be a better way to learn Scala than contributing to Akka ;-)

guidomedina:
ha, I got myself started with smaller projects like akka-kryo, Akka was too big for me 4 or 5 days ago.

akka-actor/src/main/resources/reference.conf
patriknw: I was thinking that we should keep the default as 

    loggers-dispatcher = "akka.actor.default-dispatcher"

guidomedina: @patriknw it is arguable, if most users complain that they don't want more threads they can change that but the remaining population what do you think will do? I know I will prefer a dedicated blocking I/O thread, but that's just me, what do you think is the general consensus?

guidomedina:
@patriknw @ktoso 

Commenting here to discuss:

@ktoso Patrik says he would prefer that by default it points to the default-dispatcher, what do you think is the general consensus? what will most users do? keep it at default dispatcher or move it to an I/O thread that can block?

guidomedina:
@patriknw changed it to `akka.actor.default-dispatcher`

akka-actor/src/main/resources/reference.conf
```diff
@@ -25,6 +25,9 @@ akka {
   # (akka.actor.ActorSystem.Settings, akka.event.EventStream) parameters.
   logging-filter = "akka.event.DefaultLoggingFilter"
 
+  # Specifies the default loggers dispatcher
+  loggers-dispatcher = "akka.actor.default-dispatcher"
```
johanandren: `logger` not plural right?

akka-actor/src/main/scala/akka/actor/ActorSystem.scala
```diff
@@ -176,6 +176,7 @@ object ActorSystem {
     final val LogLevel: String = getString("akka.loglevel")
     final val StdoutLogLevel: String = getString("akka.stdout-loglevel")
     final val Loggers: immutable.Seq[String] = immutableSeq(getStringList("akka.loggers"))
+    final val LoggersDispatcher: String = getString("akka.loggers-dispatcher")
```
johanandren: And same here

akka-actor/src/main/resources/reference.conf
johanandren: Ah, now I see why you thought plural, because the dispatcher will be for all if there are more than one logger specified. Ok, NVM my comments

guidomedina: @johanandren there could be more than one, as I understood from the reference.conf

guidomedina: @johanandren 

`final val Loggers: immutable.Seq[String] = immutableSeq(getStringList("akka.loggers"))`

johanandren:
LGTM

PR #20054, opened 2016-03-16
catch ActorCell creation failures for top-level actors #15947

### LINKED ISSUE #15947: Failure during top level system actor creation is not logged
When creating a top level actor and there is a failure during mailbox verification it is not logged.

rkuhn:
Previously a failure during e.g. MailboxType.create() would make the
user guardian fail, tearing down the whole system as a result. The cause
is a deep bug in handling ActorCell creation that we cannot really fix
anymore due to resulting changes in semantics, hence this fix only
targets top-level actors (where the observable difference is an
unambiguous improvement).

fixes #15947

### DIFF
--- a/akka-actor-tests/src/test/scala/akka/actor/SupervisorSpec.scala
+++ b/akka-actor-tests/src/test/scala/akka/actor/SupervisorSpec.scala
@@ -5,7 +5,6 @@
 package akka.actor
 
 import language.postfixOps
-
 import org.scalatest.BeforeAndAfterEach
 import scala.concurrent.duration._
 import akka.{ Die, Ping }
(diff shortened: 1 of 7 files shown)


akka-actor-tests/src/test/scala/akka/actor/SupervisorSpec.scala
```diff
@@ -64,10 +69,40 @@ object SupervisorSpec {
...
+      case p: Props ⇒ sender() ! context.actorOf(p)
+    }
+  }
+
+  def creator(target: ActorRef, fail: Boolean = false) = {
+    val p = Props(new Creator(target))
+    if (fail) p.withMailbox("error-mailbox") else p
+  }
+
+  val failure = new AssertionError("deliberate test failure")
```
patriknw: no risk that this is treated differently that a normal RuntimeException in `NonFatal` and such?

patriknw:
LGTM, thanks!

ktoso:
Late to the party but LGTM!
(Also [refreshed memory](http://shipilev.net/blog/2014/exceptional-performance/) if I remember that try/catch is not harmful there in our situation - looks good here I think, it really is just exceptional case)

PR #20058, opened 2016-03-16
fix endless loop race condition in NodeQueue, #19949

rkuhn:
- also fixed some forgotten copy-pasta between peekNode and pollNode
- also added JavaDoc to all methods, explaining which can be used from
  what thread
- did not fix the JDK8 improvement of using Unsafe instead of inheriting
  from AtomicReference since that inheritance is not a bad thing,
  actually


patriknw:
this is just a backport

patriknw:
Refs #19949

PR #20059, opened 2016-03-16
LoggerMailbox is calling super after cleanup

guidomedina:
LoggerMailbox is calling super after cleanup, once the logger mailbox has been cleanup by sending its messages to `Logging.StandardOutLogger` calling super might just add noise if not worse; it might try to re-log messages to dead letters after logger is dead.

### DIFF
--- a/akka-actor/src/main/scala/akka/event/LoggerMailbox.scala
+++ b/akka-actor/src/main/scala/akka/event/LoggerMailbox.scala
@@ -43,6 +43,5 @@ private[akka] class LoggerMailbox(owner: ActorRef, system: ActorSystem)
         envelope = dequeue
       }
     }
-    super.cleanUp(owner, deadLetters)
   }
 }


ktoso:
OK TO TEST

akka-actor/src/main/scala/akka/event/LoggerMailbox.scala
```diff
@@ -43,6 +43,5 @@ private[akka] class LoggerMailbox(owner: ActorRef, system: ActorSystem)
         envelope = dequeue
       }
     }
-    super.cleanUp(owner, deadLetters)
```
patriknw: what kind of noise would that be? have you seen it? after the above loop the mailbox is empty, i.e. `hasMessages` is `false`

note that `cleanUp` is called after switching out the mailbox, i.e. there will be no new messages added

guidomedina: @patriknw no I haven't seen it, I didn't know exactly when cleanUp would be called but I was thinking that maybe a message would at been inserted between end of loop and call to super (a very small chance), what you are saying that chance is then zero and that call is completely harmless, though I would still remove it as it is misleading.

WDYT?

ktoso: Shall we apply "line of code innocent until proven guilty"?

patriknw: There is a comment about this in the LoggerMailbox, otherwise we would not dare to do the loop without limits.

I think it's a good practice to call super in cleanup code, you never know if something more is added to the super class later.

guidomedina: @ktoso is common sense, testing is a good practice and everything should be tested but there are cases where mathematical logic should prevail.

This case I think is one of that, when you wrote the original code and having all the data of when a code is reproduced should you still have a robot tell you a human what your code is supposed to do?

The call, without any doubt according to @patriknw is harmless and also according to the use case ... useless, shouldn't be there.

guidomedina: @patriknw if you call super when the logger is down, super delivers to dead letters and dead letters sends to logger, if there are messages between end of loop and call to super then you are really creating an endless loop hence saying that somehow messages are still being received hence making that call not harmless anymore.

guidomedina: @patriknw @ktoso in fact it is really simple:

1. What is it that call to super does? Sends remaining messages to dead letters.
2. What dead letters does? Sends them to the logger but wait, the logger is dead

If (1) happens; how is it possible to send messages to super if they queue was just drained?
if (1) happens then (2) happens, does it make sense to send to dead-letters to later send to a dead logger?

guidomedina:
@patriknw @ktoso in fact it is really simple:

1. What is it that call to super does? Sends remaining messages to dead letters.
2. What dead letters does? Sends them to the logger but wait, the logger is dead

If (1) happens; how is it possible that there are still messages if the queue was just drained?
if (1) happens then (2) happens, does it make sense to send to dead-letters to later send them to a dead logger?

patriknw:
Currently the super call will do nothing, there will be no messages. My point is that we might add other cleanup code later and therefore it's good to keep the call to super (i.e. best practice to call super in cleanup code).

guidomedina:
Ok, close the PR then.

patriknw:
thanks, looking forward to next one

Issue #19953, opened 2016-03-06
No response produced after number of requests reach max-connections value


kstokoz:
I observe some issue with managing http connection pool in akka-http v2.4.2, 
I`m executing requests one after another in sequence, initial requests succeed and produce response until number of requests reach max-connections configuration value afterwards all requests timeout.

Some other info: https://groups.google.com/forum/#!topic/akka-user/Dw1MgZ6vgzI  

Below you can find a reproducer for this strange behavior:

```scala
import akka.actor.ActorSystem
import akka.http.scaladsl.Http
import akka.http.scaladsl.model.HttpRequest
import akka.http.scaladsl.server.Directives._
import akka.http.scaladsl.server.Route
import akka.stream.{ActorMaterializer, ActorMaterializerSettings}
import com.typesafe.config.ConfigFactory
import org.apache.commons.io.IOUtils
import org.apache.http.client.methods.HttpGet
import org.apache.http.impl.client.HttpClientBuilder

object AppConfig {
  val actorSystemConfig = ConfigFactory.parseString(
    """
      akka {
        loglevel = DEBUG
        loggers = ["akka.event.slf4j.Slf4jLogger"]
        actor {
          default-dispatcher {
            executor = "fork-join-executor"
            fork-join-executor {
              parallelism-min = 8
              parallelism-factor = 2.0
              parallelism-max = 8
            }
          }
        }
        http {
          server.parsing.max-content-length = 2048m
          client.parsing.max-content-length = 2048m

          host-connection-pool {
           max-connections = 2
           max-retries = 5
           max-open-requests = 2048
         }
       }
    }
    """
  )
}

object ReproducerApp extends App {

  implicit val system = ActorSystem("actor-system", AppConfig.actorSystemConfig)
  implicit val materializer = ActorMaterializer(ActorMaterializerSettings.create(system).withDebugLogging(true))(system)

  val route: Route = {
    (path("test") & get) {
      val response = Http().singleRequest(HttpRequest().withUri("https://gist.github.com/rklaehn/d4d3ee43443b0f4741fb"))
      complete(response)
    }
  }

  val bindingFuture = Http().bindAndHandle(route, "localhost", 12345)

  System.in.read()
  system.shutdown()
  system.awaitTermination()
}

object ReproducerClientApp extends App {

  val url = "http://localhost:12345/test"
  val client = HttpClientBuilder.create().build()

  (1 to 5).foreach { i =>
    val get = new HttpGet(url)
    val response = client.execute(get)
[...]

PR #20060, opened 2016-03-17
=htc #19953 Complete Promise in captureTermination onDownstreamFinish

leachbj:
PoolSlot uses StreamUtils.captureTermination to determine when the
associated request is completed, when a request from the pool was
subsequently used as a response by a server Flow the rendering
process would cancel the subscription to the Source but that
would not complete the Promise in StreamUtils.captureTerminate and so
the SlotEvent.RequestCompleted would not be generated.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
@@ -65,13 +65,12 @@ private[http] object StreamUtils {
     val promise = Promise[Unit]()
     val transformer = new PushStage[T, T] {
       def onPush(element: T, ctx: Context[T]) = ctx.push(element)
-      override def onUpstreamFinish(ctx: Context[T]) = {
-        promise.success(())
-        super.onUpstreamFinish(ctx)
-      }
       override def onUpstreamFailure(cause: Throwable, ctx: Context[T]) = {
         promise.failure(cause)
         ctx.fail(cause)
+      }
+      override def postStop(): Unit = {
+        promise.trySuccess(())
       }
     }
     source.transform(() ⇒ transformer) -> promise.future
--- a/akka-http-core/src/test/scala/akka/http/impl/util/StreamUtilsSpec.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/util/StreamUtilsSpec.scala
@@ -1,0 +1,48 @@
+/**
+ * Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.http.impl.util
+
+import akka.stream.ActorMaterializer
+import akka.stream.scaladsl.{Sink, Source}
+import akka.testkit.AkkaSpec
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+import scala.util.Failure
+
+class StreamUtilsSpec extends AkkaSpec {
+  implicit val materializer = ActorMaterializer()
+
+  "captureTermination" should {
+    "signal completion" when {
+      "upstream terminates" in {
+        val (newSource, whenCompleted) = StreamUtils.captureTermination(Source(List(1, 2, 3)))
+
+        newSource.runWith(Sink.ignore)
+
+        Await.result(whenCompleted, 3.seconds) shouldBe ()
+      }
+
+      "upstream fails" in {
+        val ex = new RuntimeException("ex")
+        val (newSource, whenCompleted) = StreamUtils.captureTermination(Source.failed[Int](ex))
+        intercept[RuntimeException] {
+          Await.result(newSource.runWith(Sink.head), 3.second)
+        } should be theSameInstanceAs ex
+
+        Await.ready(whenCompleted, 3.seconds).value shouldBe Some(Failure(ex))
+      }
+
+
+      "downstream cancels" in {
+        val (newSource, whenCompleted) = StreamUtils.captureTermination(Source(List(1, 2, 3)))
+
+        newSource.runWith(Sink.head)
+
+        Await.result(whenCompleted, 3.seconds) shouldBe ()
+      }
+    }
+  }
+
+}


johanandren:
OK TO TEST

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -74,6 +74,10 @@ private[http] object StreamUtils {
         promise.failure(cause)
         ctx.fail(cause)
       }
+      override def onDownstreamFinish(ctx: Context[T]): TerminationDirective = {
```
drewhk: hm, why not just override postStop then and use `promise.trySuccess(())` (because `failure` might have been called already.).

ktoso: :+1:

ktoso:
Resolves #19953

ktoso:
Could you also add a test that reproduces the issue we saw in the related ticket?
It should be easily reproducible in a ClientServerSpec using one of the samples in the issue

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
drewhk: we have that button now ;)

ktoso: lol, forgot about it, old habits die slowly :) Will do.

leachbj:
I went for the simpler unit test approach, the tests fail without the fix and just verify the contract for the promise which is what the the pool relies on.

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
leachbj: I've switched to use postStop but not sure how I can test that scenario, hints?

akka-http-core/src/test/scala/akka/http/impl/util/StreamUtilsSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+  implicit val materializer = ActorMaterializer()
+
+  "captureTermination" should {
+    "signal completion" when {
+      "upstream terminates" in {
+        val (newSource, whenCompleted) = StreamUtils.captureTermination(Source(List(1, 2, 3)))
+
+        newSource.runWith(Sink.ignore)
+
+        Await.result(whenCompleted, 100.millis) shouldBe ()
```
patriknw: `100.millis` is way to short for some CI environments, please use `remainOrDefault` (or at least 3.seconds if it's not important to fail fast)

akka-http-core/src/test/scala/akka/http/impl/util/StreamUtilsSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+      }
+
+      "upstream fails" in {
+        val ex = new RuntimeException("ex")
+        val (newSource, whenCompleted) = StreamUtils.captureTermination(Source.failed[Int](ex))
+        intercept[RuntimeException] {
+          Await.result(newSource.runWith(Sink.head), 1.second)
```
patriknw: too short

akka-http-core/src/test/scala/akka/http/impl/util/StreamUtilsSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+        } should be theSameInstanceAs ex
+
+        Await.ready(whenCompleted, 100.millis).value shouldBe Some(Failure(ex))
```
patriknw: too short

akka-http-core/src/test/scala/akka/http/impl/util/StreamUtilsSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+      }
+
+
+      "downstream cancels" in {
+        val (newSource, whenCompleted) = StreamUtils.captureTermination(Source(List(1, 2, 3)))
+
+        newSource.runWith(Sink.head)
+
+        Await.result(whenCompleted, 100.millis) shouldBe ()
```
patriknw: too short

patriknw:
LGTM apart from the test timeouts

akka-http-core/src/test/scala/akka/http/impl/util/StreamUtilsSpec.scala
leachbj: interesting, I went for that value as it seemed common other tests.  I've bumped to 3 seconds to avoid possible false negatives on slow builds.

akka-http-core/src/test/scala/akka/http/scaladsl/ClientServerSpec.scala
```diff
@@ -417,6 +417,50 @@ class ClientServerSpec extends WordSpec with Matchers with BeforeAndAfterAll wit
       }
     }
 
+    "property should capture terminations" in Utils.assertAllStagesStopped {
```
leachbj: remove me this, doesn't actually verify the fix

leachbj:
The test failure looks to be unrelated;

`Reporter completed abruptly with an exception after receiving event: SuiteCompleted(Ordinal(0, 872),VirtualProcessorTest,akka.stream.tck.VirtualProcessorTest,Some(akka.stream.tck.VirtualProcessorTest),Some(23503),Some(MotionToSuppress),Some(TopOfClass(akka.stream.tck.VirtualProcessorTest)),None,None,pool-46-thread-1,1459229311375).
java.lang.RuntimeException: unexpected event [TestIgnored(Ordinal(0, 747),VirtualProcessorTest,akka.stream.tck.VirtualProcessorTest,Some(akka.stream.tck.VirtualProcessorTest),optional_spec111_maySupportMultiSubscribe,optional_spec111_maySupportMultiSubscribe,Some(IndentedText(- optional_spec111_maySupportMultiSubscribe,optional_spec111_maySupportMultiSubscribe,1)),Some(TopOfMethod(akka.stream.tck.VirtualProcessorTest,public void akka.stream.tck.VirtualProcessorTest.optional_spec111_maySupportMultiSubscribe())),None,pool-46-thread-1,1459229289525)]
	at org.scalatest.tools.JUnitXmlReporter.unexpected(JUnitXmlReporter.scala:495)
	at org.scalatest.tools.JUnitXmlReporter.processTest(JUnitXmlReporter.scala:326)
	at org.scalatest.tools.JUnitXmlReporter.getTestsuite(JUnitXmlReporter.scala:166)
	at org.scalatest.tools.JUnitXmlReporter.writeSuiteFile(JUnitXmlReporter.scala:76)
	at org.scalatest.tools.JUnitXmlReporter.apply(JUnitXmlReporter.scala:59)
	at org.scalatest.DispatchReporter$Propagator$$anonfun$run$1.apply(DispatchReporter.scala:240)
	at org.scalatest.DispatchReporter$Propagator$$anonfun$run$1.apply(DispatchReporter.scala:239)
	at scala.collection.immutable.List.foreach(List.scala:381)
	at org.scalatest.DispatchReporter$Propagator.run(DispatchReporter.scala:239)
	at java.lang.Thread.run(Thread.java:745)`

patriknw:
LGTM
@leachbj please squash to one commit

patriknw:
Refs #19953

Issue #20061, opened 2016-03-17
failure to initialize remoting should include cause


patriknw:
It is for example impossible to see what's wrong if ssl is misconfigured.

PR #20062, opened 2016-03-17
include cause in remote init failure logging, #20061

patriknw:


patriknw:
Fixes #20061

johanandren:
LGTM

ktoso:
I assume you verified these many levels deep is what we need here?
If so, LGTM

patriknw:
yes, that was what was needed for the ssl problems that I debugged

ktoso:
Great, good to merge then IMO

Issue #20032, opened 2016-03-15
User-provided materializer will be used to create system wide objects


samueltardieu:
`Http().singleRequest()` creates or reuses a cached connection stored in a ActorSystem-wide pool. However, this connection is created using the user-provided materializer. If this materializer is tied to an actor through its `context`, other actors will no longer be able to contact the same host/port after the initial materializer has been shutdown.

This is likely to happen if the user has a `implicit val materializer = ActorMaterializer()(context)` in their actor.

ActorSystem-wide cached objects should be created using a materializer with an ActorSystem scope instead of the materializer provided by the caller.

The following example illustrates the problem:

``` scala
import akka.actor.ActorSystem
import akka.http.scaladsl.Http
import akka.http.scaladsl.model.{HttpRequest, HttpResponse}
import akka.stream.ActorMaterializer
import akka.stream.scaladsl.Flow
import org.specs2.mutable._

import scala.concurrent.Await
import scala.concurrent.duration._

class MaterializerSpec extends Specification {

  implicit val system = ActorSystem("test-system")
  implicit val dispatcher = system.dispatcher
  val systemMaterializer = ActorMaterializer()(system)

  val handler = Flow[HttpRequest].map(_ => new HttpResponse())
  val bindingFuture = Http().bindAndHandle(handler, "localhost", 0)(systemMaterializer)
  val port = Await.result(bindingFuture.map(_.localAddress.getPort), 1.second)
  val uri = s"http://localhost:$port"

  "non-global materializer should not be globally stored" in {

    {
      implicit val localMaterializer = ActorMaterializer()(system)
      // The local materializer will wrongly be used to create an ActorSystem-wide connection
      // in the cached connection pool.
      val response = Await.result(Http().singleRequest(new HttpRequest(uri = uri)), 1.second)
      localMaterializer.shutdown()
      response.status.intValue() must be equalTo 200
    }
    {
      implicit val localMaterializer = ActorMaterializer()(system)
      // The following request will fail with a timeout because the local materializer of the
      // first request has been used to create the cached connection in the connection pool.
      //
      // Commenting the local materializer shutdown above will make it work as the previous
      // materializer would still be alive.

      // Using a request to another host/port will also work as a new connection will be
      // created in the connection pool with this local materializer.
      val response = Await.result(Http().singleRequest(new HttpRequest(uri = uri)), 1.second)
      localMaterializer.shutdown()
      response.status.intValue() must be equalTo 200
    }
  }
}
```

[...]

PR #20065, opened 2016-03-17
Use an Actor-System wide materializer for cached gateways (#20032)

samueltardieu:
The materializer lifespan must not be shorter than the one of the cache it is used in.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -548,9 +548,12 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
   // every ActorSystem maintains its own connection pools
   private[http] val hostPoolCache = new ConcurrentHashMap[HostConnectionPoolSetup, Future[PoolGateway]]
 
+  // an ActorSystem-wide materializer must be used for gateways stored in cache
+  private[this] val hostPoolCacheFm = ActorMaterializer()(system)
+
   private def cachedGateway(request: HttpRequest,
                             settings: ConnectionPoolSettings, connectionContext: ConnectionContext,
-                            log: LoggingAdapter)(implicit fm: Materializer): Future[PoolGateway] =
+                            log: LoggingAdapter): Future[PoolGateway] =
     if (request.uri.scheme.nonEmpty && request.uri.authority.nonEmpty) {
       val httpsCtx = if (request.uri.scheme.equalsIgnoreCase("https")) connectionContext else ConnectionContext.noEncryption()
       val setup = ConnectionPoolSetup(settings, httpsCtx, log)
@@ -563,13 +566,13 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
     }
 
   /** INTERNAL API */
-  private[http] def cachedGateway(setup: HostConnectionPoolSetup)(implicit fm: Materializer): Future[PoolGateway] = {
+  private[http] def cachedGateway(setup: HostConnectionPoolSetup): Future[PoolGateway] = {
     val gatewayPromise = Promise[PoolGateway]()
     hostPoolCache.putIfAbsent(setup, gatewayPromise.future) match {
       case null ⇒ // only one thread can get here at a time
         val whenShuttingDown = Promise[Done]()
         val gateway =
-          try new PoolGateway(setup, whenShuttingDown)
+          try new PoolGateway(setup, whenShuttingDown)(system, hostPoolCacheFm)
           catch {
             case NonFatal(e) ⇒
               hostPoolCache.remove(setup)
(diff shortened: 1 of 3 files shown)


ktoso:
OK TO TEST

ktoso:
Thanks, yep that's one of the valid ways to fix it.
I'm worried that now it is rather inconsistent, e.g. https://github.com/akka/akka/pull/20065/files#diff-57c314af9e0369f72b0f60ff1fa56f0aL391 the `cachedPool()` methods still take a materializer, and also we should allow configuring the materializer if we indeed create one internally (configure materializer settings for `Http()`).

samueltardieu:
The materializer passed by the user to `cachedHostConnectionPool` (for example) is still used for the `gatewayClientFlow`, that's why I kept it there.

The settings can still be configured at the `ActorSystem` level. Did you think of adding another section under `akka.http` to specifically let the user overload some settings for the materializer, or to let the user specify a materializer class name for the cached gateways?

ktoso:
> The settings can still be configured at the ActorSystem level. Did you think of adding another section under akka.http to specifically let the user overload some settings for the materializer, or to let the user specify a materializer class name for the cached gateways?

Yes and no; It's possible via the top level configuration yes, it's now not possible via code (which we do want to allow – a popular use case is "get default settings and tweak it around using `copy()`")

samueltardieu:
Only two `private` method were changed (removal of the implicit `ActorMaterializer`) and none of them has a privacy access greater than the `http` package. Is that a real binary incompatibility?

ktoso:
Two things here:

1) Current state of PR is binary compatible, though MiMa does not know that – `private[http]` is visible as PUBLIC in bytecode (sic!) (unless a crazy person called those from java, but that we don't care about since it's internal API). That's why we add exceptions for such.

2) It is also very confusing (public API is now lying), as:

```scala
// that `fm` implicit is carried from all the way up in public API
  private def cachedHostConnectionPool[T](setup: HostConnectionPoolSetup)(
    implicit fm: Materializer): Flow[(HttpRequest, T), (Try[HttpResponse], T), HostConnectionPool] =
    gatewayClientFlow(setup, cachedGateway(setup)/*implicit fm was passed in here*/)
```

So now the public facing API looks like:

```scala
  def cachedHostConnectionPoolHttps[T](host: String, ...)(implicit fm: Materializer)
```

however the materializer is not used, which will lead to weird behaviour ("i changed materializer settings, but it's not picking it up, help!" :wink:).

We need to take a holistic look at all affected APIs here and see how to fix it.

samueltardieu:
- binary compatibility exception: I don't know how to add it, any pointer? ;
- materializer: it is used AFAICS to establish the `gatewayClientFlow`, which can be teared down for example by shutting down the materializer, it just won't make the cached gateway unusable anymore.

I understand what you're saying concerning the materializer settings. What setting could be changed by the user and be perceptible if used for the cached connection, and would this setting be problematic if set on the default materializer? Won't the advanced user avoid the use of global pools anyway and fine tune them?

samueltardieu:
(also note that before this patch, the settings of the *first* materializer to use a particular scheme/host/port will be used for every request with the same triplet, which is quite a disturbing situation for the user and an even worse race condition if several actors target this triplet)

ktoso:
> materializer to use a particular scheme/host/port will be used for every request with the same triplet,

Good observation, a bit weird indeed. Seems like another point to not get the materializer in any case form the implicit (remove from public API). 

Next step I think is that we should investigate if/where we should remove the implicit param, by going in from the public APIs, we may indeed want to change those.

samueltardieu:
@ktoso Shall we merge this one and open a new issue concerning the consistency of the materializers interface in `Http()` for methods using cached gateways?

samueltardieu:
Rebased

ktoso:
We should solve it in one go. I don't want to have Http in a weird state where we can't release because it's in an halfway-done inconsistent state if you know what I mean.

ktoso:
In other words: We're heading towards the right direction with this PR, just need to pull it through all the way :-)

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -548,9 +548,12 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
   // every ActorSystem maintains its own connection pools
   private[http] val hostPoolCache = new ConcurrentHashMap[HostConnectionPoolSetup, Future[PoolGateway]]
 
+  // an ActorSystem-wide materializer must be used for gateways stored in cache
+  private[this] val hostPoolCacheFm = ActorMaterializer()(system)
```
patriknw: no need to configure this materializer? should it read settings from a separate config section, with fallback to the default?

samueltardieu: Yes, I think that was what @ktoso had in mind. Can you confirm?

ktoso: Yeah, we need to pass in or allow configuring if we are to have our own here.

patriknw: ping

samueltardieu: This change is moot because of #20082 which fixes the initial problem (local materializer used) as a side effect. I've opened #20188 to that effect.

samueltardieu:
This PR has been superseded by #20082 which fixes the initial #20032 problem as a side effect of rewriting the pool gateways (among others).

PR #20070, opened 2016-03-18
some random findings

rkuhn:
from yesterday’s walkthrough:
- use `union` operator instead of `++` for Sets
- make mapMaterializedValue mandatory for all FlowOpsMat

### DIFF
--- a/akka-docs/rst/scala/code/docs/stream/TwitterStreamQuickstartDocSpec.scala
+++ b/akka-docs/rst/scala/code/docs/stream/TwitterStreamQuickstartDocSpec.scala
@@ -32,8 +32,7 @@ object TwitterStreamQuickstartDocSpec {
   //#model
 
   //#tweet-source
-  val tweets: Source[Tweet, NotUsed]
-  //#tweet-source
+  val tweets: Source[Tweet, NotUsed] //#tweet-source
   = Source(
     Tweet(Author("rolandkuhn"), System.currentTimeMillis, "#akka rocks!") ::
       Tweet(Author("patriknw"), System.currentTimeMillis, "#akka !") ::
--- a/project/MiMa.scala
+++ b/project/MiMa.scala
@@ -659,6 +659,9 @@ object MiMa extends AutoPlugin {
         ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.impl.fusing.Drop.onPush"),
         ProblemFilters.exclude[FinalClassProblem]("akka.stream.stage.GraphStageLogic$Reading"), // this class is private
 
+        // lifting this method to the type where it belongs
+        ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.scaladsl.FlowOpsMat.mapMaterializedValue"),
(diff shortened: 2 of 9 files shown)


ktoso:
LGTM, thanks!

drewhk:
LGTM

patriknw:
@rkuhn did you check for `--` also?

rkuhn:
No, I didn’t. `--` is only used in StreamLayout.validate and I replaced it by `&~`, but that requires some care because that operator binds less tightly than `-` …

patriknw:
I'd rather use `diff` than the less obvious `&~`

patriknw:
@rkuhn Binary compatibility check failed!

rkuhn:
`diff` operator used, MiMa exception added

akka-docs/rst/scala/code/docs/stream/TwitterStreamQuickstartDocSpec.scala
```diff
@@ -32,8 +32,7 @@ object TwitterStreamQuickstartDocSpec {
   //#model
 
   //#tweet-source
-  val tweets: Source[Tweet, NotUsed]
-  //#tweet-source
+  val tweets: Source[Tweet, NotUsed] //#tweet-source
```
rkuhn: wat?

ktoso: yeah, this is on purpose. We only show the type in the docs, not the full impl (as you see below - it's just a mock).

project/MiMa.scala
```diff
@@ -659,6 +659,9 @@ object MiMa extends AutoPlugin {
         ProblemFilters.exclude[DirectMissingMethodProblem]("akka.stream.impl.fusing.Drop.onPush"),
         ProblemFilters.exclude[FinalClassProblem]("akka.stream.stage.GraphStageLogic$Reading"), // this class is private
 
+        // lifting this method to the type where it belongs
+        ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.scaladsl.FlowOpsMat.mapMaterializedValue"),
```
patriknw: and this is safe because user doesn't extend `FlowOpsMat`?

patriknw:
LGTM

akka-docs/rst/scala/code/docs/stream/TwitterStreamQuickstartDocSpec.scala
rkuhn: the “wat” is about the reformatting, that makes no sense and that I didn’t ask for

ktoso: The machines are revolting against humanity - doing things we never asked them to... The end is nigh

project/MiMa.scala
ktoso: I think so, yeah

ktoso:
LGTM, thanks

Issue #20028, opened 2016-03-14
figure out whether TickSourceCancellable.cancel returning true is a bug


rkuhn:
the cancellation logic is a rather complicated dance—quite possibly more complex than it needs to be

samueltardieu:
See PR https://github.com/akka/akka/pull/20076.

samueltardieu:
For reference, the current state before the fix is:
- `cancel()` will return `true` even after the first time it is called;
- a tick might be emitted after `cancel()` has returned;
- `isCancelled` can return false after `cancel()` has returned.

PR #20076, opened 2016-03-18
Simplify cancellation of TickSource (#20028)

samueltardieu:
With this fix, the following now works:
- `tickSource.cancel()` returns `true` the first time then `false` afterwards;
- `tickSource.isCancelled()` returns `true` as soon as `tickSource.cancel()` has returned;
- `tickSource` will not emit any tick after `tickSource.cancel()` has returned.

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/TickSourceSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/TickSourceSpec.scala
@@ -101,5 +101,37 @@ class TickSourceSpec extends AkkaSpec {
       c.expectComplete()
     }
 
+    "acknowledge cancellation only once" in assertAllStagesStopped {
+      val c = TestSubscriber.manualProbe[String]()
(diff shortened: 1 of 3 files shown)


ktoso:
Resolves #20028

ktoso:
OK TO TEST

samueltardieu:
Interesting, I didn't know this constraint (cannot create callbacks in `GraphStageLogic` constructor) and it didn't get caught in my tests. I'll create the callback in `preStart` then.

drewhk:
Yes, the startup sequence for a stage is somewhat involved and there is a circular dependency between the stage logic and its interpreter which is not resolved until preStart()

samueltardieu:
@drewhk This is all for the better: I could simplify the logic further by not creating the asynchronous callback until I really need to invoke it.

rkuhn:
Thanks @samueltardieu ! Please also add tests that verify the newly supported correctness :-) (don’t need to be race-condition tests, just single-threaded correctness is enough)

samueltardieu:
Tests added, rebased.

patriknw:
LGTM

ktoso:
LGTM, thanks!

PR #20078, opened 2016-03-18
Fix documented type in authenticateOAuth2Async

### LINKED ISSUE #20077: Bad type reference in doc for authenticateOAuth2Async
From @chris-zen on gitter:

> Hi, not sure if it is a mistake, but in the following doc page for the Signature section it shows the type Authenticator[T] instead of the type AsyncAuthenticator[T]
http://doc.akka.io/docs/akka/2.4.2/scala/http/routing-dsl/directives/security-directives/authenticateOAuth2Async.html

samueltardieu:
This fixes #20077.


ktoso:
Thanks, i think this indeed was a mistake :-)

ktoso:
OK TO TEST

PR #20082, opened 2016-03-19
=htc Simplify and fix pool gateway synchronization

### LINKED ISSUE #20032: User-provided materializer will be used to create system wide objects

### LINKED ISSUE #20071: akka-http client PoolGateway stuck in spinloop, eventually using 100% CPU
Unfortunately, I don't currently know *why* or *how* it happens, but it is possible for the `state` in a `PoolGateway` to be set to a `NewIncarnation(newGatewayFuture)`, where `newGatewayFuture` is a `FastFuture.successful(..)` of it self.

In other words, at https://github.com/akka/akka/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala#L57, `this == previousIncarnation`. By attaching a debugger to the process I've added a breakpoint that doesn't suspend, but only logs `this` and `previousIncarnation`, and it keeps logging the following:

    this: akka.http.impl.engine.client.PoolGateway@79fcbcef - previous: akka.http.impl.engine.client.PoolGateway@79fcbcef
    this: akka.http.impl.engine.client.PoolGateway@1df76436 - previous: akka.http.impl.engine.client.PoolGateway@1df76436
    this: akka.http.impl.engine.client.PoolGateway@516279e7 - previous: akka.http.impl.engine.client.PoolGateway@516279e7
    this: akka.http.impl.engine.client.PoolGateway@79fcbcef - previous: akka.http.impl.engine.client.PoolGateway@79fcbcef
    this: akka.http.impl.engine.client.PoolGateway@1df76436 - previous: akka.http.impl.engine.client.PoolGateway@1df76436
    this: akka.http.impl.engine.client.PoolGateway@516279e7 - previous: akka.http.impl.engine.client.PoolGateway@516279e7

I'm guessing it's a race condition somewhere as it does not happen often, but when it does, the thread that's running calling the method will spin.

Observations from canary-ing it in production is that it can take a long time before it occurs (last time for me it took 90 hours under moderate workload for it to happen to the first thread), but just a few hours until there were four requests spinning (on a server with four cpu cores.). Hopefully that observation may help in finding the cause.

I'm seeing this using the following version(s):

    "com.typesafe.akka:akka-actor_2.11:2.4.2",
    "com.typesafe.akka:akka-http-core_2.11:2.4.2",
    "com.typesafe.akka:akka-http-experimental_2.11:2.4.2",
    "com.typesafe.akka:akka-parsing_2.11:2.4.2",
    "com.typesafe.akka:akka-stream_2.11:2.4.2",

### LINKED ISSUE #20080: Separate host connection pools might be reunited later
If the user creates two `newHostConnectionPool()` with the same settings, it gets two independent pools with a given number of `max-connections` (one `PoolGateway`/`PoolInterfaceActor` per pool).

However, if those two pools are shutdown because `idle-timeout` is reached, they will both be replaced by a new incarnation of the same `PoolGateway`/`PoolInterfaceActor` (obtained from `Http().cachedGateway(hcps)`) and will have their connections shared while they were created independently. As a consequence, whereas the user could queue up to `max-connections*2` requests originally, they can now only queue up to `max-connections` before an error is thrown.

The new pools should be created without using the pool cache.

### LINKED ISSUE #20081: Remove race condition documented in PoolInterfaceActor
The pool recreation mechanism after an idle timeout has a race condition which is documented in `PoolInterfaceActor.scala`:

```
// Note that this forwarding will stop when we receive completion from the pool flow
// (because we stop ourselves then), so there is a very small chance of a request ending
// up as a dead letter if the sending thread gets interrupted for a long time right before
// the `ref ! PoolRequest(...)` in the PoolGateway
```

This race condition should be removed, as it might lead to a lockup in user code as the request will never receive a response, even after the programmed timeouts have expired.

samueltardieu:
Rewrite the pool gateway synchronization so that:
- The collapsing chain of gateways and the cache in HttpExt have been replaced by a PoolMasterActor which takes care of orchestrating the various PoolInterfaceActor.
- The documented race condition in PoolInterfaceActor is gone. No PoolInterfaceActor will receive new requests after the gateway shutdown has been initiated (fix #20081).
- A gateway created by newHostConnectionPool is never merged with the cached pools (fix #20080).
- A gateway created by newHostConnectionPool will use the given materializer for every incarnation of the PoolInterfaceActor. Gateways pointing to cached connection pools will use an ActorSystem-wide gateway (fix #20032).
- By construction, the bug where the collapsing chain of gateways can contain dependency loops no longer exists (fix #20071).

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
@@ -4,91 +4,89 @@
 
 package akka.http.impl.engine.client
 
-import java.util.concurrent.atomic.AtomicReference
-import akka.Done
+import java.util.concurrent.atomic.AtomicLong
 
-import scala.annotation.tailrec
-import scala.concurrent.{ Future, Promise }
+import akka.Done
+import akka.actor.ActorRef
+import akka.http.impl.engine.client.PoolGateway.{ GatewayIdentifier, SharedGateway }
+import akka.http.impl.engine.client.PoolMasterActor._
 import akka.http.impl.settings.HostConnectionPoolSetup
-import akka.actor.{ Deploy, Props, ActorSystem, ActorRef }
-import akka.http.scaladsl.Http
-import akka.http.scaladsl.model.{ HttpResponse, HttpRequest }
+import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
 import akka.stream.Materializer
 
-private object PoolGateway {
-
-  sealed trait State
-  final case class Running(interfaceActorRef: ActorRef,
-                           shutdownStartedPromise: Promise[Done],
-                           shutdownCompletedPromise: Promise[Done]) extends State
-  final case class IsShutdown(shutdownCompleted: Future[Done]) extends State
-  final case class NewIncarnation(gatewayFuture: Future[PoolGateway]) extends State
-}
+import scala.concurrent.{ Future, Promise }
 
 /**
- * Manages access to a host connection pool or rather: a sequence of pool incarnations.
+ * Manages access to a host connection pool through the [[PoolMasterActor]]
  *
- * A host connection pool for a given [[HostConnectionPoolSetup]] is a running stream, whose outside interface is
- * provided by its [[PoolInterfaceActor]] actor. The actor accepts [[PoolInterfaceActor.PoolRequest]] messages
- * and completes their `responsePromise` whenever the respective response has been received (or an error occurred).
- *
- * A [[PoolGateway]] provides a layer of indirection between the pool cache and the actual
- * pools that is required to allow a pool incarnation to fully terminate (e.g. after an idle-timeout)
- * and be transparently replaced by a new incarnation if required.
- * Removal of cache entries for terminated pools is also supported, because old gateway references that
- * get reused will automatically forward requests directed at them to the latest pool incarnation from the cache.
+ * A [[PoolGateway]] is represented by its [[HostConnectionPoolSetup]] and its [[GatewayIdentifier]]. If the later
+ * is [[SharedGateway]], it means that a shared pool must be used for this particular [[HostConnectionPoolSetup]].
  */
-private[http] class PoolGateway(hcps: HostConnectionPoolSetup,
-                                _shutdownStartedPromise: Promise[Done])( // constructor arg only
-                                  implicit system: ActorSystem, fm: Materializer) {
-  import PoolGateway._
-  import fm.executionContext
+private[http] final class PoolGateway(gatewayRef: ActorRef, val hcps: HostConnectionPoolSetup, val gatewayId: GatewayIdentifier)(implicit fm: Materializer) {
 
-  private val state = {
-    val shutdownCompletedPromise = Promise[Done]()
-    val props = Props(new PoolInterfaceActor(hcps, shutdownCompletedPromise, this)).withDeploy(Deploy.local)
-    val ref = system.actorOf(props, PoolInterfaceActor.name.next())
-    new AtomicReference[State](Running(ref, _shutdownStartedPromise, shutdownCompletedPromise))
+  /**
+   * Send a request through the corresponding pool. If the pool is not running, it will be started
+   * automatically. If it is shutting down, it will restart as soon as the shutdown operation is
+   * complete and serve this request.
+   *
+   * @param request the request
+   * @return the response
+   */
+  def apply(request: HttpRequest): Future[HttpResponse] = {
+    val responsePromise = Promise[HttpResponse]()
+    gatewayRef ! SendRequest(this, request, responsePromise, fm)
+    responsePromise.future
   }
 
-  def currentState: Any = state.get() // enables test access
+  /**
+   * Start the corresponding pool to make it ready to serve requests. If the pool is already started,
+   * this does nothing. If it is being shutdown, it will restart as soon as the shutdown operation
+   * is complete.
+   *
+   * @return the gateway itself
+   */
+  def startPool(): PoolGateway = {
+    gatewayRef ! StartPool(this, fm)
+    this
+  }
 
-  def apply(request: HttpRequest, previousIncarnation: PoolGateway = null): Future[HttpResponse] =
-    state.get match {
-      case Running(ref, _, _) ⇒
-        val responsePromise = Promise[HttpResponse]()
-        ref ! PoolInterfaceActor.PoolRequest(request, responsePromise)
-        responsePromise.future
+  /**
+   * Shutdown the corresponding pool and signal its termination. If the pool is not running or is
+   * being shutting down, this does nothing,
+   *
+   * @return a Future completed when the pool has been shutdown.
+   */
+  def shutdown(): Future[Done] = {
+    val shutdownCompletedPromise = Promise[Done]()
+    gatewayRef ! Shutdown(this, shutdownCompletedPromise)
+    shutdownCompletedPromise.future
+  }
 
-      case IsShutdown(shutdownCompleted) ⇒
-        // delay starting the next pool incarnation until the current pool has completed its shutdown
-        shutdownCompleted.flatMap { _ ⇒
-          val newGatewayFuture = Http().cachedGateway(hcps)
-          // a simple set is fine here as `newGatewayFuture` will be identical for all threads getting here
-          state.set(NewIncarnation(newGatewayFuture))
-          apply(request)
-        }
+  override def toString = s"PoolGateway(hcps = $hcps)"
 
-      case x @ NewIncarnation(newGatewayFuture) ⇒
-        if (previousIncarnation != null)
-          previousIncarnation.state.set(x) // collapse incarnation chain
-        newGatewayFuture.flatMap(_(request, this))
+  // INTERNAL API (testing only)
+  private[client] def poolStatus(): Future[Option[PoolInterfaceStatus]] = {
+    val statusPromise = Promise[Option[PoolInterfaceStatus]]()
+    gatewayRef ! PoolStatus(this, statusPromise)
+    statusPromise.future
+  }
+
+  override def equals(that: Any): Boolean =
+    that match {
+      case p: PoolGateway ⇒ p.hcps == hcps && p.gatewayId == gatewayId
+      case _              ⇒ false
     }
 
-  // triggers a shutdown of the current pool, even if it is already a later incarnation
-  @tailrec final def shutdown(): Future[Done] =
-    state.get match {
-      case x @ Running(ref, shutdownStartedPromise, shutdownCompletedPromise) ⇒
-        if (state.compareAndSet(x, IsShutdown(shutdownCompletedPromise.future))) {
-          shutdownStartedPromise.success(Done) // trigger cache removal
-          ref ! PoolInterfaceActor.Shutdown
-          shutdownCompletedPromise.future
-        } else shutdown() // CAS loop (not a spinlock)
+  override def hashCode(): Int = hcps.hashCode() ^ gatewayId.hashCode()
+}
 
-      case IsShutdown(x)                    ⇒ x
+private[http] object PoolGateway {
 
-      case NewIncarnation(newGatewayFuture) ⇒ newGatewayFuture.flatMap(_.shutdownAux())
-    }
+  sealed trait GatewayIdentifier
+  case object SharedGateway extends GatewayIdentifier
+  final case class UniqueGateway(id: Long) extends GatewayIdentifier
 
-  private def shutdownAux() = shutdown() // alias required for @tailrec
+  private[this] val uniqueGatewayId = new AtomicLong(0)
+  def newUniqueGatewayIdentifier = UniqueGateway(uniqueGatewayId.incrementAndGet())
+
 }
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
@@ -1,0 +1,151 @@
+package akka.http.impl.engine.client
+
+import akka.Done
+import akka.actor.{ Actor, ActorLogging, ActorRef, DeadLetterSuppression, Deploy, NoSerializationVerificationNeeded, Props, Terminated }
+import akka.http.impl.engine.client.PoolInterfaceActor.PoolRequest
+import akka.http.impl.settings.HostConnectionPoolSetup
+import akka.http.scaladsl.HttpExt
+import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
+import akka.stream.Materializer
+
+import scala.concurrent.{ Future, Promise }
+
+/**
+ * Manages access to a host connection pool or rather: a sequence of pool incarnations.
+ *
+ * A host connection pool for a given [[HostConnectionPoolSetup]] is a running stream, whose outside interface is
+ * provided by its [[PoolInterfaceActor]] actor. The actor accepts [[PoolInterfaceActor.PoolRequest]] messages
+ * and completes their `responsePromise` whenever the respective response has been received (or an error occurred).
+ *
+ * The [[PoolMasterActor]] provides a layer of indirection between a [[PoolGateway]], which represents a pool,
+ * and the [[PoolInterfaceActor]] instances which are created on-demand and stopped after an idle-timeout.
+ *
+ * Several [[PoolGateway]] objects may be mapped to the same pool if they have the same [[HostConnectionPoolSetup]]
+ * and are marked as being shared. This is the case for example for gateways obtained through
+ * [[HttpExt.cachedHostConnectionPool]]. Some other gateways are not shared, such as those obtained through
+ * [[HttpExt.newHostConnectionPool]], and will have their dedicated restartable pool.
+ *
+ */
+private[http] final class PoolMasterActor extends Actor with ActorLogging {
+
+  import PoolMasterActor._
+
+  private[this] var poolStatus = Map[PoolGateway, PoolInterfaceStatus]()
+  private[this] var poolInterfaces = Map[ActorRef, PoolGateway]()
+
+  /**
+   * Start a new pool interface actor, register it in our maps, and watch its death. No actor should
+   * currently exist for this pool.
+   *
+   * @param gateway the pool gateway this pool corresponds to
+   * @param fm the materializer to use for this pool
+   * @return the newly created actor ref
+   */
+  private[this] def startPoolInterfaceActor(gateway: PoolGateway)(implicit fm: Materializer): ActorRef = {
+    if (poolStatus.contains(gateway)) {
+      throw new IllegalStateException(s"pool interface actor for $gateway already exists")
+    }
+    val props = Props(new PoolInterfaceActor(gateway)).withDeploy(Deploy.local)
+    val ref = context.actorOf(props, PoolInterfaceActor.name.next())
+    poolStatus += gateway -> PoolInterfaceRunning(ref)
+    poolInterfaces += ref -> gateway
+    context.watch(ref)
+  }
+
+  def receive = {
+
+    // Start or restart a pool without sending it a request. This is used to ensure that
+    // freshly created pools will be ready to serve requests immediately.
+    case s @ StartPool(gateway, materializer) ⇒
+      poolStatus.get(gateway) match {
+        case Some(PoolInterfaceRunning(_)) ⇒
+        case Some(PoolInterfaceShuttingDown(shutdownCompletedPromise)) ⇒
+          // Pool is being shutdown. When this is done, start the pool again.
+          shutdownCompletedPromise.future.onComplete(_ ⇒ self ! s)(context.dispatcher)
+        case None ⇒
+          startPoolInterfaceActor(gateway)(materializer)
+      }
+
+    // Send a request to a pool. If needed, the pool will be started or restarted.
+    case s @ SendRequest(gateway, request, responsePromise, materializer) ⇒
+      poolStatus.get(gateway) match {
+        case Some(PoolInterfaceRunning(ref)) ⇒
+          ref ! PoolRequest(request, responsePromise)
+        case Some(PoolInterfaceShuttingDown(shutdownCompletedPromise)) ⇒
+          // The request will be resent when the pool shutdown is complete (the first
+          // request will recreate the pool).
+          shutdownCompletedPromise.future.foreach(_ ⇒ self ! s)(context.dispatcher)
+        case None ⇒
+          startPoolInterfaceActor(gateway)(materializer) ! PoolRequest(request, responsePromise)
+      }
+
+    // Shutdown a pool and signal its termination.
+    case Shutdown(gateway, shutdownCompletedPromise) ⇒
+      poolStatus.get(gateway).foreach {
+        case PoolInterfaceRunning(ref) ⇒
+          // Ask the pool to shutdown itself. Queued connections will be resent here
+          // to this actor by the pool actor, they will be retried once the shutdown
+          // has completed.
+          ref ! PoolInterfaceActor.Shutdown
+          poolStatus += gateway -> PoolInterfaceShuttingDown(shutdownCompletedPromise)
+        case PoolInterfaceShuttingDown(formerPromise) ⇒
+          // Pool is already shutting down, mirror the existing promise.
+          shutdownCompletedPromise.tryCompleteWith(formerPromise.future)
+        case _ ⇒
+          // Pool does not exist, shutdown is not needed.
+          shutdownCompletedPromise.trySuccess(Done)
+      }
+
+    // Shutdown all known pools and signal their termination.
+    case ShutdownAll(shutdownCompletedPromise) ⇒
+      import context.dispatcher
+      def track(remaining: Iterator[Future[Done]]): Unit =
+        if (remaining.hasNext) remaining.next().onComplete(_ ⇒ track(remaining))
+        else shutdownCompletedPromise.trySuccess(Done)
+      track(poolStatus.keys.map(_.shutdown()).toIterator)
+
+    // When a pool actor terminate, signal its termination and remove it from our maps.
+    case Terminated(ref) ⇒
+      poolInterfaces.get(ref).foreach { gateway ⇒
+        poolStatus.get(gateway) match {
+          case Some(PoolInterfaceRunning(_)) ⇒
+            log.error("connection pool for {} has shut down unexpectedly", gateway)
+          case Some(PoolInterfaceShuttingDown(shutdownCompletedPromise)) ⇒
+            shutdownCompletedPromise.trySuccess(Done)
+          case None ⇒
+          // This will never happen as poolInterfaces and poolStatus are modified
+          // together. If there is no status then there is no gateway to start with.
+        }
+        poolStatus -= gateway
+        poolInterfaces -= ref
+      }
+
+    // Testing only.
+    case PoolStatus(gateway, statusPromise) ⇒
+      statusPromise.success(poolStatus.get(gateway))
+
+    // Testing only.
+    case PoolSize(sizePromise) ⇒
+      sizePromise.success(poolStatus.size)
+
+  }
+
+}
+
+private[http] object PoolMasterActor {
+
+  sealed trait PoolInterfaceStatus
+  final case class PoolInterfaceRunning(ref: ActorRef) extends PoolInterfaceStatus
+  final case class PoolInterfaceShuttingDown(shutdownCompletedPromise: Promise[Done]) extends PoolInterfaceStatus
+
+  final case class StartPool(gateway: PoolGateway, materializer: Materializer) extends NoSerializationVerificationNeeded
+  final case class SendRequest(gateway: PoolGateway, request: HttpRequest, responsePromise: Promise[HttpResponse], materializer: Materializer)
+    extends NoSerializationVerificationNeeded
+  final case class Shutdown(gateway: PoolGateway, shutdownCompletedPromise: Promise[Done]) extends NoSerializationVerificationNeeded with DeadLetterSuppression
+  final case class ShutdownAll(shutdownCompletedPromise: Promise[Done]) extends NoSerializationVerificationNeeded with DeadLetterSuppression
+
+  // INTERNAL API (for testing only)
+  final case class PoolStatus(gateway: PoolGateway, statusPromise: Promise[Option[PoolInterfaceStatus]]) extends NoSerializationVerificationNeeded
+  final case class PoolSize(sizePromise: Promise[Int]) extends NoSerializationVerificationNeeded
+
+}
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
@@ -15,6 +15,8 @@ import akka.stream.Materializer
 import scala.concurrent.{ Future, Promise }
 
 /**
+ * INTERNAL API
+ *
  * Manages access to a host connection pool or rather: a sequence of pool incarnations.
  *
  * A host connection pool for a given [[HostConnectionPoolSetup]] is a running stream, whose outside interface is
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
@@ -1,3 +1,7 @@
+/**
+ * Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+
 package akka.http.impl.engine.client
 
 import akka.Done
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
@@ -140,6 +140,8 @@ private[http] final class PoolMasterActor extends Actor with ActorLogging {
 
 private[http] object PoolMasterActor {
 
+  val props = Props[PoolMasterActor].withDeploy(Deploy.local)
+
   sealed trait PoolInterfaceStatus
   final case class PoolInterfaceRunning(ref: ActorRef) extends PoolInterfaceStatus
   final case class PoolInterfaceShuttingDown(shutdownCompletedPromise: Promise[Done]) extends PoolInterfaceStatus
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -589,7 +589,11 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
       case other                        ⇒ TLSPlacebo() // if it's not HTTPS, we don't enable SSL/TLS
     }
 
-  /** For testing only. */
+  /**
+   * INTERNAL API
+   *
+   * For testing only
+   */
   private[scaladsl] def poolSize: Future[Int] = {
     val sizePromise = Promise[Int]()
     poolMasterActorRef ! PoolSize(sizePromise)
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -191,7 +191,7 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
 
   // ** CLIENT ** //
 
-  private[this] val poolMasterActorRef = system.actorOf(Props(new PoolMasterActor).withDeploy(Deploy.local), "pool-master")
+  private[this] val poolMasterActorRef = system.actorOf(PoolMasterActor.props, "pool-master")
   private[this] val systemMaterializer = ActorMaterializer()
 
   /**
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -11,6 +11,7 @@ import javax.net.ssl._
 import akka.actor._
 import akka.event.{ Logging, LoggingAdapter }
 import akka.http.impl.engine.HttpConnectionTimeoutException
+import akka.http.impl.engine.client.PoolMasterActor.{ PoolSize, ShutdownAll }
 import akka.http.impl.engine.client._
 import akka.http.impl.engine.server._
 import akka.http.impl.engine.ws.WebSocketClientBlueprint
@@ -19,7 +20,7 @@ import akka.http.impl.util.{ MapError, StreamUtils }
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers.Host
 import akka.http.scaladsl.model.ws.{ Message, WebSocketRequest, WebSocketUpgradeResponse }
-import akka.http.scaladsl.settings.{ ServerSettings, ClientConnectionSettings, ConnectionPoolSettings }
+import akka.http.scaladsl.settings.{ ClientConnectionSettings, ConnectionPoolSettings, ServerSettings }
 import akka.http.scaladsl.util.FastFuture
 import akka.{ Done, NotUsed }
 import akka.stream._
@@ -190,6 +191,9 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
 
   // ** CLIENT ** //
 
+  private[this] val poolMasterActorRef = system.actorOf(Props(new PoolMasterActor).withDeploy(Deploy.local), "pool-master")
+  private[this] val systemMaterializer = ActorMaterializer()
+
   /**
    * Creates a [[akka.stream.scaladsl.Flow]] representing a prospective HTTP client connection to the given endpoint.
    * Every materialization of the produced flow will attempt to establish a new outgoing connection.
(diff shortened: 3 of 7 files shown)


ktoso:
OK TO TEST

samueltardieu:
Rebased

viktorklang:
What were the alternate solutions explored prior to choosing the lockbased one?

samueltardieu:
I tried to keep a solution based on CAS (plus a loop at shutdown time) as was done before, but I couldn't remove the race condition present at https://github.com/akka/akka/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala#L60 : if a context switch happens just before this line, we might in the meantime have sent a shutdown message to the pool interface actor (this is documented at https://github.com/akka/akka/blob/master/akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolInterfaceActor.scala#L130-L133). The result might be very bad, as the client will never be notified that the request did not get through, not even through the various timeouts set on the request.

The only bulletproof solution I found for this issue was to ensure that once the shutdown message has been sent to the interface actor, no new request can be sent to it as the actor will not have a chance to forward it back to the gateway. We really need an atomic "check gateway state / send request to pool interface actor if ok" test here. We could use an extra actor to serialize the requests, but that would be much more costly than using a short-lived lock.

viktorklang:
@samueltardieu was it considered to go through an actor instead of going though a lock?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -4,24 +4,26 @@
...
 import akka.stream.Materializer
 
+import scala.concurrent.{ Future, Promise }
+
 private object PoolGateway {
 
   sealed trait State
   final case class Running(interfaceActorRef: ActorRef,
-                           shutdownStartedPromise: Promise[Done],
+                           shutdownStartedPromise: Option[Promise[Done]],
```
viktorklang: Why is this an option rather than introducing a ShutdownStarted data type?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -39,56 +41,74 @@ private object PoolGateway {
...
-    new AtomicReference[State](Running(ref, _shutdownStartedPromise, shutdownCompletedPromise))
+    Running(ref, _shutdownStartedPromise, shutdownCompletedPromise)
   }
 
-  def currentState: Any = state.get() // enables test access
+  private[this] val lock = new ReentrantLock(true)
 
-  def apply(request: HttpRequest, previousIncarnation: PoolGateway = null): Future[HttpResponse] =
-    state.get match {
+  def currentState: Any = state // enables test access
```
viktorklang: This is not threadsafe

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -39,56 +41,74 @@ private object PoolGateway {
...
+
+  private def setCurrentState(newState: State) = {
```
viktorklang: return type

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -39,56 +41,74 @@ private object PoolGateway {
...
+    lock.unlock()
+  }
+
+  def apply(request: HttpRequest, previousIncarnation: PoolGateway = null): Future[HttpResponse] = {
+    lock.lock()
+    state match {
       case Running(ref, _, _) ⇒
         val responsePromise = Promise[HttpResponse]()
         ref ! PoolInterfaceActor.PoolRequest(request, responsePromise)
+        lock.unlock()
```
viktorklang: unlock may not get called if Running-extractor throws, or if it cannot allocate a Promise of PoolRequest or if !-throws.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -39,56 +41,74 @@ private object PoolGateway {
...
-          apply(request)
-        }
+      case IsShutdown(shutdownCompleted, recreateFromCache) ⇒
+        // Delay starting the next pool incarnation until the current pool has completed its shutdown.
+        // If this gateway was created through the gateway cache, go through the cache again.
+        val newGatewayFuture = shutdownCompleted.flatMap(_ ⇒
+          if (recreateFromCache) FastFuture.successful(new PoolGateway(hcps))
+          else Http().cachedGateway(hcps))
+        state = NewIncarnation(newGatewayFuture)
+        lock.unlock()
```
viktorklang: Multiple reasons why this might not get called

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -39,56 +41,74 @@ private object PoolGateway {
...
+    lock.lock()
+    state match {
       case x @ Running(ref, shutdownStartedPromise, shutdownCompletedPromise) ⇒
-        if (state.compareAndSet(x, IsShutdown(shutdownCompletedPromise.future))) {
-          shutdownStartedPromise.success(Done) // trigger cache removal
-          ref ! PoolInterfaceActor.Shutdown
-          shutdownCompletedPromise.future
-        } else shutdown() // CAS loop (not a spinlock)
+        state = IsShutdown(shutdownCompletedPromise.future, shutdownStartedPromise.isDefined)
+        lock.unlock()
```
viktorklang: might not get called

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -39,56 +41,74 @@ private object PoolGateway {
...
+        shutdownCompletedPromise.future
 
-      case IsShutdown(x)                    ⇒ x
+      case IsShutdown(x, _) ⇒
+        lock.unlock()
+        x
 
-      case NewIncarnation(newGatewayFuture) ⇒ newGatewayFuture.flatMap(_.shutdownAux())
+      case NewIncarnation(newGatewayFuture) ⇒
+        lock.unlock()
```
viktorklang: might not get called

samueltardieu:
@viktorklang Yes, it is mentioned in the last sentence of my previous reply.

Using an actor to replace the lock could be done in two ways:

1. We use an actor per pool interface to replace the lock. Right now, the user can keep a reference onto the gateway as long as they want, even if the gateway has been shutdown. This is why there is a mechanism to build a collapsing chain of gateways corresponding to the successive incarnations of the gateway and their respective pool interface actor. Right now (before and after the proposed change), the non-collapsed part of the chain is reclaimed by the GC when the user drop its last reference to the pool. We would have no way of shutting down the locking actor when the last reference is dropped, so we would leak actors.
2. We use a unique actor to represent the locks for all the pools. In this case, we are adding contention between the various pools while they are kept independent from each other with the lock.

Both ways seem to cause more problems than they solve.

viktorklang:
@samueltardieu The difference is that by using a single actor there is no actual blocking. whereas with the lock-based approach there is locking for both reads and writes. Also, what's the difference in performance and maintainability?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -39,56 +41,74 @@ private object PoolGateway {
...
+  private[this] var state: State = {
     val shutdownCompletedPromise = Promise[Done]()
     val props = Props(new PoolInterfaceActor(hcps, shutdownCompletedPromise, this)).withDeploy(Deploy.local)
     val ref = system.actorOf(props, PoolInterfaceActor.name.next())
-    new AtomicReference[State](Running(ref, _shutdownStartedPromise, shutdownCompletedPromise))
+    Running(ref, _shutdownStartedPromise, shutdownCompletedPromise)
   }
 
-  def currentState: Any = state.get() // enables test access
+  private[this] val lock = new ReentrantLock(true)
```
viktorklang: why true?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -39,56 +41,74 @@ private object PoolGateway {
...
-                                _shutdownStartedPromise: Promise[Done])( // constructor arg only
+                                _shutdownStartedPromise: Option[Promise[Done]] = None)( // used for cache invalidation
                                   implicit system: ActorSystem, fm: Materializer) {
   import PoolGateway._
   import fm.executionContext
 
-  private val state = {
+  private[this] var state: State = {
     val shutdownCompletedPromise = Promise[Done]()
     val props = Props(new PoolInterfaceActor(hcps, shutdownCompletedPromise, this)).withDeploy(Deploy.local)
```
viktorklang: This leaks `this` to another thread ( the actor's thread) before fully constructed.

samueltardieu:
Performance wise, the lock is held for very short periods of time (it is never used as reentrant and does not contain blocking operations as far as I can tell). The parts between the lock are (in code order) one of:

- new request when gateway is up: compare state / create Promise / send message
- new request when gateway has been shut down: compare state / map existing future  / set state
- new request when gateway has a new incarnation: compare state / map existing future / set state, followed by (at most once per old/new incarnation pair to collapse the chain): set state
- shutdown when gateway is up: compare state / set state
- shutdown when gateway has been shut down: compare state
- shutdown when gateway has a new incarnation: compare state

However, the Actor approach has a certain appeal if the actor was in charge of maintaining the relationships between the gateway and the pool interface actors and may be able to handle the state of those actors quite elegantly. I'll try come up with a comparaison after my afternoon meetings, I'll keep you posted.

viktorklang:
@samueltardieu Looking forward to hearing more. (also, time holding the lock is impacted by thread scheduling artifacts). Since were talking about things aimed at Java8, I suspect a StampedLock would be more scalable than ReentrantLock if you indeed do not rely on reentrancy.

samueltardieu:
@viktorklang Indeed the actor-based solution looks more elegant.

How/when can I stop the `PoolMasterActor` created in `HttpExt`? `Http().shutdownAllConnectionPools()` could terminate it, but the documentation implies that new pools can be started after this call. I don't know if this is a shortcoming of the previous implementation (total termination was not feasible, so document it) or if this is a feature.

viktorklang:
@samueltardieu Does it need to be stopped? (It will be stopped when the ActorSystem shuts down)

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * Start a new pool interface actor, register it in our maps, and watch its death.
+   *
+   * @param gateway the pool gateway this pool corresponds to
+   * @param fm the materializer to use for this pool
+   * @return the newly created actor ref
+   */
+  private[this] def startPoolInterfaceActor(gateway: PoolGateway)(implicit fm: Materializer): ActorRef = {
+    val props = Props(new PoolInterfaceActor(gateway)).withDeploy(Deploy.local)
+    val ref = context.system.actorOf(props, PoolInterfaceActor.name.next())
```
viktorklang: why below system?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -0,0 +1,140 @@
...
+
+  def receive = {
+
+    // Start or restart a pool without sending it a request. This is used to ensure that
+    // freshly created pools will be ready to serve requests immediately.
+    case s @ StartPool(gateway, materializer) ⇒
+      poolStatus.get(gateway) match {
+        case Some(PoolInterfaceRunning(_)) ⇒
+        case Some(PoolInterfaceShuttingDown(shutdownCompletedPromise)) ⇒
+          shutdownCompletedPromise.future.foreach(_ ⇒ self ! s)
```
viktorklang: what if this Future failed?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -0,0 +1,140 @@
...
+  private[this] var poolInterfaces = Map[ActorRef, PoolGateway]()
+
+  /**
+   * Start a new pool interface actor, register it in our maps, and watch its death.
+   *
+   * @param gateway the pool gateway this pool corresponds to
+   * @param fm the materializer to use for this pool
+   * @return the newly created actor ref
+   */
+  private[this] def startPoolInterfaceActor(gateway: PoolGateway)(implicit fm: Materializer): ActorRef = {
```
viktorklang: What if there's already an InterfaceActor for the given poolgateway?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -0,0 +1,140 @@
...
+    poolStatus += gateway -> PoolInterfaceRunning(ref)
+    poolInterfaces += ref -> gateway
+    context.watch(ref)
+  }
+
+  def receive = {
+
+    // Start or restart a pool without sending it a request. This is used to ensure that
+    // freshly created pools will be ready to serve requests immediately.
+    case s @ StartPool(gateway, materializer) ⇒
```
viktorklang: I think the receive-block might be more clear if it is modeled as a state machine.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -0,0 +1,140 @@
...
+    // Shutdown a pool and signal its termination.
+    case Shutdown(gateway, shutdownCompletedPromise) ⇒
+      poolStatus.get(gateway).foreach {
+        case PoolInterfaceRunning(ref) ⇒
+          ref ! PoolInterfaceActor.Shutdown
+          poolStatus += gateway -> PoolInterfaceShuttingDown(shutdownCompletedPromise)
+        case PoolInterfaceShuttingDown(formerPromise) ⇒
+          shutdownCompletedPromise.completeWith(formerPromise.future)
+        case _ ⇒
+          shutdownCompletedPromise.success(Done)
```
viktorklang: If the `shutdownCompletedPromise` has already been completed, then this line will throw an exception.

samueltardieu:
No, it doesn't. And shouldn't in fact.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -0,0 +1,140 @@
...
+      }
+
+    // Shutdown all known pools and signal their termination.
+    case ShutdownAll(shutdownCompletedPromise) ⇒
+      val allGateways = poolStatus.keys
+      val counter = new AtomicInteger(allGateways.size)
+      for (gateway ← allGateways)
+        gateway.shutdown().foreach(_ ⇒ if (counter.decrementAndGet() == 0) shutdownCompletedPromise.success(Done))
```
viktorklang: This: 

~~~scala
val allGateways = poolStatus.keys
 +      val counter = new AtomicInteger(allGateways.size)
 +      for (gateway ← allGateways)
 +        gateway.shutdown().foreach(_ ⇒ if (counter.decrementAndGet() == 0) shutdownCompletedPromise.success(Done))
~~~

can be written as:
~~~
def shutdown(remaining: Iterator[PoolGateway]): Unit =
if (remaining.hasNext) i.next().shutdown().onComplete { _ => shutdown(i) } else shutdownCompletedPromise.success(Done)

shutdown(poolStatus.keysIterator)
~~~

Do you also need to clear the poolStatus map?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -0,0 +1,140 @@
...
+
+    // When a pool actor terminate, signal its termination and remove it from our maps.
+    case Terminated(ref) ⇒
+      poolInterfaces.get(ref).foreach { gateway ⇒
+        poolStatus.get(gateway) match {
+          case Some(PoolInterfaceShuttingDown(shutdownCompletedPromise)) ⇒
+            shutdownCompletedPromise.success(Done)
+          case _ ⇒
```
viktorklang: Why do we ignore this case?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -0,0 +1,140 @@
...
+ * and are marked as being shared. This is the case for example for gateways obtained through
+ * [[HttpExt.cachedHostConnectionPool]]. Some other gateways are not shared, such as those obtained through
+ * [[HttpExt.newHostConnectionPool]], and will have their dedicated restartable pool.
+ *
+ */
+private[http] class PoolMasterActor extends Actor {
+
+  import PoolMasterActor._
+
+  private[this] implicit val dispatcher = context.dispatcher
```
viktorklang: I'd avoid this, rather import `context.dispatcher` as close to where the ExecutionContext is needed as possible (to make it easier to reason about what parts uses it)

viktorklang:
@samueltardieu Thanks. I like what you're doing.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: This is where they were created before. But I agree, this would make more sense to create them under the pool master now that it exists.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: It cannot fail, the shutdown will always be successful. Even if it did not take place (which should never happen), the future would just never be completed.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: This is an internal private method which is only called when the interface actor for this gateway (or an equivalent one through `.equals`) does not already exist. Would you want me to add an extra comment to it?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: It is not really one, and moreover it handles all the interface actors at once, so this would neither be easy nor elegant to model with a FSM. However, I will add comments describing the life cycle of a pool interface actor as seen by the pool master actor.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
viktorklang: No, is it either A) a bug (in which it should be handled) or B) a feature (in which case the desired outcome is not immediately obvious to me)

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
viktorklang: agreed!

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
viktorklang: You don't need to model it as an Akka FSM, `context.become` may be succinct enough :)

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
viktorklang: But that assumption is not encoded here. So if someone changes the implementation down the line, and that assumption doesn't hold anymore, it will be a silent bug, will it not?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -0,0 +1,140 @@
...
+
+  case class StartPool(gateway: PoolGateway, materializer: Materializer) extends NoSerializationVerificationNeeded
+  case class SendRequest(gateway: PoolGateway, request: HttpRequest, responsePromise: Promise[HttpResponse], materializer: Materializer)
+    extends NoSerializationVerificationNeeded
+  case class Shutdown(gateway: PoolGateway, shutdownCompletedPromise: Promise[Done]) extends NoSerializationVerificationNeeded with DeadLetterSuppression
+  case class ShutdownAll(shutdownCompletedPromise: Promise[Done]) extends NoSerializationVerificationNeeded with DeadLetterSuppression
+
+  // Testing only.
+  case class PoolStatus(gateway: PoolGateway, statusPromise: Promise[Option[PoolInterfaceStatus]]) extends NoSerializationVerificationNeeded
+  case class PoolSize(sizePromise: Promise[Int]) extends NoSerializationVerificationNeeded
```
viktorklang: Add `final` for all the case classes.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: I don't see how this would happen. The promise can be completed it three exclusive ways;
- immediately if the gateway does not exist;
- as a completion of an existing promise;
- after being put into a `PoolInterfaceShuttingDown`, and it is only extracted from it in the `Terminated` message.

Or maybe your concern is that an already completed promise could be given to this actor? This should not happen, because the promise is created in the `PoolGateway` which is a private object.

viktorklang: My point is that if it is not a desired effect to have an exception thrown here if the Promise is already completed, it should use `trySuccess`, and optionally checking the returned boolean and logging if it is unexpected.

(This is akin to not want to allow buffer overruns :))

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: No need to clear the maps: we just start the shutdown of the various gateways, and when it is done the maps will be cleared of the corresponding entries.

The alternate way you're proposing to shutdown the gateways waits for one pool to be shutdown before going to the next. This is not a good idea, as a pool might take a long time to shutdown if it is handling ongoing connections.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: An interface pool actor terminates without warning: apart from marking this pool as dead (to be able to restart it), there is nothing else we can do.

The two ignored cases are "pool running and abruptly stopped" and "unknown actor terminates". The first one might need an error in the log, the second one can never happen.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: Ok, will do.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
viktorklang: Solved:
~~~
def track(remaining: Iterator[Future[Done]]): Unit =
if (remaining.hasNext) i.next().onComplete { _ => track(i) } else shutdownCompletedPromise.success(Done)
track(poolStatus.keys.map(_.shutdown()))
~~~

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
viktorklang: Add the two skipped states and a comment as to why, :)

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: Faire enough. Will use `andThen` instead.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: Ok, will throw `IllegalStateException` in this case with an informative error message.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: Sure, but I still don't get it: this is really not a state machine, as any of those events can happen at any time. You may receive `StartPool` for a pool followed by either something else for this pool (but you don't have to) or a message for another pool. Each message is dispatched independently, even if the state of each pool could be modeled as a tiny state machine in itself, but the receive function would never change.

viktorklang: Ok, then I probably misunderstood the code. Thanks for clarifying!

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: Ok.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: I'll use this, I like the idea of chaining the futures rather than decrementing a counter.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: Ok.

samueltardieu:
@viktorklang I'll push the fixes as separate commits on the branch to ease review, to be squashed later.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -4,91 +4,85 @@
...
+   */
+  def apply(request: HttpRequest): Future[HttpResponse] = {
+    val responsePromise = Promise[HttpResponse]()
+    gatewayRef ! SendRequest(this, request, responsePromise, fm)
+    responsePromise.future
+  }
 
-  private val state = {
+  /**
+   * Start the corresponding pool to make it ready to serve requests.
```
viktorklang: Perhaps document that this is an idempotent operation?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -4,91 +4,85 @@
...
+   *
+   * @return the gateway itself
+   */
+  def startPool(): PoolGateway = {
+    gatewayRef ! StartPool(this, fm)
+    this
+  }
+
+  /**
+   * Shutdown the corresponding pool and signal its termination. The pool can be restarted by sending a new request
```
viktorklang: Document that this is an idempotent operation?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -4,91 +4,85 @@
...
-    val ref = system.actorOf(props, PoolInterfaceActor.name.next())
-    new AtomicReference[State](Running(ref, _shutdownStartedPromise, shutdownCompletedPromise))
+    gatewayRef ! Shutdown(this, shutdownCompletedPromise)
+    shutdownCompletedPromise.future
+  }
+
+  override def toString = s"PoolGateway(hcps = $hcps)"
+
+  // Testing only.
+  def poolStatus(): Future[Option[PoolInterfaceStatus]] = {
```
viktorklang: If this is truly for testing purposes, shouldn't this have an access modifier and a comment about it being INTERNAL API?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -0,0 +1,145 @@
...
+ * The [[PoolMasterActor]] provides a layer of indirection between a [[PoolGateway]], which represents a pool,
+ * and the [[PoolInterfaceActor]] instances which are created on-demand and stopped after an idle-timeout.
+ *
+ * Several [[PoolGateway]] objects may be mapped to the same pool if they have the same [[HostConnectionPoolSetup]]
+ * and are marked as being shared. This is the case for example for gateways obtained through
+ * [[HttpExt.cachedHostConnectionPool]]. Some other gateways are not shared, such as those obtained through
+ * [[HttpExt.newHostConnectionPool]], and will have their dedicated restartable pool.
+ *
+ */
+private[http] class PoolMasterActor extends Actor with ActorLogging {
```
viktorklang: final?

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -0,0 +1,145 @@
...
+   * Start a new pool interface actor, register it in our maps, and watch its death. No actor should
+   * currently exist for this pool.
+   *
+   * @param gateway the pool gateway this pool corresponds to
+   * @param fm the materializer to use for this pool
+   * @return the newly created actor ref
+   */
+  private[this] def startPoolInterfaceActor(gateway: PoolGateway)(implicit fm: Materializer): ActorRef = {
+    if (poolStatus.contains(gateway)) {
+      throw new IllegalStateException(s"pool interface actor for $gateway already exists")
```
viktorklang: Note that this will crash the actor and send it to the supervisor.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
samueltardieu: Ok.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
samueltardieu: Ok.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
samueltardieu: Yes. I can restrict it to `private[client]`, and `Http().poolSize()` to `private[scaladsl]`.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: Ok.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
samueltardieu: Yes, but I already feel that this is over-engineered: this method is only called if `poolStatus.get(gateway)` is `None`. If somebody modifies the code despite the documentation and the explicit redundant test in the method, I sure hope that this will break instead of failing silently.

samueltardieu:
@viktorklang If you don't have any more comments for the time being, tell me and I'll resubmit a squashed version.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -4,91 +4,89 @@
...
 
-      case IsShutdown(x)                    ⇒ x
+  sealed trait GatewayIdentifier
+  case object SharedGateway extends GatewayIdentifier
+  final case class UniqueGateway(id: Long) extends GatewayIdentifier
 
-      case NewIncarnation(newGatewayFuture) ⇒ newGatewayFuture.flatMap(_.shutdownAux())
-    }
+  private[this] val uniqueGatewayId = new AtomicLong(0)
+  def newUniqueGatewayIdentifier = UniqueGateway(uniqueGatewayId.incrementAndGet())
```
viktorklang: Since the gateways are all children of the same parent actor, I don't think this AtomicLong is needed anymore.
A single var in the parent actor would be sufficient

samueltardieu: The gateways are objects, which are created independently by the `Http()` object, possibly in parallel. The *pools* are children of the master actor, but gateways must be able to get a unique id that will be used to identify them.

This identifier is used to differentiate between gateways to pools with identical `HostConnectionPoolSetup`. Those pools are either shared (created through `Http().cachedHostConnectionPool(…)` for example, or by `Http().singleRequest(…)`, in which case their gateways use the common identifier `SharedGateway`. However, if the user creates two `Http().newHostConnectionPool(…)` with the same `HostConnectionPoolSetup`, those pools must not be merged as their `max-connections` parameter shall be interpreted per pool, not globally.

This is why those non-shared gateways need to have a unique identifier which, together with the `HostConnectionPoolSetup`, ensures that requests get addressed to the right pool.

viktorklang: Thanks, makes total sense. I should probably try to get some sleep and not
review more code today :)

samueltardieu:
Rebased.

drewhk:
Unstash and revisit https://github.com/akka/akka/pull/20190#issuecomment-204353829 once this gets merged

samueltardieu:
Indeed, this PR fixes #20071 by construction as the collapsing chain of gateways and the associated race condition will no longer exist.

drewhk:
Ok, good to know then

samueltardieu:
Updated description.

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -0,0 +1,151 @@
+package akka.http.impl.engine.client
```
ktoso: minor (I'll add) missing license header

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -0,0 +1,151 @@
...
+  def receive = {
+
+    // Start or restart a pool without sending it a request. This is used to ensure that
+    // freshly created pools will be ready to serve requests immediately.
+    case s @ StartPool(gateway, materializer) ⇒
+      poolStatus.get(gateway) match {
+        case Some(PoolInterfaceRunning(_)) ⇒
+        case Some(PoolInterfaceShuttingDown(shutdownCompletedPromise)) ⇒
+          // Pool is being shutdown. When this is done, start the pool again.
+          shutdownCompletedPromise.future.onComplete(_ ⇒ self ! s)(context.dispatcher)
```
ktoso: good :+1:

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolMasterActor.scala
```diff
@@ -0,0 +1,151 @@
...
+ * The [[PoolMasterActor]] provides a layer of indirection between a [[PoolGateway]], which represents a pool,
+ * and the [[PoolInterfaceActor]] instances which are created on-demand and stopped after an idle-timeout.
+ *
+ * Several [[PoolGateway]] objects may be mapped to the same pool if they have the same [[HostConnectionPoolSetup]]
+ * and are marked as being shared. This is the case for example for gateways obtained through
+ * [[HttpExt.cachedHostConnectionPool]]. Some other gateways are not shared, such as those obtained through
+ * [[HttpExt.newHostConnectionPool]], and will have their dedicated restartable pool.
+ *
+ */
+private[http] final class PoolMasterActor extends Actor with ActorLogging {
```
ktoso: we always need to mark with "INTERNAL API" as well, since for Java it looks like `public` when `private[x]` is used.

akka-http-core/src/test/scala/akka/http/impl/engine/client/ConnectionPoolSpec.scala
```diff
@@ -207,20 +210,17 @@ class ConnectionPoolSpec extends AkkaSpec("""
...
 
     "transparently restart after idle shutdown" in new TestSetup() {
       val (requestIn, responseOut, responseOutSub, hcp) = cachedHostConnectionPool[Int](idleTimeout = 1.second)
 
-      val gateway = Await.result(hcp.gatewayFuture, 500.millis)
-      val PoolGateway.Running(_, _, shutdownCompletedPromise) = gateway.currentState
-      Await.result(shutdownCompletedPromise.future, 1500.millis) // verify shutdown completed
+      val gateway = hcp.gateway
+      Await.result(gateway.poolStatus(), 1500.millis).get shouldBe a[PoolInterfaceRunning]
+      awaitCond({ Await.result(gateway.poolStatus(), 1500.millis).isEmpty }, 1500.millis)
```
ktoso: outer timeout larger, inner smaller I'd propose here

ktoso:
Looks very good, some minor things found but the logic expressed as Actor looks very nice and valid.
I want to give the HttpExt another read but need to move to somewhere else now, sounds like we may merge this today even. Thanks @samueltardieu

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -613,6 +588,13 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
       case hctx: HttpsConnectionContext ⇒ TLS(hctx.sslContext, hctx.firstSession, role, hostInfo = hostInfo)
       case other                        ⇒ TLSPlacebo() // if it's not HTTPS, we don't enable SSL/TLS
     }
+
+  /** For testing only. */
+  private[scaladsl] def poolSize: Future[Int] = {
```
ktoso: `/** INTERNAL API*/`

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -545,64 +547,37 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
...
-                            settings: ConnectionPoolSettings, connectionContext: ConnectionContext,
-                            log: LoggingAdapter)(implicit fm: Materializer): Future[PoolGateway] =
+  private def sharedGateway(request: HttpRequest, settings: ConnectionPoolSettings, connectionContext: ConnectionContext, log: LoggingAdapter): PoolGateway = {
     if (request.uri.scheme.nonEmpty && request.uri.authority.nonEmpty) {
       val httpsCtx = if (request.uri.scheme.equalsIgnoreCase("https")) connectionContext else ConnectionContext.noEncryption()
       val setup = ConnectionPoolSetup(settings, httpsCtx, log)
       val host = request.uri.authority.host.toString()
       val hcps = HostConnectionPoolSetup(host, request.uri.effectivePort, setup)
-      cachedGateway(hcps)
+      sharedGateway(hcps)
```
ktoso: right

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -190,6 +191,9 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
 
   // ** CLIENT ** //
 
+  private[this] val poolMasterActorRef = system.actorOf(Props(new PoolMasterActor).withDeploy(Deploy.local), "pool-master")
```
ktoso: `PoolMasterActor.props` to be consistent with our codebase

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -190,6 +191,9 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
...
+  private[this] val systemMaterializer = ActorMaterializer()
```
ktoso: This touches the somewhere else touched upon topic of 'then we should allow configuring it',
but we can tackle that along the way from here on

samueltardieu:
Now that you can easily squash merges, I've added fixes for your remarks as additional commits.

ktoso:
PLS BUILD

ktoso:
Great contrib, thanks a lot @samueltardieu :+1:

PR #19632, opened 2016-01-27
WIP #18919 Align Java HTTP server DSL with Scala

jypma:
This replaces the Java HTTP server DSL with a Java-8 centric one which exposes all scala DSL concepts to be usable from Java, including custom directives, (un)marshallers, rejections, headers, and type safety for path and query parameters. 

Implemented according to the direction agreed upon with @ktoso in https://github.com/akka/akka/issues/18919 . The class `JavaRouteTest` demoes the Java DSL's capabilities.

List of things to be done (somewhat in order):
- [X] Migrated extra directives: `FileAndResourceDirectives`, `CodingDirectives`,`CookieDirectives`,`MiscDirectives`,  `HostDirectives`, `SchemeDirectives`, `WebSocketDirectives`, `CacheConditionDirectives`,  `RangeDirectives`
- [X] Migrate the existing tests in `akka-http-tests` once the DSL direction is agreed 
- [X] Fix compile errors in `akka-docs`
- [X] Fill in the missing directives that haven't been migrated yet 
- [x] Discuss how to resolve the introduced `TODO` and `FIXME` markers
- [ ] Replace the current Java `RouteTestKit` (which has its own implementation different from the scala `RouteTestKit`) with a [wrapper](https://github.com/jypma/scala-japi/blob/master/akkaStreamsTest/src/main/scala/com/tradeshift/scalajapi/akka/route/RouteTestKit.scala) that delegates and re-uses the scala one.
- [ ] Update textual documentation in `akka-docs`
- [x] MUST: Add `FormFieldDirectives`
- [x] MUST: Add `SecurityDirectives`, possibly Java-model a bit of the related classesupon
- [x] MUST: Add `DebuggingDirectives`
- [x] MUST: Add `FileUploadDirectives`
- [x] MUST: Add `RespondWithDirectives`

Issue #18919, opened 2015-11-12
[akka-http] Java route DSL should follow scala route DSL more closely


jypma:
In order to improve consistency, and to help devs re-use knowledge between Java and Scala, I've created a prototype Java routing DSL that follows the scala routing DSL almost exactly. It's currently in a different [repository](https://github.com/jypma/scala-japi/pull/1) (built on top of the scala routing DSL + java http model).

This alternative has the following advantages over the current Java DSL:
- Routes are completely composable and reusable, without the need to separate routing from implementation using `handleWith`. Every directive just extracts values, and calls an inner route with them.
- Extractions can be made directly as function arguments, without needing a `RequestVal` construct.
- All Scala constructs can be immediately re-exposed in Java 8-friendly API's by adapting function arguments to their equivalents in `java.util.function`.

An example route is given in [RouteSpec](https://github.com/jypma/scala-japi/pull/1/files#diff-6f20875b165bbcb14a14d523d0ffe1f7R219), here's a short extract:
```
        return route(
            path("documents", () ->
                path(UUID(), id -> // type inferred as UUID
                    complete("document " + id)
                )
            ),
            path("people", () ->
                path(name -> // type is String
                    complete("person " + name)
                )
            ),
            path("notreally", () ->
                rejectWith(new MissingFormFieldRejection("always failing"))
            ),
            path("shouldnotfail", () ->
                handleExceptions(xHandler, () ->  // re-using PFBuilder for making ExceptionHandler
                    onSuccess(() -> throwExceptionInFuture(), value -> 
                        complete("never reaches here")
                    )
                )
            ),
            path("cookies", () ->
                param(INT(), "amount", amount ->  // routes can be picked by query parameter type, just like Scala
                    complete("cookies " + amount)
                )
            ),
            path("cookies", () ->
                param(STRING(), "amount", amount -> 
                    complete("cookies (string) " + amount)
                )
            ),
            path("bar", () ->
                extract(CUSTOM_EXTRACT, value -> 
                    complete("bar " + value)
                )
            ),
            path("custom_response", () ->
                complete(HttpResponse.create().withStatus(StatusCodes.ACCEPTED))
            ),
            path("uuid", () -> route(
                put(() ->
                    entityAs(UUID_FROM_BODY, value ->  // actual wrapped Scala Unmarshaller 
                        complete("uuid " + value)
                    )
                ),
                get(() -> {
[...]

PR #20088, opened 2016-03-20
!htt #18919 #19519 Align Java HTTP server DSL with Scala

### LINKED ISSUE #19519: Add formFieldMap , formFieldMultiMap & formFieldSeq Directives for Java DSL
follow-up of #16651

ktoso:
Taking over the new Java DSL over the finish line, based on the excellent work by Jan @jypma from PR #19632.

Resolves #19519 
Resolves #19632 
Allows for basis to reviving better (and complete) JSON support as well.

> This replaces the Java HTTP server DSL with a Java-8 centric one which exposes all scala DSL concepts to be usable from Java, including custom directives, (un)marshallers, rejections, headers, and type safety for path and query parameters. 
> 
> Implemented according to the direction agreed upon with @ktoso in https://github.com/akka/akka/issues/18919 . The class `JavaRouteTest` demoes the Java DSL's capabilities.
> 
> List of things to be done (somewhat in order):
> - [X] Migrated extra directives: `FileAndResourceDirectives`, `CodingDirectives`,`CookieDirectives`,`MiscDirectives`,  `HostDirectives`, `SchemeDirectives`, `WebSocketDirectives`, `CacheConditionDirectives`,  `RangeDirectives`
> - [X] Migrate the existing tests in `akka-http-tests` once the DSL direction is agreed 
> - [X] Fix compile errors in `akka-docs`
> - [X] Fill in the missing directives that haven't been migrated yet 
> - [x] Discuss how to resolve the introduced `TODO` and `FIXME` markers
> - [ ] Replace the current Java `RouteTestKit` (which has its own implementation different from the scala `RouteTestKit`) with a [wrapper](https://github.com/jypma/scala-japi/blob/master/akkaStreamsTest/src/main/scala/com/tradeshift/scalajapi/akka/route/RouteTestKit.scala) that delegates and re-uses the scala one.
> - [ ] Update textual documentation in `akka-docs`
> - [x] MUST: Add `FormFieldDirectives`
> - [x] MUST: Add `SecurityDirectives`, possibly Java-model a bit of the related classesupon
> - [x] MUST: Add `DebuggingDirectives`
> - [x] MUST: Add `FileUploadDirectives`
> - [x] MUST: Add `RespondWithDirectives`


jypma:
`akka-http-jackson-experimental: Binary compatibility check failed!` : that was to be expected, marshalling has changed completely.

But:

`akka-http-testkit: Binary compatibility check failed!` Why is testkit not experimental? The routes underneath that it's testing are experimental as well. And a testkit is needed for the new DSL.

ktoso:
We went a bit too extreme on enabling MiMa recently, don't worry about it.
Also, it will make writing the migration guide a bit easier, so we don't miss anything.

ktoso:
I'll do a consistency spec as next step

ktoso:
FYI, around 44 missing directives still. (though my consistency spec may be off).

akka-http/src/main/scala/akka/http/javadsl/server/directives/BasicDirectives.scala
```diff
@@ -128,6 +131,13 @@ abstract class BasicDirectives {
     D.extractExecutionContext { c ⇒ inner.apply(c).toScala })
 
   /**
+   * Extracts a single value using the given function.
+   */
+  def extract[T](extract: JFunction[RequestContext, T], inner: JFunction[T, Route]): Route = ScalaRoute {
```
jypma: are you actually planning to extract a whole Java requestcontext? I didn't do that, since it would've added a lot of Java work (that part is really scala-only). I couldn't see any real use cases for it, even though it would of course be needed if you want 100% feature parity. It would also require introducing RouteResult on the Java side.

akka-actor-tests/src/test/java/akka/util/ByteStringTest.java
```diff
@@ -20,7 +20,7 @@ public void testCreation() {
   public void testBuilderCreation() {
     final ByteStringBuilder sb = ByteString.createBuilder();
     sb.append(ByteString.fromString("Hello"));
-    sb.append(ByteString.fromString("World"));
+    sb.append(ByteString.fromString("xxx"));
```
ktoso: fun fact, we don't run javatests so this did not fail :O

akka-http/src/main/scala/akka/http/javadsl/server/directives/BasicDirectives.scala
ktoso: > are you actually planning to extract a whole Java requestcontext? I didn't do that, since it would've added a lot of Java work (that part is really scala-only).

As library authors we take all the pain ourselves such that users don't have to :wink:
It takes a special kind of masochism (and / or dedication) to care about the Java API being complete :wink: 

> I couldn't see any real use cases for it, even though it would of course be needed if you want 100% feature parity. It would also require introducing RouteResult on the Java side.

It is a pretty core concept and would be weird to skip now that I'm browsing things.

https://github.com/akka/akka/blob/master/akka-http/src/main/scala/akka/http/javadsl/server/RequestContext.scala

https://github.com/akka/akka/pull/20088/files#diff-fe1dc8b93e367e290481ad720726ef32L17

jypma:
@ktoso I'm back; if I have some time, shall I put a commit somewhere to do an attempt at a Java RequestContext? Do you want all of RouteSettings and the other settings java-lized as well?

ktoso:
That'd be awesome if you could attack the RequestContext :) I'll focus on this PR today as well. 
You can fork off from my PR and PR towards it (inception!)

patriknw:
We decided to not rush this into today's release of 2.4.3. We will complete it as soon as possible and then release another 2.4.x. Thanks for all hard work!

PR #20091, opened 2016-03-20
Fix type of tweets Source

prascuna:
# authors-filter-map won't compile otherwise

### DIFF
--- a/akka-docs/rst/scala/code/docs/stream/TwitterStreamQuickstartDocSpec.scala
+++ b/akka-docs/rst/scala/code/docs/stream/TwitterStreamQuickstartDocSpec.scala
@@ -31,7 +31,10 @@ object TwitterStreamQuickstartDocSpec {
   val akka = Hashtag("#akka")
   //#model
 
-  val tweets = Source(
+  //#tweet-source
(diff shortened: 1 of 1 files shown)


typesafehub-validator:
Hi @prascuna,

[http://www.lightbend.com/contribute/cla](http://www.lightbend.com/contribute/cla)

ktoso:
Thanks a lot, great catch!

(note: this is due to the fact that actual impl is `val tweets = Source ...`, and the one for sample here is just a mock).

We should fix this differently – remove Example0, and write:

```scala
//#tweets-source
  val tweets: Source[Tweet, NotUsed] 
  //#tweets-source
  = Source(
    Tweet(Author("rolandkuhn"), System.currentTimeMillis, "#akka rocks!") ::
 
```

on line 34.

Would you be able to change (or submit a new one) PR doing that @prascuna?

prascuna:
Done. Thanks for the hint!

ktoso:
Looks great, thanks!
Let's let the validator do it's job and merge.

ktoso:
OK TO TEST

PR #20093, opened 2016-03-21
rem #19780: Skip acks during connection handoff

patriknw:
- The problem: ACK that was targeted to an old incarnation
  was sent to the new, restarted, system with same host:port, and
  therefore resulting issues noticed as
  "Error encountered while processing system message acknowledgement buffer: 
  [-1 {}] ack: ACK[0, {}]" when restarting actor system
- The reason:
  1. The endpoint reader was about to send OutgoingAck to parent reader,
     targeted to the old system.
  2. At the same time there is an incoming connection from new system
     that triggered TakeOver in the endpoint writer, i.e. replacing
     the handle to the connection of the new system.
  3. The OutgoingAck is received by the writer, which happily sends it
     to the new handle, the new system.
- The solution: Ignore OutgoingAck during the handoff (TakeOver) process.

### DIFF
--- a/akka-remote/src/main/scala/akka/remote/Endpoint.scala
+++ b/akka-remote/src/main/scala/akka/remote/Endpoint.scala
@@ -809,6 +809,12 @@ private[remote] class EndpointWriter(
 
     case s: Send ⇒
       enqueueInBuffer(s)
+
+    case OutboundAck(_) ⇒
+    // Ignore outgoing acks during take over, since we might have
+    // replaced the handle with a connection to a new, restarted, system
+    // and the ack might be targeted to the old incarnation.
+    // See issue #19780
   }
 
   override def unhandled(message: Any): Unit = message match {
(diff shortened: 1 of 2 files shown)


ktoso:
LVGTM, and thanks for the explanation.
Much better fix than the previous one, you rock!

patriknw:
@drewhk can I merge this?

drewhk:
have not reviewed yet, doing it now.

akka-remote/src/main/scala/akka/remote/Endpoint.scala
```diff
@@ -809,6 +809,12 @@ private[remote] class EndpointWriter(
 
     case s: Send ⇒
       enqueueInBuffer(s)
+
+    case OutboundAck(_) ⇒
```
drewhk: is the case in `unhandled` still relevant after this change?

drewhk:
apart from small question, LGTM.

akka-remote/src/main/scala/akka/remote/Endpoint.scala
patriknw: yes, the case in `unhandled` is for all normal outgoing acks, this is just so that we ignore them during the take over process

patriknw:
Refs #19780

patriknw:
I will backport this

Issue #15556, opened 2014-07-18
Fix "Contents" link in HTML documentation pages


ctoomey:
Per https://groups.google.com/d/msg/akka-user/qZRD-iU_9N8/L4YVmFMdz40J: Something that would improve the usability of the (generally great!) documentation for me at least would be to have the "Contents" link go to the main contents page for whichever language (Scala or Java) documentation is being viewed, rather than going up 2 levels to the page that has links to the Scala and Java documentation table of contents pages, such that I have to click the Scala or Java link to get back to the table of contents page.

E.g., once I've visited http://akka.io/docs/, decided to view the Scala HTML documentation and clicked through to the Scala table of contents page http://doc.akka.io/docs/akka/2.3.4/scala.html, then clicked through to a particular topic (e.g., http://doc.akka.io/docs/akka/2.3.4/scala/actors.html) , when I click the Contents link I should go back to the table of contents page for Scala not to a page that prompts me to choose Scala or Java again.

rkuhn:
@ctoomey sorry for not seeing this earlier: I would definitely find the two-link proposal appealing and I have not yet investigated how to make that work.

richard-imaoka:
I added the above pull request. The change looks like below. Does it look good?

Before:
![content-old](https://cloud.githubusercontent.com/assets/7414320/13913603/cd9a461c-ef8b-11e5-9f8b-f966a60aca02.png)

After:
![content-new](https://cloud.githubusercontent.com/assets/7414320/13913604/cfb22c76-ef8b-11e5-9a68-194c209be04d.png)

ctoomey:
Thanks all for getting this done.

PR #20094, opened 2016-03-21
Add links to go to Java and Scala contents #15556

richardimaoka:
This is to fix the issue #15556.
As far as I tested the change with local sbt package-site build, it worked as expected.

Before:
content-old

After:
content-new

### DIFF
--- a/akka-docs/_sphinx/themes/akka/layout.html
+++ b/akka-docs/_sphinx/themes/akka/layout.html
@@ -114,7 +114,7 @@
                 {%- endif %}
               </li>
               <li>
-                <a href="{{ pathto(master_doc) }}">{{ _('Contents') }}</a>
+                <a href="{{ pathto('java.html', 1) }}">Java Contents</a> <span class="divider">|</span> <a href="{{ pathto('scala.html', 1) }}">Scala Contents</a>
               </li>
               <li>
                 {%- if prev %}


ktoso:
That's a good idea – could we also add some javascript magic to detect in which docs we are (if window.location contains `/scala/` we're in scala) and then show the right link?

richard-imaoka:
Thanks for the feedback. 

[layout.html](https://github.com/akka/akka/blob/master/akka-docs/_sphinx/themes/akka/layout.html#L211L213) already has a JavaScript variable, which I thought I could use:
`var lang //either "scala" or "java"`

However, a lot of pages like [What is Akka](http://doc.akka.io/docs/akka/2.4.2/intro/what-is-akka.html) don't have either "java" or "scala" in their paths. Thus, the `lang` variable = "scala", the default value.

I can tweak the code anyway - on such pages where "java" nor "scala" in their paths , do we show both Scala and Java content links?

patriknw:
LGTM (I'm fine without the javascript magic)

ktoso:
I'm happy to merge this improvement already actually.
Let's thing about the better one with js a bit more – would you be able to make a follow up PR about it?

I think I'd show both if it's neither scala/java in the URL

drewhk:
LGTM and I am also fine without the magic.

ktoso:
Thanks again @richard-imaoka !

Issue #18879, opened 2015-11-06
Missing documentation on timeouts


adamw:
Rumor has it that timeouts are now supported in akka-http client & server (v 2.0-M1)

However, the [docs](http://doc.akka.io/docs/akka-stream-and-http-experimental/2.0-M1/scala/http/client-side/connection-level.html) still claim that:

"Timeouts
Currently Akka HTTP doesn't implement client-side request timeout checking itself as this functionality can be regarded as a more general purpose streaming infrastructure feature. However, akka-stream should soon provide such a feature."

I'm also not sure if the server-side timeouts are documented in any place? (didn't find yet)

ktoso:
Thanks for reporting Adam.
Yes timeouts are in, that note in the docs should be fixed - I must have
missed it.

Rumor has it that timeouts are now supported in akka-http client & server.

However, the docs
<http://doc.akka.io/docs/akka-stream-and-http-experimental/2.0-M1/scala/http/client-side/connection-level.html>
still claim that:

"Timeouts
Currently Akka HTTP doesn't implement client-side request timeout checking
itself as this functionality can be regarded as a more general purpose
streaming infrastructure feature. However, akka-stream should soon provide
such a feature."

I'm also not sure if the server-side timeouts are documented in any place?
(didn't find yet)

—
Reply to this email directly or view it on GitHub
<https://github.com/akka/akka/issues/18879>.

ktoso:
This is solved: we're documenting them here now: http://doc.akka.io/docs/akka/snapshot/scala/http/common/timeouts.html#http-timeouts

ktoso:
code include seems broken, will fix

PR #20099, opened 2016-03-21
=doc fix code include marker for timeouts

ktoso:


ktoso:
Fixes typo found while looking at https://github.com/akka/akka/issues/18879

Issue #18650, opened 2015-10-06
DNS actor fails to be created


drewhk:
Only affects 2.4. According to user report the below problem happens with invalid hostnames.

Caused by: akka.ConfigurationException: configuration problem while creating [akka://providerintegration/system/IO-DNS/inet-address] with router dispatcher [akka.actor.default-dispatcher] and mailbox [akka.actor.default-mailbox] and routee dispatcher [akka.actor.default-dispatcher] and mailbox [akka.actor.default-mailbox]
at akka.actor.LocalActorRefProvider.actorOf(ActorRefProvider.scala:795) ~[akka-actor_2.11-2.4.0.jar:na]
at akka.remote.RemoteActorRefProvider.actorOf(RemoteActorRefProvider.scala:297) ~[akka-remote_2.11-2.4.0.jar:na]
at akka.actor.dungeon.Children$class.makeChild(Children.scala:220) ~[akka-actor_2.11-2.4.0.jar:na]
at akka.actor.dungeon.Children$class.actorOf(Children.scala:38) ~[akka-actor_2.11-2.4.0.jar:na]
at akka.actor.ActorCell.actorOf(ActorCell.scala:373) ~[akka-actor_2.11-2.4.0.jar:na]
at akka.io.SimpleDnsManager.<init>(SimpleDnsManager.scala:15) ~[akka-actor_2.11-2.4.0.jar:na]
... 18 common frames omitted

analytically:
https://gist.github.com/analytically/63381c0adfee9aeb06b9

drewhk:
So the section is actually in the config. Maybe the actor is restarted because of a failure and some weird thing happens?

patriknw:
doesn't that `akka.ConfigurationException` have a `Caused by` that would tell us what is wrong?

analytically:
Yes:

```
Caused by: akka.ConfigurationException: Configuration missing for router [akka://providerintegration/system/IO-DNS/inet-address] in 'akka.actor.deployment' section.
 	at akka.routing.FromConfig.verifyConfig(RouterConfig.scala:311) ~[akka-actor_2.11-2.4.0.jar:na]
 	at akka.routing.RoutedActorRef.<init>(RoutedActorRef.scala:40) ~[akka-actor_2.11-2.4.0.jar:na]
```

kaesler:
Seeing this one too in my test code. FWIW  Here is a simple example:

import scala.concurrent.Future
import scala.concurrent.duration._

[...]

PR #20100, opened 2016-03-21
make DNS actor work with RemoteActorRefProvider, #18650

patriknw:
- The configuration of the pool of the SimpleDnsManager
  is configured in deployment section "/IO-DNS/inet-address"
- We don't really support deployment configuration of system actors
  but here it's used and I don't think we can change that.
- It didn't work when using RemoteActorRefProvider/ClusterActorRefProvider,
  so I fixed it so that the behavior is consistent with the
  LocalActorRefProvider (verified by tests)

### DIFF
--- a/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala
+++ b/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala
@@ -263,9 +263,9 @@ private[akka] class RemoteActorRefProvider(
       val lookup =
         if (lookupDeploy)
           elems.head match {
-            case "user"   ⇒ deployer.lookup(elems.drop(1))
-            case "remote" ⇒ lookupRemotes(elems)
-            case _        ⇒ None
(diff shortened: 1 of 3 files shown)


akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala
```diff
@@ -263,9 +263,9 @@ private[akka] class RemoteActorRefProvider(
       val lookup =
         if (lookupDeploy)
           elems.head match {
-            case "user"   ⇒ deployer.lookup(elems.drop(1))
-            case "remote" ⇒ lookupRemotes(elems)
-            case _        ⇒ None
+            case "user" | "system" ⇒ deployer.lookup(elems.drop(1))
```
patriknw: here is the actual fix

patriknw:
Refs #18650

ktoso:
LGTM, thanks for explaining in detail

drewhk:
LGTM, good catch!

Issue #19693, opened 2016-02-04
FAILED akka.http.impl.engine.server.HttpServerSpec


johanandren:
"should support request timeouts which"
"have a programmatically set timeout handler"

```
java.lang.AssertionError: assertion failed: Expected request 'HTTP/1.1 500 Internal Server Error
 Server: akka-http/test
 Date: XXXX
 Content-Type: text/plain; charset=UTF-8
 Content-Length: 5

 OOPS!' but got 'HTTP/1.1 503 Service Unavailable
 Server: akka-http/test
 Date: XXXX
 Content-Type: text/plain; charset=UTF-8
 Content-Length: 105

 The s'
     at scala.Predef$.assert(Predef.scala:165)
     at akka.http.impl.engine.server.HttpServerTestSetupBase.expectResponseWithWipedDate(HttpServerTestSetupBase.scala:56)
     at akka.http.impl.engine.server.HttpServerSpec$$anonfun$1$$anonfun$apply$mcV$sp$31$$anonfun$apply$mcV$sp$36$$anon$5.<init>(HttpServerSpec.scala:769)
     at akka.http.impl.engine.server.HttpServerSpec$$anonfun$1$$anonfun$apply$mcV$sp$31$$anonfun$apply$mcV$sp$36.apply$mcV$sp(HttpServerSpec.scala:765)
     at akka.http.impl.engine.server.HttpServerSpec$$anonfun$1$$anonfun$apply$mcV$sp$31$$anonfun$apply$mcV$sp$36.apply(HttpServerSpec.scala:765)
     at akka.http.impl.engine.server.HttpServerSpec$$anonfun$1$$anonfun$apply$mcV$sp$31$$anonfun$apply$mcV$sp$36.apply(HttpServerSpec.scala:765)
     at org.scalatest.Transformer$$anonfun$apply$1.apply$mcV$sp(Transformer.scala:22)
```

https://jenkins.akka.io:8498/job/pr-validator-per-commit-jenkins/5086/consoleFull

johanandren:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/3587/consoleFull

drewhk:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/3626/consoleFull

ktoso:
I'll pick this one up today

PR #20101, opened 2016-03-21
htc #19693 fix too small timeout in test

ktoso:
The timeout was too tight here. We need to be able to trigger the `updateHandler()` part before the timeout hits (this is inherently racy with small durations here, and documented as such)


ktoso:
Resolves #19693

patriknw:
LGTM

PR #20104, opened 2016-03-22
Add Flow.foldMerge

jroper:
- This is an example of the type of interface I think would be really useful for using a hub.
- I don't expect it to be merged, certainly not in its current form, it's missing Java API, documentation, and tests are not very comprehensive. But it's also probably a good idea to generalise it, in particular, flattenMerge shares a lot in common with this, and there's probably a lower level public API that could also work with dynamic Sinks similar to GraphStage that could be offered that this could be implemented in.
- My purpose for submitting this is to demonstrate one aspect of the feature set that I think should be there.  My intended use case for this API looks something like this, imagine you have a websocket that serves real time points of interest based on what area of a map users are viewing, the points of interest are published via pubsub, with topics being the id of discrete regions of the globe, and the viewing area is mapped to a list of region ids.  So every time the user scrolls on the map, a new viewing area is sent, and this updates the subscriptions.  This could be implemented using `foldMerge` like so:

``` scala
val pubSub: PubSub[PointOfInterest] = ...
val webSocket: Flow[ViewingArea, PointOfInterest, _] =
  Flow[ViewingArea].foldMerge[RegionId, PointOfInterest](Hub.empty) { (area, hub) =>
    val regions: Seq[RegionId] = calculateRegions(area)
    val unsubscribeRegions = hub.activeSourceIds.filterNot(regions.contains)
    val subscribeRegions = regions.filterNot(hub.activeSourceIds.contains)
    val subscribeRegionSources =
      subscribeRegions.map(regionId => regionId -> pubSub.subscribe(regionId))
    hub ++ subscribeRegionSources -- unsubscribeRegions
  }
```

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowFoldMergeSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowFoldMergeSpec.scala
@@ -1,0 +1,97 @@
+/**
+ * Copyright (C) 2015-2016 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.stream.scaladsl
+
+import akka.stream.testkit.TestPublisher
+import akka.stream.ActorMaterializer
+import akka.testkit.AkkaSpec
+import org.scalatest.exceptions.TestFailedException
+
+import scala.concurrent._
+
+class FlowFoldMergeSpec extends AkkaSpec {
+  implicit val materializer = ActorMaterializer()
+  import system.dispatcher
+
+  def src10(i: Int) = Source(i until (i + 10))
+  def blocked = Source.fromFuture(Promise[Int].future)
+
+  val toSeq = Sink.seq[Int]
+  val toSet: Sink[Int, Future[Set[Int]]] = toSeq.mapMaterializedValue(_.map(_.toSet))
+
+  "A FoldMerge" must {
+
+    "work in the nominal case" in {
+      Source(List(0, 10, 20, 30))
+        .foldMerge[Int, Int](Hub.empty)((id, hub) ⇒ hub + (id -> src10(id)))
+        .runWith(toSet)
+        .futureValue should ===((0 until 40).toSet)
+    }
+
+    "not be held back by one slow stream" in {
+      Source(List(0, 10, -1, 20, 30))
+        .foldMerge[Int, Int](Hub.empty) { (id, hub) ⇒
+          if (id < 0) hub + (id -> blocked) else hub + (id -> src10(id))
+        }.take(40)
+        .runWith(toSet)
+        .futureValue should ===((0 until 40).toSet)
+    }
+
+    "propagate early failure from main stream" in {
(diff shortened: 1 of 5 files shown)


drewhk:
Instead of addSource I would rather see this provide a Sink and Source that can be materialized any number of times. Materializing the Sink it provides will simply count as registering a new input to the hub and materializing the Source will simply count as registering a listener.

jroper:
But with such an API, how would you remove a source from being an input?  It's fine if you control the source, just terminate it, but in my use case above, the source is a pubsub subscription, it's never going to terminate.  What you'd have to do is send it through some `Flow` that you create that can asynchronously receive a signal to both cancel upstream and complete downstream.  Implementing that is not a straight forward thing to do for an end user.

My reason for creating this PR is to ensure that whatever APIs we do provide, this common use case will be simple for end users to implement.  I think removing sources that you don't control will be common, reactive maps is one example of an app that would use it, or any web application where you dynamically register interest in things and then deregister interest in things, for example implementing a messaging system like gitter where you can join and leave channels.

I mean, we created reactive maps to show how simple it is to implement scalable, fault tolerant, reactive systems with our tech.  If the streaming parts of reactive maps aren't simple to implement with Akka streams, if users have to drop down to manually implementing their own GraphStage to do it, then we need to question whether we're providing the right APIs.  I'm not saying that the API that I've provided is the right API, just want to make sure that what we do provide enables this.

drewhk:
> But with such an API, how would you remove a source from being an input?

Simple, the Sink will always materialize to a handle that can be used for removing it from the hub.

jroper:
Ok, so we're looking at something like this then:

```scala
val pubSub: PubSub[PointOfInterest] = ...
val (hubSink, hubSource) = Hub[PointOfInterest]()
val sink = Sink.fold[Map[RegionId, HubHandle], ViewingArea](Map.empty) { (regions, area) =>
  val newRegions: Seq[RegionId] = calculateRegions(area)
  val unsubscribeRegions = regions.keys.filterNot(newRegions.contains)
  val subscribeRegions = newRegions.filterNot(regions.contains)
  val newRegionSubscriptions = subscribeRegions.map { regionId => 
    val source = pubSub.subscribe(regionId)
    val handle = hubSink.runWith(source)
    regionId -> handle
  }.toMap
  regions -- unsubscribeRegions ++ newRegionSubscriptions
}
val webSocket = Flow.fromSinkAndSource(sink, hubSource)
```

That doesn't look too bad.

drewhk:
Yep! And the benefit of this is that if the Source is reusable (i.e. different materializations do different things) then prefusing is possible.

drewhk:
This also allows the other pattern to be implemented (pseudocode):

```scala
def addSource(hub: Hub[T], src: Source[T]): Handle = hub.sink.runWith(src)
```

if for some reason this is preferred.

patriknw:
Closing, and created an [issue](https://github.com/akka/akka/issues/20890) for it until someone can work on it.

Issue #20095, opened 2016-03-21
FAILED: StressSpec


drewhk:
` - must use routers with high throughput *** FAILED *** (11 seconds, 309 milliseconds)
[JVM-1]   java.lang.RuntimeException: barrier failed: routers-done-14`

https://jenkins.akka.io:8498/job/akka-multi-node/1416/consoleFull

patriknw:
This can be related to https://github.com/akka/akka/pull/19897

PR #19897, opened 2016-02-26
=clu #19859 Relaxed constraints on downing old incarnation of rejoining node

Adebski:
Added information to clustering documentation about this "auto-downing" so it is more explicit - hopefully to right file. 

On my local machine some of the tests from `akka-cluster` module were failing (even w/o any changes) so we will have to see what are results from Jenkins.

PR #20105, opened 2016-03-22
harden cluster.StressSpec, #20095

patriknw:
- logs show unexpected unreachable (and downing) during the high
  throughput test
- increasing the failure-detector timeout

I have also verified that the failure should not be related to the change in https://github.com/akka/akka/pull/19897. That code path is not exercised in this part of the test.


patriknw:
Fixes #20095

drewhk:
LGTM

ktoso:
LGTM

Issue #20027, opened 2016-03-14
FAILED: LoggingReceiveSpec


ktoso:
https://jenkins.akka.io:8498/job/akka-multi-node-nightly/3599/consoleFull

Somewhat weird:

```
[info] LoggingReceiveSpec:
[info] A LoggingReceive
[info] - must decorate a Receive (17 milliseconds)
[WARN] [03/13/2016 02:40:36.397] [logging-akka.actor.default-dispatcher-5] [akka://logging/user/$a] unhandled message from Actor[akka://logging/deadLetters]: hallo
[WARN] [03/13/2016 02:40:36.432] [logging-akka.actor.default-dispatcher-5] [akka://logging/user/$$a] unhandled message from Actor[akka://logging/system/testActor-118#-1565089382]: bah
[info] - must be added on Actor if requested (36 milliseconds)
[info] - must not duplicate logging (14 milliseconds)
[info] An Actor
[info] - must log AutoReceiveMessages if requested (17 milliseconds)
[info] - must log Supervision events if requested (17 milliseconds)
[info] - must log DeathWatch events if requested *** FAILED *** (20 milliseconds)
[info]   java.lang.AssertionError: assertion failed: expected: no longer watched by but got unexpected message Debug(akka://lifecycle/user,class akka.actor.LocalActorRefProvider$Guardian,now supervising TestActor[akka://lifecycle/user/$$e])
[info]   at scala.Predef$.assert(Predef.scala:165)
[info]   at akka.testkit.TestKitBase$class.expectMsgPF(TestKit.scala:368)
[info]   at akka.testkit.TestKit.expectMsgPF(TestKit.scala:734)
[info]   at akka.event.LoggingReceiveSpec$$anonfun$3$$anonfun$apply$mcV$sp$6$$anon$6$$anonfun$5.apply$mcV$sp(LoggingReceiveSpec.scala:189)
[info]   at akka.event.LoggingReceiveSpec$$anonfun$3$$anonfun$apply$mcV$sp$6$$anon$6$$anonfun$5.apply(LoggingReceiveSpec.scala:176)
[info]   at akka.event.LoggingReceiveSpec$$anonfun$3$$anonfun$apply$mcV$sp$6$$anon$6$$anonfun$5.apply(LoggingReceiveSpec.scala:176)
[info]   at akka.testkit.TestKitBase$class.within(TestKit.scala:307)
[info]   at akka.testkit.TestKit.within(TestKit.scala:734)
[...]

PR #20107, opened 2016-03-22
harden LoggingReceiveSpec, #20027

patriknw:
- the Debug message was spilled over from previous watch,
  replaced with fishForMessage
- also cleaned up some within limits


patriknw:
Fixes #20027

drewhk:
LGTM

PR #20109, opened 2016-03-22
increase some test timeouts in http-core, #19965

### LINKED ISSUE #19965: failure: too short timeout in RequestParserSpec
Please, please, assume that the build servers are very slow (GC, virtualization, whatnot). 250 ms is way to short. Use 3s if you don't have a good reason to use a shorter timeout.

patriknw:
Fixes #19965


ktoso:
LGTM

drewhk:
LGTM

PR #20112, opened 2016-03-22
make DNS actor work with RemoteActorRefProvider, #18650 (for validation)

patriknw:
- The configuration of the pool of the SimpleDnsManager
  is configured in deployment section "/IO-DNS/inet-address"
- We don't really support deployment configuration of system actors
  but here it's used and I don't think we can change that.
- It didn't work when using RemoteActorRefProvider/ClusterActorRefProvider,
  so I fixed it so that the behavior is consistent with the
  LocalActorRefProvider (verified by tests)

(cherry picked from commit fcd7561b2f25832a32ffbfd12d92d816fb3537e3)


patriknw:
Refs #18650

patriknw:
That was `ClusterShardingLeavingSpec`, that is really not something I want to debug on 2.3. It's unrelated to this PR anyway. I'll rebuild when the ClusterClientSpec has been merged.

patriknw:
PLS BUILD

Issue #2033, opened 2014-02-22
akka-cluster script throws errors


tadasv:
The akka-cluster script isn't working. Any command I try to run I get this output:

    ./akka-cluster localhost 9999 is-available
    Error: Error processing file '/dev/fd/0' - Cannot convert result to a string.
    Akka cluster node is not available on localhost, due to

I tried this on OS X and Ubuntu Precise. I start my jar with

    -Dcom.sun.management.jmxremote.port=9999 -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false

My project deps:

    <dependencies>
            <dependency>
                <groupId>com.typesafe</groupId>
                <artifactId>config</artifactId>
                <version>1.2.0</version>
            </dependency>
            <dependency>
                <groupId>com.typesafe.akka</groupId>
                <artifactId>akka-cluster_2.10</artifactId>
                <version>2.2.3</version>
            </dependency>
            <dependency>
                <groupId>com.typesafe.akka</groupId>
                <artifactId>akka-remote_2.10</artifactId>
                <version>2.2.3</version>
            </dependency>
            <dependency>
                <groupId>org.scala-lang</groupId>
                <artifactId>scala-library</artifactId>
                <version>2.10.3</version>
            </dependency>
        </dependencies>

Tolsi:
I have the same problem, guys, how you solved it? @tadasv @leonardschneider

patriknw:
It looks like the problem exists in the latest 2.3.11 zip.
You can copy the script from http://downloads.typesafe.com/akka/akka_2.10-2.3.10.zip

Tolsi:
@patriknw nothing changed

patriknw:
Ah, I thought this was the ticket about the wrong line endings.

ewegene:
I am seeing a similar problem with akka-2.4.0.

After telling it to leave the cluster:

    $ akka-cluster localhost 9999 leave akka.tcp://ClusterSystem@10.0.0.100:2551
    Scheduling akka.tcp://ClusterSystem@10.0.0.100:2551 to LEAVE cluster

I see that it takes a couple seconds to leave the cluster, if I check it immediately it returns true:

    $ akka-cluster localhost 9999 is-available
    Checking if member node on localhost is AVAILABLE
    true

And eventually it will start throwing up that error (usually within a couple/few seconds within the leave command)

    $ akka-cluster localhost 9999 is-available
[...]

PR #20117, opened 2016-03-22
Improve error when Cluster MBean is gone #2033

raboof:
When the Cluster JMX MBean is missing jmxsh-R5.jar produces 'Cannot convert
result to a string' as an error message.

Recognize this error and give a more meaningful one back to the user.

### DIFF
--- a/akka-kernel/src/main/dist/bin/akka-cluster
+++ b/akka-kernel/src/main/dist/bin/akka-cluster
@@ -33,9 +33,13 @@ function get() {
 }
 
 function ensureNodeIsRunningAndAvailable {
-    REPLY=$(get Available) # redirects STDERR to STDOUT before capturing it
+    REPLY=$(get Available 2>&1) # redirects STDERR to STDOUT before capturing it
     if [[ "$REPLY" != *true ]]; then
-        echo "Akka cluster node is not available on $HOST, due to $REPLY"
+        if [[ "$REPLY" == *"Cannot convert result to a string." ]]; then
+          echo "Akka cluster MBean is not available on $HOST:$PORT"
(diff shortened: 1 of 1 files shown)


akka-kernel/src/main/dist/bin/akka-cluster
```diff
@@ -33,9 +33,13 @@ function get() {
 }
 
 function ensureNodeIsRunningAndAvailable {
-    REPLY=$(get Available) # redirects STDERR to STDOUT before capturing it
+    REPLY=$(get Available 2>&1) # redirects STDERR to STDOUT before capturing it
     if [[ "$REPLY" != *true ]]; then
-        echo "Akka cluster node is not available on $HOST, due to $REPLY"
+        if [[ "$REPLY" == *"Cannot convert result to a string." ]]; then
```
drewhk: is this check robust? Or can it mask errors that has nothing to do with MBean not aviailable?

raboof: That'd have to mean the `Available` field of the `akka/Cluster` MBean would have been changed to something that cannot be converted to a string.

I'm not sure whether that is even possible, but if it is, it'd require code changes to the MBean, and I don't know why we'd make such a backwards-incompatible change.

patriknw: sounds alright

patriknw:
LGTM, thanks!

patriknw:
pr validation will not check anything here

patriknw:
thanks for contributing @raboof

patriknw:
Refs #2033

Issue #19039, opened 2015-11-28
add Source.cycle


hepin1989:
There is a repeat def `Source.repeat[T](element: T)` and I think a `cycle` should be handy too ,as `Source.cycle[T](T:*)` .

`Sourcel.cycle(1,2,3)` will emit `1,2,3,1,2,3,1,2,3...`

hepin1989:
I think I could implement that via `Source(() => Iterator.continually(Seq(1,2,3).flatten)`

ktoso:
I was about to open a ticket about it yesterday :)
+1, I think it's very useful.

viktorklang:
Or: `Source.repeat(Seq(1,2,3)).mapConcat(identity)`

ktoso:
Random thought... I seem to remember Scala's Range was "pretty heavy" though can't recall completely in what way...? So since `javadsl.Source.cycle(1, 10)` wants to be the equivalent of `scaladsl.

ktoso:
mis-clicked, sorry.

viktorklang:
I don't want to be the person who is that person, but how about we name it `repeatAll` so that it autocompletes nicely?

ktoso:
cycle seems to be Rx wording. I don't mind either of them.

viktorklang:
I'm fine with `cycle` too, I was more concerned by the user.

rkuhn:
Cycling is what I do sometimes (but not often enough) :-)

agolubev:
is it similar to #16454?

PR #20125, opened 2016-03-23
#19039: Added cycle method to Source class.

kelebra:


drewhk:
OK TO TEST

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -204,6 +204,14 @@ object Source {
     })
 
   /**
+   * Create [[Source]] that will continually produce given elements in specified order.
+   */
+  def cycle[T](elements: T*): Source[T, NotUsed] = {
```
drewhk: Needs JavaDSL counterpart

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
```diff
@@ -272,6 +272,20 @@ class SourceSpec extends AkkaSpec with DefaultTimeout {
...
+  "Cycle Source" must {
+
+    "continuously generate the same sequence" in {
+      val expected = Seq(1, 2, 3, 1, 2, 3, 1, 2, 3)
+      Source.cycle(1, 2, 3).grouped(9).runWith(Sink.head).futureValue should ===(expected)
+    }
+
+    "reject creation from empty sequence" in {
+      intercept[IllegalArgumentException] {
+        Source.cycle()
```
ktoso: Let's make this not compile instead of throwing.
Signature should be:

```scala
def cycle[T](first: T, rest: T*): Soruce[T, NotUsed] = ???
// cycle(1,2,3) // ok
// cycle(1) // ok
// repeat(1) should be implemented in terms of cycle(1) I guess?
// cycle() // won't compile
```

drewhk: That is harder to feed with a `Seq` though. I would probably prefer a `Seq` version.

ktoso: Hm, yeah that's true as well, can keep the Seq then.

// note: please immutable.Seq

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
kelebra: Sure, will add it today, thanks.

kelebra: Added in the second commit.

kelebra:
Just a small question: do I need to squash these 3 commits into one?

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -91,6 +91,24 @@ object Source {
...
+   * data.add(1);
+   * data.add(2);
+   * data.add(3);
+   * Source.cycle(data);
+   * }}}
+   *
+   * Start a new 'cycled' `Source` from the given elements. The producer stream of elements
+   * will continue infinitely by repeating the same sequence of elements provided during construction.
+   */
+  def cycle[O](elements: java.util.List[O]): javadsl.Source[O, NotUsed] =
```
ktoso: That's not 1:1 with the scala one though - why not `@varargs def cycle[O](elements: O*)`?

ktoso:
Added one comment on javadsl, yes, please squash afterwards

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
kelebra: Was confused by the test which checks types relation between scaladsl and javadsl and this particular mapping: `Seq -> java.util.List`.
Will update javadsl signature to suggested version.

kelebra:
Squashed into one commit as agreed previously.
Sorry for mess around a simple change, still learning akka internals and am pretty new to open source contribution as such :)

viktorklang:
Isn't this: Source.repeat(immutableIterable).mapConcat(identity)?

kelebra:
Yes, it is and you actually proposed this implementation in issue discussion itself (#19039) but i decided to go with the one more framework independent. Do you want me to change it to `Source.repeat(Seq(1,2,3)).mapConcat(identity)`?

viktorklang:
@kelebra I think your implementation is fine. You'll want to bestow it with it's own DefaultAttributes though.

kelebra:
Added `DefaultAttributes.cycledSource` as was suggested

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
drewhk: What was the conclusion here? I thought that we agreed on the Seq version. Or if the varargs was chosen then I agree with @ktoso that the first element should be enforced statically by the argument declaration.

viktorklang: IMO shouldn't be varargs because it'll have to be copied to an immutable collection immediately in that case.

drewhk: I agree, but maybe the Java DSL should have a varargs version as it is much harder to create simple sequences in java.

viktorklang: Perhaps the easiest solution is to have it use `() => Iterator[T]`?

patriknw: I agree that `() => Iterator[T]` would be better. We don't have any other Source from varargs. It also adds the interesting twist that the factory will be invoked for each round.

patriknw:
Refs #19039

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
kelebra: To my best understanding you are suggesting to change the java dsl implementation to create a new iterator out of array (varargs) and use not `cycle` from scaladsl but `fromIterator`? 
I would also stick to the varargs version of method in javadsl.

Regarding the option of having `def cycle[T](first: T, rest: T*): Source[T, NotUsed] = ???`: I do not think it looks good from client perspective and I would rather stick to the `def cycle[T](elements: T*): Source[T, NotUsed] = ???`

patriknw: No, as I understand it, the suggestion from @viktorklang is to not use varargs, but:

    def cycle[O](f: function.Creator[java.util.Iterator[O]]): javadsl.Source[O, NotUsed]

similar for scaladsl.

However, I might have misunderstood the purpose and the use case motivates varargs?

viktorklang: @patriknw yes, exactly. I think it is better to try to harmonize the approach we take to these things.

>Regarding the option of having def cycle[T](first: T, rest: T*): Source[T, NotUsed] = ???: I do not think it looks good from client perspective and I would rather stick to the def cycle[T](elements: T*): Source[T, NotUsed] = ???

@kelebra I don't understand your argument here, would you mind elaborating?

kelebra: @viktorklang, @patriknw thanks for specifying the signature, will commit changes today.

@viktorklang, Well, as an end user I do not want actually to think about head/tail element(s). I just want to pass immutable sequence and worry about nothing :)

kelebra: @viktorklang Nevermind, commented actually the wrong thing.

Guys so as far as I understand you are fine with the following signatures, right?

**Scala**:
`def cycle[T](elements: T*): Source[T, NotUsed]`
**Java**:
`def cycle[O](f: function.Creator[java.util.Iterator[O]]): javadsl.Source[O, NotUsed]`

viktorklang: I think it should have the same-ish API for both, it () => Iterator[T] for Scala.

Reason: elements: T* would still need to be made into an immutable collection.

kelebra: @viktorklang Deal, will commit changes later today.

Just to finalize the signatures:
**Scala**:
`def cycle[T](() => Iterator[T]): Source[T, NotUsed]`
**Java**:
`def cycle[O](f: function.Creator[java.util.Iterator[O]]): javadsl.Source[O, NotUsed]`

viktorklang: @kelebra sounds terrific, thank you!

kelebra:
Done all the changes suggested in the review. Now signatures are the ones agreed before. Anything else needs to be done?

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
```diff
@@ -272,6 +272,14 @@ class SourceSpec extends AkkaSpec with DefaultTimeout {
     }
   }
 
+  "Cycle Source" must {
+
+    "continuously generate the same sequence" in {
+      val expected = Seq(1, 2, 3, 1, 2, 3, 1, 2, 3)
+      Source.cycle(() ⇒ List(1, 2, 3).iterator).grouped(9).runWith(Sink.head).futureValue should ===(expected)
```
viktorklang: What happens if the iterator is empty?

ktoso: ping here, let's add a test

ktoso:
This stage must be added to documentation:

- https://github.com/akka/akka/blob/master/akka-docs/rst/scala/stream/stages-overview.rst
- https://github.com/akka/akka/blob/master/akka-docs/rst/java/stream/stages-overview.rst

akka-stream-tests/src/test/scala/akka/stream/scaladsl/SourceSpec.scala
kelebra: Added as suggested

kelebra:
Added documentation and behavior for case of empty iterator along with test.

kelebra:
Guys, is there anything else that need to be done for this one to be merged?

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -204,6 +204,18 @@ object Source {
     })
 
   /**
+   * Create [[Source]] that will continually produce given elements in specified order.
+   *
+   * Start a new 'cycled' `Source` from the given elements. The producer stream of elements
+   * will continue infinitely by repeating the same sequence of elements provided during construction.
+   */
+  def cycle[T](f: () ⇒ Iterator[T]): Source[T, NotUsed] = {
+    require(f().nonEmpty, "Provided iterator should not be empty")
```
patriknw: @kelebra I don't think it's expected that `f` is applied here, we should not do that before materialization, i.e. the check should be later and fail the stream instead

kelebra: Actually giving it another thought: why should stream fail at all? `fromIterator` does not fail the stream in case you pass empty `Iterator` object, stream just finishes. Maybe the same should happen in this case?

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -204,6 +204,17 @@ object Source {
     })
 
   /**
+   * Create [[Source]] that will continually produce given elements in specified order.
+   *
+   * Start a new 'cycled' `Source` from the given elements. The producer stream of elements
+   * will continue infinitely by repeating the same sequence of elements provided during construction.
+   */
+  def cycle[T](f: () ⇒ Iterator[T]): Source[T, NotUsed] = fromIterator(() ⇒ {
+    if (f().isEmpty) Iterator.empty
```
patriknw: ok, I guess it's a matter of how we define it.

Reference docs and scaladocs should be updated regarding empty iterator.

Do we want `f` to be evaluated for each round or only once? Now it looks like a mix, it is checked for empty once but called each time by `Iterator.continually`.

kelebra: Will do the update on docs regarding empty iterator case. Should I update `fromIterator` docs for the same case?

I would do the following:
```
  def cycle[T](f: () ⇒ Iterator[T]): Source[T, NotUsed] = {
    val iterator = f()
    val cycledArg = if (iterator.nonEmpty) Iterator.continually(iterator).flatten else iterator
    fromIterator(() ⇒ cycledArg).withAttributes(DefaultAttributes.cycledSource)
  }
```
But unfortunately this results infinite loop in scala internals when `iterator` runs out of elements.

So it looks like that the only way to 'cycle' iterator is to do it with repeated evaluation of `f`: `Iterator.continually(f()).flatten`.

Now regarding the empty case. Again you cannot provide empty iterator to `Iterator.continually` method (see [stack overflow question](http://stackoverflow.com/questions/36818050/iterator-continually-cannot-evaluate-expression-in-scala-repl)), it looks like that we need to do 'flattening' by ourselves returning empty iterator or `f()`: `if (f().isEmpty) Iterator.empty` or `vol iterator = f() if(iterator.isEmpty) iterator else Iterator.continually(f()).flatten`.

All said above leads me to the following implementation:
```
  def cycle[T](f: () ⇒ Iterator[T]): Source[T, NotUsed] = {
    val iterator = if (f().isEmpty) Iterator.empty else Iterator.continually(f()).flatten
    fromIterator(() ⇒ iterator).withAttributes(DefaultAttributes.cycledSource)
  }
```
Please let me know if you agree with it

patriknw: That is still evaluating `f` several times and in the end there is really no way around that since `Iterator.continually` will evaluate it each time, and `f` doesn't have to return same thing every time and any such evaluation that is `empty` is bad. I think we should fail for empty, i.e. 
```scala
val iter = Iterator.continually {val i = f(); if (i.isEmpty) throw new NoSuchElementException("empty iterator") else i}.flatten
```

kelebra: Do agree. Let me update implementation and docs today.

kelebra: Implementation updated along with docs.

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -204,6 +204,17 @@ object Source {
     })
 
   /**
+   * Create [[Source]] that will continually produce given elements in specified order.
+   *
+   * Start a new 'cycled' `Source` from the given elements. The producer stream of elements
+   * will continue infinitely by repeating the same sequence of elements provided during construction.
```
patriknw: sorry for nitpicking, but this comment is not completely true, it's not necessarily the same sequence of elements. For each round the `f` is evaluated and it doesn't have to return the same thing each time

kelebra: Thanks for pointing that out! I did the update to both documentation and scaladoc.

akka-stream/src/main/scala/akka/stream/javadsl/Source.scala
```diff
@@ -91,6 +91,20 @@ object Source {
...
   /**
+   * Helper to create 'cycled' [[Source]] from iterator provider.
+   * Example usage:
+   *
+   * {{{
+   * Source.cycle(() -> Arrays.asList(1, 2, 3).iterator());
+   * }}}
+   *
+   * Start a new 'cycled' `Source` from the given elements. The producer stream of elements
+   * will continue infinitely by repeating the sequence of elements provided by function parameter.
```
kelebra: @patriknw updated as was suggested before

patriknw:
Thanks a lot for contributing. Looking great now!

patriknw:
Refs #19039

kelebra:
Thanks a lot for help and opportunity to contribute :+1:

PR #20128, opened 2016-03-23
htc #19777 don't error on possible truncation errors

### LINKED ISSUE #19777: akka-http throws OutputTruncationException if the server closes the connection fast enough
There have been other reports around `OutputTruncationException` thrown when the server closes the connection fast enough, but I don't think those have such a simple reproducible case.

Akka HTTP 2.4.2-RC3.

For this test, you must run a fast-closing server, such as:

``` bash
% socat tcp4-listen:5985,reuseaddr,fork exec:/tmp/answer.sh"
```

with a HTTP response in `/tmp/answer.sh` reading the 4 lines (including blank line) sent by Akka in the request (`GET`, `Host`, `User-Agent`, separator) and returns the response before closing the socket:

``` bash
#! /bin/bash

for i in 1 2 3 4; do
  IFS='' read -r line
done
printf "HTTP/1.1 200 OK\r\nConnection: close\r\nContent-Type: application/octet-stream\r\nContent-Length: 0\r\n\r\n"
```

I believe this behavior (dropping the connection very fast) is allowed as per RFC 7230. And the CouchDB database does it sometimes.

This test case will correctly print and display the request status (`Status: 200 OK`) but will signal an error in the log:

```
[ERROR] [02/15/2016 09:50:33.771] [default-akka.actor.default-dispatcher-2] [akka://default/user/StreamSupervisor-0/flow-1-0-unknown-operation] Error in stage [One2OneBidi]: Inner stream finished before inputs completed. Outputs might have been truncated. (akka.http.impl.util.One2OneBidiFlow$OutputTruncationException$)
```

``` scala
import akka.actor.ActorSystem
import akka.http.scaladsl.model.{HttpMethods, HttpRequest}

import scala.concurrent.Await

object ErrorTruncation extends App {

  implicit val executionContext = system.dispatcher

  val request  = HttpRequest(HttpMethods.GET, "http://localhost:5985/")
  val status = Await.result(Http().singleRequest(request), 5.seconds).status
  println(s"Status: $status")

  Http().shutdownAllConnectionPools().foreach(_ => system.terminate())
}
```

Note: reading the empty body entity does not change the error sent by Akka, so I've not included it here. A non-empty body reads just fine as well, but the error message will still be present in the log.

ktoso:
Resolves #19777

Though I am somewhat iffy about it - please review @drewhk


akka-http-core/src/main/scala/akka/http/impl/util/One2OneBidiFlow.scala
```diff
@@ -87,8 +91,13 @@ private[http] object One2OneBidiFlow {
           } else throw new UnexpectedOutputException(element)
         }
         override def onUpstreamFinish(): Unit = {
-          if (pending == 0 && isClosed(inIn) && !innerFlowCancelled) complete(outOut)
-          else throw OutputTruncationException
+          if (pending > 0 || innerFlowCancelled) {
+            if (materializer.isInstanceOf[ActorMaterializer])
+              ActorMaterializer.downcast(materializer).logger.debug(
+                "Inner stream finished before inputs completed. " +
```
drewhk: unfortunately this is not reliable, this is exactly the issue. Eliminate this whole dance and only keep the check that the emitted elements must never be more than the consumed ones.

patriknw:
looks like is not ready, and will slip the release

akka-http-core/src/main/scala/akka/http/impl/util/One2OneBidiFlow.scala
drewhk: Hmm, thinking about it, isn't the case that `(pending > 0 && innerFlowCompleted)` is definitely a truncation (and the only one that is reliable)? Since that means that we passed something to the inner flow which will never return as it has completed. Cancellation means nothing since we already passed it into the inner flow when we increased pending see https://github.com/akka/akka/pull/20128/files#diff-7572c326b2938b0c69732c6f12f643f2R65

drewhk:
We should try to finish this though, it is a rather annoying issue right now.

ktoso:
@drewhk will take it over, closing

Issue #15990, opened 2014-09-29
Add automatic redirection handling


jrudolph:
The high-level client-side API should support some form of configurable redirection handling.

Several things need to be taken into account:
 *  dealing with headers to be used for the redirected request. This has security implications.
   * user-agent [spray-user](https://groups.google.com/forum/#!topic/spray-user/1F1cSBeYPh8)
   * cookies
   * authorization
 * redirection not allowed for non-idempotent request (already implemented in spray) for some redirection status codes
 * which HostConnector to use for cross-domain redirections

Before starting on this, it would make sense to go through the mailing list / spray tickets to find further cases we might need to support.

A first solution could only support redirections to the same domain, where keeping certain headers may be safer (but: cookies with path constraints set).

Maybe also consider how browsers deal with redirections.

This is part of the bigger initiative to support a high-level HTTP client interface as tracked as #16856.

/cc @sirthias

huntc:
I've just stumbled across the need for this and will have to of course hand craft it. I just thought that you'd like to know of at least one person who needs this functionality.

francisdb:
Another case to handle would be redirect loops

CptnKirk:
Also +1 for this feature.  Did tracking of this issue get lost?  Was 1.0-RC4, now I see it has no milestone. Yet I don't see that anyone has modified the milestone since sirthias on Jun 24.

coreyauger:
Also +1 from me.

ktoso:
Logging a +1 :point_up: [February 24, 2016 3:11 PM](https://gitter.im/akka/akka?at=56cdba24bfb1cf9a7ff2badb)

We'd welcome contributions of this feature, anyone interested in helping out? :-)

RomanIakovlev:
I'd like to try to help with this, since I need it in my project. However, I'm totally new to akka source code, so I'd definitely need some initial guidance. What would be the best channel to ask questions related to this?

ktoso:
Thank a lot for offering help on that :-)
Best discussed here actually as it's nice and async, or gitter (however this week the team is mostly offline, planning the next months of development, so we may be a bit slow to respond - sorry about that).

RomanIakovlev:
Thanks, I'd be happy to help. Just not sure if I'm up to this task, but one never knows unless one tries. :)

Ok, here are the initial questions.

[...]

Issue #16856, opened 2015-02-11
High-level HTTP client support


jrudolph:
#It would be convenient if akka-http would provide common high-level HTTP functionality similar to what a browser supports in its backend. This is a collection of all the things that would belong into this category. This is a metaticket and each subfeature has its own ticket.

 - [ ] connection pooling and configurable request queuing (provided by host-level API #15681)
 - [x] request-level API (#15906)
 - [ ] automatic redirection (#15990)
 - [x] automatic retry attempts (#16852)
 - [ ] try different IP addresses on retries (#16827)
 - [ ] HTTP proxy support (for HTTP: #16853, for HTTPS: #16153)
 - [ ] authentication data store (#16854)
 - [ ] cookie store (#16855)
 - [ ] response routing DSL (#15909)
 - [ ] request building should be polished and documented (https://github.com/akka/akka/issues/18550)
 - [ ] optional: response decompression (https://github.com/akka/akka/issues/16813)
 - [ ] optional: Add support for OAuth authentication #16857

kanerogers:
Awesome. Looking forward to this.

jvican:
Me too. Great and necessary features!
If you need any help, let me know and I will help.

rkuhn:
Thanks for your offer to help, the core team will likely not get around to these anytime soon.

ktoso:
Somewhat related to https://github.com/akka/akka/issues/18550

ktoso:
For reference when designing the client side:

I bumped into very nice blog post which compared different Scala HTTP clients, among others Spray and Play: 
https://www.implicitdef.com/2015/11/19/comparing-scala-http-client-libraries.html
I agree with the article that "dead simple" should be a focus of the Client Side, and the awesome stuff should be additions.

francisdb:
Could be a joined effort together with
https://github.com/playframework/playframework/issues/4297 ?

ktoso:
Thanks for cross referencing to that ticket – I've commented there as well as you'll notice :-) 
We definitely want to share bits of code with WS, we'll see how exactly we'll be able to do that (it's SSL/TLS handling is excellent).

hepin1989:
I just encounter this problem too,and then just try the AHC.A higher level like play-ws would be really helpful,the current `singleRequest` one is not that handy.for the Play-WS,I don't think that could be split out,but for later,maybe we could have two implement of play-ws,one for the AHC one and one for the AKKA based one,anyway there is a WSAPI.

[...]

PR #20135, opened 2016-03-24
[WIP] #15990 - HTTP client auto redirects

RomanIakovlev:
This PR adds support for Akka HTTP client to automatically follow the redirects, if it's allowed by RFC. In its current form, it supports redirects to the same host:port, but not the cross-origin redirects, because it re-uses the same underlying `OutgoingConnectionBlueprint` instance as the one being requested by the HTTP client API user originally.

It's not yet fully tested because of the aforementioned limitation (in other words, I don't know if it makes sense to proceed with the current approach), however some tests are available in the `akka.http.impl.engine.client.RedirectTestSpec`. Most basic cases are covered. Particularly, the last one (marked as ignored) highlights the problem with cross-origin redirects.

The settings API (scaladsl/javadsl) and logic related to change of HTTP methods based on the redirect's status code might need some more work.

@ktoso please review.


typesafehub-validator:
Hi @RomanIakovlev,

[http://www.lightbend.com/contribute/cla](http://www.lightbend.com/contribute/cla)

ktoso:
OK TO TEST

akka-http-core/src/main/resources/reference.conf
```diff
@@ -282,14 +307,14 @@ akka.http {
...
-    max-uri-length             = 2k
-    max-method-length          = 16
+    max-uri-length = 2k
+    max-method-length = 16
     max-response-reason-length = 64
-    max-header-name-length     = 64
-    max-header-value-length    = 8k
-    max-header-count           = 64
-    max-chunk-ext-length       = 256
-    max-chunk-size             = 1m
```
ktoso: Please don't :)

RomanIakovlev: Oops, using auto-format on this file was not the best idea... :) Fixed.

akka-http-core/src/main/resources/reference.conf
```diff
@@ -170,6 +170,31 @@ akka.http {
     parsing {
       # no overrides by default, see `akka.http.parsing` for default values
     }
+
+    # Automatic redirect settings
+    redirect {
+      # `same-origin` and `cross-origin` contain the same set of settings for
```
ktoso: excellent, thanks - separating config like this looks very good.

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
```diff
@@ -0,0 +1,291 @@
...
+    else false
+  }
+
+  trait OpOutPrinter[F, _ <: Outlet[F]] {
+    def print(t: F): Unit
+  }
+
+  trait OpInPrinter[T <: Inlet[_]] {
+    def print(): Unit
+  }
```
ktoso: I'll see later if we really need those but in general let's not add ad-hoc in-middle-of-other-class type classes for printing :)

ktoso: (Can stay here while we're discussing the PR still, no worries)

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
```diff
@@ -0,0 +1,291 @@
...
+class RedirectSupportStage(settings: ClientAutoRedirectSettings, redirectMapper: RedirectMapper) extends GraphStage[BidiShape[HttpRequest, HttpRequest, HttpResponse, HttpResponse]] {
+  /**
+    * in1 is where the original messages come from
+    */
+  val requestIn: Inlet[HttpRequest] = Inlet("Input1")
+
+  /**
+    * in2 is the responses to original messages. If it's a redirect, its not forwarded upstream, but goes downstream
+    */
+  val responseIn: Inlet[HttpResponse] = Inlet("Input2")
```
ktoso: Please use this pattern for port names: `RedirectSupportStage.responseIn` / `RedirectSupportStage.requestOut` etc

ktoso: (If something fails, the name is only printed - so it's more helpful to include names like this, we're trying to standardise on this pattern now)

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
```diff
@@ -0,0 +1,291 @@
...
+      }
+
+      def doPull[T](inlet: Inlet[T])(implicit ev: OpInPrinter[Inlet[T]]) = {
+        if (!isClosed(inlet)) {
+          pull(inlet)
+        }
+        ev.print()
+      }
+
+      var inFlight = Vector.empty[HttpRequest]
```
ktoso: It doesn't seem like we're limiting the max nr of in flight messages; would that be a good idea to add here?

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
```diff
@@ -0,0 +1,291 @@
...
+        override def onPush(): Unit = {
+          printy("responseIn: onPush")
+          val response = grab(responseIn)
+          val maybeRedirect = redirectMapper(settings, inFlight.head, response)
+          if (maybeRedirect.isDefined) {
+            val redirect = maybeRedirect.get
+            printy(loopDetector.toString())
+            val loops = loopDetector(inFlight.head)
+            if (loops.contains(redirect.uri)) {
+              throw InfiniteRedirectLoopException
```
ktoso: I forgot which is better - throw or fail the stage? 
The answer depends on if we want to log this failure... I guess we do.
// cc @drewhk

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
```diff
@@ -0,0 +1,291 @@
...
+import akka.http.scaladsl.settings.ClientAutoRedirectSettings.HeadersForwardMode.{All, Only, Zero}
+import akka.stream._
+import akka.stream.scaladsl.BidiFlow
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+
+import scala.language.higherKinds
+import scala.util.control.NoStackTrace
+
+object RedirectSupportStage {
+  type RedirectMapper = (ClientAutoRedirectSettings, HttpRequest, HttpResponse) => Option[HttpRequest]
```
ktoso: Let's make it a normal class instead perhaps?
Note that you could: 

```
case class RedirectMapper(a: ClientAutoRedirectSettings, b: HttpRequest, c: HttpResponse) 
extends (ClientAutoRedirectSettings, HttpRequest, HttpResponse) => Option[HttpRequest] {
 def apply...
```

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
```diff
@@ -0,0 +1,291 @@
...
+        /**
+          * Called when the input port has a new element available. The actual element can be retrieved via the
+          * [[GraphStageLogic.grab()]] method.
+          */
+        override def onPush(): Unit = {
+          printy("responseIn: onPush")
+          val response = grab(responseIn)
+          val maybeRedirect = redirectMapper(settings, inFlight.head, response)
+          if (maybeRedirect.isDefined) {
+            val redirect = maybeRedirect.get
```
ktoso: instead of `isDefined + get` please use `maybeRedirect match { case Some(...`

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
ktoso: hm, also - would it be useful to include the url in the exception message?

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
```diff
@@ -0,0 +1,291 @@
...
+          * be called for this port.
+          */
+        override def onDownstreamFinish(): Unit = {
+          printy("responseOut: onDownstreamFinish")
+          cancel(responseIn)
+        }
+      })
+
+    }
+}
```
ktoso: note to self: re-read again

akka-http-core/src/main/scala/akka/http/impl/settings/ClientAutoRedirectSettingsImpl.scala
```diff
@@ -0,0 +1,25 @@
...
+object ClientAutoRedirectSettingsImpl extends SettingsCompanion[ClientAutoRedirectSettingsImpl]("akka.http.client.redirect") {
+
+  override def fromSubConfig(root: Config, inner: Config): ClientAutoRedirectSettingsImpl = {
+    val sameOriginConfig = inner.withFallback(root.getConfig(prefix)).getConfig("same-origin")
+    val crossOriginConfig = inner.withFallback(root.getConfig(prefix)).getConfig("cross-origin")
+    ClientAutoRedirectSettingsImpl(
+      ClientAutoRedirectSettingsItemImpl.fromSubConfig(root, sameOriginConfig),
+      ClientAutoRedirectSettingsItemImpl.fromSubConfig(root, crossOriginConfig))
+  }
+}
```
ktoso: great, thanks

akka-http-core/src/main/scala/akka/http/impl/settings/ClientAutoRedirectSettingsImpl.scala
```diff
@@ -0,0 +1,25 @@
...
+import akka.http.impl.util.SettingsCompanion
+import akka.http.javadsl.settings.ClientAutoRedirectSettingsItem
+import com.typesafe.config.Config
+
+/** INTERNAL API */
+private[akka] final case class ClientAutoRedirectSettingsImpl(
+  sameOrigin: ClientAutoRedirectSettingsItem,
+  crossOrigin: ClientAutoRedirectSettingsItem) extends akka.http.scaladsl.settings.ClientAutoRedirectSettings
+
+object ClientAutoRedirectSettingsImpl extends SettingsCompanion[ClientAutoRedirectSettingsImpl]("akka.http.client.redirect") {
```
ktoso: Hm, companion with Impl in name? Sure

akka-http-core/src/main/scala/akka/http/impl/settings/ClientConnectionSettingsImpl.scala
```diff
@@ -24,7 +24,8 @@ private[akka] final case class ClientConnectionSettingsImpl(
   requestHeaderSizeHint: Int,
   websocketRandomFactory: () ⇒ Random,
   socketOptions: immutable.Seq[SocketOption],
-  parserSettings: ParserSettings)
+  parserSettings: ParserSettings,
+  redirectSettings: akka.http.scaladsl.settings.ClientAutoRedirectSettings)
```
ktoso: AFAICS keeping binary compatibility will force us to write a constructor the same as the previous one with a default value for the new parameter here.

ktoso: (you can check that by running `mimaReportBinaryIssues`

akka-http-core/src/main/scala/akka/http/impl/util/JavaMapping.scala
```diff
@@ -172,6 +172,7 @@ private[http] object JavaMapping {
   implicit object ClientConnectionSettings extends Inherited[js.ClientConnectionSettings, akka.http.scaladsl.settings.ClientConnectionSettings]
   implicit object ConnectionPoolSettings extends Inherited[js.ConnectionPoolSettings, akka.http.scaladsl.settings.ConnectionPoolSettings]
   implicit object ParserSettings extends Inherited[js.ParserSettings, akka.http.scaladsl.settings.ParserSettings]
+  implicit object ClientAutoRedirectSettings extends Inherited[js.ClientAutoRedirectSettings, akka.http.scaladsl.settings.ClientAutoRedirectSettings]
```
ktoso: :+1:

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -258,7 +258,8 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
   def clientLayer(hostHeader: Host,
                   settings: ClientConnectionSettings,
                   log: LoggingAdapter = system.log): ClientLayer =
-    OutgoingConnectionBlueprint(hostHeader, settings, log)
+    RedirectSupportStage(settings.redirectSettings).atop(OutgoingConnectionBlueprint(hostHeader, settings, log))
```
ktoso: I think we could put that into the blueprint directly

akka-http-core/src/test/scala/akka/http/impl/engine/client/RedirectTestSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+import akka.stream.scaladsl._
+import akka.stream.testkit.Utils
+import akka.stream.{ ActorMaterializer, ActorMaterializerSettings }
+import akka.testkit.AkkaSpec
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+class RedirectTestSpec extends AkkaSpec {
+  implicit val materializer = ActorMaterializer(ActorMaterializerSettings(system).withFuzzing(true))
```
ktoso: fuzzing is enabled by default (see reference.conf in test scope), you can simply `ActorMaterializer()`

akka-http-core/src/test/scala/akka/http/impl/engine/client/RedirectTestSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+
+    "be able to handle infinite redirect loop" in Utils.assertAllStagesStopped {
+      val (_, serverHostName, serverPort) = TestUtils.temporaryServerHostnameAndPort()
+      val cycleLength = 3
+      val binding = Http().bindAndHandleSync(r ⇒ {
+        val c = r.uri.toString.reverse.takeWhile(Character.isDigit).reverse.toInt
+        HttpResponse(status = StatusCodes.MovedPermanently, headers = List(Location(s"/r${(c % cycleLength) + 1}")))
+      }, serverHostName, serverPort)
+
+      val x = Source.fromIterator(() ⇒ Iterator.from(1))
```
ktoso: hm, wouldn't `single()` be enough? The redirects are happening internally, right?

akka-http-core/src/test/scala/akka/http/impl/engine/client/RedirectTestSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+        .take(cycleLength)
+        .map(id ⇒ HttpRequest(uri = s"/r$id"))
+        .via(Http().outgoingConnection(serverHostName, serverPort))
+        .runWith(Sink.head)
+
+      a[RedirectSupportStage.InfiniteRedirectLoopException.type] should be thrownBy Await.result(x, 3.second)
+
+      binding.futureValue.unbind()
```
ktoso: `futureValue` on the unbind as well :)

akka-http-core/src/test/scala/akka/http/impl/engine/client/RedirectTestSpec.scala
```diff
@@ -0,0 +1,127 @@
...
+          p.headers.find(_.is("etag")).map(_.value.replace("\"", "").toInt).getOrElse(0)
+        })
+        .runFold(0)(_ + _)
+
+      result.futureValue(PatienceConfig(1000.seconds)) shouldEqual (N + 1) * (N / 2)
+
+      binding.futureValue.unbind()
+    }
+
+    "be able to forward selected headers to redirected requests in different origin" /*in Utils.assertAllStagesStopped */ ignore {
```
ktoso: I'll check out the branch and investigate this one then; likely not this week though – easter happened.
Sorry for the delay!

RomanIakovlev:
Thanks a bunch for the review! I'll have a break for a couple of days now, and will address your findings afterwards.

ktoso:
Excellent, thank you! Your work looks very through so far, looking forward to figuring out the missing bits and pushing it over the finish line :) Have a nice easter, cheers :rabbit:

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
2beaucoup: I can think of some more aspects that'll come up (like caching or proxying). Should they all be baked into the main blueprint?

RomanIakovlev: I thought about those other aspects as well. Maybe it makes sense to introduce something like `HighLevelHttpClient`, to wrap all those aspects and conditionally enable or disable them separately based on the settings? Of course, this potential `HighLevelHttpClient` is bigger than this PR, and there's already an issue for it: https://github.com/akka/akka/issues/16856. I just think this particular place, i.e. `atop(OutgoingConnectionBlueprint)`, is the most appropriate place to add such high-level client.

Every such feature incur some overhead, and client doesn't have to pay the price for something he/she doesn't need, like these redirects, for example. In this particular case, I think I can check the settings on both `cross-origin` and `same-origin` branches, and if both are disabled, don't add this `RedirectSupportStage` on top of the `OutgoingConnectionBlueprint` at all.

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
```diff
@@ -0,0 +1,291 @@
...
+              None
+            }
+          })
+        }
+      } else {
+        None
+      }
+    }
+
+  def sameOrigin(u1: Uri, u2: Uri): Boolean = {
```
2beaucoup:
```scala
def sameOrigin(u1: Uri, u2: Uri) =
  u2.isRelative || u1.isAbsolute && u1.authority == u2.authority
```

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
```diff
@@ -0,0 +1,291 @@
...
+      case MovedPermanently | Found | TemporaryRedirect | PermanentRedirect if m == GET || m == HEAD => Some(m)
+      case SeeOther => Some(GET)
+      case _ => None
+    }
+  }
+
+  val defaultRedirectMapper: RedirectMapper =
+    (s, req, resp) => {
+      if (resp.status.isRedirection) {
+        inferMethod(req, resp).flatMap { method =>
```
2beaucoup: The nested `flatMap`s could be replaced by a for-loop (untested):

```scala
def isAllowed(location: Uri) = {
  val setting =
    if (resp.uri.isRelative || req.uri.isAbsolute && req.uri.authority == resp.uri.authority) s.sameOrigin
    else s.crossOrigin
  setting.getAllow
}

for {
  method <- inferMethod(req, resp)
  location <- resp.header[Location]
  if isAllowed(location.uri)
} yield {
  val headers = item.getForwardHeaders match {
    case All => req.headers
    case Zero => Nil
    case Only(hs) => req.headers.filter(h => hs.contains(h.name))
  }
  req.copy(uri = location.uri, headers = headers, method = method)
}
```

akka-http-core/src/main/scala/akka/http/impl/settings/ClientConnectionSettingsImpl.scala
RomanIakovlev: Speaking of binary compatibility, I'm currently struggling with another class, namely `akka.http.impl.settings.ClientConnectionSettingsImpl`. I've been able to create a constructor without the newly added parameter, however `mimaReportBinaryIssues` also complains about other case classes methods like `copy` and `apply`. AFAIK those are compiler generated and I can't override them. Or can I? And should I?

ktoso: The settings are very smartly structured to allow us to use case classes while retaining public API binary compatibility, you'll notice that:

```scala
/** INTERNAL API */
private[akka] final case class ClientConnectionSettingsImpl(
```

So noone is supposed to touch this class => we can change it and filter out MiMa warnings that it has about it.

RomanIakovlev: Thanks, it's good to know! Then how to mark these issues as ignored? 

Besides those I only have 2 errors in binary compatibility:
```
[error]  * abstract method getRedirectSettings()akka.http.javadsl.settings.ClientAut
oRedirectSettings in class akka.http.javadsl.settings.ClientConnectionSettings is pr
esent only in current version
```
and
```
[error]  * abstract method redirectSettings()akka.http.scaladsl.settings.ClientAutoR
edirectSettings in class akka.http.scaladsl.settings.ClientConnectionSettings is pre
sent only in current version
```
These are the newly added methods. Shouldn't they also be ignored for binary compatibility purposes? I mean, adding a method shouldn't impact the binary compatibility, should it?

ktoso: > I mean, adding a method shouldn't impact the binary compatibility, should it?

It depends. Imagine you've added an abstract method (like we do here), to a class which users are meant to extend. Suddenly you broke their code when they try to upgrade.

In this case the classes are lined up such that we can indeed silence this error – as only we implement this interface.

akka-http-core/src/test/scala/akka/http/impl/engine/client/RedirectTestSpec.scala
RomanIakovlev: Do you mean `binding.futureValue.unbind().futureValue`? I don't see this in previously existing specs using the same approach.

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
RomanIakovlev: This is a good question. I'm still not sure if this collection `inFlight` is needed at all with current approach. I've already raised this question [before](https://github.com/akka/akka/issues/15990#issuecomment-198074583). I think, current implementation of the push-pull logic prevents more than 1 request in flight, i.e. I don't pull from `requestIn` before I send the reply to the previous request to the `responseOut`. It is also consistent with the in flight numbers I see during tests.

So with this current approach (always one request in flight), the `inFlight` collection is not needed. I've implemented it like this because I was a bit confused on this topic, but now it starts to become clearer. I think I should either just have 1 item in flight, and keep it in an `Option` instead of `Vector`, or change push-pull logic and truly support multiple, if limited, number of requests in flight. Which approach is preferable?

akka-http-core/src/test/scala/akka/http/impl/engine/client/RedirectTestSpec.scala
ktoso: Then those other places are likely wrong :)

ktoso:
Was not able to get to this yet, slipping to 2.4.4 (which will be released whenever we have javadsl ready and this likely too)

ktoso:
Refs #15990

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
```diff
@@ -0,0 +1,299 @@
...
+      }
+    }
+
+    def sameOrigin(u1: Uri, u2: Uri): Boolean = {
+      u2.isRelative || u1.isAbsolute && u1.authority == u2.authority
+    }
+
+    def apply(s: ClientAutoRedirectSettings, req: HttpRequest, resp: HttpResponse): Option[HttpRequest] = {
+      for {
+        location <- resp.header[Location] if resp.status.isRedirection
```
2beaucoup: I'd keep the cheap `if` check outside the for-loop.

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
```diff
@@ -0,0 +1,299 @@
...
+        method <- inferMethod(req, resp)
+        item = if (sameOrigin(req.uri, Uri(location.value))) s.sameOrigin else s.crossOrigin
```
2beaucoup: `Uri(location.value)` can be replaced with `location.uri`.

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
```diff
@@ -0,0 +1,299 @@
...
+  implicit val ResponseInPrinterResp = new OpInPrinter[Inlet[HttpResponse]] {
+    override def print(): Unit = {
+      printy("\tpulling from responseIn")
+    }
+  }
+
+  def apply(settings: ClientAutoRedirectSettings) =
+    BidiFlow.fromGraph(new RedirectSupportStage(settings, DefaultRedirectMapper()))
+
+  def apply(settings: ClientAutoRedirectSettings, redirectMapper: RedirectMapper) =
```
2beaucoup: Use default value for `redirectMapper` and remove the above method?

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
```diff
@@ -0,0 +1,299 @@
...
+          * Called when the input port has a new element available. The actual element can be retrieved via the
+          * [[GraphStageLogic.grab()]] method.
+          */
+        override def onPush(): Unit = {
+          printy("responseIn: onPush")
+          val response = grab(responseIn)
+          redirectMapper(settings, inFlight.head, response) match {
+            case Some(redirect) =>
+              printy(loopDetector.toString())
+              val loops = loopDetector(inFlight.head)
```
2beaucoup: There should be no loops in here. ;) Rename to `followed`?

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
```diff
@@ -0,0 +1,299 @@
...
+          case All => req.headers
+          case Zero => Nil
+          case Only(hs) => req.headers.filter(h => hs.exists(h.is))
+        }
+        req.withUri(location.value).withHeaders(headers).withMethod(method)
+      }
+    }
+  }
+
+  case class InfiniteRedirectLoopException(loop: immutable.Seq[Uri])
```
2beaucoup: Just `RedirectLoopException`?

akka-http-core/src/test/scala/akka/http/impl/engine/client/RedirectTestSpec.scala
```diff
@@ -0,0 +1,126 @@
...
+        } else {
+          HttpResponse(status = StatusCodes.MovedPermanently, headers = List(Location(s"/r${c + 1}")))
+        }
+      }, serverHostName, serverPort)
+
+      val N = 100
+      val result = Source.fromIterator(() ⇒ Iterator.from(1))
+        .take(N)
+        .map(id ⇒ HttpRequest(uri = s"/r$id"))
+        .via(Http().outgoingConnection(serverHostName, serverPort))
```
2beaucoup: Would be nice to use `RedirectSupportStage` directly ISO `Http().outgoingConnection` in these tests so we can get rid of the bindings.

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
RomanIakovlev: @2beaucoup In case the `if` is outside the `for` comprehension, I'd have to step out from the `Option` monad and provide the `else` block. I don't think it looks better. Here, have a look:

      if (resp.status.isRedirection) {
        for {
          location <- resp.header[Location]
          method <- inferMethod(req, resp)
          item = if (sameOrigin(req.uri, location.uri)) s.sameOrigin else s.crossOrigin
          if item.allow
        } yield {
          val headers = item.forwardHeaders match {
            case All => req.headers
            case Zero => Nil
            case Only(hs) => req.headers.filter(h => hs.exists(h.is))
          }
          req.withUri(location.value).withHeaders(headers).withMethod(method)
        }
      }
      else {
        None
      }
    }

Instead of:

      for {
        location <- resp.header[Location] if resp.status.isRedirection
        method <- inferMethod(req, resp)
        item = if (sameOrigin(req.uri, location.uri)) s.sameOrigin else s.crossOrigin
        if item.allow
      } yield {
        val headers = item.forwardHeaders match {
          case All => req.headers
          case Zero => Nil
          case Only(hs) => req.headers.filter(h => hs.exists(h.is))
        }
        req.withUri(location.value).withHeaders(headers).withMethod(method)
      }

I'd rather keep the existing version in this case. But as you wish. Let me know if you still prefer to have the `if` outside of the `for`.

akka-http-core/src/test/scala/akka/http/impl/engine/client/RedirectTestSpec.scala
RomanIakovlev: @2beaucoup Sorry, I don't understand your proposal. Could you please clarify?

RomanIakovlev:
One thing has just occurred to me. Please correct me if I'm wrong anywhere in my reasoning.

Every `OutgoingConnectionBlueprint` instance (further referred to just as "outgoing connection") has maximum 1 corresponding TCP connection at any time. This means, one outgoing connection can only be connected to 1 host. This, in turn, means it's impossible to implement cross-origin redirects inside single outgoing connection.

Next, it seems that all of the existing Akka HTTP client features are implemented as graphs. To send request to another origin means to instantiate the outgoing connection inside another outgoing connection. Or, in other words, to support cross-origin redirects we must change the graph shape in the runtime. Which is not possible, at least currently, if I understand Akka Streams correctly.

To summarize, I think such high-level HTTP client features as redirects support cannot be implemented purely in Akka Streams terms. I think it's should possible to achieve the desired behavior using pure actors, but I'm not sure it's a good way to go forward. Or, maybe it's possible to come up with some creative solution, like changing the graph shape at runtime somehow. 

I hope core team members could shed some light here.

2beaucoup:
You are right @RomanIakovlev and that's one more reason to keep this kind of functionality out of the low level blueprints (and even the default connection methods in `Http`?).

I think your stage should work for cross origin redirects by using it atop `Http.superPool`. Atop `outgoingConnection` and `cachedHostConnectionPool` it should work at least for same origin redirects.

akka-http-core/src/test/scala/akka/http/impl/engine/client/RedirectTestSpec.scala
2beaucoup: Sure. `outgoingConnection` produces a `Flow[HttpRequest, HttpResponse, Future[OutgoingConnection]]`. If you provide such a flow yourself you don't need to to bind/unbind:
```scala
val flow = RedirectSupportStage() join Flow[HttpRequest].map(r ⇒ {
  val c = r.uri.toString.reverse.takeWhile(Character.isDigit).reverse.toInt
  if (c % 2 == 0) {
    HttpResponse(entity = c.toString)
  } else {
    HttpResponse(status = StatusCodes.MovedPermanently, headers = List(Location(s"/r${c + 1}")))
  }
})
val N = 100
val result = Source.fromIterator(() ⇒ Iterator.from(1))
  .take(N)
  .map(id ⇒ HttpRequest(uri = s"/r$id"))
  .via(flow)
  .mapAsync(4)(_.entity.toStrict(1.second))
  .map { r ⇒ val s = r.data.utf8String; log.debug(s); s.toInt }
  .runFold(0)(_ + _)
```

akka-http-core/src/main/scala/akka/http/impl/engine/client/RedirectSupportStage.scala
2beaucoup: I agree that the pure monadic style looks better. But the suggested change saves the unnecessary location header lookup which will always return `None` for all non-redirect responses and is more expensive to calculate than `isRedirection `.

If only if-statements could come first in for-loops. ;)

drewhk:
I think on the per-request API level it can be implemented on the top using Future composition by simply recovering from the case where a redirect has been found with another request Future using flatMap.

In the case of Flow based APIs, a loop might be necessary, reinjecting redirected requests to the wrapped flow (with an upper limit) using the graph DSL

ktoso:
Hey there, thanks for the nice chat on ScalaDays :-)
AFAIR we agreed to close it for now, but I'd hope you'd be able to give it another shot given what we've learned from the initial attempt?

A contrib of this feature would be very awesome!

Issue #20123, opened 2016-03-22
Deprecate recoverWith and create two variants with repeating and non-repeating semantics


Falmarri:
Scala version 2.11.7 (OpenJDK 64-Bit Server VM, Java 1.8.0_74)
Akka 2.4.2

```
scala> 
Source.failed(
    new Exception("An original failure")
).recoverWith{ 
    case e => Source.single(10).concat(Source.failed(e))
}.runForeach(println)
res3: scala.concurrent.Future[akka.Done] = 
scala.concurrent.impl.Promise$DefaultPromise@127a7272

scala> 10
10
10
10
10
10
10
10
10
10
10
10
...
```

drewhk:
Why is this a bug?

ktoso:
Hm... it is weird, isn't it?

I'd expect:
`failed(ex) => no failure downstream since recovered => single(10) => the failure ex`

drewhk:
it depends on the intended semantics. I am not sure what is the expected semantic here, is it to apply the recoverWith *every time*, or just once. The every time semantic is useful because for example it allows to implement reconnect style logic. I marked this as "new" until it is figured out what recoverWith should actually mean.

drewhk:
It says in the doc:
"RecoverWith allows to switch to alternative Source on flow failure. **It will stay in effect after
a failure has been recovered so that each time there is a failure it is fed into the `pf` and a new
Source may be materialized.**"

drewhk:
So it does what it advertises

ktoso:
Huh whoa, I did not realise that's the semantics we have, missed reviewing that one maybe :-/
It is rather weird... The other one should be allowed as well IMO

drewhk:
Why is it weird? The single recover is as simple as:

```scala
src.recover { case ex => Dummy }.filter(_ != Dummy).concat(recoverSrc)
```

While the recoverWith behavior cannot be expressed by built-in combinators.

drewhk:
Hm, not exactly because above always concats the second source.

drewhk:
Maybe we can add an additional parameter capping the restart count.

ktoso:
> Maybe we can add an additional parameter capping the restart count.

Yeah (or play with naming). In general recoverWith for me is all about the Source inside - which could be another http call etc, and I'd want to retry only once etc.

drewhk:
we cannot rename it

ktoso:
[...]

PR #20137, opened 2016-03-24
#20123 Added recoverWithRetries

Falmarri:
recoverWithRetries adds a parameter to limit the number of attempts. Existing recoverWith method behavior remains unchanged.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -1220,10 +1220,16 @@ private[stream] final class Reduce[T](f: (T, T) ⇒ T) extends SimpleLinearGraph
 /**
  * INTERNAL API
  */
-private[stream] final class RecoverWith[T, M](pf: PartialFunction[Throwable, Graph[SourceShape[T], M]]) extends SimpleLinearGraphStage[T] {
+private[stream] object RecoverWith {
+  val InfiniteRetries = -1
+}
+
+private[stream] final class RecoverWith[T, M](maximumRetries: Int, pf: PartialFunction[Throwable, Graph[SourceShape[T], M]]) extends SimpleLinearGraphStage[T] {
   override def initialAttributes = DefaultAttributes.recoverWith
 
   override def createLogic(attr: Attributes) = new GraphStageLogic(shape) {
+    var attempt = 0
+
     setHandler(in, new InHandler {
       override def onPush(): Unit = push(out, grab(in))
       override def onUpstreamFailure(ex: Throwable) = onFailure(ex)
@@ -1233,7 +1239,13 @@ private[stream] final class RecoverWith[T, M](pf: PartialFunction[Throwable, Gra
       override def onPull(): Unit = pull(in)
     })
 
-    def onFailure(ex: Throwable) = if (pf.isDefinedAt(ex)) switchTo(pf(ex)) else failStage(ex)
+    def onFailure(ex: Throwable) =
+      if ((maximumRetries == RecoverWith.InfiniteRetries || attempt < maximumRetries) && pf.isDefinedAt(ex)) {
+        switchTo(pf(ex))
+        attempt += 1
+      }
+      else
+        failStage(ex)
 
     def switchTo(source: Graph[SourceShape[T], M]): Unit = {
       val sinkIn = new SubSinkInlet[T]("RecoverWithSink")
--- a/akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
+++ b/akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
@@ -795,8 +795,31 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
    * '''Cancels when''' downstream cancels
    *
    */
+  @deprecated("Use recoverWithRetries instead.", "2.4.4")
   def recoverWith[T >: Out](pf: PartialFunction[Throwable, _ <: Graph[SourceShape[T], NotUsed]]): javadsl.Flow[In, T, Mat @uncheckedVariance] =
     new Flow(delegate.recoverWith(pf))
+
+  /**
+    * RecoverWithRetries allows to switch to alternative Source on flow failure. It will stay in effect after
+    * a failure has been recovered up to `attempts` number of times so that each time there is a failure
+    * it is fed into the `pf` and a new Source may be materialized. Note that if you pass in 0, this won't
+    * attempt to recover at all. Passing in a negative number will behave exactly the same as  `recoverWith`.
+    *
+    * Since the underlying failure signal onError arrives out-of-band, it might jump over existing elements.
+    * This stage can recover the failure signal, but not the skipped elements, which will be dropped.
+    *
+    * '''Emits when''' element is available from the upstream or upstream is failed and element is available
+    * from alternative Source
+    *
+    * '''Backpressures when''' downstream backpressures
+    *
+    * '''Completes when''' upstream completes or upstream failed with exception pf can handle
+    *
+    * '''Cancels when''' downstream cancels
+    *
+    */
+  def recoverWithRetries[T >: Out](attempts: Int, pf: PartialFunction[Throwable, _ <: Graph[SourceShape[T], NotUsed]]): javadsl.Flow[In, T, Mat @uncheckedVariance] =
+    new Flow(delegate.recoverWithRetries(attempts, pf))
 
   /**
    * Terminate processing (and cancel the upstream publisher) after the given
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowRecoverWithSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowRecoverWithSpec.scala
@@ -53,7 +53,7 @@ class FlowRecoverWithSpec extends AkkaSpec {
         .expectError(ex)
     }
 
-    "be able to recover with th same unmaterialized source if configured" in assertAllStagesStopped {
+    "be able to recover with the same unmaterialized source if configured" in assertAllStagesStopped {
       val src = Source(1 to 3).map { a ⇒ if (a == 3) throw ex else a }
       src.recoverWith { case t: Throwable ⇒ src }
         .runWith(TestSink.probe[Int])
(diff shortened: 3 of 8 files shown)


ktoso:
OK TO TEST

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -1125,7 +1127,18 @@ private[stream] final class RecoverWith[T, M](pf: PartialFunction[Throwable, Gra
       override def onPull(): Unit = pull(in)
     })
 
-    def onFailure(ex: Throwable) = if (pf.isDefinedAt(ex)) switchTo(pf(ex)) else failStage(ex)
+    def onFailure(ex: Throwable) =
+      if (attempt != r && pf.isDefinedAt(ex)) {
```
drewhk: `attempt < r` is more expressive

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -1112,10 +1112,12 @@ private[stream] final class Reduce[T](f: (T, T) ⇒ T) extends SimpleLinearGraph
 /**
  * INTERNAL API
  */
-private[stream] final class RecoverWith[T, M](pf: PartialFunction[Throwable, Graph[SourceShape[T], M]]) extends SimpleLinearGraphStage[T] {
+private[stream] final class RecoverWith[T, M](r: Int, pf: PartialFunction[Throwable, Graph[SourceShape[T], M]]) extends SimpleLinearGraphStage[T] {
```
drewhk: Instead of `r` use a more descriptive name, like `maximumRetries`

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: while it is not very likely that attempt overflows to be `-1` I would still rather prefer to check for `-1` explicitly to disable checks and always retry. Also, you should name the constant to something like `InfiniteRetries`.

i.e.
```
if ((maximumRetries == InfiniteRetries || attempt < maximumRetries) && pf.isDefinedAt(ex))
```

then you don't need the extra check `(r > 0)`, just always increase the `attempt` it does not matter anyway.

drewhk:
LGTM

ktoso:
What do we do about recoverWith's behaviour though?
I do strongly feel the current behaviour is very unexpected (yes I know its documented and we missed it during review).

drewhk:
We can deprecate it, that is all we can do.

Falmarri:
@ktoso added deprecated tag to recoverWith. If 2.4.3 isn't the correct version I can update it. @drewhk updated and squashed.

Btw, is recoverWithRetries a good enough name? It's slightly verbose. I wish recoverWith is such a good name, too bad it's taken =\

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -1112,10 +1112,13 @@ private[stream] final class Reduce[T](f: (T, T) ⇒ T) extends SimpleLinearGraph
 /**
  * INTERNAL API
  */
-private[stream] final class RecoverWith[T, M](pf: PartialFunction[Throwable, Graph[SourceShape[T], M]]) extends SimpleLinearGraphStage[T] {
+private[stream] final class RecoverWith[T, M](maximumRetries: Int, pf: PartialFunction[Throwable, Graph[SourceShape[T], M]]) extends SimpleLinearGraphStage[T] {
   override def initialAttributes = DefaultAttributes.recoverWith
 
   override def createLogic(attr: Attributes) = new GraphStageLogic(shape) {
+    var attempt = 0
+    val InfiniteRetries = -1
```
patriknw: place in companion?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -1125,7 +1128,14 @@ private[stream] final class RecoverWith[T, M](pf: PartialFunction[Throwable, Gra
       override def onPull(): Unit = pull(in)
     })
 
-    def onFailure(ex: Throwable) = if (pf.isDefinedAt(ex)) switchTo(pf(ex)) else failStage(ex)
+    def onFailure(ex: Throwable) =
+      if ((maximumRetries == InfiniteRetries || attempt < maximumRetries) && pf.isDefinedAt(ex)) {
+        switchTo(pf(ex))
+      }
+      else {
```
patriknw: several superfluous braces here

ktoso:
> We can deprecate it, that is all we can do.

That's not true IMO. It is binary and source compatible, it is about fixing behaviour that is unexpected and if someone wants the "continue recovering" pointing towards this new method.

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -790,10 +790,33 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
...
 
   /**
+    * RecoverWithRetries allows to switch to alternative Source on flow failure. It will stay in effect after
+    * a failure has been recovered up to `attempts` number of times so that each time there is a failure
+    * it is fed into the `pf` and a new Source may be materialized. Note that if you pass in 0, this won't
+    * attempt to recover at all. Passing in a negative number will behave exactly the same as  `recoverWith`.
+    *
+    * Since the underlying failure signal onError arrives out-of-band, it might jump over existing elements.
+    * This stage can recover the failure signal, but not the skipped elements, which will be dropped.
+    *
```
patriknw: Add a note that the `PartialFunction` can be constructed with `akka.japi.pf.PFBuilder`.
Please also add that to other places where PF is used, e.g. `collect`.

patriknw:
LGTM

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
```diff
@@ -790,10 +790,33 @@ final class Flow[-In, +Out, +Mat](delegate: scaladsl.Flow[In, Out, Mat]) extends
    * '''Cancels when''' downstream cancels
    *
    */
+  @deprecated("Use recoverWithRetries instead.", "2.4.3")
   def recoverWith[T >: Out](pf: PartialFunction[Throwable, _ <: Graph[SourceShape[T], NotUsed]]): javadsl.Flow[In, T, Mat @uncheckedVariance] =
```
ktoso: I'm not convinced about that, what about https://github.com/akka/akka/pull/20137#issuecomment-203833534

patriknw:
by the way, don't we have a list of all ops in the reference docs that needs to be updated?

patriknw:
Refs #20123

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
ktoso: Ok, seems I'm outvoted on that change, let's keep the deprecation

akka-stream/src/main/scala/akka/stream/javadsl/Flow.scala
Falmarri: @patriknw Do you want that in this PR? Or a separate one?

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
Falmarri: Do you have a style guide somewhere? Should I remove the braces around the if {} even if the statement is on the next line? Or should I put the statement on the same line even though it's going to be longer than 120 chars?

patriknw: We use the official Scala style guide: http://docs.scala-lang.org/style/control-structures.html#curlybraces
Here, I would write it as 
```scala
    if ((maximumRetries == InfiniteRetries || attempt < maximumRetries) && pf.isDefinedAt(ex))
      switchTo(pf(ex))
    else
      failStage(ex)
```

drewhk:
> That's not true IMO. Fixing recoverWith is binary and source compatible, it is about fixing behaviour that is unexpected and if someone wants the "continue recovering" pointing towards this new method.

It is a behavior that has been explicitly documented in the scaladocs, hence, unexpected or not, this is the official semantics for now and we can't change it.

ktoso:
MiMa failed, I want to release `2.4.3` so taking over.

patriknw:
@ktoso wasn't it a real test failure, related to this feature:
```
[info] - must terminate with exception after set number of retries *** FAILED *** (32 milliseconds)
[info]   java.lang.AssertionError: assertion failed: expected OnError(akka.stream.scaladsl.FlowRecoverWithSpec$$anon$1: ex
[info] ), found OnNext(11)
[info]   at scala.Predef$.assert(Predef.scala:165)
[info]   at akka.testkit.TestKitBase$class.expectMsg_internal(TestKit.scala:350)
[info]   at akka.testkit.TestKitBase$class.expectMsg(TestKit.scala:326)
[info]   at akka.testkit.TestKit.expectMsg(TestKit.scala:734)
[info]   at akka.stream.testkit.TestSubscriber$ManualProbe.expectError(StreamTestKit.scala:361)
[info]   at akka.stream.scaladsl.FlowRecoverWithSpec$$anonfun$1$$anonfun$apply$mcV$sp$10$$anonfun$apply$mcV$sp$18.apply(FlowRecoverWithSpec.scala:131)
```

ktoso:
Oh, it seems both. It should also be added to the list of all operations we have so... Propose to slip this one.

patriknw:
I agree

Falmarri:
Hmm. I definitely passed the tests the first time... Must have screwed it up in the latest commit. Fixing now.

Falmarri:
@patriknw @ktoso ping. Just making sure this didn't fall off the radar.

patriknw:
@Falmarri there are still two problems, can you fix those?

1) DslConsistencySpec fails

2) Add `ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.stream.scaladsl.FlowOps.recoverWithRetries")` to: https://github.com/akka/akka/blob/master/project/MiMa.scala#L741

drewhk:
Yaay, passed! Any objections against merge?

patriknw:
Thanks @Falmarri

Issue #20086, opened 2016-03-20
Unable to match on ws.TextMessage.Streamed


ktoso:
It's not possible to write:

```scala
  val bindingFuture = Http().bindAndHandle({
    path("ws") {
      handleWebSocketMessages(Flow.fromSinkAndSource(
        Flow[Message]
          .collect {
            case TextMessage.Streamed(s) ⇒ //  <<<<<<<<<<<<<
              s.runForeach(s ⇒ println("STREAM: " + s))
              Strict("streaming...")

            case s @ TextMessage.Strict(msg) ⇒
              s
          }
          .map { x ⇒ { println("Strict: " + x); x } }
          .alsoTo(Sink.onComplete(ex ⇒ println("done: " + ex)))
          .to(Sink.ignore),
        Source.maybe[Message]))
    } 
  }, interface = "localhost", port = 8080)


```

as `Streamed` marked `private`.

ktoso:
(so people would not consume the streamed one, which would leave it hanging until timed-out).

ktoso:
Seems like a mistake that it's private to me, or was this intentional for some reason? 
// @sirthias @jrudolph

PR #20158, opened 2016-03-29
For Validation: =htc make ws.Message consistent and not private

ktoso:
Backport of https://github.com/akka/akka/issues/20086


ktoso:
Failure was the known (and fixed) GraphStages MiMa thingy

PR #20162, opened 2016-03-29
=str #19469 More verbose logging of stage and tcp errors

drewhk:
- Logs all exceptions thrown in GraphStages at error level
- Logs any exception leading to an abort of a tcp connection at debug level
  (cherry picked from commit 2ed7911)

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
@@ -467,7 +467,9 @@ private[stream] final class GraphInterpreter(
         logic.beforePreStart()
         logic.preStart()
       } catch {
-        case NonFatal(e) ⇒ logic.failStage(e)
+        case NonFatal(e) ⇒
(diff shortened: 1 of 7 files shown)


drewhk:
refs #19469

drewhk:
new MiMa failures, now these are actually caused by this PR

drewhk:
So where the heck should these MiMa filters go?? This was my second attempt.

akka-stream/src/main/scala/akka/stream/impl/fusing/GraphInterpreter.scala
```diff
@@ -534,7 +536,12 @@ private[stream] final class GraphInterpreter(
         catch {
           case NonFatal(e) ⇒
             if (activeStage == null) throw e
-            else activeStage.failStage(e, isInternal = true)
+            else {
+              val stage = assembly.stages(activeStage.stageId)
```
viktorklang: why a val?

drewhk:
PLS BUILD

drewhk:
Hm, InputStreamSpec. Has nothing to do with this PR but probably need to investigate separately.

PR #20164, opened 2016-03-29
str #19589 backport autoflushing in outputstream sink

ktoso:
Manual backport of the auto flushing feature added in https://github.com/akka/akka/issues/19589
Just cherry picking is not as trivial since we introduced more things (IOResult etc).

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/io/IOSinks.scala
@@ -44,7 +44,7 @@ private[akka] final class FileSink(f: File, append: Boolean, val attributes: Att
  * Creates simple synchronous (Java 6 compatible) Sink which writes all incoming elements to the given file
  * (creating it before hand if necessary).
  */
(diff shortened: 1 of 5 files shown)


drewhk:
But this changes the signature...

ktoso:
Nope, it's safe, specifically adds overload with the additional boolean (instead of changing signature to use default param, which would be breaking), internal classes constructor param we don't guarantee anything about (though you're right MiMa will panic, I'll fix that).

ktoso:
Actual perhaps failure?

```
[info] - must fill up buffer by default *** FAILED *** (42 milliseconds)
[info]   Array(-5, -17, -126) did not equal Array(76, 49, 2) (InputStreamSinkSpec.scala:150)
```

drewhk:
I believe this failure has nothing to do with the backport. @ktoso can you please rebase it? If after rebase validation passes, I opt for merging this.

ktoso:
rebased

ktoso:
Failure was unrelated https://github.com/akka/akka/issues/new

PR #20165, opened 2016-03-29
#19392: asInputStream timeout

drewhk:
(cherry picked from commit edd3dbd)


drewhk:
unrelated MiMa failure, already fixed

Issue #19976, opened 2016-03-08
feature parity of creating MediaTypes for JavaDSL


pawel-kaminski-krk:
I am using java dsl for akka-http. the original question is here https://groups.google.com/forum/?hl=en-GB#!topic/akka-user/19GPYsUzpSk but I try to summarize it :

I cannot use RawHeader for headers like contentType, contentLength when constructing httpResponse are ignored (as pointed in documentation). 

setting content-type using HttpEntity#create is just pain in java. It is super easy for common types like application/json but for a custom one to convert my, string mime back to ContentType I need to decide whether it is binary or not. 

I tried ```ContentTypes.create(MediaTypes.custom()) ``` but it creates MediaType rather than Binary, WithFixedCharset instances and there is no way to get ContentType from such custom MediaType. 

then I tried 

    ContentType.WithFixedCharset contentType = ContentTypes.create(
        **akka.http.scaladsl.model.**MediaType.applicationWithFixedCharset(
            "my-own", 
            HttpCharsets.UTF$minus8(), 
            ?)); // <---- I dont know event how to create empty Seq in java

but it suddenly is a scaladsl and I need to pass some scala charsets / scala collection ... and there is no equivalent in javadsl.

Why ContentTypes.create isn't just consume string??? maybe rawHeader??

I dont understand the hierarchy of the ContentType, the benefit of having Binary, applicationWithFixedCharset, applicationWithOpenCharset, and so on. this is not explained nowhere in docs.

ktoso:
Thanks for reporting, we'll need to add the more specific factories to javadsl. Will revisit the rest of the ticket then as well.

pawel-kaminski-krk:
If I can add one more thing, I found it quite hard to work with constant values as they are not attached directly do ContentType but to ContentTypes. even though there is a pattern other frameworks usually uses static public fields or methods directly on a type that should be passed to method.

lolski:
Can I have a go on this issue?

ktoso:
That would be great if you could pick this one up @lolski.
What has to be done here is to add methods to `javadsl.MediaTypes` that correspond to the scaladsl ones, such as:

```
object MediaType {

  def applicationBinary(subType: String, comp: Compressibility, fileExtensions: String*): Binary =
```

etc. There's quite a few of them, but porting should be rather simple, thanks in advance @lolski!

ktoso:
I'll mark as in progress to note @lolski is working on it

lolski:
My remark:

[...]

PR #20181, opened 2016-03-30
Adding HTTP Java DSL

lolski:
PR for https://github.com/akka/akka/issues/19976

### DIFF
--- a/akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
+++ b/akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
@@ -187,6 +187,119 @@ public final class MediaTypes {
     public static final MediaType.Binary VIDEO_X_SGI_MOVIE = akka.http.scaladsl.model.MediaTypes.video$divx$minussgi$minusmovie();
     public static final MediaType.Binary VIDEO_WEBM = akka.http.scaladsl.model.MediaTypes.video$divwebm();
 
+    public static MediaType.Binary applicationBinary(String subType, boolean compressible, String... fileExtensions) {
+        akka.http.scaladsl.model.MediaType.Compressibility comp = compressible ?
+                akka.http.scaladsl.model.MediaType.Compressible$.MODULE$ : akka.http.scaladsl.model.MediaType.NotCompressible$.MODULE$;
+
+        scala.collection.immutable.List<String> fileEx =
+                scala.collection.JavaConversions.asScalaBuffer(java.util.Arrays.asList(fileExtensions)).toList();
+
+        return akka.http.scaladsl.model.MediaType.applicationBinary(subType, comp, fileEx);
+    }
+
+    public static MediaType.WithFixedCharset applicationWithFixedCharset(String subType, HttpCharset charset, String... fileExtensions) {
+        akka.http.scaladsl.model.HttpCharset cs = (akka.http.scaladsl.model.HttpCharset) charset;
+
+        scala.collection.immutable.List<String> fileEx =
+                scala.collection.JavaConversions.asScalaBuffer(java.util.Arrays.asList(fileExtensions)).toList();
+
+        return akka.http.scaladsl.model.MediaType.applicationWithFixedCharset(subType, cs, fileEx);
+    }
+
+    public static MediaType.WithOpenCharset applicationWithOpenCharset(String subType, String... fileExtensions) {
+        scala.collection.immutable.List<String> fileEx =
+                scala.collection.JavaConversions.asScalaBuffer(java.util.Arrays.asList(fileExtensions)).toList();
+
+        return akka.http.scaladsl.model.MediaType.applicationWithOpenCharset(subType, fileEx);
+    }
+
+    public static MediaType.Binary audio(String subType, boolean compressible, String... fileExtensions) {
+        akka.http.scaladsl.model.MediaType.Compressibility comp = compressible ?
+                akka.http.scaladsl.model.MediaType.Compressible$.MODULE$ : akka.http.scaladsl.model.MediaType.NotCompressible$.MODULE$;
+
+        scala.collection.immutable.List<String> fileEx =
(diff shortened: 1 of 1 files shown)


ktoso:
Awesome and just in time too @lolski, thanks!

ktoso:
OK TO TEST

akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
```diff
@@ -187,6 +187,119 @@ private MediaTypes() { }
...
+                akka.http.scaladsl.model.MediaType.Compressible$.MODULE$ : akka.http.scaladsl.model.MediaType.NotCompressible$.MODULE$;
+
+        scala.collection.immutable.List<String> fileEx =
+                scala.collection.JavaConversions.asScalaBuffer(java.util.Arrays.asList(fileExtensions)).toList();
+
+        return akka.http.scaladsl.model.MediaType.applicationBinary(subType, comp, fileEx);
+    }
+
+    public static MediaType.WithFixedCharset applicationWithFixedCharset(String subType, HttpCharset charset, String... fileExtensions) {
+        akka.http.scaladsl.model.HttpCharset cs = (akka.http.scaladsl.model.HttpCharset) charset;
```
lolski: Please advise if there is a better way to convert HttpCharset from Java to Scala DSL

akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
```diff
@@ -187,6 +187,119 @@ private MediaTypes() { }
     public static final MediaType.Binary VIDEO_X_SGI_MOVIE = akka.http.scaladsl.model.MediaTypes.video$divx$minussgi$minusmovie();
     public static final MediaType.Binary VIDEO_WEBM = akka.http.scaladsl.model.MediaTypes.video$divwebm();
 
+    public static MediaType.Binary applicationBinary(String subType, boolean compressible, String... fileExtensions) {
+        akka.http.scaladsl.model.MediaType.Compressibility comp = compressible ?
```
ktoso: we'll want to mirror those in javadsl (types)

akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
```diff
@@ -187,6 +187,119 @@ private MediaTypes() { }
...
+
+        scala.collection.immutable.List<String> fileEx =
+                scala.collection.JavaConversions.asScalaBuffer(java.util.Arrays.asList(fileExtensions)).toList();
+
+        return akka.http.scaladsl.model.MediaType.video(subType, comp, fileEx);
+    }
+
+    // arguments have been reordered due to varargs having to be the last argument
+    // should we create multiple overloads of this function?
+    public static MediaType.Binary customBinary(String mainType, String subType, boolean compressible, java.util.Map<String, String> params, boolean allowArbitrarySubtypes, String... fileExtensions) {
```
lolski: Arguments are reordered due to default argument + varargs in the original ScalaDSL method

akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
ktoso: It will be casts like this indeed. It's "good" because we don't need wrappers (though it's "bad" because casting...).
In scala we have a safe utility in `JavaMapping` to make those casts safetly, I'm not sure if we can use it in java code

akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
ktoso: Sounds good, I'll have a look if the Scala one should be changed to or no need to then

akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
lolski: Sorry, which type specifically?

akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
```diff
@@ -187,6 +187,119 @@ private MediaTypes() { }
     public static final MediaType.Binary VIDEO_X_SGI_MOVIE = akka.http.scaladsl.model.MediaTypes.video$divx$minussgi$minusmovie();
     public static final MediaType.Binary VIDEO_WEBM = akka.http.scaladsl.model.MediaTypes.video$divwebm();
 
+    public static MediaType.Binary applicationBinary(String subType, boolean compressible, String... fileExtensions) {
+        akka.http.scaladsl.model.MediaType.Compressibility comp = compressible ?
+                akka.http.scaladsl.model.MediaType.Compressible$.MODULE$ : akka.http.scaladsl.model.MediaType.NotCompressible$.MODULE$;
+
+        scala.collection.immutable.List<String> fileEx =
+                scala.collection.JavaConversions.asScalaBuffer(java.util.Arrays.asList(fileExtensions)).toList();
```
patriknw: would it be possible to use `akka.japi.Util.immutableSeq` instead?

patriknw: and if that is the case there are several places to change

patriknw:
OK TO TEST

patriknw:
LGTM aside from the list conversion

akka-http-core/src/main/java/akka/http/javadsl/model/MediaTypes.java
ktoso: `Compressibility` - using the same trick we have in most javadsl - empty interface in javadsl which this one extends.
I'll take this over

ktoso:
This is very good, thanks! I'll polish the small things around class hierarchy for javadsl :+1:

PR #20190, opened 2016-03-31
http: Fix spinlock race condition in PoolGateway

nkvoll:
Removes the race condition between the side effects needed from "shutdown started" and "shutdown completed" by waiting for the shutdown started `Promise` to eventually callback the `shutdownStartedHandled` `Future`.

Usages of the `PoolGateway` that do not need special handling / side effects can use the `shutdownStartedPromise.future` as the `shutdownStartedHandled` argument, as shown in https://github.com/akka/akka/compare/master...nkvoll:fix-spinlock-race-condition-in-akka-http-client-connection-pool?expand=1#diff-57c314af9e0369f72b0f60ff1fa56f0aR574

Attempts to address: https://github.com/akka/akka/issues/20071.

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -321,8 +321,7 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
    */
   private[akka] def newHostConnectionPool[T](setup: HostConnectionPoolSetup)(
     implicit fm: Materializer): Flow[(HttpRequest, T), (Try[HttpResponse], T), HostConnectionPool] = {
-    val shutdownStartedPromise = Promise[Done]()
-    val gatewayFuture = FastFuture.successful(new PoolGateway(setup, shutdownStartedPromise, shutdownStartedPromise.future))
+    val gatewayFuture = FastFuture.successful(new PoolGateway(setup, () => ()))
     gatewayClientFlow(setup, gatewayFuture)
   }
 
@@ -568,23 +567,19 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
     val gatewayPromise = Promise[PoolGateway]()
     hostPoolCache.putIfAbsent(setup, gatewayPromise.future) match {
       case null ⇒ // only one thread can get here at a time
-        val whenShuttingDown = Promise[Done]()
-        val whenShuttingDownCompleted = Promise[Done]()
+        // holds the reference to the fastFuture, so we can remove it on the shutdown started hook of the pool gateway
+        var fastFuture: Future[PoolGateway] = null
         val gateway =
-          try new PoolGateway(setup, whenShuttingDown, whenShuttingDownCompleted.future)
+          try new PoolGateway(setup, () => hostPoolCache.remove(setup, fastFuture))
           catch {
             case NonFatal(e) ⇒
               hostPoolCache.remove(setup)
               gatewayPromise.failure(e)
               throw e
           }
-        val fastFuture = FastFuture.successful(gateway)
+        fastFuture = FastFuture.successful(gateway)
         hostPoolCache.put(setup, fastFuture) // optimize subsequent gateway accesses
         gatewayPromise.success(gateway) // satisfy everyone who got a hold of our promise while we were starting up
-        whenShuttingDown.future.onComplete { _ ⇒
-          hostPoolCache.remove(setup, fastFuture)
-          whenShuttingDownCompleted.success(Done)
-        }(fm.executionContext)
         fastFuture
 
       case future ⇒ future // return cached instance
--- a/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
+++ b/akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
@@ -321,7 +321,8 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
    */
   private[akka] def newHostConnectionPool[T](setup: HostConnectionPoolSetup)(
     implicit fm: Materializer): Flow[(HttpRequest, T), (Try[HttpResponse], T), HostConnectionPool] = {
-    val gatewayFuture = FastFuture.successful(new PoolGateway(setup, Promise()))
+    val shutdownStartedPromise = Promise[Done]()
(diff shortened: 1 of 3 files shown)


drewhk:
OK TO TEST

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -579,7 +581,10 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
         val fastFuture = FastFuture.successful(gateway)
         hostPoolCache.put(setup, fastFuture) // optimize subsequent gateway accesses
         gatewayPromise.success(gateway) // satisfy everyone who got a hold of our promise while we were starting up
-        whenShuttingDown.future.onComplete(_ ⇒ hostPoolCache.remove(setup, fastFuture))(fm.executionContext)
+        whenShuttingDown.future.onComplete { _ ⇒
+          hostPoolCache.remove(setup, fastFuture)
```
drewhk: so this is what needed the acknowledgement in a form of a second Future.

drewhk:
I have to go a bit deeper into this section of the code, I have a feeling that a simpler solution might work, I feel this extra Promise a bit ugly. Would be nice if we could get rid of it and simplify this logic.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
nkvoll: Yes, if this isn't done before the shutdown-future inside `PoolGateway` completes, it might be able to fetch itself from the `hostPoolCache`, which causes the issue.

drewhk: I got it. My only grudge is this extra Promise. Thinking if we can simplify this. For example this code can be factored as:
```
val completelyCompletedFuture = whenShuttingDown.future.andThen {
  hostPoolCache.remove(setup, fastFuture)
}
```

but it might be that we cannot return a Future only inject a Promise.

nkvoll:
It certainly feels a bit ugly, but I tried to follow the pattern of using Promises/Futures to model the callbacks.

I thought it could be simplified by not using `Promise[Done]` at all, but rather a synchronous callable, i.e something like `() => Unit` instead, but this is the first time I've looked at the code, and I'm not quite sure of the implications here :) It's sort of complicated by the FastFuture-wrapping of the cache when it's put into the `hostPoolCache`.

drewhk:
> but this is the first time I've looked at the code, and I'm not quite sure of the implications here :) 

The reason I can't give any recommendations because I am also not too comfortable with this part, so I will need a look at the code in my IDE. Not today though :(

nkvoll:
@drewhk I added a commit that replaces the Promise with a synchronous hook. The old behavior would still be possible to achieve using `new PoolGateway(setup, () => promise.complete(Done))`, but I couldn't find anything relying on it. The solution became slightly simpler when doing this (to my eyes, that is).

It did introduce a `var` in the `cachedGateway`, in order to access the `FastFuture` stored in the `hostPoolCache`, so it's possible to remove it in the hook, but it would only be a problem if something called `poolGateway.shutdown()` before the variable is assigned, which shouldn't be possible, as the same thread that's creating the instance, sets the variable and it isn't shared with anything prior to it being set.

viktorklang:
Question is which approach we should go with, given https://github.com/akka/akka/pull/20082

drewhk:
Good point, let's keep this PR stashed until the other is completed

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -567,19 +567,19 @@ class HttpExt(private val config: Config)(implicit val system: ActorSystem) exte
     val gatewayPromise = Promise[PoolGateway]()
     hostPoolCache.putIfAbsent(setup, gatewayPromise.future) match {
       case null ⇒ // only one thread can get here at a time
-        val whenShuttingDown = Promise[Done]()
+        // holds the reference to the fastFuture, so we can remove it on the shutdown started hook of the pool gateway
+        var fastFuture: Future[PoolGateway] = null
         val gateway =
-          try new PoolGateway(setup, whenShuttingDown)
+          try new PoolGateway(setup, () => hostPoolCache.remove(setup, fastFuture))
```
viktorklang: How is it safe to refer to the `fastFuture`-var in the supplied function?

nkvoll:
Is the referenced PR aware/intended to fix the same issue? Or is it a case of "it might be fixed" because it refactors the code around it? :) Thanks for your eyes on this!

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
nkvoll: I was thinking it might have to be volatile or such, but the variable is set from line 580-on and the `PoolGateway` hook is only executed when `poolGateway.shutdown()` is called. Which shouldn't happen in this time-frame (as far as I can tell) because nothing else holds a reference to the gateway instance. The `PoolInterfaceActor` can issue a `shutdown()`, but only after the idle timeout, which "shouldn't" occur before the variable has been assigned.

After thinking about it a little, it can be worked around by having the `PoolGateway` itself have a `val selfFuture = FastFuture.successful(this)`. Provided we change the `onShutdownStarted` hook to accept the `PoolGateway` instance as the first argument, we can use the `selfFuture` as the value in the `hostPoolCache`. That should make it easier to argue about.

Since the PR referenced in the conversation (https://github.com/akka/akka/pull/20190) changes a bunch of things around this code, it makes sense to only revisit this if the problem it's trying to solve still occurs after it's merged?

viktorklang: Sounds like a good plan.

samueltardieu:
@nkvoll The referenced PR goal is to fix the many problems caused by the collapsing chain of gateways (such as merging pools that have been created independently with `newHostConnectionPool` causing a too many queued connections error even if each pool was used within its defined bounds). As a consequence, the model was changed, and the error you fix here can no longer happen.

This was not the primary goal of the PR, but it is achieved as a side effect: since the cached connections are now handled by the pool master actor, we have no need to know when a pool shutdown has started (since there is no longer any need for asynchronous cache invalidation), and thus one of the problematic promises involved in the race condition has disappeared.

drewhk:
Ok, then I close this PR for now (we can reopen if needed).

Issue #19958, opened 2016-03-07
Throttle does not work when added to several flows.


jvieten:
If I add a Flow.throttle to independent streams only the last works correctly.

The attached sample shows that most of the time only the last "Invoke Pipe 2" gets written to System.out
[ThrottleTest.txt](https://github.com/akka/akka/files/161563/ThrottleTest.txt)

Could be a duplicate of #19862 
(Maybe @drewhk can help)

drewhk:
It might be very well a duplicate. I cannot look at this right now though, but I triage it for now.

drewhk:
I will take a look at this, although I don't see anything that would prevent reusability.

drewhk:
Throttle is reusable, we cannot reproduce non-reusablilty. I close this for now. Please rerun your tests after the 2.4.3 release (scheduled tomorrow) because there were multiple issues fixed with throttle.

PR #20191, opened 2016-03-31
19958: Add tests for throttle reuse

drewhk:
Throttle is reusable, so #19958 is invalid

### DIFF
--- a/akka-stream-tests/src/test/java/akka/stream/javadsl/FlowThrottleTest.java
+++ b/akka-stream-tests/src/test/java/akka/stream/javadsl/FlowThrottleTest.java
@@ -1,0 +1,50 @@
+package akka.stream.javadsl;
+
+import akka.Done;
+import akka.NotUsed;
+import akka.actor.ActorRef;
+import akka.japi.Pair;
+import akka.japi.function.Function;
+import akka.stream.*;
+import akka.testkit.JavaTestKit;
+import org.junit.ClassRule;
+import org.junit.Test;
+import scala.concurrent.duration.FiniteDuration;
+
+import java.util.*;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.TimeUnit;
+import akka.testkit.AkkaSpec;
+
+import static org.junit.Assert.assertEquals;
+
+public class FlowThrottleTest extends StreamTest {
+    public FlowThrottleTest() {
+        super(actorSystemResource);
+    }
+
+    @ClassRule
+    public static AkkaJUnitActorSystemResource actorSystemResource =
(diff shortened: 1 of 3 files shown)


akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowThrottleSpec.scala
```diff
@@ -46,6 +46,23 @@ class FlowThrottleSpec extends AkkaSpec {
...
+      Source.single(1)
+        .via(sharedThrottle)
+        .via(sharedThrottle)
+        .runWith(Sink.seq).futureValue should ===(Seq(1))
+
+      // It works with a new stream, too
+      Source.single(2)
+        .via(sharedThrottle)
+        .via(sharedThrottle)
+        .runWith(Sink.seq).futureValue should ===(Seq(2))
```
patriknw: interesting that `immutable.Seq` and `Seq` are `===` compatible

patriknw:
LGTM

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowThrottleSpec.scala
drewhk: Yep :) Was thinking about using a List, but then was in the mood of trying out `Seq` instead since it rhymes better:
 - run-with sink-seq future value
 - should equal-equal a Seq-of two

akka-stream-tests/src/test/java/akka/stream/javadsl/FlowThrottleTest.java
```diff
@@ -0,0 +1,50 @@
+package akka.stream.javadsl;
```
ktoso: license ;)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowThrottleSpec.scala
ktoso: Unless the "statically type checked triple equals" trait is imported normal equality is used even with === AFAIR isn't it?

akka-stream-tests/src/test/java/akka/stream/javadsl/FlowThrottleTest.java
drewhk: meh

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowThrottleSpec.scala
patriknw: I thought that was done by `AkkaSpec`?

ktoso:
LGTM

PR #20194, opened 2016-03-31
=htc #19834 Replaced PushStage based captureTermination with GraphStage

leachbj:


akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -63,17 +63,32 @@ private[http] object StreamUtils {
...
-    val transformer = new PushStage[T, T] {
-      def onPush(element: T, ctx: Context[T]) = ctx.push(element)
-      override def onUpstreamFailure(cause: Throwable, ctx: Context[T]) = {
-        promise.failure(cause)
-        ctx.fail(cause)
-      }
-      override def postStop(): Unit = {
-        promise.trySuccess(())
+    val transformer = new SimpleLinearGraphStage[T] {
+      override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
```
leachbj: seems like a lot more code to do something simple!  Am I missing a simpler API?

drewhk:
OK TO TEST

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -63,17 +63,32 @@ private[http] object StreamUtils {
...
+        setHandler(in, new InHandler {
+          override def onPush(): Unit = push(out, grab(in))
+
+          override def onUpstreamFinish(): Unit = {
```
drewhk: why not just put these (and onDownstreamFinish) into postStop like in the previous version?

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
drewhk: Not really. Why we can simplify the linear case (similar to previous API) it would mean that people need to learn *two* different stage APIs. Now you only need to learn one and that will work for every stage. In your own projects you can add subclasses if you like that simplify the boilerplate.

drewhk: btw, the logic should extend the inHandler and outHandler. Not just that it looks simpler, it also saves on allocations and two object headers.

drewhk: and then you can use `setHandler(in, out, this)`

drewhk:
Small nitpicks, otherwise looks good! Thank you for the contribution.

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
leachbj: Done

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
leachbj: Ah!  Didn't realise that was there.  Done.

patriknw:
LGTM

ktoso:
LGTM, thanks @leachbj !

Note that this is part of the `StreamUtils` which we'll want to remove eventually

ktoso:
Refs #19834

PR #20206, opened 2016-04-02
Eliminate doublecopy of byte array

alno:
When `CompactByteString` is created from varargs, args are copied to the new byte array in `apply(bytes: Byte*)` method, then this array is passed to `apply(bytes: Array[Byte])` there it's copied again. 

Second copy is unnecessary and may be eliminated by calling `ByteString.ByteString1C` directly. Also, it makes code more consistent, because other apply methods already call it directly on generated byte arrays.

### DIFF
--- a/akka-actor/src/main/scala/akka/util/ByteString.scala
+++ b/akka-actor/src/main/scala/akka/util/ByteString.scala
@@ -528,7 +528,7 @@ object CompactByteString {
     else {
       val ar = new Array[Byte](bytes.size)
       bytes.copyToArray(ar)
-      CompactByteString(ar)
+      ByteString.ByteString1C(ar)
     }
   }
 


ktoso:
OK TO TEST

akka-actor/src/main/scala/akka/util/ByteString.scala
```diff
@@ -528,7 +528,7 @@ object CompactByteString {
     else {
       val ar = new Array[Byte](bytes.size)
       bytes.copyToArray(ar)
-      CompactByteString(ar)
+      ByteString.ByteString1C(ar)
```
patriknw: I wonder if we can do even better by

```
else 
  ByteString.ByteString1C(bytes.toArray)
```

The varargs `bytes` is a `scala.collection.mutable.WrappedArray$ofByte` and the `toArray` is therefore a cast of the internal array. cloning is done by `ByteString1C`

alno: There is a problem - `ByteString1C` doesn't clone array, it just stores ref in field. So, if using `toArray` we'll allow creating implicitly mutable ByteString by calling `ByteString(someMutableArray: _*)`.

patriknw: ah, good point, I must have looked at wrong apply

is that copyToArray more efficient than `bytes.toArray.clone`?

alno: I'd run some benchmarks on different implementations of `apply` and received following results:

First, when passing several byte arguments `toArray.clone` seems to be slightly faster than `copyToArray`:

```
VarargsCopy          thrpt  600   82420.414 ±  707.086  ops/ms
VarargsToArrayClone  thrpt  600   85950.390 ±  684.836  ops/ms
```

Also, there is an old implementation (DoubleCopy), and toArray without copy, just for comparison:

```
VarargsDoubleCopy    thrpt  600   61192.525 ±  542.109  ops/ms
VarargsToArray       thrpt  600  132025.408 ± 1357.725  ops/ms
```
But if passing varargs from some existing seq like `ByteString(someSeq: _*)` `toArray` call becomes very expensive:

```
SeqCopy              thrpt  600   24281.646 ±  223.725  ops/ms
SeqToArrayClone      thrpt  600    2720.174 ±   20.391  ops/ms
SeqDoubleCopy        thrpt  600   19321.256 ±  164.948  ops/ms
SeqToArray           thrpt  600    2835.829 ±   23.124  ops/ms
```

I think that it's because of reflection, which is used to create array in [`TraversableOnce.toArray`](https://github.com/scala/scala/blob/v2.11.8/src/library/scala/collection/TraversableOnce.scala#L285) from `ClassTag`.

So, if second scenario important enough? If it's not, then it really make sense to just use `toArray.clone`, and I'll update the PR.

patriknw: Thanks for testing it, @alno 
I was only thinking of using `toArray.clone` in `def apply(bytes: Byte*): CompactByteString` where we know that it is an array. Or do you mean that varargs can be something else than a (wrapped) array? If that is the case I think we should stick to the copy, especially since the performance difference was not big.

alno: Yes, it can, though it's rare case.

Unlike Java, there varargs are translated to array argument, Scala varargs are translated to Seq argument, so method like `def info(a: Int*)` have signature `(Lscala/collection/Seq;)V`.

And when casting some existing seq to vararg by `someSeq : _*` it's not copied to array, just passed as is. For example:
```
scala> def info(a: Int*) = println(a.getClass)
scala> info(Vector(1,2,3,4,5): _*)
class scala.collection.immutable.Vector
```

patriknw: I see, thanks for clarifying. Let's use the copy then. Thanks.

akka-actor/src/main/scala/akka/util/ByteString.scala
```diff
@@ -528,7 +528,7 @@ object CompactByteString {
     else {
       val ar = new Array[Byte](bytes.size)
       bytes.copyToArray(ar)
-      CompactByteString(ar)
```
viktorklang: Nice catch!

Issue #20210, opened 2016-04-03
doc not generated under python3.5


kpbochenek:
As stated in title.

Stacktrack with logLevel := debug

`akka > sphinx:generateHtml`

```
[debug] Exception occurred:
[debug]   File "~/github/akka/akka-docs/_sphinx/exts/includecode.py", line 118, in run
[debug]     tabshift = min(tabcounts) if tabcounts else 0
[debug] ValueError: min() arg is an empty sequence
[debug] The full traceback has been saved in /tmp/sphinx-err-lyg_y3aa.log, if you want to report the issue to the developers.
```

For testing I removed condition and left 0, doc was generated just fine.

ktoso:
Does the fix you did yield proper docs?
Could you contribute it as PR?

Thanks!

kpbochenek:
It doesn't unfortunately.
This code should as I can see remove indentation if all lines have it but it is not happening.

but I will figure it out in a minute :)

PR #20212, opened 2016-04-03
=doc #20210 fix generation with python 3.5

kpbochenek:
in python 3.5 map is lazy by default and returns "map object"
tabcounts is an object and not a collection yet in that case.
`if tabcounts` is always true because of above.
`min` evaluates this lazy map and in case of no elements it throws exception.
but it should return 0 as it is defined in `else` case.
simple evaluation before if does the trick.

### DIFF
--- a/akka-docs/_sphinx/exts/includecode.py
+++ b/akka-docs/_sphinx/exts/includecode.py
@@ -114,7 +114,7 @@ class IncludeCode(Directive):
                 "Snippet ({}#{}) not found!".format(filename, section),
                 line=self.lineno
             )]
-        tabcounts = map(lambda l: countwhile(lambda c: c == ' ', l), nonempty)
+        tabcounts = list(map(lambda l: countwhile(lambda c: c == ' ', l), nonempty))
         tabshift = min(tabcounts) if tabcounts else 0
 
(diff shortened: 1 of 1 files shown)


ktoso:
OK TO TEST

ktoso:
So does this yield proper docs, I guess it should, right?

patriknw:
does this break anything with old installations?

kpbochenek:
It generates correct docs with python3.
Jenkins also built docs with python2 which doesn't introduce any compilation issues.
If we could look at generated doc on jenkins for python2 that would confirm everything.

patriknw:
my default `python --version` says 2.7.5, so perhaps I can try?
what should I look for?

kpbochenek:
`sbt sphinx:generateHtml`
doc is generated into: akka/target/sphinx/html/scala.html
check section persistence, subsection "Dependencies" (one line code) and "Event sourcing" (multi indentation example)
if they are nicely aligned then all worked great :)

patriknw:
yes, that looks all good, thanks!

patriknw:
LGTM

patriknw:
Refs #20210

PR #20213, opened 2016-04-04
Clarify system name requirement for cluster members

patriknw:
also stronger recommendation against auto-down


akka-cluster/src/main/scala/akka/cluster/Cluster.scala
```diff
@@ -258,6 +258,9 @@ class Cluster(val system: ExtendedActorSystem) extends Extension {
    * An actor system can only join a cluster once. Additional attempts will be ignored.
    * When it has successfully joined it must be restarted to be able to join another
    * cluster or to join the same cluster again.
+   *
+   * The name of the [[akka.actor.ActorSystem]] must be the same for all members of a
```
drewhk: make this a warning, so it is hard to miss.

drewhk: ah there is a note in the other file, ignore this.

ktoso:
LGTM

PR #20215, opened 2016-04-04
Improve documentation and configuration API when using together: sharding and distibuted data

jsuchenia:
It's hard to configure stateStoreMode per each shard in Java as one setter is missing.
Additionally it's worth to document that both features when used together have some drawbacks (shards are not recreated after clean start, only after rebalance)

### DIFF
--- a/akka-docs/rst/java/cluster-sharding.rst
+++ b/akka-docs/rst/java/cluster-sharding.rst
@@ -202,7 +202,8 @@ you use this mode. It is possible to remove ``akka-persistence`` dependency from
 is not used in user code and ``remember-entities`` is ``off``.
 Using it together with ``Remember Entities`` shards will be recreated after rebalancing, however will
 not be recreated after a clean cluster start as the Sharding Coordinator state is empty after a clean cluster
-start when using ddata mode.
+start when using ddata mode. When ``Remember Entities`` is ``on`` Sharding Region always keeps data usig persistence,
+no matter how ``State Store Mode`` is set.
 
 .. warning::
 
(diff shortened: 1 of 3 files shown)


patriknw:
OK TO TEST

patriknw:
LGTM

akka-docs/rst/java/cluster-sharding.rst
```diff
@@ -200,6 +200,8 @@ on all nodes::
 You must explicitly add the ``akka-distributed-data-experimental`` dependency to your build if
 you use this mode. It is possible to remove ``akka-persistence`` dependency from a project if it
 is not used in user code and ``remember-entities`` is ``off``.
+Using it together with ``Remember Entities`` shards will be recreated after rebalancing, however will
```
ktoso: Shouldn't this document explicitly that persistence is used for storing the remembered entries?

jsuchenia: There is a sentence that remember Entities keeps data in a persistence - and that's fine. However when you switch stateStoreMode to ddata (in our case for test purposes, globally) rememberEntities stopped working at start, but after a wake-up it was working fine (so you need to send message to one actor in a region, then all of them were started)

That was my intention to add this remark

akka-docs/rst/java/cluster-sharding.rst
```diff
@@ -200,6 +200,8 @@ on all nodes::
...
+not be recreated after a clean cluster start - as data source is empty.
```
patriknw: perhaps reword, something like "as the Sharding Coordinator state is empty after a clean cluster start when using `ddata` mode"

akka-docs/rst/java/cluster-sharding.rst
drewhk: maybe add a small fragment to the end that the state is empty because ddata is purely in-memory (that drives the point home).

johanandren:
LGTM

patriknw:
thanks

PR #20230, opened 2016-04-04
19834 convert streamutils graphstage

leachbj:


ktoso:
OK TO TEST

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -50,15 +51,17 @@ private[http] object StreamUtils {
...
+        override def onPull(): Unit = pull(in)
 
-      override def onUpstreamFailure(cause: Throwable, ctx: Context[ByteString]): TerminationDirective =
-        ctx.fail(f(cause))
+        setHandlers(in, out, this)
+      }
     }
 
-    Flow[ByteString].transform(() ⇒ transformer).named("transformError")
+    Flow[ByteString].via(transformer).named("transformError")
```
ktoso: could be replaced with `.recover { t => throw f(t)` I think?

ktoso:
Thanks, looks good - LGTM

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
leachbj: I just went with a literal translation, probably worth looking at a refactor of the callers though;

`CodingDirectives.decodeRequestWith` looks like the only user.

    effectiveDecoder.decode(request).mapEntity(StreamUtils.mapEntityError {
              case NonFatal(e) ⇒
                IllegalRequestException(
                  StatusCodes.BadRequest,
                  ErrorInfo("The request's encoding is corrupt", e.getMessage))
            })

Perhaps just refactor to 

    effectiveDecoder.decode(request).mapEntity { entity ⇒
              entity.transformDataBytes(Flow[ByteString].recover {
                case NonFatal(e) ⇒
                  throw IllegalRequestException(
                    StatusCodes.BadRequest,
                    ErrorInfo("The request's encoding is corrupt", e.getMessage))
              })
            }```

ktoso: Yes, this looks good to me – would you apply this change here?
In general we're trying to get rid of the special impls inside of StreamUtils :)

ktoso:
Excellent, thanks!

LGTM

drewhk:
LGTM

drewhk:
It only needs the commits to be squashed and then it can be merged

ktoso:
We can do the squashing now btw @drewhk https://github.com/blog/2141-squash-your-commits :-)

drewhk:
totally awesome!

PR #20234, opened 2016-04-05
The great reformatting

ktoso:
Thanks to the the efforts by @bwmcadams and @jrudolph in #19746 and https://github.com/akka/akka/pull/19936
I was able to update to latest published scalariform (sadly not the `0.2.0` - as it won't be released anyway it seems until 1.0 hits – https://github.com/scala-ide/scalariform/issues/223)

I tried to keep the formatting rules similar to what we had (and as @jrudolph found), still a bit of reformatting going on. No errors are printed while formatting AFAICS.

Note that the -> rewriting will eventually be hidden under a flag as generally speaking it is not always safe: https://github.com/scala-ide/scalariform/issues/209 I made it work for our FSM though – by adding object only, which works around the problem found in https://github.com/akka/akka/pull/19746#discussion_r58512702 (no setter is generated - that's what was blocking that other PR)

### DIFF
--- a/akka-actor/src/main/scala/akka/actor/FSM.scala
+++ b/akka-actor/src/main/scala/akka/actor/FSM.scala
@@ -315,6 +315,10 @@ trait FSM[S, D] extends Actor with Listeners with ActorLogging {
   val Event: FSM.Event.type = FSM.Event
   val StopEvent: FSM.StopEvent.type = FSM.StopEvent
 
+  object → {
+    def unapply[Z](in: (Z, Z)) = Some(in)
+  }
+
   /**
    * This extractor is just convenience for matching a (S, S) pair, including a
    * reminder what the new state is.
--- a/akka-actor/src/main/scala/akka/actor/Scheduler.scala
+++ b/akka-actor/src/main/scala/akka/actor/Scheduler.scala
@@ -44,8 +44,9 @@ trait Scheduler {
     initialDelay: FiniteDuration,
     interval: FiniteDuration,
     receiver: ActorRef,
-    message: Any)(implicit executor: ExecutionContext,
-                  sender: ActorRef = Actor.noSender): Cancellable =
+    message: Any)(implicit
+    executor: ExecutionContext,
+    sender: ActorRef = Actor.noSender): Cancellable =
     schedule(initialDelay, interval, new Runnable {
       def run = {
         receiver ! message
@@ -72,7 +73,8 @@ trait Scheduler {
   final def schedule(
     initialDelay: FiniteDuration,
     interval: FiniteDuration)(f: ⇒ Unit)(
-      implicit executor: ExecutionContext): Cancellable =
+    implicit
+    executor: ExecutionContext): Cancellable =
     schedule(initialDelay, interval, new Runnable { override def run = f })
 
   /**
@@ -105,8 +107,9 @@ trait Scheduler {
   final def scheduleOnce(
     delay: FiniteDuration,
     receiver: ActorRef,
-    message: Any)(implicit executor: ExecutionContext,
-                  sender: ActorRef = Actor.noSender): Cancellable =
+    message: Any)(implicit
+    executor: ExecutionContext,
+    sender: ActorRef = Actor.noSender): Cancellable =
     scheduleOnce(delay, new Runnable {
       override def run = receiver ! message
     })
@@ -118,7 +121,8 @@ trait Scheduler {
    * Scala API
    */
   final def scheduleOnce(delay: FiniteDuration)(f: ⇒ Unit)(
-    implicit executor: ExecutionContext): Cancellable =
+    implicit
+    executor: ExecutionContext): Cancellable =
     scheduleOnce(delay, new Runnable { override def run = f })
 
   /**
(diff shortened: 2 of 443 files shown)


ktoso:
Only MiMa filter added is: 

```
      // added → alias in FSM
      ProblemFilters.exclude[ReversedMissingMethodProblem]("akka.actor.FSM.→"),
```

(note that that's the OK one, as explained in #19746 , the bad one we don't have here - yay)

akka-actor/src/main/scala/akka/actor/FSM.scala
```diff
@@ -315,6 +315,10 @@ trait FSM[S, D] extends Actor with Listeners with ActorLogging {
   val Event: FSM.Event.type = FSM.Event
   val StopEvent: FSM.StopEvent.type = FSM.StopEvent
 
+  object → {
```
patriknw: is this really binary backwards compatible? if it would have been a `def` it wouldn't

ktoso:
WAT

```
[error] /localhome/jenkinsakka/workspace/pr-validator-per-commit-jenkins/akka-cluster/src/main/scala/akka/cluster/routing/AdaptiveLoadBalancing.scala:373: value / is not a member of (akka.actor.Address, Double)
[error]       case (addr, (sum, count)) ⇒ (addr → sum / count)
```

`->` has lower precedence than `→` but higher than `/`... we will have to disable the -> formatting if we want to upgrade.

akka-actor/src/main/scala/akka/actor/Scheduler.scala
```diff
@@ -44,8 +44,9 @@ trait Scheduler {
     initialDelay: FiniteDuration,
     interval: FiniteDuration,
     receiver: ActorRef,
-    message: Any)(implicit executor: ExecutionContext,
-                  sender: ActorRef = Actor.noSender): Cancellable =
+    message: Any)(implicit
+    executor: ExecutionContext,
+    sender: ActorRef = Actor.noSender): Cancellable =
```
2beaucoup: This looks ugly. Is it possible to keep the remaining parameter lists on the same line as it otherwise kinda hides them?
```scala
final def schedule(
    initialDelay: FiniteDuration,
    interval: FiniteDuration,
    receiver: ActorRef,
    message: Any)(implicit executor: ExecutionContext, sender: ActorRef = Actor.noSender): Cancellable =
```

akka-actor/src/main/scala/akka/actor/FSM.scala
ktoso: Hah, you're right - it adds the `public abstract akka.actor.FSM<S, D>.$u2192$ $u2192();` method of course, which breaks people extending FSM... :disappointed:

akka-actor/src/main/scala/akka/actor/Scheduler.scala
ktoso: Sadly AFAICS that's not possible with current scalariform.

ktoso:
I think I'll drop the ball on this one again, much damage and other problems here still.

akka-actor/src/main/scala/akka/io/TcpConnection.scala
```diff
@@ -189,7 +189,7 @@ private[io] abstract class TcpConnection(val tcp: TcpExt, val channel: SocketCha
 
   /** used in subclasses to start the common machinery above once a channel is connected */
   def completeConnect(registration: ChannelRegistration, commander: ActorRef,
-                      options: immutable.Traversable[SocketOption]): Unit = {
+    options: immutable.Traversable[SocketOption]): Unit = {
```
2beaucoup: I thought that each parameter starts on a new line if there's one single break in the parameter list. Why doesn't that apply here? What's the logic behind this?

ktoso: no idea actually, I thought it would. I've given up on this PR for the time being though :/

akka-http-core/src/main/scala/akka/http/impl/engine/client/PoolGateway.scala
```diff
@@ -38,9 +39,11 @@ private object PoolGateway {
...
  * get reused will automatically forward requests directed at them to the latest pool incarnation from the cache.
  */
-private[http] class PoolGateway(hcps: HostConnectionPoolSetup,
-                                _shutdownStartedPromise: Promise[Done])( // constructor arg only
-                                  implicit system: ActorSystem, fm: Materializer) {
+private[http] class PoolGateway(
+  hcps: HostConnectionPoolSetup,
+    _shutdownStartedPromise: Promise[Done])( // constructor arg only
+    implicit
+    system: ActorSystem, fm: Materializer) {
```
2beaucoup: here too

akka-actor/src/main/scala/akka/actor/Scheduler.scala
2beaucoup: Looks like this would require manual interaction. :(

akka-actor/src/main/scala/akka/io/TcpConnection.scala
2beaucoup: I see. But worth a try anyway

PR #20235, opened 2016-04-05
Fix last emit in groupedWithin

### LINKED ISSUE #20232: groupedWithin does not emit a last element when n == 1
In the following code, the first example should print up to 10, but gets up to 9.

```scala
Source(0 to 10).groupedWithin(1, 1 seconds).runForeach(println)
/*
Vector(0)
Vector(1)
Vector(2)
Vector(3)
Vector(4)
Vector(5)
Vector(6)
Vector(7)
Vector(8)
Vector(9)
*/

Source(0 to 10).groupedWithin(2, 1 seconds).runForeach(println)
/*
Vector(0, 1)
Vector(2, 3)
Vector(4, 5)
Vector(6, 7)
Vector(8, 9)
Vector(10)
*/
```

I confirmed this happened in Akka 2.4.2 and 2.4.3.

drewhk:
fixes https://github.com/akka/akka/issues/20232

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowGroupedWithinSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowGroupedWithinSpec.scala
@@ -6,11 +6,12 @@ package akka.stream.scaladsl
 import scala.collection.immutable
 import scala.concurrent.duration._
 import scala.concurrent.forkjoin.ThreadLocalRandom.{ current ⇒ random }
-import akka.stream.ActorMaterializer
-import akka.stream.ActorMaterializerSettings
+import akka.stream.{ ActorMaterializer, ActorMaterializerSettings, ThrottleMode }
 import akka.stream.testkit._
 import akka.stream.testkit.Utils._
 import akka.testkit.AkkaSpec
+
+import scala.concurrent.Await
 
 class FlowGroupedWithinSpec extends AkkaSpec with ScriptedTest {
 
@@ -133,6 +134,13 @@ class FlowGroupedWithinSpec extends AkkaSpec with ScriptedTest {
       def script = Script((TestConfig.RandomTestRange.map { _ ⇒ val x, y, z = random.nextInt(); Seq(x, y, z) -> Seq(immutable.Seq(x, y, z)) }
         :+ { val x = random.nextInt(); Seq(x) -> Seq(immutable.Seq(x)) }): _*)
       TestConfig.RandomTestRange foreach (_ ⇒ runScript(script, settings)(_.groupedWithin(3, 10.minutes)))
+    }
+
+    "group with small groups with backpressure" in {
+      Source(1 to 10)
+        .groupedWithin(1, 1.day)
+        .throttle(1, 110.millis, 0, ThrottleMode.Shaping)
+        .runWith(Sink.seq).futureValue should ===((1 to 10).map(List(_)))
     }
 
   }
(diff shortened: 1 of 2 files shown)


akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -956,13 +956,15 @@ private[stream] final class GroupedWithin[T](n: Int, d: FiniteDuration) extends
   override def initialAttributes = DefaultAttributes.groupedWithin
   val shape = FlowShape(in, out)
 
-  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new TimerGraphStageLogic(shape) {
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new TimerGraphStageLogic(shape) with InHandler with OutHandler {
```
patriknw: ah, that is a nice trick, also makes things more readable, I think

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowGroupedWithinSpec.scala
```diff
@@ -135,6 +136,13 @@ class FlowGroupedWithinSpec extends AkkaSpec with ScriptedTest {
       TestConfig.RandomTestRange foreach (_ ⇒ runScript(script, settings)(_.groupedWithin(3, 10.minutes)))
     }
 
+    "group with small groups with backpressure" in {
+      Source(1 to 10)
+        .groupedWithin(1, 1.second)
+        .throttle(1, 110.millis, 0, ThrottleMode.Shaping)
+        .runWith(Sink.seq).futureValue should ===((1 to 10).map(List(_)))
```
patriknw: how sensitive is this for GC jitter?

patriknw:
LGTM, after squash

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
drewhk: yeah, although I introduced this to save on memory (this is one of the reason for the ThreadLocal interpreter, so these methods can provide default implementations for the finish/fail methods and still be stackable traits)

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowGroupedWithinSpec.scala
drewhk: this takes 1.1 seconds normally and futureValue times out in 3 seconds. I think that should be fine?

drewhk:
Oh lol, forgot to amend last commit. Will squash after others reviewed.

akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowGroupedWithinSpec.scala
patriknw: I was more thinking about if `110.millis` could result in different grouping than expected because of a GC

drewhk: good point, I make the group time to be 1 day and then it will not interfere (it is irrelevant for this test).

patriknw: thanks

johanandren:
LGTM

Issue #19993, opened 2016-03-10
FAILED FlowThrottleSpec


johanandren:
https://jenkins.akka.io:8498/job/akka-nightly/3125/consoleFull
```
[info] - must send elements downstream as soon as time comes *** FAILED *** (655 milliseconds)
[info]   List(1) was not equal to List(1, 2) (FlowThrottleSpec.scala:181)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.MatchersHelper$.newTestFailedException(MatchersHelper.scala:160)
[info]   at org.scalatest.Matchers$ShouldMethodHelper$.shouldMatcher(Matchers.scala:6231)
[info]   at org.scalatest.Matchers$AnyShouldWrapper.should(Matchers.scala:6265)
[info]   at akka.stream.scaladsl.FlowThrottleSpec$$anonfun$2$$anonfun$apply$mcV$sp$15$$anonfun$apply$mcV$sp$35.apply(FlowThrottleSpec.scala:181)
[info]   at akka.stream.scaladsl.FlowThrottleSpec$$anonfun$2$$anonfun$apply$mcV$sp$15$$anonfun$apply$mcV$sp$35.apply(FlowThrottleSpec.scala:178)
[info]   at akka.stream.testkit.Utils$.assertAllStagesStopped(Utils.scala:25)
[info]   at akka.stream.scaladsl.FlowThrottleSpec$$anonfun$2$$anonfun$apply$mcV$sp$15.apply$mcV$sp(FlowThrottleSpec.scala:178)
[info]   at akka.stream.scaladsl.FlowThrottleSpec$$anonfun$2$$anonfun$apply$mcV$sp$15.apply(FlowThrottleSpec.scala:178)
[info]   at akka.stream.scaladsl.FlowThrottleSpec$$anonfun$2$$anonfun$apply$mcV$sp$15.apply(FlowThrottleSpec.scala:178)
...
```

PR #20242, opened 2016-04-05
=str #19993 Less tight limits in one throttle test

johanandren:
Increases the timings with 1.5 in the test failing in #19993

### DIFF
--- a/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowThrottleSpec.scala
+++ b/akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowThrottleSpec.scala
@@ -105,12 +105,12 @@ class FlowThrottleSpec extends AkkaSpec {
     }
 
     "send elements downstream as soon as time comes" in Utils.assertAllStagesStopped {
-      val probe = Source(1 to 10).throttle(2, 500.millis, 0, Shaping).runWith(TestSink.probe[Int])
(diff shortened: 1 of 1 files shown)


akka-stream-tests/src/test/scala/akka/stream/scaladsl/FlowThrottleSpec.scala
```diff
@@ -105,12 +105,12 @@ class FlowThrottleSpec extends AkkaSpec {
     }
 
     "send elements downstream as soon as time comes" in Utils.assertAllStagesStopped {
-      val probe = Source(1 to 10).throttle(2, 500.millis, 0, Shaping).runWith(TestSink.probe[Int])
+      val probe = Source(1 to 10).throttle(2, 750.millis, 0, Shaping).runWith(TestSink.probe[Int])
```
drewhk: why change this parameter? This makes the rate less.

johanandren: The idea was to increase all the timings by 0.5 to give more room for variations on the CI server, but maybe I thought about it the wrong way. (The other timings depend on this, so cannot be changed independently).

Couldn't think of a better way to stabilize this test as it is testing timing things.

ktoso:
LGTM

Issue #20228, opened 2016-04-04
Creating multiple https requests fail because of race condition


markvandertol:
I tried to use akka-http to open multiple https connections at once, but most of the time this fails with an exception. Also, when I insert a delay between the requests it will always succeed.

I used akka-http-core, version 2.4.3 on Java Hotspot 1.8.0_77 (64-bits).

Sample code to reproduce the issue:
```
import akka.actor.ActorSystem
import akka.http.scaladsl.Http
import akka.http.scaladsl.model.{ HttpRequest, Uri}
import akka.stream.ActorMaterializer

implicit val system: ActorSystem = ActorSystem("my-system")
implicit val materializer = ActorMaterializer()
implicit val ec = system.dispatcher
val http = Http()
val request = HttpRequest(uri = Uri("https://www.google.com"))
http.singleRequest(request).onComplete(r => println(r))
http.singleRequest(request).onComplete(r => println(r))
```

The Futures don't complete and the following errors are logged:
```
[ERROR] [04/04/2016 21:50:43.770] [my-system-akka.actor.default-dispatcher-2] [akka://my-system/user/StreamSupervisor-1/flow-1-2-SlotProcessorInternalConnectionFlow] Duplicated server name of type 0
akka.actor.ActorInitializationException: akka://my-system/user/StreamSupervisor-1/flow-1-2-SlotProcessorInternalConnectionFlow: exception during creation
	at akka.actor.ActorInitializationException$.apply(Actor.scala:174)
	at akka.actor.ActorCell.create(ActorCell.scala:606)
	at akka.actor.ActorCell.invokeAll$1(ActorCell.scala:461)
	at akka.actor.ActorCell.systemInvoke(ActorCell.scala:483)
	at akka.dispatch.Mailbox.processAllSystemMessages(Mailbox.scala:282)
	at akka.dispatch.Mailbox.run(Mailbox.scala:223)
	at akka.dispatch.Mailbox.exec(Mailbox.scala:234)
	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
	at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
	at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
[...]

PR #20243, opened 2016-04-06
Fix race condition when setting up client https connection #20228

markvandertol:
As discussed in #20228

Fixes a race condition when multiple connections are setup simultaneously that could result in an error when setting up a https connection or sending the wrong SNI to the server.

A better fix would have been to not pass around a mutable SSLParameters object between actors, but that cannot be done without breaking the public api.

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala
@@ -165,6 +165,22 @@ private[akka] class TLSActor(settings: ActorMaterializerSettings,
     params.setServerNames(Collections.singletonList(serverName))
   }
 
+  private def cloneSSLParameters(parameters: SSLParameters): SSLParameters = {
+    val clone = new SSLParameters()
+
+    clone.setCipherSuites(parameters.getCipherSuites)
+    clone.setProtocols(parameters.getProtocols)
+    clone.setWantClientAuth(parameters.getWantClientAuth)
+    clone.setNeedClientAuth(parameters.getNeedClientAuth)
+    clone.setEndpointIdentificationAlgorithm(parameters.getEndpointIdentificationAlgorithm)
+    clone.setAlgorithmConstraints(parameters.getAlgorithmConstraints)
+    clone.setServerNames(parameters.getServerNames)
+    clone.setSNIMatchers(parameters.getSNIMatchers)
+    clone.setUseCipherSuitesOrder(parameters.getUseCipherSuitesOrder)
+
+    clone
+  }
+
   var currentSession = engine.getSession
   applySessionParameters(firstSession)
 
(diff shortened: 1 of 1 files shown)


ktoso:
OK TO TEST

ktoso:
LGTM – Thanks for the fix (and finding it in the first place)!

The better fix mentioned is tricky to pull off, I wonder if we really would be able to do it in a nice way.
On one hand, letting users set "the usual things" on SSLContext is "known territory" for many, so it's a nice api to expose; but the mutability is a pain as seen here... We'd have to otherwise expose our own immutable APIs for passing in more config, that we apply to the context in the actual Actor I guess.

akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala
```diff
@@ -165,6 +165,22 @@ private[akka] class TLSActor(settings: ActorMaterializerSettings,
...
+    clone.setWantClientAuth(parameters.getWantClientAuth)
+    clone.setNeedClientAuth(parameters.getNeedClientAuth)
+    clone.setEndpointIdentificationAlgorithm(parameters.getEndpointIdentificationAlgorithm)
+    clone.setAlgorithmConstraints(parameters.getAlgorithmConstraints)
+    clone.setServerNames(parameters.getServerNames)
+    clone.setSNIMatchers(parameters.getSNIMatchers)
+    clone.setUseCipherSuitesOrder(parameters.getUseCipherSuitesOrder)
+
+    clone
+  }
```
2beaucoup: This can be done in one go without cloning the params if `hostInfo` is `None`:

```scala
private def applySNI: Unit = for {
  params <- params.sslParameters
  (hostname, _) <- hostInfo
} yield {
  //first copy the mutable SLLParameters before modifying to prevent race condition in `setServerNames`
  val clone = new SSLParameters()
  clone.setCipherSuites(params.getCipherSuites)
  clone.setProtocols(params.getProtocols)
  clone.setWantClientAuth(params.getWantClientAuth)
  clone.setNeedClientAuth(params.getNeedClientAuth)
  clone.setEndpointIdentificationAlgorithm(params.getEndpointIdentificationAlgorithm)
  clone.setAlgorithmConstraints(params.getAlgorithmConstraints)
  clone.setSNIMatchers(params.getSNIMatchers)
  clone.setUseCipherSuitesOrder(params.getUseCipherSuitesOrder)

  // apply the changes
  clone.setServerNames(Collections.singletonList(new SNIHostName(hostname)))
  engine.setSSLParameters(parameters)
}
```

2beaucoup: `setServerNames` is called only once here.

ktoso: I'll be touching this area in a PR in the next hour or so, will apply this hint – thanks @2beaucoup!

ktoso:
Thanks again @markvandertol :+1:

akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala
2beaucoup: :+1:

PR #20244, opened 2016-04-06
=htc #19834 convert StreamUtils.identityFinishReporter to GraphStage

leachbj:
Last of the deprecations for StreamUtils!

### DIFF
--- a/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
@@ -226,9 +226,9 @@ private[http] object StreamUtils {
    * is finished, only that the part that contains this flow has finished work.
    */
   def identityFinishReporter[T]: Flow[T, T, Future[Unit]] = {
-    Flow[T].viaMat(new GraphStageWithMaterializedValue[FlowShape[T, T], Future[Unit]] {
+    object IdentityFinishReporter extends GraphStageWithMaterializedValue[FlowShape[T, T], Future[Unit]] {
       val shape = FlowShape(Inlet[T]("identityFinishReporter.in"), Outlet[T]("identityFinishReporter.out"))
-      override def toString: String = "UniqueKillSwitchFlow"
+      override def toString: String = "IdentityFinishReporter"
 
       def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, Future[Unit]) = {
         val promise = Promise[Unit]()
@@ -247,13 +247,14 @@ private[http] object StreamUtils {
             promise.trySuccess(())
           }
 
-          setHandler(shape.in, this)
-          setHandler(shape.out, this)
+          setHandlers(shape.in, shape.out, this)
         }
 
         (stage, promise.future)
       }
-    })(Keep.right)
+    }
+
+    Flow[T].viaMat(IdentityFinishReporter)(Keep.right)
   }
 
   /**
--- a/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
@@ -4,18 +4,17 @@
 
 package akka.http.impl.util
 
-import java.util.concurrent.atomic.{AtomicBoolean, AtomicReference}
-import akka.{Done, NotUsed}
+import java.util.concurrent.atomic.{ AtomicBoolean, AtomicReference }
+import akka.NotUsed
 import akka.stream._
 import akka.stream.impl.fusing.GraphStages.SimpleLinearGraphStage
-import akka.stream.impl.{PublisherSink, SinkModule, SourceModule}
+import akka.stream.impl.{ PublisherSink, SinkModule, SourceModule }
 import akka.stream.scaladsl._
 import akka.stream.stage._
-import akka.stream.stage.GraphStageWithMaterializedValue
 import akka.util.ByteString
-import org.reactivestreams.{Processor, Publisher, Subscriber, Subscription}
+import org.reactivestreams.{ Processor, Publisher, Subscriber, Subscription }
 
-import scala.concurrent.{ExecutionContext, Future, Promise}
+import scala.concurrent.{ ExecutionContext, Future, Promise }
 
 /**
  * INTERNAL API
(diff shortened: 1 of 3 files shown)


ktoso:
OK TO TEST

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -224,38 +226,34 @@ private[http] object StreamUtils {
    * is finished, only that the part that contains this flow has finished work.
    */
   def identityFinishReporter[T]: Flow[T, T, Future[Unit]] = {
```
viktorklang: Argh, did we miss to make this a Future[akka.Done] ? :/

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -224,38 +226,34 @@ private[http] object StreamUtils {
...
 
-        override def decide(cause: Throwable): Supervision.Directive = {
-          // supervision will be implemented by #16916
-          promise.tryFailure(cause)
-          super.decide(cause)
+          override def postStop(): Unit = {
+            promise.trySuccess(())
+          }
+
+          setHandler(shape.in, this)
```
viktorklang: `setHandlers(shape.in, shape.out, this)`

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -224,38 +226,34 @@ private[http] object StreamUtils {
    * is finished, only that the part that contains this flow has finished work.
    */
   def identityFinishReporter[T]: Flow[T, T, Future[Unit]] = {
-    // copy from Sink.foreach
-    def newForeachStage(): (PushStage[T, T], Future[Unit]) = {
-      val promise = Promise[Unit]()
+    Flow[T].viaMat(new GraphStageWithMaterializedValue[FlowShape[T, T], Future[Unit]] {
```
viktorklang: Instead of creating a new GraphStageWithMaterializedValue, make it an `object IdentityFinishReporter extends GraphStageWithMaterializedValue …`

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -224,38 +226,34 @@ private[http] object StreamUtils {
...
+      val shape = FlowShape(Inlet[T]("identityFinishReporter.in"), Outlet[T]("identityFinishReporter.out"))
+      override def toString: String = "UniqueKillSwitchFlow"
```
viktorklang: This String can't be right

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -224,38 +226,34 @@ private[http] object StreamUtils {
...
 
-      val stage = new PushStage[T, T] {
-        override def onPush(elem: T, ctx: Context[T]): SyncDirective = ctx.push(elem)
+      def createLogicAndMaterializedValue(inheritedAttributes: Attributes): (GraphStageLogic, Future[Unit]) = {
+        val promise = Promise[Unit]()
```
viktorklang: Shouldn't this override `initialAttributes`?

leachbj: I don't think it needs to.  In this class the only other one that does is limitByteChunkStage which has `override def initialAttributes = Attributes.name("limitByteChunksStage")`

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
ktoso: Two good pieces of news here:
- It's in internal api, we could change if it we want to
- we want to remove the entire StreamUtils object (which contains this), since those were workarounds for missing Streams features – so the goal is to remove these workaround-thingies all together

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
viktorklang: Since InitialFinishReporter should be an object, then initialAttributes should be a val iso def (to avoid allocating the name for each use)

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -224,38 +226,34 @@ private[http] object StreamUtils {
...
-          ctx.finish()
-        }
+          override def onPull(): Unit = pull(shape.in)
 
-        override def onDownstreamFinish(ctx: Context[T]): TerminationDirective = {
-          promise.success(())
-          ctx.finish()
-        }
+          override def onUpstreamFailure(ex: Throwable): Unit = {
+            promise.failure(ex)
```
viktorklang: Shouldn't this be `tryFailure` as in the original code?

johanandren: It wasn't, we´re in `onUpstreamFailure` here, `tryFailure` is for exceptions originating from this stage.

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -223,39 +225,36 @@ private[http] object StreamUtils {
    * completion or error signal. It doesn't necessarily mean, though, that all of a streaming pipeline
    * is finished, only that the part that contains this flow has finished work.
    */
-  def identityFinishReporter[T]: Flow[T, T, Future[Unit]] = {
-    // copy from Sink.foreach
-    def newForeachStage(): (PushStage[T, T], Future[Unit]) = {
-      val promise = Promise[Unit]()
+  def identityFinishReporter[T]: Flow[T, T, Future[Done]] = {
```
johanandren: Couldn't this just be replaced by using `Flow.watchTermination` in the call sites instead of this custom stage?

leachbj: watchTermination looks pretty much like this code.

johanandren: Yes, so no need to have this duplicate implementation of it here. Let's just use that instead.

leachbj: The latest commit in the chain does that.

viktorklang: Nice find, @johanandren, and great work @leachbj !

johanandren:
LGTM. Thanks @leachbj!

akka-http-core/src/main/scala/akka/http/impl/util/StreamUtils.scala
```diff
@@ -219,46 +220,6 @@ private[http] object StreamUtils {
...
-        }
-      }
-
-      (stage, promise.future)
-    }
-    Flow[T].transformMaterializing(newForeachStage)
-  }
-
-  /**
    * Similar to Source.maybe but doesn't rely on materialization. Can only be used once.
```
patriknw: nice

patriknw:
LGTM

viktorklang:
*AMAZING* ratio of lines added vs removed while keeping the feature intact!

PR #20252, opened 2016-04-06
htp #20214 allow workaround for disabling SNI

### LINKED ISSUE #20214: Disabling per-connection SNI doesn't work as of akka 2.4.5
Hi!

In some cases I need to disable SNI entirely when using akka-http.
In 2.4.2 I used this code to create a custom connection context:
```
  protected def httpsContext: HttpsConnectionContext =
  {
    val ctx = sslContext
    val params = ctx.getDefaultSSLParameters

    /* Disable SNI if asked to */
    if (!sniEnabled) {
      params.setServerNames(Nil.asJava)
      params.setSNIMatchers(Nil.asJava)
    }

     new HttpsConnectionContext(ctx, sslParameters = Some(params))
   }
```

However as of 2.4.3 this no longer works and gives an "andshake alert:  unrecognized_name" error.

It seems this is due to https://github.com/akka/akka/pull/19858
that overrides the serverNames that I have set.

This seems like an unintended consequence, or am I missing something?

ktoso:
This resolves https://github.com/akka/akka/issues/20214 
however it depends on ssl-config 2.0 release (where we'd like to drop JDK6/7 while we're at it).

The big warnings are logged once per Http extension instance (so basically once per app), this PR will also get documentation how to do the "more right thing", i.e. not disable these things globally, but only for a given connection.

### DIFF
--- a/akka-docs/rst/java/http/client-side/https-support.rst
+++ b/akka-docs/rst/java/http/client-side/https-support.rst
@@ -45,8 +45,33 @@ to rely on the configured default client-side ``HttpsContext``.
 If no custom ``HttpsContext`` is defined the default context uses Java's default TLS settings. Customizing the
 ``HttpsContext`` can make the Https client less secure. Understand what you are doing!
 
+Detailed configuration and workarounds
+--------------------------------------
+
+Akka HTTP relies on `Typesafe SSL-Config`_ which is a library maintained by Lightbend that makes configuring
+things related to SSL/TLS much simpler than using the raw SSL APIs provided by the JDK. Please refer to its
+documentation to learn more about it.
+
+All configuration options available to this library may be set under the ``akka.ssl-context`` configuration for Akka HTTP applications.
+
+.. note::
+  When encountering problems connecting to HTTPS hosts we highly encourage to reading up on the excellent ssl-config
+  configuration. Especially the quick start sections about `adding certificates to the trust store`_ should prove
+  very useful, for example to easily trust a self-signed certificate that applications might use in development mode.
+
+.. warning::
+  While it is possible to disable certain checks using the so called "loose" settings in SSL Config, we **strongly recommend**
+  to instead attempt to solve these issues by properly configuring TLS–for example by adding trusted keys to the keystore.
+
+  If however certain checks really need to be disabled because of misconfigured (or legacy) servers that your
+  application has to speak to, instead of disabling the checks globally (i.e. in ``application.conf``) we suggest
+  configuring the loose settings for *specific connections* that are known to need them disabled (and trusted for some other reason).
+  The pattern of doing so is documented in the folowing sub-sections.
+
+.. _adding certificates to the trust store: http://typesafehub.github.io/ssl-config/WSQuickStart.html#connecting-to-a-remote-server-over-https
+
 Hostname verification
----------------------
+^^^^^^^^^^^^^^^^^^^^^
 
 Hostname verification proves that the Akka HTTP client is actually communicating with the server it intended to
 communicate with. Without this check a man-in-the-middle attack is possible. In the attack scenario, an alternative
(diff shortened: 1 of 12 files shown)


akka-http-core/src/test/scala/akka/http/scaladsl/SslConfigWarningsSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+      val p = TestProbe()
+      system.eventStream.subscribe(p.ref, classOf[Logging.LogEvent])
+
+      EventFilter.warning(start = "Detected that Server Name Indication (SNI) is disabled globally", occurrences = 1) intercept {
+        Http()(system)
+      }
+
+      // the very big warning shall be logged only once per actor system (extension)
+      EventFilter.warning(start = "Detected that Server Name Indication (SNI) is disabled globally", occurrences = 0) intercept {
+        Http()(system)
```
patriknw: Is the warning really logged each time the extension is accessed like this? Otherwise you should probably nest the event filters around one such call.

akka-http-core/src/test/scala/akka/http/scaladsl/SslConfigWarningsSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+      }
+
+      Await.ready(system.terminate(), 3.seconds)
```
patriknw: you can still use TestKit.shutdown

akka-http-core/src/test/scala/akka/http/scaladsl/SslConfigWarningsSpec.scala
ktoso: No, it's just logged once per extension (warning logging runs in constructor or HttpExt, so once per AS)

patriknw:
LGTM, aside from nitpicks

ktoso:
ssl-config released, this should pass now (adderssed comments)

ktoso:
Failed because I forgot to publish ssl-config-akka (did just ssl-config), published it now.

akka-docs/rst/java/http/client-side/https-support.rst
```diff
@@ -45,8 +45,33 @@ to rely on the configured default client-side ``HttpsContext``.
...
+
+All configuration options available to this library may be set under the `akka.ssl-context` configuration for Akka HTTP applications.
+
+.. note::
+  When encountering problems connecting to HTTPS hosts we highly encourage to reading up on the excellent ssl-config
+  configuration. Especially the quick start sections about `adding certificates to the trust store`_ should prove
+  very useful, for example to easily trust a self-signed certificate that applications might use in development mode.
+
+.. warning::
+  While it is possible to disable certain checks using the so called "loose" settings in SSL Config, we **strongly recomment**
```
johanandren: recommend

akka-docs/rst/java/http/client-side/https-support.rst
```diff
@@ -45,8 +45,33 @@ to rely on the configured default client-side ``HttpsContext``.
...
+  to instead attempt to solve these issues by properly configuring TLS–for example by adding trusted keys to the keystore.
+
+  If however certain checks really need to be disabled because of misconfigured (or legacy) servers that your
+  applicaion has to speak to, instead of disabling the checks globally (i.e. in `application.conf`) we suggest
```
johanandren: markdown code quotes in rst file?

johanandren:
LGTM except for some doc nitpicks

ktoso:
Thanks, fixed!

ktoso:
PLS BUILD

akka-docs/rst/java/http/client-side/https-support.rst
```diff
@@ -45,8 +45,33 @@ to rely on the configured default client-side ``HttpsContext``.
...
+  configuration. Especially the quick start sections about `adding certificates to the trust store`_ should prove
+  very useful, for example to easily trust a self-signed certificate that applications might use in development mode.
+
+.. warning::
+  While it is possible to disable certain checks using the so called "loose" settings in SSL Config, we **strongly recommend**
+  to instead attempt to solve these issues by properly configuring TLS–for example by adding trusted keys to the keystore.
+
+  If however certain checks really need to be disabled because of misconfigured (or legacy) servers that your
+  applicaion has to speak to, instead of disabling the checks globally (i.e. in ``application.conf``) we suggest
+  configuring more loose for *specific connections* that are known to need them disabled (and trusted for some other reason).
```
2beaucoup: "...configuring more loose for..."?

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -753,7 +754,30 @@ trait DefaultSSLContextCreation {
   protected def system: ActorSystem
   protected def sslConfig: AkkaSSLConfig
 
-  protected def createDefaultClientHttpsContext(): HttpsConnectionContext = {
+  // --- log warnings ---
+  private[this] def log = system.log
+
+  def validateAndWarnAboutLooseSettings() = {
```
2beaucoup: What is validated here?

akka-http-core/src/test/scala/akka/http/scaladsl/SslConfigWarningsSpec.scala
2beaucoup: Doesn't ssl-config itself log warnings?

akka-docs/rst/java/http/client-side/https-support.rst
ktoso: thanks, must have lost my train of thought there

akka-http-core/src/test/scala/akka/http/scaladsl/SslConfigWarningsSpec.scala
ktoso: Yes and this one is specifically added and much louder, discussed it a bit on gitter, story here is:

- ssl-config logs a warning when created and bad things noticed, it's a small oneliner (bad, once per ssl-config)
- akka logs a big fat warning when ssl-config is GLOBALLY (very bad) configured to disable things, instead we recommend setting it up for a single connection 
  - note that if you do so, you'll still get one warning message that for that ssl-config instance (the above "bad" triggers then and logs once).

This is to not hide that you're using loose settings somewhere. Worst case is the global one - 2 log messages will be shown, but the akka one specifically telling about "hey, do this per connection plz"

akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala
```diff
@@ -193,12 +169,9 @@ private[akka] class TLSActor(settings: ActorMaterializerSettings,
...
     }
-    params.sslParameters foreach { p ⇒
-      //first copy the mutable SLLParameters before modifying to prevent race condition
-      val parameters = cloneSSLParameters(p)
-      hostInfo foreach { case (host, _) ⇒ applySNI(host, parameters) }
-      engine.setSSLParameters(parameters)
-    }
+
+    // configure Server Name Indication unless ssl-config disabled it (in which case we already logged many warnings)
+    applySNI(params)
```
2beaucoup: I wonder if this stuff should rather be the responsibility of ssl-config.

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
ktoso: You're right that no validation yet here – shall I rename and go over the MiMa dance?
If we'd validate ssl things it'd be here in this callback so I'd keep as is

akka-http-core/src/main/scala/akka/http/scaladsl/Http.scala
```diff
@@ -753,7 +754,30 @@ trait DefaultSSLContextCreation {
...
+        "If you are forced to interact with a server that is behaving such that you must disable this setting, " +
+        "please disable it for a given connection instead, by configuring a specific HttpsConnectionContext " +
+        "for use only for the trusted target that hostname verification would have blocked.")
+
+    if (sslConfig.config.loose.disableSNI)
+      log.warning("Detected that Server Name Indication (SNI) is disabled globally (via ssl-config) for the Http extension! " +
+        "This is very dangerous and may expose you to man-in-the-middle attacks. " +
+        "If you are forced to interact with a server that is behaving such that you must disable this setting, " +
+        "please disable it for a given connection instead, by configuring a specific HttpsConnectionContext " +
+        "for use only for the trusted target that hostname verification would have blocked.")
```
2beaucoup: Refer to `sslConfig.config.loose.disableSNI` here? Looks like there`s some DRY potential here. ;)

akka-stream/src/main/scala/akka/stream/impl/io/TLSActor.scala
ktoso: Yeah, we discussed this at the point where this was introduced, it's a bit tricky since "where to call it" is half of the problem here.

Previously we could not have put it directly into ssl-config (it would have to use reflection or other hackery) since it was JDK6 supporting, yesterday I moved ssl-config 0.2 to JDK8 in which the `setServerNames` method is available so we could do that move... I'll open a ticket on ssl-config

ktoso:
Addressed comments and rebased

ktoso:
Oh that's silly – failure was missing `,` due to rebase in `MiMa.scala`

ktoso:
TimeoutDirectives failure, seems unrelated but want to be sure.

ktoso:
PLS BUILD

Issue #20263, opened 2016-04-08
NPE in ConsistentHashingRouter for SimpleDnsManager


patriknw:
```
2016-04-08T04:22:43Z ChristophersMBP.lan WARN  ConsistentHashingRoutingLogic ConsistentHashingRoutingLogic(akka://conductr) - Couldn't route message with consistent hash key [logs2.papertrailapp.com] due to [null]
```

Reported by @huntc

patriknw:
A workaround, until next release is out, is to use a a round robin pool instead.
In config:
```
akka.actor.deployment {
  /IO-DNS/inet-address {
    mailbox = "unbounded"
    router = "round-robin-pool"
    nr-of-instances = 4
  }
}
```

huntc:
Thanks for this. Do I need that configuration now then for every Lagom service, every Play 2.5 app and every Akka http based service?

patriknw:
I don't think so. This dns router id not used much, it seems. Note that it was not working at all in Akka 2.4.2 with ClusterActorRefProvider and we didn't see many complaints about that.

I think the usage in conductr is a special, that triggers it in the startup. I think application.conf in conductr will be enough, but you have to try it.

drewhk:
backported

PR #20264, opened 2016-04-08
lazy usage of defaultAddress in ConsistentHashingRouter, #20263

patriknw:
- because consistent hashing routing pool is used by SimpleDnsManager
  and that can be activated early, before the transport defaultAddress
  is set in the startup

### DIFF
--- a/akka-actor/src/main/scala/akka/routing/ConsistentHashing.scala
+++ b/akka-actor/src/main/scala/akka/routing/ConsistentHashing.scala
@@ -149,7 +149,17 @@ final case class ConsistentHashingRoutingLogic(
   def this(system: ActorSystem) =
     this(system, virtualNodesFactor = 0, hashMapping = ConsistentHashingRouter.emptyConsistentHashMapping)
 
-  private val selfAddress = system.asInstanceOf[ExtendedActorSystem].provider.getDefaultAddress
+  private lazy val selfAddress = {
+    // Important that this is lazy, because consistent hashing routing pool is used by SimpleDnsManager
+    // that can be activated early, before the transport defaultAddress is set in the startup.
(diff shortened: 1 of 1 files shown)


patriknw:
Refs #20263

akka-actor/src/main/scala/akka/routing/ConsistentHashing.scala
```diff
@@ -149,7 +149,10 @@ final case class ConsistentHashingRoutingLogic(
   def this(system: ActorSystem) =
     this(system, virtualNodesFactor = 0, hashMapping = ConsistentHashingRouter.emptyConsistentHashMapping)
 
-  private val selfAddress = system.asInstanceOf[ExtendedActorSystem].provider.getDefaultAddress
+  // important that this is lazy, because consistent hashing routing pool is used by SimpleDnsManager
+  // that can be activated early, before the transport defaultAddress is set in the startup
+  // see issue #20263
+  private lazy val selfAddress = system.asInstanceOf[ExtendedActorSystem].provider.getDefaultAddress
```
huntc: Could the default address still be `null` in the case of some race condition?

huntc: Perhaps `selfAddress` should be an `Option`...

patriknw: You are right, we can do better. I have changed it slightly so that if it happens the message will not be routed, but it will try again for next message, see https://github.com/akka/akka/pull/20264/files#diff-0931c33dd37d5c175c80c0b6244f3d95R162

johanandren:
LGTM

Issue #19688, opened 2016-02-03
[2.4.1][PersistentFSM] getting StateTimeout for state without timeout after restart


LoneEngineer:
Hi,

The behavior, I observed, looks like bug, but I'm not sure at which side (akka or mine). Behavior of StateTimeout for persistent FSM is not exposed in the documentation very clearly (or may be I missed something).

So the case is following: a persistentFSM actor has a short-lived state (with small timeout) and long-lived state (long or infinite timeout). After restart the actor receives StateTimeout in long-lived state (which is unexpected) if before restart actor visited short-lived state.

I tried to get minimal app (scalatest actually) which shows this behavior. It uses inmem db, but I tried with LevelDB as well - nothing changes.

https://github.com/LoneEngineer/akka-persistent-fsm-test

patriknw:
This might be related to https://github.com/akka/akka/issues/18207

Any thoughts, @leonidb ?

leonidb:
@LoneEngineer I've tried to run the tests, they pass if I run them both in one execution. 

If I run them one by one, I get a timeout in the second test, because the tests are configured to run with in memory persistence journal - so the actor is in the ```StateOne``` after recovery, so the timeout is triggered.

How do you execute the tests?

LoneEngineer:
@leonidb Hi, of course I ran them both in one execution (I tried both in sbt console (output below) and under IDEA 15.0.1).

$ sbt test
[info] Loading global plugins from /home/sergei/.sbt/0.13/plugins
[info] Updating {file:/home/sergei/.sbt/0.13/plugins/}global-plugins...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
[info] Set current project to akka-test (in build file:/home/sergei/Work/github/akka-persistent-fsm-test/)
[info] Updating {file:/home/sergei/Work/github/akka-persistent-fsm-test/}akka-persistent-fsm-test...
[info] Resolving jline#jline;2.12.1 ...
[info] Done updating.
[info] Compiling 1 Scala source to /home/sergei/Work/github/akka-persistent-fsm-test/target/scala-2.11/classes...
[info] Compiling 1 Scala source to /home/sergei/Work/github/akka-persistent-fsm-test/target/scala-2.11/test-classes...
[info] TestApp:
Creating actor before reload
[...]

PR #20266, opened 2016-04-08
=per #19688 [2.4.1][PersistentFSM] getting StateTimeout for state wit…

leonidb:
…hout timeout after restart

### DIFF
--- a/akka-persistence/src/main/scala/akka/persistence/fsm/PersistentFSMBase.scala
+++ b/akka-persistence/src/main/scala/akka/persistence/fsm/PersistentFSMBase.scala
@@ -301,15 +301,14 @@ trait PersistentFSMBase[S, D, E] extends Actor with Listeners with ActorLogging 
     handleEvent = stateFunction orElse handleEventDefault
 
   /**
-   * Verify existence of initial state and setup timers. This should be the
-   * last call within the constructor, or [[akka.actor.Actor#preStart]] and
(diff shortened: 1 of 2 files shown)


leonidb:
Calling ```initialize()``` in a constructor of ```PersistentFSM``` is redundant, as it's called after recovery. It also might lead to bugs like the one from #19688.
I've made the ```initialize()``` method package private. This will cause a compilation errors (very easy to fix) during upgrade for users who do use this method in a constructor of ```PersistentFSM```.

ktoso:
OK TO TEST

ktoso:
Uff, good that I just remembered that we're experimental here so no trouble in changing this.

Note for reviewers, this class has such note:
`This is an EXPERIMENTAL feature and is subject to change until it has received more real world testing.`

Perhaps adding a `@deprecated` on it as well would be a good idea - for people who used and called it explicitly it would show up as compile error and in the IDEs it'd show up as deprecated use.

leonidb:
@ktoso how about something like:
```scala
@deprecated("Removed from API, called internally", "2.4.4")
```

patriknw:
PLS REBUILD

patriknw:
@leonidb please add that deprecation, otherwise looks good

patriknw:
thanks, @leonidb

patriknw:
Refs #19688

PR #20267, opened 2016-04-09
str #20262 reduce should fail explicitly on empty stream

### LINKED ISSUE #20262: Reduce emits null if used with AnyRef and an empty source
Reduce initially uses a default value as aggregate:

```Scala
var aggregator: T = _
```

Which means that it is `null` for non-primitives. If upstream closes before emitting any element the `aggregator` value is emitted downstream that emits a `null` downstream which is illegal.

I think it should just close rather than pass the empty value downstream, if you want a seed that is emitted on empty upstream you can always use `fold` instead.

ktoso:
Was a fun fix for a change :-)
Resolves https://github.com/akka/akka/issues/20262

### DIFF
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -1168,6 +1168,9 @@ private[stream] final class Reduce[T](f: (T, T) ⇒ T) extends SimpleLinearGraph
         pull(in)
         setHandler(in, rest)
       }
+      override def onUpstreamFinish(): Unit = {
+        failStage(new NoSuchElementException("reduce over empty stream"))
+      }
     })
     def rest = new InHandler {
       override def onPush(): Unit = {
--- a/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
+++ b/akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
@@ -1158,34 +1158,36 @@ private[stream] final class DropWithin[T](timeout: FiniteDuration) extends Simpl
 private[stream] final class Reduce[T](f: (T, T) ⇒ T) extends SimpleLinearGraphStage[T] {
   override def initialAttributes: Attributes = DefaultAttributes.reduce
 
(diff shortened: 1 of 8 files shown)


viktorklang:
@ktoso oops, didn't see you already tackled this. Ohwell, please feel encouraged to steal my performance improvemens from https://github.com/akka/akka/pull/20268

ktoso:
No problem :) thanks for the opt, I'll steal it into here :+1:

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -1168,6 +1168,9 @@ private[stream] final class Reduce[T](f: (T, T) ⇒ T) extends SimpleLinearGraph
         pull(in)
         setHandler(in, rest)
       }
+      override def onUpstreamFinish(): Unit = {
+        failStage(new NoSuchElementException("reduce over empty stream"))
+      }
```
johanandren: So simple fix it's almost suspicious ;)

johanandren:
LGTM

akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
viktorklang: Yep, now it only needs the opts from my PR :)

akka-stream/src/main/scala/akka/stream/scaladsl/Source.scala
```diff
@@ -107,6 +107,11 @@ final class Source[+Out, +Mat](private[stream] override val module: Module)
    * The returned [[scala.concurrent.Future]] will be completed with value of the final
    * function evaluation when the input stream ends, or completed with `Failure`
    * if there is a failure signaled in the stream.
+   *
+   * If the stream is empty (i.e. completes before signalling any elements),
+   * the reduce stage will fail its downstream with an [[NoSuchElementException]],
```
patriknw: s/an/a/

patriknw:
LGTM

ktoso:
Will merge after validation

PR #20268, opened 2016-04-09
=str - #20262 - Makes reduce on empty stream throw an UOE as the stdl…

viktorklang:
…ib does


akka-stream/src/main/scala/akka/stream/impl/fusing/Ops.scala
```diff
@@ -1158,32 +1158,36 @@ private[stream] final class DropWithin[T](timeout: FiniteDuration) extends Simpl
...
+          aggregator = grab(in)
+          pull(in)
+          setHandler(in, self)
+        }
+
+        override def onUpstreamFinish(): Unit =
+          failStage(new UnsupportedOperationException("empty stream reduce"))
+      })
+
+      setHandler(out, self)
```
ktoso: I was a bit iffy about using self here but it actually looks good and should be worth it, thanks :) 
As suggested in the other PR I'll copy that optimisation :+1:

ktoso:
https://github.com/akka/akka/pull/20267 won the horse race :)

Thanks Viktor!

Issue #20273, opened 2016-04-10
EitherUnmarshaller could be a neat idea?


ktoso:
Following up to akka-user thread [Un-marshalling Either](https://groups.google.com/forum/#!topic/akka-user/LWYvvGyDLow), I had an idea that we could actually provide that one, as simplest possible "if this doesnt work, try the other one" unmarshaller.

Implementing unmarshallers is rather tricky for newbies, as seen in that comment so I think this may actually be worthwhile.

PR #20274, opened 2016-04-10
+htc #20273 EitherUnmarshaller

ktoso:
Resolves https://github.com/akka/akka/issues/20273 if we want to.
Seems like a nice addition to me but would welcome comments from @jrudolph @sirthias or @akka/akka-team 

Related blog post http://kto.so/2016/04/10/hakk-the-planet-implementing-akka-http-marshallers/

### DIFF
--- a/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
+++ b/akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
@@ -4,6 +4,7 @@
 
 package akka.http.scaladsl.unmarshalling
 
+import akka.event.Logging
 import akka.stream.Materializer
 
 import scala.util.control.{ NoStackTrace, NonFatal }
@@ -140,6 +141,15 @@ object Unmarshaller
    */
   final case class UnsupportedContentTypeException(supported: Set[ContentTypeRange])
     extends RuntimeException(supported.mkString("Unsupported Content-Type, supported: ", ", ", ""))
+
+  /** Order of parameters (`right` first, `left` second) is intentional, since that's the order we evaluate them in. */
+  final case class EitherUnmarshallingException(
+    rightClass: Class[_], right: Throwable,
+    leftClass: Class[_], left: Throwable)
+    extends RuntimeException(
+      s"Failed to unmarshal Either[${Logging.simpleName(leftClass)}, ${Logging.simpleName(rightClass)}] (attempted ${Logging.simpleName(rightClass)} first). " +
+        s"Right failure: ${right.getMessage}, " +
+        s"Left failure: ${left.getMessage}")
 
   object UnsupportedContentTypeException {
     def apply(supported: ContentTypeRange*): UnsupportedContentTypeException = UnsupportedContentTypeException(Set(supported: _*))
--- a/akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/UnmarshallingSpec.scala
+++ b/akka-http-tests/src/test/scala/akka/http/scaladsl/unmarshalling/UnmarshallingSpec.scala
@@ -4,18 +4,22 @@
 
 package akka.http.scaladsl.unmarshalling
 
+import akka.http.scaladsl.unmarshalling.Unmarshaller.EitherUnmarshallingException
 import org.scalatest.{ BeforeAndAfterAll, FreeSpec, Matchers }
 import akka.http.scaladsl.testkit.ScalatestUtils
 import akka.actor.ActorSystem
 import akka.stream.ActorMaterializer
 import akka.http.scaladsl.model._
+import scala.concurrent.duration._
+
+import scala.concurrent.Await
 
 class UnmarshallingSpec extends FreeSpec with Matchers with BeforeAndAfterAll with ScalatestUtils {
   implicit val system = ActorSystem(getClass.getSimpleName)
   implicit val materializer = ActorMaterializer()
   import system.dispatcher
 
-  "The PredefinedFromEntityUnmarshallers." - {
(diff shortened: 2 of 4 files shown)


johanandren:
I can see why you choose the right hand to be the starting point but it feels backwards to me that 
`Unmarshal(...).to[Either[String, Int]]` does not try `String` first.

viktorklang:
Why? Doesn't it make sense to pick the Right thing first, then take what's
Left? ;)

On Mon, Apr 11, 2016 at 6:11 PM, Johan Andrén <notifications@github.com>
wrote:

> I can see why you choose the right hand to be the starting point but it
> feels backwards to me that
> Unmarshal(...).to[Either[String, Int]] does not try String first.
>
> —
> You are receiving this because you are on a team that was mentioned.
> <https://github.com/akka/akka/pull/20274#issuecomment-208429526>
>



-- 
Cheers,
√

johanandren:
Sometimes the right thing is not the right thing.

viktorklang:
Make the right thing great again.

On Mon, Apr 11, 2016 at 6:14 PM, Johan Andrén <notifications@github.com>
wrote:

> Sometimes the right thing is not the right thing.
>
> —
> You are receiving this because you are on a team that was mentioned.
> <https://github.com/akka/akka/pull/20274#issuecomment-208430627>
>



-- 
Cheers,
√

ktoso:
I always drive on the right side of the road, it's the english that do it on the wrong side.

ktoso:
Hilarious puns aside, WDYT – should we stray away from the "left is error-ish" and evaluate Left-then-Right?

johanandren:
I'd vote for that, but it seems I am already in minority.

rkuhn:
Then I’ll come to your aid: when reading the english text it is “either String or Int”, in that order.

rkuhn:
(Even though the resulting semantics may be surprising due to wanting the `Left` case be a catch-all type. Yes, this fact might invalidate the approach ;-) )

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
```diff
@@ -141,6 +142,15 @@ object Unmarshaller
...
     extends RuntimeException(supported.mkString("Unsupported Content-Type, supported: ", ", ", ""))
 
+  /** Order of parameters (`right` first, `left` second) is intentional, since that's the order we evaluate them in. */
+  final case class EitherUnmarshallingException(
+    rightClass: Class[_], right: Throwable,
+    leftClass: Class[_], left: Throwable)
+    extends RuntimeException(
+      s"Failed to unmarshal Either[${Logging.simpleName(leftClass)}, ${Logging.simpleName(rightClass)}] (attempted ${Logging.simpleName(rightClass)} first). " +
+        s"Right failure: ${right.getMessage}, " +
+        s"Left failure: ${left.getMessage}")
```
2beaucoup: Include `value` here?

ktoso: Hm... yeah worth considering, the error starts getting a bit long but maybe that's fine

ktoso: Since you're in this PR – any opinion on order of evaluating? Right Left or Left Right?

ktoso:
> Then I’ll come to your aid: when reading the English text it is “either String or Int”, in that order.
This helps.

> Even though the resulting semantics may be surprising due to wanting the Left case be a catch-all type. Yes, this fact might invalidate the approach ;-) 

... but this brings us back to square one again :wink:

ktoso:
If I count correctly we're 2:2 now?

johanandren:
I have a Dr on my side though.

ktoso:
Is he on the Right side though?

Ok, all jokes aside I guess Left then Right does make sense when one reads it, and it was also counter intuitive for me while implementing it. I corrected my thinking due to knowing how Either usually is used, however keyword "corrected" here, as intuitively I expected Left then Right when I first looked at it.

I'll rework the PR to do so.

akka-http/src/main/scala/akka/http/scaladsl/unmarshalling/Unmarshaller.scala
2beaucoup: I'd find it less intuitive if we switch so my vote would be to keep the current version.

ktoso: :confused: now I'm stuck... I like the Left being "fallback" too as witnessed here, feels more correct even if perhaps less "reads like english" hm.Will try to brainstorm this with the team in the morning.

ktoso: Thanks for your input!

2beaucoup: No prob. One more argument would be that the happy (right) state should me more frequent than the exceptional (left) one so trying right first should be more efficient as it has a higher probability of success. Just my 2 cents.

johanandren:
Could it be that the ambiguity points to having it in akka http not being a good idea, since how you expect it to work is different from person to person? I think we should prefer things that are more obvious in the API (and it is already a pretty big public surface area)

ktoso:
I'm going to merge this, it's useful as is and if people are confused about order of things we can revisit that.

johanandren:
LGTM (after the fact)

jmzhang:
I was trying to unmarshal a string to an either, but a bit surprised to discover the current eitherUnmarshaller is only available for HttpEntity. Any chance it could be made even more generic?

Issue #20271, opened 2016-04-10
akka-http: Strict-Transport-Security header should include "preload" optional field


atemerev:
As per https://developer.mozilla.org/en-US/docs/Web/Security/HTTP_strict_transport_security , there is a "preload" field which is not included yet in akka-http headers model. This leads to parse errors in akka-http websocket client code.

Can be encountered in the wild, e.g. on CloudFlare services.

Example (Websocket URI): wss://ws-feed.exchange.coinbase.com 

Proposed solution: update headers model in akka-http-core Java DSL and Scala DSL to include "preload" optional field.

atemerev:
Created pull request

ktoso:
Thanks a lot for the issue and PR, it's a good idea to fix this I agree.
We'll need to figure out if we're able to do so without breaking binary compatibility

jmhofer:
Actually, [RFC 6797](https://tools.ietf.org/html/rfc6797#section-6.1) says in 6.1:

> 5.  If an STS header field contains directive(s) not recognized by
>        the UA, the UA MUST ignore the unrecognized directives, and if
>        the STS header field otherwise satisfies the above requirements
>        (1 through 4), the UA MUST process the recognized directives.

This means that that this header should not only be extended to accept `preload`, it should be fixed so that it ignores other values it doesn't know about (instead of failing with a parse error).

atemerev:
OK — I don't have the expertise of writing it correctly with Shapeless/Parboiled2, but it should be parsed as one or more tokens separated by ';', where max-age=<int> is required, includeSubDomains is optional, and the rest (including preloaded) can be safely ignored.

atemerev:
Corrected the PR to maintain binary compatibility.

PR #20275, opened 2016-04-10
Added preloaded field to Strict-Transport-Security header

atemerev:
#20271 — Google extensions for only allowing strict TLS

https://developer.mozilla.org/en-US/docs/Web/Security/HTTP_strict_transport_security#Preloading_Strict_Transport_Security


ktoso:
Good idea, though won't be binary compatible in this way, need to find another way to expose this :(

ktoso:
OK TO TEST

akka-http-core/src/main/java/akka/http/javadsl/model/headers/StrictTransportSecurity.java
```diff
@@ -11,11 +11,12 @@
...
+    public abstract boolean preload();
 
     public static StrictTransportSecurity create(long maxAge) {
-        return new akka.http.scaladsl.model.headers.Strict$minusTransport$minusSecurity(maxAge, false);
+        return new akka.http.scaladsl.model.headers.Strict$minusTransport$minusSecurity(maxAge, false, false);
     }
-    public static StrictTransportSecurity create(long maxAge, boolean includeSubDomains) {
-        return new akka.http.scaladsl.model.headers.Strict$minusTransport$minusSecurity(maxAge, includeSubDomains);
+    public static StrictTransportSecurity create(long maxAge, boolean includeSubDomains, boolean preload) {
+        return new akka.http.scaladsl.model.headers.Strict$minusTransport$minusSecurity(maxAge, includeSubDomains, preload);
```
ktoso: we'll need to retain the previous method too (because binary compat):

```scala
public static StrictTransportSecurity create(long maxAge, boolean includeSubDomains) {
  return new akka.http.scaladsl.model.headers.Strict$minusTransport$minusSecurity(maxAge, includeSubDomains, false);
}
```

ktoso:
Something weird is going on here, this surely is not binary compatible but jenkins was happy hm... Need to investigate why

akka-http-core/src/main/scala/akka/http/impl/model/parser/SimpleHeaders.scala
```diff
@@ -187,7 +187,7 @@ private[parser] trait SimpleHeaders { this: Parser with CommonRules with CommonA
   def server = rule { products ~ EOI ~> (Server(_)) }
 
   def `strict-transport-security` = rule {
-    ignoreCase("max-age=") ~ `delta-seconds` ~ optional(ws(";") ~ ignoreCase("includesubdomains") ~ push(true)) ~ EOI ~> (`Strict-Transport-Security`(_, _))
+    ignoreCase("max-age=") ~ `delta-seconds` ~ optional(ws(";") ~ ignoreCase("includesubdomains") ~ push(true)) ~ optional(ws(";") ~ ignoreCase("preload") ~ push(true)) ~ EOI ~> (`Strict-Transport-Security`(_, _, _))
```
ktoso: hm, yes and no since order does not matter, we could get `preload; includeSubdomains`

atemerev:
Since "preload" is a Google extension and not in RFC (and doesn't need to be discriminating upon in any way), I can exclude it from the model. Just added "allow extra fields" parsing rule, just ignoring it as requested in RFC.

Still have somewhat fixed order of fields, though.

johanandren:
I think this needs some test coverage, could you please add, after that LGTM

ktoso:
LGTM after adding a test indeed 👍

drewhk:
I am not sure I fully get what this does here, so a test is welcome.

ktoso:
Would you be able to add a parser test for this change and then we'll merge @atemerev ?

atemerev:
@ktoso Yes, will do it today.

ktoso:
Awesome, thank you :)

atemerev:
@ktoso Sorry, couldn't do it on Friday, a little overload. :) Either tomorrow or Monday morning is the new ETA.

ktoso:
No problem, thanks in advance!

atemerev:
@ktoso Added parser test

akka-http-core/src/test/scala/akka/http/scaladsl/model/headers/HeaderSpec.scala
```diff
@@ -62,6 +62,18 @@ class HeaderSpec extends FreeSpec with Matchers {
     }
   }
 
+  "Strict-Transport-Security should" - {
+    "provide parseFromValueString method" - {
+      "successful parse run" in {
```
johanandren: Shouldn't there be a happy case without the optional part as well?

ktoso: Added that

ktoso:
Thanks again @atemerev !

ktoso:
Added testcase and merged: https://github.com/akka/akka/commit/8892eec62fb41441ea1a3ba40f0cb8820935ef89

PR #20279, opened 2016-04-11
=act reduce substring call by adding an additional startIndex

He-Pin:
do not need to call `rest = rest.substring(index + 2)` any more.

### DIFF
--- a/akka-actor/src/main/scala/akka/event/Logging.scala
+++ b/akka-actor/src/main/scala/akka/event/Logging.scala
@@ -1076,20 +1076,23 @@ trait LoggingAdapter {
   def format(t: String, arg: Any*): String = {
     val sb = new java.lang.StringBuilder(64)
     var p = 0
-    var rest = t
+    var startIndex = 0
     while (p < arg.length) {
-      val index = rest.indexOf("{}")
+      val index = t.indexOf("{}", startIndex)
       if (index == -1) {
-        sb.append(rest).append(" WARNING arguments left: ").append(arg.length - p)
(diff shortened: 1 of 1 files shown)


johanandren:
OK TO TEST

akka-actor/src/main/scala/akka/event/Logging.scala
```diff
@@ -1076,20 +1076,23 @@ trait LoggingAdapter {
...
-        sb.append(rest.substring(0, index)).append(arg(p))
-        rest = rest.substring(index + 2)
+        sb.append(t.substring(startIndex, index))
+          .append(arg(p))
+        startIndex = index + 2
         p += 1
       }
     }
-    sb.append(rest).toString
+    sb.append(t.substring(startIndex,t.length)).toString
```
patriknw: looks like this code is not formatted by scalariform? didn't you compile it?

patriknw:
LGTM

akka-actor/src/main/scala/akka/event/Logging.scala
hepin1989: I will try again.maybe I do the PR first and `sbt test` followed.

hepin1989: @patriknw I compiled it,but it still the same.

ktoso: It may be one of the files where scalariform freaks out and stops formatting :(

ktoso:
LGTM

Issue #16035, opened 2014-10-05
io: ClosedChannelException should be expected during Selector registration


jrudolph:
This can probably happen if a connection gets closed in the same moment as an interest is registered.

See https://groups.google.com/forum/#!topic/spray-user/PskLyvsFeOU

PR #20289, opened 2016-04-11
ignore ClosedChannelException when register selector, #16035

patriknw:


patriknw:
Refs #16035

johanandren:
LGTM

patriknw:
The test failure was. 

`[info] - must not thrown on unbind after system has been shut down *** FAILED *** (15 milliseconds)
[info]   akka.stream.StreamTcpException: Tcp command [Connect(localhost/127.0.0.1:35780,None,List(),None,true)] failed
`

I don't think it's related to this change, but I'm not sure. The intention of the change is to only suppress the warning logging. As far as I understood the exception was previously caught, logged and ignored.

patriknw:
PLS REBUILD

patriknw:
PLS BUILD

PR #20292, opened 2016-04-11
=doc fix #20291 by exposing imports better

### LINKED ISSUE #20291: Imports from 2nd example in Routing DSL Overview docs are not complete
The second example from [Routing DSL Overview](http://doc.akka.io/docs/akka/2.4.3/scala/http/routing-dsl/overview.html) docs, haven't compiled until I modified imports from: 

```
    import akka.http.scaladsl.server.Directives._
    import scala.io.StdIn
     
    object WebServer {
      def main(args: Array[String]) {
        // needed for the future flatMap/onComplete in the end
        implicit val executionContext = system.dispatcher
     
        val route =
          get {
            pathSingleSlash {
              complete(HttpEntity(ContentTypes.`text/html(UTF-8)`,"<html><body>Hello world!</body></html>"))
            } ~
              path("ping") {
                complete("PONG!")
              } ~
              path("crash") {
                sys.error("BOOM!")
              }
          }
     
        // `route` will be implicitly converted to `Flow` using `RouteResult.route2HandlerFlow`
        val bindingFuture = Http().bindAndHandle(route, "localhost", 8080)
        println(s"Server online at http://localhost:8080/\nPress RETURN to stop...")
        StdIn.readLine() // let it run until user presses return
        bindingFuture
          .flatMap(_.unbind()) // trigger unbinding from the port
          .onComplete(_ ⇒ system.terminate()) // and shutdown when done
      }
    }


```

to: 

```
import akka.actor.ActorSystem
import akka.http.scaladsl.model.{ContentTypes, HttpEntity}
import akka.http.scaladsl.server.Directives._
import scala.io.StdIn
```
So probably there is an error in documentation. However because I'm quite new to both, Scala and Akka, so I'm not sure if I have done everything right.

ktoso:
Resolves #20291

### DIFF
--- a/akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
+++ b/akka-docs/rst/scala/code/docs/http/scaladsl/HttpServerExampleSpec.scala
@@ -4,9 +4,7 @@
 
 package docs.http.scaladsl
 
-import akka.actor.{ Props, ActorRef, ActorSystem }
 import akka.event.LoggingAdapter
-import akka.http.scaladsl.Http
 import akka.http.scaladsl.Http.ServerBinding
 import akka.http.scaladsl.model._
 import akka.stream.ActorMaterializer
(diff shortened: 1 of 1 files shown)


johanandren:
LGTM

PR #20294, opened 2016-04-12
lazy usage of defaultAddress in ConsistentHashingRouter, #20263 (for validation)

patriknw:
- because consistent hashing routing pool is used by SimpleDnsManager
  and that can be activated early, before the transport defaultAddress
  is set in the startup

(cherry picked from commit 78505ccddb77562a0a8bf6c4b8dee3365c367d55)

Not sure how important it is to backport this to 2.3.x, but since the previous fix for the SimpleDnsManager was backported we can backport this also.


patriknw:
Refs #20263

Issue #20298, opened 2016-04-12
Resource leak upon early reply to 100-continue request


johanandren:
When a client sends a `Expect: 100-continue ` and the server responds directly, without reading the entity, with `Connection: close` as described by the Scaladoc of `emit100ContinueResponse` in `HttpServerBluePrint`, however if the client then closes the connection (as requested) the blueprint is deadlocked and will not complete.

The reason for this is that `HttpMessageParser` absorbs the upstream finish to be able to emit a `EntityStreamError(ErrorInfo(Entity stream truncation,))` downstream. It is however not possible to get demand that will trigger that downstream, as the `PrepareRequest` stage is in entity-stream mode and will set a flag `downstreamPullWaiting` and consume said demand.

PR #20300, opened 2016-04-13
WIP Http resource leak test coverage and fix

### LINKED ISSUE #17524: Use assertAllStagesStopped in Http tests

johanandren:
Fixes #17524 and #20298

### DIFF
--- a/akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerTestSetupBase.scala
+++ b/akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerTestSetupBase.scala
@@ -71,4 +71,13 @@ abstract class HttpServerTestSetupBase {
   def send(string: String): Unit = send(ByteString(string.stripMarginWithNewline("\r\n"), "UTF8"))
 
   def closeNetworkInput(): Unit = netIn.sendComplete()
+
+  def shutdownBlueprint(): Unit = {
+    netIn.sendComplete()
+    requests.expectComplete()
+
+    responses.sendComplete()
+    netOut.expectBytes(ByteString("HTT")) // ???
+    netOut.expectComplete()
+  }
 }
--- a/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
+++ b/akka-http-core/src/main/scala/akka/http/impl/engine/server/HttpServerBluePrint.scala
@@ -385,7 +385,10 @@ private[http] object HttpServerBluePrint {
               messageEndPending = false
               push(requestPrepOut, MessageEnd)
             case MessageStartError(status, info) ⇒ finishWithIllegalRequestError(status, info)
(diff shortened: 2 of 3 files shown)


akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerTestSetupBase.scala
```diff
@@ -71,4 +71,13 @@ abstract class HttpServerTestSetupBase {
   def send(string: String): Unit = send(ByteString(string.stripMarginWithNewline("\r\n"), "UTF8"))
 
   def closeNetworkInput(): Unit = netIn.sendComplete()
+
+  def shutdownBlueprint(): Unit = {
+    netIn.sendComplete()
+    requests.expectComplete()
+
+    responses.sendComplete()
+    netOut.expectBytes(ByteString("HTT")) // ???
```
johanandren: @ktoso Wat?

johanandren:
That was a suspciously quick CI run...

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerTestSetupBase.scala
ktoso: Well, looks like beginning of the `HTTP/1.1` response... most likely this should not happen but I have not read the sources in detail here yet

johanandren: Ok, then at least there was no "Well obviously ...". I will take another look at it before this PR is done.

johanandren: Arh. Already figured it out, I have used it extensively in tests not consuming the reply, so this is of course the start of the reply... :P sorry for pinging you with such a stupid miss.

2beaucoup: But why does this succeed even if there's no response sent at all?

johanandren:
Just a note to myself about this PR, it is waiting for me to look into the weird 'HTT' response in the shutdown.

akka-http-core/src/test/scala/akka/http/impl/engine/server/HttpServerTestSetupBase.scala
johanandren: I have opened a separate ticket #20377 to look at this so we can get the resource leak fix in. Pretty sure the `HTT` behavior of the server in this case wasn't really introduced here, just noticed so it should be safe to merge.

johanandren:
@ktoso @patriknw review thanks!

ktoso:
LGTM, thanks for opening the other ticket

PR #20261, opened 2016-04-08
first prototype of Aeron Source & Sink

patriknw:
- and test program to meassure latency and throughput

PR #20302, opened 2016-04-13
Proof-of-concept for the new Remoting codenamed Artery

drewhk:
**DO NOT MERGE**

This is a skeleton, PoC for the new remoting. The purpose of this code is only to be a starting point to bootstrap the new effort, adding new features and expanding on it continuously.

Next steps:
- FlightRecorder: we need this from day 1 and must extensively dogfood on it
- establish performance baseline: we need to track regressions every week, from day 1
- high-level connection management and error handling for the streams
  - we should try plugging Aeron in very early to see if we need to adjust things

### DIFF
--- a/akka-remote/src/main/resources/reference.conf
+++ b/akka-remote/src/main/resources/reference.conf
@@ -71,6 +71,12 @@ akka {
 
   remote {
 
+    ### FIXME: Temporary switch for the PoC
+    artery {
+      enabled = off
+      port = 20200
+    }
+
     ### General settings
 
     # Timeout after which the startup of the remoting subsystem is considered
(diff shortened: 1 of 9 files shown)


drewhk:
Nothing here is set in stone, this is not even a proposal, only a scaffolding, feel free to brainstorm.

patriknw:
Shall we create a `artery-dev` branch as target? I created `remote-dev` but I can rename that.

drewhk:
I am fine with whatever branch. Maybe artery-dev is less confusing as it refers to the codename of this feature compared to the generic "remote-dev".

patriknw:
that's a great start!

drewhk:
I think we should merge this soon, so we can fork out. What should be the branch? I open a new PR.

patriknw:
Let me merge the aeron stuff in https://github.com/akka/akka/pull/20261
and then I rename the branch

drewhk:
ok sounds good!

patriknw:
new branch pushed: `artery-dev`

drewhk:
MiMa complained about the missing `useUntrustedMode` method. We actually no longer use the method, and Remoting is not configurable since 2.3 so I don't expect anyone having a custom subclass of Remote. I think it is safe to ignore this bin-compat change.

drewhk:
opening a new and fix Mima...

PR #20304, opened 2016-04-13
Proof-of-concept for the new Remoting codenamed Artery

drewhk:


drewhk:
This is only a skeleton/scaffolding for future work.

akka-remote/src/main/scala/akka/remote/Remoting.scala
```diff
@@ -227,9 +227,6 @@ private[remote] class Remoting(_system: ExtendedActorSystem, _provider: RemoteAc
       s"Attempted to quarantine address [$remoteAddress] with uid [$uid] but Remoting is not running", null)
   }
 
-  // Not used anywhere only to keep compatibility with RemoteTransport interface
-  protected def useUntrustedMode: Boolean = provider.remoteSettings.UntrustedMode
-
```
patriknw: let's say we want to merge this into master later and release something together with Akka 2.4.x, should we try to not touch things that would break then?

project/MiMa.scala
```diff
@@ -714,6 +714,36 @@ object MiMa extends AutoPlugin {
...
+        ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.engine.client.PoolGateway.apply"),
+        ProblemFilters.exclude[IncompatibleMethTypeProblem]("akka.http.impl.engine.client.PoolGateway.this"),
+        ProblemFilters.exclude[MissingClassProblem]("akka.http.impl.engine.client.PoolGateway$NewIncarnation$"),
+        ProblemFilters.exclude[MissingClassProblem]("akka.http.impl.engine.client.PoolGateway$Running$"),
+        ProblemFilters.exclude[MissingClassProblem]("akka.http.impl.engine.client.PoolGateway$IsShutdown$"),
+        ProblemFilters.exclude[DirectMissingMethodProblem]("akka.http.impl.engine.client.PoolInterfaceActor.this"),
+        ProblemFilters.exclude[MissingClassProblem]("akka.http.impl.engine.client.PoolGateway$Running"),
+        ProblemFilters.exclude[MissingClassProblem]("akka.http.impl.engine.client.PoolGateway$IsShutdown"),
+        ProblemFilters.exclude[MissingClassProblem]("akka.http.impl.engine.client.PoolGateway$NewIncarnation"),
+        ProblemFilters.exclude[MissingClassProblem]("akka.http.impl.engine.client.PoolGateway$State"),
```
patriknw: why were these added?

akka-remote/src/main/scala/akka/remote/Remoting.scala
drewhk: what would break? I don't think anyone extended RemoteTransport since 2.4 at least (in 2.2, maybe also 2.3 you could still configure a different Remoting but this is not exposed anymore). If they did, what did they do with it? It is not possible to load it into Akka. I think this is fine to remove.

project/MiMa.scala
drewhk: hm, this is because my rebase I guess. This seem to have slipped through. The artery-dev branch was much earlier than what I forked from, the above seem to be the exclusions for the PoolGateway fixes merged recently. I will remove those.

akka-remote/src/main/scala/akka/remote/Remoting.scala
drewhk: (the above method is not called anywhere)

patriknw: ah, it's not possible, I didn't know, ok then

project/MiMa.scala
patriknw: thanks

patriknw:
LGTM, after the mima cleanup

akka-remote/src/main/scala/akka/remote/artery/ArterySubsystem.scala
```diff
@@ -0,0 +1,124 @@
...
+    // TODO: lookup subchannel
+    // FIXME: Use a different envelope than the old Send, but make sure the new is handled by deadLetters properly
+    queue.offer(Send(message, senderOption, recipient, None))
+  }
+
+  def quarantine(uid: Option[Int]): Unit = ()
+
+  // Idempotent
+  def associate(): Unit = {
+    queue = Source.queue(256, OverflowStrategy.dropBuffer).to(sink).run()(materializer)
```
hepin1989: make it configurable?

patriknw: We are doing a prototype. Things will of course be of Akka quality in the end.

drewhk:
Fixed MiMa

PR #20305, opened 2016-04-13
adjustments for Scala 2.12.0-M4

patriknw:


ktoso:
LGTM, had to do the same comparator manually today too ;)

johanandren:
LGTM

SethTisue:
the Comparator workaround code added here may not be necessary anymore as of 2.12.0-RC1, because scala/scala#5307

patriknw:
thanks for the update, @SethTisue (we'll leave it as is though)

